PR #4697, opened 2019-12-13
Set contextdir to current PWD if user did not specify a context dir.

rhatdan:
This matches the behavour of buildah bud.

Grabbing the contextdir based on the path to a Dockerfile, makes little sense.
I believe this is more likely what the user would expect.  If the context
directory was should be a different directory then the user must specify it.

The man page documentation actually conflicted on this.

Fixes: https://github.com/containers/libpod/issues/4695


rhatdan:
@vrothberg @mheon @baude @TomSweeneyRedHat @jwhonce PTAL

cmd/podman/build.go
```diff
@@ -186,7 +186,9 @@ func buildCmd(c *cliconfig.BuildValues) error {
 	if err != nil {
 		return nil
 	}
-	contextDir := ""
+	// Ignore error if os.Getwd fails. contextDir.
+	// contextDir could be updated later.
+	contextDir, _ := os.Getwd()
```
vrothberg: I would prefer putting it in the `else` case. In case we're not updating it, we would mask this error. It's unlikely to ever occur but it seems cleaner to not ignore it.

vrothberg: Maybe...
```golang
if len(cliArgs) == 0 {
   a,b := os.Getwed()
} else {
   //big body below
}

vrothberg:
Can we add a system test for that? It seems that's something we've regressed on already.

PR #4699, opened 2019-12-13
Fix F30-F31 migration for Podman 1.7.0

mheon:
The earlier attempt to re-add config migration only worked with user-specified configs (podman run --config). This version works more in line with that we want - the first rootless config file will be changed from runc to crun.

Verified on my system after an F31 migration - everything seems to be working well.


mheon:
We need this for the 1.7.0 release

rhatdan:
LGTM

PR #4710, opened 2019-12-16
Use systemd/sd-daemon.h headers for systemd presence

raukadah:
Finding systemd devel packages using libsystemd does not work as
in RHEL based distro the package name is systemd-devel and for
deb/ubunutu it is libsystemd. It is also giving false result when
podman rpm is built with systemd but hack/systemd_tag.sh does not
return anything.

Moving to systemd/sd-daemon.h header files which comes from devel
packages fixes the issue.

Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>

### DIFF
--- a/hack/systemd_tag.sh
+++ b/hack/systemd_tag.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
-if pkg-config --exists libsystemd; then
-    echo systemd
+cc -E - > /dev/null 2> /dev/null << EOF
+#include <systemd/sd-daemon.h>
+EOF
+if test $? -eq 0 ; then
+	echo systemd
 fi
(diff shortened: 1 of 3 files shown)


mheon:
LGTM but would like someone more familiar with shell to double-check

hack/systemd_tag.sh
```diff
@@ -1,4 +1,7 @@
-#!/usr/bin/env bash
-if pkg-config --exists libsystemd; then
-    echo systemd
+#!/bin/bash
```
sshnaidm: why not #!/usr/bin/env bash as it's in original patch?

raukadah:
@mheon @haircommander @sshnaidm Thank you for the review, For RHEL based system, we need to install system-devel, which is fixed in current patch.

PR #4712, opened 2019-12-16
Add `untag` sub-command

saschagrunert:
Podman now supports untagging images via the `untag` sub-command for the
root and `image` commands. Testing and documentation has been added as
well.

Replaces: https://github.com/containers/libpod/pull/4693 if agreed.

### DIFF
--- a/cmd/podman/untag.go
+++ b/cmd/podman/untag.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"github.com/containers/libpod/cmd/podman/cliconfig"
+	"github.com/containers/libpod/pkg/adapter"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
+	untagCommand cliconfig.UntagValues
+
+	_untagCommand = &cobra.Command{
+		Use:   "untag [flags] IMAGE [NAME...]",
+		Short: "Remove a name from a local image",
+		Long:  "Removes one or more names from a locally-stored image.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			untagCommand.InputArgs = args
+			untagCommand.GlobalFlags = MainGlobalOpts
+			untagCommand.Remote = remoteclient
+			return untag(&untagCommand)
+		},
+		Example: `podman untag 0e3bbc2
+  podman untag imageID:latest otherImageName:latest
+  podman untag httpd myregistryhost:5000/fedora/httpd:v2`,
+	}
+)
+
+func init() {
+	untagCommand.Command = _untagCommand
+	untagCommand.SetHelpTemplate(HelpTemplate())
+	untagCommand.SetUsageTemplate(UsageTemplate())
+}
+
+func untag(c *cliconfig.UntagValues) error {
+	args := c.InputArgs
+
+	if len(args) == 0 {
+		return errors.Errorf("at least one image name needs to be specified")
+	}
+
+	runtime, err := adapter.GetRuntime(getContext(), &c.PodmanCommand)
+	if err != nil {
+		return errors.Wrapf(err, "could not create runtime")
+	}
+	defer runtime.DeferredShutdown(false)
+
+	newImage, err := runtime.NewImageFromLocal(args[0])
+	if err != nil {
+		return err
+	}
+
+	tags := args[1:]
+	if len(args) == 1 {
+		// Remove all tags if not explicitly specified
+		tags = newImage.Names()
+	}
+	logrus.Debugf("Tags to be removed: %v", tags)
+
+	for _, tag := range tags {
+		if err := newImage.UntagImage(tag); err != nil {
+			return errors.Wrapf(err, "removing %q from %q", tag, newImage.InputName)
+		}
+	}
+	return nil
+}
--- a/cmd/podman/varlink/io.podman.varlink
+++ b/cmd/podman/varlink/io.podman.varlink
@@ -860,6 +860,12 @@ method PushImage(name: string, tag: string, compress: bool, format: string, remo
 # be found, an [ImageNotFound](#ImageNotFound) error will be returned; otherwise, the ID of the image is returned on success.
 method TagImage(name: string, tagged: string) -> (image: string)
 
+# UntagImage takes the name or ID of an image in local storage as well as the
+# tag name to be removed.  If the image cannot be found, an
+# [ImageNotFound](#ImageNotFound) error will be returned; otherwise, the ID of
+# the image is returned on success.
+method UntagImage(name: string, tag: string) -> (image: string)
+
 # RemoveImage takes the name or ID of an image as well as a boolean that determines if containers using that image
 # should be deleted.  If the image cannot be found, an [ImageNotFound](#ImageNotFound) error will be returned.  The
 # ID of the removed image is returned when complete.  See also [DeleteUnusedImages](DeleteUnusedImages).
--- a/docs/source/markdown/podman-untag.1.md
+++ b/docs/source/markdown/podman-untag.1.md
@@ -1,0 +1,37 @@
+% podman-untag(1)
+
+## NAME
+podman\-untag - Removes one or more names from a locally-stored image
+
+## SYNOPSIS
+**podman untag** *image*[:*tag*] [*target-names*[:*tag*]] [*options*]
+
+**podman image untag** *image*[:*tag*] [target-names[:*tag*]] [*options*]
+
+## DESCRIPTION
+Removes one or all names of an image. A name refers to the entire image name,
+including the optional *tag* after the `:`. If no target image names are
+specified, `untag` will remove all tags for the image at once.
+
+## OPTIONS
+
+**--help**, **-h**
+
+Print usage statement
+
+## EXAMPLES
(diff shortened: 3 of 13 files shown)


rhatdan:
Related to :https://github.com/containers/libpod/issues/4320

docs/source/markdown/podman-untag.1.md
```diff
@@ -0,0 +1,37 @@
...
+## NAME
+podman\-untag - Removes one or more names from a locally-stored image
+
+## SYNOPSIS
+**podman untag** *image*[:*tag*] *target-names*[:*tag*] [*options*]
+
+**podman image untag** *image*[:*tag*] *target-names*[:*tag*] [*options*]
+
+## DESCRIPTION
+Removes one or all aliases of an image. An alias refers to the entire image
```
vrothberg: "aliases" versus "names" - I prefer sticking with names to remain consistent

docs/source/markdown/podman-untag.1.md
```diff
@@ -0,0 +1,37 @@
+% podman-untag(1)
+
+## NAME
+podman\-untag - Removes one or more names from a locally-stored image
+
+## SYNOPSIS
+**podman untag** *image*[:*tag*] *target-names*[:*tag*] [*options*]
+
+**podman image untag** *image*[:*tag*] *target-names*[:*tag*] [*options*]
```
vrothberg: The usage here and in the commands help message implies that specifying at least on name is necessary but they are optional. Putting it in brackets would fix it.

cmd/podman/untag.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
+	untagCommand cliconfig.UntagValues
+
+	_untagCommand = &cobra.Command{
+		Use:   "untag [flags] IMAGE TAG [TAG...]",
```
vrothberg: The syntax seems wrong. `untag [flags] IMAGE [NAME:[TAG]]...`?

cmd/podman/untag.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return err
+	}
+
+	tags := args[1:]
+	if len(args) == 1 {
+		// Remove all tags if not explicitly specified
+		tags = newImage.Names()
+	}
+	logrus.Debugf("tags to be removed: %v", tags)
```
vrothberg: Logs should be capitalized.

cmd/podman/untag.go
```diff
@@ -0,0 +1,71 @@
...
+
+	for _, tag := range tags {
+		msg := fmt.Sprintf("removing %q from %q", tag, newImage.InputName)
+		logrus.Debug(msg)
```
vrothberg: Why not use `Debugf(..., x, y, z)`? Logs should be capitalized.

cmd/podman/untag.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := newImage.UntagImage(tag); err != nil {
```
vrothberg: I'd prefer `UntagImage` to take a string slice. This will speed things up, especially for the remote use case.

docs/source/markdown/podman-untag.1.md
saschagrunert: Yes, changed it to `name`

docs/source/markdown/podman-untag.1.md
saschagrunert: Fixed :+1:

cmd/podman/untag.go
saschagrunert: Hm, if we say that we reference a `name` to the full name including the tag, it would be
`untag [flags] IMAGE [NAME...]`, right?

cmd/podman/untag.go
saschagrunert: I'm creating a dedicated `string` because I wanted to report the failing debug message as error some lines below :)

cmd/podman/untag.go
saschagrunert: That's a breaking API change, should I still go for it?

cmd/podman/varlink/io.podman.varlink
```diff
@@ -855,6 +855,12 @@ method PushImage(name: string, tag: string, compress: bool, format: string, remo
 # be found, an [ImageNotFound](#ImageNotFound) error will be returned; otherwise, the ID of the image is returned on success.
 method TagImage(name: string, tagged: string) -> (image: string)
 
+# UntagImage takes the name or ID of an image in local storage as well as the
+# tag name to be removed.  If the image cannot be found, an
+# [ImageNotFound](#ImageNotFound) error will be returned; otherwise, the ID of
+# the image is returned on success.
+method UntagImage(name: string, tag: string) -> (image: string)
```
TomSweeneyRedHat: You don't mention 'flags' here...

cmd/podman/untag.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
+	untagCommand cliconfig.UntagValues
+
+	_untagCommand = &cobra.Command{
+		Use:   "untag [flags] IMAGE [NAME...]",
```
TomSweeneyRedHat: do `[flags]` still hold?  I don't see a mention below about it in the md.

saschagrunert: No I think we can remove it since `untag` in specific has no flags.

cmd/podman/varlink/io.podman.varlink
saschagrunert: Yep, see https://github.com/containers/libpod/pull/4712#discussion_r359198261

cmd/podman/untag.go
vrothberg: Oh, I didn't notice that. Let's not break the API. There will be bigger changes next year where we have the chance to clean things up a bit.

cmd/podman/untag.go
vrothberg: Ah, thanks! The issue with using the same string is that by convention logs should start with upper and errors with lowercase.

Since we're returning the first error, I think we don't need to log it.

cmd/podman/untag.go
saschagrunert: SGTM

cmd/podman/untag.go
saschagrunert: Okay, I removed the log and just returned the lowercase error.

docs/source/markdown/podman-untag.1.md
```diff
@@ -0,0 +1,37 @@
...
+**--help**, **-h**
+
+Print usage statement
+
+## EXAMPLES
+
+```
+$ podman untag 0e3bbc2
+
+$ podman untag imageID:latest otherImageName:latest
```
vrothberg: s/imageID/imageName/ ... using ${ID}:tag wouldn't work as ${ID} would be interpreted as a "name".

saschagrunert: Okay, fixed that as well :+1:

vrothberg:
The linter isn't happy.

saschagrunert:
> The linter isn't happy.

Thanks for the hint, updating the PR now.

saschagrunert:
I think I have to re-add the `[flags]` indicator to make the integration tests happy.

cmd/podman/untag.go
TomSweeneyRedHat: That works for me, thanks @saschagrunert

TomSweeneyRedHat: So I'm very confused at to why the linter is requiring '[flags]` here is we have not flags?  It's a very confusing usage message.

saschagrunert: The integration tests parse this to automatically do some iterative testing:
https://github.com/containers/libpod/blob/c1a7911f37d5b9791be68f3a9e6478771d565731/test/system/015-help.bats#L45-L63

I could probably work around this somehow.

cmd/podman/untag.go
```diff
@@ -0,0 +1,67 @@
...
+		Short: "Remove a name from a local image",
+		Long:  "Removes one or more names from a locally-stored image.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			untagCommand.InputArgs = args
+			untagCommand.GlobalFlags = MainGlobalOpts
+			untagCommand.Remote = remoteclient
+			return untag(&untagCommand)
+		},
+		Example: `podman unttag 0e3bbc2
```
NevilleC: s/unttag/untag

vrothberg:
Ping. Let's get this PR through.

cmd/podman/untag.go
saschagrunert: Fixed :+1:

saschagrunert:
> Ping. Let's get this PR through.

Rebased on top of the latest master branch and applied the review suggestion.

vrothberg:
Note that we can ignore the rdoproject failures; they won't block merging for now.

TomSweeneyRedHat:
LGTM

PR #4721, opened 2019-12-17
Update `tag` documentation regarding 'alias' usage

saschagrunert:
The word `alias` is not very common when speaking about image names and
tags. So we just refer to image name as the overall identifier of an
image.

Refers to https://github.com/containers/libpod/pull/4712#discussion_r358463230

### DIFF
--- a/cmd/podman/tag.go
+++ b/cmd/podman/tag.go
@@ -12,7 +12,7 @@ var (
 
 	tagDescription = "Adds one or more additional names to locally-stored image."
 	_tagCommand    = &cobra.Command{
-		Use:   "tag [flags] IMAGE TAG [TAG...]",
+		Use:   "tag [flags] IMAGE TARGET_NAME [TARGET_NAME...]",
 		Short: "Add an additional name to a local image",
 		Long:  tagDescription,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


cmd/podman/tag.go
```diff
@@ -12,7 +12,7 @@ var (
 
 	tagDescription = "Adds one or more additional names to locally-stored image."
 	_tagCommand    = &cobra.Command{
-		Use:   "tag [flags] IMAGE TAG [TAG...]",
+		Use:   "tag [flags] IMAGE NAME [NAME...]",
```
rhatdan: Should this be TARGET-NAME?  TO be consistent with the man page, or should the man page just reference NAME?

saschagrunert: Okay, I changed it to `TARGET_NAME` to keep it consistent.

rhatdan:
LGTM

cmd/podman/tag.go
```diff
@@ -12,7 +12,7 @@ var (
 
 	tagDescription = "Adds one or more additional names to locally-stored image."
 	_tagCommand    = &cobra.Command{
-		Use:   "tag [flags] IMAGE TAG [TAG...]",
+		Use:   "tag [flags] IMAGE TARGET_NAME [TARGET_NAME...]",
```
TomSweeneyRedHat: same comment here as in the other "tag" PR, `[flags]` here but no further word about them elsewhere?

saschagrunert:
Rebased on top of the latest master branch.

TomSweeneyRedHat:
There was a test error in package for centos and I don't know how to restart it?

PR #4722, opened 2019-12-17
libpod: drop arbitrary memory limit of 4M

giuseppe:
drop the arbitrary limit of 4M for creating a container.

4M are not enough for runc to run, on Fedora it seems to require at least 7M, and it is too much for crun that requires less than 1 Mb.

### DIFF
--- a/cmd/podman/shared/create_cli.go
+++ b/cmd/podman/shared/create_cli.go
@@ -12,11 +12,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	// It's not kernel limit, we want this 4M limit to supply a reasonable functional container
-	linuxMinMemory = 4194304
-)
-
 // GetAllLabels ...
 func GetAllLabels(labelFile, inputLabels []string) (map[string]string, error) {
 	labels := make(map[string]string)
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mheon PTAL

mheon:
Hm. My only concern would be that, from what I remember, the `runc` "I don't have enough memory" errors are horribly unintuitive - you'd never know that the memory limit was too low. Should we consider adding a warning for limits of, say, 5mb and under, that the OCI runtime may not have enough memory to finish completing the container? That would be really annoying, but otherwise people will be very confused about why things aren't working.

giuseppe:
> Hm. My only concern would be that, from what I remember, the `runc` "I don't have enough memory" errors are horribly unintuitive - you'd never know that the memory limit was too low

one issue is that the 4Mb limit doesn't help much with runc, as it requires at least 7Mb now.

rhatdan:
We could make it a logrus.Warnf statement.  Then it would not show unless the user stated --log-level.
Also should we special case crun, to not need the message?

TomSweeneyRedHat:
I this configurable from libpod.conf?  If not, I think it should be.

giuseppe:
personally I don't think we should try to prevent runtime failures from Podman.  It should be the OCI runtime itself to show a better error message.

Another example:

```
# podman --runtime /usr/bin/runc run --rm  --pids-limit 1 alpine echo hi
Error: container_linux.go:346: starting container process caused "process_linux.go:319: getting the final child's pid from pipe caused \"EOF\"": OCI runtime error
```

mheon:
Fair enough. LGTM then.

Issue #4704, opened 2019-12-15
Non-stopping containers corrupt internal state with  "pod rm -f" 


ensc:
/kind bug

**Description**

When a container does not get stopped by `podman pod rm -f` information about it are not completely released and e.g. its name can not be reused anymore.  Such containers do not show up in "podman ps -a" and can not be removed.  Their`slurp4netns` network stays alive too and blocks reuse of mapped ports.

**Steps to reproduce the issue:**

1.  `podman pod create --name test1`

2.  add a container which needs some time for shutdown; e.g. `postgres` 

     ```
     podman run -d --pod test1 --name c1 postgres
     ```

3.  remove the pod forcefully and wait

     ```
     podman pod rm -f test1
     ```

     After some time you will see "cannot remove container ... did not die within timeout"

4.  repeat step 1 + 2

**Describe the results you received:**

Step 4.2 errors with

```
Error: error creating container storage: the container name "c1" is already in use by "ffa8ebb5bc118a5e3b5b872ba7eb76edd471d74bb8049a840f844d6828e1f0a2". You have to remove that container to be able to reuse that name.: that name is already in use
```

Removing it does not work either:

```
$ podman rm -f c1
Error: Failed to evict container: "": Failed to find container "c1" in state: no container with name or ID c1 found: no such container
$ podman rm -f ffa8ebb5bc118a5e3b5b872ba7eb76edd471d74bb8049a840f844d6828e1f0a2
Error: Failed to evict container: "": Failed to find container "ffa8ebb5bc118a5e3b5b872ba7eb76edd471d74bb8049a840f844d6828e1f0a2" in state: no container with name or ID ffa8ebb5bc118a5e3b5b872ba7eb76edd471d74bb8049a840f844d6828e1f0a2 found: no such container
```

A `podman ps -a` shows a running infra container but not `c1`.  

**Describe the results you expected:**

Name `c1` can be used again

**Output of `podman version`:**

```
$ podman version
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.1
  podman version: 1.6.2
host:
  BuildahVersion: 1.11.3
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.2-1.fc31.x86_64
    path: /usr/bin/conmon
[...]

PR #4723, opened 2019-12-18
Remove volumes after containers in pod remove

mheon:
When trying to reproduce #4704 I noticed that the named volumes from the Postgres containers in the reproducer weren't being removed by `podman pod rm -f` saying that the container they were attached to was still in use. This was rather odd, considering they were only in use by one container, and that container was in the process of being removed with the pod.

After a bit of tracing, I realized that the cause is the ordering of container removal when we remove a pod. Normally, it's done in removeContainer() before volume removal (which is the last thing in that function). However, when we are removing a pod, we remove containers all at once, after removeContainer has already finished - meaning the container still exists when we try to remove its volumes, and thus the volume can't be removed.

Solution: collect a list of all named volumes in use by the pod, and remove them all at once after every container in the pod is gone. This ensures that there are no dependency issues.

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -573,7 +573,7 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force bool,
 			if !volume.IsCtrSpecific() {
 				continue
 			}
-			if err := runtime.removeVolume(ctx, volume, false); err != nil && err != define.ErrNoSuchVolume && err != define.ErrVolumeBeingUsed {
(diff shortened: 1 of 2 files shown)


mheon:
CI is green. Github, however, is complaining that commits are not signed, even though they are? Little confusing.

rhatdan:
I fixed the github issue.
LGTM
@giuseppe @haircommander @saschagrunert @nalind @TomSweeneyRedHat PTAL

haircommander:
LGTM

PR #4726, opened 2019-12-18
libpod: do not use kill --all with PID namespace

giuseppe:
if the container has a PID namespace, there is no need to use kill
--all and use cgroups to read the list of processes but it is enough
to kill the main process.

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -463,7 +463,8 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force bool,
 
 	// Check that the container's in a good state to be removed
 	if c.state.State == define.ContainerStateRunning {
-		if err := c.stop(c.StopTimeout(), true); err != nil {
+		all := !HasNamespace(c.config.Spec, spec.PIDNamespace)
+		if err := c.stop(c.StopTimeout(), all); err != nil {
 			return errors.Wrapf(err, "cannot remove container %s as it could not be stopped", c.ID())
 		}
 	}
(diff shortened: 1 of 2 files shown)


rhatdan:
@giuseppe I wrote a similar PR, https://github.com/containers/libpod/pull/4727
I think we should merge the two.  The question I have is whether or not we should remove the option from c.stop() and have the function decide whether to send signal to all processes or just the primary.

I think running podman stop on a container that is not in a PID namespace is also broken.

mheon:
I'm in favor of keeping it.

For a test of what stop should do, running a Docker container with pid=host
and an exec session in that container, then stopping, should tell us. If
the exec session dies we should use --all for stop, otherwise not.

rhatdan:
You meant a Podman container. :^)

 If the exec joins the PID Namespace then it should also exit when the PID namespace goes away.

The problem is we will have to put the same code all over the place to c.stop(TIMEOUT, !hasPIDNS) 

Currently we are broken in multiple places when sharing the PID Namespace with the host.

mheon:
It looks like Docker stops all exec sessions even with `--pid=host` - based on that, agree that we should force `--all` with stop as well.

giuseppe:
I don't think dropping the `-all` changes how exec sessions are handled.

When running with a PID namespace, if the PID 1 in the container is killed, then all the processes are also killed regardless if they are exec sessions or not.

The only difference is when when we sent a signal != SIGKILL.  I still think it should be the PID 1 responsibility to propagate it, if necessary.

Another disadvantage with using `kill --all` when not needed is that it uses the freezer cgroup to suspend temporarily the container in order to avoid races (i.e. processes can be created/destroyed while the runtime sends them the signal).  An async operation like sending a signal turns into suspending the container.

rhatdan:
@giuseppe You ok with going with my version?  If not reopen PR.

Issue #4738, opened 2019-12-21
Completion in zsh


st1971:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Command completion not working as expected when using the zsh shell.

**Steps to reproduce the issue:**

1. type 'podman run'

2. hit tab key

3.

**Describe the results you received:**
3 Lines of errors.
```
_arguments:comparguments:325: invalid argument: Note:[use the environment variable 'TMPDIR' to change the temporary storage location for container images, '/var/tmp'.]
_arguments:comparguments:325: invalid argument: Note:[use the environment variable 'TMPDIR' to change the temporary storage location for container images, '/var/tmp'.]
_arguments:comparguments:325: invalid argument: Note:[use the environment variable 'TMPDIR' to change the temporary storage location for container images, '/var/tmp'.]
```

**Describe the results you expected:**
List of images

**Additional information you deem important (e.g. issue happens only occasionally):**

Problem seems to be being caused by a change to the flags section of --help output.
```
      --tmpdir string             Path to the tmp directory for libpod state content.
                                  
                                  Note: use the environment variable 'TMPDIR' to change the temporary storage location for container images, '/var/tmp'.
                                  
      --trace                     Enable opentracing output
```
The blank lines and Note were not in the help output in previous versions.  The completion script needs an additional sed statement to remove the additional notes and blank line.

Have attached a patch file with the changes that I have made locally to make this work for me.
[completions.patch.txt](https://github.com/containers/libpod/files/3991876/completions.patch.txt)

**Output of `podman version`:**

```
Version:            1.7.0-rc1
RemoteAPI Version:  1
Go Version:         go1.13.5
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.5
  podman version: 1.7.0-rc1
host:
  BuildahVersion: 1.11.6
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.2-1.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.2, commit: 186a550ba0866ce799d74006dab97969a2107979'
  Distribution:
    distribution: fedora
    version: "31"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 200000
      size: 65536
    uidmap:
    - container_id: 0
[...]

PR #4475, opened 2019-11-07
Add missing information to podman.1 man page

rhatdan:
Add missing options.
Define the use of the TMPDIR environment variable for the storing of container images
when they are being pulled to the system.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #4740, opened 2019-12-23
zsh completion: ignore multi-line output in Flags

edsantiago:
PR #4475 introduced an interesting twist on --help: a help
string that spans multiple lines. This broke zsh completion.

I'm not keen on that multi-line output, but it shouldn't
break completion. Fix is simple: look only for flag lines
beginning with '-', filter out anything else.

Fixes: #4738


rhatdan:
LGTM

rhatdan:
@mheon @saschagrunert @giuseppe @vrothberg @QiWang19 @TomSweeneyRedHat PTAL

edsantiago:
Not my catch; all credit to @st1971. @vrothberg you seem to be unclear on the concept of PTO.

PR #4743, opened 2019-12-23
add struct response for removal of images

baude:
when removing an image from storage, we should return a struct that
details what was untagged vs deleted.  this replaces the simple
println's used previously and assists in API development.

### DIFF
--- a/API.md
+++ b/API.md
@@ -149,6 +149,8 @@ in the [API.md](https://github.com/containers/libpod/blob/master/API.md) file in
 
 [func RemoveImage(name: string, force: bool) string](#RemoveImage)
 
+[func RemoveImageWithResponse(name: string, force: bool) RemoveImageResponse](#RemoveImageWithResponse)
+
 [func RemovePod(name: string, force: bool) string](#RemovePod)
 
 [func Reset() ](#Reset)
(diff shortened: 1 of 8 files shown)


rhatdan:
A couple of issues with this.  
1. You are changing Lipod API, does this mean we need to version bump.
2.  Can an image be deleted and untagged, if yes, would a user want this to be reported twice?
3.  Would scripts that were built to use old interface now break because of the special Untagged, Deleted flags?

baude:
@rhatdan responses per question

1. I dont think so personally ...
2. The behaviour of what is printed is not changed at all ... just simply how it gets printed.  Whatever we did is what we do now, which emulates docker.
3. No, due to the response of number 2.  Output did not change at all.

baude:
if folks are really worried about @rhatdan 's number one question there, I could make a wrapper method so that nothing would change in case anyone was using it.  But I know @mheon hated our existing code anyway.

vrothberg:
> if folks are really worried about @rhatdan 's number one question there, I could make a wrapper method so that nothing would change in case anyone was using it. But I know @mheon hated our existing code anyway.

I am okay with breaking the _golang_ API. I mostly care about a stable _varlink_ API at the moment. Release management will get easier when we split libpod and Podman.

PR #4745, opened 2019-12-23
[CI:DOCS]add volume endpoints for libpod

baude:
create, inspect, ls, prune, and rm


baude:
can someone crush the lgtm?

Issue #4746, opened 2019-12-23
(minor) podman kill: should do input validation


edsantiago:
Looks like podman is accepting any int, casting to ulong, and passing it along:
```console
# podman run -d --name foo alpine sleep 60
5eafd476443109ab52af53cc68a08e81f3eb7a2249c81993193b01590cd1d5b2
# podman kill -s -1 foo
ERRO[0000] unknown signal "18446744073709551615"
unknown signal "18446744073709551615"
Error: error sending signal to container 5eafd476443109ab52af53cc68a08e81f3eb7a2249c81993193b01590cd1d5b2: `/usr/bin/runc kill 5eafd476443109ab52af53cc68a08e81f3eb7a2249c81993193b01590cd1d5b2 18446744073709551615` failed: exit status 1
# podman kill -s 999 foo
ERRO[0000] container_linux.go:389: signaling init process caused "invalid argument"
container_linux.go:389: signaling init process caused "invalid argument"
Error: error sending signal to container 5eafd476443109ab52af53cc68a08e81f3eb7a2249c81993193b01590cd1d5b2: `/usr/bin/runc kill 5eafd476443109ab52af53cc68a08e81f3eb7a2249c81993193b01590cd1d5b2 999` failed: exit status 1

------
# [ expected something more like ]:
Error: Invalid signal: -1 (or 999)
```
Would it make sense to validate against `SIGRTMAX` before accepting it? 

And, for the sake of discussion, I feel it might be a nice courtesy to accept a signed int and run `abs()` on it. Those of us with fingers trained to type `kill -N` would appreciate it. 

Oh, also, `kill -s 0` throws an error (invalid signal). This is what docker does, so we might need to keep it for compatibility, but it could also be nice to replicate standard `kill -0` behavior (exit 0 if container exists).

rhatdan:
I like to get a fix on all of these.

Anyone out there looking for a quick git project should grab this.

PR #4749, opened 2019-12-26
signal parsing - better input validation

edsantiago:
The helper function we use for signal name mapping does not
check for negative numbers nor invalid (too-high) ones. This
can yield unexpected error messages:
```
   # podman kill -s -1 foo
   ERRO[0000] unknown signal "18446744073709551615"
```
This PR introduces a small wrapper for it that:

  1) Strips off a leading dash, allowing '-1' or '-HUP'
     as valid inputs; and
  2) Rejects numbers <1 or >64 (SIGRTMAX)

Also adds a test suite checking signal handling as well as
ensuring that invalid signals are rejected by the command line.

Fixes: #4746

### DIFF
--- a/cmd/podman/kill.go
+++ b/cmd/podman/kill.go
@@ -3,7 +3,7 @@ package main
 import (
 	"github.com/containers/libpod/cmd/podman/cliconfig"
 	"github.com/containers/libpod/pkg/adapter"
-	"github.com/docker/docker/pkg/signal"
+	"github.com/containers/libpod/pkg/util"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -54,7 +54,7 @@ func killCmd(c *cliconfig.KillValues) error {
 
 	// Check if the signalString provided by the user is valid
 	// Invalid signals will return err
-	killSignal, err := signal.ParseSignal(c.Signal)
+	killSignal, err := util.ParseSignal(c.Signal)
 	if err != nil {
 		return err
 	}
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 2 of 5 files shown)


edsantiago:
I've used this as an exercise in Go namespaces; and obviously I'm still missing something because I can't figure out how to import `sigrtmax` from `github.com/docker/docker/pkg/signal/signal_linux.go`. I would prefer not to merge this PR with a hardcoded `64`. Suggestions appreciated.

pkg/util/utils.go
```diff
@@ -305,6 +304,21 @@ func GetImageConfig(changes []string) (ImageConfig, error) {
...
+// Parse and validate a signal name or number
+func ParseSignal(rawSignal string) (syscall.Signal, error) {
+	// Strip off leading dash, to allow -1 or -HUP
+	basename := strings.TrimPrefix(rawSignal, "-")
+
+	signal, err := signal.ParseSignal(basename)
+	if err != nil {
+		return -1, err
+	}
+	if 1 <= signal && signal <= 64 { // FIXME: hardcoding SIGRTMAX
```
mheon: I don't see `SIGRTMAX` in the x/sys/unix library, which makes me think that it's not defined anywhere in official Go libraries. We may have to leave it here.

pkg/util/utils.go
```diff
@@ -305,6 +304,21 @@ func GetImageConfig(changes []string) (ImageConfig, error) {
...
+		return signal, nil
+	}
+	return -1, fmt.Errorf("Invalid signal: %s", rawSignal)
```
mheon: We generally prefer `errors.Errorf()` to `fmt.Errorf()` - they work identically, but the errors alias is used more consistently elsewhere

mheon:
Importing `sigrtmax` is impossible as it's lower-cased - Go's visibility rules are that functions and variables starting with a Capital are exported and available outside the package, lower-case ones are private and unusable.

mheon:
/approve

Found a nit but otherwise LGTM

pkg/util/utils.go
mheon: Also, the linter is complaining that Invalid is capitalized, Go wants error messages to start with a lowercase letter

edsantiago:
@mheon thanks for the advice. `fmt.Errorf` changed to `errors.Errorf`. Capitalization fixed, and since my only reason for capitalizing was to keep the message identical to docker's, I changed it to something I believe is more helpful anyway. The new error message required some tweaks to the test code.

edsantiago:
I have no idea what those failures mean. Giving up for today.

cmd/podman/kill.go
```diff
@@ -3,7 +3,7 @@ package main
 import (
 	"github.com/containers/libpod/cmd/podman/cliconfig"
 	"github.com/containers/libpod/pkg/adapter"
-	"github.com/docker/docker/pkg/signal"
+	"github.com/containers/libpod/pkg/util"
```
rhatdan: Always happy to see this.

PR #4751, opened 2019-12-26
The --quiet flag does not conflict with templates in ps

mheon:
To match Docker behavior, make `--quiet` and `--format` with a Go template not conflict. Instead, just turn off `--quiet` in such cases, as we'll be using Go template output instead.


edsantiago:
/lgtm
/hold
[ remove hold if/when ready to merge ]

rhatdan:
LGTM
/hold cancel

PR #4755, opened 2019-12-28
Fix race condition in kill test leading to hang

edsantiago:
When you open a FIFO for reading, but there's no writer, you hang.
This is just one of those obscure UNIXisms we all know but just
forget all too often.

My last PR was guilty of introducing such a condition; I caught
it by accident while testing other stuff. In short, the signal
container was doing 'echo DONE' as its last step, and we (BATS)
were reading the FIFO to check for it; but if the container
exited before we opened the FIFO for read, the open would hang.
This is not a hang that we can catch in the test: it would hang
the entire job forever.

Solution: there's no need for the DONE. Just remove it and
the check. (For future reference, another option would be
to 'exec 5<$fifo' immediately after spawning the log reader,
and using 'read -u 5' instead of 'read <$fifo', but that's
unwieldy and hard to maintain.)

### DIFF
--- a/test/system/130-kill.bats
+++ b/test/system/130-kill.bats
@@ -8,7 +8,7 @@ load helpers
 @test "podman kill - test signal handling in containers" {
     # Start a container that will handle all signals by emitting 'got: N'
     local -a signals=(1 2 3 4 5 6 8 10 12 13 14 15 16 20 21 22 23 24 25 26 64)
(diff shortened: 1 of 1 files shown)


edsantiago:
This approach is imperfect: if podman/runc/crun/alpine signal handling were ever to break, we still have a risk of a hang. Closing in favor of #4756 which is slightly more complex but produces much better diagnostics.

PR #4757, opened 2019-12-28
refactor network commands

baude:
move core of network commands from pkg/adapter to pkg/network to assist
with api development and remote podman commands.

### DIFF
--- a/pkg/adapter/network.go
+++ b/pkg/adapter/network.go
@@ -67,13 +67,9 @@ func (r *LocalRuntime) NetworkInspect(cli *cliconfig.NetworkInspectValues) error
 		rawCNINetworks []map[string]interface{}
 	)
 	for _, name := range cli.InputArgs {
-		b, err := network.ReadRawCNIConfByName(name)
+		rawList, err := network.InspectNetwork(name)
 		if err != nil {
 			return err
-		}
-		rawList := make(map[string]interface{})
-		if err := json.Unmarshal(b, &rawList); err != nil {
-			return fmt.Errorf("error parsing configuration list: %s", err)
 		}
 		rawCNINetworks = append(rawCNINetworks, rawList)
 	}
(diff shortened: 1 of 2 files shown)


pkg/network/network.go
```diff
@@ -148,3 +150,43 @@ func ValidateUserNetworkIsAvailable(userNet *net.IPNet) error {
 	}
 	return nil
 }
+
+// RemoveNetwork removes a given network by name.  If the network has container assoociated with it, that
```
NevilleC: s/assoociated/associated

pkg/adapter/network.go
```diff
@@ -98,7 +94,22 @@ func (r *LocalRuntime) NetworkRemove(ctx context.Context, cli *cliconfig.Network
 		if err != nil {
 			return networkRmSuccesses, networkRmErrors, err
 		}
-		if err := r.removeNetwork(ctx, name, containers, cli.Force); err != nil {
+		// We need to iterate containers looking to see if they belong to the given network
+		for _, c := range containers {
+			if util.StringInSlice(name, c.Config().Networks) {
+				// if user passes force, we nuke containers
+				if cli.Force {
```
NevilleC: maybe this could be happy path

```go
if !cli.Force {
        // Without the force option, we return an error
        return nil, nil, errors.Errorf("%q has associated containers with it. use -f to forcibly delete containers", name)
}

if err := r.RemoveContainer(ctx, c.Container, true, true); err != nil {
    return nil, nil, err
}
```

pkg/adapter/network.go
```diff
@@ -98,7 +94,22 @@ func (r *LocalRuntime) NetworkRemove(ctx context.Context, cli *cliconfig.Network
...
+					if err := r.RemoveContainer(ctx, c.Container, true, true); err != nil {
+						return nil, nil, err
+					}
+				} else {
+					// Without the the force option, we return an error
```
NevilleC: s/the the/the

pkg/adapter/network.go
```diff
@@ -98,7 +94,22 @@ func (r *LocalRuntime) NetworkRemove(ctx context.Context, cli *cliconfig.Network
...
+					return nil, nil, errors.Errorf("%q has associated containers with it. use -f to forcibly delete containers", name)
```
NevilleC: s/use/Use

baude:
this is blocking apiv2 work.  please review and merge

pkg/network/network.go
```diff
@@ -148,3 +150,43 @@ func ValidateUserNetworkIsAvailable(userNet *net.IPNet) error {
...
+	}
+	// Remove the configuration file
+	if err := os.Remove(cniPath); err != nil {
+		return errors.Wrapf(err, "failed to remove network configuration file %q", cniPath)
+	}
+	return nil
+}
+
+// InspectNetwork reads a CNI config and returns its configuration
+func InspectNetwork(name string) (map[string]interface{}, error) {
```
mheon: Can we return anything better than this? I know that CNI doesn't bother exporting most types, but can we potentially provide some sort of struct defining what contents could be?

mheon:
@baude I mean, it is Sunday of shutdown :smile:

pkg/network/network.go
baude: It is really too loosely defined to do much more.  I don't like it either.

mheon:
Ack. LGTM then.

PR #4758, opened 2019-12-29
Don't show PASS on success for gitvalidate

rhatdan:
Make test logs less verbose so that failures are easier to spot.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ export GOPROXY=https://proxy.golang.org
 
 GO ?= go
 DESTDIR ?=
-EPOCH_TEST_COMMIT ?= 2d8f1c8fda4d0a8b7c41addbc89f0b2f83908ec2
+EPOCH_TEST_COMMIT ?= $(shell git merge-base HEAD $${DEST_BRANCH:-master})
 HEAD ?= HEAD
 CHANGELOG_BASE ?= HEAD~
 CHANGELOG_TARGET ?= HEAD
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes: https://github.com/containers/libpod/issues/4581
@edsantiago PTAL

rhatdan:
@ssbarnea PTAL

rhatdan:
Ok, I thought the pipefail would exit with != 0.

edsantiago:
It does, but remember that `make` executes every separate line in a separate shell; so a lone `set -o pipefail` on its own line will not affect the next one. Very quick testing on my end shows this to be more promising:
```make
    set -o pipefail; GIT_CHECK_EXCLUDE=...
```

edsantiago:
Long rambling notes ahead, caveat lector.

First, it turns out that `vbatts/git-validation` has both `-v` and `-q` flags, and we're invoking it with `-v`. That goes a long way toward explaining the verbosity. After playing with removing `-v` and/or adding `-q`, I think the most helpful output comes from simply removing `-v`, **not** adding `-q`, and changing the `grep` in your PR to:
```
          ... | { egrep -v ' PASS$$' || test $$? = 1; }
```
(that is, use `egrep` instead of grep, and then single-quote, space, PASS, dollar, dollar, single quote). The reason for that is that just grepping out `PASS` would filter any commit message containing the string PASS - such as this one.

Going further, I think the ultimate root cause here is the hardcoding of `EPOCH_TEST_COMMIT`. Here's another of my stupid questions: why not just use `master`?
```make
.gitvalidation: .gopathok
	GIT_CHECK_EXCLUDE="./vendor:docs/make.bat" $(GOBIN)/git-validation -run DCO,short-subject,dangling-whitespace -range master..$(HEAD)
```
This produces clean concise readable results, and doesn't make me as nervous as that `pipefail - grep` chain.

edsantiago:
...or, on reflection, `master^` to better handle merges?

edsantiago:
How about this (on `Makefile:6`) and removing the `-v` on the `git-validation` call (line 470)?
```make
EPOCH_TEST_COMMIT ?= $(shell git merge-base HEAD $${DEST_BRANCH:-master})
```

rhatdan:
@edsantiago Fixed
@mheon PTAL

edsantiago:
Oh - what I was suggesting is that with the `git merge-base` you no longer need the `pipefail` nor `grep -v`

edsantiago:
LGTM! Before merging: @cevich, could you take a look at the `EPOCH` change and give it your blessing?

cevich:
>  could you take a look at the EPOCH change and give it your blessing?

I was too late/slow :cry: I think my concerns were minor though, given that we already run `git` all over the place in the Makefile and nothing broke before this merged :smile:

PR #4766, opened 2019-12-30
[CI:DOCS]more apiv2 development

baude:
misc population of methods, etc


pkg/api/server/register_info.go
```diff
@@ -3,6 +3,7 @@ package server
 import (
 	"github.com/containers/libpod/pkg/api/handlers/generic"
 	"github.com/gorilla/mux"
+	"net/http"
```
rhatdan: Wrong place

pkg/api/server/register_ping.go
```diff
@@ -3,13 +3,14 @@ package server
 import (
 	"github.com/containers/libpod/pkg/api/handlers/generic"
 	"github.com/gorilla/mux"
+	"net/http"
```
rhatdan: Wrong place

baude:
rebased and fixed up

PR #4768, opened 2019-12-30
Enable shell completion for podman commands.

### LINKED ISSUE #3878: bash-completion: list files for `podman build -f`
/kind feature

`podman build -f`[TAB] doesn't list files or completes paths. Having that would make it easier to use. Likely, other commands could benefit from that as well but I did not examine that yet.

NevilleC:
I added a command completion that generates
the shell scripts that would enable the shell
completion, using the cobra function `GenBash/ZshCompletion`.

Closes #3878

```shell
[vagrant@localhost libpod]$ podman build -f Dockerfile[tab][tab]
Dockerfile         Dockerfile.centos  Dockerfile.fedora
```
Signed-off-by: Neville Cain <neville.cain@qonto.eu>

### DIFF
--- a/cmd/podman/shell_auto_completion.go
+++ b/cmd/podman/shell_auto_completion.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	bashAutoCompletionDescription = `To load completion run
+
+podman completion bash >>~/.bashrc
+
+Reload your shell.
+
+To make it available in all your bash session, add this to your ~/.bashrc file:
+echo 'source <(podman completion bash)' >>~/.bashrc
+
+As an alternative, if 'bash-completion' is installed on your system, you can add it in:
+/etc/bash_completion.d/{your_file_name}
+`
+)
+
+var _bashCompletionCommand = &cobra.Command{
+	Use:   "completion",
+	Short: "Generates shell scripts for auto-completion",
+	// We do not intend to show this command to the user so
+	// we marked it as hidden. We should use them by using
+	// the "make completion" target to update the shell scripts
+	// enabling the auto-completion for podman.
+	Hidden:  true,
+	Example: `podman completion --help`,
+}
+
+func init() {
+	_bashCompletionCommand.AddCommand(
+		&cobra.Command{
+			Use:   "bash",
+			Short: "generate auto-completion for bash",
+			Long:  bashAutoCompletionDescription,
+			RunE: func(cmd *cobra.Command, args []string) error {
+				// Writing the shell script to stdout allows the most flexible use
+				// as the user can redirect the outputt where it needs it.
+				return rootCmd.GenBashCompletion(os.Stdout)
+			},
+			Example: `podman completion bash >>~/.bashrc`,
+		},
+	)
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,6 +103,13 @@ Well, you can now create your own branch, apply changes on it, and then submitti
 
 For further reading about branching [you can read this document](https://herve.beraud.io/containers/linux/podman/isolate/environment/2019/02/06/how-to-hack-on-podman.html).
 
+### Adding/Updating a new command
+
+When you add/update a new command, dont't forget to run 'make completion' to update
+the bash completion script.
+
+There is no need to update zsh completion script as it autogenerates completion.
+
 ## Submitting Pull Requests
 
 No Pull Request (PR) is too small! Typos, additional comments in the code,
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,9 @@ CROSS_BUILD_TARGETS := \
 
 all: binaries docs
 
+completion: binaries
+	bin/podman completion bash > completions/bash/podman
+
 default: help
 
 define PRINT_HELP_PYSCRIPT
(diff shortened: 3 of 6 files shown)


NevilleC:
WIP - just opening it to show I would like to take https://github.com/containers/libpod/issues/3878.

It seems to work but it needs to be refined. Feel free to review but keep this in mind, please.

cmd/podman/shell_auto_completion.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var _bashCompletionCommand = &cobra.Command{
+	Use:   "bash-completion",
+	Short: "Generates shell script for bash shell",
+	Long: `To load completion run
```
rhatdan: Can you add this as a hidden command.  We don't want to document this, and really do not intend users to use it.

cmd/podman/shell_auto_completion.go
```diff
@@ -0,0 +1,68 @@
+package main
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	bashAuthoCompletionDescription = `To load completion run
```
NevilleC: I need to correct the typo here `Autho` instead of `auto` - I am playing too much with authorised transaction ;)

Still WIP, don't have a lot of spare time but should be able to finish this off by the end of the week.

NevilleC:
I've just discovered it was some completion files already in the repo. I would need to check them as what I am adding here may not add any value.

rhatdan:
@edsantiago @mheon @TomSweeneyRedHat PTAL

rhatdan:
@NevilleC Is there a way to further enhance this to add some existing options.
For existence how do we handle 

```
$ podman --events-backend<tab>
file journal none
```

mheon:
I'd like Ed to take a look at the zsh work before we merge.

On the whole, I really like doing this automatically, maintaining the completions manually is really easy to mess up (I can't guarantee we've always caught PRs that added flags, but no completions)

edsantiago:
I'm really sorry but there is no way I will approve this. The entire point of my zsh completion is for it to be simple, clean, and hands-off: all the completion code generates itself based on the output of `podman --help`. The approach in this PR is unmaintainable: it requires updated any time a podman flag or command is added or changed. I can't accept that. (But, fortunately, I am not the final word in this, and others can override my opinion if they see fit.)

mheon:
We could probably automate a check during PR testing similar to how we do gofmt - run the regeneration command, fail if the git repo is unclean.

NevilleC:
> Is there a way to further enhance this to add some existing options.
> For existence how do we handle

@rhatdan Maybe, I'd need to search for this.

@edsantiago No problem, I do understand your point of view. What I liked with this approach is that the user does not really need to update the `bash` and `zsh` completion scripts manually, it would just need to use a command for it. That's what I tried to achieve here when I discovered that Cobra could generate those for us. So far I tested mainly with `bash` (as it's what I am familiar with) only and it did seem to work well.

Maybe we could keep this approach only for `bash`.

@mheon @edsantiago Let me know in which direction you want to go and I will update/close accordingly this PR.

edsantiago:
>For [example] how do we handle
>```console
>$ podman --events-backend<tab>
>file journal none
>```
To be fair, I can't handle that in zsh right now either. If the `podman --help` output were tweaked something like this:
```diff
-  --events-backend string     Events backend to use
+  --events-backend string     Events backend to use: file, journald, or none
```
...then it would be pretty straightforward to parse that and add those as options. (Or I could just hardcode them directly into the zsh completion script. But as you might guess, I won't do that).

>the user does not really need to update the bash and zsh completion scripts manually

@NevilleC the zsh completion script does not need to be touched. What I was trying to say earlier is that it autogenerates completions. When a user hits TAB, the completion script runs `podman --help` (possibly with a subcommand) and computes the possible completions automatically. It also has logic for handling directory completion, files, images, containers, pods. Cobra offers nothing even close to that.

>Maybe we could keep this approach only for bash for the moment

I'm ok with that.

rhatdan:
@edsantiago I like the idea of specifying the options in the help.

--events. ... (Valures: journald, file, none)

Something like:
[diff.txt](https://github.com/containers/libpod/files/4035395/diff.txt)

edsantiago:
I don't like the manually-maintained aspect, but I love the consistency of the format. I'd love to see that in place.

NevilleC:
@rhatdan @edsantiago @mheon  shall I rebase or close this MR as I am not sure it is something wanted anymore? Please let me know.

edsantiago:
@NevilleC I've looked at the current bash completion, and at the new cobra-generated one, and I lean strongly toward keeping the existing one for now, with the hope that some bashcompletionese speaker will be able to bring in the figure-things-out-automatically `--help` code from zsh. Reasoning: the cobra code is great for keeping every single new option up-to-date; but the existing code is carefully tuned for humans: it autocompletes images, containers, even capabilities. The latter is much friendlier and (IMO) more important than missing some options. But this is just one person's opinion.

mheon:
The loss of autocomplete for container/image names could be a dealbreaker. I don't know if that would be supported in this configuration.

rhatdan:
Right the big advance would be a way to merge the two together.  IE Have a library of calls that exists for greater expansion, perhaps in a separate bash script and then have the autogenerated parts.

But this would require some real expertise and creativity in bash to make it work, I believe.  Sadly if this was easy, I would figure people using cobra would be pointing it out.

NevilleC:
Thanks for the feedback. Closing the PR.

PR #4788, opened 2020-01-04
github stale workflow: rephrase and bump close time

vrothberg:
Rephrase the stale message to be friendlier and bump the closing time to
365 days.  The docs of the stale workflow do not indicate whether we can
not close, so a limit of 365 days seems fair.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,11 +16,9 @@ jobs:
     - uses: actions/stale@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-issue-message: 'This issue had no activity for 30 days. In the absence of activity or the "do-not-close" label, the issue will be automatically closed within 7 days.'
-        stale-pr-message: 'This pull request had no activity for 30 days. In the absence of activity or the "do-not-close" label, the pull request will be automatically closed within 7 days.'
+        stale-issue-message: 'A friendly reminder that this issue had no activity for 30 days.'
+        stale-pr-message: 'A friendly reminder that this PR had no activity for 30 days.'
         stale-issue-label: 'stale-issue'
         stale-pr-label: 'stale-pr'
         days-before-stale: 30
-        days-before-close: 7
-        exempt-issue-label: 'do-not-close'
-        exempt-pr-label: 'do-not-close'
+        days-before-close: 365


vrothberg:
@rhatdan @onlyjob PTAL

rhatdan:
LGTM

.github/workflows/stale.yml
```diff
@@ -16,11 +16,9 @@ jobs:
     - uses: actions/stale@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-issue-message: 'This issue had no activity for 30 days. In the absence of activity or the "do-not-close" label, the issue will be automatically closed within 7 days.'
-        stale-pr-message: 'This pull request had no activity for 30 days. In the absence of activity or the "do-not-close" label, the pull request will be automatically closed within 7 days.'
+        stale-issue-message: 'A friendly reminder that this issue had no activity for 30 days.'
+        stale-pr-message: 'A friendly reminder that this PR had no activity for 30 days.'
```
TomSweeneyRedHat: Are we not closing these after 7 days now?  If so, I think that should still be noted.

.github/workflows/stale.yml
```diff
@@ -16,11 +16,9 @@ jobs:
...
         stale-issue-label: 'stale-issue'
         stale-pr-label: 'stale-pr'
         days-before-stale: 30
-        days-before-close: 7
-        exempt-issue-label: 'do-not-close'
-        exempt-pr-label: 'do-not-close'
+        days-before-close: 356
```
TomSweeneyRedHat: In the description, you said both 356 and 365.  Here you've 356.  Which is it?

.github/workflows/stale.yml
vrothberg: No, we are closing after one year. The intention is to not close at all but the docs of the stale workflow do not indicate whether we can not close, so a limit of 365 days seems fair.

.github/workflows/stale.yml
vrothberg: That was my brain on vacation-mode. 365 is the right one (a year) :)

vrothberg:
@baude @mheon PTAL

mheon:
I'm in favor. We've dealt with the initial lot of issues; it's still valuable for reminding us that issues need to be closed, but most of the time it's one more thing I need to whack with `do-not-close`

TomSweeneyRedHat:
Fighting build issues, otherwise LGTM

vrothberg:
Note that tests are still missing. I will push an image to Quay that we can use for testing.

Issue #4665, opened 2019-12-09
Error: could not get runtime: please update to v2.0.1 or later: outdated conmon version (CentOS 8)


svenvogel:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug
1. podman-1.6.4 -> 94b784255cf849c6d9024def6355fc1cf67c8cbf
2. conmon 2.0.3 -> eb5fa88c26fde5ce1e3f8a1d2a8a9498b2d7dbe6

/kind feature

**Description**
i build podman on centos 8. it works but after install of the RPM and run an container with podman i get the following error. i used 2.0.3 of conmon.

`Error: could not get runtime: please update to v2.0.1 or later: outdated conmon version`

there are any hint or anybody can help whats wrong? maybe it has something to do with  #4465 but i dont know...

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**
podman-1.6.4-1575906879.git94b78425.el8.x86_64

**Additional environment details (AWS, VirtualBox, physical, etc.):**
virtual Machine on KVM

mheon:
Where is Conmon installed to? It is possible you have another Conmon installed that is being used instead of the one you've built - check to see if there are any other Conmon binaries installed at the paths listed in `conmon_path` in libpod.conf.

mheon:
In particular, I would check `/usr/libexec/podman/conmon` - I believe we used to install a bundled Conmon there, while new Conmon installs likely choose `/usr/bin`

haircommander:
you can also find the conmon that podman is choosing with `podman info` or running with `--log-level debug`

mheon:
Part of me thinks that `podman info` will also explode...

We may want to put the path we're using in the error message, to make it clearer what's going on

haircommander:
when searching for a good candidate, we actually go through every path configured and choose only the conmons that are of the right version. @svenvogel, where is the conmon 2.0.3 installed?

svenvogel:
@haircommander thanks for the answer.

podman info gives me the same
`Error: could not get runtime: please update to v2.0.1 or later: outdated conmon version`

with debug
`DEBU[0000] Reading configuration file "/usr/share/containers/libpod.conf"
[...]

PR #4795, opened 2020-01-06
packaging: validate installed rpms

ssbarnea:
Previously we builded RPMs that contained an outdated conmon which was not compatible. From now on `make-install` will also call `podman version` and `podman info` in order to perform a minimal sanity check of the installation.

Fixes: #4665

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -572,10 +572,13 @@ vendor-in-container:
 package:  ## Build rpm packages
 	## TODO(ssbarnea): make version number predictable, it should not change
 	## on each execution, producing duplicates.
-	rm -f  ~/rpmbuild/RPMS/x86_64/* ~/rpmbuild/RPMS/noarch/*
+	rm -rf build/* *.src.rpm ~/rpmbuild/RPMS/*
 	./contrib/build_rpm.sh
 
+# Remember that rpms install exec to /usr/bin/podman while a `make install`
+# installs them to /usr/local/bin/podman which is likely before. Always use
+# a full path to test installed podman or you risk to call another executable.
 package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y remove podman podman-remote
-	sudo ${PKG_MANAGER} -y clean all
 	sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
+	/usr/bin/podman version
+	/usr/bin/podman info  # will catch a broken conmon
--- a/contrib/build_rpm.sh
+++ b/contrib/build_rpm.sh
@@ -2,7 +2,7 @@
 set -euxo pipefail
 
 # returned path can vary: /usr/bin/dnf /bin/dnf ...
-pkg_manager=`command -v dnf yum | head -n1`
+pkg_manager=$(command -v dnf yum | head -n1)
 echo "Package manager binary: $pkg_manager"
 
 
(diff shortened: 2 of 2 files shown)


rhatdan:
@ssbarnea you need to sign your commits.
Fixes LGTM, but shouldn't we just be installing conmon package at this point?  Rather then building it in the spec.

Makefile
```diff
@@ -572,10 +572,13 @@ vendor-in-container:
...
 	./contrib/build_rpm.sh
 
+# Remember that rpms install exec to /usr/bin/podman while a `make install`
+# installs them to /usr/local/bin/podman which is likely before. Always use
+# a full path to test installed podman or you risk to call another executable.
 package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y remove podman podman-remote
-	sudo ${PKG_MANAGER} -y clean all
-	sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
+	sudo ${PKG_MANAGER} -y install --allowerasing ${HOME}/rpmbuild/RPMS/*/*.rpm
```
NevilleC: I don't think we should use `--allowerasing`. From what I could test, it is not an option that `yum` will understand, only `dnf`.

ssbarnea: @NevilleC I added it because at some point I encountered a case where yum picked podman-1.0 from system instead of the newly build rpm, and this is not something we want to happen.

So far none of the 3 platform jobs we had complained about that option so I do not find it as a bug. Do you know a specific platform-version what needs to be supported so I can enable an extra job for it?

ssbarnea:
Note to reviewers: is ok to see github-check that failed as it failed after successfully running the `make package-install`. That broken code is in rdo job definition and I am going to remove it from there.

contrib/build_rpm.sh
```diff
@@ -60,6 +67,3 @@ if [ -d ~/rpmbuild/BUILD ]; then
 fi
 
 rpmbuild --rebuild ${extra_arg:-} podman-*.src.rpm
-
-# clean up src.rpm as it's been built
-sudo rm -f podman-*.src.rpm
```
edsantiago: This is a significant change in behavior and isn't really explained.

edsantiago: In fact I'm uncomfortable with the `rm` being done in Makefile; what would you think of moving this up a few lines, to just before the `make -f .copr/Makefile`?

Makefile
```diff
@@ -572,10 +572,13 @@ vendor-in-container:
...
 
+# Remember that rpms install exec to /usr/bin/podman while a `make install`
+# installs them to /usr/local/bin/podman which is likely before. Always use
+# a full path to test installed podman or you risk to call another executable.
 package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y remove podman podman-remote
-	sudo ${PKG_MANAGER} -y clean all
 	sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
+	/usr/bin/podman version
+	/usr/bin/podman info  # will catch a broken conmon
```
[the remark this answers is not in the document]
ssbarnea: @edsantiago For stuff like the pre-commit does a really good job. Not to mention that we can use it to also run bashate which would point other issues. Anyone for

contrib/build_rpm.sh
```diff
@@ -2,7 +2,7 @@
 set -euxo pipefail
 
 # returned path can vary: /usr/bin/dnf /bin/dnf ...
-pkg_manager=`command -v dnf yum | head -n1`
+pkg_manager=$(command -v dnf yum | head -n1)
```
ssbarnea: That is one issue that pre-commit+bashate can identify.

baude:
lgtm if we can get the last test to pass

contrib/build_rpm.sh
```diff
@@ -53,13 +60,13 @@ export extra_arg="--without doc --without debug"
 echo ${PKGS[*]}
 sudo $pkg_manager install -y ${PKGS[*]}
 
+# clean up src.rpm as it's been built
```
edsantiago: I would've preferred changing the comment, since it's now misleading, but it's not worth rerunning CI. Maybe next time take a quick look for context when you cut and paste.

LGTM. Thank you for your work on this.

ssbarnea:
> Fixes LGTM, but shouldn't we just be installing conmon package at this point? Rather then building it in the spec.

If we are to switch to external conmon we need extra testing and it would outside the scope if this change. Not against but we need to assert availability on all rpm supported platforms (4 atm).

rhatdan:
/hold cancel
Lets move to external conmon in separate PR.

PR #4796, opened 2020-01-06
rpm: assure we build and use a rpm repository

ssbarnea:
Assures we also build a rpm repository and install rpms from inside it.

This improves testing of packaging by assuring that we test the same
way it is supposed to be used in production.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -568,8 +568,17 @@ vendor-in-container:
 package:  ## Build rpm packages
 	## TODO(ssbarnea): make version number predictable, it should not change
 	## on each execution, producing duplicates.
-	rm -f  ~/rpmbuild/RPMS/x86_64/* ~/rpmbuild/RPMS/noarch/*
+	rm -rf build/* *.src.rpm ~/rpmbuild/RPMS/*/*
 	./contrib/build_rpm.sh
 
-package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y install --allowerasing ${HOME}/rpmbuild/RPMS/*/*.rpm
+package-install: package  ## Install rpm packages via a local podman.repo
+	# We use the repository approach with max priority and without disabling
+	# system packages in order to detect if our packages are
+	# creating any conflicts.
+	ls -la ${PWD}/build/buildset
+	sudo cp -f ${PWD}/build/podman.repo /etc/yum.repos.d/podman.repo
+	# --best needed to avoid accident where old rpms ones are picked instead
+	sudo ${PKG_MANAGER} -y install --best --allowerasing --nogpgcheck podman podman-remote
+	podman version
+	# info may require sudo and will also verify conman compatibility
+	sudo podman info --log-level debug
--- a/contrib/build_rpm.sh
+++ b/contrib/build_rpm.sh
@@ -2,7 +2,7 @@
 set -euxo pipefail
 
 # returned path can vary: /usr/bin/dnf /bin/dnf ...
-pkg_manager=`command -v dnf yum | head -n1`
+pkg_manager=$(command -v dnf yum | head -n1)
 echo "Package manager binary: $pkg_manager"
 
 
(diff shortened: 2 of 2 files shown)


NevilleC:
I think some of the changes here could conflict badly with https://github.com/containers/libpod/pull/4762

ssbarnea:
@NevilleC Yep, there is some clashing "opportunity"  around thse but I think we will manage it, probably by splitting out changes in smaller bits, followed by some rebases.

My focus is to fix the package building which currently produces bad rpms. As soon we have this right it will be much easier to introduce the other changes, like making the build process more idempotent. I am really glad that you are working on this because I was a bit surprised to discover the lack of idempotency on the build part.

rhatdan:
@ssbarnea Please sign your PRs
@edsantiago @cevich PTAL

ssbarnea:
recheck-rdo

ssbarnea:
@NevilleC If you can take a look at the PR it woudl breat. It was one major bug now which I was not yet able to figure-out. Apparently `make package-install` may endup installing the podman rpm from `@System` repo instead of one from the newly added `podman` repository, even if I added the priority to it.

Disabling system repo is not an option because it contains other dependencies needed by podman. We need to assure it installs the right version, maybe we should even double check and verify the version after install to avoid a false-positive build.

Makefile
```diff
@@ -568,8 +568,17 @@ vendor-in-container:
...
-	rm -f  ~/rpmbuild/RPMS/x86_64/* ~/rpmbuild/RPMS/noarch/*
+	rm -rf build/* *.src.rpm ~/rpmbuild/RPMS/*/*
 	./contrib/build_rpm.sh
 
-package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y install --allowerasing ${HOME}/rpmbuild/RPMS/*/*.rpm
+package-install: package  ## Install rpm packages via a local podman.repo
+	# We use the repository approach with max priority and without disabling
+	# system packages in order to detect if our packages are
+	# creating any conflicts.
```
cevich: Suggestion: these comments will also display on stdout when the `package-install` recipe is executing.  Not a huge problem but does add a bit to output clutter.  Would the comment be just as useful for maintainers if placed above the target?

Makefile
```diff
@@ -568,8 +568,17 @@ vendor-in-container:
...
+	ls -la ${PWD}/build/buildset
+	sudo cp -f ${PWD}/build/podman.repo /etc/yum.repos.d/podman.repo
+	# --best needed to avoid accident where old rpms ones are picked instead
+	sudo ${PKG_MANAGER} -y install --best --allowerasing --nogpgcheck podman podman-remote
+	podman version
+	# info may require sudo and will also verify conman compatibility
```
NevilleC: s/conman/conmon

contrib/build_rpm.sh
```diff
@@ -47,7 +51,7 @@ else
 fi
 
 echo ${PKGS[*]}
-sudo $pkg_manager install -y ${PKGS[*]}
+sudo $pkg_manager install --disablerepo podman -y ${PKGS[*]}
```
NevilleC: maybe a comment to explain why we need the `--disablerepo podman`?

ssbarnea:
I will take care all comments tomorrow (and rebase). Hopefully I will find a way to solve the repo issue, which is still a mistery to me. At least on rhel-8.1 yum reports zero rpms in the newly created repo, even if the repodata files clearly indicate the two rpms that we build. Installing the rpms manually without a repository works well but I would prefer a full repo test.

edsantiago:
One problem is Epoch: In the fedpkg spec, the Epoch is set to 2. In fake-spec and in rhpkg it is unset. You can possibly get this to work in fedora-land by copying the fedpkg epoch directive:
```
 Name: podman
+%if 0%{?fedora}
+Epoch: 2
+%endif
Version: ...
```
...but then you run into:
```
Error:
 Problem: cannot install the best candidate for the job
  - nothing provides oci-systemd-hook needed by podman-2:1.7.0-1578408618.git61378e61.fc32.x86_64
```
As I seldom get tired of mentioning, it is a *really really really bad idea* to have a duplicate, unmaintained, inconsistent specfile in this repo. (Not your fault! Just a long-existing maintenance nightmare)

ssbarnea:
Closing for the moment until I get the other changes in.

ssbarnea:
@edsantiago @NevilleC @cevich Please check https://github.com/containers/libpod/pull/4815 as I started a new PR to avoid the noise from this one.

PR #4797, opened 2020-01-06
(minor) fix broken links to container-policy.json.5

edsantiago:
Two man pages had incorrect references to a nonexistent
policy.conf(5) or policy-conf(5) [dot vs dash]. Fix them.

Also checked for other broken references via:

   $ for i in registries.conf storage.conf policy.json ; do grep -R $i.5 docs/source | grep -v containers-$i;done

(No further results found. That's not a guarantee that there
aren't other broken links though).


baude:
LGTM

PR #4798, opened 2020-01-06
[CI:DOCS]add return info for prereview

baude:
begin to add return info and status codes for errors so that we can
review the apiv2


rhatdan:
LGTM
although I don't really understand what is going on.  I figure this is just input to a tool that describes what the return codes are when generating API.

baude:
> LGTM
> although I don't really understand what is going on. I figure this is just input to a tool that describes what the return codes are when generating API.

It is an intermediate step for tool input yes.  I havent figured out how the tooling quite works yet so I just took the time to record it and somewhat verify that we are doing the right things.

PR #4805, opened 2020-01-07
log: support --log-opt tag=

giuseppe:
support a custom tag to add to each log for the container.

Closes: https://github.com/containers/libpod/issues/3653

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -419,6 +419,17 @@ Logging driver specific options.  Used to set the path to the container log file
 
 `--log-opt path=/var/log/container/mycontainer.json`
 
+**--log-opt**=*tag*
+
+Set custom logging configuration.  Presently supports the `tag` option
+which specified a custom log tag for the container.  For example:
+
+`--log-opt tag="{{.ImageName}}"`
+
+It supports the same keys as `podman inspect --format`.
+
+It is currently supported only by the journald log driver.
+
 **--mac-address**=*address*
 
 Container MAC address (e.g. 92:d0:c6:0a:29:33)
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -379,6 +379,8 @@ type ContainerConfig struct {
 	CgroupParent string `json:"cgroupParent"`
 	// LogPath log location
 	LogPath string `json:"logPath"`
+	// LogTag is the tag used for logging
+	LogTag string `json:"logTag"`
 	// LogDriver driver for logs
 	LogDriver string `json:"logDriver"`
 	// File containing the conmon PID
@@ -724,6 +726,11 @@ func (c *Container) CgroupParent() string {
 // in the runtime
 func (c *Container) LogPath() string {
 	return c.config.LogPath
+}
+
+// LogTag returns the tag to the container's log file
+func (c *Container) LogTag() string {
+	return c.config.LogTag
 }
 
 // RestartPolicy returns the container's restart policy.
--- a/libpod/container_inspect.go
+++ b/libpod/container_inspect.go
@@ -107,6 +107,7 @@ type InspectContainerData struct {
 	OCIConfigPath   string                      `json:"OCIConfigPath,omitempty"`
 	OCIRuntime      string                      `json:"OCIRuntime,omitempty"`
 	LogPath         string                      `json:"LogPath"`
+	LogTag          string                      `json:"LogTag"`
 	ConmonPidFile   string                      `json:"ConmonPidFile"`
 	Name            string                      `json:"Name"`
 	RestartCount    int32                       `json:"RestartCount"`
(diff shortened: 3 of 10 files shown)


pkg/spec/createconfig.go
```diff
@@ -287,6 +287,11 @@ func (c *CreateConfig) getContainerCreateOptions(runtime *libpod.Runtime, pod *l
 		options = append(options, libpod.WithLogPath(logPath))
 	}
 
+	logTag := getLoggingTag(c.LogDriverOpt)
```
vrothberg: "Logging" vs "Log" inconsistency

docs/source/markdown/podman-create.1.md
```diff
@@ -419,6 +419,12 @@ Logging driver specific options.  Used to set the path to the container log file
 
 `--log-opt path=/var/log/container/mycontainer.json`
 
+**--log-opt**=*tag*
+
+Specify a custom log tag for the container.  For example:
+
+`--log-opt tag="{{.ImageName}}"`
```
vrothberg: What's the format? What are valid keys and values?

libpod/container_inspect.go
```diff
@@ -630,16 +631,7 @@ type InspectNetworkSettings struct {
 }
 
 // Inspect a container for low-level information
```
vrothberg: Maybe extend the comment that this method expects callers to have locked the container (if needed)

vrothberg:
https://review.rdoproject.org/zuul/buildset/fb366dee96094ad4bb02d1c41fb211b6 is failing in other PRs as well. Do you know what that is?

pkg/spec/createconfig.go
giuseppe: I've used the same convention used above for `logPath := getLoggingPath(c.LogDriverOpt)`

docs/source/markdown/podman-create.1.md
```diff
@@ -419,6 +419,16 @@ Logging driver specific options.  Used to set the path to the container log file
 
 `--log-opt path=/var/log/container/mycontainer.json`
 
+**--log-opt**=*tag*
+
+Specify a custom log tag for the container.  For example:
```
mheon: We should mention that other log options are available but not supported. Maybe "Set custom logging configuration. Presently supports the **tag** option which specified a custom log tag for the container. For example..."

pkg/spec/parse.go
```diff
@@ -144,6 +144,18 @@ func getLoggingPath(opts []string) string {
 	return ""
 }
 
+func getLoggingTag(opts []string) string {
```
mheon: Can we combine this and getLoggingPath into one function, so we don't have to double-traverse the parameters? We can also add warnings for unrecognized options

mheon:
I'm starting to really hate Go templates, but if we need to do it, we need to do it. Found a few small things, but overall looks good.

TomSweeneyRedHat:
Tests aren't hip.
Other than the comments from @vrothberg and @mheon, LGTM

rhatdan:
Is this something containers.conf should handle also?

libpod/container.go
```diff
@@ -376,6 +376,8 @@ type ContainerConfig struct {
 	CgroupParent string `json:"cgroupParent"`
 	// LogPath log location
 	LogPath string `json:"logPath"`
+	// LogPath log location
```
NevilleC: this comment seems to be a duplicate of the `LogPath` one

rhatdan:
@giuseppe could you respond to all of the questions above?

pkg/spec/parse.go
giuseppe: done

docs/source/markdown/podman-create.1.md
giuseppe: specified in the man page

giuseppe:
> @giuseppe could you respond to all of the questions above?

I was still looking into adding tests.  We currently have all tests disabled for journald and checking whether the tag is correctly added will probably need to use `journalctl`.

> Is this something containers.conf should handle also?

since it is backend specific, should we have a way to specify `log_opt` instead?

rhatdan:
We have LogSizeMax which maybe should be changed to log_opt,since this seems to be specific to file.

giuseppe:
comments are addressed, but tests are missing

giuseppe:
added a test but the journald tests are not enabled in the e2e tests.

I'll try to enable them as a follow up PR

rhatdan:
What error will I see if I try to use log-opt without journald?

giuseppe:
> What error will I see if I try to use log-opt without journald?

right now it will be silently ignored.  I think it should be fixed in conmon, I'll open a PR to address it

giuseppe:
> right now it will be silently ignored. I think it should be fixed in conmon, I'll open a PR to address it

PR opened here: https://github.com/containers/conmon/pull/115

PR #4807, opened 2020-01-07
Enable pre-commit linting

ssbarnea:
This should help use keep the codebase more consistent, and avoid sevel
whitespace related issues, or bad file permissions.

pre-commit allows us to easily introduce other linters in follow-ups,
like bashate.

Note: pre-commit tool does *not* install any git-hooks. Making commits
will will call the tool unless you deliverately tell it to install the
hooks.


TomSweeneyRedHat:
@ssbarnea LGTM, but I'm not a yaml expert.  @cevich and/or @edsantiago if you've time to peruse, that would be great.

Also this has a WIP label, but I'm not seeing WIP in the title/description.  Should the label be removed @ssbarnea ?

.tool/lint
```diff
@@ -4,6 +4,12 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+
+# Run pre-commit
+PYTHON=$(command -v python python3|head -n1)
+which pre-commit >/dev/null || $PYTHON -m pip install --user pre-commit
```
edsantiago: I'm really uncomfortable with running `pip` as part of a lint step; and I suspect that not all developers will welcome this. Could I recommend instead that you follow the convention used immediately below, i.e. erroring (or perhaps just warning) with instructions on how to install pre-commit?

test/system/helpers.bash
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
edsantiago: This is not a good idea: it might give a reader the mistaken impression that `helpers.bash` is a runnable script. It is not; anyone familiar with BATS will know this, but other will not and might waste time trying to understand its usage. The lack of a shebang here is very deliberate. Are there conventions or directives that can tell `pre-commit` to not warn here?

ssbarnea: I think that adding any kind of shebang would avoid the error. Mainly the logic inside pre-commit is simple: .sh .bash are likely scripts, so they need execute bit and a shebang line. What is inside the shebamg line is mostly up to us.

Excludes are possible but I am doing my best to avoid them as they are not easy to maintain (a single regex for all, not a list). Better to find local workaround.

.tool/lint
ssbarnea: I moved the install code to Makefile following the pattern used with other tools.

edsantiago: I did not explain myself correctly. What I meant is that automatically installing a Python tool via `pip` violates POLA, the Principle Of Least Astonishment. The make process pulls in Go modules - a Go developer will not be surprised by this. Installing Python stuff under a home directory, in non-Go places, that is unexpected. For some people, undesired.

I would suggest that the installation be done manually, a deliberate action on the part of the developer. I would suggest that if the tool is missing, you emit a warning or error with appropriate instructions.

test/system/helpers.bash
edsantiago: I did not explain myself correctly. What I mean is, a shebang may cause confusion in a future maintainer: they might think that this is a runnable script. This has the potential to waste time.

Since there doesn't seem to be a clean exclusion mechanism, please change it to something like:
```
#!/bin/bash --help "This is a BATS helper; IT IS NOT AN EXECUTABLE SCRIPT"
```

Makefile
```diff
@@ -510,6 +510,11 @@ endef
 		$(call go-get,github.com/cpuguy83/go-md2man); \
 	fi
 
+.precommitok:
+	$(PYTHON) -m pip install --user pre-commit
+	$(PYTHON) -m pre_commit --version
+	touch .precommitok
```
cevich: Yeah, not good to do this in a make file, it may not always be running inside automation either.  A better way to get this into automation is:

1.  Add `pre-commit` install during the building of our gate container:  `contrib/gate/Dockerfile`.  Wait for this PR to merge and the auto-build of container image to happen.
2. Another PR adding instructions to kick off the actual checking from `.cirrus.yml` under `gating_task` (I'm assuming these checks run quickly).

contrib/cirrus/lib.sh
```diff
@@ -294,8 +294,8 @@ setup_rootless() {
 
     cd $GOSRC
     # Guarantee independence from specific values
-    ROOTLESS_UID=$[RANDOM+1000]
-    ROOTLESS_GID=$[RANDOM+1000]
+    ROOTLESS_UID=$((RANDOM+1000))
+    ROOTLESS_GID=$((RANDOM+1000))
```
cevich: I haven't researched this much, but we have to be careful here b/c bash is potentially very old under some circumstances.  Are you certain the `$((...))` way functions in vanilla (not updated) Ubuntu 18?

.tool/lint
cevich: See my suggestion below, in comment I made about the Makefile.  We address this concern and make the thing run faster by only installing it once at image-build time instead of PR testing time.

Makefile
ssbarnea: @cevich My top priority is to ease linting on a clean developer clone, always thinking about making as easy as possible for a new-contributor to lint his change, even if only wants to fix a typo in the docs. I do not want to force him to encounter weird errors or to have to read the manual. Probably this should explain why I would want to avoid changing CI files -- these may make it run nicesly in CI but it will produce a bad user experience when you try to run locally.

Usually I would do something like ``which tool || yum install -y tool`` kind of approach, which works decently well even of non supported systems as the user will clearly see that tool is not installed and he could install it manually and run the command again.

With python installed tools is a bit more tricky because I do not want to install them at system level and when using --user the executable will go into ~/.local/bin which is not always in user PATH (i implemented this feature in fc29 myself). Using python module calling bypassed the PATH issue but has the cost of not being able to use `which` to determine if the tool is intalled or not.
 
I also happen to dislike the placeholder file so I will try to look for other options.

To simplify reviews I will keep this one for bashare and open a clean one only for pre-commit. Later we can add bashate.  I fully agree that smaller reviews are to be preffered.

This big patch is more of the result of POC work and was more than enough to get confirmation what we want to introduce these kind of checks, we only need agree on implementation details.

Makefile
```diff
@@ -149,7 +151,12 @@ endif
 	touch $@
 
 lint: .gopathok varlink_generate ## Execute the source code linter
+ifeq ($(PRE_COMMIT),)
+	$(warning pre-commit was not found, we will try to install it.)
+	$(PYTHON) -m pip install --user pre-commit
```
edsantiago: I'm sorry, I don't think I'm explaining myself properly.

As a developer, I do not want surprises from `make`. I do not want my home directory cluttered, my system altered without my consent or even awareness. This is a violation of trust.

You are welcome to issue a failure message telling the user what the problem is and how to solve it:
```
    FATAL: pre-commit tool not found. To continue, please run $(PYTHON) -m pip install --user pre-commit
```
You are welcome to add a new `install.precommit` target and encourage the user to use that. (You are NOT welcome to add that to `install.tools`, because that is so far entirely go-based and mostly required for actual building. This new thing is Python and not required for building, and as such deserves to be kept separate).

I would also encourage you to add these steps after `.tool/lint`: that way a developer at least gets the benefit of that linter even if they choose not to install pre-commit.

I hope this clarifies my position.

Makefile
```diff
@@ -149,7 +151,12 @@ endif
...
+endif
 	@echo "checking lint"
+	$(PRE_COMMIT) run -a
```
edsantiago: It doesn't work that way.

Even if we allowed the auto-install, this wouldn't work because `PRE_COMMIT` would still be null.

This isn't critical, because I'm rejecting auto-install, but is just an FYI to encourage you to look at this sort of thing in the future.

Makefile
```diff
@@ -32,8 +32,10 @@ BUILDTAGS ?= \
 	exclude_graphdriver_devicemapper \
 	seccomp \
 	varlink
-PYTHON ?= $(shell command -v python python3|head -n1)
+PYTHON ?= $(shell command -v python3 python|head -n1)
 PKG_MANAGER ?= $(shell command -v dnf yum|head -n1)
+# ~/.local/bin is not in PATH on all systems
+PRE_COMMIT ?= $(shell command -v ~/.local/bin/pre-commit pre-commit | head -n1)
```
edsantiago: Assuming I understand `pip` correctly (which is not certain, because I'm not a Python fan), this should probably be `$(PYTHON) -m site --user-base` instead of hardcoded `~/.local`

Makefile
cevich: The reality is we need both.  It makes little sense to have developers and automation depending on different linters / checkers plus their configuration and runtime environment. Otherwise we're wasting developers time by mismatched standards for acceptable code.

That is why the gate container exists, and why we have the process of running it locally [documented in (hopefully) the first place a new contributor would look.](https://github.com/containers/libpod/blob/master/CONTRIBUTING.md#go-format-and-lint).  If you can find a way so this make target **ONLY** happens inside the gate container (maybe even possibly running it for the caller), that would be acceptable.

Makefile
```diff
@@ -149,6 +151,12 @@ endif
 	touch $@
 
 lint: golangci-lint
+ifeq ($(PRE_COMMIT),)
+	@echo "FATAL: pre-commit was not found, please install it." >&2
+	@exit 2
```
edsantiago: This isn't helpful: many people will have no idea what pre-commit is or how to install it.

Please put yourself in the shoes of your target audience: take a step back, try to imagine what this will be like for others. Offer an exact command that they can copy/paste, such as a `make` target or a `pip` command.

Makefile
```diff
@@ -149,6 +151,12 @@ endif
...
+endif
+	@echo "checking lint"
```
edsantiago: This `echo` doesn't really add anything

Makefile
edsantiago: I was hoping for clearer instructions, but this will do.

edsantiago:
LGTM. Thanks for persevering on this.

rhatdan:
Thanks @ssbarnea 
/lgtm

vrothberg:
> "FATAL: pre-commit was not found, check https://pre-commit.com/ about installing it."

This just hit me. I'd wish for a make target that installs pre-commit as we do for all required tools.

ssbarnea:
@vrothberg Original proposal had that, but I was asked to remove it. I understood that there is a desire to auto-install only go build dependencies and not non-go ones. Still, feel free to reopen the subject because I agree that makefile should attempt to install it.

rhatdan:
We want this under make install.tools, I believe, as opposed to automatically being installed on first use.

cevich:
or if it's only something needed for automation, it can be installed by related scripts or in container images (e.g. the gating image).

edsantiago:
The problem is that it got added to the default `make lint` target, so it's breaking developer environments. The solution is to create a new `.install.pre-commit` target and let developers install it if they so wish. (It would also be nice to make missing-pre-commit a warning, not fatal, but I've already said enough about that).

PR #4815, opened 2020-01-08
rpm: build a repo and use it

ssbarnea:
Using a repository improves testing of podman installation instead of
using the bare rpm install method by using the likely installation
method used on production.

Depends-On: https://review.rdoproject.org/r/#/c/27527/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -575,10 +575,5 @@ package:  ## Build rpm packages
 	rm -rf build/* *.src.rpm ~/rpmbuild/RPMS/*
 	./contrib/build_rpm.sh
 
-# Remember that rpms install exec to /usr/bin/podman while a `make install`
-# installs them to /usr/local/bin/podman which is likely before. Always use
-# a full path to test installed podman or you risk to call another executable.
-package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
-	/usr/bin/podman version
-	/usr/bin/podman info  # will catch a broken conmon
+package-install: ## Install rpm packages
+	./contrib/build_rpm.sh install
--- a/contrib/build_rpm.sh
+++ b/contrib/build_rpm.sh
@@ -36,10 +36,9 @@ declare -a PKGS=(\
 if [[ $pkg_manager == *dnf ]]; then
     # We need to enable PowerTools if we want to get
     # install all the pkgs we define in PKGS
-    # PowerTools exists on centos-8 but not on fedora-30 and rhel-8
-    if (dnf -v -C repolist all|grep "Repo-id      : PowerTools" >/dev/null); then
-        sudo dnf config-manager --set-enabled PowerTools
-    fi
+    # PowerTools exists on centos-8(-stream) but not on fedora-30 and rhel-8
+    dnf -C repolist all | sed -rn 's/(.*PowerTools).*/\1/p' | \
+        xargs sudo dnf config-manager --set-enabled
 
     PKGS+=(python3-devel \
         python3-varlink \
@@ -70,3 +69,16 @@ if [ -d ~/rpmbuild/BUILD ]; then
 fi
 
 rpmbuild --rebuild ${extra_arg:-} podman-*.src.rpm
+createrepo -v ~/rpmbuild/RPMS/x86_64
+
+if [ "${1:-}" == "install" ]; then
+  	sudo ${pkg_manager} -y --repofrom podman,file://${HOME}/rpmbuild/RPMS/x86_64/ install podman podman-remote
+    # Remember that rpms install exec to /usr/bin/podman while a `make install`
+    # installs them to /usr/local/bin/podman which is likely before. Always use
+    # a full path to test installed podman or you risk to call another executable.
+	/usr/bin/podman version
+	# TODO(ssbarnea): implement a test that verifies the version to match what
+	# we build. The problem is that Version seems to be generic, maybe use
+	# of Git Commit
+	sudo /usr/bin/podman info  # will catch a broken conmon
+fi


ssbarnea:
@edsantiago @NevilleC @cevich -- interestingly this uncovers a bug related to centos-8 where rhel8 where yum decides to install the distro version instead of our code. The same bug does not happen with centos-7 and fedora-30. 

This makes me believe we have another spec issue. Advise would be welcomed.

Makefile
```diff
@@ -575,10 +575,17 @@ package:  ## Build rpm packages
...
 
+build/podman.repo: package
+
 # Remember that rpms install exec to /usr/bin/podman while a `make install`
 # installs them to /usr/local/bin/podman which is likely before. Always use
 # a full path to test installed podman or you risk to call another executable.
-package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
+package-install: package build/podman.repo  ## Install rpm packages
+	#sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
```
edsantiago: Please remove

contrib/build_rpm.sh
```diff
@@ -58,6 +58,8 @@ fi
 export extra_arg="--without doc --without debug"
 
 echo ${PKGS[*]}
+# we disable podman repo because that can become invalid during builds and is
+# not needed for dependencies.
```
edsantiago: Please remove comment: it's misleading (a holdover from previous PR, perahps).

contrib/build_rpm.sh
```diff
@@ -70,3 +72,21 @@ if [ -d ~/rpmbuild/BUILD ]; then
 fi
 
 rpmbuild --rebuild ${extra_arg:-} podman-*.src.rpm
+
+# builds build/podman.repo which should be ready to install locally
+mkdir -p build/buildset
+rm -rf build/buildset/*
```
edsantiago: A better option is to reorder:
```
rm -rf build/buildset
mkdir -p build/buildset
```

Makefile
```diff
@@ -575,10 +575,17 @@ package:  ## Build rpm packages
...
+	sudo cp -f build/podman.repo /etc/yum.repos.d/podman.repo
+	sudo ${PKG_MANAGER} -y --enablerepo podman install podman podman-remote
```
edsantiago: This approach (repo disabled by default) is much better -- but the logical extension is to remove the `cp` entirely and use `--repofrompath` instead; this way the system isn't left in an unmaintainable state. Could you look into using that, esp. seeing if it works with yum?

contrib/build_rpm.sh
```diff
@@ -70,3 +72,21 @@ if [ -d ~/rpmbuild/BUILD ]; then
...
+pushd build/buildset
+   cp -l ~/rpmbuild/RPMS/*/*.rpm .
+   createrepo .
+popd
+cat <<EOF >build/podman.repo
```
edsantiago: This is probably not absolutely critical, but the correct way to create files is to make a tempfile and rename it on success:
```
cat <<EOF >build/podman.repo.tmp.$$ && mv build/podman.repo.tmp.$$ build/podman.repo
```
With this form, if `build/podman.repo` exists it is guaranteed to be correct; without a tmp-rename, it is possible for an error (typically out-of-space but also possibly i/o error) to leave a corrupt file in place. This is a good habit to get into.

Makefile
cevich: Using sudo in a makefile leaves me slightly uncomfortable as it assumes a lot.  For example, if this target is running inside a container as root, it will fail if sudo isn't installed (perhaps for security reasons, perhaps just to keep the image small).  Would it make sense to leave out the `sudo` and just rely on the user running `sudo make package-install` when required?

rhatdan: Yes I agree, I don't think sudo should be in the Makefile.

ssbarnea: I will address all comments receives so far, but before implementing these changes I need help addresing the issue i mentioned at https://github.com/containers/libpod/pull/4815#issuecomment-572085796 -- so far in 2/4 platforms yum/dnf is unable to find the rpms from the builder repo.

ssbarnea: @rhatdan Calling make with sudo would break a huge number of things, think about temp files created as root. 

While is technically easy to detect root and fail if you did not call it as root, it would create a never ending list of problems as all make targets run as root will produce artifacts with wrong user. I really do not with to create any file owned by root in a non-root user home directory.

The jobs we run on rdo now are expecting to find a repository on `project_root/build/builset` and I would prefer not to have to alter them to look in different location.

Look inside the makefile and you will see that `package-install` target depends-on `package` for good reasons. This means that if user calls it with `sudo`, both targets will run as root. build_rpm.sh is also calling copr makefile which produces the source rpms innside build/ folder, so it would introduce root owned files to current project_root folder.

I am ok to remove sudo from makefile if we find a way to avoid messing build when someone runs `sudo make <random-target>` by mistake. I personally find more annoying the messed file ownership that the feature of not calling sudo exactly where is needed.

Maybe we should add a protection for each target and prevent it from running other than how is inteded (like some kind of ensure-root and ensure-non-root checks)?

edsantiago:
If I may ask everyone to take a step back for a moment: what problem are we trying to solve?

Podman is an open-source project, not a Fedora-RHEL one. The little I understand of this PR is that it's intended for a narrow niche testing environment. Does this code really belong in the podman repo? Might there be a better place for it, more tightly integrated with the niche environment it's intended for?

ssbarnea:
@edsantiago @rhatdan @cevich I guess none of you tried to run even the most based `sudo make help`, which produces somethink like:
```
 sudo make help
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Usage: make <target>
```

Based on this I would say that other things need to be addressed before we can remove sudo from inside the makefile.

contrib/build_rpm.sh
```diff
@@ -70,3 +69,16 @@ if [ -d ~/rpmbuild/BUILD ]; then
 fi
 
 rpmbuild --rebuild ${extra_arg:-} podman-*.src.rpm
+createrepo -v ~/rpmbuild/RPMS/x86_64
```
javierpena: For consistency with line 75, could we make this /${HOME}/rpmbuild/RPMS/x86_64 ?

cevich: IIRC, rpmbuild checks an env. var. for the preferred location.  The script should probably reference that instead of hard-coding.  Otherwise this build could break in unexpected ways.

soostdijck:
Any chance of building and incuding a new slirp4netns rpm? It seems that rootless containers do not want to run on centos 7.7 with the default supplied version.

cevich:
My understanding is that rootless containers aren't fully supported on 7, but I cannot be sure of my memory of the reason.  I want to say it has/had something to do with a user-namespace something or other feature in the kernel.

rhatdan:
Correct rootless containers on RHEL7 will not be fully supported until RHEL7.8 release.

rhatdan:
@ssbarnea Still working on this?

vrothberg:
Friendly ping.

vrothberg:
Going to close. Please reopen if the PR is still desired.

PR #4818, opened 2020-01-08
exec: fix pipes

haircommander:
In a largely anticlimatic solution to the saga of piped input from conmon, we come to this solution.

When we pass the Stdin stream to the exec.Command structure, it's immediately consumed and lost, instead of being consumed through CopyDetachable().

When we don't pass -i in, conmon is not told to create a masterfd_stdin, and won't pass anything to the container.

With both, we can do

echo hi | podman exec -til cat

and get the expected hi

fixes: https://github.com/containers/libpod/issues/4785
fixes: https://github.com/containers/libpod/issues/4326
fixes: https://github.com/containers/libpod/issues/3302

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -546,6 +546,10 @@ func (r *ConmonOCIRuntime) ExecContainer(c *Container, sessionID string, options
 		args = append(args, "-t")
 	}
 
+	if options.Streams.AttachInput {
+		args = append(args, "-i")
+	}
+
 	// Append container ID and command
 	args = append(args, "-e")
 	// TODO make this optional when we can detach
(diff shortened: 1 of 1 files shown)


haircommander:
is it a day of triumph or mourning when a bug this long lasting has this simple of a fix. I feel both ways.

mheon:
Hm. Do we need to do anything for `podman run`? I think it has this issue as well... We might need to store `-i` in the DB and use it to determine if Conmon needs `-i`?

mheon:
Damn, though, glad to see we have a fix here

haircommander:
> Hm. Do we need to do anything for `podman run`? I think it has this issue as well... We might need to store `-i` in the DB and use it to determine if Conmon needs `-i`?

hm it does seem run has the same problem, and it's not fixed here. Poking around..

giuseppe:
> echo hi | podman exec -til cat

is it expected to work with `--tty` as well?  If I add `-t` the exec hangs.

In this case, Docker fails with `the input device is not a TTY`

giuseppe:
not worth another iteration but if you are going to re-push for any reason, please shorten the commit log lines

rhatdan:
/lgtm
Is this worth putting out a point release?  Maybe wait a week to see if anything else comes up?

mheon:
Yeah, I'd want to give this a bit to see if anything else pops up with 1.7.0.

PR #4820, opened 2020-01-08
Usage messages: show possible option values

edsantiago:
...in a consistent manner: ("a"|"b"|"c")

This makes it possible (and easy) for zsh completion to
pick those out of the --help messages and offer them
as values when user hits TAB.

I chose this format because it's an already-existing
convention in cmd/podman/common.go.

Also: removed two duplicate "default: x" messages (Cobra
displays those automatically where a non-null default
is specified).

### DIFF
--- a/cmd/podman/main_local.go
+++ b/cmd/podman/main_local.go
@@ -33,7 +33,7 @@ const remote = false
 
 func init() {
 	cgroupManager := define.SystemdCgroupsManager
-	cgroupHelp := "Cgroup manager to use (cgroupfs or systemd)"
+	cgroupHelp := `Cgroup manager to use ("cgroupfs"|"systemd")`
 	cgroupv2, _ := cgroups.IsCgroup2UnifiedMode()
 	if rootless.IsRootless() && !cgroupv2 {
 		cgroupManager = ""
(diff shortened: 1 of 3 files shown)


edsantiago:
@rhatdan PTAL, this is a followup to your suggestion from this morning. It includes the relevant fixes to zsh completion (which, obviously, I tested).

cmd/podman/main_local.go
```diff
@@ -50,12 +50,12 @@ func init() {
 	if err := rootCmd.PersistentFlags().MarkHidden("default-mounts-file"); err != nil {
 		logrus.Error("unable to mark default-mounts-file flag as hidden")
 	}
-	rootCmd.PersistentFlags().StringVar(&MainGlobalOpts.EventsBackend, "events-backend", "", "Events backend to use")
+	rootCmd.PersistentFlags().StringVar(&MainGlobalOpts.EventsBackend, "events-backend", "", `Events backend to use ("file"|"journal"|"none")`)
```
NevilleC: are you sure the value is not `journald` instead of `journal`?

In docs/source/markdown/podman.1.md:
```
**--events-backend**=*type*

Backend to use for storing events. Allowed values are **file**, **journald**, and **none**.
```

edsantiago: Nice catch! Indeed, `libpod/events/events.go` confirms that the actual value is `journald`. Fixed and pushed.

Many thanks for your keen eye. Have I mentioned (lately) how much I dislike manual duplication?...

TomSweeneyRedHat:
LGTM

PR #4823, opened 2020-01-09
podman-generate-systemd --new

vrothberg:
Add a --new flag to podman-generate-systemd to create a new container
via podman-run instead of starting an existing container.

Creating a new container presents the challenge to find a reverse
mapping from a container to the CLI flags it can be created with.  We
are doing this via `(Container).Config.CreateCommand` field, which
includes a copy of the process' command from procFS at creation time.
This field may not be useful when the container was not created via the
Podman CLI (e.g., via a Python script).  Hence, we do not guarantee the
correctness of the generated files.


vrothberg:
@baude @mheon @rhatdan PTAL

@edsantiago, could you have a look at this?

docs/source/markdown/podman-generate-systemd.1.md
```diff
@@ -18,6 +18,10 @@ Note that this command is not supported for the remote client.
 
 Generate files instead of printing to stdout.  The generated files are named {container,pod}-{ID,name}.service and will be placed in the current working directory.
 
+**--generic**, **-g**
+
+Generate generic systemd files which do not require a container to pre-exist as it will create a new one via podman-run.  The generated unit files should be manually checked as there is no guarantee for correctness; the reverse-mapping from a container configuration to Podman's CLI flags may be incorrect, depending on how the container has been created.  Note that **-generic** does not support pods.
+
```
edsantiago: I had some trouble understanding the purpose of this because I misinterpreted 'exist'. Could I suggest a variation of wording, something like:
```
Generate unit files that will recreate a missing/nonexistent container at service start time;
(FIXME explain why this could be useful, e.g. if you expect vandals to 'podman rm' existing containers??)
This option relies on container configuration files, which may not map directly to podman CLI flags; the
user is encouraged to review the generated output carefully before placing in production.
```
...or something like that, assuming I've correctly understood the purpose? I also find the `--generic` name uninformative but can't think of a better option.

pkg/systemdgen/systemdgen.go
```diff
@@ -109,6 +138,41 @@ func CreateContainerSystemdUnit(info *ContainerInfo, generateFiles bool) (string
...
+			"--conmon-pidfile", "/%t/%n-pid",
+			"--cidfile", "/%t/%n-cid",
+		}
+		// Check whether the container was created via
+		// container-{create,run} or the short form.
+		index := 2
+		if info.CreateCommand[1] == "container" {
+			index = 3
+			if len(info.CreateCommand) < 4 {
+				return "", errors.Errorf("container's create command is too short or invalid: %v", info.CreateCommand)
```
edsantiago: duplication. Why not move this up, and combine with 151-153 into:
```go
    index := 2
    if len(info.CreateCommand) > 2 && info.CreateCommand[1] == "container" {
        index = 3
    }
    if len(info.CreateCommand) < index+1 {
        ...error too short
```

docs/source/markdown/podman-generate-systemd.1.md
vrothberg: Thanks for the input! I think `--new` could be much easier to understand :)

pkg/systemdgen/systemdgen.go
vrothberg: Great idea!

docs/source/markdown/podman-generate-systemd.1.md
edsantiago: I'm still having a lot of trouble understanding `--new`, but a big part of that is that I still don't understand the use case for this. The way I understand the need for this, I would find the following names a little more intuitive: `--persistent`, `--regenerate`, `--rerun`, `--retry-run`, `--template-from`. Do any of those match your intention for this option? (This is just a request for info, not for changing anything. I don't understand well enough to make actual suggestions)

vrothberg: Until now, all generated unit files expect that the container already exists. Hence, unit files only do a `podman-start` and `podman-stop`. That's fine but in some cases we want to create a new container to make the unit files more portable (in the sense that they could be used on another machine or more easily be replicated).

This PR is aiming to generate units based on the idea we were working on in Westford (see the blog post https://www.redhat.com/sysadmin/podman-shareable-systemd-services).

edsantiago: Hmmm.... `--portable`?

vrothberg: I am in favor of `--new` as I find it to illustrate best that we create a new container.

@giuseppe @TomSweeneyRedHat @rhatdan @mheon @baude WDYT?

giuseppe: I think `--new` is clearer.  `portable` related to systemd might give the idea of `portable system services` that are a different thing

edsantiago: Thank you for all the conversation. I'll admit that I still dislike `--new`, find it unclear, but this is small in the grand scheme of things. The overall idea and implementation LGTM.

TomSweeneyRedHat:
LGTM, I'm thinking a blog post on this new functionality would be most helpful.

Issue #4785, opened 2020-01-03
Docker/podman difference when piping to exec statement


mycase:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Piping to stdin of a `podman exec` statement behaves differently from piping to `docker exec`

**Steps to reproduce the issue:**

1.  On Fedora 31:
```
$ uname -a
Linux desk004.dabusiness.com 5.3.16-300.fc31.x86_64 #1 SMP Fri Dec 13 17:59:04 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
$ rpm -q docker podman
docker-1.13.1-68.git47e2230.fc30.x86_64
podman-1.6.2-2.fc31.x86_64

```
2. Open a terminal and type
```
$ sudo docker run -ti --name f31c fedora:31
```
Open a second terminal and type
```
$ echo abc | sudo docker exec -i f31c cat
abc
$
```
note the output `abc`

3. Open a terminal and type
```
$ sudo podman run -ti --name f31c fedora:31
```
Open a second terminal and type
```
$ echo abc | sudo podman exec -i f31c cat
$
```
note the absence of any output

**Describe the results you received:**
Stdin doesn't seem to get piped to `podman exec`'s argument statement, whereas it does seem to get piped to `docker exec`'s argument statement. This doesn't follow from the example above, but more complex cases, like piping a database dump to a database, and the database not having been altered in the `podman` case, would suggest this.

**Describe the results you expected:**
I would have expected `podman` and `docker` to behave in the same way

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64
[dauser@desk004 midasdb]$ podman info --debug
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.1
  podman version: 1.6.2
host:
  BuildahVersion: 1.11.3
  CgroupVersion: v1
  Conmon:
    package: conmon-2.0.2-1.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.2, commit: 186a550ba0866ce799d74006dab97969a2107979'
  Distribution:
    distribution: fedora
[...]

PR #4420, opened 2019-11-01
[CI:DOCS] make docsonly prs

baude:
skips certain aspects of prs for testing.

Signed-off-by: baude <bbaude@redhat.com>

PR #4824, opened 2020-01-09
more BATS tests

edsantiago:
- run: --name (includes 'podman container exists' tests)
- run: --pull (always, never, missing)
- build: new test for ADD URL (#4420)
- exec: new test for issue #4785 (pipe getting lost)
- diff: new test
- selinux (mostly copied from docker-autotest)

Plus a bug fix: the wait_for_output() helper would continue
checking, eventually timing out, even if the container had
already exited (probably because of an error). Fix: as
part of the loop, run 'podman inspect' and bail out if
container is not running. Include exit code and logs.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -85,4 +85,55 @@ echo $rand        |   0 | $rand
     run_podman 1 run --rm $IMAGE sh -c /bin/false
 }
 
+@test "podman run --name" {
+    randomname=$(random_string 30)
+
+    # Assume that 4 seconds gives us enough time for 3 quick tests (or at
+    # least for the 'ps'; the 'container exists' should pass even in the
+    # unlikely case that the container exits before we get to them)
+    run_podman run -d --name $randomname $IMAGE sleep 4
+    cid=$output
+
+    run_podman ps --format '{{.Names}}--{{.ID}}'
+    is "$output" "$randomname--${cid:0:12}"
+
+    run_podman container exists $randomname
+    run_podman container exists $cid
+
+    # Done with live-container tests; now let's test after container finishes
+    run_podman wait $cid
+
+    # Container still exists even after stopping:
+    run_podman container exists $randomname
+    run_podman container exists $cid
+
+    # ...but not after being removed:
+    run_podman rm $cid
+    run_podman 1 container exists $randomname
+    run_podman 1 container exists $cid
+}
+
+@test "podman run --pull" {
+    skip_if_remote "podman-remote does not emit 'Trying to pull' msgs"
+
+    run_podman run --pull=missing $IMAGE true
+    is "$output" "" "--pull=missing [present]: no output"
+
+    run_podman run --pull=never $IMAGE true
+    is "$output" "" "--pull=never [present]: no output"
+
+    # Now test with busybox, which we don't have present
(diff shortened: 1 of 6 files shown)


edsantiago:
[Test failure was because podman-remote does not emit "Trying to pull ..." messages; so there's no way to distinguish I-pulled from image-was-already-here. Decided to just `skip_if_remote`.]

edsantiago:
```console
$ ./bin/podman run --rm --privileged --userns=host quay.io/libpod/alpine_labels:latest cat -v /proc/self/attr/current
WARNING: The same type, major and minor should not be used for multiple devices.
```
Any ideas?

edsantiago:
It's not a flake: rerunning fails the same way. There's something weird about the f30 test image, and I suspect it's a real problem. Unfortunately it will be almost impossible to debug. I've submitted https://github.com/opencontainers/runtime-tools/pull/701 as a long-term future way of getting a slight clue about what's wrong, but even if that's accepted (which is unlikely because its CI is failing and I have no idea why) it would take time to vendor this into podman.

TomSweeneyRedHat:
Maybe I'm missing something, but I'm not seeing the warning when I run it "by hand" on my F30 vm with upstream Podman.  However the output is missing a carriage return, it couldn't be that could it?

```
$ podman run --rm --privileged --userns=host quay.io/libpod/alpine_labels:latest cat -v /proc/self/attr/current
unconfined_u:system_r:container_runtime_t:s0^@[tom@localhost ~]$
```

edsantiago:
No, it works fine (no warning message) on all the f30 systems I've tried. The problem is something specific to the CI one. (And no, it's not the NUL-instead-of-NL terminator. That's just the way `/proc/self/attr/current` works everywhere). Thanks for checking!

rhatdan:
sudo podman run --rm --privileged quay.io/libpod/alpine_labels:latest cat -v /proc/self/attr/current
WARNING: The same type, major and minor should not be used for multiple devices.

Something is wacky with --privileged.  Looks like a bug.

rhatdan:
Well the good thing, is this bug exists with older podman.

podman run --privileged alpine echo hello
WARNING: The same type, major and minor should not be used for multiple devices.
hello

rpm -q podman
podman-1.6.2-2.fc31.x86_64

rhatdan:
Added a little extra ouput

WARNING: The same type, major and minor should not be used for multiple devices. /dev/null

Looks like we have two `/dev/null` added.

rhatdan:
Looks like the way my host was configured.

```
sh-5.0# mknod /dev/foo c 1 3 
sh-5.0# podman run --privileged alpine echo hello
WARNING: The same type, major and minor should not be used for multiple devices.
sh-5.0# rm -f /dev/foo 
sh-5.0# podman run --privileged alpine echo hello
hello
```

edsantiago:
https://github.com/opencontainers/runtime-tools/pull/701 is merged, so some day we'll get greater visibility into what the problem is, but that could take years (new runtime-tools release, then a podman revendor). ITM I've simply special-cased the warning and ignored it. The warning [will still appear in test logs](https://api.cirrus-ci.com/v1/task/5088121775718400/logs/system_test.log), in case anyone wants to fix it, but the selinux test will not fail because of it.

PR #4825, opened 2020-01-09
Fix Makefile ref libseccomp branch as a commit

cevich:
Commit a824186ac9 mistakenly updated LIBSECCOMP_COMMIT with a branch
name instead of a commit reference.  This breaks on Ubuntu's git with
the `--detach` option, causing it to throw this error:

```
fatal: '--detach' cannot be used with '-b/-B/--orphan'
```

Fix this by using the tag `v2.3.3` which at the time of this commit, is
the current HEAD of the release-2.3 branch.

Thanks to Ed Santiago <santiago@redhat.com> for helping figure
out the error and the fix.


edsantiago:
LGTM, and special thanks for checking the tag and reporting that - it increases my confidence that there won't be surprises.

/lgtm
/hold

Issue #4833, opened 2020-01-10
rootless container - systemd user service failes on 1.70 but is working on 1.6.2


choeffer:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Followed https://www.redhat.com/sysadmin/podman-shareable-systemd-services to create rootless container systemd user service. On Fedora 31 and podman 1.6.2 it is working fine, on Fedora 31 and podman 1.7.0 it is failing.

**Steps to reproduce the issue:**

1. Create systemd user service, as described in the link above.

2. Start service.

3. Podman 1.6.2 works, 1.7.0 fails.

**Describe the results you received:**

on 1.7.0 I get -> failed (Result: timeout)

```
[fedora@fedora-31-cloud-podman user]$ systemctl --user status redhat_ex.service
● redhat_ex.service - Podman in Systemd
   Loaded: loaded (/home/fedora/.config/systemd/user/redhat_ex.service; disabled; vendor preset: enabled)
   Active: failed (Result: timeout) since Fri 2020-01-10 15:07:28 UTC; 9s ago
      CPU: 115ms

Jan 10 15:05:58 fedora-31-cloud-podman.novalocal systemd[658]: This usually indicates unclean termination of a previous run, or service implementation defici>
Jan 10 15:05:58 fedora-31-cloud-podman.novalocal podman[817]: 2020-01-10 15:05:58.421989591 +0000 UTC m=+0.060805575 container create 6513f72db0a5558629524c5>
Jan 10 15:05:58 fedora-31-cloud-podman.novalocal podman[817]: 2020-01-10 15:05:58.685485772 +0000 UTC m=+0.324301748 container init 6513f72db0a5558629524c524>
Jan 10 15:05:58 fedora-31-cloud-podman.novalocal podman[817]: 2020-01-10 15:05:58.690586238 +0000 UTC m=+0.329402224 container start 6513f72db0a5558629524c52>
Jan 10 15:05:58 fedora-31-cloud-podman.novalocal podman[817]: 6513f72db0a5558629524c5244f1956380b998fbe3d37b9c86339093d087ccf0
Jan 10 15:05:58 fedora-31-cloud-podman.novalocal systemd[658]: redhat_ex.service: New main PID 836 does not belong to service, and PID file is not owned by r>
[...]

PR #4835, opened 2020-01-10
Revert "libpod, rootless: create cgroup for conmon"

vrothberg:
This reverts commit 78e2a31943ddab6efcbc14388532dfb410780299.

Fixes: #4833

@giuseppe PTAL

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -1255,6 +1255,14 @@ func (r *ConmonOCIRuntime) moveConmonToCgroupAndSignal(ctr *Container, cmd *exec
 		mustCreateCgroup = false
 	}
 
+	if rootless.IsRootless() {
+		ownsCgroup, err := cgroups.UserOwnsCurrentSystemdCgroup()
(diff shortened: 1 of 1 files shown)


vrothberg:
@edsantiago, I'd love to have some (system?) tests that run simple containers in systemd service (root & rootless). Do you think that's possible or are there some potential issues I might not have thought of?

giuseppe:
root containers also create the cgroup for conmon?  Is the difference the fact we leave two processes running in the cgroup (slirp4netns and fuse-overlayfs)?

giuseppe:
reverting the patch will break running rootless containers from within a cgroup not owned by the rootless user, e.g. a session created by another user such as root

vrothberg:
> reverting the patch will break running rootless containers from within a cgroup not owned by the rootless user, e.g. a session created by another user such as root

Good point. I'll close the PR.

>  Is the difference the fact we leave two processes running in the cgroup (slirp4netns and fuse-overlayfs)?

Systemd doesn't like that apparently

PR #4836, opened 2020-01-10
Cirrus: Fix libpod base images going stale

cevich:
VM Base images are used as a starting point for runtime VM images.
The in-use VM base images should never be pruned, which is an
operation that potentially occurs periodically from automation
running on the master branch of the libpod repo.

However the only place which updates timestamps (blocking pruning)
of base images, occurs during runtime VM image building.  Therefor,
if images are not regularly rebuilt, it's possible their base images
go stale and are pruned.

Changes:

* Add freshly-produced base images (old ones got pruned)
* Wrap the timestamp update script to include base image names
in the update list.

Notes:

* Regularly updating base image timestamps only needs to happen
on the libpod repo's meta task, since all base images live there.
* Using a wrapper is needed to maintain compatibility with multiple
versions of the imgts container image used by other repos / branchs.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -296,7 +296,7 @@ meta_task:
     timeout_in: 10m
 
     # Cirrus-CI ignores entrypoint defined in image
-    script: '/usr/local/bin/entrypoint.sh |& ${TIMESTAMP}'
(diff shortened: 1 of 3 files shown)


cevich:
***Review note:*** The output from the meta task is the thing to watch (even if it shows green).

For example https://cirrus-ci.com/task/4972365645873152

baude:
LGTM

PR #4767, opened 2019-12-30
[WIP] Attach for APIv2

mheon:
Still very early. Does handle both terminal and non-terminal attach, but does not include the Resize endpoint. TODOs litter the code. Compile-tested only, run at your own risk.

PR #4837, opened 2020-01-10
Add an API for Attach over HTTP API

mheon:
The new APIv2 branch provides an HTTP-based remote API to Podman. The requirements of this are, unfortunately, incompatible with the existing Attach API. For non-terminal attach, we need append a header to what was copied from the container, to multiplex STDOUT and STDERR; to do this with the old API, we'd need to copy into an intermediate buffer first, to handle the headers.

To avoid this, provide a new API to handle all aspects of terminal and non-terminal attach, including closing the hijacked HTTP connection. This might be a bit too specific, but for now, it seems to be the simplest approach.

This commit provides the base API only; APIv2 has not merged yet, so this is not wired into the frontend.

I wanted to rework the rest of attach at the same time (some parts of it, particularly how we start the attach session and how we do resizing, are (in my opinion) handled much better here. That may still be on the table, but I wanted to avoid breaking existing APIs in this already massive change.

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"time"
 
@@ -374,7 +375,9 @@ type AttachStreams struct {
 	AttachInput bool
 }
 
-// Attach attaches to a container
+// Attach attaches to a container.
+// This function returns when the attach finishes. It does not hold the lock for
+// the duration of its runtime, only using it at the beginning to verify state.
 func (c *Container) Attach(streams *AttachStreams, keys string, resize <-chan remotecommand.TerminalSize) error {
 	if !c.batched {
 		c.lock.Lock()
@@ -382,6 +385,7 @@ func (c *Container) Attach(streams *AttachStreams, keys string, resize <-chan re
 			c.lock.Unlock()
 			return err
 		}
+		// We are NOT holding the lock for the duration of the function.
 		c.lock.Unlock()
 	}
 
@@ -389,8 +393,69 @@ func (c *Container) Attach(streams *AttachStreams, keys string, resize <-chan re
 		return errors.Wrapf(define.ErrCtrStateInvalid, "can only attach to created or running containers")
 	}
 
-	defer c.newContainerEvent(events.Attach)
+	c.newContainerEvent(events.Attach)
 	return c.attach(streams, keys, resize, false, nil)
+}
+
+// HTTPAttach forwards an attach session over a hijacked HTTP session.
+// HTTPAttach will consume and close the included httpCon, which is expected to
+// be sourced from a hijacked HTTP connection.
+// The cancel channel is optional, and can be used to asyncronously cancel the
+// attach session.
+// The streams variable is only supported if the container was not a terminal,
+// and allows specifying which of the container's standard streams will be
+// forwarded to the client.
+// This function returns when the attach finishes. It does not hold the lock for
+// the duration of its runtime, only using it at the beginning to verify state.
+func (c *Container) HTTPAttach(httpCon net.Conn, httpBuf *bufio.ReadWriter, streams *HTTPAttachStreams, detachKeys *string, cancel <-chan bool) error {
+	if !c.batched {
+		c.lock.Lock()
+		if err := c.syncContainer(); err != nil {
+			c.lock.Unlock()
+
+			// Write any errors to the HTTP buffer before we close.
+			hijackWriteErrorAndClose(err, c.ID(), httpCon, httpBuf)
+
+			return err
+		}
+		// We are NOT holding the lock for the duration of the function.
+		c.lock.Unlock()
+	}
+
+	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
+		toReturn := errors.Wrapf(define.ErrCtrStateInvalid, "can only attach to created or running containers")
+
+		// Write any errors to the HTTP buffer before we close.
+		hijackWriteErrorAndClose(toReturn, c.ID(), httpCon, httpBuf)
+
+		return toReturn
+	}
+
+	logrus.Infof("Performing HTTP Hijack attach to container %s", c.ID())
+
+	c.newContainerEvent(events.Attach)
+	return c.ociRuntime.HTTPAttach(c, httpCon, httpBuf, streams, detachKeys, cancel)
+}
+
+// AttachResize resizes the container's terminal, which is displayed by Attach
+// and HTTPAttach.
+func (c *Container) AttachResize(newSize remotecommand.TerminalSize) error {
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return err
+		}
+	}
+
+	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
+		return errors.Wrapf(define.ErrCtrStateInvalid, "can only resize created or running containers")
+	}
+
+	logrus.Infof("Resizing TTY of container %s", c.ID())
+
+	return c.ociRuntime.AttachResize(c, newSize)
 }
 
 // Mount mounts a container's filesystem on the host
--- a/libpod/oci.go
+++ b/libpod/oci.go
@@ -1,6 +1,9 @@
 package libpod
 
 import (
+	"bufio"
+	"net"
+
 	"k8s.io/client-go/tools/remotecommand"
 )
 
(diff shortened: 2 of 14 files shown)


mheon:
Marking WIP until APIv2 merges and I can integrate this into it. This will supercede #4767

pkg/api/handlers/containers_attach.go
```diff
@@ -0,0 +1,82 @@
...
+	hijacker, ok := w.(http.Hijacker)
+	if !ok {
+		utils.InternalServerError(w, errors.Errorf("unable to hijack connection"))
+		return
+	}
+	connection, buffer, err := hijacker.Hijack()
+	if err != nil {
+		utils.InternalServerError(w, errors.Wrapf(err, "error hijacking connection"))
+		return
+	}
```
jwhonce: This may help flush the options to make sure the client sets the raw mode
        _, _ = conn.Write([]byte{})

mheon:
Terminal attach is working. Removing WIP.

Non-terminal attach still needs to be tested.

mheon:
Actually, this still needs significant fixes, now that I look at it, so don't merge yet

mheon:
Non-terminal attach confirmed working.

Lingering issues:
- Need to write errors to the HTTP connection
- Hangup on container exit is still iffy - seems to hang for a few seconds figuring out what happened
- Need to make sure we are in a good state to attach before hijacking

mheon:
Issues 1 and 2 are addressed. Just needs swagger docs and proper state checks.

mheon:
Apparently there are query parameters. Will deal with those and swagger docs tomorrow.

mheon:
@rhatdan @vrothberg @baude @jwhonce This should be ready now, PTAL

vrothberg:
Pre-review comment: Can you squash the commits? I am mostly concerned about noise when bisecting.

libpod/oci.go
```diff
@@ -47,6 +50,23 @@ type OCIRuntime interface {
...
+	// STDOUT, STDERR), 3 null (0x00) bytes, 4-byte big endian length.
+	// If a cancel channel is provided, it can be used to asyncronously
+	// termninate the attach session. Detach keys, if given, will also cause
+	// the attach session to be terminated if provided via the STDIN
+	// channel. If they are not provided, the default detach keys will be
+	// used instead. Detach keys of "" will disable detaching via keyboard.
+	// The streams parameter may be passed for containers that did not
+	// create a terminal and will determine which streams to forward to the
+	// client.
+	HTTPAttach(ctr *Container, httpConn net.Conn, httpBuf *bufio.ReadWriter, streams *HTTPAttachStreams, detachKeys *string, cancel <-chan bool) error
```
vrothberg: Thanks for the great comments! _Really_ nice

libpod/container_api.go
```diff
@@ -410,24 +410,45 @@ func (c *Container) HTTPAttach(httpCon net.Conn, httpBuf *bufio.ReadWriter, deta
 		c.lock.Lock()
 		if err := c.syncContainer(); err != nil {
 			c.lock.Unlock()
+
+			// Write any errors to the HTTP buffer before we close.
+			if _, err2 := httpBuf.Write([]byte(err.Error())); err2 != nil {
+				logrus.Errorf("Error writing error to container %s HTTP attach connection: %v", c.ID(), err2)
```
rhatdan: You should you record the err.Error() in the logrus.Error.

libpod/container_api.go
```diff
@@ -410,24 +410,45 @@ func (c *Container) HTTPAttach(httpCon net.Conn, httpBuf *bufio.ReadWriter, deta
...
+			} else {
+				if err3 := httpBuf.Flush(); err3 != nil {
+					logrus.Errorf("Error flushing HTTP buffer for container %s attach connection: %v", c.ID(), err3)
```
rhatdan: You should you record the err.Error() in the logrus.Error.

libpod/container_api.go
```diff
@@ -410,24 +410,45 @@ func (c *Container) HTTPAttach(httpCon net.Conn, httpBuf *bufio.ReadWriter, deta
...
 		// We are NOT holding the lock for the duration of the function.
 		c.lock.Unlock()
 	}
 
 	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
+		toReturn := errors.Wrapf(define.ErrCtrStateInvalid, "can only attach to created or running containers")
+
+		// Write any errors to the HTTP buffer before we close.
+		if _, err := httpBuf.Write([]byte(toReturn.Error())); err != nil {
+			logrus.Errorf("Error writing error to container %s HTTP attach connection: %v", c.ID(), err)
```
rhatdan: You should you record the toReturnr.Error() in the logrus.Error.

libpod/oci_conmon_linux.go
```diff
@@ -482,18 +482,27 @@ func (r *ConmonOCIRuntime) UnpauseContainer(ctr *Container) error {
...
 	}
 
 	// Ensure that our contract of closing the HTTP connection is honored.
 	defer func() {
-		// TODO: we should try to write error to the HTTP connection if
-		// it was not nil
+		// Write any errors to the HTTP buffer before we close.
+		if deferredErr != nil {
+			if _, err := httpBuf.Write([]byte(deferredErr.Error())); err != nil {
+				logrus.Errorf("Error writing error to container %s HTTP attach connection: %v", ctr.ID(), err)
```
rhatdan: You should you record the deferredErr.Error() in the logrus.Error.

mheon:
Rebased, squashed, comments addressed, PTAL

rhatdan:
LGTM

libpod/container_api.go
```diff
@@ -374,25 +375,103 @@ type AttachStreams struct {
...
+		// We are NOT holding the lock for the duration of the function.
 		c.lock.Unlock()
 	}
 
 	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
 		return errors.Wrapf(define.ErrCtrStateInvalid, "can only attach to created or running containers")
 	}
 
-	defer c.newContainerEvent(events.Attach)
+	c.newContainerEvent(events.Attach)
```
haircommander: wouldn't this mean the event happens before the actual attach does? I thought the defer was there to say it happened after c.attach returned

libpod/container_api.go
```diff
@@ -374,25 +375,103 @@ type AttachStreams struct {
...
+
+	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
+		toReturn := errors.Wrapf(define.ErrCtrStateInvalid, "can only attach to created or running containers")
+
+		// Write any errors to the HTTP buffer before we close.
+		if _, err := httpBuf.Write([]byte(toReturn.Error())); err != nil {
+			logrus.Errorf("Error writing error %q to container %s HTTP attach connection: %v", err, c.ID(), err)
+		} else if err2 := httpBuf.Flush(); err2 != nil {
+			logrus.Errorf("Error flushing HTTP buffer for container %s attach connection: %v", c.ID(), err2)
+		}
```
haircommander: can we make this a function? we do this very thing up there

libpod/container_api.go
```diff
@@ -374,25 +375,103 @@ type AttachStreams struct {
...
+
+		if err := httpCon.Close(); err != nil {
+			logrus.Errorf("Error closing HTTP attach connection for container %s: %v", c.ID(), err)
+		}
```
haircommander: aformentioned function should include this too

libpod/oci_conmon_linux.go
```diff
@@ -33,6 +36,13 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const (
+	// This is Conmon's STDIO_BUF_SIZE. I don't believe we have access to it
+	// directly from the Go cose, so const it here
+	bufferSize = 8192 + 1
```
haircommander: https://github.com/containers/conmon/blob/77f4a5131bfc344ab4485cd3f1e4644bfa2122e2/runner/config/config.go#L5 :smile:

libpod/container_api.go
mheon: Yeah. I think it's better to notify as it starts, so we have a better audit-like log of what happened when

libpod/oci_conmon_linux.go
```diff
@@ -465,6 +475,136 @@ func (r *ConmonOCIRuntime) UnpauseContainer(ctr *Container) error {
...
+			if _, err := httpBuf.Write([]byte(deferredErr.Error())); err != nil {
+				logrus.Errorf("Error writing error %q to container %s HTTP attach connection: %v", deferredErr, ctr.ID(), err)
+			} else if err := httpBuf.Flush(); err != nil {
+				logrus.Errorf("Error flushing HTTP buffer for container %s attach connection: %v", ctr.ID(), err)
+			}
+		}
+
+		if err := httpConn.Close(); err != nil {
+			logrus.Errorf("Unable to close HTTP attach connection: %q", err)
+		}
```
haircommander: cc: function comment above

libpod/oci_conmon_linux.go
```diff
@@ -465,6 +475,136 @@ func (r *ConmonOCIRuntime) UnpauseContainer(ctr *Container) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Make a channel to pass errors back
+	errChan := make(chan error)
+
+	attachStdout := true
+	attachStderr := true
+	attachStdin := true
```
haircommander: for non http attach we default to false. why true here?

mheon: Ensuring that, if not specified, we attach to all streams, the default behavior

mheon:
Comments addressed

mheon:
@baude @jwhonce PTAL

PR #4839, opened 2020-01-10
libpod: fix --userns=keep-id with big UIDs

giuseppe:
when creating a keep-id namespace, we split the original user
namespace in:

|inner ns | outer ns  | size|
|------|------|------|
|    0        | 1         | ID|
|    ID       | 0         | 1|
|    ID+1     | ID+1      | availableIds - ID|

When the user ID is bigger than the number of available
subuids/subgids we fail to create the user namespace because the first
slice is bigger than the available number of IDs and the third one has
a negative size.

Fix it by not using more than the available number of IDs in the first
slice and creating the third one only if there are other IDs left.

When the user ID is bigger than the number of additional IDs, there
will be a gap between the two mappings so the IDs between the maximum
additional ID and the user ID won't be present inside of the
namespace.

Closes: https://github.com/containers/libpod/issues/4838


debarshiray:
Looks good to me! Thanks @giuseppe

PR #4842, opened 2020-01-11
[Makefile] Ensure .gopathok dependency is met for varlink

Foxboron:
When executing make in parallel, e.g `make -j8`, there is a chance steps are
executed at the same time. There is a chance .gopathok and the actual varlink
generation happening at the same time, causing a race and ultimately failing the
build.

Ensuring the .gopathok dependency is met at the actual step fixes the problem.

Signed-off-by: Morten Linderud <morten@linderud.pw>


TomSweeneyRedHat:
LGTM, nice catch/fix @Foxboron !

mheon:
LGTM

PR #4843, opened 2020-01-11
swagger documentation updates

baude:
adhere closer to the spec by using description and summary fields and
also ensuring that the id is unique to avoid collision between generic
and libpod endpoints.

also, make swagger output work with redoc which seems to display our
information better for our needs.

### DIFF
--- a/pkg/api/handlers/generic/config.go
+++ b/pkg/api/handlers/generic/config.go
@@ -1,0 +1,9 @@
+package generic
+
+// ContainerCreateResponse is the response struct for creating a container
+type ContainerCreateResponse struct {
+	// ID of the container created
+	Id string `json:"Id"`
+	// Warnings during container creation
+	Warnings []string `json:"Warnings"`
+}
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -11,10 +11,8 @@ import (
 
 func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	// swagger:operation POST /containers/create containers createContainer
-	//
-	// Create a container
-	//
 	// ---
+	// summary: Create a container
 	// produces:
 	// - application/json
 	// parameters:
@@ -24,44 +22,62 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//    description: container name
 	// responses:
 	//   '201':
-	//    description: tbd
+	//       $ref: "#/responses/ContainerCreateResponse"
 	//   '400':
 	//       "$ref": "#/responses/BadParamError"
 	//   '404':
 	//       "$ref": "#/responses/NoSuchContainer"
 	//   '409':
-	//       "$ref": "#/types/ConflictError"
+	//       "$ref": "#/responses/ConflictError"
 	//   '500':
-	//      "$ref": "#/types/InternalError"
+	//      "$ref": "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/containers/create"), APIHandler(s.Context, generic.CreateContainer)).Methods(http.MethodPost)
 	// swagger:operation GET /containers/json containers listContainers
-	//
-	// List containers
-	//
 	// ---
+	// summary: List containers
+	// description: Returns a list of containers
+	// parameters:
+	//  - in: query
+	//    name: filters
+	//    type: string
+	//    description: |
+	//       Returns a list of containers.
+	//        - ancestor=(<image-name>[:<tag>], <image id>, or <image@digest>)
+	//        - before=(<container id> or <container name>)
+	//        - expose=(<port>[/<proto>]|<startport-endport>/[<proto>])
+	//        - exited=<int> containers with exit code of <int>
+	//        - health=(starting|healthy|unhealthy|none)
+	//        - id=<ID> a container's ID
+	//        - is-task=(true|false)
+	//        - label=key or label="key=value" of a container label
+	//        - name=<name> a container's name
+	//        - network=(<network id> or <network name>)
+	//        - publish=(<port>[/<proto>]|<startport-endport>/[<proto>])
+	//        - since=(<container id> or <container name>)
+	//        - status=(created|restarting|running|removing|paused|exited|dead)
+	//        - volume=(<volume name> or <mount point destination>)
 	// produces:
 	// - application/json
 	// responses:
 	//   '200':
-	//     schema:
-	//     type: array
-	//     items:
-	//       "$ref": "#/responses/Container"
+	//        "$ref": "#/responses/DocsListContainer"
 	//   '400':
 	//       "$ref": "#/responses/BadParamError"
 	//   '500':
 	//      "$ref": "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/containers/json"), APIHandler(s.Context, generic.ListContainers)).Methods(http.MethodGet)
 	// swagger:operation POST  /containers/prune containers pruneContainers
-	//
-	// Prune unused containers
-	//
 	// ---
+	// summary: Delete stopped containers
+	// description: Remove containers not in use
 	// parameters:
 	//  - in: query
 	//    name: filters
-	//    type: map[string][]string
-	//    description: something
+	//    type: string
+	//    description:  |
+	//      Filters to process on the prune list, encoded as JSON (a `map[string][]string`).  Available filters:
+	//       - `until=<timestamp>` Prune containers created before this timestamp. The `<timestamp>` can be Unix timestamps, date formatted timestamps, or Go duration strings (e.g. `10m`, `1h30m`) computed relative to the daemon machine’s time.
+	//       - `label` (`label=<key>`, `label=<key>=<value>`, `label!=<key>`, or `label!=<key>=<value>`) Prune containers with (or without, in case `label!=...` is used) the specified labels.
 	// produces:
 	// - application/json
 	// responses:
@@ -71,10 +87,8 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//      "$ref": "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/containers/prune"), APIHandler(s.Context, generic.PruneContainers)).Methods(http.MethodPost)
 	// swagger:operation DELETE /containers/{nameOrID} containers removeContainer
-	//
-	// Delete container
-	//
 	// ---
+	// summary: Remove a container
 	// parameters:
 	//  - in: path
 	//    name: nameOrID
@@ -83,11 +97,13 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//  - in: query
 	//    name: force
 	//    type: bool
-	//    description: need something
+	//    default: false
+	//    description: If the container is running, kill it before removing it.
 	//  - in: query
 	//    name: v
 	//    type: bool
-	//    description: need something
+	//    default: false
+	//    description: Remove the volumes associated with the container.
 	//  - in: query
 	//    name: link
 	//    type: bool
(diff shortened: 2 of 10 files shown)


pkg/api/handlers/generic/config.go
```diff
@@ -0,0 +1,8 @@
+package generic
+
+type CtrCreateResponse struct {
```
vrothberg: Please comment exported types. The linter should complain about it.

vrothberg: s/CtrCreateResponse/CreateContainerResponse/ to make the type match the identifier of the function.

pkg/api/server/register_containers.go
```diff
@@ -24,44 +22,63 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
...
+	//    type: string
+	//    description: |
+	//       Returns a list of containers.
+	//        - ancestor=(<image-name>[:<tag>], <image id>, or <image@digest>)
+	//        - before=(<container id> or <container name>)
+	//        - expose=(<port>[/<proto>]|<startport-endport>/[<proto>])
+	//        - exited=<int> containers with exit code of <int>
+	//        - health=(starting|healthy|unhealthy|none)
+	//        - id=<ID> a container's ID
+	//        - isolation=(default|process|hyperv) (Windows daemon only)
```
vrothberg: s/Windows daemon only/unsupported/ ?

pkg/api/server/register_containers.go
```diff
@@ -24,44 +22,63 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
...
+	//        - volume=(<volume name> or <mount point destination>)
 	// produces:
 	// - application/json
 	// responses:
 	//   '200':
-	//     schema:
-	//     type: array
-	//     items:
-	//       "$ref": "#/responses/Container"
+	//        "$ref": "#/responses/DocsListContainer"
```
vrothberg: `DocsListContainer` ... can we introduce a consistent naming scheme? I guess that's something for a future PR but a consistent naming scheme/coding style is invaluable for reading and understanding the code.

pkg/api/server/register_containers.go
```diff
@@ -24,44 +22,63 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
...
 	// parameters:
 	//  - in: query
 	//    name: filters
-	//    type: map[string][]string
-	//    description: something
+	//    type: string
+	//    description:  |
+	//      Filters to process on the prune list, encoded as JSON (a `map[string][]string`).  Available filters:
+	//       - `until=<timestamp>` Prune containers created before this timestamp. The `<timestamp>` can be Unix timestamps, date formatted timestamps, or Go duration strings (e.g. `10m`, `1h30m`) computed relative to the daemon machine’s time.
+	//       - `label` (`label=<key>`, `label=<key>=<value>`, `label!=<key>`, or `label!=<key>=<value>`) Prune containers with (or without, in case `label!=...` is used) the specified labels.
```
vrothberg: Missing dot before "Prune containers with [...]"

pkg/api/server/register_containers.go
```diff
@@ -107,15 +124,19 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
...
-	//
-	// Inspect Container
-	//
 	// ---
+	// summary: Inspect container
+	// description: Return low-level information about a container.
 	// parameters:
 	//  - in: path
-	//    name: nameOrID
+	//    name: nameorid
```
vrothberg: Other endpoints use the camel-case while the [docker docs](https://docs.docker.com/engine/api/v1.40/#operation/ContainerDelete) name it `id`. Does the identifier impact if the argument matches?

pkg/api/server/register_containers.go
```diff
@@ -126,11 +147,10 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
...
 	//     "$ref": "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/containers/{name:..*}/json"), APIHandler(s.Context, generic.GetContainer)).Methods(http.MethodGet)
-	// swagger:operation POST /containers/{nameOrID}/kill containers killContainer
-	//
-	// Kill Container
-	//
+	// swagger:operation post /containers/{nameorid}/kill containers killcontainer
 	// ---
+	// summary: Kill container
+	// description: send a posix signal to a container, defaulting to killing to the container.
```
vrothberg: For the `signal` parameter: `Signal to send to the container as an integer or string (e.g. SIGINT)`

pkg/api/server/register_containers.go
```diff
@@ -153,10 +173,9 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//      "$ref": "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/containers/{name:..*}/kill"), APIHandler(s.Context, generic.KillContainer)).Methods(http.MethodPost)
 	// swagger:operation GET /containers/{nameOrID}/logs containers LogsFromContainer
-	//
-	// Get logs from container
-	//
 	// ---
+	// summary: Get container logs
+	// description: Get stdout and stderr logs from a container.
```
vrothberg: Do we need to mention restriction for support logging drivers? https://docs.docker.com/engine/api/v1.40/#operation/ContainerLogs mentions that only `json-file` and `journald` are supported.

pkg/api/server/register_containers.go
```diff
@@ -165,46 +184,46 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//  - in: query
 	//    name: follow
 	//    type: bool
-	//    description: needs description
+	//    description: Keep connection after returning logs.
 	//  - in: query
 	//    name: stdout
 	//    type: bool
-	//    description: needs description
+	//    description: not supported?
```
vrothberg: The `?` is confusing

pkg/api/server/register_containers.go
```diff
@@ -165,46 +184,46 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
...
 	//    name: timestamps
 	//    type: bool
-	//    description: needs description
+	//    default: false
+	//    description: Add timestamps to every log line
 	//  - in: query
 	//    name: tail
 	//    type: string
-	//    description: needs description
+	//    description: Only return this number of log lines from the end of the logs
```
vrothberg: The docker docs state
>Only return this number of log lines from the end of the logs. Specify as an integer or all to output all log lines.

Do we support "all"?

pkg/api/server/register_images.go
```diff
@@ -12,9 +12,9 @@ import (
 func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	// swagger:operation POST /images/create images createImage
 	//
-	// Create an image from an image
-	//
 	// ---
+	// summary: Create an image from an image
+	// description: Create an image by either pulling it from a registry or importing it.
```
vrothberg: The parameters do not look correct, see https://docs.docker.com/engine/api/v1.40/#operation/ImageCreate. The docker docs are also incomplete as there is an additional `message` parameter (which will soon reach the docs).

pkg/api/server/register_images.go
```diff
@@ -30,21 +30,20 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
 	//     description: unexpected error
 	//     schema:
 	//      $ref: '#/responses/GenericError'
 	r.Handle(VersionedPath("/images/create"), APIHandler(s.Context, generic.CreateImageFromImage)).Methods(http.MethodPost).Queries("fromImage", "{fromImage}")
 	// swagger:operation POST /images/create images createImage
-	//
-	// Create an image from Source
-	//
 	// ---
+	// summary: Create an image from Source
```
vrothberg: I have a hard time matching this endpoint to the API. https://docs.docker.com/engine/api/v1.40/#operation/ImageCreate has the `fromSrc` as a parameter.

pkg/api/server/register_images.go
```diff
@@ -89,32 +87,41 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	// description: Load a set of images and tags into a repository.
 	// parameters:
 	//  - in: query
 	//    name: quiet
 	//    type: bool
 	//    description: not supported
+	//  - in: body
+	//    description: tarball of container image
+	//    type: string
+	//    format: binary
```
vrothberg: https://docs.docker.com/engine/api/v1.40/#operation/ImageLoad doesn't mention this parameter. Can you elaborate?

pkg/api/server/register_images.go
```diff
@@ -141,8 +147,12 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//    description: maximum number of results
```
vrothberg: Consistency nit: some descriptions are entire sentences, some start with capital caps, other with lower.

pkg/api/server/register_images.go
```diff
@@ -238,10 +242,9 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//      $ref: "#/responses/InternalError"
 	r.Handle(VersionedPath("/images/{name:..*}/json"), APIHandler(s.Context, generic.GetImage))
 	// swagger:operation POST /images/{nameOrID}/tag images tagImage
-	//
-	// Tag an image
-	//
 	// ---
+	// summary: Tag an image
+	// description: Tag an image so that it becomes part of a repository.
```
vrothberg: Nit: "so that it becomes part of a repository" is not entirely true in the libpod world. `podman tag $ID foo` resolves to `localhost/foo` which strictly speaking isn't a real repository.

pkg/api/server/register_images.go
```diff
@@ -318,27 +319,17 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
 	*/
 
-	// swagger:operation POST /libpod/images/{nameOrID}/exists images imageExists
-	//
-	// Check if image exists in local store
-	//
+	// swagger:operation POST /libpod/images/{nameOrID}/exists images libpodImageExists
 	// ---
+	// summary: Image exists
+	// description: Check if image exists in local store
```
vrothberg: s/store/storage/

vrothberg: Check if **an** image  ... ?

vrothberg:
> The v2 merge introduced new lint errors which we need to fix first.

I traced it down. `make lint` is using the old metalinter while `make validate` is using golangci-lint which in turn has quite some linters disabled. I'll open a PR to clean it up and enable more linters.

jwhonce:
@vrothberg Short answer is spec does all defaults.

  name: debug
          schema:
            type: integer
            enum: [0, 1]
            default: 0

We would need to test, that it's parsed and then passed to the generator correctly to allow reDoc to display.

/cc @baude

pkg/api/server/register_pods.go
```diff
@@ -9,10 +9,8 @@ import (
 
 func (s *APIServer) registerPodsHandlers(r *mux.Router) error {
 	// swagger:operation GET /libpod/pods/json pods ListPods
```
jwhonce: I know libpodListPods      is not required.  But it would be more consistent.

pkg/api/server/swagger.go
```diff
@@ -131,3 +131,13 @@ type swagListContainers struct {
 		//handlers.Container
 	}
 }
+
+// To be determined
+// swagger:response tbd
+type swagTBD struct {
+}
+
+// Filters baude
```
mheon: Can the comments in this file be improved a bit? I have no idea what's going on here.

baude: no they cannot, they are only used by go-swagger to generate documentation.  the one you pick on in particular is just a test. lemme remove it.

PR #4850, opened 2020-01-13
Fix linting

vrothberg:
A series of commits to:
* Update golangci-lint and make it initially pass. There was a series of issues preventing from running it cleanly (e.g., broken contrib/perftest and upstream regressions). Pinning it to 1.18.0 for now.
* Use a blacklisting approach for linters instead of enabling them one-by-one. This makes it easier to maintain (updates will pull in new linters automatically) and it's easier to enable disabled linters.
* Last but not least, start by enabling `gocritic`.

@baude @mheon @rhatdan @TomSweeneyRedHat PTAL


libpod/oci_conmon_linux.go
```diff
@@ -586,7 +586,7 @@ func (r *ConmonOCIRuntime) ExecContainer(c *Container, sessionID string, options
 
 	// we don't want to step on users fds they asked to preserve
 	// Since 0-2 are used for stdio, start the fds we pass in at preserveFDs+3
-	execCmd.Env = append(r.conmonEnv, fmt.Sprintf("_OCI_SYNCPIPE=%d", options.PreserveFDs+3), fmt.Sprintf("_OCI_STARTPIPE=%d", options.PreserveFDs+4), fmt.Sprintf("_OCI_ATTACHPIPE=%d", options.PreserveFDs+5))
+	execCmd.Env = append(execCmd.Env, fmt.Sprintf("_OCI_SYNCPIPE=%d", options.PreserveFDs+3), fmt.Sprintf("_OCI_STARTPIPE=%d", options.PreserveFDs+4), fmt.Sprintf("_OCI_ATTACHPIPE=%d", options.PreserveFDs+5))
```
giuseppe: it doesn't seem equivalent to the previous version.

Do we need to set `execCmd.Env = r.conmonEnv` ?

vrothberg: Good catch! Yes, we need to set it like that. I got tricked with the following line which is adding `conmonEnv`.

vrothberg:
> level=error msg="Running error: no such linter \"funlen\""

Looks like the CI has another golangci-lint already installed

vrothberg:
> > level=error msg="Running error: no such linter "funlen""
> 
> Looks like the CI has another golangci-lint already installed

@cevich, `quay.io/libpod/gate:master` ships with the tools from `make install.tools`. Are you cool with removing that from the Dockerfile? This way, the gating tests will always fetch the versions the Makefile wants. I didn't want to update and overwrite the image before consulting you.

TomSweeneyRedHat:
LGTM once the CI and cevich are happy

libpod/container.log.go
```diff
@@ -56,7 +56,7 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 				continue
 			}
 			if nll.Partial() {
-				partial = partial + nll.Msg
+				partial += nll.Msg
```
mheon: I honestly thought Go did not support this syntax

mheon:
LGTM

mheon:
I'll defer merge until we get a response from Chris.

vrothberg:
> I'll defer merge until we get a response from Chris.

I think we can merge as I did not change the Dockerfile (or the image). To make it pass for now, I added an `rm $GOPATH/bin/*` to the gating tasks.

mheon:
Manually adding label so we merge

cevich:
Sorry I'm late to the party as usual :disappointed: 

> quay.io/libpod/gate:master ships with the tools from make install.tools...cool with removing that from the Dockerfile?

Yes, please make whatever changes are needed.  My understanding is the only user of that image is 'gate_task'.  The purpose of both is to reliably pass or fail as quickly as possible.  Though quickly and consistently failing, is typically undesirable :smile:  So installing at runtime is fine if the linter is a moving target.

PR #4852, opened 2020-01-13
pod: honor --infra-command

giuseppe:
if --infra-command is specified, it must have a higher precedence over
the command specified in the image.

If the image specifies an entrypoint, it must be used also with
--infra-command.

### DIFF
--- a/cmd/podman/libpodruntime/runtime.go
+++ b/cmd/podman/libpodruntime/runtime.go
@@ -169,7 +169,7 @@ func getRuntime(ctx context.Context, c *cliconfig.PodmanCommand, renumber, migra
 	}
 
 	infraCommandFlag := c.Flags().Lookup("infra-command")
-	if infraCommandFlag != nil && infraImageFlag.Changed {
+	if infraCommandFlag != nil && infraCommandFlag.Changed {
 		infraCommand, _ := c.Flags().GetString("infra-command")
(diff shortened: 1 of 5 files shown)


mheon:
Concern: I think this may unconditionally override image entrypoints if the config file does not explicitly set `infra_command = ""` (otherwise, the defaults will plug in `"/pause"`)

giuseppe:
> Concern: I think this may unconditionally override image entrypoints if the config file does not explicitly set `infra_command = ""` (otherwise, the defaults will plug in `"/pause"`)

good point.  Should we default to ""?

mheon:
As long as this doesn't break the default k8s.io pause image, I think that's a good idea

giuseppe:
> As long as this doesn't break the default k8s.io pause image, I think that's a good idea

it seems the k8s.io image has an ENTRYPOINT defined: https://github.com/kubernetes/kubernetes/blob/master/build/pause/Dockerfile

giuseppe:
> As long as this doesn't break the default k8s.io pause image, I think that's a good idea

PR opened here: https://github.com/containers/common/pull/41

giuseppe:
the correct thing to do would be to honor the setting in this order:

- the `--infra-command` value
- what is set in the image
- the default setting in the conf file.

I'll rework it and open a new PR

Issue #4687, opened 2019-12-12
Uses/requires CNI for slirp4ns


ensc:
/kind bug

**Description**

When running podman in root mode, it requires/uses CNI to build slirp4ns networks.  In rootless mode with same parameters, CNI is not used.

CNI is problematic because it messes around with `iptables` directly which breaks when they are managed externally.  For this, I need a pure userspace solution (more or less like `dockerd --iptables=false`) like slirp4ns.

**Steps to reproduce the issue:**

1.  `rm -rf /etc/cni/net.d`  (or keep it and use `podman inspect` later to see whether CNI is used)

2. run as root:

    ```
    podman run -it --network=slirp4ns fedora:31
    ```

3. run as user:

   ```
   podman run -it --network=slirp4ns fedora:31
   ```

**Describe the results you received:**

Step 2 fails with

```
ERRO[0000] CNI network "slirp4ns" not found             
```

Step 3 succeeds.

**Describe the results you expected:**

Step 2 succeeds and CNI is not used.

**Output of `podman version`:**

```
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.1
  podman version: 1.6.2
host:
  BuildahVersion: 1.11.3
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.2-1.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.2, commit: 186a550ba0866ce799d74006dab97969a2107979'
  Distribution:
    distribution: fedora
    version: "31"
  MemFree: 2040983552
  MemTotal: 32669814784
  OCIRuntime:
    name: crun
    package: crun-0.10.6-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.10.6
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  SwapFree: 4177784832
  SwapTotal: 4294963200
  arch: amd64
  cpus: 8
  eventlogger: journald
  hostname: ensc-virt.intern.sigma-chemnitz.de
  kernel: 5.3.15-300.fc31.x86_64
  os: linux
  rootless: false
  uptime: 26h 4m 9.38s (Approximately 1.08 days)
registries:
  blocked: null
  insecure: null
  search:
[...]

PR #4853, opened 2020-01-13
Do not configure CNI when slirp4netns is requested

mheon:
Our networking code bakes in a lot of assumptions about how networking should work - that CNI is *always* used with root, and that slirp4netns is *always* used only with rootless. These are not safe assumptions. This fixes one particular issue, which would cause CNI to also be run when slirp4netns was requested as root.

Fixes: #4687


AkihiroSuda:
rootlessport needs to be enable as well when slirp4netns is used?

Also, what is the usecase of this?

mheon:
I'm not honestly sure - support for slirp4netns as root was added at some point in the past, but I can't recall why. It doesn't seem to have many users.

baude:
code LGTM

rhatdan:
@AkihiroSuda Most likely running podman inside of a container.

PR #4858, opened 2020-01-14
make lint: extend checks

vrothberg:
This PR includes a series of commits to extend the linter checks to include:
* `pkg/tracing`
* `docs/`
* unit tests (i.e., `_test.go` files)

Especially checking the unit tests could make CI more robust as rigorous error checks could reveal flakes.


vrothberg:
Next round.  @TomSweeneyRedHat @rhatdan @baude @mheon PTAL

The next candidate (if time allows) is to include `test/` into the checks as well. After that, we can enable the currently disabled one-by-one.

cmd/podman/remoteclientconfig/configfile_test.go
```diff
@@ -92,14 +92,15 @@ func TestReadRemoteConfig(t *testing.T) {
 		{"nouser", args{reader: strings.NewReader(noUser)}, makeNoUserResult(), false},
 	}
 	for _, tt := range tests {
+		test := tt
```
TomSweeneyRedHat: why keep 'tt' at all, can't you drop this line and replace tt in line 94 with 'test'?

cmd/podman/remoteclientconfig/configfile_test.go
```diff
@@ -192,17 +194,18 @@ func TestRemoteConfig_GetRemoteConnection(t *testing.T) {
 		{"none", fields{Connections: blank}, args{name: "foobar"}, nil, true},
 	}
 	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
+		test := tt
```
TomSweeneyRedHat: ditto tt comment

libpod/image/image_test.go
```diff
@@ -213,7 +210,7 @@ func TestImage_RepoDigests(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for _, test := range []struct {
+	for _, tt := range []struct {
```
TomSweeneyRedHat: OK, now I'm confused.  Lint wanted tt changed to test in the other code, but test to tt here?

cmd/podman/remoteclientconfig/configfile_test.go
vrothberg: We shouldn't use loop variables in go which may leak outside the loop (e.g., by appending it to a slice outside of the loop) or when the code can run in parallel. The latter is the case for all t.Run(..)s as unit tests could run in parallel.

vrothberg: The "problem" with loop variables is that they're actually a pointer. `test := tt` allocates a new variable and copies the value of `tt`. This way, we're safe.

libpod/image/image_test.go
vrothberg: I wanted it to be consistent. `tt` as the loop variable, `test` for the variable we're using in the loop's body.

pkg/hooks/1.0.0/when_test.go
```diff
@@ -48,7 +52,8 @@ func TestHasBindMountsAnd(t *testing.T) {
 	hasBindMounts := true
 	when := When{HasBindMounts: &hasBindMounts}
 	config := &rspec.Spec{}
-	for _, containerHasBindMounts := range []bool{false, true} {
+	for _, b := range []bool{false, true} {
+		containerHasBindMounts := b
```
TomSweeneyRedHat: OK, these three b, a, and o are like tt/tests above.  I don't get it.

cmd/podman/remoteclientconfig/configfile_test.go
TomSweeneyRedHat: thx, didn't see the leak outside of the loop happening in this code, but may well have missed it.

rhatdan:
LGTM
Nice work.

.gitignore
```diff
@@ -21,11 +21,11 @@ __pycache__
 /cmd/podman/varlink/ioprojectatomicpodman.go
 /cmd/podman/varlink/iopodman.go
 .gopathok
-test/e2e/e2e.coverprofile
```
mheon: Why? I think this is useful to prevent it accidentally being committed

vrothberg: Because we ignore all `*.coverprofile` now

mheon:
The only thing pending is a Prow test, so I'll go ahead and whack this
/lgtm

haircommander:
git hooks have been dropping for prow btw, the suggestion is to spam retest until it triggers
/retest

PR #4859, opened 2020-01-14
oci_conmon: not make accessible dirs if not needed

giuseppe:
do not change the permissions mask for the rundir and the tmpdir when running a container with a user namespace and the current user is mapped inside the user namespace.
    
The change was introduced with 849548ffb8e958e901317eceffdcc2d918cafd8d, that dropped the intermediate mount namespace in favor of allowing root into the user namespace to access these directories.

Closes: https://github.com/containers/libpod/issues/4846

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -149,9 +149,23 @@ func (r *ConmonOCIRuntime) Path() string {
 	return r.path
 }
 
+// hasCurrentUserMapped checks whether the current user is mapped inside the container user namespace
+func hasCurrentUserMapped(ctr *Container) bool {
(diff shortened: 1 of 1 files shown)


rhatdan:
Very cool if this works.
LGTM

giuseppe:
> Very cool if this works.
> LGTM

so now I remember where it was failing.  If the user is not mapped inside of the user namespace, it won't have access to any temporary file we create (such as hostname, .containerenv, passwd) and the OCI runtime will fail.

The correct check is to verify if the current user is mapped inside the user namespace.

mheon:
Huh. Neat. LGTM once this goes green.

PR #4860, opened 2020-01-14
v2 api: top improvements

vrothberg:
* Use `pkg/adapter` to increase code reuse and reduce code redundancy.
* Extend swagger docs to mention AIX descriptors.
* Document the libpod endpoint which shares the same handler.

@baude @jwhonce PTAL

### DIFF
--- a/pkg/api/handlers/containers_top.go
+++ b/pkg/api/handlers/containers_top.go
@@ -4,10 +4,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/containers/libpod/cmd/podman/cliconfig"
 	"github.com/containers/libpod/libpod"
-	"github.com/containers/libpod/libpod/define"
-	"github.com/containers/libpod/pkg/adapter"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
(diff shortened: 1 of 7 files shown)


pkg/api/handlers/containers_top.go
```diff
@@ -28,23 +29,14 @@ func TopContainer(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		utils.InternalServerError(w, err)
-		return
-	}
-	if state != define.ContainerStateRunning {
-		utils.ContainerNotRunning(w, name, errors.Errorf("Container %s must be running to perform top operation", name))
-		return
-	}
+	adapterRuntime := adapter.LocalRuntime{}
+	adapterRuntime.Runtime = runtime
```
vrothberg: It would be nice to make `runtime` be an `*adapter.LocalRuntime` by default.

vrothberg:
While testing I found a small bug in the top-backend. In some cases, we fall back to exec-ing ps(1) in the container. Via the v2 api, I've seen conmon debug messages in the returned data. That was because the top-backend was using one pipe for stderr and stdout. I assume we never caught that because we don't run `top` in debug-mode. Anyhow, it's fixed now.

vrothberg:
Now with refactored code. @baude PTAL

pkg/api/server/register_containers.go
```diff
@@ -345,7 +345,7 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//  - in: query
 	//    name: ps_args
 	//    type: string
-	//    description: arguments to pass to ps such as aux
+	//    description: arguments to pass to ps such as aux. Requires ps(1) to be installed in the container if no ps(1) compatible AIX descriptors are used.
```
jwhonce: default: field should be added

PR #4861, opened 2020-01-14
oci_conmon: do not create a cgroup under systemd

giuseppe:
Detect whether we are running under systemd (if the INVOCATION_ID is set).  If Podman is running under a systemd service, we do not need to create a cgroup for conmon.

The variable is set only on new systemd versions (the version on RHEL8 seems fine and sets it), so for older versions add a `--cgroups=disabled-conmon` option to disable cgroups for conmon.

Closes: https://github.com/containers/libpod/issues/4833


docs/source/markdown/podman-run.1.md
```diff
@@ -92,8 +92,9 @@ If the host uses cgroups v1, the default is set to **host**.  On cgroups v2 the
 **--cgroups**=*mode*
 
 Determines whether the container will create CGroups.
-Valid values are *enabled* and *disabled*, which the default being *enabled*.
+Valid values are *enabled*, *disabled*, *disabled-conmon*, which the default being *enabled*.
```
mheon: I think `no-conmon` would be a better name, but I'd like other folks to agree to that before we change it.

libpod/container.go
```diff
@@ -375,6 +375,8 @@ type ContainerConfig struct {
 	// NoCgroups indicates that the container will not create CGroups. It is
 	// incompatible with CgroupParent.
 	NoCgroups bool `json:"noCgroups,omitempty"`
```
mheon: In retrospect we should have made this something other than a bool, but I think we're too late to fix that. The new question is whether we should make the new value a string instead - CgroupMode, maybe? - to allow for future expansion

docs/source/markdown/podman-run.1.md
giuseppe: @vrothberg do you have any preference on the naming?

vrothberg: I don't have a preference.

giuseppe:
how to move forward?  I rename the new option to `no-conmon` and make it a string instead of a bool?

vrothberg:
> how to move forward? I rename the new option to `no-conmon` and make it a string instead of a bool?

SGTM

rhatdan:
Can you add ("enabled"|"disabled"|"no-conmon") to the command line option?  @edsantiago uses this for generation of the completion flags.

libpod/oci_conmon_linux.go
```diff
@@ -1297,8 +1297,15 @@ func startCommandGivenSelinux(cmd *exec.Cmd) error {
 // it then signals for conmon to start by sending nonse data down the start fd
 func (r *ConmonOCIRuntime) moveConmonToCgroupAndSignal(ctr *Container, cmd *exec.Cmd, startFd *os.File) error {
 	mustCreateCgroup := true
+
 	// If cgroup creation is disabled - just signal.
-	if ctr.config.NoCgroups {
+	switch ctr.config.CgroupsMode {
```
mheon: We should still check ctr.config.NoCgroups as well (legacy reasons).

libpod/container.go
mheon: Can you mark this as deprecated while you're in here? So we know not to use it more

pkg/spec/namespaces.go
```diff
@@ -216,6 +216,9 @@ func (c *CgroupConfig) ToCreateOptions(runtime *libpod.Runtime) ([]libpod.CtrCre
 	if c.Cgroups == "disabled" {
```
mheon: IMO we should disable this and remove WithNoCgroups, and move everything to use WithCgroupsMode

mheon: We'll need to retain NoCgroups in the database and keep supporting it there for legacy reasons, but we should move all new containers over to use the new version

rhatdan: If we are depracatng it, then remove it from the docs and hide the option.  No reason for people to continue to use it.

mheon: It should continue working - we're just changing the way we store it in the database

rhatdan: But I thought we were discouraging an option?

mheon:
Just the way we used to store it in the DB - want to move from a bool to a
string

giuseppe:
tests are passing

rhatdan:
LGTM
But I thought that @mheon still has an issue?

giuseppe:
>  But I thought that @mheon still has an issue?

I think I took care of all the comments.  @mheon is there still anything pending?

giuseppe:
@mheon are you fine with the current version?

mheon:
Looking...

PR #3901, opened 2019-08-28
Cirrus: Support testing with F31

cevich:
Add F31 + Drop F29

***Depends on: #3754 #4028 #4028 #4154 #4222 #4825***

Fixes #5285
Fixes #3447

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #4870, opened 2020-01-15
Gating dockerfile

vrothberg:
A series of commits to
* Bump the base image of the gating task to F31.
* Remove gometalinter from the Makefile (not needed anymore).
* Update make .install.golangci-lint to force a specific version.

@cevich PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -481,7 +481,7 @@ uninstall:
 	GIT_CHECK_EXCLUDE="./vendor:docs/make.bat" $(GOBIN)/git-validation -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..$(HEAD)
 
 .PHONY: install.tools
-install.tools: .install.gitvalidation .install.gometalinter .install.md2man .install.ginkgo .install.golangci-lint ## Install needed tools
+install.tools: .install.gitvalidation .install.md2man .install.ginkgo .install.golangci-lint ## Install needed tools
 
 define go-get
 	env GO111MODULE=off \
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM and happy green test buttons

Makefile
```diff
@@ -498,19 +498,8 @@ endef
...
 
-.install.gometalinter: .gopathok
-	if [ ! -x "$(GOBIN)/gometalinter" ]; then \
-		$(call go-get,github.com/alecthomas/gometalinter); \
-		cd $(FIRST_GOPATH)/src/github.com/alecthomas/gometalinter; \
-		git checkout --detach e8d801238da6f0dfd14078d68f9b53fa50a7eeb5; \
-		$(GO) install github.com/alecthomas/gometalinter; \
-		$(GOBIN)/gometalinter --install; \
-	fi
-
```
cevich: Good riddance! :smile:

contrib/gate/Dockerfile
```diff
@@ -1,6 +1,5 @@
-FROM fedora:30
+FROM fedora:31
 RUN dnf -y install \
-      atomic-registries \
```
cevich: This will conflict with #3901 but not badly (Mentioning it for my own notes)

hack/install_golangci.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+$GOBIN/golangci-lint --version | grep $VERSION
```
cevich: Suggest validating at least `$VERSION` is non-empty.  Did you intended `grep -q`?  Also should probably `set -eo pipefail`.  Otherwise if `golangci-lint` dies, you'll miss it.

cevich:
*sigh* changes to Dockerfile won't be validate until after this merges.  Maybe someday I'll get around to solving that :disappointed:

vrothberg:
> _sigh_ changes to Dockerfile won't be validate until after this merges. Maybe someday I'll get around to solving that disappointed

We could build it locally and push it.

hack/install_golangci.sh
vrothberg: Great catch! I'll open a follow-up PR.

vrothberg: https://github.com/containers/libpod/pull/4879

cevich:
> We could build it locally and push it.

Yes.  However as the automation-guru here, suggesting that solution causes my Chakras to become misaligned :disappointed:

PR #4874, opened 2020-01-15
[CI:DOCS]swagger cleanup and left-hand nav

baude:
add a static tags file so we can dictate the left-hand navigation. in
doing so we now override the tag in the swagger:operation. we now have
images and images (compat) as a way to differentiate.

### DIFF
--- a/pkg/bindings/containers.go
+++ b/pkg/bindings/containers.go
@@ -109,12 +109,14 @@ func (c Connection) UnpauseContainer(nameOrID string) error {
 }
 
 func (c Connection) WaitContainer(nameOrID string) error {
-	_, err := http.Post(c.makeEndpoint(fmt.Sprintf("containers/%s/wait", nameOrID)), "application/json", nil)
+	// TODO when returns are ironed out, we can should use the newRequest approach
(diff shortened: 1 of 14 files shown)


pkg/bindings/errors.go
```diff
@@ -0,0 +1,46 @@
+package bindings
+
+import (
+	"encoding/json"
+	"github.com/sirupsen/logrus"
```
mheon: Import ordering

pkg/bindings/containers.go
```diff
@@ -109,12 +109,14 @@ func (c Connection) UnpauseContainer(nameOrID string) error {
 }
 
 func (c Connection) WaitContainer(nameOrID string) error {
-	_, err := http.Post(c.makeEndpoint(fmt.Sprintf("containers/%s/wait", nameOrID)), "application/json", nil)
+	// TODO when returns are ironed out, we can should use the newRequest approach
+	_, err := http.Post(c.makeEndpoint(fmt.Sprintf("containers/%s/wait", nameOrID)), "application/json", nil) // nolint
```
mheon: Shouldn't this have a closeResponseBody?

baude: it isnt complete, hence the note

PR #4879, opened 2020-01-16
hack/install_golangci.sh: check env vars

vrothberg:
Also make it executable and `set -eo pipefail`.

Based on a comment from @cevich -> https://github.com/containers/libpod/pull/4870#pullrequestreview-343315780

### DIFF
--- a/hack/install_golangci.sh
+++ b/hack/install_golangci.sh
@@ -1,5 +1,17 @@
-#!/bin/bash -e
+#!/bin/bash
+
+if [ -z "$VERSION" ]; then
+	echo \$VERSION is empty
+	exit 1
+fi
+
+if [ -z "$GOBIN" ]; then
+	echo \$GOBIN is empty
+	exit 1
+fi
+
 $GOBIN/golangci-lint --version | grep $VERSION
 if [ $?  -ne 0 ]; then
+	set -e
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOBIN v$VERSION
 fi


mheon:
LGTM

hack/install_golangci.sh
```diff
@@ -1,5 +1,17 @@
-#!/bin/bash -e
+#!/bin/bash
+
```
cevich: We want `set -eo pipefail` here.

hack/install_golangci.sh
```diff
@@ -1,5 +1,17 @@
...
+fi
+
+if [ -z "$GOBIN" ]; then
+	echo \$GOBIN is empty
+	exit 1
+fi
+
 $GOBIN/golangci-lint --version | grep $VERSION
 if [ $?  -ne 0 ]; then
+	set -e
```
cevich: kill this line.

hack/install_golangci.sh
vrothberg: If we do this, we'll fail in the version check below when golangci-lint is not installed. I tried that but it didn't work.

hack/install_golangci.sh
vrothberg: See comment above. This is command that should not fail, all above are fine.

cevich:
For your own knowledge, this is a bash gotcha (from the man page):

> The  return  status  of  a  pipeline  is  the exit status of the last command, unless the pipefail option is enabled. If pipefail is enabled, the pipeline's return status is the value of the last (rightmost)  command to exit with a non-zero status, or zero if all commands exit successfully.

So my understanding is without pipefail (the default) a ` foo | bar` will give you a zero-exit if `foo` fails but not `bar`.  When pipefail is enabled (not the default), then either `foo` or `bar` failing will result in a non-zero exit.

@edsantiago please correct my understanding, you know this stuff better :smile:

hack/install_golangci.sh
edsantiago: I don't have a strong opinion on setting `pipefail`, but if it's really desired, then one solution could be...

hack/install_golangci.sh
edsantiago: ...something like this:
```bash
linter=$GOBIN/golangci-ling
if [ ! -x $linter ]; then
    echo "linter unavailable: $linter"
    exit 1
fi
if ! grep "$VERSION" <<($linter --version); then
    echo "linter is the wrong version"
    $linter --version
    exit 1
fi

PR #4881, opened 2020-01-16
v2: don't block sigterm and add the service to `make binaries`

vrothberg:
@baude @jwhonce @mheon @TomSweeneyRedHat @rhatdan PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -322,7 +322,7 @@ system.test-binary: .install.ginkgo
 vagrant-check:
 	BOX=$(BOX) sh ./vagrant.sh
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
Although realize that service, should be going away soon.  I think we should move this to 
`podman service`, ASAP

vrothberg:
> LGTM
> Although realize that service, should be going away soon. I think we should move this to
> `podman service`, ASAP

I changed it in `pkg/api/server/server.go`, so a `podman service` will benefit from the change as well.

PR #4882, opened 2020-01-16
Cirrus: remove workaround for cleaning /go/bin

vrothberg:
Remove the temporary workaround for cleaning /go/bin in the gating task.
The workaround was added to make sure that we're always installing the
latest tools in `make install.tools`.  The gating image does not ship
with these pre-installed tools anymore which is now causing errors in
the gating task as the `rm` is missing the `--force` flag.


vrothberg:
@baude @mheon @cevich @TomSweeneyRedHat @rhatdan ... forgetting the `--force` was an expensive oversight. Now all open PRs must be rebased :-( Sorry for the inconvenience.

PR #4884, opened 2020-01-16
systemdgen: add --ignore flag to generic services

vrothberg:
The --ignore flag lets Podman ignore errors when a specified container
does not exist (anymore).  That's a nice addition to generic services
generated via the --new flag.  Those services create new containers and
can hence allow user to manually remove a container; may it only be by
accident.

The important part of using the --ignore flag is that Podman will exit 0
which plays nicer with most restart policies; a non-zero exit may yield
systemd to restart the entire service which is arguably wrong if the
user manually deletes the container.

If desired, users can still alter the generated files.


mheon:
LGTM

vrothberg:
Curious ... I didn't fail locally

vrothberg:
Eeeh...the e2e tests failed. Will fix it up and repush.

vrothberg:
Ready to merge

Issue #4886, opened 2020-01-16
CI flakes in test/e2e/run_signal_test.go


vrothberg:
I've seen the "signals are not forwarded to container with sig-proxy false" flake a couple of times lately. Let's have a look and investigate the test.

/var/tmp/go/src/github.com/containers/libpod/test/e2e/run_signal_test.go:135

PR #4887, opened 2020-01-16
e2e/run_signal_test.go: make it more robust

vrothberg:
Make the signal test more robust by just checking that the container's
exit code is non-zero.  There are two possible exit codes (i.e., 130 and
137) depending on how the container is being killed, which is likely
responsible for CI flakes.

Fixes: #4886


vrothberg:
Ready to merge

PR #4890, opened 2020-01-16
WIP: Swagger docs: fix inconsistencies

edsantiago:
The swagger doc headers are, if I may be blunt, evil.
There is far too much manual muckery; much of those
comments should be autogenerated.

I'm working on that: a script that parses the r.Handle()
registrations and autogenerates a set of swagger comments.
It's still a WIP, and I'd like to extend it much more
so the parameter descriptions can be written in a more
human-friendly way, but that's not going to happen
before devconf. ITM here are a ton of problems, some
real, some perhaps not, that my script has found and
fixed.

Some are (possibly minor, I don't know) capitalization
errors. Some are errors in the documented method (GET vs POST).
Some are errors in the endpoint. I believe that all of
them should be reviewed carefully and, as appropriate,
merged. (You do not need to merge this PR, but please
at least incorporate the necessary fixes into your trees).

### DIFF
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -10,7 +10,7 @@ import (
 )
 
 func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
-	// swagger:operation POST /containers/create compat containerCreate
+	// swagger:operation POST /containers/create compat createContainer
 	// ---
 	//   summary: Create a container
 	//   tags:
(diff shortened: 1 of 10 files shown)


edsantiago:
@baude @jwhonce PTAL

edsantiago:
[At least one test failure is real; I'm not going to bother fixing since this is just an informational please-steal-this-code PR]

PR #4892, opened 2020-01-17
Minor: Bugfix in upload image

cevich:
The release upload process always involves two filenames, however the
second filename might (someday) be optional.  The code allowed for this,
however input validation did not.  This change fixes the validation.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #4894, opened 2020-01-17
Tests for API v2

edsantiago:
Initial framework for testing the version 2 (HTTP) API.
Includes a collection of tests for some of the existing
endpoints. Not all tests are currently passing.

### DIFF
--- a/test/apiv2/00-TEMPLATE
+++ b/test/apiv2/00-TEMPLATE
@@ -1,0 +1,6 @@
+# -*- sh -*-
+#
+# FIXME: one-line description of the purpose of this file
+#
+
+# vim: filetype=sh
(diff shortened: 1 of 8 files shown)


edsantiago:
I'm not terribly happy with the file-descriptor muckery[1]. Suggestions welcome.

 [1] Because subtests are expected to run `podman` or other commands that may emit spurious output, and said output would affect the TAP (`ok/not ok`) stream, I've done some acrobatics to restrict stdout/stderr.

stbischof:
@rhatdan you wrote a LGTM but the /lgtm tag is missing. is this by exident?

baude:
> @rhatdan you wrote a LGTM but the /lgtm tag is missing. is this by exident?

our process generally requires two lgtms.  the second has the leading slash which indicates it is approved to merge.

rhatdan:
@baude Could you give this the aforementioned lgtm

PR #4901, opened 2020-01-18
Make it easier to override git variables

### LINKED ISSUE #4787: Build gets the wrong version, when built from tarball
Seems like the build script picks up the version of the **build system** instead...

When using something like (a branch of) [buildroot](https://buildroot.org/), that builds from dist tarballs.

It also uses "git fetch", which can be quite costly (we are using a shallow clone)

So it would be better if "make podman" didn't do all those magic git commands ?

----

The workaround is to provide the required version in the "CIRRUS_TAG" variable

```bash
    VERSION*)
        OUTPUT="${CIRRUS_TAG:-$(git fetch --tags && git describe HEAD 2> /dev/null)}"
        ;;
    NUMBER*)
        OUTPUT="$($0 VERSION | sed 's/-.*//')"
        ;;
```

Or to give up a provide both RELEASE_VERSION and RELEASE_NUMBER

```make
RELEASE_VERSION ?= $(shell hack/get_release_info.sh VERSION)
RELEASE_NUMBER ?= $(shell hack/get_release_info.sh NUMBER|sed -e 's/^v\(.*\)/\1/')
```

Like currently done for the GIT_COMMIT variable, similar issue (wrong git repo)

```make
GIT_COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),${COMMIT_NO}-dirty,${COMMIT_NO})
```

https://github.com/kubernetes/minikube/tree/master/deploy/iso/minikube-iso/package/podman

afbjorklund:
Use $VERSION instead of the "$CIRRUS_TAG".
(the other variable to set is $GIT_COMMIT)

Avoid doing an unnecessary `git fetch --tags`.
Move the NUMBER sed hacks to the shell script.

Allows building from e.g. tarball, without git.

Also allows building within a _different_ git.

Example use:

```shell
make VERSION=v1.7.0 GIT_COMMIT=b7ce1157b00af09f4a09e39b377aa3abff46ee05
```

Closes #4787

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ CHANGELOG_BASE ?= HEAD~
 CHANGELOG_TARGET ?= HEAD
 PROJECT := github.com/containers/libpod
 GIT_BASE_BRANCH ?= origin/master
+VERSION ?= $(shell git describe HEAD 2>/dev/null)
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN ?= $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 LIBPOD_IMAGE ?= libpod_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
@@ -89,7 +90,7 @@ LIBSECCOMP_COMMIT := v2.3.3
 GINKGOTIMEOUT ?= -timeout=90m
 
 RELEASE_VERSION ?= $(shell hack/get_release_info.sh VERSION)
-RELEASE_NUMBER ?= $(shell hack/get_release_info.sh NUMBER|sed -e 's/^v\(.*\)/\1/')
+RELEASE_NUMBER ?= $(shell hack/get_release_info.sh NUMBER)
 RELEASE_DIST ?= $(shell hack/get_release_info.sh DIST)
 RELEASE_DIST_VER ?= $(shell hack/get_release_info.sh DIST_VER)
 RELEASE_ARCH ?= $(shell hack/get_release_info.sh ARCH)
--- a/hack/get_release_info.sh
+++ b/hack/get_release_info.sh
@@ -19,10 +19,16 @@ unset OUTPUT
 case "$1" in
     # Wild-card suffix needed by valid_args() e.g. possible bad grep of "$(echo $FOO)"
     VERSION*)
-        OUTPUT="${CIRRUS_TAG:-$(git fetch --tags && git describe HEAD 2> /dev/null)}"
+        # v1.7.0-152-g9be6430
+        OUTPUT="${CIRRUS_TAG:-$VERSION}"
+        ;;
+    TAG*)
+        # v1.7.0
+        OUTPUT="$($0 VERSION | sed 's/-.*//')"
         ;;
     NUMBER*)
-        OUTPUT="$($0 VERSION | sed 's/-.*//')"
+        # 1.7.0
+        OUTPUT="$($0 TAG | sed -e 's/^v\(.*\)/\1/')"
         ;;
     DIST_VER*)
         OUTPUT="$(source /etc/os-release; echo $VERSION_ID | cut -d '.' -f 1)"


afbjorklund:
It would still be easier if there was a VERSION file in the repository, updated by the release script ?

This is used in other projects, such as conmon: `VERSION := $(shell cat VERSION)`

Then there is no risk of the user giving wrong information, and also the README can be updated:

`* [Latest Version: 1.6.3](https://github.com/containers/libpod/releases/latest)`

----

Too many manual steps now, and easy to miss one:

```go
// Version is the version of the build.
// NOTE: remember to bump the version at the top
// of the top-level README.md file when this is
// bumped.
const Version = "1.7.0"
```

Although this was mostly about the git clone requirement.


/assign @vrothberg

Makefile
```diff
@@ -9,6 +9,7 @@ CHANGELOG_BASE ?= HEAD~
 CHANGELOG_TARGET ?= HEAD
 PROJECT := github.com/containers/libpod
 GIT_BASE_BRANCH ?= origin/master
+VERSION ?= $(shell git describe HEAD 2>/dev/null)
```
vrothberg: Should we keep the `git fetch --tags` here?

afbjorklund: I'm not sure _when_ it is needed, but you can add it back if you want.

>        By default, any tag that points into the histories being fetched is
>        also fetched; the effect is to fetch tags that point at branches that
>        you are interested in. This default behavior can be changed by using
>        the --tags or --no-tags options or by configuring remote.<name>.tagOpt.

I had to remove it, since it was fetching **all** tags and all _their_ history...

`git clone --depth=1 --branch=2019.02.8 https://github.com/buildroot/buildroot `

afbjorklund:
> Cirrus isn't yet happy. I assume there are some leftover references in `contrib/cirrus/lib.sh`.

Pretty sure that I did not touch $RELEASE_ARCH at all.

```
[+0000s] Parsing actual_release.txt contents: X-RELEASE-INFO: podman  fedora 30 amd64
[+0000s] ************************************************
[+0000s] >>>>> FATAL: main() requires $RELEASE_ARCH to be non-empty
[+0000s] ************************************************
```

But it seems that it is indeed missing the $RELEASE_VERSION

Probably missing an `export` ?

afbjorklund:
Definitely made it _worse_ with this change, so closing for now.

cevich:
FWIW: There are a bunch of values that come from `release.txt` which is produced by the Makefile.  I had a lot of trouble getting this stuff working across both Fedora and Ubuntu, so it's easily possible there is an excessive amount of duct-tape :confused: 

I do like your motivations in this PR though, maybe try doing them one at a time instead of all at once.  IIRC, cirrus doesn't fetch tag info at all when cloning repos.  But that info. isn't needed until the build/upload release process happens.  Both of those have their own (obviously) named scripts under `contrib/cirrus/` if you want to hard-code a `git fetch` or two into them instead of the Makefile, that's fine by me.

HTH and thanks for making this effort.

afbjorklund:
I noticed in the issue that the Fedora and Ubuntu packages are still missing build information, like git commit and build time. So it is not _easy_ to get this right. Haven't checked the [opensuse](https://build.opensuse.org/project/show/devel:kubic:libcontainers:stable) packages yet...

podman-1.6.2-2.fc31.x86_64.rpm
```json
{
    "RemoteAPIVersion": 1,
    "Version": "1.6.2",
    "GoVersion": "go1.13.1",
    "GitCommit": "",
    "Built": 0,
    "OsArch": "linux/amd64"
}
```

podman_1.6.2-1\~ubuntu16.04\~ppa1_amd64.deb
```json
{
    "RemoteAPIVersion": 1,
    "Version": "1.6.2",
    "GoVersion": "go1.10.4",
    "GitCommit": "",
    "Built": 0,
    "OsArch": "linux/amd64"
}
```

cevich:
I agree it's very much not easy, I spent a very long time tinkering with this kind of thing.

cevich:
ya, and then some dumb humans went in and added some "Features" :grinning: 

I'm happy to help advise from the automation/cirrus/scripts perspective, just mention me or ping me on irc.  If it would help, I'm open to extending the fields in `release.txt` if that might make things easier.  AFAIK, that file is only consumed by one or two scripts that should be easy to update.

PR #4906, opened 2020-01-20
lint: activate bashate

ssbarnea:
- Activate bashate and resolves rule `E010`.
- Temporary disable other non-conforming rules, as they will be
  addressed in follow-ups
- Fixes python2 bug in embedded get_env_vars()
- Includes newline auto-fixed introduced by a previous merge

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,3 +10,22 @@ repos:
       - id: check-byte-order-marker
       - id: check-executables-have-shebangs
       - id: check-merge-conflict
+  - repo: https://github.com/openstack-dev/bashate.git
+    rev: "0.6.0"
+    hooks:
+      - id: bashate
+        entry: bashate --error . --ignore=E003,E002,E006,E011,E020,E041,E042
+        # Run bashate check for all bash scripts
+        # Ignores the following rules:
+        # E006: Line longer than 79 columns (as many scripts use jinja
+        #       templating, this is very difficult)
+        # E040: Syntax error determined using `bash -n` (as many scripts
+        #       use jinja templating, this will often fail and the syntax
+        #       error will be discovered in execution anyway)
+        # TODO(ssbarnea): remove temporary skips made during adoption:
+        # E002 Tab indents
+        # E010 The "do" should be on same line as for
+        # E011 Then keyword is not on same line as if or elif keyword
+        # E020 Function declaration not in format ^function name {$
+        # E041 Arithmetic expansion using $[ is deprecated for $((
+        # E042 local declaration hides errors
--- a/test/system/helpers.bash
+++ b/test/system/helpers.bash
@@ -1,3 +1,4 @@
+#!/usr/bin/env false
 # -*- bash -*-
 
 # Podman command to run; may be podman-remote
--- a/Makefile
+++ b/Makefile
@@ -536,7 +536,7 @@ validate.completions: completions/bash/podman
 	. completions/bash/podman
 	if [ -x /bin/zsh ]; then /bin/zsh completions/zsh/_podman; fi
 
-validate: gofmt .gitvalidation validate.completions golangci-lint man-page-check
+validate: lint gofmt .gitvalidation validate.completions golangci-lint man-page-check
 
 build-all-new-commits:
 	# Validate that all the commits build on top of $(GIT_BASE_BRANCH)
--- a/contrib/cirrus/build_vm_images.sh
+++ b/contrib/cirrus/build_vm_images.sh
@@ -28,8 +28,7 @@ cd "$GOSRC/$PACKER_BASE"
 # Add/update labels on base-images used in this build to prevent premature deletion
 ARGS="
 "
-for base_image_var in $BASE_IMAGE_VARS
-do
+for base_image_var in $BASE_IMAGE_VARS; do
     # See entrypoint.sh in contrib/imgts and contrib/imgprune
     # These updates can take a while, run them in the background, check later
     gcloud compute images update \
--- a/contrib/cirrus/check_image.sh
+++ b/contrib/cirrus/check_image.sh
@@ -43,8 +43,7 @@ do
         "$(systemctl list-unit-files --no-legend $REQ_UNIT)" = "$REQ_UNIT enabled" || let "NFAILS+=1"
 done
 
-for evil_unit in $EVIL_UNITS
-do
+for evil_unit in $EVIL_UNITS; do
     # Exits zero if any unit matching pattern is running
     unit_status=$(systemctl is-active $evil_unit &> /dev/null; echo $?)
     item_test "No $evil_unit unit is present or active:" "$unit_status" -ne "0" || let "NFAILS+=1"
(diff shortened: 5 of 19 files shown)


rhatdan:
LGTM
although not sure why images test failed.

mheon:
@cevich @edsantiago PTAL

TomSweeneyRedHat:
LGTM
The images tests seem to be failing everywhere in Libpod.  I'm not sure what's what there.

cevich:
looking...

test/system/helpers.bash
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bats
```
cevich: IIRC @edsantiago said doing this is bad-juju since it implies the file is a script that people should execute.  Is there a way to block the bash lint-checking for this file?

cevich:
...Most of the changes here are purely cosmetic/style related but I can live with them.  Want to get @edsantiago to confirm on `test/system/helpers.bash`.  I think he had a workaround of sorts for this finding elsewhere, but I can't remember what it was.  Otherwise if we could just ignore that file, it could be an option as well.

cevich:
Otherwise LGTM

test/system/helpers.bash
ssbarnea: My impression was that that file was a bats test file but I am more than happy to use another approach. Maybe we could change its extension? I would prefer to avoid excludes because they are hard to maintain and it would make impossible to run bashate directly, without pre-commit.

cevich: My understanding is this file is special.  We really need @edsantiago input here since he knows bats the best.

edsantiago:
We've been through this before. In your first PR I mentioned that if this change is absolutely necessary it should (1) include a command-line option that prevents invoking this as a script, and (2) a good comment explaining that this is not a user-runnable script and describing the reason for the shebang (i.e. lint). Please see your previous PR for details.

ssbarnea:
> We've been through this before. In your first PR I mentioned that if this change is absolutely necessary it should (1) include a command-line option that prevents invoking this as a script, and (2) a good comment explaining that this is not a user-runnable script and describing the reason for the shebang (i.e. lint). Please see your previous PR for details.

Did you look the proposed shebang for the file? I think it documents very clear that is not a bash script. Look again. It is a bats file, let it be known.

Bashate will and does not run scripts, bit it looks for them. Shebang line is what any executable script should have, regardless which interpreter they use.

rhatdan:
@ssbarnea Please rebase.  I believe this needs a rebase to get the images test to pass.

cevich:
> Bashate will and does not run scripts, bit it looks for them. Shebang line is what any executable script should have, regardless which interpreter they use.

I believe that's exactly the problem.  The `helpers` file is not a script nor an executable.  It's a library of sorts and specific for interpretation by bats.

@edsantiago is there any chance of being able to stick a different extension on the `helpers.bash` file (maybe like `helpers.bats` or `helpers.lib` or something)?  That may be an easier solution that trying to maintain an exclusion-list for Bashate.

edsantiago:
@ssbarnea @cevich no, there is no way to tell BATS to change the `.bash` extension for helpers. And I'm OK with that: the bug here is that some lint code is demanding that a file with the extension `.bash` include a shebang, even if the file is not `+x`. This is a linter bug, pure and simple. It is unfortunate that we have to pollute our code by working around it.

@ssbarnea please try this:
```bash
#!/usr/bin/env -S bash -c "echo This is not an actual bash script;exit 1"
#
# This is not an executable script. Do not try to invoke it. The shebang above
# is required by a broken lint package. We apologize for the inconvenience.
#
```

cevich:
> It is unfortunate that we have to pollute our code by working around it.

Yeah, this is a trade-off vs maintaining an exception list for the linter, or using a different linter.  Arguably the shebang is/will be slightly easier to maintain long-term (by doing nothing), than the list (possibly updating it if the dir/file moves), or changing to a less-popular linter .  So...the least worst evil among a pile of dead suspected-witches :confounded:

.pre-commit-config.yaml
```diff
@@ -10,3 +10,22 @@ repos:
       - id: check-byte-order-marker
       - id: check-executables-have-shebangs
       - id: check-merge-conflict
+  - repo: https://github.com/openstack-dev/bashate.git
+    rev: "0.6.0"
+    hooks:
+      - id: bashate
+        entry: bashate --error . --ignore=E003,E002,E006,E011,E020,E041,E042
```
vrothberg: Please make this part of the Makefile (i.e., `make validate`). I don't want to play ping-pong with the CI but really want a local make target to run tests.

vrothberg: By having it in `make validate`, running the linter will then automatically be part of the gating task in Cirrus.

ssbarnea: @edsantiago In regards to `.bash` extension I think you will really struggle to defend the case that this is an extension owned by bats tools. Yes, is a bug, but not in bashate, is a bug in bats that makes this owneship assumpts. They could as well use `.bats` instead which does not overlap with others.

Sure I can replace the sheband with another variant, but I am kinda become tired of nit-picking. The review allows edits from others and they are more than welcome. As far we finish it faster the better, already spend more time on it than expected.

Now I do have some questions regarding the complex nature of the makefile and how is used on CI. Maybe someone can build a mermaid diagram with the dependencies?

My expectation was that CI would run `make lint` somewhere, preferably as soon as possible. Most projects have lint job but should I guess that libpod one is called validate, right?

ssbarnea:
> #!/usr/bin/env -S bash -c "echo This is not an actual bash script;exit 1"

This is much much worse, you are using bash itself to run the file! Guess what, vscode now is now reporting various bugs inside the file from both shellcheck and bashate, because the file is now an official bash file. Your proposed shebang life made it!

ssbarnea:
For fixing the CI, i raised https://github.com/containers/libpod/pull/5022

edsantiago:
> ^-- SC1008: This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh.
> ^-- SC2096: On most OS, shebangs can only specify a single parameter.

Fair enough; it was an ugly hack. I guess if the linter wants a shebang -- and this is a linter bug, not a BATS bug -- then let's give it a shebang. `#!/bin/bash` is fine. But please, I implore, leave the two-line comment (or equivalent) near the top. It might save someone time in the future, if they're not familiar with BATS.

As for why it's a linter bug (this is low priority, but just an FYI): there is a longstanding convention of storing library functions in a non-executable file with a `.sh` extension. BATS extends this convention to `.bash`, so a `load foo` statement in a `.bats` test will translate to `source foo.bash`. Perhaps it was an imperfect decision on the part of the BATS creators, perhaps they could have used `.bats-library` or `.potrzebie` instead, but they chose `.bash` as something meaningful, that stands out as different in a directory full of `.bats` files; and at this point that is unlikely to change. I believe that a linter should understand that a non-executable file should not have a shebang. 

Thanks again for your perseverance. I realize there are a lot of details to consider; our common goal is to spend time early to get those details right, so we don't have to spend (usually more) time in the future figuring out breakages.

rhatdan:
@ssbarnea Still working on this?

vrothberg:
Friendly ping.

ssbarnea:
I am not working on this, so anyone is welcomed to takeover this.

cevich:
I think we can probably just close this for now then.  If someone wants to take up the helm, I'm happy to help as always, but I think we have 'bigger fish to fry' for the near-term.

PR #4920, opened 2020-01-21
Use cgroupv2 super magic from golang.org/x/sys/unix

saschagrunert:
We can use this constant from the already existing sys/unix package
instead of defining it by our own.


TomSweeneyRedHat:
LGTM, nice clean up

PR #4948, opened 2020-01-23
seccomp policy: expect profile in config label

vrothberg:
Move the seccomp profile from a manifest annotation to a config label.
This way, we can support it for Docker images as well and provide an
easy way to add that data via Dockerfiles.

@rhatdan @nalind PTAL

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -31,9 +31,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// seccompAnnotationKey is the key of the image annotation embedding a seccomp
+// seccompLabelKey is the key of the image annotation embedding a seccomp
 // profile.
-const seccompAnnotationKey = "io.containers.seccomp.profile"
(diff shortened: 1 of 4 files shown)


cmd/podman/shared/create.go
```diff
@@ -709,7 +709,7 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 
 	// SECCOMP
 	if data != nil {
-		if value, exists := data.Annotations[seccompAnnotationKey]; exists {
+		if value, exists := labels[seccompLabelKey]; exists {
```
mheon: Should we check both? It doesn't seem like it would hurt.

vrothberg: I would prefer to have just one place, especially when aiming to get that into the OCI. Luckily, nobody is using it yet :-)

mheon:
LGTM

vrothberg:
Just missing a "/lgtm" :)

PR #4953, opened 2020-01-23
Review corrections pass #2

baude:
Add API review comments to correct documentation and endpoints.  Also, add a libpode prune method to reduce code duplication.  Only used right now for the API but when the remote client is wired, we will switch over there too.

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -836,3 +836,44 @@ func (r *Runtime) GetLatestContainer() (*Container, error) {
 	}
 	return ctrs[lastCreatedIndex], nil
 }
+
+// PruneContainers removes stopped and exited containers from localstorage.  A set of optional filters
+// can be provided to be more granular.
+func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int64, map[string]error, error) {
+	pruneErrors := make(map[string]error)
+	prunedContainers := make(map[string]int64)
+	// We add getting the exited and stopped containers via a filter
+	containerStateFilter := func(c *Container) bool {
+		if c.PodID() != "" {
+			return false
+		}
+		state, err := c.State()
+		if err != nil {
+			logrus.Error(err)
+			return false
+		}
+		if state == define.ContainerStateStopped || state == define.ContainerStateExited {
+			return true
+		}
+		return false
+	}
+	filterFuncs = append(filterFuncs, containerStateFilter)
+	delContainers, err := r.GetContainers(filterFuncs...)
+	if err != nil {
+		return nil, nil, err
+	}
+	for _, c := range delContainers {
+		ctr := c
+		size, err := ctr.RWSize()
+		if err != nil {
+			pruneErrors[ctr.ID()] = err
+			continue
+		}
+		err = r.RemoveContainer(context.Background(), ctr, false, false)
+		pruneErrors[ctr.ID()] = err
+		if err != nil {
+			prunedContainers[ctr.ID()] = size
+		}
+	}
+	return prunedContainers, pruneErrors, nil
+}
--- a/pkg/api/handlers/containers.go
+++ b/pkg/api/handlers/containers.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"fmt"
+	"github.com/docker/docker/api/types"
 	"net/http"
 
 	"github.com/containers/libpod/libpod"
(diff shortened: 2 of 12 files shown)


libpod/container_commit.go
```diff
@@ -146,7 +146,7 @@ func (c *Container) Commit(ctx context.Context, destImage string, options Contai
 	// Workdir
 	importBuilder.SetWorkDir(c.config.Spec.Process.Cwd)
 
-	// Process user changes
+	// process user changes
```
mheon: Find/replace issue?

libpod/container_inspect.go
```diff
@@ -670,7 +670,7 @@ func (c *Container) getContainerInspectData(size bool, driverData *driver.Data)
 		return nil, err
 	}
 
-	// Process is allowed to be nil in the stateSpec
+	// process is allowed to be nil in the stateSpec
```
mheon: More find/replace problems?

libpod/define/container.go
```diff
@@ -0,0 +1,8 @@
+package define
+
+// PruneContainerResponse describes the results of a container pruning
+type PruneContainerResponse struct {
```
mheon: Why in define?

libpod/runtime_ctr.go
```diff
@@ -836,3 +836,49 @@ func (r *Runtime) GetLatestContainer() (*Container, error) {
...
+		prunedResponses []define.PruneContainerResponse
+	)
+	// We add getting the exited and stopped containers via a filter
+	containerStateFilter := func(c *Container) bool {
+		state, err := c.State()
+		if err != nil {
+			logrus.Error(err)
+			return false
+		}
+		if c.PodID() != "" {
```
mheon: Move this to the beginning, before c.State() - makes things a bit faster

libpod/define/container.go
mheon: Also, for consistency with the rest of libpod, we probably want to return:
[]string, map[string]error, int64, error
(Array of IDs of successes, map of ID to error for failed containers, total space freed, fatal errors that prevented us from pruning any container)

We can convert to this in adapter

libpod/runtime_ctr.go
```diff
@@ -836,3 +836,49 @@ func (r *Runtime) GetLatestContainer() (*Container, error) {
...
+	delContainers, err := r.GetContainers(filterFuncs...)
+	if err != nil {
+		return nil, err
+	}
+	for _, c := range delContainers {
+		ctr := c
+		response := define.PruneContainerResponse{
+			ID: ctr.ID(),
+		}
+		size, err := ctr.RootFsSize()
```
mheon: Shouldn't this be RwSize() - we're not removing the image the container is based on, so I don't think we freed that space.

pkg/api/handlers/containers.go
```diff
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"fmt"
+	"github.com/docker/docker/api/types"
```
mheon: Import ordering

libpod/runtime_ctr.go
```diff
@@ -836,3 +836,49 @@ func (r *Runtime) GetLatestContainer() (*Container, error) {
...
+		if err != nil {
+			response.PruneError = err
+			prunedResponses = append(prunedResponses, response)
+			continue
+		}
+		err = r.RemoveContainer(context.Background(), ctr, false, false)
+		response.PruneError = err
+		response.Size = size
```
mheon: Shouldn't this only be if PruneError == nil?

pkg/api/handlers/utils/containers.go
```diff
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"fmt"
+	"github.com/containers/libpod/cmd/podman/shared"
```
mheon: Import ordering

pkg/api/handlers/containers.go
```diff
@@ -192,3 +193,55 @@ func RestartContainer(w http.ResponseWriter, r *http.Request) {
...
+	if utils.IsLibpodRequest(r) {
+		var response []LibpodContainersPruneReport
+		for ctrID, size := range prunedContainers {
+			response = append(response, LibpodContainersPruneReport{ID: ctrID, SpaceReclaimed: size})
+		}
+		for ctrID, err := range pruneErrors {
+			response = append(response, LibpodContainersPruneReport{ID: ctrID, PruneError: err.Error()})
+		}
+		utils.WriteResponse(w, http.StatusOK, response)
+
```
jwhonce: If you are writing the Response here, you need to return.  The one at the end of the function will be ignored.

pkg/api/handlers/generic/containers_create.go
```diff
@@ -40,7 +40,9 @@ func CreateContainer(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Decode()"))
 		return
 	}
-
+	if len(input.HostConfig.Links) > 0 {
+		utils.Error(w, utils.ErrLinkNotSupport.Error(), http.StatusBadRequest, errors.Wrapf(utils.ErrLinkNotSupport, "bad parameter"))
```
jwhonce: We should only error if Links is given and true.

pkg/api/handlers/libpod/containers.go
```diff
@@ -117,19 +131,13 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
...
-	utils.WriteResponse(w, http.StatusNoContent, "")
-}
-
-func PruneContainers(w http.ResponseWriter, r *http.Request) {
-	// TODO Needs rebase to get  filers; Also would be handy to define
-	// an actual libpod container prune method.
-	// force
-	// filters
+	// TODO can we straight up return an int32 like this ?
+	utils.WriteResponse(w, http.StatusOK, exitCode)
```
jwhonce: Do you really want the client to really get an integer or a string with the value of the integer?

pkg/api/server/register_containers.go
```diff
@@ -301,7 +316,8 @@ func (s *APIServer) RegisterContainersHandlers(r *mux.Router) error {
 	//  - in: query
 	//    name: detachKeys
 	//    type: string
-	//    description: needs description
+	//    description: "Override the key sequence for detaching a container. Format is a single character [a-Z] or ctrl-<value> where <value> is one of: a-z, @, ^, [, , or _."
```
jwhonce: The value of description should not be quoted.

baude: it has to be because something in there murders the go swagger generation

pkg/api/handlers/generic/containers_create.go
baude: links are not bool ... they are []string so in this case, i think checking the len is ok?

pkg/api/handlers/libpod/containers.go
baude: integer i believe.

pkg/api/handlers/generic/containers_create.go
jwhonce: Ah, other places in API link is boolean.  Sorry for confusion.

pkg/api/server/register_containers.go
jwhonce: Learn something new every day.

pkg/api/handlers/generic/containers.go
```diff
@@ -47,14 +44,40 @@ func RemoveContainer(w http.ResponseWriter, r *http.Request) {
...
+		containers, err = runtime.GetAllContainers()
+	} else {
+		containers, err = runtime.GetRunningContainers()
+	}
 	if err != nil {
 		utils.InternalServerError(w, err)
 		return
 	}
-
+	if _, found := mux.Vars(r)["Limit"]; found {
```
jwhonce: "Limit" should be lowercase, if query tags are correct.

pkg/api/handlers/libpod/containers.go
jwhonce: Then how are you dealing with endianness?

baude: switched to string

PR #4964, opened 2020-01-24
apparmor: allow receiving of signals from 'podman kill'

saschagrunert:
In newer kernels, AppArmor will reject attempts to send signals to a
container because the signal originated from outside of that AppArmor
profile. Correct this by allowing all unconfined signals to be received.

Ref https://github.com/moby/moby/pull/37831/files


mheon:
LGTM but would like an ack from @vrothberg before merge

TomSweeneyRedHat:
LGTM and ditto @mheon's ack request

Issue #4783, opened 2020-01-03
Depends on obsolete/archived "github.com/fatih/camelcase"


onlyjob:
[github.com/fatih/camelcase](https://github.com/fatih/camelcase) is an unmaintained/archived project.

Let's reduce dependency on dead code please.

mheon:
Seems to be used primarily to generate headers for CLI output in commands like `ps`, `stats`, `history`. One direct invocation, six or seven calls to the function it's used in.

rhatdan:
@onlyjob do you know of an alternative?

onlyjob:
I wish I knew...

onlyjob:
The closest I can think of is [github.com/mattn/go-shellwords](https://github.com/mattn/go-shellwords)

vrothberg:
The package is _really_ small, so I forked it into libpod: https://github.com/containers/libpod/pull/4968

onlyjob:
Wouldn't it be better to switch to `mattn/go-shellwords`? It appears to be already used by Buildah and `containers/storage`.

vrothberg:
@onlyjob, `go-shellwords` seems to behave differently. It's used to split a single-string shell command into an array, while `camelcase` is splitting a camelcase borders.

PR #4968, opened 2020-01-25
fork fatih/camelcase

vrothberg:
faith/camelcase has been archived and is no longer maintained.
The package is sufficiently small and self-contained enough to
maintain it in libpod.

Fixes: #4783


TomSweeneyRedHat:
LGTM and happy test buttons.
My only question is if it would make more sense to put this under containers/camelcase rather than under libpod/pkg?

PR #4973, opened 2020-01-26
Throw error on invalid sort value

rhatdan:
We define the valid sort values, so we should throw an error
on invalid sort values.

### DIFF
--- a/cmd/podman/images.go
+++ b/cmd/podman/images.go
@@ -159,6 +159,21 @@ func imagesCmd(c *cliconfig.ImagesValues) error {
 		filters = append(filters, fmt.Sprintf("reference=%s", image))
 	}
 
+	var sortValues = map[string]bool{
+		"created":    true,
+		"id":         true,
+		"repository": true,
+		"size":       true,
+		"tag":        true,
+	}
+	if !sortValues[c.Sort] {
+		keys := make([]string, 0, len(sortValues))
+		for k := range sortValues {
+			keys = append(keys, k)
+		}
+		return errors.Errorf("invalid sort value %q,  required values: %s", c.Sort, strings.Join(keys, ", "))
+	}
+
 	opts := imagesOptions{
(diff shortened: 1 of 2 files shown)


rhatdan:
@rpjday @mheon @baude @vrothberg @TomSweeneyRedHat @jwhonce @QiWang19 PTAL

cmd/podman/images.go
```diff
@@ -159,6 +159,17 @@ func imagesCmd(c *cliconfig.ImagesValues) error {
 		filters = append(filters, fmt.Sprintf("reference=%s", image))
 	}
 
+	var sortValues = map[string]bool{
+		"id":         true,
+		"size":       true,
+		"tag":        true,
+		"repository": true,
+		"created":    true,
+	}
```
TomSweeneyRedHat: nit, could we change the above to top down alpha order?  created, id, repository, size, tag

cmd/podman/images.go
```diff
@@ -159,6 +159,17 @@ func imagesCmd(c *cliconfig.ImagesValues) error {
...
+	if !sortValues[c.Sort] {
+		return errors.Errorf("invalid sort value %q", c.Sort)
```
vrothberg: We could print valid values in the message.

cmd/podman/images.go
rhatdan: Fixed

cmd/podman/images.go
rhatdan: Fixed

rhatdan:
Added lots of tests.
@vrothberg @mheon @TomSweeneyRedHat @baude @jwhonce PTAL
This is an answer to an issue on the mailing list.

cmd/podman/images.go
TomSweeneyRedHat: ty!

PR #4980, opened 2020-01-26
[CI:DOCS]Binding overhauls

baude:
Add binding for networks and begin documentation for binding methods for godoc.  Also, add major functions to their own subpackages so reduce the amount of of method confusion.  So instead of: bindings.ListImages(), we now do a [bindings].images.List().

Also, the connection is passed to each binding method via a context to allow for future growth.

Lastly, add first set of tests.  There are a couple of things to work out for rootless tests yet.


baude:
please note that although tests are now being added, i have to not added them to CI yet.  I wish to get this PR in first; then sort out the rootless issues.  The turn on CI for them.  Wont be long but this PR is big enough.

rhatdan:
LGTM, except one include problem,

pkg/bindings/containers/containers.go
```diff
@@ -0,0 +1,259 @@
...
+
+	"github.com/containers/libpod/cmd/podman/shared"
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/pkg/bindings"
+)
+
+// List obtains a list of containers in local storage.  All parameters to this method are optional.
+// The filters are used to determine which containers are listed. The last parameter indicates to only return
+// the most recent number of containers.  The pod and size booleans indicate that pod information and rootfs
+// size information should also be included.  Any finally, the sync bool synchronizes the OCI runtime and
```
TomSweeneyRedHat: nit "Any finally" -> "Finally"

baude:
fixed imports

pkg/bindings/connection.go
```diff
@@ -60,3 +125,24 @@ func (c Connection) newRequest(httpMethod, endpoint string, httpBody io.Reader,
 	response, err := c.client.Do(req) // nolint
```
jwhonce: We should re-initiate the server connection if dropped.  At least once, maybe three times...

pkg/bindings/images/images.go
```diff
@@ -0,0 +1,189 @@
...
+	return deleted, response.Process(nil)
+}
+
+// Tag adds an additional name to locally-stored image.
+func Tag(ctx context.Context, nameOrID string) error {
+	conn, err := bindings.GetConnectionFromContext(ctx)
+	if err != nil {
+		return err
+	}
+	var ()
```
jwhonce: No variables declared.  No params sent.

pkg/bindings/images/images.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	response, err := conn.NewRequest(http.MethodPost, "/images/prune", nil, params)
+	if err != nil {
+		return deleted, err
+	}
+	return deleted, response.Process(nil)
+}
+
+// Tag adds an additional name to locally-stored image.
+func Tag(ctx context.Context, nameOrID string) error {
```
jwhonce: Don't you need the tag value here?  Or maybe an array of tags?

pkg/bindings/pods/pods.go
```diff
@@ -0,0 +1,197 @@
...
+func Exists(ctx context.Context, nameOrID string) (bool, error) {
+	conn, err := bindings.GetConnectionFromContext(ctx)
+	if err != nil {
+		return false, err
+	}
+	response, err := conn.NewRequest(http.MethodGet, fmt.Sprintf("/pods/%s/exists", nameOrID), nil, nil)
+	if err != nil {
+		return false, err
+	}
+	return response.StatusCode == http.StatusOK, err
```
jwhonce: See note above on checking status codes.

pkg/bindings/test/images_test.go
```diff
@@ -0,0 +1,92 @@
+package test_bindings
```
jwhonce: Without populating return values from the image functions these cannot be used.

pkg/bindings/pods/pods.go
```diff
@@ -0,0 +1,197 @@
...
+	response, err := conn.NewRequest(http.MethodPost, "/pods/prune", nil, nil)
+	if err != nil {
+		return err
+	}
+	return response.Process(nil)
+}
+
+// List returns all pods in local storage.  The optional filters parameter can
+// be used to refine which pods should be listed.
+func List(ctx context.Context, filters map[string][]string) (*[]libpod.PodInspect, error) {
```
jwhonce: Shouldn't this be []*libpod.PodInspect ?

pkg/bindings/containers/containers.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/pkg/bindings"
+)
+
+// List obtains a list of containers in local storage.  All parameters to this method are optional.
+// The filters are used to determine which containers are listed. The last parameter indicates to only return
+// the most recent number of containers.  The pod and size booleans indicate that pod information and rootfs
+// size information should also be included.  Finally, the sync bool synchronizes the OCI runtime and
+// container state.
+func List(ctx context.Context, filters map[string][]string, last *int, pod, size, sync *bool) ([]shared.PsContainerOutput, error) { // nolint:typecheck
```
jwhonce: This should be []*shared.PsContainerOutput to be consistent with other API calls.

pkg/bindings/images/images.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if response.StatusCode == http.StatusOK {
+		return true, nil
+	}
+	return false, nil
+}
+
+// List returns a list of images in local storage.  The all boolean and filters parameters are optional
+// ways to alter the image query.
+func List(ctx context.Context, all *bool, filters map[string][]string) ([]handlers.ImageSummary, error) {
```
jwhonce: This should be []*handlers.ImageSummary to be consistent with other API endpoints

pkg/bindings/images/images.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return &inspectedData, response.Process(&inspectedData)
+}
+
+func ImageTree(ctx context.Context, nameOrId string) error {
+	return bindings.ErrNotImplemented
+}
+
+// History returns the parent layers of an image.
+func History(ctx context.Context, nameOrID string) ([]handlers.HistoryResponse, error) {
```
jwhonce: []*handlers.HistoryResponse

pkg/bindings/images/images.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	// TODO this still needs error handling added
+	//_, err := http.Post(c.makeEndpoint("/images/loads"), "application/json", r) //nolint
+	_ = conn
+	return bindings.ErrNotImplemented
+}
+
+// Remove deletes an image from local storage.  The optional force parameter will forcibly remove
+// the image by removing all all containers, including those that are Running, first.
+func Remove(ctx context.Context, nameOrID string, force *bool) ([]map[string]string, error) {
```
jwhonce: Given this is one image, an array of maps seems wrong.

pkg/bindings/images/images.go
```diff
@@ -0,0 +1,189 @@
...
+		params["compress"] = strconv.FormatBool(*compress)
+	}
+	response, err := conn.NewRequest(http.MethodGet, fmt.Sprintf("/images/%s/get", nameOrID), nil, params)
+	if err != nil {
+		return err
+	}
+	if err := response.Process(nil); err != nil {
+		return err
+	}
+	_, err = io.Copy(w, response.Body)
```
jwhonce: I believe you need a reponse.Body.Close() after the copy.

pkg/bindings/network/network.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/containernetworking/cni/libcni"
+	"github.com/containers/libpod/pkg/bindings"
+)
+
+func Create() {}
+func Inspect(ctx context.Context, nameOrID string) (map[string]interface{}, error) {
```
jwhonce: Why use interface?  Everything has been string-ified to get here.  So you would need to convert them to xxx for this type of return.

pkg/bindings/connection.go
```diff
@@ -17,33 +25,90 @@ type APIResponse struct {
...
 
 func (c Connection) makeEndpoint(u string) string {
-	return fmt.Sprintf("%s%s", defaultConnection, u)
+	// The d character in the url is discarded and is meaningless
+	return fmt.Sprintf("http://d/%s%s", defaultConnectionPath, u)
 }
 
-func (c Connection) newRequest(httpMethod, endpoint string, httpBody io.Reader, params map[string]string) (*APIResponse, error) {
+// NewRequest assembles the http request and returns the response
+func (c Connection) NewRequest(httpMethod, endpoint string, httpBody io.Reader, params map[string]string) (*APIResponse, error) {
```
jwhonce: This should be named DoRequest() or something like that.  It doesn't return a new API Request object().

If this is changed to a variadic function we could hide more boilerplate:

conn.NewRequest(nil, params, "GET", "/images/%s/json", nameOrID)

NewRequest(r io.Reader, p map[string]string, m string, e string, a ...interface{})
  s := fmt.Sprintf(e, ...a)
  // Use s
  .

pkg/bindings/images/images.go
baude: discussed and resolved

pkg/bindings/network/network.go
baude: ah yes, the joys of working with cni ... they are interfaces ...

pkg/bindings/images/images.go
baude: discussed, keep for now

baude:
fixed up with extra sugar @jwhonce

pkg/bindings/connection.go
```diff
@@ -17,46 +25,170 @@ type APIResponse struct {
...
+	var (
+		err            error
+		response       *http.Response
+		safePathValues []string
+	)
+	// Make sure path values are http url safe
+	for _, pv := range pathValues {
+		safePathValues = append(safePathValues, url.QueryEscape(pv))
+	}
+	safeEndpoint := fmt.Sprintf(endpoint, safePathValues)
```
jwhonce: Shouldn't safePathValues be safePathValues[:] ?

baude: @jwhonce the linter puked on that.

PR #4981, opened 2020-01-26
Automatically execute podman-remote if vars set

### LINKED ISSUE #4390: Automatically fork/exec to "podman-remote" if PODMAN_VARLINK_ADDRESS is set / implement podman -H

/kind feature

**Description**

Today it seems you have to choose if you run a container locally (by running "podman run") or if you want to run a container remotely (by running "podman-remote run"). It means you don't write the exact same command depending on the mode you want to use. Docker on the other side has a single executable "docker", which queries the environment variable DOCKER_HOST to know to which deamon it shall connect to (and defaults to unix://var/run/docker.sock), whether it's a local deamon or a remote one.

Would it be possible that you:
 - Implement an implicit "fork/exec" to "podman-remote" whenever the "podman" executable is invoked, but PODMAN_VARLINK_ADDRESS environment variable is actually set, allowing to transparently use remote container just by a change of an environment variable, and without any change of which executable is launched (ie do not force users to explicitly write "podman-remote" ?
 - Implement some -H flag, eg "podman -H <podman-valink-address> run ..." similar to "docker -H <docker-host> run ...", more or less the same way than in point 1 if PODMAN_VARLINK_ADDRESS had been set ?

The use case would typically fixing the classical issues of Docker in Docker (or Podman in Podman) which is either broken or hard to set up correctly without being too limited. This case happens for us in our Jenkins Pipeline flows. The Jenkins slave itself is a Docker container, and as part of the Jenkinsfile it's not uncommon to run other Docker containers inside it. Using Docker in Docker would have many limitations, so today Jenkins relies on Docker side by side Docker, by bind mounting /var/run/docker.sock inside the Jenkins slave container. As one may want the exact same code to run the same way when run through Jenkins (so from a container), and outside Jenkins (so without top level container), it would be cool if "podman" would automatically translate to "podman-remote" just based on some environment variable, just like "docker" does.

afbjorklund:
If any of PODMAN_HOST or PODMAN_VARLINK_BRIDGE are set,
then execute "podman-remote" rather than this "podman".

Example:

```console
$ ./bin/podman version
Version:            1.7.1-dev
RemoteAPI Version:  1
Go Version:         go1.13.6
Git Commit:         263432b1db60013d5f871f371112f75e2024c88e
Built:              Sun Jan 26 18:43:10 2020
$ eval $(podman-machine env --varlink)
$ ./bin/podman version
Client:
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.13.6
Git Commit:         b7ce1157b00af09f4a09e39b377aa3abff46ee05
Built:              Sat Jan 18 15:26:34 2020

Server:
Version:            1.7.1-dev
RemoteAPI Version:  1
Go Version:         go1.12.10
Git Commit:         ce4bf33ccc80819f199bdbb6526b9adee1ad41b0-dirty
Built:              Sat Jan 18 11:01:17 2020
```

Closes #4390

### DIFF
--- a/cmd/podman/main_local.go
+++ b/cmd/podman/main_local.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"log/syslog"
 	"os"
+	"os/exec"
 	"runtime/pprof"
 	"strconv"
 	"strings"
@@ -31,7 +32,37 @@ import (
 
 const remote = false
 
+func runRemote(path string, args ...string) (int, error) {
+	c := exec.Command(path, args...)
+	c.Stdin = os.Stdin
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	if err := c.Run(); err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			waitStatus := exitError.Sys().(syscall.WaitStatus)
+			return waitStatus.ExitStatus(), nil
+		}
+		return 1, err
+	}
+	return 0, nil
+}
+
 func init() {
+	host := os.Getenv("PODMAN_HOST")
+	bridge := os.Getenv("PODMAN_VARLINK_BRIDGE")
+	address := os.Getenv("PODMAN_VARLINK_ADDRESS")
+	if host != "" || bridge != "" || address != "" {
+		program := "podman-remote"
+		if path, err := exec.LookPath(program); err == nil {
+			if rc, err := runRemote(path, os.Args[1:]...); err == nil {
+				os.Exit(rc)
+			} else {
+				fmt.Fprintf(os.Stderr, "Error running %s: %v\n", path, err)
+			}
+		} else {
+			fmt.Fprintf(os.Stderr, "Error finding %s: %v\n", program, err)
+		}
+	}
 	cgroupManager := define.SystemdCgroupsManager
 	cgroupHelp := `Cgroup manager to use ("cgroupfs"|"systemd")`
 	cgroupv2, _ := cgroups.IsCgroup2UnifiedMode()


afbjorklund:
Hmm, think I messed up the error reporting when I made it into a function:

``` console
$ ./bin/podman run busybox true
$ ./bin/podman run busybox false
Error running /usr/local/bin/podman-remote: exit status 1
```

afbjorklund:
Forgot PODMAN_VARLINK_ADDRESS, which was used in the documentation...

cmd/podman/main_local.go
```diff
@@ -31,7 +32,37 @@ import (
...
+	}
+	return 0, nil
+}
+
 func init() {
+	host := os.Getenv("PODMAN_HOST")
+	bridge := os.Getenv("PODMAN_VARLINK_BRIDGE")
+	address := os.Getenv("PODMAN_VARLINK_ADDRESS")
+	if host != "" || bridge != "" || address != "" {
+		program := "podman-remote"
```
TomSweeneyRedHat: It might be good to have a debug statement in here to just say "Attempting to connect to remote Podman" or some such.  I could see it as being potentially helpful in a debugging situation later on.

afbjorklund: So something like a `logrus.Debugf` with the command being executed perhaps.

afbjorklund: Tried adding logging, but it doesn't show up in the logs (at the debug level)

afbjorklund: @TomSweeneyRedHat : apparently this (`init`) happens _before_ logging has been set up

That is, we are still running with the default log level of InfoLevel (hiding the debug logging)

vrothberg:
Friendly ping.

afbjorklund:
Not sure if there’s anything more that can be done here, or if we should continue to rename the binary...

afbjorklund:
We are currently missing newer `podman-remote` binaries as well, so there's nothing to execute.

rhatdan:
@afbjorklund Still working on this?

afbjorklund:
@rhatdan : No, I am not sure what kind of logging that TomSweeneyRedHat wanted ?

If we output something, that will break users (the same way that the docker wrapper does)

And the logging isn't set up yet, since we call this in the `init` before we set up logging

I think we just went with everybody using `podman-remote` for the binary name instead.

vrothberg:
I am going through open pull requests at the moment. Apologies for not getting this in earlier!

Much has changed since this PR was opened and while it doesn't apply as is anymore, we can still use the idea. The new remote client is talking to Podman's new rest API, so we can ignore varlink. But auto-detecting `PODMAN_HOST` and using the remote backend could easily be done by adding a host-env check to https://github.com/containers/libpod/blob/master/cmd/podman/registry/config.go#L40.

@baude @jwhonce WDYT?

afbjorklund:
> Much has changed since this PR was opened and while it doesn't apply as is anymore, we can still use the idea. The new remote client is talking to Podman's new rest API, so we can ignore varlink.

We have already adopted to the existing policy of `podman` being native on Linux and remote on Mac and Windows ("not available" on Linux), and using varlink and sudo for the access model.

afbjorklund:
I will close this PR, and you can make a new one for the new cilent (for the same original issue).

We still have some issues with the documentation being "blank" and the > 1.6 binaries missing ?

So it seems to be a _lot_ more than just the name of the program that will need some rework here...

Can continue the "podman-remote" discussion in the #4390 issue instead of here, if needed.

Issue #4706, opened 2019-12-16
podman-remote: podman push - invalid memory address or nil pointer dereference


bszeti:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Using podman-remote on MacOS with a remote Fedora 31. Command 'podman push' drops a go error:

```
$ podman push 56d835d0d82f docker://quay.io/bszeti/podmantest:latest
panic: runtime error: invalid memory address or nil pointer dereference
```
**Steps to reproduce the issue:**

1. Pull or build and image (in my case image Id: 56d835d0d82f)
2. Try to push the image to a remote registry using podman-remote
podman push 56d835d0d82f docker://quay.io/bszeti/podmantest:latest

**Describe the results you received:**
Command drops error.
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x1ec883c]

goroutine 1 [running]:
main.pushCmd(0x33552e0, 0x0, 0x0)
	cmd/podman/push.go:97 +0x29c
main.glob..func57(0x332fb40, 0xc000393b80, 0x2, 0x2, 0x0, 0x0)
	cmd/podman/push.go:36 +0x94
github.com/spf13/cobra.(*Command).execute(0x332fb40, 0xc0000bc020, 0x2, 0x2, 0x332fb40, 0xc0000bc020)
	vendor/github.com/spf13/cobra/command.go:826 +0x465
github.com/spf13/cobra.(*Command).ExecuteC(0x3331440, 0x6, 0xc000044047, 0x31)
	vendor/github.com/spf13/cobra/command.go:914 +0x2fc
github.com/spf13/cobra.(*Command).Execute(...)
	vendor/github.com/spf13/cobra/command.go:864
main.main()
	cmd/podman/main.go:160 +0xb0
```

**Describe the results you expected:**
Successful push.

**Additional information you deem important (e.g. issue happens only occasionally):**
Same command works without issue if it's executed on the Linux host.

**Output of `podman version`:**

```
Client:
Version:            1.6.1-dev
RemoteAPI Version:  1
Go Version:         go1.12.9
OS/Arch:            darwin/amd64

Service:
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
client:
[...]

PR #4985, opened 2020-01-26
Hidden remote flags can be nil

baude:
The pull command has several options that are hidden for the remote client.  In that case, when checking to see if the flag has been flipped with .Changed, we get a nil pointer error.  Using IsSet is tolerant of this.

Fixes: #4706


rhatdan:
LGTM

mheon:
LGTM

Issue #4799, opened 2020-01-06
Healthcheck not tracked on images


astronouth7303:
/kind feature

**Description**

Looking at `podman image inspect`, the varlink IDL, and `python-podman`, I can see that containers have a healthcheck, but it is not tracked on the image.

Note that I have been consulting https://github.com/containers/libpod/blob/master/cmd/podman/varlink/io.podman.varlink this entire time, so I'm reasonably confident this issue exists in `master`.

**Steps to reproduce the issue:**

1. Create an image with a healthcheck (such as from a `Dockerfile` with a `HEALTHCHECK` line) with podman
2. Inspect it with podman

**Describe the results you received:**
No mention of a healthcheck anywhere in the image metadata

**Describe the results you expected:**
A mention in the healthcheck (probably in the `Config` section)

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:            1.6.4
RemoteAPI Version:  1
Go Version:         go1.13.5
Git Commit:         5cc92849f7fc9dd734ca2fd8f3ae8830b9a7eb26
Built:              Sat Dec 21 13:13:24 2019
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: 5cc92849f7fc9dd734ca2fd8f3ae8830b9a7eb26
  go version: go1.13.5
  podman version: 1.6.4
host:
  BuildahVersion: 1.12.0-dev
  CgroupVersion: v1
  Conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.2, commit: unknown'
  Distribution:
    distribution: debian
    version: unknown
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 842928128
  MemTotal: 8229838848
  OCIRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: |-
      runc version 1.0.0~rc9+dfsg1
      commit: 1.0.0~rc9+dfsg1-1
      spec: 1.0.1-dev
  SwapFree: 6711959552
  SwapTotal: 8467247104
[...]

PR #4986, opened 2020-01-26
inspect image healthchecks

baude:
when a docker image has a defined healthcheck, it should be displayed with inspect.  this is only valid for docker images as oci images are not aware of healthchecks.

Fixes: #4799

### DIFF
--- a/pkg/inspect/inspect.go
+++ b/pkg/inspect/inspect.go
@@ -3,6 +3,7 @@ package inspect
 import (
 	"time"
 
+	"github.com/containers/image/v5/manifest"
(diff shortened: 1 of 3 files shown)


pkg/inspect/inspect.go
```diff
@@ -1,6 +1,7 @@
 package inspect
 
 import (
+	"github.com/containers/image/v5/manifest"
```
rhatdan: Wrong place

rhatdan:
LGTM once you fix the placement of the include directive.

baude:
fixed

PR #4987, opened 2020-01-27
vendor in c/common config pkg for containers.conf

rhatdan:
vendor in latest buildah


rhatdan:
Attempting to simplify the containers.conf patch set to see if we can pass the tests.  Concentrating only on the client side.

rhatdan:
Going back to original PR>

Issue #4880, opened 2020-01-16
podman play kube does not pull images like it should


Komic:
/kind bug

`podman play kube` does not seem to be pulling updated images as directed by the `:latest` tag or `imagePullPolicy: always`.

**Steps to reproduce the issue:**

1. `podman stop/remove` an outdated container
2. `podman play kube` a .yml containing the `:latest` tag or `imagePullPolicy: always`
3. The container is recreated from the outdated image

**Output of `podman version`:**

```
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.13.5
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.5
  podman version: 1.7.0
host:
  BuildahVersion: 1.12.0
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.2-1.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.2, commit: 186a550ba0866ce799d74006dab97969a2107979'
  Distribution:
    distribution: fedora
    version: "31"
  OCIRuntime:
    name: crun
    package: crun-0.10.6-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.10.6
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  arch: amd64
  cpus: 12
  eventlogger: journald
  kernel: 5.3.16-300.fc31.x86_64
  os: linux
  rootless: false
registries:
  search:
  - docker.io
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - quay.io
store:
  ConfigFile: /etc/containers/storage.conf
  ContainerStore:
    number: 26
  GraphDriverName: overlay
  GraphOptions:
    overlay.mountopt: nodev,metacopy=on
  GraphRoot: /var/lib/containers/storage
  GraphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "true"
  ImageStore:
    number: 12
  RunRoot: /var/run/containers/storage
  VolumePath: /var/lib/containers/storage/volumes
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-1.7.0-2.fc31.x86_64
```

rhatdan:
[...]

PR #4993, opened 2020-01-27
honor pull policy in play kube

baude:
When a container specification has a pull policy, we should honor it when recreating the pods/containers from yaml.

Fixes: #4880

### DIFF
--- a/pkg/adapter/pods.go
+++ b/pkg/adapter/pods.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/containers/buildah/pkg/parse"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/libpod/cmd/podman/cliconfig"
 	"github.com/containers/libpod/cmd/podman/shared"
@@ -604,7 +605,24 @@ func (r *LocalRuntime) PlayKubeYAML(ctx context.Context, c *cliconfig.KubePlayVa
 	}
 
 	for _, container := range podYAML.Spec.Containers {
-		newImage, err := r.ImageRuntime().New(ctx, container.Image, c.SignaturePolicy, c.Authfile, writer, &dockerRegistryOptions, image.SigningOptions{}, nil, util.PullImageMissing)
+		pullPolicy := util.PullImageMissing
+		if len(container.ImagePullPolicy) > 0 {
+			pullPolicy, err = util.ValidatePullType(string(container.ImagePullPolicy))
+			if err != nil {
+				return nil, err
+			}
+		}
+		named, err := reference.ParseNormalizedNamed(container.Image)
+		if err != nil {
+			return nil, err
+		}
+		// In kube, if the image is tagged with latest, it should always pull
+		if tagged, isTagged := named.(reference.NamedTagged); isTagged {
+			if tagged.Tag() == image.LatestTag {
+				pullPolicy = util.PullImageAlways
+			}
+		}
+		newImage, err := r.ImageRuntime().New(ctx, container.Image, c.SignaturePolicy, c.Authfile, writer, &dockerRegistryOptions, image.SigningOptions{}, nil, pullPolicy)
 		if err != nil {
 			return nil, err
 		}
--- a/test/e2e/play_kube_test.go
+++ b/test/e2e/play_kube_test.go
@@ -47,6 +47,7 @@ spec:
       value: podman
     image: {{ .Image }}
     name: {{ .Name }}
+    imagePullPolicy: {{ .PullPolicy }}
     resources: {}
     {{ if .SecurityContext }}
     securityContext:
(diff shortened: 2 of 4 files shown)


jwhonce:
LGTM

pkg/adapter/pods.go
```diff
@@ -604,7 +604,14 @@ func (r *LocalRuntime) PlayKubeYAML(ctx context.Context, c *cliconfig.KubePlayVa
 	}
 
 	for _, container := range podYAML.Spec.Containers {
-		newImage, err := r.ImageRuntime().New(ctx, container.Image, c.SignaturePolicy, c.Authfile, writer, &dockerRegistryOptions, image.SigningOptions{}, nil, util.PullImageMissing)
+		pullPolicy := util.PullImageMissing
```
TomSweeneyRedHat: not used?

pkg/adapter/pods.go
```diff
@@ -604,7 +604,14 @@ func (r *LocalRuntime) PlayKubeYAML(ctx context.Context, c *cliconfig.KubePlayVa
...
+		if len(container.ImagePullPolicy) > 0 {
+			pullPolicy, err = util.ValidatePullType(string(container.ImagePullPolicy))
+			if err != nil {
+				return nil, err
+			}
+		}
+		newImage, err := r.ImageRuntime().New(ctx, container.Image, c.SignaturePolicy, c.Authfile, writer, &dockerRegistryOptions, image.SigningOptions{}, nil, pullPolicy)
```
TomSweeneyRedHat: if there's not a container.ImagePullPolicy, should you set it to the `pullPolicy` that you established prior?  It doesn't look like pullPolicy is being used for anything but what's returned to it here in the if statement.  I'm surprised lint didn't bark, or maybe I need more tea.

test/e2e/play_kube_test.go
```diff
@@ -396,4 +404,14 @@ var _ = Describe("Podman generate kube", func() {
 		Expect(logs.ExitCode()).To(Equal(0))
 		Expect(logs.OutputToString()).To(ContainSubstring("Operation not permitted"))
 	})
+
+	It("podman play kube with pull policy of never should be 125", func() {
```
haircommander: this is a great start. would you mind adding a couple more tests of the other options, so if they change behavior we catch it here too?

pkg/adapter/pods.go
haircommander: it seems to be overwritten below if ImagePullPolicy > 0

haircommander:
LGTM besides possibly more test coverage :smile:

haircommander:
also given this comment https://github.com/containers/libpod/issues/4880#issuecomment-578947664 I'd either like special case handling for `latest` or a new issue to track that we don't currently handle latest similarly

baude:
> also given this comment [#4880 (comment)](https://github.com/containers/libpod/issues/4880#issuecomment-578947664) I'd either like special case handling for `latest` or a new issue to track that we don't currently handle latest similarly

did this per our discussion too!

pkg/adapter/pods.go
```diff
@@ -604,7 +605,24 @@ func (r *LocalRuntime) PlayKubeYAML(ctx context.Context, c *cliconfig.KubePlayVa
...
+				return nil, err
+			}
+		}
+		named, err := reference.ParseNormalizedNamed(container.Image)
+		if err != nil {
+			return nil, err
+		}
+		// In kube, if the image is tagged with latest, it should always pull
+		if tagged, isTagged := named.(reference.NamedTagged); isTagged {
+			if tagged.Tag() == "latest" {
```
haircommander: It makes me sad we don't have a public variable DefaultTag or anything to use here. the best I can find is libpod/image/parts.go:70

my saddness isn't really part of this PR, but if you happened to use a variable here it'd make me not sad anymore :slightly_smiling_face:

test/e2e/play_kube_test.go
```diff
@@ -396,4 +404,86 @@ var _ = Describe("Podman generate kube", func() {
...
+		err := generateKubeYaml(getPod(withCtr(ctr)), kubeYaml)
+		Expect(err).To(BeNil())
+
+		kube := podmanTest.Podman([]string{"play", "kube", kubeYaml})
+		kube.WaitWithDefaultTimeout()
+		Expect(kube.ExitCode()).To(Equal(125))
+	})
+
+	It("podman play kube with pull policy of missing", func() {
+		ctr := getCtr(withPullPolicy("never"), withImage(BB))
```
haircommander: test name and policy name don't seem to match

test/e2e/play_kube_test.go
```diff
@@ -396,4 +404,86 @@ var _ = Describe("Podman generate kube", func() {
...
+		Expect(err).To(BeNil())
+
+		kube := podmanTest.Podman([]string{"play", "kube", kubeYaml})
+		kube.WaitWithDefaultTimeout()
+		Expect(kube.ExitCode()).To(Equal(0))
+
+		inspect = podmanTest.Podman([]string{"inspect", BB})
+		inspect.WaitWithDefaultTimeout()
+		newBBinspect := inspect.InspectImageJSON()
+		Expect(oldBBinspect[0].Digest).To(Not(Equal(newBBinspect[0].Digest)))
```
haircommander: nice, I like this approach

haircommander:
one blocking nit, one non-blocking nit, otherwise LGTM :clinking_glasses:

pkg/adapter/pods.go
baude: fixed

haircommander:
comments addressed in record time, LGTM

Issue #4473, opened 2019-11-07
Wrong network namespace when exec in container created with --net=host


faulesocke:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
When a container is created with `podman run --net=host` all instances of `podman exec` to that container do not attach to the containers network namespace but use whatever network namespace they are executed in.

**Steps to reproduce the issue:**
```
ip netns add foo
ip netns exec foo podman run --net=host -d <image>
podman exec <cid> ip link
```

**Describe the results you received:**
I receive a list of the network devices in my current network namespace.

**Describe the results you expected:**
I expect to see the network devices currently present in the containers network namespace.

**Additional information you deem important (e.g. issue happens only occasionally):**
I actually use `--net=host` to manually create a network namespace as root and then I use `ip netns exec su --login -c "podman run --net=host ..." <user>` to drop privileges and run a rootless container inside that namespace.

This allows me to put whatever I like into that namespace prior to running any container app. This is especially useful to put a macvlan or veth interface into a rootless container.

**Output of `podman version`:**

```
% podman version
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.3
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
probably not necessary
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
% xbps-query podman
architecture: x86_64
build-date: 2019-10-28 09:08 UTC
filename-sha256: 9f02213136c2e47678d2a7df463ff39b38ef0bde9e2f423e06f8d8a25ad31f0b
filename-size: 17MB
homepage: https://podman.io/
install-date: 2019-11-02 14:30 UTC
installed_size: 52MB
license: Apache-2.0
maintainer: Cameron Nemo <camerontnorman@gmail.com>
metafile-sha256: 217e71f163e554521506eee062869443aa3467d1b3a4d1a3ded0f738b0ddbc4b
pkgver: podman-1.6.2_1
repository: https://alpha.de.repo.voidlinux.org/current
shlib-requires:
	libpthread.so.0
	libgpgme.so.11
	libassuan.so.0
	libgpg-error.so.0
	libseccomp.so.2
	librt.so.1
	libdevmapper.so.1.02
	libc.so.6
short_desc: Simple management tool for containers and images
source-revisions: podman:1c4242350be
state: installed
```

[...]

PR #4996, opened 2020-01-27
[CI:DOCS]rootless exec cannot join root namespace

baude:
add a quick note in the rootless.md about usage of --net=host as rootless and subsequent podman execs.

Fixes: #4473


TomSweeneyRedHat:
LGTM
/hold cancel

PR #4997, opened 2020-01-27
Update remote client bridge documentation.

jwhonce:
* Add PODMAN_VARLINK_BRIDGE examples.

### DIFF
--- a/docs/tutorials/remote_client.md
+++ b/docs/tutorials/remote_client.md
@@ -18,7 +18,7 @@ installed on it and the varlink service activated.  You will also need to be abl
 system as a user with privileges to the varlink socket (more on this later).
 
 ## Building the remote client
-At this time, the remote-client is not being packaged for any distribution.  It must be built from
+At this time, the Podman remote-client is not being packaged for any distribution.  It must be built from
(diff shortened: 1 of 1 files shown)


docs/tutorials/remote_client.md
```diff
@@ -54,24 +61,28 @@ access to the remote system.  This limitation is being worked on.
...
-```
-Note here we are binding the Podman socket to a local TCP socket on port 1234.
-
-#### Running the remote client
-With the ssh session established, we can now run the remote client in a different terminal window. You
-must inform Podman where to look for the bound socket you created in the previous step using an
-environment variable.
+To use the remote client, we must establish an ssh connection to the Podman server.
+Using the varlink bridge, we're going to initiate a ssh tunnel to connect to the server. You
+must inform Podman where to look for the sshd service on which server.
```
TomSweeneyRedHat: I'm not a fan of the words 'we' or 'you' in a man page.  Suggest:

To use the remote client, an ssh connection to the Podman server must be established.
Using the varlink bridge, an ssh tunnel must be initiated to connect to the server.  Podman must
then be informed of the location of the sshd server on the targeted server.

Issue #2909, opened 2019-04-12
better error message required in case error in CNI plugin.


kunalkushwaha:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**
/kind bug

**Description**

In case of any error in CNI config file, the error message while creating container with network of CNI plugin simply show error `CNI network "<network-name>" not found`. This error message is confusing for end user. 

A better error message will be helpful to understand where actually error is.

_Raised by podman end user:_

**Steps to reproduce the issue:**

1. Create CNI plugin with some error.
```bash
$ cat /etc/cni/net.d/77-ipvlan.conflist 
{
    "cniVersion": "0.3.0",
    "name": "myvlan",
    "plugins": [
      {
        "type": "ipvlan",
        "master": "enp1s0", #some comments
        "ipam": {
            "type": "host-local",
            "subnet": "10.88.0.0/16",
            "routes": [
                { "dst": "0.0.0.0/0" }
            ]
        }
     }
    ]
}

```
2. create container with `myvlan` network.
```bash
$ sudo podman  run --rm -it --network=myvlan docker.io/library/alpine sh
ERRO[0000] CNI network "myvlan" not found               
Error: error configuring network namespace for container baf2fda585db0cc2c874a4bfc00bca3dd27e8402a18ec8e157124240ca887f29: CNI network "myvlan" not found
```

3. To understand complete details of error, users has to run `podman create` with `--log-level-debug` option. 
```log
DEBU[0000] overlay test mount indicated that metacopy is not being used
DEBU[0000] backingFs=extfs, projectQuotaSupported=false, useNativeDiff=true, usingMetacopy=false
WARN[0000] Error loading CNI config list file /etc/cni/net.d/77-ipvlan.conflist: error parsing configuration list: invalid character '#' looking for beginning of object key string
INFO[0000] Found CNI network podman (type=bridge) at /etc/cni/net.d/87-podman-bridge.conflist
WARN[0000] Error loading CNI config file /etc/cni/net.d/97-podman-macvlan.conf: error parsing configuration: missing 'type'
INFO[0000] Found CNI network lo (type=loopback) at /etc/cni/net.d/99-loopback.conf

```

**Describe the results you received:**
```
ERRO[0000] CNI network "myvlan" not found               
Error: error configuring network namespace for container baf2fda585db0cc2c874a4bfc00bca3dd27e8402a18ec8e157124240ca887f29: CNI network "myvlan" not found
```

[...]

PR #5008, opened 2020-01-28
display file name of bad cni conf

baude:
if one of the cni conf files is badly formatted or cannot be loaded, we now display the error as well as the filename.

Fixes: #2909


TomSweeneyRedHat:
LGTM assuming happy tests

Issue #4814, opened 2020-01-08
`make lint` fails on macos


ssbarnea:
```
$ make lint                                                                                                                                                                                                    Makefile:47: Podman is being compiled without the systemd build tag. Install libsystemd on Ubuntu or systemd-devel on rpm based distro for journald support
X11 forwarding request failed
X11 forwarding request failed
GO111MODULE=off go generate ./cmd/podman/varlink/...
../../../vendor/github.com/varlink/go/cmd/varlink-go-interface-generator/main.go:12:2: cannot find package "github.com/varlink/go/varlink/idl" in any of:
/usr/local/Cellar/go/1.13.4/libexec/src/github.com/varlink/go/varlink/idl (from $GOROOT)
/Users/ssbarnea/go/src/github.com/varlink/go/varlink/idl (from $GOPATH)
cmd/podman/varlink/generate.go:3: running "go": exit status 1
make: *** [cmd/podman/varlink/iopodman.go] Error 1
FAIL: 2
```

rhatdan:
@baude PTAL

baude:
We dont support native compilation on osx ... only cross-compile from linux.  Varlink is one reason.  @ssbarnea how do you want to deal with this/go from here.

ssbarnea:
Linting must work from any platform likely to support contributions from, even if compilation would be cross-compiled.

I was planning to do something about it myself but got busy with other stuff. Keep it open until someone fixes it.

I would consider acceptable even if only pre-commit runs on macos and the rest is skipped. The more interesting aspect is that pre-commit works but make never reaches it because chokes at the dependency target it cannot build.

If any idea about how to make that target optional for macos it would be great. 

PS. In fact make could start some containers and do the build with them for unsupported platforms, but this would require lots of changes.

If you tell me which patch to take, I can look into making a PR to address it.

medyagh:
@vrothberg this issue seem to still exist !
on mac : 

```
==> make podman-remote
Last 15 lines from /Users/medya/Library/Logs/Homebrew/podman/01.make:
make
podman-remote
uname: illegal option -- o
usage: uname [-amnprsv]
mkdir -p "/private/tmp/podman-20200210-97980-183905n/_output/src/github.com/containers"
ln -sfn "/private/tmp/podman-20200210-97980-183905n" "/private/tmp/podman-20200210-97980-183905n/_output/src/github.com/containers/libpod"
[...]

PR #5011, opened 2020-01-29
run `varlink_generate` on Linux only

vrothberg:
Running the `varlink_generate` make target on non-Linux machines is not
supported, so restrict it to Linux only.

Fixes: #4814

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -524,7 +524,10 @@ install.libseccomp.sudo:
 
 
 cmd/podman/varlink/iopodman.go: .gopathok cmd/podman/varlink/io.podman.varlink
+ifeq ("$(shell uname -o)", "GNU/Linux")
+	# Only generate the varlink code on Linux (see issue #4814).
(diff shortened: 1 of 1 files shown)


vrothberg:
@baude @ssbarnea PTAL.

Note that I do not have any non-Linux machines. It would be good to have someone check if this PR fixes #4814 or if we're stuck on a subsequent stage.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

afbjorklund:
@vrothberg: this didn't seem to work everywhere

`uname: illegal option -- o`

Would it be possible to use the standard `uname -s` ?

medyagh:
I sitll have this issue my mac version is 10.14.6

vrothberg:
-> https://github.com/containers/libpod/pull/5162

Thanks for the ping!

PR #5015, opened 2020-01-29
apiv2 binding test fixes

baude:
a recent refactor in the bindings broke the tests.  quick fixes to get them working again.

### DIFF
--- a/pkg/bindings/connection.go
+++ b/pkg/bindings/connection.go
@@ -115,11 +115,12 @@ func (c *Connection) DoRequest(httpBody io.Reader, httpMethod, endpoint string, 
 	)
 	safePathValues := make([]interface{}, len(pathValues))
 	// Make sure path values are http url safe
-	for _, pv := range pathValues {
-		safePathValues = append(safePathValues, url.QueryEscape(pv))
+	for i, pv := range pathValues {
+		safePathValues[i] = url.QueryEscape(pv)
 	}
+	// Lets eventually use URL for this which might lead to safer
+	// usage
 	safeEndpoint := fmt.Sprintf(endpoint, safePathValues...)
-
 	e := c.makeEndpoint(safeEndpoint)
 	req, err := http.NewRequest(httpMethod, e, httpBody)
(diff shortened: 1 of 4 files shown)


pkg/bindings/connection.go
```diff
@@ -113,13 +113,12 @@ func (c *Connection) DoRequest(httpBody io.Reader, httpMethod, endpoint string,
 		err      error
 		response *http.Response
 	)
-	safePathValues := make([]interface{}, len(pathValues))
+	var safePathValues []interface{}
```
jwhonce: If you keep make(), then use for loop iteration value, eg:

safePathValue[i] = url.QueryEscape(pv)

Then you don't need to append.  Shaves some time off.

pkg/bindings/connection.go
```diff
@@ -113,13 +113,12 @@ func (c *Connection) DoRequest(httpBody io.Reader, httpMethod, endpoint string,
...
 	// Make sure path values are http url safe
 	for _, pv := range pathValues {
 		safePathValues = append(safePathValues, url.QueryEscape(pv))
 	}
 	safeEndpoint := fmt.Sprintf(endpoint, safePathValues...)
```
jwhonce: Please add a TODO about converting to url library.  Thanks.

pkg/bindings/connection.go
```diff
@@ -150,8 +149,8 @@ func GetConnectionFromContext(ctx context.Context) (*Connection, error) {
 	if c == nil {
 		return nil, errors.New("unable to get connection from context")
 	}
-	conn := c.(Connection)
-	return &conn, nil
+	conn := c.(*Connection)
```
jwhonce: Why not just return the pointer?

baude: it actually fails to work work like that. it was on the casting side.

pkg/bindings/connection.go
baude: sure

baude:
updated with review comments

PR #5017, opened 2020-01-29
Makefile cleanups

### LINKED ISSUE #4829: Makefile tab completion turned very slow
I commonly use tab completion to list Make targets. This has worked out well for a long time but at some point (didn't bisect yet), the completion started to take _veeerrry_ long.

Although this might not be a big issue I suspect that the Makefile is doing too much work and may need some house cleaning.

vrothberg:
A set of cleanups for the Makefile. I'm ultimately aiming at making it faster (and fix #4829) since it got _very_ slow over time.

I bisected the performance and found that the release targets are responsible:
https://github.com/containers/libpod/blob/master/Makefile#L364-L372

@cevich, are `podman-v$(RELEASE_NUMBER).tar.gz:` and the remote counterpart used in the CI? Generating `RELEASE_NUMBER` for each invocation of the Makefile is burning most of the time. Commenting the two targets renders a `make tab tab` completion fast again. If needed, any chance we could not embed that in the target name?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,12 +45,6 @@ ifeq ($(shell go help mod >/dev/null 2>&1 && echo true), true)
 	GO_BUILD=GO111MODULE=on $(GO) build -mod=vendor
 endif
 
-ifeq (,$(findstring systemd,$(BUILDTAGS)))
-$(warning \
-	Podman is being compiled without the systemd build tag.\
-	Install libsystemd on Ubuntu or systemd-devel on rpm based distro for journald support)
-endif
-
 BUILDTAGS_CROSS ?= containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper exclude_graphdriver_overlay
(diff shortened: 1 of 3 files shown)


cevich:
>  If needed, any chance we could not embed that in the target name?

I'm not sure, looking...

cevich:
...okay, the only user (I'm aware of) for that target is `contrib/cirrus/build_release.sh`.  Conveniently for implementing your suggestion,  that script only calls:

*  `make podman-remote-${CROSS_PLATFORM}-release` (where $CROSS_PLATFORM is known)
*  `make podman.msi`
*  `make podman-release`

So the file-name targets are not needed AFAICT, so long as the above targets function (and spit out the expected files).

Great detective work BTW :grinning:

vrothberg:
Thanks, @cevich! I updated the PR. The Makefile is now fast again. git-bisect to the rescue :)

vrothberg:
@cevich @baude @mheon @rhatdan PTAL

rhatdan:
LGTM

PR #5018, opened 2020-01-29
Deprecate & remove IsCtrSpecific in favor of IsAnon

### LINKED ISSUE #5009: Upgrading from 1.6 to 1.7 deletes named persistent volumes on `podman run --rm`

/kind bug

**Description**

Named volumes automatically created by version 1.6 of podman are retained through multiple invocations of `podman run --rm`.  For example, postgres databases that specified a named volume the first time that they ran will work through multiple invocations of the container.  Named volumes that are made manually with `podman volume create` will also work through multiple invocations of the container.  https://github.com/containers/libpod/blob/master/docs/source/markdown/podman-run.1.md mentions that "If no such named volume exists, Podman will create one."  That's great.  It's worked well for me for a few months now.

The trouble is after an upgrade to version 1.7, where the same `podman run --rm` invocation will work correctly the first time, but as soon as the container is stopped the named volume is deleted in the background.  The next invocation of `podman run --rm` with that same set of command line switches will create a new volume with no data in it and mount it normally.  The postgres database from our example becomes empty.  That's the case in particular that surprised me.

```
# Podman version 1.6.2-2.fc31
/bin/podman run -it --rm --mount type=volume,src=vol16,dst=/home/user/vol16 --name voltest ubuntu:18.04 /bin/bash -c 'echo "yo" > /home/user/vol16/test.txt'
/bin/podman volume inspect vol16 | grep -i containerspecific
          "ContainerSpecific": true
/bin/podman run -it --rm --mount type=volume,src=vol16,dst=/home/user/vol16 --name voltest ubuntu:18.04 /bin/bash -c 'cat /home/user/vol16/test.txt'
yo
/bin/podman run -it --rm --mount type=volume,src=vol16,dst=/home/user/vol16 --name voltest ubuntu:18.04 /bin/bash -c 'cat /home/user/vol16/test.txt'
yo

# Do an in-place upgrade

# Podman version 1.7.0-2.fc31
/bin/podman run -it --rm --mount type=volume,src=vol16,dst=/home/user/vol16 --name voltest ubuntu:18.04 /bin/bash -c 'cat /home/user/vol16/test.txt'
yo
/bin/podman run -it --rm --mount type=volume,src=vol16,dst=/home/user/vol16 --name voltest ubuntu:18.04 /bin/bash -c 'cat /home/user/vol16/test.txt'
cat: /home/user/vol16/test.txt: No such file or directory
/bin/podman run -it --rm --mount type=volume,src=vol16,dst=/home/user/vol16 --name voltest ubuntu:18.04 /bin/bash -c 'cat /home/user/vol16/test.txt'
cat: /home/user/vol16/test.txt: No such file or directory
```

`cat: /home/user/vol16/test.txt: No such file or directory`

`yo`

Investigating this a bit shows that the meaning of the `ContainerSpecific` flag in the boltdb has changed slightly.  Whereas prior to https://github.com/containers/libpod/commit/0d623914d01bcbc10beebf2db966e17da215dfbb the `ContainerSpecific` flag was set to `true` for any auto-created volume including named ones, after that commit those are now `ContainerSpecific: false`.  Unfortunately https://github.com/containers/libpod/commit/0d623914d01bcbc10beebf2db966e17da215dfbb#diff-7b5097b329b325dc0490ec58b133a9aeR440 also changed the logic around volume deletion for containers started with `--rm`.  Now, any volume that is `ContainerSpecific: true`, regardless of named-ness gets deleted when the container stops.

As you might expect, volumes created with `podman volume create` have `ContainerSpecific: false` assigned, and are not affected by this situation.

I honestly don't know what to do about this.  I don't think that there's any other state information that we could glean from the database to determine whether it's a good idea to delete these volumes, but it seems like a really scary behaviour to mysteriously lose previously-persistent volumes.  Make it clear that automatically created volumes are different from manually created ones?  Update the code to ensure they're always made via the same codepath?  I'm open to suggestions here.

```
1.6.2 upgrading to 1.7.0
```
I'm pretty sure the change is in 1.6.3, but that's what I had to test with.

n/a

See above

This was tested on two physical machines, discovered on one running Fedora CoreOS with rpm-ostreed doing automatic updates, and then verified on one running Fedora 31's desktop spin.

mheon:
In Podman 1.6.3, we added support for anonymous volumes - fixing our old, broken support for named volumes that were created with containers. Unfortunately, this reused the database field we used for the old implementation, and toggled volume removal on for `podman run --rm` - so now, we were removing *named* volumes created with older versions of Podman.

We can't modify these old volumes in the DB, so the next-safest thing to do is swap to a new field to indicate volumes should be removed. Problem: Volumes created with 1.6.3 and up until this lands, even anonymous volumes, will not be removed. However, this is safer than removing too many volumes, as we were doing before.

Fixes #5009

### DIFF
--- a/libpod/container_commit.go
+++ b/libpod/container_commit.go
@@ -137,7 +137,7 @@ func (c *Container) Commit(ctx context.Context, destImage string, options Contai
 				if err != nil {
 					return nil, errors.Wrapf(err, "volume %s used in container %s has been removed", v.Name, c.ID())
 				}
-				if vol.IsCtrSpecific() {
+				if vol.Anonymous() {
 					importBuilder.AddVolume(v.Dest)
 				}
 			}
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -1546,17 +1546,16 @@ func WithVolumeGID(gid int) VolumeCreateOption {
 	}
 }
 
-// withSetCtrSpecific sets a bool notifying libpod that a volume was created
-// specifically for a container.
-// These volumes will be removed when the container is removed and volumes are
-// also specified for removal.
-func withSetCtrSpecific() VolumeCreateOption {
+// withSetAnon sets a bool notifying libpod that this volume is anonymous and
+// should be removed when containers using it are removed and volumes are
+// specified for removal.
+func withSetAnon() VolumeCreateOption {
 	return func(volume *Volume) error {
 		if volume.valid {
 			return define.ErrVolumeFinalized
 		}
 
-		volume.config.IsCtrSpecific = true
+		volume.config.IsAnon = true
 
 		return nil
 	}
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -319,7 +319,7 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (c *Contai
 		// The volume does not exist, so we need to create it.
 		volOptions := []VolumeCreateOption{WithVolumeName(vol.Name), WithVolumeUID(ctr.RootUID()), WithVolumeGID(ctr.RootGID())}
 		if isAnonymous {
-			volOptions = append(volOptions, withSetCtrSpecific())
+			volOptions = append(volOptions, withSetAnon())
 		}
 		newVol, err := r.newVolume(ctx, volOptions...)
 		if err != nil {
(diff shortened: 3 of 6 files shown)


rhatdan:
LGTM

PR #5019, opened 2020-01-29
Cirrus: Set EPOCH_TEST_COMMIT during gate task

cevich:
For whatever reason, this required variable is not set when `make` is
doing validation within automation.  Fix this by establishing the
value based on data provided from the CI system.  Since this data
is specific to automation executing against a PR, limit this specific
job to not run during testing of branches (post-merge).  This is safe
since we can assume the PR would not have been merged, if the basics
checked during gating did not pass.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -85,10 +85,15 @@ gce_instance:
 # quick format, lint, and unit tests on the standard platform.
 gating_task:
 
+    # Only run this on PRs, never during post-merge testing.  This is also required
+    # for proper setting of EPOCH_TEST_COMMIT value, required by validation tools.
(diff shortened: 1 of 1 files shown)


cevich:
@vrothberg I **think** this will fix the problem you noticed, where validation under automation is falsely passing.  PTAL.

cevich:
...actually, lets confirm gate is working by adding a very bad commit...

cevich:
[Yep, fails on signature check as expected.](https://cirrus-ci.com/task/6211020834209792)

(Removing the testing commit)

TomSweeneyRedHat:
LGTM assuming happy tests.  Fingers crossed for that bit.

PR #5020, opened 2020-01-29
Bindingtestrootless

baude:
thanks to giuseppe!

### DIFF
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -200,6 +200,9 @@ can_use_shortcut ()
   if (argv == NULL)
     return false;
 
+  if (strstr (argv[0], "podman") == NULL)
+    return false;
+
   for (argc = 0; argv[argc]; argc++)
     {
       if (argc == 0 || argv[argc][0] == '-')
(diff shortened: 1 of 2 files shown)


pkg/rootless/rootless_linux.c
```diff
@@ -200,6 +200,9 @@ can_use_shortcut ()
   if (argv == NULL)
     return false;
 
+  if (strstr (argv[0], "podman") == NULL)
```
rhatdan: What happens if I execute /usr/bin/podman?  Should this me just checking for paths ending in /podman or podman?

jwhonce: @rhatdan strstr() should only return NULL if "podman" is not contained in argv[0]

rhatdan: No that is fine. My lying eyes saw strcmp now I see strstr...

giuseppe: yes as long as podman is in the string then the process tries to join the user namespace.

I need to refactor this part and move the podman specific bits our of the rootless package

Issue #4809, opened 2020-01-07
Container Creation Entrypoint is a single string instead of a list


astronouth7303:
/kind bug

**Description**

The parameters to create a container (the `Create` type in the varlink schema) takes a single string. However, the OCI image configuration and the config key out of `InspectImage` can be an array.

**Steps to reproduce the issue:**

1. Try to create a container with an array entrypoint (either from an image with one or via varlink API)

**Describe the results you received:**
```
{'error': 'org.varlink.service.InvalidParameter', 'parameters': {'parameter': 'client.call.create.entrypoint'}}
```

**Describe the results you expected:**
Successful creation of a container

**Additional information you deem important (e.g. issue happens only occasionally):**
Based on https://github.com/containers/libpod/blob/bd3d8f4e282dc7a552d955bf5958b53d23937604/cmd/podman/varlink/io.podman.varlink#L318 I'm sure this is still in master.

**Output of `podman version`:**

```
Version:            1.6.4
RemoteAPI Version:  1
Go Version:         go1.13.5
Git Commit:         5cc92849f7fc9dd734ca2fd8f3ae8830b9a7eb26
Built:              Sat Dec 21 13:13:24 2019
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: 5cc92849f7fc9dd734ca2fd8f3ae8830b9a7eb26
  go version: go1.13.5
  podman version: 1.6.4
host:
  BuildahVersion: 1.12.0-dev
  CgroupVersion: v1
  Conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.2, commit: unknown'
  Distribution:
    distribution: debian
    version: unknown
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 188473344
  MemTotal: 8229838848
  OCIRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: |-
      runc version 1.0.0~rc9+dfsg1
      commit: 1.0.0~rc9+dfsg1-1
      spec: 1.0.1-dev
  SwapFree: 6091321344
[...]

PR #5022, opened 2020-01-30
Assure validate includes lint

ssbarnea:
Fixed issue where lint was not run on CI, so we missed to prevent regressions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -591,7 +591,7 @@ validate.completions: completions/bash/podman
 	if [ -x /bin/zsh ]; then /bin/zsh completions/zsh/_podman; fi
 
 .PHONY: validate
-validate: gofmt .gitvalidation validate.completions golangci-lint man-page-check
+validate: lint gofmt .gitvalidation validate.completions man-page-check
 
 .PHONY: build-all-new-commits
 build-all-new-commits:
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
The changes LGTM, but the tests are not at all happy and I'm not sure what's what there.

Makefile
```diff
@@ -536,7 +536,7 @@ validate.completions: completions/bash/podman
 	. completions/bash/podman
 	if [ -x /bin/zsh ]; then /bin/zsh completions/zsh/_podman; fi
 
-validate: gofmt .gitvalidation validate.completions golangci-lint man-page-check
+validate: lint gofmt .gitvalidation validate.completions golangci-lint man-page-check
```
edsantiago: This is probably not going to work, because the `lint` target requires `pre-commit` which may not exist on all testing environments (this was deliberate as part of #4809).

And please, if you do solve the dependency issue, do not add duplication here: you should not have both `lint` and `golangci-lint` on this line.

edsantiago:
LGTM. As I mentioned in #4807 I'm uncomfortable with this because I believe it will break `make validate` on some developer environments, but it seems to be passing CI and that's probably more important. Thanks for removing the duplication.

cevich:
not sure how to fix or re-run the `rdoproject.org/github-check` :confounded:

edsantiago:
>not sure how to fix or re-run the rdoproject.org/github-check

Has that one ever passed? I've never seen it green, and always just ignore it.

cevich:
Yeah, it does sometimes pass :smile:   We can ignore it I think.

Issue #5027, opened 2020-01-30
podman commit with different mandatory parameter than docker commit


loibpet:
/kind bug

**Description**

Ran into this problem, when I was checking if scripts developed for docker were running with podman.

podman supposed to be compatible to docker in terms of parameters. However, there are still breaking differences:

```shell
docker --version && docker commit --help
Docker version 1.13.1, build 07f3374/1.13.1

Usage:  docker commit [OPTIONS] CONTAINER [REPOSITORY[:TAG]]

Create a new image from a container's changes

Options:
  -a, --author string    Author (e.g., "John Hannibal Smith <hannibal@a-team.com>")
  -c, --change list      Apply Dockerfile instruction to the created image (default [])
      --help             Print usage
  -m, --message string   Commit message
  -p, --pause            Pause container during commit (default true)
```

whereas:

```shell
podman --version && podman commit --help
podman version 1.7.0
Create new image based on the changed container

Description:
Create an image from a container's changes. Optionally tag the image created, set the author with the --author flag, set the commit message with the --message flag, and make changes to the instructions with the --change flag.

Usage:
  podman commit [flags] CONTAINER IMAGE

Examples:
  podman commit -q --message "committing container to image" reverent_golick image-committed
  podman commit -q --author "firstName lastName" reverent_golick image-committed
  podman commit -q --pause=false containerID image-committed

Flags:
  -a, --author string        Set the author for the image committed
  -c, --change stringArray   Apply the following possible instructions to the created image (default []): CMD | ENTRYPOINT | ENV | EXPOSE | LABEL | ONBUILD | STOPSIGNAL | USER | VOLUME | WORKDIR
  -f, --format Format        Format of the image manifest and metadata (default "oci")
      --include-volumes      Include container volumes as image volumes
  -m, --message string       Set commit message for imported image
  -p, --pause                Pause container during commit
  -q, --quiet                Suppress output
```

The issue the second mandatory parameter (`IMAGE`) for the `podman` command.  This breaks any compatibility.

**Describe the results you expected:**

No additional mandatory parameter for parameter inherited from docker.

mheon:
@baude I remember we added this one waaaay back to solve a problem we were having... I want to say when we moved to using `buildah commit`?

@TomSweeneyRedHat Can buildah commit without a tag? We're using your code, so if you can, we ought to be able to

baude:
@mheon the cli should be able to do this.  lemme check.

TomSweeneyRedHat:
[...]

PR #5028, opened 2020-01-30
make image reference for commit optional

baude:
to match docker compat, the image tag should be optional.

Fixes: #5027


TomSweeneyRedHat:
LGTM
However, while testing I discovered an unrelated error with commit that should be touched up too.
https://github.com/containers/libpod/issues/5029

mheon:
LGTM but I'd like a final review from someone more familiar with images

PR #5030, opened 2020-01-30
fix longname handling for bindings

baude:
the api needs to account for image input where the image is encoded as a fqd image name.

### DIFF
--- a/pkg/api/handlers/handler.go
+++ b/pkg/api/handlers/handler.go
@@ -4,25 +4,16 @@ import (
 	"net/http"
 
 	"github.com/containers/libpod/libpod"
-	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
 )
 
 // Convenience routines to reduce boiler plate in handlers
 
-func getVar(r *http.Request, k string) string {
-	return mux.Vars(r)[k]
-}
-
 // func hasVar(r *http.Request, k string) bool {
 // 	_, found := mux.Vars(r)[k]
 // 	return found
 // }
-
-func getName(r *http.Request) string {
-	return getVar(r, "name")
-}
 
 func decodeQuery(r *http.Request, i interface{}) error {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
--- a/pkg/api/handlers/libpod/networks.go
+++ b/pkg/api/handlers/libpod/networks.go
@@ -1,0 +1,85 @@
+package libpod
+
+import (
+	"net/http"
+
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/pkg/api/handlers/utils"
+	"github.com/containers/libpod/pkg/network"
+	"github.com/gorilla/schema"
+	"github.com/pkg/errors"
+)
+
+func CreateNetwork(w http.ResponseWriter, r *http.Request) {}
+func ListNetworks(w http.ResponseWriter, r *http.Request) {
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	config, err := runtime.GetConfig()
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	configDir := config.CNIConfigDir
+	if len(configDir) < 1 {
+		configDir = network.CNIConfigDir
+	}
+	networks, err := network.LoadCNIConfsFromDir(configDir)
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	utils.WriteResponse(w, http.StatusOK, networks)
+}
+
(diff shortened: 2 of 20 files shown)


pkg/api/handlers/libpod/networks.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	name := utils.GetName(r)
+	if err := network.RemoveNetwork(name); err != nil {
+		// If the network cannot be found, we return a 404.
+		if errors.Cause(err) == network.ErrNetworkNotFound {
+			utils.Error(w, "Something went wrong", http.StatusNotFound, err)
```
jwhonce: Shouldn't message be "network not found" vs. "Something went wrong"?
You need a return after Error().

pkg/api/handlers/libpod/networks.go
```diff
@@ -0,0 +1,80 @@
...
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	name := utils.GetName(r)
+	n, err := network.InspectNetwork(name)
+	if err != nil {
+		// If the network cannot be found, we return a 404.
+		if errors.Cause(err) == network.ErrNetworkNotFound {
+			utils.Error(w, "Something went wrong", http.StatusNotFound, err)
```
jwhonce: Check message. You need return.

pkg/api/handlers/libpod/networks.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		utils.InternalServerError(w, err)
```
jwhonce: You need return

pkg/api/handlers/libpod/networks.go
```diff
@@ -0,0 +1,80 @@
...
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	name := utils.GetName(r)
+	if err := network.RemoveNetwork(name); err != nil {
+		// If the network cannot be found, we return a 404.
+		if errors.Cause(err) == network.ErrNetworkNotFound {
+			utils.Error(w, "Something went wrong", http.StatusNotFound, err)
+		}
+		utils.InternalServerError(w, err)
```
jwhonce: You need return

pkg/api/handlers/utils/handler.go
```diff
@@ -59,3 +61,19 @@ func FilterMapToString(filters map[string][]string) (string, error) {
...
+func getVar(r *http.Request, k string) string {
+	return mux.Vars(r)[k]
+}
+
+// GetName extracts the name from the mux and does unescape on
+// it.  Any error from the unescape is logged as Info because
+// the error is a user input error
+func GetName(r *http.Request) string {
+	val := getVar(r, "name")
+	name, err := url.PathUnescape(val)
```
jwhonce: 1) I would move the PathUnescape() into getVar() with the logging.  Also Wrapf() the error to provide some context on what is failing.
2) We agreed to return the value of the variable on errors not "" which this code does.

TomSweeneyRedHat:
LGTM but the RDOproject tests don't look hip.

pkg/api/handlers/utils/handler.go
```diff
@@ -59,3 +62,18 @@ func FilterMapToString(filters map[string][]string) (string, error) {
 	}
 	return string(f), nil
 }
+
+func getVar(r *http.Request, k string) string {
+	val := mux.Vars(r)[k]
+	safeVal, err := url.PathUnescape(val)
+	if err != nil {
+		logrus.Error(errors.Wrapf(err, "failed to unescape mux value %s", val))
```
jwhonce: k and url would make searching journald possible.  Are those included in the err you are wrapping?

pkg/api/handlers/handler.go
```diff
@@ -4,26 +4,17 @@ import (
 	"net/http"
```
jwhonce: Should this file now be deleted?

pkg/api/handlers/utils/handler.go
jwhonce: Does the wrapped err include the url? Otherwise the key and url would provide the complete context for debugging.

baude: stuff is still being used from this.

pkg/api/handlers/handler.go
baude: stuff is still being used by this

PR #5031, opened 2020-01-30
compat container names begin with /

baude:
when using the apiv2, containers names are preceeded with a /.  fixing this for inspect and listcontainers.


mheon:
LGTM, thanks. We might want to make this path-based in the future, to only trigger only compat endpoints not Libpod ones, but this ought to be fine for now

TomSweeneyRedHat:
LGTM assuming happy tests

Issue #5034, opened 2020-01-31
podman run --rm : Error forwarding signal 23 to container, plus stacktrace


edsantiago:
On rawhide: 
```
# podman run --rm alpine date
Fri Jan 31 12:34:45 UTC 2020
ERRO[0001] Error forwarding signal 23 to container 272087a3133469f8c302991b3b85327f56bd35d0a6536f6127b280238466b693: container has already been removed
ERRO[0001] Error forwarding signal 23 to container 272087a3133469f8c302991b3b85327f56bd35d0a6536f6127b280238466b693: container has already been removed
panic: close of closed channel

goroutine 72 [running]:
panic(0x5619d0bc2ec0, 0x5619d0e1e700)
        /usr/lib/golang/src/runtime/panic.go:1060 +0x424 fp=0xc0003e5ed0 sp=0xc0003e5e28 pc=0x5619cf4c8c24
runtime.closechan(0xc00007e9c0)
        /usr/lib/golang/src/runtime/chan.go:359 +0x253 fp=0xc0003e5f10 sp=0xc0003e5ed0 pc=0x5619cf49acf3
[100% reproducible, so I see no need to print the many hundreds of lines]
```
podman-1.7.1-0.66.dev.git36af283.fc32.x86_64
conmon-2.0.11-0.3.dev.gitccfdbb6.fc32.x86_64
5.5.0-0.rc6.git3.1.fc32.x86_64

vrothberg:
Looking into it. Installing a rawhide VM atm.

vrothberg:
Very curious to see a SIGURG flying around O.o

edsantiago:
I'm seeing this again on rawhide, but (1) on `podman run [without --rm]` and `podman rm` (not just `run --rm`) and (2) not reliably:
```
   $ podman run --name lcdcF3pEMqYnCsuPJXcHQv7b6IEK2U quay.io/libpod/alpine_labels:latest sleep 30
   $ podman rm -f lcdcF3pEMqYnCsuPJXcHQv7b6IEK2U
   4dbac6142c5d57aa113a2bf5374f264a4ac18a7bf71ec4272959ce2d4a645b05
[...]

PR #5036, opened 2020-01-31
sigproxy: return after closing the channel

vrothberg:
When stopping signal handling (e.g., to properly handle ^C) we are also
closing the signal channel.  We should really return from the go-routine
instead of continuing and risking double-closing the channel which leads
to a panic.

Fixes: #5034

### DIFF
--- a/pkg/adapter/sigproxy_linux.go
+++ b/pkg/adapter/sigproxy_linux.go
@@ -25,11 +25,17 @@ func ProxySignals(ctr *libpod.Container) {
 			}
 
 			if err := ctr.Kill(uint(s.(syscall.Signal))); err != nil {
+				// If the container dies, and we find out here,
+				// we need to forward that one signal to
+				// ourselves so that it is not lost, and then
+				// we terminate the proxy and let the defaults
+				// play out.
 				logrus.Errorf("Error forwarding signal %d to container %s: %v", s, ctr.ID(), err)
 				signal.StopCatch(sigBuffer)
 				if err := syscall.Kill(syscall.Getpid(), s.(syscall.Signal)); err != nil {
 					logrus.Errorf("failed to kill pid %d", syscall.Getpid())
 				}
+				return
 			}
 		}
 	}()


pkg/adapter/sigproxy_linux.go
```diff
@@ -30,6 +30,7 @@ func ProxySignals(ctr *libpod.Container) {
 				if err := syscall.Kill(syscall.Getpid(), s.(syscall.Signal)); err != nil {
```
mheon: Wait, what is this even doing? Why are we double-sending the signal to ourself on errors?

mheon: From git-blame, it looks like an attempt to make C-c kill Podman if sig-proxy fails because the container is stopped. I don't know if this is a good idea, or necessary at all, but it's probably not harmful

vrothberg: I had the very same question and found https://github.com/containers/libpod/pull/1086/files#r202198031.

vrothberg: But I am not sure about the correctness after all

vrothberg: Should this be the container?

vrothberg: Yes, that should be the container.

mtrmac: This makes (some) sense to me as is — as long as the container is running, attempts to gracefully terminate Podman are redirected to gracefully terminating the container. (The downside is that if the container is uncooperative, it’s difficult to terminate Podman.)

If the container dies, and we find out here, we need to forward that one signal to ourselves so that it is not lost, and then we terminate the proxy and let the defaults play out.

(Pedantically, at the time of `StopCatch`, several other signals may be queued in `sigBuffer`, and we throw them away. For the typical `SIGINT`/`SIGTERM` case this doesn’t matter much.)

vrothberg: Tests are green now. Mind dropping a LGTM if you agree to the changes?

mheon:
LGTM

TomSweeneyRedHat:
Could we add a comment similar to Miloslav's in the code to avoid future head scratching?
Otherwise LGTM

vrothberg:
> Could we add a comment similar to Miloslav's in the code to avoid future head scratching?
> Otherwise LGTM

Good point! I added a comment a repushed.

TomSweeneyRedHat:
LGTM, thanks for the comment.

rhatdan:
/lgtm
/hold
hold cancel once tests pass.

PR #5038, opened 2020-01-31
[CI:DOCS] Fix a syntax error in hack/release.sh

mheon:
When I fixed the release script up to remove epoch changes I missed this. Oops.


dougsland:
looks sane to me too.

PR #5039, opened 2020-01-31
Cirrus: Fix gate image & false-positive exits

cevich:
***Depends on:*** #5066 

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,6 +17,7 @@ env:
     DEST_BRANCH: "master"
     # Overrides default location (/tmp/cirrus) for repo clone
     GOPATH: "/var/tmp/go"
+    GOBIN: "${GOPATH}/bin"
     GOSRC: "/var/tmp/go/src/github.com/containers/libpod"
     CIRRUS_WORKING_DIR: "/var/tmp/go/src/github.com/containers/libpod"
     # The default is 'sh' if unspecified
@@ -85,17 +86,18 @@ gce_instance:
 # quick format, lint, and unit tests on the standard platform.
 gating_task:
 
-    # Only run this on PRs, never during post-merge testing.  This is also required
-    # for proper setting of EPOCH_TEST_COMMIT value, required by validation tools.
+    # Only run this on PRs, never during post-merge testing (for speed).
     only_if: $CIRRUS_BRANCH != $DEST_BRANCH
 
     env:
         CIRRUS_WORKING_DIR: "/usr/src/libpod"
-        GOPATH: "/go"
-        GOSRC: "/go/src/github.com/containers/libpod"
+        SRCPATH: "$CIRRUS_WORKING_DIR"
 
     # Runs within Cirrus's "community cluster"
     container:
+        # Note: Image has dual purpose, see contrib/gate/README.md
+        # The entrypoint.sh script ensures a prestine copy of $SRCPATH is
+        # available at $GOSRC before executing make instructions.
         image: "quay.io/libpod/gate:master"
         cpu: 4
         memory: 12
@@ -133,9 +135,9 @@ gating_task:
 
     # Verify some aspects of ci/related scripts
     ci_script:
-        - '${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/lib.sh.t |& ${TIMESTAMP}'
+        - '${GOSRC}/${SCRIPT_BASE}/lib.sh.t |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh -C ${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/packer test'
-        - '${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/cirrus_yaml_test.py |& ${TIMESTAMP}'
+        - '${GOSRC}/${SCRIPT_BASE}/cirrus_yaml_test.py |& ${TIMESTAMP}'
 
     # Verify expected bash environment (-o pipefail)
     pipefail_enabledscript: 'if /bin/false | /bin/true; then echo "pipefail fault" && exit 72; fi'
@@ -217,13 +219,14 @@ varlink_api_task:
 
     env:
         CIRRUS_WORKING_DIR: "/usr/src/libpod"
-        GOPATH: "/go"
-        GOSRC: "/go/src/github.com/containers/libpod"
+        SRCPATH: "$CIRRUS_WORKING_DIR"
+        EPOCH_TEST_COMMIT: "${CIRRUS_BASE_SHA}"  # repo clone missing this data
         # Used by tree_status.sh
         SUGGESTION: 'remove API.md, then "make varlink_api_generate" and commit changes.'
 
     # Runs within Cirrus's "community cluster"
     container:
+        # Note: Image has dual purpose, see contrib/gate/README.md
         image: "quay.io/libpod/gate:master"
         cpu: 4
         memory: 12
@@ -749,12 +752,13 @@ success_task:
 
     env:
         CIRRUS_WORKING_DIR: "/usr/src/libpod"
-        GOPATH: "/go"
-        GOSRC: "/go/src/github.com/containers/libpod"
+        SRCPATH: "$CIRRUS_WORKING_DIR"
+        EPOCH_TEST_COMMIT: "${CIRRUS_BASE_SHA}"  # repo clone missing this data
 
     container:
+        # Note: Image has dual purpose, see contrib/gate/README.md
         image: "quay.io/libpod/gate:master"
         cpu: 1
         memory: 1
 
-    success_script: '$CIRRUS_WORKING_DIR/$SCRIPT_BASE/success.sh |& ${TIMESTAMP}'
+    success_script: '/usr/local/bin/entrypoint.sh ./$SCRIPT_BASE/success.sh |& ${TIMESTAMP}'
--- a/contrib/gate/Dockerfile
+++ b/contrib/gate/Dockerfile
@@ -33,31 +33,36 @@ RUN dnf -y install \
       zip \
       && dnf clean all
 
-ENV GOPATH="/go" \
-    PATH="/go/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin" \
+ENV GOPATH="/var/tmp/go" \
+    GOBIN="/var/tmp/go/bin" \
+    PATH="/var/tmp/go/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin" \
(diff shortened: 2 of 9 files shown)


mheon:
/approve
LGTM but would like someone more familiar with bash to review

Makefile
```diff
@@ -104,6 +104,8 @@ GOBIN := $(shell $(GO) env GOBIN)
 ifeq ($(GOBIN),)
 GOBIN := $(FIRST_GOPATH)/bin
 endif
+# Required by some scripts
+export GOBIN
```
TomSweeneyRedHat: can you do this export  if it's not defined?  CI tests aren't happy.

cevich: removed

cevich:
> LGTM but would like someone more familiar with bash to review

Yeah, that and I need to improve the commit message and add some comments...

cevich:
@mheon @baude FYI- I still have some i-dotting and t-crossing to do here.  For the moment, it looks like whatever was causing download failures has passed, so I think there is less urgency to this PR.  However, I did uncover and fix a bunch of false-positive problems in our current setup.

- [ ] Confirm the `success` job works with the new gate container
- [ ] Confirm w/n a manual image build & push is needed or if the current master and PR usage of the new image will work as-is.
- [ ] Notify people to rebase their PRs if required once the new image is built and pushed.

cevich:
Note: There are a few documentation issues I need to clean up and additional local testing needed (see checklist above)

cevich:
@mheon @baude @edsantiago manual testing has shown, we will get gate-jobs failing for a short duration once this PR merges.  However, once quay builds the new image, re-running any failed jobs should fix them.

FWIW: I did experiment with trying the new code using the old container, and visa-versa but couldn't find a sane way to keep all use-cases happy.

.cirrus.yml
```diff
@@ -85,17 +86,18 @@ gce_instance:
...
 
     env:
         CIRRUS_WORKING_DIR: "/usr/src/libpod"
-        GOPATH: "/go"
-        GOSRC: "/go/src/github.com/containers/libpod"
+        SRCPATH: "$CIRRUS_WORKING_DIR"
 
     # Runs within Cirrus's "community cluster"
     container:
+        # Note: Image has dual-purpose, see contrib/gate/README.md
```
edsantiago: Nit: improper hyphen (should just be "Image has dual purpose"). Two other instances below.

contrib/gate/Dockerfile
```diff
@@ -33,31 +33,36 @@ RUN dnf -y install \
...
+# Only needed for installing build-time dependencies, then will be removed
 COPY / $GOSRC
 
-WORKDIR $GOSRC
-
 # Install dependencies
 RUN set -x && \
+    mkdir -p "$GOBIN" && \
+    mkdir -p /etc/cni/net.d && \
+    mkdir -p /etc/containers && \
```
edsantiago: Why this change? I find it more maintainable to group each mkdir with its install.

contrib/gate/entrypoint.sh
```diff
@@ -1,15 +1,23 @@
 #!/bin/bash
 
-[[ -n "$SRCPATH" ]] || \
-    ( echo "ERROR: \$SRCPATH must be non-empty" && exit 1 )
-[[ -n "$GOSRC" ]] || \
-    ( echo "ERROR: \$GOSRC must be non-empty" && exit 2 )
+set -e
+
+die() {
+    echo "${2:-FATAL ERROR (but no message given!) in ${FUNCNAME[1]}()}"
```
edsantiago: A nice habit to develop is embedding a helpful prefix in `die` messages: something that explains where the message is coming from. Usually `$(basename $0)` is good, but in this case `entrypoint.sh` is useless so it should be something like `gate` or `gate-container` or something that will help a poor user track down where the error is actually coming from.

contrib/gate/entrypoint.sh
```diff
@@ -1,15 +1,23 @@
...
+    exit ${1:-1}
+}
+
+[[ -n "$SRCPATH" ]] || die 1 "ERROR: \$SRCPATH must be non-empty"
+[[ -n "$GOPATH" ]] || die 2 "ERROR: \$$GOPATH must be non-empty"
```
edsantiago: Too many dollars.

hack/get_release_info.sh
```diff
@@ -7,7 +7,7 @@
 set -euo pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
edsantiago: DIR is no longer used, so please get rid of it

hack/install_golangci.sh
```diff
@@ -1,17 +1,16 @@
...
-	echo \$VERSION is empty
-	exit 1
-fi
+set -e
 
-if [ -z "$GOBIN" ]; then
-	echo \$GOBIN is empty
-	exit 1
-fi
+die() { echo "${1:-No error message given}"; exit 1; }
```
edsantiago: This is a good place to use `$(basename $0)` as a prefix in the `die` message.

hack/install_golangci.sh
```diff
@@ -1,17 +1,16 @@
...
-$GOBIN/golangci-lint --version | grep $VERSION
-if [ $?  -ne 0 ]; then
-	set -e
-	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOBIN v$VERSION
+BIN="$GOBIN/golangci-lint"
+if [ ! -x "$BIN" ]; then
+    echo "Installing golangci-lint v$VERSION into $GOBIN"
+    curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOBIN v$VERSION
+else
+    echo "Using existing golangci-lint v$($BIN --version | grep "$VERSION") at $BIN"
```
edsantiago: Whoa! What is this supposed to do? My guess is you want something like:
```bash
   echo "$BIN already installed: $($BIN --version)"
```

Makefile
```diff
@@ -536,7 +536,7 @@ uninstall:
 .PHONY: .gitvalidation
 .gitvalidation: .gopathok
 	@echo "Validating vs commit '$(call err_if_empty,EPOCH_TEST_COMMIT)'"
-	GIT_CHECK_EXCLUDE="./vendor:docs/make.bat" $(GOBIN)/git-validation -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..$(HEAD)
+	GIT_CHECK_EXCLUDE="./vendor:docs/make.bat" $(GOBIN)/git-validation -v -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..$(HEAD)
```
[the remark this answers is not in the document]
cevich: oh right, good catch.  I thought that would help make the 'bad commit' EPOCH_TEST_COMMIT error more informative. But I just tried again, and even with `-v` it still only exits 128 w/o a messsage :disappointed:

.cirrus.yml
cevich: got it.

contrib/gate/Dockerfile
cevich: There are fewer layers to download this way.

contrib/gate/entrypoint.sh
cevich: Great suggestion.

hack/install_golangci.sh
cevich: wrong copy-paste from earlier version.  Fixing...

edsantiago:
Looks like a real failure in `success.sh`:
```
contrib/cirrus/lib.sh: line 41: cd: /var/tmp/go/src/github.com/containers/libpod: No such file or directory
```
There are too many containers and contexts for me to figure out the right solution for this one; I'll trust that it's easy for you to do so.

cevich:
> There are too many containers and contexts for me to figure out the right solution for this one; I'll trust that it's easy for you to do so.

This is simply a reflection of what I confirmed earlier.  Due to unification of paths used in the gate-container scripts (so they're common for all CI jobs) the "old" (built from master prior to this PR) image cannot work with this PR.  @mheon will have to merge this manually.

Once this code is on master, Quay will build a new gate container that will function properly.  However, during the interumn time before it's built, this task will simply fail everywhere.

OTOH, I think having commonly used paths everywhere is a worthwhile change to suffer a few failures over.

Make sense?

edsantiago:
>Make sense?

I guess. I agree with the elimination of conflicting/duplicate/inconsistent `$GOSRC` (although there's still one unfortunate duplication that I don't know how to get rid of). I just have a hard time understanding all the container setup and wasn't sure if this error was "something got missed" or "it will get fixed when the magic container gets rebuilt". I will trust your judgment that it's the latter.  Thanks.

contrib/gate/Dockerfile
edsantiago: My priorities are maintainability >> optimizations, but I won't block because of this.

edsantiago:
LGTM except for the `mkdir` optimization. Nice work.

cevich:
@mheon @baude PTAL, this one needs a manual merge.  I've manually tested the new gate container as best as I can, but there's no simple way to get it the "old" gate image to pass with this specific PR.

mheon:
@cevich Do we have a way to back this out quickly if it goes bad? Want to make sure we aren't going to end up with a red master and no plan to resolve it

cevich:
@mheon The backout plan is: Revert this commit.  Allow quay.io to notice and build the "old" gate container.

Backout plan B: I will make a copy of the current gate container tagged "reverted_5039"

mheon:
@cevich Ack. Let's wait until I cut 1.8.0 and then get this in.

cevich:
yep, np.

Tag created: quay.io/libpod/gate:reverted_5039

cevich:
@mheon we released 1.8 already no?

mheon:
1.8.0 is out, 1.8.1 is in RC

mheon:
So yes, we should probably go ahead with this

edsantiago:
>1.8.1 is in RC

Um, how certain are you? RC1 had rootless regressions, and there haven't been any builds since then...

mheon:
We cut an RC2 tag, didn't we?

mheon:
https://github.com/containers/libpod/releases/tag/v1.8.1-rc2

Must not have been built?

edsantiago:
fc31 build [failed in koji](https://koji.fedoraproject.org/koji/buildinfo?buildID=1472551), I saw no other attempted builds

cevich:
All well and good, but **THIS** pr is needed to improve the gate job accuracy and documentation :smile:   It's also got some important fixes to avoid running container images produced from other branches.

edsantiago:
Way over my head, but LGTM in principle.

PR #5041, opened 2020-01-31
Bump to v1.8.0-rc1

mheon:
As the title says


rhatdan:
LGTM

mheon:
Gating is green, and this is rebased against latest master - I think we must have other problems if it's red elsewhere

PR #5044, opened 2020-01-31
Update XML to not embed quote in PATH on windows

### LINKED ISSUE #4335: Error in setting path on windows
I've installed the "Latest remote client for Windows"  tried it and uninstalled.

After uninstalling it the user path on my machine is left to `%PATH%;C:\Program Files (x86)\RedHat\Podman"` (in the registry under `HKEY_CURRENT_USER\Environment`) which causes problems both because I no longer have the paths set on my user path and because the `"` left in the path causes problems with other scripts

I'm not sure if the error happened during install or uninstall, if it is podman that causes it or the installer used on windows (it is the only new application I installed before having this problem), but it may be worth looking into it.

Thanks

Edited to add the reference to the registry

jwhonce:
* Remove the unsupported commend in lua script

Fixes #4335

### DIFF
--- a/docs/links-to-html.lua
+++ b/docs/links-to-html.lua
@@ -1,4 +1,3 @@
-# links-to-html.lua
 function Link(el)
   el.target = string.gsub(el.target, "%.1.md", ".html")
   return el
--- a/contrib/msi/podman.wxs
+++ b/contrib/msi/podman.wxs
@@ -33,7 +33,7 @@
     </Directory>
 
     <Property Id="setx" Value="setx.exe"/>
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@giuseppe @vrothberg @mheon @TomSweeneyRedHat @baude PTAL

docs/links-to-html.lua
```diff
@@ -1,4 +1,3 @@
-# links-to-html.lua
```
TomSweeneyRedHat: Just triple checking, did you mean to remove this @jwhonce ?

PR #5060, opened 2020-02-03
[CI:DOCS] swagger: v2: libpod/images/{import,load,pull}

vrothberg:
Note: this commit is merely adding swagger documentation and the golang
stubs and types for the proposed endpoints.  The implementation will
follow in separate individual changes in the future.

The ultimate goal is to prevent the libpod API from exposing the rather
complex /images/create endpoint from Docker and split it into easier to
implement, use and comprehend endpoints with a more narrow focus.

 # Import

Add the v2 swagger documentation for the libpod/images/import endpoint.
Note that we have intend to have separate backend and not mix it up with
load since import allows for specifying a URL instead of a local
tarball.

 # Load

Complete the v2 swagger documentation for the libpod/images/load
endpoint.  Note that we are accounting for future plans to be able to
load multiple images from one oci/docker archive by returning an array
of image-load responses.

Also move the (incomplete) implementation of the generic endpoint to the
corresponding package and create a stub for the libpod handler, which
will be implemented once there's an agreement on the proposed API.

 # Pull

Add the v2 swagger documentation for the libpod/images/pull endpoint.
Similar to the load endpoint, we return an array since more than one
image can be pulled when the `all-tags` parameter is set.

### DIFF
--- a/pkg/api/handlers/generic/images.go
+++ b/pkg/api/handlers/generic/images.go
@@ -3,6 +3,7 @@ package generic
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -314,4 +315,48 @@ func GetImages(w http.ResponseWriter, r *http.Request) {
 		summaries[j] = is
 	}
 	utils.WriteResponse(w, http.StatusOK, summaries)
+}
+
+func LoadImages(w http.ResponseWriter, r *http.Request) {
+	// TODO this is basically wrong
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+
+	query := struct {
+		Changes map[string]string `json:"changes"`
+		Message string            `json:"message"`
+		Quiet   bool              `json:"quiet"`
+	}{
+		// This is where you can override the golang default value for one of fields
+	}
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+
+	var (
+		err    error
+		writer io.Writer
+	)
+	f, err := ioutil.TempFile("", "api_load.tar")
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to create tempfile"))
+		return
+	}
+	if err := handlers.SaveFromBody(f, r); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to write temporary file"))
+		return
+	}
+	id, err := runtime.LoadImage(r.Context(), "", f.Name(), writer, "")
+	//id, err := runtime.Import(r.Context())
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to load image"))
+		return
+	}
+	utils.WriteResponse(w, http.StatusOK, struct {
+		Stream string `json:"stream"`
+	}{
+		Stream: fmt.Sprintf("Loaded image: %s\n", id),
+	})
 }
--- a/pkg/api/server/register_images.go
+++ b/pkg/api/server/register_images.go
@@ -107,7 +107,7 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//     description: no error
 	//   500:
 	//     $ref: '#/responses/InternalError'
-	r.Handle(VersionedPath("/images/load"), APIHandler(s.Context, libpod.ImportImage)).Methods(http.MethodPost)
+	r.Handle(VersionedPath("/images/load"), APIHandler(s.Context, generic.LoadImages)).Methods(http.MethodPost)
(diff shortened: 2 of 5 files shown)


pkg/api/server/register_images.go
```diff
@@ -659,10 +655,71 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	//     $ref: "#/response/LibpodImagesImportResponse"
+	//   500:
+	//     $ref: '#/responses/InternalError'
+	r.Handle(VersionedPath("/libpod/images/import"), APIHandler(s.Context, libpod.ImagesImport)).Methods(http.MethodPost)
+	// swagger:operation GET /libpod/images/pull libpod libpodImagesPull
+	// ---
+	// tags:
+	//  - images
+	// summary: Import image
+	// description: Import a previosly exported tarball as an image.
```
mheon: This doesn't seem right

pkg/api/server/register_images.go
```diff
@@ -659,10 +655,71 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	// parameters:
+	//  - in: query
+	//    name: reference
+	//    description: Mandatory reference to the image (e.g., quay.io/image/name:tag)/
+	//    type: string
+	//  - in: query
```
mheon: Wait, aren't these a cookie?

mheon:
Should I review the code here too? It sounds like it's just stubs

vrothberg:
> Should I review the code here too? It sounds like it's just stubs

No. @baude asked me to tackle swagger first to have an agreement on the API. Implementation will follow (see commit message).

pkg/api/server/register_images.go
vrothberg: It depends what we want. We can also send a base64 encoded `DockerAuthConfig` (see c/images/types.DockerAuthConfig`).

However, since we're going through SSH to a remote machine, I don't think we need to pass around token.

pkg/api/server/register_images.go
vrothberg: A funny swap :^) Thanks for catching!

pkg/api/server/register_images.go
```diff
@@ -659,10 +655,71 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	//    name: credentials
+	//    description: username:password for the registry.
+	//    type: string
+	//  - in: query
+	//    name: tls-verife
```
baude: typo

pkg/api/server/register_images.go
baude: remember, you need to support both ssh and not.  This is not a remote api; it's an api (think cockpit)

baude:
@vrothberg looking good to me ... address the few comments here and there and im happy.

pkg/api/server/register_images.go
vrothberg: Let's make this a PS in today's meeting :+1:

vrothberg:
> @vrothberg looking good to me ... address the few comments here and there and im happy.

Done. The auth/credential handling is still open.

vrothberg:
No idea what's going with gating now

vrothberg:
```
[+0048s] GIT_CHECK_EXCLUDE="./vendor:docs/make.bat" /go/bin/git-validation -run DCO,short-subject,dangling-whitespace -range 23f795786224c27f737e9043e9d513f54fa35ba8..HEAD
[+0048s] time="2020-02-03T15:25:20Z" level=error msg="mm[git] cmd: \"git --no-pager log --pretty=format:%H 23f795786224c27f737e9043e9d513f54fa35ba8..HEAD\""
[+0048s] 2020/02/03 15:25:20 exit status 128
```

... works locally and has passed before. Not sure why git-validation goes south.

cevich:
Gate-job failure issue: https://github.com/containers/libpod/issues/5063

vrothberg:
@baude @mheon @jwhonce @rhatdan ready to go

vrothberg:
> @baude @mheon @jwhonce @rhatdan ready to go

Actually, let me use this PR to add the implementations on top as well.

pkg/api/handlers/generic/images.go
```diff
@@ -315,3 +316,47 @@ func GetImages(w http.ResponseWriter, r *http.Request) {
...
+	query := struct {
+		Changes map[string]string `json:"changes"`
+		Message string            `json:"message"`
+		Quiet   bool              `json:"quiet"`
+	}{
+		// This is where you can override the golang default value for one of fields
+	}
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
```
TomSweeneyRedHat: Maybe it's my fine Irish temper, but I hate seeing "Something went wrong" in an error message.  My first thought is "NO S**T!, now tell me what went wrong, don't waste time with info that doesn't help me solve my problem.".  I'd prefer just a plain simple "Error: "

But that might just be me.

vrothberg: So true!

I fear we inherited that from Docker but I guess we can improve that for the libpod endpoints. @mheon @jwhonce @baude WDYT

jwhonce: Please do improve on those messages.  In the swagger we use one model and that populates all the examples.  Brent and I planned to improve these in both trees over time.  The Cause and Message returns have also added to the "what do we want to say here" issue.

PR #5061, opened 2020-02-03
[CI:DOCS]add apiv2 endpoints for exec

baude:
add the openapi/swagger documentation for exec.  The subcommands added are create, inspect, resize, and start.

at the time of this writing, no structure is declared for the inspect response.  once the libpod work for this is complete, we can inherit and swaggerize it.

### DIFF
--- a/pkg/api/server/register_exec.go
+++ b/pkg/api/server/register_exec.go
@@ -1,0 +1,329 @@
+package server
+
+import (
+	"net/http"
+
+	"github.com/containers/libpod/pkg/api/handlers"
+	"github.com/gorilla/mux"
+)
+
+func (s *APIServer) registerExecHandlers(r *mux.Router) error {
+	// swagger:operation POST /containers/{name}/create compat createExec
+	// ---
+	// tags:
+	//   - exec (compat)
+	// summary: Create an exec instance
+	// description: Run a command inside a running container.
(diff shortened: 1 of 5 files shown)


pkg/api/server/register_exec.go
```diff
@@ -0,0 +1,330 @@
+package server
+
+import (
+	"github.com/containers/libpod/pkg/api/handlers"
```
mheon: Import ordering

pkg/api/server/register_exec.go
```diff
@@ -0,0 +1,330 @@
...
+	// parameters:
+	//  - in: path
+	//    name: name
+	//    type: string
+	//    required: true
+	//    description: name of container
+    //  - in: body
+    //    name: control
+    //    description: Attributes for create
+    //    schema:
```
mheon: I'm probably going to have a version of this in libpod, FYI - we can likely use that for swagger generation

mheon: Though, ours might have some extra stuff that we don't want in here, so I think we should keep this until we know

mheon:
You have legitimate lint errors

baude:
all fixed up

rhatdan:
LGTM

PR #5066, opened 2020-02-03
Cirrus: Fix gate task + make lint|validate

### LINKED ISSUE #5063: Cirrus-CI: Gate job lint/validation fails with error 128
/kind bug

**Description**

For pull requests, automation executes a series of 'make' targets against a clone of the author's forked/branched code.  A required optimization in this process is to limit the checks to only encompass a commit range beginning with a a known upstream branch commit (commonly referred to as the Epoch commit).

However, depending on the starting-point for the authors fork'd branch, the Epoch commit used during these checks may not be reachable.  This results in one or more gate task checks failing.  When this fails during the git validation check, the tool or script throws an error 128

1. Fork upstream libpod repostory

2. Create a working branch, make a small change and commit.

3. Open a pull request

Gate task fails with error 128 

Gate task passes

This environment variable set in `.cirrus.yml` is significant: 

```bash
EPOCH_TEST_COMMIT: "${CIRRUS_BASE_SHA}"
```

cevich:
Fixes #5063

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -93,7 +93,6 @@ gating_task:
         CIRRUS_WORKING_DIR: "/usr/src/libpod"
         GOPATH: "/go"
         GOSRC: "/go/src/github.com/containers/libpod"
-        EPOCH_TEST_COMMIT: "${CIRRUS_BASE_SHA}"
 
     # Runs within Cirrus's "community cluster"
     container:
@@ -102,6 +101,16 @@ gating_task:
         memory: 12
 
     timeout_in: 20m
+
+    # Custom cloning is required to satisfy lint/validation needs
+    clone_script: |
+        git clone --recursive --branch=$DEST_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+        git remote update origin
+        if [[ -n "$CIRRUS_PR" ]]; then # running for a PR
+            git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
+            git checkout pull/$CIRRUS_PR
+        fi
+        git reset --hard $CIRRUS_CHANGE_IN_REPO
 
     # Don't bother going further if something is down
     networking_script: '${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/networking.sh'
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -146,6 +149,7 @@ endif
 
 .PHONY: lint
 lint: golangci-lint
+	@echo "Linting vs commit '$(call err_if_empty,EPOCH_TEST_COMMIT)'"
```
edsantiago: I don't think `EPOCH_TEST_COMMIT` is actually used here, is it?

Makefile
```diff
@@ -3,7 +3,7 @@ export GOPROXY=https://proxy.golang.org
 
 GO ?= go
 DESTDIR ?=
-EPOCH_TEST_COMMIT ?= $(shell git merge-base HEAD $${DEST_BRANCH:-master})
+EPOCH_TEST_COMMIT ?= $(shell git merge-base $${DEST_BRANCH:-master} HEAD)
```
edsantiago: What is the etymology of this variable name? Since it doesn't seem to be used anywhere else, could it be called GIT_MERGE_BASE instead?

.cirrus.yml
```diff
@@ -93,7 +93,6 @@ gating_task:
         CIRRUS_WORKING_DIR: "/usr/src/libpod"
         GOPATH: "/go"
         GOSRC: "/go/src/github.com/containers/libpod"
-        EPOCH_TEST_COMMIT: "${CIRRUS_BASE_SHA}"
```
edsantiago: This now leaves a dangling instance of `EPOCH_TEST_COMMIT` in a comment in this file; that will confuse future maintainers

Makefile
cevich: My best guess is it's used in other `containers/*` repos. and that was copy-pasted over here when it was eventually needed.  Though arguably a bad name, we might consider keeping it to preserve consistency with the other projects.

Makefile
cevich: TBH I'm not sure if it's used by the underlying linting tool or not.  I'm not opposed to removing it.

edsantiago: lint acts on a snapshot, i.e. the current source tree. It should not know or care about git history.

.cirrus.yml
cevich: don't worry, both past and future maintainers are already confused.  It's a job requirement :smile: 

joking aside, I think #5039 will make the situation slightly better.

cevich: better meaning: I added some comments

Makefile
edsantiago: I see it used in buildah also. OK, keep it, I can hold my nose.

PR #5067, opened 2020-02-03
[CI:DOCS] Update /_ping support

jwhonce:
* Add swagger support and documentation
* Promote handler as there can only ever be one
* Update swagger tags

Between verbs and paths there are four endpoints for /_ping.  I've
attempted to document them well without four copies of the details.
Time will tell.

Clients will use the headers `Libpod-API-Version` and
`Libpod-Buildha-Version` to determine if they are connecting to a Podman
service or some other compatible engine.

Client calls GET(/_ping):
 * VARLINK engines will fail with error message
 * Compatible engines will return 200 without Libpod headers
 * Podman engines will return 200 and Libpod headers


baude:
yeah LGTM

PR #5070, opened 2020-02-03
add libpod create container

baude:
add the libpod endpoint for creating a container from a createconfig.  also, add input values via body for pod create.

### DIFF
--- a/pkg/spec/createconfig.go
+++ b/pkg/spec/createconfig.go
@@ -167,6 +167,7 @@ type SecurityConfig struct {
 }
 
 // CreateConfig is a pre OCI spec structure.  It represents user input from varlink or the CLI
+// swagger:model CreateConfig
 type CreateConfig struct {
 	Annotations       map[string]string
 	Args              []string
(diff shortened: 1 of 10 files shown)


pkg/spec/createconfig.go
```diff
@@ -167,6 +167,7 @@ type SecurityConfig struct {
 }
 
 // CreateConfig is a pre OCI spec structure.  It represents user input from varlink or the CLI
+// swagger:model CreateConfig
```
mheon: I don't want to formalize this in the API - it's a terrible structure that we'd love to replace. I don't think we want to engrave this into the API for all eternity as the definitive standard that containers are built from. We should either mark this as subject to change, or do a full review on the struct and see what can be done to make it less insane to support.

baude: @mheon what exactly do you propose then? nobody has the time to review this entire structure and rewrite it.  you want to make a 100+ option list? no.  and nobody wants to rewrite all the logic to go from this to a container...

mheon: This will be an absolute nightmare to support. There are fields in there that are internal use only that are, nonetheless, public and open to use, and I have no idea what happens if those are populated outside of our tools. We are setting ourselves us for a difficult to use, impossible to support API if we expose things like this.

Way I see it, we have two options: try and rework CreateConfig into something more sane (remove internal-only options, remove fields that require parsing where we can), or do as @jwhonce suggested and add fields to the generic endpoint that we need, and compile that into a CreateConfig - this doesn't couple us to an internal representation, and we know what's in it.

PR #5074, opened 2020-02-03
Add /swagger/ endpoint to serve swagger yaml to clients

jwhonce:
The provided yaml file will describe the current Podman REST API.

### DIFF
--- a/pkg/api/server/register_swagger.go
+++ b/pkg/api/server/register_swagger.go
@@ -1,0 +1,26 @@
+package server
+
+import (
+	"net/http"
+	"os"
+
+	"github.com/gorilla/mux"
+)
+
+// DefaultPodmanSwaggerSpec provides the default path to the podman swagger spec file
+const DefaultPodmanSwaggerSpec = "/usr/share/containers/podman/swagger.yaml"
+
+// RegisterSwaggerHandlers maps the swagger endpoint for the server
+func (s *APIServer) RegisterSwaggerHandlers(r *mux.Router) error {
+	// This handler does _*NOT*_ provide an UI rather just a swagger spec that an UI could render
+	r.PathPrefix("/swagger/").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		path := DefaultPodmanSwaggerSpec
+		if p, found := os.LookupEnv("PODMAN_SWAGGER_SPEC"); found {
+			path = p
+		}
+		w.Header().Set("Content-Type", "text/yaml")
+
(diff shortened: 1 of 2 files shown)


pkg/api/server/register_swagger.go
```diff
@@ -0,0 +1,16 @@
...
+	"net/http"
+
+	"github.com/gorilla/mux"
+)
+
+// RegisterSwaggerHandlers maps the swagger endpoint for the server
+func (s *APIServer) RegisterSwaggerHandlers(r *mux.Router) error {
+	r.PathPrefix("/swagger/").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/yaml")
+		http.ServeFile(w, r, "/usr/share/podman/swagger.yaml")
```
baude: how can we deal with working from master for example? is there any way we can make this more flexible?

baude: maybe honoring an ENV var?

mheon: This will also probably live somewhere under `/usr/share/containers` once packaged

edsantiago: For clarity, would you consider renaming `/swagger/` -> `/swagger.yaml`? There are tools that expect JSON, and I'm sure they'll deal just fine with HTTP headers, but humans looking at the URL might appreciate a visual cue as to what to expect.

And yeah, it would be nice to have an override for the hardcoded path.

jwhonce: @baude  No.  I don't want us checking in a file that is generated.  Yes.  I'll make the path a default using @mheon 's input and add PODMAN_SWAGGER_SPEC.

@edsantiago This endpoint isn't for people, it's for clients to use. They expect /swagger/  . I'll add a comment to that fact.  I didn't add swagger annotations because that is a hen/chick issue.

edsantiago: Flexibility is exactly why I'm requesting it. Perhaps some tools expect `/swagger/` but here is one counterexample which demonstrates that others explicitly request `.json` or `.yaml`: https://github.com/kiwicom/schemathesis

In fifteen examples in the README, all of them explicitly request `.json` or `.yaml`. I believe that is the more user-friendly approach.

baude:
LGTM, @rhatdan ptal and merge

edsantiago:
Before pulling the trigger, can someone please ELI5 why `/swagger/` is preferable to the clear and explicit and futureproof `/swagger.yaml` ?

PR #5078, opened 2020-02-04
Skip empty hooks directories by default

saschagrunert:
We should not consider empty hooks directories because they could lead
into inappropriate error messages. As a background information, we would
like to be able to specify empty hooks directories in CRI-O without further
processing of the hooks-dir configuration string slice.

Relates to https://github.com/cri-o/cri-o/issues/3179

### DIFF
--- a/pkg/hooks/hooks.go
+++ b/pkg/hooks/hooks.go
@@ -42,11 +42,11 @@ type namedHook struct {
 // New creates a new hook manager.  Directories are ordered by
 // increasing preference (hook configurations in later directories
 // override configurations with the same filename from earlier
-// directories).
+// directories).  Empty directories will be skipped automatically.
 //
 // extensionStages allows callers to add additional stages beyond
 // those specified in the OCI Runtime Specification and to control
-// OCI-defined stages instead of delagating to the OCI runtime.  See
+// OCI-defined stages instead of delegating to the OCI runtime.  See
 // Hooks() for more information.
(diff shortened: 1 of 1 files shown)


rhatdan:
/approve
LGTM

mheon:
Why are we adding empty-string hooks directories? I'm really confused as to what's going on here - this seems somewhat questionable?

TomSweeneyRedHat:
If we should be skipping empty directories, code LGTM.  Interested in hearing the answer to @mheon's question.

rhatdan:
I believe the code search 
/usr/share/containers/oci/hooks.d
/etc/containers/oci/hooks.d
If either one of these is empty then we should ignore it.
@saschagrunert Is this the issue?

mheon:
I'm fine with not erroring on hooks directories that are specified, but don't exist, but this seems to be ignoring *the empty string* - which is very different, and looks like invalid input?

mheon:
If it's just directories that are specified not existing, we should already catch that lower in the file?

saschagrunert:
> I'm fine with not erroring on hooks directories that are specified, but don't exist, but this seems to be ignoring _the empty string_ - which is very different, and looks like invalid input?

Sorry for the confusion. Yeah it's about invalid input. I played around with CRI-O (using this part of libpod) and had multiple combinations of CLI and config parameters which more or less resulted in passing a `[]string{""}` slice to this function. I'll close it for now and check if we can solve it in CRI-O in a more meaningful way.

PR #5082, opened 2020-02-04
[CI:DOCS] Spec proposal

baude:
warning: the naming of this might change as well as the location.
    
this is a build on a PR from mheon from last year that proposes a shift from our current approach of creating containers based on the arbitrarily made createconfig.  the new approach would be to have a specification that is detached from the podman cli.  the spec could then be generated and used to make a container.  this theoretically is the beginning of a long-needed refactor involving how we get from the cli -> libpod | apiv2 -> libpod with code re-use and less duplication.
    
the intent is to build the apiv2 container creation based on this approach only.  wiring to the podman cli will happen after the fact.

### DIFF
--- a/pkg/specgen/specgen.go
+++ b/pkg/specgen/specgen.go
@@ -1,0 +1,363 @@
+package specgen
+
+import (
+	"net"
+
+	"github.com/containers/libpod/libpod"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// TODO
+type Namespace struct {
+}
+
+// ContainerBasicConfig contains the basic parts of a container.
+type ContainerBasicConfig struct {
+	// Name is the name the container will be given.
+	// If no name is provided, one will be randomly generated.
+	// Optional.
+	Name string
+	// Pod is the ID of the pod the container will join.
+	// Optional.
+	Pod string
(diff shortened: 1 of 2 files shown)


pkg/specgen/specgen.go
```diff
@@ -0,0 +1,399 @@
...
+	// OOMScoreAdj adjusts the score used by the OOM killer to determine
+	// processes to kill for the container's process.
+	// Optional.
+	OOMScoreAdj *int `json:"oom_score_adj,omitempty"`
+}
+
+// ContainerHealthCheckConfig describes a container healthcheck with attributes
+// like command, retries, interval, start period, and timeout.
+type ContainerHealthCheckConfig struct {
+	manifest.Schema2HealthConfig `json:"healthcheck,omitempty"`
```
mheon: Do we want to give this a name? In case we add extra fields related to healthchecks at some point?

baude: @mheon i couldnt think of a name that wasnt redundant.  you?

mheon: ...HealthConfig? Maybe?

baude:
updated

mheon:
Current version LGTM, we can discuss further details as the implementation comes together

PR #5083, opened 2020-02-04
v2 api: /libpod/images/{import,load,pull}

vrothberg:
Implement the v2 endpoints for pulling, loading and importing images.


vrothberg:
@jwhonce @baude PTAL

The "import" and "load" endpoints will follow soon but given my workday is about to end, you may already have a look at "pull".

pkg/api/handlers/libpod/images.go
```diff
@@ -186,6 +193,118 @@ func ImagesImport(w http.ResponseWriter, r *http.Request) {
...
 
 func ImagesPull(w http.ResponseWriter, r *http.Request) {
-	//TODO ...
-	utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.New("/libpod/images/pull is not yet implemented"))
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	query := struct {
+		Reference    string `schema:"reference"`
+		Credentials  string `schema:"credentials"`
+		OverrideOS   string `schema:"override-os"`
```
jwhonce: camelCase is preferred to dashes, to save on length.  eg., overrideOs vs override-os

pkg/api/handlers/libpod/images.go
```diff
@@ -186,6 +193,118 @@ func ImagesImport(w http.ResponseWriter, r *http.Request) {
...
+		TLSVerify: true,
+	}
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+
+	if len(query.Reference) == 0 {
```
jwhonce: I find below more obvious:

if _, found := mux.Vars(r)["reference"]; !found {
...
}

pkg/api/server/register_images.go
```diff
@@ -709,18 +709,32 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
 	// - application/json
 	// responses:
 	//   200:
 	//     $ref: "#/response/LibpodImagesPullResponse"
+	//   400:
+	//     $ref: "#/responses/BadParamError"
 	//   500:
 	//     $ref: '#/responses/InternalError'
-	r.Handle(VersionedPath("/libpod/images/pull"), APIHandler(s.Context, libpod.ImagesPull)).Methods(http.MethodPost)
+	r.Handle(VersionedPath("/libpod/images/pull"), APIHandler(s.Context, libpod.ImagesPull)).Methods(http.MethodGet)
```
jwhonce: This operation is Post not Get.

cmd/podman/pull.go
```diff
@@ -101,6 +101,7 @@ func pullCmd(c *cliconfig.PullValues) (retError error) {
 		}
 	}
 
+	// FIXME: that's a bug. What if we pass "localhost:5000/no-tag" ?
 	arr := strings.SplitN(args[0], ":", 2)
```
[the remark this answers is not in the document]
vrothberg: Fixing `cmd/podman/pull.go` is out of scope of this PR which is why I added the FIXME.

FWIW, we cannot parse docker references with `net/url.Parse`. There's a load of really nasty corner cases. Using `c/image` (as done in the PR) is the only way to be sure we're using a correct parser.

pkg/api/handlers/libpod/images.go
vrothberg: Good point, will do!

pkg/api/handlers/libpod/images.go
vrothberg: Sure, can do

pkg/api/server/register_images.go
vrothberg: Can you elaborate why? I changed it to Get because everything is done via parameters.

vrothberg: I peeked into https://www.w3schools.com/tags/ref_httpmethods.asp and agree that it should be POST.

pkg/api/handlers/libpod/images.go
vrothberg: Just about to perform the change, I actually prefer the current way. It's faster as we don't need to look up and don't duplicate the schema string but can reference directly via the struct.

vrothberg:
@mheon @rhatdan @baude @jwhonce ready for review, PTAL.

vrothberg:
```
Error initializing source docker://registry.fedoraproject.org/fedora-minimal:latest: error pinging docker registry registry.fedoraproject.org: invalid status code from registry 503 (Service Unavailable)
```
registry.fedoraproject.org is striking again

baude:
LGTM

pkg/api/server/register_images.go
```diff
@@ -688,37 +692,53 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
 	// - application/json
 	// responses:
 	//   200:
 	//     $ref: "#/responses/DocsLibpodImagesPullResponse"
+	//   400:
+	//     $ref: "#/responses/BadParamError"
 	//   500:
 	//     $ref: '#/responses/InternalError'
-	r.Handle(VersionedPath("/libpod/images/pull"), APIHandler(s.Context, libpod.ImagesPull)).Methods(http.MethodPost)
+	r.Handle(VersionedPath("/libpod/images/pull"), APIHandler(s.Context, libpod.ImagesPull)).Methods(http.MethodPut)
```
jwhonce: Documentation says Post here not Put.

vrothberg: Great catch! Fixed now.

vrothberg:
Green again. Ready to merge.

PR #5088, opened 2020-02-04
Begin exec rework

mheon:
As part of the rework of exec sessions, we need to address them independently of containers. In the new API, we need to be able to fetch them by their ID, regardless of what container they are associated with. Unfortunately, our existing exec sessions are tied to individual containers; there's no way to tell what container a session belongs to and retrieve it without getting every exec session for every container.

I originally debated just adding a registry to the database to point individual exec sessions at each container, but if we're going so far as to completely restructure exec sessions, it makes sense to store them separately as well. Existing exec sessions will be retained as-is for legacy purposes, while the new-style exec sessions will be stored separately from containers and accessed via their own API endpoints.

There's some potential performance impact to be discussed here in that we'll now need one DB call to get each exec session, and then one call per exec session to update its state, whenever we need to do an operation that needs to check if there are running exec sessions. I'm also concerned as to the circumstances of them remaining in the database; I need to verify on the Docker side as to when they are cleaned (only on reboot? on container restart as well? when the exec session finishes?).
    
This is very early; I'm only about halfway through DB changes, which will then require a refactor of existing exec code to use the new fields in the DB, and associated changes to the lifecycle of exec sessions.

TODO:

- [x] Finish updating database for new storage scheme of exec sessions
- [x] Update Exec() API in container_exec.go to use new exec session storage
- [x] Create function for checking if any exec sessions are running, and use this to replace all existing checks of that type
- [x] Update container removal to deal with new exec sessions
- [ ] Investigate when exec sessions are removed from the database and update our code to follow that lifecycle
- [ ] Investigate cleanup processes for exec sessions so that we can support detached exec

### DIFF
--- a/libpod/boltdb_state.go
+++ b/libpod/boltdb_state.go
@@ -41,6 +41,8 @@ type BoltState struct {
 //   containing the path to the container's network namespace, a dependencies
 //   bucket containing the container's dependencies, and an optional pod key
 //   containing the ID of the pod the container is joined to.
+//   After updates to include exec sessions, may also include an exec bucket
+//   with the IDs of exec sessions currently in use by the container.
 // - allCtrsBkt: Map of ID to name containing only containers. Used for
 //   container lookup operations.
 // - podBkt: Contains a sub-bucket for each pod in the state.
@@ -49,6 +51,10 @@ type BoltState struct {
 //   containers in the pod.
 // - allPodsBkt: Map of ID to name containing only pods. Used for pod lookup
 //   operations.
+// - execBkt: Map of exec session ID to exec session - contains a sub-bucket for
+//   each exec session in the DB.
+// - execRegistryBkt: Map of exec session ID to nothing. Contains one entry for
+//   each exec session. Used for iterating through all exec sessions.
 // - runtimeConfigBkt: Contains configuration of the libpod instance that
 //   initially created the database. This must match for any further instances
 //   that access the database, to ensure that state mismatches with
@@ -86,6 +92,7 @@ func NewBoltState(path string, runtime *Runtime) (State, error) {
 		allPodsBkt,
 		volBkt,
 		allVolsBkt,
+		execBkt,
 		runtimeConfigBkt,
 	}
 
@@ -171,6 +178,11 @@ func (s *BoltState) Refresh() error {
 			return err
 		}
 
+		execBucket, err := getExecBucket(tx)
+		if err != nil {
+			return err
+		}
+
 		// Iterate through all IDs. Check if they are containers.
 		// If they are, unmarshal their state, and then clear
 		// PID, mountpoint, and state for all of them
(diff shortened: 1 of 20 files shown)


mheon:
Update: looking more into exec sessions in Docker, it looks like they are removed immediately after termination - probably just the exit code is retrieved via Inspect. We will need the ability to autoremove exec sessions for detached containers. I still see some potential for leaking, but we can probably counter that by clearing them on container restart.

mheon:
DB changes should be complete. Still doesn't compile, but all functions are there. Going to convince everything to use the new exec sessions, then write unit tests for the DB.

mheon:
Exec API has been rewritten to use the new structure.
Next steps: rewrite remainder of Libpod and cmd/podman to use the new APIs, write unit tests for DB work, implement missing APIs (Inspect, Resize).

mheon:
Libpod work mostly done. It compiles, even.

Next steps: need to rewrite resize code, write unit tests for DB code.

mheon:
Added resizing. Going to chase down test failures, add some unit tests, then we can merge.

mheon:
Rebased, healthcheck errors should be fixed. Stripping WIP, should be ready for merge.

rhatdan:
@mheon Needs rebase.

mheon:
Should be ready after #5429 merges

mheon:
Rebased, repushed. @rhatdan @baude @vrothberg @giuseppe @haircommander PTAL

libpod/boltdb_state.go
```diff
@@ -895,6 +950,287 @@ func (s *BoltState) GetContainerConfig(id string) (*ContainerConfig, error) {
...
+			return err
+		}
+		ctrBucket, err := getCtrBucket(tx)
+		if err != nil {
+			return err
+		}
+
+		sessionExists := execBucket.Get(sessionID)
+		if sessionExists == nil {
+			return define.ErrNoSuchExecSession
```
baude: wrap with id?

baude: up to you i guess, if the caller wraps it, then ok too

libpod/boltdb_state_internal.go
```diff
@@ -805,6 +832,23 @@ func (s *BoltState) removeContainer(ctr *Container, pod *Pod, tx *bolt.Tx) error
...
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+		if len(sessions) != 0 {
+			return errors.Wrapf(define.ErrCtrExists, "container %s has active exec sessions: %s", ctr.ID(), strings.Join(sessions, ", "))
+		}
+	}
+
```
baude: this block and the previous block at 800 appear to be the same and also have the same error message.  any chance we could stub that out?

libpod/container_exec.go
```diff
@@ -0,0 +1,765 @@
...
+		}
+	}
+
+	session, ok := c.state.ExecSessions[sessionID]
+	if !ok {
+		return errors.Wrapf(define.ErrNoSuchExecSession, "container %s has no exec session with ID %s", c.ID(), sessionID)
+	}
+
+	if session.State != define.ExecStateRunning {
+		return errors.Wrapf(define.ErrExecSessionStateInvalid, "can only stop running exec sessions, while container %s session %s state is %q", c.ID(), session.ID(), session.State.String())
```
baude: % is not running, current state %s ?

baude:
lgtm, couple of things to think about

libpod/boltdb_state_internal.go
mheon: Actually they're completely redundant. Oops.

libpod/container_exec.go
mheon: Fixed

mheon:
Rebased on top of fixed Conmon. Fingers are crossed.

rhatdan:
@mheon is this the same issue you were seeing or a different one?

mheon:
Tests are (amazingly) passing now, save a network flake.

@TomSweeneyRedHat @rhatdan @vrothberg PTAL, let's see about getting this landed

mheon:
@rhatdan Different issue, looks like a flake in the APIv2 tests

rhatdan:
@mheon Do we want to get this into podman 1.8.2 or wait until we ship it?

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM and happy green tests buttons.

rhatdan:
/hold
/lgtm
I want @containers/virtcontainers-repo-admin to cancel the hold if he wants it in the 1.8.2 release or afterwards.

PR #5089, opened 2020-02-04
fix swagger docs and make sure docs validation runs

baude:

### DIFF
--- a/pkg/api/server/register_images.go
+++ b/pkg/api/server/register_images.go
@@ -638,25 +638,24 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	// summary: Load image
 	// description: Load an image (oci-archive or docker-archive) stream.
 	// parameters:
-	//  - in: query
-	//    name: change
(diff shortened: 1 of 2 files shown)


pkg/api/server/register_images.go
```diff
@@ -638,25 +638,25 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	//     type: string
+	//   - in: query
+	//     name: message
+	//     description: Set commit message for imported image
+	//     type: string
+	//   - in: body
+	//     name: request
+	//     description: tarball of container image
+	//     schema:
+	//       type: string
```
jwhonce: If this is tarball that is not uuencoded, I think you want to add the property of "format: binary".

pkg/api/server/register_images.go
```diff
@@ -665,31 +665,31 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	//     description: Set commit message for imported image
+	//     type: string
+	//   - in: query
+	//     name: url
+	//     description: Specify a URL instead of a tarball
+	//     type: boolean
+	//   - in: body
+	//     name: request
+	//     schema:
+	//       type: string
```
jwhonce: Again, I think we want the 'format: binary' property here.

baude:
updated

PR #5098, opened 2020-02-05
Special case memory-swap=-1

rhatdan:
We document that memory-swap==-1 means unlimited, but currently we
won't allow the user to specify the -1 value.

Fixes: https://github.com/containers/libpod/issues/5091

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -309,9 +309,13 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 		}
 	}
 	if c.String("memory-swap") != "" {
-		memorySwap, err = units.RAMInBytes(c.String("memory-swap"))
-		if err != nil {
-			return nil, errors.Wrapf(err, "invalid value for memory-swap")
+		if c.String("memory-swap") == "-1" {
+			memorySwap = -1
+		} else {
+			memorySwap, err = units.RAMInBytes(c.String("memory-swap"))
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid value for memory-swap")
+			}
 		}
 	}
 	if c.String("kernel-memory") != "" {


rhatdan:
@s-winter does this fix your problem?

s-winter:
> @s-winter does this fix your problem?

Thank you for looking into the issue. Unfortunately, I cannot tell if your change fixed my problem or not. I just tried running podman built from your fork of libpod and it fails. I tried merging containers/libpod master and it fails. I spent around an hour googling for the error messages, applying fixes and workarounds here and there and it still fails. I will look into this again when I find more time.

giuseppe:
> Thank you for looking into the issue. Unfortunately, I cannot tell if your change fixed my problem or not. I just tried running podman built from your fork of libpod and it fails. I tried merging containers/libpod master and it fails. I spent around an hour googling for the error messages, applying fixes and workarounds here and there and it still fails. I will look into this again when I find more time.

can you please share the error message?  It might be related to crun and how the `memory-swap` is mapped to cgroups v2, as it works differently from cgroup v1.

Done some work recently in this area: https://github.com/containers/crun/pull/226

s-winter:
> can you please share the error message? It might be related to crun and how the `memory-swap` is mapped to cgroups v2, as it works differently from cgroup v1.

Invoking `podman run -it --rm --memory=500m --memory-swap=-1 docker.io/library/ubuntu` I get
```
Error: invalid configuration, cannot specify resource limits without cgroups v2 and --cgroup-manager=systemd
```
When I add `--cgroup-manager=systemd` to my command line I get
```
ERRO[0000] could not find slirp4netns, the network namespace won't be configured: exec: "slirp4netns": executable file not found in $PATH 
Error: sd-bus add match: Operation not permitted: OCI runtime permission denied error
```
Adding `--net host` mutes the error for now and I get
```
Error: sd-bus add match: Operation not permitted: OCI runtime permission denied error
```

giuseppe:
> When I add `--cgroup-manager=systemd` to my command line I get

looks like there is no systemd user session that you can use.

How has the rootless user logged in?  Through sudo?

You need to make sure there is a systemd user session, you either need to login through ssh/desktop so that the pam session is created, or you need to use `loginctl enable-linger $UID`

rhatdan:
@mheon @giuseppe @baude @TomSweeneyRedHat @QiWang19 PTAL
This at least gets us closer to the solution.

giuseppe:
LGTM

Tested locally:
```sh
$ podman run --rm --memory=500m --memory-swap=-1 fedora cat /sys/fs/cgroup/memory.swap.max
max
```

s-winter:
> looks like there is no systemd user session that you can use.
> 
> How has the rootless user logged in? Through sudo?

Direct login via ssh.

> You need to make sure there is a systemd user session, you either need to login through ssh/desktop so that the pam session is created, or you need to use `loginctl enable-linger $UID`

As I wrote I directly log in via ssh. I also tried `loginctl enable-linger $UID`, but that does not change anything about the error.

I recall that I had the same issue when I was playing with podman earlier, around mid November. I do not recall what I did to resolve it back then or if I resolved it at all. But I was very happy that everything was running out of the box and without problems when I ran version 1.7.0 from the installed Ubuntu package this week. I am surprised that I'm now facing the issue again when switching to a build from libpod's master branch to test the fix proposed by rhatdan.

PR #5111, opened 2020-02-06
Move podman-service to podman-system-service

### LINKED ISSUE #5108: suggestion: rename `podman service` to avoid conflict with `docker service`
`podman service`  looks as if it was Podman's implementation of `docker service`.

I suggest renaming `podman service` to `podman daemon`, because it seems actually reimplementation of `docker daemon` that had existed in Docker 1.10 and prior versions.

mheon:
Alleviates a conflict with the unrelated `docker service` command

Fixes #5108


mheon:
@rhatdan @baude @vrothberg @jwhonce PTAL

baude:
i want to know more about this before it gets merged

PR #5112, opened 2020-02-06
Cirrus: Never run prune on other branches

cevich:
This is needed because the prune container image will be built from
other branches as they are made.  If the behavior of this or the imgts
image diverges from that of master, random VM images could be "cleaned"
unexpectedly.  By hard-coding this task to the master branch only,
it should never run anywhere else.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #5115, opened 2020-02-06
images --format compatible with docker

### LINKED ISSUE #5110: podman images --format incompatibility with docker
/kind bug

**Description**

The tags available in the `podman images --format` option is incompatible with docker. In particular, docker has both `{{.CreatedAt}}` and `{{.CreatedSince}}` while podman has neither but has `{{.Created}}` instead.

1. `docker images --format '{{.CreatedAt}}'`

2. `podman images --format '{{.CreatedAt}}'`

```
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.12.13
Git Commit:         b7ce1157b00af09f4a09e39b377aa3abff46ee05
Built:              Sun Jan 19 14:08:07 2020
```

```
debug:
  compiler: gc
  git commit: b7ce1157b00af09f4a09e39b377aa3abff46ee05
  go version: go1.12.13
  podman version: 1.7.0
host:
  BuildahVersion: 1.12.0
  CgroupVersion: v1
  Conmon:
    package: Unknown
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.10, commit: 7a830be343876ac381c965c7429a7fb9b3d7a609'
  Distribution:
    distribution: gentoo
    version: unknown
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 500
      size: 1
    - container_id: 1
      host_id: 1000000
      size: 100000
    uidmap:
    - container_id: 0
      host_id: 500
      size: 1
    - container_id: 1
      host_id: 1000000
      size: 100000
  MemFree: 2903642112
  MemTotal: 33604526080
  OCIRuntime:
    name: runc
    package: Unknown
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc9
      spec: 1.0.1-dev
  SwapFree: 2019291136
  SwapTotal: 2147479552
  arch: amd64
  cpus: 12
  eventlogger: journald
  hostname: lorien
  kernel: 5.4.15
  os: linux
  rootless: true
  slirp4netns:
    Executable: /usr/bin/slirp4netns
    Package: Unknown
    Version: |-
      slirp4netns version 0.4.2
      commit: 69153b0d1cb82216d6782179ff7c3c5e91c731a1
  uptime: 214h 10m 6.52s (Approximately 8.92 days)
registries:
  search:
  - docker.io
  - quay.io
store:
  ConfigFile: /home/bruce/.config/containers/storage.conf
  ContainerStore:
    number: 5
  GraphDriverName: btrfs
  GraphOptions: {}
  GraphRoot: /scratch/containers/storage
  GraphStatus:
    Build Version: 'Btrfs v5.4.1 '
    Library Version: "102"
  ImageStore:
    number: 27
  RunRoot: /tmp/500
  VolumePath: /scratch/containers/storage/volumes
```

```
app-emulation/libpod-1.7.0::gentoo  USE="btrfs rootless -apparmor (-selinux) (-ostree%)"
```

QiWang19:
Let valid values of --format be compatible with docker. Replace CreatedTime with CreatedAt, Created with CreatedSince.
fix #5110 

### DIFF
--- a/cmd/podman/images.go
+++ b/cmd/podman/images.go
@@ -21,16 +21,16 @@ import (
 )
 
 type imagesTemplateParams struct {
-	Repository  string
-	Tag         string
-	ID          string
-	Digest      digest.Digest
-	Digests     []digest.Digest
-	Created     string
-	CreatedTime time.Time
-	Size        string
-	ReadOnly    bool
-	History     string
+	Repository   string
+	Tag          string
+	ID           string
+	Digest       digest.Digest
+	Digests      []digest.Digest
+	CreatedAt    time.Time
+	CreatedSince string
+	Size         string
+	ReadOnly     bool
+	History      string
 }
 
 type imagesJSONParams struct {
@@ -65,7 +65,7 @@ func (a imagesSorted) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
 type imagesSortedCreated struct{ imagesSorted }
 
 func (a imagesSortedCreated) Less(i, j int) bool {
-	return a.imagesSorted[i].CreatedTime.After(a.imagesSorted[j].CreatedTime)
+	return a.imagesSorted[i].CreatedAt.After(a.imagesSorted[j].CreatedAt)
 }
 
 type imagesSortedID struct{ imagesSorted }
@@ -185,7 +185,17 @@ func imagesCmd(c *cliconfig.ImagesValues) error {
 		history:   c.History,
 	}
 
-	opts.outputformat = opts.setOutputFormat()
+	outputformat := opts.setOutputFormat()
+	// These fields were renamed, so we need to provide backward compat for
+	// the old names.
+	if strings.Contains(outputformat, "{{.Created}}") {
+		outputformat = strings.Replace(outputformat, "{{.Created}}", "{{.CreatedSince}}", -1)
+	}
+	if strings.Contains(outputformat, "{{.CreatedTime}}") {
+		outputformat = strings.Replace(outputformat, "{{.CreatedTime}}", "{{.CreatedAt}}", -1)
+	}
+	opts.outputformat = outputformat
+
 	filteredImages, err := runtime.GetFilteredImages(filters, false)
 	if err != nil {
 		return errors.Wrapf(err, "unable to get images")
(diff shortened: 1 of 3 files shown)


mheon:
LGTM but would prefer to hold until after 1.8.0 so I don't need to update release notes again

QiWang19:
> Any man page changes? Tests?
> Also can I still use the old values? If I built a script using Created or CreatedTime, will it still work?

We can add man pages like this https://docs.docker.com/engine/reference/commandline/images/#format-the-output
old value `--format {{.Created}}` won't work. Should be `{{.CreatedSince}}` now

TomSweeneyRedHat:
Looks like the same test errors here as in some other PRs.  Houston we may have a CI problem.

cmd/podman/images.go
```diff
@@ -21,16 +21,16 @@ import (
...
+	Repository   string
+	Tag          string
+	ID           string
+	Digest       digest.Digest
+	Digests      []digest.Digest
+	CreatedAt    time.Time
+	CreatedSince string
+	Size         string
+	ReadOnly     bool
+	History      string
```
TomSweeneyRedHat: Just wondering if we should still support "Created" and "CreatedTime" in case people have adjusted and have built dependencies on those?  I'd be fine if they were hidden options going forward, available for use, but not documented.

TomSweeneyRedHat:
definitely double check on the man pages and tests.  If we do drop the old names, we will need a BIG announcement in the 1.8.1 or 1.9.0 Release announcement about that.

cmd/podman/images.go
QiWang19: My concern is that keeping `Created` and `CreatedTime` in this struct for hidden options seems duplicate.

TomSweeneyRedHat: yep, it's certainly a duplicate and I'd add a comment in the code as to why it's there.  My concern is someone out there has already built a dependency on one of those and we might break them if they go away.

mheon: I would not retain duplicates in the structures - especially given how similar all these names are. We may want to look into normalization similar to what we're doing in `podman inspect --format` though.

QiWang19: @TomSweeneyRedHat @mheon Thanks. I will do that.

TomSweeneyRedHat: If Matt's hip, then I'm hip.  Thanks for the discussion!

rhatdan:
LGTM
But not sure what is happening with the remote tests.

PR #5120, opened 2020-02-06
LibpodAPI.BuildImage: don't require a name for the new image

nalind:
When we finish building an image, we try to look up its ID by looking up the image using the name that we were asked to assign to the image.  If we weren't asked to assign a name to the image, that would produce an error.  The BuildImage() API we're using returns the image's ID anyway, so we can skip the lookup and just return the ID directly.


rhatdan:
LGTM

mheon:
LGTM

PR #5122, opened 2020-02-07
Simplify image object creation

mtrmac:
This is a refactoring that should not change behavior.  Originally created primarily so that both `image.Runtime` and `image.Image` have a single constructor where I can hook the WIP new API, this:
- Simplifies creation of `Image` by name, so that each caller doesn’t have to create it manually, call `getLocalImage` which modifies one member, and then modify another member manually
- Modifies `Runtime.getImage` to return a `storage.Image` instead of wrapping it in an `image.Image` that every caller discards.
- Removes dead code in `Image.Size`

See the individual commit messages for details.

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -120,7 +120,7 @@ func (ir *Runtime) newImage(inputName string, img *storage.Image) *Image {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


baude:
LGTM ... thanks @mtrmac

PR #5125, opened 2020-02-07
[1.0] update containers/image

vrothberg:
Fixes for https://bugzilla.redhat.com/show_bug.cgi?id=1792796

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -11,7 +11,7 @@ github.com/containerd/cgroups 58556f5ad8448d99a6f7bea69ea4bdb7747cfeb0
 github.com/containerd/continuity master
 github.com/containernetworking/cni v0.7.0-alpha1
 github.com/containernetworking/plugins 1562a1e60ed101aacc5e08ed9dbeba8e9f3d4ec1
-github.com/containers/image 93bced01015eb94bec4821df1876314be8197680
+github.com/containers/image podman-1.0
 github.com/containers/storage backport_symlinks https://github.com/mheon/storage.git
 github.com/containers/psgo dc0bc9fac5b715034c4310ed4d795b3182360842
 github.com/coreos/go-systemd v14
--- a/vendor/github.com/containers/image/docker/docker_client.go
+++ b/vendor/github.com/containers/image/docker/docker_client.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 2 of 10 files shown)


vrothberg:
@mheon PTAL

vendor.conf
```diff
@@ -11,7 +11,7 @@ github.com/containerd/cgroups 58556f5ad8448d99a6f7bea69ea4bdb7747cfeb0
 github.com/containerd/continuity master
 github.com/containernetworking/cni v0.7.0-alpha1
 github.com/containernetworking/plugins 1562a1e60ed101aacc5e08ed9dbeba8e9f3d4ec1
-github.com/containers/image 93bced01015eb94bec4821df1876314be8197680
```
TomSweeneyRedHat: Even if it is correct, this seems like an odd version.  How does it tie into c/image?  This is likely time to further Tom's education in vendoring.
Beyond which you're fighting test issues.

vrothberg: I created a `podman-1.0` branch which has commit `93bced01015eb94bec4821df1876314be8197680` as the base with the backported fixes for the CVE on top.

I think that's a text book example for how bad our release management was prior to go modules as the tools were oftentimes not vendoring in releases but commits or sometimes just `master` which has no real value for maintenance tasks (e.g., bug fixes).

The RHEL update policies are very strict, so I cannot change the base (i.e., commit `93bced01015eb94bec4821df1876314be8197680`) to a proper release, so I created a dedicated branch in c/image for `podman-1.0` as we have one for `podman-1.4.2`. We could argue for vendoring the last commit in the `podman-1.0` branch in c/image but I prefer to keep the reference on the branch to make it more explicit which branch we were using (for future generations).

With go modules, it's much easier as we became more disciplined with release management and vendoring, so we can just have a `5.0-stable` branch for more recent versions of podman, buildah, etc.

>  Beyond which you're fighting test issues.

The CI doesn't start in this branch at all, so I hope that's not my fight :^)

rhatdan:
LGTM, but tests are failing.

vrothberg:
> LGTM, but tests are failing.

The CI doesn't start in this branch at all, so I hope that's not my fight :^)

rhatdan:
@vrothberg Should we just force merge then?

vrothberg:
> @vrothberg Should we just force merge then?

@mheon WDYT?

vendor.conf
TomSweeneyRedHat: Thank you.  I missed the podman-1.0 branch, now the tag in here makes sense.

mheon: I'm honestly amazed you got this to apply without Libpod changes. Bravo!

vrothberg: Need more cookies from @baude :^)

mheon: I didn't realize we were shipping a c/image this new in 1.0.0, actually. I was thinking this was older, and substantially more monkey-patched. That might be c/storage, now that I think about it...

mheon:
@vrothberg From what I understood from Chris, this isn't likely to start working soon... Given that, if we can manually do basic compile and functionality checks on RHEL8 for this branch and the 1.4 one, I think we're good to merge.

vrothberg:
Can we get this merged?

PR #5132, opened 2020-02-07
Add test cases to validate remove and list images api.

sujil02:
Add test cases to validate remove and list images api.

Includes test case to validate list image api count as we create and delete images.
Include test case to validate remove image api responses with container instance
and bogus images etc.

### DIFF
--- a/pkg/bindings/test/common_test.go
+++ b/pkg/bindings/test/common_test.go
@@ -16,6 +16,7 @@ import (
 const (
 	defaultPodmanBinaryLocation string = "/usr/bin/podman"
 	alpine                      string = "docker.io/library/alpine:latest"
+	busybox                     string = "docker.io/library/busybox:latest"
 )
 
 type bindingTest struct {
@@ -113,7 +114,38 @@ func (b *bindingTest) startAPIService() *gexec.Session {
 }
 
 func (b *bindingTest) cleanup() {
+	s := b.runPodman([]string{"stop", "-a", "-t", "0"})
+	s.Wait(45)
 	if err := os.RemoveAll(b.tempDirPath); err != nil {
 		fmt.Println(err)
 	}
+}
+
+// Pull is a helper function to pull in images
+func (b *bindingTest) Pull(name string) {
+	p := b.runPodman([]string{"pull", name})
+	p.Wait(45)
+}
+
+// Run a container and add append the alpine image to it
+func (b *bindingTest) RunTopContainer(name *string) {
+	cmd := []string{"run", "-dt"}
+	if name != nil {
+		containerName := *name
+		cmd = append(cmd, "--name", containerName)
+	}
+	cmd = append(cmd, alpine, "top")
+	p := b.runPodman(cmd)
+	p.Wait(45)
+}
+
+//  StringInSlice returns a boolean based on whether a given
+//  string is in a given slice
+func StringInSlice(s string, sl []string) bool {
+	for _, val := range sl {
+		if s == val {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/bindings/test/images_test.go
+++ b/pkg/bindings/test/images_test.go
@@ -2,10 +2,10 @@ package test_bindings
 
 import (
 	"context"
-	"fmt"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/bindings/test/common_test.go
```diff
@@ -113,7 +114,38 @@ func (b *bindingTest) startAPIService() *gexec.Session {
 }
 
 func (b *bindingTest) cleanup() {
+	s := b.runPodman([]string{"stop", "-a", "-t", "0"})
+	s.Wait(45)
```
rhatdan: Seems like a long time to wait?

pkg/bindings/test/common_test.go
```diff
@@ -113,7 +114,38 @@ func (b *bindingTest) startAPIService() *gexec.Session {
...
+// Run a container and add append the alpine image to it
+func (b *bindingTest) RunTopContainer(name *string) {
+	cmd := []string{"run", "-dt"}
+	if name != nil {
+		containerName := *name
+		cmd = append(cmd, "--name", containerName)
+	}
+	cmd = append(cmd, alpine, "top")
+	p := b.runPodman(cmd)
+	p.Wait(45)
```
rhatdan: What are you waiting for?

pkg/bindings/test/images_test.go
```diff
@@ -68,28 +69,62 @@ var _ = Describe("Podman images", func() {
 		_, err = images.GetImage(connText, data.ID[0:12], nil)
 		Expect(err).To(BeNil())
 
-		//Inspect by long name should work, it doesnt (yet) i think it needs to be html escaped
+		// Inspect by long name should work, it doesnt (yet) i think it needs to be html escaped
 		_, err = images.GetImage(connText, alpine, nil)
-		Expect(err).To(BeNil())
+		//Expect(err).To(BeNil())
```
rhatdan: Space between // and Expect?

pkg/bindings/test/images_test.go
```diff
@@ -68,28 +69,62 @@ var _ = Describe("Podman images", func() {
...
 	})
+
+	// Test to validate the remove image api
 	It("remove image", func() {
 		// Remove invalid image should be a 404
 		_, err = images.Remove(connText, "foobar5000", &false)
-		Expect(err).ToNot(BeNil())
```
rhatdan: Why is this removed?

rhatdan: Should the images.Remove above use falseFlag?

pkg/bindings/test/images_test.go
```diff
@@ -68,28 +69,62 @@ var _ = Describe("Podman images", func() {
...
+		Expect(inspectData.ID).To(Equal(response[0]["Deleted"]))
+		inspectData, err = images.GetImage(connText, "busybox", nil)
+		code, _ = bindings.CheckResponseCode(err)
+		Expect(code).To(BeNumerically("==", 404))
+
+		// Start a container with alpine image
+		var top string = "top"
+		bt.RunTopContainer(&top)
+		// we should now have a container called "top" running
+		containerResponse, err := containers.Inspect(connText, "top", &falseFlag)
```
rhatdan: What are these falseFlag and TrueFlags for?

pkg/bindings/test/common_test.go
baude: It wont wait that long if it completes.  It will wait at least that long if it does not.

pkg/bindings/test/images_test.go
baude: remove

pkg/bindings/test/images_test.go
baude: size

baude: because the binding takes a pointer because the param is optional.

pkg/bindings/test/images_test.go
umohnani8: Maybe remove this part of the test and a leave a comment that says we need to add a test for inspecting by long name once it works.

pkg/bindings/test/images_test.go
sujil02: Fixed.

PR #5149, opened 2020-02-10
Bump to v1.6.5

mheon:
This is on the v1.6 long-term support branch; v1.8.0 will remain the most recent master release, and this is not intended to replace it.

With that said:
Bump to v1.6.5, including a new version of c/image with necessary fixes.


rhatdan:
LGTM

PR #5159, opened 2020-02-10
Rewire ListContainers for APIv2 libpod

baude:
consumers of the api remarked how they would prefer a more strongly typed data structure from list containers oon the libpod side of things.  for example, events should be consumable and consistent timestamps.  also, for the sake of compatibility, it is helpful to have the json named atttributes for Id to not be ID.

listcontainers on the libpod side no longer strongly uses the the ps cli to obtain information but we do benefit from turning on the ability to list the last X containers, something CLI does not have yet. we also flipped the bit on defaulting to truncated output in the return.

thanks to the efforts of the cockpit team to help us here.

### DIFF
--- a/pkg/api/handlers/libpod/containers.go
+++ b/pkg/api/handlers/libpod/containers.go
@@ -1,16 +1,20 @@
 package libpod
 
 import (
-	"fmt"
 	"net/http"
+	"path/filepath"
+	"sort"
 	"strconv"
+	"time"
 
 	"github.com/containers/libpod/cmd/podman/shared"
 	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/pkg/api/handlers"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 func StopContainer(w http.ResponseWriter, r *http.Request) {
@@ -46,7 +50,8 @@ func RemoveContainer(w http.ResponseWriter, r *http.Request) {
 }
 func ListContainers(w http.ResponseWriter, r *http.Request) {
 	var (
-		filters []string
+		filterFuncs []libpod.ContainerFilter
+		pss         []ListContainer
 	)
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
@@ -73,19 +78,54 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 		Size:      query.Size,
 		Sort:      "",
 		Namespace: query.Namespace,
+		NoTrunc:   true,
 		Pod:       query.Pod,
 		Sync:      query.Sync,
 	}
 	if len(query.Filter) > 0 {
 		for k, v := range query.Filter {
 			for _, val := range v {
-				filters = append(filters, fmt.Sprintf("%s=%s", k, val))
+				generatedFunc, err := shared.GenerateContainerFilterFuncs(k, val, runtime)
+				if err != nil {
+					utils.InternalServerError(w, err)
+					return
+				}
+				filterFuncs = append(filterFuncs, generatedFunc)
 			}
 		}
 	}
-	pss, err := shared.GetPsContainerOutput(runtime, opts, filters, 2)
+
+	if !query.All {
+		// The default is get only running containers. Do this with a filterfunc
+		runningOnly, err := shared.GenerateContainerFilterFuncs("status", define.ContainerStateRunning.String(), runtime)
+		if err != nil {
+			utils.InternalServerError(w, err)
+			return
+		}
+		filterFuncs = append(filterFuncs, runningOnly)
+	}
+
+	cons, err := runtime.GetContainers(filterFuncs...)
 	if err != nil {
 		utils.InternalServerError(w, err)
+	}
+	if query.Last > 0 {
+		// Sort the containers we got
+		sort.Sort(psSortCreateTime{cons})
+		// we should perform the lopping before we start getting
+		// the expensive information on containers
+		if query.Last < len(cons) {
+			cons = cons[len(cons)-query.Last:]
+		}
+	}
+	for _, con := range cons {
+		listCon, err := ListContainerBatch(runtime, con, opts)
+		if err != nil {
+			utils.InternalServerError(w, err)
+			return
+		}
+		pss = append(pss, listCon)
+
 	}
 	utils.WriteResponse(w, http.StatusOK, pss)
 }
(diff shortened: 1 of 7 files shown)


baude:
@marusak the rework was greater than initially expected, but I think this is more inline with what you wanted.

```
[
  {
    "Command": [
      "top"
    ],
    "Created": 1581364244411362000,
    "Exited": false,
    "ExitedAt": -6795364578871345000,
    "ExitCode": 0,
    "Id": "9f5c2c029147593d8d149efe27b330f3291b576e05c712caae6bef446b2bb76d",
    "Image": "docker.io/library/alpine:latest",
    "IsInfra": false,
    "Labels": null,
    "Mounts": [],
    "Names": [
      "musing_colden"
    ],
    "Namespaces": {},
    "Pid": 0,
    "Pod": "",
    "PodName": "",
    "Ports": [
      {
        "hostPort": 80,
        "containerPort": 8001,
        "protocol": "tcp",
        "hostIP": ""
      }
    ],
    "Size": null,
    "StartedAt": 1581364244960983000,
    "State": "running"
  }
]
```

pkg/api/handlers/libpod/containers.go
```diff
@@ -194,3 +234,122 @@ func ShowMountedContainers(w http.ResponseWriter, r *http.Request) {
...
+	)
+
+	batchErr := ctr.Batch(func(c *libpod.Container) error {
+		conConfig = c.Config()
+		conState, err = c.State()
+		if err != nil {
+			return errors.Wrapf(err, "unable to obtain container state")
+		}
+
+		exitCode, exited, err = c.ExitCode()
```
mheon: This is part of the state, if we're retrieving c.State there's no need to get this separately

pkg/api/handlers/libpod/containers.go
```diff
@@ -194,3 +234,122 @@ func ShowMountedContainers(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			return errors.Wrapf(err, "unable to obtain container exit code")
+		}
+		startedTime, err = c.StartedTime()
```
mheon: Same as above

pkg/api/handlers/libpod/containers.go
```diff
@@ -194,3 +234,122 @@ func ShowMountedContainers(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			logrus.Errorf("error getting started time for %q: %v", c.ID(), err)
+		}
+		exitedTime, err = c.FinishedTime()
```
mheon: Same as above

pkg/api/handlers/libpod/containers.go
mheon: conState.ExitCode

pkg/api/handlers/libpod/containers.go
mheon: conState.StartedTime

pkg/api/handlers/libpod/containers.go
mheon: conState.FinishedTime

pkg/api/handlers/libpod/containers.go
mheon: Oh, nevermind, we're only getting the individual state fields, not the entire state... I guess this is what happens when we have c.State() and c.state being entirely different things

pkg/api/handlers/libpod/containers.go
```diff
@@ -194,3 +234,122 @@ func ShowMountedContainers(w http.ResponseWriter, r *http.Request) {
...
+		Names:     []string{conConfig.Name},
+		Pid:       pid,
+		Pod:       conConfig.Pod,
+		Ports:     conConfig.PortMappings,
+		Size:      size,
+		StartedAt: startedTime.UnixNano(),
+		State:     conState.String(),
+	}
+	if opts.Pod && len(conConfig.Pod) > 0 {
+		pod, err := rt.LookupPod(conConfig.Pod)
```
mheon: No need to do a Lookup here, GetPod is faster and we are guaranteed to have a full ID

pkg/api/handlers/libpod/containers.go
```diff
@@ -194,3 +234,122 @@ func ShowMountedContainers(w http.ResponseWriter, r *http.Request) {
...
+		Mounts:    ctr.UserVolumes(),
+		Names:     []string{conConfig.Name},
+		Pid:       pid,
+		Pod:       conConfig.Pod,
+		Ports:     conConfig.PortMappings,
+		Size:      size,
+		StartedAt: startedTime.UnixNano(),
+		State:     conState.String(),
+	}
+	if opts.Pod && len(conConfig.Pod) > 0 {
```
mheon: This is a little unclear - makes it seem like an array? `configConfig.Pod != ""` is a little better

pkg/api/handlers/libpod/types.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"github.com/containers/libpod/cmd/podman/shared"
+	"github.com/containers/libpod/libpod"
+	"github.com/cri-o/ocicni/pkg/ocicni"
+)
+
+// Listcontainer describes a container suitable for listing
+type ListContainer struct {
+	// Container command
+	Command []string
```
mheon: Is the lack of JSON tags deliberate?

marusak:
Thanks @baude for working on this!
The current state works for me. It still has bunch of differences between `libpod/containers/json` and `containers/json` but I guess I misunderstood the goal of having those two as similar as possible unless some difference is needed. But I plan to use `libpod` so this is no problem for us, only if once someone would want to migrate. (But I like the current state more than docker one)

Before this PR:
```
Cgroup: ""                                                                      
Command: "sh -c for i in $(..."                                                 
Created: "3 weeks ago"                                                          
CreatedAt: "2020-01-19T14:15:29.110814596+01:00"                                
ExitedAt: "2020-01-19T14:15:49.629692216+01:00"                                 
ID: "0e7c5942cdb2"                                                              
IPC: ""                                                                         
Image: "docker.io/library/fedora:latest"                                        
IsInfra: false                                                                  
Labels: Object { maintainer: "Clement Verna <cverna@fedoraproject.org>" }       
MNT: ""                                                                         
Mounts: ""                                                                      
NET: ""                                                                         
Names: "trusting_dubinsky"                                                      
PID: ""                                                                         
PIDNS: ""                                                                       
Pid: 0                                                                          
Pod: ""                                                                         
PodName: ""                                                                     
Ports: ""                                                                       
Size: null                                                                      
StartedAt: "2020-01-19T14:15:29.561671656+01:00"                                
State: 6                                                                        
Status: "Exited (0) 3 weeks ago"                                                
UTS: ""                                                                         
User: ""    
```

now:
```
Command: Array(3) [ "sh", "-c", "for i in $(seq 20); do echo $i; sleep 1; done" ]
Created: 1579439729110814700                                                    
ExitCode: 0                                                                     
Exited: true                                                                    
ExitedAt: 1579439749629692200                                                   
Id: "0e7c5942cdb2e8cbff709f6e6b8b71d001bf3508e4eaaba94a8aa46568f63304"          
Image: "docker.io/library/fedora:latest"                                        
IsInfra: false                                                                  
Labels: Object { maintainer: "Clement Verna <cverna@fedoraproject.org>" }       
Mounts: Array []                                                                
Names: Array [ "trusting_dubinsky" ]                                            
Namespaces: Object {  }                                                         
Pid: 0                                                                          
Pod: ""                                                                         
PodName: ""                                                                     
Ports: null                                                                     
Size: null                                                                      
StartedAt: 1579439729561671700                                                  
State: "exited"    
```

and `containers/json`:
```
AdjustCPUShares: false                                                          
Command: "sh -c for i in $(seq 20); do echo $i; sleep 1; done"                  
Config: null                                                                    
Created: 1579439729                                                             
Id: "0e7c5942cdb2e8cbff709f6e6b8b71d001bf3508e4eaaba94a8aa46568f63304"          
Image: "docker.io/library/fedora:latest"                                        
ImageID: "f0858ad3febdf45bb2e5501cb459affffacef081f79eaa436085c3b6d9bd46ca"     
Labels: Object { maintainer: "Clement Verna <cverna@fedoraproject.org>" }       
Mounts: null                                                                    
Name: ""                                                                        
Names: Array [ "/trusting_dubinsky" ]                                           
NetworkSettings: null                                                           
NetworkingConfig: null                                                          
Ports: null                                                                     
SizeRootFs: 194281245                                                           
State: "\u0006"                                                                 
Status: ""   
```

marusak:
.One more comment: It returns   `Created: 1579439729110814700` which is in nanoseconds. 
`/libpod/images/json` returns `Created: 1581423166` which is in seconds.
If things would be bit more consistent that would be great :)

pkg/api/handlers/libpod/containers.go
baude: Im told the compiler treats them both in the same way.  For readability, i like len() ... just personal preference i think.

mheon: Fair enough

baude:
all fixed up

mheon:
LGTM

marusak:
Thanks, these changes work for Cockpit!

pkg/api/handlers/libpod/types.go
```diff
@@ -0,0 +1,82 @@
...
+	Created int64
+	// If container has exited/stopped
+	Exited bool
+	// Time container exited
+	ExitedAt int64
+	// If container has exited, the return code from the command
+	ExitCode int32
+	// The unique identifier for the container
+	ID string `json:"Id"`
+	// Container image
```
TomSweeneyRedHat: Do you expect the name or the ID to be in here?

TomSweeneyRedHat:
LGTM once the tests get happy, they're not atm.
One minor nit of a nit in a comment that could be chased later.

PR #5168, opened 2020-02-11
Do not copy up when volume is not empty

mheon:
When Docker performs a copy up, it first verifies that the volume being copied into is empty; thus, for volumes that have been modified elsewhere (e.g. manually copying into then), the copy up will not be performed at all. Duplicate this behavior in Podman by checking if the volume is empty before copying.

Furthermore, move setting copyup to false further up. This will prevent a potential race where copy up could happen more than once if Podman was killed after some files had been copied but before the DB was updated.

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1383,17 +1383,33 @@ func (c *Container) mountNamedVolume(v *ContainerNamedVolume, mountpoint string)
 	}
 	if vol.state.NeedsCopyUp {
 		logrus.Debugf("Copying up contents from container %s to volume %s", c.ID(), vol.Name())
+
+		// Set NeedsCopyUp to false immediately, so we don't try this
+		// again when there are already files copied.
+		vol.state.NeedsCopyUp = false
+		if err := vol.save(); err != nil {
+			return nil, err
+		}
+
+		// If the volume is not empty, we should not copy up.
+		volMount := vol.MountPoint()
+		contents, err := ioutil.ReadDir(volMount)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error listing contents of volume %s mountpoint when copying up from container %s", vol.Name(), c.ID())
+		}
+		if len(contents) > 0 {
+			// The volume is not empty. It was likely modified
+			// outside of Podman. For safety, let's not copy up into
+			// it. Fixes CVE-2020-1726.
+			return vol, nil
+		}
+
 		srcDir, err := securejoin.SecureJoin(mountpoint, v.Dest)
 		if err != nil {
 			return nil, errors.Wrapf(err, "error calculating destination path to copy up container %s volume %s", c.ID(), vol.Name())
 		}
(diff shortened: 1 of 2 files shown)


mheon:
@vrothberg @rhatdan @baude @giuseppe @TomSweeneyRedHat PTAL

libpod/container_internal.go
```diff
@@ -1383,18 +1383,34 @@ func (c *Container) mountNamedVolume(v *ContainerNamedVolume, mountpoint string)
...
+		// If the volume is not empty, we should not copy up.
+		volMount := vol.MountPoint()
+		contents, err := ioutil.ReadDir(volMount)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error listing contents of volume %s mountpoint when copying up from container %s", vol.Name(), c.ID())
+		}
+		if len(contents) > 0 {
+			// The volume is not empty. It was likely modified
+			// outside of Podman. For safety, let's not copy up into
+			// it.
```
vrothberg: Maybe worth dropping the CVE in the comment (for historical reasons).

mheon: Fair point

libpod/container_internal.go
```diff
@@ -1383,18 +1383,34 @@ func (c *Container) mountNamedVolume(v *ContainerNamedVolume, mountpoint string)
...
+			return nil, err
+		}
+
+		// If the volume is not empty, we should not copy up.
+		volMount := vol.MountPoint()
+		contents, err := ioutil.ReadDir(volMount)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error listing contents of volume %s mountpoint when copying up from container %s", vol.Name(), c.ID())
+		}
+		if len(contents) > 0 {
```
rhatdan: I take it contents does not include "." and ".."

mheon: Yes, those are excluded

TristanCacqueray:
I confirm dd65ad5a fixes the issue. Here is a reproducer:

```
# step1: create the image
CTX=$(buildah from registry.fedoraproject.org/fedora:31)
buildah run $CTX sh -c 'mkdir /data; echo oops > /data/secret.txt'
buildah commit --rm $CTX fedora-data

# step2: create the volume
podman volume create test
echo readonly-file > ~/.local/share/containers/storage/volumes/test/_data/secret.txt

# step3: trigger issue
podman run -it --rm --volume test:/data:ro fedora-data id
cat ~/.local/share/containers/storage/volumes/test/_data/secret.txt
```

The last stdout line should be `readonly-file`, not `oops`

mheon:
Added a test

vrothberg:
gating isn't happy yet

mheon:
Tests seem to be coming back as flakes, not failures, this time. Might be ready with some retesting.

mheon:
It's green. @rhatdan @baude @vrothberg @giuseppe @TomSweeneyRedHat PTAL and merge

mheon:
Nevermind, need to fix one thing in tests

mheon:
Re-pushed, should be good now. Held LGTM would be appreciated.

PR #5169, opened 2020-02-11
API v2: pods: fix two incorrect return codes

edsantiago:
1) /pods/\<X\>/exists - is documented to return 204, and that's
    the correct value, but until now it has been returning 200.

 2) /pods/create - return 409 (conflict), not 500, when pod
    already exists

Also: in WriteResponse(), if code is 204 (No Content), emit
the status code only but no content-type headers nor content.

### DIFF
--- a/pkg/api/handlers/libpod/pods.go
+++ b/pkg/api/handlers/libpod/pods.go
@@ -91,7 +91,11 @@ func PodCreate(w http.ResponseWriter, r *http.Request) {
 
 	pod, err := runtime.NewPod(r.Context(), options...)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


mheon:
/approve
Changes LGTM

edsantiago:
Pushed update: only change is in `handler.go`, using a `switch` statement to handle both 204 and 304 as defined in rfc2616. Also as a way to rerun flaking tests.

edsantiago:
(one more push: fix a badly-formatted comment I added)

edsantiago:
@jwhonce @baude if you can spare a moment, I'd appreciate a quick look. I need this to converge on a working set of APIv2 tests that we can enable in CI.

PR #5171, opened 2020-02-11
Fix container filters

baude:
container filters were being double encoded (maybe triple) which resulted in the wrong encoding representation of filters being sent by the go-bindings.  Also, on the server side, Filter needed to be changed to Filter to decode properly. Finally, due to the changed return type of List Containers, the go bindings return values needed to be changed.

### DIFF
--- a/pkg/api/handlers/libpod/containers.go
+++ b/pkg/api/handlers/libpod/containers.go
@@ -56,7 +56,7 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
 		All       bool                `schema:"all"`
-		Filter    map[string][]string `schema:"filter"`
+		Filters   map[string][]string `schema:"filters"`
 		Last      int                 `schema:"last"`
 		Namespace bool                `schema:"namespace"`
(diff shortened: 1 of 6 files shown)


baude:
@marusak ptal

rhatdan:
LGTM

marusak:
Ok, I tested this and now it parses the filters, which is awesome!
However it is still a bit difficult to use:
`libpod/containers/json {all: true, filter: "id=<container_id>"}` returns all containers, ignoring the filter. Fair enough, even though doc says `Return all containers. By default, only running containers are shown` which implies to me, it is there to just be able to query all, not just running containers.

So when I drop `all:true`:
`libpod/containers/json {filter: "id=<container_id>"}` and the container is running, it returns only this one. __However__ when the container is not running, I get just `null`. I know there is also `status=(created|restarting|running|removing|paused|exited|dead)` but I would not prefer to pass this whole thing. (not sure even if possible to declare more statuses at once).

I think it should work like this:
```py
containers = allRuningContainers() # Start with just running containers
if all:
    containers = allContainers() # If all passed, assume all containers
if filter:
    containers = containers.filter(filter) # Then apply filters on containers that you have prepared
return containers
```

PR #5172, opened 2020-02-11
api: fix the CPU stats reported

giuseppe:

### DIFF
--- a/pkg/api/handlers/generic/containers_stats.go
+++ b/pkg/api/handlers/generic/containers_stats.go
@@ -67,9 +67,9 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 		preCPUStats = docker.CPUStats{
 			CPUUsage: docker.CPUUsage{
 				TotalUsage:        stats.CPUNano,
-				PercpuUsage:       []uint64{uint64(stats.CPU)},
-				UsageInKernelmode: 0,
-				UsageInUsermode:   0,
+				PercpuUsage:       stats.PerCPU,
+				UsageInKernelmode: stats.CPUSystemNano,
+				UsageInUsermode:   stats.CPUNano - stats.CPUSystemNano,
 			},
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM

marusak:
It seems not to report `percpu_usage` anymore (which is probably correct as I am on CGroupsV2).
But it still does not show `system_cpu_usage` without which it is not possible to get any useful data from this.

pkg/api/handlers/generic/containers_stats.go
```diff
@@ -67,9 +67,9 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
...
 			CPUUsage: docker.CPUUsage{
 				TotalUsage:        stats.CPUNano,
-				PercpuUsage:       []uint64{uint64(stats.CPU)},
-				UsageInKernelmode: 0,
-				UsageInUsermode:   0,
+				PercpuUsage:       stats.PerCPU,
+				UsageInKernelmode: stats.CPUSystemNano,
+				UsageInUsermode:   stats.CPUNano - stats.CPUSystemNano,
 			},
 			SystemUsage:    0,
```
marusak: I think this is what I crave for.

giuseppe: added a new patch to address it

TomSweeneyRedHat:
LGTM but it looks like our CI is still fighting.

giuseppe:
is `rdoproject.org/github-check` blocking?

Given how often it fails, should we make it optional?

rhatdan:
It is not blocking. Although I wish someone from RDO would look into what is failing?

rhatdan:
@ssbarnea PTAL

ssbarnea:
You can safely ignore the RHEL8-packaging issue, is unrelated and we already made it non-voting, until someone is fixing it.

PR #5175, opened 2020-02-11
apiv2 stream events

baude:
the events endpoint should be stream-based.  it also needed to be registered to answer and not produce 404s.

### DIFF
--- a/pkg/api/handlers/events.go
+++ b/pkg/api/handlers/events.go
@@ -1,19 +1,24 @@
 package handlers
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
-	"strings"
-	"time"
 
+	"github.com/containers/libpod/libpod/events"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 func GetEvents(w http.ResponseWriter, r *http.Request) {
+	var (
+		fromStart   bool
+		eventsError error
+	)
 	query := struct {
-		Since   time.Time           `schema:"since"`
-		Until   time.Time           `schema:"until"`
+		Since   string              `schema:"since"`
+		Until   string              `schema:"until"`
 		Filters map[string][]string `schema:"filters"`
 	}{}
 	if err := decodeQuery(r, &query); err != nil {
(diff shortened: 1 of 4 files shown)


baude:
@jwhonce @marusak

marusak:
I tested this from cockpit-podman and can confirm it fixes the 404 and it also correctly streams events. Thank you!

pkg/api/handlers/events.go
```diff
@@ -27,15 +32,30 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
...
 	}
-
-	var apiEvents = make([]*Event, len(libpodEvents))
-	for _, v := range libpodEvents {
-		apiEvents = append(apiEvents, EventToApiEvent(v))
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	for event := range eventChannel {
+		e := EventToApiEvent(event)
+		//utils.WriteJSON(w, http.StatusOK, e)
```
vrothberg: Nit: delete the line. Non-blocking to safe some energy.

pkg/api/handlers/events.go
```diff
@@ -27,15 +32,30 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
...
+		coder := json.NewEncoder(w)
+		coder.SetEscapeHTML(true)
+		if err := coder.Encode(e); err != nil {
+			logrus.Errorf("unable to write json: %q", err)
```
marusak: I am hitting this.
```
ERRO[4557] unable to write json: "write unix /run/podman/podman.sock->@: i/o timeout"
```
Running `sudo podman --log-level=debug system service -t 0` for some time (When trying to reproduce a few minutes is enough), I can still communicate through the API just fine (so the socket is still active), but I was not getting events anymore and then noticed that on every event this message is printed.

Stopping `podman system service`, starting it back up and then connecting again from cockpit-podman works fine. But it should not timeout. (I am not quite sure what is timing out, the socket obviously not, maybe my connect, or this streamer).
Have you tried running it for longer and listening to the stream?

vrothberg: Thanks for the report. I'll take a look.

vrothberg: I can reproduce locally as well.

vrothberg: This happens for other handlers (e.g., stats) as well, so it's unrelated to this PR.

vrothberg: @baude @jwhonce, have you seen that before?

PR #5178, opened 2020-02-12
stats: Expose CPU usage in API

marusak:
`stats.CPU` is a number representing CPU usage in percents. It is possible
to calculate it from other exposed fields (once they are implemented
correctly) but why to do it, when this is already calculated.

This does not break API, it just adds another field.

@baude WDYT? Is it OK to add another field like this?

### DIFF
--- a/pkg/api/handlers/generic/containers_stats.go
+++ b/pkg/api/handlers/generic/containers_stats.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/containers/libpod/libpod"
 	"github.com/containers/libpod/libpod/define"
-	"github.com/containers/libpod/pkg/api/handlers"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/containers/libpod/pkg/cgroups"
 	docker "github.com/docker/docker/api/types"
@@ -58,17 +57,18 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var preRead time.Time
-	var preCPUStats docker.CPUStats
+	var preCPUStats CPUStats
 	if query.Stream {
 		preRead = time.Now()
 		systemUsage, _ := cgroups.GetSystemCPUUsage()
-		preCPUStats = docker.CPUStats{
+		preCPUStats = CPUStats{
 			CPUUsage: docker.CPUUsage{
 				TotalUsage:        stats.CPUNano,
 				PercpuUsage:       stats.PerCPU,
 				UsageInKernelmode: stats.CPUSystemNano,
 				UsageInUsermode:   stats.CPUNano - stats.CPUSystemNano,
 			},
+			CPU:            stats.CPU,
 			SystemUsage:    systemUsage,
 			OnlineCPUs:     0,
 			ThrottlingData: docker.ThrottlingData{},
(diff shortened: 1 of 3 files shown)


pkg/api/handlers/generic/containers_stats.go
```diff
@@ -151,6 +151,7 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 						UsageInUsermode:   cgroupStat.CPU.Usage.Total - cgroupStat.CPU.Usage.Kernel,
 					},
 					SystemUsage: 0,
+					CPU: stats.CPU,
```
giuseppe: do we need something similar at line 75?

marusak: I don't think so. Those are previous stats, and I cannot think of usecase where it would make sense have it.

vendor/github.com/docker/docker/api/types/stats.go
```diff
@@ -50,6 +50,9 @@ type CPUStats struct {
 	// Online CPUs. Linux only.
 	OnlineCPUs uint32 `json:"online_cpus,omitempty"`
 
+	// Usage of CPU in %. Linux only.
+	CPU float64 `json:"cpu"`
```
mheon: We can't update vendor/ - these are basically shared library files that are automatically updated with the library in question. If we need to change this we need to bring it into our code.

marusak: right. Is there something similar I can get inspiration from?

mheon: @baude Where should we keep structs like this?

@marusak Basically we just want to copy the struct into our codebase, with a note as to where it came from. Not sure exactly where it should end up yet, though.

marusak: So I was looking into that a bit and I think I found some solution. If it is the right or best solution I cannot tell. Pushed it here for discussion. @baude WDYT?

pkg/api/handlers/generic/types.go
```diff
@@ -0,0 +1,55 @@
...
+	CPUUsage docker.CPUUsage `json:"cpu_usage"`
+
+	// System Usage. Linux only.
+	SystemUsage uint64 `json:"system_cpu_usage,omitempty"`
+
+	// Online CPUs. Linux only.
+	OnlineCPUs uint32 `json:"online_cpus,omitempty"`
+
+	// Usage of CPU in %. Linux only.
+	CPU float64 `json:"cpu"`
```
mheon: I don't really like the name here. Maybe `TotalCPU`?

mheon: `TotalCPUPercent` alternatively?

mheon:
Looks like a few lint errors

pkg/api/handlers/generic/types.go
marusak: I went with the same name that is used in `libpod/stats_config.go` for the same thing. (I assign it from that structure from field `CPU`). But surely, I can change it.

marusak:
> Looks like a few lint errors

Does [this](https://github.com/containers/libpod/pull/5178/checks?check_run_id=444406733) show me, what it actually expects? Or am I reading it wrongly?

mheon:
Yes - the `gofmt` command can also be used to directly fix the files (`-s -w` I believe)

baude:
@jwhonce are you ok with this?

jwhonce:
LGTM

PR #5183, opened 2020-02-12
rootlessport: fix potential hang

giuseppe:
write to the error pipe only in case of an error.  Otherwise we may
end up in a race condition in the select statement below as the read
from errChan happens before initComplete and the function returns
immediately nil.

Closes: https://github.com/containers/libpod/issues/5182

### DIFF
--- a/pkg/rootlessport/rootlessport_linux.go
+++ b/pkg/rootlessport/rootlessport_linux.go
@@ -149,9 +149,6 @@ func parent() error {
 	cmd.Stdout = &logrusWriter{prefix: "child"}
 	cmd.Stderr = cmd.Stdout
 	cmd.Env = append(os.Environ(), reexecChildEnvOpaque+"="+string(opaqueJSON))
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Pdeathsig: syscall.SIGTERM,
-	}
 	childNS, err := ns.GetNS(cfg.NetNSPath)
 	if err != nil {
 		return err
@@ -162,6 +159,12 @@ func parent() error {
 	}); err != nil {
 		return err
 	}
+
+	defer func() {
+		if err := syscall.Kill(cmd.Process.Pid, syscall.SIGTERM); err != nil {
+			logrus.WithError(err).Warn("kill child process")
+		}
+	}()
 
 	logrus.Info("waiting for initComplete")
 	// wait for the child to connect to the parent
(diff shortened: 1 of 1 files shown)


giuseppe:
@AkihiroSuda PTAL

pkg/rootlessport/rootlessport_linux.go
```diff
@@ -120,8 +120,8 @@ func parent() error {
 		driverErr := driver.RunParentDriver(initComplete, quit, nil)
 		if driverErr != nil {
 			logrus.WithError(driverErr).Warn("parent driver exited")
+			errCh <- driverErr
 		}
-		errCh <- driverErr
```
AkihiroSuda: How does `RunParentDriver` return nil error?

AkihiroSuda: maybe close errCh?

mheon:
LGTM once comment is addressed

TomSweeneyRedHat:
LGTM once the comment is addressed.  CI isn't buying it, probably not related.

giuseppe:
it took a while as I've found another race condition.

More details in the second patch.

giuseppe:
@nalind I see the reexec package is using `SysProcAttr.Pdeathsig`, we should probably change it as well.

pkg/rootlessport/rootlessport_linux.go
```diff
@@ -142,15 +143,12 @@ func parent() error {
...
-	cmd := exec.Command(fmt.Sprintf("/proc/%d/exe", os.Getpid()))
+	cmd := exec.Command("/proc/self/exe")
 	cmd.Args = []string{reexecChildKey}
 	cmd.Stdin = childQuitR
 	cmd.Stdout = &logrusWriter{prefix: "child"}
 	cmd.Stderr = cmd.Stdout
 	cmd.Env = append(os.Environ(), reexecChildEnvOpaque+"="+string(opaqueJSON))
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Pdeathsig: syscall.SIGTERM,
-	}
```
AkihiroSuda: what was the problem?

giuseppe: the child process is killed whenever the parent thread exits which we have no control about since it is managed by the Go runtime.

giuseppe:
PR to address the same race condition in containers/storage: https://github.com/containers/storage/pull/530

PR #5187, opened 2020-02-12
add pkg/seccomp

vrothberg:
Add pkg/seccomp to consolidate all seccomp-policy related code which is
currently scattered across multiple packages and complicating the
creatconfig refactoring.

@baude @mheon PTAL

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -22,6 +22,7 @@ import (
 	"github.com/containers/libpod/pkg/inspect"
 	ns "github.com/containers/libpod/pkg/namespaces"
 	"github.com/containers/libpod/pkg/rootless"
+	"github.com/containers/libpod/pkg/seccomp"
 	cc "github.com/containers/libpod/pkg/spec"
 	"github.com/containers/libpod/pkg/util"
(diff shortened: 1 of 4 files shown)


cmd/podman/shared/create.go
```diff
@@ -23,6 +23,7 @@ import (
 	ns "github.com/containers/libpod/pkg/namespaces"
 	"github.com/containers/libpod/pkg/rootless"
 	cc "github.com/containers/libpod/pkg/spec"
+	"github.com/containers/libpod/pkg/seccomp"
```
mheon: Does it make sense to move this under specgen as a subpackage? I'm getting concerned about the number of things we have in pkg/

vrothberg: I think it makes sense in `pkg/` while other packages in `pkg/` might very well belong into a (not yet existing) `internal/` package for stuff we don't want to expose to external users.

baude: the pkg scrub should come after specgen is used in the cli and varlink is gone imho

baude:
yeah lgtm

TomSweeneyRedHat:
LGTM once the tests get happy

PR #5075, opened 2020-02-04
Force --all when --filter is passed to podman ps

mheon:
When we filter, it should be out of all containers, not just running ones, by default - this is necessary to ensure Docker compatability.

Fixes #5050

PR #5189, opened 2020-02-12
Only set --all when a status filter is given to ps

mheon:
The changes in #5075 turn out to be too aggressive; we should only be setting --all if a status= filter is given. Otherwise only running containers are filtered.

### DIFF
--- a/cmd/podman/ps.go
+++ b/cmd/podman/ps.go
@@ -205,9 +205,15 @@ func checkFlagsPassed(c *cliconfig.PsValues) error {
 	if c.Last >= 0 && c.Latest {
 		return errors.Errorf("last and latest are mutually exclusive")
 	}
-	// Filter forces all
+	// Filter on status forces all
 	if len(c.Filter) > 0 {
-		c.All = true
+		for _, filter := range c.Filter {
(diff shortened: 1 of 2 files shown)


baude:
lgtm

cmd/podman/ps.go
```diff
@@ -205,9 +205,19 @@ func checkFlagsPassed(c *cliconfig.PsValues) error {
...
+		foundStatus := false
+		for _, filter := range c.Filter {
+			splitFilter := strings.SplitN(filter, "=", 2)
+			if strings.ToLower(splitFilter[0]) == "status" {
+				foundStatus = true
+				break
+			}
+		}
+		if foundStatus {
+			c.All = true
```
rhatdan: No need for foundStatus, just set the c.All=true above in the foundStatus=true line.

mheon: Fixed

TomSweeneyRedHat:
LGTM assuming happy tests

mheon:
Remote client exec tests are flaking a lot

TomSweeneyRedHat:
indeed on the remote tests, I've seen bucking broncos bounce less.

PR #5190, opened 2020-02-12
filtering behavior correction

baude:
when filtering containers, if a status= is provided as an input filter, then we should override the all to always be true.

### DIFF
--- a/pkg/api/handlers/libpod/containers.go
+++ b/pkg/api/handlers/libpod/containers.go
@@ -83,6 +83,8 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 		Pod:       query.Pod,
 		Sync:      query.Sync,
 	}
+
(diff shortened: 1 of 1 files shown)


baude:
@marusak fyi

mheon:
LGTM

marusak:
Wondering, if this would not suffer with the same problem as filtering with ID does. Let me explain.
(just note that `status=(created|restarting|running|removing|paused|exited|dead)`)

When `all=true` and `filters.id=<some_id>` then all containers are returned, ignoring the filter. Without `all=true` the filtering works correctly as long as the container is running, when the container is not running, it is not considered (explained [here](https://github.com/containers/libpod/pull/5171#issuecomment-585079991)). So with change in this PR wouldn't it always return all containers ignoring the status filter?

PR #5202, opened 2020-02-13
test rootless_storage_path from storage.conf

QiWang19:
Add test for PR https://github.com/containers/storage/pull/529
test rootless_storage_path from strorage.conf. If user configured rootless_storage_path in storage.conf, podman info should suggest the change.

can be merged after #5929

### DIFF
--- a/test/e2e/info_test.go
+++ b/test/e2e/info_test.go
@@ -1,8 +1,13 @@
 package integration
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os"
+	"os/exec"
+	"path/filepath"
 
+	"github.com/containers/libpod/pkg/rootless"
 	. "github.com/containers/libpod/test/utils"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -49,5 +54,42 @@ var _ = Describe("Podman Info", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(ContainSubstring("registry"))
+	})
+
+	It("podman info rootless storage path", func() {
+		if !rootless.IsRootless() {
+			Skip("test of rootless_storage_path is only meaningful as rootless")
+		}
+		SkipIfRemote()
+		oldHOME, hasHOME := os.LookupEnv("HOME")
+		defer func() {
+			if hasHOME {
+				os.Setenv("HOME", oldHOME)
+			} else {
+				os.Unsetenv("HOME")
+			}
+		}()
+		os.Setenv("HOME", podmanTest.TempDir)
+		configPath := filepath.Join(os.Getenv("HOME"), ".config", "containers", "storage.conf")
+		err := os.RemoveAll(filepath.Dir(configPath))
+		Expect(err).To(BeNil())
+
+		err = os.MkdirAll(filepath.Dir(configPath), os.ModePerm)
+		Expect(err).To(BeNil())
+
+		rootlessStoragePath := `"/tmp/$HOME/$USER/$UID"`
+		driver := `"overlay"`
+		storageOpt := `"/usr/bin/fuse-overlayfs"`
+		storageConf := []byte(fmt.Sprintf("[storage]\ndriver=%s\nrootless_storage_path=%s\n[storage.options]\nmount_program=%s", driver, rootlessStoragePath, storageOpt))
+		err = ioutil.WriteFile(configPath, storageConf, os.ModePerm)
+		Expect(err).To(BeNil())
+
+		expect := filepath.Join("/tmp", os.Getenv("HOME"), os.Getenv("USER"), os.Getenv("UID"))
+		podmanPath := podmanTest.PodmanTest.PodmanBinary
+		cmd := exec.Command(podmanPath, "info", "--format", "{{.Store.GraphRoot}}")
+		out, err := cmd.CombinedOutput()
+		fmt.Println(string(out))
+		Expect(err).To(BeNil())
+		Expect(string(out)).To(ContainSubstring(expect))
 	})
 })


rhatdan:
Testing for https://github.com/containers/storage/pull/529

test/e2e/info_test.go
```diff
@@ -49,4 +53,26 @@ var _ = Describe("Podman Info", func() {
...
+		cmd := exec.Command("sed", "-i", "-e", "/rootless_storage_path/d", "-e", fmt.Sprintf("/\\[storage\\]/a rootless_storage_path = %s", rootStoragePath), configPath)
+		err := cmd.Run()
+		Expect(err).To(BeNil())
+
+		expect := filepath.Join("/tmp", os.Getenv("HOME"), os.Getenv("USER"), os.Getenv("UID"))
+
+		session := podmanTest.Podman([]string{"info", "--format", "{{.store.GraphRoot}}"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString).Should(Equal(expect))
```
rhatdan: I think the check shouldbe.
Expect(session.OutputToString).To(ContainString(expect))

QiWang19: looks like this is not related. the result <func() string>: 0x1376d40 doesn't contains or equaal to <string>: /tmp/home/some11410dude/some11410dude
any other way to fix?

rhatdan:
@QiWang19 Whats the scoop on this.  Once this passes tests we can merge storage PR.

QiWang19:
```
configPath := "/etc/containers/storage.conf"
desPath := filepath.Join(podmanTest.TempDir, "storage.conf")
cmd := exec.Command("sudo", "cp", configPath, desPath)
```
the `cp` failed. I want to try the storage.conf under $HOME

rhatdan:
Remove the tmpdir stuff.  That must be the issue.

cp /etc/containers/storage.conf /etc/containers/storage.conf.test
And then copy it back.

QiWang19:
> Remove the tmpdir stuff. That must be the issue.
> 
> cp /etc/containers/storage.conf /etc/containers/storage.conf.test
> And then copy it back.

fails again, cp returns the warning 
```
We trust you have received the usual lecture from the local System
[+0440s] Administrator. It usually boils down to these three things:
[+0440s] 
[+0440s]     #1) Respect the privacy of others.
[+0440s]     #2) Think before you type.
[+0440s]     #3) With great power comes great responsibility.
[+0440s] 
[+0440s] sudo: no tty present and no askpass program specified
```

rhatdan:
@edsantiago Any way for us to run sudo from the test platform?

edsantiago:
I doubt it, but in any case I have grave reservations about that test. It is a really, really bad idea to modify system files as part of a test, and if it has to be done it should be with lots of safeguards. I can think of many better ways to implement this test but won't have time to get into it today; I've set a tickler for myself for Monday morning.

rhatdan:
We could bind mount over it, or maybe we just add an environment variable into libpod and test with that.

test/e2e/info_test.go
```diff
@@ -49,4 +53,32 @@ var _ = Describe("Podman Info", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.IsJSONOutputValid()).To(BeTrue())
 	})
+	It("podman info rootless storage path", func() {
+		if !rootless.IsRootless() {
+			Skip("test rootless_storage_path for rootless")
+		}
+		SkipIfRemote()
+		configPath := "/etc/containers/storage.conf"
```
edsantiago: I'm really confused. In rootless, doesn't podman use `$HOME/.config/containers/storage.conf`? Shouldn't this test be using that?

test/e2e/info_test.go
```diff
@@ -49,4 +53,32 @@ var _ = Describe("Podman Info", func() {
...
+		desPath := "/etc/containers/storage.conf.test"
+		cmd := exec.Command("sudo", "cp", configPath, desPath)
+		out, err := cmd.CombinedOutput()
+		fmt.Println(string(out))
+		Expect(err).To(BeNil())
+
+		rootStoragePath := `"/tmp/$HOME/$USER/$UID"`
```
edsantiago: Why this complicated (and unreliable) value? It's not clear to me, from the commit message nor the code, what the point of this test is. Are you testing that podman can perform environment variable interpolation in `rootless_storage_path`, or simply that it will recognize an override? If the latter, it would be much better to define `rootStoragePath := /some/randomly/generated/string` and then avoid the duplication in line 73 (setting `expect`).

test/e2e/info_test.go
```diff
@@ -49,4 +53,32 @@ var _ = Describe("Podman Info", func() {
...
+		cmd = exec.Command("sudo", "sed", "-i", "-e", "/rootless_storage_path/d", "-e", fmt.Sprintf("/\\[storage\\]/a rootless_storage_path = %s", rootStoragePath), configPath)
```
edsantiago: Again, if you're really rootless, you could just `cat` these two lines to the rootless storage.conf. (Even better: is there a command-line or envariable way to override the default `storage.conf` path? I couldn't find one)

test/e2e/info_test.go
```diff
@@ -49,4 +53,32 @@ var _ = Describe("Podman Info", func() {
...
+		Expect(err).To(BeNil())
+
+		expect := filepath.Join("/tmp", os.Getenv("HOME"), os.Getenv("USER"), os.Getenv("UID"))
+		podmanPath := podmanTest.PodmanTest.PodmanBinary
+		cmd = exec.Command(podmanPath, "info", "--format", "{{.store.GraphRoot}}")
+		out, err = cmd.Output()
+		Expect(err).To(BeNil())
+		Expect(string(out)).To(ContainSubstring(expect))
+
+		cmd = exec.Command("sudo", "cp", desPath, configPath)
```
edsantiago: I think you can and should do this test without `sudo` and especially without modifying system files. But for future reference: should you ever really need to modify a system file, this is not enough. You need to write a `restore` that is safe, robust, bulletproof - that will run even if the `Expect`s fail, if the test has an error. I don't know the Go way to do this, maybe `defer`?

test/e2e/info_test.go
QiWang19: it's testing both variable interpolation in rootless_storage_path and that rootless_storage_path override the GraphRoot value in podman.

test/e2e/info_test.go
QiWang19: I can't find that envariable. I can try to add an variable for test and use $HOME/.config/containers without sudo.

test/e2e/info_test.go
edsantiago: Ah, okay, that wasn't clear from the commit message. In that case, perhaps a good way to test both would be:
```
    path = "/tmp/$RANDOM_STRING"
    ...write that to $HOME/.config/containers/storage.conf (yes, *BEFORE* setting the string)
    putenv("RANDOM_STRING", random-string)
    ...run podman info
    ...now make sure that what you get from podman matches the interpolated $RANDOM_STRING
```

QiWang19:
tests pass. can we merge https://github.com/containers/storage/pull/529? @edsantiago @rhatdan @TomSweeneyRedHat

rhatdan:
@edsantiago the goal of rootless_storage_path is for systems that have NFS homedirs and can't run with the default storage.conf that we create in the homedir.  IE $HOME/.local/share/containers/storage will not work. 

We want to allow administrators to tell podman/buildah to create their content in /var/tmp for example. 

Imagine a university with nfs homedirs, admin could set this up and every student logging in would automatically have their first podman work rather then fail, and force every student to edit the storage.conf that was generated in the homedir

edsantiago:
@rhatdan thanks, that makes sense; but it still should be possible to test this by creating a dummy `$HOME/.config/containers/storage.conf`, no? Go initialization is weird, but I would hope that config file parsing happens before accessing storage, so this could all be tested rootless?

QiWang19:
https://github.com/containers/storage/pull/529 has merged. Do I need to keep this PR?

rhatdan:
I wish there was a way to test this.  @edsantiago Any ideas.  I think the best thing would be to add an environment variable to containers/storage to specify a different path for /etc/containers/storage.conf, then we could test this out. in Podman.  @QiWang19 Could you work on a PR to support this in container/storage.

edsantiago:
@rhatdan did my comment above not make sense? About a dummy `storage.conf` under HOME?

QiWang19:
> @rhatdan did my comment above not make sense? About a dummy `storage.conf` under HOME?

What is 'dummy storage.conf'?
copy the $HOME/.config/containers/storage.conf to another path under $HOME as dummy storage.conf , and in the test need an environment variable to specify dummy storage.conf?

QiWang19:
> @rhatdan did my comment above not make sense? About a dummy `storage.conf` under HOME?

What is 'dummy storage.conf'?
copy the $HOME/.config/containers/storage.conf to another path under $HOME as dummy storage.conf , and in the test need an environment variable to specify dummy storage.conf?

edsantiago:
I'm sorry, I guess I didn't explain myself correctly. I was proposing something like this:
```
$ rm -rf ~/.config/containers
$ mkdir ~/.config/containers
$ printf "[storage]\nrootless_storage_path=XXXXXXXXXX\n" > ~/.config/containers/storage.conf
$ run podman info, see if it took effect
```
My assumption here is that for purposes of this test, `/etc` and `$HOME` are indistinguishable, because podman initializes config settings early. That is: you do not actually need to change anything in `/etc`. This assumption may be incorrect.

rhatdan:
Sure, I guess if this works, then we could merge this PR along with a vendoring of github.com/containers/storage.

QiWang19:
tests error 
```
time="2020-02-27T16:26:22-05:00" level=error msg="'overlay' is not supported over extfs at \"/home/some14791dude/.local/share/containers/storage/overlay\""
[+0883s]       /home/some14791dude/.local/share/containers/storage
```

should the 
```
[storage]
driver="overlay"
[storage.options]
mount_program="/usr/bin/fuse-overlayfs"
```
be added to the dummy storage.conf?

edsantiago:
Why not just `driver="vfs"`? It really shouldn't matter for purposes of this test

QiWang19:
> Why not just `driver="vfs"`? It really shouldn't matter for purposes of this test

Doesn't work locally. I can try in the test.

QiWang19:
I'm vendoring a new branch https://github.com/QiWang19/storage/tree/fix_storage-path. 
The `graphroot` is empty in the dummy storage.conf, to use the `rootless_storage_path`, the variable assignment should be moved out of the `if storageOpts.GraphRoot == ""{} else {}` block. And there is a path matching issue.

QiWang19:
@rhatdan @edsantiago PTAL. Will get this in for test storage.config `rootless_storage_path` storage.config?

test/e2e/info_test.go
```diff
@@ -49,4 +54,32 @@ var _ = Describe("Podman Info", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.IsJSONOutputValid()).To(BeTrue())
 	})
+	It("podman info rootless storage path", func() {
+		if !rootless.IsRootless() {
+			Skip("test rootless_storage_path for rootless")
```
edsantiago, line 59: Perhaps it might be clearer to say "this test is only meaningful as rootless"

test/e2e/info_test.go
```diff
@@ -49,4 +54,32 @@ var _ = Describe("Podman Info", func() {
...
+		}
+		SkipIfRemote()
+		configPath := filepath.Join(os.Getenv("HOME"), ".config", "containers", "storage.conf")
+		err := os.RemoveAll(filepath.Dir(configPath))
```
edsantiago, line 63: This makes me uncomfortable. I think these tests run only in CI, and can't be run by developers, but if that ever changes this could be catastrophic for regular users: it would destroy their existing storage.conf.

Did you consider something such as this instead?
```
    preserved_home := getenv("HOME")
    (add magic fail-safe such that HOME is restored on exit or failure)
    setenv("HOME", random-tmpdir)
    ...then create the storage.conf, etc
```

rhatdan: Lets try that.

QiWang19: fixed

QiWang19:
@edsantiago @rhatdan PTAL.

test/e2e/info_test.go
```diff
@@ -50,4 +55,41 @@ var _ = Describe("Podman Info", func() {
...
+		err = ioutil.WriteFile(configPath, storageConf, os.ModePerm)
+		Expect(err).To(BeNil())
+
+		expect := filepath.Join("/tmp", os.Getenv("HOME"), os.Getenv("USER"), os.Getenv("UID"))
+		podmanPath := podmanTest.PodmanTest.PodmanBinary
+		cmd := exec.Command(podmanPath, "info", "--format", "{{.Store.GraphRoot}}")
+		out, err := cmd.CombinedOutput()
+		fmt.Println(string(out))
+		Expect(err).To(BeNil())
+		Expect(string(out)).To(ContainSubstring(expect))
```
edsantiago, line 93: It'd be better to compare the string exactly, not substring, but I'm OK with this as it is. Just make a note for future work: exact comparisons are better than substring.

PR #5204, opened 2020-02-13
apiv2 container create using specgen

baude:
this uses the specgen structure to create containers rather than the outdated createconfig.  right now, only the apiv2 create is wired up.  eventually the cli will also have to be done.

### DIFF
--- a/pkg/api/handlers/libpod/containers_create.go
+++ b/pkg/api/handlers/libpod/containers_create.go
@@ -1,0 +1,29 @@
+package libpod
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/pkg/api/handlers/utils"
+	"github.com/containers/libpod/pkg/specgen"
+	"github.com/pkg/errors"
+)
+
+// CreateContainer takes a specgenerator and makes a container. It returns
+// the new container ID on success along with any warnings.
+func CreateContainer(w http.ResponseWriter, r *http.Request) {
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	var sg specgen.SpecGenerator
+	if err := json.NewDecoder(r.Body).Decode(&sg); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Decode()"))
+		return
+	}
+	ctr, err := sg.MakeContainer(runtime)
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	response := utils.ContainerCreateResponse{ID: ctr.ID()}
+	utils.WriteJSON(w, http.StatusCreated, response)
+}
--- a/pkg/api/handlers/utils/containers.go
+++ b/pkg/api/handlers/utils/containers.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"syscall"
@@ -9,9 +10,18 @@ import (
 	"github.com/containers/libpod/cmd/podman/shared"
 	"github.com/containers/libpod/libpod"
 	"github.com/containers/libpod/libpod/define"
+	createconfig "github.com/containers/libpod/pkg/spec"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
 )
+
+// ContainerCreateResponse is the response struct for creating a container
+type ContainerCreateResponse struct {
+	// ID of the container created
+	ID string `json:"id"`
+	// Warnings during container creation
+	Warnings []string `json:"Warnings"`
+}
 
 func KillContainer(w http.ResponseWriter, r *http.Request) (*libpod.Container, error) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
@@ -118,4 +128,19 @@ func GenerateFilterFuncsFromMap(r *libpod.Runtime, filters map[string][]string) 
 		}
 	}
 	return filterFuncs, nil
+}
+
+func CreateContainer(ctx context.Context, w http.ResponseWriter, runtime *libpod.Runtime, cc *createconfig.CreateConfig) {
+	var pod *libpod.Pod
+	ctr, err := shared.CreateContainerFromCreateConfig(runtime, cc, ctx, pod)
+	if err != nil {
+		Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "CreateContainerFromCreateConfig()"))
+		return
+	}
+
+	response := ContainerCreateResponse{
+		ID:       ctr.ID(),
+		Warnings: []string{}}
+
+	WriteResponse(w, http.StatusCreated, response)
 }
--- a/pkg/spec/config_linux.go
+++ b/pkg/spec/config_linux.go
@@ -32,8 +32,8 @@ func Device(d *configs.Device) spec.LinuxDevice {
 	}
 }
 
-// devicesFromPath computes a list of devices
-func devicesFromPath(g *generate.Generator, devicePath string) error {
+// DevicesFromPath computes a list of devices
+func DevicesFromPath(g *generate.Generator, devicePath string) error {
 	devs := strings.Split(devicePath, ":")
 	resolvedDevicePath := devs[0]
 	// check if it is a symbolic link
@@ -216,7 +216,7 @@ func getDevices(path string) ([]*configs.Device, error) {
 	return out, nil
 }
 
-func (c *CreateConfig) addPrivilegedDevices(g *generate.Generator) error {
+func addPrivilegedDevices(g *generate.Generator) error {
 	hostDevices, err := getDevices("/dev")
 	if err != nil {
 		return err
@@ -280,16 +280,16 @@ func (c *CreateConfig) createBlockIO() (*spec.LinuxBlockIO, error) {
 		var lwds []spec.LinuxWeightDevice
 		ret = bio
 		for _, i := range c.Resources.BlkioWeightDevice {
-			wd, err := validateweightDevice(i)
+			wd, err := ValidateweightDevice(i)
 			if err != nil {
 				return ret, errors.Wrapf(err, "invalid values for blkio-weight-device")
 			}
-			wdStat, err := getStatFromPath(wd.path)
+			wdStat, err := GetStatFromPath(wd.Path)
 			if err != nil {
-				return ret, errors.Wrapf(err, "error getting stat from path %q", wd.path)
+				return ret, errors.Wrapf(err, "error getting stat from path %q", wd.Path)
 			}
 			lwd := spec.LinuxWeightDevice{
-				Weight: &wd.weight,
+				Weight: &wd.Weight,
 			}
 			lwd.Major = int64(unix.Major(wdStat.Rdev))
 			lwd.Minor = int64(unix.Minor(wdStat.Rdev))
@@ -347,7 +347,7 @@ func makeThrottleArray(throttleInput []string, rateType int) ([]spec.LinuxThrott
 		if err != nil {
 			return []spec.LinuxThrottleDevice{}, err
 		}
-		ltdStat, err := getStatFromPath(t.path)
+		ltdStat, err := GetStatFromPath(t.path)
 		if err != nil {
 			return ltds, errors.Wrapf(err, "error getting stat from path %q", t.path)
 		}
@@ -361,7 +361,7 @@ func makeThrottleArray(throttleInput []string, rateType int) ([]spec.LinuxThrott
 	return ltds, nil
 }
 
-func getStatFromPath(path string) (unix.Stat_t, error) {
+func GetStatFromPath(path string) (unix.Stat_t, error) {
 	s := unix.Stat_t{}
 	err := unix.Stat(path, &s)
 	return s, err
--- a/pkg/spec/spec.go
+++ b/pkg/spec/spec.go
@@ -16,9 +16,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-const cpuPeriod = 100000
+const CpuPeriod = 100000
 
-func getAvailableGids() (int64, error) {
+func GetAvailableGids() (int64, error) {
 	idMap, err := user.ParseIDMapFile("/proc/self/gid_map")
 	if err != nil {
 		return 0, err
@@ -80,7 +80,7 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 	}
 	gid5Available := true
 	if isRootless {
-		nGids, err := getAvailableGids()
+		nGids, err := GetAvailableGids()
 		if err != nil {
 			return nil, err
 		}
@@ -197,8 +197,8 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 		addedResources = true
 	}
 	if config.Resources.CPUs != 0 {
-		g.SetLinuxResourcesCPUPeriod(cpuPeriod)
-		g.SetLinuxResourcesCPUQuota(int64(config.Resources.CPUs * cpuPeriod))
+		g.SetLinuxResourcesCPUPeriod(CpuPeriod)
+		g.SetLinuxResourcesCPUQuota(int64(config.Resources.CPUs * CpuPeriod))
 		addedResources = true
 	}
 	if config.Resources.CPURtRuntime != 0 {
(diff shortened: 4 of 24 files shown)


libpod/container.go
```diff
@@ -372,11 +372,11 @@ type ContainerConfig struct {
 	StopTimeout uint `json:"stopTimeout,omitempty"`
 	// Time container was created
 	CreatedTime time.Time `json:"createdTime"`
-	// NoCgroups indicates that the container will not create CGroups. It is
+	// CgroupsMode indicates that the container will not create CGroups. It is
```
larsks: This looks like some over-enthusiastic search-and-replace...I suspect this should still read `NoCgroups`.

baude: indeed, good catch

libpod/define/errors.go
```diff
@@ -45,6 +45,7 @@ var (
 	// ErrPodFinalized indicates that the pod has already been created and
 	// cannot be modified
 	ErrPodFinalized = errors.New("pod has been finalized")
+
```
mheon: ?

pkg/api/handlers/libpod/containers_create.go
```diff
@@ -0,0 +1,30 @@
...
+		return
+	}
+	ctr, err := sg.MakeContainer(runtime)
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	response := utils.ContainerCreateResponse{ID: ctr.ID()}
+	utils.WriteJSON(w, http.StatusCreated, response)
+
```
mheon: Extra newline

pkg/api/handlers/utils/containers.go
```diff
@@ -119,3 +129,29 @@ func GenerateFilterFuncsFromMap(r *libpod.Runtime, filters map[string][]string)
 	}
 	return filterFuncs, nil
 }
+
+func CreateContainer(ctx context.Context, w http.ResponseWriter, runtime *libpod.Runtime, cc *createconfig.CreateConfig) {
+	var pod *libpod.Pod
+	ctr, err := shared.CreateContainerFromCreateConfig(runtime, cc, ctx, pod)
+	if err != nil {
+		//if strings.Contains(err.Error(), "invalid log driver") {
```
mheon: Can we get rid of this?

pkg/spec/config_linux.go
```diff
@@ -32,6 +32,11 @@ func Device(d *configs.Device) spec.LinuxDevice {
 	}
 }
 
+// DevicesFromPath is an exported wrapper for devicesFromPath
+func DevicesFromPath(g *generate.Generator, devicePath string) error {
```
mheon: Can we just rename devicesFromPath instead? I don't think we need separate exported/not exported versions

pkg/spec/spec.go
```diff
@@ -404,8 +404,13 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 	return configSpec, nil
 }
 
-func blockAccessToKernelFilesystems(config *CreateConfig, g *generate.Generator) {
-	if !config.Security.Privileged {
+// BlockAccessToKernelFilesystems is an exported wrapper for blockAccessToKernelFilesystems
+func BlockAccessToKernelFilesystems(privileged, pidModeIsHost bool, g *generate.Generator) {
```
mheon: Probably don't need a separate exported version, can just rename the other one

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/containers/libpod/libpod/define"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// MakeContainer creates a container based on the SpecGenerator
+func (s *SpecGenerator) MakeContainer(rt *libpod.Runtime) (*libpod.Container, error) {
+	var pod *libpod.Pod
+	if err := s.validate(rt); err != nil {
+		return nil, errors.Wrap(err, "failed to create container")
```
mheon: Should this be something more like "invalid config provided"?

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	options = append(options, s.createExitCommandOption(rtc, podmanPath))
+	newImage, err := rt.ImageRuntime().NewFromLocal(s.Image)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO mheon wants to talk with Dan about this
+	useImageVolumes := s.ImageVolumeMode == "bind"
+	options = append(options, libpod.WithRootFSFromImage(newImage.ID(), s.Image, useImageVolumes))
```
mheon: You know, I think useImageVolumes might be 100% unused...

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+	if len(s.Name) > 0 {
+		logrus.Debugf("setting container name %s", s.Name)
+		options = append(options, libpod.WithName(s.Name))
+	}
+	if s.Pod != "" {
+		logrus.Debugf("adding container to pod %s", s.Pod)
+		options = append(options, rt.WithPod(pod))
+	}
+
+	//if len(mounts) != 0 || len(namedVolumes) != 0 {
```
mheon: I think we can still do this with s.Mounts and s.NamedVolumes

mheon: Though... Crap, that doesn't include VolumesFrom.

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+	//
+	//if len(namedVolumes) != 0 {
+	//	options = append(options, libpod.WithNamedVolumes(namedVolumes))
+	//}
+	//
+	if len(s.Command) != 0 {
+		options = append(options, libpod.WithCommand(s.Command))
+	}
+
+	options = append(options, libpod.WithEntrypoint(s.Entrypoint))
```
mheon: Shouldn't this also be only if len(s.Entrypoint) > 0?

mheon: Hmmm. We might want to pull from the image in that case.

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+		options = append(options, libpod.WithStopTimeout(*s.StopTimeout))
+	}
+
+	if s.LogConfiguration != nil {
+		if len(s.LogConfiguration.Path) > 0 {
+			options = append(options, libpod.WithLogPath(s.LogConfiguration.Path))
+		}
+		if len(s.LogConfiguration.Options) > 0 {
+			// Note: I'm really guessing here.
+			options = append(options, libpod.WithLogTag(s.LogConfiguration.Options["tag"]))
```
mheon: Should only do if Options["tag"] exists

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+
+		if len(s.LogConfiguration.Driver) > 0 {
+			options = append(options, libpod.WithLogDriver(s.LogConfiguration.Driver))
+		}
+	}
+
+	// Security options
+
+	options = append(options, libpod.WithSecLabels(s.SelinuxOpts))
```
mheon: @rhatdan Should we only do this if explicitly set to something?

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+	// Get namespace related options
+	namespaceOptions, err := s.generateNamespaceContainerOpts(rt)
+	if err != nil {
+		return nil, err
+	}
+	options = append(options, namespaceOptions...)
+
+	// TODO NetworkNS still needs to be done!
+
+	options = append(options, libpod.WithConmonPidFile(s.ConmonPidFile))
```
mheon: I think we only do this if ConmonPidFile != ""

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+		options = append(options, libpod.WithRootFS(s.Rootfs))
+	}
+	// Default used if not overridden on command line
+
+	if s.RestartPolicy != "" {
+		if s.RestartPolicy == "unless-stopped" {
+			return nil, errors.Wrapf(define.ErrInvalidArg, "the unless-stopped restart policy is not supported")
+		}
+
+		split := strings.Split(s.RestartPolicy, ":")
```
mheon: This is wrong - RestartPolicy is already split in SpecGenerator into RestartPolicy and RestartRetries

pkg/specgen/namespaces.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+// IsPrivate indicates the namespace is private
+func (n *Namespace) IsPrivate() bool {
+	return n.NSMode == Private
+}
+
+// validate perform simple validation on the namespace to make sure it is not
+// invalid from the get-go
+func (n *Namespace) validate() error {
```
mheon: We should check that NSMode is one of the valid ones listed above

pkg/specgen/namespaces.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+	}
+	return nil
+}
+
+func (s *SpecGenerator) generateNamespaceContainerOpts(rt *libpod.Runtime) ([]libpod.CtrCreateOption, error) {
+	options := make([]libpod.CtrCreateOption, 0)
+
+	// Cgroups
+	if s.CgroupNS.IsPrivate() {
```
mheon: All of these should probably be switch statements, so we have a default for not-handled modes

pkg/specgen/oci.go
```diff
@@ -0,0 +1,260 @@
...
+
+	isRootless := rootless.IsRootless()
+	if isRootless {
+		inUserNS = true
+	}
+	if !s.UserNS.IsHost() {
+		if s.UserNS.IsContainer() || s.UserNS.IsPath() {
+			inUserNS = true
+		}
+		if s.IDMappings != nil && (len(s.IDMappings.UIDMap) > 0 || len(s.IDMappings.GIDMap) > 0) {
```
mheon: We might be able to set this to `s.UserNS.IsPrivate()` instead

pkg/specgen/create.go
baude: what do you want me to do; it is in the current code path.

mheon:
I'll look into useImageVolumes on the Libpod side, and make a PR to tear it
out of it really is unused, we can merge that before or after this, doesn't
really matter

pkg/api/handlers/libpod/containers_create.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/pkg/errors"
+)
+
+// CreateContainer takes a specgenerator and makes a container. It returns
+// the new container ID on success along with any warnings.
+func CreateContainer(w http.ResponseWriter, r *http.Request) {
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	var sg specgen.SpecGenerator
+	if err := json.NewDecoder(r.Body).Decode(&sg); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Decode()"))
```
TomSweeneyRedHat: Not a big fan of "Something went wrong".  I'd much prefer just "Error: "

pkg/api/handlers/utils/containers.go
```diff
@@ -119,3 +129,29 @@ func GenerateFilterFuncsFromMap(r *libpod.Runtime, filters map[string][]string)
...
+		//	w.Header().Set("Content-Type", "text/plain; charset=us-ascii")
+		//	w.WriteHeader(http.StatusInternalServerError)
+		//	msg := fmt.Sprintf("logger: no log driver named '%s' is registered", input.HostConfig.LogConfig.Type)
+		//	if _, err := fmt.Fprintln(w, msg); err != nil {
+		//		log.Errorf("%s: %q", msg, err)
+		//	}
+		//	//s.WriteResponse(w, http.StatusInternalServerError, fmt.Sprintf("logger: no log driver named '%s' is registered", input.HostConfig.LogConfig.Type))
+		//	return
+		//}
+		Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "CreateContainerFromCreateConfig()"))
```
TomSweeneyRedHat: Something nit

pkg/specgen/config_linux_cgo.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if scp == seccomp.PolicyImage {
+		labels, err := img.Labels(context.Background())
+		if err != nil {
+			return nil, err
+		}
+		imagePolicy := labels[seccomp.ContainerImageLabel]
+		if len(imagePolicy) < 1 {
+			return nil, errors.New("no secomppolicy defined by image")
```
TomSweeneyRedHat: nit "no secomppolicy" -> "secomppolicy is not"

pkg/specgen/create.go
```diff
@@ -0,0 +1,203 @@
...
+				return nil, errors.Wrapf(define.ErrInvalidArg, "restart policy requires a positive number of retries")
+			}
+			options = append(options, libpod.WithRestartRetries(uint(numTries)))
+		}
+		options = append(options, libpod.WithRestartPolicy(split[0]))
+	}
+
+	if s.ContainerHealthCheckConfig.HealthConfig != nil {
+		options = append(options, libpod.WithHealthCheck(s.ContainerHealthCheckConfig.HealthConfig))
+		logrus.Debugf("New container has a health check")
```
TomSweeneyRedHat: can you include the container name/id here in the message?

baude: no  because the container is not created yet

mheon: Should we just drop this, then? I don't know if there's value in printing this on the server side if we know nothing more...

mheon:
@giuseppe @rhatdan PTAL - I'd love another set of eyes on this before we merge

libpod/runtime_ctr.go
```diff
@@ -165,10 +165,10 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (c *Contai
 		ctr.ociRuntime = ociRuntime
 	}
 
-	// Check NoCgroups support
+	// Check CgroupsMode support
```
rhatdan: This comment, versus the lines below, does not make sense.

pkg/specgen/config_linux_cgo.go
rhatdan: no seccomp policy defined by image

pkg/specgen/create.go
rhatdan: I would hope the WithSecLabels ignores if nothing is set.

libpod/container.go
```diff
@@ -372,11 +372,11 @@ type ContainerConfig struct {
...
 	// incompatible with CgroupParent.  Deprecated in favor of CgroupsMode.
 	NoCgroups bool `json:"noCgroups,omitempty"`
 	// CgroupsMode indicates how the container will create cgroups
-	// (disabled, no-conmon, enabled).  It supersedes NoCgroups.
+	// (disabled, no-conmon, enabled).  It supersedes CgroupsMode.
```
mheon: Aggressive rename

libpod/container.go
mheon: Aggressive rename

libpod/container_internal.go
```diff
@@ -1140,7 +1140,7 @@ func (c *Container) stop(timeout uint) error {
 	// the containers cgroup in order to make sure the container is stopped.
 	all := !c.hasNamespace(spec.PIDNamespace)
 	// We can't use --all if CGroups aren't present.
-	// Rootless containers with CGroups v1 and NoCgroups are both cases
+	// Rootless containers with CGroups v1 and CgroupsMode are both cases
```
mheon: Aggressive rename

libpod/options.go
```diff
@@ -1112,7 +1112,7 @@ func WithCgroupParent(parent string) CtrCreateOption {
 		}
 
 		if ctr.config.NoCgroups {
-			return errors.Wrapf(define.ErrInvalidArg, "CgroupParent conflicts with NoCgroups")
+			return errors.Wrapf(define.ErrInvalidArg, "CgroupParent conflicts with CgroupsMode")
```
mheon: Aggressive Rename

libpod/runtime_ctr.go
```diff
@@ -168,7 +168,7 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (c *Contai
 	// Check NoCgroups support
 	if ctr.config.NoCgroups {
 		if !ctr.ociRuntime.SupportsNoCgroups() {
-			return nil, errors.Wrapf(define.ErrInvalidArg, "requested OCI runtime %s is not compatible with NoCgroups", ctr.ociRuntime.Name())
+			return nil, errors.Wrapf(define.ErrInvalidArg, "requested OCI runtime %s is not compatible with CgroupsMode", ctr.ociRuntime.Name())
```
mheon: Aggressive Rename

libpod/runtime_ctr.go
```diff
@@ -194,7 +194,7 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (c *Contai
 	// Otherwise, the OCI runtime won't be able to stop our container.
 	if ctr.config.NoCgroups {
 		if ctr.config.Spec.Linux == nil {
-			return nil, errors.Wrapf(define.ErrInvalidArg, "must provide Linux namespace configuration in OCI spec when using NoCgroups")
+			return nil, errors.Wrapf(define.ErrInvalidArg, "must provide Linux namespace configuration in OCI spec when using CgroupsMode")
```
mheon: Aggressive rename

PR #5206, opened 2020-02-14
Allow devs to set labels in container images for default capabilities.

rhatdan:
This patch allows users to specify the list of capabilities required
to run their container image.

Setting a image/container label "io.containers.capabilities=setuid,setgid"
tells podman that the contained image should work fine with just these two
capabilties, instead of running with the default capabilities, podman will
launch the container with just these capabilties.

If the user or image specified capabilities that are not in the default set,
the container will print an error message and will continue to run with the
default capabilities.


rhatdan:
@vrothberg This is what we talked about and I covered in my talk in BRNO.
PTAL
Need to get something similar for seccomp.json files

docs/source/markdown/podman-create.1.md
```diff
@@ -409,6 +409,16 @@ millions of trillions.
 
 Add metadata to a container (e.g., --label com.example.key=value)
 
+Users can set a special label **io.containers.capabilities=CAP1,CAP2,CAP3** that
```
mheon: I feel like this belongs elsewhere. We don't expect people to set this at `podman run` time, right? Instead we want them to load it from the image.

Should we instead be setting `io.containers.capabilities` at container creation time based on the capabilities we put in the spec, so that it will be there when we commit images?

docs/source/markdown/podman-create.1.md
```diff
@@ -989,6 +999,14 @@ can override the working directory by using the **-w** option.
 $ podman create alpine ls
 ```
 
+### Run a container with just the SETUID and SETGID Capabilities
+
+
+```
+
+$ podman run -d --label=io.containers.capabilities=setuid,setgid fedora /usr/bin/server
```
mheon: Why do I do this instead of `--cap-add` and `--cap-drop`? I don't know if this makes much sense outside of loading from images.

docs/source/markdown/podman-create.1.md
rhatdan: No the goal here is to override the defaults with something more secure.

docs/source/markdown/podman-create.1.md
rhatdan: The idea is to get this data into the container image.  And maybe I should add the data to Docker build.
The real goal is to get people to add 
LABEL io.containers.capabilities=setuid,setgid 
In their Containerfile/Dockerfile.
Then users of their container image will get this new more secure behaviour.

rhatdan:
Updated to add more information to podman build.

docs/source/markdown/podman-create.1.md
mheon: But rather than trying to get people to add the label themselves, we can make it for them from what they entered in `--cap-add` and `--cap-drop`. We know the final capabilities the container was created with, so it's pretty trivial to make the label from that.

rhatdan: But that is not natural behaviour for Building container images.  Most people, if they are going to use this feature are going to do it via a Dockerfile.  We could add the label as a side effect of --cap-add and --cap-drop, but I am not sure that would be expected by users.

mheon: I agree that most people will be using a Dockerfile - I'm mostly confused as to why people would ever want to set it directly at `podman run` time, instead of using `--cap-add` and `--cap-drop`.

pkg/spec/spec.go
```diff
@@ -315,6 +315,14 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 	}
 	configSpec := g.Config
 
+	var capRequired []string
+	for key, val := range config.Labels {
+		if key == "io.containers.capabilities" {
```
vrothberg: This should really be a constant.

test/e2e/run_security_labels.go
```diff
@@ -0,0 +1,325 @@
...
+		inspect.WaitWithDefaultTimeout()
+		Expect(inspect.ExitCode()).To(Equal(0))
+
+		ctr := inspect.InspectContainerToJSON()
+		caps := strings.Join(ctr[0].EffectiveCaps, ",")
+		Expect(caps).To(Equal("CAP_SETUID,CAP_SETGID"))
+	})
+
+	It("podman bad security labels", func() {
+		test1 := podmanTest.Podman([]string{"create", "--label", "io.containers.capabilities=sys_admin", "--name", "test1", "alpine", "echo", "test1"})
```
vrothberg: Looks like we're not testing with an image, and I believe we should.

pkg/spec/createconfig.go
```diff
@@ -112,6 +112,7 @@ type NetworkConfig struct {
 type SecurityConfig struct {
 	CapAdd                  []string // cap-add
 	CapDrop                 []string // cap-drop
+	CapRequired             []string // cap-required
```
vrothberg: Can we call it `CapBase` to be consistent with https://github.com/containers/libpod/pull/5209?

pkg/spec/security.go
```diff
@@ -123,6 +125,25 @@ func (c *SecurityConfig) ConfigureGenerator(g *generate.Generator, user *UserCon
 		return err
 	}
 
+	privCapRequired := []string{}
+
+	if !c.Privileged && len(c.CapRequired) > 0 {
```
vrothberg: I think we should always set `CapRequired`. If it's not set via the CLI or by the image, we should use the default ones.

pkg/spec/spec.go
```diff
@@ -315,6 +315,14 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 	}
 	configSpec := g.Config
 
+	var capRequired []string
```
vrothberg: Please add comments.

pkg/spec/spec.go
vrothberg: I would love the constant to be in `pkg/capabilities` as we do for the seccomp policies now in [pkg/seccomp](https://github.com/containers/libpod/blob/master/pkg/seccomp/seccomp.go).

test/e2e/run_security_labels.go
rhatdan: I commit an image and test off the image, but if you are talking about an image built with a Containerfile, I could do that as well.

pkg/spec/security.go
rhatdan: We can talk about this, but this could be a problem.  I was thinking CapRequired was something that the user or the image had specified as the capabilies required to run the application.  As  opposed to the caplist, which is the "default capabilities + CapADD - CapDrop).  If we set CapRequired == CapDefault and only overrode it when the label was set, then we would not be able to tell when the user specified at capability outside of the "default" list of capabilities.

test/e2e/run_security_labels.go
vrothberg: Didn't notice the commit below. Thanks! LGTM

pkg/spec/security.go
vrothberg: Maybe we can have a quick bluejeans later? I want to show the `pkg/seccomp` package that we can extend for these needs.

docs/source/markdown/podman-build.1.md
```diff
@@ -272,6 +272,16 @@ BUILDAH\_ISOLATION environment variable.  `export BUILDAH_ISOLATION=oci`
 
 Add an image *label* (e.g. label=*value*) to the image metadata. Can be used multiple times.
 
+Users can set a special label **io.containers.capabilities=CAP1,CAP2,CAP3** that
+indicates the default list of Linux capabilities required for the container to
+run properly. Setting this label in a container or a container image tells podman
```
TomSweeneyRedHat: podman -> Podman

docs/source/markdown/podman-build.1.md
```diff
@@ -272,6 +272,16 @@ BUILDAH\_ISOLATION environment variable.  `export BUILDAH_ISOLATION=oci`
...
+to run the container with just these capabilties. Podman will launch the container
+with just the specified capabilties, as long as this list of capabilities is a
+subset of the default list.
+
+If the user or image specified capabilities that are not in the default set, the container
```
TomSweeneyRedHat: I find "user or image" to be a bit confusing.  Could we just drop that?  "If the specified capabilities..."

TomSweeneyRedHat: ooops, if dropped, the following "that" doesn't work.   "If the specified capabilities are not in the default set, ..."

docs/source/markdown/podman-commit.1.md
```diff
@@ -86,6 +86,11 @@ $ podman commit -q --pause=false containerID image-committed
 e3ce4d93051ceea088d1c242624d659be32cf1667ef62f1d16d6b60193e2c7a8
 ```
 
+```
+$ podman commit -q --change LABEL=io.containers.capabilities=setuid,setgid epic_nobel privimage
```
TomSweeneyRedHat: You don't have the verbiage from build in here, should you?

test/e2e/run_security_labels.go
```diff
@@ -0,0 +1,325 @@
...
+		inspect := podmanTest.Podman([]string{"inspect", "test1"})
+		inspect.WaitWithDefaultTimeout()
+		Expect(inspect.ExitCode()).To(Equal(0))
+
+		ctr := inspect.InspectContainerToJSON()
+		caps := strings.Join(ctr[0].EffectiveCaps, ",")
+		Expect(caps).To(Not(Equal("CAP_SETUID,CAP_SETGID")))
+	})
+
+})
```
TomSweeneyRedHat: might be good to add build, commit and the commands other than create.

docs/source/markdown/podman-create.1.md
rhatdan: Sure, but do you just want me to not document that you can do this?

pkg/spec/spec.go
rhatdan: Will have to move this to containers/common since pkg/capabilities is in there now, Also I think this should be a list. since it might change in the future if this got standardized and they picked a different name.

test/e2e/run_security_labels.go
rhatdan: Also added a Dockerfile build

pkg/spec/createconfig.go
rhatdan: I don't think this is the same thing.  CapBase is the default capabilities for all containers.

docs/source/markdown/podman-build.1.md
rhatdan: Fixed

docs/source/markdown/podman-build.1.md
rhatdan: Fixed

docs/source/markdown/podman-commit.1.md
rhatdan: I added it.

test/e2e/run_security_labels.go
rhatdan: Yes added.

TomSweeneyRedHat:
LGTM, but definitely want a @mheon head nod

mheon:
I'm still iffy about documenting this in run and create - it seems like extra confusion since we already have ways to do this. Might want to talk about it at scrum.

rhatdan:
@mheon @vrothberg @giuseppe @baude PTAL

rhatdan:
@mheon @vrothberg @giuseppe @baude PTAL
Passes tests now and removed mentions of the label in podman run and podman create.

rhatdan:
As @TomSweeneyRedHat would say, Happy Green Buttons.   Can I get this reviewed?
@mheon @giuseppe @vrothberg @baude @QiWang19 @jwhonce

docs/source/markdown/podman-commit.1.md
mheon: This is badly formatted; should be `--change 'LABEL io.containers.capabilities=setuid,setgid'`

docs/source/markdown/podman-create.1.md
```diff
@@ -1000,6 +1000,14 @@ can override the working directory by using the **-w** option.
 $ podman create alpine ls
 ```
 
+### Run a container with just the SETUID and SETGID Capabilities
```
mheon: Suggest rename to "Set default capabilities for container to SETUID and SETGID" or similar

docs/source/markdown/podman-commit.1.md
rhatdan: Their might be a bug in the parser, but my syntax works, and yours does not.  This is also documented two lines higher.
```
$ podman commit --change "LABEL foo=bar" myweb dan1
Error: invalid syntax for --change: LABEL foo=bar

$ podman commit --change "LABEL=foo=bar" myweb dan1
Copying blob 195be5f8be1d skipped: already exists  
Copying blob 5f70bf18a086 [--------------------------------------] 0.0b / 0.0b
Copying config e87256c3fc done  
Storing signatures
e87256c3fcda3881112f19dbfda9a733bee0472354269dc62c2c5c7e81c69e1f
```


oo=bar" myweb dan1
Error: invalid syntax for --change: LABEL foo=bar

docs/source/markdown/podman-create.1.md
rhatdan: Removed these commands, since they don't make sense for podman run/create any longer.

docs/source/markdown/podman-create.1.md
```diff
@@ -1000,6 +1000,14 @@ can override the working directory by using the **-w** option.
...
+
+
+```
+
+$ podman run -d --label=io.containers.capabilities=setuid,setgid fedora /usr/bin/server
```
vrothberg:
```suggestion
$ podman create -d --label=io.containers.capabilities=setuid,setgid fedora /usr/bin/server
```

docs/source/markdown/podman-create.1.md
vrothberg:
```suggestion
### Create a container with just the SETUID and SETGID Capabilities
```

docs/source/markdown/podman-build.1.md
```diff
@@ -279,6 +279,16 @@ BUILDAH\_ISOLATION environment variable.  `export BUILDAH_ISOLATION=oci`
...
 Add an image *label* (e.g. label=*value*) to the image metadata. Can be used multiple times.
 
+Users can set a special LABEL **io.containers.capabilities=CAP1,CAP2,CAP3** in
+a Containerfile that specified the list of Linux capabilities required for the
+container to run properly. This label specified in a container image tells
+Podman to run the container with just these capabilties. Podman launches the
+container with just the specified capabilties, as long as this list of
+capabilities is a subset of the default list.
+
+If the specified capabilities are not in the default set, the container will
```
vrothberg:
```suggestion
If the specified capabilities are not in the default set, Podman will
```

vrothberg:
I am loving the new GitHub feature to suggest changes in the reviews.

docs/source/markdown/podman-create.1.md
rhatdan: Removing line.

docs/source/markdown/podman-create.1.md
rhatdan: Removing line.
We don't want to document this label in podman run or create.

rhatdan:
@vrothberg Sadly our Gating blocks it.

PR #5207, opened 2020-02-14
Fix SELinux labels of volumes

rhatdan:
If we attempt to label a volume and the file system
does not support labeling, then just warn.  SELinux
may or may not work, on the volume.

There is no way to setup a private label on a newly
created volume without using the container mountlabel.

If we don't have a mount label at the time of creation of
the volume, the only option we have is to create a shared
label.


rhatdan:
May fix: https://github.com/containers/libpod/issues/5200
Will get it further along since we don't know if SELinux labeling will prevent the container from interacting with the label of the file system.

rhatdan:
@mheon @baude @giuseppe @TomSweeneyRedHat @QiWang19 PTAL

TomSweeneyRedHat:
LGTM once the flakes in the tests are chased.

PR #5209, opened 2020-02-14
Undocker part 1)

vrothberg:
Create dedicated packages:
* pkg/signal for signal handling
* pkg/capabilities for all-things caps

Especially `pkg/capabilities` will come in handy for #5206.


vrothberg:
@rhatdan @giuseppe @mheon @baude PTAL

vrothberg:
The rdoproject.org/github-check is starting to get annoying. Although it's not blocking, it causes the CI to be red for way too long.

rhatdan:
Should this be happening in github.com/containers/common so that buildah can take advantage?

vrothberg:
> Should this be happening in github.com/containers/common so that buildah can take advantage?

I prefer to wait until we separate Podman from libpod. This will allow for Buildah to use more package from libpod.

pkg/signal/signal_linux.go
```diff
@@ -1,8 +1,14 @@
-// +build !mips,!mipsle,!mips64,!mips64le
+// +build linux
```
haircommander: should we add in a copyright line at the top, even if it wasn't present in the old file?

vrothberg: Absolutely. Thanks for noticing!

vrothberg: Done :heavy_check_mark:

TomSweeneyRedHat:
Personally, if the end goal is to get the packages into common eventually, I'd say move 'em now even if other pkg's from Libpod go later.  Your call.

rhatdan:
LGTM
@mheon @TomSweeneyRedHat @giuseppe @haircommander PTAL

PR #5210, opened 2020-02-14
WIP: Add support for pinned namespaces

saschagrunert:
The idea is to pin the namespaces to a dedicated on-disk location by not having the need for a pause container any more (analogous to the CRI-O implementation).

This is still in WIP and would need documentation, testing and some pre-defined design decisions.

#### Demo

What seems to work right now, when a `pinns` exectuable (from [CRI-O pinns](https://github.com/cri-o/cri-o/tree/master/pinns) or [pinns.rs](https://github.com/saschagrunert/pinns.rs)) is in `$PATH`.

##### Namespace life-cycle handling
```
> sudo ./bin/podman pod create -n mypod -i
b73890f9b3143b66a7b78dc2755fc36df1efd2989a3b759448448ecf72d75164

> sudo ls /var/run/libpod/ns/b73890f9b3143b66a7b78dc2755fc36df1efd2989a3b759448448ecf72d75164
ipc  net  uts

> sudo ./bin/podman pod rm mypod
b73890f9b3143b66a7b78dc2755fc36df1efd2989a3b759448448ecf72d75164

> sudo ls /var/run/libpod/ns/b73890f9b3143b66a7b78dc2755fc36df1efd2989a3b759448448ecf72d75164
ls: cannot access '/var/run/libpod/ns/b73890f9b3143b66a7b78dc2755fc36df1efd2989a3b759448448ecf72d75164': No such file or directory
```

##### Containers in pods
```
> sudo ./bin/podman pod create -n mypod -i
f16880f174021ed71861213b4ec4ed64ddc2bbe5de09405efb2a1f32b607fdee

> sudo ./bin/podman run --pod mypod -d alpine sleep infinity
26b52409249670abccc929e7e9a94eae561a888cac3e2765f5052a322070e339

> sudo ./bin/podman run --pod mypod -d alpine sleep infinity
07637abb5078d43bb6679eea14436ebb8586d0634c22ee06d568cd2a97db721e

> sudo ./bin/podman run --pod mypod -d alpine sleep infinity
fe6564f40fdbc27f57f33728eeed156f89891b4e7ccd84606207d46d7310ac14

> sudo ./bin/podman ps --ns
CONTAINER ID  NAMES           PID    CGROUPNS    IPC         MNT         NET         PIDNS       USERNS      UTS
fe6564f40fdb  charming_cerf   15084  4026531835  4026541537  4026541669  4026541540  4026541670  4026531837  4026541469
07637abb5078  laughing_nash   14929  4026531835  4026541537  4026541667  4026541540  4026541668  4026531837  4026541469
26b524092496  vibrant_wright  14281  4026531835  4026541537  4026541665  4026541540  4026541666  4026531837  4026541469
```

#### Still TODO
- Documentation
- Testing
- Find a way to provide pinns here


libpod/runtime_pod_namespaces_linux.go
```diff
@@ -0,0 +1,88 @@
+// +build linux
+
+package libpod
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/docker/docker/pkg/mount"
+	"github.com/docker/docker/pkg/symlink"
```
saschagrunert: Might clash with https://github.com/containers/libpod/pull/5209

vrothberg: Yes. We want to get rid of direct dependencies on docker/docker.

cmd/podman/pod_create.go
```diff
@@ -58,6 +58,7 @@ func init() {
 	// Remember to do this safely by checking len, etc.
 
 	flags.StringVar(&podCreateCommand.CgroupParent, "cgroup-parent", "", "Set parent cgroup for the pod")
+	flags.BoolVarP(&podCreateCommand.PinNamespaces, "pin-namespaces", "i", false, "Pin namespaces instead of using an infra container")
```
haircommander: for now, the "instead of using an infra container" isn't quite correct. I would want to update the description once we can actually drop the infra container

haircommander: oh I see that's the eventual intention of this PR, nevermind :smile:

libpod/runtime_pod_namespaces_linux.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	// Not supported by cri-o/pinns yet, but by pinns.rs
+	/* if p.config.UsePodPID {
+		args = append(args, "--pid")
+	} */
+	if p.config.UsePodUTS {
+		args = append(args, "--uts")
+	}
+
+	if err := command.New(pinns, args...).RunSilentSuccess(); err != nil {
```
vrothberg: Can we use the stdlib instead?

vrothberg: Thinking about it: couldn't we do everything in a go package with C bindings? `github.com/containers/libpod/pkg/pinns`? That would be faster, easier to maintain (as most folks don't speak rust unfortunately), and there would be no need to package a new tool for all distributions.

haircommander: cgo isn't a great option, because the go runtime and locking threads is funky. We thought about this approach, but ultimately decided exec'ing the binary was a more guarenteed way of preventing an interrupt from changing the context too much. here's more info https://github.com/containernetworking/plugins/tree/master/pkg/ns

vrothberg: That's an omnipresent problem in go but locking the thread has proven to work. I didn't yet buy the idea of adding another dependency on top.

vrothberg: Just saw https://github.com/cri-o/cri-o/tree/master/pinns now. Easier than Rust. I try to catch up with the discussion in the CRI-O PR.

saschagrunert: Don’t take my Rust stuff too serious, we had a hackweek and I’m totally open to stick to the C version if that’s the preferred one. 🙂 

I was at the same point in thinking to unshare+pin as subcommand rather than having another binary, but now I see totally the arguments against something like that. 🤔

Which kinds of namespaces do we want to unshare beside NET, IPC and UTS? What about PID (dedicated feature called “pns“ in k8s), CGROUP and the USER (permissions) ns?

mheon:
I think we need to talk about cgroup... My initial impression is yes, though.
User and PID should be supported, but opt-in - I think you should have to explicitly tell `podman pod create` you want to share those.

haircommander:
> I think we need to talk about cgroup... My initial impression is yes, though.
> User and PID should be supported, but opt-in - I think you should have to explicitly tell `podman pod create` you want to share those.

the conversation of a pod level cgroupns hasn't been resolved in k8s, it's kind of been deferred until cgroupsv2 comes along more. As such, I could see it being useful to wire up the capability in podman so we don't have to add it later and fuss with different versions of pinns

rhatdan:
This seems to be a containers.conf decision, so Distros/Administrators can make the decision.

saschagrunert:
> This seems to be a containers.conf decision, so Distros/Administrators can make the decision.

I agree.

Do we want to put `pinns` into a dedicated project, I'd be happy to support you with that but I do not have the power :grin:

Maybe @vrothberg can support here as well? Are there any open discussion points regarding `pinns`?

vrothberg:
> Do we want to put `pinns` into a dedicated project, I'd be happy to support you with that but I do not have the power grin
> 
> Maybe @vrothberg can support here as well? Are there any open discussion points regarding `pinns`?

I would love to prevent us from having (yet) another dependency and project. They need to be setup, maintained and packaged; all causing additional work for upstream and downstream.

Hope dies last :) Why can't we have a go package with C bindings doing the work? @haircommander mentioned the common issues regarding cgo, thread locking in go etc. but is there a concrete technical reason/blocker not to go with go? If there's a clear blocker, then I stop nagging :^)

`pinns` doesn't do much and is really small, which would hurt even more to create a new project etc. In case we need to go with a dedicated binary, I prefer to move it over to libpod at some point. This way, we spare us from creating a new project, setting up CI, wiring that into libpod etc. With the CRI-O/libpod migration, I believe that libpod might be a better home for it in the long term.

rhatdan:
@saschagrunert @haircommander Any movement on this?

saschagrunert:
Closing this for now since I think main questions like how we want to distribute pinns are not solved yet.

PR #5211, opened 2020-02-14
increase stop timeout to 30 seconds

vrothberg:
Increase the timeout during container stop before shooting with SIGKILL.
Users have repeatedly reported that they are hitting the timeout, in
particular on small VMs with low resources.  Setting it to 30 seconds
will decrease the chance of hitting the timeout.

@mheon @giuseppe @baude @rhatdan ... this could be a controversial change as it also increases the time to stop misbehaving containers. These could still be silenced with `podman-kill` though.

### DIFF
--- a/libpod/define/config.go
+++ b/libpod/define/config.go
@@ -19,7 +19,7 @@ var (
 const (
 	// CtrRemoveTimeout is the default number of seconds to wait after stopping a container
 	// before sending the kill signal
-	CtrRemoveTimeout = 10
+	CtrRemoveTimeout = 30
 	// DefaultTransport is a prefix that we apply to an image name
 	// to check docker hub first for the image
 	DefaultTransport = "docker://"


mheon:
I think misbehaving containers are going to be a problem, here... A lot of
programs do ignore SIGTERM. And there's no timeout flag to podman rm, so
you'll end up waiting the full 30 seconds on a sleep container to rm -f.

This seems like a good candidate for containers.conf so we don't need to
unconditionally up the timeout.

vrothberg:
> This seems like a good candidate for containers.conf so we don't need to
unconditionally up the timeout.

That seems much better than increasing the timeout here. @rhatdan WDYT?

TomSweeneyRedHat:
I'm way in on @mheon's thought of moving this to the conf file.

rhatdan:
Fine with me.  Please open a PR with containers/common to add the field.  If we could ever figure out the one last blocker, we could get containers.conf merged into libpod.

PR #5216, opened 2020-02-14
add caching for binding tests

baude:
add the ability to cache images instead of pull them.  makes tests faster and less network use when we flip on CI.

Also added list images with filter test

### DIFF
--- a/pkg/bindings/test/common_test.go
+++ b/pkg/bindings/test/common_test.go
@@ -13,10 +13,30 @@ import (
 	"github.com/pkg/errors"
 )
 
+type testImage struct {
+	name        string
+	shortName   string
+	tarballName string
+}
+
 const (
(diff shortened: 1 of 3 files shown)


pkg/bindings/test/common_test.go
```diff
@@ -13,10 +13,30 @@ import (
...
+}
+
 const (
 	defaultPodmanBinaryLocation string = "/usr/bin/podman"
-	alpine                      string = "docker.io/library/alpine:latest"
-	busybox                     string = "docker.io/library/busybox:latest"
+)
+
+var (
+	ImageCacheDir = "/tmp/podman/imagecachedir"
```
jwhonce: Would this be better as os.TempDir("", "podman/imageCachedDir")

Issue #5087, opened 2020-02-04
label option, from libpod.conf, is not being respected.


fidencio:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When playing with podman + kata, there's the need to always pass `--security-opt label=disable` to `podman-run`. A similar way to achieve that, but for all the containers, would be setting `label = false` in the libpod.conf. However, it doesn't seem to work.

**Steps to reproduce the issue:**

On a Fedora 31 machine, using cgroups v1. do:
1. `dnf install kata-runtime` 
2. set `label=false` in the libpod.conf file
3.  `podman --runtime /usr/bin/kata-runtime run -it fedora /bin/bash`

**Describe the results you received:**
Error: rpc error: code = Unknown desc = selinux label is specified in config, but selinux is disabled or not supported: OCI runtime error

**Describe the results you expected:**
Container would be started in the same way as if started using the following command-line: `podman --runtime /usr/bin/kata-runtime run --security-opt label=disable fedora /bin/bash`

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 1.7.0
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.5
  podman version: 1.7.0
host:
  BuildahVersion: 1.12.0
  CgroupVersion: v1
  Conmon:
    package: conmon-2.0.9-2.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.9, commit: 7d46f3e7711aa3578488284ae2f98b447658f086'
  Distribution:
    distribution: fedora
    version: "31"
  MemFree: 5537505280
  MemTotal: 33534525440
  OCIRuntime:
    name: crun
    package: crun-0.10.6-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.10.6
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  SwapFree: 0
  SwapTotal: 0
  arch: amd64
  cpus: 12
  eventlogger: journald
  hostname: laerte
  kernel: 5.4.13-201.fc31.x86_64
  os: linux
  rootless: false
  uptime: 149h 40m 23.83s (Approximately 6.21 days)
registries:
  search:
  - docker.io
  - registry.fedoraproject.org
[...]

PR #5225, opened 2020-02-17
config: use built-in TOML merge and adhere to label setting

vrothberg:
Podman did not check the `label` setting in the libpod.conf and ignored if the libpod.conf set it to false to disable process labels in the container. While working on that, I found a bug in the config-merge code related to a false merging of booleans. To fix that, use the built-in features of the TOML library to merge/extend config instead of manually working around that.

Fixes: #5087


vrothberg:
@mheon @rhatdan PTAL

rhatdan:
I would rather get in containers.conf, but I have not figured out the last Test failure. 
LGTM

libpod/config/config.go
```diff
@@ -439,15 +437,11 @@ func probeConmon(conmonBinary string) error {
...
-	config := &Config{} // start with an empty config
-
-	// First, try to read the user-specified config
-	if userConfigPath != "" {
-		var err error
-		config, err = readConfigFromFile(userConfigPath)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error reading user config %q", userConfigPath)
-		}
+	// Start with the default config and interatively merge fields in the system
```
TomSweeneyRedHat: nit interatively -> iteratively

libpod/config/config.go
```diff
@@ -456,44 +450,45 @@ func NewConfig(userConfigPath string) (*Config, error) {
...
 	}
 
-	// Check if we need to switch to cgroupfs and logger=file on rootless.
-	defaultConfig.checkCgroupsAndLogger()
-
-	if err := config.mergeConfig(defaultConfig); err != nil {
-		return nil, errors.Wrapf(err, "error merging default config from memory")
+	// Since runc does not currently support cgroupV2
+	// Change to default crun on first running of libpod.conf
+	// TODO Once runc has support for cgroups, this function should be removed.
```
TomSweeneyRedHat: I wonder if we should make a backlog jira card to help to remember to remove this.

TomSweeneyRedHat:
LGTM

pkg/spec/security.go
```diff
@@ -85,6 +92,10 @@ func (c *SecurityConfig) SetSecurityOpts(runtime *libpod.Runtime, securityOpts [
 		}
 	}
 
+	if !labelSet && !libpodConfig.EnableLabeling {
```
mheon: It feels like we should be handling this in Libpod - It's the libpod config we're using to set this.

vrothberg: Any idea how to detect if an item in `LabelOpts` was set by the user. I moved it here, because we have the info at hand but in libpod we had to be smarter.

mheon: If `--label` is set while the command line says labeling is disabled, IMO that's an error - SELinux has been disabled, so you can't configure it.

@rhatdan WDYT?

vrothberg: I can fix that, then it's straight forward to detect with a simple length check in libpod :+1:

rhatdan: Sounds reasonable to me.

vrothberg: I begin to believe that `pkg/spec` is the right place for it. We are querying the config there already to lookup certain options. We don't handle the labels really in `./libpod`.

vrothberg:
@rhatdan @mheon please take another look. I moved the non-CLI label setting into `pkg/spec` which to me seems the right place. If you want it somewhere else, please let me know where and I can move it.

mheon:
My thinking was mostly that, if I set `libpod.conf` to disable labelling, we should not use SELinux, no matter where the container originated - so for APIv2 containers, we should not use SELinux even if the remote host requests it.

vrothberg:
> My thinking was mostly that, if I set `libpod.conf` to disable labelling, we should not use SELinux, no matter where the container originated - so for APIv2 containers, we should not use SELinux even if the remote host requests it.

I don't know any other flag where we're doing that as the CLI always (?) wins. Especially in case of booleans we had to distinguish between unset, set and unspecified in the configurations to know if the admin cares and compare that with the CLI but we don't have that.

mheon:
I view this less as a default setting for the label flag and more as a configuration bool entirely disabling labelling. I'm not too firm on this, but it might help to have a tiebreaker here. @rhatdan @baude Thoughts?

rhatdan:
I disagree, I might want to run the bulk of my containers without SELinux separation, but a couple of untrusted ones, I want to run with SELinux.

rhatdan:
LGTM

vrothberg:
Let me add a commit on top to clarify that in the libpod.conf. The man page is not strictly clear on that and I can understand Matt's idea as well.

vrothberg:
Done :heavy_check_mark:

PR #5237, opened 2020-02-17
rootless: check if the conmon process is valid

giuseppe:
if the pause process doesn't exist and  we try to join a conmon
namespace, make sure the process still exists.  Otherwise re-create
the user namespace.


mheon:
LGTM

PR #5245, opened 2020-02-18
rootless: fix a regression when using -d

giuseppe:
when using -d and port mapping, make sure the correct fd is injected
into conmon.

Move the pipe creation earlier as the fd must be known at the time we
create the container through conmon.

Closes: https://github.com/containers/libpod/issues/5167

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -335,10 +335,13 @@ func (r *Runtime) setupRootlessPortMapping(ctr *Container, netnsPath string) (er
 		return errors.Wrapf(err, "delete file %s", logPath)
 	}
 
-	ctr.rootlessPortSyncR, ctr.rootlessPortSyncW, err = os.Pipe()
-	if err != nil {
-		return errors.Wrapf(err, "failed to create rootless port sync pipe")
+	if !ctr.config.PostConfigureNetNS {
+		ctr.rootlessPortSyncR, ctr.rootlessPortSyncW, err = os.Pipe()
+		if err != nil {
+			return errors.Wrapf(err, "failed to create rootless port sync pipe")
+		}
 	}
+
 	cfg := rootlessport.Config{
 		Mappings:  ctr.config.PortMappings,
 		NetNSPath: netnsPath,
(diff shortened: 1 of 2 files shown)


libpod/oci_conmon_linux.go
```diff
@@ -1160,6 +1160,14 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
...
 
 	if ctr.config.NetMode.IsSlirp4netns() {
+		havePortMapping := len(ctr.Config().PortMappings) > 0
+		if havePortMapping {
+			ctr.rootlessPortSyncR, ctr.rootlessPortSyncW, err = os.Pipe()
+			if err != nil {
+				return errors.Wrapf(err, "failed to create rootless port sync pipe")
+			}
+		}
+
```
TomSweeneyRedHat: should you do the `defer errorhandling` for rootlessPortSyncR/W here as you do below for rootlessSlipSyncR/W around line 1177?

libpod/networking_linux.go
```diff
@@ -355,6 +351,11 @@ func (r *Runtime) setupRootlessPortMapping(ctr *Container, netnsPath string) (er
 	cmd := exec.Command(fmt.Sprintf("/proc/%d/exe", os.Getpid()))
 	cmd.Args = []string{rootlessport.ReexecKey}
 	// Leak one end of the pipe in rootlessport process, the other will be sent to conmon
+
+	if ctr.rootlessPortSyncR != nil {
+		defer errorhandling.CloseQuiet(ctr.rootlessPortSyncR)
+	}
+
```
TomSweeneyRedHat: Do you need a similar for rootlessSlirpSyncR?

libpod/oci_conmon_linux.go
giuseppe: no, we cannot close `rootlessPortSyncR` here, it must be left open as it used later

libpod/networking_linux.go
giuseppe: do you mean `rootlessSlirpSyncW`?  No, it is already closed in `libpod/oci_conmon_linux.go`

rhatdan:
LGTM
but tests are not happy.

giuseppe:
tests are passing now

PR #5248, opened 2020-02-18
Add installation of pre-commit to Makefile

rhatdan:
pre-commit is required for running certain tests by the Makefile.
Adding pre-commit to install.tools allows users to easily use
and test podman.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ endif
 lint: golangci-lint
 	@echo "Linting vs commit '$(call err_if_empty,EPOCH_TEST_COMMIT)'"
 ifeq ($(PRE_COMMIT),)
-	@echo "FATAL: pre-commit was not found, check https://pre-commit.com/ about installing it." >&2
+	@echo "FATAL: pre-commit was not found, make .install.pre-commit to installing it." >&2
 	@exit 2
 endif
 	$(PRE_COMMIT) run -a
@@ -561,6 +561,12 @@ endef
 .PHONY: .install.golangci-lint
 .install.golangci-lint: .gopathok
 	VERSION=1.18.0 GOBIN=$(GOBIN) sh ./hack/install_golangci.sh
+
+.PHONY: .install.pre-commit
+.install.pre-commit:
+	if [ -z "$(PRE_COMMIT)" ]; then \
+		python3 -m pip install --user pre-commit; \
+	fi
 
 .PHONY: .install.md2man
 .install.md2man: .gopathok


rhatdan:
Fixes: https://github.com/containers/libpod/issues/5081

rhatdan:
@baude @mheon @vrothberg @giuseppe @TomSweeneyRedHat @edsantiago PTAL

Makefile
```diff
@@ -562,6 +562,12 @@ endef
 .install.golangci-lint: .gopathok
 	VERSION=1.18.0 GOBIN=$(GOBIN) sh ./hack/install_golangci.sh
 
+.PHONY: .install.pre-commit
+.install.pre-commit:
+	if [ ! -x "$(PRE_COMMIT)" ]; then \
```
edsantiago: I think you want something like this instead:
```bash
    if [ -z "$(PRE_COMMIT)" ]; then \
```
Reason: PRE_COMMIT is defined on line 37 as follows:
```make
```
Thus the case where it's not installed yields an empty string

edsantiago:
Oh dear. That CI failure does not look easily solved.

rhatdan:
@cevich How do I get python-pip installed into the Cirrus VMs?

edsantiago:
In Chris's absence, I'll chime in but possibly unhelpfully: instead of jumping through these hoops, would it make sense instead to leave `.install.pre-commit` out of `.install.tools`, and just change the lint error message to tell users to run that target?

edsantiago:
Looks like a flake

/retest

edsantiago:
```
 error: could not run steps: step centos-install failed: could not wait for build: the build centos-install failed after 4m27s with reason PushImageToRegistryFailed: Failed to push the image to the registry.
Registry server Address: 
Registry server User Name: serviceaccount
Registry server Email: serviceaccount@example.org
Registry server Password: <<non-empty>>
error: build error: Failed to push image: blob upload unknown
```
Still looks like a flake
/retest

rhatdan:
@mheon @baude @vrothberg @giuseppe @QiWang19 @jwhonce @TomSweeneyRedHat Can we get this one in please. Ready to merge.

cevich:
> In Chris's absence, I'll chime in but possibly unhelpfully: instead of jumping through these hoops, would

Thanks @edsantiago ya, I'm nervous of using `pip` in CI, since it bypasses the native package system.  Unless it's run inside a python virtualenv, using `pip` system-wide is a "no-no" in my book.  It can really confuse all questions of "which code am I running"?

PR #5253, opened 2020-02-18
Update to the latest version of buildah

rhatdan:


rhatdan:
@mheon @TomSweeneyRedHat @giuseppe @vrothberg @baude @jwhonce @QiWang19 PTAL
Looks like we can update buildah without the containers.conf  Merging this should help containers.conf get in.  I believe.

cmd/podman/build.go
```diff
@@ -84,7 +84,10 @@ func init() {
 	}
 
 	flag.DefValue = "true"
-	fromAndBugFlags := buildahcli.GetFromAndBudFlags(&fromAndBudValues, &userNSValues, &namespaceValues)
+	fromAndBugFlags, err := buildahcli.GetFromAndBudFlags(&fromAndBudValues, &userNSValues, &namespaceValues)
+	if err != nil {
+		logrus.Error("failed to setup podman build flags")
```
mheon: Can we include the error in here?

go.mod
```diff
@@ -4,19 +4,16 @@ go 1.12
...
 require (
 	github.com/BurntSushi/toml v0.3.1
-	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/buger/goterm v0.0.0-20181115115552-c206103e1f37
 	github.com/checkpoint-restore/go-criu v0.0.0-20190109184317-bdb7599cd87b
-	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/containernetworking/cni v0.7.2-0.20190904153231-83439463f784
 	github.com/containernetworking/plugins v0.8.5
-	github.com/containers/buildah v1.13.1
+	github.com/containers/buildah v1.14.1-0.20200219125159-7cd6f7d04842
```
TomSweeneyRedHat: Did you spin up a 1.14.1 somewhere?  The  highest release I see is 1.14.0.

TomSweeneyRedHat:
My only concern is where Buildah 1.14.1 is coming from, other than that and @mheon's request, LGTM

go.mod
rhatdan: Nope this comes from the master branch.  I have no idea where it comes from.

cmd/podman/build.go
rhatdan: Fixed

go.mod
rhatdan: I think this is being generated by go.mod somehow.  The go.mod that I added was from the master branch.  We will need to cut a release before podman gets its next release.  But this should be fine for now.

$ ./bin/podman info  | grep -i buildah
BuildahVersion: 1.15.0-dev

TomSweeneyRedHat:
@rhatdan thx.  I'm fine with that for now, but before we cut Podman.next, I definitely will want to get a new Buildah release together and vendored.

rhatdan:
@TomSweeneyRedHat Agreed.  We should have @edsantiago add a test to make sure their is never the letters `dev` in the `podman info | grep -i buildah` before we release a new podman.  Don't want that gating PRs but only releases.

rhatdan:
@mheon @giuseppe @baude @vrothberg @jwhonce @QiWang19 PTAL
Could someone merge this so that we can rebase and see where we are with the greater containers.conf PR.

Issue #5131, opened 2020-02-07
podman logs crash


yangm97:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Sometimes, `podman logs` crashes saying it has run out of memory. Looks like it's trying to load the whole file in RAM, and crashes when it doesn't fit. 

**Steps to reproduce the issue:**

1. Have a container to generate a log file that is bigger than the available RAM
2. `# podman logs -f --tail=10 some-container`

**Describe the results you received:**

```
fatal error: runtime: out of memory

runtime stack:
runtime.throw(0x1184bbe, 0x16)
        /usr/lib/go-1.10/src/runtime/panic.go:616 +0x60
runtime.sysMap(0x1f990000, 0x17b10000, 0xbee34401, 0x1e21328)
        /usr/lib/go-1.10/src/runtime/mem_linux.go:227 +0x11c
runtime.(*mheap).sysAlloc(0x1e13520, 0x17b10000, 0x2)
        /usr/lib/go-1.10/src/runtime/malloc.go:470 +0x23c
runtime.(*mheap).grow(0x1e13520, 0xbd88, 0x0)
        /usr/lib/go-1.10/src/runtime/mheap.go:907 +0x48
runtime.(*mheap).allocSpanLocked(0x1e13520, 0xbd85, 0x1e21338, 0xfffffade)
        /usr/lib/go-1.10/src/runtime/mheap.go:820 +0x30c
runtime.(*mheap).alloc_m(0x1e13520, 0xbd85, 0x101, 0x1)
        /usr/lib/go-1.10/src/runtime/mheap.go:686 +0x118
runtime.(*mheap).alloc.func1()
        /usr/lib/go-1.10/src/runtime/mheap.go:753 +0x3c
runtime.(*mheap).alloc(0x1e13520, 0xbd85, 0x13010101, 0x0)
        /usr/lib/go-1.10/src/runtime/mheap.go:752 +0x60
runtime.largeAlloc(0x17b081e0, 0x13740101, 0x137000e0)
        /usr/lib/go-1.10/src/runtime/malloc.go:826 +0x74
runtime.mallocgc.func1()
        /usr/lib/go-1.10/src/runtime/malloc.go:721 +0x38
runtime.systemstack(0x0)
        /usr/lib/go-1.10/src/runtime/asm_arm.s:349 +0x80
runtime.mstart()
[...]

PR #5263, opened 2020-02-19
implement reverse reader for log reads

baude:
in cases where the log file exceeds the available memory of a system, we had a bug that triggered an oom because the entire logfile was being read when the tail parameter was given.  this reads in chunks and is more or less memory safe.

fixes: #5131

### DIFF
--- a/hack/get_ci_vm.sh
+++ b/hack/get_ci_vm.sh
@@ -97,7 +97,7 @@ keys=[k for k in env if "ENCRYPTED" not in str(env[k])]
 for k,v in env.items():
     v=str(v)
     if "ENCRYPTED" not in v:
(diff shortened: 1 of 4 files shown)


rhatdan:
@yangm97 PTAL
LGTM
Does this do anything for when I want to read an entire log?  Do we buffer that up now?

yangm97:
> Does this do anything for when I want to read an entire log? 

Or tailing more lines than it fits on the memory. I haven't tried it yet but if I understood the code correctly, it does seem to work under that assumption.

baude:
> @yangm97 PTAL
> LGTM
> Does this do anything for when I want to read an entire log? Do we buffer that up now?

reading the entire log is done differently.  not applicable

Issue #5255, opened 2020-02-19
podman exec inherits environment variables from container process


jmou:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman exec` does not use the environment variables that the container image is built with. Instead it seems to inherit environment variables from the container process.

**Steps to reproduce the issue:**

1. Run
```
$ podman run --rm -d --name execenv debian env x=y PATH=/bin sleep 10
eefc809c55a0fb0c877f3ef331b51b4d0091dd7d5e4f8445854d0f21d45a756b
$ podman exec execenv env
```

**Describe the results you received:**

```
$ podman exec execenv env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
TERM=xterm
container=podman
HOSTNAME=
HOME=/root
x=y
```

Note $x is set and $PATH gets reset.

**Describe the results you expected:**

Compare to docker
```
$ sudo docker run --rm -d --name execenv debian env x=y PATH=/bin sleep 10
b4c0ef743608691c9f6094b68f42b6df3da822e5e1565258bf5d1756c94b5c24
$ sudo docker exec execenv env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=b4c0ef743608
HOME=/root
```

$x is unset and $PATH has a default value.

**Additional information you deem important (e.g. issue happens only occasionally):**

I am running into this issue while using containers that run processes inside Python virtualenvs. In particular, because `podman exec` inherits VIRTUAL_ENV and other variables, but PATH is reset without the virtualenv path, I end up half-in/half-out of the virtualenv.

**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.6
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.6
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.10-2.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.10, commit: 6b526d9888abb86b9e7de7dfdeec0da98ad32ee0'
  Distribution:
    distribution: fedora
    version: "31"
  IDMappings:
    gidmap:
[...]

PR #5264, opened 2020-02-19
WIP - consolidate env handling into pkg/env

vrothberg:
Env-variable related code is scattered across several packages making it
hard to maintain and extend.  Consolidate the code into a new pkg/env
package.

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -5,6 +5,7 @@ package libpod
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/binary"
 	"fmt"
 	"io"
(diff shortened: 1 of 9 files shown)


vrothberg:
Still want to move the env-parsing code into pkg/env but for now I want to throw CI at it.

vrothberg:
Fixes: #5255

libpod/oci_conmon_linux.go
```diff
@@ -1274,15 +1280,26 @@ func prepareProcessExec(c *Container, cmd, env []string, tty bool, cwd, user, se
...
 
-	hasHomeSet := false
-	for _, s := range pspec.Env {
-		if strings.HasPrefix(s, "HOME=") {
-			hasHomeSet = true
-			break
-		}
+	// We need to start with a fresh environment and should not inherit from the
+	// container's config (https://github.com/containers/libpod/issues/5255).
+	imageData, err := getImageDataFromContainer(c)
```
mheon: Do we need this? We have the full list of environment variables the container was created from in the OCI spec.

vrothberg: According to #5255 we need that to match Docker which always starts with the default + from the image.

mheon: We don't want to use the extra environment variables set by the container process, but we do want those set at container creation time.

Tested with:
Terminal 1 - `sudo docker run -t -i --rm --name test --env a=b fedora bash`
Terminal 2 - `sudo docker exec -t -i test env`

I see `a=b` in the output of exec - so we should be including everything from container create time, but nothing that was added after the container was run.

mheon:
Note to self: since Valentin is off on vacation after today, I'll take this one and finish it up.

mheon:
Closing in favor of #5270

Issue #4283, opened 2019-10-17
test: exercise podman-login + skopeo


vrothberg:
We've recently had a silent regression where podman-login used another path for storing the auth file than the defaults of c/image do. This broke Skopeo CI which is using podman-login. To prevent us from running into similar regressions in the future, we could extend our e2e tests to perform a podman-login, followed by, for instance, a skopeo-copy.

rhatdan:
@edsantiago were you every able to generate a tests for this?

edsantiago:
Oops, no - I dropped the ball on this. In fact there aren't currently _any_ `login` tests in the podman system test suite; I've never been sure how to fit those in. I'll look into it today.

rhatdan:
Excellent.

edsantiago:
I'm very close to having decent tests but have run into a snag: it seems impossible to know the path podman will use to an auth file. I'm seeing all sorts of variations like:
```
/run/containers/0/auth.json
/run/user/0/containers/auth.json
/run/user/1000/containers/auth.json
/var/tmp/run-1000/containers/auth.json
```
If I grok the code correctly, it's `vendor/github.com/containers/image/v5/pkg/docker/config/config.go:getPathToAuth()` which tries all sorts of variations depending on whether `$XDG_RUNTIME_DIR` is set or not, whether it's writable or not, and the phase of the moon. 

I need to have the path to the auth file, because I need to test what podman does with missing/corrupt files and with invalid credentials. I do not want to recreate the complicated logic in the above module. I would love to add a `podman info` field to show this path, but `getPathToAuth()` is not exported.

Is there something simple I'm missing?

rhatdan:
```
man podman login
...
       Note:  You  can  also override the default path of the authentication file by setting the REGISTRY_AUTH_FILE
       environment variable. export REGISTRY_AUTH_FILE=path
```
You can hard code it to the specific file you want.

edsantiago:
I know that - I deliberately didn't test that (yet) because it wouldn't really help with the desired skopeo case. I want to test that podman's default path works with skopeo by default. Sorry for not being clear.

edsantiago:
...but that's a good idea: the skopeo-podman-default tests don't actually need to know the auth file path; and the tests that do, can simply set it explicitly. I'll try that tomorrow if there are no other suggestions. Thanks.

edsantiago:
[...]

PR #5265, opened 2020-02-19
New login and push tests

edsantiago:
Test podman login/logout, login with wrong credentials,
auth file contents, auth file path override, push/pull,
and, if skopeo is installed, credentials sharing

Fixes: #4283

### DIFF
--- a/test/system/150-login.bats
+++ b/test/system/150-login.bats
@@ -1,0 +1,339 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# tests for podman login
+#
+
+load helpers
+
+###############################################################################
+# BEGIN one-time envariable setup
+
+# Create a scratch directory; our podman registry will run from here. We
+# also use it for other temporary files like authfiles.
+if [ -z "${PODMAN_LOGIN_WORKDIR}" ]; then
+    export PODMAN_LOGIN_WORKDIR=$(mktemp -d --tmpdir=${BATS_TMPDIR:-${TMPDIR:-/tmp}} podman_bats_login.XXXXXX)
+fi
+
+# Randomly-generated username and password
+if [ -z "${PODMAN_LOGIN_USER}" ]; then
+    export PODMAN_LOGIN_USER="user$(random_string 4)"
+    export PODMAN_LOGIN_PASS=$(random_string 15)
+fi
+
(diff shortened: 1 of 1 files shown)


mheon:
I'm going to try this, but Github may be busted enough it doesn't work
/approve

edsantiago:
New force-push: disabled all tests under podman-remote.

IMPORTANT NOTE: `skopeo` is not installed in the CI environment, so the skopeo-podman auth-sharing tests are being skipped. I don't know how to update CI setups and don't dare do so without Chris.

test/system/150-login.bats
```diff
@@ -0,0 +1,339 @@
...
+    skip "skopeo does not yet support REGISTRY_AUTH_FILE; #822"
+    authfile=${PODMAN_LOGIN_WORKDIR}/auth-$(random_string 10).json
+    rm -f $authfile
+
+    REGISTRY_AUTH_FILE=$authfile _test_skopeo_credential_sharing
+    rm -f $authfile
+}
+
+@test "podman login - shares credentials with skopeo - via --authfile" {
+    # Also test that command-line --authfile overrides envariable
```
TomSweeneyRedHat: should you skip this one right off the bat too?

TomSweeneyRedHat:
One question for consideration, LGTM.

test/system/150-login.bats
edsantiago: Haha! No, that's actually deliberate: since `--authfile=/path` should always override environment, `$REGISTRY_AUTH_FILE` should be ignored by both podman and skopeo - hence it shouldn't matter whether or not skopeo implements that or not. And if/when skopeo _does_ implement it, if for any reason it gets implemented wrong such that environment overrides command-line, this test will catch it.

Does that make sense? I had hoped that the test comment would make it clear without having to go into subtleties, but should I update the comment, and if so, can you suggest an improvement?

TomSweeneyRedHat: no worries, just me apparently reviewing too quickly.

rhatdan:
Nice job.  When @cevich gets back we can work to get skopeo installed in AMIs.
/lgtm

cevich:
> we can work to get skopeo installed in AMIs.

Got it added into my #3901 where I need to rebuild the VM images anyway to add dnsmasq.

PR #5266, opened 2020-02-19
[WIP] Add manifest subcommand

sjug:
Working towards https://github.com/containers/libpod/issues/713.
Requires https://github.com/containers/libpod/pull/5253.

Took a swing at breaking out the manifests monolith from buildah.
Quite a bit different implementation of the commands compared to podman `GetRuntime()` etc.

How much did you want to bring over from buildah vs refactor?
Do we want to lift `getStore()` from buildah to make it work as is?

I would appreciate some suggestions.

### DIFF
--- a/cmd/podman/manifest_annotate.go
+++ b/cmd/podman/manifest_annotate.go
@@ -5,10 +5,10 @@ import (
 	"strings"
 
 	"github.com/containers/buildah/manifests"
-	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/libpod/cmd/podman/cliconfig"
+	"github.com/containers/libpod/cmd/podman/libpodruntime"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -19,7 +19,7 @@ var (
 	manifestAnnotateDescription = `adds or updates information about an entry in a manifest list or image index`
 	_manifestAnnotateCommand    = &cobra.Command{
 		Use:   "annotate [flags] [manifest] [tags]",
-		Short: "Add or update information about an entry in a manifest list or image index",
+		Short: "manifest annotate",
 		Long:  manifestAnnotateDescription,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			manifestAnnotateCommand.InputArgs = args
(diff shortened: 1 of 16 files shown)


mheon:
We can't use the Buildah store methods directly as they won't respect libpod.conf; will probably need to move some of this into pkg/image

rhatdan:
@sjug First you need to sign your PRs to make the tests run.
It is fine to share as code with Buildah,  we already do this for podman build and option parsing.
The question is, could we do the same for manifest options, so we don't end up with the options in two places.
Is there anything we could do with Buildah to make it easier to vendor manifests into podman.

sjug:
I've kept the second commit to show how I replaced some key different functions from buildah to see if that's okay. Also the third commit is adding in the buildah/manifest deps.

I'm not sure why I'm failing gating?

cmd/podman/manifest_annotate.go
```diff
@@ -0,0 +1,166 @@
...
+	runtime, err := libpodruntime.GetRuntime(getContext(), &c.PodmanCommand)
+	if err != nil {
+		return errors.Wrapf(err, "could not create runtime")
+	}
+	defer runtime.DeferredShutdown(false)
+
+	store := runtime.GetStore()
+	systemContext := runtime.SystemContext()
+
+	_, listImage, err := util.FindImage(store, "", systemContext, listImageSpec)
```
baude: we typically would put this content and below into the adapter layer.  I think given your not familar with that and we are actively working on it, I'm OK with ignoring that for now and allowing us to clean this up.

baude: we typically use something like runtime.LookupImage (or something close to that.)

cmd/podman/manifest_annotate.go
```diff
@@ -0,0 +1,166 @@
+package main
```
baude: this command will need to fenced off as a local-only command until we have apiv2 endpoints for this.  maybe that should get done now? so that this is a slightly cleaner implementation?

baude:
needs man pages and bash completion additions to pass gating.

rhatdan:
@vrothberg @sjug We are working on APIV2 for this first and setting up support for it in libpod.  Once this happens we can work on the CLI for it.  @baude Opened the first PR for APIV2 yesterday.

rhatdan:
@sjug This PR is replaced with https://github.com/containers/libpod/pull/5843

Issue #5238, opened 2020-02-17
swagger: possible inconsistencies


edsantiago:
These are the inconsistencies my script sees in the `// swagger:operation` comments. I've organized them by what seems to me like a common grouping, and hope that it makes sense. Not all of these may be real issues, they're just confusing.

All files are under `pkg/api/server`. In some places I have adjusted whitespace for legibility.

### Missing `libpod` prefix

The following names in `register_containers.go` are all missing the usual `libpod` prefix:

```
  // swagger:operation GET  /libpod/containers/showmounted   libpod showMountedContainers
  // swagger:operation POST /libpod/containers/{name}/mount  libpod mountContainer
  // swagger:operation GET  /libpod/containers/{name}/exists libpod containerExists
```
(i.e. they should maybe be `libpodShowMountedContainers`, `libpodMountContainer`, `libpodContainerExists` for consistency with the 17 other paths defined here)

### Incomplete names

The following names in `register_containers.go` should maybe include a `Container` suffix?
```
  // swagger:operation POST /libpod/containers/{name}/attach libpod libpodAttach
  // swagger:operation POST /libpod/containers/{name}/resize libpod libpodResize
```
The 18 other paths all include `container` or `Container`.

### Inconsistent names (do not match the actual code).

For some reason I can't understand, `register_images.go` has a remarkable number of inconsistencies between the name on the right-hand side of the swagger line (operation ID) and the actual function being used. These inconsistencies probably aren't harmful, but they certainly make it difficult for a human to understand the relationships:
```
  // swagger:operation POST /images/create compat createImage
  r.Handle(VersionedPath("/images/create"), ... generic.CreateImageFromImage

  // swagger:operation GET /images/json compat listImages
  r.Handle(VersionedPath("/images/json"), ... generic.GetImages

  // swagger:operation POST /images/load compat importImage
  r.Handle(VersionedPath("/images/load"), ... generic.LoadImages

  // swagger:operation GET /images/{name}/history compat imageHistory
  r.Handle(VersionedPath("/images/{name}/history"), ... handlers.HistoryImage

  // swagger:operation GET /images/{name}/json compat inspectImage
  r.Handle(VersionedPath("/images/{name}/json"), ... generic.GetImage

  // swagger:operation GET /libpod/images/{name}/history libpod libpodImageHistory
  r.Handle(VersionedPath("/libpod/images/{name}/history"), ... handlers.HistoryImage

  // swagger:operation GET /libpod/images/json libpod libpodListImages
[...]

PR #5235, opened 2020-02-17
Update mux rules to allow slashes in image names

jwhonce:
%2F escaping is also supported.

Fixes #5229
Fixes #5228

Signed-off-by: Jhon Honce <jhonce@redhat.com>

PR #5267, opened 2020-02-19
Swagger: fix one incorrect comment

edsantiago:
PR #5235 updated a bunch of entrypoints from {name} to {name:.*},
requiring matching changes to swagger comments. Looks like one
got missed due to the manual nature of this work.

Have I mentioned lately that manual maintenance of duplicate
information is not a great idea?

Discrepancy caught by my script, which I would really like to
look into getting into CI, but is gating on #5238. (I would
actually not like to get this script into CI, I would prefer
to have the duplicate information be autogenerated from
the function calls themselves, but I seem to have lost
that battle)

### DIFF
--- a/pkg/api/server/register_images.go
+++ b/pkg/api/server/register_images.go
@@ -822,7 +822,7 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//   500:
 	//      $ref: '#/responses/InternalError'
 	r.Handle(VersionedPath("/libpod/images/search"), s.APIHandler(handlers.SearchImages)).Methods(http.MethodGet)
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

edsantiago:
```
[+0031s] - path param "{name:.*}" has no parameter definition
[+0031s] - path param "name" is not present in path "/libpod/images/{name:.*}"
```
I'm stumped. It looks exactly the same as the `/images/{name:.*}` change. My consistency-checking script finds it OK. @jwhonce @baude I can only assume that there's some magic setting somewhere else that I've missed.

edsantiago:
Ok, fixed. Silly me: I missed that one field here isn't duplicated, it's triplicated.

PR #5268, opened 2020-02-19
Warn user about --password cli option in login

Akasurde:
Using command line parameter --password may reveal registry password. Better to warn user about this.

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/cmd/podman/login.go
+++ b/cmd/podman/login.go
@@ -82,6 +82,10 @@ func loginCmd(c *cliconfig.LoginValues) error {
 		server = registryFromFullName(scrubServer(args[0]))
 	}
 
+	if c.Flag("password").Changed {
+		fmt.Fprintf(os.Stderr, "WARNING! Using --password via the cli is insecure. Please consider using --password-stdin\n")
+	}
+
 	sc := image.GetSystemContext("", c.Authfile, false)
 	if c.Flag("tls-verify").Changed {
 		sc.DockerInsecureSkipTLSVerify = types.NewOptionalBool(!c.TlsVerify)


cmd/podman/login.go
```diff
@@ -82,6 +82,10 @@ func loginCmd(c *cliconfig.LoginValues) error {
 		server = registryFromFullName(scrubServer(args[0]))
 	}
 
+	if c.Flag("password").Changed {
+		fmt.Println("WARNING! Using --password via the cli is insecure. Please consider using --password-stdin")
```
mheon: Do we want to send this to STDERR? Maybe use `logrus.Errorf`?

Akasurde: I am not sure if this needs to be sent to stderr since this is a warning.

rhatdan: Yes it should use stderr, for output.  Not sure about logrus.Error though since that prints out ERROR, I believe, and this really is not an error.

Akasurde: Done.

rhatdan:
LGTM

Akasurde:
@mheon @rhatdan Thanks for the review and merge.

edsantiago:
Day late dollar short, but would you consider opening a new PR using `logrus.Warn()` instead of printf?

mheon:
Warnf isn't printed by default, which could be a problem

edsantiago:
>Warnf isn't printed by default

:gaping-mouth-agog-emoji: I didn't realize that. Um, ok, never mind. Thanks for the quick response.

mheon:
I kind of want to change that, but we have a bunch of warnings that will show by default now. We'd need to do a pass over all our logging and determine what severity things should be (and then try and fix any warnings that were still being displayed)

rhatdan:
I agree, but I believe these warnings run deep.

PR #5270, opened 2020-02-19
Consolidate env handling into pkg/env

mheon:
Taking this one over from @vrothberg so he can enjoy his vacation

Original description from #5264 -
Env-variable related code is scattered across several packages making it hard to maintain and extend.  Consolidate the code into a new pkg/env package.

I modified this a bit to remove the changes to oci_conmon_linux.go as the that code tried to fix was actually a crun issue.


pkg/env/env.go
```diff
@@ -0,0 +1,123 @@
...
+				if strings.HasPrefix(part[0], name) {
+					env[part[0]] = part[1]
+				}
+			}
+		} else if val, ok := os.LookupEnv(name); ok {
+			// if only a pass-through variable is given, clean it up.
+			env[name] = val
+		}
+	}
+	return nil
```
TomSweeneyRedHat: Do we need to check in here if we have more than one equal sign?

mheon: The SplitN() call above should take care of that - splits only once

TomSweeneyRedHat: Oh man, tired reviewing eyes, thought that was Split.  Thanks

TomSweeneyRedHat:
LGTM assuming happy tests

mheon:
Rebased and repushed with fixes

edsantiago:
Weird. One of the flakes is in `Cirrus CI / special_testing_rootless TEST_REMOTE_CLIENT:false`, in `podman run --host-env environment test`:
```
test/e2e/run_test.go:253

Running: podman [options] --events-backend file --storage-driver vfs run --rm --env-host docker.io/library/alpine:latest printenv FOO
Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
```
It seems to fail three times in a row. I can't reproduce it, not on f30 or f30, not root or rootless. not with multiple settings of my `$PATH`. I can't tell from the CI name if this is f30 or f31.

edsantiago:
@mheon @baude here's a smoking-gun [failure of the duplicate-ip test](https://api.cirrus-ci.com/v1/task/6347871272304640/logs/integration_test.log#t--Podman-run-two-containers-with-the-same-IP) showing that container 2 really is running and has the expected IP as container 1.

Other failures in that log are the same printenv PATH flate.

edsantiago:
Whoa! I'm seeing two different failure messages here, on two different runs!

[First one](https://api.cirrus-ci.com/v1/task/6629346249015296/logs/integration_test.log#t--podman-run---host-env-environment-test)
```
Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
```
[Second](https://api.cirrus-ci.com/v1/task/6347871272304640/logs/integration_test.log#t--podman-run---host-env-environment-test)
```
Error: container_linux.go:349: starting container process caused "exec: \"printenv\": executable file not found in $PATH": OCI runtime command not found error
```
The second one (properly and helpfully) includes `printenv` as the not-found command. Could this be a crun vs runc difference? Can we make them *both* emit the proper form? And, even better, can we include the expansion of `$PATH` in all such messages?

mheon:
@edsantiago I don't know if the env-host stuff is a flake - this PR does refactor that code. I can't replicate locally, so I've added a commit to see if something wierd is happening with $PATH.

mheon:
As for the two errors... I bet that's crun vs runc (F30 vs F31)

mheon:
Now, the real interesting part - it succeeds on Ubuntu.

mheon:
`[+0494s] /root/bin:/var/tmp/go/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/var/tmp/go/bin`

That doesn't look to include `/bin` which is the cause of the issue. It looks like we're somehow copying in $PATH from the host - which `--env-host` should not do (and does not do on my system).

mheon:
Alright, I'm giving up for now... this one makes no sense. I'll tag @vrothberg to take a look once he's back from vacation, see what he makes of this.

vrothberg:
Having a look now.

vrothberg:
I can reproduce locally.

Issue #5256, opened 2020-02-19
/etc/resolv.conf with dnsname plugin and user namespace not using dnsname resolver


ykuksenko:
/kind bug

**Description**

When running a container that is using a user namespace attached to a non default network with dnsname plugin enabled /etc/resolv.conf contains the hosts resolvers instead of the dnsmasq resolver for the network.

**Steps to reproduce the issue:**
```
[root@f31 ~]# cat /etc/resolv.conf
# Generated by NetworkManager
search example.org
nameserver 192.168.0.200
[root@f31 ~]# useradd woot
[root@f31 ~]# podman network create woot
/etc/cni/net.d/woot.conflist
[root@f31 ~]# podman run --rm -ti -p 2000:2000 --network=woot library/alpine cat /etc/resolv.conf
search example.org
nameserver 10.89.0.1
[root@f31 ~]# podman run --rm -ti -p 2000:2000 --subuidname=woot --subgidname=woot --network=woot library/alpine cat /etc/resolv.conf
search example.org
nameserver 192.168.0.200
```

**Describe the results you received:**
The output for /etc/resolv.conf for the two runs was different, first producing the dnsmasq resolver and the second time producing the hosts resolver.

**Describe the results you expected:**
I expected /etc/resolv.conf to contain `10.89.0.1` in both cases.

**Additional information you deem important (e.g. issue happens only occasionally):**
This is on a fedora 31 server system.
**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.6
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.6
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.10-2.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.10, commit: 6b526d9888abb86b9e7de7dfdeec0da98ad32ee0'
  Distribution:
    distribution: fedora
    version: "31"
  MemFree: 4174106624
  MemTotal: 6228332544
  OCIRuntime:
    name: crun
    package: crun-0.12.1-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.12.1
      commit: df5f2b2369b3d9f36d175e1183b26e5cee55dd0a
      spec: 1.0.0
[...]

PR #5271, opened 2020-02-19
populate resolv.conf with dnsname responses when in usernamespace

baude:
when using usernamespace, dnsname respondes from cni were not making it into the containers /etc/resolv.conf because of a timing issue.  this corrects that behavior.

Fixes: #5256

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -914,6 +914,7 @@ func (c *Container) checkDependenciesRunning() ([]string, error) {
 }
 
 func (c *Container) completeNetworkSetup() error {
+	var outResolvConf []string
 	netDisabled, err := c.NetworkDisabled()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


libpod/container_internal.go
```diff
@@ -927,7 +928,37 @@ func (c *Container) completeNetworkSetup() error {
...
+	for _, cni := range state.NetworkStatus {
+		if cni.DNS.Nameservers != nil {
+			for _, server := range cni.DNS.Nameservers {
+				outResolvConf = append(outResolvConf, fmt.Sprintf("nameserver %s", server))
+			}
+		}
+	}
+	// check if we have a bindmount for resolv.conf
+	resolvBindMount := state.BindMounts["/etc/resolv.conf"]
+	if len(outResolvConf) < 1 || resolvBindMount != "" || len(c.config.NetNsCtr) > 0 {
```
mheon: I think `resolvBindMount` is flipped - we only want to return nil if it is ""

libpod/container_internal.go
```diff
@@ -927,7 +928,37 @@ func (c *Container) completeNetworkSetup() error {
...
+	}
+	for _, line := range strings.Split(string(b), "\n") {
+		// only keep things that dont start with nameserver from the old
+		// resolv.conf file
+		if !strings.HasPrefix(line, "nameserver") {
+			outResolvConf = append([]string{line}, outResolvConf...)
+		}
+	}
+	// write and return
+	return ioutil.WriteFile(resolvBindMount, []byte(strings.Join(outResolvConf, "\n")), 0644)
```
mheon: This overwrites, and doesn't delete, right?

baude: i believe that is correct.

rhatdan:
LGTM
@mheon @giuseppe @TomSweeneyRedHat @jwhonce @QiWang19 PTAL

Issue #5040, opened 2020-01-31
podman images --filter is missing 'since' support


mbaldessari:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
Docker allows you to filter images based on the since keyword:
[root@compute-0 ~]# docker images --filter since=5730ecdbe270
REPOSITORY                                               TAG                 IMAGE ID            CREATED             SIZE
192.168.24.1:8787/rhosp13/openstack-nova-compute         2019-10-31.1        63a28c287f5f        3 months ago        1.67 GB
192.168.24.1:8787/rhosp13/openstack-ceilometer-central   2019-10-31.1        ab188979b384        3 months ago        722 MB

Podman seems to not have implemented this yet:
$ sudo podman images
REPOSITORY                  TAG      IMAGE ID       CREATED       SIZE
docker.io/library/busybox   latest   6d5fcfe5ff17   5 weeks ago   1.44 MB
$ sudo podman images --filter 'since=6d5fcfe5ff17'
Error: unable to get images: invalid filter since 

**Output of `podman version`:**

```
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.13.5
OS/Arch:            linux/amd64
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-1.7.0-2.fc31.x86_64
```

mheon:
Sure, this should be doable

rhatdan:
@mbaldessari Any chance you could open a PR to add this?

mbaldessari:
I might take a stab in the next week or so, I only filed it because I noticed the discrepancy while doing some unrelated work.

rhatdan:
SGTM

rhatdan:
@mbaldessari Any chance you have time to work on this?

PR #5272, opened 2020-02-19
podman images: add --filter=since=XX

edsantiago:
Looks like a bit of a misunderstanding from early on.

Docker implements --filter=since=IMAGE. Podman implements 'after'
instead of 'since'. Add an equivalent case statement to handle
both, keeping 'after' because we have no way of knowing if it
is used in the field.

Update documentation ... and fix what looks like a complete
misinterpretation of what the code actually does: the man page
claimed that these were time fields, but I don't see any
possible incantation in which a time value works or could
work. Updated docs to reflect IMAGE usage. Also changed
nonworking '==' to single '='.

Added tests.

Fixes: #5040

### DIFF
--- a/docs/source/markdown/podman-images.1.md
+++ b/docs/source/markdown/podman-images.1.md
@@ -29,11 +29,11 @@ Filter output based on conditions provided
 
   Filters:
 
-  **after==TIMESTRING**
-    Filter on images created after the given time.Time.
+  **since=IMAGE**
+    Filter on images created after the given IMAGE (name or tag).
 
-  **before==TIMESTRING**
-    Filter on images created before the given time.Time.
+  **before=IMAGE**
+    Filter on images created before the given IMAGE (name or tag).
 
   **dangling=true|false**
     Show dangling images. Dangling images are a file system layer that was used in a previous build of an image and is no longer referenced by any active images. They are denoted with the <none> tag, consume disk space and serve no active purpose.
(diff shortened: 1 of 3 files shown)


mheon:
/approve
LGTM pending tests

edsantiago:
Force-pushed update: disable new test under podman-remote, which seems to be broken wrt `commit --quiet`

edsantiago:
@mbaldessari PTAL

docs/source/markdown/podman-images.1.md
```diff
@@ -29,11 +29,11 @@ Filter output based on conditions provided
 
   Filters:
 
-  **after==TIMESTRING**
-    Filter on images created after the given time.Time.
+  **after=IMAGE**, **since=IMAGE**
```
rhatdan: Should we stop documenting "after"  and just go with since?

edsantiago: I don't have a feeling for who uses this, but I'm OK with it. Shall I update & push?

TomSweeneyRedHat: I'd be fine with dropping "after" from the docs but keeping it as a hidden option in the code.

edsantiago: Done. Fixed, rebased, pushed. I also changed the order of the options in the `case` statement, putting "since" first to make it clearer to maintainers that it's the preferred option.

PR #5273, opened 2020-02-20
Add support for ssh:// and unix:// podman clients

jwhonce:
* Make context keys package safe
 * Add support for PODMAN_HOST and PODMAN_SSHKEY
 * Add slight increasing delay when client connections fail
 * Remove usages of path.Join(), added JoinURL(). '/' is not OS
   dependent.
 * Add support for secure connections to server


pkg/bindings/connection.go
```diff
@@ -85,26 +135,40 @@ func pingNewConnection(ctx context.Context) error {
...
 
-// newConnection takes a scheme and address and creates a connection from it
-func newConnection(scheme, address string) Connection {
-	client := http.Client{
+func sshClient(_url *url.URL, identity string) (*http.Client, error) {
+	auth, err := publicKey(identity)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to parse identity %s: %v\n", identity)
+	}
+	bastion, err := ssh.Dial("tcp",
```
AkihiroSuda: This doesn't support `~/.ssh/config` and yet `InsecureIgnoreHostKey()` is insecure.
I suggest porting over `docker system dial-stdio`.

rhatdan:
@afbjorklund PTAL

pkg/bindings/connection.go
jwhonce: @AkihiroSuda We trying to actively avoid running ssh on the client.  We did that in the previous version of the API and it was very painful for users. I can work on not using `InsecureIgnoreHostKey()`, or at least make it optional.

AkihiroSuda: Without using ssh binary, isn't it hard to support bastions?

jwhonce: sshd is only required on the bastion, no ssh binary on the client.  I updated the code to allow for known_hosts and secure connections.

AkihiroSuda: Sorry, I meant `ProxyCommand` stuff

baude:
lgtm

PR #5277, opened 2020-02-20
Beautify podman bridge CNI config

saschagrunert:
Bear with me, but we vendor that file directly into the distribution and I want to provide a well-formatted JSON. :see_no_evil: 

Applying prettier to the CNI config to fix mixed indents and improve formatting.


rhatdan:
Much prettier :^)
LGTM

PR #5282, opened 2020-02-20
Login test: use --password-stdin

edsantiago:
Great timing: this new test collided against #5268, which added
a warning about using command-line --password. CI is now going
to fail all over.

Fix: rework test to use --password-stdin. Am doing so only
in the places where output string is checked; other instances
can keep using '--password xxx' because it's simpler.

### DIFF
--- a/test/system/150-login.bats
+++ b/test/system/150-login.bats
@@ -108,8 +108,8 @@ function setup() {
 @test "podman login - basic test" {
     run_podman login --tls-verify=false \
                --username ${PODMAN_LOGIN_USER} \
-               --password ${PODMAN_LOGIN_PASS} \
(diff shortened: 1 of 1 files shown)


edsantiago:
Sorry everyone; this is super-urgent because without it CI will be blocked on all PRs. Any open PRs will probable need to be rebased once this merges.

edsantiago:
To be precise: any PRs based against e7d8bda8 or later will need to be rebased. Any open PRs based against earlier master are fine.

mheon:
/approve
LGTM pending tests

TomSweeneyRedHat:
Barn door has been closed, but LGTM
thanks for the quick turn around @edsantiago

PR #5284, opened 2020-02-20
search test on fedora registry: retry 5 times

edsantiago:
...to try to compensate for flaky host.

registry.fedoraproject.org is just not reliable. It's flaking
with 503 errors, causing massive amounts of wasted CI time
and developer effort.

There is exactly one instance of that registry in these tests.
We can't replace it with quay.io, because "search quay.io/"
(trailing slash) fails with some sort of authentication error.
So let's just try a sleep/retry cycle instead.

### DIFF
--- a/test/e2e/search_test.go
+++ b/test/e2e/search_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strconv"
 	"text/template"
+	"time"
 
 	. "github.com/containers/libpod/test/utils"
 	. "github.com/onsi/ginkgo"
@@ -165,8 +166,16 @@ registries = ['{{.Host}}:{{.Port}}']`
 	})
 
 	It("podman search v2 registry with empty query", func() {
-		search := podmanTest.Podman([]string{"search", "registry.fedoraproject.org/"})
-		search.WaitWithDefaultTimeout()
+		var search *PodmanSessionIntegration
+		for i := 0; i < 5; i++ {
+			search = podmanTest.Podman([]string{"search", "registry.fedoraproject.org/"})
+			search.WaitWithDefaultTimeout()
+			if search.ExitCode() == 0 {
+				break
+			}
+			fmt.Println("Search failed; sleeping & retrying...")
+			time.Sleep(2 * time.Second)
+		}
 		Expect(search.ExitCode()).To(Equal(0))
 		Expect(len(search.OutputToStringArray())).To(BeNumerically(">=", 1))
 	})


test/e2e/search_test.go
```diff
@@ -165,8 +166,16 @@ registries = ['{{.Host}}:{{.Port}}']`
 	})
 
 	It("podman search v2 registry with empty query", func() {
-		search := podmanTest.Podman([]string{"search", "registry.fedoraproject.org/"})
-		search.WaitWithDefaultTimeout()
+		var search
```
mheon: Compiler is yelling because it's not sure what type this is... I'm honestly not either. You might need to make this `search := podmanTest.Podman(...)` so the compiler infers the type

edsantiago: >I'm honestly not either. 

Oh, **thank you**. I've been pulling my hair out trying to figure out the type (I can't get ginkgo working on my laptop, so I can't experiment either). 

If I use `:=` won't it fall out of scope outside the `for` loop? /me doesn't grok golang.

rhatdan: var search *PodmanSessionIntegration

rhatdan: should work

mheon: Yeah, `:=` is just sugar for `var $NAME <inferred type> = ...`

edsantiago: @mheon, I know that much^Wlittle about Go; my question was, doesn't `:=` declare within curlybrace scope? If I did `session := ...`, could I use `session` in the two `Expect()`s below?

TomSweeneyRedHat:
LGTM

test/e2e/search_test.go
mheon: Yes - it'll leak into the sub-scope created by the curly braces (though you can declare another `session` within that sub-scope to shadow it in there).

edsantiago: I'm still not using the right words. Let me try this way. Will this work? I wouldn't expect it to.
```go
    for .... {
        myvar := autodeclared
        ....
    }
    if myvar ...     // <<---- referencing myvar outside of the above "for" scope
```

edsantiago: PS thank you @rhatdan, that seems to be getting it much further along.

edsantiago:
Neither test (`fedora-30` nor `ubuntu-18`) experienced an error, i.e. both succeeded on the first attempt to fedoraregistry, so the retry code never triggered.

mheon:
/approve
/lgtm

Not sure what's going on here...

edsantiago:
githubstatus.com is reporting a recently (minutes ago) resolved issue

PR #5292, opened 2020-02-21
disable generation of cni firewall plugin

baude:
it turns out that when the firewall plugin is not provided as part of the configuration, then the firewall cni plugin will dynamically figure out if it should use firewalld or iptables.

also removing this from the default configuration file


mheon:
LGTM

PR #5305, opened 2020-02-23
Add basic deadlock detection for container start/remove

mheon:
We can easily tell if we're going to deadlock by comparing lock IDs before actually taking the lock. Add a few checks for this in common places where deadlocks might occur.

This does not yet cover pod operations, where detection is more difficult (and costly) due to the number of locks being involved being higher than 2.

Also, add some error wrapping on the Podman side, so we can tell people to use `system renumber` when it occurs.

### DIFF
--- a/cmd/podman/rm.go
+++ b/cmd/podman/rm.go
@@ -4,8 +4,10 @@ import (
 	"fmt"
 
 	"github.com/containers/libpod/cmd/podman/cliconfig"
+	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/pkg/adapter"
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM, although it would have been better if tool just automatically renumbered rather then making the user do it.

mheon:
Renumber should have all containers stopped in order to work properly, so that probably won't work

TomSweeneyRedHat:
LGTM assuming happy tests

mheon:
They don't seem to have fired. I'll repush.

mheon:
This is ready

PR #5307, opened 2020-02-24
fix security-opt generate kube

### LINKED ISSUE #4950: generate kube does not include security-opt
/kind bug

**Description**

The `security-opt` is not included when generated kubernetes pod descriptions.

```
podman run --name test -v `pwd`:/host --security-opt "label=disable" docker.io/library/bash:latest ls /host
podman generate kube > test.yaml
```

The `security-opt` has no influence on the generated yaml

the `test.yaml` should contain `securityContext: seLinuxOptions`

```
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.13.5
```

```
podman-1.7.0-2.fc31.x86_64
```

QiWang19:
fix #4950
add selinux options from --security-opt from the container to generate kube result

### DIFF
--- a/libpod/kube.go
+++ b/libpod/kube.go
@@ -468,11 +468,26 @@ func generateKubeSecurityContext(c *Container) (*v1.SecurityContext, error) {
 		return nil, err
 	}
 
+	var selinuxOpts v1.SELinuxOptions
+	opts := strings.SplitN(c.config.Spec.Annotations[InspectAnnotationLabel], ":", 2)
+	if len(opts) == 2 {
+		switch opts[0] {
+		case "type":
+			selinuxOpts.Type = opts[1]
+		case "level":
+			selinuxOpts.Level = opts[1]
+		}
+	}
+	if len(opts) == 1 {
+		if opts[0] == "disable" {
+			selinuxOpts.Type = "spc_t"
+		}
+	}
+
 	sc := v1.SecurityContext{
-		Capabilities: newCaps,
-		Privileged:   &priv,
-		// TODO How do we know if selinux were passed into podman
-		//SELinuxOptions:
+		Capabilities:   newCaps,
+		Privileged:     &priv,
+		SELinuxOptions: &selinuxOpts,
 		// RunAsNonRoot is an optional parameter; our first implementations should be root only; however
 		// I'm leaving this as a bread-crumb for later
 		//RunAsNonRoot:             &nonRoot,
--- a/test/e2e/generate_kube_test.go
+++ b/test/e2e/generate_kube_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/ghodss/yaml"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 )
 
 var _ = Describe("Podman generate kube", func() {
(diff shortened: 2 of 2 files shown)


libpod/kube.go
```diff
@@ -468,11 +468,28 @@ func generateKubeSecurityContext(c *Container) (*v1.SecurityContext, error) {
...
+		case "user":
+			selinuxOpts.User = opts[1]
+		case "role":
+			selinuxOpts.Role = opts[1]
+		case "type":
+			selinuxOpts.Type = opts[1]
+		case "level":
+			selinuxOpts.Level = opts[1]
+		default:
+			return nil, errors.Errorf("unknown field of SELinuxOptions: %s", opts[0])
```
mheon: Should be a `errors.Wrapf(define.ErrInternal, ...`

libpod/kube.go
```diff
@@ -468,11 +468,28 @@ func generateKubeSecurityContext(c *Container) (*v1.SecurityContext, error) {
 		return nil, err
 	}
 
+	var selinuxOpts v1.SELinuxOptions
+	opts := strings.SplitN(c.config.Spec.Annotations[InspectAnnotationLabel], ":", 2)
+	if len(opts) == 2 {
+		switch opts[0] {
```
mheon: I'm 99% sure that this is going to be set to `disabled` in almost every use, so we need to support that

libpod/kube.go
mheon: We are also missing `filesystem` support, I think.

mheon: Oh, sorry, that second comment should go with the one above

libpod/kube.go
QiWang19: what should we do here if it's `disable`?

mheon: That's an excellent question... Can Kube YAML not request SELinux be turned off? @rhatdan

haircommander: I actually don't think it can on the kube yaml level. I am pretty sure the only way to disable selinux in a pod is to give it privilege

haircommander: here's the object definition: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#selinuxoptions-v1-core, and per: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#podsecuritycontext-v1-core if one isn't specified the runtime allocates one (if it's privileged it's given spc_t, which effectively disables)

libpod/kube.go
QiWang19: > We are also missing `filesystem` support, I think.


the fstype is not defined under v1.SELinuxOptions, should it be supported somewhere else?

libpod/kube.go
QiWang19: can it be hardcoded `selinuxOpts.Type = "spec_t"` to type:spc_t?

haircommander: yeah that can be done

libpod/kube.go
mheon: I'm not sure, but we have to support it somehow - otherwise we'll break `generate kube` on containers with it set

rhatdan:
Kube.yaml supports type not disabled.  
```
podman --security-opt label:disable 
```
means
kube.yaml
```
security_opt:
     type spc_t
```

QiWang19:
@mheon @haircommander @rhatdan PTAL

haircommander:
LGTM

rhatdan:
@FlorianLudwig PTAL

Issue #5311, opened 2020-02-24
APIv2 is now super-slow


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout 10 tcp:127.0.0.1:5001
```
First thing to note: even though the `--timeout 10` should make it exit almost instantaneously, it doesn't. It just sits there. (It eventually exits, maybe 3 minutes later; I haven't done exact timing).

Next, try a trivial query:
```
$ time curl http://localhost:5001/v1.40/version
...
curl http://localhost:5001/v1.40/version  0.01s user 0.01s system 0% cpu 1.629 total
                                                                         ^^^^^^^^^^^
```
That's 1.6 seconds for something that (in previous commits) was 0.07 seconds.

Bisecting shows #5158 to be the culprit. That's way too big a commit for me to grok, so I can't identify the root cause; sorry.

PR #5313, opened 2020-02-24
apiv2 tests: add more pod tests, timing check

edsantiago:
Looks like /libpod/pods/create has been fixed to return an
actual pod ID. Extend those tests.

Also, update timeout in the server command: it's now seconds,
not milliseconds.

Also, update FIXME comments in /pods/prune . Still doesn't
work, but clarify what we're seeing.

Also, add a new test that runs ten /info requests and
barfs if it takes more than 5 seconds.

### DIFF
--- a/test/apiv2/01-basic.at
+++ b/test/apiv2/01-basic.at
@@ -47,4 +47,19 @@ t GET info 200           \
   .DefaultRuntime=runc   \
   .MemTotal~[0-9]\\+
 
+# Timing: make sure server stays responsive
+t0=$SECONDS
+for i in $(seq 1 10); do
+    # FIXME: someday: refactor t(), separate out the 'curl' logic so we
+    #        can call it directly. Then we won't get ten annoying 'ok' lines.
+    t GET info 200
+done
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@baude @mheon @TomSweeneyRedHat @QiWang19 @jwhonce PTAL

test/apiv2/01-basic.at
```diff
@@ -47,4 +47,19 @@ t GET info 200           \
   .DefaultRuntime=runc   \
   .MemTotal~[0-9]\\+
 
+# 2020-02-24 #5158 introduces a massive slowdown, 1-2s per request.
```
TomSweeneyRedHat: I'd not leave this comment in the code.  I'd raise an issue though if not already done.

TomSweeneyRedHat:
LGTM

test/apiv2/01-basic.at
edsantiago: #5311 already filed. I'll rephrase it; I just wanted something to explain the reasoning behind this test

PR #5320, opened 2020-02-24
Add test to validate prune pod apiv2 binding.

sujil02:
Includes test to validate all, no or few pods are pruned,
if the pods are in exited state.

### DIFF
--- a/libpod/pod.go
+++ b/libpod/pod.go
@@ -88,6 +88,7 @@ type PodInspect struct {
 type PodInspectState struct {
 	CgroupPath       string `json:"cgroupPath"`
 	InfraContainerID string `json:"infraContainerID"`
+	Status           string `json:"status"`
 }
 
 // PodContainerInfo keeps information on a container in a pod
--- a/libpod/pod_api.go
+++ b/libpod/pod_api.go
@@ -407,7 +407,10 @@ func (p *Pod) Status() (map[string]define.ContainerStatus, error) {
 	if err != nil {
 		return nil, err
 	}
+	return containerStatusFromContainers(allCtrs)
+}
 
+func containerStatusFromContainers(allCtrs []*Container) (map[string]define.ContainerStatus, error) {
 	// We need to lock all the containers
 	for _, ctr := range allCtrs {
 		ctr.lock.Lock()
@@ -443,6 +446,14 @@ func (p *Pod) Inspect() (*PodInspect, error) {
 	if err != nil {
 		return &PodInspect{}, err
 	}
+	ctrStatuses, err := containerStatusFromContainers(containers)
+	if err != nil {
+		return nil, err
+	}
+	status, err := CreatePodStatusResults(ctrStatuses)
+	if err != nil {
+		return nil, err
+	}
 	for _, c := range containers {
 		containerStatus := "unknown"
 		// Ignoring possible errors here because we don't want this to be
@@ -468,6 +479,7 @@ func (p *Pod) Inspect() (*PodInspect, error) {
 		State: &PodInspectState{
 			CgroupPath:       p.state.CgroupPath,
 			InfraContainerID: infraContainerID,
+			Status:           status,
 		},
 		Containers: podContainers,
 	}
--- a/pkg/bindings/test/pods_test.go
+++ b/pkg/bindings/test/pods_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/onsi/gomega/gexec"
 )
 
-var _ = Describe("Podman images", func() {
+var _ = Describe("Podman pods", func() {
 	var (
 		bt       *bindingTest
 		s        *gexec.Session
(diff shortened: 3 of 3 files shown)


pkg/bindings/test/pods_test.go
```diff
@@ -190,7 +190,45 @@ var _ = Describe("Podman images", func() {
...
+		// Test prune all pods in exited state.
+		bt.Podcreate(&newpod)
+		err = pods.Start(connText, newpod)
+		Expect(err).To(BeNil())
+		err = pods.Start(connText, newpod2)
+		Expect(err).To(BeNil())
+		err = pods.Stop(connText, newpod, nil)
+		Expect(err).To(BeNil())
+		err = pods.Stop(connText, newpod2, nil)
+		Expect(err).To(BeNil())
```
TomSweeneyRedHat: it might be good to add a test somewhere in here to make sure you've two pods up and in the exited state before you remove them.

haircommander:
/ok-to-test

Does this label do anything anymore? All the tests have run

pkg/bindings/test/pods_test.go
```diff
@@ -190,7 +190,45 @@ var _ = Describe("Podman images", func() {
...
+		err = pods.Prune(connText)
+		Expect(err).To(BeNil())
+		podSummary, err = pods.List(connText, nil)
+		Expect(err).To(BeNil())
+		Expect(len(podSummary)).To(Equal(0))
```
baude: what about filters?

mheon:
@haircommander ok-to-test allows the ci/prow/images job to run (which, afaik, we're in the process of deprecating anyways) - so pretty soon it will do nothing, but for now it runs one test.

pkg/bindings/test/pods_test.go
sujil02: Currently, the pod inspect doesn't return the pod status I am working on updating the pod inspect bindings and will do it in the following pr.

pkg/bindings/test/pods_test.go
sujil02: The api for pod list with filters returns 500 as "not implemented". I am looking into it to know if it's a bug or needs implementation. Can we have that in next pr with updated bindings?

pkg/bindings/test/pods_test.go
sujil02: fixed

libpod/pod_api.go
```diff
@@ -432,7 +432,7 @@ func (p *Pod) Inspect() (*PodInspect, error) {
 	var (
 		podContainers []PodContainerInfo
 	)
-
+	status, _ := p.GetPodStatus()
```
umohnani8: This should happen after we update the pod, so move it to line 441.
And check for the error that is returned from this call, don't suppress it

sujil02: fixed.

libpod/pod.go
```diff
@@ -88,6 +88,7 @@ type PodInspect struct {
 type PodInspectState struct {
 	CgroupPath       string `json:"cgroupPath"`
 	InfraContainerID string `json:"infraContainerID"`
+	Status           string `json:"status"`
```
mheon: Why is this in the database? I don't see anything saving to it

libpod/pod_status.go
```diff
@@ -6,7 +6,7 @@ import "github.com/containers/libpod/libpod/define"
 // statuses of the containers in the pod.
 // Returns a string representation of the pod status
 func (p *Pod) GetPodStatus() (string, error) {
-	ctrStatuses, err := p.Status()
+	ctrStatuses, err := p.status()
```
mheon: This is not safe - we need to use the locked version to ensure mutual exclusion

mheon: (The locked version, in this case, being `Status()` - the uppercased version)

libpod/pod.go
mheon: Ahhh, nevermind - this is an Inspect struct... Alright, this ought to be fine

libpod/pod_status.go
umohnani8: The reason for the lowercase `status` here is because `GetPodStatus` is used above in `Inspect` https://github.com/containers/libpod/pull/5320/files#diff-c0092994ef899ada726c98fafdf2e830R443, which has a lock and this was causing a double lock to happen. I guess we can have a lowercase `getPodStatus` function, wdyt @mheon?

mheon: Given this, you cannot use it in `Inspect()` directly - you'll have to call `status()` and then `CreatePodStatusResults` from that

mheon: I would recommend just duplicating the status code in Inspect, it's only 2 function calls

umohnani8: alright, sounds good! @sujil02 ^^

sujil02: fixed.

mheon:
LGTM

rhatdan:
/lgtm
Again.

PR #5324, opened 2020-02-25
rootless: raise error if the process is not found

giuseppe:
we need to store the pause process PID file so that it can be re-used later.
    
commit e9dc2120925d9bc32b87ed3c4122aa40f7413db5 introduced this regression.

Closes: https://github.com/containers/libpod/issues/5246


mheon:
LGTM

PR #5326, opened 2020-02-25
Remove 1 sec delay

### LINKED ISSUE #5311: APIv2 is now super-slow

jwhonce:
* Stop closing net.Listener() twice on interrupt
 * Do not report error if closing server twice
 * Refactor reading channel to make golangci-lint happy

Fixes #5311

### DIFF
--- a/cmd/podman/service.go
+++ b/cmd/podman/service.go
@@ -143,7 +143,6 @@ func runREST(r *libpod.Runtime, uri string, timeout time.Duration) error {
 		if err != nil {
 			return errors.Wrapf(err, "unable to create socket %s", uri)
 		}
-		defer l.Close()
 		listener = &l
 	}
 	server, err := api.NewServerWithSettings(r, timeout, listener)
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

PR #5329, opened 2020-02-25
add epoch for specfile

baude:
to get the copr rpms to jive better with the fedora rpms, we need to set an epoch.

### DIFF
--- a/contrib/spec/podman.spec.in
+++ b/contrib/spec/podman.spec.in
@@ -43,6 +43,11 @@
 %global shortcommit_conmon %(c=%{commit_conmon}; echo ${c:0:7})
 
 Name: podman
+%if 0%{?fedora}
+Epoch: 99
+%else
+Epoch: 0
+%endif
 Version: 1.8.1
 Release: #COMMITDATE#.git%{shortcommit0}%{?dist}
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

larsks:
For posterity, the particular problem this is solving is that the `podman` rpms in Fedora have `Epoch: 2`:

```
$ yum info podman
Name         : podman
Epoch        : 2
Version      : 1.8.0
Release      : 2.fc31
```

The packages in the COPR repository had no epoch, so despite being a more recent build of podman they were always considered a downgrade from the stock Fedora packages (so `dnf upgrade` would always replace them).

It's  possible to resolve this by giving the COPR a lower repository priority (e.g., adding `priority=50` to `/etc/yum.repos.d/_copr:copr.fedorainfracloud.org:baude:Upstream_CRIO_Family.repo`).

This commit addresses the issue in the packages by setting `Epoch: 99` when building fedora packages, ensuring that the COPR packages will always be considered an upgrade from the stock packages.

Looks good to me.

TomSweeneyRedHat:
@larsks TYVM for the addon explanation!  Luv it!
LGTM

PR #5330, opened 2020-02-25
curb flakes in integration tests

baude:
instead of searching the fedora registry which is error prone, we instead search a local registry for the empty set search.

when running two containers with the same IP, i suspect the first container has not fully gotten its ip information back from cni when the second container fires.  rework this test such that we use nginx to make sure the container is up and running before continues which should pace the subsequent test.

### DIFF
--- a/test/e2e/run_staticip_test.go
+++ b/test/e2e/run_staticip_test.go
@@ -3,7 +3,10 @@
 package integration
 
 import (
+	"fmt"
+	"net/http"
 	"os"
+	"time"
 
 	. "github.com/containers/libpod/test/utils"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


edsantiago:
LGTM, thank you!

mheon:
There is almost no chance that CNI is waiting to give us back IP information - the network is fully configured by the time `podman start` returns. I don't think this change is going to fix our flake.

edsantiago:
```
[+0397s] Error: Error copying image to the remote destination: Error initializing destination docker://localhost:5004/my-alpine:latest: error loading registries: open test/registries.conf: no such file or directory
```
This is happening _before_ your changes. I don't get it.

PR #5333, opened 2020-02-26
CI: add API v2 tests

edsantiago:
API v2 has been quiet for a few days, and the test script is
actually passing. Let's take advantage of this opportunity
to get them running in CI.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -398,6 +398,7 @@ testing_task:
     unit_test_script: '$SCRIPT_BASE/unit_test.sh |& ${TIMESTAMP}'
     integration_test_script: '$SCRIPT_BASE/integration_test.sh |& ${TIMESTAMP}'
     system_test_script: '$SCRIPT_BASE/system_test.sh |& ${TIMESTAMP}'
+    apiv2_test_script: '$SCRIPT_BASE/apiv2_test.sh |& ${TIMESTAMP}'
     build_release_script: '$SCRIPT_BASE/build_release.sh |& ${TIMESTAMP}'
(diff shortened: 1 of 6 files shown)


cevich:
so far changes LGTM are there any additional logs which would be helpful to collect for these tests (see `logcollector.sh`)?

edsantiago:
Thanks for the reminder. There's one log file, full results of all HTTP requests, but I haven't yet had a need to look at it even in daily on-my-laptop testing. It's long, and I don't think it's worth preserving because the tests themselves provide a good dump on error.

cevich:
okay SGTM.

edsantiago:
@cevich the failing job is f30, and it's [failing](https://api.cirrus-ci.com/v1/task/6687830147596288/logs/apiv2_test.log) because the system is cgroups v1, so my test expects to see `runc`, but it's seeing `crun` instead. Is this intentional, using `crun` on f30?

TomSweeneyRedHat:
LGTM and seems to be passing everything but the problematic remote_client test

edsantiago:
@TomSweeneyRedHat that's actually a weird and real failure; see my question to @cevich above (basically, f30 system using crun, so my test is confused, as am I)

cevich:
> Is this intentional, using crun on f30?

Yes this was intentional, unless that's bad, then...errr...it was a mistake :grin:

(Meaning: I'm open to hearing arguments to change it)

cevich:
Okay, decision is: F30 should be runc only.  Working on a PR for that (IIUC, should be small/simple change).

edsantiago:
Yay - with Chris's #5342, this is now passing.

PR #3091, opened 2019-05-09
Replace root-based rootless tests

cevich:
Since CI automation is now executing all tests as a regular user, there
is no need for root-based testing to run special rootless tests.  Remove
them.

However, the root-based rootless tests did include one test for exercising
the '--rootfs' option which is needed.  Add a new general, and more through
test to replace it - meaning it will be executed as root and non-root.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #5334, opened 2020-02-26
kill test: clean up warnings; document better

edsantiago:
9f69c4eca (part of the f31 pr, #3091) semi-broke the kill test,
there's now an ugly warning:

    setup(): removing stray images quay.io/libpod/fedora-minimal:latest 7bb5a60e8a78

The comments also didn't actually explain the problem
being addressed, and included a misleading reference
to busybox.

Here we switch to using fedora-minimal only with podman-remote,
clean it up (rmi) when finished, and include an explanation in
the comments about why this is needed; making it clear that
this workaround can be removed once we get rid of podman-remote.
We also reformat back to 80 columns.

### DIFF
--- a/test/system/130-kill.bats
+++ b/test/system/130-kill.bats
@@ -6,12 +6,23 @@
 load helpers
 
 @test "podman kill - test signal handling in containers" {
+    # podman-remote and crun interact poorly in f31: crun seems to gobble up
+    # some signals.
+    # Workaround: run 'env --default-signal sh' instead of just 'sh' in
+    # the container. Since env on our regular alpine image doesn't support
+    # that flag, we need to pull fedora-minimal. See:
+    #    https://github.com/containers/libpod/issues/5004
+    # FIXME: remove this kludge once we get rid of podman-remote
+    local _image=$IMAGE
+    local _sh_cmd="sh"
+    if is_remote; then
+        _image=quay.io/libpod/fedora-minimal:latest
+        _sh_cmd="env --default-signal sh"
+    fi
+
     # Start a container that will handle all signals by emitting 'got: N'
     local -a signals=(1 2 3 4 5 6 8 10 12 13 14 15 16 20 21 22 23 24 25 26 64)
-    # The --default-signal option not available in busybox implementation of 'env' in $IMAGE
-    # needed here to ensure handling of SIGINT inside container uses the default handler
-    _IMAGE=quay.io/libpod/fedora-minimal:latest
-    run_podman run -d $_IMAGE env --default-signal sh -c \
+    run_podman run -d $_image $_sh_cmd -c \
         "for i in ${signals[*]}; do trap \"echo got: \$i\" \$i; done;
         echo READY;
         while ! test -e /stop; do sleep 0.05; done;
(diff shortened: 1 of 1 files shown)


mheon:
LGTM
/approve

test/system/130-kill.bats
```diff
@@ -70,6 +81,10 @@ load helpers
     run_podman wait $cid
     run_podman rm $cid
     wait $podman_log_pid
+
+    if [[ $_image != $IMAGE ]]; then
+        run_podman rmi $_image
+    fi
```
TomSweeneyRedHat: I'm not quite sure why you need to conditionalize this, couldn't you just always do the `rmi`?

TomSweeneyRedHat:
One question, LGTM once the tests are happy.

test/system/130-kill.bats
edsantiago: No, that would be un-nice. The entire rationale behind this test suite is to minimize pulling. In order to do so, all tests are guaranteed to have one and exactly one `$IMAGE` loaded at start-time; so it is courteous for each test to (absent any failures) leave state as they found it.

edsantiago:
Test failures: flake, I think, but am logging for posterity:
```
pkg/bindings/test/containers_test.go:238
.... sends POST /libpod/containers/sha/stop
.... sends GET /libpod/containers/name/json
....expects .State.Status to be "exited", but gets "stopped" instead
```
I've seen this before in system tests, basically `podman ps` showing "stopped" instead of "exited", but have never been able to reproduce it. This may be a real problem.

edsantiago:
The other flake is the one we've been discussing in IRC, "timed out waiting for file".

cevich:
> UPDATE (few minutes later): had to restart this once again (two restarts total). Passed after that.

Thanks @edsantiago I appreciate the help with all these "new" flakes.

test/system/130-kill.bats
```diff
@@ -6,12 +6,23 @@
...
+    # some signals.
+    # Workaround: run 'env --default-signal sh' instead of just 'sh' in
+    # the container. Since env on our regular alpine image doesn't support
+    # that flag, we need to pull fedora-minimal. See:
+    #    https://github.com/containers/libpod/issues/5004
+    # FIXME: remove this kludge once we get rid of podman-remote
+    local _image=$IMAGE
+    local _sh_cmd="sh"
+    if is_remote; then
+        _image=quay.io/libpod/fedora-minimal:latest
```
cevich: FWIW and/or maybe a helpful addition to this comment.  This image is built automatically from `./contrib/fedora-minimal/` (since the fedoraproject registry is so often flaky).

edsantiago: Thanks - I think the `quay.io` is enough to clue someone in that this is a special image; but in this test my goal is to minimize all pulls, even from quay. Every pull is a high-risk chance for a flake.

cevich: Yes I agree with the methodology.

Actually, something that occurred to me:  When the ginkgo tests run, we cache image tarballs in `/tmp` with some very stable/well-known filenames.  If those files exist, pulling could be completely avoided...just import them directly.

PR #5338, opened 2020-02-27
Vendor in latest containers/buildah

umohnani8:
Pulls in fix that sets the correct ownership on
the working directory during the build process.


umohnani8:
@mheon @baude @rhatdan PTAL
We need this in the release branches as well for the various RHEL versions, which branches should I port it too?

mheon:
We should not do a complete revendor for Buildah on stable branches; we want to just include the one patch in question. As for where, we should talk at scrum about what requirements are.

mheon:
@TomSweeneyRedHat We're planning on cutting Podman 1.8.1 tomorrow or Monday, so we might want to think about making that soon?

PR #5342, opened 2020-02-27
Force using runc in F30

cevich:
Suspect crun is being pulled in by installing podman RPM at F30 VM build time.  Attempting to fix that, and also force use of runc for Fedora < 31 at runtime.

Ref: https://github.com/containers/libpod/pull/5333#issuecomment-592048270

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ env:
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
     ###
-    _BUILT_IMAGE_SUFFIX: "libpod-5874660151656448"
+    _BUILT_IMAGE_SUFFIX: "libpod-5940307564953600"
     FEDORA_CACHE_IMAGE_NAME: "fedora-31-${_BUILT_IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "fedora-30-${_BUILT_IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-19-${_BUILT_IMAGE_SUFFIX}"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -48,8 +48,9 @@ env:
     #### Default to NOT operating in any special-case testing mode
     ####
     SPECIALMODE: "none"  # don't do anything special
-    TEST_REMOTE_CLIENT: false  # don't test remote client by default
-    ADD_SECOND_PARTITION: false  # will certainly fail inside containers
+    TEST_REMOTE_CLIENT: 'false'  # don't test remote client by default
+    ADD_SECOND_PARTITION: 'false'  # will certainly fail inside containers
+    MOD_LIBPOD_CONF: 'true'  # Update libpod.conf runtime if required by OS environment
 
     ####
     #### Credentials and other secret-sauces, decrypted at runtime when authorized.
@@ -253,6 +254,9 @@ build_each_commit_task:
         cpu: 8
         memory: "8Gb"
 
+    env:
+        MOD_LIBPOD_CONF: 'false'
+
     timeout_in: 30m
 
     setup_environment_script: '$SCRIPT_BASE/setup_environment.sh |& ${TIMESTAMP}'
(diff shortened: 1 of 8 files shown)


cevich:
> Wouldn't it be simpler to remove handle_crun

Ohohoh snap, I forgot about that damn thing, thanks for the reminder!  Yes, we need that also.

Waiting to push it's removal until VM build finishes, assuming it will pass.  It might fail due to dependency resolution on attempted removal of crun...

contrib/cirrus/setup_environment.sh
```diff
@@ -48,7 +51,7 @@ case "${OS_RELEASE_ID}" in
         if [[ "$ADD_SECOND_PARTITION" == "true" ]]; then
             bash "$SCRIPT_BASE/add_second_partition.sh"; fi
 
-        if [[ "$OS_RELEASE_VER" == "31" ]]; then
+        if [[ "$OS_RELEASE_VER" -ge "31" ]]; then
             warn "Switching io schedular to deadline to avoid RHBZ 1767539"
```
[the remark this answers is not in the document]
cevich: k.  got it, thx.

cevich:
> I don't really understand that logic anyway, since AIUI podman configures itself properly at install-time.

FWIW/IIRC: That was added b/c podman RPM brought one configuration and installing from source needed another.  So it was necessary while we had giuseppe's special 'F30 + crun' task going (since removed).

cevich:
OMG, WTH happened, ugg.  I think maybe we still need to customize `libpod.conf` for F31.  Checking a VM image manually...

cevich:
> could I suggest switching depending on cgroups version?

Considering this now, since it applies across distributions and versions.

.cirrus.yml
```diff
@@ -48,8 +48,9 @@ env:
     #### Default to NOT operating in any special-case testing mode
     ####
     SPECIALMODE: "none"  # don't do anything special
-    TEST_REMOTE_CLIENT: false  # don't test remote client by default
-    ADD_SECOND_PARTITION: false  # will certainly fail inside containers
+    TEST_REMOTE_CLIENT: 'false'  # don't test remote client by default
+    ADD_SECOND_PARTITION: 'false'  # will certainly fail inside containers
+    MOD_LIBPOD_CONF: 'true'  # Update libpod.conf runtime if required by OS environment
```
mheon: I don't believe we actually listen to `libpod.conf` for anything related to runtime. The test suite itself should be specifying runtime, manually, for each call to Podman.

cevich: Actually I think @edsantiago system-tests do use it, no?

In any case, and since we want to (soonish) have both runc/crun happening on the same OS/version.  My thinking is it's best that automation lean toward over rather than under specifying this fact.

I'm open to arguments the other way though, as it certainly is an additional complication.

cevich: Clarification: By over-specific I mean setting `/usr/share/containers/libpod.conf` ***AND*** setting `$OCI_RUNTIME` based on if cgroupsv2 is enabled in the kernel.

cevich:
So...back to an earlier fear of mine, https://github.com/containers/libpod/pull/5342/checks?check_run_id=473384963 is showing that removal of 'crun' package (or something else) in F30 is also yanking out the podman RPM.  We need that for this particular task.  Investigating...

cevich:
...confirmed.  The issue comes from podman in `updates-testing`, so we have a packaging problem.  Invoking the packaging $DIETY @lsm5, bowing to his greatness, and getting out my workaround-duct-tape.  Note: I tried this w/o updates-testing on F30, and podman pulls in runc as expected.

lsm5:
@cevich https://bodhi.fedoraproject.org/updates/FEDORA-2020-89326b1795 has `Suggests: crun` for f30.

cevich:
Damnit! So now the bindings task in F31 is getting things wrong, despite forcing use of crun.  Did something else change between building VM images on Tuesday and Today?

https://cirrus-ci.com/task/5152549762236416?command=integration_test#L625

cevich:
Errr wait, that's not the problem, it says `level=warning`.  Hrmmm...

cevich:
Re-running flaked tests..

cevich:
And again...

mheon:
/approve
LGTM

PR #5347, opened 2020-02-27
rework apiv2 wait endpoint|binding

baude:
added the ability to wait on a condition (stopped, running, paused...) for a container.  if a condition is not provided, wait will default to the stopped condition which uses the original wait code paths.  if the condition is stopped, the container exit code will be returned.

also, correct a mux issue we discovered.

### DIFF
--- a/pkg/api/handlers/generic/containers.go
+++ b/pkg/api/handlers/generic/containers.go
@@ -14,7 +14,6 @@ import (
 	"github.com/containers/libpod/pkg/api/handlers"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/containers/libpod/pkg/util"
-	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -71,7 +70,7 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 		utils.InternalServerError(w, err)
 		return
 	}
-	if _, found := mux.Vars(r)["limit"]; found {
+	if _, found := r.URL.Query()["limit"]; found {
 		last := query.Limit
 		if len(containers) > last {
 			containers = containers[len(containers)-last:]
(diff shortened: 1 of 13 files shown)


pkg/api/handlers/generic/containers.go
```diff
@@ -136,7 +135,7 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) {
 	// /{version}/containers/(name)/wait
 	exitCode, err := utils.WaitContainer(w, r)
 	if err != nil {
-		msg = err.Error()
```
jwhonce: Shouldn't err be reported somewhere?

pkg/api/handlers/libpod/containers.go
```diff
@@ -178,7 +178,6 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
 func WaitContainer(w http.ResponseWriter, r *http.Request) {
 	exitCode, err := utils.WaitContainer(w, r)
 	if err != nil {
-		utils.InternalServerError(w, err)
```
jwhonce: Shouldn't err be reported somewhere?

pkg/api/handlers/utils/containers.go
```diff
@@ -91,25 +93,33 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
...
 		return 0, err
 	}
-
-	if len(query.Condition) > 0 {
-		UnSupportedParameter("condition")
+	interval, err := time.ParseDuration("250ms")
+	if err != nil {
+		InternalServerError(w, err)
+		return 0, err
+	}
```
TomSweeneyRedHat: Not a biggy, but could you do this ParseDuration as an else for line 101?

baude: @rhatdan hates elses

pkg/api/handlers/generic/containers.go
baude: the error is sent down the pipe in utils.WaitContainer ...

pkg/api/handlers/libpod/containers.go
baude: the error is sent down the pipe in utils.WaitContainer ...

pkg/api/handlers/utils/containers.go
TomSweeneyRedHat: yeah, but without one here you're doing ParseDuration twice if you get a value fed in.  Your call.

baude: ok, sure!

Issue #5350, opened 2020-02-27
Podman hanging


astronouth7303:
/kind bug

**Description**

Might be #5291, but I'm not using systemd-spawned containers.

I'm consistently having problems with podman hanging

**Steps to reproduce the issue:**
Ok, this is a multi-terminal thing, but I'll try my best.

I'm starting with a fresh podman instance with nothing else on the system.
```
🐚 podman run --rm --publish 5984:80 nginx
<pulls image>
<waits>
```
```
$ podman ps
CONTAINER ID  IMAGE                           COMMAND               CREATED         STATUS             PORTS                 NAMES
f2ee8b412cc8  docker.io/library/nginx:latest  nginx -g daemon o...  12 seconds ago  Up 12 seconds ago  0.0.0.0:5984->80/tcp  boring_chaum
$ podman rm -f --latest
f2ee8b412cc8cb0a63652018c813d21f978cd3819bc6ef0891fc2dd97c52fa3f
```
(The command in 🐚 exits)

```
🐚 podman run --rm --publish 5984:80 nginx
<waits>
```
```
$ podman ps
<waits>
CONTAINER ID  IMAGE  COMMAND  CREATED  STATUS  PORTS  NAMES
```
```
$ podman rm -f --all
<waits>                                                                                                                     
Error: error removing container d498db43fe12dea382155bcb3c35d12c92c5328f7ea1bafccda94724f2d69d10 root filesystem: unlinkat /home/astraluma/.local/share/containers/storage/overlay-containers/d498db43fe12dea382155bcb3c35d12c92c5328f7ea1bafccda94724f2d69d10/userdata/shm: device or resource busy
```
Ctrl-C in 🐚 does work, and the other two commands run (producing the output above).

**Describe the results you expected:**

No hanging and no weird errors.

**Additional information you deem important (e.g. issue happens only occasionally):**

I'm running `sudo rm -r ~/.local/share/containers/` between runs to consistently get the above behavior.

**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.8
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.8
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v1
  Conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
[...]

PR #5354, opened 2020-02-28
build: specify input fd to buildah

giuseppe:
It solves a tight loop with poll as stdin will be initialized to
/dev/null in buildah/imagebuildah/StageExecutor.Run.

### DIFF
--- a/cmd/podman/build.go
+++ b/cmd/podman/build.go
@@ -352,6 +352,7 @@ func buildCmd(c *cliconfig.BuildValues) error {
 		ContextDirectory:        contextDir,
 		DefaultMountsFilePath:   c.GlobalFlags.DefaultMountsFile,
 		Err:                     stderr,
+		In:                      os.Stdin,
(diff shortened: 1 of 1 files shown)


baude:
lgtm

mheon:
LGTM

fweimer:
Thanks for working on this.

Shouldn't buildah stop polling if it sees an end-of-file condition? Or maybe skip polling altogether if the input isn't something that can be polled, such as a regular file?

TomSweeneyRedHat:
Giuseppe, does this solve #5350?

giuseppe:
> Shouldn't buildah stop polling if it sees an end-of-file condition? Or maybe skip polling altogether if the input isn't something that can be polled, such as a regular file?

yes I agree.  We need to fix the root cause in Buildah as well, the fix here is to make sure it goes in before we re-vendor Buildah.

I've opened an issue in Buildah and assigned myself so we don't forget: https://github.com/containers/buildah/issues/2194

PR #5355, opened 2020-02-28
vendor in c/common config pkg for containers.conf

rhatdan:


rhatdan:
Attempting this without the build section, to see if we can get everything in except for `podman build`

cmd/podman/stop.go
```diff
@@ -42,8 +41,9 @@ func init() {
 	flags.BoolVarP(&stopCommand.Ignore, "ignore", "i", false, "Ignore errors when a specified container is missing")
 	flags.StringArrayVarP(&stopCommand.CIDFiles, "cidfile", "", nil, "Read the container ID from the file")
 	flags.BoolVarP(&stopCommand.Latest, "latest", "l", false, "Act on the latest container podman is aware of")
-	flags.UintVar(&stopCommand.Timeout, "time", define.CtrRemoveTimeout, "Seconds to wait for stop before killing the container")
-	flags.UintVarP(&stopCommand.Timeout, "timeout", "t", define.CtrRemoveTimeout, "Seconds to wait for stop before killing the container")
+	flags.UintVarP(&stopCommand.Timeout, "time", "t", defaultContainerConfig.Libpod.StopTimeout, "Seconds to wait for stop before killing the container")
+	flags.UintVar(&stopCommand.Timeout, "timeout", defaultContainerConfig.Libpod.StopTimeout, "Seconds to wait for stop before killing the container")
+	markFlagHidden(flags, "timeout")
```
mheon: We should just use normalization to map timeout to time

mheon:
This still reads libpod.conf if it exists, right? We cannot break existing users by completely substituting a new configuration file, there has to be a transition period where both are valid.

rhatdan:
The problem with reading libpod.conf versus containers.conf is figuring out priority.

PR #5363, opened 2020-03-01
Add validate() for containers

mheon:
Until now, we've been validating every part of container configuration through the With... functions that set the options. This if fine when we are just validating the options to an individual function, but things get complicated once we need to validate conflicts between different options. We don't know the order in which things were passed, so we need the validation on both of the potential options that can conflict, resulting in significant code duplication. To solve this, add a validate() function for containers, and use this to check whether everything is in a good state.

We can probably move more into this function (there are other parts of container creation that also do validation of a sort) but this is a good start to simplifying our options.

### DIFF
--- a/libpod/container_validate.go
+++ b/libpod/container_validate.go
@@ -1,0 +1,99 @@
+package libpod
+
+import (
+	"github.com/containers/libpod/libpod/define"
+	"github.com/containers/libpod/pkg/rootless"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
Nice consolidation
LGTM

PR #5367, opened 2020-03-02
consolidate env handling into pkg/env

vrothberg:
Env-variable related code is scattered across several packages making it
hard to maintain and extend.  Consolidate the code into a new pkg/env
package.

Signed-off-by: Matthew Heon <matthew.heon@pm.me>

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -1,0 +1,126 @@
+// Package for processing environment variables.
+package env
+
+// TODO: we need to add tests for this package.
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// DefaultEnvVariables set $PATH, $TERM and $container.
+var DefaultEnvVariables = map[string]string{
+	"PATH":      "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+	"TERM":      "xterm",
+	"container": "podman",
+}
+
+const whiteSpaces = " \t"
+
+// ParseSlice parses the specified slice and transforms it into an environment
+// map.
+func ParseSlice(s []string) (map[string]string, error) {
+	env := make(map[string]string, len(s))
+	for _, e := range s {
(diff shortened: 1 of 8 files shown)


rhatdan:
LGTM

pkg/env/env.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// DefaultEnvVariables set $PATH, $TERM and $container.
+var DefaultEnvVariables = map[string]string{
+	"PATH":      "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+	"TERM":      "xterm",
+	"container": "podman",
+}
+
+const whiteSpaces = " \t"
```
TomSweeneyRedHat: Should there be a space in front of the tab?  I'm not sure.

rhatdan: Yes, this is either a space or a tab.

TomSweeneyRedHat: OK, I'm not sure I was clear, should this be `"\t"` or `"{tab or space}\t"` ?

vrothberg: It's used in `strings.TrimLeft` and meant to trim _tabs and whitespaces_, so we're good. `TrimLeft` interprets the string as a set of characters (in contrast to a string/sequence) for trimming.

vrothberg:
Ready to go. @TomSweeneyRedHat @baude @rhatdan @giuseppe PTAL

pkg/env/env.go
```diff
@@ -0,0 +1,126 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// DefaultEnvVariables set $PATH, $TERM and $container.
+var DefaultEnvVariables = map[string]string{
+	"PATH":      "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+	"TERM":      "xterm",
+	"container": "podman",
```
rhatdan: How about 
"container": path.Base(os.Executable())
This would make this package able to be used by non podman.

PR #5372, opened 2020-03-02
Update start stop api to use pod status function.

sujil02:
Update the start and stop api to validate pod status based on status functions
rather than iterating over all the containers in the pod. (Addressing the TODO)

### DIFF
--- a/pkg/api/handlers/libpod/pods.go
+++ b/pkg/api/handlers/libpod/pods.go
@@ -172,7 +172,6 @@ func PodStop(w http.ResponseWriter, r *http.Request) {
 			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
 		return
 	}
-	allContainersStopped := true
 	name := utils.GetName(r)
 	pod, err := runtime.LookupPod(name)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/libpod/pods.go
```diff
@@ -172,34 +172,19 @@ func PodStop(w http.ResponseWriter, r *http.Request) {
...
 	pod, err := runtime.LookupPod(name)
 	if err != nil {
 		utils.PodNotFound(w, name, err)
 		return
 	}
 
-	// TODO we need to implement a pod.State/Status in libpod internal so libpod api
-	// users dont have to run through all containers.
-	podContainers, err := pod.AllContainers()
+	status, err := pod.GetPodStatus()
```
baude: remind me, does this also account for the infra container?

sujil02: Yes, it does takes the infra container into account.  pod.GetPodStatus() and pod.AllContainers() both internally call the runtime.state.PodContainers hence no major change in the implementaion

rhatdan:
@sujil02 needs a rebase.

mheon:
/approve
LGTM on my end

PR #5373, opened 2020-03-02
exec: get the exit code from sync pipe instead of file

haircommander:
Before, we were getting the exit code from the file, in which we waited an arbitrary amount of time (5 seconds) for the file, and segfaulted if we didn't find it. instead, we should be a bit more certain conmon has sent the exit code. Luckily, it sends the exit code along the sync pipe fd, so we can read it from there

Adapt the ExecContainer interface to pass along a channel to get the pid and exit code from conmon, to be able to read both from the pipe

Also slightly change the way we report a bogus value when failing to read from the conmon pipe. we now set it to a value that cannot be a valid pid * -1, to make sure we correctly report the place an error comes from

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -297,7 +297,9 @@ func (c *Container) Exec(tty, privileged bool, env map[string]string, cmd []stri
 		// Conmon will pass a non-zero exit code from the runtime as a pid here.
 		// we differentiate a pid with an exit code by sending it as negative, so reverse
 		// that change and return the exit code the runtime failed with.
-		if pid < 0 {
+		// Make sure the value is not ErrorConmonRead, as that is a podman set bogus value
+		// and not sent by conmon (and thus has no special meaning)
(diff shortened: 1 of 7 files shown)


baude:
code lgtm

mheon:
This may cause issues with my detached exec code for APIv2 - how do we
source exit codes there?

haircommander:
> This may cause issues with my detached exec code for APIv2 - how do we source exit codes there?

it would probably take an entirely different code path. We only wait on the exit code when we are attached to it. In updating the exec exit code (for an podman exec list or something, whatever the api will be for listing a container's exec session), we could check for the existence of the exit code via file. 

If need be, we could even save the pipe fd in the db or the location of the file, and open it as needed. This PR should fit right into your exec rework PR, and we can logic through how detach works with this when you get there. But for now this solves an immediate CI flake, so I'm inclined to go forward and think about reworks later.

haircommander:
This is ready

rhatdan:
LGTM

Issue #5366, opened 2020-03-02
APIv2: Container create requires network


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
Using APIv2 to create a new container fails with `invalid network`. I just want to create a new container with using only the image name as I would do on command line.

**Steps to reproduce the issue:**

1. `sudo curl --unix-socket /run/podman/podman.sock http://d/v1.12/libpod/containers/create -H "Content-Type: application/json" --request POST --data "{\"image\":\"docker.io/library/fedora:latest\"}"`

**Describe the results you received:**
`{"cause":"invalid network \"\"","message":"invalid config provided: invalid network \"\"","response":500}`

**Describe the results you expected:**
A container would be created. This command should be equivalent to `sudo podman create docker.io/library/fedora:latest` which succeeds. 

Using local build of the current master.

baude:
i have a fix for this in my local git as I was fixing a number of issues.  I'll prune it out this morning and make a pr.

marusak:
This is still not fixed.
The same reproducer works exactly the same with `podman-1.8.2-1.fc31.x86_64` which was released yesterday. (Also tried current master to build it for myself and the same result).
Can someone please reopen this issue?

rhatdan:
@baude PTAL

PR #5374, opened 2020-03-02
add default network for apiv2 create

baude:
during container creation, if no network is provided, we need to add a default value so the container can be later started.

use apiv2 container creation for RunTopContainer instead of an exec to the system podman. RunTopContainer now also returns the container id and an error.

also, changed the use of the connections and bindings slightly to make it more convenient to write tests.

Fixes: #5366 


baude:
@sujil02 @lsm5 FYI

edsantiago:
This might be a real failure, not a flake. I keep pressing rerun, but it keeps failing, and it's not a failure I recognize.

PR #5381, opened 2020-03-03
env: set "container" to current binary

vrothberg:
Instead of hard-coding "podman" as the "container" environment variable
set it to the current binary.  This way the package isn't tied to podman
and can be used by others (e.g., cri-o).

@rhatdan PTAL

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -12,11 +12,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-// DefaultEnvVariables set $PATH, $TERM and $container.
+// DefaultEnvVariables sets $PATH and $TERM.
 var DefaultEnvVariables = map[string]string{
-	"PATH":      "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
-	"TERM":      "xterm",
-	"container": "podman",
+	"PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+	"TERM": "xterm",
 }
 
 const whiteSpaces = " \t"
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -477,7 +477,9 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 	//
 	// Precedence order (higher index wins):
 	//  1) env-host, 2) image data, 3) env-file, 4) env
-	var env map[string]string
+	env := map[string]string{
+		"container": "podman",
+	}
 
 	// Start with env-host
 	if c.Bool("env-host") {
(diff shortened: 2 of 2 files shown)


pkg/env/env.go
```diff
@@ -7,16 +7,25 @@ import (
...
-	"PATH":      "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
-	"TERM":      "xterm",
-	"container": "podman",
+	"PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+	"TERM": "xterm",
+}
+
+func init() {
+	// Set "container" to the current binary.
+	bin, err := os.Executable()
```
TomSweeneyRedHat: do we need to guard against symlinks here?

vrothberg: Good question. We don't really need to protect us from them as we only use it as a string but if we had a `docker -> podman` symlink, we would add `docker` as the value. I see a beauty in both.

What's your preference?

rhatdan: I am not sure that is true.  It would depend on how you did the symbolic link.

pkg/env/env.go
```diff
@@ -7,16 +7,25 @@ import (
...
+	if err != nil {
+		panic(fmt.Sprintf("cannot determine executable: %v", err))
```
rhatdan: Rather then panic could you make this a errors.Errorf and then set the id to "unknown"

mheon:
If container is unset, libpod will auto-set it to "libpod" - should we not
set it at all in pkg/env and instead just do it somewhere in cmd/podman?

rhatdan:
I would like to move this code out of libpod at some point, and share it with Buildah and CRI-o perhaps,

vrothberg:
> I would like to move this code out of libpod at some point, and share it with Buildah and CRI-o perhaps,

This will happen once we separate podman and libpod. I am a bit scared of moving too much code to c/common at the moment as we only really test in libpod's CI.

pkg/env/env.go
vrothberg: done

rhatdan:
LGTM

vrothberg:
Restarted the flaked tests

mheon:
Resending as GitHub seems to have lost my reply...

Strongly disagree with ctr=unknown - we do know, from the back end used to
create it. If we cannot sanely determine the binary name, we shouldn't set
anything so defaults will be used in Libpod.

Also, concern: does podman-remote now set container=podman-remote? Is it
bad if that is different between local and remote?

vrothberg:
Changed as suggested by Matt.

TomSweeneyRedHat:
LGTM if the tests get happy

vrothberg:
Thanks for the hint. I don't know why the tests are that flaky lately. Restarted :^)

PR #5387, opened 2020-03-03
Rename long cirrus job names

edsantiago:
Basically, add a 'name' tag (remote, local) to each incantation
of TEST_REMOTE_CLIENT: true/false

I have no idea if this will work or what results it'll produce.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -396,8 +396,12 @@ testing_task:
     env:
         ADD_SECOND_PARTITION: 'true'
         matrix:
-            TEST_REMOTE_CLIENT: 'true'
-            TEST_REMOTE_CLIENT: 'false'
+            - name: remote
+              env:
+                  TEST_REMOTE_CLIENT: 'true'
+            - name: local
+              env:
+                  TEST_REMOTE_CLIENT: 'false'
 
     networking_script: '${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/networking.sh'
     setup_environment_script: '$SCRIPT_BASE/setup_environment.sh |& ${TIMESTAMP}'
@@ -448,8 +452,12 @@ special_testing_rootless_task:
         ADD_SECOND_PARTITION: 'true'
         SPECIALMODE: 'rootless'  # See docs
         matrix:
-            TEST_REMOTE_CLIENT: 'true'
-            TEST_REMOTE_CLIENT: 'false'
+            - name: remote
+              env:
+                  TEST_REMOTE_CLIENT: 'true'
+            - name: local
+              env:
+                  TEST_REMOTE_CLIENT: 'false'
 
     timeout_in: 60m
 
@@ -649,8 +657,12 @@ verify_test_built_images_task:
     env:
         ADD_SECOND_PARTITION: 'true'
         matrix:
-            TEST_REMOTE_CLIENT: 'true'
-            TEST_REMOTE_CLIENT: 'false'
+            - name: remote
+              env:
+                  TEST_REMOTE_CLIENT: 'true'
+            - name: local
+              env:
+                  TEST_REMOTE_CLIENT: 'false'
         matrix:
             # Required env. var. by check_image_script
             PACKER_BUILDER_NAME: "fedora-30"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -34,11 +34,16 @@ env:
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
     ###
+    FEDORA_NAME: "fedora-31"
+    PRIOR_FEDORA_NAME: "fedora-30"
+    UBUNTU_NAME: "ubuntu-19"
+    PRIOR_UBUNTU_NAME: "ubuntu-18"
+
     _BUILT_IMAGE_SUFFIX: "libpod-5940307564953600"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-31-${_BUILT_IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CACHE_IMAGE_NAME: "fedora-30-${_BUILT_IMAGE_SUFFIX}"
-    UBUNTU_CACHE_IMAGE_NAME: "ubuntu-19-${_BUILT_IMAGE_SUFFIX}"
-    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "ubuntu-18-${_BUILT_IMAGE_SUFFIX}"
+    FEDORA_CACHE_IMAGE_NAME: "${FEDORA_NAME}-${_BUILT_IMAGE_SUFFIX}"
+    PRIOR_FEDORA_CACHE_IMAGE_NAME: "${PRIOR_FEDORA_NAME}-${_BUILT_IMAGE_SUFFIX}"
+    UBUNTU_CACHE_IMAGE_NAME: "${UBUNTU_NAME}-${_BUILT_IMAGE_SUFFIX}"
+    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "${PRIOR_UBUNTU_NAME}-${_BUILT_IMAGE_SUFFIX}"
 
     ####
     #### Variables for composing new cache-images (used in PR testing) from
@@ -370,7 +375,7 @@ image_prune_task:
 
 # This task does the unit and integration testing for every platform
 testing_task:
-
+    alias: "testing"
     depends_on:
         - "gating"
         - "vendor"
@@ -383,13 +388,18 @@ testing_task:
         $CIRRUS_CHANGE_MESSAGE !=~ '.*CI:IMG.*' &&
         $CIRRUS_CHANGE_MESSAGE !=~ '.*CI:DOCS.*'
 
-    gce_instance:
-        matrix:
-            image_name: "${FEDORA_CACHE_IMAGE_NAME}"
-            image_name: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
-            # Multiple test failures on Ubuntu 19 - Fixes TBD in future PR
-            # TODO: image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
-            image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
+    matrix:
+        - name: "test ${FEDORA_NAME}"
+          gce_instance:
+              image_name: "${FEDORA_CACHE_IMAGE_NAME}"
+        - name: "test ${PRIOR_FEDORA_NAME}"
+          gce_instance:
+              image_name: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
+        # Multiple test failures on Ubuntu 19 - Fixes TBD in future PR
+        # TODO: image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
+        - name: "test ${PRIOR_UBUNTU_NAME}"
+          gce_instance:
+              image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
 
     timeout_in: 120m
 
@@ -475,7 +485,7 @@ special_testing_rootless_task:
 
 
 special_testing_in_podman_task:
-
+    alias: "special_testing_in_podman"
     depends_on:
         - "gating"
         - "varlink_api"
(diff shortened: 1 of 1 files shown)


edsantiago:
Looks promising so far:
```diff
- testing TEST_REMOTE_CLIENT:true image:fedora-31-libpod-594030756...
+ testing image:fedora-31-libpod-5940307564953600 name:remote
```
I had hoped it would chop off the `name:`, but that's only 5 characters, and overall this gives us 11 characters max vs 24. 

Cirrus is throttling me because of too-quickly-repeated jobs; I'll wait a while before retrying.

edsantiago:
@cevich PTAL. I tried everything I could think of; there doesn't seem to be a way to get rid of the `-libpod-123456789` suffix. IMHO this is still good enough as it is.

cevich:
...found it in [containers/storage configuration](https://github.com/containers/storage/blob/master/.cirrus.yml#L54-L96).  Example of what it looks like (using Cirrus WebUI for convenience):  https://cirrus-ci.com/build/6574673093722112

Point is, the variable name possibly doesn't need to be so verbose.  For instance `$FEDORA_CACHE_IMAGE_NAME` -> `$FIMG1` and `$PRIOR_FEDORA_CACHE_IMAGE_NAME` -> `$FIMG2` (or whatever is short and makes sense).

Though I'm seeing `image: fully-spelled-out-name-122343455667` appear as well.  Not sure if that affects...where are you trying to use the unique (human readable) task name from?

edsantiago:
I'm not sure we're talking about the same thing. For the first part of this PR (`TEST_REMOTE_CLIENT`), a shorter variable name would help, but I think this approach is better. For the second part, I don't think variable names come into play at all. I was just trying to get rid of the useless `libpod-XXXXXXX` garbage that gets displayed on github results. Unfortunately I can't find any way to do so.

edsantiago:
Filed https://github.com/cirruslabs/cirrus-ci-docs/issues/589 in hopes of getting a way to display user-provided names

edsantiago:
OMG everybody look at how nice and readable those CI job names are!

TomSweeneyRedHat:
sweet!

rhatdan:
LGTM

edsantiago:
/hold

@cevich are you responsible for podbot? If so can you check if there's some unwelcome interaction between this PR and podbot?

The short version is: this PR took a lot of retries to get to pass. There were two jobs flaking over and over again: fedora-30 remote and local. Eventually, one of those finally passed. When it did, podbot pinged me (twice!) on #podman :
```
13:55:18 <podbot_50205371> edsantiago: Cirrus-CI testing successful for PR 'Rename long cirrus job names': https://github.com/containers/libpod/pull/5387
13:56:11 <podbot_56650334> edsantiago: Cirrus-CI testing successful for PR 'Rename long cirrus job names': https://github.com/containers/libpod/pull/5387
```
Two different podbots, almost a minute apart. But it was a lie: my job had not finished. The second job was still running; it finally passed about ten minutes later (and there was no corresponding podbot message on IRC).

edsantiago:
More info: doesn't look like my PR's fault. I just saw podbot spit out five identical messages for one unrelated PR:
```
20:48:33 <podbot_65422381> sgrunert: Cirrus-CI testing successful for PR 'Fix podman image sign help output': https://github.com/containers/libpod/pull/5389
20:48:33 <podbot_49421838> sgrunert: Cirrus-CI testing successful for PR 'Fix podman image sign help output': https://github.com/containers/libpod/pull/5389
20:48:33 <podbot_59304614> sgrunert: Cirrus-CI testing successful for PR 'Fix podman image sign help output': https://github.com/containers/libpod/pull/5389
20:48:33 <podbot_56459706> sgrunert: Cirrus-CI testing successful for PR 'Fix podman image sign help output': https://github.com/containers/libpod/pull/5389
20:48:38 <podbot_53830764> sgrunert: Cirrus-CI testing successful for PR 'Fix podman image sign help output': https://github.com/containers/libpod/pull/5389
```
The PR in question is green, but I wasn't paying close enough attention to check if it went green before those messages or after.

cevich:
> When it did, podbot pinged me (twice!) on #podman 

Those notices come from the `success_task` script.  It's possible for there to be multiple executions of the success task at the same time.  I believe this is caused by a combination of multiple side-effects involving dependent tasks.  I debugged it fully once, but couldn't find a simple way to guarantee only once notice is sent per sha.

This is another thing we could move into post-cirrus automation, where we can have more guarantees about the actual completion-state of the overall build.

cevich:
> doesn't look like my PR's fault.

Not I highly doubt it's anything you did.

edsantiago:
Thanks. I've seen some weird podbot messages the past 24 hours, and I concur.

/hold cancel

.cirrus.yml
```diff
@@ -367,7 +372,7 @@ image_prune_task:
 
 # This task does the unit and integration testing for every platform
 testing_task:
-
+    alias: "testing"
```
cevich: Oh neat! Didn't know about that feature.  It's a bit hidden in the docs :confused:

cevich: Oh!  Beware:  Using Alias will impact `contrib/cirrus/cirrus_yaml_test.py` in cases where the task name and alias name differ.  Should be easy to fix though.

.cirrus.yml
```diff
@@ -380,21 +385,30 @@ testing_task:
...
-        matrix:
-            image_name: "${FEDORA_CACHE_IMAGE_NAME}"
-            image_name: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
-            # Multiple test failures on Ubuntu 19 - Fixes TBD in future PR
-            # TODO: image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
-            image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
+    matrix:
+        - name: "testing ${FEDORA_NAME}"
+          gce_instance:
+              image_name: "${FEDORA_CACHE_IMAGE_NAME}"
```
cevich: Okay now that is some black-magic voodoo right there!

I don't see that documented anywhere, and yet it makes a huge impact.  Wow!

I worry this might be an unintended feature (b/c it's undocumented?), it might be worth submitting a documentation update to see how the Cirrus people respond (and to update the docs for everybody's benefit).  I'd hate this feature break/disappear in the future.  

Reg: https://github.com/cirruslabs/cirrus-ci-docs/blob/master/docs/guide/writing-tasks.md

(Plus, they REALLY appreciate help with docs as they're all non-native English speakers.)

.cirrus.yml
```diff
@@ -464,7 +482,7 @@ special_testing_rootless_task:
 
 
 special_testing_in_podman_task:
-
+    alias: "special_testing_in_podman"
```
cevich: FWIW, I don't believe the under-scores are important,  can you use actual spaces here?

Suggestion:  Could these test names be shortened even more to "p-in-p testing"?  Even "In Podman" might be good enough for this audience.

.cirrus.yml
edsantiago: I got the idea from an [already-closed issue](https://github.com/cirruslabs/cirrus-ci-docs/issues/278)

.cirrus.yml
```diff
@@ -476,12 +494,13 @@ special_testing_in_podman_task:
...
-        matrix:
-            # FIXME: Integration testing currently broken for F31 hosts
-            # Error: container_linux.go:345: starting container process caused "process_linux.go:281: applying cgroup configuration for process caused \"mountpoint for cgroup not found\"": OCI runtime error
-            # image_name: "${FEDORA_CACHE_IMAGE_NAME}"
-            image_name: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
+    matrix:
+        # FIXME: Integration testing currently broken for F31 hosts
+        # Error: container_linux.go:345: starting container process caused "process_linux.go:281: applying cgroup configuration for process caused \"mountpoint for cgroup not found\"": OCI runtime error
+        # image_name: "${FEDORA_CACHE_IMAGE_NAME}"
+        - name: "special_testing_in_podman ${PRIOR_FEDORA_NAME}"
```
cevich: Is it necessary to have the task-level `alias: "special_testing_in_podman"` also if you're setting `name` here?  Might be confusing to future maintainers as it's a bit ambiguous as to which name will be used/presented and where.  Though I can understand if this isn't fixable.

.cirrus.yml
edsantiago: I would **LOVE** to shorten even more! I just have no idea what these test names even are supposed to mean. "Special" has no meaning to me. I'm happy to change it to "in podman" (did you really want mixed case?)

.cirrus.yml
edsantiago: This is horrible but yes, it's necessary, because otherwise the only thing appearing in name is "fedora-30" with no way to distinguish it from other tasks. I believe the horribleness tradeoff is worth it for improved end-user experience.

.cirrus.yml
```diff
@@ -646,8 +668,12 @@ verify_test_built_images_task:
...
-            TEST_REMOTE_CLIENT: 'false'
+            - name: remote
+              env:
+                  TEST_REMOTE_CLIENT: 'true'
+            - name: local
+              env:
+                  TEST_REMOTE_CLIENT: 'false'
         matrix:
             # Required env. var. by check_image_script
             PACKER_BUILDER_NAME: "fedora-30"
```
cevich: It would reduce duplication if we used your new `$FEDORA_NAME` variable here for the value, and similar for the lines below.

edsantiago: Good catch, thanks

cevich:
@edsantiago this is really really fantastic, thank you again for helping with these things.  The name length was a loud complaint I heard frequently during initial implementation, but (at the time) was impossible to fix.

One minor suggestion in addition to my comments above:  Consider if we could just drop the term 'testing' entirely from the displayed names.  Everybody here knows / assumes most/all activities here are about or for "testing".  Might help get the names even shorter :grinning: 

In any case, this is a vast improvement as-is and I think the results speak for themselves.

LGTM as-is after fixing up `contrib/cirrus/cirrus_yaml_test.py`, it's fine if you prefer to do followup PRs for my other suggestions.

edsantiago:
> Consider if we could just drop the term 'testing' entirely from the displayed names

I shortened ti to `test`. I'd like to keep _some_ common prefix, because Cirrus/Github sorts results. A common prefix helps cluster everything in `testing_task` together, making it easier IMO for the eye to pick up common success/failure elements.

I would like to explore further renaming of tasks, but believe a followup PR would be better.

Thanks for your feedback! That was helpful; Fixes, cleanup, and rebase coming soon.

.cirrus.yml
cevich: Well! I would **LOVE**  to explain WTF I was thinking when I crafted such a obscure name :smile: 

The primary and most important work done by Cirrus-CI on this repo. is the big "Integration + System testing matrix".  In other words, if any of them fail, nearly all other test results can be ignored.

So "Special" in this case simply denotes automation which is operating on something other than the primary concern.  A different (named) mode of operation (they're all documented) if you will.

Note: In a perfect world, we would have these "special" tasks depend on the primary integration/system tasks.  However, that would badly hurt overall runtime, so we do the "special" stuff in parallel.

hope that helps.

cevich: Other than the variable names, nobody depends on the term "special" AFAIK.  Some other ideas:  Auxiliary, aux, other, non-standard, different, alternative, alt.

Though I question if the distinction is needed at all, developers could probably tell what it is just from the value of `$SPECIALMODE`.

.cirrus.yml
cevich: For sure it is, but since it's horrible and required, it's probably worth documenting or commenting somewhere.

Idea (not sure if possible / would work): Maybe a YAML anchor & alias could help avoid value duplication?   i.e. Anchor on `alias: "special_testing_in_podman" &something` and (somehow) alias that into `name: *something "${PRIOR_FEDORA_NAME}"` (not sure that can be done with scalars).

.cirrus.yml
cevich: Wow...that's some deep/clever searching.

cevich:
okay, follow-up PRs are perfectly fine by me, let's get this in quick so everyone can benefit as-is...

PR #5388, opened 2020-03-04
add tests for container kill and exists

lsm5:
depends on https://github.com/containers/libpod/pull/5374 . Also, tests are currently failing due to a crun issue being worked on separately.

### DIFF
--- a/pkg/bindings/test/containers_test.go
+++ b/pkg/bindings/test/containers_test.go
@@ -410,4 +410,106 @@ var _ = Describe("Podman containers ", func() {
 		_, err = containers.Top(bt.conn, cid, []string{"Me,Neither"})
 		Expect(err).To(BeNil())
 	})
+
+	It("podman bogus container does not exist in local storage", func() {
+		// Bogus container existence check should fail
+		containerExists, err := containers.Exists(bt.conn, "foobar")
+		Expect(err).To(BeNil())
+		Expect(containerExists).To(BeFalse())
+	})
+
+	It("podman container exists in local storage by name", func() {
+		// Container existence check by name should work
+		var name = "top"
+		_, err := bt.RunTopContainer(&name, &bindings.PFalse, nil)
+		Expect(err).To(BeNil())
+		containerExists, err := containers.Exists(bt.conn, name)
+		Expect(err).To(BeNil())
+		Expect(containerExists).To(BeTrue())
+	})
+
+	It("podman container exists in local storage by ID", func() {
+		// Container existence check by ID should work
+		var name = "top"
+		cid, err := bt.RunTopContainer(&name, &bindings.PFalse, nil)
+		Expect(err).To(BeNil())
+		containerExists, err := containers.Exists(bt.conn, cid)
+		Expect(err).To(BeNil())
+		Expect(containerExists).To(BeTrue())
+	})
+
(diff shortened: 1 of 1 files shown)


rhatdan:
@lsm5 needs a rebase

pkg/bindings/test/containers_test.go
```diff
@@ -380,4 +380,103 @@ var _ = Describe("Podman containers ", func() {
...
+	It("podman kill a running container by ID with TERM", func() {
+		// Killing a running container by ID with TERM should work
+		var name = "top"
+		cid, err := bt.RunTopContainer(&name, &bindings.PFalse, nil)
+		Expect(err).To(BeNil())
+		err = containers.Kill(bt.conn, cid, "SIGTERM")
+		Expect(err).To(BeNil())
+	})
+
+	// This test panics after a 10 minute timeout.
```
lsm5: @baude PTAL ^

lsm5: @baude re-ping

lsm5:
@baude @mheon all tests passed. PTAL..

rhatdan:
LGTM
@mheon @baude @vrothberg @giuseppe @edsantiago @cevich PTAL

lsm5:
> Is it really necessary to use SIGKILL so often? I was trained to use that only as a last resort. Maybe you could alternate SIGINT and SIGTERM?
> 
> Also, what would you think of adding an `Exists()` test after one or two of the kill tests? E.g. after line 483? See, the thing is, you're not really testing kill: you're testing "kill does not give me an error". Not the same thing.

@edsantiago addressed both comments. PTAL..

edsantiago:
LGTM assuming tests pass. Thank you!

edsantiago:
Test failure in [rootless local](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5381542914555904/html/system_test.log.html):
```
Error: error removing container b52d9a50bceff47592a01cde287cf54a3f072570f045998efdcb0ee507388803 root filesystem: 1 error occurred:
 	* unlinkat /home/some18944dude/.local/share/containers/storage/overlay/428c1faaae3530ee1f58f0e7c55f6e45c0eb76580e75abc9bbc08cf2e28bc8d5/merged: device or resource busy
```
I'm guessing it's a flake but once again it's not one I've seen before. Could someone please investigate?

edsantiago:
/lgtm

...but I would really like someone mount-knowledgeable (@mheon? @giuseppe?) to look at the flake above and help me understand how to address it / prevent it in the future (or, if it's a real podman problem, which I kind of think it might be, how to gather more data).

mheon:
It seems like the mountpoint is still active when we try and remove it. Hard to tell exactly what's going on without having a look at what exactly is mounted when we try and remove - could be overlay failing to unmount, could be our SHM mount?

edsantiago:
All of those are such simple tests: `podman run X`. Unfortunately I don't have a way of knowing which of those containers was the one leaving the mount behind. Would it help if I reworked the test such that it does 'podman rm` immediately after each `run` command?

lsm5:
@edsantiago everything passes now, so maybe we're good to go?

edsantiago:
I gave it my /lgtm blessing but I don't have privs to `/approve`

PR #5392, opened 2020-03-04
WIP: DO NOT MERGE: Abusing CI system - Add a task that always happens last

cevich:


.mergify.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+# Format Ref: https://doc.mergify.io/configuration.html
+pull_request_rules:
+    - name: 'After Cirrus-CI Fails'
```
cevich: Just a reference label, I don't think the name actually shows up anywhere.  Experimentation will be needed to know for sure.

.mergify.yml
```diff
@@ -0,0 +1,11 @@
...
+      conditions:
+          - '- status-failure ~= cirrus-ci/.*'
```
cevich: This says, match any pull request where the status name matches a regular-expression.  In this case, `cirrus-ci/.*` would  match all tasks with that prefix.  In this PR there is only one: https://github.com/containers/libpod/pull/5392/files#diff-0f7067c31caedf7df2739c9b07efa122R89 for testing.

.mergify.yml
```diff
@@ -0,0 +1,11 @@
...
+      actions:
+          review:
```
cevich: Create a github code-review when the condition matches (one or more cirrus-ci tasks failed).

.mergify.yml
```diff
@@ -0,0 +1,11 @@
...
+              type: 'REQUEST_CHANGES'
```
cevich: Make the code-review block merging of the PR

.mergify.yml
```diff
@@ -0,0 +1,11 @@
...
+              message: '[Mergify](mergify.io) ***review*** comment message text'
```
cevich: In the review comment section, drop whatever text we like.

cevich:
@edsantiago with me so far?

cevich:
Gahhhh....nevermind @edsantiago this won't work.  Trying something simpler instead.

cevich:
Won't work reason:  In order to post a link to a generated summary of test failures, the text must be specific to the PR/branch it relates to.  However, Mergify only operates based on the target branch's copy of `.mergify.yml` - where the text to post is defined.   I could not find any way in their documentation, such as using substitution tokens in the text, to workaround this.

PR #5394, opened 2020-03-04
[ci] Prettier Names for testing tasks

fkorotkov:
Fixes https://github.com/cirruslabs/cirrus-ci-docs/issues/589


edsantiago:
THANK YOU! I was missing the `alias` directive. With that, we are getting clean readable names in the CI review.

PR #5395, opened 2020-03-04
Cirrus: Fix fedora-minimal mirroring

cevich:
Builds of this dockerfile fail on quay.io due to not being able to pull
the base image.  Use a fully-qualified FROM name to work around this.


cevich:
Note: I Tested this Dockerfile build  works on quay, by temporarily creating a trigger for it on my branch of my libpod fork.

PR #5396, opened 2020-03-04
exec: fix error code when conmon fails

haircommander:
this is a cosmetic change that makes sure podman returns a sane error code when conmon dies underneath it

fixes the weirdness in this comment: https://github.com/containers/libpod/issues/5339#issuecomment-594836315

but unfortunately not the underlying problem.

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -340,6 +340,12 @@ func (c *Container) Exec(tty, privileged bool, env map[string]string, cmd []stri
 		if lastErr != nil {
 			logrus.Errorf(lastErr.Error())
 		}
+		// ErrorConmonRead is a bogus value set by podman to indicate reading a value from
+		// conmon failed. Since it is specifically not a valid exit code, we should set
+		// a generic error here
+		if exitCodeData.data == define.ErrorConmonRead {
+			exitCodeData.data = define.ExecErrorCodeGeneric
+		}
 		lastErr = errors.Wrapf(define.ErrOCIRuntime, "non zero exit code: %d", exitCodeData.data)
 	}
 


rhatdan:
@baude @mheon @giuseppe @vrothberg @jwhonce @edsantiago PTAL

libpod/container_api.go
```diff
@@ -341,6 +341,10 @@ func (c *Container) Exec(tty, privileged bool, env map[string]string, cmd []stri
 			logrus.Errorf(lastErr.Error())
 		}
 		lastErr = errors.Wrapf(define.ErrOCIRuntime, "non zero exit code: %d", exitCodeData.data)
+		// if we failed in podman, set the exit code to a generic error
```
edsantiago: "if we failed in **podman**"?

edsantiago:
LGTM but I'm not qualified to understand the whole subcontext here; nor do I have a reliable way to reproduce this.

edsantiago:
I spoke too soon: we _do_ have a way to test this: CI:
```
[+0223s] # time="2020-03-04T16:53:39-05:00" level=error msg="container create failed (no logs from conmon): EOF"
[+0223s] # Error: non zero exit code: -2147483649: OCI runtime error
```
I was expecting this PR to change this message?

libpod/container_api.go
haircommander: I updated the comment to be more informative

haircommander:
> I spoke too soon: we _do_ have a way to test this: CI:
> 
> ```
> [+0223s] # time="2020-03-04T16:53:39-05:00" level=error msg="container create failed (no logs from conmon): EOF"
> [+0223s] # Error: non zero exit code: -2147483649: OCI runtime error
> ```
> 
> I was expecting this PR to change this message?

uh yeah I would expect the message to change too, but I clobbered CI with a force push. Time to wait for another failure :sweat:

edsantiago:
Did you mean to check in the `.lock` file?

And, I misspoke, I wouldn't expect the message to change since the assignment happens after the message.  I'm not sure I understand how this helps...

haircommander:
> Did you mean to check in the `.lock` file?
> 
> And, I misspoke, I wouldn't expect the message to change since the assignment happens after the message. I'm not sure I understand how this helps...

jeez it must be the end of the day....

thanks for your hand holding, I dropped the .lock file and updated the location to make it actually work..

edsantiago:
Looks better:
```
[+0228s] # time="2020-03-04T17:47:55-05:00" level=error msg="container create failed (no logs from conmon): EOF"
[+0228s] # Error: non zero exit code: 125: OCI runtime error
```
Failures restarted, even though one of them is [new to me](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5572233498198016/html/integration_test.log.html#t--podman-failed-to-start-with---rm-should-delete-the-container--1). I'll just treat it as a flake and worry about it later if it recurs.
```
Running: (env: [PODMAN_VARLINK_ADDRESS=unix:/run/user/18722/io.podman-4cdaee893ade05e694bf42b3a156aa50a069802cf4bd86a35acf7e6859ce65a5]) podman [options]-remote start -l
         Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
```

edsantiago:
Why isn't this automerging?

PR #5400, opened 2020-03-05
generate systemd: remove leading slashes

vrothberg:
Remove leading slashes from the run-dir paths. It was meant to make it
explicit that we're dealing with an absolute path but user feedback has
shown that most are aware.  It also cleans up the path in the systemctl
status output.


mheon:
Sure, LGTM

PR #5402, opened 2020-03-05
docs: clarify bools in `podman --help`

vrothberg:
* clarify that --syslog and --trace expects arguments
* mention defaults in --help text
* add --trace to docs


rhatdan:
Should we document --trace or hide it as an option?  Do we want users actually using this?

vrothberg:
> Should we document --trace or hide it as an option? Do we want users actually using this?

No preference. The purist inside of me just wants consistency :^)

rhatdan:
@baude @mheon  WDYT.   I think we should remove documentation for it and hide it.

mheon:
Uhm. Hiding seems appropriate.

vrothberg:
`--trace` is now hidden. @mheon @rhatdan PTAL

rhatdan:
LGTM

PR #5409, opened 2020-03-05
WIP:Filter pods through pod list api

sujil02:
Refactored current pod list flow to libpod internal
so filter pod functionalities can be shared between api and cmd.

### DIFF
--- a/libpod/pod.go
+++ b/libpod/pod.go
@@ -2,10 +2,13 @@ package libpod
 
 import (
 	"net"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/libpod/lock"
+	"github.com/containers/libpod/pkg/util"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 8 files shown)


libpod/pod.go
```diff
@@ -304,3 +307,122 @@ func (p *Pod) GetPodStats(previousContainerStats map[string]*ContainerStats) (ma
 	}
 	return newContainerStats, nil
 }
+
+func (r *Runtime) GetPodsWithFilters(filters []string) ([]*Pod, error) {
```
mheon: We don't want to move string filters all the way into Libpod - these are an artifact of the CLI, we don't want them in the Go API. @baude Where did we end up sticking container filters?

libpod/runtime_pod.go
```diff
@@ -106,14 +107,14 @@ func (r *Runtime) Pods(filters ...PodFilter) ([]*Pod, error) {
 	for _, pod := range pods {
 		include := true
 		for _, filter := range filters {
+			fmt.Println(filter(pod))
```
mheon: Leftover debugging?

sujil02: deleted.

PR #5410, opened 2020-03-05
APIv2: compatible api fixes

st1971:
Added addition API handler registration without the need for a API version number, this mimics the structure of the docker API..
Updated GetImages handler to remove null entry at the end of the images array.
Change GetContainers to map status of configured to created.
Implemented size parameter with default of false on ListContainers handler.
Implemented size parameter on GetContainer.

### DIFF
--- a/pkg/api/handlers/generic/containers.go
+++ b/pkg/api/handlers/generic/containers.go
@@ -58,9 +58,6 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 		// override any golang type defaults
 	}
 
-	// Default Size to false
-	query.Size = false
-
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
 		return
@@ -107,9 +104,6 @@ func GetContainer(w http.ResponseWriter, r *http.Request) {
 	}{
 		// override any golang type defaults
 	}
-
-	// Default Size to false
-	query.Size = false
 
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
--- a/pkg/api/handlers/generic/containers.go
+++ b/pkg/api/handlers/generic/containers.go
@@ -101,6 +101,20 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 
 func GetContainer(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	query := struct {
+		Size bool `schema:"size"`
+	}{
+		// override any golang type defaults
+	}
+
+	// Default Size to false
+	query.Size = false
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
 
 	name := utils.GetName(r)
 	ctnr, err := runtime.LookupContainer(name)
@@ -108,7 +122,7 @@ func GetContainer(w http.ResponseWriter, r *http.Request) {
 		utils.ContainerNotFound(w, name, err)
 		return
 	}
-	api, err := handlers.LibpodToContainerJSON(ctnr)
+	api, err := handlers.LibpodToContainerJSON(ctnr, query.Size)
 	if err != nil {
 		utils.InternalServerError(w, err)
 		return
--- a/pkg/api/handlers/generic/containers.go
+++ b/pkg/api/handlers/generic/containers.go
@@ -57,6 +57,10 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 	}{
 		// override any golang type defaults
 	}
+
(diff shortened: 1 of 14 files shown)


pkg/api/handlers/generic/containers.go
```diff
@@ -57,6 +57,10 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 	}{
 		// override any golang type defaults
 	}
+
+	// Default Size to false
+	query.Size = false
```
mheon: Is this necessary? false is the default for booleans

mheon:
/approve

@jwhonce @baude PTAL... I don't know what we want to do about the duplicate endpoints with regard to swagger.

pkg/api/handlers/generic/containers.go
st1971: your right it's not necessary, should i withdraw the pr change and resubmit?

st1971:
@mheon did consider the swagger definition and i don't think it needs to change, the uris defined in swagger don't include the version information in the path, only taking about the docker compatible endpoints here and not the libpod ones

pkg/api/handlers/generic/containers.go
rhatdan: You can just remove the code and amend your PR and force push it.

pkg/api/handlers/generic/images.go
```diff
@@ -305,7 +305,8 @@ func GetImages(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Failed get images"))
 		return
 	}
-	var summaries = make([]*handlers.ImageSummary, len(images)+1)
+	// +1 removed from len(images) as this leaves a null entry at the end of the image array
```
rhatdan: No need for this comment.  It will mean nothing in the future.  Please remove the comment and just ad it to the PR.

pkg/api/handlers/types.go
```diff
@@ -347,24 +347,35 @@ func ImageDataToImageInspect(ctx context.Context, l *libpodImage.Image) (*ImageI
...
-func LibpodToContainer(l *libpod.Container, infoData []define.InfoData) (*Container, error) {
+func LibpodToContainer(l *libpod.Container, infoData []define.InfoData, sz bool) (*Container, error) {
 	imageId, imageName := l.Image()
 
 	var (
 		err        error
 		sizeRootFs int64
 		sizeRW     int64
 		state      define.ContainerStatus
+		stateStr   string
```
rhatdan: Don't define it here.  Just use `stateStr=state.String()` below.

pkg/api/handlers/types.go
```diff
@@ -347,24 +347,35 @@ func ImageDataToImageInspect(ctx context.Context, l *libpodImage.Image) (*ImageI
...
+
+	if sz {
+		if sizeRW, err = l.RWSize(); err != nil {
+			return nil, err
+		}
+		if sizeRootFs, err = l.RootFsSize(); err != nil {
+			return nil, err
+		}
+	} else {
+		sizeRW = 0
```
rhatdan: These are defaults so no reason for this code block.

pkg/api/handlers/types.go
```diff
@@ -428,6 +444,18 @@ func LibpodToContainerJSON(l *libpod.Container) (*docker.ContainerJSON, error) {
...
 	}
 
+	if sz {
+		if sizeRW, err = l.RWSize(); err != nil {
+			return nil, err
+		}
+		if sizeRootFs, err = l.RootFsSize(); err != nil {
+			return nil, err
+		}
+	} else {
```
rhatdan: No need for else block, These are defaults.

rhatdan:
Thanks @st1971 We need @jwhonce and @baude to review the API Changes.

pkg/api/handlers/generic/containers.go
st1971: Removed unneeded code

pkg/api/handlers/generic/images.go
st1971: Removed comment

pkg/api/handlers/types.go
st1971: Done

pkg/api/handlers/types.go
st1971: Done

pkg/api/handlers/types.go
st1971: The entire if sz block was wrong, have removed and changed the call to inspect to pass sz rather than it being hard coded to true, which seemed a better thing to do

pkg/api/handlers/generic/images.go
```diff
@@ -305,7 +305,7 @@ func GetImages(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Failed get images"))
 		return
 	}
-	var summaries = make([]*handlers.ImageSummary, len(images)+1)
+	var summaries = make([]*handlers.ImageSummary, len(images))
```
jwhonce: The +1 ensures we don't return an empty body. When no images are found.

st1971: just spun up an f31 vm with moby-engine, with no images pulled down a curl to /images/json returns and empty array "[]", on podman in the same state if i put the +1 back in the a curl to /images/json returns "[null]" which is causing consuming code to fail.  without the +1 podman produced the same as docker "[]".

without the +1 podman mimics the behavior of docker on this call

jwhonce: @st1971 Okay, something must have changed in the response writer.  Sorry for the noise, thanks for checking.

st1971: no worries

PR #5412, opened 2020-03-06
Allow users to set TMPDIR environment

rhatdan:
Some users have small /var/tmp directories and need to be able to specify a different location
for temporary files, which includes more space.

Fixes: https://github.com/containers/libpod/issues/5411

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -600,3 +600,12 @@ func HomeDir() (string, error) {
 	}
 	return home, nil
 }
+
+func Tmpdir() string {
+	tmpdir := os.Getenv("TMPDIR")
+	if tmpdir == "" {
+		tmpdir = "/var/tmp"
+	}
+
+	return tmpdir
+}
--- a/cmd/podman/load.go
+++ b/cmd/podman/load.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/libpod/cmd/podman/cliconfig"
 	"github.com/containers/libpod/cmd/podman/shared/parse"
 	"github.com/containers/libpod/pkg/adapter"
+	"github.com/containers/libpod/pkg/util"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
@@ -75,7 +76,7 @@ func loadCmd(c *cliconfig.LoadValues) error {
 		if terminal.IsTerminal(int(os.Stdin.Fd())) {
 			return errors.Errorf("cannot read from terminal. Use command-line redirection or the --input flag.")
 		}
-		outFile, err := ioutil.TempFile("/var/tmp", "podman")
+		outFile, err := ioutil.TempFile(util.Tmpdir(), "podman")
 		if err != nil {
 			return errors.Errorf("error creating file %v", err)
 		}
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -1066,6 +1066,8 @@ b
 **/etc/subuid**
 **/etc/subgid**
 
+NOTE: Use the environment variable `TMPDIR` to change the temporary storage location of downloaded container images. Podman defaults to use `/var/tmp`.
+
 ## SEE ALSO
 subgid(5), subuid(5), libpod.conf(5), systemd.unit(5), setsebool(8), slirp4netns(1), fuse-overlayfs(1)
 
--- a/docs/source/markdown/podman-load.1.md
+++ b/docs/source/markdown/podman-load.1.md
@@ -30,6 +30,8 @@ Read from archive file, default is STDIN.
 
 The remote client requires the use of this option.
 
+NOTE: Use the environment variable `TMPDIR` to change the temporary storage location of container images. Podman defaults to use `/var/tmp`.
+
 **--quiet**, **-q**
 
 Suppress the progress output
--- a/docs/source/markdown/podman-pull.1.md
+++ b/docs/source/markdown/podman-pull.1.md
@@ -156,6 +156,8 @@ Storing signatures
 
 	registries.conf is the configuration file which specifies which container registries should be consulted when completing image names which do not include a registry or domain portion.
 
+NOTE: Use the environment variable `TMPDIR` to change the temporary storage location of downloaded container images. Podman defaults to use `/var/tmp`.
+
 ## SEE ALSO
 podman(1), podman-push(1), podman-login(1), containers-registries.conf(5)
 
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -1335,6 +1335,8 @@ b
 
 **/etc/subgid**
 
+NOTE: Use the environment variable `TMPDIR` to change the temporary storage location of downloaded container images. Podman defaults to use `/var/tmp`.
+
 ## SEE ALSO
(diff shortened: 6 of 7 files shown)


karlyeurl:
Should `cmd/podman/load.go` line 78 be also adjusted?

rhatdan:
@karlyeurl Nice catch, I missed that one.

mheon:
LGTM

pkg/util/utils.go
```diff
@@ -600,3 +600,12 @@ func HomeDir() (string, error) {
...
 }
+
+func Tmpdir() string {
+	tmpdir := os.Getenv("TMPDIR")
+	if tmpdir == "" {
+		tmpdir = "/var/tmp"
+	}
+
+	return tmpdir
+}
```
TomSweeneyRedHat: Does this cause the `--tmpdir` global option to be ignored?

rhatdan: No.

rhatdan: There were already being ignored.

Man page for podman documents this.

**--tmpdir**

Path to the tmp directory, for libpod runtime content.

NOTE --tmpdir is not used for the temporary storage of downloaded images.  Use the environment variable `TMPDIR` to change the temporary storage location of downloaded container images. Podman defaults to use `/var/tmp`.

TomSweeneyRedHat: My bad, I'd pulled up the man page and in my rush neglected to read the note.  Ugh.
LGTM

gcs278:
Hey guys,

Just curious - is there any way to set this via configuration? Our /var/tmp is really small, and it seems a little sketchy adding TMPDIR to every scope (shell/scripts) that podman is running in. It appears the libpod.conf tmp_dir is not the same as TMPDIR which isn't the same as --tmpdir (--tmpdir != $TMPDIR != tmp_dir in libpod.conf).

Just seems like something you'd should be able to control via storage.conf or containers.conf. Our podman solution is looking a little messy with configuration files (storage.conf), environment variables (TMPDIR), some runtime arguments (--tmpdir) that we have to maintain.

Thoughts?

Thanks,

Grant

TomSweeneyRedHat:
@rhatdan would this line in /etc/containers/containers.conf do the trick?
```
# Directory for temporary files. Must be tmpfs (wiped after reboot)
tmp_dir = "/var/run/libpod"
```

mheon:
No, that's the libpod tmpdir, which is distinct.

mheon:
However, there is a field for setting environment of the runtime in containers.conf.

```
# Environment variables to be used when running the container engine (e.g., Podman, Buildah).
# For example "http_proxy=internal.proxy.company.com".
# Note these environment variables will not be used within the container.
# Set the env section under [containers] table, if you want to set environment variables for the container.
# env = []
```

Setting that would let you set the TMPDIR environment variable for all calls to Podman/Buildah.

rhatdan:
Correct
[engine]
env=["TMPDIR=/mybigdir"]

Should work.

saper:
Ran into this running `podman-1.9.3-2.module+el8.2.1+6867+366c07d6.x86_64` from fresh RHEL 8.2 - huge images did not fit into `/var/tmp` despite `TMPDIR` was set.

```
Copying blob 01baf108b52c done
  write /var/tmp/storage651390928/5: no space left on device
Error: error pulling image "docker-virtual.production.artifactory.bshg.com/pytest-base:0.15": unable to pull docker-virtual.production.artifactory.bshg.com/pytest-base:0.15: unable to pull image: Error writing blob: error storing blob to file "/var/tmp/storage651390928/5": write /var/tmp/storage651390928/5: no space left on device
```

Should https://github.com/containers/storage also be changed?

saper:
Ran into this running `podman-1.9.3-2.module+el8.2.1+6867+366c07d6.x86_64` from fresh RHEL 8.2 - huge images did not fit into `/var/tmp` despite `TMPDIR` was set.

```
Copying blob 01baf108b52c done
  write /var/tmp/storage651390928/5: no space left on device
Error: error pulling image "docker-virtual.production.xxx/base:0.15": unable to pull docker-virtual.production.xxx/base:0.15: unable to pull image: Eror writing blob: error storing blob to file "/var/tmp/storage651390928/5": write /var/tmp/storage651390928/5: no space left on device
```

Should https://github.com/containers/storage also be changed?

rhatdan:
This might be fixed in newer versions of Podman.

The containers.conf should not be an issue, if you can trigger this by doing

TMPDIR=/BIGDIR podman ...

That means that the code is not properly telling containers image where to store it's blobs.

saper:
I had `TMPDIR` set to a large directory in the environment while invoking `podman pull`, that's the point. I have installed the update from Redhat on Dec 3rd, so it should be pretty up to date as far as 1.9.x is concerned.

rhatdan:
Right, I am not sure if this was fixed in podman 1.9.  Podman 2.* should have fixes for this though.

gcs278:
> Correct
> [engine]
> env=["TMPDIR=/mybigdir"]
> 
> Should work.

Not sure if this is what @saper was saying - couldn't tell if @saper was were setting via `env` containers.conf, but the above doesn't work for me. Doing `TMPDIR=<TMPDIR> podman images` does indeed work for me. I think `[engine]` should be `[containers]`, but I tried that and it didn't work either.

Still got errors about /var/tmp.
podman version 2.0.5

rhatdan:
It should definitely be [engine] not [containers] 
[containers] will add the environment variable to the container, but not effect he engine launching the container.

rhatdan:
This code is in the main function of podman.

```
	for _, env := range cfg.Engine.Env {
		splitEnv := strings.SplitN(env, "=", 2)
		if len(splitEnv) != 2 {
			return fmt.Errorf("invalid environment variable for engine %s, valid configuration is KEY=value pair", env)
		}
		// skip if the env is already defined
		if _, ok := os.LookupEnv(splitEnv[0]); ok {
			logrus.Debugf("environment variable %s is already defined, skip the settings from containers.conf", splitEnv[0])
			continue
		}
		if err := os.Setenv(splitEnv[0], splitEnv[1]); err != nil {
			return err
		}
	}
```
Which should be setting the environment variable.

rhatdan:
Ok I found the bug.

gcs278:
Thanks Dan.

Another comment here is that - aren't we overloading the environment variable TMPDIR? I know once this fix goes through, I can set it selectively through containers.conf, but at the same time, I broke stuff by setting $TMPDIR in my bashrc. TMPDIR is used by the `mktemp` utility. 

Why can't we have a new variable named like PODMAN_TMPDIR to avoid name conflicts since they do fundamentally different things? Or what this by design?

rhatdan:
Well there are other places within GOLANG that will use TMPDIR, which makes having a different environment variable potentially confusing.

pkg/util/utils.go
kousu: Hi! Thanks for making this change. I was surprised to see my /var filling up when I thought I had locked it down to only daemons.

It's unfortunate this is inconsistent with `mktemp` which defaults to `/tmp` when TMPDIR is unset. I don't suppose you would consider making `/tmp` the default?

TomSweeneyRedHat: @kouso unless my memory banks have fogged over, I believe we originally had /tmp as the default several versions ago, but ran into a lot of issues with container builds not working as there wasn't enough space.

kousu: Interesting! Well, I suppose it depends on your setup.

I can just set up a /etc/profile.d/ file to get `mktemp` and `podman` to agree, so it's not a big deal. Thanks for taking the time to fill me in.

PR #5418, opened 2020-03-07
Shorten five more CI task names

edsantiago:
Now that we know how to do it, let's go whole hog...

Inspired by suggestions from Chris Evich on #5387


edsantiago:
Doesn't look like it worked. There's some funky interaction between `name` and foo_task and `matrix`. Not going to worry about it tonight.

PR #5424, opened 2020-03-09
use c/storage instead of docker/docker

vrothberg:
Use c/storage packages instead of docker/docker ones where possible.


rhatdan:
Nice.

Issue #5423, opened 2020-03-08
podman generate systemd creates incorrect WantedBy for user services


wdouglascampbell:
/kind bug

**Description**

I'm using podman to generate a systemd service file.  I am planning to run the container rootless and use systemd to automatically start the container.

I have enabled linger.

**Steps to reproduce the issue:**

1.  Generate systemd.
```
podman generate systemd -n postfix > ~/.config/systemd/user/container-postfix.service
```
2.  Reload User Service Units.

```
systemctl --user daemon-reload
```

3.  Enable Service.

```
systemctl --user enable container-postfix.service
```

4.  Reboot

**Describe the results you received:**

No attempt was made to start the container on reboot.

**Describe the results you expected:**

Container start attempted on boot.

**Additional information you deem important (e.g. issue happens only occasionally):**

When I edit the container-postfix.service file and change WantedBy line from:

WantedBy=multi-user.target

to:

WantedBy=default.target

and reload the configuration, re-enable, etc and reboot the container runs on boot.

Referencing https://wiki.archlinux.org/index.php/Systemd/User#Writing_user_units it seems to indicate that User Service Unit files should have WantedBy=default.target

Should podman provide an option to specify how to generate the service file?

**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.6
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.6
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v1
  Conmon:
    package: conmon-2.0.10-2.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.10, commit: 6b526d9888abb86b9e7de7dfdeec0da98ad32ee0'
  Distribution:
    distribution: fedora
    version: "31"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 3505111040
  MemTotal: 4121415680
  OCIRuntime:
    name: runc
    package: runc-1.0.0-102.dev.gitdc9208a.fc31.x86_64
    path: /usr/bin/runc
    version: |-
[...]

PR #5427, opened 2020-03-09
generate systemd: add `default.target` to INSTALL

vrothberg:
When enabling a systemd service we can specify which target will start
it by specifying it in the `[INSTALL]` section.  In case of root, this
is commonly set to `multi-user.target` which is used to start other
essential system services such as the network manager, D-BUS and more.

However, the `multi-user.target` is not enough on all systems,
especially when running rootless and enabling user services.  Multiple
users have reported issues that there isn't even an attempt to start the
service.

Setting the INSTALL target to `default.target` will fix the rootless
case.  However, `default.target` may vary among systems.  Fedora
Workstation, for instance, sets the `default.target` to the graphical
target (i.e., runlevel 5) while Fedora Server sets it to
`multi-user.target` which is on runlevel 2 and hence way earlier in the
startup sequence.

As INSTALL allows for specifying multiple INSTALL targets, we can set it
to `multi-user.target` to continue supporting existing workloads AND to
`default.target` which MAY redundantly attempt to start it at a later point;
effectively a NOP for the root case and essential for rootless.

Fixes: #5423


vrothberg:
Once merged, I believe we can close:
- https://github.com/containers/libpod/issues/5291
- https://github.com/containers/libpod/issues/5094
- https://github.com/containers/libpod/issues/5423

vrothberg:
@giuseppe @rhatdan @edsantiago PTAL

edsantiago:
Could you update `docs/source/markdown/podman-generate-systemd.1.md` and `docs/tutorials/rootless_tutorial.md` please? (I feel really dumb for having restarted the flakes before reviewing)

vrothberg:
> Could you update `docs/source/markdown/podman-generate-systemd.1.md` and `docs/tutorials/rootless_tutorial.md` please? (I feel really dumb for having restarted the flakes before reviewing)

Done. Note that I removed the section from the rootless tutorial. The example there was incorrect. As users should use `podman generate systemd` and since there's no real difference between root and rootless (we want so support both), I opted to remove the section.

vrothberg:
Not sure why GitHub is not seeing the new commit ...

vrothberg:
> Not sure why GitHub is not seeing the new commit ...

Accidentally created an upstream branch for it....

edsantiago:
LGTM. Good luck with the flakes.

vrothberg:
Shall we force merge?

mheon:
At this point, yeah, not worth waiting for the bot.

PR #5429, opened 2020-03-09
Revert exec changes

mheon:
This reverts #5373 and #5396 

We expected these to resolve Conmon-related CI flakes, but no such luck. Given this, and the fact that they interfere with the upcoming exec rework and don't seem to resolve any other issues, revert them.

Also should have the side-effect of un-breaking the build on 32-bit


mheon:
@baude @rhatdan @vrothberg @haircommander @TomSweeneyRedHat PTAL

haircommander:
/lgtm
/hold

though, we need another solution for the -1 problem in the second commit.  that can wait though.

mheon:
Alright, bot is stuck, pressing the button manual-like

PR #5437, opened 2020-03-10
Remove nonexistent --set arg from runlabel documentation

coandco:
As far as I can tell from looking through the git history of `cmd/podman/runlabel.go`, there never was a `--set` option.  I'm not sure why it was in the documentation for the command.

### DIFF
--- a/docs/source/markdown/podman-container-runlabel.1.md
+++ b/docs/source/markdown/podman-container-runlabel.1.md
@@ -85,10 +85,6 @@ created from this image.
 
 Set rootfs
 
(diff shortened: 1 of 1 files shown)


coandco:
Done.  The commit now has the proper field.

rhatdan:
@edsantiago We have a test that makes sure that all options in podman help are in man pages, I believe,  Perhaps we need a check to go the other way also, to catch things like this.

@coandco Thanks for the PR.

TomSweeneyRedHat:
LGTM

PR #5445, opened 2020-03-10
Filter pods through pod list api

sujil02:
Refactored current filter pods flow through the shared pod functions
so filter pod functionalities can be shared between api and cmd.

### DIFF
--- a/cmd/podman/pod_ps.go
+++ b/cmd/podman/pod_ps.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
@@ -13,7 +12,6 @@ import (
 	"github.com/containers/libpod/cmd/podman/shared"
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/pkg/adapter"
-	"github.com/containers/libpod/pkg/util"
 	"github.com/docker/go-units"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -28,8 +26,6 @@ const (
 	CREATED      = "Created"
 	NUM_CTR_INFO = 10
 )
-
-type PodFilter func(*adapter.Pod) bool
 
 var (
 	bc_opts shared.PsOptions
@@ -174,29 +170,23 @@ func podPsCmd(c *cliconfig.PodPsValues) error {
 
 	opts.Format = genPodPsFormat(c)
 
-	var filterFuncs []PodFilter
-	if c.Filter != "" {
-		filters := strings.Split(c.Filter, ",")
-		for _, f := range filters {
-			filterSplit := strings.Split(f, "=")
-			if len(filterSplit) < 2 {
-				return errors.Errorf("filter input must be in the form of filter=value: %s is invalid", f)
-			}
-			generatedFunc, err := generatePodFilterFuncs(filterSplit[0], filterSplit[1])
-			if err != nil {
-				return errors.Wrapf(err, "invalid filter")
-			}
-			filterFuncs = append(filterFuncs, generatedFunc)
-		}
-	}
-
 	var pods []*adapter.Pod
+
+	// If latest is set true filters are ignored.
 	if c.Latest {
 		pod, err := runtime.GetLatestPod()
 		if err != nil {
 			return err
 		}
 		pods = append(pods, pod)
+		return generatePodPsOutput(pods, opts)
+	}
+
+	if c.Filter != "" {
+		pods, err = runtime.GetPodsWithFilters(c.Filter)
+		if err != nil {
+			return err
+		}
 	} else {
 		pods, err = runtime.GetAllPods()
 		if err != nil {
@@ -204,19 +194,7 @@ func podPsCmd(c *cliconfig.PodPsValues) error {
 		}
 	}
 
-	podsFiltered := make([]*adapter.Pod, 0, len(pods))
-	for _, pod := range pods {
-		include := true
-		for _, filter := range filterFuncs {
-			include = include && filter(pod)
-		}
-
-		if include {
-			podsFiltered = append(podsFiltered, pod)
-		}
-	}
-
-	return generatePodPsOutput(podsFiltered, opts)
+	return generatePodPsOutput(pods, opts)
 }
 
 // podPsCheckFlagsPassed checks if mutually exclusive flags are passed together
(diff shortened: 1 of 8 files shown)


cmd/podman/pod_ps.go
```diff
@@ -174,49 +170,20 @@ func podPsCmd(c *cliconfig.PodPsValues) error {
...
-			generatedFunc, err := generatePodFilterFuncs(filterSplit[0], filterSplit[1])
-			if err != nil {
-				return errors.Wrapf(err, "invalid filter")
-			}
-			filterFuncs = append(filterFuncs, generatedFunc)
-		}
-	}
-
 	var pods []*adapter.Pod
-	if c.Latest {
```
mheon: I think we need to retain the `c.Latest` code - that's not a filter, but a separate flag

cmd/podman/shared/pod.go
```diff
@@ -134,4 +137,129 @@ func CreatePortBindings(ports []string) ([]ocicni.PortMapping, error) {
 	return portBindings, nil
 }
 
+func GetPodsWithFilters(r *libpod.Runtime, c *cliconfig.PodPsValues) ([]*libpod.Pod, error) {
+
+	var pods []*libpod.Pod
+	var err error
+	if c.Latest {
```
mheon: I don't know if this should be part of it - @baude Thoughts?

cmd/podman/shared/pod.go
```diff
@@ -134,4 +137,129 @@ func CreatePortBindings(ports []string) ([]ocicni.PortMapping, error) {
 	return portBindings, nil
 }
 
+func GetPodsWithFilters(r *libpod.Runtime, c *cliconfig.PodPsValues) ([]*libpod.Pod, error) {
```
mheon: Comments on public functions are always appreciated

cmd/podman/shared/pod.go
```diff
@@ -134,4 +137,129 @@ func CreatePortBindings(ports []string) ([]ocicni.PortMapping, error) {
...
+		pods, err = GenerateFilterFunctionAndGetPods(r, strings.Split(c.Filter, ","))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return pods, nil
+}
+
+func GenerateFilterFunctionAndGetPods(r *libpod.Runtime, filters []string) ([]*libpod.Pod, error) {
```
mheon: Function comment would be appreciated

mheon: Also, should this be public? Looks internal-only

cmd/podman/shared/pod.go
```diff
@@ -134,4 +137,129 @@ func CreatePortBindings(ports []string) ([]ocicni.PortMapping, error) {
...
+		filterFuncs = append(filterFuncs, generatedFunc)
+	}
+	pods, err := r.Pods(filterFuncs...)
+	if err != nil {
+		return nil, err
+	}
+
+	return pods, nil
+}
+func GeneratePodFilterFuncs(filter, filterValue string) (
```
mheon: As above - function comment, and should this be public?

pkg/bindings/test/pods_test.go
```diff
@@ -49,7 +49,6 @@ var _ = Describe("Podman pods", func() {
 		Expect(err).To(BeNil())
 		Expect(response.Config.Name).To(Equal(newpod))
 	})
-
```
mheon: Why was this removed?

pkg/bindings/test/pods_test.go
```diff
@@ -77,14 +76,42 @@ var _ = Describe("Podman pods", func() {
...
 
-		// TODO not working  Because: code to list based on filter
-		// "not yet implemented",
-		// Validate list pod with filters
-		//filters := make(map[string][]string)
-		//filters["name"] = []string{newpod}
-		//filteredPods, err := pods.List(bt.conn, filters)
-		//Expect(err).To(BeNil())
-		//Expect(len(filteredPods)).To(BeNumerically("==", 1))
+		// Validate list pod with status filter
```
mheon: Can these be put in a separate test?

mheon:
@baude PTAL

cmd/podman/shared/pod.go
sujil02: Yeah. As mentioned I am working on the documentation just wanted to get the PR out to validate the code flow.

cmd/podman/shared/pod.go
sujil02: Should I put it back in pod_ps? So, Currently, it's not like that, but do we have to apply pods filter on the latest fetched pod as well if so we will need the flag here.

pkg/bindings/test/pods_test.go
sujil02: Done.

pkg/bindings/test/pods_test.go
sujil02: A typo maybe fixed now.

cmd/podman/shared/pod.go
sujil02: fixed. made private.

cmd/podman/shared/pod.go
sujil02: fixed. made private.

cmd/podman/shared/pod.go
sujil02: Added docs.

cmd/podman/shared/pod.go
mheon: Filters probably should not affect --latest - in fact, I'd go so far as to say filters and `--latest` should conflict.

libpod/runtime_pod.go
```diff
@@ -90,18 +90,16 @@ func (r *Runtime) LookupPod(idOrName string) (*Pod, error) {
...
-		return nil, define.ErrRuntimeStopped
-	}
-
-	pods, err := r.state.AllPods()
+	pods, err := r.GetAllPods()
 	if err != nil {
 		return nil, err
 	}
 
+	return r.FilterPodsWithFilterFunc(pods, filters...)
```
mheon: Why is this changed? I don't see any uses of FilterPodsWithFilterFunc. We also dropped the runtime validity check and lock?

sujil02: Fixed now. I added this to support filters on the latest pod but this is no longer needed.  No, we did not drop that check there or was already a method [GetAllPods](https://github.com/containers/libpod/blob/master/libpod/runtime_pod.go#L121) that does the same just using that.

cmd/podman/pod_ps.go
sujil02: Fixed. The flag is retained.

mheon:
/approve
LGTM
@baude @vrothberg @TomSweeneyRedHat @rhatdan PTAL

TomSweeneyRedHat:
LGTM, nice work @sujil02!

PR #5450, opened 2020-03-10
build for amd64|arm|ppc64le

baude:
to protect against architecture specific build regressions, we cross compile for multiple architectures

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,8 +108,9 @@ endif
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 
 CROSS_BUILD_TARGETS := \
-	bin/podman.cross.darwin.amd64 \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -108,8 +108,9 @@ endif
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 
 CROSS_BUILD_TARGETS := \
-	bin/podman.cross.darwin.amd64 \
```
mheon: Are we dropping the Darwin target?

mheon: Ahhh, is it part of the `podman-rmote-darwin` build above?

baude: that can NEVER work ... in order for it to work, it must be the darwin REMOTE client.

mheon:
LGTM

PR #5460, opened 2020-03-11
man page cross-reference fixes: part 2

edsantiago:
The other direction: fix or clean up elements documented in
man pages but which did/do not exist in actual podman:

  * runlabel: add missing "-n" alias for --name

And, remove man page entries for nonexistent options:

  * podman commit: --iidfile
  * podman container runlabel: --rootfs, --storage
  * podman create: --cpu-count

There are two problems I don't know how to deal with. Both
are related to main_local.go:rootCmd.PersistentFlags() :

  1) podman-build.1.md documents --cni-config-dir and
     --runtime options, but these are not actually options
     under podman build; they are global options. The
     documentation in this man page differs from that
     under podman-build.

  2) podman ps implements a binary --namespace option,
     but this option does not (cannot?) appear in --help
     because there's a global --namespace string option
     and Cobra somehow gets confused about this.

Do we really intend for global options to be parsed on
the right-hand side of subcommands? This strikes me as
unintuitive and potentially confusing, although the
fact that it has taken me this long to discover it
suggests that it's not _that_ confusing.

Suggestions welcome. I can file issues for 1/2 above,
or simply teach my script to special-case ignore them.

### DIFF
--- a/docs/source/markdown/podman-commit.1.md
+++ b/docs/source/markdown/podman-commit.1.md
@@ -38,10 +38,6 @@ Can be set multiple times
 Set the format of the image manifest and metadata.  The currently supported formats are _oci_ and _docker_.  If
 not specifically set, the default format used is _oci_.
 
-**--iidfile**=*ImageIDfile*
-
-Write the image ID to the file.
-
 **--include-volumes**
 
 Include in the committed image any volumes added to the container by the `--volume` or `--mount` options to the `podman create` and `podman run` commands.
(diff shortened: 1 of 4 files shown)


docs/source/markdown/podman-commit.1.md
```diff
@@ -38,10 +38,6 @@ Can be set multiple times
 Set the format of the image manifest and metadata.  The currently supported formats are _oci_ and _docker_.  If
 not specifically set, the default format used is _oci_.
 
-**--iidfile**=*ImageIDfile*
-
-Write the image ID to the file.
```
mheon: We should probably support this. @edsantiago Mind opening up an issue so we remember to add this?

mheon:
@edsantiago Intermingling global and local options is definitely confusing in some cases, but convenient in others (means we can drop `--log-level=debug` anywhere and I don't have to explain to folks that it must go between `podman` and `run`, for example). The duplicate `--namespace` option is concerning to me, though. I would have thought that would be caught by Cobra.

docs/source/markdown/podman-commit.1.md
edsantiago: Done: #5461

PR #5453, opened 2020-03-10
man pages: fix inconsistencies

edsantiago:
I wrote a script to cross-reference podman --help against
man pages. It found a bunch of inconsistencies fix them:

 * options missing from man pages
 * options misspelled or misformatted in man pages (usually
   misplaced asterisks or missing dashes, but see --dns-opt)
 * one spurious comma in the actual source file --help

This is a fix in which I iterate over 'podman CMD --help'
and check for presence in man pages. The other way around
(look for flags in man pages, check podman CMD --help)
is probably impossible: there are too many special cases

Signed-off-by: Ed Santiago <santiago@redhat.com>

PR #5465, opened 2020-03-11
New test: man page cross-ref against --help

edsantiago:
New hack/man-page-option-checker, added to 'make validate',
runs 'podman [subcommand] --help' and cross-references
against man pages in docs/source/markdown/podman*.1.md

See #5453 and #5460 for instances of the problems the
script has found.

The careful reader will find an alarming number of special-case
bypasses. These are a tradeoff I am making: to get perfect
coverage with no handwaving, it would be necessary to make
drastic changes to some man pages, and I believe those would
be counterproductive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -383,6 +383,10 @@ docdir:
 .PHONY: docs
 docs: .install.md2man docdir $(MANPAGES) ## Generate documentation
 
+.PHONE: xref_helpmsgs_manpages
+xref_helpmsgs_manpages:
+	./hack/xref-helpmsgs-manpages
+
 install-podman-remote-%-docs: podman-remote docs $(MANPAGES)
 	rm -rf docs/build/remote
 	mkdir -p docs/build/remote
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -137,7 +137,7 @@ gating_task:
         - 'cd $GOSRC && ./hack/podman-commands.sh |& ${TIMESTAMP}'
         # N/B: need 'clean' so some committed files are re-generated.
         - '/usr/local/bin/entrypoint.sh clean podman-remote |& ${TIMESTAMP}'
-        - '/usr/local/bin/entrypoint.sh clean podman BUILDTAGS="exclude_graphdriver_devicemapper selinux seccomp" |& ${TIMESTAMP}'
+        - '/usr/local/bin/entrypoint.sh clean podman xref_helpmsgs_manpages BUILDTAGS="exclude_graphdriver_devicemapper selinux seccomp" |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh local-cross |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh podman-remote-darwin |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh podman-remote-windows |& ${TIMESTAMP}'
--- a/hack/man-page-checker
+++ b/hack/man-page-checker
@@ -1,16 +1,6 @@
 #!/bin/bash
 #
-# man-page-name-checker - validate and cross-reference man page names
-#
-# FIXME as of 2019-03-20 there are still four files with inconsistent names:
-#
-#    podman-logs.1.md         NAME= podman-container-logs
-#    podman-info.1.md         NAME= podman-system-info
-#    podman-rm.1.md           NAME= podman-container-rm
-#    podman-rmi.1.md          NAME= podman-image-rm
-#
-# If those four get renamed (with suitable symlink fixes), this script
-# can be enabled in CI to prevent future inconsistencies.
+# man-page-checker - validate and cross-reference man page names
 #
 
 die() {
--- a/hack/xref-helpmsgs-manpages
+++ b/hack/xref-helpmsgs-manpages
@@ -1,0 +1,307 @@
+#!/usr/bin/perl
+#
+# xref-helpmsgs-manpages - cross-reference --help options against man pages
+#
+package LibPod::CI::XrefHelpmsgsManpages;
+
+use v5.14;
+use utf8;
+
+use strict;
+use warnings;
+
+(our $ME = $0) =~ s|.*/||;
+our $VERSION = '0.1';
+
+# For debugging, show data structures using DumpTree($var)
+#use Data::TreeDumper; $Data::TreeDumper::Displayaddress = 0;
+
+###############################################################################
+# BEGIN user-customizable section
+
+# Path to podman executable
+my $Default_Podman = './bin/podman';
+my $PODMAN = $ENV{PODMAN} || $Default_Podman;
+
+# Path to podman markdown source files (of the form podman-*.1.md)
+my $Markdown_Path = 'docs/source/markdown';
+
+# END   user-customizable section
+###############################################################################
+
+use FindBin;
+
+###############################################################################
+# BEGIN boilerplate args checking, usage messages
+
+sub usage {
+    print  <<"END_USAGE";
+Usage: $ME [OPTIONS]
+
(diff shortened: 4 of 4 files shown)


edsantiago:
Looks like `make validate` runs before `make`. Any suggestions on where to move this check? It must (obviously) run after `./bin/podman` exists.

rhatdan:
Couldn't you just add it to validate and make it require docs?

edsantiago:
`docs` isn't the problem; `bin/podman` is. I need a compiled executable with which to run `podman --help`.

edsantiago:
@cevich could you PTAL at `.cirrus.yml` and see if that's an okay place to add this test? It runs in <4s, so it shouldn't be too onerous an addition to the test matrix.

TomSweeneyRedHat:
LGTM, but would like a @cevich head nod

cevich:
> docs isn't the problem; bin/podman is. I need a compiled executable with which to run podman --help.

Since we're building podman in the gate task anyway, and this is a really quick check, it's a fine place for this.  Also, it's intended for developers to be able to use the gate image locally, so supporting the check here is a good idea (assuming a working podman).  

Possible perl-script maintenance arguments aside, I do worry if it was a happy-accident perl was even installed.  It might be wise to explicitly list it in `contrib/gate/Dockerfile` for extra assurance.

Otherwise this LGTM

edsantiago:
I'm a little confused. Where does `gate` fit into anything? I added the step in `testing_task`, that doesn't have anything to do with `gating_task`, does it? Other than a dependency?

cevich:
Oh sorry it's probably ***me*** that was confused.  I thought I was looking at the `gate_task` in the diff, peeping again...

cevich:
...okay, hmmm, not sure what I was looking at before.  Taking a second look, this may not be the best place to run this check because updating man pages could be considered work on "docs" and get caught by this in `testing_task`:

```yaml
    only_if: >-
        $CIRRUS_CHANGE_MESSAGE !=~ '.*CI:IMG.*' &&
        $CIRRUS_CHANGE_MESSAGE !=~ '.*CI:DOCS.*'
```

In other words, someone edits a man page, and sets `[CI:DOCS]` on their PR, inadvertently blocking this check :disappointed:  Reminder: That workflow is designed for only-docs related updates, where we want to skip all the other binary-related testing.

Hmmm, so maybe could this be added to the `docs_task`?  That will always run on the default VM (F31 presently), and since `setup_environment.sh` doesn't run, a stable podman (from RPM) will be available, though it won't be built from source.

If you need to test against a built podman, then adding your script into a Makefile target (new?) such that it runs during `gate_task` is probably the safest.

Unless, do you disagree about my `[CI:DOCS]` concern?

edsantiago:
>Hmmm, so maybe could this be added to the docs_task?

That would solve the problem for `CI:DOCS` PRs but completely break every other PR: in general, a change that adds a 'podman foo --bar` option is accompanied by a corresponding addition of `--bar` to the man page, and if this test runs a prepackaged podman, without `foo --bar`, the check will fail.

When I thought about this last week, I came to the conclusion that `CI:DOCS` should/would be used for wording changes, rephrasing, but not significant additions/removals of flags from man pages. This may be an incorrect assumption: I haven't actually checked git history. 

>If you need to test against a built podman, then adding your script into a Makefile target (new?) such that it runs during gate_task is probably the safest.

I'm having trouble parsing this. Can you show me a code diff that you think would work?

cevich:
> but completely break every other PR

Good point.

> but not significant additions/removals of flags from man pages

Hmmm, yes, I agree, and it's purpose is less than clear as well.

> I'm having trouble parsing this. Can you show me a code diff that you think would work?

Sure, happy to...

cevich:
Maybe something like this?

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index fa5c3bb73..e7f7c8b98 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -136,8 +136,8 @@ gating_task:
         - '/usr/local/bin/entrypoint.sh podman |& ${TIMESTAMP}'
         - 'cd $GOSRC && ./hack/podman-commands.sh |& ${TIMESTAMP}'
         # N/B: need 'clean' so some committed files are re-generated.
-        - '/usr/local/bin/entrypoint.sh clean podman-remote |& ${TIMESTAMP}'
-        - '/usr/local/bin/entrypoint.sh clean podman BUILDTAGS="exclude_graphdriver_devicemapper selinux seccomp" |& ${TIMESTAMP}'
+        - '/usr/local/bin/entrypoint.sh clean podman-remote check_docs |& ${TIMESTAMP}'
+        - '/usr/local/bin/entrypoint.sh clean podman check_docs BUILDTAGS="exclude_graphdriver_devicemapper selinux seccomp" |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh local-cross |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh podman-remote-darwin |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh podman-remote-windows |& ${TIMESTAMP}'
@@ -417,7 +417,6 @@ testing_task:
     networking_script: '${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/networking.sh'
     setup_environment_script: '$SCRIPT_BASE/setup_environment.sh |& ${TIMESTAMP}'
     unit_test_script: '$SCRIPT_BASE/unit_test.sh |& ${TIMESTAMP}'
-    man_page_help_xref_script: './hack/man-page-option-checker'
     integration_test_script: '$SCRIPT_BASE/integration_test.sh |& ${TIMESTAMP} | ${LOGFORMAT} integration_test'
     system_test_script: '$SCRIPT_BASE/system_test.sh |& ${TIMESTAMP} | ${LOGFORMAT} system_test'
     apiv2_test_script: '$SCRIPT_BASE/apiv2_test.sh |& ${TIMESTAMP} | ${LOGFORMAT} apiv2_test'
diff --git a/Makefile b/Makefile
index 6104c444c..4f21c4956 100644
--- a/Makefile
+++ b/Makefile
@@ -383,6 +383,10 @@ docdir:
 .PHONY: docs
 docs: .install.md2man docdir $(MANPAGES) ## Generate documentation
 
+.PHONY: check_docs
+check_docs: docs
+       ./hack/man-page-option-checker
+
 install-podman-remote-%-docs: podman-remote docs $(MANPAGES)
        rm -rf docs/build/remote
        mkdir -p docs/build/remote
```

edsantiago:
Thanks, that looks promising. I'll give it a try once CI is fixed. (Not with podman-remote, though, no sense bothering with that).

Makefile
```diff
@@ -383,6 +383,10 @@ docdir:
 .PHONY: docs
 docs: .install.md2man docdir $(MANPAGES) ## Generate documentation
 
+.PHONE: xref_helpmsgs_manpages
+xref_helpmsgs_manpages:
```
edsantiago: I'm really unhappy with this. I would like this to depend on `bin/podman`, but since that's a PHONY, adding that dependency causes a rebuild. Is there any clean way to say that this target depends on an existing `bin/podman` but without forcing a recompile?

cevich: I don't think so, and even if it was possible as an anti-pattern it would likely cause worse problems later on.

It seems like having `bin/podman` as a phony is the real problem...and possibly w/o a good solution that doesn't involve mitigating lots of technical-debt.

This (adding check in CI) might be one of those situations where the best Makefile fix involves plugging your fingers into your ears and singing Marry Poppins :grin:

edsantiago:
I fixed a bug in the man-page handling, and rebased. Tests are green.

rhatdan:
I have spent about 15 minutes in my life working on Perl, but it LGTM

PR #5474, opened 2020-03-12
rootlessport: detect rootless-child exit

giuseppe:
otherwise the rootless parent process might wait indefinitely when the
rootless-child process exits early.


rhatdan:
LGTM

PR #5476, opened 2020-03-12
run --rmi test: make it work

edsantiago:
The recently-added 'run --rmi' test was not actually doing
what it thinks it was doing: for one, 'run_podman | grep'
is never going to work; also, the test was leaving behind
stray images.

Rework to do what I believe the intention was; and, combine
into one test (down from two) for readability.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -136,21 +136,26 @@ echo $rand        |   0 | $rand
     run_podman rmi busybox
 }
 
-# 'run --rmi' deletes the image in the end unless it's used by another container.
-@test "podman run --rmi - remove image" {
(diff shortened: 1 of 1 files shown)


mheon:
/approve
LGTM

TomSweeneyRedHat:
LGTM

rhatdan:
LGTM, 
But is there a test to actually see if rmi works?

edsantiago:
> But is there a test to actually see if rmi works?

`podman run --rmi`, or just plain `rmi`? The last two lines of the new test confirm that `run --rmi` removes the image. As for plain `rmi`, good question: it's used all throughout the tests, but I don't see an explicit confirmation. Now on my TODO list.

PR #5478, opened 2020-03-12
userns: support --userns=auto

giuseppe:
automatically pick an empty range and create an user namespace for the                                                  container.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -823,6 +823,7 @@ The following examples are all valid:
 
 Without this argument the command will be run as root in the container.
 
+**--userns**=*auto*[:OPTIONS]
 **--userns**=*host*
 **--userns**=*keep-id*
 **--userns**=container:container
@@ -831,6 +832,11 @@ Without this argument the command will be run as root in the container.
 
 Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value means user namespaces are disabled.
 
+
+- `auto`: automatically create a namespace.  It is possible to specify other options to `auto`.  The supported options are
+  **size=SIZE** to specify an explicit size for the automatic user namespace.  e.g. `--userns=auto:size=8192`.  If `size` is not specified, `auto` will guess a size for the user namespace.
+  **uidmapping=HOST_UID:CONTAINER_UID:SIZE** to force a UID mapping to be present in the user namespace.
+  **gidmapping=HOST_UID:CONTAINER_UID:SIZE** to force a GID mapping to be present in the user namespace.
 - `container`: join the user namespace of the specified container.
 - `host`: run in the user namespace of the caller. This is the default if no user namespace options are set. The processes running in the container will have the same privileges on the host as any other process launched by the calling user.
 - `keep-id`: creates a user namespace where the current rootless user's UID:GID are mapped to the same values in the container. This option is ignored for containers created by the root user.
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -862,10 +862,14 @@ Sets the username or UID used and optionally the groupname or GID for the specif
 
 Without this argument the command will be run as root in the container.
 
-**--userns**=**host**|**keep-id**|**container:**_id_|**ns:**_namespace_
+**--userns**=**auto**|**host**|**keep-id**|**container:**_id_|**ns:**_namespace_
 
 Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value means user namespaces are disabled.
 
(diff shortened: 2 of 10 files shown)


libpod/runtime_ctr.go
```diff
@@ -872,3 +874,89 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
...
+		totalgid += g.Size
+	}
+
+	u := []idtools.IDMap{{ContainerID: 0, HostID: 1, Size: totaluid}}
+	g := []idtools.IDMap{{ContainerID: 0, HostID: 1, Size: totalgid}}
+	return u, g, nil
+}
+
+// GetAutoUserNS creates an automatic user namespace
+func GetAutoUserNS(size int, imageName string, rt *Runtime) (*storage.IDMappingOptions, error) {
```
mheon: Why is size signed? Unsigned seems like it would make more sense.

giuseppe: I've picked `int` only because it is the same type used by `IDMap` in containers/storage:

```
type IDMap struct {
	ContainerID int `json:"container_id"`
	HostID      int `json:"host_id"`
	Size        int `json:"size"`
}
```

giuseppe:
tests added, ready to review

cmd/podman/shared/create.go
```diff
@@ -180,6 +181,40 @@ func CreateContainer(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 	if err != nil {
 		return nil, nil, err
 	}
+	usernsMode := ns.UsernsMode(c.String("userns"))
+	if !usernsMode.IsHost() {
+		config, err := runtime.GetConfig()
+		if err != nil {
+			return nil, nil, err
+		}
+		graphLock, err := storage.GetLockfile(filepath.Join(config.TmpDir, "userns.lock"))
```
rhatdan: Should this be moved into the storage layer?

libpod/runtime_ctr.go
rhatdan: I think this should be a function in container storage.  We might want to deal with this in Buildah and definitely will want to deal with this in CRI-O.

rhatdan:
The Used ranges of UIDs has to be recorded inside of containers/storage.  Is that happening? 
The maximum range has to be allocated from /etc/subuid and /etc/subgid.
My idea would be to have podman/CRI-o/Buildah installs reserve containers in /etc/subuid and /etc/subgid 

cat /etc/subuid
3267:100000:65536
ashleycui:200000:65536
containers:2147483646:2147483648


Then container storage would use from this range.
We have to guarantee that each container gets a different range then every other container created in container storage.
Also I would like to see us have tools in container/storage to figure out the minimum size of uids, by both getting the max UID in the storage layers, as well as reading /etc/passwd?

giuseppe:
> The Used ranges of UIDs has to be recorded inside of containers/storage. Is that happening?

I'll need to move the logic down to containers/storage.  It is not currently done.



> cat /etc/subuid
> 3267:100000:65536
> ashleycui:200000:65536
> containers:2147483646:2147483648

that means we create a new user "containers"?  Shouldn't it be assigned to root?

```
root:2147483646:2147483648
```

rhatdan:
No I think it should be just on containers, or maybe we put an invalid users name, so a user named containers could still be added.

"container auto ranges"

Bottom line we need to make sure useradd and systemd do not overlay with the range we have allocated.

docs/source/markdown/podman-create.1.md
```diff
@@ -821,13 +821,15 @@ The following examples are all valid:
...
 
+**--userns**=*auto*
 **--userns**=*host*
 **--userns**=*keep-id*
 **--userns**=container:container
 **--userns**=*ns:my_namespace*
 
 Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value means user namespaces are disabled.
 
+- `auto`: automatically create a namespace.  It is possible to specify other options to `auto`.  The only option supported currently is `size`, e.g. `--userns=auto:size=8192`.  If `size` is not specified, `auto` will automatically create a namespace with 4096 IDs available.
```
rhatdan: Should probably reference the storage.conf setting for the 4096 IDs.  Also need to eventually mention that this is the greater of this number versus the minimum number required by the image.

giuseppe: I've changed this part as it is not correct anymore.  I've done more changes on the containers/storage side.

giuseppe:
tests are green

giuseppe:
updated to containers/storage 1.18.1

giuseppe:
we need a new containers/common as well

vrothberg:
@giuseppe are we ready to go?

giuseppe:
> @giuseppe are we ready to go?

yes! :rocket:

PR #5479, opened 2020-03-12
Cirrus: Disable non-docs release processing

cevich:
Detecting when it's time to upload a release inside Cirrus-CI is really
difficult for many automation and human reasons.  Disabling it for now
until a more robust solution can be implemented


rhatdan:
/approved
LGTM
@edsantiago PTAL

Issue #3575, opened 2019-07-14
Auto update containers with podman


viggy96:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**
Is there a method of automatically updating containers when a new version is available in the repository? I use watchtower to achieve this currently with docker, and that container requires mounting the docker unix socket to work, which of course, won't work for podman. Is there another method of achieving this, or is this feature on the roadmap?

rhatdan:
Would a cron job/systemd timer work for this?  Not sure what watchtower is, but it seems line it must be just checking periodically and pulling new images.

rhatdan:
After reading https://hub.docker.com/r/v2tec/watchtower/ it looks like this would be fairly easy to implement via a systemd timer.  

The issue would be if you wanted to make this general purpose. Launch a watcher on an existing container. Being able to get the command line executed, and gernate the CLI command based on it.

viggy96:
Yeah, I want this to be general purpose for all my containers. Which is why I like that watchtower container so much, its a very simple solution to auto update all my containerised applications.

abitrolly:
Docker added `--pull=...` option to `run` recently. Will that help https://github.com/docker/cli/pull/1498 ?

rhatdan:
I don't think this will help, although it is something we should add to Podman.  @QiWang19 Could you add --pull for podman run and create.

rhatdan:
Or @abitrolly If you wanted to add the support that would be great also.

mheon:
This is being worked on.

rhatdan:
@QiWang19 What is the state of this issue?

QiWang19:
> @QiWang19 What is the state of this issue?

I haven't started working on this issue.

mheon:
Should be made substantially easier with the patch from @vrothberg to display the original creation command in `podman inspect`

delight:
> Yeah, I want this to be general purpose for all my containers. Which is why I like that watchtower container so much, its a very simple solution to auto update all my containerised applications.

In watchtower the behavior configureable over labels:

https://containrrr.github.io/watchtower/container-selection/

viggy96:
@delight That's not what I want though. I want to have podman update all my containers just like watchtower does by default.

delight:
[...]

PR #5480, opened 2020-03-12
auto update containers in systemd units

vrothberg:
Add support to auto-update containers running in systemd units as
generated with `podman generate systemd --new`.

`podman restart --auto-update` looks up containers with a specified
"io.containers.autoupdate" label (i.e., the auto-update policy).

If the label is present and set to "image", Podman reaches out to the
corresponding registry to check if the image has been updated.  We
consider an image to be updated if the digest in the local storage is
different than the one of the remote image.  If an image must be
updated, Podman pulls it down and restarts the container.  Note that the
restarting sequence relies on systemd.

At container-creation time, Podman looks up the "PODMAN_SYSTEMD_UNIT"
environment variables and stores it verbatim in the container's label.
This variable is now set by all systemd units generated by
`podman-generate-systemd` and is set to `%n` (i.e., the name of systemd
unit starting the container).  This data is then being used in the
auto-update sequence to instruct systemd (via DBUS) to restart the unit
and hence to restart the container.

Note that this implementation of auto-updates relies on systemd and
requires a fully-qualified image reference to be used to create the
container.  This enforcement is necessary to know which image to
actually check and pull.  If we used an image ID, we would not know
which image to check/pull anymore.

Fixes: #3575


vrothberg:
Note that it's still work in progress. I still need to perform some more tests and add at least some e2e tests to catch error cases.

pkg/systemd/generate/systemdgen.go
```diff
@@ -94,6 +98,7 @@ Before={{- range $index, $value := .RequiredServices -}}{{if $index}} {{end}}{{
 {{- end}}
 
 [Service]
+Environment=PODMAN_SYSTEMD_UNIT=%n
```
mheon: Can we template this to make sure the environment variable always matches EnvVariable above?

pkg/adapter/autoupdate.go
```diff
@@ -0,0 +1,194 @@
...
+			needsUpdate, err := r.newerImageAvailable(image, imageName)
+			if err != nil {
+				errs = append(errs, errors.Wrapf(err, "error auto-updating container %q: image check for %q failed", ctr.ID(), imageName))
+				continue
+			}
+			if !needsUpdate {
+				continue
+			}
+			logrus.Infof("Auto-updating container %q using image %q", ctr.ID(), imageName)
+			_, err = r.updateImage(imageName)
```
mheon: Kind of wish we had a `--pull=update` option to `podman run` so we could just make sure this was part of `generate systemd --new` units - we wouldn't need to pull the image, instead we could rely on `podman run` in the unit file.

mheon: Also - shouldn't this be up a level, in the loop over image ID? So we don't try and re-pull multiple times if I have several containers all on the same image all needing updates?

saschagrunert:
> Add support to auto-update containers running in systemd units as
> generated with `podman generate systemd --new`.

Nice work! :+1: I just have some thoughts without having to look at the code.

> `podman restart --auto-update` looks up containers with a specified
> "io.containers.autoupdate" label (i.e., the auto-update policy).

I would expect that this happens automatically, do you think it would make sense to have something like a systemd timer to really auto-update the image without manual intervention?

> If the label is present and set to "image", Podman reaches out to the
> corresponding registry to check if the image has been updated. We
> consider an image to be updated if the digest in the local storage is
> different than the one of the remote image. If an image must be
> updated, Podman pulls it down and restarts the container. Note that the
> restarting sequence relies on systemd.

Would it be possible to catch update issues? For example we stop the old container, start the new one and if the new one dies we recover back to the old version. Maybe with an additional label that the update failed?

> Note that this implementation of auto-updates relies on systemd and
> requires a fully-qualified image reference to be used to create the
> container. This enforcement is necessary to know which image to
> actually check and pull. If we used an image ID, we would not know
> which image to check/pull anymore.

I'm not sure if I get this correct, but we could just resolve the image during the initial pull, right? So if i just specify `alpine`, the image is available locally then we still could resolve it with the remote resource to obtain the full reference… (would be something like docker.io/library/alpine:latest in that case).

cmd/podman/restart.go
```diff
@@ -41,13 +40,14 @@ func init() {
 	flags.BoolVar(&restartCommand.Running, "running", false, "Restart only running containers when --all is used")
 	flags.UintVarP(&restartCommand.Timeout, "timeout", "t", define.CtrRemoveTimeout, "Seconds to wait for stop before killing the container")
 	flags.UintVar(&restartCommand.Timeout, "time", define.CtrRemoveTimeout, "Seconds to wait for stop before killing the container")
+	flags.BoolVarP(&restartCommand.AutoUpdate, "auto-update", "", false, "Restart all containers configured with an auto-update policy (experimental)")
```
saschagrunert: I don't see a reason to explicitly mark features as experimental. There is not policy around that in podman, right?

cmd/podman/shared/create.go
```diff
@@ -178,6 +180,28 @@ func CreateContainer(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 	if err != nil {
 		return nil, nil, err
 	}
+	createConfig.RawImageName = rawImageName
+
+	// (VR): Ideally we perform the checks _before_ pulling the image but that
+	// would require some bigger code refactoring of `ParseCreateOpts` and the
+	// logic here.  But as the creation code will be consolidation in the future
+	// and given auto updates are experiment, we can live with that for now. In
```
saschagrunert:
```suggestion
	// and given auto updates are experimental, we can live with that for now. In
```

pkg/autoupdate/autoupdate.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// Label denotes the container/pod label key to specify auto-update policies.
+const Label = "io.containers.autoupdate"
+
+type Policy string
+
+const (
+	// PolicyDefault is the default policy denoting no auto updates.
+	PolicyDefault Policy = "default"
```
saschagrunert: I would give the default policy a more explicit name, something like `disabled`

libpod/container.go
```diff
@@ -239,6 +239,12 @@ type ContainerConfig struct {
 	// container has been created with.
 	CreateCommand []string `json:"CreateCommand,omitempty"`
 
+	// RawImageName is the raw and unprocessed name of the image when creating
+	// the container (as specified by the user).  May or may not be set.  One
+	// use case to store this data are auto-updates where we need the _exact_
+	// name and not some normalized instance of it.
+	RawImageName string `json:"RawImageName,omitempty"`
```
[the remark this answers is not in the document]
vrothberg: That's normalized and may even be different to the input one as one image (i.e, ID) can have multi names, so we really need to save what the user specified.

libpod/options.go
```diff
@@ -673,6 +673,18 @@ func WithName(name string) CtrCreateOption {
 	}
 }
 
+// WithRawImageName sets the container config's raw image name (as specified by
+// the user).
+func WithRawImageName(name string) CtrCreateOption {
```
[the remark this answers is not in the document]
vrothberg: I like the idea. I need to check if we have the raw image name at hand for all call sites of `WithImage()`.

pkg/adapter/autoupdate.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/containers/libpod/libpod/image"
+	"github.com/containers/libpod/pkg/autoupdate"
+	"github.com/containers/libpod/pkg/systemd"
+	systemdGen "github.com/containers/libpod/pkg/systemd/generate"
+	"github.com/containers/libpod/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func (r *LocalRuntime) AutoUpdate() ([]string, []error) {
```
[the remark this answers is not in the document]
vrothberg: Sure!

pkg/adapter/autoupdate.go
vrothberg: > Kind of wish we had a `--pull=update` option to `podman run` so we could just make sure this was part of `generate systemd --new` units - we wouldn't need to pull the image, instead we could rely on `podman run` in the unit file.

I had the same thought while working on it. It would work for simple auto-updates but it would close the door for roll backs. For roll backs we have to retag the previous image ID to `RawImageName` and restart the systemd unit. If the unit would always fetch the image from the registry, this would not work.

vrothberg: > Also - shouldn't this be up a level, in the loop over image ID? So we don't try and re-pull multiple times if I have several containers all on the same image all needing updates?

Yes. We need to bundle the updates by image.

vrothberg:
> > `podman restart --auto-update` looks up containers with a specified
> > "io.containers.autoupdate" label (i.e., the auto-update policy).
> 
> I would expect that this happens automatically, do you think it would make sense to have something like a systemd timer to really auto-update the image without manual intervention?

Absolutely. That's actually the end goal to have a oneshot timer running. I'll add a .service file for that once the PR is in a better shape.

> > If the label is present and set to "image", Podman reaches out to the
> > corresponding registry to check if the image has been updated. We
> > consider an image to be updated if the digest in the local storage is
> > different than the one of the remote image. If an image must be
> > updated, Podman pulls it down and restarts the container. Note that the
> > restarting sequence relies on systemd.
> 
> Would it be possible to catch update issues? For example we stop the old container, start the new one and if the new one dies we recover back to the old version. Maybe with an additional label that the update failed?

I plan to add roll backs in the future but not with this PR. It would imply retagging the previous image ID to `RawImageName` and restart the service. Keeping the door for roll backs open also impacted the current design (see https://github.com/containers/libpod/pull/5480#discussion_r392090878).

> > Note that this implementation of auto-updates relies on systemd and
> > requires a fully-qualified image reference to be used to create the
> > container. This enforcement is necessary to know which image to
> > actually check and pull. If we used an image ID, we would not know
> > which image to check/pull anymore.
> 
> I'm not sure if I get this correct, but we could just resolve the image during the initial pull, right? So if i just specify `alpine`, the image is available locally then we still could resolve it with the remote resource to obtain the full reference… (would be something like docker.io/library/alpine:latest in that case).

I want to avoid the problems of unqualified-search-registries and all its ambiguities.

> (would be something like docker.io/library/alpine:latest in that case).

That really depends on the registries.conf, which is why I want to enforce fully-qualified references.

cmd/podman/restart.go
vrothberg: > I don't see a reason to explicitly mark features as experimental. 

If a feature is likely to change in the near future, that's something we need to communicate to users. I do not want users to depend on the current behavior until we have gained some more experience and initial feedback.  Once we feel confident (and implemented roll backs), we can consider it stable.

> There is not policy around that in podman, right?

There's no formal policy, no. We did it a couple of times recently. API v2 would be on example, seccomp-policy for images another.

pkg/autoupdate/autoupdate.go
vrothberg: Good idea!

cmd/podman/restart.go
```diff
@@ -57,6 +57,18 @@ func restartCmd(c *cliconfig.RestartValues) error {
 	}
 	defer runtime.DeferredShutdown(false)
 
+	if c.AutoUpdate {
```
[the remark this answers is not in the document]
vrothberg: Maybe, adding a new command is the way to go. It would allow for adding flags in the future (e.g., `--rollback`) and could be beneficial for docs where can give more verbose examples (without exploding the restart man page).

libpod/container.go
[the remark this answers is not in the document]
mheon: Works for me, just wanted to make sure we weren't duplicating.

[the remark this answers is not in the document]
mheon: Can we throw this in the output of `podman inspect` as well? It seems like a useful thing to know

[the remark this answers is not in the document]
vrothberg: Sure. Cool to do this in a separate PR?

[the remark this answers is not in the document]
mheon: Sure

vrothberg:
* Refined the backend
* All logic is now in pkg/autoupdate
* Added `podman auto-update` command and a man page
* Added a `podman-auto-update.{service,timer}` (untested)
* Addressed all review comments

I am satisfied with the code now. Still need to perform some more manual tests and add some e2e tests.

@edsantiago, I would love to have your eyes and hands on this PR.

vrothberg:
Note to self: create a follow-up PR to allows podman-inspect to return the new RawImageName field.

vrothberg:
Ready from my side. I will add systemd tests in the near future but want to spit that part out as I also want to actually run the generated unit files.

vrothberg:
@rhatdan @edsantiago @TomSweeneyRedHat  @baude @mheon PTAL

edsantiago:
I'm out today, but when playing with this yesterday (with a local registry) I was stumped by the lack of `--tls-verify=false` in `auto-update`. Is that worth considering? (Or perhaps you've already added it, I haven't checked this morning)

vrothberg:
> I'm out today, but when playing with this yesterday (with a local registry) I was stumped by the lack of `--tls-verify=false` in `auto-update`. Is that worth considering? (Or perhaps you've already added it, I haven't checked this morning)

I have considered it but I want users to specify everything in the `registries.conf`. In case of `--tls-verify=false`, the registry has to be marked as insecure. Using `--tls-verify` for run, create, pull etc. is fine as we will pull only one image but in case of auto-update we might pull N images from M registries/repositories.

But that's just my two cents. I am curious what others think about it.

rhatdan:
I agree I think this should be handled in registries.conf.

libpod/container_api.go
```diff
@@ -538,7 +538,7 @@ func (c *Container) Unpause() error {
 	if c.state.State != define.ContainerStatePaused {
 		return errors.Wrapf(define.ErrCtrStateInvalid, "%q is not paused, can't unpause", c.ID())
 	}
-	defer c.newContainerEvent(events.Unpause)
+	defer c.NewContainerEvent(events.Unpause)
```
mheon: I'm really hesitant to make these public; why do we need to add events outside of Libpod?

vrothberg: Maybe we should move `pkg/autoupdate` into `libpod`? I want this to show up in the events.

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -0,0 +1,46 @@
+% podman-auto-update(1)
+
+## NAME
+podman\-auto-update - Auto update containers according to their auto-update policy
```
TomSweeneyRedHat: do you need a '\' before `-update`?

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -0,0 +1,46 @@
...
+
+If the label is present and set to "image", Podman reaches out to the corresponding registry to check if the image has been updated.
+An image is considered updated if the digest in the local storage is different than the one of the remote image.
+If an image must be updated, Podman pulls it down and restarts the systemd unit executing the container.
+
+At container-creation time, Podman looks up the "PODMAN_SYSTEMD_UNIT" environment variables and stores it verbatim in the container's label.
+This variable is now set by all systemd units generated by `podman-generate-systemd` and is set to `%n` (i.e., the name of systemd unit starting the container).
+This data is then being used in the auto-update sequence to instruct systemd (via DBUS) to restart the unit and hence to restart the container.
+
+Note that this implementation of auto updates relies on systemd and requires a fully-qualified image reference (e.g., quay.io/podman/stable:latest) to be used to create the container.
```
TomSweeneyRedHat: auto updates -> auto update    # lose the 's'

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -0,0 +1,46 @@
...
+This enforcement is necessary to know which image to actually check and pull.
+If we used an image ID, we would not know which image to check/pull anymore.
```
TomSweeneyRedHat: If an Image ID was used, auto update will not know which image to check or pull.

pkg/systemd/dbus.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/containers/libpod/pkg/rootless"
+	"github.com/coreos/go-systemd/v22/dbus"
+	godbus "github.com/godbus/dbus/v5"
```
TomSweeneyRedHat: On second reading I realize this is "go-dbus".  But on the first pass through I read "god-bus" and given corona, it made me question things!

vrothberg: :rofl:

docs/source/markdown/podman-auto-update.1.md
vrothberg: Good catch. I just tested, the backslash isn't needed :)

libpod/container_api.go
mheon: I'm still iffy because `generate systemd` isn't directly integrated with Libpod right now. We're relying on external annotations to determine what unit we're in, which is iffy.

If we're doing events, I would want to see both the old and new container IDs in the event

vrothberg:
Added do-not-merge label to prevent this from getting in before cutting 1.8.1 - merging this PR would require a minor version bump (i.e., 1.9.0)

libpod/container_api.go
vrothberg: Then I prefer reverting the events changes.

vrothberg: Done.

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -0,0 +1,46 @@
...
+bc219740a210455fa27deacc96d50a9e20516492f1417507c13ce1533dbdcd9d
+
+# Generate a systemd unit for this container
+$ podman generate systemd --new --files bc219740a210455fa27deacc96d50a9e20516492f1417507c13ce1533dbdcd9d
+/home/user/containers/libpod/container-bc219740a210455fa27deacc96d50a9e20516492f1417507c13ce1533dbdcd9d.service
+
+# Load the new systemd unit and start it
+$ mv ./container-bc219740a210455fa27deacc96d50a9e20516492f1417507c13ce1533dbdcd9d.service ~/.config/systemd/user
+$ systemctl --user daemon-reload
+$ systemctl --user start container-bc219740a210455fa27deacc96d50a9e20516492f1417507c13ce1533dbdcd9d.service
```
edsantiago: This doesn't actually work as described, but it's unrelated to your PR.

First: on my test setup (f31, root), I have to `restorecon /etc/systemd/system/xxx.service`, otherwise systemctl says "no such unit".

Second, if the initial source container is started with `--name foo`, `systemctl start` will constantly fail until you `podman rm -f` the original source container:
```
Mar 18 08:22:40 ci-vm-10-0-137-242.hosted.upshift.rdu2.redhat.com podman[68087]: Error: error creating container storage: the container name "mytest" is already in use by "5d6d296390e453b59a9e109a8de78369949f1de0d2fedcb3c6c0568d4e0ceafd". You have to remove that container to be able to reuse that name.: that name is already in use
```

vrothberg: > First: on my test setup (f31, root), I have to restorecon /etc/systemd/system/xxx.service, otherwise systemctl says "no such unit".

Shouldn't it go to `/usr/lib/systemd/system/` for system services? That's where I place them and selinux behaved.

> Second, if the initial source container is started with --name foo, systemctl start will constantly fail until you podman rm -f the original source container.

That's a super helpful observation! I think, we should not restart but stop-start the services, so that everything gets properly cleaned up. I'll investigate.

Thanks a lot for testing and the feedback, @edsantiago !

edsantiago: >Shouldn't it go to /usr/lib/systemd/system/ for system services?

Probably not: that's for installed packages; the recommendation is to use `/etc/systemd/system` for local configuration. Regardless, if I (root) am sitting in my home directory when I run `podman generate systemd --files`, then `mv` that file elsewhere, SELinux will have problems because the file will be labeled `admin_home_t` instead of `systemd_unit_file_t`.

This brings up a gripe I've long had, though: why does podman write the unit file to pwd? That causes pain for users who then have to figure out the correct destination, mv it, restorecon, etc. Is it too late to fix podman so it actually figures out the correct root/rootless path and writes the file to the proper place?

>we should not restart but stop-start the services

As I recall, `systemctl stop podman-xx.service` did not help (perhaps because of the `--cidfile` option). Only `podman rm -f` (or `podman stop, podman rm`) helped.

vrothberg: > Second, if the initial source container is started with --name foo, systemctl start will constantly fail until you podman rm -f the original source container.

Misread the sentence. Yes, the initial container will remain. We want to add `--replace` which will remove existing containers in case of a name collision.

vrothberg:
@mheon, I think we can merge it. We marked it as experimental (same as v2) which doesn't require a minor version bump. Once that's in (and my todo-list is smaller) I will tackle the remaining bits (tests, better docs for systemd in general and --replace).

mheon:
Alright, SGTM

edsantiago:
Aside from the [file path and name issues](https://github.com/containers/libpod/pull/5480#discussion_r394314003), and the lack of a fallback mechanism (which I assume will be addressed some day), LGTM.

mheon:
Code LGTM

rhatdan:
@edsantiago @vrothberg Units should go under /etc not in /usr/lib. /usr/lib is for the distro, /etc/ is for the admin.

Also the reason you have an SELinux issue is the difference between mv and cp.  If you cp to the destination directory, it will be labeled correctly.  If you mv then it will maintain the label that the file was created in. If you created the unit file in your homedir, and mv it to /etc/systemd/system, then it will look to SELinux as if it s user homedir content.

Back in 2012 Dan Walsh wrote about this.
https://danwalsh.livejournal.com/56534.html

mv -Z /src/ /dest

Will do the thing you expected

edsantiago:
Thanks @rhatdan . Is it fair to say that a regular user should not have to worry about those things? I have opened #5538 with the intention that podman take care of that automatically.

PR #5482, opened 2020-03-12
remove imagefilter for varlink remote client

baude:
the api for getting images changed to use filters but back level podman versions cannot handle it.  so temporarily disabling the filtering for remote clients until the restful approach can be used.


baude:
setting to WIP because i think the first pass here will fail tests and i need to figure out which remote tests should be temporarily disabled.

PR #5484, opened 2020-03-12
turn off color-mode for bindings

baude:
the binding ginkgo tests were using color mode which throws in a bunch of ansi garbage that makes it hard to read the logs

### DIFF
--- a/contrib/cirrus/integration_test.sh
+++ b/contrib/cirrus/integration_test.sh
@@ -45,7 +45,7 @@ case "$SPECIALMODE" in
     bindings)
 	    make
         make install PREFIX=/usr ETCDIR=/etc
-	    cd pkg/bindings/test && ginkgo -r
+	    cd pkg/bindings/test && ginkgo -trace -noColor -debug  -r
 	;;
     none)
         make


edsantiago:
LGTM. I'll check in again in a while and see what the resulting log looks like.

FWIW the `make localintegration` target also includes `-progress`; I have no clue what that does but am tossing it out in case you think it could be useful.

edsantiago:
Restarted two known-flake runs; am leaving [this one](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/6227193546145792/html/integration_test.log.html) because it has errors I haven't seen before:
```
Error: cannot listen on the TCP port: listen tcp4 :5000: bind: address already in use
...
Error: Error copying image to the remote destination: Error trying to reuse blob sha256:5216338b40a7b96416b8b9858974bbe4acc3096ee60acbc4dfb1ee02aecceb10 at destination: error pinging docker registry localhost:5000: Get https://localhost:5000/v2/: x509: certificate signed by unknown authority
```

baude:
@edsantiago to be fair, i think those are true integration tests and not bindings correct? if so, there should be no corollary between this PR and those tests.

edsantiago:
Oops - no, I meant that I wanted someone else to look at those before they vanish into the ether. We seem to be gaining flakes, and I want wide awareness of them.

Sorry - didn't mean to imply causation (I see no possible way in which your PR could possibly cause those failures).

rhatdan:
@edsantiago The listen on socket 5000, looks like multiple tests running simultaneously?  Two containers trying to grate socket 5000? Just a guess, If yes, could we just randomize the port and then have a lot less chance of hitting the flake?

edsantiago:
Could be: I don't have a sense for how (if) these tests run in parallel. All of the other failures look troublesome, though: the iptables one, the sleep-1 container that didn't exit after 90s. But I'm not going to worry about them this evening.

Issue #5454, opened 2020-03-11
Podman stops accepting terminal input


computator:
/kind bug

**Description**

Podman seems to intermittently corrupt the terminal somehow while running containers. Once this happens, any new containers will not accept any input or the detach keys. The only way to regain control is by stopping the container from another terminal. When the container exits, all the keys typed into it are dumped to the shell which then causes "command not found" or other errors depending on which keys had been pressed. This issue continues to happen for any new containers started in the same terminal window. Resetting the terminal doesn't fix it, but it can be fixed either by starting a new subshell or by closing the terminal window and opening a new one.

**Steps to reproduce the issue:**

1. Start a container attached to the terminal.

2. Use the container some and then exit it

3. Repeat steps 1-2 an indeterminate number of times. At some point a newly started container will not accept any input.

4. Stop the container from another terminal (`podman stop -l`)

5. Run any container in the original terminal window (Ex: `podman run --rm -ti ubuntu`) and see it still will not accept any input.

**Additional information you deem important (e.g. issue happens only occasionally):**

I have run across this multiple times while developing and testing various containerized apps. I believe it's happened in both root and rootless mode, but I mostly use rootless. Once the issue is triggered neither mode works in that terminal window as tested with:
`podman run --rm -ti ubuntu` and `sudo podman run --rm -ti ubuntu`.

**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.10.1
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.10.1
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v1
  Conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.11, commit: '
  Distribution:
    distribution: ubuntu
    version: "18.04"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 23185395712
  MemTotal: 33688653824
  OCIRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
[...]

PR #5496, opened 2020-03-13
eat signal 23 in signal proxy

baude:
due to a change in golang-1.14 and it's changes to make go funcs with tight loops preemptive, signals are now getting "through" that never were before.

From the golang-1.14 announce:

Goroutines are now asynchronously preemptible. As a result, loops without function calls no longer potentially deadlock the scheduler or significantly delay garbage collection. This is supported on all platforms except windows/arm, darwin/arm, js/wasm, and plan9/*.

A consequence of the implementation of preemption is that on Unix systems, including Linux and macOS systems, programs built with Go 1.14 will receive more signals than programs built with earlier releases. This means that programs that use packages like syscall or golang.org/x/sys/unix will see more slow system calls fail with EINTR errors. Those programs will have to handle those errors in some way, most likely looping to try the system call again. For more information about this see man 7 signal for Linux systems or similar documentation for other systems.

### DIFF
--- a/pkg/adapter/sigproxy_linux.go
+++ b/pkg/adapter/sigproxy_linux.go
@@ -20,7 +20,10 @@ func ProxySignals(ctr *libpod.Container) {
 		for s := range sigBuffer {
 			// Ignore SIGCHLD and SIGPIPE - these are mostly likely
 			// intended for the podman command itself.
-			if s == syscall.SIGCHLD || s == syscall.SIGPIPE {
+			// SIGURG was added because of golang 1.14 and its preemptive changes
+			// causing more signals to "show up".
+			// https://github.com/containers/libpod/issues/5483
+			if s == syscall.SIGCHLD || s == syscall.SIGPIPE || s == syscall.SIGURG {
 				continue
 			}
 


pkg/adapter/sigproxy_linux.go
```diff
@@ -20,7 +20,9 @@ func ProxySignals(ctr *libpod.Container) {
 		for s := range sigBuffer {
 			// Ignore SIGCHLD and SIGPIPE - these are mostly likely
 			// intended for the podman command itself.
-			if s == syscall.SIGCHLD || s == syscall.SIGPIPE {
+			// SIGURG was added because of golang 1.14 and its preemptive changes
+			// causing more signals to "show up".
```
mheon: Can you link https://github.com/containers/libpod/issues/5454 in here so in the future we know what's going on?

mheon:
One nit, otherwise LGTM

pkg/adapter/sigproxy_linux.go
baude: done

pkg/adapter/sigproxy_linux.go
```diff
@@ -20,7 +20,10 @@ func ProxySignals(ctr *libpod.Container) {
...
+			// https://github.com/containers/libpod/issues/5483
```
TomSweeneyRedHat: might be good to also add #5454

TomSweeneyRedHat:
One suggestion, LGTM otherwise.  Nice work Mr @baude

rhatdan:
LGTM

mheon:
Mergebot must be blocked somewhere. Looking...

PR #5507, opened 2020-03-15
refactor info

baude:
the current implementation of info, while typed, is very loosely done so.  we need stronger types for our apiv2 implmentation and bindings.

### DIFF
--- a/libpod/define/info.go
+++ b/libpod/define/info.go
@@ -1,0 +1,101 @@
+package define
+
+import "github.com/containers/storage/pkg/idtools"
+
+// Info is the overall struct that describes the host system
+// running libpod/podman
+type Info struct {
+	Host       *HostInfo              `json:"host"`
+	Store      *StoreInfo             `json:"store"`
+	Registries map[string]interface{} `json:"registries"`
+}
+
+//HostInfo describes the libpod host
+type HostInfo struct {
+	Arch           string                 `json:"arch"`
+	BuildahVersion string                 `json:"buildahVersion"`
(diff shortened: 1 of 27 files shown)


baude:
This PR will not build nor pass any tests.

baude:
@mheon I know we have talked about this for a while ... the current state of info will soon block finalization of the apiv2.  I'm only looking for general confirmation here of what you think.  An actual review can come once I finish this PR.  If you don't like the approach, it makes no sense to go further.

mheon:
I'm OK with leaving it in Libpod; it seems like a reasonable API endpoint for us.

Overall approach LGTM here.

edsantiago:
I'm not really keen on the idea of breaking backward compatibility, but I missed the discussion yesterday. If it's absolutely necessary, here is a [patch to get system tests passing](https://paste.centos.org/view/a7702c6a). I believe there is still a real problem with the `info` command, though:
```console
$ ./bin/podman info --format '{{.host.arch}}'
Error: template: image:1:7: executing "image" at <.host.arch>: can't evaluate field host in type main.infoWithExtra```
$ /usr/bin/podman info --format '{{.host.arch}}'
amd64
```
Fails for anything (AFAICT) inside `{{.anything}}`

libpod/define/info.go
```diff
@@ -0,0 +1,84 @@
+package define
+
+import "github.com/containers/storage/pkg/idtools"
+
+type Info struct {
```
mheon, line 5: All structs in here are going to need doc comments

Issue #5503, opened 2020-03-15
Os is empty string


vertti:
/kind bug

**Description**

Originally filed it to buildah here https://github.com/containers/buildah/issues/2225 but it seems not to be a buildah issue.

**Steps to reproduce the issue:**

`podman build` produces images that when inspected show `Os` field as empty `""` instead of expected `linux`.

OpenShift refuses to run these containers and prints out:
```
Image operating system mismatch: image uses "", expecting "linux"
```

**Steps to reproduce the issue:**
1. Use `Dockerfile` like:
```
FROM node:12-buster

ADD package.json .
```
2. run `podman build -t website .`
3. run `podman inspect website`

**Describe the results you received:**

Check `Os` field in the output. It's `""` instead of `"linux"`

**Describe the results you expected:**

Check `Os` field in the output. It's `""` instead of `"linux"`
Running for example `podman inspect node:12-buster` will show `        "Os": "linux",`

**Additional information you deem important (e.g. issue happens only occasionally):**

While discussing on the `buildah` issue tracker, it seems buildah will get the correct `Os` field:

I also get the correct `"linux"` result with your `buildah` command. But `podman` still prints empty string for `Os`.

```
$ cat Dockerfile 
FROM node:12-buster

ADD package.json .
```
```
$ sudo buildah bud -t tom -f Dockerfile --layers .
STEP 1: FROM node:12-buster
STEP 2: ADD package.json .
--> Using cache c07c461de47b57ef1588a2456c70144ef465b90f19559f5c53e8b798fcf0a195
STEP 3: COMMIT tom
--> c07c461de47
c07c461de47b57ef1588a2456c70144ef465b90f19559f5c53e8b798fcf0a195
```
```
$ sudo buildah inspect --format '{{.Docker.OS}}' tom
linux
```
but continuing with `podman` you can see the layer hashes match:
```
$ sudo podman build -t foo .
STEP 1: FROM node:12-buster
STEP 2: ADD package.json .
--> Using cache c07c461de47b57ef1588a2456c70144ef465b90f19559f5c53e8b798fcf0a195
STEP 3: COMMIT foo
--> c07c461de47
c07c461de47b57ef1588a2456c70144ef465b90f19559f5c53e8b798fcf0a195
```
and the `Os` is still empty
```
$ sudo podman inspect foo
[
    {
        "Id": "c07c461de47b57ef1588a2456c70144ef465b90f19559f5c53e8b798fcf0a195",
[...]

PR #5508, opened 2020-03-15
add os|arch attributes when building

baude:
when building images, we can now add the os and arch of the image using overrides from the commandline.  the commandline options set sane defaults so we use those as well.

Fixes: #5503


mheon:
I like simple fixes. LGTM

PR #5516, opened 2020-03-16
Cirrus: Remove darwin/windows builds in gate-job

cevich:
It's advisable to have the initial gating job execute as quickly as
possible, weeding out simple mistakes early on, when possible.  However,
over time it has bloated to duplicate some more specific testing which
occurs in other tasks.  In this specific case the
`special_testing_cross` task.  Remove these duplicate items from the gate
job to speed things up for everyone.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -139,8 +139,6 @@ gating_task:
         - '/usr/local/bin/entrypoint.sh clean podman-remote |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh clean podman xref_helpmsgs_manpages BUILDTAGS="exclude_graphdriver_devicemapper selinux seccomp" |& ${TIMESTAMP}'
         - '/usr/local/bin/entrypoint.sh local-cross |& ${TIMESTAMP}'
-        - '/usr/local/bin/entrypoint.sh podman-remote-darwin |& ${TIMESTAMP}'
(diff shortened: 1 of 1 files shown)


cevich:
***Note:*** I'm not positive of the reason these cross-compile tasks were added to the gate job.  If it's important, I'm fine with just closing/abandoning this PR.

rhatdan:
@edsantiago @mheon @baude PTAL

edsantiago:
Are you sure the `local-cross` step is a dup? It has only been in place for three weeks (added by @baude in #5450) and IMHO it is a great addition, given the nightmares we've had with podman builds breaking due to 32-bit incompatibilities. I'm OK with moving it out into a separate task, but I don't think it should be removed.

rhatdan:
Yes we still need to run the cross check, I agree that this might not need to be gating.

cevich:
Thanks for the opinions @rhatdan and @edsantiago I'll leave the local-cross task there and just remove the true duplicates...

cevich:
@rhatdan @edsantiago PTAL.  Tests all green. Better now?

Looks like this shaves about 1.5 minutes off the `gate` job, so a bit minor, but something that affects all contributors.

edsantiago:
/lgtm

thank you!

rhatdan:
/approve
I will take 1.5 minutes, thanks.

PR #5439, opened 2020-03-10
systemd generator: force run container detached if CreateCommand has no detach param

ttys3:
since we use systemd service `Type=forking`
when we generated systemd service file using the `--new` param like
```
 sudo podman generate systemd --name --new ctrName
```
we got `ExecStart=/usr/bin/podman run --conmon-pidfile ...`
if somehow the container's `CreateCommand` has no `-d` or `--detach` param,
podman will run the container in default attached mode.
so systemd will wait the podman command exit until failed with timeout error.

quick reproduce the problem:
```
sudo podman pull nginxdemos/hello
sudo podman create --name ngxdemo -p 8086:80 nginxdemos/hello
cd /etc/systemd/system/
sudo podman generate systemd --name --new --restart-policy on-failure -t 3 --files ngxdemo
sudo podman rm ngxdemo

sudo systemctl start container-ngxdemo.service
Job for container-ngxdemo.service failed because a timeout was exceeded.
See "systemctl status container-ngxdemo.service" and "journalctl -xe" for details.
```

how to avoid the problem:

solution 1: merge this PR -_- and everything goes fine

solution 2: warn user to remember add `-d` when create the container
and user need to do is just add the `-d` param to the `create` command:
```
sudo podman pull nginxdemos/hello
sudo podman create -d --name ngxdemo -p 8086:80 nginxdemos/hello
cd /etc/systemd/system/
sudo podman generate systemd --name --new --restart-policy on-failure -t 3 --files ngxdemo
sudo podman rm ngxdemo
sudo systemctl start container-ngxdemo.service
# everything goes fine
```

PR #5514, opened 2020-03-16
fix systemd generate tests

vrothberg:
Add the `default.target` to the failing tests.

Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

PR #5517, opened 2020-03-16
Emergency fix for pr #5439

edsantiago:
PR collision with #5427

Solution: add 'default.target' to WantedBy


edsantiago:
@vrothberg beat me to it; closing in favor of #5514

PR #5528, opened 2020-03-17
Fix vendoring on master

mheon:
It looks like something snuck in and now make vendor fails on master. This should make things happy.


TomSweeneyRedHat:
LGTM

PR #5536, opened 2020-03-18
podman: avoid conmon zombie on exec

giuseppe:
conmon forks itself, so make sure we reap the first process and not
leave a zombie process.


TomSweeneyRedHat:
Changes LGTM, a couple of tests aren't happy and didn't leave many bread crumbs behind.  I've restarted.

mheon:
Good catch, LGTM

PR #5542, opened 2020-03-18
Vendor in containers/buildah v1.14.3

rhatdan:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/containernetworking/cni v0.7.2-0.20200304161608-4fae32b84921
 	github.com/containernetworking/plugins v0.8.5
-	github.com/containers/buildah v1.14.3-0.20200313154200-d26f437b2a46
+	github.com/containers/buildah v1.14.3
 	github.com/containers/common v0.5.0
 	github.com/containers/conmon v2.0.10+incompatible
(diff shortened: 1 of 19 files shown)


rhatdan:
Need this for podman 1.8.2 release.
@mheon @TomSweeneyRedHat @QiWang19 @baude @giuseppe @vrothberg PTAL

rhatdan:
@edsantiago podman info should never show a -dev branch for Buildah.

mheon:
LGTM

edsantiago:
>@edsantiago podman info should never show a -dev branch for Buildah.

I'm missing some context here. I recall discussing this a few weeks ago, in the context of _a released podman_ should never show -dev. The implication I see in the above is that _no podman ever_ should show -dev, is that correct? If the latter is correct, that should be easy to test, and I can write a hook for it. For the former, the conversation stalled because I was not aware of a reliable way to identify  "this tag/build/pr is intended for release". @cevich is currently doing some promising work along those lines, though.

If my "no podman ever" interpretation is the right one, please let me know and I will add a CI test that blocks -dev in any committed buildah in any PR.

TomSweeneyRedHat:
LGTM once tests are happy.
@edsantiago  We should keep "buildah*-dev" out of a final Podman release, but I think it's OK on a rare occasion for upstream and even RC releases.  Just hope we don't slip up an let one slip through tp a final though.

rhatdan:
@edsantiago No this is the discussion we had a few weeks ago, wanted to make sure rc1 was failing your release testing.

PR #5545, opened 2020-03-18
swagger inconsistencies: fix /libpod/events

edsantiago:
Fix a confusing inconsistency between swagger and actual
registration. I think this is the intention, but please
review carefully.


edsantiago:
@baude @jwhonce PTAL; I suggest checking out into your local git tree, and/or using the expandos to view full context (i.e. the `//swagger` comments above). I believe this is the intended behavior, but I might be mistaken.

baude:
this is covered in https://github.com/containers/libpod/pull/5525 i believe

edsantiago:
Yep, hadn't seen that, thanks. Closing.

PR #5548, opened 2020-03-19
image prune skips images with child images.

kunalkushwaha:
While the image build process, intermediate images are created. These images are also used as cache images, used in rebuilding the same images. This fix the deletion of cache images.

Fixes: https://github.com/containers/libpod/issues/5444

 Signed-off-by: Kunal Kushwaha <kunal.kushwaha@gmail.com>

### DIFF
--- a/docs/source/markdown/podman-image-prune.1.md
+++ b/docs/source/markdown/podman-image-prune.1.md
@@ -11,6 +11,8 @@ podman-image-prune - Remove all unused images from the local store
 you can delete all unused images.  Unused images are dangling images as well as any image that
 does not have any containers based on it.
 
+The image prune command does not prune cache images that only use layers that are necessary for other images.
+
 ## OPTIONS
 **--all**, **-a**
 
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -32,10 +32,10 @@ import (
 	"github.com/containers/libpod/pkg/registries"
 	"github.com/containers/libpod/pkg/util"
 	"github.com/containers/storage"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/opentracing/opentracing-go"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -845,6 +845,26 @@ func (i *Image) History(ctx context.Context) ([]*History, error) {
 // Dangling returns a bool if the image is "dangling"
 func (i *Image) Dangling() bool {
 	return len(i.Names()) == 0
+}
+
+// Intermediate returns true if the image is cache or intermediate image.
+// Cache image has parent and child.
+func (i *Image) Intermediate(ctx context.Context) (bool, error) {
+	parent, err := i.IsParent(ctx)
+	if err != nil {
+		return false, err
+	}
+	if !parent {
+		return false, nil
+	}
+	img, err := i.GetParent(ctx)
+	if err != nil {
+		return false, err
+	}
+	if img != nil {
+		return true, nil
+	}
+	return false, nil
 }
 
 // Labels returns the image's labels
--- a/libpod/image/prune.go
+++ b/libpod/image/prune.go
@@ -57,7 +57,7 @@ func generatePruneFilterFuncs(filter, filterValue string) (ImageFilter, error) {
 }
 
 // GetPruneImages returns a slice of images that have no names/unused
-func (ir *Runtime) GetPruneImages(all bool, filterFuncs []ImageFilter) ([]*Image, error) {
+func (ir *Runtime) GetPruneImages(ctx context.Context, all bool, filterFuncs []ImageFilter) ([]*Image, error) {
 	var (
 		pruneImages []*Image
 	)
@@ -74,10 +74,6 @@ func (ir *Runtime) GetPruneImages(all bool, filterFuncs []ImageFilter) ([]*Image
 			}
 		}
 
-		if len(i.Names()) == 0 {
-			pruneImages = append(pruneImages, i)
-			continue
-		}
 		if all {
 			containers, err := i.Containers()
 			if err != nil {
@@ -85,7 +81,21 @@ func (ir *Runtime) GetPruneImages(all bool, filterFuncs []ImageFilter) ([]*Image
 			}
 			if len(containers) < 1 {
 				pruneImages = append(pruneImages, i)
+				continue
 			}
+		}
+
+		//skip the cache or intermediate images
+		intermediate, err := i.Intermediate(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if intermediate {
+			continue
+		}
+
+		if i.Dangling() {
+			pruneImages = append(pruneImages, i)
 		}
 	}
 	return pruneImages, nil
@@ -111,7 +121,7 @@ func (ir *Runtime) PruneImages(ctx context.Context, all bool, filter []string) (
 		filterFuncs = append(filterFuncs, generatedFunc)
 	}
 
-	pruneImages, err := ir.GetPruneImages(all, filterFuncs)
+	pruneImages, err := ir.GetPruneImages(ctx, all, filterFuncs)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to get images to prune")
 	}
--- a/test/e2e/prune_test.go
+++ b/test/e2e/prune_test.go
@@ -87,7 +87,7 @@ var _ = Describe("Podman prune", func() {
 		Expect(podmanTest.NumberOfContainers()).To(Equal(0))
 	})
 
-	It("podman image prune none images", func() {
+	It("podman image prune skip cache images", func() {
 		SkipIfRemote()
 		podmanTest.BuildImage(pruneImage, "alpine_bash:latest", "true")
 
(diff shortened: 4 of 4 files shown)


docs/source/markdown/podman-image-prune.1.md
```diff
@@ -11,6 +11,9 @@ podman-image-prune - Remove all unused images from the local store
 you can delete all unused images.  Unused images are dangling images as well as any image that
 does not have any containers based on it.
 
+It skips the cache images i.e. intermidiate images build during image building.
```
giuseppe: typo in `intermidiate`

docs/source/markdown/podman-image-prune.1.md
```diff
@@ -11,6 +11,9 @@ podman-image-prune - Remove all unused images from the local store
...
+Cache images gets deleted along with main image deletion.
```
giuseppe: s/gets/are/

rhatdan:
Tests are failing?

docs/source/markdown/podman-image-prune.1.md
```diff
@@ -11,6 +11,9 @@ podman-image-prune - Remove all unused images from the local store
 you can delete all unused images.  Unused images are dangling images as well as any image that
 does not have any containers based on it.
 
+It skips the cache images i.e. intermediate images build during image building.
```
TomSweeneyRedHat: Suggest: "The image prune command does not prune any cache image, i.e. intermediate images built during image creation"

If you don't take that, be sure to change "bulid" -> "built"

docs/source/markdown/podman-image-prune.1.md
```diff
@@ -11,6 +11,9 @@ podman-image-prune - Remove all unused images from the local store
...
+Cache images are deleted along with main image deletion.
```
TomSweeneyRedHat: deletion -> deletion when using the podman rmi command.

and if you go with that, we probably ought to add a pointer to podman rmi.

rhatdan:
@mheon @baude @TomSweeneyRedHat @vrothberg @giuseppe @jwhonce @QiWang19 PTAL

libpod/image/image.go
```diff
@@ -847,6 +847,23 @@ func (i *Image) Dangling() bool {
 	return len(i.Names()) == 0
 }
 
+// IsCacheImage returns true if the image is cache or intermediate image.
+// Cache image has parent and child.
```
vrothberg: I am note 100 percent sure about that. @mtrmac @nalind , WDYT?

mtrmac: First, the function is misnamed. Let’s not confuse ourselves like that.

---

The bug report says
> Basically, docker image prune does not remove any images which have dependent tagged images.

This is clearly not implementing _that_, e.g. a chain of 5 derived dangling images will AFAICS cause only the most-derived image to be pruned, when all 5 should be pruned.

---

(I would also expect an O(N) mark&sweep GC over the layer tree (we are essentially pruning untagged _layers_, leaving any attached image objects around, after all, and relative parent/child relationships for the same `TopLayer` don’t really matter), this is O(N^2). Or, at the _very least_, have this `IsCacheImage` bail out early if `!parent` instead of unnecessarily computing `GetParent`.)

TomSweeneyRedHat:
@kunalkushwaha needs a rebase

libpod/image/image.go
kunalkushwaha: > This is clearly not implementing that, e.g. a chain of 5 derived dangling images will AFAICS cause only the most-derived image to be pruned, when all 5 should be pruned.

This is not clear to me.
The bug is `podman images prune` should not prune any image( tagged or untagged) which is used as an intermediate image for building finally a tagged image, So while re-build process of the tagged image, all intermediate images should be readily used as cache image.

How to find an image is an intermediate/cache image.
-  image must have child and parent image.
 - This is implemented in `IsCacheImage()` function.

How will the (chain of ) dangling images will be handled?
- If an image has no label and has no child (image derived from this) will be deleted.
- dangling image is formed when the image rebuild happens multiple times, and it leads to the previous target image untagged and not used.
- So dangling images will not have chain of dangling images. The leaf node of the image tree will be dangling.

The above implementation is based on the above points.

mtrmac: I think you are saying the same thing: to delete a chain of dangling images, one has to run the proposed implementation N times, where N is the length of the chain.

mtrmac: …and as to how a chain of dangling images can be created, consider a change to one of the early lines of a Dockerfile, not the last line. At the point that early line is processed, a new intermediate image is created, and all following images are children of that new intermediate image. The previous intermediate image, and all its children, will be dangling, as a result of that one build.

kunalkushwaha: thanks for explanation, got your point.

Those chain of dangling or dependent images used only by dangling images get deleted along with deletion of dangling image(leaf node). This is taken care by `image.Remove()` function which is used for removal of image.
https://github.com/containers/libpod/blob/e1ffac6cc73eb36640cbaf6a1a28ba44749a96d9/libpod/image/image.go#L426 
It deletes all child images which are not base image for any other image. https://github.com/containers/libpod/blob/e1ffac6cc73eb36640cbaf6a1a28ba44749a96d9/libpod/image/image.go#L444

mtrmac: Oh, I didn’t realize the list generated by GetPruneImages is not the only images that are deleted. OK, then this probably does what’s necessary.


(I still object to naming a function that returns `true` not only on cache images `IsCache`, that’s going to horribly confuse someone eventually.)

kunalkushwaha: > (I still object to naming a function that returns true not only on cache images IsCache, that’s going to horribly confuse someone eventually.)

Will `image.Intermediate` or `image.HasParentAndChild` will work?

libpod/image/image.go
```diff
@@ -847,6 +847,25 @@ func (i *Image) Dangling() bool {
 	return len(i.Names()) == 0
 }
 
+// IsCache returns true if the image is cache or intermediate image.
+// Cache image must have parent and child.
+func (i *Image) IsCache(ctx context.Context) bool {
+	parent, err := i.IsParent(ctx)
+	if err != nil || !parent {
+		return false
```
mtrmac: Is it safe to ignore `err` and allow the image to be deleted in that case?

libpod/image/image.go
mtrmac: Both work for me.

rhatdan:
@mtrmac PTAL

rhatdan:
sorry needs a rebase

docs/source/markdown/podman-image-prune.1.md
```diff
@@ -11,6 +11,8 @@ podman-image-prune - Remove all unused images from the local store
 you can delete all unused images.  Unused images are dangling images as well as any image that
 does not have any containers based on it.
 
+The image prune command does not prune any cache image, i.e. intermediate images built during image creation.
```
mtrmac, line 14: This seems incorrect; it _will_ prune intermediate images that are not parents of a tagged image, won’t it?

libpod/image/image.go
```diff
@@ -847,6 +847,25 @@ func (i *Image) Dangling() bool {
...
+	if err != nil || !parent {
+		return false, err
+	}
+
+	img, err := i.GetParent(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	if img != nil && parent {
```
mtrmac, line 863: The `parent` check is now redundant.

libpod/image/prune.go
```diff
@@ -87,6 +83,19 @@ func (ir *Runtime) GetPruneImages(all bool, filterFuncs []ImageFilter) ([]*Image
...
+		cache, err := i.Intermediate(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if cache {
+			continue
+		}
+
+		if i.Dangling() {
+			pruneImages = append(pruneImages, i)
```
mtrmac, line 97: `pruneImages` may now contain the same image twice if `all == true`.

(Non-blocking: Overall I’d prefer the control flow to cleanly separate the decision (“should this be pruned?”) from the consequence (“add to `pruneImages`”), not intermix them on the various code paths.)

kunalkushwaha: This is also fixed now.

rhatdan:
@mtrmac PTAL

docs/source/markdown/podman-image-prune.1.md
mtrmac: Still outstanding.

libpod/image/image.go
```diff
@@ -847,6 +847,27 @@ func (i *Image) Dangling() bool {
...
+		return false, err
+	}
+
+	img, err := i.GetParent(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	// note: image have already parent.
+	// if image have children, it is Intermediate image.
```
mtrmac, line 864:
```suggestion
	// Note: The image is already know to have a parent.
	// If the image also has children, it is an intermediate image.
```

docs/source/markdown/podman-image-prune.1.md
kunalkushwaha: This is fixed. 
updated text is 
```
The image prune command does not prune any valid cache image, i.e. intermediate images built during tagged image creation.
```

mtrmac: I don’t understand “valid” in that context; what would an “invalid” cache image be?

Either way the “i.e.” explanation for what the “valid cache image” means does not match what the code does.

kunalkushwaha: Writing too much explanation like explaining valid image may not look good. Will removing valid will fix it?

mtrmac: No, that just removes an attempt to fix the problem.

How about something like “does not prune cache images that only use layers that are necessary for other uses”?

kunalkushwaha: Updated it.

kunalkushwaha:
@mtrmac PTAL. All issues are resolved now.

kunalkushwaha:
> were fixed previously but now are in the code again.
> 
> What is going on here? AFAICT it is not a caching issue on my end.

I am not sure, how that happens. I have updated all of the required changes. PTAL.

rhatdan:
@mtrmac PTAL again.

PR #5550, opened 2020-03-19
rootless: fix usage with hidepid=1

giuseppe:
when /proc is mounted with hidepid=1 a process doesn't see processes
from the outer user namespace.  This causes an issue reading the
cmdline from the parent process.

To address it, always read the command line from /proc/self instead of
using /proc/PARENT_PID.


rhatdan:
LGTM

Issue #5553, opened 2020-03-19
APIv2 incompatibilities with docker-py


max-arnold:
/kind bug

**Description**

I'm trying to make the [SaltStack Docker driver](https://docs.saltstack.com/en/master/ref/modules/all/salt.modules.dockermod.html) work with Podman APIv2 (the driver uses `docker-py` under the hood).

I managed to run simple Salt commands like `docker.version`, `docker.info` and `docker.pull`, but more complex ones like `docker.ps` or `docker.run_container` are failing. I traced the failures back to the underlying APIv2 calls that aren't compatible with Docker API:

1. `sudo curl --unix-socket /run/podman/podman.sock http://d/version`

It reports `"Version":"1.8.1"`, but the Salt module expects at least `1.9.0` to decide if it should work at all.  The problem is arguable and can be fixed on Salt side by checking for `Podman Engine` component name. However, I decided to report it anyway because other tools in the wild can also make some decisions based on the version number.

2. `sudo curl --unix-socket /run/podman/podman.sock 'http://localhost/v1.40/containers/json?limit=-1&all=1&size=0&trunc_cmd=0'`

`{"cause":"runtime error: slice bounds out of range","message":"runtime error: slice bounds out of range","response":500}`

This problem is caused by the `limit=-1` default parameter for `containers()` in docker-py: https://github.com/docker/docker-py/blob/master/docker/api/container.py#L151-L153

Docker API spec doesn't list `-1` as a valid number https://docs.docker.com/engine/api/v1.40/#tag/Container However, other calls use it, so I guess it is a common convention:

```
curl -s https://docs.docker.com/engine/api/v1.40.yaml | grep -- '-1' | grep limit
        description: "Total memory limit (memory + swap). Set as `-1` to enable unlimited swap."
          Tune a container's PIDs limit. Set `0` or `-1` for unlimited, or `null` to not change.
```

3. Creating a container

`sudo curl --unix-socket /run/podman/podman.sock --header "Content-Type: application/json" --data '{"Tty": false, "OpenStdin": false, "StdinOnce": false, "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Image": "72300a873c2ca11c70d0c8642177ce76ff69ae04d61a5813ef58d40ff66e3e7c", "NetworkDisabled": false, "HostConfig": {"NetworkMode": "default"}}' 'http://localhost/v1.40/containers/create?name=ubuntu'`

It returns the following response

[...]

PR #5554, opened 2020-03-19
fix reported compat issues

baude:
honor -1 in in list containers for compatibility mode.  it is commonly used to indicate no limit.

change the json id parameter to Id in container create.

Fixes: #5553

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -87,7 +87,7 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 		utils.InternalServerError(w, err)
 		return
 	}
-	if _, found := r.URL.Query()["limit"]; found {
+	if _, found := r.URL.Query()["limit"]; found && query.Limit != -1 {
 		last := query.Limit
 		if len(containers) > last {
 			containers = containers[len(containers)-last:]
--- a/pkg/api/handlers/utils/containers.go
+++ b/pkg/api/handlers/utils/containers.go
@@ -16,7 +16,7 @@ import (
 // ContainerCreateResponse is the response struct for creating a container
 type ContainerCreateResponse struct {
 	// ID of the container created
-	ID string `json:"id"`
+	ID string `json:"Id"`
 	// Warnings during container creation
 	Warnings []string `json:"Warnings"`
 }


rhatdan:
LGTM

rhatdan:
@mheon @jwhonce @edsantiago PTAL

TomSweeneyRedHat:
LGTM

max-arnold:
Are these other occurrencies of lowercase `id` also should be`Id`?

```
libpod/pkg/api % grep -r '"id"' .
./handlers/compat/images.go:		Id             string            `json:"id"`
./handlers/compat/images.go:		Id             string            `json:"id"`
./handlers/types.go:	ID string `json:"id"`
./handlers/types.go:	ID string `json:"id"`
./handlers/types.go:	ID string `json:"id"`
./handlers/types.go:	ID             string `json:"id"`
./handlers/types.go:	ID string `json:"id"`
./handlers/types.go:	ID     string       `json:"id"`
./handlers/utils/containers.go:	ID string `json:"id"`
```

baude:
> Are these other occurrencies of lowercase `id` also should be`Id`?
> 
> ```
> libpod/pkg/api % grep -r '"id"' .
> ./handlers/compat/images.go:		Id             string            `json:"id"`
> ./handlers/compat/images.go:		Id             string            `json:"id"`
> ./handlers/types.go:	ID string `json:"id"`
> ./handlers/types.go:	ID string `json:"id"`
> ./handlers/types.go:	ID string `json:"id"`
> ./handlers/types.go:	ID             string `json:"id"`
> ./handlers/types.go:	ID string `json:"id"`
> ./handlers/types.go:	ID     string       `json:"id"`
> ./handlers/utils/containers.go:	ID string `json:"id"`
> ```

the only questionable ones i see for certain are the compat ones ... does the consumer side of those expect Id?

pkg/api/handlers/compat/containers.go
```diff
@@ -87,7 +87,7 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 		utils.InternalServerError(w, err)
 		return
 	}
-	if _, found := r.URL.Query()["limit"]; found {
+	if _, found := r.URL.Query()["limit"]; found && query.Limit != -1 {
```
edsantiago: Wouldn't `>= 0` be safer?

rhatdan: Yes, then if user sends any negative number it will work.

max-arnold:
To be honest, I haven’t looked how those ids are named in the Docker API spec. It is just a list of other possible occurrences to cross-check

jwhonce:
> > Are these other occurrencies of lowercase `id` also should be`Id`?
> > ```
> > libpod/pkg/api % grep -r '"id"' .
> > ./handlers/compat/images.go:		Id             string            `json:"id"`
> > ./handlers/compat/images.go:		Id             string            `json:"id"`
> > ./handlers/types.go:	ID string `json:"id"`
> > ./handlers/types.go:	ID string `json:"id"`
> > ./handlers/types.go:	ID string `json:"id"`
> > ./handlers/types.go:	ID             string `json:"id"`
> > ./handlers/types.go:	ID string `json:"id"`
> > ./handlers/types.go:	ID     string       `json:"id"`
> > ./handlers/utils/containers.go:	ID string `json:"id"`
> > ```
> 
> the only questionable ones i see for certain are the compat ones ... does the consumer side of those expect Id?

If the /libpod tree doesn't follow the same naming WRT to common fields then bindings with case sensitive json keys won't port without a rewrite. A quick look this morning seems to be "id" for path queries and "Id" for responses.

PR #5555, opened 2020-03-19
Add basic structure of a spec generator for pods

mheon:
This will be used for remote creation of pods initially.


mheon:
@baude PTAL

PR #5566, opened 2020-03-20
Add podman static build

saschagrunert:
We’re now able to build a static podman binary based on a custom nix
derivation. This is integrated in cirrus as well, whereas a later target
would be to provide a self-contained static binary bundle which can be
installed on any Linux x64-bit system.

Fixes: https://github.com/containers/libpod/issues/1399

Test it out yourself:
```bash
> make build-static
…
> ldd bin/podman-static
        not a dynamic executable
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ LDFLAGS_PODMAN ?= \
 	  -X $(LIBPOD)/define.buildInfo=$(BUILD_INFO) \
 	  -X $(LIBPOD)/config._installPrefix=$(PREFIX) \
 	  -X $(LIBPOD)/config._etcDir=$(ETCDIR) \
-	  -extldflags "$(LDFLAGS)"
+	  $(EXTRA_LDFLAGS)
 #Update to LIBSECCOMP_COMMIT should reflect in Dockerfile too.
 LIBSECCOMP_COMMIT := v2.3.3
 # Rarely if ever should integration tests take more than 50min,
@@ -216,6 +216,30 @@ bin/podman.cross.%: .gopathok
 	GOOS="$${TARGET%%.*}" \
 	GOARCH="$${TARGET##*.}" \
 	$(GO_BUILD) -gcflags '$(GCFLAGS)' -asmflags '$(ASMFLAGS)' -ldflags '$(LDFLAGS_PODMAN)' -tags '$(BUILDTAGS_CROSS)' -o "$@" $(PROJECT)/cmd/podman
+
+# Update nix/nixpkgs.json its latest master commit
+.PHONY: nixpkgs
+nixpkgs:
+	@nix run -f channel:nixpkgs-unstable nix-prefetch-git -c nix-prefetch-git \
+		--no-deepClone https://github.com/nixos/nixpkgs > nix/nixpkgs.json
+
+NIX_IMAGE ?= quay.io/podman/nix-podman:1.0.0
+
+# Build the nix image as base for static builds
+.PHONY: nix-image
+nix-image:
+	$(CONTAINER_RUNTIME) build -t $(NIX_IMAGE) -f Containerfile-nix .
+
+# Build podman statically linked based on the default nix container image
+.PHONY: build-static
+build-static:
+	$(CONTAINER_RUNTIME) run \
+		--rm -it \
+		-v $(shell pwd):/work \
+		-w /work $(NIX_IMAGE) \
+		sh -c "nix build -f nix && \
+			   mkdir -p bin && \
+			   cp result-*bin/bin/podman bin/podman-static"
 
 .PHONY: run-docker-py-tests
 run-docker-py-tests:
--- a/README.md
+++ b/README.md
@@ -180,3 +180,9 @@ Podman offers a Varlink-based API for remote management of containers.
 However, this API has been deprecated by the REST API.
 Varlink support is in maintenance mode, and will be removed in a future release.
 For more details, you can see [this blog](https://podman.io/blogs/2020/01/17/podman-new-api.html).
+
+## Static Binary Builds
+The Cirrus CI integration within this repository contains a `static_build` job
+which produces a static Podman binary for testing purposes. Please note that
+this binary is not officially supported with respect to feature-completeness
+and functionality and should be only used for testing.
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -820,6 +820,7 @@ success_task:
         - "test_build_cache_images"
         - "verify_test_built_images"
         - "docs"
+        - "static_build"
         # FIXME remove when all v2 tests pass
         - "integration_test_temporary"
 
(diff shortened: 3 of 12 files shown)


saschagrunert:
Just have to make the CI happy :)

vrothberg:
> This is integrated in cirrus as well, whereas a later target
would be to provide a self-contained static binary bundle which can be
installed on any Linux x64-bit system.

I feel rather objective toward that idea. There's a whole load of issues with static binaries working around NSS (breaking LDAP etc.). We don't provide static builds for that reason. If we allow for static builds, we need to make clear that this is not supported. I think we need some way in `podman info` to see if it's a static binary; issues should only be filed for dynamic ones.

Makefile
```diff
@@ -213,6 +213,30 @@ bin/podman.cross.%: .gopathok
 	GOARCH="$${TARGET##*.}" \
 	$(GO_BUILD) -gcflags '$(GCFLAGS)' -asmflags '$(ASMFLAGS)' -ldflags '$(LDFLAGS_PODMAN)' -tags '$(BUILDTAGS_CROSS)' -o "$@" $(PROJECT)/cmd/podman
 
+# Update nix/nixpkgs.json its latest master commit
+.PHONY: nixpkgs
+nixpkgs:
+	@nix run -f channel:nixpkgs-unstable nix-prefetch-git -c nix-prefetch-git \
+		--no-deepClone https://github.com/nixos/nixpkgs > nix/nixpkgs.json
+
+NIX_IMAGE ?= quay.io/crio/nix-podman:1.0.0
```
TomSweeneyRedHat: could we have this live in quay.io/podman/nix-podman:1.0.0 instead?

saschagrunert: Yes, but I do not have write permissions to that repository :+1:

saschagrunert:
> > This is integrated in cirrus as well, whereas a later target
> > would be to provide a self-contained static binary bundle which can be
> > installed on any Linux x64-bit system.
> 
> I feel rather objective toward that idea. There's a whole load of issues with static binaries working around NSS (breaking LDAP etc.). We don't provide static builds of Skopeo for that reason. If we allow for static builds, we need to make clear that this is not supported. I think we need some way in `podman info` to see if it's a static binary; issues should only be filed for dynamic ones.

Do you recap which kind of issue that was? I'm not aware of any issues for example in conjunction with the CRI-O static binary bundle.

rhatdan:
Discussion here
https://github.com/containers/skopeo/pull/670

saschagrunert:
> Discussion here
> [containers/skopeo#670](https://github.com/containers/skopeo/pull/670)

Thanks. To me it looks like an usual bug in golang which has been fixed long time ago, right? I would not say that this falls into the category of: load of issues with static binaries. WDYT?

vrothberg:
> > Discussion here
> > [containers/skopeo#670](https://github.com/containers/skopeo/pull/670)
> 
> Thanks. To me it looks like an usual bug in golang which has been fixed long time ago, right? I would not say that this falls into the category of: load of issues with static binaries. WDYT?

I don't see that bypassing NSS has been resolved. @mtrmac will know.

rhatdan:
This is an issue of support.  It is likely that a static version of podman would not work well on a system that used nsswitch to modify the way host resolution worked.  Or how UID/GID lookups worked.
It would just fall back to reading /etc/hosts and /etc/resolv.conf, /etc/passwd, /etc/shadow.

This is definitely not something we would want to 'SUPPORT',   But I have no problem providing it for people to play with.

I agree that `podman info` should display that it was built staticly to prevent confusion.

saschagrunert:
> I agree that `podman info` should display that it was built staticly to prevent confusion.

Sounds good, I'll follow up with that on Monday if there are no further concerns.

mtrmac:
AFAIK the NSS situation remains the same (either dynamically link, or use a Go-only approximation of the default behavior, loosing all NSS plugins).

Separately, statically-building device-mapper and/or LVM, if nothing else, seems pretty scary. Is it safe to mix&match versions of libraries dealing with filesystems and user data on one system?

---

My first, fairly strong, instinct is to say ”no, get tested packages from your distribution” (and that includes Nix). Admittedly I’ve been living in the distribution world for ages, so I’m not at all neutral.

And we do see that whether one makes container builds or static builds difficult, one or the other tends to happen anyway just because users feel they need one of those. So it might be, after all, worthwhile to produce one of the two just to make sure most people use a known build instead of home-grown differently-patched one. Perhaps. I have no idea how prevalent are the _uses_ of the static/container builds, just that at least one person probably makes one.

Makefile
saschagrunert: Fixed as discussed offline.

saschagrunert:
I added an additional PodmanLinkmode field to the `podman info` output.

TomSweeneyRedHat:
Code LGTM, I'll leave it to others for the should/should not debate.  Personally I think it's fine as long as it's documented.

saschagrunert:
> I still have very mixed feelings about it and I frankly don't have time (or desire) to get into Nix, but as long as I don't have to maintain it, I am cool with it 
> 
> What I would love to have is some visible statement that the nix stuff is not supported. Which further makes me wonder why to offer something that is not supported but I don't want to be the party pooper and really want some more opinions.
> 
> @mheon @baude PTAL

:man_shrugging: I think if the majority does not want to have it then feel free to close it.

rhatdan:
Since this is an upstream project, support is best effort.  We can document the short comings of this version and state what does and does not work.  But if community wants it and is willing to deal with issues, then I have no problem doing it.

jwhonce:
@saschagrunert 

I have a request https://github.com/containers/libpod/issues/5765 points out `info` is already a bit slow.  Could you change `podmanLinkmode()` to `PodmanLinkMode()` and use a build tag to just report which mode?  You would need two files linkmode_dynamic.go and linkmode_static.go which are tagged to be mutually exclusive.  That way `info` can call and pay a minimal price.

rhatdan:
I just added podman-remote-static, which was needed for CRC,  @saschagrunert is this still something you are working on?

saschagrunert:
> I just added podman-remote-static, which was needed for CRC, @saschagrunert is this still something you are working on?

Yes I updated the PR and also the suggestion from @jwhonce. Should I state somewhere explicitly that the static binary is not supported? I mean if we not ship it and podman info returns the "static" linkmode we should be probably fine already, right?

vrothberg:
> Should I state somewhere explicitly that the static binary is not supported? 

Yes, please add a note in the README. If we don't make it explicit, users will not know.

README.md
```diff
@@ -180,3 +180,9 @@ Podman offers a Varlink-based API for remote management of containers.
 However, this API has been deprecated by the REST API.
 Varlink support is in maintenance mode, and will be removed in a future release.
 For more details, you can see [this blog](https://podman.io/blogs/2020/01/17/podman-new-api.html).
+
+## Static Binary Builds
+The Cirrus CI integration within this repository contains a `static_build` job
+which produces a static Podman binary for testing purposes. Please note that
+this binary is not officially supported with respect to feature-completeness
+and functionality and should be only used for testing.
```
saschagrunert, line 188: Added the requested note here. :point_up:

vrothberg: Great, thanks a lot!

PR #5576, opened 2020-03-20
when removing networks for tests, force should be used

baude:
when removing networks in integration tests, we should should force; otherwise if the network has containers associated with it, it will fail to remove.


rhatdan:
LGTM

rhatdan:
@mheon @giuseppe @vrothberg @TomSweeneyRedHat @jwhonce @QiWang19 PTAL

TomSweeneyRedHat:
LGTM

PR #5577, opened 2020-03-20
podmanv2 volume create

baude:
add volume create

### DIFF
--- a/cmd/podmanV2/volumes/create.go
+++ b/cmd/podmanV2/volumes/create.go
@@ -1,0 +1,72 @@
+package volumes
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/containers/libpod/cmd/podmanV2/parse"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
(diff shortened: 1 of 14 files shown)


cmd/podmanV2/volumes/create.go
```diff
@@ -0,0 +1,72 @@
+package volumes
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/containers/libpod/cmd/podman/shared/parse"
```
jwhonce: Need to copy the code, you cannot pull from cmd/podman.  When we're done that whole directory is going to be rm'ed.

pkg/domain/infra/abi/volumes.go
```diff
@@ -0,0 +1,38 @@
+// +build ABISupport
+
+package abi
+
+import (
+	"context"
+
+	"github.com/containers/libpod/cmd/podman/shared"
```
jwhonce: Cannot include code from cmd/podman, need to copy it out.

jwhonce:
LGTM once you get the tests passing

PR #5579, opened 2020-03-20
Make macOS unit tests runnable

mtrmac:
With this,
```console
% CGO_ENABLED=0 go test -tags='exclude_graphdriver_btrfs btrfs_noversion   exclude_graphdriver_devicemapper seccomp containers_image_openpgp' $(go list ./... | grep -v '^github.com/containers/libpod/\(test\|dependencies\|pkg/bindings/test\)')
```
succeeds.

(OTOH the `make` target, besides being much slower, eventually fails:
```console
% CGO_ENABLED=0 make localunit BUILDTAGS='exclude_graphdriver_btrfs btrfs_noversion   exclude_graphdriver_devicemapper seccomp containers_image_openpgp'
…
Failed to compile shm:

can't load package: package github.com/containers/libpod/libpod/lock/shm: build constraints exclude all Go files in /Users/mitr/Go/src/github.com/containers/libpod/libpod/lock/shm
```
)

This should probably be extended so that `make localunit` automatically does the right thing, but this is right now good enough for me to continue working.

See the individual commits for details.

### DIFF
--- a/cmd/podman/main_local_unsupported.go
+++ b/cmd/podman/main_local_unsupported.go
@@ -1,0 +1,44 @@
+// +build !remoteclient,!linux
+
+package main
+
+// The ONLY purpose of this file is to allow the subpackage to compile. Don’t expect anything
+// to work.
+
+import (
+	"syscall"
+
+	"github.com/spf13/cobra"
+)
+
+const remote = false
+
+func setSyslog() error {
+	return nil
+}
+
+func profileOn(cmd *cobra.Command) error {
+	return nil
+}
+
+func profileOff(cmd *cobra.Command) error {
+	return nil
+}
+
+func setupRootless(cmd *cobra.Command, args []string) error {
+	return nil
+}
+
+func setRLimits() error {
+	return nil
+}
+
+func setUMask() {
+	// Be sure we can create directories with 0755 mode.
+	syscall.Umask(0022)
+}
+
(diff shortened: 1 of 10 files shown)


mheon:
LGTM. Question: what are we doing that requires the test suite to run on OSX?

mtrmac:
> LGTM. Question: what are we doing that requires the test suite to run on OSX?

I’m working on the image library. That part IIRC already works well enough on Mac, but when updating callers outside libpod/image for changed APIs it’s a hassle not to be able to even compile-check the updates locally.

There _are_ alternatives, though (I can keep a don’t-even-look-at-it PR open to trigger CI, or push to a local VM for builds), so if this is not desirable (and the part that makes `cmd/podman` non-remote build on macOS may easily be undesirable), I’m fine with rejecting (parts of?) this PR.

mheon:
Changes seem fine, so merging in its entirety works for me.

rhatdan:
Should we be running this test all the time?

mtrmac:
> Should we be running this test all the time?

There _are_ tests for the remote builds on Mac. The local builds are pretty much nonsense, and so far I seem to be the only one needing them – and I right now don’t want to spend the time automating and polishing this. If anything, I’m more inclined to add an explicit “this is a local build on Mac, abort immediately” check to start of `main`.

Issue #5586, opened 2020-03-23
APIv2: `commit` badly attached and documented


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
I am trying to use `commit` from APIv2 and having hard time due to miserable documentation. (1) First issue is that it is attached to `/commit` instead of `/libpod/commit` which would be much better for consistent experience. (2) But the main issue is that it does not mention that the request needs some body, without the body it fails:

```
sudo curl -X POST --unix-socket /run/podman/podman.sock "http://d/v1.24/commit?container=44c004db4b17&comment=message&repo=myrepo"
{"cause":"EOF","message":"Decode(): EOF","response":500}
```

Passing just empty object, solves this issue:
```
sudo curl -X POST --unix-socket /run/podman/podman.sock -H "Content-Type: application/json" --data "{}" "http://d/v1.24/commit?container=44c004db4b17&comment=message&repo=myrepo&tag=1" 
{"cause":"no such container","message":"no container with name or ID 44c004db4b17 found: no such container","response":404}
```

(3) Also I think some parameters must be required (like `container`), but the doc does not mention it.
(4) Finally, is it possible to set up 'format' the same way I could do `podman commit --format <format>`?

Using current master and reading current documentation.

marusak:
Also, could I get example how to set up two changes? For example lets say I want to set up CMD to `bash -c "echo 'hello world'"` and USER to `admin`?
Thank you

mheon:
@baude @jwhonce PTAL

baude:
1. Added libpod endpoint in #5588 
2. Was a bug, fixed
3. fixed in swagger docs
4. yup, i added that.
5. sudo curl -X POST --unix-socket /run/podman/podman.sock 'http://d/v1.24/libpod/commit?author=baude&changes=CMD%3D%2Fbin%2Ffoobar&container=foo&repo=quay.io%2Fbaude%2Fdunbar&tag=latest

marusak:
Thank you @baude for the quick fixes!
ad 5. I was mostly asking how is it possible to pass more than one `change` (comma separated, space separated, some kind of array...)? And also how it should be encoded when the value is more than one string (sometimes these things are passed as arrays, sometimes as encoded strings...)

PR #5588, opened 2020-03-23
apiv2 container commit for libpod

baude:
add an endpoint for a libpod specific commit.  add the abilility to define the format.  also update documentation accordingly.

Fixes: #5586

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -10,12 +10,15 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/containers/buildah"
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/libpod/libpod"
 	"github.com/containers/libpod/libpod/image"
+	image2 "github.com/containers/libpod/libpod/image"
 	"github.com/containers/libpod/pkg/api/handlers"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/containers/libpod/pkg/util"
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@mheon @jwhonce @marusak PTAL

pkg/api/handlers/libpod/images.go
```diff
@@ -416,3 +419,84 @@ func ImagesPull(w http.ResponseWriter, r *http.Request) {
...
+	var (
+		destImage string
+		mimeType  string
+	)
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+
+	query := struct {
+		Author    string `schema:"author"`
+		Changes   string `schema:"changes"`
```
mheon: If we're doing our own endpoint, let's restore sanity and make this a []string

pkg/api/handlers/libpod/images.go
```diff
@@ -416,3 +419,84 @@ func ImagesPull(w http.ResponseWriter, r *http.Request) {
...
+		PreferredManifestType: mimeType,
+	}
+
+	if len(query.Tag) > 0 {
+		tag = query.Tag
+	}
+	options.Message = query.Comment
+	options.Author = query.Author
+	options.Pause = query.Pause
+	options.Changes = strings.Fields(query.Changes)
```
mheon: This doesn't work, changes can contain whitespace.

pkg/bindings/containers/types.go
```diff
@@ -11,3 +11,16 @@ type LogOptions struct {
 	Timestamps *bool
 	Until      *string
 }
+
+// CommitOptions describe details about the resulting commited
+// image as defined by repo and tag. None of these options
+// are required.
+type CommitOptions struct {
+	Author  *string
+	Changes *string
```
mheon: This should be a []string to match the CLI.

mheon:
Doesn't Docker allow a full image config be passed in the body? That seems useful, to avoid the expensive and uncertain parsing of Change

baude:
> Doesn't Docker allow a full image config be passed in the body? That seems useful, to avoid the expensive and uncertain parsing of Change

We dont have the capability to do that ...

pkg/api/handlers/libpod/images.go
```diff
@@ -416,3 +419,84 @@ func ImagesPull(w http.ResponseWriter, r *http.Request) {
...
+		Format: "oci",
+	}
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	rtc, err := runtime.GetConfig()
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Decode()"))
```
TomSweeneyRedHat: I still don't like the "Something went wrong" line.

mheon: We also have `utils.InternalServerError()` that abstracts this away...

PR #5591, opened 2020-03-23
podmanv2 exit code

baude:
add ability to set and get exit code.

### DIFF
--- a/cmd/podmanV2/root.go
+++ b/cmd/podmanV2/root.go
@@ -6,6 +6,7 @@ import (
 	"path"
 
 	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/version"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


cmd/podmanV2/root.go
```diff
@@ -30,7 +31,14 @@ func init() {
...
-		fmt.Println(err)
-		os.Exit(1)
+		fmt.Fprintln(os.Stderr, "Error:", err.Error())
+	} else if registry.GetExitCode() == define.ExecErrorCodeGeneric {
+		// The exitCode modified from define.ExecErrorCodeGeneric,
+		// indicates an application
+		// running inside of a container failed, as opposed to the
+		// podman command failed.  Must exit with that exit code
+		// otherwise command exited correctly.
+		registry.SetExitCode(0)
```
rhatdan: Why would you set the exit code to 0?

baude: i didnt ....
```67165b7675 cmd/podman/main.go (Valentin Rothberg 2020-01-13 13:01:45 +0100 170)                 exitCode = 0
```

rhatdan: @vrothberg PTAL

vrothberg: Commit 67165b7675 did _not_ set that but it removed a redundant check (tautology) to make a linter happy. Digging a bit deeper in the git history brings us to a Mr. Daniel J Walsh with commit b231e3412e7d :smile:

vrothberg: As far as I read it, we reserved `125` for for podman failures (see podman(1)). So this check seems like a safety net to make sure we don't return `125` when podman behaved correctly but the container might return `125`. In that case, we're losing the container's exit code.

rhatdan: Ok, I hope that Daniel J Walsh was correct...

rhatdan:
LGTM
@vrothberg @mheon @TomSweeneyRedHat @jwhonce @QiWang19 PTAL

cmd/podmanV2/root.go
vrothberg: He usually is :^)

PR #5593, opened 2020-03-23
podmanv2 volumes

baude:
add volume commands: create, inspect, ls, prune, and rm

### DIFF
--- a/cmd/podmanV2/volumes/inspect.go
+++ b/cmd/podmanV2/volumes/inspect.go
@@ -1,0 +1,74 @@
+package volumes
+
+import (
+	"encoding/json"
+	"fmt"
+	"html/template"
+	"os"
+
+	"github.com/containers/buildah/pkg/formats"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+)
+
+var (
+	volumeInspectDescription = `Display detailed information on one or more volumes.
+
+  Use a Go template to change the format from JSON.`
+	inspectCommand = &cobra.Command{
+		Use:   "inspect [flags] VOLUME [VOLUME...]",
+		Short: "Display detailed information on one or more volumes",
+		Long:  volumeInspectDescription,
+		RunE:  inspect,
+		Example: `podman volume inspect myvol
+  podman volume inspect --all
+  podman volume inspect --format "{{.Driver}} {{.Scope}}" myvol`,
+	}
+)
+
+var (
+	inspectOpts   = entities.VolumeInspectOptions{}
+	inspectFormat string
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: inspectCommand,
+		Parent:  volumeCmd,
+	})
+	flags := inspectCommand.Flags()
+	flags.BoolVarP(&inspectOpts.All, "all", "a", false, "Inspect all volumes")
+	flags.StringVarP(&inspectFormat, "format", "f", "json", "Format volume output using Go template")
+}
+
+func inspect(cmd *cobra.Command, args []string) error {
+	if (inspectOpts.All && len(args) > 0) || (!inspectOpts.All && len(args) < 1) {
+		return errors.New("provide one or more volume names or use --all")
+	}
+	responses, err := registry.ContainerEngine().VolumeInspect(context.Background(), args, inspectOpts)
+	if err != nil {
+		return err
+	}
+	switch inspectFormat {
+	case "", formats.JSONString:
+		jsonOut, err := json.MarshalIndent(responses, "", "     ")
+		if err != nil {
+			return errors.Wrapf(err, "error marshalling inspect JSON")
+		}
+		fmt.Println(string(jsonOut))
+	default:
+		tmpl, err := template.New("volumeInspect").Parse(inspectFormat)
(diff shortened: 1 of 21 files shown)


cmd/podman/volume_prune.go
```diff
@@ -44,22 +45,29 @@ func init() {
...
+		if r.Err == nil {
+			fmt.Println(r.Id)
+		} else {
+			errs = append(errs, r)
+		}
 	}
-	if len(prunedErrors) == 0 {
-		return nil
+	if len(errs) == 0 {
+		return lastErr
```
mheon: This seems wrong. We never set lastErr?

baude: it was set to nil when declared.

pkg/api/handlers/libpod/volumes.go
```diff
@@ -113,25 +118,30 @@ func ListVolumes(w http.ResponseWriter, r *http.Request) {
...
+			Mountpoint: v.MountPoint(),
+			CreatedAt:  v.CreatedTime(),
+			Labels:     v.Labels(),
+			Scope:      v.Scope(),
+			Options:    v.Options(),
+			UID:        v.UID(),
+			GID:        v.GID(),
 		}
-		volumeConfigs = append(volumeConfigs, config)
+		volumeConfigs = append(volumeConfigs, &entities.VolumeListReport{VolumeConfigResponse: config})
```
mheon: I don't think we want entities to leak into here - struct should live in pkg/api and entities should use it

cmd/podman/volume_prune.go
mheon: But nothing after this sets it to anything *other* than nil?

mheon: Ah, I see it's set below... Hm.

pkg/api/handlers/libpod/volumes.go
baude: we agreed to this on Friday ... imho, this makes more sense as is.  @jwhonce what say you?

rhatdan:
LGTM

cmd/podmanV2/volumes/inspect.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	switch inspectFormat {
+	case "", formats.JSONString:
+		jsonOut, err := json.MarshalIndent(responses, "", "     ")
+		if err != nil {
+			return errors.Wrapf(err, "error marshalling inspect JSON")
+		}
+		fmt.Println(string(jsonOut))
+	default:
+		tmpl, err := template.New("volumeInspect").Parse(inspectFormat)
```
jwhonce: inspectFormat could be "" here.

cmd/podmanV2/volumes/list.go
```diff
@@ -0,0 +1,99 @@
...
+	// "\t" from the command line is not being recognized as a tab
+	// replacing the string "\t" to a tab character if the user passes in "\t"
+	cliOpts.Format = strings.Replace(cliOpts.Format, `\t`, "\t", -1)
+	if cliOpts.Quiet {
+		cliOpts.Format = "{{.Name}}\n"
+	}
+	headers := "DRIVER\tVOLUME NAME\n"
+	row := cliOpts.Format
+	if !strings.HasSuffix(cliOpts.Format, "\n") {
+		row = fmt.Sprintf("%s\n", row)
```
jwhonce:
```suggestion
		row += "\n"
```
Faster and uses less resources, and clearer

cmd/podmanV2/volumes/list.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	format := "{{range $, $x := . }}" + row + "{{end}}"
```
jwhonce:
```suggestion
	format := "{{range . }}" + row + "{{end}}"
```
Otherwise, user needs to add $x.ID to their format as well.

cmd/podmanV2/volumes/list.go
```diff
@@ -0,0 +1,99 @@
...
+	if !cliOpts.Quiet && !cmd.Flag("format").Changed {
+		w = tabwriter.NewWriter(os.Stdout, 12, 2, 2, ' ', 0)
+		format = fmt.Sprintf("%s%s", headers, format)
```
jwhonce:
```suggestion
		format = headers+format
```
Faster, less resources, easier to read

pkg/api/handlers/libpod/volumes.go
jwhonce: I'm with @baude entities, should be used in all the "presentation" layers.

pkg/domain/infra/abi/volumes.go
```diff
@@ -36,3 +38,108 @@ func (ic *ContainerEngine) VolumeCreate(ctx context.Context, opts entities.Volum
...
+	}
+	for _, vol := range vols {
+		reports = append(reports, &entities.VolumeRmReport{
+			Err: ic.Libpod.RemoveVolume(ctx, vol, opts.Force),
+			Id:  vol.Name(),
+		})
+	}
+	return reports, nil
+}
+func (ic *ContainerEngine) VolumeInspect(ctx context.Context, namesOrIds []string, opts entities.VolumeInspectOptions) ([]*entities.VolumeInspectReport, error) {
```
jwhonce:
```suggestion

func (ic *ContainerEngine) VolumeInspect(ctx context.Context, namesOrIds []string, opts entities.VolumeInspectOptions) ([]*entities.VolumeInspectReport, error) {
```

cmd/podmanV2/volumes/inspect.go
baude: that is caught in the case statement prior

jwhonce: Missed it

PR #5611, opened 2020-03-25
podmanv2 pod create using podspecgen

baude:
using the factory approach similar to container, we now create pods based on a pod spec generator.  wired up the podmanv2 pod create command, podcreatewithspec binding, simple binding test, and apiv2 endpoint.

also included some code refactoring as it introduced as easy circular import.

### DIFF
--- a/cmd/podmanV2/common/netflags.go
+++ b/cmd/podmanV2/common/netflags.go
@@ -1,0 +1,108 @@
+package common
+
+import (
+	"net"
+
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/libpod/pkg/rootless"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func getDefaultNetwork() string {
+	if rootless.IsRootless() {
+		return "slirp4netns"
+	}
+	return "bridge"
+}
+
+func GetNetFlags() *pflag.FlagSet {
+	netFlags := pflag.FlagSet{}
+	netFlags.StringSlice(
+		"add-host", []string{},
+		"Add a custom host-to-IP mapping (host:ip) (default [])",
+	)
+	netFlags.StringSlice(
+		"dns", []string{},
+		"Set custom DNS servers",
+	)
+	netFlags.StringSlice(
+		"dns-opt", []string{},
+		"Set custom DNS options",
+	)
+	netFlags.StringSlice(
+		"dns-search", []string{},
+		"Set custom DNS search domains",
+	)
+	netFlags.String(
+		"ip", "",
+		"Specify a static IPv4 address for the container",
+	)
+	netFlags.String(
+		"mac-address", "",
+		"Container MAC address (e.g. 92:d0:c6:0a:29:33)",
+	)
+	netFlags.String(
+		"network", getDefaultNetwork(),
+		"Connect a container to a network",
+	)
+	netFlags.StringSliceP(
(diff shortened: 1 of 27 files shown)


cmd/podmanV2/common/util.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"strconv"
+
+	"github.com/cri-o/ocicni/pkg/ocicni"
+	"github.com/docker/go-connections/nat"
+	"github.com/pkg/errors"
+)
+
+// createPortBindings iterates ports mappings and exposed ports into a format CNI understands
+func createPortBindings(ports []string) ([]ocicni.PortMapping, error) {
```
mheon: This code is in desperate need of a rewrite. Suggest we comment as much.

pkg/domain/entities/pods.go
```diff
@@ -0,0 +1,45 @@
...
+	s.InfraImage = p.InfraImage
+	s.SharedNamespaces = p.Share
+
+	// Networking config
+	s.NetNS = p.Net.Network
+	s.StaticIP = p.Net.StaticIP
+	s.StaticMAC = p.Net.StaticMAC
+	s.PortMappings = p.Net.PublishPorts
+	s.CNINetworks = p.Net.CNINetworks
+	s.DNSServer = p.Net.DNSServers
```
mheon: Does this handler dns=none?

mheon: Or dns=host... I forget what we named it

pkg/domain/infra/abi/pods.go
```diff
@@ -4,6 +4,9 @@ package abi
 
 import (
 	"context"
+
+	"github.com/containers/libpod/pkg/specgen"
+
```
mheon: Why all the whitespace between imports?

mheon:
I'm getting really uncomfortable with the amount of code duplication going on here - we're rapidly approaching the point where bugfixes are going to have to go into both Podman and Podmanv2, and making sure that happens is going to be painful.

rhatdan:
Which is why we need to get rid of podman 1.0 ASAP.  We need to start a branch for 1.0 and then make master branch the 2.0 branch.

baude:
> Which is why we need to get rid of podman 1.0 ASAP. We need to start a branch for 1.0 and then make master branch the 2.0 branch.

@rhatdan you couldnt be more right!

pkg/bindings/test/pods_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/containers/libpod/pkg/specgen"
+
```
jwhonce:
```suggestion
```

cmd/podmanV2/common/netflags.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, err
+	}
+	servers, err := cmd.Flags().GetStringSlice("dns")
+	if err != nil {
+		return nil, err
+	}
+	for _, d := range servers {
+		if d == "host" {
+			opts.DNSHost = true
```
mheon: Damn, this is actually `--dns=none` - my fault here.

From podman-run manpage:
```
       The special value none can be specified to disable creation of /etc/resolv.conf in the container by Podman.  The /etc/resolv.conf file in the image will be used without changes.
```

PR #5612, opened 2020-03-25
enable linting on v2

baude:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,7 @@ run:
     - apparmor
     - seccomp
     - selinux
+    - ABISupport
   concurrency: 6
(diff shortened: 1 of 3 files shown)


mheon:
LGTM

vrothberg:
> That doesn't seem to be enough. Run the linter on some files (e.g., `golangci-lint run pkg/domain/infra/abi/containers.go `). There is a huge amount of exported types and functions without doc comments; none of them should pass the linter.

Actually. We still disable `golint`. Running that reveals all kinds of issues I mentioned in the mail earlier. I assumed golint is part of `make lint` but it isn't. I really think we should enable it. 

/lgtm
/hold

PR #5613, opened 2020-03-25
Ensure that exec sends resize events

### LINKED ISSUE #5584: Since version 1.8.3-0.1.dev.gitaa6c8c2.fc33 all console utilities in container working only in 80x25 chars .
Since version 1.8.3-0.1.dev.gitaa6c8c2.fc33 all console utilities in container working only in 80x25 chars .

Screenshot from 2020-03-23 10-07-27

Last good version is 1.8.2-0.9.dev.gitcccd05c.fc33

mheon:
We previously tried to send resize events only after the exec session successfully started, which makes sense (we might drop an event or two that came in before the exec session started
otherwise). However, the start function blocks, so waiting actually means we send no resize events at all, which it obviously worse than losing a few.. Sending resizes before attach starts seems to work fine in my testing, so let's do that until we get bug reports that it doesn't work.

Fixes #5584

### DIFF
--- a/libpod/container_exec.go
+++ b/libpod/container_exec.go
@@ -575,7 +575,7 @@ func (c *Container) ExecResize(sessionID string, newSize remotecommand.TerminalS
 		return errors.Wrapf(define.ErrNoSuchExecSession, "container %s has no exec session with ID %s", c.ID(), sessionID)
 	}
 
-	logrus.Infof("Removing container %s exec session %s", c.ID(), session.ID())
+	logrus.Infof("Resizing container %s exec session %s to %+v", c.ID(), session.ID(), newSize)
 
 	if session.State != define.ExecStateRunning {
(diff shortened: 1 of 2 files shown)


haircommander:
LGTM,

for the record this should be fine because conmon waits to be ready to consume events on the ctrl buffer [ref](https://github.com/containers/conmon/blob/89b2478b507c6f285cd97ae8e55c85b9cafe6e81/src/ctrl.c#L62). The only thing that could happen is they start buffering if there are too many piling up in the fd before conmon consumes them, but that's wacky so I'm not worried

mheon:
We're actually sending them before Conmon is running, so there will be no `ctl` yet, and the first few will fail as such.

haircommander:
hmhmhm I see. in that case, I'd be wary of getting https://github.com/containers/libpod/issues/3946 again

mheon:
Easy, hacky way to work around this: Busywait until exec session status is Running, then start processing resize events.

Probably more proper way: add a channel parameter to ExecStartAttach that will be signalled when the exec session has been successfully started. A little hesitant to do this because I don't know what else wants this...

PR #5614, opened 2020-03-26
podmanv2 add pre-run to each commmand

baude:
each container command needs a prerune because it is not part of the containers subcommand yet


rhatdan:
LGTM

PR #5620, opened 2020-03-26
Add support for specifying CNI networks in podman play kube

cfelder:
As far as I understood it is not possible to define container network on the yaml level (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#pod-v1-core).

I am adding networking options to ``podman play cube`` though.

### DIFF
--- a/docs/source/markdown/podman-play-kube.1.md
+++ b/docs/source/markdown/podman-play-kube.1.md
@@ -36,6 +36,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--network**=*cni networks*
+
+A comma-separated list of the names of CNI networks the pod should join.
+
 **--quiet**, **-q**
 
 Suppress output information when pulling images
@@ -62,8 +66,16 @@ $ podman play kube demo.yml
 52182811df2b1e73f36476003a66ec872101ea59034ac0d4d3a7b40903b955a6
 ```
 
+CNI network(s) can be specified as comma-separated list using ``--network``
+```
+$ podman play kube demo.yml --network cni1,cni2
+52182811df2b1e73f36476003a66ec872101ea59034ac0d4d3a7b40903b955a6
+```
+
+Please take into account that CNI networks must be created first using podman-network-create(1).
+
 ## SEE ALSO
-podman(1), podman-container(1), podman-pod(1), podman-generate-kube(1), podman-play(1)
+podman(1), podman-container(1), podman-pod(1), podman-generate-kube(1), podman-play(1), podman-network-create(1)
 
 ## HISTORY
 December 2018, Originally compiled by Brent Baude (bbaude at redhat dot com)
--- a/pkg/adapter/pods.go
+++ b/pkg/adapter/pods.go
@@ -343,7 +343,7 @@ func (r *LocalRuntime) CreatePod(ctx context.Context, cli *cliconfig.PodCreateVa
 			logrus.Debugf("Pod will use host networking")
 			options = append(options, libpod.WithPodHostNetwork())
 		case "":
-			return "", errors.Errorf("invalid value passed to --net: must provide a comma-separated list of CNI networks or host")
+			return "", errors.Errorf("invalid value passed to --network: must provide a comma-separated list of CNI networks or host")
 		default:
 			// We'll assume this is a comma-separated list of CNI
 			// networks.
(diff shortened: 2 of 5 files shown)


mheon:
I don't know if we want to do this - should we be sourcing from the Kube YAML instead? @haircommander thoughts?

cfelder:
@mheon I would prefer specifying this on yaml level but I have not found an option to declare a network for a pod in the specs.

haircommander:
yeah I wouldn't go this route. There are some first class pod API features that relate to pod networking:

the[dnsConfig](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config) object would take care of `--dns` `--dns-opt` and `--dns-search`
[hostAliases](https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/#adding-additional-entries-with-hostaliases) seem to take care of `--add-hosts` and `--ip`

finally, you can replicate `--publish` on the container level with the [ports](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#containerport-v1-core) object

there doesn't seem to be a way to specify `--mac-address`, and `--network` also seems inaccessible. It is also hard to figure out how similar behavior to `--no-hosts` could be implemented.

haircommander:
I suppose `--network` could be added as a hacky way to specify a network (Because k8s api doesn't provide it). But the others should be done the k8s way IMO

cfelder:
I am more than happy to implement this in yaml if you can help me with the specification as I don't know how to specify the network options in yaml syntax.

it would be a lot easier to implement this in ``PlayKubeYAML`` in ``pkg/adapter/pods.go`` using ``libPod.WithPodNetworks``.

haircommander:
> I am more than happy to implement this in yaml if you can help me with the specification as I don't know how to specify the network options in yaml syntax.
> 

were the above links not enough to get started? do you have specific questions about them?

> it would be a lot easier to implement this in `PlayKubeYAML` in `pkg/adapter/pods.go` using `libPod.WithPodNetworks`.

As an example, when iterating through the container objects, you can access `container.Ports` and take the information defined [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#containerport-v1-core) and translate it into the equivalent values for libpod/options:WithNetNS()

does that help?

haircommander:
> > I am more than happy to implement this in yaml if you can help me with the specification as I don't know how to specify the network options in yaml syntax.
> 
> were the above links not enough to get started? do you have specific questions about them?
> 
> > it would be a lot easier to implement this in `PlayKubeYAML` in `pkg/adapter/pods.go` using `libPod.WithPodNetworks`.
> 
> As an example, when iterating through the container objects, you can access `container.Ports` and take the information defined [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#containerport-v1-core) and translate it into the equivalent values for libpod/options:WithNetNS()
> 
> does that help?

nvm on that, we already do something similar in `getPodPorts` in pkg/adapter/pods.go . but hopefully that serves as an example for the other options

cfelder:
Thanks for your remarks I changed my PR to just add ``--network`` for ``podman play kube`` in order to support specifying CNI networks which cannot be done in yaml syntax right now.

mheon:
Concur that just --network is appropriate...

We probably also want to restrict it to only listing a set of CNI networks - we should not support `--net=host` and `--net=bridge` as those are handled in the YAML.

pkg/adapter/pods.go
```diff
@@ -343,7 +343,7 @@ func (r *LocalRuntime) CreatePod(ctx context.Context, cli *cliconfig.PodCreateVa
 			logrus.Debugf("Pod will use host networking")
 			options = append(options, libpod.WithPodHostNetwork())
 		case "":
-			return "", errors.Errorf("invalid value passed to --net: must provide a comma-separated list of CNI networks or host")
+			return "", errors.Errorf("invalid value passed to --network: must provide a comma-separated list of CNI networks or host")
```
TomSweeneyRedHat: host is not mentioned  in the man page.  It should be removed here or added there.

docs/source/markdown/podman-play-kube.1.md
```diff
@@ -36,6 +36,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--network**=*cni networks*
+
+A comma-separated list of the names of CNI networks the pod should join.
+
```
TomSweeneyRedHat: an example of valid input might be nice in the example section.

TomSweeneyRedHat:
Any tests possible?  Code LGTM, just a few nitty doc things.  Thanks @cfelder !

pkg/adapter/pods.go
cfelder: This belongs to ``CreatePod`` which includes host in its man page but I've corrected this in line 601.

I just wanted to correct ``--net`` to ``--network`` here.

docs/source/markdown/podman-play-kube.1.md
cfelder: Done

vrothberg:
Ping. Let's get this in.

mheon:
/approve
LGTM

PR #5627, opened 2020-03-26
apiv2 add default network in specgen

baude:
when a network is not provided, we should set a default mode based on rootless or rootfull.

Fixes: #5366


marusak:
Thanks @baude . I checked it locally and it fixes the issue for me.

mheon:
LGTM

PR #5632, opened 2020-03-27
V2 podman image prune

jwhonce:
* Fixed header for `podman image ls`
* Implemented prune `all` flag, preserved filter method for backwards
  capability
* Updated binding tests


baude:
LGTM

PR #5634, opened 2020-03-27
V2 podman system service

jwhonce:
* Added support for system service
* Enabled linting on the varlinkapi source, needed to support V2
  service command
* Add hack/golangci-lint.sh to support linting with multiple sets of build tags

### DIFF
--- a/hack/golangci-lint.sh
+++ b/hack/golangci-lint.sh
@@ -1,0 +1,15 @@
+#!/bin/bash -e
+
+# Need to run linter twice to cover all the build tags code paths
+
+declare -A BUILD_TAGS
+BUILD_TAGS[default]="apparmor,seccomp,selinux"
+BUILD_TAGS[abi]="${BUILD_TAGS[default]},ABISupport,varlink,!remoteclient"
+BUILD_TAGS[tunnel]="${BUILD_TAGS[default]},!ABISupport,!varlink,remoteclient"
+
+[[ $1 == run ]] && shift
+
+for i in tunnel abi; do
+  echo Build Tags: ${BUILD_TAGS[$i]}
+  golangci-lint run --build-tags=${BUILD_TAGS[$i]} "$@"
+done
--- a/pkg/domain/entities/system.go
+++ b/pkg/domain/entities/system.go
@@ -1,0 +1,14 @@
+package entities
+
+import (
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+// ServiceOptions provides the input for starting an API Service
+type ServiceOptions struct {
+	URI     string         // Path to unix domain socket service should listen on
+	Timeout time.Duration  // duration of inactivity the service should wait before shutting down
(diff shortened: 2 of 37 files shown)


jwhonce:
@baude We need to discuss how to generate the varlink code. I believe that is what is giving the linter heartburn

mheon:
Are we doing Varlink in V2? Can we get away with keeping varlink support on a 1.0 branch?

baude:
> Are we doing Varlink in V2? Can we get away with keeping varlink support on a 1.0 branch?

unfortunately, we are going to have to for some amount of time. i would love to talk about how to fence it off with a build tag so we could begin deprecating in fedora.

rhatdan:
Var link will need to be supported for the length of RHEL8, I think or at least until RHEL8.6

rhatdan:
We can begin deprecating it in Fedora 33.

hack/golangci-lint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+
+# Need to run linter twice to cover all the build tags code paths
+
+declare -A BUILD_TAGS
+BUILD_TAGS[default]="apparmor,seccomp,selinux"
+BUILD_TAGS[abi]="${BUILD_TAGS[default]},ABISupport,varlink,!remoteclient"
+BUILD_TAGS[tunnel]="${BUILD_TAGS[default]},!ABISupport,!varlink,remoteclient"
```
vrothberg, line 8: Elegant!

pkg/domain/entities/system.go
```diff
@@ -0,0 +1,13 @@
+package entities
+
+import (
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+type ServiceOptions struct {
+	Uri     string
```
vrothberg, line 10: Nit: s/Uri/URI/

pkg/domain/entities/system.go
```diff
@@ -0,0 +1,13 @@
+package entities
+
+import (
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+type ServiceOptions struct {
```
vrothberg, line 9: Can we add comments to this type? Once we enable `golint` it would complain about it.

hack/golangci-lint.sh
jwhonce: Thanks

pkg/domain/entities/system.go
jwhonce: done.

rhatdan:
@jwhonce Needs rebase.

hack/golangci-lint.sh
```diff
@@ -0,0 +1,15 @@
...
+
+[[ $1 == run ]] && shift
+
+for i in tunnel abi; do
+  echo Build Tags: ${BUILD_TAGS[$i]}
```
cevich, line 13: Nit: It's good practice to always quote string arguments to `echo`.

PR #5638, opened 2020-03-27
podmanv2 checkpoint and restore

baude:
add the ability to checkpoint and restore containers on v2podman

### DIFF
--- a/pkg/api/handlers/libpod/containers.go
+++ b/pkg/api/handlers/libpod/containers.go
@@ -1,16 +1,21 @@
 package libpod
 
 import (
+	"io/ioutil"
 	"net/http"
+	"os"
 	"path/filepath"
 	"sort"
 	"strconv"
 	"time"
 
+	"github.com/containers/libpod/pkg/api/handlers/compat"
+
 	"github.com/containers/libpod/cmd/podman/shared"
 	"github.com/containers/libpod/libpod"
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
+	"github.com/containers/libpod/pkg/domain/entities"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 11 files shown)


pkg/api/handlers/libpod/containers.go
```diff
@@ -325,3 +326,84 @@ func ListContainerBatch(rt *libpod.Runtime, ctr *libpod.Container, opts shared.P
...
 	return ps, nil
 }
+
+func Checkpoint(w http.ResponseWriter, r *http.Request) {
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	query := struct {
+		Keep           bool   `schema:"keep"`
+		LeaveRunning   bool   `schema:"leaveRunning"`
+		TCPEstablished bool   `schema:"tcpEstablished"`
+		Export         string `schema:"export"`
```
mheon: We can't just pass this over - we'll be creating the export tar on the server. Expected behavior would be the client, I imagine. If export is requested we probably need to save the tar to a temporary file and stream it back to the client.

pkg/api/handlers/libpod/containers.go
```diff
@@ -325,3 +326,84 @@ func ListContainerBatch(rt *libpod.Runtime, ctr *libpod.Container, opts shared.P
...
+	}
+	utils.WriteResponse(w, http.StatusOK, entities.CheckpointReport{Id: ctr.ID()})
+}
+
+func Restore(w http.ResponseWriter, r *http.Request) {
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	query := struct {
+		Keep            bool   `schema:"keep"`
+		TCPEstablished  bool   `schema:"tcpEstablished"`
+		Import          string `schema:"import"`
```
mheon: Same as Export, but in reverse - the tar is going to be on the client, so we need to stream it to the server, save to a temporary file, and then pass that into libpod Restore.

mheon:
Suggestion: strip import/export and do them in a separate PR, rest is probably ready to merge as is.

baude:
ready for merge

pkg/api/handlers/libpod/containers.go
```diff
@@ -325,3 +330,129 @@ func ListContainerBatch(rt *libpod.Runtime, ctr *libpod.Container, opts shared.P
 	}
 	return ps, nil
 }
+
+func Checkpoint(w http.ResponseWriter, r *http.Request) {
```
mheon, line 334: Needs doc comment

mheon: Can wait till later, though, given this is just an HTTP endpoint

mheon:
LGTM

pkg/api/handlers/libpod/containers.go
baude: none of the endpoints have doc comments

rhatdan:
I want @adrianreber to make sure this works but It LGTM now. Just so we can get to testing.
/lgtm

PR #5639, opened 2020-03-27
V2 pod top

vrothberg:
The v2 endpoint for pod-top was missing, so we're first adding this endpoint and then do the plumbing for podmanV2.

While CI is cooking, I'll look into ending some tests for the endpoint but I was born after the war and don't speak perl.

### DIFF
--- a/cmd/podmanV2/pods/top.go
+++ b/cmd/podmanV2/pods/top.go
@@ -1,0 +1,90 @@
+package pods
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/psgo"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	topDescription = fmt.Sprintf(`Specify format descriptors to alter the output.
+
+	You may run "podman pod top -l pid pcpu seccomp" to print the process ID, the CPU percentage and the seccomp mode of each process of the latest pod.
+  Format Descriptors:
+    %s`, strings.Join(psgo.ListDescriptors(), ","))
+
+	topOptions = entities.PodTopOptions{}
+
+	topCommand = &cobra.Command{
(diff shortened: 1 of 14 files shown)


mheon:
LGTM

cmd/podmanV2/pods/top.go
```diff
@@ -0,0 +1,90 @@
...
+
+	topCommand = &cobra.Command{
+		Use:               "top [flags] POD [FORMAT-DESCRIPTORS|ARGS]",
+		Short:             "Display the running processes in a pod",
+		Long:              topDescription,
+		PersistentPreRunE: preRunE,
+		RunE:              top,
+		Args:              cobra.ArbitraryArgs,
+		Example: `podman top podID
+podman top --latest
```
baude: shoiuld all these be pod top?

pkg/api/server/register_pods.go
```diff
@@ -263,5 +263,36 @@ func (s *APIServer) registerPodsHandlers(r *mux.Router) error {
...
+	//    default: true
+	//    description: Stream the output
+	//  - in: query
+	//    name: ps_args
+	//    type: string
+	//    default: -ef
+	//    description: arguments to pass to ps such as aux. Requires ps(1) to be installed in the container if no ps(1) compatible AIX descriptors are used.
+	// responses:
+	//   200:
+	//     $ref: "#/responses/DockerTopResponse"
```
baude: any chance we can NOT have docker in the response description?

cmd/podmanV2/pods/top.go
vrothberg: They should. Good catch!

baude:
couple of non-code things ... else lgtm

pkg/api/server/register_pods.go
vrothberg: Sure, will change. Inherited from the containers top one, will have a look there as well.

vrothberg: Done. PTanotherLook.

rhatdan:
Needs a rebase

TomSweeneyRedHat:
LGTM once rebased.

vrothberg:
Restarted flaked tests.

vrothberg:
Ready to merge

PR #5645, opened 2020-03-27
V2 podman inspect

jwhonce:
* Demonstrates having a sub-command (inspect) bridge multiple packages 
* Expose podman container inspect
* Expose podman image inspect
* Duplicated ImageData to prevent import cycles.  Will be merged with other
  like structs after V2 migration is complete


cmd/podmanV2/images/inspect.go
```diff
@@ -1,71 +1,44 @@
...
-	flags := inspectCmd.Flags()
-	flags.BoolVarP(&inspectOpts.Latest, "latest", "l", false, "Act on the latest container podman is aware of")
-	flags.BoolVarP(&inspectOpts.Size, "size", "s", false, "Display total file size")
-	flags.StringVarP(&inspectOpts.Format, "format", "f", "", "Change the output format to a Go template")
-
-	if registry.EngineOptions.EngineMode == entities.ABIMode {
-		// TODO: This is the same as V1.  We could skip creating the flag altogether in V2...
-		_ = flags.MarkHidden("latest")
-	}
+	inspectOpts = common.AddInspectFlagSet(inspectCmd)
```
mheon: The latest flag has no meaning for `podman image inspect` so we shouldn't be adding it here

pkg/domain/entities/images.go
```diff
@@ -47,6 +49,32 @@ func (i *Image) Id() string {
 	return i.ID
 }
 
+type ImageData struct {
```
mheon: Duplicating this seems like a bad idea; we should move the backend structs into libpod/define and make this an alias for that type, as we did for container inspect.

PR #5647, opened 2020-03-28
Update podmanV2 to use containers.conf

rhatdan:
Move some default processing code around so that it can be shard between
podmanv1 and V2.

Switch to using --time as opposed to --timeout to better match Docker.


rhatdan:
@mheon @vrothberg @baude @jwhonce @TomSweeneyRedHat @QiWang19 PTAL

pkg/util/utils.go
```diff
@@ -652,3 +660,120 @@ func ValidateSysctls(strSlice []string) (map[string]string, error) {
 	}
 	return sysctl, nil
 }
+
+func DefaultContainerConfig() *config.Config {
+	return defaultContainerConfig
+}
+
+func containerConfig() *config.Config {
```
vrothberg: I think this should be an `init()` to make it explicit that this is happening whenever the package is used.

pkg/util/utils.go
```diff
@@ -652,3 +660,120 @@ func ValidateSysctls(strSlice []string) (map[string]string, error) {
 	}
 	return sysctl, nil
 }
+
+func DefaultContainerConfig() *config.Config {
```
vrothberg: Please document/comment exposed types and APIs. Currently it's passing CI because we (very unfortunately) have `golint` disabled but we need to get that back. Documenting it now will spare us some work in the future.

rhatdan:
We don't want to tie V2 to golang library.  Some of these defaults are used in client.

pkg/util/utils.go
vrothberg: I haven't heard of that yet. Do you have link? It's mentioned in [effective go](https://golang.org/doc/effective_go.html#initialization).

I prefer having such code in an `init()` mostly because of the `os.Exit(...)`. I find it more idiomatic for such cases. I wouldn't care as much without the exit.

rhatdan: I thought one of the linters was preventing it, I will try again.

pkg/util/utils.go
rhatdan: Moving these to containers.conf

rhatdan:
@baude @giuseppe @vrothberg @mheon @jwhonce @QiWang19 @TomSweeneyRedHat PTAL

mheon:
LGTM but want a signoff by @jwhonce

contrib/podmanimage/stable/Dockerfile
```diff
@@ -11,16 +11,12 @@ FROM fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN useradd podman; yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
```
TomSweeneyRedHat, line 14: I don't think "build" should change here to podman....

TomSweeneyRedHat: ditto the others.

contrib/podmanimage/stable/containers.conf
```diff
@@ -0,0 +1,11 @@
+[containers]
+netns="host"
+userns="host"
+ipcns="host"
+utsns="host"
+cgroupns="host"
+cgroups="disabled"
+[engine]
+cgroup_manager = "cgroupfs"
```
TomSweeneyRedHat, line 9: This line was under `[containers]` in Buildah.  Should it be in there or here?

rhatdan: This is an engine field.  It should be here.  Buildah does not use, I believe.

rhatdan: Actually we do need it to prevent the warning.

rhatdan:
@jwhonce PTAL

contrib/podmanimage/stable/Dockerfile
rhatdan: Why not?

TomSweeneyRedHat: I assumed "build" was expected to be in the passwd file and this might mess something up that was dependent on it.  I'm not sure where that dependency is though, or if we even need to do any useradd here.  I don't recall the genesis of it.

TomSweeneyRedHat: and, quick little innocuous changes like this have the tendency to come back and bite me later.

rhatdan: build is just the user that you use.  What I am trying to do is build a container image that can be used both by root and as rootless.

podman run --user podman quay.io/stable/podman podman info
and
podman run  quay.io/stable/podman podman info

Would work.

Having the user be "build" makes no sense, (It does for Buildah though).  This was a cut and paste from the buildahimage, and I am just fixing it.

cmd/podman/containers/container.go
```diff
@@ -20,7 +17,7 @@ var (
 		RunE:             registry.SubCommandExists,
 	}
 
-	defaultContainerConfig = getDefaultContainerConfig()
+	defaultContainerConfig = util.DefaultContainerConfig()
```
jwhonce, line 20:
```suggestion
	containerConfig = util.DefaultContainerConfig()
```
In other packages, we have containerConfig as a PodmanConfig which carries a config.Config().  I suspect we dont' want `util.DefaultContainerConfig() `here rather `registry.PodmanConfig()`.

pkg/util/utils.go
```diff
@@ -669,3 +681,7 @@ func swapSELinuxLabel(cLabel, processLabel string) (string, error) {
 	dcon["type"] = scon["type"]
 	return dcon.Get(), nil
 }
+
+func DefaultContainerConfig() *config.Config {
```
jwhonce, line 685: Should cmd/podman/registry/config.go be updated to use this rather than calling config directly?

jwhonce:
LGTM

PR #5657, opened 2020-03-30
Fix the pod name passed to cni

AlbanBedel:
When creating a network we pass down a name which end up in the K8S_POD_NAME argument to cni plugins. Currently this name is always filled with the container name, so for pods it is the name of the infra container, not really what one would expect.

This mess up with the dnsname plugin as it doesn't receive the pod name in K8S_POD_NAME, see https://github.com/containers/dnsname/issues/11. To fix this pass the pod name when the container is part of a pod, otherwise use the container name like before.

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -101,7 +101,19 @@ func (r *Runtime) configureNetNS(ctr *Container, ctrNS ns.NetNS) ([]*cnitypes.Re
 		requestedMAC = ctr.config.StaticMAC
 	}
 
-	podNetwork := r.getPodNetwork(ctr.ID(), ctr.Name(), ctrNS.Path(), ctr.config.Networks, ctr.config.PortMappings, requestedIP, requestedMAC)
+	// If we are in a pod use the pod name for the network, otherwise the container name
+	var podName string
+	if ctr.PodID() != "" {
+		pod, err := r.GetPod(ctr.PodID())
+		if err == nil {
+			podName = pod.Name()
+		}
+	}
+	if podName == "" {
+		podName = ctr.Name()
+	}
+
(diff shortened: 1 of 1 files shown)


libpod/networking_linux.go
```diff
@@ -101,7 +101,15 @@ func (r *Runtime) configureNetNS(ctr *Container, ctrNS ns.NetNS) ([]*cnitypes.Re
 		requestedMAC = ctr.config.StaticMAC
 	}
 
-	podNetwork := r.getPodNetwork(ctr.ID(), ctr.Name(), ctrNS.Path(), ctr.config.Networks, ctr.config.PortMappings, requestedIP, requestedMAC)
+	var podName string
+	pod, err := r.GetPod(ctr.PodID())
```
mheon: We should only make this call is `ctr.PodID() != ""` - this is a fairly expensive database lookup that we should avoid if we can.

AlbanBedel: Looking at the boltdb code: https://github.com/containers/libpod/blob/598bb53d46dfc85b8bcc1e3000736106f80de93e/libpod/boltdb_state.go#L1366-L1369 it just return right away if the id is empty, so that shouldn't make a big difference.

mheon: I don't want to force this decision on states, though - we aren't strictly wedded to Bolt, so a future SQLite-based state, for example, might not do that. Just a bit safer to do it here, IMO.

TomSweeneyRedHat:
LGTM once the requested check is added.

AlbanBedel:
Is there anything still preventing this PR from being merged?

TomSweeneyRedHat:
I think it's GTG, @mheon what sayest thou?

PR #5659, opened 2020-03-30
podman-save: log when more than one image is specified

vrothberg:
Podman save can only save one image.  While we are planning to support
multiple images to catch up with Docker, we still need time to get the
work done.

Instead of ignoring other arguments, log when there are more than one.

### DIFF
--- a/cmd/podman/save.go
+++ b/cmd/podman/save.go
@@ -9,6 +9,7 @@ import (
 	"github.com/containers/libpod/pkg/adapter"
 	"github.com/containers/libpod/pkg/util"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
 )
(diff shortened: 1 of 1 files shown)


cmd/podman/save.go
```diff
@@ -70,6 +71,10 @@ func saveCmd(c *cliconfig.SaveValues) error {
 		return errors.Errorf("need at least 1 argument")
 	}
 
+	if len(args) > 1 {
+		logrus.Errorf("only the first image will be saved but %d are specified", len(args))
+	}
```
mtrmac: `LocalRuntime.SaveImage` uses `c.inputArgs[1:]` as `DockerArchiveAdditionalTags`. That might be not what the command is _expected_ to do — but this breaks that.

So, either that functionality is useful, and this error needs to be removed (or reworded, or… I don’t know), or it is not, and it should be removed from the code. I guess the former.

vrothberg:
Closing for now.

PR #5664, opened 2020-03-30
rootless: make cgroup ownership detection not fatal

giuseppe:
do not fail if we cannot detect the cgroup ownership.  The detection
fails when running in a container, since the cgroup showed in
/proc/self/cgroup is not accessible, due to the runtime mounting it
directly as the cgroup root.

Closes: https://github.com/containers/libpod/issues/5488

### DIFF
--- a/cmd/podman/main_local.go
+++ b/cmd/podman/main_local.go
@@ -174,13 +174,13 @@ func setupRootless(cmd *cobra.Command, args []string) error {
 	if os.Geteuid() == 0 {
 		ownsCgroup, err := cgroups.UserOwnsCurrentSystemdCgroup()
 		if err != nil {
-			return err
-		}
-		conf, err := runtime.GetConfig()
-		if err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


cmd/podman/main_local.go
```diff
@@ -174,13 +174,13 @@ func setupRootless(cmd *cobra.Command, args []string) error {
 	if os.Geteuid() == 0 {
 		ownsCgroup, err := cgroups.UserOwnsCurrentSystemdCgroup()
 		if err != nil {
-			return err
-		}
-		conf, err := runtime.GetConfig()
-		if err != nil {
-			return err
+			logrus.Warnf("Failed to detect the owner for the current cgroup: %v", err)
```
rhatdan: Currently users will no see Warnf messages by default.  But this should only happen when running podman in container correct?

mheon:
LGTM once tests go green

cmd/podman/main_local.go
giuseppe: yes, that is the only case where it should happen

PR #5672, opened 2020-03-30
podmanv2 save image

baude:
add ability to save an image for podman v2

### DIFF
--- a/cmd/podmanV2/images/save.go
+++ b/cmd/podmanV2/images/save.go
@@ -1,0 +1,87 @@
+package images
+
+import (
+	"context"
+	"os"
+	"strings"
+
+	"github.com/containers/libpod/libpod/define"
+
+	"github.com/containers/libpod/cmd/podmanV2/parse"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/libpod/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+var validFormats = []string{define.OCIManifestDir, define.OCIArchive, define.V2s2ManifestDir, define.V2s2Archive}
+
+var (
+	saveDescription = `Save an image to docker-archive or oci-archive on the local machine. Default is docker-archive.`
+
+	saveCommand = &cobra.Command{
+		Use:               "save [flags] IMAGE",
+		Short:             "Save image to an archive",
+		Long:              saveDescription,
+		PersistentPreRunE: preRunE,
+		RunE:              save,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.Errorf("need at least 1 argument")
+			}
+			format, err := cmd.Flags().GetString("format")
+			if err != nil {
+				return err
+			}
+			if !util.StringInSlice(format, validFormats) {
+				return errors.Errorf("format value must be one of %s", strings.Join(validFormats, " "))
+			}
+			return nil
+		},
+		Example: `podman save --quiet -o myimage.tar imageID
+  podman save --format docker-dir -o ubuntu-dir ubuntu
+  podman save > alpine-all.tar alpine:latest`,
+	}
+)
+
+var (
+	saveOpts entities.ImageSaveOptions
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: saveCommand,
+	})
+	flags := saveCommand.Flags()
+	flags.BoolVar(&saveOpts.Compress, "compress", false, "Compress tarball image layers when saving to a directory using the 'dir' transport. (default is same compression type as source)")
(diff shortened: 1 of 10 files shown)


rhatdan:
LGTM

cmd/podmanV2/images/save.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/libpod/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+const (
+	ociManifestDir  = "oci-dir"
```
mheon: We can't put these somewhere common? I'm seeing a lot of magic strings from here floating around in the code

pkg/api/handlers/libpod/images.go
```diff
@@ -193,17 +210,29 @@ func ExportImage(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	rdr, err := os.Open(output)
 	if err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to read the exported tarfile"))
 		return
 	}
 	defer rdr.Close()
-	defer os.Remove(tmpfile.Name())
+	//defer os.Remove(output)
```
mheon: Is this supposed to be commented?

mheon: This also seems like it needs to be moved up above the block at line 217, and always be a RemoveAll

pkg/api/handlers/libpod/images.go
```diff
@@ -193,17 +210,29 @@ func ExportImage(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			utils.InternalServerError(w, err)
+			return
+		}
+		defer os.RemoveAll(output)
+		defer rdr.Close()
+		utils.WriteResponse(w, http.StatusOK, rdr)
+		return
+	}
+	rdr, err := os.Open(output)
```
mheon: Doesn't this immediately overwrite `rdr` as set right above?

pkg/bindings/images/images.go
```diff
@@ -145,11 +145,12 @@ func Export(ctx context.Context, nameOrID string, w io.Writer, format *string, c
...
 	}
-	if err := response.Process(nil); err != nil {
+
+	if response.StatusCode/100 == 2 || response.StatusCode/100 == 3 {
+		_, err = io.Copy(w, response.Body)
 		return err
 	}
-	_, err = io.Copy(w, response.Body)
-	return err
+	return response.Process(nil)
```
mheon: Aren't we double-reading a potentially very large tar like this?

pkg/api/handlers/libpod/images.go
baude: there is a return above this.

cmd/podmanV2/images/save.go
jwhonce: @baude `registry` would be the package, if this is only for the CLI.  Otherwise, `entities` if it is needed more further afield.

pkg/api/handlers/libpod/images.go
```diff
@@ -162,28 +163,44 @@ func PruneImages(w http.ResponseWriter, r *http.Request) {
...
-
-	tmpfile, err := ioutil.TempFile("", "api.tar")
-	if err != nil {
-		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to create tempfile"))
-		return
-	}
-	if err := tmpfile.Close(); err != nil {
-		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to close tempfile"))
+	switch query.Format {
+	case "oci-archive", "docker-archive":
```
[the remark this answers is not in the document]
jwhonce: or `entities` ?

pkg/domain/infra/tunnel/images.go
```diff
@@ -85,3 +89,54 @@ func (ir *ImageEngine) Prune(ctx context.Context, opts entities.ImagePruneOption
...
+			return errors.Errorf("%q already exists as a regular file", options.Output)
+		}
+	case os.IsNotExist(err):
+		if err := os.Mkdir(options.Output, 0755); err != nil {
+			return err
+		}
+	default:
+		return err
+	}
+	return utils2.UntarToFileSystem(options.Output, f, nil)
```
jwhonce, line 141: Are you untarring on the server vs the client?

baude:
ready for merge

pkg/api/handlers/utils/handler.go
```diff
@@ -43,6 +43,13 @@ func WriteResponse(w http.ResponseWriter, code int, value interface{}) {
 		w.Header().Set("Content-Type", "application/octet; charset=us-ascii")
 		w.WriteHeader(code)
 
+		if _, err := io.Copy(w, v); err != nil {
+			logrus.Errorf("unable to copy to response: %q", err)
+		}
+	case io.Reader:
+		w.Header().Set("Content-Type", "application/octet; charset=us-ascii")
```
jwhonce, line 50:
```suggestion
		w.Header().Set("Content-Type", "application/octet; charset=utf8")
```
or,
```go
w.Header().Set("Content-Type", "application/x-tar")
```

PR #5673, opened 2020-03-31
Fix ro bind mount flags

kolyshkin:
Found the bug by reading the source for `InitFSMounts()`, confirmed with this:

```console
$ ./bin/podman run -v /tmp:/tmp alpine true; echo $?
0

$ ./bin/podman run -v /tmp:/tmp:ro alpine true; echo $?
0

$ ./bin/podman run -v /tmp:/w0w:ro alpine true; echo $?
> Error: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/tmp\\\" to rootfs \\\"/home/kir/.local/share/containers/storage/overlay/7636ef3650fc91ee4996ccc026532bb3cff7182c0430db662fffb933e0bcadc9/merged\\\" at \\\"/home/kir/.local/share/containers/storage/overlay/7636ef3650fc91ee4996ccc026532bb3cff7182c0430db662fffb933e0bcadc9/merged/w0w\\\" caused \\\"operation not permitted\\\"\"": OCI runtime permission denied error
> 126
```

The last command is not working because in-container mount point
(i.e. mount destination, not the mount source) is used to search
for a parent mount in /proc/self/mountinfo.

And yet the following
```console
$ ./bin/podman run -v /tmp:/run/test:ro alpine true; echo $?
0
```
still works fine! Here's why:

```console
$ mount | grep -E '/run |/tmp '
tmpfs on /run type tmpfs (rw,nosuid,nodev,seclabel,mode=755)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,seclabel)
```

Fixes: 0f5ae3c

@giuseppe PTAL (also if you can suggest what to do to add a regression test for this --  as my attempt has failed).

### DIFF
--- a/pkg/spec/storage.go
+++ b/pkg/spec/storage.go
@@ -867,9 +867,9 @@ func InitFSMounts(inputMounts []spec.Mount) ([]spec.Mount, error) {
 	var mounts []spec.Mount
 	for _, m := range inputMounts {
 		if m.Type == TypeBind {
-			baseMnt, err := findMount(m.Destination, systemMounts)
+			baseMnt, err := findMount(m.Source, systemMounts)
 			if err != nil {
-				return nil, errors.Wrapf(err, "error looking up mountpoint for mount %s", m.Destination)
+				return nil, errors.Wrapf(err, "error looking up mountpoint for mount %s", m.Source)
 			}
 			var noexec, nosuid, nodev bool
 			for _, baseOpt := range strings.Split(baseMnt.Opts, ",") {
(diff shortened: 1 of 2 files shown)


kolyshkin:
OK, tests are passing with the first two commits. Adding the fixed test case.

kolyshkin:
OK unfortunately the updated tests are not revealing the bug, but I am seeing it on my f31 laptop.

Anyway, pushing the fix.

kolyshkin:
I have an alternative (better!) version; https://github.com/containers/libpod/pull/5676

PR #5675, opened 2020-03-31
podmanV2: implement pull

vrothberg:
Implement pulling images for the v2 client.  What I _really_ don't like
is the fact that we are now having a near identical code clone among
`pkg/domain/infra/abi` and `pkg/api/handlers/libpod`.  Partly because we
don't yet have a higher-level pull function and partly because we have
redudancy among `pkg/domain` and `pkg/api`.  Pull might be a high
outlier but I am concerned already by the potential of introducing more
redundancy.  I'd love `infra/abi` and `pkg/abi` to really use the
same code in the future.

### DIFF
--- a/cmd/podmanV2/images/pull.go
+++ b/cmd/podmanV2/images/pull.go
@@ -1,0 +1,140 @@
+package images
+
+import (
+	"fmt"
+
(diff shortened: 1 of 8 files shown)


vrothberg:

rhatdan:
LGTM

PR #5676, opened 2020-03-31
Fix/improve pkg/storage.InitFSMounts

kolyshkin:
### Fix

Found the bug by reading the source for `InitFSMounts()`, confirmed with:

```console
$ ./bin/podman run -v /tmp:/tmp alpine true; echo $?
0

$ ./bin/podman run -v /tmp:/tmp:ro alpine true; echo $?
0

$ ./bin/podman run -v /tmp:/w0w:ro alpine true; echo $?
> Error: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/tmp\\\" to rootfs \\\"/home/kir/.local/share/containers/storage/overlay/7636ef3650fc91ee4996ccc026532bb3cff7182c0430db662fffb933e0bcadc9/merged\\\" at \\\"/home/kir/.local/share/containers/storage/overlay/7636ef3650fc91ee4996ccc026532bb3cff7182c0430db662fffb933e0bcadc9/merged/w0w\\\" caused \\\"operation not permitted\\\"\"": OCI runtime permission denied error
> 126
```

The last command is not working because in-container mount point
(i.e. mount destination, not the mount source) is used to search
for a parent mount in /proc/self/mountinfo.

And yet the following
```console
$ ./bin/podman run -v /tmp:/run/test:ro alpine true; echo $?
0
```
still works fine! Here's why:

```console
$ mount | grep -E '/run |/tmp '
tmpfs on /run type tmpfs (rw,nosuid,nodev,seclabel,mode=755)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,seclabel)
```

Fixes: https://github.com/containers/libpod/pull/2432 (commit 0f5ae3c)
Related-to: https://github.com/containers/libpod/issues/2312

### Improve

Instead of getting mount options from /proc/self/mountinfo, which is
very costly to read/parse (and can even be unreliable), let's use
statfs(2) to figure out the flags we need.

### DIFF
--- a/pkg/spec/storage.go
+++ b/pkg/spec/storage.go
@@ -855,21 +855,19 @@ func SupercedeUserMounts(mounts []spec.Mount, configMount []spec.Mount) []spec.M
 }
 
 // Ensure mount options on all mounts are correct
-func InitFSMounts(inputMounts []spec.Mount) ([]spec.Mount, error) {
+func InitFSMounts(mounts []spec.Mount) error {
 	// We need to look up mounts so we can figure out the proper mount flags
 	// to apply.
 	systemMounts, err := pmount.GetMounts()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error retrieving system mounts to look up mount options")
+		return errors.Wrapf(err, "error retrieving system mounts to look up mount options")
 	}
 
-	// TODO: We probably don't need to re-build the mounts array
-	var mounts []spec.Mount
-	for _, m := range inputMounts {
+	for i, m := range mounts {
 		if m.Type == TypeBind {
 			baseMnt, err := findMount(m.Source, systemMounts)
 			if err != nil {
-				return nil, errors.Wrapf(err, "error looking up mountpoint for mount %s", m.Source)
+				return errors.Wrapf(err, "error looking up mountpoint for mount %s", m.Source)
 			}
 			var noexec, nosuid, nodev bool
 			for _, baseOpt := range strings.Split(baseMnt.Opts, ",") {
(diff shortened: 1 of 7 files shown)


kolyshkin:
@giuseppe PTAL

rhatdan:
LGTM, but you have compiler errors.

mheon:
/approve
/ok-to-test

I love the simplification of the mount parsing - this is much, much cleaner.

mheon:
Looks like it needs to be gated to only compile on Linux, though, it's blowing up on Darwin.

pkg/spec/storage.go
```diff
@@ -856,45 +856,22 @@ func SupercedeUserMounts(mounts []spec.Mount, configMount []spec.Mount) []spec.M
...
-	// We need to look up mounts so we can figure out the proper mount flags
-	// to apply.
-	systemMounts, err := pmount.GetMounts()
-	if err != nil {
-		return nil, errors.Wrapf(err, "error retrieving system mounts to look up mount options")
-	}
-
 	// TODO: We probably don't need to re-build the mounts array
-	var mounts []spec.Mount
+	mounts := make([]spec.Mount, len(inputMounts))
```
mheon:
```suggestion
	mounts := make([]spec.Mount, 0, len(inputMounts))
```
Shouldn't this initialize capacity to `len(inputMounts)`, but keep the index at 0, since we're appending?

kolyshkin: You're right! But the more I look into users of this function, the more I think we should just fix things in place.

Let me fix it.

kolyshkin: OK I have added a separate commit which changes the function to fix options in place, and convert its users.

kolyshkin:
* changed InitFSMounts to change options in place rather than creating a new slice
* fixed non-Linux compilation (moved getting default options to pkg/util)

kolyshkin:
> Looks like it needs to be gated to only compile on Linux, though, it's blowing up on Darwin.

done

@mheon May I ask why do we care for darwin?

kolyshkin:
Oh, any suggestions about how to add a regression test to CI for this are welcome. I tried it before (see https://github.com/containers/libpod/pull/5673/commits/e336840726c8a23c7b221609f000b6043e822f34) but a CI has passed :-\

rhatdan:
@kolyshkin containers/image which vendors in container/storage is used for skopeo on Macs.  So we want to make sure the tool chain builds for Darwin.

mheon:
The Darwin stuff in Libpod is a result of the current remote darwin client situation. Hopefully, once Podmanv2 is done, we will no longer need to have Podman build on OS X.

kolyshkin:
I fixed the issues with my patch, current CI failures are (to my best knowledge) unrelated.

mheon:
Look like known flakes. Code LGTM

rhatdan:
/lgtm
/hold
hold cancel when tests pass

kolyshkin:
@rhatdan CI is all green now

PR #5677, opened 2020-03-31
exec: fix hang if control path is deleted

giuseppe:
if the control path file is deleted, libpod hangs waiting for a reader
to open it.  Attempt to open it as non blocking until it returns an
error different than EINTR or EAGAIN.

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -575,13 +575,36 @@ func (r *ConmonOCIRuntime) HTTPAttach(ctr *Container, httpConn net.Conn, httpBuf
 	}
 }
 
+// isRetryable returns whether the error was caused by a blocked syscall or the
+// specified operation on a non blocking file descriptor wasn't ready for completion.
+func isRetryable(err error) bool {
+	if errno, isErrno := errors.Cause(err).(syscall.Errno); isErrno {
+		return errno == syscall.EINTR || errno == syscall.EAGAIN
+	}
+	return false
+}
+
+// openControlFile opens the terminal control file.
+func openControlFile(ctr *Container, parentDir string) (*os.File, error) {
+	controlPath := filepath.Join(parentDir, "ctl")
+	for i := 0; i < 600; i++ {
+		controlFile, err := os.OpenFile(controlPath, unix.O_WRONLY|unix.O_NONBLOCK, 0)
+		if err == nil {
+			return controlFile, err
+		}
+		if !isRetryable(err) {
+			return nil, errors.Wrapf(err, "could not open ctl file for terminal resize for container %s", ctr.ID())
+		}
+		time.Sleep(time.Second / 10)
+	}
+	return nil, errors.Errorf("timeout waiting for %q", controlPath)
+}
+
 // AttachResize resizes the terminal used by the given container.
 func (r *ConmonOCIRuntime) AttachResize(ctr *Container, newSize remotecommand.TerminalSize) error {
-	// TODO: probably want a dedicated function to get ctl file path?
-	controlPath := filepath.Join(ctr.bundlePath(), "ctl")
-	controlFile, err := os.OpenFile(controlPath, unix.O_WRONLY, 0)
+	controlFile, err := openControlFile(ctr, ctr.bundlePath())
 	if err != nil {
-		return errors.Wrapf(err, "could not open ctl file for terminal resize")
+		return err
 	}
 	defer controlFile.Close()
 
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@mheon @vrothberg @baude @jwhonce @QiWang19 @TomSweeneyRedHat PTAL

libpod/oci_conmon_linux.go
```diff
@@ -783,13 +783,29 @@ func (r *ConmonOCIRuntime) ExecContainer(c *Container, sessionID string, options
 	return pid, attachChan, err
 }
 
+func isRetryable(err error) bool {
```
vrothberg: A comment would be great to recall in which cases we're returning true and why.

giuseppe: added and pushed a new version

libpod/oci_conmon_linux.go
```diff
@@ -783,13 +783,31 @@ func (r *ConmonOCIRuntime) ExecContainer(c *Container, sessionID string, options
...
+// specified operation on a non blocking file descriptor wasn't ready for completion.
+func isRetryable(err error) bool {
+	if errno, isErrno := errors.Cause(err).(syscall.Errno); isErrno {
+		return errno == syscall.EINTR || errno == syscall.EAGAIN
+	}
+	return false
+}
+
 // ExecAttachResize resizes the TTY of the given exec session.
 func (r *ConmonOCIRuntime) ExecAttachResize(ctr *Container, sessionID string, newSize remotecommand.TerminalSize) error {
```
mheon: There's also an AttachResize function above; can you make the same changes in there?

libpod/oci_conmon_linux.go
```diff
@@ -783,13 +783,31 @@ func (r *ConmonOCIRuntime) ExecContainer(c *Container, sessionID string, options
...
 	// TODO: probably want a dedicated function to get ctl file path?
 	controlPath := filepath.Join(ctr.execBundlePath(sessionID), "ctl")
-	controlFile, err := os.OpenFile(controlPath, unix.O_WRONLY, 0)
-	if err != nil {
-		return errors.Wrapf(err, "could not open ctl file for terminal resize for container %s exec session %s", ctr.ID(), sessionID)
+	var controlFile *os.File
+	var err error
+
+	for {
```
mheon: Should we build in a max wait for this loop? 10 seconds or so?

giuseppe: not sure, the control file will eventually be deleted and the loop returns.  But I can add one.  Should I hard-code 10 seconds or play safer with something like 60 seconds?

rhatdan: 60 seconds,   Will this go into a tight loop if the failure happens?

libpod/oci_conmon_linux.go
```diff
@@ -783,13 +783,31 @@ func (r *ConmonOCIRuntime) ExecContainer(c *Container, sessionID string, options
...
+		controlFile, err = os.OpenFile(controlPath, unix.O_WRONLY|unix.O_NONBLOCK, 0)
+		if err == nil {
+			break
+		}
+		if !isRetryable(err) {
+			return errors.Wrapf(err, "could not open ctl file for terminal resize for container %s exec session %s", ctr.ID(), sessionID)
+		}
+		time.Sleep(time.Second / 10)
```
TomSweeneyRedHat: Can we do a logrus.info/warning or something here to let the user know we're sleeping?  10 seconds is a long time to wait at a blinking cursor, never mind 60.

libpod/oci_conmon_linux.go
TomSweeneyRedHat: I like the idea of stopping after X attempts.

libpod/oci_conmon_linux.go
```diff
@@ -575,13 +575,36 @@ func (r *ConmonOCIRuntime) HTTPAttach(ctr *Container, httpConn net.Conn, httpBuf
...
+func isRetryable(err error) bool {
+	if errno, isErrno := errors.Cause(err).(syscall.Errno); isErrno {
+		return errno == syscall.EINTR || errno == syscall.EAGAIN
+	}
+	return false
+}
+
+// openControlFile opens the terminal control file.
+func openControlFile(ctr *Container) (*os.File, error) {
+	controlPath := filepath.Join(ctr.bundlePath(), "ctl")
```
mheon: Is this correct? The bundle path differs for exec and non-exec

mheon: This actually only works for attach, not exec attach - we need to use `filepath.Join(ctr.execBundlePath(sessionID), "ctl")` for exec

giuseppe: thanks for noticing it!  Surprisingly Podman works also with the  `filepath.Join(ctr.bundlePath(), "ctl")` version

libpod/oci_conmon_linux.go
giuseppe: It is the same logic as before.  It is just to avoid a syscall to block and it happens in a goroutine.

I don't think we need to debug it

giuseppe:
rebased, need a new lgtm

PR #5686, opened 2020-03-31
podmanv2 load

baude:
enable podman load for v2

add reexec into main

add systemd build flag to v2 makefile

### DIFF
--- a/cmd/podmanV2/images/load.go
+++ b/cmd/podmanV2/images/load.go
@@ -1,0 +1,61 @@
+package images
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/libpod/image"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/spf13/cobra"
+)
+
+var (
+	loadDescription = "Loads an image from a locally stored archive (tar file) into container storage."
+	loadCommand     = &cobra.Command{
+		Use:               "load [flags] [NAME[:TAG]]",
+		Short:             "Load an image from container archive",
+		Long:              loadDescription,
+		RunE:              load,
+		Args:              cobra.MaximumNArgs(1),
+		PersistentPreRunE: preRunE,
+	}
+)
+
+var (
+	loadOpts entities.ImageLoadOptions
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: loadCommand,
+	})
+
+	loadCommand.SetHelpTemplate(registry.HelpTemplate())
+	loadCommand.SetUsageTemplate(registry.UsageTemplate())
+	flags := loadCommand.Flags()
+	flags.StringVarP(&loadOpts.Input, "input", "i", "", "Read from specified archive file (default: stdin)")
+	flags.BoolVarP(&loadOpts.Quiet, "quiet", "q", false, "Suppress the output")
+	flags.StringVar(&loadOpts.SignaturePolicy, "signature-policy", "", "Pathname of signature policy file")
+	if registry.IsRemote() {
+		_ = flags.MarkHidden("signature-policy")
+	}
+
+}
+
+func load(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		repo, err := image.NormalizedTag(args[0])
+		if err != nil {
+			return err
+		}
+		loadOpts.Name = repo.Name()
+	}
+	response, err := registry.ImageEngine().Load(context.Background(), loadOpts)
+	if err != nil {
+		return err
+	}
+	fmt.Println("Loaded image: " + response.Name)
+	return nil
+}
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -315,3 +315,24 @@ func (ir *ImageEngine) Untag(ctx context.Context, nameOrId string, tags []string
 	}
 	return nil
 }
+
+func (ir *ImageEngine) Load(ctx context.Context, opts entities.ImageLoadOptions) (*entities.ImageLoadReport, error) {
+	var (
+		writer io.Writer
+	)
+	if !opts.Quiet {
+		writer = os.Stderr
+	}
+	name, err := ir.Libpod.LoadImage(ctx, opts.Name, opts.Input, writer, opts.SignaturePolicy)
+	if err != nil {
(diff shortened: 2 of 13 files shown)


cmd/podmanV2/images/load.go
```diff
@@ -0,0 +1,73 @@
...
+
+func load(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		repo, err := image.NormalizedTag(args[0])
+		if err != nil {
+			return err
+		}
+		loadOpts.Name = repo.Name()
+	}
+	//if len(args) > 0 {
```
mheon: Can we either enable this, or at least get a comment on why it's off?

cmd/podmanV2/images/load.go
```diff
@@ -0,0 +1,62 @@
...
+	flags.StringVar(&loadOpts.SignaturePolicy, "signature-policy", "", "Pathname of signature policy file (not usually used)")
+	if registry.IsRemote() {
+		_ = flags.MarkHidden("signature-policy")
+	}
+
+}
+
+func load(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		repo, err := image.NormalizedTag(args[0])
```
jwhonce: This isn't pulling in too much of the libpod/image package, yes?

pkg/domain/infra/abi/images.go
```diff
@@ -264,3 +264,23 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
 // 	copy(report.Report.Id, id)
 // 	return &report, nil
 // }
+func (ir *ImageEngine) Load(ctx context.Context, opts entities.ImageLoadOptions) (*entities.ImageLoadReport, error) {
+	var (
+		writer io.Writer
+	)
+	if !opts.Quiet {
+		writer = os.Stderr
```
jwhonce: Writing to os.Stderr from this layer? abi shouldn't be doing any presentation work.

pkg/domain/infra/tunnel/images.go
```diff
@@ -93,3 +94,12 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
 	}
 	return &entities.ImagePullReport{Images: pulledImages}, nil
 }
+
+func (ir *ImageEngine) Load(ctx context.Context, opts entities.ImageLoadOptions) (*entities.ImageLoadReport, error) {
+	f, err := os.Open(opts.Input)
```
jwhonce: This is opening on the server not the client.

pkg/domain/infra/abi/images.go
vrothberg: The writing happens in c/image for the progress bars during copy.

pkg/domain/infra/abi/images.go
```diff
@@ -264,3 +264,23 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
...
+	name, err := ir.Libpod.LoadImage(ctx, opts.Name, opts.Input, writer, opts.SignaturePolicy)
+	if err != nil {
+		return nil, err
+	}
+	newImage, err := ir.Libpod.ImageRuntime().NewFromLocal(name)
+	if err != nil {
+		return nil, errors.Wrap(err, "image loaded but no additional tags were created")
+	}
+	if err := newImage.TagImage(opts.Name); err != nil {
+		return nil, errors.Wrapf(err, "error adding '%s' to image %q", opts.Name, newImage.InputName)
```
vrothberg, line 283: Can we use `%q` instead of `'%s'`?

pkg/domain/infra/abi/images.go
```diff
@@ -264,3 +264,23 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
...
+		writer io.Writer
+	)
+	if !opts.Quiet {
+		writer = os.Stderr
+	}
+	name, err := ir.Libpod.LoadImage(ctx, opts.Name, opts.Input, writer, opts.SignaturePolicy)
+	if err != nil {
+		return nil, err
+	}
+	newImage, err := ir.Libpod.ImageRuntime().NewFromLocal(name)
```
vrothberg, line 278: `cmd/podman/load.go` does:
``` golang
if len(imageName) > 0 {                                                                                 
        split := strings.Split(names, ",")                                                              
        newImage, err := runtime.NewImageFromLocal(split[0])                                            
        if err != nil {                                                                                 
                return err                                                                              
        }                                                                                               
        if err := newImage.TagImage(imageName); err != nil {                                            
                return errors.Wrapf(err, "error adding '%s' to image %q", imageName, newImage.InputName)
        }                                                                                               
}                                                                                                       
```

I think we should make this and the following check conditional and use `opts.Name` instead of `name`.

Ultimately, (Runtime).LoadImage(..) should really be doing this check and the tagging so users of it don't need to do that.

pkg/domain/infra/tunnel/images.go
vrothberg: It looks good to me. We're opening the file and pass it to the bindings which are throwing it over to the server. We're still on the client site.

pkg/domain/infra/abi/images.go
vrothberg: @baude, can you move these checks into `./libpod`?

cmd/podmanV2/images/load.go
```diff
@@ -0,0 +1,62 @@
...
+		if err != nil {
+			return err
+		}
+		loadOpts.Name = repo.Name()
+	}
+	response, err := registry.ImageEngine().Load(context.Background(), loadOpts)
+	if err != nil {
+		return err
+	}
+	fmt.Println("Loaded image: " + response.Name)
```
vrothberg, line 60: :100: ... in v1 we're (for some reason) expecting multiple images here but currently return only one. Once we support multi-image archives, we can rework it.

cmd/podmanV2/images/load.go
baude: I dont believe so but once we have the client done, we can examine things and perform refectors if needed.

pkg/domain/infra/abi/images.go
jwhonce: @vrothberg Okay, I'll write a future issue on that.

vrothberg:
My previous comments are still open:

* Load does _not_ work and errors out with `Error: error adding "" to image "@a187dde48cd289ac374ad8539930628314bc581a481cdb41409c9289419ddb72": repository name must have at least one component` while v1 is working.
* There is no `image load` command.

Please don't ask me to spend time reviewing if my reviews are not being addressed.

PR #5688, opened 2020-03-31
podmanv2 import

baude:
add the ability to import a container image from a container export

### DIFF
--- a/cmd/podmanV2/images/import.go
+++ b/cmd/podmanV2/images/import.go
@@ -1,0 +1,87 @@
+package images
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/containers/libpod/cmd/podmanV2/parse"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	importDescription = `Create a container image from the contents of the specified tarball (.tar, .tar.gz, .tgz, .bzip, .tar.xz, .txz).
+
+  Note remote tar balls can be specified, via web address.
+  Optionally tag the image. You can specify the instructions using the --change option.`
+	importCommand = &cobra.Command{
+		Use:               "import [flags] PATH [REFERENCE]",
+		Short:             "Import a tarball to create a filesystem image",
+		Long:              importDescription,
+		RunE:              importCon,
+		PersistentPreRunE: preRunE,
+		Example: `podman import http://example.com/ctr.tar url-image
+  cat ctr.tar | podman -q import --message "importing the ctr.tar tarball" - image-imported
+  cat ctr.tar | podman import -`,
+	}
+)
+
+var (
+	importOpts entities.ImageImportOptions
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: importCommand,
+	})
+
+	importCommand.SetHelpTemplate(registry.HelpTemplate())
+	importCommand.SetUsageTemplate(registry.UsageTemplate())
+	flags := importCommand.Flags()
+	flags.StringArrayVarP(&importOpts.Changes, "change", "c", []string{}, "Apply the following possible instructions to the created image (default []): CMD | ENTRYPOINT | ENV | EXPOSE | LABEL | STOPSIGNAL | USER | VOLUME | WORKDIR")
+	flags.StringVarP(&importOpts.Message, "message", "m", "", "Set commit message for imported image")
+	flags.BoolVarP(&importOpts.Quiet, "quiet", "q", false, "Suppress output")
+}
+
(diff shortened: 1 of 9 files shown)


pkg/domain/infra/tunnel/images.go
```diff
@@ -93,3 +94,20 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
...
+func (ir *ImageEngine) Import(ctx context.Context, opts entities.ImageImportOptions) (*entities.ImageImportReport, error) {
+	var (
+		err       error
+		sourceURL *string
+		f         *os.File
+	)
+	if opts.SourceIsURL {
+		sourceURL = &opts.Source
+	} else {
+		f, err = os.Open(opts.Source)
```
jwhonce: This is importing off the server, not the client.  Won't the user expect the file to be on their machine?

cmd/podmanV2/images/import.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	importOpts.Source = source
+	importOpts.Reference = reference
+
+	response, err := registry.ImageEngine().Import(context.Background(), importOpts)
+	if err != nil {
+		return err
+	}
+	fmt.Println(response.Id)
```
vrothberg, line 85: Unrelated to this PR. I don't get why we're handcrafting return types in `entities`. Non of these types need to be marshaled, so we could very well just return a `string` in this case (and many others). No need to change now but this is just a general comment.

pkg/domain/infra/tunnel/images.go
vrothberg: We're still on the client here and are passing it to the bindings which eventually throw it over to the server.

cmd/podmanV2/images/import.go
```diff
@@ -0,0 +1,87 @@
...
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: importCommand,
+	})
+
+	importCommand.SetHelpTemplate(registry.HelpTemplate())
+	importCommand.SetUsageTemplate(registry.UsageTemplate())
+	flags := importCommand.Flags()
+	flags.StringArrayVarP(&importOpts.Changes, "change", "c", []string{}, "Apply the following possible instructions to the created image (default []): CMD | ENTRYPOINT | ENV | EXPOSE | LABEL | STOPSIGNAL | USER | VOLUME | WORKDIR")
+	flags.StringVarP(&importOpts.Message, "message", "m", "", "Set commit message for imported image")
+	flags.BoolVarP(&importOpts.Quiet, "quiet", "q", false, "Suppress output")
```
vrothberg, line 47: Long-term proposal: I'd love to have an `--input -i` as we have for `load`. We inherited it from Docker but can remain backward compat even when adding a new option.

pkg/domain/entities/images.go
```diff
@@ -171,3 +171,16 @@ type ImagePruneReport struct {
 	Report Report
 	Size   int64
 }
+
+type ImageImportOptions struct {
```
vrothberg, line 175: Nit: if we comment all this stuff now, we don't need to do it in the future. Let's name it while it's hot?

pkg/domain/entities/images.go
```diff
@@ -171,3 +171,16 @@ type ImagePruneReport struct {
...
+	Changes     []string
+	Message     string
+	Quiet       bool
+	Reference   string
+	Source      string
+	SourceIsURL bool
+}
+
+type ImageImportReport struct {
+	Id string
```
vrothberg, line 185: s/Id/ID/ ... for the linter

cmd/podmanV2/images/import.go
```diff
@@ -0,0 +1,87 @@
...
+var (
+	importDescription = `Create a container image from the contents of the specified tarball (.tar, .tar.gz, .tgz, .bzip, .tar.xz, .txz).
+
+  Note remote tar balls can be specified, via web address.
+  Optionally tag the image. You can specify the instructions using the --change option.`
+	importCommand = &cobra.Command{
+		Use:               "import [flags] PATH [REFERENCE]",
+		Short:             "Import a tarball to create a filesystem image",
+		Long:              importDescription,
+		RunE:              importCon,
```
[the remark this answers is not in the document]
baude: i was trying to say importContainer :)

cmd/podmanV2/images/import.go
baude: we agreed in principle that everything should feed a struct and return a struct for future wiggle room

pkg/domain/entities/images.go
baude: we get into this push and pull between go templates and the linter here. leaving for now.

vrothberg:
Making chief happy.
/lgtm
/hold

cmd/podmanV2/images/import.go
```diff
@@ -0,0 +1,87 @@
...
+		PersistentPreRunE: preRunE,
+		Example: `podman import http://example.com/ctr.tar url-image
+  cat ctr.tar | podman -q import --message "importing the ctr.tar tarball" - image-imported
+  cat ctr.tar | podman import -`,
+	}
+)
+
```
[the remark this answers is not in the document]
vrothberg: Still outstanding.

PR #5689, opened 2020-03-31
podmanv2 pod inspect

sujil02:
Add the ability to inspect pod in podmanv2

### DIFF
--- a/cmd/podmanV2/pods/inspect.go
+++ b/cmd/podmanV2/pods/inspect.go
@@ -1,0 +1,64 @@
+package pods
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	jsoniter "github.com/json-iterator/go"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	inspectOptions = entities.PodInspectOptions{}
+)
+
+var (
+	inspectDescription = fmt.Sprintf(`Display the configuration for a pod by name or id
+
+	By default, this will render all results in a JSON array.`)
+
+	inspectCmd = &cobra.Command{
+		Use:     "inspect [flags] POD [POD...]",
+		Short:   "Displays a pod configuration",
+		Long:    inspectDescription,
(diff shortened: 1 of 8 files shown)


jwhonce:
@baude Should `podman inspect` be enhanced to call pod inspect now?

mheon:
IMO, no - `docker inspect` doesn't do volumes, for example, last I checked.

TomSweeneyRedHat:
LGTM assuming happy tests

cmd/podmanV2/pods/inspect.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func inspect(cmd *cobra.Command, args []string) error {
+
+	if len(args) < 1 && !inspectOptions.Latest {
+		return errors.Errorf("you must provide the name or id of a running pod")
+	}
+
+	if !inspectOptions.Latest {
+		topOptions.NameOrID = args[0]
```
vrothberg, line 52:
```suggestion
		inspectOptions.NameOrID = args[0]
```

sujil02: Fixed. Wasn't aware CI isn't in place. retested now works fine.

vrothberg:
@rhatdan @baude @mheon PTAL

cmd/podmanV2/pods/inspect.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func inspect(cmd *cobra.Command, args []string) error {
+
+	if len(args) < 1 && !inspectOptions.Latest {
+		return errors.Errorf("you must provide the name or id of a running pod")
+	}
+
+	if !inspectOptions.Latest {
+		inspectOptions.NameOrID = args[0]
```
mheon, line 52: If we're just discarding other arguments, should we ensure that there is at most 1 argument passed?

pkg/domain/entities/pods.go
```diff
@@ -3,6 +3,7 @@ package entities
 import (
 	"time"
 
+	"github.com/containers/libpod/libpod"
```
mheon, line 6: @baude We OK with this?

baude: no, i will follow up with you and sujil

PR #5696, opened 2020-04-01
Fix environment handling from containers.conf

rhatdan:

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -538,14 +538,16 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 		}
 	}
 
-	// env overrides any previous variables
-	cmdlineEnv := c.StringSlice("env")
-	if len(cmdlineEnv) > 0 {
-		parsedEnv, err := envLib.ParseSlice(cmdlineEnv)
-		if err != nil {
-			return nil, err
+	if c.IsSet("env") {
+		// env overrides any previous variables
+		cmdlineEnv := c.StringSlice("env")
+		if len(cmdlineEnv) > 0 {
+			parsedEnv, err := envLib.ParseSlice(cmdlineEnv)
+			if err != nil {
+				return nil, err
+			}
+			env = envLib.Join(env, parsedEnv)
 		}
-		env = envLib.Join(env, parsedEnv)
 	}
 
 	// LABEL VARIABLES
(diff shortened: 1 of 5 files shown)


rhatdan:
Fixes https://github.com/containers/libpod/issues/5685

cmd/podman/shared/create.go
```diff
@@ -538,18 +538,29 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
...
 
-	// env overrides any previous variables
-	cmdlineEnv := c.StringSlice("env")
-	if len(cmdlineEnv) > 0 {
-		parsedEnv, err := envLib.ParseSlice(cmdlineEnv)
-		if err != nil {
-			return nil, err
+	if c.IsSet("env") {
+		// env overrides any previous variables
+		cmdlineEnv := c.StringSlice("env")
```
mheon, line 543: Woah. Is ENV a string slice? That separates automatically on `,` which means we can't have commas in environment variables - that doesn't seem right?

mheon:
LGTM once things go green. I have some questions, but they're unrelated to this PR.

cmd/podman/shared/create.go
rhatdan: Perhaps a different issue, it has been this way for a while.

rhatdan:
@giuseppe PTAL

giuseppe:
LGTM

cmd/podman/shared/create.go
larsks: @mheon fwiw, commas in environment variables seems to work just fine.

cmd/podman/shared/create.go
```diff
@@ -538,18 +538,29 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
...
-		env = envLib.Join(env, parsedEnv)
+	}
+
+	labelsList := []string{}
+	if c.IsSet("label") {
+		labelsList = c.StringArray("label")
+	}
+	labelFiles := []string{}
+	if c.IsSet("label-files") {
+		labelFiles = c.StringSlice("label-file")
```
TomSweeneyRedHat, line 559: I think line 558 should be singular, "label-file"

TomSweeneyRedHat:
@rhatdan all kinds of test unhappiness.

cmd/podman/shared/create.go
rhatdan: Great find.

rhatdan:
Yes for some reason, my new test will not pass, although works when I run this by hand.  Will continue playing with it.

PR #5697, opened 2020-04-01
podmanv2 export

baude:
add ability to export a container to a tarball

### DIFF
--- a/cmd/podmanV2/containers/export.go
+++ b/cmd/podmanV2/containers/export.go
@@ -1,0 +1,57 @@
+package containers
+
+import (
+	"context"
+	"os"
+
+	"github.com/containers/libpod/cmd/podmanV2/parse"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+var (
+	exportDescription = "Exports container's filesystem contents as a tar archive" +
+		" and saves it on the local machine."
+
+	exportCommand = &cobra.Command{
+		Use:               "export [flags] CONTAINER",
+		Short:             "Export container's filesystem contents as a tar archive",
+		Long:              exportDescription,
+		PersistentPreRunE: preRunE,
+		RunE:              export,
+		Args:              cobra.ExactArgs(1),
+		Example: `podman export ctrID > myCtr.tar
+  podman export --output="myCtr.tar" ctrID`,
+	}
+)
+
+var (
+	exportOpts entities.ContainerExportOptions
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: exportCommand,
+	})
+	exportCommand.SetHelpTemplate(registry.HelpTemplate())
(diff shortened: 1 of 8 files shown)


pkg/api/server/register_containers.go
```diff
@@ -587,6 +587,29 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
...
+	//  - in: path
+	//    name: name
+	//    type: string
+	//    required: true
+	//    description: the name or ID of the container
+	// produces:
+	// - application/json
+	// responses:
+	//   200:
+	//     description: ok
```
jwhonce, line 606: The body should contain the tarball.

pkg/api/server/register_containers.go
```diff
@@ -1237,5 +1260,27 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
...
```
jwhonce, line 1279: Should return the tarball

cmd/podmanV2/containers/export.go
```diff
@@ -0,0 +1,57 @@
...
+	if len(exportOpts.Output) == 0 {
+		file := os.Stdout
+		if terminal.IsTerminal(int(file.Fd())) {
+			return errors.Errorf("refusing to export to terminal. Use -o flag or redirect")
+		}
+		exportOpts.Output = "/dev/stdout"
+	} else if err := parse.ValidateFileName(exportOpts.Output); err != nil {
+		return err
+	}
+	return registry.ContainerEngine().ContainerExport(context.Background(), args[0], exportOpts)
```
jwhonce, line 56: Why ship the filename upstream? A writer would fit the interface better.  ContainerExport() could just use the writer.

pkg/api/handlers/compat/containers_export.go
```diff
@@ -0,0 +1,42 @@
...
+		utils.Error(w, "failed to save the image", http.StatusInternalServerError, errors.Wrap(err, "failed to save image"))
+		return
+	}
+	rdr, err := os.Open(tmpfile.Name())
+	if err != nil {
+		utils.Error(w, "failed to read temp tarball", http.StatusInternalServerError, errors.Wrap(err, "failed to read the exported tarfile"))
+		return
+	}
+	defer rdr.Close()
+	defer os.Remove(tmpfile.Name())
```
mheon, line 40: This should be moved up to right after the creation of the temporary file

pkg/bindings/containers/containers.go
```diff
@@ -296,3 +297,20 @@ func Stop(ctx context.Context, nameOrID string, timeout *uint) error {
 	}
 	return response.Process(nil)
 }
+
+func Export(ctx context.Context, nameOrID string, w io.Writer) error {
```
mheon, line 301: Needs doc comment

pkg/domain/infra/tunnel/containers.go
```diff
@@ -210,3 +212,18 @@ func (ic *ContainerEngine) ContainerCommit(ctx context.Context, nameOrId string,
...
+	var (
+		err error
+		w   io.Writer
+	)
+	if len(options.Output) > 0 {
+		w, err = os.Create(options.Output)
+		if err != nil {
+			return err
+		}
+
```
mheon, line 226: If you spin another version of this, can you get rid of the extra whitespace here?

cmd/podmanV2/containers/export.go
baude: while i agree that would be cleaner, the libpod method call does not accept that right now.

PR #5701, opened 2020-04-02
podmanV2: implement push

vrothberg:
* Implement `podman-push` and `podman-image-push` for the podmanV2
  client.

* Tests for `pkg/bindings` are not possible at the time of writing as we
  don't have a local registry running.

* Implement `/images/{name}/push` compat endpoint. Tests are not
  implemented for this v2 endpoint.  It has been tested manually.

General note: The auth config extraction from the http header is not
implement for push.  Since it's not yet supported for other endpoints
either, I deferred it to future work.

### DIFF
--- a/cmd/podmanV2/images/push.go
+++ b/cmd/podmanV2/images/push.go
@@ -1,0 +1,127 @@
+package images
+
+import (
+	buildahcli "github.com/containers/buildah/pkg/cli"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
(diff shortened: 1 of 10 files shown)


vrothberg:
For some reason, we get `<nil>` in the bindings when using `func ImageNotFound(w http.ResponseWriter, name string, err error)` in the API. I tried it with `image exists` and get the same behavior, so it seems unrelated to the changes of this PR. Will investigate.

vrothberg:
@mheon @jwhonce @baude PTAL

baude:
yeah lgtm, lets get this in

PR #5705, opened 2020-04-02
Cleanup whether to enter user namespace

rhatdan:

### DIFF
--- a/cmd/podman/main_local.go
+++ b/cmd/podman/main_local.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"runtime/pprof"
 	"strconv"
-	"strings"
 	"syscall"
 
 	"github.com/containers/common/pkg/config"
@@ -192,7 +191,7 @@ func setupRootless(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	if os.Geteuid() == 0 || cmd == _searchCommand || cmd == _versionCommand || cmd == _mountCommand || cmd == _migrateCommand || strings.HasPrefix(cmd.Use, "help") {
+	if !executeCommandInUserNS(cmd) {
 		return nil
 	}
 
@@ -241,6 +240,25 @@ func setupRootless(cmd *cobra.Command, args []string) error {
 		os.Exit(ret)
 	}
 	return nil
+}
+
+// Most podman commands when run in rootless mode, need to be executed in the
+// users usernamespace.  This function is updated with a  list of commands that
+// should NOT be run within the user namespace.
+func executeCommandInUserNS(cmd *cobra.Command) bool {
+	if os.Geteuid() == 0 {
+		return false
+	}
+	switch cmd {
+	case _migrateCommand,
+		_mountCommand,
+		_renumberCommand,
+		_infoCommand,
+		_searchCommand,
+		_versionCommand:
+		return false
+	}
+	return true
 }
 
 func setRLimits() error {


rhatdan:
Fixes problems found in https://github.com/containers/libpod/issues/5306

cmd/podman/main_local.go
```diff
@@ -243,6 +243,27 @@ func setupRootless(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func useUserNS(cmd *cobra.Command) bool {
+	if os.Geteuid() == 0 {
+		return false
+	}
+	switch cmd {
+	case _migrateCommand,
+		_migrateCommand,
```
vrothberg, line 252:
```suggestion
```

cmd/podman/main_local.go
```diff
@@ -243,6 +243,27 @@ func setupRootless(cmd *cobra.Command, args []string) error {
...
+		_infoCommand,
+		_searchCommand,
+		_systemResetCommand,
+		_versionCommand:
+		return false
+	}
+	if strings.HasPrefix(cmd.Use, "help") {
+		return false
+	}
+	return true
```
vrothberg, line 264:
```suggestion
	return !strings.HasPrefix(cmd.Use, "help")
```

vrothberg: Can you add a comment to this check? As far as I understand it, we are checking if the "Use" string of the command starts with "help" but there are no such occurrences in the code:
```bash
$ grep "Use:" -r cmd/ | grep help | wc -l
0
```

giuseppe:
I am afraid we won't be able to rm files not owned by the rootless user without entering the user namespace.

rhatdan:
Yes I just thought of the same thing.

cmd/podman/main_local.go
rhatdan: I think this was added for commands like

podman run --help

rhatdan:
This PR no longer fixes the issue, but I think it cleans up the code and adds

cmd/podman/main_local.go
rhatdan: Fixed

cmd/podman/main_local.go
rhatdan: Done.

vrothberg: `Cmd.Use` is the wrong field for that:
```golang
// Command is just that, a command for your application.              
// E.g.  'go run ...' - 'run' is the command. Cobra requires          
// you to define the usage and description as part of your command    
// definition to ensure usability.                                    
type Command struct {                                                 
        // Use is the one-line usage message.                         
        Use string                                                    
```

vrothberg: I printf-debuged this code path. If we run `--help` we don't execute the code path. I guess we can just `return true`.

rhatdan: Nice, so I guess this code is not needed, removed.

cmd/podman/main_local.go
```diff
@@ -243,6 +243,22 @@ func setupRootless(cmd *cobra.Command, args []string) error {
...
+	case _migrateCommand,
+		_mountCommand,
+		_renumberCommand,
+		_infoCommand,
+		_searchCommand,
+		_versionCommand:
+		return false
+	}
+	return true
+}
```
TomSweeneyRedHat, line 260: I'd like a better function name and/or a function comment over what this function is supposed to do.

mheon:
One question: can we still remove containers without entering the user namespace? We may need to delete files not owned by the current user.

I was thinking we needed a way to enter the userns without joining the pause process.

cmd/podman/main_local.go
rhatdan: Suggestions?

rhatdan: Changed name to executeCommandInUserNS
and added a comment.

giuseppe:
> I was thinking we needed a way to enter the userns without joining the pause process.

We have a fallback in place for that.  When the pause process cannot be joined in some cases it is enough to just create a new namespace.  Looks like something is going wrong there when not using systemd

PR #5707, opened 2020-04-03
Prepare for crun checkpoint support

adrianreber:
This includes two small changes for using crun with checkpoint support:

 * correctly set up XDG_RUNTIME_DIR (thanks to @giuseppe for helping with this)
 * adapt checkpoint support test to work with crun and runc


giuseppe:
/approve

LGTM

rhatdan:
This looks like the checkpoint/restore tests are blowing up?

adrianreber:
It always fails with
```
CRIU version check failed: exec: "criu": executable file not found in $PATH
```

Very strange. Ah, maybe by setting up a new environment runc no longer gets the full PATH from the current shell and does not find the CRIU binary in /usr/sbin. I probably need to add /usr/sbin to the PATH passed down to runc.

rhatdan:
When I saw your change, that is what I was thinking.  You should just add the XDG_RUNTIME_DIR to an existing environment.

TomSweeneyRedHat:
LGTM assuming happy tests

PR #5709, opened 2020-04-03
podmanV2: implement search

vrothberg:
Also implement a new libpod endpoint to add more parameters and to
prevent us from converting between slices and maps and make use of
the filter parsing in the image backend.


vrothberg:
@mheon @jwhonce @baude PTAL

vrothberg:
@rhatdan PTAL

vrothberg:
Ready to merge.

cmd/podmanV2/images/search.go
```diff
@@ -0,0 +1,158 @@
...
+
+import (
+	"reflect"
+	"strings"
+
+	buildahcli "github.com/containers/buildah/pkg/cli"
+	"github.com/containers/buildah/pkg/formats"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/libpod/image"
```
rhatdan, line 11: I don't think this is allowed.
@baude Correct?

rhatdan: @mheon ?

mheon: Concur, we're trying to avoid using libpod in podmanv2

rhatdan:
We should have a tests that just greps through the code in cmd/podman/podmanV2 and looks for libpod/libpod, if it finds it, throw an error.

cmd/podmanV2/images/search.go
vrothberg: I would prefer waiting for the image-consolidation to be done (each line of code gets more expensive with every change) but sure, I can move the type somewhere else.

vrothberg: I looked into it and don't think it's a good idea to break parts out of `libpod/image`. It could really be somewhere in `pkg/` and is not tied to `libpod`. Breaking parts out would IMO makes things worse.

Can we wait for the image consolidation?

rhatdan: The issue, is a matter of size.  In that this will cause libpod to be pulled in swelling the size of the client.
But I will let @baude @jwhonce @mheon Decide on this.
One option would be, could we move libpod/image to pkg/image for the time being?

vrothberg: Ah. In that case, the image package would pull in c/storage and a couple of other rather large packages. However, `libpod/image` is already pulled in via `podman load`.

mheon: That's not really good, since separating Libpod and everything in the frontend is a core goal of this refactor.

cmd/podmanV2/images/search.go
```diff
@@ -0,0 +1,158 @@
...
+
+// imageSearch implements the command for searching images.
+func imageSearch(cmd *cobra.Command, args []string) error {
+	searchTerm := ""
+	switch len(args) {
+	case 1:
+		searchTerm = args[0]
+	default:
+		return errors.Errorf("search requires exactly one argument")
+	}
```
jwhonce, line 106:
```suggestion
```
In command declaration: Args: cobra.ExactArgs(1) will take care of this.

cmd/podmanV2/images/search.go
vrothberg: Alright. I'll work around it.

> That's not really good, since separating Libpod and everything in the frontend is a core goal of this refactor.

What is considered part of libpod? We are importing `libpod/define` in the v2 frontend. Is there a clear differentiator between what goes into `define` and what into `entities`?

I begin to see wisdom in Dan's proposal to enforce such rules in the CI.

mheon: Define is Libpod-specific structures and constants (with a few exceptions - anything that goes in the DB, for example, needs to stay in Libpod). Anything struct or constant Libpod returns that doesn't fall under one of those exceptions will probably move to define/ eventually. The important point is that define is structs only, no code, and very minimal dependencies, so we can import it anywhere without much consequence.

As to what define is vs. entities, define is on the Libpod side of the libpod/Podman split, entities is on the Podman side. When we eventually split the repo, I expect entities to be moved into the new Podman repo.

cmd/podmanV2/images/search.go
```diff
@@ -0,0 +1,158 @@
...
+		PreRunE: searchCmd.PreRunE,
+		RunE:    searchCmd.RunE,
+		Example: `podman image search --filter=is-official --limit 3 alpine
+		podman image search registry.fedoraproject.org/  # only works with v2 registries
+		podman image search --format "table {{.Index}} {{.Name}}" registry.fedoraproject.org/fedora`,
+	}
+)
+
+func init() {
+	// pull
```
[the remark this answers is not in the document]
vrothberg: C&P :) Thanks!

cmd/podmanV2/images/search.go
vrothberg: That's nice, thanks!

cmd/podmanV2/images/search.go
vrothberg: I worked around it by converting between `libpod/image` and the `entities` search result.

Regarding the desired clean separation between entities and libpod:

I had a closer look and fear that achieving the goal will require some further massaging until there's a clean cut. After some visual inspection, I ran `./dependencies/analyses/dependency-tree.sh` and found quite some (transitive) dependencies on `libpod`:

> github.com/containers/libpod/libpod, github.com/containers/libpod/libpod/define, github.com/containers/libpod/libpod/driver, github.com/containers/libpod/libpod/events, github.com/containers/libpod/libpod/image, github.com/containers/libpod/libpod/layers, github.com/containers/libpod/libpod/lock, github.com/containers/libpod/libpod/lock/file, github.com/containers/libpod/libpod/lock/shm, github.com/containers/libpod/libpod/logs, github.com/containers/libpod/libpod/logs/reversereader

There are many more low-level imports on buildah, docker, containerd, containers/storage etc.

I think that a clean separation requires a translation between low-level types and the high-level ones in `entities` (which then require copies of the low-level types).

vrothberg:

cmd/podmanV2/images/search.go
```diff
@@ -0,0 +1,157 @@
...
+// searchFlags set the flags for the pull command.
+func searchFlags(flags *pflag.FlagSet) {
+	flags.StringSliceVarP(&searchOptions.Filters, "filter", "f", []string{}, "Filter output based on conditions provided (default [])")
+	flags.StringVar(&searchOptions.Format, "format", "", "Change the output format to a Go template")
+	flags.IntVar(&searchOptions.Limit, "limit", 0, "Limit the number of results")
+	flags.BoolVar(&searchOptions.NoTrunc, "no-trunc", false, "Do not truncate the output")
+	flags.StringVar(&searchOptions.Authfile, "authfile", buildahcli.GetDefaultAuthFile(), "Path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
+	flags.BoolVar(&searchOptions.TLSVerifyCLI, "tls-verify", true, "Require HTTPS and verify certificates when contacting registries")
+	if registry.IsRemote() {
+		_ = flags.MarkHidden("authfile")
```
rhatdan, line 92: Is there a reason this is hidden?  Do we need a way to handle this in remote?  With the new API is there a way to pass this up?

vrothberg: It's hidden in other commands and in v1 as well.

There is currently now way to pass this in v2. The Docker API does that via base64 encoded authfile in the http header but we don't support that yet. I think we need to support it not only for the Docker endpoints but also for us. I suggest to cover that with remote login/logout as it really needs to be consistent across all endpoints that deal with authenticate.

Can we get this PR in?

PR #5710, opened 2020-04-03
v2podman container create

baude:
create a container in podmanv2 using specgen approach.  this is the core implementation and still has quite a bit of code commented out specifically around volumes, devices, and namespaces.  need contributions from smes on these parts.


baude:
> If you feel comfortable merging this, LGTM. I can't really review since much functionality isn't working/comment out.

@vrothberg create is a massive piece, this seems like an mvp for create and then we nibble our way to fully functional.

cmd/podmanV2/common/default.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/containers/libpod/pkg/sysinfo"
+	"github.com/opencontainers/selinux/go-selinux"
+)
+
+func getDefaultSecurityOptions() []string {
+	securityOpts := []string{}
+	if defaultContainerConfig.Containers.SeccompProfile != "" && defaultContainerConfig.Containers.SeccompProfile != parse.SeccompDefaultPath {
+		securityOpts = append(securityOpts, fmt.Sprintf("seccomp=%s", defaultContainerConfig.Containers.SeccompProfile))
+	}
+	if apparmor.IsEnabled() && defaultContainerConfig.Containers.ApparmorProfile != "" {
```
mheon, line 20: These checks can't be done on the client. Apprmor/SELinux enabled checks must be server side.

cmd/podmanV2/common/ports.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// ExposedPorts parses user and image ports and returns binding information
+func ExposedPorts(expose []string, publish []ocicni.PortMapping, publishAll bool, imageExposedPorts map[string]struct{}) ([]ocicni.PortMapping, error) {
+	containerPorts := make(map[string]string)
+
+	// TODO this needs to be added into a something that
+	// has access to an imageengine
+	// add expose ports from the image itself
+	//for expose := range imageExposedPorts {
```
mheon, line 21: We'll have to move this into specgen as a bool (PublishImagePorts maybe)

libpod/container.go
```diff
@@ -425,18 +425,6 @@ type ContainerConfig struct {
 	HealthCheckConfig *manifest.Schema2HealthConfig `json:"healthcheck"`
 }
 
-// ContainerNamedVolume is a named volume that will be mounted into the
-// container. Each named volume is a libpod Volume present in the state.
-type ContainerNamedVolume struct {
```
mheon, line 430: We can't move this. It's a database struct.

We'll have to make a copy that we can use outside of Libpod.

baude: mheon, then we will need to change the struct in specgen to something else.

baude:
updated

mheon:
LGTM. Let's get something landed and iterate from there.

stevekuznetsov:
boop

PR #5713, opened 2020-04-03
podmanv2 version

baude:
this is only for the abi side and does not print out the remote information yet.  the remote information will need to be worked on and agreed upon later.


rhatdan:
LGTM

PR #5716, opened 2020-04-04
podmanv2 ps

baude:
add the ability to list containers


pkg/domain/entities/container_ps.go
```diff
@@ -0,0 +1,167 @@
...
+	// Port mappings
+	Ports []ocicni.PortMapping
+	// Size of the container rootfs.  Requires the size boolean to be true
+	Size *shared.ContainerSize
+	// Time when container started
+	StartedAt int64
+	// State of container
+	State string
+	// Mount namespace
+	MNT string `json:"Mnt,omitempty"`
```
mheon, line 54: All these namespaces should be in a nullable separate structure which is included in this

pkg/domain/entities/container_ps.go
```diff
@@ -0,0 +1,167 @@
...
+// Listcontainer describes a container suitable for listing
+type ListContainer struct {
+	// Container command
+	Command []string
+	// Container creation time
+	Created int64
+	// If container has exited/stopped
+	Exited bool
+	// Time container exited
+	ExitedAt int64
```
mheon, line 22: Why int64? Shouldn't this be a time.Time?

pkg/domain/entities/container_ps.go
baude: it was originally in a sep struct but that does not work for output.

pkg/domain/entities/container_ps.go
```diff
@@ -0,0 +1,167 @@
...
+	// If the container is part of Pod, the Pod ID. Requires the pod
+	// boolean to be set
+	Pod string
+	// If the container is part of Pod, the Pod name. Requires the pod
+	// boolean to be set
+	PodName string
+	// Port mappings
+	Ports []ocicni.PortMapping
+	// Size of the container rootfs.  Requires the size boolean to be true
+	Size *shared.ContainerSize
```
mheon, line 48: Should move the struct into here

libpod/filters/containers.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/libpod/define"
+	"github.com/containers/libpod/pkg/timetype"
+	"github.com/containers/libpod/pkg/util"
+	"github.com/pkg/errors"
+)
+
+// GenerateContainerFilterFuncs return ContainerFilter functions based of filter.
+func GenerateContainerFilterFuncs(filter, filterValue string, r *libpod.Runtime) (func(container *libpod.Container) bool, error) {
```
mheon, line 17: Why doesn't this return `[]libpod.ContainerFilter`?

baude: because the iteration is outside this

mheon: I'm also really iffy about this being in libpod - seems to be moving CLI parsing into the Libpod API

mheon: Pod/volume filters seem to be in cmd/podman/shared

pkg/domain/entities/container_ps.go
baude: it would trigger a regression for something like:

```podman ps --ns --format "{{.MNT}}"```

pkg/domain/entities/container_ps.go
baude: that is what it was before i moved it. i am not changing any of the types in the struct.

baude:
@mheon after talking with @jwhonce , we want to keep pkg/ps until we complete this thing. we suspect more code like this will emerge and then we can find a suitable home that makes sense.

rhatdan:
LGTM

PR #5728, opened 2020-04-06
Do not error on pids.current stats if ctr.path is empty

saschagrunert:
If the ctr.path is empty, then we do not try to access
`/sys/fs/cgroup/pids/pids.current` any more because this path will be
wrong in any case. We now return and do not set the PIDs stats.

Refers to https://github.com/cri-o/cri-o/issues/3522

/assign @giuseppe

### DIFF
--- a/pkg/cgroups/pids.go
+++ b/pkg/cgroups/pids.go
@@ -44,8 +44,12 @@ func (c *pidHandler) Destroy(ctr *CgroupControl) error {
 
 // Stat fills a metrics structure with usage stats for the controller
 func (c *pidHandler) Stat(ctr *CgroupControl, m *Metrics) error {
-	var PIDRoot string
+	if ctr.path != "" {
+		// nothing we can do to retrieve the pids.current path
+		return nil
+	}
 
+	var PIDRoot string
 	if ctr.cgroup2 {
 		PIDRoot = filepath.Join(cgroupRoot, ctr.path)
 	} else {


mheon:
Looks like 3 flakes (two networking pulling images, Podman RMI with cached images in special_testing_rootless). Restarted the lot.

mheon:
LGTM but want a signoff from @giuseppe before we merge

pkg/cgroups/pids.go
```diff
@@ -46,10 +46,13 @@ func (c *pidHandler) Destroy(ctr *CgroupControl) error {
 func (c *pidHandler) Stat(ctr *CgroupControl, m *Metrics) error {
 	var PIDRoot string
 
-	if ctr.cgroup2 {
+	if ctr.cgroup2 { // nolint: gocritic
 		PIDRoot = filepath.Join(cgroupRoot, ctr.path)
-	} else {
+	} else if ctr.path != "" {
```
giuseppe, line 51: we could skip it also on cgroup v2.  Since there is no `pids.current` in the cgroup v2 root

saschagrunert: Alright, changed it as suggested.

giuseppe:
thanks, LGTM

saschagrunert:
Hey, can we create a new patch release containing this fix?

rhatdan:
We are planning on releasing podman-1.9.0-rc1 today or Monday.

saschagrunert:
Sounds good, thank you! :pray:

PR #5731, opened 2020-04-06
podmanv2 start

baude:
add the ability to start containers

### DIFF
--- a/cmd/podmanV2/containers/start.go
+++ b/cmd/podmanV2/containers/start.go
@@ -1,0 +1,87 @@
+package containers
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/containers/libpod/cmd/podmanV2/common"
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/cmd/podmanV2/utils"
+	"github.com/containers/libpod/libpod/define"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	startDescription = `Starts one or more containers.  The container name or ID can be used.`
+	startCommand     = &cobra.Command{
+		Use:     "start [flags] CONTAINER [CONTAINER...]",
+		Short:   "Start one or more containers",
+		Long:    startDescription,
+		RunE:    start,
+		PreRunE: preRunE,
+		Args:    cobra.MinimumNArgs(1),
+		Example: `podman start --latest
+  podman start 860a4b231279 5421ab43b45
+  podman start --interactive --attach imageID`,
+	}
+)
+
+var (
+	startOptions entities.ContainerStartOptions
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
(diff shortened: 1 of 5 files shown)


cmd/podmanV2/containers/start.go
```diff
@@ -0,0 +1,88 @@
+package containers
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/containers/libpod/cmd/podmanV2/utils"
+
```
mheon, line 8: Why extra newline?

cmd/podmanV2/containers/start.go
```diff
@@ -0,0 +1,88 @@
...
+	if registry.IsRemote() {
+		_ = flags.MarkHidden("latest")
+		_ = flags.MarkHidden("sig-proxy")
+	}
+
+}
+
+func start(cmd *cobra.Command, args []string) error {
+	var errs utils.OutputErrors
+	if len(args) > 1 && startOptions.Attach {
```
mheon, line 56: Should move this into ContainerStart IMO

pkg/domain/infra/abi/containers.go
```diff
@@ -509,3 +510,110 @@ func (ic *ContainerEngine) ContainerExec(ctx context.Context, nameOrId string, o
 	ec, err = terminal.ExecAttachCtr(ctx, ctr, options.Tty, options.Privileged, options.Envs, options.Cmd, options.User, options.WorkDir, &options.Streams, options.PreserveFDs, options.DetachKeys)
 	return define.TranslateExecErrorToExitCode(ec, err), err
 }
+
+func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []string, options entities.ContainerStartOptions) ([]*entities.ContainerStartReport, error) {
```
mheon, line 514: Combining start and attach into one call in the bindings is going to be iffy for tunnel, but I don't know if we can do otherwise for ABI... Eek.

baude: on the tunnel side, that is going to be ok to have multiple binding calls ... i think we do this elsewhere.

cmd/podmanV2/containers/start.go
baude: i like the idea of catching this right away.  why proceed at all under invalid conditions?

mheon: Fair enough

mheon:
LGTM

pkg/domain/infra/abi/containers.go
```diff
@@ -509,3 +510,110 @@ func (ic *ContainerEngine) ContainerExec(ctx context.Context, nameOrId string, o
...
+			// Handle non-attach start
+			// If the container is in a pod, also set to recursively start dependencies
+			report := &entities.ContainerStartReport{
+				Id:       ctr.ID(),
+				ExitCode: 125,
+			}
+			if err := ctr.Start(ctx, ctr.PodID() != ""); err != nil {
+				//if lastError != nil {
+				//	fmt.Fprintln(os.Stderr, lastError)
+				//}
```
vrothberg, line 603: This can be dropped (no reference to `lastError`).

PR #5733, opened 2020-04-06
Add pod prune for api v2

sujil02:
Add pod prune for api v2.

Add the ability to prune pods for api v2,
Includes the addition of force flag, for client-side prompt.

### DIFF
--- a/pkg/api/handlers/libpod/pods.go
+++ b/pkg/api/handlers/libpod/pods.go
@@ -232,13 +232,20 @@ func PodRestart(w http.ResponseWriter, r *http.Request) {
 func PodPrune(w http.ResponseWriter, r *http.Request) {
 	var (
 		runtime = r.Context().Value("runtime").(*libpod.Runtime)
+		reports []*entities.PodPruneReport
 	)
-	pruned, err := runtime.PrunePods()
+	responses, err := runtime.PrunePods(r.Context())
 	if err != nil {
 		utils.InternalServerError(w, err)
 		return
 	}
-	utils.WriteResponse(w, http.StatusOK, pruned)
+	for k, v := range responses {
+		reports = append(reports, &entities.PodPruneReport{
+			Err: v,
+			Id:  k,
+		})
+	}
+	utils.WriteResponse(w, http.StatusOK, reports)
 }
 
 func PodPause(w http.ResponseWriter, r *http.Request) {
--- a/pkg/api/server/register_pods.go
+++ b/pkg/api/server/register_pods.go
@@ -53,11 +53,7 @@ func (s *APIServer) registerPodsHandlers(r *mux.Router) error {
 	// - application/json
 	// responses:
 	//   200:
-	//     description: tbd
-	//     schema:
-	//       type: object
-	//       additionalProperties:
-	//         type: string
+	//     $ref: '#/responses/PodPruneReport'
 	//   400:
 	//     $ref: "#/responses/BadParamError"
 	//   409:
--- a/pkg/bindings/pods/pods.go
+++ b/pkg/bindings/pods/pods.go
@@ -98,17 +98,19 @@ func Pause(ctx context.Context, nameOrID string) (*entities.PodPauseReport, erro
 	return &report, response.Process(&report)
 }
 
-// Prune removes all non-running pods in local storage.
-func Prune(ctx context.Context) error {
+// Prune by default removes all non-running pods in local storage.
+// And with force set true removes all pods.
+func Prune(ctx context.Context) ([]*entities.PodPruneReport, error) {
+	var reports []*entities.PodPruneReport
 	conn, err := bindings.GetClient(ctx)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	response, err := conn.DoRequest(nil, http.MethodPost, "/pods/prune", nil)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	return response.Process(nil)
+	return reports, response.Process(&reports)
 }
 
 // List returns all pods in local storage.  The optional filters parameter can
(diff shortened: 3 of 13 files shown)


pkg/api/server/register_pods.go
```diff
@@ -51,13 +51,14 @@ func (s *APIServer) registerPodsHandlers(r *mux.Router) error {
...
+	//    type: boolean
+	//    description : force removal of a running pods
 	// responses:
 	//   200:
-	//     description: tbd
-	//     schema:
-	//       type: object
-	//       additionalProperties:
-	//         type: string
+	//   200:
```
vrothberg, line 60: Drop this line.

pkg/api/server/register_pods.go
```diff
@@ -51,13 +51,14 @@ func (s *APIServer) registerPodsHandlers(r *mux.Router) error {
 	// summary: Prune unused pods
 	// produces:
 	// - application/json
+	//  - in: query
+	//    name: force
+	//    type: boolean
+	//    description : force removal of a running pods
```
vrothberg, line 57:
```suggestion
	//    description : force removal of a running pod
```

pkg/bindings/pods/pods.go
```diff
@@ -98,17 +98,23 @@ func Pause(ctx context.Context, nameOrID string) (*entities.PodPauseReport, erro
 	return &report, response.Process(&report)
 }
 
-// Prune removes all non-running pods in local storage.
-func Prune(ctx context.Context) error {
+// Prune by default removes all non-running pods in local storage.
+// And with force set true removes all pods.
+func Prune(ctx context.Context, force *bool) ([]*entities.PodPruneReport, error) {
```
vrothberg, line 103: Do we _need_ force to be a pointer? If not, let's make it call by value.

cmd/podmanV2/pods/prune.go
```diff
@@ -0,0 +1,56 @@
...
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: pruneCommand,
+		Parent:  podCmd,
+	})
+	flags := pruneCommand.Flags()
+	flags.BoolVarP(&pruneOptions.Force, "force", "f", false, "Force removal of all running pods.  The default is false")
+}
+
+func prune(cmd *cobra.Command, args []string) error {
```
vrothberg, line 39: We need to check the arguments here as v1 does:

```
~ $ podman pod prune 234908234908234 23423423
Error: `podman pod prune` takes no arguments
```

pkg/api/handlers/libpod/pods.go
```diff
@@ -230,13 +230,31 @@ func PodRestart(w http.ResponseWriter, r *http.Request) {
 func PodPrune(w http.ResponseWriter, r *http.Request) {
 	var (
 		runtime = r.Context().Value("runtime").(*libpod.Runtime)
+		decoder = r.Context().Value("decoder").(*schema.Decoder)
+		reports []*entities.PodPruneReport
 	)
-	pruned, err := runtime.PrunePods()
+	query := struct {
+		Force bool `schema:"force"`
```
mheon, line 237: This does not match the usage of Force for pruning containers or volumes, we should drop this. Force is client-side-only and removes the yes/no continue prompt.

sujil02: @baude @jwhonce I thought we agreed to add the force for prune pod. 
Should I go ahead had change this to the client-side only?

cmd/podmanV2/pods/prune.go
sujil02: fixed.

pkg/bindings/pods/pods.go
sujil02: _force_ need not be a pointer. However, I noticed we have used pointers in other references of similar use case hence went with maintaining consistency eg: https://github.com/containers/libpod/blob/master/pkg/bindings/pods/pods.go#L155.

Should I still go ahead and change it?

vrothberg: Pointers are usually used to find out whether a certain value has been set on the CLI. This allows for distinguishing it from being set (true/false) and not being set (nil), which allows for using defaults in backend code. As we are not doing that in this case, there's no need to have a pointer.

pkg/api/handlers/libpod/pods.go
sujil02: Fixed the force flag. @mheon PTAL

pkg/bindings/pods/pods.go
sujil02: Thank you. Wrt today discussion force flag for pod prune is not required here hence marking resolved.

rhatdan:
LGTM
@vrothberg @mheon @TomSweeneyRedHat PTAL

cmd/podmanV2/pods/prune.go
```diff
@@ -0,0 +1,75 @@
...
+	var (
+		errs utils.OutputErrors
+	)
+	if len(args) > 0 {
+		return errors.Errorf("`%s` takes no arguments", cmd.CommandPath())
+	}
+	if !pruneOptions.Force {
+		reader := bufio.NewReader(os.Stdin)
+		fmt.Println("WARNING! This will remove all volumes not used by at least one container.")
```
vrothberg, line 52: "This will remove all volumes not used by at least one container"...looks like a copy&paste error.

sujil02: Upps. Fixed it.

rhatdan:
@sujil02 Needs a rebase.

sujil02:
> @sujil02 Needs a rebase.

Yeah on it. It Needs some test fixes as well with the rebase.

sujil02:
Done. Changed the commit messages as CLI force flag edits are no longer required.

PR #5737, opened 2020-04-06
Cirrus: Add support for Fedora 32 beta

cevich:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -34,12 +34,12 @@ env:
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
     ###
-    FEDORA_NAME: "fedora-31"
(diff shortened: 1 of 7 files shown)


cevich:
@mheon @baude @edsantiago PTAL I'm thinking maybe while things are haywire, we sneak in this PR also.  It drops F30 and and brings us up to F31 + 32 and fixes some runc/crun packaging issues.

My motivation is that this PR is a pre-req. for #5853 which we need for buildah and bringing some sanity to our Dockerfile setup.  When F32 proper is released, switching from the beta should be relatively painless.

edsantiago:
LGTM at first pass

edsantiago:
/lgtm

feel free to remove the hold when you think it's ready. (Also needs /approve, which I can't grant)

PR #5752, opened 2020-04-07
Add basic structure of output for APIv2 pod inspect

mheon:
This will replace the structs in use in libpod, which cannot be used as they are also directly involved in the database representation of pods and cannot be moved out of Libpod.


baude:
LGTM

jwhonce:
LGTM, reserve the right to complain once code arrives.

PR #5754, opened 2020-04-07
v2podman ps alter formats

baude:
in order to get the go templating to work for custom input, we now use structure methods instead of template map funcs.  this requires some manipulation of fields so that the funcs can have the proper names.


mheon:
Test failures look legitimate

pkg/domain/entities/container_ps.go
```diff
@@ -46,13 +50,61 @@ type ListContainer struct {
...
-	Size *shared.ContainerSize
+	ContainerSize *shared.ContainerSize
 	// Time when container started
 	StartedAt int64
 	// State of container
-	State string
+	ContainerState string
+}
+
+func (l ListContainer) State() string {
```
mheon, line 62: All functions in this file will need doc comments

mheon:
LGTM otherwise

pkg/domain/entities/container_ps.go
```diff
@@ -46,13 +50,69 @@ type ListContainer struct {
...
+// Size returns the rootfs and virtual sizes in human duration in
+// and output form (string) suitable for ps
+func (l ListContainer) Size() string {
+	virt := units.HumanSizeWithPrecision(float64(l.ContainerSize.RootFsSize), 3)
+	s := units.HumanSizeWithPrecision(float64(l.ContainerSize.RwSize), 3)
+	return fmt.Sprintf("%s (virtual %s)", s, virt)
+}
+
+// Names returns the container name in string format
+func (l ListContainer) Names() string {
```
jwhonce, line 94: Do we need the plural on this method since it is returning a Name rather than a slice?

pkg/domain/entities/container_ps.go
```diff
@@ -46,13 +50,69 @@ type ListContainer struct {
...
+}
+
+// CreatedAt returns the container creation time in string format.  podman
+// and docker both return a timestamped value for createdat
+func (l ListContainer) CreatedAt() string {
+	return time.Unix(l.Created, 0).String()
+}
+
+// CreateHuman allows us to output the created time in human readable format
+func (l ListContainer) CreatedHuman() string {
```
jwhonce, line 114: Should` Created() `be human and `CreatedUnix() `otherwise?  I believe `.Create` is what people would expect yes?

pkg/domain/entities/container_ps.go
baude: Names is the go template key  ... thats what v1podman and docker use

pkg/domain/entities/container_ps.go
baude: .CreatedAt is the template key for podmanv1 and docker ... it is a timestamp not human readable.  I added the created myself thinking it might be useful

PR #5758, opened 2020-04-08
Set exit codes on errors.

rhatdan:
podmanV2 run fedora /etc
Should exit with 126
podmanV2 run fedora /etc/foobar
Should exit with 127
podmanv2 run --fobar fedora echo hello
Should exit with 125
podmanv2 run fedora bash -c "exit 100"
Should exit with 100

### DIFF
--- a/cmd/podmanV2/containers/run.go
+++ b/cmd/podmanV2/containers/run.go
@@ -109,13 +109,14 @@ func run(cmd *cobra.Command, args []string) error {
 	}
 	runOpts.Spec = s
 	report, err := registry.ContainerEngine().ContainerRun(registry.GetContext(), runOpts)
+	// report.ExitCode is set by ContainerRun even it it returns an error
+	registry.SetExitCode(report.ExitCode)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
@baude @mheon @vrothberg @jwhonce PTAL

cmd/podmanV2/containers/run.go
```diff
@@ -108,13 +108,13 @@ func run(cmd *cobra.Command, args []string) error {
 	}
 	runOpts.Spec = s
 	report, err := registry.ContainerEngine().ContainerRun(registry.GetContext(), runOpts)
+	registry.SetExitCode(report.ExitCode)
```
mheon, line 111: Can we get a comment on why this has to be before the return so we don't accidentally undo this later?

mheon:
One nit otherwise LGTM

cmd/podmanV2/containers/run.go
rhatdan: Added

PR #5760, opened 2020-04-08
podmanV2: implement logs

vrothberg:
Implement the `podman {container} logs` for the v2 client. The remote
client does not yet support it.  There's some more work needed for the
rest api; some options are missing (e.g., printing names) while others
are broken (e.g., the until http parameter).

The remote parts will be tackled in a future change.

### DIFF
--- a/cmd/podmanV2/containers/logs.go
+++ b/cmd/podmanV2/containers/logs.go
@@ -1,0 +1,108 @@
+package containers
+
+import (
+	"os"
+
+	"github.com/containers/libpod/cmd/podmanV2/registry"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/libpod/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// logsOptionsWrapper wraps entities.LogsOptions and prevents leaking
+// CLI-only fields into the API types.
+type logsOptionsWrapper struct {
+	entities.ContainerLogsOptions
+
+	SinceRaw string
+}
+
+var (
+	logsOptions     logsOptionsWrapper
+	logsDescription = `Retrieves logs for one or more containers.
+
+  This does not guarantee execution order when combined with podman run (i.e., your run may not have generated any logs at the time you execute podman logs).
+`
+	logsCommand = &cobra.Command{
+		Use:     "logs [flags] CONTAINER [CONTAINER...]",
+		Short:   "Fetch the logs of one or more container",
+		Long:    logsDescription,
+		RunE:    logs,
+		PreRunE: preRunE,
+		Example: `podman logs ctrID
+  podman logs --names ctrID1 ctrID2
+  podman logs --tail 2 mywebserver
+  podman logs --follow=true --since 10m ctrID
+  podman logs mywebserver mydbserver`,
+	}
+
+	containerLogsCommand = &cobra.Command{
+		Use:     logsCommand.Use,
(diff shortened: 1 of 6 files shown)


vrothberg:
@rhatdan @jwhonce @mheon @baude PTAL

cmd/podmanV2/containers/logs.go
```diff
@@ -0,0 +1,110 @@
...
+
+	containerLogsCommand.SetHelpTemplate(registry.HelpTemplate())
+	containerLogsCommand.SetUsageTemplate(registry.UsageTemplate())
+
+	containerLogsFlags := containerLogsCommand.Flags()
+	logsFlags(containerLogsFlags)
+}
+
+func logsFlags(flags *pflag.FlagSet) {
+	flags.BoolVar(&logsOptions.Details, "details", false, "Show extra details provided to the logs")
```
rhatdan, line 83: For some reason Podmna logs has

	flags.SetInterspersed(false)

cmd/podmanV2/containers/logs.go
```diff
@@ -0,0 +1,110 @@
...
+	flags.BoolVarP(&logsOptions.Timestamps, "timestamps", "t", false, "Output the timestamps in the log")
+	flags.BoolVarP(&logsOptions.Names, "names", "n", false, "Output the container name in the log")
+
+	_ = flags.MarkHidden("details")
+}
+
+func logs(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 && logsOptions.Latest {
+		return errors.New("no containers can be specified when using 'latest'")
+	} else if !logsOptions.Latest && len(args) < 1 {
```
rhatdan, line 97: Nit: No need for else here, I believe.

pkg/domain/infra/abi/containers.go
```diff
@@ -709,3 +712,48 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
 	}
 	return &report, nil
 }
+
+func (ic *ContainerEngine) ContainerLogs(ctx context.Context, containers []string, options entities.ContainerLogsOptions) error {
+	if options.Writer == nil {
```
baude, line 717: imho this should be caught at the client level i.e. cmd/...

baude:
one nit else LGTM

cmd/podmanV2/containers/logs.go
```diff
@@ -0,0 +1,110 @@
...
+	// container logs
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode},
+		Command: containerLogsCommand,
+		Parent:  containerCmd,
+	})
+
+	containerLogsCommand.SetHelpTemplate(registry.HelpTemplate())
+	containerLogsCommand.SetUsageTemplate(registry.UsageTemplate())
+
```
jwhonce, line 77:
```suggestion
```
Setup in parent.

pkg/domain/infra/abi/containers.go
vrothberg: We're setting that always in `cmd` but I wanted to be defensive in case things change in the future.

cmd/podmanV2/containers/logs.go
vrothberg: I added it back but I think that's not needed for logs. All arguments must be container names or IDs. But one way or another, we'll error out if flag-like args are passed that we don't know.

vrothberg:
Thanks for the reviews! Ready to merge.

PR #5762, opened 2020-04-08
podmanv2 version format variable name change

baude:
change the name of the variable for version to avoid collision with other commands.


jwhonce:
@baude Did you see @vrothberg wrapper struct to prevent name collisions?

baude:
yeah but for one variable .... meh

PR #5767, opened 2020-04-08
logformat: handle apiv2 results, add anchor links

edsantiago:
apiv2 tests emit TAP-compliant output; recognize it and
highlight it the same way we do BATS tests.

Add anchor links to TAP output, so other tools (e.g.
cirrus-flake-summarize) can link to particular lines

And, remove a "-f" from "wait" in test-apiv2; looks
like there's some version of bash used in some CI VM
that doesn't grok it.

### DIFF
--- a/contrib/cirrus/logformatter
+++ b/contrib/cirrus/logformatter
@@ -220,8 +220,8 @@ END_HTML
             $cirrus_task = $1;
         }
 
-        # BATS handling
-        if ($line =~ /^1\.\.\d+$/) {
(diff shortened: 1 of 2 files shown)


edsantiago:
Tests green. Confirmation that the new code works: [this link](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5675066415382528/html/apiv2_test.log.html#t--00004) should take you to line 4 on the apiv2 log, and all lines should be colored green (for "ok"). Before this PR, all text would be gray, and the line-number anchor would not do anything.

PR #5768, opened 2020-04-08
Fix (make vendor)

mtrmac:
`(export a=b command args)` does not run `(command args)` with `a=b`, it sets `$a` to `b`, and marks variables `$a` `$command` `$args` as exported, i.e. `(command args)` is not run.

So we were not actually running `(go mod tidy)`.

Fix that.


giuseppe:
LGTM

/hold cancel

PR #5777, opened 2020-04-09
Refactor service idle support

### LINKED ISSUE #5599: APIv2: Events streaming behaves weirdly with socket activation

/kind bug

**Description**
This is collection of observations. I want to point out some finding. Also it does not work greatly with [cockpit http client](https://cockpit-project.org/guide/latest/cockpit-http.html) and I am not sure yet if this is problem on cockpit or podman side. I hope that we actually may figure out what is happening and see, what side needs fixing.

I am using current master build. Related https://github.com/containers/libpod/pull/5158

1. Enable and start `podman.socket` (`systemctl enable --now podman.socket`)
2. Notice that `podman.service` is `inactive (dead)` (this is of course expected and correct)
3. Start listening on events. `sudo curl --unix-socket /run/podman/podman.sock http://d/v1.24/events`
4. Notice that `podman.service` is `activating (start)`. (**(1)** Never `running` though - this is one interesting observation)
5. Wait for a bit to make sure it keeps streaming. After minute or two do some action and see it correctly logs it.  (like `podman run fedora` and there are 6 events (from `create` to `cleanup`)) That is all correct.
6. Stop the events streaming (by ctrl+C).
7. Notice that `podman.service` never dies **(2)**. It stays in `activating (start)` forever. (Well, at least for around 10minutes, not sure how this works, maybe there is some exponential backof?). I would guess that when no one is listening, the service should die.

Now, when I try to do the same with cockpit http client. **(3)**
Points 1. and 2. are the same. I create a new connection to `/run/podman/podman.sock` and call `/events` endpoint and attach streaming callback. It works for 5 seconds, and then it disconnects. It does not recognize that there is open connection. Looking into `systemctl status podman` I see it is in `activating (start)` for those 5 seconds and then goes back to `inactive (dead)`. Streaming with Cockpit works fine when the `podman system service` does not die (started with `-t 0`).
I am not sure if this is problem in cockpit http client and in podman not recognizing connection.

If it does not ring any bells (mostly to @jwhonce), than I try to create some simple reproducer with Cockpit. Or if there is some way how I can provide some logs, please let me know.

EDIT: A few more info, how cockpit actually works:
1. It creates [cockpit.channel](https://cockpit-project.org/guide/latest/cockpit-channels.html) with options: `{"method":"GET","path":"/v1.12/events?","superuser":true,"payload":"http-stream2","unix":"/run/podman/podman.sock"}`. [http-stream2](https://github.com/cockpit-project/cockpit/blob/master/doc/protocol.md#payload-http-stream2) documentation.

jwhonce:
* Move connection tracking into APIServer using ConnState()
* Remove Connection counters from CLI code
* Update events handler to support client not closing connection
* Improve logging messages

Fixes #5599

### DIFF
--- a/pkg/api/handlers/compat/events.go
+++ b/pkg/api/handlers/compat/events.go
@@ -1,7 +1,6 @@
 package compat
 
 import (
-	"encoding/json"
 	"fmt"
 	"net/http"
 
@@ -10,6 +9,7 @@ import (
 	"github.com/containers/libpod/pkg/api/handlers"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
 	"github.com/gorilla/schema"
+	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -48,14 +48,27 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
 	}()
 	if eventsError != nil {
 		utils.InternalServerError(w, eventsError)
+		close(eventChannel)
 		return
 	}
 
+	// If client disappears we need to stop listening for events
+	go func(done <-chan struct{}) {
+		<-done
+		close(eventChannel)
+	}(r.Context().Done())
+
+	// Headers need to be written out before turning Writer() over to json encoder
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if flusher, ok := w.(http.Flusher); ok {
+		flusher.Flush()
+	}
+
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
 	coder := json.NewEncoder(w)
 	coder.SetEscapeHTML(true)
 
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
 	for event := range eventChannel {
 		e := handlers.EventToApiEvent(event)
 		if err := coder.Encode(e); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/api/handlers/compat/events.go
```diff
@@ -48,14 +48,27 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
...
+	}(r.Context().Done())
 
+	// Headers need to be written out before turning Writer() over to json encoder
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
+	if flusher, ok := w.(http.Flusher); ok {
+		flusher.Flush()
+	}
+
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
```
mheon, line 68: Why are we doing this inside one function? This should be done once, for the entire package, at global scope

jwhonce: @mheon I'm using the `json` variable as an shortcut to the variable `jsoniter.ConfigCompatibleWithStandardLibrary`. What is the benefit of a global scoped variable? 

I just substituted jsoniter for json package in this case.  If there is some benefit of a global variable I'll be happy to cover that in a follow up PR.

mheon: Global scope replaces all uses of `encoding/json` with `jsoniter` within the packages - which, given its performance benefits and compatibility with the standard library, is desirable. We're doing this within Libpod with great success.

PR #5778, opened 2020-04-09
podmanv2 history and image remove templates

baude:
remove the use of template functions images and history to allow for straight-forward user experience.  instead of templates we use structs and struct methods.


rhatdan:
One small NIT,
LGTM
I like the simplification.

cmd/podmanV2/images/history.go
```diff
@@ -100,24 +104,50 @@ func history(cmd *cobra.Command, args []string) error {
...
-	err = tmpl.Execute(w, results.Layers)
+	err = tmpl.Execute(w, hr)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Failed to print report"))
 	}
 	w.Flush()
 	return nil
 }
+
+type historyreporter struct {
```
mheon, line 127: Any reason we can't hang the methods off entities.ImageHistoryLayer?

cmd/podmanV2/images/list.go
```diff
@@ -9,9 +9,11 @@ import (
 	"text/tabwriter"
 	"text/template"
 	"time"
+	"unicode"
+
+	"github.com/docker/go-units"
```
mheon, line 15: Extra newline

cmd/podmanV2/images/history.go
```diff
@@ -100,24 +104,50 @@ func history(cmd *cobra.Command, args []string) error {
...
+}
+
+func (h historyreporter) CreatedBy() string {
+	if len(h.ImageHistoryLayer.CreatedBy) > 45 {
+		return h.ImageHistoryLayer.CreatedBy[:45-3] + "..."
+	}
+	return h.ImageHistoryLayer.CreatedBy
+}
+
+func (h historyreporter) ID() string {
```
mheon, line 148: Does `--no-trunc` affect this?

cmd/podmanV2/images/history.go
baude: evidently the preference is that it be done like this to keep it encapsulated.

cmd/podmanV2/images/history.go
baude: yup

PR #5799, opened 2020-04-13
Add container subcommands

baude:
where applicable, add the container commands to the container sub menu

### DIFF
--- a/cmd/podmanV2/containers/unpause.go
+++ b/cmd/podmanV2/containers/unpause.go
@@ -30,6 +30,10 @@ func init() {
 	registry.Commands = append(registry.Commands, registry.CliCommand{
 		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
 		Command: unpauseCommand,
+	})
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: unpauseCommand,
 		Parent:  containerCmd,
 	})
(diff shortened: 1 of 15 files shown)


rhatdan:
LGTM

cmd/podmanV2/containers/unpause.go
```diff
@@ -27,6 +27,10 @@ var (
 )
 
 func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: unpauseCommand,
+	})
 	registry.Commands = append(registry.Commands, registry.CliCommand{
 		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
 		Command: unpauseCommand,
```
TomSweeneyRedHat, line 36: nit, it would good to switch the order of lines 36 and 37 to match the other commands.

TomSweeneyRedHat:
One nit for consideration, otherwise LGTM

rhatdan:
@jwhonce and @mheon Have already discussed this. I forget the reason.

jwhonce:
@baude I prefer separate cobra.Commands where the duplicated fields are copied, and unique fields like `Example` can be more correct.  I like to have the "fully qualified" command be the Parent and alias commands be the child. For example, `podman kill` pulls from `podman container kill`. They can live in the same containers.go file.

When the command is a different name, I have been creating new files.  For example, `images/images.go` and `images/list.go`.  Makes finding things much easier.

PR #5803, opened 2020-04-13
v2podman add container init

baude:
add the ability to init a container both local and remote

### DIFF
--- a/pkg/api/handlers/libpod/containers.go
+++ b/pkg/api/handlers/libpod/containers.go
@@ -285,3 +285,23 @@ func Restore(w http.ResponseWriter, r *http.Request) {
 	}
 	utils.WriteResponse(w, http.StatusOK, entities.RestoreReport{Id: ctr.ID()})
 }
+
+func InitContainer(w http.ResponseWriter, r *http.Request) {
+	name := utils.GetName(r)
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	ctr, err := runtime.LookupContainer(name)
+	if err != nil {
+		utils.ContainerNotFound(w, name, err)
+		return
+	}
+	err = ctr.Init(r.Context())
+	if errors.Cause(err) == define.ErrCtrStateInvalid {
+		utils.Error(w, "container already initialized", http.StatusNotModified, err)
+		return
+	}
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	utils.WriteResponse(w, http.StatusNoContent, "")
+}
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -1377,7 +1377,6 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//   500:
 	//     $ref: "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/libpod/containers/{name}/restore"), s.APIHandler(libpod.Restore)).Methods(http.MethodPost)
-
 	// swagger:operation GET /containers/{name}/changes libpod libpodChangesContainer
 	// swagger:operation GET /libpod/containers/{name}/changes compat changesContainer
 	// ---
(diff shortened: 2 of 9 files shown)


pkg/api/handlers/libpod/containers.go
```diff
@@ -285,3 +285,23 @@ func Restore(w http.ResponseWriter, r *http.Request) {
...
+	name := utils.GetName(r)
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	ctr, err := runtime.LookupContainer(name)
+	if err != nil {
+		utils.ContainerNotFound(w, name, err)
+		return
+	}
+	err = ctr.Init(r.Context())
+	if errors.Cause(err) == define.ErrCtrStateInvalid {
+		utils.Error(w, "container already initd", http.StatusNotModified, err)
```
jwhonce, line 299:
```suggestion
		utils.Error(w, "container already init'ed", http.StatusNotModified, err)
```
Seems better english.

pkg/api/server/register_containers.go
```diff
@@ -1411,6 +1410,29 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	r.HandleFunc(VersionedPath("/containers/{name}/changes"), s.APIHandler(compat.Changes))
 	r.HandleFunc("/containers/{name}/changes", s.APIHandler(compat.Changes))
 	r.HandleFunc(VersionedPath("/libpod/containers/{name}/changes"), s.APIHandler(compat.Changes))
-
+	// swagger:operation POST /libpod/containers/{name}/init libpod libpodInitContainer
+	// ---
+	// tags:
+	//  - containers
+	// summary: Initialize a container
+	// description: Performs all tasks necessary for starting the container but does not start the container.
```
jwhonce, line 1418:
```suggestion
	// description: Performs all tasks necessary for initializing the container but does not start the container.
```

pkg/domain/infra/abi/containers.go
```diff
@@ -794,3 +794,17 @@ func (ic *ContainerEngine) ContainerCleanup(ctx context.Context, namesOrIds []st
 	}
 	return reports, nil
 }
+
+func (ic *ContainerEngine) ContainerInit(ctx context.Context, namesOrIds []string, options entities.ContainerInitOptions) ([]*entities.ContainerInitReport, error) {
+	var reports []*entities.ContainerInitReport
+	ctrs, err := getContainersByContext(options.All, options.Latest, namesOrIds, ic.Libpod)
```
jwhonce, line 800:
```suggestion
	ctrs, err := getContainersByContext(options.All, options.Latest, namesOrIds, ic.Libpod)
```
Just querying the containers causes them to be initialized?  I suspect a C&P error here.  I am also confused how you are going to create the container if a "not found" error is going to return.

pkg/domain/infra/tunnel/containers.go
```diff
@@ -338,3 +338,19 @@ func (ic *ContainerEngine) ContainerDiff(ctx context.Context, nameOrId string, _
...
 }
+
+func (ic *ContainerEngine) ContainerInit(ctx context.Context, namesOrIds []string, options entities.ContainerInitOptions) ([]*entities.ContainerInitReport, error) {
+	var reports []*entities.ContainerInitReport
+	ctrs, err := getContainersByContext(ic.ClientCxt, options.All, namesOrIds)
+	if err != nil {
+		return nil, err
+	}
+	for _, ctr := range ctrs {
+		err := containers.ContainerInit(ic.ClientCxt, ctr.ID)
```
jwhonce, line 349:
```suggestion
		err := containers.ContainerInit(ic.ClientCxt, ctr.ID)
```
This looks better.

pkg/domain/infra/tunnel/containers.go
```diff
@@ -338,3 +338,19 @@ func (ic *ContainerEngine) ContainerDiff(ctx context.Context, nameOrId string, _
 func (ic *ContainerEngine) ContainerCleanup(ctx context.Context, namesOrIds []string, options entities.ContainerCleanupOptions) ([]*entities.ContainerCleanupReport, error) {
 	return nil, errors.New("not implemented")
 }
+
+func (ic *ContainerEngine) ContainerInit(ctx context.Context, namesOrIds []string, options entities.ContainerInitOptions) ([]*entities.ContainerInitReport, error) {
+	var reports []*entities.ContainerInitReport
+	ctrs, err := getContainersByContext(ic.ClientCxt, options.All, namesOrIds)
+	if err != nil {
```
jwhonce, line 345: Isn't container not found an error?

pkg/domain/infra/abi/containers.go
baude: This is correct.  If the container input is bad (i.e. not found), then it returns and does not proceed.  This is the exact behavor of v1.

pkg/domain/infra/tunnel/containers.go
baude: this behavior is same as v1, if input is bad, return before attempting any inits.

PR #5804, opened 2020-04-13
Need to set security options even if user does not specify options

rhatdan:

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -783,10 +783,12 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 		Sysctl:         sysctl,
 	}
 
+	var securityOpt []string
 	if c.Changed("security-opt") {
-		if err := secConfig.SetSecurityOpts(runtime, c.StringArray("security-opt")); err != nil {
-			return nil, err
-		}
+		securityOpt = c.StringArray("security-opt")
+	}
+	if err := secConfig.SetSecurityOpts(runtime, securityOpt); err != nil {
+		return nil, err
 	}
 
 	// SECCOMP


rhatdan:
@mheon @baude We are not applying seccomp.json rules unless the user specifies a security option.  This is needed for 1.9.0.

rhatdan:
@baude @mheon @giuseppe @vrothberg @TomSweeneyRedHat @QiWang19 PTAL

baude:
any way this can have a test so when it fails in v2 we can catch it?

rhatdan:
I think we should add this to podman info, then we could at least look for it.

rhatdan:
Only reason I caught this, was that I was playing with it to containers.conf running podman in container.

TomSweeneyRedHat:
So prior to you change, the StringArray was returning 'nil' instead of an empty array?  If so, that's subtle.  

The remote tstsaren't happy, I restarted the remote one, I don't think it's related.  A new test would also be good.

rhatdan:
The problem is the code fails over to the built in.  I guess we could check a syscall that we know is blocked by the default seccomp filters.

mount/unshare perhaps.

But for now lets get this merged, so we can get podman 1.9.0 released.

rhatdan:
@TomSweeneyRedHat Prior to this PR, the code was not setting the seccomp.json file to /usr/share/containers/containers.conf unless the user specified ANY --security-opt rule.  We want this set if the user did not override the seccomp file.

TomSweeneyRedHat:
@rhatdan got it, thanks.  I wasn't catching that `config.SetSecurityOpts()` was setting stuff to the default location if it got an empty array.

TomSweeneyRedHat:
late, but LGTM

PR #5807, opened 2020-04-14
Fix invalid container path comparison for pid cgroup

saschagrunert:
This fixes the behavior to return nil for the PIDs cgroup if the
container path is empty.

cc @giuseppe


rhatdan:
LGTM
@giuseppe @vrothberg @mheon @baude @QiWang19 @TomSweeneyRedHat PTAL

PR #5808, opened 2020-04-14
v2podman ps revert structure changes

baude:
reverting name changes to the listcontainer structure because it negatively impacted the direct consumption of the restful API.  instead we now use a local structure in the CLI to modify the output as needed.


mheon:
LGTM

pkg/ps/ps.go
```diff
@@ -158,11 +158,11 @@ func ListContainerBatch(rt *libpod.Runtime, ctr *libpod.Container, opts entities
...
-		ContainerNames: []string{conConfig.Name},
+		Names:          []string{conConfig.Name},
 		Pid:            pid,
 		Pod:            conConfig.Pod,
-		PortMappings:   conConfig.PortMappings,
-		ContainerSize:  size,
+		Ports:          conConfig.PortMappings,
+		Size:           size,
 		StartedAt:      startedTime.Unix(),
 		ContainerState: conState.String(),
```
marusak, line 167: Also `ContainerState` was `State` before.

test/apiv2/20-containers.at
```diff
@@ -21,8 +21,8 @@ t GET libpod/containers/json?all=true 200 \
   length=1 \
   .[0].Id~[0-9a-f]\\{12\\} \
   .[0].Image=$IMAGE \
-  .[0].Cmd[0]="true" \
-  .[0].ContainerState~\\\(exited\\\|stopped\\\) \
+  .[0].Command[0]="true" \
+  .[0].State~\\\(exited\\\|stopped\\\) \
```
marusak, line 25: This should catch it.

marusak:
Thanks, now it is usable as it was before. LGTM

PR #5813, opened 2020-04-14
More system test fixes on regressions

baude:
Fix more regressions between v1 and v2


cmd/podmanV2/images/rm.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/spf13/pflag"
+
```
rhatdan, line 8: Remove blank line.

PR #5814, opened 2020-04-14
v2specgen prune libpod

baude:
use libpod only in the specgen/generate package so that the remote clients do not inherit libpod bloat.


baude:
fwiw, this saved about 20% in remote client file size

pkg/specgen/generate/config_linux_cgo.go
```diff
@@ -1,11 +1,13 @@
...
 
-package specgen
+package generate
 
 import (
 	"context"
 	"io/ioutil"
 
+	"github.com/containers/libpod/pkg/specgen"
+
```
rhatdan, line 10: Delete line.

rhatdan: LGTM other then this NIT.

pkg/specgen/generate/config_linux_nocgo.go
```diff
@@ -0,0 +1,12 @@
...
+
+package generate
+
+import (
+	"github.com/containers/libpod/pkg/specgen"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func (s *specgen.SpecGenerator) getSeccompConfig(configSpec *spec.Spec) (*spec.LinuxSeccomp, error) {
+	return nil, nil
```
jwhonce, line 11: nil, nil seems odd

pkg/specgen/generate/oci.go
```diff
@@ -183,32 +184,32 @@ func (s *SpecGenerator) ToOCISpec(rt *libpod.Runtime, newImage *image.Image) (*s
 
 	// NAMESPACES
 
-	if err := s.pidConfigureGenerator(&g); err != nil {
+	if err := pidConfigureGenerator(&g, s); err != nil {
```
jwhonce, line 187:
```suggestion
	if err := pidConfigureGenerator(s, &g); err != nil {
```
A small nit since s is the receiver, I would like to see it first not last.

PR #5817, opened 2020-04-14
Fix up SELinux labeling

rhatdan:
SELinux label options processing fixes, should allow system tests to pass.

### DIFF
--- a/pkg/specgen/container_validate.go
+++ b/pkg/specgen/container_validate.go
@@ -68,18 +68,6 @@ func (s *SpecGenerator) Validate() error {
 	if len(s.CapAdd) > 0 && s.Privileged {
 		return exclusiveOptions("CapAdd", "privileged")
 	}
-	// selinuxprocesslabel and privileged are exclusive
-	if len(s.SelinuxProcessLabel) > 0 && s.Privileged {
-		return exclusiveOptions("SelinuxProcessLabel", "privileged")
-	}
-	// selinuxmounmtlabel and privileged are exclusive
-	if len(s.SelinuxMountLabel) > 0 && s.Privileged {
-		return exclusiveOptions("SelinuxMountLabel", "privileged")
-	}
-	// selinuxopts and privileged are exclusive
-	if len(s.SelinuxOpts) > 0 && s.Privileged {
-		return exclusiveOptions("SelinuxOpts", "privileged")
-	}
 	// apparmor and privileged are exclusive
 	if len(s.ApparmorProfile) > 0 && s.Privileged {
 		return exclusiveOptions("AppArmorProfile", "privileged")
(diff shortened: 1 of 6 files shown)


rhatdan:
@baude PTAL

baude:
LGTM

baude:
@mheon bc this has specgen changes, you should ptal

pkg/specgen/container_validate.go
```diff
@@ -68,18 +68,6 @@ func (s *SpecGenerator) Validate() error {
 	if len(s.CapAdd) > 0 && s.Privileged {
 		return exclusiveOptions("CapAdd", "privileged")
 	}
-	// selinuxprocesslabel and privileged are exclusive
```
mheon, line 71: Why are we turning these off? Privileged flag disables all manual SELinux config, last I checked

pkg/specgen/specgen.go
```diff
@@ -230,14 +230,6 @@ type ContainerSecurityConfig struct {
 	// If SELinux is enabled and this is not specified, a label will be
 	// automatically generated if not specified.
 	// Optional.
-	SelinuxProcessLabel string `json:"selinux_process_label,omitempty"`
```
mheon, line 233: Is it necessary to completely remove this?

pkg/specgen/container_validate.go
rhatdan: In the other code, we are setting the selinux options to "disabled" when --privileged mode is enabled.  I guess I could check over there.  This was causing an error to be thrown here.

pkg/specgen/specgen.go
rhatdan: It is not used.

mheon: Why? The original intent of SpecGen was to make things as explicit as possible, removing parsing where possible - it seems like we're regressing on this here?

rhatdan: We don't know at the time that the specgen is created what the labels are going to be.  The labels are creates in container/storage which happens after this is generate.

mheon: These were originally intended for user-specified labels from `--security-opt label=` but it doesn't look like those compile cleanly into a complete process/mount label without outside intervention

mheon:
Success failed, rerunning

LGTM otherwise

rhatdan:
@mheon That is not the way `--security-opt label=` works.
It does
```
man podman create
...
       --security-opt=option

       Security Options

              · label=user:USER     : Set the label user for the container processes

              · label=role:ROLE     : Set the label role for the container processes

              · label=type:TYPE     : Set the label process type for the container processes

              · label=level:LEVEL   : Set the label level for the container processes

              · label=filetype:TYPE : Set the label file type for the container files

              · label=disable       : Turn off label separation for the container
```

stevekuznetsov:
/hold
```
 Required status check \"success\" is cancelled
```

rhatdan:
@stevekuznetsov What is going on?

mheon:
@cevich Poke. Success task seems broken here. Rerunning not helping.

PR #5818, opened 2020-04-15
podmanV2: fix nil deref

vrothberg:
Fix a typo when looking up a flag causing a nil deref and all commands
to fail.


vrothberg:
@giuseppe @rhatdan we need to get this in asap. V2 is broken on master.

@baude @jwhonce FYI

PR #5830, opened 2020-04-15
podmanv2 fix runtime assignment

baude:
when provided by the command line, the runtime assignment needs to send the right attribute


rhatdan:
LGTM

Issue #5796, opened 2020-04-12
podman not opening port


supermar1010:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The port is supposed to be routed to the host, but is not. See the output of `podman ps`. Port 10080(it is in a different docker-compose/pod) works but 10000 and 18081(both in the same docker-compose/pod) does not, there's no error message or similiar.

Sometimes it works though, I'm not sure what triggers it that it works sometimes.

If I use a bash in the container the port 3000 is open and it returns stuff as its supposed to, so seems to be a mapping bug? In my container I'm listening on *:3000.

`lsof -i -P -n | grep LISTEN`:

`exe     2204  jc2    9u  IPv4  30349      0t0  TCP 127.0.0.1:10080 (LISTEN)`

`podman ps`:

```
[jc2@v220190910461598152 backend]$ podman ps
CONTAINER ID  IMAGE                                   COMMAND               CREATED         STATUS             PORTS              
                                 NAMES
0be12725dfb4  docker.io/library/mongo-express:latest  mongo-express         5 minutes ago   Up 5 minutes ago   127.0.0.1:10000->3000/tcp, 0.0.0.0:18081->8081/tcp  backend_mongo-express_1
6596ecacfe98  docker.io/library/mongo:latest          mongod                5 minutes ago   Up 5 minutes ago   127.0.0.1:10000->3000/tcp, 0.0.0.0:18081->8081/tcp  mongoJC2
c282bfe1789a  docker.io/library/node:latest           bash -c cp /usr/s...  5 minutes ago   Up 5 minutes ago   127.0.0.1:10000->3000/tcp, 0.0.0.0:18081->8081/tcp  backend_node_1
94a666479072  docker.io/library/nginx:alpine          nginx -g daemon o...  40 minutes ago  Up 40 minutes ago  127.0.0.1:10080->80/tcp                             nginx_web_1
```

**Steps to reproduce the issue:**

1. For me start the corresponding docker file

**Describe the results you received:**

Port is not available/open

**Describe the results you expected:**

Port is opened and listening

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:            1.8.2
RemoteAPI Version:  1
Go Version:         go1.13.6
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.6
[...]

PR #5839, opened 2020-04-16
Add SELinux volume information to troubleshoot.md

rhatdan:
This issue happens enough that we need to put it on
the top of the troubleshoot guide.


rhatdan:
@aanno PTAL

aanno:
I was able to circumvent the problem with

```
$ chcon -Rt svirt_sandbox_file_t /mnt/home/aanno/scm/scm/docker-compose-files/config/solr
```

see #5796. Perhaps it would be useful to mention that as well.

rhatdan:
@aanno That is actually not the preferred way.  The issue with what you did is you set the labeling to be shared amoungst containers.  So if one container broke out , it would be allowed to read/write your content.  If you use the :Z method, it would be private to your container.  chcon container_file_t ... is equivalent to :z.

troubleshooting.md
```diff
@@ -16,7 +16,37 @@ If they differ, please update your version of PODMAN to the latest possible
...
+objects on the shared volumes. The **z** option tells Podman that two containers
+share the volume content. As a result, Podman labels the content with a shared
+content label. Shared volume labels allow all containers to read/write content.
+The **Z** option tells Podman to label the content with a private unshared label.
+Only the current container can use a private volume.
+
+$ podman run -v ~/mycontent:/content:Z fedora touch /content/file
+
+Make sure the content is private for the container.  Do not relabel system directories.
+You can also disable SELinux separation for the container using `--security-opt label=disabled`
```
mheon, line 44: I think it's `label=disable` not `disabled`

mheon: Also, suggest saying that this can be used for the system directories that can't be relabeled?

TomSweeneyRedHat:
other than @mheon's comment LGTM

troubleshooting.md
rhatdan: Fixed

PR #5842, opened 2020-04-16
podman v2 remove bloat v2

baude:
rid ourseleves of libpod references in v2 client


libpod/define/config.go
```diff
@@ -57,3 +57,12 @@ type AttachStreams struct {
 	// If false, stdout will not be attached
 	AttachInput bool
 }
+
+// JournaldLogging is the string conmon expects to specify journald logging
+const JournaldLogging = "journald"
```
rhatdan, line 62: These need to be moved to containers.conf, but fine for now.

PR #5851, opened 2020-04-16
Provide a json variable pointing to a configured json API

jwhonce:
* All commands now using the same instance of json API
* `json` variable created in each package to prevent `encoding/json`
  from being re-introduced

### DIFF
--- a/cmd/podman/common/default.go
+++ b/cmd/podman/common/default.go
@@ -1,5 +1,9 @@
 package common
 
+import (
+	"github.com/containers/libpod/cmd/podman/registry"
+)
+
 var (
 	// DefaultHealthCheckInterval default value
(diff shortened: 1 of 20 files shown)


jwhonce:
@mheon I believe this is what you were asking for earlier.  I'll remove [WIP] if this is correct.

cmd/podmanV2/registry/config.go
```diff
@@ -20,8 +21,12 @@ const (
 
 var (
 	PodmanOptions entities.PodmanConfig
+	Json          = jsoniter.ConfigCompatibleWithStandardLibrary
 )
 
+func init() {
```
mheon, line 27: Can probably trim this

mheon:
LGTM. This is, indeed, what I was asking for.

PR #5853, opened 2020-04-16
Unify in_podman container packaging & VM packaging

cevich:
***Depends on optional:*** #5737

Significant changes in this PR:

* Separate out package update + install from the VM Image setup process, into dedicated scripts.
* Reuse VM Image packaging scripts, for `in_podman` container image (Dockerfile + Dockerfile.ubuntu).
* Add test-builds of the container images to the VM Image build process, to provide feedback upon future VM Image setup changes.

Note: Both the VM Images and in_podman container image are utilized outside of the libpod repository.  For example,


contrib/cirrus/lib.sh
```diff
@@ -449,63 +448,43 @@ systemd_banish() {
...
-    sudo rm -rf /tmp/.??*
-    sudo sync
-    sudo fstrim -av
+    $SUDO rm -rf /var/lib/cloud/instanc*
+    $SUDO rm -rf /root/.ssh/*
+    $SUDO rm -rf /etc/ssh/*key*
+    $SUDO rm -rf /etc/ssh/moduli
+    $SUDO rm -rf /home/*
+    $SUDO rm -rf /tmp/*
+    $SUDO rm -rf /tmp/.??*
```
jwhonce, line 468: -f here are ignoring errors.  `--force` is a really bad name for this flag. Is that good enough for these builds?  I get nervous when I see /etc files removed this way.

contrib/cirrus/lib.sh
```diff
@@ -449,63 +448,43 @@ systemd_banish() {
...
+    $SUDO touch /.unconfigured  # force firstboot to run
     _finalize
 }
 
 ubuntu_finalize() {
     set +e  # Don't fail at the very end
     echo "Resetting to fresh-state for usage as cloud-image."
     $LILTO $SUDOAPTGET autoremove
-    sudo rm -rf /var/cache/apt
+    $SUDO rm -rf /var/cache/apt
```
jwhonce, line 488:
```suggestion
    $SUDO rm -r /var/cache/apt
```
If you expect the cache to really be empty drop the -f.

contrib/cirrus/packer/fedora_packaging.sh
```diff
@@ -0,0 +1,139 @@
...
+        REMOVE_PACKAGES+=(runc)
+        ;;
+    *)
+        bad_os_id_ver ;;
+esac
+
+echo "Installing general build/test dependencies for Fedora '$OS_RELEASE_VER'"
+$BIGTO ooe.sh $SUDO dnf install -y ${INSTALL_PACKAGES[@]}
+
+[[ "${#REMOVE_PACKAGES[@]}" -eq "0" ]] || \
```
jwhonce, line 134:
```suggestion
[[ ${#REMOVE_PACKAGES[@]} -eq 0 ]] || \
```
Quotes not needed between `[[]]`

contrib/cirrus/packer/fedora_setup.sh
```diff
@@ -6,130 +6,29 @@
...
-install_ooe
-
-export GOPATH="$(mktemp -d)"
-trap "sudo rm -rf $GOPATH" EXIT
-
-$BIGTO ooe.sh sudo dnf update -y
+if [[ $OS_RELEASE_VER -le 31 ]]; then
+    warn "Switching io scheduler to 'deadline' to avoid RHBZ 1767539"
+    warn "aka https://bugzilla.kernel.org/show_bug.cgi?id=205447"
+    echo "mq-deadline" | sudo tee /sys/block/sda/queue/scheduler > /dev/null
```
jwhonce, line 16: Interesting...

contrib/cirrus/setup_environment.sh
```diff
@@ -47,10 +47,12 @@ case "${OS_RELEASE_ID}" in
...
         fi
 
-        warn "Switching io scheduler to 'deadline' to avoid RHBZ 1767539"
-        warn "aka https://bugzilla.kernel.org/show_bug.cgi?id=205447"
-        echo "mq-deadline" > /sys/block/sda/queue/scheduler
-        cat /sys/block/sda/queue/scheduler
+        if [[ $OS_RELEASE_VER -le 31 ]]; then
+            warn "Switching io scheduler to 'deadline' to avoid RHBZ 1767539"
+            warn "aka https://bugzilla.kernel.org/show_bug.cgi?id=205447"
+            echo "mq-deadline" > /sys/block/sda/queue/scheduler
```
jwhonce, line 53:
```suggestion
            echo "mq-deadline" > /sys/block/sda/queue/scheduler
```
Why not use:
`echo "mq-deadline" | sudo tee /sys/block/sda/queue/scheduler > /dev/null
`
from other script?

contrib/cirrus/lib.sh
cevich: In this case I think it's okay, and we **DO** want to ignore missing file/directory errors.  But if I'm overlooking something obvious to you, please let me know.

 This function runs as the very final step when building a Ubuntu or Fedora VM image, to "reset" them for a new "first" use (on next boot).  Not all these files exist on all distros. and I was hesitant to complicate things with lots of conditionals.  In any case, it's not a function for general use during testing.  Perhaps I should comment that?

My other assumption here is, the new VM images will shortly be put to use, so if there is a problem, it will soon get noticed.

cevich: Okay, I added comments to these functions.

contrib/cirrus/packer/fedora_packaging.sh
cevich: Got it, thx.

contrib/cirrus/lib.sh
cevich: Not sure I follow the suggestion.  This is another VM Image build-time thing where it's probably okay to ignore a 'does not exist' error.  Am I missing something?

contrib/cirrus/packer/fedora_setup.sh
cevich: sigh...and apparently this is also an issue in F32 https://bugzilla.redhat.com/show_bug.cgi?id=1768092 with a 5.6 kernel :cry:   Fixing...

contrib/cirrus/setup_environment.sh
cevich: Oh durrr, ya I should make a common function for this...okay fixed, thanks!

cevich:
@baude do you mind taking a quick peek at the not-required but running tests, just to verify they appear to be operating as expected?  In other words, I'm not sure what tests are expected to fail and pass or error.

cevich:
@baude @mheon ping. Can we get this PR merged soon-ish?  I've got another followup PR rolling, to update us to F32 proper and fix an SELinux-related SNAFU affecting buildah..

TomSweeneyRedHat:
It would be nice to rename all the "Dockerfile.*" files to "Containerfile.*" if possible.

cevich:
@TomSweeneyRedHat I agree, but also/first we need to convince all the docker-centric build services (like quay.io) to recognize it too :S

(I'm assuming it doesn't, maybe I'm wrong?)

TomSweeneyRedHat:
@cevich, that's the caveat.  Many (all?) of the docker-centric build services don't recognize a Containerfile, I know for a fact that quay.io does not, but I hope to at least change that.

So if there's some reliance on the file being named "Dockerfile" so it can be found by some system expecting that name, then I'd leave it as is.  I'm guessing, however, that the Dockerfile.{extension} files can be changed to "Containerfile.{extension}" as they are not likely used by a build system.

PR #5856, opened 2020-04-16
Fix bug where two configurations had been created

jwhonce:
* registry.PodmanConfig() now returns a pointer to the source of truth


baude:
LGTM

PR #5862, opened 2020-04-17
Fix up handling of image data as well as setting Linux Capabilties correctly

rhatdan:


rhatdan:
@mheon @baude @jwhonce This will fix a bunch of failures, since we are not getting data out of the images.  Also should fix test/e2e/run_security_labels.go tests.

libpod/image/image.go
```diff
@@ -847,22 +847,72 @@ func (i *Image) Dangling() bool {
...
+	return imgInspect.Env
+}
+
 // Labels returns the image's labels
-func (i *Image) Labels(ctx context.Context) (map[string]string, error) {
+func (i *Image) Labels(ctx context.Context) map[string]string {
 	imgInspect, err := i.imageInspectInfo(ctx)
 	if err != nil {
-		return nil, nil
+		return nil
```
mheon, line 908: Why are we just discarding errors? That doesn't seem sane

rhatdan: The previous code was ignoring the errors. I thought there was a reason.

mheon: Probably not? Can't think of one

rhatdan: Fixed

rhatdan:
@baude @mheon @jwhonce @TomSweeneyRedHat @QiWang19 @giuseppe Lets get this in so we can pass more tests.

TomSweeneyRedHat:
LGTM

rhatdan:
Also fixes run_seccomp.go e2e test.

rhatdan:
run_selinux_test.go also passes.

PR #5863, opened 2020-04-17
podman rmi: refactor logic

### LINKED ISSUE #5847: podmanV2 rmi issues
`podman rmi`in v1 is trying to remove _all_ specified images and records the errors while v2 errors out immediately on the first error.

vrothberg:
While this commit was initially meant to fix #5847, it has turned into a
bigger refactoring which I did not manage to break into smaller pieces:

 * Fix #5847 by refactoring the image-removal logic.

 * Make the api handler for image-removal use the ABI code. This way,
   both (i.e., ABI and Tunnel) end up using the same code.  Achieving
   this code share required to move some code around to prevent circular
   dependencies.

 * Everything in pkg/api (excluding pkg/api/types) must now only be
   accessed from code using `ABISupport`.

 * Avoid imports from entities on handlers to prevent circular
   dependencies.

 * Move `podman system service` logic into `cmd` to prevent circular
   dependencies - it depends on pkg/api.

Fixes: #5847

### DIFF
--- a/cmd/podman/system/service.go
+++ b/cmd/podman/system/service.go
@@ -90,7 +90,8 @@ func service(cmd *cobra.Command, args []string) error {
 
 	logrus.Warn("This function is EXPERIMENTAL")
 	fmt.Fprintf(os.Stderr, "This function is EXPERIMENTAL.\n")
-	return registry.ContainerEngine().RestService(registry.GetContext(), opts)
+
+	return restService(opts, cmd.Flags(), registry.PodmanConfig())
 }
 
 func resolveApiURI(_url []string) (string, error) {
--- a/pkg/domain/infra/abi/system.go
+++ b/pkg/domain/infra/abi/system.go
@@ -1,20 +1,15 @@
-// +build ABISupport
-
 package abi
 
 import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"net"
 	"os"
 	"strconv"
-	"strings"
 	"syscall"
 
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/libpod/libpod/define"
-	api "github.com/containers/libpod/pkg/api/server"
 	"github.com/containers/libpod/pkg/cgroups"
 	"github.com/containers/libpod/pkg/domain/entities"
 	"github.com/containers/libpod/pkg/rootless"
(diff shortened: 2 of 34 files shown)


vrothberg:
I did not test the remote client yet but feel free to review.

TomSweeneyRedHat:
@vrothberg gating isn't hip

vrothberg:
Quick meeting record: we agreed to generalize the consolidation/code-share issue in the PR and make pkg/api/hanlders use the infra/abi code.

jwhonce:
These are examples of where the CLI instantiates the runtime engines. 

PodmanConfig might be too generous of input, we may need to narrow the data being passed in.
https://github.com/containers/libpod/blob/master/cmd/podman/registry/registry.go#L50

cmd/podman/system/service.go
```diff
@@ -71,7 +72,21 @@ func service(cmd *cobra.Command, args []string) error {
...
-	return registry.ContainerEngine().RestService(registry.GetContext(), opts)
+
+	// Note that the `system service` command is a bit of an outlier. To
+	// avoid circular dependencies and keep the code clean, `entities` is
+	// not allowed to import `pkg/api` and any sub-packages.  To continue
+	// allowing for a dynamic dispatch and support a parallel ABI & tunnel
+	// support in a single binary, we're doing a dispatch here.
+	podConf := registry.PodmanConfig()
+	switch podConf.EngineMode {
+	case entities.TunnelMode:
```
jwhonce, line 83: FYI -- `Mode:    []entities.EngineMode{entities.ABIMode}` when registering command implies this case will/should never be called.  If you get here "bad" things are happening.

pkg/domain/infra/abi/system.go
```diff
@@ -33,42 +30,6 @@ func (ic *ContainerEngine) Info(ctx context.Context) (*define.Info, error) {
...
-	err = server.Serve()
-	if listener != nil {
-		_ = (*listener).Close()
-	}
-	return err
-}
-
 func (ic *ContainerEngine) VarlinkService(_ context.Context, opts entities.ServiceOptions) error {
 	var varlinkInterfaces = []*iopodman.VarlinkInterface{
 		iopodmanAPI.New(opts.Command, ic.Libpod),
```
jwhonce, line 35:
```suggestion
func (ic *ContainerEngine) VarlinkService(_ context.Context, opts entities.ServiceOptions) error {
	var varlinkInterfaces = []*iopodman.VarlinkInterface{
		iopodmanAPI.New(opts.Command, ic.Libpod),
```
Should this move into cmd/podman as well?

vrothberg: Yes, this would be another nice candidate. For now, this PR focuses on `podman rmi`.

cmd/podman/system/service.go
vrothberg: Great catch, thanks!

vrothberg:
swagger failures are fixed now. @jwhonce @baude, the issue I faced is that swagger went south because pkg/api now imports infra/abi. infra/abi had the build header set to ABISupport which in turn confused swagger because it couldn't parse it anymore. I just removed the headers as the imports are guarded in infra. There's now way to point swagger to some build tags and go doesn't support setting them via environment variables either.

Should be good now :+1:

baude:
ok LGTM

vrothberg:
@rhatdan @giuseppe PTAL

PR #5866, opened 2020-04-17
Log formatter: add BATS summary line

edsantiago:
BATS emits a summary line (number of tests passed/failed)...
but only on a tty or when run with --pretty! In our CI
context, with TAP output, it gives no end summary.

Fix that. Keep track of 'ok', 'not ok', and 'skipped',
and display the counts at the end.

Also: add a regression test. You don't need to review
or even read it: it's stark, and I'm not even enabling
it for CI because it almost certainly won't run due to
missing Perl library modules. It's just something I
need on my end.

### DIFF
--- a/contrib/cirrus/logformatter
+++ b/contrib/cirrus/logformatter
@@ -52,11 +52,13 @@ a.codelink:hover   { background: #000; color: #999; }
 a.timing           { text-decoration: none; }
 
 /* BATS styles */
-.bats-ok        { color: #393; }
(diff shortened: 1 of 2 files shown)


edsantiago:
@baude with no CI running I can't offer you an example of what it looks like, sorry. It'll be something like:
```
Summary:  2 Passed, 1 Failed, 1 Skipped. Total tests: 4
```

baude:
LGTM

PR #5867, opened 2020-04-17
engine mode: make it static

vrothberg:
Make the engine mode static and do not make it depend on the runtime OS
or any other runtime data.  The remote support is enabled with a static
build flag, so we can't flip it on-demand.


vrothberg:
Closing. We want to support both in the Linux client.

PR #5875, opened 2020-04-18
Cleanup network option parsing

rhatdan:
We were not handling the parsing of --ip.  This pr adds validation
checks and now will support the flag.

Move validation to the actual parsing of the network flags.

We should only parse the dns flags if the user changed them. We don't
want to pass default options if set in containers.conf to the server.
Potential for duplicating defaults.

Add support for --dns-opt flag passing

Begin handling of --network flag, although we don't have a way right now
to translate a string into a specgen.Namespace.

### DIFF
--- a/cmd/podman/common/createparse.go
+++ b/cmd/podman/common/createparse.go
@@ -1,7 +1,6 @@
 package common
 
 import (
-	"github.com/containers/libpod/cmd/podman/parse"
 	"github.com/containers/libpod/pkg/util"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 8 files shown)


rhatdan:
@mheon @jwhonce @baude How are we supposed to convert from a string to a specgen.Namespace?

rhatdan:
Once https://github.com/containers/libpod/pull/5862 is merged this PR should be able to pass
run_staticips_test.go

rhatdan:
@baude @vrothberg @giuseppe @TomSweeneyRedHat @mheon @jwhonce @QiWang19 PTAL

PR #5879, opened 2020-04-19
Pull images when doing podman create

rhatdan:
Currently podman create does not pull images if they don't
exist.  It should follow the pull policy specified by the
user.

### DIFF
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -3,6 +3,7 @@ package containers
 import (
 	"fmt"
 
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/libpod/cmd/podman/common"
 	"github.com/containers/libpod/cmd/podman/registry"
 	"github.com/containers/libpod/pkg/domain/entities"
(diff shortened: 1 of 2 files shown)


baude:
what is the default pull policy

rhatdan:
@baude "missing"

rhatdan:
@mheon @vrothberg @giuseppe @TomSweeneyRedHat @jwhonce @QiWang19 PTAL

mheon:
LGTM

cmd/podman/containers/create.go
```diff
@@ -100,3 +106,27 @@ func createInit(c *cobra.Command) error {
...
+		_, pullErr := registry.ImageEngine().Pull(registry.GetContext(), imageName, entities.ImagePullOptions{
+			Authfile: cliVals.Authfile,
+			Quiet:    cliVals.Quiet,
+		})
+		if pullErr != nil {
+			return pullErr
+		}
+	}
+	return nil
+}
```
TomSweeneyRedHat, line 132: This looks like it pulls the image even if `--pull` is not specified?  Is that correct behavior?

Also the man page fo create only lists `--pull` and not `--pull-always` or `--pull-never` should that change?

baude: docker behavior, fwiw, is to NOT pull on create and require it to be local.

rhatdan: podman v1.9 pulls automatically.

pull-always and --pull-never only exists for buildah.  Not for create.

PR #5880, opened 2020-04-19
Make `find` ignore dot files

abitrolly:
There is no need to search for sources in hidden dirs. In my case
there are some development environment files that stand in the way.


rhatdan:
LGTM
@abitrolly You need to sign your PRs/

abitrolly:
@rhatdan thanks for the reminder. Fixed.

In ideal world it should be `gateway` task who adds the comment. )

PR #5896, opened 2020-04-20
common: setting cgroup resources correctly

giuseppe:
fix some cgroups related integration tests.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -47,6 +47,12 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cmd/podman/common/specgen.go
```diff
@@ -47,6 +47,12 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	if err != nil {
 		return err
 	}
+	if s.ResourceLimits == nil {
```
mheon, line 50: Do we need logic to clear this for the rootless v1 case?

giuseppe: we need to error out when some limits are specified.

So we need to carry the limits (avoiding to set any default) and raise an error from libpod if any limit is specified but we are rootless on cgroup v1

PR #5899, opened 2020-04-20
V2 podman image tree

jwhonce:
* Basic port of V1 podman image tree ID

TODO: Refactor to return tree from service and format in presentation
layer
TODO: Support tunneling mode


baude:
lgtm

PR #5900, opened 2020-04-20
podmanv2 cp

baude:
enable podman to copy files between container and host, local mode only. this is a straight port of v1 code to v2.

### DIFF
--- a/cmd/podman/containers/cp.go
+++ b/cmd/podman/containers/cp.go
@@ -1,0 +1,55 @@
+package containers
+
+import (
+	"github.com/containers/libpod/cmd/podman/registry"
+	"github.com/containers/libpod/pkg/cgroups"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/libpod/pkg/rootless"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cpDescription = `Command copies the contents of SRC_PATH to the DEST_PATH.
+
+  You can copy from the container's file system to the local machine or the reverse, from the local filesystem to the container. If "-" is specified for either the SRC_PATH or DEST_PATH, you can also stream a tar archive from STDIN or to STDOUT. The CONTAINER can be a running or stopped container.  The SRC_PATH or DEST_PATH can be a file or directory.
+`
+	cpCommand = &cobra.Command{
+		Use:     "cp [flags] SRC_PATH DEST_PATH",
+		Short:   "Copy files/folders between a container and the local filesystem",
+		Long:    cpDescription,
+		Args:    cobra.ExactArgs(2),
+		RunE:    cp,
+		Example: "podman cp [CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
+	}
+)
+
+var (
+	cpOpts entities.ContainerCpOptions
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode},
+		Command: cpCommand,
+	})
+	flags := cpCommand.Flags()
+	flags.BoolVar(&cpOpts.Extract, "extract", false, "Extract the tar file into the destination directory.")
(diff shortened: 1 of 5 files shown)


cmd/podman/containers/cp.go
```diff
@@ -0,0 +1,57 @@
...
+
+  You can copy from the container's file system to the local machine or the reverse, from the local filesystem to the container. If "-" is specified for either the SRC_PATH or DEST_PATH, you can also stream a tar archive from STDIN or to STDOUT. The CONTAINER can be a running or stopped container.  The SRC_PATH or DEST_PATH can be a file or directory.
+`
+	cpCommand = &cobra.Command{
+		Use:     "cp [flags] SRC_PATH DEST_PATH",
+		Short:   "Copy files/folders between a container and the local filesystem",
+		Long:    cpDescription,
+		Args:    cobra.ExactArgs(2),
+		RunE:    cp,
+		Example: "[CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
```
jwhonce, line 23:
```suggestion
		Example: "podman container cp [CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
```

pkg/domain/infra/abi/cp.go
```diff
@@ -0,0 +1,435 @@
...
+	"github.com/containers/storage/pkg/idtools"
+	securejoin "github.com/cyphar/filepath-securejoin"
+	"github.com/docker/docker/pkg/archive"
+	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func (ic *ContainerEngine) ContainerCp(ctx context.Context, options entities.ContainerCpOptions) (*entities.ContainerCpReport, error) {
```
jwhonce, line 30:
```suggestion
func (ic *ContainerEngine) ContainerCp(ctx context.Context, source, destination string, options entities.ContainerCpOptions) (*entities.ContainerCpReport, error) {
```
Would be a signature more in-line with other methods

pkg/domain/infra/abi/cp.go
```diff
@@ -0,0 +1,435 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "err getting current working directory")
+		}
+		destPath = filepath.Join(dir, destPath)
+	}
+
+	if options.Source == "-" {
+		srcPath = os.Stdin.Name()
+		extract = true
+	}
```
jwhonce, line 168: If this is moved into the presentation layer is that easier for tunneling support?

pkg/domain/infra/abi/cp.go
```diff
@@ -0,0 +1,435 @@
...
+
+	_, err = io.Copy(os.Stdout, file)
+	if err != nil {
+		return errors.Wrapf(err, "error streaming file to Stdout")
+	}
+	return nil
+}
+
+func isVolumeDestName(path string, ctr *libpod.Container) (bool, string, string) {
+	separator := string(os.PathSeparator)
```
jwhonce, line 372:
```suggestion
	separator := string(os.PathSeparator)
```
This could be an issue as path is coming from a client OS where os.PathSeparator may be different than the service OS.

pkg/domain/infra/abi/cp.go
baude: long term probably, but just working the abi angle right now

pkg/domain/infra/abi/cp.go
baude: agree, long term, but this is a simple port to get us moving

rhatdan:
LGTM

pkg/domain/infra/abi/cp.go
```diff
@@ -0,0 +1,435 @@
+// +build ABISupport
```
vrothberg, line 1: Can we remove this? That's in a race with https://github.com/containers/libpod/pull/5863#issuecomment-616710773 and can lead to failing gating tests with #5863. Swagger doesn't support build tags and goes south in certain cases.

PR #5905, opened 2020-04-20
Modify namespace generation code for specgen

mheon:
Namespaces have now been changed to properly handle all cases. Spec handling code for namespaces was consolidated in a single function.

Still missing:
- Image ports
- Pod namespaces likely still broken in Podmanv2 (we need to leave namespaces defaulted in the specgen unless we have a good reason not to)

### DIFF
--- a/pkg/specgen/generate/namespaces.go
+++ b/pkg/specgen/generate/namespaces.go
@@ -2,314 +2,389 @@ package generate
 
 import (
 	"os"
+	"strings"
 
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/libpod/define"
+	"github.com/containers/libpod/pkg/cgroups"
+	"github.com/containers/libpod/pkg/rootless"
 	"github.com/containers/libpod/pkg/specgen"
-	"github.com/cri-o/ocicni/pkg/ocicni"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/specgen/generate/namespaces.go
```diff
@@ -2,293 +2,421 @@ package generate
...
-	options := make([]libpod.CtrCreateOption, 0)
-
-	// Cgroups
-	switch {
-	case s.CgroupNS.IsPrivate():
-		ns := s.CgroupNS.Value
-		if _, err := os.Stat(ns); err != nil {
-			return nil, err
+// Get the default namespace mode for any given namespace type.
+// TODO: Do we want to wire this into containers.conf? I lean towards no, that's
```
rhatdan, line 21: There are calls to do this in containers.conf now.

mheon: Containers.conf doesn't really parse directly into namespaces at this level, which is why I'm hesitant.

jwhonce:
LGTM

pkg/specgen/generate/namespaces.go
rhatdan: Users want to to be able to disable namespaces.  Something like this, for running podman inside of a container.
```
# Default way to to create a cgroup namespace for the container
# Options are:
# `private` Create private Cgroup Namespace for the container.
# `host`    Share host Cgroup Namespace with the container.
#
cgroupns = "host"
```

mheon: Which will happen on the client side and will hand off to us a properly-formatted namespace. I don't want this code to have to parse strings.

mheon:
I'm going to add containers.conf to defaults.

mheon:
We now respect containers.conf settings for default namespaces

PR #5910, opened 2020-04-21
Move Fedora 31 dependencies for building podman into separate file

abitrolly:
I had hard time trying to figure out what list is really actual. Might need more comments to mark the bare minimum required to get `podman` binary and packages that are needed for tests etc. I doubt that `nmap-ncat` is really a build dependency.


vrothberg:
The linter fails though: Fix End of Files.........................................................Failed

I assume it's the empty line in the new file.

abitrolly:
Forgot to sign as well.

contrib/dependencies.txt
```diff
@@ -0,0 +1,34 @@
+# Fedora 31 dependencies for building podman
```
rhatdan, line 1: I would not specify the Fedora 31, this will just be a hassle in the future.  Change this to just Fedora

abitrolly: Fixed. )

cevich:
> I doubt that nmap-ncat is really a build dependency.

FYI- it's not.  There are some packages which are dependencies for testing things in general.  That specific one I believe was added so `contrib/cirrus/networking.sh` could do it's job.

PR #5911, opened 2020-04-21
podman: do not set empty cgroup limit blocks

giuseppe:
refactor cgroup limits in their own function.  If there are no limits set avoid to set the block.  Basic rootless containers work now.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -23,6 +23,173 @@ import (
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM, but needs a rebase.

giuseppe:
rebased

rhatdan:
@mheon PTAL

mheon:
LGTM

rhatdan:
@mheon @giuseppe Does this need to be cherrypicked?

giuseppe:
> @mheon @giuseppe Does this need to be cherrypicked?

We don't need to backport it as Podman 1.9 has already the same logic

PR #5916, opened 2020-04-21
v2, podman: plug --userns=auto

giuseppe:
--userns=auto tests pass now.


pkg/domain/infra/runtime_libpod.go
```diff
@@ -234,6 +234,18 @@ func ParseIDMapping(mode namespaces.UsernsMode, uidMapSlice, gidMapSlice []strin
 		HostGIDMapping: true,
 	}
 
+	if mode.IsAuto() {
+		var err error
+		options.HostUIDMapping = false
+		options.HostGIDMapping = false
+		options.AutoUserNs = true
```
mheon, line 241: Is this in Specgen? I don't think we have an Auto mode for namespaces yet...

PR #5917, opened 2020-04-21
v2: implement log{in,out}

vrothberg:
Implement podman login and logout.  Smoke tests were successful but the
system tests are currently failing as we seem unable to run a registry
at the moment.

### DIFF
--- a/cmd/podman/login.go
+++ b/cmd/podman/login.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"context"
+	"os"
+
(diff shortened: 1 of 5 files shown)


vrothberg:
> does the commit miss `go.mod` or how were the files vendored?

c/common is already vendored but nothing was referencing until now. Go only vendors what's really needed (or in the import list).

vrothberg:

vrothberg:
No idea why gating doesn't return. Repushing.

PR #5918, opened 2020-04-21
v2, podman: propagate --network settings

giuseppe:
it fixes tests using --net=host

### DIFF
--- a/cmd/podman/common/netflags.go
+++ b/cmd/podman/common/netflags.go
@@ -2,9 +2,11 @@ package common
 
 import (
 	"net"
+	"strings"
 
 	"github.com/containers/libpod/cmd/podman/parse"
 	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/containers/libpod/pkg/specgen"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


rhatdan:
I think this should happen on the server side not in the client.

mheon:
Client-side sounds right, but we should try and leave the namespace defaulted whenever possible so the server can autoselect slirp for rootless, and pod mode if in a pod.

giuseppe:
should we drop `ContainerCLIOpts.Net` and just pass down strings?

mheon:
We can definitely parse from string in the client...

Maybe we should move `containers.conf` parsing into the server, and only set network-related stuff in SpecGen if the user explicitly set it. That might be better than what we're doing.

rhatdan:
The problem with doing that is the podman --help values start being incorrect.

mheon:
I think we still display the default in the client code, but we only forward to the user if the flag was set - if it's the default we don't bother

rhatdan:
@mheon Does your PR replace the need for this PR?

mheon:
I think it does

giuseppe:
> I think it does

How so?  I don't see any namespace setting from the CLI being honored.  It is also introducing other regressions (e.g. `--userns=auto` is not working again).

giuseppe:
I'll work on the user namespace part and sync with @mheon once it is ready

PR #5919, opened 2020-04-21
Lots more fixes for V2 conversion.

rhatdan:


cmd/podman/common/netflags.go
```diff
@@ -159,9 +164,7 @@ func NetFlagsToNetOptions(cmd *cobra.Command) (*entities.NetOptions, error) {
 			return nil, err
 		}
 
-		return nil, errors.Errorf("network %s is not yet supported", network)
-		// TODO How do I convert a string network to a Specgen.Namespace?
-		//		opts.Network = specgen.Namespace{NSMode: network}
+		opts.Network = specgen.NewNamespace(network)
```
mheon, line 167: This is really unlikely to work - `--net CNI1,CNI2,CNI3` will definitely not work. We need to parse this to check for the common ones (`host`, `bridge`, `slirp4netns`, `ns:$PATH` (which needs to compile into `NSMode: Path, Value: $PATH`) and then case anything else and populate the specgen's CNINetworks field with it

pkg/specgen/namespaces.go
```diff
@@ -96,3 +98,16 @@ func (n *Namespace) validate() error {
 	}
 	return nil
 }
+
+func NewNamespace(path string) Namespace {
```
mheon, line 102: This probably works for Host, but won't work for `--cgroupns=container` (needs to turn into `Private`) or `--net=ns:$PATH` (needs to turn into `path:$PATH`)

rhatdan: Thanks.

rhatdan: Ok I moved the namespace stuff to a separate PR.

rhatdan:
@baude @giuseppe @vrothberg @mheon @TomSweeneyRedHat @QiWang19 @jwhonce PTAL

cmd/podman/common/create_opts.go
```diff
@@ -11,7 +14,7 @@ type ContainerCLIOpts struct {
 	CapAdd            []string
 	CapDrop           []string
 	CGroupsNS         string
-	CGroups           string
+	CGroupsMode       string
```
baude, line 17: this variable name cannot be changed. it correlates to the create.go line 58 and is the pointer endpoint for where the cgroups user input is stored.

cmd/podman/common/create_opts.go
```diff
@@ -25,7 +28,7 @@ type ContainerCLIOpts struct {
 	CPUSetMems        string
 	Detach            bool
 	DetachKeys        string
-	Device            []string
+	Devices           []specs.LinuxDevice
```
baude, line 31: You cannot change the variable name or type in this structure as they need to match the createflags. cobra is not aware of a LinuxDevice.

cmd/podman/containers/create.go
```diff
@@ -154,3 +159,18 @@ func pullImage(imageName string) error {
 	}
 	return nil
 }
+
+func convertDevices(cmd *cobra.Command) ([]specs.LinuxDevice, error) {
```
baude, line 163: i believe this has to be done in the specgen/generate of container create to account for actually being on the server?

baude: or is the path sufficient? if so, then this probably belongs in the area where the specgen is created based on the []string list of devices.

rhatdan: There is no verification going on here.  But specgen insists on this list being a []specs.LinuxDevice.

cmd/podman/common/create_opts.go
rhatdan: Where is line 58?

cmd/podman/common/create_opts.go
rhatdan: I am not using this value in Cobra.

rhatdan: I could just remove it, and not use it to pass this to specgen  But then I need to add a function.

cmd/podman/common/create_opts.go
rhatdan: The name CGroups is too confusing for me at least.

cmd/podman/containers/create.go
rhatdan: Simplified this one a lot.  PTAL

mheon:
LGTM but I think @baude needs to approve first

cmd/podman/common/create_opts.go
baude: as i said, create.go ... you just gotta change the variable there too if you want to rename it

PR #5921, opened 2020-04-21
add entrypoint from image where needed

baude:
if the image specifies both the image and entrypoint, we need to account for that and preprend the entrypoint to the command.  this only happens if no user command and entrypoint were supplied.


TomSweeneyRedHat:
LGTM

PR #5933, opened 2020-04-22
podman, v2: handle namespaces specified on the CLI

giuseppe:
and handle differently the user namespace as it supports additional options.

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -49,8 +49,9 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"cap-drop", []string{},
 		"Drop capabilities from the container",
 	)
+	cgroupNS := ""
 	createFlags.StringVar(
-		&cf.CGroupsNS,
+		&cgroupNS,
 		"cgroupns", containerConfig.CgroupNS(),
 		"cgroup namespace to use",
 	)
@@ -247,8 +248,9 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"interactive", "i", false,
 		"Keep STDIN open even if not attached",
 	)
+	ipcNS := ""
 	createFlags.StringVar(
-		&cf.IPC,
+		&ipcNS,
 		"ipc", containerConfig.IPCNS(),
 		"IPC namespace to use",
 	)
@@ -329,8 +331,9 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"use `OS` instead of the running OS for choosing images",
 	)
 	// markFlagHidden(createFlags, "override-os")
+	pid := ""
 	createFlags.StringVar(
-		&cf.PID,
+		&pid,
 		"pid", containerConfig.PidNS(),
 		"PID namespace to use",
 	)
@@ -461,13 +464,15 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"user", "u", "",
 		"Username or UID (format: <name|uid>[:<group|gid>])",
 	)
+	userNS := ""
 	createFlags.StringVar(
-		&cf.UserNS,
+		&userNS,
 		"userns", containerConfig.Containers.UserNS,
 		"User namespace to use",
 	)
+	utsNS := ""
 	createFlags.StringVar(
-		&cf.UTS,
+		&utsNS,
 		"uts", containerConfig.Containers.UTSNS,
 		"UTS namespace to use",
 	)
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -394,8 +394,9 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"security-opt", containerConfig.SecurityOptions(),
 		"Security Options",
 	)
+	shmSize := ""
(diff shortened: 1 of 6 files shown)


cmd/podman/common/create.go
```diff
@@ -396,8 +396,8 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	)
 	createFlags.StringVar(
 		&cf.ShmSize,
-		"shm-size", containerConfig.ShmSize(),
-		"Size of /dev/shm "+sizeWithUnitFormat,
+		"shm-size", "",
```
rhatdan, line 399: Why did you take away default?

vrothberg: It's meant to know in the backend if it has been specified or not. A comment would help to make it explicit - some might want to make it consistent in the future.

rhatdan: I would prefer to move away from using the create config struct and actually checking to see if the user specified options via cobra, and then telling the user what the true default is via cobra, rather then hacking each one up.

On the Remote case we can show them as "".

giuseppe: can we address this separately?  Otherwise it currently fails on each container that has a IPC namespace with:

```
Error: invalid config provided: cannot set shmsize when creating an IPC namespace
```

rhatdan: Why can't it?  I think when creating a IPC Namespace this should be fine, when sharing an IPC Namespace it should not be allowed.

giuseppe: the error message is wrong, but the check is correct:

```
	if s.ContainerStorageConfig.ShmSize != nil && !s.ContainerStorageConfig.IpcNS.IsPrivate() {
		return errors.New("cannot set shmsize when creating an IPC namespace")
	}
```

we cannot set the shmsize if the IPC is not private.

I got confused by the error message, I get the error when using `--ipc=host`.

I've added another patch to fix the error message

mheon: Can't we check if the flag is set and only add it to SpecGen if that is true?

cmd/podman/common/specgen.go
```diff
@@ -222,55 +222,27 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
-	//	}
-	//}
-
-	// TODO Remove when done with namespaces for realz
-	// Setting a default for IPC to get this working
-	s.IpcNS = specgen.Namespace{
-		NSMode: specgen.Private,
-		Value:  "",
+	for k, v := range map[string]*specgen.Namespace{
+		c.IPC: &s.IpcNS,
```
mheon, line 226: We need to avoid setting these if they were not explicitly set by the user; we want the backend to be able to select defaults based on whether the container is in a pod and the server's containers.conf. I think we can do an easy IsSet on the flag in question, if the parsed CLI actually made it into this function.

mheon:
Code LGTM overall but I think we need to get the full parsed CLI into the backend - we need to actually be able to tell if a flag was set, not just what it's set to.

rhatdan:
Yes I agree, I think we should dump create_opts.go altogether and handle this the way we do net options.

rhatdan:
@giuseppe Please rebase to pass the integration tests.

cmd/podman/common/specgen.go
giuseppe: yes I agree.  We need to set only what was explicitly specified.

Do you prefer I do it as part of this PR or as a follow up?

mheon: This PR would be better, if you're willing.

giuseppe: pushed an updated version

giuseppe:
> @giuseppe Please rebase to pass the integration tests.

tests are passing now

cmd/podman/containers/create.go
```diff
@@ -145,6 +145,21 @@ func createInit(c *cobra.Command) error {
 	if c.Flag("no-hosts").Changed && c.Flag("add-host").Changed {
 		return errors.Errorf("--no-hosts and --add-host cannot be set together")
 	}
+	if c.Flag("userns").Changed {
```
rhatdan, line 148: Can't you handle the shm-size like this, then we can go back to revealing to the user the system default, using the standard mechanism.

pkg/specgen/container_validate.go
```diff
@@ -54,7 +54,7 @@ func (s *SpecGenerator) Validate() error {
 	}
 	// shmsize conflicts with IPC namespace
 	if s.ContainerStorageConfig.ShmSize != nil && !s.ContainerStorageConfig.IpcNS.IsPrivate() {
-		return errors.New("cannot set shmsize when creating an IPC namespace")
+		return errors.New("cannot set shmsize without creating an IPC namespace")
```
rhatdan, line 57: How about 
"cannot set shmsize  when running in the `host` IPC Namespace"
That at least will get the user to look at their host settings.

cmd/podman/containers/create.go
giuseppe: good idea.  Fixed it

pkg/specgen/container_validate.go
giuseppe: yes that is better.  Fixed

rhatdan:
LGTM

PR #5935, opened 2020-04-22
Add SELinux label selection code to pkg/utils.go

rhatdan:
We need this code in order to get it shared in CRI-O.
I do not want to vender libpod 2.0 into CRI-O yet, since
it is not released. This code is not used in podman in
v1.9.0 so this will not break anything.


rhatdan:
@mheon @umohnani8 PTAL

rhatdan:
Sure, will have to do this in master as well.

PR #5937, opened 2020-04-22
Move selinux labeling support from pkg/util to pkg/selinux

rhatdan:
The goal here is to make the package less heavy and not overload
the pkg/util.

### DIFF
--- a/pkg/selinux/selinux.go
+++ b/pkg/selinux/selinux.go
@@ -1,0 +1,40 @@
+package util
+
+import (
+	"github.com/opencontainers/selinux/go-selinux"
+)
+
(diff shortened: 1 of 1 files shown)


rhatdan:
@vrothberg PTAL

mheon:
LGTM

mheon:
Though, I don't see anything removed from pkg/util in here ?

rhatdan:
@mheon this is against the v1.9 branch not against master.  The other PR is against master and removes the SELinux stuff from utils.go.  We did not have the selinux stuff in pkg/utils.go in v1.9.0

PR #5940, opened 2020-04-22
Fix podman rm to have correct exit codes

rhatdan:
If you attempt to remove a running container is it supposed to exit with
2
If you attempt to remove a non existing container is is supposed to exit with
1

### DIFF
--- a/cmd/podman/containers/rm.go
+++ b/cmd/podman/containers/rm.go
@@ -95,11 +95,9 @@ func rm(cmd *cobra.Command, args []string) error {
 	}
 	responses, err := registry.ContainerEngine().ContainerRm(context.Background(), args, rmOptions)
 	if err != nil {
-		// TODO exitcode is a global main variable to track exit codes.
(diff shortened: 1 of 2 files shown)


cmd/podman/containers/rm.go
```diff
@@ -108,10 +106,21 @@ func rm(cmd *cobra.Command, args []string) error {
...
+
+func setExitCode(err error) {
+	cause := errors.Cause(err)
+	switch cause {
+	case define.ErrNoSuchCtr:
+		registry.SetExitCode(1)
+	case define.ErrCtrStateInvalid:
+		registry.SetExitCode(2)
+	}
+}
```
TomSweeneyRedHat, line 126: do we need to set the registry.ExitCode to something by default if it's not one of these two?  Unknown perhaps?

jwhonce: @TomSweeneyRedHat Yes, it is initialized to 125 when created.

PR #5946, opened 2020-04-22
V2 restore libpod.Shutdown() when exiting podman commands

jwhonce:

### DIFF
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -553,3 +553,10 @@ func (ir *ImageEngine) Remove(ctx context.Context, images []string, opts entitie
 
 	return
 }
+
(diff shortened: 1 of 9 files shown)


rhatdan:
LGTM

pkg/domain/infra/abi/images.go
```diff
@@ -549,3 +549,8 @@ func (ir *ImageEngine) Remove(ctx context.Context, images []string, opts entitie
 
 	return
 }
+
+// Shutdown Libpod engine
```
mheon, line 553: One of these (containers/images shutdown) is guaranteed to error, duplicate shutdown will never work.

TomSweeneyRedHat:
LGTM.  I'd like to get a head nod from @mheon to make sure his concern was addressed.

PR #5947, opened 2020-04-22
Turn on version.go except for -v check

rhatdan:
Currently podman -v does not work but the other version checks all pass.

enabling tests to that we can get more tests running in CI.


mheon:
LGTM

PR #5953, opened 2020-04-23
Don't check configuration until user input is applied

rhatdan:

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -131,8 +131,9 @@ func NewRuntime(ctx context.Context, options ...RuntimeOption) (runtime *Runtime
 	if err != nil {
 		return nil, err
 	}
+	runtime, err = newRuntimeFromConfig(ctx, conf, options...)
 	conf.CheckCgroupsAndAdjustConfig()
-	return newRuntimeFromConfig(ctx, conf, options...)
+	return runtime, err
 }
 
 // NewRuntimeFromConfig creates a new container runtime using the given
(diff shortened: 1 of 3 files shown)


rhatdan:
@edsantiago @mheon PTAL

libpod/runtime.go
```diff
@@ -131,8 +131,9 @@ func NewRuntime(ctx context.Context, options ...RuntimeOption) (runtime *Runtime
 	if err != nil {
 		return nil, err
 	}
+	runtime, err = newRuntimeFromConfig(ctx, conf, options...)
```
mheon, line 134: Is this right? We're making the runtime before we finalize the config, looks like?

rhatdan: We need to take the input in from the user.  Currently if the user overrides the default cgroupfs, we report an error saying that the --cgroup-manager needs to be cgroupfs even if the user specified --cgroup-manager=cgroupfs.

This code is just verifying that the options that the runtime will use is correct.

baude:
LGTM

rhatdan:
@edsantiago PTAL
Could you check if this PR fixes your issue with the warnings?

edsantiago:
What timeframe do you need? Will Monday be OK? I don't have a rhel8 box right now, and just wasted an hour trying to build one with the right repos to build podman. I can't get gpgme, btrfs, devmapper, and a few others; and I can't afford any more time today. I can try to find time over the weekend if it's necessary.

rhatdan:
@edsantiago No problem.  Monday is fine.

edsantiago:
Confirmed: the error message is gone. I managed to build v1.9.0 in rhel8, confirmed that my built image has the multi-line warning; built with your patch, the warning is gone. All system tests pass, root and rootless. LGTM, adding /hold in case there are any other pending reviews. Remove when desired.

/lgtm
/hold

mheon:
/hold cancel

If this fixes our spurious warnings, LGTM on my end

PR #5954, opened 2020-04-23
Makefile: install targets should not depend on build targets

lsm5:
Distro packaging often uses non-default build and linker flags.
The current Makefile cannot be used in the package build process as it
will end up rebuilding the binaries with the default flags.

PTAL @jnovy @haircommander @baude @rhatdan 

This undoes some of https://github.com/containers/libpod/pull/3757 and https://github.com/containers/libpod/pull/2923 and would likely make some people angry :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -470,24 +470,34 @@ changelog: ## Generate changelog
 .PHONY: install
 install: .gopathok install.bin install.remote install.man install.cni install.systemd  ## Install binaries to system locations
 
-.PHONY: install.remote
-install.remote: podman-remote
+.PHONY: install.remote-nobuild
+install.remote-nobuild:
 	install ${SELINUXOPT} -d -m 755 $(DESTDIR)$(BINDIR)
 	install ${SELINUXOPT} -m 755 bin/podman-remote $(DESTDIR)$(BINDIR)/podman-remote
 	test -z "${SELINUXOPT}" || chcon --verbose --reference=$(DESTDIR)$(BINDIR)/podman-remote bin/podman-remote
 
-.PHONY: install.bin
-install.bin: podman
+.PHONY: install.remote
+install.remote: podman-remote install.remote-nobuild
+
+.PHONY: install.bin-nobuild
+install.bin-nobuild:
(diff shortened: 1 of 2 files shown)


lsm5:
I guess this would need changes elsewhere in docs wherever just the install target is mentioned. I'll go through those and update this PR..

rhatdan:
Isn't there a way that the upstream can set the proper flags, so that this still works without the change.

lsm5:
> Isn't there a way that the upstream can set the proper flags, so that this still works without the change.

While some flags may be generic enough, setting all flags for each distro might be a PITA for everyone. I feel flag setting is best left in the packaging sources. Much quicker to change / customize it there.

rhatdan:
How about if we add additional targets that distros could use
make install-nobuid
Or something like that?

lsm5:
> How about if we add additional targets that distros could use
> make install-nobuid
> Or something like that?

I guess that would end up duplicating all the install rules, yes? I'm cool with it if everyone is ok with maintaining the extra stuff.

lsm5:
I guess maybe have all the rules first mentioned in -nobuild and then:

```
install.podman: podman install.podman-nobuild
```

Would that work?

rhatdan:
Yes I think that would.

lsm5:
> Yes I think that would.

Great, I'll update PR..

lsm5:
can we get some more eyes on this please?

TomSweeneyRedHat:
It LGTM, but I definitely want other eyes on it as I know @mtrmac wasn't real hip with similar changes in another project.

lsm5:
> It LGTM, but I definitely want other eyes on it as I know @mtrmac wasn't real hip with similar changes in another project.

I didn't see the skopeo response until just now. Thanks, I'll see whatsup there..

rhatdan:
/approve
I am fine with these changes.

mtrmac:
> It LGTM, but I definitely want other eyes on it as I know @mtrmac wasn't real hip with similar changes in another project.

My questions from https://github.com/containers/skopeo/pull/900 are not directly applicable here, AFAICS `make install` _could_ actually cause a build to happen.

PR #5956, opened 2020-04-23
V2 restore podman -v command

jwhonce:
* Removed extra spaces and improved error message
* Updated tests to use gomega matchers

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -78,6 +78,10 @@ func init() {
 	)
 
 	rootFlags(registry.PodmanConfig(), rootCmd.PersistentFlags())
+
+	// "version" is a local flag to avoid collisions with sub-commands that use "-v"
+	var dummyVersion bool
+	rootCmd.Flags().BoolVarP(&dummyVersion, "version", "v", false, "Version of Podman")
 }
 
 func Execute() {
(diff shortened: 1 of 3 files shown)


cmd/podman/root.go
```diff
@@ -58,7 +58,6 @@ var (
 		Long:               "Manage pods, containers and images",
 		SilenceUsage:       true,
 		SilenceErrors:      true,
-		TraverseChildren:   true,
```
mheon, line 61: I thiiiink this might disable global flags intermixing with command-level flags, which I think we still want - `podman run -t -i --rm --log-level=debug fedora ls /` should still work

jwhonce: @mheon That still works.  TraverseChildren is about parsing parent's local flags before child command.  `root` has no parents.

rhatdan:
LGTM, Nice fix.

TomSweeneyRedHat:
LGTM assuming happy tests.

cmd/podman/root.go
```diff
@@ -78,6 +78,10 @@ func init() {
 	)
 
 	rootFlags(registry.PodmanConfig(), rootCmd.PersistentFlags())
+
+	// "version" is a local flag to avoid collisions with sub-commands that use "-v"
+	var dummyVersion bool
+	rootCmd.Flags().BoolVarP(&dummyVersion, "version", "v", false, "Version of Podman")
```
giuseppe, line 84: could we use `BoolP` here and not require `dummyVersion`?

PR #5958, opened 2020-04-23
Update vendor to containers/common v0.8.2

rhatdan:
This will fix a couple of issues caused by the move to containers.conf

If a libpod.conf file still exists,  we will ignore its events_logger
definition and use "file"

If you are running rootless on cgroupsV1 we will default to host cgroupns.

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -376,6 +376,10 @@ func ParseCreateOpts(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 		}
 	}
 
+	usernsType := c.String("userns")
+	if !c.IsSet("userns") && !idmappings.HostUIDMapping {
+		usernsType = "private"
+	}
 	// Kernel Namespaces
 	// TODO Fix handling of namespace from pod
 	// Instead of integrating here, should be done in libpod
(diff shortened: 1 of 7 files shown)


rhatdan:
@mheon @vrothberg @giuseppe PTAL

giuseppe:
> If you are running rootless on cgroupsV1 we will default to host cgroupns.

that applies also to root containers.  We are doing it for two reasons: 1) avoid a breaking change with Docker and with older Podman releases, 2) not worry if the kernel supports cgroup namespaces.  With cgroup v2 we are sure the kernel is new enough to have cgroupns.

LGTM

mheon:
LGTM

mheon:
Tests are red, though

mheon:
`[Fail] Podman UserNS support [It] podman --userns=container:CTR `

Consistent, rootless local and remote - I think we may have actually broken this

rhatdan:
@giuseppe The error seems to be one container attempting to join a different container in a different usernamespace

# ./bin/podman run -d --uidmap=0:0:1 --uidmap=1:1:4998 --name test1 alpine top
10640bb12220e2e1a3d941ccaed724c9c0392216c77427b83a4f9a1fd7828b02
# ./bin/podman run --rm --userns=container:test1 --ipc=container:test1 alpine cat /proc/self/uid_map
Error: cannot setns `(null)`: Invalid argument
write to sync socket: Broken pipe: OCI runtime error

Only change in containers.conf patch that I think could effect this switching userns=private-> userns=host?

giuseppe:
> Only change in containers.conf patch that I think could effect this switching userns=private-> userns=host?

there is an issue in crun (wrong error message, but it is just aesthetic, opened a PR to address it), and we need the equivalent of https://github.com/containers/libpod/pull/5971/files#diff-876ae6fe9ffc61d00ef62cd7ee9a677cR217-R220 in Podman 1.9 after the change in containers.conf.

If there are mappings specified, the default must be changed to "private".

PR #5962, opened 2020-04-23
Fix SELinux functions names to not be repetitive

rhatdan:
Since functions are now in an selinux subpackage, they should not start with
SELinux


TomSweeneyRedHat:
LGTM assuming happy tests

PR #5971, opened 2020-04-24
v2, podman: fix and enable all run_userns_test.go tests

giuseppe:
more details in each commit.

### DIFF
--- a/pkg/namespaces/namespaces.go
+++ b/pkg/namespaces/namespaces.go
@@ -31,7 +31,7 @@ func (n CgroupMode) IsHost() bool {
 
 // IsDefaultValue indicates whether the cgroup namespace has the default value.
 func (n CgroupMode) IsDefaultValue() bool {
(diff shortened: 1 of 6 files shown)


pkg/namespaces/namespaces.go
```diff
@@ -102,6 +102,11 @@ func (n UsernsMode) IsAuto() bool {
 	return parts[0] == "auto"
 }
 
+// IsDefaultValue indicates whether the user namespace has the default value.
+func (n UsernsMode) IsDefaultValue() bool {
```
mheon, line 106: Shouldn't this also check for the explicit Default constant?

giuseppe: I've copied it from cgroupns.

I've fixed both and pushed a new version

TomSweeneyRedHat:
LGTM assuming happy tests

PR #5972, opened 2020-04-24
Do not join pod namespaces without an infra ctr

mheon:
We do not want to join pod namespaces if no infra container is present. A pod may claim it shares namespaces without an infra container (I'll take an action item to fix that - it really should not be allowed), which was tripping up our default namespace code and forcing us to try and join the namespaces of the nonexistant infra container.


mheon:
@vrothberg PTAL - should fix your tests

PR #5975, opened 2020-04-24
[v2] tests: fix pause tests

giuseppe:
adapt the tests to work with Podman v2.

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -223,7 +223,7 @@ func createPsOut() (string, string) {
 	}
 	headers := defaultHeaders
 	row += "{{.ID}}"
-	row += "\t{{.Image}}\t{{.Command}}\t{{.CreatedHuman}}\t{{.State}}\t{{.Ports}}\t{{.Names}}"
+	row += "\t{{.Image}}\t{{.Command}}\t{{.CreatedHuman}}\t{{.Status}}\t{{.Ports}}\t{{.Names}}"
 
 	if listOpts.Pod {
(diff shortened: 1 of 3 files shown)


test/utils/utils.go
```diff
@@ -167,7 +167,7 @@ func (p *PodmanTest) NumberOfPods() int {
 // This function assumes only one container is active.
 func (p *PodmanTest) GetContainerStatus() string {
 	var podmanArgs = []string{"ps"}
-	podmanArgs = append(podmanArgs, "--all", "--format={{.Status}}")
+	podmanArgs = append(podmanArgs, "--all", "--format={{.State}}")
```
mheon, line 170: This actually seems pretty significant - is this a breaking change from Docker?

giuseppe: yes, it is a breaking change with Docker

mheon: I think we need to either alias `Status` to `State` or change it back completely, then

jwhonce: We could rename:
https://github.com/containers/libpod/blob/master/cmd/podman/containers/ps.go#L268 to Status() or add a Status() method that calls State().  /cc @baude Was State() chosen for a reason?

giuseppe: thanks for the hint, I will add a function Status() that calls State() for now.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

giuseppe:
@TomSweeneyRedHat can you use "/" + "lgtm" ?

PR #5976, opened 2020-04-24
Add --os to manifest add

QiWang19:
Add --os to manifest add for overriding the os field.


rhatdan:
LGTM

PR #5978, opened 2020-04-24
Make podman container list == podman ps

rhatdan:
Also make
podman create -p PUBLISHPorts
work.

This PR fixes ps_test.go

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -18,6 +18,7 @@ import (
 	"github.com/docker/go-units"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


jwhonce:
LGTM

rhatdan:
@mheon @baude @vrothberg @TomSweeneyRedHat @giuseppe @QiWang19 PTAL

cmd/podman/containers/ps.go
```diff
@@ -47,7 +48,9 @@ func init() {
 		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
 		Command: psCommand,
 	})
-	flags := psCommand.Flags()
+	listFlagSet(psCommand.Flags())
+}
+func listFlagSet(flags *pflag.FlagSet) {
```
TomSweeneyRedHat, line 53: a nite of a nit, it would be nice to have a blank line between the last function and the start of this function.

TomSweeneyRedHat:
One nit.  The changes LGTM in general, but I'd like to get a head nod from @mheon as I'm not 100% sure on all the port processing removal.

rhatdan:
@baude PTAL

@TomSweeneyRedHat @mheon This PR was about etting the ps test to pass and fixing one of the port handling issues (Just to get the test to pass.)  The Port handling needs more work and needs to be handled on the server side, and was not intended to be fixed here.  The changes you see on port handling here were around a function on the client side to validate user input.  This function was copied from old code and was basically doing little other then the validation.

mheon:
We need to add Expose to Specgen, probably as either a `[]uint16` or `map[uint16]strct{}`, so we can handle it on the backend. Otherwise, I suppose this LGTM - we'll need the code here, in modified form, for the backend, but we can probably source from the v1.9 branch if we need to?

PR #5982, opened 2020-04-25
Enable pod ps integration tests

baude:
Enable integration tests for pod ps.

In addition, fixed bug where output was still using slice go template routines and would fail when no infra container was present.  Added integration test to prevent future regressions.


mheon:
LGTM. Thanks for taking this!

PR #5985, opened 2020-04-26
Enable these tests

rhatdan:
These tests are passing locally.


giuseppe:
pause tests are skipped on cgroup v2 because we have a wrong check for the freezer controller.  We need these fixes: https://github.com/containers/libpod/pull/5975

giuseppe:
let's merge anyway, I'll rebase on top of this PR

/lgtm

PR #5991, opened 2020-04-26
Fix typos in messages pod rm

sujil02:
Fix typos in messages

### DIFF
--- a/pkg/domain/infra/abi/pods.go
+++ b/pkg/domain/infra/abi/pods.go
@@ -236,8 +236,6 @@ func (ic *ContainerEngine) PodRm(ctx context.Context, namesOrIds []string, optio
 		err := ic.Libpod.RemovePod(ctx, p, true, options.Force)
 		if err != nil {
 			report.Err = err
-			reports = append(reports, &report)
-			continue
 		}
 		reports = append(reports, &report)
 	}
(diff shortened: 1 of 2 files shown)


pkg/domain/infra/abi/pods.go
```diff
@@ -226,15 +226,14 @@ func (ic *ContainerEngine) PodRm(ctx context.Context, namesOrIds []string, optio
 		reports []*entities.PodRmReport
 	)
 	pods, err := getPodsByContext(options.All, options.Latest, namesOrIds, ic.Libpod)
-	if err != nil {
+	if err != nil && !(options.Ignore && errors.Cause(err) == define.ErrNoSuchPod) {
```
vrothberg, line 229: I believe we need to move ignore into `getPodsByContext`. If we specify three pods and the first is missing, we still want the other two. Does `podman pod rm bogus existingA existingB` work correctly and removes the last two?

vrothberg: I had a look at the code and it's fine. `getPodsByContext` runs over all specified ctrs/pods.

vrothberg:
May already be fixed in master now.

sujil02:
@giuseppe rm pod fixes got merged with #5997 should I close this one? however, it has some help message typos fixed.

sujil02:
@rhatdan I accidentally pushed to this branch thought this was merged. no new changes are done. need lgtm again.

PR #5997, opened 2020-04-27
v2, podman: fix a bunch of "podman pod *" tests

giuseppe:


rhatdan:
Nice work
LGTM

pkg/specgen/generate/pod_create.go
```diff
@@ -46,6 +46,13 @@ func createPodOptions(p *specgen.PodSpecGenerator) ([]libpod.PodCreateOption, er
 	if len(p.HostAdd) > 0 {
 		options = append(options, libpod.WithPodHosts(p.HostAdd))
 	}
+	if len(p.DNSServer) > 0 {
```
mheon, line 49: Need to check for `--dns=none`

giuseppe: that is already managed through `NetFlagsToNetOptions`

PR #5999, opened 2020-04-27
v2, podman: fix pod tests

giuseppe:
fix remaining failures with pod tests.

Only pause and pod stats are not addressed in this PR, since we are already tracking pause and pod stats in:

https://github.com/containers/libpod/pull/5975
https://github.com/containers/libpod/pull/5966

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -385,6 +385,16 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 			g.AddLinuxGIDMapping(uint32(0), uint32(0), uint32(1))
 		}
 	}
+
+	for _, i := range c.config.Spec.Linux.Namespaces {
+		if i.Type == spec.UTSNamespace {
+			hostname := c.Hostname()
+			g.SetHostname(hostname)
+			g.AddProcessEnv("HOSTNAME", hostname)
+			break
+		}
+	}
+
 	if c.config.UTSNsCtr != "" {
 		if err := c.addNamespaceContainer(&g, UTSNS, c.config.UTSNsCtr, spec.UTSNamespace); err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


test/e2e/pod_infra_container_test.go
```diff
@@ -95,7 +94,7 @@ var _ = Describe("Podman pod create", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 
-		check := podmanTest.Podman([]string{"ps", "-a", "--no-trunc", "--ns", "--format", "{{.IPC}} {{.NET}}"})
+		check := podmanTest.Podman([]string{"ps", "-a", "--no-trunc", "--ns", "--format", "{{.Namespaces.IPC}} {{.Namespaces.NET}}"})
```
vrothberg, line 97: The "Namespaces" changes seem like a breaking change. Is this known/intended? @mheon @baude

baude: the change is intended, though we should probably use methods to keep backwards compat?

vrothberg: +1 ... then we can make sure to test the new and old format here

giuseppe: good idea.  I've added methods to keep compatibility and fixed the tests to use both versions

libpod/container_internal_linux.go
```diff
@@ -385,6 +385,16 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 			g.AddLinuxGIDMapping(uint32(0), uint32(0), uint32(1))
 		}
 	}
+
+	for _, i := range c.config.Spec.Linux.Namespaces {
+		if i.Type == spec.UTSNamespace {
```
mheon, line 390: Shouldn't we check for path here? To see if a path to the UTS namespace was given?

giuseppe: I've only moved the existing chunk.

When we join an existing container these values are overwritten in `addNamespaceContainer`

PR #6000, opened 2020-04-27
Add support for volumes-from, image volumes, init

mheon:
This should complete Podmanv2's support for volume-related flags. Most code was sourced from the old pkg/spec implementation with modifications to account for the split between frontend flags (volume, mount, tmpfs) and the backend flags implemented here.

Also enables tests for podman run with volumes


mheon:
Hold for a bit. Our entrypoint handling is broken.

mheon:
Apparently run_volume_test also has a metric ton of `podman volume create` tests in it, which is lovely. Going to have to disable for now.

mheon:
Tests should go green now

rhatdan:
LGTM

PR #6001, opened 2020-04-27
enable load integration tests

baude:
fix bug where multiple images can be loaded and have to be able to handle the return of multiple names

### DIFF
--- a/test/e2e/load_test.go
+++ b/test/e2e/load_test.go
@@ -20,7 +20,6 @@ var _ = Describe("Podman load", func() {
 	)
 
 	BeforeEach(func() {
-		Skip(v2fail)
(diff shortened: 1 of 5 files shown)


test/e2e/load_test.go
```diff
@@ -20,7 +20,7 @@ var _ = Describe("Podman load", func() {
 	)
 
 	BeforeEach(func() {
-		Skip(v2fail)
+		//Skip(v2fail)
```
giuseppe, line 23: could we drop completely this line?

baude: yes ofc my mistake

baude:
fixed

PR #6004, opened 2020-04-27
Set up ulimits for rootless containers.

rhatdan:
Currently we are setting the maximum limits for rootful podman containers,
no reason not to set them by default for rootless users as well

### DIFF
--- a/pkg/spec/spec.go
+++ b/pkg/spec/spec.go
@@ -16,6 +16,8 @@ import (
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 const CpuPeriod = 100000
@@ -533,11 +535,31 @@ func addRlimits(config *CreateConfig, g *generate.Generator) error {
 	// If not explicitly overridden by the user, default number of open
 	// files and number of processes to the maximum they can be set to
 	// (without overriding a sysctl)
-	if !nofileSet && !isRootless {
-		g.AddProcessRlimits("RLIMIT_NOFILE", kernelMax, kernelMax)
+	if !nofileSet {
+		max := kernelMax
+		current := kernelMax
+		if isRootless {
+			var rlimit unix.Rlimit
+			if err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit); err != nil {
+				logrus.Warnf("failed to return RLIMIT_NOFILE ulimit %q", err)
+			}
+			current = rlimit.Cur
+			max = rlimit.Max
+		}
+		g.AddProcessRlimits("RLIMIT_NOFILE", current, max)
 	}
-	if !nprocSet && !isRootless {
-		g.AddProcessRlimits("RLIMIT_NPROC", kernelMax, kernelMax)
+	if !nprocSet {
+		max := kernelMax
+		current := kernelMax
+		if isRootless {
+			var rlimit unix.Rlimit
+			if err := unix.Getrlimit(unix.RLIMIT_NPROC, &rlimit); err != nil {
+				logrus.Warnf("failed to return RLIMIT_NPROC ulimit %q", err)
+			}
+			current = rlimit.Cur
+			max = rlimit.Max
+		}
+		g.AddProcessRlimits("RLIMIT_NPROC", current, max)
 	}
 
 	return nil
--- a/pkg/specgen/generate/oci.go
+++ b/pkg/specgen/generate/oci.go
@@ -13,6 +13,8 @@ import (
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 func addRlimits(s *specgen.SpecGenerator, g *generate.Generator) error {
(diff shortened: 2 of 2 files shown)


pkg/spec/spec.go
```diff
@@ -533,11 +534,27 @@ func addRlimits(config *CreateConfig, g *generate.Generator) error {
 	// If not explicitly overridden by the user, default number of open
 	// files and number of processes to the maximum they can be set to
 	// (without overriding a sysctl)
-	if !nofileSet && !isRootless {
-		g.AddProcessRlimits("RLIMIT_NOFILE", kernelMax, kernelMax)
+	if !nofileSet {
+		if isRootless {
+			var rlimit unix.Rlimit
+			if err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit); err != nil {
```
mheon, line 540: I don't think we can hard-error on these; I can see cases where this won't work, but the rest of Podman will function fine.

rhatdan: Switched to warning.

pkg/spec/spec.go
```diff
@@ -533,11 +534,27 @@ func addRlimits(config *CreateConfig, g *generate.Generator) error {
...
-		g.AddProcessRlimits("RLIMIT_NPROC", kernelMax, kernelMax)
+	if !nprocSet {
+		if isRootless {
+			var rlimit unix.Rlimit
+			if err := unix.Getrlimit(unix.RLIMIT_NPROC, &rlimit); err != nil {
+				return errors.Wrapf(err, "failed to return RLIMIT_NPROC ulimit")
+			}
+			g.AddProcessRlimits("RLIMIT_NPROC", rlimit.Cur, rlimit.Max)
+		} else {
+			g.AddProcessRlimits("RLIMIT_NPROC", kernelMax, kernelMax)
```
portante, line 556: Perhaps consider a small DRY principle change?

```
if !nprocSet{
    var ( cur uint64 = kernelMax, max uint64 kernel Max )
    if isRootless {
        var rlimit unix.Rlimit
        if err := unit.Getrlimit(unix.RLIMIT_NPROC, &rlimit); err != nil {
            return errors.Wrapf(err, "failed to return RLIMIT_NPROC ulimit")
        }
        cur = rlimit.Cur, max = rlimit.Max
    }
    g.AddProcessRlimits("RLIMIT_NPROC", cur, max)
}
```

pkg/spec/spec.go
```diff
@@ -533,11 +534,27 @@ func addRlimits(config *CreateConfig, g *generate.Generator) error {
...
-		g.AddProcessRlimits("RLIMIT_NOFILE", kernelMax, kernelMax)
+	if !nofileSet {
+		if isRootless {
+			var rlimit unix.Rlimit
+			if err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit); err != nil {
+				return errors.Wrapf(err, "failed to return RLIMIT_NOFILE ulimit")
+			}
+			g.AddProcessRlimits("RLIMIT_NOFILE", rlimit.Cur, rlimit.Max)
+		} else {
+			g.AddProcessRlimits("RLIMIT_NOFILE", kernelMax, kernelMax)
```
portante, line 545: Why is a hard-coded constant of `1048576` used for `kernelMax` for both number of files and number of processes?  Is there a proper default that could be determined from the environment?

pkg/spec/spec.go
portante: And this DRY would apply to number of files as well.

pkg/specgen/generate/oci.go
```diff
@@ -41,11 +43,29 @@ func addRlimits(s *specgen.SpecGenerator, g *generate.Generator) error {
...
+	if !nprocSet {
+		if isRootless {
+			var rlimit unix.Rlimit
+			if err := unix.Getrlimit(unix.RLIMIT_NPROC, &rlimit); err == nil {
+				g.AddProcessRlimits("RLIMIT_NPROC", rlimit.Cur, rlimit.Max)
+			} else {
+				logrus.Warnf("failed to get RLIMIT_NPROC ulimit %q", err)
+			}
+		} else {
+			g.AddProcessRlimits("RLIMIT_NPROC", kernelMax, kernelMax)
```
portante, line 67: Seems like the same DRY principle can be applied here.

pkg/spec/spec.go
giuseppe: we have `--ulimit host` that tells Podman to not set any limit block in the OCI configuration file.  When there are no limit sets the OCI runtime won't apply any limit to the container so the host settings are propagated to the container.

Do we have to make that the default (AFAICS it is not yet supported by Buildah)?

pkg/spec/spec.go
rhatdan: Don't know, I was trying to fix rootless case.  I would actually think we should just use the current environment in both cases.  But not sure.  I am slightly worried about root case being able to override default MAX if the process has CAP_SYS_RESOURCE.  
```
man capabilities
...
       CAP_SYS_RESOURCE
              * Use reserved space on ext2 filesystems;
              * make ioctl(2) calls controlling ext3 journaling;
              * override disk quota limits;
              * increase resource limits (see setrlimit(2));
              * override RLIMIT_NPROC resource limit;
              * override maximum number of consoles on console allocation;
              * override maximum number of keymaps;
              * allow more than 64hz interrupts from the real-time clock;
              * raise  msg_qbytes  limit  for  a  System  V  message  queue  above  the  limit  in
                /proc/sys/kernel/msgmnb (see msgop(2) and msgctl(2));
              * allow  the RLIMIT_NOFILE resource limit on the number of "in-flight" file descrip‐
                tors to be bypassed when passing file descriptors to another process  via  a  UNIX
                domain socket (see unix(7));
              * override  the /proc/sys/fs/pipe-size-max limit when setting the capacity of a pipe
                using the F_SETPIPE_SZ fcntl(2) command.
              * use F_SETPIPE_SZ to increase the capacity of a pipe above the limit  specified  by
                /proc/sys/fs/pipe-max-size;
              * override  /proc/sys/fs/mqueue/queues_max  limit when creating POSIX message queues
                (see mq_overview(7));
              * employ the prctl(2) PR_SET_MM operation;
              * set /proc/[pid]/oom_score_adj to a value lower  than  the  value  last  set  by  a
                process with CAP_SYS_RESOURCE.
```

@giuseppe @mheon Any idea where this kernelMax came from?

giuseppe: I think it is just the default used by Docker.

https://github.com/containers/buildah/commit/a2b018430df1e8e89b23cb5bfa49e4d3517e1c2d


Also, shouldn't these changes be done in `pkg/specgen` instead of `pkg/spec`?

rhatdan:
@TomSweeneyRedHat @baude @giuseppe @mheon @vrothberg @jwhonce @QiWang19 PTAL

pkg/spec/spec.go
rhatdan: It is being done in both places.  pkg/spec is still used by varlink.

pkg/spec/spec.go
rhatdan: I would figure it was the default.  Where does the 1024,1024 come from that we have now?
I also found it curious that 

$ podman unshare grep open /proc/self/limits 
Max open files            524288               524288               files     

Why is the default to set the soft limit== harlimit in a user namespace?

giuseppe: I think we do that automatically to avoid the `max number of open files` error when binding too many ports: https://github.com/containers/libpod/commit/9126b45a3f2255b103ef60f305f8ca9ab2805b50.


The issue with rootless is that we cannot set higher limits that the user currently has, so we can't default to the values used by root as they could not be allowed in the current session.

At the same time, we risk of setting higher limits than a root container would have if defaulting to `--ulimit host`.

We'd probably have to check each default value and lower it when it is higher than what the user is allowed to use, but we can adress this part later

rhatdan: So should we merge this change or drop it?

baude:
lgtm

PR #6007, opened 2020-04-27
enable volume integration tests

baude:
enabled integration tests for volumes.  there are two exceptions that still need work because of something not yet implemented.

also, add code to deal with the fact that containers conf appears to set a local volume driver where it used to be simply blank.

### DIFF
--- a/cmd/podman/volumes/list.go
+++ b/cmd/podman/volumes/list.go
@@ -2,6 +2,7 @@ package volumes
 
 import (
 	"context"
+	"fmt"
 	"html/template"
 	"io"
 	"os"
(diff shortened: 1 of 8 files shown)


mheon:
@baude Can you also turn on the remaining tests in `run_volume_test.go`?

baude:
> @baude Can you also turn on the remaining tests in `run_volume_test.go`?

ill peek

cmd/podman/volumes/list.go
```diff
@@ -57,6 +58,9 @@ func list(cmd *cobra.Command, args []string) error {
 	if cliOpts.Quiet && cmd.Flag("format").Changed {
 		return errors.New("quiet and format flags cannot be used together")
 	}
+	if len(cliOpts.Filter) > 0 {
+		lsOpts.Filter = make(map[string][]string)
```
rhatdan, line 62: Should this be optimized with the len(cliOpts.Filter)?

 		lsOpts.Filter = make(map[string][]string. len(cliOpts.Filter))

baude:
all ready for merge

cmd/podman/volumes/list.go
baude: i think theoretically at some point you can have multiple filters with the same key in which this len assumption would be incorrect.

Issue #5915, opened 2020-04-21
native healthchecks are unreliable/inconsistent


cjeanner:
<!--
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------
Use the commands below to provide key information from your environment:
You do NOT have to include this information if this is a FEATURE REQUEST

**NOTE** A large number of issues reported against Podman are often found to already be fixed
in more current versions of the project.  Before reporting an issue, please verify the
version you are running with `podman version` and compare it to the latest release
documented on the top of Podman's [README.md](../README.md).  If they differ, please
update your version of Podman to the latest possible and retry your command before creating
an issue.

If you are filing a bug against `podman build`, please instead file a bug
against Buildah (https://github.com/containers/buildah/issues). Podman build
executes Buildah to perform container builds, and as such the Buildah
maintainers are best equipped to handle these bugs.
-->

**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
Using native podman healthcheck feature is unreliable since there are inconsistencies between live state and the state reported within "inspect". For instance:
```Bash
podman inspect neutron_api | jq '.[0]["State"]["Healthcheck"]'                                                                                                                                              
{                                                                                                                                                                                                                                             
  "Status": "healthy",
  "FailingStreak": 0,
  "Log": [
    {
      "Start": "2020-04-21T11:21:38.122154731Z",
      "End": "2020-04-21T11:21:39.199659552Z",
      "ExitCode": 0,
      "Output": ""
    },
    {
      "Start": "2020-04-21T11:21:38.917118406Z",
      "End": "2020-04-21T11:21:39.575425172Z",
      "ExitCode": 0,
      "Output": ""
    },
    {
      "Start": "2020-04-21T11:21:39.251824671Z",
      "End": "2020-04-21T11:21:39.767073492Z",
      "ExitCode": 0,
      "Output": ""
    },
    {
      "Start": "2020-04-21T11:21:47.97086488Z",
      "End": "2020-04-21T11:21:48.443987962Z",
      "ExitCode": 1,
      "Output": ""
    },
    {
      "Start": "2020-04-21T11:21:47.987444237Z",
      "End": "2020-04-21T11:21:48.497487617Z",
      "ExitCode": 0,
      "Output": ""
    }
  ]
}
```

while:
```
[CentOS-8.1 - root@undercloud ~]# podman healthcheck run neutron_api
unhealthy
```
[...]

PR #6008, opened 2020-04-27
separate healthcheck and container log paths

baude:
instead of using the container log path to derive where to put the healthchecks, we now put them into the rundir to avoid collision of health check log files when the log path is set by user.

Fixes: #5915


mheon:
LGTM

baude:
we should get this into 1.9.x

PR #6009, opened 2020-04-27
separate healthcheck and container log paths

baude:
instead of using the container log path to derive where to put the healthchecks, we now put them into the rundir to avoid collision of health check log files when the log path is set by user.

Fixes: #5915

### DIFF
--- a/libpod/healthcheck.go
+++ b/libpod/healthcheck.go
@@ -238,7 +238,7 @@ func (c *Container) updateHealthCheckLog(hcl define.HealthCheckLog, inStartPerio
 
 // HealthCheckLogPath returns the path for where the health check log is
 func (c *Container) healthCheckLogPath() string {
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

rhatdan:
Should this be back ported to podman v1.9?

cjeanner:
Oh, sad to see it moving back in a location hard to share across containers. Anyway, as long as it solves the issue, it's good for me :).
I think it should be backported to 1.9 as well.

EmilienM:
> Should this be back ported to podman v1.9?

Yes, please. We'll need it backported into RHEL 8.x if possible so our healthchecks are correctly logged again.

Thanks

baude:
guys were are way ahead of you ... see #6008

EmilienM:
> guys were are way ahead of you ... see #6008

Thanks ;-)

PR #6011, opened 2020-04-28
Fixes podman save fails when specifying an image using a digest fixes-5234

sujil02:
Adds check to parse normalized name and create docker archive dst reference
for tagged untagged image.

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -1487,14 +1487,14 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
 		}
 		manifestType = manifest.DockerV2Schema2MediaType
 	case "docker-archive", "":
-		dst := output
 		destImageName := imageNameForSaveDestination(i, source)
-		if destImageName != "" {
-			dst = fmt.Sprintf("%s:%s", dst, destImageName)
-		}
-		destRef, err = dockerarchive.ParseReference(dst) // FIXME? Add dockerarchive.NewReference
+		ref, err := dockerArchiveDstReference(destImageName)
 		if err != nil {
-			return errors.Wrapf(err, "error getting Docker archive ImageReference for %q", dst)
+			return err
+		}
+		destRef, err = dockerarchive.NewReference(output, ref)
+		if err != nil {
+			return errors.Wrapf(err, "error getting Docker archive ImageReference for %s:%v", output, ref)
 		}
 	default:
 		return errors.Errorf("unknown format option %q", format)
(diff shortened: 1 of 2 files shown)


sujil02:
@mtrmac Thank you for the help.
PTAL

libpod/image/image.go
```diff
@@ -1487,14 +1487,14 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
...
-			dst = fmt.Sprintf("%s:%s", dst, destImageName)
+		ref, err := dockerArchiveDstReference(destImageName)
+		if err != nil {
+			return err
 		}
-		destRef, err = dockerarchive.ParseReference(dst) // FIXME? Add dockerarchive.NewReference
+		destRef, err = dockerarchive.NewReference(output, ref)
 		if err != nil {
-			return errors.Wrapf(err, "error getting Docker archive ImageReference for %q", dst)
+			return errors.Wrapf(err, "error getting Docker archive ImageReference for %q", output)
```
mtrmac, line 1497: It might be useful to include both `output` and `ref.String()` here.

libpod/image/image.go
```diff
@@ -1514,6 +1514,23 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
 	return nil
 }
 
+// dockerArchiveDestReference returns a NamedTagged reference for a tagged image and nil for untagged image.
+func dockerArchiveDstReference(normalizedInput string) (reference.NamedTagged, error) {
+	if normalizedInput == "" {
+		return nil, errors.Errorf("docker-archive reference %s isn't of the form <path>[:<reference>]", normalizedInput)
```
mtrmac, line 1520: This previously didn’t fail, but just called `ParseReference` without a name = called `NewReference(output, nil)`.

(And if it were intentional, the error text does not match the situation, `normalizedInput` from `imageNameForSaveDestination` does not contain any path.)

libpod/image/image.go
```diff
@@ -1514,6 +1514,23 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
...
+	}
+	ref, err := reference.ParseNormalizedNamed(normalizedInput)
+	if err != nil {
+		return nil, errors.Wrapf(err, "docker-archive parsing reference")
```
mtrmac, line 1524: It might be useful to include `normalizedInput` in the error message (`ParseNormalizedNamed` doesn’t do that itself).

libpod/image/image.go
sujil02: Included.

libpod/image/image.go
sujil02: updated error message to have normalizedInput

libpod/image/image.go
sujil02: Update to not failing and instead `NewReference(output, nil)`. PTAL

sujil02:
> LGTM.
> 
> (Should there be tests for the newly working functionality?)

Added test. PTAL

PR #6016, opened 2020-04-28
v2, podman: fix create and entrypoint tests

giuseppe:

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -364,20 +364,20 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	s.WorkDir = workDir
 	entrypoint := []string{}
 	userCommand := []string{}
-	if ep := c.Entrypoint; len(ep) > 0 {
-		// Check if entrypoint specified is json
-		if err := json.Unmarshal([]byte(c.Entrypoint), &entrypoint); err != nil {
-			entrypoint = append(entrypoint, ep)
+	if c.Entrypoint != nil {
+		if ep := *c.Entrypoint; len(ep) > 0 {
+			// Check if entrypoint specified is json
+			if err := json.Unmarshal([]byte(*c.Entrypoint), &entrypoint); err != nil {
+				entrypoint = append(entrypoint, ep)
+			}
 		}
+		s.Entrypoint = entrypoint
 	}
-
 	var command []string
-
-	s.Entrypoint = entrypoint
 
 	// Build the command
 	// If we have an entry point, it goes first
-	if len(entrypoint) > 0 {
+	if c.Entrypoint != nil {
 		command = entrypoint
 	}
 	if len(inputCommand) > 0 {
@@ -386,9 +386,12 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 		userCommand = append(userCommand, inputCommand...)
 	}
 
-	if len(inputCommand) > 0 {
+	switch {
+	case len(inputCommand) > 0:
 		s.Command = userCommand
-	} else {
+	case c.Entrypoint != nil:
+		s.Command = []string{}
+	default:
 		s.Command = command
 	}
 
--- a/pkg/specgen/generate/container_create.go
+++ b/pkg/specgen/generate/container_create.go
@@ -130,7 +130,7 @@ func createContainerOptions(rt *libpod.Runtime, s *specgen.SpecGenerator, pod *l
 		options = append(options, libpod.WithName(s.Name))
 	}
 	if pod != nil {
-		logrus.Debugf("adding container to pod %s", pod.Name)
+		logrus.Debugf("adding container to pod %s", pod.Name())
 		options = append(options, rt.WithPod(pod))
 	}
 	destinations := []string{}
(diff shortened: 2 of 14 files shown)


pkg/specgen/generate/container_create.go
```diff
@@ -24,11 +25,28 @@ func MakeContainer(ctx context.Context, rt *libpod.Runtime, s *specgen.SpecGener
 	// If joining a pod, retrieve the pod for use.
 	var pod *libpod.Pod
 	if s.Pod != "" {
-		foundPod, err := rt.LookupPod(s.Pod)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error retrieving pod %s", s.Pod)
+		podName := s.Pod
```
mheon, line 28: Should handle this on the frontend, not the backend, IMO - if we find the "new" sigil, we make a pod first, then make a SpecGenerator with the new pod's ID

giuseppe: won't that cause all clients to implement the same logic?  Also, potentially the operation could be done atomically in the backend or at least have better cleanup on errors

mheon: I view it as a call to the existing `pod create` API with everything defaulted except name, which shouldn't be extra code.

As for atomicity... Hmm. I don't think we actually handled that in Podmanv1, so we're not obligated to, but that does not mean we can't be better.

My main objection is that I really don't want to have to parse special sigils or other command line artifacts in Specgen - I wanted the input in Specgen to be interpreted literally, no special processing, wherever possible for simplicity.

@baude @vrothberg @rhatdan Thoughts on this one?

vrothberg: I do not have strong feelings but share the general desire to keep the CLI specifics out. At some point, Podman and libpod will have separate homes and a CLI-only change should not require touching the libpod project.

rhatdan: Yes I prefer this on the frontend.

giuseppe: I've moved the handling of `new:` to `cmd/podman` and pushed a new version

giuseppe:
@mheon @mheon @rhatdan  PTAL

cmd/podman/common/specgen.go
```diff
@@ -364,20 +364,20 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
 	}
-
 	var command []string
 
-	s.Entrypoint = entrypoint
-
 	// Build the command
 	// If we have an entry point, it goes first
-	if len(entrypoint) > 0 {
+	if c.Entrypoint != nil {
```
mheon, line 380: Why are we building the command here? It seems like we should be unconditionally setting s.Command to userCommand?

giuseppe: I've needed it to differentiate between an empty entrypoint and an unspecified entrypoint.

If `entrypoint == nil` then it wasn't specified.

test/e2e/start_test.go
```diff
@@ -83,7 +82,6 @@ var _ = Describe("Podman start", func() {
 		if podmanTest.RemoteTest {
 			Skip("Container-start name check doesn't work on remote client. It always returns the full ID.")
 		}
-		Expect(session.OutputToString()).To(Equal(name))
```
mheon, line 86: Why are we dropping this? I think this still seems like something we want to pass?

giuseppe: we are printing the ID instead of the short name, now the output looks like:

```
podman start foo
98ea6e4f216f2fb4b69fff9b3a44842c38686ca685f3f55dc48c5d3fb1107be4
```

I can drop this patch but is it a requirement to print the short name?  How would we do it?  Doing the reverse mapping in the frontend?

mheon: I think the key here is that we print the input that the user gave - so if the user typed `foo` we print `foo`, if they typed `98ea` we print `98ea`

giuseppe: how would we handle multiple containers?

`podman start foo1 foo2 foo3`..

Do we have any guarantee the result keeps the same order?

mheon: Hmmm. Back in v1, we had the loop converting shortname -> container in main, so we could easily look this up... @jwhonce @baude Thoughts here?

baude: this has plagued me a little while now ... i would prefer full id but i think docker does names? the problem is with the new approach we loose context of whether we were passed a name vs an ID so we also dont know how to respond?

giuseppe: yes, Docker returns the name you've passed as argument and probably this is the reason why we have added this test.

Possible solutions if we want to keep the current behavior:
- in addition to the ID, pass back the name specified (quite ugly)
- enforce the returned list to keep the same order
- retrieve the ID from the client (do a lookup for each input)
- don't care about Docker, `create`, `run` and `stop` already print the full ID

I like the last two options more, we don't need to change the API and if the ID lookup turns out to be expensive we can drop it/optimize later.

@vrothberg any thought?

vrothberg: I checked Docker 19.03.8 which always returns what the user specified:
* start foo -> foo
* start 123 -> foo
* start $fullID -> $fullID

As pointed out above, if multiple containers are specified we need to know which containers have been started: `$ podman start 1 2 3 doesNotExist 4` ... we need to know that `doesNotExist` could not be started.

While I am not 100 percent sure if we could extract that data already, I suggest to make things a bit more explicit and embed the user input in the return type. We could extend `ContainerStartReport` with `RawInput` (or sth similar): https://github.com/containers/libpod/blob/master/pkg/domain/entities/containers.go#L228

The output in `cmd/...` could then be done by iterating over the report slice:
```golang
errors := []
for _, r := range startReports {
 if r.Err != nil { // append to errors and continue }
 if r.RawInput == "" { // LATEST
   fmt.Println(r.ID)
 }  else 
   fmt.Println(r.RawInput)
  }
}
```

baude:
lgtm

Issue #5138, opened 2020-02-09
Source IP always 127.0.0.1 in rootless Podman 1.8.0


basvdlei:
/kind bug

**Description**

For a rootless container the source IP of incoming packets on a publish port is always `127.0.0.1`. Even if the request is made from an external host.

**Steps to reproduce the issue:**

1. Start a NGINX container:

```
machine-1$ podman run -p 8888:80 docker.io/library/nginx:latest
```

2. Make a request from another node.

```
machine-2$ curl http://machine-1:8888
```

3. Look at the source ip of the request in NGINX stdout log:

```
127.0.0.1 - - [09/Feb/2020:21:54:17 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.66.0" "-"
```

**Describe the results you received:**

The logged source address is always 127.0.0.1

**Describe the results you expected:**

The logged source ip address to match the ip of the host the request was coming from.

**Additional information you deem important (e.g. issue happens only occasionally):**

In Podman 1.7 this worked as expected. And it's probably related to:

> Rootless Podman now uses Rootlesskit for port forwarding, which should greatly improve performance and capabilities

**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.6
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.6
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.10-2.fc31.x86_64
    path: /usr/libexec/crio/conmon
    version: 'conmon version 2.0.10, commit: 6b526d9888abb86b9e7de7dfdeec0da98ad32ee0'
  Distribution:
    distribution: fedora
    version: "31"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 239222784
  MemTotal: 16487555072
  OCIRuntime:
    name: crun
    package: crun-0.12.1-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.12.1
      commit: df5f2b2369b3d9f36d175e1183b26e5cee55dd0a
[...]

PR #6025, opened 2020-04-28
allow switching of port-forward approaches when rootless using slirp4netns

aleks-mariusz:
As of podman 1.8.0, because of commit da7595a, the default approach of providing port-forwarding in rootless mode has switched (and been hard-coded) to rootlessport, for the purpose of providing super performance.  The side-effect of this switch is source within the container to the port-forwarded service always appears to originate from 127.0.0.1 (see issue #5138).

This commit allows a user to specify if they want to revert to the previous approach
of leveraging slirp4netns add_hostfwd() api which, although not as stellar performance,
restores usefulness of seeing incoming traffic origin IP addresses.

The change should be transparent; when not specified, rootlessport will continue to be
used, however if specifying `--net slirp4netns:slirplisten` the old approach will be used.

Testing performed in rootless mode of relevant functionality:

<details><summary>Default behaviour scenario 1 (nothing specified, so not taking advantage of this commit original behaviour (higher performance rootlessport in use))</summary>

```
[test@test ~]$ podman run -d --name nginx -p 8080:80 docker.io/library/nginx:alpine
c0644bcd50c950a000305d0e548e3b1c40732af5c1f9bcbe1dffba244fc1a048
[test@test ~]$ podman logs -f nginx
127.0.0.1 - - [28/Apr/2020:18:19:47 +0000] "GET / HTTP/1.1" 304 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.122 Safari/537.36" "-"
```
</details>
<details><summary>Default behaviour scenario 2 (specifying just --net slirp4netns, which is default when in rootless mode)</summary>

```
[test@test ~]$ podman run -d --name nginx -p 8080:80 --net slirp4netns docker.io/library/nginx:alpine
82e93b571cb013b05f586097544c081c78335f932a5fae071d27ca0f24014cf0
[test@test ~]$ podman logs -f nginx
127.0.0.1 - - [28/Apr/2020:18:20:10 +0000] "GET / HTTP/1.1" 304 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.122 Safari/537.36" "-"
```
</details>
<details><summary>Restored (old, pre-1.8.0) behaviour, where source IP show true origin of port-forwarded traffic.</summary>

```
[test@test ~]$ podman run -d --name nginx -p 8080:80 --net slirp4netns:slirplisten docker.io/library/nginx:alpine
4c5f517d20df3bfc8767371cfefbe87835eb0a0e4e6c300c24b306c80fd3f59f
[test@test ~]$ podman logs -f nginx
192.168.0.22 - - [28/Apr/2020:18:19:14 +0000] "GET / HTTP/1.1" 304 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.122 Safari/537.36" "-"
```
</details>

Note: the above may imply the restored port-forwarding via slirp4netns is not as performant as the new rootlessport approach, however the figures shared in the [original commit](https://github.com/containers/libpod/commit/da7595a69fc15d131c9d8123d0a165bdde4232b6?diff=split) that introduced rootlessport are as follows: `socat: 5.2 Gbps, slirp4netns: 8.3 Gbps, RootlessKit: 27.3 Gbps`, which are more than sufficient for many use cases where the origin of traffic is more important than limits that cannot be reached due to bottlenecks elsewhere.

`Signed-off-by: Aleks Mariusz <m.k@alek.cx>`

### DIFF
--- a/libpod/container_validate.go
+++ b/libpod/container_validate.go
@@ -66,12 +66,9 @@ func (c *Container) validate() error {
 
 	// Rootless has some requirements, compared to networks.
 	if rootless.IsRootless() {
-		if len(c.config.Networks) > 0 {
+		if !c.config.NetMode.IsSlirp4netns() && len(c.config.Networks) > 0 {
 			return errors.Wrapf(define.ErrInvalidArg, "cannot join CNI networks if running rootless")
 		}
-
-		// TODO: Should we make sure network mode is set to Slirp if set
-		// at all?
 	}
 
 	// Can only set static IP or MAC is creating a network namespace.
--- a/pkg/namespaces/namespaces.go
+++ b/pkg/namespaces/namespaces.go
@@ -17,7 +17,9 @@ const (
 	nsType        = "ns"
 	podType       = "pod"
 	privateType   = "private"
+	rlkFwdType    = "port_handler=rootlesskit"
 	shareableType = "shareable"
+	slirpFwdType  = "port_handler=slirplisten"
 	slirpType     = "slirp4netns"
 )
 
@@ -380,7 +382,31 @@ func (n NetworkMode) IsBridge() bool {
 
 // IsSlirp4netns indicates if we are running a rootless network stack
 func (n NetworkMode) IsSlirp4netns() bool {
-	return n == slirpType
+	return n == slirpType || strings.HasPrefix(string(n), slirpType+":")
+}
+
+// IsPortForwardViaRootlessKit indicates if we are doing rootless port-forwarding via rootlesskit/rootlessport
+func (n NetworkMode) IsPortForwardViaRootlessKit() bool {
+	if !n.IsSlirp4netns() {
+		return false
+	}
+	// below here, implied IsSlirp4netns() == true
+	if !strings.Contains(string(n), ":") {
+		return true // default type
+	}
+	// below here, also contains(":") == true
+	parts := strings.SplitN(string(n), ":", 2)
+	return len(parts) == 2 && parts[1] == rlkFwdType
+}
+
+// IsPortForwardViaSlirpHostFwd indicates if we are doing rootless port-forwarding via slirp4netns add_hostfwd()
+func (n NetworkMode) IsPortForwardViaSlirpHostFwd() bool {
+	if !n.IsSlirp4netns() {
+		return false
+	}
+	// below here, implied IsSlirp4netns() == true
+	parts := strings.SplitN(string(n), ":", 2)
+	return len(parts) > 1 && parts[1] == slirpFwdType
 }
 
 // IsNS indicates a network namespace passed in by path (ns:<path>)
--- a/cmd/podman/cliconfig/config.go
+++ b/cmd/podman/cliconfig/config.go
@@ -708,7 +708,6 @@ type UntagValues struct {
 func GetDefaultConfig() *config.Config {
 	var err error
 	conf, err := config.NewConfig("")
-	conf.CheckCgroupsAndAdjustConfig()
 	if err != nil {
 		logrus.Errorf("Error loading container config %v\n", err)
 		os.Exit(1)
--- a/cmd/podman/main.go
+++ b/cmd/podman/main.go
@@ -126,6 +126,9 @@ func before(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	defaultContainerConfig.Engine.CgroupManager = MainGlobalOpts.CGroupManager
+	defaultContainerConfig.CheckCgroupsAndAdjustConfig()
+
 	if err := setupRootless(cmd, args); err != nil {
 		return err
 	}
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -161,6 +161,19 @@ type slirpFeatures struct {
 	HasEnableSeccomp       bool
 }
 
+type slirp4netnsCmdArg struct {
+	Proto     string `json:"proto,omitempty"`
+	HostAddr  string `json:"host_addr"`
+	HostPort  int32  `json:"host_port"`
+	GuestAddr string `json:"guest_addr"`
+	GuestPort int32  `json:"guest_port"`
+}
+
+type slirp4netnsCmd struct {
+	Execute string            `json:"execute"`
+	Args    slirp4netnsCmdArg `json:"arguments"`
+}
+
 func checkSlirpFlags(path string) (*slirpFeatures, error) {
 	cmd := exec.Command(path, "--help")
 	out, err := cmd.CombinedOutput()
@@ -214,6 +227,11 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) (err error) {
 	}
 	if slirpFeatures.HasEnableSeccomp {
 		cmdArgs = append(cmdArgs, "--enable-seccomp")
+	}
+	var apiSocket string
+	if havePortMapping && ctr.config.NetMode.IsPortForwardViaSlirpHostFwd() {
+		apiSocket = filepath.Join(ctr.runtime.config.Engine.TmpDir, fmt.Sprintf("%s.net", ctr.config.ID))
+		cmdArgs = append(cmdArgs, "--api-socket", apiSocket)
 	}
 
 	// the slirp4netns arguments being passed are describes as follows:
@@ -279,7 +297,11 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) (err error) {
 	}
 
 	if havePortMapping {
-		return r.setupRootlessPortMapping(ctr, netnsPath)
+		if ctr.config.NetMode.IsPortForwardViaSlirpHostFwd() {
+			return r.setupRootlessPortMappingViaSlirp(ctr, cmd, apiSocket)
+		} else {
+			return r.setupRootlessPortMappingViaRLK(ctr, netnsPath)
+		}
 	}
 	return nil
 }
@@ -330,7 +352,7 @@ func waitForSync(syncR *os.File, cmd *exec.Cmd, logFile io.ReadSeeker, timeout t
 	return nil
 }
 
-func (r *Runtime) setupRootlessPortMapping(ctr *Container, netnsPath string) (err error) {
+func (r *Runtime) setupRootlessPortMappingViaRLK(ctr *Container, netnsPath string) (err error) {
 	syncR, syncW, err := os.Pipe()
 	if err != nil {
 		return errors.Wrapf(err, "failed to open pipe")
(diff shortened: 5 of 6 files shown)


libpod/container_validate.go
```diff
@@ -66,12 +66,9 @@ func (c *Container) validate() error {
 
 	// Rootless has some requirements, compared to networks.
 	if rootless.IsRootless() {
-		if len(c.config.Networks) > 0 {
+		if !c.config.NetMode.IsSlirp4netns() && len(c.config.Networks) > 0 {
```
mheon, line 69: Why are we changing this? CNI networks will never be usable with rootless Podman's current network stack.

rhatdan:
@aleks-mariusz Thanks for the PR, You need to sign your commit.  BTW Did this already get merged upstream in the master branch.
@giuseppe PTAL

pkg/namespaces/namespaces.go
```diff
@@ -380,7 +382,31 @@ func (n NetworkMode) IsBridge() bool {
...
+// IsPortForwardViaRootlessKit indicates if we are doing rootless port-forwarding via rootlesskit/rootlessport
+func (n NetworkMode) IsPortForwardViaRootlessKit() bool {
+        if !n.IsSlirp4netns() {
+		return false
+	} else { // below here, implied IsSlirp4netns() == true
+		if !strings.Contains(string(n), ":") {
+			return true // default type
+		} else { // below here, also contains(":") == true
+			parts := strings.SplitN(string(n), ":", 2)
+			return parts[1] == rlkFwdType
```
mheon, line 397: Need to verify that len(parts) == 2 else this could segfault

libpod/container_validate.go
aleks-mariusz: I changed this because otherwise i cannot use the ":" character when using rootless mode, it thinks i'm trying to join CNI networks which this validation prohibits when rootless.

If there's a better way to enable this functionality, please let me know.

pkg/namespaces/namespaces.go
aleks-mariusz: just for my clarification, should i be checking if it's _exactly_ length 2? or is the fact that this statement falls on the other side of the if statement _else_ clause, where it definitely has to contain a ":" not sufficient?

aleks-mariusz:
> @aleks-mariusz Thanks for the PR, You need to sign your commit. 

I updated the PR to add a sign-off at the end of the description, is that enough or do i need to amend the comment itself to include a sign-off as well?

> BTW Did this already get merged upstream in the master branch.

I was asked to base this off v1.9 branch, so i create the PR against that as the source/target branch

pkg/namespaces/namespaces.go
```diff
@@ -17,7 +17,9 @@ const (
 	nsType        = "ns"
 	podType       = "pod"
 	privateType   = "private"
+	rlkFwdType    = "rootlessport"
```
AkihiroSuda, line 20: rootlessport -> rootlesskit?

AkihiroSuda:
FYI this package has same interface as rootlessport and can be used for simplifying code https://github.com/rootless-containers/rootlesskit/blob/master/pkg/port/slirp4netns/slirp4netns.go

pkg/namespaces/namespaces.go
aleks-mariusz: this was the example parameter [that @giuseppe used](https://github.com/containers/libpod/issues/5138#issuecomment-618243054).. can change it if you want the parameter to look different

aleks-mariusz:
> FYI this package has same interface as rootlessport and can be used for simplifying code https://github.com/rootless-containers/rootlesskit/blob/master/pkg/port/slirp4netns/slirp4netns.go

i simply used the code that was removed in da7595a on advice in #5138 as it was the easiest way to restore source-ip functionality (and it works).

AkihiroSuda:
Is this going to be `--network-opt`?
https://github.com/containers/libpod/issues/6064#issuecomment-623474633

mheon:
I think it's a cleaner approach than trying to add further parsing to `--network`.

`--network slirp4netns --network-opt port_forward=slirp4netns` and `--network slirp4netns --network-opt port_forward=rootlesskit` seem pretty reasonable.

We'll need a new field (probably map[string]string) in Container configuration to hold it, and we can check it during network setup; if not set, we just use the default or rootlesskit.

giuseppe:
> I think it's a cleaner approach than trying to add further parsing to `--network`.
> 
> `--network slirp4netns --network-opt port_forward=slirp4netns` and `--network slirp4netns --network-opt port_forward=rootlesskit` seem pretty reasonable.

would it be possible in future to have multiple networks?  With separate flags it will be more difficult, while having a single `--network=TYPE:OPTIONS` would make it simpler.

mheon:
I don't know - I don't think we have any ability to do that right now (`--network` can only be passed once, and we'd have to rework how we store it in the DB if we wanted to change that)

5eraph:
Well what about both: 
- keep `--network` as is  
- `--network-opt` will optionally accept network scope (global scope by default)

So sample parameter of slirp4nets outbound address would be 
`--network-opt=slirp4netns:OUTBOUND_ADDR=127.0.0.1`
This should be backward compatible with current `--network` and we can transition without significant changes later.

giuseppe:
> I don't know - I don't think we have any ability to do that right now (`--network` can only be passed once, and we'd have to rework how we store it in the DB if we wanted to change that)

true but that will be an internal change, while now we are introducing a CLI change that will be difficult to change later.

> Well what about both:
> 
>     * keep `--network` as is
> 
>     * `--network-opt` will optionally accept network scope (global scope by default)
> 
> 
> So sample parameter of slirp4nets outbound address would be
> `--network-opt=slirp4netns:OUTBOUND_ADDR=127.0.0.1`
> This should be backward compatible with current `--network` and we can transition without significant changes later.

it is possible to add multiple slirp4netns networks to the same container

aleks-mariusz:
Would it be of any value if we perhaps _also_ had an environment variable be handled in order to make this change take effect?  

I'm asking because the way i am launching podman where i really need this source-ip preserved functionality is via [podman-compose](https://github.com/containers/podman-compose), which does **not** currently support any `--network` parameter being read from the docker-compose.yml nor ever passed (not to mention it defines a pod and adds the container, so i'm not sure how the above support i added fits into that), perhaps an environment variable would be better suited to that purpose?

AkihiroSuda:
or libpod.conf?

giuseppe:
yes it should go through `containers.conf`.  I believe it already works if we go with `--network=`

aleks-mariusz:
libpod is an interesting idea, i didn't even consider that.. but won't that mean that every container will in-effect be impacted/affected by the change if it's a global (or per user) level?  is that desirable?

TBH, i'm not a _huge_ fan of the environment variable idea (it's a bit too easy to lose hours off debugging something where it turns out you set (or didn't set) an environment variable and forgot), but it was the most immediate thing that came into my mind that worked around pod definitions not supporting --network

giuseppe:
> the config file approach is an interesting idea, i didn't even consider that.. but won't that mean that every container will in-effect be impacted/affected by the change if it's a global (or per user) level? is that desirable?

that can be used exactly in the case you want to change the default.  For a single container, `--network` should be enough.  I think we shouldn't be abusing env variables more than we already do :-)

rhatdan:
/usr/share/containers/containers.conf and /etc/containers/containers.conf are global.
~/.config/containers/containers.conf is local.

aleks-mariusz:
i can take a stab at adding it to this PR if folks are happy with what i have in here so far looks good.. 

also what should i name the parameter (presuming it'd go under the `[network]` section) in the containers.conf? i am thinking `rootless_portforward_mechanism` but i tend to be a bit overly verbose :-)

mheon:
Maybe `rootless_port_handler`?

giuseppe:
> Maybe `rootless_port_handler`?

could it be `port_handler` or `slirp_port_handler`?  In theory slirp can be used from root as well

AkihiroSuda:
s/slirp_port_handler/port_handler/ because `rootlesskit` isn't really slirp. Also, future version of rootless Podman may support non-slirp network via some privileged helper like `lxc-user-nic`.

rhatdan:
@aleks-mariusz ping

aleks-mariusz:
@rhatdan  Will try to look at giuseppe's review this weekend and maybe put the config file handling in as well

aleks-mariusz:
@giuseppe i've added your review feedback to this PR now.. 

I'm going to hold off from trying to tackle the config file handling as i'm not quite sure how to this is handled (and may not be handled in this project but in [containers/common](https://github.com/containers/common/blob/master/pkg/config/config.go#L365-L376) )

Can you (or anyone) add the config file handling to this PR please (or guide me how to)

giuseppe:
> Can you (or anyone) add the config file handling to this PR please (or guide me how to)

I think we can tackle that separately as another PR.

The new code LGTM, could you just fix the golint issues reported here: https://github.com/containers/libpod/pull/6025/checks?check_run_id=660593109  ?

giuseppe:
I think the next two patches change code introduced by the first one, can you please squash the patches in one?  After that, I think it is ready to go

disaster123:
@aleks-mariusz i merged your pull request into v1.9 branch and specified --net slirp4netns:slirplisten for my rootless container. But it does not seem to work:
```
inside container:
 tcpdump -i any udp and port 10000
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on any, link-type LINUX_SLL (Linux cooked), capture size 262144 bytes

18:12:37.226011 IP localhost.47012 > localhost.10000: UDP, length 104
18:12:37.421069 IP localhost.49500 > localhost.10000: UDP, length 104
18:12:37.620377 IP localhost.51852 > localhost.10000: UDP, length 104
```

outside container:
```
tcpdump  -i any udp and port 10000
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on any, link-type LINUX_SLL (Linux cooked), capture size 262144 bytes
19:59:19.726830 IP ip4d140XXX.dynamic.kabel-deutschland.de.56728 > myserver.de.10000: UDP, length 104
19:59:20.117883 IP ip4d140XXX.dynamic.kabel-deutschland.de.56728 > myserver.de.10000: UDP, length 104
```

aleks-mariusz:
@giuseppe is there anything else needed to merge this? i would really like the config file support to be written as well as even this is a bit cumbersome to have to use and being able to use a config file would be a preferable/more global change please

giuseppe:
it looks fine, thanks.

Can you also open a PR for master?  I'd like that the feature goes first in master, and then in 1.9

aleks-mariusz:
> it looks fine, thanks.
> 
> Can you also open a PR for master? I'd like that the feature goes first in master, and then in 1.9

@giuseppe PR #6324 created

aleks-mariusz:
implemented w/ Giuseppe's help via https://github.com/containers/podman/pull/6965

PR #6028, opened 2020-04-28
User specified environment after other environments are set

rhatdan:
Users can not currently override the environment variables set by
--http-proxy

### DIFF
--- a/pkg/spec/spec.go
+++ b/pkg/spec/spec.go
@@ -326,10 +326,6 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 		}
 		defaultEnv = env.Join(env.DefaultEnvVariables, defaultEnv)
 	}
-	config.Env = env.Join(defaultEnv, config.Env)
-	for name, val := range config.Env {
-		g.AddProcessEnv(name, val)
-	}
 
 	if err := addRlimits(config, &g); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes: https://github.com/containers/libpod/issues/6017

rhatdan:
@mheon This works correctly in master.

$ export http_proxy=http://2.2.2.2:2222; ./bin/podman run -e http_proxy=http://1.1.1.1:1111 alpine printenv http_proxy
http://1.1.1.1:1111

pkg/spec/spec.go
```diff
@@ -360,6 +356,11 @@ func (config *CreateConfig) createConfigToOCISpec(runtime *libpod.Runtime, userM
 	if err := config.Cgroup.ConfigureGenerator(&g); err != nil {
 		return nil, err
 	}
+
+	config.Env = env.Join(defaultEnv, config.Env)
+	for name, val := range config.Env {
+		g.AddProcessEnv(name, val)
+	}
```
TomSweeneyRedHat, line 363: It might be good to add an inline comment here to explain why it's here to make sure it doesn't get moved up again inadvertently.

rhatdan: This is a back port.  And in code that is not used in the new version. I added a test for this situation in the master branch pull.  I think we need more tests to make sure that environment handling is correct.

mheon:
LGTM, but we're gonna miss 1.9.1

rhatdan:
@mheon Yup we will leave it in the hopper for podman 1.9.2

TomSweeneyRedHat:
LGTM

PR #6029, opened 2020-04-28
User specified environment happen after other environments are set

rhatdan:
When using varlink we want to make sure that user specified environment variables
take precedence over http-proxy environment.


rhatdan:
This PR adds a test to make sure we don't regress on this.
https://github.com/containers/libpod/issues/6017

mheon:
LGTM

PR #6035, opened 2020-04-29
rootless: move ns open before fork

giuseppe:
commit 788fdc685b00dee5ccb594bef845204250c4c123 introduced a race
where the target process dies before the child process opens the
namespace files.  Move the open before the fork so if it fails the
parent process can attempt to join a different container instead of
failing.


giuseppe:
I'll have to backport to 1.9

rhatdan:
We have already started a podman 1.9.2

mheon:
Are we backporting this to 1.9.x? I don't see a PR for it

PR #6037, opened 2020-04-29
Enable push tests

vrothberg:
A couple of fixes to make push work. Mostly related to parsing and translating --tls-verify.

### DIFF
--- a/pkg/bindings/images/images.go
+++ b/pkg/bindings/images/images.go
@@ -311,8 +311,9 @@ func Push(ctx context.Context, source string, destination string, options entiti
 	params.Set("credentials", options.Credentials)
 	params.Set("destination", destination)
 	if options.SkipTLSVerify != types.OptionalBoolUndefined {
-		val := bool(options.SkipTLSVerify == types.OptionalBoolTrue)
-		params.Set("tlsVerify", strconv.FormatBool(!val))
+		// Note: we have to verify if skipped is false.
+		verifyTLS := bool(options.SkipTLSVerify == types.OptionalBoolFalse)
+		params.Set("tlsVerify", strconv.FormatBool(verifyTLS))
 	}
 
 	path := fmt.Sprintf("/images/%s/push", source)
--- a/pkg/bindings/images/images.go
+++ b/pkg/bindings/images/images.go
@@ -310,9 +310,9 @@ func Push(ctx context.Context, source string, destination string, options entiti
 	params := url.Values{}
 	params.Set("credentials", options.Credentials)
 	params.Set("destination", destination)
-	if options.TLSVerify != types.OptionalBoolUndefined {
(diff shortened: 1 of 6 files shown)


vrothberg:
@baude @giuseppe @jwhonce @mheon @rhatdan PTAL

pkg/bindings/images/images.go
```diff
@@ -310,9 +310,9 @@ func Push(ctx context.Context, source string, destination string, options entiti
 	params := url.Values{}
 	params.Set("credentials", options.Credentials)
 	params.Set("destination", destination)
-	if options.TLSVerify != types.OptionalBoolUndefined {
-		val := bool(options.TLSVerify == types.OptionalBoolTrue)
-		params.Set("tlsVerify", strconv.FormatBool(val))
+	if options.SkipTLSVerify != types.OptionalBoolUndefined {
+		val := bool(options.SkipTLSVerify == types.OptionalBoolTrue)
```
rhatdan, line 314: This confused me, I think it would be better to do 

val := bool(options.SkipTLSVerify != types.OptionalBoolTrue)
params.Set("tlsVerify", strconv.FormatBool(val))

In my small brain, this is more understandable.

vrothberg: I understand why your brain parser stopped here. I simplified it a bit further:
```diff
-               val := bool(options.SkipTLSVerify == types.OptionalBoolTrue)
-               params.Set("tlsVerify", strconv.FormatBool(!val))
+               // Note: we have to verify if skipped is false.
+               verifyTLS := bool(options.SkipTLSVerify == types.OptionalBoolFalse)
+               params.Set("tlsVerify", strconv.FormatBool(verifyTLS))
```

vrothberg:
Fixed

mheon:
LGTM

pkg/domain/entities/images.go
```diff
@@ -183,8 +183,8 @@ type ImagePushOptions struct {
 	// SignBy adds a signature at the destination using the specified key.
 	// Ignored for remote calls.
 	SignBy string
-	// TLSVerify to enable/disable HTTPS and certificate verification.
-	TLSVerify types.OptionalBool
+	// SkipTLSVerify to skip HTTPS and certificate verification.
+	SkipTLSVerify types.OptionalBool
```
TomSweeneyRedHat, line 187: Don't we have a `--tlsverify` and not a `--skiptlsverify` option?  This var name now seems to negate that to me.  Maybe it's my turn to be confused?

vrothberg: The CLI option is parsed and set to another variable (not this one). c/image is using the skip semantics. We have this logic inverted with other commands as well (all leading to failing --tls-verify).

I opted for making the Skip logic consistent in the code. Will hopefully prevent us from doing too much brain yoga.

PR #6039, opened 2020-04-29
Enable search tests

vrothberg:
* fix up the tls-verify hiccups as done for push
* clean up a couple of search things
* enable search tests

### DIFF
--- a/cmd/podman/images/pull.go
+++ b/cmd/podman/images/pull.go
@@ -93,23 +93,22 @@ func pullFlags(flags *pflag.FlagSet) {
 
 // imagePull is implement the command for pulling images.
 func imagePull(cmd *cobra.Command, args []string) error {
-	pullOptsAPI := pullOptions.ImagePullOptions
 	// TLS verification in c/image is controlled via a `types.OptionalBool`
 	// which allows for distinguishing among set-true, set-false, unspecified
 	// which is important to implement a sane way of dealing with defaults of
 	// boolean CLI flags.
 	if cmd.Flags().Changed("tls-verify") {
-		pullOptsAPI.TLSVerify = types.NewOptionalBool(pullOptions.TLSVerifyCLI)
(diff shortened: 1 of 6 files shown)


vrothberg:
@baude @jwhonce @mheon @rhatdan @TomSweeneyRedHat PTAL

rhatdan:
LGTM

cmd/podman/images/pull.go
```diff
@@ -93,23 +93,22 @@ func pullFlags(flags *pflag.FlagSet) {
...
 // imagePull is implement the command for pulling images.
 func imagePull(cmd *cobra.Command, args []string) error {
-	pullOptsAPI := pullOptions.ImagePullOptions
 	// TLS verification in c/image is controlled via a `types.OptionalBool`
 	// which allows for distinguishing among set-true, set-false, unspecified
 	// which is important to implement a sane way of dealing with defaults of
 	// boolean CLI flags.
 	if cmd.Flags().Changed("tls-verify") {
-		pullOptsAPI.TLSVerify = types.NewOptionalBool(pullOptions.TLSVerifyCLI)
+		pullOptions.SkipTLSVerify = types.NewOptionalBool(!pullOptions.TLSVerifyCLI)
```
TomSweeneyRedHat, line 101: Maybe I've just gotten used to it over the years, but I'd rather not negate this, personal bias me thinks and ¯\_(ツ)_/¯

vrothberg: I completely share your feeling. The trouble here is that we have to negate it at some point. c/image has a skip logic while the CLI has a "do" logic. As we regressed in the backend, I moved the negation here (CLI only) in hope to never touch it again. The backend will likely more often be touched, so I think it's good to have a type/logic that follows that c/image expects rather than negating there.

TomSweeneyRedHat:
LGTM assuming happy tests

PR #5233, opened 2020-02-17
fix mandatory parameter in login/logout

QiWang19:
fix #5146
Insted of using a registry as mandatory parameter, this path allows podman to use the first registry from registries.conf.

Signed-off-by: Qi Wang <qiwan@redhat.com>

PR #6041, opened 2020-04-29
System tests: help messages: check required-arg

edsantiago:
If a usage message is of the form '... [flags] ARGNAME',
where ARGNAME is all-caps and not in brackets, it must
be a required argument. Try running podman subcommand
without ARGNAME, and make sure that podman bails out
with an informative message. (Since this message is
freeform in each subcommand, not Cobra-generated,
we have a lot of possible variations to check for).

Fix podman login/logout Use messages to indicate that
REGISTRY is now optional (as of #5233).

This test has actually been in place for over a year but
due to a typo on my part -- a missing space -- it was
not being run. "For want of a space, much testing was lost".

### DIFF
--- a/cmd/podman/login.go
+++ b/cmd/podman/login.go
@@ -19,7 +19,7 @@ type loginOptionsWrapper struct {
 var (
 	loginOptions = loginOptionsWrapper{}
 	loginCommand = &cobra.Command{
-		Use:   "login [flags] REGISTRY",
+		Use:   "login [flags] [REGISTRY]",
 		Short: "Login to a container registry",
 		Long:  "Login to a container registry on a specified server.",
 		RunE:  login,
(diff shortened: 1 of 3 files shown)


edsantiago:
Oh dear:
```
$ ./bin/podman container attach
panic: runtime error: index out of range [0] with length 0
```
@baude, could this be related to your recent commits here? 

Unfortunately I can no longer build podman on my f30 laptop, so I can't track it down myself. Guess I should try to upgrade to f32 tonight.

mheon:
LGTM once we get things stabilized.

edsantiago:
Aw man this is all over the place. container attach, commit, create, exec, that's just a few minutes of test-edit-test. We need some serious review here on input arg validation.

edsantiago:
Fixed in #6055

PR #6051, opened 2020-04-30
Fixes for test/e2e/containers_conf_test.go

rhatdan:


rhatdan:
@mheon PTAL especially at the namespace rework.

pkg/specgen/namespaces.go
```diff
@@ -215,12 +245,17 @@ func ParseUserNamespace(ns string) (Namespace, error) {
 func ParseNetworkNamespace(ns string) (Namespace, []string, error) {
 	toReturn := Namespace{}
 	var cniNetworks []string
+	// Net defaults to Slirp on rootless
+	if rootless.IsRootless() {
```
mheon, line 249: This completely overrides parsing on rootless, so it'll break rootless `--net=host`

pkg/specgen/namespaces.go
```diff
@@ -215,12 +245,17 @@ func ParseUserNamespace(ns string) (Namespace, error) {
...
+		toReturn.NSMode = Slirp
+		return toReturn, nil, nil
+	}
 	switch {
 	case ns == "slirp4netns":
 		toReturn.NSMode = Slirp
 	case ns == "pod":
 		toReturn.NSMode = FromPod
-	case ns == "bridge":
+	case ns == "bridge", ns == "":
```
mheon, line 258: I think we need a separate `case` for `""` where we check if we are rootless and return `Bridge` or `Slirp` based on that

pkg/specgen/generate/namespaces.go
```diff
@@ -49,51 +48,26 @@ func GetDefaultNamespaceMode(nsType string, cfg *config.Config, pod *libpod.Pod)
...
-		}
-		if !cgroupsv2 {
-			toReturn.NSMode = specgen.Host
-		}
-	default:
-		return toReturn, errors.Wrapf(define.ErrInvalidArg, "invalid namespace type %s passed", nsType)
+		return specgen.ParseCgroupNamespace(cfg.Containers.CgroupNS)
+	case "net":
+		ns, _, err := specgen.ParseNetworkNamespace(cfg.Containers.NetNS)
+		return ns, err
```
TomSweeneyRedHat, line 67: Maybe I'm not following the logic, but don't you need a rootless check here?  Will Slirp still be returned?

pkg/specgen/namespaces.go
rhatdan: Yes, you are right fixed.

pkg/specgen/generate/namespaces.go
rhatdan: Containers.conf takes care of setting the default to slirp4netns or bridge already.  So that should not be a problem.  Matt points out above ,that if for some reason we do not have a conf and the default is "", then we need the rootless check in ParseNetworkNamespace, which I fixed.

rhatdan:
Needs https://github.com/containers/common/pull/131 to be merged first and then this needs to be vendored into buildah.

TomSweeneyRedHat:
LGTM but want a @mheon head nod too.

PR #6054, opened 2020-04-30
CI:DOCS: Document API docs + CORS maintenance

cevich:
Signed-off-by: Chris Evich <cevich@redhat.com>

PR #6055, opened 2020-04-30
    fix commands without input

baude:
in cases where commands require input and we dont provide it, we often would segv.  This can be attributed in many cases to the subcommand not picked up the cobra Args attribute or neither had them.

this includes eds pr #6054

### DIFF
--- a/cmd/podman/containers/attach.go
+++ b/cmd/podman/containers/attach.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/containers/libpod/cmd/podman/registry"
+	"github.com/containers/libpod/cmd/podman/validate"
 	"github.com/containers/libpod/pkg/domain/entities"
 	"github.com/pkg/errors"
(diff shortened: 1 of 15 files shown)


edsantiago:
LGTM. I don't understand the build error; the first thing I did was git-pr this locally and build it.

mheon:
Probably one of the two commits does not build on its own; I'd try squashing them down to one.

LGTM otherwise.

edsantiago:
>Probably one of the two commits does not build on its own

That seems impossible at first glance; and no matter how I look at it it's still impossible. The error message makes no sense. I'm starting to think there's something broken in the build_each_commit test.

edsantiago:
Aha - `build_all_commits` is rebasing onto master, which now has #6047, which relies on `errors`. @baude, you need to rebase onto master and re-add the `pkg/errors` include.

edsantiago:
Oops. Sorry, I meant to run the full BATS suite but got distracted.

PR #6056, opened 2020-04-30
V2 Commands that require ParentNS (rootful) are report error

jwhonce:
* Updated unmount to require ParentNS


baude:
LGTM

PR #6058, opened 2020-04-30
Fix errors found in coverity scan

rhatdan:

### DIFF
--- a/libpod/container_inspect.go
+++ b/libpod/container_inspect.go
@@ -661,7 +661,7 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
 				if ns.Path != "" {
 					networkMode = fmt.Sprintf("ns:%s", ns.Path)
 				} else {
-					networkMode = "none"
+					networkMode = "private"
 				}
 				break
 			}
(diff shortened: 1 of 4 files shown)


libpod/container_inspect.go
```diff
@@ -831,11 +828,12 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
 	usernsMode := ""
 	if c.config.UserNsCtr != "" {
 		usernsMode = fmt.Sprintf("container:%s", c.config.UserNsCtr)
-	} else {
+	} else if ctrSpec.Linux != nil {
```
TomSweeneyRedHat, line 831: Are these if's really coverity issues?

TomSweeneyRedHat:
LGTM, a couple of those were scarier than anticipated.

libpod/container_inspect.go
```diff
@@ -743,22 +743,21 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
...
 			if ns.Type == spec.IPCNamespace {
-				foundIPCNS = true
 				if ns.Path != "" {
 					ipcMode = fmt.Sprintf("ns:%s", ns.Path)
 				}
 				break
 			}
 		}
-		if !foundIPCNS {
+		if ipcMode == "" {
```
mheon, line 760: Is this actually equivalent? I'm trying to work my head around the logic here, and I think it may be setting host when we're actually private?

libpod/container_inspect.go
```diff
@@ -831,11 +828,12 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
 	usernsMode := ""
 	if c.config.UserNsCtr != "" {
 		usernsMode = fmt.Sprintf("container:%s", c.config.UserNsCtr)
-	} else {
+	} else if ctrSpec.Linux != nil {
 		// Locate the spec's user namespace.
 		// If there is none, it's default - the empty string.
 		// If there is one, it's "private" if no path, or "ns:" if
 		// there's a path.
+
```
mheon, line 836: Random newline?

libpod/container_inspect.go
rhatdan: Yes coverity was complaining about the chance the ctrSpec.Linux could be nil and dereferenced later.

libpod/container_inspect.go
rhatdan: Removed

libpod/container_inspect.go
rhatdan: Good point, you actually discovered a bug.  If we have a namespace then it should be "private" else it is host.  So in the loop above we should set ipdMode = "private" if there is no path.

rhatdan:
@mheon @giuseppe @vrothberg @baude @jwhonce @QiWang19 PTAL

QiWang19:
LGTM

TomSweeneyRedHat:
LGTM

PR #6059, opened 2020-04-30
search --limit compatible with docker

QiWang19:
Check --limit range and update --limit manpage explanation.

### DIFF
--- a/docs/source/markdown/podman-search.1.md
+++ b/docs/source/markdown/podman-search.1.md
@@ -59,7 +59,7 @@ Valid placeholders for the Go template are listed below:
 
 **--limit**=*limit*
 
-Limit the number of results
+Limit the number of results. This value can be in the range between 1 and 100. The default number of results is 25.
 Note: The results from each registry will be limited to this value.
 Example if limit is 10 and two registries are being searched, the total
 number of results will be 20, 10 from each (if there are at least 10 matches in each).
--- a/cmd/podman/images/search.go
+++ b/cmd/podman/images/search.go
@@ -105,6 +105,10 @@ func imageSearch(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("search requires exactly one argument")
 	}
 
+	if searchOptions.Limit > 100 {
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM
But it looks like you have broken the tests.  Also do we know what the default is?  Should be mentioned in the the man page.

docs/source/markdown/podman-search.1.md
```diff
@@ -59,7 +59,7 @@ Valid placeholders for the Go template are listed below:
 
 **--limit**=*limit*
 
-Limit the number of results
+Limit the number of results. This value could be in the range between 1 and 100.
```
TomSweeneyRedHat, line 62: nit could -> can

TomSweeneyRedHat:
The other part of this problem that's not solved here is if there are 200 entries in the registry and the registry returns only 100 due to the limit set by the registry itself, regardless of the limit option being used or not used by the user.  So if there are 200 entries in the registry and the user asks for all but only 100 are shown, we need to tell them at the end of the output.  Similar to what Docker does and Derrick put an example in the BZ.

TomSweeneyRedHat:
And fwiw, I'm fine with merging this in the near term and then doing a separate PR for the other warning message.

TomSweeneyRedHat:
LGTM

vrothberg:
@TomSweeneyRedHat @QiWang19 @rhatdan what was the BZ? I just hit this issue (working on another BZ) and do not understand this limit.

TomSweeneyRedHat:
@vrothberg the BZ is https://bugzilla.redhat.com/show_bug.cgi?id=1827794

vrothberg:
I don't understand the hard limit of 100 if it's up to registries to decide.

vrothberg:
@mtrmac WDYT?

mtrmac:
Yeah, that’s not what the bug asked for. On the contrary, if anything, the user should be able to disable the default 25 limit to get _all_ results without any limit.

PR #6061, opened 2020-05-01
check --get-login when login

QiWang19:
Check --get-login is set in podman since it is not shared option from c/common and does not valid by the package.


rhatdan:
LGTM, Not sure what this is for?

QiWang19:
check whether --get-login is set
```
**--get-login**

Return the logged-in user for the registry.  Return error if no login is found.
```

rhatdan:
/approve
@TomSweeneyRedHat @baude @mheon @jwhonce @giuseppe PTAL

TomSweeneyRedHat:
LGTM assuming happy tests

PR #6066, opened 2020-05-01
[CI:DOC]Use full repo name in podmanimage Dockerfiles

TomSweeneyRedHat:
In the Buildah images, we had a problem where the testing image
was installed with an older version of Buildah than the stable
image.  This was apparently due to quay.io using Docker and Dockerhub
which has a version of Fedora that did not let testing
version of Buildah to be installed as it should have been.

This change fully specifies the name of the fedora image to
use.  This has not been a problem in Podman, but I'm carrying
this change here to avoid future problems.


rhatdan:
LGTM
/lgtm

PR #6076, opened 2020-05-04
image removal: refactor part 2

vrothberg:
Continue the refactoring of image removal.  I didn't manage to break all
the following changes into smaller and easier to digest commits due to
time constraints:

 * Return an error slice instead of a single error. Use multierror only
   in the client/frontend.  Reflect that in the types.

 * Use the batch image removal in the client while preserving the more
   rest-idiomatic single-image removal endpoint.

 * Add a new handler for the single-image removal endpoint to make it
   share the same code as the batch endpoint.

 * Expose bindings for the single and batch endpoints, so we can
   properly test them.

 * Add several convenience functions for error handling to
   pkg/errorhandling.

 * Set the correct error type in libpod to set the exit code to 2 when
   one or more containers are using an image.

 * Massage the bindings tests a bit and tackle compilation errors.

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -170,9 +170,9 @@ func run(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 	if runRmi {
-		_, err := registry.ImageEngine().Remove(registry.GetContext(), []string{args[0]}, entities.ImageRemoveOptions{})
-		if err != nil {
-			logrus.Errorf("%s", errors.Wrapf(err, "failed removing image"))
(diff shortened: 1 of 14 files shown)


vrothberg:
I am currently working on the bindings tests and see if I can get the to pass (currently container.Inspect issue).

vrothberg:
Polished the bindings tests a bit. Once the containers bindings tests are passing, I will add a test for the batch-removal bindings.

@baude @jwhonce @lsm5 @mheon @rhatdan PTAL

baude:
LGTM

PR #6077, opened 2020-05-04
Fix errors found when comparing podman v1 --help versus V2

rhatdan:
Mainly add missing commands to podman image, podman containers, podman system

Also fix some informations messages and descriptions.

### DIFF
--- a/cmd/podman/containers/prune.go
+++ b/cmd/podman/containers/prune.go
@@ -18,10 +18,10 @@ import (
 var (
 	pruneDescription = fmt.Sprintf(`podman container prune
 
-	Removes all stopped | exited containers`)
+	Removes all non running containers`)
 	pruneCommand = &cobra.Command{
 		Use:     "prune [flags]",
-		Short:   "Remove all stopped | exited containers",
+		Short:   "Remove all non running containers",
 		Long:    pruneDescription,
 		RunE:    prune,
 		Example: `podman container prune`,
@@ -50,7 +50,7 @@ func prune(cmd *cobra.Command, args []string) error {
 	}
 	if !force {
 		reader := bufio.NewReader(os.Stdin)
-		fmt.Println("WARNING! This will remove all stopped containers.")
+		fmt.Println("WARNING! This will remove all non running containers.")
 		fmt.Print("Are you sure you want to continue? [y/N] ")
 		answer, err := reader.ReadString('\n')
 		if err != nil {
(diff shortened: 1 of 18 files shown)


rhatdan:
@mheon @baude @vrothberg @giuseppe @TomSweeneyRedHat @QiWang19 @jwhonce PTAL

cmd/podman/containers/prune.go
```diff
@@ -18,10 +18,10 @@ import (
 var (
 	pruneDescription = fmt.Sprintf(`podman container prune
 
-	Removes all stopped | exited containers`)
+	Removes all not running containers`)
```
mheon, line 21: Should this be "non running"?

rhatdan: Yup

QiWang19:
LGTM

cmd/podman/containers/prune.go
rhatdan: Fixed

baude:
all aliased commands need to have a completely separate command for them to so that example text is contextually correct.  the team outvoted me on that, so this is not a baude edict.  you can look at some of the image or container commands as to how it should be done.  in those cases, the flags have to be broken out and into a func that handles a pflag.FlagSet ...

rhatdan:
@baude I see the correct help, I am not sure what command you are talking about?
```./bin/podman image import --help
Import a tarball to create a filesystem image

Description:
  Create a container image from the contents of the specified tarball (.tar, .tar.gz, .tgz, .bzip, .tar.xz, .txz).

  Note remote tar balls can be specified, via web address.
  Optionally tag the image. You can specify the instructions using the --change option.

Usage:
  podman image import [flags] PATH [REFERENCE]

Examples:
  podman image import http://example.com/ctr.tar url-image
  cat ctr.tar | podman -q image import --message "importing the ctr.tar tarball" - image-imported
  cat ctr.tar | podman image import -

Flags:
  -c, --change stringArray   Apply the following possible instructions to the created image (default []): CMD | ENTRYPOINT | ENV | EXPOSE | LABEL | STOPSIGNAL | USER | VOLUME | WORKDIR
  -m, --message string       Set commit message for imported image
  -q, --quiet                Suppress output
```

edsantiago:
@rhatdan I don't see any inconsistencies in your addition; the only broken one right now is `podman unpause`, which is not something you touch in this PR. Your PR LGTM.

PR #6080, opened 2020-05-04
v2 podman stats

baude:


cmd/podman/containers/stats.go
```diff
@@ -0,0 +1,244 @@
...
+		Parent:  containerCmd,
+	})
+
+	containerStatsFlags := containerStatsCommand.Flags()
+	statFlags(containerStatsFlags)
+}
+
+// stats is different in that it will assume running containers if
+// no input is given, so we need to validate differently
+func checkStatOptions(cmd *cobra.Command, args []string) error {
```
rhatdan, line 86: This seems like it should be a shared function.

libpod/util_test.go
```diff
@@ -3,14 +3,16 @@ package libpod
 import (
 	"testing"
 
+	"github.com/containers/libpod/utils"
+
```
rhatdan, line 7: Bad empty line.

pkg/domain/infra/abi/pods_stats.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/containers/libpod/utils"
+
```
rhatdan, line 8: Bad empty line

cmd/podman/containers/stats.go
```diff
@@ -0,0 +1,244 @@
...
+var (
+	statsOptions       entities.ContainerStatsOptions
+	defaultStatsRow    = "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\n"
+	defaultStatsHeader = "ID\tNAME\tCPU %\tMEM USAGE / LIMIT\tMEM %\tNET IO\tBLOCK IO\tPIDS\n"
+)
+
+func statFlags(flags *pflag.FlagSet) {
+	flags.BoolVarP(&statsOptions.All, "all", "a", false, "Show all containers. Only running containers are shown by default. The default is false")
+	flags.StringVar(&statsOptions.Format, "format", "", "Pretty-print container statistics to JSON or using a Go template")
+	flags.BoolVarP(&statsOptions.Latest, "latest", "l", false, "Act on the latest container podman is aware of")
```
TomSweeneyRedHat, line 58: podman -> Podman

cmd/podman/containers/stats.go
baude: it is an odd duck one, which is i why handled it this way.  im not sure any other funcs are like this.

pkg/domain/infra/abi/containers.go
```diff
@@ -998,3 +999,84 @@ func (ic *ContainerEngine) Shutdown(_ context.Context) {
...
+			break
+		}
+		//if strings.ToLower(format) != formats.JSONString && !c.NoReset {
+		//	tm.Clear()
+		//	tm.MoveCursor(1, 1)
+		//	tm.Flush()
+		//}
+		//if err := outputStats(reportStats, format); err != nil {
+		//	return nil, err
+		//}
```
giuseppe, line 1078: do we need the dead code here or can we delete it?

utils/utils.go
```diff
@@ -125,3 +126,21 @@ func Tar(source string) (io.ReadCloser, error) {
 	logrus.Debugf("creating tarball of %s", source)
 	return archive.Tar(source, archive.Uncompressed)
 }
+
+// RemoveScientificNotationFromFloat returns a float without any
+// scientific notation if the number has any.
+// golang does not handle conversion of float64s that have scientific
+// notation in them and otherwise stinks.  please replace this if you have
+// a better implementation.
+func RemoveScientificNotationFromFloat(x float64) (float64, error) {
```
giuseppe, line 135: do you have any example that is not handled by `"%.2f"` and it is causing problems?

baude: no, this is what was used before and i didnt care to delete it in case there was a problem...

PR #6088, opened 2020-05-05
BATS help test: check usage string

edsantiago:
Now that we've agreed that usage messages should match
what the user typed, confirm it. IOW 'podman foo --help'
should not issue a usage message for 'podman container foo'.

Fix one broken instance, 'unpause'.


mheon:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests

PR #6096, opened 2020-05-05
Add small fixes for 'podman run' from diffing inspect

mheon:
To try and identify differences between Podman v1.9 and master, I ran a series of `podman run` commands with various flags through each, then inspecting the resulting containers and diffed the inspect JSON between each. This identified a number of issues which are fixed in this PR.

In order of discovery:
- Podman v2 gave short names for images, where Podman v1 gave the fully-qualified name. Simple enough fix (get image tags and use the first one if they're available)
- The --restart flag was not being parsed correctly when a number of retries was specified. Parsing has been corrected.
- The -m flag was not setting the swap limit (simple fix to set swap in that case if it's not explicitly set by the user)
- The --cpus flag was completely nonfunctional (wired in its logic)

Tests have been added for all of these to catch future regressions.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -26,6 +26,16 @@ func getCPULimits(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string) 
 	cpu := &specs.LinuxCPU{}
 	hasLimits := false
 
+	const cpuPeriod = 100000
+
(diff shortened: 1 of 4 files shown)


mheon:
Good to merge. @baude @rhatdan @vrothberg PTAL

vrothberg:
@giuseppe PTAL

PR #6100, opened 2020-05-06
podman: split env variables in env and overrides

giuseppe:
There are three different priorities for applying env variables:

1) environment/config file environment variables
2) image's config
3) user overrides (--env)

The first and third kind are known to the client, while the image's
config is handled by the backend.

We need to carry the variables having different priorities in
different maps, so the backend can correctly build the final
environment configuration.

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -156,8 +156,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	createFlags.String("entrypoint", "",
 		"Overwrite the default ENTRYPOINT of the image",
 	)
-	createFlags.StringArrayVarP(
-		&cf.env,
+	createFlags.StringArrayP(
 		"env", "e", containerConfig.Env(),
 		"Set environment variables in container",
 	)
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -299,6 +299,14 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 		"container": "podman",
 	}
 
+	if defaultEnv, err := envLib.ParseSlice(containerConfig.GetDefaultEnv()); err != nil {
+		return errors.Wrap(err, "error parsing default env variables")
+	} else {
+		for k, v := range defaultEnv {
+			env[k] = v
+		}
+	}
+
 	// First transform the os env into a map. We need it for the labels later in
 	// any case.
 	osEnv, err := envLib.ParseSlice(os.Environ())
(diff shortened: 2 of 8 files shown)


giuseppe:
marked as Draft as I don't really like this solution.  Any suggestions?

rhatdan:
The way this should work, is the containers.conf entries on the client should not be sent to the server.  The server should grab the defaults itself.

So we should check if the "--env" option was used by the usr and send those across, if not then send []string.

The only reason we have them in the client at all is to attempt to tell the users the defaults.

cmd/podman/common/create.go
```diff
@@ -158,7 +158,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	)
 	createFlags.StringArrayVarP(
 		&cf.env,
-		"env", "e", containerConfig.Env(),
```
rhatdan, line 161: We want this.

cmd/podman/common/create.go
```diff
@@ -158,7 +158,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	)
 	createFlags.StringArrayVarP(
 		&cf.env,
```
rhatdan, line 160: Remove the autosetting,  And check if the user actually did --env.

cmd/podman/common/specgen.go
```diff
@@ -299,6 +299,14 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 		"container": "podman",
 	}
 
+	if defaultEnv, err := envLib.ParseSlice(containerConfig.Env()); err != nil {
```
rhatdan, line 302: server side should call containerConfig.GetDefaultEnv()

cmd/podman/common/create.go
giuseppe: not sure it works as expected.  If I specify just `--env` then it takes the next flag as the argument.

giuseppe: if we set the default here, it overrides the image's configuration

mheon: I think we need an IsSet check on the flag, to only set it in SpecGen if it was explicitly set by the user

giuseppe: > I think we need an IsSet check on the flag, to only set it in SpecGen if it was explicitly set by the user

I've not found a way to do it cleanly with string arrays/slices.  The Value is converted to a comma separated string, which doesn't work when there is a comma in the env variables.

cmd/podman/common/specgen.go
giuseppe: fixed

cmd/podman/common/create.go
mheon: I think that's just String Slice - String Array is a proper array, and allows commas

cmd/podman/common/create.go
giuseppe: fixed

cmd/podman/common/create.go
giuseppe: fixed

rhatdan:
Replacing this PR with mine.

PR #6102, opened 2020-05-06
Add command to build podman-remote-static

rhatdan:
CRC Team wants a static podman-remote that it can install on any linux
box.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,6 +187,10 @@ endif
 .PHONY: podman
 podman: bin/podman
 
+.PHONY: bin/podman-remote-static
+podman-remote-static: bin/podman-remote-static
+	CGO_ENABLED=0 $(GO) build $(BUILDFLAGS) -gcflags '$(GCFLAGS)' -asmflags '$(ASMFLAGS)' -ldflags '$(LDFLAGS_PODMAN_STATIC)' -tags "!ABISupport varlink containers_image_openpgp remoteclient" -o bin/podman-remote-static $(PROJECT)/cmd/podman
+
 .PHONY: bin/podman-remote
 bin/podman-remote: .gopathok $(SOURCES) go.mod go.sum $(PODMAN_VARLINK_DEPENDENCIES) ## Build with podman on remote environment
 	$(GO_BUILD) $(BUILDFLAGS) -gcflags '$(GCFLAGS)' -asmflags '$(ASMFLAGS)' -ldflags '$(LDFLAGS_PODMAN)' -tags "$(BUILDTAGS) remoteclient" -o $@ $(PROJECT)/cmd/podman


baude:
lgtm

rhatdan:
@mheon @cevich @baude How do I get this to build a version of podman-remote on the github.com/containers/libpod main page, that CRC team can grab?

mheon:
Automatically? Excellent question. I've been manually compiling, zipping, and generating checksums, then attaching to releases.

Also, we should make very clear that we document these are for CRC only if we're posting them here. I don't want to be fielding issues about these not working on other distros because people saw them available and assumed they were official.

rhatdan:
@mheon Sure we can write the wording however you want.

rhatdan:
Since it is only podman-remote, it should not as popular, you still need a full podman to use it.

TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
@mheon @giuseppe @baude @QiWang19 @jwhonce @vrothberg PTAL

cevich:
> Automatically? Excellent question. I've been manually compiling, zipping, and generating checksums, then attaching to releases.

The really really really important thing left unmentioned is: _We trust @mheon isn't doing evil things._   That doesn't automatically apply to automation, though there are some built-in security measures in Github Actions, there are almost none present for Cirrus-CI.

> How do I get this to build a version of podman-remote on the github.com/containers/libpod main page, that CRC team can grab?

Assuming you can tolerate the build happening through a 'docker build' command.  A compromise here is: We set up an automatic build for a new image on quay.io, but reference this file **ONLY** for specific branches (and/or tags).  Then you simply update the `README.md` file with a link to those tagged images.  They will pick up any backports automatically, but otherwise will have a static name and tag.

@rhatdan could that work for now?  What would you like the image called?

Containerfile.static
```diff
@@ -0,0 +1,12 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update
+RUN apt-get install -y \
+    make \
+    golang \
+    git-core \
+    libglib2.0-dev
+
+ENV GOPATH=/
+WORKDIR /src/github.com/containers/libpod
```
cevich, line 12: Did you intend to `ADD . /src/github.com/containers/libpod` and then `RUN podman-remote-static-local`?

Makefile
```diff
@@ -187,6 +187,16 @@ endif
 .PHONY: podman
 podman: bin/podman
 
+.PHONY: bin/podman-remote-static
+podman-remote-static: bin/podman-remote-static
+	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -f Containerfile.static -t podmanbuildimage .
+	${CONTAINER_RUNTIME} run --rm --security-opt label=disable -v $$(pwd):/src/github.com/containers/libpod \
+		podmanbuildimage make podman-remote-static-local
```
cevich, line 194: This is strange, I don't understand the purpose of this Makefile target.  It seems to build a container image, changes it's contents at runtime, then leave it on the developers workstation dangling, orphaned, and sad :cry: 

More likely: I don't understand what you're trying to accomplish here :confused:   Maybe put what I'm missing into a comment :grinning:

Makefile
```diff
@@ -187,6 +187,16 @@ endif
...
+
+.PHONY: bin/podman-remote-static-local
+podman-remote-static-local: bin/podman-remote-static
+	CGO_ENABLED=0 $(GO) build $(BUILDFLAGS) -gcflags '$(GCFLAGS)' -asmflags '$(ASMFLAGS)' -ldflags '$(LDFLAGS_PODMAN_STATIC)' -tags "!ABISupport varlink containers_image_openpgp remoteclient" -o bin/podman-remote-static $(PROJECT)/cmd/podman
+
```
cevich, line 199: Do we need a `install.podman-remote-static-local` target too?

Containerfile.static
rhatdan: No we volume mont that in and execute a podman or docker command to build the static binary.

Makefile
rhatdan: You can't do this on a Fedora/RHEL/Centos box, because glibc does not support static linking.  Because of this we need to use debian or ubuntu to compile the program staticly.

This code was mostly copied from skopeo, which already went through this.

So the code creates an image and then executes a container on it, that does the build.

If you are on Ubuntu or Debian you could just do the 
make podman-remote-static-local, and save the container part.

Makefile
rhatdan: No we do not want this to be installed, since we would prefer people to use dynamically linked podman-remote.  The only reason to build this is for people who want to use podman-remote on distros without compiling and packaging.  Mainly doing this for CRC.

rhatdan:
BTW I don't mind if we don't automate this, as long as we had could test somehow that after a release all of the packages are complete and in place.  Maybe have a test that runs on PRs to ensure the current release is complete.  The thing we want to solve is making sure that @mheon or whoever cuts a release, completes all of the tasks, if the tasks are not automated.

rhatdan:
We need to ensure that there is a MAC, Windows, Linux executable in place when podman-1.9.2 is released as an example.

cevich:
> If you are on Ubuntu or Debian you could just do the make podman-remote-static-local, and save the container part.

Ahh, okay, so in this case you may find [this Cirrus-C feature interesting/useful.](https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment).  Great for testing, not so much for publishing (safely).

I would suggest adding a comment to the `Makefile` and `Containerfile.static` explaining the container image is only needed temporarily, because the static-compile feature of glibc is required from a debian-like distro.  That way future maintainers/developers won't wonder why the makefile is building a thing to build a thing :smile: 

> The only reason to build this is for people who want to use podman-remote on distros without compiling and packaging. Mainly doing this for CRC.

So if I understand correctly, without automation @mheon would add a 'make podman-remote-static` to his release checklist, followed by uploading the compiled `/src/github.com/containers/libpod/bin/podman-remote-static` somewhere folks can get at it.  The container image is then discarded.  The libpod `README.md` should/will have a link to download this binary.  Have I got that all right?

rhatdan:
Yes.

rhatdan:
Actually checking this again, after I remove some buildtags that should not have been there, this PR works without the second container.

baude:
lgtm

PR #6107, opened 2020-05-06
Enables port test

sujil02:
Adds port subcommand for containers
Updates check for no args when all flag is set.

### DIFF
--- a/cmd/podman/parse/common.go
+++ b/cmd/podman/parse/common.go
@@ -30,13 +30,20 @@ func CheckAllLatestAndCIDFile(c *cobra.Command, args []string, ignoreArgLen bool
 		return errors.Errorf("--all and --latest cannot be used together")
 	}
 
+	if (argLen > 0) && specifiedAll {
+		return errors.Errorf("no arguments are needed with --all")
+	}
+
 	if ignoreArgLen {
 		return nil
 	}
-	if (argLen > 0) && (specifiedAll || specifiedLatest) {
-		return errors.Errorf("no arguments are needed with --all or --latest")
-	} else if cidfile && (argLen > 0) && (specifiedAll || specifiedLatest || specifiedCIDFile) {
-		return errors.Errorf("no arguments are needed with --all, --latest or --cidfile")
+
+	if argLen > 0 {
+		if specifiedLatest {
+			return errors.Errorf("no arguments are needed with --latest")
+		} else if cidfile && (specifiedLatest || specifiedCIDFile) {
+			return errors.Errorf("no arguments are needed with --latest or --cidfile")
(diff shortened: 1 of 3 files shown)


cmd/podman/parse/common.go
```diff
@@ -30,13 +30,17 @@ func CheckAllLatestAndCIDFile(c *cobra.Command, args []string, ignoreArgLen bool
 		return errors.Errorf("--all and --latest cannot be used together")
 	}
 
+	if (argLen > 0) && specifiedAll {
+		return errors.Errorf("no arguments are needed with --all")
```
rhatdan, line 34: Changes this to state
"ports are not allowed with the --all option"

sujil02: This is a common check and not a specific one.

sujil02:
@baude I have modified the check to validate no args with all flag set.
I re-ran the e2e test locally seemed fine, I could not spot a use case with args and all set together. However, if this is not the case I can revert back and add a special check-in ports

ref: https://github.com/containers/libpod/pull/6107/commits/3ada8dd3fe585c37259257bdfac94131e5a729d6#diff-c8b9edcdc2fa538036af3057366227a1R33

cmd/podman/parse/common.go
```diff
@@ -30,13 +30,17 @@ func CheckAllLatestAndCIDFile(c *cobra.Command, args []string, ignoreArgLen bool
...
+	}
+
 	if ignoreArgLen {
 		return nil
 	}
-	if (argLen > 0) && (specifiedAll || specifiedLatest) {
-		return errors.Errorf("no arguments are needed with --all or --latest")
-	} else if cidfile && (argLen > 0) && (specifiedAll || specifiedLatest || specifiedCIDFile) {
-		return errors.Errorf("no arguments are needed with --all, --latest or --cidfile")
+	if (argLen > 0) && specifiedLatest {
```
rhatdan, line 40: How about
```
             if (argLen > 0) {
                    if specifiedLatest {
		               return errors.Errorf("no arguments are needed with --latest")
	           } else  {
                      if cidfile && specifiedCIDFile {
                       		return errors.Errorf("no arguments are needed with --latest or --cidfile")
                     }
         }
```
}

sujil02: Yup. looks neater will do.
As this is a change in common error validation waiting for the test suite to run as wanted to be sure pulling the all flag validation before ignoreArgLen check looks good.

TomSweeneyRedHat:
LGTM once @rhatdan's comment is addressed.

sujil02:
@rhatdan  PTAL again

PR #6115, opened 2020-05-07
podman: split env variables in env and overrides

rhatdan:
There are three different priorities for applying env variables:

1) environment/config file environment variables
2) image's config
3) user overrides (--env)

The third kind are known to the client, while the default config and image's
config is handled by the backend.

### DIFF
--- a/pkg/specgen/generate/container.go
+++ b/pkg/specgen/generate/container.go
@@ -9,6 +9,7 @@ import (
 	envLib "github.com/containers/libpod/pkg/env"
 	"github.com/containers/libpod/pkg/signal"
 	"github.com/containers/libpod/pkg/specgen"
+	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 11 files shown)


rhatdan:
I believe this should replace https://github.com/containers/libpod/pull/6100

But we still have issues with --http-proxy and --env-host.

Where should these defaults come from.  In a different conversation, it was suggested that --http-proxy should come from the server, currently it is coming from the client.

I think if we want this to be used on the server, we would also need to handle --env-host on the server side as well, or we would need to send over a different set of enviornment variables from client to server.

The question is whether or not sending the environment from a MAC, Windows, Linux box over the network to a server makes sense, or would the user expect the environment from the server to be handled?

pkg/specgen/specgen.go
```diff
@@ -44,7 +44,6 @@ type ContainerBasicConfig struct {
 	Command []string `json:"command,omitempty"`
 	// Env is a set of environment variables that will be set in the
 	// container.
-	// Optional.
```
mheon, line 47: I'm pretty sure this is still Optional - the user doesn't need to specify any environment variables to successfully make a container.

mheon:
I still see potential issues with `--env-file` here - it will override image environment like this, which I think is incorrect. Though maybe we just provide paths to the files in the SpecGen? That may well be what Docker does (I don't think they parse it in the frontend)

pkg/specgen/generate/container.go
```diff
@@ -48,24 +49,28 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
...
-		for k, v := range envs {
-			if _, exists := s.Env[k]; !exists {
-				s.Env[v] = k
-			}
-		}
+	envs, err := envLib.ParseSlice(imageEnvs)
+	if err != nil {
+		return errors.Wrap(err, "Env fields from image failed to parse")
 	}
+	s.Env = envLib.Join(envLib.Join(defaultEnvs, envs), s.Env)
```
giuseppe, line 73: this won't respect the priorities, we need to ensure `--env` variables always have a higher priority than what is specified in the image.
My patch could be simplified to not propagate two different Env blocks if the env files coming from `--env-host` has a higher priority than the image's configuration, but I think as an user I'd not expect that.

rhatdan: No this does respoend it.  S.Env takes priority, which is input from the users.

defaultEnvs -- containers.conf,
envs == image
s.Env == (--env-host, env-file, --env)

envlib.Join works like append in the second parameter will replace first, if they conflict.

rhatdan:
I updated this PR to add another patch to move http-proxy to server side.

pkg/specgen/generate/container.go
rhatdan: BTW if this was not done correctly, I would expect the tests to pass.

mheon:
I doubt we actually have tests for proper parsing order in environment variables. All the `podman run` tests passed without any of this being added.

PR #6122, opened 2020-05-07
Set pids-limit to 0 on non cgroupv2 rootless systems

rhatdan:
Rootless cgroups V1 does not support setting a pids limit.

### DIFF
--- a/cmd/podman/commands.go
+++ b/cmd/podman/commands.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	"github.com/containers/buildah/pkg/parse"
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/libpod/pkg/apparmor"
 	"github.com/containers/libpod/pkg/cgroups"
 	"github.com/containers/libpod/pkg/rootless"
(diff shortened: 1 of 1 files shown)


rhatdan:
@mheon @vrothberg @giuseppe @baude @TomSweeneyRedHat PTAL

cmd/podman/commands.go
```diff
@@ -182,6 +182,8 @@ func getDefaultPidsLimit() int64 {
 		cgroup2, _ := cgroups.IsCgroup2UnifiedMode()
```
mheon, line 182: I think we also have to verify that cgroup manager is set to systemd - @giuseppe can you confirm?

TomSweeneyRedHat:
LGTM other than @mheon's question

cmd/podman/commands.go
giuseppe: yes, with cgroupfs rootless is still not able to use cgroups

rhatdan: Fixed

rhatdan:
@mheon @giuseppe PTAL

PR #6127, opened 2020-05-07
[WIP]fix pod stats flake

baude:
it appears that the pod stats flake can be attributed to the fact that the container being run is not fully running when the stats call is made.  because the stats call is in format of json, it fails when nil

### DIFF
--- a/test/e2e/stats_test.go
+++ b/test/e2e/stats_test.go
@@ -5,6 +5,7 @@ package integration
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/containers/libpod/pkg/cgroups"
 	. "github.com/containers/libpod/test/utils"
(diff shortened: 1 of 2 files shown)


mheon:
Shouldn't we be getting a better error out of Stats when the container isn't running?

test/e2e/stats_test.go
```diff
@@ -87,13 +88,27 @@ var _ = Describe("Podman stats", func() {
...
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		session = podmanTest.Podman([]string{"stats", "--all", "--no-stream", "--format", "json"})
-		session.WaitWithDefaultTimeout()
-		Expect(session.ExitCode()).To(Equal(0))
-		Expect(session.IsJSONOutputValid()).To(BeTrue())
+		for i := 0; i < 5; i++ {
+			ps := podmanTest.Podman([]string{"ps", "-q"})
+			ps.WaitWithDefaultTimeout()
+			if len(ps.OutputToStringArray()) == 1 {
```
rhatdan, line 98: Probably safer to do `if len(ps.OutputToStringArray())  > 0`

vrothberg:
I am going to close this as the stats flakes were fixed by commit 3cef598fc2b2.

PR #6130, opened 2020-05-08
set --conmon-pidfile

vrothberg:
The --conmon-pidfile was not set in the spec leading to failing systemd
units.  Also add a system test to prevent future regressions.


vrothberg:
@rhatdan PTAL

vrothberg:
Merging manually as the bot doesn't react

PR #6131, opened 2020-05-08
[1.9] add --conmon-pidfile/--cidfile system test

vrothberg:
Add a system test to make sure that pidfile and cidfile are being
generated.  Podman regressed in master and given 1.9 will be supported
for a while, let's make extra sure we're catching potential on these
flags early.


vrothberg:
@giuseppe @mheon PTAL

PR #6134, opened 2020-05-08
fix and enable systemd system tests

vrothberg:
The systemd unit test never ran in CI and was broken for various
reasons.  Fix the test to execute Podman in systemd units and to also
run generated units files.

Note: more tests will be added in the future.  The simple check for now
will prevent regressions.


rhatdan:
LGTM

rhatdan:
@baude @giuseppe @mheon @TomSweeneyRedHat PTAL and Merge

mheon:
Would really like a signoff from @edsantiago before merge

edsantiago:
Already looked this morning, but it's Friday so I didn't want to admit to working. LGTM and a welcome addition.

PR #6137, opened 2020-05-08
Fix handling of overridden paths from database

rhatdan:
If the first time you run podman in a user account you do a
su - USER, and the second time, you run as the logged in USER
podman fails, because it is not handling the tmpdir definition
in the database. This PR fixes this problem.

vendor containers/common v0.11.1

This should fix a couple of issues we have seen in podman 1.9.1
with handling of libpod.conf.

### DIFF
--- a/cmd/podman/login.go
+++ b/cmd/podman/login.go
@@ -46,7 +46,6 @@ func init() {
 
 	// Podman flags.
 	flags.BoolVarP(&loginOptions.tlsVerify, "tls-verify", "", false, "Require HTTPS and verify certificates when contacting registries")
-	flags.BoolVarP(&loginOptions.GetLoginSet, "get-login", "", false, "Return the current login user for the registry")
 	loginOptions.Stdin = os.Stdin
 	loginOptions.Stdout = os.Stdout
 	loginOptions.AcceptUnspecifiedRegistry = true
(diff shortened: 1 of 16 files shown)


rhatdan:
@mheon @giuseppe @vrothberg @baude @QiWang19 @jwhonce @TomSweeneyRedHat PTAL

cmd/podman/login.go
```diff
@@ -46,7 +46,6 @@ func init() {
 
 	// Podman flags.
 	flags.BoolVarP(&loginOptions.tlsVerify, "tls-verify", "", false, "Require HTTPS and verify certificates when contacting registries")
-	flags.BoolVarP(&loginOptions.GetLoginSet, "get-login", "", false, "Return the current login user for the registry")
```
mheon, line 49: Why are we dropping this? Seems unrelated to the rest of the PR

mheon:
One question otherwise LGTM

cmd/podman/login.go
TomSweeneyRedHat: Ditto @mheon's question.  If we are dropping it, we need a man page change.

rhatdan: @mtrmac Added this option to github/container/common/pkg/auth, so it is conflicting in podman code and no longer needed.

PR #6142, opened 2020-05-08
default to tunnel without ABISupport tag

vrothberg:
When compiling a Linux binary without ABISupport, default to use the
tunnel.  The behaviour is expected in `podman-remote`.

Also set a default for the remote flag so `podman-remote` works OOB.

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -208,7 +208,7 @@ func syslogHook() {
 
 func rootFlags(opts *entities.PodmanConfig, flags *pflag.FlagSet) {
 	// V2 flags
-	flags.StringVarP(&opts.Uri, "remote", "r", "", "URL to access Podman service")
+	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress, "URL to access Podman service")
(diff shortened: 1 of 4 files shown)


cmd/podman/root.go
```diff
@@ -208,7 +208,7 @@ func syslogHook() {
 
 func rootFlags(opts *entities.PodmanConfig, flags *pflag.FlagSet) {
 	// V2 flags
-	flags.StringVarP(&opts.Uri, "remote", "r", "", "URL to access Podman service")
+	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress, "URL to access Podman service")
```
rhatdan, line 211: Does this work on rootless and rootfull?  Or do we need a function call?

mheon:
LGTM once the question from @rhatdan is addressed

cmd/podman/root.go
vrothberg: Good point! Currently, `podman system service` does not distinguish between root and rootless. However, it seems like a follow up PR until rootless works.

TomSweeneyRedHat:
LGTM
I'm fine with merging this now as long as there's an issue/jira card spun to answer @rhatdan's question.

PR #6144, opened 2020-05-08
Fix `podman pod create --infra=false`

mheon:
We were accidentally setting incorrect defaults for the network namespace for rootless `pod create` when infra containers were not being created. This should resolve that issue.


baude:
LGTM thanks @mheon

Issue #5205, opened 2020-02-14
network inspect does not have format


medyagh:
we added podman as a driver to minikube ! and merged but I wanna add minikube's mount feature. https://github.com/kubernetes/minikube/issues/6628

I wanna do something simmilar to this in docker:
```
docker inspect --format '{{(index .IPAM.Config 0).Gateway}}' $(docker network ls --filter name=bridge --format {{.ID}})
```

but podman doesnt have format for network inspect

```
 podman network inspect --help
network inspect
Description:
  Inspect network
Usage:
  podman network inspect NETWORK [NETWORK...] [flags] 
Examples:
  podman network inspect podman
```

vrothberg:
Thanks for opening the issue, @medyagh! Are you interested in contributing this feature?

baude:
@medyagh the network stack is a major difference between podman and docker.  for example, in inspect there really is very little chance the output would look the same.  given that, would format still be helpful?

medyagh:
@baude all i need is a way to get the routable IP of the host from inside of the container.

so if I have a webserver running on the host machine, 
from inside the container, I wanna be able to ping that IP . 

in docker I get that IP using this command:
```
docker inspect --format '{{(index .IPAM.Config 0).Gateway}}' $(docker network ls --filter name=bridge --format {{.ID}})
```
i want same command for podman (so I can implement minikube mount command)

rhatdan:
This is definitely something we want.  Just need someone to take it up and run with it.
@kunalkushwaha @sujil02 Anyone interested?

kunalkushwaha:
I can pick this.

/assign

rhatdan:
@kunalkushwaha What is going on with this issue?

kunalkushwaha:
Sorry, somehow it skipped from task list. Its half done.. shall make PR soon.

medyagh:
@kunalkushwaha are there anything new about this ?

kunalkushwaha:
@medyagh Sorry for the delay in this. Due to recent client code (v2 changes), my fix is not mergeable. 
I am trying to fix this. Shall be able to do it soon.

afbjorklund:
I think it would be better to look at the container's gateway ? Or am I missing something here.

QiWang19:
This issue #5205 was fixed by #6161

PR #6161, opened 2020-05-11
`--format` and `--filter` options for `network ls` and `network inspect` command

kunalkushwaha:
This PR is for fixing bug #5205 
This adds following options to network commands 
  - `podman network ls --filter` option 
  -  `podman network inspect --format` option

NOTE: _Since the CNI conf file does not have defined structure and is read as read & stored as `map of interfaces`, its difficult to get similar output as of `docker` as specified in https://github.com/containers/libpod/issues/5205#issue-565036875. until we create an internal structure. Which may not be good idea, as it may require to sync up with libcni code. Ideally, this structure should be defined in libcni code._

PTAL: @medyagh @afbjorklund, If this helps your use case.

### DIFF
--- a/cmd/podman/networks/inspect.go
+++ b/cmd/podman/networks/inspect.go
@@ -3,6 +3,10 @@ package network
 import (
 	"encoding/json"
 	"fmt"
+	"html/template"
+	"io"
+	"os"
+	"strings"
 
 	"github.com/containers/libpod/cmd/podman/registry"
 	"github.com/containers/libpod/pkg/domain/entities"
@@ -24,12 +28,18 @@ var (
 	}
 )
 
+var (
+	networkInspectOptions entities.NetworkInspectOptions
+)
+
 func init() {
 	registry.Commands = append(registry.Commands, registry.CliCommand{
 		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
 		Command: networkinspectCommand,
 		Parent:  networkCmd,
 	})
+	flags := networkinspectCommand.Flags()
+	flags.StringVarP(&networkInspectOptions.Format, "format", "f", "", "Pretty-print network to JSON or using a Go template")
 }
 
 func networkInspect(cmd *cobra.Command, args []string) error {
@@ -41,6 +51,22 @@ func networkInspect(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	fmt.Println(string(b))
+	if strings.ToLower(networkInspectOptions.Format) == "json" || networkInspectOptions.Format == "" {
+		fmt.Println(string(b))
+	} else {
+		var w io.Writer = os.Stdout
+		//There can be more than 1 in the inspect output.
+		format := "{{range . }}" + networkInspectOptions.Format + "{{end}}"
+		tmpl, err := template.New("inspectNetworks").Parse(format)
+		if err != nil {
+			return err
+		}
+		if err := tmpl.Execute(w, responses); err != nil {
+			return err
+		}
+		if flusher, ok := w.(interface{ Flush() error }); ok {
+			return flusher.Flush()
+		}
+	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


sujil02:
It might be a good idea to add some test cases here https://github.com/containers/libpod/blob/master/test/e2e/network_test.go#L95 corresponding to new flags.

TomSweeneyRedHat:
Quick peruse LGTM.  Need tests and also updates to `./completions/bash/podman`

docs/source/markdown/podman-network-ls.1.md
```diff
@@ -14,6 +14,14 @@ Displays a list of existing podman networks. This command is not available for r
 
 The `quiet` option will restrict the output to only the network names
 
+**--format**, **-f**
+
+Pretty-print networks to JSON or using a Go template
+
+**--fiter**
```
rhatdan, line 21: Spelling.

docs/source/markdown/podman-network-inspect.1.md
```diff
@@ -9,6 +9,15 @@ podman\-network\-inspect - Displays the raw CNI network configuration for one or
...
 Display the raw (JSON format) network configuration. This command is not available for rootless users.
 
+## OPTIONS
+**--quiet**, **-q**
+
+The `quiet` option will restrict the output to only the network names
+
+**--format**, **-f**
+
+Pretty-print networks to JSON or using a Go template
```
TomSweeneyRedHat, line 19: ending period

docs/source/markdown/podman-network-inspect.1.md
```diff
@@ -9,6 +9,15 @@ podman\-network\-inspect - Displays the raw CNI network configuration for one or
 ## DESCRIPTION
 Display the raw (JSON format) network configuration. This command is not available for rootless users.
 
+## OPTIONS
+**--quiet**, **-q**
+
+The `quiet` option will restrict the output to only the network names
```
TomSweeneyRedHat, line 15: ending period

docs/source/markdown/podman-network-ls.1.md
```diff
@@ -14,6 +14,14 @@ Displays a list of existing podman networks. This command is not available for r
 
 The `quiet` option will restrict the output to only the network names
 
+**--format**, **-f**
+
+Pretty-print networks to JSON or using a Go template
```
TomSweeneyRedHat, line 19: ending period

docs/source/markdown/podman-network-ls.1.md
```diff
@@ -14,6 +14,14 @@ Displays a list of existing podman networks. This command is not available for r
...
+
+**--filter**
+
+Provide filter values (e.g. 'name=podman')
```
TomSweeneyRedHat, line 23: ending period

cmd/podman/networks/inspect.go
```diff
@@ -41,6 +51,23 @@ func networkInspect(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	fmt.Println(string(b))
+	if strings.ToLower(networkInspectOptions.Format) == "json" || networkInspectOptions.Format == "" {
+		fmt.Println(string(b))
+	} else {
+
```
QiWang19, line 57: empty line not needed

sujil02:
@rhatdan: Need further opinion. I might be wrong.

Generally from what I have seen filters are usually applied lower on the stack bypassing the filter as params down. From how its done here filters for the network cannot be used by remote clients resulting in ambiguity in the responses. This can be considered a workaround in my opinion.

rhatdan:
@sujil02 Not sure what you are saying, But I believe the filtering should be taking place on the server side not in the client.

sujil02:
What I meant was the api for network list will not be able to implement filters as it is not bonded to the `url.values()` and passed down.

QiWang19:
yes, I think this filter should be passing down. With the tunnel mode NetworkList hasn't been implemented yet, maybe for this PR is should pass to pkg/domain/infra/abi/network.go?

kunalkushwaha:
Got it. I will move the filter code from cli to `pkg`

QiWang19:
LGTM

PR #6167, opened 2020-05-11
spec: fix order for setting rlimits

giuseppe:
also make sure that the limits we set for rootless are not higher than
what we'd set for root containers.


giuseppe:
@rhatdan we need to backport to 1.9

rhatdan:
@giuseppe Please open a PR to backport.

PR #6174, opened 2020-05-11
rootless: do not set pids limits with cgroupfs

giuseppe:
and enable events tests.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -8,12 +8,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/libpod/cmd/podman/parse"
 	"github.com/containers/libpod/libpod/define"
 	ann "github.com/containers/libpod/pkg/annotations"
 	envLib "github.com/containers/libpod/pkg/env"
 	ns "github.com/containers/libpod/pkg/namespaces"
+	"github.com/containers/libpod/pkg/rootless"
 	"github.com/containers/libpod/pkg/specgen"
 	systemdGen "github.com/containers/libpod/pkg/systemd/generate"
 	"github.com/containers/libpod/pkg/util"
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

mheon:
LGTM

mheon:
Haven't seen this test failure before, might be legitimate

cmd/podman/common/specgen.go
```diff
@@ -128,18 +130,21 @@ func getIOLimits(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string) (
...
+		pids.Limit = containerConfig.PidsLimit()
+		return pids, nil
+	}
 	if c.PIDsLimit > 0 {
 		pids.Limit = c.PIDsLimit
-		hasLimits = true
-	}
-	if !hasLimits {
-		return nil, nil
+		return pids, nil
```
TomSweeneyRedHat, line 145: if `c.PidsLimit == 0` is falling through and returning `nil, nil` the right thing?

giuseppe: yes, `PidsLimit==0` means no limits

cmd/podman/common/specgen.go
```diff
@@ -128,18 +130,21 @@ func getIOLimits(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string) (
 
 func getPidsLimits(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string) (*specs.LinuxPids, error) {
```
rhatdan, line 131: It does not look like this function returns an error, should we drop the error?

giuseppe: sure, pushed a new version.

giuseppe:
tests pass now

test/e2e/run_networking_test.go
```diff
@@ -202,7 +203,7 @@ var _ = Describe("Podman run networking", func() {
 		Expect(inspect.ExitCode()).To(Equal(0))
 
 		containerConfig := inspect.InspectContainerToJSON()
-		Expect(containerConfig[0].NetworkSettings.Ports[0].HostPort).ToNot(Equal("80"))
+		Expect(containerConfig[0].NetworkSettings.Ports[0].HostPort).ToNot(Equal(80))
```
jwhonce, line 206: Is this an expected change?  Seems like it could break clients.

PR #6176, opened 2020-05-11
Some BATS cleanup: run and systemd tests

edsantiago:
run test: run positive test before negative; and actually
implement real negative tests. Note that '--conmon-pidfile'
will happily clobber an existing file; I don't know if that
is desired behavior. Also, add confirmation tests for
cidfile/pidfile, not just 'exit status is good'.

systemd test: enable rootless, and again add actual content
testing.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -161,17 +161,31 @@ echo $rand        |   0 | $rand
 # 'run --conmon-pidfile --cid-file' makes sure we don't regress on these flags.
 # Both are critical for systemd units.
 @test "podman run --conmon-pidfile --cidfile" {
-    pid=$(mktemp)
-    cid=$(mktemp)
+    pidfile=${PODMAN_TMPDIR}/pidfile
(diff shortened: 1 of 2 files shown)


edsantiago:
@vrothberg ptal

mheon:
LGTM but want a signoff from @vrothberg before merge

edsantiago:
I'd rather not merge with that giant FIXME. @vrothberg can you comment on whether or not --conmon-pidfile should clobber existing file?

vrothberg:
> I'd rather not merge with that giant FIXME. @vrothberg can you comment on whether or not --conmon-pidfile should clobber existing file?

cidfile and conmon-pidfile behave differently. AFAICS, it's expected that cidfile fails (Docker compat) while pidfile overwrites (to make conmon simpler).

edsantiago:
Thank you. I've removed the FIXME and the commented-out code. Tests are green.

PR #6181, opened 2020-05-11
add port to podman remote command

baude:

### DIFF
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -381,7 +381,29 @@ func (ic *ContainerEngine) Config(_ context.Context) (*config.Config, error) {
 }
 
 func (ic *ContainerEngine) ContainerPort(ctx context.Context, nameOrId string, options entities.ContainerPortOptions) ([]*entities.ContainerPortReport, error) {
-	return nil, errors.New("not implemented")
+	var (
+		reports    []*entities.ContainerPortReport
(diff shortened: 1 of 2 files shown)


pkg/domain/infra/tunnel/containers.go
```diff
@@ -381,7 +381,29 @@ func (ic *ContainerEngine) Config(_ context.Context) (*config.Config, error) {
...
+		namesOrIds = append(namesOrIds, nameOrId)
+	}
+	ctrs, err := getContainersByContext(ic.ClientCxt, options.All, namesOrIds)
+	if err != nil {
+		return nil, err
+	}
+	for _, con := range ctrs {
+		if con.State != define.ContainerStateRunning.String() {
+			continue
+		}
```
TomSweeneyRedHat, line 398: Not sure, but I don't think we should add them to paused containers either.

rhatdan: This code is ignoreing all non running containers including paused containers, I believe?

rhatdan:
LGTM

pkg/domain/infra/tunnel/containers.go
baude: should be

TomSweeneyRedHat: DOH, don't know what I was reading earlier, ignore that comment.

PR #6193, opened 2020-05-12
Cirrus: Include packages for containers/conmon CI

cevich:
Depends on: #6201 

This allows the containers/conmon repository to share the same VM
images produced by containers/libpod.  Included are several packages
which are downloaded only since they might otherwise interfere with
testing for some repos.  This allows stable versions to be at the ready
at testing runtime, avoiding any version updates surprising developers.

### DIFF
--- a/contrib/cirrus/packer/fedora_packaging.sh
+++ b/contrib/cirrus/packer/fedora_packaging.sh
@@ -11,6 +11,8 @@ echo "Updating/Installing repos and packages for $OS_REL_VER"
 
 source $GOSRC/$SCRIPT_BASE/lib.sh
 
+req_env_var GOSRC SCRIPT_BASE BIGTO INSTALL_AUTOMATION_VERSION FEDORA_BASE_IMAGE PRIOR_FEDORA_BASE_IMAGE
+
 # Pre-req. to install automation tooing
 $LILTO $SUDO dnf install -y git
 
@@ -35,7 +37,7 @@ fi
 
 $BIGTO ooe.sh $SUDO dnf update -y
 
-REMOVE_PACKAGES=()
+REMOVE_PACKAGES=(runc)
 INSTALL_PACKAGES=(\
     autoconf
     automake
@@ -50,8 +52,11 @@ INSTALL_PACKAGES=(\
     containernetworking-plugins
     containers-common
     criu
+    crun
+    curl
     device-mapper-devel
     dnsmasq
+    e2fsprogs-devel
     emacs-nox
     file
     findutils
@@ -60,16 +65,26 @@ INSTALL_PACKAGES=(\
     gcc
     git
     glib2-devel
+    glibc-devel
     glibc-static
     gnupg
     go-md2man
     golang
+    gpgme
     gpgme-devel
+    grubby
+    hostname
     iproute
     iptables
     jq
+    krb5-workstation
+    libassuan
     libassuan-devel
+    libblkid-devel
     libcap-devel
+    libffi-devel
+    libgpg-error-devel
+    libguestfs-tools
     libmsi1
     libnet
     libnet-devel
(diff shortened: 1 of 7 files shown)


contrib/cirrus/packer/fedora_packaging.sh
```diff
@@ -60,82 +63,107 @@ INSTALL_PACKAGES=(\
...
     iptables
     jq
+    krb5-workstation
+    libassuan
     libassuan-devel
+    libblkid-devel
     libcap-devel
+    libffi-devel
+    libgpg-error-devel
+    libguestfs-tools
```
haircommander, line 85: can we add parallel here? it's used for running bats jobs in parallel in newer versions of cri-o

contrib/cirrus/packer/ubuntu_packaging.sh
```diff
@@ -122,6 +122,10 @@ INSTALL_PACKAGES=(\
     zip
     zlib1g-dev
 )
+DOWNLOAD_PACKAGES=(\
+    cri-o-1.17_1.17.4
```
haircommander, line 126: is there the corresponding package bump in fedora?

contrib/cirrus/packer/fedora_packaging.sh
cevich: We can, but I really recommend against running tests in parallel.  Sometimes it makes sense and is safe, but often it makes debugging incredibly difficult with very little time-savings.  Lastly, our Fedora and Ubuntu setups need to be maintained fairly close, package-wise.

For example, I had a lot of problems running bats in parallel on Ubuntu for Buildah due to unsolvable bugs.  It wasted a lot of Dan and my time debugging, where ultimately I just had to add a condition to skip parallel testing on Ubuntu.  Complicating the setup, and arguably making it less reliable and harder to maintain.

Is it REALLY necessary for conmon, with substantial time-savings (or some other huge benefit)?

contrib/cirrus/packer/ubuntu_packaging.sh
cevich: Sadly, this is a really F'd up aspect of Ubuntu-land and our cri-o packages.  The version number is part of the package name (they do the same thing with the damn kernel).  I would MUCH prefer to just specify `crio`, where (as in Fedora-land) it simply picks up the latest available distro version and dependencies.

That said, since we're just downloading the package, and will install it at runtime, I can be flexible here if there is another/different solution needed.  Also remember, while conmon will share images produced from this project, it does _NOT_ have to always be using the exact same images as libpod.  In other words, each project can move to new (or old) images as needed by that project.

contrib/cirrus/packer/fedora_packaging.sh
haircommander: well, cri-o currently runs the tests in parallel all the time. It is in the cri-o code, not conmon. We'd have to carry a patch to stop cri-o from doing it. The tests are well designed to run in parallel and have been for a long time. If there is an issue there, it would have to be dealt with in cri-o as well (probably by me anyway), so I don't see the risk in adding it for conmon

cevich: Oh it's in cri-o okay, and I bet the time-savings is quite significant.  I'm okay with adding it since you'll be the one debugging :grin:   If it occurs to you during your cri-o travels, I would suggest adding a parallelism mechanism based on an env. var.  Those are easy to override in most CI systems, in case it's needed for debugging.

Here's how it's done[ in containers/storage ](https://github.com/containers/storage/blob/master/tests/test_runner.bash#L21-L24).  Having the ability to set `JOBS=1` saved me wasting lots of debugging time on Ubuntu (as essplained in the comment).  Otherwise, there's a reasonable default set based on the number of CPUs in the VM.

cevich: ...okay I added 'parallel' to both Ubuntu and Fedora VM images.

cevich:
@haircommander Note while I'm out: For this PR or anything like it (using the `CI:IMG` magic) doesn't actually need to be merged in order to utilize the output VM Images in the conmon repo.  Though eventually having them merged here (and using the new VM images) is desirable from a maintenance perspective (as the instructions for building the images is also changing).  This makes it easy to test out changes to images used elsewhere in another PR (like https://github.com/containers/conmon/pull/152), nothing is permanent until merged, and the `meta` task will do it's job of keeping the active images from being pruned.

The important thing in this PR is: Download fixed-versions of cri-o and kube along with all dependencies, but do not install them (they will break libpod testing).  Then in conmon CI, we rpm install the downloaded files.  This is to prevent package updates/changes happening at test-runtime, which would open a window of uncertainty and/or introduce unpredictable behavior.  That said, you _can_ do runtime package updates/installs, it's not strictly forbidden. just something I like so unrelated development isn't impacted.

Hopefully this all makes sense.

cevich:
Removing ***WIP*** and marking this as optional.  It has changes and new VM images needed by conmon, but nothing immediately required in libpod.  It's fine if it sits here until I get back from PTO.

cevich:
Rebased + Rebuilt all VM images.

PR #6195, opened 2020-05-12
troubleshooting: describe systemd in centos:7

ktdreyer:
Hosts with cgroups v2 cannot run systemd in `centos:7`. We described this for RHEL 7, but not CentOS 7. Clarify the implications for RHEL, CentOS, and Fedora.

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -424,9 +424,9 @@ Choose one of the following:
     * Install the fuse-overlayfs package for your Linux Distribution.
     * Add `mount_program = "/usr/bin/fuse-overlayfs"` under `[storage.options]` in your `~/.config/containers/storage.conf` file.
 
-### 17) rhel7-init based images don't work with cgroups v2
+### 17) RHEL 7-based images don't work with cgroups v2
 
-The systemd version shipped in rhel7-init doesn't have support for cgroups v2.  You'll need at least systemd 230.
+The systemd version shipped in RHEL 7 and CentOS 7 doesn't have support for cgroups v2.  You'll need at least systemd 230 inside the RHEL 7-based container image.
 
 #### Symptom
 ```console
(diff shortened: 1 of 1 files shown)


troubleshooting.md
```diff
@@ -426,7 +426,7 @@ Choose one of the following:
 
 ### 17) rhel7-init based images don't work with cgroups v2
 
-The systemd version shipped in rhel7-init doesn't have support for cgroups v2.  You'll need at least systemd 230.
+The systemd version shipped in rhel7-init and centos:7 doesn't have support for cgroups v2.  You'll need at least systemd 230.
```
mheon, line 429: Maybe we should just say "RHEL 7 and CentOS 7 images"

troubleshooting.md
```diff
@@ -437,6 +437,8 @@ Failed to get D-Bus connection: Operation not permitted
 Error: non zero exit code: 1: OCI runtime error
 ```
 
+This means that you cannot run systemd inside a CentOS 7 or RHEL 7 container if your host uses cgroups v2. RHEL 8 and Fedora use cgroups v2.
```
mheon, line 440: Fedora 31 and later

mheon:
Can you add a signoff with `git commit -s --amend --no-edit` and re-push to fix CI?

/approve
/ok-to-test

TomSweeneyRedHat:
LGTM once @mheon's points are addressed and gating passes.  Thanks @ktdreyer !

troubleshooting.md
rhatdan: RHEL 8 does not default to cgroups V2.  It defaults to cgroups V1.

ktdreyer:
I was too slow here. Thanks Dan for continuing on with https://github.com/containers/libpod/pull/6199

PR #6196, opened 2020-05-12
Remove libpod.conf from repo

mheon:
Now that we're shipping containers.conf, we don't want to provide a libpod.conf anymore. This removes libpod.conf from the repo and as many direct uses as I can find.

There are a few more mentions in the documentation, but someone more familiar with containers.conf should make those edits.


mheon:
They may cause CI to explode. Will deal with it if it happens.

baude:
yeah LGTM

TomSweeneyRedHat:
@lsm5 PTAL

mheon:
I think this will go on our master CI without much of an issue. I think it will actively explode on the 1.9 branch, so I don't think this can be backported there.

libpod.conf
```diff
@@ -1,181 +0,0 @@
-# libpod.conf is the default configuration file for all tools using libpod to
```
TomSweeneyRedHat, line 1: Should we keep the file, but just have it contain a small comment/note that its been removed as of version 1.9 and replaced by containers.conf?

lsm5:
LGTM, I might need to adjust packaging but no biggie..

rhatdan:
/lgtm
/hold
I don't think we should leave the file around.

PR #6198, opened 2020-05-12
[1.9] cgroup: skip unified if we are using v1

haircommander:
There exists a chance for a node to be booted in v1, but have reminents of v2. An example is this CRI-O ci run: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3565/test_pull_request_crio_critest_fedora/11243/

We fail because we are incorrectly writing to the unified path, because we are v1. We should not write to the unified path if we are v1

fixes test failures found in https://github.com/cri-o/cri-o/pull/3565

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -517,6 +517,10 @@ func (c *CgroupControl) AddPid(pid int) error {
 	}
 
 	for _, n := range names {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe PTAL

PR #6203, opened 2020-05-12
V2 attach bindings and test

jwhonce:
* Add ErrLostSync to report lost of sync when de-mux'ing stream
* Add logus.SetLevel(logrus.DebugLevel) when `go test -v` given
* Add context to debugging messages
* Add context to signal error returns

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -1704,6 +1704,8 @@ func httpAttachTerminalCopy(container *net.UnixConn, http *bufio.ReadWriter, cid
 	buf := make([]byte, bufferSize)
 	for {
 		numR, err := container.Read(buf)
+		logrus.Debugf("Read fd(%d) %d/%d bytes for container %s", int(buf[0]), numR, len(buf), cid)
+
 		if numR > 0 {
 			switch buf[0] {
 			case AttachPipeStdout:
--- a/pkg/bindings/containers/containers.go
+++ b/pkg/bindings/containers/containers.go
@@ -2,6 +2,8 @@ package containers
 
 import (
 	"context"
+	"encoding/binary"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -13,6 +15,10 @@ import (
 	"github.com/containers/libpod/pkg/bindings"
 	"github.com/containers/libpod/pkg/domain/entities"
 	"github.com/pkg/errors"
+)
+
+var (
+	ErrLostSync = errors.New("lost synchronization with attach multiplexed result")
 )
 
 // List obtains a list of containers in local storage.  All parameters to this method are optional.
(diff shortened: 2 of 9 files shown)


baude:
lgtm, we need this merged

TomSweeneyRedHat:
LGTM

libpod/oci_conmon_linux.go
```diff
@@ -1704,6 +1704,8 @@ func httpAttachTerminalCopy(container *net.UnixConn, http *bufio.ReadWriter, cid
 	buf := make([]byte, bufferSize)
 	for {
 		numR, err := container.Read(buf)
+		logrus.Debugf("Read fd(%d) %d/%d bytes for container %s", int(buf[0]), numR, len(buf), cid)
```
mheon, line 1707: We should drop the debugfs here, I think - this is a pretty hot loop and could be expensive even if it compiles down to a single conditional branch

pkg/bindings/containers/containers.go
```diff
@@ -333,3 +339,125 @@ func ContainerInit(ctx context.Context, nameOrID string) error {
 	}
 	return response.Process(nil)
 }
+
+// Attach attaches to a running container
+func Attach(ctx context.Context, nameOrId string, detachKeys *string, logs, stream *bool, stdin, stdout, stderr chan string) error {
```
mheon, line 344: Why are we `chan string` for the standard streams? I would think that io.File so we can directly pass in the stdin/out/err FDs would be a lot better

mheon: (More accurately, `io.Reader` for stdin, `io.Writer` or `WriteCloser` for stdout/err

pkg/bindings/containers/containers.go
```diff
@@ -333,3 +339,125 @@ func ContainerInit(ctx context.Context, nameOrID string) error {
...
+				break
+			}
+			stdout <- string(buffer)
+		}
+	}
+	return err
+}
+
+// DemuxHeader reads header for stream from server multiplexed stdin/stdout/stderr/2nd error channel
+func DemuxHeader(r io.Reader, buffer []byte) (fd, sz int, err error) {
```
mheon, line 428: Are we sure this works with terminal attach? There'll be no headers there

pkg/bindings/containers/containers.go
```diff
@@ -333,3 +339,125 @@ func ContainerInit(ctx context.Context, nameOrID string) error {
...
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	ctype := response.Header.Get("Content-Type")
+	upgrade := response.Header.Get("Connection")
+
+	buffer := make([]byte, 1024)
+	if ctype == "application/vnd.docker.raw-stream" && upgrade == "Upgrade" {
```
mheon, line 380: I think that the correct course of action here is likely to inspect the container and determine if Terminal is set, and then proceed based on that

pkg/bindings/containers/containers.go
jwhonce: I went with io.Writer.

pkg/bindings/containers/containers.go
jwhonce: Do you still send the application/vnd.docker.raw-stream header?

mheon: Yes - that is always sent, for both types of attach.

PR #6207, opened 2020-05-13
WIP - compat handlers: add X-Registry-Auth header support

vrothberg:


vrothberg:
Manual tests were successful. I am now working on testing that in the apiv2 tests.

vrothberg:
> Manual tests were successful. I am now working on testing that in the apiv2 tests.

I looked into running a local registry and found a blocker, unfortunately pretty late in the process. The Docker compat endpoints don't expose a `tlsVerify` parameter that we could set to false. Hence, the server fails pulling because we're creating our own certificate. One workaround would be to use a custom registries.conf where we're setting the local registry as insecure and that's where I am stuck.

There is no way to point Podman to a custom registries.conf. The e2e tests are using a REGISTRIES_CONF_PATH env variable but that's a NOP as there's no code using that  at all.

For now, I suggest to merge the PR as is and to trust that my local smoke tests work.

However, I believe we need a way to point Podman to a custom registries.conf. Buildah exposes a `--registries-conf` which I suggest to add to Podman as well.

vrothberg:
@baude @jwhonce @edsantiago @mheon PTAL

edsantiago:
How hard would it be to add our own, undocumented, unsupported, `tlsVerify` to the API?

vrothberg:
> How hard would it be to add our own, undocumented, unsupported, `tlsVerify` to the API?

A simple 1 minute change. If that's acceptable, this would certainly be the quickest way forward.

mheon:
Can we restrict it so it's not available outside of the tests?

vrothberg:
> Can we restrict it so it's not available outside of the tests?

We could via an env var, for instance. But I think this would make it sufficiently complex to justify adding a `--registries-conf` flag.

mheon:
We have plenty of basically-undocumented unit-testing-only flags like that, so I'm on board

pkg/api/handlers/utils/images.go
```diff
@@ -1,19 +1,50 @@
...
+
+// GetXRegistryAuth extracts a DockerAuthConfig from the request's header if
+// present. Encoding errors are logged but yield nil (for Docker compat).
+func GetXRegistryAuth(r *http.Request) *types.DockerAuthConfig {
+	authHeader := r.Header.Get(XRegistryAuthHeader)
+	authConfig := &dockerAPITypes.AuthConfig{}
+	if len(authHeader) > 0 {
+		authJSON := base64.NewDecoder(base64.URLEncoding, strings.NewReader(authHeader))
+		if err := json.NewDecoder(authJSON).Decode(authConfig); err != nil {
+			logrus.Errorf("Error parsing %q header: %v", XRegistryAuthHeader, err)
```
jwhonce, line 34: I think it would be better to return a 400 to the client than ignore the header.

vrothberg: SGTM
Just to be extra sure: Docker is also just logging which is the motivation behind that. Propagating the error is much cleaner though.

vrothberg:
Back to WIP:

We now support both, single-auth headers and multi-auth headers. Multi-auth headers are the same as the ones from projectatomic docker. This way, we are compatible with vanilla Docker _and_ the projectatomic/RHEL Docker which we can/will also use for the new libpod endpoints.

It's now wired into the compat endpoints. Libpod endpoints are still missing along with unit, apiv2 and integration tests.

vrothberg:
Dev update:
* All encoding and decoding of the authentication headers is working now.
* Bindings have been updated for pull, push along with CLI parts. We can now also set headers in the bindings requests which accounts for the high amount of changed files.
* We can send credentials AND entire auth files now. This required some massaging of c/image and lead to https://github.com/containers/image/pull/928.
* Note that there's a new `pkg/auth` that the bindings and the sever share.

I'll continue on Monday and wire in support for all commands/endpoints that support `--authfile` and or specifying credentials. Once that's done, I will have a look at testing. I tried to get a local registry running in `apiv2-test` but failed miserably. I may need @edsantiago's bash skills to get that part done in a sane way.

Since so many tests require a local registry, I would love to have `make local-registry` target which runs a `registry:2` container with "known" credentials that may be dumped at a known location. This would be nice for local development but it would also prevent us from having duplicate code for setting up a registry across the e2e, system, api and potentially bindings tests. @edsantiago, is that something you could do (assuming there's time)?

vrothberg:
Note that we first need to get https://github.com/containers/image/pull/933 in.

vrothberg:
Also, once #6269 is merged, we can add auth tests.

cmd/podman/manifest/push.go
```diff
@@ -59,7 +70,24 @@ func push(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// TLS verification in c/image is controlled via a `types.OptionalBool`
+	// which allows for distinguishing among set-true, set-false, unspecified
+	// which is important to implement a sane way of dealing with defaults of
+	// boolean CLI flags.
+	if cmd.Flags().Changed("tls-verify") {
+		manifestPushOpts.SkipTLSVerify = types.NewOptionalBool(!manifestPushOpts.TLSVerifyCLI)
+	}
+	if err := registry.ImageEngine().ManifestPush(context.Background(), args, manifestPushOpts.ManifestPushOptions); err != nil {
```
jwhonce, line 90:
```suggestion
	if err := registry.ImageEngine().ManifestPush(registry.Context(), args, manifestPushOpts.ManifestPushOptions); err != nil {
```

pkg/domain/infra/abi/play.go
```diff
@@ -147,6 +150,13 @@ func (ic *ContainerEngine) PlayKube(ctx context.Context, path string, options en
 		writer = os.Stderr
 	}
 
+	if len(options.Username) > 0 && len(options.Password) > 0 {
```
jwhonce, line 153: It is not possible to have a username without password?

pkg/domain/infra/abi/images.go
```diff
@@ -226,12 +225,11 @@ func (ir *ImageEngine) Push(ctx context.Context, source string, destination stri
 	}
 
 	var registryCreds *types.DockerAuthConfig
-	if options.Credentials != "" {
-		creds, err := util.ParseRegistryCreds(options.Credentials)
-		if err != nil {
-			return err
+	if len(options.Username) > 0 && len(options.Password) > 0 {
```
jwhonce, line 228: It is not possible to have a username without password?

pkg/domain/infra/abi/images.go
```diff
@@ -121,12 +121,11 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
 	}
 
 	var registryCreds *types.DockerAuthConfig
-	if options.Credentials != "" {
-		creds, err := util.ParseRegistryCreds(options.Credentials)
-		if err != nil {
-			return nil, err
+	if len(options.Username) > 0 && len(options.Password) > 0 {
```
jwhonce, line 124: It is not possible to have a username without password?

pkg/bindings/play/play.go
```diff
@@ -2,16 +2,45 @@ package play
...
 	"os"
 	"strconv"
 
 	"github.com/containers/image/v5/types"
 	"github.com/containers/libpod/pkg/bindings"
+	authBindings "github.com/containers/libpod/pkg/bindings/auth"
 	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
```
jwhonce, line 17: How much does this grow the remote client?

pkg/auth/auth.go
```diff
@@ -0,0 +1,182 @@
...
+	conf := imageAuthToDockerAuth(authConfig)
+	buf, err := json.Marshal(conf)
+	if err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.EncodeToString(buf), nil
+}
+
+// EncodeMultiAuthConfigs serializes the auth configurations as a base64 encoded JSON payload.
+func EncodeMultiAuthConfigs(authConfigs map[string]types.DockerAuthConfig) (string, error) {
```
jwhonce, line 78: Usually this would be encoded as multiple headers, but I think that would break here because you would loose the registry name.  Maybe add a comment to that fact?

pkg/auth/auth.go
```diff
@@ -0,0 +1,182 @@
...
+		conf = &c
+		break
+	}
+	return conf, "", nil
+}
+
+// RemoveAuthfile is a convenience function that is meant to be called in a
+// deferred statement. If non-empty, it removes the specified authfile and log
+// errors.  It's meant to reduce boilerplate code at call sites of
+// `GetXRegistryAuth`.
```
jwhonce, line 57:
```suggestion
// `GetCredentials`.
```
Maybe?

pkg/auth/auth.go
```diff
@@ -0,0 +1,182 @@
...
+func RemoveAuthfile(authfile string) {
```
jwhonce, line 58: If you change signature to include the http.Request then you could add more context to the error message.  Or, runtime.Caller() would help debugging. Currently I don't know which endpoint is included in the failure.

pkg/api/server/register_registries.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http"
+
+	"github.com/containers/libpod/pkg/api/handlers/libpod"
+	"github.com/gorilla/mux"
+)
+
+func (s *APIServer) registerRegistriesHandlers(r *mux.Router) error {
+	// swagger:operation GET /libpod/registries/json libpod libpodListRegistries
+	// ---
+	// summary: Return a list of search registries and their mirrors.
```
jwhonce, line 13: missing `tags:` field

pkg/api/server/register_registries.go
```diff
@@ -0,0 +1,37 @@
...
+	// description: Return a sorted list of all search registries and their mirrors. The list can be used to determine which credentials the server may require.
```
jwhonce, line 14: Sort criteria?

pkg/bindings/play/play.go
vrothberg: This would effectively revert the changes from @baude and rebump from 33M to 41M.

pkg/auth/auth.go
vrothberg: That and it would break the projectatomic Docker and hence RHEL 7 -> 8 upgrade path.

vrothberg:
> Is identitytoken going to be implemented in a follow on PR?

Can you elaborate on that? Identity tokens are already supported via the c/image backends and accounted for in the header.

pkg/auth/auth.go
vrothberg: Good catch!

pkg/api/server/register_registries.go
vrothberg: I really want to ditch this endpoint in favour of sending all credentials (see https://github.com/containers/image/pull/942#issue-422757846).

Getting this to work is exploding in complexity and amount of work, so I hope we can settle on the all-credentials case.

pkg/bindings/play/play.go
vrothberg: I want to go back to the initial approach of sending all credentials over to the server (see https://github.com/containers/image/pull/942#issue-422757846). The amount of work and complexity is getting out of hand.

vrothberg: This would avoid repulling the k8s stuff back in :)

pkg/domain/infra/abi/images.go
vrothberg: Only looking for username now, thanks!

vrothberg:
Looking at the `podman stats with json output` flakes now.

vrothberg:
> Looking at the `podman stats with json output` flakes now.

Fixed in https://github.com/containers/libpod/pull/6403

vrothberg:
Note: I vendored a commit of c/image (non release). I agreed with @rhatdan to do this until c/image has new release. It's acceptable for an RC at least.

rhatdan:
LGTM

PR #6214, opened 2020-05-13
[CI:DOCS]remove libpod.conf from spec

baude:
in the contrib rpm.spec.in, we no longer should try to package libpod.conf


marusak:
`contrib/build_rpm.sh` did not work without this patch, works fine with it. Thanks!

mheon:
LGTM

PR #6215, opened 2020-05-13
Use the containers.conf cni_config_dir option for inspect and delete

### LINKED ISSUE #6212: Use of constant pkg/network/config.go#CNIConfigDir
/kind bug

**Description**

The files pkg/network/files.go and pkg/api/handlers/libpod/networks.go still make use of the static constant _CNIConfigDir_ in pkg/network/config.go.

The resulting issue appears when the cni config dir is changed in libpod.conf. While `network create` and `network ls` use pkg/domain/infra/abi/network.go#getCNIConfDir, `inspect` and `remove` use pkg/network/files.go#GetCNIConfigPathByName (for `inspect` indirectly via _ReadRawCNIConfByName_).

I will fix this by replacing the use of the constant in _GetCNIConfigPathByName_ by a call to _getCNIConfDir_.

maxm123:
The pkg/network/files.go methods currently use the constant '/etc/cni/net.d'
for network handling. This results in the unability of podman-network-inspect
and podman-network-rm to locate the cni network configuration files.
This commit propagates the libpod.Runtime through the networking methods and
finally makes use of its configuration (config.Network.NetworkConfigDir).
Closes #6212

Signed-off-by: Maximilian Müller <maxm123@techie.com>

### DIFF
--- a/pkg/domain/infra/abi/network.go
+++ b/pkg/domain/infra/abi/network.go
@@ -15,26 +15,15 @@ import (
 	"github.com/pkg/errors"
 )
 
-func getCNIConfDir(r *libpod.Runtime) (string, error) {
-	config, err := r.GetConfig()
-	if err != nil {
-		return "", err
-	}
-	configPath := config.Network.NetworkConfigDir
-
-	if len(config.Network.NetworkConfigDir) < 1 {
-		configPath = network.CNIConfigDir
-	}
-	return configPath, nil
-}
-
 func (ic *ContainerEngine) NetworkList(ctx context.Context, options entities.NetworkListOptions) ([]*entities.NetworkListReport, error) {
 	var reports []*entities.NetworkListReport
-	cniConfigPath, err := getCNIConfDir(ic.Libpod)
+
+	config, err := ic.Libpod.GetConfig()
 	if err != nil {
 		return nil, err
 	}
-	networks, err := network.LoadCNIConfsFromDir(cniConfigPath)
+
+	networks, err := network.LoadCNIConfsFromDir(network.GetCNIConfDir(config))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


maxm123:
The failure happens in container tests and consists of a `panic: send on closed channel`.

rhatdan:
libpod.conf is dead, containers.conf is what we use now.

rhatdan:
@mheon PTAL

pkg/network/files.go
```diff
@@ -9,9 +9,23 @@ import (
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/allocator"
+	"github.com/containers/libpod/libpod"
 	"github.com/pkg/errors"
 )
 
+func GetCNIConfDir(r *libpod.Runtime) (string, error) {
```
rhatdan, line 16: Rather then passing in the libpod and all of the code this pulls in, just pass in the config 
containers/common/pkg/config 
*config.Config

maxm123: I put your feedback into practice and now pass the config object.

pkg/domain/infra/abi/network.go
```diff
@@ -49,8 +38,14 @@ func (ic *ContainerEngine) NetworkInspect(ctx context.Context, namesOrIds []stri
 	var (
 		rawCNINetworks []entities.NetworkInspectReport
 	)
+
+	config, err := ctx.Value("runtime").(*libpod.Runtime).GetConfig()
```
QiWang19, line 42: can this be replaced with `ic.Libpod.GetConfig()`?

maxm123: Probably, yes. I've just seen that ctx.Value("runtime").(*libpod.Runtime) around there (e.g. for network creation) and I thought it was the way to do there.

QiWang19: I think that is mostly used for remote functionality, not used under pkg/domain/infra/abi

maxm123: Okay, I update this.

TomSweeneyRedHat:
LGTM, but definitely want a @mheon head nod.

QiWang19:
LGTM

PR #6217, opened 2020-05-13
Add manifest annotate endpoint

QiWang19:
Adds manifest annotate endpoint and enables tunnel mode


rhatdan:
@jwhonce @baude PTAL

baude:
@rhatdan initially you were dead set against this endpoint, which is why it was not initially implemented.  I do not remember why exactly.  Firstly, have you changed your mind?

rhatdan:
Ok thanks for reminding me.  In this case this is modification of an existing manifest correct? 
Lets talk at standup.

PR #6220, opened 2020-05-13
V2 Update attach bindings to use Readers/Writers vs chan

jwhonce:


pkg/bindings/containers/containers.go
```diff
@@ -373,11 +378,18 @@ func Attach(ctx context.Context, nameOrId string, detachKeys *string, logs, stre
 	}
 	defer response.Body.Close()
 
-	ctype := response.Header.Get("Content-Type")
-	upgrade := response.Header.Get("Connection")
+	// if stdin != nil {
+	// 	// Copy to stdin
+	// }
```
vrothberg, line 383: Looks like we don't need this

baude:
lgtm

PR #6228, opened 2020-05-14
Default podman.spec to use crun

rhatdan:
On systems with cgroupV2 runc will not currently work
switch the default in spec file to use crun.

### DIFF
--- a/contrib/spec/podman.spec.in
+++ b/contrib/spec/podman.spec.in
@@ -80,13 +80,14 @@ BuildRequires: libselinux-devel
 BuildRequires: pkgconfig
 BuildRequires: make
 BuildRequires: systemd-devel
-Requires: runc
 Requires: skopeo-containers
 Requires: containernetworking-plugins >= 0.6.0-3
 Requires: iptables
 %if 0%{?rhel} <= 7
 Requires: container-selinux
 %else
+Requires: oci-runtime
+Recommends: crun
 Recommends: container-selinux
 Recommends: slirp4netns
 Recommends: fuse-overlayfs


rhatdan:
Should fix https://github.com/containers/libpod/issues/6226

rhatdan:
@mheon @vrothberg @giuseppe @QiWang19 @jwhonce @TomSweeneyRedHat @baude PTAL

QiWang19:
LGTM

mheon:
LGTM

mheon:
We should make sure this works on Cent 7, though; this needs to work there for the rdocheck tests

TomSweeneyRedHat:
The change looks fine to me, I'm just a little concerned about distros without v2.  Can we get a head nod from @lsm5?

rhatdan:
crun works fine on non V2 systems, I did move the crun requirements and oci-runtime requirement inside of the Fedora block.

lsm5:
LGTM RE: crun.

On Fedora, I noticed the crun package `Provides: oci-runtime = 2`. So I think we should get rid of oci-runtime from podman spec as `Requires: oci-runtime` and `Recommends: crun` together doesn't seem right.

rhatdan:
Not sure what that means, oci-runtime=2 versus oci-runtime=1?  Bottom line podman requires an OCI Runtime to be installed.  I want that to be allowed to only install runc, crun, or kata-runtime.  I would prefer to have all three of these be provides oci-runtime, and then the user of podman could remove the others as he pleases.  Without requires oci-runtime, then the user could remove all runtimes just by eliminating reccomends, and stuf will break.

Issue #3906, opened 2019-08-29
Container storage leak when the podman process is SIGKILL


TristanCacqueray:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When the podman process is SIGKILL midway, the container storage leaks and it is not obvious how to clean it up.

**Steps to reproduce the issue:**

Run this script:

```python
#!/bin/python3

import subprocess, time

argv = ["podman", "run", "--rm", "--name", "test-server",
        "-e", "PYTHONUNBUFFERED=1",
        "fedora", "python3", "-m", "http.server"]

while True:
    try:
        print("Press ctrl-C if http server start...")
        proc = subprocess.Popen(argv)
        try:
            proc.communicate()
        except KeyboardInterrupt:
            proc.terminate()
            proc.kill()
            raise
    except KeyboardInterrupt:
        subprocess.Popen(["podman", "kill", "test-server"]).wait()
    time.sleep(1)
```

Sample output:
```bash
$ python3 repro.py 
Press ctrl-C if http server start...
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
^C
Keyboard interrupt received, exiting.
Error: no container with name or ID test-server found: no such container
Press ctrl-C if http server start...
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
^C
Keyboard interrupt received, exiting.
Error: no container with name or ID test-server found: no such container
Press ctrl-C if http server start...
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
^C
Keyboard interrupt received, exiting.
Error: no container with name or ID test-server found: no such container
Press ctrl-C if http server start...
Error: error creating container storage: the container name "test-server" is already in use by "478dbcf9e5f5d262ad6b21ad3fe94907260c086b718ce29d54d05a311caf2abf". You have to remove that container to be able to reuse that name.: that name is already in use
Press ctrl-C if http server start...
Error: error creating container storage: the container name "test-server" is already in use by "478dbcf9e5f5d262ad6b21ad3fe94907260c086b718ce29d54d05a311caf2abf". You have to remove that container to be able to reuse that name.: that name is already in use
```

**Describe the results you received:**

Once this happens, to be able to re-use the `test-server` name, the container storage needs to be removed manually:
```bash
$ podman container inspect test-server
[...]

PR #6229, opened 2020-05-14
Cleanup OCI runtime before storage

mheon:
Some runtimes (e.g. Kata containers) seem to object to having us unmount storage before the container is removed from the runtime. This is an easy fix (change the order of operations in cleanup) and seems to make more sense than the way we were doing things.

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1562,21 +1562,24 @@ func (c *Container) cleanup(ctx context.Context) error {
 		lastError = errors.Wrapf(err, "error removing container %s network", c.ID())
 	}
 
+	// Remove the container from the runtime, if necessary.
+	// Do this *before* unmounting storage - some runtimes (e.g. Kata)
+	// apparently object to having storage removed while the container still
+	// exists.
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

snir911:
doesn't fix the issue :/

mheon:
Still worth merging, IMO, this makes more sense than the way we did it before.

mheon:
This does sound like a Kata bug as this point, from what I'm hearing. Podman, on stopping a container, calls the OCI runtime's `kill` (which should guarantee the container and associated VM stops), and then deletes it from the runtime, and then unmounts storage. This sounds like a correct workflow.

TomSweeneyRedHat:
LGTM
bummer it doesn't fix the bug, but a good change regardless.

mheon:
I pushed another commit that might actually resolve it, but haven't checked

snir911:
I  can confirm this is working, i had a similar version but i was not sure if doing the cleanup before setting the Removing state will make Removing state redundant (or even to regress the bug< #3906> was fixed with 25cc43c)

PR #6230, opened 2020-05-14
Fix REMOTETAGS

rhatdan:
Handle REMOTETAGS the same way for all remote commands.

This fixes issues where remote commands are not building correctly on rhel7 and centos7 systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ ETCDIR ?= /etc
 TMPFILESDIR ?= ${PREFIX}/lib/tmpfiles.d
 SYSTEMDDIR ?= ${PREFIX}/lib/systemd/system
 USERSYSTEMDDIR ?= ${PREFIX}/lib/systemd/user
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -22,6 +22,7 @@ ETCDIR ?= /etc
 TMPFILESDIR ?= ${PREFIX}/lib/tmpfiles.d
 SYSTEMDDIR ?= ${PREFIX}/lib/systemd/system
 USERSYSTEMDDIR ?= ${PREFIX}/lib/systemd/user
+REMOTETAGS := !ABISupport remoteclient exclude_graphdriver_btrfs btrfs_noversion  exclude_graphdriver_devicemapper containers_image_openpgp
```
mheon, line 25: Should we use the hack/ scripts to determine whether btrfs, seccomp, etc are included?

rhatdan: No, We don't want this code compiled into remoteclient. In a perfect world we would not hit this.  The btrfs and devicemapper flags are for containers/image and containers/storage and prevent btrfs and devicemapper from being compiled into the client.

TomSweeneyRedHat:
LGTM assuming happy tests

mheon:
Manually setting hold and LGTM

PR #6231, opened 2020-05-14
Fix two coverity issues (unchecked null return)

mheon:
Theoretically these should never happen, but it never hurts to be sure and check. Add a check to one, make the other one a create-if-not-exist (it was just adding, not checking the contents).


mheon:
@rhatdan PTAL

TomSweeneyRedHat:
Changes LGTM, but the test aren't happy atm.  I'm not sure if it's related.

PR #6232, opened 2020-05-14
system tests: small fixes for rawhide+cgroups v1

edsantiago:
Three small fixes for breaking tests on rawhide:

  1) run test: looks like runc changed the format of
     an error message, adding a colon in one place.
     runc is used on rawhide when booted in cgroups v1

  2) volumes test: difference in exit status and error
     message between runc and crun.

  3) systemd test: define XDG_RUNTIME_DIR if unset.
     podman helpfully sets this to a reasonable default,
     but the 'systemctl' commands used in this test do not.

### DIFF
--- a/test/system/160-volumes.bats
+++ b/test/system/160-volumes.bats
@@ -116,8 +116,17 @@ EOF
     chmod 755 $mountpoint/myscript
 
     # By default, volumes are mounted noexec. This should fail.
-    run_podman 126 run --rm --volume $myvolume:/vol:z $IMAGE /vol/myscript
-    is "$output" ".* OCI runtime permission denied.*" "run on volume, noexec"
+    # ARGH. Unfortunately, runc (used for cgroups v1) produces a different error
+    local expect_rc=126
+    local expect_msg='.* OCI runtime permission denied.*'
+    run_podman info --format '{{ .Host.OCIRuntime.Path }}'
+    if expr "$output" : ".*/runc"; then
+        expect_rc=1
+        expect_msg='.* exec user process caused.*permission denied'
+    fi
+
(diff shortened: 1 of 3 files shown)


test/system/160-volumes.bats
```diff
@@ -116,8 +116,17 @@ EOF
...
 
     # By default, volumes are mounted noexec. This should fail.
-    run_podman 126 run --rm --volume $myvolume:/vol:z $IMAGE /vol/myscript
-    is "$output" ".* OCI runtime permission denied.*" "run on volume, noexec"
+    # ARGH. Unfortunately, runc (used for cgroups v1) produces a different error
+    local expect_rc=126
+    local expect_msg='.* OCI runtime permission denied.*'
+    run_podman info --format '{{ .Host.OCIRuntime.Path }}'
+    if expr "$output" : ".*/runc"; then
+        expect_rc=1
```
edsantiago, line 124: @giuseppe and/or someone runc-savvy, I'd really like an extra set of eyes on this. I believe `crun` is doing the right thing, per `podman-run(1)`, but I'm uncomfortable with the unexpected `runc` exit status. I wonder if there's a way podman can trap that and convert `1` to `126`, and if so, should it?

mheon: This sounds like a Podman bug - we have code that munges the error messages (string wrangling - it's horrible and ugly but the runtime doesn't give us anything else to work with) and it probably isn't handling the runc error string properly

TomSweeneyRedHat:
LGTM
Can we adjust for the Podman bug @mheon mentions post merge?

mheon:
IMO, let's drop a TODO on that issue, then merge.

PR #6236, opened 2020-05-14
Make convenience boxed true/false easier to use

jwhonce:
Small improvement for working with the API bindings.

* changed PFalse to &false
* changed PTrue to &true

### DIFF
--- a/pkg/bindings/bindings.go
+++ b/pkg/bindings/bindings.go
@@ -11,8 +11,10 @@ package bindings
 var (
 	// PTrue is a convenience variable that can be used in bindings where
 	// a pointer to a bool (optional parameter) is required.
-	PTrue bool = true
+	pTrue = true
+	PTrue = &pTrue
 	// PFalse is a convenience variable that can be used in bindings where
 	// a pointer to a bool (optional parameter) is required.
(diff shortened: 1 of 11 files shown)


pkg/bindings/bindings.go
```diff
@@ -11,8 +11,10 @@ package bindings
 var (
 	// PTrue is a convenience variable that can be used in bindings where
 	// a pointer to a bool (optional parameter) is required.
-	PTrue bool = true
+	pTrue = true
+	PTrue = &pTrue
 	// PFalse is a convenience variable that can be used in bindings where
```
TomSweeneyRedHat, line 16: PFalse and PTrue in comments need to change to pFalse and pTrue

TomSweeneyRedHat: oops, never mind, I should look at line two for each of these.

TomSweeneyRedHat:
LGTM assuming happy tests

pkg/bindings/bindings.go
jwhonce: @TomSweeneyRedHat The comments are more meaningful now. We don't want people using pTrue or pFalse.

PR #6254, opened 2020-05-17
WIP Save authfile in container config.

neVERberleRfellerER:
This is one possible approach usable as base for https://github.com/containers/libpod/pull/6188 .

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -414,6 +414,8 @@ type ContainerConfig struct {
 
 	// HealthCheckConfig has the health check command and related timings
 	HealthCheckConfig *manifest.Schema2HealthConfig `json:"healthcheck"`
+
+	Authfile string `json:"authfile"`
 }
 
 // ContainerNamedVolume is a named volume that will be mounted into the
--- a/libpod/define/container_inspect.go
+++ b/libpod/define/container_inspect.go
@@ -510,6 +510,8 @@ type InspectContainerHostConfig struct {
 	IOMaximumIOps uint64 `json:"IOMaximumIOps"`
 	// IOMaximumBandwidth is Windows-only and not presently implemented.
 	IOMaximumBandwidth uint64 `json:"IOMaximumBandwidth"`
+	//Authfile is path to used authfile
+	Authfile string `json:"AuthFile"`
 }
 
 // InspectBasicNetworkConfig holds basic configuration information (e.g. IP
--- a/pkg/specgen/specgen.go
+++ b/pkg/specgen/specgen.go
@@ -130,6 +130,8 @@ type ContainerBasicConfig struct {
 	// Remove indicates if the container should be removed once it has been started
 	// and exits
 	Remove bool `json:"remove"`
+	//Authfile is path to authfile to use
+	Authfile string `json:"authfile,omitempty"`
 }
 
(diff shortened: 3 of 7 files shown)


libpod/container.go
```diff
@@ -414,6 +414,8 @@ type ContainerConfig struct {
 
 	// HealthCheckConfig has the health check command and related timings
 	HealthCheckConfig *manifest.Schema2HealthConfig `json:"healthcheck"`
+
+	Authfile string `json:"authfile"`
```
vrothberg, line 418: Please add a comment.

libpod/define/container_inspect.go
```diff
@@ -510,6 +510,8 @@ type InspectContainerHostConfig struct {
 	IOMaximumIOps uint64 `json:"IOMaximumIOps"`
 	// IOMaximumBandwidth is Windows-only and not presently implemented.
 	IOMaximumBandwidth uint64 `json:"IOMaximumBandwidth"`
+	//Authfile is path to used authfile
```
vrothberg, line 513:
```suggestion
	// Authfile is the path to the auth file used at container creation.
```

pkg/specgen/specgen.go
```diff
@@ -130,6 +130,8 @@ type ContainerBasicConfig struct {
 	// Remove indicates if the container should be removed once it has been started
 	// and exits
 	Remove bool `json:"remove"`
+	//Authfile is path to authfile to use
```
vrothberg, line 133:
```suggestion
	// Authfile is the path to auth file to use
```

vrothberg:
@mheon PTAL Cool to put that in the config or shall we use a label?

vrothberg:
Note that it's important to only store the auth file if it was actually specified by the user on the CLI via `--authfile`. We shouldn't store the default one. This way we are making sure that we can establish a priority for which authfile is being used during auto updates.

mheon:
@vrothberg I'll admit I haven't been paying attention to the previous discussion here - what's the logic for making this a part of container config? Pros/cons vs. using labels? My initial instinct is that we should use a label, but I could be convinced otherwise.

vrothberg:
> @vrothberg I'll admit I haven't been paying attention to the previous discussion here - what's the logic for making this a part of container config? Pros/cons vs. using labels? My initial instinct is that we should use a label, but I could be convinced otherwise.

@neVERberleRfellerER is currently working on wiring authfile support into `podman auto-update`. A `podman auto-update --authfile=...` is already a big step forward but it's not quite enough for @neVERberleRfellerER's use case. Different containers might be pulled from the same registry but from different namespaces which may require different credentials. Hence, one authfile isn't enough.

The idea now is to store the authfile at container-creation time. During auto-update, we can look up if a container was created with a non-default authfile and use that instead when contacting the registry.

I recall that you preferred storing certain data in labels rather than extending the container config. This may be such a case.

mheon:
My initial impression from this is that it should probably be a label similar to the ones we use for Inspect options (it's not strictly necessary for Libpod operation, but if it's there we'll know what to do with it).

neVERberleRfellerER:
Since https://github.com/containers/libpod/pull/6188 has been merged this isn't useful anymore. Let me know if you want authfile in container config for different reasons.
@vrothberg Thank you for everything.

PR #6256, opened 2020-05-18
Support k8s Deployment in play kube

theunrealgeek:
This PR is adding support for `Deployment` k8s Kind in addition to the already supported `Pod` kind and is in response to https://github.com/containers/libpod/issues/4093

`Deployment` kinds generally just spit out one or more `Pod`s from a template spec. Hence I have reused most of the existing code for `Pod` and just created multiple of these dictated by the replica count specified in the `Deployment` YAML.

The only breaking change in this PR is that container names from `play kube` now have the name of the pod prefixed to them since `podman` needs unique container names overall and not unique container names per-pod as is the expectation from Kubernetes.


rhatdan:
Thanks @theunrealgeek 
This looks like it breaks a bunch of tests?
@haircommander @baude @mheon @giuseppe @umohnani8 @mrunalp PTAL

cmd/podman/play/kube.go
```diff
@@ -84,7 +84,18 @@ func kube(cmd *cobra.Command, args []string) error {
 		fmt.Fprintf(os.Stderr, l)
 	}
 
-	fmt.Printf("Pod:\n%s\n", report.Pod)
+	switch len(report.Pods) {
+	case 0:
+		return nil
```
mheon, line 89: Hm. Are there cases where containers could be made, but not pods? We might want to make 0 print nothing but not return yet

haircommander:
this looks awesome @theunrealgeek ! can you add some tests in the style of the tests in `test/e2e/play_kube.go`?

cmd/podman/play/kube.go
haircommander: I actually disagree, there really shouldn't be a case containers are created and pods are not. I think return early here if no printing is needed

cmd/podman/play/kube.go
```diff
@@ -84,7 +84,18 @@ func kube(cmd *cobra.Command, args []string) error {
...
+	case 1:
+		fmt.Printf("Pod:\n")
+	default:
+		fmt.Printf("Pods:\n")
+	}
+	for _, pod := range report.Pods {
```
haircommander, line 95: is there any way to print the containers in the pods underneath their pod? Right now, this looks like it will do
```
Pods:
pod-1
pod-2
Containers:
container-1-1
container-2-1
container-1-2
container-2-2
```
where I think it'd be clearer if we did
```
Pods:
pod-1
Containers:
container-1-1
container-2-1

pod-2
Containers:
container-1-2
container-2-2
```

pkg/domain/infra/abi/play.go
```diff
@@ -359,6 +414,11 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
 	containerConfig.Image = containerYAML.Image
 	containerConfig.ImageID = newImage.ID()
 	containerConfig.Name = containerYAML.Name
+
+	if podName != "" {
```
haircommander, line 418: maybe fail here if podName is not specified?

pkg/domain/infra/abi/play.go
```diff
@@ -359,6 +414,11 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
...
+		containerConfig.Name = fmt.Sprintf("%s-%s", podName, containerYAML.Name)
+	}
+
```
TomSweeneyRedHat, line 421: We've some examples in the man page that probably ought to be touched up based on these changes.

cmd/podman/play/kube.go
theunrealgeek: Good idea, I'll make this change

pkg/domain/infra/abi/play.go
theunrealgeek: Will do

theunrealgeek:
> this looks awesome @theunrealgeek ! can you add some tests in the style of the tests in `test/e2e/play_kube_test.go`?

Didn't realise there were tests for this, I'll add some new tests for the new stuff and fix any existing ones that break.

rhatdan:
@theunrealgeek Please rebase this PR

theunrealgeek:
I'm done fixing the existing tests and addressing most of the review comments. I need to add a few more tests with a `Deployment` YAML and will hopefully update the PR with all of this tomorrow

theunrealgeek:
@haircommander Can you please take a look at this again ? I've addressed most of the comments. There are still some failing CI tests which don't seem related to anything I have touched, but if they are, please let me know and I'll take a look at them.

pkg/domain/infra/abi/play.go
theunrealgeek: @TomSweeneyRedHat I looked at [this](https://github.com/containers/libpod/blob/master/docs/source/markdown/podman-play-kube.1.md) which seemed like what you were talking about, but there doesn't seem to be anything related to the name of the containers in there which might need changing.

Should I add something about this in there ?

test/e2e/play_kube_test.go
```diff
@@ -14,6 +14,17 @@ import (
...
+kind: UnknownKind
+metadata:
+  labels:
+    app: app1
+  name: unknown
+spec:
+  hostname: unknown
+`
+
 var yamlTemplate = `
```
haircommander, line 28: nit: I would probably change this to `podYAMLTemplate` now

test/e2e/play_kube_test.go
```diff
@@ -77,14 +88,109 @@ spec:
...
+
+	_, err = f.WriteString(content)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func generateKubeYaml(pod *Pod, fileName string) error {
```
haircommander, line 194: similarly, this should probably generatePodKubeYaml

haircommander: also, if there's any redundancy in this function with generateDeploymentKubeYaml, It'd be ideal if the similarities could be put in a function.  I see you've made writeYAML, but haven't used it in generateKubeYaml

haircommander, line 194: second nit for this line: we should stick to either YAML or Yaml. I don't care which, but if you'd prefer to use YAML, can you change all of them?

haircommander:
@theunrealgeek excellent thank you. a couple more nits but all together looks great to me :+1: :smiley:

haircommander:
(also, I agree, test failures look like registry issues. I've tried retriggering them)

pkg/domain/infra/abi/play.go
TomSweeneyRedHat: @theunrealgeek my bad, I thought the output in those examples were getting a dash added to them.  Ignore this.

rhatdan:
@theunrealgeek Any progress?

mheon:
Tests are fixed on master, FYI - a rebase should pick up the fix

theunrealgeek:
@rhatdan Sorry about the delay, got involved in a prod incident on my day job.

@haircommander I've addressed your comments. Thank you for your review :)

rhatdan:
@theunrealgeek No problem, thanks for the contribution.
@haircommander @umohnani8 @baude @mrunalp PTAL

pkg/domain/infra/abi/play.go
```diff
@@ -382,7 +445,10 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
 
 	setupSecurityContext(&securityConfig, &userConfig, containerYAML)
 
-	securityConfig.SeccompProfilePath = seccompPaths.findForContainer(containerConfig.Name)
+	// Since we prefix the container name with pod name to work-around the uniqueness requirement,
+	// seccom stuff should reference the actual container name from the YAML
```
haircommander, line 449: nit: s/seccom stuff/the seccomp profile

theunrealgeek: Done :+1:

pkg/domain/entities/play.go
```diff
@@ -26,12 +26,17 @@ type PlayKubeOptions struct {
 	SeccompProfileRoot string
 }
 
-// PlayKubeReport contains the results of running play kube.
-type PlayKubeReport struct {
-	// Pod - the ID of the created pod.
-	Pod string
+// PlayKubePods represents a single pod and associated containers created by play kube
+type PlayKubePod struct {
+	ID string
```
vrothberg, line 31: Please add a comment what ID is.

pkg/domain/entities/play.go
```diff
@@ -26,12 +26,17 @@ type PlayKubeOptions struct {
 	SeccompProfileRoot string
 }
 
-// PlayKubeReport contains the results of running play kube.
-type PlayKubeReport struct {
-	// Pod - the ID of the created pod.
-	Pod string
+// PlayKubePods represents a single pod and associated containers created by play kube
```
vrothberg, line 29:
```suggestion
// PlayKubePod represents a single pod and associated containers created by play kube
```

pkg/domain/entities/play.go
theunrealgeek: @vrothberg Good catch, I've fixed it

haircommander:
LGTM :smiley:

cmd/podman/play/kube.go
```diff
@@ -92,21 +92,29 @@ func kube(cmd *cobra.Command, args []string) error {
...
-	}
-	for _, ctr := range report.Containers {
-		fmt.Println(ctr)
+	for _, pod := range report.Pods {
+		fmt.Printf("Pod:\n")
+		fmt.Println(pod.ID)
+
+		switch len(pod.Containers) {
+		case 0:
+			return nil
```
mheon, line 107: Should this be a `continue` so we print all pods in the case that there are more than 1?

mheon:
One last nit, otherwise LGTM. Clear to merge once fixed.

Issue #6261, opened 2020-05-18
API v2: delete an image


Edward5hen:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
As the [doc](http://docs.podman.io/en/latest/_static/api.html#operation/libpodRemoveImage) says, `/libpod/images/{name:.*}/remove` should delete an image with DELETE method. But I get a "no such image" 404 code.

**Steps to reproduce the issue:**

1. setup the apiv2 service

2. podman pull alpine

3. curl -v -X DELETE http://localhost:8081/v1.40/libpod/images/f70734b6a266/remove

**Describe the results you received:**
```
[root@Fedora31 ~]# podman images
REPOSITORY                 TAG      IMAGE ID       CREATED       SIZE
docker.io/library/alpine   latest   f70734b6a266   2 weeks ago   5.88 MB
[root@Fedora31 ~]# curl -v -X DELETE http://localhost:8081/v1.40/libpod/images/f70734b6a266/remove
*   Trying ::1:8081...
* TCP_NODELAY set
* connect to ::1 port 8081 failed: Connection refused
*   Trying 127.0.0.1:8081...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8081 (#0)
> DELETE /v1.40/libpod/images/f70734b6a266/remove HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.66.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Content-Type: application/json
< Date: Thu, 14 May 2020 05:12:53 GMT
< Content-Length: 182
<
{"cause":"no such image","message":"Failed to find image f70734b6a266/remove: unable to find a name and tag match for f70734b6a266/remove in repotags: no such image","response":404}
* Connection #0 to host localhost left intact
```

**Describe the results you expected:**
Remove the image successfully

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
[root@Fedora31 ~]# podman version
Version:            1.9.1
RemoteAPI Version:  1
Go Version:         go1.13.9
Built:              Mon May 11 06:37:25 2020
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
[root@Fedora31 ~]# podman info --debug
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.13.9
[...]

PR #6264, opened 2020-05-18
v2endpoint remove image path correction

baude:
the endpoint for single image removal (on the libpod side) should be as follows:

versionedPath/libpod/images/IMAGENAME

The DELETE method then signifies the removal of the image.

Fixes: #6261


pkg/api/handlers/libpod/images.go
```diff
@@ -602,6 +602,11 @@ func CommitContainer(w http.ResponseWriter, r *http.Request) {
 	options.Changes = query.Changes
 	ctr, err := runtime.LookupContainer(query.Container)
 	if err != nil {
+		fmt.Println("//////////////")
+		fmt.Println(query)
+		fmt.Println("//////////////")
+		fmt.Println("//////////////")
+		fmt.Println("//////////////")
```
vrothberg, line 609: That might be a bit too verbose :^)

Issue #6237, opened 2020-05-14
Building with podman-remote hangs


afbjorklund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Trying to do a `podman build` over podman-remote, and the connection gets stuck.

This happens with both 1.9.0 and 1.9.1, but works OK with 1.8.2. On Ubuntu 19.10

The workaround is running "sudo podman build" remotely (over ssh), which works.

But that means that the user needs to transport the entire build context themselves.

From https://github.com/kubernetes/minikube/issues/8063

Tried to debug, but not sure how ?

**Steps to reproduce the issue:**

1. `minikube start --driver podman`

2. `eval $(minikube podman-env)`

3. `podman-remote build .`

**Describe the results you received:**

```
 podman-remote build .
^C
```

**Describe the results you expected:**

```
 podman-remote build .
STEP 1: FROM busybox
STEP 2: RUN true
--> Using cache 98bc298a4242e40d67a4f37d2a2d37f2cf7ab3b71e1499a67a278bd131e20398
98bc298a4242e40d67a4f37d2a2d37f2cf7ab3b71e1499a67a278bd131e20398
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:            1.9.1
RemoteAPI Version:  1
Go Version:         go1.12.10
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman/unknown,now 1.9.1~1 amd64 [installed]
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

mheon:
@baude PTAL

baude:
can you provide the dockerfile in question?

afbjorklund:
Here it is:

```Dockerfile
FROM busybox
RUN true
```

afbjorklund:
Thanks @baude:

I tried upgrading (to `904ee39`) and it seems to have fixed the first issue.

But it now seems to hang on the second step, and eventually time-out:

```
STEP 1: FROM busybox
Getting image source signatures
Copying blob sha256:d9cbbca60e5f0fc028b13c01e70e96cc2eefe7858ccc5a33c9a8faa8b03286aa
Copying config sha256:78096d0a54788961ca68393e5f8038704b97d8af374249dc5c8faec1b8045e42
[...]

PR #6268, opened 2020-05-18
 Fix EOM for SendFile

baude:
To terminate a connection of varlink, say after sending a file, we need to send a message containing a delimiter of ':' so the client knows to hang up.

Fixes: #6237


mheon:
Gating is angry. Otherwise LGTM

PR #6269, opened 2020-05-18
New tool: hack/podman-registry, manages local registry

edsantiago:
In response to #6207: this is a helper script intended for
use in starting and stopping a local container registry.
It takes care of port, username, password assignments;
generates a self-signed certificate; and starts the
container in an isolated podman root/runroot to avoid
conflicting with the caller's environment.

Intended usage: invoke from shell script, using 'eval'
to get results into calling process environment. See
help message (-h) for invocation details. This will
work for shell scripts but will be difficult if
called from Go or C - if that is likely to happen,
I'd love to hear suggestions for alternate ways to
get the settings back to the caller.

### DIFF
--- a/hack/podman-registry
+++ b/hack/podman-registry
@@ -1,0 +1,231 @@
+#! /bin/bash
+#
+# podman-registry - start/stop/monitor a local instance of registry:2
+#
+ME=$(basename $0)
+
+###############################################################################
+# BEGIN defaults
+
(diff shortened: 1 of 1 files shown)


edsantiago:
@vrothberg please lmk if this meets your needs. If the result-passing mechanism won't work for you, I'm open to alternatives.

edsantiago:
Closed in favor of #6273

PR #6270, opened 2020-05-18
Implement detached exec

mheon:
Implement detached exec for containers (ability to run a command in a container without attaching to it).

TODO:
- Support for exec exit commands has been added, but they are not yet being used. Modify `podman container cleanup` so it can clean up an individual exec session and generate exit commands using it.
- The APIv2 endpoint will need to be changed to use the new code once exit command generation is ready

This is mostly complete as it, though. `podman exec -d` is working; it's just never getting rid of the exec session without a container restart or removal.


mheon:
Cleanup processes implemented.

mheon:
Problem: we don't know enough at time of creation to know if we should add a cleanup process to remove the exec session. Thus, this works for local, but not remote.

I'm going to fix gating, add some local-only tests, and un-mark WIP so I can tackle this issue in a separate PR.

mheon:
Test added. Dropping WIP.

mheon:
@rhatdan @baude @vrothberg @giuseppe @TomSweeneyRedHat PTAL

mheon:
Would like a nod from @jwhonce on the pkg/domain changes as well

PR #6271, opened 2020-05-19
V2 API Version Support

jwhonce:
* Update blang/semver to allow ParseTolerant() support
* Provide helper functions for API handlers to obtain client's 'version'
  path variable
* Introduce new errors:
  * version not given in path, endpoints may determine if this is a hard
    error (ErrVersionNotGiven)
  * given version not supported (ErrVersionNotSupported), only a soft
    error if the handler is going to hijack the connection
* Added unit tests for version parsing
* bindings check version on connect:
  * client <= Server API version connection is continued
  * client >= Server API version connection fails


pkg/api/handlers/handler.go
```diff
@@ -1,6 +1,14 @@
...
-	DefaultApiVersion = "1.40" // See https://docs.docker.com/engine/api/v1.40/
-	MinimalApiVersion = "1.24"
+import (
+	"github.com/blang/semver"
+)
+
+var (
+	// See https://docs.docker.com/engine/api/v1.40/
+	// libpod compat handlers are expected to honor docker API versions
+	MaxApiVersion     = semver.MustParse("1.40.0") // Maximum supported Libpod API version must be >= docker API
```
mheon, line 10: Should we split this for Compat and Libpod? I don't know if I want us to be beholden to Docker versioning for Libpod

jwhonce: If we don't at least draft with them, we're creating a blocking change moving between compat and libpod.

mheon: I think we'll already have problems with people checking Docker version (Podman v2.0.0 vs Docker 19.05), so we can't guarantee full compatibility. Though this is probably a conversation for watercooler tomorrow.

pkg/bindings/connection.go
```diff
@@ -154,8 +155,20 @@ func pingNewConnection(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		switch APIVersion.Compare(v) {
+		case 1, 0:
+			// Server's job when client version is equal or older
+			return nil
+		case -1:
+			return errors.Errorf("server API version is too old. client %s server %s", APIVersion.String(), v.String())
```
vrothberg, line 170: I enjoy quoted strings in logs:
```suggestion
			return errors.Errorf("server API version is too old. client %q server %q", APIVersion.String(), v.String())
```

pkg/api/handlers/handler.go
vrothberg, line 10: Please add doc comments to all exported variables.

vrothberg: I won't be able to join the water cooler, so I am sharing my thoughts here.

I think it's important to be able (client & server) to communicate which versions of Docker are supported and which version of Podman. If they overlap, clients may know if the server actually supports what they want and the server may not be able to tell the client if they are supported or not. Mixing the two or letting them overlap would erase (to a certain degree) the expressiveness of semver.

If a user/costumer wants to figure out which versions (of Docker) a given Podman supports, they should see that with a simple `podman info` which could list a version range for Docker and another one for the Podman/libpod endpoints. If they overlapped, the data would not answer their question. This would make the lives of support much easier, especially for customer cases.

mheon:
LGTM

PR #6273, opened 2020-05-19
Registry helper : go bindings

vrothberg:
Based on https://github.com/containers/libpod/pull/6269. Adding go-bindings.

### DIFF
--- a/hack/podman-registry-go/registry.go
+++ b/hack/podman-registry-go/registry.go
@@ -1,0 +1,98 @@
+package registry
+
+import (
+	"strings"
+
+	"github.com/containers/libpod/utils"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	imageKey = "PODMAN_REGISTRY_IMAGE"
+	userKey  = "PODMAN_REGISTRY_USER"
+	passKey  = "PODMAN_REGISTRY_PASS"
+	portKey  = "PODMAN_REGISTRY_PORT"
+)
+
+var binary = "podman-registry"
+
+// Registry is locally running registry.
+type Registry struct {
+	// Image - container image of the registry.
+	Image string
+	// User - the user to authenticate against the registry.
+	User string
+	// Password - the accompanying password for the user.
+	Password string
+	// Port - the port the registry is listening to on the host.
+	Port string
+	// running indicates if the registry is running.
+	running bool
+}
+
+// Start a new registry and return it along with it's image, user, password, and port.
+func Start() (*Registry, error) {
+	// Start a registry.
+	out, err := utils.ExecCmd(binary, "start")
+	if err != nil {
+		return nil, errors.Wrapf(err, "error running %q: %s", binary, out)
+	}
+
+	// Parse the output.
+	registry := Registry{}
+	for _, s := range strings.Split(out, "\n") {
+		if s == "" {
+			continue
+		}
+		spl := strings.Split(s, "=")
+		if len(spl) != 2 {
+			return nil, errors.Errorf("unexpected output format %q: want 'PODMAN_...=...'", s)
+		}
+		key := spl[0]
+		val := strings.TrimSuffix(strings.TrimPrefix(spl[1], "\""), "\"")
+		switch key {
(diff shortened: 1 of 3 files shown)


hack/podman-registry-go/registry_test.go
```diff
@@ -0,0 +1,31 @@
...
+		require.NoError(t, err)
+		assert.True(t, len(reg.Image) > 0)
+		assert.True(t, len(reg.User) > 0)
+		assert.True(t, len(reg.Password) > 0)
+		assert.True(t, len(reg.Port) > 0)
+		registries = append(registries, reg)
+	}
+
+	// Stop registries.
+	for _, reg := range registries {
```
edsantiago, line 27: Would it make sense to add a `teardown` that repeats this loop, but without error checks? My concern is making sure we clean up even in the case of an assertion failure.

Bonus points for doing it cleanly, e.g. by removing from `registries` in the loop below, or by having `reg.Stop()` set an already-stopped flag, etc.

hack/podman-registry-go/registry_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Start registries.
+	for i := 0; i < 3; i++ {
+		reg, err := Start()
+		require.NoError(t, err)
+		assert.True(t, len(reg.Image) > 0)
+		assert.True(t, len(reg.User) > 0)
+		assert.True(t, len(reg.Password) > 0)
+		assert.True(t, len(reg.Port) > 0)
+		registries = append(registries, reg)
```
edsantiago, line 23: If you do add a teardown as suggested below, it might make sense to reorder the above as:
````
    assert(Port)
    append(reg)
    assert(all the rest)
````
That way `teardown` will have the best possible `registries` array early on.

File: hack/podman-registry-go/registry.go [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+func Start() (*Registry, error) {
+	// Start a registry.
+	out, err := utils.ExecCmd(binary, "start")
+	if err != nil {
+		return nil, errors.Wrapf(err, "error running %q: %s", binary, out)
+	}
+
+	// Split the output.
+	outSpl := strings.Split(out, "\n")
+	if len(outSpl) != 5 {
```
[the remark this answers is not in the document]
vrothberg: That's a fair point. I took the shortcut but will change and do proper parsing.

hack/podman-registry-go/registry_test.go
vrothberg: I changed it to use a multierror. Good catch!

The registry is now also carrying an internal state field to know it was already stopped.

hack/podman-registry-go/registry_test.go
vrothberg: Accounted for that with a multi error :+1:

vrothberg:
@baude @edsantiago @jwhonce @mheon PTAL

edsantiago:
LGTM. I wish there wasn't so much duplication.

hack/podman-registry-go/registry.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, s := range strings.Split(out, "\n") {
+		if s == "" {
+			continue
+		}
+		spl := strings.Split(s, "=")
+		if len(spl) != 2 {
+			return nil, errors.Errorf("unexpected output format %q: want 'PODMAN_...=...'", s)
+		}
+		key := spl[0]
+		val := strings.TrimSuffix(strings.TrimPrefix(spl[1], "\""), "\"")
```
jwhonce, line 53:
```suggestion
		val := strings.TrimSuffix(strings.TrimPrefix(spl[1], `\`), `\`)
```
Is a little cleaner to read.

hack/podman-registry-go/registry.go
```diff
@@ -0,0 +1,98 @@
...
+
+	registry.running = true
+
+	return &registry, nil
+}
+
+// Stop the registry.
+func (r *Registry) Stop() error {
+	// Stop a registry.
+	if !r.running {
```
jwhonce, line 90: +1

Issue #5736, opened 2020-04-06
podman info: should include podman version


edsantiago:
The `podman info` command does not actually include podman's own version.

afbjorklund:
I guess it is bug-compatible with docker that way ? (it only includes server version)

afbjorklund:
Interestingly, `podman-remote info` _does_ include the podman version.  But only for the client!

The host part of the output is the same as `podman info`, in that it only mentions Buildah version.

alitvak69:
podman info --debug includes podman version

afbjorklund:
It has another block of info now, so I suppose this is fixed.

before: (without the actual date, for brevity)
```
host:
registries:
store:
```

after: (without the actual date, for brevity)
```
host:
registries:
store:
version:
```

Added in 62803f8d3bacaea9956e6ec3176b223b85d260a5

The main difference seems to be date format.


podman info (partial)
```
version:
  Built: 1589615354
  GitCommit: 343ab99b39771b5084a231a2db8d3dc969f395f9
  GoVersion: go1.13.8
  OsArch: linux/amd64
  RemoteAPIVersion: 1
  Version: 2.0.0-dev
```

podman version
```
Version:            2.0.0-dev
RemoteAPI Version:  1
Go Version:         go1.13.8
Git Commit:         343ab99b39771b5084a231a2db8d3dc969f395f9
Built:              Sat May 16 09:49:14 2020
OS/Arch:            linux/amd64
```

rhatdan:
Yes I guess we should translate the date.

dougsland:
@edsantiago looks like your requested is fixed. Could you please double check? 
If everything is working for you, please consider closing this ticket.

```
$ git clone git@github.com:containers/libpod.git && cd libpod
$ make && sudo make install

$ podman info
<snip>
version:
  APIVersion: 1
  Built: 1590582353
  BuiltTime: Wed May 27 08:25:53 2020
  GitCommit: 119e13d4bcc3e8da903da33b291f2795b0e76e71
  GoVersion: go1.14.2
  OsArch: linux/amd64
  Version: 2.0.0-dev
```

edsantiago:
LGTM, thank you!

rhatdan:
@afbjorklund Please open an issue on podman-remote info to get the other information.

PR #6275, opened 2020-05-19
Display human build date in podman info

rhatdan:
Currently we are displaying the Seconds since EPOCH
this will change to displaying date, similar to `podman version`

### DIFF
--- a/docs/source/markdown/podman-info.1.md
+++ b/docs/source/markdown/podman-info.1.md
@@ -30,128 +30,212 @@ Run podman info with plain text response:
 ```
 $ podman info
 host:
-  BuildahVersion: 1.4-dev
-  Conmon:
-    package: Unknown
-    path: /usr/libexec/podman/conmon
-    version: 'conmon version 1.12.0-dev, commit: d724f3d54ad2d95b6de741085d4990190ebfd7ff'
-  Distribution:
-    distribution: fedora
-    version: "28"
-  MemFree: 1271083008
-  MemTotal: 33074233344
-  OCIRuntime:
-    package: runc-1.0.0-51.dev.gitfdd8055.fc28.x86_64
-    path: /usr/bin/runc
-    version: 'runc version spec: 1.0.0'
-  SwapFree: 34309664768
-  SwapTotal: 34359734272
   arch: amd64
+  buildahVersion: 1.15.0
+  cgroupVersion: v1
+  conmon:
+    package: conmon-2.0.16-2.fc32.x86_64
+    path: /usr/bin/conmon
+    version: 'conmon version 2.0.16, commit: 1044176f7dd177c100779d1c63931d6022e419bd'
   cpus: 8
+  distribution:
+    distribution: fedora
+    version: "32"
+  eventLogger: file
   hostname: localhost.localdomain
-  kernel: 4.18.7-200.fc28.x86_64
+  idMappings:
+    gidmap:
+    - container_id: 0
+      host_id: 3267
+      size: 1
+    - container_id: 1
+      host_id: 100000
+      size: 65536
+    uidmap:
+    - container_id: 0
+      host_id: 3267
+      size: 1
+    - container_id: 1
+      host_id: 100000
+      size: 65536
+  kernel: 5.6.11-300.fc32.x86_64
+  linkmode: dynamic
+  memFree: 1401929728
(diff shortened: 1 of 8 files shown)


mheon:
More Quay issues, looks like. Restarted tests.

libpod/define/version.go
```diff
@@ -17,6 +18,16 @@ var (
 	buildInfo string
 )
 
+// humanVersion is an human output struct for varlink
+type humanVersion struct {
```
mheon, line 22: Why is this private? Why not `HumanVersion`, since this is used outside of Libpod?

baude:
does this get exposed on the v2 API? if so, does it impact cockpit?

TomSweeneyRedHat:
The version manpage has an example or two on it.  It would probably be good to verify and if necessary update those based on these changes.
LGTM other than prior comments.

libpod/define/version.go
rhatdan: The goal was to not expose it outside of libpod.  Just to give a prettier view to users in the podman info call.

mheon: I think it does make it to the remote API, though.

rhatdan:
Currently we are returning the EPOCH time to cockpit.  Which it can switch the display to human form.  The problem with Info is there is no good (easy) way to process this on the client side.

libpod/define/version.go
mheon: Looks like it does - I see calls to it in the APIv2 handlers

rhatdan: Another option would be to just return both BuiltTime and Built so we could have human form and epoch seconds.

rhatdan:
@TomSweeneyRedHat This is only for `podman info` not `podman version`.

rhatdan:
@baude @mheon @TomSweeneyRedHat I took another pass at this and only added the BuildTime object to Version, so it should not effect any consumers.

rhatdan:
Man page is just using current output of the podman info command.

mheon:
LGTM

rhatdan:
@mheon I added another patch to fix a bug I found while improving the docs

docs/source/markdown/podman-version.1.md
```diff
@@ -25,17 +25,18 @@ Change output format to "json" or a Go template.
...
+Git Commit:         4520664f63c3a7f9a80227715359e20069d95542
+Built:              Tue May 19 10:48:59 2020
+OS/Arch:            linux/amd64
 ```
 
 Filtering out only the version:
 ```
 $ podman version --format '{{.Client.Version}}'
-1.6.3
+2.0.0
```
TomSweeneyRedHat, line 39: See, I told you this needed updating!  ;^)

docs/source/markdown/podman-info.1.md
```diff
@@ -30,128 +30,212 @@ Run podman info with plain text response:
...
-  MemFree: 1271083008
-  MemTotal: 33074233344
-  OCIRuntime:
-    package: runc-1.0.0-51.dev.gitfdd8055.fc28.x86_64
-    path: /usr/bin/runc
-    version: 'runc version spec: 1.0.0'
-  SwapFree: 34309664768
-  SwapTotal: 34359734272
   arch: amd64
+  buildahVersion: 1.15.0
```
TomSweeneyRedHat, line 34: is this version right?

TomSweeneyRedHat:
one question, otherwise LGTM

docs/source/markdown/podman-info.1.md
rhatdan: I removed the -dev flag, to make it look better, since we will probably not update this when we deliver podman 2.0.

QiWang19:
LGTM. Will this PR and #5834 close #5736 ?

docs/source/markdown/podman-info.1.md
TomSweeneyRedHat: Ah, thanks.  I was a bit concerned we were already spitting out v1.15 somehow.

rhatdan:
Yes this will 
fixes: #5736

QiWang19:
LGTM

TomSweeneyRedHat:
LGTM 
@mheon I'll let you press the final button on this one

PR #6279, opened 2020-05-19
use internal copy rather than tar for copy fix #6003

### LINKED ISSUE #6003: podman cannot mount volumes with external symlinks
/kind bug

**Description**

Podman volumes with external symlinks does not seem to be mountable

1.  Create volume with symlink which references files outside of the volume.  For example symbolic link to ../foo

2. Try to mount the volume. 

3. Volume does not mount

Volume fails to mount

Volume should mount, and when I enter the image, I should see the symbolic link point to the file within the context of the mounted volume.  This is the behavior in docker. Also these sorts of symbolic links are very common in directory /etc which points to /usr/share or to other directories outside the volume

-----------------
Error: unable to start container 34e9c7636cd474d63741b93ecbd0dabfcab44aad263cedac4b4aa65ef93c5460: error copying content from container 34e9c7636cd474d63741b93ecbd0dabfcab44aad263cedac4b4aa65ef93c5460 into volume demo_etc: invalid symlink "/home/joe/.local/share/containers/storage/volumes/demo_etc/_data/httpd/logs" -> "../../var/log/httpd"

---------------

[joe@big-apple bench]$ podman version
Version:            2.0.0-dev
RemoteAPI Version:  1
Go Version:         go1.14.2
```

[joe@big-apple SPECS]$ 
  configFile: /home/joe/.config/containers/storage.conf
  containerStore:
    number: 9
    paused: 0
    running: 5
    stopped: 4
  graphDriverName: vfs
  graphOptions: {}
  graphRoot: /home/joe/.local/share/containers/storage
  graphStatus: {}
  imageStore:
    number: 36
  volumePath: /home/joe/.local/share/containers/storage/volumes
version:
  Built: 0
  GitCommit: ""
  GoVersion: go1.14.2
  OsArch: linux/amd64
  RemoteAPIVersion: 1
  Version: 2.0.0-dev
```

```
podman-2.0.0-0.1.dev.git8857ba2.mga8
```

joequant:
This patch uses internal copy to copy containers rather than
tar.  The problem with tar is that it does not copy symlinks pointing
outside of the volume.

This is a very tricky patch to get right because you have to get the 
ownerships and permissions right as well as handle symlinks 
correctly.  Off the shelf copy trees don't work because they dont
have the right combination of options to do the right thing.

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1411,6 +1411,129 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
 	return mountPoint, nil
 }
 
+// Copy tree with owner and perms
+
+type CopyError struct {
+	Src string
+	Dst string
+}
+
+func (e CopyError) Error() string {
+	return fmt.Sprintf("`%s -> %s` copyerror", e.Src, e.Dst)
+}
+
+func samefile(src string, dst string) bool {
+	srcInfo, _ := os.Stat(src)
+	dstInfo, _ := os.Stat(dst)
+	return os.SameFile(srcInfo, dstInfo)
+}
+
+func specialfile(fi os.FileInfo) bool {
+	return (fi.Mode() & os.ModeNamedPipe) == os.ModeNamedPipe
+}
+
+func IsSymlink(fi os.FileInfo) bool {
+	return (fi.Mode() & os.ModeSymlink) == os.ModeSymlink
+}
+
+// Copy file with owner and perms
+func (c *Container) copyWithOwnerAndPerms(src, dst string) error {
+	if samefile(src, dst) {
+		return &CopyError{src, dst}
+	}
+
+	srcStat, err := os.Lstat(src)
+	if err != nil {
+		return err
+	} else if specialfile(srcStat) {
+		return &CopyError{src, dst}
+	}
+
+	if dstInfo, err := os.Stat(dst); err == nil && dstInfo.Mode().IsDir() {
+		dst = filepath.Join(dst, filepath.Base(src))
+	} else if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	if IsSymlink(srcStat) {
+		if srcLink, err := os.Readlink(src); err != nil {
+			return err
+		} else if err := os.Symlink(srcLink, dst); err != nil {
+			return err
+		}
+	}
+
+	// Do the actual copy
+	fsrc, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer fsrc.Close()
+
+	fdst, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer fdst.Close()
+
+	if size, err := io.Copy(fdst, fsrc); err != nil {
+		return err
+	} else if size != srcStat.Size() {
+		return fmt.Errorf("%s: %d/%d copied", src, size, srcStat.Size())
+	}
+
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@nalind PTAL

libpod/container_internal.go
```diff
@@ -1411,6 +1411,129 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
...
+		if err := os.MkdirAll(dst, srcFileInfo.Mode()); err != nil {
+			return err
+		}
+	} else if !dstFileInfo.IsDir() {
+		return &CopyError{src, dst}
+	}
+
+	for _, entry := range entries {
+		srcPath := filepath.Join(src, entry.Name())
+		dstPath := filepath.Join(dst, entry.Name())
```
TomSweeneyRedHat, line 1515: I think you want to user securejoin for these

libpod/container_internal.go
```diff
@@ -1411,6 +1411,129 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
...
+	} else if size != srcStat.Size() {
+		return fmt.Errorf("%s: %d/%d copied", src, size, srcStat.Size())
+	}
+
+	if err = c.copyOwnerAndPerms(src, dst); err != nil {
+		return err
+	}
+
+	return nil
+}
```
giuseppe, line 1490: could we use CopyFile from `github.com/mrunalp/fileutils` here?

libpod/container_internal.go
```diff
@@ -1411,6 +1411,129 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
...
+	srcInfo, _ := os.Stat(src)
+	dstInfo, _ := os.Stat(dst)
+	return os.SameFile(srcInfo, dstInfo)
+}
+
+func specialfile(fi os.FileInfo) bool {
+	return (fi.Mode() & os.ModeNamedPipe) == os.ModeNamedPipe
+}
+
+func IsSymlink(fi os.FileInfo) bool {
```
nalind, line 1435: This function looks like it doesn't need to be exported.

libpod/container_internal.go
nalind: The tar-based logic would have attempted to preserve timestamps and xattrs on items it copied, and made an effort to preserve hard links among multiple items being copied.  If we don't want to lose those features (whether podman wants to keep those things could well be different than the previous implementation), they'll need to be added back here.

mheon: Xattrs are likely important. I don't know if timestamps are; need to check.

rhatdan: Hardlinks also, to save on size.

rhatdan:
@joequant Still working on this, needs a rebase and some comments addressed.

vrothberg:
Friendly ping.

rhatdan:
@mheon If you are going to work on rewriting copy, you might want to take this over.

rhatdan:
The Copy rewrite is continuing on Buildah and is progressing,  I think we should close this and work on the new Copy once it merges into Buildah.

joequant:
Thanks.  Also any ETA for when the changes will get merged in?  I'm using docker and would like to migrate over to podman.

joequant:
Can you link the issue that tracks this on buildah?  Thanks.

rhatdan:
https://github.com/containers/buildah/issues/2476

PR #6280, opened 2020-05-19
Turn off 'noexec' option by default for named volumes

mheon:
We previously enforced this for security reasons, but as Dan has explained on several occasions, it's not very valuable there (it's trivially easy to bypass) and it does seriously annoy folks trying to use named volumes. Flip the default from 'on' to 'off'.

### DIFF
--- a/test/system/160-volumes.bats
+++ b/test/system/160-volumes.bats
@@ -115,7 +115,8 @@ echo "got here -$rand-"
 EOF
     chmod 755 $mountpoint/myscript
 
-    # By default, volumes are mounted noexec. This should fail.
+    # By default, volumes are mounted exec, but we have manually added the
+    # noexec option. This should fail.
     # ARGH. Unfortunately, runc (used for cgroups v1) produces a different error
     local expect_rc=126
     local expect_msg='.* OCI runtime permission denied.*'
@@ -125,12 +126,12 @@ EOF
         expect_msg='.* exec user process caused.*permission denied'
     fi
 
-    run_podman ${expect_rc} run --rm --volume $myvolume:/vol:z $IMAGE /vol/myscript
+    run_podman ${expect_rc} run --rm --volume $myvolume:/vol:noexec,z $IMAGE /vol/myscript
     is "$output" "$expect_msg" "run on volume, noexec"
 
-    # With exec, it should pass
-    run_podman run --rm -v $myvolume:/vol:z,exec $IMAGE /vol/myscript
-    is "$output" "got here -$rand-" "script in volume is runnable with exec"
+    # With the default, it should pass
+    run_podman run --rm -v $myvolume:/vol:z $IMAGE /vol/myscript
+    is "$output" "got here -$rand-" "script in volume is runnable with default (exec)"
 
     # Clean up
     run_podman volume rm $myvolume
--- a/pkg/util/mountOpts_linux.go
+++ b/pkg/util/mountOpts_linux.go
@@ -7,7 +7,7 @@ import (
 )
 
 func getDefaultMountOptions(path string) (defaultMountOptions, error) {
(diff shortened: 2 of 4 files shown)


rhatdan:
LGTM

rhatdan:
@BenTheElder PTAL

rhatdan:
@baude @giuseppe @vrothberg @QiWang19 @jwhonce @TomSweeneyRedHat PTAL

mheon:
Apparently we have a test verifying that `noexec` is in there - let me fix that

rhatdan:
Looks like the exec/noexec test is still failing.

TomSweeneyRedHat:
LGTM assuming happy tests

BenTheElder:
ACK thanks, small note: In our case we were trying to use anonymous volumes (they're just scratch space, bound to the container lifecycle) and swap docker/podman. 

Switching to creating volumes and specifying these explicitly in the podman-specific logic and improving detection of podman-docker vs docker is working well for us and should also let us bypass https://github.com/containers/libpod/issues/4276 on older podman versions.

mheon:
Hit the system tests as well; @edsantiago mind taking a look to make sure this looks OK?

test/system/160-volumes.bats
```diff
@@ -125,13 +126,17 @@ EOF
         expect_msg='.* exec user process caused.*permission denied'
     fi
 
-    run_podman ${expect_rc} run --rm --volume $myvolume:/vol:z $IMAGE /vol/myscript
+    run_podman ${expect_rc} run --rm --volume $myvolume:/vol:noexecz $IMAGE /vol/myscript
```
edsantiago, line 129: "noexecz"? Should there be a comma or something?

mheon: This is what happens when I push before the tests finish running. Oops.

test/system/160-volumes.bats
```diff
@@ -125,13 +126,17 @@ EOF
         expect_msg='.* exec user process caused.*permission denied'
     fi
 
-    run_podman ${expect_rc} run --rm --volume $myvolume:/vol:z $IMAGE /vol/myscript
+    run_podman ${expect_rc} run --rm --volume $myvolume:/vol:noexec,z $IMAGE /vol/myscript
     is "$output" "$expect_msg" "run on volume, noexec"
 
     # With exec, it should pass
     run_podman run --rm -v $myvolume:/vol:z,exec $IMAGE /vol/myscript
     is "$output" "got here -$rand-" "script in volume is runnable with exec"
```
edsantiago, line 135: remove these

edsantiago: Sorry for brusqueness, I just wasted too much time typing, then having my comments lost because of outdated diffs, then typing again, then losing them AGAIN DAMN YOU GITHUB. Didn't want to do that again.

Anyhow, my point was, system tests aren't intended to cover every possible situation - they're just a last-minute check for basic functionality. Testing the default is a good enough check for exec; testing noexec, likewise. We don't IMHO need a full test just for parsing 'exec'.

edsantiago:
"Failed to stop: readHandshakeRecord" in Cirrus ?!

test/system/160-volumes.bats
mheon: Fixed, thanks

kallisti5:
oh wow.. this one was a obnoxious to track down and not super obvious. All of our CI started breaking.

Pretty much compiling anything within a named volume requires exec (autoconf does a bunch of gcc compiler checks which execute test binaries)

rhatdan:
@mheon I think this should be back ported to podman 1.9.3

mheon:
Sure, will do.

Removing hold, things are green
/hold cancel

PR #6291, opened 2020-05-20
oci conmon: tell conmon to log container name

haircommander:
specifying `-n=ctr-name` tells conmon to log CONTAINER_NAME=name if the log driver is journald

add this, and a test!

also, refactor the args slice creation to not append() unnecessarily.

technically, this needs https://github.com/containers/conmon/pull/154, though the test is skipped anyway

### DIFF
--- a/test/e2e/logs_test.go
+++ b/test/e2e/logs_test.go
@@ -173,6 +173,24 @@ var _ = Describe("Podman logs", func() {
 		Expect(string(out)).To(ContainSubstring("alpine"))
 	})
 
+	It("podman journald logs for container name", func() {
+		Skip("need to verify images have correct packages for journald")
+		containerName := "inside-journal"
+		logc := podmanTest.Podman([]string{"run", "--log-driver", "journald", "-d", "--name", containerName, ALPINE, "sh", "-c", "echo podman; sleep 0.1; echo podman; sleep 0.1; echo podman"})
+		logc.WaitWithDefaultTimeout()
+		Expect(logc.ExitCode()).To(Equal(0))
+		cid := logc.OutputToString()
+
+		wait := podmanTest.Podman([]string{"wait", "-l"})
+		wait.WaitWithDefaultTimeout()
+		Expect(wait.ExitCode()).To(BeZero())
+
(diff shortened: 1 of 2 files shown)


test/e2e/logs_test.go
```diff
@@ -173,6 +173,24 @@ var _ = Describe("Podman logs", func() {
 		Expect(string(out)).To(ContainSubstring("alpine"))
 	})
 
+	It("podman journald logs for container name", func() {
+		Skip("need to verify images have correct packages for journald")
```
rhatdan, line 177: You test is always skipped?

rhatdan:
LGTM
other then the issue with skipping the test.

test/e2e/logs_test.go
haircommander: all of the logs tests are always skipped

haircommander: they're actually skipped in two different spots :smile:

haircommander:
all logs tests are skipped [here](https://github.com/containers/libpod/blob/a6700147a820da5662ebe114a89c644df4fe1845/test/e2e/logs_test.go#L22)
all journald tests are skipped [here](https://github.com/containers/libpod/blob/a6700147a820da5662ebe114a89c644df4fe1845/test/e2e/logs_test.go#L160)

I have a feeling fully enabling the test will be a bit more challenging than dropping these. do you two want me to address those here @rhatdan @vrothberg ?

vrothberg:
> all logs tests are skipped here

Only for remote though.

> I have a feeling fully enabling the test will be a bit more challenging than dropping these. do you two want me to address those here @rhatdan @vrothberg ?

If you have some cycles for that, it would be great. Adding disabled tests is not ideal.

haircommander:
I have about enough cycles to attempt to enable them, but I can't promise carrying through fixing the tests if they fail from v2/not having journalctl on the images

haircommander:
(also without https://github.com/containers/conmon/pull/154 the test will fail)

haircommander:
update: ci images do not have journald. I would like to enable the tests as a follow up

PR #6300, opened 2020-05-20
govern remote attach and start

baude:
fixes a race where container would start before attach could occur resulting in an error.


mheon:
LGTM

PR #6302, opened 2020-05-20
Fix create_test for remote integration

baude:
fixes tests to pass for remote integration.  the two remaining tests that are skipped are due to lack of logs command.


rhatdan:
LGTM
one issue with this is that we are no longer testing the -l flag, but I don't think that is a big deal.

TomSweeneyRedHat:
LGTM

mheon:
Can we ensure that we do not completely drop testing of `--latest` as part of this? Adding separate, gated tests that only run local seems like a good idea.

PR #6310, opened 2020-05-20
Update Derivative API tutorial to reflect the HTTP API

mheon:
Once the HTTP API becomes stable, I think we want to drive people towards using it, as opposed to directly calling our CLI (when possible).


rhatdan:
LGTM
/approve

PR #6311, opened 2020-05-20
Fix build on OS X

mheon:
We disabled the OS X and Windows cross-building tests. This, predictably, led us to regress a bit in our ability to build for both of these.

This fixes the build on OS X and fixes one obvious Windows bug. Unfortunately, we're dragging in all of `pkg/spec` somewhere on Windows, and things are blowing up spectacularly because of it (plus a few uses of the `syscall` package in the bindings).

I've giving up for the day. This fixes OS X, but does not fully enable the cross-build CI (need Windows fixes for that).


rhatdan:
LGTM

PR #6312, opened 2020-05-21
Fix remote handling of podman images calls

rhatdan:
Enable three more tests
Fix handling of image filters

### DIFF
--- a/test/e2e/images_test.go
+++ b/test/e2e/images_test.go
@@ -131,7 +131,6 @@ var _ = Describe("Podman images", func() {
 	})
 
 	It("podman images filter by image name", func() {
-		Skip(v2remotefail)
 		podmanTest.RestoreAllArtifacts()
 		session := podmanTest.PodmanNoCache([]string{"images", "-q", ALPINE})
 		session.WaitWithDefaultTimeout()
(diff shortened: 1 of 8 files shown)


rhatdan:
@baude @mheon @jwhonce @giuseppe @vrothberg @QiWang19 @TomSweeneyRedHat PTAL

test/e2e/images_test.go
```diff
@@ -152,9 +152,7 @@ var _ = Describe("Podman images", func() {
 	})
 
 	It("podman images filter reference", func() {
-		if podmanTest.RemoteTest {
-			Skip("Does not work on remote client")
-		}
+		SkipIfRemote()
```
QiWang19, line 155: Does this enable the test?

baude: it seems like he is replacing a piece of code with a convenience function.

baude:
lgtm

test/e2e/images_test.go
rhatdan: Yes just cleaning up.
BTW These can get enabled for remote, as soon as podman-remote build works.

test/e2e/images_test.go
```diff
@@ -180,9 +178,7 @@ var _ = Describe("Podman images", func() {
 	})
 
 	It("podman images filter before image", func() {
-		if podmanTest.RemoteTest {
-			Skip("Does not work on remote client")
-		}
+		SkipIfRemote()
```
mheon, line 181: Why are all of these SkipIfRemote? A lot of them seem like they ought to work remote

rhatdan: podman build does not work, so these will not work until that is fixed.  These were blocked in podman 1.9 as well.

mheon: Ahh, ack.

PR #6318, opened 2020-05-21
Enable system prune test remote client

sujil02:
Removed the use of lastest flags as not supported in remote client
and latest flag alredy teseted in indivisual pod start stop test.

### DIFF
--- a/test/e2e/prune_test.go
+++ b/test/e2e/prune_test.go
@@ -161,7 +161,6 @@ var _ = Describe("Podman prune", func() {
 	})
 
 	It("podman system prune pods", func() {
-		Skip(v2remotefail)
 		session := podmanTest.Podman([]string{"pod", "create"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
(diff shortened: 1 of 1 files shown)


test/e2e/prune_test.go
```diff
@@ -320,9 +322,6 @@ var _ = Describe("Podman prune", func() {
 		create.WaitWithDefaultTimeout()
 		Expect(create.ExitCode()).To(Equal(0))
 
-		// Adding images should be pruned
-		podmanTest.BuildImage(pruneImage, "alpine_bash:latest", "true")
```
rhatdan, line 324: Why are you removing this?

sujil02: Build image with layers is not implemented by a remote client. And for this use case it simply added a bunch of more images using build image and checks if they are pruned with system prune command. However the addition of BB image here https://github.com/containers/libpod/pull/6318/files#diff-9e3f37b4287d658db5658028b861b77bL319 and validating it pruned here https://github.com/containers/libpod/pull/6318/files#diff-9e3f37b4287d658db5658028b861b77bL351
addresses the use case hence the test need not be skipped for that matter and is still accounts as a valid test.

rhatdan:
LGTM
Although I would like to get podman-remote build to work.

PR #6319, opened 2020-05-21
Get proper exit code when running or starting a container.

rhatdan:
When we finish running a container, we need to call wait in order
to get the exit code from the container.

### DIFF
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -360,10 +360,19 @@ func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []stri
 		report := entities.ContainerStartReport{Id: name}
 		if options.Attach {
 			report.Err = startAndAttach(ic, name, &options.DetachKeys, options.Stdin, options.Stdout, options.Stderr)
+			if report.Err == nil {
+				exitCode, err := containers.Wait(ic.ClientCxt, name, nil)
+				if err == nil {
+					report.ExitCode = int(exitCode)
+				}
+			} else {
+				report.ExitCode = define.ExitCode(report.Err)
+			}
 			reports = append(reports, &report)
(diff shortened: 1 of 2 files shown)


baude:
lgtm

mheon:
For reference, I think exit codes with `podman run --rm` will still be broken for remote, and I'm not sure exactly how to fix them.

rhatdan:
@mheon This seems to work.
```
sh-5.0# podman rm foo
Error: no container with name or ID foo found: no such container
sh-5.0# echo $?
1
sh-5.0# ./bin/podman-remote rm foo
Error: unable to find container "foo": no such container
sh-5.0# echo $?
1
sh-5.0# ./bin/podman-remote rm foo
Error: unable to find container "foo": no such container
sh-5.0# echo $?
1
sh-5.0# ./bin/podman-remote rmi foo
Error: 1 error occurred:
	* unable to find a name and tag match for foo in repotags: no such image
sh-5.0# echo $?
1
```

rhatdan:
```
sh-5.0# podman rm d05d4b492e7d81c54f7e395e1d293c7dc6045213faafcf34a2be2a3172af503d
Error: cannot remove container d05d4b492e7d81c54f7e395e1d293c7dc6045213faafcf34a2be2a3172af503d as it is running - running or paused containers cannot be removed without force: container state improper
sh-5.0# echo $?
2
sh-5.0# ./bin/podman-remote rm d05d4b492e7d81c54f7e395e1d293c7dc6045213faafcf34a2be2a3172af503d
Error: cannot remove container d05d4b492e7d81c54f7e395e1d293c7dc6045213faafcf34a2be2a3172af503d as it is running - running or paused containers cannot be removed without force: container state improper
sh-5.0# echo $?
2
```

mheon:
@rhatdan I expect `podman rm` works properly, but `podman run --rm` likely does not.

PR #6321, opened 2020-05-21
Allow to change the generated systemd unit name prefix

Luap99:
I often generate my systemd units with `podman gernerate systemd` but I dont like the default unit prefixes _"container-"_/_"pod-"_, so I have to change them manually in a extra step.
This PR allows to change the created systemd unit names with a custom prefix and a custom separator with the following flags:
- --container-prefix
- --pod-prefix
- --separator
The current behavior does not change!

If a user dont want any prefix he could do now `podman generate systemd -f -n --container-prefix "" --pod-prefix "" --separator "" POD/CONTAINER`
If you want to change the separator you can do `podman generate systemd -f -n --separator "_" POD/CONTAINER`


mheon:
@vrothberg PTAL

rhatdan:
Testing does not like your commit messages > 80 chars on first line.
Also can you squash your commits.

Luap99:
OK, thank you. I am new to this stuff.
I now squashed them and also added tests to the commit.

rhatdan:
LGTM

PR #6325, opened 2020-05-21
Turn off 'noexec' option by default for named volumes

mheon:
We previously enforced this for security reasons, but as Dan has explained on several occasions, it's not very valuable there (it's trivially easy to bypass) and it does seriously annoy folks trying to use named volumes. Flip the default from 'on' to 'off'.

This is a backport from the master branch to v1.9 branch.

### DIFF
--- a/pkg/util/mountOpts_linux.go
+++ b/pkg/util/mountOpts_linux.go
@@ -7,7 +7,7 @@ import (
 )
 
 func getDefaultMountOptions(path string) (defaultMountOptions, error) {
-	opts := defaultMountOptions{true, true, true}
+	opts := defaultMountOptions{false, true, true}
 	if path == "" {
 		return opts, nil
 	}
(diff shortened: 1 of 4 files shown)


mheon:
Ed's 160-volumes test has leaked in, but I don't think adding extra testing here is a bad thing.

mheon:
@rhatdan @vrothberg @giuseppe @baude @jwhonce @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM assuming happy tests

edsantiago:
LGTM. Thank you for bringing in the (new) tests rather than skipping them.

edsantiago:
Grumble - maybe the tests weren't such a good idea. One of the failures is almost certainly due to the CapItAliZaTion changes. The other, a failing `touch` with `userns=keep-id`, I _think_ I've seen before and is a bug in 1.9 that has since been fixed. 

I'm OK if you want to just remove those failing tests.

PR #6327, opened 2020-05-21
v2 podman-remote build

baude:
this is a very basic implementation of build.  some of the more advanced options need to be included still as well.  i think the endpoints for compat and libpod will have to split given buildahs more advanced set of options.  that should probably be done by someone more experienced with build internals.

### DIFF
--- a/pkg/api/server/register_images.go
+++ b/pkg/api/server/register_images.go
@@ -1188,5 +1188,214 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//     $ref: "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/libpod/images/{name}/changes"), s.APIHandler(compat.Changes)).Methods(http.MethodGet)
 
+	// swagger:operation POST /libpod/build libpod libpodBuildImage
+	// ---
+	// tags:
+	//  - images
+	// summary: Create image
+	// description: Build an image from the given Dockerfile(s)
+	// parameters:
+	//  - in: query
+	//    name: dockerfile
+	//    type: string
+	//    default: Dockerfile
+	//    description: |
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
@rhatdan should we switch from Dockerfile -> Containerfile through out?

pkg/api/server/register_images.go
```diff
@@ -1188,5 +1188,214 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
...
+	//           description: output from build process
+	//           example: |
+	//             (build details...)
+	//             Successfully built 8ba084515c724cbf90d447a63600c0a6
+	//   400:
+	//     $ref: "#/responses/BadParamError"
+	//   500:
+	//     $ref: "#/responses/InternalError"
+	r.Handle(VersionedPath("/libpod/build"), s.APIHandler(compat.BuildImage)).Methods(http.MethodPost)
 	return nil
```
TomSweeneyRedHat, line 1400: Probably ought to have a todo with the options not here at least.  I'm not seeing annotations or authfile for instance.

baude: @jwhonce implemented the original for compatibility only.  as I said in my  commit message, it likely needs to be split but again, this needs more expertise that I have time.  I'm hoping someone will take this and run with it merged or not.

TomSweeneyRedHat: Not seeing layers, tls-verify, and volume options probably ought to add those at least.  That should get the majority of cores stuff and we could add others later.

rhatdan: I am fine with getting a bare bones version in, and then adding options as we get time.  For now, we want to get the tests passing.

jwhonce:
LGTM for now

PR #6345, opened 2020-05-22
remote manifest test

QiWang19:
Enable remove manifest tests. Skip --purge test because remote does not support it.


rhatdan:
LGTM

PR #6353, opened 2020-05-22
enable building without `varlink` tag

lsm5:
default build without `varlink` tag
    
Issue gh#6286 was already fixed in a prior commit but the Makefile still
ran some varlink steps by default.
    
This commit makes any varlink build steps dependent on the varlink
build tag and also makes the contrib rpm spec file independent of
varlink.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ ETCDIR ?= /etc
 TMPFILESDIR ?= ${PREFIX}/lib/tmpfiles.d
 SYSTEMDDIR ?= ${PREFIX}/lib/systemd/system
 USERSYSTEMDDIR ?= ${PREFIX}/lib/systemd/user
-REMOTETAGS ?= !ABISupport remote exclude_graphdriver_btrfs btrfs_noversion  exclude_graphdriver_devicemapper containers_image_openpgp
+REMOTETAGS ?= !ABISupport remote exclude_graphdriver_btrfs btrfs_noversion exclude_graphdriver_devicemapper containers_image_openpgp
 BUILDTAGS ?= \
 	$(shell hack/apparmor_tag.sh) \
 	$(shell hack/btrfs_installed_tag.sh) \
@@ -30,8 +30,7 @@ BUILDTAGS ?= \
 	$(shell hack/selinux_tag.sh) \
 	$(shell hack/systemd_tag.sh) \
 	exclude_graphdriver_devicemapper \
-	seccomp \
-	varlink
+	seccomp
 PYTHON ?= $(shell command -v python3 python|head -n1)
 PKG_MANAGER ?= $(shell command -v dnf yum|head -n1)
 # ~/.local/bin is not in PATH on all systems
@@ -145,6 +144,9 @@ help:
 ifeq ("$(wildcard $(GOPKGDIR))","")
 	mkdir -p "$(GOPKGBASEDIR)"
 	ln -sfn "$(CURDIR)" "$(GOPKGDIR)"
+endif
+
+ifneq (,$(findstring varlink,$(BUILDTAGS)))
 	ln -sfn "$(CURDIR)/vendor/github.com/varlink" "$(FIRST_GOPATH)/src/github.com/varlink"
 endif
 	touch $@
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -634,11 +644,20 @@ endef
 		$(call go-get,github.com/cpuguy83/go-md2man); \
 	fi
 
+# $BUILD_TAGS variable is used in hack/golangci-lint.sh
 .PHONY: varlink_generate
+ifneq (or $(findstring varlink,$(BUILDTAGS)),$(findstring varlink,$(BUILD_TAGS)))
```
edsantiago, line 649: I'm super rusty on Makefile magic, but this doesn't look right to me. Are you looking for `varlink` in either of those two variables? If so would this work?
```makefile
ifneq (,$(findstring varlink,$(BUILDTAGS) $(BUILD_TAGS)))
```

lsm5: yes, I'm looking for varlink in those two. I'll give that a try, thanks!

lsm5: there seems to be some integration test issues that I'm looking into as well.

lsm5: `make lint` fails with:
```
# $BUILD_TAGS variable is used in hack/golangci-lint.sh
 .PHONY: varlink_generate
-ifneq (or $(findstring varlink,$(BUILDTAGS)),$(findstring varlink,$(BUILD_TAGS)))
+ifneq (,$(findstring varlink,$(BUILD_TAGS) $(BUILDTAGS)))
 varlink_generate: .gopathok pkg/varlink/iopodman.go ## Generate varlink
 else
 varlink_generate:
```

edsantiago: Probably some magic syntax detail I'm missing.

Well, does the `or` format work? If you run `make` with varlink in either variable, does it trigger the desired varlink_generate?

lsm5: The `or` format works well for `make lint`. I don't need to specify `varlink` buildtag for tunnel and abi as it will automatically use it from the golangci-lint.sh file. Both `make` and `make BUILDTAGS='varlink' work correctly as well.

edsantiago: OK - sorry to have muddled things. LGTM, thanks.

lsm5: btw, what you suggested *I think* should work as well, which is what seems to be suggested in the [official make doc](https://www.gnu.org/software/make/manual/make.html#Testing-Flags), but for whatever reason it doesn't seem to work in this case. Maybe more eyes would help..

lsm5:
Btw, I guess in the default case, it will now skip varlink build and tests. Let me know if CI needs to continue testing for the varlink case as well.

edsantiago:
Hmmm. CI is not happy, and it doesn't look (at first glance) like a flake:
```
[+0179s] # github.com/containers/libpod/test/endpoint [github.com/containers/libpod/test/endpoint.test]
[+0179s] test/endpoint/config.go:10: undefined: ALPINE
[+0179s] test/endpoint/setup.go:19: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:102: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:120: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:129: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:136: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:144: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:176: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:185: undefined: EndpointTestIntegration
[+0179s] test/endpoint/setup.go:195: undefined: EndpointTestIntegration
[+0179s] test/endpoint/config.go:10: too many errors
```

lsm5:
> Hmmm. CI is not happy, and it doesn't look (at first glance) like a flake:
> 
> ```
> [+0179s] # github.com/containers/libpod/test/endpoint [github.com/containers/libpod/test/endpoint.test]
> [+0179s] test/endpoint/config.go:10: undefined: ALPINE
> [+0179s] test/endpoint/setup.go:19: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:102: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:120: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:129: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:136: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:144: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:176: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:185: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/setup.go:195: undefined: EndpointTestIntegration
> [+0179s] test/endpoint/config.go:10: too many errors
> ```

err, yes that's a consequence of me wrapping test/endpoint/endpoint.go inside varlink buildtag.

Could someone please let me know if `test/endpoint` is a varlink-specific thing? Or if I should split out the varlink specific stuff in separate files?

edsantiago:
>if `test/endpoint` is a varlink-specific thing?

There is no way I can answer that but I'm going to jump in anyway: a quick `ack -i varlink test/endpoint/` produces so many hits that I have to assume it is.

lsm5:
I added conditional to skip endpoint tests if no varlink buildtag found.

edsantiago:
LGTM

rhatdan:
It would be nice to split out the varlink code into a separate Makefile.Varlink.  And then optionally include it if necessary.  That might cleanup the Makefile.

/lgtm

PR #6362, opened 2020-05-22
[CI:DOCS] Fix readthedocs link

TomSweeneyRedHat:
Touch up the link to the docs on readthedocs.  Using the fully
specified link like this will cause a CORS issue in many browsers.

Plus we're working on a Spanish variant of the site, so we probably
should point to the English variant.


mheon:
LGTM

PR #6363, opened 2020-05-22
V2 Fix interface nil checks

jwhonce:
Bugs fixes:

* Accept connections from services that don't provide version header
* Fix interface parameter nil checks. See https://golang.org/doc/faq#nil_error

### DIFF
--- a/pkg/bindings/containers/logs.go
+++ b/pkg/bindings/containers/logs.go
@@ -50,7 +50,6 @@ func Logs(ctx context.Context, nameOrID string, opts LogOptions, stdoutChan, std
 	if err != nil {
 		return err
 	}
-	defer response.Body.Close()
 
 	buffer := make([]byte, 1024)
 	for {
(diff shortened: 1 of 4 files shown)


baude:
your signing is off ... fix and we can review and merge

baude:
lgtm

pkg/bindings/containers/logs.go
```diff
@@ -50,7 +50,6 @@ func Logs(ctx context.Context, nameOrID string, opts LogOptions, stdoutChan, std
 	if err != nil {
 		return err
 	}
-	defer response.Body.Close()
```
mheon, line 53: Why was this removed?

mheon:
One question otherwise LGTM

pkg/bindings/containers/logs.go
jwhonce: @mheon It could cause a double close on the body.

PR #6366, opened 2020-05-23
Turn on Ubuntu testing

rhatdan:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -423,13 +423,12 @@ testing_task:
         - name: "test ${PRIOR_FEDORA_NAME}"
           gce_instance:
               image_name: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
-        # TODO:
-        # - name: "test ${UBUNTU_NAME}"
-        #   gce_instance:
-        #       image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
-        # - name: "test ${PRIOR_UBUNTU_NAME}"
-        #   gce_instance:
-        #       image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
+        - name: "test ${UBUNTU_NAME}"
+          gce_instance:
+              image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
+        - name: "test ${PRIOR_UBUNTU_NAME}"
+          gce_instance:
+              image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
 
     timeout_in: 120m
 


rhatdan:
@edsantiago Looks like you registry script does not work on ubuntu?

edsantiago:
Entirely my fault: I took shortcuts when writing the script, and deferred error handling "for later". I deferred too long. I apologize.

I've filed #6375 to address this; errors should now be visible to the caller. @rhatdan if that passes, please feel free to rebase onto it so we can see what's failing here.

rhatdan:
@edsantiago It looks like podman does not like the `--root` flag on Ubuntu. I have no idea what is causing this?

edsantiago:
````
../podman-registry: line 102: --root: command not found
````
It looks like `$PODMAN` is not being set. It should be set in line 17 via:
````bash
PODMAN=${PODMAN:-$(type -p podman)}
````
I'm leaving home now and can't check, but I have a vague memory that there's some code that installs podman, then deletes all the installed files. I don't know why that would happen only on Ubuntu and not Fedora though. 

My two suggestions are to (1) set `PODMAN=/path/to/podman` in whatever code invokes this script, or (2) change the default in the script to `$(dirname $0)/../bin/podman` (to use the podman built in the tree). I don't have a good sense for which is better or if there are other options.

Gotta go, sorry. HTH

**Comment by edsantiago** [comment]
Found it: [contrib/cirrus/lib.sh:remove_packaged_podman_files()](https://github.com/containers/libpod/blob/master/contrib/cirrus/lib.sh#L403). What I don't understand is how then the script would work on Fedora.

I'm leaning slightly toward this solution, if you'd like to apply it and test with it.
```diff
--- a/hack/podman-registry
+++ b/hack/podman-registry
@@ -14,7 +14,7 @@ PODMAN_REGISTRY_PASS=
 PODMAN_REGISTRY_PORT=

 # Podman binary to run
-PODMAN=${PODMAN:-$(type -p podman)}
+PODMAN=${PODMAN:-$(dirname $0)/../bin/podman}

 # END   defaults
```

rhatdan:
Trying it out.

edsantiago:
@rhatdan it's not going to work; see #6390. Sorry.

edsantiago:
Yikes! Could this be a problem with a bad golang version on ubuntu?

rhatdan:
@AkihiroSuda @giuseppe Looks like rootless kit is pulling in requirements for newer versions of golang for ubuntu versions.  What do you guys suggest we do?

AkihiroSuda:
Could you be more specific? It should work with Go >= 1.13 and probably also 1.12.

rhatdan:
When compiling the code on Ubuntu we are seeing these errors.
```
[+0124s] # github.com/containers/libpod/vendor/k8s.io/apimachinery/pkg/util/errors
[+0124s] vendor/k8s.io/apimachinery/pkg/util/errors/errors.go:99:10: undefined: errors.Is
[+0124s] # github.com/containers/libpod/vendor/github.com/rootless-containers/rootlesskit/pkg/port/builtin/parent
[+0124s] vendor/github.com/rootless-containers/rootlesskit/pkg/port/builtin/parent/parent.go:94:9: undefined: "github.com/containers/libpod/vendor/github.com/pkg/errors".Is
```

AkihiroSuda:
Bumping up https://github.com/pkg/errors works?

https://github.com/pkg/errors seems tested with Go >= 1.11: https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/.travis.yml#L4

AkihiroSuda:
Ah, `Is()` seems to require Go 1.13
https://github.com/pkg/errors/blob/v0.9.1/go113.go

rhatdan:
Is golang 1.13 available on ubuntu 19?

AkihiroSuda:
Yes https://packages.ubuntu.com/eoan/golang-1.13

rhatdan:
Ok so we probably need to update the VM images.

rhatdan:
Closing in favour of #6418

PR #6375, opened 2020-05-24
podman-registry helper script: handle errors

edsantiago:
My initial revision of the podman-registry helper script
was written in haste, with an enormous tradeoff: no
visibility into any errors. We are now paying for this
in #6366: the script is failing on Ubuntu and we
have no way of knowing why.

This PR adds a must_pass() function used for critical
steps. This runs the action silently; if the command
fails, it displays the failing command name with
full output logs, cleans up the temporary workdir,
and exits with error status.

As a reminder, the reason this is necessary is that
our script convention is to output a series of
environment variables to stdout -- we must therefore
take pains not to emit anything else to stdout.
And, unfortunately, podman and openssl tend to be
rather verbose.

### DIFF
--- a/hack/podman-registry
+++ b/hack/podman-registry
@@ -104,6 +104,24 @@ function podman() {
               "$@"
 }
 
+###############
+#  must_pass  #  Run a command quietly; abort with error on failure
+###############
+function must_pass() {
+    local log=${PODMAN_REGISTRY_WORKDIR}/log
+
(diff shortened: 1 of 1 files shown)


vrothberg:
@edsantiago, the changes have caused a regression surfacing in me being unable to authenticate with the provided credentials. Reverting the commit has solved the issue. I couldn't spot which change would cause that though.

PR #6390, opened 2020-05-26
podman-registry: fix lost credentials

edsantiago:
must_pass(), added in #6375, eats the credentials generated
via 'podman run --entrypoint htpasswd'. Run that podman
instance directly, and add explicit error check.

The error and stdout/stderr handling here has gotten
cumbersome. There must be something I'm missing that
could make it all simpler.

Also: when setting $PODMAN, default to the locally built
one -- there may not be one in $PATH.

### DIFF
--- a/hack/podman-registry
+++ b/hack/podman-registry
@@ -14,7 +14,7 @@ PODMAN_REGISTRY_PASS=
 PODMAN_REGISTRY_PORT=
 
 # Podman binary to run
-PODMAN=${PODMAN:-$(type -p podman)}
+PODMAN=${PODMAN:-$(dirname $0)/../bin/podman}
 
 # END   defaults
 ###############################################################################
@@ -176,11 +176,16 @@ function do_start() {
               -out ${AUTHDIR}/domain.crt                               \
               -subj "/C=US/ST=Foo/L=Bar/O=Red Hat, Inc./CN=localhost"
 
-    # Store credentials where container will see them
-    must_pass podman run --rm                                        \
-              --entrypoint htpasswd ${PODMAN_REGISTRY_IMAGE}         \
-              -Bbn ${PODMAN_REGISTRY_USER} ${PODMAN_REGISTRY_PASS}   \
-              > $AUTHDIR/htpasswd
+    # Store credentials where container will see them. We can't run
+    # this one via must_pass because we need its stdout.
+    podman run --rm                                               \
+           --entrypoint htpasswd ${PODMAN_REGISTRY_IMAGE}         \
+           -Bbn ${PODMAN_REGISTRY_USER} ${PODMAN_REGISTRY_PASS}   \
+           > $AUTHDIR/htpasswd
+    if [ $? -ne 0 ]; then
+        rm -rf ${PODMAN_REGISTRY_WORKDIR}
+        die "Command failed: podman run [htpasswd]"
+    fi
 
     # In case someone needs to debug
     echo "${PODMAN_REGISTRY_USER}:${PODMAN_REGISTRY_PASS}" \
--- a/Makefile
+++ b/Makefile
@@ -318,7 +318,7 @@ localunit: test/goecho/goecho varlink_generate
 	ginkgo \
 		-r \
 		$(TESTFLAGS) \
-		--skipPackage test/e2e,pkg/apparmor,test/endpoint,pkg/bindings \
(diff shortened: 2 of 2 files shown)


hack/podman-registry
```diff
@@ -14,7 +14,7 @@ PODMAN_REGISTRY_PASS=
 PODMAN_REGISTRY_PORT=
 
 # Podman binary to run
-PODMAN=${PODMAN:-$(type -p podman)}
+PODMAN=${PODMAN:-$(dirname $0)/../bin/podman}
```
vrothberg, line 17: CI doesn't like that change.

edsantiago:
> I tested it successfully on my machine.

Thank you! I've pushed what I hope might be a fix for the path issue. @vrothberg do you have any other ideas on how to set a reasonable default for PODMAN here? See ongoing discussion in #6366 for context

vrothberg:
Thanks, Ed! To see the new script+bindings in action, see https://github.com/containers/libpod/pull/6207/files#diff-596fb9579398fc722f6c7dd243736347. I am repeating myself but cannot state enough how much I love that we can easily test authentication now :partying_face: 

> Thank you! I've pushed what I hope might be a fix for the path issue. @vrothberg do you have any other ideas on how to set a reasonable default for PODMAN here? See ongoing discussion in #6366 for context

Ideally I want _everything_ to be centrally managed in the `Makefile`. All paths and all set up should be done there, so we have a reproducible way of running tests with all paths being relative to the root, without recursion. We started moving parts of the CI into shell scripts in ./contrib which IMHO make things a bit complicated. It's also hard for new contributors to get started.

Long story short: I would love to move all test invocations into the `Makefile` and make all paths relative to the project's root. Couldn't we set the `PODMAN` env var in the Makefile?

vrothberg:
LGTM

edsantiago:
>Couldn't we set the PODMAN env var in the Makefile?

SGTM but I don't know which Makefile steps would need it; feel free to fix that in a separate PR. Here I'm just trying to set a fallback default. That had originally been podman-from-$PATH but for reasons I don't understand Ubuntu does not have that. 

Argh. CI is failing again.

edsantiago:
```
/var/tmp/go/src/github.com/containers/libpod/hack/../bin/podman: No such file or directory
```
@vrothberg what am I doing wrong here? Is this test somehow running before podman gets built?

vrothberg:
@edsantiago, it looks like it :(

Adding a `make podman` here (https://github.com/containers/libpod/blob/master/contrib/cirrus/unit_test.sh#L10) could work.

edsantiago:
>Adding a make podman

Ouch. There is no way I'm doing that. Is there any way to defer this test until later?

ITM I'm going to go back to looking at Ubuntu and see if I can figure out why it's missing podman.

vrothberg:
> > Adding a make podman
> 
> Ouch. There is no way I'm doing that. Is there any way to defer this test until later?

Why not? The integration test script does a `make`, `make install` before running.

edsantiago:
@vrothberg our CI is already too fragile: I don't want to add a `make` step anywhere without completely understanding its effects.

This situation is getting a little out of hand; I'm afraid of it becoming a band-aid fest where we keep patching kludges and kludges and end up with something unmaintainable. I really need to take time to understand what is going on here. Right now everything points to a test that's running at the wrong time/place: whatever this new test is that's running on the registry script, it needs to run after there is a built podman. How can we make that happen, _without_ changing the order of how/when podman is built in CI? How can we run this test later?

vrothberg:
> How can we make that happen, without changing the order of how/when podman is built in CI? How can we run this test later?

I "assume" that the `unit_test.sh` should be moved after `integration_test.sh` but that would mean that integration tests would run before unit tests. Personally, I don't see any harm in building Podman before the unit tests - in any case, that's what's happening in the order of the CI.

edsantiago:
I would strongly urge us not to muck with anything major like this while @cevich is away. I believe there was a reason for running `unit` before `integration`; I just can't find it in git logs right now.

IIUC the root of the problem is that `make localunit` runs ginkgo, which looks for `*_test.go`, and finds this new one under `hack`.

Proposal: in `Makefile`, under `localunit`, add `--skipPackage hack`. Then add `hack` to the `ginkgo` target. This will run the `hack` tests under `localintegration`. @vrothberg would that be acceptable?

vrothberg:
> @vrothberg would that be acceptable?

Sure, whatever works for you :+1: But again, I don't think that a `make build` is something critical.

PR #6398, opened 2020-05-26
Fix builds on 32 Arches.

rhatdan:


rhatdan:
Fixes: https://github.com/containers/libpod/issues/6397

mheon:
LGTM

rhatdan:
@edsantiago PTAL

edsantiago:
LGTM, it's consistent with previous fixes, but I have no way to test it other than letting koji do so.

/lgtm
/hold

TomSweeneyRedHat:
LGTM

baude:
@rhatdan i provided a PR to this to make sure make cross-local will catch this in CI.

edsantiago:
/hold

@mheon I thought we couldn't LGTM while tests were still running?

mheon:
@edsantiago It already had a hold, so I just reapplied the LGTM

edsantiago:
Sorry @mheon, I had missed that.

/hold cancel

edsantiago:
Followup: [koji builds are back to passing](https://koji.fedoraproject.org/koji/buildinfo?buildID=1516819). Thanks everyone.

rhatdan:
@edsantiago, @baude was asking why are we building for x86 32 bit environments?

edsantiago:
>why are we building for x86 32 bit environments?

That's a packaging question, I can't possibly answer it. However:
* in looking at the spec file, I see no arch exclusions or inclusions; and
* in looking at recent koji builds, I see that i686 is a standard target even for fc33.

From that, I would guess that this is simply The Way It Is in koji. It should be possible to add an arch exclusion though.

@lsm5 , can you answer or do you know who can answer?

rhatdan:
I just thought that Fedora no longer supported that arch.

baude:
seems wasteful to continue this for i686

edsantiago:
My understanding is that f31 dropped the 32-bit _kernel_ but still builds packages for multilib. But please take anything I say with a giant lump of salt.

lsm5:
>     * in looking at the spec file, I see no arch exclusions or inclusions; and
> 
>     * in looking at recent koji builds, I see that i686 is a standard target even for fc33.

@edsantiago Idon't remember off-hand but I've had random requests to not disable any of the supported arches, that's why I removed the `ExclusiveArch`. Does the i686 build imply more time and resources for testing? I can check with fedora-devel people as well on this.

rhatdan:
No, this is out of curiosity more then anything.  I am glad we caught this, and fixed the tests to catch it earlier in CI.
Someone somewhere is bound to build podman on a 32 bit system, so we should be ready for it.

Issue #6405, opened 2020-05-27
[CI flake] podman stats --format json - no output


vrothberg:
I've experienced flakes in the podman-stats e2e tests and could reproduce it manualy:
```
libpod (master) $ ./bin/podman stats --all --no-stream --format json|jq .
[
  {
    "id": "c74d1aa69138",
    "name": "jovial_kalam",
    "cpu_percent": "--",
    "mem_usage": "0.01%",
    "mem_percent": "1.274MB / 16.65GB",
    "net_io": "-- / --",
    "block_io": "-- / --",
    "pids": "1"
  }
]
libpod (master) $ ./bin/podman stats --all --no-stream --format json|jq .
```

The second call yields no output, ultimately yielding the marshal errors we see in the CI (see below):
```
[+0953s] Running: /var/tmp/go/src/github.com/containers/libpod/bin/podman --storage-opt vfs.imagestore=/tmp/podman/imagecachedir --root /tmp/podman_test294532388/crio --runroot /tmp/podman_test294532388/crio-run --runtime /usr/bin/crun --conmon /usr/bin/conmon --cni-config-dir /etc/cni/net.d --cgroup-manager systemd --tmpdir /tmp/podman_test294532388 --events-backend file --storage-driver vfs stats --all --no-stream --format json
[+0953s] output: 
[+0953s] unexpected end of JSON input
```

I'll try to further track it down and come up with a fix.

Cc @baude

PR #6406, opened 2020-05-27
container stats: fix --no-stream race

vrothberg:
Fix a race in `podman container stats` by waiting for the client to
consume the data in the channel.  This requires a `sync.WaitGroup` (or
semaphore) in the client and to also close the channel the backend.

Fixes: #6405

@baude PTAL

### DIFF
--- a/cmd/podman/containers/stats.go
+++ b/cmd/podman/containers/stats.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 	"text/tabwriter"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


vrothberg:
Somehow related: I find the [pod stats implementation](https://github.com/containers/libpod/blob/master/cmd/podman/pods/stats.go#L77) slightly less complicated as all the streaming logic is part of `cmd/podman` while the API always returns the stats of one call. This also allows for separating CLI and API options (see https://github.com/containers/libpod/blob/master/cmd/podman/pods/stats.go#L24).

baude:
LGTM

mheon:
Nice catch!
/lgtm

PR #6410, opened 2020-05-27
specgen: fix segfault

haircommander:
we should not access the devices without checking if the resources are there


mheon:
Nice catch, LGTM

TomSweeneyRedHat:
LGTM

PR #6416, opened 2020-05-28
fix `ps --last=N`

vrothberg:
Fix `ps --last=N` to also include non-running containers. Also add an
e2e test to prevent us from regressing in the future.

### DIFF
--- a/pkg/ps/ps.go
+++ b/pkg/ps/ps.go
@@ -23,7 +23,7 @@ func GetContainerLists(runtime *libpod.Runtime, options entities.ContainerListOp
 		filterFuncs []libpod.ContainerFilter
 		pss         []entities.ListContainer
 	)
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

TomSweeneyRedHat:
@vrothberg gating barfed and left zero indication of what went wrong.  Joyous.

TomSweeneyRedHat:
LGTM other than Gating inexplicably croaking.

edsantiago:
@TomSweeneyRedHat gating runs in a container image. From quay. Nuff said.

quay is reporting [read-only mode is up](https://status.quay.io/) so I restarted gating, and it passed. I don't know if the other tests will restart automatically; looks like they won't so this and all other open PRs may need a re-push.

PR #6418, opened 2020-05-28
Enable Ubuntu tests in CI

baude:
Add updates required for ubuntu and run integration tests

### DIFF
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -39,6 +39,17 @@ done
 cd "${GOSRC}/"
 case "${OS_RELEASE_ID}" in
     ubuntu)
+	    apt-get update
+	    apt-get install -y containers-common
+	    sed -ie 's/^\(# \)\?apparmor_profile =.*/apparmor_profile = ""/' /etc/containers/containers.conf
+	    if [[ "$OS_RELEASE_VER" == "19" ]]; then
+		    apt-get purge -y --auto-remove golang*
+		    apt-get install -y golang-1.13
+		    ln -s /usr/lib/go-1.13/bin/go /usr/bin/go
+	    fi
+	    if [[ "$OS_RELEASE_VER" == "20" ]]; then
+		    apt-get install -y python-is-python3
+	    fi
         ;;
     fedora)
         # All SELinux distros need this for systemd-in-a-container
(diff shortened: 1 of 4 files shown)


mheon:
LGTM once things go green

contrib/cirrus/setup_environment.sh
```diff
@@ -39,6 +39,15 @@ done
 cd "${GOSRC}/"
 case "${OS_RELEASE_ID}" in
     ubuntu)
+	    apt-get update 
+	    if [[ "$OS_RELEASE_VER" == "19" ]]; then
+		    apt-get purge -y --auto-remove golang* 
+		    apt-get install -y golang-1.13 
+		    ln -s /usr/lib/go-1.13/bin/go /usr/bin/go
+	    fi
+	    if [[ "$OS_RELEASE_VER" == "20" ]]; then
```
TomSweeneyRedHat, line 48: should we do "> 19" here?

contrib/cirrus/setup_environment.sh
```diff
@@ -39,6 +39,15 @@ done
 cd "${GOSRC}/"
 case "${OS_RELEASE_ID}" in
     ubuntu)
+	    apt-get update 
+	    if [[ "$OS_RELEASE_VER" == "19" ]]; then
+		    apt-get purge -y --auto-remove golang*
```
edsantiago, line 44: Could I recommend quoting the asterisk?
```bash
   ... --auto-remove "golang*"
```
Without the quotes, if there are one or more files named `golang<something>` in the current directory, the shell will expand that command line to `--auto-remove golang-file1 golang-file2 etc`. I realize this is unlikely and I'm probably being pedantic.

rhatdan:
@mheon Did some of your pod handling code futher break Ubuntu?

mheon:
Not that I'm aware of? That should all be OS-agnostic

rhatdan:
@edsantiago Looks like your registry script is not working on ubuntu.

edsantiago:
This is really strange. Everything points to a difference in the Ubuntu version of BATS: environment variables that are preserved in Fedora/RHEL are being clobbered in Ubuntu. The registry is starting, but subsequent tests aren't getting the right port / testuser / password. Is there some way to find out what version of BATS is being used? It's not listed in the `package_versions` CI step.

I'm also confused about why the `remote` tests are failing: this test is disabled on podman-remote. I think my test for `podman-remote` is failing because of the new APIv2 way. But that's another TODO for me.

vrothberg:
Ubuntu:20.04 is shipping `Bats 1.2.0-dev` while F32 is on 1.1.0.

vrothberg:
Opened https://github.com/containers/libpod/pull/6515. Bats v1.1.0 worked in the Ubuntu tests in the Google cloud.

PR #6419, opened 2020-05-28
Ensure that signal names can be parsed on Windows

mheon:
To ensure the Windows and OS X remote clients can properly parse container stop signal (when given as a name e.g. SIGTERM) and set it in SpecGen, we need access to a list of Linux signal names and the numbers they map to that is available on non-Linux OSes. Fortunately, these are ABI constants that are extremely unlikely to change, so we can just take the existing constant definitions from the library and use them.

The signal numbers used here are sourced from AMD64, but should be the same for every architecture that is not Alpha, SPARC, MIPS, and PA-RISC. So `podman run --stop-signal SIGTTOU` from a Windows client to a Podman service on a SPARC host will set an incorrect stop signal, but I don't think this is a large problem.

### DIFF
--- a/pkg/signal/signal_common.go
+++ b/pkg/signal/signal_common.go
@@ -1,0 +1,41 @@
+package signal
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+// ParseSignal translates a string to a valid syscall signal.
+// It returns an error if the signal map doesn't include the given signal.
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat @QiWang19 @baude @jwhonce @vrothberg @giuseppe PTAL

PR #6435, opened 2020-05-29
check --user range for rootless containers

QiWang19:
Check --user range if it's a uid for rootless containers. Returns error if it is out of the range. From https://github.com/containers/libpod/issues/6431#issuecomment-636124686

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -3,6 +3,7 @@ package containers
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/containers/libpod/cmd/podman/common"
@@ -10,7 +11,9 @@ import (
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/pkg/domain/entities"
 	"github.com/containers/libpod/pkg/errorhandling"
+	"github.com/containers/libpod/pkg/rootless"
 	"github.com/containers/libpod/pkg/specgen"
+	"github.com/containers/libpod/pkg/util"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -90,6 +93,15 @@ func run(cmd *cobra.Command, args []string) error {
 	cliVals.Net, err = common.NetFlagsToNetOptions(cmd)
 	if err != nil {
 		return err
+	}
+
+	if rootless.IsRootless() && !registry.IsRemote() {
+		userspec := strings.SplitN(cliVals.User, ":", 2)[0]
+		if uid, err := strconv.ParseInt(userspec, 10, 32); err == nil {
+			if err := util.CheckRootlessUIDRange(int(uid)); err != nil {
+				return err
+			}
+		}
 	}
 
 	if af := cliVals.Authfile; len(af) > 0 {
--- a/pkg/util/utils_linux.go
+++ b/pkg/util/utils_linux.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"github.com/containers/libpod/pkg/rootless"
 	"github.com/containers/psgo"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


cmd/podman/containers/run.go
```diff
@@ -101,7 +103,19 @@ func run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
+	if rootless.IsRootless() {
+		if uid, err := strconv.ParseInt(cliVals.User, 10, 64); err == nil {
```
mheon, line 107: There is a good chance that the user entered a username, not a string - this would be a lot more useful if it ran as part of libpod Init() after the container is mounted and know what the UID is.

cmd/podman/containers/run.go
```diff
@@ -101,7 +103,19 @@ func run(cmd *cobra.Command, args []string) error {
...
+			uids, _, err := rootless.GetConfiguredMappings()
+			if err != nil {
+				return err
+			}
+			for _, u := range uids {
+				if uid > int64(u.Size) {
```
mheon, line 113: We should probably add 1 here (we also map in the user's own UID)

cmd/podman/containers/run.go
```diff
@@ -101,7 +103,19 @@ func run(cmd *cobra.Command, args []string) error {
...
+					return errors.Errorf("UID %d is greater then the range of UIDs available to the user. Need to modify /etc/subuid to allow this UID.", uid)
```
mheon, line 114: Suggest `errors.Errorf("requested user's UID %d is too large for the rootless user namespace", uid)`

cmd/podman/containers/run.go
rhatdan: I think it might be better to do a double check.  Since the second check has to happen after the image has been pulled.  If the user gives us a UID, then we can check right away.  If they give us one with a username, we can report back that it will not work inside of the user namespace.

QiWang19:
@mheon @rhatdan PTAL

pkg/util/utils_linux.go
```diff
@@ -52,3 +53,18 @@ func FindDeviceNodes() (map[string]string, error) {
...
+
+// CheckRootlessUIDRange checks the uid within the rootless container is in the range from /etc/subuid
+func CheckRootlessUIDRange(uid int) error {
+	uids, _, err := rootless.GetConfiguredMappings()
+	if err != nil {
+		return err
+	}
+	for _, u := range uids {
+		// add 1 since we also map in the user's own UID
+		if uid+1 > u.Size {
```
mheon, line 65: Sorry, not what I meant - should be `uid > u.Size + 1`

QiWang19: fixed

pkg/util/utils_unsupported.go
```diff
@@ -10,3 +10,8 @@ import (
 func FindDeviceNodes() (map[string]string, error) {
 	return nil, errors.Errorf("not supported on non-Linux OSes")
 }
+
+// CheckRootlessUIDRange is not implemented anywhere except Linux.
+func CheckRootlessUIDRange(uid int) error {
+	return errors.Errorf("this function is not supported on this os")
```
rhatdan, line 16: I think this should return nil. If this is called by podman-remote, we don't want the check being done.

rhatdan:
The more I think about this, we only want this checked in podman-local not on podman-remote for the client.

pkg/util/utils_linux.go
```diff
@@ -52,3 +53,18 @@ func FindDeviceNodes() (map[string]string, error) {
 
 	return nodes, nil
 }
+
+// CheckRootlessUIDRange checks the uid within the rootless container is in the range from /etc/subuid
+func CheckRootlessUIDRange(uid int) error {
+	uids, _, err := rootless.GetConfiguredMappings()
```
rhatdan, line 59: I think there should be a remote check here and return nil if remote.

QiWang19: Disabled this test under cmd/podman if it's remote

Issue #6439, opened 2020-05-30
Running a sample container Docs Fail


Davidnet:
bug

**Description**

Using the sample tutorial located at [documentation link](https://github.com/containers/libpod/blob/master/docs/tutorials/podman_tutorial.md)

If trying to run the example: 

```bash
$ podman run -dt -p 8080:8080/tcp -e HTTPD_VAR_RUN=/var/run/httpd -e HTTPD_MAIN_CONF_D_PATH=/etc/httpd/conf.d \
                       -e HTTPD_MAIN_CONF_PATH=/etc/httpd/conf \
                       -e HTTPD_CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/httpd/ \
                       registry.fedoraproject.org/f27/httpd /usr/bin/run-httpd
Trying to pull registry.fedoraproject.org/f27/httpd...
  manifest unknown: manifest unknown
Error: unable to pull registry.fedoraproject.org/f27/httpd: unable to pull image: Error initializing source docker://registry.fedoraproject.org/f27/httpd:latest: Error reading manifest latest in registry.fedoraproject.org/f27/httpd: manifest unknown: manifest unknown
```

**Steps to reproduce the issue:**

1. Run 

```bash
$ podman run -dt -p 8080:8080/tcp -e HTTPD_VAR_RUN=/var/run/httpd -e HTTPD_MAIN_CONF_D_PATH=/etc/httpd/conf.d \
                       -e HTTPD_MAIN_CONF_PATH=/etc/httpd/conf \
                       -e HTTPD_CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/httpd/ \
                       registry.fedoraproject.org/f27/httpd /usr/bin/run-httpd
```

**Describe the results you received:**

Failed to resolve the get the container

**Describe the results you expected:**

Run example and get a http server

**Output of `podman version`:**

```
Version:            1.9.2
RemoteAPI Version:  1
Go Version:         go1.14.2
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.14.2
  podmanVersion: 1.9.2
host:
  arch: amd64
  buildahVersion: 1.14.8
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.16-2.fc32.x86_64
    path: /usr/bin/conmon
[...]

PR #6441, opened 2020-05-30
[CI:DOCS] update httpd location in tutorial

TomSweeneyRedHat:
Update the reference to the httpd in the Podman tutorial.
This just bumps it to F29 for now to get things working,
I'd prefer to get something set up in quay.io to make it
more resistant to Fedora version changes.

Fixes: #6439


rhatdan:
This fixes the issue, but it is not a long term solution.  We need to find an image in ubi8 or centos.
/lgtm

Issue #1070, opened 2018-07-10
Podman exits immediately when running as non-root user


aengelke:
> /kind bug

**Description**

When running `podman` as non-root user, the program exits immediately with status code zero. This is presumably from [cmd/podman/main.go:35](https://github.com/projectatomic/libpod/blob/master/cmd/podman/main.go#L35), where exit is called when we *became* root in the user namespace. (Note that `became` is `false` if and only if an error occurred -- this is already cought above.) Also, exit code zero doesn't make sense to me.

**Steps to reproduce the issue:**

1. Run `podman` as non-root

**Describe the results you received:**

Nothing happens

**Describe the results you expected:**

Help output

**Output of `podman version`:**

```
Version:       0.6.5
Go Version:    go1.10.3
OS/Arch:       linux/amd64
```

(as included in Silverblue; version 28.20180707.0)

mheon:
Can reproduce locally here on latest master. Going to take a look today.
@giuseppe This looks like the rootless issues @umohnani8 had as well.

mheon:
I don't think this is actually that line in main, though - the parent will be killed, but the child should still be running in the user namespace created for Podman.

giuseppe:
@mheon I can assign it to myself if you are fine, I had a similar issue when using `-ti` with selinux enabled, just to be sure it is the same one I am seeing, could you try with `semodule -DB` and see if there is any AVC in the journal?

mheon:
@giuseppe Not SELinux - I'm seeing no denials even with dontaudit turned off

aengelke:
No AVC denials in the journal, independently of SELinux configuration.

mheon:
This only appears to happen with packaged Podman - can't reproduce on source builds. My earlier test with latest master was actually a 0.6.5 RPM - I was using the wrong binary. Oops.

aengelke:
Thanks for the quick fix (and the workaround)!

PR #6442, opened 2020-05-31
Shell completion

### LINKED ISSUE #6440: Shell completion

/kind feature

**Description**

The current bash autocompletion is hard to maintain and buggy. Cobra can create completion shell scripts. This was already tried in PR #4768 but dropped because the lack of autocompletion support for containers/pods/volumes/etc...

With the current cobra version 1.0.0 you can now add a custom function to do this for flags and args
**cobra documentation:**
[main completion](https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-nouns-written-in-go)
[flag completion](https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-flags-written-in-go)

With this it would be possible to create autocompletion scripts for bash, zsh, fish, and powershell!

I would take the time to implement this if this is wanted?

Luap99:
Add shell completion with cobra
 
Allow automatic generation for shell completion scripts
with the internal cobra functions (requires v1.0.0)
  
This should replace the handwritten completion scripts
and even adds support for fish and powershell(if needed?)

We can now create the scripts with
- podman completion bash
- podman completion zsh
- podman completion fish
- podman completion powershell

to test the completion run:
`source <(podman completion bash)`

I added the main completion for all commands
but this still could be improved with
better custom completion functions.
see: cmd/podman/completion/completion.go

Flag completion is for now only added to
podman generate systemd --restart-policy
for now
just to show how to implement this on flags

This is still **WIP**

Closes #6440


Luap99:
@rhatdan @TomSweeneyRedHat please take a look :)

Luap99:
Ok i played with this a bit and tested the bash, zsh and fish completions. Bash and fish are working fine but zsh is not working properly (only completes static flags). There is already a [PR](https://github.com/spf13/cobra/pull/1070) in the cobra project which fixes this Issue so this should work once this get merged.

I start adding autocompletion support for the flags

TomSweeneyRedHat:
@saschagrunert just putting this on your radar as an FYI as I know you have poked around in this area on other projects.

rhatdan:
@edsantiago PTAL  I know you have played in this area in the past.

Luap99:
I started adding completion for flags.
I created the table below so you can see my progress and how each flag gets completed.

| Command                                                                              | Main Args          | Flags                                                                                                                                                                                                               |
| ------------------------------------------------------------------------------------ | ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| [podman-attach](podman-attach.1.md)<br>[podman-container-attach](podman-attach.1.md) | running containers | **--detach-keys** -> _ctrl-_                                                                                                                                                                                        |
| [podman-auto-update](podman-auto-update.1.md)                                        | none               |                                                                                                                                                                                                                     |
| [podman-build](podman-build.1.md)<br>[podman-image-build](podman-build.1.md)         | default shell      | skipping for now because flags are defined in buildah                                                                                                                                                               |
| [podman-commit](podman-commit.1.md)<br>[podman-container-commit](podman-commit.1.md) | all containers     | **--change** -> _CMD, ENTRYPOINT, ENV, EXPOSE, LABEL, STOPSIGNAL, USER, VOLUME, WORKDIR_ <br> **--format** -> _oci,docker_ <br> **--iidfile** -> default shell <br> **--author** -> none <br> **--message** -> none |
| [podman-container-checkpoint](podman-container-checkpoint.1.md)                      | running containers | **--export** -> default shell                                                                                                                                                                                       |
| [podman-container-cleanup](podman-container-cleanup.1.md)                            | exited containers  | **--exec** -> none                                                                                                                                                                                                  |
| [podman-container-exists](podman-container-exists.1.md)                              | all containers     |                                                                                                                                                                                                                     |
| [podman-container-prune](podman-container-prune.1.md)                                | none               | **--filter** -> none                                                                                                                                                                                                |
| [podman-container-restore](podman-container-restore.1.md)                            | all containers     | **--import** -> default shell <br> **--name** -> none                                                                                                                                                               |
| [podman-container-runlabel](podman-container-runlabel.1.md)                          | all images         | **--authfile** -> default shell <br> **--cert-dir** -> default shell <br> **--creds** -> none <br> **--name** -> none                                                                                               |

I will edit this post to keep the table up to date with my progress.

cmd/podman/completion/bash.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: bashCmd,
+		Parent:  completionCmd,
+	})
+	flags := bashCmd.Flags()
+	flags.StringVarP(&completionOptions.File, "file", "f", "", "Output the completion to file rather then stdout")
```
edsantiago, line 43: It's rather THAN, not then. Please fix all copy/paste instances of this.

edsantiago:
I love the idea of this, and what I've seen of the implementation so far (but it will take me many more hours to complete review). If that zsh PR ever gets merged, then vendored in here, I would love to replace my crufty hand-written zsh completion.

My first showstopper question is: how are you testing this? I fired up a bash shell and:
```console
$ source <(./bin/podman completion bash)
$ podman p<TAB>   --> bash: _get_comp_words_by_ref: command not found
```

Luap99:
you have to use `./bin/podman p<TAB>` because it implements some custom hidden args on the podman command

edsantiago:
I did (try both `podman` and `./bin/podman`); I redacted one just for brevity. No difference.

Luap99:
ok i installed it with `sudo make install` i think the new podman version needs to be in your $PATH

edsantiago:
Solved: you need to run this first:
```console
$ source /usr/share/bash-completion/bash_completion
```

cmd/podman/completion/completion.go
```diff
@@ -0,0 +1,368 @@
...
+				//suggested "registry.fedoraproject.org/f29/httpd:latest" as
+				// - "registry.fedoraproject.org/f29/httpd:latest"
+				// - "registry.fedoraproject.org/f29/httpd"
+				// - "f29/httpd:latest"
+				// - "f29/httpd"
+				// - "httpd:latest"
+				// - "httpd"
+				paths := strings.Split(repo, "/")
+				for i := range paths {
+					sugesstionWithTag := strings.Join(paths[i:], "/")
```
edsantiago, line 136: typo: it is spelled 'suggestion' (two 'g's, only one 's' in 'est')

cmd/podman/completion/completion.go
```diff
@@ -0,0 +1,368 @@
...
+				//suggest only full repo path if no input is given
+				suggestions = append(suggestions, repo)
+			} else {
+				//suggested "registry.fedoraproject.org/f29/httpd:latest" as
+				// - "registry.fedoraproject.org/f29/httpd:latest"
+				// - "registry.fedoraproject.org/f29/httpd"
+				// - "f29/httpd:latest"
+				// - "f29/httpd"
+				// - "httpd:latest"
+				// - "httpd"
```
edsantiago, line 133: I would strongly encourage you to add image IDs: not all images have names. (You should probably check for that)

cmd/podman/completion/completion.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	pods, err := registry.ContainerEngine().PodPs(registry.GetContext(), listOpts)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	for _, pod := range pods {
+		//include name in suggestions
+		//if ids should be also completed
```
edsantiago, line 84: Yes. Container IDs aren't usually user-visible, but I can attest that it is reeeeally convenient to `podman ps`, see a container I want to operate on, and type just the first two characters of the SHA and want completion on it.

cmd/podman/completion/completion.go
```diff
@@ -0,0 +1,368 @@
...
+	switch {
+	case containers != nil && images != nil:
+		//combine both
+		return append(containers, images...), cobra.ShellCompDirectiveNoFileComp
+
+	case containers != nil:
+		return images, cobra.ShellCompDirectiveNoFileComp
+
+	case images != nil:
+		return containers, cobra.ShellCompDirectiveNoFileComp
```
edsantiago, line 321: I think the logic here is reversed

edsantiago: Actually, thinking about it, what does `append()` do with nil? Wouldn't it be cleaner to just eliminate the `switch` entirely and trust that nil will be handled correctly?

edsantiago:
Can you find a way to disable the showing of global flags on subcommands? It is the opposite of useful to see 15+ suggestions on a command that has only two flags:
```console
$ podman rmi -<TAB><TAB>
--all                --identity           --runroot=
--cgroup-manager     --identity=          --runtime
--cgroup-manager=    --log-level          --runtime=
--cni-config-dir     --log-level=         --storage-driver
--cni-config-dir=    --namespace          --storage-driver=
--conmon             --namespace=         --storage-opt
--conmon=            --network-cmd-path   --storage-opt=
--events-backend     --network-cmd-path=  --syslog
--events-backend=    --remote             --tmpdir
--force              --remote=            --tmpdir=
--help               --root               -a
--hooks-dir          --root=              -f
--hooks-dir=         --runroot            -r
```

cmd/podman/completion/completion.go
mheon: `append(nil, entry)` will produce a 1-length array containing `entry`. I haven't personally tested `append(entry, nil)` but I'm fairly certain it will produce the original array, unchanged.

edsantiago: I tested this and it seems to work perfectly fine:
```diff
index f2b787b5b..91336cdba 100644
--- a/cmd/podman/completion/completion.go
+++ b/cmd/podman/completion/completion.go
@@ -309,20 +309,7 @@ func AutocompleteContainersAndImages(cmd *cobra.Command, args []string, toComple
        containers, _ := getContainers("all")
        images, _ := getImages(toComplete)

-       switch {
-       case containers != nil && images != nil:
-               //combine both
-               return append(containers, images...), cobra.ShellCompDirectiveNoFileComp
-
-       case containers != nil:
-               return images, cobra.ShellCompDirectiveNoFileComp
-
-       case images != nil:
-               return containers, cobra.ShellCompDirectiveNoFileComp
-
-       default:
-               return nil, cobra.ShellCompDirectiveError
-       }
+       return append(containers, images...), cobra.ShellCompDirectiveNoFileComp
 }

 //Autocomplete volumes.
```
Tested with `podman inspect` and:
* no images, no containers
* 1 image, no containers
* 2 images, no containers
* 2 images, 1 container
* 2 images, 2 containers

Did not test with 0 images, 2 containers :-)

I think the above is much more maintainable; it could also be applied to several other places in the code.

Luap99: yeah thats is mush better

edsantiago:
There is a weird inconsistency in the autogenerated flag completions, and I can't figure out where it's happening:
```
    flags+=("--something=")
    two_word_flags+=("--something")
```
Note that the first one has an `=` suffix, the second does not. This produces duplicate completions, as seen in my earlier comment above (`--something` and `--something=`). I think it makes most sense to remove the `=` but as long as you stick with one format I'm OK with it.

Also, is there a way to add option completion to flags? E.g. `--cgroup-manager` currently uses file completion, it should offer only two choices.

And hey, I know I'm griping a lot, but this is really great work. I really like it and think it's much more maintainable than the current bash completions. Thank you for your work on this.

Luap99:
I think thats because these are persistent flags and can therefore used on any subcommand
`podman rmi --log-level info imagename` is completely valid

edsantiago:
It may be valid but it's completely useless, and no human will ever want to do that. The possibility is there due to an accident of history. If there is any possible way to remove those from completion, I encourage you to do so.

Luap99:
> There is a weird inconsistency in the autogenerated flag completions, and I can't figure out where it's happening:
> 
> ```
>     flags+=("--something=")
>     two_word_flags+=("--something")
> ```
> 
> Note that the first one has an `=` suffix, the second does not. This produces duplicate completions, as seen in my earlier comment above (`--something` and `--something=`). I think it makes most sense to remove the `=` but as long as you stick with one format I'm OK with it.
> 
Thats how it should be since we can both use `--log-level info` and `--log-level=info`
> Also, is there a way to add option completion to flags? E.g. `--cgroup-manager` currently uses file completion, it should offer only two choices.

Sure stared to implement this already on some flags see: https://github.com/containers/libpod/pull/6442#issuecomment-637775222

Luap99:
> It may be valid but it's completely useless, and no human will ever want to do that. The possibility is there due to an accident of history. If there is any possible way to remove those from completion, I encourage you to do so.

I will look into this.

edsantiago:
>Thats how it should be since we can both use --log-level info and --log-level=info

Understood, but our target audience are (presumably) people with more than ten minutes of Linux experience, and understand how to use flags. The clutter added by duplicate flag entries is horrendous: try `podman create -<TAB><TAB>`. It is actually evil. There should be one and only one completion convention.

>Sure stared to implement this already on some flags

Oops, I missed that - thank you!

edsantiago:
zsh: I suspect this is a bug in cobra itself, not your code: any flag whose default is empty causes a zsh completion error:
```console
zsh$ podman volume create<TAB>
_arguments:comparguments:325: invalid option definition: (*-l *--label)*-l[Set metadata for a volume (default [])]:

zsh$ podman completion zsh | grep -3 'Set metadata for a volume'   pr/6442
function _podman_volume_create {
  _arguments \
    '--driver[Specify volume driver name]:' \
    '(*-l *--label)'{\*-l,\*--label}'[Set metadata for a volume (default [])]:' \
    '(*-o *--opt)'{\*-o,\*--opt}'[Set driver specific options (default [])]:' \
    '--cgroup-manager[Cgroup manager to use ("cgroupfs"|"systemd")]:' \
    '--cni-config-dir[Path of the configuration directory for CNI networks]:' \
```
There are 35 instances of this. One easy solution is to `sed -e 's/\[\]/none/'` or `s/\(\[\|\]\)/\\\1/g`

edsantiago:
>There is already PR in cobra which should fix the zsh issues

Fair enough. I had looked at https://github.com/spf13/cobra/pull/1070 to see if it addressed this, and not found a clear indication, but it's an enormous PR and it's very likely I missed something. I will not look at zsh again for now.

Luap99:
There its mentioned https://github.com/spf13/cobra/issues/1129

cmd/podman/completion/completion.go
edsantiago: I think all image IDs should be listed, honestly: there are often cases where intermediate images have no tags, and someone running `podman rmi <TAB>` might want to see the full list of both IDs and tags. But this is just an opinion based on gut feel, and I don't have any evidence to support it. I think it would be OK to do what you think is best, and add a big FIXME comment saying that this needs UX testing. @mheon @rhatdan @TomSweeneyRedHat WDYT?

mheon: I'm a little worried that printing all IDs could get overwhelming, it's easy to accumulate a lot of dangling images by using `podman build`. I would expect us to offer tab completion once a few characters are typed, though.

marckhouzam:
> zsh: I suspect this is a bug in cobra itself, not your code: any flag whose default is empty causes a zsh completion error:
> 
> ```
> zsh$ podman volume create<TAB>
> _arguments:comparguments:325: invalid option definition: (*-l *--label)*-l[Set metadata for a volume (default [])]:
> ```

This has been reported here https://github.com/spf13/cobra/issues/1129
and I can confirm the problem will be fixed with #1070.

marckhouzam:
> > Thats how it should be since we can both use --log-level info and --log-level=info
> 
> Understood, but our target audience are (presumably) people with more than ten minutes of Linux experience, and understand how to use flags. The clutter added by duplicate flag entries is horrendous: try `podman create -<TAB><TAB>`. It is actually evil. There should be one and only one completion convention.

I agree with you about not showing both `--flag` and `--flag=`.  I've proposed to remove the `--flag=` form in https://github.com/spf13/cobra/pull/1070 which, if accepted, will affect `zsh` and `fish`.  It will not change things for `bash`, but I have something in the pipeline for `bash`.  For now, I'm not aware of a way for you to remove the `--flag=` form for `bash` without having support from Cobra (unless you grep out the `flags+=("--something=")` everywhere after generating the completion script).

If you could weigh in on the comment below to support this suggested change, it would help get it accepted.  Thanks.
https://github.com/spf13/cobra/pull/1070#discussion_r422680307

Luap99:
I have added the completion for all commands/flags except the build, login and logout flags because their are defined in buildah and containers/common. They could also be implemented here but I would prefer to define them in their repo. There is still much room for improvement in the completion functions.

I know that this has a lot of changes to look through but the biggest part contains only the flags and command definition. The main logic lives in `cmd/podman/common/completion.go`. 

This does not replace the current shell completion scripts. I would recommend to wait for this until https://github.com/spf13/cobra/pull/1146 is merged.

rhatdan:
@Luap99 Please rebase this PR?  What is the current state?

Luap99:
@rhatdan rebased

> What is the current state?

I would say the biggest part is done. There are still many completions which could be improved but this already a huge amount and i don't have enough time for now. So i would call it ready for review.

rhatdan:
@Luap99 Sadly it needs another rebase.
@edsantiago @vrothberg @QiWang19 PTAL

Luap99:
OK, done.

Please note that the zsh and bash script are not available in their final form - it would be best to rely on this PR.: https://github.com/spf13/cobra/pull/1146

Luap99:
You're right, I have removed the unused code. A large part of the changes are due to formatting and documentation.

The main advantage is that we can support bash/zsh/fish and powershell completion. This also ensures that we don't miss any flags or commands in completion, as they are automatically added (unless they are hidden).

I also noticed that there is no completion for podman-remote. You can simply do `podman-remote completion bash` with this. It even omits the unsupported flags (e.g. --latest ) but i need to figure out how to complete the remote images,containers,etc...

This would also be useful for windows powershell users. However, the powershell script supports completion only for commands and flags, but not for the arguments (containers, images, etc...). I did not test powershell yet because i dont know how to build the windows client.

edsantiago:
@Luap99 I believe the upstream changes have been committed; have they been vendored into podman yet? I would really, really like to see this PR incorporated. It is great work.

cmd/podman/completion/completion.go
```diff
@@ -0,0 +1,85 @@
...
+	completionOptions = entities.CompletionOptions{}
+	completionCmd     = &cobra.Command{
+		Use:              fmt.Sprintf("completion [flags] {%s}", strings.Join(shells, ",")),
+		Short:            "Generate shell autocompletions",
+		Long:             completionDescription,
+		TraverseChildren: true,
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return shells, cobra.ShellCompDirectiveNoFileComp
+		},
+		Args: cobra.ExactValidArgs(1),
```
marckhouzam, line 33: This is valid, but it won't do what it is meant to do.  It's not a big deal but I wanted to point it out.

Because you use `ValidArgsFunction` instead of `ValidArgs`, which is the best way in my mind too, so that you can turn off file completion for the shell parameter, the `ExactValidArgs()` function no longer knows that the valid args are, so it will just allow any arg.  But you do catch anything invalid in your code yourself.

This may be something Cobra could improve upon, but I don't have a clear idea yet.

Luap99:
@edsantiago The zsh changes are merged. Which means that zsh and fish shell completion should be fully usable. The bash completion also works but currently provides flag suggestions with and without the equal form. The generated bash script also has something like 18k lines. That's the reason I temporary replaced the go module with @marckhouzam new bash completion branch. PR: https://github.com/spf13/cobra/pull/1146.
I did this to test his changes and provide him with feedback. I can confirm that this works very well and it even enables support for descriptions in bash. While testing I noticed two bugs: https://github.com/spf13/cobra/pull/1162 and https://github.com/spf13/cobra/pull/1171. The latter one would allow podman --remote to complete remote images/containers/etc.. which would be a great improvement although it also requires a running podman endpoint for the subcommand and flag completion. But this might be something I could change if I figure out how to remote wiring exactly works.

This results in two options:

1. Wait for the bash PR to get merged.
2. Vendor in the current cobra master (there is no new release with the zsh change) and use this PR to make it the default for fish and zsh shell completion and keep using our handwritten bash script until the bash PR lands.

rhatdan:
@Luap99 This PR needs a rebase, and has cobra got the fixes to allow us to remove the WIP?

Luap99:
@rhatdan Still waiting on https://github.com/spf13/cobra/pull/1146

However I vendored this PR in for now so you can test this if you want.

## How to test this PR
bash:
`source <(bin/podman completion bash)`
zsh:
`bin/podman completion zsh -f "${fpath[1]}/_podman"` and reload the shell
fish:
`bin/podman completion fish | source`

To disable descriptions use it like this. `source <(bin/podman completion bash --no-desc)`
Make sure you do `bin/podman [TAB]`

## How it works
Cobra creates hidden command called `__complete` which will provide the completion values. The shell completion scripts are static and only call this cobra command with the current arguments. This command then provides the correct completions and the shell script just parses this output.
Full documentation: https://github.com/spf13/cobra/blob/master/shell_completions.md

## Advantages
- All commands and flags are automatically completed (unless there are hidden or deprecated). There is no need to do anything if you add a command or flag.
- The custom completion logic is written in go and there is no need to do this for each shell. Just add a `ValidArgsFunction`. to the command or flag. You can see such functions in `cmd/podman/common/completion.go`.
- Support for the fish shell.
- Eventually support for powershell (not with this PR as this also needs some cobra changes https://github.com/spf13/cobra/pull/1208).
- Full support for `podman-remote` and `podman --remote` including the parsing from `--url` and `--connection` values so that the the correct containers/images from the remote endpoint are completed.
- Fix some currently not working behavior e.g. `podman --log-level info [TAB]`

## TODO
- I have added some FIXMEs which need your attention.
- Add/Cleanup some tests for this.
- Flags for `build` do not have custom completion functions yet as there are defined in `buildah`. These flags will default to path completion. They can be easily added in buildah later on. The same goes for `login` and `logout`.

I think this is a great improvement in usability, so I really recommend that you try it.

/cc @edsantiago @mheon @TomSweeneyRedHat @vrothberg

rhatdan:
@Luap99 Lets get this into Podman even if we don't abandon the hand built completions right away.  So people can play with it.

Make `podman completions` hidden so we don't have to document and support it for other use cases other then in the completion scripts.

rhatdan:
@baude @edsantiago @mheon @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
@Luap99 once this is merged, could you please create a PR in github.com.podman.io and update the install instructions there with a small section on how to use this?  Somewhere under this [Building from scratch](https://podman.io/getting-started/installation#building-from-scratch) section of the doc.

I took a very high level scan, the changes LGTM.  I just don't have time to do a very detailed review at this point.

edsantiago:
I continue to think this is super-exciting work, but in testing it right now I notice that flag completion on subcommands is including global options:
```console
$ ./bin/podman ps -<TAB>
--all               -a  -- Show ...   <--good
--cgroup-manager        -- Cgroup ... <--BAD
...
```
Can you configure this so global options are only shown if there's no subcommand?

Luap99:
> Can you configure this so global options are only shown if there's no subcommand?

Well this is not natively supported. Persistent flags are valid in this case and can be used there. This is how most programs use them. I agree with you that this is not really that useful for podman.

The correct way to stop this would be to define the flags as local ones on the root cmd. However this would break backwards compability. Local flags could only be used like this: `podman --log-level info run ...`, `podman run --log-level info ...` would no longer work. I think this is not desirable.

I believe I could hack around this by dropping the unwanted flags on the fly if a subcmd gets completed. I will try this if you really prefer that option.

edsantiago:
The undesirable flags don't show up in `podman subcommand --help`, so there must be some internal flag that distinguishes them. Can that be used by the completion code?

Luap99:
> The undesirable flags don't show up in podman subcommand --help, so there must be some internal flag that distinguishes them. Can that be used by the completion code?

Yes I can drop the undesired flags easily but this comes at a cost. `podman ps --log[TAB]` would no longer get completed. I just want to point this out. I agree with you that hiding them could result in a better user experience.

A compromise could be that we show the persistent flags only after at least three chars are typed. That should be enough to not print to much.

edsantiago:
> I can drop the undesired flags easily but this comes at a cost. podman ps --log[TAB] would no longer get completed

That's a feature, not a bug! `--log-level` is a global option, not a subcomand option. It should never be visible on subcommands. It is, due to unfortunate backward-compatibility reasons. Maybe some scripts use it, but command-line users (the ones using completion) should be encouraged to use those only at the root level (`podman --log-level=xxx`). 

>show the persistent flags only after at least three chars are typed

I don't think they should be shown at all. (At the subcommand level).

Luap99:
@edsantiago This works as you wish now.

edsantiago:
@Luap99 you need to rebase on current master to fix CI. My apologies.

Luap99:
I have opened https://github.com/containers/common/pull/314 to implement the login/logout flag completions.

rhatdan:
@Luap99 Needs a rebase.  I would love to finally push this over the line.

Luap99:
@rhatdan Rebased

The last commit still needs a buildah release with the changes in https://github.com/containers/buildah/pull/2656. Or alternatively this could also be merged without the last commit for now as this only adds the completions for the build command.

rhatdan:
Ok lets cut a buildah release and move forward

cmd/podman/common/completion.go
```diff
@@ -0,0 +1,638 @@
...
+				// - "f29/httpd:latest"
+				// - "f29/httpd"
+				// - "httpd:latest"
+				// - "httpd"
+				paths := strings.Split(repo, "/")
+				for i := range paths {
+					suggestionWithTag := strings.Join(paths[i:], "/")
+					suggestionWithoutTag := strings.SplitN(suggestionWithTag, ":", 2)[0]
+					suggestions = append(suggestions, suggestionWithTag, suggestionWithoutTag)
+				}
```
Luap99, line 130: This logic does not work correct for images pulled via digest. I will fix this.

I wonder what the best ux for image completion would be. Currently we only complete the full reference with and without tag. I personally would prefer if we also complete the short form, e.g. `alpine`. But this could result in many displayed suggestions. With the current shortnames work I think it would make sense to display all defined shortnames and all full references for the present images.

@vrothberg WDYT? Is it possible to retrieve all defined shortnames here?

Luap99 (commented):
This is ready for review.

Please see this https://github.com/containers/podman/pull/6442#issuecomment-690987239 if you want to test it.

edsantiago:
I've been testing this out all afternoon, and it's just beautiful work. I'd really like to see this merged before the tree changes and @Luap99 needs to rebase yet again. @containers/podman-maintainers can we make an effort to review and merge this?

TomSweeneyRedHat:
Thought I'd commented on this.  
FWIW, it LGTM, but I'm not at all a subject matter expert here.

Luap99:
Rebased with the new doc format since this would have blocked CI post-merge otherwise.

edsantiago:
@Luap99 I don't think it would've blocked, since github would merge on top of master (i.e. rebasing for you) but thanks.

I didn't have a strong urge to repeat my testing this morning, and yesterday was a nice quiet PR day, so I just diff'ed against the previous submission:
```console
$ git cdif pr/6442{-rev05,}
```
...and this is the only change I see which does not look like one of yesterday's PRs:
```diff
diff --git a/cmd/podman/completion/completion.go b/cmd/podman/completion/completion.go
index 71d07082c..ead8d1f05 100644
--- a/cmd/podman/completion/completion.go
+++ b/cmd/podman/completion/completion.go
@@ -87,7 +87,7 @@ func completion(cmd *cobra.Command, args []string) error {
        }

        _, err = io.WriteString(w, fmt.Sprintf(
-               "# This file is generated with %q; see: podman-completion(1)\n", cmd.CommandPath(),
+               "\n# This file is generated with %q; see: podman-completion(1)\n", cmd.CommandPath(),
        ))
        return err
 }
```
LGTM!

Issue #6409, opened 2020-05-27
apiv2: GET /libpod/images/nonesuch/get : leaks api.tar tmpfile


edsantiago:
Given an APIv2 server running:
```console
$ ./bin/podman system service --time 999 tcp:127.0.0.1:5000 &
[1] 1414075
```
...attempting to get a nonexistent image will leave behind an empty `api.tarNNNNNN` file in `$TMPDIR`:
```console
$ ls -lt $TMPDIR/|head
....
$ curl http://localhost:5000/v1.40/libpod/images/nonesuch/get
{"cause":"no such image", ...etc }
$ ls -lt $TMPDIR/|head
-rw-------. 1 esm  esm      0 May 27 12:16 api.tar857349796
....
```
The compat API does not suffer from this problem:
```console
$ curl http://localhost:5000/v1.40/images/nonesuch/get
{"cause":"no such image", ....}
$ ls -lt $TMPDIR/|head
-rw-------. 1 esm  esm      0 May 27 12:16 api.tar857349796    (same one as before)
....
```

mheon:
@baude @jwhonce PTAL

PR #6447, opened 2020-06-01
Fix leak of empty tarball

baude:
In cases of trying to export an image, if the image was not found, we leaked an empty tarball or directory depending on the format.

Fixes: #6409


mheon:
LGTM

TomSweeneyRedHat:
LGTM

PR #6450, opened 2020-06-01
Fix breakage in upstream CI

mheon:
Alpine 3.12 seems to have renamed the `man` package - it no longer exists. Switch man-db, which I believe is the same thing.


mheon:
Swapped from man-db (pulls down 20mb) to less (pulls down 6mb)

mheon:
Closing in favor of #6447

Issue #5466, opened 2020-03-11
[1.8] rootless container with --userns=keep-id does not forward ports


duritong:
/kind bug

**Description**

I am not able to get a rootless containter that forwards port if using `--userns=keep-id`

**Steps to reproduce the issue:**

```
$ podman run --userns=keep-id -d -p 127.0.0.1:55432:80 fedora:31 bash -c 'sleep 100'
eb001430c61a16bbeaa1a5d584d219a7be0920b9536af4627de3d5b07c919498
$ ss -nlpt | grep 55432
$ 

```

**Describe the results you received:**

Container runs - but Port 55432 is not bound

**Describe the results you expected:**

Port 55432 is bound

**Additional information you deem important (e.g. issue happens only occasionally):**

If I remove `--userns=keep-id` Port gets bound and traffic forwarded:

```
$ podman run -d -p 127.0.0.1:45432:80 fedora:31 bash -c 'sleep 100' 
9dd2b3c7d6f4a9d503dd6ef445a104c164badd272727d3131f62dc7c6bd301d8
$ ss -nlpt | grep 45432
LISTEN    0         4096             127.0.0.1:45432            0.0.0.0:*        users:(("exe",pid=47914,fd=11))   

```

**Output of `podman version`:**

```
$ podman version
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.6
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```$ podman info --debug
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.6
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.10-2.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.10, commit: 6b526d9888abb86b9e7de7dfdeec0da98ad32ee0'
  Distribution:
    distribution: fedora
    version: "31"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 1279201280
  MemTotal: 16674430976
  OCIRuntime:
    name: crun
    package: crun-0.12.2.1-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
[...]

Issue #5046, opened 2020-01-31
podman exec -i truncates the STDIN input


goneri:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

Hi,

I tried to get Ansible working with pipelining. Pipelining makes use of STDIN. The following command works fine with `1.8.0-dev`:

`echo 'print("a\n");'| podman exec -i 'ubi8' '/bin/sh' -c '/usr/bin/python3.6 && sleep 0' `

But I get a failure if I try to inject something bigger (an Ansiblz). I managed to reproduce the problem with `dd` and an ubi8 container:

```
[goneri@t580-gone c]$ cat /usr/bin/vim | /usr/bin/dd of=/tmp/a
6336+1 records in
6336+1 records out
3244296 bytes (3.2 MB, 3.1 MiB) copied, 0.012102 s, 268 MB/s
[goneri@t580-gone c]$ cat /usr/bin/vim | podman exec -i 'ubi8' '/usr/bin/dd' of=/tmp/a
904+1 records in
904+1 records out
463112 bytes (463 kB, 452 KiB) copied, 0.0150833 s, 30.7 MB/s
[goneri@t580-gone c]$ cat /usr/bin/vim | podman exec -i 'ubi8' '/usr/bin/dd' of=/tmp/a
904+1 records in
904+1 records out
463112 bytes (463 kB, 452 KiB) copied, 0.0147833 s, 31.3 MB/s
[goneri@t580-gone c]$ cat /usr/bin/vim | podman exec -i 'ubi8' '/usr/bin/dd' of=/tmp/a
904+1 records in
904+1 records out
463112 bytes (463 kB, 452 KiB) copied, 0.0139259 s, 33.3 MB/s
[goneri@t580-gone c]$ cat /usr/bin/vim | podman exec -i 'ubi8' '/usr/bin/dd' of=/tmp/a
912+1 records in
912+1 records out
467208 bytes (467 kB, 456 KiB) copied, 0.0139809 s, 33.4 MB/s
```

1.

Pass a content to `podman exec -i`.

2.

The content is truncated.

**podman-1.8.0-1580495541.git4699d5e0.fc31.x86_64**

mheon:
Can you try this as root, and see if that resolves the issue? Want to see if this is definitely Podman or potentially fuse-overlayfs

goneri:
Hi @mheon,

I don't see any difference:

```
[...]

PR #6456, opened 2020-06-01
system tests : more tests

edsantiago:
- exec: add test for #5046, in which conmon swallowed chars
  on a large byte transfer

- pod: add 'pod exists' tests, both positive and negative;
  consolidate tests; add '--label', and check in 'pod inspect'
  add 'pod ps' tests

- networking: add test for #5466, in which detached run
  with --userns=keep-id would not forward a port


mheon:
LGTM

rhatdan:
More tests are good tests...
/lgtm
/hold

PR #6467, opened 2020-06-02
make env handling os dependent

baude:
environment variables are handled differently on windows vs linux.  here we split them to be handled but no actually processing of windows environment variables was done.  it can be added for future.  hoowever, now we dont get errors on windows about processing them.

### DIFF
--- a/pkg/env/env_unsupported.go
+++ b/pkg/env/env_unsupported.go
@@ -1,0 +1,8 @@
+// +build !linux,!darwin
+
+package env
+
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

pkg/env/env_unsupported.go
```diff
@@ -0,0 +1,8 @@
+// +build windows
```
mheon, line 1: Can we make this `!linux,!darwin`?

TomSweeneyRedHat:
LGTM

PR #6468, opened 2020-06-02
Enable detached exec for remote

mheon:
The biggest obstacle here was cleanup - we needed a way to remove detached exec sessions after they exited, but there's no way to tell if an exec session will be attached or detached when it's created, and that's when we must add the exit command that would do the removal. The solution was adding a delay to the exit command (5 minutes), which gives sufficient time for attached exec sessions to retrieve the exit code of the session after it exits, but still guarantees that they will be removed, even for detached sessions. This requires Conmon 2.0.17, which has the new `--exit-delay` flag.

As part of the exit command rework, we can drop the hack we were using to clean up exec sessions (remove them as part of inspect). This is a lot cleaner, and I'm a lot happier about it.

Otherwise, this is just plumbing - we need a bindings call for detached exec, and that needed to be added to the tunnel mode backend for entities.

### DIFF
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -46,6 +46,9 @@ case "${OS_RELEASE_ID}" in
 
         workaround_bfq_bug
 
+	# HACK: Need Conmon 2.0.17, currently in updates-testing on F31.
+	dnf update -y --enablerepo=updates-testing conmon
+
         if [[ "$ADD_SECOND_PARTITION" == "true" ]]; then
             bash "$SCRIPT_BASE/add_second_partition.sh"
         fi
--- a/pkg/api/handlers/compat/exec.go
+++ b/pkg/api/handlers/compat/exec.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/pkg/api/handlers"
 	"github.com/containers/libpod/pkg/api/handlers/utils"
+	"github.com/containers/libpod/pkg/specgen/generate"
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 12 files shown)


rhatdan:
You need to modify the spec file to require the version of conmon or something newer.

Well we at least need to do this for Fedora/RHEL/Centos/Ubuntu...

mheon:
Bumped COPR spec to 2.0.17

mheon:
Starting to go green, @rhatdan @vrothberg @baude @TomSweeneyRedHat @jwhonce PTAL

rhatdan:
LGTM

contrib/cirrus/setup_environment.sh
```diff
@@ -46,6 +46,9 @@ case "${OS_RELEASE_ID}" in
 
         workaround_bfq_bug
 
+	# HACK: Need Conmon 2.0.17, currently in updates-testing on F31.
+	dnf update -y --enablerepo=updates-testing conmon
```
TomSweeneyRedHat, line 50: Any way we can push this through to stable?  I'd like to get the hack part out of this.

pkg/api/handlers/compat/exec.go
```diff
@@ -104,15 +123,6 @@ func ExecInspectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	utils.WriteResponse(w, http.StatusOK, inspectOut)
```
jwhonce, line 125:
```suggestion
	utils.WriteResponse(w, http.StatusCreated, inspectOut)
```
From https://docs.docker.com/engine/api/v1.40/#operation/ContainerExec

mheon: This isn't Exec, this is ExecStart, which is a 200

contrib/cirrus/setup_environment.sh
mheon: It's in the process of pushing, but I don't know how long it takes to make it into the repos

TomSweeneyRedHat:
If we go with the enable-repo testing bit in, I'd like a V2 tagged issue spun to remove that before release.

Other than that and @jwhonce's issue, LGTM

PR #6469, opened 2020-06-02
V2 Add support for ssh authentication methods

jwhonce:
* `podman --remote ssh://<user>:<password>@<host>:<port><path>`
* `podman --remote ssh://<user>:<password>@<host>:<port><path> --identity <path> --passphrase <phrase>`
* `ssh-add <key>`
  `podman --remote ssh://<user>@<host><path>`
* Fix `podman help` to run even if podman missing components
* Prompt for passphrase on stdin IFF key is protected and passphrase
  not given via any other configuration

* cobra flags do not support optional value flags therefore refactored
  --remote to be a boolean and --url will now contain the URI to Podman
  service

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -103,6 +103,11 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 
 	cfg := registry.PodmanConfig()
 
+	// Help is a special case, no need for more setup
+	if cmd.Name() == "help" {
+		return nil
+	}
+
 	// Prep the engines
 	if _, err := registry.NewImageEngine(cmd, args); err != nil {
 		return err
@@ -150,6 +155,11 @@ func persistentPostRunE(cmd *cobra.Command, args []string) error {
 	// TODO: Remove trace statement in podman V2.1
 	logrus.Debugf("Called %s.PersistentPostRunE(%s)", cmd.Name(), strings.Join(os.Args, " "))
 
+	// Help is a special case, no need for more cleanup
+	if cmd.Name() == "help" {
+		return nil
+	}
+
 	cfg := registry.PodmanConfig()
 	if cmd.Flag("cpu-profile").Changed {
 		pprof.StopCPUProfile()
@@ -191,8 +201,11 @@ func loggingHook() {
 
 func rootFlags(opts *entities.PodmanConfig, flags *pflag.FlagSet) {
 	// V2 flags
-	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress(), "URL to access Podman service")
-	flags.StringSliceVar(&opts.Identities, "identity", []string{}, "path to SSH identity file")
+	flags.BoolVarP(&opts.Remote, "remote", "r", false, "Access remote Podman service (default false)")
+	// TODO Read uri from containers.config when available
+	flags.StringVar(&opts.Uri, "url", registry.DefaultAPIAddress(), "URL to access Podman service (CONTAINER_HOST)")
+	flags.StringSliceVar(&opts.Identities, "identity", []string{}, "path to SSH identity file, (CONTAINER_SSHKEY)")
+	flags.StringVar(&opts.PassPhrase, "passphrase", "", "passphrase for identity file (not secure, CONTAINER_PASSPHRASE), ssh-agent always supported")
 
 	cfg := opts.Config
 	flags.StringVar(&cfg.Engine.CgroupManager, "cgroup-manager", cfg.Engine.CgroupManager, "Cgroup manager to use (\"cgroupfs\"|\"systemd\")")
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -61,6 +61,9 @@ Podman and libpod currently support an additional `precreate` state which is cal
 **--identity**=*path*
 Path to SSH identity file
 
+**--passphrase**=*secret*
+pass phrase for SSH identity file
+
 **--log-level**=*level*
 
 Log messages above specified level: debug, info, warn, error (default), fatal or panic (default: "error")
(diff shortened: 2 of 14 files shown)


cmd/podman/root.go
```diff
@@ -191,8 +191,9 @@ func loggingHook() {
 
 func rootFlags(opts *entities.PodmanConfig, flags *pflag.FlagSet) {
 	// V2 flags
-	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress(), "URL to access Podman service")
-	flags.StringSliceVar(&opts.Identities, "identity", []string{}, "path to SSH identity file")
+	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress(), "URL to access Podman servic (PODMAN_HOST)")
+	flags.StringSliceVar(&opts.Identities, "identity", []string{}, "path to SSH identity file, (PODMAN_SSHKEY)")
+	flags.StringVar(&opts.PassPhrase, "passphrase", "", "passphrase for identity file (not secure, PODMAN_PASSPHRASE), ssh-agent always supported")
```
edsantiago, line 196: I really can't think of any situation in which this would be desirable.

pkg/bindings/connection.go
```diff
@@ -307,17 +338,23 @@ func (h *APIResponse) IsServerError() bool {
...
+func publicKey(path string, passphrase string) (ssh.AuthMethod, error) {
 	key, err := ioutil.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
 
 	signer, err := ssh.ParsePrivateKey(key)
 	if err != nil {
-		return nil, err
+		// TODO Should we prompt for passPhrase if Zero?
```
edsantiago, line 349: It's perfectly fine, and even common practice, to have an empty passphrase

cmd/podman/root.go
rhatdan: I would call it password  And have it support "-"  to take password from stdin.

podman login --password, -p is already supported.

cmd/podman/root.go
```diff
@@ -191,8 +191,9 @@ func loggingHook() {
 
 func rootFlags(opts *entities.PodmanConfig, flags *pflag.FlagSet) {
 	// V2 flags
-	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress(), "URL to access Podman service")
-	flags.StringSliceVar(&opts.Identities, "identity", []string{}, "path to SSH identity file")
+	flags.StringVarP(&opts.Uri, "remote", "r", registry.DefaultAPIAddress(), "URL to access Podman servic (PODMAN_HOST)")
```
rhatdan, line 194: I would prefer this be handled as two different options.  Since we can now store or soon will be able store the remoteuri in containers.conf we need a way to change podman to use the remote protocol
podman --use-remote
podman --client

Or do we change remote to be a bool and add --remote-uri

pkg/bindings/connection.go
```diff
@@ -69,23 +74,28 @@ func JoinURL(elements ...string) string {
...
 		secure bool
 	)
-	if v, found := os.LookupEnv("PODMAN_HOST"); found {
+	if v, found := os.LookupEnv("PODMAN_HOST"); found && uri == "" {
 		uri = v
 	}
 
-	if v, found := os.LookupEnv("PODMAN_SSHKEY"); found {
-		identity = []string{v}
+	if v, found := os.LookupEnv("PODMAN_SSHKEY"); found && len(identities) == 0 {
```
rhatdan, line 86: Can we use CONTAINER_SSHKEY and CONTAINER_PASSWORD?  In case we build future tools to use server?

cmd/podman/root.go
jwhonce: @rhatdan I am trying to differentiate between the password for authentication to the remote server vs the passphrase to open the key file.

@edsantiago It is up to the user to determine when it's acceptable to give the phrase on the command line.

pkg/bindings/connection.go
jwhonce: @rhatdan Sure, these are hold-overs from V1

pkg/bindings/connection.go
jwhonce: @edsantiago How about prompting in the case that the key file requires it and it's not given?

edsantiago: I would really expect that to be handled by the underlying ssh helper - it can't realistically be done anywhere else.

docs/source/markdown/podman.1.md
```diff
@@ -73,18 +76,21 @@ When namespace is set, created containers and pods will join the given namespace
 **--network-cmd-path**=*path*
 Path to the command binary to use for setting up a network.  It is currently only used for setting up a slirp4netns network.  If "" is used then the binary is looked up using the $PATH environment variable.
 
-**--remote**, **-r**=*url*
-URL to access Podman service (default "unix:/run/user/3267/podman/podman.sock")
+**--remote**, **-r**
+Access Podman service will be remote
+
+**--url**=*value*
+URL to access Podman service (default from `containers.conf` or lastly "unix:/run/user/3267/podman/podman.sock").
```
mheon, line 83: Can we be more verbose about the default? Root doesn't look there, and we should say that we use `/run/user/$UID` to be completely clear.

pkg/domain/entities/engine.go
```diff
@@ -43,11 +43,13 @@ type PodmanConfig struct {
 	EngineMode     EngineMode       // ABI or Tunneling mode
 	Identities     []string         // ssh identities for connecting to server
 	MaxWorks       int              // maximum number of parallel threads
+	PassPhrase     string           // ssh passphrase for identity for connecting to server
 	RegistriesConf string           // allows for specifying a custom registries.conf
+	Remote         bool             // Connection to API Service "remote"
```
mheon, line 48: Can the comment here be made a little more clear? "This is a connection to a remote Podman instance using the REST API" or similar?

PR #6477, opened 2020-06-03
RHEL8 and Centos8 don't have oci-runtime yet

rhatdan:
For the time being we need to just require runc
this should fix rdoproject.org/github-check


TomSweeneyRedHat:
LGTM

PR #6482, opened 2020-06-03
Ensure that image/container inspect are specialized

mheon:
We are currently able to inspect images with `podman container inspect` and containers with `podman image inspect` and neither of those seem correct. This ensures that the appropriate flags, and only the appropriate flags, are available for each specialized exec, and they can only inspect the specific type they were intended to.


edsantiago:
LGTM pending tests. I restarted a flake in bindings.

TomSweeneyRedHat:
LGTM assuming happy tests

PR #6483, opened 2020-06-03
When stopping containers locally, ensure cleanup runs

### LINKED ISSUE #5747: race: 'podman stop' does not always remove all podman mounts
One of the e2e tests, `podman list running container` in [test/e2e/mount_test.go](https://github.com/containers/libpod/blob/master/test/e2e/mount_test.go#L185), occasionally [flakes in CI](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5127055163523072/html/integration_test.log.html#t--podman-list-running-container--1). Basically, `podman mount` is finding an active mountpoint even after the container is `podman stopped`. Rerunning `podman mount` a second later finds no mounts, so it's pretty likely a race: maybe there's some cleanup that isn't happening in time. 

If this race condition **is OK**, that is, if it doesn't matter whether `podman mount` shows a dead mountpoint from a stopped container, then the e2e test must be fixed.

If this race condition **is not OK**, that is, if `podman stop` should guarantee that there are no mount points when it exits, then podman stop must be fixed.

Reproducer:

```bash
#!/bin/bash

set -e

T0=$SECONDS

while :;do
    cid=$(podman run -dt docker.io/library/alpine:latest top)
    podman mount --notruncate | grep -q $cid
    podman stop $cid > /dev/null
    m2=$(podman mount --notruncate)
    if [[ "$m2" =~ $cid ]]; then
        echo "FOO! Still mounted!"
        echo "$m2"
        echo "time = $(( $SECONDS - $T0 )) seconds"
        sleep 1
        echo
        echo "after sleep 1:"
        podman mount
        exit 1
    fi
    podman rm $cid >/dev/null
done
```

Sample run:
```console
# /tmp/mtest
FOO! Still mounted!
2b473b127c369cd11da3c775780ea2e693e17511534c1e296980a35443d14a70 /var/lib/containers/storage/overlay/5798db64ca882ed888a403a5398d1bb5d020c8c0bab719ba67a71b13673b4773/merged
time = 96 seconds

after sleep 1:
```
CI failure is f30; I can reproduce with podman-1.8.0-4.fc30.

Problem still present in rawhide: podman-1.8.3-0.75.dev.gitf7dffed.fc33

mheon:
The cleanup process was already running and ensuring that mounts and networking configuration was cleaned up on container stop, but this was async from the actual `podman stop` command which breaks some expectations - the container is still mounted at the end of `podman stop` and will be cleaned up soon, but not immediately. Fortunately, it's a trivial change to resolve this.

Fixes #5747

### DIFF
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -179,6 +179,12 @@ func (ic *ContainerEngine) ContainerStop(ctx context.Context, namesOrIds []strin
 			report.Err = err
 			reports = append(reports, &report)
 			continue
+		} else if err := con.Cleanup(ctx); err != nil {
+			// Only if no error, proceed to cleanup to ensure all
+			// mounts are removed before we exit.
+			report.Err = err
+			reports = append(reports, &report)
+			continue
 		}
 		reports = append(reports, &report)
 	}


TomSweeneyRedHat:
lint is biting you @mheon

PR #6484, opened 2020-06-03
Namespace fields were set with bogus values

rhatdan:
Namespace fields were being set with their option names
unless overriden by the user.  This was causing issues
with podman-remote on userns.


baude:
seems like a winner ++
LGTM

awesome @rhatdan

mheon:
LGTM

PR #6486, opened 2020-06-03
add socket information to podman info

baude:
this is step 1 to self-discovery of remote ssh connections.  we add a remotesocket struct to info to detect what the socket path might be.

### DIFF
--- a/docs/source/markdown/podman-info.1.md
+++ b/docs/source/markdown/podman-info.1.md
@@ -71,6 +71,9 @@ host:
       commit: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
       spec: 1.0.1-dev
   os: linux
+  remoteSocket:
+    exists: false
+    path: /run/user/1000/podman/podman.sock
   rootless: true
   slirp4netns:
     executable: /bin/slirp4netns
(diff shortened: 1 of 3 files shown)


rhatdan:
man page examples need to be updated.

pkg/domain/infra/abi/system.go
```diff
@@ -25,7 +25,36 @@ import (
...
+	socketPath := filepath.Join(xdg, "podman", "podman.sock")
+	rs := define.RemoteSocket{
+		Path:   socketPath,
+		Exists: false,
+	}
+
+	// Check if the socket exists
+	if _, err := os.Stat(socketPath); err == nil {
+		rs.Exists = true
+	}
```
jwhonce, line 51:
```suggestion
	if fi, err := os.Stat(socketPath); err == nil {
	        if fi.Mode()&os.ModeSocket == 0 {
		    rs.Exists = true
		}
	}
```

baude:
> man page examples need to be updated.

updated now

rhatdan:
LGTM

vrothberg:
Restarted the jobs. Look like Quay had a hiccup.

docs/source/markdown/podman-info.1.md
```diff
@@ -71,6 +71,9 @@ host:
       commit: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
       spec: 1.0.1-dev
   os: linux
+  remoteSocket:
+    exists: false
```
edsantiago, line 75: I think this is inaccurate: when I test it on my system, I don't get an `exists` line at all. Could it be the `omitempty` below? I'd like to see this reflect reality, either by removing the `omitempty` or removing the `false` here.

Issue #6490, opened 2020-06-04
podman-api fails with 'runtime error: invalid memory address or nil pointer dereference'


jgallucci32:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman API returns an error when querying information for certain docker image

**Steps to reproduce the issue:**

1. Start Podman service in separate session
```
podman system service
````

2. Pull image
```
podman pull docker.io/rancher/rke-tools:v0.1.56
```

3. Use cURL to get information about image using API
```
curl --unix-socket /run/podman/podman.sock http://d/v1.24/images/rancher/rke-tools:v0.1.56/json
```

**Describe the results you received:**
Results in the following error
```
{"cause":"runtime error: invalid memory address or nil pointer dereference","message":"runtime error: invalid memory address or nil pointer dereference","response":500}
```

**Describe the results you expected:**
Checking a different image such as `docker.io/library/hello-world` produces the expected results
```
# curl --unix-socket /run/podman/podman.sock http://d/v1.24/images/library/hello-world/json
[...]

PR #6491, opened 2020-06-04
/images/.../json: fix port parsing

vrothberg:
Fix a bug when parsing the `ExposedPorts` of the image that lead to
panics when the field was set.  The OCI image spec allows three valid
formats: `tcp/port`, `udp/port` and `port`

Fixes: #6490
Reported-by: @jgallucci32

### DIFF
--- a/pkg/api/handlers/types.go
+++ b/pkg/api/handlers/types.go
@@ -334,16 +334,25 @@ func ImageDataToImageInspect(ctx context.Context, l *libpodImage.Image) (*ImageI
 func portsToPortSet(input map[string]struct{}) (nat.PortSet, error) {
 	ports := make(nat.PortSet)
 	for k := range input {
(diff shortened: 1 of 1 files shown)


vrothberg:
@baude @giuseppe @jwhonce @mheon @rhatdan PTAL

rhatdan:
LGTM
But what happens when the user just specifies `port`.  Does this just default to `tcp` or `udp` or are you supposed to use both?

vrothberg:
> LGTM
> But what happens when the user just specifies `port`. Does this just default to `tcp` or `udp` or are you supposed to use both?

Then it defaults to `tcp`. That's what the spec wants:

https://github.com/containers/libpod/pull/6491/files#diff-2c1834810ca2c977e6dec2367e27701eR341-R342

PR #6493, opened 2020-06-04
V2 podman system connection

jwhonce:
* Implement command
* Refactor podman to pull from containers.conf/connections.conf by default
* Write podman-system-connection.1.md

### DIFF
--- a/cmd/podman/registry/config.go
+++ b/cmd/podman/registry/config.go
@@ -68,7 +68,6 @@ func newPodmanConfig() {
 		}
 	}
 
-	// FIXME: for rootless, add flag to get the path to override configuration
 	cfg, err := config.NewConfig("")
 	if err != nil {
 		fmt.Fprint(os.Stderr, "Failed to obtain podman configuration: "+err.Error())
@@ -83,7 +82,7 @@ func newPodmanConfig() {
 	podmanOptions = entities.PodmanConfig{Config: cfg, EngineMode: mode}
 }
 
-// SetXdgDirs ensures the XDG_RUNTIME_DIR env and XDG_CONFIG_HOME variables are set.
+// setXdgDirs ensures the XDG_RUNTIME_DIR env and XDG_CONFIG_HOME variables are set.
 // containers/image uses XDG_RUNTIME_DIR to locate the auth file, XDG_CONFIG_HOME is
 // use for the libpod.conf configuration file.
 func setXdgDirs() error {
--- a/cmd/podman/registry/config_tunnel.go
+++ b/cmd/podman/registry/config_tunnel.go
@@ -2,6 +2,13 @@
 
 package registry
 
+import (
+	"os"
+)
+
 func init() {
 	abiSupport = false
+
+	// Enforce that podman-remote == podman --remote
+	os.Args = append(os.Args, "--remote")
 }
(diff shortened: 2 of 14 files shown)


cmd/podman/registry/config.go
```diff
@@ -68,22 +68,23 @@ func newPodmanConfig() {
...
 		fmt.Fprint(os.Stderr, "Failed to obtain podman configuration: "+err.Error())
 		os.Exit(1)
 	}
 
-	cfg.Network.NetworkConfigDir = cfg.Network.CNIPluginDirs[0]
-	if rootless.IsRootless() {
-		cfg.Network.NetworkConfigDir = ""
+	cfg.Network.NetworkConfigDir = ""
+	if !rootless.IsRootless() {
+		if len(cfg.Network.CNIPluginDirs) > 0 {
```
mheon, line 79: Why this change? CNI should never enter the picture in rootless mode

cmd/podman/registry/config_tunnel.go
```diff
@@ -2,6 +2,13 @@
...
 
+import (
+	"os"
+)
+
 func init() {
 	abiSupport = false
+
+	// Enforce that podman-remote == podman --remote
+	os.Args = append(os.Args, "--remote")
```
mheon, line 13: Is there a potential problem if we duplicate this? Should we only append if --remote is not already set?

cmd/podman/root.go
```diff
@@ -103,13 +103,22 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
...
-
 	// Help is a special case, no need for more setup
 	if cmd.Name() == "help" {
 		return nil
 	}
 
+	cfg := registry.PodmanConfig()
+	if cfg.Remote {
+		if !cmd.Flag("url").Changed {
+			_ = cmd.Flags().Set("url", cfg.Config.Engine.RemoteURI)
```
rhatdan, line 114: Why not just default to these values, and only use them if user specifies remote.  That way the user will see them if he does podman run --help?

cmd/podman/registry/config.go
jwhonce: During testing with the remote client, the previous code panic'ed on the [0]

cmd/podman/registry/config_tunnel.go
jwhonce: duping, doesn't hurt. This is faster than searching the list and appending.

cmd/podman/root.go
jwhonce: D*mn, and I even had a TODO to implement this way.  Fixed.

```
$ bin/podman --help |grep identity                                                    
      --identity string           path to SSH identity file, (CONTAINER_SSHKEY) (default "/home/jhonce/.ssh/ocp-tier3") 
```

cmd/podman/registry/config.go
```diff
@@ -68,22 +68,23 @@ func newPodmanConfig() {
 		}
 	}
 
-	// FIXME: for rootless, add flag to get the path to override configuration
-	cfg, err := config.NewConfig("")
+	cfg, err := config.ReadCustomConfig()
```
rhatdan, line 71: This should not change.  This eliminates all of the defaults and reads only a custom config.  It should never be done for the global setting, only when you are customising the local.
This is why SELinux is breaking.

rhatdan:
LGTM

PR #6498, opened 2020-06-04
Ensure that containers in pods properly set hostname

### LINKED ISSUE #6494: podman pod --hostname: not working in RHEL8
```console
# podman pod create --name=mypod --hostname=foo.com
1f91b7aa5ff3f5255c31073d98474ee15e2ca1a66f89fb120c3eb21bec6ff2ab

# podman run --rm --pod mypod alpine hostname
44b350a23c72      <---- expected: foo.com

# podman run --rm --pod mypod alpine hostname
7ead774d363f      <---- different result each time
```
podman-2.0.0-0.3.rc4.module+el8.3.0+6918+d949891b.x86_64

Both root and rootless

mheon:
When we moved to the new Namespace types in Specgen, we made a distinction between taking a namespace from a pod, and taking it from another container. Due to this new distinction, some code
that previously worked for both `--pod=$ID` and `--uts=container:$ID` has accidentally become conditional on only the latter case. This happened for Hostname - we weren't properly setting it in cases where the container joined a pod. Fortunately, this is an easy fix once we know to check the condition.

Also, ensure that `podman pod inspect` actually prints hostname.

Fixes #6494

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -50,7 +50,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"Drop capabilities from the container",
 	)
 	createFlags.String(
-		"cgroupns", containerConfig.CgroupNS(),
+		"cgroupns", "",
 		"cgroup namespace to use",
 	)
 	createFlags.StringVar(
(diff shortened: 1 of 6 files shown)


mheon:
Second commit pushed with further fixes. This resolves the issue, as tested on a RHEL8 CI VM provided by @edsantiago 

The issue is the presence of a `containers.conf` file. We need to make sure we have a default containers.conf on our CI VMs so we catch things like this.

mheon:
@rhatdan PTAL. I don't think you'll like the second commit, but I don't see a better way without a big rewrite of the specgen population code.

rhatdan:
I am fine with this change. LGTM
I think this is the best for now although we could add some hanlding of remote to github.com/containers, so it only returns the values when grabbing locally.

mheon:
@vrothberg @baude @jwhonce PTAL, let's get this merged, then.

PR #6501, opened 2020-06-05
Fix handling of ThrottleWriteIOPSDevice

rhatdan:
This is causing the UBuntu tests to fail.


rhatdan:
@giuseppe @baude @jwhonce @mheon @vrothberg @TomSweeneyRedHat PTAL

giuseppe:
LGTM, I had opened exactly the same PR: https://github.com/containers/libpod/pull/6502

rhatdan:
There is still one more issue with Ubuntu.  The real problem is without Ubuntu we are not testing against Cgroups V1 right now.

PR #6504, opened 2020-06-05
Fix handling of systemd.

rhatdan:
Systemd enablement has to happen on the server side, since we need
check if the image is running systemd.

Also need to make sure user setting the StopSignal is not overriden on the
server side. But if not set and using systemd, we set it correctly.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -3,7 +3,6 @@ package common
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -285,16 +284,13 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 		s.NetNS = c.Net.Network
 	}
 
-	// STOP SIGNAL
-	signalString := "TERM"
 	if sig := c.StopSignal; len(sig) > 0 {
-		signalString = sig
+		stopSignal, err := util.ParseSignal(sig)
+		if err != nil {
+			return err
+		}
+		s.StopSignal = &stopSignal
 	}
-	stopSignal, err := util.ParseSignal(signalString)
-	if err != nil {
-		return err
-	}
-	s.StopSignal = &stopSignal
 
 	// ENVIRONMENT VARIABLES
 	//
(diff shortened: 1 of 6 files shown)


rhatdan:
@baude @mheon @giuseppe @vrothberg @QiWang19 @jwhonce @TomSweeneyRedHat PTAL
This should fix the last Ubuntu failure.  BTW This failed on Fedora but we disabled the test. We would get yelled at a lot by the Security team if this broke.

Now turned on testing for cgroupV2 also, no idea why this was disabled.

cmd/podman/common/specgen.go
```diff
@@ -439,25 +435,7 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
-	if systemd {
-		if s.StopSignal == nil {
-			stopSignal, err = util.ParseSignal("RTMIN+3")
-			if err != nil {
-				return errors.Wrapf(err, "error parsing systemd signal")
-			}
-			s.StopSignal = &stopSignal
-		}
-	}
+	s.Systemd = c.Systemd
```
giuseppe, line 438: should we leave the default still to true in `createFlags.StringVar` and set `s.Systemd` only when it is changed?

pkg/specgen/generate/container_create.go
```diff
@@ -128,7 +130,39 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
 	if s.Stdin {
 		options = append(options, libpod.WithStdin())
 	}
-	if len(s.Systemd) > 0 {
+
+	useSystemd := false
+	switch s.Systemd {
+	case "always", "true":
```
giuseppe, line 136: I think `"true"` should be handled as `""`, not `"always"`, that is the behaviour in 1.9

cmd/podman/common/specgen.go
rhatdan: The problem is figuring out if the value changed.  It is a lot of hacking for not much value.  Systemd really does not have a default, Always probably should not exist.  We should have had three values
"true", "false", "" that way if the user forces it to TRUE we setup to run systemd, like we do with Always. "" means the code determines if you are running with systemd mode.

But this is not the default and would be a breaking change. So really "true" means nothing now, or the same as "".

mheon: We could alias `true` to `always` - I think that 99% of people specifying `--systemd` intended that in the first place?

rhatdan: Well almost no one probably specifies that command other then to turn it on for debugging purposes. But lets get this fix in and we can move on.

baude:
lgtm

TomSweeneyRedHat:
LGTM

PR #6505, opened 2020-06-05
Add parallel operation to `podman stop`

mheon:
This is the other command that benefits greatly from being run in parallel, due to the potential 15-second timeout for containers that ignore SIGTERM.

While we're at it, also clean up how stop timeout is set. This needs to be an optional parameter, so that the value set when the container is created with `--stop-timeout` will be respected.


rhatdan:
LGTM

PR #6515, opened 2020-06-08
V2 enable ubuntu

vrothberg:
Meant to replace https://github.com/containers/libpod/pull/6418.

### DIFF
--- a/hack/install_bats.sh
+++ b/hack/install_bats.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -e
+
+die() { echo "${1:-No error message given} (from $(basename $0))"; exit 1; }
+
+buildDir=$(mktemp -d)
+git clone https://github.com/bats-core/bats-core $buildDir
+
+pushd $buildDir
+pwd
+git reset --hard ${VERSION}
+./install.sh /usr/local
+popd
+
+rm -rf $buildDir
--- a/test/e2e/checkpoint_test.go
+++ b/test/e2e/checkpoint_test.go
@@ -232,7 +232,8 @@ var _ = Describe("Podman checkpoint", func() {
 	})
 
 	It("podman checkpoint container with established tcp connections", func() {
-		Skip("fails on ubuntu")
+		// Broken on Ubuntu.
+		SkipIfNotFedora()
 		localRunString := getRunString([]string{redis})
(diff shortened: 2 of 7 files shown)


vrothberg:
Don't merge. Added a bad commit but want to let it run to test.

rhatdan:
@vrothberg Could you squash your commits.  "m" is a very informative commit message, but I have a `Q` about it.  :^)

vrothberg:
> @vrothberg Could you squash your commits. "m" is a very informative commit message, but I have a `Q` about it. :^)

he, I noticed too but want to wait and see if that turns tests green.

hack/install_bats.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -e
+
+die() { echo "${1:-No error message given} (from $(basename $0))"; exit 1; }
+
+buildDir=$(mktemp -d)
+git clone https://github.com/bats-core/bats-core $buildDir
+
+pushd $buildDir
+echo `pwd`
```
edsantiago, line 11: nit (minor, no need to fix): it's cleaner to just say `pwd`

vrothberg: Thanks, fixed now

test/e2e/checkpoint_test.go
```diff
@@ -232,6 +232,7 @@ var _ = Describe("Podman checkpoint", func() {
 	})
 
 	It("podman checkpoint container with established tcp connections", func() {
+		Skip("fails on ubuntu")
```
mheon, line 235: Why is this a plain Skip? Shouldn't this be a `SkipIfNotFedora()` too?

vrothberg: Inherited from #6418 but a good catch! Will update with the next push.

TomSweeneyRedHat:
LGTM assuming happy tests and @mheon's comment is addressed

edsantiago:
The BATS failure is an Ubuntu bug: they forked their own BATS and broke it. I've filed an [Ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/bats/+bug/1882542)

vrothberg:
Can someone drop a 'lgtm hold'?

edsantiago:
bindings failed. I'm assuming it's a flake. Restarted.

PR #6516, opened 2020-06-08
fix api fails with 'strconv.ParseUint: parsing "tcp": invalid syntax'

zhangguanzhang:
fix https://github.com/containers/libpod/issues/6511
https://github.com/containers/libpod/blob/master/pkg/api/handlers/compat/containers.go#L466-L471
https://github.com/containers/libpod/blob/1fcb6788a5d7471a7ca6215a40e36e21812a0f6e/pkg/api/handlers/compat/containers.go#L466-L471
`nat.NewPort`
https://github.com/docker/go-connections/blob/480d0b58242613178844e8ffabde7a5d55658f0d/nat/nat.go#L34
the first arg should be protocol，such as `splitp` is `[]string{"80","tcp"}`,
```
nat.NewPort(splitp[0], splitp[1])  ==> nat.NewPort(splitp[1], splitp[0])
```

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -465,8 +465,11 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON, 
 
 	ports := make(nat.PortSet)
 	for p := range inspect.HostConfig.PortBindings {
-		splitp := strings.Split(p, "/")
-		port, err := nat.NewPort(splitp[0], splitp[1])
+		splitp := strings.SplitN(p, "/", 2)
+		if len(splitp) != 2 {
+			return nil, errors.Errorf("PORT/PROTOCOL Format required for %q", p)
+		}
+		port, err := nat.NewPort(splitp[1], splitp[0])
 		if err != nil {
 			return nil, err
 		}


pkg/api/handlers/compat/containers.go
```diff
@@ -466,7 +466,7 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
 	ports := make(nat.PortSet)
 	for p := range inspect.HostConfig.PortBindings {
 		splitp := strings.Split(p, "/")
-		port, err := nat.NewPort(splitp[0], splitp[1])
+		port, err := nat.NewPort(splitp[1], splitp[0])
```
rhatdan, line 469: This looks like we should have an error check here as well, to make sure that the splitp is at least two items.

pkg/api/handlers/compat/containers.go
```diff
@@ -466,7 +466,7 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
 	ports := make(nat.PortSet)
 	for p := range inspect.HostConfig.PortBindings {
 		splitp := strings.Split(p, "/")
```
rhatdan, line 468:
```
splitp := strings.SplitN(p, "/",2)
if len(splitp) != 2 {
       return errors.Errorf("PORT/PROTOCOL Format required for %q",p)
}
```

rhatdan:
@mheon PTAL

zhangguanzhang:
it always return two items

pkg/api/handlers/compat/containers.go
mheon: I think having length(1) is valid - in that case, we force protocol to "tcp".

At some point, we're going to rewrite this to compile down into a SpecGen so we can maximize code reuse (and not duplicate our creation logic), so I'm not terribly concerned about how maintainable it is for now.

rhatdan:
@zhangguanzhang Should still check in-case someone eventually changes the call.

pkg/api/handlers/compat/containers.go
rhatdan: Then that use case needs to be handled.

zhangguanzhang:
https://github.com/containers/libpod/blob/a85e97952994a2803a17ee2a384b260115f18b43/pkg/specgen/generate/ports.go#L284-L302
https://github.com/containers/libpod/blob/a85e97952994a2803a17ee2a384b260115f18b43/pkg/specgen/generate/ports.go#L174-L188
https://github.com/containers/libpod/blob/a85e97952994a2803a17ee2a384b260115f18b43/libpod/container_inspect.go#L616-L619
All of this ensures that the protocol value is not null here

mheon:
Given this, it looks like it is safe, though we generally prefer to lean on the safe side to make absolutely sure we don't segfault.

zhangguanzhang:
and fix this https://github.com/containers/libpod/issues/6475

contrib/podmanimage/upstream/Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN useradd podman; yum -y update; yum -y reinstall shadow-utils; yum -y install
      cd $GOPATH/src/github.com/containernetworking/plugins; \
      ./build_linux.sh; \
      mkdir -p /usr/libexec/cni; \
-     \cp -fR bin/* /usr/libexec/cni; \
+     /usr/bin/cp -fR bin/* /usr/libexec/cni; \
```
TomSweeneyRedHat, line 58: Personally I don't think this is necessary, and I don't believe that it's at all related to the rest of this PR.  Please remove this.  If you really feel it's necessary, please change it in a separate PR so we can track the change in Git history.

zhangguanzhang: thx

jwhonce:
@zhangguanzhang Please add a signature to your commit.  Then CI will test your code.  Thank you for your contribution.

zhangguanzhang:
Signed-off-by: zhangguanzhang zhangguanzhang@qq.com

zhangguanzhang:
@jwhonce  PTAL

mheon:
/ok-to-test
/approve
LGTM
@rhatdan @vrothberg @jwhonce @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM

PR #6522, opened 2020-06-08
Add support for the unless-stopped restart policy

### LINKED ISSUE #6508: Command: "podman run --restart=" is *not* compatible w/ "docker run --restart="
/kind bug

**Description**

Installing _podman-docker_ the CLI command _podman_ should be as compatible as possible w/ docker.

There is an important docker command that is used in many dockerized systems. B/c the compatibility is not perfect in the podman command all the scripts should be changed because of podman. In practice you need an IF clause in all your scripts to check if you are running the script in a podman or a docker system, and it is not the right way in the long run.

In the Docker, you have to use _docker run --–restart=unless-stopped_ e.g for testing and sometimes in a production environment. This way you may stop a container and it **does not start after reboot**, like the _docker run --–restart=always_ does in Docker! 

But _podman run --restart=unless-stopped_ gives an error b/c this is not supported in the _libpod_.

And if you change your Docker scripts to _docker run -–restart=always_ you will lose the function you need, namely, keeping container stopped after reboot! Of course, it works in _podmad_ but not in the Docker! 

1. Install _podman-docker_ and a native _docker_

2. Run this command in the both docker and podman environments: 
$ _docker run --restart=unless-stopped_

podman fails to an error

podman should not fail

Correction: accept _--restart=unless-stopped_ using the policy _unless-stopped_ as **a synonym** to the policy _always_ in  the podman/libpod.

mheon:
We initially believed that implementing this required support for restarting containers after reboot, but this is not the case. The unless-stopped restart policy acts identically to the always restart policy except in cases related to reboot (which we do not support yet), but it does not require that support for us to implement it.

Changes themselves are quite simple, we need a new restart policy constant, we need to remove existing checks that block creation of containers when unless-stopped was used, and we need to update the manpages.

Fixes #6508

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -682,11 +682,10 @@ Restart policy will not take effect if a container is stopped via the **podman k
 
 Valid _policy_ values are:
 
-- **no**: Do not restart containers on exit;
-- **on-failure**[:*max_retries*]: Restart containers when they exit
-with a non-zero exit code, retrying indefinitely or until the optional
-*max_retries* count is hit;
-- **always**: Restart containers when they exit, regardless of status, retrying indefinitely.
+- `no`                       : Do not restart containers on exit
+- `on-failure[:max_retries]` : Restart containers when they exit with a non-zero exit code, retrying indefinitely or until the optional *max_retries* count is hit
+- `always`                   : Restart containers when they exit, regardless of status, retrying indefinitely
+- `unless-stopped`           : Identical to **always**
 
 Please note that restart will not restart containers after a system reboot.
 If this functionality is required in your environment, you can invoke Podman from a **systemd.unit**(5) file, or create an init script for whichever init system is in use.
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -676,6 +676,7 @@ Valid values are:
 - `no`                       : Do not restart containers on exit
 - `on-failure[:max_retries]` : Restart containers when they exit with a non-0 exit code, retrying indefinitely or until the optional max_retries count is hit
 - `always`                   : Restart containers when they exit, regardless of status, retrying indefinitely
+- `unless-stopped`           : Identical to **always**
 
 Please note that restart will not restart containers after a system reboot.
 If this functionality is required in your environment, you can invoke Podman from a systemd unit file, or create an init script for whichever init system is in use.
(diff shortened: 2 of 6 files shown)


rhatdan:
LGTM

rhatdan:
@HarriL PTAL
@vrothberg @giuseppe @TomSweeneyRedHat @baude @QiWang19 @jwhonce @ashley-cui PTAL

docs/source/markdown/podman-run.1.md
```diff
@@ -674,11 +674,10 @@ Restart policy will not take effect if a container is stopped via the **podman k
...
 
-- **no**: Do not restart containers on exit;
-- **on-failure**[:*max_retries*]: Restart containers when they exit
-with a non-zero exit code, retrying indefinitely or until the optional
-*max_retries* count is hit;
-- **always**: Restart containers when they exit, regardless of status, retrying indefinitely.
+- `no`                       : Do not restart containers on exit
+- `on-failure[:max_retries]` : Restart containers when they exit with a non-zero exit code, retrying indefinitely or until the optional *max_retries* count is hit
+- `always`                   : Restart containers when they exit, regardless of status, retrying indefinitely
+- `unless-stopped`           : Identical to **always**
```
TomSweeneyRedHat, line 680: the code comment below says it's not identical to always....  If unless-stopped is not supported at the moment, I'd just remove it from the man pages until it is.

rhatdan:
@mheon Would be good to get this one merged.

mheon:
Updated with a test, rebased

disaster123:
@mheon i know this is already merged but isn't this part missing:
https://github.com/mheon/libpod/blob/unless-stopped/pkg/spec/createconfig.go#L364

unless-stopped is still prevented in pkg spec

mheon:
pkg/spec is actively deprecated, so we're not really updating it anymore. I'm hoping to migrate the last users of it to the new specgen code sometime next week.

disaster123:
ah OK i fall about it when merging this one into my local v1.9 codebase

IceCodeNew:
should update the help message here:
`# podman run --help|grep restart`
>       --restart string                           Restart policy to apply when a container exits ("always"|"no"|"on-failure")

// Above command was issued on podman version 2.1.0

mheon:
Can you open a fresh issue about that so it isn't forgotten?

IceCodeNew:
Sure.

PR #6529, opened 2020-06-08
Enable IPv6 port binding

### LINKED ISSUE #5715: Add support for IPv6 port binding
/kind feature

**Description**

Would be nice to be able to bind ports to IPv6 addresses like with IPv4. In Docker works both.

`podman pod create --name=test --share net -p 2000:0000:60:1244::2:80:80/tcp` should work without an error like

Error: unable to create pod: Invalid ip address 2000:0000:60:1244::2: address 2000:0000:60:1244::2:: too many colons in address

mheon:
Two areas needed tweaking to accomplish this: port parsing and binding ports on the host.

Parsing is an obvious problem - we have to accomodate an IPv6 address enclosed by [] as well as a normal IPv4 address. It was slightly complicated by the fact that we previously just counted the number of colons in the whole port definition (a thousand curses on whoever in the IPv6 standard body decided to reuse colons for address separators), but did not end up being that bad.

Libpod also (optionally) binds ports on the host to prevent their reuse by host processes. This code was IPv4 only for TCP, and bound to both for UDP (which I'm fairly certain is not correct, and has been adjusted). This just needed protocols adjusted to read "tcp4"/"tcp6" and "udp4"/"udp6" based on what we wanted to bind to.

Fixes #5715

### DIFF
--- a/cmd/podman/common/util.go
+++ b/cmd/podman/common/util.go
@@ -71,14 +71,44 @@ func createPortBindings(ports []string) ([]specgen.PortMapping, error) {
 			return nil, errors.Errorf("invalid port format - protocol can only be specified once")
 		}
 
-		splitPort := strings.Split(splitProto[0], ":")
+		remainder := splitProto[0]
+		haveV6 := false
+
+		// Check for an IPv6 address in brackets
+		splitV6 := strings.Split(remainder, "]")
+		switch len(splitV6) {
+		case 1:
+			// Do nothing, proceed as before
+		case 2:
+			// We potentially have an IPv6 address
+			haveV6 = true
+			if !strings.HasPrefix(splitV6[0], "[") {
+				return nil, errors.Errorf("invalid port format - IPv6 addresses must be enclosed by []")
+			}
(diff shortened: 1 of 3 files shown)


cmd/podman/common/util.go
```diff
@@ -71,14 +71,44 @@ func createPortBindings(ports []string) ([]specgen.PortMapping, error) {
...
+
+		// Check for an IPv6 address in brackets
+		splitV6 := strings.Split(remainder, "]")
+		switch len(splitV6) {
+		case 1:
+			// Do nothing, proceed as before
+		case 2:
+			// We potentially have an IPv6 address
+			haveV6 = true
+			if !strings.HasPrefix(splitV6[0], "[") {
```
rhatdan, line 85: Wouldn't it be better to do this with a Regex?

mheon:
Uh-oh. All the containers/common tests are failing on ubuntu - looks like apparmor errors?

rhatdan:
It is looking for a container-default file installed for Ubuntu platform?  Needs this for apparmor

rhatdan:
@lsm5 WDYT

PR #6530, opened 2020-06-08
WIP: Enable (as in, fix) podman-remote testing

edsantiago:
podman-remote has not been tested. I suspect this PR
is going to fail catastrophically.


edsantiago:
Grrr, special_testing_rootless didn't work. This must be one of those things where I need to remake the whole VM setup. I'll look into that tomorrow.

test/e2e/cp_test.go
```diff
@@ -141,6 +141,8 @@ var _ = Describe("Podman cp", func() {
 	})
 
 	It("podman cp stdin/stdout", func() {
+		SkipIfRemote()
+		Skip("Looks like SkipIfRemote() is not working")
```
edsantiago, line 145: [For real](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/6438322283216896/html/integration_test.log.html#t--podman-cp-stdin-stdout--1). 

The code in [libpod_suite_remote_test.go](https://github.com/containers/libpod/blob/master/test/e2e/libpod_suite_remote_test.go#L22) looks OK to me. I don't understand this.

edsantiago:
@cevich, @mheon, @jwhonce, @TomSweeneyRedHat , @rhatdan, PTAL. I am sorry for not having caught this earlier.

test/e2e/cp_test.go
mheon: Um. It has to just be this one case, because we'd have a lot more errors if it wasn't working everywhere...

edsantiago: IKR. But look at the function definition for `SkipIfRemote()`:
```go
    ginkgo.Skip("This function is not enabled for remote podman")
```
Then look at that giant log file again. Do an in-page search for the string 'enabled for'.

I'm really confused.

rhatdan:
/approve
LGTM

test/e2e/cp_test.go
rhatdan: Is libpod_suite_remote_test.go  being used?

edsantiago: To my eye -- which of course is not well-versed in Go -- it has the right build tags:
```go
// +build remoteclient
```
I still don't understand why `not enabled for remote podman` doesn't appear in the test log. Many of the tests calling `SkipIfRemote()` are tagged `!remoteclient` (presumably for the v2 transition) -- but not all of them.

The one thing I don't actually see is `remoteclient` anywhere in Makefile. There's `remote`, and I recall discussion in scrum about a difference between those, but wasn't/amn't educated enough to understand the difference.

rhatdan:
We use the "remote" buildtag for podman-remote now, while we use remoteclient is for varlink.

PR #6542, opened 2020-06-09
Ensure signal validation happens first in pod kill

### LINKED ISSUE #6540: [low] podman-remote: pod kill: does not do proper signal validation
```console
$ MYSOCK=unix:/tmp/mypodmansock
$ ./bin/podman system service --timeout=0 $MYSOCK &
[1] 467524

$ ./bin/podman-remote --url $MYSOCK pod kill -s 0 nosuchpod
Error: unable to find pod "nosuchpod": no such pod
```
Expected: `Error: invalid signal: 0`

mheon:
This fixes an error in the system tests, which expect that when you try and kill a nonexistent pod with an incorrect signal, you receive an error about the signal, not the pod.

At the same time, fix a missing return statement in the bindings, which could also have caused us grief.

Fixes #6540

### DIFF
--- a/pkg/api/handlers/libpod/pods.go
+++ b/pkg/api/handlers/libpod/pods.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/libpod/pkg/util"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 func PodCreate(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
@edsantiago got time for a quick eyeball?

edsantiago:
@TomSweeneyRedHat I already gave it a once-over; did not feel comfortable LGTM'ing because I don't know `ParseSignal`, can't easily chase down the `syscall.Signal` type, so I don't know the type of `sig` in the debug statement, and could not confirm that `%d` was appropriate. Apart from that, everything looks perfect to me.

PR #6545, opened 2020-06-09
[WIP] Implement pod-network-reload

mheon:
This adds a new command, 'podman network reload', to reload the networks of existing containers, forcing recreation of firewall rules after e.g. `firewall-cmd --reload` wipes them out.

Under the hood, this works by calling CNI to tear down the existing network, then recreate it using identical settings. We request that CNI preserve the old IP and MAC address in most cases (where the container only had 1 IP/MAC), but there will be some downtime inherent to the teardown/bring-up approach. The architecture of CNI doesn't really make doing this without downtime easy (or maybe even possible...).

At present, this only works for root Podman, and only locally. I don't think there is much of a point to adding remote support (this is very much a local debugging command), but I think adding rootless support (to kill/recreate slirp4netns) could be valuable.

Still To-Do:
- Bash Completions
- Tests
- CNI is not honoring our requested IP and MAC address, need to figure that out.

### DIFF
--- a/cmd/podman/networks/reload.go
+++ b/cmd/podman/networks/reload.go
@@ -1,0 +1,65 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/containers/libpod/cmd/podman/parse"
+	"github.com/containers/libpod/cmd/podman/registry"
+	"github.com/containers/libpod/cmd/podman/utils"
+	"github.com/containers/libpod/pkg/domain/entities"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+var (
+	networkReloadDescription = `reload container networks, recreating firewall rules`
+	networkReloadCommand     = &cobra.Command{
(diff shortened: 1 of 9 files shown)


mheon:
Would like comments on the name - this is a `podman network` command that acts on containers, but `podman container reload` doesn't seem descriptive and `podman container reload-net` is a little long. Thoughts?
@rhatdan @baude @vrothberg @TomSweeneyRedHat

vrothberg:
> Would like comments on the name - this is a `podman network` command that acts on containers, but `podman container reload` doesn't seem descriptive and `podman container reload-net` is a little long. Thoughts?
> @rhatdan @baude @vrothberg @TomSweeneyRedHat

I like `podman network reload` :+1:

Luap99:
@mheon Any news on this? I believe I can figure this one out so I would like to finish it.

mheon:
If you'd like to take this over, feel free, I'm otherwise occupied working on volume plugins.

Luap99:
Replaced by #8571

Issue #5485, opened 2020-03-12
Problems with podman generate systemd


larsks:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The unit files produced by `podman generate systemd --new ...` are problematic because they will fail to start the service if the system crashes.  This happens because the container is only removed as part of `ExecStopPost`, and during a crash (/power failure/etc) that never runs.  The result is that a subsequent attempt to start the service will fail with `Error: error creating container storage: the container name "your_container_name" is already in use by ...`.

You can't simply drop the `--new`, because then the resulting unit file won't actually create the container if it doesn't exist.

What you really need to make this robust is one of two things:

- Tie the lifetime of the container to the lifetime of the `podman` command. When the `podman` exits, the container is gone. This would make a containerized service behave like a normal un-containerized service, which would work well with a process manager like `systemd`.

- Allow replacing an existing container with a new container.  E.g., introduce a `--replace` flag to `podman run` so that `podman run --replace --name foo ...` would permit create a new container named `foo` *even if one already exists*.

mheon:
@vrothberg PTAL

vrothberg:
Thanks for reaching out, @larskarlitski!

I like the idea of a `--replace` flag as it allows to generalize this problem beyond systemd.

@mheon @giuseppe @rhatdan WDYT?

giuseppe:
could we add a `ExecStartPre` that does a `rm -f`?

vrothberg:
> could we add a `ExecStartPre` that does a `rm -f`?

That will only work for containers that have been named but won't work for unnamed containers.

mheon:
No issues here, sounds reasonable to do

rhatdan:
@larsks @vrothberg Does this issue still exist?

vrothberg:
Yes. The idea is to add `--replace` to run/create which will remove an existing container in case of a name conflict. There are a couple of issues I want to tackle once the v2 work is closer to the finish line.

rhatdan:
@vrothberg Any progress on this?

vrothberg:
#6553 will fix it :+1:

PR #6553, opened 2020-06-10
--replace for containers and pods

vrothberg:
Add a `--replace` flag to `podman {create,run}` and `podman pod create` to allow for replacing and removing existing named containers and pods.

Use that flag in `generate systemd` to make the generated units more robust and remove named containers or pods even after a system crash.

Fixes: #5485


vrothberg:
@mheon, I wired the flag into `pod create` and into the systemd units as well. The big --new pod PR was merged, so I could complete --replace.

vrothberg:
@giuseppe @rhatdan @TomSweeneyRedHat PTAL

rhatdan:
LGTM

PR #6562, opened 2020-06-10
Fix saving in oci format

rhatdan:
Fixes: https://github.com/containers/libpod/issues/6544

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -21,6 +21,7 @@ import (
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	ociarchive "github.com/containers/image/v5/oci/archive"
+	"github.com/containers/image/v5/oci/layout"
 	is "github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/tarball"
 	"github.com/containers/image/v5/transports"
@@ -1475,9 +1476,10 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
 			return errors.Wrapf(err, "error getting OCI archive ImageReference for (%q, %q)", output, destImageName)
 		}
 	case "oci-dir":
-		destRef, err = directory.NewReference(output)
+		destImageName := imageNameForSaveDestination(i, source)
+		destRef, err = layout.NewReference(output, destImageName)
 		if err != nil {
-			return errors.Wrapf(err, "error getting directory ImageReference for %q", output)
+			return errors.Wrapf(err, "error saving the oci directory ImageReference for (%q, %q)", output, destImageName)
 		}
 		manifestType = imgspecv1.MediaTypeImageManifest
 	case "docker-dir":


libpod/image/image.go
```diff
@@ -1475,7 +1476,8 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
 			return errors.Wrapf(err, "error getting OCI archive ImageReference for (%q, %q)", output, destImageName)
 		}
 	case "oci-dir":
-		destRef, err = directory.NewReference(output)
+		destImageName := imageNameForSaveDestination(i, source)
+		destRef, err = layout.NewReference(output, destImageName)
```
mtrmac, line 1480: (Preserving the `// destImageName may be ""` comment would be consistent, and not lead the readers to wonder why it can be so for one and not the other.)

libpod/image/image.go
```diff
@@ -1475,7 +1476,8 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
...
 		if err != nil {
 			return errors.Wrapf(err, "error getting directory ImageReference for %q", output)
 		}
```
mtrmac, line 1483: BTW below, calling the format `docker-dir` is somewhat misleading: It’s a c/image-invented format. I don’t care about the misattribution, just noting that some people are interested to avoiding certain words :) , so maybe it might be worth adding an alias.

test/e2e/save_test.go
```diff
@@ -75,7 +75,7 @@ var _ = Describe("Podman save", func() {
 		Expect(save).To(ExitWithError())
 	})
 
-	It("podman save to directory with oci format", func() {
+	It("podman save and load to directory with oci format", func() {
```
mtrmac, line 78: Looking at the test failures:
- At least going by descriptions without digging into PR history, it’s not clear that all of them wanted to use the OCI directory format; maybe they intended to use `docker-dir`.
- `podman load` does not implement `oci-archive` (https://github.com/containers/libpod/blob/091b9b1163ecbc3e64a9152b5e57e969b4475c2a/libpod/runtime_img.go#L257 ), `podman pull oci-archive:…` might work in the meantime.

mtrmac: It’s not entirely clear that `podman load` should try all of these formats — but probably so (and leaving `podman pull` for the remote pulls and the explicit transport:… uses). It might be worth checking whether the input path is a directory instead of trying all four formats every time.

libpod/image/image.go
rhatdan: Fixed

libpod/image/image.go
rhatdan: This one we will need to let go.

test/e2e/save_test.go
rhatdan: The error I am seeing is a lack of manifest.json file?
```
# /bin/podman save --format docker-dir -o /tmp/docker-dir alpine
# ./bin/podman save --format oci-dir -o /tmp/oci-dir alpine
# ls /tmp/docker-dir
50644c29ef5a27c9a40c393a73ece2479de78325cae7d762ef3cdc19bf42dd0a  blobs       manifest.json  version
a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e  index.json  oci-layout
# ls /tmp/oci-dir
blobs  index.json  oci-layout
```
Should there be a manifest.json file in the oci-dir?

rhatdan: @mtrmac ^^

libpod/image/image.go
mtrmac: I don’t see any change

libpod/image/image.go
```diff
@@ -1475,9 +1476,10 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
 			return errors.Wrapf(err, "error getting OCI archive ImageReference for (%q, %q)", output, destImageName)
 		}
 	case "oci-dir":
-		destRef, err = directory.NewReference(output)
+		destImageName := imageNameForSaveDestination(i, source)
+		destRef, err = layout.NewReference(output, destImageName)
 		if err != nil {
-			return errors.Wrapf(err, "error getting directory ImageReference for %q", output)
+			return errors.Wrapf(err, "error saving the oci directory ImageReference for (%q, %q)", output, destImageName)
```
mtrmac, line 1482: This is not quite right — actual saving happens in `PushImageToReference` below.
```suggestion
			return errors.Wrapf(err, "error getting OCI directory ImageReference for (%q, %q)", output, destImageName)
```

test/e2e/save_test.go
mtrmac: There shouldn’t be a `manifest.json`. The missing `manifest.json` error is reported by a `dir:` source — as mentioned above, `podman load` tries to use `dir:` but not `oci:` (I incorrectly said `oci-archive:`, which _is_ tried).

rhatdan:
@QiWang19 Could you take this one over.  The issue here is on the load side not on the save.

Issue #6535, opened 2020-06-09
Missing a comprehensive guide to the new API


TomasTomecek:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

I wanted to check out the new API and play with it a bit but I am actually missing a guide (or a blog post) how to even reach the API.

[This documentation](http://docs.podman.io/en/latest/_static/api.html) is fantastic, but it's missing the step 0 - where is the HTTP api being served and how to reach it.

jgallucci32:
@TomasTomecek The new Podman API (APIv2) is based off the Docker API referenced here: https://docs.docker.com/engine/api/

In my own personal testing the API has its limitations, but it is actively under development and supports many of the same calls you would expect in Docker. As far as I can tell it only works with sockets, but I imagine HTTP will be implemented down the road.

The use of the experimental service is explained in the help command `podman system service --help`
```
Run API service

Description:
  Run an API service

Enable a listening service for API access to Podman commands.


Usage:
  podman system service [flags] [URI]

Examples:
  podman system service --time=0 unix:///tmp/podman.sock

Flags:
  -t, --time int   Time until the service session expires in seconds.  Use 0 to disable the timeout (default 5)
```

To start the service run something like this to keep it running for a long period of time
```
podman system service -t 5000 &
```
You can use cURL on the socket with the same request you would for the Docker API
```
curl --unix-socket /run/podman/podman.sock http://docker/v1.24/info
```

NOTE: If you install the package `podman-docker` it will create a symbolic link for `/var/run/docker.sock` which links to `/run/podman/podman.sock`

TomasTomecek:
@jgallucci32 thank you, this is exactly the info I was expecting to read [here](http://docs.podman.io/en/latest/_static/api.html).

PR #6564, opened 2020-06-10
[CI:DOCS] Add quick start directions to APIv2 Dock

TomSweeneyRedHat:
Adds some quick start up directions to the top of the
API v2 documentation and a few examples.
This strongly leverages comments from @jgallucci32 in #6535.

Fixes: #6535

### DIFF
--- a/pkg/api/server/docs.go
+++ b/pkg/api/server/docs.go
@@ -4,6 +4,24 @@
 // only as experimental as this point.  The endpoints, parameters, inputs, and
 // return values can all change.
 //
+// To start the service and keep it running for 5,000 seconds (-t 0 runs forever):
+//    podman system service -t 5000 &
+//
+// You can then use cURL on the socket using requests documented below.
+//
+// NOTE: if you install the package podman-docker, it will create a symbolic
+// link for /var/run/docker.sock to /run/podman/podman.sock
+//
+// See podman-service(1) for more information.
+//
+//  Quick Examples:
+//   'podman info'
+//      curl --unix-socket /run/podman/podman.sock http://d/v1.0.0/libpod/info
+//   'podman pull quay.io/containers/podman'
+//      curl -XPOST --unix-socket /run/podman/podman.sock -v 'http://d/v1.0.0/images/create?fromImage=quay.io%2Fcontainers%2Fpodman'
+//   'podman list images'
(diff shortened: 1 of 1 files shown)


pkg/api/server/docs.go
```diff
@@ -4,6 +4,24 @@
 // only as experimental as this point.  The endpoints, parameters, inputs, and
 // return values can all change.
 //
+// To start the service and keep it running for 5,000 seconds:
+//    podman system service -t 5000 &
```
mheon, line 8: Should explicitly mention 0 timeout = never stops

rhatdan:
LGTM, with @mheon Change.

pkg/api/server/docs.go
agebhar1: Maybe also add an example for 'HTTP', e.g.
```
$ podman system service --timeout 0 tcp:127.0.0.1:8080
```

pkg/api/server/docs.go
```diff
@@ -4,6 +4,24 @@
...
+// You can then use cURL on the socket using requests documented below.
+//
+// NOTE: if you install the package podman-docker, it will create a symbolic
+// link for /var/run/docker.sock to /run/podman/podman.sock
+//
+// See podman-service(1) for more information.
+//
+//  Quick Examples:
+//   'podman info'
+//      curl --unix-socket /run/podman/podman.sock http://d/v1.0.0/libpod/info
```
agebhar1, line 19: For a TCP connection the example might be:
```
curl --silent localhost:8080/v1.0.0/libpod/info 
```

rhatdan: I would really recommend that people not use tcp, since there is NO Security wrapper around this.  It is basically opening a welcome to root user without a password if you listen on a network port and run as root.  Running as non root, gives away access to your homedir.

agebhar1: Thanks for the security hint, you are totally right.

agebhar1: If I start podman system service as user from the example above (w/ or w/o timeout)
```
$ podman system service 
```

TomSweeneyRedHat:
Can this be merged now?  I'm happy to circle back to handle the outcome of the discussions with @agebhar1 , but in the short term would like to see how this looks once published.   I fear once pushed I may need to touch up formatting issues.

PR #6566, opened 2020-06-11
replace error-prone non-ASCII quotes in template config

parmentelat:
while browsing the config file template that ships with fedora32, I came across
odd-looking non-ASCII quotes and I figured I'd propose to clean this up


rhatdan:
You need to open this PR against github.com/containers/common
You are not allowed to modify code in the vendor directory.  This code comes from other repositories,

rhatdan:
Thanks @parmentelat I opened this PR in https://github.com/containers/common/pull/178

PR #6568, opened 2020-06-11
libpod: fix check for slirp4netns netns

giuseppe:
fix the check for c.state.NetNS == nil.  Its value is changed in the
first code block, so the condition is always true in the second one
and we end up running slirp4netns twice.

Closes: https://github.com/containers/libpod/issues/6538


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
These little subtle fixes always fascinate me.

PR #6576, opened 2020-06-11
Fix missing code during in_podman build

cevich:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,7 @@ ENV SCRIPT_BASE=./contrib/cirrus
 ENV PACKER_BASE=$SCRIPT_BASE/packer
 
 # Only add minimal tooling necessary to complete setup.
-ADD /$SCRIPT_BASE $GOSRC/$SCRIPT_BASE
-ADD /hack/install_catatonit.sh $GOSRC/hack/
-ADD /cni/*.conflist $GOSRC/cni/
-ADD /test/*.json $GOSRC/test/
-ADD /test/*.conf $GOSRC/test/
+ADD / $GOSRC
 WORKDIR $GOSRC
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

cevich:
This will fix the `in_podman` builds that have been failing for the last few days.

rhatdan:
/approve
@mheon @baude @giuseppe @vrothberg @TomSweeneyRedHat @jwhonce @QiWang19 @ashley-cui PTAL

Dockerfile
```diff
@@ -10,11 +10,7 @@ ENV SCRIPT_BASE=./contrib/cirrus
 ENV PACKER_BASE=$SCRIPT_BASE/packer
 
 # Only add minimal tooling necessary to complete setup.
```
markstos, line 12: @cevich This comment no longer seems accurate after the change. Is adding everything to the container from the libpod directory really the "minimal tooling to necessary complete setup"? On my checkout of `libpod` I have 126 MB in the current directory and running `make test` visibly stalled at this step for so long that I stopped to investigate.

For contrast, I ran this on the old "ADD" commands:

      du -hc ./contrib/cirrus/ ./hack/install_catatonit.sh cni/*.conflist test/*.json test/*.conf

That totaled 304K.

Did the minimal files needed for tooling really increase from 304K to 126M?

cevich: Sorry I'm late.  You're right, it's not an accurate comment.  At one point I was much more selective in the Dockerfile about only adding what was needed, but...kept missing stuff.  So I just tossed it all in and forgot about it :grin: I'll open another PR to fix this.

cevich: #6904

PR #6588, opened 2020-06-12
Reassemble filters on the server side

rhatdan:
--filter label=foo=bar, was been translated on the server side to
--filter label=foo --filter=bar

This PR fixes this back to what the user specified.

### DIFF
--- a/pkg/api/handlers/utils/images.go
+++ b/pkg/api/handlers/utils/images.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/storage"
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago @jwhonce PTAL
@jwhonce I am not sure if this should be fixed here or to fix the protocol.  IE Should filers be a 
map[string]string 
rather then
map[string][]string

rhatdan:
@baude PTAL

rhatdan:
Probably fixes https://github.com/containers/libpod/issues/6387

rhatdan:
Fixes: https://github.com/containers/libpod/issues/6580

edsantiago:
Will look on Monday. Thank you, Dan, but I hope you're able to recharge today.

rhatdan:
@edsantiago I am always charged, although my wife is yelling at me now, so I am heading to the beach. :^) And you should not be here either.

vrothberg:
> @edsantiago I am always charged, although my wife is yelling at me now, so I am heading to the beach. :^) And you should not be here either.

She was right to do so ;-) Hope you enjoyed surfing.

edsantiago:
LGTM. Given an image with `LABEL version="1.0"`:
```console
$ ./bin/podman-remote --url $MYSOCK images -f label=foo=bar
REPOSITORY  TAG     IMAGE ID  CREATED  SIZE
$ ./bin/podman-remote --url $MYSOCK images -f label=version=1..
REPOSITORY  TAG     IMAGE ID  CREATED  SIZE
$ ./bin/podman-remote --url $MYSOCK images -f label=version=1.0
REPOSITORY      TAG     IMAGE ID      CREATED         SIZE
localhost/test  latest  c46c4c460591  34 seconds ago  5.85 MB
```

/lgtm

PR #6605, opened 2020-06-15
APIv2 tests: Add some tests for podman pods

ypu:
Add some tests for podman pods subcommand:
  restart
  rm
  start
  stas
  stop
  top
  unpause

### DIFF
--- a/test/apiv2/40-pods.at
+++ b/test/apiv2/40-pods.at
@@ -25,6 +25,16 @@ t POST "libpod/pods/create (dup pod)" name=foo 409 \
 
 #t POST libpod/pods/create a=b 400 .cause='bad parameter'  # FIXME: unimplemented
 
+t POST libpod/pods/foo/start '' 200 \
+  .Errs=null \
+  .Id=$pod_id
+
+t POST libpod/pods/foo/start '' 304 \
+
+t POST libpod/pods/fakename/start '' 404 \
+  .cause="no such pod" \
+  .message="no pod with name or ID fakename found: no such pod"
+
 if root || have_cgroupsv2; then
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
/approve

rhatdan:
@edsantiago PTAL

test/apiv2/40-pods.at
```diff
@@ -35,12 +45,72 @@ else
...
+podman run -d --pod bar busybox sleep 999
+
+t POST libpod/pods/bar/stop?t=1 '' 200 \
+  .Errs=null \
+  .Id=$pod_bar_id
+
+t POST libpod/pods/bar/start '' 200
+
+t GET libpod/pods/stats?all=true 200
+is `awk -F"{" '{print NF-1}' <<< "$output"` 3 "GET libpod/pods/stats?all=true"
```
edsantiago, line 94: This and line 97 are difficult to understand; they're also fragile, because it's possible that a future version of `stats` could return nested objects. Did you know you can use `jq` itself to get an array length? I would encourage you to replace this (and line 97) with:
```bash
is $(jq '. | length' <<<"$output") 3 "stats?all=true: number of records found"
```
Note also the change in the message: there is no need to duplicate the entire `GET etc etc', because that is already shown in the previous line. This form removes unnecessary text, and also gives information on what is actually being tested (number of matches).

TomSweeneyRedHat:
LGTM once @edsantiago's concerns are addressed

test/apiv2/40-pods.at
ypu: Thanks @edsantiago . Will update it.

PR #6610, opened 2020-06-15
Add deprecated message to varlink command

jwhonce:


cmd/podman/system/varlink.go
```diff
@@ -16,11 +16,12 @@ var (
...
-		Args:  cobra.MinimumNArgs(1),
-		Short: "Run varlink interface",
-		Long:  varlinkDescription,
-		RunE:  varlinkE,
+		Use:        "varlink [flags] [URI]",
+		Args:       cobra.MinimumNArgs(1),
+		Short:      "Run varlink interface",
+		Long:       varlinkDescription,
+		RunE:       varlinkE,
+		Deprecated: "Please see `podman system service' for RESTful APIs",
```
rhatdan, line 24: Can you just make this hidden?

TomSweeneyRedHat, line 24:
```suggestion
		Deprecated: "Please see 'podman system service' for RESTful APIs",
```
different single quotes, I removed the back tick one in front, you could replace the ending one if you prefer as long as they match.

mheon:
Agree with the deprecation message, but also think we should mark this as hidden to try and force people to `podman system service --varlink`

docs/source/markdown/podman-remote.1.md
```diff
@@ -17,7 +17,7 @@ Podman uses Buildah(1) internally to create container images. Both tools share i
 (not container) storage, hence each can use or manipulate images (but not containers)
 created by the other.
 
-Podman-remote provides a local client interacting with a Podman backend node through a varlink ssh connection. In this context, a Podman node is a Linux system with Podman installed on it and the varlink service activated. Credentials for this session can be passed in using flags, environment variables, or in `podman-remote.conf`
+Podman-remote provides a local client interacting with a Podman backend node through a RESTful API tunneled through a ssh connection. In this context, a Podman node is a Linux system with Podman installed on it and the API service activated. Credentials for this session can be passed in using flags, environment variables, or in `podman-remote.conf`
```
TomSweeneyRedHat, line 20:
```suggestion
Podman-remote provides a local client interacting with a Podman backend node through a RESTful API tunneled through a SSL/TLS connection. In this context, a Podman node is a Linux system with Podman installed on it and the API service activated. Credentials for this session can be passed in using flags, environment variables, or in `podman-remote.conf`
```

rhatdan: No, we want them tunning through ssh not directly via SSL/TLS

rhatdan:
LGTM

docs/source/markdown/podman-remote.1.md
TomSweeneyRedHat: Thought we were using https?

rhatdan: Nope.  ssh is default.

PR #6619, opened 2020-06-16
[CI:DOCS] Revert "Change Varlink systemd unit to use `system service`"

mheon:
This reverts commit 1bc992bfc3a983b4d9ab53f778a545d83bcde94d.

We originally thought `podman varlink` was entirely removed, but that was not true. We originally thought that `podman system service --varlink` worked the same as `podman varlink` but that was also not true. `system service` is broken when used under systemd units, and `podman varlink` still exists and works. Revert the change to `podman system service` to fix socket-activated Varlink under systemd.

### DIFF
--- a/contrib/varlink/io.podman.service
+++ b/contrib/varlink/io.podman.service
@@ -6,7 +6,7 @@ Documentation=man:podman-varlink(1)
 
 [Service]
 Type=simple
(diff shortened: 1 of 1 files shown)


mheon:
@rhatdan @baude @vrothberg @jwhonce @giuseppe @TomSweeneyRedHat PTAL, need this for a new RC

PR #6621, opened 2020-06-16
search: allow wildcards

vrothberg:
Allow trailing wildcards in the search term.  Note that not all
registries support wildcards and it may only work with v1 registries.
Hence, we just strip `*` before parsing the domain of the input.

Fixes: bugzilla.redhat.com/show_bug.cgi?id=1846629

@mtrmac @rhatdan @TomSweeneyRedHat  @QiWang19 PTAL

### DIFF
--- a/libpod/image/search.go
+++ b/libpod/image/search.go
@@ -64,13 +64,16 @@ type SearchFilter struct {
 // SearchImages searches images based on term and the specified SearchOptions
 // in all registries.
 func SearchImages(term string, options SearchOptions) ([]SearchResult, error) {
-	// Check if search term has a registry in it
-	registry, err := sysreg.GetRegistry(term)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error getting registry from %q", term)
-	}
-	if registry != "" {
-		term = term[len(registry)+1:]
+	registry := ""
+
+	// Try to extract a registry from the specified search term.  We
+	// consider everything before the first slash to be the registry.  Note
+	// that we cannot use the reference parser from the containers/image
+	// library as the search term may container arbitrary input such as
+	// wildcards.  See bugzilla.redhat.com/show_bug.cgi?id=1846629.
+	if spl := strings.SplitN(term, "/", 2); len(spl) > 1 {
+		registry = spl[0]
+		term = spl[1]
 	}
 
 	registries, err := getRegistries(registry)
--- a/pkg/registries/registries.go
+++ b/pkg/registries/registries.go
@@ -3,12 +3,10 @@ package registries
 import (
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/libpod/pkg/rootless"
-	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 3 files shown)


vrothberg:
@baude @mheon this would be a nice to have for v2, assuming this PR is the right approach.

TomSweeneyRedHat:
Some of the tests don't look happy.

vrothberg:
> parse it manually, using whatever the semantics is 

That is the crux. I don't see a way to support arbitrary input. Especially with the search-registries. We need to extract a the domain/registry from the input, but if we support arbitrary input we just can't do that.

If we limit wildcard-support to trailing ones, we don't can keep the current parsing logic and support most of the use cases (at least all in the BZ).

vrothberg:
Marked as WIP until we figured out if trailing-only would meet the customer demands.

mtrmac:
> > parse it manually, using whatever the semantics is
> 
> That is the crux. I don't see a way to support arbitrary input. Especially with the search-registries. We need to extract a the domain/registry from the input, but if we support arbitrary input we just can't do that.

I didn’t mean “whatever the semantics of the remote registry wildcards is”, just “whatever the semantics of the argument to `podman search` is”.

Looking at https://github.com/containers/libpod/blob/88ee5d4f5f8ba6afabbb921db7b0ba7e92d59148/libpod/image/search.go#L71-L78 , we parse the input string into (`registry`, `term`), and only `term` is passed to the registry’s API (and thus has unknown semantics).

The semantics of the split into registry+term is entirely within Podman’s control. Isn’t what we want basically a “registry is the part before the first `/` if any”? I suppose we should have unit tests for all the affected cases, but I don’t think it’s much more than that.

vrothberg:
Brilliant idea to consider a registry before `/`, @mtrmac!

mtrmac:
Looking into this a bit further, the tests of `reference.Parse` do suggest that `foo/bar` is parsed as “domain == `foo`”.

OTOH there is a difference in error handling/format enforcement, for inputs like 'what#is/this\even'.

vrothberg:
> Looking into this a bit further, the tests of `reference.Parse` do suggest that `foo/bar` is parsed as “domain == `foo`”.
> 
> OTOH there is a difference in error handling/format enforcement, for inputs like 'what#is/this\even'.

The podman-search behaviour is quite curious in some cases already:
```
~ $ podman search registry.redhat.io
INDEX       NAME                                               DESCRIPTION                                       STARS   OFFICIAL   AUTOMATED
docker.io   docker.io/knqyf263/registry.redhat.io-ubi7                                                           0                  
docker.io   docker.io/abdennour/rhel                           RHEL based on registry.redhat.io/ubi8/ubi-in...   0                  
docker.io   docker.io/chengkuan/amq-streams-kafka-connect-23   Kafka MongoDB Connect based on registry.redh...   0                  
```

It really is quite challenging for parsing. Docker has such a nice life as it only normalizes to docker.io.

libpod/image/search.go
```diff
@@ -64,11 +64,11 @@ type SearchFilter struct {
 // SearchImages searches images based on term and the specified SearchOptions
 // in all registries.
 func SearchImages(term string, options SearchOptions) ([]SearchResult, error) {
-	// Check if search term has a registry in it
-	registry, err := sysreg.GetRegistry(term)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error getting registry from %q", term)
+	registry := ""
```
vrothberg, line 67: @mtrmac, does that reflect what you had in mind?

mtrmac: Yes, exactly (but I haven’t done the legwork to check that there are enough tests for the relevant input formats to know I’m not missing a corner case).

libpod/image/search.go
```diff
@@ -64,11 +64,11 @@ type SearchFilter struct {
...
+	if spl := strings.SplitN(term, "/", 2); len(spl) > 1 {
+		registry = spl[0]
 	}
+
 	if registry != "" {
 		term = term[len(registry)+1:]
 	}
```
mtrmac, line 74: (Absolutely non-blocking: It feel be a tiny bit cleaner to me if this used `spl[1]` in the condition above instead of an independent parsing code — but AFAICS it works exactly the same.)

vrothberg:
Rebased on master to unblock CI

PR #6622, opened 2020-06-16
Fix --init and --init-path

goochjj:
This passes init and init-path through to specgen, and I've tested it works after fixing the command injection.

### DIFF
--- a/pkg/specgen/generate/storage.go
+++ b/pkg/specgen/generate/storage.go
@@ -314,8 +314,8 @@ func addContainerInitBinary(s *specgen.SpecGenerator, path string) (spec.Mount, 
 	if !s.PidNS.IsPrivate() {
 		return mount, fmt.Errorf("cannot add init binary as PID 1 (PID namespace isn't private)")
 	}
-	if s.Systemd == "true" || s.Systemd == "always" {
-		return mount, fmt.Errorf("cannot use container-init binary with systemd")
+	if s.Systemd == "always" {
+		return mount, fmt.Errorf("cannot use container-init binary with systemd=always")
 	}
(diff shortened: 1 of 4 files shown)


mheon:
Changes LGTM
/approve
/ok-to-test

mheon:
Can you add a signoff? `git commit --amend -s --no-edit` should add it.

goochjj:
Signoff added, added appropriate unit tests, fixed systemd=true behavior.

mheon:
Gitvalidation apparently dislikes that the first line of your commit message is >90 characters

pkg/specgen/generate/storage.go
```diff
@@ -314,8 +314,8 @@ func addContainerInitBinary(s *specgen.SpecGenerator, path string) (spec.Mount,
 	if !s.PidNS.IsPrivate() {
 		return mount, fmt.Errorf("cannot add init binary as PID 1 (PID namespace isn't private)")
 	}
-	if s.Systemd == "true" || s.Systemd == "always" {
-		return mount, fmt.Errorf("cannot use container-init binary with systemd")
+	if s.Systemd == "always" {
```
TomSweeneyRedHat, line 317: Is this change right?  Can we used the container-init binary with systemd?

mheon: `--systemd=true` is the default from the CLI, which implies that we use systemd if requested; I think we need to ignore it when both it and `--init` are given, in favor of the more-specific `--init`. Meanwhile, `--systemd=always` forces systemd on, and that sounds like it should conflict.

mheon:
LGTM

@rhatdan @vrothberg @giuseppe @baude @TomSweeneyRedHat PTAL, let's get this landed

PR #6625, opened 2020-06-16
Add --preservefds to podman run

QiWang19:
Add --preservefds to podman run. close https://github.com/containers/libpod/issues/6458

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -61,10 +61,12 @@ func runFlags(flags *pflag.FlagSet) {
 	flags.SetNormalizeFunc(common.AliasFlags)
 	flags.BoolVar(&runOpts.SigProxy, "sig-proxy", true, "Proxy received signals to the process")
 	flags.BoolVar(&runRmi, "rmi", false, "Remove container image unless used by other containers")
+	flags.UintVar(&runOpts.PreserveFDs, "preserve-fds", 0, "Pass a number of additional file descriptors into the container")
 	if registry.IsRemote() {
 		_ = flags.MarkHidden("authfile")
 		_ = flags.MarkHidden("env-host")
 		_ = flags.MarkHidden("http-proxy")
+		_ = flags.MarkHidden("preserve-fds")
 	}
 	// Not sure we want these exposed yet.  If we do, they need to be documented in man pages
 	_ = flags.MarkHidden("override-arch")
@@ -163,6 +165,7 @@ func run(cmd *cobra.Command, args []string) error {
 	}
 	runOpts.Detach = cliVals.Detach
 	runOpts.DetachKeys = cliVals.DetachKeys
+	cliVals.PreserveFDs = runOpts.PreserveFDs
 	s := specgen.NewSpecGenerator(args[0], cliVals.RootFS)
 	if err := common.FillOutSpecGen(s, &cliVals, args); err != nil {
 		return err
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -609,6 +609,10 @@ If a container is run with a pod, and the pod has an infra-container, the infra-
 
 Run container in an existing pod and read the pod's ID from the specified file.  If a container is run within a pod, and the pod has an infra-container, the infra-container will be started before the container is.
 
+**--preserve-fds**=*N*
+
+Pass down to the process N additional file descriptors (in addition to 0, 1, 2).  The total FDs will be 3+N.
+
 **--privileged**=**true**|**false**
 
 Give extended privileges to this container. The default is **false**.
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -418,6 +418,11 @@ type ContainerConfig struct {
 
 	// HealthCheckConfig has the health check command and related timings
 	HealthCheckConfig *manifest.Schema2HealthConfig `json:"healthcheck"`
+
+	// PreserveFDs is a number of additional file descriptors (in addition
+	// to 0, 1, 2) that will be passed to the executed process. The total FDs
+	// passed will be 3 + PreserveFDs.
+	PreserveFDs uint `json:"preserveFds,omitempty"`
 }
 
 // ContainerNamedVolume is a named volume that will be mounted into the
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -1369,6 +1369,18 @@ func WithHealthCheck(healthCheck *manifest.Schema2HealthConfig) CtrCreateOption 
 	}
 }
 
+// WithPreserveFDs forwards from the process running Libpod into the container
+// the given number of extra FDs (starting after the standard streams) to the created container
+func WithPreserveFDs(fd uint) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+		ctr.config.PreserveFDs = fd
+		return nil
+	}
+}
+
 // WithCreateCommand adds the full command plus arguments of the current
 // process to the container config.
 func WithCreateCommand() CtrCreateOption {
--- a/pkg/specgen/specgen.go
+++ b/pkg/specgen/specgen.go
@@ -130,6 +130,11 @@ type ContainerBasicConfig struct {
 	// Remove indicates if the container should be removed once it has been started
 	// and exits
 	Remove bool `json:"remove"`
+	// PreserveFDs is a number of additional file descriptors (in addition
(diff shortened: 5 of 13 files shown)


rhatdan:
LGTM
/approve

edsantiago:
Please add a system test, possibly in `test/system/030-run.bats`, using something like what I messaged you earlier today. The e2e test here is not very useful: all it does is confirm that the option does not cause a syntax error. It does not test any actual functionality.

cmd/podman/containers/run.go
```diff
@@ -61,10 +61,12 @@ func runFlags(flags *pflag.FlagSet) {
 	flags.SetNormalizeFunc(common.AliasFlags)
 	flags.BoolVar(&runOpts.SigProxy, "sig-proxy", true, "Proxy received signals to the process")
 	flags.BoolVar(&runRmi, "rmi", false, "Remove container image unless used by other containers")
+	flags.UintVar(&runOpts.PreserveFDs, "preserve-fds", 0, "Pass N additional file descriptors to the container")
```
mheon, line 64: Recommend rewording to "Pass a number of additional file descriptors into the container"

libpod/container.go
```diff
@@ -414,6 +414,11 @@ type ContainerConfig struct {
 
 	// HealthCheckConfig has the health check command and related timings
 	HealthCheckConfig *manifest.Schema2HealthConfig `json:"healthcheck"`
+
+	// PreserveFDs is a number of additional file descriptors (in addition
+	// to 0, 1, 2) that will be passed to the executed process. The total FDs
+	// passed will be 3 + PreserveFDs.
+	PreserveFDs uint `json:"-"`
```
mheon, line 421: Can you add a comment that this is not in the JSON because it is not supported for remote connections?

mheon: Oh, wait, nevermind - this needs to actually have a JSON tag here, not `"-"` - that should only be in Specgen. This is for the DB, and we do want to save this to the container in the DB.

libpod/options.go
```diff
@@ -1369,6 +1369,17 @@ func WithHealthCheck(healthCheck *manifest.Schema2HealthConfig) CtrCreateOption
 	}
 }
 
+// WithPreserveFDs adds the file descriptors to the container config
```
mheon, line 1372: Suggest "WithPreserveFDs forwards from the process running Libpod into the container the given number of extra FDs (starting after the standard streams) to the created container"

pkg/specgen/specgen.go
```diff
@@ -130,6 +130,10 @@ type ContainerBasicConfig struct {
 	// Remove indicates if the container should be removed once it has been started
 	// and exits
 	Remove bool `json:"remove"`
+	// PreserveFDs is a number of additional file descriptors (in addition
+	// to 0, 1, 2) that will be passed to the executed process. The total FDs
+	// passed will be 3 + PreserveFDs.
+	PreserveFDs uint `json:"-"`
```
mheon, line 136: We do need a comment here that this is JSON:"-" because it's not available remote.

mheon: Unlike container.go this one should stay "-" as well

docs/source/markdown/podman-run.1.md
```diff
@@ -609,6 +609,10 @@ If a container is run with a pod, and the pod has an infra-container, the infra-
 
 Run container in an existing pod and read the pod's ID from the specified file.  If a container is run within a pod, and the pod has an infra-container, the infra-container will be started before the container is.
 
+**--preserve-fds**=*N*
+
+Pass down to the process N additional file descriptors (in addition to 0, 1, 2).  The total FDs will be 3+N.
```
mheon, line 614: Mention this is local-only

mheon: Actually, maybe not... I don't think we do this for --latest, do we?

QiWang19: > Actually, maybe not... I don't think we do this for --latest, do we?
>

 I don't see podman run has `--latest`. podman exec with both --latest and --preserve-fds seems working
```
$ podman exec -it --latest --preserve-fds=1 sh
/ # ls -l /proc/self/fd
total 0
lrwx------    1 root     root            64 Jun 16 19:52 0 -> /dev/pts/0
lrwx------    1 root     root            64 Jun 16 19:52 1 -> /dev/pts/0
lrwx------    1 root     root            64 Jun 16 19:52 2 -> /dev/pts/0
lrwx------    1 root     root            64 Jun 16 19:52 3 -> /null
ls: /proc/self/fd/4: cannot read link: No such file or directory
lr-x------    1 root     root            64 Jun 16 19:52 4
```


> Mention this is local-only

the `--preserve-fds` was hidden if remote in `podman exec`, so I also hide it for `podman run`. I can mention it is local-only.

mheon: I don't think we need further manpage changes, it looks like other flags hidden if remote are still in the manpages.

QiWang19:
> Please add a system test, possibly in `test/system/030-run.bats`, using something like what I messaged you earlier today. The e2e test here is not very useful: all it does is confirm that the option does not cause a syntax error. It does not test any actual functionality.

https://github.com/containers/libpod/pull/6625/files#diff-f1c2f651579cf49df98160fa0e2d50a7 test failed. @edsantiago  Can you help me diagnose if it's because the test not correct? Otherwise, my implementation may not as expected.

edsantiago:
@QiWang19 this is what the test should look like:
```bash
@test "podman run --preserve-fds" {
    skip_if_remote

    content=$(random_string 20)
    echo "$content" > $PODMAN_TMPDIR/tempfile

    run_podman run --rm -i --preserve-fds=2 $IMAGE sh -c "cat <&4" 4<$PODMAN_TMPDIR/tempfile
    is "$output" "$content" "container read input from fd 4"
}
```
Please use that. It still fails, but it does so for a real-problem reason which is being tracked in #6653 
```
 2020/06/17 15:51:44 bolt.Close(): funlock error: bad file descriptor
time="2020-06-17T15:51:44-06:00" level=error msg="failed to close libpod db: \"db file close: close /var/lib/containers/storage/libpod/bolt_state.db: bad file descriptor\""
Error: error saving container 26cd8352b73d3c07b46343c63f3b3b5c00e37f28ed685ace08788745025a0704 state: write /var/lib/containers/storage/libpod/bolt_state.db: bad file descriptor

```
Please coordinate efforts to resolve that issue.

Explanation of what I changed:
* use FD 4, not 3: I forgot that BATS uses fd 3 for internal purposes.
* Remove the NONLOCAL_IMAGE stuff. You copy-pasted that from another test. It is not necessary or desired here (it introduces a point of failure because it relies on remote registry)
* use random content, not hardcoded strings which could easily appear via other ways
* use a file in `$PODMAN_TMPDIR` - otherwise "testfile" will pollute your current working directory
* use `--rm` in the run command, to avoid having to clean up

QiWang19:
@edsantiago Thanks!

edsantiago:
@QiWang19 it looks like #6653 might not get attention soon. Since I believe your PR is correct, and the real problem is elsewhere, I'd be OK with you adding `skip "enable this once #6653 is fixed"` to the BATS test and trying to get your PR committed.

mheon:
Hold this until Monday, 2.0 is feature-frozen

rhatdan:
@mheon Are you ok to merge this now?

libpod/container.go
```diff
@@ -418,6 +418,11 @@ type ContainerConfig struct {
 
 	// HealthCheckConfig has the health check command and related timings
 	HealthCheckConfig *manifest.Schema2HealthConfig `json:"healthcheck"`
+
+	// PreserveFDs is a number of additional file descriptors (in addition
+	// to 0, 1, 2) that will be passed to the executed process. The total FDs
+	// passed will be 3 + PreserveFDs.
+	PreserveFDs uint `json:"preserveFds,omitempty"`
```
mheon, line 425: This needs to be `json:"-"`

pkg/specgen/specgen.go
mheon: This should not be `json:"-"`

libpod/container.go
mheon: Wait, I'm confused, nevermind.

pkg/specgen/specgen.go
mheon: Nevermind, I'm confused.

mheon:
LGTM, but would like final signoff from @giuseppe

QiWang19:
Thanks.
@giuseppe PTAL.

PR #6630, opened 2020-06-16
Show Anon, GID, UID in v2 volumes

ashley-cui:
Anon, GID, UID parameters previously hidden if empty in podman volume for API v2.

Status still hidden if empty, because it should not display a null string

Fixes https://github.com/containers/libpod/issues/6524


rhatdan:
LGTM
@mheon PTAL

PR #6631, opened 2020-06-16
Fix handling of old oci hooks

rhatdan:
Podman is blowing up with oci-umount hook, because
it was never rewritten to support the v1.0.0 value.

This PR adds support for the older version and cleans
up the hook handling.

### DIFF
--- a/pkg/hooks/read.go
+++ b/pkg/hooks/read.go
@@ -3,12 +3,12 @@ package hooks
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
+	old "github.com/containers/libpod/pkg/hooks/0.1.0"
 	current "github.com/containers/libpod/pkg/hooks/1.0.0"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -49,7 +49,7 @@ func read(content []byte) (hook *current.Hook, err error) {
 	}
 	reader, ok := Readers[ver.Version]
 	if !ok {
-		return nil, fmt.Errorf("unrecognized hook version: %q", ver.Version)
+		return nil, errors.Errorf("unrecognized hook version: %q", ver.Version)
 	}
 
 	hook, err = reader(content)
@@ -95,4 +95,6 @@ func ReadDir(path string, extensionStages []string, hooks map[string]*current.Ho
 
 func init() {
 	Readers[current.Version] = current.Read
+	Readers[old.Version] = old.Read
+	Readers[""] = old.Read
 }
--- a/pkg/hooks/0.1.0/hook.go
+++ b/pkg/hooks/0.1.0/hook.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"strings"
 
-	"github.com/containers/libpod/pkg/hooks"
 	current "github.com/containers/libpod/pkg/hooks/1.0.0"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 )
@@ -32,8 +31,9 @@ type Hook struct {
 	HasBindMounts *bool `json:"hasbindmounts,omitempty"`
 }
 
-func read(content []byte) (hook *current.Hook, err error) {
+func Read(content []byte) (hook *current.Hook, err error) {
 	var raw Hook
+
 	if err = json.Unmarshal(content, &raw); err != nil {
 		return nil, err
 	}
@@ -85,9 +85,4 @@ func read(content []byte) (hook *current.Hook, err error) {
 	}
 
 	return hook, nil
-}
-
-func init() {
-	hooks.Readers[""] = read
-	hooks.Readers[Version] = read
 }
--- a/pkg/hooks/0.1.0/hook_test.go
+++ b/pkg/hooks/0.1.0/hook_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestGood(t *testing.T) {
-	hook, err := read([]byte("{\"hook\": \"/a/b/c\", \"stages\": [\"prestart\"], \"cmds\": [\"sh\"]}"))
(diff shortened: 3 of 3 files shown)


rhatdan:
@edsantiago @haircommander PTAL

rhatdan:
Fixes: https://github.com/containers/libpod/issues/6629

haircommander:
LGTM

haircommander:
I have a guess you were looking for @heatmiser as well :smiley:

pkg/hooks/read.go
```diff
@@ -95,4 +96,6 @@ func ReadDir(path string, extensionStages []string, hooks map[string]*current.Ho
 
 func init() {
 	Readers[current.Version] = current.Read
+	Readers[old.Version] = old.Read
+	Readers[""] = old.Read
```
edsantiago, line 100: Well, it works, I guess, but this usage is new to me so of course I had to edit `/usr/share/containers/oci/hooks.d/oci-umount.json` and add `"version": "xx",` for several values of `xx`. Results:
| Version | Error |
| :--- | :--- |
| "0.1.0" | works fine |
| "0.1.1" | error setting up OCI Hooks ... unrecognized hook version: "0.1.1" |
| "1.0.0" | error setting up OCI Hooks ... 1.0.0: json: cannot unmarshal string into Go struct field Hook.hook of type specs.Hook |

I'll buy that my `1.0.0` is my own problem, probably because the format has changed somehow and I can't just add "version" without changing other things.

But, is there ever a possibility of a version other than `0.1.0`, `1.0.0`, or (absent)? Is there a way to specify a fallback `Reader`?

rhatdan: Sure, it's code.  I will default to current.Read

pkg/hooks/read.go
```diff
@@ -49,7 +49,7 @@ func read(content []byte) (hook *current.Hook, err error) {
 	}
 	reader, ok := Readers[ver.Version]
 	if !ok {
-		return nil, fmt.Errorf("unrecognized hook version: %q", ver.Version)
+		reader = current.Read
```
edsantiago, line 52: Thanks, I'm comforted that this is more futureproof. At the risk of annoying you even further, WDYT of:
```
    logrus.Warnf("unrecognized hook version %q; defaulting to %q reader", ver.Version, current.Version)
```
(I know it will not show by default, but could help in unforeseen errors if someone remembers to use `--log-level`)

pkg/hooks/read.go
heatmiser: Not sure if this helps...but I just tried editing `/usr/share/containers/oci/hooks.d/oci-umount.json` and add `"version": "0.1.0",` with result:

`Error: error setting up OCI Hooks: parsing hook "/usr/share/containers/oci/hooks.d/oci-umount.json": unrecognized hook version: "0.1.0"`

pkg/hooks/read_test.go
```diff
@@ -171,27 +171,6 @@ func TestUnknownDir(t *testing.T) {
 	}
 }
 
-func TestBadDir(t *testing.T) {
```
mheon, line 174: Why are we dropping this? I think we should still fail on hooks with an invalid version

pkg/hooks/read.go
mheon: I'm a little iffy here. We want to support `""` and `"0.0.1"` - I'm less sure we want to support `"-1"`, `"this is not a number"`, and `"1/0"` as valid hook versions.

pkg/hooks/read_test.go
rhatdan: In order to handle Ed's vision which is more future proof, we had to drop this test.  Let's discuss today. BTW Maybe we should just merge the original and then talk about this at standup.

edsantiago: Maybe my suggestion _isn't_ more futureproof. Maybe if there's a v2.0.0 or v1.2.3 it will be incompatible enough that podman _should_ abort early. I have no sense for these hooks, for the json format, for when/how that format changes. I just think we need to think about it in advance.

mheon: If we had a more easily parsed version number, I'd say we should accept anything with major version 1 and deny anything else, assuming semantic versioning.

heatmiser: Is there a (relatively easy) workaround that can implemented in the meantime, for users to be able to launch containers on their systems in the meantime?

edsantiago: @heatmiser you already tried adding `"version":"0.1.0",` to the json... I wonder, could you try `"1.0.0"`? I apologize for not trying it myself first.

heatmiser: Version | Error
-- | --
"0.1.0" | error setting up OCI Hooks ... unrecognized hook version: "0.1.0"
"0.1.1" | error setting up OCI Hooks ... unrecognized hook version: "0.1.1"
"1.0.0" | error setting up OCI Hooks ... 1.0.0: json: cannot unmarshal string into Go struct field Hook.hook of type specs.Hook

Fedora 31
`# podman -v`
`podman version 2.0.0-dev`
`# rpm -q podman`
`podman-2.0.0-1592321502.git908bc3fb.fc31.x86_64`
...from Copr repo for Upstream_CRIO_Family owned by baude
Recently executed package updates, not sure what previous version of podman was, but it worked
Note that I am also seeing the same error messages for `/usr/share/containers/oci/hooks.d/oci-systemd-hook.json`  as I am for `/usr/share/containers/oci/hooks.d/oci-umount.json` when I attempt to start containers via podman.

pkg/hooks/read.go
rhatdan: @heatmiser Try that with my patch and it should work.

rhatdan:
@baude @mheon @jwhonce @TomSweeneyRedHat @QiWang19 @ashley-cui PTAL and get this merged.

edsantiago:
LGTM.

PR #6614, opened 2020-06-16
Follow logs using event channel

jgallucci32:
Changes method for following logs to use an event
channel rather than a sleep timer to help prevent
race conditions.

Close #6531

Signed-off-by: jgallucci32 <john.gallucci.iv@gmail.com>

PR #6632, opened 2020-06-16
Use StopAtEOF() to avoid early termination of following logs

jgallucci32:
Signed-off-by: jgallucci32 <john.gallucci.iv@gmail.com>

### DIFF
--- a/libpod/container_log.go
+++ b/libpod/container_log.go
@@ -87,7 +87,7 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 				}
 				if state != define.ContainerStateRunning && state != define.ContainerStatePaused {
 					err := t.StopAtEOF()
-					if err != nil && fmt.Sprintf("%v",err) != "tail: stop at eof" {
+					if err != nil && fmt.Sprintf("%v", err) != "tail: stop at eof" {
 						logrus.Error(err)
 					}
 					break
--- a/libpod/container_log.go
+++ b/libpod/container_log.go
@@ -1,6 +1,7 @@
 package libpod
 
 import (
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@QiWang19 @ashley-cui @jwhonce @mheon PTAL

jgallucci32:
/hold

2 checks failed. Seem to be the sames ones as before. No better, no worse.

jgallucci32:
@edsantiago I am able to reliably get this to fail with the following commands from the test
```
podman run --rm --name logs-f-rm -dt docker.io/library/alpine:latest sh -c 'echo podman; sleep 1; echo podman'
podman logs -f logs-f-rm
podman run --rm --name logs-f-rm -dt docker.io/library/alpine:latest true

```
NOTE: You have to run the 3 commands at the same time. Any delay in running the third command results in expected behavior.

This gives a good starting point to figure out where the race condition lives.

jgallucci32:
> @edsantiago I am able to reliably get this to fail with the following commands from the test
> 
> ```
> podman run --rm --name logs-f-rm -dt docker.io/library/alpine:latest sh -c 'echo podman; sleep 1; echo podman'
> podman logs -f logs-f-rm
> podman run --rm --name logs-f-rm -dt docker.io/library/alpine:latest true
> ```
> 
> NOTE: You have to run the 3 commands at the same time. Any delay in running the third command results in expected behavior.
> 
> This gives a good starting point to figure out where the race condition lives.

Ok it looks like PR https://github.com/containers/libpod/pull/6614 fixes this condition. What I need to test is if `StopAtEOF()` in combination with event channels is the correct fix. It's possible it is waiting for DONE because event channels is faster and not completing the log to end of file.

edsantiago:
There's no race condition, this is a broken, badly written test. `logs -f` should terminate when the container terminates, we all agree on that -- there is no reason whatsoever, though, for it to guarantee that the container has been removed and cleaned up. This e2e test needs to be rewritten.

jgallucci32:
@edsantiago Which specific test needs to be rewritten?

I'm thinking of closing this PR since it proved using `Stop()` and `StopAtEOF()` had to effect and this is being reworked in a separate PR.

edsantiago:
@jgallucci32 see #6640

jgallucci32:
/Close

This has been superseded by #6644 and #6614

Issue #6578, opened 2020-06-11
podman-remote: build: creates a weird 'localhost/latest' image


edsantiago:
Usual setup for podman system service
```console
$ mkdir a
$ echo 'FROM docker.io/library/alpine:latest' >a/Containerfile
$ ./bin/podman-remote --url $MYSOCK build --layers=true -t test a
--> a24bb401329
a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e

FROM docker.io/library/alpine:latest
COMMIT test
Successfully built a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e
$ ./bin/podman images
REPOSITORY                TAG     IMAGE ID      CREATED      SIZE
localhost/test            latest  a24bb4013296  12 days ago  5.85 MB
docker.io/library/alpine  latest  a24bb4013296  12 days ago  5.85 MB
localhost/latest          latest  a24bb4013296  12 days ago  5.85 MB
^^^^^^^^^^^^^^^^
```
That last `localhost/latest` is totally weird and unexpected, and is causing failures in e2e test

baude:
this is NOT a remote problem ... it happens with `podman build` ... still checking.

baude:
disregard previous comment.

Issue #6577, opened 2020-06-11
podman-remote: build --file SUBPATH: exit 0, but NOP


edsantiago:
Usual setup for podman system server
```console
$ mkdir a
$ echo 'FROM docker.io/library/alpine:latest' >a/Containerfile
$ ./bin/podman-remote --url $MYSOCK build --layers=true -t test --file a/Containerfile a
```
This exits immediately, with exit status 0, having performed nothing. On server debug log:
```
DEBU[0449] APIHandler -- Method: POST URL: /v1.0.0/libpod/build?dockerfile=a%2FContainerfile&forcerm=1&platform=linux%2Famd64&rm=1&shmsize=67108864&t=test
INFO[0449] Request Failed(Internal Server Error): error reading info about "/dev/shm/libpod_builder405942463/build/a/Containerfile": stat /dev/shm/libpod_builder405942463/build/a/Containerfile: no such file or directory
```
Running without `--file etc` proceeds farther, but with its own weirdness which I will report after this.

TomSweeneyRedHat:
@edsantiago it could well be end of day brain for me, but I'm not able to get your command to run using the latest  upstream.  The `--url` isn't found?  I've not played with the remote side much, so could be a user error here.  Also, what did you set `$MYSOCK` to and what command did you run for the server?
```
# ./bin/podman-remote --url $MYSOCK build --layers=true -t test --file ~/Dockerfile.alpine .
Error: unknown flag: --url
```

edsantiago:
@tomsweeney sorry, I left out important steps because I was filing so many issues that day. "Usual setup" means:
```console
$ export MYSOCK=unix:/tmp/mypodmansock
$ ./bin/podman system service --timeout=0 $MYSOCK
```
That hangs (serving), so you need to set MYSOCK on the other shell.

I'm not seeing the same failure any more though. I will investigate further tomorrow.

edsantiago:
Duh. The key is that podman server can't be in the same pwd as podman client. Yep, problem still exists.

TomSweeneyRedHat:
@edsantiago I still can't get  podman-remote to find a `--url` option.  Is that bit correct?

edsantiago:
@TomSweeneyRedHat my first and only guess is that somehow you've got the build tags wrong and are building varlink instead of v2? But I have no idea how to even test that. `--version` doesn't help. Or maybe you're typing `make podman` instead of `make`? Can you `ls -l bin/podman-remote`? 

[...]

PR #6634, opened 2020-06-16
fix misc remote build issues

baude:
address problem when multiple -t were sent.  and rework remote build's tarball if a context dir is given other than ".".

Fixes: #6578
Fixes: #6577

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -48,34 +48,34 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 	defer os.RemoveAll(anchorDir)
 
 	query := struct {
-		Dockerfile  string `schema:"dockerfile"`
-		Tag         string `schema:"t"`
-		ExtraHosts  string `schema:"extrahosts"`
-		Remote      string `schema:"remote"`
-		Quiet       bool   `schema:"q"`
-		NoCache     bool   `schema:"nocache"`
-		CacheFrom   string `schema:"cachefrom"`
-		Pull        bool   `schema:"pull"`
-		Rm          bool   `schema:"rm"`
-		ForceRm     bool   `schema:"forcerm"`
-		Memory      int64  `schema:"memory"`
-		MemSwap     int64  `schema:"memswap"`
-		CpuShares   uint64 `schema:"cpushares"`  //nolint
-		CpuSetCpus  string `schema:"cpusetcpus"` //nolint
-		CpuPeriod   uint64 `schema:"cpuperiod"`  //nolint
-		CpuQuota    int64  `schema:"cpuquota"`   //nolint
-		BuildArgs   string `schema:"buildargs"`
-		ShmSize     int    `schema:"shmsize"`
-		Squash      bool   `schema:"squash"`
-		Labels      string `schema:"labels"`
-		NetworkMode string `schema:"networkmode"`
-		Platform    string `schema:"platform"`
-		Target      string `schema:"target"`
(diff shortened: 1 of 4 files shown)


edsantiago:
Confirmed: it addresses #6577 and #6578. Thank you!

pkg/api/handlers/compat/images_build.go
```diff
@@ -107,20 +107,24 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
...
 	)
-	if strings.Contains(query.Tag, ":") {
-		tokens := strings.SplitN(query.Tag, ":", 2)
-		name = tokens[0]
-		tag = tokens[1]
+	if len(query.Tag) > 0 {
+		output = query.Tag[0]
 	}
+	if len(query.Tag) > 1 {
+		additionalNames = append(additionalNames, query.Tag[1:]...)
```
rhatdan, line 117: Nit, this could just be
additionalNames := query.Tag[1:]

rhatdan:
LGTM, Fix Nit, if you find other issues.

pkg/domain/infra/tunnel/images.go
```diff
@@ -288,3 +305,64 @@ func (ir *ImageEngine) Shutdown(_ context.Context) {
...
+			return tw.Close()
+		}
+		if err != nil {
+			return err
+		}
+
+		var body io.Reader = tr
+		name := h.Name
+		data, ok, skip, err := fn(h, tr)
+		logrus.Debugf("Transform %s -> %s: data=%t ok=%t skip=%t err=%v", name, h.Name, data != nil, ok, skip, err)
```
jwhonce, line 332:
```suggestion
		logrus.Debugf("Transform %q -> %q: data=%t ok=%t skip=%t err=%v", name, h.Name, data != nil, ok, skip, err)
```
small nit for clarity.

pkg/domain/infra/tunnel/images.go
```diff
@@ -288,3 +305,64 @@ func (ir *ImageEngine) Shutdown(_ context.Context) {
...
+	for {
+		h, err := tr.Next()
+		if err == io.EOF {
+			return tw.Close()
+		}
+		if err != nil {
+			return err
+		}
+
+		var body io.Reader = tr
```
jwhonce, line 329: `tr` doesn't appear to change value in the loop.  Should be moved out of the loop.

pkg/domain/infra/tunnel/images.go
```diff
@@ -288,3 +305,64 @@ func (ir *ImageEngine) Shutdown(_ context.Context) {
...
+		name := h.Name
```
jwhonce, line 330: From archive/tar documentation:

> For forward compatibility, users that retrieve a Header from Reader.Next, mutate it in some ways, and then pass it back to Writer.WriteHeader should do so by creating a new Header and copying the fields that they are interested in preserving.

baude: im seriously hesitant to fiddle with with this as it was taken directly from imagebuilders code, which is presumably tried and true?

pkg/api/handlers/compat/images_build.go
```diff
@@ -107,20 +107,24 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
...
+	}
+	//if strings.Contains(query.Tag, ":") {
```
mheon, line 119: Can we drop this? Looks like it's no longer needed

baude:
All fixed up with what I am comfortable with ...

baude:
you know the drill ....

PR #6636, opened 2020-06-16
Re-add resource limit warnings to Specgen

mheon:
These were part of Podman v1.9, but were lost in the transition to using Specgen to create containers. Most resource limits are checked via the sysinfo package to ensure they are safe to use (the cgroup is mounted, kernel support is present, etc) and removed if not safe. Further, bounds checks are performed to ensure that values are valid.

Ensure these warnings are printed client-side when they occur. This part is a little bit gross, as it happens in pkg/infra and not cmd/podman, which is largely down to how we implemented `podman run` - all the work is done in pkg/infra and it returns only once the container has exited, and we need warnings to print *before* the container runs. The solution here, while inelegant, avoid the need to extensively refactor our handling of run.

Should fix blkio-limit warnings that were identified by the FCOS test suite.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -635,8 +635,6 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	s.Remove = c.Rm
 	s.StopTimeout = &c.StopTimeout
 
-	// TODO where should we do this?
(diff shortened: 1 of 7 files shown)


mheon:
@baude PTAL

@dustymabe FYI

rhatdan:
LGTM

mheon:
@vrothberg @giuseppe @TomSweeneyRedHat @baude PTAL and merge

TomSweeneyRedHat:
LGTM

PR #6641, opened 2020-06-17
generate systemd: `ExecStopPost` for all units

vrothberg:
Add an `ExecStopPost` run even for units generated without `--new`.
Although it may seem redundant to run `container/pod stop` twice at
first glance, we really need the post run.  If the main PID (i.e.,
conmon) is killed, systemd will not execute `ExecStop` but only the
post one.  We made this obeservation in a customer issue and could
reproduce the behavior consistently.  Hence, the post run is needed
to properly clean up when conmon is killed and it's pretty much a
NOP in all other cases.

Credits to Ulrich Obergfell for throrough and detailed analyses,
which ultimately lead to this fix.

### DIFF
--- a/pkg/systemd/generate/containers.go
+++ b/pkg/systemd/generate/containers.go
@@ -83,9 +83,7 @@ ExecStartPre={{.ExecStartPre}}
 {{- end}}
 ExecStart={{.ExecStart}}
 ExecStop={{.ExecStop}}
-{{- if .ExecStopPost}}
 ExecStopPost={{.ExecStopPost}}
-{{- end}}
 PIDFile={{.PIDFile}}
 KillMode=none
 Type=forking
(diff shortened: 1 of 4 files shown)


vrothberg:
@mheon, that's the last one of my urgent items for v2. Uli confirmed that this is working.

mheon:
LGTM

rhatdan:
LGTM
Will have to wait until the race issue is fixed, then we can get this in .

pkg/systemd/generate/containers.go
```diff
@@ -170,6 +168,7 @@ func executeContainerTemplate(info *containerInfo, options entities.GenerateSyst
 	info.EnvVariable = EnvVariable
 	info.ExecStart = "{{.Executable}} start {{.ContainerNameOrID}}"
 	info.ExecStop = "{{.Executable}} stop {{if (ge .StopTimeout 0)}}-t {{.StopTimeout}}{{end}} {{.ContainerNameOrID}}"
+	info.ExecStopPost = "{{.Executable}} stop {{if (ge .StopTimeout 0)}}-t {{.StopTimeout}}{{end}} {{.ContainerNameOrID}}"
```
giuseppe, line 171: should it be '-{{.Executable}} stop {{if (ge .StopTimeout 0)}}-t {{.StopTimeout}}{{end}} {{.ContainerNameOrID}}`  (extra - at the beginning) to ignore failures?

vrothberg: I think we're good as `stop` doesn't error if a container is already stopped.

giuseppe: is there any risk that the container is created with `--rm` so the first stop will also delete it?

vrothberg:
Repushed

PR #6666, opened 2020-06-18
podman: add new cgroup mode split

giuseppe:
add a new cgroup mode split.

When running under systemd there is no need to create yet another cgroup for the container.
    
With conmon-delegated the current cgroup will be split in two sub cgroups:
    
    - supervisor
    - container
    
The supervisor cgroup will hold conmon and the podman process, while the container cgroup is used by the OCI runtime (using the cgroupfs backend).

Closes: https://github.com/containers/libpod/issues/6400

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -78,9 +78,12 @@ If the host uses cgroups v1, the default is set to **host**.  On cgroups v2 the 
 **--cgroups**=*mode*
 
 Determines whether the container will create CGroups.
-Valid values are *enabled*, *disabled*, *no-conmon*, which the default being *enabled*.
+Valid values are *enabled*, *disabled*, *no-conmon*, *split*, which the default being *enabled*.
+
+The *enabled* option will create a new cgroup under the cgroup-parent.
 The *disabled* option will force the container to not create CGroups, and thus conflicts with CGroup options (**--cgroupns** and **--cgroup-parent**).
 The *no-conmon* option disables a new CGroup only for the conmon process.
+The *split* option splits the current cgroup in two sub-cgroups: one for conmon and one for the container payload.  It is not possible to set *--cgroup-parent* with *split*.
 
 **--cgroup-parent**=*path*
 
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -89,14 +89,16 @@ Set the cgroup namespace mode for the container.
 
 If the host uses cgroups v1, the default is set to **host**.  On cgroups v2, the default is **private**.
 
-**--cgroups**=**enabled**|**disabled**|**no-conmon**
+**--cgroups**=**enabled**|**disabled**|**no-conmon**|**split**
 
 Determines whether the container will create CGroups.
 
-Default is **enabled**. The **disabled** option will force the container
-to not create CGroups, and thus conflicts with CGroup options
-(**--cgroupns** and **--cgroup-parent**).
+Default is **enabled**.
+
+The **enabled** option will create a new cgroup under the cgroup-parent.
+The **disabled** option will force the container to not create CGroups, and thus conflicts with CGroup options (**--cgroupns** and **--cgroup-parent**).
 The **no-conmon** option disables a new CGroup only for the **conmon** process.
+The **split** option splits the current cgroup in two sub-cgroups: one for conmon and one for the container payload.  It is not possible to set **--cgroup-parent** with **split**.
 
 **--cgroup-parent**=*path*
 
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -31,6 +31,7 @@ import (
 	"github.com/containers/libpod/pkg/resolvconf"
 	"github.com/containers/libpod/pkg/rootless"
 	"github.com/containers/libpod/pkg/util"
+	"github.com/containers/libpod/utils"
 	"github.com/containers/storage/pkg/archive"
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/opencontainers/runc/libcontainer/user"
@@ -1505,8 +1506,17 @@ func (c *Container) getOCICgroupPath() (string, error) {
 	switch {
 	case (rootless.IsRootless() && !unified) || c.config.NoCgroups:
 		return "", nil
+	case c.config.CgroupsMode == cgroupSplit:
+		if c.config.CgroupParent != "" {
+			return c.config.CgroupParent, nil
+		}
+		selfCgroup, err := utils.GetOwnCgroup()
+		if err != nil {
+			return "", err
+		}
+		return filepath.Join(selfCgroup, "container"), nil
 	case c.runtime.config.Engine.CgroupManager == config.SystemdCgroupsManager:
-		// When runc is set to use Systemd as a cgroup manager, it
+		// When the OCI runtime is set to use Systemd as a cgroup manager, it
 		// expects cgroups to be passed as follows:
 		// slice:prefix:name
 		systemdCgroups := fmt.Sprintf("%s:libpod:%s", path.Base(c.config.CgroupParent), c.ID())
--- a/libpod/oci_conmon.go
+++ b/libpod/oci_conmon.go
@@ -1,0 +1,7 @@
+package libpod
+
+const (
+	// cgroupSplit is the cgroup mode for reusing the current cgroup both
+	// for conmon and for the container payload.
+	cgroupSplit = "split"
+)
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -57,7 +57,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	createFlags.StringVar(
 		&cf.CGroupsMode,
 		"cgroups", containerConfig.Cgroups(),
-		`control container cgroup configuration ("enabled"|"disabled"|"no-conmon")`,
+		`control container cgroup configuration ("enabled"|"disabled"|"no-conmon"|"split")`,
 	)
 	createFlags.StringVar(
 		&cf.CGroupParent,
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -205,6 +205,10 @@ func createInit(c *cobra.Command) error {
 		cliVals.Env = env
 	}
 
+	if c.Flag("cgroups").Changed && cliVals.CGroupsMode == "split" && registry.IsRemote() {
+		return errors.Errorf("the option --cgroups=%q is not supported in remote mode", cliVals.CGroupsMode)
+	}
+
 	// Docker-compatibility: the "-h" flag for run/create is reserved for
 	// the hostname (see https://github.com/containers/libpod/issues/1367).
 
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/libpod/libpod/lock"
 	"github.com/containers/libpod/pkg/namespaces"
 	"github.com/containers/libpod/pkg/rootless"
+	"github.com/containers/libpod/utils"
 	"github.com/containers/storage"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
@@ -1089,10 +1090,25 @@ func (c *Container) NamespacePath(linuxNS LinuxNS) (string, error) { //nolint:in
 
 // CGroupPath returns a cgroups "path" for a given container.
 func (c *Container) CGroupPath() (string, error) {
-	switch c.runtime.config.Engine.CgroupManager {
-	case config.CgroupfsCgroupsManager:
+	switch {
+	case c.config.CgroupsMode == cgroupSplit:
+		if c.config.CgroupParent != "" {
+			return "", errors.Errorf("cannot specify cgroup-parent with cgroup-mode %q", cgroupSplit)
+		}
+		cg, err := utils.GetCgroupProcess(c.state.ConmonPID)
+		if err != nil {
+			return "", err
+		}
+		// Use the conmon cgroup for two reasons: we validate the container
+		// delegation was correct, and the conmon cgroup doesn't change at runtime
+		// while we are not sure about the container that can create sub cgroups.
+		if !strings.HasSuffix(cg, "supervisor") {
+			return "", errors.Errorf("invalid cgroup for conmon %q", cg)
+		}
+		return strings.TrimSuffix(cg, "/supervisor") + "/container", nil
+	case c.runtime.config.Engine.CgroupManager == config.CgroupfsCgroupsManager:
 		return filepath.Join(c.config.CgroupParent, fmt.Sprintf("libpod-%s", c.ID())), nil
-	case config.SystemdCgroupsManager:
+	case c.runtime.config.Engine.CgroupManager == config.SystemdCgroupsManager:
 		if rootless.IsRootless() {
 			uid := rootless.GetRootlessUID()
 			parts := strings.SplitN(c.config.CgroupParent, "/", 2)
--- a/libpod/container_validate.go
+++ b/libpod/container_validate.go
@@ -34,6 +34,10 @@ func (c *Container) validate() error {
 		return errors.Wrapf(define.ErrInvalidArg, "cannot both create a network namespace and join another container's network namespace")
 	}
 
+	if c.config.CgroupsMode == cgroupSplit && c.config.CgroupParent != "" {
+		return errors.Wrapf(define.ErrInvalidArg, "cannot specify --cgroup-mode=split with a cgroup-parent")
+	}
+
 	// Not creating cgroups has a number of requirements, mostly related to
 	// the PID namespace.
 	if c.config.NoCgroups || c.config.CgroupsMode == "disabled" {
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -881,6 +881,12 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		return err
 	}
 
+	if ctr.config.CgroupsMode == cgroupSplit {
+		if err := utils.MoveUnderCgroupSubtree("supervisor"); err != nil {
+			return err
+		}
+	}
+
 	args := r.sharedConmonArgs(ctr, ctr.ID(), ctr.bundlePath(), filepath.Join(ctr.state.RunDir, "pidfile"), ctr.LogPath(), r.exitsDir, ociLog, ctr.LogDriver(), logTag)
 
 	if ctr.config.Spec.Process.Terminal {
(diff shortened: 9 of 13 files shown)


giuseppe:
@vrothberg PTAL

libpod/container_internal_linux.go
```diff
@@ -1505,6 +1506,17 @@ func (c *Container) getOCICgroupPath() (string, error) {
 	case (rootless.IsRootless() && !unified) || c.config.NoCgroups:
 		return "", nil
 	case c.runtime.config.Engine.CgroupManager == config.SystemdCgroupsManager:
+		if c.config.CgroupsMode == "no-conmon-delegated" {
```
vrothberg, line 1509: Can we make "no-conmon-delegated" a constent somewhere?

giuseppe: > The commit messages states `conmon-delegated` but the code adds a `no-conmon-delegated`. I'd guess we should update the docs as well.

Thanks, I've renamed it to `conmon-delegated`

> Do we need to update the systemd unit files in `pkg/system/generate`?

I don't think we can do that yet.  crun needs https://github.com/containers/crun/pull/409

docs/source/markdown/podman-run.1.md
```diff
@@ -89,14 +89,15 @@ Set the cgroup namespace mode for the container.
...
-**--cgroups**=**enabled**|**disabled**|**no-conmon**
+**--cgroups**=**enabled**|**disabled**|**no-conmon**|**conmon-delegated**
 
 Determines whether the container will create CGroups.
 
 Default is **enabled**. The **disabled** option will force the container
 to not create CGroups, and thus conflicts with CGroup options
 (**--cgroupns** and **--cgroup-parent**).
 The **no-conmon** option disables a new CGroup only for the **conmon** process.
+The **conmon-delegated** option reuses the current cgroup for both the conmon and the container payload, it works only on cgroup v2.
```
mheon, line 100: If it's also specific to systemd cgroups (looks like it is) we should also mention that

mheon:
Overall, I like this. I think it does break `podman stats`/`podman top` though, and it probably doesn't need to - we just need to look in the current cgroup instead

docs/source/markdown/podman-run.1.md
giuseppe: not sure it is so specific to systemd.  It can be used also without systemd, even if it is less useful

goochjj:
Is there a reason why you did cgroups 2 only?  This approach works for cgroups 1:
https://github.com/goochjj/libpod/commit/3d6ec2043b75744a5fec60088c99ba56a51f6479

I'd happily make this coexist if need be.

goochjj:
For that matter, why is cgroup-parent so draconian? It only allows items to end in .slice.

And yet this works:
"cgroupsPath": "/machine.slice/rwhod.service/container",

in config.json

Container cgroups could be implemented with --cgroup-parent /machine.slice/%n/container, if the cgroup-parent check weren't rejecting it.  This commit proves it works.

Would a better implementation be implementing --conmon-cgroup-parent /machine.slice/%n/supervisor?

giuseppe:
> Container cgroups could be implemented with --cgroup-parent /machine.slice/%n/container, if the cgroup-parent check weren't rejecting it. This commit proves it works.

we could relax them.

> Would a better implementation be implementing --conmon-cgroup-parent /machine.slice/%n/supervisor?

The current implementation works well also without systemd, i.e. you have a cgroup you own and you want to create both conmon+container cgroup in it.

giuseppe:
> Is there a reason why you did cgroups 2 only? This approach works for cgroups 1:
> [goochjj@3d6ec20](https://github.com/goochjj/libpod/commit/3d6ec2043b75744a5fec60088c99ba56a51f6479)
> 
> I'd happily make this coexist if need be.

just simplicity.  We can surely add cgroup v1 support.  It must be done for each controller separately, it is not enough to get the cgroup from the pid controller

docs/source/markdown/podman-run.1.md
giuseppe: this will also break remote.  We probably need a check to not permit the option being used as remote as it changes the cgroup of the Podman process itself

goochjj:
> > Is there a reason why you did cgroups 2 only? This approach works for cgroups 1:
> > [goochjj@3d6ec20](https://github.com/goochjj/libpod/commit/3d6ec2043b75744a5fec60088c99ba56a51f6479)
> > I'd happily make this coexist if need be.
> 
> just simplicity. We can surely add cgroup v1 support. It must be done for each controller separately, it is not enough to get the cgroup from the pid controller

I read the cgroup from the pid controller, but in the Move call, I move all of them... that should do it, right?

I don't know if pids will always be 3 - I can fix the find if need be.

giuseppe:
> > 
> 
> I read the cgroup from the pid controller, but in the Move call, I move all of them... that should do it, right?
> 
> I don't know if pids will always be 3 - I can fix the find if need be.

there could be different cgroups for each controller, the safest is to read each controller and then do the move as part of the same loop for that controller

goochjj:
> > > 
> > 
> > 
> > I read the cgroup from the pid controller, but in the Move call, I move all of them... that should do it, right?
> > I don't know if pids will always be 3 - I can fix the find if need be.
> 
> there could be different cgroups for each controller, the safest is to read each controller and then do the move as part of the same loop for that controller

... except you were using the same call to determine the name of the supervisor cgroup - so I could only return 1

So should the supervisor cgroup name be based on the pids namespace, and the move do something else?  i.e. just tack "container" onto the end?  tack "container" onto the end if it's not /?

goochjj:
@giuseppe 
https://github.com/giuseppe/libpod/pull/2

A couple thoughts here:
1. The cgroup of the container is handled by the OCI runtime, as passed into the bundle.  All conmon-delegated does is ensure that cgroup passed in the spec is "<unit>/container", which needs an additional function (GetPidCgroup()) to determine what cgroups we were passed. 
2. The cgroup of the conmon process is already handled in oci_conmon_linux... So the new function to move to a subtree shouldn't be necessary... The strategy also uses a "cgroupfs" method for moving the Pids, not a "lets create a systemd transition unit to do it" method.  That code already exists in oci_conmon_linux, it just needs the same as above - use of GetPidCgroup to construct the "<unit>/supervisor" cgroup path.  And the cgroups package already deals with "do all the controllers" logic.

goochjj:
> Overall, I like this. I think it does break `podman stats`/`podman top` though, and it probably doesn't need to - we just need to look in the current cgroup instead

FYI top is working for me - it properly shows the /slice/unit/container cgroup

Stats is not.

giuseppe:
>     1. The cgroup of the container is handled by the OCI runtime, as passed into the bundle.  All conmon-delegated does is ensure that cgroup passed in the spec is "/container", which needs an additional function (GetPidCgroup()) to determine what cgroups we were passed.
> 
>     2. The cgroup of the conmon process is already handled in oci_conmon_linux... So the new function to move to a subtree shouldn't be necessary... The strategy also uses a "cgroupfs" method for moving the Pids, not a "lets create a systemd transition unit to do it" method.  That code already exists in oci_conmon_linux, it just needs the same as above - use of GetPidCgroup to construct the "/supervisor" cgroup path.  And the cgroups package already deals with "do all the controllers" logic.

for cgroup v2 there cannot be processes in the inner groups.  We need to make sure the conmon process is already in the right sub cgroup before it starts (or have a way to communicate to conmon to move itself).  We need to do such steps before setting up the container, the previous code path cannot be used.

> ... except you were using the same call to determine the name of the supervisor cgroup - so I could only return 1

there is only one hierarchy in cgroup v2, so that is enough.

In any case, I think I can just make my PR more generic to handle cgroup v1 too, even if it is not necessary to move the conmon process in a sub cgroup

goochjj:
pr updated w/ podman stats working

I'll be the first to admit, I don't fully understand all the implications, and I don't have a cgv2 machine to test on.  I do know this all works on cgv1, and it looks simpler and reuses code paths that were already present and seem to make sense.

But I don't see how podman stats is going to work if you never pass the cgroup parent back through the container object...

Thanks!

goochjj:
FCOS appears to be running cg1/2 hybrid, and pkg/cgroups doesn't adjust /sys/fs/cgroup/unified.

Is the long term plan to fix pkg/cgroups or deprecate it?

goochjj:
I see what you're saying.

Updated PR, works on cgroups 1, hybrid, and cg2 unified

giuseppe:
I will adjust this PR to work on cgroup v1. Hybrid mode is not supported and it will never be, controllers must either be on cgroup v1 or cgroup v2. Fedora 30 mounts hybrid mode but there are no controllers enabled by default on cgroup v2 so it is equivalent to the legacy mode.  There are also a few other things I need to take care of before I can drop the WIP label

giuseppe:
I've adapted it to work on cgroup v1 too.  I've added the missing bits, like support for stats and raising an error when used in remote mode.

The option is now called `split-current` as it explains better what Podman does

docs/source/markdown/podman-create.1.md
```diff
@@ -80,7 +80,7 @@ If the host uses cgroups v1, the default is set to **host**.  On cgroups v2 the
 Determines whether the container will create CGroups.
 Valid values are *enabled*, *disabled*, *no-conmon*, which the default being *enabled*.
 The *disabled* option will force the container to not create CGroups, and thus conflicts with CGroup options (**--cgroupns** and **--cgroup-parent**).
-The *no-conmon* option disables a new CGroup only for the conmon process.
+The *no-conmon* option disables a new CGroup only for the conmon process.  The *split-current* option reuses the current cgroup for both the conmon and the container payload, it works only on cgroup v2.
```
goochjj, line 83: remove "only works for cgroup v2"

docs/source/markdown/podman-run.1.md
```diff
@@ -89,14 +89,15 @@ Set the cgroup namespace mode for the container.
...
-**--cgroups**=**enabled**|**disabled**|**no-conmon**
+**--cgroups**=**enabled**|**disabled**|**no-conmon**|**split-current**
 
 Determines whether the container will create CGroups.
 
 Default is **enabled**. The **disabled** option will force the container
 to not create CGroups, and thus conflicts with CGroup options
 (**--cgroupns** and **--cgroup-parent**).
 The **no-conmon** option disables a new CGroup only for the **conmon** process.
+The **split-current** option reuses the current cgroup for both the conmon and the container payload, it works only on cgroup v2.
```
goochjj, line 100: remove only works on cgroup v2

docs/source/markdown/podman-create.1.md
giuseppe: thanks, fixed now

docs/source/markdown/podman-run.1.md
giuseppe: fixed

goochjj:
LGTM

goochjj:
@mheon @vrothberg PTAL

mheon:
We want to hold this until Monday - we're in a semi-feature-freeze state for v2.0.0 release, which should happen in a few hours. Will do a full review once that is cut.

docs/source/markdown/podman-create.1.md
```diff
@@ -80,7 +80,7 @@ If the host uses cgroups v1, the default is set to **host**.  On cgroups v2 the
 Determines whether the container will create CGroups.
 Valid values are *enabled*, *disabled*, *no-conmon*, which the default being *enabled*.
 The *disabled* option will force the container to not create CGroups, and thus conflicts with CGroup options (**--cgroupns** and **--cgroup-parent**).
-The *no-conmon* option disables a new CGroup only for the conmon process.
+The *no-conmon* option disables a new CGroup only for the conmon process.  The *split-current* option reuses the current cgroup for both the conmon and the container payload.
```
rhatdan, line 83: Can we just call this current?  I don't see where split- adds any additional information to the user.

docs/source/markdown/podman-run.1.md
```diff
@@ -89,14 +89,15 @@ Set the cgroup namespace mode for the container.
 
 If the host uses cgroups v1, the default is set to **host**.  On cgroups v2, the default is **private**.
 
-**--cgroups**=**enabled**|**disabled**|**no-conmon**
+**--cgroups**=**enabled**|**disabled**|**no-conmon**|**split-current**
```
rhatdan, line 92: Use current.

libpod/oci_conmon.go
```diff
@@ -0,0 +1,7 @@
+package libpod
+
+const (
+	// cgroupSplitCurrent is the cgroup mode for reusing the current cgroup both
+	// for conmon and for the container payload.
+	cgroupSplitCurrent = "split-current"
```
rhatdan, line 6: Just current

giuseppe: I'd prefer `split` in the name, since the current cgroup is splitted in two (supervisor and container).  `current` sounds like the current cgroup is used without any further change.

Should I rename it to `split`?

docs/source/markdown/podman-create.1.md
giuseppe: we may need `current` in future to really mean `current` and use the cgroup where Podman is running instead of creating two sub-cgroups

TomSweeneyRedHat: `reuse-current` or perhaps just `reuse` then?

libpod/oci_conmon.go
rhatdan: // cgroupSplitCurrent is the cgroup mode for reusing the current cgroup both
	// for conmon and for the container payload.
This description looks like you are using the same cgroup for conmon and the container?

I am thoroughly confused and `split` is not helping.  Could someone write down a table of the different modes we are looking at running cgroups

Can we get a modes table to descibe each one of these modes.

giuseppe: I'll fix it.  Do you prefer if I fix the comment and explain better why it is `split-current` or just use `current`?

goochjj: <table>
<tr><th>Cgroups</th><th>Parent   Cgroup</th><th>conmon</th><th>container</th></tr>
<tr><td>disabled</td><td>any</td><td>unchanged</td><td>unchanged</td></tr>
<tr><td>enabled</td><td>blank</td><td>/machine.slice/libpod-conmon-CID.scope</td><td>/machine.slice/libpod-CID.scope</td></tr>
<tr><td>enabled</td><td>slice</td><td>/(Specified   Slice)/libpod-conmon-CID.scope</td><td>/(Specified   Slice)/libpod-CID.scope</td></tr>
<tr><td>no-conmon</td><td>default</td><td>unchanged</td><td>/machine.slice/libpod-CID.scope</td></tr>
<tr><td>no-conmon</td><td>slice</td><td>unchanged</td><td>/(Specified   Slice)/libpod-CID.scope</td></tr>
<tr><td>split-current</td><td>default</td><td>(current   cgroup)/supervisor</td><td>(current   cgroup)/container</td></tr>
<tr><td>split-current</td><td>slice</td><td>invalid   (slice/supervisor)</td><td>invalid   (slice/supervisor/slice/container)</td></tr>
</table>

Perhaps split-current should disallow cgroup-parent, or at the very least it should remove the "only a slice is valid" restriction because that validation assumes libpod will be manufacturing the scope name, which split-current doesn't do.

Split-current is nice because
1) it does what systemd-nspawn does
2) Systemd units get split nicely under the unit name, i.e.
  /system.slice/service.unit/supervisor -> conmon
  /system.slice/service.unit/container -> everything else

All processes show up nicely in `systemctl status`.  Slice can be changed in the unit if need be. (i.e. Slice=/machine.slice)

Ultimately using split-current w/ systemd units means we want to split the unit that systemd created, so I'm ok if `--cgroup-parent`` is disabled.

I get that you dislike hyphenated options.  This option already has `no-conmon`.  Current isn't correct, current is basically what "disabled" is, as is "reuse".  Split would make sense.

rhatdan: Lets call it split.

docs/source/markdown/podman-run.1.md
rhatdan: Use split

libpod/oci_conmon.go
giuseppe: renamed and pushed

docs/source/markdown/podman-run.1.md
```diff
@@ -89,14 +89,15 @@ Set the cgroup namespace mode for the container.
 
 If the host uses cgroups v1, the default is set to **host**.  On cgroups v2, the default is **private**.
 
-**--cgroups**=**enabled**|**disabled**|**no-conmon**
+**--cgroups**=**enabled**|**disabled**|**no-conmon**|**split**
 
 Determines whether the container will create CGroups.
 
 Default is **enabled**. The **disabled** option will force the container
```
mheon, line 96: Can we add that "enabled" created a new cgroup under the cgroup parent, to distinguish it from "split" which reuses the current cgroup?

docs/source/markdown/podman-run.1.md
```diff
@@ -89,14 +89,15 @@ Set the cgroup namespace mode for the container.
...
 to not create CGroups, and thus conflicts with CGroup options
 (**--cgroupns** and **--cgroup-parent**).
 The **no-conmon** option disables a new CGroup only for the **conmon** process.
+The **split** option splits the current cgroup in two sub-cgroups: one for conmon and one for the container payload.
```
mheon, line 100: Can you mention that this conflicts with cgroup-parent

mheon:
Can you add the cgroup-parent validation to `validate()` in `libpod/container_validate.go`?

docs/source/markdown/podman-run.1.md
giuseppe: added to the last version

docs/source/markdown/podman-run.1.md
giuseppe: added

giuseppe:
> Can you add the cgroup-parent validation to `validate()` in `libpod/container_validate.go`?

added and pushed a new version

giuseppe:
ready for review

mheon:
LGTM

goochjj:
LGTM

docs/source/markdown/podman-run.1.md
```diff
@@ -89,14 +89,16 @@ Set the cgroup namespace mode for the container.
...
 
-Default is **enabled**. The **disabled** option will force the container
-to not create CGroups, and thus conflicts with CGroup options
-(**--cgroupns** and **--cgroup-parent**).
+Default is **enabled**.
+
+The **enabled** option will create a new cgroup under the cgroup-parent.
+The **disabled** option will force the container to not create CGroups, and thus conflicts with CGroup options (**--cgroupns** and **--cgroup-parent**).
 The **no-conmon** option disables a new CGroup only for the **conmon** process.
+The **split** option splits the current cgroup in two sub-cgroups: one for conmon and one for the container payload.  It is not possible to set **--cgroup-parent** with **split**.
```
TomSweeneyRedHat, line 101: We really, really, really need to create a conmon man page that we can add a link to for pages like this.  @haircommander any chance of making this happen, even if the page is terribly short and sweet?

This is not a holdback for this review at this time.

haircommander: opened https://github.com/containers/conmon/issues/184

PR #6676, opened 2020-06-18
Account for non-default port number in image name

lsm5:
Previously, if an image was tagged with the format
$REGISTRY:$PORT/$REPO:$TAG,
then `podman images` would display "$PORT/$REPO:$TAG" under the "TAG"
field.

This commit correctly displays $REGISTRY:$PORT/$REPO under the
"REPOSITORY" field while the "TAG" field only displays $TAG.

Fixes: gh#6665

@baude @mheon @rhatdan @vrothberg @jwhonce PTAL

### DIFF
--- a/cmd/podman/images/list.go
+++ b/cmd/podman/images/list.go
@@ -188,7 +188,7 @@ func writeTemplate(imageS []*entities.ImageSummary) error {
 }
 
 func tokenRepoTag(tag string) (string, string) {
-	tokens := strings.SplitN(tag, ":", 2)
+	tokens := strings.Split(tag, ":")
 	switch len(tokens) {
 	case 0:
 		return tag, ""
@@ -196,6 +196,8 @@ func tokenRepoTag(tag string) (string, string) {
 		return tokens[0], ""
 	case 2:
 		return tokens[0], tokens[1]
+	case 3:
+		return tokens[0] + ":" + tokens[1], tokens[2]
 	default:
 		return "<N/A>", ""
 	}


cmd/podman/images/list.go
```diff
@@ -184,14 +184,16 @@ func writeTemplate(imageS []*entities.ImageSummary) error {
...
+	tokens := strings.Split(tag, ":")
 	switch len(tokens) {
 	case 0:
 		return tag, ""
 	case 1:
 		return tokens[0], ""
 	case 2:
 		return tokens[0], tokens[1]
+	case 3:
+		return tokens[0] + ":" + tokens[1], tokens[2]
```
rhatdan, line 196: Is this correct?  To be safe this should be 

return tokens[0] + ":" + tokens[1], strings.Join(tokens[2:],":")

rhatdan: I would figure len == 0 is an error. 1, and 2 are fine and then > 2 would be the default.

lsm5: > Is this correct? To be safe this should be
> 
> return tokens[0] + ":" + tokens[1], strings.Join(tokens[2:],":")

hmm, is that because the tag can themselves contain ":" in them?

rhatdan: I am not sure if they are prevented

lsm5: That fails already:
```
$ ./bin/podman tag docker.io/library/ubuntu foo1:20000/foo2:foo3:foo4
Error: invalid reference format

lsm5 @ nagato : ~/go/src/github.com/containers/libpod(tag-correction) $ ./bin/podman tag docker.io/library/ubuntu foo1:20000/foo2:foo3     
lsm5 @ nagato : ~/go/src/github.com/containers/libpod(tag-correction) $ 
```

rhatdan: Yes I checked the same thing, I guess we can let this go
lgtm

rhatdan:
Looks like tests are failing.

cmd/podman/images/list.go
lsm5: err I made some more changes ... made len == 3 the default and <N/A> for len == 0. PTAL ...

lsm5: went back to the original change itself as lots of failures otherwise.

lsm5:
alright, so I effectively only have a trivial README change but CI is still failing. Probably not my fault. Could someone PTAL ?

mheon:
Can you try rebasing? Lint has been acting up recently, but it seems OK on master

lsm5:
> Can you try rebasing? Lint has been acting up recently, but it seems OK on master

seems like my branch is rebased on the latest..

lsm5:
master is on a2661b1c25e280282d101f24ce478305622da11f and i'm rebased on that as well.

lsm5:
> ```
> [+0086s]  * 9a290a95c "Revert to check if tests pass" ... FAIL
> [+0086s]   - FAIL - does not have a valid DCO
> ```
> 
> ^^^ this commit isn't signed. I guess we can squash them into one?

Thanks!!

lsm5:
hmm, still a bunch of other failures

vrothberg:
I see the login tests fail in other PRs as well and they don't pass locally either ...

vrothberg:
Ha. `htpasswd` dropped from the `registry:2` image

lsm5:
all tests pass. @rhatdan could you PTAL once again for confirmation?

Issue #6664, opened 2020-06-18
APIv2: No events when deleting image


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
When I listen on events through APIv2 and delete image I don't get any event for it. I get it later on when I do some other action, that triggers some event. It feels like events are not flushed/sent, but are only queued.

**Steps to reproduce the issue:**
1. `sudo curl --unix-socket /run/podman/podman.sock http://d/v1.24/events` listen on events
2. `sudo podman run -dit fedora bash`
3. `sudo podman rmi -f eb7134a03cd6` (`eb7134a03cd6` is the fedora image that was used in ^)

**Describe the results you received:**
Checking events (step 1. was running the whole time) I don't see any about the image being removed, just about the container being stopped (but not removed).

```
{"Type":"image","Action":"pull","Actor":{"ID":"","Attributes":{"containerExitCode":"0","image":"","name":""}},"scope":"local","time":1592472737,"timeNano":1592472737248919025}
{"Type":"container","Action":"create","Actor":{"ID":"5f22309892b25090082bb6db24ad937820e6838b813c6453cae87ee299af4ed2","Attributes":{"containerExitCode":"0","image":"registry.fedoraproject.org/fedora:latest","name":"keen_pike"}},"scope":"local","time":1592472737,"timeNano":1592472737274347288}
{"Type":"container","Action":"init","Actor":{"ID":"5f22309892b25090082bb6db24ad937820e6838b813c6453cae87ee299af4ed2","Attributes":{"containerExitCode":"0","image":"registry.fedoraproject.org/fedora:latest","name":"keen_pike"}},"scope":"local","time":1592472737,"timeNano":1592472737698748338}

---------- Now I run step 3. ------------------

{"Type":"container","Action":"start","Actor":{"ID":"5f22309892b25090082bb6db24ad937820e6838b813c6453cae87ee299af4ed2","Attributes":{"containerExitCode":"0","image":"registry.fedoraproject.org/fedora:latest","name":"keen_pike"}},"scope":"local","time":1592472737,"timeNano":1592472737752666694}
[...]

PR #6677, opened 2020-06-18
Poll on events for file reading

baude:
When multiple connections are monitoring events via the remote API, the inotify in the hpcloud library seems unable to consistently send events.  Switching from inotify to poll seems to clear this up.

Fixes: #6664


mheon:
LGTM. Simple fixes are always nice

PR #6679, opened 2020-06-18
system tests: new rm, build tests

edsantiago:
- rm: confirm 'rm' and 'rm -f' on running container

 - build: shotgun test of workdir, cmd, env, labels

### DIFF
--- a/test/system/055-rm.bats
+++ b/test/system/055-rm.bats
@@ -21,6 +21,18 @@ load helpers
     run_podman 125 inspect $rand
 }
 
+@test "podman rm - running container, w/o and w/ force" {
+    run_podman run -d $IMAGE sleep 5
+    cid="$output"
+
+    # rm should fail
+    run_podman 2 rm $cid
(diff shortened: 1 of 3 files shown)


QiWang19:
LGTM.  May need a rebase to get CI regression fixed.

rhatdan:
LGTM
Rebase and we can get this merged, once the tests pass.

edsantiago:
Tests are green. Changes since first submission for review:
* helpers.bash - added code to convert `./bin/podman` to absolute path. This is necessary because one test (and I have more in mind) currently cd's into a temporary directory to test `build` with relative paths. It has the unfortunate side effect of making the podman invocation line unreadable; I will submit a subsequent PR to the log formatter to address this.
* build.bats - prepend 'l' (the letter ell) to the label name, to work around a limitation in `jq` wherein `.foo.1bar` (JSON string beginning with a digit) causes jq to choke. (The label is a random string)

PR #6680, opened 2020-06-18
Fix event cleanup when client closes connection

jwhonce:
* Refactor channel close to happen in code that creates channel
* Add context to allow stopping the publishing of events


mheon:
Fixed by #6838

PR #6684, opened 2020-06-19
CI: force registry:2.6

vrothberg:
Force using the `registry:2.6` image. 2.7 and beyond dropped the
`htpasswd` binary from the rootfs which parts of our CI depends
on.

While this is not a sustainable solution (assuming `htpasswd` is gone
for ever), it unblocks the CI for now.

### DIFF
--- a/hack/podman-registry
+++ b/hack/podman-registry
@@ -7,7 +7,7 @@ ME=$(basename $0)
 ###############################################################################
 # BEGIN defaults
 
-PODMAN_REGISTRY_IMAGE=docker.io/library/registry:2
+PODMAN_REGISTRY_IMAGE=docker.io/library/registry:2.6
 
 PODMAN_REGISTRY_USER=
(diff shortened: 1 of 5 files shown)


vrothberg:
FYI @baude @mheon

rhatdan:
LGTM
Need to figure out a long term solution.

edsantiago:
LGTM. Thank you for tracking that down.

TomSweeneyRedHat:
@cevich FYI

edsantiago:
>assuming htpasswd is gone for ever

FWIW [their own documentation](https://docs.docker.com/registry/deploying/#native-basic-auth) describes the use of `htpasswd` in their container. It is hard to imagine that we are the only ones impacted by this.

mheon:
@vrothberg FYI we probably need a backport of this to the v1.9 branch, in case we ever need to do further backports there

PR #6689, opened 2020-06-19
Advertise Conmon PID via MAINPID

goochjj:
When NOTIFY_SOCKET is passed from systemd, we
send MAINPID=<pid of conmon> so systemd monitors the correct PID.

Imported daemon/ from go-systemd dependency.

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -30,6 +30,7 @@ import (
 	"github.com/containers/libpod/utils"
 	pmount "github.com/containers/storage/pkg/mount"
 	"github.com/coreos/go-systemd/v22/activation"
+	"github.com/coreos/go-systemd/v22/daemon"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux"
 	"github.com/opencontainers/selinux/go-selinux/label"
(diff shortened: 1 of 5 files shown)


libpod/oci_conmon_linux.go
```diff
@@ -37,6 +37,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"k8s.io/client-go/tools/remotecommand"
+	"github.com/coreos/go-systemd/v22/daemon"
```
TomSweeneyRedHat, line 40: @goochjj gating doesn't like the location of this line, it needs to move up in alpha order.  I think to line 32.5

mheon:
Gating is complaining about gofmt.

@giuseppe PTAL

TomSweeneyRedHat:
Once gating and happy tests are complete
LGTM

libpod/oci_conmon_linux.go
goochjj: fixed

vrothberg:
I'll pass it do @giuseppe :^)

goochjj:
I'm not sure how to fix the vendor check.  When I run the go mod commands locally it deletes many many things.

goochjj:
I think I figured it out

vrothberg:
> I'm not sure how to fix the vendor check. When I run the go mod commands locally it deletes many many things.

We require go >= 1.1.3. If you're below that, you can run `make vendor-in-container`

libpod/oci_conmon_linux.go
```diff
@@ -1016,6 +1017,12 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		// conmon not having a pid file is a valid state, so don't set it if we don't have it
 		logrus.Infof("Got Conmon PID as %d", conmonPID)
 		ctr.state.ConmonPID = conmonPID
+		sent, err2 := daemon.SdNotify(false, fmt.Sprintf("MAINPID=%d\n", conmonPID))
+		if err2 != nil {
```
giuseppe, line 1021: sent doesn't seem to be used?

You could write as:
```suggestion
		if _, err := daemon.SdNotify(false, fmt.Sprintf("MAINPID=%d\n", conmonPID)); err != nil {
```

giuseppe:
yes I agree.  We need only for start, then LGTM

goochjj:
> yes I agree. We need only for start, then LGTM

I sent it twice, in case runc or crun passes a different MAINPID.  conmon happens before OCI runtime, start happens after.

Note also this PR is superceded by #6693

Issue #6688, opened 2020-06-19
systemd sd-notify freezes podman


goochjj:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Creating a service using Type=notify freezes everything until the container is ready.

Setting a service as Type=notify sets the NOTIFY_SOCKET, which gets passed through podman properly.  Runc and Crun then proxy that NOTIFY_SOCKET through to the container, indicating the container will signal when ready.  The whole idea being that starting a container does not equal "a container is ready", this initialization could take seconds, 10's of seconds, or minutes.  And it shouldn't matter how long it takes..

The problem is while it's in "starting" status, podman is frozen completely. `podman ps` doesn't return, `podman exec` doesn't work, even `podman info` won't return.  One partially initialized container shouldn't freeze everything, and the lack of exec makes it hard to diagnose what's going on inside the container to resolve the sd-notify issue. `podman stop` and `podman kill` appear to work, but the container is still stuck.

In addition, the MAINPID isn't set right - but we'll come back to that.

**Steps to reproduce the issue:**

1. Create a systemd service. Change type from Type=forking to Type=notify, and remove the 
PIDFile.  Add NotifyAccess=all.
SSCE
```
[Unit]
Description=NotifyTesting
Wants=network.target
After=network-online.target

[Service]
Environment=PODMAN_SYSTEMD_UNIT=%n
SyslogIdentifier=%N
ExecStartPre=-/usr/bin/podman stop %N
ExecStartPre=-/usr/bin/podman rm %N
LogExtraFields=CONTAINER_NAME=%N
ExecStart=/usr/bin/podman --log-level=debug run \
  -d --log-driver=journald \
  --init \
  --cgroups no-conmon \
  --net=host \
  --name %N \
  alpine sleep infinity
ExecStop=/usr/bin/podman stop -t 20 %N
Type=notify
NotifyAccess=all
Restart=on-failure
#Restart=always
RestartSec=30s
TimeoutStartSec=20
TimeoutStopSec=25
#KillMode=none
#Type=forking
#PIDFile=/run/podman-pid-%n
Delegate=yes
Slice=machine.slice

[Install]
WantedBy=multi-user.target default.target
```

2. Run the service.

3. Note the status:
```
[...]

PR #6693, opened 2020-06-19
Implement --sdnotify cmdline option to control sd-notify behavior

goochjj:
--sdnotify conmon-only|container|none

With "conmon-only", we send the MAINPID, and clear the NOTIFY_SOCKET so the OCI
runtime doesn't pass it into the container. We also advertise "ready" when the
OCI runtime finishes to advertise the service as ready.

With "container", we send the MAINPID, and leave the NOTIFY_SOCKET so the OCI
runtime passes it into the container for initialization, and let the container advertise further metadata.

The "none" option does what it's always done in the past - passes NOTIFY_SOCKET
to conmon's process and the OCI runtime processes, and does no manipulation or "help".

This removes the need for hardcoded CID and PID files in the command line, and
the PIDFile directive, as the pid is advertised directly through sd-notify.

Includes #6689 
References #6688

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/containers/common/pkg/auth"
 	"github.com/containers/libpod/v2/cmd/podman/registry"
+	"github.com/containers/libpod/v2/libpod/define"
 	"github.com/spf13/pflag"
 )
 
@@ -393,6 +394,11 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		&cf.RootFS,
 		"rootfs", false,
 		"The first argument is not an image but the rootfs to the exploded container",
+	)
+	createFlags.StringVar(
+		&cf.SdNotifyMode,
+		"sdnotify", define.SdNotifyModeContainer,
+		`control sd-notify behavior ("container"|"conmon"|"ignore")`,
 	)
 	createFlags.StringArrayVar(
 		&cf.SecurityOpt,
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -702,6 +702,17 @@ If specified, the first argument refers to an exploded container on the file sys
 This is useful to run a container without requiring any image management, the rootfs
 of the container is assumed to be managed externally.
 
+**--sdnotify**=**container**|**conmon**|**ignore**
+
+Determines how to use the NOTIFY_SOCKET, as passed with systemd and Type=notify.
+
+Default is **container**, which means allow the OCI runtime to proxy the socket into the
+container to receive ready notification.  Podman will set the MAINPID to conmon's pid.
+The **conmon** option sets MAINPID to conmon's pid, and sends READY when the container
+has started.  The socket is never passed to the runtime or the container.
+The **ignore** option removes NOTIFY_SOCKET from the environment for itself and child processes,
+for the case where some other process above Podman uses NOTIFY_SOCKET and Podman should not use it.
+
 **--seccomp-policy**=*policy*
 
 Specify the policy to select the seccomp profile. If set to *image*, Podman will look for a "io.podman.seccomp.profile" label in the container-image config and use its value as a seccomp profile. Otherwise, Podman will follow the *default* policy by applying the default profile unless specified otherwise via *--security-opt seccomp* as described below.
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -723,6 +723,17 @@ of the container is assumed to be managed externally.
 Note: On **SELinux** systems, the rootfs needs the correct label, which is by default
 **unconfined_u:object_r:container_file_t**.
 
+**--sdnotify**=**container**|**conmon**|**ignore**
+
+Determines how to use the NOTIFY_SOCKET, as passed with systemd and Type=notify.
+
+Default is **container**, which means allow the OCI runtime to proxy the socket into the
+container to receive ready notification.  Podman will set the MAINPID to conmon's pid.
+The **conmon** option sets MAINPID to conmon's pid, and sends READY when the container
+has started.  The socket is never passed to the runtime or the container.
+The **ignore** option removes NOTIFY_SOCKET from the environment for itself and child processes,
+for the case where some other process above Podman uses NOTIFY_SOCKET and Podman should not use it.
+
 **--seccomp-policy**=*policy*
 
 Specify the policy to select the seccomp profile. If set to *image*, Podman will look for a "io.podman.seccomp.profile" label in the container-image config and use its value as a seccomp profile. Otherwise, Podman will follow the *default* policy by applying the default profile unless specified otherwise via *--security-opt seccomp* as described below.
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -24,6 +24,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/mount"
+	"github.com/coreos/go-systemd/v22/daemon"
 	securejoin "github.com/cyphar/filepath-securejoin"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
@@ -1191,6 +1192,19 @@ func (c *Container) start() error {
 	logrus.Debugf("Started container %s", c.ID())
 
 	c.state.State = define.ContainerStateRunning
+
+	if c.config.SdNotifyMode != define.SdNotifyModeIgnore {
+		payload := fmt.Sprintf("MAINPID=%d", c.state.ConmonPID)
+		if c.config.SdNotifyMode == define.SdNotifyModeConmon {
+			payload += "\n"
+			payload += daemon.SdNotifyReady
+		}
+		if sent, err := daemon.SdNotify(false, payload); err != nil {
+			logrus.Errorf("Error notifying systemd of Conmon PID: %s", err.Error())
+		} else if sent {
+			logrus.Debugf("Notify sent successfully")
+		}
+	}
 
 	if c.config.HealthCheckConfig != nil {
 		if err := c.updateHealthStatus(define.HealthCheckStarting); err != nil {
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -30,6 +30,7 @@ import (
 	"github.com/containers/libpod/v2/utils"
 	pmount "github.com/containers/storage/pkg/mount"
 	"github.com/coreos/go-systemd/v22/activation"
+	"github.com/coreos/go-systemd/v22/daemon"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux"
 	"github.com/opencontainers/selinux/go-selinux/label"
@@ -365,8 +366,10 @@ func (r *ConmonOCIRuntime) StartContainer(ctr *Container) error {
 		return err
 	}
 	env := []string{fmt.Sprintf("XDG_RUNTIME_DIR=%s", runtimeDir)}
-	if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
-		env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
+	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
+		if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
+			env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
+		}
 	}
 	if path, ok := os.LookupEnv("PATH"); ok {
 		env = append(env, fmt.Sprintf("PATH=%s", path))
@@ -887,6 +890,12 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		}
 	}
 
+	if ctr.config.SdNotifyMode == define.SdNotifyModeIgnore {
+		if err := os.Unsetenv("NOTIFY_SOCKET"); err != nil {
+			logrus.Warnf("Error unsetting NOTIFY_SOCKET %s", err.Error())
+		}
+	}
+
 	args := r.sharedConmonArgs(ctr, ctr.ID(), ctr.bundlePath(), filepath.Join(ctr.state.RunDir, "pidfile"), ctr.LogPath(), r.exitsDir, ociLog, ctr.LogDriver(), logTag)
 
 	if ctr.config.Spec.Process.Terminal {
@@ -940,7 +949,7 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 	}
 
 	// 0, 1 and 2 are stdin, stdout and stderr
-	conmonEnv, envFiles, err := r.configureConmonEnv(runtimeDir)
+	conmonEnv, envFiles, err := r.configureConmonEnv(ctr, runtimeDir)
 	if err != nil {
 		return err
 	}
@@ -1034,6 +1043,13 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		// conmon not having a pid file is a valid state, so don't set it if we don't have it
 		logrus.Infof("Got Conmon PID as %d", conmonPID)
 		ctr.state.ConmonPID = conmonPID
+		if ctr.config.SdNotifyMode != define.SdNotifyModeIgnore {
+			if sent, err := daemon.SdNotify(false, fmt.Sprintf("MAINPID=%d", conmonPID)); err != nil {
+				logrus.Errorf("Error notifying systemd of Conmon PID: %s", err.Error())
+			} else if sent {
+				logrus.Debugf("Notify MAINPID sent successfully")
+			}
+		}
 	}
 
 	if ctr.config.PreserveFDs > 0 {
@@ -1137,7 +1153,7 @@ func prepareProcessExec(c *Container, cmd, env []string, tty bool, cwd, user, se
 
 // configureConmonEnv gets the environment values to add to conmon's exec struct
 // TODO this may want to be less hardcoded/more configurable in the future
-func (r *ConmonOCIRuntime) configureConmonEnv(runtimeDir string) ([]string, []*os.File, error) {
+func (r *ConmonOCIRuntime) configureConmonEnv(ctr *Container, runtimeDir string) ([]string, []*os.File, error) {
 	env := make([]string, 0, 6)
 	env = append(env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", runtimeDir))
 	env = append(env, fmt.Sprintf("_CONTAINERS_USERNS_CONFIGURED=%s", os.Getenv("_CONTAINERS_USERNS_CONFIGURED")))
@@ -1149,8 +1165,10 @@ func (r *ConmonOCIRuntime) configureConmonEnv(runtimeDir string) ([]string, []*o
 	env = append(env, fmt.Sprintf("HOME=%s", home))
 
 	extraFiles := make([]*os.File, 0)
-	if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
-		env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
+	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
+		if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
+			env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
+		}
 	}
 	if !r.sdNotify {
 		if listenfds, ok := os.LookupEnv("LISTEN_FDS"); ok {
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	"github.com/containers/common/pkg/config"
@@ -22,6 +23,10 @@ import (
 )
 
 // Runtime Creation Options
+var (
+	// SdNotifyModeValues describes the only values that SdNotifyMode can be
+	SdNotifyModeValues = []string{define.SdNotifyModeContainer, define.SdNotifyModeConmon, define.SdNotifyModeIgnore}
+)
 
 // WithStorageConfig uses the given configuration to set up container storage.
 // If this is not specified, the system default configuration will be used
@@ -546,6 +551,23 @@ func WithSystemd() CtrCreateOption {
 		}
 
 		ctr.config.Systemd = true
+		return nil
+	}
+}
+
+// WithSdNotifyMode sets the sd-notify method
+func WithSdNotifyMode(mode string) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+
+		// verify values
+		if len(mode) > 0 && !util.StringInSlice(strings.ToLower(mode), SdNotifyModeValues) {
+			return errors.Wrapf(define.ErrInvalidArg, "--sdnotify values must be one of %q", strings.Join(SdNotifyModeValues, ", "))
+		}
+
+		ctr.config.SdNotifyMode = mode
 		return nil
 	}
 }
--- a/pkg/specgen/container_validate.go
+++ b/pkg/specgen/container_validate.go
@@ -3,6 +3,7 @@ package specgen
 import (
 	"strings"
 
+	"github.com/containers/libpod/v2/libpod/define"
 	"github.com/containers/libpod/v2/pkg/rootless"
 	"github.com/containers/libpod/v2/pkg/util"
 	"github.com/pkg/errors"
@@ -13,6 +14,8 @@ var (
 	ErrInvalidSpecConfig = errors.New("invalid configuration")
 	// SystemDValues describes the only values that SystemD can be
 	SystemDValues = []string{"true", "false", "always"}
+	// SdNotifyModeValues describes the only values that SdNotifyMode can be
+	SdNotifyModeValues = []string{define.SdNotifyModeContainer, define.SdNotifyModeConmon, define.SdNotifyModeIgnore}
 	// ImageVolumeModeValues describes the only values that ImageVolumeMode can be
 	ImageVolumeModeValues = []string{"ignore", "tmpfs", "anonymous"}
 )
@@ -39,6 +42,10 @@ func (s *SpecGenerator) Validate() error {
 	// systemd values must be true, false, or always
 	if len(s.ContainerBasicConfig.Systemd) > 0 && !util.StringInSlice(strings.ToLower(s.ContainerBasicConfig.Systemd), SystemDValues) {
 		return errors.Wrapf(ErrInvalidSpecConfig, "--systemd values must be one of %q", strings.Join(SystemDValues, ", "))
+	}
+	// sdnotify values must be container, conmon, or ignore
+	if len(s.ContainerBasicConfig.SdNotifyMode) > 0 && !util.StringInSlice(strings.ToLower(s.ContainerBasicConfig.SdNotifyMode), SdNotifyModeValues) {
+		return errors.Wrapf(ErrInvalidSpecConfig, "--sdnotify values must be one of %q", strings.Join(SdNotifyModeValues, ", "))
 	}
 
 	//
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -81,6 +81,7 @@ type ContainerCLIOpts struct {
 	Rm                bool
 	RootFS            bool
 	SecurityOpt       []string
+	SdNotifyMode      string
 	ShmSize           string
 	StopSignal        string
 	StopTimeout       uint
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -443,6 +443,7 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	}
 
 	s.Systemd = c.Systemd
+	s.SdNotifyMode = c.SdNotifyMode
 	if s.ResourceLimits == nil {
 		s.ResourceLimits = &specs.LinuxResources{}
 	}
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -414,6 +414,8 @@ type ContainerConfig struct {
 	// sharing kernel namespaces in a pod
 	IsInfra bool `json:"pause"`
 
+	// SdNotifyMode tells libpod what to do with a NOTIFY_SOCKET if passed
+	SdNotifyMode string `json:"sdnotifyMode,omitempty"`
 	// Systemd tells libpod to setup the container in systemd mode
 	Systemd bool `json:"systemd"`
 
--- a/libpod/define/config.go
+++ b/libpod/define/config.go
@@ -75,3 +75,10 @@ const JSONLogging = "json-file"
 
 // NoLogging is the string conmon expects when specifying to use no log driver whatsoever
 const NoLogging = "none"
+
+// Strings used for --sdnotify option to podman
+const (
+	SdNotifyModeContainer = "container"
+	SdNotifyModeConmon    = "conmon"
+	SdNotifyModeIgnore    = "ignore"
+)
--- a/libpod/oci_conmon_exec_linux.go
+++ b/libpod/oci_conmon_exec_linux.go
@@ -444,7 +444,7 @@ func (r *ConmonOCIRuntime) startExec(c *Container, sessionID string, options *Ex
 	// 	}
 	// }
 
-	conmonEnv, extraFiles, err := r.configureConmonEnv(runtimeDir)
+	conmonEnv, extraFiles, err := r.configureConmonEnv(c, runtimeDir)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/specgen/generate/container_create.go
+++ b/pkg/specgen/generate/container_create.go
@@ -175,6 +175,10 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
 
 		options = append(options, libpod.WithSystemd())
 	}
+	if len(s.SdNotifyMode) > 0 {
+		options = append(options, libpod.WithSdNotifyMode(s.SdNotifyMode))
+	}
+
 	if len(s.Name) > 0 {
 		logrus.Debugf("setting container name %s", s.Name)
 		options = append(options, libpod.WithName(s.Name))
--- a/pkg/specgen/specgen.go
+++ b/pkg/specgen/specgen.go
@@ -107,6 +107,11 @@ type ContainerBasicConfig struct {
 	// If not specified, "false" will be assumed.
 	// Optional.
 	Systemd string `json:"systemd,omitempty"`
+	// Determine how to handle the NOTIFY_SOCKET - do we participate or pass it through
+	// "container" - let the OCI runtime deal with it, advertise conmon's MAINPID
+	// "conmon-only" - advertise conmon's MAINPID, send READY when started, don't pass to OCI
+	// "ignore" - unset NOTIFY_SOCKET
+	SdNotifyMode string `json:"sdnotifyMode,omitempty"`
 	// Namespace is the libpod namespace the container will be placed in.
 	// Optional.
 	Namespace string `json:"namespace,omitempty"`
--- a/test/system/260-sdnotify.bats
+++ b/test/system/260-sdnotify.bats
@@ -1,0 +1,142 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# Tests for systemd sdnotify
+#
+
+load helpers
+
+# Shared throughout this module: PID of socat process, and path to its log
+_SOCAT_PID=
+_SOCAT_LOG=
+
+function setup() {
+    skip_if_remote
+
+    # TODO: remove this once CI systems have newer crun and container-selinux
+    skip "TEMPORARY SKIP - until CI systems get new crun, container-selinux"
+
+    basic_setup
+}
+
+function teardown() {
+    unset NOTIFY_SOCKET
+
+    _stop_socat
+
+    basic_teardown
+}
+
+###############################################################################
+# BEGIN helpers
+
+# Run socat process on a socket, logging to well-known path. Each received
+# packet is logged with a newline appended, for ease of parsing the log file.
+function _start_socat() {
+    _SOCAT_LOG="$PODMAN_TMPDIR/socat.log"
+
+    rm -f $_SOCAT_LOG
+    socat unix-recvfrom:"$NOTIFY_SOCKET",fork \
+          system:"(cat;echo) >> $_SOCAT_LOG" &
+    _SOCAT_PID=$!
+}
+
+# Stop the socat background process and clean up logs
+function _stop_socat() {
+    if [[ -n "$_SOCAT_PID" ]]; then
+        kill $_SOCAT_PID
+    fi
+    _SOCAT_PID=
+
+    if [[ -n "$_SOCAT_LOG" ]]; then
+        rm -f $_SOCAT_LOG
+    fi
+}
+
+# Check that MAINPID=xxxxx points to a running conmon process
+function _assert_mainpid_is_conmon() {
+    local mainpid=$(expr "$1" : "MAINPID=\([0-9]\+\)")
+    test -n "$mainpid" || die "Could not parse '$1' as 'MAINPID=nnnn'"
+
+    test -d /proc/$mainpid || die "sdnotify MAINPID=$mainpid - but /proc/$mainpid does not exist"
+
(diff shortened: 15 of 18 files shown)


goochjj:
The more I think about this I'm not sure "none" is a valid option, though it was essentially requested...  Since "none" returns a broken situation. (MAINPID never advertised properly)... Unless there's a case when runc or crun advertises the correct MAINPID.

mheon:
I think `none` is "I am not being managed by systemd, please ignore" - we've seen cases where Podman is run as a child of something actually managed by systemd, inherits NOTIFY_SOCKET from the parent (arguably a bug in the parent, but still), and then blocks forever as it tries to notify on an already-used socket.

goochjj:
> I think `none` is "I am not being managed by systemd, please ignore" - we've seen cases where Podman is run as a child of something actually managed by systemd, inherits NOTIFY_SOCKET from the parent (arguably a bug in the parent, but still), and then blocks forever as it tries to notify on an already-used socket.

That's a reasonable interpretation.  In that case I should probably block NOTIFY_SOCKET from being passed down to the OCI runtime and conmon... correct?  That way everything will proceed as if the NOTIFY_SOCKET isn't there.

vs. what it does now, which means "libpod does nothing with NOTIFY_SOCKET other than pass it on", which is pretty similar to "container".

goochjj:
> > I think `none` is "I am not being managed by systemd, please ignore" - we've seen cases where Podman is run as a child of something actually managed by systemd, inherits NOTIFY_SOCKET from the parent (arguably a bug in the parent, but still), and then blocks forever as it tries to notify on an already-used socket.
> 
> That's a reasonable interpretation. In that case I should probably block NOTIFY_SOCKET from being passed down to the OCI runtime and conmon... correct? That way everything will proceed as if the NOTIFY_SOCKET isn't there.
> 
> vs. what it does now, which means "libpod does nothing with NOTIFY_SOCKET other than pass it on", which is pretty similar to "container".

In which case I think "ignore" is a better description.

mheon:
Ignore works for me - and yes, we should not forward the socket in that case

docs/source/markdown/podman-create.1.md
```diff
@@ -697,6 +697,17 @@ If specified, the first argument refers to an exploded container on the file sys
...
+**--sdnotify**=**container**|**conmon-only**|**ignore**
+
+Determines how to use the NOTIFY_SOCKET, as passed with systemd and Type=notify.
+
+Default is **container**, which means allow the OCI runtime to proxy the socket into the
+container to receive ready notification.  Podman will set the MAINPID to conmon's pid.
+The **conmon-only** option sets MAINPID to conmon's pid, and sends READY when the container
+has started.  The socket is never passed to the runtime or the container.
+The **ignore** option removes NOTIFY_SOCKET from the environment for itself and child processes,
+for the case where some other service above us uses NOTIFY_SOCKET but we shouldn't mess with it.
```
TomSweeneyRedHat, line 709:
```suggestion
for the case where some other service above Podman uses NOTIFY_SOCKET and Podman should not revise it.
```

docs/source/markdown/podman-run.1.md
```diff
@@ -714,6 +714,17 @@ of the container is assumed to be managed externally.
...
+**--sdnotify**=**container**|**conmon-only**|**ignore**
+
+Determines how to use the NOTIFY_SOCKET, as passed with systemd and Type=notify.
+
+Default is **container**, which means allow the OCI runtime to proxy the socket into the
+container to receive ready notification.  Podman will set the MAINPID to conmon's pid.
+The **conmon-only** option sets MAINPID to conmon's pid, and sends READY when the container
+has started.  The socket is never passed to the runtime or the container.
+The **ignore** option removes NOTIFY_SOCKET from the environment for itself and child processes,
+for the case where some other service above us uses NOTIFY_SOCKET but we shouldn't mess with it.
```
TomSweeneyRedHat, line 726: see prior

docs/source/markdown/podman-create.1.md
goochjj: revised

rhatdan:
Doesn't Podman have to respond to the ignore and set the ready state for systemd if it sees the flag and is set to ignore?  IE Systemd will never mark the unit file as ready unless someone says it is ready.

goochjj:
> Doesn't Podman have to respond to the ignore and set the ready state for systemd if it sees the flag and is set to ignore? IE Systemd will never mark the unit file as ready unless someone says it is ready.

That's true - however, currently Podman doesn't respond at all.  Runc/crun do.  Ignore is only relevant when some other service is the responsible party.


The default is "container", which is the current behavior, which would mean Runc/Crun need to respond.  This PR just makes sure the correct MAINPID is set during the conversation.

conmon-only means Podman is the responsible party - the one responsible for sending READY=1, and keeps runc/crun out of the loop.

ignore would only be used if some other service is calling podman. I.e. some service that is sd-notify enabled, and therefore they are the responsible party.  Contrived example:
```
#!/bin/bash

#I'm a service
systemd-notify --status="Starting up"

#Spawn some containers
podman run --sdnotify ignore --rm -d alpine sleep infinity
podman run --sdnotify ignore --rm -d alpine sleep infinity
podman run --sdnotify ignore --rm -d alpine sleep infinity

#Do other stuff
sleep 1

#And we're good
systemd-notify --ready
```

Right now, each podman would see the NOTIFY_SOCKET set and assume that it's responsible for telling systemd when it's ready... Which is incorrect, the overall service is the shell script. 

The other alternative is people explicitly changing the podman lines to do:
```
env -u NOTIFY_SOCKET podman run --rm -d alpine sleep infinity
```

As that's really what ignore is doing.

cmd/podman/common/create.go
```diff
@@ -393,6 +393,11 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"rootfs", false,
 		"The first argument is not an image but the rootfs to the exploded container",
 	)
+	createFlags.StringVar(
+		&cf.SdNotifyMode,
+		"sdnotify", "container",
```
rhatdan, line 398: Should probably be in containers.conf github.com/containers/common/pkg/config.  But this is fine for now.

rhatdan: These should become constants not magic strings.

cmd/podman/common/create.go
```diff
@@ -393,6 +393,11 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
...
+		`control sd-notify behavior ("container"|"conmon-only"|"ignore")`,
```
rhatdan, line 399: Change conmon-only to conmon.  The -only does not add any value, and I am not a big fan of options with "dashes" in them.

vrothberg:
> This removes the need for hardcoded CID and PID files in the command line, and
the PIDFile directive, as the pid is advertised directly through sd-notify.

Can you elaborate why this would remove the CID file? I think that's still needed to cleanly stop/remove a container in a unit.

vrothberg:
Reviewing now. @giuseppe PTAL as well

libpod/container_internal.go
```diff
@@ -1166,6 +1167,18 @@ func (c *Container) start() error {
 
 	c.state.State = define.ContainerStateRunning
 
+	if c.config.SdNotifyMode != "ignore" {
+		payload := fmt.Sprintf("MAINPID=%d\n", c.state.ConmonPID)
+		if c.config.SdNotifyMode == "conmon-only" {
```
vrothberg, line 1172: As @rhatdan mentioned in another comment. Using constants would be good. I suggest putting these consts into `libpod/define`, so we can use it in `cmd/podman` without leaking low-level dependencies into the client.

libpod/options.go
```diff
@@ -550,6 +550,18 @@ func WithSystemd() CtrCreateOption {
 	}
 }
 
+// WithSystemd turns on systemd mode in the container
+func WithSdNotifyMode(mode string) CtrCreateOption {
+	return func(ctr *Container) error {
```
vrothberg, line 555: I think we should validate `mode` here and make sure that the value is supported.

mheon: Agree

mheon:
Once comments are addressed, this LGTM

goochjj:
> > This removes the need for hardcoded CID and PID files in the command line, and
> > the PIDFile directive, as the pid is advertised directly through sd-notify.
> 
> Can you elaborate why this would remove the CID file? I think that's still needed to cleanly stop/remove a container in a unit.

CID file can still be used, it's just SystemD doesn't need it so it doesn't need to be hanging out in /run/%n-cid, it'll revert to being in userdata/

PID file can still be used, it's just SystemD doesn't need it so it doesn't need to be hanging out in /run/%n.pid, it'll revert to being in userdata/.... And we don't need to specify it in PIDFile because systemd will receive the appropriate PID via MAINPID= w/ Type=notify.

If for whatever reason people still want to use Type=forking, they can, they'll need to keep the PIDFile directive and put it somewhere it can be found... Otherwise Type=notify just makes things more clean and less hardcoded.

goochjj:
> Oh man, I really like this PR. Thanks a ton!
> 
> Just a minor comment regarding input validation. @edsantiago , could you have a look as well? I wonder if you have a suggestion how we could test this in CI.

I too was struggling with this.  I can envision the tests - as I did the testing manually, but it's going to require forking which is currently beyond my ability to do in golang. 

We'd need a thread that opens a UNIX dgram socket and captures input.  Perhaps (tmpdir)/notify.

1) Ensure podman ignores the socket
  Create notify socket
  Monitor notify socket
  set env NOTIFY_SOCKET
  Spawn podman run --rm --sd-notify=ignore alpine ls
  Verify podman returns exit 0
  Verify notify socket received absolutely no data.

2) Verify sd-notify=conmon (heres where it starts to get tricky)
  Create notify socket
  set env NOTIFY_SOCKET
  Proc1) Monitor notify socket
  Proc2) Spawn podman run --rm --sd-notify=conmon alpine ls
  Proc1) wait for receipt of MAINPID=(pid) and verify it points to conmon, descendent of Proc2
  Proc1) Wait for receipt of second MAINPID= and READY=1
  Proc2) verify exit 0

3) Verify sd-notify=container
  Create notify socket
  set env NOTIFY_SOCKET
  Proc1) Monitor notify socket
  Proc2) Spawn podman run --rm --sd-notify=container alpine ls
  Proc1) wait for receipt of MAINPID=(pid) and verify it points to conmon, descendent of Proc2
  Proc1) sleep 3 seconds
  Proc1) Find the socket - /run/runc/(CID)/notify/notify.sock or /run/crun/(CID)/notify/notify - and send READY=1 to it
  Proc1) Verify you get back MAINPID= and READY=1
  Proc2) verify exit 0

  If podman exits in Proc2 before we send the READY in Proc1, then runc/crun didn't wait for a notification.

4) Test that not specifying sd-notify does the same thing as 3
5) Test that not setting NOTIFY_SOCKET does the same thing as 1

I'm not sure how to coordinate the test between two threads like that in Go.

libpod/options.go
goochjj: @vrothberg @mheon 
Validation already happened in pkg/specgen/container_validate.go

I used the systemd option as a model - that seemed to be where the validation belonged. (i.e. see WithSystemd() directly above this hunk)

goochjj: I can add it again if you want.

pkg/specgen/container_validate.go
```diff
@@ -13,6 +13,8 @@ var (
 	ErrInvalidSpecConfig = errors.New("invalid configuration")
 	// SystemDValues describes the only values that SystemD can be
 	SystemDValues = []string{"true", "false", "always"}
+	// SdNotifyModeValues describes the only values that SdNotifyMode can be
+	SdNotifyModeValues = []string{"container", "conmon-only", "ignore"}
```
goochjj, line 17: Note: this pkg doesn't have access to libpod/define - so I can't use the magic string replacements

mheon: I think importing libpod/define here should be OK - we don't want to drag in libpod itself, but libpod/define is specifically structured to be just constants/struct definitions and very few external dependencies.

goochjj:
All comments should be resolved at this point. PTAL!

libpod/container_internal.go
```diff
@@ -1166,6 +1167,18 @@ func (c *Container) start() error {
 
 	c.state.State = define.ContainerStateRunning
 
+	if c.config.SdNotifyMode != define.SdNotifyModeIgnore {
+		payload := fmt.Sprintf("MAINPID=%d\n", c.state.ConmonPID)
+		if c.config.SdNotifyMode == define.SdNotifyModeConmon {
+			payload += "READY=1\n"
```
edsantiago, line 1173: Apologies if this is a dup - I think github ate my earlier attempt to comment here.

Is there a reason to use hardcoded `READY=1` instead of `SdNotifyReady` from `go-systemd`?

Also, are you absolutely certain about the newline? It's been a long time since I played with sdnotify but IIRC there is no trailing newline. I've confirmed this just now by playing with `systemd-notify` and `socat`

goochjj: It works both ways - so I suppose it shouldn't be \n terminated then...  Resolved.

edsantiago:
I'm having a lot of trouble with `sdnotify=container`. It is very possible that the fault is mine, because it's been many months since I last fiddled with sdnotify.

I think someone/something (crun?) is intercepting `NOTIFY_SOCKET`. Note these two commands, differing only in the setting of that envariable:
```console
$ env -u NOTIFY_SOCKET ./bin/podman run --privileged --rm -v /tmp/podman-sdnotify:/tmp/podman-sdnotify:Z alpine ls -l /tmp
total 0
srwxrwxr-x    1 root     root             0 Jun 22 17:24 podman-sdnotify
$ NOTIFY_SOCKET=/tmp/podman-sdnotify ./bin/podman run --privileged --rm -v /tmp/podman-sdnotify:/tmp/podman-sdnotify:Z alpine ls -l /tmp
total 0
srwxrwxrwx    1 root     root             0 Jun 22 17:36 notify
```
That is: when `NOTIFY_SOCKET` is unset, the (existing) socket is bind-mounted just fine in the container. When it is **set**, the bind mount doesn't work, and there's a socket called `notify`.

The context for this is that I was trying:
```console
$ socat unix-recv:/tmp/podman-sdnotify -

[in another terminal]
$ ./bin/podman run --privileged --rm --sdnotify=container -v /tmp/podman-sdnotify:/tmp/podman-sdnotify:Z fedora systemd-notify --ready
Failed to notify init system: No such file or directory
```
I have a vague memory that this was intentional, but don't remember the details. @giuseppe perhaps you remember?

goochjj:
> I'm having a lot of trouble with `sdnotify=container`. It is very possible that the fault is mine, because it's been many months since I last fiddled with sdnotify.
> 
> I think someone/something (crun?) is intercepting `NOTIFY_SOCKET`. Note these two commands, differing only in the setting of that envariable:
> 
> ```
> $ env -u NOTIFY_SOCKET ./bin/podman run --privileged --rm -v /tmp/podman-sdnotify:/tmp/podman-sdnotify:Z alpine ls -l /tmp
> total 0
> srwxrwxr-x    1 root     root             0 Jun 22 17:24 podman-sdnotify
> $ NOTIFY_SOCKET=/tmp/podman-sdnotify ./bin/podman run --privileged --rm -v /tmp/podman-sdnotify:/tmp/podman-sdnotify:Z alpine ls -l /tmp
> total 0
> srwxrwxrwx    1 root     root             0 Jun 22 17:36 notify
> ```
> 
> That is: when `NOTIFY_SOCKET` is unset, the (existing) socket is bind-mounted just fine in the container. When it is **set**, the bind mount doesn't work, and there's a socket called `notify`.
> 
> The context for this is that I was trying:
> 
> ```
> $ socat unix-recv:/tmp/podman-sdnotify -
> 
> [in another terminal]
> $ ./bin/podman run --privileged --rm --sdnotify=container -v /tmp/podman-sdnotify:/tmp/podman-sdnotify:Z fedora systemd-notify --ready
> Failed to notify init system: No such file or directory
> ```
> 
> I have a vague memory that this was intentional, but don't remember the details. @giuseppe perhaps you remember?

runc/crun take care of proxying the socket. In fact they don't bind mount it, they create a unix-datagram inside the container, and write to the one outside.

I am attempting to decipher goroutines so I can write a proper test in e2e...

Basically though:
T1: `socat unix-recvfrom:/tmp/podman-sdnotify,fork -`
T2: `env NOTIFY_SOCKET=/tmp/podman-sdnotify ./bin/podman run --rm --sdnotify=container fedora systemd-notify --ready`

T1 shows
MAINPID=22842READY=1
MAINPID=22854
MAINPID=22842

This illustrates the sequence:
  podman now sends 22842 (conmon) PID during oci_conmon_linux
  runc/crun passes READY=1 back from systemd-notify
  runc/crun passes MAINPID=22854 back from systemd-notify (which we can't monitor)
  podman sends 22842 from container start - to overwrite the wrong PID

edsantiago:
```
 MAINPID=22842READY=1
```
How are you getting this?! My entire rathole stemmed from my trying to do exactly that. I cannot get it to work, either root or rootless: `systemd-notify` in the container fails as I described above. I get `MAINPID=xxx` in the `socat` window, but never `READY=1`

edsantiago:
I'm on Fedora 32 with crun

goochjj:
correction - 

When NOTIFY_SOCKET is passed,
1) runc will create a socket in /run/runc/CID/notify/notify.sock, and then add a bind mount to /run/systemd/notify.sock (or similar).  This works as I illustrated
2) crun is creating a socket in /run/crun/CID/notify/notify.sock, but never adds the bind mount, so it crashes spectacularly.

Note that's not a problem with this PR, it's crun missing a bind mount - try on podman pre-this-PR without a sdnotify command line option - that fails too.

Just using your technique of adding the bind mount won't work - because runc and crun are explicitly blocking until THEIR socket receives a READY=1.

goochjj:
> ```
>  MAINPID=22842READY=1
> ```
> 
> How are you getting this?! My entire rathole stemmed from my trying to do exactly that. I cannot get it to work, either root or rootless: `systemd-notify` in the container fails as I described above. I get `MAINPID=xxx` in the `socat` window, but never `READY=1`

Magic!

(Or, photon, cgroups 1 + runc)  :-D

goochjj:
> correction -
> 
> When NOTIFY_SOCKET is passed,
> 
> 1. runc will create a socket in /run/runc/CID/notify/notify.sock, and then add a bind mount to /run/systemd/notify.sock (or similar).  This works as I illustrated
> 2. crun is creating a socket in /run/crun/CID/notify/notify.sock, but never adds the bind mount, so it crashes spectacularly.
> 
> Note that's not a problem with this PR, it's crun missing a bind mount - try on podman pre-this-PR without a sdnotify command line option - that fails too.
> 
> Just using your technique of adding the bind mount won't work - because runc and crun are explicitly blocking until THEIR socket receives a READY=1.

I am looking in config.json for the bind mount - which probably isn't a good test if crun is doing it after the spec is loaded. I suppose I need to inspect the mount namespace

edsantiago:
Phew - thank you. Ok, so it's `crun` then. That sure was a lot of time wasted this morning, but I'm glad the issue is identified. Please ignore my bind-mount comment, that was just a red herring, me trying to track down the root cause.

If `crun` is renaming the socket, shouldn't it reset `NOTIFY_SOCKET` in the container? I ran `printenv` and still see `NOTIFY_SOCKET=/tmp/podman-sdnotify` in the container

goochjj:
> Phew - thank you. Ok, so it's `crun` then. That sure was a lot of time wasted this morning, but I'm glad the issue is identified. Please ignore my bind-mount comment, that was just a red herring, me trying to track down the root cause.
> 
> If `crun` is renaming the socket, shouldn't it reset `NOTIFY_SOCKET` in the container? I ran `printenv` and still see `NOTIFY_SOCKET=/tmp/podman-sdnotify` in the container

Yep I agree - I'm tracking it down

goochjj:
> > Phew - thank you. Ok, so it's `crun` then. That sure was a lot of time wasted this morning, but I'm glad the issue is identified. Please ignore my bind-mount comment, that was just a red herring, me trying to track down the root cause.
> > If `crun` is renaming the socket, shouldn't it reset `NOTIFY_SOCKET` in the container? I ran `printenv` and still see `NOTIFY_SOCKET=/tmp/podman-sdnotify` in the container
> 
> Yep I agree - I'm tracking it down

yeah it's the env variable that isn't being set right.

crun is mounting the socket at /run/notify.

`env NOTIFY_SOCKET=/run/podman-sdnotify /usr/bin/podman --runtime /opt/podman/bin/crun run --rm  -it fedora sh -c 'printenv NOTIFY_SOCKET;ls -ld /run/notify; env NOTIFY_SOCKET=/run/notify systemd-notify --ready'`

I'll see if I can find that in crun

goochjj:
> > > Phew - thank you. Ok, so it's `crun` then. That sure was a lot of time wasted this morning, but I'm glad the issue is identified. Please ignore my bind-mount comment, that was just a red herring, me trying to track down the root cause.
> > > If `crun` is renaming the socket, shouldn't it reset `NOTIFY_SOCKET` in the container? I ran `printenv` and still see `NOTIFY_SOCKET=/tmp/podman-sdnotify` in the container
> > 
> > 
> > Yep I agree - I'm tracking it down
> 
> yeah it's the env variable that isn't being set right.
> 
> crun is mounting the socket at /run/notify.
> 
> `env NOTIFY_SOCKET=/run/podman-sdnotify /usr/bin/podman --runtime /opt/podman/bin/crun run --rm -it fedora sh -c 'printenv NOTIFY_SOCKET;ls -ld /run/notify; env NOTIFY_SOCKET=/run/notify systemd-notify --ready'`
> 
> I'll see if I can find that in crun

Heh, this is weird.  It's apparently assuming the socket name is "notify".

When I pass /run/podman-sdnotify, it's mounted as /run/notify
When I pass /tmp/podman-sdnotify, it's mounted as /tmp/notify

I'm assuming this has just been working, because systemd's socket filename is notify. sooooo

giuseppe:
> > 
> 
> Heh, this is weird. It's apparently assuming the socket name is "notify".
> 
> When I pass /run/podman-sdnotify, it's mounted as /run/notify
> When I pass /tmp/podman-sdnotify, it's mounted as /tmp/notify
> 
> I'm assuming this has just been working, because systemd's socket filename is notify. sooooo

both crun and runc wraps the NOTIFY_SOCKET logic and expose a different socket to the container.

When the container notifies it is ready, crun/runc reads the notification on the socket they've created and propagate it back to systemd on the correct socket.

This is to address two issues:

- workaround a systemd notify socket race where the sender of the notification cannot be correctly found with short lived processes if it is not directly monitored by systemd or in a different cgroup.

- avoid that the container communicates directly with systemd and could send some weird MAINPID= value.  So both crun/runc filter anything received from the container except `READY=1`.  I think systemd-nspawn does the same thing.

goochjj:
> > > 
> > 
> > 
> > Heh, this is weird. It's apparently assuming the socket name is "notify".
> > When I pass /run/podman-sdnotify, it's mounted as /run/notify
> > When I pass /tmp/podman-sdnotify, it's mounted as /tmp/notify
> > I'm assuming this has just been working, because systemd's socket filename is notify. sooooo
> 
> both crun and runc wrap the NOTIFY_SOCKET logic and expose a different socket to the container.
> 
> When the container notifies it is ready, crun/runc reads the notification on the socket they've created and propagate it back to systemd on the correct socket.
> 
> This is to address two issues:
> 
> * workaround a systemd notify socket race where the sender of the notification cannot be correctly found with short lived processes if it is not directly monitored by systemd or in a different cgroup.
> * avoid that the container communicates directly with systemd and could send some weird MAINPID= value.  So both crun/runc filter anything received from the container except `READY=1`.  I think systemd-nspawn does the same thing.

I have no problem with any of that, except crun is doing it wrong.

When NOTIFY_SOCKET=/run/podman-sdnotify is passed,
CRUN creates /run/crun/CID/notify/notify.sock
CRUN bind mounts /run/crun/CID/notify/notify.sock to /run/notify
CRUN starts the container with NOTIFY_SOCKET=/run/podman-sdnotify

And it doesn't work.

Bind mount appears to be mounting to $(dirname $NOTIFY_SOCKET)/notify but passes the NOTIFY_SOCKET as is.

So, runc works regardless of what NOTIFY_SOCKET is
crun only works if the socket name is notify.  i.e. /run/systemd/notify, /run/notify, /tmp/notify, /tmp/some/path/here/notify, what have you.

Since this isn't related to my PR, I don't think it should be a blocker on this issue - if it needs to be corrected there'd need to be an issue in the crun repo.

goochjj:
Could use some feedback on the e2e test. It works. It's not entirely... understandable.

I could attempt to refactor the unix dgram stuff into a helper function

Also, obviously it wouldn't be merged focused.

I'd squash these 3 commits together.

goochjj:
https://github.com/containers/crun/issues/414

mheon:
Changes LGTM, but I'd like someone else with more experience with notify to review the tests for accuracy

mheon:
@giuseppe @edsantiago PTAL

test/e2e/systemd_notify_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+var _ = Describe("Podman systemd-notify", func() {
+	var (
+		tempdir    string
+		err        error
+		podmanTest *PodmanTestIntegration
+	)
+
+	BeforeEach(func() {
+		SkipIfRootless()
```
edsantiago, line 27: This really bothers me, and I'm afraid I spent far too much time today trying to track down the AVC, never getting too far due to the distraction game.

Anyway, I'd love to see this limitation removed but I'm 99% convinced that the problem is outside the scope of your PR. This is really nice work, thank you.

goochjj: If it makes you feel any better - the binary appears to work just fine rootless in all 3 modes. I tried the test rootless but something in my environment is messed up and complaining about a lock... I had just copied that from the systemd_test.

goochjj:
Squashing commits and removing FIt

goochjj:
Yeah I'm not sure about the test failures.  From what I can tell rootless should work just fine.

I don't know how to solve this 
```
"unlinkat /tmp/podman/imagecachedir/vfs/dir/083c11ecaa71f30cd96462da54e18aa870d7cddfa199f9ff31ed8d9913d56b40/run: permission denied"
```
Or this on fedora 32
```
Running: podman [options] run fedora sh -c ls -ld $NOTIFY_SOCKET; systemd-notify --ready; printenv NOTIFY_SOCKET
         MAINPID=241378
         MAINPID=241378
         [0;1;31mFailed to notify init system: Permission denied[0m
         srwxrwxrwx. 1 root root 0 Jun 24 14:20 /tmp/podman_test883637516/notify
         /tmp/podman_test883637516/notify
         output: srwxrwxrwx. 1 root root 0 Jun 24 14:20 /tmp/podman_test883637516/notify /tmp/podman_test883637516/notify
         [AfterEach] Podman systemd-notify
```

Or the networking test that's failing on Fedora32 - which I don't think I caused.

Ubuntu tests seem to be apparmor related.

If anyone has any ideas, let me know.  Otherwise, let me know how you'd like me to proceed - if we should merge as is or if you want me to do something else.

edsantiago:
AppArmor - known problem, rebase to fix. But I don't think there's any point right now, because I don't think these tests are going to pass any time soon without @giuseppe's help.

@giuseppe, could you take a look at what crun is doing here, and how it's mounting the socket? In particular:
```console
$ NOTIFY_SOCKET=/tmp/podman-sdnotify/notify ./bin/podman run --rm --sdnotify=container fedora sh -c 'ls -lZ /tmp'
total 4
-rwx------. 1 root root system_u:object_r:fusefs_t:s0       1379 May 15 06:48 ks-script-xz5dtd94
drwx------. 2 root root unconfined_u:object_r:user_tmp_t:s0   60 Jun 25 12:17 podman-sdnotify

$ NOTIFY_SOCKET=/tmp/podman-sdnotify/notify ./bin/podman run --rm --sdnotify=container fedora sh -c 'ls -lZ /tmp/podman-sdnotify'
ls: cannot open directory '/tmp/podman-sdnotify': Permission denied

$ sudo setenforce 0
$ NOTIFY_SOCKET=/tmp/podman-sdnotify/notify ./bin/podman run --rm --sdnotify=container fedora sh -c 'ls -lZ /tmp/podman-sdnotify'
total 0
srwxrwxrwx. 1 root root unconfined_u:object_r:user_tmp_t:s0 0 Jun 25 12:17 notify
```
This happens if I try `/run/user/MYID/systemd/notify` also: the parent directory and socket are created with `user_tmp_t`, and I get an AVC when the container tries to access the socket.

giuseppe:
> @giuseppe, could you take a look at what crun is doing here, and how it's mounting the socket? In particular:

crun is mounting the socket from the `/run/user/$UID` directory.  That is a socket created by crun itself.  crun will read from this socket and propagate the notification to systemd.

That is why we see the `unconfined_u:object_r:container_var_run_t:s0` label.  Should we relax the selinux policy?  I'll try to change the label from crun, I'll let you know in a bit

giuseppe:
I've opened a PR here: https://github.com/containers/crun/pull/419

What I see now:

```
$ NOTIFY_SOCKET=/run/user/1000/notify/notify  ./bin/podman run --rm --sdnotify=container fedora ls -lZ /run/user/1000/notify
ERRO[0000] Error notifying systemd of Conmon PID: dial unixgram /run/user/1000/notify/notify: connect: no such file or directory 
ERRO[0000] Error notifying systemd of Conmon PID: dial unixgram /run/user/1000/notify/notify: connect: no such file or directory 
total 0
srwxrwxrwx. 1 root root unconfined_u:object_r:container_file_t:s0:c421,c594 0 Jun 25 13:27 notify
```

goochjj:
> AppArmor - known problem, rebase to fix. But I don't think there's any point right now, because I don't think these tests are going to pass any time soon without @giuseppe's help.

I'm rebased on origin/master... so...

goochjj:
> This happens if I try `/run/user/MYID/systemd/notify` also: the parent directory and socket are created with `user_tmp_t`, and I get an AVC when the container tries to access the socket.

It's my intention to review the mount labels in the "Make libpod/conmon handle the notify socket" branch...

goochjj:
Can someone please advise me on how I can run the integration tests locally? I can't get them to work on F32 or Ubuntu 20.04, for different reasons.

I want to fix the checks locally as pushing and having Cirrus tell me they're wrong is going to make this a lot longer to resolve. (And ginkgo focusing local works, but that's using my VM's state, not whatever state Cirrus uses)

At this point my assumptions are that container-selinux 2.137 is included because the libpodimage appears to include updates-testing, but maybe it's failing because crun 0.14 isn't packaged yet.  I want to confirm my assumptions but I can't even get that far locally.

I'm trying `make integration`, but the "ADD / $GOSRC" looks questionable to me, so I changed it to "ADD . $GOSRC" as it looks like it should add the libpod source directory into the container.

F32 dies with:
```
[2020-07-01T13:36:10+00:00] Removing:
[2020-07-01T13:36:10+00:00]  runc      x86_64      2:1.0.0-144.dev.gite6555cc.fc32      @@System      9.0 M
[2020-07-01T13:36:10+00:00]
[2020-07-01T13:36:10+00:00] Transaction Summary
[2020-07-01T13:36:10+00:00] ================================================================================
[2020-07-01T13:36:10+00:00] Remove  1 Package
[2020-07-01T13:36:10+00:00]
[2020-07-01T13:36:10+00:00] Freed space: 9.0 M
[2020-07-01T13:36:10+00:00] Running transaction check
[2020-07-01T13:36:10+00:00] Transaction check succeeded.
[2020-07-01T13:36:10+00:00] Running transaction test
[2020-07-01T13:36:10+00:00] Transaction test succeeded.
[2020-07-01T13:36:10+00:00] Running transaction
[2020-07-01T13:36:11+00:00] RPM: error: Plugin selinux: hook tsm_pre failed
[2020-07-01T13:36:11+00:00] Error: Could not run transaction.
[2020-07-01T13:36:11+00:00] <exit 1> dnf erase -y runc
##### (failed with exit: 1)
##### (exceeded 5 attempts)
Error: error building at STEP "RUN bash $PACKER_BASE/fedora_packaging.sh &&     dnf clean all &&     rm -rf /var/cache/dnf": error while running runtime: exit status 125
make: *** [Makefile:280: libpodimage] Error 125
```

Focal dies with:
```
STEP 8: RUN bash $PACKER_BASE/fedora_packaging.sh &&     dnf clean all &&     rm -rf /var/cache/dnf
ERRO[0000] systemd cgroup flag passed, but systemd support for managing cgroups is not available
systemd cgroup flag passed, but systemd support for managing cgroups is not available
error running container: error creating container for [/bin/sh -c bash $PACKER_BASE/fedora_packaging.sh &&     dnf clean all &&     rm -rf /var/cache/dnf]: : exit status 1
Error: error building at STEP "RUN bash $PACKER_BASE/fedora_packaging.sh &&     dnf clean all &&     rm -rf /var/cache/dnf": error while running runtime: exit status 1
make: *** [Makefile:280: libpodimage] Error 125
```

For that matter it seems podman doesn't detect/use systemd cgroup manager on my Focal testbed... Should these tests be run as root?

rhatdan:
@edsantiago PTAL

edsantiago:
```
systemd support for managing cgroups is not available
```
Can you try:  `setsebool -P container_manage_cgroup=1`

As for running ginkgo tests locally, sorry. I've spent a ridiculous amount of time over the last few years trying to do that and never even gotten close. It must be possible, I suppose, but for me it's actually just faster and less frustrating to let CI do it.

edsantiago:
One suggestion: since this is something I'm going to put in system tests anyway, why not just skip ginkgo and go directly to bats? I can help you write a `test/system/xxx-sdnotify.bats` test if you'd like. *THOSE* are completely trivial to run on a development system.

goochjj:
Already on container_manage_cgroup --> on

When I do sudo make integration I get
```
./hack/install_catatonit.sh
skipping ... catatonit is already installed
fatal: not a git repository (or any of the parent directories): .git
go build -ldflags '-X github.com/containers/libpod/libpod/define.gitCommit= -X github.com/containers/libpod/libpod/define.buildInfo=1593612156 -X github.com/containers/libpod/libpod/config._installPrefix=/usr/local -X github.com/containers/libpod/libpod/config._etcDir=/etc ' -tags "   selinux systemd exclude_graphdriver_devicemapper seccomp" -o test/checkseccomp/checkseccomp github.com/containers/libpod/test/checkseccomp
fatal: not a git repository (or any of the parent directories): .git
go build -ldflags '-X github.com/containers/libpod/libpod/define.gitCommit= -X github.com/containers/libpod/libpod/define.buildInfo=1593612157 -X github.com/containers/libpod/libpod/config._installPrefix=/usr/local -X github.com/containers/libpod/libpod/config._etcDir=/etc ' -o test/goecho/goecho github.com/containers/libpod/test/goecho
ginkgo -v  -tags "   selinux systemd exclude_graphdriver_devicemapper seccomp" -timeout=90m -cover -flakeAttempts 3 -progress -trace -noColor -nodes 3 -debug test/e2e/. hack/.
/bin/sh: ginkgo: command not found
make: *** [Makefile:331: ginkgo] Error 127
make: *** [Makefile:300: integration] Error 2
```


I'm game to test however.  I don't have experience with any of these tools, so I'm just guessing here. :-D

I'm on Kubernetes slack if that's easier.

goochjj:
@edsantiago This is how I'm currently testing

[notifytesting.zip](https://github.com/containers/libpod/files/4858525/notifytesting.zip)

For container, I'll test rootless with
bash notifytest-container.sh $HOME/tmp/notify.sock
bash notifytest-container.sh $HOME/tmp/a/long/path/notify.sock

Root
bash notifytest-container.sh /run/notify.sock
bash notifytest-container.sh /run/notify/notify.sock
bash notifytest-container.sh /tmp/no/particular/path/and/name

The others can run root or rootless since no bind mount takes place.

The --pids-limit 0 is an artifact of #6834 

Any help you can provide is appreciated!

edsantiago:
@goochjj what is your test environment where you can get this working? I'm just trying to get a manual confirmation that it works, and continue to get AVCs. I am using `crun` with the latest pr 419 (c04fd1508ab51d8e62bb29d40cddaf23a87e6a44) and the d46fd8ef266e84c70968a199991561c514618d6a commit of your PR. I have container-selinux-2.137.0-3.fc32. I still can't access the socket in the container:
```
$ NOTIFY_SOCKET=/tmp/podman-sdnotify/notify ./bin/podman run --rm --sdnotify=container fedora sh -c 'systemd-notify --ready'
Failed to notify init system: Permission denied

type=AVC msg=audit(1593618438.828:158200): avc:  denied  { sendto } for  pid=935032 comm="systemd-notify" path="/run/user/14904/crun/665079e2855b0ad745bec386e90b95825fbba4792ecdc710ac9a124f7ba7e38b/notify/notify" scontext=system_u:system_r:container_t:s0:c105,c402 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=unix_dgram_socket permissive=0
```

goochjj:
@edsantiago place your podman in /usr/local/bin, chcon it to match /usr/bin/
Actually, `sudo make install.bin-nobuild`
And then use /usr/local/bin.

The SELinux policy tags the podman process based on its source location

edsantiago:
Argh - missing `chcon`, that was it. Thank you!

PS location doesn't matter, all you need to do is `chcon -t container_runtime_exec_t bin/podman`

goochjj:
Interesting, when I didn't run it from /usr/local I don't think I got the process context, just the file context.  <shrug> I'm only pretending to know what I'm doing.

edsantiago:
@goochjj I submitted https://github.com/goochjj/libpod/pull/2 against your PR, it's a simple .bats file containing tests for each of your additions. Feel free to include it as part of your PR if you like.

Note that the tests are actually disabled, because they require a new conmon and container-selinux that is not available on the CI machines and probably won't be for a while. Since my tests pass, I'm pretty happy to just approve this without tests in CI. It will take time to get the CI systems updated.

goochjj:
@edsantiago  THANK YOU
So I should remove the test/e2e/ file then, ginkgo doesn't have to ALSO test it.

Should we maybe check crun and container-selinux directly? i.e
. /etc/os-release
if [ "$NAME" = "Fedora" -a "$VERSION_ID" = "32" ]; then
  VER=$(rpm -q container-selinux --queryformat "%{VERSION}")
  if [ -z "$VER" ]; then echo "Cannot find container-selinux"; exit 1; fi
  rpmdev-vercmp 2.137 "$VER"
  RET=$?
  if [ "$RET" = "12" ]; then
    #Package is 2.137.0 or newer
    true
  else
    echo "Container-selinux is older than 2.137, skipping test"
  fi
fi

Similarly look at crun --version....

edsantiago:
> So I should remove the test/e2e/ file then, ginkgo doesn't have to ALSO test it.

I'd be happy with that.

>Should we maybe check crun and container-selinux directly? 

...and add an equivalent check for Ubuntu? That's a scary amount of temporary code. Probably the cleanest option is to file an issue or maybe two: one to upgrade all CI machines, one to update the Fedora specfile to add a `Requires:` on the specific crun and container-selinux.

goochjj:
> > So I should remove the test/e2e/ file then, ginkgo doesn't have to ALSO test it.
> 
> I'd be happy with that.

Done

> > Should we maybe check crun and container-selinux directly?
> 
> ...and add an equivalent check for Ubuntu? That's a scary amount of temporary code. Probably the cleanest option is to file an issue or maybe two: one to upgrade all CI machines, one to update the Fedora specfile to add a `Requires:` on the specific crun and container-selinux.

Yeah ok you're right.

It looks like you're already conspiring for a crun build, and I know rhatdan is working through the updates-testing branch.  Is there something I need to do? What would those issues look like?

edsantiago:
I can file those once `crun` is packaged. One is an issue here, the other in the fedpkg repo for podman. I'm OOTO the rest of the week but have filed a reminder note for myself. Thanks for all your work on this!

goochjj:
Thank you!

libpod/options.go
```diff
@@ -550,6 +555,23 @@ func WithSystemd() CtrCreateOption {
 	}
 }
 
+// WithSystemd turns on systemd mode in the container
```
giuseppe, line 558: WithSdNotifyMode

libpod/oci_conmon_linux.go
```diff
@@ -1149,8 +1165,10 @@ func (r *ConmonOCIRuntime) configureConmonEnv(runtimeDir string) ([]string, []*o
 	env = append(env, fmt.Sprintf("HOME=%s", home))
 
 	extraFiles := make([]*os.File, 0)
-	if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
-		env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
+	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
+		if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
+			env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
```
giuseppe, line 1170: who will ensure the `notify` socket is accessible inside of the container?

goochjj: SdNotifyModeContainer - OCI runtime does
SdNotifyModeConmon/ignore - The notify socket ISN'T accessible in the container.

With this PR, all we do is selectively pass the NOTIFY_SOCKET to the OCI runtime.

goochjj:
Do I need to do anything else for this to be merged?

mheon:
Let me re-kick the tests, I think they just flaked.

@rhatdan @vrothberg @giuseppe @baude PTAL

vrothberg:
Restarted the last flake

rhatdan:
@giuseppe PTAL

mheon:
Two flakes, both on the Fedora repos. I'm just going to merge once this goes green.

/lgtm
/hold

rhatdan:
Thanks @goochjj Nice contribution.

PR #6591, opened 2020-06-12
Merged request to fix -f to stop following logs

jgallucci32:
This is a MERGED request from @QiWang19 and myself. It contains the test cases developed by her in addition to the modifications made to handle multiple cases. Close #6531

Fixes an issue with the first PR where a container would exit while following logs and the log tail continued to follow. This creates a subroutine which checks the state of the container and instructs the tailLog to stop when it reaches EOF.

Tested the following conditions:
* Tail and follow logs of running container
* Tail and follow logs of stopped container
* Tail and follow logs of running container which exits after some time

PR #6702, opened 2020-06-20
Stop following logs using timers

### LINKED ISSUE #6531: podman logs -f continues indefinitely on a stopped container
/kind bug

**Description**
Running `podman logs -f` on a stopped container run indefinitely and does not exit
This is a regression of Issue https://github.com/containers/libpod/issues/435 and PR https://github.com/containers/libpod/pull/437

1. Run a container which exits immediately
```
podman run -d --name hello-world hello-world
```

2. Follow logs of stopped container
```
podman logs -f hello-world
```

Command does not exit; hangs indefinitely

Command should exit.

The expected behavior of Docker is to exit when following a log of a stopped container

```
Version:      2.0.0-dev
API Version:  1
Go Version:   go1.14.3
Git Commit:   d857275901e8c1ea7515360631e5894018e17f30
Built:        Fri Jun  5 17:00:00 2020
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.15.0-dev
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.18-0.6.dev.git50aeae4.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.18-dev, commit: 51e91bbc42aaf0676bb4023fb86f00460bf7a0a2'
  cpus: 4
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: file
  hostname: gallucci-fedora.devlnk.net
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.7.0-1.fc33.x86_64
  linkmode: dynamic
  memFree: 15239475200
  memTotal: 16783962112
  ociRuntime:
    name: crun
    package: crun-0.13-2.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.13
      commit: e79e4de4ac16da0ce48777afb72c6241de870525
      spec: 1.0.0
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 4294963200
  swapTotal: 4294963200
  uptime: 22h 20m 14.55s (Approximately 0.92 days)
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - docker.io
store:
  containerStore:
    number: 6
    paused: 0
    running: 4
    stopped: 2
  graphDriverName: overlay
  graphOptions:
  graphRoot: /var/lib/containers/storage
  graphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "true"
  imageStore:
    number: 3
  runRoot: /var/run/containers/storage
  volumePath: /var/lib/containers/storage/volumes
version:
  APIVersion: 1
  Built: 1591401600
  BuiltTime: Fri Jun  5 17:00:00 2020
  GitCommit: d857275901e8c1ea7515360631e5894018e17f30
  GoVersion: go1.14.3
  OsArch: linux/amd64
  Version: 2.0.0-dev
```

```
podman-2.0.0-0.121.dev.git1fcb678.fc33.x86_64
```

Fedora 33 (Rawhide Prerelease)

jgallucci32:
This incorporates code from PR #6591 and #6614 but does not use
event channels to detect container state and rather uses timers
with a defined wait duration before calling `t.StopAtEOF()` to
ensure the last log entry is output before a container exits.

The polling interval is set to 250 milliseconds based on polling
interval defined in hpcloud/tail here:
https://github.com/hpcloud/tail/blob/v1.0.0/watch/polling.go#L117

Close #6531

Co-authored-by: Qi Wang <qiwan@redhat.com>
Signed-off-by: jgallucci32 <john.gallucci.iv@gmail.com>

### DIFF
--- a/libpod/container_log.go
+++ b/libpod/container_log.go
@@ -1,7 +1,9 @@
 package libpod
 
 import (
+	"fmt"
 	"os"
+	"time"
 
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/libpod/logs"
@@ -81,5 +83,33 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 		}
 		options.WaitGroup.Done()
 	}()
+	// Check if container is still running or paused
+	if options.Follow {
+		go func() {
+			for {
+				state, err := c.State()
+				if err != nil {
+					time.Sleep(250 * time.Millisecond)
+					tailError := t.StopAtEOF()
+					if tailError != nil && fmt.Sprintf("%v", tailError) != "tail: stop at eof" {
+						logrus.Error(tailError)
+					}
+					if errors.Cause(err) != define.ErrNoSuchCtr {
+						logrus.Error(err)
+					}
+					break
+				}
+				if state != define.ContainerStateRunning && state != define.ContainerStatePaused {
+					time.Sleep(250 * time.Millisecond)
+					tailError := t.StopAtEOF()
+					if tailError != nil && fmt.Sprintf("%v", tailError) != "tail: stop at eof" {
+						logrus.Error(tailError)
+					}
+					break
+				}
+				time.Sleep(250 * time.Millisecond)
+			}
+		}()
+	}
 	return nil
 }
--- a/libpod/container_log.go
+++ b/libpod/container_log.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/libpod/logs"
+	"github.com/hpcloud/tail/watch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -88,8 +89,8 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 		go func() {
 			for {
 				state, err := c.State()
+				time.Sleep(watch.POLL_DURATION)
 				if err != nil {
-					time.Sleep(250 * time.Millisecond)
 					tailError := t.StopAtEOF()
 					if tailError != nil && fmt.Sprintf("%v", tailError) != "tail: stop at eof" {
 						logrus.Error(tailError)
@@ -100,14 +101,12 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 					break
 				}
 				if state != define.ContainerStateRunning && state != define.ContainerStatePaused {
-					time.Sleep(250 * time.Millisecond)
 					tailError := t.StopAtEOF()
 					if tailError != nil && fmt.Sprintf("%v", tailError) != "tail: stop at eof" {
 						logrus.Error(tailError)
 					}
 					break
 				}
-				time.Sleep(250 * time.Millisecond)
 			}
 		}()
 	}
--- a/test/e2e/logs_test.go
+++ b/test/e2e/logs_test.go
@@ -311,4 +311,16 @@ var _ = Describe("Podman logs", func() {
 		logs.WaitWithDefaultTimeout()
 		Expect(logs).To(Not(Exit(0)))
 	})
+
+	It("follow output stopped container", func() {
+		containerName := "logs-f"
+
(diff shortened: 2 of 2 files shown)


jgallucci32:
@mheon @TomSweeneyRedHat @QiWang19 @edsantiago PTAL

This passed all checks first go around. The key difference with this PR is I removed the dependency on the external event module and it now runs independently using timers. To get the timing precise it needed to match the polling interval defined in hpcloud/tail here:
https://github.com/hpcloud/tail/blob/v1.0.0/watch/polling.go#L117
``` golang
func init() {
	POLL_DURATION = 250 * time.Millisecond
}
```
Additionally it needed to ensure when the container changes state, the 250 milisecond delay is applied before calling `StopAtEOF()` to avoid the race condition where the tailing stops before the last entry is output to stdout.

libpod/container_log.go
```diff
@@ -81,5 +83,33 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 		}
 		options.WaitGroup.Done()
 	}()
+	// Check if container is still running or paused
+	if options.Follow {
+		go func() {
+			for {
+				state, err := c.State()
+				if err != nil {
+					time.Sleep(250 * time.Millisecond)
```
rhatdan, line 92: Rather then sprinkle magic numbers here this should be
time.Sleep(watch.POLL_DURATION)
Then future developers would have an idea where this came from.

jgallucci32: @rhatdan Thanks for the suggestion; definitely a better way to do it. I was able to also refactor it down to a single sleep statement with no change to the logic of the loop. This has passed all checks and should be good to go now.

rhatdan:
LGTM
@edsantiago @giuseppe @vrothberg @mheon @baude PTAL

libpod/container_log.go
```diff
@@ -81,5 +84,31 @@ func (c *Container) readFromLogFile(options *logs.LogOptions, logChannel chan *l
 		}
 		options.WaitGroup.Done()
 	}()
+	// Check if container is still running or paused
+	if options.Follow {
+		go func() {
+			for {
+				state, err := c.State()
+				time.Sleep(watch.POLL_DURATION)
```
giuseppe, line 92: is there any reason why we do a `Sleep` before checking for the error?  Should the `Sleep` happen later?

test/e2e/logs_test.go
```diff
@@ -311,4 +311,16 @@ var _ = Describe("Podman logs", func() {
 		logs.WaitWithDefaultTimeout()
 		Expect(logs).To(Not(Exit(0)))
 	})
+
+	It("follow output stopped container", func() {
+		containerName := "logs-f"
+
+		logc := podmanTest.Podman([]string{"run", "--name", containerName, "-d", ALPINE})
```
edsantiago, line 318: I'm not comfortable with this usage (relying on ALPINE's entrypoint to exit immediately). Could you add an explicit command such as `true`?

libpod/container_log.go
jgallucci32: Yes. Under all conditions you must call `stopAtEOF()` when exiting the subroutine in order to break the infinite wait state `t.Lines` in the other subroutine. You must also wait up to the POLL_DURATION before calling `stopAtEOF()` or else you may run into the condition where the container has exited or does not exist and `t.Lines` is still polling for changes to the log file.

If you check commit https://github.com/containers/libpod/pull/6702/commits/8926e1f03bc82d75a295a0784c2756ea37c9b031 you will see there originally was 3 sleep statements, 2 of them after the `if` condition. This was refactored as there is no logical different between calling sleep first at the beginning versus calling it after the conditional statements and at the end of the entire loop.

test/e2e/logs_test.go
jgallucci32: Change has been made and pushed.

PR #6703, opened 2020-06-21
rootless_linux: improve error message

dougsland:
Improve the error message for rootless mode.

Git-Url: https://github.com/containers/libpod/issues/6572
Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -166,7 +166,8 @@ func GetConfiguredMappings() ([]idtools.IDMap, []idtools.IDMap, error) {
 	}
 	mappings, err := idtools.NewIDMappings(username, username)
 	if err != nil {
-		logrus.Errorf("cannot find mappings for user %s: %v", username, err)
+		logrus.Errorf(
+			"cannot find UID/GID for user %s: %v - check rootless mode in man pages.", username, err)
 	} else {
 		uids = mappings.UIDs()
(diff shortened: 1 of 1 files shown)


pkg/rootless/rootless_linux.go
```diff
@@ -166,7 +166,8 @@ func GetConfiguredMappings() ([]idtools.IDMap, []idtools.IDMap, error) {
 	}
 	mappings, err := idtools.NewIDMappings(username, username)
 	if err != nil {
-		logrus.Errorf("cannot find mappings for user %s: %v", username, err)
+		logrus.Errorf(
+			"cannot find mappings for user %s: %v, examine rootless mode in man podman", username, err)
```
mheon, line 170: This is used by more than just Podman, so this should probably be reworded - maybe `cannot find UID/GID mappings for user %s: %v - check /etc/subuid and /etc/subgid`?

dougsland: @mheon how about:
"cannot find UID/GID for user %s: %v - check rootless mode in the official documentation"

rhatdan: I would be more specific.  `check rootless mode in man pages.`

rhatdan:
LGTM

vrothberg:
Thanks for contributing!

Issue #6413, opened 2020-05-28
[APIv2] list containers: parameter limit doesn't work for non-running containers


Edward5hen:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
[The DOC](http://docs.podman.io/en/latest/_static/api.html#operation/libpodListContainers) says query parameter `limit`
> Return this number of most recently created containers, including non-running ones.

but it does not include non-running ones in my testing.

**Steps to reproduce the issue:**

1.      podman system service --time 0 tcp:127.0.0.1:8081 &> ~/server.log &

2. 
```
[root@localhost ~]# podman ps -a
CONTAINER ID  IMAGE                                                                                    COMMAND          CREATED            STATUS                        PORTS   NAMES
177a5acdc613  docker.io/library/alpine:latest                                                          sleep 100        2 hours ago        Exited (0) 2 hours ago                testctn1
63265da07da7  docker.io/library/alpine:latest                                                          /bin/sh          About an hour ago  Exited (0) About an hour ago          jolly_austin
85c424218dd6  k8s.gcr.io/pause:3.2                                                                                      About an hour ago  Up About an hour ago                  0df430f23e2f-infra
a3d39d7e7066  localhost/osbs/rsyslog:extras-rhel-7.8-containers-candidate-63044-20200511205526-x86_64  /bin/rsyslog.sh  2 hours ago        Up 2 hours ago                        rsyslog
a9324a2b2c9f  docker.io/library/alpine:latest                                                          echo hello       2 hours ago        Exited (0) 2 hours ago                testctn2
f32cc00cd445  docker.io/library/alpine:latest                                                          echo hello       2 hours ago        Exited (0) 2 hours ago                testctn3
fc44f47c82fe  docker.io/library/alpine:latest                                                          sleep 1000       2 hours ago        Exited (0) 2 hours ago                testctn
```

3.     curl -v "http://localhost:8081/v1.40/libpod/containers/json?limit=5" | jq

**Describe the results you received:**
Only return the running ones.
```
[root@localhost ~]# curl -v "http://localhost:8081/v1.40/libpod/containers/json?limit=5" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying ::1:8081...
[...]

PR #6710, opened 2020-06-22
libpod/containers/json: alias last -> limit

vrothberg:
Support both `last` and `limit` for in the containers listing endpoint.
We intended to use `limit` which is also mentioned in the docs, but the
implementation ended up using `last` as the http parameter; likely being
caused by the CLI using `--last`.  To avoid any regression, we decided
for supporting both and aliasing `last`.

Fixes: #6413

### DIFF
--- a/pkg/api/handlers/libpod/containers.go
+++ b/pkg/api/handlers/libpod/containers.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/libpod/pkg/ps"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

mheon:
I think we need to bump API version as part of this.

mheon:
1.0.0 -> 1.0.1 seems appropriate.

We should also maintain an API changelog somewhere.

vrothberg:
merge me :)

PR #6713, opened 2020-06-22
Set syslog for exit commands on log-level=debug

mheon:
We have a flag, --syslog, for telling logrus to log to syslog as well as to the terminal. Previously, this flag also set the exit command for containers to use `--syslog` (otherwise all output from exit commands is lost). I attempted to replicate this with Podman v2.0, but quickly ran into circular import hell (the flag is defined in cmd/podman, I needed it in cmd/podman/containers, cmd/podman imports cmd/podman/containers already, etc). Instead, let's just set the syslog flag automatically on `--log-level=debug` so we log exit commands automatically when debug-level logs are requested. This is consistent with Conmon and seems to make sense.


QiWang19:
LGTM

jwhonce:
hold pending tests
/hold
/lgtm

PR #6715, opened 2020-06-22
Fix conflicts between privileged and other flags

mheon:
The `--privileged` flag does not conflict with `--group-add` (this one was breaking Toolbox) and does not conflict with most parts of `--security-opt` (this was breaking Openstack).


TomSweeneyRedHat:
LGTM
assuming happy tests

Issue #6724, opened 2020-06-23
[ubuntu 20.04] Error: AppArmor profile "container-default" specified but not loaded


shaicoleman:
/kind bug

**Description**

```
$ sudo podman run -it hello-world
Error: AppArmor profile "container-default" specified but not loaded
```

**Steps to reproduce the issue:**

1. Clean install of Ubuntu Desktop 20.04 on VirtualBox, minimal installation

2. Install podman 2.0 via the Kubic project packages

3. Run `sudo podman run -it hello-world`

**Output of `podman version`:**

```
Version:      2.0.0
API Version:  1
Go Version:   go1.13.8
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.18, commit: '
  cpus: 4
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: file
  hostname: shai-VirtualBox
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.4.0-37-generic
  linkmode: dynamic
  memFree: 1991516160
  memTotal: 4126953472
  ociRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: 'runc version spec: 1.0.1-dev'
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 968105984
  swapTotal: 968105984
  uptime: 9m 35.21s
registries:
  search:
  - docker.io
  - quay.io
store:
  configFile: /etc/containers/storage.conf
  containerStore:
    number: 1
    paused: 0
    running: 0
    stopped: 1
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /var/lib/containers/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 1
  runRoot: /var/run/containers/storage
  volumePath: /var/lib/containers/storage/volumes
version:
  APIVersion: 1
  Built: 0
[...]

PR #6725, opened 2020-06-23
Fix default AppArmor profile to contain version suffix

saschagrunert:
The default AppArmor profile is now `container-default-$VERSION`, which
was previously only `container-default`.

Requires https://github.com/containers/common/pull/185
Fixes https://github.com/containers/libpod/issues/6724

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,3 +67,5 @@ require (
 	k8s.io/apimachinery v0.18.4
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 )
+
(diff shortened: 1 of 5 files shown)


vrothberg:
I think we need to fix that in c/common as c/common/pkg/apparmor is broken. See https://github.com/containers/libpod/issues/6724#issuecomment-648105172.

saschagrunert:
> I think we need to fix that in c/common as c/common/pkg/apparmor is broken. See [#6724 (comment)](https://github.com/containers/libpod/issues/6724#issuecomment-648105172).

Yes, see https://github.com/containers/common/pull/185

saschagrunert:
Okay, closing in favor of an containers/common-only solution.

PR #6726, opened 2020-06-23
system tests: invoke with abs path to podman

edsantiago:
Reversion of one part of #6679: my handling of 'realpath'
would not work when $PODMAN is 'podman-remote --url etc'.
Trying to handle that case got unmaintainable; so instead
let's just force 'make {local,remote}system' to invoke
with a full PODMAN path. This breaks down if someone
runs the tests with a manual 'bats' invocation, but I
think I'm the only one who ever does that.

Since podman path will now be very long in the logs,
add code to logformatter to abbreviate it like we do
for the ginkgo logs.

And, one thing that has bugged me for a long time:
in the error logs, show a different prompt ('#' vs '$')
to distinguish root vs rootless. This should make it
much easier to see at-a-glance whether a log file
is root or not. Add tests for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,7 @@ remoteintegration: varlink_generate test-binaries ginkgo-remote
 localsystem:
 	# Wipe existing config, database, and cache: start with clean slate.
 	$(RM) -rf ${HOME}/.local/share/containers ${HOME}/.config/containers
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

Issue #5698, opened 2020-04-01
named volume created when running with userns=keep-id has wrong ownership


stintel:
/kind bug

**Description**

When running podman with userns=keep-id and a named volume, the volume ownership is wrong.

**Steps to reproduce the issue:**

1. podman volume rm test

2. podman run --rm -it --userns=keep-id --mount 'type=volume,src=test,dst=/home/user' --workdir /home/user ubuntu:18.04

3. touch test

**Describe the results you received:**
```
1000@1da098d4a317:~$ touch test
touch: cannot touch 'test': Permission denied
1000@1da098d4a317:~$
```

**Describe the results you expected:**
```
1000@1da098d4a317:~$ touch test
1000@1da098d4a317:~$
```

**Additional information you deem important (e.g. issue happens only occasionally):**

When running with --userns=keep-id, the volume should be created with the real UID/GID of the user calling podman, so that it is writeable inside the container.

```
$ podman volume inspect test
[
     {
          "Name": "test",
          "Driver": "local",
          "Mountpoint": "/home/stijn/.local/share/containers/storage/volumes/test/_data",
          "CreatedAt": "2020-04-01T21:57:54.956706206+03:00",
          "Labels": {

          },
          "Scope": "local",
          "Options": {

          },
          "UID": 1,
          "GID": 1
     }
]
$ sudo ls -la /home/stijn/.local/share/containers/storage/volumes/test/_data
total 0
drwxr-xr-x 1 1065536 1065536  0 apr  1 21:57 .
drwx------ 1 1065536 1065536 10 apr  1 21:57 ..
```

**Output of `podman version`:**

```
Version:            1.8.0
RemoteAPI Version:  1
Go Version:         go1.13.8
Built:              Tue Mar 10 15:59:20 2020
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.8
  podman version: 1.8.0
host:
  BuildahVersion: 1.13.1
  CgroupVersion: v1
  Conmon:
    package: Unknown
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.11, commit: ff9d97a08d7a4b58267ac03719786e4e7258cecf'
  Distribution:
    distribution: gentoo
    version: unknown
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 100
      size: 1
    - container_id: 1
      host_id: 1065536
      size: 65536
    uidmap:
[...]

PR #6730, opened 2020-06-23
set correct uid/gid on volumes

vrothberg:
When setting up the volumes for a container, make sure that new volumes
are created with the correct uid/gid which we're looking up in the
container config.

Fixes: #5698

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -1141,8 +1141,13 @@ func (c *Container) IDMappings() (storage.IDMappingOptions, error) {
 
 // RootUID returns the root user mapping from container
 func (c *Container) RootUID() int {
+	return c.findUID(0)
+}
+
+// findUID returns the UID mapping of the specified uid or 0 if it couldn't be found.
+func (c *Container) findUID(uid int) int {
 	for _, uidmap := range c.config.IDMappings.UIDMap {
-		if uidmap.ContainerID == 0 {
+		if uidmap.ContainerID == uid {
 			return uidmap.HostID
 		}
 	}
@@ -1151,8 +1156,13 @@ func (c *Container) RootUID() int {
 
 // RootGID returns the root user mapping from container
 func (c *Container) RootGID() int {
+	return c.findGID(0)
+}
+
+// findUID returns the GID mapping of the specified gid or 0 if it couldn't be found.
+func (c *Container) findGID(gid int) int {
 	for _, gidmap := range c.config.IDMappings.GIDMap {
-		if gidmap.ContainerID == 0 {
+		if gidmap.ContainerID == gid {
 			return gidmap.HostID
 		}
 	}
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/buildah/pkg/chrootuser"
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/libpod/libpod/define"
 	"github.com/containers/libpod/libpod/events"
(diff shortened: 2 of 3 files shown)


vrothberg:
@rhatdan PTAL

libpod/runtime_ctr.go
```diff
@@ -308,8 +309,15 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (_ *Contai
 
 		logrus.Debugf("Creating new volume %s for container", vol.Name)
 
+		// Get the uid/gid of the container user to create the volume
+		// with the correct uid/gid. See github.com/containers/libpod/issues/5698.
+		uid, gid, _, err := chrootuser.GetUser(ctr.state.Mountpoint, ctr.config.User)
```
mheon, line 314: This is not safe - `ctr.state.Mountpoint` is unset at this point, because the container has not been mounted.

mheon: If you pass a user by *name* instead of UID, this will break horribly and use the host's `/etc/passwd` instead of the container's.

mheon: When I went over this with Sujil when he was working on it, we agreed that mounting the container this early in creation is prohibitively difficult, so he was investigating chown'ing the volume as part of copy-up.

vrothberg: Eek. What shall we do?

mheon: The best of a series of bad options was to do it as part of this block: https://github.com/containers/libpod/blob/f4c3b718eb22a161a897a6ed55d10f3a07e31aa8/libpod/container_internal.go#L1457-L1487

This only happens once, on first creation of the volume

vrothberg: > If you pass a user by _name_ instead of UID, this will break horribly and use the host's `/etc/passwd` instead of the container's.

Can we support by _name_ here at all?

mheon: If that's OK, it makes things *much* simpler

libpod/container.go
```diff
@@ -1141,8 +1141,13 @@ func (c *Container) IDMappings() (storage.IDMappingOptions, error) {
...
 // RootUID returns the root user mapping from container
 func (c *Container) RootUID() int {
+	return c.findUID(0)
+}
+
+// findUID returns the UID mapping of the specified uid or 0 if it couldn't be found.
+func (c *Container) findUID(uid int) int {
 	for _, uidmap := range c.config.IDMappings.UIDMap {
-		if uidmap.ContainerID == 0 {
+		if uidmap.ContainerID == uid {
```
giuseppe, line 1150: is the function supposed to be used only with `findUID(0)` as it seems in the current version?  Otherwise it needs to check if the uid is in the `[idmap.ContainerID; idmap.ContainerID+idmap.Size)` range and the offset must be added to the return value.

libpod/container.go
```diff
@@ -1151,8 +1156,13 @@ func (c *Container) RootUID() int {
...
 // RootGID returns the root user mapping from container
 func (c *Container) RootGID() int {
+	return c.findGID(0)
+}
+
+// findUID returns the GID mapping of the specified gid or 0 if it couldn't be found.
+func (c *Container) findGID(gid int) int {
 	for _, gidmap := range c.config.IDMappings.GIDMap {
-		if gidmap.ContainerID == 0 {
+		if gidmap.ContainerID == gid {
```
giuseppe, line 1165: same comment as above

libpod/runtime_ctr.go
giuseppe: can we move the chown later (container_internal_linux.go generateSpec)?  We already parse once the `/etc/passwd` and `/etc/group` files

giuseppe: looking at it more in details, I think a possible solution is to have an operation like `ChownIfEmpty` on the volumes that we can use once we read `/etc/passwd` from the image

vrothberg:
Closing in favour of https://github.com/containers/libpod/pull/6747.

PR #6731, opened 2020-06-23
Add JSON output field for ps

baude:
the toolbox team needs a field in our ps json that represents a human readable time.

### DIFF
--- a/pkg/domain/entities/container_ps.go
+++ b/pkg/domain/entities/container_ps.go
@@ -15,6 +15,8 @@ type ListContainer struct {
 	Command []string
 	// Container creation time
 	Created int64
+	// Human readable container creation time.
+	CreatedAt string
 	// If container has exited/stopped
 	Exited bool
 	// Time container exited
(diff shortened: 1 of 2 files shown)


jwhonce:
LGTM

mheon:
LGTM

pkg/domain/entities/container_ps.go
```diff
@@ -15,6 +15,8 @@ type ListContainer struct {
 	Command []string
 	// Container creation time
 	Created int64
+	// Human readable container creation time.
+	CreatedAt string
```
debarshiray, line 19: I wanted to point out that the `Created` and `CreatedAt` fields are switched between `ps` and `images`.

In `images`, `Created` is the human-friendly string while `CreatedAt` is the machine-readable form. In `ps`, they are the other way round.

I don't really mind it. Just wanted to be sure that it's a deliberate decision (for Docker compatibility?).

rhatdan: Do  you know if they are switched in Docker as well?

HarryMichal: Docker does not offer a JSON-formatted output.

PR #6732, opened 2020-06-23
Set stop signal to 15 when not explicitly set

mheon:
When going through the output of `podman inspect` to try and identify another issue, I noticed that Podman 2.0 was setting
StopSignal to 0 on containers by default. After chasing it through the command line and SpecGen, I determined that we were
actually not setting a default in Libpod, which is strange because I swear we used to do that. I re-added the disappeared default and now all is well again.

Also, while I was looking for the bug in SpecGen, I found a bunch of TODOs that have already been done. Eliminate the comments for these.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -535,7 +535,6 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 
 	s.SeccompPolicy = c.SeccompPolicy
 
-	// TODO: should parse out options
 	s.VolumesFrom = c.VolumesFrom
 
 	// Only add read-only tmpfs mounts in case that we are read-only and the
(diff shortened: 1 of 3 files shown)


mheon:
Let me add a test...

mheon:
Done

rhatdan:
LGTM 
But tests are not happy.

mheon:
Oops. Fixed.

rhatdan:
@mheon still broken.

mheon:
```
[+0695s]   Expected
[+0695s]       <uint>: 15
[+0695s]   to equal
[+0695s]       <int>: 15
```

I really hate Go sometimes

TomSweeneyRedHat:
LGTM 
assuming happy tests, love the int/uint error @mheon, nice to know I'm not the only one to get bit by similar

Issue #6720, opened 2020-06-23
Any plans to implement docker compat endpoints for volumes in APIv2?


maybe-sybr:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Are there any current plans to implement docker compatible endpoints in APIv2 for managing volumes. I'm currently using the Python `docker` module for portability reasons and would prefer not to have to mix in usage of the alpha looking containers/podman-py package.

rhatdan:
Yes we should have these interfaces.  Interested in opening a PR?  :^)

maybe-sybr:
#6736 is reviewable now I think. I've thrown up some questions which probably any of the maintainers here could answer (mostly relating to where some code should live).

maybe-sybr:
With #6736 merged, this should be all done now. There are a few ongoing pieces of work (notably #6765 ) but those don't require this to remain open.

PR #6736, opened 2020-06-24
APIv2: Add docker compatible volume endpoints

maybe-sybr:
This is a WIP for a simple docker compatibility endpoint for volumes in APIv2. Happy to take any feedback on the approach or any admin I'd need to do (CLA?) since this is my first contrib to libpod.

### DIFF
--- a/pkg/api/handlers/compat/volumes.go
+++ b/pkg/api/handlers/compat/volumes.go
@@ -1,0 +1,238 @@
+package compat
+
+import (
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/libpod/define"
+	"github.com/containers/libpod/pkg/api/handlers/utils"
+	"github.com/containers/libpod/pkg/domain/filters"
+	"github.com/containers/libpod/pkg/domain/infra/abi/parse"
+	docker_api_types "github.com/docker/docker/api/types"
+	docker_api_types_volume "github.com/docker/docker/api/types/volume"
+	"github.com/gorilla/schema"
+	"github.com/pkg/errors"
+)
+
+func ListVolumes(w http.ResponseWriter, r *http.Request) {
+	var (
+		decoder = r.Context().Value("decoder").(*schema.Decoder)
+		runtime = r.Context().Value("runtime").(*libpod.Runtime)
+	)
+	query := struct {
+		Filters map[string][]string `schema:"filters"`
+	}{
+		// override any golang type defaults
+	}
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+
+	// Reject any libpod specific filters since `GenerateVolumeFilters()` will
+	// happily parse them for us.
+	for filter := range query.Filters {
+		if filter == "opts" {
+			utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+				errors.Errorf("unsupported libpod filters passed to docker endpoint"))
+			return
+		}
+	}
+	volumeFilters, err := filters.GenerateVolumeFilters(query.Filters)
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+
+	vols, err := runtime.Volumes(volumeFilters...)
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	volumeConfigs := make([]*docker_api_types.Volume, 0, len(vols))
+	for _, v := range vols {
+		config := docker_api_types.Volume{
+			Name:       v.Name(),
+			Driver:     v.Driver(),
+			Mountpoint: v.MountPoint(),
+			CreatedAt:  v.CreatedTime().Format(time.RFC3339),
+			Labels:     v.Labels(),
+			Scope:      v.Scope(),
+			Options:    v.Options(),
+		}
+		volumeConfigs = append(volumeConfigs, &config)
+	}
+	response := docker_api_types_volume.VolumeListOKBody{
+		Volumes:  volumeConfigs,
+		Warnings: []string{},
+	}
+	utils.WriteResponse(w, http.StatusOK, response)
+}
+
+func CreateVolume(w http.ResponseWriter, r *http.Request) {
+	var (
+		volumeOptions []libpod.VolumeCreateOption
+		runtime       = r.Context().Value("runtime").(*libpod.Runtime)
+		decoder       = r.Context().Value("decoder").(*schema.Decoder)
+	)
+	/* No query string data*/
+	query := struct{}{}
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	// decode params from body
+	input := docker_api_types_volume.VolumeCreateBody{}
+	if err := json.NewDecoder(r.Body).Decode(&input); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Decode()"))
+		return
+	}
+
+	if len(input.Name) > 0 {
+		volumeOptions = append(volumeOptions, libpod.WithVolumeName(input.Name))
+	}
+	if len(input.Driver) > 0 {
+		volumeOptions = append(volumeOptions, libpod.WithVolumeDriver(input.Driver))
+	}
+	if len(input.Labels) > 0 {
+		volumeOptions = append(volumeOptions, libpod.WithVolumeLabels(input.Labels))
+	}
+	if len(input.DriverOpts) > 0 {
+		parsedOptions, err := parse.VolumeOptions(input.DriverOpts)
+		if err != nil {
+			utils.InternalServerError(w, err)
+			return
+		}
+		volumeOptions = append(volumeOptions, parsedOptions...)
+	}
+	vol, err := runtime.NewVolume(r.Context(), volumeOptions...)
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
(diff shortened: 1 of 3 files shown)


maybe-sybr:
Related to my question about compat endpoints for volumes at #6720

pkg/api/server/register_volumes.go
```diff
@@ -102,5 +103,35 @@ func (s *APIServer) registerVolumeHandlers(r *mux.Router) error {
...
+	//        - name=<volume-name> Matches all of volume name.
+	//
+	//      Note:
+	//        The boolean `dangling` filter is not yet implemented for this endpoint.
+	// responses:
+	//   '200':
+	//     "$ref": "#/responses/DockerVolumeList"
+	//   '500':
+	//     "$ref": "#/responses/InternalError"
+	r.Handle(VersionedPath("/volumes"), s.APIHandler(compat.ListVolumes)).Methods(http.MethodGet)
```
maybe-sybr, line 134: From the network compat endpoint registration, it seems like I might also want unversioned path handlers here?

pkg/api/handlers/compat/volumes.go
```diff
@@ -0,0 +1,68 @@
...
+			Mountpoint: v.MountPoint(),
+			CreatedAt:  v.CreatedTime().Format(time.RFC3339),
+			Labels:     v.Labels(),
+			Scope:      v.Scope(),
+			Options:    v.Options(),
+		}
+		volumeConfigs = append(volumeConfigs, &config)
+	}
+	response := docker_api_types_volume.VolumeListOKBody{
+		Volumes: volumeConfigs,
```
maybe-sybr, line 64: It seems like that gating test didn't like this line and wants me to horizontally align this with the line below. Is that right?

mheon:
No CLA required, but all commits do need a Signoff.

pkg/api/handlers/compat/volumes.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	response := docker_api_types_volume.VolumeListOKBody{
+		Volumes:  volumeConfigs,
+		Warnings: []string{},
+	}
+	utils.WriteResponse(w, http.StatusOK, response)
+}
+
+// swagger:model DockerVolumeCreate
+type DockerVolumeCreate docker_api_types_volume.VolumeCreateBody
```
maybe-sybr, line 82: I put this inline here rather than in entities because it seemed like a bad idea to put docker junk in there.

pkg/api/handlers/compat/volumes.go
```diff
@@ -0,0 +1,263 @@
...
+
+// This response definition is used for both the create and inspect endpoints
+// swagger:response DockerVolumeInfoResponse
+type swagDockerVolumeInfoResponse struct {
```
maybe-sybr, line 86: The linter complains about these swagger response defs being dead code. It's right, and I don't really know if swagger can even discover them here (IDK how it works and haven't checked out how to run the build/doc steps locally). I put them inline here since I made the choice to put the `DockerVolumeCreate` typedef in here. These have a kind of logical home in `api/server/swagger.go`, but I figured for now I'd keep everything in one place and just point out that it needs some thought from someone with more insight than me.

maybe-sybr:
This changeset implements all of the v1.40 endpoints with a few support caveats noted in the commit message. The endpoints work fine when I hit them in a project I'm working on, but I'm not testing many of the edge cases. I'm leaving this as is for now, pending some input from y'all. I'll keep an eye on the PR for any feedback.

pkg/api/server/register_volumes.go
maybe-sybr: There are unversioned path handlers now

pkg/api/handlers/compat/volumes.go
maybe-sybr: I fixed this

pkg/api/handlers/compat/volumes.go
mheon: I think this needs to move into `pkg/entities/` with the other structs, and probably be public - those two things should resolve the issue

pkg/api/handlers/compat/volumes.go
mheon: It's fine, we do that elsewhere in entities where we need to guarantee compatability

mheon:
@maybe-sybr Let me get a quick handler for the dangling filter written, should be pretty easy

mheon:
#6756

maybe-sybr:
@mheon I have a new changeset which uses your dangling filter and also excludes the libpod specific filters. I'll push a HEAD which merges in the current HEAD from #6756 and has that fixup commit for my change. This will obviously need a rebase/autosquash once your merge goes in, or a rebase/autosquash/removal of the merge commit if we want to merge this first and wait for `dangling` support to land independently.

pkg/api/handlers/compat/volumes.go
maybe-sybr: These live in `pkg/entities` now, check the latest fixup commit

pkg/api/handlers/compat/volumes.go
maybe-sybr: As above, these live in `pkg/entities` now

maybe-sybr:
Rebased on master, squashed down and I fixed up those swagger model types so they're capital cased/public. I think this will be good to go now.

pkg/api/handlers/compat/volumes.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+
+	// Reject any libpod specific filters since `GenerateVolumeFilters()` will
+	// happily parse them for us.
+	for filter := range query.Filters {
+		switch filter {
```
maybe-sybr, line 40: gocritic is bullying me into making this an `if` since we're only excluding one filter. If we need to exclude more in future, we can just rework it back into something a bit more concise like this `switch`

rhatdan:
@baude @jwhonce PTAL

rhatdan:
Thanks @maybe-sybr

pkg/api/handlers/compat/volumes.go
```diff
@@ -0,0 +1,239 @@
...
+	"time"
+
+	"github.com/containers/libpod/libpod"
+	"github.com/containers/libpod/libpod/define"
+	"github.com/containers/libpod/pkg/api/handlers/utils"
+	"github.com/containers/libpod/pkg/domain/filters"
+	"github.com/containers/libpod/pkg/domain/infra/abi/parse"
+	"github.com/gorilla/schema"
+	"github.com/pkg/errors"
+
```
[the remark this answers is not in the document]
maybe-sybr: done, force pushed the fixedup commit

TomSweeneyRedHat:
@maybe-sybr you've a gofmt error that needs fixing:
```
gofmt -s -w pkg/api/handlers/compat/volumes.go
```
then add, commit, push.

pkg/api/handlers/compat/volumes.go
```diff
@@ -0,0 +1,238 @@
...
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	name := utils.GetName(r)
+	vol, err := runtime.LookupVolume(name)
+	if err != nil {
+		utils.VolumeNotFound(w, name, err)
+		return
+	}
+	if err := runtime.RemoveVolume(r.Context(), vol, query.Force); err != nil {
```
mheon, line 185: We realized recently that our Force and the Docker engine's Force are actually different things (Docker's behavior seems to be that Force means "do not error on missing volume"; ours means "remove any not-running containers using the volume at the same time")

mheon: As such, I recommend unconditionally setting Force to false for now.

maybe-sybr: Done, will push an extra commit in a moment. I've also added handling for `force` to alter a 404 to a 204 when the named volume can't be found. That means we don't need to caveat our support or explicitly reject the query parameter when we see it. Take a look at the new diff here where you have a chance, @mheon

maybe-sybr: Well, gocritic hates that code structure. I'll make some changes

mheon:
LGTM - lot of TODOs, but we can sort them in the future. Nice work!

@baude @rhatdan @vrothberg PTAL

maybe-sybr:
Missed the sign-off on that commit, just a push to add it and hopefully pass gating now

maybe-sybr:
The swagger check in gating is complaining at me for changing some of the endpoint lines to match a change it suggested for one of them in a previous run. https://github.com/containers/libpod/pull/6736/checks?check_run_id=824743139

I would have assumed all of these should be either `compat` (and end up being volumes (compat)` in the API docs) or just `volumes` (and some magic takes place?). Any ideas on how to make it happy and have the docs render sensibly?

vrothberg:
I reads like Swagger wants ` volumes` to be at the end of the lines it complains about

maybe-sybr:
@vrothberg , check out these two checks and the diff between them:
 * https://github.com/containers/libpod/pull/6736/checks?sha=10227dbd9619eba3e2fd3747af05ebbbe9a681d5
 * https://github.com/containers/libpod/pull/6736/checks?sha=257f99bc87b73b29e3522f33b3cded1a0f6fb867 (current)
 * https://github.com/maybe-sybr/libpod/compare/10227dbd9619eba3e2fd3747af05ebbbe9a681d5..257f99bc87b73b29e3522f33b3cded1a0f6fb867

The way I read that complaint was that initially I needed to change the word `volumes` to `compat` on L111, which I did as well as making the same change to lines following it. Now it's complaining about the other lines I changed and wants them to be `volumes` again.

I think I understand what's happening through - swagger likely wants that word to be the parent of the endpoint being described on that line. So for the top level `/volumes` endpoint, that's `/` which is probably assigned to `compat` somewhere. Then all of the other endpoints live under `/volumes` so that's their parent and swagger is right to complain when I lie about them being children of the `compat` "group".

I'm going to make that change now that I have a handle on why it's telling me to do things. Will push a squashed and rebased change momentarily.

maybe-sybr:
CI looks much happier on the unsquashed version of the diff I just re-pushed. LMK if any other changes need to be made prior to merge.

PR #6738, opened 2020-06-24
APIv2:fix: Remove `/json` from compat network EPs

maybe-sybr:
Just a trivial fix to make the network endpoints match the docker enginer API spec. Hopefully the current endpoint URLs aren't considered part of the API since they've already landed in podman 2.0, and we can fix them now. If we need to, we can just add the correct endpoints paths and leave the `/json` ones vestigial.

 AFAICT all of the other registered endpoints in the APIv2 compat layer which include a `/json` suffix are correct.


rhatdan:
If these are compatability endpoints and they are not compatable, then this is a bug.

PR #6747, opened 2020-06-24
container: move volume chown after spec generation

giuseppe:
move the chown for newly created volumes after the spec generation so the correct UID/GID are known.

Closes: https://github.com/containers/libpod/issues/5698

Alternative to: github.com/containers/libpod/pull/6730

### DIFF
--- a/libpod/volume.go
+++ b/libpod/volume.go
@@ -64,6 +64,14 @@ type VolumeState struct {
 	// create time, then cleared after the copy up is done and never set
 	// again.
 	NeedsCopyUp bool `json:"notYetMounted,omitempty"`
+	// NeedsChown indicates that the next time the volume is mounted into
+	// a container, the container will chown the volume to the container process
+	// UID/GID.
+	NeedsChown bool `json:"notYetChowned,omitempty"`
+	// UIDChowned is the UID the volume was chowned to.
+	UIDChowned int `json:"uidChowned,omitempty"`
+	// GIDChowned is the GID the volume was chowned to.
+	GIDChowned int `json:"gidChowned,omitempty"`
 }
 
 // Name retrieves the volume's name
@@ -113,13 +121,33 @@ func (v *Volume) Anonymous() bool {
 }
 
 // UID returns the UID the volume will be created as.
-func (v *Volume) UID() int {
-	return v.config.UID
+func (v *Volume) UID() (int, error) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	if !v.valid {
+		return -1, define.ErrVolumeRemoved
+	}
+
+	if v.state.UIDChowned > 0 {
+		return v.state.UIDChowned, nil
+	}
+	return v.config.UID, nil
 }
 
 // GID returns the GID the volume will be created as.
-func (v *Volume) GID() int {
-	return v.config.GID
+func (v *Volume) GID() (int, error) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	if !v.valid {
+		return -1, define.ErrVolumeRemoved
+	}
+
+	if v.state.GIDChowned > 0 {
+		return v.state.GIDChowned, nil
+	}
+	return v.config.GID, nil
 }
 
 // CreatedTime returns the time the volume was created at. It was not tracked
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1015,6 +1015,12 @@ func (c *Container) init(ctx context.Context, retainRetries bool) error {
 		return err
 	}
 
+	for _, v := range c.config.NamedVolumes {
+		if err := c.chownVolume(v.Name); err != nil {
+			return err
+		}
+	}
+
 	// With the spec complete, do an OCI create
 	if err := c.ociRuntime.CreateContainer(c, nil); err != nil {
 		// Fedora 31 is carrying a patch to display improved error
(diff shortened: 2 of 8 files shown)


rhatdan:
If a second container uses the same volume, does it get chowned?

TomSweeneyRedHat:
> If a second container uses the same volume, does it get chowned?

might be good to add a test for this.

vrothberg:
`make validate` and repushing should make gating happy

giuseppe:
> If a second container uses the same volume, does it get chowned?

no, it is chowned only the first time.  And it should be race free (a second container that accesses the volume between its creation and being chowned doesn't affect it)

mheon:
I'm still convinced this makes more sense as part of the volume mount operation - keeping the separate bool makes sense, but we should put it in the same place that copy-up currently happens.

giuseppe:
> I'm still convinced this makes more sense as part of the volume mount operation - keeping the separate bool makes sense, but we should put it in the same place that copy-up currently happens.

the mount currently happens before the spec is generated (the UID/GID are not known)

mheon:
That shouldn't be true - mount + copy up of volumes only happens as part of `prepare()` which is called as part of `Start()` (or `Init()`) once the container is created

giuseppe:
> That shouldn't be true - mount + copy up of volumes only happens as part of `prepare()` which is called as part of `Start()` (or `Init()`) once the container is created

yes volumes are mounted with `mountStorage` as part of `prepare()` while `generateSpec` is called later as part of `init()`

mheon:
Ah, I think I see now - you want to do the chown later, so we don't need a duplicate lookup up UID/GID. OK.

giuseppe:
> Ah, I think I see now - you want to do the chown later, so we don't need a duplicate lookup up UID/GID. OK.

does it look good for you?

libpod/volume.go
```diff
@@ -114,11 +122,17 @@ func (v *Volume) Anonymous() bool {
 
 // UID returns the UID the volume will be created as.
 func (v *Volume) UID() int {
+	if v.state.UIDChowned > 0 {
```
mheon, line 125: These need to do a v.lock() / v.update() first - v.state is not populated until the first v.update happens

libpod/volume.go
```diff
@@ -114,11 +122,17 @@ func (v *Volume) Anonymous() bool {
...
+		return v.state.UIDChowned
+	}
 	return v.config.UID
 }
 
 // GID returns the GID the volume will be created as.
 func (v *Volume) GID() int {
+	if v.state.GIDChowned > 0 {
```
mheon, line 133: Same as above - needs a lock/update

libpod/volume.go
mheon: We probably also need an unlocked version to use inside of volume_inspect below

mheon:
Sorry for the delay. Few issues with the volume functions, otherwise LGTM.

libpod/volume.go
giuseppe: thanks, fixed in the new version

disaster123:
Does this one also affects v1.9 ? I'm always getting "permission denied" messages for containers running in rootless mode using volumes.

mheon:
Yes, this affects every Podman version, I believe.

disaster123:
OK i'll try to backport your branch to v1.9 and check whether it fixes my problems.

disaster123:
I backported this to v1.9 (which was trivial just two missing imports). But it still fails:

```/home/podman/.local/share/containers/storage/volumes/maria_cs19-mariadb-data/_data``` is the source path for the volume which is owned by user podman shouldn't this be owned by the user running mysqld inside the container?

Output / log from Container:
```
2020-06-30 14:21:14 0 [Note] mysqld (mysqld 10.3.22-MariaDB-1:10.3.22+maria~bionic) starting as process 32 ...
2020-06-30 14:21:15 0 [ERROR] mysqld: Can't create/write to file '/var/lib/mysql/aria_log_control' (Errcode: 13 "Permission denied")
2020-06-30 14:21:15 0 [ERROR] mysqld: Got error 'Can't create file' when trying to use aria control file '/var/lib/mysql/aria_log_control'
```

The volume was created by ```podman-compose```

**Comment by giuseppe** [comment]
tests are green again

**Comment by disaster123** [comment]
The output of podman-compose looks like this:
```
podman pod create --name=maria --share net
a5ccf3e089fb2cac498d807982b5e21c4b54f153c9ff4270d318c2f187a65f98
0
podman volume inspect maria_cs19-mariadb-data || podman volume create maria_cs19-mariadb-data
Error: no volume with name "maria_cs19-mariadb-data" found: no such volume
podman volume inspect maria_cs19-mariadb-logs || podman volume create maria_cs19-mariadb-logs
Error: no volume with name "maria_cs19-mariadb-logs" found: no such volume
podman run --name=cs20-mariadb -d --pod=maria --label io.podman.compose.config-hash=123 --label io.podman.compose.project=maria --label io.podman.compose.version=0.0.1 --label com.docker.compose.container-number=1 --label com.docker.compose.service=mariadb -e zabbixServer=172.24.0.7 -e enableZabbixMonitoring=yes -e innodb_read_io_threads=4 -e innodb_write_io_threads=8 -e innodb_buffer_pool_size=8192M --mount type=bind,source=/home/podman/.local/share/containers/storage/volumes/maria_cs19-mariadb-data/_data,destination=/var/lib/mysql,bind-propagation=z --mount type=bind,source=/home/podman/.local/share/containers/storage/volumes/maria_cs19-mariadb-logs/_data,destination=/var/log/mysql,bind-propagation=z --add-host mariadb:127.0.0.1 --add-host cs20-mariadb:127.0.0.1 --hostname cs20-mariadb --tty docker.contentserv.com/cs-saas-prod/mariadbprod:stable
```

giuseppe:
I don't think the `--mount type=bind` is translated to a volume.  That looks like a plain bind mount

mheon:
It's a bind mount of a Libpod-managed named volume. That's honestly bizarre, and not a good idea.

Also, there's no `--user` directive, so I question whether this patch would help - it looks like mysql is started as root, then drops caps and becomes an unprivileged user

PR #6748, opened 2020-06-24
Add containers.conf default file for windows and MAC Installs

rhatdan:
We want to add this configuration file so that users can descover
how to configure the permanent connection to a remote podman instance.

### DIFF
--- a/contrib/remote/containers.conf
+++ b/contrib/remote/containers.conf
@@ -1,0 +1,11 @@
+# The containers configuration file specifies all of the available configuration
+# command-line options/flags for container engine tools like Podman
+# but in a TOML format that can be easily modified and versioned.
+
+[engine]
+
+# Default Remote URI to access the Podman service.
+# Examples:
+#  remote rootless ssh://engineering.lab.company.com/run/user/1000/podman/podman.sock
+#  remote rootfull ssh://root@10.10.1.136:22/run/podman/podman.sock
+# remote_uri= ""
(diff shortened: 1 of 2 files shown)


rhatdan:
@ashley-cui @baude @jwhonce PTAL

How should I install this in the Makefile?  IE We want this file packaged up with the Mac and Windows installs so that it ends up in the equivalent of 
$HOME/.config/containers/containers.conf

contrib/remote/containers.conf
```diff
@@ -0,0 +1,13 @@
+# The containers configuration file specifies all of the available configuration
+# command-line options/flags for container engine tools like Podman
+# but in a TOML format that can be easily modified and versioned.
+
+[engine]
+
+# Default Remote URI to access the Podman service.
+# Examples:
+#  rootless "unix://run/user/$UID/podman/podman.sock" (Default)
+#  rootfull "unix://run/podman/podman.sock.(Default)
```
TomSweeneyRedHat, line 10: do these hold for Windows?  IDK, just asking.

ashley-cui:
@rhatdan  Looks like it would make sense to put it in `make podman-remote-release-{platform}.zip`. I haven't looked too much at windows but as long as the file is in the zip for mac, I can change the homebrew formula to install the config wherever

contrib/remote/containers.conf
rhatdan: No the Unix examples should be removed.

TomSweeneyRedHat:
LGTM 
assuming happy tests

rhatdan:
@ashley-cui @baude PTAL

mheon:
@jwhonce Will this be included in the .msi packages for Windows, or do we need to do more for that to happen?

PR #6753, opened 2020-06-24
utils: drop default mapping when running uid!=0

giuseppe:
this is a leftover from the first implementation of rootless.  This
code is never hit by podman rootless anymore as podman automatically
creates a user namespace now.

Fixes an issue with podman remote when used with uid != 0.


rhatdan:
LGTM

PR #6754, opened 2020-06-24
Docs: consistency between man / --help

edsantiago:
New functionality in hack/man-page-checker: start cross-
referencing the man page 'Synopsis' line against the
output of 'podman foo --help'. This is part 1, flag/option
consistency. Part 2 (arg consistency) is too big and will
have to wait for later.

flag/option consistency means: if 'podman foo --help'
includes the string '[flags]' in the Usage message,
make sure the man page includes '[*options*]' in its
Synopsis line, and vice-versa. This found several
inconsistencies, which I've fixed.

While doing this I realized that Cobra automatically
includes a 'Flags:' subsection in its --help output
for all subcommands that have defined flags. This
is great - it lets us cross-check against the
usage synopsis, and make sure that '[flags]' is
present or absent as needed, without fear of
human screwups. If a flag-less subcommand ever
gets extended with flags, but the developer forgets
to add '[flags]' and remove DisableFlagsInUseLine,
we now have a test that will catch that. (This,
too, caught two instances which I fixed).

I don't actually know if the new man-page-checker
functionality will work in CI: I vaguely recall that
it might run before 'make podman' does; and also
vaguely recall that some steps were taken to remedy
that.

### DIFF
--- a/cmd/podman/healthcheck/run.go
+++ b/cmd/podman/healthcheck/run.go
@@ -12,12 +12,13 @@ import (
 var (
 	healthcheckRunDescription = "run the health check of a container"
 	healthcheckrunCommand     = &cobra.Command{
-		Use:     "run [flags] CONTAINER",
-		Short:   "run the health check of a container",
-		Long:    healthcheckRunDescription,
-		Example: `podman healthcheck run mywebapp`,
-		RunE:    run,
-		Args:    cobra.ExactArgs(1),
+		Use:                   "run CONTAINER",
+		Short:                 "run the health check of a container",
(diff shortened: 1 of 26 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
I just wish the GitHub review pane did a better job of showing what really changed.  Many were big in the pane, but were really just one liners.

edsantiago:
@TomSweeneyRedHat click the gear icon in the diff page, there's a setting there to ignore whitespace

PR #6759, opened 2020-06-24
Update dockerpy test readme

sujil02:
Corresponding to https://github.com/containers/libpod/pull/6603#issuecomment-644255858 comment, updating docker py readme.
@edsantiago PTAL

### DIFF
--- a/test/test_dockerpy/README.md
+++ b/test/test_dockerpy/README.md
@@ -2,29 +2,43 @@
 
 Python test suite to validate Podman endpoints using dockerpy library
 
+Validate :-
+
+Make sure podman binary exist under libpod/bin
+
+Validate Python 3.6+ installed using `python --version `
+
+Validate docker-py module exist by using `python list`. If missing install.
+
+
 Running tests
 =============
 To run the tests locally in your sandbox:
 
-#### Run the entire test
+cd into the libpod folder and run one of the following commands:-
+
+Note:- The test can be ran as root or rootless i.e sudo is optional.
+
+#### Run the entire test suite
 
 ```
-sudo PYTHONPATH=/usr/bin/python python -m dockerpy.images
+sudo PYTHONPATH=/usr/bin/python python -m unittest discover test/
 ```
 
 Passing the -v option to your test script will instruct unittest.main() to enable a higher level of verbosity, and produce detailed output:
 
 ```
-sudo PYTHONPATH=/usr/bin/python python -m unittest -v dockerpy.images
+sudo PYTHONPATH=/usr/bin/python python -m unittest discover -v test/
 ```
+
 #### Run a specific test class
 
 ```
-sudo PYTHONPATH=/usr/bin/python python -m unittest -v dockerpy.images.TestImages
+sudo PYTHONPATH=/usr/bin/python python -m unittest -v test.test_dockerpy.test_images
 ```
 
 #### Run a specific test within the test class
 
 ```
-sudo PYTHONPATH=/usr/bin/python python -m unittest -v dockerpy.images.TestImages.test_list_images
+sudo PYTHONPATH=/usr/bin/python python -m unittest -v test.test_dockerpy.test_images.TestImages.test_list_images
 ```


test/test_dockerpy/README.md
```diff
@@ -2,29 +2,43 @@
 
 Python test suite to validate Podman endpoints using dockerpy library
 
+Validate :-
+
+Make sure podman binary exist under libpod/bin
+
+Validate Python 3.6+ installed using `python --version `
```
TomSweeneyRedHat, line 9:
```suggestion
Validate Python 3.6+ installed using `python --version`
```

test/test_dockerpy/README.md
```diff
@@ -2,29 +2,43 @@
...
+
+Validate docker-py module exist by using `python list`. If missing install.
```
TomSweeneyRedHat, line 11:
```suggestion
Validate docker-py module exists by using `python list`. If missing install.
```

test/test_dockerpy/README.md
```diff
@@ -2,29 +2,43 @@
 
 Python test suite to validate Podman endpoints using dockerpy library
 
+Validate :-
+
+Make sure podman binary exist under libpod/bin
```
TomSweeneyRedHat, line 7:
```suggestion
Make sure the Podman binary exists under libpod/bin
```

TomSweeneyRedHat: although on further thought, don't uppercase Podman there, you're searching for the binary 'podman', not the binary 'Podman'.  Keep the and exists though

test/test_dockerpy/README.md
edsantiago: I've never heard of `python list`, and it doesn't work for me:
```console
$ python list
python: can't open file 'list': [Errno 2] No such file or directory
```
Can you give more info about what `docker-py` is? I found and installed `python3-docker-4.2.0-1.fc32` but even with that, I get the following when I try running tests:
```console
$ python -m unittest discover test
...long python crap...
ImportError: cannot import name 'Client' from 'docker'
```
I think you have an environment that may be somewhat customized, and you're not aware of it.

Suggestion: I realize this is a lot of work, but one thing I do to help me place myself in the shoes of my target audience is to try my instructions on a fresh clean virt. That helps me catch assumptions I've made, so I can document them for others.

test/test_dockerpy/README.md
```diff
@@ -2,29 +2,43 @@
...
+
+
 Running tests
 =============
 To run the tests locally in your sandbox:
 
-#### Run the entire test
+cd into the libpod folder and run one of the following commands:-
+
+Note:- The test can be ran as root or rootless i.e sudo is optional.
```
edsantiago, line 20: If sudo is optional, I recommend not including it in the sample commands below. Always start with the bare minimum privilege; assume the user can figure out sudo on their own if desired.

test/test_dockerpy/README.md
```diff
@@ -2,29 +2,43 @@
...
+
+#### Run the entire test suite
 
 ```
-sudo PYTHONPATH=/usr/bin/python python -m dockerpy.images
+sudo PYTHONPATH=/usr/bin/python python -m unittest discover test/
```
edsantiago, line 25: I don't think `PYTHONPATH` does what you think it does. It is a way to extend the default search path for modules; setting it to the executable is not going to do anything.

test/test_dockerpy/README.md
```diff
@@ -2,29 +2,43 @@
...
 ```
-sudo PYTHONPATH=/usr/bin/python python -m unittest -v dockerpy.images
+sudo PYTHONPATH=/usr/bin/python python -m unittest discover -v test/
 ```
+
 #### Run a specific test class
 
 ```
-sudo PYTHONPATH=/usr/bin/python python -m unittest -v dockerpy.images.TestImages
+sudo PYTHONPATH=/usr/bin/python python -m unittest -v test.test_dockerpy.test_images
```
edsantiago, line 37: Although I can't actually get it to work either way, I think it might be nicer to document that path as `test/test_dockerpy/test_images.py` -- that is, slashes instead of dots, and `.py` at the end. People use command completion, and it's much easier to type `test/test_do<TAB>/test_im<TAB>` than to spell out the entire thing by hand (and get it wrong, because tipyng is hard).

vrothberg:
AFAIKS the PR doesn't apply anymore. Paths have changed etc.

PR #6760, opened 2020-06-24
Fix podman python docker-py tests

jwhonce:
* Refactor packaging so unittest discovery works
* Refactor tests to use python3-docker.rpm that ships with Fedora32
* Flush image cache between tests suites
* Update documentation to reflect changes

Outstanding issue:
* client.get_image() does not fail if image does not exist


baude:
lgtm

jwhonce:
@sujil02 PTAL

sujil02:
@jwhonce, thank you for the cleanup. py test much-needed this pass.
Just, not a biggie. Having a test folder again under the main test folder might not be needed, discover would work even if the dir name would be prefixed `test`. Hence I named dir  test_docker_py. And Simply `PYTHONPATH=/usr/bin/python python -m unittest discover .` under libpod/test should work
Ignore the comment if dir is needed for the better maintenance

Also, I have a failed attempt at updating the readme here #6759. @edsantiago the readme here looks better and addresses most of your comments. We can have this in and I rebase and address any remaining comments.

test/python/dockerpy/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+```shell script
+# dnf install python3-docker
+```
+
+### Run the entire test suite
+
+```shell
+# cd test/python/dockerpy
+# PYTHONPATH=/usr/bin/python python -m unittest discover .
```
edsantiago, line 17: Is this (`PYTHONPATH`) a holdover from an earlier misunderstanding in docs? Unless I'm missing something profound, none of these `PYTHONPATHs` should be necessary, and they're really misleading.

If I am missing something, could you please document it?

Also, could I suggest using a `$` prompt instead of hash, to make it obvious to a user that this does not (and should not) need root?

PR #6761, opened 2020-06-24
Print port mappings in `ps` for ctrs sharing network

mheon:
In Podman v1.9, we printed port mappings for the container, even if it shared its network namespace (and thus ports) with another container. We regressed on this in Podman v2.0, which is fixed here.


rhatdan:
LGTM

edsantiago:
Confirmed, and I have a system test in the works.

/lgtm

PR #6763, opened 2020-06-24
Fix error handling problem in APIv2 network remove

maxm123:
Fix a bug with APIv2 compat network remove to log an ErrNetworkNotFound instead of nil
The endpoint uses network.Exists which returns 'false, nil' if the network doesn't exist.
The endpoint used to pass that 'nil' to utils.Error (api/handlers) which resulted in nil dereferencing.
It now calls utils.Error with the error object ErrNetworkNotFound.

Signed-off-by: Maximilian Müller <maxm123@techie.com>

### DIFF
--- a/pkg/api/handlers/compat/networks.go
+++ b/pkg/api/handlers/compat/networks.go
@@ -285,7 +285,7 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if !exists {
-		utils.Error(w, "network not found", http.StatusNotFound, err)
+		utils.Error(w, "network not found", http.StatusNotFound, network.ErrNetworkNotFound)
 		return
 	}
 	if err := network.RemoveNetwork(config, name); err != nil {


pkg/api/handlers/compat/networks.go
```diff
@@ -285,7 +285,7 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if !exists {
-		utils.Error(w, "network not found", http.StatusNotFound, err)
+		utils.Error(w, "network not found", http.StatusNotFound, ErrNetworkNotFound)
```
mheon, line 288: It looks like this needs to be `config.ErrNetworkNotFound` and you need to import `"github.com/containers/libpod/pkg/network/config"`

mheon:
@jwhonce PTAL at the second bit

pkg/api/handlers/compat/networks.go
maxm123: Yes, I'm sorry, I had forgotten the module prefix.
During installation, by the way, I had to cope with make install.tools (taken from CONTRIBUTING.md) which fails with: `install: cannot change permissions of ‘/usr/local/libexec/bats-core’: No such file or directory` since hack/install_bats.sh makes it install to /usr/local where my user account doesn't have write access to. But binaries also compile without that.

pkg/api/handlers/utils/errors.go
```diff
@@ -26,7 +26,7 @@ func Error(w http.ResponseWriter, apiMessage string, code int, err error) {
 	log.Infof("Request Failed(%s): %s", http.StatusText(code), err.Error())
 	em := entities.ErrorModel{
 		Because:      (errors.Cause(err)).Error(),
-		Message:      err.Error(),
+		Message:      apiMessage,
```
jwhonce, line 29: This change is going to break a lot of expectations of callers.

maxm123: If the current behavior is the way you like it, I could also just remove the apiMessage parameter and update the callers.

maxm123: I just took it out as it was just an code doc issue/unused parameter which made me wonder.
Now it is just about the nil dereferencing at the network endpoint.

mheon:
/approve
/ok-to-test

This LGTM. Do you want to remove the WIP for merge?

rhatdan:
Thanks @maxm123 
/lgtm

Issue #6766, opened 2020-06-25
`podman generate systemd` would create ill systemd unit files if using `=` in the --pod flag


floating-cat:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
The `podman generate systemd` would create ill systemd unit files if I use `=` in the --pod flag.

**Steps to reproduce the issue:**

1. Run below commands:

```
podman pod create --name systemd-pod
podman create --pod=systemd-pod alpine top
podman generate systemd --new --name systemd-pod
```

2. The `podman generate systemd` will generate a systems unit file with this content:

```
ExecStart=/usr/bin/podman run --conmon-pidfile %t/container-agitated_elgamal.pid --cidfile %t/container-agitated_elgamal.ctr-id --cgroups=no-conmon --pod-id-file %t/pod-systemd-pod.pod-id -d --pod=systemd-pod alpine top
```

It uses `--pod-id-file` and `--pod` at the same time when running the podman.

**Describe the results you received:**
When running these generated systems unit files, the podman will show `Cannot specify both --pod and --pod-id-file` error and exit because of this: https://github.com/containers/libpod/blob/2df3faab7b2801cc2e3c2291b618ae530223329d/cmd/podman/common/specgen.go#L259

**Describe the results you expected:**
We can run this generated systems unit files without any problem.

**Additional information you deem important (e.g. issue happens only occasionally):**

If I use the following commands, the `podman generate systemd` will generate the correct systemd unit files.

```
podman pod create --name systemd-pod
podman create --pod systemd-pod alpine top
podman generate systemd --new --name systemd-pod
```

**Output of `podman version`:**

```
Version:      2.0.0
API Version:  1
Go Version:   go1.14.4
Git Commit:   eec482cae3289ecaad45c602629657da7062ce9c
Built:        Sat Jun 20 20:30:13 2020
OS/Arch:      linux/amd64
```

vrothberg:
Opened https://github.com/containers/libpod/pull/6768 to fix it.

PR #6768, opened 2020-06-25
generate systemd: improve pod-flags filter

vrothberg:
When generating systemd units for pods, we need to remove certain
pod-related flags from the containers' create commands.  Make sure
to account for all the syntax including a single argument with key and
value being split by `=`.

Fixes: #6766

### DIFF
--- a/pkg/systemd/generate/common.go
+++ b/pkg/systemd/generate/common.go
@@ -1,6 +1,8 @@
 package generate
 
 import (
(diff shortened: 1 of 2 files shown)


vrothberg:
@giuseppe @rhatdan PTAL

TomSweeneyRedHat:
LGTM
and happy green test buttons

PR #6773, opened 2020-06-25
docs: recommend alternatives to podman inspect

markstos:
podman inspect is problematic because there can be naming clashes. Also,
it only inspects a couple of types of objects and the docs for it didn't
help discover that several more types could be inspected as well.

To address both concerns, we deprecate `podman inspect` and update the
docs to point to to the recommend alternatives.

Issue: #6756

### DIFF
--- a/docs/source/markdown/podman-inspect.1.md
+++ b/docs/source/markdown/podman-inspect.1.md
@@ -6,14 +6,20 @@ podman\-inspect - Display a container or image's configuration
 ## SYNOPSIS
 **podman inspect** [*options*] *name* [...]
 
-**podman image inspect** [*options*] *image*
-
-**podman container inspect** [*options*] *container*
-
 ## DESCRIPTION
+
 This displays the low-level information on containers and images identified by name or ID. By default, this will render
 all results in a JSON array. If the container and image have the same name, this will return container JSON for
 unspecified type. If a format is specified, the given template will be executed for each result.
+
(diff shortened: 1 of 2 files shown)


mheon:
I don't think we want to actively deprecate, but we should push folks in the direction of the more specific inspect commands

markstos:
@mheon I've revised to add references without deprecation notice.

rhatdan:
Looking at Docker, I guess we should attempt to match their functionality.

docs/source/markdown/podman-inspect.1.md
```diff
@@ -1,20 +1,26 @@
 % podman-inspect(1)
 
 ## NAME
-podman\-inspect - Display a container or image's configuration
+podman\-inspect - Display a container or image's configuration.
```
TomSweeneyRedHat, line 4: The added period is causing a gating test error as this line no longer matches the corresponding line in podman.1.md.

markstos: I pushed an amended commit which drops the period.

rhatdan:
Thanks @markstos 
LGTM

rhatdan:
/approve
@markstos Needs a rebase.

markstos:
@rhatdan Conflicts have been resolved.

rhatdan:
Sadly you are now failing gating tests.

You need to gofmt your code.

PR #6781, opened 2020-06-25
WIP: CI:DOCS: Add document for libpod release process

cevich:


cevich:
The purpose here isn't to to add red-tape, in fact you'll see there's a whole section there about release-automation.  This doc is more about "just in case" if Matt is unavailable for an emergency release and/or a step gets missed and debugging/recovery help is needed.

@mheon @rhatdan @baude @TomSweeneyRedHat PTAL and shoot me suggestions and ideas.  

Esp.  I'm not sure where to put this new doc :grin:

hack/RELEASE.md
```diff
@@ -0,0 +1,23 @@
+# Libpod/Podman/Podman Remote Release Checklist
+
+0. Make a new branch using the name format <1.2.3>[-<epic>]
+1. Open a PR with a commit having a message in the following format:
```
mheon, line 4: Release process is two PRs: a release notes PR, and a release PR.

Release notes updates RELEASE_NOTES.md and adds release notes, as one might expect. This must merge before the second PR is created.

We want two commits in the release PR, which must be rebased on top of the latest version of the branch (IE, include the release notes PR that merged before):

First one modifies version/version.go and contrib/spec/podman.spec.in and changes the version numbers there to the desired release, plus runs `make changelog` to generate changelog. Second one modifies version/version.go and the RPM spec again, bumping the micro version and appending `-dev` to version.go.

The first commit should be PGP signed, be named v2.0.1 or similar (the version number with lowercase v prepended), and have a message the same as the name of the tag. This tag is manually pushed once the second PR merges.

mheon: For example, the tag command I used to generate the v2.0.1 tag:

`git tag -s -m 'v2.0.1' v2.0.1 <FULL_COMMIT_SHA>`

hack/RELEASE.md
```diff
@@ -0,0 +1,23 @@
+# Libpod/Podman/Podman Remote Release Checklist
+
+0. Make a new branch using the name format <1.2.3>[-<epic>]
```
mheon, line 3: Usually we do not include micro - just Major.Minor - in branch names (branches will include multiple micro releases)

cevich: Sure, but automation is really naive and prefers to be told explicitly what to do :grin:

I see some branches named like '1.2.3-rc4'.  Managing those should also be included under the release-automation purview IMHO.  

So something like [this regex-format  sounds like it would be more accurate:
``^[A-z\-]+ Release [0-9]+\.[0-9]+(\.[0-9]+(-[\w\-\.]+)?)?``](https://regex101.com/r/Jr68bt/1)

?

hack/RELEASE.md
cevich: Can all the commits come in under a single PR?  From a process-triggering perspective, that's much more simple to detect.

Rebasing the code and creating tags/branches/github-releases are all fairly well-understood puzzle-pieces for me.  I also intend to include release-validation checks in the workflow that occurs in the PR.  So you'll have quick feedback before anything goes upstream/official.

I also want to wire-in notifications at the master and release-branch level, so we know right away if something breaks.  Github-actions provides a 'retry' button just like cirrus does.  But having confidence in the automation at the PR-stage is an important goal of mine.

hack/RELEASE.md
mheon: Are you sure those aren't tags?

Release branches (aside from one unfortunate case we will never, ever repeat) have always been `vX.Y`

hack/RELEASE.md
mheon: The release notes shouldn't require any action from release automation, but if we have a process doc, they should be mentioned.

All action should be on the actual release PR, which should always be a consistent format.

cevich: Excellent, glad to hear it. That also means release-commit / PR problems can be caught by the dummy-release run inside the PR.  Or the dummy run will be missing...which will also be noticed.  Okay, I believe I know the way forward here.

hack/RELEASE.md
cevich: That regex is just matching the commit comment, so automation knows to start either a dummy release (for a PR) or a real one (for a branch).  I'm happy to make the regex whatever you prefer (hence the playground link above).  So you'd [prefer `^[A-z\-]+ Release (v[0-9]+\.[0-9]+(\.[0-9]+(-[\w\-\.]+)?)?)`](https://regex101.com/r/Jr68bt/2)

cevich: > aside from one unfortunate case we will never, ever repeat

This is a good sign that automation is needed.  It prefers to ***always*** repeat it self :grin:

mheon: We'll need special handling for release candidates (`vA.B.C-rcD`) that explicitly marks them as pre-release in Github, so we may need two regexes?

cevich: That's fine, the `(-[\w\-\.]+)?)?)` at the end should allow for that.  Practically speaking, what is the difference with a pre-release.  For example, are there steps that should not be performed?

cevich:
@mheon when you have a mo.  PTAL https://github.com/containers/libpod/pull/6781/files#diff-3545a3a52685c6e1f95c1060b4febcb4R5-R34  I think you mentioned there should be a total of 3 commits in the release PR.  Which one am I missing and where in the sequence should it appear?

hack/RELEASE.md
mheon: The biggest thing is that pre-releases must mark the Github release as "pre-release", and do not need to build Windows/OS X/etc binaries (but it's not necessarily bad if it does build them).

hack/RELEASE.md
```diff
@@ -0,0 +1,77 @@
+(FIXME: Should this file live elsewhere?)
+
+# Libpod/Podman/Podman-Remote Automated Release Workflow
+
+1. Open a new PR.  The the `HEAD` commit *must* have a summary line matching the
+   following regex: `Release (v\d+\.\d+(\.\d+(-[\w\-\.]+)?)?)`.
+
+   * The second line of the commit text (after the summary and blank line), *must*
+     match the following regex: `^\(prior release: (.+)\)`.  Where the content of
```
mheon, line 9: Do we need this? I would think we can programmatically determine this, given we're strict semantic-versioning now.

hack/RELEASE.md
```diff
@@ -0,0 +1,77 @@
+(FIXME: Should this file live elsewhere?)
+
+# Libpod/Podman/Podman-Remote Automated Release Workflow
+
+1. Open a new PR.  The the `HEAD` commit *must* have a summary line matching the
```
mheon, line 5: It's probably not going to be HEAD but HEAD~1 - the format is Release commit, then -dev release commit (bumping to a fresh development version)

mheon:
It's three commits across two PRs.

Commit 1, PR 1: Update release notes with all changes of note

Wait until that merges, then rebase, making sure that commit is included.

Commit 1, PR 2: Bump to vA.B.C (the actual release, signed)
Commit 2, PR 2: Bump to vA.B.(C+1)-dev (Bump to a development version, so further builds do not show as `vA.B.C` in `podman info` - makes absolutely sure only the release commit is identified as a full release, not a prerelease.)

hack/RELEASE.md
cevich: My concern was over properly handling `X1.Y1.Z1-123-g0987654` upgrades to `X1.Y2`, but maybe you have a different understanding?  I only invented this (Prior release) method since it seemed a simple solution.

This is the failure-mode I was considering

1. Version `1.0.0` is tagged, branched, and released.
2. Somebody finds and fixes an (untested) `1.0` -> `2.0` upgrade breakage.
3. Version `1.0.1` is created on `1.0` branch (but not tagged or released)
4. An update is released for packages pointing at `1.0.1`
4. Other changes are made
5. Version `2.0.0` is submitted for automated release testing.
6. Automation (and humans) have established, successful upgrades MUST happen starting with `1.0.1` and not `1.0`.

In other words, how does a ***potential*** future release version know what is the ***minimum*** (lowest) acceptable prior version, for which upgrades (and downgrades?) are supported.

Since we cannot retroactively change any already released tag-objects, having this data available now will assist future automated release testing.  I thought that sticking something into the text is the simplest solution which is also, itself, easily verifiable.  If you know a better way, please tell (I'm ignorant in some things git and most things RCM :smile: ).

Or...maybe this is simply an unnecessary complication and I can just put it out of my mind.  I'm okay with that as well :grin: 

@edsantiago knows a thing or three about git...maybe he has an opinion (including: *"forget about upgrade-testing, it's too hard*" :smile: )

hack/RELEASE.md
cevich: Oh! Thanks.  Does the `-dev` release need to happen in the exact same PR or could it come after?  Does automation need to do anything in support of a `-dev` release?

So it goes (in order), and all in a single PR:

1. Release notes commit
2. Release commit
3. Dev Release commit

cevich: oh nvm, I see you answered me blow...

hack/RELEASE.md
mheon: Step 3 sounds dubious - we should always be tagging releases. Any version bump that is not -dev suffixed should make a tag.

mheon: On upgrades in general - we're very lenient on this (RHEL means we have to be - we went from 1.0, to 1.4.2, to 1.6.4, to 1.9.3 - lots of skipping there). Given the way our internals work, at present it's doubtful we will need to implement a requirement that only major version N-1 be upgraded to new major version N.

cevich:
> It's three commits across two PRs.

It adds a **lot** of complexity for automation running in one PR to (reliably) depend upon previous PR's commits. Also, since we have a merge-bot in play, it's possible a merge (even accidental) can come in out-of-order.  Can the release workflow be changed to use a single PR with:

1. [HEAD^2] Update release notes with all changes of note
2. [HEAD^1] Bump to vA.B.C (the actual release, signed)
3. [HEAD] Dev Release commit Bump to vA.B.(C+1)-dev

Also, it would be far safer (and less complicated), if all the release-important things can be validated in a single release PR (before the automated release).  This workflow would guarantee no other commits can merge between 1 and 3.  Thereby:

> makes absolutely sure only the release commit is identified as a full release, not a prerelease.

Another alternative that comes to mind is:  Have release automation file the `-dev` PR, which can then follow the normal process (except that a human doesn't have to open it).

hack/RELEASE.md
edsantiago: Duplicate 'the'

hack/RELEASE.md
```diff
@@ -0,0 +1,77 @@
...
+     (for example to correct an undetected bug), the process must begin again with the next
+     symentic version.  ***Note:*** In this special case, the `Prior Release` reference text
+     of all future commits *must not* reference the errant version.
+
+5. Assuming branch-testing passes automated checks.
+
+   * A new tag and branch *will* be created by automation, referencing the "Release vX.Y.Z"
+     commit.
+
+   * The tag *will* be named like `vX.Y.Z[-extra]' and the branch *will* be named `vX.Y`.
```
edsantiago, line 57: Wrong closing quote: should be backtick, not tick

hack/RELEASE.md
```diff
@@ -0,0 +1,77 @@
...
+     The tag annotation (message) *shall* be a string of the following format:
+     ``<prior release> -> <tag name> from <originating branch>``.
+
+   * Automation *may* create a Github-Release object, reflecting the name of the tag.  All
+     available build-time artifacts from branch-testing *may* be uploaded to the object.
+
+
+# Post-release Checklist
+
+0. Go to quat/libpod/gate "builds" page.  Verify a 'gate' image build is running or completed
```
edsantiago, line 69: This is probably a typo for quay. Why make the reader have to work? Why not include the full url? https://quay.io/repository/libpod/gate?tab=builds

hack/RELEASE.md
cevich: Thanks for the feedback/details, I think I'll just drop this aspect then.  However you brought up an interesting point: It sounds like there are at least two kinds of tags:  Those which pair to a release branch, and those that don't (e.g. -dev, -rc, etc.).

Idea: Maybe, for tag annotation then, we have automation add the branch-name of the release-branch.  That way other/different automation can easily tell the difference between release-tags and other-tags.

Or do you have another idea for the tag messages that might be helpful?

hack/RELEASE.md
cevich: good idea, thanks!

hack/RELEASE.md
cevich: Okay, so automation should make a github release, named using the form 'vX.Y' _UNLESS_ it's a pre-release as indicated by finding the string 'rc' inside the match-group `(-[\w\-\.]+)?)?)`.  When it's a pre-release, then the github-release MUST include the entire tag name (so 'vX.Y.Z-rcBlAH123WHatEvER`).  Correct?

cevich:
Force-pushed updates based on discussions / feedback.

vrothberg:
@cevich I guess a rebase will resolve the flakes

cevich:
Update: This PR is *on hold* behind #6784 (which will take some time)

TomSweeneyRedHat:
I'm late to the party @cevich, but @mheon had put together a google doc a while back.  I'm not sure it's still on point, but thought I'd link you to it in case it's helpful.  https://docs.google.com/document/d/1Yj8B1liKV1lVnqfAcpDByOcB7lYZ5VI4xTLC6Jc6FFc/edit

cevich:
@TomSweeneyRedHat your not late, I'm still working my down to this.  It's the last item to be run by the new CI setup I'm working on.  When I get closer, I'm planning on re-re-re-reviewing everything here, and ensuring the doc and my understanding is up to snuff.  So if you have time to make corrections/clarifications here, it will be eventually (much) appreciated :D

cevich:
Closing this PR for now, I will likely re-open it later on once my pile of other/unrelated CI changes stabilize.

PR #6752, opened 2020-06-24
Fix inspect to display multiple label: changes

rhatdan:
If the user runs a container like

podman run --security-opt seccomp=unconfined --security-opt label=type:spc_t --security-opt label=level:s0 ...

Podman inspect was only showing the second option

This change will show

            "SecurityOpt": [
                "label=type:spc_t,label=level:s0:c60",
                "seccomp=unconfined"
            ],

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #6783, opened 2020-06-25
system tests: add pod, inspect testing

edsantiago:
Followup to #6761: confirm that 'podman ps' shows the ports
on a running container in a pod created with -p (not to
be confused with the container itself running with -p,
tested in 500-networking.bats).

While we're at it, test that the port handling itself works,
by sending random text to the container and making sure the
container receives it.

Followup to #6752: 'podman inspect' should show multiple
security opts

### DIFF
--- a/test/system/200-pod.bats
+++ b/test/system/200-pod.bats
@@ -150,6 +150,18 @@ function random_ip() {
 
     pod_id_file=${PODMAN_TMPDIR}/pod-id-file
 
+    # Randomly-assigned ports in the 5xxx and 6xxx range
+    for port_in in $(shuf -i 5000-5999);do
+        if ! { exec 3<> /dev/tcp/127.0.0.1/$port_in; } &>/dev/null; then
+            break
+        fi
+    done
+    for port_out in $(shuf -i 6000-6999);do
+        if ! { exec 3<> /dev/tcp/127.0.0.1/$port_out; } &>/dev/null; then
+            break
+        fi
+    done
+
     # Create a pod with all the desired options
     # FIXME: --ip=$ip fails:
     #      Error adding network: failed to allocate all requested IPs
@@ -161,6 +173,7 @@ function random_ip() {
                --dns        "$dns_server"                \
                --dns-search "$dns_search"                \
                --dns-opt    "$dns_opt"                   \
+               --publish    "$port_out:$port_in"         \
                --label      "${labelname}=${labelvalue}"
     pod_id="$output"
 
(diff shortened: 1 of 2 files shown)


test/system/200-pod.bats
```diff
@@ -199,6 +212,34 @@ function random_ip() {
...
 
+    # Test local port forwarding, as well as 'ps' output showing ports
+    # Run 'nc' in a container, waiting for input on the published port.
+    c_name=$(random_string 15)
+    run_podman run -d --pod mypod --name $c_name $IMAGE nc -l -p $port_in
+    cid="$output"
+
+    # Try running another container also listening on the same port.
+    run_podman 1 run --pod mypod --name dsfsdfsdf $IMAGE nc -l -p $port_in
+    is "$output" "nc: bind: Address in use" \
```
edsantiago, line 223: This might be fragile since it relies on the output of `nc`, a tool we don't own

test/system/410-selinux.bats
```diff
@@ -63,4 +63,24 @@ function check_label() {
...
+
+    run_podman run -d --name myc \
+               --security-opt seccomp=unconfined \
+               --security-opt label=type:spc_t \
+               --security-opt label=level:s0 \
+               $IMAGE sh -c 'while test ! -e /stop; do sleep 0.1; done'
+    run_podman inspect --format='{{ .HostConfig.SecurityOpt }}' myc
+    is "$output" ".*label=type:spc_t.*"   "'podman inspect' includes type"
+    is "$output" ".*label=level:s0.*"     "'podman inspect' includes level"
+    is "$output" ".*seccomp=unconfined.*" "'podman inspect' includes seccomp"
```
edsantiago, line 80: Does anyone know if the output of `inspect` is guaranteed to be consistent? It'd be nice to do this in one single line.

mheon: In ordering?

edsantiago: Yes. In lots of iterations, both of `podman inspect` and recreating the container, the only value I've seen is:
```
[label=type:spc_t,label=level:s0 seccomp=unconfined]
```
If I change the order of `level` and `type`, the `inspect` order reflects command-line order. `seccomp` is always last regardless of command-line order.

...but that doesn't prove anything; it could just be that I've been lucky so far.

edsantiago:
Logging flakes for possible future analysis

#### in-podman fedora-31 fedora-31

  2020-06-25T16:11:59  [integration_test](https://api.cirrus-ci.com/v1/task/5092539076182016/logs/integration_test.log)
*   [Podman port [It] podman port multiple ports ](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--podman-port-multiple-ports--1)
  *   [[same]](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--podman-port-multiple-ports--2)
  *   [[same]](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--podman-port-multiple-ports--3)
*   [Podman network [It] podman network rm ](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--podman-network-rm--1)
*   [Podman run with --ip flag [It] Podman run two containers with the same IP ](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--Podman-run-two-containers-with-the-same-IP--1)
*   [Podman create with --ip flag [It] Podman create two containers with the same IP ](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--Podman-create-two-containers-with-the-same-IP--1)
*   [Podman run with --ip flag [It] Podman run two containers with the same IP ](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5092539076182016/html/integration_test.log.html#t--Podman-run-two-containers-with-the-same-IP--2)

test/system/410-selinux.bats
mheon: From the way the code is currently written, the order should never change, though that's not really an API guarantee

edsantiago: That's good enough for me - we don't need an API guarantee, since the tests are tightly bound: any change to the code will be caught early. I've pushed a change that tests in one go. Thank you for checking!

rhatdan:
LGTM

PR #6570, opened 2020-06-11
Change buildtag for remoteclient to remote for testing

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #6784, opened 2020-06-25
Cirrus: Make efficient and performant

cevich:
Reimplement CI-automation to remove accumulated technical-debt and
optimize workflow.  The task-dependency graph designed goal was to
shorten it's depth and increase width (i.e. more parallelism).  A
reduction in redundant building (and 3rd party module download) was
also realized by caching `$GOPATH` and `$GOCACHE` early on.  This
cache is then reused in favor of a fresh clone of the repository
(when possible).

Note: The system tests typically execute MUCH faster than the
integration tests.  However, contrary to a fail-fast/fail-early
principal, they are executed last.  This was implemented due to
debug-ability related concerns/preferences of the primary
(golang-centric) project developers.


cevich:
@baude I've stubbed in most of what I think we discussed on IRC.  Ignore all the deleted files, the only file that's important at the moment is https://github.com/containers/libpod/blob/26424fb62d5d522602bf7aa7752b482f1416f3d2/.cirrus.yml

(You may want to [read up on YAML anchors/aliases first](https://cirrus-ci.org/guide/tips-and-tricks/#sharing-configuration-between-tasks))

cevich:
working stubs: https://github.com/containers/libpod/blob/66ac189f285965daf01c9ca3cf0879aae8df8f88/.cirrus.yml

.cirrus.yml
```diff
@@ -39,742 +28,97 @@ env:
...
-verify_test_built_images_task:
-
-    only_if: >-
-        $CIRRUS_BRANCH != $DEST_BRANCH &&
-        $CIRRUS_CHANGE_MESSAGE =~ '.*CI:IMG.*' &&
-        $CIRRUS_CHANGE_MESSAGE !=~ '.*CI:DOCS.*'
-
+        - smoke
 
+verify_builds_task:
```
cevich, line 75: note to self: I ***hate** this task name.  This task is for making sure all the externally/manually tested builds (which will be part of the release) are produced.  For example, the current 'build without cgo' and 'build for darwin or windows' items should go here.

This job runs in parallel with integration and system testing, so it's window for including items like 'build each commit' is also large.  If the task gets too long for developer feedback convenience, it can be split up as-needed later on.

cevich:
@jwhonce What's the name for the "python tests with a local server" that you mentioned during scrum?  I want to make sure it's included in the new CI-job structure.

@edsantiago mind taking a peek at the names and comments here.  You spent a lot of time figuring how to make them pretty, I don't want to revert any of your efforts.

@baude This is the set of jobs and parallelism we talked about.  Please let me know what you think.

edsantiago:
>names and comments here. You spent a lot of time figuring how to make them pretty

That work is completely reverted already: Cirrus doesn't work as I had expected, and the effect of my cleanup commit was to disable podman-remote testing. But thanks.

cevich:
>That work is completely reverted already:

Oh sorry @edsantiago I think this is my bad wording.  I meant the work you did a long while ago so the task names are more human-friendly, without a bunch of numbers and gibberish in them :smiley:

edsantiago:
That's exactly what I meant: all that work is gone. Or at least most of it. 15f273b93

edsantiago:
@rhatdan has been working hard to reenable podman-remote tests (#6570), and made incremental progress, but still no luck.

edsantiago:
In the diffs I'm looking at, I see `gating` removed. In the github PR status page, I see:
```
* gating Expected - Waiting for status to be reported  (Required)
```
(See attached screenshot)
gating-missing

What am I missing?

BTW there's no way I can review this all in one day; I'll continue tomorrow

cevich:
@edsantiago thanks for looking.  

> all that work is gone. Or at least most of it.

Aww :cry: Are you sure the `name` changes are necessary?  What you learned about ssh and envvar passing seems **MUCH** more significant.  Maybe the problem you're hitting with it is simply the nested `env` -> `matrix` -> `env` blocks...they don't look valid to me (or maybe I'm seeing things).  Is there a PR I can look at?

> In the github PR status page, I see:

Ignore those. They're added by github due to repository-wide settings we have enabled.  We'll address those when this PR is closer to merging or merged.

> In the diffs I'm looking at, I see gating removed. In the github PR status page, I see:

Sorry, I should make that more clear in the description.  This change is in a comment but easily missed: I'm proposing we rename 'gate' -> 'smoke' for several reasons, but primarily so it doesn't cause confusion with the `gating` term used in downstream testing (i.e. RHEL and Fedora modules / dist-git).

For reference/background, the dependency-graph of the tasks here is based on discussions Brent and I had.  The central idea is to run as much in parallel as possible, counting on CIrrus-CI canceling everything upon the first failure (since `success` task depends on everything passing).  Also note: I'm planning to move all VM and container image building code and configuration into a dedicated repository (so no more multi-mode runtime magic, like `CI:IMG` or `CI:DOC`).

> I can review this all in one day; I'll continue tomorrow

Please take your time, this is only a prototype for a much larger piece of work I'm **not** eager to rush through :grin:

.cirrus.yml
```diff
@@ -39,742 +25,235 @@ env:
...
-        <<: *meta_env_vars
-        GCPJSON: ENCRYPTED[4c11d8e09c904c30fc70eecb95c73dec0ddf19976f9b981a0f80f3f6599e8f990bcef93c253ac0277f200850d98528e7]
-        GCPNAME: ENCRYPTED[7f54557ba6e5a437f11283a53e71baec9ca546f48a9835538cc54d297f79968eb1337d4596a1025b14f9d1c5723fbd29]
-
-    timeout_in: 10m
-
-    script: '/usr/local/bin/entrypoint.sh |& ${TIMESTAMP}'
-
+    FEDORA_CONTAINER_FQIN: "quay.io/libpod/in_podman_fedora:${DEST_BRANCH}"
+    UBUNTU_CONTAINER_FQIN: "quay.io/libpod/in_podman_ubuntu:${DEST_BRANCH}"
```
edsantiago, line 36: `contrib/curris/lib.sh` (and two others) still reference `IN_PODMAN_IMAGE` pointing to `quay.io/libpod/in_podman:` (as in just `in_podman`, not `in_podman_distro`). Do those need to be updated? (Also, how do these `in_podman_xxx` images get created and pushed to quay?)

cevich: Oh good catch, I was thinking of moving those to a separate follow-on PR (along with relocating the VM Image build stuff).  Now I'm realizing this PR needs to depend on that.  The reason is the answer to your other question:

> how do these in_podman_xxx images get created and pushed to quay?

We adopted the "easy-way" here, and it's outlived its usefulness.  It works like this:

1. A branch (like master) changes due to a PR merge or a tag push
2. On quay.io, we have many images defined, including `in_podman`.  Many have a build-trigger defined for any branch, though `in_podman` and several others have specific triggers on named branches (for historical reasons).
3. One/more of these triggers notice the repository change, and independently fire-off `docker build ...` jobs (within quay.io service).
4. Every build pushes both a `latest and named tag (like `master`, or 'v1.6', etc) depending on the build trigger configuration.
5. Each quay.io repo has notification settings configured to e-mail me when a build fails.

cevich:
**Major problems:** With current container & VM image build workflow

1. _Sometimes_ I read failure notification mails, and go check if there's at least one green build for each branch, on each repo.  Most of the time I can't be bothered.
2. When something breaks, I scramble to decode what went wrong in the past (sometimes with limited data), and somewhat blindly change the images so they seem to work.
3. When builds are working okay images are constantly updated on quay.io.  Depending on timing, this can mean somebody's in-flight tests on a PR pull and execute different images, possibly with different behavior, and very little observe-ability.
4. When someone wants to change a container image, they have to do so locally, push with a "test" tag, and fudge a PR to use that image.  Assuming testing passes, they merge the PR, wait for quay.io to build+tag the image and hope nobody's in-flight tests break.
5. There's a constantly changing 'latest' tag, which may come from a build on a branch nobody cares about currently.  It's super-duper easy to pull and try to use `latest` (for local testing) outside the context it was built for.  i.e. "latest" doesn't actually mean latest in all environment-contexts.

-----

What I propose is a separate repository with:

1. A standalone CI/CD setup, but authorized with the same google-cloud and quay.io services.
1. Separate branches that pair with dependent repositories and their release branches.  e.g. libpod-v1.6, buildah-v1.14, etc.
2. The standard PR -> Merge workflow is used on each branch where:
    1. We build VM and container images (including in_podman_fedora, gate, imgts, etc.) in every PR, and (maybe) periodically on each branch.
    2. We run repository specific smoke tests against the produced images.  This will mainly be very simplistic, like running 'make all' and checking the exit status.  Assuming all is well, the container images are tagged and pushed to quay.io.  The VM images simply stay where they are (we use the same imgts + imgprune schemes to maintain "current" images).
    3.  With the images-repo PR still open, a PR is opened in the dependent repo. referencing the just produced/pushed  named/numbered/tagged images.
    4. Regular/Full CI tests run for the dependent that repo, validating the soundness of the new image for it's purpose.
    5. If tests fail, adjustments are made in their respective repository PRs, and life goes on.  If everything looks good, the PRs are merged (order doesn't matter), and life goes on.
3. Independently of any open image PRs, all dependent repos. continue using their defined tagged/numbered images as normal, without interruption or surprises.
5. Once the image and dependent repo. PRs merge, other PRs need to rebase to use the new images, just as it is today, further limiting developer surprises.

rhatdan:
/approve
@cevich What is happening with this?

cevich:
> @cevich What is happening with this?

Lots of "change YAML" -> "Break Cirrus" cycles :grin:   This PR is my top priority ATM.

cevich:
@baude  When u have a mo.  Mind peeking at the task names here?

[ ] Make sure they read okay
[ ] Everything you want tested, is here
[ ] Anything permutation(s) that can be dropped?

https://github.com/containers/podman/pull/6784/checks

.cirrus.yml
```diff
@@ -16,604 +8,578 @@ env:
...
-
-    # HTMLify ginkgo and bats logs
-    LOGFORMAT: "${CIRRUS_WORKING_DIR}/${SCRIPT_BASE}/logformatter"
 
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
-    ###
+    ####
     FEDORA_NAME: "fedora"
     PRIOR_FEDORA_NAME: "prior-fedora"
```
edsantiago, line 25: Nooooooo! What happened? These were nice happy human-readable names until now, why are they changing to horrible un-numbered ones? Can this be reverted?

cevich: maybe but damn, there was a good reason I had to change it and now I can't remember.  I'll try putting it back and see what breaks.

cevich:
dependency-graph

edsantiago:
Can you merge [my PR](https://github.com/containers/automation/pull/49) and try again? It might look better.

cevich:
dependency-graph

cevich:
@edsantiago slightly different...I think your tool is plenty past the 80% "good enough" mark :smile:

I see there's a typo in the "Total Success" task, and I'd like to make at least one pass through all my comments...you know, fixing all the its -> it's :grin:

Then, I know @baude doesn't like bash, and I value your feedback, but man...the diff is HUGE.  Though I'd be perfectly happy with a even a cursory review, is there some way I can make it any easier to look at?

edsantiago:
oh, yuk - that's not what I was expecting. I had it tuned for a different revision :frowning_face: . Oh well.

We talked about this in Watercooler this week. There's no way to really review it. We just have to take the plunge. I will look at it over the next two hours and give it my best shot, but unfortunately I will be net-unavailable all of tomorrow.

test/system/500-networking.bats
```diff
@@ -82,6 +82,9 @@ load helpers
 
 # "network create" now works rootless, with the help of a special container
 @test "podman network create" {
+    if ! is_rootless; then
+        skip_if_remote "FIXME: Test flakes very frequently for yet-unknown reason(s)"
```
edsantiago, line 86: To make this easier for the 7808-fixer-person to find, please say something like "FIXME: pending #7808"

cevich: oh there's an open issue? Excellent, yes I'll pop that in there.

cevich:
> We talked about this in Watercooler this week. There's no way to really review it. We just have to take the plunge. I will look at it over the next two hours and give it my best shot, but unfortunately I will be net-unavailable all of tomorrow.

Ya, it's a beast, I completely understand, and would greatly appreciate even 30-minutes of your eyeball time :grin: 

Thanks

contrib/cirrus/lib.sh
```diff
@@ -3,293 +3,188 @@
...
 # Global details persist here
-source /etc/environment  # not always loaded under all circumstances
+source /etc/environment || touch /etc/environment  # needed on Mac OSX
+
+# Under some contexts these values are not set, make sure they are.
+USER="$(whoami)"
+HOME="$(getent passwd $USER | cut -d : -f 6)"
+[[ -n "$UID" ]] || \
+    UID=$(getent passwd $USER | cut -d : -f 3)
+GID=$(getent passwd $USER | cut -d : -f 4)
```
edsantiago, line 17: Whew. I see that this is simply a move of already-existing code, but *shudder*. Some day (not this iteration!) I'd like to understand this better, see which systems it fails on, use `${...:-...}` to set only if missing, and perhaps use `id` instead of getent. ITM if you need to repush for some other reason, a quick FIXME here might be nice.

contrib/cirrus/lib.sh
```diff
@@ -3,293 +3,188 @@
...
 
-# Automation environment doesn't automatically load for Ubuntu 18
+# Automation environment doesn't automatically load for non-interactive shells
 if [[ -r '/usr/share/automation/environment' ]]; then
     source '/usr/share/automation/environment'
+    for libname in defaults anchors console_output utils; do
+        # Pre-installed into environment
+        # shellcheck disable=SC1090,SC2154
+        source $AUTOMATION_LIB_PATH/${libname}.sh
```
edsantiago, line 25: I don't see `$AUTOMATION_LIB_PATH` defined anywhere! What am I missing?

edsantiago:
First pass finished (on scripts; I didn't even look at `.cirrus.yml`). It LGTM. Some nits, but nothing I'm even going to bother mentioning. They can be fine-tuned over time. The important thing is that CI passes (mostly) and that it seems to be quicker than our current method. I think we all want to see this go live.

contrib/cirrus/lib.sh
cevich: I think this was for Ubuntu, during initial implementations, though I added the comment later on.  In hind-sight, I think I understand why this was needed and am willing to try removing it.

contrib/cirrus/lib.sh
cevich: and ***I*** see an opportunity for a more helpful comment :grinning:

contrib/cirrus/lib.sh
cevich: While attempting to remove this, I re-discovered why it's necessary and added a helpful comment.  The primary reason we need it is for the tasks which run inside a container.  In those cases, things like `$HOME` and `$USER` are ***not*** set - perhaps due to the way cirrus executes containers.

cevich:
@baude @mheon PTAL, this is ready to be manually merged.

The "Optional release test" failure is expected.  I triggered it manually to confirm it correctly fails when finding `-dev` in the output of `podman info` ("2.2.0-dev").  This is most certainly the case, since this PR is not for a podman release :smile: Normally this task won't run, unless somebody presses the "trigger" button in the WebUI.

***note:*** After/before this PR merges, I/we need to update the repo. settings to make the 'smoke' and 'total success' tasks as required (under branch protection).  All in-flight PRs will need to rebase as well, so might be good to merge this during a "quiet" period.

edsantiago:
Quick note in case you need to resubmit (please don't do so just for this): `logformatter` also works on (and vastly improves) system tests.

cevich:
> (Deleted my earlier comment. Never mind. Let's just get this in, then all of us as a team can take care of making future changes to the parts that affect us)

Agreed, parallelize the work :grinning:

rhatdan:
Needs @baude approval.

PR #6790, opened 2020-06-26
Set engine env from common config

QiWang19:
vendor commond PR https://github.com/containers/common/pull/198 to set container_engine_env

related issue  https://github.com/containers/common/issues/31

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -128,6 +128,21 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	for _, env := range cfg.Engine.Env {
+		splitEnv := strings.SplitN(env, "=", 2)
+		if len(splitEnv) != 2 {
+			return fmt.Errorf("invalid environment variable for engine %s, valid configuration is KEY=value pair", env)
+		}
+		// skip if the env is already defined
+		if _, ok := os.LookupEnv(splitEnv[0]); ok {
+			logrus.Debugf("environment variable %s is already defined, skip the settings from containers.conf", splitEnv[0])
+			continue
+		}
+		if err := os.Setenv(splitEnv[0], splitEnv[1]); err != nil {
+			return err
+		}
+	}
+
 	if cmd.Flag("cpu-profile").Changed {
 		f, err := os.Create(cfg.CPUProfile)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


QiWang19:
@rhatdan @mheon  PTAL

cmd/podman/root.go
```diff
@@ -128,6 +128,18 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	for _, env := range cfg.Engine.Env {
+		splitEnv := strings.SplitN(env, "=", 2)
+		key := splitEnv[0]
+		value := ""
+		if len(splitEnv) == 2 {
+			value = splitEnv[1]
+		}
```
rhatdan, line 137: If it is not 2, this should also be an error.

rhatdan: I see what you did, but users might accidentally override an existing environement variable with nothing.
If they want nothing then they should set it 
foo=""

rhatdan:
LGTM
@mheon @baude @giuseppe @vrothberg @TomSweeneyRedHat @ashley-cui @jwhonce PTAL

cmd/podman/root.go
```diff
@@ -128,6 +128,16 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	for _, env := range cfg.Engine.Env {
+		splitEnv := strings.SplitN(env, "=", 2)
```
mheon, line 132: Do we want this to take precedence over actual OS environment? I imagine that if I have something set in containers.conf but later do a `ENVVAR=value podman run ...` I want the version passed there to take precedence.

If so, we should only define environment variables that are not already defined.

cmd/podman/root.go
```diff
@@ -128,6 +128,16 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
...
+		if len(splitEnv) != 2 {
+			return fmt.Errorf("invalid environment variable for engine %s, valid configuration is KEY=value pair", env)
+		}
+		if err := os.Setenv(splitEnv[0], splitEnv[0]); err != nil {
```
giuseppe, line 136: should this be ` os.Setenv(splitEnv[0], splitEnv[1])` ?

QiWang19: yes. fixed....

cmd/podman/root.go
rhatdan: Yes I guess that makes sense.  @QiWang19 Please add a note to the man pages on podman (buildah) about this, also, just do a check to see if environment is set, it set write a debug message, if different, and don't override,

QiWang19:
@rhatdan @mheon PTAL

PR #6792, opened 2020-06-26
[CI:DOCS]Add swagger.yaml to docs/

baude:
Adding the swagger.yaml to the docs directory so that we can version the read-the-docs API information. also, change the links to be relative in nature and point to the new swagger.


mheon:
LGTM

mheon:
Really wish we had a way to update this automatically on release, but we can look into that later
/lgtm

PR #6794, opened 2020-06-26
Set console mode for windows

baude:
Windows terminal handling is different than darwin and linux.  It needs to have the terminal mode set to enable virtual terminal processing.  This allows colors and other things to work.

### DIFF
--- a/cmd/podman/main.go
+++ b/cmd/podman/main.go
@@ -16,7 +16,9 @@ import (
 	_ "github.com/containers/libpod/cmd/podman/system"
 	_ "github.com/containers/libpod/cmd/podman/volumes"
 	"github.com/containers/libpod/pkg/rootless"
+	"github.com/containers/libpod/pkg/terminal"
 	"github.com/containers/storage/pkg/reexec"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/terminal/console_supported.go
```diff
@@ -0,0 +1,8 @@
+// +build !windows
```
mheon, line 1: Can we rename this to `console_unix.go`?

cmd/podman/main.go
```diff
@@ -53,6 +56,10 @@ func main() {
 			}
 		}
 	}
+	if err := terminal.SetConsole(); err != nil {
+		logrus.Error(err)
+		panic(1)
```
mheon, line 61: Recommend `os.Exit` instead

cmd/podman/main.go
```diff
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/containers/libpod/pkg/terminal"
+	"github.com/sirupsen/logrus"
+
```
rhatdan, line 9: Remove blank line

pkg/terminal/console_windows.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func setConsoleMode(handle windows.Handle, flags uint32) error {
+	var mode uint32
+	err := windows.GetConsoleMode(handle, &mode)
+	if err != nil {
+		return err
+	}
+	if err := windows.SetConsoleMode(handle, mode|flags); err != nil {
+		logrus.WithError(err).Error("Failed to set console mode for cli")
```
rhatdan, line 31: Should this return the error, if not could you add a comment as to why it should not?

baude:
fixed

mheon:
LGTM

TomSweeneyRedHat:
LGTM

Issue #6793, opened 2020-06-26
auto-update no PODMAN_SYSTEMD_UNIT label found


q-b:
Hi,

While playing with podman last version on CentOS 8, I was unable to use the auto-update feature.

```
# podman auto-update
Trying to pull ...
<redacted>
... signatures
Error: 1 error occurred:
	* error auto-updating container "<redacted>": no PODMAN_SYSTEMD_UNIT label found
```

https://github.com/containers/libpod/blob/0968f25988624699f2db559f0a984cf84c8a284f/pkg/autoupdate/autoupdate.go#L184

The service file looks good (I checked it on disk and daemon-reloaded tu be sure):
```
...
[Service]
Environment=PODMAN_SYSTEMD_UNIT=%n
...
```

```
# podman inspect ...
...
            "Labels": {
                "io.containers.autoupdate": "image",
...
            "CreateCommand": [
...
                "--label",
                "io.containers.autoupdate=image",
...
```
```
# podman version
Version:      2.0.1
API Version:  1
Go Version:   go1.13.4
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

Let me know if you need more input to troubleshoot this issue!

mheon:
@vrothberg PTAL

vrothberg:
Thanks for opening the issue! 

Indeed, we regressed on it. Absolute time to extend the systemd tests...

vrothberg:
And as it looks like we regressed with 2.0.1. 2.0.0 is working...

vrothberg:
Master is fine though.

vrothberg:
I now fail to reproduce again. Very curious.

I used the following script to reproduce:
```bash
./bin/podman rm -af
CID=`./bin/podman create --label "io.containers.autoupdate=image" -d busybox:latest top`
./bin/podman generate systemd --new $CID > ~/.config/systemd/user/container-test.service
systemctl --user daemon-reload
systemctl --user stop container-test && systemctl --user start container-test && ./bin/podman inspect --format "{{.Config.Labels}}" -l
./bin/podman auto-update
systemctl --user stop container-test
```

vrothberg:
@edsantiago could you have a quick look and try to reproduce with 2.0.1?

vrothberg:
Opened https://github.com/containers/libpod/pull/6810 to add a test that would catch that case.

edsantiago:
I can't find any way to reproduce; am trying root on f32.

[...]

PR #6810, opened 2020-06-29
systemd system test: run auto-update

vrothberg:
Run `podman auto-update` in the systemd system tests.  Note that this is
a first step to at least exercise parts of `auto-update` in the CI.  The
service won't get updated just yet as we need to set up a local
registry, and push a new image.  I do not have enough time at the moment
to do that but consider this change already as an improvement.

We are experiencing some issues in #6793 w.r.t. to auto-updates but
couldn't track down the root cause yet.

### DIFF
--- a/test/system/250-systemd.bats
+++ b/test/system/250-systemd.bats
@@ -41,7 +41,7 @@ function teardown() {
     fi
 
     cname=$(random_string)
-    run_podman create --name $cname --detach $IMAGE top
+    run_podman create --name $cname --label "io.containers.autoupdate=image" --detach $IMAGE top
 
     run_podman generate systemd --new $cname
     echo "$output" > "$UNIT_FILE"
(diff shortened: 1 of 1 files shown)


edsantiago:
Testing with a local registry will be difficult unless `auto-update` can be taught to use `--tls-verify=false`

vrothberg:
> Testing with a local registry will be difficult unless `auto-update` can be taught to use `--tls-verify=false`

Good point. We could either extend the CLI and add that as a flag or we create a custom registries.conf where we mark the registry as insecure.

vrothberg:
Repushed to get tests running again. GitHub was down today.

vrothberg:
@mheon, I want this in the v2 branch as well. Just to be sure.

mheon:
We really need a backports label...

TomSweeneyRedHat:
Changes LGTM
but tests still seem to be dragging.

vrothberg:
Ready to merge :)

Issue #2669, opened 2019-03-15
podman save not saving multiple images


Prodian0013:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**
/kind bug

**Description**
When executing a `podman save` with multiple images only one image is saved.

**Steps to reproduce the issue:**

1. `podman save > images.tar docker.io/busybox:1.27.2 docker.io/metallb/controller:v0.3.1 docker.io/metallb/speaker:v0.3.1`

2. `podman import images.tar`

3. `podman images`

**Describe the results you received:**
Only the first image is saved and imported.

**Describe the results you expected:**
All three images should be saved and imported.  

The documentation states `podman save` is the equivalent to `docker save`.  Where as `docker save` can actually save multiple images at once. 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 1.1.2
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.10.2
  podman version: 1.1.2
host:
  BuildahVersion: 1.7.1
  Conmon:
    package: podman-1.1.2-2.git0ad9b6b.el7.x86_64
    path: /usr/libexec/podman/conmon
    version: 'conmon version 1.14.0-dev, commit: 6e07c13bf86885ba6d71fdbdff90f436e18abe39-dirty'
  Distribution:
    distribution: '"centos"'
    version: "7"
  MemFree: 96034816
  MemTotal: 3973865472
  OCIRuntime:
    package: runc-1.0.0-59.dev.git2abd837.el7.centos.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.0'
  SwapFree: 2147217408
  SwapTotal: 2147479552
  arch: amd64
  cpus: 2
  hostname: controller.lan
  kernel: 3.10.0-957.5.1.el7.x86_64
  os: linux
  rootless: false
  uptime: 59m 48.22s
insecure registries:
  registries: []
registries:
  registries:
  - registry.access.redhat.com
  - docker.io
  - registry.fedoraproject.org
  - quay.io
  - registry.centos.org
store:
  ConfigFile: /etc/containers/storage.conf
  ContainerStore:
    number: 0
  GraphDriverName: overlay
  GraphOptions: null
  GraphRoot: /var/lib/containers/storage
  GraphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
[...]

PR #6811, opened 2020-06-29
podman load/save: support multi-image docker archive

vrothberg:
Fixes: #2669


vrothberg:
Need to get https://github.com/containers/image/pull/975 in before.

libpod/image/pull.go
```diff
@@ -113,6 +113,52 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
+		}
+		return ir.getSinglePullRefPairGoal(srcRef, digest)
+	}
+
+	// Need to load in all the repo tags from the manifest
+	res := []pullRefPair{}
+	for _, dst := range manifest.RepoTags {
+		// Check if image exists and give a warning of untagging
+		localImage, err := ir.NewFromLocal(dst)
+		imageID := strings.TrimSuffix(manifest.Config, ".json")
```
mtrmac, line 143: (I know this is pre-existing.)

https://github.com/moby/moby/blob/749d90e10f989802638ae542daf54257f3bf71f2/image/spec/v1.2.md#combined-image-json--filesystem-changeset-format does not say the file name of `manifest.Config` is a digest; this should read the manifest and use `ConfigInfo`, or maybe exactly `c/image/Manifest.ImageID` like c/image/storage does.

libpod/image/image.go
```diff
@@ -169,15 +170,169 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+		if err := multiDest.Finalize(ctx); err != nil {
+			if finalErr == nil {
+				finalErr = err
+			} else {
+				finalErr = errors.Wrap(finalErr, err.Error())
+			}
+		}
+	}()
+
+	manifestMIMEType := multiDest.SupportedManifestMIMETypes()[0]
```
mtrmac, line 204: Why is this necessary? `c/image/copy` should automatically convert.

libpod/image/image.go
```diff
@@ -169,15 +170,169 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+
+	// Check whether c/image's progress bars use stderr or stdout.  Use
+	// stderr in case we need to be quiet or if the output is set to
+	// stdout.  If the output is set of stout, any log message there would
```
mtrmac, line 208: s/Check whether … use/Decide whether … should use/ ?

libpod/image/image.go
```diff
@@ -169,15 +170,169 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+		// Unless we referred to an ID, add the input as a tag.
+		id := localImage.ID()
+		if strings.HasPrefix(id, nameOrID) {
+			continue
+		}
+		tag, err := NormalizedTag(nameOrID)
+		if err != nil {
+			return err
+		}
+		tags = append(tags, tag.String())
```
mtrmac, line 243: `tag` already is a `reference.Named`, only for this to lose that and use a string representation, and `GetAdditionalTags` later to undo that again (and silently skip non-tag values it seems).

libpod/image/image.go
```diff
@@ -169,15 +170,169 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+		imageTags[localImage] = tags
+	}
+
+	// Now copy the images one-by-one.
+	copyOptions := getCopyOptions(sys, writer, nil, nil, SigningOptions{}, manifestMIMEType, nil)
+	copyOptions.DestinationCtx.SystemRegistriesConfPath = registries.SystemRegistriesConfPath()
```
mtrmac, line 249: (I know this is pre-existing.)

Set this when creating `sys`, and `getCopyOptions` will AFAICS propagate that.

libpod/image/image.go
mtrmac: Was anything changed in the comment or code? Also, s/stout/stdout/`.

vrothberg: The comment didn't (my bad) but the code did.

vrothberg:
> The changes to `libpod/image/pull.go` were dropped — is that intentional?
> 
> In a sense they really don’t belong into this PR, OTOH fixing the existing fairly broken `podman pull docker-archive:…` code would be nice.

That was intentional as I want us to focus on the multi-image part first. Once that's in, we can tackle the next issue - unless you feel strongly to tackle both at once.

mtrmac:
I’m fine with having the PR narrowly scoped; what is/isn’t necessary to keep the Podman repo maintainable is up to the judgment of Podman maintainers.

pkg/domain/entities/images.go
```diff
@@ -269,11 +269,21 @@ type ImageImportReport struct {
...
-	Format   string
-	Output   string
-	Quiet    bool
+	// Format of saving the image: oci-archive, oci-dir (directory with oci
+	// manifest type), docker-archive, docker-dir (directory with v2s2
+	// manifest type).
+	Format string
+	// MultiImageArchive denotes if the created archive shall include more
+	// than one image.
+	MultiImageArchive bool
```
mtrmac, line 282: This affects behavior of the “tags” parameter of `ImageEngine.Save`, which is not a part of these options… possible but complicated, starting with “this field is difficult to document”.

libpod/image/image.go
```diff
@@ -173,15 +174,208 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+		}
+		finalErr = errors.Wrap(finalErr, err.Error())
+	}()
+
+	// Decide whether c/image's progress bars use stderr or stdout.  Use
+	// stderr in case we need to be quiet or if the output is set to
+	// stdout.  If the output is set of stout, any log message there would
+	// corrupt the tarfile.
+	writer := os.Stdout
+	if quiet || outputFile == os.Stdout.Name() {
```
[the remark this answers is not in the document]
vrothberg: I can move it up to `cmd/podman` :+1:

libpod/image/image.go
```diff
@@ -173,15 +174,208 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+		iData, exists := imageMap[id]
+		if !exists {
+			imageQueue = append(imageQueue, id)
+			iData = &imageData{Image: localImage}
+			imageMap[id] = iData
+		}
+
+		// Unless we referred to an ID, add the input as a tag.
+		if strings.HasPrefix(id, nameOrID) {
+			continue
```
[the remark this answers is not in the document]
vrothberg: Sure, I can change.

I see this case as an early return, where I prefer `if cond { return } // do sth` over `if !cond { // do sth }`.

libpod/image/image.go
```diff
@@ -173,15 +174,208 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+		}
+		tag, err := NormalizedTag(nameOrID)
+		if err != nil {
+			return err
+		}
+		refTagged, _ := tag.(reference.NamedTagged)
```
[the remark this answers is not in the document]
vrothberg: I guess we need to error out.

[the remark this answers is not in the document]
vrothberg: Or ignore?

libpod/image/image.go
```diff
@@ -173,15 +174,208 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile
...
+
+	return nil
+}
+
+func (ir *Runtime) loadMultiImageArchive(ctx context.Context, sys *types.SystemContext, ref types.ImageReference, writer io.Writer) (loadedImages []string, finalErr error) {
+	if ref.Transport().Name() != DockerArchive {
+		return nil, errors.Errorf("wrong transport %q: only %q supports multi-image archives", ref, DockerArchive)
+	}
+
+	archReader, err := archive.NewReader(sys, ref.StringWithinTransport())
```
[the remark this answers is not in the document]
vrothberg: Right. I believe we can consolidate quite a lot into `pullGoalFromImageReference` actually.

libpod/image/image.go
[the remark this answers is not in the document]
mtrmac: Yeah, it’s really a personal preference, and the after-`continue` case here is not really big. “Absolutely non-blocking.”

libpod/image/image.go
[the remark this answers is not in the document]
mtrmac: Error out; we can’t process users’ input.

(Ideally, the CLI level should do the parsing, but that would require us to introduce a dockerReferenceOrID type… and ugh, Golang. This is early enough that just returning an error at this point is works for me.)

libpod/image/pull.go
```diff
@@ -114,64 +114,113 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
-				logrus.Errorf("the image %s already exists, renaming the old one with ID %s to empty string", dst, localImage.ID())
-			}
+	return refPairs, nil
+}
 
-			pullInfo, err := ir.getPullRefPair(srcRef, dst)
-			if err != nil {
-				return nil, err
+// getPullRefPairFromDockerReference returns a slice of pullRefPairs for the specified docker reference.
+func (ir *Runtime) getPullRefPairsFromDockerReference(ctx context.Context, ref types.ImageReference, sc *types.SystemContext) ([]pullRefPair, error) {
```
mtrmac, line 176:
```suggestion
func (ir *Runtime) getPullRefPairsFromDockerArchiveReference(ctx context.Context, ref types.ImageReference, sc *types.SystemContext) ([]pullRefPair, error) {
```

libpod/image/pull.go
```diff
@@ -114,64 +114,113 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
+	// Let's first assume it's a file-only reference.
+	reader, err := archive.NewReader(sc, ref.StringWithinTransport())
```
mtrmac, line 178: No; the caller knows whether the intent is to read everything in a file, or to “pull” a single image. No need to guess.

libpod/image/pull.go
```diff
@@ -114,64 +114,113 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
+		}()
+		return ir.getPullRefPairsFromDockerArchiveReader(ctx, reader, ref, sc)
+	}
+
+	var destName string
+	// Determine the destination. If it's a tagged item, we
+	// can use it directly, otherwise we need to determine
+	// the image digest.
+	if dkrRef := ref.DockerReference(); dkrRef != nil {
+		destName = ref.DockerReference().String()
```
mtrmac, line 193: Doesn’t this change behavior to always use the `docker.io/library` normalization? I’d expect `reader.ManifestTagsForReference` to be invoked here.

libpod/image/pull.go
```diff
@@ -114,64 +114,113 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
+	} else {
+		destName, err = getImageDigest(ctx, ref, sc)
```
mtrmac, line 195: Same here, the previous behavior was to use a tag from the manifest.

libpod/image/pull.go
```diff
@@ -114,64 +114,113 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
-			pullInfo, err := ir.getPullRefPair(srcRef, dst)
-			if err != nil {
-				return nil, err
+// getPullRefPairFromDockerReference returns a slice of pullRefPairs for the specified docker reference.
+func (ir *Runtime) getPullRefPairsFromDockerReference(ctx context.Context, ref types.ImageReference, sc *types.SystemContext) ([]pullRefPair, error) {
+	// Let's first assume it's a file-only reference.
+	reader, err := archive.NewReader(sc, ref.StringWithinTransport())
+	if err == nil {
+		defer func() {
+			if err := reader.Close(); err != nil {
```
mtrmac, line 181: This closes the `reader` before any references are used; that won’t work.

libpod/image/pull.go
```diff
@@ -114,64 +114,113 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
+
+// pullGoalFromImageReference returns a pull goal for a single ImageReference, depending on the used transport.
+func (ir *Runtime) pullGoalFromImageReference(ctx context.Context, srcRef types.ImageReference, imgName string, sc *types.SystemContext) (*pullGoal, error) {
+	span, _ := opentracing.StartSpanFromContext(ctx, "pullGoalFromImageReference")
+	defer span.Finish()
+
+	// supports pulling from docker-archive, oci, and registries
+	switch srcRef.Transport().Name() {
+	case DockerArchive:
+		pairs, err := ir.getPullRefPairsFromDockerReference(ctx, srcRef, sc)
```
mtrmac, line 216: (Ideally, I’d expect an `archive.Reader` created somewhere around here to persist until the pull is done. OTOH that’s an optimization that can wait for a separate PR.)

libpod/image/pull_test.go
```diff
@@ -150,33 +150,25 @@ func TestPullGoalFromImageReference(t *testing.T) {
...
 			"docker-archive:testdata/docker-unnamed.tar.xz",
 			[]expected{{"@ec9293436c2e66da44edb9efb8d41f6b13baf62283ebe846468bc992d76d7951", "@ec9293436c2e66da44edb9efb8d41f6b13baf62283ebe846468bc992d76d7951"}},
-			false,
+			true,
 		},
 		{ // RepoTags is a [docker.io/library/]name:latest, normalized to the short format.
-			"docker-archive:testdata/docker-name-only.tar.xz",
-			[]expected{{"localhost/pretty-empty:latest", "localhost/pretty-empty:latest"}},
+			"docker-archive:testdata/docker-name-only.tar.xz:pretty-empty:latest",
+			[]expected{{"docker.io/library/pretty-empty:latest", "docker.io/library/pretty-empty:latest"}},
```
mtrmac, line 157: The `expected.dstName` here seems wrong.

libpod/image/pull_test.go
```diff
@@ -150,33 +150,25 @@ func TestPullGoalFromImageReference(t *testing.T) {
...
 		{ // RepoTags has multiple items for a single image
-			"docker-archive:testdata/docker-two-names.tar.xz",
+			"docker-archive:testdata/docker-two-names.tar.xz:example.com/empty:latest",
 			[]expected{
-				{"localhost/pretty-empty:latest", "localhost/pretty-empty:latest"},
 				{"example.com/empty:latest", "example.com/empty:latest"},
 			},
 			true,
 		},
-		{ // FIXME: Two images in a single archive - only the "first" one (whichever it is) is returned
```
mtrmac, line 173: I’d expect a test to remain with `len(expected) == 0` to test that this is rejected.

libpod/runtime_img.go
```diff
@@ -278,9 +278,16 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
+		newImages, err = r.ImageRuntime().LoadAllImagesFromDockerArchive(ctx, inputFile, signaturePolicy, writer)
+		if err == nil {
+			return getImageNames(newImages), nil
+		}
+	}
+
 	for _, referenceFn := range []func() (types.ImageReference, error){
 		func() (types.ImageReference, error) {
-			return dockerarchive.ParseReference(inputFile) // FIXME? We should add dockerarchive.NewReference()
+			return dockerarchive.ParseReference(inputFile)
```
mtrmac, line 290: So now `name` is completely ignored for `docker-archive:` AFAICS.

Instead, this should do `ParseReference(inputFile+":"+name)`, or even better `ParseNormalizedNamed`+`dockerachive.NewReference` (would reject a "name" `@0`)

vrothberg: This was breaking backwards compat (and tests).
`$ podman load -i foo.tar name:tag` is supposed to load the archive and tag the (single) image with `name:tag`.

mtrmac: How does that work, when `name` is _not used at all_ now?

libpod/image/pull.go
vrothberg: But we don't without breaking backwards compatibility with `podman pull docker-archive:...` which expects all RepoTags to be loaded. Or am I missing something in the c/image PR that I could use?

mtrmac: I thought we agreed that using `podman pull` to pull more than one _image_ would be dropped. If it is a single-image archive, `ManifestTagsForReference` lists all tags for that image.

libpod/image/pull.go
vrothberg: It's working in this case, I am enforcing the original reference.

What I am trying to do here is to extract all RepoTags from a given reference. I didn't find a way other than using an `archive.Reader`.

libpod/image/pull.go
vrothberg: Initially yes, but I thought we agreed yesterday to keep extracting all RepoTags to keep the current behaviour.

libpod/image/pull.go
mtrmac: > It's working in this case, I am enforcing the original reference.

Ah, yes. But that’s a very localized override, so `getPullRefPairsFromDockerArchiveReader` also accepts multi-image archives, and silently creates (that-original-reference, any-tag-for-any-image-in-archive) pairs AFAICS.

> What I am trying to do here is to extract all RepoTags from a given reference.

`NewReaderForReference`+`ManifestTagsForReference`.

> I didn't find a way other than using an `archive.Reader`.

Yes, that’s intentional, to make it annoyingly difficult to extract `RepoTags` without creating the `archive.Reader`, so that the easy way to use the API does not involve creating the temporary archive twice. 

(If you _really don’t want to_ keep a `Reader` alive throughout the pull, `docker archive.New{Index,}Reference` does allow creating reader-independent references for the individual names/images in an archive. But even a _short-lived_ Reader for the purpose of the heuristics creating ref pairs would speed up things like `getImageDigest`, so please use that, and maybe disassociate from the `Reader` only for the ref pair.)

libpod/image/pull.go
mtrmac: I thought that meant “all tags _for a single image_” — after all, current `podman pull` only ever uses `manifest[0]`. There’s no precedent for `podman pull` actually reading more than one image AFAICS (if only because docker/archive never made that possible).

libpod/image/pull.go
vrothberg: Fixed

libpod/image/pull.go
vrothberg: Thanks again for the guidance. The reader now persists across the entire pull (moved back the cleanUp idea we had earlier).

libpod/image/pull.go
vrothberg: Now done. Thanks for pushing me to do the right thing :)

libpod/image/pull.go
vrothberg: fixed

libpod/runtime_img.go
vrothberg: That's done after in https://github.com/containers/podman/blob/master/pkg/domain/infra/abi/images.go#L465.

It's quite a spaghetti :spaghetti:

mtrmac: That uses a _return value_ from this function. I can’t see any way the `name` parameter to this function affects the return value in the docker-archive case.

vrothberg:
@baude @jwhonce PTAL at the API parts

libpod/runtime_img.go
mtrmac: Oops, sorry; it does use `opts.Name` to decide what to tag — but still, this function does not use it to decide what to pull.

Either the `name` parameter indicates what to pull, and in that case it should do so for `docker-archive`, or it doesn’t, and in that case it should be dropped from this function, changing behavior of all the other transports.

vrothberg: If we do `input+":"+name` the semantics change. Before, we were loading whatever is inside and add `name` as a tag to it. Now we would expect `name` to be present in the image.

mtrmac: Aargh, so this is a Podman invention, not a Docker compatibility CLI.

We have different behavior in different transports, and none of `docs/podman-load.1`, `cmd/podman/images/load.go`, `ImageLoadOptions` actually document what the “name” field means. What a _mess_.

I can’t spend a lot of time on this now; I guess cleaning that up (if it should be cleaned up) is well outside the scope of this PR, but I’d at least like to insist on very explicit documentation  (end-user and internal throughout the API / data structures) of what the expected behavior actually is.

vrothberg: I will add a comment and open an issue so we don't forget about it.

pkg/api/handlers/libpod/images.go
```diff
@@ -234,6 +234,76 @@ func ExportImage(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+
+	// References are mandatory!
+	if len(query.References) == 0 {
```
jwhonce, line 258: You could use .Queries("References", "*") when registering this endpoint. The framework would take care of the validation.  "Format doesn't work as well, since you have a non-Zero default for that parameter.

libpod/runtime_img.go
zhangguanzhang: > 我将添加评论并打开一个问题，这样我们就不会忘记它。

https://github.com/containers/podman/issues/7337 I think it's the same issue

vrothberg:
Removed WIP as we're vendoring c/image directly.

NOTE: we are vendoring c/image@master and need to cut a c/image release before cutting Podman v2.1 (Cc @mheon)

vrothberg:
Added more tests for `pull` to exercise the new syntax for docker-archives.

vrothberg:
storage@v1.23.1 is causing a regression in the `apiv2` tests....

vrothberg:
Opened https://github.com/containers/storage/issues/697 and https://github.com/containers/image/pull/1031 to temporarily revert the change in c/image.

vrothberg:
@edsantiago are https://github.com/containers/podman/pull/6811/checks?check_run_id=1007722626 flakes?

@mtrmac, I am running out of time. If you find some cycles, feel free to take over this PR and push it over the finish line.

edsantiago:
@vrothberg those show every sign of being #7195 but I've never seen so many in one run.

vrothberg:
It's green, it's green!

@rhatdan, let's merge before the CI gods stop shining upon me head.

libpod/image/pull.go
```diff
@@ -114,65 +128,76 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
...
+		destNames = append(destNames, destName)
+	} else {
+		for i := range destNames {
+			ref, err := NormalizedTag(destNames[i])
+			if err != nil {
+				return nil, err
+			}
+			destNames[i] = ref.String()
+		}
+	}
```
mtrmac, line 153: (Aesthetically I’m not a great fan of using `destNames` both for the archive tags and for storage references. It works fine, sure.)

libpod/image/pull.go
```diff
@@ -277,9 +302,11 @@ func cleanErrorMessage(err error) string {
 }
 
 // doPullImage is an internal helper interpreting pullGoal. Almost everyone should call one of the callers of doPullImage instead.
+// Note that the pullGoal is cleaned up.  Some of it's resources may not be available anymore.
```
mtrmac, line 305:
```suggestion
// Note that the pullGoal is cleaned up.  Some of its resources may not be available anymore.
```

mtrmac, line 305: This works just fine, but it feels fragile. Anyone adding an extra error/return path to callers of `doPullImage` is fairly likely to overlook this. Requiring owners of a `pullGoal` to explicitly `defer goal.cleanUp()` would make this easier to see.

libpod/image/pull_test.go
```diff
@@ -169,18 +177,56 @@ func TestPullGoalFromImageReference(t *testing.T) {
...
+			[]expected{},
+			true,
+			false,
 			true,
 		},
 
 		// == oci-archive:
-		{"oci-archive:/dev/this-does-not-exist", nil, false}, // Input does not exist.
-		{"oci-archive:/dev/null", nil, false},                // Input exists but does not contain a manifest.
+		{"oci-archive:/dev/this-does-not-exist", nil, false, false, false}, // Input does not exist.
```
mtrmac, line 228: I know in the error cases the `compareSourceRefs` value does not matter, but will later be hard to tell whether the `false` is an intentional property of `oci-archive` or not. “Everything but docker-archive has this set to `true`” is a simpler story to tell.

libpod/image/pull_test.go
```diff
@@ -243,12 +299,19 @@ func TestPullGoalFromImageReference(t *testing.T) {
...
+			}
 			require.NoError(t, err, c.srcName)
 			require.Len(t, res.refPairs, len(c.expected), c.srcName)
 			for i, e := range c.expected {
 				testDescription := fmt.Sprintf("%s #%d", c.srcName, i)
 				assert.Equal(t, e.image, res.refPairs[i].image, testDescription)
-				assert.Equal(t, srcRef, res.refPairs[i].srcRef, testDescription)
+				if c.compareSourceRefs {
+					// They're not equal in case of a docker-archive.
+					assert.Equal(t, srcRef, res.refPairs[i].srcRef, testDescription)
```
mtrmac, line 313: This fails just because it is comparing pointers for equality, doesn’t it? Could this instead compare `transports.ImageName(…)` values, allowing us to get rid of the `compareSourceRefs` option?

libpod/image/pull_test.go
```diff
@@ -243,12 +299,19 @@ func TestPullGoalFromImageReference(t *testing.T) {
 		if len(c.expected) == 0 {
 			assert.Error(t, err, c.srcName)
 		} else {
+			if c.mustFail {
```
mtrmac, line 302: There’s already the `len(c.expected) == 0` check just above. Either just use it, or remove it and convert all its users to `mustFail`. Things like
```go
{"docker-archive:/dev/this-does-not-exist", nil, false, false, false}, // Input does not exist.
```
above are internally inconsistent.

mtrmac (commented):
ACK (full review now), just some maintainability suggestions.

mtrmac:
Also this vendors an unreleased version of c/image. Is that OK?

mheon:
OK for now, but we'll need to get a c/image release out in the next several weeks for Podman 2.1.0 so we don't cut a full release with a prerelease c/image

libpod/image/pull.go
vrothberg: I changed it as suggested. Very thoughtful!

libpod/image/pull_test.go
vrothberg: Done and dropped a comment.

libpod/image/pull_test.go
vrothberg: Fixed, thanks a lot!

libpod/image/pull_test.go
vrothberg: Much better!

vrothberg:
Final polish based on @mtrmac's comments pushed.

go.mod
```diff
@@ -60,8 +60,10 @@ require (
...
 	golang.org/x/net v0.0.0-20200707034311-ab3426394381
 	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
-	golang.org/x/sys v0.0.0-20200728102440-3e129f6d46b1
+	golang.org/x/sys v0.0.0-20200810151505-1b9f1253b3ed
 	k8s.io/api v0.0.0-20190620084959-7cf5895f2711
 	k8s.io/apimachinery v0.19.0
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 )
+
+replace github.com/containers/image/v5 => github.com/containers/image/v5 v5.5.2-0.20200902171422-1c313b2d23e0
```
rhatdan, line 69: Isn't this a WIP until this is released?  I think we should release containers/image so that @mheon can release pdoman 2.1.0

vrothberg: @mheon acked above (see https://github.com/containers/podman/pull/6811#issuecomment-688523677). We just need to release before v2.1.0 which is still a bit in the future.

mheon: @rhatdan We're still waiting on a few more PRs in Podman land (I don't want to cut a 2.1.0 without https://github.com/containers/podman/pull/7452 because that's blocking some folks from moving off 1.9) so we have a bit of time.

mtrmac (commented):
LGTM.

rhatdan:
/lgtm
But this Podman can not be released until containers/image is updated.

maybe-sybr:
Is there any follow up work planned to bring the other formats into line with this? In particular I'm interested in `oci-dir`.

vrothberg:
> Is there any follow up work planned to bring the other formats into line with this? In particular I'm interested in `oci-dir`.

The `oci` transports already supports multiple image for storing images. However, there's no easy way to use it as a source without knowing the exact images in the directory: `Error initializing source oci:/home/vrothberg/foo:: more than one image in oci, choose an image`

@nalind @mtrmac I haven't checked the code but assume that the `oci` transport may need some wiring w.r.t. to image indexes to make it work?

mtrmac:
A similar list/get-reference API seems natural; the thing I’m not currently sure about is multi-arch support; that somehow uses  OCI indices, and I’m not quite sure how. IIRC it’s not a nested-index design, so the top-level index might be a multi-arch image set, not a multi-image archive. @nalind probably knows more.

nalind:
When we merged manifest list support, the `oci` transports stopped assuming that a manifest being added to them wasn't a list, so entries they contain can now themselves be either lists, or images that would be included in a list, and tag ("org.opencontainers.image.ref.name") annotations stored in the top-level index can be attached to either.  They no longer attempt to auto-generate `Platform` information for their contents, either.

Copying using `skopeo copy --all` from a source that is a manifest list to an `oci` location produces this:
```
- index.json (1 entry which may have an "org.opencontainers.image.ref.name" annotation)
- blobs/
  - sha256/
    - manifest list as a blob (only entry in index.json)
    - manifests referred to by the manifest list, as blobs
    - blobs referred to by those manifests
```

Copying from another location into the same `oci` location adds another entry to the top-level `index.json`, and either the single manifest or the manifest list and some number of the manifests it refers to, along with other blobs, to the `blobs/sha256` subdirectory, so you get either:
```
- index.json (2 entries which may have "org.opencontainers.image.ref.name" annotations)
- blobs/
  - sha256/
    - two manifest lists as blobs (first and second entries in index.json)
    - manifests referred to by either manifest list, as blobs
    - blobs referred to by those manifests
```
or
```
- index.json (2 entries which may have "org.opencontainers.image.ref.name" annotations)
- blobs/
  - sha256/
    - a manifest list as a blob (first entry in index.json)
    - manifests referred to by the manifest list, as blobs
    - a manifest as a blob (second entry in index.json)
    - blobs referred to by those manifests
```

When using an `oci` location as a source, if there's exactly one item (whether it's a list or not is irrelevant at this point) listed in the `oci` location's `index.json`, not specifying a tag when referring to the layout as a source will return that one item as a starting point, and if that's a manifest list, the library's copying logic either copies everything in the list or some subset, just as it would if it were reading from a registry and starting with the same item.

If there's more than one item in the `oci` location, then they really need to have "org.opencontainers.image.ref.name" annotations on them, because otherwise we don't know how to specify which of them to start copying from, list or not, since references for those transports currently don't have a form that incorporates digests, and we don't have a way to list their contents.

PR #6814, opened 2020-06-29
Fix saving in oci format

### LINKED ISSUE #6544: Podman save oci-dir produces non-OCI layout
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

/kind bug

**Description**

`podman save --format oci-dir -o existing_dir` does not produce a OCI layout of the blobs in the `existing_dir` directory. I am not sure if this is by design, in which case, it would be nice to name this option differently.

```
$ mkdir podman_debian_dir
$ podman save --format oci-dir -o podman_debian_dir debian:latest
```

The layout of the results look like this:
```
$ ls -al
total 116436
drwxr-xr-x 2 nisha nisha      4096 Jun  9 07:06 .
drwxr-xr-x 4 nisha nisha      4096 Jun  9 07:05 ..
-rw-r--r-- 1 nisha nisha       579 Jun  9 07:06 42f7b5ed30dd502a394c6377a72e2e07e198c145e3bb749c780262d84bfa990e
-rw-r--r-- 1 nisha nisha 119207936 Jun  9 07:06 8803ef42039dcbe936755e9baae4bb7b19cb0fb6a438eb3992950cd0afef8e4f
-rw-r--r-- 1 nisha nisha       350 Jun  9 07:06 manifest.json
-rw-r--r-- 1 nisha nisha        33 Jun  9 07:06 version
```

I would expect the layout to be the same as when you do:
```
$ podman save --format oci-archive -o debian.tar debian:latest
$ tar xvf debian.tar -C podman_debian_dir
```
Like:
```
$ ls -al
total 20
drwxr-xr-x 3 nisha nisha 4096 Jun  9 06:49 .
drwxr-xr-x 4 nisha nisha 4096 Jun  9 07:05 ..
drwxr-xr-x 3 nisha nisha 4096 Jun  9 06:49 blobs
-rw-r--r-- 1 nisha nisha  274 Jun  9 06:49 index.json
-rw-r--r-- 1 nisha nisha   31 Jun  9 06:49 oci-layout
```

```
$ podman version
Version:            1.9.3
RemoteAPI Version:  1
Go Version:         go1.10.1
```

```
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.10.1
  podmanVersion: 1.9.3
host:
  arch: amd64
  buildahVersion: 1.14.9
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.16, commit: '
  cpus: 2
  distribution:
    distribution: ubuntu
    version: "18.04"
  eventLogger: file
  hostname: ubuntu
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.3.0-53-generic
  memFree: 789159936
  memTotal: 4098334720
  ociRuntime:
    name: runc
    package: 'containerd.io: /usr/bin/runc'
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc10
      commit: dc9208a3303feef5b3839f4323d9beb36df0a9dd
      spec: 1.0.1-dev
  os: linux
  rootless: true
  slirp4netns:
    package: 'slirp4netns: /usr/bin/slirp4netns'
    version: |-
      slirp4netns version 0.4.3
      commit: unknown
  swapFree: 1349640192
  swapTotal: 2147479552
  uptime: 149h 1m 52.11s (Approximately 6.21 days)
registries:
  search:
  - docker.io
  - quay.io
store:
  configFile: /home/nisha/.config/containers/storage.conf
  containerStore:
    number: 0
    paused: 0
    running: 0
    stopped: 0
  graphDriverName: vfs
  graphOptions: {}
  graphRoot: /home/nisha/.local/share/containers/storage
  graphStatus: {}
  imageStore:
    number: 2
  runRoot: /run/user/1000
  volumePath: /home/nisha/.local/share/containers/storage/volumes
```

```
$ apt list podman
Listing... Done
podman/unknown,now 1.9.3~1 amd64 [installed]
```

VMware Fusion VM

QiWang19:
- fix saving&loading oci format. Close #6544
- support loading using image name without "localhost/" prefix when reading from ociarchive/dir saved from this semantics

### DIFF
--- a/libpod/runtime_img.go
+++ b/libpod/runtime_img.go
@@ -8,9 +8,15 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/containers/buildah/imagebuildah"
+	"github.com/containers/image/v5/directory"
 	"github.com/containers/image/v5/docker/reference"
+	ociarchive "github.com/containers/image/v5/oci/archive"
+	"github.com/containers/image/v5/oci/layout"
+	"github.com/containers/image/v5/types"
+
 	"github.com/containers/libpod/v2/libpod/define"
 	"github.com/containers/libpod/v2/libpod/image"
 	"github.com/containers/libpod/v2/pkg/util"
@@ -18,9 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
-	"github.com/containers/image/v5/directory"
 	dockerarchive "github.com/containers/image/v5/docker/archive"
-	ociarchive "github.com/containers/image/v5/oci/archive"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
@@ -256,28 +260,48 @@ func DownloadFromFile(reader *os.File) (string, error) {
 
 // LoadImage loads a container image into local storage
 func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer io.Writer, signaturePolicy string) (string, error) {
-	var newImages []*image.Image
-	src, err := dockerarchive.ParseReference(inputFile) // FIXME? We should add dockerarchive.NewReference()
-	if err == nil {
-		newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
-	}
-	if err != nil {
-		// generate full src name with specified image:tag
-		src, err := ociarchive.NewReference(inputFile, name) // imageName may be ""
-		if err == nil {
-			newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
-		}
-		if err != nil {
-			src, err := directory.NewReference(inputFile)
-			if err == nil {
-				newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
+	var (
+		newImages []*image.Image
+		err       error
+		src       types.ImageReference
+	)
+
+	for _, referenceFn := range []func() (types.ImageReference, error){
+		func() (types.ImageReference, error) {
+			return dockerarchive.ParseReference(inputFile) // FIXME? We should add dockerarchive.NewReference()
+		},
+		func() (types.ImageReference, error) {
+			return ociarchive.NewReference(inputFile, name) // name may be ""
+		},
+		func() (types.ImageReference, error) {
+			// prepend "localhost/" to support local image saved with this semantics
+			if !strings.Contains(name, "/") {
+				return ociarchive.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
 			}
-			if err != nil {
-				return "", errors.Wrapf(err, "error pulling %q", name)
+			return nil, nil
+		},
+		func() (types.ImageReference, error) {
+			return directory.NewReference(inputFile)
+		},
+		func() (types.ImageReference, error) {
+			return layout.NewReference(inputFile, name)
+		},
+		func() (types.ImageReference, error) {
+			// prepend "localhost/" to support local image saved with this semantics
+			if !strings.Contains(name, "/") {
+				return layout.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			}
+			return nil, nil
+		},
+	} {
+		src, err = referenceFn()
+		if err == nil && src != nil {
+			if newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer); err == nil {
+				return getImageNames(newImages), nil
 			}
 		}
 	}
-	return getImageNames(newImages), nil
+	return "", errors.Wrapf(err, "error pulling %q", name)
 }
 
 func getImageNames(images []*image.Image) string {
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -21,6 +21,7 @@ import (
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	ociarchive "github.com/containers/image/v5/oci/archive"
+	"github.com/containers/image/v5/oci/layout"
 	is "github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/tarball"
 	"github.com/containers/image/v5/transports"
@@ -1483,9 +1484,10 @@ func (i *Image) Save(ctx context.Context, source, format, output string, moreTag
 			return errors.Wrapf(err, "error getting OCI archive ImageReference for (%q, %q)", output, destImageName)
 		}
 	case "oci-dir":
-		destRef, err = directory.NewReference(output)
+		destImageName := imageNameForSaveDestination(i, source)
+		destRef, err = layout.NewReference(output, destImageName) // destImageName may be ""
 		if err != nil {
-			return errors.Wrapf(err, "error getting directory ImageReference for %q", output)
+			return errors.Wrapf(err, "error getting the OCI directory ImageReference for (%q, %q)", output, destImageName)
 		}
 		manifestType = imgspecv1.MediaTypeImageManifest
 	case "docker-dir":


QiWang19:
replace #6562

libpod/runtime_img.go
```diff
@@ -273,7 +274,13 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
 				newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
 			}
 			if err != nil {
-				return "", errors.Wrapf(err, "error pulling %q", name)
+				src, err := layout.NewReference(inputFile, "")
```
QiWang19, line 277: @mtrmac PTAL
 `src, err := layout.NewReference(inputFile, "")`, why the`podman load` can work with "", but error if pass to `name`?
` unable to pull oci:input_file:alpine: Error initializing source oci:input_file:alpine: no descriptor found for reference "alpine"`

mtrmac: @QiWang19 `""` means “verify that the index contains only one manifest, and use that one”. Passing a name allows reading from an index that contains any number of manifests, but the name must match the `org.opencontainers.image.ref.name` annotation exactly (that annotation is a string with allowed characters and no defined normalization semantics).

So, what’s the value of that annotation in the created index? Is it there at all? Does it perhaps say `localhost/alpine` or something like that?

QiWang19: hm,yes, if I load from the oci-dir it was saved, it says `localhost/alpine`
```
$ podman save --format oci-dir -o podman-dir alpine
$ podman load -i podman-dir alpine

$ podman images
REPOSITORY            TAG     IMAGE ID      CREATED      SIZE
localhost/podman-dir  latest  0f5f445df8cc  4 weeks ago  5.85 MB
localhost/alpine      latest  0f5f445df8cc  4 weeks ago  5.85 MB
```

mtrmac: That’s messy… the local image _is_ named `localhost/alpine` in c/storage, and we do want to find that image when the user says `alpine`.  So, using the `localhost/alpine` name when writing to the archive makes _some_ sense.

OTOH when reading from the archive it’s not at all clean to automatically change the name, because the names in OCI don’t have any semantics like that.

Conceptually I think it would be clean to use just `alpine` when writing to `oci-dir:`, and because it is basically new with this PR, we could do that — but it is more important to me that `oci-dir:` and `oci-archive:` behave consistently, and we can’t quite change the behavior of `oci-archive:` now.

I suppose we could, when reading from the archive, try both the user input unmodified and user input with `localhost/` prepended if there is no registry component.

QiWang19: > Passing a name allows reading from an index that contains any number of manifests, but the name must match the `org.opencontainers.image.ref.name` annotation exactly (that annotation is a string with allowed characters and no defined normalization semantics).
> 

The passing name error  happens if `org.opencontainers.image.ref.name` contains a tag, it this expected?

```
podman save --format oci-dir -o podman-dir docker.io/library/alpine:3.5
podman load -i podman-dir docker.io/library/alpine:3.5
Error initializing source oci:podman-dir:docker.io/library/alpine: no descriptor found for reference "docker.io/library/alpine"
```

mtrmac: That doesn’t look expected — but it also suggests that the tag was stripped somewhere in Podman.

libpod/runtime_img.go
```diff
@@ -268,12 +271,27 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
 			newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
 		}
 		if err != nil {
-			src, err := directory.NewReference(inputFile)
-			if err == nil {
-				newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
+			if strings.Contains(name, "/") {
```
mtrmac, line 274: `!Contains` I imagine

libpod/runtime_img.go
```diff
@@ -268,12 +271,27 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
+				// prepend "localhost/" to support local image saved with this semantics
+				src, err = ociarchive.NewReference(inputFile, filepath.Join("localhost", name))
```
mtrmac, line 276: Not `filepath.Join`, that is semantically incorrect and it would use backslashes on Windows. Even using `"localhost/" + ` would be better.

But there already is some code around for dealing with this, at least this should use `DefaultLocalRegistry`.

libpod/runtime_img.go
```diff
@@ -268,12 +271,27 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
+				}
 			}
 			if err != nil {
-				return "", errors.Wrapf(err, "error pulling %q", name)
+				src, err := directory.NewReference(inputFile)
+				if err == nil {
+					newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
+				}
+				if err != nil {
+					src, err := layout.NewReference(inputFile, name)
```
mtrmac, line 287: OCI expanded directories and OCI archives should behave the same WRT the prefixing with `localhost/`.

libpod/runtime_img.go
```diff
@@ -268,12 +271,27 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
+					if err == nil {
+						newImages, err = r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
+					}
+					if err != nil {
+						return "", errors.Wrapf(err, "error pulling %q", name)
+					}
+				}
```
mtrmac, line 294: BTW this series of ever-deeper-nesting `if`s is very repetitive and fairly hard to read, especially if there were any copy&paste errors it would be difficult to notice.

Please consider consolidating this, either by building a `[]types.ImageReference` in advance (for all possible cases, skipping ones that fail) and iterating over it, or maybe
```go
for _, referenceFn := []range func() (types.ImageReference, error) {
    func() (
        types.ImageReference, error) { return dockerarchive.ParseReference(inputFile) // FIXME? We should add dockerarchive.NewReference()
    },
    …
} {
   ref, err := referenceFn()
   if err == nil {
      if newImages, err = …; err == nil {
          return …
       }
   }
}

**Comment by QiWang19** [comment]
@mtrmac PTAL

File: libpod/runtime_img.go [hunk 6]
```diff
@@ -256,28 +260,54 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+		err       error
+		src       types.ImageReference
+	)
+
+	for _, referenceFn := range []func() (types.ImageReference, error){
+		func() (types.ImageReference, error) {
+			return dockerarchive.ParseReference(inputFile) // FIXME? We should add dockerarchive.NewReference()
+		},
+		func() (types.ImageReference, error) {
+			if src, err = ociarchive.NewReference(inputFile, name); err == nil { // imageName may be ""
```
mtrmac, line 274: This can be just `return ociarchive.NewReference…`. Same in the other copy.

libpod/runtime_img.go
```diff
@@ -256,28 +260,54 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
-			if err != nil {
-				return "", errors.Wrapf(err, "error pulling %q", name)
+			return src, err
+		},
+		func() (types.ImageReference, error) {
+			// prepend "localhost/" to support local image saved with this semantics
+			if !strings.Contains(name, "/") {
+				return ociarchive.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			}
+			return src, err
```
mtrmac, line 284: These `src, err` values come from some earlier iteration… It might be cleaner to return `nil, nil` and check for `src == nil` in the loop.

Same in the other copy.

libpod/runtime_img.go
mtrmac, line 274:
```suggestion
			if src, err = ociarchive.NewReference(inputFile, name); err == nil { // name may be ""
```

mtrmac: (Also, this `src, err = ` assignment modifies the variables in the loop. It does work fine with the code as it is now, but it’s hard to maintain with an interaction at a distance like that.)

QiWang19:
@rhatdan @mheon @vrothberg PTAL

PR #6821, opened 2020-06-30
Set TMPDIR to /var/tmp by default if not set

rhatdan:
Containers/image will use TMPDIR for the location of pulled layer blobs.
If TMPDIR is not set, it will use /tmp.  Since this is known to be of
limited space on most systems, we change the default to /var/tmp
if the user has not told the tools where to store temporary files.

### DIFF
--- a/cmd/podman/main.go
+++ b/cmd/podman/main.go
@@ -29,6 +29,11 @@ func main() {
 		return
 	}
 
+	// Hard code TMPDIR functions to use /var/tmp, if user did not override
+	if _, ok := os.LookupEnv("TMPDIR"); !ok {
+		os.Setenv("TMPDIR", "/var/tmp")
+	}
+
 	cfg := registry.PodmanConfig()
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1769918

TomSweeneyRedHat:
Changes LGTM, but tests are spitting up.
You did this same change in https://github.com/containers/libpod/commit/772af18205eacf95ae55a4167f7f48e917bcd5e1, did it get removed mistakenly?  I'm confused about that part.

mheon:
LGTM, but I doubt this is actually the fix to that BZ. The version that failed QE had this code, and even without it we pass if TMPDIR was explicitly overridden. If there's something going on, it's happening elsewhere.

rhatdan:
@TomSweeneyRedHat It was not included in the rewrite.

PR #6827, opened 2020-06-30
Disable SELinux labeling if privileged and user does not specify labels

rhatdan:
The previous patch mistakenly turned on SELinux even when --privileged.

This patch will disable SELinux, if the user specified --privileged and
did not specify any SELinux options.

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -196,9 +196,7 @@ func CreateContainer(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 }
 
 func parseSecurityOpt(config *cc.CreateConfig, securityOpts []string, runtime *libpod.Runtime) error {
-	var (
-		labelOpts []string
-	)
+	var labelOpts []string
 
 	if config.PidMode.IsHost() {
 		labelOpts = append(labelOpts, label.DisableSecOpt()...)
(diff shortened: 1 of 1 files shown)


rhatdan:
@mheon PTAL

mheon:
LGTM

mheon:
Force-merging since CI is down

cjeanner:
Hello,
I think this kind of revert to the previous situation where podman doesn't honour the "--security-opt" when "--privileged" is passed. Doing a test right now, and, well.... we're back to the previous state: "--privileged --security-opt label=filetype:container_share_t" doesn't apply the filetype (still container_file_t in the container).

cjeanner:
(scratch my comment - there's another issue, --pid=host is preventing the application of --security-opt)

rhatdan:
@cjeanner Please be more specific.  --pid=host will cause the SELinux labels to not be applied, is that the problem you are seeing?  What command exacly are you having problems with?

cjeanner:
@rhatdan hello! so everything is explained here, with a reproducer: https://bugzilla.redhat.com/show_bug.cgi?id=1846364#c23
Lemme know if you need more infos.

rhatdan:
https://github.com/containers/libpod/pull/6832

PR #6828, opened 2020-06-30
Fix event cleanup when client closes connection

mheon:
Supercedes #6680

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Podman presently only supports running containers on Linux. However, we are buil
 
 ## Roadmap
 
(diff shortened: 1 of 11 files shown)


mheon:
@edsantiago Any chance you remember if these are the same errors #6680 saw?

edsantiago:
Yep, that's exactly it. Now where did I write down my reproducer? It's a trivial two-liner.

mheon:
@edsantiago No need, I don't think I'm brave enough to try and salvage this. I'll go with a different approach.

edsantiago:
@mheon check your email, date 24 June, subject 'PR 6680 reproducer in 1minutetip'. I wrote down a lot of notes there.

PR #6829, opened 2020-06-30
Add username to /etc/passwd inside of container if --userns keep-id

rhatdan:
If I enter a continer with --userns keep-id, my UID will be present
inside of the container, but most likely my user will not be defined.

This patch will take information about the user and stick it into the
container.

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -278,6 +278,9 @@ type ContainerConfig struct {
 	User string `json:"user,omitempty"`
 	// Additional groups to add
 	Groups []string `json:"groups,omitempty"`
+	// AddCurrentUserPasswdEntry indicates that the current user passwd entry
+	// should be added to the /etc/passwd within the container
+	AddCurrentUserPasswdEntry bool `json:"addCurrentUserPasswdEntry,omitempty"`
 
 	// Namespace Config
 	// IDs of container to share namespaces with
@@ -786,7 +789,10 @@ func (c *Container) Hostname() string {
 
 // WorkingDir returns the containers working dir
 func (c *Container) WorkingDir() string {
-	return c.config.Spec.Process.Cwd
+	if c.config.Spec.Process != nil {
+		return c.config.Spec.Process.Cwd
+	}
+	return "/"
 }
 
 // State Accessors
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"os/user"
 	"path"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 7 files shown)


libpod/container.go
```diff
@@ -284,13 +284,14 @@ type ContainerConfig struct {
...
-	PIDNsCtr    string `json:"pidNsCtr,omitempty"`
-	UserNsCtr   string `json:"userNsCtr,omitempty"`
-	UTSNsCtr    string `json:"utsNsCtr,omitempty"`
-	CgroupNsCtr string `json:"cgroupNsCtr,omitempty"`
+	IPCNsCtr     string `json:"ipcNsCtr,omitempty"`
+	MountNsCtr   string `json:"mountNsCtr,omitempty"`
+	NetNsCtr     string `json:"netNsCtr,omitempty"`
+	PIDNsCtr     string `json:"pidNsCtr,omitempty"`
+	UserNsCtr    string `json:"userNsCtr,omitempty"`
+	UserNsKeepId bool
```
mheon, line 292: This should be elsewhere, have a JSON tag, and a comment describing what it does

libpod/options.go
```diff
@@ -844,6 +844,19 @@ func WithPIDNSFrom(nsCtr *Container) CtrCreateOption {
 	}
 }
 
+// WithUserNSKeepID indicates that container should add user entry to passwd
+// file, since the UID will be mapped into the container, via user namespace
+func WithUserNSKeepId() CtrCreateOption {
```
mheon, line 849: Can we rename this? Maybe "WithAddCurrentUserToPasswd" or something more clear about what it does?

libpod/container_internal_linux.go
```diff
@@ -1430,46 +1431,55 @@ func (c *Container) generatePasswd() (string, error) {
...
-		return "", nil
-	}
-	// Lookup the user to see if it exists in the container image
-	_, err = lookup.GetUser(c.state.Mountpoint, userspec)
-	if err != nil && err != user.ErrNoPasswdEntries {
-		return "", err
-	}
-	if err == nil {
-		return "", nil
+	originPasswdFile := filepath.Join(c.state.Mountpoint, "/etc/passwd")
```
vrothberg, line 1437: I think it's time to either break this code into a separate function or move it into a `pkg/passwd`.

It was already really hard to brain-parse the code before but now we have four nested branches that let my brain parser overflow.

Breaking the code into a separate function would simplify it (as we can use returns), we could add unit tests and ideally add more comments.

libpod/container.go
```diff
@@ -278,6 +278,9 @@ type ContainerConfig struct {
 	User string `json:"user,omitempty"`
 	// Additional groups to add
 	Groups []string `json:"groups,omitempty"`
+	// AddCurrentUserPasswdEntry indicates that the current user passwd entry
+	// should be added to the /etc/passwd within the container
+	AddCurrentUserPasswdEntry bool
```
mheon, line 283: Can this have an `omitempty` so we minimize the amount of stuff saved to the DB?

libpod/container_internal_linux.go
rhatdan: Done

vrothberg:
Looks like a bug in the tests. I'll check out the PR and have a look.

debarshiray:
Shouldn't it also have created the group? :)

PR #6832, opened 2020-07-01
Don't disable selinux labels if user specifies a security opt

rhatdan:
Currenty if the user specifies --pid=host or --ipc=host or --privileged
then we disable SELinux labeling. If the user however specifies
--security-opt label:...  Then we assume they want to leave SELinux enabled
and know what they are doing.

This PR will leave SELinux enabled if a user specifies a --security-opt label
option.

### DIFF
--- a/cmd/podman/shared/create.go
+++ b/cmd/podman/shared/create.go
@@ -198,9 +198,7 @@ func CreateContainer(ctx context.Context, c *GenericCLIResults, runtime *libpod.
 func parseSecurityOpt(config *cc.CreateConfig, securityOpts []string, runtime *libpod.Runtime) error {
 	var labelOpts []string
 
-	if config.PidMode.IsHost() {
-		labelOpts = append(labelOpts, label.DisableSecOpt()...)
-	} else if config.PidMode.IsContainer() {
+	if config.PidMode.IsContainer() {
 		ctr, err := runtime.LookupContainer(config.PidMode.Container())
 		if err != nil {
 			return errors.Wrapf(err, "container %q not found", config.PidMode.Container())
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

TomSweeneyRedHat:
LGTM

cjeanner:
Hello!
Thank you for this patch - I'm testing it right now, and see it will force a new change on our side:
IF we pass ```--privileged --pid=host```   THEN it disable selinux as expected.
IF we pass ```--privileged --pid=host --security-opt label=....```  THEN it will NOT disable selinux and just apply the labels.

This means we'll need to ensure we actually do pass ```--security-opt label=type:spc_t``` when we pass the "--pid=host". Not a big deal, but it's something we have to do, right?

For instance:
```
podman run --rm --privileged --net=host --pid=host docker.io/library/centos:8 cat /proc/self/attr/current
unconfined_u:system_r:spc_t:s0

podman run --rm --privileged --net=host --pid=host --security-opt label=level:s0 docker.io/library/centos:8 cat /proc/self/attr/current
system_u:system_r:container_t:s0

podman run --rm --privileged --net=host --pid=host --security-opt label=level:s0 --security-opt label=type:spc_t docker.io/library/centos:8 cat /proc/self/attr/current
system_u:system_r:spc_t:s0
```
This would ensure the process is actually running as privileged.

Do you confirm this is the intended behaviour? If so, I think it would be fine, as long as we do the edition job on our side.

Cheers,

C.

rhatdan:
Let's put it this way.  If you set the --security-opt label=* flag, then SELinux will be enabled, bottom line.
Currently 
--privileged, --ipc=host and --pid=host disable SELinux separation.  The reason we do this, is most applications will be blocked from access to the different types exposed by the --ipc=host and --pid=host flags.  If the user specifies --security-opt label=* then we assume that they understand how SELinux works and will do the right thing with the flags.

Bottom line is yes if you set the --secuirty-opt label=* and want SELinux to not interface with container separation, then  you MUST set the spc_t type --security-opt label=type:spc_t

cjeanner:
Thank you Dan for the details! After the first test, I can confirm this last patch seems to solve our issues, and, nice addition, we apparently don't need to do more changes on our side.
Thank you again!

mheon:
I'm going to go ahead and merge - we can land further changes in additional PRs if we need to

PR #6833, opened 2020-07-01
Do not disable labelling on pid=host if overridden

mheon:
If the user manually provides SELinux security opts, prefer those to unconditionally disabling when `--pid=host` and `--ipc=host` are passed.


mheon:
Closing in favor of #6832

PR #6836, opened 2020-07-01
Add --tz flag to create, run

### LINKED ISSUE #5128: Add flag to podman run/create to automatically set the timezone in container to match host.
Something like?
--tz host
--tz image

Then in containers conf, we could allow setting this by default.

Thoughts?

ashley-cui:
--tz flag sets timezone inside container
Can be set to IANA timezone as well as `local` to match host machine

Closes #5128

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -449,6 +449,11 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"tty", "t", false,
 		"Allocate a pseudo-TTY for container",
 	)
+	createFlags.StringVar(
+		&cf.Timezone,
+		"tz", containerConfig.TZ(),
+		"Set timezone in container",
+	)
 	createFlags.StringSliceVar(
 		&cf.UIDMap,
 		"uidmap", []string{},
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -815,6 +815,10 @@ interactive shell. The default is false.
 Note: The **-t** option is incompatible with a redirection of the Podman client
 standard input.
 
+**--tz**=*timezone*
+
+Set timezone in container. This flag takes area-based timezones, GMT time, as well as `local`, which sets the timezone in the container to match the host machine. See `/usr/share/zoneinfo/` for valid timezones.
+
 **--uidmap**=*container_uid:host_uid:amount*
 
 UID map for the user namespace.  Using this flag will run the container with user namespace enabled.  It conflicts with the `--userns` and `--subuidname` flags.
@@ -1034,6 +1038,14 @@ the uids and gids from the host.
 
 ```
 $ podman create --uidmap 0:30000:7000 --gidmap 0:30000:7000 fedora echo hello
+```
+
+### Configure timezone in a container
+
+```
+$ podman create --tz=local alpine date
+$ podman create --tz=Asia/Shanghai alpine date
+$ podman create --tz=US/Eastern alpine date
 ```
 
 ### Rootless Containers
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -856,6 +856,10 @@ interactive shell. The default is **false**.
 **NOTE**: The **-t** option is incompatible with a redirection of the Podman client
 standard input.
 
+**--tz**=*timezone*
+
+Set timezone in container. This flag takes area-based timezones, GMT time, as well as `local`, which sets the timezone in the container to match the host machine. See `/usr/share/zoneinfo/` for valid timezones.
+
 **--uidmap**=*container_uid*:*host_uid*:*amount*
 
 Run the container in a new user namespace using the supplied mapping. This option conflicts
@@ -1317,6 +1321,14 @@ using global options.
 
 ```
 podman --log-level=debug --storage-driver overlay --storage-opt "overlay.mount_program=/usr/bin/fuse-overlayfs" run busybox /bin/sh
+```
+
+### Configure timezone in a container
+
+```
+$ podman run --tz=local alpine date
+$ podman run --tz=Asia/Shanghai alpine date
+$ podman run --tz=US/Eastern alpine date
 ```
 
 ### Rootless Containers
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -424,6 +424,10 @@ type ContainerConfig struct {
 	// to 0, 1, 2) that will be passed to the executed process. The total FDs
 	// passed will be 3 + PreserveFDs.
 	PreserveFDs uint `json:"preserveFds,omitempty"`
+
+	// Timezone is the timezone inside the container.
+	// Local means it has the same timezone as the host machine
+	Timezone string `json:"timezone,omitempty"`
 }
 
 // ContainerNamedVolume is a named volume that will be mounted into the
@@ -1247,4 +1251,9 @@ func (c *Container) AutoRemove() bool {
 		return false
 	}
 	return c.Spec().Annotations[define.InspectAnnotationAutoremove] == define.InspectResponseTrue
+}
+
+func (c *Container) Timezone() string {
+	return c.config.Timezone
+
 }
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1241,6 +1241,31 @@ func (c *Container) makeBindMounts() error {
 		c.state.BindMounts["/etc/hostname"] = hostnamePath
 	}
 
+	// Make /etc/localtime
+	if c.Timezone() != "" {
+		if _, ok := c.state.BindMounts["/etc/localtime"]; !ok {
+			var zonePath string
+			if c.Timezone() == "local" {
+				zonePath, err = filepath.EvalSymlinks("/etc/localtime")
+				if err != nil {
+					return errors.Wrapf(err, "error finding local timezone for container %s", c.ID())
+				}
+			} else {
+				zone := filepath.Join("/usr/share/zoneinfo", c.Timezone())
+				zonePath, err = filepath.EvalSymlinks(zone)
+				if err != nil {
+					return errors.Wrapf(err, "error setting timezone for container %s", c.ID())
+				}
+			}
+			localtimePath, err := c.copyTimezoneFile(zonePath)
+			if err != nil {
+				return errors.Wrapf(err, "error setting timezone for container %s", c.ID())
+			}
+			c.state.BindMounts["/etc/localtime"] = localtimePath
+
+		}
+	}
+
 	// Make .containerenv
 	// Empty file, so no need to recreate if it exists
 	if _, ok := c.state.BindMounts["/run/.containerenv"]; !ok {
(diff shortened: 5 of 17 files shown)


ashley-cui:
@rhatdan PTAL

docs/source/markdown/podman-run.1.md
```diff
@@ -856,6 +856,14 @@ interactive shell. The default is **false**.
 **NOTE**: The **-t** option is incompatible with a redirection of the Podman client
 standard input.
 
+**--tz**=*timezone*
+
+Set timezone in container. This flag takes area-based timezones, GMT time, as well as `host`, which sets the timezone in the container to match the host machine. See `/usr/share/zoneinfo/` for valid timezones.
+
+$ podman create --tz=host
+$ podman create --tz=Asia/Shanghai
+$ podman create --tz=US/Eastern
```
TomSweeneyRedHat, line 865: s/b podman run

docs/source/markdown/podman-create.1.md
```diff
@@ -815,6 +815,15 @@ interactive shell. The default is false.
 Note: The **-t** option is incompatible with a redirection of the Podman client
 standard input.
 
+**--tz**=*timezone*
+
+Set timezone in container. This flag takes area-based timezones, GMT time, as well as `host`, which sets the timezone in the container to match the host machine. See `/usr/share/zoneinfo/` for valid timezones.
+
+$ podman create --tz=host
```
rhatdan, line 822: Isn't this local?

docs/source/markdown/podman-run.1.md
```diff
@@ -856,6 +856,14 @@ interactive shell. The default is **false**.
 **NOTE**: The **-t** option is incompatible with a redirection of the Podman client
 standard input.
 
+**--tz**=*timezone*
+
+Set timezone in container. This flag takes area-based timezones, GMT time, as well as `host`, which sets the timezone in the container to match the host machine. See `/usr/share/zoneinfo/` for valid timezones.
+
+$ podman create --tz=host
```
rhatdan: tz=local

cmd/podman/common/create.go
```diff
@@ -449,6 +449,11 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"tty", "t", false,
 		"Allocate a pseudo-TTY for container",
 	)
+	createFlags.StringVar(
+		&cf.TZ,
+		"tz", containerConfig.TZ(),
```
rhatdan, line 454: What does this do in podman-remote?

docs/source/markdown/podman-run.1.md
```diff
@@ -856,6 +856,14 @@ interactive shell. The default is **false**.
 **NOTE**: The **-t** option is incompatible with a redirection of the Podman client
 standard input.
 
+**--tz**=*timezone*
+
+Set timezone in container. This flag takes area-based timezones, GMT time, as well as `host`, which sets the timezone in the container to match the host machine. See `/usr/share/zoneinfo/` for valid timezones.
+
+$ podman run --tz=host
```
rhatdan, line 863: tz=local

libpod/container.go
```diff
@@ -424,6 +424,9 @@ type ContainerConfig struct {
 	// to 0, 1, 2) that will be passed to the executed process. The total FDs
 	// passed will be 3 + PreserveFDs.
 	PreserveFDs uint `json:"preserveFds,omitempty"`
+
+	//Tz is the timezone inside the container
+	TZ string `json:"timezone"`
```
rhatdan, line 429: omitempty

libpod/container_internal_linux.go
```diff
@@ -1241,6 +1242,26 @@ func (c *Container) makeBindMounts() error {
...
+				return errors.Wrapf(err, "error finding local timezone for container %s", c.ID())
+			}
+		} else {
+			zonePath = filepath.Join("/usr/share/zoneinfo", c.TZ())
+		}
+		localtimePath, err := c.copyTimezoneFile(zonePath)
+		if err == nil {
+			c.state.BindMounts["/etc/localtime"] = localtimePath
+		} else {
+			logrus.Warning("Official timezone not set, using custom")
```
rhatdan, line 1260: This is an error not a warning, return the error.
if err != nil {
 			return errors.Wriap(err, "Official timezone not set, using custom")
}

test/e2e/run_test.go
```diff
@@ -1047,4 +1047,25 @@ USER mail`
...
+		Expect(session.OutputToString()).To(ContainSubstring("HST"))
+
+		session = podmanTest.Podman([]string{"run", "--tz", "local", "--rm", ALPINE, "date", "+'%H %Z'"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		t := time.Now()
+		z, _ := t.Zone()
+		h := strconv.Itoa(t.Hour())
+		Expect(session.OutputToString()).To(ContainSubstring(z))
+		Expect(session.OutputToString()).To(ContainSubstring(h))
```
rhatdan, line 1068: Could you add a test with container.conf set to something and make sure it works.

test/e2e/create_test.go
```diff
@@ -471,4 +471,27 @@ var _ = Describe("Podman create", func() {
 		Expect(len(data)).To(Equal(1))
 		Expect(data[0].Config.StopSignal).To(Equal(uint(15)))
 	})
+
+	It("podman create --tz", func() {
+		session := podmanTest.Podman([]string{"create", "--tz", "foo", "--name", "bad", ALPINE, "date"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Not(Equal(0)))
+
+		session = podmanTest.Podman([]string{"run", "--tz", "Pacific/Honolulu", "--name", "zone", ALPINE, "date"})
```
rhatdan, line 480: You should not be using run, you should use create followed by start.

cmd/podman/common/create.go
ashley-cui: ah, I haven't done anything in the remote area, should I add the remote implementation to this PR as well?

test/e2e/run_test.go
ashley-cui: @rhatdan Do you know which test files has examples of a test with container.conf? I'm trying to find examples

docs/source/markdown/podman-create.1.md
mheon, line 822: Examples should go at the bottom of the manpage

libpod/container.go
```diff
@@ -1248,3 +1251,8 @@ func (c *Container) AutoRemove() bool {
 	}
 	return c.Spec().Annotations[define.InspectAnnotationAutoremove] == define.InspectResponseTrue
 }
+
+func (c *Container) TZ() string {
```
mheon, line 1255: Needs a comment

libpod/container.go
mheon: Please rename to "Timezone" as part of this struct - brevity is good for CLI options, not good for code clarity.

libpod/container_internal_linux.go
```diff
@@ -1241,6 +1242,26 @@ func (c *Container) makeBindMounts() error {
 		c.state.BindMounts["/etc/hostname"] = hostnamePath
 	}
 
+	// Make /etc/localtime
+	if _, ok := c.state.BindMounts["/etc/localtime"]; !ok {
```
mheon, line 1246: Only do this is timezone is not set to "" - we should not modify behavior for containers not explicitly passing the flag.

libpod/container_internal_linux.go
```diff
@@ -1533,3 +1554,32 @@ func (c *Container) getOCICgroupPath() (string, error) {
...
+	}
+	defer dest.Close()
+	_, err = io.Copy(dest, src)
+	if err != nil {
+		return "", err
+	}
+
+	//Copy SELinux location file permissions
+	copylabel, err := selinux.FileLabel(zonePath)
+	if err := label.Relabel(localtimeCopy, copylabel, false); err != nil {
```
mheon, line 1580: Please use `c.config.MountLabel` instead for the label

mheon: Also, ensure this is chown'd to `c.RootUID()` and `c.RootGID()`

libpod/options.go
```diff
@@ -1525,6 +1525,16 @@ func withSetAnon() VolumeCreateOption {
 	}
 }
 
+func WithTZ(path string) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+		ctr.config.TZ = path
```
mheon, line 1533: We should validate here that the given timezone actually exists.

pkg/specgen/generate/container_create.go
```diff
@@ -135,6 +135,17 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
 		options = append(options, libpod.WithStdin())
 	}
 
+	if s.TZ != "" {
+		if s.TZ != "local" {
+			zonePath := filepath.Join("/usr/share/zoneinfo", s.TZ)
+			_, err := os.Stat(zonePath)
+			if err != nil {
+				return nil, err
```
mheon, line 143: More descriptive error here would be very nice

pkg/specgen/specgen.go
```diff
@@ -135,6 +135,8 @@ type ContainerBasicConfig struct {
 	// passed will be 3 + PreserveFDs.
 	// set tags as `json:"-"` for not supported remote
 	PreserveFDs uint `json:"-"`
+
+	TZ string `json:"timezone,omitempty"`
```
mheon, line 139: Needs a comment

test/e2e/run_test.go
```diff
@@ -1047,4 +1047,25 @@ USER mail`
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(strings.Contains(session.OutputToString(), groupName)).To(BeTrue())
 	})
+
+	It("podman run --tz", func() {
+		session := podmanTest.Podman([]string{"run", "--tz", "foo", "--rm", ALPINE, "date"})
```
mheon, line 1052: Can we verify this inside the container as well, make sure the timezone has been correctly detected?

ashley-cui: The honolulu test should make sure the timezone has been correctly detected since it checks if the timezone is 
HST (hawaii standard time) while the input is Pacific/Honolulu. I'm not sure how to do verify the timezone inside the container though.

mheon:
I see that a lot of the timezones are symlinks in non-Red Hat OSes; we might need to do some sort of resolution on the link to get the correct target. For example: In Debian, the `GB` timezone is a symlink (`GB -> Europe/London`). We should also have some sort of detection to ensure we don't accidentally mount in an entire folder if the user specifies `Australia` as a timezone, instead of `Australia/Brisbane`

ashley-cui:
@mheon Ooh, thanks, great point! I'll add that in now, a simple IsDir and EvalSymlinks should suffice?

mheon:
Yeah, those should be fine

ashley-cui:
@mheon @rhatdan fixed it up and added a few more tests, PTAL

rhatdan:
LGTM

test/e2e/containers_conf_test.go
```diff
@@ -211,4 +211,13 @@ var _ = Describe("Podman run", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.LineInOuputStartsWith("search")).To(BeFalse())
 	})
+
+	It("podman run containers.conf timezone", func() {
```
baude, line 215: what happens if i run this command with bad input? like abc?

baude: nvm, i see in other tests you check that!

cmd/podman/common/create.go
baude: if you don't do it here, i would exect you to wire up the tz stuff asap for the api (remote) side.

ashley-cui: I can definitely do that, but it might take a day or so to dig around and understand the api side. Would you prefer it in a new pr or add to this one?

ashley-cui: @baude uhh it looks like somehow i actually got this wired to the api remote side without knowing i did. pleasant surprise(?)

PR #6842, opened 2020-07-02
Pids-limit should only be set if the user set it

rhatdan:
Currently we are sending over pids-limits from the user even if they
never modified the defaults.  The pids limit should be set at the server
side unless modified by the user.

This issue has led to failures on systems that were running with cgroups V1.

### DIFF
--- a/pkg/specgen/generate/container.go
+++ b/pkg/specgen/generate/container.go
@@ -10,6 +10,7 @@ import (
 	envLib "github.com/containers/libpod/v2/pkg/env"
 	"github.com/containers/libpod/v2/pkg/signal"
 	"github.com/containers/libpod/v2/pkg/specgen"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
@@ -166,6 +167,21 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 	if len(s.SelinuxOpts) == 0 {
 		if err := setLabelOpts(s, r, s.PidNS, s.IpcNS); err != nil {
 			return nil, err
+		}
+	}
+
+	// If caller did not specify Pids Limits load default
+	if s.ResourceLimits == nil || s.ResourceLimits.Pids == nil {
+		if s.CgroupsMode != "disabled" {
+			limit := rtc.PidsLimit()
+			if limit != 0 {
+				if s.ResourceLimits == nil {
+					s.ResourceLimits = &spec.LinuxResources{}
+				}
+				s.ResourceLimits.Pids = &spec.LinuxPids{
+					Limit: limit,
+				}
+			}
 		}
 	}
 
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -330,8 +330,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"pid", "",
 		"PID namespace to use",
 	)
-	createFlags.Int64Var(
-		&cf.PIDsLimit,
+	createFlags.Int64(
 		"pids-limit", containerConfig.PidsLimit(),
 		"Tune container pids limit (set 0 for unlimited, -1 for server defaults)",
 	)
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -66,7 +66,7 @@ type ContainerCLIOpts struct {
 	OverrideArch      string
 	OverrideOS        string
 	PID               string
-	PIDsLimit         int64
+	PIDsLimit         *int64
 	Pod               string
 	PodIDFile         string
 	PreserveFDs       uint
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -7,14 +7,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containers/common/pkg/config"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/libpod/v2/cmd/podman/parse"
 	"github.com/containers/libpod/v2/libpod/define"
 	ann "github.com/containers/libpod/v2/pkg/annotations"
 	envLib "github.com/containers/libpod/v2/pkg/env"
 	ns "github.com/containers/libpod/v2/pkg/namespaces"
-	"github.com/containers/libpod/v2/pkg/rootless"
 	"github.com/containers/libpod/v2/pkg/specgen"
 	systemdGen "github.com/containers/libpod/v2/pkg/systemd/generate"
 	"github.com/containers/libpod/v2/pkg/util"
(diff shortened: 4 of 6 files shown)


rhatdan:
@giuseppe @mheon PTAL

mheon:
I don't like this because it disables the default PID limit entirely, even for root containers. We should preserve the default limit where possible (root and cgroups v2 rootless)

rhatdan:
@mheon The default limit should be handled on the server side, if the user does not specify a root limit then pkg/specgen should use the default.

mheon:
I don't see code for that anywhere. `pkg/specgen` takes the resource limits as they were given and uses them unmodified.

rhatdan:
@mheon Added  the defaults.

pkg/specgen/generate/container.go
```diff
@@ -169,6 +170,19 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 		}
 	}
 
+	// If caller did not specify Pids Limits load default
+	if s.ResourceLimits.Pids == nil {
+		if s.CgroupsMode != "disabled" {
```
mheon, line 175: We also need to make sure we are not rootless + cgroups v1

pkg/specgen/generate/container.go
```diff
@@ -169,6 +170,19 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 		}
 	}
 
+	// If caller did not specify Pids Limits load default
+	if s.ResourceLimits.Pids == nil {
```
mheon, line 174: This could null-pointer, I think - `s.ResourceLimits` may be null.

pkg/specgen/generate/container.go
rhatdan: The rtc.PidsLimit() takes care of this.

pkg/specgen/generate/container.go
rhatdan: Fixed

pkg/specgen/generate/container.go
```diff
@@ -169,6 +170,19 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 		}
 	}
 
+	// If caller did not specify Pids Limits load default
+	if s.ResourceLimits != nil && s.ResourceLimits.Pids == nil {
```
mheon, line 174: Do we want to make ResourceLimits if it was null and Pids is not empty?

mheon: (I think this will be the case for rootless v2, for example)

goochjj:
Does not work.
```
(focal)mrwizard@FocalCG1Dev:~/src/podman
$ bin/podman --runtime /usr/local/bin/crun-debug run --rm -it alpine sh
Error: cannot set limits without cgroups: OCI runtime error
jq .linux.resources /run/user/1000/crun/config.json
{
  "pids": {
    "limit": 4096
  }
}
```

pkg/specgen/generate/container.go
goochjj: rtc.PidsLimit() is returning 4096 on my cgroups v1 machine, rootless, even though the default does not show in run --help.

mheon:
Confirmed here as well.

mheon:
Ah, I think I know what's going on here. The code in containers/common for determining default limit is broken.

```
func (c *Config) PidsLimit() int64 {
        if unshare.IsRootless() {
                if c.Engine.CgroupManager == SystemdCgroupsManager {
                        cgroup2, _ := cgroupv2.Enabled()
                        if cgroup2 {
                                return c.Containers.PidsLimit
                        }
                        return 0
                }
        }
        return sysinfo.GetDefaultPidsLimit()
}
```

Notice that if cgroup manager is set to cgroupfs, we will drop through the conditionals and return `sysinfo.GetDefaultPidsLimit` - and v1 rootless defaults to cgroupfs.

goochjj:
> Ah, I think I know what's going on here. The code in containers/common for determining default limit is broken.
> 
> ```
> func (c *Config) PidsLimit() int64 {
>         if unshare.IsRootless() {
>                 if c.Engine.CgroupManager == SystemdCgroupsManager {
>                         cgroup2, _ := cgroupv2.Enabled()
>                         if cgroup2 {
>                                 return c.Containers.PidsLimit
>                         }
>                         return 0
>                 }
>         }
>         return sysinfo.GetDefaultPidsLimit()
> }
> ```
> 
> Notice that if cgroup manager is set to cgroupfs, we will drop through the conditionals and return `sysinfo.GetDefaultPidsLimit` - and v1 rootless defaults to cgroupfs.

if it's cgroupfs, we return 0 - that's what I see

goochjj:
> > Ah, I think I know what's going on here. The code in containers/common for determining default limit is broken.
> > ```
> > func (c *Config) PidsLimit() int64 {
> >         if unshare.IsRootless() {
> >                 if c.Engine.CgroupManager == SystemdCgroupsManager {
> >                         cgroup2, _ := cgroupv2.Enabled()
> >                         if cgroup2 {
> >                                 return c.Containers.PidsLimit
> >                         }
> >                         return 0
> >                 }
> >         }
> >         return sysinfo.GetDefaultPidsLimit()
> > }
> > ```
> > 
> > 
> > Notice that if cgroup manager is set to cgroupfs, we will drop through the conditionals and return `sysinfo.GetDefaultPidsLimit` - and v1 rootless defaults to cgroupfs.
> 
> if it's cgroupfs, we return 0 - that's what I see

NVM you're right.

goochjj:
> > > Ah, I think I know what's going on here. The code in containers/common for determining default limit is broken.
> > > ```
> > > func (c *Config) PidsLimit() int64 {
> > >         if unshare.IsRootless() {
> > >                 if c.Engine.CgroupManager == SystemdCgroupsManager {
> > >                         cgroup2, _ := cgroupv2.Enabled()
> > >                         if cgroup2 {
> > >                                 return c.Containers.PidsLimit
> > >                         }
> > >                         return 0
> > >                 }
> > >         }
> > >         return sysinfo.GetDefaultPidsLimit()
> > > }
> > > ```
> > > 
> > > 
> > > Notice that if cgroup manager is set to cgroupfs, we will drop through the conditionals and return `sysinfo.GetDefaultPidsLimit` - and v1 rootless defaults to cgroupfs.
> > 
> > 
> > if it's cgroupfs, we return 0 - that's what I see
> 
> NVM you're right.

The return 0 needs to move out one level probably.

goochjj:
Confirmed this fixes it:
```
--- a/vendor/github.com/containers/common/pkg/config/default.go
+++ b/vendor/github.com/containers/common/pkg/config/default.go
@@ -489,8 +489,8 @@ func (c *Config) PidsLimit() int64 {
                        if cgroup2 {
                                return c.Containers.PidsLimit
                        }
-                       return 0
                }
+               return 0
        }
        return sysinfo.GetDefaultPidsLimit()
 }
```

goochjj:
Though I'm not sure why cgroups v2 can't set a pids limit in cgroupfs mode - if the pids controller is in our cgroup, and we have write access to it, why can't we?

rhatdan:
@goochjj Ask @giuseppe I don't remember why.
Anyways lets get this in.

goochjj:
This merges a bunch of vendor dependencies which breaks the build.

Could you remove the runc bump from 1.0.0-rc90 to rc91?  It breaks things in buildah due to changes in libcontainer...

IMHO that bump should be a separate PR, allowing for fixes to that issue to be tracked under their own issue/PR.

I've verified removing the runc bump fixes the compile.

goochjj:
https://github.com/rhatdan/libpod/pull/6

goochjj:
@rhatdan https://github.com/goochjj/libpod/tree/rhatdan-pids-limit

Rebased, with runc rc90 and common 0.15.2

goochjj:
My branch updated and ready if #6897 is merged.

mheon:
I'm going to give this a final test on Debian to verify it fixes v1, but LGTM based on what I see now.

mheon:
Confirmed working on Debian with v1
/lgtm

PR #6845, opened 2020-07-02
stop podman service in e2e tests

baude:
when running e2e tests, each test knows to stop its service when running remote; however, during setup and teardown remote services were not being killed when we were done with them.


baude:
@rhatdan merge and rebase this should fix the hanging problem.

mheon:
LGTM

PR #6855, opened 2020-07-05
podman ps truncate the command

Luap99:
With a long create command the output from ps is basically unreadable.

This is a regression that was introduced with Podman 2.0.

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -309,7 +309,13 @@ func (l psReporter) Status() string {
 
 // Command returns the container command in string format
 func (l psReporter) Command() string {
-	return strings.Join(l.ListContainer.Command, " ")
+	command := strings.Join(l.ListContainer.Command, " ")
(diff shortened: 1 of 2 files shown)


cmd/podman/containers/ps.go
```diff
@@ -309,7 +309,13 @@ func (l psReporter) Status() string {
 
 // Command returns the container command in string format
 func (l psReporter) Command() string {
-	return strings.Join(l.ListContainer.Command, " ")
+	command := strings.Join(l.ListContainer.Command, " ")
+	if !noTrunc {
+		if len(command) > 17 {
+			return command[0:17] + "..."
```
mheon, line 315: I think you have an off-by-1 here, should be 0:16

Luap99: I looked at cmd/podman/shared/container.go in v1.9.3 branch and there it is 17.

rhatdan:
@vrothberg ptal

PR #6869, opened 2020-07-06
[CI:IMG] Bump crun to 1.14 on CI VMs

edsantiago:
This is an empty commit intended solely to get CI to rebuild
the VMs using crun 1.14 (up from 1.13). Twin goals:

  1) Be able to test #6693 (--sdnotify option); and
  2) Get rid of 'cgroup.freeze' CI flakes.

These are fixed by crun PRs 419 and 423 respectively.

### DIFF
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -59,7 +59,7 @@ case "${OS_RELEASE_ID}" in
         workaround_bfq_bug
 
 	# HACK: Need Conmon 2.0.17, currently in updates-testing on F31.
-	dnf update -y --enablerepo=updates-testing conmon
+	$BIGTO dnf update -y --enablerepo=updates-testing conmon
 
         if [[ "$ADD_SECOND_PARTITION" == "true" ]]; then
             bash "$SCRIPT_BASE/add_second_partition.sh"
--- a/Makefile
+++ b/Makefile
@@ -629,7 +629,7 @@ endef
 .PHONY: .install.ginkgo
 .install.ginkgo: .gopathok
 	if [ ! -x "$(GOBIN)/ginkgo" ]; then \
-		$(GO_BUILD) -o ${GOPATH}/bin/ginkgo ./vendor/github.com/onsi/ginkgo/ginkgo ; \
+		$(call go-get,github.com/onsi/ginkgo); \
 	fi
 
 .PHONY: .install.gitvalidation


edsantiago:
Looks like something is pretty badly broken; and it looks like ginkgo, which suggests #6859, which I have no possible way of debugging or fixing. So I give up. @cevich, heads-up that your next CI:IMG may fail.

cevich:
@edsantiago yes thank you, I've got a PR kicking around to address this.

cevich:
xref #6822

contrib/cirrus/setup_environment.sh
```diff
@@ -59,7 +59,7 @@ case "${OS_RELEASE_ID}" in
         workaround_bfq_bug
 
 	# HACK: Need Conmon 2.0.17, currently in updates-testing on F31.
-	dnf update -y --enablerepo=updates-testing conmon
+	$BIGTO dnf update -y --enablerepo=updates-testing conmon
```
cevich, line 62: This script is more intended for small adjustments that need to happen at runtime.  Since you're building new VM images here, might as well include this change into: `contrib/cirrus/packer/*_packaging.sh`.  Currently fedora is setup (I believe in `contrib/cirrus/packer/fedora_setup.sh`) to NOT use updates-testing on F31.  I'm not opposed to undoing that logic so updates-testing is used everywhere.

FWIW: I'm also not opposed to having this live in `setup_environment.sh` for the short-term.  Whatever you think is easiest/best.

edsantiago: @cevich when you say "might as well include this change", what are you referring to? Do you mean **my** change, that is, the addition of `$BIGTO` to a flaky command? Or do you mean e0d940463475b907d56aa548401f8ba916cec21b, the `dnf` command itself added 2020-06-02 to get conmon 2.0.17? The reason behind my change (adding `$BIGTO`) is that CI jobs have been failing frequently in this particular step. I wanted a quick band-aid for that, without having to look into all VMs and distros to see if conmon 2.0.17 has made it into stable everywhere.

Since there seem to be multiple `CI:IMG` PRs right now, and since CI is hosed anyway, I'm going to let this PR sit idle for now; I will revisit once things stabilize.

cevich: > I wanted a quick band-aid for that,

Ahh okay, then `setup_environment.sh` is fine, and you probably don't even need to go through the `CI:IMG` shenanigans either.  My other comments are !quick-band-aid, so it's okay to ignore them :smile:

edsantiago:
Whew... in addition to the ginkgo mess, CI is failing because there's a test for a `varlink` binary in `contrib/cirrus/check_image.sh`:
```bash
item_test "The varlink executable is present" -x "$(type -P varlink)" || let "NFAILS+=1"
```
Closing again; this is beyond my abilities.

PR #6870, opened 2020-07-06
Pids-limit should only be set if the user set it

rhatdan:
Currently we are sending over pids-limits from the user even if they
never modified the defaults.  The pids limit should be set at the server
side unless modified by the user.

This issue has led to failures on systems that were running with cgroups V1.


rhatdan:
@mheon @giuseppe @vrothberg This completes the fix for the pids-limit problem for v2.0.

mheon:
LGTM once this goes green

PR #6878, opened 2020-07-07
Add support for Filter query parameter to list images api

skorhone:
Docker api version 1.24 uses a query parameter named Filter for filtering images by names. In more recent versions of api name filter is in filters query parameter with other filters

This patch adds a mapping that translates Filter query parameter to Filters={"reference": [""]}

Patch resolves issue: #6797

### DIFF
--- a/pkg/api/handlers/utils/images.go
+++ b/pkg/api/handlers/utils/images.go
@@ -60,6 +60,7 @@ func GetImages(w http.ResponseWriter, r *http.Request) ([]*image.Image, error) {
 		All     bool
 		Filters map[string][]string `schema:"filters"`
 		Digests bool
+		Filter  string // Docker 1.24 compatibility
 	}{
 		// This is where you can override the golang default value for one of fields
 	}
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/utils/images.go
```diff
@@ -76,8 +77,16 @@ func GetImages(w http.ResponseWriter, r *http.Request) ([]*image.Image, error) {
 		err    error
 	)
 
-	if len(query.Filters) > 0 {
-		for k, v := range query.Filters {
+	queryFilters := query.Filters
+	if len(query.Filter) > 0 { // Docker 1.24 compatibility
```
vrothberg, line 81: Here, we need to make sure that we only accept this parameter for the docker-compat endpoints. There's a helper function to check if the request for libpod:
```golang
if !IsLibpodRequest(r) && en(query.Filter) > 0 { ...`

skorhone:
@vrothberg I'll have a look at those tests.

Also, it seems that currently name/reference filter is a bit too relaxed. It does seem to allow partial matches for names and tags. If it's much of an issue with name, but tag matching should be strict. Filter for image:v1 should not match with image:v10 like it currently does.

pkg/api/handlers/utils/images.go
skorhone: Makes sense. Added this check.

skorhone:
@vrothberg Now there's both positive and negative test cases

rhatdan:
@baude @mheon @QiWang19 @ashley-cui PTAL

PR #6881, opened 2020-07-07
fix race condition in `libpod.GetEvents(...)`

vrothberg:
Fix a race that could cause read errors to be masked.  Masking such
errors is likely to report red herrings since users don't see that
reading failed for some reasons but that a given event could not be
found.

### DIFF
--- a/libpod/events.go
+++ b/libpod/events.go
@@ -3,6 +3,7 @@ package libpod
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mheon:
LGTM, and a very good catch!

@baude @giuseppe @rhatdan PTAL

baude:
nice!
/hold

PR #6882, opened 2020-07-07
Revert runc bump to rc91

goochjj:
Lets separate out the dependbot update - common here, runc elsewhere.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/containers/libpod
 
 go 1.13
 
+replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc90
+
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/blang/semver v3.5.1+incompatible
--- a/go.sum
+++ b/go.sum
@@ -42,15 +42,12 @@ github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA
 github.com/cespare/xxhash v1.1.0/go.mod h1:XrSqR1VqqWfGrhpAt58auRo0WTKS1nRRg3ghfAqPWnc=
 github.com/checkpoint-restore/go-criu v0.0.0-20190109184317-bdb7599cd87b h1:T4nWG1TXIxeor8mAu5bFguPJgSIGhZqv/f0z55KCrJM=
 github.com/checkpoint-restore/go-criu v0.0.0-20190109184317-bdb7599cd87b/go.mod h1:TrMrLQfeENAPYPRsJuq3jsqdlRh3lvi6trTZJG8+tho=
-github.com/checkpoint-restore/go-criu/v4 v4.0.2/go.mod h1:xUQBLp4RLc5zJtWY++yjOoMoB5lihDt7fai+75m+rGw=
-github.com/cilium/ebpf v0.0.0-20200507155900-a9f01edf17e3/go.mod h1:XT+cAw5wfvsodedcijoh1l9cf7v1x9FlFB/3VmF/O8s=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd h1:qMd81Ts1T2OTKmB4acZcyKaMtRnY5Y44NuXGX2GFJ1w=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd/go.mod h1:sE/e/2PUdi/liOCUjSTXgM1o87ZssimdTWN964YiIeI=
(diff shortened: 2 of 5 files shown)


goochjj:
#6850

vrothberg:
I don't see other vendor changes. Did you run `make vendor` to update the dependencies?

goochjj:
I specifically undid the runc bump to rc91. It's breaking the build.

It's my assumption (since runc has experimental cgroups v2 now, and breaks the build, specifically in buildah), that the best path forward would be to merge the common 0.15.2 update, but NOT merge the runc bump to rc91.  Then runc compliance can be done in a separate PR, and only involves setting this back to rc91.

IMHO #6842 needs to be merged pronto because rootless is broken for cgroups v1.
#6842 needs common 0.15.2
#6842 did a make vendor, and got runc-rc91 which broke things.

Hence I'm trying to get #6842 out the door. Now, I've done a branch of @rhatdan's fix with this backreved, so right now if #6842 were merged (plus my fix) this would be the end result anyway.  I'm just unclear, direction-wise, if the plan is to
1) introduce this commit as a dependency update, either including this commit (to exclude runc) or squashing it
2) rebase 6842 so it's clear that his pids-limit commit is a strictly pids-limit commit, NOT a dependency update
3) rebase later dependency commits or have dependabot create another commit or introduce an issue to update to runc-rc91, realizing it's going to fail tests and require fixing

Regardless, the title of this PR is "Bump github.com/containers/common from 0.15.1 to 0.15.2 #6850" but it ACTUALLY ALSO updates runc to rc91... Which looks like an unintended consequence.

Please advise on procedure!

goochjj:
It's possible I'm doing this completely wrong.

It looks like containers/common was bumped to require runc 1.0.0rc91, and so no matter what I do it's going to keep importing that, unless I manually override it and check out rc90 in the vendor directory, which is probably less than ideal.

The real problem IMHO is that runc libcontainer had breaking changes in a minor release.

Advice on how this can be expedited welcome!

goochjj:
This will do it.

goochjj:
@vrothberg what do you think, should I just supercede #6850 with my own PR to bump containers/common to 0.15.2, squash this with the one dependabot did?  Then the tests can run here.

I can then open an issue for the build breakage w/ runc rc91 - which is in buildah, so should I do that issue in the buildah repo?  Or here?

mheon:
@goochjj Definitely file an issue against the Buildah repo.

As for the runc thing - it looks like it's just being dragged in by c/common, so I think we need to undo the vendor there, then revendor c/common?

goochjj:
@mheon True, but adding the replace line supercedes what common says - so it's a nice stop-gap to prevent needing common 0.15.3

This is the thing
common requires runc rc91 (just because it was bumped)
libpod requires common (so it gets rc91)
buildah requires common 0.14 - so it doesn't get the version of runc that has a problem
but libpod gets buildah, and has a NEWER runc.

goochjj:
Root cause is runc rc91 changed the way devicerules work because of the differences between cgroups v1 and v2.

Which causes:
```
vendor/github.com/containers/buildah/run_linux.go:142:72: cannot use d.DeviceRule.Permissions (type configs.DevicePermissions) as type string in argument to g.AddLinuxResourcesDevice
```

(It might break other things for all I know)

So

Possible solutions
1) Bump c/common dependency back to rc90, reversion, revendor, and let everything trickle down, so libpod will inherit rc90 from common.
2) Temp replace rc91 with rc90 in libpod, since it's the intersection of c/common and c/buildah that causes it, and either project is fine alone.
3) Update buildah to require c/common 0.15.2 and then fix the breakage

I think...?

Basically that means buildah is going to end up with code that EITHER works on rc90 or rc91 - (unless a conditional can be used) so I'm not sure what the best plan is to version these things properly.

giuseppe:
Let's skip runc rc91 as there is a regression that breaks cgroup v1 on systemd

goochjj:
That's fine, but skip where - c/common and reversion or use this PR - which blocks it in libpod?

goochjj:
Root cause:
https://github.com/opencontainers/runc/blob/819fcc687efb6e99d980a257f1dc8d92d02ba5bd/libcontainer/configs/device.go#L37

Device permissions are now a custom type

This commit
https://github.com/goochjj/libpod/commit/2b64224a8e715125a7038d0a61303a650244ed94

shows what needs to change in various libpod-things to support that.

mheon:
Requiring a fresh version of c/common is probably OK. We're not likely to cut a fresh release off master in the next few weeks (we're cutting a 2.0.2 on the v2.0 branch, which has an old-enough c/common that it should avoid this) so we have time to fix this in master (or potentially even fix this in Buildah and then vendor that) before we're in danger of releasing a broken build.

I am sort of interested as to how this snuck in without being detected by CI, though.

goochjj:
I saw it in the CI for Dan's pids-limit patch.

goochjj:
I imagine that:
1) Buildah's CI isn't including the bad version of runc since it depends on older c/common
2) Runc CI wouldn't expose it because all of runc's bits were updated to use the new custom type
3) c/commons CI (does it have CI?) wouldn't expose it because it largely has common code and probably does poke around in libcontainer


So libpod as the intersection of all those points is the one that exposed it...

goochjj:
conversation moved to issue
Will be solved by c/common new version

PR #6895, opened 2020-07-07
Add support for overlay volume mounts in podman.

QiWang19:
Add support for overlay volume mounts in podman.

### DIFF
--- a/cmd/podman/common/volumes.go
+++ b/cmd/podman/common/volumes.go
@@ -34,43 +34,43 @@ var (
 // Does not handle image volumes, init, and --volumes-from flags.
 // Can also add tmpfs mounts from read-only tmpfs.
 // TODO: handle options parsing/processing via containers/storage/pkg/mount
-func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bool) ([]spec.Mount, []*specgen.NamedVolume, error) {
+func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bool) ([]spec.Mount, []*specgen.NamedVolume, []*specgen.OverlayVolume, error) {
 	// Get mounts from the --mounts flag.
 	unifiedMounts, unifiedVolumes, err := getMounts(mountFlag)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	// Next --volumes flag.
-	volumeMounts, volumeVolumes, err := getVolumeMounts(volumeFlag)
+	volumeMounts, volumeVolumes, overlayVolumes, err := getVolumeMounts(volumeFlag)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	// Next --tmpfs flag.
 	tmpfsMounts, err := getTmpfsMounts(tmpfsFlag)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	// Unify mounts from --mount, --volume, --tmpfs.
 	// Start with --volume.
 	for dest, mount := range volumeMounts {
 		if _, ok := unifiedMounts[dest]; ok {
-			return nil, nil, errors.Wrapf(errDuplicateDest, dest)
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, dest)
 		}
 		unifiedMounts[dest] = mount
 	}
 	for dest, volume := range volumeVolumes {
 		if _, ok := unifiedVolumes[dest]; ok {
-			return nil, nil, errors.Wrapf(errDuplicateDest, dest)
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, dest)
 		}
 		unifiedVolumes[dest] = volume
 	}
 	// Now --tmpfs
 	for dest, tmpfs := range tmpfsMounts {
 		if _, ok := unifiedMounts[dest]; ok {
-			return nil, nil, errors.Wrapf(errDuplicateDest, dest)
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, dest)
 		}
 		unifiedMounts[dest] = tmpfs
 	}
@@ -101,15 +101,29 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
 		}
 	}
 
-	// Check for conflicts between named volumes and mounts
+	// Check for conflicts between named volumes, overlay volumes, and mounts
 	for dest := range unifiedMounts {
 		if _, ok := unifiedVolumes[dest]; ok {
-			return nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+		}
+		if _, ok := overlayVolumes[dest]; ok {
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
 		}
 	}
 	for dest := range unifiedVolumes {
 		if _, ok := unifiedMounts[dest]; ok {
-			return nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+		}
+		if _, ok := overlayVolumes[dest]; ok {
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+		}
+	}
+	for dest := range overlayVolumes {
+		if _, ok := unifiedMounts[dest]; ok {
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+		}
+		if _, ok := unifiedVolumes[dest]; ok {
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
 		}
 	}
 
@@ -119,7 +133,7 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
 		if mount.Type == TypeBind {
 			absSrc, err := filepath.Abs(mount.Source)
 			if err != nil {
-				return nil, nil, errors.Wrapf(err, "error getting absolute path of %s", mount.Source)
+				return nil, nil, nil, errors.Wrapf(err, "error getting absolute path of %s", mount.Source)
 			}
 			mount.Source = absSrc
 		}
@@ -129,8 +143,12 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
 	for _, volume := range unifiedVolumes {
 		finalVolumes = append(finalVolumes, volume)
 	}
+	finalOverlayVolume := make([]*specgen.OverlayVolume, 0)
+	for _, volume := range overlayVolumes {
+		finalOverlayVolume = append(finalOverlayVolume, volume)
+	}
 
-	return finalMounts, finalVolumes, nil
+	return finalMounts, finalVolumes, finalOverlayVolume, nil
 }
 
 // getMounts takes user-provided input from the --mount flag and creates OCI
@@ -465,9 +483,10 @@ func getNamedVolume(args []string) (*specgen.NamedVolume, error) {
 	return newVolume, nil
 }
 
-func getVolumeMounts(volumeFlag []string) (map[string]spec.Mount, map[string]*specgen.NamedVolume, error) {
+func getVolumeMounts(volumeFlag []string) (map[string]spec.Mount, map[string]*specgen.NamedVolume, map[string]*specgen.OverlayVolume, error) {
 	mounts := make(map[string]spec.Mount)
 	volumes := make(map[string]*specgen.NamedVolume)
+	overlayVolumes := make(map[string]*specgen.OverlayVolume)
 
 	volumeFormatErr := errors.Errorf("incorrect volume format, should be [host-dir:]ctr-dir[:option]")
 
@@ -481,7 +500,7 @@ func getVolumeMounts(volumeFlag []string) (map[string]spec.Mount, map[string]*sp
 
 		splitVol := strings.Split(vol, ":")
 		if len(splitVol) > 3 {
-			return nil, nil, errors.Wrapf(volumeFormatErr, vol)
+			return nil, nil, nil, errors.Wrapf(volumeFormatErr, vol)
 		}
 
 		src = splitVol[0]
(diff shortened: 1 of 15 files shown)


libpod/container.go
```diff
@@ -447,6 +449,19 @@ type ContainerNamedVolume struct {
 	Options []string `json:"options,omitempty"`
 }
 
+// ContainerOverlayVolume is a overlay volume that will be mounted into the
+// container. Each volume is a libpod Volume present in the state.
+type ContainerOverlayVolume struct {
+	// Destination is the absolute path where the mount will be placed in the container.
+	Dest string `json:"dest"`
+	// Type specifies the mount kind.
+	Type string `json:"type,omitempty" platform:"linux,solaris"`
```
mheon, line 458: Why is this here? I think we know type based on the fact that it's an overlay mount?

libpod/container.go
```diff
@@ -447,6 +449,19 @@ type ContainerNamedVolume struct {
...
+	// Source specifies the source path of the mount.
+	Source string `json:"source,omitempty"`
```
mheon, line 460: Can't we have multiple sources in an overlay mount?

libpod/container_internal.go
```diff
@@ -1425,7 +1425,7 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
 		defer func() {
 			if deferredErr != nil {
 				if err := c.unmountSHM(c.config.ShmDir); err != nil {
-					logrus.Errorf("Error unmounting SHM for container %s after mount error: %v", c.ID(), err)
+					logrus.Errorf(err.Error())
```
mheon, line 1428: Why did this change?

libpod/container_internal.go
```diff
@@ -1582,11 +1582,16 @@ func (c *Container) cleanupStorage() error {
 
 	for _, containerMount := range c.config.Mounts {
 		if err := c.unmountSHM(containerMount); err != nil {
-			if cleanupErr != nil {
-				logrus.Errorf("Error unmounting container %s: %v", c.ID(), cleanupErr)
-			}
+			logrus.Errorf(err.Error())
```
mheon, line 1585: Again, I don't think this should have changed

libpod/container_internal_linux.go
```diff
@@ -56,7 +57,9 @@ func (c *Container) mountSHM(shmOptions string) error {
 func (c *Container) unmountSHM(mount string) error {
 	if err := unix.Unmount(mount, 0); err != nil {
 		if err != syscall.EINVAL && err != syscall.ENOENT {
-			return errors.Wrapf(err, "error unmounting container %s SHM mount %s", c.ID(), mount)
+			logrus.Warnf("container %s failed to unmount %s : %v", c.ID(), mount, err)
```
mheon, line 60: I think this should stay an error

libpod/container_internal_linux.go
```diff
@@ -56,7 +57,9 @@ func (c *Container) mountSHM(shmOptions string) error {
...
+		} else {
+			logrus.Debugf("container %s failed to unmount %s : %v", c.ID(), mount, err)
```
mheon, line 62: This is a duplicate of what was printed below

libpod/container_internal_linux.go
```diff
@@ -316,6 +319,19 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 		}
 	}
 
+	// Add overlay volumes
+	for _, overlayVol := range c.config.OverlayVolumes {
+		contentDir, err := overlay.TempDir(c.config.StaticDir, c.RootUID(), c.RootGID())
```
mheon, line 324: Should this be the root UID/GID or the UID/GID of the user running in the container?

libpod/container_internal_linux.go
```diff
@@ -316,6 +319,19 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create TempDir in the %s directory", c.config.StaticDir)
+		}
+		overlayMount, err := overlay.Mount(contentDir, overlayVol.Source, overlayVol.Dest, c.RootUID(), c.RootGID(), c.runtime.store.GraphOptions())
```
mheon, line 328: It looks like this discards the options from the overlayVol?

libpod/options.go
```diff
@@ -1394,6 +1394,41 @@ func WithNamedVolumes(volumes []*ContainerNamedVolume) CtrCreateOption {
 	}
 }
 
+// WithOverlayVolumes adds the given overlay volumes to the container.
+func WithOverlayVolumes(volumes []*ContainerOverlayVolume) CtrCreateOption {
+	return func(ctr *Container) error {
```
mheon, line 1399: All of this should be done in Specgen - `WithOverlayVolumes` should just populate `c.config.OverlayVolumes`

pkg/specgen/specgen.go
```diff
@@ -432,6 +434,10 @@ type NamedVolume struct {
 	Options []string
 }
 
+// OverlayVolume holds information about a overlay volume that will be mounted into
+// the container.
+type OverlayVolume spec.Mount
```
mheon, line 439: This doesn't make sense, a lot of fields in spec.Mount are not used.

pkg/spec/storage.go
```diff
@@ -445,7 +445,7 @@ func getBindMount(args []string) (spec.Mount, error) {
 			}
 			setExec = true
 			newMount.Options = append(newMount.Options, kv[0])
-		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z":
+		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z", "O":
```
mheon, line 448: This shouldn't be necessary since we aren't handling `:O` as part of a regular volume

pkg/specgen/specgen.go
```diff
@@ -190,6 +190,8 @@ type ContainerStorageConfig struct {
 	// there are conflicts.
 	// Optional.
 	Volumes []*NamedVolume `json:"volumes,omitempty"`
+	// Overlay volumes are named volumes that will be added to the container.
```
mheon, line 193: Please note that this is Optional.

cmd/podman/common/volumes.go
```diff
@@ -104,12 +104,12 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
 	// Check for conflicts between named volumes and mounts
```
mheon, line 104: We need conflict checks for the new overlay volumes, as we have for named volumes and normal bind mounts right now.

cmd/podman/common/volumes.go
```diff
@@ -496,34 +501,51 @@ func getVolumeMounts(volumeFlag []string) (map[string]spec.Mount, map[string]*sp
...
 			// This is not a named volume
-			newMount := spec.Mount{
-				Destination: cleanDest,
-				Type:        string(TypeBind),
-				Source:      src,
-				Options:     options,
-			}
-			if _, ok := mounts[newMount.Destination]; ok {
-				return nil, nil, errors.Wrapf(errDuplicateDest, newMount.Destination)
+			for _, o := range options {
```
mheon, line 522: This loop makes no sense. We should check every option to determine if `O` is present, and then add *after* the loop based on whether it is or is not.

libpod/container_internal_linux.go
QiWang19: the overlay pkg has hardcoded options, @rhatdan can other options be used here?

rhatdan: Theoretically yes, they could be added. I guess you could specify RO, but a readonly overlay makes little sense.
The private, shared, slave might be able to be passed down.
For now we could throw an error when they are passed in, and wait for someone to complain.  I am not sure how well these options would work.

libpod/options.go
QiWang19: Do you mean `cmd/podman/common/specgen.go`?  Why can't keep this like `WithNamedVolumes` above?

mheon: Hmm. Didn't realize that we checked again there.

We should retain the ProcessOptions part, but move the destination check to container_validate.go and have it check both named volumes and overlay volumes.

libpod/container_internal_linux.go
mheon: I'd be fine with not allowing options at all here - just remove them from the struct, and error at the command line when someone tries to pass them.

mheon: (We can always re-add them later)

libpod/container.go
QiWang19: Can multiple sources be used by allowing multiple `-v source:dest` with same dest?

mheon: I remember we had a discussion about this. @rhatdan Did we decide to allow multiple sources? If there's any possibility, we should make this an array.

rhatdan: If we want that, we probably need a different command, maybe something with --mount.  I think for --volume we should keep it simple and not complicate the option parsing to much.

QiWang19: Change back to treat multiple overlay sources with the same dest as conflict

QiWang19:
@mheon PTAL

cmd/podman/common/volumes.go
```diff
@@ -101,15 +101,21 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
...
 	for dest := range unifiedMounts {
 		if _, ok := unifiedVolumes[dest]; ok {
-			return nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+		}
+		if _, ok := overlayVolumes[dest]; ok {
+			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
 		}
 	}
 	for dest := range unifiedVolumes {
```
mheon, line 113: We need another one of these for overlay volumes, to check that overlay volume destinations are not defined for named volumes or mounts.

cmd/podman/common/volumes.go
```diff
@@ -496,34 +507,55 @@ func getVolumeMounts(volumeFlag []string) (map[string]spec.Mount, map[string]*sp
...
+				}
 			}
-			if _, ok := mounts[newMount.Destination]; ok {
-				return nil, nil, errors.Wrapf(errDuplicateDest, newMount.Destination)
+			if overlayFlag {
+				// This is a overlay volume
+				newOverlayVol := new(specgen.OverlayVolume)
+				newOverlayVol.Destination = cleanDest
+				newOverlayVol.Source = src
+				newOverlayVol.Options = options
```
mheon, line 542: Why are we setting this if we explicitly require no options above?

cmd/podman/common/volumes.go
```diff
@@ -496,34 +507,55 @@ func getVolumeMounts(volumeFlag []string) (map[string]spec.Mount, map[string]*sp
...
+				if _, ok := overlayVolumes[newOverlayVol.Destination]; ok {
```
mheon, line 543: Why are we doing this here? This should be done as part of `parseVolumes` where the other conflict checks are done.

mheon: Ah, nevermind, this is for internal conflicts.

docs/source/markdown/podman-create.1.md
```diff
@@ -930,6 +939,37 @@ content label. Shared volume labels allow all containers to read/write content.
...
+
+  Subsequent executions of the container will see the original source directory
+content, any changes from previous container executions no longer exists.
+
+  One use case of the overlay mount is sharing the package cache from the
+host into the container to allow speeding up builds.
+
+  Note:
+
+     - The `O` flag conflicts with the `Z` and `z` flags.
```
mheon, line 960: It conflicts with all options now, doesn't it?

libpod/container.go
```diff
@@ -447,6 +449,17 @@ type ContainerNamedVolume struct {
...
 
+// ContainerOverlayVolume is a overlay volume that will be mounted into the
+// container. Each volume is a libpod Volume present in the state.
+type ContainerOverlayVolume struct {
+	// Destination is the absolute path where the mount will be placed in the container.
+	Dest string `json:"dest"`
+	// Source specifies the source path of the mount.
+	Source string `json:"source,omitempty"`
+	// Options are fstab style mount options.
+	Options []string `json:"options,omitempty"`
```
mheon, line 460: Didn't we remove volumes above?

mheon: Sorry, remove options* above

pkg/util/mountOpts.go
```diff
@@ -100,12 +100,12 @@ func ProcessOptions(options []string, isTmpfs bool, sourcePath string) ([]string
 				return nil, errors.Wrapf(ErrDupeMntOption, "only one of 'rbind' and 'bind' can be used")
 			}
 			foundBind = true
-		case "z", "Z":
+		case "z", "Z", "O":
```
mheon, line 103: We should probably enforce that O is only allowed if it's the only option here?

cmd/podman/common/volumes.go
QiWang19: Do you mean the above checking len(options)? the above is not requring no options, it's requiring that only one option is allowed.

pkg/util/mountOpts.go
QiWang19: Added a case to handle 'O', but can we leave the `ProcessOptions` as it is without checking the 'O' option, since the error threw earlier when paring the CLI?

libpod/container.go
QiWang19: Are you suggesting changing the comment and variable name to option, or removing the Options field?

cmd/podman/common/volumes.go
mheon: Yes - and that option is `O` which we should be discarding. no options that are not O are permissable, so Overlay Volumes don't use options right now.

QiWang19:
@mheon PTAL

QiWang19:
@mheon @rhatdan PTAL

test/e2e/run_volume_test.go
```diff
@@ -450,4 +451,63 @@ VOLUME /test/`
 		Expect(data[0].Mounts[0].Source).To(Equal("/tmp"))
 		Expect(data[0].Mounts[0].Destination).To(Equal("/test"))
 	})
+
+	It("podman run with overlay volume flag", func() {
```
mheon, line 455: Can we get some tests on conflicts? Adding an overlay volume at the same path as a named volume or bind mount or tmpfs should error.

libpod/container_internal_linux.go
```diff
@@ -319,6 +320,29 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
 	}
 
+	// Add overlay volumes
+	uid := c.RootUID()
+	gid := c.RootGID()
+	idMappings, err := c.IDMappings()
+	if err != nil {
+		return nil, err
+	}
+	if len(idMappings.UIDMap) > 0 && len(idMappings.GIDMap) > 0 {
```
mheon, line 330: Do we need this? I thought RootUID and RootGID already looked through the mappings?

mheon:
One more comment, then LGTM

@rhatdan @vrothberg @giuseppe @baude PTAL

libpod/container_internal_linux.go
QiWang19: Add this because https://github.com/containers/podman/pull/6895#discussion_r451133191, is this the right way to get the uid/gid of the user running in the container?

rhatdan: @giuseppe ^^

rhatdan:
LGTM

libpod/container_internal_linux.go
mheon: I'd like this answered before merge, otherwise LGTM

QiWang19: @giuseppe PTAL

rhatdan:
@giuseppe We are waiting for your review so we can get this merged.

libpod/container_internal_linux.go
giuseppe: yes, I don't think we need the extra check, RootUID/RootGID are enough.

Also it seems to not check the container id but just get the first id available, it can be different than root as the mappings are not ordered

giuseppe:
> @giuseppe We are waiting for your review so we can get this merged.

reviewed and left a comment

mheon:
LGTM

QiWang19:
@mheon @rhatdan PTAL

rhatdan:
Great work @QiWang19 getting this merged.

PR #6896, opened 2020-07-07
Fix container and pod create commands for remote create

mheon:
In `podman inspect` output for containers and pods, we include the command that was used to create the container. This is also used by `podman generate systemd --new` to generate unit files.

With remote podman, the generated create commands were incorrect since we sourced directly from os.Args on the server side, which was guaranteed to be `podman system service` (or some variant thereof). The solution is to pass the command along in the Specgen or PodSpecgen, where we can source it from the client's os.Args.

This will still be VERY iffy for mixed local/remote use (doing a `podman --remote run ...` on a remote client then a `podman generate systemd --new` on the server on the same container will not work, because the `--remote` flag will slip in) but at the very least the output of `podman inspect` will be correct. We can look into properly handling `--remote` (parsing it out would be a little iffy) in a future PR.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -401,6 +401,9 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	}
 	var command []string
 
+	// Include the command used to create the container.
+	s.ContainerCreateCommand = os.Args
+
 	// Build the command
 	// If we have an entry point, it goes first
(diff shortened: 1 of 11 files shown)


mheon:
@baude PTAL, this should fix one of the remote tests

rhatdan:
LGTM

rhatdan:
@vrothberg @baude @giuseppe @QiWang19 @ashley-cui @TomSweeneyRedHat @jwhonce PTAL

PR #6897, opened 2020-07-08
Bump github.com/containers/common from 0.15.1 to 0.15.2 runc rc90

goochjj:
Bumps [github.com/containers/common](https://github.com/containers/common) from 0.15.1 to 0.15.2.
- [Release notes](https://github.com/containers/common/releases)
- [Commits](https://github.com/containers/common/compare/v0.15.1...v0.15.2)

JJG: Pin Runc to v1.0.0-rc90 with replace until we can fix compilation w/ buildah

Signed-off-by: dependabot-preview[bot] <support@dependabot.com>
Signed-off-by: Joseph Gooch <mrwizard@dok.org>

This PR is an alternate to the dependabot issued one - that does the make vendor step, and pins runc at rc90 so we can continue development without blocking on Buildah.

PR #6898, opened 2020-07-08
[WIP] Runc v1.0.0-rc91 fixes

goochjj:
Building off of #6897 ,  bumps the rev up to RC91, and then fixes things.

Two of the commits actually depend in Buildah, the other two are proper here.  The PR would look less scary if it were properly based on #6897's branch but it's in my repo so I don't think it would look right.

Consider after 6897 is merged.

/hold


rhatdan:
Replaced with https://github.com/containers/podman/pull/6906

Issue #6899, opened 2020-07-08
[APIv2] events?filters return 500 


skywalker512:
/kind bug

**Description**

treafik v2.2.1 will  call on `/v1.24/events?filters=%7B%22type%22%3A%7B%22container%22%3Atrue%7D%7D` but return 500 

**Steps to reproduce the issue:**

1. podman system service -t 0 tcp:0.0.0.0:9999

2. start treafik with `endpoint = "tcp://127.0.0.1:9999" ` config

**Describe the results you received:**

```
{
    "cause": "runtime error: index out of range [0] with length 0",
    "message": "runtime error: index out of range [0] with length 0",
    "response": 500
}
```

**Describe the results you expected:**

```
{"Type":"container","Action":"stop","Actor":{"ID":"7f6e4f3b81349a57bd253cebb20f7b4fa98597d331f23fe767f862dcc21ece34","Attributes":{"containerExitCode":"0","image":"docker.io/library/traefik:latest","name":"traefik-main"}},"scope":"local","time":1594180142,"timeNano":1594180142023436891}
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.2
API Version:  1
Go Version:   go1.14
Git Commit:   201c9505b88f451ca877d29a73ed0f1836bb96c7-dirty
Built:        Thu Jan  1 08:00:01 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.19-dev, commit: ce64ca841e1b12954cb3c972e0c06ec69a7b31ce'
  cpus: 2
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: file
  hostname: iZ2ze8mxxl9wg4rrytkeuqZ
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.4.0-31-generic
  linkmode: static
  memFree: 1043369984
  memTotal: 4127342592
  ociRuntime:
    name: runc
    package: Unknown
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc91
      spec: 1.0.2-dev
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
[...]

PR #6902, opened 2020-07-08
events endpoint: fix panic and race condition

vrothberg:
Fix a potential panic in the events endpoint when parsing the filters
parameter.  Values of the filters map might be empty, so we need to
account for that instead of uncondtitionally accessing the first item.

Also apply a similar fix for race conditions as done in commit f4a2d25c0fca:

	Fix a race that could cause read errors to be masked.  Masking
	such errors is likely to report red herrings since users don't
	see that reading failed for some reasons but that a given event
	could not be found.

Fixes: #6899

### DIFF
--- a/pkg/api/handlers/compat/events.go
+++ b/pkg/api/handlers/compat/events.go
@@ -1,9 +1,9 @@
 package compat
 
 import (
-	"context"
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/containers/libpod/v2/libpod"
 	"github.com/containers/libpod/v2/libpod/events"
@@ -17,10 +17,10 @@ import (
 
 func GetEvents(w http.ResponseWriter, r *http.Request) {
 	var (
-		fromStart   bool
-		eventsError error
-		decoder     = r.Context().Value("decoder").(*schema.Decoder)
-		runtime     = r.Context().Value("runtime").(*libpod.Runtime)
+		fromStart bool
+		decoder   = r.Context().Value("decoder").(*schema.Decoder)
+		runtime   = r.Context().Value("runtime").(*libpod.Runtime)
+		json      = jsoniter.ConfigCompatibleWithStandardLibrary // FIXME: this should happen on the package level
 	)
 
 	query := struct {
@@ -33,11 +33,16 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
 	}
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
 		utils.Error(w, "Failed to parse parameters", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse parameters for %s", r.URL.String()))
+		return
 	}
 
 	var libpodFilters = []string{}
 	if _, found := r.URL.Query()["filters"]; found {
 		for k, v := range query.Filters {
+			if len(v) == 0 {
+				utils.Error(w, "Failed to parse parameters", http.StatusBadRequest, errors.Errorf("empty value for filter %q", k))
+				return
+			}
 			libpodFilters = append(libpodFilters, fmt.Sprintf("%s=%s", k, v[0]))
 		}
 	}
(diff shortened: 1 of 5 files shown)


vrothberg:
@baude @mheon PTAL  before of the spaghetti

pkg/api/handlers/compat/events.go
```diff
@@ -46,18 +52,21 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
 		fromStart = true
 	}
 
+	// Note: the code looks a bit like spaghetti as a couple of things
+	// happen concurrently.  The events API does not have an error channel,
+	// so reading from the events channel and checking for an error must
+	// be handled us here.  We may also stream the events, which adds yet
```
rhatdan, line 58: nit `be handled here...`

rhatdan:
LGTM, one nit.

pkg/api/handlers/compat/events.go
vrothberg: Fixed, thanks!

pkg/api/handlers/compat/events.go
```diff
@@ -68,24 +77,52 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
...
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
-	if flusher, ok := w.(http.Flusher); ok {
-		flusher.Flush()
-	}
+	// Write the events from the channel.
+	go func() {
+		first := true
+		var coder *jsoniter.Encoder
+		json := jsoniter.ConfigCompatibleWithStandardLibrary
```
mheon, line 84: Why do we need to redefine this? This should be set for the entire package

pkg/api/handlers/compat/events.go
```diff
@@ -68,24 +77,52 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
...
+			w.WriteHeader(http.StatusOK)
+			if flusher, ok := w.(http.Flusher); ok {
+				flusher.Flush()
+			}
 
-	for event := range eventChannel {
-		e := entities.ConvertToEntitiesEvent(*event)
-		if err := coder.Encode(e); err != nil {
-			logrus.Errorf("unable to write json: %q", err)
+			coder = json.NewEncoder(w)
```
mheon, line 95: Why are these in writeHeader? They seem like they ought to be part of initializing variables above

pkg/api/handlers/compat/events.go
```diff
@@ -68,24 +77,52 @@ func GetEvents(w http.ResponseWriter, r *http.Request) {
...
+			coder.SetEscapeHTML(true)
 		}
-		if flusher, ok := w.(http.Flusher); ok {
-			flusher.Flush()
+
+		for event := range eventChannel {
```
mheon, line 99: Do we have a guarantee this terminates in non-error conditions? I don't think that the channel is closed automatically when events run out, so we could block forever here.

pkg/api/handlers/compat/events.go
vrothberg: Let me rewrite a bit more. The code is hard to parse as is.

pkg/api/handlers/compat/events.go
vrothberg: No idea. I will add a TODO as I consider it to be out of context of this PR. I don't want to change to much in a fix.

vrothberg:
@mheon, I did another round of refactoring and believe we're at a fairly elegant place now

pkg/api/handlers/compat/events.go
mheon: I was going to say that I was fairly certain it was already set, but it looks like I'm wrong. Which is actually quite concerning, perf-wise - we have a super-efficient JSON encoder/decoder but we're just not using it in the API, which deals heavily with JSON.

TomSweeneyRedHat:
@vrothberg tests are falling down all over.

vrothberg:
Almost there. I'll take another look. Want to get this in eventually

rhatdan:
@giuseppe @mheon @baude @jwhonce @ashley-cui @QiWang19 @TomSweeneyRedHat PTAL

vrothberg:
@mheon we need that in v2.0

vrothberg:
Please wait before merging. I will add another commit to fix https://github.com/containers/podman/issues/6899.

vrothberg:
> Please wait before merging. I will add another commit to fix #6899.

Done. @rhatdan can you take another look?

pkg/api/handlers/compat/events.go
```diff
@@ -15,77 +16,127 @@ import (
...
+	if err := json.Unmarshal(raw, &filters); err == nil {
+		for filterKey, filterSlice := range filters {
+			for _, filterValue := range filterSlice {
+				libpodFilters = append(libpodFilters, fmt.Sprintf("%s=%s", filterKey, filterValue))
+			}
+		}
+		return libpodFilters
+	}
+
+	return nil
```
rhatdan, line 56: This function should return an err correct?  If the second unmarshal fails, we got bad data

vrothberg: Good catch, yes! I'll wrap it up and repush in a bit.

pkg/api/handlers/compat/events.go
```diff
@@ -15,77 +16,132 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// filtersFromRequests extracts the "filters" parameter from the specified
+// http.Request.  The paramater can either be a `map[string][]string` as done
```
TomSweeneyRedHat, line 20: nit of a nit, only change if you've other changes inbound.
```suggestion
// http.Request.  The parameter can either be a `map[string][]string` as done
```

TomSweeneyRedHat:
LGTM but I want a @mheon head nod before merging.

Issue #4251, opened 2019-10-13
Podman Pull Images always error when docker not


benyaminl:
/kind bug
Hello. I'm quite frustated with podman when trying to pull images. I take hours and hours to take images but always fail. This's the podman version
```
host:
  BuildahVersion: 1.9.0
  Conmon:
    package: podman-1.4.4-4.el7.centos.x86_64
    path: /usr/libexec/podman/conmon
    version: 'conmon version 0.3.0, commit: unknown'
  Distribution:
    distribution: '"centos"'
    version: "7"
  MemFree: 2073849856
  MemTotal: 3806199808
  OCIRuntime:
    package: containerd.io-1.2.10-3.2.el7.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc8+dev
      commit: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
      spec: 1.0.1-dev
  SwapFree: 3877470208
  SwapTotal: 3892310016
  arch: amd64
  cpus: 4
  hostname: thinkpad.localdomain
  kernel: 5.3.5-1.el7.elrepo.x86_64
  os: linux
  rootless: true
  uptime: 30h 5m 1.65s (Approximately 1.25 days)
registries:
  blocked: null
  insecure: null
  search:
  - registry.access.redhat.com
  - docker.io
  - registry.fedoraproject.org
  - quay.io
  - registry.centos.org
store:
  ConfigFile: /home/ben/.config/containers/storage.conf
  ContainerStore:
    number: 1
  GraphDriverName: vfs
  GraphOptions: null
  GraphRoot: /home/ben/.local/share/containers/storage
  GraphStatus: {}
  ImageStore:
    number: 3
  RunRoot: /tmp/1000
  VolumePath: /home/ben/.local/share/containers/storage/volumes

```

I tried to pull docker.io/benyaminl/lap7-dev:lumen
When I'm on docker it's running okay, but with podman always error. 
```
Copying blob 1e50d3b4ec49 done
Copying blob 3f410603f6a7 done
Copying blob 0d163c86f78f done
Copying blob 9375fe5d3314 done
Copying blob 32c10dbc9ed1 done
Copying blob a2badc3a0924 done
Copying blob a017eb278f96 done
Copying blob 9550adab8c09 done
[...]

PR #6905, opened 2020-07-08
Retry pulling image

QiWang19:
- Update the containers/common
- Use shared retry functions from containers/common. Wrap the inner helper in the retry function. Functions pullimage failed with retriable error will default maxretry 3 times using exponential backoff.
related: https://github.com/containers/podman/issues/4251#issuecomment-654463281

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -14,6 +14,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containers/common/pkg/retry"
 	cp "github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/directory"
 	dockerarchive "github.com/containers/image/v5/docker/archive"
@@ -74,6 +75,8 @@ type InfoImage struct {
 	// Layers stores all layers of image.
 	Layers []LayerInfo
 }
+
+const maxRetry = 3
 
 // ImageFilter is a function to determine whether a image is included
 // in command output. Images to be outputted are tested using the function.
@@ -158,7 +161,7 @@ func (ir *Runtime) New(ctx context.Context, name, signaturePolicyPath, authfile 
 	if signaturePolicyPath == "" {
 		signaturePolicyPath = ir.SignaturePolicyPath
 	}
-	imageName, err := ir.pullImageFromHeuristicSource(ctx, name, writer, authfile, signaturePolicyPath, signingoptions, dockeroptions, label)
+	imageName, err := ir.pullImageFromHeuristicSource(ctx, name, writer, authfile, signaturePolicyPath, signingoptions, dockeroptions, &retry.RetryOptions{MaxRetry: maxRetry}, label)
 	if err != nil {
 		return nil, errors.Wrapf(err, "unable to pull %s", name)
 	}
@@ -176,7 +179,7 @@ func (ir *Runtime) LoadFromArchiveReference(ctx context.Context, srcRef types.Im
 	if signaturePolicyPath == "" {
 		signaturePolicyPath = ir.SignaturePolicyPath
 	}
-	imageNames, err := ir.pullImageFromReference(ctx, srcRef, writer, "", signaturePolicyPath, SigningOptions{}, &DockerRegistryOptions{})
+	imageNames, err := ir.pullImageFromReference(ctx, srcRef, writer, "", signaturePolicyPath, SigningOptions{}, &DockerRegistryOptions{}, &retry.RetryOptions{MaxRetry: maxRetry})
 	if err != nil {
 		return nil, errors.Wrapf(err, "unable to pull %s", transports.ImageName(srcRef))
 	}
--- a/libpod/image/pull.go
+++ b/libpod/image/pull.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/containers/common/pkg/retry"
 	cp "github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/directory"
 	"github.com/containers/image/v5/docker"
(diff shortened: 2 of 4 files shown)


rhatdan:
@QiWang19 Please fixup you PR explanation,  Explain what is going to happen

rhatdan:
Do we need this for podman create and podman run?

QiWang19:
This retry also works for create and run since the retry function wrapped the inner helper method that is used by create and run.

rhatdan:
Great.

libpod/image/pull.go
```diff
@@ -310,9 +310,12 @@ func (ir *Runtime) doPullImage(ctx context.Context, sc *types.SystemContext, goa
 				return nil, err
 			}
 		}
-
-		_, err = cp.Image(ctx, policyContext, imageInfo.dstRef, imageInfo.srcRef, copyOptions)
-		if err != nil {
+		destRef := imageInfo.dstRef
```
rhatdan, line 313: Nit: Why do you need to use variables?

libpod/image/utils.go
```diff
@@ -174,3 +182,63 @@ func imageNameForSaveDestination(img *Image, imgUserInput string) string {
 	}
 	return fmt.Sprintf("%s%s", prepend, imgUserInput)
 }
+
+func retryIfNecessary(operation func() error) error {
```
rhatdan, line 186: Can you put this code into containers/common, since podman, Buildah and Skopeo are all going to use it?

rhatdan: This functionality along with @ParkerVR pulling of shortnames should be put into common/pkg/image  shared functions.

QiWang19: I can draft a PR in c/common, will this be common/pkg/image/image.go or pull.go? @ParkerVR?

ParkerVR: I have been testing locally with pkg/image/pull.go, which I think makes more sense for consistency.

rhatdan:
@QiWang19 Open the PR against containers/common

libpod/image/pull.go
QiWang19: Gating tests complain about using `imageInfo` in the retry loop cause it's a variable from the outer loop.

libpod/image/utils.go
rhatdan: containers/common PR has been merged.

rhatdan:
@QiWang19 containers/common merged, can you update this PR?

rhatdan:
LGTM
@ParkerVR @mtrmac @mheon @vrothberg PTAL

libpod/image/image.go
```diff
@@ -14,6 +14,7 @@ import (
 	"syscall"
 	"time"
 
+	commonimage "github.com/containers/common/pkg/image"
```
TomSweeneyRedHat, line 17: Is @mtrmac happy with this package name?  I thought he raised concerns over in common (https://github.com/containers/common/pull/229) and I'd like to settle that before we move forward with this PR.

TomSweeneyRedHat:
Changes LGTM , but I want a head nod from @mtrmac on the package naming before moving forward.

libpod/image/pull.go
mtrmac: If that warning is about the loop variable changing in next iteration, affecting the closure, an `imageInfo := imageInfo` might work just as well.

libpod/image/image.go
mtrmac: That really depends on what else is planned to go into that package.

The retry code is somewhat image-independent (but not entirely, it does check for registry errors, and drag in a registry subpackage) and could be used for other network operations, so I don’t think it’s a good name for a package containing _only_ this code. A larger package containing several more image-related utilities might make more sense, OTOH it’s currently not clear to me that that would have to be a single package versus a set of utility subpackages (from which only a subset could be used to minimize dependencies in consumers).

QiWang19: This package plans to merge the work from  @ParkerVR pulling of shortnames .

mtrmac: Yeah — is there value in having both in a single subpackage? Skopeo won’t use the short name part, at least.

(I don’t really want to make a big deal out of this, the c/common has no API guarantee, it’s not _that much_ code I expect, and the repos are tightly coupled enough that we can rearrange any time. I guess dragging in the full containers.conf config file, or some user-observable non-optional initialization behavior, into Skopeo would be the point I’d start feeling stronger about this.)

mheon: I don't object to having them both in the same package, but I do object to the name. Importing both `github.com/containers/libpod/v2/libpod/image` and `github.com/containers/common/pkg/image` in the same file is horribly confusing and renaming one at import time does not make things much better.

QiWang19: thanks @mtrmac . I moved it to a single pkg. PTAL https://github.com/containers/common/pull/237

rhatdan: I definitely do not like the name retry, it is too small, I am leaning torwards containers/common/pkg/pull.

ParkerVR: I agree with Dan - but do we want more organization with something like containers/common/pkg/image/pull ? This might give us more space for the future with less refactoring

rhatdan:
@QiWang19 What is going on with this PR?  Can we move this to no longer WIP?

rhatdan:
/approve
LGTM
@QiWang19 Can you remove [wip] from description?

QiWang19:
[wip] removed

rhatdan:
@QiWang19 is skopeo and Buildah now using the same code from containers/common?

QiWang19:
> @QiWang19 is skopeo and Buildah now using the same code from containers/common?

No, they are not using c/common. Will open PRs for that.

PR #6911, opened 2020-07-08
Bors: Initial implementation

cevich:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -92,6 +92,9 @@ gating_task:
     # Only run this on PRs, never during post-merge testing (for speed).
     only_if: $CIRRUS_BRANCH != $DEST_BRANCH
 
+    # see bors.toml
+    skip: $CIRRUS_BRANCH =~ ".*\.tmp"
+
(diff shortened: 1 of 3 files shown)


cevich:
@baude @mheon In case you want it, I believe the only other step needed is to point the bors app at this repo and set the bot permissions on it's settings page.

rhatdan:
Personally I am not a big fan of Bors, I find it much slower the /lgtm in Podman, bors r+ seems to go off into the weeds and can take hours.

cevich:
All valid criticisms.  I'd also add: It's written in haskel, a language (I'm assuming) not many people speak.  Bors long-term survive-ability may be limited.  There is another, more modern bot + green-tree service we could use (don't recall the name off-hand).  It will also manage issue and PR labeling.  Unfortunately, it's configuration is rather tricky and it doesn't play very nicely with github-actions (not disastrous, rather it refuses to run under some specific circumstances).

I'll close this for now, but if you want me to prototype the other service here or buildah or elsewhere, just let me know.

PR #6913, opened 2020-07-09
Fix: Hijacking endpoints to follow rfc 7230 semantics

skorhone:
After this patch v2 hijacking endpoints, exec/start and containers/attach follow rfc 7230 specification.

Connection will only be upgraded, if client specifies upgrade headers:

For tcp connections:

```
Connection: Upgrade
Upgrade: tcp
```

For unix socket connections:

```
Connection: Upgrade
Upgrade: sock
```

There are currently no checks if upgrade type actually matches with available protocols. Implementation just echoes protocol that client requested

### DIFF
--- a/pkg/api/handlers/compat/containers_attach.go
+++ b/pkg/api/handlers/compat/containers_attach.go
@@ -1,8 +1,12 @@
 package compat
 
 import (
+	"bufio"
 	"fmt"
+	"io"
+	"net"
 	"net/http"
+	"strings"
 
 	"github.com/containers/libpod/v2/libpod"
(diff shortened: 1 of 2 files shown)


skorhone:
Is there an image for long running container that I could use for testing execs? There seems to be no e2e tests for exec.

Would be nice, if recommended images were documented.

mheon:
We'd generally do an alpine container and run `top` inside.

skorhone:
This request is included also in #6918

jwhonce:
Merged #6918

PR #6914, opened 2020-07-09
Fix: Remove hijacked connections from active connections list

skorhone:
StateHijacked is a terminal state. If hijacked connection is registered as an active connection, connection will never be unregistered. This causes two issues

First issue is that active connection counters are off.

Second issue is a resource leak caused by connection object that is stored to a map.

After this patch hijacked connections are no longer visible in counters. If a counter for hijacked connections is required, podman must track connections returned by Hijacker.Hijack()

It might make sense to develop abstraction layer for hijacking - and move all hijacking related code to a
separate package. Hijacking code is prone to resource leaks and it should be thoroughly tested.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -255,14 +255,14 @@ func (t *IdleTracker) ConnState(conn net.Conn, state http.ConnState) {
 	oldActive := len(t.active)
 	logrus.Debugf("IdleTracker %p:%v %d/%d connection(s)", conn, state, t.ActiveConnections(), t.TotalConnections())
 	switch state {
-	case http.StateNew, http.StateActive, http.StateHijacked:
+	case http.StateNew, http.StateActive:
 		t.active[conn] = struct{}{}
 		// stop the timer if we transitioned from idle
 		if oldActive == 0 {
 			t.timer.Stop()
 		}
 		t.total++
-	case http.StateIdle, http.StateClosed:
+	case http.StateIdle, http.StateClosed, http.StateHijacked:
 		delete(t.active, conn)
 		// Restart the timer if we've become idle
 		if oldActive > 0 && len(t.active) == 0 {


mheon:
My concern here would be that this potentially causes the server to think we have 0 connections when there is only a hijacked connection active - we use the active connection counter to determine whether to shut the server down when in socket-activated mode so this could cause attach sessions to be auto-terminated after a timeout.

skorhone:
@mheon I'm not sure which is worse. This patch doesn't intend to fix underlying problem related to tracking hijacked connections. As I mentioned in description, hijacked connections need to be tracked using some other mechanism. 

Short explanation:

Once you hijack connection you will not get http.StateClosed signal. hijacking means that http server is no longer in control of connection's lifcecycle. Hijacking is meant for switching http protocol to some other protocol. This explains why http.StateHijacked is a terminal signal and it also explains why procedure is called hijacking.

I'd much rather implement connection tracking for hijacked connections as a separate pull request. It's likely to require considerable effort to do all required changes.

Btw, doesn't this all mean that your socket activated server will never get terminated, if there has been even a single hijacked connection.

If you want, I can add a counter for hijacked connections to this patch - it will keep incrementing, until someone adds proper tracking.

mheon:
I think we should probably do what you're suggesting here and remove them from active connections, but ideally we'd transition them to some sort of separate tracker at that point to make sure we know they're alive.

skorhone:
I'll see, if there's something I can do for those. I can't find any hooks that I could directly attach a listener to for those connections. I think the best option would be to overwrite default close function returned for hijacked connections.

I've prepared code for this in my other pull request. I changed both v2 apis that are using attach to do hijacking using a single function. I think that would be ideal place to do required wiring.

We could:
* Ensure that IdleTracker is passed as a reference to all http requests
* Remove http connection registration on StateHijacked and increment hijacked connection count by one
* Create a wrapper for connection
  * Add a reference to IdleTracker (received from http request)
  * Override close method to call IdleTracker (once and only once)

mheon:
@jwhonce PTAL

skorhone:
This request is included also in #6918

jwhonce:
Merged #6918

PR #6916, opened 2020-07-09
log API: add context to allow for cancelling

vrothberg:
Add a `context.Context` to the log APIs to allow for cancelling
streaming (e.g., via `podman logs -f`).  This fixes issues for
the remote API where some go routines of the server will continue
writing and produce nothing but heat and waste CPU cycles.

### DIFF
--- a/cmd/podman/containers/logs.go
+++ b/cmd/podman/containers/logs.go
@@ -32,6 +32,8 @@ var (
 		Long:  logsDescription,
 		Args: func(cmd *cobra.Command, args []string) error {
 			switch {
+			case registry.IsRemote() && logsOptions.Latest:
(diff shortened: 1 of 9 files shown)


vrothberg:
Marking as WIP as I remote tests are disabled, so I want to add a couple more commits to this PR so we can actually test what I produced :^)

rhatdan:
@vrothberg Why are remote tests disabled?  We enabled them last week?

rhatdan:
LGTM

vrothberg:
> @vrothberg Why are remote tests disabled? We enabled them last week?

I was referring to some of the log tests which were disabled for remote.

rhatdan:
@ashley-cui @QiWang19 @TomSweeneyRedHat @baude @giuseppe @nalind PTAL

PR #6918, opened 2020-07-09
Fix: Correct connection counters for hijacked connections

skorhone:
This patch fixes connection counters for v2 endpoints
    
Idletracker was moved to a new package to prevent package cycle. Hijacking code still remains in wrong place and should be moved later to isolated package

This pull request is based on two other pull requests that are already pending (#6913 and #6914). Feel free to close those requests, if you want everything as a single patch
    
### DIFF
--- a/pkg/api/server/handler_api.go
+++ b/pkg/api/server/handler_api.go
@@ -37,6 +37,7 @@ func (s *APIServer) APIHandler(h http.HandlerFunc) http.HandlerFunc {
 			c := context.WithValue(r.Context(), "decoder", s.Decoder) //nolint
 			c = context.WithValue(c, "runtime", s.Runtime)            //nolint
 			c = context.WithValue(c, "shutdownFunc", s.Shutdown)      //nolint
+			c = context.WithValue(c, "idletracker", s.idleTracker)    //nolint
 			r = r.WithContext(c)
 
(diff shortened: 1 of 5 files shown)


pkg/api/server/idletracker/idletracker.go
```diff
@@ -0,0 +1,74 @@
+package idletracker
```
skorhone, line 1: New package was required to break dependency cycle. I think it makes sense to have this tracker in isolated package

pkg/api/server/handler_api.go
```diff
@@ -37,6 +37,7 @@ func (s *APIServer) APIHandler(h http.HandlerFunc) http.HandlerFunc {
 			c := context.WithValue(r.Context(), "decoder", s.Decoder) //nolint
 			c = context.WithValue(c, "runtime", s.Runtime)            //nolint
 			c = context.WithValue(c, "shutdownFunc", s.Shutdown)      //nolint
+			c = context.WithValue(c, "idletracker", s.idleTracker)    //nolint
```
skorhone, line 40: Added idletracker to context so that it can be accessed during hijacking

skorhone:
Rebased this and two others + corrected history

jwhonce:
LGTM

Issue #6927, opened 2020-07-10
Not able to run k8s node e2e tests 


harche:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
Kubernetes [node e2e tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/e2e-node-tests.md) execute `docker info --format '{{json .}}'` command before starting with the tests. I see the support was added for `GO template {{ json . }}` in podman a while back in https://github.com/containers/podman/pull/2688 but I am still seeing the issue. 

I am trying to execute node e2e tests on Fedora 32 with podman (via podman-docker.norach)   

**Steps to reproduce the issue:**

1. clone github.com/kubernetes/kubernetes

2. make test-e2e-node

**Describe the results you received:**

```sh
CGROUPS_PIDS: enabled
Validating package...
F0710 11:20:35.434641   90024 e2e_node_suite_test.go:151] system validation failed: [failed executing "docker info --format '{{json .}}'"\noutput: Error: template: info:1: function "json" not defined
\nerror: exit status 125]
```

```sh
# docker info --format '{{json .}}'
Error: template: info:1: function "json" not defined
```

**Describe the results you expected:**
 `docker info --format '{{json .}}'` should work to make sure we are able to execute k8s e2e tests. 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
# podman version 
Version:      2.0.2
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64

```
Also, what's up with that `Built` date? 
**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.18-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.18, commit: 6e8799f576f11f902cd8a8d8b45b2b2caf636a85'
  cpus: 32
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: harpatil-node-tests
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.6.6-300.fc32.x86_64
  linkmode: dynamic
  memFree: 120888156160
  memTotal: 126625615872
  ociRuntime:
    name: crun
    package: crun-0.14-2.fc32.x86_64
[...]

PR #6929, opened 2020-07-10
version/info: format: allow more json variants

vrothberg:
Allow more variants to yield json output for `podman version` and
`podman info`.  Instead of comparing strings, use a regex and add
unit and e2e tests.

Fixes: #6927

### DIFF
--- a/cmd/podman/parse/json.go
+++ b/cmd/podman/parse/json.go
@@ -1,0 +1,9 @@
+package parse
+
+import "regexp"
+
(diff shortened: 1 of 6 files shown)


vrothberg:
@baude, PTAL. Same behaviour as Docker now although `docker info --format json` yiels `json`.

mheon:
LGTM

QiWang19:
LGTM

rhatdan:
Not crazy about the cut and pasting of all the forms of json in three places, but
/lgtm

PR #6932, opened 2020-07-10
Don't setup AppArmor provile for privileged pods

rhafer:
This is essentially db218e7162c2 forward-ported to specgen


mheon:
LGTM
/approve
@rhatdan @baude @vrothberg PTAL

saschagrunert:
Feel free to lift the hold if we wanna do the test as follow-up.

rhatdan:
Let's do the test in a separate PR.

Just create a --privileged container and then inspect it to see if AppArmor is enabled, should be enough.
/hold cancel

vrothberg:
Yes, please add tests. I really don't like merging PRs without tests. We have proven to regress on untested code.

saschagrunert:
> Yes, please add tests. I really don't like merging PRs without tests. We have proven to regress on untested code.

No worries, I spoke with @rhafer about this PR some hours ago and we will think of some (at least basic) e2e tests in conjunction with AppArmor.

PR #6952, opened 2020-07-13
add systemd to fedora image

baude:
fedora removed the systemd package from its standard container image causing our systemd pid1 test to fail.

thanks to dwalsh for fix!

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -392,7 +392,6 @@ install_test_configs() {
     # as the default).  This config prevents allocation of network address space used
     # by default in google cloud.  https://cloud.google.com/vpc/docs/vpc#ip-ranges
     install -v -D -m 644 $SCRIPT_BASE/99-do-not-use-google-subnets.conflist /etc/cni/net.d/
-    install -v -D -m 644 ./test/policy.json /etc/containers/
     install -v -D -m 644 ./test/registries.conf /etc/containers/
 }
 
--- a/test/e2e/config_amd64.go
+++ b/test/e2e/config_amd64.go
@@ -5,9 +5,11 @@ var (
 	STORAGE_OPTIONS          = "--storage-driver vfs"
 	ROOTLESS_STORAGE_FS      = "vfs"
 	ROOTLESS_STORAGE_OPTIONS = "--storage-driver vfs"
-	CACHE_IMAGES             = []string{ALPINE, BB, fedoraMinimal, nginx, redis, registry, infra, labels, healthcheck}
+	CACHE_IMAGES             = []string{ALPINE, BB, fedoraMinimal, nginx, redis, registry, infra, labels, healthcheck, ubi_init, ubi_minimal}
 	nginx                    = "quay.io/libpod/alpine_nginx:latest"
 	BB_GLIBC                 = "docker.io/library/busybox:glibc"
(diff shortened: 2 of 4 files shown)


mheon:
Should we cache a different image that already includes systemd to avoid
hitting the Fedora repos? Maybe ubi8-init?

mheon:
LGTM.

mheon:
@rhatdan @vrothberg @giuseppe PTAL and merge

vincent-olivert-riera:
@baude , you commit summary seems a little bit misleading: `add system to fedora image`

You are not adding `systemd` (I guess you made a typo) to the `fedora` image. What you are actually doing is replacing `fedora` by `ubi8-init` as the image you use for the "systemd PID1" test.

Perhaps a commit summary like this would be more appropriate?

`test/e2e/systemd_test.go: replace fedora image with ubi8-init`

edsantiago:
```
[+0287s] Caching registry.access.redhat.com/ubi8-init at /tmp/ubi8-init.tar...Running: /usr/bin/podman --root /tmp/crio --runroot /tmp/crio-run --runtime /usr/bin/crun --conmon /usr/libexec/podman/conmon --cni-config-dir /etc/cni/net.d --cgroup-manager cgroupfs --tmpdir /tmp --events-backend file --storage-driver=vfs pull registry.access.redhat.com/ubi8-init
[+0287s] Trying to pull registry.access.redhat.com/ubi8-init...
[+0287s]   A signature was required, but no signature exists
[+0287s] Error: unable to pull registry.access.redhat.com/ubi8-init: Source image rejected: A signature was required, but no signature exists
```
As best I can tell, this is happening because there's a mysterious magical invisible `--signature-policy test/policy.json` in the `podman pull` command. The only possible place I see this happening is https://github.com/containers/podman/blob/master/test/e2e/common_test.go#L266 but although I see that being set, I don't see it being actually referenced anywhere.

rhatdan:
Why was policy.json removed.  I updated my PR with this code but added back the policy.json change, and was able to pass the test locally.

baude:
@rhatdan that policy,json gets added to each system in CI.  I am not surprised that it passed locally at all.  With that policy.json though, you cannot pull anything from registry.access because it expects a signature and ... then are not signed or least not with a usable as-is signature.

rhatdan:
Removing the signature falls back to Reject all though.

vrothberg:
Continuing over here: https://github.com/containers/podman/pull/6966

edsantiago:
The story so far: what we need to do is remove this line: https://github.com/containers/podman/blob/3d3359086ef0ed0833a8ac8e9d06d9934d835139/contrib/gate/Dockerfile#L30 and then (waves hands) rebuild the images and push them onto quay.

vrothberg:
> The story so far: what we need to do is remove this line:
> 
> https://github.com/containers/podman/blob/3d3359086ef0ed0833a8ac8e9d06d9934d835139/contrib/gate/Dockerfile#L30
> and then (waves hands) rebuild the images and push them onto quay.

Currently pushing an image including this change. I'll restart the `in-podman*` PRs once the image is on Quay.

edsantiago:
Woot!

For posterity: my comment above (re: gating Dockerfile) was incorrect. That should still be done, but it wasn't the cause of the CI failure. The correct solution, found by @vrothberg and @baude, was to manually rebuild and push the `in_podman:master` image.

PR #5344, opened 2020-02-27
CI: format cirrus logs

edsantiago:
This introduces a new cirrus helper script, logformatter.
Usage is:

        [commands...] | logformatter TEST-NAME

It reformats its input into a readable, highlighed, linkable
form. Some features:

   - boring stuff (timestamps, standard podman options) is
     deemphasized
   - important stuff (warnings, errors) is emphasized
   - in-page links to the actual failures
   - active links to source files
   - jumps to bottom of page on load, because that's where
     the errors are. (All errors are linked)

Add it to select test commands (integration, system) and
add a new artifacts_html, run in the 'always' block, which
uploads generated *.log.html into Cirrus; from there we
generate a live URL that can be viewed in browser.

Unfortunately, due to security concerns in Cirrus, it is
not currently possible to make the link a live one.

Kludge: add a line of dashes after Restoring images; without this,
the first test ("systemd PID 1") has no dashes before it, so
logformatter doesn't see it.

Signed-off-by: Ed Santiago <santiago@redhat.com>

PR #6955, opened 2020-07-13
logformatter: update MAGIC BLOB string

edsantiago:
Fallout from libpod->podman repo name move: the HTML logs
created by logformatter are no longer accessible. They
render as:

    https://storage.googleapis.com/SECRET-5385732420009984-fcae48/artifacts/containers/podman/6313596734930944/html/integration_test.log.html

(yes, "SECRET" instead of "cirrus-ci". I have no idea what
that's about, but that's not the focus of this PR, and I'm
going to ignore it for the time being and hope it goes away).

The focus of this PR is to change "podman" in the URL above
back to "libpod", because that's what the Google storage
bucket is called. This is unfortunate hardcoding, because
the way the code was written it used $CIRRUS_REPO_FULL_NAME
as defined by Cirrus.

I have no idea if this will work. It's possible that instead
of this, the proper change is to update the magic blob string.
Chris is OOTO this week, so I'll just try random fixes in
hopes of getting something to work.

### DIFF
--- a/contrib/cirrus/logformatter
+++ b/contrib/cirrus/logformatter
@@ -418,10 +418,27 @@ END_HTML
         }
     }
 
-    # FIXME: if Cirrus magic envariables are available, write a link to results
+    # If Cirrus magic envariables are available, write a link to results.
+    # FIXME: it'd be so nice to make this a clickable live link.
+    #
+    # STATIC_MAGIC_BLOB is the name of a google-storage bucket. It is
+    # unlikely to change often, but if it does you will suddenly start
+    # seeing errors when trying to view formatted logs:
+    #
+    #    AccessDeniedAccess denied.Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object.
+    #
+    # This happened in July 2020 when github.com/containers/libpod was
+    # renamed to podman. If something like that ever happens again, you
+    # will need to get the new magic blob value from:
+    #
+    #   https://console.cloud.google.com/storage/browser?project=libpod-218412
+    #
+    # You will also probably need to set the bucket Public by clicking on
+    # the bucket name, then the Permissions tab. This is safe, since this
+    # project is fully open-source.
     if ($have_formatted_log && $ENV{CIRRUS_TASK_ID}) {
         my $URL_BASE          = "https://storage.googleapis.com";
-        my $STATIC_MAGIC_BLOB = "cirrus-ci-5385732420009984-fcae48";
+        my $STATIC_MAGIC_BLOB = "cirrus-ci-6707778565701632-fcae48";
         my $ARTIFACT_NAME     = "html";
 
         my $URL = "${URL_BASE}/${STATIC_MAGIC_BLOB}/artifacts/$ENV{CIRRUS_REPO_FULL_NAME}/$ENV{CIRRUS_TASK_ID}/${ARTIFACT_NAME}/${outfile}";
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -71,7 +71,7 @@ env:
     # Service-account client_email - needed to build images
     SERVICE_ACCOUNT: ENCRYPTED[702a8e07e27a6faf7988fcddcc068c2ef2bb182a5aa671f5ccb7fbbfb891c823aa4a7856fb17240766845dbd68bd3f90]
     # Service account username part of client_email - for ssh'ing into VMs
-    GCE_SSH_USERNAME: ENCRYPTED[d579f2d3000bb678c9af37c3615e92bcf3726e9afc47748c129cef23ee799faaafd4baba64048329205d162069d90060]
+    GCE_SSH_USERNAME: 'cirrus-ci'
 
 # Default VM to use unless set or modified by task
 gce_instance:


mheon:
LGTM

edsantiago:
Well, aside from the fact that it had no effect, but thanks.

It looks like I may have to change the "magic blob" string. The source of that is [this comment by cevich](https://github.com/containers/podman/pull/5344#issuecomment-592680801) on #5344, in particular:

> IIRC, the `STATIC_MAGIC_BLOB` is formed from some combination of an internal github ID and the google project number.  I/We could do more research to expose the magic better if needed.

I've looked everywhere I can think of, and cannot find any documentation on how this magic string is generated nor what its new value may be. I've looked in the source tree, in github history, IRC logs, mail logs, and even the googleapi web UI.

I'm afraid this may be unsolvable until @cevich returns.

edsantiago:
/hold

@cevich  I need help again please. Something somewhere is filtering `cirrus-ci` and replacing it with `SECRET`:
```
https://storage.googleapis.com/SECRET-6707778565701632-fcae48/artifacts/containers/podman/5595074811658240/html/integration_test.log.html
```
(Source: [sample cirrus log](https://cirrus-ci.com/task/5595074811658240?command=integration_test#L8354), happening in both integration and system test logs.)

If I copy that URL, manually change `SECRET` to `cirrus-ci`, request fails with:
```
AccessDeniedAccess denied.Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object.
```
The latter could maybe be related to the `Public access` setting [here](https://console.cloud.google.com/storage/browser?project=libpod-218412): the new blob says "Subject to object ACLs", whereas the old bucket said "Public to internet". If you agree that this is the difference, could you make the magic necessary change please?

But then that still leaves us with the SECRET thing. Will that be fixed if the ACL issue is fixed? Or is it something else?

cevich:
Yes I think you're right about the access problem, I have an idea the other thing is related to my copy-paste encrypting all our "secret" values (due to project rename).  I believe one of them is literally "cirruc-ci" and that's why it's being masked...

cevich:
...https://github.com/containers/podman/blob/master/.cirrus.yml#L74 try that one, kill the encrypted string and replace with 'cirruc-ci'.  That will either break horribly or fix the `SECRET` problem :grin:   I'll see about fixing the bucket permissions right now...

edsantiago:
Thank you! PR updated and pushed.

cevich:
...okay, access is fixed and I disabled the old bucket, made it read-only.  There's no way to change the bucket name or add a description, otherwise I would have done that as well.  Okay now, where to document these arcane incantations.

contrib/cirrus/logformatter
```diff
@@ -418,10 +418,13 @@ END_HTML
         }
     }
 
-    # FIXME: if Cirrus magic envariables are available, write a link to results
+    # If Cirrus magic envariables are available, write a link to results.
+    # FIXME: it'd be so nice to make this a clickable live link.
+    # Magic Blob can be obtained from:
+    #   https://console.cloud.google.com/storage/browser?project=libpod-218412
```
cevich, line 424: How about we just document this here.  Something like:

```
    # Note: The value for STATIC_MAGIC_BLOB is the name of a google-storage bucket.
    # They can be obtained from: https://console.cloud.google.com/storage/browser?project=libpod-218412
    # The bucket is created on first-use by Cirrus-CI when setting up a new repository, and will remain
    # for the life of the project.  Bucket names must be globally-unique, so Cirrus-CI appends it's internal
    # ID for the repository and a unique nonce (maybe the github project ID?).
    # By default, these buckets are set private.  However since the contents
    # are public (due to this being an OSS repository), it's fine to make the bucket
    # public.  This is done by adding an entry via the google-storage WebUI:
    # https://console.cloud.google.com/storage/browser/<STATIC_MAGIC_BLOB>?project=<PROJECT_ID>
    # under the 'permissions' tab, for example:
    #     Members: 'allUsers'; Role(s): 'Storage Object Viewer'.
```

edsantiago: Thanks. If tests pass, I'll update the comment and push again.

edsantiago:
WHEW! Cirrus logs now display working links at the bottom. (Sample [integration](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5110197549006848/html/integration_test.log.html), [system](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5110197549006848/html/system_test.log.html)). Thank you @cevich for your help. This was a double whammy(with the coincident SECRET problem) and I couldn't possibly have gotten anywhere without your experience.

/hold cancel

cevich:
Happy to help, and wishing I could have documented some of these things better...while putting out fires from the renaming...oof :confounded: In hindsight, at least things turned out to be failing a logical way (for me).  I'll those any day, over unpredictable/random type problems :grinning:

edsantiago:
From my perspective, this is ready to merge. To fully confirm that it works, click on any of the CI results that include system or integration tests, then expand that log, go to the bottom, and confirm (copy-paste unfortunately) that the formatted-results log works and displays the correct, readable log results.

PR #6956, opened 2020-07-13
Include infra container information in `pod inspect`

mheon:
We had a field for this in the inspect data, but it was never being populated. Because of this, `podman pod inspect` stopped showing port bindings (and other infra container settings). Add code to populate the infra container inspect data, and add a test to ensure we don't regress again.


sshnaidm:
thanks!

PR #6957, opened 2020-07-13
Mask out /sys/dev to prevent information leak from the host

rhatdan:


rhatdan:
@mrunalp PTAL

giuseppe:
/retest

LGTM

TristanCacqueray:
It seems like this is causing trouble for graphical applications such as `glxgears` that used to work when running inside a podman container started with `--device /dri`. Is there an option to keep `/sys/dev` in the container?

rhatdan:
@TristanCacqueray Please open an issue for this?  So that we can discuss it there.

PR #6959, opened 2020-07-13
[CI:DOCS] Remove outdated seccomp policy

mheon:
Some time ago, we moved the Seccomp policy (and related setup code) to a place where all our tools could share it [1]. We did not, however, remove the in-repo seccomp.json file. Over the last year or so, the in-repo seccomp policy has become progressively more and more outdated, with no effort made to maintain it (because what sense is there in keeping a duplicate?). Today, a friend came to me and asked if a Podman container could access keyctl, assuming it could not because he was reading the outdated Seccomp policy which does not allow it. Since it's becoming clear that this file is doing no good and actively causing confusion, let's just drop it.

[1] https://github.com/seccomp/containers-golang


mheon:
@lsm5 Poke - does this look like it will break packaging anywhere?

lsm5:
> @lsm5 Poke - does this look like it will break packaging anywhere?

It won't. We're using a separate seccomp.json in containers-common subpackage.

mheon:
@rhatdan @vrothberg @baude PTAL

PR #6324, opened 2020-05-21
allow switching of port-forward approaches when rootless/using slirp4netns

aleks-mariusz:
This PR is [by request](https://github.com/containers/libpod/pull/6025#issuecomment-631316276) against master (the other [original PR](https://github.com/containers/libpod/pull/6025) was against the 1.9 branch as that's what i was asked to base it off of).

---

As of podman 1.8.0, because of commit da7595a, the default approach of providing port-forwarding in rootless mode has switched (and been hard-coded) to rootlessport, for the purpose of providing super performance.  The side-effect of this switch is source within the container to the port-forwarded service always appears to originate from 127.0.0.1 (see issue #5138).

This commit allows a user to specify if they want to revert to the previous approach
of leveraging slirp4netns add_hostfwd() api which, although not as stellar performance,
restores usefulness of seeing incoming traffic origin IP addresses.

The change should be transparent; when not specified, rootlessport will continue to be
used, however if specifying `--net slirp4netns:slirplisten` the old approach will be used.

Testing performed in rootless mode of relevant functionality:

<details><summary>Default behaviour scenario 1 (nothing specified, so not taking advantage of this commit original behaviour (higher performance rootlessport in use))</summary>

```
[test@test ~]$ podman run -d --name nginx -p 8080:80 docker.io/library/nginx:alpine
c0644bcd50c950a000305d0e548e3b1c40732af5c1f9bcbe1dffba244fc1a048
[test@test ~]$ podman logs -f nginx
127.0.0.1 - - [28/Apr/2020:18:19:47 +0000] "GET / HTTP/1.1" 304 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.122 Safari/537.36" "-"
```
</details>
<details><summary>Default behaviour scenario 2 (specifying just --net slirp4netns, which is default when in rootless mode)</summary>

```
[test@test ~]$ podman run -d --name nginx -p 8080:80 --net slirp4netns docker.io/library/nginx:alpine
82e93b571cb013b05f586097544c081c78335f932a5fae071d27ca0f24014cf0
[test@test ~]$ podman logs -f nginx
[...]

Issue #6912, opened 2020-07-09
rootless: support launching slirp4netns without --disable-host-loopback


AkihiroSuda:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Support launching slirp4netns without `--disable-host-loopback`, so that the localhost ports on the host can be accessed as 10.0.2.2 from containers.

For security, the default MUST remain to use --disable-host-loopback.

Originally requested by @tmds https://github.com/containers/podman/issues/1733#issuecomment-655969836

Corresponding code: https://github.com/containers/podman/blob/9eac75a967e3459b32b1acd220afb49b60f5e5aa/libpod/networking_linux.go#L218-L220

rhatdan:
Sure, what would you suggest the CLI for this should be?

tmds:
suggestions: `--enable-host-loopback`/`--add-host-loopback`.

mheon:
I'm still of the opinion we need a generic network options CLI - `--network-opt slirp4netns:enable-host-loopback,disable-seccomp` or similar

tmds:
It would be nice to be able to pass options to `slirp4netns` from the command line.

First class features should also be exposed through direct arguments, without referencing `slirp4netns`.

giuseppe:
let's add new flags for slirp4netns using: https://github.com/containers/podman/pull/6965

giuseppe:
I've added a patch to #6963 

@AkihiroSuda will this help with rootless kind?

AkihiroSuda:
I guess you meant #6965?

> rootless kind

I'm not sure. At least kind with rootless Docker doesn't seem to require this.

giuseppe:
Yes I think the main issue is that with podman containers cannot communicate with each other when running rootless

tmds:
@AkihiroSuda @giuseppe I'm not sure how to use this.

I have a webserver running on localhost:
```
$ curl http://localhost:5000
Hello World!
```

I create a container that is on a network. I'd like the container to be able to the localhost webserver.
```
$ podman network create mynet
$ podman run -ti --network mynet busybox /bin/sh
/ # wget http://10.0.2.2:5000
Connecting to 10.0.2.2:5000 (10.0.2.2:5000)
wget: can't connect to remote host (10.0.2.2): Network is unreachable
```

What argument should I add, and to which command?

AkihiroSuda:
[...]

PR #6965, opened 2020-07-14
allow switching of port-forward approaches in rootless/using slirp4netns

giuseppe:
As of podman 1.8.0, because of commit da7595a, the default approach of providing
port-forwarding in rootless mode has switched (and been hard-coded) to rootlessport,
for the purpose of providing super performance. The side-effect of this switch is
source within the container to the port-forwarded service always appears to originate
from 127.0.0.1 (see issue #5138).

This commit allows a user to specify if they want to revert to the previous approach
of leveraging slirp4netns add_hostfwd() api which, although not as stellar performance,
restores usefulness of seeing incoming traffic origin IP addresses.

The change should be transparent; when not specified, rootlessport will continue to be
used, however if specifying --net slirp4netns:slirplisten the old approach will be used.

Note: the above may imply the restored port-forwarding via slirp4netns is not as
performant as the new rootlessport approach, however the figures shared in the original
commit that introduced rootlessport are as follows:
slirp4netns: 8.3 Gbps,
RootlessKit: 27.3 Gbps,
which are more than sufficient for many use cases where the origin of traffic is more
important than limits that cannot be reached due to bottlenecks elsewhere.

Signed-off-by: Aleks Mariusz <m.k@alek.cx>

Follow up for https://github.com/containers/podman/pull/6324

### DIFF
--- a/cmd/podman/common/netflags.go
+++ b/cmd/podman/common/netflags.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"net"
+	"strings"
 
 	"github.com/containers/libpod/v2/cmd/podman/parse"
 	"github.com/containers/libpod/v2/libpod/define"
@@ -164,11 +165,18 @@ func NetFlagsToNetOptions(cmd *cobra.Command) (*entities.NetOptions, error) {
 			return nil, err
 		}
 
+		parts := strings.SplitN(network, ":", 2)
+
 		ns, cniNets, err := specgen.ParseNetworkNamespace(network)
 		if err != nil {
 			return nil, err
 		}
 
+		if len(parts) > 1 {
+			opts.NetworkOptions = make(map[string][]string)
+			opts.NetworkOptions[parts[0]] = strings.Split(parts[1], ",")
+			cniNets = nil
+		}
 		opts.Network = ns
 		opts.CNINetworks = cniNets
 	}
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -553,7 +553,10 @@ Valid values are:
 - `<network-name>|<network-id>`: connect to a user-defined network, multiple networks should be comma separated
 - `ns:<path>`: path to a network namespace to join
 - `private`: create a new namespace for the container (default)
-- `slirp4netns`: use slirp4netns to create a user network stack.  This is the default for rootless containers
+- `slirp4netns[:OPTIONS,...]`: use slirp4netns to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
+  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
+  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
+  **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
 
 **--network-alias**=*alias*
 
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -561,7 +561,10 @@ Valid _mode_ values are:
 - _network-id_: connect to a user-defined network, multiple networks should be comma separated;
 - **ns:**_path_: path to a network namespace to join;
 - `private`: create a new namespace for the container (default)
-- **slirp4netns**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.
+- **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
+  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
+  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
+  **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
 
 **--network-alias**=*alias*
 
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -417,6 +417,7 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	s.DNSOptions = c.Net.DNSOptions
 	s.StaticIP = c.Net.StaticIP
 	s.StaticMAC = c.Net.StaticMAC
+	s.NetworkOptions = c.Net.NetworkOptions
 	s.UseImageHosts = c.Net.NoHosts
 
 	s.ImageVolumeMode = c.ImageVolume
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -341,6 +341,8 @@ type ContainerConfig struct {
 	Networks []string `json:"networks,omitempty"`
 	// Network mode specified for the default network.
 	NetMode namespaces.NetworkMode `json:"networkMode,omitempty"`
+	// NetworkOptions are additional options for each network
+	NetworkOptions map[string][]string `json:"network_options,omitempty"`
 
 	// Image Config
 
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -173,6 +173,19 @@ type slirpFeatures struct {
 	HasEnableSeccomp       bool
 }
 
+type slirp4netnsCmdArg struct {
+	Proto     string `json:"proto,omitempty"`
+	HostAddr  string `json:"host_addr"`
+	HostPort  int32  `json:"host_port"`
+	GuestAddr string `json:"guest_addr"`
+	GuestPort int32  `json:"guest_port"`
+}
+
+type slirp4netnsCmd struct {
+	Execute string            `json:"execute"`
+	Args    slirp4netnsCmdArg `json:"arguments"`
+}
+
 func checkSlirpFlags(path string) (*slirpFeatures, error) {
 	cmd := exec.Command(path, "--help")
 	out, err := cmd.CombinedOutput()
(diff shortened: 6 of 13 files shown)


aleks-mariusz:
Hi @giuseppe to answer the question from the other thread (where i thought @AkihiroSuda was waiting for a @mheon as he last posted the review), i am interested in this feature still, however i'd be more interested if this was config file driven rather than command line parameter.. someone offered to look into doing this after the command line support was in, as it's beyond my famliarity with the code base (not to mention this is the first golang code i've ever touched).

libpod/container_validate.go
```diff
@@ -70,7 +70,7 @@ func (c *Container) validate() error {
 
 	// Rootless has some requirements, compared to networks.
 	if rootless.IsRootless() {
-		if len(c.config.Networks) > 0 {
+		if !c.config.NetMode.IsSlirp4netns() {
```
mheon, line 73: I'm still against merging with this here - there's probably something on the frontend that's erroneously detecting `slirp4netns:rootlesskit` as a CNI network and adding it.

aleks-mariusz: understand, and i'm open to ideas what/where else may need to be adjusted to not need this "enabling" snippet in container_validate.go.. 

i don't however quite understand what is wrong with this logic, why would the lenght of networks given ever be a non-empty array, networks shouldn't be able to be specified.. the fixing of the erroneous detection of that parameter seems like a separate bug that needs to be fixed/handled in a different PR, no ?

giuseppe: @mheon sorry my mistake.  I had fixed the detection but forgot to switch the condition here.  Pushed a new version with `if len(c.config.Networks) > 0`.

I am currently experimenting running CNI plugins inside of the rootless environment.  So I cannot promise this line will stay around for long :-)

mheon: Fair enough - I'd be extremely interested to see proper rootless CNI!

giuseppe: it will still be limited.  We first need to create a network namespace managed by slirp, and from that network namespace create new ones using CNI.  I am still unsure how to handle its UX nicely

giuseppe:
tests are green

mheon:
I'm still a little iffy on the throwing options in the namespace itself. We have other requests to allow setting arbitrary options on networks (https://github.com/containers/podman/issues/6912) - what would you think about having a separate entry (`map[string][]string` - map of network name (slirp4netns, bridge, etc) to options) in the database for network options? We can keep the existing CLI, but split on the `:` and anything after that is split on `,` and turns into a network option?

aleks-mariusz:
pardon if i'm overstepping my place here, but the direction of some of the comments above to me seems to have some feature-creep potential..  the original goal of this PR was mainly to allow the source IP to be preserved, and while i of course have no issue with trying to do that in the cleanest way possible, it seems that some of the above discussions are starting to border on what i'd consider being better addressed/handled in a separate PR :-) which is all fine and good, but i make no promises i'll be able to assist (since it's a bit unclear who these suggestions are with) and i'd need to respectfully put my hand back down if there's an implied call for volunteers for these further changes..

giuseppe:
@mheon is it something we can add to `NetOptions`?

mheon:
@giuseppe That would work for me - I basically want a way to handle the next request after this one without duplicating all the plumbing.

giuseppe:
> @giuseppe That would work for me - I basically want a way to handle the next request after this one without duplicating all the plumbing.

thanks.  That was indeed a great idea.  It looks much cleaner and it was easy to add a new option on top of it.

docs/source/markdown/podman-create.1.md
```diff
@@ -553,7 +553,10 @@ Valid values are:
 - `<network-name>|<network-id>`: connect to a user-defined network, multiple networks should be comma separated
 - `ns:<path>`: path to a network namespace to join
 - `private`: create a new namespace for the container (default)
-- `slirp4netns`: use slirp4netns to create a user network stack.  This is the default for rootless containers
+- `slirp4netns[:OPTIONS,...]`: use slirp4netns to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
+  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
+  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
+  **allow_host_loopback=slirp4netns**: Allow the slirp4netns to reach the host localhost.
```
AkihiroSuda, line 559: **allow_host_loopback**=*true|false*

docs/source/markdown/podman-run.1.md
```diff
@@ -561,7 +561,10 @@ Valid _mode_ values are:
 - _network-id_: connect to a user-defined network, multiple networks should be comma separated;
 - **ns:**_path_: path to a network namespace to join;
 - `private`: create a new namespace for the container (default)
-- **slirp4netns**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.
+- **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
+  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
+  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
+  **allow_host_loopback=slirp4netns**: Allow the slirp4netns to reach the host localhost.
```
AkihiroSuda, line 567: `**allow_host_loopback**=*true|false*`

AkihiroSuda:
LGTM but a nit in the man page. Thanks!

docs/source/markdown/podman-create.1.md
giuseppe: thanks, fixed now.

docs/source/markdown/podman-run.1.md
```diff
@@ -561,7 +561,10 @@ Valid _mode_ values are:
 - _network-id_: connect to a user-defined network, multiple networks should be comma separated;
 - **ns:**_path_: path to a network namespace to join;
 - `private`: create a new namespace for the container (default)
-- **slirp4netns**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.
+- **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
+  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
+  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
+  **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host localhost.  Default to false.
```
AkihiroSuda, line 567: Maybe better to document the host loopback IP (10.0.2.2). Can be a follow-up PR.

giuseppe: thanks, fixed now.  Tests were still broken, so I needed to fix the PR anyway

rhatdan:
@mheon PTAL

cmd/podman/common/netflags.go
```diff
@@ -164,11 +165,17 @@ func NetFlagsToNetOptions(cmd *cobra.Command) (*entities.NetOptions, error) {
 			return nil, err
 		}
 
+		parts := strings.SplitN(network, ":", 2)
```
mheon, line 168: Recommendation: If length on this is 2, we should probably throw away `cniNets` below (options should only be given if you're not specifying more than 1 seems reasonable)

mheon:
One comment, and it would be nice to have a test; otherwise LGTM

cmd/podman/common/netflags.go
giuseppe: fixed

giuseppe:
> One comment, and it would be nice to have a test; otherwise LGTM

took care of the comment and added tests

TomSweeneyRedHat:
LGTM
and all happy green test buttons.

TomSweeneyRedHat:
Or they were green until @giuseppe added the latest at the same time as my comment.

mheon:
/hold
One flake

giuseppe:
tests are green now

/hold cancel

aleks-mariusz:
@giuseppe thanks so much for driving this forward and finally getting it all preped, perfected, and merged!

quick follow up question, did this implement config-file handling as well? if so what params would need to go into which conf file to fix the original source-ip=127.0.0.1 issue that spawned this?

giuseppe:
> quick follow up question, did this implement config-file handling as well? if so what params would need to go into which conf file to fix the original source-ip=127.0.0.1 issue that spawned this?

no, that part is not implemented yet, it is possible to specify the new option only through the CLI

mheon:
Yes - we'll need a separate field for this in containers.conf

storrgie:
Did this make it into 2.0.3?

mheon:
Negative - 2.0.3 is bugfixes only, this is a new feature.

I think we'll probably start cutting 2.1.0 release candidates soon, though - this week or next.

DavidePrincipi:
Thanks for this PR! 

The `jvb` component of Jitsi is an UDP server that needs IP addresses are preserved. 

On podman 2.2.1 I can run

    $ podman pod create --network=slirp4netns:port_handler=slirp4netns --publish 2086:80 --publish 10000:10000/udp --name=jitsi ...

https://gist.github.com/DavidePrincipi/6a45a7cfaf367115931eff3f50d1a681

aleks-mariusz:
> Negative - 2.0.3 is bugfixes only, this is a new feature.
> 
> I think we'll probably start cutting 2.1.0 release candidates soon, though - this week or next.

@mheon does this imply the conf support is available as well? if so can you please point to the param needed and which file it needs to go into? this is a follow up to my ask in July to @giuseppe

mheon:
The `netns` field in `containers.conf` should be what you want - setting that to `slirp4netns:port_handler=slirp4netns` should enable this by default

barosl:
> The `netns` field in `containers.conf` should be what you want - setting that to `slirp4netns:port_handler=slirp4netns` should enable this by default

This doesn't seem to work. I tried the following:

```toml
[containers]
netns = 'slirp4netns:port_handler=slirp4netns'
```

rktjmp:
> > The `netns` field in `containers.conf` should be what you want - setting that to `slirp4netns:port_handler=slirp4netns` should enable this by default
> 
> This doesn't seem to work. I tried the following:
> 
> ```toml
> [containers]
> netns = 'slirp4netns:port_handler=slirp4netns'
> ```

I believe you want to use

```
[engine]
network_cmd_options=["port_handler=slirp4netns"]
```

Which sets `Default options to pass to the slirp4netns binary.`

The manpage also shows `network_cmd_options["enable_ipv6=true",]`, for the option, which implies to me it's the default value, but the actual description for `enable_ipv6` says the default is false, so... not sure. I guess just be aware that the option exists if something breaks.

Also be aware that unfortunately you cant mix the `port_handler` with user-defined networks, so the option might not apply in your case:

```
• port_handler=slirp4netns:
Use  the  slirp4netns  port forwarding, it is slower than
rootlesskit but preserves the correct source IP address.
This port handler cannot be used for user-defined networks.
```

PR #6966, opened 2020-07-14
Fix systemd pid 1 test

vrothberg:
Fedora removed the systemd package from its standard container image
causing our systemd pid1 test to fail.  Replacing usage of fedora to
ubi-init.

Adding ubi images to the cache for local tests.


test/policy.json
```diff
@@ -10,12 +10,6 @@
                 {
                     "type": "reject"
                 }
-            ],
```
edsantiago, line 13: This isn't going to work. First, the "show trust" test is going to fail because it explicitly needs this line. Second, you're not removing the `install` command from lib.sh, so all `ubi` pulls are going to fail. Third, the `in-podman` images still have the bad `policy.json` so those are going to fail anyway.

Sorry!

vrothberg: Why do we need remove the install command?

edsantiago: Oh, good point, I guess with this change you wouldn't need to. Points (1) and (3) still remain though.

PR #6971, opened 2020-07-14
play-kube: add suport for "IfNotPresent" pull type

TristanCacqueray:
This change prevents this exception when loading a pod spec
using the "IfNotPresent" pull policy:
  Error: invalid pull type "IfNotPresent"

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -555,7 +555,7 @@ func ValidatePullType(pullType string) (PullType, error) {
 	switch pullType {
 	case "always":
 		return PullImageAlways, nil
-	case "missing":
+	case "missing", "IfNotPresent":
 		return PullImageMissing, nil
 	case "never":
 		return PullImageNever, nil


mheon:
LGTM
/approve
@haircommander @baude @rhatdan PTAL

pkg/util/utils.go
```diff
@@ -557,6 +557,8 @@ func ValidatePullType(pullType string) (PullType, error) {
 		return PullImageAlways, nil
 	case "missing":
 		return PullImageMissing, nil
+	case "IfNotPresent":
```
TristanCacqueray, line 560: perhaps pullType should be forced to lower() ?

pkg/util/utils.go
```diff
@@ -557,6 +557,8 @@ func ValidatePullType(pullType string) (PullType, error) {
 		return PullImageAlways, nil
 	case "missing":
```
haircommander, line 558: we can do `case "missing", "IfNotPresent":` here instead

pkg/util/utils.go
rhatdan: Does Kube allow ifnotpresent?

rhatdan:
/approve
LGTM

vrothberg:
Thanks for the fix!

PR #6973, opened 2020-07-14
[CI:DOCS]Do not copy policy.json into gating image

baude:
test/policy.json should not need to be copied into the gating image


edsantiago:
LGTM

PR #6974, opened 2020-07-14
docs: user namespace can't be shared in pods

sshnaidm:
When running "podman pod create --share user" the errors appears:
Error: User sharing functionality not supported on pod level
Fix docs and remove 'user' from shareable parameters.


giuseppe:
I think the issue is that we are not able yet to create a user namespace with `pod create`.

Although, since we are not able to create a new user namespace, it is implicit that the containers are using the same user namespace (a container can still create a child namespace with `--uidmapping` but from the parent PoV that makes no differerence).  I think we can safely drop the error message.

rhatdan:
/approve
/lgtm
Thanks @sshnaidm

PR #6975, opened 2020-07-14
Error on rootless mac and ip addresses

baude:
When creating a pod or container where a static MAC or IP address is provided, we should return a proper error and exit as 125.

Fixes: #6972


mheon:
LGTM

PR #6977, opened 2020-07-14
Preserve passwd on container restart

### LINKED ISSUE #6953: podman exec: Error: unable to find user $USER: no matching entries in passwd file

/kind bug

**Description**

Trying to exec into a container with '-u $USER' fails to look up the user that is automatically added by `--userns=keep-id --user root:root`

in one terminal:

```
$ podman run  --userns=keep-id --user root:root -it --rm --name foof registry.fedoraproject.org/fedora:32 /bin/bash
```

in another terminal:

```
$ podman exec -it -u $USER foof /bin/bash
Error: unable to find user otaylor: no matching entries in passwd file
```

This is a distillation of a problem toolbox triggers; I see this on Fedora Silverblue rawhide, @mheon was able to reproduce on his system

Version:      2.1.0-dev
API Version:  1
Go Version:   go1.14.3
Built:        Wed Dec 31 19:00:00 1969
OS/Arch:      linux/amd64

mheon:
We added code to create a `/etc/passwd` file that we bind-mount into the container in some cases (most notably, `--userns=keep-id` containers). This, unfortunately, was not persistent, so user-added users would be dropped on container restart. Changing where we store the file should fix this.

Further, we want to ensure that lookups of users in the container use the right /etc/passwd if we replaced it. There was already logic to do this, but it only worked for user-added mounts; it's easy enough to alter it to use our mounts as well.

Unfortunately, this is not enough for a full fix. You still cannot run `useradd` within the container because it wants to rename the `/etc/passwd` file (impossible with bind mounts).

Fixes #6953

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -28,7 +28,6 @@ import (
 	securejoin "github.com/cyphar/filepath-securejoin"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
-	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

mheon:
Toolbox also considers the lack of a working `useradd` a breaking change, so this isn't going to be enough.

Won't hurt to merge, though.

rhatdan:
Needs a rebase.

mheon:
Done

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #6835, opened 2020-07-01
fix API: Create container with an invalid configuration

zhangguanzhang:
Signed-off-by: zhangguanzhang <zhangguanzhang@qq.com>
fix https://github.com/containers/libpod/issues/6799
if use API to create containers, this will be null
```
        "Path": "",
        "Args": null,
```

PR #6978, opened 2020-07-14
APIv2 tests: fix race condition causing CI flake

edsantiago:
A newly-added test in #6835 was flaking in CI with:

   not ok 143 [20-containers] DELETE libpod/containers/SHA
   500 cannot remove container <sha> as it is running - running or paused containers cannot be removed without force: container state improper

Root cause: DELETE being run immediately after container start.
Although the container is short-lived, it does take time to
run and exit.

Solution: wait for container to exit (should be quick) before
deleting. This gives us a new test for the /wait endpoint.

Also: tweaked some comments for readability, removed unnecessary
container ps, added actual container status checks, and added
actual message checks to another test that was merely checking
exit status.

### DIFF
--- a/test/apiv2/20-containers.at
+++ b/test/apiv2/20-containers.at
@@ -51,17 +51,19 @@ cid=$(jq -r '.[0].Id' <<<"$output")
 
 t DELETE libpod/containers/$cid 204
 
-# Ensure that API does not occur: Create Container creates an invalid and the container fails to start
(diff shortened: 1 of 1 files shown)


edsantiago:
Sample flake: https://api.cirrus-ci.com/v1/task/4641659449245696/logs/apiv2_test.log

edsantiago:
@baude ptal

PR #6979, opened 2020-07-14
system tests: check for masked-device leaks

edsantiago:
PR #6957 added a new path (/sys/devs) to an existing list
of masked mount points which an unprivileged container
should not be able to access. Here we add a test for
those: run 'stat' on those devices in the container,
and make sure that they are dummies.

This is kind of kludgy, and relies on heuristics that
may not be 100% accurate. It also adds duplication,
a list that must be kept in sync with the original
list in pkg/specgen/generate/config_linux.go.
I'd love to hear suggestions on how to do it better.

### DIFF
--- a/test/system/400-unprivileged-access.bats
+++ b/test/system/400-unprivileged-access.bats
@@ -1,6 +1,7 @@
 #!/usr/bin/env bats   -*- bats -*-
 #
 # Tests #2730 - regular users are not able to read/write container storage
+# Tests #6957 - /sys/dev (et al) are masked from unprivileged containers
 #
 
 load helpers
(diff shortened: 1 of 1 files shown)


edsantiago:
Note to reviewers: I considered adding a test with `--privileged`, to make sure those paths are *not* masked, but decided it wasn't worth the time. If you think it is, please lmk

mheon:
I don't think we really care that the paths are un-masked; I've never seen anyone explicitly relying on that part of privileged. Probably not worth a test.

test/system/400-unprivileged-access.bats
```diff
@@ -97,4 +98,68 @@ EOF
...
+            # FIXME: this is a horrible almost-worthless test! It does not
+            # actually check for files in the directory (expect: zero),
+            # merely for the nonexistence of any subdirectories! It relies
+            # on the observed (by Ed) fact that all the masked directories
+            # contain further subdirectories on the host. If there's ever
+            # a new masked directory that contains only files, this test
+            # will silently pass without any indication of error.
+            # If you can think of a better way to do this check,
+            # please feel free to fix it.
+            is "$nlinks" "2" "$path: directory link count"
```
giuseppe, line 158: wouldn't `ls -A $path` or `find $path -mindepth 1` (and make sure the output is empty) be sufficient for this check?

edsantiago: Yes... but then I'd have to do it in a loop, and I'm very mindful of run time (perhaps too mindful?) Every second adds up, and ten `podman run` executions add up to over one second. With just one `podman run`, it's \<0.01.  There are times when I can't avoid multiple podman invocations but here I wanted to add a good-enough test without adding too much time to the already-long runs.

giuseppe: it makes sense, thanks for the explanation

TomSweeneyRedHat:
LGTM
and happy green test buttons.

PR #6991, opened 2020-07-15
Make changes to /etc/passwd on disk for non-read only

### LINKED ISSUE #6953: podman exec: Error: unable to find user $USER: no matching entries in passwd file

mheon:
Bind-mounting /etc/passwd into the container is problematic becuase of how system utilities like `useradd` work. They want to make a copy and then rename to try to prevent breakage; this is, unfortunately, impossible when the file they want to rename is a bind mount. The current behavior is fine for read-only containers, though, because we expect useradd to fail in those cases.

Instead of bind-mounting, we can edit /etc/passwd in the container's rootfs. This is kind of gross, because the change will show up in `podman diff` and similar tools, and will be included in images made by `podman commit`. However, it's a lot better than breaking important system tools.

Fixes #6953

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1458,11 +1458,26 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, error) {
 	if uid == 0 {
 		return "", nil
 	}
+
 	u, err := user.LookupId(strconv.Itoa(rootless.GetRootlessUID()))
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to get current user")
 	}
-	return fmt.Sprintf("%s:x:%s:%s:%s:%s:/bin/sh\n", u.Username, u.Uid, u.Gid, u.Username, c.WorkingDir()), nil
+
+	// Lookup the user to see if it exists in the container image.
+	_, err = lookup.GetUser(c.state.Mountpoint, u.Username)
+	if err != User.ErrNoPasswdEntries {
+		return "", err
+	}
+
+	// If the user's actual home directory exists, or was mounted in - use
+	// that.
+	homeDir := c.WorkingDir()
+	if MountExists(c.config.Spec.Mounts, u.HomeDir) {
+		homeDir = u.HomeDir
+	}
+
+	return fmt.Sprintf("%s:x:%s:%s:%s:%s:/bin/sh\n", u.Username, u.Uid, u.Gid, u.Username, homeDir), nil
 }
 
 // generateUserPasswdEntry generates an /etc/passwd entry for the container user
@@ -1488,11 +1503,8 @@ func (c *Container) generateUserPasswdEntry() (string, error) {
 
 	// Lookup the user to see if it exists in the container image
 	_, err = lookup.GetUser(c.state.Mountpoint, userspec)
-	if err != nil && err != User.ErrNoPasswdEntries {
+	if err != User.ErrNoPasswdEntries {
 		return "", err
-	}
-	if err == nil {
-		return "", nil
 	}
 
 	if groupspec != "" {
@@ -1542,6 +1554,32 @@ func (c *Container) generatePasswd() (string, error) {
 	if pwd == "" {
 		return "", nil
 	}
+
+	// If we are *not* read-only - edit /etc/passwd in the container.
+	// This is *gross* (shows up in changes to the container, will be
+	// committed to images based on the container) but it actually allows us
+	// to add users to the container (a bind mount breaks useradd).
+	// We should never get here twice, because generateUserPasswdEntry will
+	// not return anything if the user already exists in /etc/passwd.
+	if !c.IsReadOnly() {
+		containerPasswd, err := securejoin.SecureJoin(c.state.Mountpoint, "/etc/passwd")
+		if err != nil {
+			return "", errors.Wrapf(err, "error looking up location of container %s /etc/passwd", c.ID())
+		}
+
+		f, err := os.OpenFile(containerPasswd, os.O_APPEND|os.O_WRONLY, 0600)
+		if err != nil {
+			return "", errors.Wrapf(err, "error opening container %s /etc/passwd", c.ID())
+		}
+		defer f.Close()
+
+		if _, err := f.WriteString(pwd); err != nil {
+			return "", errors.Wrapf(err, "unable to append to container %s /etc/passwd", c.ID())
+		}
+
+		return "", nil
+	}
+
 	originPasswdFile := filepath.Join(c.state.Mountpoint, "/etc/passwd")
 	orig, err := ioutil.ReadFile(originPasswdFile)
 	if err != nil && !os.IsNotExist(err) {
--- a/test/e2e/run_passwd_test.go
+++ b/test/e2e/run_passwd_test.go
@@ -35,27 +35,27 @@ var _ = Describe("Podman run passwd", func() {
 	})
 
 	It("podman run no user specified ", func() {
-		session := podmanTest.Podman([]string{"run", BB, "mount"})
+		session := podmanTest.Podman([]string{"run", "--read-only", BB, "mount"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.LineInOutputContains("passwd")).To(BeFalse())
 	})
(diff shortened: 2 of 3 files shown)


libpod/container_internal_linux.go
```diff
@@ -1458,11 +1458,38 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, error) {
...
-	return fmt.Sprintf("%s:x:%s:%s:%s:%s:/bin/sh\n", u.Username, u.Uid, u.Gid, u.Username, c.WorkingDir()), nil
+
+	// Lookup the user to see if it exists in the container image.
+	_, err = lookup.GetUser(c.state.Mountpoint, u.Username)
+	if err != nil && err != User.ErrNoPasswdEntries {
+		return "", err
+	}
+	if err == nil {
+		return "", nil
+	}
```
edsantiago, line 1474: Could these two `if`s be refactored into one?
```go
if err != User.ErrNoPasswdEntries {
    return "", err
}
```

libpod/container_internal_linux.go
```diff
@@ -1458,11 +1458,38 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, error) {
...
+	if !hasHomedir {
+		ctrHomedir, err := securejoin.SecureJoin(c.state.Mountpoint, u.HomeDir)
+		if err == nil {
+			if s, err := os.Stat(ctrHomedir); err == nil && s.IsDir() {
+				hasHomedir = true
+			}
+		}
+	}
+	if hasHomedir {
+		homeDir = u.HomeDir
```
edsantiago, line 1489: Would it make sense to set WORKINGDIR here too (if unset)?
```console
$ ./bin/podman run -it --rm --userns=keep-id --security-opt label=disable -v /home/esm:/home/esm alpine_labels sh
/ $ pwd
/         <--- I kind of expected /home/esm
/ $ tail -1 /etc/passwd
esm:x:14904:14904:esm:/home/esm:/bin/sh     <--- it was detected
/ $ ls -la /home/esm
[as expected]                            <---- and it exists
```

libpod/container_internal_linux.go
```diff
@@ -1458,11 +1458,38 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, error) {
...
 	}
+
 	u, err := user.LookupId(strconv.Itoa(rootless.GetRootlessUID()))
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to get current user")
 	}
-	return fmt.Sprintf("%s:x:%s:%s:%s:%s:/bin/sh\n", u.Username, u.Uid, u.Gid, u.Username, c.WorkingDir()), nil
+
+	// Lookup the user to see if it exists in the container image.
+	_, err = lookup.GetUser(c.state.Mountpoint, u.Username)
```
edsantiago, line 1468: Corner case: would it make sense to do lookup by UID instead of name? This is super-unlikely, but if the podman user has a UID already defined in the image's `/etc/passwd`, appending a new entry is going to be mostly a NOP: `ls` and `ps` and anything that does lookups is going to show the preexisting username.

libpod/container_internal_linux.go
mheon: I'm reluctant to reset workdir given it could have been deliberately set by the user - I don't have a way of telling that this deep in the code.

libpod/container_internal_linux.go
mheon: I considered that, and decided it's probably not worth the potential performance hit of a duplicate lookup.

libpod/container_internal_linux.go
rhatdan: If workingdir == "" then we know it was never set.

rhatdan:
Nice work @mheon, I was thinking of doing this myself.  Glad you did it.

libpod/container_internal_linux.go
mheon: By this point, in the spec, it's already been defaulted to `/` if it was not set by the user.

rhatdan:
Does this work for both use cases.

if I specify a --user 1234, we add an entry for 1234 to /etc/passwd.

If you use --userns=keepid we add your entry to /etc/passwd.

mheon:
Should work for both - this happens after we've already generated both entries for `/etc/passwd` and the only difference is where we write them to (a new file or the existing passwd)

rhatdan:
@owtaylor FYI

mheon:
Added a further test and un-broke the existing passwd tests (used read-only mode to force us back to the old bind mount behavior).

libpod/container_internal_linux.go
```diff
@@ -1458,11 +1458,35 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, error) {
...
```
mtrmac, line 1468: (A pedantic note no-one should care about: POSIX does allow all-numeric usernames, with the username different from the UID value. In user input like the chown CLI, where the input is ambiguous, the input is defined to be interpreted as a username if it exists, in preference to UID.

Either way `lookup.GetUser` parsing the string and guessing whether the caller meant a name or an UID does the wrong thing, when we know the input is a name.)

mtrmac, line 1468: What if the user does exist inside the container, but with a different UID?

libpod/container_internal_linux.go
```diff
@@ -1458,11 +1458,35 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, error) {
...
+	if err != User.ErrNoPasswdEntries {
+		return "", err
+	}
+
+	// If the user's actual home directory exists, or was mounted in - use
+	// that.
```
mtrmac, line 1474: This is interesting, but somewhat out of scope.

- Is it _safe_? I can easily imagine containers that assume that `~` == `WORKDIR`.
- Also, is it OK to give the container this information about the outside? There is the precedent of telling the container the user name already, sure.

libpod/container_internal_linux.go
mtrmac: (I find it easier to read with the `err == nil` case handled explicitly, but either is fine.)

mheon:
Finally green

libpod/container_internal_linux.go
mheon: The primary case I was trying to capture here was pet containers (Toolbox) which explicitly bind-mount in the user's `/home`. I think we're safe to set in that case, given the user's /home was mounted in explicitly at container creation.

Dropping the case where we set `/home` if it exists in the container rootfs would simplify things and prevent the leakage when the home directory was not explicitly added, so I think I'll do that.

mheon:
@rhatdan @vrothberg @TomSweeneyRedHat @jwhonce PTAL and merge

rhatdan:
LGTM
I think we should think of some corner cases like @mtrmac Talks about, but we have those already.

libpod/container_internal_linux.go
mheon: I think we end up with duplicate entries in that case, with this one being the second of the two. I don't think that's *completely* disastrous, but it's definitely not ideal.

libpod/container_internal_linux.go
mheon: Update: Dropped down to just an explicit mount.

libpod/container_internal_linux.go
rhatdan: ok

rhatdan:
@vrothberg @giuseppe @baude @jwhonce @TomSweeneyRedHat @ashley-cui @QiWang19 PTAL

edsantiago:
LGTM

Issue #6933, opened 2020-07-10
Error: invalid config provided: AppArmorProfile and privileged are mutually exclusive options


polygamma:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`sudo podman run --privileged hello-world` leads to
```
Error: invalid config provided: AppArmorProfile and privileged are mutually exclusive options
```
when using Ubuntu 20.04.

I was informed by a colleague of mine about this problem, when he tried to use an image of us, and could not get it to work.

Using Arch Linux, I could not reproduce the problem.

**Steps to reproduce the issue:**

To reproduce it, I installed Ubuntu 20.04 in a VirtualBox VM where the error appears, too.

1. Download the latest Ubuntu 20.04 image from https://releases.ubuntu.com/20.04/ and install Ubuntu 20.04 in a VirtualBox VM

2. Update the system and install curl: `sudo apt-get update && sudo apt-get dist-upgrade && sudo apt-get install curl`

3. Install Podman

```bash
source /etc/os-release
sudo sh -c "echo 'deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list"
curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key | sudo apt-key add -
sudo apt-get update
sudo apt-get install podman
```

4. Run the command `sudo podman run --privileged hello-world` and get the error.

I want to explicitly state at this point, that I did nothing else, besides updating the system, installing curl and installing Podman.

**Describe the results you received:**

I could not run the `hello-world` image when using `sudo` and `--privileged`.

**Describe the results you expected:**

I want to be able to run the `hello-world` image, even when using `sudo` and `--privileged`.

**Output of `podman version`:**

```
Version:      2.0.2
API Version:  1
Go Version:   go1.14.2
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `sudo podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.18, commit: '
  cpus: 1
[...]

PR #6992, opened 2020-07-15
Support default profile for apparmor

rhatdan:
Currently you can not apply an ApparmorProfile if you specify
--privileged.  This patch will allow both to be specified
simultaniosly.

By default Apparmor should be disabled if the user
specifies --privileged, but if the user specifies --security apparmor:PROFILE,
with --privileged, we should do both.

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -405,7 +405,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	)
 	createFlags.StringArrayVar(
 		&cf.SecurityOpt,
-		"security-opt", containerConfig.SecurityOptions(),
+		"security-opt", []string{},
 		"Security Options",
 	)
 	createFlags.String(
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -512,10 +512,8 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 				s.ContainerSecurityConfig.SelinuxOpts = append(s.ContainerSecurityConfig.SelinuxOpts, con[1])
 				s.Annotations[define.InspectAnnotationLabel] = strings.Join(s.ContainerSecurityConfig.SelinuxOpts, ",label=")
 			case "apparmor":
-				if !c.Privileged {
-					s.ContainerSecurityConfig.ApparmorProfile = con[1]
-					s.Annotations[define.InspectAnnotationApparmor] = con[1]
-				}
+				s.ContainerSecurityConfig.ApparmorProfile = con[1]
+				s.Annotations[define.InspectAnnotationApparmor] = con[1]
 			case "seccomp":
 				s.SeccompProfilePath = con[1]
 				s.Annotations[define.InspectAnnotationSeccomp] = con[1]
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -41,7 +41,6 @@ case "${OS_RELEASE_ID}" in
     ubuntu)
 	    apt-get update
 	    apt-get install -y containers-common
-	    sed -ie 's/^\(# \)\?apparmor_profile =.*/apparmor_profile = ""/' /etc/containers/containers.conf
 	    if [[ "$OS_RELEASE_VER" == "19" ]]; then
 		    apt-get purge -y --auto-remove golang*
 		    apt-get install -y golang-1.13
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -250,7 +250,7 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 	}
 
 	// Apply AppArmor checks and load the default profile if needed.
-	if !c.config.Privileged {
+	if len(c.config.Spec.Process.ApparmorProfile) > 0 {
 		updatedProfile, err := apparmor.CheckProfileAndLoadDefault(c.config.Spec.Process.ApparmorProfile)
 		if err != nil {
 			return nil, err
--- a/pkg/specgen/container_validate.go
+++ b/pkg/specgen/container_validate.go
@@ -81,10 +81,6 @@ func (s *SpecGenerator) Validate() error {
 	if len(s.CapAdd) > 0 && s.Privileged {
 		return exclusiveOptions("CapAdd", "privileged")
 	}
-	// apparmor and privileged are exclusive
-	if len(s.ApparmorProfile) > 0 && s.Privileged {
-		return exclusiveOptions("AppArmorProfile", "privileged")
-	}
 	// userns and idmappings conflict
 	if s.UserNS.IsPrivate() && s.IDMappings == nil {
 		return errors.Wrap(ErrInvalidSpecConfig, "IDMappings are required when not creating a User namespace")
--- a/pkg/specgen/generate/oci.go
+++ b/pkg/specgen/generate/oci.go
@@ -285,13 +285,6 @@ func SpecGenToOCI(ctx context.Context, s *specgen.SpecGenerator, rt *libpod.Runt
 		}
 	}
 
-	// SECURITY OPTS
-	g.SetProcessNoNewPrivileges(s.NoNewPrivileges)
-
-	if !s.Privileged {
-		g.SetProcessApparmorProfile(s.ApparmorProfile)
-	}
-
 	BlockAccessToKernelFilesystems(s.Privileged, s.PidNS.IsHost(), &g)
 
 	for name, val := range s.Env {
--- a/pkg/specgen/generate/security.go
+++ b/pkg/specgen/generate/security.go
@@ -3,6 +3,7 @@ package generate
 import (
 	"strings"
 
+	"github.com/containers/common/pkg/apparmor"
 	"github.com/containers/common/pkg/capabilities"
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/libpod/v2/libpod"
@@ -56,6 +57,28 @@ func setLabelOpts(s *specgen.SpecGenerator, runtime *libpod.Runtime, pidConfig s
 	return nil
 }
 
+func setupApparmor(s *specgen.SpecGenerator, rtc *config.Config, g *generate.Generator) error {
+	hasProfile := len(s.ApparmorProfile) > 0
+	if !apparmor.IsEnabled() {
+		if hasProfile {
+			return errors.Errorf("Apparmor profile %q specified, but Apparmor is not enabled on this system", s.ApparmorProfile)
+		}
+		return nil
+	}
+	// If privileged and caller did not specify apparmor profiles return
+	if s.Privileged && !hasProfile {
+		return nil
+	}
+	if !hasProfile {
+		s.ApparmorProfile = rtc.Containers.ApparmorProfile
+	}
+	if len(s.ApparmorProfile) > 0 {
+		g.SetProcessApparmorProfile(s.ApparmorProfile)
+	}
+
+	return nil
+}
+
 func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator, newImage *image.Image, rtc *config.Config) error {
 	var (
 		caplist []string
@@ -105,6 +128,13 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 			}
 		}
 	}
+
+	g.SetProcessNoNewPrivileges(s.NoNewPrivileges)
+
+	if err := setupApparmor(s, rtc, g); err != nil {
+		return err
+	}
+
 	configSpec := g.Config
 	configSpec.Process.Capabilities.Bounding = caplist
 
--- a/test/e2e/run_apparmor_test.go
+++ b/test/e2e/run_apparmor_test.go
@@ -1,0 +1,158 @@
+// +build !remote
+
+package integration
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/containers/common/pkg/apparmor"
+	. "github.com/containers/libpod/v2/test/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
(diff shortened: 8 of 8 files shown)


rhatdan:
@vrothberg @saschagrunert PTAL.

I don't have access right now to systems that run with apparmor enabled. but I think this will fix some of the issues we have with it right now.

I can hand this PR off to an Intern to deal with it, if neither of you want to take it and run with it.
Basically we need tests to make sure that we only create the AppArmor profile if the user specifies it,
containers.conf specifies it and the user does not specify --privileged.

rhatdan:
The goal is to make sure we fix https://github.com/containers/podman/issues/6933

mheon:
LGTM, but someone who actually know the Apparmor code should review and merge

rhatdan:
Still flying blind, but added a new run_apparmor_test.go test, no way to test this locally, at this point, so will see how it does.

rhatdan:
@saschagrunert Could you try this out on an Apparmor enabled machine.  The tests are failing only on the default test for Ubuntu.  Testing locally on my Ubuntu VM, they pass.  All of the other tests, pass, I have no idea why the default test is failing.

Basically it creates a container and then makes sure the default apparmor policy is defined in the inspect code.

saschagrunert:
> @saschagrunert Could you try this out on an Apparmor enabled machine. The tests are failing only on the default test for Ubuntu. Testing locally on my Ubuntu VM, they pass. All of the other tests, pass, I have no idea why the default test is failing.
> 
> Basically it creates a container and then makes sure the default apparmor policy is defined in the inspect code.

Works locally on my side. This line looks suspicious:

https://github.com/containers/podman/blob/344a791ac2019b078dd334ef0a492ee37fe0a937/contrib/cirrus/setup_environment.sh#L44

We can also add something like 
```
> systemctl status --no-pager apparmor
> aa-status
```
to debug the current apparmor status.

rhatdan:
Thanks @saschagrunert Nice catch.  Does containers-common on ubuntu automatically install an /etc/containers/containers.conf?

saschagrunert:
> Thanks @saschagrunert Nice catch. Does containers-common on ubuntu automatically install an /etc/containers/containers.conf?

Yes, the file looks like this on Ubuntu 20.04:

```
root@74a8be14edff:/# cat /etc/containers/containers.conf
# The containers configuration file specifies all of the available configuration
# command-line options/flags for container engine tools like Podman & Buildah,
# but in a TOML format that can be easily modified and versioned.

# Please refer to containers.conf(5) for details of all configuration options.
# Not all container engines implement all of the options.
# All of the options have hard coded defaults and these options will override
# the built in defaults. Users can then override these options via the command
# line. Container engines will read containers.conf files in up to three
# locations in the following order:
#  1. /usr/share/containers/containers.conf
#  2. /etc/containers/containers.conf
#  3. $HOME/.config/containers/containers.conf (Rootless containers ONLY)
#  Items specified in the latter containers.conf, if they exist, override the
# previous containers.conf settings, or the default settings.

[containers]

# List of devices. Specified as
# "<device-on-host>:<device-on-container>:<permissions>", for example:
# "/dev/sdc:/dev/xvdc:rwm".
# If it is empty or commented out, only the default devices will be used
#
# devices = []

# List of volumes. Specified as
# "<directory-on-host>:<directory-in-container>:<options>", for example:
# "/db:/var/lib/db:ro".
# If it is empty or commented out, no volumes will be added
#
# volumes = []

# Used to change the name of the default AppArmor profile of container engine.
#
# apparmor_profile = "container-default"

# List of annotation. Specified as
# "key=value"
# If it is empty or commented out, no annotations will be added
#
# annotations = []

# Default way to to create a cgroup namespace for the container
# Options are:
# `private` Create private Cgroup Namespace for the container.
# `host`    Share host Cgroup Namespace with the container.
#
# cgroupns = "private"

# Control container cgroup configuration
# Determines  whether  the  container will create CGroups.
# Options are:
# `enabled`   Enable cgroup support within container
# `disabled`  Disable cgroup support, will inherit cgroups from parent
# `no-conmon` Container engine runs run without conmon
#
# cgroups = "enabled"

# List of default capabilities for containers. If it is empty or commented out,
# the default capabilities defined in the container engine will be added.
#
# default_capabilities = [
#    "AUDIT_WRITE",
#    "CHOWN",
#    "DAC_OVERRIDE",
#    "FOWNER",
#    "FSETID",
#    "KILL",
#    "MKNOD",
#    "NET_BIND_SERVICE",
#    "NET_RAW",
#    "SETGID",
#    "SETPCAP",
#    "SETUID",
#    "SYS_CHROOT",
# ]

# A list of sysctls to be set in containers by default,
# specified as "name=value",
# for example:"net.ipv4.ping_group_range = 0 1000".
#
# default_sysctls = [
#  "net.ipv4.ping_group_range=0 1000",
# ]

# A list of ulimits to be set in containers by default, specified as
# "<ulimit name>=<soft limit>:<hard limit>", for example:
# "nofile=1024:2048"
# See setrlimit(2) for a list of resource names.
# Any limit not specified here will be inherited from the process launching the
# container engine.
# Ulimits has limits for non privileged container engines.
#
# default_ulimits = [
#  “nofile”=”1280:2560”,
# ]

# List of default DNS options to be added to /etc/resolv.conf inside of the container.
#
# dns_options = []

# List of default DNS search domains to be added to /etc/resolv.conf inside of the container.
#
# dns_searches = []

# Set default DNS servers.
# This option can be used to override the DNS configuration passed to the
# container. The special value “none” can be specified to disable creation of
# /etc/resolv.conf in the container.
# The /etc/resolv.conf file in the image will be used without changes.
#
# dns_servers = []

# Environment variable list for the conmon process; used for passing necessary
# environment variables to conmon or the runtime.
#
# env = [
#    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
# ]

# Pass all host environment variables into the container.
#
# env_host = false

# Path to OCI hooks directories for automatically executed hooks.
#
# hooks_dir = [
# “/usr/share/containers/oci/hooks.d”,
# ]

# Default proxy environment variables passed into the container.
# The environment variables passed in include:
# http_proxy, https_proxy, ftp_proxy, no_proxy, and the upper case versions of
# these. This option is needed when host system uses a proxy but container
# should not use proxy. Proxy environment variables specified for the container
# in any other way will override the values passed from the host.
#
# http_proxy = true

# Run an init inside the container that forwards signals and reaps processes.
#
# init = false

#  Container init binary, if init=true, this is the init binary to be used for containers.
#
# init_path = "/usr/libexec/podman/catatonit"

# Default way to to create an IPC namespace (POSIX SysV IPC) for the container
# Options are:
# `private` Create private IPC Namespace for the container.
# `host`    Share host IPC Namespace with the container.
#
# ipcns = "private"

# Flag tells container engine to whether to use container separation using
# MAC(SELinux)labeling or not.
# Flag is ignored on label disabled systems.
#
# label = true

# Logging driver for the container. Available options: k8s-file and journald.
#
# log_driver = "k8s-file"

# Maximum size allowed for the container log file. Negative numbers indicate
# that no size limit is imposed. If positive, it must be >= 8192 to match or
# exceed conmon's read buffer. The file is truncated and re-opened so the
# limit is never exceeded.
#
# log_size_max = -1

# Default way to to create a Network namespace for the container
# Options are:
# `private` Create private Network Namespace for the container.
# `host`    Share host Network Namespace with the container.
# `none`    Containers do not use the network
#
# netns = "private"

# Create /etc/hosts for the container.  By default, container engine manage
# /etc/hosts, automatically adding  the container's  own  IP  address.
#
# no_hosts = false

# Maximum number of processes allowed in a container.
#
# pids_limit = 2048

# Default way to to create a PID namespace for the container
# Options are:
# `private` Create private PID Namespace for the container.
# `host`    Share host PID Namespace with the container.
#
# pidns = "private"

# Path to the seccomp.json profile which is used as the default seccomp profile
# for the runtime.
#
# seccomp_profile = "/usr/share/containers/seccomp.json"

# Size of /dev/shm. Specified as <number><unit>.
# Unit is optional, values:
# b (bytes), k (kilobytes), m (megabytes), or g (gigabytes).
# If the unit is omitted, the system uses bytes.
#
# shm_size = "65536k"

# Default way to to create a UTS namespace for the container
# Options are:
# `private`        Create private UTS Namespace for the container.
# `host`    Share host UTS Namespace with the container.
#
# utsns = "private"

# Default way to to create a User namespace for the container
# Options are:
# `auto`        Create unique User Namespace for the container.
# `host`    Share host User Namespace with the container.
#
# userns = "host"

# Number of UIDs to allocate for the automatic container creation.
# UIDs are allocated from the “container” UIDs listed in
# /etc/subuid & /etc/subgid
#
# userns_size=65536

# The network table contains settings pertaining to the management of
# CNI plugins.

[network]

# Path to directory where CNI plugin binaries are located.
#
# cni_plugin_dirs = ["/usr/libexec/cni"]

# Path to the directory where CNI configuration files are located.
#
# network_config_dir = "/etc/cni/net.d/"

[engine]

# Cgroup management implementation used for the runtime.
# Valid options “systemd” or “cgroupfs”
#
# cgroup_manager = "systemd"

# Environment variables to pass into conmon
#
# conmon_env_vars = [
#        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
# ]

# Paths to look for the conmon container manager binary
#
# conmon_path = [
#        "/usr/libexec/podman/conmon",
#        "/usr/local/libexec/podman/conmon",
#        "/usr/local/lib/podman/conmon",
#        "/usr/bin/conmon",
#        "/usr/sbin/conmon",
#        "/usr/local/bin/conmon",
#        "/usr/local/sbin/conmon"
# ]

# Specify the keys sequence used to detach a container.
# Format is a single character [a-Z] or a comma separated sequence of
# `ctrl-<value>`, where `<value>` is one of:
# `a-z`, `@`, `^`, `[`, `\`, `]`, `^` or `_`
#
# detach_keys = "ctrl-p,ctrl-q"

# Determines whether engine will reserve ports on the host when they are
# forwarded to containers. When enabled, when ports are forwarded to containers,
# ports are held open by as long as the container is running, ensuring that
# they cannot be reused by other programs on the host. However, this can cause
# significant memory usage if a container has many ports forwarded to it.
# Disabling this can save memory.
#
# enable_port_reservation = true

# Selects which logging mechanism to use for container engine events.
# Valid values are `journald`, `file` and `none`.
#
# events_logger = "journald"

# Default transport method for pulling and pushing for images
#
# image_default_transport = "docker://"

# Default command to run the infra container
#
# infra_command = "/pause"

# Infra (pause) container image name for pod infra containers.  When running a
# pod, we start a `pause` process in a container to hold open the namespaces
# associated with the  pod.  This container does nothing other then sleep,
# reserving the pods resources for the lifetime of the pod.
#
# infra_image = "k8s.gcr.io/pause:3.2"

# Specify the locking mechanism to use; valid values are "shm" and "file".
# Change the default only if you are sure of what you are doing, in general
# "file" is useful only on platforms where cgo is not available for using the
# faster "shm" lock type.  You may need to run "podman system renumber" after
# you change the lock type.
#
# lock_type** = "shm"

# Default engine namespace
# If engine is joined to a namespace, it will see only containers and pods
# that were created in the same namespace, and will create new containers and
# pods in that namespace.
# The default namespace is "", which corresponds to no namespace. When no
# namespace is set, all containers and pods are visible.
#
# namespace = ""

# Whether to use chroot instead of pivot_root in the runtime
#
# no_pivot_root = false

# Number of locks available for containers and pods.
# If this is changed, a lock renumber must be performed (e.g. with the
# 'podman system renumber' command).
#
# num_locks = 2048

# Whether to pull new image before running a container
# pull_policy = "missing"

# Directory for persistent engine files (database, etc)
# By default, this will be configured relative to where the containers/storage
# stores containers
# Uncomment to change location from this default
#
# static_dir = "/var/lib/containers/storage/libpod"

# Directory for temporary files. Must be tmpfs (wiped after reboot)
#
# tmp_dir = "/var/run/libpod"

# Directory for libpod named volumes.
# By default, this will be configured relative to where containers/storage
# stores containers.
# Uncomment to change location from this default.
#
# volume_path = "/var/lib/containers/storage/volumes"

# Default OCI runtime
#
# runtime = "runc"

# List of the OCI runtimes that support --format=json.  When json is supported
# engine will use it for reporting nicer errors.
#
# runtime_supports_json = ["crun", "runc", "kata"]

# List of the OCI runtimes that supports running containers without cgroups.
#
# runtime_supports_nocgroups = ["crun"]

# List of the OCI runtimes that supports running containers with KVM Separation.
#
# runtime_supports_kvm = ["kata"]

# Paths to look for a valid OCI runtime (runc, runv, kata, etc)
[engine.runtimes]
# runc = [
#        "/usr/bin/runc",
#        "/usr/sbin/runc",
#        "/usr/local/bin/runc",
#        "/usr/local/sbin/runc",
#        "/sbin/runc",
#        "/bin/runc",
#        "/usr/lib/cri-o-runc/sbin/runc",
# ]

# crun = [
#            "/usr/bin/crun",
#            "/usr/sbin/crun",
#            "/usr/local/bin/crun",
#            "/usr/local/sbin/crun",
#            "/sbin/crun",
#            "/bin/crun",
#            "/run/current-system/sw/bin/crun",
# ]

# kata = [
#            "/usr/bin/kata-runtime",
#            "/usr/sbin/kata-runtime",
#            "/usr/local/bin/kata-runtime",
#            "/usr/local/sbin/kata-runtime",
#            "/sbin/kata-runtime",
#            "/bin/kata-runtime",
#            "/usr/bin/kata-qemu",
#            "/usr/bin/kata-fc",
# ]

# Number of seconds to wait for container to exit before sending kill signal.
#stop_timeout = 10

# The [engine.runtimes] table MUST be the last entry in this file.
# (Unless another table is added)
# TOML does not provide a way to end a table other than a further table being
# defined, so every key hereafter will be part of [runtimes] and not the main
# config.
```

rhatdan:
Ok, I don't believe we are shipping this file in Fedora, only /usr/share/containers/containers.conf.

rhatdan:
@saschagrunert Looks like that was the problem.  Thanks for finding it.

mheon:
LGTM

Let's get this merged so we can backport to the v2.0 branch for 2.0.3 - I'd like to have this fix in for the Ubuntu and Debian folks.

Issue #6995, opened 2020-07-16
`podman play kube` doesn't ignore dockerfile ENTRYPOINT


kaiyasuda:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
`podman play kube` doesn't ignore dockerfile ENTRYPOINT when command parameter set.  just append.
This difference between Podman and Kubernetes means there is an incompatibility between the Kubernetes YAML files that work with Podman and those that work with Kubernetes.

**Steps to reproduce the issue:**

1. make pod with `podman play kube`

```
# podman play kube ipc.yml
```

```:ipc.yml
# cat ipc.yml
apiVersion: v1
kind: Pod
metadata:
  name: mc2
spec:
  containers:
  - name: producer
    image: allingeek/ch6_ipc
    command: ["./ipc", "-producer"]
  - name: consumer
    image: allingeek/ch6_ipc
    command: ["./ipc", "-consumer"]
  restartPolicy: Never
```

./ipc is binary file to demonstrate Inter Process Communication(IPC) in pod.
This binary requires argments either "-consumer" or "-producer".

**Describe the results you received:**

but these containers don't work.
because argments of ./ipc is "./ipc -consumer" or  "./ipc -producer."

```
# podman container ls -a --format "{{.Names}} {{.Command}}"
NAMES                 COMMAND
mc2-producer     ./ipc ./ipc -prod...
mc2-consumer    ./ipc ./ipc -cons...
```

**Describe the results you expected:**

According to [Kubernetes Documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes),

>If you supply a command and args, the default Entrypoint and the default Cmd defined in the Docker image are ignored. Your command is run with your args.

but podman doesn't ignore dockerfile ENTRYPOINT, just append.

This difference between Podman and Kubernetes means there is an incompatibility between the Kubernetes YAML files that work with Podman and those that work with Kubernetes.

**Output of `podman version`:**

```
# podman version
Version:      2.0.2
API Version:  1
Go Version:   go1.13.4
Built:        Wed Dec 31 19:00:00 1969
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```:yml
# podman info --debug
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.19-1.el8.x86_64
    path: /usr/bin/conmon
[...]

Issue #5730, opened 2020-04-06
play kube doesn't support container args


athena-metis:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

play kube doesn't support the "args" config option for a container in a YAML pod spec.

Behaviour of command and args are defined here:
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

In the code it always uses the image entry point (expected), then if a command is specified it is appended to the image entrypoint, otherwise if defined the image command is used (expected).
However there is no handling of args, which if specified without a command should replace the image Cmd, or if specified with a command should be appended to the command, neither of which as handled.
https://github.com/containers/libpod/blob/1fe2fbb42114b9072a1caf359beff63042df90fd/pkg/adapter/pods.go#L882-L894

**Steps to reproduce the issue:**

1. Add args to a container in a pod spec YAML

2. Use play kube to start the pod

3. Use podman inspect to inspect the running container

**Describe the results you received:**
Cmd from inspect is composed of the image entry point and the image Cmd

**Describe the results you expected:**
Cmd from inspect is composed of the image entry point and the args in the containers definition in the pod spec.

**Additional information you deem important (e.g. issue happens only occasionally):**
The inelegant work around is to include the arguments in the containers command option, but it would be better if play kube behaved more like kubernetes

**Output of `podman version`:**
```
podman version 1.8.2
```

**Output of `podman info --debug`:**

```
N/A
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-1.8.2-2.fc32.x86_64
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
physical

mheon:
Definitely sounds like a bug. Are you willing to do a PR to fix this?

rhatdan:
@kunalkushwaha @sujil02 FYI?

kunalkushwaha:
If no one working.. I can pick it up this too.

athena-metis:
@kunalkushwaha I did start on this but happy to leave it to you to pick up. I was struggling with adding appropriate tests as there don't appear to be any unit test suites covering any of the code in pkgs, and it would take several e2e tests for coverage which are quite time consuming to run so didn't want to be making many additions here. It would also probably required a different image to be used than the ones currently used for e2e as it needs one which defines an endpoint in the image

kunalkushwaha:
[...]

PR #7002, opened 2020-07-16
fix podman play kube doesn't override dockerfile ENTRYPOINT

zhangguanzhang:
fix https://github.com/containers/podman/issues/6995 https://github.com/containers/podman/issues/5730 https://github.com/containers/podman/issues/7019

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -453,11 +453,16 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
 
 	containerConfig.Command = []string{}
 	if imageData != nil && imageData.Config != nil {
-		containerConfig.Command = append(containerConfig.Command, imageData.Config.Entrypoint...)
+		containerConfig.Command = imageData.Config.Entrypoint
 	}
 	if len(containerYAML.Command) != 0 {
-		containerConfig.Command = append(containerConfig.Command, containerYAML.Command...)
-	} else if imageData != nil && imageData.Config != nil {
+		containerConfig.Command = containerYAML.Command
+	}
+	// doc https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
+	if len(containerYAML.Args) != 0 {
+		containerConfig.Command = append(containerConfig.Command, containerYAML.Args...)
+	} else if len(containerYAML.Command) == 0 {
+		// Add the Cmd from the image config only if containerYAML.Command and containerYAML.Args are empty
 		containerConfig.Command = append(containerConfig.Command, imageData.Config.Cmd...)
 	}
 	if imageData != nil && len(containerConfig.Command) == 0 {
--- a/test/e2e/play_kube_test.go
+++ b/test/e2e/play_kube_test.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	. "github.com/containers/libpod/v2/test/utils"
@@ -48,6 +49,10 @@ spec:
   {{ range . }}
   - command:
     {{ range .Cmd }}
+    - {{.}}
+    {{ end }}
+    args:
+    {{ range .Arg }}
     - {{.}}
     {{ end }}
     env:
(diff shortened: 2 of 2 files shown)


rhatdan:
@haircommander @mheon PTAL

haircommander:
thanks @zhangguanzhang LGTM

haircommander:
actually, I lied, I have a nit

pkg/domain/infra/abi/play.go
```diff
@@ -453,12 +453,17 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
...
+		containerConfig.Command = imageData.Config.Entrypoint
 	}
 	if len(containerYAML.Command) != 0 {
-		containerConfig.Command = append(containerConfig.Command, containerYAML.Command...)
-	} else if imageData != nil && imageData.Config != nil {
-		containerConfig.Command = append(containerConfig.Command, imageData.Config.Cmd...)
+		containerConfig.Command = containerYAML.Command
+	}
+	if imageData != nil && imageData.Config != nil {
+		if len(containerYAML.Args) != 0 {
```
haircommander, line 462: we shouldn't have this block here. it should be outside of the imageData block. We should set it to containerYAML.Args regardless of whether imageData is defined.

zhangguanzhang: ok,I will change it

pkg/domain/infra/abi/play.go
```diff
@@ -453,11 +453,14 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
 
 	containerConfig.Command = []string{}
 	if imageData != nil && imageData.Config != nil {
-		containerConfig.Command = append(containerConfig.Command, imageData.Config.Entrypoint...)
+		containerConfig.Command = imageData.Config.Entrypoint
 	}
 	if len(containerYAML.Command) != 0 {
-		containerConfig.Command = append(containerConfig.Command, containerYAML.Command...)
-	} else if imageData != nil && imageData.Config != nil {
+		containerConfig.Command = containerYAML.Command
```
TomSweeneyRedHat, line 459: should this still be an append, or do we want to take the yaml over the config data?  I think this is correct as is, but just wanted to double check.

zhangguanzhang: pod's command should override the image's Entrypoint
pod's args should override the image's Command

zhangguanzhang: could see this doc
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

| Description	| Docker field name	| Kubernetes field name	|
| :----- | :----- | :----- |
|The command run by the container	| Entrypoint	| command	|
|The arguments passed to the command	| Cmd	| args |

zhangguanzhang:
Please do not merge until I add the following unit tests tomorrow

zhangguanzhang:
all has done

rhatdan:
LGTM
Thanks @zhangguanzhang

zhangguanzhang:
@mheon @haircommander @TomSweeneyRedHat  PTAL

pkg/domain/infra/abi/play.go
```diff
@@ -453,11 +453,16 @@ func kubeContainerToCreateConfig(ctx context.Context, containerYAML v1.Container
...
+	}
+	// doc https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
+	if len(containerYAML.Args) != 0 {
+		containerConfig.Command = append(containerConfig.Command, containerYAML.Args...)
+	} else if len(containerYAML.Command) == 0 {
+		// Only containerYAML.Command can be appended when it is equals 0
```
haircommander, line 465: this comment seems a bit off. I'd recommend:
`Add the Cmd from the image config only if containerYAML.Command and containerYAML.Args are empty`

haircommander:
sorry, another nit, then LGTM

pkg/domain/infra/abi/play.go
zhangguanzhang: I have submitted the modification, and I will wait for the successful execution of CI later

rhatdan:
/lgtm
/hold
Thanks @zhangguanzhang

zhangguanzhang:
I think there is still a missing unit test for `ENTRYPOINT` and I will try to submit it tomorrow. Do not merge until then

zhangguanzhang:
now all has done

rhatdan:
Thanks @zhangguanzhang 
Nice work
/lgtm

Issue #6989, opened 2020-07-15
Non root user in docker.io/eclipse-mosquitto container cannot write to any directory under podman 2.0.2


Lalufu:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The docker.io/eclipse-mosquitto container contains a binary that drops privileges to UID/GID 1883. When run under podman 2.0.2 that user is not anle to write to any file/directory, regardless of permissions. When run under podman 1.8.2 things work.

**Steps to reproduce the issue:**

podman 2.0.2:

```
$ sudo podman run -it --rm docker.io/eclipse-mosquitto:1.6 /bin/sh
/ # cat > /tmp/mosquitto.conf <<EOF
> log_dest file /tmp/mosquitto.log
> EOF                          
/ # ls -ld /tmp                
drwxrwxrwt    1 root     root            28 Jul 15 19:02 /tmp
/ # /usr/sbin/mosquitto -c /tmp/mosquitto.conf
1594839785: Error: Unable to open log file /tmp/mosquitto.log for writing.
```

podman 1.8.2:

```
$ sudo podman run -it --rm docker.io/eclipse-mosquitto:1.6 /bin/sh
/ # cat > /tmp/mosquitto.conf <<EOF
> log_dest file /tmp/mosquitto.log
> EOF
/ # /usr/sbin/mosquitto -c /tmp/mosquitto.conf
^C
/ # ls -ln /tmp
total 8
-rw-r--r--    1 0        0               33 Jul 15 19:05 mosquitto.conf
-rw-------    1 1883     1883           253 Jul 15 19:05 mosquitto.log
```

**Describe the results you received:**
Writing files as the 1883 user inside the container fails even in locations that should succeed under 2.0.2

**Describe the results you expected:**
Same as under 1.8.2, the write should succeed.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

1.8.2:
```
Version:            1.8.2
RemoteAPI Version:  1
Go Version:         go1.14
OS/Arch:            linux/amd64
```

2.0.2:
```
Version:      2.0.2
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
debug:                         
  compiler: gc                 
  git commit: ""               
  go version: go1.14           
  podman version: 1.8.2        
host:                          
[...]

PR #7005, opened 2020-07-17
abi: set default umask and rlimits

giuseppe:
the code got lost in the migration to podman 2.0, reintroduce it.

Closes: https://github.com/containers/podman/issues/6989

### DIFF
--- a/cmd/podman/early_init_linux.go
+++ b/cmd/podman/early_init_linux.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+
+	"github.com/containers/libpod/v2/libpod/define"
+	"github.com/pkg/errors"
+)
+
+func setRLimits() error {
+	rlimits := new(syscall.Rlimit)
+	rlimits.Cur = define.RLimitDefaultValue
+	rlimits.Max = define.RLimitDefaultValue
+	if err := syscall.Setrlimit(syscall.RLIMIT_NOFILE, rlimits); err != nil {
+		if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, rlimits); err != nil {
+			return errors.Wrapf(err, "error getting rlimits")
+		}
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM

rhatdan:
Fixes: https://github.com/containers/podman/issues/6989

rhatdan:
@giuseppe good catch.

cmd/podman/root.go
```diff
@@ -171,6 +171,13 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if !registry.IsRemote() {
+		err := registry.ContainerEngine().EarlyProcessSetup(registry.Context())
```
mheon, line 175: This might be better as part of `cobra.OnInitialize()` - we're already configuring logging there

mheon: (Most notably, removes the need for a public function to do the work)

giuseppe: are we able there to distinguish between the remote and the ABI case?

giuseppe: well probably it makes no sense to distinguish, we can run this in both cases

cmd/podman/early_init_linux.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"os"
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+func setRLimits() error {
+	rlimits := new(syscall.Rlimit)
+	rlimits.Cur = 1048576
```
vrothberg, line 13: Nit: Can we make this a constant in `define`? It's also being used in `pkg/spec`.

giuseppe: yes that is a good idea.  Do we know where we took this magic number and what it means?

giuseppe: called `define.RLimitDefaultValue` for now

PR #6946, opened 2020-07-13
[WIP] Add support for Umask

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #6958, opened 2020-07-13
system tests: new tests for run, exec

edsantiago:
- Issue #6735 : problem with multiple namespaces; confirms
   combinations of --userns=keep-id, --privileged, --user=XX

 - Issue #6829 : --userns=keep-id will add a /etc/passwd entry

 - Issue #6593 : podman exec, with --userns=keep-id, errors
   (test is currently skipped because issue remains live)

Signed-off-by: Ed Santiago <santiago@redhat.com>

PR #7006, opened 2020-07-17
Add --umask flag for create, run

ashley-cui:
--umask sets the umask inside the container
Defaults to 0022

Replaces #6946

Co-authored-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -437,6 +437,9 @@ type ContainerConfig struct {
 	// Timezone is the timezone inside the container.
 	// Local means it has the same timezone as the host machine
 	Timezone string `json:"timezone,omitempty"`
+
+	// Umask is the umask inside the container.
+	Umask string `json:"umask,omitempty"`
 }
 
 // ContainerNamedVolume is a named volume that will be mounted into the
@@ -1276,5 +1279,8 @@ func (c *Container) AutoRemove() bool {
 
 func (c *Container) Timezone() string {
 	return c.config.Timezone
+}
 
+func (c *Container) Umask() string {
+	return c.config.Umask
 }
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	. "github.com/containers/libpod/v2/test/utils"
 	. "github.com/onsi/ginkgo"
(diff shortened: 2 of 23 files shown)


test/e2e/create_test.go
```diff
@@ -499,4 +499,33 @@ var _ = Describe("Podman create", func() {
...
+		inspect = podmanTest.Podman([]string{"inspect", "umask-short"})
+		inspect.WaitWithDefaultTimeout()
+		data = inspect.InspectContainerToJSON()
+		Expect(len(data)).To(Equal(1))
+		Expect(data[0].Config.Umask).To(Equal("0022"))
+
+		session = podmanTest.Podman([]string{"create", "--umask", "9999", "--name", "bad", ALPINE})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Not(Equal(0)))
+	})
```
TomSweeneyRedHat, line 530: You're testing both create and confirming with inspect in here which is good.  It might make sense to also add a part of this to the inspect test too, even though it would be redundant.  I'll let @mheon make the call there.

TomSweeneyRedHat:
All kinds of test unhappiness @ashley-cui

rhatdan:
The random tests are blowing up with
time="2020-07-17T16:18:53-04:00" level=error msg="error starting some container dependencies"
time="2020-07-17T16:18:53-04:00" level=error msg="\"Invalid Umask Value: strconv.ParseUint: parsing \\\"\\\": invalid syntax\""

rhatdan:
This looks like there is something wrong with the runtime-tools vendor, that is causing this issue.

```
# ./bin/podman pod create -p 8080:80 --name test1
d164202b886bb247f5b32b36ba0d6aa92aa14ab050a09eadc2fde0875cf1d13d
# ./bin/podman run --pod test1 fedora echo top
Error: open `/proc/1513113/ns/net`: No such file or directory: OCI runtime command not found error
```

@giuseppe @mheon PTAL

rhatdan:
@ashley-cui I think you might want
```
diff --git a/libpod/container_internal_linux.go b/libpod/container_internal_linux.go
index 53b7f31c4..ed0a0f2da 100644
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -341,12 +341,13 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
                g.SetProcessGID(uint32(execUser.Gid))
        }
 
-       decVal, err := strconv.ParseUint(c.config.Umask, 8, 32)
-       if err != nil {
-               return nil, errors.Wrapf(err, "Invalid Umask Value")
+       if c.config.Umask != "" {
+               decVal, err := strconv.ParseUint(c.config.Umask, 8, 32)
+               if err != nil {
+                       return nil, errors.Wrapf(err, "Invalid Umask Value: %q", c.config.Umask)
+               }
+               g.SetProcessUmask(uint32(decVal))
        }
-       g.SetProcessUmask(uint32(decVal))
-
        // Add addition groups if c.config.GroupAdd is not empty
        if len(c.config.Groups) > 0 {
                gids, err := lookup.GetContainerGroups(c.config.Groups, c.state.Mountpoint, overrides)
```

ashley-cui:
@rhatdan already updated :) the runtime-tools i vendored in myself, so it might be a vendoring error caused by me..

libpod/container.go
```diff
@@ -435,6 +435,9 @@ type ContainerConfig struct {
 	// Timezone is the timezone inside the container.
 	// Local means it has the same timezone as the host machine
 	Timezone string `json:"timezone,omitempty"`
+
+	// Umask is the umask inside the container.
+	Umask string `json:"umask,omitempty"`
```
mheon, line 440: Why are we storing this as a string? We should use whatever type the OCI spec uses - probably an int? Parsing should be done when we set the umask.

rhatdan: The value is actually an octal.  Storing it as an int, may make it hard to convert and hard to view in inspect.

test/e2e/create_test.go
```diff
@@ -499,4 +499,33 @@ var _ = Describe("Podman create", func() {
...
 	})
 
+	It("podman create --umask", func() {
+		session := podmanTest.Podman([]string{"create", "--name", "default", ALPINE})
+		session.WaitWithDefaultTimeout()
+		inspect := podmanTest.Podman([]string{"inspect", "default"})
+		inspect.WaitWithDefaultTimeout()
+		data := inspect.InspectContainerToJSON()
+		Expect(len(data)).To(Equal(1))
+		Expect(data[0].Config.Umask).To(Equal("0022"))
```
rhatdan, line 509: Coud you add a test for 

podman run --umask 0077 fedora umask
And make sure it outputs 0077

libpod/container.go
ashley-cui: yep, setting umask to 0022 would actually show up as 18 if stored as an int. i got stuck on this for a good hour or two when implementing LOL

ashley-cui:
I'm unable to reproduce locally the system tests that are failing but I confirmed that umask doesn't work on ubuntu at all. Any pointers on how to proceed? @rhatdan

edsantiago:
The system-test failures look suspiciously like something I may have introduced in #6958 - probably a bad interaction with a missing-cleanup bug in one of the integration tests. PR in progress.

The umask-on-ubuntu issue looks real though. FW(little)IW my guess is it's runc vs crun.

ashley-cui:
@edsantiago thanks! any idea who i should talk to about the ubuntu stuff?

edsantiago:
#7026 should fix the system-test flake. I'm sorry, I have no experience with Ubuntu and no idea how to investigate there. If you want to look into crun vs runc, you can reboot a Fedora system into cgroups v1 mode by booting with `systemd.unified_cgroup_hierarchy=0`

test/e2e/run_test.go
```diff
@@ -1081,4 +1081,34 @@ USER mail`
...
+		Expect(session.OutputToString()).To(Equal("0077"))
+
+		session = podmanTest.Podman([]string{"run", "--umask", "22", "--rm", ALPINE, "sh", "-c", "umask"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(Equal("0022"))
+
+		session = podmanTest.Podman([]string{"run", "--umask", "9999", "--rm", ALPINE, "sh", "-c", "umask"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Not(Equal(0)))
```
edsantiago, line 1112: Ed's standard rant: it's important to check error messages, not just error codes:
```go
    Expect(session.ErrorToString()).To(ContainSubstring("Invalid umask"))
```
Without that, the test could be failing for infinite unrelated reasons: too often a misspelled option or command name yields the expected nonzero exit status, but the test isn't actually testing anything.

Aside from that, lgtm. Nice set of tests.

ashley-cui: Sounds good, will keep in mind for future tests too :)

Issue #7021, opened 2020-07-20
podman.service: KillMode=process leaks pause process, breaks subsequent API invocations


martinpitt:
/kind bug
**Description**

podman.service, in particular the user one, uses `KillMode=process`.  When the unit stops, this leaves behind a process

    admin      20185  1.3  2.4 1281680 50484 ?       Ssl  05:31   0:00 /usr/bin/podman system service

Not only is this a resource leak, it actively breaks the next time the API gets invoked.

**Steps to reproduce the issue:**

1. Make sure user's podman service is enabled:

       systemctl  --user enable --now podman.socket

2. Make an API request to start podman.service:

       curl --unix $XDG_RUNTIME_DIR/podman/podman.sock http://localhost/

3. Log out the user

4. Ensure in `loginctl` (as root) that the user session is gone. But `pgrep -au` *username* still shows a leftover process:
```
# pgrep -au admin
17536 podman
```

5. Log back in as the user

6. Do another API request, same curl command

**Describe the results you received:**

In (6), the API request hangs. In journalctl you see an repeated restart/failure:
```
systemd[19935]: Starting Podman API Service...
systemd[19935]: podman.service: Succeeded.
systemd[19935]: podman.service: Unit process 19981 (podman pause) remains running after unit stopped.
systemd[19935]: Finished Podman API Service.
systemd[19935]: podman.service: Found left-over process 19981 (podman pause) in control group while starting unit. Ignoring.
systemd[19935]: This usually indicates unclean termination of a previous run, or service implementation deficiencies.
```

**Describe the results you expected:**

podman.service stops cleanly, and restarts work.

**Additional information you deem important (e.g. issue happens only occasionally):**

I noticed this when enabling the user-container tests in cockpit-podman in Fedora dist-git gating. 
[pull request](https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/9#)

As a workaround, I drop `KillMode` from the unit:
```
# avoid leftover user podman processes between login sessions
mkdir -p /etc/systemd/user/podman.service.d
printf '[Service]\nKillMode=\n' > /etc/systemd/user/podman.service.d/cleanup.conf
```

**Output of `podman version`:**
```
Version:      2.1.0-dev
API Version:  1
Go Version:   go1.14.4
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.0-dev
[...]

PR #7023, opened 2020-07-20
contrib/systemd cleanups

vrothberg:
This PR includes a series of fixes and clean ups for the systemd service and socket. Please refer to the individual commits for details.

Fixes: #7021


vrothberg:
@giuseppe @jwhonce @mheon  @rhatdan PTAL

contrib/systemd/system/podman.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Podman API Service
 Requires=podman.socket
-After=podman.socket
+Wants=network.target
+After=podman.socket network-online.target
```
martinpitt, line 5: As former systemd developer/maintainer I recommend to not do that. podman.service does *not* need networking at startup, only for some operations -- but that shouldn't affect the boot order. network-online.target is a rather terrible and brittle hack, see `man systemd.special`.

`Wants=network.target` is also unnecessary here. Your unit shouldn't activate that -- at most you'd have `After=network.target`, but that's not even required as podman does not require the network to shut down cleanly.

vrothberg:
Thanks for reviewing and the feedback, @martinpitt. Your comments are addressed in the latest state.

rhatdan:
@goochjj PTAL

TomSweeneyRedHat:
LGTM

goochjj:
LGTM

PR #7026, opened 2020-07-20
BATS tests: more resilient remove_same_dev_warning

edsantiago:
Some CI tests are flaking in the SELinux test, possibly because
there's a new variation of the "multiple devices" warning I hadn't
seen before:

   WARNING: Creating device "/dev/null" with same type, major and minor as existing "/dev/foodevdir/null".

Solution: in remove_same_dev_warning(), remove "multiple" from
the match string.

Also: fix a Go test that wasn't cleaning up after itself. And
add an actual test to it, not just check-exit-status.

### DIFF
--- a/test/e2e/run_device_test.go
+++ b/test/e2e/run_device_test.go
@@ -75,11 +75,17 @@ var _ = Describe("Podman run device", func() {
 
 	It("podman run device host device and container device parameter are directories", func() {
 		SkipIfRootless()
-		SystemExec("mkdir", []string{"/dev/foodevdir"})
-		SystemExec("mknod", []string{"/dev/foodevdir/null", "c", "1", "3"})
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Changes LGTM, but tests are somewhat less than happy.

mheon:
`[+0278s] Error: cannot stat /dev/foodevdir:/dev/bar: stat /dev/foodevdir: no such file or directory`

Coming out of Podman. It's wierd, considering that I'd expect the SystemExec earlier to make the device node to also fail, given it's in that directory - maybe it is and we're just not checking the exit code?

edsantiago:
I bet I shouldn't have added the `RemoveAll()` to `AfterEach()`. I bet it's some sort of parallelization issue. I'll see if I can do it as a `defer` instead.

PR #7028, opened 2020-07-21
docs: Fix formatting mistake

xordspar0:
Part of this section was a code block, and part of it was absorbed into the preceding normal paragraph.

Before/after:

image

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -1120,7 +1120,7 @@ required for VPN, without it containers need to be run with the --network=host f
 
 Environment variables within containers can be set using multiple different options:  This section describes the precedence.
 
-Precedence Order:
+Precedence order (later entries override earlier entries):
 
 - **--env-host** : Host environment of the process executing Podman is added.
 - Container image : Any environment variables specified in the container image.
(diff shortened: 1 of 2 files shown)


docs/source/markdown/podman-create.1.md
```diff
@@ -1121,13 +1121,11 @@ required for VPN, without it containers need to be run with the --network=host f
 Environment variables within containers can be set using multiple different options:  This section describes the precedence.
 
 Precedence Order:
```
mheon, line 1123: Suggestion: maybe we should add "from least to greatest" here to indicate what the actual ordering is.

mheon:
Can you make the same change to `podman-run.1.md` in the same directory?

rhatdan:
You should also mention 
--http-proxy=true|false
Which will set some environment variables as well.

rhatdan:
We really should have a test that makes sure the hierarchy works, and stays working.

docs/source/markdown/podman-create.1.md
xordspar0: Great idea.

xordspar0:
I've made most of the changes. I'm looking into how --http-proxy=true|false fits in now.

mheon:
LGTM once the HTTP proxy change makes it in.

xordspar0:
Done. Please take a look at what I did for the --http-proxy docs. I can squash these if needed, but I think it makes sense to keep them separate.

xordspar0:
For anyone curious, the --http-proxy and the --host-envs code are in the same function: https://github.com/containers/podman/blob/be5219ab9529c596421dd19bb61cdd2ced267ece/vendor/github.com/containers/common/pkg/config/config.go#L706-L722

rhatdan:
Thanks @xordspar0 for the PR.

PR #7031, opened 2020-07-21
unit tests: root check

vrothberg:
The unit tests currently require running as root.  This has caused some
confusion that justifies adding a root check to `make localunit` and
error out for non-root users instead of starting the tests deemed to
fail.

### DIFF
--- a/hack/check_root.sh
+++ b/hack/check_root.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+if ! [ $(id -u) = 0 ]; then
+   echo "Please run as root! '$@' requires root privileges."
+   exit 1
+fi
--- a/Makefile
+++ b/Makefile
@@ -307,6 +307,7 @@ testunit: libpodimage ## Run unittest on the built image
 
 .PHONY: localunit
 localunit: test/goecho/goecho varlink_generate
+	hack/check_root.sh make localunit
 	ginkgo \
 		-r \
 		$(TESTFLAGS) \


vrothberg:
@edsantiago @jwhonce @mheon @rhatdan PTAL

hack/check_root.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+if ! [ $(id -u) = 0 ]; then
```
edsantiago, line 2: It took me a long while to see the negation; this might be a little more readable
```bash
if [ "$(id -u)" != "0" ]; then
```

vrothberg:
> DAE find it a little alarming that unit tests won't run rootless?

It is and we need to address this. This change will make it more explicit.

hack/check_root.sh
vrothberg: Done, thanks for checking!

rhatdan: I just found yesterday that bats sets an environment varible EUID, which means you would not need to execute a program

if [ "${EUID}" != "0" ]; then

rhatdan:
/lgtm
But I would like to get some of the unit tests to run as rootless.

vrothberg:
> /lgtm
> But I would like to get some of the unit tests to run as rootless.

I created a Jira card for that :+1:

hack/check_root.sh
edsantiago: (bash, not bats). But yes, that's a good point. I've always been a little reluctant to trust shell-defined variables (`$HOME`, `$LOGNAME`) but this one is explicitly documented as readonly in bash(1) so I think it's ok. For future reference, since this just merged.

PR #7040, opened 2020-07-21
Do not print an error message on non-0 exec exit code

mheon:
This was added with an earlier exec rework, and honestly is very confusing. Podman is printing an error message, but the error had nothing to do with Podman; it was the executable we ran inside the container that errored, and per `podman run` convention we should set the Podman exit code to the process's exit code and print no error.

### DIFF
--- a/libpod/container_exec.go
+++ b/libpod/container_exec.go
@@ -729,10 +729,6 @@ func (c *Container) Exec(config *ExecConfig, streams *define.AttachStreams, resi
 		return -1, err
 	}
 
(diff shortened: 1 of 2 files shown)


jwhonce:
LGTM

mheon:
I'm kind of expecting some test failures out of healthchecks, this happened the last time I tried to remove this.

rhatdan:
LGTM, fix the tests, this looks to be the correct behaviour.

vrothberg:
Restarted the flaked test

PR #7047, opened 2020-07-22
CI: attempt to fix flake in login test

edsantiago:
Fixes: #5212

...or at least I hope it does. The symptom seems to be that
INTEGRATION_ROOT is not being defined in some code flows.
This PR blindly implements a suggestion from Miloslav,
setting INTEGRATION_ROOT in one more place.

We won't actually know for a long time if this works or
not, because the test failure is a flake.

Signed-off-by: Ed Santiago <santiago@redhat.com>

Issue #5212, opened 2020-02-14
CI flakes with podman-login


vrothberg:
I hit podman-login flakes quite frequently in the passed which suggests to have a look at it and track down potential sources for the flakes.

[+1077s] [Fail] Podman login and logout [BeforeEach] podman login and logout with flag --authfile 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:58
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:101
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout with --tls-verify 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:157
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout with flag --authfile 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:135
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:105
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout with flag --authfile 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:135
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:105
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout with --cert-dir 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:177
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout with --cert-dir 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:177
[+1077s] 
[+1077s] [Fail] Podman login and logout [It] podman login and logout with --cert-dir 
[+1077s] /var/tmp/go/src/github.com/containers/libpod/test/e2e/login_logout_test.go:177

rhatdan:
@vrothberg Any movement on this issue?

[...]

PR #7048, opened 2020-07-22
Turn on a bunch more remote tests

rhatdan:
We need to be more specific about the remote tests we turn off.

### DIFF
--- a/test/e2e/port_test.go
+++ b/test/e2e/port_test.go
@@ -1,5 +1,3 @@
-// +build !remote
-
 package integration
 
 import (
@@ -49,6 +47,7 @@ var _ = Describe("Podman port", func() {
 	})
 
 	It("podman port -l nginx", func() {
+		SkipIfRemote()
 		session, cid := podmanTest.RunNginxWithHealthCheck("")
 		Expect(session.ExitCode()).To(Equal(0))
 
@@ -64,6 +63,7 @@ var _ = Describe("Podman port", func() {
 	})
 
 	It("podman container port  -l nginx", func() {
+		SkipIfRemote()
 		session, cid := podmanTest.RunNginxWithHealthCheck("")
 		Expect(session.ExitCode()).To(Equal(0))
 
@@ -79,6 +79,7 @@ var _ = Describe("Podman port", func() {
 	})
 
 	It("podman port -l port nginx", func() {
+		SkipIfRemote()
 		session, cid := podmanTest.RunNginxWithHealthCheck("")
 		Expect(session.ExitCode()).To(Equal(0))
 
@@ -127,18 +128,18 @@ var _ = Describe("Podman port", func() {
 		lock2 := GetPortLock("5001")
 		defer lock2.Unlock()
 
-		setup := podmanTest.Podman([]string{"run", "-dt", "-p", "5000:5000", "-p", "5001:5001", ALPINE, "top"})
+		setup := podmanTest.Podman([]string{"run", "--name", "test", "-dt", "-p", "5000:5000", "-p", "5001:5001", ALPINE, "top"})
 		setup.WaitWithDefaultTimeout()
 		Expect(setup.ExitCode()).To(BeZero())
 
 		// Check that the first port was honored
-		result1 := podmanTest.Podman([]string{"port", "-l", "5000"})
+		result1 := podmanTest.Podman([]string{"port", "test", "5000"})
 		result1.WaitWithDefaultTimeout()
 		Expect(result1.ExitCode()).To(BeZero())
 		Expect(result1.LineInOuputStartsWith("0.0.0.0:5000")).To(BeTrue())
 
 		// Check that the second port was honored
-		result2 := podmanTest.Podman([]string{"port", "-l", "5001"})
+		result2 := podmanTest.Podman([]string{"port", "test", "5001"})
 		result2.WaitWithDefaultTimeout()
 		Expect(result2.ExitCode()).To(BeZero())
 		Expect(result2.LineInOuputStartsWith("0.0.0.0:5001")).To(BeTrue())
--- a/test/e2e/run_entrypoint_test.go
+++ b/test/e2e/run_entrypoint_test.go
@@ -1,5 +1,3 @@
-// +build !remote
-
 package integration
 
 import (
@@ -46,6 +44,7 @@ CMD []
 	})
 
 	It("podman run entrypoint", func() {
+		SkipIfRemote()
 		dockerfile := `FROM docker.io/library/alpine:latest
 ENTRYPOINT ["grep", "Alpine", "/etc/os-release"]
 `
@@ -57,6 +56,7 @@ ENTRYPOINT ["grep", "Alpine", "/etc/os-release"]
 	})
 
 	It("podman run entrypoint with cmd", func() {
+		SkipIfRemote()
 		dockerfile := `FROM docker.io/library/alpine:latest
 CMD [ "-v"]
 ENTRYPOINT ["grep", "Alpine", "/etc/os-release"]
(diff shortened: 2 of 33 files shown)


rhatdan:
This PR will turn on another 100+ tests for podman --remote.

@edsantiago @vrothberg @giuseppe @mheon @baude @jwhonce @TomSweeneyRedHat @ashley-cui @QiWang19 PTAL

BTW, I think podman remote is not able to see images that are stored as localhost/image.

This is why the builds followed by runs are failing.

test/e2e/generate_systemd_test.go
```diff
@@ -1,5 +1,3 @@
-// +build !remote
```
edsantiago, line 1: Honestly, is there any point to enabling this one?

vrothberg: `generate systemd` isn't supported for remote clients, so I think we can leave it fully disabled

test/e2e/run_entrypoint_test.go
```diff
@@ -46,6 +44,7 @@ CMD []
 	})
 
 	It("podman run entrypoint", func() {
+		SkipIfRemote()
```
edsantiago, line 47: I'm not sure why this and the subsequent ones are skipped? They're all tagged as `foobar.com/something` so the `localhost` issue shouldn't apply. And I don't understand what you mean by `localhost` anyway, I'm playing with podman-remote on my system (with master) and am not having problems building, tagging, or running images with unqualified names...?

rhatdan: I just wanted to get as many in as possible.  I did not review why they were blowing up. I was hoping to assign each one to others who are closer to the issue, to investigate dropping the SkipIfRemote().

test/e2e/generate_systemd_test.go
rhatdan: Reverted.

rhatdan:
Once we have this merged, we need to open a card for each test*go for someone to examine the skipremote tests to see if they could pass on remote systems.

edsantiago:
I opened RUN-973 this morning for that, although it's not as fine-grain as your idea.

One suggestion: if more tests choke, could you add a one-liner comment next to `SkipIfRemote` indicating (very briefly) the nature of the failure? That could make it easier to find common failure causes without doing full runs. Thanks.

Given the understanding that these skips are temporary, this PR LGTM.

edsantiago:
[CI error](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4561900547604480/html/integration_test.log.html) is a flake, #5212, and is exactly what I'm trying to address with #7047.

test/e2e/port_test.go
```diff
@@ -127,18 +128,18 @@ var _ = Describe("Podman port", func() {
...
 		defer lock2.Unlock()
 
-		setup := podmanTest.Podman([]string{"run", "-dt", "-p", "5000:5000", "-p", "5001:5001", ALPINE, "top"})
+		setup := podmanTest.Podman([]string{"run", "--name", "test", "-dt", "-p", "5000:5000", "-p", "5001:5001", ALPINE, "top"})
 		setup.WaitWithDefaultTimeout()
 		Expect(setup.ExitCode()).To(BeZero())
 
 		// Check that the first port was honored
-		result1 := podmanTest.Podman([]string{"port", "-l", "5000"})
+		result1 := podmanTest.Podman([]string{"port", "test", "5000"})
```
TomSweeneyRedHat, line 136: should this still be `-l`?

test/e2e/run_networking_test.go
```diff
@@ -265,10 +263,10 @@ var _ = Describe("Podman run networking", func() {
 	})
 
 	It("podman run network expose ports in image metadata", func() {
-		session := podmanTest.Podman([]string{"create", "-dt", "-P", nginx})
+		session := podmanTest.Podman([]string{"create", "--name", "test", "-dt", "-P", nginx})
 		session.Wait(90)
 		Expect(session.ExitCode()).To(Equal(0))
-		results := podmanTest.Podman([]string{"inspect", "-l"})
+		results := podmanTest.Podman([]string{"inspect", "test"})
```
TomSweeneyRedHat, line 269: Should this really change?

test/e2e/run_restart_test.go
```diff
@@ -35,11 +33,12 @@ var _ = Describe("Podman run restart containers", func() {
...
 
 	It("Podman start after successful run", func() {
-		session := podmanTest.Podman([]string{"run", ALPINE, "ls"})
+		SkipIfRemote()
+		session := podmanTest.Podman([]string{"run", "--name", "test", ALPINE, "ls"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 
-		session2 := podmanTest.Podman([]string{"start", "--attach", "--latest"})
+		session2 := podmanTest.Podman([]string{"start", "--attach", "test"})
```
TomSweeneyRedHat, line 41: another latest question

test/e2e/port_test.go
edsantiago: `-l` doesn't work with `remote`

test/e2e/top_test.go
```diff
@@ -55,22 +53,22 @@ var _ = Describe("Podman top", func() {
 	})
 
 	It("podman top on container", func() {
-		session := podmanTest.Podman([]string{"run", "-d", ALPINE, "top", "-d", "2"})
+		session := podmanTest.Podman([]string{"run", "--name", "test", "-d", ALPINE, "top", "-d", "2"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 
-		result := podmanTest.Podman([]string{"top", "-l"})
+		result := podmanTest.Podman([]string{"top", "test"})
```
TomSweeneyRedHat, line 60: again latest

test/e2e/port_test.go
TomSweeneyRedHat: I think it's important to have the latest tested still then.  Can we spin up new tests that skip remote with latest still?

edsantiago: Can you clarify whether you mean "it is important to test -l", or "it is important to test -l in this precise exact port test"? There are still a ton of `-l` tests.

edsantiago:
Failure ([log](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5687800454447104/html/integration_test.log.html)) seems to be consistent, in [It] podman rm -v removes anonymous volume : 
```
podman [options]-remote --remote --url unix:/... volume list --quiet
b959d6c406db5ae242f34ded602dd6bb05ab631529c583b9b9a4122352939d1e    <<---- expected empty string
```
Recommendation: disable that test with a `SkipIfRemote()  // TODO: remote rm -v does not seem to work`

test/e2e/run_restart_test.go
rhatdan: Sure if we want to have one test to test --latest option, but I believe there is more value in testing these tests as remote.

test/e2e/run_networking_test.go
rhatdan: Yes, we do we care about testing the -l or --latest inspect flag when testing networking code.  We want to test the networking code for remote.

test/e2e/port_test.go
rhatdan: We should probably have one test for each command on --latest but no more.  Just like we should test every flag.

rhatdan: I just checked there are three other tests in port_test.go that test `-l`, I would argue that we don't need this many, but these tests had -l in their test name, so I left them and skipped local.

TomSweeneyRedHat: @edsantiago  My main thing is we've coverage.  After hearing our potential low coverage numbers the other day, I hate to remove tests that potentially keep our numbers up a bit.  Not so much these particular tests, but I'm just overall leery about removing potential branches of testing without at least verifying we still have coverage.

rhatdan:
Well we actually have very good testing in the integration and e2e,  Where we don't have good coverage is on the unit level.  Especially in libpod.  Basically we need to test all of the external functions,with unit tests.

PR #7065, opened 2020-07-23
WIP - code coverage

vrothberg:
Work in progress to get code coverage into the CI.

Unit tests are not sufficient as most code is tested via the e2e tests (and the system tests).

Working:
* unit tests
* system tests


test/system/helpers.bash
```diff
@@ -137,11 +137,16 @@ function run_podman() {
...
     esac
 
+    local BIN=$PODMAN
+    if [[ ! -z "$COVERAGE" ]]; then
+        BIN="$PODMAN -test.coverprofile=coverprofile.system.$(random_string 20) COVERAGE"
+    fi
+
     # stdout is only emitted upon error; this echo is to help a debugger
-    echo "$_LOG_PROMPT $PODMAN $*"
+    echo "$_LOG_PROMPT $BIN $@"
```
edsantiago, line 146: It's kind of complicated, and I'll explain one day over a beer, but please leave that as dollar-asterisk

test/system/helpers.bash
```diff
@@ -137,11 +137,16 @@ function run_podman() {
         '?')             expected_rc=  ; shift;;  # ignore exit code
     esac
 
+    local BIN=$PODMAN
+    if [[ ! -z "$COVERAGE" ]]; then
```
edsantiago, line 141: I find positive tests more readable than negative: `-n` is better here
```bash
    if [[ -n "$COVERAGE" ]]; then
```

test/system/helpers.bash
vrothberg: Yes and yes :-) Thanks for checking!

test/system/helpers.bash
vrothberg: Much better, thanks!

vrothberg:
One step further: e2e tests can now run with coverage as well. Maybe needs more massaging until they turn green but I am optimistic.

vrothberg:
What's going on here? I just pushed...

vrothberg:
Eeeek. I don't see it merged in master, which is good.

vrothberg:
Nice...integration are green now. Only system tests are going south. So close...

Makefile
```diff
@@ -36,6 +35,15 @@ PKG_MANAGER ?= $(shell command -v dnf yum|head -n1)
 # ~/.local/bin is not in PATH on all systems
 PRE_COMMIT = $(shell command -v bin/venv/bin/pre-commit ~/.local/bin/pre-commit pre-commit | head -n1)
 
+COVERAGE_PATH ?= $(shell pwd)/.coverage
+COVERAGE_PROFILE ?= $(shell pwd)/coverage.txt
+export COVERAGE_PATH
+export COVERAGE_PROFILE
+$(shell mkdir -p ${COVERAGE_PATH})
+
+CODECOVE_TOKEN=52d2da88-a97c-442b-8c9f-1f77a5a26b7c
```
edsantiago, line 44: Yikes. Shouldn't this be a Cirrus secret?

edsantiago:
I gave this a good try: perl/awk with `seen` hashing is much slower than `sort -u` (like, 70s vs 45s for ~1k cover files). I tried perusing the files manually to see if my eye could detect useful patterns for a custom tool. No luck. This is a pretty hard problem.

Makefile
vrothberg: Absolutely. I can remove it now and generate a new token if you want to. I thought it's fine for testing purposes.

vrothberg:
> I gave this a good try: perl/awk with `seen` hashing is much slower than `sort -u` (like, 70s vs 45s for ~1k cover files). I tried perusing the files manually to see if my eye could detect useful patterns for a custom tool. No luck. This is a pretty hard problem.

Thanks a lot for checking, Ed, I appreciate it!

The last remaining idea I have is to not fully "divide and conquer" but to stop at a certain threshold. Instead of merging _N_ profiles into 1, we could merge _N_ profiles into _M_ and upload them. The upload script supports that and Codecov is apparently smart enough to combine multiple uploads on the server side.

I guess the crux is that we don't need to find the optimum. We're doing a `sort -u` to reduce redundancy and hence trim down the size. AFAIK Codecov has a size limit 100 MB per uploaded file. We could short cut with something similar to a `sort -u --until=100MB` where we're sorting until we hit the specified threshold.

Simple to code in a high-level language but I didn't find a way with coreutils.

vrothberg:
@edsantiago, I regenerated (and removed) the token. I also wrote a small go script which merges the profiles in 26 secs on my machine:
```
libpod (test-coverage) $ time go run hack/coverage/coverage.go -dir ~/profiles -profile coverag
e.txt
Merged 6776 profiles (11.6 GB) into "coverage.txt" (3.5 MB)

real    0m25.975s
user    0m25.246s
sys     0m4.490s
```

contrib/cirrus/coverage.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+source $(dirname $0)/lib.sh
+
+req_env_var GOSRC
```
cevich, line 7: I would recommend adding 'CODECOV_TOKEN' to the line above.

vrothberg: Done, thanks!

vrothberg:
Closing for now. It's a massive amount of work and I don't see it getting a higher priority in the near future.

PR #7073, opened 2020-07-23
podman save use named pipe

### LINKED ISSUE #7017: podman save --format=oci-archive fails to write to a pipe

/kind bug

**Description**

I wanted a quick and dirty way to copy an image from one computer to another. The manual for `podman save` says that it writes to `stdout` by default and the manual for `podman load` says that it reads from `stdin` by default so I assumed that piping the output of `podman save` to the destination via ssh would work, but podman save --format=oci-archive fails to write to a pipe.

1. `C="$(buildah from scratch)"`

2. `buildah commit "$C" bug`

3. `podman save --format=oci-archive bug | cat >/dev/null`

Error: error getting OCI archive ImageReference for ("/dev/stdout", "localhost/bug"): lstat /proc/2102/fd/pipe:[34531]: no such file or directory

The image archive is written to stdout.

- This bug happens every time
- Without --format=oci-archive the command does not error

```
Version:      2.0.2
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 03:00:00 1970
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.18-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.18, commit: 6e8799f576f11f902cd8a8d8b45b2b2caf636a85'
  cpus: 2
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: hostname
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 4.19.132-1.pvops.qubes.x86_64
  linkmode: dynamic
  memFree: 106500096
  memTotal: 479236096
  ociRuntime:
    name: crun
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      spec: 1.0.0
  os: linux
  remoteSocket:
  rootless: true
  slirp4netns:
    version: |-
      slirp4netns version 1.1.4
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 2
  swapFree: 997453824
  swapTotal: 1073737728
  uptime: 14m 16.04s
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - docker.io
store:
  configFile: /home/user/.config/containers/storage.conf
  containerStore:
    number: 4
    paused: 0
    running: 0
    stopped: 4
  graphDriverName: overlay
  graphOptions:
    overlay.mount_program:
      Version: |-
        fusermount3 version: 3.9.1
        fuse-overlayfs: version 1.1.0
        FUSE library version 3.9.1
  graphRoot: /home/user/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 5
  volumePath: /home/user/.local/share/containers/storage/volumes
version:
  APIVersion: 1
  Built: 0
  BuiltTime: Thu Jan  1 03:00:00 1970
  GitCommit: ""
  GoVersion: go1.14.3
  OsArch: linux/amd64
  Version: 2.0.2

```

```
podman-2.0.2-1.fc32.x86_64
```

QubesOS https://github.com/QubesOS

QiWang19:
podman save uses named pipe as output path, not directly using /dev/stdout.
fix #7017

### DIFF
--- a/cmd/podman/images/save.go
+++ b/cmd/podman/images/save.go
@@ -5,10 +5,9 @@ import (
 	"os"
 	"strings"
 
-	"github.com/containers/podman/v2/libpod/define"
-
 	"github.com/containers/podman/v2/cmd/podman/parse"
 	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/pkg/errors"
@@ -83,9 +82,10 @@ func saveFlags(flags *pflag.FlagSet) {
 
 }
 
-func save(cmd *cobra.Command, args []string) error {
+func save(cmd *cobra.Command, args []string) (finalErr error) {
 	var (
-		tags []string
+		tags      []string
+		succeeded = false
 	)
 	if cmd.Flag("compress").Changed && (saveOpts.Format != define.OCIManifestDir && saveOpts.Format != define.V2s2ManifestDir && saveOpts.Format == "") {
 		return errors.Errorf("--compress can only be set when --format is either 'oci-dir' or 'docker-dir'")
@@ -95,7 +95,22 @@ func save(cmd *cobra.Command, args []string) error {
 		if terminal.IsTerminal(int(fi.Fd())) {
 			return errors.Errorf("refusing to save to terminal. Use -o flag or redirect")
 		}
-		saveOpts.Output = "/dev/stdout"
+		pipePath, cleanup, err := setupPipe()
+		if err != nil {
+			return err
+		}
+		if cleanup != nil {
+			defer func() {
+				errc := cleanup()
+				if succeeded {
+					writeErr := <-errc
+					if writeErr != nil && finalErr == nil {
+						finalErr = writeErr
+					}
+				}
+			}()
+		}
+		saveOpts.Output = pipePath
 	}
 	if err := parse.ValidateFileName(saveOpts.Output); err != nil {
 		return err
@@ -103,5 +118,9 @@ func save(cmd *cobra.Command, args []string) error {
 	if len(args) > 1 {
 		tags = args[1:]
 	}
-	return registry.ImageEngine().Save(context.Background(), args[0], tags, saveOpts)
+	err := registry.ImageEngine().Save(context.Background(), args[0], tags, saveOpts)
+	if err == nil {
+		succeeded = true
+	}
+	return err
 }
--- a/cmd/podman/images/utils_linux.go
+++ b/cmd/podman/images/utils_linux.go
@@ -1,0 +1,47 @@
+package images
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// setupPipe for fixing https://github.com/containers/podman/issues/7017
+// uses named pipe since containers/image EvalSymlinks fails with /dev/stdout
+// the caller should use the returned function to clean up the pipeDir
+func setupPipe() (string, func() <-chan error, error) {
+	errc := make(chan error)
+	pipeDir, err := ioutil.TempDir(os.TempDir(), "pipeDir")
+	if err != nil {
+		return "", nil, err
+	}
+	pipePath := filepath.Join(pipeDir, "saveio")
+	err = unix.Mkfifo(pipePath, 0600)
+	if err != nil {
+		if e := os.RemoveAll(pipeDir); e != nil {
+			logrus.Errorf("error removing named pipe: %q", e)
+		}
+		return "", nil, errors.Wrapf(err, "error creating named pipe")
+	}
+	go func() {
+		fpipe, err := os.Open(pipePath)
+		if err != nil {
+			errc <- err
+			return
+		}
+		_, err = io.Copy(os.Stdout, fpipe)
+		fpipe.Close()
+		errc <- err
+	}()
+	return pipePath, func() <-chan error {
+		if e := os.RemoveAll(pipeDir); e != nil {
+			logrus.Errorf("error removing named pipe: %q", e)
+		}
+		return errc
+	}, nil
+}
--- a/cmd/podman/images/utils_unsupported.go
+++ b/cmd/podman/images/utils_unsupported.go
@@ -1,0 +1,7 @@
+// +build !linux
+
+package images
+
+func setupPipe() (string, func() <-chan error, error) {
+	return "/dev/stdout", nil, nil
+}
--- a/test/system/120-load.bats
+++ b/test/system/120-load.bats
@@ -26,6 +26,16 @@ verify_iid_and_name() {
     is "$new_img_name" "$1"   "Name & tag of restored image"
 }
 
+@test "podman save to pipe and load" {
+    # We can't use run_podman because that uses the BATS 'run' function
+    # which redirects stdout and stderr. Here we need to guarantee
+    # that podman's stdout is a pipe, not any other form of redirection
+    $PODMAN save --format oci-archive $IMAGE | cat >$PODMAN_TMPDIR/test.tar
+    [ $status -eq 0 ]
+
+    run_podman load -i $PODMAN_TMPDIR/test.tar
+}
+
 
 @test "podman load - by image ID" {
     # FIXME: how to build a simple archive instead?


QiWang19:
@mtrmac I'm not sure if this works. Could you PTAL?

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("--compress can only be set when --format is either 'oci-dir' or 'docker-dir'")
 	}
 	if len(saveOpts.Output) == 0 {
-		fi := os.Stdout
-		if terminal.IsTerminal(int(fi.Fd())) {
-			return errors.Errorf("refusing to save to terminal. Use -o flag or redirect")
+		pipeFile := "/tmp/saveio"
```
mtrmac, line 99: This must not use a predictable name; generate a random one similarly to `mktemp`.

… and there probably isn’t a helper to do that in the standard library. 😢 Using `ioutil.TempFile`+`os.Remove`+`Mkfifo` would not be safe.

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("--compress can only be set when --format is either 'oci-dir' or 'docker-dir'")
 	}
 	if len(saveOpts.Output) == 0 {
-		fi := os.Stdout
-		if terminal.IsTerminal(int(fi.Fd())) {
```
mtrmac, line 95: This should probably remain.

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
...
-			return errors.Errorf("refusing to save to terminal. Use -o flag or redirect")
+		pipeFile := "/tmp/saveio"
+		saveOpts.Output = pipeFile
+		err := os.Remove(pipeFile)
+		if err != nil && !os.IsNotExist(err) {
+			return errors.Wrapf(err, "error removing named pipe")
 		}
-		saveOpts.Output = "/dev/stdout"
+		err = syscall.Mkfifo(pipeFile, 0666)
```
mtrmac, line 105: 0600

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return errors.Wrapf(err, "error creating named pipe")
+		}
+		defer os.Remove(pipeFile)
+		wg := sync.WaitGroup{}
+		wg.Add(1)
```
mtrmac, line 111: Nothing uses this `WaitGroup` yet; eventually, the caller should block on this goroutine — but it should probably also collect an `error`, so a `chan error` might be sufficient.

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
...
+		go func() {
+			fpipe, _ := os.OpenFile(pipeFile, os.O_RDWR, os.ModeNamedPipe)
```
mtrmac, line 113: `OpenFile` and the mode is only necessary when creating a file, and file types can’t be used for the mode parameter anyway.

Code using pipes is also easier to read if it is clear whether it is opening for reading/writing.

So, just `os.Open(pipeFile)` I think.

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
...
+			defer fpipe.Close()
+			for {
+				reader := bufio.NewReader(fpipe)
+				_, err = io.Copy(os.Stdout, reader)
+				if err != nil {
+					logrus.Errorf("error copying data to /dev/stdout, %v", err)
+					break
+				}
+				time.Sleep(100 * time.Millisecond)
```
mtrmac, line 122: This shouldn’t be necessary. Instead, open a write-only file descriptor before creating this goroutine (then reads don’t fail with EOF; multiple write-only file descriptors for a single pipe are OK, as long as they don’t write simultaneously), and close that file descriptor only when the export is done to trigger EOF in the goroutine.

cmd/podman/images/save.go
```diff
@@ -91,11 +96,33 @@ func save(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return errors.Wrapf(err, "error creating named pipe")
+		}
+		defer os.Remove(pipeFile)
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		go func() {
+			fpipe, _ := os.OpenFile(pipeFile, os.O_RDWR, os.ModeNamedPipe)
+			defer fpipe.Close()
+			for {
```
mtrmac, line 115: How does this loop ever terminate? Not on EOF, not on removing `pipeFile`. So, only on process exit, I think — but the main goroutine does not know when this one is done.

cmd/podman/images/save.go
giuseppe: I don't think we should delete an existing file as it could be used by another podman instance.  Wouldn't it be enough to generate a random name and try again on `EEXIST`?

edsantiago: One standard pattern is to `mktemp` a directory, not a file, and create the pipe/socket/whatever therein. Implementation and cleanup are left as an exercise for the reader.

mtrmac: Oh, that’s a great trick! Good idea.

cmd/podman/images/save.go
```diff
@@ -91,11 +97,46 @@ func save(cmd *cobra.Command, args []string) error {
...
+		pipeDir, err := ioutil.TempDir(os.TempDir(), "pipeDir")
+		if err != nil {
+			return err
+		}
+		saveOpts.Output = filepath.Join(pipeDir, pipeFile)
+		err = syscall.Mkfifo(saveOpts.Output, 0600)
+		if err != nil {
+			return errors.Wrapf(err, "error creating named pipe")
+		}
+		defer os.Remove(saveOpts.Output)
```
mtrmac, line 110: All of `pipeDir` should be removed.

cmd/podman/images/save.go
```diff
@@ -91,11 +97,46 @@ func save(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("--compress can only be set when --format is either 'oci-dir' or 'docker-dir'")
 	}
 	if len(saveOpts.Output) == 0 {
-		fi := os.Stdout
-		if terminal.IsTerminal(int(fi.Fd())) {
+		pipeFile := "saveio"
```
mtrmac, line 100: (Absolutely non-blocking: This has a single user and can be inlined — or make the file name a top-level constant.)

cmd/podman/images/save.go
```diff
@@ -91,11 +97,46 @@ func save(cmd *cobra.Command, args []string) error {
...
+			return errors.Wrapf(err, "error creating named pipe")
+		}
+		defer os.Remove(saveOpts.Output)
+		errc := make(chan error)
+		fpipe, err := os.OpenFile(saveOpts.Output, os.O_RDWR, os.ModeNamedPipe)
+		defer fpipe.Close()
+		if err != nil {
+			return err
+		}
+		if terminal.IsTerminal(int(fpipe.Fd())) {
```
mtrmac, line 117: This should be testing the original standard output (we _know_ a FIFO is not a TTY).

cmd/podman/images/save.go
```diff
@@ -91,11 +97,46 @@ func save(cmd *cobra.Command, args []string) error {
...
+			for {
+				reader := bufio.NewReader(fpipe)
+				_, err = io.Copy(os.Stdout, reader)
+				if err != nil {
+					errc <- err
+					logrus.Errorf("error copying data to /dev/stdout, %v", err)
+					break
+				}
+			}
+			errc <- err
```
mtrmac, line 130: This is unreachable. (I can’t see why the `for { }` is necessary — if `io.Copy` gets an EOF, it should be done.)

cmd/podman/images/save.go
```diff
@@ -91,11 +97,46 @@ func save(cmd *cobra.Command, args []string) error {
...
+		}()
+		select {
+		case err := <-errc:
```
mtrmac, line 133: This will about never trigger here AFAICS; this should happen _after_ sending all the output, and be blocking (modulo timeouts/cancellation) at that point (but it can only be blocking if `fpipe` is closed before, to trigger EOF in the goroutine).

QiWang19:
The documentation exampl, are you suggesting this should fail?
 ```
$ podman save > alpine-all.tar alpine
```

**Comment by mtrmac** [comment]
> The documentation exampl, are you suggesting this should fail?
> 
> ```
> $ podman save > alpine-all.tar alpine
> ```

I’m not quite sure what this refers to, but no, not in general. I was some ways to trigger failures that should be detected (note that `/dev/full` is a specially behaving device, not a random meaningless name).

File: cmd/podman/images/save.go (hunk 12 above)
Comment (QiWang19) [reply cmd/podman/images/save.go:129]: The `io.Copy` does not return EOF error. How can I trigger that?

Comment (mtrmac) [reply cmd/podman/images/save.go:129]: Pipes return EOF when there are no descriptors open for writing.

Comment (mtrmac) [reply cmd/podman/images/save.go:129]: I’m sorry, I have finally tried this myself, and it turns out I have missed an important point: `open(2)` blocks _either_ the reader or writer until both are present. So we don’t need (and in fact can’t create) the long-term writer descriptor. OTOH that makes the code even simpler:


```go
package main

import (
	"fmt"
	"io"
	"io/ioutil"
	"os"
	"path/filepath"
	"syscall"
)

func fatal(err error) {
	fmt.Fprintf(os.Stderr, "Error: %v", err)
	os.Exit(1)
}

func pipeCopyGoroutine(pipePath string, errc chan<- error) {
	fmt.Fprintf(os.Stderr, "||Opening %s for reading ...\n", pipePath)
	pipeReader, err := os.Open(pipePath)
	if err != nil {
		fatal(err)
	}
	fmt.Fprintf(os.Stderr, "||Read open, copying...\n")
	n, err := io.Copy(os.Stdout, pipeReader)
	fmt.Fprintf(os.Stderr, "||Copied (%v, %v), closing reader... \n", n, err)
	pipeReader.Close()
	fmt.Fprintf(os.Stderr, "||Sending status...\n")
	errc <- err
	fmt.Fprintf(os.Stderr, "||Status sent, goroutine done\n")
}

func setupPipe() (string, func() error) {
	pipeDir, err := ioutil.TempDir(os.TempDir(), "pipeDir")
	if err != nil {
		fatal(err)
	}
	pipePath := filepath.Join(pipeDir, "saveio")
	err = syscall.Mkfifo(pipePath, 0600)
	if err != nil {
		fatal(err)
	}
	fmt.Fprintf(os.Stderr, "Created %s\n", pipePath)

	errc := make(chan error)
	go pipeCopyGoroutine(pipePath, errc)
	return pipePath, func() error {
		fmt.Fprintf(os.Stderr, "Waiting for goroutine...\n")
		err := <-errc
		fmt.Fprintf(os.Stderr, "Goroutine returned %v\n", err)
		os.RemoveAll(pipeDir)
		return err
	}
}

func main() {
//	contents := []byte("Hello pipe\n")
	contentsLen := 1024*1024
	contents := make([]byte, contentsLen, contentsLen)
	for i := 0; i < contentsLen; i+= 1 { contents[i] = 'X' }
	contents[contentsLen-1] = '\n'


	pipePath, cleanup := setupPipe()
	fmt.Fprintf(os.Stderr, "Writing to pipe ...\n")
	ioutil.WriteFile(pipePath, contents, 0600)
	fmt.Fprintf(os.Stderr, "Wrote to pipe, cleaning up...\n")
	err := cleanup()
	fmt.Fprintf(os.Stderr, "Done, error %v\n", err)
}
```
(the debug logging is probably excessive; OTOH do note that this is missing almost all the necessary error handling).

E.g.:
```console
$ go run f.go | sed 's/^XX.*XX$/XX...XX/'
Created /tmp/pipeDir917705942/saveio
Writing to pipe ...
||Opening /tmp/pipeDir917705942/saveio for reading ...
||Read open, copying...
Wrote to pipe, cleaning up...
Waiting for goroutine...
||Copied (1048576, <nil>), closing reader... 
||Sending status...
||Status sent, goroutine done
Goroutine returned <nil>
Done, error <nil>
XX...XX
$
$ go run f.go > /dev/full
Created /tmp/pipeDir330418467/saveio
Writing to pipe ...
||Opening /tmp/pipeDir330418467/saveio for reading ...
||Read open, copying...
||Copied (0, write /dev/stdout: no space left on device), closing reader... 
||Sending status...
Wrote to pipe, cleaning up...
Waiting for goroutine...
Goroutine returned write /dev/stdout: no space left on device
Done, error write /dev/stdout: no space left on device
$
$ go run f.go | (read -rN 1; exec <&-; sleep 5) | sed 's/^XX.*XX$/XX...XX/'
Created /tmp/pipeDir557014252/saveio
Writing to pipe ...
||Opening /tmp/pipeDir557014252/saveio for reading ...
||Read open, copying...
signal: broken pipe
$
$ go run f.go | (sleep 5; cat) | wc
Created /tmp/pipeDir869522253/saveio
Writing to pipe ...
||Opening /tmp/pipeDir869522253/saveio for reading ...
||Read open, copying...
Wrote to pipe, cleaning up...
Waiting for goroutine...
||Copied (1048576, <nil>), closing reader... 
||Sending status...
||Status sent, goroutine done
Goroutine returned <nil>
Done, error <nil>
      1       1 1048576
```

Note that the short-read test requires content larger than the kernel’s pipe buffer, i.e. the short “Hello pipe” content is ~always written successfully even if the consumer ignores it.

QiWang19:
@mtrmac PTAL

cmd/podman/images/save.go
```diff
@@ -95,13 +98,24 @@ func save(cmd *cobra.Command, args []string) error {
...
 	if len(args) > 1 {
 		tags = args[1:]
 	}
-	return registry.ImageEngine().Save(context.Background(), args[0], tags, saveOpts)
+	err = registry.ImageEngine().Save(context.Background(), args[0], tags, saveOpts)
+	if err != nil {
+		return err
+	}
+	if cleanup != nil {
+		defer cleanup()
```
mtrmac, line 118: No, `defer` just after a successful `setupPipe()` - so that e.g. a `ValidateFileName` failure does not cause the command to exit without removing the temporary directory. And make sure the command does not hang in that case.

cmd/podman/images/save.go
```diff
@@ -95,7 +99,16 @@ func save(cmd *cobra.Command, args []string) error {
 		if terminal.IsTerminal(int(fi.Fd())) {
 			return errors.Errorf("refusing to save to terminal. Use -o flag or redirect")
 		}
-		saveOpts.Output = "/dev/stdout"
+		pipePath, cleanup, err = setupPipe()
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err = cleanup(); err != nil {
```
mtrmac, line 107: AFAICS `cleanup()` blocks until the pipe is opened and closed, which may never happen on some of the error paths, and this would just hang.

Maybe use `context.Context` to terminate that wait, or just expose `errc` and have this function inject a fallback “wait canceled” error that is then ignored in favor of the real cause.

QiWang19: @mtrmac PTAL

cmd/podman/images/save.go
```diff
@@ -95,7 +99,17 @@ func save(cmd *cobra.Command, args []string) error {
...
 			return errors.Errorf("refusing to save to terminal. Use -o flag or redirect")
 		}
-		saveOpts.Output = "/dev/stdout"
+		pipePath, cleanup, err = setupPipe()
+		if err != nil {
+			return err
+		}
+		defer func() {
+			errc := cleanup()
+			if err := <-errc; err != nil && err.Error() != "wait canceled" {
```
rhatdan, line 108: Can you make "wait canceled" into a constant error, rather then a string.  Call it Success.

cmd/podman/images/utils_linux.go
```diff
@@ -0,0 +1,49 @@
+package images
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
```
rhatdan, line 8: Please use golang.org/x/sys/unix instead of syscall

rhatdan:
Can you add some tests?

QiWang19:
@mtrmac PTAL, will this avoid the hangs?

QiWang19:
> Can you add some tests?


The command needs test is `podman save --format oci-archive alpine | cat> test.tar` 
I haven't foundd ways to test the pipe using ginkgo or in system tests

cmd/podman/images/utils_linux.go
```diff
@@ -0,0 +1,51 @@
...
+		if err != nil {
+			errc <- err
+			return
+		}
+		if _, err = io.Copy(os.Stdout, fpipe); err != nil {
+			errc <- err
+			return
+		}
+		fpipe.Close()
+		errc <- errWaitCancel
```
mtrmac, line 43: Why is this reporting any error at all when the goroutine has succeeded?

cmd/podman/images/utils_linux.go
```diff
@@ -0,0 +1,51 @@
...
+	}()
+	return pipePath, func() <-chan error {
+		if e := os.RemoveAll(pipeDir); e != nil {
+			logrus.Errorf("error removing named pipe: %q", e)
+		}
+		return errc
```
mtrmac, line 49: OK, removing the blocking from this function completely and letting the caller block on `errc`, or not, is a nicely simple way to make that choice without an extra channel. But then the caller would actually have to make that choice.

cmd/podman/images/save.go
```diff
@@ -95,7 +98,17 @@ func save(cmd *cobra.Command, args []string) error {
 		if terminal.IsTerminal(int(fi.Fd())) {
 			return errors.Errorf("refusing to save to terminal. Use -o flag or redirect")
 		}
-		saveOpts.Output = "/dev/stdout"
+		pipePath, cleanup, err = setupPipe()
+		if err != nil {
+			return err
+		}
+		defer func() {
+			errc := cleanup()
```
mtrmac, line 106: In the `utils_unsupported` case, `cleanup` can be `nil`.

mtrmac:
> > Can you add some tests?
> 
> The command needs test is `podman save --format oci-archive alpine | cat> test.tar`
> I haven't foundd ways to test the pipe using ginkgo or in system tests

https://github.com/containers/podman/blob/43f277156c52141c89025ed6a3854f870db38033/test/system/015-help.bats#L19 seems to suggest it should be possible.

In Go, It might be possible to use `os.Pipe` + `cmd.Exec` to run the second command; the `PodmanAsUserBase` call stack would have to be extended with a `stdout` parameter. That might not be worth the effort.

edsantiago:
>seems to suggest it should be possible

Without going into too much gory detail... it's not.The BATS `run` function redirects stdout and stderr, no choice about it. `run_podman`, as a courtesy for debugging test failures, `echo`s the captured result to stdout, and that can (with caveats) be used in some cases.

That cannot work for this PR because the critical point is that the `podman` command itself have the stdout filehandle as a pipe. I have given @QiWang19 a one-time permission to run `$PODMAN` directly in a new test, without going through `run_podman`, as long as there is a big scary DO NOT EVER DO THIS warning.

mtrmac:
> > seems to suggest it should be possible
> 
> Without going into too much gory detail... it's not.

Thanks!

cmd/podman/images/save.go
QiWang19: how to use context.Context to terminate?

mtrmac: The way the `cleanup` function now returns a channel, a separate termination context might not be necessary:
```go
func save(…) (finalErr error) {
succeeded := false
…
if /* use pipe */ {
    pipePath, cleanup, err := setupPipe()
    …
    if cleanup != nil {
        defer func() {
            channel := cleanup()
            if succeeded {
                writeErr := <- channel
                if writeErr != nil && finalErr == nil { finalErr = writeErr }
            }
        }
   }
}
…
err := registry.ImageEngine().Save(context.Background(), args[0], tags, saveOpts)
if err == nil { succeeded = true }
return err
}
```

cmd/podman/images/save.go
```diff
@@ -83,9 +82,13 @@ func saveFlags(flags *pflag.FlagSet) {
 
 }
 
-func save(cmd *cobra.Command, args []string) error {
+func save(cmd *cobra.Command, args []string) (finalErr error) {
 	var (
-		tags []string
+		tags      []string
+		pipePath  string
```
mtrmac, line 88: (Non-blocking: `pipePath`, `cleanup`, (`err`?) can be just initialized with `:=` on first use without a forward declaration I think.)

cmd/podman/images/utils_linux.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+var errWaitCancel = errors.New("wait canceled")
+
+// setupPipe() for fixing https://github.com/containers/podman/issues/7017
+// uses named pipe since containers/image EvalSymlinks fails with /dev/stdout
+func setupPipe() (string, func() <-chan error, error) {
```
mtrmac, line 18: Non-blocking: The semantics of the cleanup callback is now complex enough that it might be worth documenting. (OTOH it’s not _that_ much code and it has a single user.)

QiWang19:
@edsantiago @mtrmac @rhatdan @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM
thanks for the work on this @QiWang19 and @mtrmac !

PR #7075, opened 2020-07-23
[CI:DOCS] BZ1860126 - Fix userns defaults in run man page

TomSweeneyRedHat:
Addresses the multiple "default" userns values found
in the podman-run(1) man page:  http://docs.podman.io/en/latest/markdown/podman-run.1.html.

This in response to: https://bugzilla.redhat.com/show_bug.cgi?id=1860126
which this PR wil fix.

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -905,20 +905,22 @@ Ulimit options. You can use **host** to copy the current configuration from the 
 
 Sets the username or UID used and optionally the groupname or GID for the specified command.
 
-Without this argument the command will be run as root in the container.
+Without this argument, the command will run as the user specified in the container image. Unless overridden by a `USER` command in the Containerfile or by a value passed to this option, this user generally defaults to root.
+
+When a user namespace is not in use, the UID and GID used within the container and on the host will match. When user namespaces are in use, however, the UID and GID in the container may correspond to another UID and GID on the host. In rootless containers, for example, a user namespace is always used, and root in the container will by default correspond to the UID and GID of the user invoking Podman.
 
 **--userns**=**auto**|**host**|**keep-id**|**container:**_id_|**ns:**_namespace_
 
-Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value means user namespaces are disabled.
+Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value ("") means user namespaces are disabled unless an explicit mapping is set with they `--uidmapping` and `--gidmapping` options.
 
 - **auto**: automatically create a namespace.  It is possible to specify other options to `auto`.  The supported options are
   **size=SIZE** to specify an explicit size for the automatic user namespace.  e.g. `--userns=auto:size=8192`.  If `size` is not specified, `auto` will guess a size for the user namespace.
   **uidmapping=HOST_UID:CONTAINER_UID:SIZE** to force a UID mapping to be present in the user namespace.
   **gidmapping=HOST_UID:CONTAINER_UID:SIZE** to force a GID mapping to be present in the user namespace.
-- **host**: run in the user namespace of the caller. This is the default if no user namespace options are set. The processes running in the container will have the same privileges on the host as any other process launched by the calling user.
+- **host**: run in the user namespace of the caller. The processes running in the container will have the same privileges on the host as any other process launched by the calling user (default).
 - **keep-id**: creates a user namespace where the current rootless user's UID:GID are mapped to the same values in the container. This option is ignored for containers created by the root user.
 - **ns**: run the container in the given existing user namespace.
-- **private**: create a new namespace for the container (default)
+- **private**: create a new namespace for the container.
 - **container**: join the user namespace of the specified container.
 
 This option is incompatible with **--gidmap**, **--uidmap**, **--subuid** and **--subgid**.
--- a/pkg/namespaces/namespaces.go
+++ b/pkg/namespaces/namespaces.go
@@ -91,7 +91,7 @@ func (n UsernsMode) IsHost() bool {
 	return n == hostType
 }
 
-// IsKeepID indicates whether container uses a mapping where the (uid, gid) on the host is lept inside of the namespace.
+// IsKeepID indicates whether container uses a mapping where the (uid, gid) on the host is kept inside of the namespace.
 func (n UsernsMode) IsKeepID() bool {
 	return n == "keep-id"
 }


TomSweeneyRedHat:
@giuseppe @mheon, please verify my changes are in line.  That's what I could gronk from the code, but the default userns value setting is not easily parseable.  

Also I did not test the change concerning the USER in the containerfile, but believe that is the case.  Please holler if not, and perhaps we/I should add a test to verify?  I'm not sure we have one atm.

docs/source/markdown/podman-run.1.md
```diff
@@ -901,17 +901,19 @@ Ulimit options. You can use **host** to copy the current configuration from the
 
 Sets the username or UID used and optionally the groupname or GID for the specified command.
 
-Without this argument the command will be run as root in the container.
+Without this argument the command will be run as root in the container, unless a USER has been
```
mheon, line 904: Maybe this should read "Without this argument the command will be run as root on the host unless a user was specified by the container image or the **--user** option"

docs/source/markdown/podman-run.1.md
```diff
@@ -901,17 +901,19 @@ Ulimit options. You can use **host** to copy the current configuration from the
...
+Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value ("") means user namespaces are disabled.
 
 - **auto**: automatically create a namespace.  It is possible to specify other options to `auto`.  The supported options are
   **size=SIZE** to specify an explicit size for the automatic user namespace.  e.g. `--userns=auto:size=8192`.  If `size` is not specified, `auto` will guess a size for the user namespace.
   **uidmapping=HOST_UID:CONTAINER_UID:SIZE** to force a UID mapping to be present in the user namespace.
   **gidmapping=HOST_UID:CONTAINER_UID:SIZE** to force a GID mapping to be present in the user namespace.
-- **host**: run in the user namespace of the caller. This is the default if no user namespace options are set. The processes running in the container will have the same privileges on the host as any other process launched by the calling user.
+- **host**: run in the user namespace of the caller. The processes running in the container will have the same privileges on the host as any other process launched by the calling user.
```
mheon, line 916: This is the default, not private

docs/source/markdown/podman-run.1.md
```diff
@@ -901,17 +901,19 @@ Ulimit options. You can use **host** to copy the current configuration from the
...
 
-Without this argument the command will be run as root in the container.
+Without this argument the command will be run as root in the container, unless a USER has been
+set in the Containerfile.  If the USER was set in the Containerfile, it will be used to process
+commands within the container.
 
 **--userns**=**auto**|**host**|**keep-id**|**container:**_id_|**ns:**_namespace_
 
-Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value means user namespaces are disabled.
+Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value ("") means user namespaces are disabled.
```
giuseppe, line 910: the last sentence is not correct, `An empty value ("") means user namespaces are disabled unless an explicit mapping is set with --uidmapping and --gidmapping`

docs/source/markdown/podman-run.1.md
akostadinov: How about running as root vs non-root? Is this default the same for non-root too? I thought that running as non-root creates a new userns.

docs/source/markdown/podman-run.1.md
rhatdan: Not root on the host, since it could be rootless or running in a user namespace. It will also run as the User of the image.  So the wording of this should be

Without this argument the command will run as the user specified in the container image.  Note that this user often defaults to root.  The actual UID of the process running in the container is based on the User namespace the container is running in.  By default if you are running a rootful Podman, the container will run as the hosts root.  If you are running the container as a rootless user, the root inside of the container is actually the users UID.

docs/source/markdown/podman-run.1.md
rhatdan: This is difficult to explain.

Root inside of the rootless container is your UID, so theoretically it has the same privs as a process run by your user.
Bottom line all containers will run within the same user namespace, and you will gain no separation advantage between containers launched by the same user.  All of the processes in containers will run within the same range of UIDs that other containers you launch run with.

```
$ podman run --userns=host alpine cat /proc/self/uid_map
         0       3267          1
         1     100000      65536
$ podman run --userns=host alpine cat /proc/self/uid_map
         0       3267          1
         1     100000      65536
```

docs/source/markdown/podman-run.1.md
TomSweeneyRedHat: I think it's important to get the UESR/--user option thought in.  So I've tweaked the comment from @rhatdan to:
```
Without this argument, the command will run as the user specified in the container image. Unless overridden by a `USER` command in the Containerfile or by a value passed to this option, this user generally defaults to root. The actual UID of the process running in the container is based on the User namespace the container is running in. By default, if you are running a rootful Podman, the container will run using the UID of the host's root. If you are running the container as a rootless user, by default the UID used by root inside of the container is the rootless user's UID.
```

docs/source/markdown/podman-run.1.md
TomSweeneyRedHat: > This is the default, not private

Thanks, I was looking at the wrong field with a default "Private" value (userns mode) when I change this.  I've changed the default to `host`.

docs/source/markdown/podman-run.1.md
TomSweeneyRedHat: Thanks @giuseppe .  I've added ` unless an explicit mapping is set with --uidmapping and --gidmapping` to the end of that sentence as suggested.

rhatdan:
LGTM

akostadinov:
btw I can't see an other option than `host` work on Fedora 32:
```
[avalon@fedoraw flexy-containerized]$ podman run --rm=true -ti --user=root --userns=private registry.access.redhat.com/ubi8/openjdk-11 /bin/bash -l
Error: --userns "private" is not valid
[avalon@fedoraw flexy-containerized]$ podman run --rm=true -ti --user=root --userns=auto registry.access.redhat.com/ubi8/openjdk-11 /bin/bash -l
Error: --userns "auto" is not valid
[avalon@fedoraw flexy-containerized]$ podman run --rm=true -ti --user=root --userns=auto:size=8192 registry.access.redhat.com/ubi8/openjdk-11 /bin/bash -lError: --userns "auto:size=8192" is not valid
[avalon@fedoraw flexy-containerized]$ podman run --rm=true -ti --user=root --userns=host registry.access.redhat.com/ubi8/openjdk-11 /bin/bash -l
[root@f5bca840ffa6 ~]# exit
logout
[avalon@fedoraw flexy-containerized]$ podman --version
podman version 1.7.0
```

rhatdan:
You should update the version of podman you are using on fedora 32.

docs/source/markdown/podman-run.1.md
```diff
@@ -901,20 +901,20 @@ Ulimit options. You can use **host** to copy the current configuration from the
...
 
-Without this argument the command will be run as root in the container.
+Without this argument, the command will run as the user specified in the container image. Unless overridden by a `USER` command in the Containerfile or by a value passed to this option, this user generally defaults to root. The actual UID of the process running in the container is based on the User namespace the container is running in. By default, if you are running a rootful Podman, the container will run using the UID of the host's root. If you are running the container as a rootless user, by default the UID used by root inside of the container is the rootless user's UID.
 
 **--userns**=**auto**|**host**|**keep-id**|**container:**_id_|**ns:**_namespace_
 
-Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value means user namespaces are disabled.
+Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value ("") means user namespaces are disabled unless an explicit mapping is set with --uidmapping and --gidmapping.
```
TomSweeneyRedHat, line 908: I'm going to change and re-push a slightly amended second sentence:
```suggestion
Set the user namespace mode for the container.  It defaults to the **PODMAN_USERNS** environment variable.  An empty value ("") means user namespaces are disabled unless an explicit mapping is set with the `--uidmapping` and `--gidmapping` options.
```

TomSweeneyRedHat:
All green and I think I've addressed all comments.  @mheon, could you do a final review and push if it LGTY please?

docs/source/markdown/podman-run.1.md
```diff
@@ -901,20 +901,20 @@ Ulimit options. You can use **host** to copy the current configuration from the
 
 Sets the username or UID used and optionally the groupname or GID for the specified command.
 
-Without this argument the command will be run as root in the container.
+Without this argument, the command will run as the user specified in the container image. Unless overridden by a `USER` command in the Containerfile or by a value passed to this option, this user generally defaults to root. The actual UID of the process running in the container is based on the User namespace the container is running in. By default, if you are running a rootful Podman, the container will run using the UID of the host's root. If you are running the container as a rootless user, by default the UID used by root inside of the container is the rootless user's UID.
```
mheon, line 904: I don't know if this paragraph makes much sense. We shouldn't be focusing on the user the container is run as, but that root in the container is or is not root on the host.

TomSweeneyRedHat: This whole paragraph started from a question ni the BZ:  "Namely should lack of `--user` option result in container running as specified by `USER` in Dockerfile? This is what I observe.'

I started the first bit of the USER, then Dansuggested the bit about the UID too and I tweaked it slightly.  I'm happy to make it anything that folks see fit.  Would it work if I just remove the last bit talking about the UID?

```
Without this argument, the command will run as the user specified in the container image. Unless overridden by a `USER` command in the Containerfile or by a value passed to this option, this user generally defaults to root. 
```

mheon: This sounds like two separate issues.

We need to beef up the documentation of `--user` to detail what happens there; and we need to improve the documentation here to specify what happens when user namespaces are involved. Intermingling the two here seems nonsensical from a reader's perspective, especially given that user namespaces do not just alter the mapping of the user the container is run as, but all users in the container.

TomSweeneyRedHat: That is what I've attempted to do here.  Apparently I've missed the mark.  Please give me suggested changes as I'm not sure what to change at this point.

mheon: I think we need to separate the bit about user/Dockerfile USER and put it in the description of the `--user` flag. This section should focus entirely on how a userns will result in users inside the container having different IDs from users outside the container, how this depends on the exact UID mappings of the user namespace, and how user namespaces are default for rootless and optional for root. Will write more tomorrow.

TomSweeneyRedHat: @mheon, this is currently under the --user option, not the --userns option.  Did you get a chance to tickle the keyboard with your thoughts?

mheon: Having this bit under `--user` and the rest under `--userns` does satisfy my request.

Still, the documentation here seems overly focused on root. Should probably be more generic - "When a user namespace is not in use, the UID and GID used within the container and on the host will match. When user namespaces are in use, however, the UID and GID in the container may correspond to another UID and GID on the host. In rootless containers, for example, a user namespace is always used, and root in the container will by default correspond to the UID and GID of the container invoking Podman."

TomSweeneyRedHat: Matt, I've removed this:  
```
When a user namespace is not in use, the UID and GID used within the container and on the host will match. 
When user namespaces are in use, however, the UID and GID in the container may correspond to another 
UID and GID on the host. In rootless containers, for example, a user namespace is always used, and root in
the container will by default correspond to the UID and GID of the container invoking Podman.
```
From the original paragraph.  I then took what you have above and made it a new second paragraph under user, BUT, I changed the second from last word from `container` to `user`

PTAL

mheon:
LGTM, thanks!

rhatdan:
@TomSweeneyRedHat Great job.
/lgtm

PR #7077, opened 2020-07-24
Lower missing runtime from Errorf to Infof

rhatdan:
Creating a container with --runtime=...  Can cause a situation where all of the tools that look at containers starts generating an ugly error message that
does not effect the running of the tool, but gives the users a bad experience.

Also users can not fix this situation easily without removing the container or
figuring out how to edit containers.conf.

From looking at the problem, it does not look like a real issue and should just be printed at the Info level.

### DIFF
--- a/libpod/oci_missing.go
+++ b/libpod/oci_missing.go
@@ -45,8 +45,8 @@ func getMissingRuntime(name string, r *Runtime) OCIRuntime {
 		return runtime
 	}
 
-	// Once for each missing runtime, we want to error.
-	logrus.Errorf("OCI Runtime %s is in use by a container, but is not available (not in configuration file or not installed)", name)
+	// Once for each missing runtime, we want to print info.
+	logrus.Infof("OCI Runtime %s is in use by a container, but is not available (not in configuration file or not installed)", name)
 
 	newRuntime := new(MissingRuntime)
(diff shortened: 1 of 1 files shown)


rhatdan:
@mheon @giuseppe @vrothberg @TomSweeneyRedHat @baude @jwhonce @ashley-cui @QiWang19 PTAL

rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1860176

mheon:
This is actually a serious issue for some commands. There's no way to do a
'podman start' or 'podman kill' or 'podman rm -f' without a working
runtime. We can probably get away with doing this, but the error message
that the missing runtime prints will need to be improved to tell people
what happened and how to fix it

rhatdan:
Should we make it more difficult to get into this state?

mheon:
I'm thinking that's a good idea.

Instead of making a missing runtime if it's not in the config, we can first look it up in $PATH, and if we find it we use it - if not, we fall back to MissingRuntime and the error message.

rhatdan:
@mheon Could you take this one over, since you understand what is going on here better then me.

mheon:
Sure!

mheon:
I think we can close this in favor of #7126

PR #7081, opened 2020-07-24
fix bug podman sign storage path

### LINKED ISSUE #6993: (podman image sign) fails with “Use a supported scheme” if the directory is explicitly specified or the default
/kind bug

**Description**

For now, filing just to have a place to track this.

```console
$ podman --version
podman version 1.6.4
$ podman image sign --directory /var/lib/atomic/sigstore --sign-by … docker://…
…
ERRO[0000] invalid signature storage /var/lib/atomic/sigstore: writing to /var/lib/atomic/sigstore is not supported. Use a supported scheme
```

From a quick skim of the code, this seems became worse since, in that this message is triggered on many more code paths (and `--directory` is entirely ignored?!).

- `isValidSigStoreDir` is misnamed: it’s actually `localPathFromURI`
- … and it is called on strings that are paths, not URIs.

### LINKED ISSUE #6994: (podman image sign) uses an unpredictable image name for the destination file
/kind bug

**Description**

For now, filling primarily to have a place to track this.

If a specific image is available locally, using several different repo@digest values, `podman image sign` will use that value instead of the name the user specified for storing the resulting signature.
 https://github.com/containers/podman/blob/54d16f3b5fc4210c7c46e4cca240f988fa5f39f5/pkg/domain/infra/abi/images.go#L637 (or the equivalent in earlier versions) should instead use the repository from `srcRef` and the digest of `getManifest`, like `c/image/docker.{configuredSignatureStorageBase,signatureStorageURL}` .

QiWang19:
- fix the bud podman not using specified --directory as signature storage.
- use manifest and image referce to set repo@digest.
close #6994
close #6993


pkg/domain/infra/abi/images.go
```diff
@@ -608,15 +606,6 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 			return nil, errors.Errorf("cannot determine canonical Docker reference for destination %s", transports.ImageName(rawSource.Reference()))
 		}
 
-		// create the signstore file
```
rhatdan, line 611: Do we not need to create the signstore file?

QiWang19: Move the comments down to L635 where the sigstore is created. Deleted lines of code are not used  since the repo and digest can be got from image reference.

rhatdan:
@mtrmac @vrothberg @mheon @baude @TomSweeneyRedHat @ashley-cui @jwhonce @giuseppe PTAL
LGTM

TomSweeneyRedHat:
LGTM
and I concur with @vrothberg on an added test or two.

pkg/domain/infra/abi/images.go
```diff
@@ -747,28 +736,27 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
 		// create signature
 		newSig, err := signature.SignDockerManifest(getManifest, dockerReference.String(), mech, options.SignBy)
 		if err != nil {
 			return nil, errors.Wrapf(err, "error creating new signature")
 		}
-
-		trimmedDigest := strings.TrimPrefix(repos[0], strings.Split(repos[0], "/")[0])
-		sigStoreDir = filepath.Join(sigStoreDir, strings.Replace(trimmedDigest, ":", "=", 1))
+		// create the signstore file
+		sigStoreDir = fmt.Sprintf("%s@%s=%s", filepath.Join(sigStoreDir, repo), manifestDigest.Algorithm(), manifestDigest.Hex())
```
mtrmac, line 759: (Consider introducing a separate `signatureDir` here, again to avoid overloading a single variable with multiple meanings.)

pkg/domain/infra/abi/images.go
```diff
@@ -725,63 +721,66 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
+						sigStoreURI = registryInfo.SigStore
+					}
+				}
 
+				if sigStoreURI == "" {
+					if registryConfigs.DefaultDocker == nil {
+						logrus.Debugf(" No signature storage configuration found for %s", rawSource.Reference().DockerReference().String())
+						continue
+					}
+					if sigStoreURI = registryConfigs.DefaultDocker.SigStoreStaging; sigStoreURI == "" {
```
mtrmac, line 745: I was expecting `HaveMatchRegistry` to handle the `DefaultDocker` setting — apparently it doesn’t, but I maybe it should be modified that way, also for `podman trust show` to correctly (?) report the right signature store path in the default case.

Either way, it would be simpler to have a single `if registryInfo == nil { registryInfo = registryConfigs.DefaultDocker }` check and a single `SigStoreStaging`/`SigStore` decision.

pkg/domain/infra/abi/images.go
```diff
@@ -725,63 +721,66 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
 				if registryInfo != nil {
-					if sigStoreDir = registryInfo.SigStoreStaging; sigStoreDir == "" {
-						sigStoreDir = registryInfo.SigStore
+					if sigStoreURI = registryInfo.SigStoreStaging; sigStoreURI == "" {
+						sigStoreURI = registryInfo.SigStore
+					}
+				}
 
+				if sigStoreURI == "" {
+					if registryConfigs.DefaultDocker == nil {
```
mtrmac, line 741: This is not quite the right condition; `defaultDocker` can exist, but with both `SigStore` and `SigStoreStaging` empty.

pkg/domain/infra/abi/images.go
```diff
@@ -725,63 +721,66 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
+						logrus.Debugf(" No signature storage configuration found for %s", rawSource.Reference().DockerReference().String())
```
mtrmac, line 742: Shouldn’t this fail instead of just ignoring the image?

test/e2e/image_sign_test.go
```diff
@@ -0,0 +1,63 @@
...
+	BeforeEach(func() {
+		tempdir, err = CreateTempDirInTempDir()
+		if err != nil {
+			os.Exit(1)
+		}
+		podmanTest = PodmanTestCreate(tempdir)
+		podmanTest.Setup()
+		podmanTest.SeedImages()
+
+		tempGUNPGHOME := filepath.Join(podmanTest.TempDir, "tmpGPG")
```
mtrmac, line 31:
```suggestion
		tempGNUPGHOME := filepath.Join(podmanTest.TempDir, "tmpGPG")
```

test/e2e/image_sign_test.go
```diff
@@ -0,0 +1,63 @@
...
+		err := os.Mkdir(tempGUNPGHOME, os.ModePerm)
+		Expect(err).To(Equal(nil))
+
+		origGNUPGHOME := os.Getenv("GNUPGHOME")
+		defer os.Setenv("GNUPGHOME", origGNUPGHOME)
```
mtrmac, line 36: This is called at the end of the `BeforeEach()` callback; it needs to happen in `AfterEach` instead.

test/e2e/image_sign_test.go
```diff
@@ -0,0 +1,63 @@
...
+		cmd := exec.Command("gpg", "--import", "sign/secret-key.asc")
+		err := cmd.Run()
+		Expect(err).To(BeNil())
+		sigDir := filepath.Join(podmanTest.TempDir, "test-sign")
+		err = os.MkdirAll(sigDir, os.ModePerm)
+		Expect(err).To(BeNil())
+		session := podmanTest.Podman([]string{"image", "sign", "--directory", sigDir, "--sign-by", "foo@bar.com", "docker://library/alpine"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		sigDigestDir := "alpine@sha256=185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321"
```
mtrmac, line 59: Does this work long-term? Looking at `CACHE_IMAGES`, `alpine:latest` is pulled from an external registry, so the digest might change over time.

pkg/domain/infra/abi/images.go
```diff
@@ -725,63 +721,60 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
 				if registryInfo != nil {
-					if sigStoreDir = registryInfo.SigStoreStaging; sigStoreDir == "" {
-						sigStoreDir = registryInfo.SigStore
-
+					if sigStoreURI = registryInfo.SigStoreStaging; sigStoreURI == "" {
+						sigStoreURI = registryInfo.SigStore
 					}
 				}
+				if sigStoreURI == "" {
+					return nil, errors.Errorf(" No signature storage configuration found for %s", rawSource.Reference().DockerReference().String())
```
mtrmac, line 740: The leading space is probably not wanted.

mheon: Also should not start with a capital

PR #7083, opened 2020-07-25
replace the html/template package with text/template

Luap99:
Currently some commands use the html/template package.
This can lead to invalid output.
e.g. `system df --verbose` will print `&lt;none&gt;`
instead of `<none>` with an untaged image.


rhatdan:
LGTM
@vrothberg @mheon @baude @jwhonce @giuseppe @TomSweeneyRedHat PTAL

PR #7085, opened 2020-07-26
Cleanup handling of podman mount/unmount

rhatdan:
We should default to the user name unmount rather then the internal
name of umount.

Also User namespace was not being handled correctly. We want to inform
the user that if they do a mount when in rootless mode that they have
to be first in the podman unshare state.


rhatdan:
Splitting this PR away from the bigger podmain image mount PR.  Hopefully this will help me figure out what is causing the podman tag/untag issues, that are preventing that PR from merging.

pkg/rootless/rootless_linux.c
```diff
@@ -210,6 +210,13 @@ can_use_shortcut ()
           ret = false;
           break;
         }
+
+      if (strcmp (argv[argc], "container") == 0 &&
+	   strcmp (argv[argc+1], "mount") == 0)
```
edsantiago, line 215: index out of range

rhatdan: Fixed

docs/source/markdown/links/podman-unmount.1
```diff
@@ -1 +0,0 @@
-.so man1/podman-umount.1
```
TomSweeneyRedHat, line 1: Was this intentionally deleted?

TomSweeneyRedHat:
One question, and tests as @vrothberg mentioned would be good.  Otherwise LGTM

docs/source/markdown/links/podman-unmount.1
rhatdan: Yes, It is actually renamed.  podman-umount.1 now links to man1/podman-unmount.1

rhatdan:
Tests added, but I could not get them to run locally.  Will see what happens here.

cmd/podman/registry/config.go
```diff
@@ -16,6 +16,7 @@ import (
 
 const (
 	ParentNSRequired = "ParentNSRequired"
+	UserNSRequired   = "UserNSRequired"
```
mheon, line 19: Could this be renamed? Maybe "RootRequired"?

rhatdan: I don't think that is helpful.  How about UnshareUserNSRequired Or UnshareNSRequired

mheon: UnshareNSRequired SGTM

mheon:
LGTM once the UserNSRequired field is renamed

vrothberg:
Restarted flakes

Issue #7078, opened 2020-07-24
events Rest API changes in v1.12


KKoukiou:
Previously this used to work, now it returns this error:

```
# curl -s --unix-socket /run/podman/podman.sock http://d/v1.12/events
{"cause":"unexpected end of JSON input","message":"Failed to parse parameters for /v1.12/events: unexpected end of JSON input","response":400}
```

Tested with podman-2.0.3-1.module+el8.3.0+7450+9ea086fe.x86_64

(breaks RHEL-8-3 gating)

rhatdan:
Ok that is bad.

mheon:
I am wondering if this was the parameter type change from @vrothberg from `map[string][]stiring` to `map[string]map[string]bool` for Docker compat.

I'm also curious as to why the Docker endpoint is being used here - thought Cockpit was using our endpoints?

vrothberg:
I've tested it with Docker but will take a look right now.

marusak:
> I'm also curious as to why the Docker endpoint is being used here - thought Cockpit was using our endpoints?

This was the last one where we used compat one, now we switched to libpod one - https://github.com/cockpit-project/cockpit-podman/pull/456

However it is broken in both:
```
$ sudo curl -s --unix-socket /run/podman/podman.sock http://d/v1.12/events
{"cause":"unexpected end of JSON input","message":"Failed to parse parameters for /v1.12/events: unexpected end of JSON input","response":400}
$ sudo curl -s --unix-socket /run/podman/podman.sock http://d/v1.12/libpod/events
{"cause":"unexpected end of JSON input","message":"Failed to parse parameters for /v1.12/libpod/events: unexpected end of JSON input","response":400}
```

I updated podman yesterday from fedora repos and now cannot use c-podman at all.
```
$ rpm -q podman
podman-2.0.3-1.fc32.x86_64
```

vrothberg:
Thanks, @marusak!

martinpitt:
Can you please make a new release with this? 2.0.3 now made it into Fedora 32 as well, and it's completely breaking cockpit-podman. Thanks!

marusak:
[...]

PR #7088, opened 2020-07-27
API events: fix parsing error

vrothberg:
Fix an error where an absent "filters" parameter led to JSON parsing
errors.

Fixes: #7078

### DIFF
--- a/test/apiv2/01-basic.at
+++ b/test/apiv2/01-basic.at
@@ -68,4 +68,8 @@ else
     _show_ok 0 "Time for ten /info requests" "<= 5 seconds" "$delta_t seconds"
 fi
 
+# Simple events test (see #7078)
+t GET "events?stream=false"  200
+t GET "libpod/events?stream=false"  200
+
 # vim: filetype=sh
--- a/pkg/api/handlers/compat/events.go
+++ b/pkg/api/handlers/compat/events.go
@@ -29,8 +29,14 @@ func filtersFromRequest(r *http.Request) ([]string, error) {
 		compatFilters map[string]map[string]bool
 		filters       map[string][]string
 		libpodFilters []string
+		raw           []byte
 	)
-	raw := []byte(r.Form.Get("filters"))
+
+	if _, found := r.URL.Query()["filters"]; found {
+		raw = []byte(r.Form.Get("filters"))
+	} else {
+		return []string{}, nil
+	}
 
 	// Backwards compat with older versions of Docker.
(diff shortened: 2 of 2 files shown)


vrothberg:
@jwhonce @mheon @rhatdan PTAL

vrothberg:
I am under the impression that the tests in `test/api2/rest_api` are not executed. They should have caught the regression but didn't and I couldn't find any wiring in the CI.

vrothberg:
Added a test the apiv2 test to prevent future regressions.

test/apiv2/01-basic.at
```diff
@@ -68,4 +68,8 @@ else
     _show_ok 0 "Time for ten /info requests" "<= 5 seconds" "$delta_t seconds"
 fi
 
+# Simple events test (see #7078)
+t GET "events?stream=false"  200
+t GET "libpod/events?stream=false"  200
+
```
edsantiago, line 74: Friendly reminder to check actual results, not just status code. From a quick look at the returned JSON, might I suggest:
```bash
t GET "events?stream=false"        200 .Type=system .scope=local .time~[0-9]\\+
t GET "libpod/events?stream=false" 200 .Type=system .scope=local .time~[0-9]\\+
```

vrothberg: Thanks for checking! It doesn't return any data here though. I really just want to exercise this specific code which we can't with bindings as the "filters" param is set unconditionally afaics.

edsantiago: PS last week I looked into adding tests for `/events` but gave up because of the `follow` nature - I didn't want tests to hang. If you have hints for how to write non-hanging tests (looks like `?stream=false` is the magic token I was missing) I would like to look into that again. Thank you.

vrothberg: Yes, `?stream=false` would be a good candidate for the tests here :+1:

edsantiago: >It doesn't return any data here 

Weird! As in: when you run the test I suggested, it fails?? That alarms me a little bit: I tried my tests root and rootless.

vrothberg: It now succeeds without `?filters=...`. It failed before this change with a regression I introduced. The other tests always set "filters", so here I just wanted to make sure that the code path is executed.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

mheon:
LGTM

PR #7097, opened 2020-07-27
[CI:DOCS] apiv2 fix volumes not included field

QiWang19:
Do not use volume from docker since UsageData field is not need. It's nullable in docker API and expensive to add.

### DIFF
--- a/pkg/domain/entities/volumes.go
+++ b/pkg/domain/entities/volumes.go
@@ -59,6 +59,42 @@ type VolumeConfigResponse struct {
 	Anonymous bool `json:"Anonymous"`
 }
 
+// VolumeInfo Volume list response
+// swagger:model VolumeInfo
+type VolumeInfo struct {
+
+	// Date/Time the volume was created.
+	CreatedAt string `json:"CreatedAt,omitempty"`
+
+	// Name of the volume driver used by the volume. Only supports local driver
+	// Required: true
+	Driver string `json:"Driver"`
+
+	// User-defined key/value metadata.
+	// Always included
+	Labels map[string]string `json:"Labels"`
+
+	// Mount path of the volume on the host.
+	// Required: true
+	Mountpoint string `json:"Mountpoint"`
+
+	// Name of the volume.
+	// Required: true
+	Name string `json:"Name"`
+
+	// The driver specific options used when creating the volume.
+	// Required: true
+	Options map[string]string `json:"Options"`
+
+	// The level at which the volume exists.
+	// Libpod does not implement volume scoping, and this is provided solely for
+	// Docker compatibility. The value is only "local".
+	// Required: true
+	Scope string `json:"Scope"`
+
+	// TODO: We don't include the volume `Status` for now
+}
+
 type VolumeRmOptions struct {
 	All   bool
 	Force bool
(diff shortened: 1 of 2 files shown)


pkg/domain/entities/volumes.go
```diff
@@ -59,6 +59,55 @@ type VolumeConfigResponse struct {
...
+	Status map[string]interface{} `json:"Status,omitempty"`
+}
+
+// VolumeListOKBody Volume list response
+// swagger:model VolumeListOKBody
+type VolumeListOKBody struct {
+
+	// List of volumes
+	// Required: true
+	Volumes VolumeInfo `json:"Volumes"`
```
mheon, line 108: Should this be an array?

pkg/domain/entities/volumes.go
```diff
@@ -59,6 +59,55 @@ type VolumeConfigResponse struct {
...
+
+	// The driver specific options used when creating the volume.
+	// Required: true
+	Options map[string]string `json:"Options"`
+
+	// The level at which the volume exists. Either `global` for cluster-wide, or `local` for machine level.
+	// Required: true
+	Scope string `json:"Scope"`
+
+	// Low-level details about the volume, provided by the volume driver.
```
mheon, line 92: It's completely omitted in Podman for now, we should mention this.

pkg/domain/entities/volumes.go
```diff
@@ -59,6 +59,55 @@ type VolumeConfigResponse struct {
...
 
+// swagger:model VolumeInfo
+type VolumeInfo struct {
+
+	// Date/Time the volume was created.
+	CreatedAt string `json:"CreatedAt,omitempty"`
+
+	// Name of the volume driver used by the volume.
+	// Required: true
+	Driver string `json:"Driver"`
```
mheon, line 70: Libpod only supports the Local driver, should mention this.

pkg/domain/entities/volumes.go
```diff
@@ -59,6 +59,55 @@ type VolumeConfigResponse struct {
...
+
+	// User-defined key/value metadata.
+	// Required: true
+	Labels map[string]string `json:"Labels"`
```
mheon, line 74: This is optional

mheon: Actually, now that I think about it... this is returned by the server, so required/optional might not make sense.

"Always included" vs "Optional" maybe?

pkg/domain/entities/volumes.go
```diff
@@ -59,6 +59,55 @@ type VolumeConfigResponse struct {
...
+
+	// Name of the volume.
+	// Required: true
+	Name string `json:"Name"`
+
+	// The driver specific options used when creating the volume.
+	// Required: true
+	Options map[string]string `json:"Options"`
+
+	// The level at which the volume exists. Either `global` for cluster-wide, or `local` for machine level.
```
mheon, line 88: This is always set to "local" for Podman, we do not support global volumes (they're a swarm thing)

TomSweeneyRedHat:
Should this PR really have the `[CI:DOCS]` tag in the subject?

TomSweeneyRedHat:
LGTM once @mheon 's concerns are addressed.

Issue #7100, opened 2020-07-28
IPC + PID Namespace from host does not work in combination


mario19911:
**Description**

I am trying to run two containers, both with --ipc=host and --pid=host. With this combination I am not able to see the shared memories from container A in container B. When I remove --pid=host it is working.

**Steps to reproduce the issue:**

1. Create a container which creates a shared memory (shm_server from https://www.cs.cf.ac.uk/Dave/C/node27.html)

2. Startup container A with podman run -it --ipc=host --pid=host --rm /bin/bash

3. Startup shm_server in container and check it with ipcs -m

4. Startup the second container with podman run -it --ipc=host --pid=host --rm /bin/bash and check the same with ipcs -m (cannot see the shared memory created in container A)

5. Startup the second container again with podman run -it --ipc=host  --rm /bin/bash and check ipcs -m --> does work

**Describe the results you received:**
When both namespaces are set to "host" I cannot see the generated shared memories.

**Describe the results you expected:**
To see the shared memory of container A in container B.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.3
API Version:  1
Go Version:   go1.14.2
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.18, commit: '
  cpus: 1
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: file
  hostname: mario-VirtualBox
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.4.0-42-generic
  linkmode: dynamic
  memFree: 138342400
  memTotal: 2084356096
  ociRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: 'runc version spec: 1.0.1-dev'
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 383512576
  swapTotal: 470282240
  uptime: 16h 29m 40.51s (Approximately 0.67 days)
registries:
  search:
  - docker.io
  - quay.io
store:
[...]

PR #7109, opened 2020-07-28
Specifying --ipc=host --pid=host is broken

rhatdan:
For some reason we were overwriting memory when handling both
--pid=host and --ipc=host.  Simplified the code to handle this
correctly, and add test to make sure it does not happen again.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -186,6 +186,46 @@ func getMemoryLimits(s *specgen.SpecGenerator, c *ContainerCLIOpts) (*specs.Linu
 	return memory, nil
 }
 
+func setNamespaces(s *specgen.SpecGenerator, c *ContainerCLIOpts) error {
+	var err error
+
+	if c.PID != "" {
+		s.PidNS, err = specgen.ParseNamespace(c.PID)
+		if err != nil {
+			return err
+		}
+	}
+	if c.IPC != "" {
+		s.IpcNS, err = specgen.ParseNamespace(c.IPC)
+		if err != nil {
+			return err
+		}
+	}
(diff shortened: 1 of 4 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/7100

cmd/podman/common/specgen.go
```diff
@@ -252,18 +252,21 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
 	s.Expose = expose
 
-	for k, v := range map[string]*specgen.Namespace{
-		c.IPC:       &s.IpcNS,
-		c.PID:       &s.PidNS,
-		c.UTS:       &s.UtsNS,
-		c.CGroupsNS: &s.CgroupNS,
-	} {
-		if k != "" {
-			*v, err = specgen.ParseNamespace(k)
```
vrothberg, line 262: This looks like a classic loop-variable overwrite/leak. Good catch!

mheon:
LGTM

mheon:
Set lgtm/hold labels

TomSweeneyRedHat:
LGTM
assuming happy tests

cmd/podman/common/specgen.go
giuseppe: I think the issue was that there are collisions with the keys (both c.IPC and c.PID have the value "host")

both

debarshiray:
I believe that this will also fix https://github.com/containers/toolbox/issues/501 - Toolbox uses `podman create --ipc host --pid host ...`

Thanks for the fix, @rhatdan (and the tests)!

Issue #7099, opened 2020-07-28
podman run --env-host: inconsistency between man page, actual behavior


edsantiago:
podman-run(1) documents the ENVIRONMENT precedence order as follows:
- **--env-host**: Host environment of the process executing Podman is added.
- **--http-proxy**: ....
- Container image: Any environment variables specified in the container image.

However, in practice:
```console
$ printf "FROM alpine\nENV E=defined-in-image\n" >ddd/Containerfile
$ ./bin/podman build -t foo ddd
...
$ ./bin/podman run foo printenv E
defined-in-image
$ ./bin/podman run --env-host foo printenv E
defined-in-image
$ E=from-host ./bin/podman run --env-host foo printenv E
from-host
```
I don't know if the bug is in the man page or the code. This is master @ 2b7bc9b10

If someone fixes this, no need to add tests: that's what I'm working on and how I caught it. I'll be submitting a PR tomorrow, with this particular test commented out.

rhatdan:
Bug in the code.
@ashley-cui PTAL
Anything the user specifies overrides the defaults.

rhatdan:
BTW This would be a good test.

rhatdan:
Oops this is a bug in man page.
--env-host and --http-proxy should override the image.

rhatdan:
I tested this by hand and the code seems to work correctly.

edsantiago:
OK, I'll fix the man page in my PR.

PR #7094, opened 2020-07-27
When chowning we should not follow symbolic link

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #7110, opened 2020-07-28
System tests: add environment, volume tests

edsantiago:
Tests for #7094, in which symlinks in a volume would
cause chown errors and nonrunnable containers.

Tests for environment variable precedence, now
include --env-host and proxy settings

Fix a bug caught by covscan in helpers.t ('source'
path would fail if path included spaces).

Fix podman-run man page: it was incorrect in stating
precedence between in-image environment and --env-host.

Fixes: #7099

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -1411,9 +1411,9 @@ required for VPN, without it containers need to be run with the **--network=host
 Environment variables within containers can be set using multiple different options,
 in the following order of precedence (later entries override earlier entries):
 
-- **--env-host**: Host environment of the process executing Podman is added.
-- **--http-proxy**: By default, several environment variables will be passed in from the host, such as **http_proxy** and **no_proxy**. See **--http-proxy** for details.
 - Container image: Any environment variables specified in the container image.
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

edsantiago:
```console
$ podman run --rm --env-host [...] build_test
/bin/mydefaultcmd: line 4: printenv: not found
/bin/mydefaultcmd: line 4: grep: not found
/bin/mydefaultcmd: line 4: sed: not found
[ rc=127 (** EXPECTED 0 **) ]
```
Eeek! I have a sneaking suspicion that this might be `$PATH` (possibly `/usr/bin` vs `/bin`) but I can't reproduce on my end. Has anyone run into problems like this before when using `--env-host`?

mheon:
It is most likely $PATH leaking in from the host... You may need to manually redefine it in the container.

edsantiago:
Pushed a change that sets it in the CMD. (Changing it in the container image would, I assume, be useless because of --env-host). This is an interesting corner case; I Think it may merit discussion.

rhatdan:
@mheon @TomSweeneyRedHat @vrothberg @giuseppe @baude @jwhonce @ashley-cui @QiWang19 PTAL

mheon:
LGTM
/approve

QiWang19:
LGTM

Issue #7010, opened 2020-07-17
podman --remote does not work


Luap99:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman --remote` does not work.
It looks like the options `--remote` and `--url` are simply ignored.
`podman-remote` works

**Steps to reproduce the issue:**

`podman --remote --url "abcd" ps -a`

**Describe the results you received:**
list of local containers
```
CONTAINER ID  IMAGE                            COMMAND               CREATED      STATUS                  PORTS   NAMES
02633c480efb  docker.io/library/alpine:latest  ls -la /test          2 weeks ago  Exited (0) 2 weeks ago          sad_wozniak
29502a4d9e70  docker.io/library/alpine:latest  top                   2 weeks ago  Created                         condescending_newton
```
**Describe the results you expected:**
list of remote containers or an error if the service is not running or the url was invalid
e.g. `Error: Get "http://d/v1.0.0/libpod../../../_ping": dial unix ///run/podman/podman.sock: connect: no such file or directory`

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.1.0-dev
API Version:  1
Go Version:   go1.14.4
Git Commit:   10c5f241231a55a90a37b9f578c84a6bb23ae33e
Built:        Fri Jul 17 22:19:23 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.0-dev
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.18, commit: '
  cpus: 16
  distribution:
    distribution: debian
    version: unknown
  eventLogger: file
  hostname: pc
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.7.0-1-amd64
  linkmode: dynamic
  memFree: 1521647616
  memTotal: 16802889728
  ociRuntime:
    name: runc
    package: 'runc: /usr/bin/runc'
    path: /usr/bin/runc
    version: |-
[...]

PR #7118, opened 2020-07-28
Support DOCKER_HOST and DOCKER_CERT_PATH

jwhonce:
The assumed format for DOCKER_HOST is `<schema>://<host>[:<port>][path]` .

If the ssh schema is used for DOCKER_HOST, "/run/podman/podman.sock"
is appended if path is not given.  All other schemas are assumed to
resolve to a destination running a Podman service as given.

usage: `$ DOCKER_HOST=ssh://root@localhost:22 bin/podman ps --all`

Issue: GH-7010

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path"
 	"runtime"
@@ -236,7 +237,7 @@ func loggingHook() {
 
 func rootFlags(cmd *cobra.Command, opts *entities.PodmanConfig) {
 	cfg := opts.Config
-	uri, ident := resolveDestination()
+	uri, ident := activeDestination()
 
 	lFlags := cmd.Flags()
 	lFlags.BoolVarP(&opts.Remote, "remote", "r", false, "Access remote Podman service (default false)")
@@ -289,13 +290,19 @@ func rootFlags(cmd *cobra.Command, opts *entities.PodmanConfig) {
 	}
 }
 
-func resolveDestination() (string, string) {
-	if uri, found := os.LookupEnv("CONTAINER_HOST"); found {
-		var ident string
-		if v, found := os.LookupEnv("CONTAINER_SSHKEY"); found {
-			ident = v
+func activeDestination() (string, string) {
+	if uri, found := os.LookupEnv("DOCKER_HOST"); found {
+		u, err := url.Parse(uri)
+		if err != nil {
+			return registry.DefaultAPIAddress(), ""
 		}
-		return uri, ident
+
+		// To handle rootless would require querying the service.
+		// `podman system connection add` already has all that code.
+		if u.Scheme == "ssh" && u.Path == "" {
+			u.Path = "/run/podman/podman.sock"
+		}
+		return u.String(), os.Getenv("DOCKER_CERT_PATH")
 	}
 
 	cfg, err := config.ReadCustomConfig()
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -64,6 +64,13 @@ Path to ssh identity file. If the identity file has been encrypted, podman promp
 If no identity file is provided and no user is given, podman defaults to the user running the podman command.
 Podman prompts for the login password on the remote server.
 
+Identity value resolution precedence:
+ - command line value
+ - environment variable `DOCKER_CERT_PATH`, if `DOCKER_HOST` is found
+ - environment variable `CONTAINER_SSHKEY`, if `CONTAINER_HOST` is found
+ - `containers.conf`
+ - `""`
+
 **--log-level**=*level*
 
 Log messages above specified level: debug, info, warn, error (default), fatal or panic (default: "error")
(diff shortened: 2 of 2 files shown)


cmd/podman/root.go
```diff
@@ -289,13 +290,19 @@ func rootFlags(cmd *cobra.Command, opts *entities.PodmanConfig) {
 	}
 }
 
-func resolveDestination() (string, string) {
-	if uri, found := os.LookupEnv("CONTAINER_HOST"); found {
```
mheon, line 293: Shouldn't we continue supporting the old variables too?

rhatdan:
@afbjorklund PTAL

cmd/podman/root.go
jwhonce: Environment variables CONTAINER_* support comes via the cfg.ActiveDestination() call.  The code here existed before that refactor and was duplicaticated.

docs/source/markdown/podman.1.md
```diff
@@ -80,7 +82,27 @@ Path to the command binary to use for setting up a network.  It is currently onl
 Access Podman service will be remote
 
 **--url**=*value*
-URL to access Podman service (default from `containers.conf`, rootless "unix://run/user/$UID/podman/podman.sock" or as root "unix://run/podman/podman.sock).
+URL to access Podman service (default from `containers.conf`, rootless `unix://run/user/$UID/podman/podman.sock` or as root `unix://run/podman/podman.sock`).
```
TomSweeneyRedHat, line 85: nit
```suggestion
URL to access the Podman service (default from `containers.conf`, rootless `unix://run/user/$UID/podman/podman.sock` or as root `unix://run/podman/podman.sock`).
```

docs/source/markdown/podman.1.md
```diff
@@ -80,7 +82,27 @@ Path to the command binary to use for setting up a network.  It is currently onl
...
+
+The `CONTAINER_HOST` or `DOCKER_HOST` environment variables override the default.
+ - `CONTAINER_HOST` is of the format `<schema>://[<user[:<password>]@]<host>[:<port>][<path>]`
+ - `DOCKER_HOST` is of the format `<schema>://<host>[:<port>]`
+
+Details:
+ - `user` will default to either `root` or current running user
+ - `password` has no default
+ - `host` must be provided and is either the IP or name of the machine hosting the podman service
```
TomSweeneyRedHat, line 94:
```suggestion
 - `host` must be provided and is either the IP or name of the machine hosting the Podman service
```

docs/source/markdown/podman.1.md
TomSweeneyRedHat, line 85: We generally list root, then rootless locations, I'd reverse the ordering here.

docs/source/markdown/podman.1.md
```diff
@@ -64,6 +64,8 @@ Path to ssh identity file. If the identity file has been encrypted, podman promp
 If no identity file is provided and no user is given, podman defaults to the user running the podman command.
 Podman prompts for the login password on the remote server.
 
+The `CONTAINER_SSHKEY` or `DOCKER_CERT_PATH` will override the default. Each is mapped to `CONTAINER_HOST` or `DOCKER_HOST`, respectively.
```
TomSweeneyRedHat, line 67: I'm not sure what this is saying.  Maybe:

```suggestion
If defined, either the `CONTAINER_SSHKEY` or the `DOCKER_CERT_PATH` environment variables will override the default path for the ssh identity file. Each is mapped to the `CONTAINER_HOST` and the `DOCKER_HOST`, environment variables respectively.
```
I'd add more on how the mapping works, but honestly IDK how that works and I'm not sure the average reader will.  Can you elaborate?

mheon:
Code LGTM

cmd/podman/root.go
```diff
@@ -289,13 +290,19 @@ func rootFlags(cmd *cobra.Command, opts *entities.PodmanConfig) {
...
-		var ident string
-		if v, found := os.LookupEnv("CONTAINER_SSHKEY"); found {
-			ident = v
+func activeDestination() (string, string) {
+	if uri, found := os.LookupEnv("DOCKER_HOST"); found {
```
rhatdan, line 294: What abouit CONTAINER_HOST and CONTAINER_SSHKEY?

AkihiroSuda:
This is confusing and incompatible. Please consider choosing other variable name.

rhatdan:
@AkihiroSuda Not sure what variable names you are saying CONTAINER_HOST and CONTAINER_SSHKEY?  These are already released, Although really not documented.

cmd/podman/root.go
rhatdan: Ok I see that these are defined in the connection code.

vrothberg:
> This is confusing and incompatible. Please consider choosing other variable name.

@AkihiroSuda, can you elaborate on that?

AkihiroSuda:
> $ DOCKER_HOST=ssh://root@localhost:22 bin/podman ps --all

I'd expect this command to connect to the Docker REST API via `ssh <USER>@<HOST> -- docker system dial-stdio`, which typically proxies `/var/run/docker.sock`.

jwhonce:
@AkihiroSuda Our customers have asked us to support DOCKER_HOST in the podman code to simplify supporting both products in their environment.

Your example provides a value for DOCKER_HOST consistent with this PR. The actual implementation of passing the bits between client and service is different between podman and docker.  We leverage systemd's socket activation protocol on the service side vs writing/embedding a proxy.

AkihiroSuda:
> The actual implementation of passing the bits between client and service is different between podman and docker.

I think ssh:// should be only supported in $CONTAINER_HOST then.

When ssh:// is specified in $DOCKER_HOST, Podman should error out or at least it should print logrus.Warn("Connecting to Podman, not Docker")

afbjorklund:
> @afbjorklund PTAL

I don't have any plans of using DOCKER_HOST for connecting to _podman_, but welcome not having to specify socket path...

Suppose there could be a problem with path if you want to support the regular system socket and rootless user sockets
But as we are mostly using it as cri-o interoperability, defaulting the path to `/run/podman/podman.sock` would be fine.

> - `user` will default to either `root` or current running user

> - `path` defaults to either `/run/podman/podman.sock`, or `/run/user/<uid>/podman/podman.sock` if running rootless.

When using something like minikube or vagrant, it uses a regular user for ssh access and then sudo/group for access.
This means that this proposed scheme would fail to detect the path, since it would try to run rootless if user is not "root".

It seems like DOCKER_CERT_PATH only works for certificates and not for ssh keys, so used `ssh-agent` for those instead.
Not sure where to configure host key handling, so I guess that has to be handled outside of this client ssh configuration ?

----

I documented my "machine" replacement here: https://boot2podman.github.io/2020/07/22/machine-replacement.html

`export CONTAINER_HOST=ssh://vagrant@127.0.0.1:2222/run/podman/podman.sock`

`export DOCKER_HOST=ssh://vagrant@127.0.0.1:2222 # /var/run/docker.sock`

Trying to specify the path for docker (in DOCKER_HOST) currently causes a runtime error, so it is just a comment above.

This was as a replacement for the deprecated podman-machine/docker-machine, with the old varlink and port 2376 access...
Previously it (docker) would use something like `tcp://192.168.99.100:2376`, but this is now deprecated in favor of `ssh://`

There are multiple network interfaces, so it prefers tunneling ssh over 2222 rather than just using `ssh://192.168.99.100:22`

rhatdan:
@jwhonce Please rebase this PR.

rhatdan:
I am taking this PR over,  #7620 to get it merged.

rhatdan:
Well actually I am rethinking the DOCKER_HOST support altogether and just grabbing the connection documentation.

Issue #6987, opened 2020-07-15
CI flake: run --preserve-fds: attach / unixpacket / bad file descriptor


edsantiago:
Flake seen three times on CI for #6958. Passed on re-run:
```
Running: podman [options] run --preserve-fds 1 docker.io/library/alpine:latest ls
time="2020-07-14T16:51:22-04:00" level=error msg="unable to close socket: \"close unixpacket @->/tmp/podman_test328676400/socket/7d5388edb2bed82a1bfbd8d549e58fa977154dfd03c5f6dc4ceadac06597f15c/attach: bad file descriptor\""
Error: error attaching to container 7d5388edb2bed82a1bfbd8d549e58fa977154dfd03c5f6dc4ceadac06597f15c: read unixpacket @->/tmp/podman_test328676400/socket/7d5388edb2bed82a1bfbd8d549e58fa977154dfd03c5f6dc4ceadac06597f15c/attach: read: bad file descriptor
```
Log: https://api.cirrus-ci.com/v1/task/6481256420474880/logs/integration_test.log#t--podman-run---preserve-fds

This was in "test fedora-31". cirrus-flake-xref finds no other instances of this in the last three weeks.

The error looks suspiciously close to #5769 (fixed last month), but that was `exec`; this is `run`. Possibly because of #6625

edsantiago:
@QiWang19 PTAL

edsantiago:
Seeing it on f32 now. Three failures, so it caused a CI flake. @QiWang19 PTAL.

* fedora-32-podman-6439450735542272 : test fedora-32
  * PR #7085
    * [07-27 17:36](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5376945888165888/html/integration_test.log.html#t--podman-run---preserve-fds--1)
    * [07-27 17:36](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5376945888165888/html/integration_test.log.html#t--podman-run---preserve-fds--2)
    * [07-27 17:36](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5376945888165888/html/integration_test.log.html#t--podman-run---preserve-fds--3)

QiWang19:
This flake should be fixed by https://github.com/containers/podman/pull/7120

PR #7120, opened 2020-07-28
Fix close fds of run --preserve-fds

QiWang19:
Test flakes mentioned in #6987 might be caused by uncorrect closing of file descriptor.
Remove the code to close file descriptors for `podman run` since it may close those used by other processes.

@giuseppe PTAL. The same flakes happened to podman exec once and got fixed, but does not work for `podman run`, is this a feasible fix?

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -954,9 +954,12 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		return err
 	}
 
+	var filesToClose []*os.File
 	if ctr.config.PreserveFDs > 0 {
 		for fd := 3; fd < int(3+ctr.config.PreserveFDs); fd++ {
-			cmd.ExtraFiles = append(cmd.ExtraFiles, os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd)))
+			f := os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd))
+			filesToClose = append(filesToClose, f)
+			cmd.ExtraFiles = append(cmd.ExtraFiles, f)
 		}
 	}
 
@@ -1052,14 +1055,10 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		}
 	}
 
-	if ctr.config.PreserveFDs > 0 {
-		for fd := 3; fd < int(3+ctr.config.PreserveFDs); fd++ {
-			// These fds were passed down to the runtime.  Close them
-			// and not interfere
-			if err := os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd)).Close(); err != nil {
-				logrus.Debugf("unable to close file fd-%d", fd)
-			}
-		}
+	// These fds were passed down to the runtime.  Close them
+	// and not interfere
+	for _, f := range filesToClose {
+		errorhandling.CloseQuiet(f)
 	}
 
 	return nil


rhatdan:
@giuseppe PTAL, We are seeing random flakes, we believe caused by this code.  Do you know if it is necessary.
Basically we are seeing file descriptors like boltdb database being closed when this code runs.

@edsantiago @mheon PTAL

giuseppe:
If we remove this code, we end up keeping the fd open and possibly cause the other end of the pipe to hang indefinitely.

Are we making sure the fds are really opened in the tests when we specify `--preserve-fds=X`?  In bash it can be easily done with:

```
$ exec 3</dev/null
$ podman run --preserve-fds=1 ....
```

if we don't do that, then preserve-fds will pass down and close fd 3 opened by podman.

edsantiago:
It's not the bash tests that are failing (AFAICT those have never flaked), it's the Ginkgo tests. And [that code](https://github.com/containers/podman/blob/8f7ed50cb20c04bbbb7c4907a183c07912d4bffb/test/e2e/run_test.go#L1042) really looks a lot like the one you fixed for the `exec` flake.

QiWang19:
the uni test opens the file 
https://github.com/containers/podman/blob/288ebec6e737c105fa0ef43412de4e0a8997feb9/test/e2e/run_test.go#L1043

giuseppe:
could you check if the following patch has the same effect?

```
diff --git a/libpod/oci_conmon_linux.go b/libpod/oci_conmon_linux.go
index bd6af5281..48cf02749 100644
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -954,9 +954,12 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
                return err
        }
 
+       var filesToClose []*os.File
        if ctr.config.PreserveFDs > 0 {
                for fd := 3; fd < int(3+ctr.config.PreserveFDs); fd++ {
-                       cmd.ExtraFiles = append(cmd.ExtraFiles, os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd)))
+                       f := os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd))
+                       filesToClose = append(filesToClose, f)
+                       cmd.ExtraFiles = append(cmd.ExtraFiles, f)
                }
        }
 
@@ -1052,14 +1055,10 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
                }
        }
 
-       if ctr.config.PreserveFDs > 0 {
-               for fd := 3; fd < int(3+ctr.config.PreserveFDs); fd++ {
-                       // These fds were passed down to the runtime.  Close them
-                       // and not interfere
-                       if err := os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd)).Close(); err != nil {
-                               logrus.Debugf("unable to close file fd-%d", fd)
-                       }
-               }
+       // These fds were passed down to the runtime.  Close them
+       // and not interfere
+       for _, f := range filesToClose {
+               errorhandling.CloseQuiet(f)
        }
```

if it works fine for you, please feel free to amend it in your PR.  I think libpod/oci_conmon_exec_linux.go needs the same fix.

giuseppe:
LGTM, can we get the same fix for libpod/oci_conmon_exec_linux.go  ?

QiWang19:
> LGTM, can we get the same fix for libpod/oci_conmon_exec_linux.go ?

Yes, will open a PR for that.

rhatdan:
/lgtm
@QiWang19 Did you open the second PR?

Issue #7107, opened 2020-07-28
CLI inconsistency with docker build


georgmu:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. Use podman as a docker replacement
```
ln -s /usr/local/bin/docker /usr/bin/podman
```

2. Build a new docker container from stdin
```
echo -e "FROM debian:testing\nRUN /bin/echo 'Test'" | docker build -t debian-test -
```

**Describe the results you received:**

With podman as docker replacement, I get the following error:
```
Error: error reading info about "/var/tmp/buildah510720429/var/tmp/buildah510720429/Dockerfile": stat /var/tmp/buildah510720429/var/tmp/buildah510720429/Dockerfile: no such file or directory
```

**Describe the results you expected:**

With docker, the container is built and tagged:
```
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM debian:testing
testing: Pulling from library/debian
73b408cb748b: Pull complete 
Digest: sha256:3a901d53cb199478770f3fafd8cdce130b602e3f9c7695516702a39f4ab926ca
Status: Downloaded newer image for debian:testing
 ---> 03c237d6b0a4
Step 2/2 : RUN /bin/echo 'Test'
 ---> Running in 1bf2c746e69d
Test
Removing intermediate container 1bf2c746e69d
 ---> e78872fbeaef
Successfully built e78872fbeaef
Successfully tagged debian-test:latest
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.3
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

Podman works by using `podman build ... -f -`, but this does not work with docker. Since there are scripts which are primarily targeted at docker (e.g. the travis-ci script in systemd) and use the docker-compatible syntax, I think podman is the right place to align its CLI with docker.

vrothberg:
Thanks for opening the issue, @georgmu!

@TomSweeneyRedHat @nalind PTAL

rhatdan:
This works:

echo -e "FROM debian:testing\nRUN /bin/echo 'Test'" | buildah bud -t debian-test -

So this looks like a podman bug.

TomSweeneyRedHat:
[...]

PR #7113, opened 2020-07-28
fix build err while reading from stdin

zhangguanzhang:
fix https://github.com/containers/podman/issues/7107
Signed-off-by: zhangguanzhang <zhangguanzhang@qq.com>

Issue #6518, opened 2020-06-08
CI flake: Panic in Spec Teardown: send on closed channel


edsantiago:
cirrus-flake-xref is reporting three instances of this since May 26:
```
Panic in Spec Teardown (AfterEach) [2.677 seconds]
         Podman containers 
         /var/tmp/go/src/github.com/containers/libpod/pkg/bindings/test/containers_test.go:19
           podman wait to pause|unpause condition [AfterEach]
           /var/tmp/go/src/github.com/containers/libpod/pkg/bindings/test/containers_test.go:282
         
           Test Panicked
           send on closed channel
           /usr/lib/golang/src/runtime/panic.go:969
         
           Full Stack Trace
           panic(0x10397a0, 0x134e430)
           	/usr/lib/golang/src/runtime/panic.go:969 +0x166
           github.com/containers/libpod/pkg/bindings/test.glob..func5.20.2(0xc0000b2a38, 0xc00004dc90, 0xc000319ae8, 0xc000011180)
           	/var/tmp/go/src/github.com/containers/libpod/pkg/bindings/test/containers_test.go:308 +0xc8
           created by github.com/containers/libpod/pkg/bindings/test.glob..func5.20
           	/var/tmp/go/src/github.com/containers/libpod/pkg/bindings/test/containers_test.go:304 +0x6cc
```
Log links:
* [PR 6356](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/4615038243700736/html/integration_test.log.html)
* [PR 6453](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/5392977415634944/html/integration_test.log.html)
* [PR 6515](https://storage.googleapis.com/cirrus-ci-5385732420009984-fcae48/artifacts/containers/libpod/4547364012687360/html/integration_test.log.html)

Link to [containers_test.go:308](https://github.com/containers/libpod/blob/a5facca78fed010552339a6644dd8ef1bb0fa229/pkg/bindings/test/containers_test.go#L308).

All failures have been in `fedora-32 special_testing_bindings`.

edsantiago:
Still happening, and (as far as my logs can tell) still only on f32:

Podman containers  [AfterEach] podman wait to pause|unpause condition
* fedora-32 : special_testing_bindings
  * PR #6733
[...]

PR #7121, opened 2020-07-28
Fix building from http or '-' options

rhatdan:
When copying from a URL, podman will download and create a context
directory in a temporary file.  The problem was that this directory
was being removed as soon as the function that created it was returned.

Later the build code would look for content in the temporary directory
and fail to find it, blowing up the build.

By pulling the extraction code back into the build function, we keep the
temporary directory around until the build completes.

### DIFF
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -177,6 +177,27 @@ Labels.$label_name | $label_value
     run_podman rmi -f build_test
 }
 
+@test "podman build - stdin test" {
+    if is_remote && is_rootless; then
+        skip "unreliable with podman-remote and rootless; #2972"
+    fi
+
+    # Random workdir, and multiple random strings to verify command & env
+    workdir=/$(random_string 10)
+    PODMAN_TIMEOUT=240 run_podman build -t build_test - << EOF
+FROM  $IMAGE
+RUN mkdir $workdir
+WORKDIR $workdir
+RUN /bin/echo 'Test'
+EOF
+    is "$output" ".*STEP 5: COMMIT" "COMMIT seen in log"
+
+    run_podman run --rm build_test pwd
+    is "$output" "$workdir" "pwd command in container"
+
+    run_podman rmi -f build_test
+}
+
 function teardown() {
     # A timeout or other error in 'build' can leave behind stale images
     # that podman can't even see and which will cascade into subsequent
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -138,36 +138,9 @@ func build(cmd *cobra.Command, args []string) error {
 		return errors.New("cannot specify --squash, --squash-all and --layers options together")
 	}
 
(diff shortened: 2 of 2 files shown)


rhatdan:
An alternative to https://github.com/containers/podman/pull/7113

rhatdan:
fixes #7107

test/system/070-build.bats
```diff
@@ -177,6 +177,20 @@ Labels.$label_name | $label_value
     run_podman rmi -f build_test
 }
 
+@test "podman build - stdin test" {
+    if is_remote && is_rootless; then
+        skip "unreliable with podman-remote and rootless; #2972"
+    fi
+
+    PODMAN_TIMEOUT=240 run_podman build -t build_test - << EOF
+FROM alpine
```
edsantiago, line 186: Could you make this $IMAGE (instead of alpine), to avoid hitting a registry?

test/system/070-build.bats
```diff
@@ -177,6 +177,20 @@ Labels.$label_name | $label_value
...
+RUN /bin/echo 'Test'
+EOF
+    is "$output" ".*STEP 3: COMMIT" "COMMIT seen in log"
+
```
edsantiago, line 190: Why not add a `podman run` test? Verify that `is "$output" "Test" "blah blah"`? Costs nothing, and could (maybe??) catch some weird future error.

test/system/070-build.bats
rhatdan: If that works, sure.

test/system/070-build.bats
rhatdan: Added

TomSweeneyRedHat:
Looks like a panic in one of the tests, I don't know if it's a known issue or not:

```
[+0535s]   /usr/lib/golang/src/runtime/panic.go:969
[+0535s] 
[+0535s]   Full Stack Trace
[+0535s]   panic(0x1043c20, 0x135acf0)
[+0535s]   	/usr/lib/golang/src/runtime/panic.go:969 +0x166
[+0535s]   github.com/containers/podman/v2/pkg/bindings/test.glob..func5.20.2(0xc0001228f8, 0xc0001effa0, 0xc000bd61f8, 0xc000df0950)
[+0535s]   	/var/tmp/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:308 +0xc8
[+0535s]   created by github.com/containers/podman/v2/pkg/bindings/test.glob..func5.20
[+0535s]   	/var/tmp/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:304 +0x6cc
```

edsantiago:
It's a known flake, #6518. It is really hammering us hard lately. I've restarted the bindings test.

rhatdan:
@baude @mheon @giuseppe @vrothberg @TomSweeneyRedHat @edsantiago @jwhonce PTAL

QiWang19:
This patch also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1858862

Issue #6653, opened 2020-06-17
podman exec --preserve-fds=2: panic


edsantiago:
I know, I know, this is undefined, but it probably shouldn't panic:
```console
$ ./bin/podman run --name test -d alpine top
3e96d0911deb459b8549fcc31b0124a13d18340b5dfbcc7c68791723acc0696c
$ ./bin/podman exec --preserve-fds=2 -i test date
runtime: epollwait on fd 4 failed with 9
fatal error: runtime: netpoll failed

runtime stack:
runtime.throw(0x1a568ef, 0x17)
        /usr/lib/golang/src/runtime/panic.go:1116 +0x72
runtime.netpoll(0x0, 0x0)
        /usr/lib/golang/src/runtime/netpoll_epoll.go:123 +0x363
runtime.findrunnable(0xc00005f800, 0x0)
        /usr/lib/golang/src/runtime/proc.go:2126 +0xc60
runtime.schedule()
        /usr/lib/golang/src/runtime/proc.go:2520 +0x2fc
runtime.park_m(0xc0005a2300)
        /usr/lib/golang/src/runtime/proc.go:2690 +0x9d
runtime.mcall(0x800000)
        /usr/lib/golang/src/runtime/asm_amd64.s:318 +0x5b
[lots and lots more]
```
With just `1`, behavior is inconsistent:
```console
$ ./bin/podman exec --preserve-fds=1 -i test date
Wed Jun 17 17:08:50 UTC 2020
Error: read /home/esm/.local/share/containers/storage/overlay-containers/3e96d0911deb459b8549fcc31b0124a13d18340b5dfbcc7c68791723acc0696c/userdata/bb1ef4c6c8a7b4ab51338ec78fe8f5696cdf3d36a3b00665d5bb229cb2801933/exit/3e96d0911deb459b8549fcc31b0124a13d18340b5dfbcc7c68791723acc0696c: bad file descriptor
```
or
```console
$ ./bin/podman exec --preserve-fds=1 -i test date
Wed Jun 17 17:09:09 UTC 2020
2020/06/17 11:09:09 bolt.Close(): funlock error: bad file descriptor
ERRO[0000] failed to close libpod db: "db file close: close /home/esm/.local/share/containers/storage/libpod/bolt_state.db: bad file descriptor"
```
or sometimes even no error at all
```console
$ ./bin/podman exec --preserve-fds=1 -i test date
Wed Jun 17 17:09:10 UTC 2020
```
[...]

PR #7125, opened 2020-07-28
validate fds --preserve-fds

QiWang19:
validate file descriptors passed from `podman run` and `podman exec` --preserve-fds.

### DIFF
--- a/cmd/podman/containers/exec.go
+++ b/cmd/podman/containers/exec.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	envLib "github.com/containers/podman/v2/pkg/env"
+	"github.com/containers/podman/v2/pkg/rootless"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -109,6 +110,12 @@ func exec(_ *cobra.Command, args []string) error {
 	}
 
 	execOpts.Envs = envLib.Join(execOpts.Envs, cliEnv)
+
+	for fd := 3; fd < int(3+execOpts.PreserveFDs); fd++ {
+		if !rootless.IsFdInherited(fd) {
+			return errors.Errorf("file descriptor %d is not available - the preserve-fds option requires that file descriptors must be passed", fd)
+		}
+	}
 
 	if !execDetach {
 		streams := define.AttachStreams{}
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -125,6 +125,11 @@ func run(cmd *cobra.Command, args []string) error {
 	if err := createInit(cmd); err != nil {
 		return err
 	}
+	for fd := 3; fd < int(3+runOpts.PreserveFDs); fd++ {
+		if !rootless.IsFdInherited(fd) {
+			return errors.Errorf("file descriptor %d is not available - the preserve-fds option requires that file descriptors must be passed", fd)
+		}
+	}
 
 	imageName := args[0]
 	if !cliVals.RootFS {
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -225,6 +225,16 @@ can_use_shortcut ()
   return ret;
 }
 
+int
+is_fd_inherited(int fd)
+{
+  if (open_files_set == NULL || fd > open_files_max_fd || fd < 0)
+  {
+    return 0;
+  }
+  return FD_ISSET(fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])) ? 1 : 0;
+}
+
 static void __attribute__((constructor)) init()
 {
   const char *xdg_runtime_dir;
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -32,6 +32,7 @@ extern uid_t rootless_gid();
 extern int reexec_in_user_namespace(int ready, char *pause_pid_file_path, char *file_to_read, int fd);
 extern int reexec_in_user_namespace_wait(int pid, int options);
 extern int reexec_userns_join(int pid, char *pause_pid_file_path);
+extern int is_fd_inherited(int fd);
 */
 import "C"
 
@@ -519,4 +520,9 @@ func ConfigurationMatches() (bool, error) {
 	}
 
 	return matches(GetRootlessGID(), gids, currentGIDs), nil
+}
+
+// IsFdInherited checks whether the fd is opened and valid to use
+func IsFdInherited(fd int) bool {
+	return int(C.is_fd_inherited(C.int(fd))) > 0
 }
--- a/pkg/rootless/rootless_unsupported.go
+++ b/pkg/rootless/rootless_unsupported.go
@@ -64,3 +64,8 @@ func GetConfiguredMappings() ([]idtools.IDMap, []idtools.IDMap, error) {
 func ReadMappingsProc(path string) ([]idtools.IDMap, error) {
 	return nil, nil
 }
+
+// IsFdInherited checks whether the fd is opened and valid to use
+func IsFdInherited(fd int) bool {
+	return false
+}
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -1063,6 +1063,13 @@ USER mail`
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman run --preserve-fds invalid fd", func() {
+		session := podmanTest.Podman([]string{"run", "--preserve-fds", "2", ALPINE})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Not(Equal(0)))
+		Expect(session.ErrorToString()).To(ContainSubstring("file descriptor 3 is not available"))
+	})
+
 	It("podman run --privileged and --group-add", func() {
 		groupName := "kvm"
 		session := podmanTest.Podman([]string{"run", "-t", "-i", "--group-add", groupName, "--privileged", fedoraMinimal, "groups"})
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -63,7 +63,6 @@ echo $rand        |   0 | $rand
 
 # 'run --preserve-fds' passes a number of additional file descriptors into the container
 @test "podman run --preserve-fds" {
-    skip "enable this once #6653 is fixed"
     skip_if_remote
 
     content=$(random_string 20)


libpod/oci_conmon_linux.go
```diff
@@ -956,6 +956,11 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 
 	if ctr.config.PreserveFDs > 0 {
 		for fd := 3; fd < int(3+ctr.config.PreserveFDs); fd++ {
+			f := os.NewFile(uintptr(fd), fmt.Sprintf("fd-%d", fd))
```
mheon, line 959: I think we need to do this in `cmd/podman` - we probably have at least this many open FDs of our own by this point (c/storage locks, etc)

QiWang19:
Add validation code. But this does not fix panic #6653
and system tests still fail.
```
[+0116s] # #|     FAIL: container read input from fd 4
[+0116s] # #| expected: 'MSdO93BIGQj5i44nOS4D'
[+0116s] # #|   actual: 'time="2020-07-29T15:16:22-04:00" level=error msg="unable to close file fd-4: \"close fd-4: bad file descriptor\""'
[+0116s] # #|         > 'time="2020-07-29T15:16:22-04:00" level=error msg="unable to cleanup network for container a22fc3e3b4aed17632f3861dfa409fadce23f10b40395526679b67f2a2d1c952: \"error closing network namespace for container a22fc3e3b4aed17632f3861dfa409fadce23f10b40395526679b67f2a2d1c952: Failed to close \\\"/run/user/30150/netns/cni-b4c7ec7e-c192-bc0a-c8e5-992b3c85d99f\\\": close /run/user/30150/netns/cni-b4c7ec7e-c192-bc0a-c8e5-992b3c85d99f: bad file descriptor\""'
```

mheon:
Go may well be happy to create a new file from an invalid FD - we probably need a test that the FD is actually valid.

rhatdan:
Yes look like it succeeds.

```
package main

import (
	"fmt"
	"os"
)

func main() {
	fd1 := os.NewFile(1, "stdout")
	if fd1 == nil {
		fmt.Println("stdout Failed")
	}
	fd5 := os.NewFile(500, "failure")
	if fd5 == nil {
		fmt.Println("failure succeeded")
	} else {
		fmt.Println("failure failed", fd5)
	}

}
```

edsantiago:
Just a heads-up: please see #7144, it looks germane.

rhatdan:
This works.
```
package main

import (
	"fmt"
	"golang.org/x/sys/unix"
)

func main() {

	if _, err := unix.FcntlInt(1, unix.F_GETFL, 0); err != nil {
		fmt.Println(1, err)
	}
	if _, err := unix.FcntlInt(400, unix.F_GETFL, 0); err != nil {
		fmt.Println(400, err)
		if err == unix.EBADF {
			fmt.Println("BINGO")
		} else {
			fmt.Println("BoNGO")
		}

	}

```

rhatdan:
The call seems to return err = 0 on success.  Not nil, but I think as long as we check for EBADF, then we should be safe.

QiWang19:
I didn't see the output from this success case, I guess the err is nil, can't we just return err as long as the err != nil, without checking EBADF?
```
if _, err := unix.FcntlInt(1, unix.F_GETFL, 0); err != nil {
		fmt.Println(1, err)
	}
```

rhatdan:
The preserve-fds should fail on remote client and should also not be executed on Windows.
When I was testing my test program, I was constantly seeing err != nil, Not sure why.  But I added the check for EBADF to make sure the error we were looking for was correct.

rhatdan:
How about?
	if _, err := unix.FcntlInt(fd, unix.F_GETFL, 0); err != nil && err != syscall.Errno(0) {
	             return errors.Wrapf(err, "failed to verify %d file descriptor", fd)
	}

cmd/podman/containers/run.go
```diff
@@ -125,6 +126,12 @@ func run(cmd *cobra.Command, args []string) error {
 	if err := createInit(cmd); err != nil {
 		return err
 	}
+	var fd uint
+	for fd = 3; fd < 3+runOpts.PreserveFDs; fd++ {
+		_, err := unix.FcntlInt(uintptr(fd), unix.F_GETFL, 0); err != nil && err != syscall.Errno(0) {
+			return err
```
rhatdan, line 132: I think this should be a wrapped error to tell the user which FD is invalid.

rhatdan: The FcntlInt code needs to be in a util_linux.go branch.
Does podman-remote throw an error when you do --preserver-fds?

QiWang19: Added utils_linux.go  `--preserver-fds` is disabled in remote.

rhatdan:
Your going to need verifyfd in the utils_unsupported.go as well.

cmd/podman/containers/run.go
rhatdan: I think you want to have similar code in exec.go and perhaps attach.go as well?

rhatdan:
Almost there, now I need a test to make sure it errors out with the correct error

podman run --preserve-fd 2 alpine true

Should return an error containing bad fd.

cmd/podman/containers/utils_linux.go
```diff
@@ -0,0 +1,15 @@
...
+	"syscall"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+func verifyFD(fd int) error {
+	if _, err := unix.FcntlInt(uintptr(fd), unix.F_GETFL, 0); err != nil && err != syscall.Errno(0) {
+		return errors.Wrapf(err, "failed to verify %d file descriptor", fd)
+	}
```
giuseppe, line 13: this won't detect fds that are opened by Podman.

QiWang19: If run `podman run --preserve-fds`, it's possible that the file descriptor I want to pass has already opened by podman, so this won't detect that?

giuseppe:
I don't think it can be solved at this level.

Files can be opened by the Go runtime so we don't know what FDs are valid and what are not.

rootless already stores at startup what ds were open before the Go runtime kicks in.  I think we could take advantage of that and be sure that these fds were already opened when the program started (`open_files_set` in rootless_linux.c).

rhatdan:
Do we have access to these within golang?

giuseppe:
> Do we have access to these within golang?

we will need to provide a wrapper, something like:

```diff
diff --git a/pkg/rootless/rootless_linux.c b/pkg/rootless/rootless_linux.c
index d3e43e44d..6d353d2e2 100644
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -225,6 +225,14 @@ can_use_shortcut ()
   return ret;
 }
 
+int
+is_fd_inherited(int fd)
+{
+  if (open_files_set == NULL || fd >= open_files_max_fd || fd < 0)
+    return false;
+  return FD_ISSET (fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])) ? 1 : 0;
+}
+
 static void __attribute__((constructor)) init()
 {
   const char *xdg_runtime_dir;
diff --git a/pkg/rootless/rootless_linux.go b/pkg/rootless/rootless_linux.go
index b1f200cc2..2fe915cab 100644
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -32,6 +32,7 @@ extern uid_t rootless_gid();
 extern int reexec_in_user_namespace(int ready, char *pause_pid_file_path, char *file_to_read, int fd);
 extern int reexec_in_user_namespace_wait(int pid, int options);
 extern int reexec_userns_join(int pid, char *pause_pid_file_path);
+extern int is_fd_inherited(int fd);
 */
 import "C"
 
@@ -495,3 +496,10 @@ func ConfigurationMatches() (bool, error) {
 
        return matches(GetRootlessGID(), gids, currentGIDs), nil
 }
+
+func IsFdInherited(fd int) bool {
+       if int(C.is_fd_inherited(C.int(fd))) > 0 {
+               return true
+       }
+       return false
+}
diff --git a/pkg/rootless/rootless_unsupported.go b/pkg/rootless/rootless_unsupported.go
index 1499b737f..203d56acb 100644
--- a/pkg/rootless/rootless_unsupported.go
+++ b/pkg/rootless/rootless_unsupported.go
@@ -64,3 +64,7 @@ func GetConfiguredMappings() ([]idtools.IDMap, []idtools.IDMap, error) {
 func ReadMappingsProc(path string) ([]idtools.IDMap, error) {
        return nil, nil
 }
+
+func IsFdInherited(fd int) bool {
+       return false
+}

```

QiWang19:
@giuseppe Are you going to open PR to fix this? Do you mind I grab the code above PR?

giuseppe:
> @giuseppe Are you going to open PR to fix this? Do you mind I grab the code above PR?

please go forward and use it in this PR

QiWang19:
@giuseppe Any ideas why the test fail? `podman run --preserve-fds` & `podman exec preserve fds sanity check` https://api.cirrus-ci.com/v1/task/5511032351227904/logs/integration_test.log

giuseppe:
> @giuseppe Any ideas why the test fail? `podman run --preserve-fds` & `podman exec preserve fds sanity check` https://api.cirrus-ci.com/v1/task/5511032351227904/logs/integration_test.log

looks like there is an issue in the code I've provided you, can you try with this version for `is_fd_inherited`?

```

int
is_fd_inherited(int fd)
{
  if (open_files_set == NULL || fd > open_files_max_fd || fd < 0)
    return 0;

  return FD_ISSET(fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])) ? 1 : 0;
}
```

QiWang19:
@edsantiago @rhatdan @giuseppe PTAL

rhatdan:
/approve
LGTM
If you need to do another PR, you should check the error message in your test to make sure it says something like
`failed to verify %d file descriptor`

cmd/podman/containers/exec.go
```diff
@@ -110,6 +111,12 @@ func exec(_ *cobra.Command, args []string) error {
 
 	execOpts.Envs = envLib.Join(execOpts.Envs, cliEnv)
 
+	for fd := 3; fd < int(3+execOpts.PreserveFDs); fd++ {
+		if !rootless.IsFdInherited(fd) {
+			return errors.Errorf("failed to verify %d file descriptor", fd)
```
edsantiago, line 116: This is confusing. Shouldn't it read `file descriptor %d`?

cmd/podman/containers/run.go
```diff
@@ -125,6 +125,11 @@ func run(cmd *cobra.Command, args []string) error {
 	if err := createInit(cmd); err != nil {
 		return err
 	}
+	for fd := 3; fd < int(3+runOpts.PreserveFDs); fd++ {
+		if !rootless.IsFdInherited(fd) {
+			return errors.Errorf("failed to verify %d file descriptor", fd)
```
edsantiago, line 130: same here

cmd/podman/containers/exec.go
edsantiago: Actually" failed to verify" itself doesn't really make sense to me as an end user. How about "file descriptor %d is not open" or "...is not available for I/O" or something along those lines? @giuseppe since you have the best understanding of what this code is actually validating, could you suggest a phrasing that is both technically accurate and helpful to the end user?

rhatdan: The code is verifying it a file descriptor actually exists.  I think "file descriptor %d is not open"  is correct.
Users are telling podman to leak file descriptor 3 and 4 in to a container, and if 3 and 4 are not open, then something went wrong.

mheon: Maybe `"file descriptor %d is not available - the preserve-fds option requires that file descriptors must be passed"`

QiWang19:
@edsantiago @rhatdan @mheon PTAL

rhatdan:
LGTM

PR #7131, opened 2020-07-29
Fix `podman image search` missing description

rhafer:
`podman image search` returned wrong results for the image "Description" as
it was mapped to the wrong field ("ID") in the search results.


rhafer:
> Code LGTM, but let's modify/add a test for this failure-case.

Agreed, I am currently looking at that.

rhatdan:
/lgtm
/approve
Lets get this in, and handle the test in a different PR.

PR #7133, opened 2020-07-29
Stop pause process if runtime shuts down

saschagrunert:
To avoid having a stale process running after invocations like `podman ps` we now shutdown the pause process on runtime shutdown.

@giuseppe do we have to add any checks for running pods/containers? I do not wanna break-up anything while removing the pause.pid

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -562,17 +562,30 @@ func (r *Runtime) Shutdown(force bool) error {
 
 	r.valid = false
 
-	// Shutdown all containers if --force is given
-	if force {
-		ctrs, err := r.state.AllContainers()
-		if err != nil {
-			logrus.Errorf("Error retrieving containers from database: %v", err)
-		} else {
-			for _, ctr := range ctrs {
-				if err := ctr.StopWithTimeout(r.config.Engine.StopTimeout); err != nil {
-					logrus.Errorf("Error stopping container %s: %v", ctr.ID(), err)
-				}
+	ctrs, err := r.state.AllContainers()
+	if err != nil {
+		return errors.Wrap(err, "retrieve containers from database")
+	}
+	hasRunningContainers := false
+	for _, ctr := range ctrs {
+		// Shutdown all containers if --force is given
+		if force {
+			if err := ctr.StopWithTimeout(r.config.Engine.StopTimeout); err != nil {
+				logrus.Errorf("Error stopping container %s: %v", ctr.ID(), err)
 			}
+		}
+
+		state, err := ctr.State()
+		if err != nil {
+			logrus.Errorf("Unable to retrieve container state: %v", err)
+		}
+		if state == define.ContainerStateRunning {
+			hasRunningContainers = true
+		}
+	}
+	if !hasRunningContainers {
+		if err := stopPauseProcess(); err != nil {
+			return errors.Wrap(err, "stop pause process")
 		}
 	}
 


saschagrunert:
Let's see if this breaks anything…

TomSweeneyRedHat:
@saschagrunert looks like a few tests aren't hip

mheon:
We call runtime.Shutdown() unconditionally at the end of all commands, so this is going to kill the pause process far too aggressively. It definitely needs to be gated on having running no running containers.

saschagrunert:
I agree and will add a check if no containers are running before shutdown.

libpod/runtime.go
```diff
@@ -562,18 +562,31 @@ func (r *Runtime) Shutdown(force bool) error {
...
+		state, err := ctr.State()
+		if err != nil {
+			logrus.Errorf("Unable to retrieve container state: %v", err)
+		}
+		if state == define.ContainerStateRunning {
+			hasRunningContainers = true
+		}
+	}
+	if !hasRunningContainers {
+		if err := stopPauseProcess(); err != nil {
```
rhatdan, line 587: Isn't this opening us up for a new race condition?  What happens if I kill the pause process while another container is starting up and joining the pause process?

saschagrunert: Isn't this covered by the runtime lock in Line 556?

mheon: That lock is purely single-process - it's basically just for mutable state within a single instance of Libpod, it does not give us any guarantee there isn't another Podman process in the process of starting a container.

giuseppe: yeah that is the reason why it is kept running :(  it can potentially open a race condition and two containers end up in two differerent user namespaces.

It would be nice if systemd could keep the fd available for us, without needing any pause process.

The idea would be to pass the namespace file descriptor to the systemd user instance and not require a pause process (at least when systemd is available) at all.

I've opened an issue here: https://github.com/systemd/systemd/issues/16318

saschagrunert: Okay sounds good thank you for the clarification.

saschagrunert:
Unfortunately I do not see a way forward with this PR, yet. Closing for now.

rhatdan:
@giuseppe @haircommander Would the pinns work going on in CRI-O help with this at all?

giuseppe:
> @giuseppe @haircommander Would the pinns work going on in CRI-O help with this at all?

no, rootless users are not able to create mounts required for pinning namespaces in the host mount ns.  It can be used only once we are already in the new user+mount namespace

rhatdan:
Ok, I wonder if this is something we should bring up with the kernel team?  IE Is there a security problem without creating a mount to pin a namespace?

giuseppe:
> Ok, I wonder if this is something we should bring up with the kernel team? IE Is there a security problem without creating a mount to pin a namespace?

it will make easier to leak namespaces.  Currently it is enough to kill all the processes for a user and we are sure there are no namespaces left around.

sampie:
The problem still occurs on Ubuntu 20.10 and CentOS 7. In clustered computer environments programs should not left running in anode after a work completes. I am wondering what happens with SLURM if a job leaves a process running. Does it prevent job from completing or cause any other problems. Has this behavior been tested with SLURM?

mheon:
We still don't have a way of reliably shutting down the pause process without racing against potential new instances of Podman starting, and this is unlikely to be fixed without that.

sampie:
It seems the related feature request to systemd is still open, so I guess there is still hope that a feasible technical solution will be eventually found to fix this issue.

PR #7138, opened 2020-07-29
Cirrus: Add python packages to images

cevich:
Python packages per @jwhonce request and Ubuntu 19.10 golang 1.13 for @rhatdan in buildah (per https://github.com/containers/buildah/pull/2462#issuecomment-659012120 )


rhatdan:
LGTM
/approve
@vrothberg @mheon @TomSweeneyRedHat @jwhonce @ashley-cui @QiWang19 @edsantiago PTAL

contrib/cirrus/packer/ubuntu_packaging.sh
```diff
@@ -144,22 +150,23 @@ $BIGTO ooe.sh $SUDOAPTGET upgrade
...
     $SUDO ln -s /var/cache/apt/archives "$PACKAGE_DOWNLOAD_DIR"
-    $LILTO ooe.sh $SUDOAPTGET install --download-only ${DOWNLOAD_PACKAGES[@]}
-    ls -la "$PACKAGE_DOWNLOAD_DIR/"
+    $LILTO ooe.sh $SUDOAPTGET install --download-only "${DOWNLOAD_PACKAGES[@]}"
 fi
 
-echo "Installing runtime tooling"
-# Save some runtime by having these already available
+echo "Configuring Go environment"
+$SUDO ln -s /usr/lib/go-1.14/bin/go /usr/bin/go
```
edsantiago, line 162: This is a little scary, can you explain why it's needed? Is there another way to do it? And if not, would it be a good idea to use `ln -sf` in case there's already a `/usr/bin/go`?

contrib/cirrus/packer/fedora_packaging.sh
```diff
@@ -37,7 +37,16 @@ fi
...
 $BIGTO ooe.sh $SUDO dnf update -y
 
+# While both runc and crun are needed in the support of different distros
+# and environments, packaging will need to instal one and suggesting the other.
+# This can cause some ambiguity for automation, where more concrete, predictable
+# behavior is desirable.  Instead of fighting with or complicating the packaging
+# itself, workaround the problem.  Always removing 'runc' in Fedora since it's
+# exclusively cgroupsv2 focused as of release 31.  However, ensure the runc
+# is downloaded to $PACKAGE_DOWNLOAD_DIR for the minority of use-cases where
+# automation needs to swap it in.
```
edsantiago, line 47: I'm having a lot of trouble parsing this. Might I suggest something like:
```
# Fedora, as of 31, uses cgroups v2 by default. runc does not support 
# cgroups v2, only crun does. (As of 2020-07-30 runc support is
# forthcoming but not even close to ready yet). To ensure a reliable
# runtime environment, force-remove runc if it is present.
# (Something something something automation, whatever that means)
```

edsantiago:
Is this for `test/apiv2/rest_api`? Did you verify (is there a way to verify) if those tests run on the new VMs? I have a vague memory that the python tests required some sort of `docker.py` package -- @jwhonce is that no longer the case?

contrib/cirrus/packer/ubuntu_packaging.sh
cevich: Sadly, there are multiple versions of golang available in Ubuntu for each release and this is how they're suppose to be installed system-wide.  Otherwise using them requires some (more fugly) env. var. monkeying.  Though the -sf isn't a bad idea, would you be okay with me adding the `f` but not re-rolling the VM images?  I can also add a comment to that line.

edsantiago: Oh heck yes. None of my comments (except the docker one, which I think is obsolete knowledge) require re-spinning CI:IMG. A comment would be appreciated.

contrib/cirrus/packer/fedora_packaging.sh
cevich: good suggestions.

FWIW: "something something" == other repos need runc (and other packages) so we download and cache them at build-time so they can be installed at runtime w/o version/update surprises :grin:

cevich:
> Is this for test/apiv2/rest_api?

Yes...@jwhonce needs to take a look.  Possibly play / confirm his stuff works in a `hack/get_ci_vm.sh fedora-31-podman-6530021898584064` (assuming if it works in 31, it'll work in 32)

contrib/cirrus/packer/ubuntu_packaging.sh
cevich: okay, done.

contrib/cirrus/packer/fedora_packaging.sh
cevich: okay, got it.

cevich:
Force-pushed just comment updates.  Would like the final /lgtm from @jwhonce to confirm these are the only python modules needed.

jwhonce:
@cevich No joy running the hack script:

```
.
:
ERROR: (gcloud.compute.ssh) Could not fetch resource:
 - The resource 'projects/libpod-218412/zones/us-central1-a/instances/jhonce-fedora-31-podman-6530021898584064' was not found
:
.
The following instances will be deleted. Any attached disks configured
 to be auto-deleted will be deleted unless they are attached to any 
other instances or the `--keep-disks` flag is given and specifies them
 for keeping. Deleting a disk is irreversible and any data on the disk
 will be lost.
 - [jhonce-fedora-31-podman-6530021898584064] in [us-central1-a]

Do you want to continue (Y/n)?  y

ERROR: (gcloud.compute.instances.delete) Failed to fetch some instances:
 - The resource 'projects/libpod-218412/zones/us-central1-a/instances/jhonce-fedora-31-podman-6530021898584064' was not found
```

Has `fedora-31-podman-6530021898584064` changed? How can I find the latest value?

cevich:
@jwhonce how are we lookin' here?

Issue #7140, opened 2020-07-29
podman-remote: images --format .History : unexpected cruft


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman-remote images alpine --format '{{ .History }}'
a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e, <missing>
4fab981df737963d83b548a3a6f4e7b43f718eb88567ba5ec7921f636a93d7bc, <missing>, <missing>, <missing>, <missing>, <missing>, <missing>, <missing>, <missing>
```
Compare to plain podman:
```console
$ ./bin/podman images alpine --format '{{ .History }}'
             <--- these are two empty lines

```

PR #7147, opened 2020-07-29
Return NamesHistory when returning remote images

rhatdan:
We are returning bogus data in podman-remote images --format json.
This change will match the same data returned my podman images --format json.


rhatdan:
Fixes: https://github.com/containers/podman/issues/7140

mheon:
Does this match what the Docker compat handler expects?

rhatdan:
I have no idea, But what we were returning was bogus. This matches our API

edsantiago:
Confirmed: the BATS 'image history' test now works against this build. Thanks.

mheon:
LGTM, then.

PR #7152, opened 2020-07-29
fix pod creation with "new:" syntax

### LINKED ISSUE #7087: rootless containers cannot connect using the local host on the same pod
/kind bug

**Description**
I'm running version 2.0.3 so maybe it's a new thing because the documentation says it should work.

Running on Fedora 32 with rootless containers.
I'm creating two containers in the same pod. They should connect one another using the local host, but it does not work.

1. podman run -d --pod=new:test docker.io/library/nginx:latest 

2. podman run --rm -it --pod=test fedora curl http://localhost

curl: (7) Failed to connect to localhost port 80: Connection refused

The default page from the server

```
podman version 2.0.3
```

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    path: /usr/bin/conmon
  cpus: 4
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: ip-10-0-0-111.ec2.internal
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.6.7-300.fc32.x86_64
  linkmode: dynamic
  memFree: 11909259264
  memTotal: 16780132352
  ociRuntime:
    name: crun
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      spec: 1.0.0
  os: linux
  remoteSocket:
  rootless: true
  slirp4netns:
    version: |-
      slirp4netns version 1.1.4
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 2
  swapFree: 0
  swapTotal: 0
  uptime: 5h 58m 39.45s (Approximately 0.21 days)
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - docker.io
store:
  configFile: /home/fedora/.config/containers/storage.conf
  containerStore:
    number: 7
    paused: 0
    running: 5
    stopped: 2
  graphDriverName: overlay
  graphOptions:
    overlay.mount_program:
      Version: |-
        fusermount3 version: 3.9.1
        fuse-overlayfs: version 1.1.0
        FUSE library version 3.9.1
  graphRoot: /home/fedora/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 7
  volumePath: /home/fedora/.local/share/containers/storage/volumes
version:
  APIVersion: 1
  Built: 0
  BuiltTime: Thu Jan  1 00:00:00 1970
  GitCommit: ""
  GoVersion: go1.14.3
  OsArch: linux/amd64
  Version: 2.0.3

```

```
podman-2.0.3-1.fc32.x86_64
```

Luap99:
When you execute podman create/run with the --pod new:\<name\> syntax
the pod was created but the namespaces where not shared and
therefore containers could not communicate over localhost.

Add the default namespaces and pass the network options to the
pod create options.

Closes #7087

### DIFF
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -124,7 +124,7 @@ func create(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if _, err := createPodIfNecessary(s); err != nil {
+	if _, err := createPodIfNecessary(s, cliVals.Net); err != nil {
 		return err
 	}
 
@@ -283,7 +283,7 @@ func openCidFile(cidfile string) (*os.File, error) {
 // createPodIfNecessary automatically creates a pod when requested.  if the pod name
 // has the form new:ID, the pod ID is created and the name in the spec generator is replaced
 // with ID.
-func createPodIfNecessary(s *specgen.SpecGenerator) (*entities.PodCreateReport, error) {
+func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions) (*entities.PodCreateReport, error) {
 	if !strings.HasPrefix(s.Pod, "new:") {
 		return nil, nil
 	}
@@ -292,11 +292,10 @@ func createPodIfNecessary(s *specgen.SpecGenerator) (*entities.PodCreateReport, 
 		return nil, errors.Errorf("new pod name must be at least one character")
 	}
 	createOptions := entities.PodCreateOptions{
-		Name:  podName,
-		Infra: true,
-		Net: &entities.NetOptions{
-			PublishPorts: s.PortMappings,
-		},
+		Name:          podName,
+		Infra:         true,
+		Net:           netOpts,
+		CreateCommand: os.Args,
 	}
 	s.Pod = podName
 	return registry.ContainerEngine().PodCreate(context.Background(), createOptions)
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -176,7 +176,7 @@ func run(cmd *cobra.Command, args []string) error {
 	}
 	runOpts.Spec = s
 
-	if _, err := createPodIfNecessary(s); err != nil {
+	if _, err := createPodIfNecessary(s, cliVals.Net); err != nil {
 		return err
 	}
 
--- a/pkg/specgen/generate/namespaces.go
+++ b/pkg/specgen/generate/namespaces.go
@@ -462,6 +462,10 @@ func specConfigureNamespaces(s *specgen.SpecGenerator, g *generate.Generator, rt
 func GetNamespaceOptions(ns []string) ([]libpod.PodCreateOption, error) {
 	var options []libpod.PodCreateOption
 	var erroredOptions []libpod.PodCreateOption
+	if ns == nil {
+		//set the default namespaces
+		ns = strings.Split(specgen.DefaultKernelNamespaces, ",")
+	}
 	for _, toShare := range ns {
 		switch toShare {
 		case "cgroup":
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -812,7 +812,11 @@ USER mail`
 	})
 
 	It("podman run --pod automatically", func() {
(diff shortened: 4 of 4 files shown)


rhatdan:
LGTM
@mheon @baude @haircommander @ashley-cui PTAL

cmd/podman/containers/create.go
```diff
@@ -292,11 +292,11 @@ func createPodIfNecessary(s *specgen.SpecGenerator) (*entities.PodCreateReport,
...
 	createOptions := entities.PodCreateOptions{
-		Name:  podName,
-		Infra: true,
-		Net: &entities.NetOptions{
-			PublishPorts: s.PortMappings,
-		},
+		Name:          podName,
+		Infra:         true,
+		Net:           netOpts,
+		Share:         strings.Split(specgen.DefaultKernelNamespaces, ","),
```
mheon, line 298: It looks like we need to make sure that the defaults are used in Specgen if this is unset.

Luap99: [there?](https://github.com/containers/podman/blob/c66ce8d67f9eb53a0997da39999ed8629d3252e3/pkg/specgen/generate/namespaces.go#L462)

mheon: Yes - that looks right. If the argument to that is nil, we should return the defaults. A 0-length non-nil array would indicate nothing is shared.

Luap99: OK, updated.

cmd/podman/containers/create.go
```diff
@@ -292,11 +292,10 @@ func createPodIfNecessary(s *specgen.SpecGenerator) (*entities.PodCreateReport,
...
 	createOptions := entities.PodCreateOptions{
-		Name:  podName,
-		Infra: true,
-		Net: &entities.NetOptions{
-			PublishPorts: s.PortMappings,
-		},
+		Name:          podName,
+		Infra:         true,
+		Net:           netOpts,
+		CreateCommand: os.Args,
```
mheon, line 298: Probably best to leave this empty - it's mainly used for `podman generate systemd --new`, and it could have misleading results there

mheon:
One more comment, LGTM once it's addressed

cmd/podman/containers/create.go
Luap99: I believe it is correct to set here. I checked `podman generate systemd --new` and it throws an error that you have to use `podman pod create`. This can later be used to implement this.

mheon:
Needs a rebase to fix CI

cmd/podman/containers/create.go
mheon: @vrothberg Thoughts here?

vrothberg: It looks good to me. As @Luap99 mentioned, `generate systemd` requires a pod to be generated via `pod create` and will error out otherwise (see https://github.com/containers/podman/blob/master/pkg/systemd/generate/pods.go#L265).

Setting it here will be more accurate as `generate systemd` will otherwise error out with "no create command found" which would be less informative than "pod does not appear to be created via `podman pod create`"

mheon: Alright. We should be good to merge after a rebase, then.

TomSweeneyRedHat:
LGTM
assuming happy tests

Luap99:
One test still failed! A Flake?

mheon:
Looking...

mheon:
I've never seen this one before - going to assume it's a flake and restart...

Luap99:
Now it timed out! :(

Luap99:
@mheon Test is green :)

Issue #7114, opened 2020-07-28
podman-remote: commit: not tagging


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman-remote run --name a alpine true
$ ./bin/podman-remote commit a b
f95ed969832cf49b012d33ffa3e4e020df892d3acde8fb9e6167c176ec4d25eb
$ ./bin/podman-remote  images
REPOSITORY                    TAG     IMAGE ID      CREATED             SIZE
<none>                        <none>  f95ed969832c  About a minute ago  5.85 MB    <---- missing tags
docker.io/library/alpine      latest  a24bb4013296  8 weeks ago         5.85 MB
```
Expected:
```
REPOSITORY                    TAG     IMAGE ID      CREATED        SIZE
localhost/b                   latest  feb57509edd1  4 seconds ago  5.85 MB
docker.io/library/alpine      latest  a24bb4013296  8 weeks ago    5.85 MB
```
master @ 288ebec6e, root & rootless

edsantiago:
It's a string length problem??!? If I commit as `bb` (two characters), it works fine:
```console
$ ./bin/podman-remote commit a bb
72c80484d8679099965be00d4f0e108bb878946a02aa917520120d501fdb1992
$ ./bin/podman-remote images
REPOSITORY                    TAG     IMAGE ID      CREATED        SIZE
localhost/bb                  latest  72c80484d867  4 seconds ago  5.85 MB
```

ashley-cui:
closed via #7158

PR #7158, opened 2020-07-30
Handle single character images

rhatdan:
Currently you can only specify multiple character for image names
when executing podman-remote commit

podman-remote commit a b
Will complete, but will save the image without a name.

podman-remote commit a bb
Works.

This PR fixes and now returns an error if the user doees not specify an
image name to commit to.

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -594,11 +594,9 @@ func CommitContainer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// I know mitr hates this ... but doing for now
-	if len(query.Repo) > 1 {
+	if len(query.Repo) > 0 {
 		destImage = fmt.Sprintf("%s:%s", query.Repo, tag)
 	}
-
 	commitImage, err := ctr.Commit(r.Context(), destImage, options)
 	if err != nil && !strings.Contains(err.Error(), "is not running") {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrapf(err, "CommitFailure"))
--- a/test/e2e/commit_test.go
+++ b/test/e2e/commit_test.go
@@ -49,6 +49,21 @@ var _ = Describe("Podman commit", func() {
 		Expect(StringInSlice("foobar.com/test1-image:latest", data[0].RepoTags)).To(BeTrue())
 	})
 
+	It("podman commit single letter container", func() {
+		_, ec, _ := podmanTest.RunLsContainer("test1")
+		Expect(ec).To(Equal(0))
+		Expect(podmanTest.NumberOfContainers()).To(Equal(1))
+
+		session := podmanTest.Podman([]string{"commit", "test1", "a"})
+		session.WaitWithDefaultTimeout()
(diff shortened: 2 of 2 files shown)


rhatdan:
fixes: https://github.com/containers/podman/issues/7114
@edsantiago PTAL

mheon:
Can you add a test?

giuseppe:
LGTM

pkg/api/handlers/libpod/images.go
```diff
@@ -594,11 +591,11 @@ func CommitContainer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// I know mitr hates this ... but doing for now
-	if len(query.Repo) > 1 {
-		destImage = fmt.Sprintf("%s:%s", query.Repo, tag)
+	if len(query.Repo) == 0 {
```
TomSweeneyRedHat, line 594: I'm always leery about " equals some number"  could you convert to `< 1`?

edsantiago:
@rhatdan CI failures look real, in API tests:
```
not ok 154 [20-containers] POST libpod/commit?container=myfoo [] : status
#  expected: 200
#    actual: 500
```
I have to guess that your change somehow broke a corner case in APIv2.

edsantiago:
Yep, looks like 'podman commit containername`, with no imagename, is perfectly valid:
```console
$ ./bin/podman commit --help
   podman commit [flags] CONTAINER [IMAGE]
./bin/podman run --name a alpine true                                                                                                         pr/7158
$ ./bin/podman commit a
Copying blob 50644c29ef5a skipped: already exists
Copying blob e624f929ee9f done
Copying config d8a029cf84 done
Storing signatures
d8a029cf8407bc6406697214a1bbdfe8729a845e99323f5bfd590e6ea73c974b
```
...but with this PR, that usage is no longer being allowed.

edsantiago:
@cevich we're seeing a lot of these, on many different PRs and over a long period of time (all last week):
```
Uploading 3 artifacts for /var/tmp/go/src/github.com/containers/podman/*.log.html
Uploaded /var/tmp/go/src/github.com/containers/podman/apiv2_test.log.html
Failed to upload artifact file /var/tmp/go/src/github.com/containers/podman/integration_test.log.html: EOF
```
Does it look familiar? Is there anything you can think of that can make this more robust?

cevich:
> Does it look familiar? Is there anything you can think of that can make this more robust?

Yes, I've heard of this, and from multiple containers repositories (i.e. it's not just a problem in one google-project or github project or cirrus-ci configuration).  Since yours represents "even more repeated complaints" now, I best bring it to the attention of CIrrus support.  That's not a guarantee that can/will do anything about it, but maybe they can help find where the problem comes from (so we can avoid it somehow).

edsantiago:
@cevich here are three just from this PR, all demonstrating the same three-line failure: 

https://cirrus-ci.com/task/5071485364600832
https://cirrus-ci.com/task/5339766201778176
https://cirrus-ci.com/task/5621241178488832

Here's one with a zero-line failure:

https://cirrus-ci.com/task/4508535411179520

cevich:
@edsantiago thanks, I'll get those right over to them.

PR #7162, opened 2020-07-30
rootless: do not ignore errors if mappings are specified

giuseppe:
when setting up the user namespace do not ignore errors from newuidmap/newgidmap if there are mappings configured.

The single user mapping is a fallback only when there are not mappings specified for the user.

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -224,6 +224,10 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (bool,
 	uidsMapped := false
 	if uids != nil {
 		err := tryMappingTool("newuidmap", pid, os.Geteuid(), uids)
+		// If some mappings were specified, do not ignore the error
+		if err != nil && len(uids) > 0 {
+			return false, -1, err
+		}
 		uidsMapped = err == nil
 	}
 	if !uidsMapped {
@@ -246,6 +250,10 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (bool,
 	gidsMapped := false
 	if gids != nil {
 		err := tryMappingTool("newgidmap", pid, os.Getegid(), gids)
+		// If some mappings were specified, do not ignore the error
+		if err != nil && len(gids) > 0 {
+			return false, -1, err
+		}
 		gidsMapped = err == nil
 	}
 	if !gidsMapped {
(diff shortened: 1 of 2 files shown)


pkg/rootless/rootless_linux.go
```diff
@@ -110,6 +110,9 @@ func tryMappingTool(tool string, pid int, hostID int, mappings []idtools.IDMap)
 	args := []string{path, fmt.Sprintf("%d", pid)}
 	args = appendTriplet(args, 0, hostID, 1)
 	for _, i := range mappings {
+		if hostID >= i.HostID && hostID < i.HostID+i.Size {
+			return errors.Errorf("invalid configuration: the specified mapping %d:%d includes the host ID", i.HostID, i.Size)
```
edsantiago, line 114: Does "host ID" mean UID? If so, might it be clearer to say `...includes host UID (%d)`?

giuseppe: the same code is used both for UID and GID, that is why I used the generic ID :-)

I can pass down the information whether it is a UID or a GID and give a clearer error if you'd like

giuseppe: added and pushed a new version

edsantiago: Ah, thanks, that makes sense. Yes, please, I think users would appreciate a clearer message. Even after all this time, I still don't understand subuid/subgid, and I really would have no idea what a message like that would mean. Something like "the specified sub%cid mapping ... overlaps %cID %d" would make it unambiguous and easy to track down. Thank you!

pkg/rootless/rootless_linux.go
```diff
@@ -110,6 +114,13 @@ func tryMappingTool(tool string, pid int, hostID int, mappings []idtools.IDMap)
 	args := []string{path, fmt.Sprintf("%d", pid)}
 	args = appendTriplet(args, 0, hostID, 1)
 	for _, i := range mappings {
+		if hostID >= i.HostID && hostID < i.HostID+i.Size {
+			what := "UID"
+			if !uid {
+				what = "GID"
+			}
+			return errors.Errorf("invalid configuration: the specified mapping %d:%d includes the host %s", i.HostID, i.Size, what)
```
edsantiago, line 122: As an ignorant 1980s-era UNIX user, I would still find this message confusing and unhelpful. Please remember that subuid/subgid is a newfangled fancy thing that old people do not know about, and especially please remember that podman may be running those tools secretly in the background. I have no idea what a "mapping" is, because I the user played no part in setting it up. Could you at least include something like `/etc/sub%sid` or even just `%sid`, with a `ToLower(what)` or equivalent? It's not a huge deal, few people will ever run into this, but those who do will probably appreciate our respect for their time.

giuseppe: fixed

mheon:
LGTM!

Issue #7130, opened 2020-07-29
Wrong file ownership on volumes initialized from container image


praiskup:
/kind bug

**Description**

The bind-mounted, freshly initialized volume contains correct files from the original
container image - but wrong permissions, when some directory has `rwx------`
permissions.

**Steps to reproduce the issue:**

```
$ podman build .
STEP 1: FROM fedora:32
STEP 2: RUN useradd testuser &&     mkdir /var/cache/testuser &&     touch /var/cache/testuser/datafile &&     chown -R testuser:testuser /var/cache/testuser &&     chmod -R 0700 /var/cache/testuser
STEP 3: COMMIT
--> 498a900473b
498a900473b241651fe9e2e7e078efaca9b59d7060a6a4b5724a97a113cbf10c
$ podman run --rm -ti 498a900473b241651fe9e2e7e078efaca9b59d7060a6a4b5724a97a113cbf10c bash
[root@c1dcda852158 /]# ls /var/cache/testuser/ -lah
total 0
drwx------. 2 testuser testuser 22 Jul 29 07:46 .
drwxr-xr-x. 3 root     root     22 Jul 29 07:46 ..
-rwx------. 1 testuser testuser  0 Jul 29 07:46 datafile
[root@c1dcda852158 /]# exit
$ podman run -v newvolume:/var/cache/testuser --rm -ti 498a900473b241651fe9e2e7e078efaca9b59d7060a6a4b5724a97a113cbf10c bash
[root@0c218c6e4777 /]# ls /var/cache/testuser/ -lah
total 0
drwx------. 2 root root 22 Jul 29 07:47 .
drwxr-xr-x. 3 root root 22 Jul 29 07:46 ..
-rwx------. 1 root root  0 Jul 29 07:46 datafile
[root@0c218c6e4777 /]#
```

**Describe the results you received:**

You can see in that output that file ownership differs in the two `podman run` scenarios,
one run with `-v` and one without.

**Output of `podman version`:**

```
$ podman version
Version:      2.0.3
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
$ podman info --debug
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
[...]

PR #7164, opened 2020-07-30
volumes: do not recurse when chowning

giuseppe:
keep the file ownership when chowning and honor the user namespace
mappings.

Closes: https://github.com/containers/podman/issues/7130


edsantiago:
This breaks one of the BATS tests, but it's almost certainly because I had no idea what I was doing when I wrote the test. If this is truly the new desired behavior, please edit `test/system/070-build.bats` starting at line 218.

edsantiago:
Oh, and if you're still around, there's no point in waiting for CI. It is going to hard-fail on this.

mheon:
LGTM once tests are adjusted

PR #7168, opened 2020-07-31
Fix close fds of exec --preserve-fds

QiWang19:
Add same fix as #7120
Fix the closing of fds from --preserve-fds to avoid the operation on unrelated fds.


mheon:
LGTM

PR #7176, opened 2020-07-31
Ensure WORKDIR from images is created

mheon:
A recent crun change stopped the creation of the container's working directory if it does not exist. This is arguably correct for user-specified directories, to protect against typos; it is definitely not correct for image WORKDIR, where the image author definitely intended for the directory to be used.

This makes Podman create the working directory and chown it to container root, if it does not already exist, and only if it was specified by an image, not the user.


mheon:
@giuseppe PTAL

mheon:
(This PR also includes a large swath of comment updates to fields in Container, because the lack of descriptiveness annoyed me)

pkg/specgen/generate/container_create.go
```diff
@@ -238,6 +238,17 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
...
+	// If the user did not set an entrypoint but the image did, ensure it is
+	// created.
+	if s.Entrypoint == nil && img != nil {
+		newEntry, err := img.Entrypoint(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if len(newEntry) > 0 {
+			options = append(options, libpod.WithCreateWorkingDir())
+		}
```
giuseppe, line 250: shouldn't we look at `img.WorkingDir(...)`?  Why does it depend on the entrypoint?

rhatdan:
Does the owner of the WORKDIR depend on the user of the image or the root of the image?
```
cat Containerfile
from fedora
run useradd dwalsh
user dwalsh
workdir /home/dwalsh
```
Is work dir supposed to be owned by dwalsh, or by root?

mheon:
In Docker, it's always owned by root.

pkg/specgen/generate/container_create.go
mheon: Oops, fixed

TomSweeneyRedHat:
LGTM
assuming happy tests

mheon:
Image in the E2E test from a `podman build` isn't being picked up by subsequent `podman run` - trying to debug now.

mheon:
The problem appears to be with remote - all the tests that build images are SkipIfRemote. I made this one SkipIfRemote as well.

giuseppe:
LGTM

vrothberg:
We merged a rather exotic commit message (going through commits atm)

mheon:
Yeah, that one was my fault... I made a debug commit to figure out why the tests were failing, and when I fixed it I did a `git commit -a --amend --no-edit` to bring the changes it, completely forgetting that I have a debug commit on top...

The good news is that the actual things that should not be merged were wiped out by that `git commit --amend`.

PR #7177, opened 2020-07-31
Ensure libpod/define does not include libpod/image

mheon:
The define package under Libpod is intended to be an extremely minimal package, including constants and very little else. However, as a result of some legacy code, it was dragging in all of libpod/image (and, less significantly, the util package). Fortunately, this was just to ensure that error constants were not duplicating, and there's nothing preventing us from importing in the other direction and keeping libpod/define free of dependencies.

### DIFF
--- a/libpod/define/errors.go
+++ b/libpod/define/errors.go
@@ -2,23 +2,20 @@ package define
 
 import (
 	"errors"
-
-	"github.com/containers/podman/v2/libpod/image"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
Much better.

TomSweeneyRedHat:
LGTM
I love little changes like this with big impacts, well done @mheon

PR #7181, opened 2020-08-01
[CI:DOCS] Change recommended systemd unit path for root.

Luap99:
`/usr/lib/systemd/system` should only be used by the package manager
administrators should use: `/etc/systemd/system` or `/usr/local/lib/systemd/system`

see: man [systemd.unit](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Unit%20File%20Load%20Path)


TomSweeneyRedHat:
LGTM
but would like a head nod from @lsm5 and/or @vrothberg

rhatdan:
Nice catch @Luap99 Thanks.
/lgtm
/approve

lsm5:
i'm late but LGTM.

Issue #7184, opened 2020-08-02
Podman API returns 500 in case network is not found instead of 404


orenbm:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Running `docker-compose build <service-name>` doesn't create the network properly. This happens because `docker-compose build` (or any other `docker-compose` command that creates an image) will start by creating a network. It will run `GET /v1.24/networks/test_default` and will get a 500 response. However, Docker returns 404 in this case (which makes sense as the network is not found).

**Steps to reproduce the issue:**

1. Start podman service with `podman system service -t 0 tcp:0.0.0.0:8080`

2. Run `curl` command to get a non-existing network:
```
curl http://localhost:8080/networks/non-existing-network
```

**Describe the results you received:**
```
{"cause":"network not found","message":"unable to find network configuration for non-existing-network: network not found","response":500}
```

**Describe the results you expected:**
```
{"cause":"network not found","message":"unable to find network configuration for non-existing-network: network not found","response":400}
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 2.0.4
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.20-1.el8.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.20, commit: ee5acee39338dff8d32a921a46909359315197b6'
  cpus: 2
  distribution:
    distribution: '"rhel"'
    version: "8.2"
  eventLogger: file
  hostname: ip-172-31-13-47.eu-central-1.compute.internal
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.18.0-193.el8.x86_64
  linkmode: dynamic
  memFree: 303247360
  memTotal: 3959033856
  ociRuntime:
    name: runc
    package: runc-1.0.0-145.rc91.git24a3cf8.el8.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.2-dev'
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 0
  swapTotal: 0
  uptime: 89h 52m 11.99s (Approximately 3.71 days)
registries:
  search:
  - registry.fedoraproject.org
[...]

PR #7188, opened 2020-08-02
API returns 500 in case network is not found instead of 404

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/7184

### DIFF
--- a/pkg/api/handlers/compat/networks.go
+++ b/pkg/api/handlers/compat/networks.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containers/podman/v2/libpod"
+	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/api/handlers/utils"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/domain/infra/abi"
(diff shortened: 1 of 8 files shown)


pkg/api/handlers/compat/networks.go
```diff
@@ -2,6 +2,7 @@ package compat
 
 import (
 	"encoding/json"
+	"github.com/containers/podman/v2/libpod/define"
```
orenbm, line 5: looks like this should go in the batch that starts on line 13

pkg/api/handlers/libpod/networks.go
```diff
@@ -2,13 +2,13 @@ package libpod
 
 import (
 	"encoding/json"
+	"github.com/containers/podman/v2/libpod/define"
```
orenbm, line 5: looks like this should go in the batch that starts on line 8

pkg/network/files.go
```diff
@@ -3,6 +3,7 @@ package network
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/containers/podman/v2/libpod/define"
```
orenbm, line 6: looks like this should go in the batch that starts on line 13

pkg/network/network.go
```diff
@@ -2,6 +2,7 @@ package network
 
 import (
 	"encoding/json"
+	"github.com/containers/podman/v2/libpod/define"
```
orenbm, line 5: looks like this should go in the batch that starts on line 11

test/apiv2/35-networks.at
```diff
@@ -0,0 +1,8 @@
+# -*- sh -*-
+#
+# network-related tests
+#
+
+t GET /networks/non-existing-network 404
+
+# vim: filetype=sh
```
orenbm, line 8: missing newline at end of file

zhangguanzhang:
@orenbm thanks

zhangguanzhang:
@vrothberg PTAL

pkg/network/config.go
```diff
@@ -20,10 +19,6 @@ const (
 	DefaultPodmanDomainName = "dns.podman"
 )
 
-var (
-	ErrNetworkNotFound = errors.New("network not found")
```
rhatdan, line 24: This should have been left and just a redirect to define.ErrNetowrkNotFound, otherwise you are changing the API.

PR #7141, opened 2020-07-29
Add test case for description being present in search result

rhafer:
Test for a specific static image and match the description to avoid
regression like https://github.com/containers/podman/pull/7131

Also apply the fix above to the api implementation.

Signed-off-by: Ralf Haferkamp <rhafer@suse.com>

PR #7191, opened 2020-08-03
Remove some unnecessary []byte to string conversions

saschagrunert:
Some calls to `Sprintf("%s")` can be avoided by using direct string
type assertions.

Indirect follow-up of https://github.com/containers/podman/pull/7141


mheon:
LGTM
/hold cancel

Issue #7180, opened 2020-08-01
user podman service leaks "pause" process, breaks API after idle timeout


martinpitt:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Stopping user's podman.service, or letting it time out, often leaves behind a "podman pause" process.

**Steps to reproduce the issue:**

1. Log in as user, and clean slate. There are no running containers, and service is not running:

        systemctl --user stop podman.service; systemctl --user start podman.socket
        # make sure there are no podman processes
        ps ux|grep podman

2. Make an API request:

        $ curl --silent --show-error --unix-socket $XDG_RUNTIME_DIR/podman/podman.sock http://d/libpod/_ping
        OK

3. Check processes. podman.service is running:
```
$ systemctl --user status podman
● podman.service - Podman API Service
     Loaded: loaded (/usr/lib/systemd/user/podman.service; disabled; vendor preset: disabled)
     Active: active (running) since Sat 2020-08-01 03:11:59 EDT; 1s ago
TriggeredBy: ● podman.socket
       Docs: man:podman-api(1)
   Main PID: 59799 (podman)
      Tasks: 14 (limit: 1286)
     Memory: 64.0M
        CPU: 145ms
     CGroup: /user.slice/user-1000.slice/user@1000.service/podman.service
             ├─59799 /usr/bin/podman system service
             ├─59805 /usr/bin/podman system service
             └─59809 /usr/bin/podman
```

   There is also a zombie process (but that's not the main issue of this bug report):
```
$ ps ux | grep podman
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
admin      59799  1.2  4.4 1206952 50508 ?       Ssl  03:11   0:00 /usr/bin/podman system service
admin      59805  1.2  4.6 1060976 51756 ?       Sl   03:11   0:00 /usr/bin/podman system service
admin      59808  0.0  0.0      0     0 ?        Zs   03:11   0:00 [podman] <defunct>
admin      59809  0.2  2.4  54832 28048 ?        S    03:11   0:00 /usr/bin/podman
```
4. Wait  a few seconds until podman.service times out, check with `systemctl --user status podman`

**Describe the results you received:**

After that I expect no running podman processes any more.

**Describe the results you expected:**

```
$ ps ux|grep podman
admin      59809  0.0  2.4  54832 28048 ?        S    03:11   0:00 /usr/bin/podman

[...]

Issue #6660, opened 2020-06-18
APIv2: Query after killing rootless hangs forever


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
In Cockpit we test how the UI behaves when podman dies,when only one is running (root/rootless)... To do that, we stop/kill sockets, services and processes. There seems to be a problem with rootless where it is not possible to connect once the process was terminated.

**Steps to reproduce the issue:**
1. `./script.sh` [script.sh](https://github.com/containers/libpod/files/4797131/script.sh.txt)
Note 1: It may succeed the first time (especially if freshly booted system), then you might need to re-run it
Note 2: Drop `--user` from all calls, change the unix path in the last call and run it as root, it all works as expected

**Describe the results you received:**
The script hangs forever. If you ^C it and then call `curl --unix-socket /run/user/1000/podman/podman.sock http://d/v1.24/libpod/info` bunch of times, eventually it succeeds. (other calls don't hang, they just `Couldn't connect to server`.

**Describe the results you expected:**
1. The call would not hang.
2. The call would always succeed on first, second or any other try (as does root one)

**Output of `podman version`:**

```
Version:      2.0.0-rc6
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0-dev
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.17-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.17, commit: bb8e273f5925c1a51737644637ef65d094a67ab1'
  cpus: 4
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: zahlmaschine
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 1935008
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 1935008
      size: 65536
  kernel: 5.6.11-300.fc32.x86_64
  linkmode: dynamic
  memFree: 391274496
  memTotal: 16418353152
  ociRuntime:
    name: crun
    package: crun-0.13-2.fc32.x86_64
    path: /usr/bin/crun
    version: |-
[...]

PR #7192, opened 2020-08-03
rootless: system service joins immediately the namespaces

giuseppe:
when there is a pause process running, let the "system service" podman
instance join immediately the existing namespaces.

Closes: https://github.com/containers/podman/issues/7180
Closes: https://github.com/containers/podman/issues/6660

### DIFF
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -205,7 +205,7 @@ can_use_shortcut ()
 
       if (strcmp (argv[argc], "mount") == 0
           || strcmp (argv[argc], "search") == 0
-          || strcmp (argv[argc], "system") == 0)
+          || (strcmp (argv[argc], "system") == 0 && argv[argc+1] && strcmp (argv[argc+1], "service") != 0))
         {
           ret = false;
           break;


martinpitt:
@giuseppe: I'm happy to test this -- do the CI processes spit out RPMs or debs somewhere which I can use?

giuseppe:
@vrothberg do you know if we generate a RPM automatically?

@martinpitt if it would work for you, I can provide a static binary

martinpitt:
@giuseppe : Yes, an x86_64 binary is just fine. I do all tests in temporary VMs anyway, so I can mess them up however much I like :)

giuseppe:
@martinpitt uploaded here: [podman.gz](https://github.com/containers/podman/files/5014660/podman.gz)

martinpitt:
I confirm that with this binary, the `podman pause` process still persists, the zombie still exists, but the API now continues to work after podman.service timed out. It gets restarted correctly. Thank you!

:+1:

giuseppe:
@vrothberg can I get the lgtm back? :-)

vrothberg:
:)

Issue #7190, opened 2020-08-03
Drop [Install] section from podman.service


martinpitt:
/kind bug

/kind feature

**Description**

`podman.service` is socket activated through `podman.socket`. It should not have its own `[Install]` section, it does not make sense to `systemctl enable podman.service`. This is defined [here](https://github.com/containers/podman/blob/master/contrib/systemd/system/podman.service#L12).

This leads to podman.service always running on a Debian system, as Debian's policy is to enable/start running services by default.

Can you please drop this?

**Output of `podman version`:**

    podman version 2.0.4

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman/unstable,now 2.0.4+dfsg1-1 amd64 [installed]
```

vrothberg:
Thanks for the report. I'll open a PR in a minute.

PR #7193, opened 2020-08-03
podman.service: drop install section

vrothberg:
podman.service is socket activated through podman.socket. It should not
have its own [Install] section, it does not make sense to systemctl
enable podman.service.

This leads to podman.service always running on a Debian system, as
Debian's policy is to enable/start running services by default.

We don't want a daemon :^)

Fixes: #7190
Reported-by: @martinpitt

### DIFF
--- a/contrib/systemd/system/podman.service
+++ b/contrib/systemd/system/podman.service
@@ -8,7 +8,3 @@ StartLimitIntervalSec=0
 [Service]
 Type=simple
 ExecStart=/usr/bin/podman system service
-
-[Install]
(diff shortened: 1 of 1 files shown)


vrothberg:
@ashley-cui @baude @giuseppe @jwhonce @mheon @rhatdan @TomSweeneyRedHat PTAL

martinpitt:
Thanks @vrothberg , much appreciated!

TomSweeneyRedHat:
LGTM

PR #7199, opened 2020-08-03
Restore "table" --format from V1

jwhonce:
* --format "table {{.field..." will print fields out in a table with
  headings.  Table keyword is removed, spaces between fields are
  converted to tabs.
* Update parse.MatchesJSONFormat()'s regex to be more inclusive
* Add report.Headers(), obtain all the field names to be used as
  column names, a map of field name to column name may be provided
  to override the field names.

Note: This PR provides the functions to support this feature.  Once merged, follow-on PR's will implement the necessary changes in the podman commands.

### DIFF
--- a/cmd/podman/report/format.go
+++ b/cmd/podman/report/format.go
@@ -1,0 +1,68 @@
+package report
+
+import (
+	"reflect"
+	"strings"
+)
+
+// tableReplacer will remove 'table ' prefix and clean up tabs
+var tableReplacer = strings.NewReplacer(
+	"table ", "",
+	`\t`, "\t",
+	`\n`, "\n",
+	" ", "\t",
+)
+
+// escapedReplacer will clean up escaped characters from CLI
+var escapedReplacer = strings.NewReplacer(
+	`\t`, "\t",
+	`\n`, "\n",
+)
+
+// NormalizeFormat reads given go template format provided by CLI and munges it into what we need
+func NormalizeFormat(format string) string {
+	f := format
+	// two replacers used so we only remove the prefix keyword `table`
+	if strings.HasPrefix(f, "table ") {
+		f = tableReplacer.Replace(f)
+	} else {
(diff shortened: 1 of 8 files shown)


rhatdan:
@jwhonce Do you have time to get back to this one now?

cmd/podman/report/format.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"reflect"
+	"strings"
+)
+
+// TableReplacer will remove 'table ' prefix and clean up tabs
+var TableReplacer = strings.NewReplacer(
+	"table ", "",
+	`\t`, "\t",
+	" ", "\t",
```
ashley-cui, line 12:
```suggestion
	"\\t ", "\t",
```
Docker formatting will take in a \t in the command line and change it into a tab. I think double backslash escapes the backslash escape, so that this will work?

```
$ sudo docker ps --format "table {{.Names}}\t{{.Status}}\t{{.RunningFor}}\t{{.Image}}\t{{.Ports}}"
NAMES                 STATUS                            CREATED             IMAGE               PORTS
condescending_gould   Up About a minute                 2 minutes ago       nginx:alpine        80/tcp
```

baude:
comments from @ashley-cui but other lgtm

cmd/podman/report/format.go
jwhonce: @ashley-cui Back quotes in go are raw string literals so \`\t\` will cover your use case. The double quotes are interpreted string literals so `"\t"` is replaced by the tab character.

Unless you are asking for three runes `\ + t + <sp>` to be replaced by tab, that is in your comment and the current replacer will not make that change but I can add it if you need it.

ashley-cui: ah okay! didn't see the backquotes, LGTM

rhatdan:
@jwhonce Testing is not happy.

ashley-cui:
LGTM

Issue #7022, opened 2020-07-20
APIv2: No events on image build


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
`podman build` does not produce any event in the REST API.

**Steps to reproduce the issue:**

1. `sudo curl --unix-socket /run/podman/podman.sock http://d/v1.24/events` - start listening on events

2. ` sudo podman build .` (some simple Dockerfile)

3. `sudo podman images -a` shows a new image, but there is no events in the step 1. (`sudo podman pull` produces event)

**Describe the results you received:**

**Describe the results you expected:**
Some events that the image list has changed.

**Output of `podman version`:**

```
Version:      2.0.1
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

vrothberg:
Thanks for opening an issue, @marusak. I can reproduce: both `podman build` and `podman-remote build` do not yield an event.

vrothberg:
I found some time to have a closer look. It seems there is no `build` event at all. @mheon did we have some in previous versions?

mheon:
I don't see how we would, given that the events code lives in Libpod and
image build is entirely in Buildah land

vrothberg:
> I don't see how we would, given that the events code lives in Libpod and image build is entirely in Buildah land

I think it's pretty similar to pulling, which happens in c/image. We could track down all buildah API calls in the code and record events there. Wouldn't that work?

mheon:
The problem specifically with build is that we immediately hand off from our CLI to the Buildah CLI - the other calls are usually a lot more integrated with the rest of our code (`commit` on images, for example - uses Buildah code, but just library calls, not handing off the entire responsibility, so we can embed the events calls).

We could potentially generate an event before we ever call Buildah, but I don't think we know some important things at that point (ID of the final image being the big one)

rhatdan:
Let's integrate something intoBuildah, that it can call functions on events.  Would seem to be the best way.

mheon:
We should move the events code into c/common to allow Buildah to also call into it.

mheon:
We will have issues if they want to do container events, though, because we can't see those in `podman ps` or the list API, so Cockpit will get very confused.

vrothberg:
> We will have issues if they want to do container events, though, because we can't see those in `podman ps` or the list API, so Cockpit will get very confused.

[...]

PR #7202, opened 2020-08-03
add event for image build

baude:
upon image build completion, a new image type event is written for "build". more intricate details, like pulling an image, that might be done by build must be implemented in different vendored packages only after libpod is split from podman.

Fixes: #7022


libpod/image/image.go
```diff
@@ -1570,6 +1570,16 @@ func (i *Image) GetHealthCheck(ctx context.Context) (*manifest.Schema2HealthConf
 	return configBlob.ContainerConfig.Healthcheck, nil
 }
 
+// newBuildEvent creates a new event based on completion of a built image
+func (ir *Runtime) NewImageBuildCompleteEvent(idOrName string) {
```
mheon, line 1574: Should we just put this in `libpod/runtime_img.go` as a method on Runtime instead of ImageRuntime? It's less consistent, but that way it stays in the same file as the only use of it...

baude: I waffled on that too.  In the end, I figured it should "sit' with the other methods.  I could go either way depending on how insistent we are.

mheon: My basic thought it that we should make it private so nobody accidentally stumbles on the API and makes spurious image events

rhatdan: Let's make it private, we can always make it public in the future if necessary without breaking API.

baude:
ready for merge when tests pass

rhatdan:
LGTM

rhatdan:
@vrothberg @giuseppe @mheon @TomSweeneyRedHat @jwhonce @ashley-cui @QiWang19 PTAL

PR #7205, opened 2020-08-03
docker-compose uses application/tar

baude:
even though the official documentation suggests that application/x-tar should be used for tar files, it seems docker-compose uses application/tar.  we now accept them and issue a warning.

Fixes: #7185

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containers/podman/v2/pkg/api/handlers/utils"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/gorilla/schema"
+	"github.com/sirupsen/logrus"
 )
 
 func BuildImage(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

pkg/api/handlers/compat/images_build.go
```diff
@@ -33,7 +34,13 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if hdr, found := r.Header["Content-Type"]; found && len(hdr) > 0 {
-		if hdr[0] != "application/x-tar" {
+		contentType := hdr[0]
+		switch contentType {
+		case "application/tar":
+			logrus.Warnf("tar file content type is  %s, should use \"application/x-tar\" content type", contentType)
```
orenbm, line 40: [`application/x-gtar` is also an alias of `application/x-tar`](http://help.dottoro.com/lapuadlp.php). Do we want to add the same behaviour for that header as well? let it work and write a warning.

baude: i wouldnt reject a pr for it.  but otherwise, how about we wait until we are asked for it ?

orenbm: I see no harm in implementing this in another PR but i can also live with not implementing it for now. `application/tar` was important for supporting docker-compose, I am not aware of a tool that uses `application/x-gtar`.

PR #7207, opened 2020-08-03
Improve error message when creating a pod/ctr with the same name

Luap99:
Check if there is an pod or container an return
the appropriate error message instead of blindly
return 'container exists' with `podman create` and
'pod exists' with `podman pod create`.

Example:
```
$ podman pod create --name test
f0ebce3c66d4634e92b64bcd42c006b3538f6e519f640b84efdff6e1b692b397
$ podman create --name test alpine
Error: name test is in use: container already exists
$ ./bin/podman create --name test alpine
Error: name "test" is in use: pod already exists
```

### DIFF
--- a/libpod/boltdb_state.go
+++ b/libpod/boltdb_state.go
@@ -2347,11 +2347,19 @@ func (s *BoltState) AddPod(pod *Pod) error {
 		// Check if we already have something with the given ID and name
 		idExist := idsBkt.Get(podID)
 		if idExist != nil {
-			return errors.Wrapf(define.ErrPodExists, "ID %s is in use", pod.ID())
+			err = define.ErrPodExists
+			if allPodsBkt.Get(idExist) == nil {
+				err = define.ErrCtrExists
+			}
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
Thanks @Luap99

rhatdan:
LGTM

libpod/boltdb_state.go
```diff
@@ -2347,11 +2347,19 @@ func (s *BoltState) AddPod(pod *Pod) error {
 		// Check if we already have something with the given ID and name
 		idExist := idsBkt.Get(podID)
 		if idExist != nil {
-			return errors.Wrapf(define.ErrPodExists, "ID %s is in use", pod.ID())
+			err = define.ErrPodExists
+			if allPodsBkt.Get(idExist) == nil {
+				err = define.ErrCtrExists
+			}
+			return errors.Wrapf(err, "ID \"%s\" is in use", pod.ID())
```
mheon, line 2354: For reference, `%q` produces the same output as `\"%s\"` - but since they're identical, there's no real need to change it.

PR #7208, opened 2020-08-03
Accept multiple Content-Types for building an image

orenbm:
Signed-off-by: orenbm oren.benmeir@cyberark.com
Fixes: #7185

Currently we accept only the 'application/x-tar' header when sending
the tar file for building an image. However, [this header has the aliases
'application/tar' and 'application/x-gtar'](http://help.dottoro.com/lapuadlp.php) 
so we should accept those as well.

Furthermore, [docker-compose uses 'application/tar'](https://docs.docker.com/engine/api/v1.25/#operation/ImageBuild) so accepting this content-type
will take us one step further in supporting docker-compose with the Podman API.


mheon:
I believe @baude has a fix out already - #7205

orenbm:
thanks @mheon , didn't see that. I'll leave a comment there and close this one.

Issue #6804, opened 2020-06-27
Feature: run/exec `--docker-compatible-check`


bbros-dev:
/kind feature

**Description**
As a user of 3rd-party tools that use Docker in scripts
I want to easily see the changes need to be made to replace `docker` with `podman`
I want to check docker compatibility without downloading or building images
I want a report that solves incompatibilities
I want a report that shows potential issues
I want a report that shows solutions for potential issues

~~**Steps to reproduce the issue:**~~
**Describe the results you received:**

Currently no such helper exists.

**Describe the results you expected:**

Ideally the report would show for known incompatibilities:
A. The `podman` option/argument that replaces a particular `docker` option/argument
Ideally the report would show for potential issues:
A. A description of the possible issue
B. A description of the solution

Example workflow
1. Given a docker command:
    ```bash
    /usr/bin/docker run -i -t --rm \
      --user 1000:1000
      --volume /dev:/dev \
      --volume /usr/local/src/firecracker:/firecracker:z \
      --env OPT_LOCAL_IMAGES_PATH=/firecracker/build \
      --env PYTHONDONTWRITEBYTECODE=1 \
      fcuvm/dev:v16 bash
    ```

2. When I change to this `podman` command
    ```bash
    /usr/bin/podman run --docker-compatibility-check \
      -i -t --rm \
      --user 1000:1000
      --volume /dev:/dev \
      --volume /usr/local/src/firecracker:/firecracker:z \
      --env OPT_LOCAL_IMAGES_PATH=/firecracker/build \
      --env PYTHONDONTWRITEBYTECODE=1 \
      fcuvm/dev:v16 bash
    ```

3. I receive some such report:

```bash
Incompatibilities: None
Potential Issues: 2

Potential Issue 1:
    The use of `--volume dev:dev` can lead to failure when run rootless and/or privilege-free.
    The failure can be an error report about accessing /dev/console.
Solution: 
    See podman.io/solution/device-tree-sharing

Potential Issue 2:
    The use of `--user NNN:MMM` can lead to failure when run rootless and/or privilege-free.
    The failure can be an error about operating system denied access to a folder or file.
Solution: 
    Add the podman specific option `--userns=keep-id`. 
    For more details see https://podman.io/solution/user-name-space
```

**Additional information you deem important (e.g. issue happens only occasionally):**

This issue arises immediately for any user switching from Docker to Podman.

**Output of `podman version`:**

```
$ podman version
Version:      2.0.0
API Version:  1
[...]

PR #7209, opened 2020-08-03
podman: support --mount type=devpts

giuseppe:
Allow to create a devpts mount.

This is useful for containers that bind mount /dev/ from the host but
at the same time want to create a terminal.

It can be used as:

podman run -v /dev:/dev --mount type=devpts,target=/dev/pts ...

Closes: https://github.com/containers/podman/issues/6804


rhatdan:
Should this happen automatically?  Should this be the default, if you volume mount in /dev:/dev?

Issue #7211, opened 2020-08-03
Podman --remote argument ordering can instead run local Podman


mheon:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Argument ordering with `podman --remote` can cause Podman to not run remote, and instad execute locally.

**Steps to reproduce the issue:**

1. `podman run -t -i --rm --name testctr fedora bash`

2. `podman system service --log-level=debug`

3. `podman exec --log-level=debug --remote exec -t -i testctr bash`

**Describe the results you received:**

Note that `podman exec` debug logs indicate that it started Conmon, and there are no debug logs out of `podman system service`

**Describe the results you expected:**

Debug logs for `podman system service` indicate that `podman exec` executed remotely.

**Additional information you deem important (e.g. issue happens only occasionally):**

N/A

**Output of `podman version`:**

Master

jwhonce:
@mheon From previous discussions `--remote` must be given before subcommand. Otherwise, it gets eaten before the subcommand can see it. https://github.com/containers/podman/pull/7212 does allow `--remote` to be given anywhere between `podman` and subcommand.

mheon:
@jwhonce Oops, I mistyped the example - omit the first `exec`

It should be `podman --log-level=debug --remote ...`

PR #7212, opened 2020-08-03
Refactor parsing to not require --remote to be first flag

jwhonce:
Use cobra.Command.FParseErrWhitelist to no longer require --remote to be
the first argument in flags when using CLI

* tweaked test utils to prevent panics

Fixes: #7211

### DIFF
--- a/cmd/podman/registry/remote.go
+++ b/cmd/podman/registry/remote.go
@@ -5,22 +5,24 @@ import (
 	"sync"
 
 	"github.com/containers/podman/v2/pkg/domain/entities"
-	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

edsantiago:
LGTM.

@vrothberg I agree with the need for a test, but this is hard to test because it requires muckery (detecting that the podman command is `podman-remote`, then strip off the `-remote` (but preserve any other arguments present), then run that in various combinations. That's much easier in bash than go, so I have a system test in the works for it. My test fails (as expected) on master, passes when run against this code. I will submit it as part of a system-test update, probably early Monday.

jwhonce:
@edsantiago Thanks!

/hold cancel

PR #7215, opened 2020-08-04
images: speed up lists

vrothberg:
Listing images has shown increasing performance penalties with an
increasing number of images.  Unless `--all` is specified, Podman
will filter intermediate images.  Determining intermediate images
has been done by finding (and comparing!) parent images which is
expensive.  We had to query the storage many times which turned it
into a bottleneck.

Instead, create a layer tree and assign one or more images to nodes that
match the images' top layer.  Determining the children of an image is
now exponentially faster as we already know the child images from the
layer graph and the images using the same top layer, which may also be
considered child images based on their history.

On my system with 510 images, a rootful image list drops from 6 secs
down to 0.3 secs.

Also use the tree to compute parent nodes, and to filter intermediate
images for pruning.

### DIFF
--- a/libpod/image/filters.go
+++ b/libpod/image/filters.go
@@ -29,6 +29,26 @@ func CreatedBeforeFilter(createTime time.Time) ResultFilter {
 	}
 }
 
+// IntermediateFilter returns filter for intermediate images (i.e., images
+// with children and no tags).
+func (ir *Runtime) IntermediateFilter(ctx context.Context, images []*Image) (ResultFilter, error) {
+	tree, err := ir.layerTree()
+	if err != nil {
+		return nil, err
+	}
+	return func(i *Image) bool {
+		if len(i.Names()) > 0 {
+			return true
+		}
+		children, err := tree.children(ctx, i, false)
+		if err != nil {
+			logrus.Error(err.Error())
+			return false
+		}
+		return len(children) == 0
+	}, nil
+}
+
 // CreatedAfterFilter allows you to filter on images created after
 // the given time.Time
 func CreatedAfterFilter(createTime time.Time) ResultFilter {
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -856,26 +856,6 @@ func (i *Image) Dangling() bool {
 	return len(i.Names()) == 0
 }
 
-// Intermediate returns true if the image is cache or intermediate image.
-// Cache image has parent and child.
-func (i *Image) Intermediate(ctx context.Context) (bool, error) {
-	parent, err := i.IsParent(ctx)
-	if err != nil {
-		return false, err
-	}
-	if !parent {
-		return false, nil
-	}
-	img, err := i.GetParent(ctx)
-	if err != nil {
-		return false, err
-	}
-	if img != nil {
-		return true, nil
-	}
-	return false, nil
-}
-
 // User returns the image's user
 func (i *Image) User(ctx context.Context) (string, error) {
 	imgInspect, err := i.inspect(ctx, false)
@@ -1214,7 +1194,7 @@ func splitString(input string) string {
 // the parent of any other layer in store. Double check that image with that
 // layer exists as well.
 func (i *Image) IsParent(ctx context.Context) (bool, error) {
-	children, err := i.getChildren(ctx, 1)
+	children, err := i.getChildren(ctx, false)
 	if err != nil {
 		if errors.Cause(err) == ErrImageIsBareList {
 			return false, nil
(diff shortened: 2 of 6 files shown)


mheon:
I don't know if I like making this part of the signature - asking someone
writing an app against Libpod to know what to do here (and to properly
handle cases where images are created and removed from under us) doesn't
seem right. Maybe a cache around getting images from the store?

mheon:
@vrothberg That would be a good compromise.

baude:
lgtm

vrothberg:
Let's wait for @mtrmac's head nod before merging. The performance diff is almost too good to be true/correct

vrothberg:
CI was green. Now with intermediate updates.

@baude @mheon @mtrmac PTAL

vrothberg:
@rhatdan @nalind PTAL

TomSweeneyRedHat:
LGTM, would like a head nod from @nalind

libpod/image/filters.go
```diff
@@ -29,6 +29,26 @@ func CreatedBeforeFilter(createTime time.Time) ResultFilter {
 	}
 }
 
+// IntermediateFilter returns filter for intermediate images (i.e., images
+// without children).
```
mtrmac, line 33: This doesn’t mention the tag check.

libpod/image/image.go
```diff
@@ -856,24 +856,14 @@ func (i *Image) Dangling() bool {
 	return len(i.Names()) == 0
 }
 
-// Intermediate returns true if the image is cache or intermediate image.
-// Cache image has parent and child.
+// Intermediate returns true if the image is a cache or intermediate image
+// (i.e., it has children and a parent).
 func (i *Image) Intermediate(ctx context.Context) (bool, error) {
```
mtrmac, line 861: Wasn’t the only caller removed as part of this PR? (And if not, I’d prefer the caller to have to supply the `layerTree` to encourage caching it over multiple calls.)

libpod/image/filters.go
```diff
@@ -29,6 +29,26 @@ func CreatedBeforeFilter(createTime time.Time) ResultFilter {
...
+func (ir *Runtime) IntermediateFilter(ctx context.Context, images []*Image) (ResultFilter, error) {
```
mtrmac, line 34: This is not new, but it’s confusing how `Runtime.IntermediateFilter` and `layerTree.isIntermediate` have different semantics. It would be nice for the two to use different names — not that I know _what_ names.

libpod/image/image.go
```diff
@@ -1289,58 +1279,11 @@ func areParentAndChild(parent, child *imgspecv1.Image) bool {
 
 // GetParent returns the image ID of the parent. Return nil if a parent is not found.
 func (i *Image) GetParent(ctx context.Context) (*Image, error) {
-	var childLayer *storage.Layer
-	images, err := i.imageruntime.GetImages()
+	tree, err := i.imageruntime.layerTree()
```
mtrmac, line 1282: I’d prefer the caller to have to supply the `layerTree` to encourage caching it over multiple calls.

libpod/image/image.go
```diff
@@ -1357,60 +1300,12 @@ func (i *Image) GetChildren(ctx context.Context) ([]string, error) {
 
 // getChildren returns a list of at most "max" imageIDs that depend on the image
 func (i *Image) getChildren(ctx context.Context, max int) ([]string, error) {
```
mtrmac, line 1302: (I’d prefer the caller, incl. the caller of the higher-level `GetChildren` to have to supply the `layerTree` to encourage caching it over multiple calls.)

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+	"context"
+
+	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+// layerTree is an internal representation of local layers.
+type layerTree struct {
+	// nodes is the actual layer tree.
+	nodes map[string]*layerNode
```
mtrmac, line 13: Is the map key a layer ID? Document that, please.

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+	// ociCache is cache for Image.ID -> OCI Image.
+	ociCache map[string]*ociv1.Image
+}
+
+// node returns a layerNode for the specified layer.
+func (t *layerTree) node(layer string) *layerNode {
```
mtrmac, line 19: Is the map key a layer ID? Document that, please.

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+			continue
+		}
+		parent := tree.node(layer.Parent)
+		node.parent = parent
+		parent.children = append(parent.children, node)
+	}
+
+	// Now assign the images to each (top) layer.
+	for i := range images {
+		img := images[i] // do not leak loop variable outside the scope
```
mtrmac, line 77: Does this actually make a difference? There isn’t a closure that would survive the loop iteration.

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+// layerNode is a node in a layerTree.  It's ID is the key in a layerTree.
+type layerNode struct {
+	children []*layerNode
+	images   []*Image
+	parent   *layerNode
+}
+
+// layerTree extracts a layerTree from the layers in the local storage and
+// relates them to the specified images.
+func (ir *Runtime) layerTree() (*layerTree, error) {
```
mtrmac, line 48: It would be nice if `GetLayersMapWithImageInfo` were rebuilt on top of this. OTOH, sure, not in scope.

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+		if err != nil {
+			return false, err
+		}
+		// History check.
+		return areParentAndChild(parentOCI, childOCI), nil
+	}
+
+	// First check images where parent is also the parent layer.
+	for _, childNode := range node.children {
+		for _, childImage := range childNode.images {
```
mtrmac, line 129: Instead of a `checkParent` helper, or in addition to it, make this whole loop a `addPossibleChildrenFromNode` helper, and it will not have to be duplicated.

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+			return nil, errors.Errorf("top layer %s of image %s not found in layer tree", img.TopLayer(), img.ID())
+		}
+		node.images = append(node.images, img)
+	}
+
+	return &tree, nil
+}
+
+// children returns the image IDs of all children. Use `max` to specify if you
+// want to query for all (i.e., -1) or a specific amount of children.  Child
```
mtrmac, line 93: The only caller that wants a limit uses `0`.

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,211 @@
...
+func (t *layerTree) parent(ctx context.Context, child *Image) (*Image, error) {
+	if child.TopLayer() == "" {
+		return nil, nil
+	}
+	node, exists := t.nodes[child.TopLayer()]
+	if !exists {
+		return nil, errors.Errorf("layer not found in layer tree: %q", child.TopLayer())
+	}
+
+	if node.parent == nil {
```
mtrmac, line 170: This is incorrect; the parent could be in `node.images`.

libpod/image/layer_tree.go
mtrmac: s/map key/parameter/

vrothberg:
Updated. Thanks for the great review!

@mtrmac, I created a Jira card to perform some follow-up work to wire in the layer into more APIs (and to also have a closer look at pkg/api/handlers).

libpod/image/layer_tree.go
mtrmac: I don’t think so; that line reads the contents of the `img` variable _at that point_ and appends the pointer to `node.images`. If `img` changes later, that doesn’t affect `node.images`.

The copy/new scope is only necessary when the evaluation of a variable is _deferred_ after the iteration terminates, e.g. `runOneSecondLater(func() { consume(img) })`

libpod/image/layer_tree.go
```diff
@@ -0,0 +1,222 @@
...
+		for _, childImage := range node.images {
+			isChild, err := checkParent(childImage)
+			if err != nil {
+				return foundChildren, err
+			}
+			if isChild {
+				foundChildren = true
+				children = append(children, childImage.ID())
+				if all {
+					return foundChildren, nil
```
mtrmac, line 143: (Absolutely no-blocking: A `break` would be a tiny bit easier to follow for me.)

rhatdan:
/lgtm
/hold
@vrothberg You can either fix @mtrmac nit in this PR or a follow on PR, you can decide to remove the hold.

vrothberg:
/hold cancel

I can fix it with the follow-up work :+1:

vrothberg:
@mheon, I'd love to get this into the v2.0 branch. Cool with it?

Issue #6064, opened 2020-05-01
Allowing rootless containers to be binded to a specific IP


Diniboy1123:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

By the nature of POSIX it's impossible to bind an application to a specific interface as far as I know, especially if one user doesn't have root. Though it should be possible to send the container traffic to  a specific IP address.

In my case the goal would be to have all the container traffic sent thru a wg0 wireguard VPN interface that always gets the same static IP and can be used without root by tools like curl simply. Assuming it would just work fine by redirecting slirp4netns traffic there. But I couldn't find an option for that.

**Steps to reproduce the issue:**

Not relevant by the nature of the FR.

**Describe the results you received:**

Not relevant by the nature of the FR.

**Describe the results you expected:**

Not relevant by the nature of the FR.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

Any should be fine.

**Output of `podman info --debug`:**

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

**Additional environment details (AWS, VirtualBox, physical, etc.):** Physical host machine running Fedora 31.

5eraph:
Hi, I was working on this rootless-containers/slirp4netns#200  and it should be included in slirp4netns v1.1.0.

Although we will have to integrate some way with libpod and I am not sure how. `slirp4netns` options are currently not directly exposed - libpod rather calls preferred options on its own. 

We need to expose slirp4netns in non intrusive way and possibly even expose iptables POSTROUTE to provide this feature for non rootless containers as well.

I assume we could extend [ContainerNetworkConfig](https://github.com/containers/libpod/blob/f6f71724949c22cf89a44015c29bd5a144db7390/pkg/specgen/specgen.go#L289-L352) to hold outbound addresses and expose them through flags. But I am not sure whether it is braking change considering standards libpod follows.

I would like to hear some thoughts and suggestions from contributors.

mheon:
I think we're at the point of needing an extra flag on `podman run` for extra network configuration, as seen by the PR to switch rootless port forwarding backend from @aleks-mariusz - a `podman run --network-opt KEY=VALUE` might be appropriate here, to pass general network configuration into the container without further polluting the incredibly complicated parsing of `--network`

rhatdan:
Yes I agree this calls for another option.

5eraph:
[...]

PR #7216, opened 2020-08-04
support outbound-addr

5eraph:
Changes to support binding to specific outbound address. Mentioned in #6064. Follow ups to #6324 and #6965.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -560,9 +560,14 @@ Valid values are:
 - `ns:<path>`: path to a network namespace to join
 - `private`: create a new namespace for the container (default)
 - `slirp4netns[:OPTIONS,...]`: use slirp4netns to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
-  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
-  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
-  **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
+  - **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`). Default is false.
+  - **enable_ipv6=true|false**: Enable IPv6. Default is false. (Required for `outbound_addr6`).
+  - **outbound_addr=INTERFACE**: Specify the outbound interface slirp should bind to (ipv4 traffic only).
+  - **outbound_addr=IPv4**: Specify the outbound ipv4 address slirp should bind to.
+  - **outbound_addr6=INTERFACE**: Specify the outbound interface slirp should bind to (ipv6 traffic only).
+  - **outbound_addr6=IPv6**: Specify the outbound ipv6 address slirp should bind to.
+  - **port_handler=rootlesskit**: Use rootlesskit for port forwarding. Default.
+  - **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
 
 **--network-alias**=*alias*
 
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -171,6 +171,8 @@ type slirpFeatures struct {
 	HasMTU                 bool
 	HasEnableSandbox       bool
 	HasEnableSeccomp       bool
+	HasOutboundAddr        bool
+	HasIPv6                bool
 }
 
 type slirp4netnsCmdArg struct {
@@ -197,6 +199,8 @@ func checkSlirpFlags(path string) (*slirpFeatures, error) {
 		HasMTU:                 strings.Contains(string(out), "--mtu"),
 		HasEnableSandbox:       strings.Contains(string(out), "--enable-sandbox"),
 		HasEnableSeccomp:       strings.Contains(string(out), "--enable-seccomp"),
+		HasOutboundAddr:        strings.Contains(string(out), "--outbound-addr"),
+		HasIPv6:                strings.Contains(string(out), "--enable-ipv6"),
 	}, nil
 }
 
@@ -225,21 +229,64 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
 
 	isSlirpHostForward := false
 	disableHostLoopback := true
+	enableIPv6 := false
+	outboundAddr := ""
+	outboundAddr6 := ""
+
 	if ctr.config.NetworkOptions != nil {
 		slirpOptions := ctr.config.NetworkOptions["slirp4netns"]
 		for _, o := range slirpOptions {
-			switch o {
-			case "port_handler=slirp4netns":
-				isSlirpHostForward = true
-			case "port_handler=rootlesskit":
-				isSlirpHostForward = false
-			case "allow_host_loopback=true":
-				disableHostLoopback = false
-			case "allow_host_loopback=false":
-				disableHostLoopback = true
+			parts := strings.Split(o, "=")
+			option, value := parts[0], parts[1]
+
+			switch option {
+			case "port_handler":
+				switch value {
+				case "slirp4netns":
+					isSlirpHostForward = true
+				case "rootlesskit":
+					isSlirpHostForward = false
+				default:
+					return errors.Errorf("unknown port_handler for slirp4netns: %q", value)
+				}
+			case "allow_host_loopback":
+				switch value {
+				case "true":
+					disableHostLoopback = false
+				case "false":
+					disableHostLoopback = true
+				default:
+					return errors.Errorf("invalid value of allow_host_loopback for slirp4netns: %q", value)
+				}
+			case "enable_ipv6":
+				switch value {
+				case "true":
+					enableIPv6 = true
+				case "false":
+					enableIPv6 = false
+				default:
+					return errors.Errorf("invalid value of enable_ipv6 for slirp4netns: %q", value)
+				}
+			case "outbound_addr":
+				ipv4 := net.ParseIP(value)
+				if ipv4 == nil || ipv4.To4() == nil {
+					_, err := net.InterfaceByName(value)
+					if err != nil {
+						return errors.Errorf("invalid outbound_addr %q", value)
+					}
+				}
+				outboundAddr = value
+			case "outbound_addr6":
+				ipv6 := net.ParseIP(value)
+				if ipv6 == nil || ipv6.To4() != nil {
+					_, err := net.InterfaceByName(value)
+					if err != nil {
+						return errors.Errorf("invalid outbound_addr6: %q", value)
+					}
+				}
+				outboundAddr6 = value
 			default:
 				return errors.Errorf("unknown option for slirp4netns: %q", o)
-
 			}
 		}
 	}
@@ -260,6 +307,30 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
 	}
 	if slirpFeatures.HasEnableSeccomp {
 		cmdArgs = append(cmdArgs, "--enable-seccomp")
+	}
+
+	if enableIPv6 {
+		if !slirpFeatures.HasIPv6 {
+			return errors.Errorf("enable_ipv6 not supported")
+		}
+		cmdArgs = append(cmdArgs, "--enable-ipv6")
+	}
+
+	if outboundAddr != "" {
+		if !slirpFeatures.HasOutboundAddr {
+			return errors.Errorf("outbound_addr not supported")
+		}
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr=%s", outboundAddr))
+	}
+
+	if outboundAddr6 != "" {
+		if !slirpFeatures.HasOutboundAddr || !slirpFeatures.HasIPv6 {
+			return errors.Errorf("outbound_addr6 not supported")
+		}
+		if !enableIPv6 {
+			return errors.Errorf("enable_ipv6=true is required for outbound_addr6")
+		}
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr6=%s", outboundAddr6))
 	}
 
 	var apiSocket string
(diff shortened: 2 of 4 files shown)


libpod/networking_linux.go
```diff
@@ -225,21 +226,55 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
 
 	isSlirpHostForward := false
 	disableHostLoopback := true
+	var outboundAddr string
+	outboundAddr = nil
```
zhangguanzhang, line 230: The default string type of golang is null, which is "", can not use nil in string

libpod/networking_linux.go
```diff
@@ -225,21 +226,55 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
+
+	var outboundAddr6 string
+	outboundAddr6 = nil
```
zhangguanzhang, line 233: The default string type of golang is null, which is "", can not use nil in string

libpod/networking_linux.go
```diff
@@ -262,6 +297,14 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
 		cmdArgs = append(cmdArgs, "--enable-seccomp")
 	}
 
+	if outboundAddr != nil {
```
zhangguanzhang, line 300: The default string type of golang is null, which is "", can not use nil in string

libpod/networking_linux.go
```diff
@@ -262,6 +297,14 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr=%s", outboundAddr))
+	}
+
+	if outboundAddr6 != nil {
```
zhangguanzhang, line 304: The default string type of golang is null, which is "", can not use nil in string

libpod/networking_linux.go
```diff
@@ -225,21 +226,55 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
+					if err != nil {
+						return errors.Wrapf(err, "invalid outbound_addr")
+					}
+				}
+				outboundAddr = ipv4OrInterface
+				continue
+			case slirpFeatures.HasOutboundAddr && strings.Contains(o, "outbound_addr6="):
+				ipv6OrInterface := string(o[15:])
+				ipv6 := net.ParseIP(ipv6OrInterface)
+				if ipv6 == nil || ipv6.to4() != nil {
```
zhangguanzhang, line 268: The default string type of golang is null, which is "", can not use nil in string

libpod/networking_linux.go
```diff
@@ -225,21 +226,55 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
 			case "allow_host_loopback=true":
 				disableHostLoopback = false
+				continue
 			case "allow_host_loopback=false":
 				disableHostLoopback = true
+				continue
+			}
+			// Parametrized options
+			switch {
+			case slirpFeatures.HasOutboundAddr && strings.Contains(o, "outbound_addr="):
```
mheon, line 254: We should merge these two switch statements - there's no real reason to have two. Probably want to split O on `=` and do a comparison on the first element.

5eraph: merged

libpod/networking_linux.go
5eraph: fixed

zhangguanzhang:
you must add the test/e2e/

5eraph:
Yes I know @zhangguanzhang, will do. I did not have time to test in my env so far, I will include tests later. That's why it is WIP so far. I did not expect that fast reviews.

Thank you both for reviews and suggestions.

5eraph:
The test I added yesterday is not correct. I will fix it today.

I wonder do we need ipv6 test as well? Slirp4netns does not test ipv6 yet. 
Also any chance we can include this in version  2.1?

libpod/networking_linux.go
```diff
@@ -262,6 +297,20 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
+			return errors.Wrapf(err, "outbound_addr not supported")
+		}
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr=%s", outboundAddr))
+	}
+
+	if outboundAddr6 != "" {
+		if !slirpFeatures.HasOutboundAddr {
+			return errors.Wrapf(err, "outbound_addr6 not supported")
+		}
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr6=%s", outboundAddr6))
```
AkihiroSuda, line 311: You need to run `slirp4netns` with `--enable-ipv6`

5eraph: thanks, added

5eraph:
OK, I finished and tested changes and everything works as expected. But ... I am not sure about tests. We need to check whether we are running podman on slirp4netns with `outbound-addr`, but right now there is no exposed function in the test suite to do this. (Or did I miss it?) Anyway I implemented it with SystemExec and this should be sufficient for now, but it just does not feel clean enough. Any suggestions?

mheon:
For testing - I would restrict testing to rootless only, which guarantees it will be run on a slirp4netns system. You'll probably have to use localhost as the source address, I don't think we can guarantee a stable external IP on the test systems.

5eraph:
Yes that is what I did in latest commit. Regarding restriction to rootless only `--network slirp4netns` should guarantee slirp4netns system or am I wrong?

mheon:
Yes - you're right, that would guarantee slirp would be used, even as root

libpod/networking_linux.go
```diff
@@ -262,6 +297,21 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
+		}
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr=%s", outboundAddr))
+	}
+
+	if outboundAddr6 != "" {
+		if !slirpFeatures.HasOutboundAddr {
+			return errors.Errorf("outbound_addr6 not supported")
+		}
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--outbound-addr6=%s", outboundAddr6))
+		cmdArgs = append(cmdArgs, "--enable-ipv6")
```
AkihiroSuda, line 312: We should also have a way to enable ipv6 without explicitly specifying the outbound address

5eraph: Alright I exposed `enable_ipv6`. I was not sure about it being automatically enabled with `outbound_addr6`. It does not feel transparent enough imo (I am not fan of enabling flags because of others require them). So right now it works only when `enable_ipv6=true` otherwise shows error with explanation. 
Is it sufficient or would you like me to implement it some other way? @AkihiroSuda

5eraph:
Alright this should be complete as is. Not sure about IPv6 tests as these are not implemented in slirp4netns either. Except for that we should be good to go.

/assign @mrunalp

libpod/networking_linux.go
```diff
@@ -225,21 +229,64 @@ func (r *Runtime) setupRootlessNetNS(ctr *Container) error {
...
+			case "allow_host_loopback":
+				switch value {
+				case "true":
+					disableHostLoopback = false
+				case "false":
+					disableHostLoopback = true
+				default:
+					return errors.Errorf("invalid value of allow_host_loopback for slirp4netns: %q", value)
+				}
+			case "enable_ipv6":
```
mheon, line 261: Can you add this to the manpages?

5eraph: sure

libpod/networking_linux.go
AkihiroSuda: Looks good

mheon:
/approve

Restarted tests in case they flaked.

mheon:
Tests seem green.
@rhatdan @baude @vrothberg @giuseppe PTAL

mheon:
LGTM, for reference

docs/source/markdown/podman-create.1.md
```diff
@@ -563,6 +563,11 @@ Valid values are:
   **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
   **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
   **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
+  **enable_ipv6=true|false**: Enable IPv6.  Default to false. (Required for `outbound_addr6`)
+  **outbound_addr=IPv4**: specify outbound ipv4 address slirp should bind to
+  **outbound_addr=INTEFACE**: specify outbound interface slirp should bind to (ipv4 traffic only)
+  **outbound_addr6=IPv6**: specify outbound ipv6 address slirp should bind to
+  **outbound_addr6=INTERFACE**: specify outbound interface slirp should bind to (ipv6 traffic only)
```
TomSweeneyRedHat, line 570: It would be good to have this all in alphabetical order.

docs/source/markdown/podman-create.1.md
```diff
@@ -563,6 +563,11 @@ Valid values are:
   **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
   **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
   **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
+  **enable_ipv6=true|false**: Enable IPv6.  Default to false. (Required for `outbound_addr6`)
```
TomSweeneyRedHat, line 566:
```suggestion
  **enable_ipv6=true|false**: Enable IPv6.  Default is false. (Required for `outbound_addr6`).
```

docs/source/markdown/podman-create.1.md
```diff
@@ -563,6 +563,11 @@ Valid values are:
...
+  **outbound_addr=IPv4**: specify outbound ipv4 address slirp should bind to
```
TomSweeneyRedHat, line 567:
```suggestion
  **outbound_addr=IPv4**: Specify the outbound ipv4 address slirp should bind to.
```

docs/source/markdown/podman-create.1.md
```diff
@@ -563,6 +563,11 @@ Valid values are:
...
+  **outbound_addr=INTEFACE**: specify outbound interface slirp should bind to (ipv4 traffic only)
```
TomSweeneyRedHat, line 568:
```suggestion
  **outbound_addr=INTEFACE**: Specify theoutbound interface slirp should bind to (ipv4 traffic only).
```

docs/source/markdown/podman-create.1.md
```diff
@@ -563,6 +563,11 @@ Valid values are:
...
+  **outbound_addr6=IPv6**: specify outbound ipv6 address slirp should bind to
```
TomSweeneyRedHat, line 569:
```suggestion
  **outbound_addr6=IPv6**: Specify the outbound ipv6 address slirp should bind to.
```

docs/source/markdown/podman-create.1.md
TomSweeneyRedHat, line 570:
```suggestion
  **outbound_addr6=INTERFACE**: Specify the outbound interface slirp should bind to (ipv6 traffic only).
```

docs/source/markdown/podman-create.1.md
5eraph: updated

docs/source/markdown/podman-create.1.md
5eraph: updated

docs/source/markdown/podman-create.1.md
5eraph: updated

docs/source/markdown/podman-create.1.md
5eraph: updated

docs/source/markdown/podman-create.1.md
5eraph: updated

docs/source/markdown/podman-create.1.md
5eraph: done

5eraph:
Hi @TomSweeneyRedHat, reuqested docs changes implemented.

docs/source/markdown/podman-create.1.md
```diff
@@ -560,9 +560,14 @@ Valid values are:
 - `ns:<path>`: path to a network namespace to join
 - `private`: create a new namespace for the container (default)
 - `slirp4netns[:OPTIONS,...]`: use slirp4netns to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
-  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
-  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
-  **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
+  - **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
```
TomSweeneyRedHat, line 563: if you get any other comments that need tending, this was in the existing lines and I didn't pick up on it.  A minor, minor nit.
```suggestion
  - **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default is false.
```

TomSweeneyRedHat:
tests aren't hip.
One possible small nit to chase if you're ambitious or have other changes to make.
LGTM otherwise.

docs/source/markdown/podman-create.1.md
5eraph: fixed as well 🙂

5eraph:
@TomSweeneyRedHat comment updated. Well I do not have anything else for now which has to be included in this pull request, so from my side we are good to go, unless you would like to add something else ofc. 🙂

rhatdan:
/lgtm
Thanks @5eraph

PR #7217, opened 2020-08-04
Remove duplicated code

rhatdan:
We have duplicated alias handling, removing.

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -518,21 +518,3 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	)
 	return &createFlags
 }
-
-func AliasFlags(_ *pflag.FlagSet, name string) pflag.NormalizedName {
-	switch name {
-	case "healthcheck-command":
-		name = "health-cmd"
-	case "healthcheck-interval":
-		name = "health-interval"
-	case "healthcheck-retries":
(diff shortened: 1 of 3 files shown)


cmd/podman/common/create.go
```diff
@@ -518,21 +518,3 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
...
-	case "healthcheck-interval":
-		name = "health-interval"
-	case "healthcheck-retries":
-		name = "health-retries"
-	case "healthcheck-start-period":
-		name = "health-start-period"
-	case "healthcheck-timeout":
-		name = "health-timeout"
-	case "net":
-		name = "network"
```
edsantiago, line 535: `utils/alias.go` includes a "timeout" alias for "time". As best I can tell this is fine, since neither `containers/create.go` nor `run.go` register either option. A few manual tests seem to confirm: `--timeout/--time` continue to yield "unknown flag" errors. I just want to point this out in case there's something I've missed.

rhatdan: Thanks for checking. I was looking for similar, but only looked to make sure what I was removing was present in Alias.

TomSweeneyRedHat: Yeah, I looked at that too.  If the timeout/time definition was here, but not below, then that would be an issue, but it looked OK to me as is.

PR #7218, opened 2020-08-04
Build without Volume Mount

rhdojun:
as it's currently not working for me and maybe for others.
But I don't know about other people's opinion, I find building this way
even more elegent as the build tooling writer has more
control over what to actually take.

```
$ python -c "import this"
The Zen of Python, by Tim Peters

Beautiful is better than ugly.
Explicit is better than implicit. <- i.e. it's this
Simple is better than complex.
[...]
```

I'll still continue debugging the volume mounting though, no worries.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ENV GOSRC=$GOPATH/src/github.com/containers/podman
 ENV SCRIPT_BASE=./contrib/cirrus
 ENV PACKER_BASE=$SCRIPT_BASE/packer
 
-ADD / $GOSRC
+ADD . $GOSRC
 WORKDIR $GOSRC
 
 # Re-use repositories and package setup as in VMs under CI
--- a/Makefile
+++ b/Makefile
@@ -267,7 +267,10 @@ libpodimage: ## Build the libpod image
 
 .PHONY: dbuild
 dbuild: libpodimage
-	${CONTAINER_RUNTIME} run --name=${LIBPOD_INSTANCE} --privileged -v ${PWD}:/go/src/${PROJECT} --rm ${LIBPOD_IMAGE} make all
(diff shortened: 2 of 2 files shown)


mheon:
This does require that we rebuild the image every time, though.

I tend to think we should go the other way - not add the source to the image, but instead exclusively mount it in.

rhdojun:
Did some debugging, aaand:

> This does require that we rebuild the image every time, though.

Local build artifacts [should be in the build image](https://github.com/containers/podman/pull/7218/commits/0202eabe06f293eccef5dfb9e93322f7d443c765#diff-3254677a7917c6c01f55212f86c57fbfR12) the next round as well.

> I tend to think we should go the other way - not add the source to the image, but instead exclusively mount it in.

That might be an option as well. This doesn't come without disadvantages though, as of course it potentially modifies the local filesystem as well in undesired ways if there's a bug in the tooling. (E.g. with my shell script I also locally deleted my podman completely at one step, which then also stopped me from building a new podman.)

vrothberg:
@mheon @rhdojun, ping :) I let you decide about a direction.

rhdojun:
I personally understand that both directions have their pros and cons but felt much more comfortable with the copy step when using podman as build tool.

Also looked around a little what other parties/solutions would do, and found that [Docker's multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) go a similar path:

```
FROM buildenv:latest
//build app

FROM deployenv:latest
COPY --from=0 /go/src/github.com/alexellis/href-counter/app     <- here
//use app
```

rhatdan:
@mheon @rhdojun We need a final decision on this.  @mheon if you want to open this for discussion at meetup next week, we can do it then.

mheon:
We completely removed the endpoint, AFAIK, so this is likely no longer
relevant.

Issue #7127, opened 2020-07-28
podman-remote: bunch of commands need to reject -l


edsantiago:
`-l` doesn't work with podman-remote, but a lot of commands accept it anyway:
```console
$ ./bin/podman-remote ps -l
CONTAINER ID  IMAGE   COMMAND  CREATED  STATUS  PORTS   NAMES

$  ./bin/podman-remote diff -l
Error: invalid character 'M' looking for beginning of value

$ ./bin/podman-remote inspect -l
[]

$ ./bin/podman-remote [ wait, start, stop, kill, port, and probably more ]
```

PR #7221, opened 2020-08-04
remove --latest for all remote commands

baude:
instead of hiding the latest options for podman-remote or catching an error if podman --remote <cmd> -l is used, we no longer add the latest option to any remote command.  podman will error with a "unknown flag" option.

Fixes: #7127

### DIFF
--- a/cmd/podman/validate/args.go
+++ b/cmd/podman/validate/args.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/containers/podman/v2/cmd/podman/registry"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


cmd/podman/validate/args.go
```diff
@@ -98,6 +102,9 @@ func CheckAllLatestAndCIDFile(c *cobra.Command, args []string, ignoreArgLen bool
 // CheckAllLatestAndPodIDFile checks that --all and --latest are used correctly.
 // If withIDFile is set, also check for the --pod-id-file flag.
 func CheckAllLatestAndPodIDFile(c *cobra.Command, args []string, ignoreArgLen bool, withIDFile bool) error {
+	if registry.IsRemote() {
```
rhatdan, line 105: I think you still want to do this check to make sure the user did not specify --all and --podidfile foobar.

TomSweeneyRedHat:
Tests aren't hip @baude 
LGTM
once @rhatdan's question is addressed.

cmd/podman/validate/args.go
baude: the current logic does not indicate that does it? i looked at this a couple of times.

vrothberg: I think we still need it as we also check for `args`.

mheon: I think he's right, this function does more than just check latest so we can't completely bypass

cmd/podman/validate/latest.go
```diff
@@ -7,9 +7,8 @@ import (
 
 func AddLatestFlag(cmd *cobra.Command, b *bool) {
 	// Initialization flag verification
-	cmd.Flags().BoolVarP(b, "latest", "l", false,
-		"Act on the latest container podman is aware of\nNot supported with the \"--remote\" flag")
-	if registry.IsRemote() {
-		_ = cmd.Flags().MarkHidden("latest")
+	if !registry.IsRemote() {
+		cmd.Flags().BoolVarP(b, "latest", "l", false,
+			"Act on the latest container podman is aware of\nNot supported with the \"--remote\" flag")
```
vrothberg, line 12: I had troubles brain parsing the error message.
Could we reduce it to "--lastet is not supported for remote calls" or something similar?

edsantiago:
Treating all failures as flakes. fedora-31 was the usual 3-line cirrus error. fedora-32 was:
```
Trying to pull registry.access.redhat.com/ubi8-init...
           Get "https://registry.access.redhat.com/v2/": net/http: TLS handshake timeout
         Error: unable to pull registry.access.redhat.com/ubi8-init: Error initializing source docker://registry.access.redhat.com/ubi8-init:latest: error pinging docker registry registry.access.redhat.com: Get "https://registry.access.redhat.com/v2/": net/http: TLS handshake timeout
```

cmd/podman/validate/args.go
```diff
@@ -47,6 +48,9 @@ func IDOrLatestArgs(cmd *cobra.Command, args []string) error {
 // CheckAllLatestAndCIDFile checks that --all and --latest are used correctly.
 // If cidfile is set, also check for the --cidfile flag.
 func CheckAllLatestAndCIDFile(c *cobra.Command, args []string, ignoreArgLen bool, cidfile bool) error {
+	if registry.IsRemote() {
+		return nil
```
vrothberg, line 52: I still think this check is still needed. We are also looking at `args`

baude:
this is ready

cmd/podman/validate/args.go
baude: ok, lemme see.

vrothberg:
Restarted three failed jobs. Looked like flakes.

PR #7223, opened 2020-08-04
Unconditionally retrieve pod names via API

### LINKED ISSUE #7214: "PodName" property in Containers object never filled by the Rest API
/kind bug

**Description**

When listing containers through the Rest API there is PodName property which should be there according to the documentation but it's always empty string. "Pod" property contains the Pod ID and can be used to identify the pods instead.

1.  sudo curl --unix-socket /run/podman/podman.sock http://d/v1.24/libpod/containers/json | jq
``` json  
{
    "Command": [
      "bash"
    ],
...
    "Pod": "8cc10866fa4c22029afc01846525ed96fa726384ad95884b7d99e9a09ed4c642",
    "PodName": "",
...
  }
]
```

```
podman version 2.0.4
```

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    path: /usr/bin/conmon
  cpus: 4
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: sequioa
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.7.8-200.fc32.x86_64
  linkmode: dynamic
  memFree: 241860608
  memTotal: 11945615360
  ociRuntime:
    name: runc
    package: runc-1.0.0-144.dev.gite6555cc.fc32.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc10+dev
      commit: fbdbaf85ecbc0e077f336c03062710435607dbf1
      spec: 1.0.1-dev
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 4742705152
  swapTotal: 6039793664
  uptime: 26h 3m 13.98s (Approximately 1.08 days)
registries:
  localhost:5000:
    Blocked: false
    Insecure: true
    Location: localhost:5000
    MirrorByDigestOnly: false
    Mirrors: []
    Prefix: localhost:5000
  search:
  - localhost:5000
  - docker.io
  - registry.fedoraproject.org
  - quay.io
  - registry.access.redhat.com
  - registry.centos.org
store:
  containerStore:
    number: 2
    paused: 0
    running: 2
    stopped: 0
  graphDriverName: overlay
  graphOptions:
  graphRoot: /var/lib/containers/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "true"
  imageStore:
    number: 8
  runRoot: /var/run/containers/storage
  volumePath: /var/lib/containers/storage/volumes
version:
  APIVersion: 1
  Built: 0
  BuiltTime: Thu Jan  1 01:00:00 1970
  GitCommit: ""
  GoVersion: go1.14.6
  OsArch: linux/amd64
  Version: 2.0.4

```

```
podman-2.0.4-1.fc32.x86_64
```

mheon:
The ListContainers API previously had a Pod parameter, which determined if pod name was returned (but, notably, not Pod ID, which was returned unconditionally). This was fairly confusing, so we decided to deprecate/remove the parameter and return it unconditionally.

To do this without serious performance implications, we need to avoid expensive JSON decodes of pod configuration in the DB. The way our Bolt tables are structured, retrieving name given ID is actually quite cheap, but we did not expose this via the Libpod API. Add a new GetName API to do this.

Fixes #7214

### DIFF
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -661,11 +661,10 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//    type: boolean
 	//    description: Include namespace information
 	//    default: false
-	//  - in: query
 	//    name: pod
 	//    type: boolean
 	//    default: false
-	//    description: Include Pod ID and Name if applicable
+	//    description: Ignored. Previously included details on pod name and ID that are currently included by default.
 	//  - in: query
 	//    name: size
 	//    type: boolean
--- a/pkg/bindings/containers/containers.go
+++ b/pkg/bindings/containers/containers.go
@@ -24,7 +24,7 @@ var (
 // the most recent number of containers.  The pod and size booleans indicate that pod information and rootfs
 // size information should also be included.  Finally, the sync bool synchronizes the OCI runtime and
 // container state.
-func List(ctx context.Context, filters map[string][]string, all *bool, last *int, pod, size, sync *bool) ([]entities.ListContainer, error) { // nolint:typecheck
+func List(ctx context.Context, filters map[string][]string, all *bool, last *int, size, sync *bool) ([]entities.ListContainer, error) { // nolint:typecheck
 	conn, err := bindings.GetClient(ctx)
 	if err != nil {
 		return nil, err
@@ -36,9 +36,6 @@ func List(ctx context.Context, filters map[string][]string, all *bool, last *int
 	}
 	if last != nil {
 		params.Set("limit", strconv.Itoa(*last))
-	}
-	if pod != nil {
-		params.Set("pod", strconv.FormatBool(*pod))
 	}
 	if size != nil {
 		params.Set("size", strconv.FormatBool(*size))
--- a/pkg/bindings/test/containers_test.go
+++ b/pkg/bindings/test/containers_test.go
@@ -510,7 +510,7 @@ var _ = Describe("Podman containers ", func() {
 		Expect(err).To(BeNil())
 		_, err = bt.RunTopContainer(&name2, bindings.PFalse, nil)
 		Expect(err).To(BeNil())
-		containerLatestList, err := containers.List(bt.conn, nil, nil, &latestContainers, nil, nil, nil)
+		containerLatestList, err := containers.List(bt.conn, nil, nil, &latestContainers, nil, nil)
 		Expect(err).To(BeNil())
 		err = containers.Kill(bt.conn, containerLatestList[0].Names[0], "SIGTERM")
 		Expect(err).To(BeNil())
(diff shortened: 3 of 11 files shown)


pkg/bindings/test/containers_test.go
```diff
@@ -755,8 +755,23 @@ var _ = Describe("Podman containers ", func() {
 		// Validate list container with id filter
 		filters := make(map[string][]string)
 		filters["id"] = []string{cid}
-		c, err := containers.List(bt.conn, filters, bindings.PTrue, nil, nil, nil, nil)
+		c, err := containers.List(bt.conn, filters, bindings.PTrue, nil, nil, nil)
 		Expect(err).To(BeNil())
 		Expect(len(c)).To(Equal(1))
 	})
+
+	It("List containers always includes pod information", func() {
```
baude, line 763: dont we need a test in the test/endpoint that would protect this?

mheon: Do we? I think the bindings test captures it adequately - verifies it's in the response?

pkg/api/server/register_containers.go
```diff
@@ -661,11 +661,10 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//    type: boolean
 	//    description: Include namespace information
 	//    default: false
-	//  - in: query
 	//    name: pod
 	//    type: boolean
 	//    default: false
-	//    description: Include Pod ID and Name if applicable
+	//    description: Ignored. Previously included details on pod name and ID that are currently included by default.
```
TomSweeneyRedHat, line 667: Is this a change that might mess up someone who's still expecting to find the Pod ID and name here?  I don't think so, but I'm not 100% sure.

rhatdan: They will get the data whether they specify the option or not.

rhatdan:
LGTM

pkg/bindings/containers/containers.go
```diff
@@ -24,7 +24,7 @@ var (
 // the most recent number of containers.  The pod and size booleans indicate that pod information and rootfs
 // size information should also be included.  Finally, the sync bool synchronizes the OCI runtime and
 // container state.
-func List(ctx context.Context, filters map[string][]string, all *bool, last *int, pod, size, sync *bool) ([]entities.ListContainer, error) { // nolint:typecheck
```
vrothberg, line 27: I prefer to keep it as it would break existing users. I'd rather mark it as deprecated and remove it with 3.0.

mheon: This is just in the bindings code; do we consider that public API? The actual API still handles the request without issue.

baude: i think we'd be bad players to break existing users. on the other hand, we do not want to mislead them.  how would you feel about leaving it but noting in the docs (above the func) that it is a no-op ?

vrothberg: I really want the bindings to be stable. @lsm5 is writing a blog post about them as we wanted to promote the bindings to Go developers. The blog does not use containers.List but could have.

mheon:
Bindings have been restored

mheon:
Rebased, should go green

Issue #7119, opened 2020-07-28
podman-remote: run --rm exits before container is truly removed


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman-remote run --name myc alpine date;./bin/podman-remote commit myc myi;./bin/podman rm -a
Tue Jul 28 17:30:18 UTC 2020
3840e1d6deaadebc70f317f765e6ea015b67d222e575ff77f920f8f67adbaa23
480a2a915ba10c1eb8da260d8f721e07eb1093ba7baa125c86a04c3aab8a14f6

$ ./bin/podman-remote run --rm myi date;./bin/podman-remote rmi myi
Tue Jul 28 17:31:00 UTC 2020
Error: 1 error occurred:
        * could not remove image 3840e1d6deaadebc70f317f765e6ea015b67d222e575ff77f920f8f67adbaa23 as it is being used by 1 containers: image is being used
```
This is two sets of multiple commands on one line. It's not important to have the first set in one line, it just makes it easier to up-arrow to repeat, which you need to do because this is a race condition hence it doesn't always trigger. The second command, of course, must be done as a oneliner because the timing window is short.

A manual `podman-remote rmi myi`, a second or so after the above, works fine.

master @ 288ebec6e, tried only rootless because it would be a pain to add sudos to all the above and I see no point to bothering. I can test root if necessary.

mheon:
We'll need to make the remote version of `podman run` call the Delete endpoint (with Force set to false) on the container once it exits iff `--rm` is set to true. If we get a "container is running" error, the user detached manually with the detach keys, and that's not an error.

edsantiago:
I believe this was fixed by #7224 but the 'Fixes' line was in a comment, not OP

PR #7224, opened 2020-08-04
Handle podman-remote run --rm

rhatdan:
We need to remove the container after it has exited for
podman-remote run --rm commands.  If we don't remove this
container at this step, we open ourselves up to race conditions.

### DIFF
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -500,9 +500,6 @@ func (ic *ContainerEngine) ContainerList(ctx context.Context, options entities.C
 }
 
 func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.ContainerRunOptions) (*entities.ContainerRunReport, error) {
-	if opts.Rm {
-		logrus.Info("the remote client does not support --rm yet")
-	}
 	con, err := containers.CreateWithSpec(ic.ClientCxt, opts.Spec)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


rhatdan:
@edsantiago @mheon PTAL

edsantiago:
Is this intended to fix #7119? If so could you `grep 7119 test/system/*.bats` and remove all matching lines?

mheon:
LGTM

edsantiago:
Flake in [special_testing_rootless](https://github.com/containers/podman/pull/7224/checks?check_run_id=946573502):
```console
 $ podman-remote --url unix:/tmp/podman.Hu25Ai run --rm quay.io/libpod/alpine_labels:latest cat -v /proc/self/attr/current
[+0224s] # read unixpacket @->/run/user/14096/libpod/tmp/socket/32f534034d29cb66f3d3d341a7ddbec1bb13244ee426497a2ffed12579e4f1fb/attach: read: connection reset by peer
```
I saw that on one CI run in #7111 but was unable to reproduce it (despite a lot of effort) hence did not file an issue.

The socket in question is a conmon one, not the podman-remote socket. I suspect there is a troublesome interaction between conmon and podman-remote but I don't know how to even file an issue without a reproducer.

edsantiago:
LGTM but (if this really is intended as a fix for #7119) I'd really like those tests enabled

edsantiago:
FWIW I pulled the PR, removed the `7119` skips from bats tests, and let them run. Root and rootless, seven runs each over an hour or so, no failures. That inspires confidence.

In case it helps:
```console
$ perl -i -ne 'print unless /7119/' test/system/*bats
```

rhatdan:
Fixes: #7119

TomSweeneyRedHat:
LGTM
assuming happy tests

edsantiago:
CI passed all in one go, no flakes.

/lgtm

PR #7227, opened 2020-08-04
run, create: add new security-opt proc-opts 

giuseppe:
it allows to customize the options passed down to the OCI runtime for
setting up the /proc mount.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -511,6 +511,8 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 			}
 
 			switch con[0] {
+			case "proc-opts":
+				s.ProcOpts = strings.Split(con[1], ",")
 			case "label":
 				// TODO selinux opts and label opts are the same thing
 				s.ContainerSecurityConfig.SelinuxOpts = append(s.ContainerSecurityConfig.SelinuxOpts, con[1])
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -756,6 +756,9 @@ Security Options
 - `seccomp=unconfined` : Turn off seccomp confinement for the container
 - `seccomp=profile.json` :  White listed syscalls seccomp Json file to be used as a seccomp filter
 
+- `proc-opts=OPTIONS` : Comma separated list of options to use for the /proc mount.  More details for the
+  possible mount options are specified at **proc(5)** man page.
+
 Note: Labeling can be disabled for all containers by setting label=false in the **containers.conf** (`/etc/containers/containers.conf` or `$HOME/.config/containers/containers.conf`) file.
 
 **--shm-size**=*size*
@@ -1168,7 +1171,7 @@ b
 NOTE: Use the environment variable `TMPDIR` to change the temporary storage location of downloaded container images. Podman defaults to use `/var/tmp`.
 
 ## SEE ALSO
-**subgid**(5), **subuid**(5), **containers.conf**(5), **systemd.unit**(5), **setsebool**(8), **slirp4netns**(1), **fuse-overlayfs**(1).
+**subgid**(5), **subuid**(5), **containers.conf**(5), **systemd.unit**(5), **setsebool**(8), **slirp4netns**(1), **fuse-overlayfs**(1), **proc**(5)**.
 
 ## HISTORY
 October 2017, converted from Docker documentation to Podman by Dan Walsh for Podman <dwalsh@redhat.com>
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -774,6 +774,8 @@ Security Options
 - **no-new-privileges**: Disable container processes from gaining additional privileges
 - **seccomp=unconfined**: Turn off seccomp confinement for the container
 - **seccomp**=_profile.json_:  Allowed syscall list seccomp JSON file to be used as a seccomp filter
+- **proc-opts**=_OPTIONS_ : Comma separated list of options to use for the /proc mount.  More details
+  for the possible mount options are specified at **proc(5)** man page.
 
 Note: Labeling can be disabled for all containers by setting **label=false** in the **containers.conf**(5) file.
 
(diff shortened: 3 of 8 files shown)


pkg/specgen/generate/oci.go
```diff
@@ -18,6 +18,18 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+func setProcfsOpts(s *specgen.SpecGenerator, g *generate.Generator) {
+	if s.ProcfsOpts == nil {
+		return
+	}
+	for i := range g.Config.Mounts {
+		if g.Config.Mounts[i].Destination == "/proc" {
+			g.Config.Mounts[i].Options = s.ProcfsOpts
```
mheon, line 27: Should we append here? Or try and merge them?

libpod/container_config.go
```diff
@@ -99,6 +99,8 @@ type ContainerSecurityConfig struct {
 	// AddCurrentUserPasswdEntry indicates that the current user passwd entry
 	// should be added to the /etc/passwd within the container
 	AddCurrentUserPasswdEntry bool `json:"addCurrentUserPasswdEntry,omitempty"`
+	// ProcfsOpts are the options used for the proc mount.
+	ProcfsOpts []string `json:"procfsOpts,omitempty"`
```
mheon, line 103: I think we can drop this, since it's already in the OCI spec

libpod/container_inspect.go
```diff
@@ -333,6 +333,8 @@ func (c *Container) generateInspectContainerConfig(spec *spec.Spec) *define.Insp
 		ctrConfig.Umask = c.config.Umask
 	}
 
+	ctrConfig.ProcfsOpts = c.config.ProcfsOpts
```
mheon, line 336: We should just be able to locate the container's procfs mount and copy the options from it.

pkg/specgen/generate/oci.go
giuseppe: I think it is better to replace them, how would we deal with something like "nodev", and then "dev"?

TomSweeneyRedHat:
Tests aren't hip @giuseppe

test/e2e/run_test.go
```diff
@@ -803,6 +803,17 @@ USER mail`
...
 
+	It("podman run --procfs-opts", func() {
+		session := podmanTest.Podman([]string{"run", "--procfs-opts", "nosuid,exec,gid=10", fedoraMinimal, "findmnt", "-noOPTIONS", "/proc"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		output := session.OutputToString()
+		// gid= is showed with the user namespace mappings, so ignore the value
+		Expect(output).To(ContainSubstring("gid="))
+		Expect(output).To(ContainSubstring("nosuid"))
+		Expect(output).To(Not(ContainSubstring("exec")))
```
TomSweeneyRedHat, line 814: looks odd that exec is not expected but the other values are?

giuseppe: the kernel won't show the default `exec`.  It would show `noexec` if it was specified

rhatdan:
Does this work if I specify multip --proc-opts each with a comma separated list?

giuseppe:
> Does this work if I specify multip --proc-opts each with a comma separated list?

yes, you can have `--procfs-opts a,b --procfs-opts c,d` and they will handled as `a,b,c,d`

AkihiroSuda:
Can this be a security-opt?

giuseppe:
> Can this be a security-opt?

security-opt are used now for selinux, apparmor and seccomp.   The procfs mount options don't seem too related.  I thought of `--mount` but I've preferred to not re-define the entire proc mount just for changing few mount options.

@rhatdan what do you think?

AkihiroSuda:
hidepid and subset opts in kernel 5.8 looks security-opt-ish

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9ff7258575d5fee011649d20cc56de720a395191

rhatdan:
I kind of like extending the security-opt rather then adding a new very specific option, that hardly anyone will use.
Making this more of a security feature also gives us the ability to make some noise.

cmd/podman/common/specgen.go
```diff
@@ -527,6 +527,8 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 			}
 
 			switch con[0] {
+			case "procfs-opts":
```
rhatdan, line 530: nit: Lets call this "procopts" or "proc-opts"
People understand proc no need for fs.

docs/source/markdown/podman-create.1.md
```diff
@@ -750,6 +750,9 @@ Security Options
 - `seccomp=unconfined` : Turn off seccomp confinement for the container
 - `seccomp=profile.json` :  White listed syscalls seccomp Json file to be used as a seccomp filter
 
+- `procfs-opts=OPTIONS` : Comma separated list of options to use for the /proc mount.
+
```
rhatdan, line 754: Can you mention the security options here?

docs/source/markdown/podman-create.1.md
```diff
@@ -750,6 +750,9 @@ Security Options
 - `seccomp=unconfined` : Turn off seccomp confinement for the container
 - `seccomp=profile.json` :  White listed syscalls seccomp Json file to be used as a seccomp filter
```
rhatdan, line 752: Remove empty line.

docs/source/markdown/podman-create.1.md
giuseppe: these are kernel specific.  Is it fine to add something like: `More details for the possible mount options are specified at **proc(5)** man page.` ?  So we don't have to worry to keep it updated

rhatdan:
@giuseppe Please rebase to get the tests to pass.

giuseppe:
> @giuseppe Please rebase to get the tests to pass.

tests are green

docs/source/markdown/podman-create.1.md
```diff
@@ -755,6 +755,9 @@ Security Options
...
 
+- `proc-opts=OPTIONS` : Comma separated list of options to use for the /proc mount.  More details for the
+  possible mount options are specified at **proc(5)** man page.
```
TomSweeneyRedHat, line 759: proc(5) is not specified in the "See Also" section of this page and probably should be.

docs/source/markdown/podman-run.1.md
```diff
@@ -768,6 +768,8 @@ Security Options
 - **no-new-privileges**: Disable container processes from gaining additional privileges
 - **seccomp=unconfined**: Turn off seccomp confinement for the container
 - **seccomp**=_profile.json_:  Allowed syscall list seccomp JSON file to be used as a seccomp filter
+- **proc-opts**=_OPTIONS_ : Comma separated list of options to use for the /proc mount.  More details
+  for the possible mount options are specified at **proc(5)** man page.
```
TomSweeneyRedHat, line 772: Ditto "not in See Also".

TomSweeneyRedHat:
Couple man page nits, otherwise
LGTM

docs/source/markdown/podman-create.1.md
giuseppe: fixed and pushed a new version

giuseppe:
> Couple man page nits, otherwise
> LGTM

pushed a fixed version

PR #7236, opened 2020-08-05
Ensure that exec errors write exit codes to the DB

### LINKED ISSUE #6893: podman-remote exec return codes incorrect
Several integration tests that are current disabled review that podman-remote sets exit codes of 0 when commands have failed.

exec foobar /etc should return a 126 (returns 0 today)
exec foobar notarealcommand should return a 127 (returns 0 today)

These can be verified with exec_test.go: 229 and :240

mheon:
In local Podman, the frontend interprets the error and exit code given by the Exec API to determine the appropriate exit code to set for Podman itself; special cases like a missing executable receive special exit codes.

Exec for the remote API, however, has to do this inside Libpod itself, as Libpod will be directly queried (via the Inspect API for exec sessions) to get the exit code. This was done correctly when the exec session started properly, but we did not properly handle cases where the OCI runtime fails before the exec session can properly start. Making two error returns that would otherwise not set exit code actually do so should resolve the issue.

Fixes #6893

### DIFF
--- a/libpod/container_exec.go
+++ b/libpod/container_exec.go
@@ -415,6 +415,13 @@ func (c *Container) ExecHTTPStartAndAttach(sessionID string, httpCon net.Conn, h
 
 	execOpts, err := prepareForExec(c, session)
 	if err != nil {
+		session.State = define.ExecStateStopped
+		session.ExitCode = define.ExecErrorCodeGeneric
+
+		if err := c.save(); err != nil {
+			logrus.Errorf("Error saving container %s exec session %s after failure to prepare: %v", err, c.ID(), session.ID())
+		}
+
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


edsantiago:
Um, I just pulled this and I don't see the 126/127 status? Is there still something left to be done?

And if there is, would you mind adding this to the system tests, since there are no tests in e2e?
```diff
diff --git a/test/system/075-exec.bats b/test/system/075-exec.bats
index b2c49510a..540540bab 100644
--- a/test/system/075-exec.bats
+++ b/test/system/075-exec.bats
@@ -21,6 +21,13 @@ load helpers

     run_podman exec $cid rm -f /$rand_filename

+    # Specially defined situations: exec a dir, or no such command.
+    # We don't check the full error message because runc & crun differ.
+    run_podman 126 exec $cid /etc
+    is "$output" ".*permission denied"  "podman exec /etc"
+    run_podman 127 exec $cid /no/such/command
+    is "$output" ".*such file or dir"   "podman exec /no/such/command"
+
     run_podman wait $cid
     is "$output" "0"   "output from podman wait (container exit code)"

```

mheon:
`sudo ./bin/podman --remote exec -t -i testctr /etc` is returning 126 for me.

Are you running the server with `--log-level=debug`? That screws up the logs (Conmon decides to insert things into them) and can break this.

mheon:
Are you sure `exec /no/such/command` should be 127? I think that too is a 126

edsantiago:
>Are you sure exec /no/such/command should be 127? I think that too is a 126

126 and 127 are specified in podman-run(1):
```
       126 The contained command cannot be invoked
       ...
       127 The contained command cannot be found
```

mheon:
Ah, you're right. Fixing.

edsantiago:
Re: my earlier comment ("duh, not working for me"): PEBKAC. I forgot to restart the server after `make`. Sorry about that, it's working much better now.

TomSweeneyRedHat:
LGTM

edsantiago:
/lgtm

Thanks @mheon

PR #7239, opened 2020-08-05
Fix handling of working dir

rhatdan:
Buildah and podman build can create images without a working dir.

FROM fedora
WORKDIR /test

If you build this image with caching twice, the second time the image
will not have a working dir.

Similarly if you execute

podman run --workdir /foobar fedora

It blows up since the workingdir is not created automatically.

Finally there was duplicated code for getting the workingdir
out of an image, that this PR removes.

### DIFF
--- a/pkg/specgen/generate/container_create.go
+++ b/pkg/specgen/generate/container_create.go
@@ -241,13 +241,7 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
 	// If the user did not set an workdir but the image did, ensure it is
 	// created.
 	if s.WorkDir == "" && img != nil {
-		newWD, err := img.WorkingDir(ctx)
-		if err != nil {
-			return nil, err
-		}
-		if newWD != "" {
-			options = append(options, libpod.WithCreateWorkingDir())
-		}
+		options = append(options, libpod.WithCreateWorkingDir())
 	}
 	if s.StopSignal != nil {
 		options = append(options, libpod.WithStopSignal(*s.StopSignal))
--- a/pkg/specgen/container_validate.go
+++ b/pkg/specgen/container_validate.go
@@ -142,11 +142,6 @@ func (s *SpecGenerator) Validate() error {
 		return err
 	}
 
-	// The following are defaults as needed by container creation
-	if len(s.WorkDir) < 1 {
-		s.WorkDir = "/"
-	}
-
 	// Set defaults if network info is not provided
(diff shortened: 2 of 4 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests

rhatdan:
Fixes containers/buildah#2402

pkg/specgen/generate/container_create.go
```diff
@@ -245,9 +245,12 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
...
 		}
-		if newWD != "" {
-			options = append(options, libpod.WithCreateWorkingDir())
-		}
+		s.WorkDir = newWD
+	}
+	if s.WorkDir == "" {
+		s.WorkDir = "/"
+	} else {
+		options = append(options, libpod.WithCreateWorkingDir())
```
mheon, line 253: I thought we decided to only do this for entrypoint specified by the image - if the user specifies `/ect` accidentally we would still error on that?

pkg/specgen/generate/container_create.go
```diff
@@ -245,9 +245,12 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
 		if err != nil {
 			return nil, err
 		}
-		if newWD != "" {
-			options = append(options, libpod.WithCreateWorkingDir())
-		}
+		s.WorkDir = newWD
```
mheon, line 248: I don't think we should be modifying the specgen here - maybe in CreateContainer, not createContainerOptions?

pkg/specgen/generate/container_create.go
rhatdan: Ok then I need to remove the test.

pkg/specgen/generate/container_create.go
rhatdan: The problem is WithCreateWorkingDir works off of that field.  Or we have to do the check twice.

rhatdan: Ok reworked it placing the WorkDir check into CreateContainer.

rhatdan:
Mitigates: https://github.com/containers/buildah/issues/2475

rhatdan:
@ashley-cui FYI

rhatdan:
@mheon Turns out runc will create workingdir on the image, if it does not exist, while crun will not.  Changed test to only run on crun systems.

mheon:
Yeah, it's a lovely difference.

PR #7245, opened 2020-08-06
enable use of iidfile for remote build

baude:
podman/buildah build allows for --iidfile which stores the id of the newly built image.  the api more or less was able to handle this but it wasnt being sent by the go bindings.  it also was not documented on the libpod side of the documentation because of a copy and paste with compat.

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -80,6 +80,7 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		Target      string   `schema:"target"`
 		Outputs     string   `schema:"outputs"`
 		Registry    string   `schema:"registry"`
+		IIDFile     string   `schema:"iidfile"`
 	}{
 		Dockerfile:  "Dockerfile",
 		Tag:         []string{},
@@ -112,7 +113,6 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest, err)
 		return
 	}
-
 	var (
 		output          string
 		additionalNames []string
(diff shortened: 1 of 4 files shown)


baude:
@rhatdan @mheon @jwhonce @vrothberg PTAL and lets decide on what level API change this is ... if at all.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

pkg/api/handlers/compat/images_build.go
```diff
@@ -80,6 +80,7 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		Target      string   `schema:"target"`
 		Outputs     string   `schema:"outputs"`
 		Registry    string   `schema:"registry"`
+		IIDFile     string   `schema:"iidfile"`
```
mheon, line 83: Why are we passing this through the API? This seems like it ought to be done by the frontend?

mheon: (Or are we supposed to be creating the ID file on the server? That seems strange...)

vrothberg: I would also expect this to be handled entirely by the client rather than the server.

baude: it cannot, as implemented, be handled by the client.  buildah does the writing of this to file.  it is hardwired.

rhatdan: Doesn't buildah hand back to podman build the ID Can't it send the ID back to the client side so the id can be written to a file?

rhatdan: Is this now having the client write out the IIDfile?

vrothberg: Not yet.

I have a hard time imaging scenarios where the iidfile could be useful on the server. I am usually using such files for systemd services or scripts, but all of them would run on the client.
```bash
$ podman-remote build --iidfile=/tmp/foo ...
$ podman-remote run `cat /tmp/foo`
```
... this would work only if the service and client are on the same machine.

rhatdan: If this does not work on my MAC with the server running on a different machine, then this is not fixed.

vrothberg:
@baude, still on it?

rhatdan:
Since @baude is out, I think someone else should take this over.  This is a fairly critical one in my mind, might allow us to enable more tests also.

baude:
closing in favor of #7679

Issue #7271, opened 2020-08-10
generate systemd invalid memory address panic


jtligon:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I've managed to get a container up and running with a device passed in. Now I'm trying to generate a systemd unit so thiat it will start the container when I power cycle to box. running `podman generate systemd ...` gives me the error pasted below.

**Steps to reproduce the issue:**

1. Start a container named motioneye

2. `sudo podman generate systemd motioneye -f --restart-policy restart-always`

3. boom

**Describe the results you received:**

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x56382de2c04d]

goroutine 1 [running]:
panic(0x56382e3f4a60, 0x56382f3f3750)
	/usr/lib/golang/src/runtime/panic.go:1064 +0x471 fp=0xc000595b98 sp=0xc000595ae0 pc=0x56382cc52301
runtime.panicmem(...)
	/usr/lib/golang/src/runtime/panic.go:212
runtime.sigpanic()
	/usr/lib/golang/src/runtime/signal_unix.go:695 +0x3de fp=0xc000595bc8 sp=0xc000595b98 pc=0x56382cc6964e
github.com/containers/libpod/cmd/podman/generate.systemd(0x56382f411c60, 0xc0001b7e40, 0x1, 0x4, 0x0, 0x0)
	/builddir/build/BUILD/podman-2.0.4/_build/src/github.com/containers/libpod/cmd/podman/generate/systemd.go:58 +0x15d fp=0xc000595c88 sp=0xc000595bc8 pc=0x56382de2c04d
github.com/containers/libpod/vendor/github.com/spf13/cobra.(*Command).execute(0x56382f411c60, 0xc0001ae030, 0x4, 0x4, 0x56382f411c60, 0xc0001ae030)
	/builddir/build/BUILD/podman-2.0.4/_build/src/github.com/containers/libpod/vendor/github.com/spf13/cobra/command.go:838 +0x455 fp=0xc000595d60 sp=0xc000595c88 pc=0x56382d647455
github.com/containers/libpod/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x56382f420360, 0xc0001a8020, 0x56382e3f1a60, 0x56382f4c9388)
[...]

PR #7272, opened 2020-08-10
generate systemd: fix error handling

vrothberg:
Fix a bug in the error handling which returned nil instead of an error
and ultimately lead to nil dereferences in the client.  To prevent
future regressions, add a test and check for the error message.

Fixes: #7271

### DIFF
--- a/test/e2e/generate_systemd_test.go
+++ b/test/e2e/generate_systemd_test.go
@@ -53,6 +53,18 @@ var _ = Describe("Podman generate systemd", func() {
 		Expect(session).To(ExitWithError())
 	})
 
+	It("podman generate systemd bad restart-policy value", func() {
+		session := podmanTest.Podman([]string{"create", "--name", "foobar", "alpine", "top"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"generate", "systemd", "--restart-policy", "bogus", "foobar"})
+		session.WaitWithDefaultTimeout()
+		Expect(session).To(ExitWithError())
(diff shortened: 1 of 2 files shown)


test/e2e/generate_systemd_test.go
```diff
@@ -53,6 +53,18 @@ var _ = Describe("Podman generate systemd", func() {
...
 
+	It("podman generate systemd bad restart-policy value", func() {
+		session := podmanTest.Podman([]string{"create", "--name", "foobar", "alpine", "top"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"generate", "systemd", "--restart-policy", "bogus", "foobar"})
+		session.WaitWithDefaultTimeout()
+		Expect(session).To(ExitWithError())
+		found, _ := session.ErrorGrepString("Error: bogus is not a valid restart policy")
```
jtligon, line 64: shouldn't bogus be arg[3]?

vrothberg: Do you want to change the error message to point to the exact argument?

jtligon: well, the bad value passed in? unless thats a bad idea? arg[4]?

vrothberg: The passed value is part of the error message already. It's "Error: $bad-value ...".

rhatdan:
LGTM

test/e2e/generate_systemd_test.go
jtligon: gotcha, I'm just bad at reading the code. I saw the string on 64 and thought it was hard coded to bogus. I see its part of the return of ExitWithError(), my bad 😊 nvm

PR #7274, opened 2020-08-10
In podman 1.* regression on --cap-add

rhatdan:
In podman 1.0 if  you executed a command like:

podman run --user dwalsh --cap-add net_bind_service alpine nc -l 80

It would work, and the user dwalsh would get the capability,  in
podman 2.0, only root and the binding set gets the capability.

This change restores us back to the way podman 1.0 worked.


rhatdan:
runtime-tools vendoring is currently broken upstream.

vrothberg:
> runtime-tools vendoring is currently broken upstream.

The linter is complaining but I wonder why the linter is running on the vendored packages at all (parsing, type checks?)

pkg/specgen/generate/security.go
```diff
@@ -144,10 +144,14 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 		configSpec.Process.Capabilities.Inheritable = caplist
 		configSpec.Process.Capabilities.Ambient = caplist
 	} else {
-		configSpec.Process.Capabilities.Effective = []string{}
+		userCaps, err := capabilities.NormalizeCapabilities(s.CapAdd)
```
rhatdan, line 147: @giuseppe PTAL

rhatdan:
@mheon @vrothberg @baude @jwhonce @ashley-cui @QiWang19 PTAL

TomSweeneyRedHat:
Tests aren't at all happy.
Changes LGTM, but I'd like a head nod from @giuseppe before merging.

pkg/specgen/generate/security.go
```diff
@@ -144,10 +144,14 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
...
+		if err != nil {
+			return errors.Wrapf(err, "capabilities requested by user are not valid: %q", strings.Join(s.CapAdd, ","))
+		}
+		configSpec.Process.Capabilities.Effective = userCaps
 		configSpec.Process.Capabilities.Permitted = []string{}
 		configSpec.Process.Capabilities.Inheritable = []string{}
```
mheon, line 153: Should we be setting this too? It seems strange that they can't be inherited if we're giving them to the first process (esp. considering we set all 4 about for root)

rhatdan: Yes I am not sure if we should set all 4 or just effective, or any combination.

docs/source/markdown/podman-build.1.md
```diff
@@ -736,6 +745,52 @@ $ podman build -f dev/Containerfile https://10.10.10.1/podman/context.tar.gz
...
+*.doc
+!Help.doc
+```
+
+Exclude all doc files except Help.doc from the image.
+
+This functionality is compatible with the handling of .dockerignore files described here:
+
+https://docs.docker.com/engine/reference/builder/#dockerignore-file
+
```
giuseppe, line 793: same here

docs/source/markdown/podman-build.1.md
```diff
@@ -752,6 +807,8 @@ If you are using `useradd` within your build script, you should pass the `--no-l
 podman(1), buildah(1), containers-registries.conf(5), crun(8), runc(8), useradd(8)
 
 ## HISTORY
+Aug 2020, Additiona options and .dockerignore added by Dan Walsh <dwalsh@redhat.com>
+
```
giuseppe, line 811: seems unrelated.  And typo in `Additiona` :-)

pkg/specgen/generate/security.go
```diff
@@ -144,11 +138,22 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 		configSpec.Process.Capabilities.Inheritable = caplist
 		configSpec.Process.Capabilities.Ambient = caplist
```
giuseppe, line 139: not related to the patch, but why do we set Ambient capabilities for root containers?

docs/source/markdown/podman-build.1.md
```diff
@@ -351,6 +351,15 @@ another process.
...
 
+**--omit-timestamp** *bool-value*
+
+Set the create timestamp to epoch 0 to allow for deterministic builds (defaults to false).
+By default, the created timestamp is changed and written into the image manifest with every commit,
+causing the image's sha256 hash to be different even if the sources are exactly the same otherwise.
+When --omit-timestamp is set to true, the created timestamp is always set to the epoch and therefore not
+changed, allowing the image's sha256 to remain the same. All files committed to the layers of the image
+will get the epoch 0 timestamp.
+
```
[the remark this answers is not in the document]
rhatdan: Yes it is related to the update vendor of buildah. I will split this out into a separate PR to make it clearer.  Without this documentation the update will fail.

docs/source/markdown/podman-build.1.md
rhatdan: Answered above.

pkg/specgen/generate/security.go
rhatdan: I am not quite sure how to handle these different caps, But logging in as root, I see that capamb and capinh by default are all 0's so we probably should not touch them in either case.

PR #7281, opened 2020-08-10
Ensure correct propagation for cgroupsv1 systemd cgroup

mheon:
On cgroups v1 systems, we need to mount /sys/fs/cgroup/systemd into the container. We were doing this with no explicit mount propagation tag, which means that, under some circumstances, the shared mount propagation could be chosen - which, combined with the fact that we need a mount to mask /sys/fs/cgroup/systemd/release_agent in the container, means we would leak a never-ending set of mounts under /sys/fs/cgroup/systemd/ on container restart.

Fortunately, the fix is very simple - hardcode mount propagation to something that won't leak.


rhatdan:
LGTM
@vrothberg @baude @edsantiago @QiWang19 @vrothberg @jwhonce @giuseppe @TomSweeneyRedHat PTAL

mheon:
Rebased

Issue #6721, opened 2020-06-23
podman does not support both tag and digest for pulling images


Jamstah:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Docker will accept a tag and a digest to pull an image. It will pull that specific digest only, and completely ignore the tag.

**Steps to reproduce the issue:**

```
[jammy@ibm007470 ~]$ skopeo inspect docker://docker.io/fedora:32 | jq .Digest
"sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326"
[jammy@ibm007470 ~]$ podman pull docker.io/fedora:32@sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326
Error: invalid image reference "docker.io/fedora:32@sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326"
[jammy@ibm007470 ~]$ docker pull docker.io/fedora:32@sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326
sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326: Pulling from library/fedora
0169c1449c16: Downloading  4.832MB/70.73MB
```

**Describe the results you received:**
Podman does not pull

**Describe the results you expected:**
Podman pulls successfully

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
[jammy@ibm007470 ~]$ podman version
Version:            1.9.3
RemoteAPI Version:  1
Go Version:         go1.14.2
OS/Arch:            linux/amd64
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-1.9.3-1.fc32.x86_64
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Laptop

vrothberg:
Thanks for opening the issue, @Jamstah!

@mtrmac PTAL

mtrmac:
(Just linking earlier discussions:) The problem space has been described in https://github.com/containers/buildah/issues/1407#issuecomment-472575647 . An implementation https://github.com/cri-o/cri-o/pull/3060 , with the PR review pointing out gotchas with simpler implementations.

rhatdan:
@mtrmac still working on this?

ParkerVR:
[...]

PR #7282, opened 2020-08-10
Img Pull - Error pass through

ParkerVR:
Did some digging on https://github.com/containers/podman/issues/6721
Realized it was a feature not yet supported rather than a bug, changed error to pass through higher error for better diagnosis

Example output due to change:

    Original
 ➜  $ podman pull docker.io/fedora:32@sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326
Error: invalid image reference "docker.io/fedora:32@sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326"

    Modified
➜  #$podman pull docker.io/fedora:32@sha256:e69b5a62ce23c673885bddc94e6679c9b2af683059637ceddb9cff458537a326
Error: Docker references with both a tag and digest are currently not supported

### DIFF
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -226,7 +226,7 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
 	if err != nil {
 		imageRef, err = alltransports.ParseImageName(fmt.Sprintf("%s%s", dockerPrefix, rawImage))
 		if err != nil {
-			return nil, errors.Errorf("invalid image reference %q", rawImage)
+			return nil, errors.Wrapf(err, "invalid image reference %q", rawImage)
 		}
 	}
 


rhatdan:
LGTM
@mheon @vrothberg @baude @giuseppe @jwhonce @ashley-cui @QiWang19 PTAL

pkg/domain/infra/abi/images.go
```diff
@@ -226,7 +226,7 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
 	if err != nil {
 		imageRef, err = alltransports.ParseImageName(fmt.Sprintf("%s%s", dockerPrefix, rawImage))
 		if err != nil {
-			return nil, errors.Errorf("invalid image reference %q", rawImage)
```
mheon, line 229: Does this produce really verbose errors?

If they get too gnarly, we may want to make this only happen when `--log-level=debug` is passed, and we return the previous output otherwise.

pkg/domain/infra/abi/images.go
```diff
@@ -226,7 +226,7 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
...
+			return nil, err
```
QiWang19, line 229: maybe use errrors.Wrapf?
`return nil, errors.Wrapf(err, "invalid image reference %q", rawImage)`

pkg/domain/infra/abi/images.go
ParkerVR: Not particularly, here are some examples:

`Invalid image name "%s", expected colon-separated transport:reference`, imgName
`Invalid image name "%s", unknown transport "%s"`, imgName, parts[0]

mheon: Alright. I think we should include Qi's suggestion about wrapping, but once that's done LGTM

QiWang19:
LGTM

Issue #7103, opened 2020-07-28
podman pod stop and podman pod start don't reopen ports to containers


fp7:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
After `podman pod restart <name>` the containers will be still accessible from the host.
After `podman pod stop <name>; podman pod start <name>` they are not

**Steps to reproduce the issue:**

1. `podman pod create --name test -p "9200:9200"`

2. `podman run -d --pod test  -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.6.2`

3. `podman pod stop test`

3. `podman pod start test`

4. `curl localhost:9200/_cat`

**Describe the results you received:**
I get `curl: (7) Failed to connect to localhost port 9200: Connection refused`

**Describe the results you expected:**
I would expect the port to be reachable

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.3
API Version:  1
Go Version:   go1.14.4
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.18, commit: '
  cpus: 8
  distribution:
    distribution: ubuntu
    version: "18.04"
  eventLogger: file
  hostname: tp
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 4.15.0-112-generic
  linkmode: dynamic
  memFree: 16008376320
  memTotal: 41937190912
  ociRuntime:
    name: runc
    package: 'containerd.io: /usr/bin/runc'
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc10
      commit: dc9208a3303feef5b3839f4323d9beb36df0a9dd
      spec: 1.0.1-dev
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
[...]

PR #7283, opened 2020-08-10
Ensure pod infra containers have an exit command

mheon:
Most Libpod containers are made via `pkg/specgen/generate` which includes code to generate an appropriate exit command which will handle unmounting the container's storage, cleaning up the container's network, etc. There is one notable exception: pod infra containers, which are made entirely within Libpod and do not touch pkg/specgen. As such, no cleanup process, network never cleaned up, bad things can happen.

There is good news, though - it's not that difficult to add this, and it's done in this PR. Generally speaking, we don't allow passing options directly to the infra container at create time, but we do (optionally) proxy a pre-approved set of options into it when we create it. Add ExitCommand to these options, and set it at time of pod creation using the same code we use to generate exit commands for normal containers.

### DIFF
--- a/libpod/pod_api.go
+++ b/libpod/pod_api.go
@@ -11,20 +11,20 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// Start starts all containers within a pod
-// It combines the effects of Init() and Start() on a container
+// Start starts all containers within a pod.
+// It combines the effects of Init() and Start() on a container.
 // If a container has already been initialized it will be started,
 // otherwise it will be initialized then started.
 // Containers that are already running or have been paused are ignored
 // All containers are started independently, in order dictated by their
 // dependencies.
-// An error and a map[string]error are returned
+// An error and a map[string]error are returned.
 // If the error is not nil and the map is nil, an error was encountered before
-// any containers were started
+// any containers were started.
 // If map is not nil, an error was encountered when starting one or more
 // containers. The container ID is mapped to the error encountered. The error is
-// set to ErrCtrExists
-// If both error and the map are nil, all containers were started successfully
+// set to ErrPodPartialFail.
+// If both error and the map are nil, all containers were started successfully.
 func (p *Pod) Start(ctx context.Context) (map[string]error, error) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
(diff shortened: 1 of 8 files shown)


TomSweeneyRedHat:
Anyway to add a test?

ashley-cui:
LGTM, tested and it fixes #7103


```
[acui@localhost podman]$ ./bin/podman pod create --name test -p 5000:5000
564f217152dcceccc8f8200d8d4060f1b7602e64db104d07b8d3ee2dbc9ccb22

[acui@localhost podman]$ ./bin/podman create -d --pod test -p 5000:5000 registry:2
9c4d770b0cefd11b112ca13e415a2cd231fe1cb1866754fbc4d6eaa204902a42

[acui@localhost podman]$ ./bin/podman pod start test
564f217152dcceccc8f8200d8d4060f1b7602e64db104d07b8d3ee2dbc9ccb22

[acui@localhost podman]$ curl localhost:5000

[acui@localhost podman]$ ./bin/podman pod stop test
564f217152dcceccc8f8200d8d4060f1b7602e64db104d07b8d3ee2dbc9ccb22

[acui@localhost podman]$ ./bin/podman pod start test
564f217152dcceccc8f8200d8d4060f1b7602e64db104d07b8d3ee2dbc9ccb22

[acui@localhost podman]$ curl localhost:5000
```

mheon:
Pushed an update to fix pods where no infra container was created.

mheon:
Rebased, hopefully CI will be happy now

rhatdan:
Sad trombone.

mheon:
I think the bindings test is wrong, will fix tomorrow morning. Not sure what's going on with rootless remote and F31 - they seem to be consistently failing, but initial glance doesn't say why.

mheon:
Bindings tests should be fixed

mheon:
Current thinking: we have a race similar to https://github.com/containers/podman/issues/5747 but with stopping pods, not containers - and it's causing the bindings tests to fail as we hit the pod for status info before the cleanup process can run.

rhatdan:
Can you add a podman wait?

mheon:
I'm looking into a similar solution to what we used in #5747 - force a cleanup before the API call returns.

mheon:
Added code to ensure cleanup happens before the Pod Stop API call finishes.

libpod/pod_api.go
```diff
@@ -138,10 +138,82 @@ func (p *Pod) StopWithTimeout(ctx context.Context, cleanup bool, timeout int) (m
...
+		ctr.lock.Lock()
+
+		if err := ctr.syncContainer(); err != nil {
+			ctr.lock.Unlock()
+			ctrErrors[ctr.ID()] = err
+			continue
+		}
+
+		// Ignore containers that are running/paused
+		if !ctr.ensureState(define.ContainerStateConfigured, define.ContainerStateCreated, define.ContainerStateStopped, define.ContainerStateExited) {
```
TomSweeneyRedHat, line 185: Not for this review, but it might be nice to create a `isContainerPausedOrRunning()` utility function.  I think we've this big blob of test in a few places.  But not for this PR.

TomSweeneyRedHat:
LGTM
one comment for future consideration and happy green test buttons

mheon:
Tests are green!

@rhatdan @baude @giuseppe @TomSweeneyRedHat PTAL and merge, would like to get this backported to 2.0 for 2.0.5

PR #7289, opened 2020-08-11
V2 backports

vrothberg:
Backport a number of fixes:

 * Fix close fds of exec --preserve-fds
 * fix pod creation with "new:" syntax
 * Fix podman service --valink timeout
 * Add versioned _ping endpoint
 * Change recommended systemd unit path for root.
 * API returns 500 in case network is not found instead of 404
 * podman.service: drop install section
 * Handle podman-remote run --rm
 * correct go-binding key for volumes
 * Reenable remote system tests
 * implement the exitcode when start a container with attach
 * Do not set host IP on ports when 0.0.0.0 requested
 * Missing return after early exit
 * docker-compose uses application/tar
 * rootless: system service joins immediately the namespaces
 * fix bug podman sign storage path
 * podman-remote send name and tag
 * Ensure that exec errors write exit codes to the DB
 * fix podman logs --tail when log is bigger than pagesize
 * image list: speed up
 * generate systemd: fix error handling

TODO:
* @mheon, I did not backport https://github.com/containers/podman/commit/333d9af77a65d860b5fd788805cd4c0f2cd232df yet as it's based on the config refactoring. I let you decide if we should pull in the refactor as well or fix it in the old code base. Once done, we need to get @rhatdan's https://github.com/containers/podman/pull/7239 in as well which depends on some of the code.


vrothberg:
@baude @giuseppe @jwhonce @mheon @QiWang19 @rhatdan @zhangguanzhang 

PTAL ... went through the log and picked up commits for 2.0

mheon:
I'm a little iffy about the userns keep-id makes changes to `/etc/passwd` patches - those felt feature-esque to me so I didn't backport them. But if you don't mind including them despite this, I'm OK with it.

vrothberg:
> I'm a little iffy about the userns keep-id makes changes to `/etc/passwd` patches - those felt feature-esque to me so I didn't backport them. But if you don't mind including them despite this, I'm OK with it.

@rhatdan WDYT?

zhangguanzhang:
I'm OK with it.

TomSweeneyRedHat:
I'd lean towards not including the user-id that @mheon questioned  too.  However, I'm OK if it goes in if that's the decision.  Otherwise:
LGTM

vrothberg:
I also wasn't 100 percent sure about the passwd nature but as https://github.com/containers/podman/issues/6953 was marked as a bug, I figure to pull that in. Maybe it was a bug by an earlier feature commit though.

rhatdan:
/lgtm

I am fine with adding them.

PR #7290, opened 2020-08-11
Show c/storage (Buildah/CRI-O) containers in ps

rhatdan:
The `podman ps --all` command will now show containers that
are under the control of other c/storage container systems and
the new `ps --external` option will show only containers that are
in c/storage but are not controlled by libpod.

In the below examples, the '*working-container' entries were created
by Buildah.

```
podman ps -a
CONTAINER ID  IMAGE                             COMMAND  CREATED       STATUS                   PORTS  NAMES
9257ef8c786c  docker.io/library/busybox:latest  ls /etc  8 hours ago   Exited (0) 8 hours ago          gifted_jang
d302c81856da  docker.io/library/busybox:latest  buildah  30 hours ago  external                        busybox-working-container
7a5a7b099d33  localhost/tom:latest              ls -alF  30 hours ago  Exited (0) 30 hours ago         hopeful_hellman
01d601fca090  localhost/tom:latest              ls -alf  30 hours ago  Exited (1) 30 hours ago         determined_panini
ee58f429ff26  localhost/tom:latest              buildah  33 hours ago  external                        alpine-working-container

podman ps --external
CONTAINER ID  IMAGE                             COMMAND  CREATED       STATUS    PORTS  NAMES
d302c81856da  docker.io/library/busybox:latest  buildah  30 hours ago  external         busybox-working-container
ee58f429ff26  localhost/tom:latest              buildah  33 hours ago  external         alpine-working-container

```

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -13,6 +13,7 @@ import (
 	tm "github.com/buger/goterm"
 	"github.com/containers/buildah/pkg/formats"
 	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/cmd/podman/utils"
 	"github.com/containers/podman/v2/cmd/podman/validate"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/cri-o/ocicni/pkg/ocicni"
@@ -56,9 +57,9 @@ func init() {
 func listFlagSet(flags *pflag.FlagSet) {
 	flags.BoolVarP(&listOpts.All, "all", "a", false, "Show all the containers, default is only running containers")
 	flags.StringSliceVarP(&filters, "filter", "f", []string{}, "Filter output based on conditions given")
+	flags.BoolVar(&listOpts.Storage, "storage", false, "Show containers in storage not controlled by Podman")
 	flags.StringVar(&listOpts.Format, "format", "", "Pretty-print containers to JSON or using a Go template")
 	flags.IntVarP(&listOpts.Last, "last", "n", -1, "Print the n last created containers (all states)")
-	flags.BoolVar(&listOpts.Namespace, "namespace", false, "Display namespace information")
 	flags.BoolVar(&listOpts.Namespace, "ns", false, "Display namespace information")
 	flags.BoolVar(&noTrunc, "no-trunc", false, "Display the extended information")
 	flags.BoolVarP(&listOpts.Pod, "pod", "p", false, "Print the ID and name of the pod the containers are associated with")
@@ -69,6 +70,7 @@ func listFlagSet(flags *pflag.FlagSet) {
 
 	sort := validate.Value(&listOpts.Sort, "command", "created", "id", "image", "names", "runningfor", "size", "status")
 	flags.Var(sort, "sort", "Sort output by: "+sort.Choices())
+	flags.SetNormalizeFunc(utils.AliasFlags)
 }
 func checkFlags(c *cobra.Command) error {
 	// latest, and last are mutually exclusive.
@@ -102,6 +104,14 @@ func checkFlags(c *cobra.Command) error {
 	if listOpts.Watch > 0 && listOpts.Latest {
 		return errors.New("the watch and latest flags cannot be used together")
 	}
+	cfg := registry.PodmanConfig()
+	if cfg.Engine.Namespace != "" {
+		if c.Flag("storage").Changed && listOpts.Storage {
+			return errors.New("--namespace and --storage flags can not both be set")
+		}
+		listOpts.Storage = false
+	}
+
 	return nil
 }
 
--- a/docs/source/markdown/podman-ps.1.md
+++ b/docs/source/markdown/podman-ps.1.md
@@ -32,11 +32,17 @@ all the containers information.  By default it lists:
 
 **--all**, **-a**
 
-Show all the containers, default is only running containers
+Show all the containers created by Podman, default is only running containers.
+
+Note: Podman shares containers storage with other tools such as Buildah and CRI-O. In some cases these `external` containers might also exist in the same storage. Use the `--storage` option to see these external containers. External containers show the 'storage' status.
 
 **--pod**, **-p**
 
 Display the pods the containers are associated with
+
+**--storage**
+
+Display external containers that are not controlled by Podman but are stored in containers storage.  These external containers are generally created via other container technology such as Buildah or CRI-O and may depend on the same container images that Podman is also using.  External containers are denoted with either a 'buildah' or 'storage' in the COMMAND and STATUS column of the ps output. Only used with the --all option.
 
 **--no-trunc**
 
@@ -174,11 +180,20 @@ CONTAINER ID  IMAGE                            COMMAND    CREATED        STATUS 
 
 ```
 
+```
+$ podman ps --storage -a
+CONTAINER ID  IMAGE                             COMMAND  CREATED      STATUS  PORTS  NAMES
+69ed779d8ef9f  redis:alpine  "redis-server"  25 hours ago  Created                   6379/tcp  k8s_container1_podsandbox1_redhat.test.crio_redhat-test-crio_1
+38a8a78596f9  docker.io/library/busybox:latest  buildah  2 hours ago  storage        busybox-working-container
+fd7b786b5c32  docker.io/library/alpine:latest   buildah  2 hours ago  storage        alpine-working-container
+f78620804e00  scratch                           buildah  2 hours ago  storage        working-container
+```
+
 ## ps
 Print a list of containers
 
 ## SEE ALSO
-podman(1)
+podman(1), buildah(1), crio(8)
 
 ## HISTORY
 August 2017, Originally compiled by Urvashi Mohnani <umohnani@redhat.com>
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -8,11 +8,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/buildah"
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/libpod/events"
 	"github.com/containers/podman/v2/pkg/cgroups"
 	"github.com/containers/podman/v2/pkg/rootless"
+	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/docker/go-units"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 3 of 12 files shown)


cmd/podman/containers/ps.go
```diff
@@ -56,10 +57,10 @@ func init() {
 func listFlagSet(flags *pflag.FlagSet) {
 	flags.BoolVarP(&listOpts.All, "all", "a", false, "Show all the containers, default is only running containers")
 	flags.StringSliceVarP(&filters, "filter", "f", []string{}, "Filter output based on conditions given")
+	flags.BoolVar(&listOpts.External, "external", true, "Show containers in storage not controlled by Podman")
```
mheon, line 60: Think we should use `--storage` here for consistency with `podman rm --storage`

rhatdan: I am fine with this.

libpod/runtime_ctr.go
```diff
@@ -905,3 +906,67 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
 	}
 	return prunedContainers, pruneErrors, nil
 }
+
+// StorageContainers returns a list of containers from containers/storage that
+// are not currently known to Podman.  The list of containers passed in are all
+// of the containers known to Podman from the GetContainers() function.
+func (r *Runtime) StorageContainers() ([]*Container, error) {
```
mheon, line 913: We can't make this return Libpod containers - having a bunch of semi-functional things wandering around calling themselves a libpod.Container is asking for trouble (they don't have valid locks, for example, we'll panic the moment someone tries to call a real Libpod function on one of them). I think we need a separate API for this, that returns something like []*StorageContainer - we can translate those into the PS container struct in pkg/domain, so the code in PS will remain unchanged.

docs/source/markdown/podman-ps.1.md
```diff
@@ -32,12 +32,19 @@ all the containers information.  By default it lists:
...
 
 **--pod**, **-p**
 
 Display the pods the containers are associated with
 
+**--storage**
+
+Display containers that are not controlled by Podman but are stored in containers storage.  These containers
+are generally created via other container technology such as Buildah or CRI-O and may depend on the
+same container images that Podman is using.  These containers are denoted with a 'NA' in the COMMAND and
```
TomSweeneyRedHat, line 45: Does the 'NA' still apply?  Your example shows 'Buildah`

docs/source/markdown/podman-ps.1.md
```diff
@@ -32,12 +32,19 @@ all the containers information.  By default it lists:
 
 **--all**, **-a**
 
-Show all the containers, default is only running containers
+Show all the containers including those created by other container technologies such as Buildah and CRI-O.  The default displays only running containers created by Podman.  Containers not created by Podman have an 'NA' status.
```
TomSweeneyRedHat, line 35: ditto 'NA' as to below.

docs/source/markdown/podman-ps.1.md
rhatdan: Nope, good catch.

libpod/runtime_ctr.go
```diff
@@ -905,3 +907,32 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
...
+
+// StorageContainers returns a list of containers from containers/storage that
+// are not currently known to Podman.  The list of containers passed in are all
+// of the containers known to Podman from the GetContainers() function.
+func (r *Runtime) StorageContainers() ([]storage.Container, error) {
+	var retCtrs []storage.Container
+
+	// We only have a store when doing all, storage or size, so skip this otherwise.
+	if r.store == nil {
+		return retCtrs, nil
```
mheon, line 919: This should probably be an error

libpod/runtime_ctr.go
```diff
@@ -905,3 +907,32 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
...
+	}
+
+	storeContainers, err := r.store.Containers()
+	if err != nil {
+		return retCtrs, errors.Wrapf(err, "error reading list of all containers")
+	}
+	for _, container := range storeContainers {
+		if _, err := r.state.Container(container.ID); err == nil {
```
mheon, line 927: Can use `HasContainer` instead - is a lot faster

libpod/runtime_ctr.go
rhatdan: Are you sure?  All of the other checks ignore this issue and just return success?

mheon: Having no store is something that we deliberately did ourselves - we should never be calling this in a case where the store is not set up, so returning an error seems correct

TomSweeneyRedHat:
@rhatdan gating issue, looks like a man page needs tweaking.

rhatdan:
Tests are green.
@mheon @giuseppe @vrothberg @TomSweeneyRedHat @QiWang19 @ashley-cui @baude @jwhonce PTAL

QiWang19:
LGTM

rhatdan:
Tests are still green.  Can I get a review on this one?
@mheon @giuseppe @vrothberg @TomSweeneyRedHat @QiWang19 @baude @jwhonce PTAL

docs/source/markdown/podman-ps.1.md
```diff
@@ -32,12 +32,18 @@ all the containers information.  By default it lists:
 
 **--all**, **-a**
 
-Show all the containers, default is only running containers
+Show all the containers including those created by other container technologies such as Buildah and CRI-O.  The default displays only running containers maintained by Podman.  External containers will have a 'storage' status.
```
edsantiago, line 35: I'd really like this to read " including non-running ones and those created..." but it is really not worth another CI run for that. Maybe a followup CI docs pr.

edsantiago:
/hold

Possible bug. Will report in full in a few minutes, but I need time to investigate. Please do not merge yet.

cmd/podman/containers/ps.go
```diff
@@ -56,9 +57,9 @@ func init() {
 func listFlagSet(flags *pflag.FlagSet) {
 	flags.BoolVarP(&listOpts.All, "all", "a", false, "Show all the containers, default is only running containers")
 	flags.StringSliceVarP(&filters, "filter", "f", []string{}, "Filter output based on conditions given")
+	flags.BoolVar(&listOpts.Storage, "storage", true, "Show containers in storage not controlled by Podman")
```
edsantiago, line 60: Should default to false

pkg/ps/ps.go
```diff
@@ -68,7 +69,31 @@ func GetContainerLists(runtime *libpod.Runtime, options entities.ContainerListOp
 			return nil, err
 		}
 		pss = append(pss, listCon)
+	}
+
+	if options.All && options.Storage {
```
edsantiago, line 74: Should be `||`, not `&&`'

mheon:
I'd prefer it if we took the all flag out of the picture entirely, and only
gated this behind the storage flag.

edsantiago:
>I'd prefer it if we took the all flag out of the picture entirely

I was ambivalent at first -- it does significantly change the behavior of `ps -a` -- but the more I think about it the more I really want it in `ps -a`.

Disturbingly often I find myself in annoying situations where `podman rm` and `rmi` start complaining at me about "in use by a container" or something like that. `podman ps -a` shows nothing. I've trained myself to run `buildah containers` and `buildah rm`, and that usually fixes things (typically leftovers from a failed `podman build`). I think normal humans might not knoow to do this, and would appreciate `ps -a` showing something useful.

This still leaves the problem of removing those containers. `podman rm` and `rm --storage` aren't actually very helpful here. But we can address that later.

mheon:
I would really prefer the opposite, because I don't want people to see these containers and then try using them with other Podman commands. The output of `ps` may not be as complete as it should be, but I strongly suspect people will not be able to identify what is a storage container and what is a Podman container on first glance, and they will try and interact with these containers using the standard Podman commands (`podman start`, `podman rm`, `podman inspect`, etc) and will get a series of wonderfully contradictory "that container does not exist" errors. We need to make this an explicit thing people opt into to prevent normal operation of Podman from becoming horribly confusing.

This is really a no-win scenario - we can either show them and confuse people as to why they don't act like normal Podman containers, or not show them and confuse people as to why they see errors that their image is "in use".

edsantiago:
That's a good point. I defer to you in this matter. And I agree that it's a no-win.

edsantiago:
Proposed test: run a long build with a short timeout:
```bash
@test "podman ps --storage" {
    PODMAN_TIMEOUT=5 run_podman 124 build -t thiswillneverexist - <<EOF
FROM $IMAGE
RUN sleep 15
EOF

    run_podman ps --storage
    is "${lines[1]}" \
       "[0-9a-f]\{12\} \+$IMAGE *buildah .* seconds ago .* storage .* ${PODMAN_TEST_IMAGE_NAME}-working-container" \
       "podman ps --storage"

    run_podman rm --storage -f "${lines[1]:0:12}"
}
```
Requires minor surgery to `helpers.bash`, so please don't include in your PR. I'm commenting here just to see if anyone has better ideas on how to test. I don't like this because it adds 15 seconds to the already-long system test run time. Any other ideas?

rhatdan:
We can discuss this next week, but I am preparing to make these containers first class within podman, IE allow podman ps -a to show them by default.  Allow `podman rm` to remove them by default.  Allow `podman diff` to show the difference by default (Something requested by CRI-O).

Users do not understand these and do not understand how to use them.  If we make --storage the default and properly show that they are not libpod containers, I believe the user experience will be better for the users.

```
$ ./bin/podman ps -a
CONTAINER ID  IMAGE   COMMAND  CREATED  STATUS  PORTS   NAMES
$ buildah from alpine
alpine-working-container
 $ ./bin/podman ps -a
CONTAINER ID  IMAGE                            COMMAND  CREATED        STATUS   PORTS   NAMES
84c690fc961c  docker.io/library/alpine:latest  buildah  2 seconds ago  storage          alpine-working-container
$ ./bin/podman rm alpine-working-container
alpine-working-container
```

These containers are unknown by libpod but Podman should be able to deal with them.
We can allow the storage flag to be set in containers.conf to allow us to get back to the current behaviour if users and distros want, but I think that these containers not being usable by default is a User Interface failure.

TomSweeneyRedHat:
I second the thought of a post scrum chat on how ps should show the variety of containers next week, fwiw.

mheon:
Concur. I still have serious doubts (we can maybe get a few commands to work in an OK-ish way, but the vast majority of Podman will never work with these containers)

vrothberg:
Just a friendly ping to bump the issue up the mailboxes. Let's chat in tomorrows water cooler.

rhatdan:
@vrothberg @giuseppe @mheon @baude @jwhonce @ashley-cui @QiWang19 This is less controversial now since the --format is defaulted to false.

docs/source/markdown/podman-ps.1.md
mheon: Can we undo this change until we change the default?

rhatdan:
I added more documentation to explain what these are to the podman build man page and tried to standardize on the term `external` containers in all of the effected man pages.

mheon:
LGTM

mheon:
Restarted two tests (hopefully flakes)

vrothberg (commented):
/lgtm

PR #7291, opened 2020-08-11
[CI:DOCS] cross-reference .rst files

edsantiago:
There are a bunch of *.rst files in docs/source, linking sometimes
to man pages and sometimes to other .rst files. These files each
have entries of the following form:
   :doc:`foo <link-to-foo>` Description of foo

...for all podman sub and sub-subcommands 'foo'.

Read all .rst files and make sure that:

  - all entries in a given file are in alphabetical order
  - all link-to-foo targets point to existing doc files
  - every subcommand known by 'podman help' has a corresponding
    doc entry in a .rst file

### DIFF
--- a/docs/source/managecontainers.rst
+++ b/docs/source/managecontainers.rst
@@ -37,9 +37,9 @@ Manage Containers
 
 :doc:`port <markdown/podman-port.1>` List port mappings or a specific mapping for the container
 
-:doc:`restart <markdown/podman-restart.1>` Restart one or more containers
-
 :doc:`prune <markdown/podman-container-prune.1>` Remove all stopped containers
+
+:doc:`restart <markdown/podman-restart.1>` Restart one or more containers
 
 :doc:`restore <markdown/podman-container-restore.1>` Restores one or more containers from a checkpoint
 
--- a/docs/source/Commands.rst
+++ b/docs/source/Commands.rst
@@ -98,7 +98,7 @@ Commands
 
 :doc:`top <markdown/podman-top.1>` Display the running processes of a container
 
-:doc:`umount <markdown/podman-umount.1>` Unmounts working container's root filesystem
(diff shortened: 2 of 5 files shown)


edsantiago:
This is a two-commit PR, and I encourage you to review it as such.

The first commit is relatively minor: simply adds a sort-check to one of the man page tests. It's a little more complicated than desired, because I changed error status to a global. It makes sense for....

...The second commit, which is the one that checks `.rst` files. In this commit I also fixed the problems reported by the new script functionality.

docs/source/managecontainers.rst
```diff
@@ -37,10 +37,10 @@ Manage Containers
 
 :doc:`port <markdown/podman-port.1>` List port mappings or a specific mapping for the container
 
-:doc:`restart <markdown/podman-restart.1>` Restart one or more containers
-
 :doc:`prune <markdown/podman-container-prune.1>` Remove all stopped containers
 
+:doc:`restart <markdown/podman-restart.1>` Restart one or more containers
+
```
TomSweeneyRedHat, line 43: should line 40 and 44 be higher up?  'podman-container-*'

TomSweeneyRedHat:
Nice work @edsantiago.  I have a love/hate relationship with these kinds of checks.  I hate it when I mess up and they ding me, but I love it that they ding me because things are messed up.
The script part LGTM, but I think you'll unlease some ordering errors that will need further help.  Will add a LGTM once those are cleaned up.

docs/source/managecontainers.rst
edsantiago: I'm sorting on the subcommand name, not the link target. 'prune' < 'restart'

edsantiago: (does that address your concern?)

TomSweeneyRedHat: My bad, I was liking at the link targets there.

mheon:
LGTM
/approve

PR #7292, opened 2020-08-11
Wait for reexec to finish when fileOutput is nil

jdieter:
Currently, we're not cleaning up up after ourselves when `fileOutput` is nil. This patch fixes that.

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -233,6 +233,11 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (_ boo
 	if pid < 0 {
 		return false, -1, errors.Errorf("cannot re-exec process")
 	}
+	defer func() {
+		if retErr != nil {
+			C.reexec_in_user_namespace_wait(pidC, 0)
+		}
+	}()
 
 	uids, gids, err := GetConfiguredMappings()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/rootless/rootless_linux.go
```diff
@@ -294,6 +294,11 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (_ boo
 	}
 
 	if fileOutput != nil {
+		ret := C.reexec_in_user_namespace_wait(pidC, 0)
+		if ret < 0 {
+			return false, -1, errors.New("error waiting for the re-exec process")
+		}
+
```
TomSweeneyRedHat, line 301: IDK, but will ask.  We have a couple of other returns before the end of this function at lines 311 and 315.  Do we need to call there too?

jdieter: I think we probably should.  Can we `defer` a function that runs `reexec_in_user_namespace_wait`?  Would there be a problem if we ran it against the same PID in rapid succession?

TomSweeneyRedHat: @jdieter if we did the defer, we'd need to capture the status of the call and return appropriately based on that.  I'm not sure that's doable with defer, @mheon or @vrothberg might have some smarter thoughts on that or my question in general.

rhatdan:
@giuseppe PTAL

pkg/rootless/rootless_linux.go
giuseppe: we could probably record the exit error and run it only when retErr != nil.  What do you think?

rhatdan: Makes sense to me.

jdieter: Ok, I'm not very proficient in Go, but tell me if my lastest push has what you're looking for.  I've setup a deferred function that waits for the child if `retErr != nil`.  I don't bother with the return value from `reexec_in_user_namespace_wait` because we presumably are more interested in passing whatever error is causing us to bail.

rhatdan:
@giuseppe PTAL

PR #7293, opened 2020-08-11
Fix hang when `path` doesn't exist

jdieter:
I'm not sure if this is an OS-specific issue, but on CentOS 8, if `path` doesn't exist, this hangs while waiting to read from this socket, even though the socket is closed by the `reexec_in_user_namespace`.  Switching to a pipe fixes the problem, and pipes shouldn't be an issue since this is Linux-specific code.

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -389,13 +389,11 @@ func TryJoinFromFilePaths(pausePidPath string, needNewNamespace bool, paths []st
 			lastErr = nil
 			break
 		} else {
-			fds, err := unix.Socketpair(unix.AF_UNIX, unix.SOCK_DGRAM, 0)
+			r, w, err := os.Pipe()
 			if err != nil {
 				lastErr = err
 				continue
 			}
-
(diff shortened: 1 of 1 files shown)


pkg/rootless/rootless_linux.go
```diff
@@ -389,14 +389,11 @@ func TryJoinFromFilePaths(pausePidPath string, needNewNamespace bool, paths []st
 			lastErr = nil
 			break
 		} else {
-			fds, err := unix.Socketpair(unix.AF_UNIX, unix.SOCK_DGRAM, 0)
-			if err != nil {
+			if r, w, err := os.Pipe(); err != nil {
```
zhangguanzhang, line 392:
```suggestion
			r, w, err := os.Pipe()
			if err != nil {
```

zhangguanzhang:
you need rebase your commit count from 2 to 1

mheon:
@giuseppe PTAL

rhatdan:
LGTM
But need @giuseppe approval.

jdieter:
Thanks so much!

PR #7295, opened 2020-08-11
podman-remote fixes for msi and client

baude:
correct small typo that sets the path on windows via the msi xml.

in the remote client, prompt for SSH password when no identity or alternate means of authentication are provided.

### DIFF
--- a/cmd/podman/system/connection/add.go
+++ b/cmd/podman/system/connection/add.go
@@ -124,6 +124,7 @@ func add(cmd *cobra.Command, args []string) error {
 		cfg.Engine.ServiceDestinations = map[string]config.Destination{
 			args[0]: dst,
 		}
+		cfg.Engine.ActiveService = args[0]
 	} else {
 		cfg.Engine.ServiceDestinations[args[0]] = dst
 	}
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@ashley-cui @jwhonce PTAL

contrib/msi/podman.wxs
```diff
@@ -24,16 +24,15 @@
...
+              <File Id="520C6E17-77A2-4F41-9611-30FA763A0702" Name="podman.exe" Source="bin/podman-remote-windows.exe" KeyPath="yes"/>
             </Component>
           </Directory>
         </Directory>
       </Directory>
     </Directory>
 
     <Property Id="setx" Value="setx.exe"/>
-    <CustomAction Id="ChangePath" ExeCommand="PATH &quot;%PATH%;[INSTALLDIR] &quot;" Property="setx" Execute="deferred" Impersonate="yes" Return="check"/>
+    <CustomAction Id="ChangePath" ExeCommand="PATH &quot;%PATH%;[INSTALLDIR]&quot;" Property="setx" Execute="deferred" Impersonate="yes" Return="check"/>
```
Luap99, line 35: @baude I think using setx isn't a good option. It doesn't remove the entry on uninstall.

I stumbled across this SO post: https://stackoverflow.com/questions/1931586/can-anyone-give-me-a-example-of-modifying-windows-environment-system-variables-i
wix toolset doc: https://wixtoolset.org/documentation/manual/v3/xsd/wix/environment.html

So i think the correct way would be to use this line: `<Environment Id="PATH" Name="PATH" Value="[INSTALLDIR]" Permanent="no" Part="last" Action="set" System="yes" />` and I think it has to be placed between line 27 and 28.

baude: @Luap99 thanks! I'll take a look.  Many, many, thanks

baude: @jwhonce this is starting to smell like a difference between wixl and wix?

Luap99: @baude Unfortunately the environment key is not supported by wixl. I found this PR: https://gitlab.gnome.org/GNOME/msitools/-/merge_requests/11

jwhonce: @Luap99 Yes, would love to do this the "right" way once wixl supports it. /cc @baude

cmd/podman/system/connection/add.go
```diff
@@ -181,12 +182,20 @@ func getUDS(cmd *cobra.Command, uri *url.URL) (string, error) {
 		authMethods = append(authMethods, ssh.PublicKeysCallback(a.Signers))
 	}
 
-	config := &ssh.ClientConfig{
+	if len(authMethods) == 0 {
+		pass, err := terminal.ReadPassword("Login password:")
```
jwhonce, line 186:
```suggestion
		pass, err := terminal.ReadPassword(fmt.Sprintf("%s's login password:", uri.User.Username())
```

PR #7306, opened 2020-08-12
Change /sys/fs/cgroup/systemd mount to rprivate

mheon:
I used the wrong propagation first time around because I forgot that rprivate is the default propagation. Oops. Switch to rprivate so we're using the default.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -626,7 +626,7 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
 			Destination: "/sys/fs/cgroup/systemd",
 			Type:        "bind",
 			Source:      "/sys/fs/cgroup/systemd",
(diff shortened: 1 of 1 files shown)


mheon:
@vrothberg @baude @giuseppe @rhatdan @TomSweeneyRedHat PTAL

rhatdan:
I could argue this either way, but I guess rprivate is more secure.  The problem with rprivate is if the administrator was to mount something on /sys/fs/cgroup after the container was created, it would not see the mountpoint with rprivate.  This is not likely to happen so LGTM

mheon:
Hm. Let me double-check, actually, make sure this doesn't interfere with the release_agent masked path we mount under it.

TomSweeneyRedHat:
LGTM
if things checkout out @mheon

rhatdan:
/lgtm
/hold
@mheon Cancel this when you are sure.  I don't believe it will be an issue.

mheon:
/hold cancel

We should be fine

PR #7308, opened 2020-08-12
Add support for setting the CIDR when using slirp4netns

hamzadis:
This adds support for the --cidr parameter that is supported
by slirp4netns since v0.3.0. It allows the user to change the
IP range that is used for the network inside the container.

Changing the IP range can fix address conflicts with outside
networks or, which is the issue I had, with the virtualbox NAT
network mode which uses the same range (10.0.2.0/24).

This was referenced in https://github.com/rootless-containers/slirp4netns/issues/70 but
never got resolved.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -563,6 +563,7 @@ Valid values are:
 - `private`: create a new namespace for the container (default)
 - `slirp4netns[:OPTIONS,...]`: use slirp4netns to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
   - **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`). Default is false.
+  - **cidr=CIDR**: Specify ip range to use for this network. (Default is `10.0.2.0/24`).
   - **enable_ipv6=true|false**: Enable IPv6. Default is false. (Required for `outbound_addr6`).
   - **outbound_addr=INTERFACE**: Specify the outbound interface slirp should bind to (ipv4 traffic only).
   - **outbound_addr=IPv4**: Specify the outbound ipv4 address slirp should bind to.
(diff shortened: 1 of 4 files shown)


docs/source/markdown/podman-create.1.md
```diff
@@ -562,6 +562,7 @@ Valid values are:
 - `ns:<path>`: path to a network namespace to join
 - `private`: create a new namespace for the container (default)
 - `slirp4netns[:OPTIONS,...]`: use slirp4netns to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
+  - **cidr=CIDR**: Specify ip range to use for this network. (Default is `10.0.2.0/24`).
```
mheon, line 565: This should probably go one line down (alphabetical order)

docs/source/markdown/podman-run.1.md
```diff
@@ -570,9 +570,15 @@ Valid _mode_ values are:
 - **ns:**_path_: path to a network namespace to join;
 - `private`: create a new namespace for the container (default)
 - **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
-  **port_handler=rootlesskit**: Use rootlesskit for port forwarding.  Default.
-  **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
-  **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).  Default to false.
+  - **cidr=CIDR**: Specify ip range to use for this network. (Default is `10.0.2.0/24`).
```
mheon, line 573: Same - one line down, please

mheon:
Minor nits on manpages, otherwise LGTM
/approve

@rhatdan @vrothberg @giuseppe @TomSweeneyRedHat PTAL

docs/source/markdown/podman-create.1.md
hamzadis: Moved.

rhatdan:
This LGTM, but I would like to see this wired into containers.conf, so that users could permanently change this. I would also like to see the other slirp options available in containers.conf.

Then demonstrate the CIDR correctly via the --help, and hard code it in containers.conf rather then inside of the code.

PR #7310, opened 2020-08-12
Makefile: do not hardcode btrfs exclusion tags

lsm5:


TomSweeneyRedHat:
LGTM assuming happy tests

rhatdan:
We don't want any of these in the remote command, They should have no effect, but we don't want to have btrfs compiled if it leaks into remote.

lsm5:
hmm, @baude did you agree with this change yesterday, or did I misread what you said?

PR #7311, opened 2020-08-12
Support sighup reload configuration files

QiWang19:
Support podman service sighup reload configuration files(containers.conf, registries.conf, storage.conf).

### DIFF
--- a/cmd/podman/system/service_abi.go
+++ b/cmd/podman/system/service_abi.go
@@ -5,8 +5,12 @@ package system
 import (
 	"context"
 	"net"
+	"os"
+	"os/signal"
 	"strings"
 
+	"github.com/containers/podman/v2/cmd/podman/utils"
+	"github.com/containers/podman/v2/libpod"
 	api "github.com/containers/podman/v2/pkg/api/server"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/domain/infra"
@@ -39,6 +43,7 @@ func restService(opts entities.ServiceOptions, flags *pflag.FlagSet, cfg *entiti
 		return err
 	}
 
+	startWatcher(rt)
 	server, err := api.NewServerWithSettings(rt, opts.Timeout, listener)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


mheon:
This also needs to reload the Libpod state to account for potential changed
storage/database paths. We also potentially loaded a fresh database, so we
need to re-do the Libpod initialization process (checking if the alive file
is created and refreshing the state + creating it if it is not present).

mheon:
Also, the configured and default OCI runtimes may have changed, so we need
to refresh those as well.

cmd/podman/system/service_abi.go
```diff
@@ -55,3 +60,23 @@ func restService(opts entities.ServiceOptions, flags *pflag.FlagSet, cfg *entiti
...
+// startWatcher starts a new SIGHUP go routine for the current config.
+func startWatcher(rt *libpod.Runtime) {
+	// Setup the signal notifier
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, utils.Hup)
+
+	go func() {
+		for {
+			// Block until the signal is received
+			<-ch
```
vrothberg, line 73: Could we add a debug message here as well? This could come in handy for support.

cmd/podman/utils/signals_linux.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// Platform specific signal synonyms
+var (
+	Hup os.Signal = unix.SIGHUP
```
vrothberg, line 13:
```suggestion
	SIGHUP os.Signal = unix.SIGHUP
```

Might be easier to understand when the identifier is consistent.

libpod/runtime.go
```diff
@@ -816,3 +818,48 @@ func (r *Runtime) mergeDBConfig(dbConfig *DBConfig) {
 func (r *Runtime) EnableLabeling() bool {
 	return r.config.Containers.EnableLabeling
 }
+
+func (r *Runtime) Reload() error {
```
vrothberg, line 822: Please add a comment.

mheon:
Woah woah woah. Were my comments here ignored? This *does not work* if anything related to the OCI runtime or storage paths changed.

QiWang19:
Can I have more details about what is specifically the LIbpod state and storage paths that need to update? 
Sorry missed that in this PR, I can set up another PR to fix it.

mheon:
We need to:
- Reload the Libpod database to account for potential changed paths
- Perform the Libpod initialization process (checking the Alive file, creating it if it does not exist, refreshing containers if it does not exist)
- Load new OCI runtimes from the configuration file and set a new default OCI runtime

QiWang19:
> We need to:
> 
>     * Reload the Libpod database to account for potential changed paths
> 
>     * Perform the Libpod initialization process (checking the Alive file, creating it if it does not exist, refreshing containers if it does not exist)
> 
>     * Load new OCI runtimes from the configuration file and set a new default OCI runtime

Does this just need to call makeruntime https://github.com/containers/podman/blob/80d2c0135052ee8f93abd59d9e63e5699fb94288/libpod/runtime.go#L241 after reloading the configuration files?

Issue #7294, opened 2020-08-11
Containers started using socket-activated APIv2 die from systemd activation timeout


danopia:
/kind bug

**Description**

When creating and starting containers specifically via the `/run/podman/podman.sock` endpoint, the containers are stopped when `podman.service` gets shut down due to socket activation timeout. This manifests as containers stopping 10s after the last APIv2 request is made. The shutdown is clean; `podman ps -a` shows exit code 0 for them.

Opening an APIv2 events stream or issuing APIv2 requests every ~5 seconds causes the containers to stay alive longer; but if those requests stop, the containers still stop then.

**Steps to reproduce the issue:**

Minimal case assuming you have podman's systemd units installed already: 

1. Check you have the socket unit ready: `systemctl enable --now podman.socket`

1. Create a sleep container:
    1. `curl --unix-socket /run/podman/podman.sock -XPOST http://d/v1.0.0/libpod/images/pull'?reference=k8s.gcr.io/pause:3.2'`
    1. `curl --unix-socket /run/podman/podman.sock -XPOST http://d/v1.0.0/libpod/containers/create --header 'content-type: application/json' --data '{"image":"k8s.gcr.io/pause:3.2","name":"bg-test"}'`

1. Start watching events: `podman events &`

1. Start the container: `curl --unix-socket /run/podman/podman.sock -XPOST http://d/v1.0.0/libpod/containers/bg-test/start`

**Describe the results you received:**

```
2020-08-11 14:52:21.101145183 +0000 UTC container init 1d1018fa7c028dd9341df4fd26d3b9d04a719ca4b4913da872f9c6d408a831e5 (image=k8s.gcr.io/pause:3.2, name=bg-test)
2020-08-11 14:52:21.114121047 +0000 UTC container start 1d1018fa7c028dd9341df4fd26d3b9d04a719ca4b4913da872f9c6d408a831e5 (image=k8s.gcr.io/pause:3.2, name=bg-test)
[ ... 10 seconds pass ... ]
2020-08-11 14:52:31.215222381 +0000 UTC container died 1d1018fa7c028dd9341df4fd26d3b9d04a719ca4b4913da872f9c6d408a831e5 (image=k8s.gcr.io/pause:3.2, name=bg-test)
[...]

Issue #7195, opened 2020-08-03
CI flake: podman-remote: no output from container (and "does not exist in database"?)


edsantiago:
This is a bad report. I have no reproducer nor any real sense for what's going on.

I'm seeing consistent flakes in #7111 . The failing test is always "podman run : user namespace preserved root ownership" which is simply a quick loop of `podman run` commands. The last set of failures all looked like:
```
[+0136s] # # podman-remote --url ... run --rm --user=100 --userns=keep-id quay.io/libpod/alpine_labels:latest stat -c %u:%g:%n /etc
[+0136s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+0136s] # #|     FAIL: run  --user=100 --userns=keep-id (/etc)   <<<--- these flags are not always the same
[+0136s] # #| expected: '0:0:/etc'
[+0136s] # #|   actual: ''
[+0136s] # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
In two of the three most recent failures, in teardown, there's a `podman rm -a -f` that barfs with:
```
[+0136s] # Error: container d0ea34aaeffe07cc3e3f7f79933372a2bab825ef97e17c580eb1e1e94b2ac7e7 does not exist in database: no such container
```
Logs: [fedora 32](https://cirrus-ci.com/task/5457991551942656?command=system_test#L202), [fedora 31](https://cirrus-ci.com/task/4754304110166016?command=system_test#L202), [special testing rootless](https://cirrus-ci.com/task/5598729040297984?command=system_test#L173).

In an even earlier run, in [special_testing_rootless](https://cirrus-ci.com/task/6706268498821120?command=system_test#L224), there was a different error in a different test:
```
[+0257s] not ok 74 podman volume with --userns=keep-id
[+0257s] # $ /var/tmp/go/src/github.com/containers/podman/bin/podman-remote --url unix:/tmp/podman.fpYZ0P run --rm -v /tmp/podman_bats.LsZK3v/volume_O8zRoMsGmt:/vol:z quay.io/libpod/alpine_labels:latest stat -c %u:%s /vol/myfile
[+0257s] # read unixpacket @->/run/user/23298/libpod/tmp/socket/1bee53f8e19e2b03ff773e504fead7f5994b8be5545b315a73a3d2cef290f567/attach: read: connection reset by peer
[...]

PR #7312, opened 2020-08-13
podman.service: use sdnotify

vrothberg:
Commit 2b6dd3fb4384 set the killmode of the podman.service to the
systemd default which ultimately lead to the problem that systemd
will kill *all* processes inside the unit's cgroup and hence kill
all containers whenever the service is stopped.

Fix it by setting the type to sdnotify and the killmode to process.
`podman system service` will send the necessary notify messages
when the SDNOTIFY_SOCKET is set and unset it right after to prevent
the backend and container runtimes from jumping in between and send
messages as well.

Fixes: #7294

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"net"
 	"net/http"
@@ -17,6 +18,7 @@ import (
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/api/server/idletracker"
 	"github.com/coreos/go-systemd/v22/activation"
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
@@ -147,8 +149,31 @@ func newServer(runtime *libpod.Runtime, duration time.Duration, listener *net.Li
 	return &server, nil
 }
 
-// Serve starts responding to HTTP requests
+// If the NOTIFY_SOCKET is set, communicate the PID and readiness, and
+// further unset NOTIFY_SOCKET to prevent containers from sending
+// messages and unset INVOCATION_ID so conmon and containers are in
+// the correct cgroup.
+func setupSystemd() {
+	if len(os.Getenv("NOTIFY_SOCKET")) == 0 {
+		return
+	}
+	payload := fmt.Sprintf("MAINPID=%d", os.Getpid())
+	payload += "\n"
+	payload += daemon.SdNotifyReady
+	if sent, err := daemon.SdNotify(true, payload); err != nil {
+		logrus.Errorf("Error notifying systemd of Conmon PID: %s", err.Error())
+	} else if sent {
+		logrus.Debugf("Notify sent successfully")
+	}
+
+	if err := os.Unsetenv("INVOCATION_ID"); err != nil {
+		logrus.Errorf("Error unsetting INVOCATION_ID: %s", err.Error())
+	}
+}
+
+// Serve starts responding to HTTP requests.
 func (s *APIServer) Serve() error {
+	setupSystemd()
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
 	errChan := make(chan error, 1)
--- a/contrib/systemd/system/podman.service
+++ b/contrib/systemd/system/podman.service
@@ -6,5 +6,6 @@ Documentation=man:podman-system-service(1)
 StartLimitIntervalSec=0
 
 [Service]
-Type=simple
+Type=notify
+KillMode=process
 ExecStart=/usr/bin/podman system service


vrothberg:
@martinpitt @goochjj PTAL

pkg/api/server/server.go
```diff
@@ -147,7 +149,27 @@ func newServer(runtime *libpod.Runtime, duration time.Duration, listener *net.Li
...
+	}
+	payload := fmt.Sprintf("MAINPID=%d", os.Getpid())
+	payload += "\n"
+	payload += daemon.SdNotifyReady
+	if sent, err := daemon.SdNotify(false, payload); err != nil {
+		logrus.Errorf("Error notifying systemd of Conmon PID: %s", err.Error())
+	} else if sent {
+		logrus.Debugf("Notify sent successfully")
+	}
+	if err := os.Unsetenv("NOTIFY_SOCKET"); err != nil {
```
giuseppe, line 167: could we use `daemon.SdNotify(true, ...)` and avoid the `Unsetenv`?

vrothberg: Good catch :+1:

vrothberg:
This looks good to me now. Podman will now be stopped as systemd knows it's PID. Also systemd is not complaining about old conmons hanging around but it actually reports them. Below you can see what I'm referring to. I ran a bunch of containers, waited for the service to timeout and then did an `image list` to fire up the service again:
```
[root@nebuchadnezzar libpod]# systemctl status podman.service
● podman.service - Podman API Service
     Loaded: loaded (/usr/local/lib/systemd/system/podman.service; enabled; vendor preset: disabled)
     Active: active (running) since Thu 2020-08-13 13:18:14 CEST; 3s ago
TriggeredBy: ● podman.socket
       Docs: man:podman-system-service(1)
   Main PID: 1198502 (podman)
      Tasks: 27 (limit: 19009)
     Memory: 39.0M
        CPU: 191ms
     CGroup: /system.slice/podman.service
             ├─1197016 /usr/bin/conmon --api-version 1 -c b3ef44b91d1108f23c04f19b61b59c8ec1d6f57918ccca6a84bc64bf86eac834 
             ├─1197240 /usr/bin/conmon --api-version 1 -c fb6e0ae46293a97bf166dc3a76e17f549b949b28a6d66601ca27b220ce9431ec 
             ├─1197955 /usr/bin/conmon --api-version 1 -c da360f41e2d854e4c12b0f061b230c04a5f4f7e1abc408e60ef15ec1efc57c9c 
             ├─1198029 /usr/bin/conmon --api-version 1 -c d931f5e0bea1dc2acad2c1c32bedad7cc5b5571d4b5048a1f5d5fbae2ba76d60 
             ├─1198096 /usr/bin/conmon --api-version 1 -c 1b84665fa739d408146bfbf03ab775c9bc18c7b648e13d848bd832dcfb432312 
             ├─1198162 /usr/bin/conmon --api-version 1 -c 1ba81c25316fb91e2f585969a6cc6a2df4f9352fa3e44fa931fe8981f5f89308 
             ├─1198229 /usr/bin/conmon --api-version 1 -c aaabfd54a0fd3e3513ab91c73aaf91e9184df9f420150c1b2e9cd5a2a013d893 
             └─1198502 /usr/local/bin/podman system service
```

mheon:
LGTM

vrothberg:
Please wait with merging until we have more acks on it.

rhatdan:
I thought you had two.  But you have control anyways.

pkg/api/server/server.go
```diff
@@ -147,8 +149,26 @@ func newServer(runtime *libpod.Runtime, duration time.Duration, listener *net.Li
 	return &server, nil
 }
 
-// Serve starts responding to HTTP requests
+// If the SDNOTIFY_SOCKET is set, communicate the PID and readiness, and
+// further unset the SDNOTIFY_SOCKET to prevent containers from sending
```
edsantiago, line 153: I'm still struggling to understand the entire ramifications of this PR, but while I do so, you may want to remove the misleading `SD` from the envariable name in the comments above.

giuseppe:
> This looks good to me now. Podman will now be stopped as systemd knows it's PID. Also systemd is not complaining about old conmons hanging around but it actually reports them. Below you can see what I'm referring to. I ran a bunch of containers, waited for the service to timeout and then did an `image list` to fire up the service again:

why do we have conmon processes in the podman.service cgroup?  I thought we moved them to a separate cgroup

vrothberg:
> why do we have conmon processes in the podman.service cgroup? I thought we moved them to a separate cgroup

Could it be because we're unsetting NOTIFY_SOCKET?

giuseppe:
> Could it be because we're unsetting NOTIFY_SOCKET?

that should not happen (at least I don't see we are using it internally in Podman).  I don't think the issue is introduced with this change.  Or are those old containers that are configured with --cgroup-mode=split?

vrothberg:
> > Could it be because we're unsetting NOTIFY_SOCKET?
> 
> that should not happen (at least I don't see we are using it internally in Podman). I don't think the issue is introduced with this change. Or are those old containers that are configured with --cgroup-mode=split?

They are created with the remote client. Maybe we're missing that in the remote code paths?

giuseppe:
> They are created with the remote client. Maybe we're missing that in the remote code paths?

this patch seems to do the trick:
```
diff --git a/pkg/api/server/server.go b/pkg/api/server/server.go
index 18b48a3f6..2a3e44e04 100644
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -153,6 +153,7 @@ func (s *APIServer) Serve() error {
        signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
        errChan := make(chan error, 1)
 
+       os.Unsetenv("INVOCATION_ID")
        go func() {
                <-s.idleTracker.Done()
                logrus.Debugf("API Server idle for %v", s.idleTracker.Duration)
```

vrothberg:
> this patch seems to do the trick:

Tested successfully, thanks @giuseppe !

vrothberg:
Allright, now as we've ironed out the last remaining fart, I am good to merge. Thanks all!

goochjj:
LGTM

TomSweeneyRedHat:
The change LGTM, but it looks like you may have encountered a real error in the tests.

mheon:
F31 looks like one of our remote flakes...

/lgtm
/hold

edsantiago:
yes, it's the #7195 flake

TomSweeneyRedHat:
Thanks, @QiWang19 's #7311 suffers the same issue.

edsantiago:
/hold

investigating something. Please do not merge yet.

edsantiago:
All clear. The problem I was looking at turned out to be caused by something else.

Issue #7302, opened 2020-08-12
Podman vs Docker networking: IPv4, IPv6 and Dual Stack


aojea:
/kind bug
/kind feature

**Description**

I do not think that podman and docker network will behave similar, simple because this is a long history between containers networking, CNM/libnetwork vs CNI https://kubernetes.io/blog/2016/01/why-kubernetes-doesnt-use-libnetwork/

Docker network ALWAYS use IPv4, if you enable IPv6 then it creates dual-stack networks..

```
docker network inspect kind
[
    {
        "Name": "kind",
        "Id": "979f492554fd8b8dc94d58f8e989dd5878ea410b4ff0b75457359d00cfa19ee6",
        "Created": "2020-07-28T13:01:45.860343319+02:00",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": true,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "172.18.0.0/16",
                    "Gateway": "172.18.0.1"
                },
                {
                    "Subnet": "fc00:f853:ccd:e793::/64",
                    "Gateway": "fc00:f853:ccd:e793::1"
                }
            ]
        },

```

With `podman network` you can only have single stack networks for the containers, because you can only specify one range as parameter:

```
$ sudo podman network inspect kind
(snipped)
        "bridge": "cni-podman5",
        "hairpinMode": true,
        "ipMasq": true,
        "ipam": {
          "ranges": [
            [
              {
                "gateway": "fc00:8866:27d0:bd7e::1",
                "subnet": "fc00:8866:27d0:bd7e::/64"
              }
            ]
          ],
          "routes": [
```

CNI allows dual-stack configurations, something we already have in KIND, as you can see in this snippet

https://github.com/kubernetes-sigs/kind/blob/afe54f7eb52c902ff5bdb23e522bc38e00a71e27/images/kindnetd/cmd/kindnetd/cni.go#L34-L64

I'm happy to submit a PR to implement this, but I'm interested in hearing what are the expected semantics for `podman network`, 
I do see value in creating IPv6 only containers, but in the other hand I recognize that this is far from being a reality and most people use dual stack on containers.

* Should we pluralize the parameters for ranges in the `podman network create` 
https://github.com/containers/podman/blob/588df9039700e53eb3d8222c15494a5dd727e4f5/pkg/domain/entities/network.go#L40-L50 ?
[...]

PR #7314, opened 2020-08-13
IPv6 default route

aojea:
The networking configuration was assuming always that the default
gateway was 0.0.0.0/0, that does not work if we have IPv6
containers.

We must generate the corresponding default route depending of the
IP family of the subnet used.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

### DIFF
--- a/pkg/domain/infra/abi/network.go
+++ b/pkg/domain/infra/abi/network.go
@@ -191,7 +191,7 @@ func createBridge(r *libpod.Runtime, name string, options entities.NetworkCreate
 	var plugins []network.CNIPlugins
 	var routes []network.IPAMRoute
 
-	defaultRoute, err := network.NewIPAMDefaultRoute()
+	defaultRoute, err := network.NewIPAMDefaultRoute(network.IsIPv6(subnet.IP))
 	if err != nil {
 		return "", err
 	}
--- a/pkg/network/netconflist.go
+++ b/pkg/network/netconflist.go
@@ -6,6 +6,11 @@ import (
 	"path/filepath"
 )
 
+const (
+	defaultIPv4Route = "0.0.0.0/0"
+	defaultIPv6Route = "::/0"
+)
+
 // NcList describes a generic map
 type NcList map[string]interface{}
 
@@ -86,9 +91,13 @@ func NewIPAMRoute(r *net.IPNet) IPAMRoute { //nolint:interfacer
 }
 
 // NewIPAMDefaultRoute creates a new IPAMDefault route of
-// 0.0.0.0/0
-func NewIPAMDefaultRoute() (IPAMRoute, error) {
-	_, n, err := net.ParseCIDR("0.0.0.0/0")
+// 0.0.0.0/0 for IPv4 or ::/0 for IPv6
+func NewIPAMDefaultRoute(isIPv6 bool) (IPAMRoute, error) {
+	route := defaultIPv4Route
+	if isIPv6 {
+		route = defaultIPv6Route
+	}
+	_, n, err := net.ParseCIDR(route)
 	if err != nil {
 		return IPAMRoute{}, err
 	}
--- a/pkg/network/ip.go
+++ b/pkg/network/ip.go
@@ -12,3 +12,8 @@ func CalcGatewayIP(ipn *net.IPNet) net.IP {
 	nid := ipn.IP.Mask(ipn.Mask)
 	return ip.NextIP(nid)
 }
+
+// IsIPv6 returns if netIP is IPv6.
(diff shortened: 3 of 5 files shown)


aojea:
/assign @mheon @mrunalp 

This allows to create IPv6 clusters with KIND
```
NAME                 STATUS   ROLES    AGE    VERSION   INTERNAL-IP              EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION           CONTAINER-RUNTIME
kind-control-plane   Ready    master   4m9s   v1.18.6   fc00:681e:5d82:3c05::2   <none>        Ubuntu 20.04 LTS   5.7.10-201.fc32.x86_64   containerd://1.4.0-beta.1-85-g334f567e
```

go.mod
```diff
@@ -65,4 +65,5 @@ require (
 	k8s.io/api v0.18.6
 	k8s.io/apimachinery v0.18.6
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+	k8s.io/utils v0.0.0-20190221042446-c2654d5206da
```
mheon, line 68: I wonder if this adds significantly to our binary size - we've had bad experiences with K8S dependencies ballooning the Podman executable by 5+ MB in size before...

pkg/network/netconflist.go
```diff
@@ -86,13 +91,14 @@ func NewIPAMRoute(r *net.IPNet) IPAMRoute { //nolint:interfacer
...
-	if err != nil {
-		return IPAMRoute{}, err
+// 0.0.0.0/0 for IPv4 or ::/0 for IPv6
+func NewIPAMDefaultRoute(isIPv6 bool) IPAMRoute {
+	_, n, _ := net.ParseCIDR(defaultIPv4Route)
+	if isIPv6 {
+		_, n, _ = net.ParseCIDR(defaultIPv6Route)
+
 	}
-	return NewIPAMRoute(n), nil
```
aojea, line 95: we can't return an error parsing a constant

pkg/domain/infra/abi/network.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/containers/podman/v2/pkg/network"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/pkg/errors"
+
+	utilnet "k8s.io/utils/net"
```
aojea, line 19: it's a small dependency and very useful to deal with isIPv6 things

go.mod
aojea: yeah, I know the pain, this is very small and without indirect dependencies. Is just 2 files but very useful helpers, at least we are using it a lot in k8s and ovn-kube

pkg/network/netconflist.go
```diff
@@ -86,13 +91,14 @@ func NewIPAMRoute(r *net.IPNet) IPAMRoute { //nolint:interfacer
...
-// 0.0.0.0/0
-func NewIPAMDefaultRoute() (IPAMRoute, error) {
-	_, n, err := net.ParseCIDR("0.0.0.0/0")
-	if err != nil {
-		return IPAMRoute{}, err
+// 0.0.0.0/0 for IPv4 or ::/0 for IPv6
+func NewIPAMDefaultRoute(isIPv6 bool) IPAMRoute {
+	_, n, _ := net.ParseCIDR(defaultIPv4Route)
+	if isIPv6 {
+		_, n, _ = net.ParseCIDR(defaultIPv6Route)
```
mheon, line 98: I wonder if we should keep the error but panic instead of returning it - maybe even make this an `init()` to guarantee that we don't make a working Podman if we accidentally add a bad constant?

go.mod
aojea: forget to mention that we are going to need this for all the dual stack stuff
https://github.com/containers/podman/issues/7302

pkg/network/netconflist.go
aojea: now that you mention we can get rid of this and return a constant directly in pkg/domain/infra/abi/network.go

aojea: will keep it as it was, with the err, to make the diff smaller, as you say it can deviate in the future but right now there is no  big gain here

go.mod
aojea: I've removed dependency, if we need it later it can be revisited but the bits that we need does not justify vendoring

aojea:
@mheon PTAL, I've removed the dependencies, kept the same signature for `NewIPAMDefaultRoute` and added an e2e test

mheon:
/approve
LGTM

Issue #7309, opened 2020-08-12
Is there any way to override the storage.conf options?


ziXet:
I would like to run some podman commands with different options in the storage.conf.

Without this option I need to use "sed" to comment and uncomment specific lines in the storage.conf before and after running my special podman command which is not ideal with multiprocessing.

I'd prefer to run something like this:
`podman image pull --storage storage.options.additionalimagestores=["/storage"]`

Is it possible?

giuseppe:
podman --storage-opt...?  Is that what you would like to use?

ziXet:
> podman --storage-opt...? Is that what you would like to use?

Thanks for the response.
I would like to override the value of additionalimagestores in the storage.conf. I think --storage-opt is for overriding "storage driver options". 

Please correct me if I'm wrong

giuseppe:
I think you are right.  I don't see a way to change the additionalimagestores using `--storage-opt`.

Anyway, I think that allowing a different additionalimagestores per container could cause some issues.  For example how would we find the image referred by the container if the stores can change?  We would need to store the configuration per each container

ziXet:
My use case is a bit special. I would like to pull the images to the additional storage using: 
`podman image pull myimage:v1 --root /storage`
If I run this command while I have `additionalstorage=['/storage']` in my storage.conf, I get this error:
`Error: lock "/storage/overlay-images/images.lock" is not a read-only lock`
To solve this issue, I need to set the `additionalstorage=[]` or comment it out. after pulling a readonly image, I can put the '/storage' back to the additionalstorage.

rhatdan:
I am working on a better fix for this, but this will work.

podman image pull myimage:v1 --storage-opt=overlay2.mountopt=nodev --root /storage

ziXet:
> I am working on a better fix for this, but this will work.
> 
> podman image pull myimage:v1 --storage-opt=overlay.mountopt=nodev --root /storage

Thanks Dan! it actually worked!

rhatdan:
I have a PR for this, just need to get it merged.

rhatdan:
This is now fixed in master branch.

PR #7315, opened 2020-08-13
Allow users to override all storage options at the command line

rhatdan:
podman --root /var/lib/shared --storage-opt="" pull alpine

Tells podman to ignore the storage options defined in storage.conf.

This would allow users to configure additional stores, and still use
podman to update the additional store storage.

### DIFF
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -138,7 +138,7 @@ specify additional options via the `--storage-opt` flag.
 
 #### **--storage-opt**=*value*
 
-Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all.
+Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all. Use --storage-opt="" to ignore all storage options in the configuration files.
 
 #### **--syslog**=*true|false*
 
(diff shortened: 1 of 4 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/7309

TomSweeneyRedHat:
A test would be good.
Otherwise LGTM

mheon:
Concur that a test would be nice

rhatdan:
The problem is there is no good way to test with storage.conf that I can see, without overwriting the system configuration files.
@edsantiago Any ideas?  We basically need a system wide storage.conf with storage-opt set, and then we need to test that this
podman --storage-opt "" info --format "{{.Store.GraphOptions}}"
Ends up equal to "".

edsantiago:
I was just looking at this PR and pondering that very question. I've long wanted a way to test config files, but really the only sane way to do that is with a `--config-file-path=xxx` command-line option. With rootless it could be possible to override by redefining `$HOME` or $XDG_SOMETHING`, but that would have cascading effects because of image/container storage. 

For a case like this, where you don't need `podman run` or anything that actually uses images, I think you could:
```bash
@test ..... {
    fake_conf=$PODMAN_TMPDIR/home/.config/containers/containers.conf
    mkdir -[ $(dirname $fake_conf)
    cat >$fake_conf <<EOF
blah blah blah
EOF

    HOME=$PODMAN_TMPDIR/home run_podman --storage-opt="" info ...
```
Disclaimer: I typed code directly into a browser without testing. That is always a bad idea.

rhatdan:
Sadly this PR will only work for rootfull podman

rhatdan:
Need https://github.com/containers/common/pull/262 in order to add tests.

vrothberg:
@rhatdan, let's get things moving before we the PR has birthday :)

rhatdan:
I have fixed this via other means.

adamish:
Another possible solution could be podman providing a command that modified the storage.conf file in a safe and consistent way, something like
```
podman conf add-additionalimagestore "/foo"
podman conf remove-additionalimagestore "/foo"
```

rhatdan:
That would be a very complex tool and would need to be modified each time a new config entry was added to containers.conf.  If there was a simple way to autogenerate the tool, I might consider it. but this seems like you are building a toml editor.

Issue #7316, opened 2020-08-13
podman run -d: hangs when $NOTIFY_SOCKET is set


edsantiago:
```console
# export NOTIFY_SOCKET=/tmp/mypodmansocket
# socat unix-recvfrom:"$NOTIFY_SOCKET",fork system:"(cat;echo)" &
[1] 104770
# podman run -d --sdnotify=container alpine sh -c 'sleep 10'
2020/08/13 16:50:03 socat[104857] E sendto(8, 0x55de50ff26b0, 14, 0, AF=1 "<anon>", 0): Transport endpoint is not connected
95fb97931bbf3c3b773dddff07f9fea80b341293fb39e6d63e3becbd3937099f
2020/08/13 16:50:13 socat[104864] E sendto(8, 0x55de50ff26b0, 14, 0, AF=1 "<anon>", 0): Transport endpoint is not connected
```
Note the ten-second difference in the `socat` timestamps; that is because the container is `sleep 10`. Change it to `sleep 30`, you get a 30-second delay.

What I expected: since this is `run -d` (detached), I expected podman to detach immediately and let the container deal with sdnotify.

(How I found this: trying to run `systemd-notify` in a `fedora:latest` container. Ha ha, silly me, they removed systemd-notify from that image). Ergo, I think this is counterintuitive behavior if a user has a container that never makes it to `sdnotify`. I really don't expect `podman run -d` to hang forever.

edsantiago:
Oooh! Try running `podman ps` or even `podman info` while there's a hung `podman run -d`. It too will hang. (`podman images` is fine).

mheon:
Is this the same issue as https://github.com/containers/podman/issues/6688

edsantiago:
Yes, it looks like a different manifestation of the same problem. I'd say "feel free to close", but given how badly #6688 has been neglected, I'm going to leave it open as a reminder that this really is an unpleasant and unacceptable bug,

giuseppe:
I don't think it is a bug.  `podman run` waits for the container to notify when it is ready.  If the container is never ready what should we do?  It is not even Podman fault at this point, the OCI runtime is handling the NOTIFY_SOCKET

edsantiago:
I'm fine with the container hanging. I'm not fine with `podman ps` or `podman info` or other podman commands hanging.

rhatdan:
Yes, I think we need a way to work around the lock.

mheon:
[...]

PR #7317, opened 2020-08-13
system tests: enable sdnotify tests

edsantiago:
Oops. PR #6693 (sdnotify) added tests, but they were disabled
due to broken crun on f31. I tried for three weeks to get a
magic CI:IMG PR to update crun on the CI VMs ... but in that
time I forgot to actually enable those new tests.

This PR removes a 'skip'. It also changes the test image,
from fedora:latest to :31, because (sigh) :latest removed
the systemd-notify tool.

WARNING WARNING WARNING: the symptom of a missing systemd-notify
is that podman will hang forever, not even stopped by the timeout
command in podman_run! (Filed: #7316). This means that if the
sdnotify-in-container test ever fails, the symptom will be that
Cirrus itself will time out (2 hours?). This is horrible. I
don't know what to do about it other than push for a fix for 7316.

### DIFF
--- a/test/system/260-sdnotify.bats
+++ b/test/system/260-sdnotify.bats
@@ -12,8 +12,15 @@ _SOCAT_LOG=
 function setup() {
     skip_if_remote
 
-    # TODO: remove this once CI systems have newer crun and container-selinux
-    skip "TEMPORARY SKIP - until CI systems get new crun, container-selinux"
+    # Skip if systemd is not running
+    systemctl list-units &>/dev/null || skip "systemd not available"
+
+    # sdnotify fails with runc 1.0.0-3-dev2 on Ubuntu. Let's just
+    # assume that we work only with crun, nothing else.
+    run_podman info --format '{{ .Host.OCIRuntime.Name }}'
+    if [[ "$output" != "crun" ]]; then
+        skip "this test only works with crun, not '$output'"
+    fi
 
     basic_setup
 }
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

edsantiago:
Note: @vrothberg's #7312 mucks with NOTIFY_SOCKET, and I pushed this PR before that one merged. I did, of course, test (root, rootless; my f32 laptop and a rawhide VM) that both work together. I can rebase and push if anyone is nervous.

edsantiago:
What is `verify_test_built_images_task` and why is it running instead of all the other things like `testing_task`?  @cevich?

edsantiago:
@cevich this verify_test_built_images thing, could it have to do with this change in 576ce0f1b5012bec7d0a5648b1d8d752e42fba03 ?
```diff
 # Update metadata on VM images referenced by this repository state
 meta_task:

-    depends_on:
-        - "gating"
-        - "vendor"
-        - "varlink_api"
-        - "build_each_commit"
-        - "build_without_cgo"
-
```

cevich:
> What is `verify_test_built_images_task` and why is it running instead of all the other things like `testing_task`? @cevich?

@edsantiago that should only run when building new VM images, (i.e. `CI:IMG` magic).  I don't believe it will function properly in any other context.

edsantiago:
@cevich if you click 'View details' on this PR (next to openshift-merge-robot), you'll see a slew of `verify-*` jobs.

Aw, crap. I bet they triggered because I include CI:IMG in the commit message body. This is why a bad PR got merged: CI never actually ran. Filed #7374. Thanks for the insight, that's what I was missing.

PR #7333, opened 2020-08-17
Use `bash` binary from env instead of /bin/bash for scripts

saschagrunert:
It's not possible to run any of the scripts on distributions which do
have `bash` not in `/bin`. This is being fixed by using `/usr/bin/env
bash` instead.


saschagrunert:
> LGTM but `./vendor` must remain untouched.

Ah yep sure, thank you for the hint. :+1:

rhatdan:
@saschagrunert What distribution is having issues?

PR #7341, opened 2020-08-17
e2e tests: use actual temp dirs, not "/tmp/dir"

edsantiago:
One of the --iidfile tests was flaking:

   Error: failed to write image ID to file "/tmp/dir/idFile": open /tmp/dir/idFile: no such file or directory

Root cause: test was actually not mkdir'ing /tmp/dir. Test was
mostly passing because _other_ tests in the suite were mkdir'ing
it, but once in a while this test ran before the others.

Solution: fixed this test to use CreateTempDirInTempDir(). And,
since hardcoded tempdirs are bad practice, grepped for '"dir"'
and fixed all other instances too.

### DIFF
--- a/test/e2e/build_test.go
+++ b/test/e2e/build_test.go
@@ -127,8 +127,10 @@ var _ = Describe("Podman build", func() {
 		defer Expect(os.Chdir(cwd)).To(BeNil())
 
 		// Write target and fake files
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

mheon:
LGTM

PR #7342, opened 2020-08-17
WIP: TESTING: DO NOT MERGE

cevich:


cevich:
@edsantiago I think the spot you're seeing the "hang" is a red-herring, I think something's happening on another level to cause this.  Look at the example from this PR: https://cirrus-ci.com/task/5905973485240320?command=system_test#L156  Execution simply/literally stops w/o notice, until the "BIG" timeout clobbers it.  All the other tests passed, so...geeze...the _only_ thing I can think is something must be happening on the cirrus side.  Maybe related to the underlying VMs being preemptable and a secondary bug related to output processing when that happens?  https://cloud.google.com/compute/docs/instances/preemptible

cevich:
....Actually, no...scratch that.  Our VMs are ***NOT*** preemptable...we use the expensive ones :confused:

edsantiago:
FWIW: acting on the possibility that the `env bash` PR could be responsible, I rebased my two PRs on the commit prior to that. No luck on either one: both ran into the 60- and 120-minute timeout, same tests, same symptoms.

cevich:
> FWIW: acting on the possibility that the env bash PR could be responsible

Excellent, thanks for trying that.  One more question-mark item we can check off the list.

***ONE*** of my tests also hit the 120min timeout: https://github.com/containers/podman/pull/7342/checks?check_run_id=995072852 but appears almost totally different from yours.   I'm taking this as meaning the problem isn't specific to your PR, and we should basically expect to see this problem appear everywhere.

I'm going to try and re-run it to see if the behavior changes at all (IIRC, yours always failed for at least 3-5 retries).

cevich:
Problem fixed by https://github.com/containers/podman/pull/7354

PR #7343, opened 2020-08-17
Ensure DefaultEnvVariables is used in Specgen

mheon:
When we rewrote Podman's pkg/spec, one of the things that was lost was our use of a set of default environment variables, that ensure all containers have at least $PATH and $TERM set.

While we're in the process of re-adding it, change it from a variable to a function, so we can ensure the Join function does not overwrite it and corrupt the defaults.

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -12,13 +12,15 @@ import (
 	"github.com/pkg/errors"
 )
 
-// DefaultEnvVariables sets $PATH and $TERM.
-var DefaultEnvVariables = map[string]string{
-	"PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
-	"TERM": "xterm",
-}
-
 const whiteSpaces = " \t"
+
+// DefaultEnvVariables returns a default environment, with $PATH and $TERM set.
+func DefaultEnvVariables() map[string]string {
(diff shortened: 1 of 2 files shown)


mheon:
@rhatdan @baude @giuseppe @TomSweeneyRedHat PTAL

cmd/podman/common/specgen.go
```diff
@@ -308,9 +308,8 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	//
 	// Precedence order (higher index wins):
 	//  1) env-host, 2) image data, 3) env-file, 4) env
-	env := map[string]string{
-		"container": "podman",
-	}
+	env := envLib.DefaultEnvVariables()
```
rhatdan, line 311: It should be getting these fields from containers/common

mheon: We can change that in a future release; for now, we need this merged so we can get it in 2.0.5

mheon: Refactored a bit.

I don't think these make sense for containers.conf, we want to ensure a few basic environment variables are always populated - the user shouldn't really need to ever set this, outside of the existing Env section of containers.conf.

mheon:
@rhatdan Updated to remove parts that conflict with #7355 - PTAL

PR #7344, opened 2020-08-17
Cirrus: Increase integration-testing timeout

cevich:
Observed timeout problems hitting some integration-testing
tasks differently than others.  Given the current `Makefile`
has a ginkgo timeout of 90-minutes, the task timeout for
integration tests should be longer. Increase the timeout
of the main integration-test running tasks to the (default)
120min global valie in `.cirrus.yml`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -417,8 +417,6 @@ testing_task:
           gce_instance:
               image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
 
-    timeout_in: 120m
-
     env:
         ADD_SECOND_PARTITION: 'true'
         matrix:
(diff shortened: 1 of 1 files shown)


edsantiago:
I'm really not sure about this. If we were seeing:
```
log log log
lots of log
tests almost finished
Timed out!
```
...then I'd feel totally comfortable bumping up the timeout here. But we're not. We're seeing a weird unexplained timeout with no logs. I was really hoping we could figure out _why_ these timeouts are happening.

cevich:
@edsantiago that's a separate problem, though related.  While looking back through failed tasks on folks PRs, I did notice the condition you describe.  Upon investigation, I noticed that the 'testing' task uses a timeout of 120min, but the `special-testing-rootless` (which also runs integration tests) has a timeout half as long.  Same deal with `in_podman` task.  This PR simply brings all the tasks running integration tests more in-line in terms of timeouts, by having them all use the (default) 120min value.

Note: This is a task-level timeout, it doesn't take the place of the built-in 90-minute ginkgo timeout (in the Makefile).  Though it would be nice to not have the two interfere destructively.

cevich:
hmmm...maybe I need a better commit message.

rhatdan:
/approve
LGTM Makes sense to me.
I will leave it to @edsantiago to decide to merge.

PR #7346, opened 2020-08-17
Don't limit the size on /run for systemd based containers

rhatdan:
We had a customer incident where they ran out of space on /run.

If you don't specify size, it will be still limited to 50% or memory
available in the cgroup the container is running in.  If the cgroup is
unlimited then the /run will be limited to 50% of the total memory
on the system.

### DIFF
--- a/cmd/podman/common/volumes.go
+++ b/cmd/podman/common/volumes.go
@@ -88,17 +88,11 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
 			if _, ok := unifiedVolumes[dest]; ok {
 				continue
 			}
-			localOpts := options
-			if dest == "/run" {
-				localOpts = append(localOpts, "noexec", "size=65536k")
-			} else {
-				localOpts = append(localOpts, "exec")
-			}
 			unifiedMounts[dest] = spec.Mount{
 				Destination: dest,
 				Type:        TypeTmpfs,
 				Source:      "tmpfs",
-				Options:     localOpts,
+				Options:     options,
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


rhatdan:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1869322

mheon:
I'm a little uncomfortable with this given that we don't set a memory limit by default. Maybe cap at 1gb or so by default on systems with >2gb of ram and no memory limit set?

rhatdan:
tmpfs already have this limit set and users can currently max out memory on /dev/shm,  So this is just eliminating a potential bug, for not much advantage.

tmpfs is controlled by kernel to be 50% of available memory, in cgroup or in host, depending on which one is lower.

TomSweeneyRedHat:
Does this line need to change too?  https://github.com/containers/podman/blob/master/pkg/spec/storage.go#L174

mheon:
We should not be changing anything in pkg/spec without a very good reason

TomSweeneyRedHat:
Can there be tests added?

rhatdan:
@TomSweeneyRedHat What do you want me to test?  That the size is not set?

TomSweeneyRedHat:
@rhatdan my thinking was to set above and below the old limit and make sure that the value was set appropriately.

jwhonce:
LGTM

cmd/podman/common/volumes.go
```diff
@@ -88,17 +88,11 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
 			if _, ok := unifiedVolumes[dest]; ok {
 				continue
 			}
-			localOpts := options
-			if dest == "/run" {
-				localOpts = append(localOpts, "noexec", "size=65536k")
```
edsantiago, line 93: I may have missed context here, but if I understand @TomSweeneyRedHat's suggestion, should that have been to remove only the `size=` portion, but not the `noexec`?

rhatdan: I mentioned this in the PR

`Also /run is mounted on the host as exec, so no reason for us to mount
it noexec.`

edsantiago: Oh, it's in the commit message. Sorry, I missed that.

/lgtm

Issue #7285, opened 2020-08-10
systemd unit with --new fails to control or update the container


IMetZach:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**
1. Deploy my container as root:  
```
# podman create -d --name pihole --label "io.containers.autoupdate=image" --network host -e TZ="America/New York" --privileged -v "/opt/pihole/etc-pihole/:/etc/pihole/:z" -v "/opt/pihole/etc-dnsmasq.d/:/etc/dnsmasq.d/:z" --dns=127.0.0.1 --dns=1.1.1.1 --hostname pi.hole -e VIRTUAL_HOST="pi.hole" -e PROXY_LOCATION="pi.hole" -e ServerIP="127.0.0.1" pihole/pihole:latest
```  
2. Generate a systemd unit file:  
```
# podman generate systemd --new --files pihole
/root/container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service
# mv /root/container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service /etc/systemd/system/container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service
# systemctl daemon-reload
```  
3. Start the container.  
4. Notice the failure to start  
```
# systemctl start container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service
Job for container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service failed because the control process exited with error code.
See "systemctl status container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service" and "journalctl -xe" for details.
```
5. Check the output of the journal:  
```
# journalctl -xe -u container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service
-- The job identifier is 15188 and the job result is done.
Aug 10 13:47:28 pihole.redziput.lan systemd[1]: container-28850079097aa28a5bedd61c7f0b03b0a66b1fbe67f90179805e9fcd04ce23e8.service: Start request repeated too quickly.
[...]

PR #7350, opened 2020-08-18
generate systemd: quote arguments with whitespace

vrothberg:
Make sure that arguments with whitespace are properly quoted so they are
interpreted as one (and not multiple ones) by systemd.

Now `-e tz="america/new york"` will be generated as `-e "tz=america/new york"`.
The quotes are moving but the argument is still correct.

Fixes: #7285

### DIFF
--- a/pkg/systemd/generate/common.go
+++ b/pkg/systemd/generate/common.go
@@ -1,6 +1,7 @@
 package generate
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -52,4 +53,16 @@ func filterPodFlags(command []string) []string {
 		processed = append(processed, s)
 	}
 	return processed
+}
+
+// quoteArguments makes sure that all arguments with at least one whitespace
+// are quoted to make sure those are interpreted as one argument instead of
+// multiple ones.
+func quoteArguments(command []string) []string {
+	for i := range command {
+		if strings.ContainsAny(command[i], " \t") {
+			command[i] = strconv.Quote(command[i])
+		}
+	}
+	return command
 }
--- a/pkg/systemd/generate/pods_test.go
+++ b/pkg/systemd/generate/pods_test.go
@@ -75,7 +75,7 @@ Before=container-1.service container-2.service
 Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 ExecStartPre=/bin/rm -f %t/pod-123abc.pid %t/pod-123abc.pod-id
-ExecStartPre=/usr/bin/podman pod create --infra-conmon-pidfile %t/pod-123abc.pid --pod-id-file %t/pod-123abc.pod-id --name foo --replace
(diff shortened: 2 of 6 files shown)


vrothberg:
@mheon a late candidate for 2.0.

@edsantiago, I need a head nod from you before we can merge. Can you think of a case where the changes could lead to a regression?

rhatdan:
LGTM

pkg/systemd/generate/common.go
```diff
@@ -53,3 +55,18 @@ func filterPodFlags(command []string) []string {
...
+
+// regexWhitespace matches string that contain at least one whitespace.
+var regexWhitespace = regexp.MustCompile(`^.*\s.*$`)
+
+// quoteArguments makes sure that all arguments with at least one whitespace
+// are quoted to make sure those are interpreted as one argument instead of
+// multiple ones.
+func quoteArguments(command []string) []string {
+	for i := range command {
+		if regexWhitespace.MatchString(command[i]) {
```
edsantiago, line 67: regex seems like such overkill. Did you consider `string.ContainsAny` with `"<space><tab>"`?

pkg/systemd/generate/pods_test.go
```diff
@@ -118,7 +118,7 @@ WantedBy=multi-user.target default.target`
 				StopTimeout:      10,
 				PodmanVersion:    "CI",
 				RequiredServices: []string{"container-1", "container-2"},
-				CreateCommand:    []string{"podman", "pod", "create", "--name", "foo"},
+				CreateCommand:    []string{"podman", "pod", "create", "--name", "foo", "bar=arg with space"},
```
edsantiago, line 121: I can't quickly figure out whether `strconv.Quote()` handles quotes themselves. Could you change this to `"bar=arg \"with space\""` and have the test itself guarantee that quotes are preserved?

pkg/systemd/generate/common.go
vrothberg: Good idea, I'll change!

pkg/systemd/generate/pods_test.go
vrothberg: Quotes are preserved. I added additional unit tests. Cool with these?
https://github.com/containers/podman/pull/7350/files#diff-1b78823cd7c48fb19dc371e2eb287729R32-R55

edsantiago: Yes! Nice!

vrothberg: Great, thanks!

Can you /lgtm /hold?

mheon:
Nice catch, LGTM

TomSweeneyRedHat:
LGTM
and I think #7195 is biting this too

edsantiago:
No, it's the CI hang. @vrothberg you need to rebase on master, I'm sorry.

vrothberg:
merge me :)

PR #7353, opened 2020-08-18
Cleanup handling of podman mount/unmount

rhatdan:
We should default to the user name unmount rather then the internal
name of umount.

Also User namespace was not being handled correctly. We want to inform
the user that if they do a mount when in rootless mode that they have
to be first in the podman unshare state.


rhatdan:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1857542

rhatdan:
This is a back port from upstream

mheon:
I just built master and tried to `podman mount` in a `podman unshare` shell (both obviously as rootless) and got:
```Error: cannot run command "podman mount" in rootless mode```

I am pretty certain that this is still broken and these patches will not improve the situation.

mheon:
Nevermind. I'm an idiot - used the wrong Podman binary.

LGTM, but CI is red.

TomSweeneyRedHat:
LGTM

PR #7354, opened 2020-08-18
Re-disable sdnotify tests to try to fix CI

edsantiago:
Some CI tests are hanging, timing out in 60 or 120 minutes.
I wonder if it's #7316, the bug where all podman commands
hang forever if NOTIFY_SOCKET is set?


rhatdan:
/approve
LGTM

PR #7355, opened 2020-08-18
Use environment from containers.conf

rhatdan:
podman needs to use the environment settings in containers.conf
when setting up the containers.

Also host environment variables should be relative to server side
not the client.

### DIFF
--- a/pkg/specgen/specgen.go
+++ b/pkg/specgen/specgen.go
@@ -43,6 +43,13 @@ type ContainerBasicConfig struct {
 	// image's configuration.
 	// Optional.
 	Command []string `json:"command,omitempty"`
+	// EnvHost indicates that the host environment should be added to container
+	// Optional.
+	EnvHost bool `json:"env_host,omitempty"`
+	// EnvHTTPProxy indicates that the http host proxy environment variables
+	// should be added to container
+	// Optional.
+	HTTPProxy bool `json:"httpproxy,omitempty"`
 	// Env is a set of environment variables that will be set in the
 	// container.
 	// Optional.
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -307,10 +307,11 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	// ENVIRONMENT VARIABLES
 	//
 	// Precedence order (higher index wins):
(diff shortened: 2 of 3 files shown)


rhatdan:
@mheon I believe this is a better fix then https://github.com/containers/podman/pull/7343

pkg/specgen/specgen.go
```diff
@@ -43,6 +43,13 @@ type ContainerBasicConfig struct {
 	// image's configuration.
 	// Optional.
 	Command []string `json:"command,omitempty"`
+	// EnvHost indicates that the host environment should be added to container
+	// Optional.
+	EnvHost bool `json:"env_host,omitempty"`
```
mheon, line 48: Why are we sourcing this from the server, instead of the host? This doesn't make sense, IMO

mheon:
This does not solve the problem #7343 is trying to solve - ensuring that $PATH and $TERM are always populated in containers we make.

pkg/specgen/specgen.go
rhatdan: If you are on a MAC or Windows box you want this from the server.  The Proxies are setup on the Server, Proxies on the Client might not work while you are on the server machines.

rhatdan:
PATH will always be set.  If we need TERM set always the change should go into containers common. Not be hard coded in Podman.

mheon:
I don't see any guarantee that we set a default $PATH if one is not provided by the image.

mheon:
And IMO, we should just not support `--env=host` on Windows and OS X - it makes no sense to source from the server, that's not what the description of the flag tells me it will do and it feels like a security issue, leaking server environment into containers.

edsantiago:
rebase on master >= 748e8829d to fix CI

rhatdan:
Looking at the code --env-host and --http-proxy is already blocked from podman-Remote calls.

TomSweeneyRedHat:
LGTM

rhatdan:
@mheon @baude @giuseppe @vrothberg @QiWang19 @ashley-cui @jwhonce PTAL
This is ready to go in.

QiWang19:
LGTM

mheon:
Wait a moment, I thought we agreed to drop `env=host` from specgen?

rhatdan:
env=host was dropped from podman-remote.

PR #7356, opened 2020-08-18
[2.0] vendor c/image v5.5.2

vrothberg:
Enable pagination until the search result reaches the limit, instead of
returning default 100 limit from registry API.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1866153

### DIFF
--- a/cmd/podman/images/search.go
+++ b/cmd/podman/images/search.go
@@ -103,10 +103,6 @@ func imageSearch(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("search requires exactly one argument")
 	}
 
-	if searchOptions.Limit > 100 {
-		return errors.Errorf("Limit %d is outside the range of [1, 100]", searchOptions.Limit)
-	}
-
 	// TLS verification in c/image is controlled via a `types.OptionalBool`
 	// which allows for distinguishing among set-true, set-false, unspecified
 	// which is important to implement a sane way of dealing with defaults of
--- a/docs/source/markdown/podman-search.1.md
+++ b/docs/source/markdown/podman-search.1.md
@@ -59,7 +59,7 @@ Valid placeholders for the Go template are listed below:
 
 **--limit**=*limit*
 
-Limit the number of results. This value can be in the range between 1 and 100. The default number of results is 25.
+Limit the number of results (default 25).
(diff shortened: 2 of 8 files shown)


vrothberg:
@baude @mheon @QiWang19 PTAL

mheon:
LGTM

QiWang19:
LGTM

TomSweeneyRedHat:
LGTM
May have missed the PR in my disaster of an inbox, but is this going into upstream too?

QiWang19:
Do we need  to remove the CLI check in this PR? Or I can open a PR after this vendor PR gets merged.
https://github.com/containers/podman/blob/1deb4d1d70efb6d62f4fe5e735c94523f930b6d7/cmd/podman/images/search.go#L106-L108

mheon:
This needs a rebase on latest master, I think.

mheon:
Ah, wait, this is against 2.0 - interesting.

mheon:
Hm. 
`[+0868s] Error: container_linux.go:349: starting container process caused "error adding seccomp rule for syscall socket: requested action matches default action of filter": OCI runtime error`

Exclusively on runc... I think we solved this with a runc bump on master?

rhatdan:
Yup I am seeing the same failure on my v2.0 tests.

@cevich PTAL

vrothberg:
> Do we need to remove the CLI check in this PR? Or I can open a PR after this vendor PR gets merged.
> 
> https://github.com/containers/podman/blob/1deb4d1d70efb6d62f4fe5e735c94523f930b6d7/cmd/podman/images/search.go#L106-L108

Good catch, @QiWang19! I blindly vendored but can fix it here directly :+1:

cevich:
> Yup I am seeing the same failure on my v2.0 tests.
> @cevich PTAL

Yep, I'm on it.  I think #7365 might fix this, but it's just a guess.  I don't think anyone is aware of the actual underlying cause, but I was told it only affects Ubuntu VMs, Bent tried replacing runc, and so the other main difference is an older golang version (on this branch).  If that doesn't work out...sheesh, I dunno where to turn next :confused:

vrothberg:
Rebased on @baude's fixes. Thanks!

PR #7357, opened 2020-08-18
image sign using per user registries.d

QiWang19:
Support per user  ~/.config/containers/registries.d to allow rootless image sign configurations.

### DIFF
--- a/docs/source/markdown/podman-image-sign.1.md
+++ b/docs/source/markdown/podman-image-sign.1.md
@@ -9,7 +9,9 @@ podman-image-sign - Create a signature for an image
 ## DESCRIPTION
 **podman image sign** will create a local signature for one or more local images that have
 been pulled from a registry. The signature will be written to a directory
-derived from the registry configuration files in /etc/containers/registries.d. By default, the signature will be written into /var/lib/containers/sigstore directory.
+derived from the registry configuration files in `$HOME/.config/containers/registries.d` if it exists,
+otherwise `/etc/containers/registries.d` (unless overridden at compile-time), see **containers-registries.d(5)** for more information.
+By default, the signature will be written into `/var/lib/containers/sigstore` for root and `$HOME/.local/share/containers/sigstore` for non-root users
 
 ## OPTIONS
 
@@ -38,7 +40,8 @@ Sign the busybox image with the identify of foo@bar.com with a user's keyring an
 ## RELATED CONFIGURATION
 
 The write (and read) location for signatures is defined in YAML-based
-configuration files in /etc/containers/registries.d/.  When you sign
+configuration files in /etc/containers/registries.d/ for root,
+or $HOME/.config/containers/registries.d for non-root users.  When you sign
 an image, Podman will use those configuration files to determine
 where to write the signature based on the the name of the originating
 registry or a default storage value unless overridden with the --directory
@@ -52,6 +55,9 @@ When signing an image preceded with the registry name 'privateregistry.example.c
 the signature will be written into sub-directories of
 /var/lib/containers/sigstore/privateregistry.example.com. The use of 'sigstore' also means
 the signature will be 'read' from that same location on a pull-related function.
+
+## SEE ALSO
+containers-registries.d(5)
 
 ## HISTORY
 November 2018, Originally compiled by Qi Wang (qiwan at redhat dot com)
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -26,16 +26,12 @@ import (
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	domainUtils "github.com/containers/podman/v2/pkg/domain/utils"
 	"github.com/containers/podman/v2/pkg/rootless"
-	"github.com/containers/podman/v2/pkg/trust"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/storage"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
-
-// SignatureStoreDir defines default directory to store signatures
-const SignatureStoreDir = "/var/lib/containers/sigstore"
 
 func (ir *ImageEngine) Exists(_ context.Context, nameOrID string) (*entities.BoolReport, error) {
 	_, err := ir.Libpod.ImageRuntime().NewFromLocal(nameOrID)
@@ -707,12 +703,6 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 	sc := ir.Libpod.SystemContext()
 	sc.DockerCertPath = options.CertDir
 
-	systemRegistriesDirPath := trust.RegistriesDirPath(sc)
-	registryConfigs, err := trust.LoadAndMergeConfig(systemRegistriesDirPath)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error reading registry configuration")
-	}
-
 	for _, signimage := range names {
 		err = func() error {
 			srcRef, err := alltransports.ParseImageName(signimage)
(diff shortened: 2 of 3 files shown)


rhatdan:
@QiWang19 Rebase and you should be able to pass the tests.

docs/source/markdown/podman-image-sign.1.md
```diff
@@ -9,7 +9,8 @@ podman-image-sign - Create a signature for an image
 ## DESCRIPTION
 **podman image sign** will create a local signature for one or more local images that have
 been pulled from a registry. The signature will be written to a directory
-derived from the registry configuration files in /etc/containers/registries.d. By default, the signature will be written into /var/lib/containers/sigstore directory.
+derived from the registry configuration files in /etc/containers/registries.d for root user,
+or $HOME/.config/containers/registries.d for non-root user. By default, the signature of root user will be written into /var/lib/containers/sigstore directory.
```
rhatdan, line 13: Where will signatures be written for non root users?

$HOME/.local/share/containers/sigstore?

QiWang19: non-root users should create their YAML files under the $HOME/.config/containers/registries.d. Do we provide the default config file like default.yaml?

rhatdan: Non root users are not going to know how to do this, so we need to setup good defaults.  podman should automatically detect that there is no default.yaml, and create it and store signatures in $HOME/.local/share/containers/sigstore.  If default.yaml exists, then it will follow those rules.  I would guess.

QiWang19: Set the default sigstore to `$HOME/.local/share/containers/sigstore` if the config file is using `/var/lib/containers/sigsore`. Also hard code this path if rootless user has no configurations at all.

QiWang19:
@mtrmac @rhatdan PTAL

rhatdan:
LGTM
@mtrmac PTAL

rhatdan:
@mtrmac @vrothberg PTAL

docs/source/markdown/podman-image-sign.1.md
```diff
@@ -38,7 +40,7 @@ Sign the busybox image with the identify of foo@bar.com with a user's keyring an
 ## RELATED CONFIGURATION
 
 The write (and read) location for signatures is defined in YAML-based
-configuration files in /etc/containers/registries.d/.  When you sign
+configuration files in /etc/containers/registries.d/ for root user, or $HOME/.config/containers/registries.d for non-root user.  When you sign
```
vrothberg, line 43:
```suggestion
configuration files in /etc/containers/registries.d/ for root, or $HOME/.config/containers/registries.d for non-root users.  When you sign
```

docs/source/markdown/podman-image-sign.1.md
```diff
@@ -9,7 +9,9 @@ podman-image-sign - Create a signature for an image
 ## DESCRIPTION
 **podman image sign** will create a local signature for one or more local images that have
 been pulled from a registry. The signature will be written to a directory
-derived from the registry configuration files in /etc/containers/registries.d. By default, the signature will be written into /var/lib/containers/sigstore directory.
+derived from the registry configuration files in `$HOME/.config/containers/registries.d` if exists, otherwise `/etc/containers/registries.d` (unless overridden at compile-time), see **containers-registries.d(5)** for more information.
+By default, the signature will be written into `/var/lib/containers/sigstore` from `/etc/containers/registries.d/default.yaml`, if the rootless user has no privilege to write the signature, the YAML format files
+under `$HOME/.config/containers/registries.d` can be configured, otherwise, `$HOME/.local/share/containers/sigstore` will be used.
```
vrothberg, line 14: The comment doesn't seem to match the code. We don't seem to check for permissions but use the rootless paths if they exist even for root.

vrothberg (commented):
I really want a head nod from @mtrmac. I may very well miss details/side effects I am may not be aware of.

pkg/trust/trust.go
```diff
@@ -77,14 +85,18 @@ func DefaultPolicyPath(sys *types.SystemContext) string {
...
 func RegistriesDirPath(sys *types.SystemContext) string {
 	systemRegistriesDirPath := "/etc/containers/registries.d"
-	if sys != nil {
-		if sys.RegistriesDirPath != "" {
-			return sys.RegistriesDirPath
-		}
-		if sys.RootForImplicitAbsolutePaths != "" {
-			return filepath.Join(sys.RootForImplicitAbsolutePaths, systemRegistriesDirPath)
-		}
+	userRegistriesDir := filepath.FromSlash(".config/containers/registries.d")
```
mtrmac, line 88: Should these two be a top-level commented variables like the others?

pkg/domain/infra/abi/images.go
```diff
@@ -727,24 +724,25 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
-				sigStoreDir, err = localPathFromURI(sigStoreURI)
-				if err != nil {
-					return nil, errors.Wrapf(err, "invalid signature storage %s", sigStoreURI)
+			}
+			if sigStoreURI == "" {
+				// set the default sigstore for rootless or root if it is not configured
+				if rootless.IsRootless() {
+					sigStoreURI = trust.RootlessSignatureStoreURL()
+				} else {
+					sigStoreURI = trust.DefaultSignatureStoreURL()
```
mtrmac, line 739: Dealing with URLs here seems to be an unnecessary complication; move the `localPathFromURI` call to the `registryInfo != nil` case, and this can just work with the `sigStoreDir` path.

mtrmac (commented):
I’d prefer for this to be merged only after or together with https://github.com/containers/image/pull/1035 to make sure the two are consistent.

_Right now_, the code LGTM, though the documentation does not match the code and needs updating.

---

The path lookup logic is complex enough that it probably makes sense to export `c/image/docker.configuredSignatureStorageBase` (with a better name) now, and use it here instead of maintaining a second copy of the code. (While overall we do want  readers/writers to use the c/image API to access signatures instead of raw access, this has clearly not stopped `podman sign`, and modifying `podman sign` to use a `c/image/docker` `ImageDestination` would now incompatibly introduce a requirement to be able to connect to the registry.)

rhatdan:
@QiWang19 This needs a rebase.  Is this still being worked on?

QiWang19:
@rhatdan still being worked on, waiting to get this commit from c/image https://github.com/containers/image/commit/1a0dda734e3bed323a9674e274b379cf841fe8b1

rhatdan:
@QiWang19 you should have the containers/image PR in podman now.

QiWang19:
@mtrmac @vrothberg PTAL

pkg/trust/trust.go
```diff
@@ -76,15 +83,17 @@ func DefaultPolicyPath(sys *types.SystemContext) string {
...
-		if sys.RegistriesDirPath != "" {
-			return sys.RegistriesDirPath
-		}
-		if sys.RootForImplicitAbsolutePaths != "" {
-			return filepath.Join(sys.RootForImplicitAbsolutePaths, systemRegistriesDirPath)
-		}
+	if sys != nil && sys.RegistriesDirPath != "" {
+		return sys.RegistriesDirPath
+	}
+	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
```
rhatdan, line 89: Should there be a rootless check here?

QiWang19: To keep consistent with part of the logic of c/image [newConfigWrapper ](https://github.com/containers/image/blob/f1a6f9b50d0286bf208935d4b9ab48117294ac31/pkg/sysregistriesv2/system_registries_v2.go#L351), if per-user dir exists, direclty use it, without checking whether its rootless or not

rhatdan: But what happens if the directory does not exists and you are rootless.  I would figure we still want to create it.

QiWang19: you mean need to create the $HOME/.config/containers/registries.conf.d if rootless? This  RegistriesDirPath only used by `podman image trust show` to see if the configuration for specific registry exist, so we don't change it if no  configuration exists. `podman image sign` will create the place for signature storage.

rhatdan: Ok I thought that was what was calling this.

QiWang19:
@mtrmac PTAL

QiWang19:

docs/source/markdown/podman-image-sign.1.md
```diff
@@ -9,7 +9,9 @@ podman-image-sign - Create a signature for an image
 ## DESCRIPTION
 **podman image sign** will create a local signature for one or more local images that have
 been pulled from a registry. The signature will be written to a directory
-derived from the registry configuration files in /etc/containers/registries.d. By default, the signature will be written into /var/lib/containers/sigstore directory.
+derived from the registry configuration files in `$HOME/.config/containers/registries.d` if exists,
```
TomSweeneyRedHat, line 12:
```suggestion
derived from the registry configuration files in `$HOME/.config/containers/registries.d` if it exists,
```

docs/source/markdown/podman-image-sign.1.md
```diff
@@ -9,7 +9,9 @@ podman-image-sign - Create a signature for an image
...
+otherwise `/etc/containers/registries.d` (unless overridden at compile-time), see **containers-registries.d(5)** for more information.
```
TomSweeneyRedHat, line 13: looks like we don't list `containers-registries-d(5)` at the bottom of this man page and I think we should.

mtrmac (commented):
These code changes LGTM.

To be consistent, though, I think this PR should also update `getPolicyShowOutput`  the other user of `trust.LoadAndMergeConfig`+`trust.HaveMatchRegistry` (and then delete them), so that that function also correctly shows the new built-in default paths.

A downside to that is that that function will now always show a sigstore, even for X-R-S-S registries (and there isn’t a way to tell whether a registry supports X-R-S-S without connecting to it), but that’s probably better than not showing a default sigstore path that would get used.

QiWang19:
> These code changes LGTM.
> 
> To be consistent, though, I think this PR should also update `getPolicyShowOutput`, the other user of `trust.LoadAndMergeConfig`+`trust.HaveMatchRegistry` (and then delete them), so that that function also correctly shows the new built-in default paths.
> 

replace the above with `docker.SignatureStorageBaseURL` needs pass in an ImageReference, is there a good way to form an imageReference from the transports configurations in policy.json?

mtrmac:
> > These code changes LGTM.
> > To be consistent, though, I think this PR should also update `getPolicyShowOutput`, the other user of `trust.LoadAndMergeConfig`+`trust.HaveMatchRegistry` (and then delete them), so that that function also correctly shows the new built-in default paths.
> 
> replace the above with `docker.SignatureStorageBaseURL` needs pass in an ImageReference, is there a good way to form an imageReference from the transports configurations in policy.json?

Ouch… no. There is one case that works ( a `policy.json` scope that includes a tag or digest); all the rest does not:
- For a host[:port]/…/repo scope, it would be possible to look up e.g. …/repo:latest, but that might not be correct if :latest has a different configuration section from the rest of the repo.
- For a host[:port], it is impossible to look up without inventing a repo+tag/digest, and that again could clash with a more specific configuration section.
- In the worst case, there will be new support for *.domain.com and the like in `policy.json`, so there could be multiple equally right answers if a.domain.com and b.domain.com are configured with different sigstores.

The last one is essentially unfixable given the current `trust.Policy` data structure and CLI output, so I suppose returning nothing if the existing code doesn’t find an exact match is as close as we can get.

mtrmac (approved):
LGTM.

umohnani8:
Changes LGTM, @QiWang19 needs a rebase though

QiWang19:
Needs lgtm label.

umohnani8:
/lgtm
/hold

@mheon should we hold for the github-check failure?

mheon:
We can't do anything about the RDO check, but it's not a blocking failure, we can merge without it.
/hold cancel

Issue #7360, opened 2020-08-18
podman-remote exec -i, with stdin redirection, hangs


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman-remote run -d --name foo alpine sleep 20
60a88de15c87d44872b20f2716cd529de4ca0471527151456049dbc3dfeaa927

$ ./bin/podman-remote exec -i foo cat <<<"hi there"
hi there
<<<--- hangs here (no prompt, no nothing) until container sleep finishes. ^D does nothing.
```
Without `-i` there is no hang, but there is no output either.

Regular (non-remote) podman works fine: output is as expected, and the `exec` returns immediately.

master @ 748e8829d, root & rootless

vrothberg:
@mheon, do you have cycles to tackle it? My guts tell me you'll know where to punch :)

mheon:
Maybe? I'll self-assign, and we'll see how much I can get done this week.

rhatdan:
@mheon Any movement on this?

mheon:
Not yet.

edsantiago:
Still present in ab1d3a58d4b98bc948d7344fc27c680ca326086f; removing stale-issue label.

rhatdan:
@mheon Reminder.

mheon:
I will try and get to this in January.

edsantiago:
Issue still present in master @ a47515008b3c0dde9f161faa0d9f455ca6772393

mheon:
I wonder if Conmon 2.0.26 has resolved this - will re-test with it.

rhatdan:
I am assuming this is fixed in Main, reopen if I am mistaken.

edsantiago:
Unfortunately, this is not fixed. There is a very simple reproducer, it is still failing on master. There are two system tests being skipped right now with "FIXME: pending #7360", they hang too. Reopening.

christian-korneck:
also seems to happen when using the `docker` cli over ssh to a podman server (3.0.1)

```
$ echo uname | docker -H ssh://<server> exec -i <container> sh
Linux
[hangs until aborted...]
```

docker cli: 20.10.5, build 55c4c88, macOS 11.2.3 x86_64
podman: 3.0.1, fedora 33 x86_64

edsantiago:
[...]

Issue #7337, opened 2020-08-17
podman load -i Path-With-Caps label : invalid reference format: repository name must be lowercase


edsantiago:
Only when `--format=oci-archive`:
```console
$ podman pull alpine
$ podman save --format oci-archive alpine:latest >| /tmp/FOO.tar
$ podman load -i /tmp/FOO.tar foo                                                                                                              bats (2)
Error: error pulling "foo": unable to pull dir:/tmp/FOO.tar: error determining pull goal for image "dir:/tmp/FOO.tar": error parsing dest reference name "localhost/tmp/FOO.tar": error parsing named reference "localhost/tmp/FOO.tar": invalid reference format: repository name must be lowercase
```
Only the exact sequence above fails. The following are fine:
* `podman save` without `--format oci-archive`
* `podman load -i /tmp/FOO.tar` without `foo` (label)
* `podman save` to an all-lower-case path

EDIT: meant to add that this also fails if any directory component has caps: `/tmp/FOO/bar.tar`

master @ 47108e2e0

rhatdan:
@QiWang19 PTAL

zhangguanzhang:
on Centos8 and use the code of master to compile, it works ok
```
[root@centos8 ~]# ./podman/bin/podman images
REPOSITORY               TAG     IMAGE ID      CREATED      SIZE
docker.io/library/nginx  alpine  736a7d71fd38  5 weeks ago  23.1 MB
[root@centos8 ~]# ./podman/bin/podman save --format oci-archive docker.io/library/nginx:alpine > /tmp/FOO.tar
[root@centos8 ~]# ./podman/bin/podman rmi docker.io/library/nginx:alpine
Untagged: docker.io/library/nginx:alpine
Deleted: 736a7d71fd384010a9a6b3ed7fda1aad73f510fb596d08c783ccdba21b6bf3f5
[root@centos8 ~]# ./podman/bin/podman load -i /tmp/FOO.tar
Getting image source signatures
Copying blob 3eee30c545e4 done  
Copying blob 01a49528d795 done  
Copying blob 88a5e8859cbc done  
Copying blob aace0d553d4a done  
Copying blob 77ca9fc22518 done  
Copying config 736a7d71fd done  
Writing manifest to image destination
Storing signatures
Loaded image(s): docker.io/library/nginx:alpine
[root@centos8 ~]# ./podman/bin/podman images
REPOSITORY               TAG     IMAGE ID      CREATED      SIZE
[...]

Issue #7273, opened 2020-08-10
podman --help: spurious "Usage of remote" in stderr


edsantiago:
```console
$ podman --help
Usage of remote:
  -r, --remote
Manage pods, containers and images
....
```
The "Usage of remote" is jarring and unnecessary. It also goes to `stderr`, not stdout. `git bisect` blames 98da2fa80657fecb6b083eaf68974d2ee2f84e6a. @jwhonce PTAL

PR #7362, opened 2020-08-18
system tests: enable more remote tests; cleanup

edsantiago:
info, images, run, networking tests: remove some skip_if_remote()s
that were added in the varlink days. All of these tests now seem
to work with APIv2.

help test: check that first output line from 'podman --help'
is the program description (regression check for #7273).

load test: clean up stray images, rewrite test to make it conform
to existing convention. In the process, discover and file #7337

exec test (and networking): file #7360, and add FIXME comment
to skip()s suggesting evaluating those tests once that is fixed.

pod test: now that #6328 is fixed, use 'podman pod inspect --format'
instead of relying on jq

Various other tests: add an explanation of why test is disabled
so we can more easily distinguish "this will never be meaningful
under remote" vs "hey, doesn't work for now, but maybe someday".

### DIFF
--- a/test/system/005-info.bats
+++ b/test/system/005-info.bats
@@ -3,8 +3,6 @@
 load helpers
 
 @test "podman info - basic test" {
-    skip_if_remote "capitalization inconsistencies"
-
     run_podman info
 
     expected_keys="
@@ -28,8 +26,6 @@ runRoot:
 }
 
 @test "podman info - json" {
-    skip_if_remote "capitalization inconsistencies"
-
     run_podman info --format=json
 
     expr_nvr="[a-z0-9-]\\\+-[a-z0-9.]\\\+-[a-z0-9]\\\+\."
--- a/test/system/010-images.bats
+++ b/test/system/010-images.bats
@@ -75,8 +75,6 @@ Size        | [0-9]\\\+
 }
 
 @test "podman images - filter" {
-    skip_if_remote "podman commit -q is broken in podman-remote"
-
     run_podman inspect --format '{{.ID}}' $IMAGE
     iid=$output
 
--- a/test/system/015-help.bats
+++ b/test/system/015-help.bats
@@ -178,6 +178,14 @@ function check_help() {
     # Called with no args -- start with 'podman --help'. check_help() will
     # recurse for any subcommands.
     check_help
+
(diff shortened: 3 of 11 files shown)


rhatdan:
LGTM
I like more tests.

QiWang19:
LGTM

edsantiago:
This is really alarming. The failure is:
```
# $ podman load -i /tmp/podman_bats.VjElw1/myimage-5bJ5lXXL.tar
...
# $ podman images -a --format {{.ID}} {{.Repository}}:{{.Tag}}
# dd8b0a90b986 quay.io/libpod/alpine_labels:latest
# #|     FAIL: Image ID of loaded image == original
# #| expected: '4fab981df737'
# #|   actual: 'dd8b0a90b986'
```
That is: the image restored by `podman load` has the _same tag_, but _different image ID_. It is failing consistently across fedora and ubuntu. The returned image ID is always the same: `dd8b0`.

I was able to reproduce this once, only once, on my laptop just now. I then tried to `git bisect`, thinking it might've been something pulled into master today ... and nope, now I can't reproduce it on my laptop at all. 

Possibly very stupid question: is there any possibility that this may have been a time- or network-dependent failure? That there was something in the ether for an hour or two, causing these tests to fail in CI and on my setup, but then it cleared up? What exactly goes into the image ID outside of values in the tarball itself?

TomSweeneyRedHat:
@edsantiago it feels like the image somehow got refreshed sometime after the archive was made, but before the check was done.  I'm not sure how that could happen though looking at the code path.

edsantiago:
'podman save --format oci-archive' is the culprit. Filed: #7371

edsantiago:
This is ready for review. Big change since initial submission is in `120-load.bats`, it still does a load but does not compare IID (due to #7371). I'm still not 100% comfortable with this test as it is, because it's going to leave the system in a funky state: the IID of `alpine_labels`, after running this test, will differ from the actual IID of `alpine_labels` as pulled from `quay.io`. I'm totally OK with someone saying "don't do that", I could cobble up something that save/loads a generated image instead. If there is belief that #7371 will be fixed in a reasonable time, then I'm OK with just waiting instead.

mheon:
LGTM on my side

PR #7372, opened 2020-08-19
podman: add option --cgroup-conf

giuseppe:
it allows to manually tweak the configuration for cgroup v2.
    
we will expose some of the options in future as single options (e.g. the new memory knobs), but for now add the more generic
--cgroup-conf mechanism for maximum control on the cgroup configuration.

OCI specs change: https://github.com/opencontainers/runtime-spec/pull/1040

Requires: https://github.com/containers/crun/pull/459

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -104,6 +104,10 @@ The **split** option splits the current cgroup in two sub-cgroups: one for conmo
 
 Path to cgroups under which the cgroup for the container will be created. If the path is not absolute, the path is considered to be relative to the cgroups path of the init process. Cgroups will be created if they do not already exist.
 
+**--cgroup-conf**=*KEY=VALUE*
+
+When running on cgroup v2, specify the cgroup file to write to and its value.  For example **--cgroup-conf=memory.high=1073741824** sets the memory.high limit to 1GB.
+
 **--cidfile**=*file*
 
 Write the container ID to *file*.
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -516,5 +516,10 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 		"seccomp-policy", "default",
 		"Policy for selecting a seccomp profile (experimental)",
 	)
+	createFlags.StringSliceVar(
+		&cf.CgroupConf,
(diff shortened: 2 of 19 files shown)


docs/source/markdown/podman-run.1.md
```diff
@@ -909,6 +909,10 @@ The following example maps uids 0-1999 in the container to the uids 30000-31999
 
 Ulimit options. You can use **host** to copy the current configuration from the host.
 
+**--unified**=*KEY=VALUE*
```
AkihiroSuda, line 912: `--cgroup-unified` might be more descriptive

mheon:
Should we look into exposing any commonly-set versions of this as specific CLI options? This doesn't seem like a very easy/intuitive flag to use, but some of the advanced settings are going to be useful and should probably be documented independently.

giuseppe:
> Should we look into exposing any commonly-set versions of this as specific CLI options? This doesn't seem like a very easy/intuitive flag to use, but some of the advanced settings are going to be useful and should probably be documented independently.

yes definitely.  We will need to expose some of them in a more user friendly way.  This is just to make sure we have full control on the cgroup if required, since the OCI specs permit it

rhatdan:
I would prefer this to be hidden for now. since we have no idea what would be useful or not, or I guess we could state it as experimental.  We also need to know if the oci runtime supports it.

docs/source/markdown/podman-run.1.md
TomSweeneyRedHat: I concur, especially since we have cgroupns and crgoup-parent already.

giuseppe:
> I would prefer this to be hidden for now. since we have no idea what would be useful or not, or I guess we could state it as experimental. We also need to know if the oci runtime supports it.

I am fine with hiding the feature for now, or marking it experimental but the idea is that we don't want to know what each option really means.  The OCI runtime will take care of that.  `--unified` is to give complete control over the cgroup configuration.  Today for example it is not possible to configure the hugetlb controller because we are not exposing the knobs for doing that.

rhatdan:
@giuseppe Why don't we add a option for that then?  I can understand future proofing the tool, but if there is a primary feature of cgroups available in the kernel, we should expose an option for that controller.  That would greatly increase the number of people who would use the option.  Do we set sane defaults for Hugetlb.

PR #7382, opened 2020-08-19
Move pod jobs to parallel execution

mheon:
Make Podman pod operations that do not involve starting containers (which needs to be done in a specific order) use the same parallel operation code we use to make `podman stop` on large numbers of containers fast. We were previously stopping containers in a pod serially, which could take up to the timeout (default 15 seconds) for each container - stopping 100 containers that do not respond to SIGTERM would take 25 minutes.

To do this, refactor the parallel operation code a bit to remove its dependency on libpod (damn circular import restrictions...) and use parallel functions that just re-use the standard container API operations - maximizes code reuse (previously each pod handler had a separate implementation of the container function it performed).

This is a bit of a palate cleanser after fighting CI for two days - nice to be able to return to a land of sanity.

### DIFF
--- a/pkg/parallel/parallel.go
+++ b/pkg/parallel/parallel.go
@@ -1,6 +1,7 @@
 package parallel
 
 import (
+	"context"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


pkg/parallel/parallelctr/ctr.go
```diff
@@ -0,0 +1,42 @@
+package parallelctr
```
jwhonce, line 1: Why stutter name on enclosing directory and file?

mheon: Because I am not very creative and could not think of a better name.

mheon:
This should (finally) go green now

mheon:
It's green. @baude @rhatdan @vrothberg @TomSweeneyRedHat @jwhonce PTAL

pkg/parallel/parallel.go
```diff
@@ -42,3 +43,32 @@ func SetMaxThreads(threads uint) error {
...
+
+		err := fn()
+
+		jobControl.Release(1)
+
+		retChan <- err
+	}()
+
+	return retChan
+}
```
TomSweeneyRedHat, line 74: Really neat function.  At first I was looking all over above for the channel close.

TomSweeneyRedHat:
LGTM

PR #7386, opened 2020-08-20
fix podman load oci-archive's tar with name

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/7337
fixes this problem:
```
$ podman pull alpine
$ podman save --format oci-archive alpine:latest > /tmp/FOO.tar
$ podman load -i /tmp/FOO.tar foo
Error: error pulling "foo": unable to pull dir:/tmp/FOO.tar: error determining pull goal for image "dir:/tmp/FOO.tar": error parsing dest reference name "localhost/tmp/FOO.tar": error parsing named reference "localhost/tmp/FOO.tar": invalid reference format: repository name must be lowercase
```

### DIFF
--- a/libpod/image/pull.go
+++ b/libpod/image/pull.go
@@ -201,6 +201,24 @@ func (ir *Runtime) pullGoalFromImageReference(ctx context.Context, srcRef types.
 		}, nil
 
 	case OCIArchive:
+		// imgName such as "/tmp/FOO.tar:" or "/tmp/FOO.tar:domain.com/foo:tag1"
+		imgName := strings.TrimSuffix(srcRef.StringWithinTransport(), ":")
+		imgNameSli := strings.SplitN(imgName, ":", 2)
+		if len(imgNameSli) == 2 {
+			// Fixes: https://github.com/containers/podman/issues/7337
+			// use the empty name to test the manifest's length at first
+			testSrcRef, err := ociarchive.NewReference(imgNameSli[0], "")
+			if err != nil {
+				return nil, err
+			}
+			_, err = ociarchive.LoadManifestDescriptor(testSrcRef)
+			if err == nil {
+				// err is equal nil, this means the manifest's length is 1
+				// so could use the name override it
+				return ir.getSinglePullRefPairGoal(testSrcRef, imgNameSli[1])
+			}
+		}
+
 		// retrieve the manifest from index.json to access the image name
 		manifest, err := ociarchive.LoadManifestDescriptor(srcRef)
 		if err != nil {
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -452,7 +452,11 @@ func (ir *ImageEngine) Load(ctx context.Context, opts entities.ImageLoadOptions)
 	if !opts.Quiet {
 		writer = os.Stderr
 	}
-	name, err := ir.Libpod.LoadImage(ctx, opts.Name, opts.Input, writer, opts.SignaturePolicy)
+	iName := opts.Name
+	if opts.Tag != "" {
+		iName += ":" + opts.Tag
+	}
+	name, err := ir.Libpod.LoadImage(ctx, iName, opts.Input, writer, opts.SignaturePolicy)
 	if err != nil {
 		return nil, err
 	}
--- a/test/system/120-load.bats
+++ b/test/system/120-load.bats
@@ -51,8 +51,7 @@ verify_iid_and_name() {
     run_podman images $fqin --format '{{.Repository}}:{{.Tag}}'
     is "$output" "$fqin" "image preserves name across save/load"
 
-    # FIXME: when/if 7337 gets fixed, load with a new tag
-    if false; then
+    # load with a new tag, when src file is oci-archive
     local new_name=x$(random_string 14 | tr A-Z a-z)
     local new_tag=t$(random_string 6 | tr A-Z a-z)
     run_podman rmi $fqin
@@ -60,7 +59,6 @@ verify_iid_and_name() {
     run_podman load -i $archive $fqin
     run_podman images $fqin --format '{{.Repository}}:{{.Tag}}'
     is "$output" "$fqin" "image can be loaded with new name:tag"
-    fi
 
     # Clean up
     run_podman rmi $fqin


rhatdan:
@mtrmac @vrothberg PTAL

test/system/120-load.bats
```diff
@@ -38,9 +47,26 @@ verify_iid_and_name() {
...
+    verify_name localhost/foo:test
+    run_podman rmi localhost/foo:test
+
+    run_podman load -i $archive test.com/foo
+    verify_name test.com/foo:latest
+    run_podman rmi test.com/foo:latest
+
+    run_podman load -i $archive test.com/foo:test
+    verify_name test.com/foo:test
+    run_podman rmi test.com/foo:test
```
edsantiago, line 67: This is a nice set of tests, but they're actually unrelated to the purpose of this particular test block ("save to pipe"). They would belong better in a separate `@test`. I would also like to see an explicit test for CAPS in the archive name. So maybe something like:
```bats
@test "podman save/load: oci-archive" {
    # FIXME: remove this if #7371 gets fixed
    safe_archive=$PODMAN_TMPDIR/restoreme.tar
    run_podman save -o $safe_archive $IMAGE

    archive=$PODMAN_TMPDIR/MYIMAGE-$(random_string 8).tar

    run_podman save --format oci-archive -o $archive $IMAGE
    run_podman rmi $IMAGE

    function _load_and_verify() {
        run_podman load -i $archive "$1"
        run_podman images -a --format '{{.Repository}}:{{.Tag}}'
        is "$output" "$2" "podman load, with tag $1"
        run_podman rmi "$2"
    }

    _load_and_verify "foo" "localhost/foo:latest"
    _load_and_verify "foo:test" "localhost/foo:test"
    _load_and_verify "test.com/foo" "test.com/foo:latest"
    _load_and_verify "test.com/foo:test" "test.com/foo:test"

    # Restore original test image with real IID (FIXME: #7371)
    run_podman load -i $safe_archive
}

Comment (zhangguanzhang) [reply test/system/120-load.bats:67]: I got it

**Comment by zhangguanzhang** [comment]
ci has some errors, I will try to fix it

File: libpod/image/pull.go [hunk 2]
```diff
@@ -176,6 +176,24 @@ func (ir *Runtime) pullGoalFromImageReference(ctx context.Context, srcRef types.
 		}, nil
 
 	case OCIArchive:
+		// imgName such as "oci-archive:/tmp/FOO.tar:localhost/foo:latest"
+		// or "oci-archive:/tmp/FOO.tar:domain.com/foo:tag1", "oci-archive:/tmp/FOO.tar"
+		imgNameSli := strings.SplitN(imgName, ":", 3)
```
mtrmac, line 181: - Don’t use `imgName`, except maybe for the destination; start with the well-typed `srcRef`.
- `oci-archive:` paths can include `:` on Windows.

(Eventually this should get a `docker/archive.Reader`-like interface similar to the concept of #6811 . But that’s much more work.)

libpod/image/pull.go
```diff
@@ -176,6 +176,24 @@ func (ir *Runtime) pullGoalFromImageReference(ctx context.Context, srcRef types.
...
+		if len(imgNameSli) == 3 {
+			// Fixes: https://github.com/containers/podman/issues/7337
```
mtrmac, line 183: - How does this relate to #7337 _at all_, when the previous version did not use the path component of `srcRef` for `dest` either AFAICS?
- Why is fixing the single-image case enough? Can’t multi-image archives use upper-case names as well?

libpod/runtime_img.go
```diff
@@ -283,14 +283,12 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
 			return dockerarchive.ParseReference(inputFile) // FIXME? We should add dockerarchive.NewReference()
 		},
 		func() (types.ImageReference, error) {
-			return ociarchive.NewReference(inputFile, name) // name may be ""
```
mtrmac, line 286: - This is incorrect. OCI reference annotations have no semantics other than a set of allowed characters, and it’s perfectly possible fo the annotation in the image to have no slash. It must be possible to refer to such an image.
- `oci:` and `oci-archive:` should behave consistently; this changes the two in different ways.
- I can’t see any relationship to #7337 either way. One goal per PR, please; or at least one commit per concept; or _at the very least_ write down in the PR description what else the PR does!

libpod/runtime_img.go
```diff
@@ -299,11 +297,12 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
 		func() (types.ImageReference, error) {
+			iName := name // name may be ""
 			// prepend "localhost/" to support local image saved with this semantics
-			if !strings.Contains(name, "/") {
-				return layout.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			if iName != "" && !strings.Contains(name, "/") {
+				iName = fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name)
 			}
-			return nil, nil
+			return layout.NewReference(inputFile, iName)
```
mtrmac, line 305: (In both the `ociarchive` and `layout` code above:)
- AFAICS this is unrelated to #7337 , and probably belongs to a separate PR.
- If the separate `NewReference(inputFile, name)` attempt remains (as it should), this case should _only_ try prepending `DefaultLocalRegistry`. It’s a good point that in the `""` case that shouldn’t happen, but in that case these functions should return `nil, nil`, not try to load the archive the second time using the same reference that has already failed.

pkg/domain/infra/abi/images.go
```diff
@@ -452,7 +452,8 @@ func (ir *ImageEngine) Load(ctx context.Context, opts entities.ImageLoadOptions)
 	if !opts.Quiet {
 		writer = os.Stderr
 	}
-	name, err := ir.Libpod.LoadImage(ctx, opts.Name, opts.Input, writer, opts.SignaturePolicy)
+	imgName := opts.Name + ":" + opts.Tag
+	name, err := ir.Libpod.LoadImage(ctx, imgName, opts.Input, writer, opts.SignaturePolicy)
```
mtrmac, line 456: Hum. Paging  @mheon @jwhonce 
- ⚠️  *What’s the semantics of `ImageLoadOptions`*? We must establish that and document that. (See also https://github.com/containers/podman/pull/6811#discussion_r473009515 )
- Why is it safe to change this behavior now?
- `pkg/domain/infra/tunnel.ImageEngine.Load` uses `Name`+`Tag` in yet another different way. Which one is correct?

mtrmac (changes requested):
I really don’t understand what this is fixing and how. This is the third PR with, what seems to me, wildly different approach to supposedly the same problem.

You probably understand the sitaution, after spending days on this, but without writing it down we don’t know whether we are seeing the same thing or considering the same goal (and at least I can’t spend the same days to retrace all your steps now.)

Please make a clear record for the future:
- As a part of the _code base_:
  - End-user documentation: what is the goal / intended behavior of the feature
  - Programmer documentation: What is the semantics of the involved functions / data structures.
  - (If it is not obvious, which should not be the case after the above:) How the implementation works and matches the function contracts.
- As part of this PR / process:
  - What is failing. Was the previous goal of the code incorrect? Or just incorrectly developed?
    - (#7337 only lists an error reported by the `dir:` transport. If this is focusing on `oci-archive:`, please write down what is the problem encountered when using `oci-archive:` that this is supposed to be fixing. So far I’m just guessing based on the the very different PRs and I can’t see any connection to upper-case file names.)
  - How is this PR fixing this

(Others: See https://github.com/containers/podman/pull/6811#discussion_r473009515 for some context around this, it’s not at all clear what we are doing here.)

edsantiago:
>I really don’t understand what this is fixing and how. This is the third PR with, what seems to me, wildly different approach to supposedly the same problem

@mtrmac there is strong precedent for the following to just work:
```console
# podman load -i FOO.tar myname
# podman load -i FOO.tar myname:mytag
```
[ref: existing tests](https://github.com/containers/podman/blob/d254fa4c35d81992084ddad300adf487239808db/test/system/120-load.bats#L99).

This is currently broken when `FOO.tar` is an oci-archive. As a podman user, I do not wish to care what format a tarball is in, I just want `podman load` to work as it has always worked. And I don't want to have to worry about upper-case vs lower-case tar file names. Does that address your concern?

mtrmac:
- No-one has actually written down how that fails internally. The quoted error code mentioning `dir:` is of course unrelated, it’s just the one `LoadImage` happens to report. How does it _actually_ fail, what are we fixing?
- If this is related to use of capital letters in file names, and that failing substantially similarly as the `dir:` case on creating a storage reference with capital letters (is that the case??), AFAICS that just doesn’t really match what the PRs are doing. Nothing in the existing code is using the file name in a Docker-reference context (in particular as the second parameter to `getSinglePullRefPairGoal`) already.
- The PRs somehow hint at ignoring the user-specified name (“myname” above), and using data from the archive instead, but “myname” does not use capital letters.

I’m not disputing that things are broken, and it’s entirely fair to say that I should do my own research; I’m afraid I can’t spend that time this week; and looking at how all over the place the previous PRs have been, I’d like the PR author to show the analysis that has already been done.

(It’s also quite possible I’ve just been an idiot and I’ve been overlooking an entirely obvious problem, and rejecting the clearly correct way to solve it. In that case, showing the existing analysis should hopefully also help.)

zhangguanzhang:
> * No-one has actually written down how that fails internally. The quoted error code mentioning `dir:` is of course unrelated, it’s just the one `LoadImage` happens to report. How does it _actually_ fail, what are we fixing?
> * If this is related to use of capital letters in file names, and that failing substantially similarly as the `dir:` case on creating a storage reference with capital letters (is that the case??), AFAICS that just doesn’t really match what the PRs are doing. Nothing in the existing code is using the file name in a Docker-reference context (in particular as the second parameter to `getSinglePullRefPairGoal`) already.
> * The PRs somehow hint at ignoring the user-specified name (“myname” above), and using data from the archive instead, but “myname” does not use capital letters.
> 
> I’m not disputing that things are broken, and it’s entirely fair to say that I should do my own research; I’m afraid I can’t spend that time this week; and looking at how all over the place the previous PRs have been, I’d like the PR author to show the analysis that has already been done.
> 
> (It’s also quite possible I’ve just been an idiot and I’ve been overlooking an entirely obvious problem, and rejecting the clearly correct way to solve it. In that case, showing the existing analysis should hopefully also help.)

https://github.com/containers/podman/blob/516196f09677819b72d03e068fb8094b28e273aa/libpod/runtime_img.go#L285-L318
When importing images, these methods will be tried from front to back
In fact, when we import the oci-archive format and use name coverage， neither the second method nor the third method passed normally, so the error message is the e thrown by the last method.

When the passed image name is not empty, the final method returns an error
https://github.com/containers/image/blob/22129f19a77890293540e67843f28860f5a77895/oci/layout/oci_transport.go#L180-L215
so this case will return a error
https://github.com/containers/podman/blob/master/libpod/image/pull.go#L178

zhangguanzhang:
and the title of the https://github.com/containers/podman/issues/7337 is wrong，Actually fix this problem:
```
$ podman pull alpine
$ podman save --format oci-archive alpine:latest > /tmp/FOO.tar
$ podman load -i /tmp/FOO.tar foo
Error: error pulling "foo": unable to pull dir:/tmp/FOO.tar: error determining pull goal for image "dir:/tmp/FOO.tar": error parsing dest reference name "localhost/tmp/FOO.tar": error parsing named reference "localhost/tmp/FOO.tar": invalid reference format: repository name must be lowercase
```

rhatdan:
@zhangguanzhang Are you still working on this one?

zhangguanzhang:
> @zhangguanzhang Are you still working on this one?
You can discuss this with @QiWang19 , I'm going on vacation

rhatdan:
@zhangguanzhang Thanks enjoy your vacation, I just returned.

rhatdan:
@zhangguanzhang What is going on with this?

zhangguanzhang:
> @zhangguanzhang What is going on with this?

@QiWang19 Is there any progress on your side regarding this matter recently?

libpod/image/pull.go
```diff
@@ -201,6 +201,24 @@ func (ir *Runtime) pullGoalFromImageReference(ctx context.Context, srcRef types.
 		}, nil
 
 	case OCIArchive:
+		// imgName such as "/tmp/FOO.tar:" or "/tmp/FOO.tar:domain.com/foo:tag1"
+		imgName := strings.TrimSuffix(srcRef.StringWithinTransport(), ":")
```
mtrmac, line 205: What does the `TrimSuffix` do?

libpod/image/pull.go
mtrmac: Still no answer to this.

mtrmac (changes requested):
> Even hypothetically granting the status quo of different transports behaving in different ways, and as a more limited case: either the `name` parameter to `Runtime.LoadImage` _for `oci-archive:` images_ specifies what image inside the archive to use (and in that case it should be used for _all_ OCI archives, whether single-image or multi-image), or it doesn’t (and in that case it should be ignored for _all_ OCI archives, whether single-image or multi-image. It makes no sense to me to have a parameter that means different things for single-image and multi-image archives , when most of the call stack has no idea which one it is.

Please make a clear record for the future:
- As a part of the _code base_:
  - End-user documentation: what is the goal / intended behavior of the feature
  - Programmer documentation: What is the semantics of the involved functions / data structures.
  - (If it is not obvious, which should not be the case after the above:) How the implementation works and matches the function contracts.
- As part of this PR / process:
  - What is failing. Was the previous goal of the code incorrect? Or just incorrectly developed?
    - (#7337 only lists an error reported by the `dir:` transport. If this is focusing on `oci-archive:`, please write down what is the problem encountered when using `oci-archive:` that this is supposed to be fixing. So far I’m just guessing based on the the very different PRs and I can’t see any connection to upper-case file names.)
  - How is this PR fixing this

zhangguanzhang:
> > Even hypothetically granting the status quo of different transports behaving in different ways, and as a more limited case: either the `name` parameter to `Runtime.LoadImage` _for `oci-archive:` images_ specifies what image inside the archive to use (and in that case it should be used for _all_ OCI archives, whether single-image or multi-image), or it doesn’t (and in that case it should be ignored for _all_ OCI archives, whether single-image or multi-image. It makes no sense to me to have a parameter that means different things for single-image and multi-image archives , when most of the call stack has no idea which one it is.
> 
> Please make a clear record for the future:
> 
> * As a part of the _code base_:
>   
>   * End-user documentation: what is the goal / intended behavior of the feature
>   * Programmer documentation: What is the semantics of the involved functions / data structures.
>   * (If it is not obvious, which should not be the case after the above:) How the implementation works and matches the function contracts.
> * As part of this PR / process:
>   
>   * What is failing. Was the previous goal of the code incorrect? Or just incorrectly developed?
>     
>     * (#7337 only lists an error reported by the `dir:` transport. If this is focusing on `oci-archive:`, please write down what is the problem encountered when using `oci-archive:` that this is supposed to be fixing. So far I’m just guessing based on the the very different PRs and I can’t see any connection to upper-case file names.)
>   * How is this PR fixing this

Is this feature needed or not planned to support this feature?

mtrmac:
> Is this feature needed or not planned to support this feature?

_Which_ “this”?

zhangguanzhang:
> > Is this feature needed or not planned to support this feature?
> 
> _Which_ “this”?

podman load -i /tmp/foo.tar foo:testTag

QiWang19:
#7337 seems not caused by Path-With-Caps. The issue can be regenerated even with a path using lower cases.
@zhangguanzhang Are you trying to fix the [name:tag]` inconsistencies problem #7387? From the comments there the optional argument is going to be dropped in Podman v3.0. I have no idea what is a good fix for now.
@mtrmac @vrothberg WDTY?

mtrmac:
> @zhangguanzhang Are you trying to fix the [name:tag]` inconsistencies problem #7387? From the comments there the optional argument is going to be dropped in Podman v3.0. I have no idea what is a good fix for now.
> @mtrmac @vrothberg WDTY?

For the name inconsistencies, I guess if no-one makes an executive decision one way or the other to use one of the semantics and possibly break users, we can always at least document the current behavior (both for the API and the CLI), and document both the behavior _and_ that it is deprecated. In general I don’t think that changing small aspects of the inconsistent behavior would be materially better, as long as an inconsistency remains.

OTOH… doesn’t some of the API come directly from Docker? At least that one should have clearer semantics, giving us a specific target that _must_ work. That might involve actually implementing that in a reliable way (a `nameToSearchForInArchive` / `nameToTagTheResultWithLocally` parameter that does exactly what the API means), and we could preserve the current mess for CLI users (`legacyNameParameterWithConfusingSemantics`) — or it might be just enough to push us to implement only the Docker API-compatible behavior and break the other one.

Issue #7373, opened 2020-08-19
Error message needed when creating container in pod with -p option


TomSweeneyRedHat:
Pulling this from a discussion by @Luap99 in #7364

IMO, we have to error on the podman create command. I've seen some people run into it and not realize the error.

This should not work:
```
$ podman pod create --name t1
$ podman create --pod t1 -p 8080:80 nginx
```
The create command should note the -p option when being used in a pod and at the very least issue a warning if not an error and disallow it totally.

Luap99:
Closing since this got already fixed by #7390

PR #7388, opened 2020-08-20
fix pod creation with "new:" syntax followup + allow hostname

Luap99:
Fixes: 4c75fe3f70ed ("fix pod creation with "new:" syntax")

Commit 4c75fe3f70ed passes all net options to the pod but forgot
to unset the options for the container creation. This leads to
erros when using flags like `--ip` since we tried setting
the ip on the pod and container which obviously fails.

I didn't notice the bug because we don't throw an error when
specifing port bindings on a container which joins the pods
network namespace. (#7373)

Also allow the use of `--hostname` and pass that option to the
pod and unset it for the container. The container has to use
the pods hostname anyway. This would error otherwise.

Added tests to prevent regression.

### DIFF
--- a/test/e2e/run_networking_test.go
+++ b/test/e2e/run_networking_test.go
@@ -531,8 +531,8 @@ var _ = Describe("Podman run networking", func() {
 		SkipIfRemote()
 		SkipIfRootless()
 		netName := "podmantestnetwork"
-		ipAddr := "10.20.30.128"
-		create := podmanTest.Podman([]string{"network", "create", "--subnet", "10.20.30.0/24", netName})
+		ipAddr := "10.25.30.128"
+		create := podmanTest.Podman([]string{"network", "create", "--subnet", "10.25.30.0/24", netName})
 		create.WaitWithDefaultTimeout()
 		Expect(create.ExitCode()).To(BeZero())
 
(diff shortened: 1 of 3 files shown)


Luap99:
@mheon PTAL Commit 4c75fe3 is already in the v2 branch

test/e2e/run_networking_test.go
```diff
@@ -541,4 +541,19 @@ var _ = Describe("Podman run networking", func() {
 		Expect(run.ExitCode()).To(BeZero())
 		Expect(run.OutputToString()).To(ContainSubstring(ipAddr))
 	})
+
+	It("podman run with new:pod and static-ip", func() {
+		SkipIfRemote()
+		SkipIfRootless()
+		netName := "podmantestnetwork2"
+		ipAddr := "10.20.40.128"
+		create := podmanTest.Podman([]string{"network", "create", "--subnet", "10.20.40.0/24", netName})
```
mheon, line 550: Please remove this at the end of the test - these are not namespaces or automatically removed at the end of each test.

Also, a different subnet that is less likely to collide with other tests would be a good idea.

Luap99: What subnet would you recommend? something in 172.16... or 192.168... ?

mheon: 10.0.0.0 is fine, we use it elsewhere. We do just some 10.20.x.x subnets elsewhere, though. Maybe 10.25.40.0?

Luap99: There is a test directly above which uses 10.20.30.0/24 and does not remove the network. Should I also change this test while I'm at this?

mheon: Yes please!

mheon:
Restarted two flakes.

@baude @rhatdan @vrothberg @giuseppe PTAL, would like to get this one into 2.0.5 - it's an annoying little regression from 1.9

TomSweeneyRedHat:
LGTM

PR #7376, opened 2020-08-19
Cirrus: special-case CI colon-IMG and colon-DOCS only in subject

edsantiago:
As of a few minutes ago (relative to this commit), Cirrus
defines the CIRRUS_CHANGE_TITLE envariable as "First line
of CIRRUS_CHANGE_MESSAGE"[1]. Replace all conditionals
accordingly.

 [1] https://github.com/cirruslabs/cirrus-ci-docs/commit/f8d2530c602709a24d9113691a43e6a20f7020b9

Reasoning: up until this PR, the presence of CI:IMG
or CI:DOCS *in the body* of the commit message would trigger
those magic CI code flows. This violates POLA, and actually
led to a bad PR (#7317) being merged because CI never ran.

Fixes: #7374

Signed-off-by: Ed Santiago <santiago@redhat.com>

PR #7390, opened 2020-08-20
error when adding container to pod with network information

baude:
because a pod's network information is dictated by the infra container at creation, a container cannot be created with network attributes.  this has been difficult for users to understand.  we now return an error when a container is being created inside a pod and passes any of the following attributes:

* static IP (v4 and v6)
* static mac
* ports -p (i.e. -p 8080:80)
* exposed ports (i.e. 222-225)
* publish ports from image -P


pkg/specgen/container_validate.go
```diff
@@ -37,6 +37,23 @@ func (s *SpecGenerator) Validate() error {
 		}
 	}
 
+	// Containers being added to a pod cannot have certain network attributes
+	// associated with them because those should be on the infra container.
+	if len(s.Pod) > 0 {
```
mheon, line 42: Suggest adding `&& s.NetNS.Mode == FromPod` (verifies we are not a `--net=host` container among other things)

pkg/specgen/container_validate.go
```diff
@@ -37,6 +37,23 @@ func (s *SpecGenerator) Validate() error {
...
+		if s.StaticIP != nil || s.StaticIPv6 != nil {
+			return errors.Wrap(define.ErrNetworkOnPodContainer, "static ip addresses must be defined when the pod is created")
+		}
+		if s.StaticMAC != nil {
+			return errors.Wrap(define.ErrNetworkOnPodContainer, "MAC addresses must be defined when the pod is created")
+		}
+		if len(s.CNINetworks) > 0 {
+			return errors.Wrap(define.ErrNetworkOnPodContainer, "networks must be defined when the pod is created")
+		}
+		if len(s.PortMappings) > 0 || len(s.Expose) > 0 || s.PublishExposedPorts {
```
mheon, line 52: I think `Expose` is technically OK, that's just for Buildah

mheon: (Adds Expose lines to images committed from the container)

TomSweeneyRedHat:
Does this handle https://github.com/containers/podman/issues/7373?

pkg/specgen/container_validate.go
```diff
@@ -37,6 +37,23 @@ func (s *SpecGenerator) Validate() error {
...
+	// associated with them because those should be on the infra container.
+	if len(s.Pod) > 0 && s.NetNS.NSMode == FromPod {
+		if s.StaticIP != nil || s.StaticIPv6 != nil {
+			return errors.Wrap(define.ErrNetworkOnPodContainer, "static ip addresses must be defined when the pod is created")
+		}
+		if s.StaticMAC != nil {
+			return errors.Wrap(define.ErrNetworkOnPodContainer, "MAC addresses must be defined when the pod is created")
+		}
+		if len(s.CNINetworks) > 0 {
+			return errors.Wrap(define.ErrNetworkOnPodContainer, "networks must be defined when the pod is created")
```
Luap99, line 50: AFAIK This can't happen. If I use `--net foobar` the container won't join the pod namespace.

test/e2e/create_test.go
```diff
@@ -542,4 +542,61 @@ var _ = Describe("Podman create", func() {
...
+		pod.WaitWithDefaultTimeout()
+		Expect(pod.ExitCode()).To(BeZero())
+
+		session := podmanTest.Podman([]string{"create", "--pod", name, "--ip", "192.168.1.2", ALPINE, "top"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).ToNot(BeZero())
+	})
+
+	It("create container in pod with mac should fail", func() {
+		name := "createwithstaticmac"
```
Luap99, line 558: same as above: `SkipIfRootless()`

TomSweeneyRedHat:
Yep, it does fix #7376 
LGTM
assuming happy test

test/e2e/create_test.go
```diff
@@ -542,4 +542,61 @@ var _ = Describe("Podman create", func() {
...
+		Expect(session.ExitCode()).ToNot(BeZero())
+	})
+
+	It("create container in pod with network should fail", func() {
+		name := "createwithnetwork"
+		pod := podmanTest.RunTopContainerInPod("", "new:"+name)
+		pod.WaitWithDefaultTimeout()
+		Expect(pod.ExitCode()).To(BeZero())
+
+		session := podmanTest.Podman([]string{"create", "--pod", name, "--network", "foobar", ALPINE, "top"})
```
Luap99, line 574: also needs: `SkipIfRootless()`
Does foobar exists?
If we join another network we also should be allowed to pass `-p 80:80` again. So this might be worth to test.

pkg/specgen/container_validate.go
mheon: Theoretically people could pass this via the REST API, so this would catch that.

Luap99: Ah, didn't knew you can set the netns separately via the api.

PR #7416, opened 2020-08-23
[CI:DOCS] fix swagger api docs

Luap99:
Separate the volume endpoints into compat and libpod,
as it is done for the other endpoints.

Move the libpod image push endpoint to images.


rhatdan:
@mheon @jwhonce @baude PTAL

baude:
lgtm, @jwhonce ptal

rhatdan:
@mheon @jwhonce @ashley-cui PTAL

PR #7428, opened 2020-08-24
Ensure pod REST API endpoints include ctr errors

mheon:
The APIv2 pod endpoints that operate on multiple containers, such as Start, Kill, Pause, Unpause, do not report errors encountered by individual containers, because they incorrectly assume that any error is fatal. The documentation for the Libpod API calls notes, however, that ErrPodPartialFail will *always* be returned if any container failed; so we need to ignore that error and continue to collating and returning container errors.

### DIFF
--- a/pkg/api/handlers/libpod/pods.go
+++ b/pkg/api/handlers/libpod/pods.go
@@ -135,8 +135,8 @@ func PodStop(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	var errs []error //nolint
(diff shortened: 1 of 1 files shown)


mheon:
This is likely only a partial solution, we're seeing JSON unmarshal errors with individual container errors after this; @baude is on the case there.

edsantiago:
It's #7195. Restarting.

mheon:
Rebased, should go green

PR #7439, opened 2020-08-25
run test: use explicit uid/gid

edsantiago:
Fixes Fedora gating test failure: if the host tests are running
under UID 1000, --userns=keep-id will (of course) add current
user as 1000, and the in-container 'adduser' will assign 1001.
To prevent that, assign UID 4242 (and hope that that's not
our calling user's UID).

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -286,10 +286,10 @@ echo $rand        |   0 | $rand
     cid="$output"
 
     gecos="$(random_string 6) $(random_string 8)"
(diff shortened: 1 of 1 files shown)


mheon:
/approve
LGTM

rhatdan:
@edsantiago Did you make this change to master branch?

edsantiago:
no, I'd like to make it cleaner and better for master (this was an emergency). I'd like to, for instance, check that this is not the UID of the current user. I also have some other small BATS cleanup to make. PR to come today or tomorrow.

Issue #7444, opened 2020-08-25
podman images: panic


edsantiago:
I don't suppose this one will be easily reproduced. I was infinite-looping a `podman build` / `podman rmi` for purposes of reproducing #7148. I ^Ced it (in the middle of `build`), ran `podman images`, and boom:
```console
STEP 4: RUN /bin/echo hello
hello
^C
# ../bin/podman images
ERRO[0000] error reading image "11a9efdcc082f29e92296c1fe71d91d96bc198e7681d8a23955a1b93303fc079" as image: error locating item named "manifest" for image with ID "11a9efdcc082f29e92296c1fe71d91d96bc198e7681d8a23955a1b93303fc079": file does not exist
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe0 pc=0x12a96c2]

goroutine 1 [running]:
github.com/containers/podman/v2/libpod/image.areParentAndChild(0xc000189d40, 0x0, 0xc00038f350)
        libpod/image/image.go:1249 +0x32
github.com/containers/podman/v2/libpod/image.(*layerTree).children.func1(0xc0000bc480, 0x268e36cbaa7147e7, 0xc000189d40, 0x0)
        libpod/image/layer_tree.go:127 +0xae
github.com/containers/podman/v2/libpod/image.(*layerTree).children.func2(0xc000184340, 0x1ddfa00, 0x0, 0x0)
        libpod/image/layer_tree.go:135 +0xbd
github.com/containers/podman/v2/libpod/image.(*layerTree).children(0xc00049cb80, 0x1ddfa00, 0xc00038f350, 0xc0004fa900, 0x0, 0x0, 0x0, 0x0, 0x1da0900, 0xc000201000)
        libpod/image/layer_tree.go:153 +0x253
github.com/containers/podman/v2/libpod/image.(*Runtime).IntermediateFilter.func1(0xc0004fa900, 0x0)
        libpod/image/filters.go:43 +0x61
github.com/containers/podman/v2/libpod/image.FilterImages(0xc000446240, 0x4, 0x4, 0xc00051fa28, 0x1, 0x1, 0xc00000e7e0, 0x0, 0x0)
        libpod/image/filters.go:145 +0x77
[...]

PR #7448, opened 2020-08-25
fix panic when checking len on nil object

baude:
issue #7444 describes a problem where an image does not have a manifest file and cannot be processed by our library correctly.  the origin of the panic is because we are checking the len of a nil object's attribute.  this is a temporary fix to protect from the panic in the future.  the origin of the problem is more interesting and requires more work when the code author returns from pto.


mheon:
LGTM

edsantiago:
LGTM, and I appreciate your having saved me the effort of applying the patch on my test VM.

rhatdan:
@baude Please rebase to fix flakes.

vrothberg:
Added a backport label (Cc @mheon )

PR #7452, opened 2020-08-25
Refactor API build endpoint to be more compliant

### LINKED ISSUE #7136: podman-remote: build: does not emit 'STEP'
Also, other output is out of order, and includes unexpected cruft.

In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ printf "FROM alpine\nRUN echo hi\n" >Containerfile
$ ./bin/podman-remote build -t foo -f Containerfile .
hi               <----- what is this doing here? I expected it much farther down below
Copying blob sha256:50644c29ef5a27c9a40c393a73ece2479de78325cae7d762ef3cdc19bf42dd0a
Copying blob sha256:1846819bbfecae935016434cedca6048bea8129532da628c1b5a88c29a77ea66
Copying config sha256:44c079bfdef6a5d51438995fa9dce6e70add2204c2d2dacfc0165e0379d9e650
Storing signatures
--> 44c079bfdef
44c079bfdef6a5d51438995fa9dce6e70add2204c2d2dacfc0165e0379d9e650

FROM alpine          <--- expected: STEP 1
RUN echo hi          <--- expected: STEP 2
COMMIT foo           <--- expected: STEP 3
Successfully built 44c079bfdef6a5d51438995fa9dce6e70add2204c2d2dacfc0165e0379d9e650
```
Comparison against non-remote:
```console
$ ./bin/podman build -t foo -f Containerfile .
STEP 1: FROM alpine
STEP 2: RUN echo hi
hi
STEP 3: COMMIT foo
--> fa3f90c1de0
fa3f90c1de0be86d636cc70d4ca5d817c9d2292f7cb52a681a0c27afdaf30c52
```

### LINKED ISSUE #7137: podman-remote: build: does not handle errors
`podman-remote build`, if there are errors, will happily exit 0 and with no output.

In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ printf "FROM alpine\nRUN false\n" > Containerfile
$ ./bin/podman-remote build -t foo -f Containerfile .
$ echo $?
0
```
Expected:
```console
$ ./bin/podman build -t foo -f Containerfile .
STEP 1: FROM alpine
STEP 2: RUN false
Error: error building at STEP "RUN false": error while running runtime: exit status 1
$ echo $?
125
```
(bonus: there is no output in the `system server` window, either)

### LINKED ISSUE #7602: macos: podman 2.0.6 remote support is broken
While recent versions of podman improved the remoting  support they also introduced some critical bugs, like

```bash
export CONTAINER_HOST=ssh://root@example.com/run/podman/podman.sock
podman build .
# returns success (0), no output, but it failed to do anything
```

Trying to increase verbosity has not effects.

Both client (macos) and remote (fedora-32) are using same version of podman 2.0.6

jwhonce:
* Refactor channel.WriteCloser() to encapsulate the channel
* Refactor endpoint to "live" stream buildah output channels over API
  rather then buffering output
* Refactor bindings/tunnel build code for endpoint changes
* Cleanup initiating extra image engine
* Restore build system tests

Fixes #7136
Fixes #7137
Fixes #7602


rhatdan:
Can you either add tests or turn on remote tests to guarantee the two referenced issues, are permanently fixed.

edsantiago:
There are lots of disabled tests in `test/system/070-build.bats`. You can search for 7136/7137, remove those skips, and run via:
```console
$ PODMAN=$(pwd)/bin/podman-remote bats test/system/070-build.bats
```
(setting up a system service is left as an exercise for the reader).

As of this writing, the tests all fail. One common factor is that the new code is not spitting out a `COMMIT` line (my tests expect to see `COMMIT`). There's another few weird errors, I'll let you sort those out.

pkg/bindings/images/build.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// Build creates an image using a containerfile reference
+func Build(ctx context.Context, containerFiles []string, tarfile io.Reader, options entities.BuildOptions) (*entities.BuildReport, error) {
+	conn, err := bindings.GetClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	params := url.Values{}
+	params.Set("dockerfile", containerFiles[0])
```
TomSweeneyRedHat, line 28: Not for  this review, just for rumination for later, could we change this to containerfile?

TomSweeneyRedHat:
Lint is not your friend atm @jwhonce

pkg/bindings/images/build.go
jwhonce: @TomSweeneyRedHat the parameter name here is part the API, changing it would break compatibility.

jwhonce:
@rhatdan @edsantiago Remote tests restored.

test/system/070-build.bats
```diff
@@ -32,9 +31,8 @@ EOF
 
 # Regression from v1.5.0. This test passes fine in v1.5.0, fails in 1.6
 @test "podman build - cache (#3920)" {
-    skip_if_remote "FIXME: pending #7136"
-    if is_remote && is_rootless; then
-        skip "unreliable with podman-remote and rootless; #2972"
+    if is_rootless; then
```
edsantiago, line 34: I don't remember the reason for this one, but if I read the logic correctly, and if you've really fixed this for remote (safe bet, since the original skip was varlink) -- shouldn't you just remove the `if/skip` entirely? As it was, it would run rootless-local, and now it won't.

edsantiago:
@jwhonce you might want to rebase to pick up @mheon's fix for 7195. I'm not saying CI will fail, but yeah, I'm saying CI will likely fail. Sounds like a mafia threat.

edsantiago:
Tests are unhappy. I'm seeing a spurious `$PACKER_BASE/fedora_packaging.sh &&   ` (and similar strings) which closely match other stuff I'm reviewing on another repo, so now I don't know if my browser is corrupt. Are you seeing those strings too in the build test logs? Anyhow, build tests are failing.

edsantiago:
ok this is eerie. The string `bash $PACKER_BASE/fedora_packaging.sh` appears in  https://github.com/jwhonce/libpod/blob/0975f694c390f7b7f72a46493be081a1ee895d74/contrib/cirrus/packer/fedora_setup.sh#L22

How can that be interfering with a `podman build` ??

test/system/070-build.bats
```diff
@@ -145,7 +135,7 @@ echo "\$1"
 printenv | grep MYENV | sort | sed -e 's/^MYENV.=//'
 EOF
 
-    # For overriding with --env-file
+    # For overridding with --env-file
```
edsantiago, line 138: Typo introduced: "overriding" was/is correct

test/system/070-build.bats
```diff
@@ -194,8 +184,14 @@ EOF
                build_test
     is "${lines[0]}" "$workdir" "container default command: pwd"
     is "${lines[1]}" "$s_echo"  "container default command: output from echo"
+
     is "${lines[2]}" "$s_env1"  "container default command: env1"
-    is "${lines[3]}" "$s_env2"  "container default command: env2"
+
+    # remote client environment is not forwarded to host
+    if ! is_remote; then
```
edsantiago, line 191: I think it would be useful to make this an if-else, so we actually confirm that podman-remote doesn't somehow manage to start sending envariables:
```bash
if is_remote; then
    is "${lines[3]}" "this-should-be-overridden-by-env-host" "podman-remote does not send local environment"
else
    ...
```

test/system/070-build.bats
```diff
@@ -206,7 +202,11 @@ EOF
               printenv http_proxy https_proxy ftp_proxy
     is "${lines[0]}" "http-proxy-in-env-file"  "env-file overrides env"
     is "${lines[1]}" "https-proxy-in-env-file" "env-file sets proxy var"
-    is "${lines[2]}" "ftp-proxy-from-env"      "ftp-proxy is passed through"
+
+    # remote client environment is not forwarded to host
+    if ! is_remote; then
```
edsantiago, line 207: Ditto (in this case, check that output is `""`)

edsantiago (changes requested):
Spelling nit, and two suggestions for tests

test/system/070-build.bats
jwhonce: IDE dictionary or spelling check is broken on that word.  I'll revert.

edsantiago (approved):

ashley-cui (commented):
Tested against #7602, LGTM

PR #7456, opened 2020-08-26
spec: modprobe fuse with --device .*/fuse

giuseppe:
If the container uses the /dev/fuse device, attempt to load the fuse
kernel module first so that nested containers can use it.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1872240

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -231,6 +231,10 @@ Note: if the user only has access rights via a group, accessing the device
 from inside a rootless container will fail. The **crun**(1) runtime offers a
 workaround for this by adding the option **--annotation run.oci.keep_original_groups=1**.
 
+Podman may load kernel modules required for using the specified
+device. The devices that podman will load modules when necessary are:
+/dev/fuse.
+
 **--device-cgroup-rule**=rule
 
 Add a rule to the cgroup allowed devices list
--- a/pkg/specgen/generate/config_linux.go
+++ b/pkg/specgen/generate/config_linux.go
@@ -248,6 +248,13 @@ func addDevice(g *generate.Generator, device string) error {
 		}
 		g.Config.Mounts = append(g.Config.Mounts, devMnt)
 		return nil
+	} else if src == "/dev/fuse" {
+		// if the user is asking for fuse inside the container
+		// make sure the module is loaded.
+		f, err := unix.Open(src, unix.O_RDONLY|unix.O_NONBLOCK, 0)
+		if err == nil {
+			unix.Close(f)
+		}
 	}
 	dev.Path = dst
 	g.AddDevice(*dev)
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -215,6 +215,10 @@ Note: if the user only has access rights via a group, accessing the device
 from inside a rootless container will fail. The **crun**(1) runtime offers a
 workaround for this by adding the option **--annotation run.oci.keep_original_groups=1**.
 
+Podman may load kernel modules required for using the specified
+device. The devices that podman will load modules when necessary are:
+/dev/fuse.
+
 **--device-cgroup-rule**="type major:minor mode"
 
 Add a rule to the cgroup allowed devices list. The rule is expected to be in the format specified in the Linux kernel documentation (Documentation/cgroup-v1/devices.txt):


rhatdan:
LGTM
@mheon @vrothberg @TomSweeneyRedHat @baude @QiWang19 @ashley-cui @jwhonce PTAL

mheon:
This needs documentation somewhere - the manpages, definitely, but maybe more than that? And definitely a debug logging message. We should make it obvious that this command can alter system configuration.

pkg/specgen/generate/config_linux.go
```diff
@@ -225,6 +226,11 @@ func addDevice(g *generate.Generator, device string) error {
 	if err != nil {
 		return err
 	}
+	if src == "/dev/fuse" {
+		// if the user is asking for fuse inside the container
+		// make sure the module is loaded.
+		exec.Command("modprobe", "fuse").Run()
```
mheon, line 232: Should we make this conditional on being root? I don't think rootless will have the ability to modprobe

mheon: Ah, is it that this function only happens as root, because rootless bind-mounts devices?

TomSweeneyRedHat: And does this change the host or the container?  I thought the problem was the host side and I'd be hesitant to change the kernel on the host without some kind of confirmation from the user.

giuseppe:
> This needs documentation somewhere - the manpages, definitely, but maybe more than that? And definitely a debug logging message. We should make it obvious that this command can alter system configuration.

we are already doing it for overlay, aufs and devmapper.

pkg/specgen/generate/config_linux.go
giuseppe: rootless containers need it too, but I agree it is pointless to try it since they miss permissions for doing it.  Moving to run only as root.

rhatdan:
Is there a way we can trigger this automatically in Podman.  Currently the kernel loads some kernel modules automatically on first use of a device.  SELinux is blocking this for the container.

I wonder if podman just opened and closed the /dev/fuse device, if it would trigger the load and this would work in rootless and rootfull mode.  It also would be less invasive then this change.

I believe the issue here is that the confined container is the first on on the system to use /dev/fuse, so container_t process triggers a kernel module load, which is blocked.

If podman triggers the load then everything is happy.

vrothberg:
@giuseppe @rhatdan is this good to merge?

mheon:
I'm still really iffy about this one. It feels incorrect for us to be making changes like this without explicit user request...

rhatdan:
I would prefer to check if having podman attempt to use the device would cause the kernel to load the module rather then loading the module directly.  This would then fix this problem in both rootless and rootfull mode.

rhatdan:
I believe if podman just opened and closed the device it would trigger the kernel loading the fuse module.  Here is an experiment I
did to get this to happen.

```
# modprobe -r fuse
# lsmod | grep fuse
# su dwalsh
$ lsmod | grep fuse
$ cat /dev/fuse
cat: /dev/fuse: Operation not permitted
$ lsmod | grep fuse
fuse                  139264  1
```

rhatdan:
Perhaps we should just run through the list of devices handed to the container and just open and close them, ignoring errors.
This would then trigger the kernel to load any kernel modules attached to the devices.  It would work in rootful and rootless mode, and would not be subject to SELinux blocking the access inside of the container.

mheon:
If that works, it seems safe enough and won't hurt performance much.

docs/source/markdown/podman-run.1.md
```diff
@@ -231,6 +231,8 @@ Note: if the user only has access rights via a group, accessing the device
 from inside a rootless container will fail. The **crun**(1) runtime offers a
 workaround for this by adding the option **--annotation run.oci.keep_original_groups=1**.
 
+When running as root, Podman may load kernel modules required for using the specified device.
```
TomSweeneyRedHat, line 234: IDK, but think we might want to list the modules we might load.  Same thought for create.  Maybe:
```suggestion
When running as root, Podman may load kernel modules required for using the specified device.  The modules that Podman will load when necessary are:  /dev/fuse
```

giuseppe:
thinking more of it, it seems we are just trying to workaround selinux.  Do you think it could be controlled by a boolean whether containers can load modules on demand (assuming anyway they have access to the device) or simply change the policy to not block such request in any case?

rhatdan:
I really do  not want to change the SELinux, because it has supported this for years.  We don't have fine grained control over what modules the kernel will load. We either allow it or we do not.  I believe that this could lead to a container causing the kernel to load a vulnerable kernel module that the container could then use to attack the kernel. Have you attempted to just open and close the /dev/fuse device by podman to see if this fixes the issue.

giuseppe:
> Have you attempted to just open and close the /dev/fuse device by podman to see if this fixes the issue.

yes that works but it still seems like a workaround.  What should we do in the case of `-v /dev:/dev`?  Should we attempt to open recusrively every file under `/dev`?

We could probably just improve the error message in fuse-overlayfs and detect when `fuse` is not loaded

rhatdan:
No we don't want to walk the /dev, I agree fixing the issue with fuse-overlay would be good,  I originally thought of  just doing to Open/Close on all devices added with --dev, but someone pointed out that this could lead to other side effects that we do not want.

rhatdan:
I believe we should special case the /dev/fuse just because we document this to users, and this can easily happen on a fresh booted system.

giuseppe:
> I believe we should special case the /dev/fuse just because we document this to users, and this can easily happen on a fresh booted system.

I am fine if we special case `/dev/fuse`, I was mostly worried about the cost of doing it for every device and how to handle the recursive case.  I've changed the patch to open the file instead of attempting modprobe.

pkg/specgen/generate/config_linux.go
```diff
@@ -248,6 +248,13 @@ func addDevice(g *generate.Generator, device string) error {
 		}
 		g.Config.Mounts = append(g.Config.Mounts, devMnt)
 		return nil
+	} else if src == "/dev/fuse" {
+		// if the user is asking for fuse inside the container
+		// make sure the module is loaded.
+		f, err := unix.Open(src, unix.O_RDONLY|unix.O_NONBLOCK, 0)
+		if err != nil {
```
rhatdan, line 255: Shouldn't this be 
if err == nil {
	unix.Close(f)
}

docs/source/markdown/podman-run.1.md
rhatdan: This should be 
The devices that podman will load modules when necessary are: /dev/fuse

giuseppe: I dropped "When running as root@ as the device is opened as rootless as well

pkg/specgen/generate/config_linux.go
giuseppe: yes!  Thanks for noticing it

rhatdan:
LGTM

vrothberg (commented):
LGTM

PR #7457, opened 2020-08-26
[CI:DOCS] Making docs build on mac

ashley-cui:
sed syntax on mac is different

### DIFF
--- a/docs/remote-docs.sh
+++ b/docs/remote-docs.sh
@@ -88,6 +88,16 @@ function pub_pages() {
     done
 }
 
+## sed syntax is different on darwin and linux
+## sed --help fails on mac, meaning we have to use mac syntax
+function sed_os(){
+    if sed --help > /dev/null 2>&1 ; then
+        $(sed -i "$@")
+    else
+        $(sed -i "" "$@")
+    fi
+}
+
 ## rename renames podman-remote.ext to podman.ext, and fixes up contents to reflect change
 function rename (){
     if [[ "$PLATFORM" != linux ]]; then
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

docs/remote-docs.sh
```diff
@@ -88,21 +88,30 @@ function pub_pages() {
     done
 }
 
+## sed syntax is different on darwin and linux
+function sed_os(){
+    if [[ "$OSTYPE" == "darwin"* ]]; then
```
edsantiago, line 93: When possible, I like to test for functionality rather than platform. Is there a way to, say, run `sed --help` and differentiate the Mac version from regular sed?

ashley-cui: Sure, `sed --help` doesn't work on a mac so i can base it off of the exit code for `sed --help` ?

edsantiago: That's really alarming! But yes, I think it would be safe to correlate "sed --help fails" with "this is not GNU sed, therefore it needs a special invocation of -i".

edsantiago:
Oops! Forgot to mention: if you edit the PR title, adding `[CI:DOCS]`, you receive a get-out-of-CI-hell-free card.  I don't know if you also need to change the subject of your git commit, but if you're going to re-push, might as well.

TomSweeneyRedHat:
LGTM
and happy green test buttons

TomSweeneyRedHat:
@ashley-cui close, next time try
/hold cancel

PR #7459, opened 2020-08-26
[CI:IMG] try adding tor and removing skopeo-containers

lsm5:

Trial run to check if `contrib/cirrus/packer/README.how-to-update-cirrus-vms` works.


lsm5:
@cevich is moving to using containers/automation-images. So, I'll close this..

PR #7464, opened 2020-08-26
BATS: fix corner case in --userns=keep-id test

edsantiago:
The test that does 'adduser' in a keep-id container had a
really dumb bug: if the user running the test has UID 1000,
then podman itself (via keep-id) will add the "1000" passwd
entry, and the in-container "adduser" will allocate 1001,
making our test fail. This triggered in f31/f32 podman gating
tests, but (?!?) never in rawhide gating tests.

Solution: explicitly feed a UID to adduser. Make sure that
it's not the same as the UID of the current user.

Also (unrelated): fix a ridiculous "run mkdir || die". At
the time I wrote that I probably had no idea how BATS works.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -294,11 +294,22 @@ echo $rand        |   0 | $rand
     run_podman run -d --userns=keep-id $IMAGE sh -c 'while ! test -e /stop; do sleep 0.1; done'
     cid="$output"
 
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

TomSweeneyRedHat:
LGTM
and oh so subtle.   Nice catch.

giuseppe (approved):

edsantiago:
omg tests are green! Thank you so much @mheon!

Issue #7468, opened 2020-08-27
podman  generate kube without hostAliases


zhangguanzhang:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

/kind feature

**Description**

**Steps to reproduce the issue:**

1.create pod
```
podman pod create --name test \
  --add-host test.test.com:127.0.0.1 \
  --add-host test1.test.com:127.0.0.1
podman run -d --pod test --name test2  docker.io/library/nginx:alpine
```
2.  generate kube
```
podman  generate kube test
```

**Describe the results you received:**
the `spec.hostAliases` is null

**Describe the results you expected:**
```
  hostAliases:
  - hostnames:
    - test.test.com
    ip: 127.0.0.1
  - hostnames:
    - test1.test.com
    ip: 127.0.0.1
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
$ podman --version
podman version 2.0.5
```

**Output of `podman info --debug`:**

```
(paste your output here)
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
(paste your output here)
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

Yes/No

**Additional environment details (AWS, VirtualBox, physical, etc.):**

bcstinch:
Does this fix #7442 as well? will hostAliases translate back into a podman pod from the kube yaml?

zhangguanzhang:
after https://github.com/containers/podman/pull/7469 be merged, I will take a look at https://github.com/containers/podman/issues/7442

> Does this fix #7442 as well? will hostAliases translate back into a podman pod --add-host from the kube yaml?

PR #7469, opened 2020-08-27
fix podman generate kube with HostAliases

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/7468

### DIFF
--- a/libpod/kube.go
+++ b/libpod/kube.go
@@ -48,12 +48,22 @@ func (p *Pod) GenerateForKube() (*v1.Pod, []v1.ServicePort, error) {
 		return nil, servicePorts, errors.Errorf("pod %s only has an infra container", p.ID())
 	}
 
+	extraHost := make([]v1.HostAlias, 0)
 	if p.HasInfraContainer() {
 		infraContainer, err := p.getInfraContainer()
 		if err != nil {
 			return nil, servicePorts, err
 		}
-
+		for _, host := range infraContainer.config.ContainerNetworkConfig.HostAdd {
+			hostSli := strings.SplitN(host, ":", 2)
+			if len(hostSli) != 2 {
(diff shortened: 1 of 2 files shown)


libpod/kube.go
```diff
@@ -48,19 +48,33 @@ func (p *Pod) GenerateForKube() (*v1.Pod, []v1.ServicePort, error) {
...
 			return nil, servicePorts, err
 		}
 		servicePorts = containerPortsToServicePorts(ports)
 	}
 	pod, err := p.podWithContainers(allContainers, ports)
+	if err != nil {
+		return nil, servicePorts, err
+	}
+	pod.Spec.HostAliases = extraHost
 	return pod, servicePorts, err
```
rhatdan, line 78: Can you change this to return nil, rather then err, to make it clearer to future  developers that this is success.

rhatdan:
@mheon @vrothberg @haircommander @giuseppe @mrunalp @baude PTAL

libpod/kube.go
zhangguanzhang: servicePorts []v1.ServicePort   //nolint
nolint, so I should'nt use nil, and line 47 
```
return nil, servicePorts, errors.Erro....
```

zhangguanzhang: I got it wrong🤣, now it's handled

TomSweeneyRedHat:
LGTM
and happy green test buttons.  How did that happen?

zhangguanzhang:
I think the latest code of other people is also green in ci after merging, which may be fixed by a previous submission

rhatdan:
We need someone with a better handle on Kubernetes to review this.
@mheon @vrothberg @haircommander @giuseppe @mrunalp @baude @umohnani8  PTAL

haircommander:
LGTM

umohnani8:
LGTM

Issue #7441, opened 2020-08-25
rootless: can't run systemd in a container: Failed to create control group


dustymabe:
/kind bug

**Description**

On Fedora CoreOS (cgroups v1) I'm trying to build and run a systemd
container via rootless podman and it fails:

```
[core@localhost build]$ cat Dockerfile 
FROM registry.fedoraproject.org/fedora:32
RUN dnf -y install systemd httpd
RUN systemctl enable httpd
EXPOSE 80
CMD [ "/sbin/init" ]

[core@localhost build]$ podman build --quiet -t httpd .
c19db941eee8c74ef5fff230ef5f5c292fe4f98409501e6f246a0b6f17f44485

[core@localhost build]$ podman run -it --rm --name httpd httpd
systemd v245.7-1.fc32 running in system mode. (+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=unified)
Detected virtualization podman.
Detected architecture x86-64.

Welcome to Fedora 32 (Container Image)!

Set hostname to <7f8b837350b4>.
Failed to create /user.slice/user-1000.slice/session-1.scope/init.scope control group: Permission denied
Failed to allocate manager object: Permission denied
[!!!!!!] Failed to allocate manager object.
Exiting PID 1...

[core@localhost build]$ getsebool container_manage_cgroup
container_manage_cgroup --> on
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
[core@localhost build]$ podman version
Version:      2.0.5
API Version:  1
Go Version:   go1.14.6
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64

[core@localhost build]$ rpm -q podman
podman-2.0.5-1.fc32.x86_64
```

**Output of `podman info --debug`:**

<details>

```
[core@localhost build]$ podman info --debug
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.19, commit: 5dce9767526ed27f177a8fa3f281889ad509fea7'
  cpus: 2
  distribution:
[...]

PR #7475, opened 2020-08-27
abi: trim init command

baude:
without it the systemd detection fails.

Closes: https://github.com/containers/podman/issues/7441


mheon:
LGTM

PR #7477, opened 2020-08-27
[CI:IMG] use ubuntu crun-0.14.1~3 with race fix

lsm5:


edsantiago:
This is the seccomp failure. Okay, who remembers how to fix that?

rhatdan:
Most likely you need to update the seccomp.json file and crun to match.

lsm5:
@cevich is moving to using containers/automation-images. So, I'll close this..

PR #7480, opened 2020-08-27
Switch to containers/common for seccomp

saschagrunert:
The seccomp/containers-golang library is not maintained any more and we
should stick to containers/common.


TomSweeneyRedHat:
LGTM
assuming happy tests

baude:
i dont understand this PR and what it is doing ... why do we need to add the seccomp.json to podman?  It should be provided by a more common package.  Again, this was just a glance, and i might be missing something very obvious.

saschagrunert:
> i dont understand this PR and what it is doing ... why do we need to add the seccomp.json to podman? It should be provided by a more common package. Again, this was just a glance, and i might be missing something very obvious.

Nono, I just removed the `github.com/seccomp/containers-golang` dependency in favor of a bumped containers/common, because we moved all the source code from seccomp/containers-golang to containers/common. It should not change any behavior at all.

PR #7485, opened 2020-08-27
Cirrus: Obsolete CI:IMG process & related files

cevich:
All VM-building functionality has been migrated to
https://github.com/containers/automation_images

Some container-build functions are still maintained here
but are on a very-short list to also be migrated to
the repository linked above.


cevich:
@edsantiago @rhatdan @baude @TomSweeneyRedHat PTAL, this is ready to go in.

.cirrus.yml
```diff
@@ -360,30 +342,32 @@ meta_task:
...
-
-    env:
-        <<: *meta_env_vars
-        GCPJSON: ENCRYPTED[766916fedf780cbc16ac3152f7f73c5d9dcf64768fc6e80b0858c5badd31e7b41f3c864405c814189fd340e5a056ba18]
-        GCPNAME: ENCRYPTED[d6869741209b8cf380adb8a3858cbce4542c9cf115452fcd2024a176b08fce10112e8bf0fbcc2f0033e7b87ef4342b3a]
-
-    timeout_in: 10m
-
-    script: '/usr/local/bin/entrypoint.sh |& ${TIMESTAMP}'
+# image_prune_task:
```
rhatdan, line 348: Is there a reason this is just commented out and not deleted?

TomSweeneyRedHat:
LGTM
but definitely want @edsantiago to give this a thorough review.

.cirrus.yml
cevich: I'm not 100% sure if I want to re-enable it temporarily while moving it over to `automation_images` repo.  It's quite a bit of work to move, so depending on how long that takes, it might need to be switched back on temporarily to clean out a "too large" set of stale/orphan/testing VM images.  I'm not dead-set against simply removing it, if you think it's distracting.

contrib/cirrus/lib.sh
```diff
@@ -61,22 +61,17 @@ CONTINUOUS_INTEGRATION="${CONTINUOUS_INTEGRATION:-false}"
...
-# Exists to support manual single-image building in case of emergency
-export PACKER_BUILDS="${PACKER_BUILDS:-ubuntu-20,ubuntu-19,fedora-32,fedora-31}"
-# Google cloud provides these, we just make copies (see $SCRIPT_BASE/README.md) for use
-export UBUNTU_BASE_IMAGE="ubuntu-2004-focal-v20200506"
-export PRIOR_UBUNTU_BASE_IMAGE="ubuntu-1910-eoan-v20200211"
-# Manually produced base-image names (see $SCRIPT_BASE/README.md)
-export FEDORA_BASE_IMAGE="fedora-cloud-base-32-1-6-1588257430"
-export PRIOR_FEDORA_BASE_IMAGE="fedora-cloud-base-31-1-9-1588257430"
-export BUILT_IMAGE_SUFFIX="${BUILT_IMAGE_SUFFIX:--$CIRRUS_REPO_NAME-${CIRRUS_BUILD_ID}}"
+export BUILT_IMAGE_SUFFIX="${_BUILT_IMAGE_SUFFIX:-$CIRRUS_REPO_NAME-${CIRRUS_BUILD_ID}}"
```
edsantiago, line 72: Is this default (`reponame-buildid`) still valid nowadays when the convention is 'c followed by a bignum' ?

cevich: This is some legacy garbage from when `hack/get_ci_vm.sh` didn't know how to interpret `.cirrus.yml` (now it does).  I wonder if maybe I should try getting rid of this line + `s/BUILD_IMAGE_SUFFIX/_BUILD_IMAGE_SUFFIX/` everywhere.

It'll take running the tests again, but maybe removing this legacy garbage (finally) is worth.  I'd also more seriously consider killing off the commented section Dan was concerned about (above).  I'm thinking it's more distraction than help, even if I need to re-enable it.

cevich:
Fixed both concerns, rebased, force pushed.

rhatdan:
/approve
LGTM

.cirrus.yml
```diff
@@ -151,25 +150,15 @@ gating_task:
...
-        - name: "build in_podman image ${UBUNTU_NAME}"
-          container:
-              dockerfile: Dockerfile.ubuntu
-        - name: "build gate image $DEST_BRANCH branch"
-          container:
-              dockerfile: contrib/gate/Dockerfile
-
     container:
-        dockerfile: Dockerfile
+        dockerfile: contrib/gate/Dockerfile
```
edsantiago, line 161: The string `Dockerfile.ubuntu` no longer appears anywhere in the source tree. Would it make sense to remove the top-level `Dockerfile.ubuntu` now? (It is much harder to check if it's also safe to remove the top-level `Dockerfile`, but I tried some heuristics and I think it might also be safe to remove it.

.cirrus.yml
```diff
@@ -724,16 +588,14 @@ success_task:
         - "build_without_cgo"
         - "container_image_build"
         - "meta"
-        - "image_prune"
+        #- "image_prune"
```
edsantiago, line 591: I'd feel much better if this was removed, since the corresponding `_task` was also removed

.cirrus.yml
cevich: Good catch, yes the Ubuntu dockerfile can go.  I'm afraid to remove the `Dockerfile` though, I seem to recall some build-related test (somewhere) needing at least a simple one.

edsantiago (commented):
(Commenting via review, because I can't comment on the line itself).

`.cirrus.yml` line 52 defined `BUILT_IMAGE_SUFFIX` (without a leading underscore). That seems to be part of what your PR removes; and I no longer see anything in `contrib` that references this variable. Could you remove it, to prevent future maintainer confusion?

.cirrus.yml
cevich: woops.

cevich:
> Could you remove it, to prevent future maintainer confusion?

Yes, that whole block except one line can be removed...

edsantiago:
Should you also remove the rest of the files in `contrib/cirrus/packer`, and all mentions of `$PACKER` in `contrib/cirrus/lib.sh`, and mention of `systemd_banish.sh`?

cevich:
> remove the rest of the files in contrib/cirrus/packer, and all mentions

Oh! I thought I got that...fixing...

.cirrus.yml
```diff
@@ -66,8 +56,8 @@ env:
     ####
     #### Credentials and other secret-sauces, decrypted at runtime when authorized.
     ####
-    # Freenode IRC credentials for posting status messages
-    IRCID: ENCRYPTED[0c4a3cc4ecda08bc47cd3d31592be8ae5c2bd0151bf3def00a9afd139ef1ab23a1bd0523319d076c027f9749ddb1f3c8]
```
edsantiago, line 70: Did you mean to remove this?

cevich: Ya, I oinked out all the IRC stuff. It doesn't seem to be used/noticed much.

cevich: Found some more calls into IRC stuff, and removed.  We should be good now...

edsantiago: Is/was that the "[Action Recommended]" podbot messages? Darn. Now that they're not spamflooding the channel, I've actually been checking those - they're a good source of news about flakes. Is there another place where I can get reports of those failures? (Like, is the mailing-list thing working?)

edsantiago: Anyhow, if you need to push again (please don't just for this), maybe it might be a good idea to update the commit message as well, to reflect removing the IRC stuff.

Dockerfile
```diff
@@ -1,21 +1,2 @@
...
-WORKDIR $GOSRC
-
-# Re-use repositories and package setup as in VMs under CI
-RUN bash $PACKER_BASE/fedora_packaging.sh && \
-    dnf clean all && \
-    rm -rf /var/cache/dnf
-
-# Mirror steps taken under CI
-RUN bash -c 'source $GOSRC/$SCRIPT_BASE/lib.sh && install_test_configs'
+ADD /* /var/tmp/go/src/github.com/containers/podman/
```
edsantiago, line 2: This makes it look like you know what the purpose of this file is. Since Dockerfiles allow comments, would you mind adding one such as "this top-level Dockerfile is used by XXX for purposes of YYY"?

cevich: I don't really know for sure, it's only a guess that a copy of the repo. might be useful.

I s'pose the other option here is, I could delete `Dockerfile` and then we watch and see what (if anything) breaks...

edsantiago:
Thank you for your patience with my nitpicking. This is good cleanup, I can find no other loose ends. LGTM!

cevich:
Thanks for your thoroughness :+1:

PR #7493, opened 2020-08-28
[WIP] Added shortnames feature

ParkerVR:
Linked with https://github.com/containers/common/pull/253
TODO: move this code to common, skip configure message when fully qualified name is provided

### DIFF
--- a/libpod/image/pull.go
+++ b/libpod/image/pull.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/containers/common/pkg/config/shortnames"
 	"github.com/containers/common/pkg/retry"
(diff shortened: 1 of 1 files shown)


libpod/image/pull.go
```diff
@@ -238,7 +239,17 @@ func (ir *Runtime) pullImageFromHeuristicSource(ctx context.Context, inputName s
 		if srcTransport != nil && srcTransport.Name() != DockerTransport {
 			return nil, err
 		}
+
+		fullnames, err := shortnames.Search(inputName)
```
mtrmac, line 243: The integration should happen inside `ir.pullGoalFromPossiblyUnqualifiedName`, which already starts with a check whether the input specifies a registry (and in that case the short name lookup should not be involved).

PR #7143, opened 2020-07-29
bindings: skip flaky pause/unpause test

edsantiago:
The "podman wait to pause|unpause condition" test is failing
several times a day, always a flake. Issue #6518.

Disable it until the cause can be identified and fixed.

Signed-off-by: Ed Santiago <santiago@redhat.com>

Issue #5325, opened 2020-02-25
special_testing_bindings_task, `podman remove a paused container by id with force` test fails


cevich:
/kind bug

**Description**

On F31:

```
...cut...
[+0504s] time="2020-02-25T09:50:42-05:00" level=debug msg="Paused container fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2"
[+0504s] time="2020-02-25T09:50:44-05:00" level=debug msg="APIHandler -- Method: DELETE URL: /v1.24/libpod/containers/fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2?force=true&vols=false (conn 1/4)"
[+0504s] time="2020-02-25T09:50:44-05:00" level=debug msg="Sending signal 9 to container fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2"
[+0504s] the container `fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2` is not running: No such process
[+0504s] time="2020-02-25T09:50:44-05:00" level=info msg="Request Failed(Internal Server Error): `/usr/bin/crun resume fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2` failed: exit status 1"
[+0504s] fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2
[+0504s] 
• Failure [8.980 seconds]
[+0504s] Podman containers 
[+0504s] 
/var/tmp/go/src/github.com/containers/libpod/pkg/bindings/test/containers_test.go:15
[+0504s]   
podman remove a paused container by id with force [It]
[+0504s]   
/var/tmp/go/src/github.com/containers/libpod/pkg/bindings/test/containers_test.go:177
[+0504s] 
[+0504s]   
Expected
[+0504s]       <utils.ErrorModel>: {
[+0504s]           Because: "`/usr/bin/crun resume fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2` failed: exit status 1",
[+0504s]           Message: "`/usr/bin/crun resume fad95b817abd7f418bc075d9738828d348050fd62db276676d144230e09133d2` failed: exit status 1",
[+0504s]           ResponseCode: 500,
[+0504s]       }
[+0504s]   to be nil
[+0504s] 
...
```

**Steps to reproduce the issue:**

1. `make`
2. `cd pkg/bindings/test && ginkgo -r`

**Describe the results you received:**

Test fails

[...]

PR #7514, opened 2020-08-31
bindings: reenable flaky(?) pause/unpause test

edsantiago:
Reference: #6518, a very-frequently-flaking CI test, disabled
a month ago (#7143) because it was triggering so often in CI.
Unfortunately, that seems to have simply swept the problem
under the rug. AFAICT nobody has bothered to look at the
root bug, so let's just reenable. If the problem persists,
I'll let annoyed developers squeaky-wheel 6158 so there's
some incentive to fix it. If the problem has miraculously
gone away in the last month, that's a win too.

(This test failure does not reproduce on my laptop, nor
does it lend itself to devising a simple reproducer on
a test VM.)

Also: since #5325 appears to have been closed as fixed,
remove a 'Skip' that references it. Unfortunately this
also requires removing a lot of other cruft. This was
an incidental oh-by-the-way addition that I thought
would be trivial but ended up causing a much larger diff.

### DIFF
--- a/pkg/bindings/test/containers_test.go
+++ b/pkg/bindings/test/containers_test.go
@@ -2,7 +2,6 @@ package test_bindings
 
 import (
 	"net/http"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rhatdan:
/approve
LGTM

PR #7517, opened 2020-08-31
use crio runc on CICID ubuntu

baude:
when running CICD on Ubuntu where no cgroups v2, we need to use a newer runc for things like seccomp and the default ubuntu runc
 is not new enough.


rhatdan:
/hold cancel
LGTM

PR #7525, opened 2020-09-01
[v1.9] Fix login using authconfig credentials

crevetor:
In 1.9 authconfig credentials login is broken because if no username or password was provided we use the authconfig credentials only if they don't provide a username, which is incorrect.


crevetor:
Closing this one. Will reopen with DCO.

PR #7526, opened 2020-09-01
Fix authconfig authentication in 1.9

crevetor:
In the current 1.9 branch, if you pass an authfile with valid credentials, it is ignored.

This patch fixes it by checking that the username from the authfile is not empty in order to use those credentials.

Signed-off-by: Antoine Reversat <a.reversat@gmail.com>

### DIFF
--- a/cmd/podman/login.go
+++ b/cmd/podman/login.go
@@ -131,7 +131,7 @@ func loginCmd(c *cliconfig.LoginValues) error {
 	}
 
 	// If no username and no password is specified, try to use existing ones.
-	if c.Username == "" && password == "" && authConfig.Username == "" && authConfig.Password != "" {
+	if c.Username == "" && password == "" && authConfig.Username != "" && authConfig.Password != "" {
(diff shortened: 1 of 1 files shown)


mheon:
The v1.9 branch is mostly decommissioned at this point, and we're really only fixing security issues in it. Is it not possible to update to v2.0?

crevetor:
I'm looking into upgrading to v2 but since 1.9 is what ships in RHEL 8.2 I'd suggest that, given the complexity of this patch, it would be a good idea to merge it.

rhatdan:
@crevetor We don't plan on ever updating the 1.9 branch on RHEL, the next release on RHEL will be podman 2.0.*  And is coming out in November.  If you want to get this fix into RHEL then it needs to go into the master branch.

crevetor:
@rhatdan the fix is already in the master branch since this whole part of the code got reworked. Thanks for the heads up that 1.9 won't be updated on RHEL. Do with this PR whatever you want.

Issue #7521, opened 2020-09-01
Server/service fails to detect client disconnect in stats REST API


hakonhall:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The server does not detect client disconnect with the stats REST API, taking up resources linear in the the number of historical requests.

**Steps to reproduce the issue:**

1.

```
sudo podman run -d --name c1 --entrypoint /bin/sleep docker.io/library/alpine 12345
```

2.

Start the service in terminal A:

```
sudo podman --root /var/vespa/podman system service -t 0
```

3.

In Terminal B:

```
sudo curl --unix-socket /run/podman/podman.sock http://h/containers/c1/stats
```

and then `control-c` after, say, 10s.

**Describe the results you received:**

Every 4 seconds, the server responds with stats and prints a warning:

```
http: superfluous response.WriteHeader call from github.com/containers/libpod/pkg/api/handlers/utils.WriteJSON (handler.go:64)
```

This is annoying but looks harmless. After `control-c` (in terminal B), the message is printed a few more times, the following is printed instead every 4 seconds:

```
http: superfluous response.WriteHeader call from github.com/containers/libpod/pkg/api/handlers/utils.WriteJSON (handler.go:64)
ERRO[0276] unable to write json: "write unix /run/podman/podman.sock->@: write: broken pipe" 
```

**Describe the results you expected:**

I expect the server to detect that the client has disconnected, and stop trying to send a response.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
$ podman version
Version:            1.9.3
RemoteAPI Version:  1
Go Version:         go1.13.4
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
(if needed)
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-1.9.3-2.module+el8.2.1+6867+366c07d6.x86_64
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

No and yes

**Additional environment details (AWS, VirtualBox, physical, etc.):**

(if needed)

mheon:
The API in 1.9.3 is *very* early - any chance you can retest with Podman v2.0.5?

hakonhall:
On RHEL 8 that's the latest I have available.  The description in https://podman.io/getting-started/installation would install an even older version, so it requires building from scratch.

[...]

PR #7527, opened 2020-09-01
APIv2 container stats - send header once

baude:
when sending container stats, we should only set and send the response header once.  using approach developed by jhonce for his image build refactor.

Fixes: #7521

### DIFF
--- a/pkg/api/handlers/compat/containers_stats.go
+++ b/pkg/api/handlers/compat/containers_stats.go
@@ -75,6 +75,21 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	flush := func() {
+		if flusher, ok := w.(http.Flusher); ok {
+			flusher.Flush()
+		}
+	}
+
+	// set up the header information and send it
+	w.WriteHeader(http.StatusOK)
+	w.Header().Add("Content-Type", "application/json")
+	flush()
+
+	// set up an encoder for streaming the data back
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(true)
+
 	for ok := true; ok; ok = query.Stream {
 		// Container stats
 		stats, err := ctnr.GetContainerStats(stats)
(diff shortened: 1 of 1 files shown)


baude:
@jwhonce ptal

pkg/api/handlers/compat/containers_stats.go
```diff
@@ -174,11 +189,13 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 			ID:       stats.ContainerID,
 			Networks: net,
 		}
-
-		utils.WriteJSON(w, http.StatusOK, s)
-		if flusher, ok := w.(http.Flusher); ok {
-			flusher.Flush()
+		// encode data
+		if err := enc.Encode(s); err != nil {
+			utils.Error(w, "failed stat encoding", http.StatusBadRequest, err)
```
jwhonce, line 194: You have already send the StatusOK so you cannot send StatusBadRequest here.

baude: what should i use @jwhonce

vrothberg (commented):
I don't see the connection to #7521. I would expect a fix for #7521 to check in the loop if the context has been cancelled (i.e., connection has been closed) and return.

vrothberg:
> I don't see the connection to #7521. I would expect a fix for #7521 to check in the loop if the context has been cancelled (i.e., connection has been closed) and return.

See [events](https://github.com/containers/podman/blob/master/pkg/api/handlers/compat/events.go#L120) for an example.

vrothberg:
> > I don't see the connection to #7521. I would expect a fix for #7521 to check in the loop if the context has been cancelled (i.e., connection has been closed) and return.
> 
> See [events](https://github.com/containers/podman/blob/master/pkg/api/handlers/compat/events.go#L120) for an example.

I think we should borrow the idea from events ^ which takes care of both: sending the header only once _and_ checking for closed connections.

vrothberg:
@baude, I can take over if you want to

pkg/api/handlers/compat/containers_stats.go
jwhonce: Most you can do is log error.

vrothberg:
> @baude, I can take over if you want to

As I have your permission to take over issues, I'll close it and wrap up a new PR :)

PR #7532, opened 2020-09-02
We should not be mounting /run as noexec when run with --systemd

rhatdan:
The system defaults /run to "exec" mode, and we default --read-only
mounts on /run to "exec", so --systemd should follow suit.


mheon:
LGTM

rhatdan:
@TomSweeneyRedHat @edsantiago @ashley-cui @QiWang19 @baude @jwhonce PTAL

QiWang19:
LGTM

ashley-cui:
LGTM

edsantiago:
LGTM. Confirmed via:
```console
$ /bin/podman run --read-only --rm alpine sh -c "cp /bin/date /run;/run/date"
sh: /run/date: Permission denied

$ ./bin/podman run --read-only --rm quay.io/edsantiago/testimage:20200902 sh -c "cp /bin/date /run;/run/date"
Wed Sep  2 17:37:50 UTC 2020
```
(likewise with `--systemd=always` instead of `--read-only`. Without either flag, both `/bin` and `./bin` (dot slash bin) work equally.

PR #7534, opened 2020-09-02
Migrate away from docker.io

edsantiago:
CI and system tests currently pull some images from docker.io.
Eliminate that, by:

  - building a custom image containing much of what we need
    for testing; and
  - copying other needed images to quay.io

The principal change is to create a new quay.io/libpod/testimage,
using the new test/system/build-testimage script, instead of
relying on quay.io/libpod/alpine_labels. We also switch to
using a hardcoded :YYYYMMDD tag, instead of :latest, in an
attempt to futureproof our CI. This image includes 'httpd'
from busybox-extras, which we use in our networking test
(previously we had to pull and run busybox from docker.io).

The testimage can and should be extended as needed for future
tests, e.g. adding test file content or other useful tools.

For the '--pull' tests which require actually pulling from
the registry, I've created an image with the same name but
tagged :00000000 so it will never be pulled by default.
Since this image is only used minimally, it's just busybox.

Unfortunately there remain two cases we cannot solve in
this tiny alpine-based image:

  1) docker registry
  2) systemd

For those, I've (manually) run:

    podman pull [ docker.io/library/registry:2.7 | registry.fedoraproject.org/fedora:31 ]
    podman tag !$ quay.io/...
    podman push !$

...and amended the calling tests accordingly.

I've tried to make the the smallest reasonable diff, not the
smallest possible one. I hope it's a reasonable tradeoff.


edsantiago:
Marked `WIP` because I'd like feedback on the image name: `testimage` seems too undescriptive.

Until naming is resolved, images live in my personal `quay.io` namespace instead of `quay.io/libpod`.

test/system/500-networking.bats
```diff
@@ -12,7 +12,7 @@ load helpers
     random_2=$(random_string 30)
 
     HOST_PORT=8080
-    SERVER=http://localhost:$HOST_PORT
+    SERVER=http://127.0.0.1:$HOST_PORT
```
edsantiago, line 15: This is completely weird. When running podman as root, 'localhost' no longer seems to work, at least with alpine: `wget` barfs with ECONNREFUSED. It works fine with rootless. I gave up trying to track it down, and explicit IPv4 home seems like a better idea anyway.

test/system/helpers.bash
```diff
@@ -5,9 +5,9 @@ PODMAN=${PODMAN:-podman}
 
 # Standard image to use for most tests
 PODMAN_TEST_IMAGE_REGISTRY=${PODMAN_TEST_IMAGE_REGISTRY:-"quay.io"}
-PODMAN_TEST_IMAGE_USER=${PODMAN_TEST_IMAGE_USER:-"libpod"}
-PODMAN_TEST_IMAGE_NAME=${PODMAN_TEST_IMAGE_NAME:-"alpine_labels"}
-PODMAN_TEST_IMAGE_TAG=${PODMAN_TEST_IMAGE_TAG:-"latest"}
+PODMAN_TEST_IMAGE_USER=${PODMAN_TEST_IMAGE_USER:-"edsantiago"}
+PODMAN_TEST_IMAGE_NAME=${PODMAN_TEST_IMAGE_NAME:-"testimage"}
```
edsantiago, line 9: Please offer suggestions here.

test/system/260-sdnotify.bats
```diff
@@ -116,8 +116,7 @@ function _assert_mainpid_is_conmon() {
 @test "sdnotify : container" {
     # Sigh... we need to pull a humongous image because it has systemd-notify.
     # FIXME: is there a smaller image we could use?
-    _FEDORA=registry.fedoraproject.org/fedora:31
-
+    local _FEDORA="$PODMAN_TEST_IMAGE_REGISTRY/$PODMAN_TEST_IMAGE_USER/fedora:31"
```
rhatdan, line 119: Should we update this to 32?  Or eliminate the version altogether?

test/system/helpers.bash
rhatdan: I am fine with testimage?

test/system/build-testimage
```diff
@@ -0,0 +1,59 @@
...
+
+set -ex
+
+# Please document the reason for all flags, apk's, and anything non-obvious
+#
+#    --squash-all    : needed by 'tree' test in 070-build.bats
+#    busybox-extras  : provides httpd needed in 500-networking.bats
+#
+podman rmi -f testimage &> /dev/null || true
+podman build --squash-all -t testimage - <<EOF
```
rhatdan, line 35: Would we want to switch this to --timestamp=0 or some other timestamp, to keep it consistent once this is available to Podman?

rhatdan (approved):
A few questions, and need to get rid of the edsantiago references 
but overall LGTM

test/system/build-testimage
edsantiago: I considered that but couldn't think of a reason - it'll be useful to have a number of tools built-in, and it would be **super** useful to figure out a way some day to include `registry` and systemd, but if/when that happens we'll change the `YMD` tag. Do you see value in reproducible/comparable images?

test/system/helpers.bash
edsantiago: Okay, going once...

test/system/500-networking.bats
rhatdan: Weird, but localhost is hard coded within container/image and podman to represent locally built images.

vrothberg:
@edsantiago, what's the motivation to migrate away? I assume the latest policy changes of Docker Hub but it would be nice to mention it in the commit message for historical reasons.

vrothberg (commented):
Other than Dan's comments, LGTM

edsantiago:
>what's the motivation to migrate away? I assume the latest policy changes of Docker Hub but it would be nice to mention it in the commit message for historical reasons.

Yes, Docker Hub pull limit. I updated commit message to reflect that. 

Since `testimage` seems OK, I've pushed those images to `quay.io/libpod`, made them public, updated the reference in `helpers.bash`, and pushed.

Issue #7462, opened 2020-08-26
Compatibility: Add filters for listing networks


minioin:
I tried to run k3d with podman system service but k3d returned error with ``filters for listing networks is not implemented ``. As I looked at docker [api](https://docs.docker.com/engine/api/v1.40/#operation/NetworkList), it lists filters as an option.

This features is one of the first thing I encountered while running [k3d under podman ](https://github.com/rancher/k3d/issues/84).

Luap99:
The problem is that podman handles networking differently from docker. The podman backend only supports `name` and `plugin` as valid filter keys for networks. I guess that's the reason why the api don't allow filters.

minioin:
Considering that we already need a system service for docker compatibility, how is it difficult to maintain a list of networks to implement this feature? What is blocking this api exactly?

Luap99:
@mheon stated that he is looking to improve this in the future https://github.com/containers/podman/issues/7080#issuecomment-664021523

@minioin Do you know what filter k3d is using? I think we can at least implement the name filter for now.

rhatdan:
@minioin Time is the only thing blocking us.  Getting community to help build it is always helpful.

iwilltry42:
@Luap99 , in k3d we're only filtering by name: https://github.com/rancher/k3d/blob/main/pkg/runtimes/docker/network.go#L47-L54
```Go
        // (1) configure list filters
	args := filters.NewArgs()
	args.Add("name", name)

	// (2) get filtered list of networks
	networkList, err := docker.NetworkList(ctx, types.NetworkListOptions{
		Filters: args,
	})
```

rhatdan:
@iwilltry42 @minioin  @Luap99 Can we close this issue now since your PR has been merged?

mheon:
I'm going to go ahead and close this. We're not 100% there yet, but we can open separate, more specific issues for any remaining missing filters.

PR #7536, opened 2020-09-02
APIv2 Add network list filtering

Luap99:
Add the filter option to the libpod endpoint.
Add support for the name filter on the docker endpoint.

Add apiv2 tests for the network list endpoints.
Enable podman network integration tests for remote.

### DIFF
--- a/pkg/api/server/register_networks.go
+++ b/pkg/api/server/register_networks.go
@@ -61,6 +61,11 @@ func (s *APIServer) registerNetworkHandlers(r *mux.Router) error {
 	//  - networks (compat)
 	// summary: List networks
 	// description: Display summary of network configurations
+	// parameters:
+	//  - in: query
+	//    name: filters
+	//    type: string
+	//    description: JSON encoded value of the filters (a map[string][]string) to process on the networks list. Only the name filter is supported.
 	// produces:
 	// - application/json
 	// responses:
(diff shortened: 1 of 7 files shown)


Luap99:
Ref #7462

rhatdan:
@baude @jwhonce @mheon PTAL

QiWang19:
LGTM

pkg/api/server/register_networks.go
```diff
@@ -106,7 +111,7 @@ func (s *APIServer) registerNetworkHandlers(r *mux.Router) error {
 	//    required: true
 	//    description: the name of the network
 	//  - in: query
-	//    name: Force
+	//    name: force
```
vrothberg, line 114: Nit (for cleaner potential cherry-picking): Can you break this change out into a separate commit?

vrothberg (commented):
Other than the nit, LGTM.

But I want a head nod from @baude and/or @jwhonce

pkg/api/server/register_networks.go
mheon: I wouldn't worry. We're past the point that anything should be going into 2.0 without a critical bug being behind it.

vrothberg (commented):
LGTM, given tests are passing I am good to merge

@mheon @rhatdan PTAL

PR #7533, opened 2020-09-02
[WIP]update vm images

baude:
pick up new crun and crio-runc

Signed-off-by: Brent Baude <bbaude@redhat.com>

PR #7538, opened 2020-09-03
Update VM images for new crun; adapt Cap tests to work with new kernel

edsantiago:
(This is an adoption of #7533 because Brent is on PTO).

Pick up new crun and crio-runc.

Also: renames from useful fedora-32 and -31 to less-useful
names; presumably this is needed by something-something in
the new VM setup.

Also: tweak two e2e tests to more properly handle a kernel
(5.8.4) with a greater set of capabilities than what we
or crun can yet handle.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -34,12 +34,12 @@ env:
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
     ###
-    FEDORA_NAME: "fedora-32"
-    PRIOR_FEDORA_NAME: "fedora-31"
-    UBUNTU_NAME: "ubuntu-20"
-    PRIOR_UBUNTU_NAME: "ubuntu-19"
+    FEDORA_NAME: "fedora"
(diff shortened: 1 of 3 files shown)


edsantiago:
I'm ashamed at the number of pushes I had to do. Should I ever need to deal with ginkgo again (and, sigh, it seems likely), I will take time to figure out a way to run it locally.

Test seems to work. See, e.g., [test fedora RCLI:false](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4980896967163904/html/integration_test.log.html#t--podman-privileged-CapEff--1) which shows different `CapEff` for host & container. What that _doesn't_ show is if my logic is correct, e.g. the test could just as easily pass if `ParseUint()` is returning `1` for each value. So please, dear reviewer, apply a careful eye to that code.

rhatdan:
LGTM
@baude @cevich @mheon PTAL

Issue #7513, opened 2020-08-31
ignore_chown_errors is ignored in global storage.conf


plopresti:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

ignore_chown_errors = "true" does not work if I put it in /etc/containers/storage.conf. It does work if I put it in ~/.config/containers/storage.conf.

**Steps to reproduce the issue:**

```
$ rm ~/.config/containers/storage.conf
$ cat /etc/containers/storage.conf
[storage]
  driver = "overlay"
  [storage.options]
    mount_program = "/bin/fuse-overlayfs"
    [storage.options.overlay]
      ignore_chown_errors = "true"
$ podman info | grep ignore_chown_errors
$ cp /etc/containers/storage.conf ~/.config/containers/.
$ podman info | grep ignore_chown_errors
    overlay.ignore_chown_errors: "true"
```

**Describe the results you received:**
"pm info | grep ignore_chown_errors" shows nothing even though I set ignore_chown_errors in /etc/containers/storage.conf

**Describe the results you expected:**
I expect the ignore_chown_errors setting in /etc/containers/storage.conf to be honored

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.5
API Version:  1
Go Version:   go1.13.14
Built:        Fri Aug 28 11:06:26 2020
OS/Arch:      linux/amd64

```

**Output of `podman info --debug`:**

```
$ pm info --debug
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.20-1.el7.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.20, commit: 2eef3028b91df012a014c0b8ceb3202b3d2222bb'
  cpus: 48
  distribution:
    distribution: '"rhel"'
    version: "7.8"
  eventLogger: file
  hostname: lex3.kla-tencor.com
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 100
      size: 1
    uidmap:
    - container_id: 0
      host_id: 39327
      size: 1
  kernel: 3.10.0-1127.19.1.el7.x86_64
  linkmode: dynamic
  memFree: 396886724608
  memTotal: 404129320960
  ociRuntime:
    name: runc
    package: runc-1.0.0-103.dev.el7.x86_64
    path: /usr/bin/runc
    version: |-
[...]

PR #7549, opened 2020-09-06
[2.0] Vendor in containers/storage v1.20.4

rhatdan:
This will allow the passing down of the ignore_chown_errors flag from
/etc/containers/storage.conf for rootless containers.


rhatdan:
@mheon @vrothberg @baude @giuseppe @TomSweeneyRedHat @QiWang19 @ashley-cui @jwhonce PTAL

rhatdan:
Fixes: https://github.com/containers/podman/issues/7513

vrothberg (approved):
LGTM

I think this calls for tests to prevent future regressions (Cc @edsantiago). @rhatdan, do you have an idea how we could easily test it? I guess we had to set up a new uid/gid mapping ideally with a range of 1.

rhatdan:
@vrothberg the problem with testing this is we would need to modify the host system that is running the test to make it work, which @edsantiago frowns upon.

vrothberg:
> @vrothberg the problem with testing this is we would need to modify the host system that is running the test to make it work, which @edsantiago frowns upon.

That's what I feared. Couldn't we run it in a container?

rhatdan:
Yes doing this in a privileged container might make sense.  Do we have any examples of this though?

vrothberg:
> Yes doing this in a privileged container might make sense. Do we have any examples of this though?

I think we could reuse the podman images from Quay.io and mount a handcrafted /etc/sub{g,u}id? Then we'd need an image with more than one uids/gids.

mheon:
Let's keep this as just a vendor update. If we want to add tests, we should do it against master first.

Issue #7545, opened 2020-09-05
Apparmor status incorrectly detected on Arch/Manjaro


rrichardson:
/kind bug

**Description**

When running `--security-opt apparmor=unconfined` I get this error: 
`Error: Apparmor profile "unconfined" specified, but Apparmor is not enabled on this system`

However, Apparmor is enabled on my system
```
apparmor module is loaded.
5 profiles are loaded.
5 profiles are in enforce mode.
```

So I'm guessing perhaps I am missing some flag that allows podman or the container to detect this? 
This is the command I'm executing.. Actually it's `kind` 0.8.1 that is executing it

I'm running Manjaro/Arch linux : Linux ricktop 5.4.60-2-MANJARO #1 SMP PREEMPT Sat Aug 22 14:35:41 UTC 2020 x86_64 GNU/Linux  

**Steps to reproduce the issue:**

1.  Enable Apparmor

2. Run `podman run --security-opt apparmor=unconfined fedora-minimal cat /proc/self/uid_map`

**Describe the results you received:**

`Error: Apparmor profile "unconfined" specified, but Apparmor is not enabled on this system`

**Describe the results you expected:**
Something else. 

**Additional information you deem important (e.g. issue happens only occasionally):**

This happens whether I run with or without `--privileged` and with or without `sudo`
So it seems like it's a sanity check that occurs fairly early on. 

**Output of `podman version`:**
```
➜ podman version
Version:      2.0.4
API Version:  1
Go Version:   go1.14.6
Git Commit:   cab27868766a3b01ac14ff46180af8644c41d1cd
Built:        Fri Jul 31 15:31:25 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.20, commit: 13244db638cf987c415298a3c23393ae5abeb885'
  cpus: 12
  distribution:
    distribution: manjaro
    version: unknown
  eventLogger: file
  hostname: ricktop
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 10000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 10000
      size: 65536
  kernel: 5.4.60-2-MANJARO
[...]

PR #7553, opened 2020-09-07
Fix unconfined AppArmor profile usage for unsupported systems

saschagrunert:
If we select "unconfined" as AppArmor profile, then we should not error
even if the host does not support it at all. This behavior has been
fixed and a corresponding e2e test has been added as well.

Fixes https://github.com/containers/podman/issues/7545

### DIFF
--- a/pkg/specgen/generate/security.go
+++ b/pkg/specgen/generate/security.go
@@ -60,7 +60,7 @@ func setLabelOpts(s *specgen.SpecGenerator, runtime *libpod.Runtime, pidConfig s
 func setupApparmor(s *specgen.SpecGenerator, rtc *config.Config, g *generate.Generator) error {
 	hasProfile := len(s.ApparmorProfile) > 0
 	if !apparmor.IsEnabled() {
-		if hasProfile {
+		if hasProfile && s.ApparmorProfile != "unconfined" {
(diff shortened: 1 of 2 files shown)


pkg/specgen/generate/security.go
```diff
@@ -60,8 +60,8 @@ func setLabelOpts(s *specgen.SpecGenerator, runtime *libpod.Runtime, pidConfig s
 func setupApparmor(s *specgen.SpecGenerator, rtc *config.Config, g *generate.Generator) error {
 	hasProfile := len(s.ApparmorProfile) > 0
 	if !apparmor.IsEnabled() {
-		if hasProfile {
-			return errors.Errorf("Apparmor profile %q specified, but Apparmor is not enabled on this system", s.ApparmorProfile)
+		if hasProfile && s.ApparmorProfile != "unconfined" {
+			return errors.Errorf("AppArmor profile %q specified, but it is not enabled on this system", s.ApparmorProfile)
```
vrothberg, line 64: `it` refers to the profile but not to AppArmor. Changing the error message should be in a separate commit but I suggest keeping it as is.

vrothberg (commented):
Besides the error message, LGTM :+1:

pkg/specgen/generate/security.go
saschagrunert: Yes sure, reverted that line.

PR #7554, opened 2020-09-07
make image parent check more robust

vrothberg:
Follow up on issue #7444 and make the parent checks more robust.
We can end up with an incoherent storage when, for instance, a
build has been killed.

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -1247,11 +1247,14 @@ func areParentAndChild(parent, child *imgspecv1.Image) bool {
 	// candidate parent's diff IDs, which together would have
 	// controlled which layers were used
 
-	// issue #7444 describes a panic where the length of child.RootFS.DiffIDs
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

vrothberg:
In case we backport this to 2.0 (we I think we should) we need to cherry-pick commit a6f85861df406ae7d44c25c1cbeecc1d267b9f43 first.

vrothberg:
@baude @mheon PTAL

vrothberg:
@mheon, can I backport or do we need a BZ?

mheon:
At this point, we're very late in the cycle for 8.3 - we can talk at WC tomorrow, but if you think this is necessary we definitely need a BZ.

vrothberg:
> At this point, we're very late in the cycle for 8.3 - we can talk at WC tomorrow, but if you think this is necessary we definitely need a BZ.

Thanks! I created BZ 1876576 :+1:

Issue #7384, opened 2020-08-19
`podman stop $cid1 $cid2` successfully stops both but sometimes errors that $cid2 does not exist


asottile:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

This one is pretty hard to track down and I haven't isolated a highly-reproducible scenario however I am able to observe this about ~10% of the time with my given workload.

The output below shows the failure scenario and the output of `podman container inspect $cid` for both of the containers just before running `podman stop $cid1 $cid2`

It appears that sometimes when stopping both containers podman successfully stops both but then loses track of the second container causing an error.

I can also reproduce this when running `podman stop $cid1` and `podman stop $cid2` separately

**Steps to reproduce the issue:**

I'm still working on more reproducible results but I'm currently running something similar to this in a loop and get an error every ~10 or so runs.  I will comment / edit if I find a more reproducible run and without private components.

```bash
cid1=$(podman run --rm --detach -v "$workdir:/work:rw" -v "$input:/input:ro" -v "$pc:/pc:rw" $img sleep infiinty)
cid2=$(podman run --net=none --rm --detach -v "$workdir:/work:rw" -v "$input:/input:ro" -v "$pc:/pc:rw" $img sleep infiinty)
# several podman exec -i "$cid1"... and one  podman exec "$cid2" ...
podman status "$cid1"
podman status "$cid2"
podman stop "$cid1" "$cid2"
```

**Describe the results you received:**

#### `podman inspect $cid1` (right before running stop)
<details>
<pre><code>
[
  {
    "Id": "295e59ca74f4486cf27dece22be6d7b16ef0d9e54e5053418fb7ebf910b20ac8",
    "Created": "2020-08-19T14:54:21.826048257-07:00",
    "Path": "dumb-init",
    "Args": [
      "--",
      "sleep",
      "infinity"
    ],
    "State": {
      "OciVersion": "1.0.2-dev",
      "Status": "running",
      "Running": true,
      "Paused": false,
      "Restarting": false,
      "OOMKilled": false,
      "Dead": false,
      "Pid": 4613,
      "ConmonPid": 4592,
      "ExitCode": 0,
      "Error": "",
      "StartedAt": "2020-08-19T14:54:26.251034733-07:00",
      "FinishedAt": "0001-01-01T00:00:00Z",
      "Healthcheck": {
        "Status": "",
        "FailingStreak": 0,
        "Log": null
      }
    },
[...]

PR #7557, opened 2020-09-08
podman stop: do not cleanup for auto-removal

vrothberg:
Do not perform a container clean up for containers configured for
auto-removal (e.g., via `podman run --rm`).  There is a small race
window with the other process performing the removal where a clean up
during podman-stop may fail since the container has already been removed
and cleaned up.  As the removing process will clean up the container,
we don't have to do it during podman-stop.

Fixes: #7384

### DIFF
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -174,6 +174,12 @@ func (ic *ContainerEngine) ContainerStop(ctx context.Context, namesOrIds []strin
 				return err
 			}
 		}
+		if c.AutoRemove() {
+			// Issue #7384: if the container is configured for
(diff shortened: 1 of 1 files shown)


vrothberg:
@mheon PTAL ... I couldn't reproduce the issue on my machine (I assume it's too fast)

rhatdan:
Should we continue to remove the container and then just check if the container was already removed?

vrothberg:
> Should we continue to remove the container and then just check if the container was already removed?

(*Container).CleanUp() doesn't remove the container but cleans up namespaces and other resources. The other process is already taking care of that, so it would be redundant to continue doing it in case of auto-removal.

mheon:
LGTM

PR #7559, opened 2020-09-08
Makefile: add cross compile targets for more arches

lsm5:
This commit adds additional targets for which we provide binary packages
on fedora and other distros.

@vrothberg @mheon @baude @rhatdan @TomSweeneyRedHat


lsm5:
gotta check if Cirrus. checks for these targets, I don't see linux/$other_arch listed in `cross_task`

lsm5:
> gotta check if Cirrus checks for these targets, I don't see linux/$other_arch listed in `cross_task`

ah ok, I do see a `/usr/local/bin/entrypoint.sh local-cross to ensure build success on other arches. So, I'm guessing this PR should be good to go.

@cevich PTAL.

vrothberg (commented):
LGTM

Issue #7340, opened 2020-08-17
podman-remote: run --rm: Error removing container


edsantiago:
Another podman-remote flake, probably a race condition. I don't think this is the same as #7195 because when it happens I actually get output.
```console
$ podman rm -a;podman volume rm -a
$ expect=$(./bin/podman-remote run --rm -v testvol1:/etc/apk alpine_labels ls /etc/apk)
$ while :;do got=$(./bin/podman-remote run --rm -v testvol1:/etc/apk alpine_labels ls /etc/apk); test "$got" = "$expect" || echo "got: $got";done
ERRO[0000] Error removing container 96e24969a069225db3cfdc27a724172fac29465ac911fa2974f68a94f3706b28: no container with name or ID 96e24969a069225db3cfdc27a724172fac29465ac911fa2974f68a94f3706b28 found: no such container
ERRO[0000] Error removing container e8834ac68b23e1720beed42e68b3e6de250629af7bd0295b9c9f62bf179c4912: container e8834ac68b23e1720beed42e68b3e6de250629af7bd0295b9c9f62bf179c4912 does not exist in database: no such container
ERRO[0000] Error removing container 3e4fcf50ba72d454ba81548dbfdff17d10a439001bb1a28a76f7165ede45eb4d: no container with name or ID 3e4fcf50ba72d454ba81548dbfdff17d10a439001bb1a28a76f7165ede45eb4d found: no such container
ERRO[0000] Error removing container 933bb8451d833c6c04399315edb91a8fa4900991f4f4f8f73c5cf0ac9e5f6ff9: container 933bb8451d833c6c04399315edb91a8fa4900991f4f4f8f73c5cf0ac9e5f6ff9 does not exist in database: no such container
^C
```
Note that there are two distinct error messages:
* no container with name or ID ...
* container ... does not exist in database

Exit status, FWIW, is 0 even with the error.

master @ 47108e2e0, root and rootless

I've seen it even without the `-v testvol`, but it reproduces much more quickly with `-v`. As in, less than five seconds to get an error message on my f32 laptop.

edsantiago:
CI flakes in the field:

Podman run with volumes [It] podman named volume copyup
* fedora-32 : special_testing_rootless
  * PR #7152
    * [07-31 13:34](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5869546022109184/html/integration_test.log.html#t--podman-named-volume-copyup--1)
[...]

PR #7561, opened 2020-09-08
remote run: fix error checks

vrothberg:
As error types are not preserved on the client side (due to marshaling),
we cannot use `errors.Cause(...)` and friends but, unfortunately, have
to fall back to looking for substring the error messages.

Change the error checks in remote run to do substring matches and fix
issue #7340.

Fixes: #7340

### DIFF
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -19,6 +19,7 @@ import (
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/bindings/containers"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/errorhandling"
 	"github.com/containers/podman/v2/pkg/specgen"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -537,8 +538,8 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
 		// de-spaghetti the code.
 		defer func() {
 			if err := containers.Remove(ic.ClientCxt, con.ID, bindings.PFalse, bindings.PTrue); err != nil {
-				if errors.Cause(err) == define.ErrNoSuchCtr ||
-					errors.Cause(err) == define.ErrCtrRemoved {
+				if errorhandling.Contains(err, define.ErrNoSuchCtr) ||
+					errorhandling.Contains(err, define.ErrCtrRemoved) {
 					logrus.Warnf("Container %s does not exist: %v", con.ID, err)
 				} else {
 					logrus.Errorf("Error removing container %s: %v", con.ID, err)
@@ -556,7 +557,7 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
 
 	// Determine why the wait failed.  If the container doesn't exist,
 	// consult the events.
-	if !strings.Contains(waitErr.Error(), define.ErrNoSuchCtr.Error()) {
+	if !errorhandling.Contains(waitErr, define.ErrNoSuchCtr) {
 		return &report, waitErr
 	}
 
--- a/pkg/errorhandling/errorhandling.go
+++ b/pkg/errorhandling/errorhandling.go
@@ -57,3 +57,11 @@ func CloseQuiet(f *os.File) {
 		logrus.Errorf("unable to close file %s: %q", f.Name(), err)
 	}
 }
+
+// Contains checks if err's message contains sub's message. Contains should be
+// used iff either err or sub has lost type information (e.g., due to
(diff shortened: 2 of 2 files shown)


pkg/domain/infra/tunnel/containers.go
```diff
@@ -509,31 +509,23 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
...
-				report.ExitCode = int(exitCode)
-			}
-		}
-	} else {
+	if opts.Detach {
 		err = containers.Start(ic.ClientCxt, con.ID, nil)
+	} else {
+		err = startAndAttach(ic, con.ID, &opts.DetachKeys, opts.InputStream, opts.OutputStream, opts.ErrorStream)
 	}
+	// Wait
```
rhatdan, line 517: Nit this comment should be in the else block.

rhatdan (approved):
LGTM other the tini nit.

QiWang19:
LGTM

pkg/domain/infra/tunnel/containers.go
```diff
@@ -509,31 +509,22 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
...
-			if errors.Cause(err) == define.ErrNoSuchCtr ||
-				errors.Cause(err) == define.ErrCtrRemoved {
-				logrus.Warnf("Container %s does not exist: %v", con.ID, err)
-			} else {
-				logrus.Errorf("Error removing container %s: %v", con.ID, err)
-			}
+	} else {
+		// Wait
+		var exitCode int32
+		exitCode, err = containers.Wait(ic.ClientCxt, con.ID, nil)
```
mheon, line 522: We should not be doing this if `--detach` was given.

vrothberg: Oh man, fair point! I'll polish a bit more

vrothberg:
@edsantiago PTAL

edsantiago:
@vrothberg did you try the reproducer in #7340? I'm still seeing it:
```console
$ while :;do got=$(./bin/podman-remote run --rm -v testvol1:/etc/apk alpine_labels ls /etc/apk); test "$got" = "$expect" || echo "got: $got";done
Error: no container with name or ID 447c18f8966f912c98e113257a980351c94e47c0d4c5692cd83efc011058b6ef found: no such container
Error: no container with name or ID be5a8aa6cc0be4c6f5d7add69e748498ee29fd2c465e399a14ac678ed8d4a40f found: no such container
Error: no container with name or ID 354abc85b5046fcb82bed15099f4c0cfd9cf4a7004dd57cb787daf0dee75bb7e found: no such container
```

vrothberg:
> @vrothberg did you try the reproducer in #7340? I'm still seeing it:

I ran it before for 20 minutes and didn't catch it anymore. I will recheck, maybe the last change screwed things up again. Thanks for checking.

/hold

TomSweeneyRedHat:
and some possible test issues.

vrothberg:
Looks like there's more unpack. Both, tunnel and abi, are doing (`if opts.Rm`):
```
if err := ic.Libpod.RemoveContainer(ctx, ctr, false, true); err != nil {       
        if errors.Cause(err) == define.ErrNoSuchCtr ||                         
                errors.Cause(err) == define.ErrCtrRemoved {                    
                logrus.Errorf("Container %s does not exist: %v", ctr.ID(), err)
        } else {                                                               
                logrus.Errorf("Error removing container %s: %v", ctr.ID(), err)
        }                                                                      
}                                                                              
```

Both are suspect to a race with the `podman container cleanup --rm` process. However, the tunnel parts run into the additional issue that we're not preserving the error types (we're losing them over the wire).

@mheon, do you know if the seemingly redundant removal in `ContainerRun` serves a purpose I may not be aware of? Seeing the CI failures, there might be something going on with volumes. Are `(libpod).RemoveContainer()` and `container cleanup --rm` identical?

mheon:
@vrothberg Basically, we have to be sure that the container is completely removed when the `podman run` process exits to prevent races - but we also need the cleanup process, because the `podman run` process can be killed, or the user can manually detach causing it to exit. So we need both. As you surmise, `RemoveContainer()` and `cleanup --rm` are the same thing - we're just invoking in duplicate to be certain the container is removed.

Generally speaking, the increasing priority of terminal-interactive tasks means that in the local case `podman run` will run before the cleanup process, so it will be able to call `RemoveContainer()` first. The cleanup process will fail, but it's not really logging its results anywhere, so that's perfectly fine. I imagine remote makes this considerably more complicated...

vrothberg:
Thanks, @mheon! Once https://github.com/containers/podman/pull/7574 is merged, I rebase and rework this PR.

vrothberg:
On top of the latest changes, it was sufficient to just fix the error checking.

@mheon @rhatdan PTanotherL

Issue #7250, opened 2020-08-07
podman logs colours the terminal persistently


tanwald:
/kind bug

**Description**

The font colour of the terminal session continues to be the same as the last log entry. Only `reset` or `echo -ne '\033c'` or similar resets the colour.

If the command is part of a script this behaviour can get confusing...

**Steps to reproduce the issue:**

1. `podman logs <container>`

**Describe the results you received:**

The font colour of the terminal session continues to be the same as the last log entry. 

**Describe the results you expected:**

!(The font colour of the terminal session continues to be the same as the last log entry.)

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.4
API Version:  1
Go Version:   go1.14.6
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.19, commit: 5dce9767526ed27f177a8fa3f281889ad509fea7'
  cpus: 16
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: tanwald1
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.7.11-200.fc32.x86_64
  linkmode: dynamic
  memFree: 46242598912
  memTotal: 67210756096
  ociRuntime:
    name: crun
    package: crun-0.14.1-1.fc32.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      commit: 598ea5e192ca12d4f6378217d3ab1415efeddefa
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.1.4-1.fc32.x86_64
    version: |-
      slirp4netns version 1.1.4
[...]

PR #7565, opened 2020-09-08
Terminal reset on logs complete

ParkerVR:

Response for https://github.com/containers/podman/issues/7250#event-3742016507
Although, I did this without a reproducer and assumed it would work. @tanwald @mheon if you know of any I will test it out

Detects if the os.Stdout is a terminal (or cygwin terminal) and resets colors on complete if so. (also adds an extra newline)

### DIFF
--- a/cmd/podman/containers/logs.go
+++ b/cmd/podman/containers/logs.go
@@ -1,12 +1,14 @@
 package containers
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/containers/podman/v2/cmd/podman/registry"
 	"github.com/containers/podman/v2/cmd/podman/validate"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/util"
+	"github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


mheon:
Vendor check is angry - you might want to swap to `terminal.IsTerminal()` which is already vendored?

cmd/podman/containers/logs.go
```diff
@@ -105,5 +107,10 @@ func logs(_ *cobra.Command, args []string) error {
 		logsOptions.Since = since
 	}
 	logsOptions.Writer = os.Stdout
-	return registry.ContainerEngine().ContainerLogs(registry.GetContext(), args, logsOptions.ContainerLogsOptions)
+
+	err := registry.ContainerEngine().ContainerLogs(registry.GetContext(), args, logsOptions.ContainerLogsOptions)
+	if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
+		fmt.Printf("\033[0m\n")
```
TomSweeneyRedHat, line 113: Could you define this using a variable?  `ResetAll = "\033[0m"`  or perhaps pull in https://godoc.org/github.com/whitedevops/colors and use from there?  I think defining our own constant is probably OK rather than pulling in colors.

Luap99:
I understand why someone want this fixed by podman but I don't think it would make much sense.
1. Someone might have set their color code to red for example and they wouldn't want it to be reset by podman.
2. Other linux utilities like `cat` and `tail` don't do this either. Just try:
```
$ echo -e '\033[0;31mtest' > tmp
$ cat tmp
```
3. I think it is very unlikely to end up in this state. If a program writes color codes to the terminal they would normally reset them.
4. You can reset this yourself automatically for all programs if you put the reset code into your `$PS1`.

rhatdan:
I agree, lets close this as won't fix.  I don't like resetting the terminal on completion, since this might reset from a known setting.

PR #7568, opened 2020-09-09
Fix podman ps -l

rhatdan:
Currently the --latest flag is ignored by podman ps command.

### DIFF
--- a/test/e2e/ps_test.go
+++ b/test/e2e/ps_test.go
@@ -104,11 +104,13 @@ var _ = Describe("Podman ps", func() {
 		SkipIfRemote()
 		_, ec, _ := podmanTest.RunLsContainer("")
 		Expect(ec).To(Equal(0))
+		_, ec, _ = podmanTest.RunLsContainer("")
+		Expect(ec).To(Equal(0))
 
-		result := podmanTest.Podman([]string{"ps", "--latest"})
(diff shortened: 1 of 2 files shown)


rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1877182

test/e2e/ps_test.go
```diff
@@ -104,11 +104,13 @@ var _ = Describe("Podman ps", func() {
 		SkipIfRemote()
 		_, ec, _ := podmanTest.RunLsContainer("")
 		Expect(ec).To(Equal(0))
+		_, ec, _ = podmanTest.RunLsContainer("")
+		Expect(ec).To(Equal(0))
```
zhangguanzhang, line 108: why the same code run twice

vrothberg: Before `ps -l` would display all containers. To test for this case, we need at least two running containers.

vrothberg (commented):
LGTM

rhatdan:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1877188

test/e2e/ps_test.go
rhatdan: We need two running containers.  The bug is that --latest should only show the 1 latest containers.  The test was failing to find the bug, since there was only one running container.

Issue #7157, opened 2020-07-30
Units created with "podman generate systemd --new" fail due missing podman state cleanup before start


yangm97:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

There's an issue with "undead" containers created with `--rm` from withn a shell script launched by a systemd unit which clog up state, and possibly corrupt state when too many of those are present. 
They show up on `ps -a` as `Created`. 

`container prune` on 1.6.x (plus armbian stretch) doesn't clean them, on 2.0.x (plus armbian buster) it almost always cleans all of those "undead" containers, except for when it fails with errors such as 

```
ERRO[0009] error unmounting /var/lib/containers/storage/overlay/bla/merged: invalid argument
``` 

followed by

```
Error: error creating overlay mount to /var/lib/containers/storage/overlay/bla/merged: no such file or directory
```

(btw isn't it weird that a prune operation is trying to mount stuff?)

**Steps to reproduce the issue:**

I couldn't narrow down where this is coming, it's one of those issues which happen "on its own" and "when it feels like doing so" 🙃, so the reproduction I could come up with was to mimick my exact environment using the ansible playbook and scripts from this gist: 

https://gist.github.com/yangm97/573b86a8b82181723aea96f70e690a98

This play will:
- Create the container wrapper script and symlinks
- Generate and persist a yggdrasil config from a j2 template
- Create a systemd unit which consumes the symlink and config above

Having that set up:
1. Try to abuse the host? Forced shutdowns maybe?

**Describe the results you received:**

```
$ sudo podman ps -a
CONTAINER ID  IMAGE                                                         COMMAND               CREATED       STATUS          PORTS   NAMES
019b650b5b0c  docker.io/yangm97/yggdrasil-go:latest                         -useconffile /etc...  6 weeks ago   ...

[truncated output]

1631fe0bddc3  docker.io/yangm97/yggdrasil-go:latest                         -useconffile /etc...  2 weeks ago   Up 2 weeks ago          yggdrasil.1594606637
20c48c933470  docker.io/yangm97/yggdrasil-go:latest                         -useconffile /etc...  3 months ago  

[truncated output]
```

**Describe the results you expected:**

Containers created with `--rm` should always be removed.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.2
API Version:  1
Go Version:   go1.14.4
[...]

PR #7571, opened 2020-09-09
generate systemd: catch `--name=foo`

vrothberg:
The systemd generator looks for certain flags in the containers' create
commands to determine which flags need to be added.  In case of named
containers, the generator adds the `--replace` flag to prevent name
conflicts at container creation.  Fix the generator to not only cover
the `--name foo` syntax but also the `--name=foo` one.

Fixes: #7157


rhatdan:
LGTM

Issue #7135, opened 2020-07-29
podman-remote: kill -s: hangs


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman pull fedora-minimal

$ ./bin/podman-remote run -d --name foo fedora-minimal env --default-signal sh -c "trap 'echo GOT HERE' 2;while :;do sleep 1;done"
ddd0303f467f170f46ca92e9b8c8b7ce61ca5574ad50de42ea82c4ee2a831cfa

$ time ./bin/podman-remote kill -s 2 foo
^C     <--- I got impatient before waiting 10 minutes
./bin/podman-remote kill -s 2 foo  0.12s user 0.09s system 0% cpu 8:34.25 total

$ ./bin/podman-remote logs foo
GOT HERE           <--- ok, so the signal DID make it to container
```
Using `fedora-minimal` and `env` because of #5004. 

It also hangs with `-s 3`, which indicates that `trap` is not the problem. It also hangs with just plain `sleep 999`, I just did the trap thing so I could see if the container was getting the signal.

edsantiago:
Still happening, master @ 575b3a3d8ab4f1f72d58b71aa4d534eca3a5d8f9

vrothberg:
I'll take a stab at this one. Thanks for providing a reproducer, @edsantiago !

vrothberg:
#7572 to the rescue

PR #7572, opened 2020-09-09
remote kill: don't wait for the container to stop

vrothberg:
Invert the branch logic to match the comment.  Docker seems to wait for
the container while Podman does not.

Fixes: #7135

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -174,7 +174,7 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrapf(err, "unable to kill Container %s", name))
 	}
 
-	if utils.IsLibpodRequest(r) {
+	if !utils.IsLibpodRequest(r) {
 		// the kill behavior for docker differs from podman in that they appear to wait
 		// for the Container to croak so the exit code is accurate immediately after the
 		// kill is sent.  libpod does not.  but we can add a wait here only for the docker
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"syscall"
 
 	"github.com/containers/podman/v2/libpod"
 	"github.com/containers/podman/v2/libpod/define"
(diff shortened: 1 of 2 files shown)


edsantiago:
Would you mind removing the "FIXME: pending #7135" skip in `test/system/130-kill.bats`? (I know it kills the elegance of a one-character patch, sorry!)

vrothberg:
> Would you mind removing the "FIXME: pending #7135" skip in `test/system/130-kill.bats`? (I know it kills the elegance of a one-character patch, sorry!)

Done!

@jwhonce, the kill-wait code has been added with commit d924494f561b but I found no sign in docker that it would actually wait. Did the behaviour change over time?

vrothberg:
> @jwhonce, the kill-wait code has been added with commit [d924494](https://github.com/containers/podman/commit/d924494f561bb878a2b3a7ce438d87ecb934b5fb) but I found no sign in docker that it would actually wait. Did the behaviour change over time?

Ah, Docker is only waiting on sigkill:
```golang
// ContainerKill sends signal to the container
// If no signal is given (sig 0), then Kill with SIGKILL and wait
// for the container to exit.
// If a signal is given, then just send it to the container and return.
func (daemon *Daemon) ContainerKill(name string, sig uint64) error {
        container, err := daemon.GetContainer(name)
        if err != nil {
                return err
        }

        if sig != 0 && !signal.ValidSignalForPlatform(syscall.Signal(sig)) {
                return fmt.Errorf("The %s daemon does not support signal %d", runtime.GOOS, sig)
        }

        // If no signal is passed, or SIGKILL, perform regular Kill (SIGKILL + wait())
        if sig == 0 || syscall.Signal(sig) == syscall.SIGKILL {
                return daemon.Kill(container)
        }
        return daemon.killWithSignal(container, int(sig))
}
```

vrothberg:
/hold

Let me fix that before

pkg/api/handlers/compat/containers.go
```diff
@@ -169,16 +170,16 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	if utils.IsLibpodRequest(r) {
-		// the kill behavior for docker differs from podman in that they appear to wait
-		// for the Container to croak so the exit code is accurate immediately after the
-		// kill is sent.  libpod does not.  but we can add a wait here only for the docker
-		// side of things and mimic that behavior
+	// Docker waits for the container to stop if the signal is 0 or
+	// SIGKILL.
+	if !utils.IsLibpodRequest(r) && (signal == 0 || syscall.Signal(signal) == syscall.SIGKILL) {
```
rhatdan, line 182: Why would signal=0 stop the container?  I though signal==0 means just check if the PID exists?

vrothberg: `signal == 0` in docker terms means default which in turn is SIGKILL

edsantiago: That's my understanding too: sig0 is never delivered to the target, it's only used as a check by calling code (for existence, also for EPERM). Good catch.

rhatdan: Ok.

rhatdan:
LGTM

mheon:
LGTM

mheon:
I'm going to hold/LGTM and hope tests go green.
/hold
/lgtm

vrothberg:
@mheon, can you /lgtm again. I accidentally added another commit on top but Cirrus is smart enough to cache results :)

PR #7573, opened 2020-09-09
Make oom-score-adj actually work

rhatdan:
During the redesign of podman 2.0, we dropped the support for --oom-score-adj.
Test for this flag was bogus and thus passing when it was broken.

Basically just need to set the value in the spec.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1877187


mheon:
LGTM

TomSweeneyRedHat:
LGTM

Issue #7117, opened 2020-07-28
podman-remote: container killed by 'rm -f', exit code is 0


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman-remote run --name foo alpine sleep 30
```
In window 3:
```console
$ ./bin/podman-remote rm -f foo    [pauses 10 seconds]
<sha>
```
Go back to window 2. Container has exited, but rc=0 (expected: 137). `podman stop foo` works as expected - kills the container, and the `run` process (window 2) exits 137.

Related: #3795

master @ 288ebec6e, root & rootless

mheon:
We need to adapt the fix for #3795 (reading events) for remote - right now it's local Podman only

zhangguanzhang:
I test this on 2.0.5, It still exists

zhangguanzhang:
https://github.com/containers/podman/blob/master/pkg/domain/infra/tunnel/containers.go#L481
containers has be removed by `podman-remote rm -f foo`, so this will response a 404
https://github.com/containers/podman/blob/master/pkg/bindings/containers/containers.go#L263
so line 481's `exitCode` will be 0

vrothberg:
I opened https://github.com/containers/podman/pull/7574 to fix it. This required quite some massaging but it's passing on my end.

PR #7574, opened 2020-09-09
remote run: consult events for exit code

vrothberg:
After attaching to a container, we wait for the container to finish and
return its exit code.  Waiting for the container may not always succeed,
for instance, when the container has been force removed by another
process.  In such case, we have to look at the *last* container-exit
event.

Also refactor the `ContainerRun` method a bit to return early on errors
and de-spaghetti the code.

Enable the remote-disabled system test.

Fixes: #7117

### DIFF
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -8,11 +8,13 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/podman/v2/libpod/define"
+	"github.com/containers/podman/v2/libpod/events"
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/bindings/containers"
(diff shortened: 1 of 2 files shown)


vrothberg:
@edsantiago feel free to stress-test

vrothberg:
@jwhonce @mheon @rhatdan @TomSweeneyRedHat PTAL

rhatdan:
LGTM

pkg/domain/infra/tunnel/containers.go
```diff
@@ -507,33 +509,83 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
...
-		if err == nil {
-			exitCode, err := containers.Wait(ic.ClientCxt, con.ID, nil)
-			if err == nil {
-				report.ExitCode = int(exitCode)
+
+	if opts.Rm {
+		// Defer the removal, so we can return early if needed and
+		// de-spaghetti the code.
+		defer func() {
+			if err := containers.Remove(ic.ClientCxt, con.ID, bindings.PFalse, bindings.PTrue); err != nil {
```
mheon, line 517: I think we don't want to do this in the opts.Detach case - for that, we rely entirely on the cleanup process. Should be enough to move this block down to 536

mheon:
One thing, otherwise LGTM

TomSweeneyRedHat:
LGTM
after @mheon's comment is addressed

pkg/domain/infra/tunnel/containers.go
vrothberg: Good point. The if start-attach failed, the error should also go on debug (as done locally).

vrothberg:
Ready to merge.

PR #7575, opened 2020-09-09
Fix podman container runlabel --display

rhatdan:
Current podman container runlabel --display is being ignored.

This is just supposed to display the command that would be run, and
then exit, but instead is actually running the command.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1877186

### DIFF
--- a/pkg/domain/infra/abi/containers_runlabel.go
+++ b/pkg/domain/infra/abi/containers_runlabel.go
@@ -36,6 +36,11 @@ func (ic *ContainerEngine) ContainerRunlabel(ctx context.Context, label string, 
 		return err
 	}
 
+	if options.Display {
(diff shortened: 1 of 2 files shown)


pkg/domain/infra/abi/containers_runlabel.go
```diff
@@ -36,6 +36,11 @@ func (ic *ContainerEngine) ContainerRunlabel(ctx context.Context, label string,
 		return err
 	}
 
+	if options.Display {
+		fmt.Printf("command: %s\n", strings.Join(append([]string{os.Args[0]}, cmd[1:]...), " "))
```
vrothberg, line 40: To double check: is the `command: ` prefix what we had in pre 2.0 time?

rhatdan: This code was cut and paste from the 1.9.0 base.

TomSweeneyRedHat:
LGTM

PR #7577, opened 2020-09-09
podman container runlabel should pull the image if it does not exist

rhatdan:
Since --pull is deprecated, remove it from help and hide if from --help
Also set it to true by default.

Share image pull code betweern podman image pull and podman container runlabel.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1877181

### DIFF
--- a/cmd/podman/containers/runlabel.go
+++ b/cmd/podman/containers/runlabel.go
@@ -30,7 +30,7 @@ var (
 		RunE:  runlabel,
 		Args:  cobra.MinimumNArgs(2),
 		Example: `podman container runlabel run imageID
-  podman container runlabel --pull install imageID arg1 arg2
+  podman container runlabel install imageID arg1 arg2
   podman container runlabel --display run myImage`,
 	}
 )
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
@rhatdan lint isn't hip

cmd/podman/containers/runlabel.go
```diff
@@ -61,11 +60,8 @@ func init() {
 	_ = flags.MarkHidden("opt1")
 	_ = flags.MarkHidden("opt2")
 	_ = flags.MarkHidden("opt3")
+	_ = flags.MarkHidden("pull")
 	_ = flags.MarkHidden("signature-policy")
-
-	if err := flags.MarkDeprecated("pull", "podman will pull if not found in local storage"); err != nil {
```
vrothberg, line 66: Please keep the flag deprecated. Otherwise, users using it won't be notified ("Flag --pull has been deprecated, podman will pull if not found in local storage") on that it's a NOP.

pkg/domain/infra/abi/images.go
```diff
@@ -294,6 +294,10 @@ func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entiti
 	return &entities.ImagePullReport{Images: foundIDs}, nil
 }
 
+func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, options entities.ImagePullOptions) (*entities.ImagePullReport, error) {
+	return pull(ir.Libpod.ImageRuntime(), ctx, rawImage, options)
```
vrothberg, line 298: This change seems out of context of the bug fix.

vrothberg (commented):
A nit and the request to keep the deprecation notification.

pkg/domain/infra/abi/images.go
rhatdan: The basic idea was to use one consistent way of pulling an image.  I found bugs in the runlabel pull code, that were different then podman pull, so rather then cut and paste in the fixes, I created a new shared function that both could use.

cmd/podman/containers/runlabel.go
rhatdan: Added back

rhatdan:
Lint actually caught a bug.

TomSweeneyRedHat:
LGTM

rhatdan:
@mheon @vrothberg @baude @jwhonce @QiWang19 @ashley-cui PTAL

QiWang19:
LGTM

Issue #7547, opened 2020-09-05
OCI runtime permission denied when trying to use --userns container:id


j0057:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I'm trying to run a container as a sidecar to another container, by sharing the user and net namespaces using the `--user container:id` and `--net container:id` options. This gives an "OCI runtime permission denied" error.

**Steps to reproduce the issue:**

1. Start the main container: `sudo podman run --rm -it --name test --uidmap 0:60000:1000 --gidmap 0:60000:1000 --net sys alpine:3.12`

2. Try to start the sidecar container: `sudo podman run --rm -it --name test-sidecar --userns container:test --net container:test alpine:3.12`

**Describe the results you received:**

An error message: `Error: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:47: preparing rootfs caused: permission denied: OCI runtime permission denied error`

**Describe the results you expected:**

Both the `test` and `test-sidecar` containers should run successfully and be visible from the host system as running under uid 60000.

**Additional information you deem important (e.g. issue happens only occasionally):**

Repro step 2 works if I remove the `--userns` flag, but then the main process runs as root not user 60000; it gives exactly the same error message if I run it with `--uidmap 0:60000:1000` instead.

**Output of `podman version`:**

```
Version:      2.0.6
API Version:  1
Go Version:   go1.15
Git Commit:   27362ba1ad8879ea71610fa68a651a1651e0180f
Built:        Tue Sep  1 21:43:39 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.20, commit: 13244db638cf987c415298a3c23393ae5abeb885'
  cpus: 8
  distribution:
    distribution: arch
    version: unknown
  eventLogger: file
  hostname: muon
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.8.5-arch1-1
  linkmode: dynamic
  memFree: 18484121600
  memTotal: 33629425664
  ociRuntime:
    name: runc
    package: Unknown
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc92
      commit: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
[...]

PR #7578, opened 2020-09-09
libpod: read mappings when joining a container userns

giuseppe:
when joining an existing container user namespace, read the existing
mappings so the storage can be created with the correct ownership.

Closes: https://github.com/containers/podman/issues/7547

Signed-off-by: Giuseppe Scrivano <giuseppe@scrivano.org>


TomSweeneyRedHat:
LGTM
assuming happy tests

PR #7588, opened 2020-09-10
tests/e2e: Add Toolbox-specific test cases

HarryMichal:
In the past, Toolbox[0] has been affected by several of Podman's
bugs/changes of behaviour. This is one of the steps to assure that as
Podman progresses, Podman itself and subsequently Toolbox do not regress.
One of the other steps is including Toolbox's system tests in Podman's
gating systems (which and to what extent is yet to be decided on).

The tests are trying to stress parts of Podman that Toolbox needs for
its functionality: permission to handle some system files, correct
values/permissions/limits in certain parts, management of users and
groups, mounting of paths,.. The list is most likely longer and
therefore more commits will be needed to control every aspect of the
Toolbox/Podman relationship :).

Some test cases in test/e2e/toolbox_test.go rely on some tools being
present in the base image[1]. That is not the case with the common
ALPINE image or the basic Fedora image.

Some tests might be duplicates of already existing tests. I'm more in
favor of having those duplicates. Thanks to that it will be clear what
functionality/behaviour Toolbox requires.

[0] https://github.com/containers/toolbox
[1] https://github.com/containers/toolbox/#image-requirements

I'm opening the PR to get some feedback if these tests seem okay. I'm 100% sure that these are not all the tests cases we'd like to get covered but I think it's better to have something to start with and then pile on top of it :).

/cc @debarshiray @edsantiago

### DIFF
--- a/test/e2e/toolbox_test.go
+++ b/test/e2e/toolbox_test.go
@@ -1,0 +1,368 @@
+package integration
+
+/*
+	toolbox_test.go is under the care of the Toolbox Team.
+
+	The tests are trying to stress parts of Podman that Toolbox[0] needs for
+	its functionality.
+
+	[0] https://github.com/containers/toolbox
+
+	Info about test cases:
+	- some tests rely on a certain configuration of a container that is done by
+		executing several commands in the entry-point of a container. To make
+		sure the initialization had enough time to be executed,
+		WaitContainerReady() after the container is started.
+
+	- in several places there's an invocation of 'podman logs' It is there mainly
+		to ease debugging when a test goes wrong (during the initialization of a
+		container) but sometimes it is also used in the test case itself.
+
+	Maintainers (Toolbox Team):
+	- Ondřej Míchal <harrymichal@fedoraproject.org>
+	- Debarshi Ray <rishi@fedoraproject.org>
+
+	Also available on Freenode IRC on #silverblue or #podman
+*/
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"os/user"
+	"strconv"
+	"strings"
+	"syscall"
+
+	. "github.com/containers/podman/v2/test/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Toolbox-specific testing", func() {
+	var (
+		tempdir    string
+		err        error
+		podmanTest *PodmanTestIntegration
+	)
+
+	BeforeEach(func() {
+		tempdir, err = CreateTempDirInTempDir()
+		if err != nil {
+			os.Exit(1)
+		}
+		podmanTest = PodmanTestCreate(tempdir)
+		podmanTest.Setup()
+		podmanTest.SeedImages()
+	})
+
+	AfterEach(func() {
+		podmanTest.Cleanup()
+		f := CurrentGinkgoTestDescription()
+		processTestResult(f)
+	})
+
+	It("podman run --dns=none - allows self-management of /etc/resolv.conf", func() {
+		var session *PodmanSessionIntegration
+
+		session = podmanTest.Podman([]string{"run", "--dns", "none", ALPINE, "sh", "-c",
+			"rm -f /etc/resolv.conf; touch -d '1970-01-01 00:02:03' /etc/resolv.conf; stat -c %s:%Y /etc/resolv.conf"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring("0:123"))
+	})
+
+	It("podman run --no-hosts - allows self-management of /etc/hosts", func() {
+		var session *PodmanSessionIntegration
+
+		session = podmanTest.Podman([]string{"run", "--no-hosts", ALPINE, "sh", "-c",
+			"rm -f /etc/hosts; touch -d '1970-01-01 00:02:03' /etc/hosts; stat -c %s:%Y /etc/hosts"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring("0:123"))
+	})
+
+	It("podman create --ulimit host + podman exec - correctly mirrors hosts ulimits", func() {
+		if podmanTest.RemoteTest {
+			Skip("Ulimit check does not work with a remote client")
+		}
+		var session *PodmanSessionIntegration
+		var containerHardLimit int
+		var rlimit syscall.Rlimit
+		var err error
+
+		err = syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit)
+		Expect(err).To(BeNil())
+		fmt.Printf("Expected value: %d", rlimit.Max)
+
+		session = podmanTest.Podman([]string{"create", "--name", "test", "--ulimit", "host", ALPINE,
+			"sleep", "1000"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"start", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"exec", "test", "sh", "-c",
+			"ulimit -H -n"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		containerHardLimit, err = strconv.Atoi(strings.Trim(session.OutputToString(), "\n"))
+		Expect(err).To(BeNil())
+		Expect(containerHardLimit).To(BeNumerically(">=", rlimit.Max))
+	})
+
+	It("podman create --ipc=host --pid=host + podman exec - correct shared memory limit size", func() {
+		// Comparison of the size of /dev/shm on the host being equal to the one in
+		// a container
+		if podmanTest.RemoteTest {
+			Skip("Shm size check does not work with a remote client")
+		}
+		var session *PodmanSessionIntegration
+		var cmd *exec.Cmd
+		var hostShmSize, containerShmSize int
+		var err error
+
+		// Because Alpine uses busybox, most commands don't offer advanced options
+		// like "--output" in df. Therefore the value of the field 'Size' (or
+		// ('1K-blocks') needs to be extracted manually.
+		cmd = exec.Command("df", "/dev/shm")
+		res, err := cmd.Output()
+		Expect(err).To(BeNil())
+		lines := strings.SplitN(string(res), "\n", 2)
+		fields := strings.Fields(lines[len(lines)-1])
+		hostShmSize, err = strconv.Atoi(fields[1])
+		Expect(err).To(BeNil())
+
+		session = podmanTest.Podman([]string{"create", "--name", "test", "--ipc=host", "--pid=host", ALPINE,
+			"sleep", "1000"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"start", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"exec", "test",
+			"df", "/dev/shm"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		lines = session.OutputToStringArray()
+		fields = strings.Fields(lines[len(lines)-1])
+		containerShmSize, err = strconv.Atoi(fields[1])
+		Expect(err).To(BeNil())
+
+		// In some cases it may happen that the size of /dev/shm is not exactly
+		// equal. Therefore it's fine if there's a slight tolerance between the
+		// compared values.
+		Expect(hostShmSize).To(BeNumerically("~", containerShmSize, 100))
+	})
+
+	It("podman create --userns=keep-id --user root:root - entrypoint - entrypoint is executed as root", func() {
+		var session *PodmanSessionIntegration
+
+		session = podmanTest.Podman([]string{"run", "--userns=keep-id", "--user", "root:root", ALPINE,
+			"id"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring("uid=0(root) gid=0(root)"))
+	})
+
+	It("podman create --userns=keep-id + podman exec - correct names of user and group", func() {
+		var session *PodmanSessionIntegration
+		var err error
+
+		currentUser, err := user.Current()
+		Expect(err).To(BeNil())
+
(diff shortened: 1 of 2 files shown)


test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,298 @@
...
+		f := CurrentGinkgoTestDescription()
+		processTestResult(f)
+	})
+
+	It("podman run --dns=none - allows self-management of /etc/resolv.conf", func() {
+		var session *PodmanSessionIntegration
+
+		session = podmanTest.Podman([]string{"run", "--dns", "none", ALPINE,
+			"rm", "/etc/resolv.conf", ";",
+			"touch", "/etc/resolv.conf"})
```
edsantiago, line 59: Does this syntax actually work? The usual convention seems to be:
```go
    podmanTest...( "sh", "-c", "this;that;theother")
```
...but this was just a detour for bringing me to my favorite rant: "run something and only check exit code" is not a test. Some typos can lead to a command passing but not actually doing anything. A much better idea is to check for unusual strings. For instance:
```
sh -c "rm /etc/resolv.conf && touch -d '1970-01-01 00:02:03' /etc/resolv.conf && stat -c %s:%Y /etc/resolv.conv"
```
...and confirm that you get an exact match on `0:123`

HarryMichal: > Does this syntax actually work? The usual convention seems to be:

Yes, I was not sure entirely how to go about the commands in the entry-point. Your idea seems to be the right one.

> ...but this was just a detour for bringing me to my favorite rant: "run something and only check exit code" is not a test. 

Roger! The next iteration should fix this. Thank you!

rhatdan:
LGTM
@edsantiago @mheon @vrothberg @TomSweeneyRedHat @ashley-cui @QiWang19 PTAL

edsantiago:
@rhatdan I'm waiting for CI churn

mheon:
We may want to add fedoraToolbox to the cached images - it's a ~60mb pull for three separate tests.

rhatdan:
Yes adding this to cached images makes sense.

HarryMichal:
I still need to finish the test-cases before this can be merged. I keep running into problems :).

Is it on me to cache the image or is that something done by you, guys? I wanted to avoid relying on the `fedora-toolbox` image because the way it is stored in the registry is stupid and requires a bit of maintenance (the full name of the image is `fxx/fedora-toolbox` where `xx` stands for the version number of Fedora instead of being called just `fedora-toolbox`). But for the maintenance, there are contacts to me or @debarshiray.

TomSweeneyRedHat:
Great idea @HarryMichal , thanks for taking this on.
Changes LGTM in general, but I'll lean on @edsantiago 's opinion.
Tests at the moment are not happy.

HarryMichal:
The tests are failing on just one test case and I'm not sure if it is due to the test case not being correct or it's actually hitting a bug. The `--ulimit host` option in `podman create` is supposed to set ulimits in the container to match the host but the hard limit in the container is twice is large than on the host.

@rhatdan, what do you think?

edsantiago:
@HarryMichal see `setrlimit(2)`:
```
       The  soft  limit  is  the value that the kernel enforces for the corre‐
       sponding resource.  The hard limit acts  as  a  ceiling  for  the  soft
       limit:  an  unprivileged process may set only its soft limit to a value
       in the range from 0 up to the hard limit, and (irreversibly) lower  its
       hard   limit.    A  privileged  process  (under  Linux:  one  with  the
       CAP_SYS_RESOURCE capability in the initial  user  namespace)  may  make
       arbitrary changes to either limit value.
```

HarryMichal:
Oh, ok, that's something I did not know about. Thank you @edsantiago! In that case, I'll drop the test-case.

edsantiago:
Well... what was the reason behind your need for this test case? You could still check that container limit is not less than host, or you could leave the `==` test for rootless only (rootless, of course, cannot bump up limits). It depends on what this test was hoping to accomplish.

HarryMichal:
> what was the reason behind your need for this test case?

It's used for Toolbox so I included it.

> You could still check that container limit is not less than host, or you could leave the == test for rootless only (rootless, of course, cannot bump up limits). It depends on what this test was hoping to accomplish.

The first option sounds good to me.

HarryMichal:
I believe the failed test is a flake.

An image cache was mentioned... is that something that needs to be done in the PR, too? Apart from that, I think this is good to go.

@edsantiago

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,346 @@
...
+			Skip("Shm size check does not work with a remote client")
+		}
+		var session *PodmanSessionIntegration
+		var cmd *exec.Cmd
+		var hostShmSize, containerShmSize int
+		var err error
+
+		re, err := regexp.Compile("([0-9]+)")
+
+		cmd = exec.Command("df", "-h", "/dev/shm")
```
edsantiago, line 121: Future maintainers will appreciate a comment here such as "df 'Size' field on host".

And, while I was in the process of trying to grok this I wondered: why the `df -h` and regex and `Atoi`? Why not just run `df` (no `-h`) on host and container, and string-compare `Size`?

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+		session = podmanTest.Podman([]string{"exec", "test",
+			"id"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		ok, _ := session.GrepString(expectedOutput)
+		Expect(ok).To(BeTrue())
+	})
+
+	It("podman create --userns=keep-id - entrypoint - adding user with useradd and then removing his password", func() {
```
edsantiago, line 195: Let's try to keep it gender-neutral - "the" or "their password" would be more appropriate

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,346 @@
...
+			fmt.Sprintf("useradd --home-dir %s --shell %s --uid %s %s; passwd --delete %s; sleep 1000",
+				homeDir, shell, uid, username, username)})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"start", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		time.Sleep(3 * time.Second)
```
edsantiago, line 214: I see a lot of these sleeps. Tests are already close to the CI timeout limit, and every second counts. Are these necessary? If so, would 1 second be OK? (And can you document in a comment why they're necessary?)

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,346 @@
...
+			username,
+			uid,
+			gid,
+			homeDir,
+			shell)
+		fmt.Printf("Expected line in /etc/passwd: %s\n", expectedUser)
+
+		session = podmanTest.Podman([]string{"create", "--name", "test", "--userns=keep-id", "--user", "root:root", fedoraToolbox, "sh", "-c",
+			fmt.Sprintf("useradd --home-dir %s --shell %s --uid %s %s; groupadd --gid %s %s; usermod --append --groups wheel --home %s --shell %s --uid %s --gid %s %s; sleep 1000",
+				badHomeDir, badShell, badUID, username, gid, groupName, homeDir, shell, uid, gid, username)})
```
edsantiago, line 283: As a courtesy to future maintainers who might not want to count and match `%s`s, could you maybe split this into three separate `Sprintf`s, one for each add/mod command? That would make it very easy for the eye to see which `%s` corresponds to which variable.

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+		session = podmanTest.Podman([]string{"start", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		time.Sleep(3 * time.Second)
+
+		session = podmanTest.Podman([]string{"logs", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
edsantiago, line 295: As long as you're running `logs`, wouldn't it be a good idea to check for error output from the `baduser` add?

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+		session = podmanTest.Podman([]string{"start", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		time.Sleep(1 * time.Second)
+
+		session = podmanTest.Podman([]string{"logs", "test"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
edsantiago, line 344: Similarly to above `logs` comment, wouldn't it be a good idea to make sure that output here is empty? Any output at all could be an important warning that could be missed.

edsantiago (commented):
Nice work, LGTM with a few suggestions. And, is there a reason for the separate commits? It might make sense to squash.

Thanks for your perseverance on this@

test/e2e/toolbox_test.go
HarryMichal: I added several comments to make the test case more clear.

I agree with you that the use `df -h` is indeed kinda weird and instead I updated the test to use `df` without the `-h` option. Therefore the regexp won't be necessary anymore. But I'd like to keep `Atoi` because in some cases the size of `/dev/shm` is not exactly equal (I found that out when troubleshooting the tests in the CI). The values being numerical gives the ability to add to the check some tolerance.

test/e2e/toolbox_test.go
HarryMichal: Fixed.

test/e2e/toolbox_test.go
HarryMichal: I agree that the timeout is too long. My motivation was to give the entry-point enough time to be executed. It definitely can be just a second or even less. But I've got a different idea of how to go about this. Instead of explicitly waiting I switched the invocation of `podman logs` with `podman exec` that happens quite often. Actually... now I'm wondering if that will fix it :sweat_smile: .

Anyway, the timeout can be definitely lower but some is definitely necessary.

test/e2e/toolbox_test.go
HarryMichal: Fixed.

test/e2e/toolbox_test.go
HarryMichal: `baduser` does not imply the invocation of `useradd` should fail. I added a comment that clarifies that fact. Nevertheless, I added checks after the invocation of `podman logs` to make sure it is ok.

test/e2e/toolbox_test.go
HarryMichal: Fixed.

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+	It("podman run --no-hosts - allows self-management of /etc/hosts", func() {
+		var session *PodmanSessionIntegration
+
+		session = podmanTest.Podman([]string{"run", "--no-hosts", ALPINE, "sh", "-c",
+			"rm -f /etc/hosts; touch -d '1970-01-01 00:02:03' /etc/hosts; stat -c %s:%Y /etc/hosts"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		ok, _ := session.GrepString("0:123")
+		Expect(ok).To(BeTrue())
```
edsantiago, line 85: For future work, I would encourage you to get in the habit of seeing `BeTrue()` as unmitigated evil. There are ginkgo functions that do string checks; the benefit of those is that when something goes wrong, you get nice actual error messages ("expected this, saw that") which actually lets a human being figure out what happened.

Unfortunately it's a lot of work to figure out what those are, because ginkgo is not human-friendly. So it's not worth re-pushing and going through CI again. But there will be a next time, and I hope you'll keep that in mind.

edsantiago (changes requested):
One suggestion for next time, but otherwise LGTM. Thanks for your work on this.

edsantiago (approved):

edsantiago:
@containers/podman-maintainers  PTAL

edsantiago:
@HarryMichal our CI environment changed today; you're going to need to rebase in order for us to merge this.

test/e2e/toolbox_test.go
HarryMichal: I'll keep that in mind in the future. I took inspiration from the other test files and more-or-less copied what I saw there. It was also my first time working with ginkgo.

HarryMichal:
I must say that the names of the tests are much clearer now :). Good work!

edsantiago:
Did someone press re-run on a failed test here? If so, I think we have a problem.

When I checked earlier this morning I saw a [test failure that looked real](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5310199206510592/html/int-remote-fedora-31-root-host.log.html#t--podman-create---userns-keep-id---entrypoint---adding-user-with-useradd-and-then-removing-their-password--1). It is now gone, but there has been no push. That means this is a flake, and is going to come back to haunt us later on.

The errors are both in `Toolsbox-specific testing`, two different tests, each failed three times in a row. The failures are the ginkgo anti-helpful ones, but @HarryMichal I beg you to please look into them before merging this.

edsantiago (changes requested):
I'm going to see if this blocks merging. I want to understand the failures because I really do not want them to make their way into master.

test/e2e/toolbox_test.go
```diff
@@ -0,0 +1,373 @@
...
+		useradd := fmt.Sprintf("useradd --home-dir %s --shell %s --uid %s %s",
+			badHomeDir, badShell, badUID, username)
+		groupadd := fmt.Sprintf("groupadd --gid %s %s",
+			gid, groupName)
+		usermod := fmt.Sprintf("usermod --append --groups wheel --home %s --shell %s --uid %s --gid %s %s",
+			homeDir, shell, uid, gid, username)
+
+		session = podmanTest.Podman([]string{"create", "--name", "test", "--userns=keep-id", "--user", "root:root", fedoraToolbox, "sh", "-c",
+			fmt.Sprintf("%s; %s; %s; sleep 1000", useradd, groupadd, usermod)})
+		session.WaitWithDefaultTimeout()
```
edsantiago, line 302: @HarryMichal I'm pretty sure the error is here: this is racy. Could you look into using `WaitContainerReady()`? You would need to `echo READY` before the `sleep`, and adjust the `podman logs = nil` test below, and also have to do this on all other tests; but I think that's the easiest safest way to proceed.

HarryMichal: `WaitContainerReady()` does what I was trying to accomplish by calling `sleep`. I did not know about it. Thank you! I'll adjust the code to make use of it.

HarryMichal:
I tried to make use of `WaitContainerReady()` and while going through the tests I also got rid of `ok, _ := session.GrepString()` + `Expect(ok).To(BeTrue())` and replaced those lines with `Expect(session.OutputToString()).To(ContainSubstring())`.

HarryMichal:
Ah! I forgot to adjust most of the tests to account for the fact the logs are not empty most of the time now because of the use of `WaitContainerReady()`.

edsantiago (approved):

edsantiago:
Nice work! Let's get this in.
/lgtm
/hold cancel

Issue #7580, opened 2020-09-10
CI flake: podman run --conmon-pidfile: conmon not running?


edsantiago:
One of the system tests does `podman run --conmon-pidfile=PATH` then reads pid from PATH and confirms that `/proc/pid/exe` is a symlink pointing to `.*/conmon`. This test has flaked twice in the past month, with:
```
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: conmon pidfile (= PID 100491) points to conmon process
#| expected: '.*/conmon'
#|   actual: ''
#\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
The two flakes:

* fedora-32 : special_testing_rootless
  * PR #7539
    * [09-05 07:13](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4787272123940864/html/system_test.log.html)
* fedora-32 : special_testing_rootless [remote]
  * PR #7346
    * [08-18 17:42](https://api.cirrus-ci.com/v1/task/5004720299311104/logs/system_test.log)

Source code: https://github.com/containers/podman/blob/a82de0e3a039bb23d3a9bdb0a1cb9a56d5cec5df/test/system/030-run.bats#L178

This is so infrequent that it's hard to know where to start. I'm filing because:
* Someone might perhaps think of a race condition in the conmon-pidfile code (e.g. perhaps the code is not flushing/closing the output file, so my test is reading an incomplete PID? Maybe the PID above should've been 100491X?)
* Someone could suggest a way for me to instrument the test, to make the next failure easier to diagnose?

rhatdan:
This looks like the old race condition where we were not getting stdout back on the remote end.  I believe that is fixed?

edsantiago:
You mean #7195? That seems doubtful because one of the failures is non-remote, and regardless, the missing output is from `readlink` which is not running under podman. Did I misunderstand?

rhatdan:
Ok, I did not read the actual code, I just saw the failure and it looked familiar.

rhatdan:
Would this happen if $conmon_pid nolonger existed?

edsantiago:
Yes. Is  that possible in a `sleep infinity` container?

rhatdan:
One would hope not, but that is the symptom

rhatdan:
[...]

Issue #7371, opened 2020-08-19
podman save --format oci-archive: loses image ID


edsantiago:
This may not be a bug, but I think it is, because `podman save` without `--format oci-archive` preserves image ID.
```console
$ cat >oci-check.sh <<EOF
#!/bin/bash

image=${1?Usage: $0 FQIN}

OCI_FORMAT="--format oci-archive"

./bin/podman rm  -a >/dev/null
./bin/podman rmi -a >/dev/null
./bin/podman pull -q "$image" >/dev/null
./bin/podman images --noheading
rm -f /tmp/foo.tar
./bin/podman save $OCI_FORMAT -o /tmp/foo.tar "$image"
./bin/podman rmi -a >/dev/null
./bin/podman load -q -i /tmp/foo.tar   >/dev/null
./bin/podman images --noheading
EOF
$ chmod 755 ./oci-check.sh

!! basically just "podman pull", "podman save", and "podman load". Note differing image IDs.
$ for i in docker.io/library/busybox docker.io/library/alpine:latest quay.io/libpod/alpine_labels:latest;do ./oci-check.sh $i;echo;done
docker.io/library/busybox  latest  018c9d7b792b  3 weeks ago  1.45 MB
docker.io/library/busybox  latest  c71b90eb5b17  3 weeks ago  1.44 MB

docker.io/library/alpine  latest  a24bb4013296  2 months ago  5.85 MB
docker.io/library/alpine  latest  0f5f445df8cc  2 months ago  5.85 MB

quay.io/libpod/alpine_labels  latest  4fab981df737  21 months ago  4.69 MB
quay.io/libpod/alpine_labels  latest  dd8b0a90b986  21 months ago  4.68 MB

```
Commenting out `OCI_FORMAT=` makes all the IIDs agree.

master @ 7e2a1b34c

vrothberg:
@mtrmac, do you think it's a bug or feature request/enhancement?

mtrmac:
In general, neither. The change ultimately happens in https://github.com/containers/image/blob/1c313b2d23e039a81a64bbced1b2b427d60429e1/image/docker_schema2.go#L78 , and as the comment says, the conversion from v2s2 to OCI is lossy (IIRC OCI does not have health checks).

[...]

PR #7590, opened 2020-09-10
system tests: cleanup

edsantiago:
- run tests: better "skip" message for docker-archive test;
  remove FIXME, document that podman-remote doesn't support it

- run tests: instrument the --conmon-pidfile test in hopes
  of tracking down flake #7580: cross-check pidfile against
  output of 'podman inspect', and add some debug messages
  that will only be seen on test failure.

- load tests: the pipe test: save and load a temporary tag,
  not $IMAGE. Primary reason is because of #7371, in which
  'podman load' assigns a new image ID (instead of preserving
  the saved one). This messes with our image management, and
  it turns out to be nonfixable.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -189,9 +189,19 @@ echo $rand        |   0 | $rand
 
     is "$(< $cidfile)" "$cid" "contents of cidfile == container ID"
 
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat (commented):
LGTM

PR #7592, opened 2020-09-10
Force Attach() to send a SIGWINCH and redraw

### LINKED ISSUE #6253: podman attach prints nothing until modification
**Description**

If I attach to a container the terminal is still empty until:

* I press a key to change the output on the screen (like enter in a shell)
* I resize the screen

I could not find any other issue with that problem.

/kind bug

1. `podman run --name test -it debian bash`
2. `podman attach test`

I would expect that there is an output without modify the screen size or the content. 
`docker` handles this case better. 

```
Version:            1.9.1
RemoteAPI Version:  1
Go Version:         go1.11.6
```

mheon:
Basically, we want to force the application in the container to (iff the container was made with a terminal) redraw said terminal immediately after an attach completes, so the fresh Attach session will be able to see what's going on (e.g. will have a shell prompt). Our current attach functions are unfortunately geared more towards `podman run` than `podman attach` and will start forwarding resize events *immediately* instead of waiting until the attach session is alive (much safer for short-lived `podman run` sessions, but broken for the `podman attach` case). To avoid a major rewrite, let's just manually send a SIGWINCH after attach succeeds to force a redraw.

Fixes #6253

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/libpod/events"
+	"github.com/containers/podman/v2/pkg/signal"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
LGTM

Issue #7583, opened 2020-09-10
CI flake:   podman run with new:pod and static-ip: Error adding network: failed to set bridge addr


edsantiago:
The symptom is:
```
  podman run with new:pod and static-ip
 ...
Running: podman [options] network create --subnet 10.25.40.0/24 podmantestnetwork2
/etc/cni/net.d/podmantestnetwork2.conflist
Running: podman [options] run -t -i --rm --pod new:testpod --net podmantestnetwork2 --ip 10.25.40.128 docker.io/library/alpine:latest ip addr
time="2020-09-03T15:37:25-05:00" level=error msg="Error adding network: failed to set bridge addr: \"cni-podman1\" already has an IP address different from 10.25.40.1/24"
time="2020-09-03T15:37:25-05:00" level=error msg="Error while adding pod to CNI network \"podmantestnetwork2\": failed to set bridge addr: \"cni-podman1\" already has an IP address different from 10.25.40.1/24"
time="2020-09-03T15:37:25-05:00" level=error msg="error starting some container dependencies"
time="2020-09-03T15:37:25-05:00" level=error msg="\"error configuring network namespace for container 6091376fe654ca512bc885c89c0026c76549f8a47e174c37819e77f8c5678cfe: failed to set bridge addr: \\\"cni-podman1\\\" already has an IP address different from 10.25.40.1/24\""
Error: error starting some containers: internal libpod error
```
These are all three-time failures, causing the entire CI job to fail, probably because the first failure leaves things in a bad state for the test reruns -- subsequent attempts fail with:
```
Running: podman [options] network create --subnet 10.25.40.0/24 podmantestnetwork2
Error: network 10.25.40.0/24 is already being used by a cni configuration
```
So, at a minimum, I would suggest fixing the test so as to clean up after itself.

Flake history:

* fedora-32 : test fedora-32
  * PR #7408
    * [08-21 22:29](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5483323134836736/html/integration_test.log.html#t--podman-run-with-new-pod-and-static-ip--1)
    * [08-21 22:29](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5483323134836736/html/integration_test.log.html#t--podman-run-with-new-pod-and-static-ip--2)
[...]

PR #7593, opened 2020-09-10
run_networking e2e test: add cleanup to some tests

edsantiago:
Problem: if either of the two "podman network create" tests
fail, all subsequent retries will also fail because the
created network has not been cleaned up (so "network create"
will fail with EEXIST).

Solution: run "podman network rm" as deferred cleanup instead
of in each test.

This is NOT a fix for #7583 - it is just a way to allow
ginkgo to retry a failing test.


TomSweeneyRedHat:
LGTM

vrothberg (commented):
good catch!
LGTM

Issue #7167, opened 2020-07-31
'--infra-command' parameter in pod create not working as expected


KKoukiou:
/kind bug

**Description**

Running the following command `podman pod create --infra-image alpine --infra-command 'sleep 1000' --name pod-1`
will create `pod-1` with and infra container whose command is `/bin/bash` and not `sleep 1000` as specified in the arguments.

I would expect that the infra container created would now have an EntryPoint equal to `sleep `1000`. Instead I get the following configuration.

**Additional information you deem important (e.g. issue happens only occasionally):**
``` bash
[root@localhost ~]# podman ps --all
CONTAINER ID  IMAGE                            COMMAND  CREATED         STATUS                         PORTS   NAMES
...
a7b81b9a77e3  docker.io/library/alpine:latest  /bin/sh  3 minutes ago   Exited (0) About a minute ago          23a396950741-infra

[root@localhost ~]# podman pod inspect pod-1
{
    "Id": "23a3969507419bbe8a095d3313e79286e66a9ddd57bcb1ee125318d4ad4b02bd",
    "Name": "pod-1",
    "Created": "2020-07-31T07:53:18.994608342-04:00",
    "CreateCommand": [
        "podman",
        "pod",
        "create",
        "--infra-image",
        "alpine",
        "--infra-command",
        "sleep 1000",
        "--name",
        "pod-1"
    ],
    "State": "Stopped",
    "Hostname": "pod-1",
    "CreateCgroup": false,
    "CgroupParent": "machine.slice",
    "CgroupPath": "machine.slice/machine-libpod_pod_23a3969507419bbe8a095d3313e79286e66a9ddd57bcb1ee125318d4ad4b02bd.slice",
    "CreateInfra": false,
    "InfraContainerID": "a7b81b9a77e30c7c783dddf9af11b1f617b84eef12f36115053dd469270a4e80",
    "SharedNamespaces": [
        "ipc",
        "net",
        "uts"
    ],
    "NumContainers": 2,
    "Containers": [
        {
            "Id": "9db93de08b23d0fb2814cae0df653e75207e747638fe0d7a1cd00e3764640fc7",
            "Name": "test-pod-1-system",
            "State": "configured"
        },
        {
            "Id": "a7b81b9a77e30c7c783dddf9af11b1f617b84eef12f36115053dd469270a4e80",
            "Name": "23a396950741-infra",
            "State": "stopped"
        }
    ]
}

```

**Output of `podman version`:**

```
2.0.2
```

[...]

PR #7621, opened 2020-09-13
Fix podman pod create --infra-command and --infra-image

rhatdan:
Currently infr-command and --infra-image commands are ignored
from the user.  This PR instruments them and adds tests for
each combination.

### DIFF
--- a/cmd/podman/pods/create.go
+++ b/cmd/podman/pods/create.go
@@ -55,8 +55,8 @@ func init() {
 	flags.StringVar(&createOptions.CGroupParent, "cgroup-parent", "", "Set parent cgroup for the pod")
 	flags.BoolVar(&createOptions.Infra, "infra", true, "Create an infra container associated with the pod to share namespaces with")
 	flags.StringVar(&createOptions.InfraConmonPidFile, "infra-conmon-pidfile", "", "Path to the file that will receive the POD of the infra container's conmon")
-	flags.StringVar(&createOptions.InfraImage, "infra-image", containerConfig.Engine.InfraImage, "The image of the infra container to associate with the pod")
-	flags.StringVar(&createOptions.InfraCommand, "infra-command", containerConfig.Engine.InfraCommand, "The command to run on the infra container when the pod is started")
+	flags.String("infra-image", containerConfig.Engine.InfraImage, "The image of the infra container to associate with the pod")
+	flags.String("infra-command", containerConfig.Engine.InfraCommand, "The command to run on the infra container when the pod is started")
 	flags.StringSliceVar(&labelFile, "label-file", []string{}, "Read in a line delimited file of labels")
 	flags.StringSliceVarP(&labels, "label", "l", []string{}, "Set metadata on pod (default [])")
 	flags.StringVarP(&createOptions.Name, "name", "n", "", "Assign a name to the pod")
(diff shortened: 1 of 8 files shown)


rhatdan:
Fixes https://github.com/containers/podman/issues/7167

libpod/pod.go
```diff
@@ -70,6 +70,13 @@ type PodConfig struct {
 
 	// ID of the pod's lock
 	LockID uint32 `json:"lockID"`
+
+	// InfraImage is the image a pod infra container will use to manage
+	// namespaces.
+	InfraImage string `json:"InfraImage,omitempty"`
```
mheon, line 76: Throw these in InfraContainerConfig

TomSweeneyRedHat:
You've some lint unhappiness in here @rhatdan

libpod/pod.go
```diff
@@ -70,6 +70,13 @@ type PodConfig struct {
 
 	// ID of the pod's lock
 	LockID uint32 `json:"lockID"`
+
+	// InfraImage is the image a pod infra container will use to manage
+	// namespaces.
+	InfraImage string `json:"infraImage,omitempty"`
```
mheon, line 76: Can you take these out of PodConfig now that they're in InfraContainerConfig?

rhatdan: Light dawns on marble head.  Fixed.

TomSweeneyRedHat:
Still fighting test issues @rhatdan

rhatdan:
This is ready to go in,
@TomSweeneyRedHat @mheon @vrothberg @giuseppe @ashley-cui @QiWang19 @jwhonce @baude PTAL

cmd/podman/pods/create.go
```diff
@@ -104,6 +103,18 @@ func create(cmd *cobra.Command, args []string) error {
 		createOptions.Share = nil
 	} else {
 		createOptions.Share = strings.Split(share, ",")
+		if cmd.Flag("infra-command").Changed {
+			createOptions.InfraCommand, err = cmd.Flags().GetString("infra-command")
```
vrothberg, line 107: Why is that necessary? It seems rather exotic. A comment would be very helpful.

vrothberg: Same below.

vrothberg (commented):
Other than the nit, LGTM

cmd/podman/pods/create.go
rhatdan: Not sure it is exotic, since this type of code is all over the cli.  Basically we want to make sure the user made a change using the CLI, rather then sending the default settings over to the server.  That way the server can differentiate if the user wanted a change or just to take defaults.

vrothberg: So we wouldn't take the defaults from the client containers.conf but from the server?

rhatdan: Yes

vrothberg: What's the reasoning for that? Do we document that somewhere?

I think there's some inconsistency among the commands w.r.t. to which defaults come from the client and server.

I am not against the idea but just want to understand if we have some policy for that decision.

rhatdan: All defaults should come from the server.  Any defaults that come from the client are a bug.  But I agree there are some coming from the client.

rhatdan: @vrothberg I now have comments, PTAL

TomSweeneyRedHat (commented):
LGTM

PR #7629, opened 2020-09-14
Makefile: Remove broken targets

cevich:
The configuration for this container has moved to the
containers/automation_images repository.  Leverage the
new location to guarantee synchronization with containers
used in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ PROJECT := github.com/containers/podman
 GIT_BASE_BRANCH ?= origin/master
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN ?= $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
-LIBPOD_IMAGE ?= libpod_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 LIBPOD_INSTANCE := libpod_dev
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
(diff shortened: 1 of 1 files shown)


cevich:
@mheon @edsantiago I need some help here: I don't like this fix, any suggestions?

(Context: We removed /Dockerfile b/c it referenced removed packer stuff)

Another idea: Instead of building from the makefile, what if I simply pointed `${LIBPOD_IMAGE}` at 'quay.io/libpod/fedorta_podman:latest'

cevich:
Note to me: There is no `quay.io/libpod/fedora_podman:latest`, would need to update automation to push one (easy task)

mheon:
Is there any value in making the automation repo a submodule of Podman if we need to check it out here?

Makefile
```diff
@@ -266,7 +266,9 @@ clean: ## Clean artifacts
 
 .PHONY: libpodimage
 libpodimage: ## Build the libpod image
-	${CONTAINER_RUNTIME} build -t ${LIBPOD_IMAGE} .
+	-rm -rf $@
+	git clone https://github.com/containers/automation_images.git $@
+	${CONTAINER_RUNTIME} build -t ${LIBPOD_IMAGE} -f podman/Containerfile $@
```
edsantiago, line 271: Sorry - I'm going to need a whole lot more context to understand this. What is this used for? How is it used? What magic setup is needed? I tried running these commands and just get a failure on STEP 1, almost certainly because I'm not defining `BASE_NAME` and `BASE_TAG`. How does this compensate for that?

I'm also uncertain about the dependency between repos: any change to Containerfile in that repo will have to be sync'ed to this Makefile in the podman repo, and worse, possibly that has to be done on all active maintenance branches. Is there any way to build this image from the `automation_images` repo instead? (Is that what your recent comment suggests?)

I'm also really uncomfortable about running a git checkout within a git checkout. Did you consider:
```make
    LIBPODIMAGE_TMP=$(shell mktemp -d -t $@.tmp.XXXXXXX); git clone ... && ${CONTAINER_RUNTIME} ... $$LIBPODIMAGE_TMP; rm -rf $$LIBPODIMAGE_TMP
```

cevich:
>  I'm going to need a whole lot more context to understand this.

@edsantiago context: "Hey cevich! make dbuild is broken" - that's all I got :confused: Looking at the Makefile, it seems several targets try to use the locally built image. Perhaps a secondary-problem:  I'm not aware of any automated tests that check these targets.

> Is there any way to build this image from the automation_images repo instead? (Is that what your recent comment suggests?)

Yes and yes.  Basically I just need to make the build scripting push a `:latest` image if executing against a branch.  Or, point quay.io's auto-builder service at it.  Another idea would be to parse out the `_BUILT_IMAGE_SUFFIX` value from `.cirrus.yml` and use that as the image tag (this would be more synchronized tag than `latest`).

> really uncomfortable about running a git checkout within a git checkout.

Me too, and @mheon idea of a submodule would basically fit the same bill.

cevich:
Clarification: "Another idea would be to parse out the"...from a Makefile target or script.

cevich:
Okay, here's the latest idea: Slurp out the tag from `.cirrus.yml` and use the pre-built image.  This seems to work (I tested `make shell`) but the image isn't a very human-friendly size (i.e. it's huge).

Poking further...a lot of the make targets that use `${LIBPOD_IMAGE}` are horribly broken for use on Fedora (esp. the volume-mounts). They assume you're running as root and always disable SELInux :confounded: 

In fact, I'm not sure how they ever functioned as intended...which begs the question: @mheon can we simply delete these targets :D  That would be the easiest solution of all :smiley:

mheon:
At least one person is using them, but I agree that we need to clean them up. I think the SELinux thing is normal/expected (we can't reasonably expect to relabel the directory the user is building out of) but the rest needs to be fixed so it actually works.

cevich:
> At least one person is using them

@mheon are you sure?  Near as I can tell, none of these targets can or did work as intended, for a very long time.  Besides SELinux, none of them specify a working directory, and so the repository volume-mount is completely useless.  The container will try to run 'make' from `/`.  Even the previous `Dockerfile` specified a working-dir different from the volume-mount (I just checked).  

If it's only one person...can I/we ask them if it's okay to delete:

* `dbuild*`
* `test`
* `integration`
* `shell`
* `testunit`

?

edsantiago:
`dbuild` dates back to a031b83, November 2017, "Initial checkin from CRI-O repo". It has had only the most trivial of edits since then. How sure are we that someone is using this?

cevich:
Update: I killed off `dbuild*` and kinda-sorta mostly fixed the other targets.  The unittests run and pass, but integration fails.  

@edsantiago WDYT about my calling a python script from Makefile to parse `.cirrus.yml`?  I'm slightly nervous because it assumes python3 and python3-pyyaml are both installed.  However, the alternative of parsing YAML with bash or grep/cut seems even more fugly.  I think this is a preferable solution to using 'latest' - since that's a very ambiguous tag (never mind it not currently existing).

mheon:
@edsantiago Pretty sure we have at least one user, because someone asked me about it not working this morning in Google Chat

edsantiago:
@cevich if this is something that runs in CI, we can guarantee having `python3` and any required modules. If this isn't in CI, well, I still think we can assume that the entire known universe has python3 by now.

@mheon excellent, that means we can identify that party and bring them into this conversation. Out of curiosity, do you know if it has *ever* worked for them?

cevich:
@edsantiago my guess is "no", these targets all look to me like they're intended for human-use.  Automation does things differently in each case:

* Build for darwin inside a container (kind of duplicates `podman-remote-darwin`)
* Build all binaries inside a container
* Run only integration tests inside a container
* Run only unit tests inside a container
* Run both unit and integration tests inside a container
* Drop into a shell inside a container

edsantiago:
>my guess is "no",

I'm not sure I understand your response. Your guess about what?

mheon:
@edsantiago Yes, though I understand that he had to make changes to the Makefile to get it working which became https://github.com/containers/podman/pull/7218

cevich:
> I'm not sure I understand your response. Your guess about what?

These targets *do not* run in any automation that I am aware of, so I'm guessing they're intended only for human-consumption.

edsantiago:
My takeaway:
* podman team does not use these
* podman team doesn't even know anything about these

Therefore:
* podman team should not maintain or support these

cevich:
In anticipation of "Okay, let's delete them", I rebased against upstream and nuked these (and two more) broken targets.

edsantiago:
LGTM but could you squash and update the PR title?

Makefile
```diff
@@ -264,46 +263,6 @@ clean: ## Clean artifacts
...
-.PHONY: dbuild-podman-remote
-dbuild-podman-remote: libpodimage
-	${CONTAINER_RUNTIME} run --name=${LIBPOD_INSTANCE} --privileged -v ${PWD}:/go/src/${PROJECT} --rm ${LIBPOD_IMAGE} $(GOBUILD) -ldflags '$(LDFLAGS_PODMAN)' -tags "$(REMOTETAGS)" -o bin/podman-remote ./cmd/podman
-
-.PHONY: dbuild-podman-remote-darwin
-dbuild-podman-remote-darwin: libpodimage
-	${CONTAINER_RUNTIME} run --name=${LIBPOD_INSTANCE} --privileged -v ${PWD}:/go/src/${PROJECT} --rm ${LIBPOD_IMAGE} env GOOS=darwin $(GOBUILD) -ldflags '$(LDFLAGS_PODMAN)' -tags "${REMOTETAGS}" -o bin/podman-remote-darwin ./cmd/podman
-
-.PHONY: test
-test: libpodimage ## Run tests on built image
```
mheon, line 284: We can't ditch `make test` - if we want to remove the dockerized version, we should convert this to run the full suite locally.

cevich: Okay, I can implement this as a PHONY for all the `local*` targets.

cevich:
@mheon @edsantiago okay, this should be good to go.

TomSweeneyRedHat:
@cevich it looks like this has been pending testing for 3 hours with no progress?

cevich:
> @cevich it looks like this has been pending testing for 3 hours with no progress?

The title `[ci skip]` does what it says.  Here, I can do this to make it show green things...

Makefile
```diff
@@ -322,6 +281,9 @@ localunit: test/goecho/goecho varlink_generate
 	$(GO) tool cover -func=${COVERAGE_PATH}/coverprofile > ${COVERAGE_PATH}/functions
 	cat ${COVERAGE_PATH}/functions | sed -n 's/\(total:\).*\([0-9][0-9].[0-9]\)/\1 \2/p'
 
+.PHONY: test
+test: localunit localintegration remoteintegration localsystem remotesystem  ## Run unit, integration, and system tests.
```
edsantiago, line 285: I wonder if `make` will try to parallelize these? That would make for a complete disaster. (I'm not going to even try it, because half of the above tests fail catastrophically on my laptop). Anyhow, I get the feeling that even without parallel execution this will not work for many people.

This is tough. I really like the idea of `make test`, it's just the right thing to do, but I have no way to actually verify or write such a thing.

cevich: Make does not parallelize by default, you have to pass `-j`.

I'm in the same boat, no way to see if this works.  I have the feeling that the number of people who actually use this target is so small that either:

* They will complain if it's broken
* They will fix it if it's broken
* Both :smile:

vrothberg (commented):
Can we run the full CI on this change? Just to be extra sure we don't have any side effects.

cevich:
Yep, if it makes you feel more comfortable :smile:

cevich:
rebased and force-pushed

PR #7633, opened 2020-09-14
Usability: prevent "-l" with arguments

edsantiago:
Add new system check confirming that "podman foo -l arg"
throws an error; and fix lots of instances where code
was not doing this check.

I'll probably need to add something similar for --all but
that can wait.

### DIFF
--- a/cmd/podman/containers/logs.go
+++ b/cmd/podman/containers/logs.go
@@ -37,7 +37,7 @@ var (
 			case registry.IsRemote() && len(args) > 1:
 				return errors.New(cmd.Name() + " does not support multiple containers when run remotely")
 			case logsOptions.Latest && len(args) > 0:
-				return errors.New("no containers can be specified when using 'latest'")
+				return errors.New("--latest and containers cannot be used together")
 			case !logsOptions.Latest && len(args) < 1:
 				return errors.New("specify at least one container name or ID to log")
 			}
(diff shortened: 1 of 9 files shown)


cmd/podman/containers/top.go
```diff
@@ -91,6 +91,9 @@ func top(cmd *cobra.Command, args []string) error {
 	if len(args) < 1 && !topOptions.Latest {
 		return errors.Errorf("you must provide the name or id of a running container")
 	}
+	if len(args) > 0 && topOptions.Latest {
+		return errors.Errorf("--latest and containers cannot be used together")
+	}
```
Luap99, line 96: `podman top` shouldn't check this. This is valid:
```
$ bin/podman top -l pid user
PID   USER
1     root
```

cmd/podman/pods/top.go
```diff
@@ -68,6 +68,9 @@ func top(_ *cobra.Command, args []string) error {
 	if len(args) < 1 && !topOptions.Latest {
 		return errors.Errorf("you must provide the name or id of a running pod")
 	}
+	if len(args) > 0 && topOptions.Latest {
+		return errors.Errorf("--latest and pods cannot be used together")
+	}
```
Luap99, line 73: Same here.

cmd/podman/containers/top.go
edsantiago: Good catch, thank you. Fixed.

rhatdan:
LGTM
@mheon @baude @vrothberg @TomSweeneyRedHat @ashley-cui @QiWang19 PTAL

TomSweeneyRedHat:
LGTM

Issue #7407, opened 2020-08-22
Podman auto-update with --pull=always fails


andrew-kennedy:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I have setup a few rootless containers using systemd with the auto update label, and they fail to update using podman auto-update with errors such as the following.

**Steps to reproduce the issue:**

1. Create various containers in a pod.

2. Use podman generate to create rootless systemd units to start these containers. 

3. Start the containers using `systemctl --user start pod-test-pod.service`

4. Attempt to `podman auto-update`

**Describe the results you received:**
I received many errors like this for each container:

```
	* error auto-updating container "e322c98940c218419dde53384aaf0c639a430b049d2e33b59613dca9e644d7d5": image check for "90a12b5c957bab94f57aadfd9e655d414ab69443495f22ef430152181ab0aede" failed: invalid repository name (90a12b5c957bab94f57aadfd9e655d414ab69443495f22ef430152181ab0aede), cannot specify 64-byte hexadecimal strings
```

**Describe the results you expected:**

I expected the containers to auto update.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.4
API Version:  1
Go Version:   go1.14.6
Built:        Wed Dec 31 16:00:00 1969
OS/Arch:      linux/amd64```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.19, commit: 5dce9767526ed27f177a8fa3f281889ad509fea7'
  cpus: 24
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: yggdrasil.ak.codes
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.7.15-200.fc32.x86_64
  linkmode: dynamic
  memFree: 4345331712
  memTotal: 33604354048
  ociRuntime:
    name: crun
    package: crun-0.14.1-1.fc32.x86_64
    path: /usr/bin/crun
[...]

Issue #7404, opened 2020-08-21
RFE: Sort the API endpoints alphabetically in the documentation.


Luap99:
/kind feature

**Description**

The endpoints inside the RestAPI Documentation at https://docs.podman.io/en/latest/_static/api.html are grouped in categories which is good. But within each category the endpoints seem to be in a random order (see below). It would be nice if they were sorted alphabetically.

![image](https://user-images.githubusercontent.com/45212748/90923667-5c63fe00-e3ee-11ea-9573-95a5f18316b8.png)

rhatdan:
SGTM. Could one of you open a PR to fix this.

Luap99:
I have looked at the swagger documentation but didn't found a way to achieve the desired result. Maybe someone else finds/knows a way to do this.

rhatdan:
@baude @jwhonce Any ideas?

vrothberg:
https://stackoverflow.com/questions/24951268/sort-api-methods-in-swagger-ui looks promising.

vrothberg:
But I don't know how our docs are wired. @baude, in case you tackle it, I'd like to watch over your shoulders so I know what to tweak where.

jwhonce:
@TomSweeneyRedHat @baude 

From a quick read:
 - it appears we need to set properties on the redoc script. Manipulation of the yaml does not change the presentation order
 - we currently retrieve the redoc script on-demand from html

To fix:
 - We need to either set properties when running redoc script
 - Or, build json application combining redoc with properties and our swagger. Application provides most control on future changes and locks us into a version of redoc.
 - Either way, we need to update operationId's in the swagger:operation comments to maintain desired sort order, examples that would work:
   - createNetworkCompat
   - createNetworkLibpod
   - <verb><object><differentiator>

TomSweeneyRedHat:
@jwhonce it doesn't get published to podman.io, it gets published to readthedocs via some make magic that @baude put together.  We then do a redirect of docs.podman.io to readthedocs.  Also readthedocs has some settngs you can select to make that look seamless.

eithe suggestion sounds at least slightly painful, I'll let @baude flip the coin.

rhatdan:
@Luap99 @TomSweeneyRedHat @baude @jwhonce Anyone have a chance to look at this?

TomSweeneyRedHat:
Think we need a decision first from @baude and/or @umohnani8 about which direction to go as requested in @jwhonce 's comment a few back.

baude:
[...]

PR #7636, opened 2020-09-15
run/create: record raw image

vrothberg:
Record the user-specified "raw" image name in the SpecGenerator, so we
can pass it along to the config when creating a container.  We need a
separate field as the image name in the generator may be set to the
ID of the previously pulled image - ultimately the cause of #7404.

Reverting the image name from the ID to the user input would not work
since "alpine" for pulling iterates over the search registries in the
registries.conf but looking up "alpine" normalizes to
"localhost/alpine".

Recording the raw-image name directly in the generator was the best of
the options I considered as no hidden magic from search registries or
normalizations (that may or may not change in the future) can interfere.
The auto-update backend enforces that the raw-image name is a
fully-qualified reference, so we need to worry about that in the front
end.

Fixes: #7407

### DIFF
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -110,7 +110,9 @@ func create(cmd *cobra.Command, args []string) error {
 	}
 
 	imageName := args[0]
+	rawImageName := ""
 	if !cliVals.RootFS {
+		rawImageName = args[0]
 		name, err := pullImage(args[0])
 		if err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


vrothberg:
@mheon @rhatdan PTAL

A ten line change that took me quite a while...

vrothberg:
Changed the system test to prevent future regressions.

rhatdan:
LGTM
Need to get @mheon Approval

mheon:
LGTM

TomSweeneyRedHat:
LGTM
special_testing-rootless is failing, I'm not sure if that's a known issue or not.

Issue #7263, opened 2020-08-08
Regression: /events API response headers aren't flushed until first event


danopia:
/kind bug

**Description**

In Podman 2.0.2 (and likely earlier), requests to `/v1.0.0/libpod/events` would immediately return an HTTP 200 OK response header, and then sleep waiting for the first event to happen.

Starting in Podman 2.0.3, requests to the same API seem to just "hang" without any server response, until something happens that generates an event.

This means that code that processes the event stream can't be assured that the event stream is actually healthy. A possible workaround is doing something that causes events (creating and deleting a pod) in parallel to setting up the event stream.

**Steps to reproduce the issue:**

Assuming Podman 2.0.4:

1. `podman system service tcp:127.0.0.1:8410`

2. `wget -S -O- http://127.0.0.1:8410/v1.0.0/libpod/events`

**Describe the results you received:**

```
Connecting to localhost (localhost)|127.0.0.1|:8410... connected.
HTTP request sent, awaiting response... 
```

**Describe the results you expected:**

```
Connecting to localhost (localhost)|127.0.0.1|:8410... connected.
HTTP request sent, awaiting response... 
  HTTP/1.1 200 OK
  Content-Type: application/json
  Date: Sat, 08 Aug 2020 18:38:38 GMT
  Transfer-Encoding: chunked
```

**Additional information you deem important (e.g. issue happens only occasionally):**

In Podman 2.0.3, the /events API was temporarily broken. This can be worked around by adding `?filter={}`.

Since this is a regression report, here are several docker commands to show the change between versions:

Podman 2.0.2 gives a 200 OK before stalling (GOOD):
```
# docker run --privileged --rm -it mgoltzsche/podman:2.0.2 sh -euxc "podman system service tcp:127.0.0.1:8410 & sleep 1; wget -O- -T2 -S http://127.0.0.1:8410/v1.0.0/libpod/events'?filters={}'"
+ sleep 1
+ podman system service tcp:127.0.0.1:8410
+ wget -O- -T2 -S 'http://127.0.0.1:8410/v1.0.0/libpod/events?filters={}'
Connecting to 127.0.0.1:8410 (127.0.0.1:8410)
  HTTP/1.1 200 OK
  Content-Type: application/json
  Date: Sat, 08 Aug 2020 18:41:42 GMT
  Connection: close
  Transfer-Encoding: chunked
  
writing to stdout
wget: download timed out
```

Podman 2.0.3 stalls before responding at all (BAD):
```
[...]

PR #7637, opened 2020-09-15
events endpoint: header: do not wait for events

vrothberg:
Do not wait for events to occur before writing the OK header.
Events can take an unknown amount of time to occur and clients
do not need to wait until then to know if the connection is
good.

Fixes: #7263


vrothberg:
@baude @jwhonce PTAL

rhatdan:
LGTM

Issue #7639, opened 2020-09-15
rootless:  cgroupsv1: 2.1.0rc1: trying to run systemd in container fails with OCI runtime permission denied error


dustymabe:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Running systemd inside a rootless podman container seems to be broken on cgroupsv1. We added [a test](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/podman/rootless-systemd) to our FCOS CI for this case and [it fails](https://github.com/coreos/fedora-coreos-config/pull/615) with `podman-2.1.0-0.2.rc1.fc32.x86_64`.

You can see the test failure easily by doing:

```
# build container as non-root user
cd $(mktemp -d)
cat <<EOF > Containerfile
FROM registry.fedoraproject.org/fedora:32
RUN dnf -y install systemd httpd \
&& dnf clean all \
&& systemctl enable httpd
ENTRYPOINT [ "/sbin/init" ]
EOF
podman build -t httpd .
```

```
[core@cosa-devsh ~]$ podman run -it --name httpd --publish 8080:80 httpd
Error: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/sys/fs/cgroup/systemd\\\" to rootfs \\\"/var/home/core/.local/share/containers/storage/overlay/a82a726dabf3f6d5116cb7eb7348428a93bbb17d664a54fae12fb7ce20eaa0f3/merged\\\" at \\\"/var/home/core/.local/share/containers/storage/overlay/a82a726dabf3f6d5116cb7eb7348428a93bbb17d664a54fae12fb7ce20eaa0f3/merged/sys/fs/cgroup/systemd\\\" caused \\\"operation not permitted\\\"\"": OCI runtime permission denied error
```

This is not SELinux (I'm running in permissive mode).

In a debug session with some team members it was determined that systemd inside the container really has nothing to do with the problem because it happens even when systemd isn't in the container or trying to be run:

```
[core@cosa-devsh log]$ podman run -it --systemd=always --rm registry.fedoraproject.org/fedora:32
[...]

PR #7640, opened 2020-09-15
Fix --systemd=always regression

rhatdan:
The kernel will not allow you to modify existing mount flags on a volume
when bind mounting it to another place.  Since /sys/fs/cgroup/systemd is
mounted noexec on the host, it needs to be mounted with the same flags
in the rootless container.

Fixes: https://github.com/containers/podman/issues/7639

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -635,7 +635,7 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
 			Destination: "/sys/fs/cgroup/systemd",
 			Type:        "bind",
 			Source:      "/sys/fs/cgroup/systemd",
-			Options:     []string{"bind", "nodev", "nosuid", "rprivate"},
+			Options:     []string{"bind", "nodev", "noexec", "nosuid", "rprivate"},
 		}
 		g.AddMount(systemdMnt)
 		g.AddLinuxMaskedPaths("/sys/fs/cgroup/systemd/release_agent")


rhatdan:
@dustymabe PTAL
@mheon @vrothberg @giuseppe @baude @QiWang19 @TomSweeneyRedHat @jwhonce PTAL

edsantiago:
For future reference: looks like this was overzealous cleanup in #7532

libpod/container_internal_linux.go
```diff
@@ -635,7 +635,7 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
 			Destination: "/sys/fs/cgroup/systemd",
 			Type:        "bind",
 			Source:      "/sys/fs/cgroup/systemd",
-			Options:     []string{"bind", "nodev", "nosuid", "rprivate"},
+			Options:     []string{"bind", "nodev", "noexec", "nosuid", "rprivate"},
```
mheon, line 638: Any chance we can grab flags from the host mount to make sure they match?

edsantiago: I was just starting to look to see if this was possible! This would be the preferred solution.

rhatdan: That would be nice, but I would prefer to do that in a different PR and get this one merged.

edsantiago:
I don't think this needs the `backport` tag:

https://github.com/containers/podman/blob/547464172c09859c2a5cba69e2aea3db91a17000/libpod/container_internal_linux.go#L602-L606

rhatdan:
@edsantiago needs back port to podman 2.1 release. So I guess we should update. Not sure if @mheon Is just going to push everything in master into 2.1 rc2 or not.

mheon:
Yeah, plan is that we don't branch until we have an RC that satisfies FCOS
and Toolbox, so no backports needed yet

TomSweeneyRedHat:
For the short term
LGTM

edsantiago:
/lgtm
/hold
(tests are green; hold is in case @rhatdan wants to wait for feedback from @dustymabe )

libpod/container_internal_linux.go
dustymabe: Do we want to track this work in an issue or something to make sure it doesn't drop through the cracks?

TomSweeneyRedHat: @dustymabe I made a quick and dirty issue: https://github.com/containers/podman/issues/7661

Issue #7651, opened 2020-09-16
`podman images -a` breaks after pulling <image>@<digest>


AkihiroSuda:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman images -a` breaks after pulling IMAGE@DIGEST

**Steps to reproduce the issue:**
```console
$ podman pull docker.io/library/hello-world@sha256:4cf9c47f86df71d48364001ede3a4fcd85ae80ce02ebad74156906caff5378bc
Trying to pull docker.io/library/hello-world@sha256:4cf9c47f86df71d48364001ede3a4fcd85ae80ce02ebad74156906caff5378bc...
Getting image source signatures
Copying blob 0e03bdcc26d7 done  
Copying config bf756fb1ae [======================================] 1.5KiB / 1.5KiB
Writing manifest to image destination
Storing signatures
bf756fb1ae65adf866bd8c456593cd24beb6a0a061dedf42b26a993176745f6b

$ podman images -a
Error: error parsing repository tag "docker.io/library/hello-world:sha256:4cf9c47f86df71d48364001ede3a4fcd85ae80ce02ebad74156906caff5378bc":: invalid reference format
```

**Describe the results you received:**
Failed as shown above

**Describe the results you expected:**
Shouldn't fail

**Output of `podman version`:**

```
Version:      2.1.0-dev
API Version:  1
Go Version:   go1.15.2
Git Commit:   b9c47fa40bc503e155a528174974293c6b053ec2
Built:        Wed Sep 16 18:05:37 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /usr/local/bin/conmon
    version: 'conmon version 2.0.21, commit: 35a2fa83022e56e18af7e6a865ba5d7165fa2a4a'
  cpus: 2
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: journald
  hostname: suda-ws01
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1001
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1001
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.4.0-47-generic
  linkmode: dynamic
[...]

PR #7654, opened 2020-09-16
image list: return all associated names

vrothberg:
Always return all associated names / repo tags of an image and fix a bug
with malformed repo tags.

Previously, Podman returned all names only with `--all` but this flag
only instructs to list intermediate images and should not alter
associated names.  With `--all` Podman queried the repo tags of an image
which splits all *tagged* names into repository and tag which is then
reassembled to eventually be parsed again in the frontend.  Lot's of
redundant CPU heat and buggy as the reassembly didn't consider digests
which ultimately broke parsing in the frontend.

Fixes: #7651

### DIFF
--- a/test/e2e/images_test.go
+++ b/test/e2e/images_test.go
@@ -241,6 +241,38 @@ WORKDIR /test
 		Expect(result.OutputToStringArray()).Should(HaveLen(0), "dangling image output: %q", result.OutputToString())
 	})
 
+	It("podman pull by digest and list --all", func() {
+		// Prevent regressing on issue #7651.
+		digestPullAndList := func(noneTag bool) {
+			session := podmanTest.Podman([]string{"pull", ALPINEAMD64DIGEST})
+			session.WaitWithDefaultTimeout()
+			Expect(session).Should(Exit(0))
+
+			result := podmanTest.Podman([]string{"images", "--all", ALPINEAMD64DIGEST})
+			result.WaitWithDefaultTimeout()
+			Expect(result).Should(Exit(0))
+
+			found, _ := result.GrepString("<none>")
+			if noneTag {
+				Expect(found).To(BeTrue())
+			} else {
+				Expect(found).To(BeFalse())
+			}
+		}
+		// No "<none>" tag as tagged alpine instances should be present.
+		session := podmanTest.Podman([]string{"pull", ALPINELISTTAG})
+		session.WaitWithDefaultTimeout()
+		Expect(session).Should(Exit(0))
+		digestPullAndList(false)
+
+		// Now remove all images, re-pull by digest and check for the "<none>" tag.
+		session = podmanTest.Podman([]string{"rmi", "-af"})
+		session.WaitWithDefaultTimeout()
+		Expect(session).Should(Exit(0))
+
+		digestPullAndList(true)
+	})
+
 	It("podman check for image with sha256: prefix", func() {
 		session := podmanTest.Podman([]string{"inspect", "--format=json", ALPINE})
 		session.WaitWithDefaultTimeout()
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

mtrmac (commented):
- If this is the last caller of `ReposToMap`, shouldn’t it be removed? That would **EDIT <s>allow</s>avoid** similar confusion for any future callers.
- Is it really correct to put digests to the Repo**Tags** field? (I can’t quickly find documentation of the type). See also `ImageToImageSummary`. OTOH `tokenRepoTag` does seem to be ready to handle digests.

vrothberg:
>     * If this is the last caller of `ReposToMap`, shouldn’t it be removed? That would **EDIT allowavoid** similar confusion for any future callers.

Good catch, just removed it.

>     * Is it really correct to put digests to the Repo**Tags** field? (I can’t quickly find documentation of the type). See also `ImageToImageSummary`. OTOH `tokenRepoTag` does seem to be ready to handle digests.

I think that's okay. RepoTags is part of the entities return type where we have full control over. We can change the name to better reflect the semantics (and also document the fields) in the future.

mtrmac:
> RepoTags is part of the entities return type where we have full control over. We can change the name to better reflect the semantics (and also document the fields) in the future.

AFAICT it is visible in the JSON output of `podman images`, but I may well have overlooked something; I didn’t read the code end-to-end.

Either way, displaying unexpected digests is better than not displaying anything at all, so this is clearly an improvement.

test/e2e/images_test.go
```diff
@@ -241,6 +241,17 @@ WORKDIR /test
...
 
+	It("podman pull by digest and list --all", func() {
+		session := podmanTest.Podman([]string{"pull", ALPINEAMD64DIGEST})
+		session.WaitWithDefaultTimeout()
+		Expect(session).Should(Exit(0))
+
+		// Prevent regressing on issue #7651.
+		result := podmanTest.Podman([]string{"images", "--all"})
+		result.WaitWithDefaultTimeout()
+		Expect(result).Should(Exit(0))
```
edsantiago, line 252: You know my feelings about "run a test and only check exit code, and just blindly la-la-la about its actual output". I'm writing my own system test anyway, so this will be covered, but I do wish I could instill the habit of checking results.

edsantiago (approved):
Seems to work. I'm a little confused that when I pull `same-image-I-already-have@same-sha` I then get two lines, one with a `<none>` tag that can only be `rmi'ed` (untagged) by `rmi ...@sha`, but I can live with that.

mheon:
@edsantiago Probably a bug, but a separate one. IMO, file it and we can fix later.

TomSweeneyRedHat:
And targets:  https://bugzilla.redhat.com/show_bug.cgi?id=1879622

TomSweeneyRedHat:
Changes LGTM
A test failed, but it looks to be a network hiccup.

vrothberg:
> And targets: https://bugzilla.redhat.com/show_bug.cgi?id=1879622

Thanks! I referenced the BZ in the backport.

test/e2e/images_test.go
vrothberg: Thanks for the reminder, @edsantiago! It was not on purpose but happened in a hurry. Since it's not merged, I will improve the test.

vrothberg:
> Seems to work. I'm a little confused that when I pull `same-image-I-already-have@same-sha` I then get two lines, one with a `<none>` tag that can only be `rmi'ed` (untagged) by `rmi ...@sha`, but I can live with that.

Good catch! Docker prints the a single item only (in case there's a "real" tag).

vrothberg:
/hold
Don't merge for now. The backport is merged, to fix the more serious error. So we have time here to take care of beautifying.

vrothberg:
> Seems to work. I'm a little confused that when I pull `same-image-I-already-have@same-sha` I then get two lines, one with a `<none>` tag that can only be `rmi'ed` (untagged) by `rmi ...@sha`, but I can live with that.

Got that sorted out as well.

Issue #7628, opened 2020-09-14
mount argument order


darkdragon-001:
/kind bug

**Description**

Order of `type=`, `src=` and `target=` should not matter.

**Steps to reproduce the issue:**

1. Mount with `type=` _after_ `target=`

**Describe the results you received:**

Invalid syntax error.

**Describe the results you expected:**

Should work without problems.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
2.1.0-dev
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.0-dev
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-0.3.dev.git5a6b2ac.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21-dev, commit: 5c03a23398b94cf869f071128631ef7fd9153b3b'
  cpus: 8
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: file
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.8-300.fc33.x86_64
  linkmode: dynamic
  memFree: 31894863872
  memTotal: 41713786880
  ociRuntime:
    name: runc
    package: runc-1.0.0-279.dev.gitdedadbf.fc33.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc92+dev
      commit: c9a9ce0286785bef3f3c3c87cd1232e535a03e15
      spec: 1.0.2-dev
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.1.4-1.fc33.x86_64
    version: |-
      slirp4netns version 1.1.4+dev
      commit: eecccdb96f587b11d7764556ffacfeaffe4b6e11
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 3
      libseccomp: 2.5.0
  swapFree: 4294963200
  swapTotal: 4294963200
  uptime: 27m 19.88s
registries:
  search:
[...]

PR #7655, opened 2020-09-16
--mount: support arbitrary mount-argument order

vrothberg:
Support an arbitrary order in which arguments are specified to the
`--mount` flag.  Previously, Podman expected `type=...` to come
first which was breaking compatibility with Docker.

Note that this is the ground work to default to "volume" (again Docker
compat).  However, this will require some further massaging as we have
to assign a name.

Fixes: #7628

### DIFF
--- a/cmd/podman/common/volumes.go
+++ b/cmd/podman/common/volumes.go
@@ -28,6 +28,7 @@ var (
 	errDuplicateDest = errors.Errorf("duplicate mount destination")
 	optionArgError   = errors.Errorf("must provide an argument for option")
 	noDestError      = errors.Errorf("must set volume destination")
+	errInvalidSyntax = errors.Errorf("incorrect mount format: should be --mount type=<bind|tmpfs|volume>,[src=<host-dir|volume-name>,]target=<ctr-dir>[,options]")
 )
 
 // Parse all volume-related options in the create config into a set of mounts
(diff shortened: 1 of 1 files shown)


vrothberg:
@baude @mheon @rhatdan PTAL

mheon:
LGTM

cmd/podman/common/volumes.go
```diff
@@ -156,25 +178,13 @@ func getMounts(mountFlag []string) (map[string]spec.Mount, map[string]*specgen.N
...
-	errInvalidSyntax := errors.Errorf("incorrect mount format: should be --mount type=<bind|tmpfs|volume>,[src=<host-dir|volume-name>,]target=<ctr-dir>[,options]")
-
-	// TODO(vrothberg): the manual parsing can be replaced with a regular expression
-	//                  to allow a more robust parsing of the mount format and to give
-	//                  precise errors regarding supported format versus supported options.
 	for _, mount := range mountFlag {
-		arr := strings.SplitN(mount, ",", 2)
-		if len(arr) < 2 {
-			return nil, nil, errors.Wrapf(errInvalidSyntax, "%q", mount)
+		// TODO: Docker defaults to "volume" if no mount type is specified.
```
rhatdan, line 182: Why not just make this default in findMountType?

vrothberg: See commit message:

> Note that this is the ground work to default to "volume" (again Docker
compat). However, this will require some further massaging as we have
to assign a name.

Defaulting to "volume" fails as volumes need to be named. I am sure we can tackle it much easier now but refrained from doing that in the bug fix.

mheon: Volumes don't technically need to be named. If you pass libpod a named volume with no name, we will make an anonymous volume (pseudorandom ID as name)

vrothberg: Curious. I got an error when playing with the default. Maybe it was a red herring.

TomSweeneyRedHat:
LGTM
I think we can move this along now with a follow up if necessary from @vrothberg

PR #7658, opened 2020-09-16
[v2.0.5-rhel] image list: return all associated names

vrothberg:
Always return all associated names / repo tags of an image and fix a bug
with malformed repo tags.

Previously, Podman returned all names only with `--all` but this flag
only instructs to list intermediate images and should not alter
associated names.  With `--all` Podman queried the repo tags of an image
which splits all *tagged* names into repository and tag which is then
reassembled to eventually be parsed again in the frontend.  Lot's of
redundant CPU heat and buggy as the reassembly didn't consider digests
which ultimately broke parsing in the frontend.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1879622

### DIFF
--- a/libpod/image/utils.go
+++ b/libpod/image/utils.go
@@ -86,33 +86,6 @@ func hasTransport(image string) bool {
 	return strings.Contains(image, "://")
 }
 
-// ReposToMap parses the specified repotags and returns a map with repositories
-// as keys and the corresponding arrays of tags or digests-as-strings as values.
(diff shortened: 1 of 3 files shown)


vrothberg:
GitHub action to prefix the title is failing. I'll take a look tomorrow.

mheon:
I think we can merge without the check succeeding (success task doesn't depend on it) which is good.

Issue #7196, opened 2020-08-03
[APIv2] Container log not provided


psakar:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Container log not provided 
**Steps to reproduce the issue:**

Docker

CLI
docker create --interactive --tty --name busybox docker.io/library/busybox:latest sh -c "echo Hello"
docker start busybox
docker logs busybox
docker inspect busybox > /tmp/test-busybox-docker.txt
docker rm busybox 

REST
curl --unix-socket /var/run/docker.sock -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/create?name=busybox' --data '{"name":"busybox2","Image":"docker.io/library/busybox:latest", "Tty":true, "Cmd":["sh", "-c", "echo Hello"]}'
curl --unix-socket /var/run/docker.sock -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/busybox/start'
curl --unix-socket /var/run/docker.sock --output - -H "Content-Type: application/json" "http://127.0.0.1:8880/containers/busybox/logs?stdout=true&stderr=true"
docker logs busybox
docker inspect busybox > /tmp/test-busybox-docker-api.txt
curl --unix-socket /var/run/docker.sock -X DELETE -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/busybox'

Podman

podman --log-level=debug system service -t 0 tcp:localhost:8880 2>&1>/tmp/podman-rest.log &

CLI
podman create --interactive --tty --name busybox docker.io/library/busybox:latest sh -c "echo Hello"
podman start busybox
podman logs busybox
podman inspect busybox > /tmp/test-busybox-podman.txt
podman rm busybox 

REST
curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/create?name=busybox' --data '{"name":"busybox2","Image":"docker.io/library/busybox:latest", "Tty":true, "Cmd":["sh", "-c", "echo Hello"]}'
curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/busybox/start'
curl --output - -H "Content-Type: application/json" "http://127.0.0.1:8880/containers/busybox/logs?stdout=true&stderr=true"
podman logs busybox
podman inspect busybox > /tmp/test-busybox-podman-api.txt
[...]

PR #7660, opened 2020-09-16
Fix remote logs

ashley-cui:
Docker compatibility  - logs endpoint does not write stream headers if container has a tty

There are still some issues with logging behavior that affects this, we need to sort that out in the backend before we can know for certain if this is the right fix

Fixes: https://github.com/containers/podman/issues/7196

### DIFF
--- a/pkg/api/handlers/compat/containers_logs.go
+++ b/pkg/api/handlers/compat/containers_logs.go
@@ -105,6 +105,18 @@ func LogsFromContainer(w http.ResponseWriter, r *http.Request) {
 
 	var frame strings.Builder
 	header := make([]byte, 8)
+
+	writeHeader := true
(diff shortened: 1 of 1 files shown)


ashley-cui:
@vrothberg tagging so you have the link to this PR, thanks for helping me on this

ashley-cui:
Update: this should fix remote logs the best we can: adding a tty to a container makes the logs add return carriage symbols, which is a tty quirk, so this should be an accpetable fix. See: https://github.com/moby/moby/issues/37366#issuecomment-401157643

rhatdan:
LGTM
@mheon @jwhonce @baude PTAL

vrothberg (commented):
LGTM but I want a head not from @baude and/or @jwhonce

Issue #7656, opened 2020-09-16
restartPolicy doesn't seem to work with `podman play kube`


dtitov:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I'd like to achieve the behavior when failed containers (exited with non-zero exit code) are automatically restarted by Podman in a deployment run by `podman play kube`. AFAIK, the default option is `always` (if nothing else is explicitly specified). However, when my container dies (`Exited (1) 10 minutes ago`), it doesn't get automatically restarted. I tried explicitly adding `always` option: `restartPolicy: Always`, but it changed nothing, exited container remains exited without any visible attempts to restart it.

**Steps to reproduce the issue:**

1. Create a Kubernetes file with a pod having a container that fails upon the start.

2. `podman play kube`

**Describe the results you received:**
 Container fails and doesn't get restarted.

**Describe the results you expected:**
Podman tries to restart the exited container.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:            1.6.4
RemoteAPI Version:  1
Go Version:         go1.13.4
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.13.4
  podman version: 1.6.4
host:
  BuildahVersion: 1.12.0-dev
  CgroupVersion: v1
  Conmon:
    package: conmon-2.0.6-1.module_el8.2.0+305+5e198a41.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.6, commit: a2b11288060ebd7abd20e0b4eb1a834bbf0aec3e'
  Distribution:
    distribution: '"centos"'
    version: "8"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 11370590208
  MemTotal: 16644763648
  OCIRuntime:
    name: runc
    package: runc-1.0.0-65.rc10.module_el8.2.0+305+5e198a41.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.1-dev'
  SwapFree: 0
  SwapTotal: 0
  arch: amd64
  cpus: 4
  eventlogger: journald
[...]

PR #7671, opened 2020-09-17
handle the restartPolicy for play kube and generate kube

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/7656

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -299,6 +299,18 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
 		return nil, err
 	}
 
+	var ctrRestartPolicy string
+	switch podYAML.Spec.RestartPolicy {
+	case v1.RestartPolicyAlways:
+		ctrRestartPolicy = libpod.RestartPolicyAlways
+	case v1.RestartPolicyOnFailure:
+		ctrRestartPolicy = libpod.RestartPolicyOnFailure
+	case v1.RestartPolicyNever:
+		ctrRestartPolicy = libpod.RestartPolicyNo
+	default: // Default to Always
+		ctrRestartPolicy = libpod.RestartPolicyAlways
+	}
+
 	containers := make([]*libpod.Container, 0, len(podYAML.Spec.Containers))
 	for _, container := range podYAML.Spec.Containers {
 		pullPolicy := util.PullImageMissing
@@ -326,6 +338,7 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
 		if err != nil {
 			return nil, err
 		}
+		conf.RestartPolicy = ctrRestartPolicy
 		ctr, err := createconfig.CreateContainerFromCreateConfig(ctx, ic.Libpod, conf, pod)
 		if err != nil {
 			return nil, err
--- a/test/e2e/play_kube_test.go
+++ b/test/e2e/play_kube_test.go
@@ -46,6 +46,7 @@ metadata:
 {{ end }}
 
 spec:
+  restartPolicy: {{ .RestartPolicy }}
   hostname: {{ .Hostname }}
   hostAliases:
 {{ range .HostAliases }}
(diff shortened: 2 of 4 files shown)


pkg/domain/infra/abi/play.go
```diff
@@ -299,6 +299,18 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
...
+	var ctrRestartPolicy string
+	switch podYAML.Spec.RestartPolicy {
+	case v1.RestartPolicyAlways:
+		ctrRestartPolicy = libpod.RestartPolicyAlways
+	case v1.RestartPolicyOnFailure:
+		ctrRestartPolicy = libpod.RestartPolicyOnFailure
+	case v1.RestartPolicyNever:
+		ctrRestartPolicy = libpod.RestartPolicyNo
+	default: // Default to Always
+		ctrRestartPolicy = libpod.RestartPolicyAlways
```
mheon, line 311: Is this the default? I remember the upstream issue mentioning OnFailure was the default for Kube. @haircommander Would you happen to know this?

mheon:
/approve
One question but otherwise LGTM

pkg/domain/infra/abi/play.go
zhangguanzhang:
```
[root@k8s-m1 ~]# kubectl explain pod.spec.restartPolicy
KIND:     Pod
VERSION:  v1

FIELD:    restartPolicy <string>

DESCRIPTION:
     Restart policy for all containers within the pod. One of Always, OnFailure,
     Never. Default to Always. More info:
     https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy
```

test/e2e/play_kube_test.go
```diff
@@ -293,13 +296,14 @@ type HostAlias struct {
...
 	p := Pod{
-		Name:        defaultPodName,
-		Hostname:    "",
-		HostAliases: nil,
-		Ctrs:        make([]*Ctr, 0),
-		Volumes:     make([]*Volume, 0),
-		Labels:      make(map[string]string),
-		Annotations: make(map[string]string),
+		Name:          defaultPodName,
+		RestartPolicy: "",
```
haircommander, line 300: we may want the default here to be `never` as that matches the behavior the tests were written to expect

zhangguanzhang: got it

rhatdan:
Could you make sure that the restart policy is also written to the yaml file if `podman generate kube` is run?

haircommander:
> Could you make sure that the restart policy is also written to the yaml file if `podman generate kube` is run?

a difficulty here is podman's restartPolicy is container level, k8s is pod level.

mheon:
Restart policy for pods would be a fun RFE, I might get ambitious and code that up at some point.

zhangguanzhang:
> Could you make sure that the restart policy is also written to the yaml file if `podman generate kube` is run?

It had be done

zhangguanzhang:
ci had error, but no reason 😥, @vrothberg @TomSweeneyRedHat  PTAL

vrothberg:
Restarted the job. Thanks for the ping!

zhangguanzhang:
> Restarted the job. Thanks for the ping!

thanks

pkg/domain/infra/abi/play.go
dtitov: @mheon, if you are referring to my issue, then I took this info from the [this doc](http://docs.podman.io/en/latest/markdown/podman-generate-systemd.1.html):
```
–restart-policy=policy

Set the systemd restart policy. The restart-policy must be one of: “no”, “on-success”, “on-failure”, “on-abnormal”, “on-watchdog”, “on-abort”, or “always”. The default policy is on-failure.
```

But now I see that it's a wrong doc. In Kubernetes the default restart policy is indeed `Always`.

rhatdan:
Thanks @zhangguanzhang 
/lgtm

Issue #7673, opened 2020-09-17
podman pod create --infra-command: does not seem to be working


edsantiago:
I think #7621 may have been incompletely tested?
```console
$ ./bin/podman pod create --name=mypod --infra-command=/pause
c572562a28a4c5fe6c5b61cec38144aeb127efc76f2892ea71801aa8ae9225d7
$ ./bin/podman run --pod mypod alpine true
Error: cannot get namespace path unless container 39ce86bd690311fb5501a944b210ebca346b86a990b439ad6f8386ac6821bdca is running: container is stopped
$ ./bin/podman logs 39c
{"msg":"executable path not specified","level":"error","time":"2020-09-17T14:59:59.000091139Z"}
$  ./bin/podman inspect --format '{{.Config.Entrypoint}}' 39c
/pause
```
master @ 031ddf9c8476202c6b0c810c6c0f6a7a0040c035; rootless on f32

vrothberg:
@rhatdan PTAL

edsantiago:
FWIW I see one difference between `podman pod create` and `podman pod create --infra-command=/pause`:

This is a normal, good, infra container:
```console
$ ./bin/podman inspect 9db --format '{{.Path}}:{{.Args}}'
/pause:[/pause]
```
This is with `--infra-command=/pause` (which should theoretically be a NOP):
```console
$ ./bin/podman inspect 5e2 --format '{{.Path}}:{{.Args}}'
:[] 
```

PR #7675, opened 2020-09-17
fix the .Path and .Args when use the infra-command

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/7673

### DIFF
--- a/libpod/runtime_pod_infra_linux.go
+++ b/libpod/runtime_pod_infra_linux.go
@@ -50,7 +50,11 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
 				entryPoint = config.Entrypoint
 				entryCmd = config.Entrypoint
 			}
+		} else { // so use the InfraCommand
+			entrypointSet = true
+			entryCmd = entryPoint
 		}
+
 		if len(config.Cmd) > 0 {
 			// We can't use the default pause command, since we're
 			// sourcing from the image. If we didn't already set an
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@zhangguanzhang Remote tests do not seem to be happy

zhangguanzhang:
I will fix this when I have time

edsantiago:
All tests are unhappy; but FWIW it now passes my (new, unsubmitted) `--infra-image + --infra-command` test. You're on the right track, thank you!

edsantiago:
I have a system test (BATS) that tests this fairly thoroughly, so don't feel any need to write one. I will submit my PR as soon as yours merges (because, obviously, it will not pass on master right now).

zhangguanzhang:
> 我有一个系统测试（BATS），可以对它进行全面的测试，因此不需要编写任何文件。您的合并后，我会立即提交我的PR（因为显然，它现在不会继续通过Master）。

Does this mean I should remove the e2e I wrote？

edsantiago:
No, those are a good idea - multiple levels of testing are good. What I meant was, if you were thinking of writing a system test, you don't need to :grin:

zhangguanzhang:
My local test is normal, I don’t know why Args is empty in ci

edsantiago:
You are using `Expect(check1` when you should be using `check2` :wink:

zhangguanzhang:
> You are using `Expect(check1` when you should be using `check2` 😉

oops, wrong copy😅

edsantiago (approved):
I tested against my new BATS pod infra tests; all passes. Nice work.

rhatdan:
Thanks @zhangguanzhang 
/lgtm

PR #7679, opened 2020-09-17
enable --iidfile for podman-remote build

baude:
for podman-remote build operations, the iidfile, when used, needs to write the file to the client's local filesystem.

### DIFF
--- a/pkg/domain/infra/tunnel/images.go
+++ b/pkg/domain/infra/tunnel/images.go
@@ -306,7 +306,22 @@ func (ir *ImageEngine) Config(_ context.Context) (*config.Config, error) {
 }
 
 func (ir *ImageEngine) Build(_ context.Context, containerFiles []string, opts entities.BuildOptions) (*entities.BuildReport, error) {
-	return images.Build(ir.ClientCxt, containerFiles, opts)
+	report, err := images.Build(ir.ClientCxt, containerFiles, opts)
+	if err != nil {
+		return nil, err
+	}
+	// For remote clients, if the option for writing to a file was
+	// selected, we need to write to the *client's* filesystem.
+	if len(opts.IIDFile) > 0 {
+		f, err := os.Create(opts.IIDFile)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


jwhonce:
@baude Code looks good, but I'm wondering if it should live in the bindings? Or even in the cmd tree vs. this deep in libraries?

baude:
i personally think this is fine ... if we have strong feelings, it can be moved.

mheon:
Bindings feels a little wierd - they already pass back the ID, I don't think there's a need for them to also make a file with it.

TomSweeneyRedHat:
LGTM

mheon:
LGTM

pkg/domain/infra/tunnel/images.go
```diff
@@ -306,7 +306,22 @@ func (ir *ImageEngine) Config(_ context.Context) (*config.Config, error) {
 }
 
 func (ir *ImageEngine) Build(_ context.Context, containerFiles []string, opts entities.BuildOptions) (*entities.BuildReport, error) {
-	return images.Build(ir.ClientCxt, containerFiles, opts)
+	report, err := images.Build(ir.ClientCxt, containerFiles, opts)
```
rhatdan, line 309: I think this should just be done in the client, and should not be supported as a BuildOption, IE IIDFile should just be done in podman client.

vrothberg: Although it would be cleaner in the client, our hands are a bit tied as `entities.BuildahOptions` is just an alias for the one from `imagebuildah` (see below). Since the buildah code is taking care of creating the file it (for the local client), we had to special case the iidfile creation in the client for the remote one. I think it may be easier to just leave the code here.

```golang 
// BuildOptions describe the options for building container images.
type BuildOptions struct {                                         
        imagebuildah.BuildOptions                                  
}                                                                  
```

vrothberg (commented):
LGTM

rhatdan:
Not thrilled but I guess our hands are forced.
/hold cancel

PR #7687, opened 2020-09-18
system tests: new tests

edsantiago:
- podman network create: new test

- podman pull by-sha + podman images -a (#7651)

- podman image mount: new test

- podman pod: --infra-image and --infra-command (#7167)

For convenience and robustness, build a new testimage
containing a custom file /home/podman/testimage-id
with contents YYYYMMDD (same as image tag). The
image-mount test checks that this file exists and
has the desired content. New testimage also includes
a dummy 'pause' executable, for testing pod infra.

Updates from testimage:20200902 to :20200917

### DIFF
--- a/test/system/010-images.bats
+++ b/test/system/010-images.bats
@@ -112,4 +112,51 @@ Labels.created_at | 20[0-9-]\\\+T[0-9:]\\\+Z
     run_podman rm  mytinycontainer
 }
 
+# Regression test for https://github.com/containers/podman/issues/7651
+# in which "podman pull image-with-sha" causes "images -a" to crash
+@test "podman images -a, after pulling by sha " {
+    # Get a baseline for 'images -a'
+    run_podman images -a
+    local images_baseline="$output"
+
+    # Get the digest of our local test image. We need to do this in two steps
+    # because 'podman inspect' only works reliably on *IMAGE ID*, not name.
+    # See https://github.com/containers/podman/issues/3761
+    run_podman inspect --format '{{.Id}}' $IMAGE
+    local iid="$output"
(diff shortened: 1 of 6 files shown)


vrothberg (commented):
LGTM, nice work!

edsantiago:
```
Error: network 10.254.134.0/24 is already being used by a cni configuration
```
Sigh, I was afraid of that. I'll have to get back to this on Monday. ITM, any suggestions on how to test for a safe rfc1918 subnet?

mheon:
I believe there's a file somewhere in the repo that we use to give Cirrus VMs a list of unusable subnets for `network create` (nets already in use by GCP) - maybe consulting with that would help reduce the possibility of this?

edsantiago:
Unfortunately that doesn't help when running tests in the Fedora or RHEL gating environments. I need a bulletproof way to check the local environment.

edsantiago:
Looking at [contrib/cirrus/99-do-not-use-google-subnets.conflist](https://github.com/containers/podman/blob/master/contrib/cirrus/99-do-not-use-google-subnets.conflist) I found that the whole `10.128-10.255` block is reserved for the Google environment. Since this may be subject to change without notice, I altered the network test to use the class B block (`172.16-172.31`) instead. This is less common, but still not any sort of guarantee that the subnet is free. 

I considered adding retries, detecting "being used by a cni configuration" and rerunning the random_rfc1918 loop, but that's still fragile and introduces its own risk.

I considered parsing the output of `podman network ls` -> `podman network inspect` and, um, yeah, no.

Any better ideas?

jwhonce:
We can refine if network selection becomes an issue

/lgtm

Issue #7664, opened 2020-09-17
[low pri] podman image umount -a: emits unexpected output


edsantiago:
`podman umount -a` emits one line of output (image ID) for each available image, no matter whether they're mounted, not mounted, never been mounted:
```console
# podman rmi -a
# podman pull -q quay.io/libpod/testimage:20200916
610a6ea7b347b81b2051aa364f80e5cd08863760605f3dd76207df0ec6a24bae
# podman image umount -a
610a6ea7b347b81b2051aa364f80e5cd08863760605f3dd76207df0ec6a24bae
# podman pull -q quay.io/libpod/testimage:20200902
423caf5c7e48eb8222662cf77347740cc5bdd326d937563aa3647c5ccbe76759
# podman image umount -a
610a6ea7b347b81b2051aa364f80e5cd08863760605f3dd76207df0ec6a24bae
423caf5c7e48eb8222662cf77347740cc5bdd326d937563aa3647c5ccbe76759
# podman image mount
[ no output]
```
`podman-image-unmount(1)` does not describe this behavior. I find it weird: it's fine to output an image ID if it was actually mounted, but this doesn't fit into my preconceived expectations. If it *is* the desired behavior, could it be documented?

master @ 8d7e7954546efb344d677e0c3c86d9d045a3de4f

rhatdan:
Yes it should be cleaned up to match the mount command.  The thing is it could get slow.  Since umount is not guaranteed to fully unmount the mount point.  Since the mount point could be mounted multiple times.  Currently I think 
`podman container unmount` and `podman image unmount` are just listing the images that it called unmount on.  I don't think containers/storage gives us an error if a container or image is not mounted when you call unmount, it just returns success.

rhatdan:
This is fixed in the main branch.

PR #7691, opened 2020-09-18
Fix podman image unmount to only report images unmounted

rhatdan:
Currently `podman image unmount` report every image that is mounted
when it unmounts them. We should only report unmounted actually mounted images.


rhatdan:
@edsantiago PTAL

rhatdan:
Fixes:https://github.com/containers/podman/issues/7664

rhatdan:
Fixes:https://github.com/containers/podman/issues/7664

TomSweeneyRedHat:
tests look awful red here @rhatdan

PR #7694, opened 2020-09-18
Preserve groups in exec sessions in ctrs with --user

mheon:
Podman wants to guarantee that exec sessions retain the groups of the container they are started in, unless explicitly overridden
by the user. This guarantee was broken for containers where the `--user` flag was specified; this patch resolves that.

Somewhere in the Exec rewrite for APIv2, I changed the location where the container's User is passed into the exec session (similar to groups, we also want to preserve user unless overridden). The lower-level Exec APIs already handled setting user and group appropriately if not specified when the exec session was created, but I added duplicate code to handle this higher in the stack - and that code only handled setting user, not supplemental groups, breaking support in that specific case.

Two things conspired to make this one hard to track down: first, things were only broken if the container explicitly set a user; otherwise, the container user would still appear to be unset to the lower-level code, which would properly set supplemental
groups (this tricked our existing test into passing). Also, the `crun` OCI runtime will add the groups without prompting, which further masked the problem there. I debated making `runc` do the same, but in the end it's better to fix this in Podman - it's better to be explicit about what we want done so we will work with all OCI runtimes.

### DIFF
--- a/libpod/container_exec.go
+++ b/libpod/container_exec.go
@@ -980,11 +980,6 @@ func prepareForExec(c *Container, session *ExecSession) (*ExecOptions, error) {
 		capList = capabilities.AllCapabilities()
 	}
 
-	user := c.config.User
-	if session.Config.User != "" {
-		user = session.Config.User
-	}
-
 	if err := c.createExecBundle(session.ID()); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


jwhonce:
LGTM

TomSweeneyRedHat:
All kinds of red in the tests @mheon

mheon:
One flake, otherwise green.

PR #7708, opened 2020-09-21
Cirrus: Fix decrypt failure

cevich:
I Noted that the last cron-based branch-build had failed with an error from Cirrus-CI regarding decryption of variables.  Upon examination, found several encrypted values to be invalid.  Replaced the critical ones, and disabled any useless/disabled values/facilities (i.e. IRC notifications, and GCS release uploads).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -55,13 +55,13 @@ env:
     #### Credentials and other secret-sauces, decrypted at runtime when authorized.
     ####
     # Freenode IRC credentials for posting status messages
-    IRCID: ENCRYPTED[e87bba62a8e924dc70bdb2b66b16f6ab4a60d2870e6e5534ae9e2b0076f483c71c84091c655ca239101e6816c5ec0883]
+    IRCID: disabled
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mheon @baude @vrothberg @edsantiago PTAL

mheon:
/approve
LGTM

edsantiago:
I have no way of reviewing this. I have confirmed that the `a28` string appears in the `master` copy of `.cirrus.yml`, in a context that suggests that it is equivalent usage to the one in this branch. None of the other strings (GCP) appear in master. I can only assume that if CI passes, the strings must be valid.

cevich:
> I can only assume that if CI passes, the strings must be valid.

This is the case.  I found in my investigation, if ***ANY*** of the strings fail to decrypt, Cirrus-CI will not dercrypt anything and just  try to pass the `ENCRYPTED[blah1234]`...causing things to break very badly.

Thanks for taking a look @edsantiago

PR #7712, opened 2020-09-21
add missing return for compat kill

baude:
on an error condition in kill for the compatibility layer, we were missing a return.


mheon:
LGTM. Really wish we had a linter that could catch this for us

Issue #7685, opened 2020-09-18
[makefile] - /bin/sh: c: command not found


zhangguanzhang:
```
$ make help
/bin/sh: c: command not found
make: [Makefile:140: help] Error 127 (ignored)
```

rhatdan:
Works for me on Fedora 33
```
make help
Usage: make <target>
 * 'binaries' - Build podman
 * 'changelog' - Generate changelog
 * 'clean' - Clean artifacts
 * 'docs' - Generate documentation
 * 'gofmt' - Verify the source code gofmt
 * 'install' - Install binaries to system locations
 * 'integration' - Execute integration tests
 * 'libpodimage' - Build the libpod image
 * 'local-cross' - Cross local compilation
 * 'package' - Build rpm packages
 * 'package-install' - Install rpm packages
 * 'shell' - Run the built image and attach a shell
 * 'test' - Run tests on built image
 * 'testunit' - Run unittest on the built image
 * 'varlink_generate' - Generate varlink
```

zhangguanzhang:
```
[root@Centos8 ~]# cd podman
[root@Centos8 ~/podman]# make help
/bin/sh: c: command not found
make: [Makefile:140: help] Error 127 (ignored)
[root@Centos8 ~/podman]# make --version
GNU Make 4.2.1
Built for x86_64-redhat-linux-gnu
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
[root@Centos8 ~/podman]# cat /etc/redhat-release 
CentOS Linux release 8.2.2004 (Core)
```

afbjorklund:
Most likely you don't have python installed ?

```Makefile
help:
        @$(PYTHON) -c "$$PRINT_HELP_PYSCRIPT" < $(MAKEFILE_LIST)
```

Reproducer:
```
 make help PYTHON=""
/bin/sh: 1: c: not found
make: [Makefile:140: help] Error 127 (ignored)

```

zhangguanzhang:
> Most likely you don't have python installed ?
> 
> ```makefile
> PYTHON ?= $(shell command -v python3 python|head -n1)
> help:
>         @$(PYTHON) -c "$$PRINT_HELP_PYSCRIPT" < $(MAKEFILE_LIST)
> ```
> 
[...]

PR #7714, opened 2020-09-21
Print nice error message when python is not installed

rhatdan:
If user runs `make help` on a system without python, we should print a
decent message.

Fixes: https://github.com/containers/podman/issues/7685

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,8 +135,13 @@ export PRINT_HELP_PYSCRIPT
 err_if_empty = $(if $(strip $($(1))),$(strip $($(1))),$(error Required variable $(1) value is undefined, whitespace, or empty))
 
 .PHONY: help
+ifneq (, ${PYTHON})
 help:
 	@$(PYTHON) -c "$$PRINT_HELP_PYSCRIPT" < $(MAKEFILE_LIST)
+else
+help:
+	$(error python required for 'make help', executable not found)
+endif
 
 .gopathok:
 ifeq ("$(wildcard $(GOPKGDIR))","")


Makefile
```diff
@@ -135,8 +135,13 @@ export PRINT_HELP_PYSCRIPT
 err_if_empty = $(if $(strip $($(1))),$(strip $($(1))),$(error Required variable $(1) value is undefined, whitespace, or empty))
 
 .PHONY: help
+ifneq (, ${PYTHON})
```
TomSweeneyRedHat, line 138: Does this work if Python isn't installed?  I'm getting:

```
# python3 python|head -n1
python3: can't open file 'python': [Errno 2] No such file or directory
```
and think this will force the else to fire, or am I badly parsing the command Makefile will call?

rhatdan: Yes that is the idea.

TomSweeneyRedHat: Right, and this is probably my confusion, but I thought on a no python machine this was resolving to `if '' ne 'python3: can't open...'` and then on a machine with python `if '' ne 'pythonv3.1' so the else statement wouldn't be ever entered?

edsantiago: @TomSweeneyRedHat where are you seeing that `can't open file`? Every way I've tried to test the `PYTHON` definition:

https://github.com/containers/podman/blob/fffcc25d8dfab8c6059c229177286f462d909b8d/Makefile#L33

...gives the right result. `command -v` on bash and even `/bin/sh` on fedora and alpine returns empty string if neither python3 nor python are installed. Can you give more context on the failure you're seeing?

rhatdan: Correct, here is an example, added the "<" and ">" to help you visulize
```
echo "<$(command -v python4 | head -n1)>"
<>
```

rhatdan:
@TomSweeneyRedHat @baude @mheon @ashley-cui @QiWang19 @jwhonce @giuseppe @vrothberg @saschagrunert @zhangguanzhang PTAL

vrothberg (commented):
LGTM

Would love a head nod from @jwhonce

saschagrunert (approved):

Makefile
TomSweeneyRedHat: @edsantiago I may have shot myself in the foot, I'd dropped the 'command -v` part of the command to get that error.
@rhatdan thanks for the clarification.

TomSweeneyRedHat (commented):
LGTM

zhangguanzhang:
test ok in my machine，LGTM

PR #7716, opened 2020-09-21
podman-remote untag image is broken

rhatdan:
There seems to be some questionable code in NamesHistory which is not
returning the true value of the history.  Basically this ends up truncating
the last item on the list which causes the test to fail.

Not sure if this code is still necessary or what it breaks.

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -566,19 +566,22 @@ func UntagImage(w http.ResponseWriter, r *http.Request) {
 		utils.ImageNotFound(w, name, errors.Wrapf(err, "Failed to find image %s", name))
 		return
 	}
-	tag := "latest"
-	if len(r.Form.Get("tag")) > 0 {
-		tag = r.Form.Get("tag")
+	tags := []string{}
+	if len(r.Form.Get("repo")) > 0 {
+		repo := r.Form.Get("repo")
+		tag := "latest"
+		if len(r.Form.Get("tag")) > 0 {
+			tag = r.Form.Get("tag")
+		}
+		tags = []string{fmt.Sprintf("%s:%s", repo, tag)}
+	} else {
+		tags = newImage.Names()
 	}
-	if len(r.Form.Get("repo")) < 1 {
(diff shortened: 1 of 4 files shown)


rhatdan:
@mheon PTAL,

mheon:
I think we need @vrothberg or @baude or someone else more familiar with the image code to comment here... I don't really know why this was added. Tests are red without it, though.

vrothberg:
@saschagrunert added that code. It sounds reasonable to not display a name twice.

I am currently trying to reproduce the untag issue as I fail to see a relation at the moment.

vrothberg:
It looks like a symptomatic fix. Local and remote untag should execute the same code. If there's a hiccup in remote, I suspect there's something else going on.

vrothberg:
Looking closer at the untag endpoint, there are a couple issues:

* It always requires a `repo` which the local client does not. `podman untag $ID` will remove _all_ tags. That's currently not working and maybe something we need to add to the tests.
* There are parsing errors:
```
podman (master) $ ./bin/podman-remote images
REPOSITORY                TAG     IMAGE ID      CREATED       SIZE
docker.io/library/alpine  latest  a24bb4013296  3 months ago  5.85 MB
docker.io/library/alpine  2       a24bb4013296  3 months ago  5.85 MB
podman (master) $ ./bin/podman-remote untag docker.io/library/alpine:latest
Error: Failed to find image docker.io/library/alpine:latest: unable to find 'docker.io/library/alpine:latest' in local storage: no such image
podman (master) $ ./bin/podman-remote untag docker.io/library/alpine:latest
Error: Failed to find image docker.io/library/alpine:latest: unable to find 'docker.io/library/alpine:latest' in local storage: no such image
```
... and probably more

vrothberg:
Untagging all is still broken as the endpoint forces a repo to be specified. I'll wrap up something for that and link it here.

vrothberg:
It bit me in the tests in https://github.com/containers/podman/pull/7823

vrothberg:
https://github.com/containers/podman/pull/7840

rhatdan:
Replaced with https://github.com/containers/podman/pull/7840

PR #7717, opened 2020-09-21
Fix up attach tests for podman remote

rhatdan:
When we execute podman-remote attach, we were not checking if the
container was in the correct state, this is leading to timeouts and
we had turned off remote testing.

Also added an IfRemote() function so we can turn on more tests when
using the "-l" flag for local, but use container name for remote.


TomSweeneyRedHat:
LGTM

rhatdan:
@vrothberg  @baude @mheon @ashley-cui @QiWang19 @jwhonce @giuseppe PTAL

vrothberg (commented):
LGTM

PR #7718, opened 2020-09-21
Sign multi-arch images

QiWang19:
podman image sign handles muti-arch images.
--all option to create signature for each manifest from the image manifest list.

close:  https://bugzilla.redhat.com/show_bug.cgi?id=1770037

### DIFF
--- a/cmd/podman/images/sign.go
+++ b/cmd/podman/images/sign.go
@@ -47,6 +47,7 @@ func init() {
 	certDirFlagName := "cert-dir"
 	flags.StringVar(&signOptions.CertDir, certDirFlagName, "", "`Pathname` of a directory containing TLS certificates and keys")
 	_ = signCommand.RegisterFlagCompletionFunc(certDirFlagName, completion.AutocompleteDefault)
+	flags.BoolVarP(&signOptions.All, "all", "a", false, "Sign all the manifests of the multi-architecture image")
 }
 
 func sign(cmd *cobra.Command, args []string) error {
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -28,6 +28,8 @@ import (
 	"github.com/containers/podman/v2/pkg/rootless"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/storage"
+	dockerRef "github.com/docker/distribution/reference"
+	"github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -718,9 +720,9 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 					logrus.Errorf("unable to close %s image source %q", srcRef.DockerReference().Name(), err)
 				}
 			}()
-			getManifest, _, err := rawSource.GetManifest(ctx, nil)
+			topManifestBlob, manifestType, err := rawSource.GetManifest(ctx, nil)
 			if err != nil {
-				return errors.Wrapf(err, "error getting getManifest")
+				return errors.Wrapf(err, "error getting manifest blob")
 			}
 			dockerReference := rawSource.Reference().DockerReference()
 			if dockerReference == nil {
@@ -743,34 +745,34 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 					return err
 				}
 			}
-			manifestDigest, err := manifest.Digest(getManifest)
+			manifestDigest, err := manifest.Digest(topManifestBlob)
 			if err != nil {
 				return err
 			}
 
-			// create signature
-			newSig, err := signature.SignDockerManifest(getManifest, dockerReference.String(), mech, options.SignBy)
-			if err != nil {
-				return errors.Wrapf(err, "error creating new signature")
-			}
-			// create the signstore file
-			signatureDir := fmt.Sprintf("%s@%s=%s", sigStoreDir, manifestDigest.Algorithm(), manifestDigest.Hex())
-			if err := os.MkdirAll(signatureDir, 0751); err != nil {
-				// The directory is allowed to exist
-				if !os.IsExist(err) {
-					logrus.Error(err)
-					return nil
+			if options.All {
+				if !manifest.MIMETypeIsMultiImage(manifestType) {
+					return errors.Errorf("%s is not a multi-architecture image (manifest type %s)", signimage, manifestType)
 				}
-			}
-			sigFilename, err := getSigFilename(signatureDir)
-			if err != nil {
-				logrus.Errorf("error creating sigstore file: %v", err)
+				list, err := manifest.ListFromBlob(topManifestBlob, manifestType)
+				if err != nil {
+					return errors.Wrapf(err, "Error parsing manifest list %q", string(topManifestBlob))
+				}
+				instanceDigests := list.Instances()
+				for _, instanceDigest := range instanceDigests {
+					digest := instanceDigest
+					man, _, err := rawSource.GetManifest(ctx, &digest)
+					if err != nil {
+						return err
+					}
+					if err = putSignature(man, mech, sigStoreDir, instanceDigest, dockerReference, options); err != nil {
+						return errors.Wrapf(err, "error storing signature for %s, %v", dockerReference.String(), instanceDigest)
+					}
+				}
 				return nil
 			}
-			err = ioutil.WriteFile(filepath.Join(signatureDir, sigFilename), newSig, 0644)
-			if err != nil {
-				logrus.Errorf("error storing signature for %s", rawSource.Reference().DockerReference().String())
-				return nil
+			if err = putSignature(topManifestBlob, mech, sigStoreDir, manifestDigest, dockerReference, options); err != nil {
+				return errors.Wrapf(err, "error storing signature for %s, %v", dockerReference.String(), manifestDigest)
 			}
 			return nil
 		}()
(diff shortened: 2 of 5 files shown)


rhatdan:
@QiWang19 Needs a rebase.

rhatdan:
@QiWang19 Needs a rebae.

QiWang19:
I'll rework this one after #7357 being merged.

> @QiWang19 Needs a rebae.

QiWang19:
@mtrmac PTAL.

cmd/podman/images/sign.go
```diff
@@ -47,6 +47,7 @@ func init() {
 	certDirFlagName := "cert-dir"
 	flags.StringVar(&signOptions.CertDir, certDirFlagName, "", "`Pathname` of a directory containing TLS certificates and keys")
 	_ = signCommand.RegisterFlagCompletionFunc(certDirFlagName, completion.AutocompleteDefault)
+	flags.BoolVarP(&signOptions.All, "all", "a", false, "Allow signing all the manifest of the image manifest list")
```
mtrmac, line 50: “Allow” seems unnecessary.

Maybe say “multi-architecture image” instead of the implementation jargon.

(Similarly in the man page.)

pkg/domain/infra/abi/images.go
```diff
@@ -718,7 +719,7 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 					logrus.Errorf("unable to close %s image source %q", srcRef.DockerReference().Name(), err)
 				}
 			}()
-			getManifest, _, err := rawSource.GetManifest(ctx, nil)
+			getManifest, manifestType, err := rawSource.GetManifest(ctx, nil)
```
mtrmac, line 722: (Unrelated: `s/getManifest/manifestBlob/` , or maybe  …`/topManifest[Blob]/`, would be more readable.)

pkg/domain/infra/abi/images.go
```diff
@@ -748,28 +749,38 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
-			}
-			// create the signstore file
-			signatureDir := fmt.Sprintf("%s@%s=%s", sigStoreDir, manifestDigest.Algorithm(), manifestDigest.Hex())
-			if err := os.MkdirAll(signatureDir, 0751); err != nil {
-				// The directory is allowed to exist
-				if !os.IsExist(err) {
-					logrus.Error(err)
-					return nil
+			if options.All {
+				list, err := manifest.ListFromBlob(getManifest, manifestType)
```
mtrmac, line 753: Probably check `manifest.MIMETypeIsMultiImage` first to give the user a meaningful error.

pkg/domain/infra/abi/images.go
```diff
@@ -806,3 +817,32 @@ func localPathFromURI(url *url.URL) (string, error) {
...
+
+func createSignature(m []byte, dockerReference string, mech signature.SigningMechanism, keyIdentity string) ([]byte, error) {
+	newSig, err := signature.SignDockerManifest(m, dockerReference, mech, keyIdentity)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating new signature")
+	}
+	return newSig, nil
+}
+
+func putSignature(sigStoreDir string, signature []byte, instanceDigest digest.Digest) error {
```
mtrmac, line 829: (More of a reminder of an option than a review comment: `createSignature`+`putSignature` can be a single function instead of two. Entirely up to you, this works just fine.)

pkg/domain/infra/abi/images.go
```diff
@@ -806,3 +817,32 @@ func localPathFromURI(url *url.URL) (string, error) {
 	}
 	return url.Path, nil
 }
+
+func createSignature(m []byte, dockerReference string, mech signature.SigningMechanism, keyIdentity string) ([]byte, error) {
```
mtrmac, line 821: Please use the original `reference.Named` instead of a `string` for `dockerReference`, to keep strong typing as long as possible.

(Absolutely non-blocking: Consider also passing all of `options` instead of `keyIdentity`. Either way works fine.)

pkg/domain/infra/abi/images.go
```diff
@@ -748,28 +749,38 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
+			// create signature
+			newSig, err := createSignature(getManifest, dockerReference.String(), mech, options.SignBy)
 			if err != nil {
-				logrus.Errorf("error storing signature for %s", rawSource.Reference().DockerReference().String())
+				return err
+			}
+			// create the signstore file
+			if err = putSignature(sigStoreDir, newSig, manifestDigest); err != nil {
+				logrus.Errorf("error storing signature for %s, %v", dockerReference.String(), err)
 				return nil
```
mtrmac, line 784: (Absolutely non-blocking: This is pre-existing and more or less unrelated, still: I don’t understand why some errors are fatal, and returned, and some are  only logged and otherwise ignored.)

pkg/domain/infra/abi/images.go
```diff
@@ -806,3 +817,32 @@ func localPathFromURI(url *url.URL) (string, error) {
...
+	return newSig, nil
+}
+
+func putSignature(sigStoreDir string, signature []byte, instanceDigest digest.Digest) error {
+	signatureDir := fmt.Sprintf("%s@%s=%s", sigStoreDir, instanceDigest.Algorithm(), instanceDigest.Hex())
+	if err := os.MkdirAll(signatureDir, 0751); err != nil {
+		// The directory is allowed to exist
+		if !os.IsExist(err) {
+			logrus.Errorf("error creating directory %s: %s", signatureDir, err)
+			return err
```
mtrmac, line 835: The caller is already calling `logrus.Errorf` on this error. Generally I would expect such a decision of whether to return or log the error to be maintained at a single level of the call stack.

(Same for the other errors.)

pkg/domain/infra/abi/images.go
mtrmac, line 829: (Absolutely non-blocking: It might feel a tiny bit more readable to have the “destination” parameters (directory+digest) together, instead of wedging the “content” `signature` parameter in the middle.)

mtrmac (commented):
ACK, just some nits.

pkg/domain/infra/abi/images.go
QiWang19: If some errors occur before `putSignature`,  return the error since the `putSignature` is not useful in that situation. `putSignature` just logrus the error so it can continue to next image inside the for loop, but this may lead to failing silently.

mtrmac: From the point of the user, either the command succeeded in doing everything required, or not. If it doesn’t do everything, the user must notice and just _somehow_; if the command reports that failure in two different ways, I can’t see how that is mores useful than doing it in one way.

And just logging the error makes automation much more difficult — especially if users don’t encounter that case and don’t even _know_ they should be checking for logs.

pkg/domain/infra/abi/images.go
mtrmac: Oops, this is not possible if the error reporting regimes are different.

cmd/podman/images/sign.go
```diff
@@ -47,6 +47,7 @@ func init() {
 	certDirFlagName := "cert-dir"
 	flags.StringVar(&signOptions.CertDir, certDirFlagName, "", "`Pathname` of a directory containing TLS certificates and keys")
 	_ = signCommand.RegisterFlagCompletionFunc(certDirFlagName, completion.AutocompleteDefault)
+	flags.BoolVarP(&signOptions.All, "all", "a", false, "Sign all the manifests of the multi-architecture image manifest list")
```
mtrmac, line 50: Just “of the multi-architecture image”? (In both places)

pkg/domain/infra/abi/images.go
```diff
@@ -743,34 +744,36 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
-			// create the signstore file
-			signatureDir := fmt.Sprintf("%s@%s=%s", sigStoreDir, manifestDigest.Algorithm(), manifestDigest.Hex())
-			if err := os.MkdirAll(signatureDir, 0751); err != nil {
-				// The directory is allowed to exist
-				if !os.IsExist(err) {
-					logrus.Error(err)
-					return nil
+			if options.All {
+				if !manifest.MIMETypeIsMultiImage(manifestType) {
+					return errors.Errorf("manifest type %s, %s is not a multi-architecture image", manifestType, signimage)
```
mtrmac, line 754: “%s is not a multi-architecture image (manifest type %s)” would prioritize the more relevant information.

pkg/domain/infra/abi/images.go
```diff
@@ -743,34 +744,36 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
+				}
+				list, err := manifest.ListFromBlob(topManifestBlob, manifestType)
+				if err != nil {
+					return errors.Wrapf(err, "Error parsing manifest list %q", string(topManifestBlob))
```
mtrmac, line 758: (I’m not sure that dumping the full manifest here is warranted, but it may well be useful.)

pkg/domain/infra/abi/images.go
```diff
@@ -743,34 +744,36 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
+				}
+				instanceDigests := list.Instances()
+				for _, instanceDigest := range instanceDigests {
+					digest := instanceDigest
+					man, _, err := rawSource.GetManifest(ctx, &digest)
+					if err != nil {
+						return err
+					}
+					if err = putSignature(man, mech, sigStoreDir, instanceDigest, dockerReference, options); err != nil {
+						return errors.Wrapf(err, "error storing signature for %s, %v", dockerReference.String())
```
mtrmac, line 768: (Validation detects an error)

The context should include `instanceDigest`.

pkg/domain/infra/abi/images.go
```diff
@@ -743,34 +744,36 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
-			if err != nil {
-				logrus.Errorf("error creating sigstore file: %v", err)
 				return nil
 			}
-			err = ioutil.WriteFile(filepath.Join(signatureDir, sigFilename), newSig, 0644)
-			if err != nil {
-				logrus.Errorf("error storing signature for %s", rawSource.Reference().DockerReference().String())
-				return nil
+
+			// create signature and save to the signstore file
```
mtrmac, line 774: (Absolutely non-blocking: This comment is now a bit surprising — why is this one worth a comment and the other `putSignature` call not? It almost suggests that there is a difference.

Maybe write a documentation string for the `putSignature` function instead.)

pkg/domain/infra/abi/images.go
mtrmac: I’m sorry, it should include _both_ `dockerReference` and `instanceDigest`.

pkg/domain/infra/abi/images.go
```diff
@@ -743,34 +744,34 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
...
-			if err != nil {
-				logrus.Errorf("error creating sigstore file: %v", err)
 				return nil
 			}
-			err = ioutil.WriteFile(filepath.Join(signatureDir, sigFilename), newSig, 0644)
-			if err != nil {
-				logrus.Errorf("error storing signature for %s", rawSource.Reference().DockerReference().String())
-				return nil
+			if err = putSignature(topManifestBlob, mech, sigStoreDir, manifestDigest, dockerReference, options); err != nil {
+				return errors.Wrapf(err, "error storing signature for %v", manifestDigest)
```
mtrmac, line 774: Same here, include the `dockerReference` (it directly relates to user input, and disambiguates if there are multiple images), and _maybe_ the manifest digest.

mtrmac (commented):
LGTM.

TomSweeneyRedHat:
LGTM
Very nice work @QiWang19 !

zhangguanzhang:
need approve label 😀

PR #7719, opened 2020-09-22
podman save: fix redirect of multi-images

vrothberg:
Fix a bug when saving multi-image archives where the progress bars from
copying images was accidentally written to the archive and hence
corrupted it.

### DIFF
--- a/test/system/120-load.bats
+++ b/test/system/120-load.bats
@@ -147,4 +147,45 @@ verify_iid_and_name() {
        "Diagnostic from 'podman load' without redirection or -i"
 }
 
+@test "podman load - multi-image archive" {
+    img1="quay.io/libpod/testimage:00000000"
+    img2="quay.io/libpod/testimage:20200902"
+    archive=$PODMAN_TMPDIR/myimage-$(random_string 8).tar
+
+    run_podman pull $img1
+    run_podman pull $img2
+
+    run_podman save -m -o $archive $img1 $img2
+    run_podman rmi -f $img1 $img2
+    run_podman load -i $archive
+
+    run_podman image exists $img1
+    run_podman image exists $img2
+    run_podman rmi -f $img1 $img2
+}
+
+@test "podman load - multi-image archive with redirect" {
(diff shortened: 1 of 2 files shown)


vrothberg:
@edsantiago please have a look at the system tests.

rhatdan:
LGTM

test/system/120-load.bats
```diff
@@ -147,4 +147,43 @@ verify_iid_and_name() {
        "Diagnostic from 'podman load' without redirection or -i"
 }
 
+@test "podman load - multi-image archive" {
+    alpine="docker.io/library/alpine:latest"
+    busybox="docker.io/library/busybox:latest"
```
edsantiago, line 152: We no longer use docker.io - we can't. We use `quay.io/libpod/testimage:<some tag>`. I believe you could accomplish this with
```bash
    img1="quay.io/libpod/testimage:00000000"
    img2="quay.io/libpod/testimage:20200902"
```
Both have been used by prior system tests, hence need to exist forever. If you actually need images that differ in name as well as tag, you could use `testdigest_v2s2:20200210` or `:latest`. That is a tiny image that should be quick and safe to pull.

Or, since one of my top goals in system tests is to minimize pulls (because they're a cause of flakes), you could instead `podman build` with custom tags.

vrothberg: Thanks, Ed! I went with pulling.

test/system/120-load.bats
```diff
@@ -147,4 +147,43 @@ verify_iid_and_name() {
...
+
+    run_podman pull $img1
+    run_podman pull $img2
+
+    run_podman save -m -o $archive $img1 $img2
+    run_podman rmi -af
+    run_podman load -i $archive
+
+    run_podman image exists $img1
+    run_podman image exists $img2
```
edsantiago, line 163: Please `rmi` anything you bring in. Sorry, I should've caught that earlier.

vrothberg: Thanks, fixed now :)

mheon:
LGTM

PR #7721, opened 2020-09-22
Add option to not switch cgroup when user doesn't own it

zv0n:
Hello,
I've added a command line option which makes it possible to not switch cgroup when user doesn't own it.

We need this change in our environment because we're using podman for PBS jobs and if podman changes cgroup to user.slice of user who started the job rather than keep the PBS' cgroup, we can't properly monitor how much resources the job is using.

I hope this change looks alright to you and can be merged

### DIFF
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -28,6 +28,10 @@ The CGroup manager to use for container cgroups. Supported values are cgroupfs o
 Note: Setting this flag can cause certain commands to break when called on containers previously created by the other CGroup manager type.
 Note: CGroup manager is not supported in rootless mode when using CGroups Version V1.
 
+**--cgroup-check-user**=*boolean*
+
+Whether or not to check if user owns current CGroup. If set to true and user doesn't own current CGroup, try switching to a CGroup the user owns. (Default: `true`)
+
 **--cni-config-dir**
 Path of the configuration directory for CNI networks.  (Default: `/etc/cni/net.d`)
 
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -188,7 +188,7 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 	// 3) command doesn't require Parent Namespace
 	_, found := cmd.Annotations[registry.ParentNSRequired]
 	if !registry.IsRemote() && rootless.IsRootless() && !found {
-		err := registry.ContainerEngine().SetupRootless(registry.Context(), cmd)
+		err := registry.ContainerEngine().SetupRootless(registry.Context(), cmd, cfg.CgroupCheck)
 		if err != nil {
 			return err
 		}
@@ -256,6 +256,7 @@ func rootFlags(cmd *cobra.Command, opts *entities.PodmanConfig) {
 
 	pFlags := cmd.PersistentFlags()
 	pFlags.StringVar(&cfg.Engine.CgroupManager, "cgroup-manager", cfg.Engine.CgroupManager, "Cgroup manager to use (\"cgroupfs\"|\"systemd\")")
+	pFlags.BoolVar(&opts.CgroupCheck, "cgroup-check-user", true, "If user doesn't own current cgroup, change cgroup to user.slice (if exists)")
 	pFlags.StringVar(&opts.CPUProfile, "cpu-profile", "", "Path for the cpu profiling results")
 	pFlags.StringVar(&opts.ConmonPath, "conmon", "", "Path of the conmon binary")
 	pFlags.StringVar(&cfg.Engine.NetworkCmdPath, "network-cmd-path", cfg.Engine.NetworkCmdPath, "Path to the command for configuring the network")
--- a/pkg/domain/entities/engine.go
+++ b/pkg/domain/entities/engine.go
@@ -42,6 +42,7 @@ type PodmanConfig struct {
 	CPUProfile     string           // Hidden: Should CPU profile be taken
 	EngineMode     EngineMode       // ABI or Tunneling mode
 	Identity       string           // ssh identity for connecting to server
+	CgroupCheck    bool             // whether to check ownership of cgroup
 	MaxWorks       int              // maximum number of parallel threads
 	RegistriesConf string           // allows for specifying a custom registries.conf
 	Remote         bool             // Connection to Podman API Service will use RESTful API
--- a/pkg/domain/entities/engine_container.go
+++ b/pkg/domain/entities/engine_container.go
@@ -69,7 +69,7 @@ type ContainerEngine interface {
 	PodStop(ctx context.Context, namesOrIds []string, options PodStopOptions) ([]*PodStopReport, error)
 	PodTop(ctx context.Context, options PodTopOptions) (*StringSliceReport, error)
 	PodUnpause(ctx context.Context, namesOrIds []string, options PodunpauseOptions) ([]*PodUnpauseReport, error)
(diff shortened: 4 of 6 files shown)


rhatdan:
@giuseppe PTAL

rhatdan:
Not sure what PBS is?

zv0n:
@rhatdan Portable Batch System
basically if user wants to compute something that will take a long time/lot of resources, they submit their script to PBS which then decides when/on which server to run the script (based on current load on the servers). Then when the job is finished, output of the script is save in the user's home directory and all temporary files are cleaned up.

The problem is that when podman changes CGroup to user's user.slice PBS can't keep track of the processes spawned from within podman, so it can't properly monitor resource use or clean up after itself

rhatdan:
@giuseppe is on PTO right now, but I thought we had a mechanism to tell the container to not use cgroups, which seems like what you want here?

docs/source/markdown/podman.1.md
```diff
@@ -28,6 +28,10 @@ The CGroup manager to use for container cgroups. Supported values are cgroupfs o
 Note: Setting this flag can cause certain commands to break when called on containers previously created by the other CGroup manager type.
 Note: CGroup manager is not supported in rootless mode when using CGroups Version V1.
 
+**--cgroup-check-user**=*boolean*
```
mheon, line 31: I feel like this should be part of the `--cgroup` flag to `podman run` and `podman create` - we should never be swapping cgroups otherwise.

mheon:
@rhatdan Yes, `--cgroups=disabled` (though this only supports crun right now)

giuseppe:
@zv0n as @mheon pointed out, we already have `--cgroups=disabled`.  Does that work for your use case or is there anything missing?

zv0n:
I think we've tried that with 1.9 and it didn't quite do the trick. I don't have access to our environment right now, but I'll test it out tomorrow and let you know what does/doesn't work

rhatdan:
If it does not work, lets work on it to see if we can fix whatever is broken to make it work.

zv0n:
Hi, I've tested `--cgroups disabled` and it doesn't work. I suspect it's because the portion of SetupRootles, which changes CGroup, only checks whether CGroupManager is systemd and even if CGroupManager ISN'T systemd, if the current system has systemd as init, the function changes CGroup.

Side note - I was also informed that some of our users want shared pid namespace (`--pid host`) for easy communication between containers, which isn't possible with `--cgroups disabled`. I realize this is a very very edge case usage, so I understand if you don't want an additional command line parameter for that, but if it'd be possible to add this option (have cgroups enabled, but skip cgroup switch to user.slice) to config file for example, we'd really appreciate it :)

giuseppe:
> Hi, I've tested `--cgroups disabled` and it doesn't work. I suspect it's because the portion of SetupRootles, which changes CGroup, only checks whether CGroupManager is systemd and even if CGroupManager ISN'T systemd, if the current system has systemd as init, the function changes CGroup.

I think we should skip the check when `--cgroups=disabled`.




> Side note - I was also informed that some of our users want shared pid namespace (`--pid host`) for easy communication between containers, which isn't possible with `--cgroups disabled`. I realize this is a very very edge case usage, so I understand if you don't want an additional command line parameter for that, but if it'd be possible to add this option (have cgroups enabled, but skip cgroup switch to user.slice) to config file for example, we'd really appreciate it :)

How would we be able to terminate the container without cgroups and pid namespace?  The idea with `--pid=host` is that we can get the list of processes from the cgroup.  If we have no cgroup and no pid namespace, it won't be possible to terminate a container, as processes can escape.

zv0n:
I see, well I guess we'll just tell our users that they can't have everything they want :D

> I think we should skip the check when `--cgroups=disabled`.

Should I create an issue then? I think changes in this PR are irrelevant at this point and I don't know from which variable to get cgroups setting

rhatdan:
@zv0n Can you modify your patch to just not do the check if --cgroups=disabled  We should get that partial fix in for now.

zv0n:
@rhatdan I've tried going through the code and the only way I see that this could be done is if the cgroups option was moved to rootflags. Right now it is in command flags and I don't see a way to access those from root.go. I've tried moving the cgroups option to rootFlags, but don't know from where to get correct default value, could you please point me in the right direction? Also would moving cgroups to rootFlags be acceptable to you?

rhatdan:
@giuseppe WDYT?

mheon:
Moving `--cgroups` to a root flag does not seem feasible to me.

Is there any way to rework the logic in `setupRootless()` to not require a flag?

giuseppe:
Maybe `--cgroup-manager`?

zv0n:
@giuseppe so something like `--cgroup-manager disabled`, which would also automatically set `--cgroups disabled`?

giuseppe:
> @giuseppe so something like `--cgroup-manager disabled`, which would also automatically set `--cgroups disabled`?

yes, I think that could work.  @mheon what do you think?

zv0n:
After trying to implement this, I think it's not possible to change `--cgroups` based on `--cgroups-manager` because when `--cgroups` is being parsed from command line the function creates a new separate configuration object from default values. (`cmd/podman/common/create.go:14` calls `PodmanConfig()` which in turn calls `newPodmanConfig()`)

zv0n:
So, anyone here have any idea how to achieve this? Because so far, short of completely rewriting the CLI parsing system, I don't see a way :D

rhatdan:
@mheon @baude PTAL

mheon:
I don't actually see what you're saying with being unable to use `--cgroup-manager` to set `--cgroups`. I don't follow how we're creating a new config object - it's called in a `sync.Once` to ensure we only ever ensure one config?

zv0n:
@mheon I might be wrong, I haven't programmed in Go much, but to me it seems that when `flags.AddFlagSet(common.GetCreateFlags(&cliVals)` is called in `cmd/podman/containers/run.go` a new config `containerConfig` is created in `cmd/podman/common/crate.go`.

This new config is then used on lines 58-62 in `create.go` to provide default value for `cf.CGroupsMode`.

This is a problem as I'd like to edit `containerConfig` based on value set to `cfg.Engine.CgroupManager` in `cmd/podman/root.go`, so `containerConfig.Cgroups()` would return `disabled` in appropriate situation. But since the `containerConfig` var is created in `create.go` I can't change anything in it from `root.go` (at least I think, again, I'm not a Go Pro (heh :D)).

I also can't change the value based on `cfg.Engine.CgroupManager` without extensive rewrites. As the code exists now, it only expects `cgroupfs` or `systemd`. So my idea for this PR was:

If `--cgroupsManager` is set to `disabled`, change `containerConfig` so `containerConfig.Cgroups()` would return `disabled`, then change `cfg.Engine.CgroupManager` to its default value `cgroupfs` so as to not break existing code.

But as I have said, I'm running into issues with changing `containerConfig`.

If I'm doing something wrong/stupidly, please tell me, I'd like to learn, thanks!

mheon:
I strongly suspect you're looking in the wrong place - there is no Engine field in `entities.PodmanConfig`, which is what we care about. I think you're looking at the containers/common config, which is loaded in from configuration file. That is then overridden by the `entities.PodmanConfig` settings in `cmd/podman/registry` that are sourced from CLI flags.

zv0n:
So, should I create an issue for this? This looks like it would be a substantial change to the code base and I don't feel confident enough in my Go/Cgroup knowledge to do this myself :D

mheon:
Sure.

Issue #7726, opened 2020-09-22
Podman manifest inspect doesn't work with existing images


fatherlinux:
/kind bug

**Description**

Podman manifest inspect only works with manifests created with podman, not with manifests that are created with docker and other tools. This can be tested by trying any image on a registry server.

**Steps to reproduce the issue:**

Any of these three commands will fail:

podman manifest inspect docker.io/library/busybox

podman manifest inspect registry.access.redhat.com/ubi8/ubi-minimal

podman manifest inspect registry.access.redhat.com/ubi8/ubi

**Describe the results you received:**

All three of the above command will fail with the following error:

Error: error inspect manifest registry.access.redhat.com/ubi8/ubi: loading manifest "registry.access.redhat.com/ubi8/ubi": unable to load manifest list: unsupported format "application/vnd.docker.distribution.manifest.v2+json": manifest type not supported

**Describe the results you expected:**
I would expect the "podman manifest inspect" command to work with pre-existing images, just like it works with manifests created by podman. Here's an example of this command functioning with a manifest created by podman:

[root@fedora ~]# podman manifest create mylist:v1.11
74ad8d03f4ed4802b3f9f3a715e5c75c17b574dc5c65370667de4ba0b89d78e1
[root@fedora ~]# podman manifest inspect mylist:v1.11
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": null
}

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.6
API Version:  1
Go Version:   go1.14.6
Built:        Tue Sep  1 15:26:51 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.19, commit: 5dce9767526ed27f177a8fa3f281889ad509fea7'
  cpus: 1
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: fedora.dc2.crunchtools.com
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.6.6-300.fc32.x86_64
  linkmode: dynamic
  memFree: 892121088
  memTotal: 2075738112
  ociRuntime:
    name: crun
[...]

PR #7735, opened 2020-09-22
fix allowing inspect manifest of non-local image

QiWang19:
Add support of `podman manifest inspect` returning manifest list of non-local manifest.
Close https://github.com/containers/podman/issues/7726

### DIFF
--- a/pkg/api/handlers/libpod/manifests.go
+++ b/pkg/api/handlers/libpod/manifests.go
@@ -6,11 +6,13 @@ import (
 
 	"github.com/containers/buildah/manifests"
 	copy2 "github.com/containers/image/v5/copy"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/podman/v2/libpod"
 	"github.com/containers/podman/v2/libpod/image"
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/api/handlers/utils"
+	"github.com/containers/podman/v2/pkg/domain/infra/abi"
 	"github.com/gorilla/schema"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
@@ -48,17 +50,18 @@ func ManifestCreate(w http.ResponseWriter, r *http.Request) {
 func ManifestInspect(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 	name := utils.GetName(r)
-	newImage, err := runtime.ImageRuntime().NewFromLocal(name)
-	if err != nil {
-		utils.ImageNotFound(w, name, err)
+	imageEngine := abi.ImageEngine{Libpod: runtime}
+	inspectReport, inspectError := imageEngine.ManifestInspect(r.Context(), name)
+	if inspectError != nil {
+		utils.Error(w, "Something went wrong.", http.StatusNotFound, inspectError)
 		return
 	}
-	data, err := newImage.InspectManifest()
-	if err != nil {
-		utils.InternalServerError(w, err)
+	var list manifest.Schema2List
+	if err := json.Unmarshal(inspectReport, &list); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "Unmarshal()"))
 		return
 	}
-	utils.WriteResponse(w, http.StatusOK, data)
+	utils.WriteResponse(w, http.StatusOK, &list)
 }
 
 func ManifestAdd(w http.ResponseWriter, r *http.Request) {
--- a/pkg/domain/infra/abi/manifest.go
+++ b/pkg/domain/infra/abi/manifest.go
@@ -3,6 +3,7 @@
 package abi
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -11,15 +12,17 @@ import (
 	"strings"
 
 	"github.com/containers/buildah/manifests"
+	buildahManifests "github.com/containers/buildah/pkg/manifests"
+	"github.com/containers/buildah/util"
 	buildahUtil "github.com/containers/buildah/util"
 	cp "github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
 	libpodImage "github.com/containers/podman/v2/libpod/image"
 	"github.com/containers/podman/v2/pkg/domain/entities"
-	"github.com/containers/podman/v2/pkg/util"
 	"github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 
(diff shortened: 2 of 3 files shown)


test/e2e/manifest_test.go
```diff
@@ -49,6 +49,12 @@ var _ = Describe("Podman manifest", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman manifest inspect", func() {
+		session := podmanTest.Podman([]string{"manifest", "inspect", "docker://library/busybox"})
```
TomSweeneyRedHat, line 53: Just to be completely sure, could you also try pulling the image specified as it was in the issue.

`podman manifest inspect docker.io/library/busybox`

test/e2e/manifest_test.go
```diff
@@ -49,6 +49,12 @@ var _ = Describe("Podman manifest", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman manifest inspect", func() {
+		session := podmanTest.PodmanNoCache([]string{"manifest", "inspect", "docker://library/busybox"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
TomSweeneyRedHat, line 55:
```suggestion
		Expect(session.ExitCode()).To(Equal(0))
		
		session := podmanTest.PodmanNoCache([]string{"manifest", "inspect", "docker.io/library/busybox"})
		session.WaitWithDefaultTimeout()
		Expect(session.ExitCode()).To(Equal(0))
```

test/e2e/manifest_test.go
```diff
@@ -49,6 +49,12 @@ var _ = Describe("Podman manifest", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman manifest inspect", func() {
+		session := podmanTest.PodmanNoCache([]string{"manifest", "inspect", "docker://library/busybox"})
```
mheon, line 53: Is this an image we have *cached*? The bug here is that we are failing to run `podman manifest list` on local images we have already pulled, but who manifests were not created by Podman?

QiWang19: > The bug here is that we are failing to run `podman manifest list` on local images we have already pulled, but who manifests were not created by Podman?

yes, this needs to be fixed. And the steps to reproduce the issue is trying to inspect the manifest on a registry server. So manifest inspect is supposed to work no matter the image is in local storage or not.

rhatdan:
@nalind @mtrmac @vrothberg PTAL

pkg/domain/infra/abi/manifest.go
```diff
@@ -49,20 +52,64 @@ func (ir *ImageEngine) ManifestInspect(ctx context.Context, name string) ([]byte
...
-	image, err := ir.Libpod.ImageRuntime().New(ctx, name, "", "", nil, nil, libpodImage.SigningOptions{}, nil, util.PullImageMissing)
+	if newImage, err := ir.Libpod.ImageRuntime().NewFromLocal(name); err == nil {
+		// return the manifest in local storage
+		if list, err := newImage.InspectManifest(); err == nil {
+			buf, err := json.MarshalIndent(list, "", "    ")
+			if err != nil {
+				return buf, errors.Wrapf(err, "error rendering manifest for display")
+			}
+			return buf, nil
+		} else if errors.Cause(err) != buildahManifests.ErrManifestTypeNotSupported {
```
vrothberg, line 63: Why are continuing if the type isn't supported? A comment would help.

QiWang19:

pkg/api/handlers/libpod/manifests.go
```diff
@@ -48,17 +50,18 @@ func ManifestCreate(w http.ResponseWriter, r *http.Request) {
...
+	inspectReport, inspectError := imageEngine.ManifestInspect(r.Context(), name)
+	if inspectError != nil {
+		utils.Error(w, "Something went wrong.", http.StatusNotFound, inspectError)
 		return
 	}
-	data, err := newImage.InspectManifest()
-	if err != nil {
-		utils.InternalServerError(w, err)
+	var list manifest.Schema2List
+	if err := json.Unmarshal(inspectReport, &list); err != nil {
```
nalind, line 60: Does this need to handle OCI index format?  If so, does it need to convert the index to a Docker manifest list, or should it just pass through what it got back unmodified?

If the manifest's type is not already in the Docker manifest list format, and we need to return a result in that format, we'd want to use `manifest.ListFromBlob()` to produce a list object, call the list's `ConvertToMIMEType()` method to get a second list object of the `manifest.DockerV2ListMediaType` type, and then call the second list's `Serialize()` method to re-encode it as JSON.

QiWang19: Add the manifest type check in pkg/domain/infra/abi/manifest.go.

QiWang19:
@nalind @containers/podman-maintainers PTAL

pkg/domain/infra/abi/manifest.go
```diff
@@ -41,28 +44,82 @@ func (ir *ImageEngine) ManifestCreate(ctx context.Context, names, images []strin
...
 // ManifestInspect returns the content of a manifest list or image
 func (ir *ImageEngine) ManifestInspect(ctx context.Context, name string) ([]byte, error) {
-	dockerPrefix := fmt.Sprintf("%s://", docker.Transport.Name())
-	_, err := alltransports.ParseImageName(name)
+	if newImage, err := ir.Libpod.ImageRuntime().NewFromLocal(name); err == nil {
+		// return the manifest in local storage
+		if list, err := newImage.InspectManifest(); err == nil {
+			buf, err := json.MarshalIndent(list, "", "    ")
+			if err != nil {
+				return buf, errors.Wrapf(err, "error rendering manifest for display")
```
jwhonce, line 52: Please add `name` to error. Much needed context in multi-tenant log.

pkg/domain/infra/abi/manifest.go
```diff
@@ -41,28 +44,82 @@ func (ir *ImageEngine) ManifestCreate(ctx context.Context, names, images []strin
...
 
-	list, err := image.InspectManifest()
-	if err != nil {
-		return nil, errors.Wrapf(err, "loading manifest %q", name)
 	}
-	buf, err := json.MarshalIndent(list, "", "    ")
+	err = json.Indent(&b, result, "", "    ")
 	if err != nil {
-		return buf, errors.Wrapf(err, "error rendering manifest for display")
+		return nil, errors.Wrapf(err, "error rendering manifest for display")
```
jwhonce, line 120: Adding context to the error will help debugging in multi-tenant logs

test/e2e/manifest_test.go
```diff
@@ -49,6 +49,16 @@ var _ = Describe("Podman manifest", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman manifest inspect", func() {
+		session := podmanTest.Podman([]string{"manifest", "inspect", BB})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.PodmanNoCache([]string{"manifest", "inspect", "docker.io/library/busybox"})
```
jwhonce, line 57:
```suggestion
		session = podmanTest.PodmanNoCache([]string{"manifest", "inspect", "quay.io/library/busybox"})
```
Is this possible?

test/e2e/manifest_test.go
```diff
@@ -49,6 +49,16 @@ var _ = Describe("Podman manifest", func() {
...
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
jwhonce, line 59:
```suggestion
		Expect(session).To(Exit(0))
```
Is more BDD

test/e2e/manifest_test.go
QiWang19: It's possible to inspect the manifest from quay.io, but I guess there's no library/busybox repository, do you have other suggestions?

QiWang19:

test/e2e/manifest_test.go
jwhonce: @QiWang19 I did a grep of the tests directory and found other tests using `quay.io/libpod/alpine_labels:latest` does that meet your needs?

QiWang19: `quay.io/libpod/alpine_labels:latest` is not a manifest list type

PR #7754, opened 2020-09-23
Fix a bug where log-driver json-file was made no logs

mheon:
When we added the None log driver, it was accidentally added in the middle of a set of Fallthrough stanzas which all should have
led to k8s-file, so that JSON file logging accidentally caused no logging to be selected instead of k8s-file.

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -1330,10 +1330,10 @@ func (r *ConmonOCIRuntime) sharedConmonArgs(ctr *Container, cuuid, bundlePath, p
 	switch logDriver {
 	case define.JournaldLogging:
 		logDriverArg = define.JournaldLogging
-	case define.JSONLogging:
-		fallthrough
 	case define.NoLogging:
 		logDriverArg = define.NoLogging
+	case define.JSONLogging:
+		fallthrough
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@edsantiago @TomSweeneyRedHat @baude @QiWang19 @ashley-cui @jwhonce PTAL

edsantiago:
Why don't we have tests for this? Okay, I'll take the heat for that one, but can we hold off on merging this until I look into it and write one?

QiWang19:
LGTM

edsantiago:
Never mind. I haven't the slightest idea how to test this.

mheon:
@edsantiago Yeah, I really have no idea about testing this one. I suppose we could try and stat the log path listed in `podman inspect` but there are issues with that (it'd just not exist in the `none` case?)

TomSweeneyRedHat:
LGTM

edsantiago:
>stat the log path listed in `podman inspect`

Thanks @mheon, that's what I was missing. I have a test now, but your PR is (mostly) green so I'll add my test in a separate  PR.

One quick nit: in podman-remote, the "not supported" message goes out on the server side only. Is there a way to make it go to the client instead?

PR #7758, opened 2020-09-23
system tests: new test for run --log-driver

edsantiago:
Tests all (current) values for --log-driver=X, and one test
for invalid value. For those drivers that write a local
file (json-file, k8s-file), test that the file exists and
contains results of the expected form (timestamp, stdout,
'F' for 'F'ull line, and the expected string output.

For json-file, confirm that podman issues a "Choosing k8s-file"
warning (only on local. On podman-remote, the warning goes only
to the server's stderr).

Written in response to #7754 in which driver=json-file was
falling through to 'none' instead of 'k8s-file'.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -61,8 +61,8 @@ echo $rand        |   0 | $rand
     is "$tests_run" "$(grep . <<<$tests | wc -l)" "Ran the full set of tests"
 }
 
-@test "podman run - globle runtime option" {
-    skip_if_remote "runtime flag is not passing over remote"
+@test "podman run - global runtime option" {
+    skip_if_remote "runtime flag is not passed over remote"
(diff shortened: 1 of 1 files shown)


edsantiago:
To bring this explicitly to everyone's attention: there is a bug somewhere in which `podman-remote run --log-driver=json-file ...` emits the warning line **on the server**, not the client.

rhatdan:
LGTM

Luap99:
@edsantiago journald driver could be tested with something like this:
```
$ podman run --log-driver journald --conmon-pidfile /tmp/conmonpid alpine echo teststring
teststring
$ journalctl --output cat  _PID=$(cat /tmp/conmonpid)
teststring

```

edsantiago:
@Luap99 thank you - that may be a TBD for a future test, but not this one: it would make it more complicated and fragile.

PR #7759, opened 2020-09-23
Allow filtering on pod label values

xordspar0:
Before this PR, filters of the form `podman pod ps --filter label=app=myapp` were not working. The results would include all pods that contained the `app` label with any value. Looking at the code, this makes sense. It appears that the second `=` and everything after it were getting truncated.

The strange part is that there was a test for this that was passing even before my change. However, when I changed the test to look exactly like my usecase, it did fail until I applied my patch. I haven't yet figured out why this is. If anyone can figure it out, let me know.

### DIFF
--- a/cmd/podman/pods/ps.go
+++ b/cmd/podman/pods/ps.go
@@ -73,7 +73,7 @@ func pods(cmd *cobra.Command, _ []string) error {
 	if cmd.Flag("filter").Changed {
 		psInput.Filters = make(map[string][]string)
 		for _, f := range inputFilters {
-			split := strings.Split(f, "=")
+			split := strings.SplitN(f, "=", 2)
 			if len(split) < 2 {
 				return errors.Errorf("filter input must be in the form of filter=value: %s is invalid", f)
 			}
--- a/test/e2e/pod_ps_test.go
+++ b/test/e2e/pod_ps_test.go
@@ -212,17 +212,17 @@ var _ = Describe("Podman ps", func() {
 		Expect(ec).To(Equal(0))
 
 		_, ec, podid2 := podmanTest.CreatePodWithLabels("", map[string]string{
-			"io.podman.test.label": "value1",
-			"io.podman.test.key":   "irrelevant-value",
+			"app":                "myapp",
+			"io.podman.test.key": "irrelevant-value",
 		})
 		Expect(ec).To(Equal(0))
 
 		_, ec, podid3 := podmanTest.CreatePodWithLabels("", map[string]string{
-			"io.podman.test.label": "value2",
+			"app": "test",
 		})
 		Expect(ec).To(Equal(0))
 
-		session := podmanTest.Podman([]string{"pod", "ps", "--no-trunc", "--filter", "label=io.podman.test.key", "--filter", "label=io.podman.test.label=value1"})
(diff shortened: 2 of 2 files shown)


rhatdan:
Thanks @xordspar0 
LGTM
/approve
@TomSweeneyRedHat @mheon @ashley-cui @QiWang19 @vrothberg @jwhonce @giuseppe @baude PTAL

cmd/podman/pods/ps.go
```diff
@@ -73,7 +73,7 @@ func pods(cmd *cobra.Command, _ []string) error {
 	if cmd.Flag("filter").Changed {
 		psInput.Filters = make(map[string][]string)
 		for _, f := range inputFilters {
-			split := strings.Split(f, "=")
+			split := strings.SplitN(f, "=", 2)
```
TomSweeneyRedHat, line 76: nice fix, been bit by this elsewhere....

TomSweeneyRedHat:
LGTM

ashley-cui (commented):
LGTM

test/e2e/pod_ps_test.go
```diff
@@ -212,17 +212,17 @@ var _ = Describe("Podman ps", func() {
...
 		Expect(ec).To(Equal(0))
 
 		_, ec, podid3 := podmanTest.CreatePodWithLabels("", map[string]string{
-			"io.podman.test.label": "value2",
+			"app": "test",
 		})
 		Expect(ec).To(Equal(0))
 
-		session := podmanTest.Podman([]string{"pod", "ps", "--no-trunc", "--filter", "label=io.podman.test.key", "--filter", "label=io.podman.test.label=value1"})
+		session := podmanTest.Podman([]string{"pod", "ps", "--no-trunc", "--filter", "label=app", "--filter", "label=app=myapp"})
```
xordspar0, line 225: I understand why this test was passing now! I didn't notice that the two filters in the old test were for different label keys. `label=...key` and `label=...label=value1`. This test was passing because there is only one pod that has a `key` label and a `label` label, not because the value matched.

With that in mind, my change actually changes the meaning of this test. I'm not sure what the intention of the old test was, but it didn't test filtering on label values. The new test does that, but maybe it misses something else. Thoughts?

xordspar0: In case anyone is curious, I found this by breaking out my trusty debugger.

1. I ran the debugger with `dlv exec bin/podman -- pod ps --no-trunc --filter label=io.podman.test.key --filter label=io.podman.test.label=value1`, copying the command line arguments from the test exactly.
2. I set a breakpoint on the following line (just after the filter command line arg processing): https://github.com/containers/podman/blob/a6b300ef7ee19da0c590af9bbcd1a35de7fa8c84/cmd/podman/pods/ps.go#L83
3. I ran `print psInput.Filters` in the debugger. I expected to see something like `["io.podman.test.label", "io.podman.test.label"]`, but that's not what I saw! Instead, this was the output:
  ```
map[string][]string [
	"label": [
		"io.podman.test.key",
		"io.podman.test.label",
	], 
]
```

I was surprised to see filters for two different labels, which made me realize that I had made a mistake when reading the test code earlier.

mheon:
/hold cancel
LGTM

xordspar0:
Yay! :tada:

PR #7760, opened 2020-09-23
Remove some SkipIfRootless flags from tests

rhatdan:
We need to get more tests running in rootless mode.  Since cgroupsV2 allows
management of cgroups in rootless environments a lot of more tests can be run.

### DIFF
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -11,7 +11,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/containers/podman/v2/pkg/cgroups"
 	. "github.com/containers/podman/v2/test/utils"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/mrunalp/fileutils"
@@ -50,7 +49,6 @@ var _ = Describe("Podman run", func() {
 	})
 
 	It("podman run a container based on a complex local image name", func() {
-		SkipIfRootless()
 		imageName := strings.TrimPrefix(nginx, "quay.io/")
 		session := podmanTest.Podman([]string{"run", imageName, "ls"})
 		session.WaitWithDefaultTimeout()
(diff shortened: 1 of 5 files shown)


QiWang19:
LGTM

rhatdan:
@edsantiago PTAL turns on a few more tests, we need to get the same fix for skipifrootless() that you did for skipifremote()
I will continue attempting to turn on rootless tests, and get @giuseppe to look at the cgroups failures on rootless.

@containers/podman-maintainers PTAL and merge.

test/e2e/run_test.go
```diff
@@ -601,27 +585,24 @@ USER bin`
...
-		Expect(session.OutputToString()).To(Equal("uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),18(audio),20(dialout),26(tape),27(video),777,65533(nogroup)"))
+		Expect(session.LineInOutputContains("777,65533(nogroup)")).To(BeTrue())
 	})
 
 	It("podman run with user (default)", func() {
-		SkipIfRootless()
 		session := podmanTest.Podman([]string{"run", "--rm", ALPINE, "id"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		Expect(session.OutputToString()).To(Equal("uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)"))
```
mheon, line 624: This seems a lot less specific

edsantiago:
@rhatdan I looked at this yesterday (while CI was still red) and it looked promising, although I agree with @mheon's concern re: dropping the string check. I can't do a full review today, sorry, am heading OOTO soon for the whole day.

test/e2e/run_test.go
rhatdan: Yes, they are very different in rootless versus rootful, but I do not know why.  But the current test is fragile since the ALPINE image can change in the future.  We just want to make sure that the container is running as root.

But you could very well be pointing out a rootless bug.

```
$ podman (pullpolicy) $ sudo podman run --rm alpine id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
$ podman (pullpolicy) $ podman run --rm alpine id
uid=0(root) gid=0(root)
```
I still think we should get this in, and I will open an issue above.

edsantiago: >the ALPINE image can change in the future

I am establishing what I hope is a robust convention in which `quay.io/libpod/testimage:YYYYMMDD` is an alpine-based image, entirely under our control, to avoid precisely this sort of problem. See https://github.com/containers/podman/blob/master/test/system/build-testimage

rhatdan: Opened an issue on this one, I believe we should merge this for and fix this specific issue later.
https://github.com/containers/podman/issues/7782

PR #7762, opened 2020-09-23
HTTP Attach: Wait until both STDIN and STDOUT finish

mheon:
In the old code, there was a chance that we could return when only one of STDIN or STDOUT had finished - this could lead to us dropping either input to the container, or output from it, in the case that one stream terminated early.

To resolve this, use separate channels to return STDOUT and STDIN errors, and track which ones have returned cleanly to ensure that we need bith in order to return from the HTTP attach function and pass control back to the HTTP handler (which would assume we exited cleanly and close the client's attach connection).

### DIFF
--- a/libpod/oci_conmon_exec_linux.go
+++ b/libpod/oci_conmon_exec_linux.go
@@ -537,9 +537,6 @@ func attachExecHTTP(c *Container, sessionID string, r *http.Request, w http.Resp
 		}
 	}()
 
-	// Make a channel to pass errors back
-	errChan := make(chan error)
-
 	attachStdout := true
 	attachStderr := true
 	attachStdin := true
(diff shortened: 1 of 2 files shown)


baude:
lgtm, confirmed it fixed the problem

libpod/oci_conmon_exec_linux.go
```diff
@@ -613,19 +613,38 @@ func attachExecHTTP(c *Container, sessionID string, r *http.Request, w http.Resp
...
+			if outDone && inDone {
+				return nil
+			}
+		case err := <-stdinChan:
+			if err != nil {
+				return err
+			}
+
+			inDone = true
+			if outDone && inDone {
```
rhatdan, line 643: Nit: Putting these three lines at the bottom would make this cleaner and not have duplicated code.

libpod/oci_conmon_linux.go
```diff
@@ -690,28 +690,47 @@ func (r *ConmonOCIRuntime) HTTPAttach(ctr *Container, req *http.Request, w http.
...
+			if outDone && inDone {
+				return nil
+			}
+		case err := <-stdinChan:
+			if err != nil {
+				return err
+			}
+
+			inDone = true
+			if outDone && inDone {
```
rhatdan, line 728: Ditto:

rhatdan:
LGTM other then tests failures.

mheon:
Do not merge yet if this goes green, have a testing commit on top.

mheon:
Alright, dropped the test commit. This should be ready once tests go green.

TomSweeneyRedHat:
LGTM

PR #7763, opened 2020-09-23
system tests: helpers: safer parse_table

edsantiago:
The parse_table() helper has until now dumbly split lines
on every single '|' character. This prevents us from running
simple tests such as 'cgroupManager: (systemd|cgroupfs)'.

We now use an ugly but robust sed expression to split
on '|' but *only* when surrounded by spaces and/or beginning
or end of line. This is safe because, for readability, all
tables already keep the '|' symbols well separated from
table content.

Add tests. And, the whole reason behind this, add
an actual real test for cgroupManager and cgroupVersion.

### DIFF
--- a/test/system/005-info.bats
+++ b/test/system/005-info.bats
@@ -19,8 +19,8 @@ graphRoot:
 graphStatus:
 imageStore:\\\s\\\+number: 1
 runRoot:
-cgroupManager:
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

vrothberg (commented):
LGTM

PR #7798, opened 2020-09-28
Use local image if input image is a manifest list

### LINKED ISSUE #7233: Inconsistent treatment of manifest lists in "podman run"

/kind bug

**Description**

A `podman run` with the image pointing to a local manifest list will fail.

A `podman run` with the image pointing to a manifest list in a registry will succeed.

At best this is confusing. At worst, it means there is no way to extend a manifest list, since it will be treated differently if created locally or pulled, and I have to remove it in order to use it.

Reproducer script:

```sh
#!/bin/bash -ex

M=docker.io/avikivity/test-manifest

podman pull docker.io/fedora:32
podman manifest create "$M"
podman manifest add --all "$M" docker.io/fedora:32
podman manifest push --all "$M" docker://"$M"
podman run --rm "$M" echo y || echo failed running image from manifest
podman rmi "$M"
podman run --rm "$M" echo y && echo succeeded running image from manifest
```

1. Edit the reproducer, substituting an unused name that you can push to for `M`. 

2. Run the reproducer

3. Wonder why the same `podman run` command fails or succeeds depending on local state

```
failed running image from manifest
succeeded running image from manifest
```

```
succeeded running image from manifest
```

```
Version:      2.0.4
API Version:  1
Go Version:   go1.14.6
Built:        Thu Jan  1 02:00:00 1970
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.15.0
  cgroupVersion: v2
  conmon:
    path: /usr/bin/conmon
  cpus: 12
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: avi
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.7.8-200.fc32.x86_64
  linkmode: dynamic
  memFree: 2141855744
  memTotal: 33542152192
  ociRuntime:
    name: crun
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      spec: 1.0.0
  os: linux
  remoteSocket:
  rootless: true
  slirp4netns:
    version: |-
      slirp4netns version 1.1.4
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 2
  swapFree: 15166337024
  swapTotal: 16869486592
  uptime: 359h 10m 34.1s (Approximately 14.96 days)
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - docker.io
store:
  configFile: /home/avi/.config/containers/storage.conf
  containerStore:
    number: 2
    paused: 0
    running: 1
    stopped: 1
  graphDriverName: overlay
  graphOptions:
    overlay.mount_program:
      Version: |-
        fusermount3 version: 3.9.1
        fuse-overlayfs: version 1.1.0
        FUSE library version 3.9.1
  graphRoot: /home/avi/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 33
  volumePath: /home/avi/.local/share/containers/storage/volumes
version:
  APIVersion: 1
  Built: 0
  BuiltTime: Thu Jan  1 02:00:00 1970
  GitCommit: ""
  GoVersion: go1.14.6
  OsArch: linux/amd64
  Version: 2.0.4
```

```
podman-2.0.4-1.fc32.x86_64
```

QiWang19:
If run&create image returns error: image contains manifest list, not a runnable image, find the local image that has digest matching the digest from the list and use the image from local storage for the command.

close #7233

### DIFF
--- a/pkg/specgen/generate/container.go
+++ b/pkg/specgen/generate/container.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/podman/v2/pkg/specgen"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
@@ -33,7 +34,43 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 
 		_, mediaType, err := newImage.Manifest(ctx)
 		if err != nil {
-			return nil, err
+			if errors.Cause(err) != image.ErrImageIsBareList {
+				return nil, err
+			}
+			// if err is not runnable image
+			// use the local store image with repo@digest matches with the list, if exists
+			manifestByte, manifestType, err := newImage.GetManifest(ctx, nil)
+			if err != nil {
+				return nil, err
+			}
+			list, err := manifest.ListFromBlob(manifestByte, manifestType)
+			if err != nil {
+				return nil, err
+			}
+			images, err := r.ImageRuntime().GetImages()
+			if err != nil {
+				return nil, err
+			}
+			findLocal := false
+			listDigest, err := list.ChooseInstance(r.SystemContext())
+			if err != nil {
+				return nil, err
+			}
+			for _, img := range images {
+				for _, imageDigest := range img.Digests() {
+					if imageDigest == listDigest {
+						newImage = img
+						s.Image = img.ID()
+						mediaType = manifestType
+						findLocal = true
+						logrus.Debug("image contains manifest list, using image from local storage")
+						break
+					}
+				}
+			}
+			if !findLocal {
+				return nil, image.ErrImageIsBareList
+			}
 		}
 
 		if s.HealthConfig == nil && mediaType == manifest.DockerV2Schema2MediaType {
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -609,4 +609,21 @@ var _ = Describe("Podman create", func() {
 		Expect(session.ExitCode()).ToNot(BeZero())
 	})
 
+	It("create use local store image if input image contains a manifest list", func() {
+		session := podmanTest.Podman([]string{"pull", BB})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(BeZero())
+
+		session = podmanTest.Podman([]string{"manifest", "create", "mylist"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"manifest", "add", "--all", "mylist", BB})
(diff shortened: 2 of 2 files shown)


rhatdan:

QiWang19:
@nalind PTAL

pkg/specgen/generate/container.go
```diff
@@ -33,7 +34,40 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
...
+						if imageDigest == listDigest {
+							newImage = img
+							s.Image = img.ID()
+							findLocal = true
+							logrus.Debug("image contains manifest list, using image from local storage")
+							break
+						}
+					}
+				}
+			}
```
nalind, line 67: This will match any local image which is also in the list, so it solves the problem when only one of them is present.  If I've pulled more than one using the repository name and their respective digests, this will choose the first one in the list, when the one for the current architecture (if one is present) is listed later in `images`.  Instead of checking for any of the values returned by `list.Instances()`, can we use `list.ChooseInstance()` to select the one that would be correct for the current architecture, and only accept an image that matches the digest it returns?

QiWang19: fixed.

QiWang19:
@nalind @containers/podman-maintainers  PTAL

nalind:
LGTM

pkg/specgen/generate/container.go
```diff
@@ -33,7 +34,42 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 
 		_, mediaType, err := newImage.Manifest(ctx)
```
rhatdan, line 35: Do you have to set this mediaType somewhere?

QiWang19: mediaType will be used to check if its a manifest.DockerV2Schema2MediaType
https://github.com/containers/podman/pull/7798/files#diff-5a91791543058d94a51dab8d8294ca7cR75

rhatdan: But if the newImage.Manifest(ctx) check fails, the nediaType is never set.

nalind: Ah, good catch.  If we do find a suitable image and update `newImage` to point to it, we need to make sure that `mediaType` is updated to the value that correctly describes its manifest.

QiWang19: If newImage.Manifest(ctx) fails, use the mediaType of the local image.

PR #7799, opened 2020-09-28
Make all Skips specify a reason

rhatdan:
Always use CGROUPV2 rather then reading from system all the time.

### DIFF
--- a/test/e2e/common_test.go
+++ b/test/e2e/common_test.go
@@ -599,6 +599,24 @@ func (p *PodmanTestIntegration) CreateSeccompJson(in []byte) (string, error) {
 	return jsonFile, nil
 }
 
+func SkipIfRootlessCgroupsV1(reason string) {
+	if len(reason) < 5 {
+		panic("SkipIfRootlessCgroupsV1 must specify a reason to skip")
+	}
+	if os.Geteuid() != 0 && !CGROUPSV2 {
+		Skip("[rootless]: " + reason)
+	}
+}
+
+func SkipIfRootless(reason string) {
+	if len(reason) < 5 {
+		panic("SkipIfRootless must specify a reason to skip")
+	}
+	if os.Geteuid() != 0 {
+		ginkgo.Skip("[rootless]: " + reason)
+	}
+}
+
 func SkipIfNotFedora() {
 	info := GetHostDistributionInfo()
 	if info.Distribution != "fedora" {
@@ -610,21 +628,21 @@ func isRootless() bool {
 	return os.Geteuid() != 0
 }
 
-func SkipIfCgroupV1() {
-	cgroupsv2, err := cgroups.IsCgroup2UnifiedMode()
-	Expect(err).To(BeNil())
-
-	if !cgroupsv2 {
-		Skip("Skip on systems with cgroup V1 systems")
+func SkipIfCgroupV1(reason string) {
+	if len(reason) < 5 {
+		panic("SkipIfCgroupV1 must specify a reason to skip")
+	}
+	if !CGROUPSV2 {
+		Skip(reason)
 	}
 }
 
-func SkipIfCgroupV2() {
-	cgroupsv2, err := cgroups.IsCgroup2UnifiedMode()
-	Expect(err).To(BeNil())
-
-	if cgroupsv2 {
-		Skip("Skip on systems with cgroup V2 systems")
+func SkipIfCgroupV2(reason string) {
+	if len(reason) < 5 {
+		panic("SkipIfCgroupV2 must specify a reason to skip")
+	}
+	if CGROUPSV2 {
+		Skip(reason)
 	}
 }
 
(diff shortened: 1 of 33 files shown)


test/e2e/common_test.go
```diff
@@ -599,6 +599,24 @@ func (p *PodmanTestIntegration) CreateSeccompJson(in []byte) (string, error) {
 	return jsonFile, nil
 }
 
+func SkipIfRootlessCgroupsV1(reason string) {
+	if reason == "" {
+		panic("SkiptIfRootlessCgroupsV1 must specify a reason to skip")
```
edsantiago, line 604: Typo, 'Skipt' has a spurious 't'. (Ibid below)

test/e2e/common_test.go
```diff
@@ -599,6 +599,24 @@ func (p *PodmanTestIntegration) CreateSeccompJson(in []byte) (string, error) {
...
+	}
+	if os.Geteuid() != 0 && !CGROUPSV2 {
+		Skip("[rootless]: " + reason)
+	}
+}
+
+func SkipIfRootless(reason string) {
+	if reason == "" {
```
edsantiago, line 612: How about `if len(reason) < 5` instead? (And, will `len()` handle nil?) When doing this type of check, I sometimes worry about people bypassing it with `" "` or `"x"` or `"foo"`. (Never deliberately, it's just that sometimes we write code in a hurry and try to get it passing with the best intentions to fix it later). Ibid below.

TomSweeneyRedHat:
Tests aren't hip @rhatdan

edsantiago (changes requested):
Mostly LGTM, but needs lint fixes (removing unused cgroups test), and a typo fix, and a suggestion.

test/e2e/common_test.go
rhatdan: Fixed

test/e2e/run_device_test.go
```diff
@@ -72,7 +72,7 @@ var _ = Describe("Podman run device", func() {
 	})
 
 	It("podman run device host device and container device parameter are directories", func() {
-		SkipIfRootless() // Can not create devices in /dev in rootless mode
+		SkipIfRootless("Can not create devices in /dev in rootless mode")
```
edsantiago, line 75: nit: "Cannot" is one word

test/e2e/run_test.go
```diff
@@ -368,16 +368,16 @@ USER bin`
 	})
 
 	It("podman run sysctl test", func() {
-		SkipIfRootless() // Network sysclts are not avalable root rootless
+		SkipIfRootless("Network sysclts are not avalable root rootless")
```
edsantiago, line 371: typo (transposition): should be `sysctls`

test/e2e/common_test.go
rhatdan: Fixed

test/e2e/run_device_test.go
rhatdan: Fixed

test/e2e/run_test.go
rhatdan: Fixed

test/e2e/libpod_suite_remote_test.go
```diff
@@ -24,16 +24,10 @@ func IsRemote() bool {
...
-}
-
-func SkipIfRootlessCgroupsV1() {
-}
-
-func SkipIfRootless() {
-	if os.Geteuid() != 0 {
-		ginkgo.Skip("This function is not enabled for rootless podman")
+	if reason == "" {
+		panic("SkiptIfRemote must specify a reason to skip")
```
edsantiago, line 28: One more Skipt-with-a-t

edsantiago (commented):
So close!

edsantiago (approved):

edsantiago:
(I'm not going to block because of that typo. If CI passes, and no other problems are found, I'm fine with merging)

test/e2e/libpod_suite_remote_test.go
rhatdan: I am sure we will need another pass, since I can not tests the tests locally right now.

test/e2e/pod_stats_test.go
```diff
@@ -16,10 +16,6 @@ var _ = Describe("Podman pod stats", func() {
 	)
 
 	BeforeEach(func() {
-		if os.Geteuid() != 0 {
```
vrothberg, line 19: This test does not work rootless. Can we do a skip rootless here?

The underlying issue is that we're always using `--cgroup-manager=cgroupfs` when executing Podman and that doesn't work as an ordinary user.

rhatdan: Re-added, although it would be good to figure out how we can get these tests to run with rootless and cgroups v2.

vrothberg: Thanks!

Absolutely agree on getting to run in rootless mode.

TomSweeneyRedHat:
LGTM
and dare I say "happy green test buttons"?

edsantiago:
Glad to see this go in; I think it'll really help us gain some coverage.
/lgtm
/hold

PR #7802, opened 2020-09-28
Fix handling of Ambient/Inheritable caps for non root user

rhatdan:
On F33 we are getting permission denied when changing default
capabilities of containers.  I believe Ineritable and Ambient
capabilities have to match Permissited and potentially Effective.

### DIFF
--- a/pkg/specgen/generate/security.go
+++ b/pkg/specgen/generate/security.go
@@ -136,6 +136,7 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 		configSpec.Process.Capabilities.Effective = caplist
 		configSpec.Process.Capabilities.Permitted = caplist
 		configSpec.Process.Capabilities.Inheritable = caplist
+		configSpec.Process.Capabilities.Ambient = caplist
 	} else {
 		userCaps, err := capabilities.NormalizeCapabilities(s.CapAdd)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe PTAL

rhatdan:
@baude this should fix the issue we have been seeing on F33.

saschagrunert (approved):

lsm5:
I tried a scratch build with this patch and it doesn't give me the capset error anymore. So LGTM.

lsm5 (approved):

vrothberg (commented):
Good catch. I keep forgetting the definitions of the different sets but after re-reading capabilities(7) the patch does exactly what's described for ordinary users :+1: 

LGTM

Still want a final head nod from @giuseppe

giuseppe:
alternative solution: https://github.com/containers/podman/pull/7820

I've checked what Docker 19 does so now Podman sets the same capabilities

rhatdan:
@giuseppe Have you tried yours against Fedora 33?

rhatdan:
https://github.com/containers/podman/pull/7820 is more secure

micah:
It seems like the capabilities system has had a few fairly major breakages in the recent releases. Are there tests that cover these cases, so they can be caught in the future by the CI?

rhatdan:
@micah we are adding them,  But one of them was newer kernel with crun bug, which was hard to catch.

PR #7647, opened 2020-09-16
Refactor remote pull to provide progress

jwhonce:
Fixes #7543

Tested with:

$ (echo '# start'; podman-remote pull nginx ) 2>&1 | ts '[%Y-%m-%d %H:%M:%.S]'
$ (echo '# start'; podman pull nginx ) 2>&1 | ts '[%Y-%m-%d %H:%M:%.S]'

Signed-off-by: Jhon Honce <jhonce@redhat.com>

PR #7451, opened 2020-08-25
Send HTTP Hijack headers after successful attach

mheon:
Our previous flow was to perform a hijack before passing a connection into Libpod, and then Libpod would attach to the container's attach socket and begin forwarding traffic.

A problem emerges: we write the attach header as soon as the attach complete. As soon as we write the header, the client assumes that all is ready, and sends a Start request. This Start may be processed *before* we successfully finish attaching, causing us to lose output.

The solution is to handle hijacking inside Libpod. Unfortunately, this requires a downright extensive refactor of the Attach and HTTP Exec StartAndAttach code. I think the result is an improvement in some places (a lot more errors will be handled with a proper HTTP error code, before the hijack occurs) but other parts, like the relocation of printing container logs, are just *bad*. Still, we need this fixed now to get CI back into good shape...

Fixes #7195

PR #7803, opened 2020-09-28
System tests: reenable some skipped tests

edsantiago:
- pause test: enable when rootless + cgroups v2
   (was previously disabled for all rootless)

 - run --pull: now works with podman-remote
   (in #7647, thank you @jwhonce)

 - various other run/volumes tests: try reenabling
   It looks like #7195 was fixed (by #7451? I'm not
   sure if I'm reading the conversation correctly).
   Anyway, remove all the skip()s on 7195. Only time
   will tell if it's really fixed)

Also:

 - new test for podman image tree --whatrequires
   (because TIL). Doesn't work with podman-remote.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -132,8 +132,6 @@ echo $rand        |   0 | $rand
 }
 
 @test "podman run --pull" {
-    skip_if_remote "podman-remote does not emit 'Trying to pull' msgs"
-
     run_podman run --pull=missing $IMAGE true
     is "$output" "" "--pull=missing [present]: no output"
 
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM if the tests pass
/approve

TomSweeneyRedHat:
@edsantiago looks like some network tests are failing.  I don't believe it's due to these changes.
LGTM
assuming happy tests

edsantiago:
@TomSweeneyRedHat yep, I'm having a field day filing issues right now

edsantiago:
New force push. No changes to my tests, but I did piggyback in a fix for a new e2e test that was not properly cleaning up after itself. And, filed two race-condition issues, #7807 and #7808.

test/e2e/network_test.go
```diff
@@ -301,11 +301,13 @@ var _ = Describe("Podman network", func() {
 		session := podmanTest.Podman([]string{"network", "create", netName1})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(BeZero())
+		defer podmanTest.removeCNINetwork(netName1)
```
Luap99, line 304: Unfortunately this will blow up.
We cannot remove the network twice.
The test already removes them.

edsantiago: @Luap99 good point. I've filed #7809 and will revert my attempt. Thanks.

Issue #7740, opened 2020-09-23
podman/docker compatibility during volume creation


rsommer:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
When a named volume is created via docker and the volume does already exist, no error ist returned. podman on the other hand returns `Error: volume with name test already exists: volume already exists`

**Steps to reproduce the issue:**

docker:
```
$ docker volume create test
test
$ docker volume create test
test
```

podman:
```
$ podman volume create test
test
$ podman volume create test
Error: volume with name test already exists: volume already exists
```

**Describe the results you received:**
podman throws an error when the volume does already exist.

**Describe the results you expected:**
podman just "returns" the existing volume, being compatible to docker. The behaviour was discovered while trying to replace the docker daemon in gitlab ci. The cache volumes used by the gitlab runner are (re-)created using the same volume name, leading to the shown error.

**Output of `podman version`:**

```
Version:      2.1.0
API Version:  2.0.0
Go Version:   go1.14
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.20, commit: '
  cpus: 2
  distribution:
    distribution: debian
    version: "10"
  eventLogger: journald
  hostname: cirunnertest01
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1221
      size: 1
    - container_id: 1
      host_id: 2459296
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1221
      size: 1
    - container_id: 1
      host_id: 2459296
      size: 65536
  kernel: 4.19.0-10-amd64
  linkmode: dynamic
  memFree: 209768448
  memTotal: 2066657280
  ociRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: |-
      runc version 1.0.0~rc6+dfsg1
      commit: 1.0.0~rc6+dfsg1-3
      spec: 1.0.1
  os: linux
  remoteSocket:
    path: /run/user/1221/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
[...]

PR #7804, opened 2020-09-28
fix for compatibility volume creation

baude:
in the compatibility layer, creating a volume with a name that already does not result in an error.  instead a 201 response with the existing volume's information is returned. while it seems like a bug on the part of docker and they agree, no attempt has been made to fix it in five years.  See https://github.com/moby/moby/issues/16068

Fixes: #7740

### DIFF
--- a/pkg/api/handlers/compat/volumes.go
+++ b/pkg/api/handlers/compat/volumes.go
@@ -93,6 +93,29 @@ func CreateVolume(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// See if the volume exists already
+	existingVolume, err := runtime.GetVolume(input.Name)
+	if err != nil && errors.Cause(err) != define.ErrNoSuchVolume {
+		utils.InternalServerError(w, err)
+		return
+	}
+
+	// if using the compat layer and the volume already exists, we
+	// must return a 201 with the same information as create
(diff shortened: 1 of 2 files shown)


test/e2e/run_passwd_test.go
```diff
@@ -108,7 +108,7 @@ USER 1000`
 	It("podman run numeric user not specified in container modifies group", func() {
 		session := podmanTest.Podman([]string{"run", "--read-only", "-u", "20001", BB, "mount"})
 		session.WaitWithDefaultTimeout()
-		Expect(session.ExitCode()).To(Equal(0))
+		pgExpect(session.ExitCode()).To(Equal(0))
```
jwhonce, line 111:
```suggestion
		pgExpect(session.ExitCode()).To(Equal(0))
```
What is pgExpect?

baude: leftover cruft from tests ... this should be no change. updated commit

psgreco:
Just tested and this solves https://github.com/containers/podman/issues/7740 for me

rhatdan:
I love bug for bug compatibility.

jwhonce:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #7805, opened 2020-09-28
Journald log driver test

Luap99:
Test that the journald log driver writes to journald and
that we can read it with journalctl.


Luap99:
@edsantiago PTAL

edsantiago:
LGTM. I was concerned that we might run CI on non-systemd hosts, but it looks like we're OK in that respect. Thanks for the PR.

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #7811, opened 2020-09-28
Ignore containers.conf sysctl when namespaces set to host

rhatdan:
If user sets namespace to host, then default sysctls need to be ignored
that are specific to that namespace.

--net=host ignore sysctls that begin with net.
--ipc=host ignore fs.mqueue
--uts=host ignore kernel.domainname and kernel.hostname

### DIFF
--- a/test/e2e/containers_conf_test.go
+++ b/test/e2e/containers_conf_test.go
@@ -179,6 +179,12 @@ var _ = Describe("Podman run", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(ContainSubstring("1000"))
+
+		// Ignore containers.conf setting if --net=host
+		session = podmanTest.Podman([]string{"run", "--rm", "--net", "host", fedoraMinimal, "cat", "/proc/sys/net/ipv4/ping_group_range"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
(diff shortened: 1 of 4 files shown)


rhatdan:
@baude @giuseppe PTAL
This should fix the F33 issue with default sysctls.

giuseppe (approved):
LGTM

test/e2e/run_test.go
```diff
@@ -373,6 +373,11 @@ USER bin`
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(ContainSubstring("net.core.somaxconn = 65535"))
+
+		// network sysctls should fail if --net=host is set
```
vrothberg, line 377: Should do the same for IPC and UTS

test/e2e/containers_conf_test.go
```diff
@@ -179,6 +179,12 @@ var _ = Describe("Podman run", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(ContainSubstring("1000"))
+
+		// Ignore containers.conf setting if --net=host
```
vrothberg, line 183: Should do the same for IPC and UTS

vrothberg (commented):
LGTM but I'd like to increase test coverage by testing all namespaces.

TomSweeneyRedHat:
LGTM
extra testing per @vrothberg would be welcomed

PR #7812, opened 2020-09-28
Make the e2e test network cleanup more robust.

### LINKED ISSUE #7809: test/e2e : podman network remove with two networks: needs fail-safe cleanup
test/e2e/network_test.go:"podman network remove with two networks" does not properly clean up after itself if the test fails partway (e.g. if `net1` succeeds but `net2` fails).

Luap99:
We need to check if the network exists before trying to
remove them. This allows the `removeCNINetwork` function
always to be called. This is needed by tests which are trying
to remove the network manually in order to prevent flakes.

Fixes #7809

### DIFF
--- a/test/e2e/network_create_test.go
+++ b/test/e2e/network_create_test.go
@@ -56,9 +56,16 @@ func genericPluginsToPortMap(plugins interface{}, pluginType string) (network.Po
 }
 
 func (p *PodmanTestIntegration) removeCNINetwork(name string) {
-	session := p.Podman([]string{"network", "rm", "-f", name})
+	// Some tests remove the network manual but they should
+	// still defer this function in order to prevent flakes.
+	// So only remove the network if it exists.
+	session := p.Podman([]string{"network", "inspect", name})
 	session.WaitWithDefaultTimeout()
-	Expect(session.ExitCode()).To(BeZero())
+	if session.ExitCode() == 0 {
+		session = p.Podman([]string{"network", "rm", "-f", name})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(BeZero())
+	}
 }
 
 func removeNetworkDevice(name string) {
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -135,7 +135,6 @@ var _ = Describe("Podman network", func() {
 	})
 
 	It("podman network rm", func() {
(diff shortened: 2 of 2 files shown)


Luap99:
@edsantiago PTAL

test/e2e/network_create_test.go
```diff
@@ -56,9 +56,16 @@ func genericPluginsToPortMap(plugins interface{}, pluginType string) (network.Po
...
+	// still defer this function in order to prevent flakes.
+	// So only remove the network if it exists.
+	session := p.Podman([]string{"network", "inspect", name})
 	session.WaitWithDefaultTimeout()
-	Expect(session.ExitCode()).To(BeZero())
+	if session.ExitCode() == 0 {
+		session = p.Podman([]string{"network", "rm", "-f", name})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(BeZero())
+	}
```
edsantiago, line 68: Sorry for the late review; I had a concern and wanted to sleep on it.

My concern is that many of the e2e tests are already hard to read, and `podman network inspect` will add thirty more lines to already-long output. What would you think of just always force-running `podman network rm -f`, and just not bothering to check exit status? This is cleanup, not a test: the tests that need to check `network rm` already do so explicitly. 

That leaves the possibility that `network rm -f` could fail for a reason other than `network not found`. I don't know how likely that is, or how serious. (I suspect that it would lead to further failures that would be caught, but don't really know). So, I'm not filing this as a please-change, more like a please-discuss. @mheon @giuseppe @rhatdan WDYT?

rhatdan: Well for consistency we could make podman rm network exit with 1 when the name does not exists like we do with containers.
```
$ podman network rm foobar1; echo $?
Error: unable to find network configuration for foobar1: network not found
125
$ podman container rm foobar1; echo $?
Error: no container with name or ID foobar1 found: no such container
1
podman image rm foobar1; echo $?
Error: 1 error occurred:
	* unable to find a name and tag match for foobar1 in repotags: no such image
1

```
Then we could check if the exit code was greater then 1?

edsantiago: Oh, I really like that. 125 troubles me, because it (should be) reserved for podman-internal errors.

rhatdan:
@Luap99 My PR has merged, I think it would be best to check the exit code status on removing the network and then ignore failures if the exit code is 1.

PR #7815, opened 2020-09-28
Add X-Registry-Config support

jwhonce:
* Refactor auth pkg to support X-Registry-Config
* Refactor bindings to support X-Registry-Config
* Refactor build endpoint to support X-Registry-Config. Supports:
  * --creds
  * --authfile
* Added X-Reference-Id Header to http.Request to support log event
  correlation
* Log headers from http.Request


jwhonce:
@vrothberg Please take a look

pkg/auth/auth.go
```diff
@@ -15,21 +16,96 @@ import (
...
+// the necessary authentication information for libpod operations
+func GetCredentials(r *http.Request) (*types.DockerAuthConfig, string, HeaderAuthName, error) {
+	has := func(key HeaderAuthName) bool { hdr, found := r.Header[string(key)]; return found && len(hdr) > 0 }
+	switch {
+	case has(XRegistryConfigHeader):
+		c, f, err := getConfigCredentials(r)
+		return c, f, XRegistryConfigHeader, err
+	case has(XRegistryAuthHeader):
+		c, f, err := getAuthCredentials(r)
+		return c, f, XRegistryConfigHeader, err
```
vrothberg, line 41: That should be XRegistryAuthHeader.

It looks like the 2nd return value is only used for error messages. Can we simplify further by embedding it in the error message here?

pkg/bindings/images/build.go
```diff
@@ -93,6 +94,19 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
 		params.Set("labels", l)
 	}
 
+	var (
+		headers map[string]string
+		err     error
+	)
+	if options.SystemContext.DockerAuthConfig != nil {
```
vrothberg, line 101: This could be a nil deref on `options.SystemContext` so we need an additional check for that.

pkg/auth/auth.go
```diff
@@ -51,38 +127,75 @@ func GetCredentials(r *http.Request) (*types.DockerAuthConfig, string, error) {
...
 
-	header := make(map[string]string)
-	header[XRegistryAuthHeader] = content
+	if len(authConfigs) == 0 {
+		return "", nil
+	}
+	return encodeMultiAuthConfigs(authConfigs)
+}
+
+// headerAuth returns a map with the XRegistryAuthHeader set which can
```
vrothberg, line 181:
```suggestion
// headerAuth returns a base64 encoded map with the XRegistryAuthHeader set which can
```

pkg/auth/auth.go
```diff
@@ -51,38 +127,75 @@ func GetCredentials(r *http.Request) (*types.DockerAuthConfig, string, error) {
...
+		return nil, err
+	}
+
+	if len(content) > 0 {
+		return map[string]string{string(headerName): content}, nil
+	}
+	return nil, nil
+}
+
+// headerConfig returns a map with the XRegistryConfigHeader set which can
```
vrothberg, line 154:
```suggestion
// headerConfig returns a base64 encoded map with the XRegistryConfigHeader set which can
```

pkg/auth/auth.go
```diff
@@ -15,21 +16,96 @@ import (
...
+func (h HeaderAuthName) String() string { return string(h) }
+
 // XRegistryAuthHeader is the key to the encoded registry authentication
 // configuration in an http-request header.
-const XRegistryAuthHeader = "X-Registry-Auth"
+const XRegistryAuthHeader HeaderAuthName = "X-Registry-Auth"
+
+// XRegistryConfigHeader is the key to the encoded registry authentication
+// configuration in an http-request header.
+const XRegistryConfigHeader HeaderAuthName = "X-Registry-Config"
```
vrothberg, line 29: Can you drop a comment that this refers to `build` only?

vrothberg (commented):
Just a couple of nits.

Can you document in `pkg/auth` why we have two kinds headers?

It's also worth mentioning that X-Registry-Config and X-Registry-Auth _can_ essentially be the same thing: multi-auth headers. The Atomic Docker in old RHEL and Fedora etc. has used X-Registry-Auth for two things:
1. Vanilla Docker compatible single auth
2. Atomic multi-auth headers to transparently support search registries where even a pull may contact multiple registries

Docker seems to have X-Registry-Config for the multi-auth case where only `build` may pull from more than one registry. This may confuse future generations so I think we document that explicitly in the code.

* libpod uses X-Registry-Auth only where it can be a single and a multi-auth header
* compat-layer supports X-Registry-Config for build

As libpod already supports the multi-auth header but not the X-Registry-Config, I think we should break `/libpod/build` into a separate handler. We could also only look for X-Registry-Config if the request is for compat.

In other words: Can we use X-Registry-Config only for compat build but not for libpod? Libpod already supports multi-auth headers and using X-Registry-Config now would break libpod/build.

vrothberg:
@jwhonce, did you check how Atomic Docker deals with the auth header for build?

We may need a fallback on the server side and check for both headers to be compatible with vanilla Docker (Config) and Atomic (potentially multi-auth Auth).

pkg/api/handlers/compat/images_push.go
```diff
@@ -49,9 +49,9 @@ func PushImage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	authConf, authfile, err := auth.GetCredentials(r)
+	authConf, authfile, key, err := auth.GetCredentials(r)
 	if err != nil {
-		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse %q header for %s", auth.XRegistryAuthHeader, r.URL.String()))
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "Failed to parse %q header for %s", key, r.URL.String()))
```
TomSweeneyRedHat, line 54: I hate the "Something went wrong" message with a passion.  Can we tell them what went wrong?  Perhaps "Unable to retrieve credentials" or similar?

TomSweeneyRedHat: Or I see below you've "Failed to retrieve repository credentials", that works too.

pkg/auth/auth.go
jwhonce: Fixed bug.  I could move error handling, but it would really break segregation of duties as I would need to format the http body here.

pkg/auth/auth.go
jwhonce: I added "As of API..." messages.

pkg/api/handlers/compat/images_push.go
jwhonce: @vrothberg 

> @jwhonce, did you check how Atomic Docker deals with the auth header for build?
> 
> We may need a fallback on the server side and check for both headers to be compatible with vanilla Docker (Config) and Atomic (potentially multi-auth Auth).

auth.GetCredentials() pulls whichever header is provided and returns a libpod compatible structure.  So we're covered.

vrothberg (commented):
LGTM, nice work @jwhonce !

rhatdan:
@jwhonce Is this still a WIP PR?

PR #7820, opened 2020-09-29
capabilities: always set ambient and inheritable

giuseppe:
change capabilities handling to reflect what docker does.

Bounding: set to caplist
Inheritable: set to caplist
Effective: if uid != 0 then clear; else set to caplist
Permitted: if uid != 0 then clear; else set to caplist
Ambient: clear

### DIFF
--- a/pkg/specgen/generate/security.go
+++ b/pkg/specgen/generate/security.go
@@ -131,12 +131,13 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 	}
 
 	configSpec := g.Config
+	configSpec.Process.Capabilities.Ambient = []string{}
 	configSpec.Process.Capabilities.Bounding = caplist
+	configSpec.Process.Capabilities.Inheritable = caplist
 
 	if s.User == "" || s.User == "root" || s.User == "0" {
 		configSpec.Process.Capabilities.Effective = caplist
 		configSpec.Process.Capabilities.Permitted = caplist
-		configSpec.Process.Capabilities.Inheritable = caplist
 	} else {
 		userCaps, err := capabilities.NormalizeCapabilities(s.CapAdd)
 		if err != nil {
--- a/test/e2e/cp_test.go
+++ b/test/e2e/cp_test.go
@@ -269,11 +269,11 @@ var _ = Describe("Podman cp", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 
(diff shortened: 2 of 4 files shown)


giuseppe:
the test failure seems correct to me, why would we allow chown to work with `-u USER` and no caps added?

```
		setup := podmanTest.RunTopContainer("testctr")
		setup.WaitWithDefaultTimeout()
		Expect(setup.ExitCode()).To(Equal(0))

		session = podmanTest.Podman([]string{"exec", "-u", "testuser", "testctr", "touch", "testfile"})
```

TomSweeneyRedHat:
LGTM
assuming happy tests

pkg/specgen/generate/security.go
```diff
@@ -131,19 +131,16 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
...
 		configSpec.Process.Capabilities.Permitted = caplist
-		configSpec.Process.Capabilities.Inheritable = caplist
 	} else {
-		userCaps, err := capabilities.NormalizeCapabilities(s.CapAdd)
-		if err != nil {
-			return errors.Wrapf(err, "capabilities requested by user are not valid: %q", strings.Join(s.CapAdd, ","))
-		}
-		configSpec.Process.Capabilities.Effective = userCaps
-		configSpec.Process.Capabilities.Permitted = userCaps
+		configSpec.Process.Capabilities.Effective = []string{}
```
rhatdan, line 142: This is a breaking change.
In podman 1.9 we ran users with the permitted and effective capability when doing a --cap-add
This allows for containers like

$ podman run --cap-add net-bind-host --user 1234 fedora httpd

To work without ever needing a setuid application.

giuseppe: I wasn't aware of such difference with Docker.

In this case, should we use your PR instead?

rhatdan: Lets just restore this section:
```
} else {
		userCaps, err := capabilities.NormalizeCapabilities(s.CapAdd)
		if err != nil {
			return errors.Wrapf(err, "capabilities requested by user are not valid: %q", strings.Join(s.CapAdd, ","))
		}
		configSpec.Process.Capabilities.Effective = userCaps
		configSpec.Process.Capabilities.Permitted = userCaps
```
And go forward with this PR.

giuseppe: fixed

giuseppe:
/hold

let's make sure tests are still passing :-)

lsm5:
@giuseppe this likely needs to be backported to v2.1

giuseppe:
backport: https://github.com/containers/podman/pull/7898

Issue #6381, opened 2020-05-26
looking up local images should consider search registries


juhp:
/kind bug

**Description**

`podman run <imagename>` chose an older local image.

**Steps to reproduce the issue:**

1. `podman pull docker.io/fedora:33 && podman pull registry.fedoraproject.org/fedora:33`
2. `podman run -it --rm fedora:33`
3. check which image is chosen

**Describe the results you received:**
For me the older image was run (in my case my docker.io/fedora:33 was several months old).

**Describe the results you expected:**
Newer image to used.
Or is this a security feature?

**Output of `podman version`:**

```
Version:            1.9.2
RemoteAPI Version:  1
Go Version:         go1.13.10
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.13.10
  podmanVersion: 1.9.2
host:
  arch: amd64
  buildahVersion: 1.14.8
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.16-2.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.16, commit: aec991fec16dc45935de184f2ea06a6ffca200a0'
  cpus: 8
  distribution:
    distribution: fedora
    version: "31"
  eventLogger: file
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.6.13-200.fc31.x86_64
  memFree: 8420085760
  memTotal: 16416141312
  ociRuntime:
    name: crun
    package: crun-0.13-2.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.13
      commit: e79e4de4ac16da0ce48777afb72c6241de870525
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.0.1-1.fc31.x86_64
    version: |-
      slirp4netns version 1.0.1
      commit: 6a7b16babc95b6a3056b33fb45b74a6f62262dd4
      libslirp: 4.1.0
[...]

PR #7823, opened 2020-09-29
image look up: consult registries.conf

vrothberg:
When looking up local images, take the unqualified-serach registries of
the registries.conf into account (on top of "localhost/").

Also extend the integration tests to prevent future regressions.

Fixes: #6381

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -410,48 +410,92 @@ func (ir *Runtime) getLocalImage(inputName string) (string, *storage.Image, erro
 	if inputName == "" {
 		return "", nil, errors.Errorf("input name is blank")
 	}
+
 	// Check if the input name has a transport and if so strip it
 	dest, err := alltransports.ParseImageName(inputName)
 	if err == nil && dest.DockerReference() != nil {
 		inputName = dest.DockerReference().String()
 	}
 
-	img, err := ir.getImage(stripSha256(inputName))
+	// Early check for fully-qualified images and (short) IDs.
+	img, err := ir.store.Image(stripSha256(inputName))
 	if err == nil {
-		return inputName, img, err
+		return inputName, img, nil
 	}
 
-	// container-storage wasn't able to find it in its current form
-	// check if the input name has a tag, and if not, run it through
-	// again
+	// Note that it's crucial to first decompose the image and check if
+	// it's a fully-qualified one or a "short name".  The latter requires
+	// some normalization with search registries and the
+	// "localhost/prefix".
 	decomposedImage, err := decompose(inputName)
 	if err != nil {
+		// We may have a storage reference.  We can't parse it to a
+		// reference before. Otherwise, we'd normalize "alpine" to
+		// "docker.io/library/alpine:latest" which would break the
+		// order in which we should query local images below.
+		if ref, err := is.Transport.ParseStoreReference(ir.store, inputName); err == nil {
+			img, err = is.Transport.GetStoreImage(ir.store, ref)
+			if err == nil {
+				return inputName, img, nil
+			}
+		}
 		return "", nil, err
 	}
 
-	// The image has a registry name in it and we made sure we looked for it locally
-	// with a tag.  It cannot be local.
+	// The specified image is fully qualified, so it doesn't exist in the
+	// storage.
 	if decomposedImage.hasRegistry {
+		// However ... we may still need to normalize to docker.io:
+		// `docker.io/foo` -> `docker.io/library/foo`
+		if ref, err := is.Transport.ParseStoreReference(ir.store, inputName); err == nil {
+			img, err = is.Transport.GetStoreImage(ir.store, ref)
+			if err == nil {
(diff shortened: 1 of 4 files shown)


vrothberg:
@edsantiago maybe something to consider for the system tests as well?

vrothberg:

libpod/image/image.go
```diff
@@ -410,48 +410,86 @@ func (ir *Runtime) getLocalImage(inputName string) (string, *storage.Image, erro
...
-	// grab all the local images
+	// Last resort: look at the repotags of all images and try to find a
+	// match.
 	images, err := ir.GetImages()
 	if err != nil {
 		return "", nil, err
 	}
 
-	// check the repotags of all images for a match
+	logrus.Errorf("HERE")
```
QiWang19, line 488:
```suggestion
```

vrothberg: Thanks!

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests.
Is this work related to the shortname work that @ParkerVR has been pursuing?

ParkerVR:
> 
> 
> LGTM
> assuming happy tests.
> Is this work related to the shortname work that @ParkerVR has been pursuing?

Yes, Valentin took over on this project. @vrothberg lmk if you'd like help with tests or additional features, I quite liked working on this

vrothberg:
> Is this work related to the shortname work that @ParkerVR has been pursuing?

It's unrelated to the "secure short names" work, which focuses on how we pull from registries. This PR focuses on how we locate images in the container storage.

vrothberg:
>  @vrothberg lmk if you'd like help with tests or additional features, I quite liked working on this

Much appreciated, thanks, @ParkerVR! I didn't start yet as we're still in bug-hunting mode but I expect to pick it up soon :)

vrothberg:
I now skip the tests for remote as `podman-remote untag` still has issues. I'll tackle them now.

vrothberg:
/hold
Let's wait for untag to be fixed first. I don't want to merge it untested for the remote client.

vrothberg:
https://github.com/containers/podman/pull/7840

vrothberg:
Looking good now :) Untag is fixed and the tests turn green.

libpod/image/image.go
```diff
@@ -410,48 +410,92 @@ func (ir *Runtime) getLocalImage(inputName string) (string, *storage.Image, erro
...
 		return "", nil, errors.Errorf("input name is blank")
 	}
+
 	// Check if the input name has a transport and if so strip it
 	dest, err := alltransports.ParseImageName(inputName)
 	if err == nil && dest.DockerReference() != nil {
 		inputName = dest.DockerReference().String()
 	}
 
-	img, err := ir.getImage(stripSha256(inputName))
```
mtrmac, line 419: This removes the only caller of `getImage` with a name AFAICS; the remaining  _know_ that they are using a full image ID; so, `getImage` should become `getImageByID` and use either `NewStoreReference(store, nil, id)` (more explicit about the intent) or perhaps `store.Image` (bypasses c/image, but allows names as input), without the overhead/risk of `ParseStoreReference` allowing other inputs.

libpod/image/image.go
```diff
@@ -410,48 +410,92 @@ func (ir *Runtime) getLocalImage(inputName string) (string, *storage.Image, erro
...
-	// grab all the local images
+	// Last resort: look at the repotags of all images and try to find a
+	// match.
 	images, err := ir.GetImages()
 	if err != nil {
 		return "", nil, err
 	}
 
-	// check the repotags of all images for a match
+	decomposedImage, err = decompose(inputName)
```
mtrmac, line 495: This was already computed around line 430.

pkg/registries/registries.go
```diff
@@ -1,5 +1,10 @@
 package registries
 
+// TODO: this package should not exist anymore.  Users should either use
+// c/image's `sysregistriesv2` package directly OR, even better, we cache a
+// config in libpod's image runtime so we don't need to parse the
+// registries.conf files redundantly.
```
mtrmac, line 6: `sysregistriesv2` is already caching the files.

mtrmac (commented):
ACK, thanks for adding the comments!

libpod/image/image.go
vrothberg: Should have added a comment. This was needed as the decomposedImage.referenceWithRegistry above has side effects.

libpod/image/image.go
vrothberg: Very good catch. I can open a follow-up PR to clean up :+1:

libpod/image/image.go
mtrmac: I can’t see any side effects — `referenceWithRegistry` just reads `ip.hasRegistry` and calls `ip.unnormalizedRef.String()` AFAICS; if it had side effects, would it even be possible to call it in a loop?

PR #7770, opened 2020-09-24
Properly handle podman run --pull command

rhatdan:
Currently the --pull missing|always|never is ignored

This PR implements this for local API.  For remote we
need to default to pullpolicy specified in the containers.conf
file.

Also fixed an issue when images were matching other images names
based on prefix, causing images to always be pulled.

I had named an image myfedora and when ever I pulled fedora, the system
thought that it there were two images named fedora since it was checking
for the name fedora as well as the prefix fedora.  I changed it to check
for fedora and the prefix /fedora, to prefent failures like I had.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #7824, opened 2020-09-29
DO NOT MERGE! Revert "Properly handle podman run --pull command"

edsantiago:
DO NOT MERGE! This is a dummy PR for the sole purpose of
running CI tests.

I really need to understand why my system tests did not catch
the "podman run --pull" problem addressed by #7770. So this
commit leaves the new tests in place, but reverts the code
changes. That should make these tests fail, and I will be
better able to understand why my tests failed to catch it.

I thought I had ginkgo working on my laptop, but running:

   ginkgo -trace -debug -r -focus 'with --pull' test/e2e

...only yields success. I suspect there's something deeper
going on here but can only find out by running the full CI
test suite. Sorry to be wasting cycles.

This reverts commit 1b5853e64794403d80c4d339c97b61dd63dd093a.

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -363,7 +363,7 @@ func GetCreateFlags(cf *ContainerCLIOpts) *pflag.FlagSet {
 	)
 	createFlags.StringVar(
 		&cf.Pull,
-		"pull", containerConfig.Engine.PullPolicy,
+		"pull", "missing",
 		`Pull image before creating ("always"|"missing"|"never")`,
 	)
 	createFlags.BoolVarP(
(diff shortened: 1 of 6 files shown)


rhatdan:
@edsantiago I think the common case was covered before my PR which is why your tests were passing. The !alltags path was not being handled as well.

if strings.HasSuffix(dName, "/"+searchName) && dSuspiciousTagValueForSearch == searchSuspiciousTagValueForSearch {
--------------------------------------^

This change is what was causing all of the failures I was seeing.  Basically I had two images within my storage myfedora and fedora, and because of this everytime I tried to run a fedora container, the code was reaching out to the registry.

edsantiago:
@rhatdan ack, I have a test for that right now, and it is (as expected) failing in pre-7770, passing in post. I will submit a PR within the hour. Thank you for the clarification!

edsantiago:
Submitted: #7828 to add new tests.

Issue #7484, opened 2020-08-27
systemd user script `--healthcheck-command ...` always has Status: starting


bbros-dev:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman healthcheck command never changes status from 'starting' state.

```bash
$ podman inspect rqlite-5.4.0 | jq '.[0]["State"]["Healthcheck"]'
{
  "Status": "starting",
  "FailingStreak": 0,
  "Log": null
}
```

**Steps to reproduce the issue:**

1. Generate systemd unit file

```bash
podman pull rqlite/rqlite:5.4.0
podman create --name rqlite-5.4.0 rqlite/rqlite:5.4.0
podman generate systemd --files --new --name --restart-policy=always rqlite-5.4.0
```

2. Alter `ExecStart` to:

```
ExecStart=/usr/bin/podman run \
  --conmon-pidfile %t/container-rqlite-5.4.0.pid \
  --cidfile %t/container-rqlite-5.4.0.ctr-id \
  --cgroups=no-conmon \
  -d \
  --replace \
  --publish 4001:4001 \
  --publish 4002:4002 \
  --healthcheck-command 'CMD-SHELL curl http://localhost:4001 && curl http://localhost:4002 && echo "Okay" && exit 0 || exit 1' \
  --healthcheck-start-period 5s \
  --healthcheck-retries 5 \
  --name rqlite-5.4.0 \
  rqlite/rqlite:5.4.0
```

3. Enable and start the service

```bash
cp container-rqlite-5.4.0.service $HOME/.config/systemd/user/container-rqlite-5.4.0.service
systemctl --user enable container-rqlite-5.4.0
systemctl --user start container-rqlite-5.4.0
systemctl --user status container-rqlite-5.4.0
```

**Describe the results you received:**

```bash
$ podman inspect rqlite-5.4.0 | jq '.[0]["State"]["Healthcheck"]'
{
  "Status": "starting",
  "FailingStreak": 0,
  "Log": null
}
```

**Describe the results you expected:**

1. Some log entries showing the healthcheck-command ran
1. Healthcheck:Status should be Healthy

**Additional information you deem important (e.g. issue happens only occasionally):**

Reproducible.

**Output of `podman version`:**

```
$ podman version
Version:      2.0.4
API Version:  1
Go Version:   go1.14.4
Built:        Thu Jan  1 10:00:00 1970
OS/Arch:      linux/amd64
```

[...]

PR #7827, opened 2020-09-29
healthchecks: return systemd-run error

vrothberg:
In case `systemd-run` errors when creating transient unit files (and
timers), create an error based on the combined output from stdout and
stderr.  Using the error from `exec.Command` contains the exit code
only which is not useful to debug (see #7484).

### DIFF
--- a/libpod/healthcheck_linux.go
+++ b/libpod/healthcheck_linux.go
@@ -35,9 +35,8 @@ func (c *Container) createTimer() error {
 	conn.Close()
 	logrus.Debugf("creating systemd-transient files: %s %s", "systemd-run", cmd)
 	systemdRun := exec.Command("systemd-run", cmd...)
-	_, err = systemdRun.CombinedOutput()
-	if err != nil {
-		return err
+	if output, err := systemdRun.CombinedOutput(); err != nil {
+		return errors.Errorf("%s", output)
 	}
 	return nil
 }


vrothberg:
@baude PTAL

libpod/healthcheck_linux.go
```diff
@@ -35,9 +35,8 @@ func (c *Container) createTimer() error {
 	conn.Close()
 	logrus.Debugf("creating systemd-transient files: %s %s", "systemd-run", cmd)
 	systemdRun := exec.Command("systemd-run", cmd...)
-	_, err = systemdRun.CombinedOutput()
-	if err != nil {
-		return err
+	if output, err := systemdRun.CombinedOutput(); err != nil {
+		return errors.Errorf("%s", output)
```
rhatdan, line 39: Can you do a 
  return errors.Wrapf(err, output)

vrothberg: `err` is really meaningless. It will give a confusing `exit 1`.

rhatdan: Ok,

rhatdan: LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

vrothberg:
merge me :)

PR #7828, opened 2020-09-29
System tests: corner case for run --pull

edsantiago:
Obscure corner case in which 'podman run --pull=never alpine'
will actually pass *with no alpine image* if there's an
image named "myalpine". (i.e. a substring match, not full
string match). Fixed in #7770 but the tests that were added
there do not actually test that.

This adds a double-duty test for that as well as making
sure that 'run --pull=never SHORTNAME' (implicit :latest)
does not match our existing :YYYYMMDD image; then one
more quick test to make sure that if we tag as :latest,
the same --pull=never succeeds.


edsantiago:
Oh, forgot to mention: this new test (correctly) fails in pre-7770 podman.

rhatdan:
LGTM
/approve

TomSweeneyRedHat:
LGTM

PR #7831, opened 2020-09-29
We already set container=podman environment variable

rhatdan:
Only need to set container, no need for containers

### DIFF
--- a/pkg/specgen/generate/container.go
+++ b/pkg/specgen/generate/container.go
@@ -75,8 +75,8 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 	if err != nil {
 		return nil, errors.Wrap(err, "error parsing fields in containers.conf")
 	}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Does it hurt to leave it other than the cost of the check?  Seems like a nice fail-safe if we ever mess up the envvar definition.

mheon:
I think the check needs to remain, but should be `"container"` instead of `"containers"` for consistency.

rhatdan:
This is ready to go in.

saschagrunert (approved):
LGTM

edsantiago:
LGTM but is the `""` check really the desired intention? That is: do you want to allow a user to override it?
```console
$ ./bin/podman run -it --rm -e container=foo $iii sh
container# echo $container
foo    <--- should podman allow user to override?
```

rhatdan:
@edsantiago Yes, we want to allow users to override the environment variable, we should only be setting it iff the user and image did NOT set it.

Issue #7808, opened 2020-09-28
race: remote,rootless: podman network create followed by run


edsantiago:
I've seen this one twice so far:
```console
$ podman-remote network create --subnet 172.16.206.0/24 testnet-TrgvHhKQAK
/etc/cni/net.d/testnet-TrgvHhKQAK.conflist
$ podman-remote run --rm --network testnet-TrgvHhKQAK quay.io/libpod/testimage:20200917 ip a
Error: error preparing container 8efdbe7260f5e082c2a256ee0781a23fecbcd0776d5ac1b43d924d39cfae5086 for attach: error configuring network namespace for container 8efdbe7260f5e082c2a256ee0781a23fecbcd0776d5ac1b43d924d39cfae5086: CNI network "testnet-TrgvHhKQAK" not found
[ rc=127 (** EXPECTED 0 **) ]
```

Log: [from PR 7803, ubuntu remote](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4668854372663296/html/system_test.log.html)

Again, I'm not going to spend any time right now trying to reproduce (acting on the assumption that this will be an easy one to track down). I will try to keep track of flake occurrences and log them here.

edsantiago:
I wonder if the race is in CNI? And I wonder if this is related? It's a very frequent flake in `podman network rm`, the classic symptom being:
```
Running: podman [options] network ls --quiet
podman-integrationtest
Running: podman [options] network rm podman-integrationtest
podman-integrationtest
Running: podman [options] network ls --quiet
podman-integrationtest
remove /home/some18764dude/.config/cni/net.d/eaac2f093ac23f24b699d68d1118f986b8620b03e3d5974e14c12147e06dfdab.conflist: no such file or directory
```
Recent instances (all one-offs, no triplets which cause a complete CI abort):

Podman network [It] podman network rm
* fedora : in-podman fedora
  * PR #7676
    * [09-17 13:05](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6365935319646208/html/integration_test.log.html#t--podman-network-rm--1)
* fedora : special_testing_rootless
  * PR #7777
    * [09-25 04:55](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4733700711645184/html/integration_test.log.html#t--podman-network-rm--1)
* fedora : test fedora
  * PR #7621
[...]

PR #7835, opened 2020-09-29
Fix some flakes in the e2e network tests.

Luap99:
The network test created config files with random filenames
but the network name was static. Since the tests can run in
parallel podman was not able to distinguish the networks.
We need to make sure that each test has its own config file
and network name. This helps to prevent unnecessary flakes.

I believe this should fix the flakes described in this comment: https://github.com/containers/podman/issues/7808#issuecomment-700283694

### DIFF
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -25,6 +25,42 @@ func removeConf(confPath string) {
 	}
 }
 
+// generateNetworkConfig generates a cni config with a random name
+// it returns the network name and the filepath
+func generateNetworkConfig(p *PodmanTestIntegration) (string, string) {
+	// generate a random name to preven conflicts with other tests
+	name := "net" + stringid.GenerateNonCryptoID()
+	path := filepath.Join(p.CNIConfigDir, fmt.Sprintf("%s.conflist", name))
+	conf := fmt.Sprintf(`{
+		"cniVersion": "0.3.0",
+		"name": "%s",
+		"plugins": [
+		  {
+			"type": "bridge",
+			"bridge": "cni1",
+			"isGateway": true,
+			"ipMasq": true,
+			"ipam": {
+				"type": "host-local",
+				"subnet": "10.99.0.0/16",
+				"routes": [
+					{ "dst": "0.0.0.0/0" }
+				]
+			}
+		  },
+		  {
+			"type": "portmap",
+			"capabilities": {
+			  "portMappings": true
+			}
+		  }
+		]
+	}`, name)
+	writeConf([]byte(conf), path)
+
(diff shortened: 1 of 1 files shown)


Luap99:
@edsantiago PTAL

rhatdan:
@mheon @baude PTAL

edsantiago:
Nice work. LGTM on first pass, I'll take a closer look in the morning. Thank you for taking this on.

test/e2e/network_test.go
```diff
@@ -25,6 +25,42 @@ func removeConf(confPath string) {
...
+		"name": "%s",
+		"plugins": [
+		  {
+			"type": "bridge",
+			"bridge": "cni1",
+			"isGateway": true,
+			"ipMasq": true,
+			"ipam": {
+				"type": "host-local",
+				"subnet": "10.99.0.0/16",
```
edsantiago, line 45: Thank you for your patience with me. I'm not sure this is going to work, because podman doesn't just trigger on name conflicts, also on subnet conflicts. I don't believe podman (or CNI?) will allow two networks with the same IP range. 

I wonder if you could use `"10.%d.0.0/16", 99 + GinkgoParallelNode()`? This is going to need weigh-in from networking and ginkgo SMEs, I fear.

Luap99: The tests never run this network config. It is only used for network ls/inspect/rm.
That's why we shouldn't run into subnet conflicts. But generell speaking it would definitely  better to generate random subnets to prevent misuse.

edsantiago: Okay... I've convinced myself (via reading) that none of these tests (currently) do anything other than `ls` and `rm`; and furthermore that no other tests in `test/e2e` (currently) use the `10.99/16` subnet; and, finally (via manual testing) that `podman network create` does not (currently) do post-hoc testing of existing `.conflist` files and throw an error if two existing files conflict. Whew! So in short, yes, I believe this is likely to address some of the network flakes we're seeing, without introducing new problems. But would you consider documenting this potentially fragile set of assumptions? Some future day the tests will break again, and someone might notice the same-subnet question; a comment may save them some time. Thank you.

rhatdan:
@Luap99 needs a rebase.

edsantiago:
Errors (3) look like docker.io flakes:
```
Error: unable to pull docker.io/library/busybox:latest: Error initializing source docker://busybox:latest: Requesting bear token: invalid status code from registry 500 (Internal Server Error)
(followed by huge stack trace)
```
Restarted all three with crossed finger.

edsantiago:
/lgtm

Thank you @Luap99 !

PR #7836, opened 2020-09-29
Search repository tags using --list-tags

QiWang19:
For fix of BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1684263
Add --list-tags to podman search to return a table the repository tags.

### DIFF
--- a/cmd/podman/images/search.go
+++ b/cmd/podman/images/search.go
@@ -85,6 +85,7 @@ func searchFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&searchOptions.NoTrunc, "no-trunc", false, "Do not truncate the output")
 	flags.StringVar(&searchOptions.Authfile, "authfile", auth.GetDefaultAuthFile(), "Path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
 	flags.BoolVar(&searchOptions.TLSVerifyCLI, "tls-verify", true, "Require HTTPS and verify certificates when contacting registries")
+	flags.BoolVar(&searchOptions.ListTags, "list-tags", false, "List the tags of the input registry")
 }
 
 // imageSearch implements the command for searching images.
@@ -99,6 +100,10 @@ func imageSearch(cmd *cobra.Command, args []string) error {
 
 	if searchOptions.Limit > 100 {
 		return errors.Errorf("Limit %d is outside the range of [1, 100]", searchOptions.Limit)
+	}
+
+	if searchOptions.ListTags && len(searchOptions.Filters) != 0 {
+		return errors.Errorf("filters are not applicable to list tags result")
 	}
 
 	// TLS verification in c/image is controlled via a `types.OptionalBool`
@@ -119,12 +124,19 @@ func imageSearch(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
 	if len(searchReport) == 0 {
 		return nil
 	}
 
 	hdrs := report.Headers(entities.ImageSearchReport{}, nil)
 	row := "{{.Index}}\t{{.Name}}\t{{.Description}}\t{{.Stars}}\t{{.Official}}\t{{.Automated}}\n"
+	if searchOptions.ListTags {
+		if len(searchOptions.Filters) != 0 {
+			return errors.Errorf("filters are not applicable to list tags result")
+		}
+		row = "{{.Name}}\t{{.Tag}}\n"
+	}
 	if cmd.Flags().Changed("format") {
 		row = report.NormalizeFormat(searchOptions.Format)
 	}
--- a/docs/source/markdown/podman-search.1.md
+++ b/docs/source/markdown/podman-search.1.md
@@ -56,6 +56,9 @@ Valid placeholders for the Go template are listed below:
 | .Stars          | Star count of image          |
 | .Official       | "[OK]" if image is official  |
 | .Automated      | "[OK]" if image is automated |
+| .Tag            | Repository tag               |
+
+Note: use .Tag only if the --list-tags is set.
 
 **--limit**=*limit*
 
@@ -64,6 +67,12 @@ Note: The results from each registry will be limited to this value.
 Example if limit is 10 and two registries are being searched, the total
 number of results will be 20, 10 from each (if there are at least 10 matches in each).
 The order of the search results is the order in which the API endpoint returns the results.
+
+**--list-tags**
+
+List the available tags in the repository for the specified image.
+**Note:** --list-tags requires the search term to be a fully specified image name.
+The result contains the Image name and its tag, one line for every tag associated with the image.
 
 **--no-trunc**
 
@@ -139,6 +148,15 @@ fedoraproject.org   registry.fedoraproject.org/f25/kubernetes-node              
 fedoraproject.org   registry.fedoraproject.org/f25/kubernetes-proxy                              0
 fedoraproject.org   registry.fedoraproject.org/f25/kubernetes-scheduler                          0
 fedoraproject.org   registry.fedoraproject.org/f25/mariadb                                       0
+```
+
+```
+$ podman search --list-tags  registry.redhat.io/rhel
+NAME                      TAG
+registry.redhat.io/rhel   7.3-74
+registry.redhat.io/rhel   7.6-301
+registry.redhat.io/rhel   7.1-9
+...
 ```
 Note: This works only with registries that implement the v2 API. If tried with a v1 registry an error will be returned.
 
--- a/libpod/image/search.go
+++ b/libpod/image/search.go
@@ -2,11 +2,13 @@ package image
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
 	sysreg "github.com/containers/podman/v2/pkg/registries"
 	"github.com/pkg/errors"
@@ -34,6 +36,8 @@ type SearchResult struct {
 	Official string
 	// Automated indicates if the image was created by an automated build.
 	Automated string
+	// Tag is the image tag
+	Tag string
 }
 
 // SearchOptions are used to control the behaviour of SearchImages.
@@ -49,6 +53,8 @@ type SearchOptions struct {
 	Authfile string
 	// InsecureSkipTLSVerify allows to skip TLS verification.
 	InsecureSkipTLSVerify types.OptionalBool
+	// ListTags returns the search result with available tags
+	ListTags bool
 }
 
 // SearchFilter allows filtering the results of SearchImages.
@@ -147,6 +153,15 @@ func searchImageInRegistry(term string, registry string, options SearchOptions) 
 	// every types.SystemContext, and to compute the value just once in one
 	// place.
 	sc.SystemRegistriesConfPath = sysreg.SystemRegistriesConfPath()
+	if options.ListTags {
+		results, err := searchRepositoryTags(registry, term, sc, options)
+		if err != nil {
+			logrus.Errorf("error listing registry tags %q: %v", registry, err)
+			return []SearchResult{}
+		}
+		return results
+	}
+
 	results, err := docker.SearchRegistry(context.TODO(), sc, registry, term, limit)
 	if err != nil {
 		logrus.Errorf("error searching registry %q: %v", registry, err)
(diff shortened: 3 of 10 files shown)


docs/source/markdown/podman-search.1.md
```diff
@@ -56,6 +56,9 @@ Valid placeholders for the Go template are listed below:
 | .Stars          | Star count of image          |
 | .Official       | "[OK]" if image is official  |
 | .Automated      | "[OK]" if image is automated |
+| .Tag            |Repository tag                |
```
TomSweeneyRedHat, line 59:
```suggestion
| .Tag            | Repository tag               |
```

docs/source/markdown/podman-search.1.md
```diff
@@ -140,6 +147,36 @@ fedoraproject.org   registry.fedoraproject.org/f25/kubernetes-proxy
...
+registry.redhat.io/rhel   7.7-348
+registry.redhat.io/rhel   7.7
+registry.redhat.io/rhel   7.4
+registry.redhat.io/rhel   7.5
+registry.redhat.io/rhel   7.5-409.1533127727
+registry.redhat.io/rhel   7.8
+registry.redhat.io/rhel   7.5-404
+registry.redhat.io/rhel   7.4-105
+registry.redhat.io/rhel   7.2-35
+```
```
TomSweeneyRedHat, line 179: Although this shows what gets returned, for the sake of brevity, I think I'd just show 3 or 4 lines and then a new line with `...` or `(multiple lines removed for brevity)`

docs/source/markdown/podman-search.1.md
```diff
@@ -65,6 +68,10 @@ Example if limit is 10 and two registries are being searched, the total
 number of results will be 20, 10 from each (if there are at least 10 matches in each).
 The order of the search results is the order in which the API endpoint returns the results.
 
+**--list-tags**
+
+Search tags available in the repository. Note: the result only contains the Image name and its tag.
```
TomSweeneyRedHat, line 73: As we can use search on `registry.redhat.io/` and get all of the images in the repository:

```suggestion
List the available tags in the repository for each image found. Note: the result contains the Image name and its tag, one line for every tag associated with the image.
```

docs/source/markdown/podman-search.1.md
TomSweeneyRedHat: And I'm going back/forth on this.  Do others have thoughts?

libpod/image/search.go
```diff
@@ -147,6 +153,45 @@ func searchImageInRegistry(term string, registry string, options SearchOptions)
 	// every types.SystemContext, and to compute the value just once in one
 	// place.
 	sc.SystemRegistriesConfPath = sysreg.SystemRegistriesConfPath()
+	if options.ListTags {
+		dockerPrefix := fmt.Sprintf("%s://", docker.Transport.Name())
+		imageRef, err := alltransports.ParseImageName(fmt.Sprintf("%s/%s", registry, term))
+		if err == nil && imageRef.Transport().Name() != docker.Transport.Name() {
+			logrus.Errorf("reference %q must be a docker reference", term)
```
TomSweeneyRedHat, line 160: @rhatdan got a less Docker way to say this?  Perhaps:
```suggestion
			logrus.Errorf("reference %q must be a valid container image reference", term)
```

test/e2e/search_test.go
```diff
@@ -423,4 +423,15 @@ registries = ['{{.Host}}:{{.Port}}']`
...
+	It("podman search repository tags", func() {
+		search := podmanTest.Podman([]string{"search", "--list-tags", "--limit", "30", "docker.io/library/alpine"})
+		search.WaitWithDefaultTimeout()
+		Expect(search.ExitCode()).To(Equal(0))
+		Expect(len(search.OutputToStringArray())).To(Equal(31))
+
+		search = podmanTest.Podman([]string{"search", "--filter=is-official", "--list-tags", "docker.io/library/alpine"})
+		search.WaitWithDefaultTimeout()
+		Expect(search.ExitCode()).To(Not(Equal(0)))
+	})
```
TomSweeneyRedHat, line 436: need to add a search for a registry too with a backslash only, no image name.  For that and for this test, add a test to make sure the returned lines are greater than 1 and perhaps greater than 2 or 3.

docs/source/markdown/podman-search.1.md
rhatdan: Yes let's list 3 and then ...

libpod/image/search.go
```diff
@@ -147,6 +153,45 @@ func searchImageInRegistry(term string, registry string, options SearchOptions)
 	// every types.SystemContext, and to compute the value just once in one
 	// place.
 	sc.SystemRegistriesConfPath = sysreg.SystemRegistriesConfPath()
+	if options.ListTags {
```
vrothberg, line 156: Can we break this out into a new function?

libpod/image/search.go
vrothberg: I think "docker" is more accurate in containers-transports(5) terms as all of the transports refer to a container image.

pkg/api/handlers/libpod/images.go
```diff
@@ -591,6 +591,7 @@ func SearchImages(w http.ResponseWriter, r *http.Request) {
 		NoTrunc   bool     `json:"noTrunc"`
 		Filters   []string `json:"filters"`
 		TLSVerify bool     `json:"tlsVerify"`
+		ListTags  bool     `json:"listTags"`
```
vrothberg, line 594: We also need to extend the swagger docs in `pkg/api/server/register_images.go`

libpod/image/search.go
TomSweeneyRedHat: @vrothberg sometimes I reall don't like it when you're right...  Let's leave it then, thx for the follow up.

QiWang19:

cmd/podman/images/search.go
```diff
@@ -86,6 +86,7 @@ func searchFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&searchOptions.NoTrunc, "no-trunc", false, "Do not truncate the output")
 	flags.StringVar(&searchOptions.Authfile, "authfile", auth.GetDefaultAuthFile(), "Path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
 	flags.BoolVar(&searchOptions.TLSVerifyCLI, "tls-verify", true, "Require HTTPS and verify certificates when contacting registries")
+	flags.BoolVar(&searchOptions.ListTags, "list-tags", false, "Search the tags of the input registry")
```
rhatdan, line 89: List the tags ...

cmd/podman/images/search.go
```diff
@@ -126,6 +127,19 @@ func imageSearch(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if searchOptions.ListTags {
+		if len(searchOptions.Filters) != 0 {
+			return errors.Errorf("filters are not applicable to search tags result")
```
rhatdan, line 132: applicable to `list` tags result

libpod/image/search.go
```diff
@@ -207,6 +217,45 @@ func searchImageInRegistry(term string, registry string, options SearchOptions)
 	return paramsArr
 }
 
+func searchRepositoryTags(registry, term string, sc *types.SystemContext, options SearchOptions) []SearchResult {
```
rhatdan, line 220: This should return real errors, and not just write them to the screen.

libpod/image/search.go
```diff
@@ -207,6 +217,45 @@ func searchImageInRegistry(term string, registry string, options SearchOptions)
...
+	dockerPrefix := fmt.Sprintf("%s://", docker.Transport.Name())
+	imageRef, err := alltransports.ParseImageName(fmt.Sprintf("%s/%s", registry, term))
+	if err == nil && imageRef.Transport().Name() != docker.Transport.Name() {
+		logrus.Errorf("reference %q must be a docker reference", term)
```
rhatdan, line 224: Return this error

QiWang19: fixed

rhatdan:
LGTM

rhatdan:

TomSweeneyRedHat:
Looks like there are still comments from @vrothberg and me that have not yet been addressed....

QiWang19:
> Looks like there are still comments from @vrothberg and me that have not yet been addressed....

fixed that in pkg/api/server/register_images.go

libpod/image/search.go
TomSweeneyRedHat: @QiWang19 this was the @vrothberg breakout.  So the change he was asking for was what you did with `searchRepositoryTags()` function below?  If so, I wonder why GitHub didn't not that.  Maybe because this line didn't change?

test/e2e/search_test.go
TomSweeneyRedHat: I see you added the test with lines >2, ty!  Could you also add a test that lists the tags for `docker.io/library/` (no image, just a backslash).  According to the doc, it's allowed:  `podman search registry.fedoraproject.org/` and if it is NOT allowed with the `--list-tags` option, it would be good to test that.

QiWang19: Test added. It's not allowed, no tags will be returned for `docker.io/library/`

TomSweeneyRedHat: I'm fine with not allowing it, but we really should add a note to the `--list-tags` option in the man page saying that, and preferably raise some kind of error message telling the user that they need to specify an image or at least return an "image not found" kind of message.

TomSweeneyRedHat: Hope I'm not being too painful @QiWang19 !

QiWang19: the error would be "library/" must be a docker reference". Add to the doc "the search term should be repository name". Would this work?
$ bin/podman search --list-tags docker.io/library/
ERRO[0000] error listing registry tags "docker.io": reference "library/" must be a docker reference

TomSweeneyRedHat: OK, first off, I'd still like to see a test for `docker.io/library/` (or some other registry) and then test for the appropriate error message.  I'm finding the man page to be very confusing at the moment, will try to suggest a rewrite.

docs/source/markdown/podman-search.1.md
```diff
@@ -65,6 +68,11 @@ Example if limit is 10 and two registries are being searched, the total
 number of results will be 20, 10 from each (if there are at least 10 matches in each).
 The order of the search results is the order in which the API endpoint returns the results.
 
+**--list-tags**
+
+List the available tags in the repository for each image found. Note: --list-tags requires the search term to be a full
+repository name. If not specified the registry server, default registries will be searched through (in /etc/containers/registries.conf).  The result contains the Image name and its tag, one line for every tag associated with the image.
+
```
TomSweeneyRedHat, line 75: This is all very contradictory.  You're saying at the start that you'll list all tags for each image found in the repository, and then that the full name must be used.  I think from our conversations that the image name with the repository must be supplied.  I.e.  you must do `docker.io/library/alpine` and you can't do `docker.io/library/` or `alpine`.  Given that assumption:
```suggestion
List the available tags in the repository for the specified image. **Note:** --list-tags requires the search term to be a fully
specified image name. The result contains the Image name and its tag, one line for every tag associated with the image.
```

QiWang19: fixed

test/e2e/search_test.go
QiWang19: search does not return error, it logs errors to standard logger and returns the 0 entry of the result.

docs/source/markdown/podman-search.1.md
QiWang19: I think the docker.io/library/alpine refer to a repository name, if append a tag to it, it will refer to an image?

TomSweeneyRedHat: @QiWang19 yeah, it's confusing and I may not have helped matters.  When you push a container image to a registry, it creates a repository from it.  How's this?

```
List the available tags in the repository for the specified image. **Note:** --list-tags requires the search term to be a fully
specified image. The result contains the Image name and its tag, one line for every tag found in the repository.
```

test/e2e/search_test.go
TomSweeneyRedHat: Sounds good and it looks like you've added a test for that, TY!

docs/source/markdown/podman-search.1.md
TomSweeneyRedHat: :+1: 
nicely done @QiWang19

TomSweeneyRedHat (commented):
LGTM

PR #7840, opened 2020-09-30
fix remote untag

vrothberg:
Fix the remote client to untag all tags of the specified image.
Instead of querying the image on the client side, support the
case where both, repo and tag, are empty and remove all tags.

Reuse the ABI implementation where possible.  In retrospective,
the libpod untag endpoint should support a slice of strings to
batch remove tags rather than reaching out for each tag individually.

Enable the skipped test.

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -560,24 +560,41 @@ func CommitContainer(w http.ResponseWriter, r *http.Request) {
 func UntagImage(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 
-	name := utils.GetName(r)
-	newImage, err := runtime.ImageRuntime().NewFromLocal(name)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


vrothberg:
@rhatdan, that's an alternative approach which I prefer as it's getting the remote client closer to the local one. I wished the untag endpoint supported a slice of strings so we could have a 1:1 mapping to the local one.

@baude @jwhonce PTAL

rhatdan:
I actually just finished something very similar, although I like yours better with using the ABI untag in the end.

rhatdan:
LGTM

giuseppe (approved):
LGTM

vrothberg:
@edsantiago, let me you know if you want me to extend the system tests as well (I can open a PR after). I just checked and it seems we're not executing `untag $image` without additional tag args to remote all tags.

edsantiago:
>extend the system tests as well

@vrothberg what kind of question is that????? :wink: It sounds like writing system tests isn't as approachable as I had hoped, so please reach out and let me know what stumbling blocks you encounter.

PS the flake you restarted is #7808. I'm going to have to disable that test because it seems to be hitting us very hard in CI.

PR #7825, opened 2020-09-29
Fix handling of remove of bogus volumes, networks and Pods

rhatdan:
In podman containers rm and podman images rm, the commands
exit with error code 1 if the object does not exists.

This PR implements similar functionality to volumes, networks, and Pods.

Similarly if volumes or Networks are in use by other containers, and return
exit code 2.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #7847, opened 2020-09-30
Make the e2e test network cleanup more robust.

### LINKED ISSUE #7809: test/e2e : podman network remove with two networks: needs fail-safe cleanup

rhatdan:
We need to handle removal of non existing network.

This allows the `removeCNINetwork` function
always to be called. This is needed by tests which are trying
to remove the network manually in order to prevent flakes.

Fixes #7809

### DIFF
--- a/test/e2e/network_create_test.go
+++ b/test/e2e/network_create_test.go
@@ -58,7 +58,7 @@ func genericPluginsToPortMap(plugins interface{}, pluginType string) (network.Po
 func (p *PodmanTestIntegration) removeCNINetwork(name string) {
 	session := p.Podman([]string{"network", "rm", "-f", name})
 	session.WaitWithDefaultTimeout()
-	Expect(session.ExitCode()).To(BeZero())
+	Expect(session.ExitCode()).To(BeNumerically("<=", 1))
 }
 
 func removeNetworkDevice(name string) {
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -135,7 +135,6 @@ var _ = Describe("Podman network", func() {
 	})
 
 	It("podman network rm", func() {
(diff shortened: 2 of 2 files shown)


rhatdan:
Replaces: https://github.com/containers/podman/pull/7812
@Luap99 @edsantiago PTAL

test/e2e/network_create_test.go
```diff
@@ -58,7 +58,7 @@ func genericPluginsToPortMap(plugins interface{}, pluginType string) (network.Po
 func (p *PodmanTestIntegration) removeCNINetwork(name string) {
 	session := p.Podman([]string{"network", "rm", "-f", name})
 	session.WaitWithDefaultTimeout()
-	Expect(session.ExitCode()).To(BeZero())
+	Expect(session.ExitCode() < 2).To(BeTrue())
```
edsantiago, line 61: I'm not a ginkgo expert nor fan, but would you consider `Expect(session.ExitCode()).To(BeNumerically("<=",1))` ? I believe that will give better diagnostics in case of error. (Those `false was not true` diagnostics always make me struggle to try to find out what happened).

rhatdan: Sure, I was trying to discover the correct syntax for this.

edsantiago: I just remembered seeing `Numerically` in the past, then grep'ed for it in `test/e2e`.

edsantiago:
LGTM. Thanks for your cleanup work in #7825

rhatdan:

vrothberg (commented):
/lgtm
/hold

PR #7848, opened 2020-09-30
Fix two e2e tests

cevich:
These tests are failing due to slightly wider coverage in #6784 but otherwise should be fixed anyway.

### DIFF
--- a/test/e2e/common_test.go
+++ b/test/e2e/common_test.go
@@ -599,19 +599,21 @@ func (p *PodmanTestIntegration) CreateSeccompJson(in []byte) (string, error) {
 	return jsonFile, nil
 }
 
-func SkipIfRootlessCgroupsV1(reason string) {
+func checkReason(reason string) {
 	if len(reason) < 5 {
-		panic("SkipIfRootlessCgroupsV1 must specify a reason to skip")
+		panic("Test must specify a reason to skip")
 	}
+}
+
+func SkipIfRootlessCgroupsV1(reason string) {
+	checkReason(reason)
 	if os.Geteuid() != 0 && !CGROUPSV2 {
 		Skip("[rootless]: " + reason)
 	}
 }
 
 func SkipIfRootless(reason string) {
-	if len(reason) < 5 {
-		panic("SkipIfRootless must specify a reason to skip")
-	}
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

rhatdan:

test/e2e/exec_test.go
```diff
@@ -286,7 +286,7 @@ var _ = Describe("Podman exec", func() {
 	It("podman exec preserves container groups with --user and --group-add", func() {
 		SkipIfRemote("FIXME: This is broken SECCOMP Failues?")
 
-		dockerfile := `FROM fedora-minimal
+		dockerfile := `FROM registry.fedoraproject.org/fedora-minimal
```
TomSweeneyRedHat, line 289: I've run into troubles with this repo in the past, will keep my fingers crossed.

TomSweeneyRedHat:
LGTM

test/e2e/common_test.go
```diff
@@ -629,23 +631,33 @@ func isRootless() bool {
...
-		panic("SkipIfCgroupV2 must specify a reason to skip")
-	}
+	checkReason(reason)
 	if CGROUPSV2 {
 		Skip(reason)
 	}
 }
 
+func isContainerized() bool {
+	if os.Getenv("container") == "podman" {
```
nalind, line 648: Is this the only value of "container" for which the test calling this function should be skipped?

ashley-cui:
LGTM

test/e2e/common_test.go
rhatdan: Well we are running within Podman, but making this just check `if os.Getenv('container") != "" ` is probably better.

test/e2e/exec_test.go
cevich: This is (hopefully) only temporary until Brent can get local caching working.

test/e2e/common_test.go
cevich: I believe so.  Though perhaps a `os.Getenv("container") != ""` would be safer.  I'll fix it, thanks.

vrothberg (commented):
LGTM

PR #7834, opened 2020-09-29
Don't disable Go modules when generating varlink

xordspar0:
From a fresh install of Fedora 33 Beta and a fresh clone of the repo, `make` fails with the following error when Go modules are disabled:

    # Only generate the varlink code on Linux (see issue #4814).
    GO111MODULE=off go generate ./pkg/varlink/...
    ../../vendor/github.com/varlink/go/cmd/varlink-go-interface-generator/main.go:12:2: cannot find package "github.com/varlink/go/varlink/idl" in any of:
    	/usr/lib/golang/src/github.com/varlink/go/varlink/idl (from $GOROOT)
    	/home/test/src/podman/_output/src/github.com/varlink/go/varlink/idl (from $GOPATH)
    pkg/varlink/generate.go:3: running "go": exit status 1
    make: *** [Makefile:646: pkg/varlink/iopodman.go] Error 1

PR #7854, opened 2020-09-30
Makefile: Fix condition for varlink_generate

rhafer:
This should fix the call to the `or` function. Previously the condition
always evaluated to the empty string independed of the BULILDTAGS.
So `go generate varlink` was allways called.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -620,7 +620,7 @@ endef
 
 # $BUILD_TAGS variable is used in hack/golangci-lint.sh
 .PHONY: varlink_generate
-ifneq (or $(findstring varlink,$(BUILDTAGS)),$(findstring varlink,$(BUILD_TAGS)))
+ifneq (,$(or $(findstring varlink,$(BUILDTAGS)),$(findstring varlink,$(BUILD_TAGS))))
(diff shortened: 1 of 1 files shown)


rhafer:
Hm, I guess this needs https://github.com/containers/podman/pull/7834 to avoid the failure in the gating job

Makefile
```diff
@@ -620,7 +620,7 @@ endef
 
 # $BUILD_TAGS variable is used in hack/golangci-lint.sh
 .PHONY: varlink_generate
-ifneq (or $(findstring varlink,$(BUILDTAGS)),$(findstring varlink,$(BUILD_TAGS)))
+ifneq (,$(or $(findstring varlink,$(BUILDTAGS)),$(findstring varlink,$(BUILD_TAGS))))
```
rhafer, line 623: I am still confused about how this is supposed to work for the `make golangci-lint` target. (See the gating failure in https://cirrus-ci.com/task/5884302384693248). It's true that `hack/golangci-lint.sh` sets the `BUILD_TAGS` variable, but that happens after this part here is executed. So it will never have an effect here.

I guess this only ever worked because the condition was broken?

rhatdan:
Since we are dropping support for varlink soon, I am going to close this PR.

PR #7856, opened 2020-09-30
podman-remote does not support most of the global flags

rhatdan:
podman-remote --help is showing a bunch of global flags that it
does not support

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -154,34 +154,35 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	if cmd.Flag("cpu-profile").Changed {
-		f, err := os.Create(cfg.CPUProfile)
-		if err != nil {
-			return errors.Wrapf(err, "unable to create cpu profiling file %s",
-				cfg.CPUProfile)
+	if !registry.IsRemote() {
+		if cmd.Flag("cpu-profile").Changed {
+			f, err := os.Create(cfg.CPUProfile)
+			if err != nil {
+				return errors.Wrapf(err, "unable to create cpu profiling file %s",
+					cfg.CPUProfile)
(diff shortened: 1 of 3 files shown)


rhatdan:
@edsantiago Your man page checker missed these global flags for podman-remote man page.

edsantiago:
Man page checker does not test podman-remote.md. That may be a holdover from the varlink days; I'll take a look (tomorrow) and see what it'd take to enable.

jwhonce:
@rhatdan Your fix is the correct method, you just need to filter through the sub-commands that are trying to now unnecessarily hide those flags.

vrothberg (commented):
While we're at it, I think that the man pages should be updated as well. A comment indicating that these options are not supported on the remote client may be enough.

rhatdan:
Podman remote man page does not refer to these options.
@vrothberg you want me to change the flags to say that each global option does not effect the remote server on the regular podman.1.md?

vrothberg:
> Podman remote man page does not refer to these options.
> @vrothberg you want me to change the flags to say that each global option does not effect the remote server on the regular podman.1.md?

I didn't check `podman-remote`. I think we're good as the options are not mentioned there.

edsantiago:
Update: without a completely different set of man pages, or at least a robust well-established convention for marking options/subcommands in man pages as remote-only, there's little hope for `hack/xref-helpmsgs-manpages` to check `podman-remote`. Even with a simple exclusion of `-l/--latest` I see 90 warnings, in both options (`run --authfile`, `rm -i`) and subcommands (`system migrate/renumber/reset`, `mount/umount/unshare`).

rhatdan:

QiWang19:
LGTM

ashley-cui (commented):
LGTM

PR #7864, opened 2020-10-01
Volume prune should not pass down the force flag

rhatdan:
podman volume prune -f

Should just tell the prune command to not prompt for confirmation.
It should not be passing the prune flag into the API.


rhatdan:
@vrothberg Would this be an API breaking change?

vrothberg (commented):
LGTM

We're only changing an internal API which is OK.

mheon:
LGTM

Issue #7837, opened 2020-09-29
remote: flake: podman rm (more than 1 container): no such container


edsantiago:
Seen this one at least twice in CI. Here's a reproducer, albeit not a great one:
```console
# cat testme
#!/bin/bash

set -ex

while :;do
    podman-remote run    -v /myvol1 --name c1 alpine true
    podman-remote run -d -v /myvol2 --name c2 alpine true
    podman-remote wait c2
    podman-remote rm c1 c2
done

# T0=$SECONDS;./testme;T1=$SECONDS;echo;echo $((T1-T0))
...
+ podman-remote run -v /myvol1 --name c1 alpine true
+ podman-remote run -d -v /myvol2 --name c2 alpine true
c1e9e53e217d2c081044f63de0fe4f41a9537072680d6b1b14cbb344e9590711
+ podman-remote rm c1 c2
c1e9e53e217d2c081044f63de0fe4f41a9537072680d6b1b14cbb344e9590711
Error: no container with name or ID c1e9e53e217d2c081044f63de0fe4f41a9537072680d6b1b14cbb344e9590711 found: no such container
```
Will usually fail within 5 minutes, but once it took 20.  The `wait` is not actually necessary, but for some reason it makes the script fail more quickly. I'm pretty sure the `-v` is necessary, I couldn't get failures without it (but maybe I just didn't wait long enough).

CI failures: [today](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6699446727606272/html/system_test.log.html), and [Aug 28](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6471035019067392/html/system_test.log.html)

vrothberg:
Thanks a ton for the reproducer, @edsantiago!

I'll take a look. This one has been quite annoying.

vrothberg:
I can reproduce and after an audit of the code I suspected that the issue is not on the server-side removal but somehow corrupted data on the client. So I started logging which containers the client wants to remove:
```
ERRO[0000] *** removing "c1c9a6f8cf12ce6252064aba9371c72ba9bb044b5b9e997cbfc87f4904f71875"
ERRO[0000] *** removing "c1c9a6f8cf12ce6252064aba9371c72ba9bb044b5b9e997cbfc87f4904f71875"
```
... and indeed the client tries to remove the same container twice although it should be two different IDs (for the two containers).

Digging further.

[...]

PR #7867, opened 2020-10-01
remote: fix name and ID collisions of containers and pods

vrothberg:
Fix the look up of containers and pods in the remote client.  User input
can refer to both, names or IDs of containers and pods, so there is a fair
chance of collisions (e.g., "c1" name with a "c1...." ID).

Those collisions are well handled (and battle tested) in the local
client which is directly using the libpod backend.  Hence, the remote
client should not attempt to introduce its own logic to prevent bugs and
divergence between the local and the remote clients.  To prevent
collisions such as in #7837, do a container/pod inspect on the
user-provided input to find the corresponding ID and eventually do full
ID comparisons to avoid potential collisions with names.

Note that this has a cost that I am not entirely happy with.  Looking at
issue #7837, the collisions are happening when removing the two
containers.  Remote container removal is now very chatty with the server
as it first queries for all containers, then iterates over the provided
names or IDs and does a remote inspect to figure out the IDs and find a
matching container object.  However, remote removal could just pass the
names and IDs directly to the batch removal endpoint.  Querying for all
containers could be prevented if the batch removal endpoint would remove
all if the slice is empty.

In other words, the bug is fixed but there's room for performance
improvements.

Fixes: #7837

### DIFF
--- a/pkg/domain/infra/tunnel/helpers.go
+++ b/pkg/domain/infra/tunnel/helpers.go
@@ -2,74 +2,118 @@ package tunnel
 
 import (
 	"context"
-	"strings"
 
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/bindings/containers"
 	"github.com/containers/podman/v2/pkg/bindings/pods"
 	"github.com/containers/podman/v2/pkg/domain/entities"
-	"github.com/containers/podman/v2/pkg/util"
+	"github.com/containers/podman/v2/pkg/errorhandling"
 	"github.com/pkg/errors"
 )
 
+// FIXME: the `ignore` parameter is very likely wrong here as it should rather
+//        be used on *errors* from operations such as remove.
 func getContainersByContext(contextWithConnection context.Context, all, ignore bool, namesOrIDs []string) ([]entities.ListContainer, error) {
-	var (
-		cons []entities.ListContainer
-	)
 	if all && len(namesOrIDs) > 0 {
 		return nil, errors.New("cannot lookup containers and all")
 	}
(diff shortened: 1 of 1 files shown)


vrothberg:
@containers/podman-maintainers PTAL ... something to look for at other places as well  (e.g., volumes, networks, etc.).

pkg/domain/infra/tunnel/helpers.go
```diff
@@ -2,74 +2,114 @@ package tunnel
...
+				filtered = append(filtered, pod)
 				found = true
 				break
 			}
+
 		}
+
 		if !found {
-			return nil, errors.Wrapf(define.ErrNoSuchPod, "unable to find pod %q", nameOrID)
+			return nil, errors.Wrapf(define.ErrNoSuchCtr, "unable to find pod %q", nameOrID)
```
vrothberg, line 111: Copy error ... should be Pod :)

vrothberg:
Maybe, container removal needs a similar batch-removal endpoint as we have for images?

mheon:
I don't think fixing this by adding a batch endpoint is a good idea. This
code is used in dozens of places and we'd need new endpoints for each. I
think we may want to investigate improving the list endpoint so we only
need to make one call - adding a nameOrId filter maybe?

mheon:
This change LGTM though. We can optimize later.

vrothberg:
> I don't think fixing this by adding a batch endpoint is a good idea. This code is used in dozens of places and we'd need new endpoints for each. I think we may want to investigate improving the list endpoint so we only need to make one call - adding a nameOrId filter maybe?

Sounds good to me. I had the same idea  to extend the list endpoint (see code comments).

vrothberg:
Failures look legit, will have a look

vrothberg:
Updated, pod tests should pass now. Had to add some `errorhandlig.Containes` checks.

vrothberg:
@baude @jwhonce @rhatdan PTAL

Issue #7782, opened 2020-09-25
Groups do not seem to be getting assigned by default in rootless mode versus rootful mode.


rhatdan:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

$ podman (pullpolicy) $ sudo podman run --rm alpine id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
$ podman (pullpolicy) $ podman run --rm alpine id
uid=0(root) gid=0(root)

rhatdan:
@giuseppe @mheon @vrothberg Ideas?

giuseppe:
we are not setting additional gids for rootless: https://github.com/containers/podman/blob/master/libpod/container_internal_linux.go#L417-L422

I think the reason is that it won't work when there are no multiple gids available, or not enough to cover all the additional gids

rhatdan:
You are saying that if the range of GIDs  is not mapped into the user namespace?  If so should we make this smarter and allow it to check the range of UIDs and make sure they will work and not blow up the container, then warn about missing gids.

PR #7871, opened 2020-10-01
Add additionalGIDs from users in rootless mode

rhatdan:
There is a risk here, that if the GID does not exists
within the User Namespace the container will fail to start.

This is only likely to happen in HPC Envioronments, and I think
we should add a field to disable it for this environment,
Added a FIXME for this issue.

We currently have this problem with running a rootfull container within
a user namespace, it will fail if the GID is not available.

I looked at potentially checking the usernamespace that you are assigned
to, but I believe this will be very difficult to code up and to figure out.

Fixes https://github.com/containers/podman/issues/7782

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -415,8 +415,9 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 	}
 
 	// Look up and add groups the user belongs to, if a group wasn't directly specified
-	if !rootless.IsRootless() && !strings.Contains(c.config.User, ":") {
+	if !strings.Contains(c.config.User, ":") {
 		for _, gid := range execUser.Sgids {
+			// FIXME: We need to add a flag to containers.conf to not add these for HPC Users.
 			g.AddProcessAdditionalGid(uint32(gid))
 		}
 	}


rhatdan:
@giuseppe PTAL

giuseppe:
I had started looked at it, but haven't finished the patch yet.

Basically, we could reuse `GetAvailableGids()` from `pkg/specgen/generate/oci.go` to know how many gids are available.  At least we can filter out gids that are outside of the `0 to GetAvailableGids()` range

rhatdan:
@giuseppe if you want to take this PR over, your welcome to it.

rhatdan:
Lets get this in and then have @giuseppe add an update when he is ready.

QiWang19:
LGTM

mheon:
LGTM as long as we don't cut a release before this is fixed.

TomSweeneyRedHat:
If we have more to do, can we add a TODO comment please?

rhatdan:
There already is a FIXME in the code.

TomSweeneyRedHat:
@rhatdan , yeah there's a "FIXME", but not a "TODO".  Dope slap here, I'll get myself before you get me.  I'll add a /lgtm, but will do a /hold in case @mheon wants to wait on this by chance.

giuseppe:
follow-up: https://github.com/containers/podman/pull/7882

PR #7873, opened 2020-10-01
logformatter: add Synopsis at top of each page

edsantiago:
At the top of each generated page, add a Synopsis table with:

      PR number/name, and link to github
      Author name(s)
      Test name (fedora/ubuntu, rootless, etc)
      Cirrus task ID, and link to Cirrus
      The value of $SPECIALMODE

This is all we can get from the Cirrus environment in
which logformatter runs; we can't get things like
cgroup manager or username that the test runs under.

Note that the table is at the top, which is usually
invisible because we autoscroll to the bottom on
page load. I tentatively think that top is a more
natural place for this info than bottom, but am
willing to listen to arguments against.

Also, one minor tweak: highlight podman commands in
the BATS output. The idea is to make it easier for the eye
to spot those, then copy/paste them to find a reproducer.

### DIFF
--- a/contrib/cirrus/logformatter
+++ b/contrib/cirrus/logformatter
@@ -31,6 +31,12 @@ our $CSS = <<'END_CSS';
 /* wrap long lines - don't require user to scroll right */
 pre        { line-break: normal; overflow-wrap: normal; white-space: pre-wrap; }
 
+/* synopsis table at top */
+table.synopsis { border: none; border-collapse: collapse; margin-left: 2em; margin-top: 2ex; }
+.synopsis th   { font-weight: normal; font-size: 110%; text-align: right; }
+.synopsis td   { font-weight: bold;   font-size: 120%; font-family: monospace; }
+
+/* test results */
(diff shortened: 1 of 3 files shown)


edsantiago:
Here are [sample results](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6243910802472960/html/system_test.log.html#top). I don't know why `Author` didn't show up -- it did in a [previous iteration](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6049612588908544/html/system_test.log.html#top) (this latter link also shows the effect of the podman command highlighting on error output).

I'm not delighted with it, and can already think of a few ways to make it more readable, but I need to head out for the whole day. And this is going to need to change anyway once #6784 goes in. Feel free to merge if it looks good, otherwise it can wait til next week.

rhatdan:
LGTM

QiWang19:
LGTM

PR #7874, opened 2020-10-01
Podman containers/pods prune should throw an error if user adds args

rhatdan:
We are not currently checking if a user accidently adds an argument to
all podman * prune commands.

### DIFF
--- a/cmd/podman/containers/prune.go
+++ b/cmd/podman/containers/prune.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/containers/podman/v2/cmd/podman/registry"
 	"github.com/containers/podman/v2/cmd/podman/utils"
+	"github.com/containers/podman/v2/cmd/podman/validate"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


edsantiago:
Wait, what?? I have an explicit test for this. And I just tried:
```console
$ ./bin/podman container prune sdfsdf
Error: `podman container prune` takes no arguments
$ ./bin/podman pod prune sdfsdf
Error: `podman pod prune` takes no arguments
```
Can you give an example of what this fixes, and what my tests missed?

Luap99:
https://github.com/containers/podman/blob/5d22eb02f95f28a87ed263afe28b7ff4bf2f6fee/cmd/podman/containers/prune.go#L48-L50
https://github.com/containers/podman/blob/5d22eb02f95f28a87ed263afe28b7ff4bf2f6fee/cmd/podman/pods/prune.go#L44-L46
It looks like we do the validation ourself in the run function.

The changes LGTM if you remove those lines.

rhatdan:
Thanks @Luap99 
@edsantiago I was just looking at the code of prunes because we had an issue that claimed `podman volume prune` did not do this, after I checked, I figured I would check the other prunes, and came upon these and figured I would throw up a PR so they all did it the same way.
I missed the fact that we were checking in the code.  And have now removed the code checks.

edsantiago:
Thanks @rhatdan and @Luap99 , I'm (hugely) relieved that my tests didn't miss something. This PR is good cleanup, a more natural way to catch the error. LGTM.

TomSweeneyRedHat:
LGTM

PR #7882, opened 2020-10-02
libpod: check there are enough gids before adding them

giuseppe:
when running rootless, check there are enough gids in the current user namespace before adding supplementary gids from /etc/group.

Follow-up for baede7cd2776b1f722dcbb65cff6228eeab5db44

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
 	"net"
 	"os"
 	"os/user"
@@ -35,6 +36,7 @@ import (
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/podman/v2/utils"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/idtools"
 	securejoin "github.com/cyphar/filepath-securejoin"
 	runcuser "github.com/opencontainers/runc/libcontainer/user"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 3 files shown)


libpod/container_internal_linux.go
```diff
@@ -416,9 +417,25 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
 	// Look up and add groups the user belongs to, if a group wasn't directly specified
 	if !strings.Contains(c.config.User, ":") {
-		for _, gid := range execUser.Sgids {
-			// FIXME: We need to add a flag to containers.conf to not add these for HPC Users.
-			g.AddProcessAdditionalGid(uint32(gid))
+		var haveEnoughGids bool
+
+		if !rootless.IsRootless() {
+			// Running as root, assume there are enough GIDs.
+			haveEnoughGids = true
```
rhatdan, line 424: This seems to be still error prone.  I can run a root container with a user namespace with only 100 GIDs?
podman run --usermap=0:100000:100 ...

libpod/container_internal_linux.go
```diff
@@ -416,9 +417,25 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
+		} else {
+			// Check whether the current user namespace has enough gids available.
+			availableGids, err := rootless.GetAvailableGids()
+			if err != nil {
+				return nil, errors.Wrapf(err, "cannot read number of available GIDs")
+			}
+			// Do not check what values, as the user can specify an arbitrary user namespace
+			// to accomodate the requested gids.
+			haveEnoughGids = availableGids > int64(len(execUser.Sgids))
```
rhatdan, line 433: I don't understand this?  
Don't we want to check each GID in the execUser.Sgids > avialbaleGids?

libpod/container_internal_linux.go
```diff
@@ -416,9 +417,25 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
+		}
+		if haveEnoughGids {
+			for _, gid := range execUser.Sgids {
```
rhatdan, line 436: We should add a logrus.Warnf whenever we hit a situation without enough GIDs.

libpod/container_internal_linux.go
giuseppe: I tried to avoid looking into the user namespace mapping, but I agree the current check doesn't make much sense.

I've changed it to look at each gid and see if it is present in the namespace (current or configured for the container)

rhatdan:
LGTM
Except you are not passing gating.

giuseppe:
tests are green

rhatdan:

QiWang19:
LGTM

PR #7891, opened 2020-10-02
This PR allows users to remove external containers directly

rhatdan:
Currenly if a user specifies the name or ID of an external storage
container, we report an error to them.

buildah from scratch
working-container-2
podman rm working-container-2
Error: no container with name or ID working-container-2 found: no such container

Since the user specified the correct name and the container is in storage we
force them to specify --storage to remove it. This is a bad experience for the
user.

This change will just remove the container from storage.  If the container
is known by libpod, it will remove the container from libpod as well.

If the user species the --storage option, and the container is known to libpod
the command will fail.

Also cleaned documented options that are not available to podman-remote.

### DIFF
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -273,16 +273,6 @@ func (ic *ContainerEngine) ContainerRestart(ctx context.Context, namesOrIds []st
 
 func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string, options entities.RmOptions) ([]*entities.RmReport, error) {
 	reports := []*entities.RmReport{}
-	if options.Storage {
-		for _, ctr := range namesOrIds {
-			report := entities.RmReport{Id: ctr}
-			if err := ic.Libpod.RemoveStorageContainer(ctr, options.Force); err != nil {
-				report.Err = err
-			}
-			reports = append(reports, &report)
-		}
-		return reports, nil
-	}
 
 	names := namesOrIds
 	for _, cidFile := range options.CIDFiles {
@@ -294,6 +284,22 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
 		names = append(names, id)
 	}
 
+	// Attempt to remove named containers directly from storage, if container is defined in libpod
+	// this will fail and code will fall through to removing the container from libpod.`
+	tmpNames := []string{}
+	for _, ctr := range names {
+		report := entities.RmReport{Id: ctr}
+		if err := ic.Libpod.RemoveStorageContainer(ctr, options.Force); err != nil {
+			// remove container names that we successfully deleted
+			tmpNames = append(tmpNames, ctr)
+		} else {
+			reports = append(reports, &report)
+		}
+	}
+	if len(tmpNames) < len(names) {
+		names = tmpNames
+	}
+
 	ctrs, err := getContainersByContext(options.All, options.Latest, names, ic.Libpod)
 	if err != nil && !(options.Ignore && errors.Cause(err) == define.ErrNoSuchCtr) {
 		// Failed to get containers. If force is specified, get the containers ID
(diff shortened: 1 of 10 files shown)


pkg/domain/infra/abi/containers.go
```diff
@@ -294,6 +284,27 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
 		names = append(names, id)
 	}
 
+	// Attempt to remove named containers directly from storage, if container is defined in libpod
+	// this will fail, if options.Storage is not specified, code will fall through to removing
+	// the container from libpod.`
+	i := 0
```
mheon, line 290: Please just make a new array - the memory savings/speed optimization here is going to be minor and these is a lot less readable than appending failed names to a fresh array.

pkg/domain/infra/abi/containers.go
```diff
@@ -294,6 +284,27 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
...
+	for _, ctr := range names {
+		report := entities.RmReport{Id: ctr}
+		if err := ic.Libpod.RemoveStorageContainer(ctr, options.Force); err != nil {
+			// remove container names that we successfully deleted
+			names[i] = ctr
+			i++
+			report.Err = err
+			if options.Storage {
+				reports = append(reports, &report)
+				return reports, nil
```
mheon, line 300: This doesn't seem correct - the old `--storage` behavior was to try all images, not fail when only one failed.

pkg/domain/infra/abi/containers.go
rhatdan: Fixed

pkg/domain/infra/abi/containers.go
rhatdan: No it was not, the code, would attempt to remove the nameOrIDs from containerStorage.  But now it will do it if the container exists in storage and not in libpod.  Making the storage flag useless.  I am going to just hide it and remove it from docs.

mheon: What I am saying is that the new code will return an error after *any single image* fails. The old code tried *all* images, even if one failed.

rhatdan: I think we just remove the storage option altogether, if the image does not exists in libpod, it will report and error now, and it will be removed if it is just in container/storage.

TomSweeneyRedHat:
I'm a bit confused by the change.  I thought we wanted to not allow this so that someone would have to think before removing a storage container....  I could see changing the error message to be more explanatory in that regard, but are we all comfortable with just removing a container that wasn't created by Podman?  I'm going to think some people will no like that and I'm not sure which is the lesser evil if you will.

rhatdan:
@TomSweeneyRedHat The only way to remove a container would be to fully specify the name.  This is  not going to happen via --latest or --all.  You basically would have to find the name of the container via podman ps --storage or an error message perhaps in podman stop.  

Then you need to do podman stop UUID. The chance of someone doing this by mistake is nil.

Currently the UI does

```
# buildah from alpine
alpine-working-container
# podman rm alpine-working-container
Error: no container with name or ID alpine-working-container found: no such container
```

Which is  a lie, and changing this to something that said to me to add --storage just feels like a bad UI.
It is not like the user just typed these random characters and accidentally hit a container name.

mheon:
Does that example actually work? I seem to remember 'rm --storage'
requiring a full ID to work.

rhatdan:
Yes this example works  It is the primary reason for this PR.

This test added in this PR makes the change.
 @test "podman rm container from storage" {

pkg/domain/infra/abi/containers.go
```diff
@@ -294,6 +284,22 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
...
+			// remove container names that we successfully deleted
+			tmpNames = append(tmpNames, ctr)
+		} else {
+			reports = append(reports, &report)
+		}
+	}
+	if len(tmpNames) < len(names) {
+		names = tmpNames
+	}
+
```
mheon, line 302: Should we have a check here to return immediately if len(names) == 0 (everything was a storage container)?

TomSweeneyRedHat (commented):
LGTM

QiWang19:
LGTM with mheon's suggestion addressed.

pkg/domain/infra/abi/containers.go
rhatdan: No because we have not processed --all or --latest yet.
But there was a bug below.

Issue #7776, opened 2020-09-25
podman is unresponsive


ocafebabe:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

podman isn't working as expected after installing it on a system with a previous docker installation

**Steps to reproduce the issue:**

1. Install docker.io

2. Uninstall docker.io

3. Install podman using Ubuntu 20.04 Kubic repository

**Describe the results you received:** freeze

**Describe the results you expected:** working condition

**Additional information you deem important (e.g. issue happens only occasionally):**

Linux Mint 20 (Ubuntu 20.04)

**Output of `podman version`:** 

```
podman version 2.1.0
```

**Output of `podman info --debug`:**

```
N/A (doesn't work)

Output of: strace podman info --debug

newfstatat(AT_FDCWD, "/home/cbourque/.local/bin/newuidmap", 0xc00065b968, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/home/cbourque/bin/newuidmap", 0xc00065ba38, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/local/sbin/newuidmap", 0xc00065bb08, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/local/bin/newuidmap", 0xc00065bbd8, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/sbin/newuidmap", 0xc00065bca8, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/bin/newuidmap", 0xc00065bd78, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/sbin/newuidmap", 0xc00065be48, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/bin/newuidmap", 0xc00065bf18, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/games/newuidmap", 0xc00065e038, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/local/games/newuidmap", 0xc00065e108, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/snap/bin/newuidmap", 0xc00065e1d8, 0) = -1 ENOENT (No such file or directory)
[...]

PR #7892, opened 2020-10-02
[2.1] rootless: fix hang when newidmap is not installed

giuseppe:
when newidmap is not installed the code would hit the
reexec_in_user_namespace_wait code and wait for the child process to
be terminated.  The child process is blocked waiting on the w pipe.

So make sure to unblock the child process first and then clean it up.

Closes: https://github.com/containers/podman/issues/7776

Signed-off-by: Giuseppe Scrivano <giuseppe@scrivano.org>
(cherry picked from commit 7147c935aa29dca2f89083bfbe2c058a0ca69857)


rhatdan:
LGTM

PR #7894, opened 2020-10-02
WIP: Get safe subnets for integration testing

Luap99:
Currently if someone needs to add a network test with a
specific subnet they have to choose one manually and have
to make sure that this network is not already used. This
is error prone.

Even if the CI passes this can still lead to flakes since
the tests are run parallel and networks could intercept
each other. The proper solution for this is to make sure
that each test uses a globally unique subnet.

This commit implements a GetSafeIPv4Subnet function which
should be used to get a custom subnet for your tests.
The function ensures that the generated subnet is globally
unique across the ginkgo test nodes.

In order for this to work each test which uses a custom
subnet has to use this function. It's up to the maintainers
to make sure new test will use this function.

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -81,13 +81,13 @@ func GetFreeNetwork(config *config.Config) (*net.IPNet, error) {
 			return nil, err
 		}
 		logrus.Debugf("checking if network %s intersects with other cni networks", nextNetwork.String())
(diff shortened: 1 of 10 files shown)


Luap99:
I break the CI on purpose. I want to check if the function works correct .

rhatdan:
@Luap99 Needs a rebase.

rhatdan:
@Luap99 Another month has gone by, what do you want to do with this PR?

Luap99:
Closing for now. Maybe I will find the time to revisit this at some point.

PR #7901, opened 2020-10-02
compat: images/create: fix tag parsing

vrothberg:
The `tag` parameter of the compat `images/create` endpoint can be both,
a tag and a digest.  Fix parsing of the parameter to detect digests and
use the appropriate `@` separator.


baude:
nice job LGTM

Issue #7886, opened 2020-10-02
Processes are not visible to other containers in the pod


ah83:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When i start multiple containers having the same pod with a shared pid namespace, the container can only see it's own pids.
The shared pid namespace works, because every container in the pod has a different pid number and not pid 1.
Also _lsns_ shows that the pid namespace is shared between the containers.
After some debugging i found out that ps is blocked by selinux, because it can not access the pids from other containers in the proc filesystem.

**Steps to reproduce the issue:**

1. Create pod with pid namespace
```
podman pod create --name test --share=pid
```
2. Create first container running sleep
```
podman run --pod test -it --rm centos:8 sleep 1000
```

3. Create second container with bash and run ps aux
```
podman run --pod test -it --rm centos:8 bash
[root@ed6eb2547653 /]# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root          56  6.3  0.0  12024  3332 pts/0    Ss   08:20   0:00 bash
root          70  0.0  0.0  43960  3356 pts/0    R+   08:20   0:00 ps aux
```

**Describe the results you received:**
A listing of /proc in the second container shows pid 1 and the pids of the first container.
```
[root@ed6eb2547653 /]# ls -l /proc  | head -n 10 
total 0
dr-xr-xr-x.  9 root root    0 Oct  2 08:18 1
dr-xr-xr-x.  9 root root    0 Oct  2 08:18 21
dr-xr-xr-x.  9 root root    0 Oct  2 08:20 56
dr-xr-xr-x.  9 root root    0 Oct  2 08:21 74
dr-xr-xr-x.  9 root root    0 Oct  2 08:21 75
drwxrwxrwt.  2 root root   40 Oct  2 08:20 acpi
-r--r--r--.  1 root root    0 Oct  2 08:21 buddyinfo
dr-xr-xr-x.  4 root root    0 Oct  2 08:18 bus
-r--r--r--.  1 root root    0 Oct  2 08:21 cgroups
```
Denied by selinux
```
cd /proc/21/
bash: cd: /proc/21/: Permission denied
```
Listing of lables
```
 ls -lZ /proc  | head -n 10
total 0
dr-xr-xr-x.  9 root root system_u:system_r:container_t:s0:c507,c511         0 Oct  2 08:18 1
[...]

PR #7902, opened 2020-10-02
Add SELinux support for pods

rhatdan:
All containers within a Pod need to run with the same SELinux
label, unless overwritten by the user.

Also added a bunch of SELinux tests to make sure selinux labels
are correct on namespaces.

Fixes: https://github.com/containers/podman/issues/7886

### DIFF
--- a/pkg/specgen/generate/container_create.go
+++ b/pkg/specgen/generate/container_create.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/podman/v2/pkg/specgen"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/storage"
+	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


pkg/specgen/generate/container_create.go
```diff
@@ -272,6 +273,19 @@ func createContainerOptions(ctx context.Context, rt *libpod.Runtime, s *specgen.
 	// Security options
 	if len(s.SelinuxOpts) > 0 {
 		options = append(options, libpod.WithSecLabels(s.SelinuxOpts))
+	} else {
+		if pod != nil {
+			// duplicate the security options from the pod
+			processLabel, err := pod.ProcessLabel()
+			if err != nil {
+				return nil, err
```
mheon, line 281: This should probably be nonfatal, making pods without an infra container is perfectly valid.

rhatdan: There will be no SELinux label though, although we have to figure out how to get one.

rhatdan: I checked inside of the code where it checks for an infra container, if no infra then it will return "" for process label, which will get SELinux to fall back to default, which is to label the containers differently.
Added a test for this use case.

rhatdan:

jwhonce:
LGTM

vrothberg (commented):
/lgtm

test/e2e/run_selinux_test.go
```diff
@@ -182,4 +182,115 @@ var _ = Describe("Podman run", func() {
...
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(Not(Equal(label1)))
+	})
+
+	It("podman test --pid=host", func() {
+		session := podmanTest.Podman([]string{"run", "--pid=host", ALPINE, "cat", "/proc/self/attr/current"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring("spc_t"))
+	})
```
edsantiago, line 281: I cannot get this to work. I'm trying to write an equivalent system test, and no matter what, I get:
```console
$ ./bin/podman run --pid=host alpine cat /proc/self/attr/current
unconfined_u:system_r:container_runtime_t:s0%
$ ls -lZ bin/podman
-rwxrwxr-x. 1 esm esm system_u:object_r:container_runtime_exec_t:s0 54319936 Oct  5 15:08 bin/podman
```
(As root, I get `spc_t` as expected). master @ f48b1639348da2e398c4e30e10f542dec02c69d5, with container-selinux-2.145.0-1.fc32. 

I must be missing something simple. Any suggestions, please?

PR #7908, opened 2020-10-03
fix podman container exists and diff for storage containers

rhatdan:
Current these commands only check if a container exists in libpod. With
this fix, the commands will also check if they are in containers/storage.

This allows users to look at differences within a buildah or CRI-O container.

Currently buildah diff does not exists, so this helps out in that situation
as well as in CRI-O since the cri does not implement a diff command.

### DIFF
--- a/cmd/podman/containers/exists.go
+++ b/cmd/podman/containers/exists.go
@@ -12,10 +12,10 @@ var (
 	containerExistsDescription = `If the named container exists in local storage, podman container exists exits with 0, otherwise the exit code will be 1.`
 
 	existsCommand = &cobra.Command{
-		Use:   "exists CONTAINER",
+		Use:   "exists [flags] CONTAINER",
 		Short: "Check if a container exists in local storage",
 		Long:  containerExistsDescription,
-		Example: `podman container exists containerID
+		Example: `podman container exists --external containerID
   podman container exists myctr || podman run --name myctr [etc...]`,
 		RunE:                  exists,
 		Args:                  cobra.ExactArgs(1),
@@ -29,10 +29,19 @@ func init() {
 		Command: existsCommand,
 		Parent:  containerCmd,
 	})
+	flags := existsCommand.Flags()
+	flags.Bool("external", false, "Check external storage containers as well as Podman containers")
 }
 
 func exists(cmd *cobra.Command, args []string) error {
-	response, err := registry.ContainerEngine().ContainerExists(context.Background(), args[0])
+	external, err := cmd.Flags().GetBool("external")
+	if err != nil {
+		return err
+	}
+	options := entities.ContainerExistsOptions{
+		External: external,
+	}
+	response, err := registry.ContainerEngine().ContainerExists(context.Background(), args[0], options)
 	if err != nil {
 		return err
 	}
--- a/completions/bash/podman
+++ b/completions/bash/podman
@@ -2468,7 +2468,6 @@ _podman_rm() {
 	-i
 	--latest
 	-l
-	--storage
 	--volumes
 	-v
     "
@@ -2696,7 +2695,7 @@ _podman_ps() {
 	  --pod -p
 	  --quiet -q
 	  --size -s
-	  --storage
+	  --external
 	  --namespace --ns
 	  --sync
      "
@@ -3168,7 +3167,11 @@ _podman_container_exists() {
     "
 
     local boolean_options="
-  "
+	  --external
+	  -h
+	  --help
+    "
+
     case "$cur" in
 	-*)
 	    COMPREPLY=($(compgen -W "$boolean_options $options_with_args" -- "$cur"))
--- a/completions/bash/podman
+++ b/completions/bash/podman
@@ -755,10 +755,8 @@ _podman_attach() {
      --detach-keys
      "
      local boolean_options="
-	  --help
-	  -h
-	  --latest
-	  -l
+	  --help -h
+	  --latest -l
 	  --no-stdin
 	  --sig-proxy
      "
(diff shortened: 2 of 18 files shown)


rhatdan:
@mrunalp PTAL

test/system/030-run.bats
```diff
@@ -432,4 +432,19 @@ json-file | f
...
 
+@test "podman container exists" {
+    rand=$(random_string 30)
+    run_podman 1 container exists myctr
+
+    run_podman create --name myctr $IMAGE /bin/true
+    run_podman container exists myctr
+
+    # Create a container that podman does not know about
+    run buildah from $IMAGE
```
edsantiago, line 443: We cannot guarantee having buildah in the test environment. I have a test for this (storage containers) in the works, but (sigh) it's in a side branch and I forgot to submit it last week. Basically it runs 'podman build', with a `RUN sleep 15` but with `PODMAN_TIMEOUT=5` so the build is likely to leave behind a buildah container. Unfortunately it also requires other changes to `helpers.bash`. May I submit that next week?

rhatdan: What do we need to do to make sure buildah is installed?  It would seem a better solution especially as I add more container/storage features into podman,

edsantiago: in CI: go through the build-new-VMs dance. In Fedora/RHEL: add `Requires: buildah` to the `-tests` subpackage.

mheon:
I do not agree with 'exists'. Openstack requested that command to make sure
containers were present and *ready to use* so they did not need to recreate
them. We are breaking the intent of the command if it reports that a
container is present but said container is Buildah or CRI-O.

rhatdan:
Again, these will not happen by accident, you have to check by name if a container exists. I would be willing to add a flag to say we want a specific type, (But I would guess no one would ever use it, but by default we should do what human beings would expect).

mheon:
It looks like Github managed to lose my response email, so pasting here:

My concern is that the exists command will answer true after a partially failed deletion, where we remove the container in the database but fail to remove it in c/storage. This has come up on a regular basis in Openstack. They use scripts to check that containers exist and work with them, recreating if necessary. I think what you are doing has real potential to break this workflow.

rhatdan:
@edsantiago this tests just hit the timeout also "8 seconds"

rhatdan:
This looks like it is ready to merge.

rhatdan:
@mheon PTAL, this now does not do exists by default, you have to specify --external.

mheon:
LGTM

cmd/podman/containers/exists.go
```diff
@@ -12,10 +12,10 @@ var (
 	containerExistsDescription = `If the named container exists in local storage, podman container exists exits with 0, otherwise the exit code will be 1.`
 
 	existsCommand = &cobra.Command{
-		Use:   "exists CONTAINER",
+		Use:   "exists [flags] CONTAINER",
```
TomSweeneyRedHat, line 15: You've 'flags' here and 'options' in the man page, I think it should be options everywhere.
```suggestion
		Use:   "exists [options] CONTAINER",
```

completions/bash/podman
```diff
@@ -2696,7 +2695,7 @@ _podman_ps() {
 	  --pod -p
 	  --quiet -q
 	  --size -s
-	  --storage
+	  --external
```
TomSweeneyRedHat, line 2698: since you've another nit to fix, can you alphabetize this please?  Also namespace on the line below.

docs/source/markdown/podman-ps.1.md
```diff
@@ -34,13 +34,13 @@ all the containers information.  By default it lists:
...
 
-Note: Podman shares containers storage with other tools such as Buildah and CRI-O. In some cases these `external` containers might also exist in the same storage. Use the `--storage` option to see these external containers. External containers show the 'storage' status.
+Note: Podman shares containers storage with other tools such as Buildah and CRI-O. In some cases these `external` containers might also exist in the same storage. Use the `--external` option to see these external containers. External containers show the 'storage' status.
 
 **--pod**, **-p**
 
 Display the pods the containers are associated with
 
-**--storage**
+**--external**
```
TomSweeneyRedHat, line 43: alphabetize this option please

cmd/podman/containers/exists.go
rhatdan: Let's do this in a different PR.  We exchange flags and options all over the place, and one of the tests is looking for flags.

docs/source/markdown/podman-container-exists.1.md
```diff
@@ -17,22 +17,30 @@ was an issue accessing the local storage.
 **-h**, **--help**
 Print usage statement
 
+**--external**=*true|false*
```
TomSweeneyRedHat, line 20: move above help please

completions/bash/podman
rhatdan: I would prefer to do this in a separate PR.  Lets go through the entire contrib and alphabetize it all.

test/system/030-run.bats
TomSweeneyRedHat: @edsantiago is this OK now?

TomSweeneyRedHat:
I'm fine with picking the man page nits in another PR.  If @edsantiago thinks the testing end is gtg, then LGTM.

cmd/podman/containers/exists.go
rhatdan: grep -r Use.*'\[flags' cmd | wc -l
82

docs/source/markdown/podman-container-exists.1.md
rhatdan: Done

docs/source/markdown/podman-ps.1.md
rhatdan: Done

test/system/030-run.bats
rhatdan: Tests would not pass if it was not.

rhatdan:
@containers/podman-maintainers @saschagrunert This is ready to merge.

saschagrunert (approved):
/lgtm

test/system/030-run.bats
edsantiago: It's OK in CI, but gating tests are failing because of missing buildah. On my list for today.

PR #7916, opened 2020-10-05
[v2.0.5-rhel] compat: images/create: fix tag parsing

vrothberg:
The `tag` parameter of the compat `images/create` endpoint can be both,
a tag and a digest.  Fix parsing of the parameter to detect digests and
use the appropriate `@` separator.

Backport of commit 6a291942c253fcb13fc2ee0990f2137ca3584270.

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -20,8 +20,24 @@ import (
 	"github.com/containers/libpod/v2/pkg/util"
 	"github.com/docker/docker/api/types"
 	"github.com/gorilla/schema"
+	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
+
+// mergeNameAndTagOrDigest creates an image reference as string from the
+// provided image name and tagOrDigest which can be a tag, a digest or empty.
+func mergeNameAndTagOrDigest(name, tagOrDigest string) string {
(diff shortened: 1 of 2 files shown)


vrothberg:
/hold

Let's wait for @jnovy's ACK.

jnovy:
Thanks Valentin. While I see no problem with inclusion of this fix in 8.3.1 - where we have podman-2.1.1 already - for 8.3.0 and the 2.0.5-rhel branch it is way too late. Only blockers/regressions which need to go through approval process can be fixed in that branch now via 0day update. This is very unlikely to happen for this one.

For 8.3.1 I filed https://bugzilla.redhat.com/show_bug.cgi?id=1884668 to assure it will go through QE once 8.3.1 is tested.

vrothberg:
> Thanks Valentin. While I see no problem with inclusion of this fix in 8.3.1 - where we have podman-2.1.1 already - for 8.3.0 and the 2.0.5-rhel branch it is way too late. Only blockers/regressions which need to go through approval process can be fixed in that branch now via 0day update. This is very unlikely to happen for this one.
> 
> For 8.3.1 I filed https://bugzilla.redhat.com/show_bug.cgi?id=1884668 to assure it will go through QE once 8.3.1 is tested.

Thanks for clarifying, @jnovy!

I'll close this one here and open against v2.1.

PR #7919, opened 2020-10-05
Fix handling of CheckRootlessUIDRange

rhatdan:
If I have multiple ranges of UIDs specified in the /etc/subuid, this check
blows up and incorrectly blocks the use of --user flag.

### DIFF
--- a/pkg/util/utils_linux.go
+++ b/pkg/util/utils_linux.go
@@ -60,11 +60,12 @@ func CheckRootlessUIDRange(uid int) error {
 	if err != nil {
 		return err
 	}
+	total := 0
 	for _, u := range uids {
-		// add 1 since we also map in the user's own UID
-		if uid > u.Size+1 {
-			return errors.Errorf("requested user's UID %d is too large for the rootless user namespace", uid)
-		}
+		total += u.Size
+	}
+	if uid > total {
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

pkg/util/utils_linux.go
```diff
@@ -60,11 +60,12 @@ func CheckRootlessUIDRange(uid int) error {
...
+	total := 0
 	for _, u := range uids {
-		// add 1 since we also map in the user's own UID
-		if uid > u.Size+1 {
-			return errors.Errorf("requested user's UID %d is too large for the rootless user namespace", uid)
-		}
+		total += u.Size
+	}
+	if uid > total {
+		return errors.Errorf("requested user's UID %d is too large for the rootless user namespace", uid)
```
TomSweeneyRedHat, line 68: Does it make sense to keep this if statement inside of the for still to have it break ASAP?

rhatdan: No because we need to add all of the ranges together to see if UID is less then.  We could check on each loop, I guess to see if it is less then, but that seems to be an optimization with little value.  There are not going to be large numbers of ranges in the /etc/subuid file.

PR #7920, opened 2020-10-05
Cirrus: Fix branch-validation failure

cevich:
When validating code on a branch, determining a starting commit to check
from isn't as straightforward as it would seem.  Default to using the
SHA from last time CI was green.  If for some reason that isn't
available, use an obviously wrong value to cause an intentional
failure.


edsantiago:
LGTM, but we'll need to keep an eye on this and possibly fine-tune.

cevich:
Repushed with fix for shellcheck error.

PR #7926, opened 2020-10-05
Cirrus: Skip deep testing on branches

cevich:
Previous to this commit, the entire suite of CI tasks run in a PR, run
again for every merge (a.k.a. branch push).  This wastes time and
resources with substantively overlapping testing.  The primary reason
to test on branch-push, is providing coverage for merge-semantics.
In other words, problems introduced due to the sequence of PR merging.
For this purpose, the vast majority of problems can be caught quickly by
a small subset of automated tests.  If deeper debugging is necessary,
then opening a test-PR is a small price to ask for the enormous amount
of time/resource savings with more limited branch-push testing.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -87,6 +87,7 @@ ext_svc_check_task:
 automation_task:
     alias: 'automation'
     name: "Check Automation"
+    skip: &branch "$CIRRUS_PR == '' || $CIRRUS_TAG != ''" # Don't run for branches
     container: *smallcontainer
     env:
         TEST_FLAVOR: automation
@@ -231,6 +232,7 @@ validate_task:
 bindings_task:
     name: "Test Bindings"
     alias: bindings
+    skip: *branch
     depends_on:
         - build
     gce_instance: *standardvm
@@ -419,6 +421,7 @@ local_integration_test_task: &local_integration_test_task
     # Integration-test task name convention:
     # <int.|sys.> <podman|remote> <Distro NV> <root|rootless>
     name: &std_name_fmt "$TEST_FLAVOR $PODBIN_NAME $DISTRO_NV $PRIV_NAME $TEST_ENVIRON"
+    skip: *branch
     alias: local_integration_test
     depends_on:
         - unit_test
@@ -456,6 +459,7 @@ remote_integration_test_task:
 container_integration_test_task:
     name: *std_name_fmt
     alias: container_integration_test
+    skip: *branch
     depends_on:
         - unit_test
     matrix: &fedora_vm_axis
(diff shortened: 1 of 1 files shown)


edsantiago:
Perhaps I'm the only one... or perhaps I'm just ignorant... but the use of "branches" here seems really confusing to me. Maybe it's a Cirrus-ism or Github-ism? To me—and this assumes that I'm understanding this correctly—it would make more sense to label this as "Skip deep testing on merge action" or "on post-merge". Or have I missed something?

To put it another way: I want our team to have a subset of people familiar-enough with Cirrus to be able to understand and review these PRs, but without being utter SMEs like you. I think the term "branch" makes sense to SMEs but not us dabblers. Can we have terminology that makes sense to both groups?

cevich:
@edsantiago I respect your motivation here.  Really it's a "branch-push" that triggers a build (a.k.a. "job" or "run of all the tests").  Contrast this with Tag-push or a PR-push triggering events.  A branch-push could happen because of a PR merge or because someone manually ran `git push upstream HEAD:master` (depending on the github repository settings at the moment).  Though generally speaking, on this specific repo., with the current repo. settings, it ***will*** be a PR-merge that updates a branch.

So my hesitation to use the term "post-merge" is avoiding any ambiguity confusion WRT the other terms (PR-push and Tag-push) and other actions which could trigger testing.

cevich:
Almost forgot:  Cirrus has support for cron-triggered branch jobs as well.  So there's another way branch testing can happen.

edsantiago:
OK - so it's just me who needs to get used to that terminology. I guess it LGTM then. For ease of visual review, and for posterity, attaching 
flow diagram with a quick hack showing `[SKIPPABLE]` for newly added skips.

cevich:
Thanks Ed.

edsantiago:
> Cirrus has support for cron-triggered branch jobs as well.

Reflecting a bit more on this, I can see the desire to skip remote integration testing on post-merge, but I would think a nightly cron job would be low-impact and we might want as much testing as we can get. I'm not saying we should enable that right now, but if we do want to, would it be possible to selectively enable all tests via cron? Maybe some way to pass an envariable to `.cirrus.yml`?

edsantiago (approved):
/lgtm
/hold

cevich:
> would it be possible to selectively enable all tests via cron?

@edsantiago yes, `$CIRRUS_CRON` will be non-empty if the branch build (job) was triggered by cron.  If you'd like to figure out the `skip:` logic to make that work, please (please!) take that liberty.  I would undoubtedly make it way more complicated than necessary :grin: 

Note:  Contrary to the docs, `$CRON_BRANCH`will be non-empty for both a PR-push AND branch-push.  For a PR it will be set to `pull/<number>` :frowning_face:

PR #7929, opened 2020-10-05
Nits

kolyshkin:
Lots of small changes, needs a careful review...

### 1. pkg/cgroups/createCgroupv2Path: nits
    
1. Check the path validity before trying to read the cgroup.controllers.  
2. Do not hardcode "/sys/fs/cgroup".
3. Simplify creating the "+this +that" string.
4. Do not wrap ioutil.WriteFile error.

### 2. Remove excessive error wrapping

In case os.Open[File], os.Mkdir[All], ioutil.ReadFile and the like
fails, the error message already contains the file name and the
operation that fails, so there is no need to wrap the error with
something like "open %s failed".
    
While at it
    
- replace a few places with os.Open, ioutil.ReadAll with
ioutil.ReadFile.
    
- replace errors.Wrapf with errors.Wrap for cases where there
are no %-style arguments.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -835,13 +835,13 @@ func (c *Container) checkpointRestoreLabelLog(fileName string) error {
 
 	logFile, err := os.OpenFile(dumpLog, os.O_CREATE, 0600)
 	if err != nil {
-		return errors.Wrapf(err, "failed to create CRIU log file %q", dumpLog)
+		return errors.Wrap(err, "failed to create CRIU log file")
 	}
 	if err := logFile.Close(); err != nil {
-		logrus.Errorf("unable to close log file: %q", err)
+		logrus.Error(err)
 	}
 	if err = label.SetFileLabel(dumpLog, c.MountLabel()); err != nil {
-		return errors.Wrapf(err, "failed to label CRIU log file %q", dumpLog)
+		return err
 	}
 	return nil
 }
@@ -919,7 +919,7 @@ func (c *Container) checkpoint(ctx context.Context, options ContainerCheckpointO
 func (c *Container) importCheckpoint(input string) error {
 	archiveFile, err := os.Open(input)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to open checkpoint archive %s for import", input)
+		return errors.Wrap(err, "Failed to open checkpoint archive for import")
 	}
 
 	defer archiveFile.Close()
@@ -1116,7 +1116,7 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
 			// Only do this if a rootfs-diff.tar actually exists
 			rootfsDiffFile, err := os.Open(rootfsDiffPath)
 			if err != nil {
-				return errors.Wrapf(err, "Failed to open root file-system diff file %s", rootfsDiffPath)
+				return errors.Wrap(err, "Failed to open root file-system diff file")
 			}
 			defer rootfsDiffFile.Close()
 			if err := c.runtime.ApplyDiffTarStream(c.ID(), rootfsDiffFile); err != nil {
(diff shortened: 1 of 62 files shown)


libpod/container_internal_linux.go
```diff
@@ -835,13 +835,13 @@ func (c *Container) checkpointRestoreLabelLog(fileName string) error {
 
 	logFile, err := os.OpenFile(dumpLog, os.O_CREATE, 0600)
 	if err != nil {
-		return errors.Wrapf(err, "failed to create CRIU log file %q", dumpLog)
+		return err
```
mheon, line 838: I think the CRIU log file bit of the error might be useful here.

libpod/container_internal_linux.go
```diff
@@ -1125,16 +1125,10 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
...
-				return errors.Wrapf(err, "Failed to open deleted files file %s", deletedFilesPath)
-			}
-			defer deletedFilesFile.Close()
-
 			var deletedFiles []string
-			deletedFilesJSON, err := ioutil.ReadAll(deletedFilesFile)
+			deletedFilesJSON, err := ioutil.ReadFile(deletedFilesPath)
 			if err != nil {
-				return errors.Wrapf(err, "Failed to read deleted files file %s", deletedFilesPath)
+				return errors.Wrapf(err, "Failed to read deleted files file")
```
mheon, line 1131: F in Failed needs to be lowercased... I'm actually surprised Lint didn't catch this before your PR.

mheon:
@giuseppe PTAL at the `pkg/cgroups` bit please

mheon:
Few nits, but LGTM overall

libpod/oci_conmon_linux.go
```diff
@@ -157,15 +157,13 @@ func newConmonOCIRuntime(name string, paths []string, conmonPath string, runtime
...
-				runtime.exitsDir)
+			return nil, errors.Wrapf(err, "error creating OCI runtime exit files directory")
 		}
 	}
 	if err := os.MkdirAll(runtime.socketsDir, 0750); err != nil {
 		// The directory is allowed to exist
 		if !os.IsExist(err) {
-			return nil, errors.Wrapf(err, "error creating OCI runtime attach sockets directory %s",
-				runtime.socketsDir)
+			return nil, errors.Wrap(err, "error creating OCI runtime attach sockets directory")
```
TomSweeneyRedHat, line 166: For these directory errors, will the directory that can't be created be included in the error message somewhere else?  I think that's important to include in the message.

rhatdan: The err returned my os.MkdirAll will include the path.

rhatdan (approved):
LGTM other then @mheon issue.

libpod/container_internal_linux.go
kolyshkin: fixed

libpod/container_internal_linux.go
kolyshkin: Will address by a separate commit (in this PR).

kolyshkin: There are other 100+ places like this :)

See the new commit

kolyshkin: Some of the remaining cases can be seen by running
```console
[kir@kir-rhat libpod]$ git grep -F errors. | grep -v ^vendor | grep '"[A-Z]'
```

but I don't have the courage to fix.

mheon: Well that's awfully concerning - I thought we had a linter to catch this, but I'm betting we lost it in the move to golangci-lint.

giuseppe:
LGTM

giuseppe (approved):

vrothberg (commented):
LGTM, thanks @kolyshkin !

PR #7930, opened 2020-10-05
Fix the "err: cause" order of OCI runtime errors

xordspar0:
Previously, the order of OCI error messages was reversed, so that the
type of error was listed as the cause. For example:

    Error: writing file `cpu.cfs_quota_us`: Invalid argument: OCI runtime error

This error message makes it seem like "OCI runtime error" is the
argument that was invalid. In fact, "OCI runtime error" is the error and
"writing file ..." is the cause. With this change, the above message
reads:

    Error: OCI runtime error: writing file `cpu.cfs_quota_us`: Invalid argument

### DIFF
--- a/libpod/oci_util.go
+++ b/libpod/oci_util.go
@@ -112,19 +112,28 @@ func bindPorts(ports []ocicni.PortMapping) ([]*os.File, error) {
 func getOCIRuntimeError(runtimeMsg string) error {
 	includeFullOutput := logrus.GetLevel() == logrus.DebugLevel
 
+	runtimeMsg = strings.Trim(runtimeMsg, "\n")
+
 	if match := regexp.MustCompile("(?i).*permission denied.*|.*operation not permitted.*").FindString(runtimeMsg); match != "" {
 		errStr := match
 		if includeFullOutput {
 			errStr = runtimeMsg
 		}
-		return errors.Wrapf(define.ErrOCIRuntimePermissionDenied, "%s", strings.Trim(errStr, "\n"))
+
+		runtimeErr := errors.New(errStr)
+		return errors.Wrapf(runtimeErr, "%v", define.ErrOCIRuntimePermissionDenied)
 	}
+
 	if match := regexp.MustCompile("(?i).*executable file not found in.*|.*no such file or directory.*").FindString(runtimeMsg); match != "" {
 		errStr := match
 		if includeFullOutput {
 			errStr = runtimeMsg
 		}
-		return errors.Wrapf(define.ErrOCIRuntimeNotFound, "%s", strings.Trim(errStr, "\n"))
+
+		runtimeErr := errors.New(errStr)
+		return errors.Wrapf(runtimeErr, "%v", define.ErrOCIRuntimeNotFound)
 	}
-	return errors.Wrapf(define.ErrOCIRuntime, "%s", strings.Trim(runtimeMsg, "\n"))
+
+	runtimeErr := errors.New(runtimeMsg)
+	return errors.Wrapf(runtimeErr, "%v", define.ErrOCIRuntime)
 }
--- a/libpod/define/errors.go
+++ b/libpod/define/errors.go
@@ -1,171 +1,181 @@
 package define
 
 import (
-	"errors"
+	"strings"
 )
+
+type LibpodError string
+
+func (e LibpodError) Error() string {
+	return string(e)
+}
+
+func (e LibpodError) Is(target error) bool {
+	return strings.HasPrefix(target.Error(), string(e))
+}
 
 var (
 	// ErrNoSuchCtr indicates the requested container does not exist
-	ErrNoSuchCtr = errors.New("no such container")
+	ErrNoSuchCtr = LibpodError("no such container")
 
 	// ErrNoSuchPod indicates the requested pod does not exist
-	ErrNoSuchPod = errors.New("no such pod")
+	ErrNoSuchPod = LibpodError("no such pod")
 
 	// ErrNoSuchImage indicates the requested image does not exist
-	ErrNoSuchImage = errors.New("no such image")
+	ErrNoSuchImage = LibpodError("no such image")
 
 	// ErrNoSuchTag indicates the requested image tag does not exist
-	ErrNoSuchTag = errors.New("no such tag")
+	ErrNoSuchTag = LibpodError("no such tag")
 
 	// ErrNoSuchVolume indicates the requested volume does not exist
-	ErrNoSuchVolume = errors.New("no such volume")
+	ErrNoSuchVolume = LibpodError("no such volume")
 
 	// ErrNoSuchNetwork indicates the requested network does not exist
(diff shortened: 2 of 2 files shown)


rhatdan:
@mheon PTAL

libpod/oci_util.go
```diff
@@ -112,19 +112,28 @@ func bindPorts(ports []ocicni.PortMapping) ([]*os.File, error) {
...
+
 	if match := regexp.MustCompile("(?i).*permission denied.*|.*operation not permitted.*").FindString(runtimeMsg); match != "" {
 		errStr := match
 		if includeFullOutput {
 			errStr = runtimeMsg
 		}
-		return errors.Wrapf(define.ErrOCIRuntimePermissionDenied, "%s", strings.Trim(errStr, "\n"))
+
+		runtimeErr := errors.New(errStr)
+		return errors.Wrapf(runtimeErr, "%v", define.ErrOCIRuntimePermissionDenied)
```
mheon, line 124: I'm against this - we want `define.ErrOCIRuntimePermissionDenied` to be the cause of the error, so we can check against it with `errors.Cause` later

xordspar0: First of all, I'm a little disappointed that there isn't something like `errors.WrapWithError(cause, wrapper error)`. Since there isn't, I defined our own error type, `LibpodError`, that will respond to `errors.Is` appropriately.

```go
annotatedError := errors.Wrapf(runtimeErr, "%v", define.ErrOCIRuntime)
fmt.Println(annotatedError)                                       // OCI runtime error: writing file `cpu.cfs_quota_us`: Invalid argument
fmt.Println(errors.Cause(annotatedError))                         // writing file `cpu.cfs_quota_us`: Invalid argument
fmt.Println(errors.Cause(annotatedError) == define.ErrOCIRuntime) // false
fmt.Println(errors.Is(annotatedError, define.ErrOCIRuntime))      // true

moreAnnotatedError := errors.Wrap(annotatedError, "request failed")
fmt.Println(errors.Is(annotatedError, define.ErrOCIRuntime))      // true
```

---

I admit that I'm not super happy with [the implementation of LibpodErr.Is](https://github.com/xordspar0/podman/blob/04084a2c8e622d19664383e1f717168ac824aaa0/libpod/define/errors.go#L13-L15):

```go
func (e LibpodError) Is(target error) bool {
	return strings.HasPrefix(target.Error(), string(e))
}
```

Just checking for a string prefix is not very precise, but I couldn't figure out a way to insert the actual error into the chain. I'll keep thinking about this.

xordspar0: So to be specific, I'm proposing that we use `errors.Is` instead of `errors.Cause` to identify what kind of error it is. Then it doesn't matter where in the chain the `define.ErrOCIRuntimePermissionDenied` error is.

rhatdan:
The goal should be to improve the user experience, which means make the error messages easier to decipher.  Of course another huge goal would be to make sure the code continues to work.  Since this does not effect remote, then it would be fairly easy for us to change.  But what about the calls to errors.Cause versus errors.Is for non libpod.Errors?

mheon:
I really don't like forcing a distinction on `errors.Cause()` for some errors versus `errors.Is()` for other errors.

Would it be easier to go about this another way, and when we print the errors in cmd/podman, detect if errors.Cause() is an OCI runtime error, and strip the suffix from the error string if so?

xordspar0:
> a distinction on errors.Cause() for some errors versus errors.Is()

I would vote for using `errors.Is()` for everything. In the official [Go 1.13 errors API](https://blog.golang.org/go1.13-errors) that was based on github.com/pkg/errors there isn't an equivalent of `errors.Cause()`. You just always check if the error you care about is _anywhere_ in the chain and you don't care if it's in a specific position.

I think this is a better solution because it reduces coupling. It makes it easier to change internal details of libpod without disrupting calling code. For example, let's say I use `errors.Cause()` to detect a `define.ErrOCIRuntimePermissionDenied` error. What if someday the code that generated the error calls a system call and wants to wrap the error returned by the kernel, such as EPERM? That would break code that uses `errors.Cause()`.

Another issue with using `errors.Cause()` is that there can be only one ultimate cause at the end of the chain. If you want to detect a specific kind of error, you have to put it at the end of the chain, even if it isn't really the cause but more of a description. That's what's happening here with "Error: writing file `cpu.cfs_quota_us`: Invalid argument: OCI runtime error".

I'm not the Go or errors expert in the room, and I have a lot to learn, but I can't think of a time when I would want to use `errors.Cause()` when I have `errors.Is()` available.

> Would it be easier to go about this another way, and when we print the errors in cmd/podman, detect if errors.Cause() is an OCI runtime error, and strip the suffix from the error string if so?

We wouldn't have to think about consequences to the surrounding code if we did it that way. I can do that if that's what you prefer.

rhatdan:
@kolyshkin @vrothberg PTAL

vrothberg:
I'm okay with migrating the code base from "github.com/pkg/errors" to what newer Go standard libs provide. But that should be separate from tackling the `OCI runtime error` issue addressed in this PR.

I don't like creating a new `LibpodError` type as I see beauty in having consistency across the code base.

However, I guess that `pkg/errors` will continue to be used (in new PRs) for a long time as it has been used for a really long time. Go was quite late to improve error handling, and the muscle memory is just so strong :^)

xordspar0:
> Would it be easier to go about this another way, and when we print the errors in cmd/podman, detect if errors.Cause() is an OCI runtime error, and strip the suffix from the error string if so?

> I don't like creating a new LibpodError type as I see beauty in having consistency across the code base.

Alright, then we can fix this at the other end when we print out the error. I'll open a new PR.

xordspar0:
Superseded by https://github.com/containers/podman/pull/7968

xordspar0:
In case anyone in the future (probably me) wants to know how we got from "let's switch the order of these two errors" to a +72 -53 patch, here's what happened:

```
errors.Wrapf(define.ErrOCIRuntime, "%s", strings.Trim(runtimeMsg, "\n"))
```

1. ErrOCIRuntime is listed as the cause, but the runtimeMsg is the real cause. Just switch the arguments.
2. errors.Wrapf only takes on error argument: the cause. The rest is just a formatted string. Fine, we'll just convert the error to a string with "%v".
3. If we just take the string from define.ErrOCIRuntime instead of inserting the actual error into the chain, then we can't check for it with errors.Is or errors.Cause.
4. The [Go 1.13 errors API](https://golang.org/pkg/errors/#Is) provides a way to wrap errors and still check for them programatically later: define your own error type and implement the method `func (m MyError) Is(target error) bool`.

So that's what I did. The error message is in the right order and you can still check if this is an OCI runtime error. But it required a pretty big change, plus callers now need to use `errors.Is()` instead of `errors.Cause()`, which causes a ripple effect through the whole codebase. It was a nice idea but it ended up way too complicated for the problem I wanted to solve.

rhatdan:
I agree that we should move to the Golang handling of this, but it will be a massive job, since the code is covered with errors.Cause calls up and down the stack.

Issue #7434, opened 2020-08-25
[Question] Is there a way to limit the size of the logs?


yamrzou:
/kind feature

Is there a way to make sure container logs do not grow indefinitely?
Would it be safe to remove ~/.local/share/containers/storage/overlay-containers/*/userdata/ctr.log while the container is running?

mheon:
There is a maximum log size for containers, in bytes, defined in `containers.conf` - it is defaulted to -1, unlimited. I believe setting that should accomplish what you're looking for?

TomSweeneyRedHat:
If so, this might be a good entry for the troubleshooting guide....

yamrzou:
Thank you, yes that's should accomplish what I'm looking for.

rhatdan:
Labeling this as containers.conf, since I should write this up in the forthcoming blog on containers.conf.

Luap99:
I just want to add that docker uses `--log-opt` for this. https://docs.docker.com/config/containers/logging/json-file/#options
`docker run -it --log-opt max-size=10m --log-opt max-file=3 alpine ash`
This might be worth mentioning.

vrothberg:
@rhatdan, did you add that to the blog? Can we close the issue?

rhatdan:
I have not written any blogs yet.  Too much vacationing.

rhatdan:
I opened this related PR. https://github.com/containers/conmon/issues/211

pmoogi-redhat:
Hello @rhatdan .  For my simulations on log-loss i needed to change limit on log size - so created changes in my repo as https://github.com/containers/podman/compare/master...pmoogi-redhat:supportforlogsizemax. It is found working fine as expected. i did some validation on log files written onto disk. This is my first change to podman repo. Kindly let me know if PR can be submitted here reflecting this change in case you find it useful and if it doesn't hurt other modules which are dependent on this repo.

mheon:
That change looks fine on the whole, feel free to make a PR and we can see about getting it landed.

rhatdan:
Yes @pmoogi-redhat Open a PR.

PR #7931, opened 2020-10-05
Support max_size logoptions

rhatdan:
Docker supports log-opt max_size and so does conmon (ALthough poorly).
Adding support for this allows users to at least make sure their containers
logs do not become a DOS vector.

Fixes: https://github.com/containers/podman/issues/7434


rhatdan:
@haircommander I think we need to fix conmon to at least handle this properly.  It should keep two times max size.  Or we need to implement the method to handle multiple log files options.

TomSweeneyRedHat:
LGTM
assuming happy tests

Issue #7490, opened 2020-08-28
hosts file not populated when using user namespace


st1971:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
when running a rooted container with either --uidmap or --subuidname /etc/hosts does not have an entry for the container.

**Steps to reproduce the issue:**

1. sudo podman run --rm -it --uidmap=0:100000:5000 --hostname=mycontainer fedora:32 /bin/bash -c "cat /etc/hosts"

2.

3.

**Describe the results you received:**
Output of the content of /etc/hosts without an entry for mycontainer
```
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```

**Describe the results you expected:**
Output of the content of /etc/hosts with an entry for mycontainer, as you get without the --uidmap or --subuidname flag
```
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
10.88.0.164	my-container
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.5
API Version:  1
Go Version:   go1.14.6
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.19, commit: 5dce9767526ed27f177a8fa3f281889ad509fea7'
  cpus: 8
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: xxxxxx.lan
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.7.17-200.fc32.x86_64
  linkmode: dynamic
  memFree: 27110072320
  memTotal: 33613930496
  ociRuntime:
    name: crun
    package: crun-0.14.1-3.fc32.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      commit: 598ea5e192ca12d4f6378217d3ab1415efeddefa
      spec: 1.0.0
[...]

PR #7937, opened 2020-10-06
Populate /etc/hosts file when run in a user namespace

rhatdan:
We do not populate the hostname field with the IP Address
when running within a user namespace.

Fixes https://github.com/containers/podman/issues/7490

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -976,6 +976,21 @@ func (c *Container) completeNetworkSetup() error {
 			}
 		}
 	}
+	// check if we have a bindmount for /etc/hosts
+	if hostsBindMount, ok := state.BindMounts["/etc/hosts"]; ok && len(c.cniHosts()) > 0 {
+		ctrHostPath := filepath.Join(c.state.RunDir, "hosts")
+		if hostsBindMount == ctrHostPath {
+			// read the existing hosts
+			b, err := ioutil.ReadFile(hostsBindMount)
+			if err != nil {
+				return err
+			}
+			if err := ioutil.WriteFile(hostsBindMount, append(b, []byte(c.cniHosts())...), 0644); err != nil {
+				return err
+			}
+		}
+	}
+
 	// check if we have a bindmount for resolv.conf
 	resolvBindMount := state.BindMounts["/etc/resolv.conf"]
 	if len(outResolvConf) < 1 || resolvBindMount == "" || len(c.config.NetNsCtr) > 0 {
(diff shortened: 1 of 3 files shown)


libpod/container_internal.go
```diff
@@ -976,6 +976,19 @@ func (c *Container) completeNetworkSetup() error {
...
 		}
 	}
+	// check if we have a bindmount for /etc/hosts
+	hostsBindMount := state.BindMounts["/etc/hosts"]
+	if hostsBindMount != "" && len(c.cniHosts()) > 0 {
+		// read the existing hosts
+		b, err := ioutil.ReadFile(hostsBindMount)
+		if err != nil {
+			return err
+		}
```
TomSweeneyRedHat, line 986: if /etc/hosts doesn't exist, do we want to fail or just continue on?

libpod/container_internal.go
```diff
@@ -976,6 +976,19 @@ func (c *Container) completeNetworkSetup() error {
 			}
 		}
 	}
+	// check if we have a bindmount for /etc/hosts
+	hostsBindMount := state.BindMounts["/etc/hosts"]
```
mheon, line 980: Could do `_, hastBindMount := state.BindMounts...` to get a bool directly instead

libpod/container_internal.go
rhatdan: If this fails, this means that the system is going to fail later when it attempts to mount the file.  So yes we want to fail

mheon: Should error - something has gone seriously wrong in container creation if BindMounts refers to a path that does not exist.

libpod/container_internal.go
```diff
@@ -976,6 +976,19 @@ func (c *Container) completeNetworkSetup() error {
...
+	if hostsBindMount != "" && len(c.cniHosts()) > 0 {
```
mheon, line 981: This should also only be done if `c.config.PostConfigureNetNS` is set.

rhatdan: This check happened above
	if !c.config.PostConfigureNetNS || netDisabled {
		return nil
	}
So this will not happen if c.config.PostConfigureNetNS is not set or net was disabled.

libpod/container_internal.go
```diff
@@ -976,6 +976,18 @@ func (c *Container) completeNetworkSetup() error {
...
 		}
 	}
+	// check if we have a bindmount for /etc/hos1ts
+	if hostsBindMount, ok := state.BindMounts["/etc/hosts"]; ok && len(c.cniHosts()) > 0 {
+		// read the existing hosts
+		b, err := ioutil.ReadFile(hostsBindMount)
+		if err != nil {
+			return err
+		}
+		if err := ioutil.WriteFile(hostsBindMount, append(b, []byte(c.cniHosts())...), 0644); err != nil {
```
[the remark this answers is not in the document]
rhatdan: I added a check to make sure the path is an internal to the container path.  
If there is a volume mount by the user, or by a shared container network namespace, we will not add the name mapping.

test/e2e/run_networking_test.go
```diff
@@ -477,6 +477,17 @@ var _ = Describe("Podman run networking", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	FIt("podman run --uidmap /etc/hosts contains --hostname", func() {
```
giuseppe, line 480: FIt -> It

giuseppe (commented):
a fix for the test, otherwise LGTM

test/e2e/run_networking_test.go
rhatdan: oops.  Too early in the morning I guess.

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM
but the ubuntu-20 root host test is having issues with some of the networking tests, but I couldn't see what, I've rerun the test in hope that it's a flake.

Issue #7807, opened 2020-09-28
podman network create: race: error reading */conflist: ENOENT


edsantiago:
There seems to be a race condition causing flakes when multiple tests run 'podman network create/rm' at once:
```
  podman network remove with two networks
...
Running: podman [options] network create net2
Error: in /etc/cni/net.d/testNetTwoCNI2.conflist: error reading /etc/cni/net.d/testNetTwoCNI2.conflist: open /etc/cni/net.d/testNetTwoCNI2.conflist: no such file or directory
```
Probable cause: `podman inspect container two CNI networks` test running at just the right time, and running its `network rm` at just the wrong moment.

Log: [from PR 7803](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6498441721282560/html/integration_test.log.html#t--podman-network-remove-with-two-networks--1)

edsantiago:
(I'm not going to bother trying to reproduce this. I'm gambling that it's trivial to find and fix the race. If it isn't, please LMK and I'll dive into it further).

mheon:
It should be possible to get the tests to use separate CNI config
directories - have a per-test directory and a global one with the main
configs. Might be worth investigating to prevent these kinds of issues
entirely

edsantiago:
@mheon by "should be possible" do you mean "we currently have existing functionality, e.g. a --confdir option or envariable, that the tests can use"? Or "it should be possible to add such functionality"?

mheon:
The functionality exists but isn't wired into the CLI last I checked
(config file only). Should not be difficult to add but would require code
changes.

Luap99:
Actually we support this already: `podman --cni-config-dir ~/test network ls`

rhatdan:
@edsantiago Is that what you want?

edsantiago:
@rhatdan I haven't looked into it as closely at it deserves. My impression is that it might help with CI flakes but still leave us exposed to real bugs: I do suspect there's a real race condition in which (perhaps):
1. process 1 (`podman network create`) reads the conflist dir; then
1. something removes one of those files; then
1. process 1 tries to read one of the conflist files it read in step 1, and fails.

One possibility is for `podman network create` to lock the cni dir (yuk); another might be to ignore ENOENT. But again, this is just pure guesswork on my part.

Luap99:
`podman network create` is indeed very racy.

Just try this:
```
$ podman network create & podman network create
[...]

PR #7943, opened 2020-10-06
prevent unpredictable results with network create|remove

baude:
due to a lack of "locking" on cni operations, we could get ourselves in trouble when doing rapid creation or removal of networks.  added a simple file lock to deal with the collision and because it is not considered a performent path, use of the file lock should be ok.  if proven otherwise in the future, some generic shared memory lock should be implemented for libpod and also used here.

Fixes: #7807


pkg/api/handlers/compat/networks.go
```diff
@@ -305,8 +306,9 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "network not found", http.StatusNotFound, define.ErrNoSuchNetwork)
 		return
 	}
-	if err := network.RemoveNetwork(config, name); err != nil {
+	if err := network.RemoveNetwork(config, name, config.Engine.TmpDir); err != nil {
```
rhatdan, line 309: Does this path make sense?  Should this be the cni path?  IE Could multiple containers create content by using different TmpDirs?

baude: this is using libpod's tmpdir to get a lock for all of cni.  at this point, containers dont matter.  the cni path is for where the plugins live on the filesystem so we know how/where to call them as they are not part of path.  the lock in question is "global, i need cni to not change for a second" lock.

pkg/domain/infra/abi/network.go
```diff
@@ -115,6 +116,14 @@ func (ic *ContainerEngine) NetworkCreate(ctx context.Context, name string, optio
 		err      error
 		fileName string
 	)
+	config, err := ic.Libpod.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	// Acquire a lock for CNI
+	if _, err := storage.GetLockfile(filepath.Join(config.Engine.TmpDir, network.LockFileName)); err != nil {
```
vrothberg, line 124: This does not yet acquire the lock.

`storage.GetLockfile` returns a `Locker, error` and we currently ignore the `Locker`.

To acquire the lock, we need to:
```golang
lock, err := storage.GetLockfile(...)
...
lock.Lock()
defer lock.Unlock()
```

pkg/network/network.go
```diff
@@ -167,7 +169,11 @@ func ValidateUserNetworkIsAvailable(config *config.Config, userNet *net.IPNet) e
 
 // RemoveNetwork removes a given network by name.  If the network has container associated with it, that
 // must be handled outside the context of this.
-func RemoveNetwork(config *config.Config, name string) error {
+func RemoveNetwork(config *config.Config, name string, lockPath string) error {
+	// Acquire a lock for network removal
+	if _, err := storage.GetLockfile(filepath.Join(lockPath, LockFileName)); err != nil {
+		return err
+	}
```
vrothberg, line 176: Same issue as above.

I think it's _very_ risky to use the lock here _and_ outside of the package. It suggests that code outside of the package (i.e., `createMacVLAN` and `createBridge`) that needs the lock should rather live inside the package. Callers shouldn't need to worry about concurrency issues; the risk of missing to lock or running into deadlocks is high.

baude: yeah, i waffled on this as well.  I'm not sure I would categorize the risk similar but it is a fair comment.  I'll consider a move.

vrothberg (commented):
LGTM, nice work!

baude:
merge me

PR #7944, opened 2020-10-06
Cirrus: Fix obtaining a CI VM

cevich:
Also removed automatic exection of setup_environment.sh since most
people using this script are podman developers (not automation/CI
folks).  If executing the automation scripts is necessary, manual
attendance to required variables like `$TEST_FLAVOR` is mandatory.

### DIFF
--- a/hack/get_ci_vm.sh
+++ b/hack/get_ci_vm.sh
@@ -1,49 +1,82 @@
 #!/usr/bin/env bash
 
+#
+# For help and usage information, simply execute the script w/o any arguments.
+#
+# This script is intended to be run by podman developers who need to debug
+# problems specifically related to Cirrus-CI automated testing.  However,
+# because it's only loosely coupled to the `.cirrus.yml` configuration, it must
+# orchestrate VMs in GCP directly.  This means users need to have
+# pre-authorization (access) to manipulate google-cloud resoures.  Additionally,
+# there are no guarantees it will remain in-sync with other automation-related
+# scripts.  Therefore it may not always function for everybody in every
+# future scenario without updates/modifications/tweaks.
+
 set -e
 
-RED="\e[1;36;41m"
-YEL="\e[1;33;44m"
+RED="\e[1;31m"
+YEL="\e[1;32m"
 NOR="\e[0m"
 USAGE_WARNING="
-${YEL}WARNING: This will not work without local sudo access to run podman,${NOR}
-         ${YEL}and prior authorization to use the libpod GCP project. Also,${NOR}
-         ${YEL}possession of the proper ssh private key is required.${NOR}
+${YEL}WARNING: This will not work without podman,${NOR}
+         ${YEL}and prior authorization to use the libpod GCP project.${NOR}
 "
-# TODO: Many/most of these values should come from .cirrus.yml
+# These values come from .cirrus.yml gce_instance clause
 ZONE="${ZONE:-us-central1-a}"
 CPUS="2"
 MEMORY="4Gb"
 DISK="200"
 PROJECT="libpod-218412"
 GOSRC="/var/tmp/go/src/github.com/containers/podman"
-GCLOUD_IMAGE=${GCLOUD_IMAGE:-quay.io/cevich/gcloud_centos:latest}
-GCLOUD_SUDO=${GCLOUD_SUDO-sudo}
+GIT_REPO="https://github.com/containers/podman.git"
+
+# Container image with necessary runtime elements
+GCLOUD_IMAGE="${GCLOUD_IMAGE:-docker.io/google/cloud-sdk:alpine}"
+GCLOUD_CFGDIR=".config/gcloud"
+
+SCRIPT_FILENAME=$(basename ${BASH_SOURCE[0]})
+HOOK_FILENAME="hook_${SCRIPT_FILENAME}"
 
 # Shared tmp directory between container and us
-TMPDIR=$(mktemp -d --tmpdir $(basename $0)_tmpdir_XXXXXX)
+TMPDIR=$(mktemp -d --tmpdir ${SCRIPT_FILENAME}_tmpdir_XXXXXX)
 
-LIBPODROOT=$(realpath "$(dirname $0)/../")
+show_usage() {
+    echo -e "\n${RED}ERROR: $1${NOR}"
+    echo -e "${YEL}Usage: $SCRIPT_FILENAME <image_name>${NOR}"
+    echo ""
+    if [[ -r ".cirrus.yml" ]]
+    then
+        echo -e "${YEL}Some possible image_name values (from .cirrus.yml):${NOR}"
+        image_hints
+        echo ""
+        echo -e "${YEL}Optional:${NOR} If a $HOME/$GCLOUD_CFGDIR/$HOOK_FILENAME executable exists during"
+        echo "VM creation, it will be executed remotely after cloning"
+        echo "$GIT_REPO. The"
+        echo "current local working branch name and commit ID, will be provided as"
+        echo "it's arguments."
+    fi
+    exit 1
+}
+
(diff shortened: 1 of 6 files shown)


cevich:
@edsantiago when you have a moment, if you could give this a try?  I spent quite a lot of time modernizing this (still) hacky script to run entirely using rootless podman.  No more sudo requirement :smiley:  I haven't tested out the "new user" one-time setup experience, so I'm curious how it works (or fails) for you.

edsantiago:
After ~20m of pulling a 700M image, it just errors out (lots of repetitions of the same thing). The relevant part seems to be:
```
There was a problem refreshing your current auth tokens: ('invalid_grant: Bad Request
```
I don't see a README or obvious code comments explaining what to do, so I'll try again in the morning. Thanks for tackling this.

edsantiago:
Friendly suggestion:
```diff

diff --git a/hack/get_ci_vm.sh b/hack/get_ci_vm.sh
index ab667aacf..2e9a95cf1 100755
--- a/hack/get_ci_vm.sh
+++ b/hack/get_ci_vm.sh
@@ -165,10 +165,24 @@ mkdir -p $TMPDIR/.ssh
 mkdir -p {$HOME,$TMPDIR}/.config/gcloud/ssh
 chmod 700 {$HOME,$TMPDIR}/.config/gcloud/ssh $TMPDIR/.ssh

-# Attempt to determine if named 'libpod' gcloud configuration exists
-showrun $PGCLOUD info > $TMPDIR/gcloud-info
-if egrep -q "Account:.*None" $TMPDIR/gcloud-info
-then
+# Returns true if user has run an 'init' and has a valid token for this gcloud
+function has_valid_token() {
+    # Attempt to determine if named 'libpod' gcloud configuration exists
+    # (FIXME: what does this comment mean anyway?)
+    if $PGCLOUD info | egrep -q 'Account:.*None'; then
+        return 1
+    fi
+
+    # It's possible for 'gcloud info' to list expired credentials,
+    # e.g. 'ERROR:  ... invalid grant: Bad Request'
+    if $PGCLOUD auth print-access-token |& grep ERROR; then
+        return 1
+    fi
+
+    return 0
+}
+
+if ! has_valid_token; then
     echo -e "\n${YEL}WARNING: Can't find gcloud configuration for libpod, running init.${NOR}"
     echo -e "         ${RED}Please choose \"#1: Re-initialize\" and \"login\" if asked.${NOR}"
     showrun $PGCLOUD init --project=$PROJECT --console-only --skip-diagnostics
```

Unfortunately I'm still not going to be able to use or even test this script: the `Transferring tarball` step currently was claiming a 56-minute ETA, so I ^C'ed.

cevich:
@edsantiago thanks for taking a peek...

> so I'll try again in the morning

Did that message stop spewing after trying again? Your latest comment says you got to the "upload" stage, so that makes me think you progressed pass the gcloud initialization.

hmmm, so it seems there's some big assumptions on available network bandwidth...Valentine hit this as well, but it was more about the VM being located in the us-central zone, and so hard to work with.  I found a workaround there, and the image-download should be a one-time cost.  The upload part is quite a bit trickier though :confused: 

I've found it immensely useful to have a copy of my current local repo. on the remote VM to play with.  Combined with all the CI-env. var. work the script does for you, makes it really easy to tinker on automation-specific stuff, but outside of Cirrus-CI.  Perhaps if instead of uploading a full tarball, the script could leverage the VM's fat-pipe.  For example, by initially cloning the podman repo. directly, and then somehow only uploading the developer's diff with their local changes.

Any suggestions how to do that?  Could a `git push` from the developers workstation, be used to update the clone on the VM?  I'm worried the script would need to make some big assumptions about the local state, like what branch it's on.

edsantiago:
>Did that message stop spewing after trying again?

No. See my diff above; I had to instrument the code so it would detect "bad token" and run the re-init code. Only after that did I get to tar-upload.

>Any suggestions how to [push local git tree, changes-only]

Wow, interesting question. It sounds really tricky, figuring out branch state, local changes (committed and uncommitted), and dealing with `gcloud ssh/scp` on top of all that. No, I have no suggestions, sorry!

cevich:
> Wow, interesting question. It sounds really tricky, figuring out branch state, local changes (committed and uncommitted), and dealing with gcloud ssh/scp on top of all that. No, I have no suggestions, sorry!

I can get the VM's external IP address, but without monkeying with copying the developers ssh public keys, or using an (insecure) shared key, setting up normal (non-gcloud) `ssh` is a PITA :frowning_face: 

Hmmm...Yeah, tricky...maybe it's  possible to partially solve this.  I can do a repo clone easy enough on the VM, that would provide a starting place.  Then perhaps it's okay to just rely on the developer to get the remote code in-line with their needs.  Let me try that on for size...

cevich:
@edsantiago okay how about this, the only initial setup it does now is a full clone of the upstream repository on the VM.  Oh, and I added your suggested "bad token" fix as well.

edsantiago:
Hmmm, how about adding an early `podman pull` of the monster-huge image? 

Reason: I just tried running the script, and it just hung. Likely reason is that my podman testing had rmi'ed the monster image, so there was a pull going on that I didn't see. I have it cached (since last night), so I just ^C'ed and `podman load`ed, and all is working now. But the hang gave no hint of what was going on.

cevich:
Good idea...

cevich:
Updated:  Found a 7x smaller gcloud image to use :smile: added suggested `podman pull...` and also added ability for some personal VM customization through a "hook" script.  @edsantiago I think you'll find this version ***MUCH*** faster :grinning:

hack/get_ci_vm.sh
```diff
@@ -1,49 +1,81 @@
...
-PGCLOUD="$GCLOUD_SUDO podman run -it --rm -e AS_ID=$UID -e AS_USER=$USER --security-opt label=disable -v $TMPDIR:$HOME -v $HOME/.config/gcloud:$HOME/.config/gcloud -v $HOME/.config/gcloud/ssh:$HOME/.ssh -v $LIBPODROOT:$LIBPODROOT $GCLOUD_IMAGE --configuration=libpod --project=$PROJECT"
-SCP_CMD="$PGCLOUD compute scp"
+[[ "$UID" -ne 0 ]] || \
+    show_usage "Must execute script as a regular (non-root) user."
 
+[[ "${LIBPODROOT#$HOME}" != "$LIBPODROOT" ]] || \
+    show_usage "Clone of containers/podman must be a subdirectory of \$HOME ($HOME)"
+
+# Disable SELinux labeling to allow read-only mounting of repository files
+PGCLOUD="podman run -it --rm --security-opt label=disable -v $TMPDIR:$TMPDIR -v $HOME:/root:ro -v $HOME/.config/gcloud:/root/.config/gcloud -v $HOME/.config/gcloud/ssh:/root/.ssh -v $LIBPODROOT:$LIBPODROOT:ro $GCLOUD_IMAGE gcloud --configuration=libpod --project=$PROJECT"
```
edsantiago, line 72: Bind-mounting `$HOME` makes me queasy. Could you keep the old way (mounting a tmpdir), and change the convention so the hook lives in `.config/gcloud`?

cevich: Yeah, I'm not very comfortable with it either.  Yep, that's a fine place for the hook.

cevich:
@baude mind taking a peek at this.  It's working a bit different from before, in the interest of both speed and bandwidth use.  On the other hand, it's a lot more flexible with a new hook-script for every developer to customize the VM setup to their own needs.

edsantiago:
New image pulls in only two minutes, much nicer, thank you! The script is now working fine for me. LGTM!

cevich:
> The script is now working fine for me.

Fantastic, I'm glad to hear it.  This has ***REALLY*** come in handy for debugging automation scripting problems and helping reproduce those "it only happens in CI" problems.  Thanks for the review and helpful feedback.

PR #7945, opened 2020-10-07
[WIP] Still remove image even if getting parent fails

rhatdan:

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -627,7 +627,7 @@ func (i *Image) TopLayer() string {
 func (i *Image) Remove(ctx context.Context, force bool) error {
 	parent, err := i.GetParent(ctx)
 	if err != nil {
-		return err
+		logrus.Warnf("Failed to get image parent: %s", err)
 	}
 	if _, err := i.imageruntime.store.DeleteImage(i.ID(), true); err != nil {
 		return err


rhatdan:
Might fix: https://bugzilla.redhat.com/show_bug.cgi?id=1884043

vrothberg (commented):
Can you elaborate how this could cause https://bugzilla.redhat.com/show_bug.cgi?id=1884043?

mheon:
Are you sure reading the *parent* is the problem? I don't get that out of the BZ you linked. To me, it looks like the image he's trying to *remove* is missing a manifest, so we can't read it.

rhatdan:
I am guessing.  I was thinking that that function was failing before the image was attempted to be deleted.

rhatdan:
```
 podman image prune -a

Error:failed to prune image: error reading image 

"cc7639456e4a4c922cb086cf35f297427b3e8c38c14716533f67c66d70dd052a" as image: error locating item named "manifest" for image with ID "cc7639456e4a4c922cb086cf35f297427b3e8c38c14716533f67c66d70dd052a": file does not exist
```
I got got up in the error line above "error reading image", I did not see where removing the image would attempt to read, but the GetParent above might.

mheon:
We need to read the image to get a copy of it so we can remove it - `Remove()` is a method on an `Image` struct, so we need to `NewFromLocal()` first to get a copy of said image. I am betting it's `NewFromLocal` that is failing.

rhatdan:
Could be, this is one it would be a lot easier to diagnose if we had a reporter.  It looks like something got screwed up in the manifest, and we are blocking because of this, but should be able to just force remove it.

PR #7948, opened 2020-10-07
Fix ps port output

saschagrunert:
When defining multiple ports (same src/dst) like `-p 80:80 -p 443:443`
then podman will not show the complete output on `podman ps` (only
`0.0.0.0:80->80/tcp` in the example). This also applies to port ranges.

This patch refactors the port loop by pre-checking for ranges and
displaying them correctly to the end user.

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -371,12 +371,6 @@ func (l psReporter) CreatedHuman() string {
 // portsToString converts the ports used to a string of the from "port1, port2"
 // and also groups a continuous list of ports into a readable format.
 func portsToString(ports []ocicni.PortMapping) string {
-	type portGroup struct {
-		first int32
-		last  int32
-	}
-	portDisplay := []string{}
-
 	if len(ports) == 0 {
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


mheon:
FYI, Multiple host ports being mapped to the same container port is allowed - I disallowed it in Podman 2.0, and then we got a bug that it worked in 1.9, so it was re-added.

saschagrunert:
CI should succeed now, looks like a network flake:
```
SSL/TLS to mirror.umd.edu:443
Attempt 1 of 9 (retry on non-zero exit):
    + openssl s_client -quiet -no_ign_eof -connect mirror.umd.edu:443 
140063535470400:error:0200206E:system library:connect:Connection timed out:crypto/bio/b_sock2.c:110:
140063535470400:error:2008A067:BIO routines:BIO_connect:connect error:crypto/bio/b_sock2.c:111:
connect:errno=110
    exit(1)
    Failure! Sleeping 1.000000 seconds
Attempt 2 of 9 (retry on non-zero exit):
140052998735680:error:0200206E:system library:connect:Connection timed out:crypto/bio/b_sock2.c:110:
140052998735680:error:2008A067:BIO routines:BIO_connect:connect error:crypto/bio/b_sock2.c:111:
connect:errno=110
    exit(1)
    Failure! Sleeping 2.000000 seconds
Attempt 3 of 9 (retry on non-zero exit):
140471372003136:error:0200206E:system library:connect:Connection timed out:crypto/bio/b_sock2.c:110:
140471372003136:error:2008A067:BIO routines:BIO_connect:connect error:crypto/bio/b_sock2.c:111:
connect:errno=110
    exit(1)
    Failure! Sleeping 4.000000 seconds
Attempt 4 of 9 (retry on non-zero exit):
140427988354880:error:0200206E:system library:connect:Connection timed out:crypto/bio/b_sock2.c:110:
140427988354880:error:2008A067:BIO routines:BIO_connect:connect error:crypto/bio/b_sock2.c:111:
connect:errno=110
    exit(1)
    Failure! Sleeping 8.000000 seconds
```

Ready for review @containers/podman-maintainers PTAL

rhatdan:
@cevich Any idea what is going on with the CI system?

saschagrunert:
Tests are succeeding now.

cevich:
> CI should succeed now, looks like a network flake:

Dang I missed it.  Was it the `Ext. services` task that was failing or something else?

mheon:
VM Agent timed out. Was different than the example you linked in your email.

cevich:
Oh okay, that's "normal" / occasional / expected failure due to cloud-cost reduction measure on the Cirrus-CI side (i.e. nothing we can do about it).  It should have automatically re-tried (and presumably passed).  If it happens repeatedly...def. let me know.

saschagrunert:
> > CI should succeed now, looks like a network flake:
> 
> Dang I missed it. Was it the `Ext. services` task that was failing or something else?

Yes, it was mostly this one :) 

(to be honest I had to fix the e2e tests as well)

saschagrunert:
Found another bug, if the host range is different from the container range. Fixed this too :heavy_check_mark:

rhatdan:
LGTM
@vrothberg @mheon PTAL

rhatdan:
Lets get this reviewed.

test/e2e/ps_test.go
```diff
@@ -425,6 +425,31 @@ var _ = Describe("Podman ps", func() {
...
+			ALPINE, "top"},
+		)
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"ps", "--format", "{{.Ports}}"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.OutputToString()).To(ContainSubstring(
+			"0.0.0.0:3000-3001->3000-3001/tcp, 0.0.0.0:3100-3102->4000-4002/tcp, 0.0.0.0:6000-6066->6000-6066/tcp, 0.0.0.0:8000->8080/tcp, 0.0.0.0:30080->30080/tcp, 0.0.0.0:30443->30443/tcp",
+		))
```
cevich, line 450: This seems slightly fragile, as it seems conceivable the sorting could change over time.  For example, sorting ranges before single-port mappings could flip.  While meaningless to humans, the test maybe unnecessarily sensitive to the order.  Is the cost of output-checking each item separately, worth the benefit of a less pedantic (but still effective) test?

test/e2e/ps_test.go
```diff
@@ -425,6 +425,31 @@ var _ = Describe("Podman ps", func() {
 		Expect(session.OutputToString()).To(ContainSubstring("0.0.0.0:2000-2006"))
 	})
 
+	It("podman ps test with multiple port range", func() {
```
cevich, line 428: (Not sure if one already exists, didn't look) Please consider adding a small/quick negative test to confirm podman throws an error if an overlapping range is given.  For example `-p 1000-2000:2000-3000` with `-p 1999-2999:3001-4001` and `-p 3000-4000:4001:5001`.  Reason being, it would be disastrous if something like this slipped through, creating a difficult to debug ambiguity for users.

test/e2e/ps_test.go
saschagrunert: My idea was to test the sorting as well because we explicitly sort it in the code. The fix mainly targets for use cases where people add multiple port ranges to a single Podman command, which is right now broken. Not sure how to improve the test to be easier to reflect. Do you have any idea?

test/e2e/ps_test.go
saschagrunert: Sure, added a test like that. :+1:

saschagrunert:
Some recent changes broke the port mappings behavior…
```
>  [sudo] ./bin/podman run -it -p 6000-6066 alpine
Error: cannot listen on the TCP port: listen tcp4 :53: bind: address already in use
```

test/e2e/ps_test.go
cevich: Oh okay, that's fine if the intention is to also verify sorting.  In that case I would as you to consider putting that in a comment?  That would be really helpful to some future developer who manages to break both sorting and fail the test  :grin:

saschagrunert:
It's green, @containers/podman-maintainers PTAL

rhatdan:
LGTM
Lets get this in.

QiWang19 (commented):
LGTM

cmd/podman/containers/ps.go
```diff
@@ -440,22 +517,27 @@ func comparePorts(i, j ocicni.PortMapping) bool {
...
+		if first != last {
+			group = fmt.Sprintf("%s-%d", group, last)
+		}
+		return group
 	}
-	return fmt.Sprintf("%s/%s", group, groupType)
+	hostGroup := group(firstHost, lastHost)
+	ctrGroup := group(firstCtr, lastCtr)
+
+	return fmt.Sprintf("%s:%s->%s/%s", ip, hostGroup, ctrGroup, groupType)
```
mheon, line 542: I think this bit handles an empty IP incorrectly - looks like it will still print the `:` after it even if there is no IP set.

saschagrunert: We only pass the `fmt.Sprintf("%s/%s", hostIP, first.Protocol)` to this function (line 496), which contains the `hostIP`. The `hostIP` is pre-checked for an empty string and still set to `0.0.0.0` (477-480).

Issue #7934, opened 2020-10-06
REST APIv2 - rootless - hostname not added to /etc/hosts


psakar:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Container created via REST APIv2 is unable to resolve its hostname because the assigned hostname is missing in container /etc/hosts

**Steps to reproduce the issue:**

```
# start podman REST API
podman --log-level=debug system service -t 0 tcp:localhost:8880 2>&1>/tmp/podman-rest.log &

curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/create?name=test' --data '{"name":"test","Image":"localhost/test:latest", "Cmd":["/bin/sh", "-c", "tail -2 /etc/hosts"]}'
curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/test/start'
curl --output - -H "Content-Type: application/json" "http://127.0.0.1:8880/containers/test/logs?stdout=true&stderr=true"
```

**Describe the results you received:**
Last two lines from file /etc/hosts from my laptop

**Describe the results you expected:**
```
# used by slirp4netns
10.0.2.100	3ff0b3519d00 test
```
(Value 3ff0b3519d00 would be different ...)

**Additional information you deem important (e.g. issue happens only occasionally):**
May be because netmode is not set ? Applying following patch fixes the issue (most probably wrong, just to test)
```
diff --git a/pkg/api/handlers/compat/containers_create.go b/pkg/api/handlers/compat/containers_create.go
index 0579da8de..6b92935b9 100644
--- a/pkg/api/handlers/compat/containers_create.go
+++ b/pkg/api/handlers/compat/containers_create.go
@@ -144,7 +144,7 @@ func makeCreateConfig(ctx context.Context, containerConfig *config.Config, input
 		IPAddress:    "",
 		LinkLocalIP:  nil, // docker-only
 		MacAddress:   input.MacAddress,
-		// NetMode:      nil,
+		NetMode:      namespaces.NetworkMode("slirp4netns"),
 		Network:      input.HostConfig.NetworkMode.NetworkName(),
 		NetworkAlias: nil, // docker-only now
 		PortBindings: input.HostConfig.PortBindings,
```

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.13.15
Built:        Wed Sep 30 21:51:15 2020
[...]

PR #7952, opened 2020-10-07
APIv2 compatibility rootless network fix

baude:
when using the compatibility mode as rootless, containers that were created were not setting their host names correctly due to the netmode not being set.

Fixes: #7934

### DIFF
--- a/pkg/api/handlers/compat/containers_create.go
+++ b/pkg/api/handlers/compat/containers_create.go
@@ -14,8 +14,10 @@ import (
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/api/handlers/utils"
 	"github.com/containers/podman/v2/pkg/namespaces"
+	"github.com/containers/podman/v2/pkg/rootless"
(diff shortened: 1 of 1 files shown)


mheon:
Can you add a test?

baude:
> Can you add a test?

ewww .... i dont know if we test rootless calls directly.  @edsantiago any idea?

edsantiago:
Uh-oh, looks like we lost `test/apiv2` in the Cirrus switchover:
```diff
-    apiv2_test_script: '$SCRIPT_BASE/apiv2_test.sh |& ${TIMESTAMP} | ${LOGFORMAT} apiv2_test'
```
@cevich is that easy for you to fix? Maybe add it into one of the local tests?

Anyhow, once that's resolved, this could be tested via code in `test/apiv2`.

jwhonce:
Fix LGTM, now for tests and CI updates...

baude:
the missing test run directly impacts this.  @edsantiago would you feel ok about merging this and adding the test when we have that test suite running again?

Issue #7751, opened 2020-09-23
Detaching from container doesn't print newline


mohd-akram:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. Run `podman run -it fedora /bin/bash`

2. Press `ctrl-p` `ctrl-q` to detach

**Describe the results you received:**

```
[root@6aef033eef57 /]# [user@host ~]$
```

**Describe the results you expected:**

```
[root@6aef033eef57 /]# read escape sequence
[user@host ~]$
```

**Additional information you deem important (e.g. issue happens only occasionally):**

On Docker, the additional message is printed to stderr.

**Output of `podman version`:**

```
2.0.6
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.19-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.19, commit: 5dce9767526ed27f177a8fa3f281889ad509fea7'
  cpus: 8
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: razrlab-am
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.7-200.fc32.x86_64
  linkmode: dynamic
  memFree: 679784448
  memTotal: 15745339392
  ociRuntime:
    name: crun
    package: crun-0.14.1-5.fc32.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      commit: 598ea5e192ca12d4f6378217d3ab1415efeddefa
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.1.4-1.fc32.x86_64
    version: |-
      slirp4netns version 1.1.4
      commit: b66ffa8e262507e37fca689822d23430f3357fe8
      libslirp: 4.3.1
[...]

PR #7953, opened 2020-10-07
Container detach message (Docker Consistency)

ParkerVR:
For https://github.com/containers/podman/issues/7751

After a 'docker attach', <ctrl+p> <ctrl+q> detaches with message 'read escape sequence'. This was not occurring for me on 'docker run -it' or 'podman run -it' so I modified the podman behavior so this occurs only on attach as well.

### DIFF
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -588,6 +588,9 @@ func (ic *ContainerEngine) ContainerAttach(ctx context.Context, nameOrID string,
 	if err != nil && errors.Cause(err) != define.ErrDetach {
 		return errors.Wrapf(err, "error attaching to container %s", ctr.ID())
 	}
+	// This write maintains compatibility with Docker on <ctrl+p> <ctrl+q> detach:
+	// 'run -it x' does not newline but 'attach x' prints this message and newlines
+	os.Stdout.WriteString("read escape sequence\n")
 	return nil
 }
 


pkg/domain/infra/abi/containers.go
```diff
@@ -312,7 +312,7 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
 					reports = append(reports, &report)
 					continue
 				}
-				report.Err = errors.Wrapf(err, "failed to evict container: %q", id)
+				report.Err = errors.Wrapf(err, "Failed to evict container: %q", id)
```
jwhonce, line 315: error elements should start with lowercase not uppercase. If you need this exception for compatibility please add a comment.

pkg/domain/infra/abi/containers.go
```diff
@@ -582,6 +582,7 @@ func (ic *ContainerEngine) ContainerAttach(ctx context.Context, nameOrID string,
 	if err != nil && errors.Cause(err) != define.ErrDetach {
 		return errors.Wrapf(err, "error attaching to container %s", ctr.ID())
 	}
+	os.Stdout.WriteString("read escape sequence\n")
```
jwhonce, line 585: I would add a comment why this is happening.

jwhonce (changes requested):

pkg/domain/infra/abi/containers.go
```diff
@@ -604,7 +605,7 @@ func checkExecPreserveFDs(options entities.ExecOptions) error {
 	if options.PreserveFDs > 0 {
 		entries, err := ioutil.ReadDir("/proc/self/fd")
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "unable to read /proc/self/fd")
```
rhatdan, line 608: Doesn't ioutil give you this information in ReadDir

rhatdan:
```
package main

import (
	"fmt"
	"io/ioutil"
)

func main() {
	_, err := ioutil.ReadDir("/tmp/missing")
	fmt.Println(err)
}
```
go run /tmp/t.go 
open /tmp/missing: no such file or directory

rhatdan:
@ParkerVR Needs a rebase.

ParkerVR:
determined necessary, an alternative method was used (https://github.com/containers/podman/pull/7955)

Issue #7659, opened 2020-09-16
podman save capability to remove signatures


jdockter:
/kind feature

**Description**
When using podman save, via opm tool (which just shells out to podman), we have no way to save images that are signed.
```
sudo opm registry add -b "docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373" -d index.db -c podman
...
...
INFO[0004] running podman save                           bundles="[docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373]"
ERRO[0004] Error: unable to save "docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373": Error copying image to the remote destination: Can not copy signatures: Storing signatures for docker tar files is not supported  bundles="[docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373]"
ERRO[0004] permissive mode disabled                      bundles="[docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373]" error="error loading bundle from image: error saving image: Error: unable to save \"docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373\": Error copying image to the remote destination: Can not copy signatures: Storing signatures for docker tar files is not supported\n. exit status 125"
Error: error loading bundle from image: error saving image: Error: unable to save "docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373": Error copying image to the remote destination: Can not copy signatures: Storing signatures for docker tar files is not supported
. exit status 125
```

At just a podman level the same is seen.
```
podman save -o test.tar docker.io/ibmcom/ibm-cp-integration-bundle@sha256:47461a7543c71108a578178efa11507910eaf895113aab3612bf21300e598373
[...]

PR #7956, opened 2020-10-07
Allow save image remove-signatures

QiWang19:
remove signatures in podman save since the image formats do not support signatures
Close: #7659 


docs/source/markdown/podman-save.1.md
```diff
@@ -48,6 +48,10 @@ Allow for creating archives with more than one image.  Additional names will be
 
 Suppress the output
 
+**--remove-signatures**
+
+Do not save signatures, if any, from source image. This is necessary when saving a signed image to a format which does not support signatures. The default is **true**.
```
TomSweeneyRedHat, line 53:
```suggestion
Do not save signatures, if any, from the source image. This is necessary when saving a signed image to a format that does not support signatures. The default is **true**.
```

TomSweeneyRedHat:
Tests and bash completions?

QiWang19:
bash completions added. I am figuring out how to test this.

jwhonce (changes requested):
@QiWang19 You have new option coded as available to remote users but not coded in the tunnel or bindings packages

test/e2e/save_test.go
```diff
@@ -116,6 +119,48 @@ var _ = Describe("Podman save", func() {
...
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"pull", "--tls-verify=false", "localhost:5000/alpine"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		outfile := filepath.Join(podmanTest.TempDir, "temp.tar")
+		save := podmanTest.Podman([]string{"save", "remove-signatures=true", "-o", outfile, "localhost:5000/alpine"})
+		save.WaitWithDefaultTimeout()
```
jwhonce, line 160: Can you verify that signatures are not included?

QiWang19: I haven't found an easy way to verify that. Needs config the registries.d/default.yaml to and gpg keys to verify the signature. Updated the test but not sure if that can work.

test/e2e/save_test.go
```diff
@@ -116,6 +120,71 @@ var _ = Describe("Podman save", func() {
 		Expect(save).To(ExitWithError())
 	})
 
+	It("podman save remove signature", func() {
+		SkipIfRemote("podman-remote doesn't suppport tls-verify option")
```
rhatdan, line 124: Podman remote now supports tls-verify.

rhatdan (approved):
When would a user ever specify --remove-signatures=false?

QiWang19:
> When would a user ever specify --remove-signatures=false?

The user will use  --remove-signatures=false, current image format doesn't support pushing signatures to it.

test/e2e/save_test.go
QiWang19: podman remove push --tls-verify works, but still see `unknown flag: --tls-verify` with podman pull,
in the test **int remote ubuntu-19 root host**
```
[+2546s] Running: /var/tmp/go/src/github.com/containers/podman/bin/podman-remote --remote --url unix:/run/podman/podman-1695fe65d29f17c37c0cd8754dbdd796e29036e2adea66ec67ab760001892f06.sock --remote --url unix:/run/podman/podman-1695fe65d29f17c37c0cd8754dbdd796e29036e2adea66ec67ab760001892f06.sock pull --tls-verify=false --signature-policy=sign/policy.json localhost:5000/alpine
[+2546s] Error: unknown flag: --tls-verify
```

rhatdan: I opened a PR to fix this.

https://github.com/containers/podman/pull/8042

rhatdan:
@QiWang19 Shouldn't the code do this automatically, IE Does the code know which image stores support signatures?  Would the user just get an error telling them to remove signatures and now they need to run the command again with the flag?  Does not seem like a good UI experience?

QiWang19:
On the suggestion under the issue https://github.com/containers/podman/issues/7659#issuecomment-694121445, the remove-signatures defaults to true and avoids the user from getting the error.

rhatdan:
@QiWang19 Sure,  but under what circumstance would the user ever set this to False.  When would the user not want to remove-signatures?  Bottom line is there a situation where a user would ever type in 
--remove-signatures=false or true?  
If we want signatures to be pushed to a registry?

QiWang19:
If we want signatures to be pushed to a registry, `podman push` also has `remove-signatures` defaults to false. This patch only applies to `podman save`, current format (oci-archive, oci-dir, docker-dir) doesn't support save with signatures.

rhatdan:
When would I ever do a 
podman save --remove-signatures=false?

QiWang19:
> When would I ever do a
> podman save --remove-signatures=false?

Remove the option and remove the signature in the code, the current image format doesn't support saving signatures.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
Would like a head nod from @jwhonce before merging as he'd asked for a change.

jwhonce (approved):

PR #7957, opened 2020-10-07
Cirrus: rootless system tests: local, not remote

edsantiago:
The new CI setup (accidentally?) ran rootless system tests
with podman-remote instead of local, and only on fedora.
This PR runs rootless system tests using podman local,
and on all four flavors (fedora + ubuntu).

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -117,7 +117,7 @@ passthrough_envars(){
     for envname in $(awk 'BEGIN{for(v in ENVIRON) print v}' | \
                          grep -Ev "SETUP_ENVIRONMENT" | \
                          grep -Ev "$SECRET_ENV_RE" | \
-                         grep -E "$PASSTHROUGH_ENV_RE"); do
+                         grep -E "$PASSTHROUGH_ENV_RE" | sort); do
 
             envval="${!envname}"
             [[ -n $(tr -d "$xchars" <<<"$envval") ]] || continue
@@ -175,16 +175,17 @@ setup_rootless() {
     # transferred to the test-user.
     msg "Configuring rootless user's environment variables:"
 
+    envfile="/home/$ROOTLESS_USER/cirrus-task-environment"
     (
         echo "# Added by ${BASH_SOURCE[0]} ${FUNCNAME[0]}()"
         echo "export SETUP_ENVIRONMENT=1"
-    ) >> "/home/$ROOTLESS_USER/.bashrc"
+    ) > $envfile
 
     while read -r env_var_val; do
-        echo "export $env_var_val" >> "/home/$ROOTLESS_USER/.bashrc"
+        echo "export $env_var_val" >> $envfile
     done <<<"$(passthrough_envars)"
-    chown $ROOTLESS_USER:$ROOTLESS_USER "/home/$ROOTLESS_USER/.bashrc"
-    cat "/home/$ROOTLESS_USER/.bashrc" | indent 2
+    chown $ROOTLESS_USER:$ROOTLESS_USER $envfile
+    indent 2 <$envfile
 
     msg "Ensure the systems ssh process is up and running within 5 minutes"
     systemctl start sshd
@@ -192,6 +193,37 @@ setup_rootless() {
            -o UserKnownHostsFile=/dev/null \
            -o StrictHostKeyChecking=no \
            -o CheckHostIP=no true
+}
+
+# This should be the first thing run from any test. It's only really
+# applicable for rootless, because we've ssh'ed in and don't have a
+# working set of environment variables.
+load_cirrus_task_environment() {
+    envfile="$HOME/cirrus-task-environment"
+
+    # The usual case: we already have the environment we need
+    if [ ! -e $envfile ]; then
+        return
+    fi
+
+    # We have an env file; but make sure we read it only once
+    if [ -n "$_initialized_cirrus_task_environment" ]; then
+        return
+    fi
+
+    # First time through. This implies that we're not root and that
+    # we have no environment. But, for sanity's sake, confirm. If
+    # a canary envariable is defined, we don't want to override.
+    # shellcheck disable=SC2154
+    if [ -n "$TEST_FLAVOR" ]; then
+        msg "*** WARNING! load_cirrus_task_environment: \$TEST_FLAVOR is already defined ($TEST_FLAVOR), before reading $envfile! Bad things may happen!"
+    fi
+
+    # shellcheck disable=SC1090
+    . $envfile
+
+    # Just in case runner.sh (for what reason??) tries to call us again.
+    _initialized_cirrus_task_environment=1
 }
 
 install_test_configs() {
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -247,6 +247,9 @@ dotest() {
 }
 
 msg "************************************************************"
+# BEFORE DOING ANYTHING AT ALL: make sure we have the env. needed for our job
+load_cirrus_task_environment
+
 # Required to be defined by caller
 # shellcheck disable=SC2154
 msg "Runner executing $TEST_FLAVOR $PODBIN_NAME-tests as $PRIV_NAME on $DISTRO_NV($OS_REL_VER)"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -539,12 +539,12 @@ rootless_system_test_task:
     alias: rootless_system_test
     depends_on:
       - rootless_integration_test
-    matrix: *fedora_vm_axis
+    matrix: *platform_axis
     gce_instance: *standardvm
     env:
         TEST_FLAVOR: sys
         PRIV_NAME: rootless
-        PODBIN_NAME: remote
+        PODBIN_NAME: podman
     clone_script: *noop  # Comes from cache
     gopath_cache: *ro_gopath_cache
     setup_script: *setup


edsantiago:
Map showing differences (top right): 
cirrus-map-pr7957

edsantiago:
@cevich I'm stumped - ubuntu tests are [failing because environment variables are lost over ssh](https://github.com/containers/podman/pull/7957/checks?check_run_id=1222587361). That doesn't confuse me - what confuses me is, how can this possibly work on Fedora?

edsantiago:
[Found it](https://cirrus-ci.com/task/4708224811139072?command=setup#L183-L186):
```bash
    # If not running interactively, don't do anything
    case $- in
        *i*) ;;
          *) return;;
    esac
```
Well, that's annoying. Fixing it is going to require the VM dance, so it'll have to wait until next week.

edsantiago:
@cevich PTAL (low priority). CI is still running but it seems to have passed "sys podman ubuntu-* rootless host", the tests in question. Even so, there's always something I've missed when dealing with CI...

cevich:
> Even so, there's always something I've missed when dealing with CI...

Have no fear, I've committed an entire hemisphere of my brain to remembering all the moving parts :grinning: 

I believe the other missing piece here is `/etc/environment`.  We drop all our important env. vars into there during `setup_environment.sh` and then hope and pray it gets loaded just in the nick of time.  It's easily possible I forgot something.  

The non-interactive thing you found is also real problem.  TBH I very much don't care for the way Ubuntu does shell setup.

In any case, loading `lib.sh` guarantees `/etc/environment` is loaded, so as long as the needed env. vars are set in there, it should be all well and good once `runner.sh` kicks in.

cevich:
Oh there's code here...lemmie peek at that too...

contrib/cirrus/lib.sh
```diff
@@ -194,6 +195,37 @@ setup_rootless() {
...
 }
 
+# This should be the first thing run from any test. It's only really
+# applicable for rootless, because we've ssh'ed in and don't have a
+# working set of environment variables.
+load_cirrus_task_environment() {
+    envfile="$HOME/cirrus-task-environment"
+
+    # The usual case: we already have the environment we need
+    if [ ! -e $envfile ]; then
```
cevich, line 205: Would this be an improvement:

`if [[ ! -e $envfile ]] && ((SETUP_ENVIRONMENT)); then`

contrib/cirrus/runner.sh
```diff
@@ -247,6 +247,9 @@ dotest() {
 }
 
 msg "************************************************************"
+# BEFORE DOING ANYTHING AT ALL: make sure we have the env. needed for our job
+load_cirrus_task_environment
+
```
cevich, line 252: This will be called multiple times...see around line 231 and possibly (rootless inside-container testing someday?) around line 116.   Maybe make the call conditional based on the current value of `$UID` and `$PRIV_NAME`?

cevich:
Hmmmm, trying to think of simplifying the (already complex) situation...What about dropping the loading function, and instead just have `lib.sh` always, conditionally load the user env. file?

We know that whenever we see `$UID != 0` and `$PRIV_NAME == rootless`, we're very likely executing `runner.sh` intentionally as a user for some rootless testing.  I cannot think of any current or possible future setup where that won't be the case.

contrib/cirrus/runner.sh
edsantiago: I'm really not sure how to interpret your feedback. Line 231 is literally the exact reason for all of this. So, of course we want `load_cirrus_task_environment` to be called in that case. And my invocation here is deliberate, because `runner.sh` is the common entry point and this `load` takes place before the subtask selection. 

To expand on my reasoning here: I found a bug in the current implementation. The current implementation adds environment variables to `.bashrc`, and this is a NOP on Ubuntu (not to mention bad practice, because it's a hidden side effect). My PR fixes that bug, by preserving necessary environment in a safe known place, then loading it at precisely the place where it's needed.

In reference to your later "have `lib.sh` always load it" comment, again I deliberately chose not to do that: `lib.sh` gets called from a lot of scripts, not all of which may want this side effect (but I haven't examined each one). I want `load_cirrus_task_environment` to be a deliberate intentional action requested by a script, in particular, by `runner.sh` which is the common delegating entry point. I want a future maintainer to see that and understand immediately that `runner.sh` is guaranteed to have the Cirrus envariables it relies on.

Could I ask you to take a fresh look? It's clear that one of us is missing something. If it's still me, could I ask you to reexplain your concerns with maybe more context that I may not be understanding?

cevich: > Could I ask you to take a fresh look? 

Of course, anytime...

cevich:
@edsantiago did you happen to take a peek at what actually ends up in the user's `$envfile`?  Assuming a VM is never used for more than one kind of test (it isn't) or style of test (it isn't) I'm wondering why everything can't simply exist in /etc/environment?  Otherwise I'm okay with your solution, as long as a simpler one is not possible within a reasonable amount of time.  Taking a quick poke at this with `get_ci_vm.sh`...

edsantiago:
>did you happen to take a peek at what actually ends up in the user's $envfile?

Only indirectly, by looking at the [debug output from setup_rootless()](https://cirrus-ci.com/task/5919914307354624?command=setup#L174). (Also indirectly, from the fact that CI passed, including Ubuntu rootless).

I considered using `/etc/environment` but quickly rejected that because there's already a lot of code in `setup_environment.sh` that appends to that file, but nothing that actually resets it. The script code also includes 'export', which according to my reasing of `environment(5)` is invalid. I just didn't want to go down that rabbit hole.

cevich:
...here's what I got:
```
[root@cevich-fedora-c4948709391728640 podman]# cat /home/some4292dude/cirrus-task-environment
# Added by contrib/cirrus/lib.sh setup_rootless()
export SETUP_ENVIRONMENT=1
export CGROUP_MANAGER=systemd
export CI=false
export CIRRUS_BASE_SHA=YOU_FOUND_A_BUG
export CIRRUS_BUILD_ID=861602520879
export CIRRUS_CI=false
export CIRRUS_REPO_NAME=podman
export CIRRUS_WORKING_DIR=/var/tmp/go/src/github.com/containers/podman
export DISTRO_NV=fedora-32
export GOCACHE=/var/tmp/go/cache/go-build
export GOPATH=/var/tmp/go
export GOSRC=/var/tmp/go/src/github.com/containers/podman
export OCI_RUNTIME=/usr/bin/crun
export PODBIN_NAME=podman
export PRIV_NAME=rootless
export ROOTLESS_USER=some4292dude
export SCRIPT_BASE=./contrib/cirrus
export TEST_ENVIRON=host
export TEST_FLAVOR=sys

[root@cevich-fedora-c4948709391728640 podman]# cat /etc/environment
export TEST_FLAVOR=sys
export PRIV_NAME=rootless
export PODBIN_NAME=podman
export TEST_ENVIRON=host
export DISTRO_NV=fedora-32
export OCI_RUNTIME=/usr/bin/crun
export CGROUP_MANAGER=systemd
export ROOTLESS_USER='some4292dude'
```

Notably missing are all of the other default `$CIRRUS_*` variables.  Even still, I can't think of a reason why having those set system-wide (`/etc/environment`) would cause problems.  The values never change within the scope/lifetime of the VM.  I also don't see any variables/values that need to be different on the VM when executing as root vs rootless.  The important one that must never get hard-coded is `$CONTAINER` but IIRC, we already filter for that.

So given the very limited and specific scope for a VM now (running one test in a single format) I think it's okay to just stick everything in `/etc/environment`.  I dunno about 'export' vs non in that file, so I assumed the worse-case (sub-process needs the var and doesn't get it).

cevich:
...Oh in fact, there's a bug.  `export SETUP_ENVIRONMENT=1` is suppose to be the 'canary' for `setup_environment.sh`.  It should be set in `/etc/environment' and isn't.  That's really only a minor thing though, protecting users of `get_ci_vm.sh` from doing a double-setup (will never happen under normal CI conditions).

edsantiago:
@cevich what I'm trying to say is that I believe the use of `/etc/environment` in these scripts is broken (the format is clearly documented as "VAR=VALUE"). I just don't feel like dealing with it right now because I just know that it will be another all-day or many-day expedition. My goal here was simply to get rootless tests working on Ubuntu.

cevich:
I'm okay with not fixing the use of `export` in `/etc/environment`, for now, that does seem likely to break things.  I also don't want a many-day expedition.  Are you willing to try re-using `/etc/environment`? If it doesn't work, then I'm okay with the more complex solution.

edsantiago:
I've just spent an hour trying to understand `/etc/environment`, and I give up. It's not going to get me where I need to go.

For posterity, my findings:
* variables defined in the file _are_ global (i.e. no `export` is needed); and
* `export` seems to be ignored (eg `export FOO=bar` sets `$FOO` in a login shell); and
* the file is _not consulted_ on `ssh` through a multiplexed session, but as expected this can be overridden via `ssh -o 'ControlMaster no' -o 'ControlPath /dev/null'`. (I do not expect that this would be a problem with the CI use of ssh); and
* I tested only on f32, and have no idea how this would work on Ubuntu.

I do believe that the current use of `/etc/environment` in `contrib/cirrus/setup_environment.sh` is improper and risky. It is owned by and intended for `pam`, it should not have `export` directives, and it should not be `sourced` by client code. But our CI setup is too fragile for me to dare touching any of that code. Since #7996 does most of what I want (enable system tests), and since I don't really care about Ubuntu, I'm just going to close this. I do hope some future maintainer fixes the environment mess, and maybe that'll be me some lazy day, but not now.

cevich:
> spent an hour trying to understand

A future maintainer could benefit from your research if it were encoded in the form of a jira card...even if unassigned.  Otherwise it's doubtful a future maintainer will remember to look in a closed PR for a "TODO"

PR #7964, opened 2020-10-08
short-name aliasing

vrothberg:
Based on https://github.com/containers/image/pull/1061 to allow for testing and playing around.


libpod/image/pull.go
```diff
@@ -401,18 +402,19 @@ func (ir *Runtime) pullGoalFromPossiblyUnqualifiedName(inputName string) (*pullG
...
-	refPairs := make([]pullRefPair, 0, len(searchRegistries))
-	for _, registry := range searchRegistries {
-		ref, err := decomposedImage.referenceWithRegistry(registry)
-		if err != nil {
-			return nil, err
-		}
-		imageName := ref.String()
-		srcRef, err := docker.ParseReference("//" + imageName)
+
+	candidates, err := shortnames.Resolve(nil, inputName, false)
```
mheon, line 406: If nothing else, I love that we finally have the resolution code in c/image so resolving short names will finally be consistent across all our tools.

vrothberg: I still hope we'll get a common image library done that Buildah, Podman and CRI-O can share.

vrothberg: s/library/package/ :)

vrothberg:
@baude @cevich ... it looks like the upper system tests were cancelled because the integration test of another $distro-$version flaked. I cannot restart them in the UI. Do you know how to restart them?

vrothberg:
Ah ! It seems they're restarting on their own, awesome!

edsantiago:
Sorry if this is a stupid question, but how are you planning to handle podman-remote? Right now it just prompts on the server side (which the remote user can never see). I've tried my best to search for 'remote' in this PR.

vrothberg:
> Sorry if this is a stupid question, but how are you planning to handle podman-remote? Right now it just prompts on the server side (which the remote user can never see). I've tried my best to search for 'remote' in this PR.

We don't (yet?) have special plans for remote. Depending on the configuration, the service endpoint may either fall back to trying out all search registries or it will error out.

Wiring prompting into the remote client would be quite a challenge. We had to do all the magic to remote-control the terminal for prompting _and_ change an unknown amount of interfaces in the code to pass down the remote stream through all code paths that may (or may not) reach short-name resolution.

vrothberg:

Note that you need to set `CONTAINERS_SHORT_NAME_ALIASING=on` until we move the main branch over to Podman 3.0 dev.

vrothberg:

Note that this PR vendors a non-release Buildah. My weekend is coming closer and we don't need to block until a Buildah release for this to be merged (or even for an RC).

mheon:
LGTM

vrothberg:
@rhatdan why did you re-wip?

rhatdan:
Because of the Buildah vendor.

rhatdan:
You could merge if you want, but we can not release until Buildah version bump.

vrothberg:
@rhatdan, we don't need to block this PR on the Buildah release. We can vendor Buildah again later. I'll be on PTO on Monday and don't want to block Podman.

PR #7965, opened 2020-10-08
Cirrus: Use google mirror for docker.io

cevich:
In Nov. 2020 Docker inc. will be limiting (in some unknown way) the rate of image pulls.  In order for this limit to not randomly break automated testing, or require extensive code changes, we need to introduce a transparent mirror.  Fortunately Google provides a transparent mirror of the docker hub, and we're already using GCP for all our test-VMs and containers.  Unfortunately, measuring the use/non-use of the mirror is very difficult w/o achieve without test-breaking operational changes.  However, based on test-failures discovered in the making of this PR, the mirror does appear to be in-use and transparent to (most) tests.

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -208,6 +208,7 @@ install_test_configs() {
     # as the default).  This config prevents allocation of network address space used
     # by default in google cloud.  https://cloud.google.com/vpc/docs/vpc#ip-ranges
     install -v -D -m 644 $SCRIPT_BASE/99-do-not-use-google-subnets.conflist /etc/cni/net.d/
+
     install -v -D -m 644 ./test/registries.conf /etc/containers/
 }
 
--- a/test/registries.conf
+++ b/test/registries.conf
@@ -1,10 +1,10 @@
 # Note that changing the order here may break tests.
-[registries.search]
-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
+unqualified-search-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
 
-[registries.insecure]
-registries = []
-
-#blocked (docker only)
-[registries.block]
-registries = []
+[[registry]]
+# In Nov. 2020, Docker rate-limits image pulling.  To avoid hitting these
+# limits while testing, always use the google mirror for qualified and
+# unqualified `docker.io` images.
+# Ref: https://cloud.google.com/container-registry/docs/pulling-cached-images
+prefix="docker.io"
+location="mirror.gcr.io"
--- a/contrib/cirrus/logcollector.sh
+++ b/contrib/cirrus/logcollector.sh
@@ -45,8 +45,10 @@ case $1 in
                     containernetworking-plugins \
                     containers-common \
                     criu \
+                    crun \
                     golang \
                     podman \
+                    runc \
                     skopeo \
                     slirp4netns \
         )
@@ -56,9 +58,7 @@ case $1 in
                 PKG_LST_CMD='rpm -q --qf=%{N}-%{V}-%{R}-%{ARCH}\n'
                 PKG_NAMES+=(\
                     container-selinux \
-                    crun \
                     libseccomp \
-                    runc \
                 )
                 ;;
             ubuntu*)
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -178,6 +178,8 @@ case "$TEST_FLAVOR" in
             remove_packaged_podman_files
             make install PREFIX=/usr ETCDIR=/etc
         fi
+
+        install_test_configs
         ;;
     vendor) make clean ;;
     release) ;;
--- a/test/e2e/manifest_test.go
+++ b/test/e2e/manifest_test.go
@@ -51,11 +51,13 @@ var _ = Describe("Podman manifest", func() {
 	})
 
 	It("podman manifest inspect", func() {
-		session := podmanTest.Podman([]string{"manifest", "inspect", BB})
+		// Something NOT on docker hub
+		fqin := "k8s.gcr.io/pause@sha256:f78411e19d84a252e53bff71a4407a5686c46983a2c2eeed83929b888179acea"
+		session := podmanTest.Podman([]string{"manifest", "inspect", fqin})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 
-		session = podmanTest.PodmanNoCache([]string{"manifest", "inspect", "docker.io/library/busybox"})
+		session = podmanTest.PodmanNoCache([]string{"manifest", "inspect", fqin})
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
 	})


cevich:
@vrothberg PTAL, I've succeeded in making podman really, really, really mad :grin: 

```
Error: unable to pull docker.io/library/alpine:latest: Error initializing source docker://alpine:latest: error loading registries 
configuration: error loading registries configuration "/var/tmp/go/src/github.com/containers/podman/test/registries.conf": 
mixing sysregistry v1/v2 is not supported
```

vrothberg:
`test/registries.conf` is in the old v1 format and you append v2 options to it.

`
[registries.search]                                                
registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
`
Should be turned into
`
unqualified-search-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
`

But I suggest to configure the mirror there as well. I see beauty in the CI and my local tests to execute the same code.

cevich:
> test/registries.conf is in the old v1 format and you append v2 options to it.

Interesting, that wasn't a line I changed.  I wonder if this `test/registries.conf` was maybe not even used before?

vrothberg:
> @vrothberg interesting, so I see there [are several tests which fail](https://cirrus-ci.com/task/5617039488319488), and the error mentions `mirror.gcr.io` but otherwise I don't see confirmation that images are being pulled from the mirror. For example:
> 
> ```
> [+0205s] Caching docker.io/library/alpine:latest at /tmp/alpine-latest.tar...Running: /var/tmp/go/src/github.com/containers/podman/bin/podman --root /tmp/crio --runroot /tmp/crio-run --runtime /usr/bin/crun --conmon /usr/bin/conmon --cni-config-dir /etc/cni/net.d --cgroup-manager systemd --tmpdir /tmp --events-backend file --storage-driver vfs pull docker.io/library/alpine:latest
>          Trying to pull docker.io/library/alpine:latest...


That's good news! Using mirrors is transparent to Podman, in the sense that Podman does not know about them. Whatever mirror we end up pulling from, the image will be tagged as the one initially referenced. So it's good that we see `mirror.gcr.io` in some errors as it indicates it's being used.

Note that the debug logs reveal which mirrors we pull from.

cevich:
> Note that the debug logs reveal which mirrors we pull from.

Ahh that is what I was hoping to learn - how to observe it.  I want to run at least once manually, to verify the mirror is in use.  Not because I don't believe you, but because I don't blindly trust my own work...

cevich:
@baude PTAL, I marked the one problematic test as skipped, referencing the issue I opened.

contrib/cirrus/lib.sh
```diff
@@ -208,6 +208,20 @@ install_test_configs() {
...
+    # use the globally installed file.
+    # Ref: https://cloud.google.com/container-registry/docs/pulling-cached-images
+    cat >> ./test/registries.conf <<EOF
+# Added by ${FUNCNAME[0]} in ${BASH_SOURCE[0]}
+[[registry]]
+location="docker.io"
+
+[[registry.mirror]]
+location="mirror.gcr.io"
+EOF
```
vrothberg, line 224: I prefer to change `./test/registries.conf` directly. This way, we use the same registry everywhere.

vrothberg (commented):
I'd love the system tests to use `test/registries.conf` as well. Can we export `REGISTRIES_CONF_PATH` in the Makefile?

Cc @edsantiago

contrib/cirrus/lib.sh
cevich: That's effectively what we get by installing the file system-wide (next line).  I too am not a big fan of having multiple potential locations for the configuration, as it makes maintenance and debugging difficult.  In a perfect world, I'd rather not rely on setting the env. var.  at all.  It's would be simpler if the e2e tests defaulted to using the installed system configuration.  I'm just not knowledgeable enough of their inner-workings to affect that change :disappointed:

cevich: oh, reading this again, you mean change the file instead of the script?  I was about to do that previously but was concerned that `mirror.gcr.io` was something special for google-cloud VMs.  Sure we can run that way...

cevich:
> which is a manifest list that isn't on Docker Hub?

Oh, good idea!

cevich:
Rebased / force-pushed with suggested changes.

rhatdan:
LGTM
@vrothberg PTAL

test/registries.conf
```diff
@@ -1,10 +1,20 @@
...
-[registries.search]
-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
+unqualified-search-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
 
 [registries.insecure]
 registries = []
 
 #blocked (docker only)
 [registries.block]
 registries = []
```
vrothberg, line 9:
```suggestion
```

test/registries.conf
```diff
@@ -1,10 +1,20 @@
...
+# In Nov. 2020, Docker rate-limits image pulling.  To avoid hitting these
+# limits while tesing, use the provided google cache.  Integration tests
+# make use of the copy of this file under `./test/`, however system tests
+# use the globally installed file.
+# Ref: https://cloud.google.com/container-registry/docs/pulling-cached-images
+[[registry]]
+location="docker.io"
+
+[[registry.mirror]]
+location="mirror.gcr.io"
```
vrothberg, line 20:
```suggestion
```

test/registries.conf
```diff
@@ -1,10 +1,20 @@
...
 [registries.block]
 registries = []
+
+# In Nov. 2020, Docker rate-limits image pulling.  To avoid hitting these
+# limits while tesing, use the provided google cache.  Integration tests
+# make use of the copy of this file under `./test/`, however system tests
+# use the globally installed file.
+# Ref: https://cloud.google.com/container-registry/docs/pulling-cached-images
+[[registry]]
+location="docker.io"
```
vrothberg, line 17: Assuming we don't want to pull from `docker.io` at all, we should do the following:
```suggestion
prefix="docker.io"
location="mirror.gcr.io"
```

If we configure `gcr` as mirror, we can still fall back to using docker.io.

test/registries.conf
cevich: oh, did you change your mind or did I misinterpret https://github.com/containers/podman/pull/7965#discussion_r507538219 ?

cevich: oh nvm, I see your comment below.

test/registries.conf
cevich: Oh...interesting idea, I like it.

cevich:
Rebased / force-pushed w/ Valentin's suggestions.

test/registries.conf
```diff
@@ -1,10 +1,9 @@
...
-[registries.insecure]
-registries = []
-
-#blocked (docker only)
-[registries.block]
-registries = []
+# In Nov. 2020, Docker rate-limits image pulling.  To avoid hitting these
+# limits while testing, always use the google mirror for qualified and
+# unqualified `docker.io` images.
+# Ref: https://cloud.google.com/container-registry/docs/pulling-cached-images
```
vrothberg, line 7: Isn't there a `[[registry]]` missing?

cevich: Easily possible, I may have misinterpreted your https://github.com/containers/podman/pull/7965#discussion_r509202113 comment.  Though...it seems to work without it, is that indication of a bug?

cevich:
@vrothberg rebased / force-pushed with the `[[registry]]` addition.

---

I've spent a lot of time considering the observation of these transparent operations.    It's really (***REALLY***) bugging me that using a mirror is **so** transparent.  In general things that function transparently, can also (necessarily) fail transparently.  It's possible that a single podman test encounters a bug, where the mirror should but is not used, eventually leading to a flake far-removed from it's cause :cry:  Consider:

1. A PR is merged that introduces a transparent mirror-pull related bug in containers/storage (randomly chosen).
2. Nobody notices because mirror-use is transparent, and some/most other podman tests are successfully use the mirror.
3. All subsequent podman PRs have mirroring transparently fallback to direct pull from `docker.io` for a one/handful of affected tests.  They all continue to pass, so nobody notices.
4. Some future podman PR on a busy Monday, encounters the `docker.io` rate-limit and fails with an error completely unrelated to the change subject.

While arguably this is an unlikely situation, the fact of the causal-disconnect could mean months or years pass before it's noticed (i.e. super-bad situation we must avoid).  Ideally we need the ability to assert on a (relevant) test-by-test basis, that the transparent mirror is in fact used when expected.  That's the only fool-proof (and automated) way of detecting a containers-code or test-bug relating to use/not of the mirror.

I propose addition of an env. var. to the lowest-level "image-pull" code-path(s) that when set, makes the output *slightly* more verbose about the host being contacted.  This small addition would be easy to unit-test, and something we can easily check across integration and system-tests in every containers repo.  For example:

```
$ export VERBOSE_PULL=1
$ podman pull docker.io/busybox:latest
Contacting mirror.gcr.io...
Trying to pull docker.io/library/busybox:latest...
Copying blob 26cc0bd3ca52 skipped: already exists
Copying blob a3ed95caeb02 done
...
```

@containers/podman-maintainers What do you think?

Note: "cevich is overly paranoid" is a valid answer :smile:

cevich:
@vrothberg tests are very much not happy after adding `[[registry]]`, [did I spell it wrong? ](https://github.com/containers/podman/pull/7965/files#diff-5d9f44e81404359a2b0bc17f7b98e8bb86ec381eab88b992a7abbf6b5d9e6ea0R4)

vrothberg:
> @vrothberg tests are very much not happy after adding `[[registry]]`, [did I spell it wrong? ](https://github.com/containers/podman/pull/7965/files#diff-5d9f44e81404359a2b0bc17f7b98e8bb86ec381eab88b992a7abbf6b5d9e6ea0R4)

`$ REGISTRIES_CONFIG_PATH=`pwd`/test/registries.conf podman --log-level=debug pull docker.io/alpine` ... works on my machine and it pulls from gcr.

vrothberg:
> I've spent a lot of time considering the observation of these transparent operations. It's really (_**REALLY**_) bugging me that using a mirror is **so** transparent. In general things that function transparently, can also (necessarily) fail transparently. 

I think that's a very fair point. It would definitely be more user friendly to display not only the reference we're pulling but also from where.

@mtrmac WDYT? I guess c/image is where we had to tackle it.

vrothberg:
@cevich it looks like we're hitting issues related to digests:
```
[+0011s] #  response: {"cause":"manifest unknown: Failed to fetch \"sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f\" from request \"/v2/library/alpine/manifests/sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f\".","message":"unable to pull docker.io/library/alpine@sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f: Error initializing source docker://alpine@sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f: Error reading manifest sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f in mirror.gcr.io/library/alpine: manifest unknown: Failed to fetch \"sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f\" from request \"/v2/library/alpine/manifests/sha256:acd3ca9941a85e8ed16515bfc5328e4e2f8c128caa72959a58a127b7801ee01f\".","response":500}
```
Looks like the image we're looking for does not exit on gcr. I think we need to have a look at each failure and see what's going on. In other words, your changes work but some images are missing on gcr.

Note that Docker Hub delayed the upcoming changes to mid '21. I think we should decouple this change (which is clearly working) from changing the tests. My proposal is to change the config to:
```
[[registry]]
location="docker.io"
[[registry.mirror]]
location="mirror.gcr.io"
```

That's what you had initially and it allows to falling back to `docker.io` if the image is not present on gcr. We can create a new work item to complete the migration.

@baude PTAL

cevich:
> We can create a new work item to complete the migration.

Okay, SGTM, and then I can drop the test "fix" from this PR as you suggested as well...

cevich:
@vrothberg errr, I just read [Note that Docker Hub delayed the upcoming changes to mid '21.](https://www.docker.com/blog/docker-hub-image-retention-policy-delayed-and-subscription-updates/).  At the end, there's a section that says they ***are*** going to enforce pull-limits (in less than a week) :cry: 

Assuming this is true, we need to keep the exclusive mirroring in place.  That implies (for this PR) we either skip the failing tests or fix them inline.  Is the digest-pulling problem caused by the mirror not retaining them or something?

mtrmac:
> > I've spent a lot of time considering the observation of these transparent operations. It's really (_**REALLY**_) bugging me that using a mirror is **so** transparent. In general things that function transparently, can also (necessarily) fail transparently.
> 
> I think that's a very fair point. It would definitely be more user friendly to display not only the reference we're pulling but also from where.

It’s a bit awkward in that even the generic `copy.Image` code doesn’t know about mirrors (in fact, the `Trying to pull` message comes from Podman, primarily to expose what short name search does).

What we would ideally want to do is to significantly extend the progress API to actually support a wide range of events, _and_ to allow individual transports to contribute events. That’s a fair bit of work, and API change, and all that. So a “short-term” (clearly forever actually) mirror-specific mini-feature would probably be more appropriate, as long as it is easy enough to maintain.

----

On c/image level, I’m rather unhappy with adding environment variables that impact behavior of a library in a way non-obvious to callers; a `types.SystemContext` option to log mirrors in _some_ way (increased Logrus level? just an `io.Writer`? a `chan DockerPullEndpointAttempt`?) would be fine with me, and if Podman wanted to set that based on an environment variable, fine.

(Alternatively, just running `podman --log-level=debug` should be cheap enough as far as the pull process is concerned, certainly compared to the network I/O, although just starting Podman does produce a fair bit of output.)

edsantiago:
I'm giving up for today, without actually having got anywhere. Current stumbling block is `busybox:glibc`, which is pulled by shortname in https://github.com/containers/podman/blob/3a62b4d6b70aaf68a443359319eb3fce7b101b87/test/e2e/pull_test.go#L43-L44 . As best I can tell, that particular image is not cached in `mirror.gcr.io`, and I can't figure out any way to force-cache it (do we have access to that mirror? If we don't, then IMO this exercise in using it is simply postponing a pain point because anything not under our control cannot be relied upon). 

I've tried many variations of `[[registry.mirror]]` trying to point to `quay.io` but the problem is that something somewhere adds a `/library` to `docker.io`, such that pulling `busybox:glibc` becomes `docker.io/library/busybox:glibc`, which if I redirect to `quay.io/libpod` becomes `quay.io/libpod/library/busybox:glibc` which fails with:
```
Mirrors also failed: [quay.io/libpod/library/registry:2.6: Requesting bear token: invalid status code from registry 400 (Bad Request)]):
```
(presumably because `quay.io` doesn't support multi-level hierarchies). 

All for today. I'll see if the morning brings fresh insights.

vrothberg:
> I've tried many variations of `[[registry.mirror]]` trying to point to `quay.io` but the problem is that something somewhere adds a `/library` to `docker.io`, such that pulling `busybox:glibc` becomes `docker.io/library/busybox:glibc`, which if I redirect to `quay.io/libpod` becomes `quay.io/libpod/library/busybox:glibc` which fails with:

```TOML
[[registry]]
prefix="docker.io"
location="quay.io/libpod"

[[registry]]
prefix="docker.io/library"
location="quay.io/libpod"
```

@edsantiago, the upper snippet should handle the described scenario. The longest matching prefix will be selected and hence `docker.io/library/busybox:glibc` would be rewritten to `quay.io/libpod/busybox:glibc`.

cevich:
> just running podman --log-level=debug should be cheap enough

@mtrmac we have a TON of tests which both display and consume stderr/stdout from podman.  Anything that significantly impacts output stands a good chance of needlessly breaking lots of tests, or at best blowing up the size of our test logs (which already require secondary processing for humans to look at).

Re: Add support in c/image and an option in podman.  We could potentially face (or will face) the same `docker.io` problem with buildah testing.  Instead of an env. var., what about simply always including the mirror information in the output?  That might break a handful of tests here and there, but having this information always available is useful for humans in some contexts (i.e. not just automated testing).

cevich:
> @edsantiago, the upper snippet should handle the described scenario. 

@vrothberg want me to set that here (since this PR touches that file already), then I think all we need is to just manually move some images over to quay, no?

edsantiago:
I'm on it, see #8165

cevich:
@edsantiago OIC whatcha doin', okay I won't touch this PR's commits then.  In case you didn't know: As long as the sha's (from rebasing) match up in your PR to this one, merging your's (when it's done) will automatically merge this one.

rhatdan:
Since https://github.com/containers/podman/pull/8165 merged, I am closing this PR.

cevich:
Thanks Dan.  Curious...I thought it would have automatically closed this one since the commit SHAs were the same. No matter.

Issue #7950, opened 2020-10-07
REST API Method Not Allowed with Podman 2.0.5 and Grype


fatherlinux:
/kind bug

**Description**

Using Grype (https://toolbox.anchore.io/) with CentOS Stream 8 and Podman 2.0.5, the Docker Compatible REST endpoint returns a Method not Found error.

**Steps to reproduce the issue:**

1.grype registry.access.redhat.com/ubi8/ubi

Or

1. grype centos

**Describe the results you received:**

From terminal: 
```
[root@centos-stream ~]# grype centos
 ✔ Vulnerability DB     [no update available]
 ✔ Pulled image         
 ⠋ Loading image        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ [requesting image from docker]
[0018] ERROR failed to catalog: could not fetch image 'centos': unable to save image tar: Error response from daemon: Method Not Allowed
```

From debug log:

```
INFO[0071] Failed Request: (405:Method Not Allowed) for GET:'/v1.40/images/get?names=centos'
```

**Describe the results you expected:**

Results similar to video here: https://toolbox.anchore.io/

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.5
API Version:  1
Go Version:   go1.14.7
Built:        Thu Sep 24 13:58:13 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.20-2.module_el8.3.0+475+c50ce30b.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.20, commit: 1019ecdeda3936be22162bb1cca308192145de53'
  cpus: 1
  distribution:
    distribution: '"centos"'
    version: "8"
  eventLogger: file
  hostname: centos-stream.dc2.crunchtools.com
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.18.0-236.el8.x86_64
  linkmode: dynamic
  memFree: 347070464
  memTotal: 1905127424
  ociRuntime:
    name: runc
    package: runc-1.0.0-68.rc92.module_el8.3.0+475+c50ce30b.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.2-dev'
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
[...]

PR #7966, opened 2020-10-08
add compatibility endpoint for exporting multiple images

baude:
with the recent inclusion of dealing with multiple images in a tar archive, we can now add a compatibility endpoint that was missing images/get?names=one,two.

Fixes: #7950

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -55,6 +55,7 @@ func ExportImage(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to create tempfile"))
 		return
 	}
+	defer os.Remove(tmpfile.Name())
 	if err := tmpfile.Close(); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to close tempfile"))
 		return
@@ -69,7 +70,6 @@ func ExportImage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	defer rdr.Close()
-	defer os.Remove(tmpfile.Name())
 	utils.WriteResponse(w, http.StatusOK, rdr)
 }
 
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
Code LGTM

It would be nice to get a test for that in the apiv2 tests.

pkg/api/handlers/compat/images.go
```diff
@@ -398,3 +398,45 @@ func LoadImages(w http.ResponseWriter, r *http.Request) {
...
+		utils.InternalServerError(w, err)
+		return
+	}
+	rdr, err := os.Open(tmpfile.Name())
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to read the exported tarfile"))
+		return
+	}
+	defer rdr.Close()
+	defer os.Remove(tmpfile.Name())
```
jwhonce, line 440: Should not this be at line#430 otherwise you may leave the file on errors.

baude:
yeah, we need them wired up, they are missing rn

pkg/api/handlers/compat/images.go
```diff
@@ -398,3 +398,45 @@ func LoadImages(w http.ResponseWriter, r *http.Request) {
...
+	tmpfile, err := ioutil.TempFile("", "api.tar")
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to create tempfile"))
+		return
+	}
+	if err := tmpfile.Close(); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to close tempfile"))
+		return
+	}
+	if err := runtime.ImageRuntime().SaveImages(r.Context(), images, "docker-archive", tmpfile.Name(), false); err != nil {
```
jwhonce, line 430: I do not see where in SaveImages() the `repositories` file is created.

baude: i personally see that as out of scope.  i would like to treat it as a bug when "we" hit it. cool with that?

baude:
test now added

PR #7967, opened 2020-10-08
Cirrus: Restore APIv2 Testing

### LINKED ISSUE #7954: CI: looks like we lost test-apiv2
As best I can tell, it looks like we lost `test/apiv2` in the big Cirrus switchover. It needs to be added back.

cevich:
Fixes #7954

This was mistakenly left out of 2c9084e2245

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -398,6 +398,7 @@ docker-py_test_task:
     main_script: *main
     always: *artifacts
 
+
 # Does exactly what it says, execute the podman unit-tests on all primary
 # platforms and release versions.
 unit_test_task:
@@ -414,6 +415,30 @@ unit_test_task:
     setup_script: *setup
     main_script: *main
     always: *artifacts
+
+
+apiv2_test_task:
+    name: "APIv2 test on $DISTRO_NV"
+    alias: apiv2_test
+    depends_on:
+        - validate
+    gce_instance: *standardvm
+    env:
+        <<: *stdenvars
+        TEST_FLAVOR: apiv2
+    clone_script: *noop  # Comes from cache
+    gopath_cache: *ro_gopath_cache
+    setup_script: *setup
+    main_script: *main
+    always: &logs_artifacts
+        <<: *artifacts
+        package_versions_script: '$SCRIPT_BASE/logcollector.sh packages'
+        ginkgo_node_logs_script: '$SCRIPT_BASE/logcollector.sh ginkgo'
+        df_script: '$SCRIPT_BASE/logcollector.sh df'
+        audit_log_script: '$SCRIPT_BASE/logcollector.sh audit'
+        journal_script: '$SCRIPT_BASE/logcollector.sh journal'
+        podman_system_info_script: '$SCRIPT_BASE/logcollector.sh podman'
+
 
 # Execute the podman integration tests on all primary platforms and release
 # versions, as root, without involving the podman-remote client.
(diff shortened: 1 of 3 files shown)


rhatdan:
@baude PTAL

rhatdan:
How come this 
APIv2 test on fedora-32 

Does not have a green check mark?

cevich:
> Does not have a green check mark?

Oh thanks for reminding me.  I marked it 'ignore failure', since some of the test items are not working and I need to open an issue on it...

cevich:
@baude PTAL and would you mind taking/assigning #7972?

rhatdan:
So we have a checking and the egg situation on this PR.  If we merge this PR we will need another PR to turn enforcing mode on.
I wish the Skip if failed was a little more visual that the tests actually failed.
If these tests succeed do they get a check mark?  IE Does the grey square box indicated a failed, ignored check?

cevich:
> So we have a checking and the egg situation on this PR.

Aye, hence the issue :disappointed: 

> If these tests succeed do they get a check mark?

Hmmm, I would only be guessing.  I believe the effect of that option is for Cirrus to NOT report any value back, so I think it remains a gray-square no matter what.  

@rhatdan one idea could be...just leave this PR sit here until the tests are fixed, and/or rebasing it into the fixing PR.

cevich:
Note: I added a separate commit which makes the test ignore failures.  So once they're green, that commit can just be reverted.

rhatdan:
@baude @mheon Most of the breakage here is in the Network Stack

edsantiago:
@cevich try rebasing on #7994

cevich:
Sure, happy to...

edsantiago:
My bad; I tested only rootless on my laptop, not root. #8000 should fix.

edsantiago:
@cevich third time's the charm? Please rebase at your convenience, let's see if #8000 can get tests passing.

cevich:
rebased.

cevich:
@edsantiago in case you didn't realize: It works the other way too, you can temporarily rebase the "fix" pr ontop of this one.  You can make the temporary test even faster by marking irreverent tasks with `skip: $CI == $CI` in `.cirrus.yml`

edsantiago:
I think you accidentally rebased on an old master, can you try current master?

cevich:
> I think you accidentally rebased on an old master, can you try current master?

git says `everything up to date`...maybe my clone or config is messed up...

cevich:
...oof, git config is messed up.

edsantiago:
Failure in `int remote ubuntu-20 root host` was a CI timeout; restarted

cevich:
Ugg, `Failed in 01:30:50`...these remote tasks are running awful close to the line.  I know some folks have a mind to make these tests run quicker.  If we keep seeing this, it's probably worth bumping up the timeout by another 10-20 minutes :disappointed:

edsantiago:
AFAIK nobody has even the slightest intention of making anything run quicker. If anything, it's the opposite. So, yeah, bumping the timeout might be necessary.

cevich:
> AFAIK nobody has even the slightest intention of making anything run quicker.

@baude spoke last week or the one before about speeding them up.  IIRC was something about moving calls from `PodmanNoCache()` -> `Podman()`, eh?  I tried to look if there's a card for this...but jira is not behaving ATM :disappointed:

cevich:
@mheon @baude confirmed, I don't see any jira card for this activity.  Mind creating one, so the wish is retained in writing? :smile:

PR #7968, opened 2020-10-08
Print the correct underlying cause for OCI errors

xordspar0:
Previously, the order of OCI error messages was reversed, so that the
type of error was listed as the cause. For example:

    Error: writing file `cpu.cfs_quota_us`: Invalid argument: OCI runtime error

This error message makes it seem like "OCI runtime error" is the
argument that was invalid. In fact, "OCI runtime error" is the error and
"writing file ..." is the cause. With this change, the above message
reads:

    Error: writing file `cpu.cfs_quota_us`: Invalid argument

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/podman/v2/cmd/podman/registry"
 	"github.com/containers/podman/v2/cmd/podman/validate"
+	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/parallel"
 	"github.com/containers/podman/v2/pkg/rootless"
@@ -84,7 +85,7 @@ func init() {
 
 func Execute() {
 	if err := rootCmd.ExecuteContext(registry.GetContextWithOptions()); err != nil {
-		fmt.Fprintln(os.Stderr, "Error:", err.Error())
+		fmt.Fprintln(os.Stderr, formatError(err))
 	} else if registry.GetExitCode() == registry.ExecErrorCodeGeneric {
 		// The exitCode modified from registry.ExecErrorCodeGeneric,
 		// indicates an application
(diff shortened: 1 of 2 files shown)


cmd/podman/root.go
```diff
@@ -84,7 +85,15 @@ func init() {
...
 	if err := rootCmd.ExecuteContext(registry.GetContextWithOptions()); err != nil {
-		fmt.Fprintln(os.Stderr, "Error:", err.Error())
+		var message string
+		if err.Cause() == define.OCIRuntimeError {
+			// OCIRuntimeErrors include the reason for the failure in the
+			// second to last message in the error chain.
+			message = strings.TrimSuffix(err.Error(), ": " + define.OCIRuntimeError.Error())
+		} else {
+			message = err.Error()
+		}
```
xordspar0, line 95: An even simpler version of this would be:

```suggestion
		message := strings.TrimSuffix(err.Error(), ": " + define.OCIRuntimeError.Error())
```

Since `strings.TrimSuffix` returns the string unchanged if the suffix is not found, it's not necessary to guard it with an if statement. Still, I thought that maybe there we'll want to add more transformations in the future. Thoughts? <abbr title="you're not gonna need it">YAGNI</abbr>?

xordspar0: Actually, I think I will do one more transformation right now.

xordspar0:
@mheon @vrothberg PTAL

cmd/podman/root.go
```diff
@@ -84,7 +85,15 @@ func init() {
 
 func Execute() {
 	if err := rootCmd.ExecuteContext(registry.GetContextWithOptions()); err != nil {
-		fmt.Fprintln(os.Stderr, "Error:", err.Error())
+		var message string
+		if errors.Cause(err) == define.ErrOCIRuntime {
+			// OCIRuntimeErrors include the reason for the failure in the
+			// second to last message in the error chain.
+			message = strings.TrimSuffix(err.Error(), ": "+define.ErrOCIRuntime.Error())
```
mheon, line 92: Hm. Can we preface this with `define.ErrOCIRuntime.Error()` - so it's "Error: OCI Runtime error: writing file `cpu.cfs_quota_us`: Invalid argument"? Seems a little bit clearer

xordspar0: Done

cmd/podman/root.go
xordspar0: Now I am doing more transformation, so the if guard is needed.

TomSweeneyRedHat:
@xordspar0  Any chance you could whip up a test to make sure the error message is formatted correctly going forward?

rhatdan:
LGTM

vrothberg (commented):
LGTM

vrothberg:
> @xordspar0 Any chance you could whip up a test to make sure the error message is formatted correctly going forward?

A test would be good. I guess a unit tests would be sufficient.

xordspar0:
Sure. For a unit test I'll need to break the logic out to a separate function that can take arguments. Maybe that's better anyway.

xordspar0:
The tests are all passing. This should be good to merge. There's something wrong with Cirrus that makes it keep losing its connection with the agent nodes, but other than that everything is green.

xordspar0:
:tada: Thanks for the thoughtful reviews and suggestions, everyone!

PR #7970, opened 2020-10-08
Store cgroup manager on a per-container basis

### LINKED ISSUE #7830: Containers created with cgroup_manager=cgroupfs fail obscurely with new default settings

/kind bug

**Description**

At some point in the past, apparently libpod wrote a libpod.conf that included `cgroup_manager=cgroupfs` (small chance that I added it manually as a workaround for some issue in the past.)

When I upgraded to podman-2.1.1, I got the new warning:

```
 Ignoring libpod.conf EventsLogger setting %q. Use containers.conf if you want to change this setting and remove libpod.conf files.
```

I inspected `libpod.conf`, determined that there was nothing in there that looked like local configuration, and deleted it. At this point, my old containers no longer started, with the error message:

 ```
$ podman start fedora-toolbox-33
Error: unable to start container "72f11d1d13db67827c20861e1c3017a9e4fce4a75ffdb0c067dad6b34bdc1f19": sd-bus call: Invalid argument: OCI runtime error
```

After an hour or so of investigation, noticing that new containers did start, and diffing debug output between new and old containers, I determined that it was possible to start my old containers with a manual --cgroup-manager=cgroupfs.

it was pointed out to me that there's a note in the podman(1) man page: Note: Setting this flag can cause certain commands to break when called on containers previously created by the other CGroup manager type." - but even if I had found this note, it would not necessarily been clear to me that "certain commands" includes starting the container - the wording sounds like it's referring to edge cases.

```
$ podman --cgroup-manager=cgroupfs run --name=test1 fedora:32 echo hi
hi
$ podman start test1
Error: unable to start container "ac41f25244ff62da8d3702d389ca11e41fc74fc464d7cfa31914580817225978": sd-bus call: Invalid argument: OCI runtime error
```

One of the following:

 1. The container runs properly with the systemd cgroup manager - from a user perspective, it's hard to see why this setting can't be used with a previously created container. 
 2. libpod detects that the container was created with the cgroupfs cgroup manager and automatically uses that when starting it
 3. libpod gives a clear error message saying what is going wrong, and how to fix it.

From a toolbox perspective, 1 or 2 would be better, since the user doesn't invoke podman directly, and just expects their previously created toolbox to keep working. Having toolbox detect the situation and add `--cgroup-manager` seems less than ideal.

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.1
Built:        Sun Sep 27 09:43:05 2020
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-0.3.dev.git5a6b2ac.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21-dev, commit: 5c03a23398b94cf869f071128631ef7fd9153b3b'
  cpus: 8
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.11-300.fc33.x86_64
  linkmode: dynamic
  memFree: 22865244160
  memTotal: 33439457280
  ociRuntime:
    name: crun
    package: crun-0.15-3.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.15
      spec: 1.0.0
  os: linux
  remoteSocket:
  rootless: true
  slirp4netns:
    package: slirp4netns-1.1.4-4.dev.giteecccdb.fc33.x86_64
    version: |-
      slirp4netns version 1.1.4+dev
      commit: eecccdb96f587b11d7764556ffacfeaffe4b6e11
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 3
      libseccomp: 2.5.0
  swapFree: 21082660864
  swapTotal: 21082660864
  uptime: 1h 42m 20.93s (Approximately 0.04 days)
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - docker.io
store:
  configFile: /var/home/otaylor/.config/containers/storage.conf
  containerStore:
    number: 20
    paused: 0
    running: 1
    stopped: 19
  graphDriverName: overlay
  graphOptions:
    overlay.mount_program:
      Package: fuse-overlayfs-1.1.2-1.fc33.x86_64
      Version: |-
        fusermount3 version: 3.9.3
        fuse-overlayfs: version 1.1.0
        FUSE library version 3.9.3
  graphRoot: /var/home/otaylor/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 216
  volumePath: /var/home/otaylor/.local/share/containers/storage/volumes
version:
  APIVersion: 2.0.0
  Built: 1601214185
  BuiltTime: Sun Sep 27 09:43:05 2020
  GitCommit: ""
  GoVersion: go1.15.1
  OsArch: linux/amd64
  Version: 2.1.1
```

```
$ rpm -q podman
warning: Found bdb Packages database while attempting sqlite backend: using bdb backend.
podman-2.1.1-2.fc33.x86_64
```

Yes

mheon:
When we create a container, we assign a cgroup parent based on the current cgroup manager in use. This parent is only usable with the cgroup manager the container is created with, so if the default cgroup manager is later changed or overridden, the container will not be able to start.

To solve this, store the cgroup manager that created the container in container configuration, so we can guarantee a container with a systemd cgroup parent will always be started with systemd cgroups.

Unfortunately, this is very difficult to test in CI, due to the fact that we hard-code cgroup manager on all invocations of Podman in CI.

Fixes #7830

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -888,9 +888,22 @@ func (c *Container) NamespacePath(linuxNS LinuxNS) (string, error) { //nolint:in
 	return fmt.Sprintf("/proc/%d/ns/%s", c.state.PID, linuxNS.String()), nil
 }
 
+// CgroupManager returns the cgroup manager used by the given container.
+func (c *Container) CgroupManager() string {
+	cgroupManager := c.config.CgroupManager
+	if cgroupManager == "" {
+		cgroupManager = c.runtime.config.Engine.CgroupManager
+	}
+	return cgroupManager
+}
+
 // CGroupPath returns a cgroups "path" for a given container.
 func (c *Container) CGroupPath() (string, error) {
+	cgroupManager := c.CgroupManager()
+
 	switch {
+	case c.config.NoCgroups || c.config.CgroupsMode == "disabled":
+		return "", errors.Wrapf(define.ErrNoCgroups, "this container is not creating cgroups")
 	case c.config.CgroupsMode == cgroupSplit:
 		if c.config.CgroupParent != "" {
 			return "", errors.Errorf("cannot specify cgroup-parent with cgroup-mode %q", cgroupSplit)
@@ -906,9 +919,9 @@ func (c *Container) CGroupPath() (string, error) {
 			return "", errors.Errorf("invalid cgroup for conmon %q", cg)
 		}
 		return strings.TrimSuffix(cg, "/supervisor") + "/container", nil
-	case c.runtime.config.Engine.CgroupManager == config.CgroupfsCgroupsManager:
+	case cgroupManager == config.CgroupfsCgroupsManager:
 		return filepath.Join(c.config.CgroupParent, fmt.Sprintf("libpod-%s", c.ID())), nil
-	case c.runtime.config.Engine.CgroupManager == config.SystemdCgroupsManager:
+	case cgroupManager == config.SystemdCgroupsManager:
 		if rootless.IsRootless() {
 			uid := rootless.GetRootlessUID()
 			parts := strings.SplitN(c.config.CgroupParent, "/", 2)
@@ -922,7 +935,7 @@ func (c *Container) CGroupPath() (string, error) {
 		}
 		return filepath.Join(c.config.CgroupParent, createUnitName("libpod", c.ID())), nil
 	default:
-		return "", errors.Wrapf(define.ErrInvalidArg, "unsupported CGroup manager %s in use", c.runtime.config.Engine.CgroupManager)
+		return "", errors.Wrapf(define.ErrInvalidArg, "unsupported CGroup manager %s in use", cgroupManager)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #7971, opened 2020-10-08
BlobInfoCacheDir is set incorrectly when copying images

rhatdan:
It is not set based on the root image directory, and always
points at the defaults.  This change will get it to follow
filepath.Join(ir.store.GraphRoot(), "cache") set from libpod.


rhatdan:

TomSweeneyRedHat:
LGTM

mheon:
LGTM

PR #7975, opened 2020-10-09
Restore --format table...

jwhonce:
Following commands:
* systemd generate
* networks inspect
* pod stats
  * Fixed test where format was quoted and then quoted again
  * Fixed bug where output never printed '--' on missed reads
* pod ps


rhatdan:
LGTM

PR #7983, opened 2020-10-09
Include CNI networks in inspect output when not running

mheon:
We were only including the CNI Network fields in the output of `podman inspect` when the container was not running. It's simple enough to fix (populate with empty structs, since we can't fill anything without a CNI response to get IP address assigned, etc).

This is necessary for Docker compatibility.


mheon:
@baude PTAL

jwhonce:
LGTM once tests are happy

TomSweeneyRedHat:
changes LGTM, but tests look to be having some real issues.

mheon:
Restarting tests, looks like it was mostly flakes in uploading artifacts?

PR #7984, opened 2020-10-09
pkg/spec: fix a confusing error message

nalind:
When we try, but fail, to load the default seccomp profile, say that, instead of suggesting that we tried to load a profile with no name.


mheon:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Looks like two of the tests were automatically canceled here?  We'd weird CI problems on Bulidah today, I'm blaming @nalind for both!  ;^)

mheon:
/hold

Restarted the cancelled test

PR #7987, opened 2020-10-09
Restore --format table support

jwhonce:
* system df
* events
  * fix error handling from go routine
* system info
* version
* volume inspect


cmd/podman/system/version.go
```diff
@@ -2,14 +2,15 @@ package system
 
 import (
 	"fmt"
+	"html/template"
```
ashley-cui, line 5: should this be `text/template` ?

jwhonce: yup, auto-import for the fail

cmd/podman/system/df.go
```diff
@@ -143,78 +124,94 @@ func printSummary(reports *entities.SystemDfReport, userFormat string) error {
...
 	volumeRow := "{{.VolumeName}}\t{{.Links}}\t{{.Size}}\n"
-	format = volumeHeaders + "{{range . }}" + volumeRow + "{{end}}"
-	return writeTemplate(w, format, dfVolumes)
+	format = "{{range . }}" + volumeRow + "{{end}}"
+	return writeTemplate(cmd, w, hdrs, format, dfVolumes)
 }
 
-func writeTemplate(w io.Writer, format string, output interface{}) error {
-	tmpl, err := template.New("dfout").Parse(format)
+func writeTemplate(cmd *cobra.Command, w *tabwriter.Writer, hdrs []map[string]string, format string,
```
TomSweeneyRedHat, line 200: Maybe it's my twisted mind, and it's a nit at best, but I think any I/O function should have the I/O object as the first param, not further down the line.  i.e.: 
```suggestion
func writeTemplate(w *tabwriter.Writer, cmd *cobra.Command, hdrs []map[string]string, format string,
```
My $.02, YMMV

cmd/podman/system/version.go
```diff
@@ -41,31 +42,38 @@ func version(cmd *cobra.Command, args []string) error {
...
-		_, err = io.WriteString(os.Stdout, s+"\n")
-		return err
-	case cmd.Flag("format").Changed:
-		out := formats.StdoutTemplate{Output: versions, Template: versionFormat}
-		err := out.Out()
+		fmt.Println(s)
+		return nil
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
```
TomSweeneyRedHat, line 54: Unless they're on top of each other, I really dislike short variable names, especially so one char ones.  Searching for them is just too painful when updating/fixining two years from now.

cmd/podman/system/df.go
jwhonce: I used this order because that is the order from the other cobra functions

ashley-cui:
LGTM besides @TomSweeneyRedHat 's comment

PR #7995, opened 2020-10-12
[CI:DOCS] Remove a note that the HTTP API is not yet stable.

mheon:
It's been out for quite a while now, I think we can safely call it stable.


mheon:
@cevich Did we disable ci:docs?

mheon:
/hold

We have flakes

TomSweeneyRedHat:
LGTM

cevich:
> @cevich Did we disable ci:docs?

In a way yes...it was never re-implemented.  In a pinch, you can use `[ci skip]` for largely the same effect.  I'll work up a PR to add it back in...

cevich:
...https://github.com/containers/podman/pull/7998

PR #7797, opened 2020-09-28
[CI:DOCS] Add section about current differences

rsommer:
There are some differences in behaviour between docker and podman which should be documented to help developers while transitioning.

PR #7998, opened 2020-10-12
[CI:DOCS] Cirrus: Ability to skip most tests for docs updates

cevich:
***Warning***: `skip` has non-obvious side-effects vs `only_if`:
https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution

The skip instruction can give a false sense of security by always
marking tasks as passed in the UI, even if they didn't run. In
contrast, the `only_if` condition will avoid creating the task
all -together; therefore, a problematic task's absense is more likely to
be noticed if it introduced a problem.

***Note:*** I left the OSX build in because it actually does some platform-specific docs-manipulation we want exercised.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -143,6 +143,7 @@ smoke_task:
 build_task:
     alias: 'build'
     name: 'Build for $DISTRO_NV'
+    only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
     gce_instance: &standardvm
         image_project: libpod-218412
         zone: "us-central1-a"
@@ -200,12 +201,14 @@ build_task:
             path: ./*.html
             type: text/html
 
+
 # Confirm the result of building on at least one platform appears sane.
 # This confirms the binaries can be executed, checks --help vs docs, and
 # other essential post-build validation checks.
 validate_task:
     name: "Validate $DISTRO_NV Build"
     alias: validate
+    only_if: *not_docs
     depends_on:
         - ext_svc_check
         - automation
@@ -227,11 +230,13 @@ validate_task:
     main_script: *main
     always: *artifacts
 
+
 # Exercise the "libpod" API with a small set of common
 # operations to ensure they are functional.
 bindings_task:
     name: "Test Bindings"
     alias: bindings
+    only_if: *not_docs
     skip: *branch
     depends_on:
         - build
@@ -268,6 +273,7 @@ swagger_task:
 endpoint_task:
     name: "Test Endpoint"
     alias: endpoint
+    only_if: *not_docs
     depends_on:
         - build
     container: *smallcontainer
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

TomSweeneyRedHat:
Is this taking over for #7797?  Looks like it's working better.

edsantiago:
Looks like we collided. I went with `skip` because it was easy for my map maker to parse and because positive logic (`skip if PR =~ CI:DOCS`) is more readable and maintainable than negative (`only if PR !~ CI:DOCS`).

.cirrus.yml
```diff
@@ -143,6 +143,7 @@ smoke_task:
 build_task:
     alias: 'build'
     name: 'Build for $DISTRO_NV'
+    only_if: $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
```
edsantiago, line 146: Would you consider making this a macro, e.g. `&not_docs`, for readability (and DRY) below?

edsantiago:
LGTM, modulo one request for DRY. I like how this approach doesn't have all the nasty gray bullet clutter that `skip` does. Map attached for those who find it useful for review.
cirrus-map-pr7998

cevich:
> I went with skip because it was easy for my map maker to parse and because positive logic

@edsantiago warning...danger..danger: `skip` has non-obvious side-effects vs `only_if`:
https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution

The `skip` instruction can give a false sense of security by always marking tasks as passed in the UI, even if they didn't run.  Therefore, this includes if the PR actually breaks something in one of those green-but-skipped tasks.

Consider the position of some future CI-failure analyst:  They must go trolling back through multiple CI builds, trying to find the one that broke something.  If the culprit was a `[CI:DOCS]` task, they are much more likely to notice if it did **not** run the broken task, as opposed to seeing **all tasks are all green**.   That's why I deliberately chose to use `only_if` over `skip`, the former doesn't create the task at all, the later does and then lies about it's status.

.cirrus.yml
cevich: good idea

cevich:
Rebased and updated w/ Ed's anchor/alias suggestion.  I also updated the commit and PR message to make clear the reasoning behind `only_if` vs `skip` usage.

cevich:
Clarification: What I mean by false-success status can be seen in this PRs twin, in the WebUI: https://cirrus-ci.com/build/6731773184835584

cevich:
@edsantiago is there some reason why we're holding on this? Did I forget something important again?

edsantiago:
No reason to hold, I just use that convention to allow you to hold-cancel or to solicit further review; i.e. to leave that decision up to you.

cevich:
Oh whew! okay, thanks.

PR #7999, opened 2020-10-12
Add a shutdown signal handler

### LINKED ISSUE #7941: Container storage creation fragile

/kind bug

**Description**

`podman` is fragile, and there are various points where interrupting an operation with ctrl+c will leave podman's state undefined, and cause follow-on breakage.

podman should make every effort to be in a position to roll-forwards or roll-back on invocation from any state-change which was in progress but incomplete from a previous run (taking into account that it may be running in parallel with other instances of itself).

For example, I inadvertently interrupted a container starting, I assume during volume creation.  I now get this:

```
Error: error creating container storage: the container name "buildweb-app-admin.webapp-config-1.55-r1" is already in use by "2c68b0f01168b42da187d9bfca25552913a2a6d8e8f3ccddc9c33a4a798e70e9". You have to remove that container to be able to reuse that name.: that name is already in use

# podman rm -v 2c68b0f01168b42da187d9bfca25552913a2a6d8e8f3ccddc9c33a4a798e70e9
Error: no container with name or ID 2c68b0f01168b42da187d9bfca25552913a2a6d8e8f3ccddc9c33a4a798e70e9 found: no such container

# podman rm -v buildweb-app-admin.webapp-config-1.55-r1
Error: no container with name or ID buildweb-app-admin.webapp-config-1.55-r1 found: no such container
```

... and no amount of system/volume pruning seems to be able to fix this issue :(

I can poke around on the filesystem to try to fix this (... or simply never use that container name ever again!!), but `podman` shouldn't let things get into this state.

Could locking be added to determine whether a partially-completed resource is under the control of a simultaneous still-running task, or is stale and should be removed?

Ideally, podman should be able to be interrupted at _any_ point during execution, and still able to operate when re-run (or, at the very least, correctly clean-up any partial state on a `system prune`).

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14.7
Built:        Sat Oct  3 11:01:42 2020
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 35a2fa83022e56e18af7e6a865ba5d7165fa2a4a'
  cpus: 8
  distribution:
    distribution: gentoo
    version: unknown
  eventLogger: file
  hostname: dellr330
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.8.9-gentoo
  linkmode: dynamic
  memFree: 2285371392
  memTotal: 8063447040
  ociRuntime:
    name: crun
    package: Unknown
    path: /usr/bin/crun
    version: |-
      crun version 0.15-dirty
      spec: 1.0.0
      +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 25193656320
  swapTotal: 25769787392
  uptime: 182h 14m 13.03s (Approximately 7.58 days)
registries:
  search:
  - docker.io
  - docker.pkg.github.com
  - quay.io
store:
  containerStore:
    number: 16
    paused: 0
    running: 16
    stopped: 0
  graphDriverName: overlay
  graphOptions:
    overlay.ignore_chown_errors: "false"
  graphRoot: /space/podman/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 264
  runRoot: /space/podman/run
  volumePath: /space/podman/volumes
version:
  APIVersion: 2.0.0
  Built: 1601719302
  BuiltTime: Sat Oct  3 11:01:42 2020
  GitCommit: 9f6d6ba0b314d86521b66183c9ce48eaa2da1de2
  GoVersion: go1.14.7
  OsArch: linux/amd64
  Version: 2.1.1

```

Yes

mheon:
We need a unified package for handling signals that shut down Libpod and Podman. We need to be able to do different things on receiving such a signal (`system service` wants to shut down the service gracefully, while most other commands just want to exit) and we need to be able to inhibit this shutdown signal while we are waiting for some critical operations (e.g. creating a container) to finish. We also want to turn off this handling when `--sig-proxy` is active.

Fixes #7941

### DIFF
--- a/libpod/shutdown/handler.go
+++ b/libpod/shutdown/handler.go
@@ -1,0 +1,105 @@
+package shutdown
+
+import (
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	stopped         bool
+	sigChan         chan os.Signal
+	cancelChan      chan bool
+	handlers        map[string]func() error
+	shutdownInhibit sync.RWMutex
+)
+
+// Start begins handling SIGTERM and SIGINT and will run the given on-signal
+// handlers when one is called. This can be cancelled by calling Stop().
+func Start() error {
+	if sigChan != nil && !stopped {
+		// Already running, do nothing.
+		return nil
+	}
+
+	sigChan = make(chan os.Signal, 1)
+	cancelChan = make(chan bool, 1)
+	stopped = false
+
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		select {
+		case <-cancelChan:
+			signal.Stop(sigChan)
+			close(sigChan)
+			close(cancelChan)
+			stopped = true
+			return
+		case sig := <-sigChan:
+			logrus.Infof("Received shutdown signal %v, terminating!", sig)
+			shutdownInhibit.Lock()
+			for name, handler := range handlers {
+				logrus.Infof("Invoking shutdown handler %s", name)
+				if err := handler(); err != nil {
+					logrus.Errorf("Error running shutdown handler %s: %v", name, err)
+				}
+			}
+			shutdownInhibit.Unlock()
+			return
+		}
+	}()
+
+	return nil
+}
+
+// Stop the shutdown signal handler.
+func Stop() error {
+	if cancelChan == nil {
+		return errors.New("shutdown signal handler has not yet been started")
+	}
+	if stopped {
+		return nil
+	}
+
+	cancelChan <- true
+
+	return nil
+}
+
+// Temporarily inhibit signals from shutting down Libpod.
+func Inhibit() {
+	shutdownInhibit.RLock()
+}
+
+// Stop inhibiting signals from shutting down Libpod.
+func Uninhibit() {
+	shutdownInhibit.RUnlock()
+}
+
+// Register registers a function that will be executed when Podman is terminated
+// by a signal.
+func Register(name string, handler func() error) error {
+	if _, ok := handlers[name]; ok {
+		return errors.Errorf("handler with name %s already exists", name)
+	}
+
+	handlers[name] = handler
+
+	return nil
+}
+
+// Unregister un-registers a given shutdown handler.
+func Unregister(name string) error {
+	if _, ok := handlers[name]; !ok {
+		return errors.Errorf("no handler with name %s found", name)
+	}
+
+	delete(handlers, name)
+
+	return nil
+}
--- a/libpod/shutdown/handler.go
+++ b/libpod/shutdown/handler.go
@@ -11,17 +11,20 @@ import (
 )
 
 var (
-	stopped         bool
-	sigChan         chan os.Signal
-	cancelChan      chan bool
-	handlers        map[string]func() error
+	stopped    bool
+	sigChan    chan os.Signal
+	cancelChan chan bool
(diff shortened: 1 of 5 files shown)


mheon:
Potential improvements:
- Pass the signal into the shutdown handler so we can switch behaviour based on SIGINT vs SIGTERM
- Order shutdown handlers LIFO so we could use both the server and libpod handlers as the same time, server handler coming first

mheon:
@jwhonce PTAL, I'd like to make sure I got signal handling for the server right.

libpod/shutdown/handler.go
```diff
@@ -0,0 +1,113 @@
...
+	sigChan         chan os.Signal
+	cancelChan      chan bool
+	handlers        map[string]func() error
+	shutdownInhibit sync.RWMutex
+)
+
+// Start begins handling SIGTERM and SIGINT and will run the given on-signal
+// handlers when one is called. This can be cancelled by calling Stop().
+func Start() error {
+	if sigChan != nil && !stopped {
```
jwhonce, line 24: It would appear to be a race on reading/setting `stopped`

libpod/shutdown/handler.go
```diff
@@ -0,0 +1,113 @@
...
+// Register registers a function that will be executed when Podman is terminated
+// by a signal.
+func Register(name string, handler func() error) error {
+	if handlers == nil {
+		handlers = make(map[string]func() error)
+	}
+
+	if _, ok := handlers[name]; ok {
+		return errors.Errorf("handler with name %s already exists", name)
+	}
```
jwhonce, line 93:
```suggestion
```
Why not just reset handler to new func?

pkg/api/server/server.go
```diff
@@ -180,8 +180,20 @@ func setupSystemd() {
...
+		return err
+	}
+	if err := shutdown.Register("server", func() error {
+		return s.Shutdown()
+	}); err != nil {
+		return err
+	}
+	// Unregister the libpod handler, which just calls exit(1).
+	// Ignore errors if it doesn't exist.
+	_ = shutdown.Unregister("libpod")
```
jwhonce, line 195: Seems smelly to remove someone elses handler by name...

libpod/shutdown/handler.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Unregister un-registers a given shutdown handler.
+func Unregister(name string) error {
+	if handlers == nil {
+		handlers = make(map[string]func() error)
+	}
+
+	if _, ok := handlers[name]; !ok {
+		return errors.Errorf("no handler with name %s found", name)
+	}
```
jwhonce, line 108:
```suggestion
```
Should removing a non-existing handler really be an error? Especially if non-owner code is going to be removing handlers.

libpod/shutdown/handler.go
mheon: I'd like to allow multiple handlers at the same time - theoretically the Libpod handler could run after the server handler and do any extra shutdown needed there.

mheon:
@jwhonce PTAL again, comments should be fixed

jwhonce:
Restarted test flakes

/lgtm

PR #8000, opened 2020-10-12
APIv2 tests: try again to fix them

edsantiago:
CI discovered that a lot of networking tests are failing; my
fault, for not having run my tests as root on my laptop.
Disable those.

Also: bump up the ten-request time limit, from 5 to 7 seconds.
Looks like something keeps getting slower and slower, but I
guess there's not much we can do about it.

Also: when we get a mismatch response code (e.g. 500 when we
expect 200), dump the response body and skip any subsequent
response checks.

### DIFF
--- a/test/apiv2/01-basic.at
+++ b/test/apiv2/01-basic.at
@@ -68,10 +68,13 @@ for i in $(seq 1 10); do
 done
 t1=$SECONDS
 delta_t=$((t1 - t2))
-if [ $delta_t -le 5 ]; then
-    _show_ok 1 "Time for ten /info requests ($delta_t seconds) <= 5s"
+
+# Desired number of seconds in which we expect to run.
+want=7
+if [ $delta_t -le $want ]; then
+    _show_ok 1 "Time for ten /info requests ($delta_t seconds) <= ${want}s"
 else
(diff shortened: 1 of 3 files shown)


test/apiv2/01-basic.at
```diff
@@ -68,7 +68,7 @@ for i in $(seq 1 10); do
 done
 t1=$SECONDS
 delta_t=$((t1 - t2))
-if [ $delta_t -le 5 ]; then
+if [ $delta_t -le 7 ]; then
     _show_ok 1 "Time for ten /info requests ($delta_t seconds) <= 5s"
```
mheon, line 72: Should this be 7 seconds?

edsantiago: Argh. That's exactly why I hate duplication. Fixed, and refactored so it shouldn't happen again the next time we have to bump up the time limit. Thank you for catching that.

edsantiago:
Cirrus flake again, `Failed to stop: readHandshakeRecord` (not one I had seen before). Restarted.

edsantiago:
Yet another `CI agent stopped responding!`. Restarted, again.

rhatdan:
LGTM

Issue #8004, opened 2020-10-13
podman run --privileged --userns=keep-id does not properly set HOME env


seanmcneil3:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When running a container in privileged mode with userns=keep-id, the output of the command "id" is correct, but behavior is slightly off with podman version 2.1.1. In previous version, the environment variable HOME would be properly set to the namespace. It is now set as HOME=/root.

**Steps to reproduce the issue:**

1. Create a container with a minimum of a bash shell environment. I've chosen from centos:7

2.Start a shell with the container, such as with:

podman run --rm -it --privileged --userns=keep-id --volume="/home:/home" --workdir="$PWD" centos:7 bash

3. In the container, execute "echo $HOME"

**Describe the results you received:**

I see "/root".

**Describe the results you expected:**

I expect to see the path for the user I run the podman container from. This is a deviation from previous releases.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

podman version 2.1.1

**Output of `podman info --debug`:**

host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 0f53fb68333bdead5fe4dc5175703e22cf9882ab'
  cpus: 16
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: desk.mcneil.com
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
  kernel: 5.8.13-300.fc33.x86_64
  linkmode: dynamic
  memFree: 13702463488
  memTotal: 33680375808
  ociRuntime:
    name: crun
    package: crun-0.15-5.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.15
      commit: 56ca95e61639510c7dbd39ff512f80f626404969
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
[...]

Issue #7139, opened 2020-07-29
rootless: pods: unlinkat, EBUSY


edsantiago:
Are pods supposed to work with podman-remote? There's a race condition somewhere in rootless pods. (There's one in rootful, too, but I'm having a really hard time getting a reproducer).

In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ cat >foo.sh <<EOF
#!/bin/bash

./bin/podman-remote pod create --infra=true --name=foo

cid=$(./bin/podman-remote run -d   --pod foo alpine sleep 1)
./bin/podman-remote run --rm --pod foo alpine true

sleep 1
/bin/time ./bin/podman-remote rm $cid

/bin/time ./bin/podman-remote pod rm foo
EOF

$ bash -x foo.sh
+ ./bin/podman-remote pod create --infra=true --name=foo
4e182754caf7e065917a258a9a5ebc4c29498df1f2381b165c10616c458ebedf
++ ./bin/podman-remote run -d --pod foo alpine sleep 1
+ cid=82822d1b6544692162118f553b857a10490a6a18ebd2ae6ecb1c088eec661c37
+ ./bin/podman-remote run --rm --pod foo alpine true
+ sleep 1
+ /bin/time ./bin/podman-remote rm 82822d1b6544692162118f553b857a10490a6a18ebd2ae6ecb1c088eec661c37
Error: error removing container 82822d1b6544692162118f553b857a10490a6a18ebd2ae6ecb1c088eec661c37 root filesystem: 1 error occurred:
        * unlinkat /home/esm/.local/share/containers/storage/overlay/3ba0c1f06abbda6296b7fd183b776c212bb5cf352100ea1ff6e93bd543204037/merged: device or resource busy


Command exited with non-zero status 125
0.03user 0.03system 0:10.12elapsed 0%CPU (0avgtext+0avgdata 29276maxresident)k   <--- note: 10s
0inputs+0outputs (0major+2127minor)pagefaults 0swaps
+ /bin/time ./bin/podman-remote pod rm foo
Error: error removing container 59f9f424ad65a88f4b55af3873a00259f74eeb8e4ab56e6bf1e01f041e822d84 root filesystem: 1 error occurred:
        * unlinkat /home/esm/.local/share/containers/storage/overlay/02b1322ceed80312346340312c1068d832727ac30a3683c6c6b2cb22073c40fa/merged: device or resource busy


Command exited with non-zero status 125
[...]

PR #8013, opened 2020-10-13
Setup HOME environment when using --userns=keep-id

rhatdan:
Currently the HOME environment is set to /root if
the user does not override it.

Also walk the parent directories of users homedir
to see if it is volume mounted into the container,
if yes, then set it correctly.

Fixes: https://github.com/containers/podman/issues/8004

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1717,11 +1717,35 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, int, int, error) {
 	// If the user's actual home directory exists, or was mounted in - use
 	// that.
 	homeDir := c.WorkingDir()
-	if MountExists(c.config.Spec.Mounts, u.HomeDir) {
-		homeDir = u.HomeDir
+	hDir := u.HomeDir
+	for hDir != "/" {
+		if MountExists(c.config.Spec.Mounts, hDir) {
+			homeDir = u.HomeDir
+			break
+		}
+		hDir = filepath.Dir(hDir)
+	}
+	if homeDir != u.HomeDir {
+		for _, hDir := range c.UserVolumes() {
+			if hDir == u.HomeDir {
+				homeDir = u.HomeDir
+				break
+			}
+		}
+	}
+	// Set HOME environment if not already set
+	hasHomeSet := false
+	for _, s := range c.config.Spec.Process.Env {
+		if strings.HasPrefix(s, "HOME=") {
+			hasHomeSet = true
+			break
+		}
+	}
+	if !hasHomeSet {
+		c.config.Spec.Process.Env = append(c.config.Spec.Process.Env, fmt.Sprintf("HOME=%s", homeDir))
 	}
 
-	return fmt.Sprintf("%s:*:%s:%s:%s:%s:/bin/sh\n", u.Username, u.Uid, u.Gid, u.Username, homeDir), uid, rootless.GetRootlessGID(), nil
+	return fmt.Sprintf("%s:*:%s:%s:%s:%s:/bin/sh\n", u.Username, u.Uid, u.Gid, u.Name, homeDir), uid, rootless.GetRootlessGID(), nil
 }
 
 // generateUserPasswdEntry generates an /etc/passwd entry for the container user
--- a/test/e2e/toolbox_test.go
+++ b/test/e2e/toolbox_test.go
@@ -365,4 +365,16 @@ var _ = Describe("Toolbox-specific testing", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(ContainSubstring("READY"))
 	})
+
+	It("podman run --userns=keep-id check $HOME", func() {
+		var session *PodmanSessionIntegration
+
+		currentUser, err := user.Current()
+		Expect(err).To(BeNil())
+		session = podmanTest.Podman([]string{"run", "-v", fmt.Sprintf("%s:%s", currentUser.HomeDir, currentUser.HomeDir), "--userns=keep-id", fedoraToolbox, "sh", "-c", "echo $HOME"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring(currentUser.HomeDir))
+	})
+
 })


libpod/container_internal_linux.go
```diff
@@ -1717,11 +1717,27 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, int, int, error) {
 	// If the user's actual home directory exists, or was mounted in - use
 	// that.
 	homeDir := c.WorkingDir()
-	if MountExists(c.config.Spec.Mounts, u.HomeDir) {
-		homeDir = u.HomeDir
+	hDir := u.HomeDir
+	for hDir != "/" {
```
TomSweeneyRedHat, line 1721: u.HomeDir is a string no?
```suggestion
	if hDir != "/" {
```

libpod/container_internal_linux.go
```diff
@@ -1717,11 +1717,27 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, int, int, error) {
...
+		if MountExists(c.config.Spec.Mounts, hDir) {
+			homeDir = u.HomeDir
+			break
+		}
+		hDir = filepath.Dir(hDir)
```
TomSweeneyRedHat, line 1726: me thinks
```suggestion
		homeDir = filepath.Dir(hDir)
```
hDir doesn't seem to be referred to after this line otherwise.

libpod/container_internal_linux.go
rhatdan: This code is attempting to walk up the directories to check if their is a mountpoint from the host that covers this directory.

For example if my homedir was /home/engineering/dwalsh.
I want to check if 
 /home/engineering/dwalsh
 /home/engineering
 /home
Is mounted into the container, if yes then I can use /home/engineering/dwalsh as the homedir, rather then CWD.

libpod/container_internal_linux.go
rhatdan: If hDir is volume mounted into the container then
```
+                       homeDir = u.HomeDir
+                       break
```
Is executed which sets the homeDir to the directory from the host as opposed to the CWD.

libpod/container_internal_linux.go
rhatdan: No, we want to look through the directories.

mheon: I don't know if this is really wanted - when I originally wrote this, I wanted it to be specific to ensure that we didn't leak information about the user on the host into the container unless they were explicitly mounting their home

test/e2e/toolbox_test.go
```diff
@@ -365,4 +365,16 @@ var _ = Describe("Toolbox-specific testing", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(ContainSubstring("READY"))
 	})
+
+	It("podman create --userns=keep-id - entrypoint - adding user with useradd and then removing their password", func() {
+		var session *PodmanSessionIntegration
+
+		currentUser, err := user.Current()
+		Expect(err).To(BeNil())
+		session = podmanTest.Podman([]string{"run", "--name", "-v", fmt.Sprintf("%s:%s:Z", currentUser.HomeDir, currentUser.HomeDir), "test", "--userns=keep-id", fedoraToolbox, "sh", "-c", "echo $HOME"})
```
edsantiago, line 374: I haven't looked at the CI failures, but I'm guessing that `--name -v` is not going to work well

edsantiago:
There's a bug somewhere, I don't have time right nowww to delve into it:
```console
$ ./bin/podman run --userns=keep-id quay.io/libpod/testimage:20200929 printenv HOME
/home/podman
```
(my username on this system is 'esm', not 'podman'. There is a valid `/home/podman` directory on this image, but no `podman` account).

rhatdan:
@edsantiago 
try
./bin/podman run -it -v /home/esm:/home/esm --userns=keep-id alpine sh

Sadly I don't think I can handle just creating the internal /home/dir.

edsantiago:
I was just trying that, and it works. I was curious why plain `-v /home/esm` (named volume, not a bind mount) didn't work though. But it's not that big a deal. I'm also curious about the `testimage` case, this is where the image specifies a WORKDIR. I can see not wanting to override that for cwd, but I do have a slight inclination toward setting `$HOME` to user's homedir even if there is a workdir.

rhatdan:
I am not sure of what the correct thing to do is.  It would simplify the code quite a bit.

rhatdan:
Setting $HOME to point at an non existing homedir, might be problematic.

edsantiago:
>Setting $HOME to point at an non existing homedir, might be problematic.

Agreed. My concern is: image has workdir. User runs container with `-v homedir`. Initial cwd is workdir (not controversial), but in this implementation, $HOME also points to workdir, even though `/home/user` exists. I am inclined to think that `$HOME` should be `/home/user` no matter whether a workdir is defined or not.

edsantiago:
Yuk - failure looks horrible, but I can't see how it could be related to your PR. Am restarting.
```
         time="2020-10-14T14:56:24-05:00" level=error msg="error joining network namespace for container bfa5af4e01cc6886e8de0e7218e6c68acbf8d3569e7f1bb8e34d960079723913: error retrieving network namespace at /proc/83941/ns/net: failed to Statfs \"/proc/83941/ns/net\": permission denied"
```
Will assume that my flake-monitor will gather it up, add it to the list, and I'll eventually notice if this one is a common one.

rhatdan:
@edsantiago I now cover your case as well.
```
./bin/podman run -v /home/dwalsh:/home/dwalsh -it --rm --userns=keep-id alpine printenv HOME
/home/dwalsh
$ ./bin/podman run -v /home/dwalsh -it --rm --userns=keep-id alpine printenv HOME
/home/dwalsh
$ ./bin/podman run -it --rm --userns=keep-id alpine printenv HOME
/
```

edsantiago:
That looks really good, thank you. I've got tests in the works, will submit PR as soon as yours merges.

rhatdan:
More tests.

```
$ ./bin/podman run -it --rm --userns=keep-id alpine grep dwalsh /etc/passwd
dwalsh:*:3267:3267:Daniel J Walsh:/:/bin/sh
$ ./bin/podman run -it -v /home/dwalsh --rm --userns=keep-id alpine grep dwalsh /etc/passwd
dwalsh:*:3267:3267:Daniel J Walsh:/home/dwalsh:/bin/sh
$ ./bin/podman run -it -v /home/dwalsh:/home/dwalsh --rm --userns=keep-id alpine grep dwalsh /etc/passwd
dwalsh:*:3267:3267:Daniel J Walsh:/home/dwalsh:/bin/sh
$ ./bin/podman run -it -v /home:/home --rm --userns=keep-id alpine grep dwalsh /etc/passwd
dwalsh:*:3267:3267:Daniel J Walsh:/home/dwalsh:/bin/sh
$ ./bin/podman run -it -v /home --rm --userns=keep-id alpine grep dwalsh /etc/passwd
dwalsh:*:3267:3267:Daniel J Walsh:/:/bin/sh
```

TomSweeneyRedHat:
@rhatdan looks like you're still on the bucking bronco of a test system.

edsantiago:
One is a network/Cirrus flake, I restarted. The [other](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5040643233808384/html/sys-podman-fedora-31-rootless-host.log.html) looks like #7139, which I'm starting to see a lot of these days. Restarted also.

vrothberg (commented):
LGTM

rhatdan:
@giuseppe @saschagrunert PTAL and merge.

saschagrunert (approved):
/lgtm

PR #8022, opened 2020-10-14
refactor api compatibility container creation to specgen

baude:
when using the compatibility layer to create containers, it used code paths to the pkg/spec which is the old implementation of containers.  it is error prone and no longer being maintained.  rather that fixing things in spec, migrating to specgen usage seems to make the most sense.  furthermore, any fixes to the compat create will not need to be ported later.

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -1,6 +1,15 @@
 package common
 
-import "github.com/containers/podman/v2/pkg/domain/entities"
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/containers/podman/v2/pkg/api/handlers"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/specgen"
+)
 
 type ContainerCLIOpts struct {
 	Annotation        []string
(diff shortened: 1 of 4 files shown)


cmd/podman/common/create_opts.go
```diff
@@ -111,3 +121,276 @@ type ContainerCLIOpts struct {
...
+			hostport, err := strconv.Atoi(pb.HostPort)
+			if err != nil {
+				return nil, nil, err
+			}
+			tmpPort := specgen.PortMapping{
+				HostIP:        pb.HostIP,
+				ContainerPort: uint16(port.Int()),
+				HostPort:      uint16(hostport),
+				Range:         0,
+				Protocol:      "", // TODO we have no protocol information on ports here. Should we do both?
```
mheon, line 230: I think protocol might be container in ContainerPort?

mheon: Contained in*

mheon:
One thing, otherwise LGTM. Less `pkg/spec` makes me very happy.

baude:
fixed

cmd/podman/common/create_opts.go
```diff
@@ -111,3 +121,276 @@ type ContainerCLIOpts struct {
...
+		Rm:                cc.HostConfig.AutoRemove,
+		SecurityOpt:       cc.HostConfig.SecurityOpt,
+		ShmSize:           strconv.Itoa(int(cc.HostConfig.ShmSize)),
+		StopSignal:        cc.Config.StopSignal,
+		StoreageOpt:       stringMaptoArray(cc.HostConfig.StorageOpt),
+		Sysctl:            stringMaptoArray(cc.HostConfig.Sysctls),
+		Systemd:           "true", // podman default
+		TmpFS:             stringMaptoArray(cc.HostConfig.Tmpfs),
+		TTY:               cc.Config.Tty,
+		//Ulimit:            cc.HostConfig.Ulimits,            // ask dan, no documented format
```
rhatdan, line 351: What happens when you uncomment this?

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

mheon:
One APIv2 test is failing

test/apiv2/20-containers.at
```diff
@@ -207,7 +207,7 @@ t DELETE containers/$cid 204
 t DELETE containers/$cid_top 204
 
 # test the apiv2 create, shouldn't ignore the ENV and WORKDIR from the image
-t POST containers/create '"Image":"'$ENV_WORKDIR_IMG'","Env":["testKey1"]' 201 \
+t POST containers/create '"Image":"'$ENV_WORKDIR_IMG'","Env":["testKey1=foo"]' 201 \
```
zhangguanzhang, line 210: https://github.com/containers/podman/blob/93af493ccb13b3d93717bf9c805f820c9e3a2fba/test/apiv2/test-apiv2#L154-L163
the line `sed -e 's/\(.*\)=\(.*\)/"\1":"\2"/' <<<$i` will convert `=` to `:`, so don't need `testKey1=foo`

cmd/podman/common/create_opts.go
nalind: Most likely this corresponds to the argument values described in https://github.com/containers/buildah/blob/master/docs/buildah-bud.md.

test/apiv2/20-containers.at
```diff
@@ -206,16 +206,6 @@ t POST containers/${cid_top}/stop "" 204
...
-# test the apiv2 create, shouldn't ignore the ENV and WORKDIR from the image
-t POST containers/create '"Image":"'$ENV_WORKDIR_IMG'","Env":["testKey1"]' 201 \
-  .Id~[0-9a-f]\\{64\\}
-cid=$(jq -r '.Id' <<<"$output")
-t GET containers/$cid/json 200 \
-  .Config.Env~.*REDIS_VERSION= \
-  .Config.Env~.*testKey1= \
-  .Config.WorkingDir="/data" # default is /data
-t DELETE containers/$cid 204
-
```
zhangguanzhang, line 218: cannot delete these lines

edsantiago: @zhangguanzhang the problem is that when you submitted your initial PR, APIv2 tests were disabled in CI. They weren't running at all. When we tried to reenable those tests, the newly-added ones broke. I never reviewed that PR, and right now it is hard to guess the original intention, so for now the best approach is just to delete them and then add something back later that will work. I am really sorry for the hassle.

zhangguanzhang: ok, I will submit these after enabled APIv2 test

zhangguanzhang:
LGTM

PR #8021, opened 2020-10-14
Fix indentation for `podman pod inspect`

mheon:
I fixed `podman inspect` in another PR, but it looks like `podman pod inspect` was also affected.

I did verify `podman network inspect` and `podman volume inspect` and both of those are still fine, though.

PR #8011, opened 2020-10-13
Restore indent on JSON from `podman inspect`

mheon:
I don't know when this was disabled, but it's very hard to read without it.

PR #182, opened 2018-01-03
Run by shortname

baude:
While pulling by shortname (fedora-minimal) worked, running a container
by the short name did not due to a logic error.

Signed-off-by: baude <bbaude@redhat.com>

PR #8025, opened 2020-10-14
Bats

edsantiago:
System test additions

 - run --userns=keep-id: confirm that $HOME gets set (#8013)

 - inspect: confirm that JSON output is a sane number of
   lines (10 or more), not an unreadable one-liner (#8011
   and #8021). Do so with image, pod, network, volume
   because the code paths might be different.

 - cgroups: confirm that 'run' preserves cgroup manager (#7970)

 - sdnotify: reenable tests, and hope CI doesn't hang. This
   test was disabled on August 18 because CI jobs were hanging
   and timing out. My suspicion was that it was #7316, which
   in turn seems to have hinged on conmon #182. The latter
   was merged on Sep 16, so let's cross our fingers and see
   what happens.

Also: remove inaccurate warning from a networking test.

And, wow, fix is_cgroupsv2(), it has never actually worked.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1717,11 +1717,35 @@ func (c *Container) generateCurrentUserPasswdEntry() (string, int, int, error) {
 	// If the user's actual home directory exists, or was mounted in - use
 	// that.
 	homeDir := c.WorkingDir()
-	if MountExists(c.config.Spec.Mounts, u.HomeDir) {
-		homeDir = u.HomeDir
+	hDir := u.HomeDir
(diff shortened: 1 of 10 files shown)


edsantiago:
WARNING! This is based on top of #8013. If that changes before merging, this PR will conflict. I'm doing this just so I can get the testing process started.

edsantiago:
/hold
CI is green, but please don't merge before #8013

vrothberg (commented):
LGTM

rhatdan:
/lgtm
/hold cancel
#8013 is merged.

PR #8029, opened 2020-10-15
Fix podman image trust show --raw output

Luap99:


Luap99:
@edsantiago Is this a known flake in APIv2 test?
```
not ok 53 [01-basic] Time for ten /info requests
#  expected: <= 7 seconds
#    actual: 8 seconds
```
https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6514321641439232/html/apiv2-podman-fedora-32-root-host.log.html

edsantiago:
@Luap99 something bad is going on with the APIv2 test. The timeout used to be 5 seconds for 10 `/info` calls. Just this week I bumped it up to 7 seconds because the timeout was failing (6). Now it's failing (8).

The reason this test is there is that at some earlier point the API got broken, with a delay on each call. I have no way of knowing if the current breakage is a similar delay in the API or if it's a slow CI machine. @jwhonce @baude are you aware of any new API code that could've reintroduced the delay? Can you please PTAL? I will be OOTO today with no network access.

rhatdan:
LGTM

test/e2e/trust_test.go
```diff
@@ -74,4 +74,18 @@ var _ = Describe("Podman trust", func() {
...
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.IsJSONOutputValid()).To(BeTrue())
+	})
+
+	It("podman image trust show --raw", func() {
+		session := podmanTest.Podman([]string{"image", "trust", "show", "--raw"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.IsJSONOutputValid()).To(BeTrue())
```
QiWang19, line 89: Can we verify some fields of the output string, like `Expect(session.OutputToString()).To(ContainSubstring("signed"))` above?

QiWang19 (commented):
LGTM

saschagrunert (approved):
LGTM

jwhonce:
@edsantiago If memory serves @baude included more network details this week

QiWang19:
May need an approval label to get this merged?

PR #8037, opened 2020-10-15
Make Podman-managed bind mounts unbindable

mheon:
This helps prevent potential mount loops when a user bind-mounts a large portion of the host (including bits that Podman uses) into a container using a shared mount propagation.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -343,7 +343,7 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 			Type:        "bind",
 			Source:      srcPath,
 			Destination: dstPath,
-			Options:     []string{"bind", "private"},
+			Options:     []string{"bind", "unbindable"},
 		}
 		if c.IsReadOnly() && dstPath != "/dev/shm" {
 			newMount.Options = append(newMount.Options, "ro", "nosuid", "noexec", "nodev")


TomSweeneyRedHat:
Does this help https://bugzilla.redhat.com/show_bug.cgi?id=1886827 - podman: containers fail to start with error "no space left on device" due to huge number of nested bind mounts (/run/.containerenv and /run/secrets) in container host's mount namespace

mheon:
It may. I have not yet confirmed that it resolves the issues.

rhvgoyal:
@mheon, which bind is being made unbindable here? "/run"? I mean makeing "/run" private should be enough to make sure that any mounts under it by container do not propagate to host.

rhvgoyal:
If we are making a mount unbindable after the mount "/run/secrets", I think its already too late because it will have propagated to host anyway by then

mheon:
@rhvgoyal This should make all bind mounts of single files that are directly managed by Podman (`/etc/hosts`, `/etc/resolv.conf`, `/etc/hostname`, `/run/.containerenv`, and a few more I'm forgetting) use `unbindable`. It doesn't touch `/run/secrets` yet, which lives in a separate place.

I think this is probably more correct than `private` propagation, even if it does not resolve the given BZ. I do want to verify that VM-based runtimes like Kata don't mind the mount option, though, it's a little atypical.

mheon:
I'll make a separate PR for the secrets mounts next week, and do more complete testing to see if this actually does resolve the situation.

rhvgoyal:
> @rhvgoyal This should make all bind mounts of single files that are directly managed by Podman (`/etc/hosts`, `/etc/resolv.conf`, `/etc/hostname`, `/run/.containerenv`, and a few more I'm forgetting) use `unbindable`. It doesn't touch `/run/secrets` yet, which lives in a separate place.
> 
> I think this is probably more correct than `private` propagation, even if it does not resolve the given BZ. I do want to verify that VM-based runtimes like Kata don't mind the mount option, though, it's a little atypical.

IIUC, "unbindable" says a mount is private. So submount under is will not propagate. Also unbindable will make sure that it can't be bind mounted.

So if you are mounting a file inside container at "/etc/hosts", and then making it unbindable, what problem is this solving. Are you trying to prevent somebody else from bind mount /etc/hosts further?

I am just not able to understand what problem you are solving and how unbindable is solving it. That's why asking all these questions, sorry.

mheon:
Looking more into this, I think you're right - this won't fix our issue. Closing.

PR #8042, opened 2020-10-16
--tls-verify and --authfile should work for all remote commands

rhatdan:
These options are now fully supported in the remote API and should no
longer be hidden and/or documented as non supported.

### DIFF
--- a/test/e2e/search_test.go
+++ b/test/e2e/search_test.go
@@ -237,7 +237,6 @@ registries = ['{{.Host}}:{{.Port}}']`
 	})
 
 	It("podman search attempts HTTP if registry is in registries.insecure and force secure is false", func() {
-		SkipIfRemote("--tls-verify is not supported on podman-remote search")
 		if podmanTest.Host.Arch == "ppc64le" {
 			Skip("No registry image for ppc64le")
 		}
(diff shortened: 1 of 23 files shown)


test/e2e/search_test.go
```diff
@@ -278,7 +281,6 @@ registries = ['{{.Host}}:{{.Port}}']`
 	})
 
 	It("podman search doesn't attempt HTTP if force secure is true", func() {
-		SkipIfRemote("--tls-verify is not supported on podman-remote search")
```
TomSweeneyRedHat, line 281: @edsantiago I just want to restate what a brilliant change adding this kind of text to the SkipIfRemote function is proving to be.  TYVM once again.

TomSweeneyRedHat (commented):
LGTM
once the redness is cleared from the tests.

rhatdan:
Some of these tests are still broken, but when I test by hand it seems to work fine.  I don't have time to fix figure out what is going on here, but would like to get this merged to actually fix the handling of the --tls-verify and --authfile flags.  

I will open an issue and have @QiWang19 Work on fixing the three tests.

QiWang19:
one of the error is the test flake. https://github.com/containers/podman/issues/7139. I rebased this PR based on the master branch but failed to reproduce that unnlinkat error.

jwhonce:
LGTM

tests appear to be timing out on podman info
```
[+0008s] not ok 53 [01-basic] Time for ten /info requests
[+0008s] #  expected: <= 7 seconds
[+0008s] #    actual: 8 seconds
```

PR #8043, opened 2020-10-16
Fix host to container port mapping for simple ranges

saschagrunert:
Signed-off-by: Sascha Grunert <sgrunert@suse.com>This fixes the issue that a simple port range should map 1:1 from the
host to the container. For example this fails without applying the
patch:

```
> podman run -it -p 6000-6066 alpine
Error: cannot listen on the TCP port: listen tcp4 :53: bind: address already in use
```

It tried to map a range with len 67 from port 0 on the host, which is
not what users expect. This regression is not included in v2.1.1.

Blocks https://github.com/containers/podman/pull/7948


rhatdan:
LGTM

saschagrunert:
So it looks like that the new intended behavior is to choose a random host port if we just specify `-p 80`? I'm wondering why we change this, because in v2.1.1 it behaves like a 1:1 mapping. I'm confused, it looks like that (if the randomness is intended) the port range is broken, since it starts at a random port but continues the range at port `0`…

rhatdan:
Should work the same was as Docker.

mheon:
Podman 1.9 and below would randomly assign if given only a single number,
not a host and container port pair - we broke this in 2.0 and are only
fixing it now.

mheon:
Though, it sounds like the behavior for a range is broken now. I'll verify
what the intended behavior is once I can get in front of a computer.

saschagrunert:
Yes, the range is broken like this:
```
> ./bin/podman run --log-level=debug --rm -it -p 6000-6066 alpine
…
DEBU[0000] rootlessport: time="2020-10-16T15:09:12+02:00" level=info msg="exposing ports [{42549 6000 tcp } {1 6001 tcp } {2 6002 tcp } {3 6003 tcp } {4 6004 tcp } {5 6005 tcp } {6 6006 tcp } {7 6007 tcp } {8 6008 tcp } {9 6009 tcp } {10 6010 tcp } {11 6011 tcp } {12 6012 tcp } {13 6013 tcp } {14 6014 tcp } {15 6015 tcp } {16 6016 tcp } {17 6017 tcp } {18 6018 tcp } {19 6019 tcp } {20 6020 tcp } {21 6021 tcp } {22 6022 tcp } {23 6023 tcp } {24 6024 tcp } {25 6025 tcp } {26 6026 tcp } {27 6027 tcp } {28 6028 tcp } {29 6029 tcp } {30 6030 tcp } {31 6031 tcp } {32 6032 tcp } {33 6033 tcp } {34 6034 tcp } {35 6035 tcp } {36 6036 tcp } {37 6037 tcp } {38 6038 tcp } {39 6039 tcp } {40 6040 tcp } {41 6041 tcp } {42 6042 tcp } {43 6043 tcp } {44 6044 tcp } {45 6045 tcp } {46 6046 tcp } {47 6047 tcp } {48 6048 tcp } {49 6049 tcp } {50 6050 tcp } {51 6051 tcp } {52 6052 tcp } {53 6053 tcp } {54 6054 tcp } {55 6055 tcp } {56 6056 tcp } {57 6057 tcp } {58 6058 tcp } {59 6059 tcp } {60 6060 tcp } {61 6061 tcp } {62 6062 tcp } {63 6063 tcp } {64 6064 tcp } {65 6065 tcp } {66 6066 tcp }]"
…
```

mheon:
The behaviour in 1.9 appears to be randomly assigning a port for every one in the range. We may want to take ranges without a host port assigned, and split them into individual ports.

cmd/podman/common/util.go
```diff
@@ -200,6 +200,10 @@ func parseSplitPort(hostIP, hostPort *string, ctrPort string, protocol *string)
 			}
 			newPort.HostPort = hostStart
 		}
+	} else {
+		// Assume a direct host to container mapping if the provided host port
+		// is nil.
```
TomSweeneyRedHat, line 205: Is this documented on a man page(s) somewhere?

TomSweeneyRedHat:
All kinds of test unhappiness.  This looks like something that should be documented on any affected man pages too.

cmd/podman/common/util.go
saschagrunert: I think this does not reflect the new implementation target. I'll remove this code part.

saschagrunert:
The failing use case should now work:
```
> ./bin/podman run --log-level=debug --rm -it -p 8080-8085 alpine 2>&1 | rg "exposing"
time="2020-10-19T11:25:34+02:00" level=debug msg="rootlessport: time=\"2020-10-19T11:25:34+02:00\" level=info msg=\"exposing ports [{41295 8080 tcp } {41296 8081 tcp } {41297 8082 tcp } {41298 8083 tcp } {41299 8084 tcp } {41300 8085 tcp }]\"\n"
```

mheon:
Didn't think we already had a test to cover this, but it looks like we do.

/approve
LGTM

Issue #7979, opened 2020-10-09
podman-remote run and podman-remote start --attach does not handle detaching with detach keys correctly


Luap99:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman-remote run` and `podman-remote start --attach` does not handle detaching with the detach keys `<ctrl-p><ctrl-q>` correctly.

**Steps to reproduce the issue:**

For `podman-remote run`:

1. Start the podman service

2. `podman-remote run -it alpine sh`

3. detach with `<ctrl-p><ctrl-q>`

For `podman-remote start --attach`:

1. Start the podman service

2. `podman-remote create -it --name test alpine sh`

3. `podman-remote start --attach test`

4. detach with `<ctrl-p><ctrl-q>`

**Describe the results you received:**

The process does not exit.

**Describe the results you expected:**

The process should exit with exitcode 0.

**Additional information you deem important (e.g. issue happens only occasionally):**

With `podman-remote start --attach` the error happens only if the container is not already started.
It also works fine with `podman-remote attach`

If I hit enter after detaching I get this error: `ERRO[0003] failed to write input to service: write unix @->/run/user/1000/podman/podman.sock: write: broken pipe`
So it looks like it does detach and closes the pipe but the process does not exit. It instead waits for the container to exit.

**Output of `podman version`:**

```
$ podman-remote version
Client:
Version:      2.2.0-dev
API Version:  2.0.0
Go Version:   go1.14.9
Git Commit:   71d675a164ee64111db1a9214673a3498b3aa123
Built:        Fri Oct  9 13:10:16 2020
OS/Arch:      linux/amd64

Server:
Version:      2.2.0-dev
API Version:  2.0.0
Go Version:   go1.14.9
Git Commit:   71d675a164ee64111db1a9214673a3498b3aa123
Built:        Fri Oct  9 13:10:05 2020
OS/Arch:      linux/amd64
```

andylibrian:
Let me look into this

/assign

PR #8045, opened 2020-10-16
Update podman-remote run and start to handle detach keys

andylibrian:
As described in the issue #7979, `podman-remote run` and `podman-remote start --attach` currently do not handle detach keys. This PR fixes the issue by handling the detach keys and exit cleanly with status 0.

### DIFF
--- a/pkg/bindings/containers/attach.go
+++ b/pkg/bindings/containers/attach.go
@@ -19,6 +19,7 @@ import (
 	"github.com/containers/podman/v2/pkg/bindings"
 	sig "github.com/containers/podman/v2/pkg/signal"
 	"github.com/containers/podman/v2/utils"
+	"github.com/moby/term"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


rhatdan:
Any chance of tests for this?

Luap99:
Thanks, @andylibrian.
I can confirm this fixes my issue.

andylibrian:
I can't seem to find existing tests that send detach keys. I attempted to write a test that send the keys but `ErrDetach` wasn't returned. Any pointer or advice would be great.

rhatdan:
I thought we could inspect this data, but turns out that these fields should not be supported in podman create and are not stored in the database.  Making this a lot more difficult to test.

I opened a PR to fix my misunderstanding.
https://github.com/containers/podman/pull/8053

/approve
/lgtm

rhatdan:
Thanks @andylibrian

PR #8046, opened 2020-10-16
Cirrus: Simplify setting/passing env. vars.

cevich:


cevich:
Note: The static-build failure will be dealt with in a separate PR

contrib/cirrus/lib.sh
```diff
@@ -108,9 +109,10 @@ bigto() { err_retry 7 5670 "" "$@"; }  # 12 minutes max
 
 # Print shell-escaped variable=value pairs, one per line, based on
 # variable name matching a regex.  This is intended to support
-# passthrough of CI variables from host -> container or from root -> user.
-# For all other vars. we rely on tooling to load this library from inside
-# the container or as rootless user to pickup the remainder.
+# passthrough of variables from Cirrus-CI for several different
+# testing setup situations (like contianer and rootless). Otherwise
+# Cirrus-CI operations always run as root on the system, so this is
+# not necesary.
```
cevich, line 115: note to me: get rid of this new comment addition, it's wrong and unnecessary (commented elsewhere).  Keep function description.

contrib/cirrus/setup_environment.sh
```diff
@@ -184,5 +190,16 @@ case "$TEST_FLAVOR" in
     *) die_unknown TEST_FLAVOR
 esac
 
-# Must be the very last command.  Establishes successful setup.
-echo 'export SETUP_ENVIRONMENT=1' >> /etc/environment
+# Must be the very last command.  Prevents setup from running twice.
+echo 'SETUP_ENVIRONMENT=1' >> /etc/ci_environment
+
+# Env. vars from Cirrus-CI and setup_environment.sh should
+# available to all users on the system.
```
cevich, line 197: note to me: comment from old iteration, re-word entirely.

cevich:
@edsantiago I had some time waiting for other tests to run so I threw this together, incorporating some of your expressed concerns over `/etc/environment`.  Hopefully you'll agree this is a much more POLA and robust way of doing things.

contrib/cirrus/setup_environment.sh
```diff
@@ -123,15 +128,16 @@ case "$PRIV_NAME" in
...
+            echo "$_suns" >> /etc/ci_environment
+            source /etc/ci_environment
         fi
         ;;
     rootless)
-        _ru="export ROOTLESS_USER='${ROOTLESS_USER:-some${RANDOM}dude}'"
-        echo "$_ru" >> /etc/environment
-        source /etc/environment
+        _ru="ROOTLESS_USER='${ROOTLESS_USER:-some${RANDOM}dude}'"
+        echo "$_ru" >> /etc/ci_environment
```
edsantiago, line 139: I was partly OK with this approach until getting here.

As you'll recall from my (abandoned) PR last week, my approach was to write user environment into a user file (in the rootless user's home directory). This to me guarantees that two CI jobs running on the same VM, one root and one rootless, will never step on each others' environments. With this line right here, we break that guarantee: a root job could run, read `/etc/ci_environment`, and get this setting. Or, worse, since there's a random element, get an inconsistent setting. (I'm not exactly sure how these systems are set up, and perhaps it's just one VM per job, but my point remains: a future maintainer might not know either, and should not know; the code should make it obvious that there can be no conflicts).

contrib/cirrus/setup_environment.sh
```diff
@@ -123,15 +128,16 @@ case "$PRIV_NAME" in
         if [[ "$TEST_ENVIRON" == "container" ]] && ((container)); then
             # There's no practical way to detect userns w/in a container
             # affected/related tests are sensitive to this variable.
-            _suns='export SKIP_USERNS=1'
-            echo "$_suns" >> /etc/environment
-            source /etc/environment
+            warn "Disabling usernamespace integration testing"
+            _suns='SKIP_USERNS=1'
+            echo "$_suns" >> /etc/ci_environment
+            source /etc/ci_environment
```
edsantiago, line 134: All these `echo / source` -- why? Why not just `SKIP_USERNS=1` or `ENV=VALUE`? Unless I'm missing something big (and I've looked very hard), the entire purpose of lines 198-202 at the end is to write all that environment. It really bothers me to see all this convolution, because it makes me wonder, what am I missing? 

I would really like to see all of these simplified to just `ENV=whatever`. If there is an important reason why it is necessary to define an intermediate variable which is used in only one place, then (riskily) append to a global /etc file, then source that file, I would suggest adding a big fat comment to each instance -- or, better yet, rethinking it and writing it in a clean readable way.

edsantiago (commented):
I really dislike the use of a `/etc` file for something that should only be necessary for rootless tests; especially given that root could pick up unwanted variables from it. Would you consider changing the convention to `/home/$ROOTLESS_USER/ci_environment`? 

A few other comments inline. Thanks.

contrib/cirrus/lib.sh
```diff
@@ -6,25 +6,24 @@
...
+# items every time this library is utilized.
+source /etc/profile
+source /etc/environment
+if [[ -r "/etc/ci_environment" ]]; then source /etc/ci_environment; fi
+USER="$(whoami)"
+HOME="$(getent passwd $USER | cut -d : -f 6)"
+# Some platforms set and make this read-only
+[[ -n "$UID" ]] || \
+    UID=$(getent passwd $USER | cut -d : -f 3)
+GID=$(getent passwd $USER | cut -d : -f 4)
```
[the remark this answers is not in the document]
cevich: Ahh,yep, it use to be used but you're right, that was dropped.

cevich:
> I really dislike the use of a /etc file for something that should only be necessary for rootless tests

This change isn't just for rootless tests, it's for ALL tests.  It's suppose to do what it says on the tin:  Consolidate all control variables into one-place for simplicity and debug-ability.  This way nobody ever has to guess where these are defined on the system for any given context.  It also makes logging them easy, separate from the "current" env. vars. printout.

cevich:
> It's suppose to do what it says on the tin

Oh, woops.  I thought I had filled out the commit message more than that...

contrib/cirrus/setup_environment.sh
cevich: Cleaned up.

contrib/cirrus/setup_environment.sh
cevich: > two CI jobs running on the same VM

The "single-use disposable" aspect is a core tenant.  Therefor if two job ever run on a VM, it's an orchestration bug by definition.

> a future maintainer might not know

There's a comment at...errr...looks like it got edited or lost in the shuffle.  I'll put it back, you're right, it's a point that must never get lost on any maintainer.

cevich:
rebase/force-push to include https://github.com/containers/podman/pull/8074

rhatdan:
LGTM
@edsantiago @baude PTAL

contrib/cirrus/setup_environment.sh
```diff
@@ -61,7 +79,7 @@ case "$CG_FS_TYPE" in
             # This is necessary since we've built/installed from source,
             # which uses runc as the default.
             warn "Forcing testing with crun instead of runc"
-            echo "export OCI_RUNTIME=crun" >> /etc/environment
+            echo "OCI_RUNTIME=crun" >> /etc/ci_environment
```
edsantiago, line 82: I really dislike how in some places (above this line) you just `echo >>/etc/ci_environment`, and in other places (below this line) you both set the envariable and then `echo` it. Aside from DRY, it makes me wonder why some envariables need that treatment and some don't. Could you just make them all a single `echo >>`, then `source /etc/ci_environment` at the end?

cevich: Oh, I didn't notice the extraneous echo's, they must be left over from previous implementation...hmmm,  actually, I think you're referring to the `export` then `echo`.  Okay, simplifying those...

contrib/cirrus/setup_environment.sh
```diff
@@ -91,26 +109,22 @@ case "$TEST_ENVIRON" in
     host)
         if [[ "$OS_RELEASE_ID" == "fedora" ]]; then
             # The e2e tests wrongly guess `--cgroup-manager cgroupfs`
-            msg "Forcing CGROUP_MANAGER=systemd"
-            _cgm="export CGROUP_MANAGER=systemd"
-            echo "$_cgm" >> /etc/environment
-            source /etc/environment
+            warn "Forcing CGROUP_MANAGER=systemd"
+            echo "CGROUP_MANAGER=$CGROUP_MANAGER" >> /etc/ci_environment
```
edsantiago, line 113: This isn't going to work

contrib/cirrus/setup_environment.sh
```diff
@@ -91,26 +109,22 @@ case "$TEST_ENVIRON" in
...
             modprobe ip6table_nat || :
             modprobe iptable_nat || :
         else
             # The e2e tests wrongly guess `--cgroup-manager systemd`
-            msg "Forcing CGROUP_MANAGER=cgroupfs"
-            _cgm="export CGROUP_MANAGER=cgroupfs"
-            echo "$_cgm" >> /etc/environment
-            source /etc/environment
+            warn "Forcing CGROUP_MANAGER=cgroupfs"
+            echo "CGROUP_MANAGER=$CGROUP_MANAGER" >> /etc/ci_environment
```
edsantiago, line 127: nor this

cevich: woopsie...

cevich: ...got it.  Good catch.

cevich:
Hey! tests all passed in a single go (no re-runs).  Addressed all of @edsantiago concerns.

PR #8053, opened 2020-10-17
podman create doesn't support creating detached containers

rhatdan:
Detached containers and detach keys are only created with the podman run, i
exec, and start commands.  We do not store the detach key sequence or the
detach flags in the database, nor does Docker. The current code was ignoreing
these fields but documenting that they can be used.

Fix podman create man page and --help output to no longer indicate that
--detach and --detach-keys works.

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -63,6 +63,8 @@ func runFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&runOpts.SigProxy, "sig-proxy", true, "Proxy received signals to the process")
 	flags.BoolVar(&runRmi, "rmi", false, "Remove container image unless used by other containers")
 	flags.UintVar(&runOpts.PreserveFDs, "preserve-fds", 0, "Pass a number of additional file descriptors into the container")
+	flags.BoolVarP(&runOpts.Detach, "detach", "d", false, "Run container in background and print container ID")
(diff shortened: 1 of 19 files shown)


Luap99:
LGTM
It looks like there are tests which used these (useless) flags.

cmd/podman/containers/run.go
```diff
@@ -76,7 +76,20 @@ func init() {
...
+			"detach", "d", false,
+			"Run container in background and print container ID",
+		)
+		flags.StringVar(
+			&cliVals.DetachKeys,
+			"detach-keys", containerConfig.DetachKeys(),
+			"Override the key sequence for detaching a container. Format is a single character `[a-Z]` or a comma separated sequence of `ctrl-<value>`, where `<value>` is one of: `a-cf`, `@`, `^`, `[`, `\\`, `]`, `^` or `_`",
+		)
+	}
 	flags := runCommand.Flags()
```
Luap99, line 91: Why the extra function? This could be added to the `runFlags` function (line 58).

test/e2e/wait_test.go
```diff
@@ -65,7 +65,7 @@ var _ = Describe("Podman wait", func() {
 		session.Wait(20)
 	})
 
-	It("podman container wait on latest container", func() {
+	FIt("podman container wait on latest container", func() {
```
Luap99, line 68:
```suggestion
	It("podman container wait on latest container", func() {
```

cmd/podman/containers/run.go
rhatdan: Thanks, I was rushing to get this done.  This is a much cleaner solution.

test/e2e/wait_test.go
rhatdan: oops

rhatdan:
@containers/podman-maintainers this is ready to merge.

mheon:
LGTM

PR #8056, opened 2020-10-18
Make invalid image name error more specific

xordspar0:
Previously, using an invalid image name would produce an error like
this:

    Error: error encountered while bringing up pod test-pod-0: invalid reference format

This message didn't specify that there was an problem with an image
name, and it didn't specify which image name had a problem if there were
multiple. Now the error reads:

    Error: error encountered while bringing up pod test-pod-0: Failed to parse image "./myimage": invalid reference format

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -341,7 +341,7 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
 		}
 		named, err := reference.ParseNormalizedNamed(container.Image)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrapf(err, "Failed to parse image %q", container.Image)
 		}
 		// In kube, if the image is tagged with latest, it should always pull
 		if tagged, isTagged := named.(reference.NamedTagged); isTagged {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
Thanks for the PR and the test.

rhatdan:

pkg/domain/infra/abi/play.go
```diff
@@ -341,7 +341,7 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
 		}
 		named, err := reference.ParseNormalizedNamed(container.Image)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrapf(err, "Failed to parse image \"%s\"", container.Image)
```
vrothberg, line 344: Using `%q` would be more idiomatic:
```suggestion
			return nil, errors.Wrapf(err, "Failed to parse image %q", container.Image)
```

vrothberg (commented):
LGTM
Just a non-blocking nit.

pkg/domain/infra/abi/play.go
xordspar0: Nice, I like that better. I updated my patch to use that.

QiWang19:
LGTM

TomSweeneyRedHat:
LGTM

PR #8061, opened 2020-10-19
System tests: remove some misleading 'run's

edsantiago:
The BATS 'run' directive is really quite obnoxious; for the
most part we really don't want to use it. Remove some uses
that snuck in last week, and remove one test (exists) that
can more naturally be piggybacked into an rm test.

While we're at it: in setup(), look for and delete stray
external (buildah) containers. This will be important if
any of the external-container tests fails; this way we
don't leave behind a state that causes subsequent tests
to fail.


vrothberg (commented):
LGTM

PR #8065, opened 2020-10-19
Tests: Fix common flakes, and improve apiv2 test log

edsantiago:
- apiv2 - the 'ten /info requests' test is flaking often,
  taking ~8 seconds (our limit is 7, up from 5 a few weeks
  ago). Brent suggested that the first /info call might be
  expensive, because it needs to access storage. So, let's
  prime it by running one /info outside the timing loop.
  And, because even that continues to fail, bump it up
  to 10 seconds and file #8076 to track the slowdown.

- toolbox test - WaitForReady() has timed out, even on one
  occasion causing a run failure because it failed 3 times.
  Solution: bump up timeout from 2s to 5s. Not really great,
  but CI systems are underpowered, and it's not unreasonable
  that 2s might be too low.

- sdnotify test - add a 'podman wait' between stop & rm.
  This may prevent a "cannot rm container as it is running"
  race condition.

While working on this, Brent and I noticed a few ways that
test-apiv2 logging can be improved:

- test name: when request is POST, display the jsonified
  parameters, not the original input ones. This should
  make it much easier to reproduce failures.

- use curl's "--write-out" option to capture http code,
  content type, and request time. We were getting the
  first two via grep from logged headers; this is cleaner.
  And there was no other way to get timing. We now include
  the timing as X-Response-Time in the log file.

- abort on *any* curl error, not just 7 (cannot connect).
  Any error at all from curl is bad news.

### DIFF
--- a/test/apiv2/01-basic.at
+++ b/test/apiv2/01-basic.at
@@ -59,7 +59,10 @@ t GET info 200                \
   .DefaultRuntime~.*$runtime  \
   .MemTotal~[0-9]\\+
 
-# Timing: make sure server stays responsive
+# Timing: make sure server stays responsive.
+# Because /info may need to check storage, it may be slow the first time.
+# Let's invoke it once to prime caches, then run ten queries in a timed loop.
+t GET info 200
 t0=$SECONDS
 for i in $(seq 1 10); do
     # FIXME: someday: refactor t(), separate out the 'curl' logic so we
@@ -70,7 +73,8 @@ t1=$SECONDS
 delta_t=$((t1 - t2))
 
 # Desired number of seconds in which we expect to run.
-want=7
+# FIXME: 10 seconds is a lot! PR #8076 opened to investigate why.
+want=10
 if [ $delta_t -le $want ]; then
     _show_ok 1 "Time for ten /info requests ($delta_t seconds) <= ${want}s"
 else
--- a/test/apiv2/test-apiv2
+++ b/test/apiv2/test-apiv2
@@ -179,7 +179,7 @@ function t() {
     # POST requests require an extra params arg
     if [[ $method = "POST" ]]; then
         curl_args="-d $(jsonify $1)"
-        testname="$testname [$1]"
+        testname="$testname [$curl_args]"
         shift
     fi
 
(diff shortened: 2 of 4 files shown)


edsantiago:
Flakes seen in PR #8061:

#### int podman ubuntu-19 root host prior-ubuntu

  2020-10-19T09:32:49  [main](https://api.cirrus-ci.com/v1/task/6253884286959616/logs/main.log)
*   [Toolbox-specific testing [It] podman create --userns=keep-id - entrypoint - adding user with useradd and then removing their password ](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6253884286959616/html/int-podman-ubuntu-19-root-host.log.html#t--podman-create---userns-keep-id---entrypoint---adding-user-with-useradd-and-then-removing-their-password--1)
*   [Toolbox-specific testing [It] podman create --userns=keep-id - entrypoint - modifying existing user with usermod - add to new group, change home/shell/uid ](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6253884286959616/html/int-podman-ubuntu-19-root-host.log.html#t--podman-create---userns-keep-id---entrypoint---modifying-existing-user-with-usermod---add-to-new-group--change-home-shell-uid--1)
*   [Toolbox-specific testing [It] podman create --userns=keep-id - entrypoint - adding user with useradd and then removing their password ](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6253884286959616/html/int-podman-ubuntu-19-root-host.log.html#t--podman-create---userns-keep-id---entrypoint---adding-user-with-useradd-and-then-removing-their-password--2)
  *   [[same]](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6253884286959616/html/int-podman-ubuntu-19-root-host.log.html#t--podman-create---userns-keep-id---entrypoint---adding-user-with-useradd-and-then-removing-their-password--3)

edsantiago:
Update: added one new flake fix, a `podman wait` intended to prevent a [flake I had not seen before](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6204933605163008/html/sys-podman-fedora-31-root-host.log.html):
```
# podman exec 645fbdd77d1c741c4273d345a6394644dbcfc6e1ead931a7ce65b1a573b1faf3 touch /stop
# podman rm 645fbdd77d1c741c4273d345a6394644dbcfc6e1ead931a7ce65b1a573b1faf3
Error: cannot remove container 645fbdd77d1c741c4273d345a6394644dbcfc6e1ead931a7ce65b1a573b1faf3 as it is running - running or paused containers cannot be removed without force: container state improper
```

edsantiago:
Well, [that settles that](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6123108304158720/html/apiv2-podman-fedora-32-root-host.log.html#t--00054). Priming `/info` does not resolve the ten-requests flake. This time it took 9 seconds.

I'm not sure what to do. The last time this happened (and the reason for the test), the delay was super obvious when running tests manually on my laptop. There was an actual lag between each command. This time, I'm not seeing that. Tests seem to run in real time, pausing only on slow actions.

edsantiago:
Pretty major overhaul from the original submission. Commit message, PR title, and PR description amended. Basic diffs are: upped the timeout from 7s to 10, and added better logging (including curl response time) in test-apiv2

rhatdan:
LGTM

zhangguanzhang:
LGTM

PR #8067, opened 2020-10-19
Ensure that hostname is added to hosts with net=host

### LINKED ISSUE #8054: hostname is (still) not added to /etc/hosts

/kind bug

**Description**

Despite issue #1745, I'm still not seeing the container hostname injected into `/etc/hosts` - I only realised this recently on trying to launch a Java-based container, and seeing an error regarding not being able to resolve the local hostname (... which is untrue - it resolves just fine, it just doesn't existing in /etc/hosts).

```
$ cat /etc/hosts | sed 's/#.*$//' | grep -v '^\s*$'
127.0.0.1       localhost
::1             localhost

$ cat /etc/hosts
# /etc/hosts: Local Host Database
#
# This file describes a number of aliases-to-address mappings for the for 
# local hosts that share this file.
#
# The format of lines in this file is:
#
# IP_ADDRESS    canonical_hostname      [aliases...]
#
#The fields can be separated by any number of spaces or tabs.
#
# In the presence of the domain name service or NIS, this file may not be 
# consulted at all; see /etc/host.conf for the resolution order.
#

# IPv4 and IPv6 localhost aliases
127.0.0.1       localhost
::1             localhost

#
# Imaginary network.
#10.0.0.2               myname
#10.0.0.3               myfriend
#
# According to RFC 1918, you can use the following IP networks for private 
# nets which will never be connected to the Internet:
#
#       10.0.0.0        -   10.255.255.255
#       172.16.0.0      -   172.31.255.255
#       192.168.0.0     -   192.168.255.255
#
# In case you want to be able to connect directly to the Internet (i.e. not 
# behind a NAT, ADSL router, etc...), you need real official assigned 
# numbers.  Do not try to invent your own network numbers but instead get one 
# from your network provider (if any) or from your regional registry (ARIN, 
# APNIC, LACNIC, RIPE NCC, or AfriNIC.)
#
```

This happens both when `podman` is and isn't executed with a `--hostname` parameter - but (and this could be key) all of this set of containers are being launched with `--network=host` (and as root) - is the hostname injection logic only working when container networking is private and/or when run as a non-root user?

(It doesn't appear to make a difference whether the user within the container is `root` or unprivileged)

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14.7
Built:        Sat Oct  3 11:01:42 2020
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 35a2fa83022e56e18af7e6a865ba5d7165fa2a4a'
  cpus: 8
  distribution:
    distribution: gentoo
    version: unknown
  eventLogger: file
  hostname: dellr330
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.8.9-gentoo
  linkmode: dynamic
  memFree: 1425133568
  memTotal: 8063447040
  ociRuntime:
    name: crun
    package: Unknown
    path: /usr/bin/crun
    version: |-
      crun version 0.15
      spec: 1.0.0
      +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 23662653440
  swapTotal: 25769787392
  uptime: 469h 28m 41.44s (Approximately 19.54 days)
registries:
  search:
  - docker.io
  - docker.pkg.github.com
  - quay.io
store:
  containerStore:
    number: 23
    paused: 0
    running: 22
    stopped: 1
  graphDriverName: overlay
  graphOptions:
    overlay.ignore_chown_errors: "false"
  graphRoot: /space/podman/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 1765
  runRoot: /space/podman/run
  volumePath: /space/podman/volumes
version:
  APIVersion: 2.0.0
  Built: 1601719302
  BuiltTime: Sat Oct  3 11:01:42 2020
  GitCommit: 9f6d6ba0b314d86521b66183c9ce48eaa2da1de2
  GoVersion: go1.14.7
  OsArch: linux/amd64
  Version: 2.1.1

```

Yes

mheon:
When a container uses --net=host the default hostname is set to the host's hostname. However, we were not creating any entries in `/etc/hosts` despite having a hostname, which is incorrect.

Fixes #8054

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1541,11 +1541,31 @@ func (c *Container) getHosts() string {
 		}
 	}
 
-	if c.config.NetMode.IsSlirp4netns() {
-		// When using slirp4netns, the interface gets a static IP
-		hosts += fmt.Sprintf("# used by slirp4netns\n%s\t%s %s\n", "10.0.2.100", c.Hostname(), c.Config().Name)
-	}
 	hosts += c.cniHosts()
+
+	// If not making a network namespace, add our own hostname.
+	if c.Hostname() != "" {
+		if c.config.NetMode.IsSlirp4netns() {
+			// When using slirp4netns, the interface gets a static IP
+			hosts += fmt.Sprintf("# used by slirp4netns\n%s\t%s %s\n", "10.0.2.100", c.Hostname(), c.config.Name)
+		} else {
+			hasNetNS := false
+			for _, ns := range c.config.Spec.Linux.Namespaces {
+				if ns.Type == spec.NetworkNamespace {
+					hasNetNS = true
+					break
+				}
+			}
+			if !hasNetNS {
+				// 127.0.1.1 and host's hostname to match Docker
+				osHostname, err := os.Hostname()
+				if err != nil {
+					osHostname = c.Hostname()
+				}
+				hosts += fmt.Sprintf("127.0.1.1 %s\n", osHostname)
+			}
+		}
+	}
 	return hosts
 }
 
--- a/pkg/specgen/generate/namespaces.go
+++ b/pkg/specgen/generate/namespaces.go
@@ -342,7 +342,7 @@ func specConfigureNamespaces(s *specgen.SpecGenerator, g *generate.Generator, rt
 				return errors.Wrapf(err, "error looking up container to share uts namespace with")
 			}
 			hostname = utsCtr.Hostname()
-		case s.NetNS.NSMode == specgen.Host || s.UtsNS.NSMode == specgen.Host:
+		case (s.NetNS.NSMode == specgen.Host && hostname == "") || s.UtsNS.NSMode == specgen.Host:
 			tmpHostname, err := os.Hostname()
 			if err != nil {
 				return errors.Wrap(err, "unable to retrieve hostname of the host")
--- a/test/e2e/run_networking_test.go
+++ b/test/e2e/run_networking_test.go
@@ -571,4 +571,19 @@ var _ = Describe("Podman run networking", func() {
 		podrm.WaitWithDefaultTimeout()
 		Expect(podrm.ExitCode()).To(BeZero())
 	})
+
+	It("podman run net=host adds entry to /etc/hosts", func() {
+		run := podmanTest.Podman([]string{"run", "--net=host", ALPINE, "cat", "/etc/hosts"})
+		run.WaitWithDefaultTimeout()
+		Expect(run.ExitCode()).To(BeZero())
+		Expect(strings.Contains(run.OutputToString(), "127.0.1.1")).To(BeTrue())
+	})
+
+	It("podman run with --net=host and --hostname sets correct hostname", func() {
+		hostname := "testctr"
+		run := podmanTest.Podman([]string{"run", "--net=host", "--hostname", hostname, ALPINE, "hostname"})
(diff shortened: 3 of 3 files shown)


zhangguanzhang:
seem docker don't have this?
```
[root@k8s-m1 ~]# docker run --hostname test111  --net host nginx:alpine cat /etc/hosts | grep test111
[root@k8s-m1 ~]# docker info
Client:
 Debug Mode: false

Server:
 Containers: 13
  Running: 9
  Paused: 0
  Stopped: 4
 Images: 22
 Server Version: 19.03.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: systemd
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: b34a5c8af56e510852c35414db4c1f4fa6172339
 runc version: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.8.9-1.el7.elrepo.x86_64
 Operating System: CentOS Linux 7 (Core)
 OSType: linux
 Architecture: x86_64
 CPUs: 1
 Total Memory: 1.943GiB
 Name: k8s-m1
 ID: EIUG:BTLZ:UK4M:P6C5:L2IW:UXEA:ZTL5:6CHJ:I3CL:G2UM:SGQ5:LG6S
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: zhangguanzhang
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Registry Mirrors:
  https://fz5yth0r.mirror.aliyuncs.com/
 Live Restore Enabled: false

```

mheon:
Docker is unconditionally forcing the entry in `/etc/hosts` to use the hostname of the host system, not the container. I'll adjust to match this.

mheon:
New changes: we now respect `--hostname` if passed, even if `--net=host` was given, and the hostname added to `/etc/hosts` when `--net=host` is set will always be the host system's hostname, not the container's hostname - both match Docker.

TomSweeneyRedHat:
code LGTM
any notes/changes to man pages needed?

rhatdan:
Shouldn't --net=host and --hostname conflict?  
I don't believe a container should be able to change the hosts namespace, unless it has NET_ADMIN capability?

mheon:
@rhatdan I don't think we are changing the host, since the container does have a UTS namespace in this case - it's just missing the net namespace.

rhatdan:
Ah OK, I guess they should conflict with the --uts=host, but no one ever does this.

PR #8069, opened 2020-10-20
fix: neutral value for MemorySwappiness

matejvasek:
Setting MemorySwappiness to -1, since that value is interpreted as "not set".

Otherwise there is issue if not running with cgroups v2 (invalid configuration, cannot specify resource limits without cgroups v2 and --cgroup-manager=systemd).


zhangguanzhang:
need go fmt, and need to rebase the commits to 1 commit

matejvasek:
@zhangguanzhang there is just single commit isn't it? I tried to run `go fmt` but it didn't change anything.

zhangguanzhang:
seem works

rhatdan:
LGTM

vrothberg (commented):
LGTM

PR #8078, opened 2020-10-20
APIv2 compatibility network connect|disconnect

baude:
Add endpoints for the compat layer for network connect and disconnect. As of now, these two endpoints do nothing to change the network state of a container.  They do some basic data verification and return the proper 200 response.  This at least allows for scripts to work on the compatibility layer instead of getting 404s.

### DIFF
--- a/libpod/events/filters.go
+++ b/libpod/events/filters.go
@@ -73,8 +73,8 @@ func generateEventUntilOption(timeUntil time.Time) func(e *Event) bool {
 }
 
 func parseFilter(filter string) (string, string, error) {
-	filterSplit := strings.Split(filter, "=")
-	if len(filterSplit) != 2 {
+	filterSplit := strings.SplitN(filter, "=", 2)
+	if len(filterSplit) == 1 {
 		return "", "", errors.Errorf("%s is an invalid filter", filter)
 	}
 	return filterSplit[0], filterSplit[1], nil
--- a/pkg/api/handlers/compat/networks.go
+++ b/pkg/api/handlers/compat/networks.go
@@ -20,6 +20,7 @@ import (
 	dockerNetwork "github.com/docker/docker/api/types/network"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 func InspectNetwork(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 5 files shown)


libpod/events/filters.go
```diff
@@ -72,21 +72,15 @@ func generateEventUntilOption(timeUntil time.Time) func(e *Event) bool {
 	}
 }
 
-func parseFilter(filter string) (string, string, error) {
-	filterSplit := strings.Split(filter, "=")
-	if len(filterSplit) != 2 {
-		return "", "", errors.Errorf("%s is an invalid filter", filter)
-	}
-	return filterSplit[0], filterSplit[1], nil
+func parseFilter(filter string) (string, string) {
```
mheon, line 75: This still needs to return (string, string, error). You need to check for length == 2 (otherwise we segfault on filterSplit[1] if the user passes an invalid filter)

pkg/api/handlers/compat/networks.go
```diff
@@ -289,6 +289,10 @@ func CreateNetwork(w http.ResponseWriter, r *http.Request) {
 	utils.WriteResponse(w, http.StatusOK, report)
 }
 
+// Disconnect removes a container from a network
+// TODO: For now this func is a no-op that checks the container name, network name, and
```
mheon, line 293: This comment is in the wrong place

pkg/api/handlers/compat/networks.go
```diff
@@ -312,3 +316,75 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if _, err = runtime.LookupContainer(netDisconnect.Container); err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
+			utils.ContainerNotFound(w, netDisconnect.Container, err)
+			return
+		}
+		utils.Error(w, "unable to lookup container", http.StatusInternalServerError, err)
+		return
+	}
+	utils.WriteResponse(w, http.StatusOK, "")
```
mheon, line 389: We need to write some sort of log here that the endpoint is unimplemented and do nothing

pkg/api/handlers/compat/networks.go
```diff
@@ -312,3 +316,75 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if _, err = runtime.LookupContainer(netConnect.Container); err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
+			utils.ContainerNotFound(w, netConnect.Container, err)
+			return
+		}
+		utils.Error(w, "unable to lookup container", http.StatusInternalServerError, err)
+		return
+	}
+	utils.WriteResponse(w, http.StatusOK, "")
```
mheon, line 355: We need to write some sort of log here that the endpoint is unimplemented and did nothing

rhatdan: I don't see a reason to fill the logs with this information. We documented it in the code.  Can we document this in the API?

mheon: Our users aren't going to be reading the API documentation or code documentation, they're going to be running apps agains the compat API and wondering why some calls return success but do not do what was asked.

pkg/api/handlers/compat/networks.go
```diff
@@ -312,3 +314,81 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
...
+	if _, err = runtime.LookupContainer(netConnect.Container); err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
+			utils.ContainerNotFound(w, netConnect.Container, err)
+			return
+		}
+		utils.Error(w, "unable to lookup container", http.StatusInternalServerError, err)
+		return
+	}
+	logrus.Warnf("network connect endpoint is not fully implemented - tried to connect container %s to network %s", netConnect.Container, name)
+	utils.WriteResponse(w, http.StatusOK, "")
```
jwhonce, line 354: A 200 should return a body at least "OK", unless this is for compatibility...

pkg/api/handlers/compat/networks.go
```diff
@@ -312,3 +314,81 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
...
+	if _, err = runtime.LookupContainer(netDisconnect.Container); err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
+			utils.ContainerNotFound(w, netDisconnect.Container, err)
+			return
+		}
+		utils.Error(w, "unable to lookup container", http.StatusInternalServerError, err)
+		return
+	}
+	logrus.Warnf("network disconnect endpoint is not fully implemented - tried to connect container %s to network %s", netDisconnect.Container, name)
+	utils.WriteResponse(w, http.StatusOK, "")
```
jwhonce, line 393: A 200 should return a body at least "OK", unless this is for compatibility...

pkg/api/server/register_networks.go
```diff
@@ -98,6 +98,63 @@ func (s *APIServer) registerNetworkHandlers(r *mux.Router) error {
...
+	//    required: true
+	//    description: the name of the network
+	//  - in: body
+	//    name: create
+	//    description: attributes for connecting a container to a network
+	//    schema:
+	//      $ref: "#/definitions/NetworkConnectRequest"
+	// responses:
+	//   200:
+	//     description: no error
```
jwhonce, line 122:
```suggestion
	//     description: OK
```

pkg/api/server/register_networks.go
```diff
@@ -98,6 +98,63 @@ func (s *APIServer) registerNetworkHandlers(r *mux.Router) error {
...
+	//    required: true
+	//    description: the name of the network
+	//  - in: body
+	//    name: create
+	//    description: attributes for disconnecting a container from a network
+	//    schema:
+	//      $ref: "#/definitions/NetworkDisconnectRequest"
+	// responses:
+	//   200:
+	//     description: no error
```
jwhonce, line 150:
```suggestion
	//     description: OK
```

pkg/api/handlers/compat/networks.go
```diff
@@ -9,6 +9,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
```
rhatdan, line 13: Remove line break

rhatdan (approved):
LGTM other then one nit.

baude:
all fixed, can i get a merge or hold merge?

pkg/api/handlers/compat/networks.go
```diff
@@ -312,3 +313,81 @@ func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
...
+	}
+	config, err := runtime.GetConfig()
+	if err != nil {
+		utils.InternalServerError(w, err)
+		return
+	}
+	name := utils.GetName(r)
+	exists, err := network.Exists(config, name)
+	if err != nil {
+		utils.InternalServerError(w, err)
```
mheon, line 337: Is this and no such container actually a 500? Feels like they should be 404

mheon: Ah, we do that below.

mheon:
LGTM

pkg/api/handlers/compat/networks.go
baude: ah! it turns out that both container and network are 404s.  Ill fix that.

baude: upon review, the code is correct.  Exists returns false, nil if it cannot find the network.  So this error check is actually correct. And the false, nil is handled in the next condition.

PR #8080, opened 2020-10-20
Cirrus: Double CPUs in default VM setup

cevich:
This should help all tests to execute faster.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -147,7 +147,7 @@ build_task:
     gce_instance: &standardvm
         image_project: libpod-218412
         zone: "us-central1-a"
-        cpu: 2
+        cpu: 4
         memory: "4Gb"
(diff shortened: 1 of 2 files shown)


edsantiago:
Doesn't look like it makes any difference for the [ten /info requsts test](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6748849035280384/html/apiv2-podman-fedora-32-root-host.log.html#t--00053):
```
ok 53 [01-basic] Time for ten /info requests (7 seconds) <= 7s
```
I can imagine that more CPUs could help the `int` tests, which run three in parallel, but it's not going to help `sys` or APIv2

cevich:
> I can imagine that more CPUs could help the int tests

Curiously, it doesn't seem to make much/any difference there either.

Closing this then since it would (literally) be a waste of money.

Issue #7213, opened 2020-08-04
[APIv2] Adding a container to a just 'Created' pod will start it but not 'start' event will be sent for the pod


KKoukiou:
/kind bug

**Description**

There is not 'start' event for pods, which got started after bring created by adding a running container to these. 

**Steps to reproduce the issue:**

1. $ podman pod create --name=user_one

2. $ sudo podman pod ps 
POD ID        NAME                  STATUS   CREATED        # OF CONTAINERS  INFRA ID
4d1b9950b7c8  user_one              Created  4 minutes ago  1                58f0cc683006

2. $ podman run -dit --pod user_one fedora bash

**Describe the results you received:**
 The user_one pod will now be Running but there is not event received.

**Describe the results you expected:**
Expected to receive `start` event for the pod.

**Output of `podman version`:**

```
podman version 2.0.4
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-2.0.4-1.fc32.x86_64
```

mheon:
We aren't technically starting the *pod* - we're starting the infra container for the pod. If there are other containers in the pod, we do not start them, only the infra container. As such, I tend to think that we should not be generating a "pod start" event.

KKoukiou:
> We aren't technically starting the _pod_ - we're starting the infra container for the pod. If there are other containers in the pod, we do not start them, only the infra container. As such, I tend to think that we should not be generating a "pod start" event.

I see what you are saying, but could you still emit a "pod start" event (or some event) when the PODS status changes from "Created" to "Running", which is what happens in this case?

mheon:
Maybe we need a better state for the pod here, to indicate it's only partially running - because you could run a `podman pod start` on a pod in that state, and it would still start containers within the pod.

mheon:
I'm going to bring this up at scrum tomorrow and ask for opinions from the team. I think we may end up with a new state for pods that are only partially running.

rhatdan:
I believe having another state makes sense.  Although I do not have a good name for that state.

mheon:
Kube makes me lean towards `Degraded` as a good name.

rhatdan:
Works for me.

rhatdan:
@mheon Did you ever have a chance to work on this?

mheon:
No - will get to it this week.

rhatdan:
@KKoukiou @mheon Is this still an issue or did the linked merge fix it?

rhatdan:
[...]

PR #8081, opened 2020-10-20
Add a Degraded state to pods

mheon:
Make a distinction between pods that are completely running (all containers running) and those that have some containers going, but not all, by introducing an intermediate state between Stopped and Running called Degraded. A Degraded pod has at least one, but not all, containers running; a Running pod has all containers running.

First step to a solution for #7213.

### DIFF
--- a/libpod/define/podstate.go
+++ b/libpod/define/podstate.go
@@ -10,9 +10,12 @@ const (
 	PodStateExited = "Exited"
 	// PodStatePaused indicates the pod has been paused
 	PodStatePaused = "Paused"
-	// PodStateRunning indicates that one or more of the containers in
-	// the pod is running
+	// PodStateRunning indicates that all of the containers in the pod are
+	// running.
 	PodStateRunning = "Running"
+	// PodStateDegraded indicates that at least one, but not all, of the
+	// containers in the pod are running.
+	PodStateDegraded = "Degraded"
 	// PodStateStopped indicates all of the containers belonging to the pod
 	// are stopped.
 	PodStateStopped = "Stopped"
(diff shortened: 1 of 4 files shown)


mheon:
Looks to be going green, @containers/podman-maintainers PTAL

rhatdan:
Could you add a test for this?

Setup/run a pod with two containers.
Get the pod state.
Stop on of the containers
Get the pod state
Stop the last container
Get the pod state.

mheon:
Sure, tests SGTM

vrothberg (commented):
LGTM

I think the description w.r.t. "running" are not 100 percent accurate. To me, "running" implies that the container is running and has not stopped. But in case of a pod, some container may have finished already (and hence stopped). Would "all containers have been started" be more accurate?

mheon:
Re-pushed with a test

vrothberg:
> I think the description w.r.t. "running" are not 100 percent accurate. To me, "running" implies that the container is running and has not stopped. But in case of a pod, some container may have finished already (and hence stopped). Would "all containers have been started" be more accurate?

@mheon can you reply? If I start a pod with two containers, and one of them exists, is the pod's state "degraded"?

mheon:
Yes. I want a clear distinction between all running and partially running.

If you want to make a further distinction between all running and some
naturally stopped (versus never started) we might could add another state,
though I have some concerns about how we'd distinguish (we don't have the
exit code at this point so I can't tell if the container exited cleanly,
and I don't think we can actually tell if the container exited on this run
of the pod and not the last time the pod was ran).

vrothberg:
@mheon, thanks!

> If you want to make a further distinction between all running and some
naturally stopped (versus never started) we might could add another state,

I _think_ that counting which containers have been started would be more accurate. Would it?

mheon:
My concern is what happens after the first time the pod is started and stopped - after that, all we have is a mix of Exited and Running, and it becomes very difficult to tell what's going on (did the user start one or two container but not the whole pod? Did the user start the whole pod, but half of it exited? We really can't tell, because the containers start in Exited state from the first time the pod was run, and end in Exited state after they finish).

Issue #8024, opened 2020-10-14
podman build --pull (--pull behaviour is different from docker)


asottile:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

the behaviour of `podman build --pull` differs from that of `docker build --pull` (the behaviour in docker is more like that of `podman build --pull-always`)

**Steps to reproduce the issue:**

1. have an outdated base image

2. `podman build --pull ...`

**Describe the results you received:**

It does not pull the base image before building

**Describe the results you expected:**

I expect it to pull the base image before building

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Built:        Wed Dec 31 16:00:00 1969
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
$ podman info --debug
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.20, commit: '
  cpus: 5
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: journald
  hostname: babibox
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.4.0-48-generic
  linkmode: dynamic
  memFree: 3150012416
  memTotal: 8348299264
  ociRuntime:
    name: runc
    package: 'cri-o-runc: /usr/lib/cri-o-runc/sbin/runc'
    path: /usr/lib/cri-o-runc/sbin/runc
    version: 'runc version spec: 1.0.2-dev'
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: 'slirp4netns: /usr/bin/slirp4netns'
    version: |-
      slirp4netns version 1.1.4
      commit: unknown
      libslirp: 4.3.1-git
      SLIRP_CONFIG_VERSION_MAX: 3
[...]

PR #8085, opened 2020-10-20
Fix pull method selection

TomSweeneyRedHat:
When using 'podman build --pull=true', the image was not pulled
if the image being pulled was present locally, but a newer version
was in the repository.  It was only pulled if there was no image
in local storage.

In addition, the pull-never option was ignored.  The line
`if flags.Pull{` at line 244 was always returning true
negating the default pullPolicy of PullNever.

Reworked the algorthim for the selection process.  Now
PullIfNewer is set to the default, and then we set the
pullPolicy appropriately based on the other flags
passed into this routine.

As an FYI, logic run in the calling functions ensures
that we have only one pull flag in the command.

Addresses: #8024

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -240,12 +240,16 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 		}
 	}
 
-	pullPolicy := imagebuildah.PullNever
-	if flags.Pull {
+	pullPolicy := imagebuildah.PullIfNewer
+	if c.Flags().Changed("pull") && !flags.Pull {
 		pullPolicy = imagebuildah.PullIfMissing
 	}
 	if flags.PullAlways {
 		pullPolicy = imagebuildah.PullAlways
+	}
+
+	if flags.PullNever {
+		pullPolicy = imagebuildah.PullNever
 	}
 
 	args := make(map[string]string)


cmd/podman/images/build.go
```diff
@@ -240,14 +240,18 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
...
+	if c.Flags().Changed("pull") && flags.Pull == false {
 		pullPolicy = imagebuildah.PullIfMissing
 	}
 	if flags.PullAlways {
 		pullPolicy = imagebuildah.PullAlways
 	}
 
+	if flags.PullNever {
+		pullPolicy = imagebuildah.PullNever
+	}
```
asottile, line 253: I'm not quite sure this is quite right either 🤔 

I think the logic should be:

```go
pullPolicy :=  imagebuildah.PullIfNewer

if c.Flags().Changed("pull") {
   if flags.Pull == false {
       pullPolicy = imagebuildah.PullNever
   } else {
       pullPolicy = imagebuildah.PullAlways
  }
}
if flags.PullAlways {
    pullPolicy = imagebuildah.PullAlways
}
if flags.PullNever {
    pullPolicy = imagebuildah.PullNever
}
```

that is:
- default: pull if newer
- `--pull` => pull always (like docker)
- `--pull=true` => pull always
- `--pull=false` => pull never
- `--pull-always` => pull always
- `--pull-never` => pull never

TomSweeneyRedHat: @asottile Maybe I need to retest, but the last time I looked at Docker it did:

```
default:  pull if missing or newer
--pull => pull if missing or newer
--pull=true => pull if missing or newer
--pull=false => pull if missing
--pull-always => pull always
--pull-never => pull never
```
I think I replicate the above with the current logic.    Personally even if different than Docker, I'd rather default, `--pull`, and `--pull=true` pull the image if missing or newer.   The `--pull=false` I could have sworn on Docker was a pull if missing too.  I'll double check that in the morning.

Thoughts from others?

asottile: docker doesn't have `--pull-always` / `--pull-never` and docker's `--pull` acts like `--pull-always`

baude:
lgtm

cmd/podman/images/build.go
mheon: So it sounds like we have one more fix needed here (`--pull` should be always instead of missing/newer) and we're good?

asottile:
this isn't correct and should not have merged

mheon:
@asottile Concur, but it should be a 1-line change to fix; let me spin up a PR for it.

mheon:
@asottile https://github.com/containers/podman/pull/8189

PR #8092, opened 2020-10-21
Podman build should default to not usins stdin

rhatdan:
Currently we leak stdin into podman builds, which can lead
to issues like run commands inside of the container waiting for
user input.

We should not take input from users other then if the user specifies
podman build -f - or podman build -, which are taken care of in other code.


mheon:
LGTM

rhatdan:
@nalind PTAL

nalind:
LGTM, mirrors https://github.com/containers/buildah/pull/2708.

Issue #8091, opened 2020-10-21
podman run --cidfile without --detach only writes the id after completion


FSMaxB-dooshop:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. `podman pull alpine` (or any container for that matter) 
2. `podman run -it --cidfile cid alpine sh` (keep the command running)
3. In another terminal: `cat cid`
4. Exit the shell in alpine, so that the container stops
5. `cat cid`

**Describe the results you received:**
In step 3, the `cid` file doesn't contain a container id, only in step 5.

**Describe the results you expected:**
In step 3 there should have already been a container id.

This is becaue the ID only gets written after running the container has finished, which is only after the container was stopped when not passing `--detach`:
https://github.com/containers/podman/blob/94873a237ab52db27916b8954e489fe780eea069/cmd/podman/containers/run.go#L188-L201

Ideally this should be fixed so that the cid is written for an already running container.
If this can't be easily fixed, or won't be fixed, at least the documentation should be updated to contain that information!

**Output of `podman version`:**

```
podman version 2.1.1
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 35a2fa83022e56e18af7e6a865ba5d7165fa2a4a'
  cpus: 48
  distribution:
    distribution: arch
    version: unknown
  eventLogger: journald
  hostname: max-threadripper-arch
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
  kernel: 5.9.1-arch1-1
  linkmode: dynamic
  memFree: 51220316160
  memTotal: 67374821376
  ociRuntime:
    name: runc
    package: Unknown
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc92
      commit: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
      spec: 1.0.2-dev
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
[...]

PR #8094, opened 2020-10-21
The cidfile should be created when the container is created

rhatdan:
Currently if you run an interactive session of podman run and
specifiy the --cidfile option, the cidfile will not get created
until the container finishes running.  If you run a detached
container, it will get created right away.  This Patch creates
the cidfile as soon as the container is created.  This could allow
other tools to use the cidefile on all running containers.

Fixes: https://github.com/containers/podman/issues/8091

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -638,3 +638,18 @@ func ValidateSysctls(strSlice []string) (map[string]string, error) {
 func DefaultContainerConfig() *config.Config {
 	return containerConfig
 }
+
+func CreateCidFile(cidfile string, id string) error {
+	cidFile, err := OpenExclusiveFile(cidfile)
+	if err != nil {
+		if os.IsExist(err) {
+			return errors.Errorf("container id file exists. Ensure another container is not using it or delete %s", cidfile)
+		}
+		return errors.Errorf("error opening cidfile %s", cidfile)
+	}
+	if _, err = cidFile.WriteString(id); err != nil {
+		logrus.Error(err)
+	}
+	cidFile.Close()
+	return nil
+}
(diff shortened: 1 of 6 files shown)


pkg/util/utils.go
```diff
@@ -638,3 +638,17 @@ func ValidateSysctls(strSlice []string) (map[string]string, error) {
 func DefaultContainerConfig() *config.Config {
 	return containerConfig
 }
+
+func OpenCidFile(cidfile string) (*os.File, error) {
+	if cidfile == "" {
+		return nil, nil
```
zhangguanzhang, line 644: may be need to new an error when `cidfile==""`?

rhatdan: Removed check, and just let it fall through to get error at next level.

mheon:
There has to be another set of logic for creating the files, as it does
(sort of) work now. I think you need to delete the existing code else we
might error on trying to make a duplicate CID file.

rhatdan:
@mheon I did remove the existing code, from run.go.  This was working but only on exit.  So podman run -d ... would work the way  you expect.  The issue was podman run -ti ... would only create the cidfile when you exited from the container.

rhatdan:
@mheon Sorry you were right, some of my changes did not make it into the original PR.

TomSweeneyRedHat:
A bunch of red marks on the tests @rhatdan

pkg/util/utils.go
```diff
@@ -638,3 +638,19 @@ func ValidateSysctls(strSlice []string) (map[string]string, error) {
...
+	if err != nil {
+		if os.IsExist(err) {
+			return errors.Errorf("container id file exists. Ensure another container is not using it or delete %s", cidfile)
+		}
+		return errors.Errorf("error opening cidfile %s", cidfile)
+	}
+	if _, err = cidFile.WriteString(id); err != nil {
+		logrus.Error(err)
+	}
+	cidFile.Sync()
```
giuseppe, line 653: the `Sync()` is not needed here.  If the node crashes the PID is not valid on the next boot anyway

mheon:
LGTM once tests start passing and comment from @giuseppe is addressed.

giuseppe (approved):
LGTM

PR #8096, opened 2020-10-21
Tests: Check different log driver can work with podman logs

ypu:
Add a check step in podman run --log-driver test. Prefer to add it here as it already has a loop to cover all different drivers
in this test.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -413,6 +413,17 @@ json-file | f
         else
             is "$output" "" "LogPath (driver=$driver)"
         fi
+
+        if [[ $driver != 'none' ]]; then
+            run_podman logs myctr
+            is "$output" "$msg" "check that podman logs works as expected"
+        else
+            run_podman 125 logs myctr
+            if ! is_remote; then
+                is "$output" ".*this container is using the 'none' log driver, cannot read logs.*" \
+                   "podman logs does not work with none log driver"
+            fi
+        fi
         run_podman rm myctr
     done < <(parse_table "$tests")
 


test/system/030-run.bats
```diff
@@ -413,6 +413,16 @@ json-file | f
         else
             is "$output" "" "LogPath (driver=$driver)"
         fi
+
+	run_podman ? logs myctr
```
rhatdan, line 417: Tab issue?

rhatdan (approved):

test/system/030-run.bats
ypu: Thanks for reminder. Updated.

test/system/030-run.bats
```diff
@@ -413,6 +413,16 @@ json-file | f
         else
             is "$output" "" "LogPath (driver=$driver)"
         fi
+
+        run_podman ? logs myctr
+        if [[ $driver != 'none' ]]; then
+            is "$output" "$msg" "check that podman logs works as expect"
```
TomSweeneyRedHat, line 419: nit, only if you need to make other changes:
```suggestion
            is "$output" "$msg" "check that podman logs works as expected"
```

test/system/030-run.bats
```diff
@@ -413,6 +413,16 @@ json-file | f
...
+            [[ $status == 0 ]]
```
edsantiago, line 420: The bats `[[` usage is evil. Please do this instead:
```bash
if [[ $driver == 'none' ]]; then
    run_podman 125 logs myctr
   ...
else
    run_podman logs myctr
   ...
fi
```

ypu:
Thanks @TomSweeneyRedHat and @edsantiago for your comments. Update the code, also fixed the check failed for --log-driver none in remote test.

edsantiago:
/lgtm

Thank you!

Issue #8082, opened 2020-10-20
Podman prints wrong image name in inspection when image has two tags


sshnaidm:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When an image has two different tags that points to the same layer, Podman prints wrong image information in container inspection.

**Steps to reproduce the issue:**

1.
```
podman run -dit --name bugtest docker.io/library/httpd:2
podman inspect bugtest | grep -i image
        "Image": "3dd970e6b110c8cbcec63e05a91e3cefd23c76a780fcb78c33979153f044b2d4",
        "ImageName": "docker.io/library/httpd:2",
            "Image": "docker.io/library/httpd:2",

podman rm -f bugtest
podman pull docker.io/library/httpd:latest
podman run -dit --name bugtest docker.io/library/httpd:2
podman inspect bugtest | grep -i image
        "Image": "3dd970e6b110c8cbcec63e05a91e3cefd23c76a780fcb78c33979153f044b2d4",
        "ImageName": "docker.io/library/httpd:latest",
            "Image": "docker.io/library/httpd:latest",
                "org.opencontainers.image.stopSignal": "28"
```

Images have the same Id:
```
$ podman image inspect httpd:2 | grep Id
        "Id": "3dd970e6b110c8cbcec63e05a91e3cefd23c76a780fcb78c33979153f044b2d4",
$ podman image inspect httpd:latest | grep Id
        "Id": "3dd970e6b110c8cbcec63e05a91e3cefd23c76a780fcb78c33979153f044b2d4",
```

**Describe the results you received:**

Wrong Image name in inspection.

**Describe the results you expected:**
When running `podman run -dit --name bugtest docker.io/library/httpd:2` and then `podman inspect bugtest` I'd like to get `"Image": "docker.io/library/httpd:2"` and not `"Image": "docker.io/library/httpd:latest"`, even if this is the same image actually.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.13.15
Built:        Wed Sep 30 22:51:15 2020
OS/Arch:      linux/amd64

```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.21-2.fc31.x86_64
[...]

PR #8098, opened 2020-10-21
container create: record correct image name

vrothberg:
Record the correct image name when creating a container by using the
resolved image name if present.  Otherwise, default to using the first
available name or an empty string in which case the image must have been
referenced by ID.

Fixes: #8082

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -469,7 +469,7 @@ func (ir *Runtime) getLocalImage(inputName string) (string, *storage.Image, erro
 		if err != nil {
 			return "", nil, err
 		}
-		img, err := ir.store.Image(ref.String())
+		img, err := ir.store.Image(reference.TagNameOnly(ref).String())
 		if err == nil {
 			return ref.String(), img, nil
 		}
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -473,4 +473,23 @@ json-file | f
     run_podman kill $cid
 }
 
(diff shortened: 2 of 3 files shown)


test/system/030-run.bats
```diff
@@ -473,4 +473,23 @@ json-file | f
     run_podman kill $cid
 }
 
+# Regression test for issue #8082
+@test "podman run : look up correct image name" {
```
vrothberg, line 477: @edsantiago PTAL

rhatdan:
LGTM

libpod/image/image.go
```diff
@@ -469,7 +469,7 @@ func (ir *Runtime) getLocalImage(inputName string) (string, *storage.Image, erro
 		if err != nil {
 			return "", nil, err
 		}
-		img, err := ir.store.Image(ref.String())
+		img, err := ir.store.Image(reference.TagNameOnly(ref).String())
```
vrothberg, line 472: @baude @mheon ... this should have a quite positive effect on container creation. If the user input wasn't tagged or digested, this lookup always failed, so we fell back to looking at all tags of all images (below in the code).

Now, we normalize to ":latest" which should have a noticeable performance boost on nodes with many images. It's also fixed in the short-names PR but it was necessary to get this fix working as well.

vrothberg:

mheon:
LGTM

TomSweeneyRedHat:
LGTM

PR #8101, opened 2020-10-21
Add hostname to /etc/hosts for --net=none

### LINKED ISSUE #8095: hostname is not added to /etc/hosts with network mode ‘none’

/kind bug

**Description**

Following on (hopefully for the last time!) from #8054, there’s potentially one last remaining use-case: if an image is launched with `—network none` for isolation purposes, the container hostname isn’t written to `/etc/hosts` and so certain applications (such as Java) which require this report that they can’t detect their own hostname.

It’s definitely an edge-case, since without networking the container can’t do much with the hostname it detects, and it’s arguably a Java issue for attempting to over-validate the hostname it’s given - but is there a case to write the container hostname to `/etc/hosts` even without networking, for the sake of completeness as much as anything?

Yes - with patch from #8054

mheon:
This does not match Docker, which does not add hostname in this case, but it seems harmless enough.

Fixes #8095

### DIFF
--- a/test/e2e/run_networking_test.go
+++ b/test/e2e/run_networking_test.go
@@ -584,6 +584,14 @@ var _ = Describe("Podman run networking", func() {
 		run := podmanTest.Podman([]string{"run", "--net=host", "--hostname", hostname, ALPINE, "hostname"})
 		run.WaitWithDefaultTimeout()
 		Expect(run.ExitCode()).To(BeZero())
-		Expect(strings.Contains(run.OutputToString(), "testctr")).To(BeTrue())
+		Expect(strings.Contains(run.OutputToString(), hostname)).To(BeTrue())
+	})
+
+	It("podman run with --net=none adds hostname to /etc/hosts", func() {
+		hostname := "testctr"
+		run := podmanTest.Podman([]string{"run", "--net=none", "--hostname", hostname, ALPINE, "hostname"})
+		run.WaitWithDefaultTimeout()
+		Expect(run.ExitCode()).To(BeZero())
+		Expect(strings.Contains(run.OutputToString(), hostname)).To(BeTrue())
 	})
 })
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Changes LGTM
I'm just leary about varying from Docker functionality...

test/e2e/run_networking_test.go
```diff
@@ -584,6 +584,14 @@ var _ = Describe("Podman run networking", func() {
...
-		Expect(strings.Contains(run.OutputToString(), "testctr")).To(BeTrue())
+		Expect(strings.Contains(run.OutputToString(), hostname)).To(BeTrue())
+	})
+
+	It("podman run with --net=none adds hostname to /etc/hosts", func() {
+		hostname := "testctr"
+		run := podmanTest.Podman([]string{"run", "--net=none", "--hostname", hostname, ALPINE, "hostname"})
+		run.WaitWithDefaultTimeout()
+		Expect(run.ExitCode()).To(BeZero())
+		Expect(strings.Contains(run.OutputToString(), hostname)).To(BeTrue())
```
edsantiago, line 595: What exactly is this test supposed to accomplish? Actually, what exactly is this PR supposed to accomplish? As I read the description and the code -- and of course I may just be confused -- I expected the given hostname to appear in the container's `/etc/hosts` file. That is not what I'm seeing:
```console
$ ./bin/podman run --net=none --hostname abcde alpine  cat /etc/hosts
127.0.0.1       localhost localhost.localdomain
::1             localhost localhost.localdomain
```

mheon: That sounds like a bug, because it should be in there

edsantiago: So, none of the code added in this PR is ever actually invoked, because of:
https://github.com/containers/podman/blob/7774f63319be64a0a96fb7ec7c9f0b7c28faac03/libpod/container_internal_linux.go#L1243
(This is what calls `generateHosts()`, and it's not called if `netDisabled`).

PR #8103, opened 2020-10-21
filter events by labels

baude:
adding the ability to filter evens by the container labels.  this requires that container labels be added to the events data being recorded and subsequently read.

### DIFF
--- a/libpod/events.go
+++ b/libpod/events.go
@@ -26,6 +26,12 @@ func (c *Container) newContainerEvent(status events.Status) {
 	e.Name = c.Name()
 	e.Image = c.config.RootfsImageName
 	e.Type = events.Container
+
+	e.Details = events.Details{
+		ID:         e.ID,
+		Attributes: c.Labels(),
+	}
+
 	if err := c.runtime.eventer.Write(e); err != nil {
 		logrus.Errorf("unable to write pod event: %q", err)
 	}
--- a/libpod/events/config.go
+++ b/libpod/events/config.go
@@ -36,6 +36,18 @@ type Event struct {
 	Time time.Time
 	// Type of event that occurred
 	Type Type
+
+	Details
+}
+
+// Details describes specifics about certain events, specifically around
+// container events
+type Details struct {
+	// ID is the event ID
+	ID string
+	// Attributes can be used to describe specifics about the event
+	// in the case of a container event, labels for example
+	Attributes map[string]string
 }
 
 // EventerOptions describe options that need to be passed to create
--- a/libpod/events/events.go
+++ b/libpod/events/events.go
@@ -69,7 +69,14 @@ func (e *Event) ToHumanReadable() string {
 	var humanFormat string
 	switch e.Type {
 	case Container, Pod:
-		humanFormat = fmt.Sprintf("%s %s %s %s (image=%s, name=%s)", e.Time, e.Type, e.Status, e.ID, e.Image, e.Name)
+		humanFormat = fmt.Sprintf("%s %s %s %s (image=%s, name=%s", e.Time, e.Type, e.Status, e.ID, e.Image, e.Name)
+		// check if the container has labels and add it to the output
+		if len(e.Attributes) > 0 {
+			for k, v := range e.Attributes {
+				humanFormat += fmt.Sprintf(", %s=%s", k, v)
+			}
+		}
+		humanFormat += ")"
 	case Image:
 		humanFormat = fmt.Sprintf("%s %s %s %s %s", e.Time, e.Type, e.Status, e.ID, e.Name)
 	case System:
--- a/libpod/events/filters.go
+++ b/libpod/events/filters.go
@@ -55,6 +55,24 @@ func generateEventFilter(filter, filterValue string) (func(e *Event) bool, error
 		return func(e *Event) bool {
 			return string(e.Type) == filterValue
 		}, nil
+
+	case "LABEL":
+		return func(e *Event) bool {
+			var found bool
+			// iterate labels and see if we match a key and value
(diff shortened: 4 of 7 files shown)


libpod/events.go
```diff
@@ -26,6 +26,11 @@ func (c *Container) newContainerEvent(status events.Status) {
 	e.Name = c.Name()
 	e.Image = c.config.RootfsImageName
 	e.Type = events.Container
+
+	details := events.Details{ID: e.ID}
```
rhatdan, line 30:
```
e.Details := events.Details{
             ID: e.ID,
             Attributes: c.Labels(),
}
```

libpod/events/journal_linux.go
```diff
@@ -46,6 +47,15 @@ func (e EventJournalD) Write(ee Event) error {
...
 			m["PODMAN_EXIT_CODE"] = strconv.Itoa(ee.ContainerExitCode)
 		}
+		// If we have container labels, we need to convert them to a string so they
+		// can be recorded with the event
+		if len(ee.Details.Attributes) > 0 {
+			b, err := json.Marshal(ee.Details.Attributes)
+			if err != nil {
+				return err
+			}
+			m["PODMAN_LABELS"] = string(b)
```
rhatdan, line 57: Should this be LIBPOD_LABELS?

rhatdan (approved):
Needs tests.

libpod/events/events.go
```diff
@@ -69,7 +69,14 @@ func (e *Event) ToHumanReadable() string {
 	var humanFormat string
 	switch e.Type {
 	case Container, Pod:
-		humanFormat = fmt.Sprintf("%s %s %s %s (image=%s, name=%s)", e.Time, e.Type, e.Status, e.ID, e.Image, e.Name)
+		humanFormat = fmt.Sprintf("%s %s %s %s (image=%s, name=%s", e.Time, e.Type, e.Status, e.ID, e.Image, e.Name)
```
mheon, line 72: Missing a closing paren in the string

mheon: Nevermind, I see where it gets added

libpod/events/config.go
```diff
@@ -36,6 +36,18 @@ type Event struct {
...
+}
+
+// Details describes specifics about certain events, specifically around
+// container events
+type Details struct {
+	// ID is the event ID
+	ID string
+	// Attributes can be used to describe specifics about the event
+	// in the case of a container event, labels for example
+	Attributes map[string]string
```
mheon, line 50: Can this be anything that is not a label? Wondering if we shouldn't just call this Labels...

libpod/events/config.go
```diff
@@ -36,6 +36,18 @@ type Event struct {
 	Time time.Time
 	// Type of event that occurred
 	Type Type
+
+	Details
```
mheon, line 40: Why do we need the embedded struct? Why not just put Attributes in here?

libpod/events/journal_linux.go
baude: no, precedent is already podman

libpod/events/config.go
baude: i believe it can be something that is not labels.

libpod/events/events.go
baude: look later in the code

libpod/events/config.go
baude: I was following the docker event structure to some degree.

mheon: I think this is overcomplicating matters, I don't see any case where we'll reuse this sub-struct.

baude:
updated and added system test

test/system/090-events.bats
```diff
@@ -0,0 +1,13 @@
...
+# tests for podman events functionality
+#
+
+load helpers
+
+@test "events with a filter by label" {
+    rand=$(random_string 30)
+    run_podman 0  run --label foo=bar --name test-$rand --rm $IMAGE ls
+    run_podman 0 events --filter type=container --filter container=test-$rand --filter label=foo=bar --filter event=start --stream=false
+    is "$output" ".*foo=bar" "check for label event on container with label"
```
edsantiago, line 12: You don't need the `0`, it's implicit. And I'd prefer a label of `foo=$random_label`, because it's just too easy for `foo=bar` to (e.g.) trigger on a previously-run container or something like that. But you done good :-), thank you for writing a test

libpod/events/filters.go
```diff
@@ -55,6 +55,24 @@ func generateEventFilter(filter, filterValue string) (func(e *Event) bool, error
...
 			return string(e.Type) == filterValue
 		}, nil
+
+	case "LABEL":
+		return func(e *Event) bool {
+			var found bool
+			// iterate labels and see if we match a key and value
+			for eventKey, eventValue := range e.Attributes {
+				filterValueSplit := strings.SplitN(filterValue, "=", 2)
+				// if the filter isnt right, just return false
```
TomSweeneyRedHat, line 65: nit of a nit to chase only if you've other edits
```suggestion
				// if the filter isn't right, just return false
```

test/system/090-events.bats
baude: I believe that having a random container name does the same thing.  This event looks for a start event on container with a label of foo=bar on a unique.

vrothberg (commented):
/hold
/lgtm

test/system/090-events.bats
edsantiago: Right -- assuming that the container filtering code is working as intended. (I get paid to be paranoid). I'll fix this in a followup PR.

PR #8104, opened 2020-10-21
set compat network driver default

baude:
when using the compatibility endpoint for creating a network, if the driver is not provided, we need to set it to the default network driver ... which is bridge.


mheon:
LGTM

PR #8107, opened 2020-10-22
Cirrus: Collect runner.sh stats

cevich:
On several occasions, there have been questions about CPU/Memory/IO trends in testing over time.  Start collecting this data for all jobs, using a common/stable format so that trending analysis can be performed within/across multiple Cirrus-CI builds.  This PR doesn't add any related tooling, it simply arranges for the collection of the data.

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -253,6 +253,16 @@ dotest() {
         |& logformatter
 }
 
+# Nearly every task in .cirrus.yml makes use of this shell script
+# wrapped by /usr/bin/time to collect runtime statistics.  Because the
+# --output option is used to log stats to a file, every child-process
+# inherits an open FD3 pointing at the log.  However, some testing
+# operations depend on making use of FD3, and so it must be explicitly
+# closed here (and for all further child-processes).
+# STATS_LOGFILE assumed empty/undefined outside of Cirrus-CI (.cirrus.yml)
+# shellcheck disable=SC2154
+exec 3<&-
+
 msg "************************************************************"
 # Required to be defined by caller
 # shellcheck disable=SC2154
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,6 +17,9 @@ env:
     CIRRUS_SHELL: "/bin/bash"
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
+    # Runner statistics log file path/name
+    STATS_LOGFILE_SFX: 'runner_stats.log'
+    STATS_LOGFILE: '$GOSRC/${CIRRUS_TASK_NAME}-${STATS_LOGFILE_SFX}'
 
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
@@ -75,7 +78,11 @@ ext_svc_check_task:
         TEST_FLAVOR: ext_svc
         CTR_FQIN: ${FEDORA_CONTAINER_FQIN}
     setup_script: &setup '$GOSRC/$SCRIPT_BASE/setup_environment.sh'
-    main_script: &main '$GOSRC/$SCRIPT_BASE/runner.sh'
+    main_script: &main '/usr/bin/time --verbose --output="$STATS_LOGFILE" $GOSRC/$SCRIPT_BASE/runner.sh'
+    always: &runner_stats
+        runner_stats_artifacts:
+            path: ./*-${STATS_LOGFILE_SFX}
+            type: text/plain
 
 
 # Execute some quick checks to confirm this YAML file and all
@@ -91,6 +98,7 @@ automation_task:
         TEST_ENVIRON: container
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # This task use to be called 'gating', however that name is being
@@ -126,6 +134,7 @@ smoke_task:
         make install.tools
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # N/B: This task is critical.  It builds all binaries and release archives
@@ -186,6 +195,7 @@ build_task:
     setup_script: *setup
     main_script: *main
     always: &binary_artifacts
+        <<: *runner_stats
         gosrc_artifacts:
             path: ./*  # Grab everything in top-level $GOSRC
             type: application/octet-stream
@@ -220,6 +230,7 @@ validate_task:
     clone_script: *noop
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # Exercise the "libpod" API with a small set of common
@@ -239,6 +250,7 @@ bindings_task:
     clone_script: *noop  # Comes from cache
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # Build the "libpod" API documentation `swagger.yaml` for eventual
@@ -275,6 +287,7 @@ endpoint_task:
     clone_script: *full_clone  # build-cache not available to container tasks
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # Check that all included go modules from other sources match
@@ -293,6 +306,7 @@ vendor_task:
     clone_script: *full_clone  # build-cache not available to container tasks
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # There are several other important variations of podman which
@@ -394,6 +408,7 @@ docker-py_test_task:
     clone_script: *noop  # Comes from cache
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 # Does exactly what it says, execute the podman unit-tests on all primary
@@ -412,6 +427,7 @@ unit_test_task:
     gopath_cache: *ro_gopath_cache
     setup_script: *setup
     main_script: *main
+    always: *runner_stats
 
 
 apiv2_test_task:
@@ -428,6 +444,7 @@ apiv2_test_task:
     setup_script: *setup
     main_script: *main
     always: &logs_artifacts
+        <<: *runner_stats
         # Required for `contrib/cirrus/logformatter` to work properly
         html_artifacts:
             path: ./*.html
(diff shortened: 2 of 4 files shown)


cevich:
@edsantiago not urgent, but PTAL when you have a second.  I'm slightly concerned about burying your `logformatter` URL above the `time --verbose` output.  Unf. `time` needs to run at a high-level, since `runner.sh` calls itself as a user and inside a container for several tests.  Is this worth worrying about?

If so, I was thinking to use the `--output=file` option, and teach `logcollector.sh` how to scoop it up.

edsantiago:
It's not clear to me what problem this is trying to solve. ISTR it had something to do with the Cirrus agent timeouts, but didn't you solve that by keeping the VMs around and looking at console logs? In what context will these timings be useful, and how?

(Please do not answer inline. If you still see a real use for this, please answer in the commit message, so a future maintainer can understand this)

cevich:
> It's not clear to me what problem this is trying to solve.

A valid point, I should probably spell that out:smile: 

It's true, resource use was originally suspected in the agent-stopped-responding problem.  However, there have been other cases where (especially memory usage) has been a concern - None of these VMs have swap, by design.    So this is mainly about retaining a record of CPU/Memory/Disk usage trends long-term, for high-level analysis.  That's why having it in the log output or in an artifact file doesn't matter...so long as it is consistent going forward.

cevich:
PR description updated.

edsantiago:
I would've preferred the actual git commit message (humans run `git log`, they seldom visit the PR web page) but that doesn't seem worth the suffering of going through CI again. What do you need from me? It LGTM, but the PR is marked Draft/WIP.

edsantiago:
OMG this just completely changed from a 1-liner to something super complicated. I withdraw my LGTM. Will take another look later.

cevich:
> I withdraw my LGTM. Will take another look later.

Yes please.  I decided to go the "store into file" route, since it doesn't bury your logformatter URL and (in theory) an artifact file should be easier to post-process if it's separates from other content.

cevich:
> I would've preferred the actual git commit message

I had thought the PR description becomes part of the merge-commit message, but looking back, it seems only the PR title is included :disappointed:   Since I'm re-running tests anyway, I'll update the commit message as well...

edsantiago:
Well, this is an interesting monkey wrench! `time`, with the `--output` option, uses fd3:
```console
$ bash -c '/usr/bin/time --verbose --output=/tmp/foo.log ls -l /proc/self/fd'                                                                         -
total 0
lrwx------. 1 esm esm 64 Nov  5 14:23 0 -> /dev/pts/11
lrwx------. 1 esm esm 64 Nov  5 14:23 1 -> /dev/pts/11
lrwx------. 1 esm esm 64 Nov  5 14:23 2 -> /dev/pts/11
l-wx------. 1 esm esm 64 Nov  5 14:23 3 -> /tmp/foo.log
```
The tests, though, expect a normal situation in which only stdin/stdout/stderr are open. 

I think the easiest, sanest way to work around this would be to add `exec 3<&-` to the top of `runner.sh`, with a HUGE comment explaining what it does and why.

cevich:
> Well, this is an interesting monkey wrench!

I'm more than happy to have provided at least some of the entertainment :rofl:  Wow, what a good catch!

So IIUC: Some tests/scripts/processes assume (4.e.g.) /proc/self/fd/3 is available where in fact /usr/bin/time has it open.  This causes said tests/scripts/processes to tuck-tail-twix-legs and run for the hills?

Geeze...well ya, closing the FD in the script, with a comment, is going to be far easier than filing a "bug" against /usr/bin/time :smiley: (assuming the necessary convincing is even attainable)

TBH, I don't think fixing this in `runner.sh` is such a bad thing...literally everything executes through that path (which is also how we get timing stats on everything).

contrib/cirrus/runner.sh
```diff
@@ -269,6 +269,20 @@ msg "************************************************************"
...
+# operations depend on making use of FD3, and so it must be explicitly
+# closed here (and for all further child-processes).
+if [[ -n "$STATS_LOGFILE" ]] && \
+    [[ "$(readlink /proc/self/fd/3)" == "$STATS_LOGFILE" ]]; then
+    warn "Closing FD 3 - open /usr/bin/time output log file"
+    exec 3<&-
+elif [[ -r "/proc/self/fd/3" ]]; then
+    warn "Found FD 3 open - this may negativly impact testing"
+fi
+
```
cevich, line 285: @edsantiago hopefully this is through enough :smile:

contrib/cirrus/runner.sh
```diff
@@ -269,6 +269,20 @@ msg "************************************************************"
...
 
+# Nearly every task in .cirrus.yml makes use of this shell script
+# wrapped by /usr/bin/time to collect runtime statistics.  Because the
+# --output option is used to log stats to a file, every child-process
+# inherits an open FD3 pointing at the log.  However, some testing
+# operations depend on making use of FD3, and so it must be explicitly
+# closed here (and for all further child-processes).
+if [[ -n "$STATS_LOGFILE" ]] && \
+    [[ "$(readlink /proc/self/fd/3)" == "$STATS_LOGFILE" ]]; then
+    warn "Closing FD 3 - open /usr/bin/time output log file"
```
edsantiago, line 280: Could I suggest not bothering with the if/warn stuff, and just doing a plain `exec 3<&-`? I see no reason whatsoever to do any checks nor to inform the user.

cevich: I think it's helpful in the case where the script is run manually after `hack/get_ci_vm.sh`.  It's also important because the cause (`/usr/bin/time --output`) is so physically distant from the mitigation.  I can easily envision a scenario where somebody tinkers with `.cirrus.yml` such that one of these warnings is the only clue as to missing expectations.  This then (hopefully) draws needed attention toward the comment.

edsantiago: I humbly suggest that anyone playing with fd3 is either (a) very smart and can figure it out, or (b) very stupid and deserves what they get. Please don't add complexity for a super-rare case.

cevich:
Stats-wise, using time in [front of the ssh client is basically useless.](https://cirrus-ci.com/task/5907934293000192?command=time#L3-L2).  This makes sense as a new session (process group) is created.  All the other cases seem to look okay, [including the "inside a container" case:](https://cirrus-ci.com/task/6611621734776832?command=time#L5).

contrib/cirrus/runner.sh
cevich: > very smart and can figure it out

Normally I would agree, except my experience is that virtually none of the podman golang-devs even try to look at the bash scripts.  They look at the e2e test results, *maybe* the log output, then panic and "ask Chris".

> very stupid and deserves what they get

That doesn't strike me as very compassionate toward future maintainers.  They are by definition "stupid" when it comes to the automation scripts.  

---

Point of fact:  Since `exec 3<&-` always works, even when not open, adding conditional warnings are the only method available to provide _any_ clues as to what might be happening.  When it comes to providing *some* possibly helpful  information vs remaining silent, even for a rare (but detectable) case: It's easier to get noticed in log output, than require digging through unfamiliar automation scripts.

---

Where is the compromise solution-space?

What if I remove the conditionals, and just leave a persistent warning message that specifically references `/proc/self/fd/3` and `$STATS_LOGFILE`?

Other options?

cevich:
@edsantiago I can understand the CPU/Memory being different for podman vs podman-remote, but does this seem reasonable/justifiable to you:

[sys podman fedora-33 root host](https://cirrus-ci.com/task/6470884246421504?command=time#L5) wall-clock time compared to [sys podman fedora-33 root host.](https://cirrus-ci.com/task/6752359223132160?command=time#L5)

In my mind, they should be virtually identical, no?

(just an observation)

edsantiago:
>sys podman fedora-33 root host wall-clock time compared to sys podman fedora-33 root host.

(I'm making the assumption that you intended to write "remote" in the second link, not "podman").

Are you asking, is something wrong with `time`? That seems unlikely, since the results it reports match those listed in the `main` accordion tab by Cirrus.

Are you asking, why do podman-remote tests run more quickly than podman-local? I believe that's because podman-remote has a lot of `skip`s. From a quick glance, I know that some of those tests (--help, podman login) take tens of seconds, so it seems plausible that that could explain some of the difference. I'm not sure I want to spend the effort to confirm test-by-test though.

If I misunderstood your question, I apologize, and could you please elaborate?

contrib/cirrus/runner.sh
edsantiago: Please do what you think is right. My comments were based on my suspicion that there is no possible scenario in which anyone will ever want to run e2e tests with fd3 intentionally open, compounded by my dislike of unnecessary unmaintainable confusing complexity. I am stepping back now and will comment no further.

cevich:
> (I'm making the assumption that you intended to write "remote" in the second link, not "podman").

Woops.

>  I believe that's because podman-remote has a lot of skips. From a quick glance, I know that some of those tests (--help, podman login) take tens of seconds, so it seems plausible that that could explain some of the difference. I'm not sure I want to spend the effort to confirm test-by-test though.

That answers my query, thanks.

contrib/cirrus/runner.sh
cevich: I appreciate your feedback, and I respect we're both trying to "do the right thing".  I'm just not always certain what that is.

I've rebased and force-pushed w/ only a warning message here.  I'll peep at the log output, then give it a last think before I call this ready for final review...

cevich: ...re-pushed with shorter warning message, re-positioned so it appears right after Cirrus-CI prints the command-line (which includes `/usr/bin/time`).

cevich: Hmmm...nope, seeing it in the logs now, I'm convinced you were right to suggest remaining silent.  Removing the warning message...

contrib/cirrus/runner.sh
```diff
@@ -253,6 +253,16 @@ dotest() {
...
 }
 
+# Nearly every task in .cirrus.yml makes use of this shell script
+# wrapped by /usr/bin/time to collect runtime statistics.  Because the
+# --output option is used to log stats to a file, every child-process
+# inherits an open FD3 pointing at the log.  However, some testing
+# operations depend on making use of FD3, and so it must be explicitly
+# closed here (and for all further child-processes).
+# STATS_LOGFILE assumed empty/undefined outside of Cirrus-CI (.cirrus.yml)
+# shellcheck disable=SC2154
```
edsantiago, line 263: You probably don't need this shellcheck directive any more, but it's not worth re-pushing for this.

edsantiago (approved):
LGTM

contrib/cirrus/runner.sh
cevich: woops.  I'll kill it next time I come across it.

edsantiago:
Sorry for late response, this was churning in CI when I last looked at it.

/lgtm
/hold

PR #8109, opened 2020-10-22
fix: /image/{name or id}/json returns RootFS layers

matejvasek:

### DIFF
--- a/pkg/api/handlers/types.go
+++ b/pkg/api/handlers/types.go
@@ -273,10 +273,12 @@ func ImageDataToImageInspect(ctx context.Context, l *libpodImage.Image) (*ImageI
 	}
 
 	rootfs := docker.RootFS{}
-	rootfs.Type = info.RootFS.Type
-	rootfs.Layers = make([]string, 0, len(info.RootFS.Layers))
-	for _, layer := range info.RootFS.Layers {
-		rootfs.Layers = append(rootfs.Layers, string(layer))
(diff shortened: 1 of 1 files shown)


mheon:
/approve
LGTM
@jwhonce @baude @vrothberg PTAL

rhatdan:
Thanks @matejvasek 
/lgtm
/hold

jwhonce:
LGTM

matejvasek:
I just added some nil checks, `info.RootFS` should be always non-`nil`, not sure about `info.RootFS.Layers`. But better save than sorry.

pkg/api/handlers/types.go
```diff
@@ -271,6 +271,15 @@ func ImageDataToImageInspect(ctx context.Context, l *libpodImage.Image) (*ImageI
 	if err != nil {
 		return nil, err
 	}
+
+	rootfs := docker.RootFS{}
+	if info.RootFS != nil && info.RootFS.Layers != nil {
```
TomSweeneyRedHat, line 276: Should this be split?  i.e. test RootFS for nil and then set RootFS.Type if it is not nil, then test for RootFS.Layers and set that if it is not nil.

matejvasek: I fix that. It turned out that it's OK to call `len()` and `range` on nil slice.

rhatdan:
Thanks @matejvasek 
/lgtm
/hold
/approve

PR #8112, opened 2020-10-22
Drop name argument from Load API

### LINKED ISSUE #7337: podman load -i Path-With-Caps label : invalid reference format: repository name must be lowercase

QiWang19:
Drop the name argument to Load API.  Splitting the code for load single and multiple images into two functions, LoadImageFromSingleImageArchive  LoadAllImageFromArchive
Specify in the document the usage of the optional name argument of the `podman load` is tagging an additional image.
Close #7337

```console
// docker-archive and oci-archive
$ podman load -i ap2.tar f00
Copying blob 364dcac827e6 done  
Copying config 9785808968 done  
Storing signatures
Loaded image(s): docker.io/library/alpine
$ podman images
REPOSITORY                TAG     IMAGE ID      CREATED     SIZE
docker.io/library/alpine  latest  9785808968ba  7 days ago  5.84 MB
localhost/f00             latest  9785808968ba  7 days ago  5.84 MB
```


rhatdan:
LGTM

libpod/runtime_img.go
```diff
@@ -295,27 +294,19 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
-		},
-		func() (types.ImageReference, error) {
-			// prepend "localhost/" to support local image saved with this semantics
-			if !strings.Contains(name, "/") {
-				return ociarchive.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			if name != "" {
+				return nil, errors.Errorf("oci format does not support additional name %s", name)
 			}
-			return nil, nil
+			return ociarchive.NewReference(inputFile, name)
```
TomSweeneyRedHat, line 300:
```suggestion
			return ociarchive.NewReference(inputFile, "")
```

libpod/runtime_img.go
```diff
@@ -295,27 +294,19 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
-		},
-		func() (types.ImageReference, error) {
-			// prepend "localhost/" to support local image saved with this semantics
-			if !strings.Contains(name, "/") {
-				return layout.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			if name != "" {
+				return nil, errors.Errorf("oci format does not support additional name %s", name)
 			}
-			return nil, nil
+			return layout.NewReference(inputFile, name)
```
TomSweeneyRedHat, line 309:
```suggestion
			return layout.NewReference(inputFile, "")
```

libpod/runtime_img.go
TomSweeneyRedHat: I just think it makes it a bit more definitive, first read through I was wondering why you checked for not an empty  string then passed the var.  YMMV

docs/source/markdown/podman-load.1.md
```diff
@@ -4,13 +4,12 @@
...
 ## SYNOPSIS
-**podman load** [*options*] [*name*[:*tag*]]
+**podman load** [*options*]
 
-**podman image load** [*options*] [*name*[:*tag*]]
+**podman image load** [*options*]
 
 ## DESCRIPTION
 **podman load** loads an image from either an **oci-archive** or a **docker-archive** stored on the local machine into container storage. **podman load** reads from stdin by default or a file if the **input** option is set.
-You can also specify a name for the image if the archive does not contain a named reference, of if you want an additional name for the local image.
```
vrothberg, line 13: I think we should keep that but document which formats support it.

libpod/runtime_img.go
```diff
@@ -295,27 +294,19 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
 		},
 		func() (types.ImageReference, error) {
-			return layout.NewReference(inputFile, name)
-		},
-		func() (types.ImageReference, error) {
-			// prepend "localhost/" to support local image saved with this semantics
-			if !strings.Contains(name, "/") {
-				return layout.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			if name != "" {
+				return nil, errors.Errorf("oci format does not support additional name %s", name)
```
vrothberg, line 307: That seems unrelated. Why's that needed?

vrothberg (commented):
I want a head nod from @mtrmac on it. The code is tricky and my eyes are tired :)

mtrmac (commented):
A quick impression is that this makes sense, still I’d like this PR to go further:

- **More extensive refactoring:**: AFAICS the `name` parameter is now _only_ used for a comparison with `""` (and an error message that is probably now very misleading and should be updated). So, should the callers be passing it in at all? The `name` parameter either shouldn’t exist, or it should exist and be used, or it should be replaced by a boolean (or possibly by splitting this function into two, `LoadImageFromSingleImageArchive` vs. `LoadAllImageFromArchive`, or whatever is best for the callers). And so on all the way up the call stack, including possibly dropping the parameter from the API.

- **Documentation:** While this probably resolves the ambiguity in actual behavior, will it be clear to future readers, and API users? What are we doing here, what does the `name` field mean after all? (Its mere _presence_ seems to have a non-obvious effect for `docker-archive`; and in the other formats we just don’t allow it instead of having a similar non-obvious effect. Is that right?) https://github.com/containers/podman/pull/7386#pullrequestreview-501854215 is not quite the level of detail needed here because the code does make more sense, but still broadly those categories of things would be nice to record for posterity _somewhere_ — at least for the API it should be clearly documented what the name/tag parameters mean. (Or if the parameters are removed, there will be nothing left to document, sure.)

---

To make sure it is not forgotten, https://github.com/containers/podman/pull/7386#pullrequestreview-501854215 says this should be dropped in Podman 3.0.  Is this the right time/branch?

libpod/runtime_img.go
```diff
@@ -295,27 +294,19 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
...
 		},
 		func() (types.ImageReference, error) {
-			return ociarchive.NewReference(inputFile, name) // name may be ""
-		},
-		func() (types.ImageReference, error) {
-			// prepend "localhost/" to support local image saved with this semantics
-			if !strings.Contains(name, "/") {
-				return ociarchive.NewReference(inputFile, fmt.Sprintf("%s/%s", image.DefaultLocalRegistry, name))
+			if name != "" {
+				return nil, errors.Errorf("oci format does not support additional name %s", name)
```
mtrmac, line 298: BTW this will never be seen by users, the way `err` is currently maintained by the loop. So the UI might need a bit more thought (or perhaps just give up and return `nil, nil` here — probably not, but that does depend on what the refactoring in callers ends up doing).

QiWang19:
> To make sure it is not forgotten, [#7386 (review)](https://github.com/containers/podman/pull/7386#pullrequestreview-501854215) says this should be dropped in Podman 3.0. Is this the right time/branch?

This branch is not meant to drop it. This patch wants to keep the support for docker-archives but removes the description from the document before Podman 3.0.

mtrmac:
> This branch is not meant to drop it. This patch wants to keep the support for docker-archives but removes the description from the document before Podman 3.0.

Ah, right, that comment was about removing the “tag the result with $this name” parameter, while this removes the “load $this image from an OCI archive” functionality. Still, is that OK to do before 3.0? Or was it completely broken already?

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac PTAL

QiWang19:

rhatdan:
LGTM, but this is API Breakage so it has to wait til we branche for 3.0.

rhatdan:
@QiWang19 Please rebase and we can merge.

QiWang19:

PR #8121, opened 2020-10-23
set resources only when specified

baude:
when using the compatibility endpoint to create a container, we should only set certain resources when we are provided a value for them or we result in fields with zero values.

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -293,66 +293,61 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig) (*Cont
 	// like with start. We believe this is just a difference in podman/compat
 	cliOpts := ContainerCLIOpts{
 		//Attach:            nil, // dont need?
-		Authfile:          "",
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

cmd/podman/common/create_opts.go
```diff
@@ -363,10 +358,37 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig) (*Cont
...
 	}
 
+	if len(cc.HostConfig.BlkioWeightDevice) > 0 {
+		devices := make([]string, 0, len(cc.HostConfig.BlkioWeightDevice))
+		for _, d := range cc.HostConfig.BlkioWeightDevice {
+			devices = append(devices, d.String())
+		}
+		cliOpts.BlkIOWeightDevice = devices
+	}
+	if cc.HostConfig.BlkioWeight > 0 {
```
TomSweeneyRedHat, line 368: to keep in alpha order, this if should be above BlkioWeightDevice

Issue #6050, opened 2020-04-30
Docker API /containers/{id}/archive


abitrolly:
/kind bug

Accessing `/containers/{id}/archive` on 1.9.1 gives 404. I guess it is not implemented yet.

https://docs.docker.com/engine/api/v1.38/#operation/ContainerArchive

mheon:
Yep, a few remote endpoints (those for copy, exec, and I believe the network commands) are still being worked on.

abitrolly:
@mheon can't wait for it to use https://github.com/buildpacks/pack/issues/564

baude:
@abitrolly https://docs.docker.com/engine/api/v1.40/#operation/ContainerArchiveInfo is this what you are after?  if so, you are correct, this is a bug.  would you be willing to contribute?

abitrolly:
I am not sure. While it seems pretty interesting, my motivation is still pretty low as I am not backed by any big company and on the personal scale I also won't be able to use it, because `podman` 2.0 is not ready, and API fixes are not backported to 1.9.x

rhatdan:
@abitrolly Does the master branch work for you now?

mheon:
This is not implemented yet. We added the endpoints, but they just throw "not implemented" errors for now.

Definitely on our radar, though, but will probably be combined with a rewrite of CLI `podman cp`

mheon:
Changing to kind/feature as this is a new feature

vrothberg:
@lsm5 are you interested in looking into it? I think we can start with adding a docker/compat endpoint first.

AFAIUI we have to mount the container and the archive the specified path inside the container and throw it back over the wire.

mheon:
@vrothberg This is part of the copy work, so I'll be on this one once we're off bug week.

vrothberg:
@mheon, sorry, I didn't notice you just self-assigned yesterday.

yan12125:
I noticed that the API endpoint to upload an archive in Podman uses different HTTP method than the corresponding API in Docker. The former uses POST and the latter uses PUT. Is the difference intentional?

[1] https://docs.podman.io/en/latest/_static/api.html#operation/putArchive
[2] https://docs.docker.com/engine/api/v1.40/#operation/PutContainerArchive

jwhonce:
@yan12125 That is a bug on podman's part.  /cc @mheon

yan12125:
[...]

PR #8126, opened 2020-10-24
Implement containers/{id or name}/archive api

matejvasek:
This is heavily in progress, but I could use any good advice.

### DIFF
--- a/pkg/api/handlers/compat/containers_archive.go
+++ b/pkg/api/handlers/compat/containers_archive.go
@@ -173,7 +173,7 @@ func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, 
 	}
 
 	defer func() {
-		if err := ctr.Unmount(true); err != nil {
+		if err := ctr.Unmount(false); err != nil {
 			logrus.Warnf("failed to unmount container %s", ctrName)
 		}
 	}()
--- a/pkg/api/handlers/compat/containers_archive.go
+++ b/pkg/api/handlers/compat/containers_archive.go
@@ -49,6 +49,7 @@ func handleHeadOrGet(w http.ResponseWriter, r *http.Request, decoder *schema.Dec
 		utils.Error(w, "Bad Request.", http.StatusBadRequest, errors.Wrap(err, "couldn't decode the query"))
 		return
 	}
+
 	if query.Path == "" {
 		utils.Error(w, "Bad Request.", http.StatusBadRequest, errors.New("missing `path` parameter"))
 		return
@@ -64,11 +65,13 @@ func handleHeadOrGet(w http.ResponseWriter, r *http.Request, decoder *schema.Dec
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, err)
 		return
 	}
+
 	mountPoint, err := ctr.Mount()
 	if err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to mount the container"))
 		return
 	}
+
 	defer func() {
 		if err := ctr.Unmount(true); err != nil {
 			logrus.Warnf("failed to unmount container %s: %q", containerName, err)
@@ -88,14 +91,18 @@ func handleHeadOrGet(w http.ResponseWriter, r *http.Request, decoder *schema.Dec
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to get stats about file"))
 		return
 	}
+
 	if len(stats) <= 0 || len(stats[0].Globbed) <= 0 {
 		errs := make([]string, 0, len(stats))
+
 		for _, stat := range stats {
 			if stat.Error != "" {
 				errs = append(errs, stat.Error)
 			}
 		}
+
 		utils.Error(w, "Not found.", http.StatusNotFound, fmt.Errorf("file doesn't exist (errs: %q)", strings.Join(errs, ";")))
+
 		return
 	}
 
@@ -114,6 +121,7 @@ func handleHeadOrGet(w http.ResponseWriter, r *http.Request, decoder *schema.Dec
 				errors.Wrapf(err, "error getting IDMappingOptions"))
 			return
 		}
+
 		destOwner := idtools.IDPair{UID: os.Getuid(), GID: os.Getgid()}
 
 		opts := copier.GetOptions{
@@ -125,6 +133,7 @@ func handleHeadOrGet(w http.ResponseWriter, r *http.Request, decoder *schema.Dec
 		}
 
 		w.WriteHeader(http.StatusOK)
+
 		err = copier.Get(mountPoint, "", opts, []string{filepath.Join(mountPoint, path)}, w)
 		if err != nil {
 			logrus.Error(errors.Wrapf(err, "failed to copy from the %s container path %s", containerName, query.Path))
@@ -156,11 +165,13 @@ func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, 
 		utils.Error(w, "Not found", http.StatusNotFound, errors.Wrapf(err, "the %s container doesn't exists", ctrName))
 		return
 	}
+
 	mountPoint, err := ctr.Mount()
 	if err != nil {
 		utils.Error(w, "Something went wrong", http.StatusInternalServerError, errors.Wrapf(err, "failed to mount the %s container", ctrName))
 		return
 	}
+
 	defer func() {
 		if err := ctr.Unmount(true); err != nil {
 			logrus.Warnf("failed to unmount container %s", ctrName)
@@ -172,11 +183,13 @@ func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, 
 		utils.Error(w, "Something went wrong", http.StatusInternalServerError, err)
 		return
 	}
+
 	idMappingOpts, err := ctr.IDMappings()
 	if err != nil {
 		utils.Error(w, "Something went wrong", http.StatusInternalServerError, errors.Wrapf(err, "error getting IDMappingOptions"))
 		return
 	}
+
 	destOwner := idtools.IDPair{UID: int(user.UID), GID: int(user.GID)}
 
 	opts := copier.PutOptions{
@@ -193,6 +206,7 @@ func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, 
 	}
 
 	w.WriteHeader(http.StatusOK)
+
 	err = copier.Put(mountPoint, filepath.Join(mountPoint, path), opts, r.Body)
 	if err != nil {
 		logrus.Error(errors.Wrapf(err, "failed to copy to the %s container path %s", ctrName, query.Path))
@@ -222,10 +236,12 @@ func statsToHeader(stats *copier.StatForItem) (string, error) {
 
 	buff := bytes.NewBuffer(make([]byte, 0, 128))
 	base64encoder := base64.NewEncoder(base64.StdEncoding, buff)
+
 	_, err = base64encoder.Write(jsonBytes)
 	if err != nil {
 		return "", err
 	}
+
 	err = base64encoder.Close()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


matejvasek:
I duplicate some code form `cp.go`.

Maybe I could create some shared package that would be used by both api and cli.

At first I was thinking of calling
```go
func (ic *ContainerEngine) ContainerCp(ctx context.Context, source, dest string, options entities.ContainerCpOptions) (*entities.ContainerCpReport, error)
```
, but that would require creating unnecessary temporary files.

matejvasek:
relates to #6050

matejvasek:
I noticed that `podman-cp` does evaluate symlinks.
Meaning
```shell
# alink -> a
podman cp xyz:/adir/alink .
```
Will copy file that the link points to, i.e user ends up with the `a` file in `cwd`.
However `docker` (and the API v2) doesn't evaluate symlinks.
This may affect may ability to reuse code from `cp.go` used by CLI :disappointed:

mheon:
@matejvasek When I was looking into implementing this, I was using https://github.com/containers/buildah/tree/master/copier - it's a new package designed for secure copy mimicking Docker behavior. The intention was to implement the archive endpoints using it, then swap `podman cp` over to use the same code.

mheon:
I was using https://github.com/containers/buildah/blob/master/copier/copier.go#L248 for copying out of, and https://github.com/containers/buildah/blob/master/copier/copier.go#L290 for copying into, to be specific.

matejvasek:
@mheon how much have you done implementing it? Does it make sense for me to continue on this PR?

matejvasek:
That Buildah pkg looks nice, it also has `Stats` for `HEAD` method.

mheon:
@matejvasek Feel free to continue, I didn't get very far - got pulled off quickly in favor of other work.

matejvasek:
CLI `podman-cp` allows to optionally pause a container. What APIv2 endpoint should do, pause or not pause?

matejvasek:
I noticed that `docker cp` doesn't create non-existent directories. e.g `docker cp afile xyz:/some/non/existent/dir` will fail. However `copier.Put()` is creating them, is that OK? Another thing I noticed: `podman-cp` also creates missing dirs on path, but the created directories are owned by `root`, whereas `copier.Put()` uses correct user which is nice.

matejvasek:
also with regard to symlinks: when getting a sole symlink directly by `copier.Get()` the method writes the file that the symlink points to into io.Writer (not the symlink itself). Docker doesn't do that, however getting a sole symlink doesn't really make a sense, so I think that's OK. When downloading a directory the symliks are downloaded as expected (as symlinks).

mheon:
Pause was an attempt at fixing a CVE that turned out to not be strictly required; I don't believe it's necessary here.

mheon:
@matejvasek FYI, we're fine with merging this without support for copying into volumes, at least initially - we can resolve that later, once the basic functionality is in place.

matejvasek:
@mheon @baude I rebased and added support for volumes.

matejvasek:
It still has one issue, when calling PUT the file is not visible immediately in running container, it's required to stop and start the container again to see the file.
/cc @nalind

matejvasek:
> It still has one issue, when calling PUT the file is not visible immediately in running container, it's required to stop and start the container again to see the file.

I noticed that `buildah copy` has the same issue. Maybe I won't be able to use `copier` pkg, and I will have to use raw `archive` pkg.

mheon:
That sounds very strange... @nalind Any thoughts?

matejvasek:
@mheon is `copier.Put()` working for you?

matejvasek:
I don't know if it changes anything, but I am running cgroups v1, not v2.

matejvasek:
I can see the same issue with `archiver` too, maybe something is wrong with my machine.

matejvasek:
Maybe I am doing mistake while testing, the file is upload correctly if I invoke API via `cURL`, but using `docker` with `DOCKER_HOST` seems to be buggy.

matejvasek:
Looks like very first call to API works (uploaded file is visible in running container almost immediately), but subsequent calls don't work well, there is error about unmounting in log:
```
WARN[0030] failed to unmount container a1d36bb594ae: "cannot unmount storage for container a1d36bb594ae2836ebdc063e443aa1ec40e91805a4241005f3c138635aa527fa as it is running or paused: container state improper" 
```
This explains why direct upload via `cURL` to freshly started container works, but `docker` CLI doesn't: `docker` CLI is doing two queries: first `HEAD` and then `PUT`.

matejvasek:
@nalind  This is really really weird. It seems that problem is call to `HEAD` -- `copier.Stat()`. When `copier.Stat()` is invoked subsequent `PUT`s -- `copier.Put()` don't work well. When I removed `copier.Stat()` from `HEAD` handler, upload works with no problem (I don't have to restart a container to see newly copied files).

matejvasek:
@nalind I am probably wrong about the `Stats()` thing. It seems to randomly not work.

matejvasek:
Sometimes copied files are immediately visible in running container, sometimes not and restart is required.

vrothberg (commented):
Nice work!

I will help getting this PR merged. Until now, I only read the code but I'll test it in a bit.

What we need is a bunch of tests in `./test/apiv2` to make sure we have something to prevent regressions.

Also, please add _verbose_ comments to the code and the functions.

vrothberg:
Continuing to drop some ideas.

Here are some ideas for testing:

* head: running curl with `--head` and making sure the "X-Docker-Container-Path-Stat" is set. Bonus points for base64 decoding and looking at the fields with `jq`.
* get with single file: untar the output, make sure the file is there and that it's identical to the one from the container. You could compare sha256 sums.
* get with entire directory: same as for file but compare the sha256 sums of the directories (sha of all sha sums of all files).
* put ... same as get but the other way around

pkg/api/handlers/compat/containers_archive.go
```diff
@@ -1,12 +1,377 @@
...
+		}
+	} else {
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
+func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, runtime *libpod.Runtime) {
+	query := struct {
+		Path string `schema:"path"`
+		// TODO handle params below
```
vrothberg, line 150: I am cool to merge this PR with the TODOs open. I can pick these work items up once merged.

vrothberg:
> @matejvasek When I was looking into implementing this, I was using https://github.com/containers/buildah/tree/master/copier - it's a new package designed for secure copy mimicking Docker behavior. The intention was to implement the archive endpoints using it, then swap `podman cp` over to use the same code.

I think we should swap the order. `podman cp` should first be ported to use `buildah/copier`. Then podman-remote can use the ABI implementation as we do for other endpoints.

If we first do the remote part and then tackle the local ones, we had to consolidate the code again.

pkg/api/handlers/compat/containers_archive.go
```diff
@@ -1,12 +1,377 @@
...
+
+	err = copier.Put(mountPoint, filepath.Join(mountPoint, path), opts, r.Body)
+	if err != nil {
+		logrus.Error(errors.Wrapf(err, "failed to copy to the %s container path %s", ctrName, query.Path))
+		return
+	}
+}
+
+func statsToHeader(stats *copier.StatForItem) (string, error) {
+	statsDTO := struct {
```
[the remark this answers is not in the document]
matejvasek: Interesting, I created the structure completely independently of that.

pkg/api/handlers/compat/containers_archive.go
```diff
@@ -1,12 +1,377 @@
...
+		return
+	}
+
+	statHeader, err := statsToHeader(stats[0].Results[stats[0].Globbed[0]])
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, err)
+		return
+	}
+
+	w.Header().Add("X-Docker-Container-Path-Stat", statHeader)
```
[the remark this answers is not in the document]
matejvasek: I saw in doc for api v1.3x example of `GET` returning the header. Also when I try to do `GET` on docker deamon (19.03.13) it does return the header.

matejvasek:
@nalind  I think I figured why I had issues with copying to running container. I was calling `ctr.Unmount(true);` meaning unmount was forced, when I changed it to non-forced it seems to work (`podman-cp` uses that too).

pkg/api/handlers/compat/containers_archive.go
matejvasek: Great. I don't know how I would fix that quickly. `buildah/copier` doesn't have the option for `NoOverwriteDirNonDir`. But `github.com/containers/storage/pkg/chrootarchive` `Untar()` has the option I could use that instead of `buildah/copier`.

pkg/api/handlers/compat/containers_archive.go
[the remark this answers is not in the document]
matejvasek: Since `docker` does adds the header also for `GET` and would prefer to keep it. At worst it shouldn't hurt.

matejvasek:
@vrothberg @mheon I tried to use `buildah/copier` in `podman-cp` but I decided not to go further with it. https://github.com/matejvasek/podman/commit/7bb4d8f80748695895e2cd0ca2d5cacf7d3dbfe0

test/apiv2/23-containersArchive.at
```diff
@@ -0,0 +1,28 @@
...
+
+HELLO_TAR="${TMPD}/hello.tar"
+
+podman run -d --name "${CTR}" "${IMAGE}" top
+
+t HEAD "containers/nonExistentCtr/archive?path=%2F" 404
+t HEAD "containers/${CTR}/archive?path=%2Fnon%2Fexistent%2Fpath" 404
+t HEAD "containers/${CTR}/archive?path=%2Fetc%2Fpasswd" 200
+
+curl "http://127.0.0.1:$PORT/containers/${CTR}/archive?path=%2Ftmp" -X PUT --upload-file "${HELLO_TAR}"
```
matejvasek, line 27: Seems there is no support for `PUT` yet. Also the testing suite expects that bodies are `JSON`s but in this case its tar stream. So I am just doing really simple test here.

Another thing is that it seems I can't use `podman-exec`. I tried:
```bash
if ! podman exec "${CTR}" cat "/etc/passwd" ; then
  echo FAIL;
fi
```
And it prints "FAIL" but it shouldn't the passwd should be there.

matejvasek: Also how can I easily access headers?

vrothberg:
> @vrothberg @mheon I tried to use `buildah/copier` in `podman-cp` but I decided not to go further with it, since `podman-cp` works as it is. [matejvasek@ba558f2](https://github.com/matejvasek/podman/commit/ba558f278625fb4779bb9a76541c1f381e3ffa85)

Thanks! Once your PR is merged, I will start plumbing with `podman-cp`. It's a really good base.

test/apiv2/23-containersArchive.at
vrothberg: Sorry for not checking before. I think that would require some bash wizardry. I suggest for now to limit tests to `HEAD` only. We can tackle the rest once the PR is merged.

matejvasek: I added some more tests, but I am kinda bypassing test utility. Have you ever considered using `Go` rather than `bash` for `api` testing?

test/apiv2/23-containersArchive.at
```diff
@@ -0,0 +1,73 @@
...
+nc='\e[0m'
+
+podman pull $IMAGE &>/dev/null
+
+# Ensure clean slate
+podman rm -a -f &>/dev/null
+
+CTR="ArchiveTestingCtr"
+
+TMPD=$(mktemp -d)
```
matejvasek, line 16: Maybe I could use `${WORKSPACE}` instead of this? I don't know.

pkg/api/handlers/compat/containers_archive.go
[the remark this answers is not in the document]
vrothberg: I agree. Thanks for clarifying.

test/apiv2/23-containersArchive.at
vrothberg: LGTM

vrothberg (commented):
LGTM
@baude @mheon PTAL

Thanks a lot for the great work, @matejvasek !

TomSweeneyRedHat:
LGTM
thanks @matejvasek !

baude:
@matejvasek thanks!

baude:
Do not merge until after 2.2

abitrolly:
I wonder if that could speed up https://github.com/wagoodman/dive when using `podman`?

PR #8127, opened 2020-10-24
Improve setupSystemd, grab mount options from the host

### LINKED ISSUE #7661: Grab mount flags for systemd work from the host mount
Per discussion in this PR: https://github.com/containers/podman/pull/7640, we want to grab the flags from the host mount as noted by @mheon [here](https://github.com/containers/podman/pull/7640#discussion_r488848768) and seconded by @edsantiago .

andylibrian:
For systemd with cgroupsv1, the mount options is currenlty hard coded:

```
[]string{"bind", "nodev", "noexec", "nosuid", "rprivate"} 
```

As discussed in PR #7640, we want to grab the mount options from the host.

fixes #7661

I'm not quite sure if there is any existing test that can hit cgroupsv1 only, to cover this scenario specifically.


libpod/container_internal_linux.go
```diff
@@ -38,6 +38,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
 	securejoin "github.com/cyphar/filepath-securejoin"
+	"github.com/moby/sys/mountinfo"
```
rhatdan, line 41: Could you use  	"github.com/containers/storage/pkg/mount"

rhatdan (approved):

libpod/container_internal_linux.go
andylibrian: Ah, sure. I wasn't aware of that. I've just pushed the update. Let me know if I should squash the commits.

libpod/container_internal_linux.go
```diff
@@ -668,11 +669,30 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
...
+		var hostMount *mount.Info
+		for _, m := range hostMounts {
+			if m != nil && m.Mountpoint == "/sys/fs/cgroup/systemd" {
+				hostMount = m
+				break
+			}
+		}
+
+		if hostMount == nil {
+			return errors.New("/sys/fs/cgroup/systemd is not mounted on the host")
```
mheon, line 686: Can you make this `errors.Wrapf(define.ErrInternal, "/sys/fs/cgroup/systemd is not mounted on the host")` - we generally try to have all errors from Libpod wrap one of our defined errors whenever we can.

mheon:
One nit otherwise LGTM

rhatdan:
Fix @mheon issue and LGTM

andylibrian:
I've just applied @mheon suggestion

libpod/container_internal_linux.go
```diff
@@ -668,11 +669,30 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
...
+			if m != nil && m.Mountpoint == "/sys/fs/cgroup/systemd" {
+				hostMount = m
+				break
+			}
+		}
+
+		if hostMount == nil {
+			return errors.Wrapf(define.ErrInternal, "/sys/fs/cgroup/systemd is not mounted on the host")
+		}
+
```
giuseppe, line 688: this is a bit expensive.

Could we just use `statfs("/sys/fs/cgroup/systemd")` and read `f_flags` from there?

rhatdan:
@andylibrian Could you make the change @giuseppe requested so we can get this merged.

andylibrian:
I've just pushed an update to use statfs to read the mount options. With `statfs` we need to build the options from flags like this:

```go

		if statfs.Flags&unix.MS_NODEV == unix.MS_NODEV {
			mountOptions = append(mountOptions, "nodev")
		}
		if statfs.Flags&unix.MS_NOEXEC == unix.MS_NOEXEC {
			mountOptions = append(mountOptions, "noexec")
		}
		if statfs.Flags&unix.MS_NOSUID == unix.MS_NOSUID {
			mountOptions = append(mountOptions, "nosuid")
		}
		if statfs.Flags&unix.MS_RDONLY == unix.MS_RDONLY {
			mountOptions = append(mountOptions, "ro")
		}
```

This works on `ubuntu 20.04.1 LTS (Focal Fossa)` but I'm not sure if there is any more flags needed for other cases.

@rhatdan @giuseppe

libpod/container_internal_linux.go
```diff
@@ -668,11 +668,30 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
...
+		if statfs.Flags&unix.MS_RDONLY == unix.MS_RDONLY {
+			mountOptions = append(mountOptions, "ro")
+		}
+
 		systemdMnt := spec.Mount{
 			Destination: "/sys/fs/cgroup/systemd",
 			Type:        "bind",
 			Source:      "/sys/fs/cgroup/systemd",
-			Options:     []string{"bind", "nodev", "noexec", "nosuid", "rprivate"},
+			Options:     mountOptions,
```
mheon, line 694: I think we should probably force `rprivate` even if not used by the host - @giuseppe @rhatdan Agree?

rhatdan: Yes.  We should always add rprivate

libpod/container_internal_linux.go
```diff
@@ -668,11 +668,30 @@ func (c *Container) setupSystemd(mounts []spec.Mount, g generate.Generator) erro
 		}
 		g.AddMount(systemdMnt)
 	} else {
+		var statfs unix.Statfs_t
+		if err := unix.Statfs("/sys/fs/cgroup/systemd", &statfs); err != nil {
+			return errors.Wrapf(err, "/sys/fs/cgroup/systemd is not mounted on the host")
```
giuseppe, line 673: if it is not mounted, could we just use a set of default mount options?

giuseppe (commented):
just a minor comment, could you please squash your patches into one?

libpod/container_internal_linux.go
mheon: Does that make sense given that we're trying to mount the same directory into the container? It seems like we ought to fail if we can't actually mount the cgroups in.

andylibrian:
I've just pushed the update as suggested, squashed into a single commit as requested. There is still a discussion about whether we should throw an error or fallback to default mount options in case systemd is not mounted.

mheon:
I think we're good to merge as-is; LGTM

mheon:

vrothberg (commented):
LGTM, thanks!

Issue #7959, opened 2020-10-07
Rootless --cpus results in Permission Denied


jam01:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
Executing a simple `podman run -it --rm -p 9000:9000 --memory 1G --cpus 1 localhost/bench-camel:latest` results in 
```
Error: opening file `cpu.max` for writing: Permission denied: OCI runtime permission denied error
```
Executing under sudo works correctly.

This issue seemed related https://github.com/containers/crun/issues/489 but if I remove the `--cpus` flag the memory limit is applied correctly, so it's something specific to cpus.

**Steps to reproduce the issue:**

1. Trying to limit cpus with `--cpus`

**Describe the results you received:**
Permissions denied error

**Describe the results you expected:**
Successfully limit cpus

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14.9
Built:        Wed Sep 30 14:31:11 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-2.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 81d18b6c3ffc266abdef7ca94c1450e669a6a388'
  cpus: 8
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: journald
  hostname: jam01.jam01
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.12-200.fc32.x86_64
  linkmode: dynamic
  memFree: 1624256512
  memTotal: 16672882688
  ociRuntime:
    name: crun
    package: crun-0.15-5.fc32.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.15
      commit: 56ca95e61639510c7dbd39ff512f80f626404969
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
[...]

PR #8134, opened 2020-10-25
[CI:DOCS] Document how to enable CPU limit delegation

xordspar0:
This adds documentation around the CPU limit delegation issue as discussed in #7959.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -107,6 +107,10 @@ Limit the CPU CFS (Completely Fair Scheduler) period
 
 Limit the container's CPU usage. This flag tell the kernel to restrict the container's CPU usage to the period you specify.
 
+On some systems, changing the CPU limits may not be allowed for non-root
+users. For more details, see
+https://github.com/containers/podman/blob/master/troubleshooting.md#26-running-containers-with-cpu-limits-fails-with-a-permissions-error
+
 **--cpu-quota**=*limit*
 
 Limit the CPU CFS (Completely Fair Scheduler) quota
@@ -114,6 +118,10 @@ Limit the CPU CFS (Completely Fair Scheduler) quota
 Limit the container's CPU usage. By default, containers run with the full
 CPU resource. This flag tell the kernel to restrict the container's CPU usage
 to the quota you specify.
+
+On some systems, changing the CPU limits may not be allowed for non-root
+users. For more details, see
+https://github.com/containers/podman/blob/master/troubleshooting.md#26-running-containers-with-cpu-limits-fails-with-a-permissions-error
 
 **--cpu-rt-period**=*microseconds*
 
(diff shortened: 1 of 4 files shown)


rhatdan:
Thanks @xordspar0 
LGTM
@TomSweeneyRedHat PTAL

docs/source/markdown/podman-run.1.md
```diff
@@ -1454,6 +1463,36 @@ The **fuse-overlay** package provides a userspace overlay storage driver, otherw
...
+
+If you want to enable CPU limit delegation for all users, you can create the
+file `/etc/systemd/system/user@.service.d/delegate.conf` with the contents:
+
+    [Service]
+    Delegate=memory pids cpu io
+
+After logging out and loggin back in, you should have permission to set CPU
+limits.
+
```
giuseppe, line 1495: I think it is fine to have it documented only in `troubleshooting.md`

docs/source/markdown/podman-create.1.md
```diff
@@ -1153,6 +1162,36 @@ The fuse-overlay package provides a userspace overlay storage driver, otherwise
...
+
+If you want to enable CPU limit delegation for all users, you can create the
+file `/etc/systemd/system/user@.service.d/delegate.conf` with the contents:
+
+    [Service]
+    Delegate=memory pids cpu io
+
+After logging out and loggin back in, you should have permission to set CPU
+limits.
+
```
giuseppe, line 1194: same here, I think it is fine to have it documented only in `troubleshooting.md`

troubleshooting.md
```diff
@@ -644,3 +644,39 @@ $ podman run --read-only --rootfs /path/to/rootfs ....
 
 Another option would be to create an overlay file system on the directory as a lower and then
 then allow podman to create the files on the upper.
+
+### 26) Running containers with CPU limits fails with a permissions error
+
+On some systems, non-root users do not have CPU limit delegation permissions.
```
giuseppe, line 650: Could we be more specific here? We could say the issue happens with systemd as by default it doesn't enable all the available controllers for unprivileged users

xordspar0: It's not *all* systems that have systemd. Fedora 33 does have CPU delegation enabled by default. It's really only systems that have cgroups v2 enabled but don't configure the delegations, which as far as I know is only Fedora 31 and 32. How specific should we be?

docs/source/markdown/podman-run.1.md
xordspar0: Ok. What should be in the man pages? A note saying that if you have issues with setting limits you should check troubleshooting.md?

troubleshooting.md
giuseppe: the proposed solution works only on systems managed by systemd, that is why I was pointing out we could specify the issue is specific to systemd.  Either way, it is just a nit, I am fine as it is now.

docs/source/markdown/podman-run.1.md
giuseppe: yes, I'd just add a pointer to the troubleshooting.md page to avoid maintaining the same information in 3 different places

troubleshooting.md
xordspar0: I see. That makes sense.

docs/source/markdown/podman-run.1.md
xordspar0: Ok, that sounds good. Is troubleshooting.md hosted on the website or should I just link to https://github.com/containers/podman/blob/master/troubleshooting.md?

PR #8142, opened 2020-10-25
Fix "Error: no such container"

kannkyo:
When executing podman logs as root, I get only "Error: no such container".

```bash
$ sudo podman logs --latest
Error: no such container
$ podman logs --latest
(snip)
[Sun Oct 25 16:38:59.076398 2020] [ssl:warn] [pid 1:tid 140704353533312] AH01882: Init: this version of mod_ssl was compiled against a newer library (OpenSSL 1.1.1b FIPS  26 Feb 2019, version currently loaded is OpenSSL 1.1.1 FIPS  11 Sep 2018) - may result in undefined or erroneous behavior
(snip)
```

### DIFF
--- a/docs/tutorials/podman_tutorial.md
+++ b/docs/tutorials/podman_tutorial.md
@@ -60,7 +60,7 @@ curl http://<IP_address>:8080
 ### Viewing the container's logs
 You can view the container's logs with Podman as well:
 ```console
-$ sudo podman logs --latest
+$ podman logs --latest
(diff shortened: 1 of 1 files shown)


mheon:
We need to either remove all use of sudo (and mark the checkport/restore section of the tutorial as "root only") or re-add it and direct folks to the dedicated rootless tutorial; the current state of the turotial is basically unusable.

rhatdan:
@kannkyo Please remove all instances of sudo from the tutorial.

Removing just one only moves the problem down the line.

If we remove all, then the commands will run fine as root or non root.

rhatdan:
Replaced by https://github.com/containers/podman/pull/8356

PR #8143, opened 2020-10-25
    enable ipv6 networks

aojea:
enable the ipv6 flag in podman network to be able to create dual-stack network for containers.
 
This is required to be compatible with docker, where --ipv6 really means dual stack.
    
podman, unlike docker, support IPv6 only containers since  07e3f1bba9674c0cb93a0fa260930bfebbf75728.
    
## How to use it

- Create a network

```sh
$ podman network create --ipv6 --subnet fd00:ff:123::/112 test
/home/aojea/.config/cni/net.d/test.conflist
```

- Run a container in the new created network and check the ips and routes assigned
```sh
$ podman/bin/podman run -it --network test alpine ash
/ # ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
3: eth0@if4: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP 
    link/ether e6:3d:d7:4f:77:6c brd ff:ff:ff:ff:ff:ff
    inet 10.88.2.2/24 brd 10.88.2.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fd00:ff:123::2/112 scope global 
       valid_lft forever preferred_lft forever
    inet6 fe80::e43d:d7ff:fe4f:776c/64 scope link 
       valid_lft forever preferred_lft forever
/ # ip route
default via 10.88.2.1 dev eth0 
10.88.2.0/24 dev eth0 scope link  src 10.88.2.2 
/ # ip -6 route
fd00:ff:123::/112 dev eth0  metric 256 
fe80::/64 dev eth0  metric 256 
default via fd00:ff:123::1 dev eth0  metric 1024 
ff00::/8 dev eth0  metric 256 
```

Signed-off-by: Antonio Ojea <aojea@redhat.com>

Fixes: https://github.com/containers/podman/issues/7302


rhatdan:
Thanks @aojea 
LGTM if tests pass.
@mheon @baude PTAL

AkihiroSuda:
rootless-cni-infra needs to be started with IPv6-enabled slirp4netns?
https://github.com/containers/podman/blob/master/libpod/rootless_cni_linux.go#L285

aojea:
> rootless-cni-infra needs to be started with IPv6-enabled slirp4netns?
> https://github.com/containers/podman/blob/master/libpod/rootless_cni_linux.go#L285

I was trying to find the problem, thanks.
I will take back the "skip" then, and it can be done in another PR

aojea:
> rootless-cni-infra needs to be started with IPv6-enabled slirp4netns?

why the previous test, for IPv4 only, works then?

aojea:
@AkihiroSuda it fails because of this
```
[/sbin/ip6tables -t nat -S --wait]: exit status 3: modprobe: can't change directory to '/lib/modules': No such file or directory\nip6tables v1.8.4 (legacy): can't initialize ip6tables table `nat': Table does not exist (do you need to insmod?)\nPerhaps ip6tables or your kernel needs to be upgraded.\n\n"
```

do you know how can I enable it?

libpod/network/create.go
```diff
@@ -41,60 +42,120 @@ func Create(name string, options entities.NetworkCreateOptions, r *libpod.Runtim
...
+			return errors.Wrapf(err, "failed to get last IP address from ip-range")
+		}
+		if !subnet.Contains(firstIP) || !subnet.Contains(lastIP) {
+			return errors.Errorf("the ip range %s does not fall within the subnet range %s", ipRange.String(), subnet.String())
+		}
+	}
+
+	// if network is provided and if gateway is provided, make sure it is "in" network
+	if gateway != nil && !subnet.Contains(gateway) {
+		return errors.Errorf("gateway %s is not in valid for subnet %s", gateway.String(), subnet.String())
```
giuseppe, line 76: should the error be "gateway %s is not valid for subnet %s" ?

aojea: indeed

aojea: heh, is carrying it over, it was like this before
https://github.com/containers/podman/pull/8143/files#diff-1203e6d536d51fdb4f55e4a605d74cc02986bb5b934027f576dbf8faf91ff6b9L73

aojea: fixed

aojea:
any idea why this is failing now?
`PR title check / update_pr (pull_request_target`

dcbw:
@mccv1r0 can you take a look at this too?

QiWang19:
> any idea why this is failing now?
> `PR title check / update_pr (pull_request_target`

@aojea There's a PR fixing this. Will pass after it gets merged.

docs/source/markdown/podman-network-create.1.md
```diff
@@ -63,6 +67,19 @@ Create a network named *newnet* that uses *192.5.0.0/16* for its subnet.
 /etc/cni/net.d/newnet.conflist
 ```
 
+Create a dual-stack network named *newDualnet* that uses an IPv6 range *2001:db8::/64* for its subnet.
+The IPv4 range is autogenerated.
```
mccv1r0, line 71: Why use `2001:db8::/64?   Can't the IPv6 range can be autogenerated too?  Preferably from a ULA and not a global range that someday might belong to someone?

aojea: this is the current docker behaviour, you have to specify the IPv6 range to create the network, the `2001:db8::/64`is just for documentation purposes, the user can specify any range meanwhile there is no conflict

mccv1r0: For docs it makes sense to use [IPv6 Address Prefix Reserved for Documentation](https://tools.ietf.org/html/rfc3849)

When using create, maybe generate IPv6 prefix (from ULA) when the prefix is not supplied by user?  I know docker doesn't do this but docker also doesn't have rootless either.  We can improve on docker.     Otherwise unless a prefix is provided to the users, they'll just take `2001:db8::/64` because it was in the doc.

Today, for IPv4, every podman system uses [10.88.0.0/16](https://github.com/containers/podman/blob/master/cni/87-podman-bridge.conflist).  I'm suggesting that this conflict be avoided with IPv6.

mheon: We should probably add a default-prefix field in containers.conf for storing a v6 prefix to use (and, for that matter, an IPv4 prefix - I believe we are presently hardcoding our auto-assigned v4 prefixes to be the first open /24 in 10.89.0.0/16)

mccv1r0: When `containers.conf` is created for the first time, I suggest to use a [Pseudo-Random Global ID](https://tools.ietf.org/html/rfc4193#section-3.2.2) to pick the prefix.  The Subnet ID for each network can pick the first open value from there.

If user re-install, they get new ULA, which is what happens with CPE gear that follow [RFC7084](https://tools.ietf.org/html/rfc7084) get factory reset.

aojea: so, how https://podman.io/blogs/2020/07/01/rest-versioning.html applies here?

> Podman’s REST API is split into two halves: one providing a Docker-compatible API, and a Libpod API providing support for Podman’s unique features such as pods.

`podman network create` has to be docker compatible or no? because if it does and we modify the behaviour to assign an IPv6 when docker fails, we are breaking the compatibility (I'm not entering the discussion now on which one is better, just which is correct )

mheon: The CLI is not strictly required to be Docker compatible, but we try to match them wherever possible for consistency and ease of migration; when we break compatibility, we should do so with good reason.

The REST API's compat endpoints are required to be strictly Docker compatible.

aojea: I personally think that changing the defaults or the behaviour respect docker should not be on the scope of this PR, since that will require more discussion, i.e, Michael suggestion is really interesting https://github.com/containers/podman/pull/8143#discussion_r512126735 but that differs from IPv4 behaviour, where you always have a well-known-subnet and most CIs and user leverage (is kind of an API right now)

docs/source/markdown/podman-network-create.1.md
```diff
@@ -49,6 +49,10 @@ Macvlan connection.
 
 The subnet in CIDR notation.
 
+**--ipv6**
+
+Enable IPv6 (Dual Stack) networking
```
TomSweeneyRedHat, line 54:
```suggestion
Enable IPv6 (Dual Stack) networking.
```

docs/source/markdown/podman-network-create.1.md
rhatdan: yes lets talk about changes of defaults in a different PR.  I am fine with adding something to containers.conf to implement the change.  If @mccv1r0 defaults is a better idea, we could switch to that default and then allow users to switch containers.conf into Docker mode.

rhatdan:
@giuseppe @mheon @vrothberg PTAL
Would be a nice feature to get this into podman 2.2
Nice work @aojea

aojea:
kindly ping, any feedback on this?

rhatdan:
@mheon PTAL

docs/source/markdown/podman-network-create.1.md
mheon: Wait, do we use the same /64 for *every* new v6 network by default? I think we do need some logic to increment that if we already have a network with that subnet, otherwise we wait until CNI tries to start a container in the network to blow up with an error...

mheon:
One last comment on IPv6 default subnet selection - I don't think we have to figure out autogeneration, but I do think we need to ensure that we don't make networks with duplicate IPv6 subnets. If I call `podman network create --ipv6` twice and do not specify a subnet, either the second call should error that the default subnet is in use and you have to select a new one, or it should automatically select another subnet not in use.

mheon:
I would also love a test that `podman network create --ipv6 --subnet` with the same subnet twice will error.

docs/source/markdown/podman-network-create.1.md
aojea: Current pr implement the docker behaviour, there is no defaulting for ipv subnets https://github.com/containers/podman/pull/8143#discussion_r512068637

I'm happy to keep contributing and follow up this discussion in a new pr/issue/meeting
https://github.com/containers/podman/pull/8143#discussion_r512142451

aojea:
> One last comment on IPv6 default subnet selection - I don't think we have to figure out autogeneration, but I do think we need to ensure that we don't make networks with duplicate IPv6 subnets. If I call `podman network create --ipv6` twice and do not specify a subnet, either the second call should error that the default subnet is in use and you have to select a new one, or it should automatically select another subnet not in use.

`podman network create --ipv6` without a subnet is an error, same as docker does

mheon:
That's not what I got out of the manpages? They mention a specific subnet. I'm OK with merging this assuming the mention of the subnet in the manpages is removed, and a test for two networks with the same subnet is added.

aojea:
> That's not what I got out of the manpages? They mention a specific subnet. I'm OK with merging this assuming the mention of the subnet in the manpages is removed, and a test for two networks with the same subnet is added.

clarified the manpage and added testcases for creation of network with duplicate subnets

mheon:
LGTM

Needs one more to merge - @containers/podman-maintainers PTAL

TomSweeneyRedHat:
Tests don't look happy....

mccv1r0 (approved):

aojea:
> rdoproject.org/github-check — github-check status: failure

at least the rh job fails with the following error

>  Problem: problem with installed package podman-catatonit-2.0.5-5.module+el8.3.0+8221+97165c3f.x86_64
  - package podman-catatonit-2.0.5-5.module+el8.3.0+8221+97165c3f.x86_64 requires podman = 2.0.5-5.module+el8.3.0+8221+97165c3f, but none of the providers can be installed
  - cannot install both podman-2.2.0-1605002806.gitaabf28a1.el8.x86_64 and podman-2.0.5-5.module+el8.3.0+8221+97165c3f.x86_64
  - conflicting requests

daiaji:
When I use `podman network create --ipv6 --subnet 1::/86` test to create a network, I always get this error...

> Error: network 1::/86 is being used by a network interface

Should I manually set the ipv6 address for the host instead of obtaining it automatically?

And the ISP of my home network only provides me with random IPv6 prefixes and IPv6-PD. Does podman support this?

aojea:
> --subnet 1::/86

that subnet is not very common and in a reserved range, https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml

I suggest to go with an ULA subnet as in the example and the docs I´ve added
```
$ podman network create --ipv6 --subnet fd00:ff:123::/112 test
/home/aojea/.config/cni/net.d/test.conflist
```

> Should I manually set the ipv6 address for the host instead of obtaining it automatically?

the ipv6 of the host depends on your network, what you define with the ` podman network create` is the subnet that podman will create to run its containers


> And the ISP of my home network only provides me with random IPv6 prefixes and IPv6-PD. Does podman support this?

podman defines a network in your host, is on the user to decide how and which technology to use to cascade the ISP prefixes to the host where podman is running

daiaji:
> podman defines a network in your host, is on the user to decide how and which technology to use to cascade the ISP prefixes to the host where podman is running

Is there any documentation?  I want to see.

In fact, if podman cannot automatically obtain the ipv6 prefix, it may be impossible to use ipv6 in my network environment.

daiaji:
> --subnet 1::/86

Don't mind, that is just an example, just to cover up the global unicast address I actually use.



> podman defines a network in your host, is on the user to decide how and which technology to use to cascade the ISP prefixes to the host where 

podman is running
Is there any documentation?  I want to see.

But in fact, if podman cannot automatically obtain the ipv6 prefix, maybe in my network environment, ipv6 cannot be used at all, after all, the prefixes distributed by ISP are always changing.

rhatdan:
@mccv1r0 WDYT?

mccv1r0:
> Should I manually set the ipv6 address for the host instead of obtaining it automatically?
>
> And the ISP of my home network only provides me with random IPv6 prefixes and IPv6-PD. Does podman support this?

Yes set it manually.   There isn't support for obtaining it automatically.

What your ISP is doing is typical.  How one obtains an IPv6 address from ISP, IT department, Cloud etc. is out of scope for podman.   DHCPv6 Prefix Delegation helps, but not everyone uses DHCP.  DHCPv6 is also performed on interface that podman has no idea about, and using a DHCPv6 client that is very dependent on what was installed on the specific version of Linux itself.

It could be possible someday to pick one DHCPv6 client (Network Manager?) and automate things such that a delegated prefix gets used by an invocation of `podman network create` but nothing in the works yet.

daiaji:
> > Should I manually set the ipv6 address for the host instead of obtaining it automatically?
> > And the ISP of my home network only provides me with random IPv6 prefixes and IPv6-PD. Does podman support this?
> 
> Yes set it manually. There isn't support for obtaining it automatically.
> 
> What your ISP is doing is typical. How one obtains an IPv6 address from ISP, IT department, Cloud etc. is out of scope for podman. DHCPv6 Prefix Delegation helps, but not everyone uses DHCP. DHCPv6 is also performed on interface that podman has no idea about, and using a DHCPv6 client that is very dependent on what was installed on the specific version of Linux itself.
> 
> It could be possible someday to pick one DHCPv6 client (Network Manager?) and automate things such that a delegated prefix gets used by an invocation of `podman network create` but nothing in the works yet.

daiaji:
Excuse me, until now, there is still no progress in setting the ipv6 address automatically?

mheon:
If you don't require static v6 IPs, dual-stack v4 and v6 and single-stack v6 are both working fine right now.

PR #8146, opened 2020-10-26
new "image" mount type

vrothberg:
Add a new "image" mount type to `--mount`.  The source of the mount is 
the name or ID of an image.  The destination is the path inside the    
container.  Image mounts further support an optional `rw,readwrite`    
parameter which if set to "true" will yield the mount writable inside  
the container.  Note that no changes are propagated to the image mount 
on the host (which in any case is read only).                          
                                                                       
Mounts are overlay mounts.  To support read-only overlay mounts, vendor
a non-release version of Buildah.                                      


mheon:
I think you're missing most of the conflict resolution logic for volumes
with the same destination - check container_validate.go in Libpod.

vrothberg:
> I think you're missing most of the conflict resolution logic for volumes with the same destination - check container_validate.go in Libpod.

Thanks for checking! I stopped at this point as I want to check with @rhatdan if the mounts should be RO or if we make use of overlay mounts. The latter may allow for more use cases.

vrothberg:
@mheon @rhatdan @mrunalp please have a look.

I'll update the man pages once we're good to go. Image volumes are now always overlay mounts so that they're writeable inside the container. There's currently a rootless overlay-mount bug during cleanup, so you may ignore errors on container removal.

vrothberg:
@edsantiago feel free to check the system tests if there's a potential blocker from your side

libpod/container_internal_linux.go
```diff
@@ -368,6 +368,29 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
 	}
 
+	// Add image volumes as overlay mounts
+	for _, volume := range c.config.ImageVolumes {
+		// Mount the specified image.
+		img, err := c.runtime.ImageRuntime().NewFromLocal(volume.Source)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error creating image volume %q:%q", volume.Source, volume.Dest)
+		}
+		mountPoint, err := img.Mount(nil, "")
```
rhatdan, line 378: Not sure if SELinux will block the image without a label. But need to verify this will work. On the other hand this had better work, since the mounts will be shared between different containers.

libpod/container_internal_linux.go
```diff
@@ -368,6 +368,29 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "error creating image volume %q:%q", volume.Source, volume.Dest)
```
rhatdan, line 380: sed s/creating/mounting

libpod/container_internal_linux.go
```diff
@@ -368,6 +368,29 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
+		}
+
+		contentDir, err := overlay.TempDir(c.config.StaticDir, c.RootUID(), c.RootGID())
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create TempDir in the %s directory", c.config.StaticDir)
+		}
+		overlayMount, err := overlay.Mount(contentDir, mountPoint, volume.Dest, c.RootUID(), c.RootGID(), c.runtime.store.GraphOptions())
```
rhatdan, line 387: This should only be done if the user requested an overlay mount.

libpod/container_internal_linux.go
rhatdan: Need to handle read/only mounts

rhatdan (approved):

cmd/podman/common/volumes.go
```diff
@@ -99,28 +99,37 @@ func parseVolumes(volumeFlag, mountFlag, tmpfsFlag []string, addReadOnlyTmpfs bo
...
 		if _, ok := unifiedMounts[dest]; ok {
-			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+			return nil, nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
 		}
 		if _, ok := unifiedVolumes[dest]; ok {
-			return nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+			return nil, nil, nil, nil, errors.Wrapf(errDuplicateDest, "conflict at mount destination %v", dest)
+		}
+	}
+	for dest := range unifiedImageVolumes {
```
mheon, line 127: We can't just do these two - we need to check against overlay volumes as well.

mheon:
Please ensure these new mounts are present in the output of `podman inspect`

libpod/runtime_ctr.go
```diff
@@ -579,6 +579,17 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force bool,
 
 	c.newContainerEvent(events.Remove)
 
+	for _, v := range c.config.ImageVolumes {
```
mheon, line 582: This should be part of container cleanup, not container removal. We want to unmount these when the container stops.

libpod/runtime_ctr.go
```diff
@@ -579,6 +579,17 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force bool,
...
+		// Unmount the specified image.
+		img, err := c.runtime.ImageRuntime().NewFromLocal(v.Source)
+		if err != nil {
+			return errors.Wrapf(err, "error unmounting image volume %q:%q", v.Source, v.Dest)
+		}
+		if err := img.Unmount(false); err != nil {
```
mheon, line 588: Does this properly handle the case where the image is already unmounted? We need to be certain that case does not cause us to be unable to remove containers.

vrothberg: Thanks for checking. Yes, that case is covered. If an image is already umounted, we don't get an error.

libpod/container_internal_linux.go
vrothberg: I am tempted to always use overlay mounts.

cmd/podman/common/volumes.go
vrothberg: Good catch! I'll simplify the checks a bit and get it polished.

libpod/container_internal_linux.go
rhatdan: The problem with overlay mounts, is it could leave the user with the impression that their changes are permanent.  If they get image is read/only then they would know right away.

edsantiago:
@vrothberg have you looked into the failing tests? The common factor is root system tests, on all OS, local and remote, and the failing test is `touch /image-mount/read-only` (it is succeeding, but the test expects it to fail)

vrothberg:
> @vrothberg have you looked into the failing tests? The common factor is root system tests, on all OS, local and remote, and the failing test is `touch /image-mount/read-only` (it is succeeding, but the test expects it to fail)

Thanks for checking! I'll wrap things up in a moment. I wanted to use overlay mounts but those are RW by default, so it had to update the overlay package in Buildah to support RO mounts.

Now, the image mounts are RO (again) and RW by choice.

vrothberg:
Note: this PR vendors a non-release Buildah. I think that's okay as we need to get short-names support in before cutting Podman v2.2.

libpod/container.go
```diff
@@ -235,6 +235,19 @@ type ContainerOverlayVolume struct {
 	Source string `json:"source,omitempty"`
 }
 
+// ContainerImageVolume is a volume based on a container image.  The container
+// image is first mounted on the host and is then bind-mounted into the
+// container.  An ImageVolume is always mounted read only.
```
mheon, line 240: That "always mounted read-only" bit seems incorrect

libpod/container_inspect.go
```diff
@@ -192,14 +192,9 @@ func (c *Container) getContainerInspectData(size bool, driverData *driver.Data)
 // Get inspect-formatted mounts list.
 // Only includes user-specified mounts. Only includes bind mounts and named
 // volumes, not tmpfs volumes.
-func (c *Container) getInspectMounts(namedVolumes []*ContainerNamedVolume, mounts []spec.Mount) ([]define.InspectMount, error) {
+func (c *Container) getInspectMounts(namedVolumes []*ContainerNamedVolume, imageVolumes []*ContainerImageVolume, mounts []spec.Mount) ([]define.InspectMount, error) {
 	inspectMounts := []define.InspectMount{}
 
-	// No mounts, return early
-	if len(c.config.UserVolumes) == 0 {
```
mheon, line 199: We need to retain this - if it's no longer accurate because image volumes/overlay volumes aren't being added to it, they should be added to UserVolumes.

vrothberg:
Thanks for the reviews, @mheon. Wrapped it up and remote sys tests should be green in a bit.

TomSweeneyRedHat:
LGTM

PR #8147, opened 2020-10-26
pr update action: fix errors on master branch

vrothberg:
The action fails on the master branch as the regex does not match.
The error in this scenario is unfortunate and not of much value as
we do not want to change PR titles on the master branch.

To fix it, entirely disable the action on the master branch which
in restrospective may be a better approach as we do not fire off the
action.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -3,6 +3,8 @@
 name: "PR title check"
 
 on: pull_request_target
+  branches:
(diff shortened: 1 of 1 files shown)


vrothberg:
@rhatdan @mheon PTAL. Apologies for the noise! I dearly hope that's fixing it now - we'll find out _after_ it's being merged.

QiWang19 (commented):
LGTM

mheon:
LGTM

PR #8151, opened 2020-10-26
image list: check for all errors

vrothberg:
For unknown historical reasons, some errors were ignored when listing
images.  I assume that the basic assumption was that if we can properly
list images, we can also successfully compute their sizes which turned
out to be wrong.

### DIFF
--- a/pkg/domain/infra/abi/images_list.go
+++ b/pkg/domain/infra/abi/images_list.go
@@ -5,6 +5,7 @@ import (
 
 	libpodImage "github.com/containers/podman/v2/libpod/image"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/pkg/errors"
 )
 
 func (ir *ImageEngine) List(ctx context.Context, opts entities.ImageListOptions) ([]*entities.ImageSummary, error) {
@@ -43,12 +44,21 @@ func (ir *ImageEngine) List(ctx context.Context, opts entities.ImageListOptions)
 			VirtualSize:  img.VirtualSize,
 			RepoTags:     img.Names(), // may include tags and digests
 		}
-		e.Labels, _ = img.Labels(context.TODO())
+		e.Labels, err = img.Labels(ctx)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error retrieving label for image %q: you may need to remove the image to resolve the error", img.ID())
+		}
 
-		ctnrs, _ := img.Containers()
+		ctnrs, err := img.Containers()
+		if err != nil {
+			return nil, errors.Wrapf(err, "error retrieving containers for image %q: you may need to remove the image to resolve the error", img.ID())
+		}
 		e.Containers = len(ctnrs)
 
-		sz, _ := img.Size(context.TODO())
+		sz, err := img.Size(ctx)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error retrieving size of image %q: you may need to remove the image to resolve the error", img.ID())
+		}
 		e.Size = int64(*sz)
 
 		summaries = append(summaries, &e)


edsantiago:
Close:
```console
$ ./bin/podman images
Error: top layer 1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73 of image a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a not found in layer tree
$ ./bin/podman images -a
ERRO[0000] Error getting image size: unable to determine size: error locating layer with ID "1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73": layer not known
REPOSITORY                TAG       IMAGE ID      CREATED       SIZE
quay.io/libpod/testimage  20200929  766ff5a3a7e4  3 weeks ago   7.75 MB
quay.io/libpod/fedora     31        a7a37f74ff86  5 months ago  0 B
```

vrothberg:
@edsantiago, I update the errors. Does that meet your expectations?

edsantiago:
With 9029493 I get:
```console
$ ./bin/podman images
Error: top layer 1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73 of image a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a not found in layer tree
$ ./bin/podman images -a
Error: error retrieving size of image "a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a": you may need to remove the image to resolve the error: unable to determine size: error locating layer with ID "1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73": layer not known
$ ./bin/podman image inspect a7a
Error: stat /home/esm/.local/share/containers/storage/overlay/1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73: no such file or directory

**Comment by vrothberg** [comment]
I think we should split that in two work items. First, we need to get the nil deref fixed. Dealing with a corrupted image seems to open a whole universe of ugly messages.

The latter will be tricky.

**Comment by edsantiago** [comment]
Okay, in order of decreasing priority (IMHO):
1. `podman images -a` should not crash
2. `podman images` should at a very minimum show existing valid images
3. `podman images` should deal with rotten images. Maybe show them as errored, maybe ignore them, maybe ignore them with a warning, I don't know.

But right now I have a system that is not in good shape, and I don't want any end users winding up in the same situation.

**Comment by vrothberg** [comment]
Sounds reasonable to me. This PR gets us to 1).

2) and 3) will need some tinkering in the backend. I like the idea to show only "healthy" images and list the others separately.

So far, I have seen such corrupted images only when builds, or image commits in general (also happens during pull), have been killed or been interrupted (OEM kills, power outage, etc.). In that case, removing the image is the only option I can think of. But I'd leave that decision open to users.

@mtrmac @rhatdan PTAL

**Comment by edsantiago** [comment]
Is there some easy way to get `podman images` to emit something at least partly useful? Right now (9029493) I, a simple naive user, am getting the `top layer` error with no indication of what is wrong; and it is not obvious to me that I should run `podman images -a` to get a different slightly less unhelpful error message.

**Review by mtrmac — COMMENTED** [review]
The error checks LGTM.

I haven’t checked in detail whether removing the image is likely to resolve the error – or if it is even always possible; it does seem reasonable just based on the quoted error messages.

**Comment by edsantiago** [comment]
It is not (possible to remove the image):
```console
$ podman rmi a7a3
Error: 1 error occurred:
        * top layer 1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73 of image a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a not found in layer tree
```
Even `-a -f` fails:
```console
$ podman rmi -a -f
Error: 3 errors occurred:
        * top layer 1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73 of image a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a not found in layer tree
        * top layer 1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73 of image a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a not found in layer tree
        * unable to delete all images, check errors and re-run image removal if needed
```
I could `rm -rf ~/.local/share/containers` but I don't think we should force end users to do (or even know about) this.

edsantiago:
I'm tentatively ok with this PR, but please remove the 'Fixes' comment from both the commit message and the Github description.

vrothberg:
> I'm tentatively ok with this PR, but please remove the 'Fixes' comment from both the commit message and the Github description.

done

mheon:
LGTM
/hold cancel

PR #8156, opened 2020-10-26
Add network aliases for containers to DB

mheon:
This adds the database backend for network aliases. Aliases are additional names for a container that are used with the CNI dnsname plugin - the container will be accessible by these names in addition to its name. Aliases are allowed to change over time as the container connects to and disconnects from networks.

Aliases are implemented as another bucket in the database to register all aliases, plus two buckets for each container (one to hold connected CNI networks, a second to hold its aliases). The aliases are only unique per-network, to the global and per-container aliases buckets have a sub-bucket for each CNI network that has aliases, and the aliases are stored within that sub-bucket. Aliases are formatted as alias (key) to container ID (value) in both cases.

Three DB functions are defined for aliases: retrieving current aliases for a given network, setting aliases for a given network, and removing all aliases for a given network.

This is still only partially done. Remaining work:
- Implement the remove aliases function for BoltDB
- Implement aliases in the in-memory state
- Add unit testing for the new aliases functionality in the DB
- Add accessors to Container to get aliases
- Add an accessor to get the container's IP for a specific CNI network

### DIFF
--- a/libpod/boltdb_state.go
+++ b/libpod/boltdb_state.go
@@ -1021,7 +1021,7 @@ func (s *BoltState) GetNetworkAliases(ctr *Container, network string) ([]string,
 
 		inNetwork := ctrNetworkBkt.Get([]byte(network))
 		if inNetwork == nil {
-			return errors.Wrapf(define.ErrNoAliases, "container %s is not part of network %s, no aliases found")
+			return errors.Wrapf(define.ErrNoAliases, "container %s is not part of network %s, no aliases found", ctr.ID(), network)
 		}
 
 		ctrAliasesBkt := dbCtr.Bucket(aliasesBkt)
@@ -1103,12 +1103,14 @@ func (s *BoltState) GetAllNetworkAliases(ctr *Container) (map[string][]string, e
 			}
 
 			netAliases := []string{}
-			aliases[string(network)] = netAliases
 
-			return netAliasesBkt.ForEach(func(alias, v []byte) error {
+			_ = netAliasesBkt.ForEach(func(alias, v []byte) error {
 				netAliases = append(netAliases, string(alias))
 				return nil
 			})
+
+			aliases[string(network)] = netAliases
+			return nil
 		})
 	})
 	if err != nil {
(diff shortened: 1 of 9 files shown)


libpod/boltdb_state_internal.go
```diff
@@ -46,26 +48,28 @@ const (
...
-	podIDKey           = []byte(podIDName)
-	namespaceKey       = []byte(namespaceName)
+	networksBkt        = []byte(networksName)
+
+	configKey     = []byte(configName)
+	stateKey      = []byte(stateName)
+	netNSKey      = []byte(netNSName)
+	containersBkt = []byte(containersName)
+	podIDKey      = []byte(podIDName)
+	namespaceKey  = []byte(namespaceName)
```
TomSweeneyRedHat, line 72: any chance these could be alphabetized?

libpod/boltdb_state.go
```diff
@@ -969,6 +972,184 @@ func (s *BoltState) AllContainers() ([]*Container, error) {
...
+			return nil
+		})
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return aliases, nil
+}
+
```
TomSweeneyRedHat, line 1037: Maybe I missed it, but will there be a `podman list network aliases {network}` or similar command/option?

mheon: No. They'll show up in `podman inspect` on the container they're attached to.

mheon:
Stripping WIP. Still missing some bits, but we can add them as part of the future Network Connect / Network Disconnect work.

rhatdan:
@baude PTAL

baude:
@rhatdan PTAL

libpod/boltdb_state_internal.go
mheon: Right now they're in the order they're created in, DB-wise.

baude:
LGTM and i verified this with the aliases code and it all still works.

ashley-cui:
LGTM (from an initial read-through)

mheon:
This is green
@baude

Issue #6510, opened 2020-06-07
podman rmi gives "layer not known" when executed in parallel


doc-olliday:
<!--
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------
Use the commands below to provide key information from your environment:
You do NOT have to include this information if this is a FEATURE REQUEST

**NOTE** A large number of issues reported against Podman are often found to already be fixed
in more current versions of the project.  Before reporting an issue, please verify the
version you are running with `podman version` and compare it to the latest release
documented on the top of Podman's [README.md](../README.md).  If they differ, please
update your version of Podman to the latest possible and retry your command before creating
an issue.

If you are filing a bug against `podman build`, please instead file a bug
against Buildah (https://github.com/containers/buildah/issues). Podman build
executes Buildah to perform container builds, and as such the Buildah
maintainers are best equipped to handle these bugs.
-->

**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

We use a Jenkins pipeline with parallel stages. Each stage builds, tags, pushs and removes a container image. In **about half of the pipeline runs** the `rmi` command fails with
`layer not known`.
We suspect a race condition during the parallel runs.
Running the same pipeline on "older" centos7 machines with docker-ce installed works without any problem.

In the pipeline script the command `docker` is used to be compatible with the "old" workers. On the actual workers `docker` is a symlink to the `podman` executable.

Many thanks for your help!

**Steps to reproduce the issue:**

1. Execute the following Jenkins pipeline on a physical worker node:

The original pipeline is more complex with 7 images build in parallel. It was stripped down for debugging and the problem was reproduced with this version:

```
pipeline{
    agent any
    options{
        ansiColor('xterm')
        disableConcurrentBuilds()
        gitLabConnection('My GitLab')
        skipDefaultCheckout()
        timestamps()
    }
    environment{
        DOCKER_TAG =        "201907_bn${BUILD_NUMBER}"
        DOCKER_NAMESPACE =  "podmantest"
    }
    stages{
        stage("Build-Phase"){
            parallel{
                stage("Build Knoedel 1"){
                    steps{
                        script{
                            buildUnit("knoedel1", "${DOCKER_NAMESPACE}", "${DOCKER_TAG}").call()
                        }
                    }
                }
                stage("Build Knoedel 2"){
                    steps{
                        script{
[...]

PR #8165, opened 2020-10-27
Move from docker.io

edsantiago:
Followon to #7965 (mirror registry). mirror.gcr.io doesn't
cache all the images we need, and I can't find a way to
add to its cache, so let's just use quay.io for those
images that it can't serve.

Tools used:
  skopeo copy --all docker://docker.io/library/alpine:3.10.2 \
                    docker://quay.io/libpod/alpine:3.10.2

...and also:

    docker.io/library/alpine:3.2
    docker.io/library/busybox:latest
    docker.io/library/busybox:glibc
    docker.io/library/busybox:1.30.1
    docker.io/library/redis:alpine
    docker.io/libpod/alpine-with-bogus-seccomp:label
    docker.io/libpod/alpine-with-seccomp:label
    docker.io/libpod/alpine_healthcheck:latest
    docker.io/libpod/badhealthcheck:latest

Since most of those were new quay.io/libpod images, they required
going in through the quay.io GUI, image, settings, Make Public.

### DIFF
--- a/test/e2e/exists_test.go
+++ b/test/e2e/exists_test.go
@@ -39,6 +39,7 @@ var _ = Describe("Podman image|container exists", func() {
 		Expect(session).Should(Exit(0))
 	})
 	It("podman image exists in local storage by short name", func() {
+		Skip("FIXME-8165: shortnames don't seem to work with quay (#8176)")
 		session := podmanTest.Podman([]string{"image", "exists", "alpine"})
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
--- a/test/e2e/images_test.go
+++ b/test/e2e/images_test.go
@@ -41,8 +41,8 @@ var _ = Describe("Podman images", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
 		Expect(len(session.OutputToStringArray())).To(BeNumerically(">", 2))
-		Expect(session.LineInOuputStartsWith("docker.io/library/alpine")).To(BeTrue())
-		Expect(session.LineInOuputStartsWith("docker.io/library/busybox")).To(BeTrue())
+		Expect(session.LineInOuputStartsWith("quay.io/libpod/alpine")).To(BeTrue())
+		Expect(session.LineInOuputStartsWith("quay.io/libpod/busybox")).To(BeTrue())
 	})
 
 	It("podman images with no images prints header", func() {
@@ -62,8 +62,8 @@ var _ = Describe("Podman images", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
 		Expect(len(session.OutputToStringArray())).To(BeNumerically(">", 2))
-		Expect(session.LineInOuputStartsWith("docker.io/library/alpine")).To(BeTrue())
-		Expect(session.LineInOuputStartsWith("docker.io/library/busybox")).To(BeTrue())
+		Expect(session.LineInOuputStartsWith("quay.io/libpod/alpine")).To(BeTrue())
+		Expect(session.LineInOuputStartsWith("quay.io/libpod/busybox")).To(BeTrue())
 	})
 
 	It("podman images with multiple tags", func() {
@@ -80,13 +80,13 @@ var _ = Describe("Podman images", func() {
 		session = podmanTest.PodmanNoCache([]string{"images"})
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
-		session.LineInOutputContainsTag("docker.io/library/alpine", "latest")
-		session.LineInOutputContainsTag("docker.io/library/busybox", "glibc")
-		session.LineInOutputContainsTag("foo", "a")
-		session.LineInOutputContainsTag("foo", "b")
-		session.LineInOutputContainsTag("foo", "c")
-		session.LineInOutputContainsTag("bar", "a")
-		session.LineInOutputContainsTag("bar", "b")
+		Expect(session.LineInOutputContainsTag("quay.io/libpod/alpine", "latest")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("quay.io/libpod/busybox", "latest")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/foo", "a")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/foo", "b")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/foo", "c")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/bar", "a")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/bar", "b")).To(BeTrue())
 		session = podmanTest.PodmanNoCache([]string{"images", "-qn"})
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
@@ -98,8 +98,8 @@ var _ = Describe("Podman images", func() {
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))
 		Expect(len(session.OutputToStringArray())).To(BeNumerically(">", 2))
-		Expect(session.LineInOuputStartsWith("docker.io/library/alpine")).To(BeTrue())
-		Expect(session.LineInOuputStartsWith("docker.io/library/busybox")).To(BeTrue())
+		Expect(session.LineInOuputStartsWith("quay.io/libpod/alpine")).To(BeTrue())
+		Expect(session.LineInOuputStartsWith("quay.io/libpod/busybox")).To(BeTrue())
 	})
 
 	It("podman empty images list in JSON format", func() {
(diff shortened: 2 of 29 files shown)


edsantiago:
Just trying to see what fails in CI, and what I need to focus on.

test/e2e/images_test.go
```diff
@@ -152,10 +152,10 @@ var _ = Describe("Podman images", func() {
 
 	It("podman images filter reference", func() {
 		podmanTest.RestoreAllArtifacts()
-		result := podmanTest.PodmanNoCache([]string{"images", "-q", "-f", "reference=docker.io*"})
+		result := podmanTest.PodmanNoCache([]string{"images", "-q", "-f", "reference=quay.io*"})
 		result.WaitWithDefaultTimeout()
 		Expect(result).Should(Exit(0))
-		Expect(len(result.OutputToStringArray())).To(Equal(2))
+		Expect(len(result.OutputToStringArray())).To(Equal(3))
```
edsantiago, line 158: Because `nginx` was already fetched from quay; and now alpine+busybox are as well.

test/e2e/ps_test.go
```diff
@@ -232,12 +232,19 @@ var _ = Describe("Podman ps", func() {
...
+		result := podmanTest.Podman([]string{"ps", "-q", "--no-trunc", "-a", "--filter", "ancestor=quay.io/libpod/alpine:latest"})
+		result.WaitWithDefaultTimeout()
+		Expect(result.ExitCode()).To(Equal(0))
+		Expect(result.OutputToString()).To(Equal(cid))
+
+		// Query just by image name, without :latest tag
+		result = podmanTest.Podman([]string{"ps", "-q", "--no-trunc", "-a", "--filter", "ancestor=quay.io/libpod/alpine"})
 		result.WaitWithDefaultTimeout()
 		Expect(result.ExitCode()).To(Equal(0))
+		Expect(result.OutputToString()).To(Equal(cid))
```
edsantiago, line 247: I've taken the liberty of making this an actual test, with output checking and all (not just "it exits zero")

test/e2e/images_test.go
```diff
@@ -80,13 +80,13 @@ var _ = Describe("Podman images", func() {
...
-		session.LineInOutputContainsTag("foo", "c")
-		session.LineInOutputContainsTag("bar", "a")
-		session.LineInOutputContainsTag("bar", "b")
+		Expect(session.LineInOutputContainsTag("quay.io/libpod/alpine", "latest")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("quay.io/libpod/busybox", "latest")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/foo", "a")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/foo", "b")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/foo", "c")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/bar", "a")).To(BeTrue())
+		Expect(session.LineInOutputContainsTag("localhost/bar", "b")).To(BeTrue())
```
edsantiago, line 89: This was a complete NOP. Plus, `LineInOutputContainsTag()` was broken. See below.

test/utils/utils.go
```diff
@@ -383,7 +379,10 @@ func tagOutputToMap(imagesOutput []string) map[string]string {
 		if len(tmp) < 2 {
 			continue
 		}
-		m[tmp[0]] = tmp[1]
```
edsantiago, line 386: This was completely broken: for a list containing `foo:a`, `foo:b`, `foo:c`, ... well, you can figure it out.

test/utils/utils.go
```diff
@@ -383,7 +379,10 @@ func tagOutputToMap(imagesOutput []string) map[string]string {
...
+		if m[tmp[0]] == nil {
```
edsantiago, line 382: WARNING: please review carefully, because I don't speak Golang, and there is almost certainly a more idiomatic way to do this.

edsantiago (commented):
I'm getting more and more uncomfortable with this whole situation. Here are some comments about sections I believe need special attention. More later.

edsantiago:
When I first heard about this mirroring approach, I thought it would be: "we will set up our own mirror, and cache what we need from docker.io".

What I now understand is (and please correct me if I'm wrong): "we will use mirror.gcr.io which is owned and set up by some random party out of our control, and it includes a small handful of the images we need, but not all of them, and we have no way to populate it with what we need, so we have to use quay anyway". 

If the latter is true; and given that most of this PR is `s/docker/quay/` because the gcr mirror is so useless, should we just skip the mirroring thing entirely and just go full-on with a quay replacement?

edsantiago:
Well, miracle of miracles, tests are passing. I'm still not sure whether to proceed with a full `quay.io` replacement, but I would like feedback on this as it is now.

test/e2e/exists_test.go
```diff
@@ -39,6 +39,7 @@ var _ = Describe("Podman image|container exists", func() {
 		Expect(session).Should(Exit(0))
 	})
 	It("podman image exists in local storage by short name", func() {
+		Skip("FIXME-8165: shortnames don't seem to work with quay")
```
mheon, line 42: This seems... pretty large? Any idea what's going on?

test/e2e/images_test.go
```diff
@@ -240,6 +240,7 @@ WORKDIR /test
 	})
 
 	It("podman pull by digest and list --all", func() {
+		Skip("FIXME-8165: 'rmi -af' fails with 'layer not known'")
```
mheon, line 243: This one makes me deeply uncomfortable - it's that easy to break our storage in such a way that we can't fix it with `rmi`?

test/e2e/tree_test.go
```diff
@@ -35,7 +35,7 @@ var _ = Describe("Podman image tree", func() {
 
 	It("podman image tree", func() {
 		SkipIfRemote("Does not work on remote client")
-		dockerfile := `FROM docker.io/library/busybox:latest
+		dockerfile := `FROM quay.io/libpod/busybox:latest
```
mheon, line 38: I wonder if we don't have builds that are just using shortnames, not pre-pulled images.

If we completely rip `docker.io` out of registries.conf, I wonder how much breaks...

test/e2e/exists_test.go
edsantiago: I don't, not on this or any of the other two FIXMEs. They worry me too, but my goal was to get tests passing and then dive deeper. I lost too much time on the `map` problem. I can dive deeper into these other problems today, I just want to know if the approach in this PR is reasonable.

mheon: I have no issue with the approach - it seems like a quick and easy fix to get things migrated away from Dockerhub.

test/e2e/images_test.go
edsantiago: This was much, much harder to reproduce than I had expected. Added script to https://github.com/containers/podman/issues/6510#issuecomment-717967486

test/e2e/exists_test.go
edsantiago: @mheon see #8176. Another one that was unexpectedly difficult to reproduce.

edsantiago:
@containers/podman-maintainers PTACL ("C" for "Careful"). Some particular points to note:
* This is based on #7965, but not entirely: I took the liberty of `git rebase -i` to remove one unnecessary commit and to combine two edits to the registries.conf file
  * Speaking of #7965, that one has a 'crun/runc' commit that I don't really understand why it's part of this. @cevich can you comment on that?
* The `search` test still relies heavily on `docker.io`. Fixing that is left as an exercise for a future PR.
* I would feel happier if someone fixed #8171 in a better way than mine, but maybe mine is good enough for now
* There are two new `Skip()s`, both of them real podman bugs (#6510, #8176) related to vfs.

November 1 is this weekend. I don't think this PR is perfect, but I think it's a major step toward getting our CI to pass once docker cuts us off.

I will be out all of tomorrow (Thursday) with no means of communication. I will likely check in Friday morning but cannot promise a full day at work.

rhatdan:
LGTM
I wonder if we should add a job that looks for docker.io in PRs for tests, to make sure they don't leak back in.

vrothberg:
> LGTM
> @containers/podman-maintainers PTAL
> I wonder if we should add a job that looks for docker.io in PRs for tests, to make sure they don't leak back in.

That can't happen in the e2e tests (see https://github.com/containers/podman/pull/8165/commits/53fe386da047cce6a9e9f381b2dd73335babe305) as references to docker.io will either go to the gcr mirror or quay.io/libpod.

LGTM but a third pair of eyes would be appreciated before merging

test/e2e/exists_test.go
```diff
@@ -39,6 +39,7 @@ var _ = Describe("Podman image|container exists", func() {
 		Expect(session).Should(Exit(0))
 	})
 	It("podman image exists in local storage by short name", func() {
+		Skip("FIXME-8165: shortnames don't seem to work with quay (#8176)")
```
TomSweeneyRedHat, line 42: @vrothberg FYI for short names work

test/e2e/systemd_test.go
```diff
@@ -59,7 +59,7 @@ WantedBy=multi-user.target
 			Expect(stop.ExitCode()).To(Equal(0))
 		}()
 
-		create := podmanTest.Podman([]string{"create", "--name", "redis", "redis"})
+		create := podmanTest.Podman([]string{"create", "--name", "redis", redis})
```
TomSweeneyRedHat, line 62: Just double-checking is removing the quotes here intended/valid?

TomSweeneyRedHat:
Just one question, otherwise LGTM and it's all green.  I'm fine with merging now and fixing later if necessary.

test/e2e/exists_test.go
vrothberg: Thanks, Tom!

test/e2e/systemd_test.go
baude: should be yes

test/e2e/exists_test.go
edsantiago: @TomSweeneyRedHat @vrothberg late followup: this does not seem to be related to shortnames. #8176 is too weird for me to describe in a helpful title.

test/e2e/systemd_test.go
edsantiago: @TomSweeneyRedHat @baude, late followup, yes, unquoting "redis" is deliberate: that gives us a variable that references an FQIN.

PR #8166, opened 2020-10-27
Allow users to mount with unbindable flag

rhatdan:
Allow users to specify unbindable on volume command line

Switch internal mounts to rprivate to help prevent leaks.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -541,7 +541,7 @@ Current supported mount TYPEs are **bind**, **volume**, **image**, **tmpfs** and
 
 	      · ro, readonly: true or false (default).
 
-	      · bind-propagation: shared, slave, private, rshared, rslave, or rprivate(default). See also mount(2).
+	      · bind-propagation: shared, slave, private, unbindable, rshared, rslave, runbindable, or rprivate(default). See also mount(2).
 
 	      . bind-nonrecursive: do not setup a recursive bind mount.  By default it is recursive.
 
@@ -962,7 +962,7 @@ The _options_ is a comma delimited list and can be:
 
 * **rw**|**ro**
 * **z**|**Z**
-* [**r**]**shared**|[**r**]**slave**|[**r**]**private**
+* [**r**]**shared**|[**r**]**slave**|[**r**]**private**[**r**]**unbindable**
 * [**r**]**bind**
 * [**no**]**exec**
 * [**no**]**dev**
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
 way mount propagation and that is mounts done on host under that volume
 will be visible inside container but not the other way around. <sup>[[1]](#Footnote1)</sup>
 
-To control mount propagation property of volume one can use `:[r]shared`,
-`:[r]slave` or `:[r]private` propagation flag. Propagation property can
-be specified only for bind mounted volumes and not for internal volumes or
-named volumes. For mount propagation to work source mount point (mount point
-where source dir is mounted on) has to have right propagation properties. For
-shared volumes, source mount point has to be shared. And for slave volumes,
-source mount has to be either shared or slave. <sup>[[1]](#Footnote1)</sup>
+To control mount propagation property of a volume one can use the [**r**]**shared**,
+[**r**]**slave**, [**r**]**private** or the [**r**]**unbindable** propagation flag.
+Propagation property can be specified only for bind mounted volumes and not for
+internal volumes or named volumes. For mount propagation to work the source mount
+point (the mount point where source dir is mounted on) has to have the right propagation
+properties. For shared volumes, the source mount point has to be shared. And for
+slave volumes, the source mount point has to be either shared or slave.
+<sup>[[1]](#Footnote1)</sup>
 
 If you want to recursively mount a volume and all of its submounts into a
 container, then you can use the `rbind` option.  By default the bind option is
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -567,7 +567,7 @@ Current supported mount TYPEs are **bind**, **volume**, **image**, **tmpfs** and
 
 	      · ro, readonly: true or false (default).
 
-	      · bind-propagation: shared, slave, private, rshared, rslave, or rprivate(default). See also mount(2).
+	      · bind-propagation: shared, slave, private, unbindable, rshared, rslave, runbindable, or rprivate(default). See also mount(2).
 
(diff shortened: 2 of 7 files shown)


rhatdan:
According to mount man page

An unbindable mount is a `private`  mount  which cannot be cloned through a bind operation.  
So this should be tighter then before.

mheon:
@rhatdan This doesn't work - I established this in the BZ with Vivek, in the previous PR I opened (https://github.com/containers/podman/pull/8037)

rhvgoyal:
If leaking mount is your problem, then making mounts "rprivate" should be enough. 

Making it unbindable will only help if you don't want users/applications of containers to further bind mount existing mounts points you are setting up for the container. Is that what you want to achieve?

rhatdan:
BindMounts destinations in containers are:
/etc/resolv.conf
/etc/hosts
/etc/passwd
/etc/group
/etc/hostname
/run/.containerenv
/run/secrets
/etc/localtime
/dev/shm (If sharing containers IPC Namespace)

The source of some of these can be shared, but the destination is never shared as a source of another mount.

libpod/container_internal_linux.go
```diff
@@ -343,7 +343,7 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 			Type:        "bind",
 			Source:      srcPath,
 			Destination: dstPath,
-			Options:     []string{"bind", "private"},
+			Options:     []string{"bind", "unbindable"},
```
giuseppe, line 346: should it be `rprivate` instead?

rhatdan: Switched

giuseppe (approved):
LGTM

rhatdan:
The main change here now is allowing unbindable mounts.

vrothberg (commented):
It still leaking with `--volume`. Can we add a test or two?

Tested with `podman run -v $FOO:/run -d ...` and `ls $FOO` will show `secrets`.

mheon:
Secrets is handled elsewhere - in containers/common, I think?

rhatdan:
This PR is not going to block the leak, it can not.  The leak happens because /run is mounted in as shared.  The caller has to mount it in as :private or :rprivate to stop the leak.  If you volume mount in the mount point we default to whatever the system default is.  In the case of Fedora it is either shared or slave.

rhatdan:
@mheon @vrothberg I changed the description of this PR.  Since we can not block the leak, I just talk about adding the unbindable flag support for volume mounts

vrothberg (commented):
LGTM

vrothberg:
/lgtm hold

Restarted the flake

docs/source/markdown/podman-create.1.md
```diff
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
...
 will be visible inside container but not the other way around. <sup>[[1]](#Footnote1)</sup>
 
-To control mount propagation property of volume one can use `:[r]shared`,
-`:[r]slave` or `:[r]private` propagation flag. Propagation property can
-be specified only for bind mounted volumes and not for internal volumes or
-named volumes. For mount propagation to work source mount point (mount point
-where source dir is mounted on) has to have right propagation properties. For
-shared volumes, source mount point has to be shared. And for slave volumes,
-source mount has to be either shared or slave. <sup>[[1]](#Footnote1)</sup>
+To control mount propagation property of volume one can use [**r**]**shared**,
```
TomSweeneyRedHat, line 1051:
```suggestion
To control mount propagation property of a volume one can use the [**r**]**shared**,
```

docs/source/markdown/podman-create.1.md
```diff
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
...
+[**r**]**slave**, [**r**]**private** or [**r**]**unbindable** propagation flag.
```
TomSweeneyRedHat, line 1052:
```suggestion
[**r**]**slave**, [**r**]**private** or the [**r**]**unbindable** propagation flag.
```

docs/source/markdown/podman-create.1.md
```diff
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
...
+Propagation property can be specified only for bind mounted volumes and not for
+internal volumes or named volumes. For mount propagation to work source mount
```
TomSweeneyRedHat, line 1054:
```suggestion
internal volumes or named volumes. For mount propagation to work the source mount
```

docs/source/markdown/podman-create.1.md
```diff
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
...
+point (mount point where source dir is mounted on) has to have right propagation
```
TomSweeneyRedHat, line 1055:
```suggestion
point (the mount point where source dir is mounted on) has to have the right propagation
```

docs/source/markdown/podman-create.1.md
```diff
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
...
+properties. For shared volumes, source mount point has to be shared. And for
```
TomSweeneyRedHat, line 1056:
```suggestion
properties. For shared volumes, the source mount point has to be shared. And for
```

docs/source/markdown/podman-create.1.md
```diff
@@ -1048,13 +1048,14 @@ visible on host and vice versa. Making a volume `slave` enables only one
...
+slave volumes, source mount has to be either shared or slave.
```
TomSweeneyRedHat, line 1057:
```suggestion
slave volumes, the source mount point has to be either shared or slave.
```

docs/source/markdown/podman-run.1.md
```diff
@@ -1099,12 +1099,13 @@ way mount propagation and that is mounts done on host under that volume
...
-named volumes. For mount propagation to work source mount point (mount point
-where source dir is mounted on) has to have right propagation properties. For
-shared volumes, source mount point has to be shared. And for slave volumes,
-source mount has to be either shared or slave. <sup>[[1]](#Footnote1)</sup>
+[**r**]**slave**, [**r**]**private** or [**r**]**unbindable** propagation flag.
+Propagation property can be specified only for bind mounted volumes and not for
+internal volumes or named volumes. For mount propagation to work source mount
+point (mount point where source dir is mounted on) has to have right propagation
+properties. For shared volumes, source mount point has to be shared. And for
+slave volumes, source mount has to be either shared or slave.
```
TomSweeneyRedHat, line 1107: See prior comments

QiWang19 (commented):
/lgtm

Issue #8068, opened 2020-10-19
Cirrus CI: agent stopped responding


edsantiago:
This is now far and away the most common CI flake.

* fedora : Test Bindings
  * PR #7961
    * [10-08 00:33]()
  * PR #7940
    * [10-06 13:43]()
  * PR #7853
    * [10-12 10:08]()
* fedora : int podman fedora-32 root host
  * PR #8061
    * [10-19 11:39]()
  * PR #8044
    * [10-17 10:28]()
  * PR #8039
    * [10-15 23:50]()
  * PR #8028
    * [10-15 05:41]()
  * PR #8021
    * [10-14 15:19]()
    * [10-14 14:39]()
  * PR #8013
    * [10-14 17:57]()
    * [10-14 17:31]()
  * PR #7996
    * [10-12 13:50]()
  * PR #7992
    * [10-16 05:53]()
  * PR #7987
    * [10-13 21:22]()
  * PR #7977
    * [10-09 07:08]()
    * [10-09 06:45]()
  * PR #7969
    * [10-08 15:56]()
  * PR #7967
    * [10-13 12:29]()
  * PR #7949
    * [10-07 13:22]()
  * PR #7919
    * [10-05 11:35]()
    * [10-05 11:12]()
  * PR #7904
    * [10-05 14:39]()
  * PR #7891
    * [10-09 07:29]()
    * [10-09 06:57]()
* fedora : int remote fedora-32 root host
  * PR #8030
    * [10-15 07:58]()
    * [10-15 06:55]()
  * PR #8018
    * [10-14 13:30]()
  * PR #8003
    * [10-13 17:33]()
  * PR #7987
    * [10-13 21:17]()
  * PR #7969
    * [10-08 16:09]()
  * PR #7966
    * [10-08 14:06]()
  * PR #7961
    * [10-08 01:07]()
  * PR #7690
    * [10-09 23:26]()
* fedora : sys podman fedora-32 root host
  * PR #8061
    * [10-19 12:40]()
  * PR #7980
    * [10-12 04:01]()
  * PR #7966
    * [10-08 14:07]()
  * PR #7910
    * [10-07 12:04]()
  * PR #7904
    * [10-05 15:26]()
  * PR #7845
    * [10-07 12:06]()
  * PR #7382
    * [10-07 11:12]()
* fedora : sys remote fedora-32 root host
  * PR #8030
    * [10-15 09:49]()
    * [10-15 09:30]()
    * [10-15 09:11]()
[...]

PR #8169, opened 2020-10-27
DO NOT MERGE: CIRRUS-CI AGENT TESTING

cevich:
See https://github.com/containers/podman/issues/8068#issuecomment-716573893


cevich:
Seeing very similar looking panics on F31 VMs...Will simply include all OS's in the fix...

PR #8170, opened 2020-10-27
Remove build \!remote flags from test

rhatdan:
Add some more tests, document cases where remote will not work
Add FIXMEs for tests that should work on podman-remote but currently
do not.

### DIFF
--- a/test/e2e/containers_conf_test.go
+++ b/test/e2e/containers_conf_test.go
@@ -1,5 +1,3 @@
-// +build !remote
-
 package integration
 
 import (
@@ -31,6 +29,10 @@ var _ = Describe("Podman run", func() {
 		podmanTest.Setup()
 		podmanTest.SeedImages()
 		os.Setenv("CONTAINERS_CONF", "config/containers.conf")
+		if IsRemote() {
+			podmanTest.RestartRemoteService()
+		}
+
 	})
 
 	AfterEach(func() {
@@ -80,12 +82,14 @@ var _ = Describe("Podman run", func() {
 	})
 
 	It("podman Capabilities in containers.conf", func() {
-		os.Setenv("CONTAINERS_CONF", "config/containers.conf")
 		cap := podmanTest.Podman([]string{"run", ALPINE, "grep", "CapEff", "/proc/self/status"})
 		cap.WaitWithDefaultTimeout()
 		Expect(cap.ExitCode()).To(Equal(0))
 
 		os.Setenv("CONTAINERS_CONF", "config/containers-ns.conf")
+		if IsRemote() {
+			podmanTest.RestartRemoteService()
+		}
 		session := podmanTest.Podman([]string{"run", "busybox", "grep", "CapEff", "/proc/self/status"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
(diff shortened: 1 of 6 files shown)


rhatdan:
More tests for podman-remote to keep @edsantiago happy.  :^)

test/e2e/push_test.go
```diff
@@ -163,6 +165,7 @@ var _ = Describe("Podman push", func() {
 	})
 
 	It("podman push to docker-archive", func() {
+		SkipIfRemote("Rootless push does not support docker-archive transport")
```
ashley-cui, line 168: Should this be SkipIfRootless? Or should it say Remote push?

rhatdan: Yes this should say Remote push does not support.  Will wait to see what tests pass.

TomSweeneyRedHat:
Changes LGTM, but tests aren't hip.

edsantiago:
Test failures are all timeouts. Restarted.

edsantiago:
Timeouts again. I noticed today that in my own PR, remote tests were taking 1h24. Given that the CI timeout is 90m, I think we have to speed up remote tests before we can merge this.

rhatdan:
@baude I think we have to wait for you to speed up remote tests.

rhatdan:
@baude These tests are still failing on 1.30 minutes timeout?

edsantiago:
Weird: the logs in all four `int remote` cases show that it successfully finished all tests, and was dumping the timing results. It seems extraordinarily unlikely that in all four cases the tests completed, then a timeout happened in a `printf` loop that should take less than a second. I'm suspecting some sort of cleanup timeout: is it possible that something in this PR is leaving behind a process, perhaps an unkillable one, and the e2e tests are hanging on `wait` at the end?

rhatdan:
@edsantiago That would make sense.  But I am not sure which one.  I could start to break these up into separate PRs to try to narrow it down.

rhatdan:
@edsantiago Split the PR in 2 to see if one can pass.
https://github.com/containers/podman/pull/8379

rhatdan:
This adds a bunch more tests.

vrothberg (commented):
LGTM

test/e2e/containers_conf_test.go
```diff
@@ -81,11 +83,17 @@ var _ = Describe("Podman run", func() {
 
 	It("podman Capabilities in containers.conf", func() {
 		os.Setenv("CONTAINERS_CONF", "config/containers.conf")
+		if IsRemote() {
+			podmanTest.RestartRemoteService()
+		}
```
edsantiago, line 88: This is probably just ginkgo-ignorance on my part, but: given that `BeforeEach()` already does this (Setenv,Restart), why is it necessary in this case, and lines 104-107, 118-121, and 132-135? (93-97 is correct because it changes the envariable).

rhatdan: It most likely is not.

rhatdan: Removed

edsantiago:
This was a [horrible scary set of failures](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5933768118632448/html/int-remote-ubuntu-19-root-host.log.html) but it turned out to be a flake. Has anyone seen these before?

mheon:
`         Error: unable to obtain cgroup stats: open /sys/fs/cgroup/pids/pids.current: no such file or directory`

Believe this one's an old friend, no?

edsantiago:
This one, I don't recognize. `cgroup.freeze` is my old CI-flake friend, not this one. The closest I can find is #5728 but that's over six months old (and closed).

Issue #8105, opened 2020-10-21
REST API v2 rootless - PublishAllPorts not working


psakar:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Container created with PublishAllPorts=true does not bind ports

**Steps to reproduce the issue:**

1. Create nginx container and publish all ports
```
podman --log-level=debug system service -t 0 tcp:localhost:8880 2>&1>/tmp/podman-rest.log &

curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/create?name=nginx' --data '{"name":"nginx","Image":"nginx:alpine", "HostConfig":{"PublishAllPorts":true}}'
curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8880/containers/nginx/start'
```

**Describe the results you received:**
Port binding is missing (checked with netstat and by inspecting container)
```
HostConfig": {
            "Binds": [],
            "CgroupMode": "private",
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "",
                "Config": null
            },
            "NetworkMode": "slirp4netns",
            "PortBindings": {},
```

**Describe the results you expected:**
Port binding is created
```
        "HostConfig": {
            "Binds": [],
            "CgroupMode": "private",
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "",
                "Config": null
            },
            "NetworkMode": "slirp4netns",
            "PortBindings": {
                "80/tcp": [
                    {
                        "HostIp": "",
                        "HostPort": "39995"
                    }
                ]
            },
```
Please note host port will be most probably different

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 2.1.0
```

psakar:
Related to https://github.com/containers/podman/issues/7927

mheon:
Strongly suspect this is already fixed on master thanks to the work from
@baude

psakar:
I tried to test it with master, but I'm not able to run my test using testcontainers against podman built from master - attempt to start the container results in 
```
 Request Failed(Internal Server Error): open io.max: No such file or directory: OCI runtime command not found error 
{"cause":"OCI runtime command not found error","message":"open io.max: No such file or directory: OCI runtime command not found error","response":500}
```

psakar:
[...]

PR #8174, opened 2020-10-28
Podman often reports OCI Runtime does not exist, even if it does

rhatdan:
When the OCI Runtime tries to set certain settings in cgroups
it can get the error "no such file or directory",  the wrapper
ends up reporting a bogus error like:

```
 Request Failed(Internal Server Error): open io.max: No such file or directory: OCI runtime command not found error
{"cause":"OCI runtime command not found error","message":"open io.max: No such file or directory: OCI runtime command not found error","response":500}
```

On first reading of this, you would think the OCI Runtime (crun or runc) were not found.  But the error is actually reporting

message":"open io.max: No such file or directory

Which is what we want the user to concentrate on.

### DIFF
--- a/libpod/define/errors.go
+++ b/libpod/define/errors.go
@@ -138,15 +138,15 @@ var (
 
 	// ErrOCIRuntimePermissionDenied indicates the OCI runtime attempted to invoke a command that returned
 	// a permission denied error
-	ErrOCIRuntimePermissionDenied = errors.New("OCI runtime permission denied error")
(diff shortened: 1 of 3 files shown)


rhatdan:
Related to https://github.com/containers/podman/issues/8105

mheon:
LGTM

jwhonce:
/lgtm
/hold
pending test results

TomSweeneyRedHat:
LGTM

Issue #8176, opened 2020-10-28
podman image exists, with vfs.imagestore and image from quay, fails


edsantiago:
I have no idea how to even describe this, so someone please feel free to change the title to a more descriptive one.

Discovered during testing of #8165 -- a podman setup using `vfs` and `vfs.imagestore` will, under certain circumstances involving images from `quay.io`, falsely claim false on `image exists`. A reproducer is worth a thousand of my words:
```bash
#!/bin/bash

set -ex

PROOT=/dev/shm/mypodman
rm -rf $PROOT
mkdir -p $PROOT

# What's really weird is that this test *succeeds* with docker.io
IMAGE=quay.io/libpod/alpine:latest
#IMAGE=docker.io/library/alpine:latest

# Pull something in order to initialize imagestore
# Without this initial pull, the last pull below fails with:
# Error: error opening "/dev/shm/mypodman/cache/vfs-images/images.lock": no such file or directory
PODMAN="./bin/podman --root $PROOT/cache --storage-driver vfs"
$PODMAN pull -q $IMAGE

# And now, switch --root to something empty, and set imagestore to the above.
# Pull the same image, and boom.
# If you comment out the 'pull' below, 'image exists' succeeds.
PODMAN="./bin/podman --storage-opt vfs.imagestore=$PROOT/cache --root $PROOT/root --storage-driver vfs"
$PODMAN pull -q $IMAGE

# Shows two images, differing only in the rightmost 'R/O' column
$PODMAN images -a

# This fails!
$PODMAN image exists alpine
```

edsantiago:
I don't think this is fixed; the [original test continues to fail](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5429634319777792/html/int-podman-fedora-31-rootless-host.log.html#t--podman-image-exists-in-local-storage-by-short-name--1). So does the reproducer in comment 0.

rhatdan:
This test is being screwed up by the mirroring.

If you run images, you see that you have these images.
REPOSITORY             TAG     IMAGE ID      CREATED        SIZE     R/O
quay.io/libpod/alpine  latest  961769676411  15 months ago  5.85 MB  false
quay.io/libpod/alpine  latest  961769676411  15 months ago  5.85 MB  true

But the image exists is looking for 
docker.io/library/alpine

Since the image exists does not go through the mirror code, it fails.

rhatdan:
[...]

PR #8178, opened 2020-10-28
NewFromLocal can return multiple images

rhatdan:
If you use additional stores and pull the same image into
writable stores, you can end up with the situation where
you have the same image twice. This causes image exists
to return the wrong error.  It should return true in this
situation rather then an error.

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -44,9 +44,14 @@ func ImageExists(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 	name := utils.GetName(r)
 
-	_, err := runtime.ImageRuntime().NewFromLocal(name)
+	ir := abi.ImageEngine{Libpod: runtime}
+	report, err := ir.Exists(r.Context(), name)
 	if err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusNotFound, errors.Wrapf(err, "failed to find image %s", name))
+		return
+	}
+	if !report.Value {
+		utils.Error(w, "Something went wrong.", http.StatusNotFound, errors.Wrapf(nil, "failed to find image %s", name))
 		return
 	}
 	utils.WriteResponse(w, http.StatusNoContent, "")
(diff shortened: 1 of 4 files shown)


rhatdan:
Fixes https://github.com/containers/podman/issues/8176

pkg/api/handlers/libpod/images.go
```diff
@@ -44,12 +44,14 @@ func ImageExists(w http.ResponseWriter, r *http.Request) {
...
-	_, err := runtime.ImageRuntime().NewFromLocal(name)
+	ir := abi.ImageEngine{Libpod: runtime}
+	report, err := ir.Exists(r.Context(), name)
 	if err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusNotFound, errors.Wrapf(err, "failed to find image %s", name))
 		return
 	}
-	utils.WriteResponse(w, http.StatusNoContent, "")
+
+	utils.WriteResponse(w, http.StatusOK, report)
```
mheon, line 54: This looks like a breaking change to the API

pkg/domain/infra/abi/images.go
```diff
@@ -39,8 +39,14 @@ const SignatureStoreDir = "/var/lib/containers/sigstore"
 
 func (ir *ImageEngine) Exists(_ context.Context, nameOrID string) (*entities.BoolReport, error) {
 	_, err := ir.Libpod.ImageRuntime().NewFromLocal(nameOrID)
-	if err != nil && errors.Cause(err) != define.ErrNoSuchImage {
-		return nil, err
+	if err != nil {
+		if errors.Cause(err) != define.ErrMultipleImages {
```
mheon, line 43: Shouldn't this be `==` instead of `!=`

pkg/api/handlers/libpod/images.go
rhatdan: I don't know, I believe the previous version was broken.  @jwhonce WDYT?

pkg/domain/infra/abi/images.go
rhatdan: Yes

pkg/api/handlers/libpod/images.go
rhatdan: Ok I went back to original API, although I still believe it is slightly broken, but usable.

rhatdan:
@edsantiago PTAL

pkg/api/handlers/libpod/images.go
```diff
@@ -44,11 +44,16 @@ func ImageExists(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 	name := utils.GetName(r)
 
-	_, err := runtime.ImageRuntime().NewFromLocal(name)
+	ir := abi.ImageEngine{Libpod: runtime}
+	report, err := ir.Exists(r.Context(), name)
```
vrothberg, line 48: :100: for using the ABI

vrothberg (commented):
LGTM

PR #8185, opened 2020-10-29
Add support for mounting external containers

rhatdan:
Continue progress on use of external containers.

This PR adds the ability to mount, umount and list the
storage containers whether they are in libpod or not.

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -918,6 +918,56 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
 	return prunedContainers, pruneErrors, nil
 }
 
+// MountStorageContainer mounts the storage container's root filesystem
+func (r *Runtime) MountStorageContainer(id string) (string, error) {
+	if _, err := r.GetContainer(id); err == nil {
+		return "", errors.Wrapf(define.ErrCtrExists, "ctr %s is a libpod container", id)
+	}
+	container, err := r.store.Container(id)
+	if err != nil {
+		return "", err
+	}
+	mountPoint, err := r.store.Mount(container.ID, "")
+	if err != nil {
+		return "", errors.Wrapf(err, "error mounting storage for container %s", id)
+	}
+	return mountPoint, nil
+}
+
+// UnmountStorageContainer unmounts the storage container's root filesystem
+func (r *Runtime) UnmountStorageContainer(id string, force bool) (bool, error) {
+	if _, err := r.GetContainer(id); err == nil {
+		return false, errors.Wrapf(define.ErrCtrExists, "ctr %s is a libpod container", id)
+	}
+	container, err := r.store.Container(id)
+	if err != nil {
+		return false, err
+	}
+	return r.store.Unmount(container.ID, force)
+}
+
+// MountedStorageContainer returns whether a storage container is mounted
+// along with the mount path
+func (r *Runtime) IsStorageContainerMounted(id string) (bool, string, error) {
+	var path string
+	if _, err := r.GetContainer(id); err == nil {
+		return false, "", errors.Wrapf(define.ErrCtrExists, "ctr %s is a libpod container", id)
+	}
+
+	mountCnt, err := r.storageService.MountedContainerImage(id)
+	if err != nil {
+		return false, "", err
+	}
+	mounted := mountCnt > 0
+	if mounted {
+		path, err = r.storageService.GetMountpoint(id)
+		if err != nil {
+			return false, "", err
+		}
+	}
+	return mounted, path, nil
+}
+
 // StorageContainers returns a list of containers from containers/storage that
 // are not currently known to Podman.
 func (r *Runtime) StorageContainers() ([]storage.Container, error) {
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -1058,11 +1058,23 @@ func (ic *ContainerEngine) ContainerMount(ctx context.Context, nameOrIDs []strin
 			os.Exit(ret)
 		}
 	}
-	ctrs, err := getContainersByContext(options.All, options.Latest, nameOrIDs, ic.Libpod)
+	reports := []*entities.ContainerMountReport{}
+	// Attempt to mount named containers directly from storage,
+	// this will fail and code will fall through to removing the container from libpod.`
+	names := []string{}
+	for _, ctr := range nameOrIDs {
+		report := entities.ContainerMountReport{Id: ctr}
+		if report.Path, report.Err = ic.Libpod.MountStorageContainer(ctr); report.Err != nil {
+			names = append(names, ctr)
+		} else {
+			reports = append(reports, &report)
+		}
+	}
+
+	ctrs, err := getContainersByContext(options.All, options.Latest, names, ic.Libpod)
 	if err != nil {
 		return nil, err
 	}
-	reports := make([]*entities.ContainerMountReport, 0, len(ctrs))
 	for _, ctr := range ctrs {
 		report := entities.ContainerMountReport{Id: ctr.ID()}
 		report.Path, report.Err = ctr.Mount()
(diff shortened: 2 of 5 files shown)


rhatdan:
This is now passing tests.

libpod/runtime_ctr.go
```diff
@@ -918,6 +918,47 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
...
+	mountPoint, err := r.store.Mount(container.ID, "")
+	if err != nil {
+		return "", errors.Wrapf(err, "error mounting storage for container %s", id)
+	}
+	return mountPoint, nil
+}
+
+// UnmountStorageContainer unmounts the storage container's root filesystem
+func (r *Runtime) UnmountStorageContainer(id string, force bool) (bool, error) {
+	container, err := r.store.Container(id)
```
mheon, line 936: We should probably check here is this is a Libpod container

libpod/runtime_ctr.go
```diff
@@ -918,6 +918,47 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
 	return prunedContainers, pruneErrors, nil
 }
 
+// MountStorageContainer mounts the storage container's root filesystem
+func (r *Runtime) MountStorageContainer(id string) (string, error) {
+	container, err := r.store.Container(id)
```
mheon, line 923: Same - we should check is this is actually a Libpod container

libpod/runtime_ctr.go
```diff
@@ -918,6 +918,47 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
...
+	if err != nil {
+		return false, err
+	}
+	return r.store.Unmount(container.ID, force)
+}
+
+// MountedStorageContainer returns whether a storage container is mounted
+// along with the mount path
+func (r *Runtime) MountedStorageContainer(id string) (bool, string, error) {
```
mheon, line 945: Suggest renaming to "IsStorageContainerMounted"

libpod/runtime_ctr.go
```diff
@@ -918,6 +918,47 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
...
+	var path string
+
+	mountCnt, err := r.storageService.MountedContainerImage(id)
```
mheon, line 948: Should check if this is a storage container

pkg/domain/infra/abi/containers.go
```diff
@@ -1058,11 +1058,23 @@ func (ic *ContainerEngine) ContainerMount(ctx context.Context, nameOrIDs []strin
...
 		}
 	}
-	ctrs, err := getContainersByContext(options.All, options.Latest, nameOrIDs, ic.Libpod)
+	reports := []*entities.ContainerMountReport{}
+	// Attempt to mount named containers directly from storage,
+	// this will fail and code will fall through to removing the container from libpod.`
+	names := []string{}
+	for _, ctr := range nameOrIDs {
+		report := entities.ContainerMountReport{Id: ctr}
+		if report.Path, report.Err = ic.Libpod.MountStorageContainer(ctr); report.Err != nil {
```
mheon, line 1067: I don't think this is safe - this feels like it should come *after* the Libpod mount logic?

libpod/runtime_ctr.go
```diff
@@ -918,6 +918,56 @@ func (r *Runtime) PruneContainers(filterFuncs []ContainerFilter) (map[string]int
 	return prunedContainers, pruneErrors, nil
 }
 
+// MountStorageContainer mounts the storage container's root filesystem
+func (r *Runtime) MountStorageContainer(id string) (string, error) {
```
mheon, line 922: You note in the test that this doesn't work rootless - should we check for that in all of the API functions, so we can return sane errors?

pkg/domain/infra/abi/containers.go
rhatdan: With the check to see if this is a libpod container, this should work fine now.

libpod/runtime_ctr.go
rhatdan: Well it will work rootless if you do a podman unshare. The problem is we can not do this in the test suite easily.
podman unshare podman mount alpine-working-container
Will work fine.

mheon: Can we check if we're inside the user namespace?

rhatdan: This issue has existed for all podman mounts not just external mounts.  

There already is a check in the podman command.
podman mount -l
Error: cannot run command "podman mount" in rootless mode, must execute `podman unshare` first

rhatdan:
@mheon PTAL again.

TomSweeneyRedHat:
This very much smells like a future blog post once it's merged.
Changes LGTM, but definitely want a @mheon head nod.

test/system/060-mount.bats
```diff
@@ -151,4 +150,31 @@ load helpers
     run_podman rm -f $cid
 }
 
+@test "podman mount external container - basic test" {
+    # Only works with root (FIXME: does it work with rootless + vfs?)
+    skip_if_rootless "mount does not work rootless"
+    skip_if_remote "mounting remote is meaningless"
+
+    f_path=/tmp/tmpfile_$(random_string 8)
+    f_content=$(random_string 30)
```
edsantiago, line 159: Neither of these are used; please remove

test/system/060-mount.bats
```diff
@@ -151,4 +150,31 @@ load helpers
...
+
+    test -d $mount_path
+
+    # Make sure that 'podman mount' (no args) returns the expected path
+    run_podman mount --notruncate
+
+    reported_mountpoint=$(echo "$output" | awk '{print $2}')
+    is $reported_mountpoint $mount_path "mountpoint reported by 'podman mount'"
+
+    # umount, and make sure files are gone
```
edsantiago, line 175: Misleading comment: there's no files-gone test. Friendly suggestion:
```diff
diff --git a/test/system/060-mount.bats b/test/system/060-mount.bats
index 0dce35430..b6f1561f7 100644
--- a/test/system/060-mount.bats
+++ b/test/system/060-mount.bats
@@ -155,16 +155,16 @@ load helpers
     skip_if_rootless "mount does not work rootless"
     skip_if_remote "mounting remote is meaningless"

-    f_path=/tmp/tmpfile_$(random_string 8)
-    f_content=$(random_string 30)
-
     # Create a container that podman does not know about
     external_cid=$(buildah from $IMAGE)

     run_podman mount $external_cid
     mount_path=$output

+    # Test image will always have this file, and will always have the tag
     test -d $mount_path
+    is $(< "$mount_path/home/podman/testimage-id") "$PODMAN_TEST_IMAGE_TAG"  \
+       "Contents of well-known file in image"

     # Make sure that 'podman mount' (no args) returns the expected path
     run_podman mount --notruncate
@@ -174,6 +174,10 @@ load helpers

     # umount, and make sure files are gone
     run_podman umount $external_cid
+    if [ -d "$mount_path" ]; then
+        die "'podman umount' did not umount"
+    fi
+
     buildah rm $external_cid
 }

```

edsantiago (commented):
LGTM, some suggestions for the test.

Incidentally, I notice that `podman umount -a` does not umount these. Should it? Given that `podman mount` lists them, I think it violates POLA for `umount -a` not to see them.
```console
$ sudo buildah from quay.io/libpod/testimage:20200929
testimage-working-container
$ sudo ./bin/podman mount testimage-working-container
/var/lib/containers/storage/overlay/f154f71a9590e6cefbdb1ea3b27c1b20e6ce8d1dea1f22774f50915d456e5746/merged
$ sudo ./bin/podman mount
cdc8c6b8d4a6  /var/lib/containers/storage/overlay/f154f71a9590e6cefbdb1ea3b27c1b20e6ce8d1dea1f22774f50915d456e5746/merged
$ sudo ./bin/podman umount -a
$ sudo ./bin/podman mount
cdc8c6b8d4a6  /var/lib/containers/storage/overlay/f154f71a9590e6cefbdb1ea3b27c1b20e6ce8d1dea1f22774f50915d456e5746/merged
^^^^^^ UNEXPECTED: I had thought that umount -a would take care of it
$ sudo ./bin/podman umount cdc
$ sudo ./bin/podman mount
[now, no output]
```

test/system/060-mount.bats
rhatdan: Love cut and paste errors.

rhatdan:
@edsantiago Since I don't believe podman mount -a should mount external containers, podman umount -a should not unmount them.  I added documentation of this to the podman mount and podman umount man pages.  PTAL

edsantiago:
>@edsantiago Since I don't believe podman mount -a should mount external containers, podman umount -a should not unmount them. I added documentation of this to the podman mount and podman umount man pages. PTAL

I 100% agree about `mount -a`, but I'm still concerned about `umount`. If I as a user run `podman umount -a`, I really expect it to unmount everything listed by `podman mount`. I can envision situations in which a user runs a script or tool which (unbeknownst to them)  does an external mount and leaves it behind. The user then tries to clean up and is stymied because they can't easily unmount. I find it jarring and confusing. If you feel strongly about `umount -a` not umounting -a, would you consider adding a warning like "will not unmount external container xyz"??

rhatdan:
I am willing to go along with umount -a unmounting all, my only argument against is based on symmetry.  I think for the most part very few if any users will ever do mount -a and unmount -a, anyways.

rhatdan:
@edsantiago podman umount -a will now unmount --external containers.

edsantiago:
Thank you! LGTM.

PR #8186, opened 2020-10-29
Create the default root API address path

### LINKED ISSUE #8184: Default rootful path to podman.sock not present
Found while creating slides.

We need to make sure that the default path is created `os.MkdirAll(...)`. I had a quick look but couldn't find the right place.
```
$ sudo ./bin/podman system service 
ERRO[0000] /run/podman/podman.sock                      
Error: unable to create socket unix:/run/podman/podman.sock: listen unix /run/podman/podman.sock: bind: no such file or directory
```
@jwhonce PTAL

Luap99:
Fixes #8184

### DIFF
--- a/cmd/podman/system/service.go
+++ b/cmd/podman/system/service.go
@@ -131,20 +131,17 @@ func resolveAPIURI(_url []string) (string, error) {
 		if srvArgs.Varlink {
 			socketName = "io.podman"
 		}
-		socketDir := filepath.Join(xdg, "podman", socketName)
-		if _, err := os.Stat(filepath.Dir(socketDir)); err != nil {
-			if os.IsNotExist(err) {
-				if err := os.Mkdir(filepath.Dir(socketDir), 0755); err != nil {
-					return "", err
-				}
-			} else {
-				return "", err
-			}
+		socketPath := filepath.Join(xdg, "podman", socketName)
+		if err := os.MkdirAll(filepath.Dir(socketPath), 0700); err != nil {
+			return "", err
 		}
-		return "unix:" + socketDir, nil
+		return "unix:" + socketPath, nil
 	case srvArgs.Varlink:
 		return registry.DefaultVarlinkAddress, nil
 	default:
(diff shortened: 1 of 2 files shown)


cmd/podman/system/service.go
```diff
@@ -131,20 +131,17 @@ func resolveAPIURI(_url []string) (string, error) {
...
-		if _, err := os.Stat(filepath.Dir(socketDir)); err != nil {
-			if os.IsNotExist(err) {
-				if err := os.Mkdir(filepath.Dir(socketDir), 0755); err != nil {
-					return "", err
-				}
-			} else {
-				return "", err
-			}
+		socketPath := filepath.Join(xdg, "podman", socketName)
+		if err := os.MkdirAll(filepath.Dir(socketPath), 0755); err != nil {
```
mheon, line 135: Should this be more restrictive than 755? We may want to 700 for security reasons

mheon:
@vrothberg @jwhonce PTAL

cmd/podman/system/service.go
```diff
@@ -131,20 +131,17 @@ func resolveAPIURI(_url []string) (string, error) {
 		if srvArgs.Varlink {
 			socketName = "io.podman"
 		}
-		socketDir := filepath.Join(xdg, "podman", socketName)
-		if _, err := os.Stat(filepath.Dir(socketDir)); err != nil {
```
vrothberg, line 135: I don't think that'll fix https://github.com/containers/podman/issues/8184 entirely as it affects root (as well).

Luap99: root is handled below in the default case

vrothberg: Time to move away from the screen :) Thanks!

cmd/podman/system/service.go
vrothberg: 700 would make sense. Others shouldn't access it.

rhatdan:
LGTM
/approve

TomSweeneyRedHat:
Nice rework @Luap99 .
One test is failing, I've restarted, the only output was a 404 error from it, which I'm assuming is a flake.
LGTM once tests are hip.

PR #8187, opened 2020-10-29
Restore --format table header support

jwhonce:
Restore V1 `--format table... ` header support 

The orIginal code was not always consistent on when headers should be rendered, I followed the tests and v1.9 code here.

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -12,6 +12,7 @@ import (
 
 	tm "github.com/buger/goterm"
 	"github.com/containers/common/pkg/report"
+	"github.com/containers/podman/v2/cmd/podman/parse"
 	"github.com/containers/podman/v2/cmd/podman/registry"
 	"github.com/containers/podman/v2/cmd/podman/utils"
 	"github.com/containers/podman/v2/cmd/podman/validate"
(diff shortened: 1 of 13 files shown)


test/trust_set_test.json
```diff
@@ -5,4 +5,4 @@
         }
     ],
     "transports": null
-}
+}
```
ashley-cui, line 8: CI unhappy here :(

jwhonce: My IDE has different ideas on that file and keeps updating it. :-(

ashley-cui (commented):
LGTM, tested some of them and the table format works :)

jwhonce:
restarting tests "CI agent stopped responding!"

cmd/podman/containers/ps.go
```diff
@@ -180,24 +176,22 @@ func ps(cmd *cobra.Command, args []string) error {
 	switch {
 	case report.IsJSON(listOpts.Format):
 		return jsonOut(listContainers)
-	case listOpts.Quiet:
+	case listOpts.Quiet && !cmd.Flags().Changed("format"):
```
rhatdan, line 179: Should we just throw an error in this situation.  Then you wouldn't need to check later.

If the user specifies --quiet and --format, we should tell them  this is not allowed.

mheon: I think that we have to allow this because Docker did, and people actually came to expect that it worked...

Generally speaking, `--format` overrides `--quiet`.

PR #8189, opened 2020-10-29
Fix the `--pull` flag to `podman build` to match Docker

mheon:
The behavior should be as follows: Unset, pull if missing by default, obey the `--pull-never` and `--pull-always` flags. Set to false, pull never. Set to true, pull always.

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -241,8 +241,12 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 	}
 
 	pullPolicy := imagebuildah.PullIfNewer
-	if c.Flags().Changed("pull") && !flags.Pull {
-		pullPolicy = imagebuildah.PullIfMissing
+	if c.Flags().Changed("pull") {
+		if flags.Pull {
+			pullPolicy = imagebuildah.PullAlways
+		} else {
+			pullPolicy = imagebuildah.PullNever
+		}
 	}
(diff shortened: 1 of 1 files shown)


asottile (approved):
perfect!

cmd/podman/images/build.go
```diff
@@ -241,8 +241,12 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
...
 
 	pullPolicy := imagebuildah.PullIfNewer
-	if c.Flags().Changed("pull") && !flags.Pull {
-		pullPolicy = imagebuildah.PullIfMissing
+	if c.Flags().Changed("pull") {
+		if flags.Pull {
+			pullPolicy = imagebuildah.PullAlways
+		} else {
+			pullPolicy = imagebuildah.PullNever
+		}
```
TomSweeneyRedHat, line 249: This does not comply to our docs.  We claim to do a "PullIfMissing" here in the man pages and a blog post or two.

Yes, we differ from Docker here, but personally I think it's a better difference the way it had just changed to.   Regardless, if we want to exactly match Docker, this change doesn't do it.  `docker build --pull=false` does a PullIfMissing, not PullNever.  

docker build  => PullIfMissing
docker build --pull => PullAlways
docker bulid --pull = true => PullAlways
docker build --pull = false => PullIfMissing

rhatdan: Ok can you fix the --o pullifmissing side of this.

PR #8191, opened 2020-10-29
When container stops, drop sig-proxy errors to infos

mheon:
The sig-proxy code is set up to error on failing to forward signals to a container. This is reasonable in cases where the container is running, but something strange went wrong - but when the Kill fails because the container is stopped, we shouldn't bother with aggressive Error logging since this is an expected part of the container lifecycle - it stops, and then `podman run` also stops, but there is a timing window in between where signals will fail to be proxied, and we should not print angry errors during that.


rhatdan:
LGTM
other then the compiler issues.

TomSweeneyRedHat:
All kinds of test unhappiness, otherwise LGTM

zhangguanzhang:
LGTM

Issue #6365, opened 2020-05-23
podman unable to limit memory (-m flag) on Ubuntu/Debian distros


richpeaua:
/kind bug

**Description**

Was trying to test drive Hashicorp Nomad podman driver on Debian/Ubuntu machines and I ran into an interesting issue. Deployment of containers, via Nomad, fail, yet if I just ran `podman run <image>` it worked fine. 

The errors I was getting were **"failed to write -1 to /sys/fs/cgroup/memory/machine.slice/libpod-2e61c7b46bc2aeed6dadecb07583e97e03ffcc694.scope/memory.memsw.limit_in_bytes"**. 

I looked into that **libpod-*-.scope**  cgroup directory and it's missing the **memory.memsw.limit_in_bytes** resource control file

![image](https://user-images.githubusercontent.com/34759261/82720937-3790ab00-9c86-11ea-9873-5731d20c5840.png)

I then deployed to a Centos8 machine and it worked just fine. I followed up by checking to see if that file was present and sure enough it was.

![image](https://user-images.githubusercontent.com/34759261/82721777-82aebc00-9c8e-11ea-8035-9d0e281524a3.png)

Due to this difference in the cgroup control files podman, and thus Nomad, is unable to limit memory in the Debian/Ubuntu distros

**Steps to reproduce the issue:**

1. sudo podman run -m=40m hello-world (on Ubuntu16-20 or Debian/9/10)

**Describe the results you received:**

`Your kernel does not support swap limit capabilities,or the cgroup is not mounted. Memory limited without swap.
Error: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:415: setting cgroup config for procHooks process caused \\\"failed to write \\\\\\\"-1\\\\\\\" to \\\\\\\"/sys/fs/cgroup/memory/machine.slice/libpod-8639bb6811f6bec5282ee72dac48abb6a647ce7daf84dd000b195dc4fe6f4df7.scope/memory.memsw.limit_in_bytes\\\\\\\": open /sys/fs/cgroup/memory/machine.slice/libpod-8639bb6811f6bec5282ee72dac48abb6a647ce7daf84dd000b195dc4fe6f4df7.scope/memory.memsw.limit_in_bytes: permission denied\\\"\"": OCI runtime permission denied error`

**Describe the results you expected:**

Output of hello-world container/successful job deployment status on Nomad

**Additional information you deem important (e.g. issue happens only occasionally):**

[...]

PR #8197, opened 2020-10-30
specgen, cgroup2: check whether memory swap is enabled

giuseppe:
add a similar check to what we do on cgroup v1.

### DIFF
--- a/pkg/specgen/generate/validate.go
+++ b/pkg/specgen/generate/validate.go
@@ -7,15 +7,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Verify resource limits are sanely set, removing any limits that are not
-// possible with the current cgroups config.
-func verifyContainerResources(s *specgen.SpecGenerator) ([]string, error) {
+// Verify resource limits are sanely set when running on cgroup v1.
+func verifyContainerResourcesCgroupV1(s *specgen.SpecGenerator) ([]string, error) {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
Doesn't this fix an issue?

pkg/specgen/generate/validate.go
```diff
@@ -1,32 +1,28 @@
...
 
 	if s.ResourceLimits == nil {
 		return warnings, nil
 	}
 
 	if s.ResourceLimits.Unified != nil {
-		if !cgroup2 {
-			return nil, errors.New("Cannot use --cgroup-conf without cgroup v2")
-		}
+		return nil, errors.New("Cannot use --cgroup-conf without cgroup v2")
```
mheon, line 25: Don't we still need the v2 check here?

mheon: Ah, I see, we have separate functions for v1 and v2

giuseppe: `verifyContainerResourcesCgroupV1` is called only on cgroup v1, so the extra check is superflous

mheon:
LGTM

rhatdan:
Fixes: https://github.com/containers/podman/issues/6365

rhatdan:
@giuseppe Tests are not happy.

PR #8204, opened 2020-10-30
Add test/apiv2/rest_api tests to make target

jwhonce:
* renamed old API tests to not be discovered, they do not pass
* Updated the API tests to use a pristine storage configuration
* Skipped attach test, it needs to be re-written
* Code re-formatter with `black` formatter

### DIFF
--- a/test/apiv2/rest_api/__init__.py
+++ b/test/apiv2/rest_api/__init__.py
@@ -1,0 +1,132 @@
+import configparser
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+
+
+class Podman(object):
+    """
+    Instances hold the configuration and setup for running podman commands
+    """
+
+    def __init__(self):
+        """Initialize a Podman instance with global options"""
+        binary = os.getenv("PODMAN", "bin/podman")
+        self.cmd = [binary, "--storage-driver=vfs"]
+
+        cgroupfs = os.getenv("CGROUP_MANAGER", "cgroupfs")
+        self.cmd.append(f"--cgroup-manager={cgroupfs}")
+
+        if os.getenv("DEBUG"):
+            self.cmd.append("--log-level=debug")
+
+        self.anchor_directory = tempfile.mkdtemp(prefix="podman_restapi_")
+        self.cmd.append("--root=" + os.path.join(self.anchor_directory, "crio"))
+        self.cmd.append("--runroot=" + os.path.join(self.anchor_directory, "crio-run"))
+
+        os.environ["REGISTRIES_CONFIG_PATH"] = os.path.join(
+            self.anchor_directory, "registry.conf"
+        )
+        p = configparser.ConfigParser()
+        p.read_dict(
+            {
+                "registries.search": {"registries": "['docker.io']"},
+                "registries.insecure": {"registries": "[]"},
+                "registries.block": {"registries": "[]"},
+            }
+        )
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
@jwhonce looks like at least some of the tests are kicking up their heals.

jwhonce:
@cevich I suspect the errors are caused by python version mismatch. Which version of python should I have targeted? I'm on fc33 so the code currently runs on python 3.9

jwhonce:
> @jwhonce looks like at least some of the tests are kicking up their heals.

@TomSweeneyRedHat yeah, I suspect a change in the python libraries. I need to get with @cevich Monday.

test/apiv2/rest_api/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self):
+        """Initialize a Podman instance with global options"""
+        binary = os.getenv("PODMAN", "bin/podman")
+        self.cmd = [binary, "--storage-driver=vfs"]
+
+        cgroupfs = os.getenv("CGROUP_MANAGER", "cgroupfs")
+        self.cmd.append(f"--cgroup-manager={cgroupfs}")
+
+        if os.getenv("DEBUG"):
+            self.cmd.append(["--log-level=debug"])
```
edsantiago, line 24: This is the culprit. `$DEBUG` is set in the CI environment, and this statement converts `cmd` into a list of lists. Fix: remove the brackets. (And, if I may make a recommendation, use a less ambiguous envariable such as `$PODMAN_REST_API_DEBUG`)

edsantiago (changes requested):

cevich:
> Which version of python should I have targeted? I'm on fc33 so the code currently runs on python 3.9

My recommendation is to shoot low, not high.  There's no rhyme/reason across distros, and Ubuntu/debian tends to lag begind, esp. the LTS and/or "Prior Ubuntu" flavor (whichever is older).  IIRC, 3.7 is a "safe place" for the next year or so (but worth checking).

test/apiv2/rest_api/__init__.py
cevich: Ahwww dangit :confounded: This is the second time my "brilliant" choice of `$DEBUG` for the shared automation library has broken something.  Before it gets even more out-of-hand, would either you or @edsantiago be willing to [fix the problem at the source?](https://github.com/containers/automation/blob/master/common/lib/defaults.sh#L10).  I'm very much focused on getting F33 VMs working w/ podman CI, then I need to focus on branch-failure notifications.  Help would be much appreciated.

edsantiago: The bug is in the code. `$DEBUG` triggered it, but was not the cause. 

I think we should make a constant effort to use descriptive conflict-free envariables, but there is no rush to fix the CI setup.

cevich: still `$DEBUG` was a poor choice on my part...in a long consistent line of poor name choices :disappointed: Frankly, I simply no longer trust myself to come up with good names.  It's just not in my DNA :cry:

jwhonce: @edsantiago Good catch! I copied the code from our other repo's and missed that.

baude:
lgtm

rhatdan (approved):
/lgtm

Issue #8131, opened 2020-10-24
--root flag for users without home directory (system/utility users) cause error on /home/user stat


aipec:
/kind bug

**Description**

When running anything in rootless mode, without a home folder causes an error:
`cannot resolve /home/new_user: lstat /home/new_user: no such file or directory`,
even with the flag `--root=/my/dir`

**Steps to reproduce the issue:**

1. sudo useradd --system new_user

2. sudo -u new_user podman run busybox

**Describe the results you received:**
I receive the error above:
`cannot resolve /home/new_user: lstat /home/new_user: no such file or directory`

**Describe the results you expected:**
With some flag, eg`--root=/some/root`, the command should succeed and work as if `/some/root` were `new_user``s home directory.

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14.9
Built:        Wed Sep 30 15:31:11 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-2.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 81d18b6c3ffc266abdef7ca94c1450e669a6a388'
  cpus: REDACTED
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: journald
  hostname: REDACTED
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.15-201.fc32.x86_64
  linkmode: dynamic
  memFree: REDACTED
  memTotal: REDACTED
  ociRuntime:
    name: crun
    package: crun-0.15-5.fc32.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.15
      commit: 56ca95e61639510c7dbd39ff512f80f626404969
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
[...]

PR #8205, opened 2020-10-31
Only use container/storage/pkg/unshare.HomeDir()

rhatdan:
We are resolving the homedir of the user in many different
places.  This Patch consolodates them to use container/storage
version.

This PR also fixes a failure mode when the homedir does not
exists, and the user sets a root path.  In this situation
podman should continue to work. Podman does not require a users
homedir to exist in order to run.

Fixes: https://github.com/containers/podman/issues/8131

### DIFF
--- a/libpod/oci_conmon_exec_linux.go
+++ b/libpod/oci_conmon_exec_linux.go
@@ -444,10 +444,7 @@ func (r *ConmonOCIRuntime) startExec(c *Container, sessionID string, options *Ex
 	// 	}
 	// }
 
-	conmonEnv, extraFiles, err := r.configureConmonEnv(c, runtimeDir)
-	if err != nil {
-		return nil, nil, err
-	}
+	conmonEnv, extraFiles := r.configureConmonEnv(c, runtimeDir)
 
 	var filesToClose []*os.File
 	if options.PreserveFDs > 0 {
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -32,6 +32,7 @@ import (
 	"github.com/containers/podman/v2/pkg/rootless"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/podman/v2/utils"
+	"github.com/containers/storage/pkg/homedir"
 	pmount "github.com/containers/storage/pkg/mount"
 	"github.com/coreos/go-systemd/v22/activation"
 	"github.com/coreos/go-systemd/v22/daemon"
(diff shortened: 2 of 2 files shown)


rhatdan:
@edsantiago WDYT of adding a system test to make sure podman will work without a homedir.

```
sudo useradd -r new_user
sudo -u new_user HOME=/tmp podman run --root /tmp  info
```

TomSweeneyRedHat:
Change LGTM
One of the tests barfed, looks like a flake and I've restarted.
Love the consolidation @rhatdan, very nice change.

rhatdan:
Tests are passing.

pkg/util/utils_supported.go
```diff
@@ -74,19 +75,15 @@ func GetRootlessConfigHomeDir() (string, error) {
...
-			if err != nil {
-				rootlessConfigHomeDirError = errors.Wrapf(err, "cannot resolve %s", home)
-				return
-			}
-			tmpDir := filepath.Join(resolvedHome, ".config")
-			st, err := os.Stat(tmpDir)
-			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && st.Mode().Perm() >= 0700 {
-				cfgHomeDir = tmpDir
-			}
+		cfgHomeDir, err := unshare.HomeDir()
```
vrothberg, line 78: Shouldn't we use `c/storage/pkg/homedir` instead?

I'm a bit confused as I have not known about `unshare.HomeDir`. We usually use `homedir` in other places. Should we consolidate the unshare one into homedir?

vrothberg: Also note that `unshare` does not look at XDG.

rhatdan: Yes you are right, and I also found another bug in the code, where rootlessConfigHomeDirError need to be an external variable.

rhatdan:
This should be ready to merge.

vrothberg (commented):
LGTM

saschagrunert (approved):
/lgtm

edsantiago:
>@edsantiago WDYT of adding a system test to make sure podman will work without a homedir.

Sorry for the late response; it took me a while to organize my ideas.

I don't like the idea of `useradd` in the system tests: these are intended to be run by regular people, too, and creating an account on someone's system is kind of impolite. But in *gating* tests we have all manner of control over the system, and I can see having a system test that does "if running-as-root and well-known no-home-dir account exists, then run this test". I've added it to my TODO list.

PR #8207, opened 2020-10-31
Centralize cores and period/quota conversion code

### LINKED ISSUE #8113: Duplicated millicores <--> quota/period code
I have an architecture question. I recently added [support for resource limits to kube play](https://github.com/containers/podman/commit/a413d4d77f1701a62b9bb2e7beb00145544e2306). In that patch, I added code to convert between k8s "millicores" and CPU quota and period. I just noticed that there is duplicated code for doing this conversion in at least 3 places now (see the references below).

Wouldn't it be good for these to share the same code and reduce the risk of the implementations getting out of sync? All of these are in different packages; where would a good place for the common code to live?

https://github.com/containers/podman/blob/2adc1b284d7f61083d19e82822f79ea14c14de2d/cmd/podman/common/specgen.go#L28-L37

https://github.com/containers/podman/blob/2adc1b284d7f61083d19e82822f79ea14c14de2d/libpod/kube.go#L330

https://github.com/containers/podman/blob/2adc1b284d7f61083d19e82822f79ea14c14de2d/pkg/domain/infra/abi/play.go#L40

https://github.com/containers/podman/blob/2adc1b284d7f61083d19e82822f79ea14c14de2d/pkg/domain/infra/abi/play.go#L517-L522

xordspar0:
Fixes #8113.

This first draft is functional, but I would like feedback on the interface of the functions I wrote. Returning two related integers leaves too many opportunities for the caller to make a mistake. The two integers are different types. Maybe that means the compiler will catch mistakes, but most likely it will produce confusing error messages.

```go
func CoresToPeriodAndQuota(cores float64) (uint64, int64)
func PeriodAndQuotaToCores(period uint64, quota int64) float64
```

An alternative I considered is only returning the quota. The default period would have to be looked up in a const value. I'm not satisfied with this because it means the value returned by CoresToQuota is meaningless until the user looks up the period. This also wouldn't improve the PeriodAndQuotaToCores function.

```go
const DefaultCPUPeriod uint64 = 100000
func CoresToQuota(cores float64) int64
func PeriodAndQuotaToCores(period uint64, quota int64) float64
```

Finally, I also considered using a custom struct. This would give the caller named fields instead of anonymous return values. However, this seems overly complicated to implement for what are basically one-line functions.

```go
type CFSParams struct {
	Period uint64
	Quota int64
}

func CoresToPeriodAndQuota(cores float64) CFSParams
func (c *CFSParams) ToCores() float64
```

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -25,11 +25,8 @@ func getCPULimits(c *ContainerCLIOpts) *specs.LinuxCPU {
 	cpu := &specs.LinuxCPU{}
 	hasLimits := false
 
-	const cpuPeriod = 100000
-
 	if c.CPUS > 0 {
-		quota := int64(c.CPUS * cpuPeriod)
-		period := uint64(cpuPeriod)
+		period, quota := util.CoresToPeriodAndQuota(c.CPUS)
 
 		cpu.Period = &period
 		cpu.Quota = &quota
--- a/libpod/kube.go
+++ b/libpod/kube.go
@@ -327,7 +327,7 @@ func containerToV1Container(c *Container) (v1.Container, []v1.Volume, error) {
 			period := *c.config.Spec.Linux.Resources.CPU.Period
 
 			if quota > 0 && period > 0 {
-				cpuLimitMilli := int64(1000 * float64(quota) / float64(period))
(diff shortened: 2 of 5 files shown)


mheon:
LGTM

rhatdan:
Thanks @xordspar0 
/lgtm

PR #8210, opened 2020-11-01
Match build pull functionality with Docker's

TomSweeneyRedHat:
This is a continuation of #8189 and #8085.

When doing a `docker build` command, if the `--pull` command is not specified
or set to `false` the pullOption used is `PullifMissing`.  This causes the
build to pull the image only if it is not present in local storage.  It also will
raise an error if the image is not found in the registry (or the registry is down),
even if the image is present in local storage.

If the `--pull` command IS specified or specified with an argument of `true`, the
build will always pull the image from the registries.  This uses the pullOption
`PullAlways`.  It also will raise an error if the image is not found in the registry,
even if the image is present in local storage.

These changes now brings the pull functionality for `podman build` into line
with `docker build`.

However, I consider this to be a breaking change.  Previously if you did
`podman build --pull`, `podman build` or `podman build --pull = true`, then
the image would be pulled from the registry if there was not an image in
local storage or if the image in the registry was newer than the one in
local storage.  An error would *NOT* be raised if there was not an image in
the registry or the registry was down as long as there was a copy in the local
storage.  An error would be raised if the image could not be retrieved from
both the registry and local storage.  This is the PullOption `PullIfNewer`.

I believe this also differs from what Buildah does at this time but I'm too
beat to chase that down at the moment.

Personally I'd like to use the `PullIfNewer` for at least `--pull` and
`--pull=true` so that you don't get an error if the registry has a network
hiccup and the image is already stored locally.  But this differs from Docker.

I'd like to post scrum about this at our next stand up to make sure we're
all on the same page about the ramifications of this change.

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -240,13 +240,9 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 		}
 	}
 
-	pullPolicy := imagebuildah.PullIfNewer
-	if c.Flags().Changed("pull") {
-		if flags.Pull {
-			pullPolicy = imagebuildah.PullAlways
-		} else {
-			pullPolicy = imagebuildah.PullNever
-		}
+	pullPolicy := imagebuildah.PullIfMissing
+	if c.Flags().Changed("pull") && flags.Pull {
+		pullPolicy = imagebuildah.PullAlways
 	}
 	if flags.PullAlways {
 		pullPolicy = imagebuildah.PullAlways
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@asottile here we go again! :smiley:

asottile (approved):
🤦 I see -- the `--pull=false` case is the change.  weird that docker just kinda ignores that argument!

rhatdan:
LGTM

rhatdan:

QiWang19 (commented):
LGTM

vrothberg (commented):
LGTM

PR #8213, opened 2020-11-02
Use osusergo build tag for static build

saschagrunert:
We now use the `osusergo` build tag to not use the glibc functions which
occur in the warnings but them from golang the os/user package.


vrothberg (commented):
LGTM
@kolyshkin PTAL

PR #8217, opened 2020-11-02
specgen: add support for ambient capabilities

giuseppe:
if the kernel supports ambient capabilities (Linux 4.3+), also set them when running with euid != 0.

This is different that what Moby does, as ambient capabilities are never set.


vrothberg (commented):
LGTM

Note that `podman top -l capamb` would also work for testing (or demos).

zhangguanzhang:
LGTM

vrothberg (commented):
/lgtm

PR #8218, opened 2020-11-02
[WIP] load/save oci-archive multi images

### LINKED ISSUE #4646: Error loading OCI archive with multiple manifest descriptors

/kind bug

**Description**

I want to be able to tag an oci archive multiple times (similar to the results of `podman tag`, but automatically for oci tars that are loaded into podman). To do this I duplicate the manifest descriptor in `index.json` and give it a different image ref name, but when I try to load the new oci archive into podman, I get an error. 

1. Run:
```
podman pull gcr.io/distroless/java
podman save -o distroless.tar --format oci-archive gcr.io/distroless/java
mkdir distroless-oci
tar -xvf distroless.tar -C distroless-oci
```

2. Duplicate the manifest descriptor in `distroless-oci/index.json` and modify the `org.opencontainers.image.ref.name`, so the contents are as follows:

```
{
  "schemaVersion":2,
  "manifests":[
    {
      "mediaType":"application/vnd.oci.image.manifest.v1+json",
      "digest":"sha256:3a5a334aff603a9f5914640802c4d909c9df285e504c05968d51105adfaa161c",
      "size":821,
      "annotations":{
        "org.opencontainers.image.ref.name":"gcr.io/distroless/java"
      }
    },
    {
      "mediaType":"application/vnd.oci.image.manifest.v1+json",
      "digest":"sha256:3a5a334aff603a9f5914640802c4d909c9df285e504c05968d51105adfaa161c",
      "size":821,
      "annotations":{
        "org.opencontainers.image.ref.name":"gcr.io/distroless/java:tag2"
      }
    }
  ]
}
```

3. Run: `tar -czf new-distroless.tar distroless-oci/`

4. Run either `podman load < new-distroless.tar` or `podman pull oci-archive:new-distroless.tar`

Using `podman load < new-distroless.tar`:

```
  open /var/tmp/podman675604817/manifest.json: not a directory
Error: error pulling "": unable to pull dir:/var/tmp/podman675604817: unable to pull image: Error determining manifest MIME type for dir:/var/tmp/podman675604817: open /var/tmp/podman675604817/manifest.json: not a directory
```

Using `podman pull oci-archive:new-distroless.tar`:

```
Error: error pulling image "oci-archive:jib-image.tar": unable to pull oci-archive:jib-image.tar: error determining pull goal for image "oci-archive:jib-image.tar": error loading manifest for {"jib-image.tar" "/usr/local/google/home/tcordle/IdeaProjects/hellogradle/oci/jib-image.tar" ""}: error loading index: more than one image in oci, choose an image
```

`podman images` loads the oci archive successfully, and lists both `gcr.io/distroless/java:latest` and `gcr.io/distroless/java:tag2`.

I'm essentially trying to replicate what the `RepoTags` field in a docker tar manifest does when you `docker load` the tar. (https://github.com/opencontainers/image-spec/issues/796)

```
Version:            1.6.4-dev
RemoteAPI Version:  1
Go Version:         go1.11.6
Git Commit:         c673ff8cb63c71a39e1a1d561373af42bfea7f24
Built:              Wed Nov 20 11:16:05 2019
```

```
debug:
  compiler: gc
  git commit: c673ff8cb63c71a39e1a1d561373af42bfea7f24
  go version: go1.11.6
  podman version: 1.6.4-dev
host:
  BuildahVersion: 1.11.5
  CgroupVersion: v1
  Conmon:
    package: Unknown
    path: /usr/local/libexec/podman/conmon
    version: 'conmon version 2.0.4-dev, commit: f6d23b51a3c56cccd4e13b4eea361e98e1855802'
  Distribution:
    distribution: debian
    version: rodete
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 89939
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 535966
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 15417282560
  MemTotal: 67534381056
  OCIRuntime:
    name: runc
    package: 'containerd.io: /usr/bin/runc'
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc8+dev
      commit: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
      spec: 1.0.1-dev
  SwapFree: 68631392256
  SwapTotal: 68631392256
  arch: amd64
  cpus: 12
  eventlogger: journald
  hostname: tcordle33.nyc.corp.google.com
  kernel: 4.19.67-2rodete2-amd64
  os: linux
  rootless: true
  slirp4netns:
    Executable: /usr/bin/slirp4netns
    Package: Unknown
    Version: |-
      slirp4netns version 0.4.2
      commit: 69153b0d1cb82216d6782179ff7c3c5e91c731a1
  uptime: 510h 47m 45.11s (Approximately 21.25 days)
registries:
  blocked: null
  insecure: null
  search:
  - docker.io
  - registry.fedoraproject.org
  - registry.access.redhat.com
store:
  ConfigFile: /usr/local/google/home/tcordle/.config/containers/storage.conf
  ContainerStore:
    number: 0
  GraphDriverName: vfs
  GraphOptions: {}
  GraphRoot: /usr/local/google/home/tcordle/.local/share/containers/storage
  GraphStatus: {}
  ImageStore:
    number: 0
  RunRoot: /run/user/535966
  VolumePath: /usr/local/google/home/tcordle/.local/share/containers/storage/volumes
```

QiWang19:
Allow load/save oci-archive format multi images.
```
$ bin/podman save -o two-images.tar -m --format oci-archive gcr.io/distroless/java docker.io/library/alpine
$ bin/podman load -i two-images.tar 
```
close #4646 


libpod/image/image.go
```diff
@@ -257,31 +241,132 @@ func (ir *Runtime) SaveImages(ctx context.Context, namesOrIDs []string, format s
...
+			if err != nil {
+				return err
+			}
 
-		img := imageMap[id]
-		copyOptions := getCopyOptions(sys, writer, nil, nil, SigningOptions{RemoveSignatures: removeSignatures}, "", img.tags)
-		copyOptions.DestinationCtx.SystemRegistriesConfPath = registries.SystemRegistriesConfPath()
+			img := imageMap[id]
+			copyOptions := getCopyOptions(sys, writer, nil, nil, SigningOptions{RemoveSignatures: removeSignatures}, "", img.tags)
+			copyOptions.DestinationCtx.SystemRegistriesConfPath = registries.SystemRegistriesConfPath()
```
zhangguanzhang, line 269: need table format

vrothberg (commented):
Early feedback from testing:

We need to distinguish IDs from names. Currently, IDs are treated as names and stored in annotations which leads to:
```
$ podman load -i ...
Loaded image(s): localhost/79fd58dc7611, localhost/d6e46aa2470d
```

vrothberg (commented):
I dropped some comments on the c/image PR: https://github.com/containers/image/pull/1072#pullrequestreview-522416830

I think there are two things we need to address:
1. We need to support multiple names or none.
2. We need/should be consistent with the reference syntax of docker-archive which identifies images by index: https://github.com/containers/image/blob/master/docs/containers-transports.5.md#docker-archivepathdocker-referencesource-index

QiWang19:
@containers/podman-maintainers @vrothberg @mtrmac  PTAL

libpod/runtime_img.go
```diff
@@ -288,6 +288,11 @@ func (r *Runtime) LoadImage(ctx context.Context, name, inputFile string, writer
 		if err == nil {
 			return getImageNames(newImages), nil
 		}
+
+		newImages, err = r.ImageRuntime().LoadAllImagesFromOCIArchive(ctx, inputFile, signaturePolicy, writer)
+		if err == nil {
+			return getImageNames(newImages), nil
+		}
```
mtrmac, line 295: These “try N different formats, succeed on any one” functions have frequently caused user confusion.

- Would it make sense to have the user specify the format? I guess auto-detection is a desirable feature…
- The error reporting definitely needs improving. If the user submits a docker-archive format that is mostly (or completely) valid but the load fails, the docker-archive-related error _must_ be available to the user.
- (Extra thought, probably not for this PR: It might be interesting to have an “autodetect archive format” helper so that we can detect the format separately (or report a clear error that the format is unrecognized), and then just do one load attempt with a simple error reporting structure. Or, well, we could go even further and share a single “decompress and untar” operation across all formats — right now we would decompress the multiple gigabytes separately for each format.)
- … actually, the fact that we have to decompress the multiple-gigabyte archive in every format until we find one that works is a pretty big cost, and IMHO a fairly strong reason to at least _allow_ the user to specify the format, even we still wanted to support auto-detection.

mtrmac (commented):
From a very quick skim, this results in aa very large function that is going to be very hard to maintain (just read, but primarily keep consistent). It would be nice to share more of the format-independent infrastructure.

Most importantly the `namesOrIDs` parsing code should be shared if at all possible, to ensure the semantics of that parameter is consistent across archive formats — even if the actual copy implementation needs to be different because one can benefit from `DockerArchiveAdditionalTags`. But also the things like converting `imageNames` to `newImages` would probably benefit from sharing (or, well, _eliminating_; the copy already has a storage reference so converting it into a name and back is pointless extra work), and I worry a bit about `newImageEvent` which is not mechanically essential, it might be too easy to forget to update a copy.

(Very tentative: it might be worth splitting the archive save/load code, along with the archive-specific helpers, into a separate file that can be easier to understand as a whole — if you do that, please make it a separate commit that changes _nothing_ in the code.)

zhangguanzhang:
Conflicting files
```
libpod/runtime_img.go
```

QiWang19:
@mtrmac PTAL, Added an archivehelper to decide the format according to the manifest.json to index.json, it this a way to detect the format?

mtrmac:
> @mtrmac PTAL, Added an archivehelper to decide the format according to the manifest.json to index.json, it this a way to detect the format?

That’s a neat idea, for the cases where this code does work it’s cheap enough, and for the expensive case (compressed archive) we can just not do that much work to guess. But because it _can_ fail, AFAICS there still must be the fall-back logic to try all of the formats (well, all of the archive formats or all of the directory formats, no need for both).

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
-			return layout.NewReference(inputFile, "")
-		},
-	} {
-		src, err := referenceFn()
-		if err == nil && src != nil {
-			if newImages, err := r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer); err == nil {
-				return getImageNames(newImages), nil
+// imageFormat detects the image foramt of the input archive or directory
+// returns "", if can not decide the format
+func imageFormat(inputFile string) (string, error) {
```
mtrmac, line 343: `guessImageFormat` maybe, to emphasize that it can fail to make a decision?

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+	path, err := explicitfilepath.ResolvePathToFullyExplicit(inputFile)
```
mtrmac, line 344: Does this do anything that matters for this function?

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+	if err != nil {
+		return "", errors.Wrapf(err, "error detecting input file format")
+	}
+	if !archive.IsArchivePath(inputFile) {
```
mtrmac, line 348: An `os.Stat` + `IsDir` would be a bit cheaper.

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+		return "", errors.Wrapf(err, "error detecting input file format")
+	}
+	reader := tar.NewReader(f)
+	for {
+		header, err := reader.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return "", errors.Wrapf(err, "error detecting archive format")
```
mtrmac, line 374: Is this the path taken on a compressed archive? That should probably return `"", nil` (and the caller must handle that) to indicate that it didn’t make a guess.

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+		}
+		if header.Typeflag == tar.TypeDir {
+			continue
+		}
+		if header.Name == "index.json" {
+			return image.OCIArchive, nil
+		}
+		if header.Name == "manifest.json" {
```
mtrmac, line 382: Apparently for compatibility with some `docker-archive` writers, `path.Clean(header.Name)` must be used; I guess using it for both cases wouldn’t hurt.

libpod/image/image.go
```diff
@@ -210,82 +194,208 @@ func (ir *Runtime) SaveImages(ctx context.Context, namesOrIDs []string, format s
...
-		src, err := is.Transport.NewStoreReference(img.imageruntime.store, nil, id)
-		if err != nil {
-			return errors.Wrapf(err, "error getting source imageReference for %q", img.InputName)
+	return nil
+}
+
+// LoadAllImagesFromOCIArchive loads all images from the docker archive that
+// fileName points to.
+func (ir *Runtime) LoadAllImagesFromOCIArchive(ctx context.Context, fileName string, signaturePolicyPath string, writer io.Writer) ([]string, error) {
+	sc := GetSystemContext(signaturePolicyPath, "", false)
```
mtrmac, line 357: `LoadAllImagesFromDockerArchive` has an `if signaturePolicyPath == ""` check. Which one is right?

libpod/image/image.go
```diff
@@ -210,82 +194,208 @@ func (ir *Runtime) SaveImages(ctx context.Context, namesOrIDs []string, format s
...
+				if err != nil {
+					return err
+				}
+			}
+			_, err = cp.Image(ctx, policyContext, dest, src, copyOptions)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
mtrmac, line 349:
```go
default:
    return errors.New("Internal error: supposedly unreachable code reached with format %q", …)
```
or something like that?

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
 
 // LoadImage loads a container image into local storage
 func (r *Runtime) LoadImage(ctx context.Context, inputFile string, writer io.Writer, signaturePolicy string) (string, error) {
-	if newImages, err := r.LoadAllImageFromArchive(ctx, writer, inputFile, signaturePolicy); err == nil {
+	format, err := imageFormat(inputFile)
+	if err != nil {
+		return "", err
+	}
```
mtrmac, line 284: What if format can not be determined?

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+	if src != nil {
+		newImages, err := r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
+		if err != nil {
+			return "", err
+		}
 		return getImageNames(newImages), nil
 	}
-	return "", err
+
+	return "", errors.Wrapf(err, "error pulling image")
```
mtrmac, line 338: This is incorrect if `src == nil` (both “couldn’t guess” and “coding error”)

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+		if err != nil {
+			return "", err
+		}
+		return strings.Join(newImageNames, ", "), nil
+	}
+	return "", errors.Errorf("error loading multipal images from the archive")
+}
+
+// LoadImageFromSingleImageDir loads image from the archive of single image that inputFile points to.
+func (r *Runtime) LoadImageFromSingleImageDir(ctx context.Context, writer io.Writer, inputFile, signaturePolicy, format string) (string, error) {
```
mtrmac, line 315: This can be package-private AFAICS.

libpod/runtime_img.go
```diff
@@ -276,46 +278,112 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+		}
+		return strings.Join(newImageNames, ", "), nil
+	case image.OCIArchive:
+		newImageNames, err := r.ImageRuntime().LoadAllImagesFromOCIArchive(ctx, inputFile, signaturePolicy, writer)
+		if err != nil {
+			return "", err
+		}
+		return strings.Join(newImageNames, ", "), nil
+	}
+	return "", errors.Errorf("error loading multipal images from the archive")
```
mtrmac, line 311: This does not say much about why things failed or what the user should do.

libpod/image/image.go
```diff
@@ -329,16 +439,8 @@ func (ir *Runtime) LoadAllImagesFromDockerArchive(ctx context.Context, fileName
...
 	}
 
-	newImages := make([]*Image, 0, len(imageNames))
-	for _, name := range imageNames {
-		newImage, err := ir.NewFromLocal(name)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error retrieving local image after pulling %s", name)
-		}
-		newImages = append(newImages, newImage)
-	}
```
mtrmac, line 339: Nice. 

Absolutely non-blocking: If I’m checking correctly, doing the same to `LoadFromArchiveReference` would allow most of the `Load` functions involved in this PR to return `[]string`, and two of the callers of `Runtime.LoadImage` would eliminate a `strings.Split` (and the third one would now contain the only `strings.Join`).

mtrmac (commented):
I’m still rather unhappy about the length and duplication in `Runtime.SaveImages`, see earlier.

As mentioned a few times, if the guessing can fail, this still needs to try the relevant alternatives. Maybe the guessing could be a separate performance feature PR, I’m not sure whether that would help or hinder.

mtrmac:
> @mtrmac PTAL, Added an archivehelper to decide the format according to the manifest.json to index.json, it this a way to detect the format?

(The detection is definitely useful, but not something I think _must_ be in this PR; originally I only asked for not throwing away errors from all the attempts. If detection can fail and we still might end up guessing, the “not throwing away errors” part will continue to be necessary.)

zhangguanzhang:
need gofmt the files

QiWang19:
@mtrmac Dropped the format detection from this PR. And the collections of the errors from `LoadAllImageFromArchive` and `LoadImageFromSingleImageArchive` using the `appendErr`, Could you take a look?

rhatdan:
@QiWang19 needs rebase.

zhangguanzhang:
still have `Conflicting files`

mtrmac:
> still have `Conflicting files`

This can not be merged as-is because it vendors an unmerged c/image branch, so that’s not an urgent concern _at this point_.

libpod/runtime_img.go
```diff
@@ -276,24 +277,51 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
 func (r *Runtime) LoadImageFromSingleImageArchive(ctx context.Context, writer io.Writer, inputFile, signaturePolicy string) (string, error) {
-	var err error
+	var (
+		src       types.ImageReference
+		err       error
+		latestErr error
+	)
+	appendErr := func(e error) {
+		if latestErr != nil {
+			latestErr = errors.Wrapf(latestErr, "error pulling images: %v\n", e)
```
mtrmac, line 320: This will prefix some errors but not all? If I am reading this correctly,
```
error pulling images: error pulling images: error1
error2
error3
```
---

`s/latestErr/allErrors/` maybe?

---

Most importantly, compare `pkg/errorhandling`, or maybe use it. I don’t know what are all of the coding conventions in Podman, but this does look like a problem appropriate for a common solution.

libpod/runtime_img.go
```diff
@@ -308,14 +336,17 @@ func (r *Runtime) LoadImageFromSingleImageArchive(ctx context.Context, writer io
 			return layout.NewReference(inputFile, "")
 		},
 	} {
-		src, err := referenceFn()
+		if src, err = referenceFn(); err != nil {
+			appendErr(err)
+		}
 		if err == nil && src != nil {
```
mtrmac, line 342: (This can now be an `else if` to be a tiny bit clearer.)

libpod/runtime_img.go
```diff
@@ -276,24 +277,51 @@ func DownloadFromFile(reader *os.File) (string, error) {
 
 // LoadImage loads a container image into local storage
 func (r *Runtime) LoadImage(ctx context.Context, inputFile string, writer io.Writer, signaturePolicy string) (string, error) {
-	if newImages, err := r.LoadAllImageFromArchive(ctx, writer, inputFile, signaturePolicy); err == nil {
+	newImages, err := r.LoadAllImageFromArchive(ctx, writer, inputFile, signaturePolicy)
+	if err == nil {
 		return newImages, nil
 	}
+	logrus.Warnf("Error trying pulling multi-arch images: %v", err)
```
mtrmac, line 284: - s/multi-arch images/multi-image archives/
- For some formats like `dir`, `LoadAllImageFromArchive` will not even try, so it’s grating to output a warning and then succeed.
- This is the _only_ caller of `LoadAllImageFromArchive` as well as `LoadImageFromSingleImageArchive`, so it seems that there should only be one set of attempts, one list of errors (and the two archive formats should only be attempted once, not once via the multi-image reader and once via the non-reader path).

libpod/runtime_img.go
```diff
@@ -276,24 +277,51 @@ func DownloadFromFile(reader *os.File) (string, error) {
...
+			latestErr = errors.Wrapf(latestErr, "%v\n", e)
+		} else {
+			latestErr = e
+		}
+	}
+	newImageNames, err := r.ImageRuntime().LoadAllImagesFromDockerArchive(ctx, inputFile, signaturePolicy, writer)
+	if err == nil {
+		return strings.Join(newImageNames, ", "), nil
+	}
+	appendErr(errors.Wrapf(err, "error trying multi docker aechive"))
```
mtrmac, line 302: “archive”, and this probably could benefit from a bit of rewording (users don’t care much about the “multi” part) — I haven’t tried this to see what the eventual full error text looks like.

(In both cases).

libpod/image/image.go
```diff
@@ -244,48 +240,143 @@ func (ir *Runtime) SaveImages(ctx context.Context, namesOrIDs []string, format s
 			if isTagged {
 				iData.tags = append(iData.tags, refTagged)
 			}
+		} else {
+			iData.destNames = []string{""}
```
mtrmac, line 244: This shouldn’t throw away any previous tags. Maybe add a single `imageSpecifiedByID` boolean?

libpod/image/image.go
```diff
@@ -210,8 +194,19 @@ func (ir *Runtime) SaveImages(ctx context.Context, namesOrIDs []string, format s
...
+	}
+	defer func() {
+		if err := policyContext.Destroy(); err != nil {
+			logrus.Errorf("failed to destroy policy context: %q", err)
+		}
+	}()
+
 	// extend an image with additional tags
 	type imageData struct {
+		destNames []string
```
mtrmac, line 209: If “extend an image”, the `*Image` should probably remain first.

I’m not 100% happy with the two duplicate name arrays, but I also can’t see much of an alternative. (Moving the docker-archive tag parsing into the docker-archive specific loop would probably add too much complexity.)

libpod/image/image.go
```diff
@@ -244,48 +240,143 @@ func (ir *Runtime) SaveImages(ctx context.Context, namesOrIDs []string, format s
...
-		img := imageMap[id]
-		copyOptions := getCopyOptions(sys, writer, nil, nil, SigningOptions{RemoveSignatures: removeSignatures}, "", img.tags)
+		copyOptions := getCopyOptions(sys, writer, nil, nil, SigningOptions{RemoveSignatures: removeSignatures}, "", nil)
 		copyOptions.DestinationCtx.SystemRegistriesConfPath = registries.SystemRegistriesConfPath()
+		// copy the image one by one
+		for _, id := range imageQueue {
+			img := imageMap[id]
+			// For copying, we need a source reference that we can create
+			// from the image.
+			src, err := is.Transport.NewStoreReference(img.imageruntime.store, nil, id)
```
mtrmac, line 310: This lookup can AFAICS also happen in the shared `imageQueue` construction code.

mtrmac (commented):
ACK in broad strokes, mostly “local” suggestions.

libpod/image/image.go
QiWang19: The src,  `storageReference` is not exported. The imageQueue may not be able to hold a field for it.

QiWang19:
@mtrmac PTAL

vrothberg:
@QiWang19, do you still have time to tackle it? No worries in case not. I am going through some open PRs and issues at the moment.

QiWang19:
> @QiWang19, do you still have time to tackle it? No worries in case not. I am going through some open PRs and issues at the moment.

Yes, I will find time to rebase this one and cooperate it with https://github.com/containers/image/pull/1072.

vrothberg:
Closing the PR as it must now be addressed in c/common/libimage. Thanks, @QiWang19, for working on this!

PR #8201, opened 2020-10-30
Remove search limit since pagination support

QiWang19:
Remove the search limit check since the c/image v5.6.0 supports pagination and can give result over 100 entries.
The same change of code has been in the v2.0 branch. https://github.com/containers/podman/pull/7356
Signed-off-by: Qi Wang <qiwan@redhat.com>

PR #8220, opened 2020-11-02
Fix podman search tests

cevich:
Depends on #8201


baude:
we just have an outtage ... i think this is a good thing to search rh registries? what do others think?

cevich:
ya, probably more important issues for us to focus on.

PR #8225, opened 2020-11-02
Improve error messages from failing tests

debarshiray:
Using a function like ContainSubstring is better because if the test
fails it will log a descriptive error that includes the actual string
generated during the test. This is more helpful than a function like
BeTrue that will only indicate that an assertion failed without giving
further details of the failure.

Signed-off-by: Debarshi Ray <rishi@fedoraproject.org>

### DIFF
--- a/test/e2e/exec_test.go
+++ b/test/e2e/exec_test.go
@@ -87,14 +87,12 @@ var _ = Describe("Podman exec", func() {
 		session := podmanTest.Podman([]string{"exec", "--env", "FOO=BAR", "test1", "printenv", "FOO"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ := session.GrepString("BAR")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(Equal("BAR"))
 
 		session = podmanTest.Podman([]string{"exec", "--env", "PATH=/bin", "test1", "printenv", "PATH"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ = session.GrepString("/bin")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(Equal("/bin"))
 	})
 
 	It("podman exec os.Setenv env", func() {
@@ -107,8 +105,7 @@ var _ = Describe("Podman exec", func() {
 		session := podmanTest.Podman([]string{"exec", "--env", "FOO", "test1", "printenv", "FOO"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ := session.GrepString("BAR")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(Equal("BAR"))
 		os.Unsetenv("FOO")
 	})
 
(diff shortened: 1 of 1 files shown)


test/e2e/exec_test.go
```diff
@@ -87,14 +87,12 @@ var _ = Describe("Podman exec", func() {
 		session := podmanTest.Podman([]string{"exec", "--env", "FOO=BAR", "test1", "printenv", "FOO"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ := session.GrepString("BAR")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(ContainSubstring("BAR"))
```
rhatdan, line 90: It is better to have
 		Expect(session.OutputToString()).To(Equal("BAR"))
That way we don't match FOOBAR

test/e2e/exec_test.go
```diff
@@ -87,14 +87,12 @@ var _ = Describe("Podman exec", func() {
...
 
 		session = podmanTest.Podman([]string{"exec", "--env", "PATH=/bin", "test1", "printenv", "PATH"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ = session.GrepString("/bin")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(ContainSubstring("/bin"))
```
rhatdan, line 95: Same here.

test/e2e/exec_test.go
```diff
@@ -107,8 +105,7 @@ var _ = Describe("Podman exec", func() {
 		session := podmanTest.Podman([]string{"exec", "--env", "FOO", "test1", "printenv", "FOO"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ := session.GrepString("BAR")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(ContainSubstring("BAR"))
```
rhatdan, line 108: Same

test/e2e/exec_test.go
```diff
@@ -199,14 +195,12 @@ var _ = Describe("Podman exec", func() {
 		session := podmanTest.Podman([]string{"exec", "--workdir", "/tmp", "test1", "pwd"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		match, _ := session.GrepString("/tmp")
-		Expect(match).Should(BeTrue())
+		Expect(session.OutputToString()).To(ContainSubstring("/tmp"))
```
rhatdan, line 198: Same

test/e2e/exec_test.go
```diff
@@ -280,7 +274,7 @@ var _ = Describe("Podman exec", func() {
 		exec := podmanTest.Podman([]string{"exec", "-ti", ctrName2, "id"})
 		exec.WaitWithDefaultTimeout()
 		Expect(exec.ExitCode()).To(Equal(0))
-		Expect(strings.Contains(exec.OutputToString(), fmt.Sprintf("%s(%s)", gid, groupName))).To(BeTrue())
+		Expect(exec.OutputToString()).To(ContainSubstring(fmt.Sprintf("%s(%s)", gid, groupName)))
```
rhatdan, line 277: Should this equal?

rhatdan (approved):

test/e2e/exec_test.go
debarshiray: Fixed.

test/e2e/exec_test.go
debarshiray: Fixed.

test/e2e/exec_test.go
debarshiray: The output of `id` would at least also have the UID, so we can't just use `Equal` without further changes to the test.

test/e2e/exec_test.go
debarshiray: Fixed.

debarshiray:
Thanks for the review!

test/e2e/exec_test.go
debarshiray: Fixed.

test/e2e/exec_test.go
rhatdan: I am fine with this, then.

rhatdan (approved):
LGTM

PR #8226, opened 2020-11-02
Test $HOME when it's parent is bind mounted with --userns=keep-id

debarshiray:
When --userns=keep-id is used, Podman is supposed to set up the home
directory of the user inside the container to match that on the host
as long as the home directory or any of its parents are marked as
volumes to be bind mounted into the container.

Currently, the test only considers the case where the home directory
itself is bind mounted into the container. It doesn't cover the Podman
code that walks through all the bind mounts looking for ancestors in
case the home directory itself wasn't specified as a bind mount.

Therefore, this improves the existing test added in commit
6ca8067956128585 ("Setup HOME environment when using --userns=keep-id")

Signed-off-by: Debarshi Ray <rishi@fedoraproject.org>

### DIFF
--- a/test/e2e/toolbox_test.go
+++ b/test/e2e/toolbox_test.go
@@ -30,10 +30,12 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path"
 	"strconv"
 	"strings"
 	"syscall"
 
+	"github.com/containers/podman/v2/pkg/rootless"
 	. "github.com/containers/podman/v2/test/utils"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/e2e/toolbox_test.go
```diff
@@ -371,10 +372,21 @@ var _ = Describe("Toolbox-specific testing", func() {
...
 		Expect(session.OutputToString()).To(ContainSubstring(currentUser.HomeDir))
+
+		location := path.Dir(currentUser.HomeDir)
+		volumeArg := fmt.Sprintf("%s:%s", location, location)
+		session = podmanTest.Podman([]string{"run",
+			"--userns=keep-id",
+			"--volume", volumeArg,
+			fedoraToolbox, "sh", "-c", "echo $HOME"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
TomSweeneyRedHat, line 388: This line looks to be failing in the CI....

debarshiray: Yes, the test doesn't seem to like being run as root.

giuseppe (approved):
LGTM

vrothberg:
Thanks, @debarshiray !

giuseppe:
> /lgtm cancel

was it supposed to be /hold cancel? :-)

vrothberg:
> > /lgtm cancel
> 
> was it supposed to be /hold cancel? :-)

I really need to grab a coffee

giuseppe:
> I really need to grab a coffee

double please :-D

/hold cancel

debarshiray:
Thanks for the reviews!

Issue #8001, opened 2020-10-13
podman run --override-arch do not download the image with correct architecture if the image with other architecture is exists.


lisongmin:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`docker run --override-arch` do not download the image with correct architecture if the image with other architecture is exists.

**Steps to reproduce the issue:**

1. enable multiarch support on x86_64

```bash
$ sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
```

2. ensure golang image is removed

```bash
$ podman image rm golang
```

3. create a container of golang with native architecture. We can see that image is x86_64, which is expected.

```bash
$ podman run -ti --rm golang ldd /bin/ls
Trying to pull docker.io/library/golang...
Getting image source signatures
Copying blob fb320882041b skipped: already exists
Copying blob 71e126169501 skipped: already exists
Copying blob 03f1c9932170 skipped: already exists
Copying blob 57df1a1f1ad8 skipped: already exists
Copying blob 1af28a55c3f3 skipped: already exists
Copying blob f4773b341423 skipped: already exists
Copying blob 24b0ad6f9416 [--------------------------------------] 0.0b / 0.0b
Copying config 05c8f6d253 done
Writing manifest to image destination
Storing signatures
        linux-vdso.so.1 (0x00007ffe6ef2b000)
        libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x00007fac06d25000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fac06b64000)
        libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007fac06af0000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fac06aeb000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fac06f77000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fac06aca000)
```

4. run another container with `--override-arch=arm64`. The arch is x86_64, which is **unexpected**

```bash
$ podman run -ti --rm --override-arch arm64 golang ldd /bin/ls
        linux-vdso.so.1 (0x00007fff2c9cb000)
        libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x00007f040364c000)
[...]

PR #8228, opened 2020-11-03
pull: account for platform

vrothberg:
When pulling an image, account for the specified OS and architecture
when looking up local images.  While a local image may be found based
on the specified name, the platform may be different from what the
user desires.  In that case, do not use the local image but continue
pulling.

Also remove pull-policy logic from the client.  That'll reduce one
roundtrip for the remote client and reduces code scattering.  The
backend should be the single source of truth for pull-policy handling.

Fixes: #8001

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -142,6 +142,25 @@ func (ir *Runtime) NewFromLocal(name string) (*Image, error) {
 	return ir.newImage(updatedInputName, localImage), nil
 }
 
+// matchesPlatform returns true if the image matches the specified os, arch and
+// variant.  Note that os, arch and variant are ignored if they are empty.
+func (i *Image) matchesPlatform(ctx context.Context, os, arch, variant string) (bool, error) {
+	data, err := i.inspect(ctx, false)
+	if err != nil {
+		return false, errors.Wrap(err, "error determining image's platform")
+	}
+	if os != "" && data.Os != os {
+		return false, nil
+	}
+	if arch != "" && data.Architecture != arch {
+		return false, nil
+	}
+	if variant != "" && data.Variant != variant {
+		return false, nil
+	}
+	return true, nil
+}
+
 // New creates a new image object where the image could be local
 // or remote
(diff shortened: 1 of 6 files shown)


vrothberg:

@nalind, @TomSweeneyRedHat ... that's fixing the issue we chatted about yesterday

nalind:
The new `Image.matchesPlatform()` seems a bit simple compared to the current version of similar logic in the image library (short version: ARM is complicated).  Does it produce the correct results on arm64 systems?

vrothberg:
> The new `Image.matchesPlatform()` seems a bit simple compared to the current version of similar logic in the image library (short version: ARM is complicated). Does it produce the correct results on arm64 systems?

Good catch. I forgot about "variant"! Is that what you were referring to?

nalind:
> > The new `Image.matchesPlatform()` seems a bit simple compared to the current version of similar logic in the image library (short version: ARM is complicated). Does it produce the correct results on arm64 systems?
> 
> Good catch. I forgot about "variant"! Is that what you were referring to?

More vaguely than I should have, but yeah, variant.

vrothberg:
Curious. Jobs didn't get queued for three hours.

rhatdan:
@vrothberg Could you rebase this and see if it passes tests?

libpod/image/image.go
```diff
@@ -142,6 +142,25 @@ func (ir *Runtime) NewFromLocal(name string) (*Image, error) {
...
+	if err != nil {
+		return false, errors.Wrap(err, "error determining image's platform")
+	}
+	if os != "" && data.Os != os {
+		return false, nil
+	}
+	if arch != "" && data.Architecture != arch {
+		return false, nil
+	}
+	if variant != "" && data.Variant != arch {
```
nalind, line 158: Comparing `data.Variant` with `arch` instead of `variant` looks wrong.

vrothberg: Thanks for catching!

vrothberg:
I am currently focusing on the copy work. I will go back to this PR after.

vrothberg:
@rhatdan was faster to fix it :)

PR #8230, opened 2020-11-03
Ensure that --net=host/pod/container conflicts with -p

mheon:
Setting port mappings only works when CNI is configuring our network (or slirp4netns, in the rootless case). This is not the case with `--net=host`, `--net=container:`, and joining the network namespace of the pod we are part of. Instead of allowing users to do these things and then be confused why they do nothing, let's actually return an error.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -522,4 +522,9 @@ json-file | f
     run_podman untag $IMAGE $newtag $newtag2
 }
 
+@test "podman run with --net=host and --port prints warning" {
+    run_podman run -d --rm -p 8080 --net=host $IMAGE ls > /dev/null
+    is "$output" ".*Port mappings have been discarded as one of the Host, Container, Pod, and None network modes are in use"
+}
+
 # vim: filetype=sh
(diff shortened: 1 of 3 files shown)


mheon:
Realization: we should block `--net=none` too.

mheon:
Apparently we were doing this in our tests, with the registry server

mheon:
Re-pushed and dropped to warning. This matches Docker, which lets you start the container but prints a warning that your container didn't actually make any port mappings.

mheon:
I think I need to get a system test written for this

mheon:
Test added, good to merge @containers/podman-maintainers

rhatdan:
LGTM

test/system/030-run.bats
```diff
@@ -522,4 +522,9 @@ json-file | f
     run_podman untag $IMAGE $newtag $newtag2
 }
 
+@test "podman run with --net=host and --port prints warning" {
+    run_podman run -d --rm -p 8080 --net=host $IMAGE ls > /dev/null
```
edsantiago, line 526: The `>/dev/null` here is a NOP. Not worth resubmitting for, though. Just an FYI for next time. (All output is captured in `$output`)

PR #8231, opened 2020-11-03
fedora rootless cpu settings

baude:
fedora does not have the the ability in rootless to set cpu limits.
this requires a simple fix for fedora 33 to pass ci tests.


vrothberg (commented):
LGTM

PR #8236, opened 2020-11-03
Update CI tests to run python docker library against API

jwhonce:
* Update reference to docker-py to docker to reflect change in library
  name
* Update tests to create storage sandbox
* Enable all tests that endpoints support
* Refactor containers/{id}/rename to return 404 not 500
* Refactor tests to use quay.io vs. docker.io


jwhonce:
@baude To reproduce issue:

```shell
DEBUG=1 PODMAN=./bin/podman python3 -m unittest -v \
test.python.docker.test_containers.TestContainers.test_inspect_container 2>&1 |tee /tmp/errors
```

jwhonce:
@cevich It looks like we need a `dnf install python3-docker` on the images

cevich:
> It looks like we need a dnf install python3-docker on the images

This is not normally a problem...just that right now I've got 6 browser tabs open, tracking major image-updates related for Ubuntu and and Fedora across three repositories :scream: 

Hmmm, for the short-term I think it's okay to `dnf install` it right before CI runs your tests, along with a comment to remove it once `python3-docker` is in the VM images.  That'll reduce the short-term testing burden on me, get you the package, though at the cost of some medium/long-term instability.  However, I intend to take care of that instability w/in the coming weeks (I'll add a jira card).

cevich:
https://github.com/containers/automation_images/pull/36

Makefile
```diff
@@ -356,8 +356,11 @@ remotesystem:
 
 .PHONY: localapiv2
 localapiv2:
+	# TODO Remove once VM's with dependency
+	./hack/install_dependencies.sh python3-docker
```
cevich, line 360: I'm not in favor of platform-specific calls like this in the general-use `Makefile`.  Since the problem is specific to CI, and short-term, would you mind doing something like this instead:

```diff
diff --git a/contrib/cirrus/runner.sh b/contrib/cirrus/runner.sh
index 084b196a9..b7e7ab852 100755
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -63,6 +63,12 @@ function _run_unit() {
 }

 function _run_apiv2() {
+    # TODO Remove once VM's with dependency
+    if [[ "$OS_RELEASE_ID" == "fedora" ]]; then
+        dnf install -y python3-docker
+    else
+        apt-get -qq -y install python3-docker
+    fi
     make localapiv2 |& logformatter
 }

```

jwhonce: @cevich Happy to make the changes. Do rhel and centos use "$OS_RELEASE_ID" == "fedora" ?

cevich: They don't but...we don't run RHEL or CentOS in Cirrus-CI, only Ubuntu and Fedora :grinning: 

In fact (IIRC) this particular job only ever runs on Fedora.

hack/install_dependencies.sh
```diff
@@ -0,0 +1,18 @@
...
+}
+
+for rpm in "$@"; do
+    if rpm -q "$rpm" >/dev/null; then
+        continue
+    fi
+
+    sudo dnf install -y "$rpm"
+    rpm -V "$rpm"
+done
```
cevich, line 18: Dead-code addition?

mheon:
I can't ack the python bits, but the rest LGTM

PR #8237, opened 2020-11-03
Add tip re. typical root cause of "Exec format error" to troubleshooting.md

vorburger:
I've ran into an `exec container process '/bin/sh': Exec format error` and it took me a moment to figure out the root cause, so I thought perhaps mentioning this in the troubleshooting doc of this project could be useful to others in the future.

Background in https://github.com/openMF/community-app/issues/3323 due to https://github.com/timbru31/docker-ruby-node/issues/564.

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -680,3 +680,9 @@ file `/etc/systemd/system/user@.service.d/delegate.conf` with the contents:
 
 After logging out and loggin back in, you should have permission to set CPU
 limits.
+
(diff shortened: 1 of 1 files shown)


vorburger:
> @vorburger vorburger force-pushed the vorburger:patch-1 branch from cd27e90 to 99df7c9 now

added missing `Signed-off-by: ` footer.

jwhonce:
LGTM

PR #8238, opened 2020-11-03
Use regex for "pod ps" name filter to match "ps" behavior

joelsmith:
`podman ps --filter name=containername` supports specifying a regular expression as the container name. This provides flexibility to match exact names using `^` and `$` to anchor the string. For example, a filter of `name=hello` would match containers `hello` and `othello`, but a filter of `name=^hello$` will only match a container named `hello`. This is a good and useful feature.

By contrast, `podman pod ps --filter name=podname` does not use regular expressions. Instead, it matches any pod name which contains the filter string.  As a result, there is no way to filter on `hello` without also matching `othello`.

This patch makes the behavior consistent across `podman ps` and `podman pod ps`.

### DIFF
--- a/libpod/filters/pods.go
+++ b/libpod/filters/pods.go
@@ -1,6 +1,7 @@
 package lpfilters
 
 import (
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -78,7 +79,11 @@ func GeneratePodFilterFunc(filter, filterValue string) (
 		}, nil
 	case "name":
 		return func(p *libpod.Pod) bool {
-			return strings.Contains(p.Name(), filterValue)
+			match, err := regexp.MatchString(filterValue, p.Name())
+			if err != nil {
+				return false
+			}
+			return match
 		}, nil
 	case "status":
 		if !util.StringInSlice(filterValue, []string{"stopped", "running", "paused", "exited", "dead", "created"}) {
--- a/test/e2e/pod_ps_test.go
+++ b/test/e2e/pod_ps_test.go
@@ -107,6 +107,28 @@ var _ = Describe("Podman ps", func() {
 		Expect(result.ExitCode()).To(Equal(0))
 	})
 
+	It("podman pod ps filter name regexp", func() {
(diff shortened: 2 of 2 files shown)


rhatdan:
Thanks @joelsmith 
LGTM
/approve

mheon:
To confirm - `podman ps` also uses this regex-based matching for names? Wasn't aware we'd added that. Huh.

joelsmith:
> To confirm - `podman ps` also uses this regex-based matching for names? Wasn't aware we'd added that. Huh.

Yes. This patch basically copies the code at https://github.com/containers/podman/blob/master/libpod/filters/containers.go#L41

TomSweeneyRedHat:
LGTM
assuming happy tests

libpod/filters/pods.go
```diff
@@ -78,7 +79,11 @@ func GeneratePodFilterFunc(filter, filterValue string) (
 		}, nil
 	case "name":
 		return func(p *libpod.Pod) bool {
-			return strings.Contains(p.Name(), filterValue)
+			match, err := regexp.MatchString(filterValue, p.Name())
+			if err != nil {
```
saschagrunert, line 83: I think we either return the error here or we at least warn log to the user that the provided regex does not compile. The new behavior probably needs documentation in one of the man pages, too.

umohnani8: +1 to log a warning here

umohnani8:
changes LGTM

libpod/filters/pods.go
TomSweeneyRedHat: @joelsmith could you update by adding a warning please and we'll move this along?

joelsmith: Sure. Do you also want me to add a warning in the `podman ps` code path that I copied from? ( https://github.com/containers/podman/blob/master/libpod/filters/containers.go#L41 )

TomSweeneyRedHat: @joelsmith  WHAT? Do you want consistency across the code?   Actually, that would be awesome, TYVM!  ;^)

Issue #7852, opened 2020-09-30
run files under /tmp can be clobbered by systemd.tmpfiles


industrial-sloth:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
I have several long-running, little-used podman containers running under Centos 7.8. These are running rootless, and are launched by systemd with `User=` flags. The systemd unit files were created by podman, with `User=`, `Group=` and `WorkingDirectory=` flags manually added.

After some period of time (which I now believe to be approximately 10 days), some of these containers would go into a "zombie" state, where they no longer appeared under `podman ps`, but did appear in a "created" state under `podman ps --all`. Under some circumstances the containers would continue running, seemingly fine other than not showing up in `podman ps`; in other cases the container networking would die.

The important factor appears to be that the `runRoot` is under `/tmp`. Stock Centos runs `systemd-tmpfiles` daily, and by default, files under `/tmp` that have not been created, accessed, or modified in the past 10 days get blown away. 

**Steps to reproduce the issue:**
1. launch a container with runfiles under `/tmp`. The container should be idle, without doing anything that would lead to access or modification times being updated for its runfiles. Verify that it shows up under `podman ps`.

2. modify `/usr/lib/tmpfiles.d/tmp.conf` so that the `/tmp` cleanup line reads (for instance) `v /tmp 1777 root root 1m` rather than the default of `v /tmp 1777 root root 10d`. Wait a minute, then run `sudo /usr/bin/systemd-tmpfiles --clean`. Alternatively, do nothing besides waiting 11 days, and this will happen automatically. :)

3. rerun `podman ps`. 

**Describe the results you received**
Container will no longer show up under `podman ps`; the underlying processes appear to still be running, but networking, etc will be hosed. The particular behavior seen here seems to depend on whether anything has happened to cause access or modification times to be updated for files under `/tmp`, as recent file accesses/modifications keep `systemd-tmpfiles` from blowing those files away.

**Describe the results you expected**
Container stays happy.

**Output of `podman version`:**
```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15
Built:        Mon Sep 28 02:07:17 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**
```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.21-1.el7.x86_64
[...]

PR #8241, opened 2020-11-04
Use /tmp/podman-run-* for backup XDG_RUNTIME_DIR

rhatdan:
We need to block systemd from cleaning up this directory
by dropping a /usr/lib/tmpfiles.d/podman.conf file in place.

Fixes: https://github.com/containers/podman/issues/7852

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -493,6 +493,8 @@ install.bin-nobuild:
 	install ${SELINUXOPT} -d -m 755 $(DESTDIR)$(BINDIR)
 	install ${SELINUXOPT} -m 755 bin/podman $(DESTDIR)$(BINDIR)/podman
 	test -z "${SELINUXOPT}" || chcon --verbose --reference=$(DESTDIR)$(BINDIR)/podman bin/podman
+	install ${SELINUXOPT} -m 755 -d ${DESTDIR}${TMPFILESDIR}
+	install ${SELINUXOPT} -m 644 contrib/tmpfile/podman.conf ${DESTDIR}${TMPFILESDIR}/podman.conf
 
 .PHONY: install.bin
 install.bin: podman install.bin-nobuild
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -493,6 +493,8 @@ install.bin-nobuild:
 	install ${SELINUXOPT} -d -m 755 $(DESTDIR)$(BINDIR)
 	install ${SELINUXOPT} -m 755 bin/podman $(DESTDIR)$(BINDIR)/podman
 	test -z "${SELINUXOPT}" || chcon --verbose --reference=$(DESTDIR)$(BINDIR)/podman bin/podman
+	install ${SELINUXOPT} -m 755 -d ${DESTDIR}${TMPFILESDIR}
```
TomSweeneyRedHat, line 496: For historical purposes, could you add a comment here explaining this?

TomSweeneyRedHat:
Weird test error, and feels a bit kludgey.

rhatdan:
@TomSweeneyRedHat What feels kludgy?

Makefile
rhatdan: You want a comment in the man page explaining why we are installing podman.conf, or do you want a comment in podman.conf telling you what the line in their is for?  Telling systemd to not prune /tmp/podman-run-*?

TomSweeneyRedHat: I was thinking of just a comment here explaining the change with a pointer to the issue.  If you think a man page change would serve better, and/or a comment in `/tmp/podman-run-*` I'm fine with either or both.  Just so that two years from now someone doesn't remove it without realizing it's need.

TomSweeneyRedHat:
Just dropping a file in a dir to keep it from being deleted feels kludgy to me.  Yeah, not our code doing that and I don't have a better suggestion.

Makefile
rhatdan: I am just thinking the message should go into the podman.conf file.  Not here.  If I am removing content, I am likely to read the file.  Added explanation to podman.conf file.

contrib/tmpfile/podman.conf
```diff
@@ -0,0 +1,4 @@
+# /tmp/podman-run-* directory can contain content for Podman containers that have run
+# for many days. This following line prevents systemd from removing this content.
+x /tmp/podman-run-.*
+d /run/podman 0700 root root
```
TomSweeneyRedHat, line 4: :+1:

TomSweeneyRedHat:
LGTM
assuming happy tests which aren't too hip atm

ashley-cui (commented):
LGTM

PR #8245, opened 2020-11-04
rootless container creation settings

baude:
when running container creation as rootless on the compatibility layer, we need to make sure settings are not being done for memory and memory swappiness

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -7,7 +7,9 @@ import (
 	"strings"
 
 	"github.com/containers/podman/v2/pkg/api/handlers"
+	"github.com/containers/podman/v2/pkg/cgroups"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/rootless"
 	"github.com/containers/podman/v2/pkg/specgen"
 )
 
@@ -129,7 +131,7 @@ func stringMaptoArray(m map[string]string) []string {
 
 // ContainerCreateToContainerCLIOpts converts a compat input struct to cliopts so it can be converted to
 // a specgen spec.
-func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig) (*ContainerCLIOpts, []string, error) {
+func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroupsManager string) (*ContainerCLIOpts, []string, error) {
 	var (
 		capAdd     []string
 		cappDrop   []string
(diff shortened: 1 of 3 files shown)


baude:
this PR is in support of the tests that @jwhonce is readying for submission which is why i didnt add any.

baude:
@jwhonce this is for you

cmd/podman/common/create_opts.go
```diff
@@ -377,7 +387,7 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig) (*Cont
 		cliOpts.MemoryReservation = strconv.Itoa(int(cc.HostConfig.MemoryReservation))
 	}
 
-	if cc.HostConfig.MemorySwap > 0 {
+	if cc.HostConfig.MemorySwap > 0 && !rootless.IsRootless() {
```
rhatdan, line 390: These are valid for rootless, if the cgroupv2 is configured correctly.  @giuseppe WDYT?

cmd/podman/common/create_opts.go
```diff
@@ -346,16 +347,25 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig) (*Cont
...
+		User:             cc.Config.User,
+		UserNS:           string(cc.HostConfig.UsernsMode),
+		UTS:              string(cc.HostConfig.UTSMode),
+		Mount:            mounts,
+		Volume:           volumes,
+		VolumesFrom:      cc.HostConfig.VolumesFrom,
+		Workdir:          cc.Config.WorkingDir,
+		Net:              &netInfo,
+	}
+	if !rootless.IsRootless() {
```
mheon, line 359: All of this should probably be happening in `pkg/specgen/generate/validate.go` in the resource validation block there.

baude: i might agree that the defaults part could ... but not the actual add

cmd/podman/common/create_opts.go
baude: ok, lets change the conditional then.

cmd/podman/common/create_opts.go
mheon: No, we should be discarding even user-set values there, and adding warnings.

cmd/podman/common/create_opts.go
```diff
@@ -401,7 +414,7 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig) (*Cont
 		cliOpts.Restart = policy
 	}
 
-	if cc.HostConfig.MemorySwappiness != nil {
+	if cc.HostConfig.MemorySwappiness != nil && !rootless.IsRootless() {
```
giuseppe, line 417: this also needs to account for cgroupv2.

The condition should be: `rootless.IsRootless() && cgroupsv2 && cgroupDriver == "systemd'`

cmd/podman/common/specgen.go
```diff
@@ -168,7 +169,7 @@ func getMemoryLimits(s *specgen.SpecGenerator, c *ContainerCLIOpts) (*specs.Linu
 		memory.Kernel = &mk
 		hasLimits = true
 	}
-	if c.MemorySwappiness >= 0 {
+	if c.MemorySwappiness >= 0 && !rootless.IsRootless() {
```
mheon, line 172: Are we sure this is correct? Is this allowed on cgroupsv2? @giuseppe

pkg/specgen/container_validate.go
```diff
@@ -145,6 +146,15 @@ func (s *SpecGenerator) Validate() error {
 	//	return errors.New("unrecognized option for cgroups; supported are 'default', 'disabled', 'no-conmon'")
 	//}
 
+	//set ulimits if not rootless
+	if len(s.ContainerResourceConfig.Rlimits) < 1 && !rootless.IsRootless() {
```
mheon, line 150: Is this correct? @rhatdan Don't we have separate logic for determining rlimits based on the maximum allowed?

baude: fixed

jwhonce:
LGTM

PR #8251, opened 2020-11-05
network aliases for container creation

baude:
podman can now support adding network aliases when running containers
(--network-alias).  It requires an updated dnsname plugin as well as an
updated ocicni to work properly.

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -105,6 +105,13 @@ func (r *Runtime) configureNetNS(ctr *Container, ctrNS ns.NetNS) ([]*cnitypes.Re
 	podName := getCNIPodName(ctr)
 
 	podNetwork := r.getPodNetwork(ctr.ID(), podName, ctrNS.Path(), ctr.config.Networks, ctr.config.PortMappings, requestedIP, requestedMAC)
+	aliases, err := ctr.runtime.state.GetAllNetworkAliases(ctr)
+	if err != nil {
+		return nil, err
+	}
+	if len(aliases) > 0 {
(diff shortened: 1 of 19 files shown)


pkg/specgen/specgen.go
```diff
@@ -328,6 +328,9 @@ type ContainerCgroupConfig struct {
 // ContainerNetworkConfig contains information on a container's network
 // configuration.
 type ContainerNetworkConfig struct {
+	// Aliases are a list of network-scoped aliases for container
+	// Optional
+	Aliases []string `json:"aliases"`
```
mheon, line 333: This should be map[string]string to properly support aliases via HTTP API (that allows them to be set per-network)

libpod/networking_linux.go
```diff
@@ -105,6 +105,9 @@ func (r *Runtime) configureNetNS(ctr *Container, ctrNS ns.NetNS) ([]*cnitypes.Re
 	podName := getCNIPodName(ctr)
 
 	podNetwork := r.getPodNetwork(ctr.ID(), podName, ctrNS.Path(), ctr.config.Networks, ctr.config.PortMappings, requestedIP, requestedMAC)
+	if aliases := ctr.config.NetworkAliases; len(aliases) > 0 {
```
mheon, line 108: No - we need to use `ctr.state.GetAllNetworkAliases(ctr)`

baude:
changing this to a WIP because we cannot proceed without a fix in OCICNI -> https://github.com/cri-o/ocicni/pull/81

rhatdan:
LGTM

baude:
merge me

PR #8252, opened 2020-11-05
migrate play kube to spec gen

baude:
we need to migrate play kube away from using the old container creation
method.  the new approach is specgen and this aligns play kube with
container creation in the rest of podman.

### DIFF
--- a/pkg/domain/infra/abi/play_test.go
+++ b/pkg/domain/infra/abi/play_test.go
@@ -6,33 +6,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
-
-var configMapList = []v1.ConfigMap{
-	{
-		TypeMeta: metav1.TypeMeta{
-			Kind: "ConfigMap",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "bar",
-		},
-		Data: map[string]string{
-			"myvar": "bar",
-		},
-	},
-	{
-		TypeMeta: metav1.TypeMeta{
-			Kind: "ConfigMap",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "foo",
-		},
-		Data: map[string]string{
-			"myvar": "foo",
-		},
-	},
-}
 
 func TestReadConfigMapFromFile(t *testing.T) {
 	tests := []struct {
@@ -55,11 +30,11 @@ data:
 			false,
 			"",
 			v1.ConfigMap{
-				TypeMeta: metav1.TypeMeta{
+				TypeMeta: v12.TypeMeta{
 					Kind:       "ConfigMap",
 					APIVersion: "v1",
 				},
-				ObjectMeta: metav1.ObjectMeta{
+				ObjectMeta: v12.ObjectMeta{
 					Name: "foo",
 				},
 				Data: map[string]string{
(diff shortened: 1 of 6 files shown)


pkg/specgen/generate/kube/kube.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func ToPodGen(ctx context.Context, podName string, podYAML *v1.PodTemplateSpec) (*specgen.PodSpecGenerator, error) {
+	p := specgen.NewPodSpecGenerator()
+	p.Name = podName
+	p.Labels = podYAML.ObjectMeta.Labels
+	// TODO we only configure Process namespace. We also need to account for Host{IPC,Network,PID}
```
mheon, line 23: For this, we'd remove the namespace from SharedNamespaces and then set it to host on all containers

pkg/specgen/generate/kube/kube.go
```diff
@@ -0,0 +1,311 @@
...
+	s.Pod = podID
+
+	setupSecurityContext(s, containerYAML)
+
+	// Since we prefix the container name with pod name to work-around the uniqueness requirement,
+	// the seccomp profile should reference the actual container name from the YAML
+	// but apply to the containers with the prefixed name
+	s.SeccompProfilePath = seccompPaths.FindForContainer(containerYAML.Name)
+
+	var err error
```
mheon, line 72: This is unnecessary - we have a `:=` on err two lines below

pkg/specgen/generate/kube/kube.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return nil, err
+	}
+	s.WorkDir = "/"
+	if imageData != nil && imageData.Config != nil {
+		if imageData.Config.WorkingDir != "" {
+			s.WorkDir = imageData.Config.WorkingDir
+		}
+		s.Command = imageData.Config.Entrypoint
+		s.Labels = imageData.Config.Labels
```
mheon, line 119: This bit, WorkingDir, and StopSignal are unnecessary - I know we need to rework Command and Entrypoint for Kube compat, but the rest should be handled by Specgen

pkg/specgen/generate/kube/kube.go
```diff
@@ -0,0 +1,311 @@
...
+		if !exists {
+			return nil, errors.Errorf("Volume mount %s specified for container but not configured in volumes", volume.Name)
+		}
+		if err := parse.ValidateVolumeCtrDir(volume.MountPath); err != nil {
+			return nil, errors.Wrapf(err, "error in parsing MountPath")
+		}
+		mount := spec.Mount{
+			Type:        "bind",
+			Destination: hostPath,
+			Source:      volume.MountPath,
```
mheon, line 175: I think that Options should be unconditionally set to `[]string{"bind"}` here

pkg/specgen/generate/kube/kube.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	if containerYAML.SecurityContext.Privileged != nil {
+		s.Privileged = *containerYAML.SecurityContext.Privileged
+	}
+
+	if containerYAML.SecurityContext.AllowPrivilegeEscalation != nil {
+		s.NoNewPrivileges = !*containerYAML.SecurityContext.AllowPrivilegeEscalation
+	}
+
+	if seopt := containerYAML.SecurityContext.SELinuxOptions; seopt != nil {
```
mheon, line 203: @rhatdan PTAL at this bit

pkg/specgen/generate/kube/kube.go
baude: it isnt, i tested

pkg/specgen/generate/kube/kube.go
baude: it does not appear to need to be

pkg/specgen/generate/kube/kube.go
baude: he confirmed this

baude:
looks like i have to move around the unit tests

pkg/specgen/generate/kube/kube.go
mheon: Well that sounds like a bug in SpecGen

rhatdan:
Fixes: https://github.com/containers/podman/issues/8211

pkg/specgen/generate/kube/kube.go
baude: given this is only a code migration, I am going to not fix todos in the code that existed prior.

baude:
merge me

mheon:
LGTM

mheon:
They really need to put the close and comment buttons a little further apart

pkg/domain/infra/abi/play_test.go
```diff
@@ -6,34 +6,9 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
haircommander, line 9: I'm not sure I love this change, why did you do it?

haircommander:
one note, not blocking, otherwise LGTM

pkg/domain/infra/abi/play_test.go
baude: whats the concern? the new alias?

haircommander: yeah! v12 doesn't mean very much to me, whereas metav1 (despite not having a ton of meaning) is a little more clear

PR #8260, opened 2020-11-06
Use CPP, CC and flags in dep check scripts

mbargull:
Similar to https://github.com/containers/buildah/pull/2751:

#### What this PR does / why we need it:

Allow build system without standard `cc` to successfully run the dependency checking helper scripts from the Makefile.
This supports custom compilers specified by the common `CC` environment variable, preprocessors given as `CPP` and additional preprocessor flags from `CPPFLAGS`.
Additionally, additional flags from `CFLAGS` and `LDFLAGS` are considered for compiling/linking.
Overall, this facilitates cross-compilation and similar setups (e.g., for building Conda packages at https://gitub.com/conda-forge).

#### How to verify it

Temporarily remove `cc` from `PATH` and point `CC`/`CPP` to a different compiler.
Install header files, e.g., `btrfs/ioctl.h`, in a custom location and set `CPPFLAGS=-I/custom/location`.
Add `-Wl,--as-needed` to `LDFLAGS` (which makes library order important (again) while linking) to confirm the order of the inputs `-` and `-ldevmapper` is correct in `libdm_tag.sh`.

### DIFF
--- a/hack/btrfs_installed_tag.sh
+++ b/hack/btrfs_installed_tag.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-cc -E - > /dev/null 2> /dev/null << EOF
+${CPP:-${CC:-cc} -E} ${CPPFLAGS} - > /dev/null 2> /dev/null << EOF
 #include <btrfs/ioctl.h>
 EOF
 if test $? -ne 0 ; then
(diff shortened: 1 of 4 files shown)


mbargull:
(Force pushed to include a description in the initial commit message.)

mbargull:
This comment for the second commit got lost due to the `rebase`/`push --force`:
> This changes vendored sources. I assume we don't want that, correct?
I only added it for completeness' sake, but made it a separate commit so it can be easily removed.

mbargull:
> This changes vendored sources. I assume we don't want that, correct?
> I only added it for completeness' sake, but made it a separate commit so it can be easily removed.

```
tree is dirty, please sync the vendor.conf and commit all changes.
```
Ah, perfect, you have CI tests for that ^^. I'll remove that commit.

mheon:
LGTM

QiWang19 (commented):
LGTM

Issue #7906, opened 2020-10-03
More flexible options when combining '--rm' and '--restart'


srcshelton:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

With `podman-2.1.1` with patch #7895 , running a container with `--rm` and `--restart on-failure` results in the output:

```
Error: the --rm option conflicts with --restart, when the restartPolicy is not "" and "no"
```

So a couple of things:

* Should there be a value within the first set if quotes?  Should the final 'and' read 'or'?

* Surely if should be possible to combine `--rm` and `--restart on-failure` (specifically) - the idea being that the container would be removed when externally killed by via `podman stop` (or equivalent), but would be restarted if the container's PID 1 aborts for any other internally-triggered reason (and/or results in a non-zero exit code)?

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14.7
Git Commit:   9f6d6ba0b314d86521b66183c9ce48eaa2da1de2
Built:        Sat Oct  3 00:59:40 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 35a2fa83022e56e18af7e6a865ba5d7165fa2a4a'
  cpus: 8
  distribution:
    distribution: gentoo
    version: unknown
  eventLogger: file
  hostname: dellr330
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.8.9-gentoo
  linkmode: dynamic
  memFree: 2429820928
  memTotal: 8063447040
  ociRuntime:
    name: crun
    package: Unknown
    path: /usr/bin/crun
    version: |-
      crun version 0.15-dirty
      commit: 56ca95e61639510c7dbd39ff512f80f626404969
      spec: 1.0.0
      +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 25387175936
  swapTotal: 25769787392
  uptime: 96h 19m 31.76s (Approximately 4.00 days)
registries:
  search:
[...]

PR #8263, opened 2020-11-06
Allow containers to --restart on-failure with --rm

rhatdan:
Fixes: https://github.com/containers/podman/issues/7906

### DIFF
--- a/cmd/podman/common/createparse.go
+++ b/cmd/podman/common/createparse.go
@@ -9,7 +9,7 @@ import (
 // by validate must not need any state information on the flag (i.e. changed)
 func (c *ContainerCLIOpts) validate() error {
 	var ()
-	if c.Rm && c.Restart != "" && c.Restart != "no" {
+	if c.Rm && (c.Restart != "" && c.Restart != "no" && c.Restart != "on-failure") {
 		return errors.Errorf(`the --rm option conflicts with --restart, when the restartPolicy is not "" and "no"`)
 	}
 
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -714,3 +714,17 @@ func (c *Container) Restore(ctx context.Context, options ContainerCheckpointOpti
 	defer c.newContainerEvent(events.Restore)
 	return c.restore(ctx, options)
 }
+
+// Indicate whether or not the container should restart
+func (c *Container) ShouldRestart(ctx context.Context) bool {
+	logrus.Debugf("Checking if container %s should restart", c.ID())
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return false
+		}
+	}
+	return c.shouldRestart()
+}
--- a/pkg/bindings/containers/containers.go
+++ b/pkg/bindings/containers/containers.go
@@ -390,3 +390,15 @@ func ContainerInit(ctx context.Context, nameOrID string) error {
 	}
 	return response.Process(nil)
 }
+
+func ShouldRestart(ctx context.Context, nameOrID string) (bool, error) {
+	conn, err := bindings.GetClient(ctx)
+	if err != nil {
+		return false, err
+	}
+	response, err := conn.DoRequest(nil, http.MethodPost, "/containers/%s/shouldrestart", nil, nil, nameOrID)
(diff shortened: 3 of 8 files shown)


libpod/container_api.go
```diff
@@ -714,3 +714,39 @@ func (c *Container) Restore(ctx context.Context, options ContainerCheckpointOpti
 	defer c.newContainerEvent(events.Restore)
 	return c.restore(ctx, options)
 }
+
+// Indicate whether or not the container should restart
+func (c *Container) ShouldRestart(ctx context.Context) bool {
```
mheon, line 719: Did you duplicate all of this logic? It looks like it was copied from container_internal

pkg/domain/infra/abi/containers.go
```diff
@@ -913,7 +913,7 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
 	} else {
 		report.ExitCode = int(ecode)
 	}
-	if opts.Rm {
+	if opts.Rm && !ctr.ShouldRestart(ctx) {
```
mheon, line 916: We also need to fix the cleanup process to not remove

libpod/container_api.go
rhatdan: It is, I will use the same function in both places.

pkg/domain/infra/abi/containers.go
rhatdan: I did but somehow it got lost from the patch.  It is back now.

cmd/podman/common/createparse.go
```diff
@@ -9,7 +9,7 @@ import (
 // by validate must not need any state information on the flag (i.e. changed)
 func (c *ContainerCLIOpts) validate() error {
 	var ()
-	if c.Rm && c.Restart != "" && c.Restart != "no" {
+	if c.Rm && (c.Restart != "" && c.Restart != "no" && c.Restart != "on-failure") {
```
TomSweeneyRedHat, line 12: It would be nice to have "on-failure" defined in a variable somewhere rather than a bare string scattered about.  Not for this pr though.

Luap99: It is defined in the `libpod` package as `RestartPolicyOnFailure`.
The same for `no` it should be `RestartPolicyNo`.
https://github.com/containers/podman/blob/dc58d4e2858bf6688a775277adf8f775afc30e73/libpod/container.go#L99

rhatdan:

baude:
lgtm

QiWang19 (commented):
LGTM

pkg/bindings/containers/containers.go
```diff
@@ -390,3 +390,15 @@ func ContainerInit(ctx context.Context, nameOrID string) error {
 	}
 	return response.Process(nil)
 }
+
+func ShouldRestart(ctx context.Context, nameOrID string) (bool, error) {
```
mheon, line 394: Why do we need this in the API and Bindings? This should be local-only, `podman container cleanup` should never be invoked remotely

rhatdan: In pkg/domain/infra/tunnel/containers.go when we do a non detached container, we Remove the container when it completes. Added this ShouldRestart check to make sure that this does not happen.

mheon: I don't know if it's worth adding a complete new API endpoint to handle this. Maybe we should just disable removing the container in that case if it detects that the container has this restart policy?

rhatdan: Then we end up with the race conditions again.  But for only a small subset of calls.

rhatdan: I removed the bindings and just stopped rm the container in run if the restart policy is set to something other then no.

cmd/podman/common/createparse.go
TomSweeneyRedHat: @rhatdan are you going to change per @Luap99 's suggestion?

TomSweeneyRedHat:
Tests aren't hip.

pkg/bindings/containers/containers.go
rhatdan: Removing the bindings proved to fail testing. So I added them back.

PR #8264, opened 2020-11-06
Show error on bad name filter in podman ps

joelsmith:
See comments in #8238 where it was decided that a message should be logged when a bad name regex is used for filtering for `podman ps` and `podman pod ps`.

### DIFF
--- a/libpod/filters/containers.go
+++ b/libpod/filters/containers.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/podman/v2/pkg/timetype"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


joelsmith:
Is this what people had in mind? I'm not very familiar with podman conventions on error logging, so please steer me in the right direction.

rhatdan:
LGTM
I would drop them down to Warnf, but right now we don't log warnings by default to the terminal.

TomSweeneyRedHat:
Changes LGTM, but tests aren't happy.   I don't think due to these changes and I've restarted.

rhatdan:
Thanks @joelsmith 
/approve
/lgtm

Issue #7942, opened 2020-10-06
podman run -t issues CRLF, making for confusing 'podman logs' output


edsantiago:
In window 1 (needed to differentiate podman-remote from podman, in logs):
```console
$ bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman run --name foo -t alpine sh -c 'echo 1;echo 2'
1
2
$ ./bin/podman logs foo|cat -vT
1^M       <--- CR
2^M
$ ./bin/podman-remote logs foo|cat -vT
1         <--- podman-remote emits those as LF; two lines for the price of one

2

$ for i in "" "-remote";do ./bin/podman$i logs foo|wc -l;done                                                                                         -
2
4
```
master @ defb754945b3f99c1d786dac95d9b17b24f55e59, but seems to be in v2.1 and possibly older as well. Root as well as rootless.

This is, I believe, causing [flakes in a podman logs test](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5425581623869440/html/int-remote-ubuntu-19-root-host.log.html#t--tail-800-lines--1)

zhangguanzhang:
same with https://github.com/containers/podman/issues/7196#issuecomment-669092535 https://github.com/containers/conmon/issues/195#issuecomment-669225956

edsantiago:
Indeed - thank you. This was a very hard problem to search for!

eiffel-fl:
I can try to investigate on this issue but can someone give me more information (if someone has some)?
Indeed, I did not understand if this problem was fixed by #7660 and where exactly this problem occurs (`podman-remote` or `conmon`).

zhangguanzhang:
> If this issue can be a good first one then I can try to investigate on this issue but can someone give me more information (if someone has some)?
> Indeed, I did not understand if this problem was fixed by #7660 and where exactly this problem occurs (`podman-remote` or `conmon`).
The root cause is this
```
[psakar@localhost podman]$ podman run --rm --tty busybox echo 1 | cat -A
1^M$
    [psakar@localhost podman]$ podman run --rm --tty busybox echo 1 ; echo 2 | cat -A
1
2$
[psakar@localhost podman]$ 
```

eiffel-fl:
Thank you, I do not promise to come back with a PR but I will investigate on this to try to narrow the root cause.

eiffel-fl:
Unfortunately, I do not come with a solution and maybe more problems...

[...]

PR #8272, opened 2020-11-08
Test podman-remote logs works the same as podman logs

rhatdan:
Fixes: https://github.com/containers/podman/issues/7942

### DIFF
--- a/test/e2e/logs_test.go
+++ b/test/e2e/logs_test.go
@@ -337,4 +337,22 @@ var _ = Describe("Podman logs", func() {
 		Expect(results).To(Exit(0))
 		Expect(results.OutputToString()).To(Equal("podman podman podman"))
 	})
+
+	It("Make sure logs match expected length", func() {
+		logc := podmanTest.Podman([]string{"run", "-t", "--name", "test", ALPINE, "sh", "-c", "echo 1; echo 2"})
+		logc.WaitWithDefaultTimeout()
+		Expect(logc).To(Exit(0))
+
+		wait := podmanTest.Podman([]string{"wait", "test"})
+		wait.WaitWithDefaultTimeout()
+		Expect(wait).To(Exit(0))
+
+		results := podmanTest.Podman([]string{"logs", "test"})
+		results.WaitWithDefaultTimeout()
+		Expect(results).To(Exit(0))
+		outlines := results.OutputToStringArray()
+		Expect(len(outlines)).To(Equal(2))
+		Expect(outlines[0]).To(Equal("1\r"))
+		Expect(outlines[1]).To(Equal("2\r"))
+	})
 })


mheon:
LGTM

giuseppe (approved):
LGTM

mheon:
LGTM, whenever you fix the unprintable characters issue

rhatdan:
@edsantiago @jwhonce fixed this issue with a different PR, but I added this test to make sure that we don't regress.

edsantiago:
I can't entirely be sure that this test is valid, because I don't know if Go `OutputToStringArray()` works the same way as tty output and/or shell pipeline would Did you test this against a pre-jhonce-fix build to make sure it fails? If not, I'll try to figure out what PR that was and will test on my end.

rhatdan:
@edsantiago yes this was failing badly before @jwhonce fix.  I wrote it with my fix, and then found that my fix was not good enough.  My original fix looked good from a human perspective, IE When you ran with a terminal.  But when you ran through this test it would blow up.  Digging deeper we saw that podman-remote was trimming control chars.

edsantiago:
I got your reply after I had already started testing; and I still think there's something missing. I don't think this test is actually testing what you think it's testing. I checked out 8f33751b5, which is the commit immediately prior to @jwhonce's #8282. I built podman and podman-remote, confirmed that `./bin/podman-remote logs foo`  gives me the (erroneous) multi-line output. Then I tried patching `test/e2e/logs_test.go` and running:
```console
$ ~/go/bin/ginkgo -trace -tags remote -debug -progress -r -focus "logs match expected length" test/e2e
...
Running: podman-remote [...] run -t --name test quay.io/libpod/alpine:latest sh -c echo 1; echo 2
output: 1 2
Running: podman-remote [...] wait test
output: 0
Running: podman-remote [...] logs test
output: 1 2
```
...and it passes. That is against what I would expect. I really think that the e2e framework, or Go, or something is silently stripping the `^M`s away, so this test will pass no matter what.

I would be very grateful if someone could confirm my findings by testing against a pre-#8282 commit, and either tell me where I went wrong, or (if I'm correct) suggest a way to make this test do what it claims to do.

Am placing hold on this because I really think it's a NOP test.

/hold

edsantiago:
Found it. The problem is that `OutputToStringArray()` splits on newline, so (of course) it's always going to return two array elements.

The following patch makes the test fail on 8f33751b5 (pre-8282) and pass on 1491b20f0 (post-8282):
```diff
diff --git a/test/e2e/logs_test.go b/test/e2e/logs_test.go
index db48e50ad..a749a86ff 100644
--- a/test/e2e/logs_test.go
+++ b/test/e2e/logs_test.go
@@ -350,6 +350,9 @@ var _ = Describe("Podman logs", func() {
                results := podmanTest.Podman([]string{"logs", "test"})
                results.WaitWithDefaultTimeout()
                Expect(results).To(Exit(0))
-               Expect(len(results.OutputToStringArray())).To(Equal(2))
+               outlines := results.OutputToStringArray()
+               Expect(len(outlines)).To(Equal(2))
+               Expect(outlines[0]).To(Equal("1\r"))
+               Expect(outlines[1]).To(Equal("2\r"))
        })
 })
```
I find it counterintuitive, because I was expecting the opposite (pre-8282 = `^M`, post-8282 = no `^M`) but I haven't looked closely enough at #8282. So please double- and triple-check my logic.

Issue #7241, opened 2020-08-05
podman-remote: there's a hang somewhere


edsantiago:
Another semi-useless report with no reproducer nor actual details.

There's a hang somewhere in podman-remote. It's causing flakes in CI all over the place. It does not seem to be deterministic, it's happening in lots of different situations.

Examples:
* From #7238 - looks like it's in podman wait - [log](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5867996646211584/html/system_test.log.html)
* From #7220 - in pause/unpause - [log](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6663941797445632/html/system_test.log.html)
* In post-merge testing - looks like wait again - [log](https://cirrus-ci.com/task/5826897567285248)

Sorry for sparse details, I'm trying very hard to be done for the week and am failing dismally. Please add examples here as you see them, I think this is going to be an ugly one.

edsantiago:
Another wait https://cirrus-ci.com/task/6571684087988224

edsantiago:
This script will reproduce it, although not quickly, you have to do it in a loop and wait possibly hours:
```console
$ cat >7241.sh <<EOF
set -e
./bin/podman-remote run -d --name foo alpine sh -c 'touch /foo;while test -e /foo; do sleep 1;done'
cid=foo
./bin/podman-remote logs $cid
./bin/podman-remote exec $cid /etc             || true
./bin/podman-remote exec $cid no-such-command  || true
./bin/podman-remote exec $cid rm /foo
timeout -v 15 ./bin/podman-remote wait $cid
./bin/podman-remote rm $cid
EOF

$ while bash -x 7241.sh;do echo;echo;done
...copious output... eventually dies
+ timeout -v 15 ./bin/podman-remote wait foo
timeout: sending signal TERM to command ‘./bin/podman-remote’

$ podman ps -a
CONTAINER ID  IMAGE                            COMMAND               CREATED         STATUS                     PORTS   NAMES
c87ff4b0836f  docker.io/library/alpine:latest  sh -c touch /foo;...  15 minutes ago  Exited (0) 10 minutes ago          foo
```
I find it interesting that CREATED was 15 minutes ago, but it exited 10. Implying that it took 5 minutes between the `exec rm /foo` and the time the container exited. 

I'm heading out for the day, but am leaving the container as it is in case anyone has suggestions for me to `inspect` or check logs or anything.

[...]

PR #8277, opened 2020-11-09
Reenable 'exec' test on remote

edsantiago:
It was 'skip'ped due to frequent flakes (#7241). I just tried
running the 7241 reproducer on my laptop for one hour, and
saw no failures, so let's reenable this in CI and see if it
comes back.

I really hate problems that "go away" on their own without
being explicitly acknowledged and fixed.

### DIFF
--- a/test/system/075-exec.bats
+++ b/test/system/075-exec.bats
@@ -6,8 +6,6 @@
 load helpers
 
 @test "podman exec - basic test" {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

edsantiago:
CI failures are real. Filed #8281.

edsantiago:
Abandoning; leaving it up to the #8281-fixer to enable this test.

PR #8284, opened 2020-11-09
system tests: skip journald tests on RHEL8

edsantiago:
(actually, on any system exhibiting the symptom wherein
journalctl --user fails due to insufficient permissions,
which for all practical purposes means only RHEL8)

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -436,6 +436,16 @@ json-file | f
 @test "podman run --log-driver journald" {
     skip_if_remote "We cannot read journalctl over remote."
 
+    # We can't use journald on RHEL as rootless, either: rhbz#1895105
+    if is_rootless; then
+        run journalctl -n 1
+        if [[ $status -ne 0 ]]; then
+            if [[ $output =~ permission ]]; then
+                skip "Cannot use rootless journald on this system"
+            fi
+        fi
+    fi
+
     msg=$(random_string 20)
     pidfile="${PODMAN_TMPDIR}/$(random_string 20)"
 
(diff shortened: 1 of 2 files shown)


edsantiago:
Tested by applying patch to podman-tests-2.1.1-3.module+el8.3.1+8686+2a59bca3.x86_64 on RHEL8. Results:
```console
$ bats /usr/share/podman/test/system/03*bats
 - podman run --log-driver journald (skipped: Cannot use rootless journald on this system)
 ...
 - podman logs over journald (skipped: Cannot use rootless journald on this system)
```
After adding testuser to `systemd-journal` group:
```console
 ✓ podman run --log-driver journald
 ...
 ✓ podman logs over journald
```
(i.e. they now run and pass).

rhatdan:
LGTM, I have a feeling centos8 will not work well without this patch either.

test/system/030-run.bats
```diff
@@ -436,6 +436,16 @@ json-file | f
...
 
+    # We can't use journald on RHEL as rootless, either: rhbz#1895105
+    if is_rootless; then
+        run journalctl -n 1
+        if [[ $status -ne 0 ]]; then
+            if [[ $output =~ permission ]]; then
+                skip "Cannot use rootless journald on this system"
+            fi
+        fi
+    fi
```
TomSweeneyRedHat, line 447: Would it be good to move this to a `skip_if_rootless_rhel` type of function?

TomSweeneyRedHat:
Changes LGTM, one question for consideration.

test/system/030-run.bats
edsantiago: Sorry for the too-late response. As much as I hate duplication, on two-offs I often just hold my nose and leave it. Should there ever be a need for a third exception, then yes, absolutely, refactor.

edsantiago:
We need to get this into RHEL8. Do I need to backport it into a maint branch, or will RHEL8 pick up a new build from here? @jnovy?

PR #8286, opened 2020-11-09
make network name uniq for dnsname tests

baude:
ed identified that the dnsname integration test does not use a unique
name and therefore cannot be cleaned up.  this was made worse by a
improper defer statement to remove the network should the test fail.


edsantiago:
LGTM

TomSweeneyRedHat:
LGTM
but the ubuntu-19 root host test failed.  I'll restart in case it's flakey.

Issue #8265, opened 2020-11-06
Podman stats does not show memory limits set per container


alosadagrande:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Wanted to limit the memory used by a container, however the memory limit it is not visualized in the stats command.

Podman is running on a F33 on a raspberry pi3 (aarch64)

**Steps to reproduce the issue:**

1. Run a container limiting the memory to use:

```sh
# podman run -d --rm --memory=64m busybox sleep 1000000000000000
ff622b233d98cc083529ba01e9f937e9483eebfc1ad52ffef35cbecde4b1ee93
```

2. Verify the stats

```sh
# podman stats

ID            NAME            CPU %   MEM USAGE / LIMIT  MEM %   NET IO          BLOCK IO           PIDS
7ccb9e39ee32  home-assistant  7.91%   219.3MB / 981.1MB  22.35%  -- / --         276.8MB / 39.99MB  92
ff622b233d98  angry_edison    --      1.069MB / 981.1MB  0.11%   628B / 1.646kB  757.8kB / 0B       1
```

**Describe the results you received:**

Podman stats shows the limit of the container to be the total amount of memory of the rpi3.

**Describe the results you expected:**

I would expect to see the limit set explicitly for the container. Basically, the same it shows in my laptop (Fedora x86_64)

**Additional information you deem important (e.g. issue happens only occasionally):**

This issue seems to show only in my ARM server.

**Output of `podman version`:**

```
# podman --version
podman version 2.1.1
```

**Output of `podman info --debug`:**

```
# podman info --debug
host:
  arch: arm64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.aarch64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 66d50d2f6c81e28bacd99ee8e8d222f028eef63f'
  cpus: 4
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.8.18-300.fc33.aarch64
  linkmode: dynamic
  memFree: 189890560
  memTotal: 981086208
  ociRuntime:
    name: crun
    package: crun-0.15-5.fc33.aarch64
    path: /usr/bin/crun
    version: |-
      crun version 0.15
[...]

PR #8290, opened 2020-11-10
use container cgroups path

vrothberg:
When looking up a container's cgroup path, parse /proc/[PID]/cgroup.
This will work across all cgroup managers and configurations and is 
supported on cgroups v1 and v2.                                     

Fixes: #8265

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -1,22 +1,18 @@
 package libpod
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
-	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/containernetworking/cni/pkg/types"
 	cnitypes "github.com/containernetworking/cni/pkg/types/current"
-	"github.com/containers/common/pkg/config"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/libpod/lock"
-	"github.com/containers/podman/v2/pkg/rootless"
-	"github.com/containers/podman/v2/utils"
 	"github.com/containers/storage"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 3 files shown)


vrothberg:
@giuseppe PTAL

mheon:
Are we sure this is a consistent thing? AFAIK we pass that path into the OCI runtime as-is for container creation? I'm worried it will be different on runc vs crun

libpod/oci_util.go
```diff
@@ -28,7 +28,7 @@ type ociError struct {
 
 // Create systemd unit name for cgroup scopes
 func createUnitName(prefix string, name string) string {
-	return fmt.Sprintf("%s-%s.scope", prefix, name)
+	return fmt.Sprintf("%s-%s.scope/container", prefix, name)
```
giuseppe, line 31: it is a crun specific feature.  It would not work with runc or another runtime.

I think the cgroup should be read from the process `/proc/PID/cgroup` file.

rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
LGTM

vrothberg:
Merge me

libpod/container.go
```diff
@@ -912,44 +908,23 @@ func (c *Container) CgroupManager() string {
...
 
-			return filepath.Join(parts[0], fmt.Sprintf("user-%d.slice/user@%d.service/user.slice/%s", uid, uid, dir), createUnitName("libpod", c.ID())), nil
-		}
-		return filepath.Join(c.config.CgroupParent, createUnitName("libpod", c.ID())), nil
-	default:
-		return "", errors.Wrapf(define.ErrInvalidArg, "unsupported CGroup manager %s in use", cgroupManager)
+	// Read /proc/[PID]/cgroup and look at the first line.  cgroups(7)
+	// nails it down to three fields with the 3rd pointing to the cgroup's
+	// path which works both on v1 and v2.
+	procPath := fmt.Sprintf("/proc/%d/cgroup", c.state.PID)
```
mheon, line 918: OK, so this merged when I wasn't looking. I'm pretty sure this completely ignores cgroup parent? I don't think this actually works.

mheon: OK, I begin to see. You split the portion that actually generates paths for creating the container out... Hm.

mheon: This is still very broken as an accessor though. It doesn't lock the container and sync state so it is useless if called outside of Libpod, and it doesn't check container state so we can't guarantee the container is running and PID is nonzero. This needs to either be made an internal-only function or have the locking/syncing and state check added (we'd probably need an internal, unlocked version as well in that case).

vrothberg: I think it makes sense to make it more robust. Currently I don't see how it could be called on a non-running container in Podman but for libpod, yes. May error out when it's not running?

mheon: Error seems appropriate, yep

vrothberg: https://github.com/containers/podman/pull/8375

PR #8292, opened 2020-11-10
[CI:DOCS] Add anchors for flag names on docs.podman.io

### LINKED ISSUE #7300: RFE: anchors in docs.podman.io's man pages would be nice to have
I'm not sure it is possible, or easily doable, but it would be nice to have anchors on the man pages that are shown in docs.podman.io.  For instance, I was reviewing @ashley-cui 's blog on timezones and I wanted to link to the tz option there, with something like: `http://docs.podman.io/en/latest/markdown/podman-run.1.html#--tz` but I could not.  I had to use a link to the top of the page, and it's a long scroll down.

Luap99:
Change the docs markdown so that flag names will be h4 headers.
Sphinx will automatically add anchors to headers. Add css to
make sure the flag names are not to big compared to the text.

The man pages also still renders fine but it looks a bit different.

Fixes #7300 

old man page:
```
OPTIONS
       --detach-keys=sequence

       Specify the key sequence for detaching a container. Format is a single character [a-Z] or one or
       more  ctrl-<value>  characters where <value> is one of: a-z, @, ^, [, , or _. Specifying "" will
       disable this feature. The default is ctrl-p,ctrl-q.

       --latest, -l

       Instead of providing the container name or ID, use the last created container. If you use  meth‐
       ods  other than Podman to run containers such as CRI-O, the last started container could be from
       either of those methods.

       The latest option is not supported on the remote client.

       --no-stdin

       Do not attach STDIN. The default is false.

       --sig-proxy=true|false

       Proxy received signals to the process (non-TTY mode only). SIGCHLD, SIGSTOP, and SIGKILL are not
       proxied. The default is true.

```
new man page:
```
OPTIONS
   --detach-keys=sequence
       Specify the key sequence for detaching a container. Format is a single character [a-Z] or one or
       more  ctrl-<value>  characters where <value> is one of: a-z, @, ^, [, , or _. Specifying "" will
       disable this feature. The default is ctrl-p,ctrl-q.

   --latest, -l
       Instead of providing the container name or ID, use the last created container. If you use  meth‐
       ods  other than Podman to run containers such as CRI-O, the last started container could be from
       either of those methods.

       The latest option is not supported on the remote client.

   --no-stdin
       Do not attach STDIN. The default is false.

   --sig-proxy=true|false
       Proxy received signals to the process (non-TTY mode only). SIGCHLD, SIGSTOP, and SIGKILL are not
       proxied. The default is true.

```
the look on docs.podman.io
gnome-shell-screenshot-190GT0

Github also adds anchors, see: https://github.com/Luap99/libpod/blob/doc-anchors/docs/source/markdown/podman-attach.1.md

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -1,0 +1,7 @@
+h2,h3,h4,h5,h6{
+    /*  Sphinx doesn't respect using a h4 header without using h1,h2,h3.
+        The flag names will be rendered as h2 and by default way to big
+        so make the font size for all headers except h1 smaller.
+    */
+    font-size: 120% !important;
+}
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -53,5 +53,8 @@ html_theme = 'alabaster'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+html_css_files = [
+    'custom.css',
+]
 
 # -- Extension configuration -------------------------------------------------
--- a/docs/source/markdown/podman-attach.1.md
+++ b/docs/source/markdown/podman-attach.1.md
@@ -18,22 +18,22 @@ Configure the keys sequence using the **--detach-keys** option, or specifying
 it in the **containers.conf** file: see **containers.conf(5)** for more information.
 
 ## OPTIONS
-**--detach-keys**=*sequence*
+#### **--detach-keys**=*sequence*
 
 Specify the key sequence for detaching a container. Format is a single character `[a-Z]` or one or more `ctrl-<value>` characters where `<value>` is one of: `a-z`, `@`, `^`, `[`, `,` or `_`. Specifying "" will disable this feature. The default is *ctrl-p,ctrl-q*.
 
-**--latest**, **-l**
+#### **--latest**, **-l**
 
 Instead of providing the container name or ID, use the last created container. If you use methods other than Podman
 to run containers such as CRI-O, the last started container could be from either of those methods.
 
 The latest option is not supported on the remote client.
 
-**--no-stdin**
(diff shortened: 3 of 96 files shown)


Luap99:
@rhatdan @TomSweeneyRedHat PTAL

baude:
lgtm

TomSweeneyRedHat:
I'm not seeing the change on the attach [page](http://docs.podman.io/en/latest/markdown/podman-attach.1.html).  Lookin at the source, I'm still seeing `<h2>` tags:
```
<h2><strong>–latest</strong>, <strong>-l</strong><a class="headerlink" href="#latest-l" title="Permalink to this headline">¶</a></h2>
```
Is it just taking a long time to push out there, or is there some under the covers resolution error?  @baude?

Luap99:
@TomSweeneyRedHat This is correct. I cannot change this. That's why I changed the css to match h4.

TomSweeneyRedHat:
@Luap99 so I'm confused.  You are saying that the css needs to be adjusted before the anchors will show on the docs.podman.io pages?  I like the change that you did here, but I'm not sure that it closed #7300.  The point of that issue is they wanted a clickable icon to show up in the page so they could create a link directly to a particular option in the man page.

Luap99:
@TomSweeneyRedHat If you hover the flag name you can click on the pilcrow on the right side of the flag name to generate the link.

rhatdan:
Nice work @Luap99

TomSweeneyRedHat:
@Luap99 thanks!  I was hovering on the left side of each and not seeing the link prior.  Is there any way to flip the link to the other side of the line?  I think that's where most folks would look for it and that's where it is in podman.io and buildah.io.

PR #8298, opened 2020-11-10
Add support for network connect / disconnect to DB

mheon:
Convert the existing network aliases set/remove code to network connect and disconnect. We can no longer modify aliases for an existing network, but we can add and remove entire networks. As part of this, we need to add a new function to retrieve current aliases the container is connected to (we had a table for this as of the first aliases PR, but it was not externally exposed).

At the same time, remove all deconflicting logic for aliases. Docker does absolutely no checks of this nature, and allows two containers to have the same aliases, aliases that conflict with container names, etc - it's just left to DNS to return all the IP addresses, and presumably we round-robin from there? Most tests for the existing code had to be removed because of this.


mheon:

baude:
LGTM, @rhatdan please review and merge

libpod/rootless_cni_linux.go
```diff
@@ -40,8 +40,12 @@ const (
...
 // AllocRootlessCNI does not lock c. c should be already locked.
 func AllocRootlessCNI(ctx context.Context, c *Container) (ns.NetNS, []*cnitypes.Result, error) {
-	if len(c.config.Networks) == 0 {
-		return nil, nil, errors.New("allocRootlessCNI shall not be called when len(c.config.Networks) == 0")
+	networks, err := c.networks()
+	if err != nil {
+		return nil, nil, err
+	}
+	if len(networks) == 0 {
+		return nil, nil, errors.New("allocRootlessCNI shall not be called when container has not joined any CNI networks")
```
TomSweeneyRedHat, line 48: nit to consider if you get other changes,   I'm not a big fan of function names in an error message and "shall" doesn't really fit as the function did do so if it got to here.  Suggest: 
```suggestion
		return nil, nil, errors.New("Allocating a rootless CNI can not be completed when the container has not yet joined any CNI networks")
```

libpod/rootless_cni_linux.go
```diff
@@ -77,8 +81,12 @@ func AllocRootlessCNI(ctx context.Context, c *Container) (ns.NetNS, []*cnitypes.
...
 // DeallocRootlessCNI does not lock c. c should be already locked.
 func DeallocRootlessCNI(ctx context.Context, c *Container) error {
-	if len(c.config.Networks) == 0 {
-		return errors.New("deallocRootlessCNI shall not be called when len(c.config.Networks) == 0")
+	networks, err := c.networks()
+	if err != nil {
+		return err
+	}
+	if len(networks) == 0 {
+		return errors.New("deallocRootlessCNI shall not be called when container has not joined any CNI networks")
```
TomSweeneyRedHat, line 89: similar to alloc above

libpod/rootless_cni_linux.go
rhatdan: I agree

mheon: Fixed

libpod/rootless_cni_linux.go
mheon: Fixed

mheon:
Restarted some tests, looked like flakes.

mheon:
This is ready @containers/podman-maintainers

Issue #8296, opened 2020-11-10
CI: need to reenable docs checks


edsantiago:
CI v2.0 removed the man page vs --help check:
```
-        - '/usr/local/bin/entrypoint.sh clean podman xref_helpmsgs_manpages BUILDTAGS="exclude_graphdriver_devicemapper selinux seccomp" |& ${TIMESTAMP}'
```
This needs to be reenabled ASAP. Unfortunately, the situation is now completely broken, the `hack/xref-helpmsgs-manpages` script is failing, I will need to spin up an emergency PR to fix it.

cevich:
@edsantiago re-assigned to you since my urgently requested fix, is effectively completely nullified by #8299

PR #8034, opened 2020-10-15
Switch help messages from using [flags] to [options]

rhatdan:
Want to have man pages match commands, since we have lots of printed
man pages with using Options, we will change the command line to use
Options in --help.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #8297, opened 2020-11-10
Cirrus: Run validation tests in CI:DOCS mode

cevich:

PR #8299, opened 2020-11-10
[CI:DOCS] Restore man page cross-checker

edsantiago:
Somewhere in the CIv2 migration we lost the man page vs --help
cross-checker. Add it back, by adding it into the man-page-check
Makefile target; this is part of 'make validate', which is run
in CI even on CI:DOCS PRs.

As happens when CI doesn't run, things broke. Man pages got out
of sync with --help. This PR:

 1) Fixes hack/xref-helpmsgs-manpages to deal with the new
    "Options" (instead of "Flags") form of podman help. #8034
    did part of that, but one of my review comments was
    accidentally left out.

 2) Fixes hack/xref-helpmsgs-manpages to deal with the new
    option syntax in man pages, post- #8292, in which each
    option is preceded by four hashes so as to make them
    HTML <h4> elements with named anchors.

 3) Fixes man pages that #8292 accidentally missed.

Fixes: #8296

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -50,6 +50,10 @@ function _run_validate() {
     # Confirm compile via prior task + cache
     bin/podman --version
     bin/podman-remote --version
+
+    # FIXME FIXME FIXME: 2020-11-10: remove once Fedora 33 has FindBin
+    perl -MFindBin -e 0 &>/dev/null || dnf -y install perl-FindBin
+
     make validate  # Some items require a build
 }
 
--- a/docs/source/markdown/podman-play-kube.1.md
+++ b/docs/source/markdown/podman-play-kube.1.md
@@ -42,6 +42,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+#### **--log-driver**=driver
+
+Set logging driver for all created containers.
+
 #### **--network**=*cni networks*
 
 A comma-separated list of the names of CNI networks the pod should join.
--- a/docs/source/markdown/podman-pod-create.1.md
+++ b/docs/source/markdown/podman-pod-create.1.md
@@ -63,7 +63,7 @@ The image that will be created for the infra container. Default: "k8s.gcr.io/pau
 
 Set a static IP for the pod's shared network.
 
-**-l**, **--label**=*label*
+#### **-l**, **--label**=*label*
 
 Add metadata to a pod (e.g., --label com.example.key=value).
 
@@ -75,7 +75,7 @@ Read in a line delimited file of labels.
 
 Set a static MAC address for the pod's shared network.
 
-**-n**, **--name**=*name*
+#### **-n**, **--name**=*name*
 
 Assign a name to the pod.
 
@@ -96,6 +96,10 @@ Set network mode for the pod. Supported values are
   - **port_handler=rootlesskit**: Use rootlesskit for port forwarding. Default.
   - **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
 
+#### **--network-alias**=strings
+
+Add a DNS alias for the container. When the container is joined to a CNI network with support for the dnsname plugin, the container will be accessible through this name from other containers in the network.
+
 #### **--no-hosts**=**true**|**false**
 
 Disable creation of /etc/hosts for the pod.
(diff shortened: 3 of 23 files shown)


edsantiago:
THIS IS GOING TO FAIL CI. At least I hope so. I've deliberately introduced two errors, because I want to make sure that the validation check will catch them. ("I want to see a negative before I see a positive." "What's that going to prove?" "Indulge me.")

I would've liked to make this a CI-colon-DOCS PR but cannot, because CI-colon-DOCS is currently a NOP: it doesn't even check docs. This is kind of bad, and will need to be fixed, but it's left as an exercise for another reader. (See #8297)

edsantiago:
Followup: as expected (and desired), initial commit [failed to validate](https://cirrus-ci.com/task/5457930671620096?command=main#L104). After pushing with corrected `.md` files, it [passed](https://cirrus-ci.com/task/4857872737632256). This is ready for review.

TomSweeneyRedHat:
LGTM
assuming happy tests

mheon:
LGTM

mheon:
Should this be CI:DOCS? Does CI:DOCS run your check?

edsantiago:
@mheon see comment 1. CI:DOCS is currently broken.

cevich:
@edsantiago pls rebase on the latest #8297 which includes (verified) working CI:DOCS support.

edsantiago:
@cevich sorry, #8297 includes an incorrect (and soon-to-be-broken) incantation in `runner.sh`. This PR solves the `xref-manpages` problem in a different, better way.

docs/source/markdown/podman-play-kube.1.md
```diff
@@ -42,6 +42,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+#### **--log-driver**=driver
+
+Logging driver for the container
+
```
edsantiago, line 48: This is one that snuck in while the CI tests were disabled. I just copied the --help output; it may need expansion.

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -96,6 +96,10 @@ Set network mode for the pod. Supported values are
   - **port_handler=rootlesskit**: Use rootlesskit for port forwarding. Default.
   - **port_handler=slirp4netns**: Use the slirp4netns port forwarding.
 
+#### **--network-alias**=strings
+
+Add network-scoped alias for the container
+
```
edsantiago, line 102: Ditto (re: needing expansion)

contrib/cirrus/runner.sh
```diff
@@ -50,6 +50,10 @@ function _run_validate() {
     # Confirm compile via prior task + cache
     bin/podman --version
     bin/podman-remote --version
+
+    # FIXME FIXME FIXME: 2020-11-10: remove once Fedora 33 has FindBin
+    perl -MFindBin -e 0 &>/dev/null || dnf -y install perl-FindBin
+
```
cevich, line 56: I'm on the fence, risk-wise.  Do you think it makes sense to instead of the above,  make use of VM image suffix 'c5402398833246208', since https://github.com/containers/automation_images/pull/39 is finished?

edsantiago: I think you know my answer to that: I like small reviewable PRs, not complicated multi-element ones. This one is unfortunately too big already, but every element is necessary. I can submit a new PR once this merges, to use the new VM images and remove the dnf-install.

edsantiago:
Rebased on master, which now includes #8297, which now means I can use CI-colon-DOCS. Confirmed that even with CI-colon-docs, [validate runs, and runs the xref script, and passes](https://cirrus-ci.com/task/5162292536934400?command=main#L104). Merge me please.

docs/source/markdown/podman-play-kube.1.md
mheon: Um. Probably "Set logging driver for all created containers"?

docs/source/markdown/podman-pod-create.1.md
mheon: "Add a DNS alias for the container. When the container is joined to a CNI network with support for the dnsname plugin, the container will be accessible through this name from other containers in the network.

TomSweeneyRedHat:
LGTM

docs/source/markdown/podman-pod-create.1.md
edsantiago: Updated and pushed, thanks.

docs/source/markdown/podman-play-kube.1.md
edsantiago: Updated and pushed, thanks.

TomSweeneyRedHat (commented):
LGTM
will let @mheon push the merge button.  Would like this is soon if possible, I suspect we've another manpage PR inbound.

PR #8301, opened 2020-11-11
Change podman build --pull=true to PullIfMissing

TomSweeneyRedHat:
One last tweak to the man page for 'build --pull' and after
further testing against Docker, one slight change to the
pull policy.  First I changed `--pull=false` from PullNever
to PullIfMissing.  This matches Docker and will pull the
image if it's not present rather than erroring.  We've
the `--pull-never` option if someone wants the pull to
not do an actual pull and to error if the image isn't
local.

Then for the man page, I'd a much bigger change, in the
initial PR, I've backed most of that out and just
added a tweak.
Hopefully, this puts this portion of the pull work behind
us for a while.

### DIFF
--- a/docs/source/markdown/podman-build.1.md
+++ b/docs/source/markdown/podman-build.1.md
@@ -438,9 +438,9 @@ When the option is specified or set to "true", pull the image from the first
 registry it is found in as listed in registries.conf.  Raise an error if not
 found in the registries, even if the image is present locally.
 
-If the option is disabled (with *--pull=false*), or not specified, pull the
+If the option is disabled (with *--pull=false*) or not specified, pull the
 image from the registry only if the image is not present locally. Raise an
-error if the image is not found in the registries.
+error if the image is not found in the registries and is not present locally.
 
 #### **--pull-always**
 
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -249,7 +249,7 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 	}
 
 	if flags.PullNever {
-		pullPolicy = imagebuildah.PullNever
+		pullPolicy = imagebuildah.PullIfMissing
 	}
 
 	args := make(map[string]string)


vrothberg (commented):
LGTM

docs/source/markdown/podman-build.1.md
```diff
@@ -438,10 +438,14 @@ When the option is specified or set to "true", pull the image from the first
 registry it is found in as listed in registries.conf.  Raise an error if not
 found in the registries, even if the image is present locally.
 
-If the option is disabled (with *--pull=false*), or not specified, pull the
+If the option is disabled (with *--pull=false*), pull the
 image from the registry only if the image is not present locally. Raise an
 error if the image is not found in the registries.
 
+If the option is not specified, pull the image from the registry if it is not
+present locally or if there is a newer version on the registry.  Raise an
```
rhatdan, line 446: This is technically incorrect.  It pulls if the registry version is different, not necessarily newer.

Image I create a container image and push it to a registry as :latest, later I find there is a big bug in my version, and remove it, causing the previous version to be :latest.  Now if someone had pulled in my buggy image, and later pulls again they will revert back to the older image, even though they were PullIfNewer.

Container/Image has no way to determine whether or not one image is "NEWER" then another image, it can only check if the digest is different.  We recently received a bug on this misunderstanding from a user, so we should not be documenting it incorrectly in the man pages and other docs.

TomSweeneyRedHat: @vrothberg I thought you had recently changed the check to pull only if the image had a newer date stamp, not that it had changed.  Am I misremembering?

umohnani8:
LGTM

docs/source/markdown/podman-build.1.md
rhatdan: I don't know, but as @mtrmac pointed out in a different issue/pr this is a bad idea, since it will not remove a bad image unless someone pushes a "new" version.  As opposed to removing the old "new" image.

vrothberg: > @vrothberg I thought you had recently changed the check to pull only if the image had a newer date stamp, not that it had changed. Am I misremembering?

An earlier version of the shortnames PR for Buildah had that but I reverted that part based on the feedback. A digest check is really the way forward. Some images claim to be have been born before me which I doubt :^)

TomSweeneyRedHat:
Please re-review from the start.  Much smaller change, plus I've added a Wiki [page](https://github.com/containers/podman/wiki/Podman-Build-Policies-Explained) to explain the pull policies if you're interested in more depth.

PR #8304, opened 2020-11-11
Cleanup error reporting

rhatdan:
The error message reported is overlay complicated and the added test does not
really help the user.

Currently the error looks like:

podman run -p 80:80 fedora echo hello
Error: failed to expose ports via rootlessport: "cannot expose privileged port 80, you might need to add "net.ipv4.ip_unprivileged_port_start=0" (currently 1024) to /etc/sysctl.conf, or choose a larger port number (>= 1024): listen tcp 0.0.0.0:80: bind: permission denied\n"

After this change

./bin/podman run -p 80:80 fedora echo hello
Error: cannot expose privileged port 80, you might need to add "net.ipv4.ip_unprivileged_port_start=0" (currently 1024) to /etc/sysctl.conf, or choose a larger port number (>= 1024): listen tcp 0.0.0.0:80: bind: permission denied

Control chars have been eliminated.

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -556,7 +556,7 @@ func (r *Runtime) setupRootlessPortMappingViaRLK(ctr *Container, netnsPath strin
 		if stdoutStr != "" {
 			// err contains full debug log and too verbose, so return stdoutStr
 			logrus.Debug(err)
-			return errors.Errorf("failed to expose ports via rootlessport: %q", stdoutStr)
+			return errors.Errorf("rootlessport " + strings.TrimSuffix(stdoutStr, "\n"))
 		}
 		return err
 	}


rhatdan:
Also opened PR https://github.com/rootless-containers/rootlesskit/pull/192 for this message.

TomSweeneyRedHat:
LGTM

zhangguanzhang:
LGTM

libpod/networking_linux.go
```diff
@@ -556,7 +556,7 @@ func (r *Runtime) setupRootlessPortMappingViaRLK(ctr *Container, netnsPath strin
 		if stdoutStr != "" {
 			// err contains full debug log and too verbose, so return stdoutStr
 			logrus.Debug(err)
-			return errors.Errorf("failed to expose ports via rootlessport: %q", stdoutStr)
+			return errors.Errorf(strings.TrimSuffix(stdoutStr, "\n"))
```
mheon, line 559: I'm a little iffy about removing any reference to RootlessPort - we want to leave some breadcrumbs so developers can find where the error came from.

rhatdan: Well I found this code by grep'ing through the vendor directory.  Not based on the first part.
Would you be ok with:
return errors.Errorf("rootlessport: " + strings.TrimSuffix(stdoutStr, "\n"))

mheon: Sure, I'm ok with that. I just want something that I can easily grep for to figure out where an error message is coming rom

libpod/networking_linux.go
```diff
@@ -556,7 +556,7 @@ func (r *Runtime) setupRootlessPortMappingViaRLK(ctr *Container, netnsPath strin
 		if stdoutStr != "" {
 			// err contains full debug log and too verbose, so return stdoutStr
 			logrus.Debug(err)
-			return errors.Errorf("failed to expose ports via rootlessport: %q", stdoutStr)
+			return errors.Errorf("rootlessport " + strings.TrimSuffix(stdoutStr, "\n"))
```
mheon, line 559: Missing a `:` after `rootlessport`

rhatdan:

libpod/networking_linux.go
rhatdan: It reads better without it.  I removed it intentionally.

PR #8312, opened 2020-11-12
Cirrus: Update VM Images; Both Fedora and Ubuntu "prior" flavors run with CGroupsV1 & runc

cevich:
***TODO:*** Open issues for disabled tests

- [x] Disable blkio.weight test on Ubuntu (#8737)
- [x] Disable pod stats tests in containerized Fedora w/ CGroupsV1 (#8732)
- [x] Disable remote pod stats on net=host post (#8685)
- [x] Disable rootless pod stats tests w/ CgroupV1 (#8734)
- [x] Disable F32 rootless system testing (#8729)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,11 +26,11 @@ env:
     ####
     FEDORA_NAME: "fedora-33"
     PRIOR_FEDORA_NAME: "fedora-32"
-    UBUNTU_NAME: "ubuntu-20"
-    PRIOR_UBUNTU_NAME: "ubuntu-19"
+    UBUNTU_NAME: "ubuntu-2010"
+    PRIOR_UBUNTU_NAME: "ubuntu-2004"
 
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c4704091098054656"
+    IMAGE_SUFFIX: "c6233039174893568"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -41,7 +41,7 @@ fi
 
 OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
 # GCE image-name compatible string representation of distribution _major_ version
-OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | cut -d '.' -f 1)"
+OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
 # Combined to ease soe usage
 OS_REL_VER="${OS_RELEASE_ID}-${OS_RELEASE_VER}"
 # This is normally set from .cirrus.yml but default is necessary when
--- a/test/e2e/pod_stats_test.go
+++ b/test/e2e/pod_stats_test.go
@@ -179,7 +179,8 @@ var _ = Describe("Podman pod stats", func() {
 
 	It("podman stats on net=host post", func() {
 		// --net=host not supported for rootless pods at present
-		SkipIfRootlessCgroupsV1("Pause stats not supported in cgroups v1")
+		// problem with sysctls being passed to containers of the pod.
+		SkipIfCgroupV1("Bug: Error: sysctl net.ipv4.ping_group_range is not allowed in the hosts network namespace: OCI runtime error")
 		podName := "testPod"
 		podCreate := podmanTest.Podman([]string{"pod", "create", "--net=host", "--name", podName})
 		podCreate.WaitWithDefaultTimeout()
--- a/test/e2e/pod_stats_test.go
+++ b/test/e2e/pod_stats_test.go
@@ -17,9 +17,7 @@ var _ = Describe("Podman pod stats", func() {
 	)
 
 	BeforeEach(func() {
-		if os.Geteuid() != 0 {
-			SkipIfCgroupV2("--cgroup-manager=cgroupfs which doesn't work in rootless mode")
-		}
+		SkipIfRootless("Tests fail with both CGv1/2 + required --cgroup-manager=cgroupfs")
 		if isContainerized() {
 			SkipIfCgroupV1("All tests fail Error: unable to load cgroup at ...: cgroup deleted")
 		}
--- a/test/e2e/pod_stats_test.go
+++ b/test/e2e/pod_stats_test.go
@@ -20,6 +20,9 @@ var _ = Describe("Podman pod stats", func() {
 		if os.Geteuid() != 0 {
 			SkipIfCgroupV2("--cgroup-manager=cgroupfs which doesn't work in rootless mode")
 		}
+		if isContainerized() {
+			SkipIfCgroupV1("All tests fail Error: unable to load cgroup at ...: cgroup deleted")
+		}
 
 		tempdir, err = CreateTempDirInTempDir()
 		if err != nil {
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -493,7 +493,9 @@ USER bin`
 				Skip("Kernel does not support blkio.weight")
 			}
 		}
-
+		if podmanTest.Host.Distribution == "ubuntu" {
+			Skip("Ubuntu <= 20.10 lacks BFQ scheduler")
+		}
 		if CGROUPSV2 {
 			// convert linearly from [10-1000] to [1-10000]
 			session := podmanTest.Podman([]string{"run", "--rm", "--blkio-weight=15", ALPINE, "sh", "-c", "cat /sys/fs/cgroup/$(sed -e 's|0::||' < /proc/self/cgroup)/io.bfq.weight"})
--- a/contrib/cirrus/logcollector.sh
+++ b/contrib/cirrus/logcollector.sh
@@ -53,7 +53,7 @@ case $1 in
                     slirp4netns \
         )
         case $OS_RELEASE_ID in
-            fedora*)
+            fedora)
                 cat /etc/fedora-release
                 PKG_LST_CMD='rpm -q --qf=%{N}-%{V}-%{R}-%{ARCH}\n'
                 PKG_NAMES+=(\
@@ -61,7 +61,7 @@ case $1 in
                     libseccomp \
                 )
                 ;;
-            ubuntu*)
+            ubuntu)
                 cat /etc/issue
                 PKG_LST_CMD='dpkg-query --show --showformat=${Package}-${Version}-${Architecture}\n'
                 PKG_NAMES+=(\
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -41,7 +41,7 @@ function _run_automation() {
     req_env_vars CI DEST_BRANCH IMAGE_SUFFIX TEST_FLAVOR TEST_ENVIRON \
                  PODBIN_NAME PRIV_NAME DISTRO_NV CONTAINER USER HOME \
                  UID AUTOMATION_LIB_PATH SCRIPT_BASE OS_RELEASE_ID \
-                 OS_RELEASE_VER CG_FS_TYPE
+                 CG_FS_TYPE
     bigto ooe.sh dnf install -y ShellCheck  # small/quick addition
     $SCRIPT_BASE/shellcheck.sh
 }
@@ -64,12 +64,6 @@ function _run_unit() {
 }
 
 function _run_apiv2() {
-    # TODO Remove once VM's with dependency
-    if [[ "$OS_RELEASE_ID" == "fedora" ]]; then
-        dnf install -y python3-docker
-    else
-        apt-get -qq -y install python3-docker
-    fi
     make localapiv2 |& logformatter
 }
 
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -73,7 +73,9 @@ case "$CG_FS_TYPE" in
         if ((CONTAINER==0)); then
             warn "Forcing testing with runc instead of crun"
             if [[ "$OS_RELEASE_ID" == "ubuntu" ]]; then
-                echo "OCI_RUNTIME=/usr/lib/cri-o-runc/sbin/runc" >> /etc/ci_environment
+                # Need b/c using cri-o-runc package from OBS
+                echo "OCI_RUNTIME=/usr/lib/cri-o-runc/sbin/runc" \
+                    >> /etc/ci_environment
             else
                 echo "OCI_RUNTIME=runc" >> /etc/ci_environment
             fi
@@ -102,8 +104,8 @@ fi
 
 # Which distribution are we testing on.
 case "$OS_RELEASE_ID" in
-    ubuntu*) ;;
-    fedora*)
+    ubuntu) ;;
+    fedora)
         if ((CONTAINER==0)); then
             msg "Configuring / Expanding host storage."
             # VM is setup to allow flexibility in testing alternate storage.
@@ -123,10 +125,15 @@ esac
 # shellcheck disable=SC2154
 case "$TEST_ENVIRON" in
     host)
-        if [[ "$OS_RELEASE_ID" == "fedora" ]]; then
-            # The e2e tests wrongly guess `--cgroup-manager cgroupfs`
+        # The e2e tests wrongly guess `--cgroup-manager` option
+        # shellcheck disable=SC2154
+        if [[ "$CG_FS_TYPE" == "cgroup2fs" ]] || [[ "$PRIV_NAME" == "root" ]]
+        then
             warn "Forcing CGROUP_MANAGER=systemd"
             echo "CGROUP_MANAGER=systemd" >> /etc/ci_environment
+        else
+            warn "Forcing CGROUP_MANAGER=cgroupfs"
+            echo "CGROUP_MANAGER=cgroupfs" >> /etc/ci_environment
         fi
         ;;
     container)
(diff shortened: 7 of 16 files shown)


cevich:
Status: [Ubuntu 20.10 image's runc version is too damn old.](https://github.com/containers/automation_images/pull/34#issuecomment-728342014)

rhatdan:
@cevich Can we switch to using crun?

cevich:
@rhatdan Yes, I believe we can...or at least make an attempt.  I was actually looking into this last night, it seems that [systemd in Ubuntu enables cgroupsv2 the same way as Fedora (just not by default).](http://manpages.ubuntu.com/manpages/groovy/man1/systemd.1.html)

@mheon @baude @giuseppe Q: Assuming it can be done, are all'y'all okay with the ***only*** testing of `runc`, occurring on "prior" Ubuntu (20.04 LTS) VMs?  This implies we would drop testing with `runc` in the future, when the 20.04 VMs go out of style.

rhatdan:
Well maybe at some point someone will update ubuntu version of runc.  We could also switch one of the fedora tests to use runc.

cevich:
> We could also switch one of the fedora tests to use runc.

This is true...and based on past experience, ***much*** easier to obtain necessary package updates.

cevich:
Note, need to open issues re: newly skipped tests:
    * Disable blkio.weight test on Ubuntu
    * Disable pod stats tests on Fedora w/ CGroupsV1 || container

cevich:
Oof! This is the best I can do, I really need help sorting through why/fixing [the 11x F32 (CGv1) rootless tests](https://cirrus-ci.com/task/6093961506848768) are failing.

cevich:
@containers/podman-maintainers PTAL.  This was a weeks-long ordeal to get the images both building and passing most tests (here).  I've added a number of commits to disable problematic tests, most of which seem like real problems that need fixing.  However, I'm holding off on opening corresponding issues until I receive review feedback.   Thanks in advance.

.cirrus.yml
```diff
@@ -509,7 +509,13 @@ rootless_integration_test_task:
...
     depends_on:
         - unit_test
-    matrix: *fedora_vm_axis
+    # Disabled: F32 rootless + CgroupsV1 fails 55/127 tests
+    #matrix: *fedora_vm_axis
+    env:
+        DISTRO_NV: ${FEDORA_NAME}
+        _BUILD_CACHE_HANDLE: ${FEDORA_NAME}-build-${CIRRUS_BUILD_ID}
+        VM_IMAGE_NAME: ${FEDORA_CACHE_IMAGE_NAME}
+        CTR_FQIN: ${FEDORA_CONTAINER_FQIN}
```
cevich, line 518: Logs: https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5688847692464128/html/sys-podman-fedora-32-rootless-host.log.html

.cirrus.yml
```diff
@@ -509,7 +509,13 @@ rootless_integration_test_task:
     skip: *branch
     depends_on:
         - unit_test
-    matrix: *fedora_vm_axis
+    # Disabled: F32 rootless + CgroupsV1 fails 55/127 tests
```
cevich, line 512: Logs: https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5688847692464128/html/sys-podman-fedora-32-rootless-host.log.html

contrib/cirrus/lib.sh
```diff
@@ -41,7 +41,7 @@ fi
 
 OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
 # GCE image-name compatible string representation of distribution _major_ version
-OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | cut -d '.' -f 1)"
+OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
```
cevich, line 44: Note: This is needed to support *two* major-version `20` Ubuntu VMs.  For example, previously `$OS_RELEASE_VER` was `20`, now it's `2010`.  This enables simple version comparisons in scripts, since bash does not like floating-point numbers.

TomSweeneyRedHat:
@cevich got some test failures to check still.  The changes look good to me, but you're stretching my CI knowledge with those changes.

cevich:
> some test failures to check still. 

Woopsie...I'll fix and re-run...

> The changes look good to me, but you're stretching my CI knowledge with those changes.

Oh no, don't worry.  It's actually completely black-magic this time :grin:

.cirrus.yml
```diff
@@ -559,7 +559,13 @@ rootless_system_test_task:
     only_if: *not_docs
     depends_on:
       - rootless_integration_test
-    matrix: *fedora_vm_axis
+    # disabled: f32 rootless + cgroupsv1 fails 55/127 tests
+    #matrix: *fedora_vm_axis
```
cevich, line 563: Logs: https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5688847692464128/html/sys-podman-fedora-32-rootless-host.log.html

rhatdan:
LGTM
The  tests requiring turning off is unfortunate, but we can examine them in the future.  I am hoping cgroupv1 is going away and most of them involve handling that.

vrothberg (commented):
LGTM.  Thanks for the great work, @cevich !

Having a card to re-enable the tests would be great.

cevich:
> The tests requiring turning off is unfortunate, but we can examine them in the future

Believe me, I tried everything I could. It seems like the tests have grown accustom to CGv2.  I'll open issues to track each one of the "disabled" commits, so we don't forget to look during a future bug-scrub.

cevich:
Ooof, and it seems the net I cast was too small for `pod stats on net=host post`.  It seems to be flaking (for some reason) b/c I swore it was passing on the last run...anyway, there's a merge conflict on `.cirrus.yml`, so I need to rebase anyway.

cevich:
Rebased w/ `SkipIfCgroupV1()` use for the `pod stats on net=host post` instead of only skipping for rootless.

cevich:
Ooof, spelled `SkipIfCgroupV1()` wrongly as "groups" (with an "s") :disappointed:

edsantiago:
I'm sorry, I've only just now looked at this, and all I can say is WTF? Why are we disabling tests? And why is there no issue filed about this "libpod_parent" problem? Or have I missed it? (I searched podman and conmon issues).

My gut reaction is that this is a non-problem, an artifact of the recent move from loglevel=error to =warning (but I have no way of confirming). If this is so, I don't think the solution is to disable tests. What am I missing? Where is the root cause of this (the libpod_parent problem) being tracked?

cevich:
> Why are we disabling tests?

@edsantiago In order to maintain separation of concerns (and avoid an overly bloated PR) it's best if we keep the test-issues and image-update items distinct. For each category of disablement, I added a commit explaining "why", and intend to file issues on each item once this is merged (hence the checklist in the description).  In (nearly) every case, the problems are caused due to the VM Images changing:

* F32 from CGv2 -> v1 (and crun -> runc).
* In Ubuntu 20.10 from CGv1 -> CGv2 (and runc -> crun)

Dan mentioned in standup recently, there is a need to exercise CGv1 in Fedora, as some folks still depend on it (e.g. to support staggered migration from docker -> podman).  The change in Ubuntu (while non-default) is there to prepare CI/testing for future changes in coming months (i.e. `21.04` expected in Apr.).

edsantiago:
This is already a bloated PR. And the correct order of execution is (1) file issues, (2) reference them in this PR, and then, only then, if everyone is happy, merge this PR. We do not postpone the filing of issues.

I still need to understand if these failures are real or not. I'll try to set up an f32 cgroups v1 VM today to find out.

cevich:
> And the correct order

We're doing things differently here, as it's more practical for the team (multiple people are needed to debug the issues).  There are other PRs blocking on having these updated images.  They cannot wait on a myriad of fixes followed by required eleven-billion (more) CI-runs.  With the holidays and people coming/going over the next month, PR's would remain blocked for an unacceptable amount of time.

edsantiago:
I've confirmed that podman works just fine on cgroups v1, it's just the warning message that screws up test checks. Please apply this diff and reenable tests:
```diff
diff --git a/test/system/helpers.bash b/test/system/helpers.bash
index f782de080..26625a4e8 100644
--- a/test/system/helpers.bash
+++ b/test/system/helpers.bash
@@ -145,11 +145,26 @@ function run_podman() {
         '?')             expected_rc=  ; shift;;  # ignore exit code
     esac

+    # FIXME FIXME FIXME: workaround for #8721
+    # When running rootless under cgroups v1, podman emits a useless warning:
+    #    Failed to add conmon to cgroupfs sandbox cgroup: error creating cgroup for memory: mkdir /sys/fs/cgroup/memory/libpod_parent: permission denied
+    # This is an artifact of #8479, in which log-level changed from error
+    # to warn. The warning had been hidden all these years. Podman works
+    # just fine, but we cannot tolerate those warnings in our fine-tuned
+    # output checks. For purposes of #8312 (the PR that adds Ubuntu 20.10
+    # and sets up some cgroups v1 testing), we work around the problem
+    # via the stick-head-in-sand approach.
+    # Please remove this workaround once #8721 is fixed.
+    local LOGLEVEL_OVERRIDE=
+    if is_cgroupsv1; then
+        LOGLEVEL_OVERRIDE="--log-level=error"
+    fi
+
     # stdout is only emitted upon error; this echo is to help a debugger
     echo "$_LOG_PROMPT $PODMAN $*"
     # BATS hangs if a subprocess remains and keeps FD 3 open; this happens
     # if podman crashes unexpectedly without cleaning up subprocesses.
-    run timeout --foreground -v --kill=10 $PODMAN_TIMEOUT $PODMAN "$@" 3>/dev/null
+    run timeout --foreground -v --kill=10 $PODMAN_TIMEOUT $PODMAN $LOGLEVEL_OVERRIDE "$@" 3>/dev/null
     # without "quotes", multiple lines are glommed together into one
     if [ -n "$output" ]; then
         echo "$output"
```

edsantiago:
Here's the diff in downloadable form, to make it easier to apply [pr8312-loglevel-fix.txt](https://github.com/containers/podman/files/5691781/pr8312-loglevel-fix.txt)

edsantiago:
Looking at the logs... is this happening only on fedora? Not on ubuntu, even with cgroups v1? If so, the `if` condition should be changed to `if is_cgroupsv1 && SOMETHING-SOMETHING-CHECK-FOR-FEDORA`. That way we will have the default log level on ubuntu.

cevich:
> If so, the if condition should be changed to

You know as well as any...A fix-in-concept is not the same as a fix that passes CI.  I would really prefer not to delay this PR with additional cycles:  ***Other PRs have been blocked on this one for weeks***.  It will be much easier, more team time-efficient, to support parallel debugging and fixing [of disabled tests] by merging this PR and opening dedicated "fix" PRs, [like this one.](https://github.com/containers/podman/pull/8729)

edsantiago:
I am not going to sign off on anything that mass-disables tests.

cevich:
@rhatdan @baude we need an arbiter here.  New images and this PR is holding up Dan's #8508.  I'm gone for the next two weeks, and am assuming the team is unwilling to wait another month for test-fixes to go in before this PR merges.  @mheon was made aware of the situation last week before he left.  Nobody is happy about disabling tests, but it seems a necessary [short-term] thing.  I'm working on opening tracking issues w/ links to failures in logs.  I'm unable to envision a different PR situation that doesn't involve holding progress hostage in some way :disappointed:

edsantiago:
Better solution: @giuseppe has submitted #8726 to fix the warning message at the source; this is the proper solution, obviating the need for my kludgy workaround. If that gets merged today, you could rebase on it and remove 4dfb29a9af2ee87c492affe27264cf22535d1111. The rest of the changes mostly LGTM, although I'm going to make some suggestions inline.

test/e2e/pod_stats_test.go
```diff
@@ -17,8 +17,9 @@ var _ = Describe("Podman pod stats", func() {
 	)
 
 	BeforeEach(func() {
-		if os.Geteuid() != 0 {
-			SkipIfCgroupV2("--cgroup-manager=cgroupfs which doesn't work in rootless mode")
+		SkipIfRootless("Tests fail with both CGv1/2 + required --cgroup-manager=cgroupfs")
+		if isContainerized() {
+			SkipIfCgroupV1("All tests fail Error: unable to load cgroup at ...: cgroup deleted")
```
edsantiago, line 22: A better way to word these messages is `Skip...("FIXME: #12345: Tests failing due to xyz")`, where 12345 is a PR you've filed showing an exact reproducer along with its symptoms. This has many benefits: first and foremost, it tracks the problem. Second, once the issue is fixed, it is easy for the developer to grep for the issue number and re-enable the test. Finally, it makes it immediately obvious to a maintainer that the given `Skip` is a real problem, not a shrug-unsupported like the ones above. 

This is not worth blocking on, but since you have to resubmit the PR anyway once #8726 is fixed, I'd encourage you to address this one.

test/e2e/pod_stats_test.go
```diff
@@ -176,7 +177,8 @@ var _ = Describe("Podman pod stats", func() {
 
 	It("podman stats on net=host post", func() {
 		// --net=host not supported for rootless pods at present
-		SkipIfRootlessCgroupsV1("Pause stats not supported in cgroups v1")
+		// problem with sysctls being passed to containers of the pod.
+		SkipIfCgroupV1("Bug: Error: sysctl net.ipv4.ping_group_range is not allowed in the hosts network namespace: OCI runtime error")
```
edsantiago, line 181: Same thing (`Skip` should be informative and reference a filed issue)

test/e2e/run_test.go
```diff
@@ -493,7 +493,9 @@ USER bin`
 				Skip("Kernel does not support blkio.weight")
 			}
 		}
-
+		if podmanTest.Host.Distribution == "ubuntu" {
```
edsantiago, line 496: This is fragile, and would be better expressed as a (TBD) check against `/sys/block/sda/queue/scheduler`, but I'll settle for a FIXME comment referencing the BZ (sorry, I couldn't find the BZ in `setup_environment.sh`).

test/e2e/pod_stats_test.go
cevich: It's a good suggestion, I'll keep that in mind.

test/e2e/run_test.go
cevich: I actually considered this, but it only turns out slightly less fragile: Who's to say the host has a SCSI block device?  Could be virtio, ide, nvme, usb, etc.  

I think the proper fix, is to (somehow) examine the set of available io-schedulers in the kernel...which I'm not eager to attempt (and do it badly).  I think it's better to throw in a short-term kludge and rely on someone more golang-savvy to solve properly later.

cevich: (I'm still working on opening issues)

cevich: ...in fact, it looks like _almost_ the correct skip is already here: `if _, err := os.Stat("/sys/fs/cgroup/blkio/blkio.weight"); os.IsNotExist(err) {`.  However, in the case of Ubuntu 20.10 w/ CGroupsV2, it doesn't cause the test to be skipped.  More, deeper debugging is necessary.  I'm working on opening an issue now.

cevich:
Okay, issues have been filed covering all cases of temporarily disabled tests.  The only exception are tests covered by the `Disable incompatible rootless + CGroupsV1 tests` commit.  Those are actually incompatible and must remain disabled, Giuseppe already looked and gave a head-nod.

edsantiago:
Thank you! #8726 is halfway through CI, maybe an hour left. I'm monitoring it closely to restart flakes quickly; will ping you once it's merged.

cevich:
> maybe an hour left.

Gee wiz, I've been saying the exact same thing on this PR...for about a month :smile: 

If only it were that simple...keep in mind _THIS_ PR will then need to be rebased, and rerun all the tests again...with fingers crossed nothing got missed.  I dunno if the team wants to wait that long or not.

edsantiago:
#8726 is merged!

edsantiago:
Tricky. I'm on it.

edsantiago:
Fix (tested on f32 cgroupsv1 VM; also on my f32 cgroupsv2 laptop):
```diff
diff --git a/test/system/600-completion.bats b/test/system/600-completion.bats
index 1e43cdc41..d1244c7d7 100644
--- a/test/system/600-completion.bats
+++ b/test/system/600-completion.bats
@@ -8,6 +8,17 @@

 load helpers

+# Returns true if we are able to podman-pause
+function _can_pause() {
+    # Even though we're just trying completion, not an actual unpause,
+    # podman barfs with:
+    #    Error: unpause is not supported for cgroupv1 rootless containers
+    if is_rootless && is_cgroupsv1; then
+        return 1
+    fi
+    return 0
+}
+
 function check_shell_completion() {
     local count=0

@@ -70,8 +81,13 @@ function check_shell_completion() {
                     ;;

                 *CONTAINER*)
+                    # podman unpause fails early on rootless cgroupsv1
+                    if [[ $cmd = "unpause" ]] && ! _can_pause; then
+                        continue 2
+                    fi
+
                     run_completion "$@" $cmd "${extra_args[@]}" ""
-                    is "$output" ".*-$random_container_name${nl}" "Found expected container in suggestions"
+                    is "$output" ".*-$random_container_name${nl}" "Found expected container in suggestions for '$cmd'"

                     match=true
                     # resume
@@ -212,7 +228,9 @@ function _check_completion_end() {
     run_podman create --name created-$random_container_name $IMAGE
     run_podman run --name running-$random_container_name -d $IMAGE top
     run_podman run --name pause-$random_container_name -d $IMAGE top
-    run_podman pause pause-$random_container_name
+    if _can_pause; then
+        run_podman pause pause-$random_container_name
+    fi
     run_podman run --name exited-$random_container_name -d $IMAGE echo exited

     # create pods for each state
```
[downloadable diff](https://github.com/containers/podman/files/5702976/pr8312-completion-fix.txt)

@Luap99 PTAL

cevich:
@edsantiago tip, if you're using the WebUI: you can drag-n-drop files directly into the comment box.  It will upload and automatically splat in the necessary markdown.

edsantiago:
I'm a crusty old UNIX user from the 1980s. I've heard of drag & drop, but it requires newfangled window managers or something like that which I've never bothered upgrading to. Now get off my lawn!

cevich:
> Now get off my lawn!

Nownow ---> I <---- also grew up in the 80's, but with *M$ DOS*...shudder. Hell, I only graduated (downgraded) from a flip-phone just a few years ago.   If ***I*** can learn to clicky-clicky...so can you :sweat_smile: 

The truly magical part is the uploading...It seems to go into some special storage location associated with the repository, but which is otherwise inaccessible (in the WebUI).  Like this:

[somefile.txt](https://github.com/containers/podman/files/5703622/somefile.txt)

cevich:
@edsantiago we got *really* lucky here.  Something _just_ fell over on the Cirrus-CI side at the end of the run.  The "Success" task (a noop)  normally starts within seconds, it took a very slow 1:25 to get going.  Anyway, thanks for your help clearing up the system-testing issue.  I recommend we don't push our luck any further before the hollow-days break.  No need to subject people to Ubuntu 19, or hold up Dan's PR any longer.

edsantiago:
/lgtm
/hold

(Disclosure: I did the vast majority of review yesterday; right now I just reviewed the diffs between this and the prior PR revision). This is a large enough PR that I'd like one more pair of eyes before merging, hence the hold.

Nice work @cevich . Thank you for your patience with my intransigence.

rhatdan:
/hold cancel
LGTM

PR #8307, opened 2020-11-11
Ensure we do not double-lock the same volume in create

mheon:
When making containers, we want to lock all named volumes we are adding the container to, to ensure they aren't removed from under us while we are working. Unfortunately, this code did not account for a container having the same volume mounted in multiple places so it could deadlock. Add a map to ensure that we don't lock the same name more than once to resolve this.

Fixes #8221

PR #8325, opened 2020-11-12
system tests: various

edsantiago:
- images: confirm that 'podman images' emits headings
  even if there are no images present. Intended to
  replace e2e test which is difficult to get working
  under podman-remote.

- build: add test for #8092, podman-build gobbling stdin.

- volumes: add test for #8307, double-lock on same volume

### DIFF
--- a/test/system/010-images.bats
+++ b/test/system/010-images.bats
@@ -3,10 +3,18 @@
 load helpers
 
 @test "podman images - basic output" {
-    run_podman images -a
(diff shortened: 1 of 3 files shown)


baude:
LGTM, thanks @edsantiago

mheon:
LGTM

edsantiago:
Well, caught a real bug. Filed #8332

rhatdan:
@edsantiago  #8332 merged

PR #8327, opened 2020-11-12
test for buildah version in container images.

rhatdan:
Check to see if we are recording the version of buildah
used to build the image as a label in the image.

Also we should make sure the filter "since" works.
We are only testing "after", which we don't document.

### DIFF
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -221,6 +221,11 @@ EOF
     run_podman run --rm build_test pwd
     is "$output" "$workdir" "pwd command in container"
 
+    # Determine buildah version, so we can confirm it gets into Labels
+    run_podman info --format '{{ .Host.BuildahVersion }}'
+    is "$output" "[1-9][0-9.-]\+" ".Host.BuildahVersion is reasonable"
+    buildah_version=$output
+
     # Confirm that 'podman inspect' shows the expected values
     # FIXME: can we rely on .Env[0] being PATH, and the rest being in order??
     run_podman image inspect build_test
@@ -239,6 +244,7 @@ Cmd[0]             | /bin/mydefaultcmd
 Cmd[1]             | $s_echo
 WorkingDir         | $workdir
 Labels.$label_name | $label_value
+Labels.\"io.buildah.version\" | $buildah_version
 "
 
     parse_table "$tests" | while read field expect; do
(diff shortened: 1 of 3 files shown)


test/system/070-build.bats
```diff
@@ -312,6 +312,23 @@ EOF
...
+
+    cat >$tmpdir/Dockerfile <<EOF
+FROM $IMAGE
+WORKDIR /tmp/dan
+EOF
+    run_podman info --format '{{ .Host.BuildahVersion }}'
+    version=$output
+    run_podman build -t build_test $tmpdir
+    run_podman image inspect --format '{{.Labels}}' build_test
+    is "$output" ".*$version" "failed to find buildah version in labels"
```
edsantiago, line 327: I love tests as much as the next person[*], but why add a new test here? `podman build` is expensive, and we already have the huge monster "podman build - workdir, cmd, env, label" test just a few lines above. Why not just add a buildah-version check to the 'tests=' table?

 [*] ok, not true. I might like them more than the next person.

edsantiago: Friendly suggestion:
```diff
diff --git a/test/system/070-build.bats b/test/system/070-build.bats
index 0741357ed..064cad965 100644
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -221,6 +221,11 @@ EOF
     run_podman run --rm build_test pwd
     is "$output" "$workdir" "pwd command in container"

+    # Determine buildah version, so we can confirm it gets into Labels
+    run_podman info --format '{{ .Host.BuildahVersion }}'
+    is "$output" "[1-9][0-9.-]\+" ".Host.BuildahVersion is reasonable"
+    buildah_version=$output
+
     # Confirm that 'podman inspect' shows the expected values
     # FIXME: can we rely on .Env[0] being PATH, and the rest being in order??
     run_podman image inspect build_test
@@ -239,6 +244,7 @@ Cmd[0]             | /bin/mydefaultcmd
 Cmd[1]             | $s_echo
 WorkingDir         | $workdir
 Labels.$label_name | $label_value
+Labels.\"io.buildah.version\" | $buildah_version
 "

     parse_table "$tests" | while read field expect; do
```
Aligning the columns (pipe symbol) is left as an exercise for the reader, should he wish to do so.

This was much, much harder than I expected it to be (because of the dots in the label name, and `jq` wanting to interpret them)

rhatdan: SGTM

TomSweeneyRedHat:
LGTM

rhatdan:

ashley-cui (commented):
LGTM

PR #8328, opened 2020-11-12
[CI:DOCS] Add missing --now in systemctl start command

FreedomBen:
If you don't use --now or also run systemctl --user start podman.socket
then the socket won't start until a reboot.  Much easier to just start
the socket at the same time as enabling it.

### DIFF
--- a/docs/tutorials/mac_win_client.md
+++ b/docs/tutorials/mac_win_client.md
@@ -29,9 +29,9 @@ $ brew install podman
 
 ### Enable the Podman service on the server machine.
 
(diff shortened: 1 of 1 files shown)


mheon:
@FreedomBen Can you add a signoff to your commit? `git commit --amend -s` and then force-pushing should do the job.

/approve

FreedomBen:
@mheon no problem, done!

rhatdan:
/approve
/lgtm
Thanks @FreedomBen

Issue #8317, opened 2020-11-12
Commands Table format messed up in docs.podman.io


TomSweeneyRedHat:
I've not idea off the top of my head how to fix this, but did want to note that the table under the Commands heading is one long continuous line.  http://docs.podman.io/en/latest/markdown/podman.1.html#commands  

Something has obviously gone wrong translating the md to html, but I know not where and haven't divd in.

rhatdan:
Perhaps one of our community people have an idea?
@Luap99 @zhangguanzhang Any ideas?

Luap99:
@TomSweeneyRedHat I also noticed this.

I found this Issue: https://github.com/readthedocs/recommonmark/issues/3

This could be a possible solution: https://pypi.org/project/sphinx-markdown-tables/

Luap99:
Fixed by #8335

PR #8331, opened 2020-11-13
[CI:DOCS] Fix Commands table

TomSweeneyRedHat:
The markdown table that lists all of the available commands
on the toplevel Podman man page does not resolve correctly
when converted to html and moved to docs.podman.io.

So as html is valid in a markdown document, I've converted
the table to use html instead in hope that it will translate better.

### DIFF
--- a/docs/source/Commands.rst
+++ b/docs/source/Commands.rst
@@ -3,7 +3,7 @@
 Commands
 ========
 
-:doc:`Podman <markdown/podman.1>` (Pod Manager) Global Options
+:doc:`Podman <markdown/podman.1>` (Pod Manager) Global Options, Environment Variables, Exit Status, Files, and more
 
 :doc:`attach <markdown/podman-attach.1>` Attach to a running container
 
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -199,62 +199,232 @@ the exit codes follow the `chroot` standard, see below:
 
 ## COMMANDS
 
-| Command                                          | Description                                                                 |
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
@edsantiago PTAL.  I have a thought that you have a test that runs through this table to validate man pages.  Please let me know if this change messes that up or not.

TomSweeneyRedHat:
Addresses: #8317

edsantiago:
Yeah, that's going to fail Validate. BTW there are several other man pages that include tables and might also be having problems: basically, any subcommand (podman image X, podman container X).

edsantiago:
So, if I may add my humble two cents: I think this is the wrong approach. The new form is unmaintainable, I'd even go as far as saying anti-maintainable. It is the opposite of what markdown is intended to provide. It's also not going to work that well with plaintext manpages. 

IIUC the problem we are trying to solve is that some HTML pages look icky. Is that correct? If so, I would encourage us to take a step back, not break the markdown pages which have been working remarkably well for us so far, and try to look for a solution closer to the problem itself. How are the web pages in question generated? What are the tools? Do they have an option for generating HTML tables from markdown ones? (i.e. this **has** to be an already-solved problem that we can piggyback on). If there's no way: can we insert filters? Can we add a sed or perl filter that translates markdown tables into HTML tags? 

I'm reluctant to spend my evening on this, and I have a hard commitment tomorrow. I'm sorry. I do hope we can find a less disruptive, more maintainable solution.

Luap99:
I opened #8335 which address the render problem so we don't have to change the doc format.

rhatdan:
Closing for https://github.com/containers/podman/pull/8335

PR #8335, opened 2020-11-13
[CI:DOCS] Fix markdown tables on docs.podman.io

Luap99:
Sphinx with recommonmark cannot render markdown tables at all.
There is a python package called `sphinx-markdown-tables` which
adds the markdown table support to recommonmark.
https://pypi.org/project/sphinx-markdown-tables/

By utilising this package we don't have to change our doc format.


rhatdan:
LGTM
@TomSweeneyRedHat @edsantiago @vrothberg @giuseppe PTAL

edsantiago:
@Luap99 do you have a way we can see results?

edsantiago:
PS that's not to say I don't believe it'll work, I'm just a careful reviewer who likes to be thorough. Nice work, thank you!

Luap99:
@edsantiago 

Install the dependencies 
```
dnf install python-sphinx python-recommonmark
pip install sphinx-markdown-tables
```

Build the site
```
cd docs
make html
```
Watch the result
```
cd build/html/markdown/
xdg-open podman.1.html
```

TomSweeneyRedHat:
@Luap99 Thanks for putting this together.  I saw your comment about sphinx in the issue, but didn't quickly see where to drop the changes in our site.  I speak HTML (and being a weird InDUHvidual) prefer it to MD, so the other change was quick and took the middle man out for me.  However if this works, I'm very happy to take it as clonking @edsantiago 's verification script alone was not good.

TomSweeneyRedHat:
FWIW, I'd to do the following to get the dependencies setup.

```
dnf install python-sphinx pip3
pip3 install python-recommonmark
pip3 install sphinx-markdown-tables
```

TomSweeneyRedHat:
but then the build failed in my sandbox....

rhatdan:
Needs a rebase.

rhatdan:
Fixes: https://github.com/containers/podman/issues/8317

TomSweeneyRedHat:
@Luap99 this did fix the table structure thanks!
However, the side bar on the home page is now a little askew.  The "Join Community" button now has a return in it and it doesn't fit in the purple box now.  I can't imagine this change would cause that, but I don't think we've had any other functional changes there recently.  Thoughts?

edsantiago:
@TomSweeneyRedHat I assume by "home page" you mean [https://podman.io/](https://podman.io/) ? (I see a "Join Community" button there in a purple box). If so, can you paste a screen shot and perhaps describe your browser? On my setup (Firefox) I can't get the "Join Community" text to spill out of the purple box no matter how narrow I make my browser window. (Snapshot below). But before spending more time on this, let's please all make sure we're talking about the same URL.
podman-join-community

TomSweeneyRedHat:
@edsantiago yes, that's the page, podman.io.  
I just pulled the page up again to get a screen shot of it to add here, and it's displaying fine now in my newly updated Firefox.  Prior il looked like there was a carriage return after the word "Join" and then "Community" was underneath is partially in the purple and mostly in the black.

So I think at this point it was just some kind of cache issue with Firefox at the time, so just ignore it, and sorry for the fire drill!

PR #8339, opened 2020-11-14
Wrap missing container errors with container ID

rhatdan:
While playing around with podman system df, I saw that my container
database was in bad state.  Basically podman new about containers
that were no longer in container/storage.

The podman system df was just erroring out early stating "container does not exist"
with no indicator of which container.

This Patch wraps the podman system df errors to indicate which container does not exist.
It also logs errors on containers that get into this state, but continues on to work on
all containers.

### DIFF
--- a/pkg/domain/infra/abi/system.go
+++ b/pkg/domain/infra/abi/system.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/podman/v2/pkg/rootless"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/podman/v2/utils"
+	"github.com/containers/storage"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


mheon:
The OCI runtime container deletion should not be returning c/storage errors
- it shouldn't be interacting with c/storage at all, only calling OCI
runtime delete?

rhatdan:
Yes you are right.  I am not sure why I got an error when trying to delete. and now I cleaned up my repository.

rhatdan:
This one is ready to go in.

mheon:
LGTM

ashley-cui (commented):
LGTM

QiWang19 (commented):
/lgtm

Issue #8341, opened 2020-11-14
Volume filtering works differently and misses existing label


afbjorklund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Volume filtering works differently in podman, when compared to docker.

**Steps to reproduce the issue:**

1. minikube start --driver podman

2. sudo -n podman volume ls --filter label=name.minikube.sigs.k8s.io=minikube

3.

**Describe the results you received:**

```
```

**Describe the results you expected:**

```
DRIVER      VOLUME NAME
local       minikube
```

**Additional information you deem important (e.g. issue happens only occasionally):**

The --filter doesn't work with label:

`minikube start --driver podman`

```console
$ sudo -n podman volume ls --filter label=name.minikube.sigs.k8s.io=minikube
$ sudo -n podman volume ls 
DRIVER      VOLUME NAME
local       minikube
$ sudo -n podman volume inspect minikube
[
     {
          "Name": "minikube",
          "Driver": "local",
          "Mountpoint": "/var/lib/containers/storage/volumes/minikube/_data",
          "CreatedAt": "2020-11-14T15:13:11.855745076+01:00",
          "Labels": {
               "created_by.minikube.sigs.k8s.io": "true",
               "name.minikube.sigs.k8s.io": "minikube"
          },
          "Scope": "local",
          "Options": {
               
          },
          "UID": 0,
          "GID": 0,
          "Anonymous": false
     }
]
```

It works OK with the docker driver.

`minikube start --driver docker`

```console
$ docker volume ls --filter label=name.minikube.sigs.k8s.io=minikube
DRIVER              VOLUME NAME
local               minikube
$ docker volume inspect minikube
[
    {
        "CreatedAt": "2020-11-14T15:22:40+01:00",
        "Driver": "local",
        "Labels": {
            "created_by.minikube.sigs.k8s.io": "true",
            "name.minikube.sigs.k8s.io": "minikube"
        },
        "Mountpoint": "/var/lib/docker/volumes/minikube/_data",
        "Name": "minikube",
        "Options": {},
        "Scope": "local"
    }
]


```
**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
(paste your output here)
```

[...]

PR #8345, opened 2020-11-14
Add support for volume ls --filter label=key=value

afbjorklund:
Supposed to be able to search for labels with a given value.

Previously it meant searching for label key and label value:

--filter label=key --filter label=value

Add some documentation and integration tests for it as well.

Fixes: #8341


Luap99 (commented):
LGTM

PR #8346, opened 2020-11-14
Improve the shell completion api

Luap99:
One main advantage of the new shell completion logic is that
we can easly parse flags and adjust based on the given flags
the suggestions. For example some commands accept the
`--latest` flag only if no arguments are given.

This commit implements this logic in a simple maintainable way
since it reuses the already existing `Args` function in the
cmd struct.

I also refactored the `getXXX` function to match based on the
namei/id which could speed up the shell completion with many
containers, images, etc...

I also added the degraded status to the valid pod status
filters which was implemented in #8081.

### DIFF
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


Luap99:
Basically typing `podman start --latest [TAB]` won't recommend container names while `podman start [TAB]` does.
This applies to all command which are using a `Args` function.

rhatdan:
LGTM
/approve

rhatdan:
@edsantiago PTAL

Luap99:
I pushed a small update to handle the `ExactArgs(2)` or `MinimumNArgs(2)` functions correct. Otherwise `podman cp [TAB]` would have failed.

cmd/podman/common/completion.go
```diff
@@ -159,76 +216,144 @@ func getNetworks() ([]string, cobra.ShellCompDirective) {
...
+	// We have to append toComplete to the args otherwise the
+	// argument count would not match the expected behavior
+	if err := cmd.Args(cmd, append(args, toComplete)); err != nil {
+		var a, b int
+		var s string
+		// Special case if we use ExactArgs(2) or MinimumNArgs(2),
+		// They will error if we try to complete the first arg.
+		// Lets parse the common error and compare if we have less args than
+		// required. In this case we are fine and should provide completion.
+		_, err = fmt.Sscanf(err.Error(), "%s %d arg(s), received %d", &s, &a, &b)
```
edsantiago, line 238: `s` is never used; I'm surprised Go doesn't warn about that. And while you're changing this, what would you think of changing `a` and `b` to `need`/`got` or something similar?

cmd/podman/common/completion.go
```diff
@@ -22,62 +23,109 @@ var (
...
 	listOpts.All = true
+	listOpts.Pod = true
 
-	if status != "all" {
-		listOpts.Filters = map[string][]string{"status": {status}}
+	/* 	The api doesn't handle several different statuses correct see:
+		https://github.com/containers/podman/issues/8344
+
+	if len(status) > 0 {
+		listOpts.Filters["status"] = statuses
```
edsantiago, line 38: I'm a little uncomfortable with a commented-out code block of this size: first, it's syntactically wrong (`status` should be `statuses`); second, it may grow even staler as time goes by. I appreciate the intention, but would it be better spelled out as a more generic TODO comment instead? (Same applies below)

edsantiago (commented):
The refactoring LGTM - I'm finding situations where it doesn't work as I expected (`podman top`, `podman container inspect -l`) but it's still sooooo much better than what we had before. This can be refined over time. Thanks for your initiative in taking this on.

Two very minor suggestions, left up to your discretion.

Luap99:
@edsantiago Updated. My error parsing didn't really work as expected. I have to clean the differences in the error message first. I thought `%s` could handle that.

Luap99:
Not all commands use a meaningful `Args` function. e.g. `podman top`

I have to go through them one by one to see were I can improve this.

edsantiago:
Yes, that makes sense, and it can be fixed over time. No rush, this is already very helpful.

edsantiago:
The tight coupling to (and massaging of) cobra messages bothers me a little, but I can't think of a better way to do it. On the whole this LGTM but I'd like confirmation review from at least one more @containers/podman-maintainers team member.

PR #8348, opened 2020-11-16
[WIP] Add pre-dump support for checkpoint

chen-zhuohan:
We can migrate container to other machine with less downtime if we can migrate container with two step, first migrating the most memory images, the second migrating the left.
For this purpose, we push the new pr.

### DIFF
--- a/cmd/podman/containers/checkpoint.go
+++ b/cmd/podman/containers/checkpoint.go
@@ -47,6 +47,8 @@ func init() {
 	flags.BoolVarP(&checkpointOptions.Keep, "keep", "k", false, "Keep all temporary checkpoint files")
 	flags.BoolVarP(&checkpointOptions.LeaveRunning, "leave-running", "R", false, "Leave the container running after writing checkpoint to disk")
 	flags.BoolVar(&checkpointOptions.TCPEstablished, "tcp-established", false, "Checkpoint a container with established TCP connections")
+	flags.BoolVar(&checkpointOptions.PreDump, "pre-dump", false, "Dump container's memory information only, leave the container running after this")
+	flags.BoolVar(&checkpointOptions.WithPrevious, "with-parent", false, "Checkout container with pre-dump images")
 	flags.BoolVarP(&checkpointOptions.All, "all", "a", false, "Checkpoint all running containers")
 	flags.StringVarP(&checkpointOptions.Export, "export", "e", "", "Export the checkpoint image to a tar.gz")
 	flags.BoolVar(&checkpointOptions.IgnoreRootFS, "ignore-rootfs", false, "Do not include root file-system changes when exporting")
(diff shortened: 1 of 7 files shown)


vrothberg (commented):
@adrianreber PTAL

adrianreber:
@chen-zhuohan Very happy to see that you are trying to integrate pre-dump support into Podman. I see that you added the label [WIP] so I am not yet commenting on the code as it still seems to be missing a couple of things.

Generally I think it does not make sense to introduce pre-dump checkpointing without the support to restore it.

To understand what you are trying to do it would be important to have much more detailed commit messages and the corresponding documentation around checkpoint/restore also would need to be updated.

I would also prefer if the help text of '--pre-dump' uses the words checkpointing instead of dumping. CRIU uses the word 'dump' a lot, but I think it would be important to have consistent wording in the Podman user interface which should always use checkpointing instead of dumping. I guess the option '--pre-dump' can keep the name, but the help text should not use it.

I have been working on pre-dump and post-dump a  lot in CRIU and runc and I am happy to help with any questions about it.

It is also necessary to detect if you are running with runc oder crun. crun currently does not yet support pre-dump in contrast to runc.

I also just recently talked with @rst0git about this and @rst0git mentioned criu-image-streamer. Not sure if you looked into this.

adrianreber:
Just saw, that the ticket is closed. Sorry for the long comment then.

YangdaXing:
> @chen-zhuohan Very happy to see that you are trying to integrate pre-dump support into Podman. I see that you added the label [WIP] so I am not yet commenting on the code as it still seems to be missing a couple of things.
> 
> Generally I think it does not make sense to introduce pre-dump checkpointing without the support to restore it.
> 
> To understand what you are trying to do it would be important to have much more detailed commit messages and the corresponding documentation around checkpoint/restore also would need to be updated.
> 
> I would also prefer if the help text of '--pre-dump' uses the words checkpointing instead of dumping. CRIU uses the word 'dump' a lot, but I think it would be important to have consistent wording in the Podman user interface which should always use checkpointing instead of dumping. I guess the option '--pre-dump' can keep the name, but the help text should not use it.
> 
> I have been working on pre-dump and post-dump a lot in CRIU and runc and I am happy to help with any questions about it.
> 
> It is also necessary to detect if you are running with runc oder crun. crun currently does not yet support pre-dump in contrast to runc.
> 
> I also just recently talked with @rst0git about this and @rst0git mentioned criu-image-streamer. Not sure if you looked into this.

@adrianreber，Hi, I'm the chen-zhuohan‘s classmate.

 We are interested in containers living migration and we want to reduce the amount of data transferred in order to reduce the migration time. 

We think the `pre-dump` option, supported by the criu, can help us to track the memory. So we want to integrate the pre-dump support and test the restore support to see whether it can work. 

We will be appreciated if you can give us some suggestions whether on container migration or source code modification.

PR #8355, opened 2020-11-16
add network connect|disconnect compat endpoints

baude:
this enables the ability to connect and disconnect a container from a
given network. it is only for the compatibility layer. some code had to
be refactored to avoid circular imports.

additionally, tests are being deferred temporarily due to some
incompatibility/bug in either docker-py or our stack.

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 	"syscall"
 	"time"
@@ -20,6 +21,7 @@ import (
 	cnitypes "github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containers/podman/v2/libpod/define"
+	"github.com/containers/podman/v2/libpod/network"
 	"github.com/containers/podman/v2/pkg/errorhandling"
 	"github.com/containers/podman/v2/pkg/netns"
 	"github.com/containers/podman/v2/pkg/rootless"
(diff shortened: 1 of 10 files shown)


libpod/networking_linux.go
```diff
@@ -981,3 +983,139 @@ func (w *logrusDebugWriter) Write(p []byte) (int, error) {
...
 }
+
+// DisconnectContainerFromNetwork removes a container from its CNI network
+func (r *Runtime) DisconnectContainerFromNetwork(nameOrID, netName string, force bool) error {
+	ctr, err := r.LookupContainer(nameOrID)
+	if err != nil {
+		return err
+	}
+
+	networks, err := ctr.networksByNameIndex()
```
mheon, line 994: Could I convince you to take everything from here on, and turn it into `func (c *Container) NetworkDisconnect(...)` and then have this function call that?

libpod/networking_linux.go
```diff
@@ -981,3 +983,139 @@ func (w *logrusDebugWriter) Write(p []byte) (int, error) {
...
+}
+
+// ConnectContainerToNetwork connects a container to a CNI network
+func (r *Runtime) ConnectContainerToNetwork(nameOrID, netName string, aliases []string) error {
+	ctr, err := r.LookupContainer(nameOrID)
+	if err != nil {
+		return err
+	}
+
+	networks, err := ctr.networksByNameIndex()
```
mheon, line 1048: Same as above, splitting this bit on into a per-container function? I'd like to be able to call this on an existing container

baude: @mheon has agreed to allow me to defer this until my next PR

libpod/networking_linux.go
baude: @mheon has agreed to allow me to defer this until my next PR

baude:
i need a LGTM and /lgtm

jwhonce:
LGTM

PR #8363, opened 2020-11-16
Add an option to control if play kube should start the pod

AlbanBedel:
Having play kube start the pod is not always appropriate, one might
for example like to have the pod running as a set of systemd services.
This new option allow to skip starting the pod.

### DIFF
--- a/docs/source/markdown/podman-play-kube.1.md
+++ b/docs/source/markdown/podman-play-kube.1.md
@@ -58,6 +58,10 @@ Suppress output information when pulling images
 
 Directory path for seccomp profiles (default: "/var/lib/kubelet/seccomp"). (Not available for remote commands)
 
+#### **--start**=*true|false*
+
+Start the pod after creating it, set to false to only create it.
+
 #### **--tls-verify**=*true|false*
 
 Require HTTPS and verify certificates when contacting registries (default: true). If explicitly set to true,
--- a/cmd/podman/play/kube.go
+++ b/cmd/podman/play/kube.go
@@ -22,6 +22,7 @@ type playKubeOptionsWrapper struct {
 
 	TLSVerifyCLI   bool
 	CredentialsCLI string
+	StartCLI       bool
 }
 
 var (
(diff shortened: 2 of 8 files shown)


docs/source/markdown/podman-play-kube.1.md
```diff
@@ -64,6 +64,10 @@ Require HTTPS and verify certificates when contacting registries (default: true)
 then TLS verification will be used. If set to false, then TLS verification will not be used. If not specified,
 TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf.
 
+#### **--create-only**, **-c**
```
TomSweeneyRedHat, line 67: move this up to line 38 so the options will be listed alphabetically (with the exception of help)

TomSweeneyRedHat:
@AlbanBedel  you've a lint error and I don't think you signed your PR which our CI requires (`git commit --amend -s` will take care of the signature).

Otherwise the change LGTM

Luap99 (commented):
This also has to be wired into the api and remote client.

Server logic here: pkg/api/handlers/libpod/play.go
Client logic here: pkg/bindings/play/play.go
API docs here: pkg/api/server/register_play.go

rhatdan:
How about the option be --now, similar to systemctl enable service --now.

The default is true, to be backwards compatible and --now=false, gives you this functionality.

AlbanBedel:
> How about the option be --now, similar to systemctl enable service --now.
> 
> The default is true, to be backwards compatible and --now=false, gives you this functionality.

In the context of a command named `play` I would find `--now` a bit confusing without further explanations. What about `--start-now=false`?

rhatdan:
I am not a big fan of options with a "-" inside of them, which is one of the reasons, I wanted to make the change.

AlbanBedel:
In this new version:
* Renamed the option to `start` as I find `now` to not be self explanatory enough
* Added the new option to the client, server and API documentation
* Fixed the lint errors
* Moved the man page entry to the right position
* Reworked the commit message

@TomSweeneyRedHat The commit has my SOB, I just removed it from the PR description.

rhatdan:
LGTM
Thanks @AlbanBedel

TomSweeneyRedHat:
The ubuntu-19 root tests aren't happy.  Not sure if it's a known issue or not.

TomSweeneyRedHat:
LGTM
assuming happy tests

Issue #8344, opened 2020-11-14
status filters for podman ps and podman pod ps should be inclusive


Luap99:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I want to filter the output for `podman ps`  to only contain created and exited containers for example.

Since filters are exclusive I cannot do this.

**Steps to reproduce the issue:**

1. `podman create alpine`

2. `podman ps --filter status=created,status=exited`

**Describe the results you received:**
empty output

**Describe the results you expected:**
show all created and exited containers

**Output of `podman version`:**

```
$ bin/podman version
Version:      2.2.0-dev
API Version:  2.0.0
Go Version:   go1.14.10
Git Commit:   4eb9c28433f8aa1827f844881efb1ae1d15b0860
Built:        Sat Nov 14 16:50:57 2020
OS/Arch:      linux/amd64
```

zhangguanzhang:
libpod/runtime_ctr.go:804-808
```golang
	for _, ctr := range ctrs {
		include := true
		for _, filter := range filters {
			include = include && filter(ctr)
		}

		if include {
			ctrsFiltered = append(ctrsFiltered, ctr)
		}
	}
```
This code determines that if one of the filter conditions is not met, it will not match.

rhatdan:
Do you know how Docker handles this?

Luap99:
It does work with docker
```
# docker ps --filter status=created --filter status=exited 
CONTAINER ID        IMAGE               COMMAND             CREATED              STATUS                       PORTS               NAMES
42aec3a7874a        alpine              "sleep 1000"        58 seconds ago       Exited (137) 5 seconds ago                       affectionate_shaw
be48cb2a08aa        alpine              "/bin/sh"           About a minute ago   Created                                          vibrant_black
20bcf202d6d9        alpine              "/bin/sh"           About a minute ago   Created                                          pensive_clarke
```

rhatdan:
Since the issue reported is fixed.  I am closing this issue.  We will follow along with #8376

PR #8368, opened 2020-11-17
Fixes: status filters for podman ps and podman pod ps should be inclusive

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/8344#issuecomment-728830411

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -802,13 +802,13 @@ func (r *Runtime) GetContainersWithoutLock(filters ...ContainerFilter) ([]*Conta
 	ctrsFiltered := make([]*Container, 0, len(ctrs))
 
 	for _, ctr := range ctrs {
-		include := true
 		for _, filter := range filters {
-			include = include && filter(ctr)
-		}
-
-		if include {
-			ctrsFiltered = append(ctrsFiltered, ctr)
+			// status filters for podman ps and podman pod ps should be inclusive
+			// issue: https://github.com/containers/podman/issues/8344
+			if filter(ctr) {
+				ctrsFiltered = append(ctrsFiltered, ctr)
+				break
+			}
 		}
 	}
 
--- a/test/e2e/ps_test.go
+++ b/test/e2e/ps_test.go
@@ -303,6 +303,22 @@ var _ = Describe("Podman ps", func() {
 		Expect(psAll.OutputToString()).To(Equal(psFilter.OutputToString()))
 	})
 
+	It("podman status filters should be inclusive", func() {
+		ctr := podmanTest.Podman([]string{"run", "-t", "-i", ALPINE, "ls", "/"})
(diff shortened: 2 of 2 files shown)


rhatdan:
/approve
LGTM
@Luap99 PTAL

libpod/runtime_ctr.go
```diff
@@ -802,13 +802,13 @@ func (r *Runtime) GetContainersWithoutLock(filters ...ContainerFilter) ([]*Conta
...
-			include = include && filter(ctr)
-		}
-
-		if include {
-			ctrsFiltered = append(ctrsFiltered, ctr)
+			// status filters for podman ps and podman pod ps should be inclusive
+			// issue: https://github.com/containers/podman/issues/8344
+			if filter(ctr) {
+				ctrsFiltered = append(ctrsFiltered, ctr)
+				continue
```
vrothberg, line 810: Shouldn't this be a `break` to stop iterating over the filters?

zhangguanzhang: got it

Luap99 (commented):
This does not address the `podman pod ps` part.

Luap99:
Also doing more testing with docker some filters appear to be inclusive (status) while other are exclusive (label). I need to see if I find something where this is documented.

zhangguanzhang:
> Also doing more testing with docker some filters appear to be inclusive (status) while other are exclusive (label). I need to see if I find something where this is documented.

got it

test/e2e/ps_test.go
```diff
@@ -303,6 +303,22 @@ var _ = Describe("Podman ps", func() {
...
 
+	It("podman status filters should be inclusive", func() {
+		ctr := podmanTest.Podman([]string{"run", "-t", "-i", ALPINE, "ls", "/"})
+		ctr.WaitWithDefaultTimeout()
+		Expect(ctr.ExitCode()).To(Equal(0))
+
+		ctr2 := podmanTest.Podman([]string{"container", "create", ALPINE})
+		ctr2.WaitWithDefaultTimeout()
+		Expect(ctr2.ExitCode()).To(Equal(0))
+
```
vrothberg, line 314: Can we add another "exited" container? I think we need one item that doesn't match the filter.

Luap99:
The docker logic is quit complex: https://github.com/moby/moby/blob/5c10ea6ae80589c2d56d618df54927e92b0b0136/daemon/list.go#L410

In general filters with the same key are inclusive, e.g. `--filter name=con1 --filter name=con2` returns con1 and con2. The only exception to this seems to be `label` which is exclusive.
Filters with different keys are exclusive, e.g. `--filter name=con1 --filter status=created` will only return con1 if it has the status created.

zhangguanzhang:
I took a look, this logic is too complicated for me🙄。May need someone else to implement this feature

Luap99:
@zhangguanzhang I will.

PR #8370, opened 2020-11-17
Remove some more excessive wrapping and stuttering

rhatdan:
Stop over wrapping API Calls

The API calls will return an appropriate error, and this wrapping
just makes the error message look like it is stuttering and a
big mess.

### DIFF
--- a/cmd/podman/system/reset.go
+++ b/cmd/podman/system/reset.go
@@ -13,7 +13,7 @@ import (
 	"github.com/containers/podman/v2/cmd/podman/validate"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/domain/infra"
-	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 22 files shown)


cmd/podman/system/reset.go
```diff
@@ -57,7 +57,7 @@ WARNING! This will remove:
 Are you sure you want to continue? [y/N] `)
 		answer, err := reader.ReadString('\n')
 		if err != nil {
-			fmt.Println(errors.Wrapf(err, "error reading input"))
```
vrothberg, line 60: That'll change from stdout to stderr.

I am okay with it but want to highlight for others who might think differently.

vrothberg (commented):
Just one nit but LGTM.

I like your ongoing error crusade

cmd/podman/system/reset.go
rhatdan: I think this is correct, since it is reporting and error.

mheon:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
/lgtm
/hold
will let @mheon cancel in case there's any hold for the rc

mheon:
/hold cancel

Should be good to go.

PR #8375, opened 2020-11-17
container cgroup path

vrothberg:
Before querying for a container's cgroup path, make sure that the
container is synced.  Also make sure to error out if the container
isn't running.

@mheon PTAL

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -921,13 +921,33 @@ func (c *Container) CgroupManager() string {
 	return cgroupManager
 }
 
-// CGroupPath returns a cgroups "path" for a given container.
+// CGroupPath returns a cgroups "path" for the given container.
+// Note that the container must be running.  Otherwise, an error
+// is returned.
 func (c *Container) CGroupPath() (string, error) {
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+		if err := c.syncContainer(); err != nil {
+			return "", errors.Wrapf(err, "error updating container %s state", c.ID())
+		}
+	}
+	return c.cGroupPath()
+}
+
+// cGroupPath returns a cgroups "path" for the given container.
+// Note that the container must be running.  Otherwise, an error
+// is returned.
+// NOTE: only call this when owning the container's lock.
+func (c *Container) cGroupPath() (string, error) {
 	if c.config.NoCgroups || c.config.CgroupsMode == "disabled" {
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

libpod/container.go
```diff
@@ -906,12 +906,26 @@ func (c *Container) CgroupManager() string {
...
 func (c *Container) CGroupPath() (string, error) {
 	if c.config.NoCgroups || c.config.CgroupsMode == "disabled" {
 		return "", errors.Wrapf(define.ErrNoCgroups, "this container is not creating cgroups")
 	}
 
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+		if err := c.syncContainer(); err != nil {
+			return "", errors.Wrapf(err, "error updating container %s state", c.ID())
```
TomSweeneyRedHat, line 921: Would it be better to add to or replace the message with the action we're taking here?  Perhaps:
```suggestion
			return "", errors.Wrapf(err, "error updating container %s state when retrieving the path for CGroup ", c.ID())
```

TomSweeneyRedHat:
One question otherwise
LGTM

libpod/container.go
vrothberg: @mheon WDYT? AFAICS this error message is used across many places. But I do see wisdom in customizing them a bit to discriminate one from another.

mheon: I don't mind - it doesn't hurt - but if Go was a more expressive language, this entire block of code would be a macro so we wouldn't have to copy and paste everywhere.

mheon: (In which case, it really would be guaranteed to be the same error message everywhere...)

mheon:
Tests are red

rhatdan:
@vrothberg Please rebase.

vrothberg:
> @vrothberg Please rebase.

Thanks for the reminder. I almost forgot about the PR.

vrothberg:
Rebased. @mheon, I think that's a good one for 2.2.1 as well?

mheon:
Sure, once tests go green.

vrothberg:
Trying to wrap my head around how this PR could cause:
```
[+3742s] Error: error configuring CNI network plugin: could not create new watcher too many open files
```
Cannot reproduce locally.

vrothberg:
@mheon green :)

PR #8376, opened 2020-11-17
Align the podman ps --filter behavior with docker

### LINKED ISSUE #8344: status filters for podman ps and podman pod ps should be inclusive

Luap99:
All of our filters worked exclusive resulting in `--filter status=created --filter status=exited` to return nothing.

In docker filters with the same key work inclusive with the only exception being `label` which is exclusive. Filters with different keys always work exclusive.

This PR aims to match the docker behavior with podman.

Fixes #8344

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -98,14 +98,6 @@ func checkFlags(c *cobra.Command) error {
 	if listOpts.Last >= 0 && listOpts.Latest {
 		return errors.Errorf("last and latest are mutually exclusive")
 	}
-	// Filter on status forces all
-	for _, filter := range filters {
-		splitFilter := strings.SplitN(filter, "=", 2)
-		if strings.ToLower(splitFilter[0]) == "status" {
-			listOpts.All = true
-			break
-		}
-	}
 	// Quiet conflicts with size and namespace and is overridden by a Go
 	// template.
 	if listOpts.Quiet {
--- a/libpod/filters/containers.go
+++ b/libpod/filters/containers.go
@@ -1,7 +1,6 @@
 package lpfilters
 
 import (
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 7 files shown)


Luap99:
I will do the same for `podman pod ps` next but this can go in a different PR

rhatdan:
LGTM

libpod/filters/containers.go
```diff
@@ -11,101 +10,133 @@ import (
...
 		}, nil
 	case "before":
-		ctr, err := r.LookupContainer(filterValue)
-		if err != nil {
-			return nil, errors.Errorf("unable to find container by name or id of %s", filterValue)
+		var createTime time.Time
+		for _, filterValue := range filterValues {
+			ctr, err := r.LookupContainer(filterValue)
+			if err != nil {
+				return nil, errors.Errorf("unable to find container by name or id of %s", filterValue)
```
saschagrunert, line 117:
```suggestion
				return nil, errors.Wrapf(err, "unable to find container by name or id of %s", filterValue)
```

libpod/filters/containers.go
```diff
@@ -11,101 +10,133 @@ import (
...
 		}, nil
 	case "since":
-		ctr, err := r.LookupContainer(filterValue)
-		if err != nil {
-			return nil, errors.Errorf("unable to find container by name or id of %s", filterValue)
+		var createTime time.Time
+		for _, filterValue := range filterValues {
+			ctr, err := r.LookupContainer(filterValue)
+			if err != nil {
+				return nil, errors.Errorf("unable to find container by name or id of %s", filterValue)
```
saschagrunert, line 133:
```suggestion
				return nil, errors.Wrapf(err, "unable to find container by name or id of %s", filterValue)
```

saschagrunert (commented):
Just two nits, otherwise LGTM

cmd/podman/containers/ps.go
```diff
@@ -98,14 +98,6 @@ func checkFlags(c *cobra.Command) error {
 	if listOpts.Last >= 0 && listOpts.Latest {
 		return errors.Errorf("last and latest are mutually exclusive")
 	}
-	// Filter on status forces all
```
mheon, line 101: It's not immediately clear what's replacing this - do we still ensure that all is set?

Luap99: Yes it is checked later in the code path again.
https://github.com/containers/podman/blob/a6f7afd41729edbe0d8cbf21bab5b475c6a13f9e/pkg/ps/ps.go#L40-L44

libpod/filters/containers.go
Luap99: Actually I can just return `err` here. The error from `LookupContainer` is `no container with name or ID %s found: no such container`. Wrapping this wouldn't provide more information

cmd/podman/containers/ps.go
mheon: Wow. Wonder how that slipped in. OK, thanks!

rhatdan:
Thanks @Luap99 
/lgtm

PR #8379, opened 2020-11-17
Remove build \!remote flags from test phase 2

rhatdan:
Add some more tests, document cases where remote will not work
Add FIXMEs for tests that should work on podman-remote but currently
do not.

### DIFF
--- a/pkg/api/handlers/compat/networks.go
+++ b/pkg/api/handlers/compat/networks.go
@@ -289,25 +289,40 @@ func CreateNetwork(w http.ResponseWriter, r *http.Request) {
 
 func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
-	config, err := runtime.GetConfig()
-	if err != nil {
-		utils.InternalServerError(w, err)
+	ic := abi.ContainerEngine{Libpod: runtime}
+
+	query := struct {
+		Force bool `schema:"force"`
+	}{
+		// This is where you can override the golang default value for one of fields
+	}
+
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
 		return
 	}
+
+	options := entities.NetworkRmOptions{
+		Force: query.Force,
+	}
+
 	name := utils.GetName(r)
-	exists, err := network.Exists(config, name)
+	reports, err := ic.NetworkRm(r.Context(), []string{name}, options)
 	if err != nil {
-		utils.InternalServerError(w, err)
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "remove Network create"))
 		return
 	}
-	if !exists {
-		utils.Error(w, "network not found", http.StatusNotFound, define.ErrNoSuchNetwork)
+	report := reports[0]
+	if report.Err != nil {
+		if errors.Cause(report.Err) == define.ErrNoSuchNetwork {
+			utils.Error(w, "network not found", http.StatusNotFound, define.ErrNoSuchNetwork)
+			return
+		}
+		utils.InternalServerError(w, report.Err)
 		return
 	}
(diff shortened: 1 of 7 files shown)


edsantiago:
I see that someone pressed Re-run. I don't think that's going to help: these look like very real errors: [lots in ubuntu](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6638347772755968/html/int-remote-ubuntu-20-root-host.log.html) and only [push in fedora](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5512447865913344/html/int-remote-fedora-32-root-host.log.html)

rhatdan:
@edsantiago I found a race condition on podman-remote network rm --force, that is causing the network remove failures.

rhatdan:

pkg/api/handlers/compat/networks.go
```diff
@@ -289,25 +289,40 @@ func CreateNetwork(w http.ResponseWriter, r *http.Request) {
...
-		return
-	}
-	if !exists {
-		utils.Error(w, "network not found", http.StatusNotFound, define.ErrNoSuchNetwork)
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "remove Network create"))
 		return
 	}
-	if err := network.RemoveNetwork(config, name); err != nil {
-		utils.InternalServerError(w, err)
+	report := reports[0]
```
mheon, line 316: Should do a length check to make sure it is at least len(1) so we don't panic

pkg/api/handlers/compat/networks.go
```diff
@@ -289,25 +289,40 @@ func CreateNetwork(w http.ResponseWriter, r *http.Request) {
...
+
+	query := struct {
+		Force bool `schema:"force"`
+	}{
+		// This is where you can override the golang default value for one of fields
+	}
+
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
```
edsantiago, line 302: I'm not going to block just because of this... but if you need to resubmit for any reason, could there maybe be a better diagnostic than "Something went wrong"?

edsantiago:
Tentatively LGTM aside from two questionable "Something went wrong" messages. I appreciate the `FIXME` in `SkipIfRemotes` that should work, that makes it easy to grep for them. I'll let others slash-lgtm tho.

pkg/api/handlers/compat/networks.go
rhatdan: Fixed

pkg/api/handlers/compat/networks.go
rhatdan: Fixed, although this message is in all over the code.  I think this is what Docker did when there was an unexpected error.

pkg/api/handlers/compat/networks.go
```diff
@@ -289,25 +289,44 @@ func CreateNetwork(w http.ResponseWriter, r *http.Request) {
 
 func RemoveNetwork(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
-	config, err := runtime.GetConfig()
-	if err != nil {
-		utils.InternalServerError(w, err)
+	ic := abi.ContainerEngine{Libpod: runtime}
+
+	query := struct {
+		Force bool `schema:"force"`
```
QiWang19, line 295: Does this parameter need document in pkg/api/server/register_networks.go?

pkg/api/handlers/compat/networks.go
TomSweeneyRedHat: @edsantiago I thoroughly concur and have been dinging those as I encounter them.  It makes me grind my teeth every time I spot it.

TomSweeneyRedHat:
LGTM once @QiWang19 's question is answered/addressed.

ashley-cui (commented):
LGTM

pkg/api/handlers/compat/networks.go
rhatdan: It was already documented, just not implemented.

PR #8380, opened 2020-11-17
Client call to /play/kube incorrectly set tlsVerify

AlbanBedel:
The API parameter `tlsVerify` should be the invert of the internal option `SkipTLSVerify`, fix this conversion.

### DIFF
--- a/pkg/bindings/play/play.go
+++ b/pkg/bindings/play/play.go
@@ -30,7 +30,7 @@ func Kube(ctx context.Context, path string, options entities.PlayKubeOptions) (*
 	params.Set("network", options.Network)
 	params.Set("logDriver", options.LogDriver)
 	if options.SkipTLSVerify != types.OptionalBoolUndefined {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
Would be nice to have a test for this.

TomSweeneyRedHat:
Good catch @AlbanBedel 
An added regression test would be most welomed!

QiWang19 (commented):
LGTM

AlbanBedel:
I don't think I'll be able to provide a test for this bug in a timely manner. It's surely not a major security issue as only users that explicitly set `--tls-verify=true` are affected, but I would still expect such a bug to be fixed ASAP and backported to all affected stable releases if there is such a process.

rhatdan:
/lgtm
@QiWang19 Could you throw together a test for this.

rhatdan:
Thanks @AlbanBedel

PR #8222, opened 2020-11-02
[CI:DOCS] Correct numbering typo

sethjones:

PR #8337, opened 2020-11-13
Cirrus: Invalidate static cache on VM image update

cevich:
Fixes: #8313

It's important to periodically update the nix cache (about 1GB in size).
If not, it can grow stale and has been observed causing task failures.
Associating the nix cache update with a VM/Container image update,
ensures it happens first in PRs where environment-related failures
are less surprising.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #8389, opened 2020-11-18
[CI:DOCS] correct numbering typo

rhatdan:
Based on PR from @sethjones.

Replaces: https://github.com/containers/podman/pull/8222

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -457,7 +457,7 @@ Attempts to run podman result in
 
 One workaround is to disable Secure Boot in your BIOS.
 
-### 20) error creating libpod runtime: there might not be enough IDs available in the namespace
+### 19) error creating libpod runtime: there might not be enough IDs available in the namespace
 
 Unable to pull images
 


rhatdan:
The CI/CD System is boloxed.  Forcing merge.

rhatdan:
@cevich CI/CD is not in a good state.

edsantiago:
Cirrus error, in big red block, is:
```
Failed to start: Invalid value for field 'resource.machineType': 'https://www.googleapis.com/compute/v1/projects/libpod-218412/zones/us-central1-a/machineTypes/'. The URL is malformed.
```
Grepping in my workdir, I find some instances of `machineType` but only in logs, never (AFAICT) in any source files. It always looks like the above URL, except the above is missing `/custom-something-something` at the end. 

What I find most intriguing is that, in my logs, the `machineType` line in json comes right after one called `labelFingerprint` ... and the last podman PR is #8337 which touched `.cirrus.yml` in something called `fingerprint_script`. Could just be coincidence, but I wonder if it's worth backing out #8337? I expect that PR would have to be force-merged as well and I don't know how to do that.

Issue #7706, opened 2020-09-21
RFE: add '--userns=...' option to 'pod create' command


ensc:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Please add the  '--userns=xxx' option to `podman pod create` so that it is possible to create (fully working) containers within a pod.

E.g. atm (podman-2.0.6-1.fc32.x86_64) it is not possible to do

```
podman pod create --name userns-test
podman run --pod=userns-test --user=root --rm -it --userns=keep-id busybox nc -l -p 22
```

(--> fails with "bind: Permission denied") while the same for a single container works:

```
podman run                   --user=root --rm -it --userns=keep-id busybox nc -l -p 22
```

[see https://github.com/containers/podman/issues/3993 too]

mlezajic:
Is there any chance to get this adressed somewhere in the near future? We really need to be able to add containers to a pod while keeping the current user id. Using podman v1.9.3 (go1.13.4) on RHEL v8.3.1-5 machine we ecounter this (as mentioned above):

`podman run --userns=keep-id --pod dev-pod-1 -dit --name centos 0d120b6ccaa8`

results in

`Error: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"sysfs\\\" to rootfs \\\"/home/fwd/.local/share/containers/storage/overlay/ab48d245cd5b06ca1b46c80ed7f64b4d6a5c23be27a6d79ee8321e5d18b5337a/merged\\\" at \\\"/sys\\\" caused \\\"operation not permitted\\\"\"": OCI runtime permission denied error
`

rhatdan:
This seems logical that we should associate the usernamespace with the pod and then any containers that get attached would be placed within that user namespace.

dominic-p:
Is this a regression? I could swear I was able to use `--userns=keep-id` inside a pod recently, but now I'm running into the same permission error.

mheon:
No - this has never worked.

fcabralpacheco:
@dominic-p 
@mheon 
I've been using containers with "--userns=keep-id" inside pods for some time on Fedora (Silverblue), and it works. The infra container will not be running with "--userns" ("UsernsMode"), but that it's not a problem for some use cases (I think).

@ensc 
> podman pod create --name userns-test
podman run --pod=userns-test --user=root --rm -it --userns=keep-id busybox nc -l -p 22

[...]

PR #8393, opened 2020-11-18
[wip] Add support for pod inside of user namespace.

rhatdan:
Specifically we need to be able to launch a pod with
--userns=keep-id

### DIFF
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -284,12 +284,19 @@ func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions
 	if len(podName) < 1 {
 		return nil, errors.Errorf("new pod name must be at least one character")
 	}
+
+	userns, err := specgen.ParseUserNamespace(cliVals.UserNS)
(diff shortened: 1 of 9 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/7706

rhatdan:
@umohnani8 Could you take this over?  I don't have the time to complete it.

rhatdan:
@umohnani8 Do you have a chance to take this over?

rhatdan:
The basic issue, we are seeing is people adding containers to pods where they use additional fields that do not match the pods infra container.

For example we have an issue where a user creates a pod, and then adds a container to the pod, within a user namespace, and the /dev/mqueue from the infra container fails to mount on the container, because the /dev/mqueue inside of the pod does not match the usernamespace inside of the container.

Bottom line, I believe we have to be able to at least configure the usernamespace of the pod.

umohnani8:
Will be taking over this work.

umohnani8:
Replaced by https://github.com/containers/podman/pull/10589

PR #8395, opened 2020-11-18
Align the podman pod ps --filter behavior with podman ps

Luap99:
Filters with the same key work inclusive with the only exception being
`label` which is exclusive. Filters with different keys always work exclusive.

Also update the documentation with the new behavior.


rhatdan:
LGTM
/approve

QiWang19 (commented):
LGTM

PR #8398, opened 2020-11-18
Shell completion for podman ps and podman pod ps --filter

Luap99:
Add all available filter options for `podman ps` and `podman
pod ps` to the completions. Refactor the code a bit to make it
easier to handle key value pairs in completions. The `completeKeyValues` function  helps to reduce code duplication.

Also make use of the new filter logic in the completions.

### DIFF
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -23,90 +23,75 @@ var (
 	LogLevels = []string{"debug", "info", "warn", "error", "fatal", "panic"}
 )
 
-func getContainers(toComplete string, statuses ...string) ([]string, cobra.ShellCompDirective) {
+type completeType int
+
+const (
+	// complete names and IDs after two chars
+	completeDefault completeType = iota
+	// only complete IDs
+	completeIDs
+	// only complete Names
+	completeNames
+)
+
+type keyValueCompletion map[string]func(s string) ([]string, cobra.ShellCompDirective)
+
+func getContainers(toComplete string, cType completeType, statuses ...string) ([]string, cobra.ShellCompDirective) {
 	suggestions := []string{}
 	listOpts := entities.ContainerListOptions{
 		Filters: make(map[string][]string),
 	}
 	listOpts.All = true
 	listOpts.Pod = true
+	if len(statuses) > 0 {
+		listOpts.Filters["status"] = statuses
+	}
 
-	// TODO: The api doesn't handle several different statuses correct see:
-	// https://github.com/containers/podman/issues/8344
-	// Instead of looping over the statuses we should be able to set
-	// listOpts.Filters["status"] = statuses
-
-	var containers []entities.ListContainer
-	var err error
-	if len(statuses) == 0 {
-		containers, err = registry.ContainerEngine().ContainerList(registry.GetContext(), listOpts)
-		if err != nil {
-			cobra.CompErrorln(err.Error())
-			return nil, cobra.ShellCompDirectiveError
-		}
-	} else {
-		for _, s := range statuses {
-			listOpts.Filters["status"] = []string{s}
-			res, err := registry.ContainerEngine().ContainerList(registry.GetContext(), listOpts)
-			if err != nil {
-				cobra.CompErrorln(err.Error())
-				return nil, cobra.ShellCompDirectiveError
-			}
-			containers = append(containers, res...)
-		}
+	containers, err := registry.ContainerEngine().ContainerList(registry.GetContext(), listOpts)
+	if err != nil {
+		cobra.CompErrorln(err.Error())
+		return nil, cobra.ShellCompDirectiveError
 	}
 
 	for _, c := range containers {
-		// include ids in suggestions if more then 2 chars are typed
-		if len(toComplete) > 1 && strings.HasPrefix(c.ID, toComplete) {
+		// include ids in suggestions if cType == completeIDs or
+		// more then 2 chars are typed and cType == completeDefault
+		if ((len(toComplete) > 1 && cType == completeDefault) ||
+			cType == completeIDs) && strings.HasPrefix(c.ID, toComplete) {
 			suggestions = append(suggestions, c.ID[0:12]+"\t"+c.PodName)
 		}
 		// include name in suggestions
-		if strings.HasPrefix(c.Names[0], toComplete) {
+		if cType != completeIDs && strings.HasPrefix(c.Names[0], toComplete) {
 			suggestions = append(suggestions, c.Names[0]+"\t"+c.PodName)
 		}
 	}
 	return suggestions, cobra.ShellCompDirectiveNoFileComp
 }
 
-func getPods(toComplete string, statuses ...string) ([]string, cobra.ShellCompDirective) {
+func getPods(toComplete string, cType completeType, statuses ...string) ([]string, cobra.ShellCompDirective) {
 	suggestions := []string{}
 	listOpts := entities.PodPSOptions{
 		Filters: make(map[string][]string),
 	}
+	if len(statuses) > 0 {
+		listOpts.Filters["status"] = statuses
+	}
 
-	// TODO: The api doesn't handle several different statuses correct see:
-	// https://github.com/containers/podman/issues/8344
-	// Instead of looping over the statuses we should be able to set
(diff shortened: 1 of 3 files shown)


Luap99:
@edsantiago PTAL

cmd/podman/common/completion.go
```diff
@@ -23,90 +23,73 @@ var (
...
+		return nil, cobra.ShellCompDirectiveError
 	}
 
 	for _, c := range containers {
-		// include ids in suggestions if more then 2 chars are typed
-		if len(toComplete) > 1 && strings.HasPrefix(c.ID, toComplete) {
+		// include ids in suggestions if cType == completeIDs or
+		// more then 2 chars are typed and cType == completeDefault
+		if ((len(toComplete) > 1 && cType == completeDefault) ||
+			cType == completeIDs) && strings.HasPrefix(c.ID, toComplete) {
```
edsantiago, line 58: Does Go have the concept of bitmasks? All these `==` and `!=` and `&&` are a little awkward for my eye to read; it seems they would be clearer with `if cType & completeIDs` or `& completeNames`

Luap99: Even if I use bitmasks I still would need to differentiate between completeIDs and completeDefault so it would look exactly the same.
Or I'm missing something?

edsantiago: Because of the `need two characters` requirement? Something like this could handle it:
```go
const (
       completeNames = 1
       completeIDs   = 2
       completePartialIDs = 4
 )
...
               if ((len(toComplete) > 1 && cType & completePartialIDs) ||
                       cType & completeIDs) && strings.HasPrefix(c.ID, toComplete) {
```
Unfortunately that's not as readable as I'd expected and I can't quickly think of a good alternative. OK, never mind then.

cmd/podman/common/completion.go
```diff
@@ -763,3 +746,77 @@ func AutocompleteSDNotify(cmd *cobra.Command, args []string, toComplete string)
 	types := []string{"container", "conmon", "ignore"}
 	return types, cobra.ShellCompDirectiveNoFileComp
 }
+
+// AutocompletePsFilters - Autocomplete ps filter options.
+func AutocompletePsFilters(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	id, name, status, ancestor := "id=", "name=", "status=", "ancestor="
+	before, since, volume, health := "before=", "since=", "volume=", "health="
+	filters := []string{id, name, "label=", "exited=", status, ancestor, before, since, volume, health, "until="}
+	switch {
```
edsantiago, line 755: I'm finding it really hard to deal with the duplication below, and the hardcoded `[stringlength:]`. I know Go has maps, does it have closures? Could you do something like:
```
    map...
       "id=": func(s string) { getContainers(s, completeIDs) },
       "name=": func(s string) { getContainers(s, completeNames) },
       "ancestor=": func(s string) { getImages(s) },
       ...
)
...
    found := false
    for each item in map:
       if strings.HasPrefix(toComplete, item.lhs) {
            strip_arg = toComplete[len(item.lhs):]
            item.func(strip_arg)
            found = true
        }
    if ! found {
         handle status, health, etc
```

Luap99: Something like that should be possible. Great idea.

Luap99: @edsantiago PTAL I think it should be way better now. Although I wish go had macros so I wouldn't have to type `func(s string) ([]string, cobra.ShellCompDirective) {}` all time.

edsantiago: That is much, much nicer, thank you! (And I agree that macros would make it even better). Looking into it some more. Thank you for your patience.

cmd/podman/common/completion.go
```diff
@@ -23,90 +23,75 @@ var (
...
+		// include ids in suggestions if cType == completeIDs or
+		// more then 2 chars are typed and cType == completeDefault
+		if ((len(toComplete) > 1 && cType == completeDefault) ||
+			cType == completeIDs) && strings.HasPrefix(c.ID, toComplete) {
 			suggestions = append(suggestions, c.ID[0:12]+"\t"+c.PodName)
 		}
 		// include name in suggestions
-		if strings.HasPrefix(c.Names[0], toComplete) {
+		if cType != completeIDs && strings.HasPrefix(c.Names[0], toComplete) {
 			suggestions = append(suggestions, c.Names[0]+"\t"+c.PodName)
```
edsantiago, line 65: I know this is not your addition, but what is the `"\t"+` intended to accomplish? On my system (zsh) it produces a very weird-looking completion list:
```console
$ ./bin/podman ps -a --filter id=<TAB>
id=276af9483c22  id=a722b8b4c064  id=5aa479ac915d  id=dc7cecab1356
id=3821798b331e  id=e00da78572a9  id=ed308f016b78  id=1a2b7290886c
id=c5e5427ef032  id=413e9b86c8d0                                    -- mypod      <<< this is weird
id=145f126fd181                                                                id=7b254fe8286b     <<< so is the extra spacing
```
Similarly with `--name=<TAB>`. If I remove the tab part and recompile, I get a (normal, expected) list of just IDs.

Luap99: That's the description part. It will show you the pod name if the container is in a pod. I don't use zsh but with bash it looks fine:
```
$ bin/podman ps --filter id=
0da6096d149d                 57e08662abe8                 936761df281d                 bfe79af8a30d                 d2e3bc52e2bf                 
355bcc085321  (test)         5b408e8c01e0                 a73f295f2f73  (angry_allen)  c6cc999b6c71 
```

Luap99: How does it look if you do `podman rm [TAB]`?

Luap99: That's my zsh output:
```
% bin/podman ps --filter id=
id=355bcc085321  -- test
id=a73f295f2f73  -- angry_allen
id=0da6096d149d  id=57e08662abe8  id=5b408e8c01e0  id=936761df281d  id=bfe79af8a30d  id=c6cc999b6c71  id=d2e3bc52e2bf
```
The formatting works ok for me although I think bash and also fish are looking better.

Luap99: You can also globally turn off descriptions for podman if you want.

edsantiago: I don't get those extra bits (parenthesized pod mane) in bash, only CIDs; there must be a bash option that I'm not setting. `podman rm <TAB>` gives me similar-looking results:
```
$ ./bin/podman rm <TAB>
2d9d3c73021a-infra  gifted_dhawan
mystifying_neumann  unruffled_hypatia  dreamy_kowalevski  keen_perlman
xenodochial_ellis   reverent_cannon    tender_dijkstra    kind_hellman  -- mypod
mystifying_buck                                                                romantic_burnell
```
FWIW **all** of those containers were run with `--pod mypod`. This may be an upstream issue in completion internals.

edsantiago: >That's my zsh output:

That looks good, it's what I tend to expect with zsh additional-comment info. OK, I'll assume it's something broken on my end, and will not pursue it further for now. Thanks for the confirmation that it looks good on your end.

Luap99: OK I see zsh groups the suggestion with the same description. That's useful:
```
% bin/podman ps --filter id=0da6096d149d
id=355bcc085321  id=82939027f0ef  id=45a4a1acd5b6  -- test                                                                                                                                   
id=a73f295f2f73                                    -- angry_allen                                                                                                                            
id=0da6096d149d            id=57e08662abe8            id=5b408e8c01e0            id=936761df281d            id=bfe79af8a30d            id=c6cc999b6c71            id=d2e3bc52e2bf
```

Luap99: @edsantiago I have turned off descriptions in bash by default since they are natively supported. You would need to run `source <(bin/podman completion bash)` to test.

edsantiago:
LGTM but I'd like an extra set of eyes on it. Really nice work @Luap99 , thank you.

PS I restarted `int podman fedora-32 root container`, it failed in `network rm` and it can only be a flake.

edsantiago:
Tests are green; @containers/podman-maintainers PTAL

PR #8405, opened 2020-11-18
Fix sed regex to update version in version/version.go

jwhonce:


mheon:
LGTM

mheon:
But someone more familiar with sed should take a look - maybe @edsantiago

edsantiago:
parentheses in sed give me a headache too. But, LGTM: I tried it on my command-line, manually setting LOCAL_VERSION, and it produces what I think is the desired effect.

Issue #7857, opened 2020-09-30
Eclipse Docker Tooling support


mlegenovic:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Docker compatibility endpoint `/images/json` does not comply with the Docker API specification. Eclipse Docker plugin calls `/images/json?all=true` and expects this structure: [https://github.com/dmandalidis/docker-client/blob/master/src/main/java/org/mandas/docker/client/messages/Image.java](https://github.com/dmandalidis/docker-client/blob/master/src/main/java/org/mandas/docker/client/messages/Image.java). Podman returns a JSON like this:
```
$ curl http://localhost:8080/images/json?all=true | jq                            
[
  {
    "Id": "50541622f4f179450f4acec7d16964499525932a81263eafb91e699671f58ee4",
    "RepoTags": [
      "docker.io/library/redis:6"
    ],
    "Created": "2020-07-22T02:15:29.603086742Z",
    "Size": 107545246,
    "Names": [
      "docker.io/library/redis:6"
    ],
    "Digest": "sha256:e495ab323824662760af192548758d93365ca2e796eb0ef54b7bf3a2bd775df5",
    "Digests": [
      "sha256:e495ab323824662760af192548758d93365ca2e796eb0ef54b7bf3a2bd775df5",
      "sha256:7d1ebef417593273b2f7b137f615393b21723d93568868c7fe5aad0401469152"
    ],
    "History": [
      "50541622f4f179450f4acec7d16964499525932a81263eafb91e699671f58ee4",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>",
      "<missing>"
    ]
  }
]
```
Deserialisation fails with following error message:

> An internal error occurred during: "Refreshing Images...".
> Entity input stream has already been closed.

**Steps to reproduce the issue:**

1. Call /images/json?all=true

2. Compare result with expected response schema described here: [https://docs.docker.com/engine/api/v1.40/#operation/ImageList](https://docs.docker.com/engine/api/v1.40/#operation/ImageList)

3. Podman does not return all required fields

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.0.6
API Version:  1
[...]

PR #8409, opened 2020-11-19
Docker compat API fixes

mlegenovic:
These are the first fixes that are needed for development environments like
Eclipse or IntelliJ, that have Docker plug-ins and use the Docker API to speak
with container engine (#7857)


zhangguanzhang:
need format

mheon:
CI is complaining that pkg/domain/entities/images.go needs a `gofmt -s -w` run.

pkg/api/handlers/types.go
```diff
@@ -215,7 +215,7 @@ func ImageToImageSummary(l *libpodImage.Image) (*entities.ImageSummary, error) {
 		Created:      l.Created().Unix(),
 		Size:         int64(*size),
 		SharedSize:   0,
-		VirtualSize:  l.VirtualSize,
```
mheon, line 218: I think this is correct as-is, though the VirtualSize output itself may be bugged right now.

jwhonce:
@mlegenovic  Thanks for the PR. To save you an additional hurdle, when you push the formatted code you need to sign your PR.

For example...
git commit -a --amend -s
git push --force

rhatdan:
@mlegenovic Still working on this?  Is this still a WIP or can we remove this flag?

mlegenovic:
@rhatdan The only reason why I set the WIP flag is because the issue is not yet solved with these changes. But the changes are completely independent and can be merged, so WIP can be removed.

rhatdan:
Great @mlegenovic Only issue then is @mheon comments on Virtual Size. Can you revert that change and then we can get this reviewed and into the package.

pkg/api/handlers/types.go
mlegenovic: I absolutely agree. The only reason why I did it that way is because my Go knowledge is minimal, so I couldn't figure out why it's wrong now. So I looked at what inspect command does and took it from there.

rhatdan:
LGTM

PR #8410, opened 2020-11-19
Allow multiple --network flags for podman run/create

Luap99:
We allow a container to be connected to several cni networks
but only if they are listed comma sperated. This is not intuitive
for users especially since the flag parsing allows multiple string
flags but only would take the last value. see: spf13/pflag#72

Also get rid of the extra parsing logic for pods. The invalid options
are already handled by `pkg/specgen`.

A test is added to prevent a future regression.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -584,7 +584,7 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **container:**_id_: reuse another container's network stack;
 - **host**: use the Podman host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
-- _network-id_: connect to a user-defined network, multiple networks should be comma separated;
+- **cni-network**: connect to a user-defined network, multiple networks should be comma-separated or they can be specified with multiple uses of the **--network** option;
 - **ns:**_path_: path to a network namespace to join;
 - **private**: create a new namespace for the container (default)
 - **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack.  This is the default for rootless containers.  It is possible to specify these additional options:
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -610,7 +610,7 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **container:**_id_: reuse another container's network stack;
 - **host**: use the Podman host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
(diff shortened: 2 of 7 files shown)


mheon:
I'm not strictly opposed, but if we do this we need to be very sure the manpage makes it clear that this is only for multiple networks - I'm afraid of users thinking that `podman run --net slirp4netns --net host` and similar invocations are valid.

rhatdan:
Could we add a failure mode where the user did do `--net=host --net=container:foobar`

TomSweeneyRedHat:
The code LGTM, but share @mheon's thought on man page updates for this.

zhangguanzhang:
> Could we add a failure mode where the user did do `--net=host --net=container:foobar`

agree

Luap99:
Updated. This is now the error.
```
$ podman run --network host --network foo-a alpine
Error: network conflict between type host and bridge
```
I also removed the extra pod parsing logic. The invalid options are already handled in `pkg/specgen`

mheon:
/approve
Restarted one test which looked flaked.
LGTM

docs/source/markdown/podman-create.1.md
```diff
@@ -584,7 +584,7 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **container:**_id_: reuse another container's network stack;
 - **host**: use the Podman host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
-- _network-id_: connect to a user-defined network, multiple networks should be comma separated;
+- **cni-network**: connect to a user-defined network, multiple networks should be comma separated or can specified with multiple uses of **--network**;
```
TomSweeneyRedHat, line 587:
```suggestion
- **cni-network**: connect to a user-defined network, multiple networks should be comma-separated or they can be specified with multiple uses of the **--network** option;
```

docs/source/markdown/podman-run.1.md
```diff
@@ -610,7 +610,7 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **container:**_id_: reuse another container's network stack;
 - **host**: use the Podman host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
-- _network-id_: connect to a user-defined network, multiple networks should be comma separated;
+- **cni-network**: connect to a user-defined network, multiple networks should be comma separated or can specified with multiple uses of **--network**;
```
TomSweeneyRedHat, line 613:
```suggestion
- **cni-network**: connect to a user-defined network, multiple networks should be comma-separated or they can be specified with multiple uses of the **--network** option;
```

Luap99:
Tests are green.

TomSweeneyRedHat:
LGTM
Thanks @Luap99 !

Issue #8411, opened 2020-11-19
aarch64 (Raspberry pi) fails to install podman container to network, error: rootless-podman-network-sandbox


rupert160:
Can somebody explain why podman networking fails to connect my container to the virtual network?
Fedora 33 aarch64, the following works fine:
```
podman run -dt --name hello-nginx nginx
```
However creating a container within a pod does not work:
```
podman network create podnet1 --subnet 192.168.0.1/24 --driver bridge

podman pod create --replace \
	--publish 53/tcp,53/udp,67/udp,80/tcp,443/tcp \
	--hostname ph2 \
	--name ph2 \
	--network podnet1

podman run -dt --pod ph2 --name hello-nginx2 nginx
```

ERRO[0000] error starting some container dependencies   
ERRO[0000] "cannot find rootless-podman-network-sandbox image for arm" 
Error: error starting some containers: internal libpod error

additionally attaching a container to the same `podnet1` network fails
```
podman run -dt --network podnet1 --name hello-nginx2 --publish 8080:80/tcp nginx
```
Error: cannot find rootless-podman-network-sandbox image for arm

It would appear that https://github.com/AkihiroSuda/podman-network-create-for-rootless-podman
has not been created for ARM or aarch64 - but I'm not sure. I cannot pull this image at least.

mheon:
I'm pretty sure we're maintaining this one ourselves in Quay - @AkihiroSuda correct?
@TomSweeneyRedHat Do you know what architectures Quay builds us for by default?

AkihiroSuda:
It is built manually, only for amd64 currently

TomSweeneyRedHat:
@mheon, not ours on quay yet.  @barthy1 built Skopeo images with different arch types for us in quay.  I can't check at the moment as quay isn't happy, but I'm pretty sure our Podman images there by default are X86_64 as that's what they're built on.

barthy1:
looks like update [here](https://github.com/containers/podman/blob/2aa6a8577d088271251604b028100f00a442f5cd/libpod/rootless_cni_linux.go#L27) is required to support other archs. So far only amd64 version is available

barthy1:
and of course multi-arch build of `quay.io/libpod/rootless-cni-infra` is requried

rupert160:
Thanks @barthy1 - I've tried to `podman pull quay.io/libpod/rootless-cni-infra` but the container can't be found... is this what you meant?

barthy1:
@rupert160 
[...]

PR #8415, opened 2020-11-19
Build multi-arch rootless-cni-infra on GitHub Actions

AkihiroSuda:
The image is built as a multi-arch OCI tar.gz and uploaded as a GitHub Artifact.

The image is NOT pushed to quay automatically.
~~A maintainer can use `skopeo` to push the artifact archive to quay.~~ (**EDIT**: might not be true for multi-arch...)

Relates to #8411

### DIFF
--- a/.github/workflows/rootless-cni-infra.yaml
+++ b/.github/workflows/rootless-cni-infra.yaml
@@ -1,0 +1,36 @@
+name: Build rootless-cni-infra
+on: [push, pull_request]
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: "Register QEMU to /proc/sys/fs/binfmt_misc"
+      run: docker run --privileged --rm tonistiigi/binfmt --install all
+    - name: "Fetch buildx binary"
+      run: |
+        wget -O buildx https://github.com/docker/buildx/releases/download/v0.4.2/buildx-v0.4.2.linux-amd64
+        chmod +x buildx
+    - name: "Initialize buildx"
+      run: |
+        ./buildx create --name cross --platform=amd64,arm,arm64,s390x,ppc64le --use
+        ./buildx inspect --bootstrap
+    - name: "Build rootless-cni-infra.tar.gz (Multi-arch OCI archive)"
+      run: |
+        ./buildx build \
+          --output type=oci,dest=rootless-cni-infra.tar \
+          --platform amd64,arm,arm64,s390x,ppc64le \
+          -f contrib/rootless-cni-infra/Containerfile \
+          contrib/rootless-cni-infra
+        gzip -9 rootless-cni-infra.tar
+    - name: "Print SHA256SUM of rootless-cni-infra.tar.gz"
+      run: |
+        sha256sum rootless-cni-infra.tar.gz
+    - name: "Upload rootless-cni-infra.tar.gz as a GitHub Artifact"
+      uses: actions/upload-artifact@v1
+      with:
+        name: rootless-cni-infra.tar.gz
+        path: rootless-cni-infra.tar.gz
+    - name: "Notice"
+      run: |
+        echo "The image is NOT pushed to quay. To push the image to quay, run skopeo manually with the artifact archive."
--- a/contrib/rootless-cni-infra/Containerfile
+++ b/contrib/rootless-cni-infra/Containerfile
@@ -4,22 +4,22 @@ ARG CNI_VERSION=v0.8.0
 ARG CNI_PLUGINS_VERSION=v0.8.7
 ARG DNSNAME_VESION=v1.0.0
 
-FROM golang:${GOLANG_VERSION}-alpine${ALPINE_VERSION} AS golang-base
+FROM --platform=$BUILDPLATFORM golang:${GOLANG_VERSION}-alpine${ALPINE_VERSION} AS golang-base
 RUN apk add --no-cache git
 
-FROM golang-base AS cnitool
+FROM --platform=$BUILDPLATFORM golang-base AS cnitool
 RUN git clone https://github.com/containernetworking/cni /go/src/github.com/containernetworking/cni
 WORKDIR /go/src/github.com/containernetworking/cni
 ARG CNI_VERSION
 RUN git checkout ${CNI_VERSION}
-RUN go build -o /cnitool ./cnitool
+RUN GOARCH=$TARGETARCH go build -o /cnitool ./cnitool
 
-FROM golang-base AS dnsname
+FROM --platform=$BUILDPLATFORM golang-base AS dnsname
 RUN git clone https://github.com/containers/dnsname /go/src/github.com/containers/dnsname
 WORKDIR /go/src/github.com/containers/dnsname
 ARG DNSNAME_VERSION
 RUN git checkout ${DNSNAME_VERSION}
-RUN go build -o /dnsname ./plugins/meta/dnsname
+RUN GOARCH=$TARGETARCH go build -o /dnsname ./plugins/meta/dnsname
 
 FROM alpine:${ALPINE_VERSION}
 RUN apk add --no-cache curl dnsmasq iptables ip6tables iproute2
@@ -33,4 +33,4 @@ COPY rootless-cni-infra /usr/local/bin
 ENV CNI_PATH=/opt/cni/bin
 CMD ["sleep", "infinity"]
 
-ENV ROOTLESS_CNI_INFRA_VERSION=3
+ENV ROOTLESS_CNI_INFRA_VERSION=4


AkihiroSuda:
> A maintainer can use skopeo to push the artifact archive to quay.

This might not be true for multi-arch?

rhatdan:
@cevich @baude @mheon PTAL

AkihiroSuda:
Can Red Hat configure GitHub Actions to allow pushing to quay?

TomSweeneyRedHat:
@barthy1 would mind taking a peak at this?
@AkihiroSuda I think you're looking for something like what @cevich did in Skopeo https://github.com/containers/skopeo/pull/1104

TomSweeneyRedHat:
And FWIW, here's the PR that @barthy1 put together for Skopeo, I've not looked at this PR to see if it's using a similar technique or not, but would like the ending images to be similar.  https://github.com/containers/skopeo/pull/1066

barthy1:
Hi @AkihiroSuda Don't you consider to use github actions to push multi-arch image to `quay.io/libpod/rootless-cni-infra` instead of generating archive? Of course this should be triggered not by pull_request, but push to master, I guess?

AkihiroSuda:
That would be better. Needs assistance from Red Hat for adding quay token to GitHub Actions.

barthy1:
@TomSweeneyRedHat in this PR the approach is - github action + emulation with `docker buildx` 
For Skopeo it was Travis + native hardware, as emulation one failed to build Skopeo. 

If emulation approach works for all listed architectures in this PR, I guess, it's the best way to build multi-arch image. To push to `quay.io/libpod` @AkihiroSuda need your (or @cevich )  assistance to provide push credentials for Github Actions, see https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository

The only one disadvantage I can see for the future extensions, if some podman image needs to be built with native hardware (as it was for Skopeo), then you will need to support 2 different build environments - github actions and Travis

AkihiroSuda:
Can we get a separate quay account for GitHub Actions? Also, I think the account should be maintained by maintainers rather than me, so that the account can be potentially reused for other image as well in future.

mheon:
This sounds sounds like it's going to require collaboration from @cevich to get the Github Actions piece working and @TomSweeneyRedHat who maintains the Quay account last I checked

TomSweeneyRedHat:
@barthy1 @AkihiroSuda I'm trying to get my head around this.  So this looks to produce the same images that @barthy1 did in Skopeo in the attached PR, but they're in a tar file that needs to be untarred and then the images pushed to quay.io via a GitHub action?  Currently, in Skopeo there are v-1.2.0-* tags where the '*' is amd64, ppc64le, and s390x.  It looks like this would create arm and arm64 variants too.  I don't see a version in the change, would this change add the Podman version number too, or is that work for the GitHub action?

We don't currently have travis here in this repo, but if we did, would it be better to build these the way the @barthy1 did using Travis?

I'm sure @cevich would be happy to add the GitHub actions, but I believe he's enjoying PTO at the moment and probably won't be back until early next week.

@AkihiroSuda and @barthy1 ty both very much for all your help on this in both projects.

AkihiroSuda:
> but they're in a tar file that needs to be untarred and then the images pushed to quay.io via a GitHub action

It is built as a tar file just because I don't have the registry token in CI to push the image to quay

TomSweeneyRedHat:
@AkihiroSuda thanks for the info.  @cevich is away until next week, let's touch base on this again then.

cevich:
I'm back, and happy to create/add needed credentials  I'll take a peek at this PR later today.

cevich:
I took a look, and am concerned on several fronts.

- It seems dangerous/failure-prone to rely on github-actions and action-steps which pull from docker hub.  Since this is running for every single PR, I'm afraid we'll rapidly run into the new rate-limits.
- It depends heavily/entirely on docker-centric third-party tooling which the podman maintainers are unlikely to ever embrace.  This creates a conflict of interest where the multi-arch images could be "held hostage" by the very technology podman is attempting to disrupt.

The second point is the real kicker beceause, `runs-on: ubuntu-20.04` will not be forever.  At some point we will ***have*** to update the CI-host and (likely) all tooling.  If there is a conflict anywhere in the ubuntu -> docker -> buildx -> QEMU space, then production of new podman images halts...at least until (worst-case) a podman maintainer (starting from scratch) learns and debugs the entire workflow enough to apply necessary fixes.

While I'm not a big fan of Travis-CI...the approach used @barthy1 for skopeo is free from dependence on intermediate tooling and simple enough that lay-people can understand it from start to finish.   Since native hardware is utilized, it only requires a few generic docker operations (which could easily be replicated with future podman), and a simple(ish) Makefile.

That all said, I'm sensitive to and appreciate the amount of work that's already been committed here.  Clearly it works as designed, and just needs some secrets added (which we're happy to create/provide).  On the other hand, and in accordance with the "sunken-cost fallacy", I wonder if the greater-good might be best served by re-implementing this to more closely resemble the skopeo setup.

I will support the decision of the group.

AkihiroSuda:
> It depends heavily/entirely on docker-centric third-party tooling which the podman maintainers are unlikely to ever embrace. This creates a conflict of interest where the multi-arch images could be "held hostage" by the very technology podman is attempting to disrupt.

What if we use buildctl directly instead of docker buildx CLI?

cevich:
> What if we use buildctl directly instead of docker buildx CLI?

That's (also) not something I'm familiar with.  Poking around on google, it seems to be the nuts-and-bolts behind buildkit?

I thought about this situation last night, and believe my main concern is over long-term maintainability.    My other argument over tooling/dependency hegemony seems a lesser concern.  This root of this is that there's a HUGE disparity in the number of people working on podman and those working on CI/CD infra. and tooling.  I'm not above relying on the community to help with maintenance, that's a positive force in my book.  Rather, we need to accept the historically-likely possibility that if/when broken for any reason, and nobody is around with skills to fix them quickly, these workflows may simply be disabled or deleted :disappointed: 

One idea that could help:  Document the heck out of it.  If a future podman-developer (not automation guru) can read a doc. and be able to upgrade/update/replace components of the image-build workflow w/o trolling around the internet much, that would be an acceptable compromise.  Perhaps that would be less work than re-implementing this to use the Travis+Native HW approach used by skopeo?

What are @rhatdan and @TomSweeneyRedHat and @mheon opinions here?  Am I being overly paranoid, and should go back into my little corner to snuggle my bash scripts?

.github/workflows/rootless-cni-infra.yaml
```diff
@@ -0,0 +1,36 @@
...
+        ./buildx create --name cross --platform=amd64,arm,arm64,s390x,ppc64le --use
+        ./buildx inspect --bootstrap
+    - name: "Build rootless-cni-infra.tar.gz (Multi-arch OCI archive)"
+      run: |
+        ./buildx build \
+          --output type=oci,dest=rootless-cni-infra.tar \
+          --platform amd64,arm,arm64,s390x,ppc64le \
+          -f contrib/rootless-cni-infra/Containerfile \
+          contrib/rootless-cni-infra
+        gzip -9 rootless-cni-infra.tar
```
cevich, line 25: I'm not sure the compression is worthwhile here, `actions/upload-artifact` will create a compressed zip-file automatically.

.github/workflows/rootless-cni-infra.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - name: "Print SHA256SUM of rootless-cni-infra.tar.gz"
+      run: |
+        sha256sum rootless-cni-infra.tar.gz
```
cevich, line 28: Perhaps this is my own ignorance: What is the purpose of calculating/printing the sha256sum of the tarball?

.github/workflows/rootless-cni-infra.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - name: "Upload rootless-cni-infra.tar.gz as a GitHub Artifact"
+      uses: actions/upload-artifact@v1
+      with:
+        name: rootless-cni-infra.tar.gz
+        path: rootless-cni-infra.tar.gz
+    - name: "Notice"
+      run: |
+        echo "The image is NOT pushed to quay. To push the image to quay, run skopeo manually with the artifact archive."
```
cevich, line 36: Were you thinking of implementing the push to quay, or is that something I could/should do as a followup PR?

I'm okay either way, but since I need to create the accounts and add the secrets, it might be easier if done as a followup.

AkihiroSuda:
Another option I'm thinking of is embedding the Containerfile into podman binary, and let podman build it on first run. This should work on any arch, and yet we no longer need to maintain the quay image.

mheon:
I like this - it resolve the distribution problem completely.

AkihiroSuda:
> I like this - it resolve the distribution problem completely.

Sorry, which is "this"?

mheon:
Your latest suggestion, building the image automatically on first run.

cevich:
> we will have a drop-in replacement for the existing docker buildx solution

Oh! Wasn't aware of that.  Yes, that negates my buildx + docker conflict-of-interest argument completely.  

>  Your latest suggestion, building the image automatically on first run.

I like this idea also, as we're all familiar with Makefiles and it sounds like only basic tooling would be required.  It also means writing less documentation :grin:

cevich:
> Also, this option is beneficial for internetless deployment.

Based on my experience working the RHEL-support front-lines:  That benefit should not be under-stated. It's a ***very*** critical/important benefit to a great number of often non-advertised uses.  It's also "one less thing to flake/break" should there be a networking failure at an inopportune time.

TomSweeneyRedHat:
Maybe I'm missing the point of the embedded Containerfile in the Podman binary, but it sounds like the end-user would have to install Podman first, then know how to make the image for their arch using the Podman binary?  From what I've seen, most people just want to grab an image from some "official" location on quay.io and go from there.

I love @barthy1's work on Skopeo, but we no longer have travis here in Buildah or Podman.  I'm not sure how easily that work would translate here without reviving travis and I'd rather not do that.  I really don't have a leaning towards the approaches discussed here as long as it documented so even I could understand it, and that we end up with an image for each of the arches that @AkihiroSuda created in this PR that automatically get out to quay.io.  

Then whatever we do here, I'd like to also do on the Buildah project.  If it makes sense to do the same kind of work on the Skopeo project, I'd be OK with replacing the work @barthy1 has provided there, but I'd prefer just adding the one or two archs to @barthy1's work on Skopeo that are here in this work but not yet in Skopeo.

Hopefully, my thoughts here are somewhat on track, I've been completely buried in emails/PR's this week and it's all blurring at this point.

cevich:
@TomSweeneyRedHat you're not far off.  Yes I'm not too happy about reviving Travis, but it is possible.  

> embedded Containerfile

No...as I understand it, we would be embedding the multi-arch image (the tarball) into the release.  Nothing gets built at package install time, only perhaps `podman load`ed (if that).

> From what I've seen, most people just want to grab an image from some "official" location on quay.io and go from there.

Yes, and this is the tail-end of my multi-dependency + complexity fear:  Users don't want some months old image, they assume it's "recent".  But if the automation is broken and nobody knows how to fix it, nothing newer will get pushed to quay :confounded:  It becomes 100% manual building ***and*** pushing at that point :disappointed:

mheon:
I'm going to bring this up for discussion with the team at scrum tomorrow, because we really need a plan here. We are preparing to ship a release of Podman including support for this in RHEL, which means we need a plan for distributing this image via RHEL channels, so this needs to be solved sooner rather than later.

mheon:
We deferred some of the discussion to Monday, but from the options we looked at, the most promising seems to be moving builds to OBS and then pushing to Quay from there - OBS possesses builders for all major architectures, the capability to autobuild on new commits (though this may require us to move the containerfile over to its own repository so builds are only triggered by changes to the image itself), and we're already very familiar with it as we use it to distribute the latest version of Podman for non-Red Hat OSes.

AkihiroSuda:
My current opinion is to just place the OCI archive as /usr/share/podman/rootless-cni-infra.tgz .

It doesn't need multi-arch builder, and doesn't need Internet either.

rhatdan:
I like that idea.

rhatdan:
After some discussion, I believe the best solution would be to build a container image off of the host OS. We should not need an image at all.  Just use the hosts CNI plugins and friends.  (We should do this for the pause container as well).

Someone needs to work out the logistics of this, but theoretically we should be able to do something like:

podman run --rootfs / echo hello

Doing this from rootless mode would allow us to do everything your container image requires, without needing to download or load an image, and using up all the extra disk space. (Or adding the image to the podman package.)

AkihiroSuda:
I agree that --rootfs / is the correct solution, but implementation may take long time. 

I think we need a short-time solution as well (i.e. local OCI archive under /usr/share/podman)

rhatdan:
Yes I think we should ship a Containerfile that could be used to build the image and to allow pulling form container registry.  But long term, we need to add support for OS based container images.

AkihiroSuda:
> I agree that --rootfs / is the correct solution, but implementation may take long time.

This might be easy. `runc spec; mount --rbind / ./rootfs; runc run foo` inside userns works without extra work.

AkihiroSuda:
opened a proposal for imageless rootless-cni-infra https://github.com/containers/podman/issues/8709

rhatdan:
Very nice.

PR #8416, opened 2020-11-19
more shell completion improvements

Luap99:
* podman image ls --filter
* podman network ls --filter
* podman volume ls --filter
* podman network connect/disconnect
* podman events --filter

If we return `ShellCompDirectiveError` to the shell the shell will
provide path completion. In none of that cases we want path completion
so it will be better to return `ShellCompDirectiveNoFileComp` instead
and log the error in case we need it.


Luap99:
@edsantiago Another one :)

Luap99:

rhatdan:
LGTM
/approve

Issue #8397, opened 2020-11-18
New flake: unable to obtain cgroup stats: open /sys/fs/cgroup/pids/pids.current: ENOENT


edsantiago:
This one started just yesterday, and it's hitting us all over the place. The basic gist is:
```console
$ podman-remote [options] pod stats --no-stream ...
Error: unable to obtain cgroup stats: open /sys/fs/cgroup/pids/pids.current: no such file or directory
```
Only on ubuntu-19 and -20; only root, not rootless. Seems to happen both local and remote. What changed yesterday?

------------------------

Podman pod stats [It] podman stats on running pods
* gce_instance:prior-ubuntu : int podman ubuntu-19 root host
  * PR #8363
    * [11-17 14:44](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5552157690691584/html/int-podman-ubuntu-19-root-host.log.html#t--podman-stats-on-running-pods--1)
    * [11-17 14:44](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5552157690691584/html/int-podman-ubuntu-19-root-host.log.html#t--podman-stats-on-running-pods--2)
    * [11-17 14:44](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5552157690691584/html/int-podman-ubuntu-19-root-host.log.html#t--podman-stats-on-running-pods--3)
* gce_instance:prior-ubuntu : int remote ubuntu-19 root host
  * PR #8170
    * [11-18 11:47](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5933768118632448/html/int-remote-ubuntu-19-root-host.log.html#t--podman-stats-on-running-pods--1)
    * [11-18 11:47](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5933768118632448/html/int-remote-ubuntu-19-root-host.log.html#t--podman-stats-on-running-pods--2)
    * [11-18 11:47](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5933768118632448/html/int-remote-ubuntu-19-root-host.log.html#t--podman-stats-on-running-pods--3)
* gce_instance:ubuntu : int podman ubuntu-20 root host
  * PR #8371
[...]

PR #8422, opened 2020-11-20
fix container cgroup lookup

vrothberg:
When running on cgroups v1, `/proc/{PID}/cgroup` has multiple entries,
each pointing potentially to a different cgroup.  Some may be empty,
some may point to parents.

The one we really need is the libpod-specific one, which always is the
longest path.  So instead of looking at the first entry, look at all and
select the longest one.

Fixes: #8397

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // CgroupfsDefaultCgroupParent is the cgroup parent for CGroupFS in libpod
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

libpod/container.go
```diff
@@ -920,19 +921,39 @@ func (c *Container) CGroupPath() (string, error) {
 		return "", errors.Wrapf(define.ErrNoCgroups, "this container is not creating cgroups")
 	}
 
-	// Read /proc/[PID]/cgroup and look at the first line.  cgroups(7)
-	// nails it down to three fields with the 3rd pointing to the cgroup's
-	// path which works both on v1 and v2.
+	// Read /proc/[PID]/cgroup and find the *longest* cgroup entry.  That's
+	// needed to account for hacks in cgroups v1, where each line in the
+	// file could potentially point to a cgroup.  The longest one, however,
```
rhatdan, line 926: This seems pretty arbitrary, but if it fixes our Flake, 

LGTM

PR #8426, opened 2020-11-20
Do not ignore infra command from config files

mheon:
It looks like this was previously removed because the default hard-coded `/pause` so we would never take into account the image config. I've removed the default in c/common and re-added support to check config files.

While we're at it, fix ENTRYPOINT support - we should not be setting this if we got ENTRYPOINT from the image.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1853455

### DIFF
--- a/libpod/runtime_pod_infra_linux.go
+++ b/libpod/runtime_pod_infra_linux.go
@@ -34,40 +34,56 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
 	// Set Pod hostname
 	g.Config.Hostname = p.config.Hostname
 
+	var options []CtrCreateOption
+
+	// Command: If user-specified, use that preferentially.
+	// If not set and the config file is set, fall back to that.
+	var infraCtrCommand []string
+	if p.config.InfraContainer.InfraCommand != nil {
+		logrus.Debugf("User-specified infra container entrypoint %v", p.config.InfraContainer.InfraCommand)
+		infraCtrCommand = p.config.InfraContainer.InfraCommand
+	} else if r.config.Engine.InfraCommand != "" {
+		logrus.Debugf("Config-specified infra container entrypoint %s", r.config.Engine.InfraCommand)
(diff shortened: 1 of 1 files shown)


mheon:
@rhatdan Do we have tests for containers.conf where we modify it and verify the changes took place? And if so, where do they live, so I can add another one for this?

go.sum
```diff
@@ -96,8 +96,8 @@ github.com/containernetworking/plugins v0.8.7/go.mod h1:R7lXeZaBzpfqapcAbHRW8/CY
 github.com/containers/buildah v1.18.0 h1:mWEm013LVNGecF++sYo0T7fe/4pqMas/PQxQ/qviC68=
 github.com/containers/buildah v1.18.0/go.mod h1:qHLk7RUL7cHfA7ve1MKkZ6cyKUxHD0YxiLJcKY+mJe8=
 github.com/containers/common v0.26.3/go.mod h1:hJWZIlrl5MsE2ELNRa+MPp6I1kPbXHauuj0Ym4BsLG4=
-github.com/containers/common v0.27.0 h1:+QlYEOitVYtU9/x8xebRgxdGqt4sLaIqV6MBOns+zLk=
-github.com/containers/common v0.27.0/go.mod h1:ZTswJJfu4aGF6Anyi2yON8Getda9NDYcdIzurOEHHXI=
+github.com/containers/common v0.27.1-0.20201120113506-fdcaa52cf3b9 h1:2mgVIr7x2hPztzvmSlLpv7YkiiLU8ETqf3kmv8xO8Ck=
```
TomSweeneyRedHat, line 99: Is @rhatdan making a v0.28.0 anytime soon?

TomSweeneyRedHat:
LGTM
but definitely want someone with more infra knowledge than I to look too.

go.sum
rhatdan: I need to cut one soon, but I was waiting for other PRs to finish or not.

mheon:
Tests failures are just flakes, @containers/podman-maintainers PTAL

mheon:
Green now

Issue #8414, opened 2020-11-19
podman ps --sort created : not always sorting by created


edsantiago:
About one in five tries, this gives me the wrong (unexpected) result:
```console
$ img=quay.io/libpod/testimage:20200929
$ ./bin/podman run -d --name c1 $img true;./bin/podman run -d --name c2 $img true;./bin/podman ps -a --sort created
faf3c3ebd4c764dad41ecee6ae0aafb077ef0ae9181f03c05235c8445de779fa
a381df699c40d4a46f4a678b2758154af375bcb1aee445526171afbf3d9f6e76
CONTAINER ID  IMAGE                              COMMAND  CREATED       STATUS                             PORTS   NAMES
a381df699c40  quay.io/libpod/testimage:20200929  true     1 second ago  Exited (0) Less than a second ago          c2
faf3c3ebd4c7  quay.io/libpod/testimage:20200929  true     1 second ago  Exited (0) 1 second ago                    c1
```
I expect to see c1 before c2, always. It's not a timing issue with `-d`, because `c1.Created < c2.Created`:
```console
$ ./bin/podman inspect --format '{{.Name}} {{.Created}}' c1 c2
c1 2020-11-19 13:30:14.806452 -0700 MST
c2 2020-11-19 13:30:14.948358079 -0700 MST
```
It's not a flake in `ps`, because for any given run of two containers, `podman ps -a --sort created` will always return the same order, always always, at least for values of "always" less than a ten-minute `while` loop.

In every instance when I see this, the CID of c2 is lexically `<` CID(c1); so much so that it seems improbable (many tens of repeated runs). I don't know if this is helpful. The sorting code looks good to my eye: https://github.com/containers/podman/blob/a18365c908d45a8ee9348c5e32a240a7b9a4091b/pkg/domain/entities/container_ps.go#L139-L141

I don't know where else to look; am hoping that someone can take a quick look at the sorting code and see something obvious.

PR #8427, opened 2020-11-20
Handle ps container created field as a time.Time

rhatdan:
In the current code we were translating the created time
from a time.Time to a unix epoch, this was leading to a loss
of precession, and some unexpected results where the sorting
order of containers was misordered because of the precession loss.

Fixes: https://github.com/containers/podman/issues/8414

If we pass around created as time.Time, we do not loose the precission.

### DIFF
--- a/pkg/domain/entities/container_ps.go
+++ b/pkg/domain/entities/container_ps.go
@@ -3,6 +3,7 @@ package entities
 import (
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/containers/podman/v2/pkg/ps/define"
 	"github.com/cri-o/ocicni/pkg/ocicni"
@@ -14,7 +15,7 @@ type ListContainer struct {
 	// Container command
 	Command []string
 	// Container creation time
-	Created int64
+	Created time.Time
 	// Human readable container creation time.
 	CreatedAt string
 	// If container has exited/stopped
@@ -137,7 +138,7 @@ func (a psSortedSize) Less(i, j int) bool {
 type PsSortedCreateTime struct{ SortListContainers }
 
 func (a PsSortedCreateTime) Less(i, j int) bool {
-	return a.SortListContainers[i].Created < a.SortListContainers[j].Created
+	return a.SortListContainers[i].Created.Before(a.SortListContainers[j].Created)
 }
 
 func SortPsOutput(sortBy string, psOutput SortListContainers) (SortListContainers, error) {
--- a/pkg/ps/ps.go
+++ b/pkg/ps/ps.go
@@ -180,7 +180,7 @@ func ListContainerBatch(rt *libpod.Runtime, ctr *libpod.Container, opts entities
 
 	ps := entities.ListContainer{
 		Command:   conConfig.Command,
-		Created:   conConfig.CreatedTime.Unix(),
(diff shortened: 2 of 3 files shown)


rhatdan:
@edsantiago PTAL

pkg/ps/ps.go
```diff
@@ -301,5 +301,5 @@ func (a SortPSContainers) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
 type SortPSCreateTime struct{ SortPSContainers }
 
 func (a SortPSCreateTime) Less(i, j int) bool {
-	return a.SortPSContainers[i].Created > a.SortPSContainers[j].Created
+	return a.SortPSContainers[i].Created.Before(a.SortPSContainers[j].Created)
```
TomSweeneyRedHat, line 304: I'm a bit confused.  Was the ">" correct for the original version of this "Less()" function?  If so, does this need to be `After` rather than `Before`, or are you correcting a bad sort?

Back to my tea cup.

pkg/domain/entities/container_ps.go
```diff
@@ -14,7 +15,7 @@ type ListContainer struct {
 	// Container command
 	Command []string
 	// Container creation time
-	Created int64
+	Created time.Time
```
mheon, line 18: @rhatdan @jwhonce Is this going over the wire? As in, is this a breaking change to our JSON format?

pkg/ps/ps.go
jwhonce: @TomSweeneyRedHat I'm a mistake in original code see https://github.com/containers/podman/pull/8427/files#diff-6fb97e57a192c9eae49da7e6a2ca7783e8d30eaf4d323cf55b56b28861016eb4L140

jwhonce (approved):
LGTM

pkg/ps/ps.go
TomSweeneyRedHat: @jwhonce "I'm a mistake in original code", I wasn't aware you've gone full Tron in this project.  That's a bit scary! :smile: 
Thx for the 411

TomSweeneyRedHat:
LGTM
as long as @jwhonce is comfortable with the date type change.

pkg/domain/entities/container_ps.go
rhatdan: I don't believe this is going over the wire.  The time.Time came over the wire though.

rhatdan: But this may be a breaking change for anyone who has vendored in /pkg/ directories. So we might have to wait to fix this until 3.0.

rhatdan:
@vrothberg  Do you agree this is a breaking change, wait for 3.0>

edsantiago:
Well, the good news is, in one hour of loop testing I got no failures. LGTM.

BTW I have a test for it in the works, will submit PR some time later, possibly with a `FIXME 8427` comment if this can't merge yet.

rhatdan:
We are holding up the merge of this PR until after 2.2 release.

rhatdan:
@containers/podman-maintainers PTAL This is ready to merge.

saschagrunert (approved):
/lgtm

PR #8431, opened 2020-11-20
Make c.networks() list include the default network

mheon:
This makes things a lot more clear - if we are actually joining a CNI network, we are guaranteed to get a non-zero length list of networks.

We do, however, need to know if the network we are joining is the default network for inspecting containers as it determines how we populate the response struct. To handle this, add a bool to indicate that the network listed was the default network, and only the default network.

Also, add further differentiation in the DB between containers that have, and do not have, a networks listing. This is necessary to ensure we can tell the difference between containers that disconnected from all their networks and containers created before we added the new table.


baude:
tested and it works

baude:
code LGTM

mheon:
Restarted one flake

PR #8436, opened 2020-11-20
Enable remote shell completion without a running endpoint

Luap99:
The problem is that we always unconditionally setup up the
`ContainerEngine/ImageEngine`. This requires an running
endpoint. Most completions (e.g. flag names) do not need
them and should not fail. This commit makes sure we only
setup the engines as needed in the completions.


rhatdan (approved):
LGTM

TomSweeneyRedHat:
LGTM

Issue #8433, opened 2020-11-20
APIv2 - created container has invalid path and args


psakar:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Container create via REST api is not able to start because it has invalid path and arguments
**Steps to reproduce the issue:**

1. start podman rest service
```
podman --log-level=debug system service -t 0 tcp:localhost:8880 2>&1>/tmp/podman-rest.log &
```
2. create container
```
[...]

Issue #7927, opened 2020-10-05
Support TestContainers


gastaldi:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

This is a follow-up of https://github.com/testcontainers/testcontainers-java/issues/2088. 
It would be nice if podman could be used with TestContainers.

mheon:
Can you link any current issues that need to be resolved here?

zhangguanzhang:
this issue https://github.com/containers/podman/issues/7923 
I was unable to recreate on master, I believe it was taken care of in https://github.com/containers/podman/commit/cde367c1c0b348366fcaf4d8fac3b753de8456d5

ruddy32:
"//vat/run/docker.sock" is not setup dynamically when API start.
This seems to be a problem with TestContainer.

rhatdan:
You need to install podman-docker, for this to happen.

bsideup:
Hey @rhatdan,

does it mean that all known issues with Podman and Testcontainers are fixed now and that 100% of TC's test suite passes with Podman? 🎉

rhatdan:
I have no idea.  But we need specific issues reported to what is broken.  No one on the Podman team is concentrating on testing TestContainers.  Community members have been taking the lead on this, and reporting issues, that we will attempt to fix ASAP.

psakar:
I was able to successfully run test involving postgresql db using podman built from 2.2.0.RC1 with fixes from  https://github.com/containers/podman/pull/8440, https://github.com/containers/podman/pull/8429 and https://github.com/containers/podman/pull/8423

podman (rootless) started with 
```
podman system service -t 0 tcp:localhost:8880 &
```

test containers started with following environment variables set 
```
DOCKER_HOST=tcp://localhost:8880
TESTCONTAINERS_RYUK_DISABLED=true
```

ruddy32:
Applying changes to **/usr/lib/systemd/user/podman.service** with
`ExecStart=/usr/bin/podman system service -t 0 tcp://localhost:8880`
and remove socket Requires and After instruction

... and updating maven build configuration with
`			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-surefire-plugin</artifactId>
				<configuration>
					<systemPropertyVariables>
						<java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
					</systemPropertyVariables>
					<environmentVariables>
[...]

PR #8440, opened 2020-11-21
APIv2 - wrong command and args for created container

psakar:
fix wrong check for entrypoint which caused wrong path and args for created container
fixes https://github.com/containers/podman/issues/8433


psakar:
@mheon You was right. It was the place you have pointed at

psakar:
Related to https://github.com/containers/podman/issues/7927

TomSweeneyRedHat:
All kinds of test unhappiness, not sure if it's related.

mheon:
Theory: the existing code is correct, but we're getting an entrypoint of an empty array (instead of nil), even when the client doesn't pass the entrypoint parameter in the input.

psakar:
@mheon I've seen empty string in entrypoint during debug (in Intellij Idea) AFAICR . I'll try to advance my go skills and debug failing test to see where the problem is.

psakar:
@mheon you was right. Empty value can be provided for entrypoint signaling the image entry point should be overridden. 
I've fixed the PR to use nil as default for entrypoint parsed from operation input

TomSweeneyRedHat:
LGTM
and happy green test buttons

Issue #8374, opened 2020-11-17
podman-remote cannot send tar content with correct dockerfile


kazimsarikaya:
Hello,
There is a two bug while sending tar to remote with podman-remote. 

First one is at: https://github.com/containers/podman/blob/master/pkg/bindings/images/build.go#L143 
~~~
params.Set("dockerfile", filepath.Base(containerFiles[0]))
~~~
docker file can be at anywhere so I don't think we set dockerfile parameter as only the file name. 

The second one is at: https://github.com/containers/podman/blob/master/pkg/bindings/images/build.go#L219
~~~
if !info.Mode().IsRegular() || path == s {
~~~
sources[0] is dockerfile and **path==s** always will be true. hence dockerfile will not be added to tar.

I use my own fork with tar compression with gzip and dockerignore parsing at https://github.com/kazimsarikaya/podman/tree/fix-send-tar 

I didn't create any push request however because I use additional gzip and dockerignore enhancements.

zhangguanzhang:
could you show the steps to reproduce it?

kazimsarikaya:
I started a podman system on a linux machine with:
~~~
/usr/local/bin/podman system service --time=0 tcp:192.168.99.114:1928
~~~

on osx machine I send a build request with:

~~~
declare -x CONTAINER_HOST="tcp://192.168.99.114:1928"
podman-remote  build -f docker/test.Dockerfile --platform linux/amd64  . 
~~~

And after sending context, the returned result is about not finding **test.Dockerfile**. Dockerfile path is changed. It should be **docker/test.Dockerfile**. The line 

~~~
params.Set("dockerfile", filepath.Base(containerFiles[0]))
~~~

removes directroy name. Probably the aim is always sending dockerfile if it is also removed by .dockerignore.

And also I reading/testing the code, nTar functions parameter sources contains docker file at index 0. for loop's first try with filepath.Walk, the line 

~~~
if !info.Mode().IsRegular() || path == s {
~~~

always returns true. because path and s is both the dockerfile. I found it when I implement .dockerignore when sending context as tar file.

The problem occurs when dockerfiles are inside subdirectories.

zhangguanzhang:
https://github.com/kazimsarikaya/podman/commit/656526463d447f67933d600d14383ed5bb8b21c6
@rhatdan PTAL

rhatdan:
@zhangguanzhang Are you opening a PR?

kazimsarikaya:
I open an pr.

[...]

PR #8456, opened 2020-11-23
podman remote send tar

kazimsarikaya:
It'is a pr about the issue #8374

### DIFF
--- a/pkg/bindings/images/build.go
+++ b/pkg/bindings/images/build.go
@@ -2,6 +2,7 @@ package images
 
 import (
 	"archive/tar"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"io"
@@ -18,6 +19,7 @@ import (
 	"github.com/containers/podman/v2/pkg/auth"
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/storage/pkg/fileutils"
 	"github.com/docker/go-units"
 	"github.com/hashicorp/go-multierror"
 	jsoniter "github.com/json-iterator/go"
@@ -138,12 +140,38 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
 	entries := make([]string, len(containerFiles))
 	copy(entries, containerFiles)
 	entries = append(entries, options.ContextDirectory)
-	tarfile, err := nTar(entries...)
+
+	excludes := options.Excludes
+	if len(excludes) == 0 {
+		excludes, err = parseDockerignore(options.ContextDirectory)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	tarfile, err := nTar(excludes, entries...)
 	if err != nil {
+		logrus.Errorf("cannot tar container entries %v error: %v", entries, err)
 		return nil, err
 	}
 	defer tarfile.Close()
-	params.Set("dockerfile", filepath.Base(containerFiles[0]))
+
+	containerFile, err := filepath.Abs(entries[0])
+	if err != nil {
+		logrus.Errorf("cannot find absolute path of %v: %v", entries[0], err)
+		return nil, err
+	}
+	contextDir, err := filepath.Abs(entries[1])
+	if err != nil {
+		logrus.Errorf("cannot find absolute path of %v: %v", entries[1], err)
+		return nil, err
+	}
+
+	if strings.HasPrefix(containerFile, contextDir+string(filepath.Separator)) {
+		containerFile = strings.TrimPrefix(containerFile, contextDir+string(filepath.Separator))
+	}
+
+	params.Set("dockerfile", containerFile)
 
 	conn, err := bindings.GetClient(ctx)
 	if err != nil {
@@ -200,52 +228,116 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
 	}
 }
 
-func nTar(sources ...string) (io.ReadCloser, error) {
+func nTar(excludes []string, sources ...string) (io.ReadCloser, error) {
+	pm, err := fileutils.NewPatternMatcher(excludes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error processing excludes list %v", excludes)
+	}
+
 	if len(sources) == 0 {
 		return nil, errors.New("No source(s) provided for build")
 	}
 
 	pr, pw := io.Pipe()
-	tw := tar.NewWriter(pw)
+	gw := gzip.NewWriter(pw)
+	tw := tar.NewWriter(gw)
 
 	var merr error
 	go func() {
 		defer pw.Close()
+		defer gw.Close()
 		defer tw.Close()
 
 		for _, src := range sources {
-			s := src
-			err := filepath.Walk(s, func(path string, info os.FileInfo, err error) error {
+			s, err := filepath.Abs(src)
+			if err != nil {
+				logrus.Errorf("cannot stat one of source context: %v", err)
+				merr = multierror.Append(merr, err)
+				return
+			}
+
+			err = filepath.Walk(s, func(path string, info os.FileInfo, err error) error {
 				if err != nil {
 					return err
 				}
-				if !info.Mode().IsRegular() || path == s {
-					return nil
-				}
 
-				f, lerr := os.Open(path)
-				if lerr != nil {
-					return lerr
+				if path == s {
+					return nil // skip root dir
 				}
 
 				name := strings.TrimPrefix(path, s+string(filepath.Separator))
-				hdr, lerr := tar.FileInfoHeader(info, name)
-				if lerr != nil {
-					f.Close()
-					return lerr
+
+				excluded, err := pm.Matches(filepath.ToSlash(name)) // nolint:staticcheck
+				if err != nil {
+					return errors.Wrapf(err, "error checking if %q is excluded", name)
 				}
-				hdr.Name = name
-				if lerr := tw.WriteHeader(hdr); lerr != nil {
-					f.Close()
-					return lerr
+				if excluded {
+					return nil
 				}
 
-				_, cerr := io.Copy(tw, f)
-				f.Close()
-				return cerr
+				if info.Mode().IsRegular() { // add file item
+					f, lerr := os.Open(path)
+					if lerr != nil {
+						return lerr
+					}
+
+					hdr, lerr := tar.FileInfoHeader(info, name)
+					if lerr != nil {
+						f.Close()
+						return lerr
+					}
+					hdr.Name = name
+					if lerr := tw.WriteHeader(hdr); lerr != nil {
+						f.Close()
+						return lerr
+					}
+
+					_, cerr := io.Copy(tw, f)
+					f.Close()
+					return cerr
+				} else if info.Mode().IsDir() { // add folders
+					hdr, lerr := tar.FileInfoHeader(info, name)
+					if lerr != nil {
+						return lerr
+					}
+					hdr.Name = name
+					if lerr := tw.WriteHeader(hdr); lerr != nil {
+						return lerr
+					}
+				} else if info.Mode()&os.ModeSymlink != 0 { // add symlinks as it, not content
+					link, err := os.Readlink(path)
+					if err != nil {
+						return err
+					}
+					hdr, lerr := tar.FileInfoHeader(info, link)
+					if lerr != nil {
+						return lerr
+					}
+					hdr.Name = name
+					if lerr := tw.WriteHeader(hdr); lerr != nil {
+						return lerr
+					}
+				} //skip other than file,folder and symlinks
+				return nil
 			})
 			merr = multierror.Append(merr, err)
 		}
 	}()
 	return pr, merr
+}
+
+func parseDockerignore(root string) ([]string, error) {
+	ignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
+	if err != nil && !os.IsNotExist(err) {
+		return nil, errors.Wrapf(err, "error reading .dockerignore: '%s'", root)
+	}
+	rawexcludes := strings.Split(string(ignore), "\n")
+	excludes := make([]string, 0, len(rawexcludes))
+	for _, e := range rawexcludes {
+		if len(e) == 0 || e[0] == '#' {
+			continue
+		}
+		excludes = append(excludes, e)
+	}
+	return excludes, nil
 }
--- a/test/e2e/build_test.go
+++ b/test/e2e/build_test.go
@@ -234,7 +234,7 @@ RUN printenv http_proxy`
 	})
 
 	It("podman build and check identity", func() {
-		session := podmanTest.Podman([]string{"build", "-f", "Containerfile.path", "--no-cache", "-t", "test", "build/basicalpine"})
+		session := podmanTest.Podman([]string{"build", "-f", "build/basicalpine/Containerfile.path", "--no-cache", "-t", "test", "build/basicalpine"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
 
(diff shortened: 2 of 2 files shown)


rhatdan:
Could you rebase your commit, looks like you submitted a merge PR with your code.

Also you need to sign your PR.

git rebase -i origin
git commit --amend -s -a
git push --force

zhangguanzhang:
```
Skipped directories tunnel: pkg/api
pkg/bindings/images/build.go:264:2: Consider preallocating `excludes` (prealloc)
	var excludes []string
	^
pkg/bindings/images/build.go:220:14: ineffectual assignment to `err` (ineffassign)
			excludes, err := parseDockerignore(s)
			          ^
pkg/bindings/images/build.go:322:13: S1003: should use strings.Contains(".$", string(ch)) instead (gosimple)
		} else if strings.Index(".$", string(ch)) != -1 {
```
The format is a bit problematic

kazimsarikaya:
interesting fails
1) the test code at  https://github.com/containers/podman/blob/eaf7a26b3c8a129ff429841453f2827d6bc34d85/test/e2e/build_test.go#L197 has a bug. when inspect fails it should not continue, however it access the data[0] and panics.

2) build succeeded, however inspect failed.
```
         Running: podman-remote [options] build build/basicalpine --iidfile /tmp/podman_test185355255/idFile
         STEP 1: FROM alpine
         Completed short name "alpine" with unqualified-search registries (origin: /etc/containers/registries.conf)
[+0800s] Getting image source signatures
         Copying blob sha256:9d16cba9fb961d1aafec9542f2bf7cb64acfc55245f9e4eb5abecd4cdc38d749
[+0801s] Copying config sha256:961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
         Storing signatures
         STEP 2: COMMIT
         --> 96176967641
         961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
         Running: podman-remote [options] inspect 961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
         
         []
         Error: error inspecting object: no such object: "961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4\n"
```

I don't think they are related to me, am I wrong?

zhangguanzhang:
seem iidfile not work with this

kazimsarikaya:
> seem iidfile not work with this

No, It is not about iidfile. iidfile succeeded. podman-remote execute inspect command with contents of iidfile.  I also use this patched version at my envrionment without any exception. 

And, an interesting thing is, I did not understand why this test https://github.com/containers/podman/blob/eaf7a26b3c8a129ff429841453f2827d6bc34d85/test/e2e/build_test.go#L175 have been run 3 times. It'is the failed test. Firstly, I think it is about default behavior for failed tests, however the other failed tests have been run only once. 

It'is possible to get podman system service log file at debug level for that test? In my environment I cannot reproduce that.

zhangguanzhang:
@rhatdan  PTAL

rhatdan:
It looks like the tests never checks if inspect succeeds.

You need to add this, to make sure it succeeds, which I believe it did not.

rhatdan:
"961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4\n"
I believe the problem is the extra "\n" in the iidfile.

kazimsarikaya:
sorry for that,the git append all commits. :(

kazimsarikaya:
after rebase from upstream (this repo), the remote tests still fails with 

```
Error: error inspecting object: no such object: "961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4\n"
```

the iidfile contains extra \n at end and of it. and inspect fails. What should I do now?

kazimsarikaya:
I found the problem and fixed. it.

static build failed, but I do not think it is really about this pr.

kazimsarikaya:
> after rebase from upstream (this repo), the remote tests still fails with
> 
> ```
> Error: error inspecting object: no such object: "961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4\n"
> ```
> 
> the iidfile contains extra \n at end and of it. and inspect fails. What should I do now?

also i figure out the problem, rebase had created a conflict with commit https://github.com/containers/podman/commit/b5e15f18d01219f666cd30623fb3ec73f338d040 hence the tests were failed.

zhangguanzhang:
need add a test unit for it

kazimsarikaya:
> need add a test unit for it

ok, I can write tests. however I didn't understand tests folders structure to put unit tests for changes. Can you give some advice?

kazimsarikaya:
> > need add a test unit for it
> 
> ok, I can write tests. however I didn't understand tests folders structure to put unit tests for changes. Can you give some advice?

Is it enough to create a test case inside https://github.com/containers/podman/blob/master/test/e2e/build_test.go for building some dockers with dummy content?

I did not find a kind of tests for per function like build_test.go for functions inside build.go

rhatdan:
Yes the e2e tests is ok.

kazimsarikaya:
> Yes the e2e tests is ok.

@rhatdan 

I planed these tests:

1) build container/dockerfile inside not root of contextdir.

2a) put .containerignore file for filtering (both file and folder ignoring)
2b) put .dockerignore file for filtering (both file and folder ignoring)

priority of .containerignore will be more than .dockerignore.

Is there a easy way to check layer contents? for ignore I need to check layer content.

```
FROM alpine:latest
ADD . /testignore/ #need to check this layer's content
```
I think these tests are enough. However I am open to advises.

rhatdan:
SGTM

kazimsarikaya:
@rhatdan while I am working on tests I found this line is interesting: https://github.com/containers/podman/blob/master/test/e2e/build_test.go#L237

the -f parameter has arg Containerfile.path but it is inside **build/basicalpine** , I think parameter should be **build/basicalpine/Containerfile.path** like at https://github.com/containers/podman/blob/master/test/e2e/build_test.go#L203

I will also fix that.

pkg/bindings/images/build.go
```diff
@@ -143,7 +146,15 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
 		return nil, err
 	}
 	defer tarfile.Close()
-	params.Set("dockerfile", filepath.Base(containerFiles[0]))
+
+	containerFile, _ := filepath.Abs(entries[0])
+	contextDir, _ := filepath.Abs(entries[1])
```
zhangguanzhang, line 151: need handle the err

kazimsarikaya: Ok, I will handle it. Also I found that If Dockerfile/Containerfile is symlink outside context dir it could not be sended. I will look for that. Also podman adds symlinks inside context dir into container, however podman-remote not. I will fix this too.

kazimsarikaya: the none followed symlink structure is like that: 
image

hello3 is missing when building with podman-remote

zhangguanzhang: nice job

kazimsarikaya:
OK, now podman-remote build behavior is nearly same as docker. Only this kind of builds are failing:
Let assume folder hierarchy is same as below

```
.
├── Containerfile
└── test1
    ├── Containerfile -> ../Containerfile
    ├── hello
    ├── hello2
    ├── hello3 -> hello2
    └── subdir
```

These docker builds commands works well
```
docker build -t test -f test1/Containerfile test1
cd test1
docker build -t test -f Containerfile .
```

However podman failes beacause a depedency forcing symlink targets should be inside context dir. The related source line is: 

https://github.com/containers/storage/blob/44cf7aab45e8f84f0aca3766f6e485288d349af4/pkg/archive/archive.go#L668-L672



Also:
A test  named as **int podman fedora-32 root host** failed. however it's blood brothers **int podman fedora-33 root host**, **int podman ubuntu-2004 root host** and **int podman ubuntu-2010 root host** passed. I didn't understand why.

@zhangguanzhang can you trigger failed and cancelled tests again?

zhangguanzhang:
I'm on a business trip these days and may need someone else to do these things

zhangguanzhang:
you could clikc the detail and `View more details on Cirrus CI` Cirrus CI website, and you could see the button which named `re-run`

kazimsarikaya:
> you could clikc the detail and `View more details on Cirrus CI` Cirrus CI website, and you could see the button which named `re-run`

@zhangguanzhang thank you for replay. I re-run failed and skipped tests. probably I should not start optional release test, sorry for that.

So, what should I do now?

pkg/bindings/images/build.go
```diff
@@ -206,46 +226,226 @@ func nTar(sources ...string) (io.ReadCloser, error) {
...
+	}
+	// try with .dockerignore
+	logrus.Debugf(".containerignore not found try .dockerignore")
+	dignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("error reading .dockerignore: '%s'", err)
+	}
+	if err == nil {
+		logrus.Debugf(".dockerignore found using it")
+	}
```
zhangguanzhang, line 357: The code here does not describe the judgment of the situation when both files exist and the description of related documents

pkg/bindings/images/build.go
```diff
@@ -206,46 +226,226 @@ func nTar(sources ...string) (io.ReadCloser, error) {
...
+
+	ignore := string(cignore) + "\n" + string(dignore)
+
+	ignoreLines := strings.Split(ignore, "\n")
+	excludes := make([]string, 0, len(ignoreLines))
+	for _, e := range ignoreLines {
+		if len(e) == 0 || e[0] == '#' {
+			continue
```
zhangguanzhang, line 365: need trim left space

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		// Write target and fake files
+		targetPath, err := CreateTempDirInTempDir()
+		if err != nil {
+			os.Exit(1)
+		}
+		targetSubPath := filepath.Join(targetPath, "subdir")
+		err = os.Mkdir(targetSubPath, 0755)
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 265:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+			Expect(os.RemoveAll(targetPath)).To(BeNil())
+		}()
+
+		// make cwd as context root path
+		Expect(os.Chdir(targetPath)).To(BeNil())
+
+		session := podmanTest.Podman([]string{"build", "-t", "test", "-f", "Containerfile", targetSubPath})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		ok, _ := session.GrepString("/test/dummy")
```
zhangguanzhang, line 289:
```suggestion
		ok, err := session.GrepString("/test/dummy")
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		cwd, err := os.Getwd()
+		Expect(err).To(BeNil())
+
+		podmanTest.AddImageToRWStore(ALPINE)
+
+		// Write target and fake files
+		targetPath, err := CreateTempDirInTempDir()
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 311:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		targetSubPath := filepath.Join(targetPath, "subdir")
+		err = os.Mkdir(targetSubPath, 0755)
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 316:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		cwd, err := os.Getwd()
+		Expect(err).To(BeNil())
+
+		podmanTest.AddImageToRWStore(ALPINE)
+
+		// Write target and fake files
+		targetPath, err := CreateTempDirInTempDir()
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 355:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+
+		containerfilePath := filepath.Join(targetPath, "Containerfile")
+		err = ioutil.WriteFile(containerfilePath, []byte(containerfile), 0644)
+		Expect(err).To(BeNil())
+
+		targetSubPath := filepath.Join(targetPath, "subdir")
+		err = os.Mkdir(targetSubPath, 0755)
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 369:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		cwd, err := os.Getwd()
+		Expect(err).To(BeNil())
+
+		podmanTest.AddImageToRWStore(ALPINE)
+
+		// Write target and fake files
+		targetPath, err := CreateTempDirInTempDir()
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 471:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		targetSubPath := filepath.Join(targetPath, "subdir")
+		err = os.Mkdir(targetSubPath, 0755)
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 476:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/build_test.go
```diff
@@ -244,4 +244,276 @@ RUN printenv http_proxy`
...
+		dummyFile := filepath.Join(targetSubPath, "dummy")
+		err = ioutil.WriteFile(dummyFile, []byte("dummy"), 0644)
+		Expect(err).To(BeNil())
+
+		emptyDir := filepath.Join(targetSubPath, "emptyDir")
+		err = os.Mkdir(emptyDir, 0755)
+		if err != nil {
+			os.Exit(1)
+		}
```
zhangguanzhang, line 485:
```suggestion
		Expect(err).To(BeNil())
```

pkg/bindings/images/build.go
kazimsarikaya: ok, wont use this approach. from documentation I found that there is command line argument **ignorefile** https://github.com/containers/podman/blob/master/docs/source/markdown/podman-build.1.md#--ignorefile 

however, I cannot found its reference at BuildOptions. Do you have any idea?

zhangguanzhang:
```
./bin/podman build --help |& grep -- --ignore
      --ignorefile string                            path to an alternate .dockerignore file
```
could see it at the latest code

kazimsarikaya: Ok, let me explain, the problem is **--ignorefile** is **buildah**'s internal option as used here: https://github.com/containers/buildah/blob/59cd527d6d40b5a77d7f3e6560b90c2ef7102c93/cmd/buildah/bud.go#L357-L363 buildah reads ignorefile and sets its lines to **options.Excludes**.  **podman** appends buildah's bud args to its build args converts some of them to **imagebuildah.BuildOptions**. And now --ignorefile never applied into Excludes. I put code to check Excludes if empty. If empty than retry read .dockerignore at: https://github.com/containers/podman/pull/8456/files#diff-43fd97ffbb9420a1f310fa0f6a07e2c6ea1d4deb0eefc76ae13812102c11c65fR144

pkg/bindings/images/build.go
```diff
@@ -200,52 +228,116 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
...
+	rawexcludes := strings.Split(string(ignore), "\n")
+	excludes := make([]string, 0, len(rawexcludes))
+	for _, e := range rawexcludes {
+		if len(e) == 0 || e[0] == '#' {
+			continue
+		}
+		excludes = append(excludes, e)
+	}
+	return excludes, nil
+}
```
TomSweeneyRedHat, line 343: Could/Should we call `./vendor/github.com/openshift/imagebuilder/builder.go:// ParseDockerIgnore ` instead?  Seems to be the same code, I'm just not sure if we should include an imagebuilder dependency, even though the Buildah code has one already that's in the build functionality.

pkg/bindings/images/build.go
```diff
@@ -200,52 +228,116 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
...
+					hdr, lerr := tar.FileInfoHeader(info, link)
+					if lerr != nil {
+						return lerr
+					}
+					hdr.Name = name
+					if lerr := tw.WriteHeader(hdr); lerr != nil {
+						return lerr
+					}
+				} //skip other than file,folder and symlinks
+				return nil
```
TomSweeneyRedHat, line 321: We shouldn't allow for symlinks that point to files/directory higher than the context directory.  i.e. if the context directory is `/home/user/runpodman` than pointing to `/home/user` should be invalid and I'm not seeing checks for this?

pkg/bindings/images/build.go
kazimsarikaya: There is two reason for that. 

1) the --ignorefile parameter does not have effect. Firstly I tried to get IgnoreFile option parameter, however It is internal parameter of buildah. Normally this method should have a second paramter named as ignorefile. Then if the parameter is empty string it should use .dockerignore, else should use given ignore file. At my previous comments I mention that. podman build command uses buildah's bud commands parameters at its parameters however it does not handle all of them. Hence I previously asked how to get "--ignorefile" argument value. Then modify the code.

2) I don't know why, using imagebuilder.ParseDockerIgnore breaks down tests. especially vendoring. 

When I looked at buildah's bud command, the --ignorefile parsed at cmd and setted to the options.Excludes. We need to do same as that? I asked it before, However I didnot get a proper answer.

pkg/bindings/images/build.go
kazimsarikaya: This is more interesting to. I also asked for that too. However no proper answer. podman's local build does allow symlinks outside the contextdir and only copies symkink's itself. However containers/storage untar method checks if symlink target is outsided or not. hence i didnot put a check here. 

Also docker build allows this kind of symlinks. podman's build command should be identical to docker's build or not?

The problem is like that. When developer intentionally  creates symlinks such as targeted /etc /var etc? 

Local podman build's allow this however remote podman not. 

Previously I asked that how we handle this?

kazimsarikaya:
@TomSweeneyRedHat  I write some comments to your reviews. I switched docker to podman nowadays. I use OSX development and remote servers for containers. After this switch lots of my docker builds started to fail. It's not about parameters/arguments. I am ok with that. However podman does not handle context dir and containerfile as docker handles it. 

I also asked, however I didnot get an answer about tihs question too: the **-f** parameter of docker build reads file contents and sends it to the remote side. hence this behavior allows developers to symlink dockerfile to outside contextdir. It's ok to know that podman will support it or not. However I didnot get an answer too.

Lot's of things are unclear. Hence I do not know what should I do. I need a image build tool whose behaviors are identical to dockers'. The aim of podman is that or not?

rhatdan:
@vrothberg @mheon @baude @jwhonce PTAL

pkg/bindings/images/build.go
TomSweeneyRedHat: @kazimsarikaya If there's a check within c/storage already, then I'm fine with that and you don't need to add one here.

Others may have different opinions, but from what I've been told, symlinks outside of the context-directory can cause potential security issues.  We do indeed try to match Docker's functionality, but not if it creates a security hole in the process.

kazimsarikaya: @TomSweeneyRedHat yes, if i send symlink outside the context-dir podman which runs at remote host returns exception because of https://github.com/containers/storage/blob/master/pkg/archive/archive.go#L668-L672 and build fails.

rhatdan:
@kazimsarikaya Podman should support -f.  Pretty much we should support everything that docker build does, if not it is a bug or should be documented as a difference.

kazimsarikaya:
> @kazimsarikaya Podman should support -f. Pretty much we should support everything that docker build does, if not it is a bug or should be documented as a difference.

ok, podman remote does not support -f parameter as following symlinks and send send them separately from context dir.

rhatdan:
@kazimsarikaya Could you open an issue on that, or maybe better a PR?  :^)

kazimsarikaya:
@rhatdan i can fix that issue with with pr too. because i also handle several docker -f differences with this pr.

also i am waiting an answer too. I commented on reviews however i didnot get proper answer. The problem is that:

* buildah has a internal parameter, not exists in buildconfig, --ignorefile 
* buildah parses that parameter before creating buildconfig and appends that ignore lines to buildconfig's excluedes parameter. 
* however podman did not achive that. 

I mention that at comment: https://github.com/containers/podman/pull/8456#discussion_r551134761 also @TomSweeneyRedHat  asked that at https://github.com/containers/podman/pull/8456#discussion_r551542859 and I also  replied with the same question.

I can easily to perform changes to  the cmd  before https://github.com/containers/podman/blob/master/cmd/podman/images/build.go#L403

pkg/bindings/images/build.go
rhatdan: SGTM

pkg/bindings/images/build.go
rhatdan: The point is about remote attack.  If I put a symbolic link into the copy command and the path points to /etc/shadow on the host, I don't want that added to the container image.  So I think on the podman-remote build we need to make sure that links in the  context directory stay within the directory.  Are you saying that Docker allows you to send across a sympbolic link to the server that points to the hosts /etc/shadow, and can embed it into the container image?

pkg/bindings/images/build.go
rhatdan: ./bin/podman-remote build --help | grep ignore
      --ignorefile string                            path to an alternate .dockerignore file

I guess the question is whether the ignorefile path should go over or the contents?

pkg/bindings/images/build.go
kazimsarikaya: Screen Shot 2021-01-14 at 00 54 41


For symlinks you only create a symlink, not copy the content. Hence for your example, the shadow will be linked to the /etc/shadow inside container, not the content of build host.

pkg/bindings/images/build.go
kazimsarikaya: No problem is, only buildah parses --ignorefile with https://github.com/containers/buildah/blob/59cd527d6d40b5a77d7f3e6560b90c2ef7102c93/cmd/buildah/bud.go#L357-L363 however podman doesnot do that. hence imagebuildah.BuildOptions.Excludes ara always empty. we should also put parsing ignorefile before https://github.com/containers/podman/blob/master/cmd/podman/images/build.go#L403

pkg/bindings/images/build.go
rhatdan: Ok SGTM

pkg/bindings/images/build.go
rhatdan: Yes I think the parsing should happen on the client side.

kazimsarikaya: Ok I will add cmd and re-push pr.

pkg/bindings/images/build.go
kazimsarikaya: However problem is at https://github.com/containers/storage/blob/master/pkg/archive/archive.go#L668-L672 hence this problem isnot related to this pr. hence it seems this conversation is resolved.

jwhonce:
LGTM

kazimsarikaya:
@rhatdan i am work on --ignorefile argument and making docker -f compilant. probably today i will finished them.

rhatdan:
@kazimsarikaya We can do that as a follow up PR.  As long as there are no changes to CLI and API. Today is feature freeze.

kazimsarikaya:
ok then i will make them a new pr. 
/hold cancel

Issue #8454, opened 2020-11-23
podman network create: add option to set custom network driver settings, eg like MTU


pschiffe:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Docker command `docker network create` has the `-o, --opt map              Set driver specific options (default map[])` option.

With podman, one have to modify the config file in `/etc/cni/` once the network is created to change its settings, which is not very automation friendly.

Related ansible collection RFE: https://github.com/containers/ansible-podman-collections/issues/147

rhatdan:
@pschiffe interested in opening a PR to fix this?

afbjorklund:
This would be useful for minikube as well, I can probably do a PR for it

rhatdan:
You got it.

afbjorklund:
Most likely it will be a simple option to "network create", rather than -o com.docker.network.driver.mtu=1500

afbjorklund:
Seems to be working:

```
$ sudo ./bin/podman network create podman1
/etc/cni/net.d/podman1.conflist
$sudo ./bin/podman network create podman2 --mtu 5000
/etc/cni/net.d/podman1.conflist
```

27: cni-podman1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu **1500** qdisc noqueue state UP group default qlen 1000

29: cni-podman2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu **5000** qdisc noqueue state UP group default qlen 1000

pschiffe:
Looking great and good enough for me, thank you!

afbjorklund:
Syntax now changed to:

`podman network create podman2 --opt mtu=5000`

pschiffe:
Resolved via #8457

Issue #7989, opened 2020-10-10
create network --label and run --restart unless-stopped are not compatible with Docker.


HarriL:
**Description**
/kind bug
You can not just install podman-docker and expect podman to be compatible.

 I installed podman-docker and got four compatibility issues from a Docker compatible installation kit.
	1. Build Bridge
	Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
	Error: unknown flag: --label
	2. Start container1
	Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
	Error: the unless-stopped restart policy is not supported: invalid argument
	3. Start container2
	Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
	Error: the unless-stopped restart policy is not supported: invalid argument
	4. Start container3
	Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
	Error: the unless-stopped restart policy is not supported: invalid argument 

**Steps to reproduce the issue:**
1. Install CentOS 8.2.2004
2. Install podman-docker
3. Run `docker network create --label "abc" --label "xyz" target-name
4. Run docker run ... --restart unless-stopped ...

**Describe the results you received:**
3. Error: unknown flag: --label
4. Error: the unless-stopped restart policy is not supported: invalid argument
**Describe the results you expected:**
3. Accept metadata "abc" and "zyz" in the network container as Docker defines and as other type of containers does.
4. The error message is wrong, this is not a restart. Actually this is **not restart** ! And Podman already works as this argument asks it to do! This argument means that after the "podman stop" command do not start this container in a reboot. And that is how podman works already. Remove the error and just do nothing. This way the behavior of podman is correct and the argument is handled "correctly".

**Additional information you deem important (e.g. issue happens only occasionally):**

#### Raises error:
> Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
> 	Error: unknown flag: --label

#### Expected to handle metadata argument(s): 
--label argument1 --label argument2
#### Docker definition: [Link to Docker CLI reference](https://docs.docker.com/engine/reference/commandline/network_create/)

The second one seems to be a  misunderstanding of what does Docker's argument unless-stopped really means/does.
#### Raises error:
	Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
	Error: the unless-stopped restart policy is not supported: invalid argument
[...]

PR #8457, opened 2020-11-23
Add podman network create flag for bridge mtu

afbjorklund:
See #8454

### DIFF
--- a/cmd/podman/networks/create.go
+++ b/cmd/podman/networks/create.go
@@ -30,6 +30,7 @@ var (
 var (
 	networkCreateOptions entities.NetworkCreateOptions
 	labels               []string
+	opts                 []string
 )
 
 func networkCreateFlags(cmd *cobra.Command) {
@@ -38,6 +39,10 @@ func networkCreateFlags(cmd *cobra.Command) {
 	driverFlagName := "driver"
 	flags.StringVarP(&networkCreateOptions.Driver, driverFlagName, "d", "bridge", "driver to manage the network")
 	_ = cmd.RegisterFlagCompletionFunc(driverFlagName, common.AutocompleteNetworkDriver)
+
+	optFlagName := "opt"
+	flags.StringArrayVarP(&opts, optFlagName, "o", []string{}, "Set driver specific options (default [])")
+	_ = cmd.RegisterFlagCompletionFunc(optFlagName, completion.AutocompleteNone)
 
 	gatewayFlagName := "gateway"
 	flags.IPVar(&networkCreateOptions.Gateway, gatewayFlagName, nil, "IPv4 or IPv6 gateway for the subnet")
@@ -92,6 +97,10 @@ func networkCreate(cmd *cobra.Command, args []string) error {
 	networkCreateOptions.Labels, err = parse.GetAllLabels([]string{}, labels)
 	if err != nil {
 		return errors.Wrap(err, "failed to parse labels")
+	}
+	networkCreateOptions.Options, err = parse.GetAllLabels([]string{}, opts)
+	if err != nil {
+		return errors.Wrapf(err, "unable to process options")
 	}
 	response, err := registry.ContainerEngine().NetworkCreate(registry.Context(), name, networkCreateOptions)
 	if err != nil {
--- a/docs/source/markdown/podman-network-create.1.md
+++ b/docs/source/markdown/podman-network-create.1.md
@@ -30,8 +30,9 @@ Driver to manage the network (default "bridge").  Currently only `bridge` is sup
 
 Set driver specific options.
 
-For the `bridge` driver the following options are supported: `mtu`.
+For the `bridge` driver the following options are supported: `mtu` and `vlan`.
 The `mtu` option sets the Maximum Transmission Unit (MTU) and takes an integer value.
+The `vlan` option assign VLAN tag and enables vlan\_filtering. Defaults to none.
 
 #### **--gateway**
 
--- a/docs/source/markdown/podman-network-create.1.md
+++ b/docs/source/markdown/podman-network-create.1.md
@@ -26,6 +26,13 @@ resolution.
 
 Driver to manage the network (default "bridge").  Currently only `bridge` is supported.
 
+#### **--opt**=*option*, **-o**
+
+Set driver specific options.
+
+For the `bridge` driver the following options are supported: `mtu`.
+The `mtu` option sets the Maximum Transmission Unit (MTU) and takes an integer value.
+
 #### **--gateway**
 
 Define a gateway for the subnet. If you want to provide a gateway address, you must also provide a
--- a/libpod/network/create.go
+++ b/libpod/network/create.go
@@ -92,6 +92,14 @@ func parseMTU(mtu string) (int, error) {
 	return m, nil
 }
 
+// parseVlan parses the vlan option
+func parseVlan(vlan string) (int, error) {
+	if vlan == "" {
+		return 0, nil // default
+	}
+	return strconv.Atoi(vlan)
+}
+
 // createBridge creates a CNI network
 func createBridge(name string, options entities.NetworkCreateOptions, runtimeConfig *config.Config) (string, error) {
 	var (
@@ -170,6 +178,11 @@ func createBridge(name string, options entities.NetworkCreateOptions, runtimeCon
 		return "", err
 	}
 
+	vlan, err := parseVlan(options.Options["vlan"])
+	if err != nil {
+		return "", err
+	}
+
 	// obtain host bridge name
 	bridgeDeviceName, err := GetFreeDeviceName(runtimeConfig)
 	if err != nil {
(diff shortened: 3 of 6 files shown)


rhatdan:
You need to update man pages

rhatdan:
Need tests, and I don't think this will work on podman-remote right now.

rhatdan:
@baude @mheon PTAL

cmd/podman/networks/create.go
```diff
@@ -42,6 +42,8 @@ func networkCreateFlags(cmd *cobra.Command) {
 
 	flags.BoolVar(&networkCreateOptions.Internal, "internal", false, "restrict external access from this network")
 
+	flags.IntVar(&networkCreateOptions.MTU, "mtu", 0, "MTU")
```
mheon, line 45: Should we do a more general options flag? It seems like there could be quite a lot of these.

mheon: (The Docker extremely-long-option format is probably unnecessary, but something like `--bridge-option mtu=8000`?

afbjorklund: I suppose there couldn't be _many_ more than what the CNI "bridge" exposes, but hopefully the user (OP) has more input...

pschiffe: In my use case I need only the MTU setting to sync it with an underlying network.
When looking at the bridge driver parameters, there's maybe one more good candidate for this - the `vlan` tag parameter..

docs/source/markdown/podman-network-create.1.md
```diff
@@ -45,6 +45,10 @@ must be used with a *subnet* option.
 Create a *Macvlan* based connection rather than a classic bridge.  You must pass an interface name from the host for the
 Macvlan connection.
 
+#### **--mtu**
+
+The MTU (integer, optional).
```
TomSweeneyRedHat, line 50: Chunked this together form Wikipedia, is this correct or needs adjusting?  I think we need a bit more info than what's here at the moment.

```suggestion
The Maximum Transmission Unit (MTU), the size of the largest protocol data unit that can be communicated in a single network-layer transaction (integer, optional).
```

cmd/podman/networks/create.go
rhatdan: I agree with @mheon Lets do this as --bridge-option and just add the two for now, which leaves us room for more in the future without exploding out CLI.

afbjorklund: The other commands use "opt" (instead of option), for instance the one from "podman volume create"

```
  -o, --opt stringArray   Set driver specific options (default [])
```

Like the ones from "run":

```
      --dns-opt strings                          Set custom DNS options
      --log-opt strings                          Logging driver options
      --security-opt stringArray                 Security Options
```

afbjorklund:
> Need tests, and I don't think this will work on podman-remote right now.

I will need some help with this (integration test, and podman remote support)

docs/source/markdown/podman-network-create.1.md
afbjorklund: Hopefully the middle road is OK, with the new options

cmd/podman/networks/create.go
rhatdan: SGTM

rhatdan:
@afbjorklund When you get this to a state that you are happy with, we can get someone to take it over if you want.

docs/source/markdown/podman-network-create.1.md
```diff
@@ -26,6 +26,13 @@ resolution.
 
 Driver to manage the network (default "bridge").  Currently only `bridge` is supported.
 
+#### **--opt**=*option*, **-o**
+
+Set driver specific options.
+
+For the `bridge` driver the following options are supported: `mtu`.
+The `mtu` option sets the Maximum Transmission Unit (MTU) and takes an integer value.
```
TomSweeneyRedHat, line 34: :+1:

afbjorklund:
I think this should work, and the flag matches the "network" which also uses drivers.

```
Usage:
  podman volume create [options] [NAME]

Options:
      --driver string     Specify volume driver name (default "local")
  -l, --label strings     Set metadata for a volume (default [])
  -o, --opt stringArray   Set driver specific options (default [])
```

```
Usage:
  podman network create [options] [NETWORK]

Options:
  -d, --driver string     driver to manage the network (default "bridge")
  -o, --opt stringArray   Set driver specific options (default [])
```

And then we don't need another option, for the second driver ?

----


Was there another flag that should be added as well ? I think "vlan" was mentioned

https://www.cni.dev/plugins/main/bridge/

* forceAddress (boolean, optional): Indicates if a new IP address should be set if the previous value has been changed. Defaults to false.
* **mtu (integer, optional): explicitly set MTU to the specified value.** Defaults to the value chosen by the kernel.
* promiscMode (boolean, optional): set promiscuous mode on the bridge. Defaults to false.
* **vlan (int, optional): assign VLAN tag.** Defaults to none.

The remaining feature is --label, but as mentioned that needs some more work:

https://github.com/containers/podman/issues/7989#issuecomment-725434675

Minikube is using labels, to identify which volumes and networks are OK to remove...

`--filter='label=created_by.minikube.sigs.k8s.io=true'`

The workaround is for the user to clean them up manually...

Volume does have --label, but the --filter was missing value. (~~#8341~~)

cmd/podman/networks/create.go
```diff
@@ -27,6 +29,9 @@ var (
 
 var (
 	networkCreateOptions entities.NetworkCreateOptions
+	opts                 = struct {
+		Opts []string
+	}{}
```
Luap99, line 34: Why is this a struct? A normal slice should be enough.

libpod/network/create.go
```diff
@@ -155,6 +164,11 @@ func createBridge(r *libpod.Runtime, name string, options entities.NetworkCreate
 		ipMasq = false
 	}
 
+	mtu, err := parseMTU(options.Options["mtu"])
+	if err != nil {
+		return "", err
+	}
+
```
Luap99, line 171: We should be parsing all options and return and error for unknown ones. `podman network create -o bogus` should not work.

libpod/network/create.go
```diff
@@ -80,6 +81,14 @@ func validateBridgeOptions(options entities.NetworkCreateOptions) error {
 
 }
 
+// parseMTU parses the mtu option
+func parseMTU(mtu string) (int, error) {
+	if mtu == "" {
+		return 0, nil // default
+	}
+	return strconv.Atoi(mtu)
```
Luap99, line 89: Should we error for negative values? I can do `-o mtu=-1` and it will be added to the config.

Luap99 (commented):
`podman-remote` works since we are using the full `NetworkCreateOptions` struct as json in the body.

For the integration tests I recommend adding this to `test/e2e/network_create_test.go`:
```
It("podman network create with mtu option", func() {
	net := "mtu-test"
	nc := podmanTest.Podman([]string{"network", "create", "--opt", "mtu=9000", net})
	nc.WaitWithDefaultTimeout()
	Expect(nc.ExitCode()).To(BeZero())
	defer podmanTest.removeCNINetwork(net)

	nc = podmanTest.Podman([]string{"network", "inspect", net})
	nc.WaitWithDefaultTimeout()
	Expect(nc.ExitCode()).To(BeZero())
	Expect(nc.OutputToString()).To(ContainSubstring(`"mtu": 9000,`))
})
```

cmd/podman/networks/create.go
afbjorklund: copy/paste only (it used to have "Labels" too)

libpod/network/create.go
afbjorklund: Good point, that should probably be validated. Supposedly CNI would also do some validation

libpod/network/create.go
afbjorklund: That would make a little more sensitive to version skew, but is probably the right thing

afbjorklund: This would also require podman to get all the available options for all the available drivers

Luap99: I'm thinking of something similar to this:
```
for k,val := range options.Options {
    switch k {
        case "mtu":
            parsemtu(val)
        default:
            return errors.Errorf("unsupported option %s",k)
    }
}
```
It's easy for someone to have a typo. We should error right away instead of having the user to debug why his config is not working as expected.

afbjorklund: But these are plugins, right ? So it could be extended ?

afbjorklund: We could do it for the bridge plugin like indicated

Luap99 (commented):
LGTM

docs/source/markdown/podman-network-create.1.md
```diff
@@ -26,6 +26,14 @@ resolution.
 
 Driver to manage the network (default "bridge").  Currently only `bridge` is supported.
 
+#### **--opt**=*option*, **-o**
+
+Set driver specific options.
+
+For the `bridge` driver the following options are supported: `mtu` and `vlan`.
+The `mtu` option sets the Maximum Transmission Unit (MTU) and takes an integer value.
+The `vlan` option assign VLAN tag and enables vlan\_filtering. Defaults to none.
```
TomSweeneyRedHat, line 35: Just triple checking, I think the '\` is intentional and needed?

TomSweeneyRedHat:
LGTM

docs/source/markdown/podman-network-create.1.md
afbjorklund: Sorry which one ? The backslash ? Editor was complaining about the "naked" slash, so I escaped it... Maybe needlessly so

TomSweeneyRedHat: oops, messed up the md prior.  `vlan\_filtering` is what I was asking about.  Seems to resolve OK, I'm just not sure it's necessary.

zhangguanzhang:
may be need to rebase to 1 commit?

rhatdan:
@afbjorklund needs a rebase, and then we can get this in.

PR #8460, opened 2020-11-23
WIP: Improve the cli args validation

Luap99:
Make use of the cobra `Args` function to validate the cli arguments.
This already exists for most commands. Using the `Args` function has
two advantages. First, it helps to reduce code duplication since most
commands can share the same `Args` function. Second, the shell
completion logic also uses this to provide better completions.
e.g. `podman start --latest [TAB]` should not complete containers

Remove useless definitions of `DisableFlagsInUseLine`. This is already
set for all commands in `cmd/podman/main.go`. Also remove
`TraverseChildren` since this has only affect on the root command.

### DIFF
--- a/cmd/podman/containers/attach.go
+++ b/cmd/podman/containers/attach.go
@@ -18,7 +18,7 @@ var (
 		Short:             "Attach to a running container",
 		Long:              attachDescription,
 		RunE:              attach,
-		Args:              validate.IDOrLatestArgs,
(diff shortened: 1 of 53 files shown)


rhatdan:
@Luap99 What is going on with this PR? at least needs a rebase.

rhatdan:
@Luap99 Another month, should we close this PR?

Luap99:
Yes, lets close for now

PR #8461, opened 2020-11-23
Implement shell completion for podman top

Luap99:


TomSweeneyRedHat:
LGTM
but would like a head nod from @edsantiago

PR #8468, opened 2020-11-24
[CI:DOCS] Tutorials:mac-win-client - Fix command ensuring sshd is enabled

lucendio:
`-s, --signal` requires a value and is probably not intended to be here since it would send a signal to the `sshd` process.


mheon:
/approve
@lucendio Can you add a signoff to your commit - `git commit --amend -s` and re-pushing should do the trick.

TomSweeneyRedHat:
LGTM
once signed, thanks @lucendio !

rhatdan:
/lgtm
Thanks @lucendio

PR #8470, opened 2020-11-24
Not use local image create/add manifest 

QiWang19:
Avoid using the image from load storage for `manifest create` and `manifest add`
since the local image does not include other entries of the list from the registry.
`--all` flag of `manifest create` and `manifest add` can not add all of the lists as expected.

### DIFF
--- a/libpod/image/manifests.go
+++ b/libpod/image/manifests.go
@@ -2,13 +2,14 @@ package image
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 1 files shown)


rhatdan:
Manifests tests are failing.

QiWang19:

saschagrunert (approved):
LGTM

vrothberg (commented):
/lgtm

Issue #6192, opened 2020-05-12
Provide metadata to identify a container from inside


debarshiray:
/kind feature

**Description**

Sometimes I come across people who are looking for ways to identify their [Toolbox](https://github.com/containers/toolbox) containers while sitting inside them. One use-case is to customize the shell prompt, just like one of those fancy Git prompts that identify the current branch and such.

I don't know what the ideal format would be.

Flatpak puts a INI-style file with stanzas at `/.flatpak-info` that describes the container. JSON is another option, but it isn't something that you can parse using the barebones POSIX shell utilities.

**Output of `podman version`:**

```
Version:            1.9.1
RemoteAPI Version:  1
Go Version:         go1.13.10
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.13.10
  podmanVersion: 1.9.1
host:
  arch: amd64
  buildahVersion: 1.14.8
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.15-1.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.15, commit: 4152e6044da92e0c5f246e5adf14c85f41443759'
  cpus: 4
  distribution:
    distribution: fedora
    version: "31"
  eventLogger: file
  hostname: kolache
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.4.14-200.notify.fc31.x86_64
  memFree: 1459437568
  memTotal: 8107806720
  ociRuntime:
    name: crun
    package: crun-0.13-2.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.13
      commit: e79e4de4ac16da0ce48777afb72c6241de870525
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-0.4.0-20.1.dev.gitbbd6f25.fc31.x86_64
    version: |-
      slirp4netns version 0.4.0-beta.3+dev
[...]

PR #8476, opened 2020-11-25
Add containerenv information to /run/.containerenv

rhatdan:
We have been asked to leak some information into the container
to indicate:
* The name and id of the container
* The version of podman used to launch the container
* The image name and ID the container is based on.
* Whether the container engine is running in rootless mode.

Fixes: https://github.com/containers/podman/issues/6192

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -26,9 +26,12 @@ Several files will be automatically created within the container. These include
 _/etc/hosts_, _/etc/hostname_, and _/etc/resolv.conf_ to manage networking.
 These will be based on the host's version of the files, though they can be
 customized with options (for example, **--dns** will override the host's DNS
-servers in the created _resolv.conf_). Additionally, an empty file is created in
-each container to indicate to programs they are running in a container. This file
-is located at _/run/.containerenv_.
+servers in the created _resolv.conf_). Additionally, a container environment
+file is created in each container to indicate to programs they are running in a
+container. This file is located at _/run/.containerenv_. When using the
+--privileged flag the .containerenv contains name/value pairs indicating the
+container engine version, whether the engine is running in rootless mode, the
+container name and id, as well as the image name and id that the container is based on.
 
 When running from a user defined network namespace, the _/etc/netns/NSNAME/resolv.conf_
 will be used if it exists, otherwise _/etc/resolv.conf_ will be used.
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -35,6 +35,7 @@ import (
 	"github.com/containers/podman/v2/pkg/rootless"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/containers/podman/v2/utils"
+	"github.com/containers/podman/v2/version"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
 	securejoin "github.com/cyphar/filepath-securejoin"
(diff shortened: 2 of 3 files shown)


libpod/container_internal_linux.go
```diff
@@ -1426,8 +1427,24 @@ func (c *Container) makeBindMounts() error {
 	// Make .containerenv
 	// Empty file, so no need to recreate if it exists
```
AkihiroSuda, line 1427: No longer empty

libpod/container_internal_linux.go
```diff
@@ -1426,8 +1427,24 @@ func (c *Container) makeBindMounts() error {
...
+		}
+		imageID, imageName := c.Image()
+
 		// Empty string for now, but we may consider populating this later
-		containerenvPath, err := c.writeStringToRundir(".containerenv", "")
+		// Populate the .containerenv with container information
+		containerenv := fmt.Sprintf(`\
+engine="podman-%s"
+name=%q
+id=%q
```
AkihiroSuda, line 1441: Exposing ID by default might be potentially dangerous. Could this be optional?

mheon:
I'm a little iffy about exposing full version numbers. Usual practice for
hardening is to disable that wherever possible to make it harder to
identify if your instance is affected by vulnerabilities. Full image ID
also has the same issue. I would not do those by default.

Also, I think your leading \ in the string you write to the file is
unnecessary.

libpod/container_internal_linux.go
rhatdan: Please explain how container ID is risky? 

We can add something to containers.conf to disable this, but I am not sure it is worth the effort.

rhatdan:
I am not crazy about security by obscurity. It is likely pretty easier for the processes within a container to figure out if the system is vulnerable.  Doing a checksum of a couple of executables makes it easy to figure out if you are vulnerable.

rhatdan:
@cgwalters @debarshiray @juanje @alexlarsson PTAL

libpod/container_internal_linux.go
AkihiroSuda: An attacker might want to know the container ID to determine the path of the attacker-crafted malicious file in the container rootfs.

Such path would be useful for the attacker when the attacker somehow managed to gain write access to e.g. /sys/kernel/uevent_helper.

debarshiray:
>> I'm a little iffy about exposing full version numbers. Usual practice for
>> hardening is to disable that wherever possible to make it harder to
>> identify if your instance is affected by vulnerabilities.
>
> It is likely pretty easier for the processes within a container to figure out
> if the system is vulnerable. Doing a checksum of a couple of executables
> makes it easy to figure out if you are vulnerable.

As long as images have the package (eg., RPM) database, then one can just do `rpm -q ...`. However, I can imagine images that don't have the database present.

rhatdan:
@debarshiray You want these fields for logging, and better reporting to the user, what is going on, correct?

docs/source/markdown/podman-run.1.md
```diff
@@ -26,9 +26,11 @@ Several files will be automatically created within the container. These include
 _/etc/hosts_, _/etc/hostname_, and _/etc/resolv.conf_ to manage networking.
 These will be based on the host's version of the files, though they can be
 customized with options (for example, **--dns** will override the host's DNS
-servers in the created _resolv.conf_). Additionally, an empty file is created in
-each container to indicate to programs they are running in a container. This file
-is located at _/run/.containerenv_.
+servers in the created _resolv.conf_). Additionally, an container environment
```
TomSweeneyRedHat, line 29:
```suggestion
servers in the created _resolv.conf_). Additionally, a container environment
```

docs/source/markdown/podman-run.1.md
```diff
@@ -26,9 +26,11 @@ Several files will be automatically created within the container. These include
...
+file is created in each container to indicate to programs they are running in a
+container. This file is located at _/run/.containerenv_ and containes name/value
```
TomSweeneyRedHat, line 31:
```suggestion
container. This file is located at _/run/.containerenv_ and contains name/value
```

TomSweeneyRedHat:
Couple nits in docs.  Once that and the security question is solved, LGTM.

debarshiray:
> You want these fields for logging, and better reporting to the user, what
> is going on, correct?

Yes, correct. eg., there is an interest to customize the shell prompt and such things to indicate which container and/or image is in play.

rhatdan:
@AkihiroSuda We are now only leaking the information in --privileged mode.

AkihiroSuda (approved):

TomSweeneyRedHat:
LGTM
one of the tests isn't hip and I'm not sure why.
Do you need to touch up https://github.com/containers/buildah/pull/2821 for the security issue @AkihiroSuda raised here?

Issue #8477, opened 2020-11-25
podman build produces "SHELL is not supported for OCI image format"


svdHero:
Hi there,

I have Dockerfiles containing the command
```
SHELL ["/bin/bash", "-c"] 
```
for setting the default shell to bash.

When I try to build images from said Dockerfiles with podman version 2.1.1, I get the error message:
"SHELL is not supported for OCI image format".

What should I do? Do I have to modify the Dockerfiles or is there another way to do this? I thought podman was a drop-in replacement for docker. Am I missing something or doing something wrong?

rhatdan:
You can use the --format docker flag, to support this option.  This option did not make it into the OCI specification.
Podman defaults to building OCI images.

rhatdan:
If you use buildah bud, you would get this message.
WARN SHELL is not supported for OCI image format, [/bin/bash -c] will be ignored. Must use `docker` format

svdHero:
Great. I did not know about the `--format` option. Thank you.
I will try to avoid non-OCI commands for new Containerfiles.

Ferryistaken:
Is there some kind of alternative to the SHELL command with OCI containers? I would like to build an OCI image but I don't really know how to transition.

mheon:
I can't find any documentation on what the `SHELL` command actually does - it's been removed from all current Dockerfile syntax references, which makes me think it's deprecated.

I think it may be similar to the `ENTRYPOINT` instruction?

Ferryistaken:
I tried building with `buildah` instead of `podman build` and it worked. It makes me think that I now have a docker formate image instead of an OCI image. How do I check if my image is OCI? Is there some tool that automatically does that?

mheon:
I believe `ManifestType` in `podman inspect` is what you're looking for. `application/vnd.docker.distribution.manifest.v2+json` for DockerV2, `application/vnd.oci.image.manifest.v1+json` for OCIv1

johnziebro:
> **You can use the --format docker flag, to support this option**. This option did not make it into the OCI specification.

I have a very similar warning with:
`CMD ["start.sh", "jupyter", "lab"]`
> WARN[0217] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format 

Do you have an example of how to use the paramater? I tried 
`docker build -t --format docker jupyterlab-datascience ./datascience`
> Error: accepts at most 1 arg(s), received 3

nezartarbin:
[...]

PR #8479, opened 2020-11-25
Drop default log-level from error to warn

rhatdan:
Our users are missing certain warning messages that would
make debugging issues with Podman easier.

For example if you do a podman build with a Containerfile
that contains the SHELL directive, the Derective is silently
ignored.

If you run with the log-level warn you get a warning message explainging
what happened.

$ podman build --no-cache -f /tmp/Containerfile1 /tmp/
STEP 1: FROM ubi8
STEP 2: SHELL ["/bin/bash", "-c"]
STEP 3: COMMIT
--> 7a207be102a
7a207be102aa8993eceb32802e6ceb9d2603ceed9dee0fee341df63e6300882e

$ podman --log-level=warn build --no-cache -f /tmp/Containerfile1 /tmp/
STEP 1: FROM ubi8
STEP 2: SHELL ["/bin/bash", "-c"]
STEP 3: COMMIT
WARN[0000] SHELL is not supported for OCI image format, [/bin/bash -c] will be ignored. Must use `docker` format
--> 7bd96fd25b9
7bd96fd25b9f755d8a045e31187e406cf889dcf3799357ec906e90767613e95f

These messages will no longer be lost, when we default to WARNing level.

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -71,7 +71,7 @@ var (
 		DisableFlagsInUseLine: true,
 	}
 
-	logLevel       = "error"
+	logLevel       = "warn"
 	useSyslog      bool
 	requireCleanup = true
 )
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -22,7 +22,7 @@ var (
 	// ChangeCmds is the list of valid Change commands to passed to the Commit call
 	ChangeCmds = []string{"CMD", "ENTRYPOINT", "ENV", "EXPOSE", "LABEL", "ONBUILD", "STOPSIGNAL", "USER", "VOLUME", "WORKDIR"}
 	// LogLevels supported by podman
-	LogLevels = []string{"debug", "info", "warn", "error", "fatal", "panic"}
+	LogLevels = []string{"debug", "info", "warn", "warning", "error", "fatal", "panic"}
 )
 
 type completeType int
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/containers/psgo v1.5.1
 	github.com/containers/storage v1.24.1
 	github.com/coreos/go-systemd/v22 v22.1.0
-	github.com/cri-o/ocicni v0.2.1-0.20201102180012-75c612fda1a2
+	github.com/cri-o/ocicni v0.2.1-0.20201125151022-df072ea5421c
 	github.com/cyphar/filepath-securejoin v0.2.2
 	github.com/davecgh/go-spew v1.1.1
 	github.com/docker/distribution v2.7.1+incompatible
@@ -73,5 +73,3 @@ require (
 	k8s.io/apimachinery v0.19.4
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 )
-
-replace github.com/cri-o/ocicni => github.com/cri-o/ocicni v0.2.1-0.20201109200316-afdc16ba66df
--- a/go.sum
+++ b/go.sum
@@ -124,8 +124,8 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:ma
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/creack/pty v1.1.9 h1:uDmaGzcdjhF4i/plgjmEsriH11Y0o7RKapEf/LDaM3w=
 github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
-github.com/cri-o/ocicni v0.2.1-0.20201109200316-afdc16ba66df h1:c35uRFkER07nAkB1X21e+PI5xO21SOyI6G7tdfvz1z4=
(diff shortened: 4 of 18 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/8477

rhatdan:
We might have to lower some Warning messages to Info or Debug, if they are not useful to the users.
Will see what the CI system finds.

rhatdan:
Buildah currently defaults to Warn level.

mheon:
Please do not merge this before 2.2 is cut. This is definitely a 3.0 change.

rhatdan:
Agreed, which is why I set the flag.

rhatdan:
Need https://github.com/cri-o/ocicni/pull/82 merged before we can move on with this.
Basically we need to make sure vendored applications do not Warn, when there is nothing the user can do about it.

cmd/podman/root.go
```diff
@@ -71,7 +71,7 @@ var (
 		DisableFlagsInUseLine: true,
 	}
 
-	logLevel       = "error"
+	logLevel       = "warn"
```
TomSweeneyRedHat, line 74: Is this something we let the user toggle via containers.conf?

TomSweeneyRedHat:
LGTM assuming happy tests

cmd/podman/root.go
rhatdan: Maybe, eventually although if we do this correct, I don't think they would want to, or ever need to.

rhatdan:
@mheon This change is causing some strange test differences.  
In logs_test "streaming output" test.

The container ends up in the "stopped" state rather then the "exited" state.  Any idea why a change of 
loglevel would cause this problem?

mheon:
It seems like whatever you did managed to break the cleanup process - the
container is probably not being unmounted either. Maybe a bad CLI argument
to said cleanup process? Ed wrote a script that intercepted the call to the
cleanup process and printed the arguments it was called with which could
help?

rhatdan:
I am seeing it locally also.  Will see if I can execute the cleanup command and if it fails.

edsantiago:
FWIW this fails root system tests on my laptop, lots of unexpected warnings:
```
time="2020-11-30T07:49:39-07:00" level=warning msg="Failed to add conmon to systemd sandbox cgroup: Unit libpod-conmon-3adb97f2b360f615f1bff55a0753624132f4ad5a551ceec87dc5504795c85a28.scope already exists."
```
and
```
time="2020-11-30T07:50:06-07:00" level=warning msg="Timed out stopping container 02c95917b41829234935840639cd69e5330922ae4336ff07cdc8fd82d5f89462, resorting to SIGKILL"
```

rhatdan:
The e2e test failures were caused on the cleanup generation.  
Basically
podman --log-level warn ...

Would generate a
podman --log-level warning container cleanup.

Then the cleanup line would silently fail,because "warning" was not accepted by the CLI.  This is now fixed.

rhatdan:
@edsantiago do you think we should just igore the warnings.  These seem to be valid warnings, that we have not been exposing to users up until now.  Perhaps we should drop these warnings back to Info if there is nothing the user can do about them, and they do not cause breakage?

mheon:
The stop one should definitely drop to Info.

edsantiago:
If my tests are suddenly seeing warnings that were never there before, then end users will suddenly start seeing warnings that were never there before. I would expect some/many of those to react with concern, and to waste time investigating, and to be upset when it turns out to be just an artifact of log level.

rhatdan:
@edsantiago the goal is to bubble the Warnings up to the user, when the user can take action.  If the user can not take action, then I believe that these should be dropped down to info.

The example that spawned this is building an OCI image with a Containerfile field that is not supported.  Currently we warn and then build the image without the field.  It would be better to warn the users so that they could fix their Containerfile without the option or build with --format docker.

Hopefully we will quickly flush out the log messages that should not be presented to the user.

edsantiago:
System tests are going to fail. On my system (f32) I'm still seeing:
```console
$ sudo ./bin/podman run -d --rm --name foo alpine sleep 20
9d9677fa601d46a99da0ed72ee3e5e7bea794e61c456243ca957ceec0ef27a8a
$ sudo ./bin/podman exec foo true
WARN[0000] Failed to add conmon to systemd sandbox cgroup: Unit libpod-conmon-9d9677fa601d46a99da0ed72ee3e5e7bea794e61c456243ca957ceec0ef27a8a.scope already exists.
```
I see this only as root, never as rootless. On the previous CI run, I saw failures in rootless but not root. That concerns me. And I really don't think users need or want to see this.

edsantiago:
Looking at [int podman f32 root container](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5182224708403200/html/int-podman-fedora-32-root-container.log.html) log, I see a lot of:
```
time="2020-11-30T16:23:35Z" level=warning msg="Error updating pod 363314e20620c674476220359e0e0c13cc03412c35136922eae4adbe9cc24d30 conmon cgroup /libpod_parent/363314e20620c674476220359e0e0c13cc03412c35136922eae4adbe9cc24d30/conmon PID limit: open /sys/fs/cgroup/libpod_parent/363314e20620c674476220359e0e0c13cc03412c35136922eae4adbe9cc24d30/conmon/pids.max: no such file or directory"
```

rhatdan:
@giuseppe Is this expected?  IE Should I drop the Warning to Info?

rhatdan:
@edsantiago @mheon PTAL

vrothberg (commented):
LGTM

Let's get it in, so it can cook for a while on our dev machines.

QiWang19 (commented):
LGTM

edsantiago:
No changes in tests? LGTM!

Issue #6244, opened 2020-05-15
Unsupported flag "--platform", multi arch support


onny:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

**Steps to reproduce the issue:**

1. ``docker run -it --name archlinux-cupsd --platform linux/amd64 archlinux``

**Describe the results you received:**

Unsupported flag "--platform"

**Describe the results you expected:**

Emulate target platform using Qemu. This seems to be an experimental feature on Docker but I would like to use it on Podman.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
ERRO[0000] cannot find mappings for user picloud: open /etc/subuid: no such file or directory 
Version:            1.9.2
RemoteAPI Version:  1
Go Version:         go1.14.2
Git Commit:         de2379a75863bdaf3ce8d414045d0b52c544f49a
Built:              Thu May 14 00:47:07 2020
OS/Arch:            linux/arm64
```

**Output of `podman info --debug`:**

```
ERRO[0000] cannot find mappings for user picloud: open /etc/subuid: no such file or directory 
debug:     
  compiler: gc 
  gitCommit: de2379a75863bdaf3ce8d414045d0b52c544f49a
  goVersion: go1.14.2         
  podmanVersion: 1.9.2
host:                         
  arch: arm64          
  buildahVersion: 1.14.8
  cgroupVersion: v1                                                        
  conmon:          
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.16, commit: a97780984207f29652af90fa3dc4e8e7576548e7'
  cpus: 4          
  distribution:           
    distribution: archarm
    version: unknown      
  eventLogger: file                                                        
  hostname: picloud   
  idMappings:    
    gidmap:                       
    - container_id: 0                                                      
      host_id: 1002               
      size: 1                                                              
    uidmap:                                                                
    - container_id: 0
      host_id: 1002          
      size: 1                   
  kernel: 5.0.11-1-ARCH    
  memFree: 1498345472               
  memTotal: 4169707520
  ociRuntime:  
   name: runc                      
    package: Unknown                                                       
    path: /usr/bin/runc                                                    
    version: |-                   
      runc version 1.0.0-rc10
      commit: dc9208a3303feef5b3839f4323d9beb36df0a9dd
      spec: 1.0.1-dev
  os: linux
  rootless: true
  slirp4netns:
[...]

PR #8488, opened 2020-11-25
Add support for --platform

rhatdan:
For docker compatibility we need to support --platform
flag.

podman create --platform
podman run --platform
podman pull --platform

Since we have --override-os and --override-arch already
this can be done just by modifying the client to split
the --platform call into os and arch and then pass those
options to the server side.

Fixes: https://github.com/containers/podman/issues/6244


vrothberg (commented):
LGTM, let's merge after 2.2

umohnani8:
LGTM

PR #8494, opened 2020-11-26
More docker compat API fixes

mlegenovic:
Fixes wrong VirtualSize, ParentId, Architecture, Author, Os and OsVersion value

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -1222,6 +1222,11 @@ func (i *Image) inspect(ctx context.Context, calculateSize bool) (*inspect.Image
 		}
 	}
 
+	parent, err := i.ParentID(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	repoTags, err := i.RepoTags()
 	if err != nil {
 		return nil, err
@@ -1248,6 +1253,7 @@ func (i *Image) inspect(ctx context.Context, calculateSize bool) (*inspect.Image
 
 	data := &inspect.ImageData{
 		ID:           i.ID(),
+		Parent:       parent,
 		RepoTags:     repoTags,
 		RepoDigests:  repoDigests,
 		Comment:      comment,
@@ -1258,10 +1264,12 @@ func (i *Image) inspect(ctx context.Context, calculateSize bool) (*inspect.Image
 		Config:       &ociv1Img.Config,
 		Version:      info.DockerVersion,
 		Size:         size,
-		VirtualSize:  size,
-		Annotations:  annotations,
-		Digest:       i.Digest(),
-		Labels:       info.Labels,
+		// This is good enough for now, but has to be
+		// replaced later with correct calculation logic
+		VirtualSize: size,
+		Annotations: annotations,
+		Digest:      i.Digest(),
+		Labels:      info.Labels,
 		RootFS: &inspect.RootFS{
 			Type:   ociv1Img.RootFS.Type,
 			Layers: ociv1Img.RootFS.DiffIDs,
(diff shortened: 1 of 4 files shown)


libpod/image/image.go
```diff
@@ -923,6 +923,11 @@ func (i *Image) Size(ctx context.Context) (*uint64, error) {
 	return nil, errors.Wrap(err, "unable to determine size")
 }
 
+//VirtualSize returns the virtual size of the image
+func (i *Image) VirtualSize(ctx context.Context) (*uint64, error) {
+	return i.Size(ctx)
```
mheon, line 928: The doubled call to `i.Size()` should be avoided - it's a very expensive operation.

mlegenovic: @mheon What would you propose:
- to make a new field which is lazy initialized, and if called a second time no calculation is needed (I suppose image size cannot change)
- to remove VirtualSize function, and to always use Size because both attributes always have the same values. At the moment Size == VirtualSize is more or less everywhere I saw in code, but I do not know if this is just temporary solution, or if this will change. I did VirtualSize function just to have an abstraction for calculation, but if both values are always the same, it is not needed.

Or do you have any other ideas?

mheon:
For now I would say remove VirtualSize and reuse size, with a comment about
how this is not strictly correct and should be investigated later.

mheon:
Test failures are consistent and look legitimate, but the error doesn't seem to make much sense - missing top layers for images we built ourselves make no sense?

@vrothberg Mind taking a look here?

mlegenovic:
@mheon I assume the problem is in GetParent function. Reason for the assumption is this comment:

```
// FIXME: GetParent() panics
// parent, err := l.GetParent(context.TODO())
```
which I have deleted because of small refactoring. GetParent is now called in my change-request every time when ParentId is determined. That is one of the fixed fields.

GetParent is also called in Remove function, but maybe this function is not called in a test, and therefore problem never occurs.

I will try to find out something with the debugger (just had no time to do it before).

mlegenovic:
It is very strange. If I run podman command that is executed in test-code by hand, everything works. Then I saw test-code calls podman with vfs as storage backend, and I thought maybe that is the problem... but it is not. If I run test with `go test <all_args>` it fails.

What I noticed is, `Runtime.layerTree()` function has this `tree.nodes` map. If you run podman command by hand, content of this map is different (it has items) then if it is called by test-code (no items in map), but both are inspecting the same image.

mheon:
This sounds like it could be because of the added store for image caching. We use an extra store for cached images (as described in https://www.redhat.com/sysadmin/image-stores-podman) in CI to speed things up (with an option to manually disable said caching for tests it breaks). We may need to disable the cache for these tests.

mlegenovic:
Can I disable this cache somehow locally? I run test with
`$ go test -v test/e2e/libpod_suite_test.go test/e2e/common_test.go test/e2e/config.go test/e2e/config_amd64.go test/e2e/build_test.go`
Is there an option to disable it?

mheon:
@baude is the expert on this, so he might be able to help

mlegenovic:
@mheon I have found the reason for the failed test. "podman build and check identity" test in `test/e2e/build_test.go` calls podman once with `podmanTest.Podman()` and a second time with `podmanTest.PodmanNoCache()`. This combination does not work. It works if you run both commands with `podmanTest.Podman()` or with `podmanTest.PodmanNoCache()`, otherwise `layers` array is empty, and test fails as consequence.

rhatdan:
@mlegenovic Could you rebase please.

mheon:
The one failure looks like a flake
/approve

jwhonce:
LGTM

PR #8266, opened 2020-11-07
Adding support for podman play volumes

aidan-plenert-macdonald:
To resolve https://github.com/containers/podman/issues/5788

Issue #5788, opened 2020-04-10
Play kube support for volumes created with `podman volume create`


frezbo:
Play kube support for volumes created with `podman volume create`

/kind feature

**Description**
Play kube support for volumes created with `podman volume create`

rhatdan:
This currently does not work?  What happens?
@kunalkushwaha @sujil02 PTAL

mheon:
We have support for bind-mount (and I think tmpfs) volumes, but not named volumes. Those generate additional bits of Kubernetes YAML, which we have to deal with in both `generate kube` and `play kube` (and there was some debate as to exactly what YAML was required). Probably not particularly difficult, but someone more familiar with Kube would be a good fit for this.

kunalkushwaha:
I can look into this.

rhatdan:
@kunalkushwaha Any progress on this?

rhatdan:
@mheon @haircommander Any ideas on this?

mheon:
On our end, the real difficulty is figuring out the format for Kube volume YAML. Once we have that, I think it's a pretty simple mapping.

haircommander:
named volumes sound like [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir). am I correct in saying so?

mheon:
Hmm. That does sound pretty close. I'll take a look.

frezbo:
@haircommander That seems fine to me too, since podman runs on the host.

rhatdan:
@ryanchpowell Could you work with @haircommander on this?

rhatdan:
@mheon @ashley-cui Could you take a look?

xordspar0:
The problem with emptyDir is that its lifetime is the same as the pod. When the pod is created, a new empty volume is created. When the pod is destroyed, the volume is deleted forever. emptyDir is a scratch space for running applications. It's similar to `/tmp`, and it can optionally be memory-backed, similar to `/tmp`.

podman volumes are independent of pods; they can be created and destroyed on their own. This is more similar to Kubernetes [PersistentVolumeClaims](https://kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim), which is an API designed for abstracting over vendor-specific storage like S3 or Google Cloud Storage.

aidan-plenert-macdonald:
[...]

PR #8497, opened 2020-11-27
Add support for persistent volume claims in kube files 

AlbanBedel:
This PR add simple support for persistent volume claims (PVC) to kube files by simply using a named volume with the name of the claim. The first commit rework the code for volumes in kube to allow supporting several types and the second one then add the new volume type.

This should solve bug #5788, but it doesn't overlap with PR #8266 which afaict is about creating the volumes and not using them.

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v2/libpod"
 	"github.com/containers/podman/v2/libpod/image"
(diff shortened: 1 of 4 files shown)


rhatdan:
Thanks @AlbanBedel 
You need to gofmt your code.

@haircommander @mheon PTAL

AlbanBedel:
I just found about `make gofmt`, CONTRIBUTING.md doesn't mention it and I had no luck with lint and validate in the gate container up to now. Currently it doesn't start at all (ERROR: Expecting libpod repository root at /usr/src/libpod (gate container entrypoint)) and last time I tried it froze my machine after a few minutes and then the container crashed.

zhangguanzhang:
maybe need rebase?

baude:
lgtm

Issue #5376, opened 2020-03-02
feature request: network create does not have "--label"


medyagh:
in minikube we recently added podman as a driver, however we can not create a network with a label like for docker.

the reason for creating a network with a label is, minikube delete should be able to clean up after itself and delete all the created containers and networks.

```
 podman network create podman1

Flags:
      --disable-dns      disable dns plugin
  -d, --driver string    driver to manage the network (default "bridge")
      --gateway ip       IPv4 or IPv6 gateway for the subnet
      --internal         restrict external access from this network
      --ip-range ipNet   allocate container IP from range
      --macvlan string   create a Macvlan connection based on this device
      --subnet ipNet     subnet in CIDR format
```

is there any chance to match docker on this feature ?

```
Usage:	docker network create [OPTIONS] NETWORK

Create a network

Options:
      --attachable           Enable manual container attachment
      --aux-address map      Auxiliary IPv4 or IPv6 addresses used by Network driver (default map[])
      --config-from string   The network from which copying the configuration
      --config-only          Create a configuration only network
  -d, --driver string        Driver to manage the Network (default "bridge")
      --gateway strings      IPv4 or IPv6 Gateway for the master subnet
      --ingress              Create swarm routing-mesh network
      --internal             Restrict external access to the network
      --ip-range strings     Allocate container ip from a sub-range
      --ipam-driver string   IP Address Management Driver (default "default")
      --ipam-opt map         Set IPAM driver specific options (default map[])
      --ipv6                 Enable IPv6 networking
      --label list           Set metadata on a network
  -o, --opt map              Set driver specific options (default map[])
      --scope string         Control the network's scope
      --subnet strings       Subnet in CIDR format that represents a network segment
```

baude:
is it not sufficient to podman network rm -f <name_of_network>?

mheon:
Can you clarify the use case here a bit? Are you using labels and filtering with `podman network ls --filter` to identify what networks need to be removed?

rhatdan:
@kunalkushwaha Interested in working on this? @sujil02 ?

hitchhooker:
> Can you clarify the use case here a bit? Are you using labels and filtering with `podman network ls --filter` to identify what networks need to be removed?
`podman network ls --filter ` does there even exist such flag as --filter for network? I keep receiving unknown flag error when trying to replace docker with podman for matrix-docker-ansible-deploy. 

`podman network ls -q --filter='name=^{{ matrix_podman_network }}$'
[...]

PR #8505, opened 2020-11-28
podman network label support

Luap99:
Add label support for podman network create. Use the `args`
field in the cni config file to store the podman labels.
Use `podman_labels` as key name and store the labels as
map[string]string.

For reference: https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config
https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md#network-configuration

Example snippet:

```
...
"args": {
	"podman_labels": {
		"key1":"value1",
		"key2":"value2"
	}
}
...
```

Make podman network list support several filters. Supported filters are name,
plugin, driver and label. Filters with different keys work exclusive. Several label
filters work exclusive and the other filter keys are working inclusive.

Also adjust the compat api to support labels in network create and list.

Breaking changes:

- podman network ls -f shortform is used for --filter instead --format
This matches docker and other podman commands (container ps, volume ps)

- libpod network list endpoint filter parameter is removed. Instead the
filters paramter should be used as json encoded map[string][]string.

### DIFF
--- a/cmd/podman/networks/create.go
+++ b/cmd/podman/networks/create.go
@@ -6,9 +6,11 @@ import (
 
 	"github.com/containers/common/pkg/completion"
 	"github.com/containers/podman/v2/cmd/podman/common"
+	"github.com/containers/podman/v2/cmd/podman/parse"
 	"github.com/containers/podman/v2/cmd/podman/registry"
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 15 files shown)


Luap99:
@baude @mheon  PTAL

Luap99:
Fixes #5376

rhatdan:
The code LGTM, I will let @mheon @baude Comment on the design of how you are storing the labels.
Also @mccv1r0 PTAL

zhangguanzhang:
LGTM

mccv1r0:
LGTM

mheon:
LGTM on my end. We'll merge once 2.2 final is release (probably in a few hours)

PR #7607, opened 2020-09-11
Implement SD-NOTIFY proxy in conmon

goochjj:
Requires https://github.com/containers/conmon/pull/182

Sets up bind mounts for a sd-notify socket, and passes it to the env of the OCI runtime.  NOTIFY_SOCKET is removed from the environment so the OCI runtime does not process the socket, conmon proxies it instead.

Discussion #6688  #7316

PR #8508, opened 2020-11-29
Implement SD-NOTIFY proxy in conmon

rhatdan:
This leverages conmon's ability to proxy the SD-NOTIFY socket.
This prevents locking caused by OCI runtime blocking, waiting for
SD-NOTIFY messages, and instead passes the messages directly up
to the host.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -422,6 +422,7 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 			newMount.Options = append(newMount.Options, "ro", "nosuid", "noexec", "nodev")
 		}
 		if !MountExists(g.Mounts(), dstPath) {
+			logrus.Warnf("Mounting %q to %q", srcPath, dstPath)
 			g.AddMount(newMount)
 		} else {
 			logrus.Infof("User mount overriding libpod mount at %q", dstPath)
@@ -691,6 +692,11 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 	}
 	if !foundContainerEnv {
 		g.AddProcessEnv("container", "libpod")
+	}
+
+	// Set notify
+	if c.config.SdNotifyMode == define.SdNotifyModeContainer {
+		g.AddProcessEnv("NOTIFY_SOCKET", "/run/notify/notify.sock")
 	}
 
 	cgroupPath, err := c.getOCICgroupPath()
@@ -1624,6 +1630,25 @@ rootless=%d
 			c.state.BindMounts[dest] = src
 		}
 	}
+
+	// Bind notify proxy
+	if c.config.SdNotifyMode == define.SdNotifyModeContainer {
+		notifyDir := filepath.Join(c.bundlePath(), "notify")
+		logrus.Debugf("checking notify %q dir", notifyDir)
+		if err := os.MkdirAll(notifyDir, 0755); err != nil {
+			if !os.IsExist(err) {
+				return errors.Wrapf(err, "unable to create notify %q dir", notifyDir)
+			}
+		}
+		if err := label.Relabel(notifyDir, c.MountLabel(), true); err != nil {
+			return errors.Wrapf(err, "relabel failed %q", notifyDir)
+		}
+		logrus.Debugf("add bindmount notify %q dir", notifyDir)
+		if _, ok := c.state.BindMounts["/run/notify"]; !ok {
+			c.state.BindMounts["/run/notify"] = notifyDir
+		}
+	}
+
 	return nil
 }
 
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -364,11 +364,6 @@ func (r *ConmonOCIRuntime) StartContainer(ctr *Container) error {
 		return err
 	}
 	env := []string{fmt.Sprintf("XDG_RUNTIME_DIR=%s", runtimeDir)}
-	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
-		if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
-			env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
-		}
-	}
 	if path, ok := os.LookupEnv("PATH"); ok {
 		env = append(env, fmt.Sprintf("PATH=%s", path))
 	}
@@ -1215,11 +1210,6 @@ func (r *ConmonOCIRuntime) configureConmonEnv(ctr *Container, runtimeDir string)
 	}
 
 	extraFiles := make([]*os.File, 0)
-	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
-		if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
-			env = append(env, fmt.Sprintf("NOTIFY_SOCKET=%s", notify))
-		}
-	}
 	if !r.sdNotify {
 		if listenfds, ok := os.LookupEnv("LISTEN_FDS"); ok {
 			env = append(env, fmt.Sprintf("LISTEN_FDS=%s", listenfds), "LISTEN_PID=1")
@@ -1252,6 +1242,12 @@ func (r *ConmonOCIRuntime) sharedConmonArgs(ctr *Container, cuuid, bundlePath, p
 			rFlags = append(rFlags, "--runtime-arg", arg)
 		}
 		args = append(args, rFlags...)
+	}
+
+	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
+		if notify, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
+			args = append(args, fmt.Sprintf("--sdnotify-socket=%s", notify))
+		}
 	}
 
 	if ctr.CgroupManager() == config.SystemdCgroupsManager && !ctr.config.NoCgroups && ctr.config.CgroupsMode != cgroupSplit {
--- a/libpod/diff.go
+++ b/libpod/diff.go
@@ -14,6 +14,7 @@ var initInodes = map[string]bool{
 	"/etc/resolv.conf":   true,
 	"/proc":              true,
 	"/run":               true,
+	"/run/notify":        true,
 	"/run/.containerenv": true,
 	"/run/secrets":       true,
 	"/sys":               true,


rhatdan:
Replaces: https://github.com/containers/podman/pull/7607

rhatdan:
@goochjj I will take this one over and try to finish it up.

Thanks for the PR.

rhatdan:
@haircommander Is this problem caused by older conmon still?

haircommander:
yeah it looks like it

rhatdan:
@cevich PTAL

libpod/container_internal_linux.go
```diff
@@ -574,6 +575,11 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 		g.AddProcessEnv("container", "libpod")
 	}
 
+	// Set notify
+	if c.config.SdNotifyMode == define.SdNotifyModeContainer {
+		g.AddProcessEnv("NOTIFY_SOCKET", "/run/notify/notify.sock")
```
giuseppe, line 580: it is managed by the OCI runtime (both runc and crun do it).  `define.SdNotifyModeContainer` should simply propagate the `NOTIFY_SOCKET` env variable to the conmon process

libpod/container_internal_linux.go
```diff
@@ -1442,6 +1448,24 @@ func (c *Container) makeBindMounts() error {
...
+		}
+		if err := label.Relabel(notifyDir, c.MountLabel(), true); err != nil {
+			return errors.Wrapf(err, "relabel failed %q", notifyDir)
+		}
+		logrus.Debugf("add bindmount notify %q dir", notifyDir)
+		if _, ok := c.state.BindMounts["/run/notify"]; !ok {
+			c.state.BindMounts["/run/notify"] = notifyDir
+		}
+	}
+
```
giuseppe, line 1468: similarly here, we need the OCI runtime to take care of it.  If we pass down the notify socket directly to the container, there is a race condition between the notification being sent and systemd reading it.

libpod/container_internal_linux.go
rhatdan: So, really this patch is not needed at all? Other then maybe adding /run/notify to the ignore list for diff?

libpod/container_internal_linux.go
goochjj: @giuseppe We're proxying. PODMAN is now in charge of notification.  NOT the OCI runtime. That's the point.

goochjj: PODMAN creates /run/notify/
PODMAN tells OCI runtime to bind-mount /run/notify to the container's setup folder
CONMON forwards /run/notify/notify.sock ( in container, by way of bind mount ) to --sdnotify-socket on host

OCI runtime is NOT taking care of it.

libpod/container_internal_linux.go
goochjj: Again, incorrect, see below.

Container needs this env variable because OCI runtime will NOT have a NOTIFY_SOCKET env var.  We're circumventing OCI container notify socket support. (On purpose)

mheon: The intent here is to work around the frankly broken behavior of OCI runtimes, which by standard do not exit until successfully notifying. This causes the OCI runtime to, during container start, hang around for potentially 10s of seconds - during which the container is locked by Podman to ensure that no other operations are triggered, and thus is unusable (will cause `podman ps` to hang, etc). We can't easily fix OCI runtimes, but we *can* work around them entirely.

giuseppe: I see.  OK it makes sense.

giuseppe (approved):
LGTM

cevich:
> @cevich PTAL

Not sure what is needed from me.  I peeked at the testing results and they all appear related to the changes here, no?

mheon:
We may need new VM images with a newer Conmon.

rhatdan:
Yes @cevich new vms need latest conmon.

cevich:
> Yes @cevich new vms need latest conmon.

And newer Ubuntu :frowning_face:   Can be part of WIP #8312 (from https://github.com/containers/automation_images/pull/34)?

cevich:
Note, "No" is acceptable: in that case I'll roll up new images separate from my Ubuntu work, and they can be brought in via this PR.

rhatdan:
@cevich depends on when I can get new Ubuntu.  A couple of days, I am fine with it. weeks, I have a problem.

cevich:
Sounds fair, if we get into next week and I'm still stuck, I'll make a new PR to just refresh the current images.  There is a risk it will fail with Ubuntu 19, as we're on borrowed-time WRT the package repositories (19 is EOL) remaining available.

Otherwise/also, I've just finished building fresh and am trying those out right now in #8538...

cevich:
...yeah, we've got problems with the new VMs that will probably take a while to solve.  But maybe we have some luck here, b/c IIRC brent built new VMs to incorporate docker compose.  It's not yet in master, @rhatdan try this see if it works:

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index 0fa51be63..be4c0e459 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ env:
     PRIOR_UBUNTU_NAME: "ubuntu-19"

     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c5402398833246208"
+    IMAGE_SUFFIX: "c4704091098054656"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
```

cevich:
Oh! I can check what version of conmon is in ubuntu-c4704091098054656...looks like `2.0.20~1`.  Is that new enough?

rhatdan:
I think we need 2.0.21

cevich:
> I think we need 2.0.21

okay.  Lokesh has a new OBS repo with libseccomp 2.5.1 and other updated packages for us.  Building images now...

cevich:
...testing new images.

cevich:
***Update***: I'm down to ~2-5 failing tests on the new images.  These failures seem to be caused by "too old" package versions.  We're asking a lot of Lokesh to maintain all these special Ubuntu repos. in the first place, so I'm afraid to ask for updates to the (now defunct) 19.10 repo, in addition to 20.04 and 20.10.  At the same time, I don't want to give you a steaming pile to work with, so I'm doing by best to get all the tests passing (PR #8312) first.

cevich:
***Update:***Getting very close with the new images, though I've had to add skips to workaround a few broken tests/situations.

cevich:
***Update***: The suite of podman CI tests should be passing now, so this PR is waiting on review and merge of #8312

cevich:
@rhatdan please rebase this and you should be good.  ***Reminder***: F32 & U20.04 are CGv1(runc),  F33 & U20.10 are CGv2(crun).

rhatdan:
@haircommander @goochjj This should be working according to @cevich but tests are still failing?

edsantiago:
```
conmon: option parsing failed: Unknown option --sdnotify-socket=/tmp/podman_test678427295/notify
```

rhatdan:
Which means we still don't have an updated conmon.

cevich:
> Which means we still don't have an updated conmon.

F33 VM Image has `conmon-2.0.21-3.fc33-x86_64`
F32 VM Image has `conmon-2.0.21-3.fc33-x86_64`
(These are found by hitting the `View more details on Cirrus CI` link, and looking under the "package versions" drop-down).

You previously commented

> I think we need 2.0.21

So dang :disappointed: I'm assuming that is what @lsm5 was informed of and delivered on, it's unfortunate if even that's too old.  I dunno what can be done before the holiday shutdown...Unless someone else can bump the packages, I'm happy to do whatever I can to rebuild images *or* (it's not pretty) but sticking an in-line `dnf update -y conmon` into `setup_environment.sh` is an available short-term option.

rhatdan:
@haircommander WDYT?

haircommander:
> @haircommander WDYT?

oh my goodness, I am really sorry everyone. I really thought the sdnotify support was added in 2.0.21, but it actually hasn't been released yet :upside_down_face: Totally my bad, I should have inspected more thoroughly

haircommander:
included here: 
https://github.com/containers/conmon/pull/220

I should be able to bump the packages in fedora after I tag the release

rhatdan:
@haircommander I merged your change.

haircommander:
they're building now!

haircommander:
I've submitted the builds to testing! @cevich is there anything else you'd need on the packaging side to get the new conmon into the VMs?

cevich:
> I've submitted the builds to testing! @cevich is there anything else you'd need on the packaging side to get the new conmon into the VMs?

Actually now that https://github.com/containers/automation_images/pull/34 has merged, all the hard work is done.  In fact, I have another PR over there that needs to rebuild images before it merges anyway, I can just provide the suffix from that.

***N/B:*** F32 will not pick up packages from updates-testing, only F33 will.  Tests will need to be adjusted accordingly...or some hackish things will be needed to get the new package into F32 at runtime (ick!).

Building now...

cevich:
...Nope, @haircommander it's not picking it up, even with updates-testing enabled I'm only seeing `conmon-2.0.21-3.fc33.x86_64`  :disappointed:

I dunno enough about the packaging side to be any help here, other than building new VM images on request.  Though since it's a pretty low-level dependency, @rhatdan if you want to temporarily stick in a `rpm -Uvh <URL>` somewhere in `setup_environment.sh`, that should at least get you through the holidays :grin: 

Otherwise, I'm happy to try again later today or tomorrow.  Building images just to get updated packages is super-duper easy to do.  The hardest part is waiting for the ~20min process to finish :smile:

cevich:
(the reason I took sooooo long before, was because I was also replacing the EOL Ubuntu 19 image)

rhatdan:
The package has to get to the mirrors.  Should be there by tomorrow.

cevich:
> The package has to get to the mirrors. Should be there by tomorrow.

Oh okay, I'll rebuild the VM images first thing in the morning and confirm...

cevich:
...seeing `conmon-2.0.22-2.fc33.x86_64` in F33 now.  [Building updated images...](https://github.com/containers/automation_images/pull/42)

cevich:
...finished.  @rhatdan try images with suffix `c6135072380157952`.  Though IIUC, only F33 will have the updated conmon.  Unless @haircommander miraculously knows how to update Ubuntu 20.10 in the OBS testing repo.

haircommander:
ubuntu 20.10 has it already actually :smiley: @lsm5 ++ for his automation

cevich:
> ubuntu 20.10 has it already

Nice!  Well then the new image should have picked it up as well.

rhatdan:
@cevich Looks like it is still missing.

rhatdan:
fixes: https://github.com/containers/podman/issues/7316

cevich:
[Running another image build now, in case something else changed over the holiday break.](https://cirrus-ci.com/build/6524344056676352)

cevich:
...done.  Latest images have suffix: `c6524344056676352`

cevich:
Oof! Well...the new images don't seem to be causing any SD-NOTIFY related failures, so that's progress :smile:  Let me know how I can help address the unrelated problems coming from the VM-image update.  They don't seem to be in short-supply these days :disappointed:

cevich:
OIC...the checkpoint failures do seem related:

`Error: error stat'ing file `/tmp/podman_test477832898/crio/vfs-containers/...blah.../userdata/notify`: No such file or directory: OCI not found`

Maybe the sd-notify support could be turned off / disabled for just that one failing test?  Or do we need an updated CRIU with a fix?

cevich:
Okay, same `c6524344056676352` images [deployed in this podman PR.](https://github.com/containers/podman/pull/8899)

rhatdan:
Not going to make the 3.0 release.  Will have to wait until 3.1

mheon:
What's the status here? I see we have a commit changing the default type of `generate systemd` units over to notify in the v3.3 branch - are we still using the OCI runtime to provide sdnotify in that case, or is conmon being used? If we're using the OCI runtime, I think we need to revert for 3.3, until this PR can merge.

rhatdan:
I have no idea of the state.  I grabbed this PR to try to push it over the edge a while back. but I have no idea what is going on at this state.  If @mheon or someone else wants to finish this, I am open to them taking it.

flyingflo:
What is missing besides of a bunch of failing tests? These seem to fail on many pull requests, however. :man_shrugging:  Don't know why..

libpod/oci_conmon_linux.go
```diff
@@ -1254,6 +1244,12 @@ func (r *ConmonOCIRuntime) sharedConmonArgs(ctr *Container, cuuid, bundlePath, p
 		args = append(args, rFlags...)
 	}
 
+	if ctr.config.SdNotifyMode == define.SdNotifyModeContainer {
```
vrothberg, line 1247: Don't we also have to unset NOTIFY_SOCKET to prevent runc and crun from blocking?  Not sure if I understood it correctly though.

rhatdan: Well we don't want to pass these on to the OCI Runtimes, yes, since we don't want them handling it.

vrothberg:
@rhatdan suggested that I take over this PR to kick it over the finish line, closing.

Issue #8507, opened 2020-11-29
podman-2.2.0_rc2 no longer allows later parameters to override those prior


srcshelton:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

In all podman releases up until `2.1.1` (possibly `2.2.0_rc1`) It was possible to provide the same parameter multiple times, and the final instance of the parameter would that which was acted upon.

It was therefore possible to define a podman invocation with a whole series of parameters into a string or array, and then run the container in a different way (perhaps with an override to perform a configuration check) simply by appending further values onto the end of the original variable, e.g.

```bash
dockeropts=(
        #--log-level debug
        --cap-drop ALL
        --cap-add SETGID
        --cap-add SETUID
        --interactive
        --mount type=bind,source=/var/run/syslog-ng/log,destination=/dev/log
        --mount type=bind,source=/etc/${PN}/,destination=/etc/${PN}/,ro=true
        --mount type=bind,source=/var/lib/${PN}/,destination=/var/lib/${PN}
        --mount type=bind,source=/var/log/${PN}/,destination=/var/log/${PN}
        --mount type=bind,source=/var/run/${PN}/,destination=/var/run/${PN}
        --name "openrc-${PN}-${PV}"

        --network host

        --restart on-failure
        #--rm
        --tty
)
dockerimage="service.${CATEGORY}.${PN}:${PV}"

name="containerised ${PN}"
command='podman'
command_args="--log-level=info run ${dockeropts[@]} ${dockerimage} -c ${CONFFILE}"

...

        ebegin "Checking configuration in '${CONFFILE}'"
        podman run "${dockeropts[@]}" \
                        --cap-drop ALL \
                        --network none \
                        --name "openrc-${PN}-${PV}-checkconfig" \
                        --replace \
                        --restart no \
                        --rm \
                "${dockerimage}" -c "${CONFFILE}" -n || rc=${?}
```

... but whereas this worked in all prior releases, in `2.2.0_rc2` podman instead outputs:

```
Error: network conflict between type host and none
```

On a personal note, this change breaks a whole bunch of my scripts - but in general, it's a significant change in behaviour from previous releases.  Whilst it's not impossible to filter strings or arrays to remove duplicates, it's more processing which was not previously required.

If this change is intentional, then could it please at least be relegated to a warning (with the previous behaviour maintained of acting on the last parameter specified) rather than being treated as an error?

**Output of `podman version`:**

```
[...]

PR #8514, opened 2020-11-30
Revert "Allow multiple --network flags for podman run/create"

Luap99:
Reverts containers/podman#8410

As described in issue #8507 this commit contains a breaking change which is not wanted in v2.2.

We can discuss later if we want this in 3.0 or not.


mheon:
LGTM once things go green. We need this in for 2.2 final.

Issue #8528, opened 2020-12-01
vendoring non-upstream cobra


vrothberg:
We are currently vendoring a non-upstream version of cobra (see https://github.com/containers/podman/blob/master/go.mod#L79):
```
[...]
replace github.com/spf13/cobra => github.com/Luap99/cobra v1.0.1-0.20201110155035-83a59186c706
```

That's not sustainable and is breaking Debian packaging. I'd loved to get that fixed quickly.

@Luap99, could you take care of it?  I think you're the cobra expert and will know what to do :)

@mheon @baude FYI ... may be worth a 2.2.1 :^)

Luap99:
@vrothberg Why is it exactly failing? 

The problem is that upstream cobra is missing a feature and bug fixes that I rely on. Commit df4bf5c584f264bdefef5cb30d85c036260eff8f mentions the necessary PRs. Unfortunately there is no indication when these are merged upstream.

vrothberg:
@Luap99, many distributions are not using the `./vendor` directory.  Each dependency is put in a dedicated source package, similar to how traditional C programs are built and distributed.

We currently vendor a fork of a non-released version of project we do not control.  That can cause all kinds of problems (e.g., we miss potential security fixes) with the immediate consequence of breaking downstream packaging.  Downstream packagers won't create a source package for your github fork, so they're blocked.

I missed this part when reviewing https://github.com/containers/podman/pull/6442 unfortunately.

Lets check what we can do to resolve the issue. Do you see a workaround that would avoid the dependency? How much work would it be?

Luap99:
Well that only leaves the option to revert to the latest cobra release. It's doable but will hurt completion experience.

vrothberg:
> Well that only leaves the option to revert to the latest cobra release. It's doable but will hurt completion experience.

Can you elaborate on which functionality would be missing?

Luap99:
It's not easy to explain for zsh and fish we are fine just small bug fixes that are missing. Bash however would have bigger problems, it would recommend every flag with and without the equal sign. Also persistent flags will be shown on subcommands which is not wrong but it will add even more flag suggestions. And it

I guess the best trade off is to add the old bash script back and keep the things for zsh and fish.

Luap99:
I will open a PR

vrothberg:
@Luap99, would the reduced bash functionality be superior to the old script? How long would it take to get your patches upstream?

I would _love_ to keep the great completions you added. Just checking which options we have.

[...]

PR #8534, opened 2020-12-01
Revert the custom cobra vendor

Luap99:
Vendor in the latest cobra release v1.1.1

This will hurt the completion experience but is required for
proper packaging, see: #8528.

The best solution is to keep the current scripts since they
work fine with cobra v1.1.1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -475,15 +475,6 @@ changelog: ## Generate changelog
 	$(shell cat $(TMPFILE) >> changelog.txt)
 	$(shell rm $(TMPFILE))
 
-completions: binaries
-	install ${SELINUXOPT} -d -m 755 completions/{bash,zsh,fish}
-	./bin/podman completion bash --no-desc -f completions/bash/podman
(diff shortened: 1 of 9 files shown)


vrothberg:
I suggest to wait before merging.  Reverting would be sad.  Your work is fantastic and I am optimistic we can keep it.

I dropped a comment in https://github.com/spf13/cobra/pull/1258 to pull in Eric who can help us speed things up upstream.  Vendoring a non-release commit is already a huge improvement over vendoring a fork.

vrothberg:
I like this approach.  We keep the great bash completions but use the upstream cobra!

LGTM

Luap99:
The best thing is to keep the current scripts, they work fine with cobra v1.1.1

One problem with that is that `podman completion` will not generate the best possible scripts. However this is fine since we store the best scripts in `./completions/` and these are used for packaging. I don't think many users want to use this command anyway.
And we would also add one small bug: https://github.com/spf13/cobra/pull/1258

umohnani8:
LGTM
@baude @rhatdan @mheon PTAL

Issue #8057, opened 2020-10-19
Cannot read container mount files


Zeranoe:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I'm trying to use postgres in a podman container with a custom mounted data directory. I'm able to create the container and the database works, but I lose the ability to read any files in the database data folder from the launching user account.

**Steps to reproduce the issue:**

1. `mkdir data`

2. `podman run --publish 5432 -e POSTGRES_PASSWORD=password -v ./data:/var/lib/postgresql/data:Z -it --rm postgres:13`

**Describe the results you received:**

Running `ls data` gives a permissions denied, but running `sudo ls -l data` shows that all the files inside were created with 100998:100998, and have `drwx` permissions.

**Describe the results you expected:**

With postgres running the processes as non-root, this is the result I expect.

**Additional information you deem important (e.g. issue happens only occasionally):**

So far I've tried setting `--user root` and other combinations to attempt to get the container to run the postgres process as my user ID, but nothing seems to work and I'm at a loss as to how this is supposed to operate.

The following seem to be related to this:
* https://www.redhat.com/sysadmin/rootless-podman-makes-sense
* https://github.com/docker-library/postgres/pull/253

The article seems very close, but in reverse (unable to write inside the image).

**Output of `podman version`:**

`podman version 2.1.1`

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-2.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 81d18b6c3ffc266abdef7ca94c1450e669a6a388'
  cpus: 8
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: journald
  hostname: f32
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.15-201.fc32.x86_64
  linkmode: dynamic
  memFree: 6770266112
  memTotal: 16662794240
  ociRuntime:
    name: crun
    package: crun-0.15-5.fc32.x86_64
    path: /usr/bin/crun
    version: |-
[...]

PR #8535, opened 2020-12-01
BATS: add ping test, ps filters, multi-option

edsantiago:
- run test : tweaks to recently-added network-conflict test:
   * remove "-d" in run
   * confirm exact warning text, and also that container
     runs successfully
   * test multiple --net options (regression #8057)

 - images, run, build, exec tests: add multiple-flag
   testing for various flags, confirming as appropriate
   whether options are overridden or accumulated.

 - ps test : add --filter and --sort tests

 - pod test: run 'ping' inside container (confirms that
   container gets PING capability)

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -449,7 +449,9 @@ json-file | f
     msg=$(random_string 20)
     pidfile="${PODMAN_TMPDIR}/$(random_string 20)"
 
-    run_podman run --name myctr --log-driver journald --conmon-pidfile $pidfile $IMAGE echo $msg
+    # Multiple --log-driver options to confirm that last one wins
+    run_podman run --name myctr --log-driver=none --log-driver journald \
+               --conmon-pidfile $pidfile $IMAGE echo $msg
 
     journalctl --output cat  _PID=$(cat $pidfile)
     is "$output" "$msg" "check that journalctl output equals the container output"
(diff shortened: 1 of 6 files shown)


TomSweeneyRedHat:
LGTM
and happy green test buttons.

test/system/030-run.bats
```diff
@@ -533,8 +537,15 @@ json-file | f
 }
 
 @test "podman run with --net=host and --port prints warning" {
-    run_podman run -d --rm -p 8080 --net=host $IMAGE ls > /dev/null
```
edsantiago, line 536: Quick followup on this: it turns out, `>/dev/null` is _not_ a NOP, it's worse: it causes all output from `run_podman` to vanish, which is anti-helpful when this test flakes, as it did twice in #8536! 

[remote f33](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6648492636504064/html/sys-remote-fedora-33-root-host.log.html#t--00037)

[remote ubuntu-20](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5733698962194432/html/sys-remote-ubuntu-20-root-host.log.html#t--00037)

mheon: The intent is to remove all non-stderr output. We only want to capture the output of STDERR for this test. I believe you told me this was the best way to accomplish that?

Regarding the flake, I'll bet is the `--rm` conflicting with the test's own cleanup logic.

edsantiago: Oh, that makes sense (the rm conflict). 

Re: BATS and stdout/stderr, unfortunately, there's no way to split them within a BATS `run` context (including `run_podman`). It's a BATS design tradeoff that works really well for the 95% case. The way I rewrote the test in this PR, I check for both the warning and some stdout; but do not (and cannot) check that the warning truly goes to stderr.

PR #8540, opened 2020-12-01
compat create should use bindings

baude:
the volumes provided is seemingly useless representing what volumes
should be added to a container. instead, the host config bindings should
be used as they acurately describe the src/dest and options for
bindings.


mheon:
LGTM

umohnani8:
LGTM

PR #8541, opened 2020-12-01
Fix potential race condition in testing

rhatdan:
The It("podman wait to pause|unpause condition"... test is
flaking every so often when a messages is sent in the second
function to a channel.  It is my believe that in between the time
the first function sends a message to the channel and before it closes
the channel the second errChan=make() has happened.  This would mean that
the fist function closes the second errChan, and then when the second
function sends a message to the second errChan, it fails and blows up with
the error you are seeing.

By creating a different variable for the second channel, we eliminate the race.

Fixes: https://github.com/containers/podman/issues/6518

### DIFF
--- a/pkg/bindings/test/containers_test.go
+++ b/pkg/bindings/test/containers_test.go
@@ -290,17 +290,17 @@ var _ = Describe("Podman containers ", func() {
 		Expect(wait).To(BeNil())
 		Expect(exitCode).To(BeNumerically("==", -1))
 
-		errChan = make(chan error)
+		unpauseErrChan := make(chan error)
 		go func() {
 			defer GinkgoRecover()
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago PTAL

mheon:
LGTM

edsantiago:
STWATTM(*)

 * Seems Totally Worth A Try To Me

TomSweeneyRedHat:
LGTM

rhatdan:
@debarshiray What do you think if we only give this information out in --privileged mode?  Or have a config flag and containers.conf to say whether we reveal it.

Some engineers are raising concerns about the information leak, being a vector to attach the system.

edsantiago:
@rhatdan I think this was meant for #6192 ?

PR #8543, opened 2020-12-01
Do not use "true" after "syslog" in exit commands

mheon:
Instead of being interpreted as an argument to the boolean flag, the 'true' is being intepreted as the Podman command to be run - so we're trying to run `podman true`, which does not exist. This causes the cleanup command to fail when `--log-level=debug` is set, so containers are not cleaned up or removed.

Separate, but worth looking into later: the errors we get on trying `podman true` with any flags are terrible - if you just type `podman true` you get a quite sane "Unrecognized command" error, but if you try `podman true --rm` you get an "unknown flag --rm" error - which makes very little sense given the command itself doesn't exist.

### DIFF
--- a/pkg/specgen/generate/container_create.go
+++ b/pkg/specgen/generate/container_create.go
@@ -388,7 +388,7 @@ func CreateExitCommandArgs(storageConfig storage.StoreOptions, config *config.Co
 	}
 
 	if syslog {
-		command = append(command, "--syslog", "true")
+		command = append(command, "--syslog")
 	}
 	command = append(command, []string{"container", "cleanup"}...)
 


mheon:
@edsantiago PTAL, I finally tracked down the `podman run --log-level=debug -d --rm` errors.

jwhonce:
LGTM

edsantiago:
Nice catch!

Was there ever an issue filed about this? I can't seem to find one. And, what would a test for this look like?

mheon:
I can't think of a test that isn't horribly racey - with `--rm` you can normally rely on the container being gone after the command exits, but that's out the window with `-d`. We could do a 5 second sleep and then check if it still exists, but that might not be enough on slow CI systems.

For an issue, I looked but could not find one.

edsantiago:
Sounds about right. But for posterity, would you mind editing the description and adding the reproducer? It would be nice to have for future reference, but I no longer remember that far back.

mheon:
Done

mheon:
(I edited the commit message not the description)

rhatdan:
The podman true versus podman true --rm, seems likely to be a problem in cobra?

jwhonce:
> The podman true versus podman true --rm, seems likely to be a problem in cobra?

Whether the flag is anchored to a command vs. global also comes into play here.  I suspect --syslog=true would have also resolved the issue.  This also came up during the --remote support, which we anchored to the podman command vs. global.

mheon:
`--syslog=true` would probably work, but these flags go through multiple levels of parsing (they get passed through Conmon first via `--exit-command-arg`) so I try to avoid making them any more complex than they have to be.

Issue #7865, opened 2020-10-01
podman network ls format doesn't work


tmds:
`podman network ls --format` doesn't work:

```
$ podman network ls  --format {{.ID}}
Error: template: listNetworks:1:14: executing "listNetworks" at <.ID>: can't evaluate field ID in type network.ListPrintReports
```

version:
```
podman.x86_64                                                         2:2.1.0-2.fc32                                                          @updates-testing
```

dzerhusen:
Where do you got the `{{.ID}}` from? I must say, i am only using version 2.0.4. 

`podman network ls --format {{.Name}}`
works for example.

tmds:
> podman network ls --format {{.Name}}

That works for me too. I got the `{{.ID}}` by copy-paste.

Do you know how I can see the full json?

zhangguanzhang:
```
./bin/podman network ls --format json
```
there is no filed named `ID`

mheon:
I believe this is a difference between us and Docker for now. We have plans
for a partial rewrite of the 'podman network's commands to address this but
it could take a while before we get to it.

tmds:
I had tried, `--format {{json .}}`. I'm not sure if that is supposed to work.

@mheon should I close this issue? or do you want to keep it open?

mheon:
Hm. I'll tag it as a feature request and it can stay open.

rhatdan:
Hopefully someone from the community can take this on.

eiffel-fl:
I can take a look to this issue.
My first research leads me to `type NetworkConfigList` in `libcni`.

I will finish to setup my development environment for `podman` and I will investigate more.

rhatdan:
You got it.  Ping us if you need any help.

eiffel-fl:
Maybe I underestimate the problem.
At first sight I just though the ID was not printed, so adding `ID string` to `type NetworkConfigList` should have solve the problem.
It solves it partly, indeed there is no more error.
But the ID is empty and I have the feeling that networks do not have ids in podman.
Even using the REST API through `curl` does not print the ID ([see](https://framabin.org/p/?1f5e399bc3e69914#qqNwbzQHLgdyK2pw2LWyCWhqhSZgd7GQQyTQ72kBABk=)).

I can try to add ID to networks but the modifications will be deeper and take more time.
If you have advices so my modifications don't break anything and be as restrained as possible I will take them.

mheon:
[...]

PR #8549, opened 2020-12-02
Add support for network ids

Luap99:
The network ID is not stored. It is just the sha256 hash from
the network name. There is a risk of a potential hash collision.
However it's very unlikely and even if we hit this it will
complain that more than network with this ID exists.

The main benefit is that the compat api can have proper
network ID support. Also this adds the support for
`podman network ls --format "{{.ID}}"` and `--filter id=<ID>`.

It also ensures that we can do network rm \<ID> and network
inspect \<ID>.

Since we use a hash this commit is backwards compatible even for
already existing networks.

### DIFF
--- a/cmd/podman/networks/list.go
+++ b/cmd/podman/networks/list.go
@@ -37,6 +37,7 @@ var (
 var (
 	networkListOptions entities.NetworkListOptions
 	filters            []string
+	noTrunc            bool
 )
 
 func networkListFlags(flags *pflag.FlagSet) {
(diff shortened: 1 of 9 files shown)


Luap99:
Fixes #7865
Replaces #7962

Luap99:
@baude @mheon PTAL

rhatdan:
LGTM
Although I will wait for @mheon @baude response.

mheon:
LGTM. It's not really perfect (the collision issue has already been noted in the PR) but it is good enough, and anything to bring us closer to parity on the networking side is a good thing.

baude:
i love it

PR #8550, opened 2020-12-02
Fix shell completion for ps --filter ancestor

Luap99:
The `ancestor` option was missing an equal sign. Therefore
the completion did not work as expected.


zhangguanzhang:
LGTM

Issue #8544, opened 2020-12-01
APIv2: running containers fails when "NetworkMode" is "default" (which is the default value docker-py uses)


riyad:
/kind bug

**Description**

Trying to find regressions from Podman 2.1.1 by exercising the APIv2 trough [docker-py](https://github.com/docker/docker-py)'s test suite (see #5386) I came across a regressions in the [`CreateContainerTest.test_group_id_ints`](https://github.com/docker/docker-py/blob/master/tests/integration/api_container_test.py#L221) test.

A container is incidentally created with `"NetworkMode": "default"`. Trying to start it later fails with _"no net configurations found in /etc/cni/net.d"_.

`"NetworkMode"` is set to `"default"` as a [default/fallback value by docker-py](https://github.com/docker/docker-py/blob/432819766014ce396e7507c52246ff5c8a11cdbc/docker/types/containers.py#L341) and has been  for [at least 4 years](https://github.com/docker/docker-py/blame/27e3fa3e2e2a198ae95399629c2d492368c95965/docker/types/containers.py#L345). Although neither documented in the [CLI](https://docs.docker.com/engine/reference/run/#network-settings) nor in the [API docs](https://docs.docker.com/engine/api/v1.40/#operation/ContainerCreate) (under HostConfig->NetworkMode) `"default"` seems to get [special treatment](https://github.com/moby/moby/blob/e1bba7456d9902d04eec8c402f852d0ae17d123c/api/types/container/host_config.go#L124) within the Docker engine (but I'm not sure how far it goes).

**Steps to reproduce the issue:**

```shell
# inside the podman repo
./bin/podman system service -t 0
```
In another shell
```shell
curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/containers/create' -H 'Content-Type: application/json' -d '{"Tty": false, "OpenStdin": false, "StdinOnce": false, "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Cmd": ["id", "-G"], "Image": "alpine:3.10", "NetworkDisabled": false, "HostConfig": {"NetworkMode": "default", "GroupAdd": ["1000", "1001"]}}'

# put the "Id" value from response in the URL
curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/containers/8d510dd4d4a9cd5a2917c58e4dbc7f177091724e1f88625af5b3cd0f75096485/start'
```

[...]

PR #8551, opened 2020-12-02
Support --network=default as if it was private

rhatdan:
Docker defines an option of "default" which means to
use the default network.  We should support this with
the same code path as --network="".

This is important for compatibility with the Docker API.

Fixes: https://github.com/containers/podman/issues/8544


TomSweeneyRedHat:
@rhatdan new tests need help

rhatdan:
This should fix another issue with "Docker" APIV2 comparability.

saschagrunert (approved):
LGTM

PR #8552, opened 2020-12-02
target is not tag

baude:
remove mistaken use of target being used for tag


mheon:
LGTM

PR #8556, opened 2020-12-02
Use Libpod tmpdir for pause path

### LINKED ISSUE #8539: Podman 2.2.0 is broken on Ubuntu Focal

/kind bug

**Description**

1. Install Podman 2.2.0 using documented method/repo: https://podman.io/getting-started/installation

2. Execute: `podman info`

3.

```
podman -v
podman version 2.2.0
podman info
error creating temporary file: No such file or directory
Error: error setting up the process: open /tmp/podman-run-1001/libpod/pause.pid: no such file or directory
```

Info output...

It works with sudoer accounts...

```
podman version
error creating temporary file: No such file or directory
Error: error setting up the process: open /tmp/podman-run-1001/libpod/pause.pid: no such file or directory
```

```
podman info --debug
error creating temporary file: No such file or directory
Error: error setting up the process: open /tmp/podman-run-1001/libpod/pause.pid: no such file or directory
```

`podman/unknown,now 2.2.0~1 amd64 [installed]`

Yes

physical

mheon:
Previously, we always computed pause path from the Rootless runtime directory. Problem: this does not match the behavior of Libpod when the directory changes. Libpod will continue to use the previous directory, cached in the database; Pause pidfiles will swap to the new path. This is problematic when the directory needs to exist to write the pidfile, and Libpod is what creates the directory.

There are two potential solutions - allow the pause pidfile to move and just make the directory when we want to write it, or use the cached Libpod paths for a guaranteed location. This patch does the second, because it seems safer - we will never miss a previously-existing pidfile because the location is now consistent.

Fixes #8539

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -472,7 +472,7 @@ func makeRuntime(ctx context.Context, runtime *Runtime) (retErr error) {
 		// we will need to access the storage.
 		if os.Geteuid() != 0 {
 			aliveLock.Unlock() // Unlock to avoid deadlock as BecomeRootInUserNS will reexec.
-			pausePid, err := util.GetRootlessPauseProcessPidPath()
+			pausePid, err := util.GetRootlessPauseProcessPidPathGivenDir(runtime.config.Engine.TmpDir)
 			if err != nil {
 				return errors.Wrapf(err, "could not get pause process pid file path")
 			}
@@ -536,6 +536,15 @@ func makeRuntime(ctx context.Context, runtime *Runtime) (retErr error) {
 	}
 
 	return nil
+}
+
+// TmpDir gets the current Libpod temporary files directory.
+func (r *Runtime) TmpDir() (string, error) {
+	if !r.valid {
+		return "", define.ErrRuntimeStopped
+	}
+
+	return r.config.Engine.TmpDir, nil
 }
 
 // GetConfig returns a copy of the configuration used by the runtime
--- a/pkg/util/utils_supported.go
+++ b/pkg/util/utils_supported.go
@@ -99,11 +99,22 @@ func GetRootlessConfigHomeDir() (string, error) {
 }
 
 // GetRootlessPauseProcessPidPath returns the path to the file that holds the pid for
-// the pause process
+// the pause process.
+// DEPRECATED - switch to GetRootlessPauseProcessPidPathGivenDir
 func GetRootlessPauseProcessPidPath() (string, error) {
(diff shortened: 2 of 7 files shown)


mheon:
@giuseppe @rhatdan PTAL

rhatdan:
LGTM
But this is a breaking change and requires 3.0

mheon:
This is a critical bugfix - 2.2 broke Podman for anyone who does not have $XDG_RUNTIME_DIR defined.

mheon:
Re-pushed again to fix cross-compile

libpod/runtime.go
```diff
@@ -472,7 +472,7 @@ func makeRuntime(ctx context.Context, runtime *Runtime) (retErr error) {
 		// we will need to access the storage.
 		if os.Geteuid() != 0 {
 			aliveLock.Unlock() // Unlock to avoid deadlock as BecomeRootInUserNS will reexec.
-			pausePid, err := util.GetRootlessPauseProcessPidPath()
+			pausePid, err := util.GetRootlessPauseProcessPidPath(runtime.config.Engine.TmpDir)
```
rhatdan, line 475: Add new new function and depracate old function, then we don't break ABI
util.RootlessPausePidPath(runtime.config.Engine.TmpDir)

mheon:
@rhatdan Re-pushed with fixes.

pkg/util/utils_supported.go
```diff
@@ -99,11 +99,22 @@ func GetRootlessConfigHomeDir() (string, error) {
 }
 
 // GetRootlessPauseProcessPidPath returns the path to the file that holds the pid for
-// the pause process
+// the pause process.
+// DEPRECATED - switch to GetRootlessPauseProcessPidPathGivenDir
 func GetRootlessPauseProcessPidPath() (string, error) {
 	runtimeDir, err := GetRuntimeDir()
```
rhatdan, line 105: NIT: Should call the new interface from the old one, for completeness.  Fix if forced to repush.
LGTM

mheon: They're actually different on the `filepath.Join()` bits - the old one manually appends `libpod` to try and find the Libpod tmpdir.

rhatdan: I was thinking of:

GetRootlessPauseProcessPidPathGivenDir(filepath.Join(runtimeDir, "libpod"))

TomSweeneyRedHat:
LGTM

QiWang19 (commented):
LGTM

PR #8557, opened 2020-12-02
add commas between mount options

baude:
when formatting mount options into a string for the compat container create, the options need to be comma delimited.

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -204,10 +204,10 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	for _, m := range cc.HostConfig.Mounts {
 		mount := fmt.Sprintf("type=%s", m.Type)
 		if len(m.Source) > 0 {
-			mount += fmt.Sprintf("source=%s", m.Source)
+			mount += fmt.Sprintf(",source=%s", m.Source)
 		}
 		if len(m.Target) > 0 {
-			mount += fmt.Sprintf("dest=%s", m.Target)
+			mount += fmt.Sprintf(",dst=%s", m.Target)
 		}
 		mounts = append(mounts, mount)
 	}


cmd/podman/common/create_opts.go
```diff
@@ -202,14 +202,14 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	// TODO options
 	mounts := make([]string, 0, len(cc.HostConfig.Mounts))
 	for _, m := range cc.HostConfig.Mounts {
-		mount := fmt.Sprintf("type=%s", m.Type)
+		mount := fmt.Sprintf("type=%s,", m.Type)
 		if len(m.Source) > 0 {
-			mount += fmt.Sprintf("source=%s", m.Source)
+			mount += fmt.Sprintf("source=%s,", m.Source)
```
edsantiago, line 207: Wouldn't it be simpler to prepend the comma here, and remove the `TrimSuffix()` below?
```go
    mount += fmt.Sprintf(",source=%s", m.Source)
                          ^
```

mheon:
LGTM

cmd/podman/common/create_opts.go
```diff
@@ -204,10 +204,10 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	for _, m := range cc.HostConfig.Mounts {
 		mount := fmt.Sprintf("type=%s", m.Type)
 		if len(m.Source) > 0 {
-			mount += fmt.Sprintf("source=%s", m.Source)
+			mount += fmt.Sprintf(",source=%s", m.Source)
 		}
 		if len(m.Target) > 0 {
-			mount += fmt.Sprintf("dest=%s", m.Target)
+			mount += fmt.Sprintf(",dst=%s", m.Target)
```
edsantiago, line 210: Is the change from `dest` to `dst` deliberate, and if so does it merit a note in the commit message?

baude: it is delib, but for a one handed guy rn, i choose to not mention it

PR #8561, opened 2020-12-02
Do not mount sysfs as rootless in more cases

mheon:
We can't mount sysfs as rootless unless we manage the network namespace. Problem: slirp4netns is now creating and managing a network namespace separate from the OCI runtime, so we can't mount sysfs in many circumstances. The `crun` OCI runtime will automatically handle this by falling back to a bind mount, but `runc` will not, so we didn't notice until RHEL gating tests ran on the new branch.

### DIFF
--- a/pkg/specgen/generate/oci.go
+++ b/pkg/specgen/generate/oci.go
@@ -165,7 +165,7 @@ func SpecGenToOCI(ctx context.Context, s *specgen.SpecGenerator, rt *libpod.Runt
 			inUserNS = true
 		}
 	}
-	if inUserNS && s.NetNS.IsHost() {
+	if inUserNS && s.NetNS.NSMode != specgen.NoNetwork {
 		canMountSys = false
 	}
 
--- a/test/e2e/run_memory_test.go
+++ b/test/e2e/run_memory_test.go
@@ -38,7 +38,7 @@ var _ = Describe("Podman run memory", func() {
 		var session *PodmanSessionIntegration
 
 		if CGROUPSV2 {
-			session = podmanTest.Podman([]string{"run", "--memory=40m", ALPINE, "sh", "-c", "cat /sys/fs/cgroup/$(sed -e 's|0::||' < /proc/self/cgroup)/memory.max"})
+			session = podmanTest.Podman([]string{"run", "--memory=40m", "--net=none", ALPINE, "sh", "-c", "cat /sys/fs/cgroup/$(sed -e 's|0::||' < /proc/self/cgroup)/memory.max"})
 		} else {
 			session = podmanTest.Podman([]string{"run", "--memory=40m", ALPINE, "cat", "/sys/fs/cgroup/memory/memory.limit_in_bytes"})
 		}
@@ -55,7 +55,7 @@ var _ = Describe("Podman run memory", func() {
 		var session *PodmanSessionIntegration
 
 		if CGROUPSV2 {
-			session = podmanTest.Podman([]string{"run", "--memory-reservation=40m", ALPINE, "sh", "-c", "cat /sys/fs/cgroup/$(sed -e 's|0::||' < /proc/self/cgroup)/memory.low"})
+			session = podmanTest.Podman([]string{"run", "--memory-reservation=40m", "--net=none", ALPINE, "sh", "-c", "cat /sys/fs/cgroup/$(sed -e 's|0::||' < /proc/self/cgroup)/memory.low"})
 		} else {
 			session = podmanTest.Podman([]string{"run", "--memory-reservation=40m", ALPINE, "cat", "/sys/fs/cgroup/memory/memory.soft_limit_in_bytes"})
 		}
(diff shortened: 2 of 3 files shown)


edsantiago:
Just curious: don't the Ubuntu CI VMs use runc? I'm wondering how/if we could have caught this earlier, and how we can test that this will work in RHEL.

mheon:
@giuseppe PTAL

@edsantiago That is an excellent question. I think they do. Possibly a newer version than RHEL runc, that has its own workaround for this?

mheon:
Memory tests are definitely broken, they're looking at things in `/sys/fs/cgroup`...

pkg/specgen/generate/oci.go
```diff
@@ -165,7 +165,7 @@ func SpecGenToOCI(ctx context.Context, s *specgen.SpecGenerator, rt *libpod.Runt
 			inUserNS = true
 		}
 	}
-	if inUserNS && s.NetNS.IsHost() {
+	if inUserNS && s.NetNS.NSMode != specgen.NoNetwork {
```
giuseppe, line 168: doesn't it work when the NSMode is set to private?

mheon: I suspect it doesn't because we still make the namespace in Podman in that case

mheon: I'll check again to verify.

mheon: You are entirely correct, it does work.

mheon:
Re-pushed with fix.

mvollmer:
Thanks for this fix! We ran into this while testing cockpit-podman on our rhel-8-4 image, which should be representative for a generic RHEL 8.4 installation.

And now I am confused about runc vs crun. What should be in use in a bone stock RHEL 8 installation, runc or crun? In other words, would a normal RHEL 8 podman user run into this bug, or not?

I ask because I am worried that we are unknowingly testing some less interesting configuration of RHEL 8.

Thanks!

mheon:
The default is `runc` - `crun` only became available with 8.3.0, AFAIK, and there's no way we're making it default before 9.0

mvollmer:
> The default is `runc` - `crun` only became available with 8.3.0, AFAIK, and there's no way we're making it default before 9.0

Thanks, so we are doing it right!

pkg/specgen/generate/oci.go
```diff
@@ -165,7 +165,7 @@ func SpecGenToOCI(ctx context.Context, s *specgen.SpecGenerator, rt *libpod.Runt
 			inUserNS = true
 		}
 	}
-	if inUserNS && s.NetNS.IsHost() {
+	if inUserNS && !(s.NetNS.NSMode == specgen.NoNetwork || s.NetNS.NSMode == specgen.Private) {
```
rhatdan, line 168: I am not sure this is correct?  And I think we should be using the functions in namespaces?  I am not sure if all cases are handled.

NSMode == "default" 
NSMode == ""
are the same as
NSMode == "private"

mheon: No, they're not. We explicitly do not want to trigger this with slirp4netns networking, and both of those are trigger slirp.

mheon: If `private` also triggers slirp, then we'll need to explicitly disable it as well.

mheon: I'm going to revert back to NoNetwork. It's guaranteed to be safe. We can loosen this later if we need to.

mheon:
Re-pushed with fixes for tests - I disabled networking on a lot of tests that accessed `/sys/fs/cgroup` as cgroupsv2 to bypass the sysfs problem.

pkg/specgen/generate/oci.go
rhatdan: So we need to have a check added to indicate whether or not we are using slirp4netns networking.

mheon: I'm actually not sure if CNI + user namespaces is safe either. It's probably better to leave disabled.

rhatdan:
@giuseppe PTAL

mheon:
@containers/podman-maintainers PTAL, this is needed for 2.2.1

saschagrunert (approved):
LGTM :+1:

giuseppe:
this PR caused a regression as reported here: https://github.com/systemd/systemd/pull/17902#issuecomment-758517847

```
podman run  --rm fedora grep cgroup /proc/self/mountinfo
1893 1891 0:27 /../../../../../.. /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime - cgroup2 cgroup2 rw,seclabel,nsdelegate
```

Do you remember what was the failure we saw without this commit?

giuseppe:
@mvollmer do you have a reproducer for the issue you were seeing?

mheon:
@giuseppe Was causing failures to run rootless Podman on cgroupsv1 + runc (most notably on the RHEL8 test suite, but Ubuntu and other v1 distros were also affected).

giuseppe:
I've opened a PR to see if we catch any of these failures in the CI: https://github.com/containers/podman/pull/8949

jonasbb:
@giuseppe Would #8949 also help with #8712? #8712 has steps to reproduce.

giuseppe:
> @giuseppe Would #8949 also help with #8712? #8712 has steps to reproduce.

yes

jonasbb:
Glad to hear. Then I hope this PR or a similar one can be merged for the next release.

PR #8562, opened 2020-12-02
WIP: test docker-compose

edsantiago:
git-fetch a set of docker-compose examples

Add a mechanism for 'curl'ing to verify that service is up.

For each docker-compose directory:
  1) test that it comes up
  2) use 'curl' to confirm that the service is up
  3) bring it back down


edsantiago:
@baude this isn't working, and I'm at a complete loss to understand why. Basically, the `curl` step is failing. If I put a pause in the script (included in this WIP), I have all the time in the world to run `lsof -i -n -P | grep 3000` and confirm that a podman process is listening on `*:3000`; but no variation of `curl http://127.0.0.1:3000/` will give me any results. I've tried removing the `--cni-config-dir` option, no difference. I'm stepping away from it for the rest of today. Maybe you'll have better luck.

edsantiago:
Progress so far: nil. Absolutely no clue. My findings, in case they're helpful:
* it has to do with `--root`; the `--cni-config-dir` option is irrelevant, as is pre-copying `87-podman*conflist` into the tmp cni dir.
* it only happens with docker-compose; I have been unable to manually create a container, using `sudo ./bin/podman`, that recreates this behavior
* if I remove the `--root` and `--runroot` from the script, it works as expected. The port responds on localhost.

Could this be a weird interaction between `--root` and APIv2?

rhatdan:
@edsantiago @baude Is this still valid, or has another version of this been merged?

edsantiago:
OOPS! Sorry, this was just a launching board for @baude to use in his tests (Which are already merged). Closing.

PR #8563, opened 2020-12-02
podman, exec: move conmon to the correct cgroup

giuseppe:
move the conmon process to the conmon cgroup also on exec.

The previous implementation would fail to move the conmon process as
the systemd unit already exists so its creation would fail.

When the unit cannot be created, attempt to directly join the cgroup
instead.

### DIFF
--- a/utils/utils_supported.go
+++ b/utils/utils_supported.go
@@ -43,6 +43,15 @@ func RunUnderSystemdScope(pid int, slice string, unitName string) error {
 	ch := make(chan string)
 	_, err = conn.StartTransientUnit(unitName, "replace", properties, ch)
 	if err != nil {
+		// On errors check if the cgroup already exists, if it does move the process there
+		if props, err := conn.GetUnitTypeProperties(unitName, "Scope"); err == nil {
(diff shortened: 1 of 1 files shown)


mheon:
LGTM once tests go green

TomSweeneyRedHat:
LGTM
assuming happy tests.

utils/utils_supported.go
```diff
@@ -156,9 +171,21 @@ func MoveUnderCgroupSubtree(subtree string) error {
 		}
 		defer f.Close()
 
-		for _, pid := range bytes.Split(processes, []byte("\n")) {
-			if _, err := f.Write(pid); err != nil {
-				logrus.Warnf("Cannot move process %s to cgroup %q", pid, newCgroup)
+		if len(processes) > 0 {
+			for _, pid := range processes {
+				if _, err := f.Write([]byte(fmt.Sprintf("%d\n", pid))); err != nil {
+					logrus.Warnf("Cannot move process %s to cgroup %q", pid, newCgroup)
```
zhangguanzhang, line 177:
```suggestion
					logrus.Warnf("Cannot move process %d to cgroup %q", pid, newCgroup)
```

PR #8564, opened 2020-12-02
BATS: add new load test

edsantiago:
Looks like #7337 was fixed (by #8112). Reenable a disabled
test for it; and make it actually work. Confirmed that
newly-added test fails on d45676549 (the commit before #8112).


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #8570, opened 2020-12-03
rewrite container copy

vrothberg:
* Add a new `pkg/copy` to centralize all container-copy related code.

* The new code is based on Buildah's `copier` package.

* The compat `/archive` endpoints use the new `copy` package.

* Update docs and add several new tests.

* Includes many fixes, most notably, the look-up of volumes and mounts.

Breaking changes:

 * Podman is now expecting that container-destination paths exist.
   Before, Podman created the paths if needed.  Docker does not do
   that and I believe Podman should not either as it's a recipe for
   masking errors.  These errors may be user induced (e.g., a path
   typo), or internal typos (e.g., when the destination may be a
   mistakenly unmounted volume).  Let's keep the magic low for such
   a security sensitive feature.


vrothberg:
@edsantiago, I would love to have your feedback on the system tests.  I added _a lot_ of them.  Given `podman cp` is very security sensitive, I thought that having a decent coverage in the e2e would is desirable.

cmd/podman/containers/cp.go
```diff
@@ -43,7 +40,7 @@ var (
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Extract the tar file into the destination directory.")
-	flags.BoolVar(&cpOpts.Pause, "pause", copyPause(), "Pause the container while copying")
```
mheon, line 46: We should probably deprecate this - it's an unnecessary addition (we fixed the CVE it was supposed to resolve in a different way).

vrothberg: Sounds good to me.  Maybe something for a follow up PR when working on `remote cp`.  Will spare us from adding the parameters.

I think we can leave it on the CLI but make it a NOP?  Or full removal given 3.0 would allow us to?

rhatdan: Leave it but remove documentation, and hide the option from --help.

test/system/065-cp.bats
```diff
@@ -7,6 +7,312 @@
...
+    run_podman 125 cp $srcdir/IdoNotExist cpcontainer:/tmp
+    is "$output" 'Error: ".*/IdoNotExist" could not be found on the host'
+
+    # Container path does not exist.  Notice that the error message shows how
+    # the specified container is resolved.
+    run_podman 125 cp $srcdir/hostfile cpcontainer:/IdoNotExist/
+    is "$output" 'Error: "/IdoNotExist/" could not be found on container.*(resolved to .*/IdoNotExist.*'
+
+    run_podman rm -f cpcontainer
+}
```
edsantiago, line 71: Any time I see a huge list of commands with only very teeny differences between them, I think: time for a table. Friendly suggestion:
```diff
diff --git a/test/system/065-cp.bats b/test/system/065-cp.bats
index 789267fcf..8e993ecf4 100644
--- a/test/system/065-cp.bats
+++ b/test/system/065-cp.bats
@@ -14,58 +14,54 @@ load helpers
     mkdir -p $srcdir
     # Create multiple files which prevents some (expensive) clean ups via
     # podman-exec.
-    echo "This file is on the host" > $srcdir/hostfile
-    echo "This file is on the host (1)" > $srcdir/hostfile1
-    echo "This file is on the host (2)" > $srcdir/hostfile2
+    local -a randomcontent=(
+        random-0-$(random_string 10)
+        random-1-$(random_string 15)
+        random-2-$(random_string 20)
+    )
+    echo "${randomcontent[0]}" > $srcdir/hostfile0
+    echo "${randomcontent[1]}" > $srcdir/hostfile1
+    echo "${randomcontent[2]}" > $srcdir/hostfile2

     run_podman run -d --name cpcontainer --workdir=/srv $IMAGE sleep infinity
-
-    # Copy to root.
-    run_podman cp $srcdir/hostfile cpcontainer:/
-    run_podman exec cpcontainer cat /hostfile
-    is "$output" "This file is on the host"
-
-    run_podman cp $srcdir/hostfile cpcontainer:/anotherbase.txt
-    run_podman exec cpcontainer cat /anotherbase.txt
-    is "$output" "This file is on the host"
-
-    # Copy to a non-root directory (varied input syntax).
-    run_podman cp $srcdir/hostfile cpcontainer:/tmp
-    run_podman exec cpcontainer cat /tmp/hostfile
-    is "$output" "This file is on the host"
-
-    run_podman cp $srcdir/hostfile1 cpcontainer:/tmp/
-    run_podman exec cpcontainer cat /tmp/hostfile1
-    is "$output" "This file is on the host (1)"
-    run_podman exec cpcontainer rm /tmp/hostfile
-
-    run_podman cp $srcdir/hostfile2 cpcontainer:/tmp/.
-    run_podman exec cpcontainer cat /tmp/hostfile2
-    is "$output" "This file is on the host (2)"
-
-    run_podman cp $srcdir/hostfile cpcontainer:/tmp/anotherbase.txt
-    run_podman exec cpcontainer cat /tmp/anotherbase.txt
-    is "$output" "This file is on the host"
-
-    # Copy to a relative path -> must resolve to the container's work dir
-    # (i.e., /srv).
     run_podman exec cpcontainer mkdir /srv/subdir
-    run_podman cp $srcdir/hostfile cpcontainer:subdir
-    run_podman exec cpcontainer cat /srv/subdir/hostfile
-    is "$output" "This file is on the host"

-    run_podman cp $srcdir/hostfile cpcontainer:anotherbase.txt
-    run_podman exec cpcontainer cat /srv/anotherbase.txt
-    is "$output" "This file is on the host"
+    # format is: <id> | <destination arg to cp> | <full dest path> | <test name>
+    # where:
+    #    id        is 0-2, one of the random strings/files
+    #    dest arg  is the right-hand argument to 'podman cp' (may be implicit)
+    #    dest path is the full explicit path we expect to see
+    #    test name is a short description of what we're testing here
+    tests="
+0 | /                    | /hostfile0            | copy to root
+0 | /anotherbase.txt     | /anotherbase.txt      | copy to root, new name
+0 | /tmp                 | /tmp/hostfile0        | copy to /tmp
+1 | /tmp                 | /tmp/hostfile1        | copy to /tmp
+2 | /tmp                 | /tmp/hostfile2        | copy to /tmp
+0 | /tmp/hostfile2       | /tmp/hostfile2        | overwrite previous copy
+0 | /tmp/anotherbase.txt | /tmp/anotherbase.txt  | copy to /tmp, new name
+0 | anotherbase.txt      | /srv/anotherbase.txt  | copy to workdir, new name
+0 | subdir               | /srv/subdir/hostfile0 | copy to workdir/subdir
+"
+
+    # Copy one of the files into container, exec+cat, confirm the file
+    # is there and matches what we expect
+    while read id dest dest_fullname description; do
+        run_podman cp $srcdir/hostfile$id cpcontainer:$dest
+        run_podman exec cpcontainer cat $dest_fullname
+        is "$output" "${randomcontent[$id]}" "$description (cp -> ctr:$dest)"
+    done < <(parse_table "$tests")

     # Host path does not exist.
     run_podman 125 cp $srcdir/IdoNotExist cpcontainer:/tmp
-    is "$output" 'Error: ".*/IdoNotExist" could not be found on the host'
+    is "$output" 'Error: ".*/IdoNotExist" could not be found on the host' \
+       "copy nonexistent host path"

     # Container path does not exist.  Notice that the error message shows how
     # the specified container is resolved.
-    run_podman 125 cp $srcdir/hostfile cpcontainer:/IdoNotExist/
-    is "$output" 'Error: "/IdoNotExist/" could not be found on container.*(resolved to .*/IdoNotExist.*'
+    run_podman 125 cp $srcdir/hostfile0 cpcontainer:/IdoNotExist/
+    is "$output" 'Error: "/IdoNotExist/" could not be found on container.*(resolved to .*/IdoNotExist.*' \
+       "copy into nonexistent path in container"

     run_podman rm -f cpcontainer
 }
```
[cp-test.txt](https://github.com/containers/podman/files/5637173/cp-test.txt)

test/system/065-cp.bats
```diff
@@ -7,6 +7,312 @@
...
+
+    # Copy to a non-root directory (varied input syntax).
+    run_podman cp $srcdir/hostfile cpcontainer:/tmp
+    run_podman exec cpcontainer cat /tmp/hostfile
+    is "$output" "This file is on the host"
+
+    run_podman cp $srcdir/hostfile1 cpcontainer:/tmp/
+    run_podman exec cpcontainer cat /tmp/hostfile1
+    is "$output" "This file is on the host (1)"
+    run_podman exec cpcontainer rm /tmp/hostfile
```
edsantiago, line 40: My suggested diff does not do this 'rm'. It wasn't clear to this reader what the purpose of that was.

test/system/065-cp.bats
```diff
@@ -7,6 +7,312 @@
 
 load helpers
 
+@test "podman cp file from host to container" {
+    skip_if_remote "podman-remote does not yet handle cp"
+
+    srcdir=$PODMAN_TMPDIR/cp-test-file-host-to-ctr
+    mkdir -p $srcdir
+    # Create multiple files which prevents some (expensive) clean ups via
+    # podman-exec.
```
edsantiago, line 16: Can you elaborate on this? I didn't understand what you mean by cleanup, or why it's expensive. Do you mean that this helps avoid having to `podman exec rm /path` ?

vrothberg: Yes, it was to avoid too many `exec`s. They're expensive and I was worried the tests would consume too much time.

test/e2e/common_test.go
```diff
@@ -317,7 +317,7 @@ func (p *PodmanTestIntegration) createArtifact(image string) {
 	fmt.Printf("Caching %s at %s...", image, destName)
 	if _, err := os.Stat(destName); os.IsNotExist(err) {
 		pull := p.PodmanNoCache([]string{"pull", image})
-		pull.Wait(90)
+		pull.Wait(240)
```
rhatdan, line 320: Was this a debugging change?

vrothberg: No, there's another commit to bump it.

I was hitting this limit too frequently at home (low bandwitdh).  It's used to fetch the images for testing and since they're big, 90 seconds are not always enough :^)

Luap99: +1 I'm facing the same issue when running tests.

test/system/065-cp.bats
vrothberg: Thanks add!

I can apply this diff if you want me to.  My bash skills are limited, so I tend to play it safe :^)

edsantiago: If you agree that it's more maintainable, then yes, please apply it. (You might notice that I added one extra test that was not in the original. IT was a one-line addition instead of multiple error-prone ones).

rhatdan: I like @edsantiago idea, apply the diff.

pkg/copy/copy.go
```diff
@@ -0,0 +1,188 @@
+package copy
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	buildahCopiah "github.com/containers/buildah/copier"
```
TomSweeneyRedHat, line 9: Love it!

pkg/copy/copy.go
```diff
@@ -0,0 +1,188 @@
...
+//
+// Please make sure to add verbose comments when changing code to make the
+// lifes of future readers easier.
+//
+// ****************************************************************************
+
+// Copy the source item to destination.  Use extract to untar the source if
+// it's a tar archive.
+func Copy(source *CopyItem, destination *CopyItem, extract bool) error {
+	// First, do the man-page dance begin.  See podman-cp(1) for details.
```
TomSweeneyRedHat, line 31: "dance begin" or "dance again"?

pkg/copy/copy.go
```diff
@@ -0,0 +1,188 @@
...
+// ********************************* NOTE *************************************
+//
+// Most security bugs are caused by attackers playing around with symlinks
+// trying to escape from the container onto the host and/or trick into data
+// corruption on the host.  Hence, file operations on containers (including
+// *stat) should always be handled by `github.com/containers/buildah/copier`
+// which makes sure to evaluate files in a chroot'ed environment.
+//
+// Please make sure to add verbose comments when changing code to make the
+// lifes of future readers easier.
```
haircommander, line 24: s/lifes/lives :)

TomSweeneyRedHat:
I may have missed the convo, but any thoughts on moving copier out of Buildah and into c/common?  This is one area that it would have been nice if Buildah/Podman used the same containers, we could make a bunch more of this code common between the two.

test/system/065-cp.bats
```diff
@@ -7,6 +7,312 @@
...
+    srcdir=$PODMAN_TMPDIR/cp-test-stdin
+    mkdir -p $srcdir
+    rand_filename=$(random_string 20)
+    rand_content=$(random_string 50)
+    echo $rand_content > $srcdir/$rand_filename
+    chmod 644 $srcdir/$rand_filename
+
+    # Now tar it up!
+    tar_file=$PODMAN_TMPDIR/archive.tar.gz
+    tar -zvcf $tar_file $srcdir
```
edsantiago, line 87: Absolute paths in tarballs make me nervous. Would you consider using `-C` instead, with corresponding changes below (`s/$srcdir/cp-test-stdin/`, possibly with a refactoring of the hardcoded `cp-test-stdin`, or even better, a randomly-generated dir name? Random is IMHO better than misleading, and the 'stdin' here is misleading):
```bash
    tar -C $PODMAN_TMPDIR -zvcf $tar_file .
```

test/system/065-cp.bats
```diff
@@ -7,6 +7,312 @@
...
+    run_podman rm -f cpcontainer
+}
+
+
+@test "podman cp file from container to host" {
+    skip_if_remote "podman-remote does not yet handle cp"
+
+    srcdir=$PODMAN_TMPDIR/cp-test-file-ctr-to-host
+    mkdir -p $srcdir
+
```
edsantiago, line 119: I meant to make time to refactor this, too, into table-driven form; but time got away from me. Given the example above, would you mind using that as a template to refactor this and the next two?

pkg/copy/copy.go
vrothberg: I hoped somebody would spot it :^)

pkg/copy/copy.go
vrothberg: :) thanks!

vrothberg:
> I may have missed the convo, but any thoughts on moving copier out of Buildah and into c/common? This is one area that it would have been nice if Buildah/Podman used the same containers, we could make a bunch more of this code common between the two.

SGTM

test/system/065-cp.bats
vrothberg: Sure, thanks!

test/system/065-cp.bats
vrothberg: It works like a charm and is much more readable. Thanks, Ed!

rhatdan:
No need to block this PR.  I am not crazy about moving Copier out of Buidlah since it is really tied to buildah[-copy/COPY functions in buildah.

test/system/065-cp.bats
vrothberg: done

vrothberg:
Changed the system tests to be table driven where applicable :+1:

vrothberg:

TomSweeneyRedHat:
Looks like the ' rdoproject.org/github-check' is failing here and in other PRs this morning with a retry limit failure of F31. `podman-package-fedora-31`.  

I'm not sure if there's anything we can do with that.

TomSweeneyRedHat:
LGTM
FWIW, I was not suggesting moving copier pkg to common as part of this work, just something to consider for the future.

PR #8571, opened 2020-12-03
Implement pod-network-reload

Luap99:
This adds a new command, 'podman network reload', to reload the
networks of existing containers, forcing recreation of firewall
rules after e.g. `firewall-cmd --reload` wipes them out.

Under the hood, this works by calling CNI to tear down the
existing network, then recreate it using identical settings. We
request that CNI preserve the old IP and MAC address in most
cases (where the container only had 1 IP/MAC), but there will be
some downtime inherent to the teardown/bring-up approach. The
architecture of CNI doesn't really make doing this without
downtime easy (or maybe even possible...).

At present, this only works for root Podman, and only locally.
I don't think there is much of a point to adding remote support
(this is very much a local debugging command), but I think adding
rootless support (to kill/recreate slirp4netns) could be
valuable.

### DIFF
--- a/cmd/podman/networks/reload.go
+++ b/cmd/podman/networks/reload.go
@@ -1,0 +1,69 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/containers/podman/v2/cmd/podman/common"
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/cmd/podman/utils"
+	"github.com/containers/podman/v2/cmd/podman/validate"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+var (
+	networkReloadDescription = `reload container networks, recreating firewall rules`
+	networkReloadCommand     = &cobra.Command{
+		Use:   "reload [options] [CONTAINER...]",
+		Short: "Reload firewall rules for one or more containers",
+		Long:  networkReloadDescription,
+		RunE:  networkReload,
+		Args: func(cmd *cobra.Command, args []string) error {
+			return validate.CheckAllLatestAndCIDFile(cmd, args, false, false)
+		},
+		ValidArgsFunction: common.AutocompleteContainers,
+		Example: `podman network reload --latest
+  podman network reload 3c13ef6dd843
+  podman network reload test1 test2`,
+		Annotations: map[string]string{
+			registry.ParentNSRequired: "",
+		},
+	}
+)
+
+var (
+	reloadOptions entities.NetworkReloadOptions
+)
+
+func reloadFlags(flags *pflag.FlagSet) {
+	flags.BoolVarP(&reloadOptions.All, "all", "a", false, "Reload network configuration of all containers")
+}
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode},
+		Command: networkReloadCommand,
+		Parent:  networkCmd,
+	})
+	reloadFlags(networkReloadCommand.Flags())
+	validate.AddLatestFlag(networkReloadCommand, &reloadOptions.Latest)
+}
+
+func networkReload(cmd *cobra.Command, args []string) error {
+	responses, err := registry.ContainerEngine().NetworkReload(registry.Context(), args, reloadOptions)
+	if err != nil {
+		return err
+	}
+
+	var errs utils.OutputErrors
+	for _, r := range responses {
+		if r.Err == nil {
+			fmt.Println(r.Id)
+		} else {
+			errs = append(errs, r.Err)
+		}
+	}
+
+	return errs.PrintErrors()
+}
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -639,6 +639,32 @@ func (c *Container) Sync() error {
 	return nil
 }
 
+// ReloadNetwork reconfigures the container's network.
+// Technically speaking, it will tear down and then reconfigure the container's
+// network namespace, which will result in all firewall rules being recreated.
+// It is mostly intended to be used in cases where the system firewall has been
+// reloaded, and existing rules have been wiped out. It is expected that some
+// downtime will result, as the rules are destroyed as part of this process.
+// At present, this only works on root containers; it may be expanded to restart
+// slirp4netns in the future to work with rootless containers as well.
+// Requires that the container must be running or created.
+func (c *Container) ReloadNetwork() error {
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return err
+		}
+	}
+
+	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
+		return errors.Wrapf(define.ErrCtrStateInvalid, "cannot reload network unless container network has been configured")
+	}
+
+	return c.reloadNetwork()
+}
+
 // Refresh is DEPRECATED and REMOVED.
 func (c *Container) Refresh(ctx context.Context) error {
 	// This has been deprecated for a long while, and is in the process of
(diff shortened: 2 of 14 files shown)


libpod/networking_linux.go
```diff
@@ -820,6 +834,63 @@ func getContainerNetNS(ctr *Container) (string, error) {
...
+		}
+	}
+
+	// teardownCNI will error if the iptables rules do not exists and this is the case after
+	// a firewall reload. The purpose of network reload is to recreate the rules if they do
+	// not exists so we should ignore possible errors.
+	_ = r.teardownCNI(ctr)
+	// FIXME: There two bad logrus.Error in the ocicni lib we should remove them
+	// https://github.com/cri-o/ocicni/blob/5243a9f23f8a9c1ab706cf1632760cb49eb2277f/pkg/ocicni/ocicni.go#L640
+	// https://github.com/cri-o/ocicni/blob/5243a9f23f8a9c1ab706cf1632760cb49eb2277f/pkg/ocicni/ocicni.go#L774
```
Luap99, line 886:
```
$ sudo bin/podman network reload --all
ERRO[0000] Error deleting network: running [/usr/sbin/iptables -t nat -D POSTROUTING -s 10.88.0.48 -j CNI-2210851933fa8e49e3ef4f38 -m comment --comment name: "podman" id: "b1b538e8bc4078fc3ee1c95b666ebc7449b9a97bacd15bcbe464a29e1be59c1c" --wait]: exit status 2: iptables v1.8.4 (legacy): Couldn't load target `CNI-2210851933fa8e49e3ef4f38':No such file or directory

Try `iptables -h' or 'iptables --help' for more information. 
ERRO[0000] Error while removing pod from CNI network "podman": running [/usr/sbin/iptables -t nat -D POSTROUTING -s 10.88.0.48 -j CNI-2210851933fa8e49e3ef4f38 -m comment --comment name: "podman" id: "b1b538e8bc4078fc3ee1c95b666ebc7449b9a97bacd15bcbe464a29e1be59c1c" --wait]: exit status 2: iptables v1.8.4 (legacy): Couldn't load target `CNI-2210851933fa8e49e3ef4f38':No such file or directory

Try `iptables -h' or 'iptables --help' for more information. 
ERRO[0000] Error deleting network: running [/usr/sbin/iptables -t nat -D POSTROUTING -s 10.88.0.47 -j CNI-520223cf89670efdff5d88d8 -m comment --comment name: "podman" id: "fe7e8eca56f844ec33af10f0aa3b31b44a172776e3277b9550a623ed5d96e72b" --wait]: exit status 2: iptables v1.8.4 (legacy): Couldn't load target `CNI-520223cf89670efdff5d88d8':No such file or directory

Try `iptables -h' or 'iptables --help' for more information. 
ERRO[0000] Error while removing pod from CNI network "podman": running [/usr/sbin/iptables -t nat -D POSTROUTING -s 10.88.0.47 -j CNI-520223cf89670efdff5d88d8 -m comment --comment name: "podman" id: "fe7e8eca56f844ec33af10f0aa3b31b44a172776e3277b9550a623ed5d96e72b" --wait]: exit status 2: iptables v1.8.4 (legacy): Couldn't load target `CNI-520223cf89670efdff5d88d8':No such file or directory

Try `iptables -h' or 'iptables --help' for more information. 
b1b538e8bc4078fc3ee1c95b666ebc7449b9a97bacd15bcbe464a29e1be59c1c
fe7e8eca56f844ec33af10f0aa3b31b44a172776e3277b9550a623ed5d96e72b
```
These errors are terrible the end user doesn't care and the command still works

Luap99:
@mheon PTAL

test/system/500-networking.bats
```diff
@@ -116,4 +116,50 @@ load helpers
...
+
+    # check that we still have the same mac and ip
+    run_podman inspect $cid --format "{{.NetworkSettings.IPAddress}}"
+    is "$output" "$ip" "IP address changed after podman network reload"
+    run_podman inspect $cid --format "{{.NetworkSettings.MacAddress}}"
+    is "$output" "$mac" "MAC address changed after podman network reload"
+
+    # check that we can still curl
+    run curl -s $SERVER/index.txt
+    is "$output" "$random_1" "curl 127.0.0.1:/index.txt"
```
edsantiago, line 162: Really nice test. May I suggest a final `run_podman rm -f $cid` for cleanup?

Luap99: Sure, I thought this would be done automatically

edsantiago: It is, but with a warning message (which I am usually the only one to see). When I initially created this test framework I decided to use the default podman root (`/var/lib/containers` etc) instead of an ad-hoc `--root/--runroot` in a tmpdir; the reason was to closely reflect a default end-user setup. As part of that, I chose to require each test to clean up after itself, sort of as an extra level of confidence against unexpected side effects. It has actually been helpful at times in catching `rm` race conditions.

libpod/networking_linux.go
Luap99: I opened https://github.com/cri-o/ocicni/pull/83 to remove these errors

libpod/container_api.go
```diff
@@ -639,6 +639,32 @@ func (c *Container) Sync() error {
...
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return err
+		}
+	}
+
+	if !c.ensureState(define.ContainerStateCreated, define.ContainerStateRunning) {
+		return errors.Wrapf(define.ErrCtrStateInvalid, "cannot reload network unless container network has been configured")
+	}
```
Luap99, line 663: @mheon Why did you accept `ContainerStateCreated`? Shouldn't we only do this for running containers?

mheon: Created containers have had their network namespaces created already, so they also may need to have their network re-configured. Internally, we have ContainerStateConfigured (container exists in Libpod but has not been mounted or created in the OCI runtime) and ContainerStateCreated (container has been mounted, network has been configured, container is created in the runtime - but has not yet been started).

Luap99: OK, thanks for the clarification.

cmd/podman/networks/reload.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+	}
+)
+
+var (
+	reloadOptions entities.NetworkReloadOptions
+)
+
+func reloadFlags(flags *pflag.FlagSet) {
+	flags.BoolVarP(&reloadOptions.All, "all", "a", false, "Reload network configuration of all containers")
```
zhangguanzhang, line 40: I see that line `podman network reload --latest`
so there need to add a `latest` option

Luap99: Latest is added with `validate.AddLatestFlag(networkReloadCommand, &reloadOptions.Latest)` below

rhatdan:
@mheon @baude PTAL

libpod/networking_linux.go
```diff
@@ -820,6 +834,63 @@ func getContainerNetNS(ctr *Container) (string, error) {
...
+				ctr.requestedMAC = requestedMAC
+				logrus.Debugf("Going to preserve container %s MAC address %s", ctr.ID(), ctr.requestedMAC.String())
+			}
+		}
+	}
+
+	// teardownCNI will error if the iptables rules do not exists and this is the case after
+	// a firewall reload. The purpose of network reload is to recreate the rules if they do
+	// not exists so we should ignore possible errors.
+	_ = r.teardownCNI(ctr)
```
mheon, line 883: Should we be ignoring the error here? Feels like we should at least logrus it.

mheon:
One nit, otherwise LGTM

mheon:
Thanks for taking this one over @Luap99

libpod/networking_linux.go
Luap99: see https://github.com/containers/podman/pull/8571#discussion_r535167473 I don't want to display this for users by default. Maybe logrus.Info is good enough?

mheon: Maybe we should string-parse the error and discard any `No such file or directory` errors, and treat others as real errors?

Luap99: Updated, PTAL

mheon:
LGTM!

rhatdan:
/lgtm

Nice work @Luap99

PR #8582, opened 2020-12-03
[CI:DOCS] The --net=container flag to Buildah is deprecated

mheon:
It's very confusing that `podman run --net=container` joins the network namespace of another container - but `podman build --net=container` creates a private network namespace. We've standardized on `--net=private` for this, and will eventually remove `podman build --net=container` as an alias for `--net=private`.

### DIFF
--- a/docs/source/markdown/podman-build.1.md
+++ b/docs/source/markdown/podman-build.1.md
@@ -412,7 +412,9 @@ Valid _mode_ values are:
 container full access to local system services such as D-bus and is therefore
 considered insecure.
 - **ns:**_path_: path to a network namespace to join.
-- **private**: create a new namespace for the container (default).
+- **private**: create a new namespace for the container (default). The
+**container** network mode is an alias for **private**, but has been deprecated
+and will be removed in a future release of Podman.
 
 #### **--no-cache**
 


docs/source/markdown/podman-build.1.md
```diff
@@ -416,7 +416,9 @@ Valid _mode_ values are:
 container full access to local system services such as D-bus and is therefore
 considered insecure.
 - **ns:**_path_: path to a network namespace to join.
-- **private**: create a new namespace for the container (default).
+- **private**: create a new namespace for the container (default). The
+**container** network mode is an alias for **private**, but has been deprecated
```
rhatdan, line 420: Don't document it at all, just remove it.  It has been removed from buildah bud documentation for a while.  This was an over site.  We can still work with the option, but no need to mention it.

mheon: Should I target this against the 2.2 branch instead, and we can completely remove in 3.0?

TomSweeneyRedHat: I like the thought, but will let @rhatdan decide.

rhatdan: Sure

mheon: Updated PR to point at the v2.2 branch, which you can apparently do without recreating, which is kind of neat.

TomSweeneyRedHat: Neat!  new learning for the day for me too!

TomSweeneyRedHat (commented):
LGTM

PR #8596, opened 2020-12-04
Use PasswordCallback instead of Password for ssh

### LINKED ISSUE #8498: macos: podman asks for login even when remote does not allow password authentications
/kind bug

**Description**

It seams that podman-remote goes into interactive password prompts regardless if remote servers allow them or not, which is different from ssh implementation and also breaks automation.

Please note that this problem seams to be particular to "podman" cli on macos (which is in fact podman-remote), but when trying to reproduce the same login prompt from a linux host while using the podman-remote command, it did fail correctly with a "no supported methods remain". Both versions were the same (2.1.1).

```
# proof server does not allow password authentication
ssbarnea@imac: ~
$ SSH_AUTH_SOCK= ssh -i /dev/null -F /dev/null -o PubkeyAuthentication=no root@leno
root@leno: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
FAIL: 255

# proof that podman asks for password when it should not do so
ssbarnea@imac: ~
$ CONTAINER_HOST=ssh://root@leno/run/podman/podman.sock CONTAINER_SSHKEY=/Users/ssbarnea/.ssh/id_rsa_insecure
 podman --log-level=debug ps
INFO[0000] podman filtering at log level debug
DEBU[0000] Called ps.PersistentPreRunE(podman --log-level=debug ps)
Login password:
Time: 0h:00m:06s
```

podman version 2.1.1
```
(paste your output here)
```

Irellevant because it does not work due to current bug
```
Login password:
```

```
2.1.1 installed with brew on macos
```

Related: https://github.com/containers/podman/issues/7806

afbjorklund:
Currently asking for login password, even if not supported by
the ssh server. So wait with prompt until actually requested.

Closes #8498

### DIFF
--- a/pkg/bindings/connection.go
+++ b/pkg/bindings/connection.go
@@ -207,11 +207,11 @@ func sshClient(_url *url.URL, secure bool, passPhrase string, identity string) (
 		authMethods = append(authMethods, ssh.Password(pw))
 	}
 	if len(authMethods) == 0 {
-		pass, err := terminal.ReadPassword("Login password:")
-		if err != nil {
-			return Connection{}, err
+		callback := func() (string, error) {
+			pass, err := terminal.ReadPassword("Login password:")
+			return string(pass), err
 		}
-		authMethods = append(authMethods, ssh.Password(string(pass)))
+		authMethods = append(authMethods, ssh.PasswordCallback(callback))
 	}
 
 	port := _url.Port()


TomSweeneyRedHat:
LGTM
assuming happy tests
Nice work @afbjorklund , TYVM!

afbjorklund:
Did manual tests, not sure if it had any regression suite (a bit of a hassle with ssh server and all)

ssbarnea (approved):
Thanks! Clearly this approach is much better as most systems are configured with passwords disabled.

afbjorklund:
For some reason, machine uses both a wellknown password **and** a ssh key for the VM...

Better leave the back door open, or something. I blame Docker (same as exporting /home)

Luap99 (approved):
LGTM

mheon:
This LGTM, but I'd like an ack from @jwhonce

PR #8597, opened 2020-12-04
Close image rawSource when each loop ends

### LINKED ISSUE #8583: (podman image sign) uses an ImageSource after closing it
After https://github.com/containers/podman/blob/70284b18cc7d4dc478c439cdda60ba64cfec060b/pkg/domain/infra/abi/images.go#L725 , all future uses of `rawSource` are invalid.

QiWang19:
Previously close rawSouce in the middle makes future use of rawSource invalid.
Move the rawSource.Close() to the end of each loop.

Close #8583 

### DIFF
--- a/pkg/domain/infra/abi/images.go
+++ b/pkg/domain/infra/abi/images.go
@@ -714,83 +714,90 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 	}
 
 	for _, signimage := range names {
-		srcRef, err := alltransports.ParseImageName(signimage)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error parsing image name")
-		}
-		rawSource, err := srcRef.NewImageSource(ctx, sc)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error getting image source")
-		}
-		err = rawSource.Close()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


QiWang19:
@mtrmac PTAL

mtrmac (commented):
Clearly an improvement, thanks for tackling this!

Ideally the loop body would be moved into a separate function (or an internal anonymous function, `func(){}()` and `defer` would be used to ensure the source is closed on all error paths.

pkg/domain/infra/abi/images.go
```diff
@@ -722,10 +722,6 @@ func (ir *ImageEngine) Sign(ctx context.Context, names []string, options entitie
 		if err != nil {
 			return nil, errors.Wrapf(err, "error getting image source")
 		}
-		err = rawSource.Close()
```
rhatdan, line 725: There is a chance at rawsouce not being closed.  I agree with Miloslav, to move this to a function and 
defer rawSource.Close()

rhatdan:
LGTM, although I would have created a separate named function, but this works.

mtrmac (approved):
LGTM.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
/lgtm
/hold
holding for tests

QiWang19:
Needs approved label.

PR #8599, opened 2020-12-04
Repeat system pruning until there is nothing removed

rhatdan:

### DIFF
--- a/cmd/podman/system/prune.go
+++ b/cmd/podman/system/prune.go
@@ -73,32 +73,31 @@ Are you sure you want to continue? [y/N] `, volumeString)
 			return nil
 		}
 	}
+
 	// TODO: support for filters in system prune
 	response, err := registry.ContainerEngine().SystemPrune(context.Background(), pruneOptions)
 	if err != nil {
 		return err
 	}
-	// Print pod prune results
-	fmt.Println("Deleted Pods")
-	err = utils.PrintPodPruneResults(response.PodPruneReport)
+	// Print container prune results
+	err = utils.PrintContainerPruneResults(response.ContainerPruneReport, true)
 	if err != nil {
 		return err
 	}
-	// Print container prune results
-	fmt.Println("Deleted Containers")
-	err = utils.PrintContainerPruneResults(response.ContainerPruneReport)
+	// Print pod prune results
+	err = utils.PrintPodPruneResults(response.PodPruneReport, true)
 	if err != nil {
 		return err
 	}
 	// Print Volume prune results
 	if pruneOptions.Volume {
-		fmt.Println("Deleted Volumes")
-		err = utils.PrintVolumePruneResults(response.VolumePruneReport)
+		err = utils.PrintVolumePruneResults(response.VolumePruneReport, true)
 		if err != nil {
 			return err
 		}
 	}
 	// Print Images prune results
-	fmt.Println("Deleted Images")
-	return utils.PrintImagePruneResults(response.ImagePruneReport)
+	utils.PrintImagePruneResults(response.ImagePruneReport, true)
+
+	return nil
 }
--- a/docs/source/markdown/podman-system-prune.1.md
+++ b/docs/source/markdown/podman-system-prune.1.md
@@ -1,7 +1,7 @@
 % podman-system-prune(1)
 
 ## NAME
-podman\-system\-prune - Remove all unused container, image and volume data
+podman\-system\-prune - Remove all unused pod, container, image and volume data
 
 ## SYNOPSIS
 **podman system prune** [*options*]
@@ -16,7 +16,7 @@ By default, volumes are not removed to prevent important data from being deleted
 ## OPTIONS
 #### **--all**, **-a**
 
-Remove all unused images not just dangling ones.
+Recursively remove all unused pod, container, image and volume data (Maximum 50 iterations.)
 
 #### **--force**, **-f**
 
--- a/cmd/podman/containers/prune.go
+++ b/cmd/podman/containers/prune.go
@@ -78,5 +78,5 @@ func prune(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	return utils.PrintContainerPruneResults(responses)
+	return utils.PrintContainerPruneResults(responses, false)
 }
--- a/cmd/podman/images/prune.go
+++ b/cmd/podman/images/prune.go
@@ -71,5 +71,5 @@ Are you sure you want to continue? [y/N] `)
 		return err
 	}
 
-	return utils.PrintImagePruneResults(results)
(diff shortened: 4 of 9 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/7990

rhatdan:
@srcshelton PTAL

srcshelton:
Okay, let me patch this into podman-2.2.0 and I'll check it out.

I'll do system rebuild without and check that multiple runs are required, and then again with to confirm it's fixed.

This may take a good number of hours…

(Although it does look as if it can't fail to resolve the problem!)

cmd/podman/system/prune.go
```diff
@@ -73,32 +73,55 @@ Are you sure you want to continue? [y/N] `, volumeString)
...
-	fmt.Println("Deleted Containers")
-	err = utils.PrintContainerPruneResults(response.ContainerPruneReport)
-	if err != nil {
-		return err
-	}
-	// Print Volume prune results
-	if pruneOptions.Volume {
-		fmt.Println("Deleted Volumes")
-		err = utils.PrintVolumePruneResults(response.VolumePruneReport)
+	for {
```
mheon, line 76: Recommendation: We should put a max iteration count on this - if we haven't removed everything in, say, 25 loops, we probably have a problem.

srcshelton: Just a quick note to agree, but suggest a larger number: I've seen 40-odd manual iterations before… so set the bar at 50?

TomSweeneyRedHat: Yeah, I'm not at all a fan of a naked for loop.

TomSweeneyRedHat: If each iteration takes a beat, we might want to add a message like "Working" that adds a dot after each iteration or after some number of iterations.

rhatdan: Each successful interaction will print purged lines, So no need for heartbeat.

srcshelton:
... a couple of questions:

* The earlier patch to repeatedly remove stale images when using `podman image prune` was said to use the same code-path as the image-pruning stage of `podman system prune` - is this not actually the case?

* Did the `podman image prune` patch also have a maximum number of iterations set - and, if so, do both commands now share a consistent limit?

srcshelton:
Also, I just did another `system prune` which removed a _large_ number of images, and the output froze for quite some time with only half of the last hash to be removed written to the terminal - could output-buffering be disabled for this operation (as I assume that an output buffer was filed when half the hash had been printed, and the delay in further output was the buffer not flushing again until filled)?

(It could also be a symptom of another issue, such as the host system locking temporarily due to high I/O?  I didn't see any other signs to suggest this being the case, though - other interactive tasks running at the same time weren't affected, and it was a long pause - at least 30s or so once I'd noticed it, but could have been much longer since I was doing other things and only checking the output occasionally)

docs/source/markdown/podman-system-prune.1.md
```diff
@@ -16,7 +16,7 @@ By default, volumes are not removed to prevent important data from being deleted
 ## OPTIONS
 #### **--all**, **-a**
 
-Remove all unused images not just dangling ones.
+Recusively remove all unused images not just dangling ones. (Maximum 50 iterations.)
```
TomSweeneyRedHat, line 19:
```suggestion
Recursively remove all unused images, not just dangling ones. (Maximum 50 iterations.)
```

TomSweeneyRedHat:
Code LGTM, a couple of small doc nits.  Any chance to get a test geared towards this?

rhatdan:
@srcshelton `podman system prune` calls the same internal function to prune images as `podman image prune`

As it does with podman volume prune, container prune, pod prune.  So there should not be a difference.  I don't think the pause was caused by the output printing, but some kind of locking in the storage layer.  Is some other operation was running on your system, there is a good chance that the removal of content was blocked on a lock.

srcshelton:
Interesting - would a wait on a lock cause the output to pause half-way through printing an image hash, though?

(Even if it is a lock-wait, unbuffered output would at least slow full lines to be printed whilst waiting?)

Whilst there were other contains running when the pause in output occurred, these had been running for several days at this point, and there was no other podman activity: nothing was starting or stopping or creating/updating images.

My question was just trying to understand how image pruning was fixed, and system pruning uses the same internal function, and yet system pruning has needed fixing separately? And whether there's a recursion limit for image pruning and, if so, whether the two methods to prune images are using the same limit, or whether they differ (potentially causing future confusion!)?

rhatdan:
All of the printing is being done with fmt.Println().  We could change this, but it does seem like a corner case.

I did not do anything for image pruning.  I have just added a loop on system prune to try to prune content again, since one pass of pruning could free up other pruning.

docs/source/markdown/podman-system-prune.1.md
```diff
@@ -16,7 +16,7 @@ By default, volumes are not removed to prevent important data from being deleted
 ## OPTIONS
 #### **--all**, **-a**
 
-Remove all unused images not just dangling ones.
+Recursively remove all unused images, not just dangling ones. (Maximum 50 iterations.)
```
vrothberg, line 19: "all unused images" ... I think we remove more than just images

cmd/podman/system/prune.go
```diff
@@ -73,32 +74,58 @@ Are you sure you want to continue? [y/N] `, volumeString)
...
-	err = utils.PrintContainerPruneResults(response.ContainerPruneReport)
-	if err != nil {
-		return err
-	}
-	// Print Volume prune results
-	if pruneOptions.Volume {
-		fmt.Println("Deleted Volumes")
-		err = utils.PrintVolumePruneResults(response.VolumePruneReport)
+
+	const MAX = 50
```
vrothberg, line 78: It seems we're now working around the fact that `registry.ContainerEngine().SystemPrune(...)` isn't doing it's job correctly.  Having the code scattered between `cmd/podman` and `pkg/domain/...` seems like a recipe for trouble.

Could we move all the logic to `registry.ContainerEngine().SystemPrune(...)`?  I think it should remove _all data_, even if there are more than 50 iterations.  It's not friendly to use.  If I do a `rm -rf *` I don't want to `ls` afterwards to check if things were really removed.

vrothberg (commented):
Code LGTM

saschagrunert (approved):
LGTM

saschagrunert:
Ah alright:
```
Inconsistent subcommand descriptions:
  podman-system-prune.1.md         = 'Remove all unused pod, container, image and volume data'
  podman-system.1.md               = 'Remove all unused container, image and volume data'
Please ensure that the NAME section of podman-system-prune.1.md
matches the subcommand description in podman-system.1.md
```

PR #8604, opened 2020-12-04
Initial implementation of volume plugins

### LINKED ISSUE #4304: NFS volumes created using podman volume create do not support specifying volume driver name

/kind bug

**Description**
After filing bug reports #4249, #4248, and #4247, and seeing all 3 resolved by the release of 1.6.2, we resumed trying to use `podman volume create` to create and mount NFS-backed volumes as originally announced in 1.6.1.

While re-creating an NFS-backed volume using 1.6.2, we're unable specify the volume driver name.

Documentation covering NFS volumes is non-existent, so we cannot tell if this is user error or broken code. As a part of resolving this issue, please consider adding NFS test cases to your build pipeline and expanding the Examples section of `podman-volume-create.1.md` with more examples, including NFS ones.

1. Create a volume backed by an NFS filesystem while trying to specify the volume driver name (guessing at the NFS invocation here... again, there are no NFS examples in the documentation to reference). 
```
[root@testhost ~]# podman volume create --driver=local --opt type=nfs --opt o=addr=192.168.2.126,rw --opt device=:/exports/test/ podman_vol_create_test2
Error: error running volume create option: not yet implemented
```

`podman volume create` does not allow us to specify a volume driver name during volume creation.

`podman volume create` allows us to specify a volume driver name during volume creation.

```
[root@testhost ~]# podman version
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.12.10
```

```
[root@testhost ~]# podman info --debug
debug:
  compiler: gc
  git commit: ""
  go version: go1.12.10
  podman version: 1.6.2
host:
  BuildahVersion: 1.11.3
  CgroupVersion: v1
  Conmon:
    package: conmon-2.0.1-1.fc30.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.1, commit: 4346fbe0b2634b05857973bdf663598081240374'
  Distribution:
    distribution: fedora
    version: "30"
  MemFree: 29456236544
  MemTotal: 33539690496
  OCIRuntime:
    name: runc
    package: runc-1.0.0-93.dev.gitb9b6cc6.fc30.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc8+dev
      commit: e3b4c1108f7d1bf0d09ab612ea09927d9b59b4e3
      spec: 1.0.1-dev
  SwapFree: 16840126464
  SwapTotal: 16840126464
  arch: amd64
  cpus: 8
  eventlogger: journald
  hostname: testhost
  kernel: 5.2.18-200.fc30.x86_64
  os: linux
  rootless: false
  uptime: 168h 7m 19.81s (Approximately 7.00 days)
registries:
  blocked: null
  insecure: null
  search:
  - docker.io
  - registry.fedoraproject.org
  - quay.io
  - registry.access.redhat.com
  - registry.centos.org
store:
  ConfigFile: /etc/containers/storage.conf
  ContainerStore:
    number: 1
  GraphDriverName: overlay
  GraphOptions:
  GraphRoot: /var/lib/containers/storage
  GraphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "true"
  ImageStore:
    number: 2
  RunRoot: /var/run/containers/storage
  VolumePath: /var/lib/containers/storage/volumes
```

```
[root@testhost ~]# rpm -q podman
podman-1.6.2-2.fc30.x86_64
```

Fresh minimal install of Fedora 30 with:
`yum -y install vim nfs-utils buildah`
`yum -y distro-sync --enablerepo=updates-testing podman`

mheon:
This implements support for mounting and unmounting volumes backed by volume plugins. Support for actually retrieving plugins requires a pull request to land in containers.conf and then that to be vendored, and as such is not yet ready. Given this, this code is only compile tested. However, the code for everything past retrieving the plugin has been written - there is support for creating, removing, mounting, and unmounting volumes, which should allow full functionality once the c/common PR is merged.

A major change is the signature of the MountPoint function for volumes, which now, by necessity, returns an error. Named volumes managed by a plugin do not have a mountpoint we control; instead, it is managed entirely by the plugin. As such, we need to cache the path in the DB, and calls to retrieve it now need to access the DB (and may fail as such).

Notably absent is support for SELinux relabelling and chowning these volumes. Given that we don't manage the mountpoint for these volumes, I am extremely reluctant to try and modify it - we could easily break the plugin trying to chown or relabel it.

Fixes #4304

### DIFF
--- a/docs/source/markdown/podman-volume-create.1.md
+++ b/docs/source/markdown/podman-volume-create.1.md
@@ -17,7 +17,7 @@ driver options can be set using the **--opt** flag.
 
 #### **--driver**=*driver*
 
-Specify the volume driver name (default local).
+Specify the volume driver name (default **local**). Setting this to a value other than **local** Podman will attempt to create the volume using a volume plugin with the given name. Such plugins must be defined in the **volume_plugins** section of the **containers.conf**(5) configuration file.
 
 #### **--help**
 
@@ -30,13 +30,14 @@ Set metadata for a volume (e.g., --label mykey=value).
 #### **--opt**=*option*, **-o**
 
 Set driver specific options.
-For the default driver, `local`, this allows a volume to be configured to mount a filesystem on the host.
+For the default driver, **local**, this allows a volume to be configured to mount a filesystem on the host.
 For the `local` driver the following options are supported: `type`, `device`, and `o`.
 The `type` option sets the type of the filesystem to be mounted, and is equivalent to the `-t` flag to **mount(8)**.
 The `device` option sets the device to be mounted, and is equivalent to the `device` argument to **mount(8)**.
 The `o` option sets options for the mount, and is equivalent to the `-o` flag to **mount(8)** with two exceptions.
 The `o` option supports `uid` and `gid` options to set the UID and GID of the created volume that are not normally supported by **mount(8)**.
-Using volume options with the `local` driver requires root privileges.
+Using volume options with the **local** driver requires root privileges.
+When not using the **local** driver, the given options will be passed directly to the volume plugin. In this case, supported options will be dictated by the plugin in question, not Podman.
 
 ## EXAMPLES
 
@@ -53,7 +54,8 @@ $ podman volume create --label foo=bar myvol
 ```
 
 ## SEE ALSO
-podman-volume(1), mount(8)
+**podman-volume**(1), **mount**(8), **containers.conf**(5)
 
 ## HISTORY
+January 2020, updated with information on volume plugins by Matthew Heon <mheon@redhat.com>
 November 2018, Originally compiled by Urvashi Mohnani <umohnani@redhat.com>
--- a/libpod/boltdb_state_internal.go
+++ b/libpod/boltdb_state_internal.go
@@ -497,6 +497,21 @@ func (s *BoltState) getVolumeFromDB(name []byte, volume *Volume, volBkt *bolt.Bu
 		}
 	}
 
+	// Retrieve volume driver
+	if volume.UsesVolumeDriver() {
+		plugin, err := s.runtime.getVolumePlugin(volume.config.Driver)
+		if err != nil {
+			// We want to fail gracefully here, to ensure that we
+			// can still remove volumes even if their plugin is
+			// missing. Otherwise, we end up with volumes that
+			// cannot even be retrieved from the database and will
+			// cause things like `volume ls` to fail.
+			logrus.Errorf("Volume %s uses volume plugin %s, but it cannot be accessed - some functionality may not be available: %v", volume.Name(), volume.config.Driver, err)
+		} else {
+			volume.plugin = plugin
+		}
+	}
+
 	// Get the lock
 	lock, err := s.runtime.lockManager.RetrieveLock(volume.config.LockID)
 	if err != nil {
--- a/libpod/define/errors.go
+++ b/libpod/define/errors.go
@@ -35,6 +35,10 @@ var (
 	// aliases.
 	ErrNoAliases = errors.New("no aliases for container")
 
+	// ErrMissingPlugin indicates that the requested operation requires a
+	// plugin that is not present on the system or in the configuration.
+	ErrMissingPlugin = errors.New("required plugin missing")
+
 	// ErrCtrExists indicates a container with the same name or ID already
 	// exists
 	ErrCtrExists = errors.New("container already exists")
--- a/libpod/runtime_volume_linux.go
+++ b/libpod/runtime_volume_linux.go
@@ -11,7 +11,9 @@ import (
 
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/libpod/events"
+	volplugin "github.com/containers/podman/v2/libpod/plugin"
 	"github.com/containers/storage/pkg/stringid"
+	pluginapi "github.com/docker/go-plugins-helpers/volume"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -53,6 +55,14 @@ func (r *Runtime) newVolume(ctx context.Context, options ...VolumeCreateOption) 
 		return nil, errors.Wrapf(define.ErrVolumeExists, "volume with name %s already exists", volume.config.Name)
 	}
 
+	// Plugin can be nil if driver is local, but that's OK - superfluous
+	// assignment doesn't hurt much.
+	plugin, err := r.getVolumePlugin(volume.config.Driver)
+	if err != nil {
+		return nil, errors.Wrapf(err, "volume %s uses volume plugin %s but it could not be retrieved", volume.config.Name, volume.config.Driver)
+	}
+	volume.plugin = plugin
+
 	if volume.config.Driver == define.VolumeDriverLocal {
 		logrus.Debugf("Validating options for local driver")
 		// Validate options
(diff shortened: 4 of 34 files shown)


mheon:
https://github.com/containers/common/pull/371 had the c/common changes I need. From there, need a release, then I can vendor it and add support for actually retrieving plugins.

mheon:
Dropping WIP. All features are now implemented.

Problem: testing. I'm going to start working towards local testing now, but getting the test environment running in CI is going to be a very interesting prospect given that it needs a separate executable/process for the plugin and containers.conf modifications.

mheon:
Rebased to fix conflicts, started adding documentation to the man pages.

mheon:
Made significant prgress on testing using https://github.com/mheon/testvol/ today - verified that it is possible to create and inspect volumes using the plugin. Mounting then is unfortunately hanging, but that appears to be happening in the volume plugin, not Podman, which is promising.

rhatdan:
Great, even if we can get this partially working for next week, we can continue to fix it up after release, as long as we add no new functionality for 3.0.

mheon:
Hanging issue identified as relating to copy up step - something about the way we copy files from the container into the volume *strongly* disagrees with the plugin implementation. Disabled copy-up on volumes for now, and things seem to be working otherwise. Lingering issues: does not work with SELinux (we don't chcon the mountpoint because we don't own the mountpoint; I'm still on the fence as to whether or not this is safe).

mheon:
CI should now go green. There are still no tests for plugins, though; I'm working on those now.

Meanwhile, the rest of this should be ready for review. @containers/podman-maintainers PTAL

docs/source/markdown/podman-volume-create.1.md
```diff
@@ -17,7 +17,7 @@ driver options can be set using the **--opt** flag.
 
 #### **--driver**=*driver*
 
-Specify the volume driver name (default local).
+Specify the volume driver name (default **local**). Setting this to a value other than **local** Podman will attempt to create the volume using a volume plugin with the given name. Such plugins must be defined in the **volume_plugins** section of the **containers.conf** configuration file.
```
vrothberg, line 20:
```suggestion
Specify the volume driver name (default **local**). Setting this to a value other than **local** Podman will attempt to create the volume using a volume plugin with the given name. Such plugins must be defined in the **volume_plugins** section of the **containers.conf(5)** configuration file.
```

libpod/boltdb_state_internal.go
```diff
@@ -497,6 +497,16 @@ func (s *BoltState) getVolumeFromDB(name []byte, volume *Volume, volBkt *bolt.Bu
 		}
 	}
 
+	// Retrieve volume driver
+	if volume.UsesVolumeDriver() {
+		plugin, err := s.runtime.getVolumePlugin(volume.config.Driver)
+		if err != nil {
+			logrus.Errorf("Volume %s uses volume plugin %s, but it cannot be accessed - some functionality may not be available: %v", volume.Name(), volume.config.Driver, err)
```
vrothberg, line 504: I wonder why errors are non-fatal.  Does Docker behave the same way?

Could you add a comment?

libpod/define/errors.go
```diff
@@ -34,6 +34,10 @@ var (
 	// aliases.
 	ErrNoAliases = errors.New("no aliases for container")
 
+	// ErrMissingPlugin indicates that the requested operation requires a
+	// plugin that is not present on the system or in the configuration.
+	ErrMissingPlugin = errors.New("required plugin missing")
```
vrothberg, line 39:
```suggestion
	ErrMissingVolumePlugin = errors.New("required volume plugin missing")
```

To leave space for other types of plugins in the future.

libpod/runtime_volume_linux.go
```diff
@@ -66,25 +75,62 @@ func (r *Runtime) newVolume(ctx context.Context, options ...VolumeCreateOption)
...
-	}
-	if err := os.Chown(fullVolPath, volume.config.UID, volume.config.GID); err != nil {
-		return nil, errors.Wrapf(err, "error chowning volume directory %q to %d:%d", fullVolPath, volume.config.UID, volume.config.GID)
-	}
-	if err := LabelVolumePath(fullVolPath); err != nil {
-		return nil, err
+	// Now we get conditional: we either need to make the volume in the
+	// volume plugin, or on disk if not using a plugin.
+	if volume.plugin != nil {
+		// We can't chown, or relabel, or similar the path the volume is
```
vrothberg, line 81: Might be worth breaking the two branches into separate functions. The branches are big and have quite some comments.

vrothberg:
I think we also need to tweak `podman cp` a bit: https://github.com/containers/podman/blob/master/pkg/domain/infra/abi/containers_stat.go#L141

If the volume is using a driver, we may need to mount the volume and not point outside of the container's mount point.

rhatdan:
@mheon isn't the mount point within container runtime storage for SELinux labeling?  What AVCs you are seeing?

I would figure the container volume is just following SELinux rules and should respect the :Z and :z flags,  If those flags are not specified then no SELinux should be done on the volume at all.

mheon:
@vrothberg Exposing Mount/Unmount outside of Libpod would be required, and I'm very hesitant to do that in this PR - this needs to be ready and landed very soon so I can move on to other features. Sounds like a lovely bug for a future PR, though.

@rhatdan No, the mountpoint is managed by the volume plugin, not us. We do not tell them where to mount, and we do not tell them how to mount - so I'm concerned that, by changing SELinux labels, we may remove the volume plugin's access to the mountpoint. I don't think there's a dedicated SELinux type for volume plugins, so I'm honestly unsure what label they're using. Maybe the container engine one?

libpod/define/errors.go
mheon: I think having one is probably fine - "missing plugin" on a Volume command is pretty distinct from "missing plugin" in a network command

libpod/boltdb_state_internal.go
mheon: This is all to ensure that, if the plugin goes away, Podman does not become nonfunctional - we can still remove the volume from the plugin.

mheon: Well, from Podman, but not from the Plugin.

mheon:
@rhatdan If I read https://github.com/moby/moby/pull/20834 right, Docker is not SELinux relabelling these volumes either - looks like it mostly requires SELinux to be disabled

rhatdan:
Yup.

libpod/boltdb_state_internal.go
mheon: Done

mheon:
Tests added. Once I work out the issues and this goes green it is good to merge.

mheon:
@containers/podman-maintainers This should go green now. PTAL.

mheon:
@baude Seeing errors restoring the image - 
```
           github.com/containers/podman/v2/test/e2e.(*PodmanTestIntegration).RestoreArtifact(0xc0008c8fc8, 0x171f28f, 0x2c, 0x0, 0x0)
```

Mind taking a quick look?

umohnani8:
changes LGTM @baude @rhatdan PTAL

mheon:
It's green!

TomSweeneyRedHat:
LGTM
but definitely want another head nod or three on this one.

PR #8609, opened 2020-12-05
Prefer read/write images over read/only images

rhatdan:
With additional stores there is a risk that you could have
multiple images with the same name.  IE An older image in a
read/only store versus a newer version in the read/write store.

This patch will ignore multiple images with the same name iff
one is read/write and all of the others are read/only.  It will
still return errors if only multiple read/only images or multiple
read/write images exists. Multipl read/write images should be
impossible.

Fixes: https://github.com/containers/podman/issues/8176

### DIFF
--- a/libpod/image/utils.go
+++ b/libpod/image/utils.go
@@ -20,7 +20,11 @@ import (
 // a match on name:tag
 func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, error) {
 	_, searchName, searchSuspiciousTagValueForSearch := search.suspiciousRefNameTagValuesForSearch()
-	var results []*storage.Image
+	type Candidate struct {
+		name  string
+		image *Image
+	}
+	var candidates []Candidate
 	for _, image := range images {
 		for _, name := range image.Names() {
 			d, err := decompose(name)
@@ -29,23 +33,52 @@ func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, er
 				continue
 			}
 			_, dName, dSuspiciousTagValueForSearch := d.suspiciousRefNameTagValuesForSearch()
-			if dName == searchName && dSuspiciousTagValueForSearch == searchSuspiciousTagValueForSearch {
-				results = append(results, image.image)
+			if dSuspiciousTagValueForSearch != searchSuspiciousTagValueForSearch {
 				continue
 			}
-			// account for registry:/somedir/image
-			if strings.HasSuffix(dName, "/"+searchName) && dSuspiciousTagValueForSearch == searchSuspiciousTagValueForSearch {
-				results = append(results, image.image)
-				continue
+			if dName == searchName || strings.HasSuffix(dName, "/"+searchName) {
+				candidates = append(candidates, Candidate{
+					name:  name,
+					image: image,
+				})
 			}
 		}
 	}
-	if len(results) == 0 {
-		return &storage.Image{}, errors.Errorf("unable to find a name and tag match for %s in repotags", searchName)
-	} else if len(results) > 1 {
-		return &storage.Image{}, errors.Wrapf(define.ErrMultipleImages, searchName)
+	if len(candidates) == 0 {
+		return nil, errors.Errorf("unable to find a name and tag match for %s in repotags", searchName)
 	}
-	return results[0], nil
+
+	// If more then one candidate and the candidates all have same name
+	// and only one is read/write return it.
+	// Othewise return error with the list of candidates
+	if len(candidates) > 1 {
+		var (
+			rwImage    *Image
+			rwImageCnt int
+		)
+		names := make(map[string]bool)
+		for _, c := range candidates {
+			names[c.name] = true
+			if !c.image.IsReadOnly() {
+				rwImageCnt++
+				rwImage = c.image
+			}
+		}
+		// If only one name used and have read/write image return it
+		if len(names) == 1 && rwImageCnt == 1 {
+			return rwImage.image, nil
+		}
+		keys := []string{}
+		for k := range names {
+			keys = append(keys, k)
+		}
+		if rwImageCnt > 1 {
+			return nil, errors.Wrapf(define.ErrMultipleImages, "found multiple read/write images %s", strings.Join(keys, ","))
+		} else {
+			return nil, errors.Wrapf(define.ErrMultipleImages, "found multiple read/only images %s", strings.Join(keys, ","))
+		}
+	}
+	return candidates[0].image.image, nil
 }
 
 // getCopyOptions constructs a new containers/image/copy.Options{} struct from the given parameters, inheriting some from sc.
--- a/test/e2e/exists_test.go
+++ b/test/e2e/exists_test.go
@@ -38,7 +38,6 @@ var _ = Describe("Podman image|container exists", func() {
 		Expect(session).Should(Exit(0))
 	})
 	It("podman image exists in local storage by short name", func() {
-		Skip("FIXME-8165: shortnames don't seem to work with quay (#8176)")
 		session := podmanTest.Podman([]string{"image", "exists", "alpine"})
 		session.WaitWithDefaultTimeout()
 		Expect(session).Should(Exit(0))


rhatdan:
@edsantiago @mtrmac PTAL

rhatdan:

vrothberg (commented):
We need a test for it. Too easy to regress if we change change code.

edsantiago:
We already have a test; it's disabled. @rhatdan can you try removing the `Skip` and see if CI passes? https://github.com/containers/podman/blob/70284b18cc7d4dc478c439cdda60ba64cfec060b/test/e2e/exists_test.go#L41

libpod/image/utils.go
```diff
@@ -20,7 +20,32 @@ import (
...
+
+	// This function prefers read/write images over read/only images, if they match
+	appendImage := func(results []*storage.Image, image *Image) []*storage.Image {
+		if foundRW {
+			// If readwrite image exists ignore readonly images
+			if image.IsReadOnly() {
+				return results
+			}
+			// Append readwrite images exists ignore readonly images
+			return append(results, image.image)
```
mtrmac, line 36: This, as well as the `return results`, is hard to follow for me.

Just have this function do the `results = append(results)` without the two call sites having to duplicate that.

(Absolutely non-blocking: It would also be nice to have the “business as usual, append” case only once.)

libpod/image/utils.go
```diff
@@ -20,7 +20,32 @@ import (
...
+	var (
+		results []*storage.Image
+		foundRW bool
+	)
+
+	// This function prefers read/write images over read/only images, if they match
+	appendImage := func(results []*storage.Image, image *Image) []*storage.Image {
+		if foundRW {
+			// If readwrite image exists ignore readonly images
+			if image.IsReadOnly() {
```
mtrmac, line 32: So if, searching for `alpine`, there is a read-only `example.com/alpine`, and a read-write `example.net/alpine`, this will silently ignore the ambiguity and only return the read-write one. That doesn’t look right to me.

BTW `results[]` can only legitimately contain 0 or 1 element, so eliminating the array and checking for equality (or rather, image ID identity? I’m not at all sure) should be fairly easy.

mtrmac (commented):
If https://github.com/containers/podman/issues/8176#issuecomment-738908444 is right that the `define.ErrMultipleImages` check doesn’t work, this PR does not yet fix all of that bug report.

libpod/image/utils.go
rhatdan: I believe results could container more then 1, if you had X additional stores, each one with a readonly  alpine, then we would have X results.

rhatdan: On your first comment, I would say that I believe the goal of the search is to find the image that the application should use.  In the case of multiple images with only one Read/Write, then the Read/Write image should be used.  If there are multiple read/only or multiple read/writes, then we get the ambiguity.

mtrmac: I meant that the _return value_ of the function is at most one value, or the function returns an error. So collecting more than one already guarantees that there will be an error.

WRT preferring read-write:
- AFAICT after a fair bit of looking, the purpose of this heuristic search is completely undocumented. It’s a bit tempting to guess that it only exists as a workaround for some inaccurate/incorrect search that used to exist elsewhere (in /usr/bin/atomic quite likely). So I’m very reluctant to add any more rules to the heuristic (although given I have no idea why this exists, I should have very little say).
- The original report already starts with `podman image exists`, which really does not care about read-write/read-only one way or the other. `podman run` can, AFAIK, also work just fine with read-only images (wasn’t that the whole point of adding additional read-only stores?). I’d even argue that _most_ users don’t care one bit whether the store is in a read-write store or not.

edsantiago: I'm not entirely following this conversation, so maybe my question is obvious/inapplicable: what is the intended purpose of `podman exists`? From my perspective, `podman exists foo` is a tool that tells me whether or not `podman run foo` will work. (Or `podman inspect foo`, or so on). Does this approach preserve that behavior?

mtrmac: To the extent both use `getLocalImage`, which I think is the case, both will be affected equally by whatever change is made.

rhatdan: Ok I simplified the function to return a read/write image, if it finds one, otherwise it will return the read/only image, if it exists.  Will return error if no match.

Ignores the case where there are multiple R/O images with the same name.

mtrmac: It also ignores the case where there are multiple ambiguous R/O images with different names, and multiple ambiguous R/W images with different names, as well as the earlier-discussed R/O vs. R/W name ambiguity, and makes `ErrMultipleImages` impossible to reach without fixing up consumers that expect it.

The simplification of the check is nice, though.

---

I was suggesting
```go
var res *storage.Image
…
if image /* matches */ {
    if res != nil && { // is ID the right thing to check across stores???
        return define.ErrMultipleImages
    }
    if res == nil || res.IsReadOnly { // image may be RO or RW
           res = image
    }
````
+ more or less the current after-the-loop code checking `res`.

---

+ the “podman image exists” check for `ErrMultipleImages` needs either fixing or removing (I can’t immediately remember which one), either way.

Comment (mtrmac) [reply libpod/image/utils.go]: (Updated the suggestion with an ID condition I lost when copy&pasting.)

**Comment by rhatdan** [comment]
@mtrmac PTAL
I changed to return the multiple "R/O" images error if R/W image is not found and there are multiple images.

**Comment by mtrmac** [comment]
@rhatdan Still https://github.com/containers/podman/pull/8609#discussion_r537551703 .

**Comment by rhatdan** [comment]
> So if, searching for alpine, there is a read-only example.com/alpine, and a read-write example.net/alpine, this will silently ignore 
> the ambiguity and only return the read-write one. That doesn’t look right to me.
With additional stores this can happen, since the underlying containers/storage will show an older version of alpine. In this case we will prefer the read/write store, since this was either pulled in a newer version of ":latest" or rebuilt/tagged a newer version of latest.  This is the situation that @edsantiago created to see this problem in the first place.

**Comment by mtrmac** [comment]
> With additional stores this can happen, since the underlying containers/storage will show an older version of alpine.

I’m not talking about different _versions_, but _entirely different images_. These are `example.com/alpine` and `something-entirely-different-and-not-example.com/alpine`.

(Your argument for accepting the newer image if both have exactly the same tag matching the user input, instead of failing on any change to the image, sounds very reasonable. I’m not at all sure about matching by ID. But that’s not my primary concern.)

File: libpod/image/utils.go [hunk 3]
```diff
@@ -29,23 +33,57 @@ func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, er
...
 			}
-			// account for registry:/somedir/image
-			if strings.HasSuffix(dName, "/"+searchName) && dSuspiciousTagValueForSearch == searchSuspiciousTagValueForSearch {
-				results = append(results, image.image)
-				continue
+			c := Candidate{
+				name:  dName,
+				image: image,
+			}
+			if dName == searchName {
```
mtrmac, line 43: AFAICS this can still be ambiguous because `dName` does not include the registry.

rhatdan: If I used name instead of dname as the key, would that work?

mtrmac: AFAICS no (two different values in the `HasSuffix` case, but primarily, there’s no need to worry about that.

You have already written code that produces the `candidates` data, and a uniqueness condition consuming that data. Why not just process the whole loop, and have a _single_ implementation of a _single_ condition? The loop is in-memory-only, with trivial cost.

libpod/image/utils.go
```diff
@@ -29,23 +33,50 @@ func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, er
...
-			if dName == searchName && dSuspiciousTagValueForSearch == searchSuspiciousTagValueForSearch {
-				results = append(results, image.image)
+			if dSuspiciousTagValueForSearch != searchSuspiciousTagValueForSearch {
 				continue
 			}
-			// account for registry:/somedir/image
-			if strings.HasSuffix(dName, "/"+searchName) && dSuspiciousTagValueForSearch == searchSuspiciousTagValueForSearch {
-				results = append(results, image.image)
-				continue
+			c := Candidate{
```
mtrmac, line 39: (This can be moved inside the condition, or even inlined.)

libpod/image/utils.go
```diff
@@ -20,7 +20,11 @@ import (
 // a match on name:tag
 func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, error) {
 	_, searchName, searchSuspiciousTagValueForSearch := search.suspiciousRefNameTagValuesForSearch()
-	var results []*storage.Image
+	type Candidate struct {
+		name  string
+		image *Image
```
mtrmac, line 25: (This can probably store a `*storage.Image` directly.)

libpod/image/utils.go
```diff
@@ -29,23 +33,50 @@ func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, er
...
+		var (
+			rwName  string
+			rwImage *Image
+		)
+		names := make(map[string]bool)
+		for _, c := range candidates {
+			names[c.name] = true
+			if !c.image.IsReadOnly() {
+				if rwImage != nil {
+					return nil, errors.Wrapf(define.ErrMultipleImages, "multiple read/write images found %s and %s", rwName, c.name)
```
mtrmac, line 63: Shouldn’t this still return all of the `keys` values? There might be _even_ more matching names, some RW, some RO.

(Sure, the previous version did not return all the values; if this is trying to do so, it should do so reliably.)

libpod/image/utils.go
```diff
@@ -29,23 +33,50 @@ func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, er
...
+		}
+		// If only one name used and have read/write image return it
+		if len(names) == 1 && rwImage != nil {
+			return rwImage.image, nil
+		}
+		keys := []string{}
+		for k := range names {
+			keys = append(keys, k)
+		}
+		return nil, errors.Wrapf(define.ErrMultipleImages, "found %s", keys)
```
mtrmac, line 77: Does this format as expected, or is a `strings.Join` necessary?

libpod/image/utils.go
```diff
@@ -29,23 +33,50 @@ func findImageInRepotags(search imageParts, images []*Image) (*storage.Image, er
...
 	}
-	return results[0], nil
+	return nil, errors.Errorf("unable to find a name and tag match for %s in repotags", searchName)
```
mtrmac, line 79: (Absolutely non-blocking: I found the previous version, with the `len() == 0` case at the front, a bit more readable. All the conditions are together; the way it is now, I have to scroll back and see what this case is about.

mtrmac (commented):
ACK overall.

---

Still outstanding:
> If https://github.com/containers/podman/issues/8176#issuecomment-738908444 is right that the `define.ErrMultipleImages` check doesn’t work, this PR does not yet fix all of that bug report.

or maybe split that into a separate tracked issue?

libpod/image/utils.go
rhatdan: It needs to be Image so I can check IsReadOnly.

libpod/image/utils.go
rhatdan: Fixed

libpod/image/utils.go
rhatdan: Looks better with strings.Join, fixed.

libpod/image/utils.go
rhatdan: Rearranged.

libpod/image/utils.go
rhatdan: Moved

Issue #8442, opened 2020-11-21
Feature: Static IP with "podman play kube --network=<network> --ip=<IP>"


hoonetorg:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

It would be useful if ```podman play kube``` would support using static IP for the generated pod.
The usecase for me is a dhcphelper pod which uses the host network wants to talk to a app pod, which has dnsmasq running in it. Forwarding DHCP requests from dhcphelper pod to app pod with dnsmasq container is easy by using a static IP.

**Steps to reproduce the issue:**

1. ```podman play kube app.yaml --network=app --ip=192.168.2.253```

**Describe the results you received:**

```Error: unknown flag: --ip```

**Describe the results you expected:**

```
Pod:
553d4b...
Containers:
xda643...
381d9f...
y1c49f...
43459c...
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.20, commit: '
  cpus: 2
  distribution:
    distribution: debian
    version: "10"
  eventLogger: journald
  hostname: youdontneedtoknow
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.19.0-12-amd64
  linkmode: dynamic
  memFree: 877731840
  memTotal: 4138450944
  ociRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: |-
      runc version 1.0.0~rc6+dfsg1
      commit: 1.0.0~rc6+dfsg1-3
      spec: 1.0.1
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 0
  swapTotal: 0
  uptime: 188h 3m 58.18s (Approximately 7.83 days)
registries:
  search:
  - docker.io
  - quay.io
store:
[...]

PR #8617, opened 2020-12-06
add static ip to kube play

mier85:
Signed-off-by: Michael Ernst <mier85@gmail.com>

this is implementing https://github.com/containers/podman/issues/8442

tried running lint & validate & tests, but they all failed before and after my commit. ( even tried the gate which is still partially in the CONTRIBUE.md but then realized that this seems to be no longer a thing )

### DIFF
--- a/cmd/podman/play/kube.go
+++ b/cmd/podman/play/kube.go
@@ -2,6 +2,7 @@ package pods
 
 import (
 	"fmt"
+	"net"
 	"os"
 
 	"github.com/containers/common/pkg/auth"
@@ -23,6 +24,7 @@ type playKubeOptionsWrapper struct {
 	TLSVerifyCLI   bool
 	CredentialsCLI string
 	StartCLI       bool
+	StaticIPCLI    string
 }
 
 var (
@@ -75,6 +77,10 @@ func init() {
 	flags.StringVar(&kubeOptions.Authfile, authfileFlagName, auth.GetDefaultAuthFile(), "Path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
 	_ = kubeCmd.RegisterFlagCompletionFunc(authfileFlagName, completion.AutocompleteDefault)
 
+	staticIPFlagName := "ip"
+	flags.StringVar(&kubeOptions.StaticIPCLI, staticIPFlagName, "", "Static IP address to assign to this pod")
+	_ = kubeCmd.RegisterFlagCompletionFunc(staticIPFlagName, completion.AutocompleteDefault)
+
 	if !registry.IsRemote() {
 
 		certDirFlagName := "cert-dir"
@@ -117,6 +123,14 @@ func kube(cmd *cobra.Command, args []string) error {
 		}
 		kubeOptions.Username = creds.Username
 		kubeOptions.Password = creds.Password
+	}
+
+	if kubeOptions.StaticIPCLI != "" {
+		ipAddress := net.ParseIP(kubeOptions.StaticIPCLI)
+		if ipAddress == nil {
+			return fmt.Errorf("failed parsing ip address: %s", kubeOptions.StaticIPCLI)
+		}
+		kubeOptions.StaticIP = ipAddress
 	}
 
 	report, err := registry.ContainerEngine().PlayKube(registry.GetContext(), args[0], kubeOptions.PlayKubeOptions)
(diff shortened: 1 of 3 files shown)


zhangguanzhang:
Lack of documentation and unit tests

mier85:
@zhangguanzhang I will add the documentation in https://github.com/containers/podman/blob/master/docs/source/markdown/podman-play-kube.1.md and some unit tests in the file you pointed to.

zhangguanzhang:
👌

cmd/podman/play/kube.go
```diff
@@ -23,6 +24,7 @@ type playKubeOptionsWrapper struct {
 	TLSVerifyCLI   bool
 	CredentialsCLI string
 	StartCLI       bool
+	StaticIPCLI    string
```
Luap99, line 27: Not needed see other comment

cmd/podman/play/kube.go
```diff
@@ -75,6 +77,10 @@ func init() {
 	flags.StringVar(&kubeOptions.Authfile, authfileFlagName, auth.GetDefaultAuthFile(), "Path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
 	_ = kubeCmd.RegisterFlagCompletionFunc(authfileFlagName, completion.AutocompleteDefault)
 
+	staticIPFlagName := "ip"
+	flags.StringVar(&kubeOptions.StaticIPCLI, staticIPFlagName, "", "Static IP address to assign to this pod")
+	_ = kubeCmd.RegisterFlagCompletionFunc(staticIPFlagName, completion.AutocompleteDefault)
+
```
Luap99, line 83: You can just use an IP flag. No need for the extra parsing logic
```suggestion
	staticIPFlagName := "ip"
	flags.IPVar(&kubeOptions.StaticIP, staticIPFlagName, nil, "Static IP address to assign to this pod")
	_ = kubeCmd.RegisterFlagCompletionFunc(staticIPFlagName, completion.AutocompleteNone)

```

cmd/podman/play/kube.go
```diff
@@ -119,6 +125,14 @@ func kube(cmd *cobra.Command, args []string) error {
...
 	}
 
+	if kubeOptions.StaticIPCLI != "" {
+		ipAddress := net.ParseIP(kubeOptions.StaticIPCLI)
+		if ipAddress == nil {
+			return fmt.Errorf("failed parsing ip address: %s", kubeOptions.StaticIPCLI)
+		}
+		kubeOptions.StaticIP = ipAddress
+	}
+
```
Luap99, line 135: Not needed see other comment

Luap99 (commented):
Thanks for the contribution @mier85 

I left some comment to simplify this a bit.

`podman play kube` can be run via the api and podman-remote. So you also need to add this option to the api and remote client.

The api parameter is defined here: https://github.com/containers/podman/blob/0c2a43b99db8a4fd75412a277da6de2731017d3e/pkg/api/handlers/libpod/play.go

Documentation for the api parameter has to be added here: https://github.com/containers/podman/blob/0c2a43b99db8a4fd75412a277da6de2731017d3e/pkg/api/server/register_play.go

Parameter for the remote client has to be added here: https://github.com/containers/podman/blob/master/pkg/bindings/play/play.go

cmd/podman/play/kube.go
mier85: flags will make things much easier, will update that

rhatdan:
@mier85 Still working on this?  We need to get this in by next week, for the 3.0 release, if we are going to get it done.

mier85:
@rhatdan still working on it. Will try to get this done ASAP this week. 

But I have a question. As several Pods could be spawned when deploying ( e.g. replicas in a kubernetes manifest ) , a single static IP wouldn't work. In this case, should be assign a range starting with the IP or should we fail and ask for several IPs or should this be a "single container only" option?

rhatdan:
I am no kubernetes expert, but I would figure fail would be the correct thing to do.

mheon:
Making static IPs incompatible with Deployments, at least initially, sounds like a fine solution.

hoonetorg:
@mier85: i see it that way: 
if kubernetes yaml is of type pod, only one pod will be deployed. if the kind is deployment than replicas is one, also only 1 container can be deployed. that are two valid scenarios for play kube, where a static ip could be used.
the thing with the range is probably a bit too difficult for the initial shot.

vrothberg:
Friendly ping. Are we good to continue?

rhatdan:
@umohnani8 Could you take this one over?

Luap99:
Replaced by #10043

Issue #8558, opened 2020-12-02
Both RootfsImageName and RootfsImageId must be set if either is set


rzlourenco:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Since I upgraded to 2.2.0, I can no longer run containers from successful intermediate build steps.  

**Steps to reproduce the issue:**

```
$ podman build .
[...]
STEP 12: USER ...
--> Using cache 5f166a88503a2a3ef76bb6dba7510795c5e7c664f0f17e181e78ae280ba5e5ee
--> 5f166a88503
STEP 13: RUN ...
[fails]
$ podman run -it --rm 5f166a88503 /bin/sh
Error: both RootfsImageName and RootfsImageID must be set if either is set: invalid argument
```

With `--log-level=debug`:

```
$ podman --log-level=debug run -it --rm 5f166a88503 /bin/sh
time="2020-12-02T16:07:38Z" level=info msg="podman filtering at log level debug"
time="2020-12-02T16:07:38Z" level=debug msg="Called run.PersistentPreRunE(podman --log-level=debug run -it --rm 5f166a88503 /bin/sh)"
time="2020-12-02T16:07:38Z" level=debug msg="Reading configuration file \"/usr/share/containers/containers.conf\""
[...]

PR #8623, opened 2020-12-07
container create: do not clear image name

vrothberg:
When creating a container, do not clear the input-image name before
looking up image names.  Also add a regression test.

Fixes: #8558

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -536,6 +536,17 @@ json-file | f
     run_podman untag $IMAGE $newtag $newtag2
 }
 
+# Regression test for issue #8558
+@test "podman run on untagged image: make sure that image metadata is set" {
+    run_podman inspect $IMAGE --format "{{.ID}}"
+    imageID="$output"
+
+    run_podman untag $IMAGE
+    run_podman run --rm $imageID ls
+
+    run_podman tag $imageID $IMAGE
+}
+
 @test "Verify /run/.containerenv exist" {
 	run_podman run --rm $IMAGE ls -1 /run/.containerenv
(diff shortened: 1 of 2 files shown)


rhatdan:
This will need to be back ported to 2.2.1

LGTM

test/system/030-run.bats
```diff
@@ -536,6 +536,17 @@ json-file | f
     run_podman untag $IMAGE $newtag $newtag2
 }
 
+# Regression test for issue #8558
+@test "podman run on untagged image: make sure that image metadata is set" {
+    run_podman inspect $IMAGE --format "{{.ID}}"
+    imageID="$output"
+
+    run_podman untag $IMAGE
+    run_podman run --rm $imageID ls
```
edsantiago, line 545: I find it helpful in regression tests to include the problem being fixed/tested. For example:
```bash
    # check exit status only; prior to #8623 this would error out with:
    # Error: both RootfsImageName and RootfsImageID must be set if either is set: invalid argument
```
First: should the error ever recur, this would help a maintainer understand how to fix it.

Second: it really isn't clear what this test is doing, which could lead to some future maintainer "helpfully" refactoring it or changing the invocation in some subtle way such that the test would no longer catch the original problem. A comment helps future maintainers understand that the code should not be touched.

Oh well, merged, too late now - but possibly for the backport, and for future such instances.

vrothberg: Thanks for the feedback. I will take care to better document in the future.

vrothberg: https://github.com/containers/podman/pull/8627

PR #8624, opened 2020-12-07
Docker compat API - containers create ignores the name

mlegenovic:
/containers/create compat endpoint does not set the name
correctly (#7857)


rhatdan:
LGTM
@jwhonce @baude @mheon @vrothberg PTAL

PR #8629, opened 2020-12-07
Do not error on installing duplicate shutdown handler

mheon:
Installing a duplicate shutdown handler fails, but if a handler with the same name is already present, we should be set to go. There's no reason to print a user-facing error about it.

This comes up almost nowhere because Podman never makes more than one Libpod runtime, but there is one exception (`system reset`) and the error messages, while harmless, were making people very confused (we got several bug reports that `system reset` was nonfunctional).


rhatdan:
LGTM

saschagrunert (approved):
LGTM

jwhonce:
/hold
/lgtm

pending tests

TomSweeneyRedHat:
LGTM
/hold cancel

PR #8630, opened 2020-12-07
Add systempaths=unconfined option

umohnani8:
Add the systempaths=unconfined option to --security-opt
to match the docker options for unmasking all the paths
that are masked by default.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -531,6 +531,13 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 			case "seccomp":
 				s.SeccompProfilePath = con[1]
 				s.Annotations[define.InspectAnnotationSeccomp] = con[1]
+			// this option is for docker compatibility, it is the same as unmask=ALL
+			case "systempaths":
+				if con[1] == "unconfined" {
+					s.ContainerSecurityConfig.Unmask = append(s.ContainerSecurityConfig.Unmask, []string{"ALL"}...)
+				} else {
+					return fmt.Errorf("invalid systempaths option %q, only `unconfined` is supported", con[1])
+				}
 			case "unmask":
 				s.ContainerSecurityConfig.Unmask = append(s.ContainerSecurityConfig.Unmask, strings.Split(con[1], ":")...)
 			default:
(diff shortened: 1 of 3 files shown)


mheon:
LGTM

TomSweeneyRedHat:
LGTM

umohnani8:
@rhatdan PTAL

docs/source/markdown/podman-run.1.md
```diff
@@ -893,6 +893,7 @@ Security Options
 - **no-new-privileges**: Disable container processes from gaining additional privileges
 - **seccomp=unconfined**: Turn off seccomp confinement for the container
 - **seccomp**=_profile.json_: Allowed syscall list seccomp JSON file to be used as a seccomp filter
+- **systempaths=unconfined**: Unmask all the paths that as masked by defaults. This is the same as **unmask=ALL**.
```
rhatdan, line 896: No need to document it, I would just leave it undocumented,  Since we have the greater feature.  This is only there for docker compatibility.

BTW Any time you update the -run man page, you probably need to update the -create man page.

rhatdan:
@cyphar PTAL

docs/source/markdown/podman-run.1.md
umohnani8: Removed from the doc and update the podman create doc as well.

cmd/podman/common/specgen.go
```diff
@@ -531,6 +531,11 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 			case "seccomp":
 				s.SeccompProfilePath = con[1]
 				s.Annotations[define.InspectAnnotationSeccomp] = con[1]
+			// this option is for docker compatibility, it is the same as unmask=ALL
+			case "systempaths":
+				if con[1] == "unconfined" {
```
cyphar, line 536: Maybe given an error if `con[1] != "unconfined"`? Otherwise LGTM.

rhatdan:
Could you also add a test for this, so we don't regress?

cmd/podman/common/specgen.go
umohnani8: Fixed

PR #8632, opened 2020-12-07
Change name of imageVolumes in container config JSON

### LINKED ISSUE #8613: Breaking Change in Container Config will cause JSON Decode Failures with Older Containers

/kind bug

**Description**

In the very old, pre-1.0 days before Libpod had proper named volume support, we emulated image volumes using bind mounts or tmpfs mounts. Part of this support involved a field in container configuration, `ImageVolumes` (type string), that determined whether we were using a tmpfs or bind mount. Libpod eventually grew support for named volumes, and the field was deprecated and became unused. I removed it in Podman 1.8 with https://github.com/containers/podman/commit/e3a549b7b1191a1667373cf60f8d08b96bbce7ba.

@vrothberg recently added a new, completely separate feature, where images can be mounted into containers as volumes (e.g. for security inspection). This again adds an `ImageVolumes` field (https://github.com/containers/podman/blob/master/libpod/container_config.go#L138) but now with a different struct type. These are completely incompatible in JSON terms and means that any containers created with Podman < 1.8.0 will be completely unusable with Podman 2.2.0.

1. Check out and build Podman 1.6.4 (any pre-1.8 will work, but we still support 1.6.4 in RHEL).

2. Create a container (`podman create`, `podman run`, doesn't really matter.

3. Upgrade to Podman 2.2.0 and run `podman ps`

Podman will not be able to read the container from the database due to a JSON decode error

Podman works as expected

The old `ImageVolumes` was not `omitempty`, so every container pre-removal is guaranteed to have it.

mheon:
Podman pre-1.8 also included a field with this name, which was a String. Podman 2.2.0 added a new field reusing the name but as a Struct. This completely broke JSON decode for pre-1.8 containers in Podman 2.2, resulting in completely broken behavior.

Re-name the JSON field and add a note that the old name should not be re-used to prevent this problem from re-occurring. This will still result in containers from 2.2.0 being broken (specifically, containers with image volumes will have them disappear) but this is the lesser of two evils. 

Fixes #8613

### DIFF
--- a/libpod/container_config.go
+++ b/libpod/container_config.go
@@ -135,7 +135,13 @@ type ContainerRootFSConfig struct {
 	// OverlayVolumes lists the overlay volumes to mount into the container.
 	OverlayVolumes []*ContainerOverlayVolume `json:"overlayVolumes,omitempty"`
 	// ImageVolumes lists the image volumes to mount into the container.
-	ImageVolumes []*ContainerImageVolume `json:"imageVolumes,omitempty"`
+	// Please note that this is named ctrImageVolumes in JSON to
+	// distinguish between these and the old `imageVolumes` field in Podman
+	// pre-1.8, which was used in very old Podman versions to determine how
+	// image volumes were handled in Libpod (support for these eventually
+	// moved out of Libpod into pkg/specgen).
+	// Please DO NOT re-use the `imageVolumes` name in container JSON again.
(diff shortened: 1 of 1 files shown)


mheon:
@containers/podman-maintainers PTAL, needed for 2.2.1

vrothberg (commented):
LGTM

TomSweeneyRedHat (commented):
LGTM

PR #8633, opened 2020-12-07
Bindings refactor

baude:
this is step one of refactoring our golang binaries.  we will no be
using structs to pass optional options.  required options will still
arguments to the binding itself.

the structs then have a generator to create helper functions which
should then be added to the git repo.

### DIFF
--- a/pkg/bindings/generator/generator.go
+++ b/pkg/bindings/generator/generator.go
@@ -1,0 +1,234 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+	"text/template"
+	"time"
+)
+
+var bodyTmpl = `package {{.PackageName}}
+
+import (
+{{range $import := .Imports}}	{{$import}}
+{{end}}
+
+)
+
+/*
+This file is generated automatically by go generate.  Do not edit.
+
+Created {{.Date}}
+*/
+
+// Changed
+func (o *{{.StructName}}) Changed(fieldName string) bool {
+	r := reflect.ValueOf(o)
+	value := reflect.Indirect(r).FieldByName(fieldName)
+	return !value.IsNil()
+}
+
+// ToParams
+func (o *{{.StructName}}) ToParams() (url.Values, error) {
+	params := url.Values{}
+	if o == nil {
+		return params, nil
+	}
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
+	s := reflect.ValueOf(o)
+	if reflect.Ptr == s.Kind() {
+		s = s.Elem()
+	}
+	sType := s.Type()
+	for i := 0; i < s.NumField(); i++ {
+		fieldName := sType.Field(i).Name
+		if !o.Changed(fieldName) {
+			continue
+		}
+		f := s.Field(i)
+		if reflect.Ptr == f.Kind() {
+			f = f.Elem()
+		}
+		switch f.Kind() {
+		case reflect.Bool:
+			params.Set(fieldName, strconv.FormatBool(f.Bool()))
+		case reflect.String:
+			params.Set(fieldName, f.String())
+		case reflect.Int, reflect.Int64:
+			// f.Int() is always an int64
+			params.Set(fieldName, strconv.FormatInt(f.Int(), 10))
+		case reflect.Slice:
+			typ := reflect.TypeOf(f.Interface()).Elem()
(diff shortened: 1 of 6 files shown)


pkg/domain/entities/params/generator/generator.go
```diff
@@ -0,0 +1,174 @@
...
+	"text/template"
+)
+
+var bodyTmpl = `package {{.PackageName}}
+
+import (
+{{range $import := .Imports}}	{{$import}}
+{{end}}
+)
+`
```
jwhonce, line 22: I would add a comment that this file is auto-generated and a timestamp for when it was written.

pkg/domain/entities/params/generator/generator.go
```diff
@@ -0,0 +1,174 @@
...
+
+func writeChangedFunc(f *os.File, structName string) error {
+	changedStruct := struct {
+		StructName string
+	}{
+		StructName: structName,
+	}
+
+	t, err := template.New(structName).Parse(changed)
```
jwhonce, line 49: You should move this parsing out of function and use MustParse(). No reason to build and keep a template per field.

pkg/domain/entities/params/generator/generator.go
```diff
@@ -0,0 +1,174 @@
...
+					tmplStruct := struct {
+						Name       string
+						StructName string
+						Type       string
+					}{
+						Name:       name,
+						StructName: inputStructName,
+						Type:       fieldType,
+					}
+					t, err := template.New(name).Parse(fieldTmpl)
```
jwhonce, line 135: I would move this parsing out of the loop. Again no need to parse and store a template per field.

Should the field tags be copied over as well from the source?

jwhonce (commented):
Should we add a SetValue() and AddValue() methods? I know dealing with the field types will be more difficult but we would be moving the conversion code into the generated layer. Could reduce errors.

This would be easier if url.Values{} exposed more container-like operations.

pkg/domain/entities/params/generator/generator.go
baude: field tags?

pkg/bindings/images/rm.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/domain/entities/params/images"
```
vrothberg, line 12: Maybe we can reflect on the package structures.  If we have `params/images` and `bindings/images`, users must chose which to use and rename the other.  I would _love_ the params and bindings to be in one flat package.  If others feel differently, could we flatten the `params` package?

vrothberg: Any comments?

pkg/bindings/images/types.go
```diff
@@ -0,0 +1,9 @@
+package images
+
+//go:generate go run ../generator/generator.go ImageRemove
+// ImageRemove are optional options for image removal
+type ImageRemove struct {
+	// Forces removes all containers based on the image
+	Force  *bool
```
jwhonce, line 7: If we need to override a field name vs. parameter, we could put a something like
```
Force *bool `param:"FoRcE"`
```
and then write that value into the generated code for `ToParam()` use.

pkg/bindings/images/types.go
```diff
@@ -0,0 +1,9 @@
+package images
+
+//go:generate go run ../generator/generator.go ImageRemove
+// ImageRemove are optional options for image removal
+type ImageRemove struct {
```
jwhonce, line 5:
```suggestion
type Remove struct {
```
We know from package this is for Images.

pkg/bindings/generator/generator.go
```diff
@@ -0,0 +1,234 @@
...
+				}
+				for _, val := range s {
+					params.Add(fieldName, val)
+				}
+			default:
+				return nil, errors.Errorf("unknown slice type %s", f.Kind().String())
+			}
+		case reflect.Map:
+			lowerCaseKeys := make(map[string][]string)
+			// I dont know if this code is needed anymore, TBD
```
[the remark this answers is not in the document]
baude: i want to deal with this when I hit something with filters so I for sure get this right and the regression tests will hold us accountable.

pkg/bindings/images/types.go
baude: will do in later pr

jwhonce:
LGTM

vrothberg (commented):
LGTM, really nice work, @baude!

PR #8635, opened 2020-12-07
Add system test for shell completion

Luap99:
There exists a unit test to ensure that shell completion functions are
defined. However there was no check about the quality of the provided
shell completions. Lets change that.

The idea is to create a general test that makes sure we are suggesting
containers,pods,images... for the correct commands. This works by
reading the command use line and checking for each arg if we provide
the correct suggestions for this arg.

It includes the following tests:
- flag suggestions if [options] is set
- container, pod, image, network, volume, registry completion
- path completion for the appropriate arg KEYWORDS (`PATH`,`CONTEXT`,etc.)
- no completion if there are no args
- completion for more than one arg if it ends with `...]`

The test does not cover completion values for flags and not every arg KEYWORD
is supported. This is still a huge improvement and covers most use cases.

This test spotted several inconsistencies between the completion and the
command use line. All of them have been adjusted to make the test pass.

The biggest advantage is that the completions always match the latest
command changes. So if someone changes the arguments for a command this
ensures that the completions must be adjusted.

### DIFF
--- a/cmd/podman/system/unshare.go
+++ b/cmd/podman/system/unshare.go
@@ -14,7 +14,7 @@ import (
 var (
 	unshareDescription = "Runs a command in a modified user namespace."
 	unshareCommand     = &cobra.Command{
-		Use:                   "unshare [COMMAND [ARG ...]]",
+		Use:                   "unshare [COMMAND [ARG...]]",
 		DisableFlagsInUseLine: true,
 		Short:                 "Run a command in a modified user namespace",
 		Long:                  unshareDescription,
--- a/test/system/600-completion.bats
+++ b/test/system/600-completion.bats
@@ -1,0 +1,272 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# Test podman shell completion
+#
+# Shell completion is provided via the cobra library
+# It is implement by calling a hidden subcommand called "__complete"
+#
+
+load helpers
+
+function check_shell_completion() {
+    local count=0
+
+    # Newline character; used for confirming string output
+    local nl="
+"
+
+    for cmd in $(_podman_commands "$@"); do
+        # Human-readable podman command string, with multiple spaces collapsed
+        name="podman"
+        if is_remote; then
+            name="podman-remote"
+        fi
+        command_string="$name $* $cmd"
+        command_string=${command_string//  / } # 'podman  x' -> 'podman x'
+
+        run_podman "$@" $cmd --help
+        local full_help="$output"
+
+        # The line immediately after 'Usage:' gives us a 1-line synopsis
+        usage=$(echo "$full_help" | grep -A1 '^Usage:' | tail -1)
+        [ -n "$usage" ] || die "podman $cmd: no Usage message found"
+
+        # If usage ends in '[command]', recurse into subcommands
+        if expr "$usage" : '.*\[command\]$' >/dev/null; then
+            check_shell_completion "$@" $cmd
+            continue
+        fi
+
+        # Trim to command path so we only have the args
+        args="${usage/$command_string/}"
+        # Trim leading whitespaces
+        args="${args#"${args%%[![:space:]]*}"}"
+
+        # Extra args is used to match the correct argument number for the command
+        # This is important because some commands provide different suggestions based
+        # on the number of arguments.
+        extra_args=()
+
+        for arg in $args; do
+
+            match=false
+            i=0
+            while true; do
+
+                case $arg in
+
+                # If we have options than we need to check if we are getting flag completion
+                "[options]")
+                    # skip this for remote it fails if a command only has the latest flag e.g podman top
+                    if ! is_remote; then
+                        run_completion "$@" $cmd "--"
+                        # If this fails there is most likely a problem with the cobra library
+                        is "${lines[0]}" "--.*" "Found flag in suggestions"
+                        [ ${#lines[@]} -gt 2 ] || die "No flag suggestions"
+                        _check_completion_end NoFileComp
+                    fi
+                    # continue the outer for args loop
+                    continue 2
+                    ;;
+
+                *CONTAINER*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
+                    is "$output" ".*-$random_container_name${nl}" "Found expected container in suggestions"
+
+                    match=true
+                    # resume
+                    ;;&
+
+                *POD*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
+                    is "$output" ".*-$random_pod_name${nl}" "Found expected pod in suggestions"
+                    _check_completion_end NoFileComp
+
+                    match=true
+                    # resume
+                    ;;&
+
+                *IMAGE*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
+                    is "$output" ".*localhost/$random_image_name:$random_image_tag${nl}" "Found expected image in suggestions"
+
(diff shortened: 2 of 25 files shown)


jwhonce:
@Luap99  LGTM, once you get past linter

TomSweeneyRedHat:
LGTM
assuming happy tests

Luap99:
@edsantiago PTAL

edsantiago:
@Luap99 I have been, all day, while juggling a bunch of other priorities. I hope to have some feedback by EOD, oops, it's EOD already. I hope to have some feedback within the next hour, at which time I have a hard cutoff. 

Really nice work as always. My feedback is so far very minor.

Luap99:
@edsantiago No problem, take your time. I just added support `podman-remote` with the force-push.

edsantiago:
My day has run long. I realize this is rude, but may I just offer this diff instead of one-by-one comments? Basically: typo fixes, a suggestion for your EOL check, a fix for an improper `untag`, and cleanup of the `pause` image:

[8635-suggestions.txt](https://github.com/containers/podman/files/5662239/8635-suggestions.txt)

This is just a first pass over the course of a long day; I'd like to make another pass tomorrow. 

One question: do you think it would be possible/easy to just run a subset of completion tests by default? This takes 37 seconds on my laptop, meaning it will take over a minute in CI. That's not a huge problem, but it's enough to make one consider.

Thanks as always. This is really nicely done.

Luap99:
> One question: do you think it would be possible/easy to just run a subset of completion tests by default? This takes 37 seconds on my laptop, meaning it will take over a minute in CI. That's not a huge problem, but it's enough to make one consider.

I would say the full test should run on every PR but it would be enough to only run this on one distro as root but for both `podman` and `podman-remote`. I don't want anyone breaking my completion logic.

test/system/600-completion.bats
```diff
@@ -0,0 +1,260 @@
...
+    run_podman volume create $random_volume_name
+
+
+    # $PODMAN may be a space-separated string, e.g. if we include a --url.
+    # Split it into its components and preserve any other args if present.
+    local -a podman_as_array=($PODMAN)
+    local    podman_path=${podman_as_array[0]}
+    local -a podman_args=("${podman_as_array[@]:1}")
+    # __completeNoDesc must be the fist arg if we running the completion cmd
+    PODMAN_COMPLETION="$podman_path __completeNoDesc ${podman_as_array[@]:1}"
```
edsantiago, line 231: Mix of temporary variables. TBH I don't think `podman_path` and `podman_args` add that much value, and it's just as readable to `s/$podman_path/${podman_as_array[0]}/` here and delete lines 228-229. Or, `s/${podman_as_array[@]:1}/${podman_args[@]}/`. Either is fine with me.

Also, `s/fist/first/` in line 230

test/system/600-completion.bats
```diff
@@ -0,0 +1,260 @@
...
+
+                *IMAGE*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
+                    is "$output" ".*localhost/$random_image_name:$random_image_tag${nl}" "No images in shell suggestions"
+
+                    # check that we complete the image with and without tag after at least one char is typed
+                    run_completion "$@" $cmd "${extra_args[@]}" "${random_image_name:0:1}"
+                    is "$output" ".*$random_image_name:$random_image_tag${nl}" "Image with tag not in shell suggestions"
+                    is "$output" ".*$random_image_name${nl}" "Image without tag not in shell suggestions"
+
```
edsantiago, line 95: (Possibly for a future enhancement): capture image ID in setup below, then test completion against `${random_image_id:0:2}`; confirm that `${...:0:12}` appears as a suggested completion. (Also confirm that it does _not_ appear in the default completion). Again, this is fine to leave for the future.

test/system/600-completion.bats
```diff
@@ -0,0 +1,260 @@
...
+                        is "${lines[0]}" "--.*" "First suggestion is not a flag"
+                        [ ${#lines[@]} -gt 2 ] || die "No flag suggestions"
+                    fi
+                    # continue the outer for args loop
+                    continue 2
+                    ;;
+
+                *CONTAINER*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
+                    is "$output" ".*-$random_container_name${nl}" "No containers in shell suggestions"
```
edsantiago, line 73: I find these descriptions confusing, because everywhere else in the tests I have used positive assertions and these are negative. I would kind of like to see that convention maintained here, but I won't block the PR because of it.
```bash
    is "$output" "..." "Found expected container in suggestions"
```

test/system/600-completion.bats
```diff
@@ -0,0 +1,260 @@
...
+                    is "$output" ".*$random_volume_name${nl}" "No volumes in shell suggestions"
+
+                    match=true
+                    # resume
+                    ;;&
+
+                *REGISTRY*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
+                    ### FIXME how can we get the configured registries?
+                    is "${lines[-1]}" "Completion ended with directive: ShellCompDirectiveNoFileComp" "Registries should have ShellCompDirectiveNoFileComp set"
```
edsantiago, line 119: Right around three or four is when I start thinking, hmmm, this should probably be refactored (and in this case there's eight, because the long-complicated-string is duplicated on each test line). I think it might make sense to add something like
```bash
function _check_completion_end() {
    is "${lines[-1]}" "Completion ended with directive: ShellCompDirective$1" "Final line of completion (FIXME: make this a better description)"
}

...
    _check_completion_end "NoFileComp"
```

test/system/helpers.bash
```diff
@@ -521,5 +521,15 @@ function remove_same_dev_warning() {
     output=$(printf '%s\n' "${lines[@]}")
 }
 
+# run 'podman help', parse the output looking for 'Available Commands';
+# return that list.
+function podman_commands() {
```
edsantiago, line 526: If you need to resubmit, could you prepend an underscore (`_podman_commands`) to mark this as a local internal helper function?

cmd/podman/system/unshare.go
```diff
@@ -14,7 +14,7 @@ import (
 var (
 	unshareDescription = "Runs a command in a modified user namespace."
 	unshareCommand     = &cobra.Command{
-		Use:                   "unshare [COMMAND [ARG ...]]",
+		Use:                   "unshare [COMMAND] [ARG...]",
```
edsantiago, line 17: This is not really accurate: a naïve reader could easily interpret that as "COMMAND -or- ARG" (which is of course meaningless). If at all possible, I'd like to see the original usage preserved.

edsantiago:
Meetings for the next two hours; I need to shift gears now. 

All my comments are relatively minor, but I think they add up to where I'd like to see a new push. Again, this is really really nice work, thank you.

>I  would say the full test should run on every PR but it would be enough to only run this on one distro as root but for both podman and podman-remote

Understood. We have no mechanism for doing that, so we'll just have to suffer on all runs.

Luap99:
> > I  would say the full test should run on every PR but it would be enough to only run this on one distro as root but for both podman and podman-remote
> 
> Understood. We have no mechanism for doing that, so we'll just have to suffer on all runs.

@edsantiago I could add `skip_if_rootless` and `skip_if_no_selinux` that should work, no?

Luap99:
Ok, I added your requested changes. This should be this ready now.

edsantiago:
/lgtm
/hold

Thanks again, and thanks for your patience with me

PR #8638, opened 2020-12-07
Pass full NetworkMode to ParseNetworkNamespace

kwiesmueller:
When creating a container using a docker client pointed to the compat api, setting the HostConfig.NetworkMode to something like `container:8a16f7abf12aee67342f3e5a9323c6ec33e70ac2e6365c8e4866a45a5b80478d` there is an error returned:

```
Error response from daemon: command rootless-cni-infra [alloc 0be27a5fb43a79dd3f5f2cd046e37cb961b5d57cc6719c53833d2645b7ae83fb container:8a16f7abf12aee67342f3e5a9323c6ec33e70ac2e6365c8e4866a45a5b80478d kubevirt-registry] in container 61a91ef143b3d9901b4f6f9a2d4f9bee1f6c27238ca3e3040139b01f8d662c2a failed with status 1, stdout="", stderr="no net configuration with name \"container:8a16f7abf12aee67342f3e5a9323c6ec33e70ac2e6365c8e4866a45a5b80478d\" in /etc/cni/net.d\n"
```

After looking at the code, it seems like https://github.com/kwiesmueller/podman/blob/master/pkg/specgen/namespaces.go#L284 checks for the passed in `ns` to have a `container:` prefix. But the `NetworkMode.NetworkName()` [method only returns `container`](https://github.com/moby/moby/blob/master/api/types/container/hostconfig_unix.go#L11).

Note that this fix is probably wrong, it solved it for me locally but I'm not familiar enough with what's going on yet and what this might cause.

/hold

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -237,7 +237,7 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	}
 
 	// netMode
-	nsmode, _, err := specgen.ParseNetworkNamespace(cc.HostConfig.NetworkMode.NetworkName())
+	nsmode, _, err := specgen.ParseNetworkNamespace(string(cc.HostConfig.NetworkMode))
 	if err != nil {
 		return nil, nil, err
 	}


Luap99:
@baude @jwhonce PTAL

mheon:
This looks more correct on first glance. I think discarding the error is probably wrong, we should at least be logging it.

rhatdan:
Thanks @kwiesmueller 
You have to sign your commits
git commit -a -s --amend
git push --force

LGTM

PR #8639, opened 2020-12-07
Fix network ls --filter invalid value flake

Luap99:
The filter is only validated when at least one network exists.


Luap99:
Example failed log: https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4709062841466880/html/int-podman-fedora-32-rootless-host.log.html

rhatdan:
LGTM

mheon:
LGTM

Issue #7838, opened 2020-09-29
generate kube should use tag


langdon:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When you use `podman generate kube` you expect to be able to run the result immediately. However, possibly depending on how you created the pod, you generate a yaml file that uses `:latest` for the tag. If you turn around and try to `play kube` that file, it fails trying to pull container images that the user has because it can't pull something with a `:latest` tag. `generate kube` should create yaml that is immediately runnable. To top it off, there is no indication of the issue being the tag unless you search and find #5186

**Steps to reproduce the issue:**

1. manually create containers in a pod called my-pod
1. podman generate kube my-pod > my-pod.yaml
1. podman pod rm -f my-pod
1. podman play kube my-pod.yaml

**Describe the results you received:**

```
Trying to pull registry.access.redhat.com/simpler-tools:latest...
  name unknown: Repo not found
</snip>
```

**Describe the results you expected:**
```
Pod: <UUID>
Containers: 
<UUID>
<UUID>
```
**Additional information you deem important (e.g. issue happens only occasionally):**

To make matters a bit more annoying, it does, in fact, create the pod with no UUID in the output or any indication it did so. Which, in order to do further testing/debugging, you have to manually remove. 

**Output of `podman version`:**

```
F32: 2.0.6
RHEL8: 1.9.3
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-2.0.6-1.fc32.x86_64
podman-1.9.3-2.module+el8.2.1+6867+366c07d6.x86_64
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

Yes

**Additional environment details (AWS, VirtualBox, physical, etc.):**

rhatdan:
@ashley-cui PTAL

umohnani8:
Tried this with podman master and it works fine for me. I created a pod with a fedora container running in it and was able to `podman play kube <yaml-file>`
```
podman play kube mypod.yaml
Trying to pull registry.fedoraproject.org/fedora:latest...
Getting image source signatures
Copying blob 22cef3226003 [--------------------------------------] 0.0b / 0.0b
Copying config e6b57c0d9f done  
Writing manifest to image destination
Storing signatures
Pod:
9e6ceeada01e996cd2cd28ec782e1ba3b6f31f0c4c71ea3e1138b182fd9cff2b
Container:
[...]

PR #8641, opened 2020-12-07
Do not pull if image domain is localhost

umohnani8:
With podman play kube, podman would always attempt to
pull if the image has the :latest tag. But this would
fail if the image was built locally and given latest
as the tag. Images build with podman and buildah have
localhost as the domain, so check if the domain is localhost.
If that is the case, then don't attempt a pull.

Fixes https://github.com/containers/podman/issues/7838

### DIFF
--- a/docs/source/markdown/podman-play-kube.1.md
+++ b/docs/source/markdown/podman-play-kube.1.md
@@ -15,6 +15,8 @@ Ideally the input file would be one created by Podman (see podman-generate-kube(
 
 Note: HostPath volume types created by play kube will be given an SELinux private label (Z)
 
+Note: If the `:latest` tag is used, Podman will attempt to pull the image from a registry. If the image was built locally with Podman or Buildah, it will have `localhost` as the domain, in that case, Podman will use the image from the local store even if it has the `:latest` tag.
+
 ## OPTIONS
 
 #### **--authfile**=*path*
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -212,8 +212,10 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
 			return nil, errors.Wrapf(err, "Failed to parse image %q", container.Image)
 		}
 		// In kube, if the image is tagged with latest, it should always pull
+		// but if the domain is localhost, that means the image was built locally
+		// so do not attempt a pull.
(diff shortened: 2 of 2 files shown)


umohnani8:
@rhatdan @mheon @baude PTAL

When using a locally built image, `podman generate kube` sets the container image as `localhost/builtimage:tag` so this check works in figuring out whether to pull the image or not.

mheon:
LGTM

docs/source/markdown/podman-play-kube.1.md
```diff
@@ -15,6 +15,8 @@ Ideally the input file would be one created by Podman (see podman-generate-kube(
 
 Note: HostPath volume types created by play kube will be given an SELinux private label (Z)
 
+Note: If the `:latest` tag is used, podman will attempt to pull the image from a registry. If the image was built locally with podman/buildah, it will have `localhost` as the domain, in that case podman will use the image from the local store even if it has the `:latest` tag.
```
TomSweeneyRedHat, line 18:
```suggestion
Note: If the `:latest` tag is used, Podman will attempt to pull the image from a registry. If the image was built locally with Podman or Buildah, it will have `localhost` as the domain, in that case, Podman will use the image from the local store even if it has the `:latest` tag.
```

TomSweeneyRedHat:
a few man page nits, otherwise LGTM

docs/source/markdown/podman-play-kube.1.md
umohnani8: Fixed

TomSweeneyRedHat (commented):
LGTM

vrothberg (commented):
Could we change the pull policy to "pull if newer"?

The `localhost/` prefixing of Podman and Buildah ignores the possibility of a registry running on `localhost`.

vrothberg:
> Could we change the pull policy to "pull if newer"?

The idea behind that proposal is that we'd be pulling an image if it exists on a registry (and has a different digest) and otherwise use the local one.

zhangguanzhang:
ci task failed, need someone help to restart them, code LGTM

vrothberg:
Why do you merge when my comments weren't addressed?

umohnani8:
oops, I should have commented that I am working on addressing @vrothberg's comments (spoke with him on slack). I will open a follow up PR to address those.

Issue #8651, opened 2020-12-08
Publishing more than one port range with automatically assigned ports leads to port mapping conflict


znerol:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Using the `--publish` flag it is possible to either specify the host port range explicitly or omit it. In the latter case the container runtime will allocate a random port range. If multiple `--publish` flags are used with multiple port ranges with host port range omitted, the container fails to start.

**Steps to reproduce the issue:**

1. Run a container and specify more than one port range without explicit host port range mapping.
    ```
    podman run --rm --publish 3000-3001 --publish 4000-4001 k8s.gcr.io/pause
    ```

**Describe the results you received:**

```
Error: conflicting port mappings for host port 1 (protocol tcp)
```

**Describe the results you expected:**

Container starting with two random host port ranges allocated to the container port ranges.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

Reproducible with both 2.2.1 and master branch

```
podman version
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.14
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

```
./bin/podman version
Version:      3.0.0-dev
API Version:  3.0.0
Go Version:   go1.14.10
Git Commit:   47d2a4be2af5143c9619a65ec641daa1025130a3
Built:        Tue Dec  8 21:47:44 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

Very likely not relevant.

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

compiled from source and installed from [kubic](https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/podman)

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

Yes

**Additional environment details (AWS, VirtualBox, physical, etc.):**

mheon:
I think #8652 will also fix this one, let me add a test

PR #8652, opened 2020-12-08
Correct port range logic for port generation

### LINKED ISSUE #8650: Publishing one randomized host port will lead to all host ports being randomized

/kind bug

**Description**

Using the `--publish` flag it is possible to either specify the host port explicitly or omit it. In the latter case the container runtime will allocate a random port. If multiple `--publish` flags are used, and at least one of them omits the host port, then _all_ host ports will be allocated randomly.

1. Run a container and specify one port with explicit host port mapping and one without.
    ```
    $ podman run --rm --publish 3000:3000 --publish 1234 k8s.gcr.io/pause
    ```
2. Examine the port mapping:
    ```
    $ podman port -l
    ```

```
$ podman port -l
3000/tcp -> 0.0.0.0:33205
1234/tcp -> 0.0.0.0:33273
```

```
$ podman port -l
3000/tcp -> 0.0.0.0:3000
1234/tcp -> 0.0.0.0:NNNNN
```

```
$ podman version
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.14
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

Not relevant (except if the bug isn't reproducible on less exotic systems).

Installed from [kubic](https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/podman)

Yes

Physical machine.

Running with `--log-level debug`, the following relevant lines appear:

```
DEBU[0000] Adding port mapping from 3000 to 3000 length 1 protocol "" 
DEBU[0000] Adding port mapping from 0 to 1234 length 1 protocol "" 
DEBU[0000] parsed reference into "[vfs@/home/lo/.local/share/containers/storage+/run/user/1000/containers]k8s.gcr.io/pause:latest" 
[...]
DEBU[0000] using systemd mode: false                    
DEBU[0000] Successfully assigned container port 3000 to host port 45799 (IP  Protocol tcp) 
DEBU[0000] Successfully assigned container port 1234 to host port 34203 (IP  Protocol tcp) 
```

Thus it looks like the command line parsing works correctly in [parseSplitPort](https://github.com/containers/podman/blob/47d2a4be2af5143c9619a65ec641daa1025130a3/cmd/podman/common/util.go#L206) but fails somewhere between there an the subsequent [specgen](https://github.com/containers/podman/blob/cec0e5749745f8773d69db3d75395a0c2390bce9/pkg/specgen/generate/ports.go#L226)

If only explicit port mappings are specified (i.e., one '--publish 3000:3000` flag) then the loglines read:

```
DEBU[0000] Adding port mapping from 3000 to 3000 length 1 protocol "" 
```

No lines are logged from `specgen/generate/parsePortMapping` at all. So, maybe the problem is in the code block within [if postAssignPort](https://github.com/containers/podman/blob/cec0e5749745f8773d69db3d75395a0c2390bce9/pkg/specgen/generate/ports.go#L174). When that branch is hit, then it looks like _all_ host ports are being randomized.

mheon:
The existing logic (Range > 0) always triggered, because range is guaranteed to be at least 1 (a single port has a range of 1, a two port range (e.g. 80-81) has a range of 2, and so on). As such this could cause ports that had a host port assigned to them by the user to randomly assign one instead.

Fixes #8650


mheon:
Now also fixes #8651

mheon:

Luap99 (commented):
LGTM

rhatdan:

baude:
nice work @mheon

PR #8663, opened 2020-12-09
archive endpoint massaging

vrothberg:
A series of commits to improve the archive endpoint and laying further groundwork for `remote cp`:
* pkg/copy: introduce a Copier
* archive: move stat-header handling into copy package
* pkg/copy: add parsing API

@jwhonce PTAL ... I figured to commit early to make your live easier


rhatdan:
LGTM

PR #8666, opened 2020-12-09
podman play should honour both ENTRYPOINT and CMD from image

cruwe:
Assigning the entrypoint  to `command` may be in line with the Kubernetes `command` and `args` pattern, but breaks normal behaviour when creating the container without passing any of those.

I have WIPed this, because I am testing a larger deployment, am uneasy about what to find next and don't want to mess up history with one-liners

### DIFF
--- a/pkg/specgen/generate/kube/kube.go
+++ b/pkg/specgen/generate/kube/kube.go
@@ -106,17 +106,20 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
 
 	// TODO: We dont understand why specgen does not take of this, but
 	// integration tests clearly pointed out that it was required.
-	s.Command = []string{}
+	// s.Command = []string{}
 	imageData, err := newImage.Inspect(ctx)
 	if err != nil {
 		return nil, err
 	}
 	s.WorkDir = "/"
+
+	// TODO: determine if `newImage.Inspect(ctx)` can ever return without `err`
+	// TODO: and have `imageData` or `imageData.Config` equal `nil`
+	// TODO: if it can, determine how to handle that error generally and globally
 	if imageData != nil && imageData.Config != nil {
 		if imageData.Config.WorkingDir != "" {
 			s.WorkDir = imageData.Config.WorkingDir
 		}
-		s.Command = imageData.Config.Entrypoint
 		s.Labels = imageData.Config.Labels
 		if len(imageData.Config.StopSignal) > 0 {
 			stopSignal, err := util.ParseSignal(imageData.Config.StopSignal)
@@ -126,13 +129,29 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
 			s.StopSignal = &stopSignal
 		}
 	}
-	if len(containerYAML.Command) != 0 {
-		s.Command = containerYAML.Command
-	}
+
 	// doc https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
-	if len(containerYAML.Args) != 0 {
-		s.Command = append(s.Command, containerYAML.Args...)
+
+	if len(containerYAML.Command) == 0 && len(containerYAML.Args) == 0 {
+		s.Entrypoint = imageData.Config.Entrypoint
+		s.Command = imageData.Config.Cmd
 	}
+
+	if len(containerYAML.Command) != 0 && len(containerYAML.Args) == 0 {
+		s.Entrypoint = containerYAML.Command
+		s.Command = nil
+	}
+
+	if len(containerYAML.Command) == 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = imageData.Config.Entrypoint
+		s.Command = containerYAML.Args
+	}
+
+	if len(containerYAML.Command) != 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = containerYAML.Command
+		s.Command = containerYAML.Args
+	}
+
 	// FIXME,
 	// we are currently ignoring imageData.Config.ExposedPorts
 	if containerYAML.WorkingDir != "" {
--- a/test/e2e/play_kube_test.go
+++ b/test/e2e/play_kube_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -490,6 +491,12 @@ func getDeployment(options ...deploymentOption) *Deployment {
 
 type deploymentOption func(*Deployment)
 
+func withDeploymentName(name string) deploymentOption {
+	return func(deployment *Deployment) {
+		deployment.Name = name
+	}
+}
+
 func withDeploymentLabel(k, v string) deploymentOption {
 	return func(deployment *Deployment) {
 		deployment.Labels[k] = v
@@ -752,6 +759,18 @@ type EnvFrom struct {
 	From string
 }
 
+type ContainerConfig struct {
+	Cmd        []string `json:"Cmd,omitempty"`
+	Entrypoint string   `json:"Entrypoint"`
+	// we do not yer care for the rest
+}
+
+type ContainerState struct {
+	Args   []string        `json:"Args,omitempty"`
+	Config ContainerConfig `json:"Config"`
+	// we do not yer care for the rest
+}
+
 func milliCPUToQuota(milliCPU string) int {
 	milli, _ := strconv.Atoi(strings.Trim(milliCPU, "m"))
 	return milli * defaultCPUPeriod
@@ -803,8 +822,15 @@ var _ = Describe("Podman play kube", func() {
 
 	})
 
-	It("podman play kube test correct command", func() {
-		pod := getPod()
+	// NOTICE: what is called k8s spec in the following section is described at the k8s docs at
+	// https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
+
+	It("podman play kube - k8s spec, 1st item, when you _do not_ supply command _or_ args", func() {
+		// given an image with `ENTRYPOINT` and `CMD` defined and a pod-spec _without any_,
+		// when the container is instantiated from the image, it should use `Entrypoint` and `Cmd`
+		// from the image and `Args` should equal `Cmd`
+
+		pod := getPod(withCtr(getCtr(withImage(redis), withCmd(nil), withArg(nil))))
 		err := generateKubeYaml("pod", pod, kubeYaml)
 		Expect(err).To(BeNil())
 
@@ -812,16 +838,35 @@ var _ = Describe("Podman play kube", func() {
 		kube.WaitWithDefaultTimeout()
 		Expect(kube.ExitCode()).To(Equal(0))
 
-		inspect := podmanTest.Podman([]string{"inspect", getCtrNameInPod(pod), "--format", "'{{ .Config.Cmd }}'"})
+		inspect := podmanTest.Podman([]string{"inspect", getCtrNameInPod(pod)})
 		inspect.WaitWithDefaultTimeout()
 		Expect(inspect.ExitCode()).To(Equal(0))
-		// Use the defined command to override the image's command
-		correctCmd := "[" + strings.Join(defaultCtrCmd, " ") + " " + strings.Join(defaultCtrArg, " ")
-		Expect(inspect.OutputToString()).To(ContainSubstring(correctCmd))
+
+		var containerState []ContainerState
+		err = json.Unmarshal([]byte(inspect.OutputToString()), &containerState)
+		Expect(err).To(BeNil())
+
+		// these we know from having had a look at the redis image
+		imgEntrypoint := `docker-entrypoint.sh`
+		imgCmd := []string{`redis-server`}
+		derivedArgs := imgCmd
+
+		containerEntrypoint := containerState[0].Config.Entrypoint
+		containerCmd := containerState[0].Config.Cmd
+		containerArgs := containerState[0].Args
+
+		Expect(containerEntrypoint).To(Equal(imgEntrypoint))
+		Expect(containerCmd).To(Equal(imgCmd))
+		Expect(containerArgs).To(Equal(derivedArgs))
 	})
 
-	It("podman play kube test correct command with only set command in yaml file", func() {
-		pod := getPod(withCtr(getCtr(withCmd([]string{"echo", "hello"}), withArg(nil))))
+	It("podman play kube - k8s spec, 2nd item, when you supply command _but no_ args", func() {
+		// given an image with `ENTRYPOINT` and `CMD` defined and a pod-spec with the `command`
+		// defined, when the container is instantiated from the image, it should replace any
+		// `Entrypoint` and `Cmd` and start with `command` _without_ further `Args`.
+
+		overrideCmd := []string{"echo", "hello"}
+		pod := getPod(withCtr(getCtr(withImage(redis), withCmd(overrideCmd), withArg(nil))))
 		err := generateKubeYaml("pod", pod, kubeYaml)
 		Expect(err).To(BeNil())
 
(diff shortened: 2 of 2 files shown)


mheon:
I'm not opposed but we should make this conditional, so it's not a breaking change. Maybe a `--use-cmd` flag?

cruwe:
I am starting to wonder: Am I missing some important context or intend which I do not know about?

cruwe:
I think I now got the intent. I am not satisfied with the verbosity, but I think the code should closely mirror the spec to prevent misunderstandings.

What is the project's position on duplication vs readability?

mheon:
I am a firm believer in readability being the first consideration (I am admittedly biased by the amount of code review I do :smile: )

pkg/specgen/generate/kube/kube.go
```diff
@@ -126,13 +125,31 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
...
+	if len(containerYAML.Command) == 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = imageData.Config.Entrypoint
+		s.Command = nil
 		s.Command = append(s.Command, containerYAML.Args...)
 	}
+
+	if len(containerYAML.Command) != 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = containerYAML.Command
+		s.Command = nil
+		s.Command = append(s.Command, containerYAML.Args...)
```
zhangguanzhang, line 150: I think these two lines are easier to read when combined into the following line:
```
s.Command = containerYAML.Args
```

pkg/specgen/generate/kube/kube.go
```diff
@@ -126,13 +125,31 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
...
+
+	if len(containerYAML.Command) != 0 && len(containerYAML.Args) == 0 {
+		s.Entrypoint = containerYAML.Command
+		s.Command = nil
+		}
+
+	if len(containerYAML.Command) == 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = imageData.Config.Entrypoint
+		s.Command = nil
 		s.Command = append(s.Command, containerYAML.Args...)
```
zhangguanzhang, line 144: I think these two lines are easier to read when combined into the following line:
```
s.Command = containerYAML.Args
```

zhangguanzhang:
and code need go fmt

pkg/specgen/generate/kube/kube.go
cruwe: put a comment there to clarify the intent

pkg/specgen/generate/kube/kube.go
cruwe: likewise

cruwe:
> and code need go fmt

that I do not get. I did and the IDE does at any commit. Nothing changed. Could you point me in the direction?

pkg/specgen/generate/kube/kube.go
```diff
@@ -126,13 +125,31 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
...
 		}
 	}
-	if len(containerYAML.Command) != 0 {
-		s.Command = containerYAML.Command
-	}
+
 	// doc https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
-	if len(containerYAML.Args) != 0 {
+
+	if len(containerYAML.Command) == 0 && len(containerYAML.Args) == 0 {
```
cruwe, line 131: btw, I am pretty sure that the initialization with `[]string{}` from L109 is not necessary. It runs fine here without and `s.ContainerBasicConfig.Command` and `.Entrypoint`, which in theory should suffer from the same problem, are initialized just fine.

pkg/specgen/generate/kube/kube.go
cruwe: ah, got it ... yes, you're right

zhangguanzhang:
you could right click the changed files., and choose the gofmt，or use `gofmt -w` command

cruwe:
Without exerting undue pressure, I'd like to point out that this PR fixes a regression:

Take for instance `postgresql` or `redis` from the docker-library: Both use the `docker-entrypoint.sh` -> `exec "$@"` pattern to do some chores at start-up time before doing a fork-replace into the main application. Starting the containers as a pod, it is reasonable (and defined by the standard mentioned as a comment to the code), that the container will start without having to explicitly pass `command` and `args`.

This behaviour is shown by `docker`, by `podman run`, by kubernetes and openshift, as also  byh `podman play kube` at version 2.1 and i strongly suspect the versions before, although I have not tested that. It stopped with `podman` v2.2 and the code I am PRing against.

We use the `podman play` feature a lot and I think highly of the idea to replace the moving and often elusive target of `docker-compose.yml` with the most sensible idea of k8s pod specs. Accordingly, I am most interested to get this code reviewed and eventually merged.

Please, tell me what I can do to have the process sped up. Thank you very much for your time and consideration.

zhangguanzhang:
you cloud see this for more detail : https://github.com/containers/podman/pull/8666/checks?check_run_id=1529080539
you code need to go fmt, and git commits need to rebase to 1 commit
ci will continue to run after that

cruwe:
> you cloud see this for more detail : https://github.com/containers/podman/pull/8666/checks?check_run_id=1529080539
> 
> you code need to go fmt, and git commits need to rebase to 1 commit.
> ci will continue to run after that.

@zhangguanzhang : Thanks!

cruwe:
I am willing to work on the failing tests. I am not particularly fluent in golang and absolutely new to ginko, so I'll definitely need some guidance. 

However, before I start working, could we reach consensus on the general idea to honour entrypoint and command from the image and be strict about the the k8s specs? @zhangguanzhang , @mheon, could you help me reaching that consensus?

mheon:
This behavior cannot be default, because it would be a breaking chance from the way things are now. I am OK with doing it, but it will need to be gated behind a flag.

daiaji:
This did cause the yaml I exported using the `podman generate kube` command to be unable to deploy using the `podman play kube` command.

cruwe:
Hi,

I already feared that there might be a severe misunderstanding present here, perhaps on my part, I do not know. I am not trying to implement a feature here, I observed an undocumented breaking change, assumed that to be an unintended side-effect of some major refactoring and attempted, not least because I am/was relying on the feature, to come up with a fix.

Consider the following pod definition:
```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: try
spec:
  containers:
    - name: try
      image: docker.io/library/redis:6
```
The image comes with "pre-packaged" `ENTRYPOINT` and `CMD`. 

On podman v2.1 (please disregard the naming of my binaries), it starts just fine. The version is
```sh
$ ./bin/podman-v2.1 version
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.13.8
Built:        Tue Dec 15 19:28:07 2020
OS/Arch:      linux/amd64
```
and, the container starting fine, I have nothing to tell here.


On podman v2.2, it constantly crashes (without any logs, which I would supply otherwise), because the new `play` disregards the `CMD` statement from the image, if nothing has been passed from the pod-spec:

On
```sh
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.13.8
Git Commit:   a0d478edea7f775b7ce32f8eb1a01e75374486cb
Built:        Tue Dec 15 19:24:20 2020
OS/Arch:      linux/amd64
```
 `podman inspect` yileds
```sh
$ ./bin/podman-v2.2 inspect try_pod-try \
| jq '.[] | {"args": .Args, "cmd": .Config.Cmd, "entrypoint": .Config.Entrypoint }'
{
  "args": [
    "docker-entrypoint.sh"
  ],
  "cmd": [
    "docker-entrypoint.sh"
  ],
  "entrypoint": "docker-entrypoint.sh"
}
```
That certainly can never start any redis.

So, I still do not understand if this behaviour is really intentional (please bear with me should I lack understanding). 

Is the new behaviour intentional and canonical? Then, I'd politely ask to mention a breaking change in the Changelogs, still politely, even if I needlessly spent hours chasing my crashing pods in my testing environment. In addition, it should then be documented that users MUST pass `command` and `args` in their pod-spec, even if they think both could be taken from the image. Futhermore, the comment referring to the kubernetes specification in the `generate` code should be removed, because the code the does not implement the k8s specs, which, for instance, demand that `ENTRYPOINT` and `CMD` are taken from the image when nothing has been passed.

If it is not intended to behave different than the k8s spec specifies, then I fail (perhaps I am thinking wrong here?) to understand why a _correct_ behaviour should be gated behind a flag, when the _incorrect_ behaviour is the default.

I hope I did not let emotions run here, I certainly do not want to offend.  I (still?) firmly believe the behaviour of v2.2 is buggy and I am ready to stand corrected should someone explain to me disregarding `CMD` from the image is correct.

In any case, thank you for your patience. (I am on CET, so I will not be active any more for my instance of today.)

mheon:
OK, this is not the impression I got from the initial PR - sorry. Yes, we did do a major rewrite of `play kube` logic for 2.2.x, so this could well be a regression fix. If it is, there is no need to add a flag.

If this is a regression, we need a test to catch it in the future; would you be willing to write one?

cruwe:
> If this is a regression, we need a test to catch it in the future; would you be willing to write one?

Hi, there already are tests, which do not catch at least one case described in the k8s spec. I'll give it a try and try to work in parallel to what and how @zhangguanzhang solved it in July.

Cheers!

zhangguanzhang:
try ro change https://github.com/containers/podman/blob/master/test/e2e/play_kube_test.go#L1248-L1267 to:
```golang
	It("podman play kube deployment more than 1 replica test correct command", func() {
		var i, numReplicas int32
		numReplicas = 5
		deployment := getDeployment(withReplicas(numReplicas))
		err := generateKubeYaml("deployment", deployment, kubeYaml)
		Expect(err).To(BeNil())

		kube := podmanTest.Podman([]string{"play", "kube", kubeYaml})
		kube.WaitWithDefaultTimeout()
		Expect(kube.ExitCode()).To(Equal(0))

		podNames := getPodNamesInDeployment(deployment)
		correctCmd := "[" + strings.Join(defaultCtrArg, " ") + "]"
		correctEntrypoint := defaultCtrCmd
		for i = 0; i < numReplicas; i++ {
			inspect := podmanTest.Podman([]string{"inspect", getCtrNameInPod(&podNames[i]), "--format", "'{{ .Config.Cmd }}'"})
			inspect.WaitWithDefaultTimeout()
			Expect(inspect.OutputToString()).To(Equal(correctCmd))
			inspect = podmanTest.Podman([]string{"inspect", getCtrNameInPod(&podNames[i]), "--format", "'{{ .Config.Entrypoint }}'"})
			inspect.WaitWithDefaultTimeout()
			Expect(inspect.OutputToString()).To(Equal(correctEntrypoint))
		}
	})
```

test/e2e/play_kube_test.go
```diff
@@ -803,76 +816,194 @@ var _ = Describe("Podman play kube", func() {
...
 		Expect(inspect.ExitCode()).To(Equal(0))
-		// Use the defined command to override the image's command
-		correctCmd := "[" + strings.Join(defaultCtrCmd, " ") + " " + strings.Join(defaultCtrArg, " ")
-		Expect(inspect.OutputToString()).To(ContainSubstring(correctCmd))
+
+		var containerState []ContainerState
+		err = json.Unmarshal([]byte(inspect.OutputToString()), &containerState)
+		if err != nil {
+			panic(err)
+		}
```
zhangguanzhang, line 843:
```suggestion
		Expect(err).To(BeNil())
```

test/e2e/play_kube_test.go
```diff
@@ -803,76 +816,194 @@ var _ = Describe("Podman play kube", func() {
...
 		Expect(inspect.ExitCode()).To(Equal(0))
-		// this image's ENTRYPOINT is called `docker-entrypoint.sh`
-		// so result should be `docker-entrypoint.sh + withArg(...)`
-		Expect(inspect.OutputToString()).To(ContainSubstring(`[docker-entrypoint.sh echo hello]`))
+
+		var containerState []ContainerState
+		err = json.Unmarshal([]byte(inspect.OutputToString()), &containerState)
+		if err != nil {
+			panic(err)
+		}
```
zhangguanzhang, line 923:
```suggestion
		Expect(err).To(BeNil())
```

zhangguanzhang:
and need rebase the commits to one commit

cruwe:
@zhangguanzhang : `Expect(err).To(BeNil())` is much nicer, thanks.

I changed the deployment test to test replica count. The idea is, that pod structure has already been tested at pod level, so at deployment level, I suggest to test what's new / different there and that's replica count.

cruwe (commented):
> This did cause the yaml I exported using the `podman generate kube` command to be unable to deploy using the `podman play kube` command.
@daiaji: Could you give an example? I'd like to have a look.

test/e2e/play_kube_test.go
```diff
@@ -803,76 +822,186 @@ var _ = Describe("Podman play kube", func() {
...
+		containerCmd := containerState[0].Config.Cmd
+		containerArgs := containerState[0].Args
+
+		var nilArray []string
+
+		Expect(containerEntrypoint).NotTo(Equal(imgEntrypoint))
+		Expect(containerCmd).NotTo(Equal(imgCmd))
+
+		Expect(containerEntrypoint).To(Equal(strings.Join(overrideCmd, " ")))
+		Expect(containerCmd).To(Equal(nilArray))
```
zhangguanzhang, line 899: a nit 
```suggestion
		Expect(len(containerCmd)).To(Equal(0))
```

pkg/specgen/generate/kube/kube.go
```diff
@@ -126,13 +125,29 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
...
+
+	if len(containerYAML.Command) == 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = imageData.Config.Entrypoint
+		s.Command = containerYAML.Args
+	}
+
+	if len(containerYAML.Command) != 0 && len(containerYAML.Args) != 0 {
+		s.Entrypoint = containerYAML.Command
+		s.Command = containerYAML.Args
+	}
```
bziemons, line 149: Hi! can we boil this down to a simple "set entrypoint/command from `imageData` and override from `containerYAML` when set" ?  I haven't seen this PR before I started fixing this and this would be my patch of this file: https://github.com/containers/podman/compare/master...bziemons:patch-kube-play-8428.patch

I also want to emphasize that the `if imageData != nil && imageData.Config != nil {` above is probably there for a reason and accessing `imageData.Config` without it is probably not a good idea.

zhangguanzhang: agree with @bziemons, line: 132,142 need the the imageData != nil && imageData.Config != nil

cruwe: > Hi! can we boil this down to a simple "set entrypoint/command from `imageData` and override from `containerYAML` when set" ? I haven't seen this PR before I started fixing this and this would be my patch of this file: https://github.com/containers/podman/compare/master...bziemons:patch-kube-play-8428.patch

Yes, we can do that. It is functionally equivalent and less code, but it lacks the clarity in mirroring the k8s specs that I wished to express. I feel that your proposition warrants some line of comments to establish that link, but apart from that, I wouldn't complain.

> I also want to emphasize that the `if imageData != nil && imageData.Config != nil {` above is probably there for a reason and accessing `imageData.Config` without it is probably not a good idea.

I fear I lost the context here. Are you referring to L115 before setting `s.WorkDir = imageData.Config.WorkingDir`? I did not notice that, but I supsect it is not necessary, because `imageData` is populated with `imageData, err := newImage.Inspect(ctx)` at L110. Any failure to get the `imageData` returns there.

In addition, I cannot imagine how to start any container without having both `imageData` and `imageData.Config`. Accordingly, the corresponding check, if it really is necessary, is misplaced when setting single values and should be performed at the "global" level somewhar at `libpod/image/image.go`, `Inspect` or when examining the return value  of it's invocation at L110.

cruwe: >> Hi! can we boil this down to a simple "set entrypoint/command from `imageData` and override from `containerYAML` when set" ? I haven't seen this PR before I started fixing this and this would be my patch of this file: https://github.com/containers/podman/compare/master...bziemons:patch-kube-play-8428.patch

> I also want to emphasize that the `if imageData != nil && imageData.Config != nil {` above is probably there for a reason and accessing `imageData.Config` without it is probably not a good idea.

Ok, now I get it. That part, if it is really necessary, about which I am yet not sure, should error much higher in the chain.

bziemons: I agree with you and the if should have gotten a comment from its writer about why it is needed. Even if it isn't needed it wouldn't make sense to make the code more confusing about it (I know that wasn't your intention ;) ). Thanks for accepting my feedback though

haircommander:
alternative fix is here https://github.com/containers/podman/pull/8807

cruwe:
Fine. I am really disappointed how this got handled.

rhatdan:
@cruwe Thanks for your PR. I am sorry about the conflicting PRs.

PR #8669, opened 2020-12-09
security: honor systempaths=unconfined for ro paths

giuseppe:
we must honor systempaths=unconfined also for read-only paths, as
Docker does:

proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0


rhatdan:
LGTM
@umohnani8 PTAL

umohnani8:
LGTM

PR #8671, opened 2020-12-09
RHEL: Do not pull rootless CNI infra image, and instead request that user build it.

mheon:
Instead, we want to advise users to manually build the image. We cannot distribute the existing image for RHEL 8.3.1, and the feature will be tech preview, so this degraded user experience will have to be sufficient until we can get a better solution in place.
    
Ref: https://issues.redhat.com/browse/RUN-1127
    
Please note that this is a RHEL only change and should not be included in non-RHEL branches.

Also, cherry-pick #8652 into the RHEL tree.


mheon:
@baude @TomSweeneyRedHat @rhatdan PTAL

rhatdan:
@mheon Looks like you have a PR in here that is not supposed to be here on Ports.

mheon:
@rhatdan Deliberate cherry-pick - that's a pretty significant bugfix we should get into RHEL. We're not frozen yet, so I grabbed it.

rhatdan:
LGTM

libpod/rootless_cni_linux.go
```diff
@@ -233,14 +225,13 @@ func ensureRootlessCNIInfraContainerRunning(ctx context.Context, r *Runtime) (*C
...
-		return nil, errors.Errorf("cannot find rootless-podman-network-sandbox image for %s", runtime.GOARCH)
-	}
+	imageName := "rootless-cni-infra"
 	logrus.Debugf("rootless CNI: ensuring image %q to exist", imageName)
-	newImage, err := r.ImageRuntime().New(ctx, imageName, "", "", nil, nil,
-		image.SigningOptions{}, nil, util.PullImageMissing)
+	newImage, err := r.ImageRuntime().NewFromLocal(imageName)
 	if err != nil {
+		if errors.Cause(err) == define.ErrNoSuchImage {
+			return nil, errors.Errorf("rootless CNI infra image not present - please build image from https://github.com/containers/podman/blob/v2.2.1/contrib/rootless-cni-infra/Containerfile and tag as %q", imageName)
```
TomSweeneyRedHat, line 233: I know it's long already, but could we give them the command to use to build the image instead of just the Containerfile?

mheon: I don't think so - the actual build process is getting very complicated because we can't just do a `podman build $URL` (you need to pull down not just the containerfile, but the directory containing it, and then do a `podman build` on that)

mheon: We have a separate card to document this elsewhere, which is where I think that sort of documentation will need to land.

vrothberg:
@AkihiroSuda PTAL.

At some point we'd love to ship multi-arch images but the local build seems like the best way forward for now.

vrothberg:
> @AkihiroSuda PTAL.
> 
> At some point we'd love to ship multi-arch images but the local build seems like the best way forward for now.

Ah, that's a RHEL-only branch

test/e2e/network_create_test.go
```diff
@@ -74,6 +74,7 @@ var _ = Describe("Podman network create", func() {
 	)
 
 	BeforeEach(func() {
+		SkipIfRootless("rootless CNI is tech preview in 8.3.1")
```
AkihiroSuda, line 77: 8.3.1 -> RHEL 8.3.1

AkihiroSuda:
> Ref: https://issues.redhat.com/browse/RUN-1127

I don't have visibility in this URL, but I'm wondering Red Hat could build and upload the infra image by themselves if they have concern in the image built outside Red Hat.

libpod/rootless_cni_linux.go
AkihiroSuda: What about letting Podman itself execute `git clone ...  && git checkout ... && podman build -t rootless-cni-infra .`

vrothberg:
> > Ref: https://issues.redhat.com/browse/RUN-1127
> 
> I don't have visibility in this URL, but I'm wondering Red Hat could build and upload the infra image by themselves if they have concern in the image built outside Red Hat.

I think that's the long-term goal. Thanks for checking!

libpod/rootless_cni_linux.go
mheon: Problem is disconnected environments - we can't guarantee the system has a working connection to the internet.

mheon:
Re-pushed. Added some extra docs to the README on Github about building, which should help people get started.

@rhatdan @TomSweeneyRedHat PTAL

mheon:
Good to merge

vrothberg (commented):
LGTM

mheon:

AkihiroSuda (approved):

PR #8681, opened 2020-12-10
Fix panic in libpod images exists endpoint

Luap99:
The libpod images exists endpoint panics when called with
a non existing image and therefore returns 500 as status
code instead of the expected 404.

A test is added to ensure it is working.


baude:
LGTM

mheon:
LGTM

PR #8685, opened 2020-12-10
Ignore containers.conf sysctls when sharing namespaces

mheon:
The existing code prevents containers.conf default sysctls from being added if the container uses a host namespace. This patch expands that to not just host namespaces, but also *shared* namespaces - so we never modify another container's (or a pod's) namespaces without being explicitly directed to do so by the user.

### DIFF
--- a/pkg/specgen/generate/security.go
+++ b/pkg/specgen/generate/security.go
@@ -178,6 +178,10 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 
 	g.SetRootReadonly(s.ReadOnlyFilesystem)
 
+	noUseIPC := s.IpcNS.NSMode == specgen.FromContainer || s.IpcNS.NSMode == specgen.FromPod || s.IpcNS.NSMode == specgen.Host
+	noUseNet := s.NetNS.NSMode == specgen.FromContainer || s.NetNS.NSMode == specgen.FromPod || s.NetNS.NSMode == specgen.Host
+	noUseUTS := s.UtsNS.NSMode == specgen.FromContainer || s.UtsNS.NSMode == specgen.FromPod || s.UtsNS.NSMode == specgen.Host
+
 	// Add default sysctls
 	defaultSysctls, err := util.ValidateSysctls(rtc.Sysctls())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/specgen/generate/security.go
```diff
@@ -178,6 +178,10 @@ func securityConfigureGenerator(s *specgen.SpecGenerator, g *generate.Generator,
 
 	g.SetRootReadonly(s.ReadOnlyFilesystem)
 
+	noUseIPC := s.IpcNS.NSMode == specgen.FromContainer || s.IpcNS.NSMode == specgen.FromPod || s.IpcNS.NSMode == specgen.Host
```
rhatdan, line 181: Why are you tying IPC to Network namespace?

mheon: I don't think I am? These are all `s.IpcNS`? We're checking for `--ipc=container:` and pod IPC.

rhatdan: Ok I read it wrong.

rhatdan:
Looks like the tests is failing for real.

mheon:
Concur. It looks like the pod isn't adding the default containers.conf sysctls... It seems like it's not a bad idea to add them, though.

rhatdan:
They should be added when the pod/infra container is created

cevich:
***Note***: Now I'm seeing this [pass for F32 host](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6075875785441280/html/int-podman-fedora-32-root-host.log.html#t--podman-stats-on-net-host-post--1), but [fail for F32 remote](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5724032064552960/html/int-remote-fedora-32-root-host.log.html#t--podman-stats-on-net-host-post--1).

mheon:
Pushed a fresh patch that will have pod infra containers add the relevant sysctls.

mheon:
Needs a merge. @containers/podman-maintainers PTAL

vrothberg (commented):
LGTM

cevich:
Maybe I missed it, but shouldn't the disabled tests also be re-enabled?

Ref: https://github.com/cevich/podman/commit/e6fbc15f26b2a609936dfc11732037c70ee14cba

mheon:
I wasn't aware that there were tests disabled - must have happened while I was on PTO?

cevich:
Correct, I disabled them so we could get my "new images" PR merged before this fix went in.

cevich:
Hmmm...looking...actually nvm, on master the only thing we have in this test is `SkipIfRootlessCgroupsV1()` which is probably fine.

Issue #7746, opened 2020-09-23
[flake] remote,rootless: create+start+ps: container does not exist in database


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ P() { ./bin/podman-remote "$@"; }
$ P create --name test1 --rm alpine foo;P start test1;P ps -aq
a7a5315b7a4e6d9b40e2955774197644fb771730721b4a9be1ac4128d47c242d
Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
Error: container a7a5315b7a4e6d9b40e2955774197644fb771730721b4a9be1ac4128d47c242d does not exist in database: no such container
```
It's a flake (race condition): don't expect it to happen every time. For me it's one out of 3-5 runs on my laptop.

Rootless only, AFAICT: I could not get it to reproduce with `P() { sudo ...}`. 

master @ 5cedd830f7275e8dc3382502908b846bfa57a3b8 on f32.

While I'm at it, the error message is different between `podman` and `podman-remote`, does this matter?
```
podman: Error: unable to start container "911e628f4232690efc7a7efad770491312c94a4a4f8fb164feb53a1b41ef4a8b": executable file not found in $PATH: No such file or directory: OCI runtime command not found error
remote: Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
```

Only seems to happen when creating an error container (`foo`, i.e. command not found). Does not happen with `true` ... except, now, as root!
```console
$ P() { sudo ./bin/podman-remote "$@"; }
$ P create --name test1 --rm alpine true;P start test1;P ps -aq
3c9be0ca2306b17a63040c24a4dbea85a283fa4ff080fc3ad5ce5d1852158d2c
test1
Error: container 3c9be0ca2306b17a63040c24a4dbea85a283fa4ff080fc3ad5ce5d1852158d2c does not exist in database: no such container
```
(very very infrequently; it takes a lot of repetitions to trigger)

rhatdan:
@edsantiago Any update on this issue?

edsantiago:
Issue is still present; failing about one in ten using reproducer in comment 0 (in a `while` loop):
```
fe1e51c49e5839fa85e999dd7fdae8b1ff96c139b827e4a530171b592baf86a4
[...]

PR #8688, opened 2020-12-10
Handle --rm when starting a container

rhatdan:
podman start should follow the same behaviour as podman run when removing a
container.

Fixes: https://github.com/containers/podman/issues/7746


rhatdan:
@edsantiago @mheon I am not sure this is the correct behaviour.

IE Should failing to start, remove an autoremove container?

# podman run -rm fedora foo
#  podman create --name test1 --rm alpine foo;podman start test1

Both remove it, so for consistancy

 podman-run create --name test1 --rm alpine foo;podman-run start test1

Will also.

edsantiago:
@rhatdan I think we should probably preserve the behavior documented in `podman-create(1)`:

>Note that the container will not be removed when it could not be created or started successfully. This allows the user to inspect the container after failure.

edsantiago:
Ugh. Thinking it over some more, I'm uncertain what "started successfully" means. I could argue that getting to the point where the container says "dude, no such command" _is_ a successful execution of the container. 

This is too tricky for me to think about at the end of a long week...

mheon:
From everything we've done so far, I think the expectation is that autoremove always removes, even containers that error. Recommendation: we should retain existing behavior and just remove that bit from the manpage.

rhatdan:
@mheon @edsantiago Tests are passing, we might as well work the same way in podman as podman-remote and remove the containers when start completes, whether it succeeds or fails.

edsantiago:
SGTM but I agree with @mheon: those two sentences need to be removed from the manpage.

vrothberg (commented):
Code LGTM but I'd love a test or two.

TomSweeneyRedHat:
LGTM
and happy green test buttons

PR #8689, opened 2020-12-11
Add volume prune --filter support

### LINKED ISSUE #8672: podman volume prune --filter missing
Docker has a --filter for podman volume prune.  We should match.  

Once this is added we should also pass the --filter from podan system --prune down to the volume prune there.

bblenard:
This change adds support for the `--filter` / `?filters` arguments on
the `podman volume prune` subcommand.

  * Adds ParseFilterArgumentsIntoFilters helper for consistent
    Filter string slice handling
  * Adds `--filter` support to podman volume prune cli
  * Adds `?filters...` support to podman volume prune api
  * Updates apiv2 / e2e tests

Closes #8672

Signed-off-by: Baron Lenardson <lenardson.baron@gmail.com>

### DIFF
--- a/docs/source/markdown/podman-volume-prune.1.md
+++ b/docs/source/markdown/podman-volume-prune.1.md
@@ -8,7 +8,8 @@ podman\-volume\-prune - Remove all unused volumes
 
 ## DESCRIPTION
 
-Removes all unused volumes. You will be prompted to confirm the removal of all the
+Removes unused volumes. By default all unused volumes will be removed, the **--filter** flag can
+be used to filter specific volumes. You will be prompted to confirm the removal of all the
 unused volumes. To bypass the confirmation, use the **--force** flag.
 
 
@@ -17,6 +18,17 @@ unused volumes. To bypass the confirmation, use the **--force** flag.
 #### **--force**, **-f**
 
 Do not prompt for confirmation.
+
+#### **--filter**
+
+Filter volumes to be pruned. Volumes can be filtered by the following attributes:
+
+- dangling
+- driver
+- label
+- name
+- opt
+- scope
 
 #### **--help**
 
@@ -29,6 +41,8 @@ Print usage statement
 $ podman volume prune
 
 $ podman volume prune --force
+
+$ podman volume prune --filter label=mylabel=mylabelvalue
 ```
 
 ## SEE ALSO
--- a/pkg/bindings/volumes/volumes.go
+++ b/pkg/bindings/volumes/volumes.go
@@ -75,7 +75,7 @@ func List(ctx context.Context, filters map[string][]string) ([]*entities.VolumeL
 }
 
 // Prune removes unused volumes from the local filesystem.
-func Prune(ctx context.Context) ([]*entities.VolumePruneReport, error) {
+func Prune(ctx context.Context, filters map[string][]string) ([]*entities.VolumePruneReport, error) {
 	var (
 		pruned []*entities.VolumePruneReport
 	)
@@ -83,7 +83,15 @@ func Prune(ctx context.Context) ([]*entities.VolumePruneReport, error) {
 	if err != nil {
 		return nil, err
 	}
-	response, err := conn.DoRequest(nil, http.MethodPost, "/volumes/prune", nil, nil)
+	params := url.Values{}
+	if len(filters) > 0 {
+		strFilters, err := bindings.FiltersToString(filters)
+		if err != nil {
+			return nil, err
+		}
+		params.Set("filters", strFilters)
+	}
+	response, err := conn.DoRequest(nil, http.MethodPost, "/volumes/prune", params, nil)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/domain/entities/volumes.go
+++ b/pkg/domain/entities/volumes.go
@@ -1,6 +1,7 @@
 package entities
 
 import (
+	"net/url"
 	"time"
 
 	docker_api_types "github.com/docker/docker/api/types"
@@ -107,6 +108,12 @@ type VolumeRmReport struct {
 
 type VolumeInspectReport struct {
 	*VolumeConfigResponse
+}
+
+// VolumePruneOptions describes the options needed
+// to prune a volume from the CLI
+type VolumePruneOptions struct {
+	Filters url.Values `json:"filters" schema:"filters"`
 }
 
 type VolumePruneReport struct {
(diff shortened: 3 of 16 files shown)


pkg/domain/entities/volumes.go
```diff
@@ -109,6 +110,12 @@ type VolumeInspectReport struct {
 	*VolumeConfigResponse
 }
 
+// VolumePruneOptions describes the options needed
+// to prune a volume from the CLI
+type VolumePruneOptions struct {
+	Filters url.Values `json:"filters" schema:"filters"`
```
bblenard, line 116: I was conflicted whether to use `url.Values` or the underlying type of `map[string][]string` for the filters. I saw it both ways throughout the codebase. I settled on `url.Values` but I don't feel strongly either way. I thought I remember seeing some mention of a refactor for filters so hopefully the inconsistency can be fixed then.

Luap99: I feel like we should use `map[string][]string` everywhere.
@baude @jwhonce WDYT?

Luap99 (commented):
Thanks for the contribution @bblenard.

This looks really good.
You still have to document the `--filter` flag in `docs/source/markdown/podman-volume-prune.1.md` and fix the bindings test in `pkg/bindings/test/volumes_test.go`.

pkg/bindings/volumes/volumes.go
```diff
@@ -75,15 +75,15 @@ func List(ctx context.Context, filters map[string][]string) ([]*entities.VolumeL
...
+func Prune(ctx context.Context, opts entities.VolumePruneOptions) ([]*entities.VolumePruneReport, error) {
 	var (
 		pruned []*entities.VolumePruneReport
 	)
 	conn, err := bindings.GetClient(ctx)
 	if err != nil {
 		return nil, err
 	}
-	response, err := conn.DoRequest(nil, http.MethodPost, "/volumes/prune", nil, nil)
+	response, err := conn.DoRequest(nil, http.MethodPost, "/volumes/prune", opts.Filters, nil)
```
Luap99, line 86: This is not correct. You have to add the filters to a new params map. Otherwise you will lose the `filters` key
```
params := url.Values{}
if opts.Filters != nil {
	filterString, err := bindings.FiltersToString(opts.Filters)
	if err != nil {
		return nil, err
	}
	params.Set("filters", filterString)
}
response, err := conn.DoRequest(nil, http.MethodPost, "/volumes/prune", params, nil)
```

pkg/domain/entities/volumes.go
bblenard: Once @baude / @jwhonce add their input I can fix the other things you mentioned @Luap99 (since the type will effect the changes you requested, even though its not much different).

pkg/bindings/volumes/volumes.go
bblenard: :+1: Thanks for the feedback. How would I have found this error on my own? I tried to update the tests in `test/apiv2/30-volumes.at` and they were passing. Those tests had the params map with filter key encoded in the url manually so I'm guessing thats why they worked. Was there another test(s) I could have ran on my dev box? I ran the following things to test on my machine

1. Manual `bin/podman` commands to test functionality
2. `sudo ./test-apiv2 30-volumes`
3. `go test -v test/e2e/config.go test/e2e/config_amd64.go test/e2e/common_test.go test/e2e/libpod_suite_test.go test/e2e/volume_prune_test.go`
4. `make lint` in the gate container

(you can see my test output in the [issue](https://github.com/containers/podman/issues/8672) for this PR)

bblenard:
> Thanks for the contribution @bblenard.
> 
> This looks really good.
> You still have to document the `--filter` flag in `docs/source/markdown/podman-volume-prune.1.md` and fix the bindings test in `pkg/bindings/test/volumes_test.go`.

Documented the filter flag here: https://github.com/containers/podman/pull/8689/commits/41861e39d53ef9ac388f6440f34073ef07386034

pkg/bindings/volumes/volumes.go
Luap99: Our CI would have catched  this. The test in `test/e2e/volume_prune_test.go` has to be run twice, for `podman` and `podman-remote`. This code path is only taken with `podman-remote`.

You can run the remote integration tests with `make remoteintegration` but make sure to add `FIt` instead of `It` in front of  your test otherwise the full test suite will be run but don't forget to remove the `FIt` before you commit your changes.

pkg/domain/entities/volumes.go
Luap99: Just keep it for now. It will be reworked soon.

docs/source/markdown/podman-volume-prune.1.md
```diff
@@ -18,6 +19,17 @@ unused volumes. To bypass the confirmation, use the **--force** flag.
...
+#### **--filter**
+
+Filter volumes to be pruned. Volumes can be filtered by the following attributes:
+
+- name
+- driver
+- scope
+- label
+- opt
+- dangling
```
TomSweeneyRedHat, line 31: Alphabetize this list please

pkg/bindings/volumes/volumes.go
bblenard: Fixed: https://github.com/containers/podman/pull/8689/commits/0a39fdaa2abf0c5905d6665f42ce74ffe4401093

docs/source/markdown/podman-volume-prune.1.md
bblenard: Fixed: https://github.com/containers/podman/pull/8689/commits/c8a7b9a4ac1aff5e479dd7e333f5aea2bd03dff4

bblenard:
Should I squash my feedback commits so the CI pipeline goes? (I think that is the issue)

mheon:
Yes, please

Luap99 (approved):
LGTM

pkg/domain/infra/abi/system.go
```diff
@@ -214,7 +214,7 @@ func (ic *ContainerEngine) SystemPrune(ctx context.Context, options entities.Sys
 			systemPruneReport.ImagePruneReport.Report.Id = append(systemPruneReport.ImagePruneReport.Report.Id, results...)
 		}
 		if options.Volume {
-			volumePruneReport, err := ic.pruneVolumesHelper(ctx)
+			volumePruneReport, err := ic.pruneVolumesHelper(ctx, nil)
```
rhatdan, line 217: This should be passing down the prune filters.

podman system prune --filter should be passed down to volume helper as well.

bblenard: Yes, that was my mistake. I can open a PR for that tonight.

rhatdan: Thanks, 
BTW great job on this PR.

bblenard: @rhatdan PR opened: https://github.com/containers/podman/pull/8724

Issue #8547, opened 2020-12-01
Wrong image tag is used when creating a container from an image with multiple tags


riyad:
/kind bug

**Description**

Trying to find regressions from Podman 2.1.1 by exercising the APIv2 trough [docker-py](https://github.com/docker/docker-py)'s test suite (see #5386) I came across a regressions in the [`ListContainersTest.test_list_containers`](https://github.com/docker/docker-py/blob/master/tests/integration/api_container_test.py#L25) test.

When listing containers (i.e. `GET /v1.40/containers/json?all=1`) the `"Image"` field has values like `"docker.io/library/sha256:be4e4bea2c2e15b403bb321562e78ea84b501fb41497472e91ecb41504e8a27c" which throws off any code that expects it to be the same as what the container was created with (here: _"alpine:3.10"_).

What "breaks" the listing containers test is another test called [`ImageCollectionTest.test_save_and_load`](https://github.com/docker/docker-py/blob/1757c974fa3a05b0e9b783af85242b18df09d05d/tests/integration/models_images_test.py#L100) which saves an image (also `"alpine:3.10"` :raised_eyebrow: ) into a file and loads it into the local image repository again.

NOTE: when loading the image Podman adds an extra tag in the form of `"docker.io/library/sha256:be4e4bea2c2e15b403bb321562e78ea84b501fb41497472e91ecb41504e8a27c"` to the image where the SHA corresponds to the actual image id. Now the image has two tags! It seems Docker doesn't add the extra tag.

When we then run the original test again (i.e. [`ListContainersTest.test_list_containers`](https://github.com/docker/docker-py/tree/master/tests/integration/api_container_test.py)) it's supposed to create a container from the _"alpine:3.10"_ image. Here's where the weird thing happens: the container's the image is saved (into the BoltDB) as `"docker.io/library/sha256:be4e4bea2c2e15b403bb321562e78ea84b501fb41497472e91ecb41504e8a27c"` (same image, but wrong tag).

**Steps to reproduce the issue:**

The following (trans-)script shows what happens. It can be run in the Python interpreter (assuming docker-py is installed).

```python
import docker
import tempfile


TEST_IMG = 'alpine:3.10'


client = docker.from_env()

# POST http+docker://localhost/v1.40/images/create
image = client.images.pull(TEST_IMG)

#
# adapted from ImageCollectionTest.test_save_and_load (can be run against Podman 2.1.1 or 2.2-dev)
#

[...]

PR #8690, opened 2020-12-11
Fix Wrong img name used when creating a container from an image which had multi names

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/8547


mheon:
LGTM

PR #8691, opened 2020-12-11
Fix some network compat api problems

Luap99:
Network create could panic when used with a json body like this:
`{"Name":"net","IPAM":{"Config":[]}}`

The network scope for list and inspect should not be empty. It can
be swarm, global or local. We only support local networks so
hardcode this field to local.


mheon:
LGTM

Issue #3759, opened 2019-08-08
Rebooted, dangling file in /var/lib/cni/networks/podman prevents container starting 


space88man:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I did a `systemctl reboot`; encountered a orphan `/var/lib/cni/networks/podman/10.88.0.20` file
which prevented the container from starting.
```
ERRO[0000] Error adding network: failed to allocate for range 0: requested IP address 10.88.0.20 is not available in range set 10.88.0.1-10.88.255.254 
ERRO[0000] Error while adding pod to CNI network "podman": failed to allocate for range 0: requested IP address 10.88.0.20 is not available in range set 10.88.0.1-10.88.255.254 
Error: unable to start container "freeswitch-init_1": error configuring network namespace for container 7690a7bc4960b76799d906dbc07a32d174a5e75581886349d3604d455e093cf7: failed to allocate for range 0: requested IP address 10.88.0.20 is not available in range set 10.88.0.1-10.88.255.254
```

**Steps to reproduce the issue:**

1. Upgraded Fedora 30 host, rebooted, tried to start a container

2.

3.

**Describe the results you received:**
```
INFO[0000] Found CNI network podman (type=bridge) at /etc/cni/net.d/87-podman-bridge.conflist 
DEBU[0000] Made network namespace at /var/run/netns/cni-76dce0ef-599f-b625-7710-a7eeef4159e5 for container 7690a7bc4960b76799d906dbc07a32d174a5e75581886349d3604d455e093cf7 
INFO[0000] Got pod network &{Name:freeswitch-init_1 Namespace:freeswitch-init_1 ID:7690a7bc4960b76799d906dbc07a32d174a5e75581886349d3604d455e093cf7 NetNS:/var/run/netns/cni-76dce0ef-599f-b625-7710-a7eeef4159e5 PortMappings:[] Networks:[podman] NetworkConfig:map[podman:{IP:10.88.0.20}]} 
INFO[0000] About to add CNI network cni-loopback (type=loopback) 
[...]

PR #8692, opened 2020-12-11
[CI:DOCS] Cleanup CNI Networks on reboot

rhatdan:
CNI sometimes leaves Network information in /var/lib/cni/networks
when the system crashes or containers do not shut down properly.

This PR will cleanup these left over files, so that container engines
will get a clean enviroment when the system reboots.

Related to: https://github.com/containers/podman/issues/3759

### DIFF
--- a/contrib/tmpfile/podman.conf
+++ b/contrib/tmpfile/podman.conf
@@ -1,4 +1,5 @@
 # /tmp/podman-run-* directory can contain content for Podman containers that have run
 # for many days. This following line prevents systemd from removing this content.
-x /tmp/podman-run-.*
-d /run/podman 0700 root root
+x /tmp/podman-run-*
+D! /run/podman 0700 root root
+D! /var/lib/cni/networks


rhatdan:
This is really not a DOCS only change, but the CI system does not test this functionality.  Machines require a reboot to make sure this works properly.

Luap99:
We try to remove these files already, no?
 https://github.com/containers/podman/blob/dd954781e6e308a0bbecfaf6699b41426100a58d/libpod/container_internal.go#L604-L613
https://github.com/containers/podman/blob/dd954781e6e308a0bbecfaf6699b41426100a58d/libpod/container_internal.go#L631

mheon:
I don't think existing code is working. I took a quick look through my `/var/lib/cni/networks/podman` yesterday and had a few hundred orphaned IP address reservation files. That's probably uncommonly high because I'm a developer and actively recreating bugs and such, but it still speaks to our deletion efforts on that front being less than successful.

rhatdan:
@Luap99 @mheon @containers/podman-maintainers PTAL
Can we get this merged.  I think this is safe whether or not the other code is correct.  Bottom line is these files can be left around if the machine suddenly crashes, and we want them cleaned up.

Luap99 (commented):
LGTM

PR #8693, opened 2020-12-11
podman: drop checking valid rootless UID

giuseppe:
do not check whether the specified ID is valid in the user namespace.

crun handles this case[1], so the check in Podman prevents to get to
the OCI runtime at all.

$ podman run --user 10:0 --uidmap 0:0:1 --rm -ti fedora:33 sh -c 'id; cat /proc/self/uid_map'
uid=10(10) gid=0(root) groups=0(root),65534(nobody)
        10          0          1

[1] https://github.com/containers/crun/pull/556

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -3,7 +3,6 @@ package containers
 import (
 	"fmt"
 	"os"
-	"strconv"
 	"strings"
 
 	"github.com/containers/common/pkg/completion"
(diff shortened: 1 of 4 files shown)


giuseppe:
/hold

let's hold it at least until https://github.com/containers/crun/pull/556 is merged

rhatdan:
What happens if you use runc?

rhatdan:
crun is merged, this LGTM, but am concerned abour runc, do we need documentation on this feature only being available to crun?

rhatdan:
Great.

umohnani8:
LGTM

PR #8696, opened 2020-12-11
podman events allow future time for --until

### LINKED ISSUE #8694: podman uses wrong event log file path on the first run with new storage location

/kind bug

**Description**

Podman uses the wrong file path for the event log file when run in a new storage location but only for the first call after that the correct location is used.

window 1:
1. `podman --events-backend file events`

window 2:

2. `tmp=$(mktemp -d)`

3. `podman --events-backend file --tmpdir $tmp --runroot $tmp/runroot --root $tmp/root create --name t1 alpine`

4. `podman --events-backend file --tmpdir $tmp --runroot $tmp/runroot --root $tmp/root create --name t2 alpine`

5. `podman --events-backend file --tmpdir $tmp --runroot $tmp/runroot --root $tmp/root events --since 1h`

First event is logged in the standard event log file path. Second event is logged in the new tmp path location.

Both events should be logged in the new location.

master @ dd954781e6e308a0bbecfaf6699b41426100a58d

Luap99:
The podman events aren't read until the given timestamp if the
timestamp is in the future. It just reads all events until now
and exits afterwards.
This does not make sense and does not match docker. The correct
behavior is to read all events until the given time is reached.

This fixes a bug where the wrong event log file path was used
when running first time with a new storage location.
Fixes #8694

This also fixes the events api endpoint which only exited when
an error occurred. Otherwise it just hung after reading all events.

### DIFF
--- a/libpod/events/events.go
+++ b/libpod/events/events.go
@@ -216,8 +216,5 @@ func (e EventLogFile) getTail(options ReadOptions) (*tail.Tail, error) {
 		reopen = false
 	}
 	stream := options.Stream
-	if len(options.Until) > 0 {
-		stream = false
-	}
 	return tail.TailFile(e.options.LogFilePath, tail.Config{ReOpen: reopen, Follow: stream, Location: &seek, Logger: tail.DiscardingLogger, Poll: true})
 }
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM
@baude @mheon PTAL

umohnani8:
LGTM

Issue #8305, opened 2020-11-11
Implement "scan" command to surface known package vulnerabilities


wagoodman:
**Is this a BUG REPORT or FEATURE REQUEST?**

/kind feature

**Description**
It would be ideal to have a `scan` command that surfaces known vulnerabilities for packages within container images, mimicking the new [`docker scan` command](https://docs.docker.com/engine/scan/). 

I work on a CLI tool, [Grype](https://github.com/anchore/grype), that is written in go that can fulfill this functionality. Today grype does not depend on shelling out to get information for packages and vulnerabilities --all functionality is contained within the grype binary, with all dependencies 100% in go (no CGO required). Between this and the fact that grype was meant from the beginning to be used as both a CLI tool and a library makes it an ideal candidate for incorporating vulnerability detection functionality for a `scan` command.

There seem to be two ways to approach integrating grype and podman to provide the `podman scan` functionality:

1. **Pull in grype as an in-source-tree go dependency**, embedding grype into the podman binary. I have a [branch prototyping out the necessary changes](https://github.com/containers/podman/compare/master...wagoodman:add-scan-cmd-with-grype-intree).

2. **Manage the grype binary separately**. In this approach podman would download and manage the grype binary in an internal podman directory and invoke grype by shelling out to the binary and obtaining the results from `json` output.

In both approaches:
- the `ImageEngine.Save` functionality would be used within the `scan` command to obtain an image archive and feed into grype.
- the vulnerability database (a sqlite file) would be managed by grype in an internal podman directory, updated daily from publicly available data derived from https://ancho.re/v1/service/feeds .

I would be happy to collaborate with folks here on the desired approach and make contributions to grype and podman as-needed!

vrothberg:
Thanks for opening the issue!

We recently added a new image mount type which was motivated by your use case (i.e., image scanning). Doing a `podman run --mount type=image,fedora,/fedora` will mount the "fedora" image at `/fedora` inside the container.

So you can use the mechanism to use pretty much any scanner on any image. I have not had a look at `docker scan` yet but it sounds like something opinionated with a baked in scanner. There are many scanners out there in the wild.

In other words. If you have scanner and want to run it, you can put it into an image, and use it to scan other images.

vrothberg:
@rhatdan PTAL

rhatdan:
I like the idea of having a podman scan command to match Docker, but I would not want it to be opinionated.  
@baude @mheon Thoughts?

mheon:
[...]

PR #8706, opened 2020-12-13
[WIP] add image scan

wagoodman:
This is a draft implementation for issue #8305 (implement `podman scan` command).

### DIFF
--- a/cmd/podman/images/scan.go
+++ b/cmd/podman/images/scan.go
@@ -1,0 +1,156 @@
+package images
+
+import (
+	"context"
+	"fmt"
+	"github.com/containers/podman/v2/cmd/podman/common"
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/specgen"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"os"
+	"strings"
+)
+
+var (
+	scanDescription = `Scan a container image for known vulnerabilities.  The image name or digest can be used.`
+	scanCommand     = &cobra.Command{
+		Use:     "scan [run-options] IMAGE [scanner-options]",
+		Short:   "Scan a container image for known vulnerabilities",
+		Long:    scanDescription,
+		RunE:    scan,
+		Example: `podman scan centos:latest`,
+	}
+
+	imageScanCommand     = &cobra.Command{
+		Use:     "scan [run-options] IMAGE [scanner-options]",
+		Short:   "Scan a container image for known vulnerabilities",
+		Long:    scanDescription,
+		RunE:    scan,
+		Example: `podman image scan centos:latest`,
+	}
+)
+
+var (
+	scanOptions = entities.ImageScanOptions{
+		ContainerRunOptions: entities.ContainerRunOptions{
+			OutputStream: os.Stdout,
+			ErrorStream:  os.Stderr,
+			Rm: true,
+		},
+	}
+
+	cliVals common.ContainerCLIOpts
+)
+
+func scanFlags(cmd *cobra.Command) {
+	flags := cmd.Flags()
+
+	common.DefineCreateFlags(cmd, &cliVals)
+	common.DefineNetFlags(cmd)
+
+	// TODO: derive default from containers.conf configuration
+	scannerFlagName := "scanner"
+	flags.StringVar(&scanOptions.ScannerImage, scannerFlagName, "docker.io/library/grype:latest", "The vulnerability scanner container image to use")
+}
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: scanCommand,
+	})
+	scanFlags(scanCommand)
+
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: imageScanCommand,
+		Parent:  imageCmd,
+	})
+	scanFlags(imageScanCommand)
+
+}
+
+func scan(cmd *cobra.Command, args []string) error {
+	var err error
+
+	if len(args) < 1 {
+		return fmt.Errorf("requires exactly one image, given %+v", args)
+	}
+	var imageToScan = args[0]
+	var mountTarget = "/podman-scan-image-mount"
+	var scanMount = fmt.Sprintf("type=image,src=%s,target=%s", imageToScan, mountTarget)
+
+	cliVals.Mount = append(cliVals.Mount, scanMount)
+
+	// allow parsing to continue with the first arg being the scanner image (not the image to scan)
+	args[0] = scanOptions.ScannerImage
+
+	cliVals.Net, err = common.NetFlagsToNetOptions(cmd)
+	if err != nil {
+		return err
+	}
+
+	if af := cliVals.Authfile; len(af) > 0 {
+		if _, err := os.Stat(af); err != nil {
+			return err
+		}
+	}
+
+	s := specgen.NewSpecGenerator(scanOptions.ScannerImage, false)
+	if err := common.FillOutSpecGen(s, &cliVals, args); err != nil {
+		return err
+	}
+	s.RawImageName = scanOptions.ScannerImage
+	scanOptions.Spec = s
+
+	// TODO: support templating in config
+	// now force the scanner to use the mount as the first argument
+	s.Command = append([]string{"dir:"+mountTarget}, s.Command...)
+
+	if _, err := createPodIfNecessary(s, cliVals.Net); err != nil {
+		return err
+	}
+
+	report, err := registry.ContainerEngine().ContainerRun(registry.GetContext(), scanOptions.ContainerRunOptions)
+	// report.ExitCode is set by ContainerRun even it it returns an error
+	if report != nil {
+		registry.SetExitCode(report.ExitCode)
+	}
+	if err != nil {
+		return err
+	}
+
+	if scanOptions.Detach {
+		fmt.Println(report.Id)
+		return nil
+	}
+	return nil
+
+}
+
+// createPodIfNecessary automatically creates a pod when requested.  if the pod name
+// has the form new:ID, the pod ID is created and the name in the spec generator is replaced
+// with ID.
+func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions) (*entities.PodCreateReport, error) {
+	if !strings.HasPrefix(s.Pod, "new:") {
+		return nil, nil
+	}
+	podName := strings.Replace(s.Pod, "new:", "", 1)
+	if len(podName) < 1 {
+		return nil, errors.Errorf("new pod name must be at least one character")
+	}
+	createOptions := entities.PodCreateOptions{
+		Name:          podName,
+		Infra:         true,
+		Net:           netOpts,
+		CreateCommand: os.Args,
+		Hostname:      s.ContainerBasicConfig.Hostname,
+	}
+	// Unset config values we passed to the pod to prevent them being used twice for the container and pod.
+	s.ContainerBasicConfig.Hostname = ""
+	s.ContainerNetworkConfig = specgen.ContainerNetworkConfig{}
+
+	s.Pod = podName
+	return registry.ContainerEngine().PodCreate(context.Background(), createOptions)
+}
+
--- a/cmd/podman/images/scan.go
+++ b/cmd/podman/images/scan.go
@@ -2,13 +2,9 @@ package images
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"github.com/containers/common/pkg/completion"
-	"github.com/containers/podman/v2/cmd/podman/parse"
-	"io/ioutil"
 	"os"
-	"strings"
 	"text/template"
 
 	"github.com/containers/podman/v2/cmd/podman/common"
@@ -35,7 +31,6 @@ var (
 func scanFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 
-	// TODO: derive default from containers.conf configuration
 	flagName := "scanner"
 	flags.StringVarP(
 		&scanOptions.ScannerImage,
@@ -49,14 +44,6 @@ func scanFlags(cmd *cobra.Command) {
 		&scanOptions.MountPoint,
 		flagName, "m", "/podman-scan-image-mount",
 		"Mount-point of the target image to be scanned",
-	)
-	_ = cmd.RegisterFlagCompletionFunc(flagName, completion.AutocompleteNone)
-
-	flagName = "format"
-	flags.StringVarP(
-		&scanOptions.Format,
-		flagName, "f", "squash",
-		"The image format of the target image to be scanned (options: oci-dir, oci-archive, docker-dir, docker-archive, squash)",
 	)
 	_ = cmd.RegisterFlagCompletionFunc(flagName, completion.AutocompleteNone)
 
@@ -77,52 +64,6 @@ func init() {
 		Parent:  imageCmd,
 	})
 	scanFlags(imageScanCommand)
-}
-
-func prepareImageMount(image string) (func(), string, error) {
-	var location string
-	var mountOptions string
-	var cleanupFn func()
-
-	switch {
-	case strings.HasSuffix(scanOptions.Format, "-archive"):
-		file, err := ioutil.TempFile("", "podman-scan-image-archive")
-		if err != nil {
-			return nil, "", fmt.Errorf("unable to create temporary image archive: %w", err)
-		}
-		cleanupFn = func() {
-			os.Remove(file.Name())
-		}
-		location = file.Name()
-		mountOptions = fmt.Sprintf("type=bind,source=%s,target=%s,ro=true", location, scanOptions.MountPoint)
-	case strings.HasSuffix(scanOptions.Format, "-dir"):
-		dir, err := ioutil.TempDir("dir", "podman-scan-image-dir")
-		if err != nil {
-			return nil, "", fmt.Errorf("unable to create temporary image archive: %w", err)
-		}
-		cleanupFn = func() {
-			os.RemoveAll(dir)
-		}
-		location = dir
-		mountOptions = fmt.Sprintf("type=bind,source=%s,target=%s,ro=true", location, scanOptions.MountPoint)
-	case scanOptions.Format == "squash":
-		mountOptions = fmt.Sprintf("type=image,src=%s,target=%s,readwrite=false", image, scanOptions.MountPoint)
-		return nil, mountOptions, nil
-	default:
-		return nil, "", fmt.Errorf("unknown image save format: %q", scanOptions.Format)
-	}
-
-	if err := parse.ValidateFileName(location); err != nil {
-		return cleanupFn, "", err
-	}
-
-	tempSaveOpts := entities.ImageSaveOptions{
-		Format: scanOptions.Format,
-		Output: location,
-	}
-
-	return cleanupFn, mountOptions, registry.ImageEngine().Save(context.Background(), image, nil, tempSaveOpts)
-
 }
 
 func scan(cmd *cobra.Command, args []string) error {
@@ -149,14 +90,7 @@ func scan(cmd *cobra.Command, args []string) error {
 		scannerArgs[i] = buf.String()
 	}
 
-	// save the image (if necessary)
-	cleanupFn, mountOptions, err := prepareImageMount(imageToScan)
-	if cleanupFn != nil {
-		defer cleanupFn()
-	}
-	if err != nil {
-		return err
-	}
+	mountOptions := fmt.Sprintf("type=image,src=%s,target=%s,readwrite=false", imageToScan, scanOptions.MountPoint)
 
 	ctrConfig := registry.PodmanConfig()
 
--- a/cmd/podman/images/scan.go
+++ b/cmd/podman/images/scan.go
@@ -1,117 +1,105 @@
 package images
 
 import (
-	"context"
 	"fmt"
+	"os"
+
 	"github.com/containers/podman/v2/cmd/podman/common"
 	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/specgen"
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"os"
-	"strings"
 )
 
 var (
-	scanDescription = `Scan a container image for known vulnerabilities.  The image name or digest can be used.`
-	scanCommand     = &cobra.Command{
-		Use:     "scan [run-options] IMAGE [scanner-options]",
-		Short:   "Scan a container image for known vulnerabilities",
-		Long:    scanDescription,
-		RunE:    scan,
-		Example: `podman scan centos:latest`,
-	}
-
 	imageScanCommand     = &cobra.Command{
 		Use:     "scan [run-options] IMAGE [scanner-options]",
 		Short:   "Scan a container image for known vulnerabilities",
-		Long:    scanDescription,
+		Long:    `Scan a container image for known vulnerabilities. The image name or digest can be used.`,
+		Args: cobra.MinimumNArgs(1),
 		RunE:    scan,
 		Example: `podman image scan centos:latest`,
 	}
+
+	scanOptions entities.ImageScanOptions
 )
 
-var (
-	scanOptions = entities.ImageScanOptions{
-		ContainerRunOptions: entities.ContainerRunOptions{
-			OutputStream: os.Stdout,
-			ErrorStream:  os.Stderr,
-			Rm: true,
-		},
-	}
-
-	cliVals common.ContainerCLIOpts
-)
 
 func scanFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 
-	common.DefineCreateFlags(cmd, &cliVals)
-	common.DefineNetFlags(cmd)
-
 	// TODO: derive default from containers.conf configuration
 	scannerFlagName := "scanner"
-	flags.StringVar(&scanOptions.ScannerImage, scannerFlagName, "docker.io/library/grype:latest", "The vulnerability scanner container image to use")
+	flags.StringVar(&scanOptions.ScannerImage, scannerFlagName, "docker.io/anchore/grype:latest", "The vulnerability scanner container image to use")
 }
+
 func init() {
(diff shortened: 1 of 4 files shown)


cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+	common.DefineCreateFlags(cmd, &cliVals)
+	common.DefineNetFlags(cmd)
+
+	// TODO: derive default from containers.conf configuration
+	scannerFlagName := "scanner"
+	flags.StringVar(&scanOptions.ScannerImage, scannerFlagName, "docker.io/library/grype:latest", "The vulnerability scanner container image to use")
+}
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
```
vrothberg, line 59:
```suggestion
		Mode:    []entities.EngineMode{entities.ABIMode},
```

vrothberg: Same below.  Let's ignore the existence of the remote client for now :)

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+	})
+	scanFlags(scanCommand)
+
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: imageScanCommand,
+		Parent:  imageCmd,
+	})
+	scanFlags(imageScanCommand)
+
```
vrothberg, line 70:
```suggestion
```

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func scan(cmd *cobra.Command, args []string) error {
+	var err error
+
+	if len(args) < 1 {
```
vrothberg, line 76: Other commands use Cobra's arg checks to make sure we have a consistent error reporting.  There's an `Args` field in the `cobra.Command` struct that we can set to `cobra.ExactArgs(1)`.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+		Short:   "Scan a container image for known vulnerabilities",
+		Long:    scanDescription,
+		RunE:    scan,
+		Example: `podman image scan centos:latest`,
+	}
+)
+
+var (
+	scanOptions = entities.ImageScanOptions{
+		ContainerRunOptions: entities.ContainerRunOptions{
```
vrothberg, line 37: I don't think we should give full access to `ContainerRunOptions`.  I don't want `image scan` to be a way of running containers but to scan a specific image.  The fact that we're running a container should be hidden from the user as much as possible so that they can focus on the security scans and not worry about the (gigantic) amount of options available for running a container.

Let's start with a minimal use case.  If we need more functionality, we can add it in the future.

vrothberg:
Continuing the conversation here:

> Things left todo:
> 
>     * make a PR in `github.com/containers/common` to expose a new configurable in `containers.conf`. I haven't identified where the change should be (maybe [here](https://github.com/containers/common/blob/master/pkg/config/config.go#L74)?) or what it may look like yet.

I think it should be part of the `[engine]` as it's rather something for Podman to interpret.  I think three knobs would make sense:
1. Image-scan image
2. Image-scan command
3. Image-scan mount point (i.e., where the image is being mounted)

> Questions:
> 
>     * are we OK with create/net commands being exposed on the scan command? I couldn't think of a reason not to include them.

I prefer to keep things as simple as possible.   Exposing all container run/create options can easily overwhelm users as it's not easy to find out which options relate to `image scan`.   I suggest to remove the run/create options and add them selectively if we need to in the future.

Thanks a lot for tackling this work, I really appreciate it!

cmd/podman/images/scan.go
wagoodman: This feedback helps a lot, thanks! I was waffling between a minimum profile (as you're suggesting) and a opening up the run options (as implemented now)... I'll make the adjustments :+1:

cmd/podman/images/scan.go
wagoodman: I was thinking about swapping this to `MinimumNArgs` to keep allowing for arguments to be passed to the run command (the scanner). Do you think we should allow this here in the CLI? or only allow extra options through configuration?

vrothberg: I suggest to start with exactly one arg for now. If we need more in the future, we can change :)

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
+package images
+
+import (
+	"context"
+	"fmt"
+	"github.com/containers/podman/v2/cmd/podman/common"
```
rhatdan, line 6: These should be separated.  The base imports from the gihub imports

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+
+func scan(cmd *cobra.Command, args []string) error {
+	var err error
+
+	if len(args) < 1 {
+		return fmt.Errorf("requires exactly one image, given %+v", args)
+	}
+	var imageToScan = args[0]
+	var mountTarget = "/podman-scan-image-mount"
+	var scanMount = fmt.Sprintf("type=image,src=%s,target=%s", imageToScan, mountTarget)
```
rhatdan, line 81: Should default to read/only.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/containers/podman/v2/pkg/specgen"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"os"
+	"strings"
+)
+
+var (
+	scanDescription = `Scan a container image for known vulnerabilities.  The image name or digest can be used.`
+	scanCommand     = &cobra.Command{
```
rhatdan, line 18: I don't think we should be adding any new primary commands.

This should be 

podman image scan ...
But no podman scan.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// allow parsing to continue with the first arg being the scanner image (not the image to scan)
+	args[0] = scanOptions.ScannerImage
+
+	cliVals.Net, err = common.NetFlagsToNetOptions(cmd)
+	if err != nil {
+		return err
+	}
+
+	if af := cliVals.Authfile; len(af) > 0 {
```
rhatdan, line 93: I think we should force the users to run these commands on previously pulled images, and not deal with pulling them within the command.  Users should have to pull the scanner image and pull the image to scan.  Let's keep this simple to start.

rhatdan:
It might be best to start with the man pages, to conclude the user experience.

$ podman image scan registry.fedoraproject.org/fedora

$ podman image scan --scanner quay.io/podman/scanner registry.fedoraproject.org/fedora

What does the user get as output?  Is there a scan report?   A json file?

We used to have the command `atomic scan`  can we steal something from there?

cmd/podman/images/scan.go
wagoodman: righto --the default for an `image` mount type is readonly (`readwrite=false`) but explicit is more clear anyway

cmd/podman/images/scan.go
wagoodman: I feel that since the scan tool will be agnostic that the user will want to pass options to change behavior (output verbosity, format options, scanner profiles/policies, etc.). The CLI is a natural way to achieve this:

```
# where anything after "--" is passed as arguments to the scanner container
podman image scan myimage:latest -- -o json -v --fail-on critical
```

That being said, I'm open for other options how to achieve this... maybe this can be replaced with `container.conf` options (or similar)? Where do you think these options should go in the context of podman configuration?

wagoodman:
@vrothberg @rhatdan Thanks for the comments! I Found some time this afternoon to tackle some of this. I added a man page entry to show the basic behavior. I wanted to highlight a couple thoughts:

- regarding scanner input, the "standard" point for providing input would be for the tool image to respond to a `PODMAN_SCAN_MOUNT` environment variable. This doesn't seem ideal, but also seems better than assuming that the scanner takes positional arguments tailored to be a path to a directory. Any idea's on how to smooth this over a bit --keep it agnostic yet more flexibly configured?
- regarding the report output, given that the `podman image scan` command will be a generic spot to plugin various tooling then trying to normalize to a specific report format seems wrong. I feel that the tool should behave as freely as possible to express the results (even if say a TUI is brought up for the user to interact with). In this way if the tool is report-oriented then the user can run `podman image scan myimage:latest > report.txt` or similar to get the output. If the tool is acting as a quality gate, then `podman image scan myimage:latest` should exit with the same exit code as the scanner tool.
- regarding `atomic`, I'm not certain! I took a quick look and it wasn't apparent what could/should be pulled from that code base --let me know if you have some specific thoughts here.

Shout out if you had more thoughts / different ideas about the user experience!

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/containers/podman/v2/pkg/specgen"
+	"github.com/spf13/cobra"
+)
+
+var (
+	imageScanCommand     = &cobra.Command{
+		Use:     "scan [run-options] IMAGE [scanner-options]",
+		Short:   "Scan a container image for known vulnerabilities",
+		Long:    `Scan a container image for known vulnerabilities. The image name or digest can be used.`,
+		Args: cobra.MinimumNArgs(1),
```
vrothberg, line 20: The files needs formatting via `gofmt`.  `make validate` should do this automatically.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/specgen"
+	"github.com/spf13/cobra"
+)
+
+var (
+	imageScanCommand     = &cobra.Command{
+		Use:     "scan [run-options] IMAGE [scanner-options]",
+		Short:   "Scan a container image for known vulnerabilities",
+		Long:    `Scan a container image for known vulnerabilities. The image name or digest can be used.`,
```
vrothberg, line 19:
```suggestion
		Long:    `Scan a container image for known vulnerabilities.`,
```
Image IDs should work as well.  I'd rather explain some of the details.  For instance, that the scanner itself is a container image and that the specified image to scan will be mounted inside.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/libpod/define"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/containers/podman/v2/pkg/specgen"
+	"github.com/spf13/cobra"
+)
+
+var (
+	imageScanCommand     = &cobra.Command{
+		Use:     "scan [run-options] IMAGE [scanner-options]",
```
vrothberg, line 17:
```suggestion
		Use:     "scan [options] IMAGE [scanner-options]",
```

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,117 @@
...
+func scan(cmd *cobra.Command, args []string) error {
+	imageToScan := args[0]
+	mountTarget := "/podman-scan-image-mount"
+	// allow parsing to continue with the first arg being the scanner image (not the image to scan)
+	args[0] = scanOptions.ScannerImage
+
+	ctrConfig := registry.PodmanConfig()
+
+	ctrOpts := common.ContainerCLIOpts{
+		Mount: []string{fmt.Sprintf("type=image,src=%s,target=%s,readwrite=false", imageToScan, mountTarget)},
```
vrothberg, line 55: I don't think that any options but `Mount` are needed.  The backend should automatically add defaults in case the fields aren't set.

docs/source/markdown/podman-image-scan.1.md
```diff
@@ -0,0 +1,43 @@
...
+
+## RETURN VALUE
+The scanner tool report.
+
+## OPTIONS
+
+#### **--scanner**, **-s** [*image*]
+
+A container image with the scanner tool installed and configured as the entrypoint and 
+reads the PODMAN_SCAN_MOUNT environment variable to determine where to scan.
```
vrothberg, line 21: I prefer `PODMAN_SCAN_MOUNT` to be another flag/option on the CLI (e.g., `--mount-point`).  Mixing inputs on CLI flags and env variables can easily get confusing.

pkg/domain/entities/images.go
```diff
@@ -377,3 +377,7 @@ type ImageUnmountReport struct {
 	Err error
 	Id  string // nolint
 }
+
+type ImageScanOptions struct {
```
vrothberg, line 381: Please add comments to the struct and its fields.

Not all types in the `entities` package are commented yet but they're super helpful in IDEs and docs.

cmd/podman/images/scan.go
vrothberg: In case I am wrong, please add a comment why these options are needed.

vrothberg:
Thanks, @wagoodman!  Really nice progress.

>     * regarding scanner input, the "standard" point for providing input would be for the tool image to respond to a `PODMAN_SCAN_MOUNT` environment variable. This doesn't seem ideal, but also seems better than assuming that the scanner takes positional arguments tailored to be a path to a directory. Any idea's on how to smooth this over a bit --keep it agnostic yet more flexibly configured?

I'd add another CLI flag (e.g., `--mount[-point]`).

>     * regarding the report output, given that the `podman image scan` command will be a generic spot to plugin various tooling then trying to normalize to a specific report format seems wrong. I feel that the tool should behave as freely as possible to express the results (even if say a TUI is brought up for the user to interact with). In this way if the tool is report-oriented then the user can run `podman image scan myimage:latest > report.txt` or similar to get the output. If the tool is acting as a quality gate, then `podman image scan myimage:latest` should exit with the same exit code as the scanner tool.
> 
>     * regarding `atomic`, I'm not certain! I took a quick look and it wasn't apparent what could/should be pulled from that code base --let me know if you have some specific thoughts here.
> 
> 
> Shout out if you had more thoughts / different ideas about the user experience!

vrothberg:
@wagoodman, are you still working on it?

Please let us know in case you don't find time anymore. We can take the commit and kick it over the finish line.

wagoodman:
@vrothberg I should be back on this next week, took a month off to welcome a new little one into the family 👶😊

vrothberg:
> @vrothberg I should be back on this next week, took a month off to welcome a new little one into the family babyblush.

Wonderful news, congratulations!

wagoodman:
@vrothberg thanks! The little dude (Miles) is doing well and we're starting to get more rest with each passing day :)

I've also had some time to think about features of the `podman image scan` command a bit more:

1. It seems that since `podman image scan` is going to be agnostic to which tool is used, it could be expanded for the tool purpose as well. That is, the original idea was to mimic `docker scan` in terms of showing vulnerabilities, however, since the tooling doing the image scan is swappable it doesn't have to be constrained to a single use case. It would be way more useful if you could plug in different kinds of image scanners (e.g. vulnerabilities, secrets scanner, security linters).

2. The image mount approach works quite well, however, I think there is good reason to allow for additional image presentations to the scanner. 

   a. Since the image mount is the squashed representation of the image filesystem it doesn't show files that were removed in a previous layer but still exist in an inner-layer (say if built from a dockerfile) --this is important information for something like a secrets scanner, where the scanner needs to be able to see all files in all layers, even if they are not visible from filesystem of say a running container derived from the image in question.

   b. There are several scanners that already exist, and they take different formats as input: a plain filesystem as would be seen within a container instantiated from the image, a Docker tar archive from a `docker image save`, an OCI tar archive, an OCI directory layout. Since podman already has the ability to present the contents of an image in all of these formats it makes sense to allow for scanners in the `podman image scan` command to have access to the image in the format of choice.

3. The current prototype assumes that the scanner image needs to respond to the `PODMAN_SCAN_MOUNT ` environment variable --this really isn't ideal since the tool needs to change to be used as a `podman image scan` scanner. The problem is that there needs to be a way to feed configuration such as mountpoint and image format details into the scanner tool in a flexible fashion... I think I have a good way to express that, I'll update the manpage changes to reflect this, but the summary is to bring templating into arguments that are passed into the scanner.

   For instance, the `grype` tool is used as so when using a tar archive:
   ```
   docker image save -o image.tar myimage:latest

   grype docker-tar:./image.tar -o json
   ```

   It would be ideal to configure `podman image scan myimage:latest` to perform something along the lines of:
   ```
   podman image scan myimage:latest --mountpoint /tmp/image.tar --format docker-archive --scanner anchore/grype:latest -- 'docker-tar:{{ .Mountpoint }}' -o json
   ```

   Taking this as static configuration from disk would be most ideal, but allowing the options from the CLI should be first-class as well. 


Again, I'll update the branch as I prototype out these ideas. I'm interested in your feedback in the above points!

wagoodman:
@vrothberg just pushed the man page updates + updated the code to reflect the changes. What's missing is how to provide all of these options via configuration (instead of needing to provide all of the options via command line every time). If we go this direction, would you know the best place for this configuration? (maybe somewhere in containers.conf?)

vrothberg:
> @vrothberg thanks! The little dude (Miles) is doing well and we're starting to get more rest with each passing day :)

Wonderful!

> I've also had some time to think about features of the `podman image scan` command a bit more:
> 
>     1. It seems that since `podman image scan` is going to be agnostic to which tool is used, it could be expanded for the tool purpose as well. That is, the original idea was to mimic `docker scan` in terms of showing vulnerabilities, however, since the tooling doing the image scan is swappable it doesn't have to be constrained to a single use case. It would be way more useful if you could plug in different kinds of image scanners (e.g. vulnerabilities, secrets scanner, security linters).

Absolutely right!  I think that's worth mentioning in the docs that the "scanners" mustn't necessarily be limited to a security-only context.  Some scanners may want to list all installed .dep/rpm/python/js packages.

>     2. The image mount approach works quite well, however, I think there is good reason to allow for additional image presentations to the scanner.
>        a. Since the image mount is the squashed representation of the image filesystem it doesn't show files that were removed in a previous layer but still exist in an inner-layer (say if built from a dockerfile) --this is important information for something like a secrets scanner, where the scanner needs to be able to see all files in all layers, even if they are not visible from filesystem of say a running container derived from the image in question.

This seems like a quite advanced yet important use case.  I suggest starting with the simple use case of the fully expanded root FS first. We can extend it in the future.  It would require some plumbing in the containers/storage library.  But I really like the idea. @rhatdan PTAL

>        b. There are several scanners that already exist, and they take different formats as input: a plain filesystem as would be seen within a container instantiated from the image, a Docker tar archive from a `docker image save`, an OCI tar archive, an OCI directory layout. Since podman already has the ability to present the contents of an image in all of these formats it makes sense to allow for scanners in the `podman image scan` command to have access to the image in the format of choice.

Sounds good to me.  It should be easy to support as `podman run/create`, for instance, already support all these transports. So you can do `podman run oci:fedora ls`.

>     3. The current prototype assumes that the scanner image needs to respond to the `PODMAN_SCAN_MOUNT ` environment variable --this really isn't ideal since the tool needs to change to be used as a `podman image scan` scanner. The problem is that there needs to be a way to feed configuration such as mountpoint and image format details into the scanner tool in a flexible fashion... I think I have a good way to express that, I'll update the manpage changes to reflect this, but the summary is to bring templating into arguments that are passed into the scanner.
>        For instance, the `grype` tool is used as so when using a tar archive:
[...]
> Again, I'll update the branch as I prototype out these ideas. I'm interested in your feedback in the above points!

Feel free to update the PR and ping me.  I am interested in your ideas!

rhatdan:
@wagoodman Are you still working on this?

wagoodman:
@vrothberg 

> This seems like a quite advanced yet important use case. I suggest starting with the simple use case of the fully expanded root FS first.

re: supporting more than the root FS, the PR has this implemented already here https://github.com/containers/podman/pull/8706/files#diff-0d9d4382690094eec9c6d7dbe310c7a320686911170923af27d6d44a927603b0R82 (just by leveraging the ready-made `registry.ImageEngine().Save(...)`). Are we saying to take this out?

Also, regarding `#3` from https://github.com/containers/podman/pull/8706/files#issuecomment-776755461 , is the `containers.conf` the best place for this kind of this configuration? Or more generally, where should the configuration for this new "scan" options reside?

vrothberg:
> @vrothberg
> 
> > This seems like a quite advanced yet important use case. I suggest starting with the simple use case of the fully expanded root FS first.
> 
> re: supporting more than the root FS, the PR has this implemented already here https://github.com/containers/podman/pull/8706/files#diff-0d9d4382690094eec9c6d7dbe310c7a320686911170923af27d6d44a927603b0R82 (just by leveraging the ready-made `registry.ImageEngine().Save(...)`). Are we saying to take this out?

I'd vote to take this out, yes.  We can add it in the future if there is a sufficiently big need for it.  In my experience, simplicity is the most important attribute in terms of usability and also maintainability.

> Also, regarding `#3` from https://github.com/containers/podman/pull/8706/files#issuecomment-776755461 , is the `containers.conf` the best place for this kind of this configuration? Or more generally, where should the configuration for this new "scan" options reside?

The link points to nirvana, so I miss some context.  `containers.conf` sounds like a good candidate in this context though.  But note that we can add support for containers.conf in a future PR.  That may help reduce the scope of the PR.

wagoodman:
@vrothberg righto, I've removed the format option. Let me know if you think there is any missing functionality that we should be considering or other areas that need to be updated.

cmd/podman/images/scan.go
vrothberg: Still open.

cmd/podman/images/scan.go
vrothberg: Still open.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,158 @@
...
+	_ = cmd.RegisterFlagCompletionFunc(flagName, completion.AutocompleteNone)
+
+	flagName = "env"
+	flags.StringSliceVarP(
+		&scanOptions.Env,
+		flagName, "e", nil,
+		"Set environment variables in scanner tool container",
+	)
+	_ = cmd.RegisterFlagCompletionFunc(flagName, completion.AutocompleteNone)
+
```
vrothberg, line 57: Remove empty line.

pkg/domain/entities/images.go
```diff
@@ -374,3 +374,15 @@ type ImageUnmountReport struct {
...
+	// ScannerImage is the scanner tool image to use to scan the target image
+	ScannerImage string
+	// MountPoint is the path of the volume mount of the target image within the scanner container
+	MountPoint string
+	// Format of saving the image: oci-archive, oci-dir (directory with oci
+	// manifest type), docker-archive, docker-dir (directory with v2s2
+	// manifest type), or "squash" (the squashed representation of the image file tree)
+	Format string
+	Env    []string
```
vrothberg, line 387: Comment for `Env` is missing.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,158 @@
...
+		scannerArgs[i] = buf.String()
+	}
+
+	mountOptions := fmt.Sprintf("type=image,src=%s,target=%s,readwrite=false", imageToScan, scanOptions.MountPoint)
+
+	ctrConfig := registry.PodmanConfig()
+
+	ctrOpts := common.ContainerCLIOpts{
+		Mount:             []string{mountOptions},
+		ImageVolume:       common.DefaultImageVolume,
```
vrothberg, line 99: @mheon, do we need to set all these options here or does `FillOutSpecGen` take care of it?  I am a bit worried of code duplication and potential divergence in the future.

docs/source/markdown/podman-image-scan.1.md
```diff
@@ -0,0 +1,70 @@
...
+```
+podman image scan --scanner anchore/grype:latest alpine:latest
+```
+
+Pass arguments to the scanner:
+```
+podman image scan alpine:latest -- 'dir:{{ .Mountpoint }}' -o json -vv
+```
+
+Change the image format presented to the scanner:
```
vrothberg, line 64: This section can be removed.

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,158 @@
...
+	scanOptions entities.ImageScanOptions
+)
+
+func scanFlags(cmd *cobra.Command) {
+	flags := cmd.Flags()
+
+	flagName := "scanner"
+	flags.StringVarP(
+		&scanOptions.ScannerImage,
+		flagName, "s", "docker.io/anchore/syft:latest",
```
vrothberg, line 37: Can we pin it to a specific version of `syft` in contrast to using `latest`?  `latest` is a moving target which makes it hard to test, verify and rely on.

vrothberg (commented):
Thanks!

I think it's time to discuss testing. I suggest to upload a small image to `quay.io/libpod` that we can use for testing and then run `podman image scan` on it.

I suggest to create a new test in `test/system`. @edsantiago can help us polish the tests.

cmd/podman/images/scan.go
mheon: Using the CLI parsing code for this is a bit iffy because it was written to expect CLI-provided defaults for flags.

mheon: I would say we probably need it as written. Directly using Specgen would remove the necessity.

vrothberg:
Friendly ping.

docs/source/markdown/podman-image-scan.1.md
```diff
@@ -0,0 +1,70 @@
...
+A container image with the scanner tool installed and configured as the entrypoint. The scanner tool should exit with
+a non-zero code to indicate a failing scan.
+
+## EXAMPLE
+
+Use the default scanner to scan alpine:latest
+```
+podman image scan alpine:latest
+
+#... the scanner report is output to stdout / stderr>
```
cdoern, line 51: show an example output so the user knows what to expect

cmd/podman/images/scan.go
```diff
@@ -0,0 +1,158 @@
...
+		OutputStream: os.Stdout,
+		ErrorStream:  os.Stderr,
+		InputStream:  os.Stdin,
+		Rm:           true,
+		Spec:         s,
+		SigProxy:     true,
+		DetachKeys:   ctrConfig.DetachKeys(),
+	}
+
+	report, err := registry.ContainerEngine().ContainerRun(registry.GetContext(), runOpts)
```
cdoern, line 143: I might be misunderstanding what this does, but why is running a container w/ the image necessary? It looks like `docker scan` just gives you information about the image?

cdoern:
@wagoodman, still working on this?

wagoodman:
@vrothberg (and @cdoern ) friendly pong! I don't think I'll have cycles anytime in the near future to complete this unfortunately. Please feel free close this PR, continue this work without me, or open another one to take it's place 👍 .

vrothberg:
> @vrothberg (and @cdoern ) friendly pong! I don't think I'll have cycles anytime in the near future to complete this unfortunately. Please feel free close this PR, continue this work without me, or open another one to take it's place +1 .

Thanks for your work on it!  I'll close the PR so others can finish.

PR #8714, opened 2020-12-14
RHEL gating tests: more journald exceptions

edsantiago:
Followup to #8284, due to my not having noticed #8096.

RHEL gating tests are failing again due to rhbz#1895105, the
one where we can't run journalctl rootless on RHEL. #8284 fixed
this for some RHEL builds of older podman, but I missed #8096
which added yet another logs test.

This brings us to three journalctl exceptions, which means
it gets complicated because I have to refactor it all.

**THIS IS NOT SUSTAINABLE**. We need some way to have a similar
setup in CI, with a permission-less rootless login, so we don't
add yet another logs test some day and discover, months later,
that it doesn't work on RHEL and then have to go into crisis
mode.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -415,13 +415,18 @@ json-file | f
         fi
 
         if [[ $driver != 'none' ]]; then
-            run_podman logs myctr
-            is "$output" "$msg" "check that podman logs works as expected"
+            if [[ $driver = 'journald' ]] && journald_unavailable; then
+                # Cannot perform check
+                :
+            else
+                run_podman logs myctr
+                is "$output" "$msg" "podman logs, with driver '$driver'"
+            fi
         else
             run_podman 125 logs myctr
             if ! is_remote; then
                 is "$output" ".*this container is using the 'none' log driver, cannot read logs.*" \
-                   "podman logs does not work with none log driver"
+                   "podman logs, with driver 'none', should fail with error"
             fi
         fi
         run_podman rm myctr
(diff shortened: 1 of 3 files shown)


edsantiago:
@jnovy FYI

edsantiago:
@containers/podman-maintainers PTAL. Suggestions welcome on how we can catch this sort of problem earlier.

I tested the new code on a RHEL VM, and the journal tests are indeed skipped when run rootless.

mheon:
LGTM

mheon:
Would podman-in-a-container catch the journald case? We wouldn't have access to the host systemd, so I presume journald would blow up.

edsantiago:
Hmmm. Looks like we lost `in_podman` testing when we switched to CI v2. @cevich was that intentional?

@mheon I'm not sure how to test, but my gut feel is that P-in-P would not catch this: the symptom we're looking for is:
```console
$ journalctl -n 1
Hint: You are currently not seeing messages from other users and the system.
      Users in the 'systemd-journal' group can see all messages. Pass -q to
      turn off this notice.
No journal files were opened due to insufficient permissions.
```
This happens when `journalctl` is run as rootless on RHEL on an account that is not in `wheel` or other special magic groups. IIRC, even when we were doing P-in-P testing, the tests ran as root inside the container so any journalctl error would be different, I think?

cevich:
> Looks like we lost in_podman testing when we switched to CI v2.

It's there, just has a new name, `container_integration_test`...errr, where is `container_system_test`?  I swore it was there...huh...apparently only in my imagination.   Okay I'll add it...must have got lost in the shuffle.

PR #8715, opened 2020-12-14
Podman image bindings for 3.0

baude:
Begin the migration of the image bindings for podman 3.0.  this includes
the use of options for each binding.  build was intentionally not
converted as I believe it needs more discussion before migration.
specifically, the build options themselves.

also noteworthly is that the remove image and remove images bindings
were merged into one.  the remove images (or batch remove) has one
downside in that the errors return no longer adhere to http return
codes.  this should be discussed and reimplemented in subsequent code.

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -265,8 +265,14 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 
 	// Format is mandatory! Currently, we only support multi-image docker
 	// archives.
+	if len(query.References) > 1 && query.Format != define.V2s2Archive {
+		utils.Error(w, "unsupported format", http.StatusInternalServerError, errors.Errorf("multi-image archives must use format of %s", define.V2s2Archive))
+		return
+
+	}
+
 	switch query.Format {
-	case define.V2s2Archive:
+	case define.V2s2Archive, define.OCIArchive:
 		tmpfile, err := ioutil.TempFile("", "api.tar")
 		if err != nil {
 			utils.Error(w, "unable to create tmpfile", http.StatusInternalServerError, errors.Wrap(err, "unable to create tempfile"))
@@ -277,6 +283,13 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 			utils.Error(w, "unable to close tmpfile", http.StatusInternalServerError, errors.Wrap(err, "unable to close tempfile"))
 			return
 		}
+	case define.OCIManifestDir, define.V2s2ManifestDir:
+		tmpdir, err := ioutil.TempDir("", "save")
+		if err != nil {
+			utils.Error(w, "unable to create tmpdir", http.StatusInternalServerError, errors.Wrap(err, "unable to create tempdir"))
+			return
+		}
+		output = tmpdir
 	default:
 		utils.Error(w, "unsupported format", http.StatusInternalServerError, errors.Errorf("unsupported format %q", query.Format))
 		return
@@ -287,7 +300,7 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 	opts := entities.ImageSaveOptions{
 		Compress:          query.Compress,
 		Format:            query.Format,
-		MultiImageArchive: true,
+		MultiImageArchive: len(query.References) > 1,
 		Output:            output,
 		RemoveSignatures:  true,
 	}
@@ -414,7 +427,6 @@ func PushImage(w http.ResponseWriter, r *http.Request) {
 	}{
 		// This is where you can override the golang default value for one of fields
 	}
-
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
 		return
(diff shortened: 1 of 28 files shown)


pkg/bindings/images/types_diffoptions.go
```diff
@@ -0,0 +1,86 @@
+package images
```
mheon, line 1: Suggestion: Can you stuff an `_` between `diff` and `options` in the filenames?

pkg/api/handlers/libpod/images.go
```diff
@@ -311,7 +311,7 @@ func ImagesLoad(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
-		Reference string `schema:"reference"`
+		Name string `schema:"name"`
```
vrothberg, line 314: I prefer to not introduce breaking changes to the endpoints.  Renaming a parameter has IMHO little benefit to users but may break them.

pkg/bindings/images/pull.go
```diff
@@ -23,26 +21,28 @@ import (
 // `rawImage` must be a reference to a registry (i.e., of docker transport or be
 // normalized to one).  Other transports are rejected as they do not make sense
 // in a remote context. Progress reported on stderr
-func Pull(ctx context.Context, rawImage string, options entities.ImagePullOptions) ([]string, error) {
+func Pull(ctx context.Context, rawImage string, options *PullOptions) ([]string, error) {
+	if options == nil {
+		t := new(PullOptions)
+		options = t
```
vrothberg, line 27:
```suggestion
		options = new(PullOptions)
```

pkg/bindings/images/diff.go
```diff
@@ -9,7 +9,11 @@ import (
 )
 
 // Diff provides the changes between two container layers
-func Diff(ctx context.Context, nameOrID string) ([]archive.Change, error) {
+func Diff(ctx context.Context, nameOrID string, options *DiffOptions) ([]archive.Change, error) {
+	if options == nil {
```
vrothberg, line 13: Great idea to allow `nil`!

I think it's a nice compromise between avoiding future breaking changes (if we have the need to at fields to the currently empty struct) and being easy to use: `Diff(a, b, nil)` is nicer than `Diff(a, b, &images.DiffOptions)`.

pkg/bindings/images/rm.go
```diff
@@ -3,17 +3,19 @@ package images
...
 
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/errorhandling"
 )
 
-// BachtRemove removes a batch of images from the local storage.
-func BatchRemove(ctx context.Context, images []string, opts entities.ImageRemoveOptions) (*entities.ImageRemoveReport, []error) {
+// Remove removes an image from the local storage.  Use optional force option to remove an
```
vrothberg, line 13:
```suggestion
// Remove removes one or more images from the local storage.  Use optional force option to remove an
```

pkg/bindings/images/types_diffoptions.go
vrothberg: +1

vrothberg (commented):
I wonder if we could remove the bindings tests.  As the bindings are used the remote client we are testing the bindings extensively.  Removing them would reduce maintenance burden and save some CI cycles.

pkg/api/handlers/libpod/images.go
```diff
@@ -277,6 +283,13 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 			utils.Error(w, "unable to close tmpfile", http.StatusInternalServerError, errors.Wrap(err, "unable to close tempfile"))
 			return
 		}
+	case define.OCIManifestDir, define.V2s2ManifestDir:
+		tmpdir, err := ioutil.TempDir("", "save")
+		if err != nil {
+			utils.Error(w, "unable to create tmpdir", http.StatusInternalServerError, errors.Wrap(err, "unable to create tempdir"))
```
TomSweeneyRedHat, line 289:
```suggestion
			utils.Error(w, "unable to create tmpdir", http.StatusInternalServerError, errors.Wrap(err, "unable to create tmpdir"))
```

pkg/bindings/images/images.go
```diff
@@ -305,28 +292,28 @@ func Push(ctx context.Context, source string, destination string, options entiti
...
+	params.Set("term", term)
 
-	if opts.SkipTLSVerify != types.OptionalBoolUndefined {
-		// Note: we have to verify if skipped is false.
-		verifyTLS := bool(opts.SkipTLSVerify == types.OptionalBoolFalse)
-		params.Set("tlsVerify", strconv.FormatBool(verifyTLS))
+	// Note: we have to verify if skipped is false.
+	if options.SkipTLSVerify != nil {
+		params.Del("SkipTLSVerify")
+		params.Set("tlsVerify", strconv.FormatBool(!options.GetSkipTLSVerify()))
```
TomSweeneyRedHat, line 312: Looks like you do this tls check in a number of places.  Would it make sense to do a utility function?

pkg/bindings/images/types.go
```diff
@@ -1,8 +1,193 @@
...
+	Filters map[string][]string
+}
+
+//go:generate go run ../generator/generator.go TagOptions
+// TagOptions are optional options for tagging images
+type TagOptions struct {
+}
+
+//go:generate go run ../generator/generator.go UntagOptions
+// UntagOptions are optional options for tagging images
```
TomSweeneyRedHat, line 82: tagging or untagging?

vrothberg (commented):
LGTM

Issue #8718, opened 2020-12-14
CI flake: sdnotify-conmon test: getting MAINPID=1 instead of READY


edsantiago:
Symptom: the `sdnotify : conmon` test in `260-sdnotify.bats` is flaking a lot:
```
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: final output from sdnotify
# #| expected: 'READY=1'
# #|   actual: 'MAINPID=98229'
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This is a reminder for Ed to take a look at https://github.com/containers/podman/blob/a226e6ed318ae7b7af16081b787cc80718e73e2d/test/system/260-sdnotify.bats#L104

* gce_instance:fedora : sys podman fedora-33 root host
  * PR #8699
    * [12-12 14:18](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5758580110393344/html/sys-podman-fedora-33-root-host.log.html)
  * PR #8693
    * [12-11 10:56](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4508173396606976/html/sys-podman-fedora-33-root-host.log.html)
* gce_instance:fedora : sys podman fedora-33 rootless host
  * PR #8689
    * [12-12 22:08](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5683119867035648/html/sys-podman-fedora-33-rootless-host.log.html)
  * PR #8676
    * [12-10 18:33](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5210519332192256/html/sys-podman-fedora-33-rootless-host.log.html)
* gce_instance:prior-fedora : sys podman fedora-32 root host
  * PR #8714
    * [12-14 09:53](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6436920483905536/html/sys-podman-fedora-32-root-host.log.html)
* gce_instance:prior-fedora : sys podman fedora-32 root host
  * PR #7357
[...]

PR #8720, opened 2020-12-14
system tests: the catch-up game

edsantiago:
- run test: minor cleanup to .containerenv test. Basically,
  make it do only two podman-runs (they're expensive) and
  tighten up the results checks

- ps test: add ps -a --storage. Requires small tweak to
  run_podman helper, so we can have "timeout" be an expected
  result

- sdnotify test: workaround for #8718 (seeing MAINPID=xxx as
  last output line instead of READY=1). As found by the
  newly-added debugging echos, what we are seeing is:

      MAINPID=103530
      READY=1
      MAINPID=103530

  It's not supposed to be that way; it's supposed to be just
  the first two. But when faced with reality, we must bend
  to accommodate it, so let's accept READY=1 anywhere in
  the output stream, not just as the last line.

### DIFF
--- a/test/system/040-ps.bats
+++ b/test/system/040-ps.bats
@@ -82,4 +82,43 @@ load helpers
     run_podman rm -a
 }
 
+@test "podman ps -a --storage" {
+    skip_if_remote "ps --storage does not work over remote"
+
+    # Setup: ensure that we have no hidden storage containers
+    run_podman ps --storage -a
+    is "${#lines[@]}" "1" "setup check: no storage containers at start of test"
+
+    # Force a buildah timeout; this leaves a buildah container behind
+    PODMAN_TIMEOUT=5 run_podman 124 build -t thiswillneverexist - <<EOF
+FROM $IMAGE
+RUN sleep 30
+EOF
+
+    run_podman ps -a
+    is "${#lines[@]}" "1" "podman ps -a does not see buildah container"
+
+    run_podman ps --storage -a
+    is "${#lines[@]}" "2" "podman ps -a --storage sees buildah container"
+    is "${lines[1]}" \
+       "[0-9a-f]\{12\} \+$IMAGE *buildah .* seconds ago .* storage .* ${PODMAN_TEST_IMAGE_NAME}-working-container" \
+       "podman ps --storage"
+
+    cid="${lines[1]:0:12}"
+
+    # 'rm -a' should be a NOP
+    run_podman rm -a
+    run_podman ps --storage -a
+    is "${#lines[@]}" "2" "podman ps -a --storage sees buildah container"
+
+    # This is what deletes the container
+    # FIXME: why doesn't "podman rm --storage $cid" do anything?
+    run_podman rm -f "$cid"
+
+    run_podman ps --storage -a
+    is "${#lines[@]}" "1" "storage container has been removed"
+}
+
+
+
 # vim: filetype=sh
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -548,27 +548,33 @@ json-file | f
 }
 
 @test "Verify /run/.containerenv exist" {
-	run_podman run --rm $IMAGE ls -1 /run/.containerenv
-	is "$output" "/run/.containerenv"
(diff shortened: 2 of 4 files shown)


edsantiago:
@containers/podman-maintainers PTAL (low priority). I have updated the commit message, PR description, and code thanks to a stroke of luck: what was intended as a debug-printf for analyzing flakes, actually triggered right here, so I was able to gather data on #8718 and add a workaround. (I'm not going to claim it as fixed though).

test/system/040-ps.bats
```diff
@@ -82,4 +82,43 @@ load helpers
...
+
+    cid="${lines[1]:0:12}"
+
+    # 'rm -a' should be a NOP
+    run_podman rm -a
+    run_podman ps --storage -a
+    is "${#lines[@]}" "2" "podman ps -a --storage sees buildah container"
+
+    # This is what deletes the container
+    # FIXME: why doesn't "podman rm --storage $cid" do anything?
```
rhatdan, line 115: --storage is really just a flag to add to --all.  The combination removes all --storage containers and all podman containers.

rhatdan:
LGTM

test/system/040-ps.bats
edsantiago: But `--storage` doesn't actually seem to do anything at all: `rm -a` removes storage containers just by itself. And from looking at the code, I don't see `--storage` assigned to any option variable that is checked anywhere (but I could be wrong):
https://github.com/containers/podman/blob/999d40d2c76992cfd2d8c0827b7b0d00c4a2a661/cmd/podman/containers/rm.go#L71-L75

rhatdan: Your right we don't refer to these as --storage but as --external in the CLI. And I don't see any use of --external in podman rm.  
I guess we only use it in podman ps and podman container exists.

But podman rm can remove --external containers, if they are specifies by name.  But not --all

edsantiago: So.... playing with this a little more... I'm getting more confused. There seem to be two types of buildah containers: complete ones, and incomplete ones? Let me see if I can express this in a way that makes sense.

Complete buildah containers can be seen and removed by podman without `-f`:
```console
$ buildah from quay.io/libpod/testimage:20200929
testimage-working-container
$ ./bin/podman rm testimage-working-container
testimage-working-container      <--- yay, success
```
Incomplete buildah containers can only be removed with `-f`:
```console
$ printf "FROM quay.io/libpod/testimage:20200929\nRUN sleep 30\n"| timeout -v 5 podman build -t thiswillneverexist -
STEP 1: FROM quay.io/libpod/testimage:20200929
STEP 2: RUN sleep 30
timeout: sending signal TERM to command ‘podman’
$ ./bin/podman rm testimage-working-container
Error: no container with name or ID testimage-working-container found: no such container
$ ./bin/podman ps -a --storage
CONTAINER ID  IMAGE                              COMMAND  CREATED         STATUS   PORTS   NAMES
ef0cfdd5fc92  quay.io/libpod/testimage:20200929  buildah  51 seconds ago  storage          testimage-working-container
$ ./bin/podman rm ef0cfdd5fc92
Error: no container with name or ID ef0cfdd5fc92 found: no such container
$ ./bin/podman rm -f ef0cfdd5fc92
ef0cfdd5fc92
```
As someone clueless about storage internals, I find this perplexing. It makes me think I might need to extend this new test so it checks both cases, complete and incomplete containers (or whatever the proper name is).

@rhatdan, @nalind, does this surprise you, or is it just a "duh, of course" that I'm not understanding?

rhatdan: I think ./bin/podman rm ef0cfdd5fc92 failing is a bug.  I am not sure why it failed.

edsantiago: It's 100% reproducible though. There must be something different between the two types of containers; I just don't know how to even look for such a difference.

edsantiago: Filed #8735

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
LGTM

PR #8723, opened 2020-12-14
Clean up temporary file.

matejvasek:
The `/images/load` APIv2 endpoint creates temporary files that are not being removed.

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -22,6 +22,7 @@ import (
 	"github.com/gorilla/schema"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // mergeNameAndTagOrDigest creates an image reference as string from the
@@ -386,6 +387,12 @@ func LoadImages(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to create tempfile"))
 		return
 	}
+	defer func() {
+		err := os.Remove(f.Name())
+		if err != nil {
+			logrus.Errorf("Failed to remove temporary file: %v.", err)
+		}
+	}()
 	if err := SaveFromBody(f, r); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to write temporary file"))
 		return


matejvasek:
I run out of disc space because of those files.

pkg/api/handlers/compat/images.go
```diff
@@ -386,6 +388,12 @@ func LoadImages(w http.ResponseWriter, r *http.Request) {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to create tempfile"))
 		return
 	}
+	defer func() {
+		err := os.Remove(f.Name())
+		if err != nil {
+			logrus.Warnf("Failed to remove temporary file: %s.", f.Name())
```
TomSweeneyRedHat, line 394: I wonder if this should be Warnf or Debugf....

rhatdan: It should be a Errorf, and it should print out the error.  BTW when you print out the Error, it will include the f.Name(), so dont add it.

logrus.Errorf("Failed to remove temporary file: %v". err)

pkg/api/handlers/compat/images.go
```diff
@@ -9,6 +9,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
```
rhatdan, line 13: No blank line

rhatdan (approved):
@vdemeester Nice catch just a couple of changes.

pkg/api/handlers/compat/images.go
matejvasek: The blank line was added by `goimports`.

matejvasek: I could try deleting it, but don't know what linter would think about it.

rhatdan:
@matejvasek Can you please squash your commits into a single commit.

git rebase -i origin 
git push --force

matejvasek:
@rhatdan why is rdoproject.org/github-check failing?

rhatdan:
@matejvasek ignore it, not related to this PR.  This failing is a non blocking failure.

rhatdan:
But to answer your question, we have switched our defaults to crun and this is not available on centos8, which is why it is failing.  I have a PR into containers/common to not hard code the default oci runtime but to search for one, this should eventually fix this problem once it gets merged into podman

TomSweeneyRedHat:
LGTM

pkg/api/handlers/compat/images.go
TomSweeneyRedHat: @rhatdan, Good call.  My noggin was thinking last night that we'd be printing an error with every file we attempted to remove.  It will only print on ones we're trying to remove and fail to do so.

PR #8725, opened 2020-12-15
Allow users to specify TMPDIR in containers.conf

rhatdan:
Currently we hard code TMPDIR environment variable to /var/tmp
if it is not set in the Environment. This causes TMPDIR environment
variable to be ignored if set in containers.conf.

This change now uses the host environment TMPDIR, followed by
containers.conf and then hard codes TMPDIR, if it was not set.

### DIFF
--- a/cmd/podman/main.go
+++ b/cmd/podman/main.go
@@ -31,11 +31,6 @@ func main() {
 		return
 	}
 
-	// Hard code TMPDIR functions to use /var/tmp, if user did not override
-	if _, ok := os.LookupEnv("TMPDIR"); !ok {
-		os.Setenv("TMPDIR", "/var/tmp")
-	}
-
 	rootCmd = parseCommands()
 
 	Execute()
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -178,6 +178,10 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
+	// Hard code TMPDIR functions to use /var/tmp, if user did not override
+	if _, ok := os.LookupEnv("TMPDIR"); !ok {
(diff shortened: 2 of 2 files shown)


rhatdan:
Truly fixes https://github.com/containers/podman/pull/5412

rhatdan:

mheon:
LGTM

QiWang19 (commented):
LGTM

ashley-cui (commented):
/lgtm

edsantiago:
I meant to ask earlier if there was any way to test this. I assume that this change updates the precedence, but I couldn't find a way to confirm. How is `$TMPDIR` visible to a podman command-line user?

edsantiago:
Self-followup: found a way using `unshare`:
```console
$ cat ~/.config/containers/containers.conf
[engine]
env = ["TMPDIR=/this/is/set/in/containers.conf"]
$ env -u TMPDIR ./bin/podman unshare /usr/bin/env|grep TMPD
TMPDIR=/this/is/set/in/containers.conf      <--- Good. Pre-8725, this prints /var/tmp
$ TMPDIR=/this/is/set/in/user/environment ./bin/podman unshare /usr/bin/env|grep TMPD
TMPDIR=/this/is/set/in/user/environment     <--- Good. Same thing pre-8725
```
Day late and dollar short, but /lgtm :-)

I do not plan to write a system test for this: I try hard to avoid mucking with `.conf` files in system tests due to the risk of not cleaning up properly and causing cascading failures. If anyone objects, and believes we need a regression test for this, please let me know.

rhatdan:
Sorry I could not figure a good way of testing this out.  So I instrumented the tool to see what TMPDIR was set to and then tried every scenario I could try.

Issue #8721, opened 2020-12-14
rootless: cgroups v1: Failed to add conmon to cgroupfs sandbox cgroup: error creating cgroup for SOMETHING


edsantiago:
Fallout from #8479 (changing log-level from error to warn): `podman run`, rootless, on a cgroups v1 system, emits unhelpful warnings:
```console
$ podman run quay.io/libpod/testimage:20200929 date
WARN[0000] Failed to add conmon to cgroupfs sandbox cgroup: error creating cgroup for memory: mkdir /sys/fs/cgroup/memory/libpod_parent: permission denied
Mon Dec 14 21:39:58 UTC 2020
```
The command works fine, exits 0 and everything, but the warning is harmful to CI tests as well as to users who neither know nor care about libpod_parent and stuff like that.

The precise message differs. These are variants I see:
```
...creating cgroup for cpu: mkdir /sys/fs/cgroup/cpu/libpod_parent: permission denied
...creating cgroup for blkio: mkdir /sys/fs/cgroup/blkio/libpod_parent: permission denied
...creating cgroup for pids: mkdir /sys/fs/cgroup/pids/libpod_parent: permission denied
```
Please either fix the condition leading to the warning, or remove the warning.

rhatdan:
@giuseppe @haircommander @mheon WDYT?

giuseppe:
Proposed fix here: https://github.com/containers/podman/pull/8726

PR #3832, opened 2019-08-16
Cirrus: Confirm networking more

cevich:
Signed-off-by: Chris Evich <cevich@redhat.com>

PR #8726, opened 2020-12-15
libpod, conmon: change log level for rootless

giuseppe:
Change the log level when running as rootless when moving conmon to a
different cgroup.

Closes: https://github.com/containers/podman/issues/8721

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -1433,6 +1433,14 @@ func (r *ConmonOCIRuntime) moveConmonToCgroupAndSignal(ctr *Container, cmd *exec
 	}
 
 	if mustCreateCgroup {
+		// Usually rootless users are not allowed to configure cgroupfs.
+		// There are cases though, where it is allowed, e.g. if the cgroup
+		// is manually configured and chowned).  Avoid detecting all
+		// such cases and simply use a lower log level.
+		logLevel := logrus.WarnLevel
+		if rootless.IsRootless() {
+			logLevel = logrus.InfoLevel
+		}
 		// TODO: This should be a switch - we are not guaranteed that
 		// there are only 2 valid cgroup managers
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

libpod/oci_conmon_linux.go
```diff
@@ -1433,6 +1433,14 @@ func (r *ConmonOCIRuntime) moveConmonToCgroupAndSignal(ctr *Container, cmd *exec
 	}
 
 	if mustCreateCgroup {
+		// Usually rootless users are not allowed to configure cgroupfs.
+		// There are cases though, where it is allowed, e.g. if the cgroup
+		// is manually configured and chowned).  Avoid detecting all
```
edsantiago, line 1438: Trailing whitespace is causing CI failure

edsantiago:
@cevich this is barfing trying to connect to `mirror.chpc.utah.edu:443`. That host is in  https://github.com/containers/podman/blob/999d40d2c76992cfd2d8c0827b7b0d00c4a2a661/contrib/cirrus/required_host_ports.txt#L8 with the commit in question being #3832, which unfortunately only says "Confirm networking more" but gives no indication of why this host is important. Do you have any recollection of why this host is needed? Can we remove it?

edsantiago:
@giuseppe the string "chpc" does not appear anywhere else in the podman source tree, leading me to believe that this check is unnecessary. For purposes of getting CI to pass, I would recommend editing the above file and removing the chpc line, updating the commit message appropriately with something like "unrelated: remove unused mirror from list of required host/ports: the host is unreachable due to DNS misconfiguration, and it doesn't look like we need it for anything anyway"

cevich:
> Can we remove it?

Yes, drop it.  It was a host mirroring something we use to depend on at runtime, likely a repo. for a package install or a container image.

PR #8727, opened 2020-12-15
Cirrus: Implement containerized system testing

cevich:
Fixes https://github.com/containers/podman/pull/8714#issuecomment-744527897

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -67,6 +67,7 @@ gcp_credentials: ENCRYPTED[a28959877b2c9c36f151781b0a05407218cda646c7d047fc556e4
 ext_svc_check_task:
     alias: 'ext_svc_check'  # int. ref. name - required for depends_on reference
     name: "Ext. services"  # Displayed Title - has no other significance
+    skip: $CI == $CI
     # Default/small container image to execute tasks with
     container: &smallcontainer
         image: ${CTR_FQIN}
@@ -90,7 +91,7 @@ ext_svc_check_task:
 automation_task:
     alias: 'automation'
     name: "Check Automation"
-    skip: &branch "$CIRRUS_PR == '' || $CIRRUS_TAG != ''" # Don't run for branches
+    skip: $CI == $CI
     container: *smallcontainer
     env:
         TEST_FLAVOR: automation
@@ -107,6 +108,7 @@ automation_task:
 smoke_task:
     alias: 'smoke'
     name: "Smoke Test"
+    skip: $CI == $CI
     container: &bigcontainer
         image: ${CTR_FQIN}
         # Leave some resources for smallcontainer
@@ -132,6 +134,8 @@ smoke_task:
         fi
         cd $CIRRUS_WORKING_DIR
         make install.tools
+        make podman
+        make podman-remote
     setup_script: *setup
     main_script: *main
     always: *runner_stats
@@ -209,6 +213,7 @@ build_task:
 validate_task:
     name: "Validate $DISTRO_NV Build"
     alias: validate
+    skip: $CI == $CI
     depends_on:
         - ext_svc_check
         - automation
@@ -237,7 +242,7 @@ bindings_task:
     name: "Test Bindings"
     alias: bindings
     only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
-    skip: *branch
+    skip: $CI == $CI
     depends_on:
         - build
     gce_instance: *standardvm
@@ -256,6 +261,7 @@ bindings_task:
 swagger_task:
     name: "Test Swagger"
     alias: swagger
+    skip: $CI == $CI
     depends_on:
         - build
     container: *smallcontainer
@@ -275,6 +281,7 @@ swagger_task:
 vendor_task:
     name: "Test Vendoring"
     alias: vendor
+    skip: $CI == $CI
     depends_on:
         - build
     container: *smallcontainer
@@ -295,6 +302,7 @@ vendor_task:
 alt_build_task:
     name: "$ALT_NAME"
     alias: alt_build
+    skip: $CI == $CI
     only_if: *not_docs
     depends_on:
         - build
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

edsantiago:
This is going to be a tricky one. The `.cirrus.yml` changes LGTM, with one probably-trivial question. The hard part is going to be updating the container image to un-comment-out the `mount_program.*fuse-overlayfs` line in `/etc/containers/storage.conf` and then forcing `--cgroup-manager=cgroupfs` (to avoid "Error: cannot open sd-bus: No such file or directory: OCI runtime command not found error"). Then figuring out how to deal with "write unixgram @aefbb->/run/systemd/journal/socket: sendmsg: no such file or directory". And then, of course, running tests and seeing what fails. This is going to need to wait until next year.

.cirrus.yml
```diff
@@ -568,6 +569,18 @@ remote_system_test_task:
         PODBIN_NAME: remote
 
 
+container_system_test_task:
+    <<: *local_system_test_task
+    alias: container_system_test
+    skip: *branch
```
edsantiago, line 575: Trivial low-priority question: is this `*branch` deliberate? In looking at the Cirrus graph, I notice that this is the only one of the right-hand boxes (system tests) that has it set; all the others -- rootless, local, remote -- skip on CI:DOCS but not on BRANCH. I don't see a strong reason to go either way, I'm just asking because it's a curious inconsistency.
cirrus-map-pr8727

edsantiago: Is it just me, or is the graph beginning to look like a... um... cetacean?

cevich:
> ...<bunch of hard stuff>...This is going to need to wait until next year.

Or I ***COULD*** just close this PR :grin: 

TBH, this doesn't sound *too* awful.  The containerized tests are executed through a re-run of `setup_environment.sh` and `runner.sh` inside the container.  The hardest part will be me writing all the comments and not spelling something wrong :wink: 

>  is this *branch deliberate

Yes, I stuck it there because the containerized integration tests have it.  All it does is bypass containerized testing when a PR merges or a branch is (otherwise) pushed.  Long ago this was "okay-ed", since these tests are low-risk/low-value at the branch-level (assuming they passed PR testing).  TBH, we probably could skip a lot more testing on branches, that would improve the cron-based fail-rate as well as make Matt's life easier come release-time.

> Is it just me, or is the graph beginning to look like a... um... cetacean?

Are you're ocular orbs spherical today?  Clearly it's a Elephant standing in line at the McDonalds drive-through, during a McRib BOGO sale*


    * By one get one only valid when meal-deal is super-sized, and includes a super-extra-large-gulp milkshake

cevich:
Might as well get the ball rolling while waiting for other test results...

cevich:
...hmmm, well that's no good:  *gopath_cache* is failing to restore `/var/tmp/go`, but the keys are the correct values from the corresponding build jobs :thinking:

cevich:
```level=error msg="unable to write pod event: \"write unixgram @002ca->/run/systemd/journal/socket: sendmsg: no such file or directory\"```

@edsantiago this seems to be the main item left failing.  Could this be fixed by a bind-mount of that socket from the host, conditional on containerized execution of the system-tests??

**Comment by cevich** [comment]
now `Failed to add conmon to systemd sandbox cgroup: dial unix /run/systemd/private: connect: no such file or directory`

Hmmm...welp, I dislike sharing something called 'private', but let's try the same medicine there...

**Comment by cevich** [comment]
...hmmm, that doesn't appear to work.  With `JOURNALD_ARGS="-v /dev/log:/dev/log -v /var/run/systemd/journal/socket:/var/run/systemd/journal/socket -v /run/systemd/private:/run/systemd/private"` now we get:

```
level=warning msg="Failed to add conmon to systemd sandbox cgroup: Process with ID 65992 does not exist."
[+0205s] # Error: cannot open sd-bus: No such file or directory: OCI not found
```

**Comment by rhatdan** [comment]
I think you could change the events-backend  from journald to file.  I believe the issue is attempts to connect to systemd socket from inside of a container.

**Comment by vrothberg** [comment]
Is there a way to execute system tests without waiting for others to complete before?

**Comment by edsantiago** [comment]
No. @cevich started on a mechanism, #8754, which I believe adds undesirable complexity to the CI yaml for too little return.

**Comment by cevich** [comment]
> I think you could change the events-backend from journald to file.

Thanks @rhatdan trying it with `none` to see if there is any affect...

(I'm not sure where `file` goes, might be something we want to grab as an artifact for debugging).

**Comment by cevich** [comment]
> Is there a way to execute system tests without waiting for others to complete before?

The logic is way to complex since it also needs to not break `[CI:DOCS]`.  For future ref. @vrothberg:  Temporarily adding `skip: $CI == $CI` to all/most of the tasks is the easiest way.  Like in [the 'DO NOT MERGE' commit](https://github.com/containers/podman/pull/8727/commits/88f16e1d1c0c8789d5734d4a2156410358b75282) I made here.

**Comment by vrothberg** [comment]
Thanks!

**Comment by cevich** [comment]
@edsantiago would it be a PITA to modify the system-tests such that:

When `$container` is defined non-empty, always use `podman --events-backend=file` for every command?

I'm pretty sure the setting of $container is podman (maybe docker) built-in behavior, so it should be a reliable indicator.

**Comment by cevich** [comment]
(f4c89a6 is my naive attempt)

**Comment by edsantiago** [comment]
That's not going to work with podman-remote. Don't even bother.

**Comment by cevich** [comment]
:man_shrugging: CI is already "bothering"...but it told me it doesn't mind :smile:

**Comment by cevich** [comment]
@edsantiago hmmm, maybe was worth. Even with that option I still see errors like:

```
/var/tmp/go/src/github.com/containers/podman/bin/podman --events-backend=file --events-backend=file run --name running-u8m0JU3ooEKMsupVpGj6vjbdi7zgjI -d quay.io/libpod/testimage:20200929 top
[+0202s] # time="2020-12-18T14:21:25Z" level=warning msg="Failed to add conmon to systemd sandbox cgroup: dial unix /run/systemd/private: connect: no such file or directory"
[+0202s] # Error: cannot open sd-bus: No such file or directory: OCI not found
```

So it seems we're going to need a *much* more complicated solution :sweat:

cevich:
Giving up on this, it's too complicated to be properly implemented anytime soon.

PR #8729, opened 2020-12-15
Cirrus: Fix Disabled F32 rootless system testing

cevich:
This reverts commit 4dfb29a9af2ee87c492affe27264cf22535d1111.

Thanks to @edsantiago for the patch.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,11 +26,11 @@ env:
     ####
     FEDORA_NAME: "fedora-33"
     PRIOR_FEDORA_NAME: "fedora-32"
-    UBUNTU_NAME: "ubuntu-20"
-    PRIOR_UBUNTU_NAME: "ubuntu-19"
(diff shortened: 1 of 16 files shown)


cevich:
@edsantiago wrote:

> Looking at the logs... is this happening only on fedora? Not on ubuntu, even with cgroups v1? If so, the if condition should be changed to if is_cgroupsv1 && SOMETHING-SOMETHING-CHECK-FOR-FEDORA. That way we will have the default log level on ubuntu.

Correct, from the prior #8312 run, both are green:

* [sys podman ubuntu-2010 host](https://cirrus-ci.com/task/5301819599486976)
* [sys podman ubuntu-2004 host](https://cirrus-ci.com/task/6427719506329600)

cevich:
(waiting for initial run w/ suggested patch to verify before narrowing focus of `if is_cgroupsv1` conditional)

cevich:
Closing in favor of #8726

Issue #4207, opened 2019-10-06
Support "podman-remote cp"


sshnaidm:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Add support for `podman-remote` to copy files from/to containers.

**Describe the results you expected:**
```
podman-remote cp /local_path/file container:/path/file
podman-remote cp container:/path/file /local_path/file
```

**Additional information you deem important (e.g. issue happens only occasionally):**

It will unblock remote containers execution in ansible podman plugin: https://github.com/ansible/ansible/pull/63166

rhatdan:
This is still being worked and definitely something we want.

vrothberg:
> This is still being worked and definitely something we want.

@rhatdan, still valid with the new v2 API coming?

rhatdan:
Well we want it in V2 API.

vrothberg:
> Well we want it in V2 API.

Note that the request mentions varlink explicitly.

sshnaidm:
> > Well we want it in V2 API.
> 
> Note that the request mentions varlink explicitly.

I think you can easily remove "varlink" from the description. The main goal is to have it remotely, no matter with which API

vrothberg:
> I think you can easily remove "varlink" from the description. The main goal is to have it remotely, no matter with which API

Perfect! I did that and assigned @baude who is the driving force behind the new v2 API.

rhatdan:
Yes this will be done in V2.

afbjorklund:
I used the following workaround, for missing remote functionality:

`podman cp /foo container:/bar `--> `podman-remote exec -i container tee /bar < /foo`

Even if it is fixed in Podman 2.0, clients will be older 1.x for a while.

We will still use the regular "cp" command, when running locally.

rhatdan:
This is still being worked.

rhatdan:
@sshnaidm @afbjorklund @vrothberg @baude Where are we on this one?  We now have API V2 and podman cp should be using it.

afbjorklund:
Here is what I get, with podman-remote 2.0.6/podman 2.0.5:

```
 podman-remote cp /tmp/foo focused_knuth:foo
Error: unrecognized command `podman-remote cp`
Try 'podman-remote --help' for more information.
 podman --remote cp /tmp/foo focused_knuth:foo
Error: unrecognized command `podman cp`
Try 'podman --help' for more information.
```

[...]

PR #8747, opened 2020-12-16
remote copy

vrothberg:
Implement `podman-remote cp` and break out the logic from the previously
added `pkg/copy` into it's basic building blocks and move them up into
the `ContainerEngine` interface and `cmd/podman`.

The `--pause` flag is now deprecated and turned into a nop.

Note that this commit is vendoring a non-release version of Buildah to
pull in updates to the copier package.


vrothberg:
One remaining bug that I will debug with Nalin.

cmd/podman/containers/cp.go
```diff
@@ -1,16 +1,25 @@
...
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 var (
-	cpDescription = `Command copies the contents of SRC_PATH to the DEST_PATH.
+	cpDescription = `Copy the contents of SRC_PATH to the DEST_PATH.
 
-  You can copy from the container's file system to the local machine or the reverse, from the local filesystem to the container. If "-" is specified for either the SRC_PATH or DEST_PATH, you can also stream a tar archive from STDIN or to STDOUT. The CONTAINER can be a running or stopped container.  The SRC_PATH or DEST_PATH can be a file or directory.
+  You can copy from the container's file system to the local machine or the reverse, from the local filesystem to the container.  If "-" is specified for either the SRC_PATH or DEST_PATH, you can also stream a tar archive from STDIN or to STDOUT.  The CONTAINER can be a running or stopped container.  The SRC_PATH or DEST_PATH can be a file or a directory.
```
edsantiago, line 22: It is no longer acceptable to double-space after periods. (It's a habit that has taken me a long time to break, but I accept the justification).

rhatdan: I though this was just those of us that have actually used a typewriter.

vrothberg: I am too old to change, @edsantiago

vrothberg: > I thought this was just those of us that have actually used a typewriter.

I learned on a typewriter :)  I always use double spaces for text and now my brain parser is painfully dependent on it.

pkg/domain/infra/abi/containers_stat.go
```diff
@@ -0,0 +1,245 @@
...
+		// join the working dir with "/", and the add the containerPath
+		// to it.
+		pathRelativeToContainerMountPoint = filepath.Join(filepath.Join("/", container.WorkingDir()), containerPath)
+	}
+	// NOTE: the secure join makes sure that we follow symlinks.  This way,
+	// we catch scenarios where the container path symlinks to a volume or
+	// bind mount.
+	// FIXME
+	resolvedPathOnTheContainerMountPoint := filepath.Join(mountPoint, pathRelativeToContainerMountPoint)
+	//	if err != nil {
```
rhatdan, line 118: Why is this code here?

rhatdan (approved):
Impressive work

pkg/domain/infra/abi/containers_stat.go
vrothberg: An old artifact, thanks for catching!

cmd/podman/containers/cp.go
```diff
@@ -40,24 +55,306 @@ var (
...
+		if !containerInfo.IsDir {
+			target = filepath.Dir(target)
+		}
+
+		copyFunc, err := registry.ContainerEngine().ContainerCopyFromArchive(registry.GetContext(), container, target, reader)
+		if err != nil {
+			return err
+		}
+		if err := copyFunc(); err != nil {
+			return errors.Wrap(err, "error copying to container")
```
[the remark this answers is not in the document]
vrothberg: The terms can be confusing. We are copying "to" a container "from" a tar archive. So the error message is correct.

cmd/podman/containers/cp.go
```diff
@@ -40,24 +55,316 @@ var (
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Extract the tar file into the destination directory.")
-	flags.BoolVar(&cpOpts.Pause, "pause", true, "Pause the container while copying")
+	flags.BoolVar(&cpOpts.Pause, "pause", true, "Pause the container while copying (deprecated)")
+	_ = flags.MarkHidden("pause")
```
Luap99, line 59: Is there any reason why you did not use `flags.MarkDeprecated`

vrothberg: I forgot its existence :)  If I need another push, I'll fix it :+1:

rhatdan: Does MarkDeprecated hide the option?  I don't want users to see the option. Just want it to work so that there scripts would continue to work.

vrothberg: Yes, that will hide the option but note that it's already hidden.

Luap99: Yes it is hidden. If you still use it the message you specify with `MarkDeprecated` is emitted to stderr.
See `podman image diff --archive alpine > /dev/null` for example.

test/system/065-cp.bats
```diff
@@ -99,18 +94,19 @@ load helpers
...
     # Test extract on non archive.
     run_podman cp --extract=true $srcdir/$rand_filename cpcontainer:/foo.txt
+    run_podman exec cpcontainer cat /foo.txt
+    is "$output" "$rand_content"
 
-    # Cannot extract an archive to a file!
-    run_podman 125 cp --extract=true $tar_file cpcontainer:/tmp/file.txt
-    is "$output" 'Error: cannot extract archive .* to file "/tmp/file.txt"'
+    # Extract an archive to a file.
+    run_podman cp --extract=true $tar_file cpcontainer:/tmp/file.txt
```
edsantiago, line 101: This is a useless test:  `/tmp/$dirname` has already been created (lines 91-92 above).

Worse, I find this command and comment confusing. What does it mean to "extract an archive to a file"? I would expect this to fail, and don't understand why it doesn't.

By adding the `exec` below I can see that podman seems to be stripping off `file.txt` and extracting into the parent directory. I think this is counterintuitive, and even dare say it is wrong behavior; but that's not my decision to make. If it is correct, or if it's impossible to fix, I think the test should be amended in some way (either with the `exec rm` or, preferably, with an extraction into a different directory) and, also, the comment should be expanded to clarify this weird behavior.

```diff
diff --git a/test/system/065-cp.bats b/test/system/065-cp.bats
index b1c3fbf8a..db9fbfff6 100644
--- a/test/system/065-cp.bats
+++ b/test/system/065-cp.bats
@@ -98,6 +98,7 @@ load helpers
     is "$output" "$rand_content"

     # Extract an archive to a file.
+    run_podman exec cpcontainer rm -rf /tmp/$dirname
     run_podman cp --extract=true $tar_file cpcontainer:/tmp/file.txt
     run_podman exec cpcontainer cat /tmp/$dirname/$rand_filename
     is "$output" "$rand_content"
```

edsantiago (commented):
LGTM, nice work, but there's one broken test and very weird (unexpected) behavior.

cmd/podman/containers/cp.go
vrothberg: Actually, using `MarkDeprecated(...)` is pretty ugly as it prints the warning twice for rootless podman (due to re-exec):
```
podman (run-950) $ ./bin/podman cp --pause=false Makefile test:/
Flag --pause has been deprecated, pause has been deprecated and and has no effect
Flag --pause has been deprecated, pause has been deprecated and and has no effect
```

vrothberg:
Repushed.  Now both, --pause and --export, are deprecated and turned into nops.  I didn't use cobra's `MarkDeprecated` as it's printing the warning twice for rootless podman due to the reexec.

cmd/podman/containers/cp.go
```diff
@@ -39,25 +54,312 @@ var (
...
+func containerMustExist(container string) error {
+	exists, err := registry.ContainerEngine().ContainerExists(registry.GetContext(), container, entities.ContainerExistsOptions{})
+	if err != nil {
+		return err
+	}
+	if !exists.Value {
+		return errors.Errorf("container %q does not exist", container)
+	}
+	return nil
+}
```
TomSweeneyRedHat, line 103: Maybe I have the exists() function from the new API stuck in my head, but I thought we'd similar functionality elsewhere in libpod?  I looked but didn't find it quickly.  This feels like a function that should live in util.

This is fine here for now, but something to ruminate on.

TomSweeneyRedHat:
Changes LGTM
One question for consideration and unfortunately a lot of red in the tests.

cmd/podman/containers/cp.go
vrothberg: Yes, there's a ContainerExists function.  I thought to put that into  a dedicated function (plus the checks) to consolidate the logic a bit further.

vrothberg:
> Changes LGTM

Thanks for checking!

> One question for consideration and unfortunately a lot of red in the tests.

Should turn green now.  Removing the options made the help checks fail (as there are no options any more).

rhatdan:
/lgtm
/hold
Go on vacation now...

vrothberg:
> Go on vacation now...

One more day to go :)

vrothberg:
One more time ... Ubuntu remote tests failed.  Rebased.

vrothberg:
Note to self: close https://github.com/containers/podman/issues/4207 once this is merged

vrothberg:
Tried to reproduce in the Google Cloud but it worked :hear_no_evil: Let's pray to the CI gods.

vrothberg:
Oh man, that's frustrating.

baude:
congrats @vrothberg

Issue #8680, opened 2020-12-10
"toolbox reset" nukes files that don't belong to it


hadess:
From https://github.com/containers/toolbox/issues/648:

Running `toolbox reset` (which just calls `podman system reset`) just nuked a whole bunch of files that didn't belong to it under /run/user/1000 including the session D-Bus socket. Here's me really hoping that it didn't nuke any other files that might have been linked from there. Note that it will likely nuke files under `/run/user/1000/doc/` which are files exported by the xdg-desktop-portal document portal for use with Flatpak. Deleting files there deletes the actual files that users used in the Flatpak containers.

```sh
$ find /run/user/1000/
/run/user/1000/
/run/user/1000/gnupg
/run/user/1000/gnome-shell-disable-extensions
/run/user/1000/gnome-shell
/run/user/1000/gnome-shell/runtime-state-LE.:0
/run/user/1000/wayland-0
/run/user/1000/wayland-0.lock
/run/user/1000/.mutter-Xwaylandauth.XN0MV0
/run/user/1000/ICEauthority
/run/user/1000/gnome-session-leader-fifo
/run/user/1000/dconf
/run/user/1000/dconf/user
/run/user/1000/gvfs
/run/user/1000/keyring
/run/user/1000/keyring/ssh
/run/user/1000/keyring/pkcs11
/run/user/1000/keyring/control
/run/user/1000/pulse
/run/user/1000/pulse/pid
/run/user/1000/pulse/native
/run/user/1000/pipewire-0
/run/user/1000/bus
/run/user/1000/systemd
/run/user/1000/systemd/transient
/run/user/1000/systemd/transient/vte-spawn-8f341ea3-9205-4427-aeeb-6b4093fdb785.scope
/run/user/1000/systemd/transient/dbus-:1.2-org.gnome.Calendar@0.service
/run/user/1000/systemd/transient/dbus-:1.2-org.gnome.Characters.BackgroundService@0.service
/run/user/1000/systemd/transient/dbus-:1.2-org.gnome.Calculator.SearchProvider@0.service
/run/user/1000/systemd/transient/dbus-:1.2-org.gnome.ControlCenter.SearchProvider@0.service
/run/user/1000/systemd/transient/dbus-:1.2-org.freedesktop.portal.IBus@1.service
/run/user/1000/systemd/transient/app-gnome-gnome\x2dsoftware\x2dservice-2095.scope
[...]

PR #8750, opened 2020-12-16
Don't accidently remove XDG_RUNTIME_DIR when reseting storage

rhatdan:
In certain cases XDG_RUNTIME_DIR was deleted by accident based on
settings in the storage.conf. This patch verifies that when doing
a storage reset, we don't accidently remove XDG_RUNTIME_DIR.

Fixes: https://github.com/containers/podman/issues/8680

### DIFF
--- a/libpod/reset.go
+++ b/libpod/reset.go
@@ -77,18 +77,35 @@ func (r *Runtime) Reset(ctx context.Context) error {
 		}
 	}
 
+	xdgRuntimeDir := filepath.Clean(os.Getenv("XDG_RUNTIME_DIR"))
 	_, prevError := r.store.Shutdown(true)
-	if err := os.RemoveAll(r.store.GraphRoot()); err != nil {
+	graphRoot := filepath.Clean(r.store.GraphRoot())
+	if graphRoot == xdgRuntimeDir {
 		if prevError != nil {
 			logrus.Error(prevError)
 		}
-		prevError = err
+		prevError = errors.Errorf("failed to remove runtime graph root dir %s, since it is the same as XDG_RUNTIME_DIR", graphRoot)
+	} else {
+		if err := os.RemoveAll(graphRoot); err != nil {
+			if prevError != nil {
+				logrus.Error(prevError)
+			}
+			prevError = err
+		}
 	}
-	if err := os.RemoveAll(r.store.RunRoot()); err != nil {
+	runRoot := filepath.Clean(r.store.RunRoot())
+	if runRoot == xdgRuntimeDir {
 		if prevError != nil {
 			logrus.Error(prevError)
 		}
-		prevError = err
+		prevError = errors.Errorf("failed to remove runtime root dir %s, since it is the same as XDG_RUNTIME_DIR", runRoot)
+	} else {
+		if err := os.RemoveAll(runRoot); err != nil {
+			if prevError != nil {
+				logrus.Error(prevError)
+			}
+			prevError = err
+		}
 	}
 	runtimeDir, err := util.GetRuntimeDir()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


libpod/reset.go
```diff
@@ -77,18 +77,35 @@ func (r *Runtime) Reset(ctx context.Context) error {
 		}
 	}
 
+	xdgRuntimeDir := os.Getenv("XDG_RUNTIME_DIR")
 	_, prevError := r.store.Shutdown(true)
-	if err := os.RemoveAll(r.store.GraphRoot()); err != nil {
+	graphRoot := r.store.GraphRoot()
+	if graphRoot == xdgRuntimeDir {
```
hadess, line 83: I'm not very familiar with Go, so is that a string equality, or does it handle minor differences between paths as they might have been defined, eg. with or without a trailing `/`.

libpod/reset.go
```diff
@@ -77,18 +77,35 @@ func (r *Runtime) Reset(ctx context.Context) error {
 		}
 	}
 
+	xdgRuntimeDir := os.Getenv("XDG_RUNTIME_DIR")
```
hadess, line 80: The documentation `XDG_RUNTIME_DIR` says:
```
 If $XDG_RUNTIME_DIR is not set applications should fall back to a replacement directory with similar capabilities and print a warning message. Applications should use this directory for communication and synchronization purposes and should not place larger files in it, since it might reside in runtime memory and cannot necessarily be swapped out to disk.
```

So this should probably fallback to `/run/user/<uid>`.

libpod/reset.go
rhatdan: Just string equality.

libpod/reset.go
```diff
@@ -77,18 +77,35 @@ func (r *Runtime) Reset(ctx context.Context) error {
...
+		if err := os.RemoveAll(graphRoot); err != nil {
+			if prevError != nil {
+				logrus.Error(prevError)
+			}
+			prevError = err
+		}
 	}
-	if err := os.RemoveAll(r.store.RunRoot()); err != nil {
+	runRoot := r.store.RunRoot()
+	if runRoot == xdgRuntimeDir {
```
hadess, line 97: It might be worth checking whether the "run root" is correctly prefixed, if it is a child of xdgRuntimeDir.

rhatdan: Why would that matter?

hadess: It matters if you can only do string comparisons. The new default configuration is `$XDG_RUNTIME_DIR/overlays`, it used to be `$XDG_RUNTIME_DIR`. You still don't want to nuke `$XDG_RUNTIME_DIR/`.

rhatdan: Right the case above will only remove $XDG_RUNTIME_DIR/overlays which is what we want.  It really does not mapper if XDG_RUNTIME_DIR/overlays or /random/path/overlays

libpod/reset.go
```diff
@@ -98,13 +115,19 @@ func (r *Runtime) Reset(ctx context.Context) error {
 	if tempDir == runtimeDir {
 		tempDir = filepath.Join(tempDir, "containers")
 	}
-	if err := os.RemoveAll(tempDir); err != nil {
+	if filepath.Clean(tempDir) == xdgRuntimeDir {
 		if prevError != nil {
 			logrus.Error(prevError)
 		}
-		prevError = err
+		prevError = errors.Errorf("failed to remove runtime tmpdir %s, since it is the same XDG_RUNTIME_DIR", tempDir)
```
TomSweeneyRedHat, line 122: nit
```suggestion
		prevError = errors.Errorf("failed to remove runtime tmpdir %s, since it is the same as XDG_RUNTIME_DIR", tempDir)
```

libpod/reset.go
```diff
@@ -77,18 +77,35 @@ func (r *Runtime) Reset(ctx context.Context) error {
...
+		}
 	}
-	if err := os.RemoveAll(r.store.RunRoot()); err != nil {
+	runRoot := filepath.Clean(r.store.RunRoot())
+	if runRoot == xdgRuntimeDir {
 		if prevError != nil {
 			logrus.Error(prevError)
 		}
-		prevError = err
+		prevError = errors.Errorf("failed to remove runtime root dir %s, since it is the same XDG_RUNTIME_DIR", runRoot)
```
TomSweeneyRedHat, line 101:
```suggestion
		prevError = errors.Errorf("failed to remove runtime root dir %s, since it is the same as XDG_RUNTIME_DIR", runRoot)
```

TomSweeneyRedHat:
Couple small nits, LGTM once cleared.

libpod/reset.go
rhatdan: The filepath.Clean() will cleanup extra "/" and other cruft in the paths.

TomSweeneyRedHat (commented):
LGTM
I like the filepath clean bit, will have to try and remember that for the future.  Good call @hadess

PR #8754, opened 2020-12-16
Cirrus: Add [CI:SYS] magic to focus on system tests

cevich:
Since the system tests normally execute almost at the end of the chain,
it's difficult to develop/debug them.  Add support for a magic string,
when appearing in the title, skips through most other testing.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -236,7 +236,7 @@ validate_task:
 bindings_task:
     name: "Test Bindings"
     alias: bindings
-    only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
+    only_if: &not_docs_sys $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' && $CIRRUS_CHANGE_TITLE !=~ '.*CI:SYS.*'
     skip: *branch
     depends_on:
         - build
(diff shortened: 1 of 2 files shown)


edsantiago:
Who is the target audience for this, and is it worth the increase in complexity? I'm having a really hard time grokking the `not_doc_sys` term, and also the logic. I find myself repeatedly looking back up to check the expression, and struggling with it each time. My brain simply does not handle negative logic (skip if this and that) well, and I don't think I'm atypical.

I understand that you have a particular problem you want to resolve in the short term, with containerized tests, but can you convince me that this complication is going to be so useful in the long term that it outweighs the lack of readability and maintainability in `.cirrus.yml`?

cevich:
> I understand that you have a particular problem you want to resolve in the short term, with containerized tests, but can you convince me that this complication is going to be so useful in the long term that it outweighs the lack of readability and maintainability

Maybe?  This is why I split it out into a separate PR for more testing/beating on (notice the PR is a "draft").  My intentions/audience is very simple:

    Help you or a developer that only needs to tweak a system-test, avoid needing 
    to wait an hour for all the integration tests to complete first.

The negative logic really hurts my brain also...but is unavoidable due to the way Cirrus-CI behaves WRT the 'skip' and 'only' options.  Worse...there's actually a really tricky bug introduced here that breaks the system tests when `[CI:SYS]` is enabled :disappointed: 

So, I was leaning toward working more on this...but I really don't want to debug the caching problem it creates.

PR #8774, opened 2020-12-18
[CI:DOCS] Document location of backend events file

cevich:
It may be useful to know the general location of this file, for backup/archival purposes.

### DIFF
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -44,7 +44,9 @@ Path of the conmon binary (Default path is configured in `containers.conf`)
 
 #### **--events-backend**=*type*
 
-Backend to use for storing events. Allowed values are **file**, **journald**, and **none**.
+Backend to use for storing events. Allowed values are **file**, **journald**, and
+**none**. When *file* is specified, the events are stored under a subdirectory
+of the *tmpdir* location (see **--tmpdir** below).
 
 #### **--help**, **-h**
 


cevich:
Oof, I edited `podman.1` which seems to have been a bad thing :disappointed:

docs/source/markdown/podman.1.md
```diff
@@ -44,7 +44,11 @@ Path of the conmon binary (Default path is configured in `containers.conf`)
 
 #### **--events-backend**=*type*
 
-Backend to use for storing events. Allowed values are **file**, **journald**, and **none**.
+Backend to use for storing events. Allowed values are **file**, **journald**, and
+**none**. When *file* is specified, the events are stored under the
+**$XDG_RUNTIME_DIR/libpod** tree. The exact sub-directory/location may vary
+depending on podman being run as root, or not.  Accessing this file while podman
```
vrothberg, line 50:
```suggestion
depending on Podman being run as root, or not.  Accessing this file while Podman
```

docs/source/markdown/podman.1.md
```diff
@@ -44,7 +44,11 @@ Path of the conmon binary (Default path is configured in `containers.conf`)
 
 #### **--events-backend**=*type*
 
-Backend to use for storing events. Allowed values are **file**, **journald**, and **none**.
+Backend to use for storing events. Allowed values are **file**, **journald**, and
+**none**. When *file* is specified, the events are stored under the
+**$XDG_RUNTIME_DIR/libpod** tree. The exact sub-directory/location may vary
```
mheon, line 49: This is not correct for Podman as root - we default to `/var/run/libpod` in that case.

mheon: A better way to word this would be "events are stored under the Podman temporary files directory (`tmpdir`)

cevich: okay, thanks.  My concern then is, if someone is specifying `file`, there's a good chance they will (eventually) be concerned about where the file is (or will be).  Especially since the definition changes for root vs rootless.

Perhaps, is "The podman temporary directory" defined someplace else in the docs?  If so, I would feel more comfortable simply referring to it as you suggest.

cevich: ...wait, is this the same `tmpdir` as used/described for the `--tmpdir` option?

cevich:
How about that?

mheon:
That works fine.

PR #8775, opened 2020-12-18
podman v3 container bindings

baude:
convert the golang container bindings to all use options so that changes
in the future are more managable.


vrothberg (commented):
LGTM

PR #8781, opened 2020-12-18
Add support for checkpoint/restore of containers with volumes

rst0git:
When migrating a container with volumes, the content of the volumes is required to be available on the destination machine.

This PR enables checkpoint/restore of containers with named volumes by including the content of volumes in checkpoint file. When restoring a container, the associated volumes are recreated and their content is restored accordingly.

In addition, this PR adds `--ignore-volumes` option to disable this feature.

Examples:

In the following example the content of associated volumes is included in checkpoint tar.gz file.
```console
# podman container checkpoint --export <checkpoint>.tar.gz <container>
```

This example doesn't include the content of volumes in the tar.gz file. This can be used to avoid unnecessary I/O operations when running checkpoint and restore on the same host.
```console
# podman container checkpoint --ignore-volumes --export <checkpoint>.tar.gz <container>
```

In the following example the container volumes will be recreated and their content restored. Note that if volumes with the same name already exist on the system, or the content of volumes is missing in the tar.gz, Podman will exit with an error.
```console
# podman container restore --import <checkpoint>.tar.gz
```

Volumes associated with container must already exist. Podman will not create them or to restore their content.
```console
# podman container restore --ignore-volumes --import <checkpoint>.tar.gz
```

Resolves https://github.com/checkpoint-restore/criu/issues/1314 https://github.com/checkpoint-restore/criu/issues/826

### DIFF
--- a/docs/source/markdown/podman-container-checkpoint.1.md
+++ b/docs/source/markdown/podman-container-checkpoint.1.md
@@ -52,6 +52,12 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
 to explicitly disable including changes to the root file-system into
 the checkpoint archive file.
 
+#### **--ignore-volumes**
+
+This option must be used in combination with the **--export, -e** option.
+When this option is specified, the content of volumes associated with
+the container will not be included into the checkpoint tar.gz file.
+
 ## EXAMPLE
 
 podman container checkpoint mywebserver
--- a/docs/source/markdown/podman-container-restore.1.md
+++ b/docs/source/markdown/podman-container-restore.1.md
@@ -85,6 +85,13 @@ exported checkpoint with **--name, -n**.
 
 Using **--ignore-static-mac** tells Podman to ignore the MAC address if it was
 configured with **--mac-address** during container creation.
+
+#### **--ignore-volumes**
+
+This option must be used in combination with the **--import, -i** option.
+When restoring containers from a checkpoint tar.gz file with this option,
+the content of associated volumes will not be restored.
+
 ## EXAMPLE
 
 podman container restore mywebserver
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -791,8 +791,8 @@ func (c *Container) addNamespaceContainer(g *generate.Generator, ns LinuxNS, ctr
 }
 
 func (c *Container) exportCheckpoint(options ContainerCheckpointOptions) error {
-	if (len(c.config.NamedVolumes) > 0) || (len(c.Dependencies()) > 0) {
-		return errors.Errorf("Cannot export checkpoints of containers with named volumes or dependencies")
+	if len(c.Dependencies()) > 0 {
+		return errors.Errorf("Cannot export checkpoints of containers with dependencies")
 	}
 	logrus.Debugf("Exporting checkpoint image of container %q to %q", c.ID(), options.TargetFile)
 
@@ -870,6 +870,47 @@ func (c *Container) exportCheckpoint(options ContainerCheckpointOptions) error {
 		}
 	}
 
+	// Folder containing archived volumes that will be included in the export
+	expVolDir := filepath.Join(c.bundlePath(), "volumes")
+
+	// Create an archive for each volume associated with the container
+	if !options.IgnoreVolumes {
+		if err := os.MkdirAll(expVolDir, 0700); err != nil {
+			return errors.Wrapf(err, "error creating volumes export directory %q", expVolDir)
+		}
+
+		for _, v := range c.config.NamedVolumes {
+			volumeTarFilePath := filepath.Join("volumes", v.Name+".tar")
+			volumeTarFileFullPath := filepath.Join(c.bundlePath(), volumeTarFilePath)
+
+			volumeTarFile, err := os.Create(volumeTarFileFullPath)
+			if err != nil {
+				return errors.Wrapf(err, "error creating %q", volumeTarFileFullPath)
+			}
+
+			volume, err := c.runtime.GetVolume(v.Name)
+			if err != nil {
+				return err
+			}
+
+			input, err := archive.TarWithOptions(volume.MountPoint(), &archive.TarOptions{
+				Compression:      archive.Uncompressed,
+				IncludeSourceDir: true,
+			})
+			if err != nil {
+				return errors.Wrapf(err, "error reading volume directory %q", v.Dest)
+			}
+
+			_, err = io.Copy(volumeTarFile, input)
+			if err != nil {
+				return err
+			}
+			volumeTarFile.Close()
+
+			includeFiles = append(includeFiles, volumeTarFilePath)
+		}
+	}
+
 	input, err := archive.TarWithOptions(c.bundlePath(), &archive.TarOptions{
 		Compression:      archive.Gzip,
 		IncludeSourceDir: true,
@@ -897,6 +938,10 @@ func (c *Container) exportCheckpoint(options ContainerCheckpointOptions) error {
 
 	os.Remove(rootfsDiffPath)
 	os.Remove(deleteFilesList)
+
+	if !options.IgnoreVolumes {
+		os.RemoveAll(expVolDir)
+	}
 
 	return nil
 }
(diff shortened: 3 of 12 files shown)


rst0git:
cc @adrianreber

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -52,6 +52,12 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
 to explicitly disable including changes to the root file-system into
 the checkpoint archive file.
 
+#### **ignore-volumes**
+
+When this option is specified, the content of volumes associated with container
```
adrianreber, line 57: ... with **the** container... ?

docs/source/markdown/podman-container-restore.1.md
```diff
@@ -85,6 +85,12 @@ exported checkpoint with **--name, -n**.
 
 Using **--ignore-static-mac** tells Podman to ignore the MAC address if it was
 configured with **--mac-address** during container creation.
+
+#### **--ignore-volumes**
+
+When restoring container from a checkpoint tar.gz file, do not restore the content of
```
adrianreber, line 91: When restoring container**s**

pkg/domain/infra/abi/containers.go
```diff
@@ -540,7 +542,7 @@ func (ic *ContainerEngine) ContainerRestore(ctx context.Context, namesOrIds []st
 
 	switch {
 	case options.Import != "":
-		cons, err = checkpoint.CRImportCheckpoint(ctx, ic.Libpod, options.Import, options.Name)
+		cons, err = checkpoint.CRImportCheckpoint(ctx, ic.Libpod, options.Import, options.Name, options.IgnoreVolumes)
```
adrianreber, line 545: In exportCheckpoint() you changed the function to only have the `options` parameter instead of one parameter for each used option. I guess you could to the same here.

adrianreber:
@rstgit Thanks a lot. Looks correct. You should probably add a test for this.

In the output of '--help' you also just talk about 'volumes'. Not sure if we need to mention 'named volumes' at those places.

What about the shell completion files? Are they automatically created or do they also need updating?

libpod/container_internal_linux.go
```diff
@@ -1170,6 +1215,29 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
 		return err
 	}
 
+	// Restore the content of named volumes. The volumes were created in setupContainer().
+	if !options.IgnoreVolumes {
+		for _, v := range c.config.NamedVolumes {
```
mheon, line 1220: This is not safe. We need to make sure that the volume is empty before we start copying into it - we've had CVEs about this before.

docs/source/markdown/podman-container-checkpoint.1.md
rst0git: Fixed, thanks!

docs/source/markdown/podman-container-restore.1.md
rst0git: Fixed, thanks!

pkg/domain/infra/abi/containers.go
rst0git: Thank you for suggesting that. I have applied this change in an additional commit.

rst0git:
Hi @adrianreber,

> You should probably add a test for this.

I've updated the PR to include a test case.
                                                                                                                                                      
> In the output of '--help' you also just talk about 'volumes'. Not sure if we need to mention 'named volumes' at those places.                       
                                                                                                                                                      
Thank you for pointing this out, I should have made it clearer. The initial goal was to extend the checkpoint/restore functionality with support for containers such as ['mysql'](https://github.com/docker-library/mysql/blob/1612c574f921f7f1afc13763ce9011fd8e130b19/8.0/Dockerfile.debian#L87) that use the [VOLUME](https://docs.docker.com/engine/reference/builder/#volume) instruction. This instruction creates a volume with specified name that must be unique on the host. However, such volumes are also described in the documentation as ["anonymously named"](http://docs.podman.io/en/latest/markdown/podman-create.1.html#volume-v-source-volume-host-dir-container-dir-options) because they have randomly generated name and will be removed when the container is removed. In addition, when volumes are created with *names* they are not *anonymous*, and they are not removed by the `--rm` option.                                                                                                                              
                                                                                                                                                      
The changes in this PR extend Podman to be able to checkpoint/restore *named*, *anonymously named*, and *anonymous* volumes.

To avoid unnecessary confusion, I used only the word "volumes" in the output of `--help`, but I'm open to suggestions that would make it easier to understand.

> What about the shell completion files? Are they automatically created or do they also need updating?

Yes, the shell completions appear to process the output of `--help`, hence they are updated automatically.

libpod/container_internal_linux.go
rst0git: Hi Matthew, thank you for your comment. I assume that you are referring to [CVE-2020-1726](https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2020-1726), which was fixed in https://github.com/containers/podman/commit/c140ecdc9b416ab4efd4d21d14acd63b6adbdd42.

Please correct me if I'm wrong, but I believe restoring container volumes is slightly different because we need to ensure that all files in a volume are "exactly" the same as in the time when the checkpoint was created. Otherwise CRIU is likely [exit with an error](https://criu.org/Validate_files_on_restore).

Thus, if a volume already exists Podman, would exit and suggest to the user to use the `--ignore-volumes` option. When this option is used, Podman will not copy any data to any volumes.

The only case when we are copying data into a volume is when the volume has been just created, and therefore empty.

mheon: Are you sure about this? I see no guarantees that we won't use existing volumes, overwriting their contents in the process. Podman will certainly create the volumes if they do not exist, but there is no protection for already-existing volumes.

rst0git: > Are you sure about this? I see no guarantees that we won't use existing volumes

I have introduced the following condition with this PR.
Would that be sufficient to protect overwriting the contents of already-existing volumes?
```diff
diff --git a/pkg/checkpoint/checkpoint_restore.go b/pkg/checkpoint/checkpoint_restore.go
index 90f629354..f6cd3b38f 100644
--- a/pkg/checkpoint/checkpoint_restore.go
+++ b/pkg/checkpoint/checkpoint_restore.go
@@ -54,6 +54,7 @@ func CRImportCheckpoint(ctx context.Context, runtime *libpod.Runtime, restoreOpt
                        "rootfs-diff.tar",
                        "network.status",
                        "deleted.files",
+                       "volumes",
                },
        }
        dir, err := ioutil.TempDir("", "checkpoint")
@@ -83,8 +84,21 @@ func CRImportCheckpoint(ctx context.Context, runtime *libpod.Runtime, restoreOpt
        }
 
        // This should not happen as checkpoints with these options are not exported.
-       if (len(config.Dependencies) > 0) || (len(config.NamedVolumes) > 0) {
-               return nil, errors.Errorf("Cannot import checkpoints of containers with named volumes or dependencies")
+       if len(config.Dependencies) > 0 {
+               return nil, errors.Errorf("Cannot import checkpoints of containers with dependencies")
+       }
+
+       // Volumes included in the checkpoint should not exist
+       if !restoreOptions.IgnoreVolumes {
+               for _, vol := range config.NamedVolumes {
+                       exists, err := runtime.HasVolume(vol.Name)
+                       if err != nil {
+                               return nil, err
+                       }
+                       if exists {
+                               return nil, errors.Errorf("volume with name %s already exists. Use --ignore-volumes to not restore content of volumes", vol.Name)
+                       }
+               }
        }
 
        ctrID := config.ID
```

mheon: Yes, such a check would be sufficient.

rst0git: Thank you Matthew!

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -52,6 +52,12 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
 to explicitly disable including changes to the root file-system into
 the checkpoint archive file.
 
+#### **--ignore-volumes**
+
+This option is available only in combination with **--export, -e**.
```
TomSweeneyRedHat, line 57: Soft suggest as it's really always "available", just not usable...
```suggestion
This option must be used in combination with the **--export, -e** option.
```

docs/source/markdown/podman-container-restore.1.md
```diff
@@ -85,6 +85,13 @@ exported checkpoint with **--name, -n**.
 
 Using **--ignore-static-mac** tells Podman to ignore the MAC address if it was
 configured with **--mac-address** during container creation.
+
+#### **--ignore-volumes**
+
+This option is available only in combination with **--import, -i**. When restoring
```
TomSweeneyRedHat, line 91: same comment as export

rhatdan:
@adrianreber PTAL

adrianreber:
Looks good. Thanks!

mheon:
LGTM

Issue #8512, opened 2020-11-30
Podman ps don't accept --filter pod


PavelSosin-320:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman ps and Podman container ls commands don't accept --filter options
 #https://github.com/containers/podman/pull/1345
**Steps to reproduce the issue:**
1. Create named Pod
2. Run image and assign it to Pod using Podman run --pod podName .... imageName command 
3. Start pod podName
4. use Podman pod ls to see that the pod is running
6. Run Podman ps or Podman container ls --pod command to see that the container is running and is assigned to the pod
7. run Podman ps or Podman container ls with ---filter or --filter "pod=podname" option to see only containers assigned to the podName pod.
**Describe the results you received:**
Error: pod is an invalid filter regardless of used option format --filter=pod 123 --filter "pod=123", etc
**Describe the results you expected:**
List and status of containers running in the pod name pod following in the documentation and 
Tracking bug #https://github.com/containers/podman/pull/1345
**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.1.1
```

**Output of `podman info --debug`:**

```
podman info --debug
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.20, commit: '
  cpus: 4
  **distribution:
    distribution: Ubuntu
    version: "20.04"**
  eventLogger: journald
  hostname: MSI
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.19.128-microsoft-standard
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
apt list podman
Listing... Done
podman/now 2.1.1~2 amd64 [installed,local]
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

No, I suppose 2.2 must be available

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Ubuntu 20.04 distro with Kube Kubernetes Microk8c package installed and systemd is enabled.
sudo apt-get upgrade podman
Reading package lists... Done
Building dependency tree
Reading state information... Done
**podman is already the newest version (2.1.1~2)**

Luap99:
[...]

PR #8804, opened 2020-12-22
add pod filter for ps

baude:
adds the ability to filter containers based on the filter "pod".  the
value can be a pod name or its full or partial id.

Fixes: #8512

### DIFF
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -690,6 +690,7 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//        - `label`=(`key` or `"key=value"`) of an container label
 	//        - `name=<name>` a container's name
 	//        - `network`=(`<network id>` or `<network name>`)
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

pkg/api/server/register_containers.go
```diff
@@ -72,6 +72,7 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//        - name=<name> a container's name
 	//        - network=(<network id> or <network name>)
 	//        - publish=(<port>[/<proto>]|<startport-endport>/[<proto>])
+	//        - pod=(<pod id> or <pod name>
```
Luap99, line 75: This is only added for compat but not the libpod endpoint.

Luap99 (commented):
One nit otherwise LGTM

pkg/api/server/register_containers.go
baude: u are right.  fixed!

rhatdan:
CI System seems to be hung.  Forcing merge.

Issue #8384, opened 2020-11-18
add configuration option for disabling new keyring


gzm55:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

In a rootless container with lean capabilities, to create a container by `podman run`, we have problems of no permissions to call pivot_root and create new keyring. In containers.conf, we could disable pivot_root, but we miss an option to disable keyring.

rhatdan:
These look like they are blocked by seccomp rules.  Could you give the actual podman command you are trying to execute?

gzm55:
1. create docker container:

```bash
docker run --rm -it --user test-user \
                   --device /dev/fuse \
                   --cap-add SYS_ADMIN \
                   an-oci-toolset-image sh
```

2. in the rootless ADMIN container, create a nested container:

```bash
podman run --rm -it --net=host --pid=host busybox echo ok
```

As @rhatdan said, the problem is that the `keyctl()` syscall is blocked by seccomp, and according to the [docker document](https://docs.docker.com/engine/security/seccomp/), this syscall is blocked to prevent containers from using the kernel keyring, which is not namespaced. I can confirm that, using an special seccomp profile with `keyctl` enabled, the nested container works.

This syscall is blocked for security concern, so I think we should either namspacize the keyring, or provide an option to reuse the current keyring just like each runtime does. The latter one should be reasonable for nested rootless container.

rhatdan:
Docker is wrong, keyctl is namespaced.  This seccomp statement is dated.
Podman allow these sysctls by default.  
Since this is more about how docker sets up the containers that run podman, it is not a podman issue.  So I am going to close this issue.
BTW Why not run podman natively or podman inside of podman?

gzm55:
> Docker is wrong, keyctl is namespaced. This seccomp statement is dated.
> Podman allow these sysctls by default.
> Since this is more about how docker sets up the containers that run podman, it is not a podman issue. So I am going to close this issue.
> BTW Why not run podman natively or podman inside of podman?

OK, thanks to figure out the solution.

Some old ci systems are still used to build images inside a docker container, so we need nested containers. And this system communicate to daemon via unix sock. Later we would like to try the new podman rest api, this way should introduce the native rootless nested containers to our systems.

gzm55:
[...]

PR #8806, opened 2020-12-22
Pass down EnableKeyring from containers.conf to conmon

rhatdan:
We have a new field in containers.conf that tells whether
or not we want to generate a new keyring in a container.

This field was being ignored.  It now will be followed and
passed down to conmon.

Fixes: https://github.com/containers/podman/issues/8384


rhatdan:

ashley-cui (commented):
LGTM

PR #8807, opened 2020-12-22
play kube: fix args/command handling

### LINKED ISSUE #8803: `podman play kube` (rootless) does not set `command` and `args` as specfied at the k8s specs
/kind bug

**Description**

https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes describes how pods started by k8s are expected to behave when `CMD` and `ENTRYPOINT` are (possibly) defined at the image and `command` and `args` passed (optionally) from the pod declaration.

When assuming that `podman kube play` should behave as k8s and podman 2.1 do, this is a bug.

If it should not, which I have difficulties understanding, but, missing deeper insights into the project, cannot rule out, it is a breaking change from 2.1 to 2.2 and should be documented as such.

I have suggested a resolution at #8666, which currently and from my perspective, is stalling. I can only guess for the reasons  (hoping not to hurt feelings), but I feel it is not being recognized as either being a bug or being a breaking change.

Personally, I feel that discussing about a fortnight about the resolution of either a bug or a breaking change is too long.
If I have trodden on any feet, I ask to be excused, I do not mean to be disrespectful.

The problem is that containers from images, which have both `ENTRYPOINT` and `CMD` set at the image level, and pod declarations which (legitimately, IMHO) omit `command` and `args`, the container does not and cannot start.

Consider the following pod definition:
```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: try
spec:
  containers:
    - name: try
      image: docker.io/library/redis:6
```

With podman 2.2, on the other hand, it crashes, exhibiting
```bash
./bin/podman-v2.2 inspect try_pod-try \
| jq '.[] | {"args": .Args, "cmd": .Config.Cmd, "entrypoint": .Config.Entrypoint }'
{
  "args": [
    "docker-entrypoint.sh"
  ],
  "cmd": [
    "docker-entrypoint.sh"
  ],
  "entrypoint": "docker-entrypoint.sh"
}
```

With podman 2.1, the pod generating mechanism takes both `command` and `args` from the image and starts just fine, which I would expect to continue with 2.2 deriving from my assumption above. If my assumption is false, I'd like to have a loud and clear warning in the changelogs and be grateful for an explanation.

#8666 has further details.

haircommander:
when neither yaml.Args nor yaml.Command are specified, we should use the entrypoint and cmd from the image.

update the tests to cover this and another case (both args and command are specified).
use the registry image instead of redis, as it has both an entrypoint and command specified.
update the documentation around this handling to hopefully prevent regressions and confusion.

fixes #8803

### DIFF
--- a/pkg/specgen/generate/kube/kube.go
+++ b/pkg/specgen/generate/kube/kube.go
@@ -112,11 +112,18 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
 		return nil, err
 	}
 	s.WorkDir = "/"
+	// We will use "Docker field name" internally here to avoid confusion
+	// and reference the "Kubernetes field name" when referencing the YAML
+	// ref: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
+	entrypoint := []string{}
+	cmd := []string{}
 	if imageData != nil && imageData.Config != nil {
 		if imageData.Config.WorkingDir != "" {
 			s.WorkDir = imageData.Config.WorkingDir
 		}
-		s.Command = imageData.Config.Entrypoint
+		// Pull entrypoint and cmd from image
+		entrypoint = imageData.Config.Entrypoint
+		cmd = imageData.Config.Cmd
 		s.Labels = imageData.Config.Labels
 		if len(imageData.Config.StopSignal) > 0 {
 			stopSignal, err := util.ParseSignal(imageData.Config.StopSignal)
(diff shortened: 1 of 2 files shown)


baude:
lgtm

bziemons:
this does not use the spec's entrypoint (`s.Entrypoint`) at all - I am guessing that can be seen on `podman inspect` and that would be irritating?

haircommander:
> this does not use the spec's entrypoint (`s.Entrypoint`) at all - I am guessing that can be seen on `podman inspect` and that would be irritating?

it seems to be a more invasive change of behavior to start using entrypoint. I'm investigating that now

haircommander:
> > this does not use the spec's entrypoint (`s.Entrypoint`) at all - I am guessing that can be seen on `podman inspect` and that would be irritating?
> 
> it seems to be a more invasive change of behavior to start using entrypoint. I'm investigating that now

I will address this in a follow up, it may need more conversation

/hold cancel

zhangguanzhang (approved):

rhatdan:
Not sure what is going on here.  Manually merging.

AlbanBedel:
This bug is quiet a severe break of functionality, can this PR be backported for v2.2.2?

rhatdan:
We are doing RC1 for podman 3.0 today.

PR #8808, opened 2020-12-22
play kube: set entrypoint when interpreting Command

haircommander:
We now set Entrypoint when interpeting the image Entrypoint (or yaml.Command)
and Command when interpreting image Cmd (or yaml.Args)

This change is kind of breaking because now checking Config.Cmd won't return
the full command, but only the {cmd,args}.

Adapt the tests to this change as well

addresses comment: https://github.com/containers/podman/pull/8807#issuecomment-749723307

Note, this contains a breaking change, as now you need to check both Entrypoint and Command for the full container command

carries https://github.com/containers/podman/pull/8807

### DIFF
--- a/pkg/specgen/generate/kube/kube.go
+++ b/pkg/specgen/generate/kube/kube.go
@@ -106,24 +106,20 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
 
 	// TODO: We don't understand why specgen does not take of this, but
 	// integration tests clearly pointed out that it was required.
-	s.Command = []string{}
 	imageData, err := newImage.Inspect(ctx)
 	if err != nil {
 		return nil, err
 	}
 	s.WorkDir = "/"
-	// We will use "Docker field name" internally here to avoid confusion
-	// and reference the "Kubernetes field name" when referencing the YAML
-	// ref: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
-	entrypoint := []string{}
-	cmd := []string{}
+	// Entrypoint/Command handling is based off of
+	// https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
 	if imageData != nil && imageData.Config != nil {
 		if imageData.Config.WorkingDir != "" {
 			s.WorkDir = imageData.Config.WorkingDir
 		}
 		// Pull entrypoint and cmd from image
-		entrypoint = imageData.Config.Entrypoint
-		cmd = imageData.Config.Cmd
+		s.Entrypoint = imageData.Config.Entrypoint
+		s.Command = imageData.Config.Cmd
 		s.Labels = imageData.Config.Labels
 		if len(imageData.Config.StopSignal) > 0 {
 			stopSignal, err := util.ParseSignal(imageData.Config.StopSignal)
@@ -135,16 +131,15 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
 	}
 	// If only the yaml.Command is specified, set it as the entrypoint and drop the image Cmd
 	if len(containerYAML.Command) != 0 {
-		entrypoint = containerYAML.Command
-		cmd = []string{}
+		s.Entrypoint = containerYAML.Command
+		s.Command = []string{}
 	}
 	// Only override the cmd field if yaml.Args is specified
 	// Keep the image entrypoint, or the yaml.command if specified
 	if len(containerYAML.Args) != 0 {
-		cmd = containerYAML.Args
+		s.Command = containerYAML.Args
 	}
 
-	s.Command = append(entrypoint, cmd...)
 	// FIXME,
 	// we are currently ignoring imageData.Config.ExposedPorts
 	if containerYAML.WorkingDir != "" {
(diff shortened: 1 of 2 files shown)


test/e2e/play_kube_test.go
```diff
@@ -815,43 +815,109 @@ var _ = Describe("Podman play kube", func() {
...
 		Expect(inspect.ExitCode()).To(Equal(0))
-		// this image's ENTRYPOINT is called `docker-entrypoint.sh`
-		// so result should be `docker-entrypoint.sh + withArg(...)`
-		Expect(inspect.OutputToString()).To(ContainSubstring(`[docker-entrypoint.sh echo hello]`))
+		Expect(inspect.OutputToString()).To(ContainSubstring(`echo`))
+
+		inspect = podmanTest.Podman([]string{"inspect", getCtrNameInPod(pod), "--format", "'{{ .Config.Cmd }}'"})
+		inspect.WaitWithDefaultTimeout()
+		Expect(inspect.ExitCode()).To(Equal(0))
+		Expect(inspect.OutputToString()).To(ContainSubstring(`hello`))
```
zhangguanzhang, line 920:
```suggestion
		Expect(inspect.OutputToString()).To(Equal(`[hello]`))
```
Maybe this is better

zhangguanzhang:
otherwise LGTM

test/e2e/play_kube_test.go
```diff
@@ -815,43 +815,109 @@ var _ = Describe("Podman play kube", func() {
...
 		// Use the defined command to override the image's command
-		correctCmd := "[" + strings.Join(defaultCtrCmd, " ") + " " + strings.Join(defaultCtrArg, " ")
-		Expect(inspect.OutputToString()).To(ContainSubstring(correctCmd))
+		Expect(ep).To(ContainSubstring(strings.Join(defaultCtrCmd, " ")))
+		Expect(cmd).To(ContainSubstring(strings.Join(defaultCtrArg, " ")))
 	})
 
-	It("podman play kube test correct command with only set command in yaml file", func() {
-		pod := getPod(withCtr(getCtr(withCmd([]string{"echo", "hello"}), withArg(nil))))
+	// If you do not supply command or args for a Container, the defaults defined in the Docker image are used.
```
rhatdan, line 830: s/Docker/Container/g

test/e2e/play_kube_test.go
```diff
@@ -815,43 +815,109 @@ var _ = Describe("Podman play kube", func() {
...
+		Expect(inspect.OutputToString()).To(ContainSubstring(`echo hello`))
+
+		inspect = podmanTest.Podman([]string{"inspect", getCtrNameInPod(pod), "--format", "'{{ .Config.Cmd }}'"})
+		inspect.WaitWithDefaultTimeout()
+		Expect(inspect.ExitCode()).To(Equal(0))
+		// an empty command is reported as '[]'
+		Expect(inspect.OutputToString()).To(ContainSubstring(`[]`))
 	})
 
+	// If you supply only args for a Container, the default Entrypoint defined in the Docker image is run with the args that you supplied.
```
rhatdan, line 878: sed 's/Docker/Container/g'

test/e2e/play_kube_test.go
```diff
@@ -815,43 +815,109 @@ var _ = Describe("Podman play kube", func() {
...
+		Expect(inspect.OutputToString()).To(ContainSubstring(`/entrypoint.sh`))
+
+		inspect = podmanTest.Podman([]string{"inspect", getCtrNameInPod(pod), "--format", "'{{ .Config.Cmd }}'"})
+		inspect.WaitWithDefaultTimeout()
+		Expect(inspect.ExitCode()).To(Equal(0))
+		Expect(inspect.OutputToString()).To(ContainSubstring(`echo hello`))
+	})
+
+	// If you supply a command and args,
+	// the default Entrypoint and the default Cmd defined in the Docker image are ignored.
```
rhatdan, line 901: sed 's/Docker/Container/g'

pkg/specgen/generate/kube/kube.go
```diff
@@ -106,17 +106,21 @@ func ToSpecGen(ctx context.Context, containerYAML v1.Container, iid string, newI
 
 	// TODO: We dont understand why specgen does not take of this, but
 	// integration tests clearly pointed out that it was required.
-	s.Command = []string{}
 	imageData, err := newImage.Inspect(ctx)
 	if err != nil {
 		return nil, err
 	}
 	s.WorkDir = "/"
+	// We will use "Docker field name" internally here to avoid confusion
```
rhatdan, line 114: Docker field name?

Why do we have the "Docker" adgective here?

rhatdan (approved):

test/e2e/play_kube_test.go
haircommander: I chose this because I copied these comments straight from the k8s docs. I'll update them though

pkg/specgen/generate/kube/kube.go
haircommander: it's the language used in the k8s docs

rhatdan: Just cause they use bad language does not mean we have to. :^)

haircommander: updated :)

test/e2e/play_kube_test.go
haircommander: Equal doesn't work because there are some odd characters from the output.  I did change it to contain the [] though

TomSweeneyRedHat:
Changes LGTM
but a rebase is needed.

PR #8810, opened 2020-12-23
Fix podman build --logfile

hshiina:
A opened file object of a logfile gets lost because the variable
`logfile` is redefined in a `if` block. This fix stops redefining
the variable.

A system test for `--logfile` is also added.

Signed-off-by: Hironori Shiina <Hironori.Shiina@fujitsu.com>


rhatdan:
Thanks @hshiina 
LGTM
/approve

rhatdan:

Luap99 (approved):
LGTM

PR #8816, opened 2020-12-23
rootless: automatically split userns ranges

giuseppe:
writing to the id map fails when an extent overlaps multiple mappings in the parent user namespace:

$ cat /proc/self/uid_map
         0       1000          1
         1     100000      65536
$ unshare -U sleep 100 &
[1] 1029703
$ printf "0 0 100\n" | tee /proc/$!/uid_map
0 0 100
tee: /proc/1029703/uid_map: Operation not permitted

This limitation is particularly annoying when working with rootless containers as each container runs in the rootless user namespace, so a command like:

$ podman run --uidmap 0:0:2 --rm fedora echo hi
Error: writing file `/proc/664087/gid_map`: Operation not permitted: OCI permission denied

would fail since the specified mapping overlaps the first mapping (where the user id is mapped to root) and the second extent with the additional IDs available.

Detect such cases and automatically split the specified mapping with the equivalent of:

$ podman run --uidmap 0:0:1 --uidmap 1:1:1 --rm fedora echo hi
hi

A fix has already been proposed for the kernel[1], but even if it accepted it will take time until it is available in a released kernel, so fix it also in pkg/rootless.

[1] https://lkml.kernel.org/lkml/20201203150252.1229077-1-gscrivan@redhat.com/

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -529,6 +529,13 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 		}
 	}
 
+	availableUIDs, availableGIDs, err := rootless.GetAvailableIDMaps()
+	if err != nil {
+		return nil, err
+	}
+	g.Config.Linux.UIDMappings = rootless.MaybeSplitMappings(g.Config.Linux.UIDMappings, availableUIDs)
+	g.Config.Linux.GIDMappings = rootless.MaybeSplitMappings(g.Config.Linux.GIDMappings, availableGIDs)
+
 	// Hostname handling:
 	// If we have a UTS namespace, set Hostname in the OCI spec.
 	// Set the HOSTNAME environment variable unless explicitly overridden by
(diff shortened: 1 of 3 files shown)


rhatdan:
What happens if I do

podman run --uidmap 0:0:1000 --rm fedora echo hi

giuseppe:
> What happens if I do
> 
> podman run --uidmap 0:0:1000 --rm fedora echo hi

it splits like:

```
$ bin/podman run --uidmap 0:0:1000 --rm fedora cat /proc/self/uid_map
         0          0          1
         1          1        999
```

rhatdan:
I think you should modify your example to use my use case, so that it is clear you are not going to put 1000 singe id maps.

rhatdan:
LGTM

giuseppe:
@mheon PTAL

mheon:
LGTM, but you have a merge conflict. Rebase and we can get this landed.

TomSweeneyRedHat:
LGTM
once rebased

giuseppe:
rebased :arrow_up:

PR #8818, opened 2020-12-23
Add pre-checkpoint and restore with previous

chen-zhuohan:
Add --pre-checkpoint option to checkpoint and --with-previous option to restore, for less dump-time when migrating


rhatdan:
First can you squash your commits.
Second you have to sign your commits.
git commit -a --amend -s
git push --force

@adrianreber PTAL

pkg/domain/infra/abi/containers.go
```diff
@@ -484,13 +484,14 @@ func (ic *ContainerEngine) ContainerCheckpoint(ctx context.Context, namesOrIds [
...
 		cons []*libpod.Container
 	)
-	checkOpts := libpod.ContainerCheckpointOptions{
-		Keep:           options.Keep,
-		TCPEstablished: options.TCPEstablished,
-		TargetFile:     options.Export,
-		IgnoreRootfs:   options.IgnoreRootFS,
-		KeepRunning:    options.LeaveRunning,
-	}
+	checkOpts := libpod.ContainerCheckpointOptions{}
```
rhatdan, line 487: Why did you change the structure?

pkg/domain/infra/abi/containers.go
```diff
@@ -520,16 +521,15 @@ func (ic *ContainerEngine) ContainerRestore(ctx context.Context, namesOrIds []st
...
-	restoreOptions := libpod.ContainerCheckpointOptions{
-		Keep:            options.Keep,
-		TCPEstablished:  options.TCPEstablished,
-		TargetFile:      options.Import,
-		Name:            options.Name,
-		IgnoreRootfs:    options.IgnoreRootFS,
-		IgnoreStaticIP:  options.IgnoreStaticIP,
-		IgnoreStaticMAC: options.IgnoreStaticMAC,
-	}
+	restoreOpts := libpod.ContainerRestoreOptions{}
```
rhatdan, line 524: Why change?

rhatdan (approved):

libpod/container_api.go
```diff
@@ -674,25 +674,43 @@ func (c *Container) Refresh(ctx context.Context) error {
 
 // ContainerCheckpointOptions is a struct used to pass the parameters
```
mheon, line 675: Comment needs to be updated

libpod/container_api.go
```diff
@@ -674,25 +674,43 @@ func (c *Container) Refresh(ctx context.Context) error {
...
 	TCPEstablished bool
 	// TargetFile tells the API to read (or write) the checkpoint image
 	// from (or to) the filename set in TargetFile
 	TargetFile string
+	// IgnoreRootfs tells the API to not export changes to
+	// the container's root file-system (or to not import)
+	IgnoreRootfs bool
+}
+
+type ContainerCheckpointOptions struct {
```
mheon, line 691: Needs a documentation comment

libpod/container_api.go
```diff
@@ -674,25 +674,43 @@ func (c *Container) Refresh(ctx context.Context) error {
...
+	// checkpoint archive a new name should be used for the
+	// restored container
+	Name string
+	// Pre Checkpoint container and leave container running
+	PreCheckPoint bool
+	// Dump container with Pre-dump images
+	WithPrevious bool
+}
+
+type ContainerRestoreOptions struct {
```
mheon, line 707: Needs a documentation comment

libpod/container_internal.go
```diff
@@ -134,6 +134,11 @@ func (c *Container) CheckpointPath() string {
 	return filepath.Join(c.bundlePath(), "checkpoint")
 }
 
+// CheckpointPath returns the path to the directory containing the checkpoint
```
mheon, line 137: This function is called `PreCheckPointPath` not `CheckpointPath` - comment needs adjusting

pkg/domain/infra/abi/containers.go
chen-zhuohan: Because there are lots of different between checkpoint and restore. With more Features or changes, the code would be not clear. So I want to distinguish between them. If it has other reason to keep one struct, or just not necessary for the pr, I can change it back.

libpod/container_api.go
chen-zhuohan: thanks, I will add it later!

chen-zhuohan:
Sorry for all, I push the wrong pr😭, I will push another pr to finish the feature and check more.

adrianreber:
@chen-zhuohan even if you already closed this a few comments.

First, I am very happy to see this implemented in Podman, great. Second, there a few non functional changes in this PR which seem unnecessary. I saw added commas, and structure renaming which should at least be separate commits or pull requests.

If I understood your code correctly this adds the possibility to do:

1. podman container checkpoint --pre-checkpoint --export=pre-export.tar.gz
2. scp pre-export.tar.gz destination:
3. podman container checkpoint --with-previous --export=export.tar.gz
4. scp export.tar.gz destination:
5. ssh destination
6. podman container restore --import-previous=pre-export.tar.gz --import=export.tar.gz

Is this correct? You should really list the steps your PR wants to do in the description to make it easier to understand.

If this is the result of this PR it sounds like the right steps to do. The container can keep on running between step 1 and step 3 and the potentially long transfer of the first checkpoint does not require a downtime as the container can keep on running and only the, hopefully, smaller checkpoint image needs to transferred in step 4.

My main question is, how would a pre-copy migration with multiple pre-copy iterations look like. With the current PR it would not be possible and it would be nice to see how this PR could be extended to allow multiple pre-copy iterations (if necessary).

Could Podman even be the place to decide how many pre-copy steps are necessary. Maybe have some thresholds in Podman to decide how many pre-copy iterations Podman should do? And if discussing reduction of migration downtime it is also important to think about post-copy based migration. I am not saying that everything needs to happen in the same PR, but maybe it would be good to have a plan how pre-copy and post-copy can work together. There is also the questions if this should use the CRIU page-server to stream the memory pages from one host to another. I guess I never implemented anything like this until now myself, because it is not clear to me how this can be implemented in such way to be open for multiple pre-copy iterations and how to include post-copy.

Lot's of questions. Happy to discuss possible implementations with you.

chen-zhuohan:
> @chen-zhuohan even if you already closed this a few comments.
> 
> First, I am very happy to see this implemented in Podman, great. Second, there a few non functional changes in this PR which seem unnecessary. I saw added commas, and structure renaming which should at least be separate commits or pull requests.
> 
> If I understood your code correctly this adds the possibility to do:
> 
> 1. podman container checkpoint --pre-checkpoint --export=pre-export.tar.gz
> 2. scp pre-export.tar.gz destination:
> 3. podman container checkpoint --with-previous --export=export.tar.gz
> 4. scp export.tar.gz destination:
> 5. ssh destination
> 6. podman container restore --import-previous=pre-export.tar.gz --import=export.tar.gz
> 
> Is this correct? You should really list the steps your PR wants to do in the description to make it easier to understand.
> 
> If this is the result of this PR it sounds like the right steps to do. The container can keep on running between step 1 and step 3 and the potentially long transfer of the first checkpoint does not require a downtime as the container can keep on running and only the, hopefully, smaller checkpoint image needs to transferred in step 4.
> 
> My main question is, how would a pre-copy migration with multiple pre-copy iterations look like. With the current PR it would not be possible and it would be nice to see how this PR could be extended to allow multiple pre-copy iterations (if necessary).
> 
> Could Podman even be the place to decide how many pre-copy steps are necessary. Maybe have some thresholds in Podman to decide how many pre-copy iterations Podman should do? And if discussing reduction of migration downtime it is also important to think about post-copy based migration. I am not saying that everything needs to happen in the same PR, but maybe it would be good to have a plan how pre-copy and post-copy can work together. There is also the questions if this should use the CRIU page-server to stream the memory pages from one host to another. I guess I never implemented anything like this until now myself, because it is not clear to me how this can be implemented in such way to be open for multiple pre-copy iterations and how to include post-copy.
> 
> Lot's of questions. Happy to discuss possible implementations with you.

Happy to get your comment. 

There are many solutions and tips for faster container migration between physical server, such as NFSNetwork File System(NFS), Remote Direct Memory Access(RDMA) and CRIU page-server you mentioned. Multiple pre-copy iterations seem great, and I and my classmate are forcing on the research if two-step checkpoint can reduce dump time, including image(tar.gz) transport time, the last checkpoint time, restore time. It's hard to determine multiple pre-copy iterations can reduce dump time.  Pre-copy iterations is also complex for implementations, so at the beginning, it isn't in our plan. 

CRIU page-serve seems great, but probably it should be implement by runc or crun first,. Then podman call the api and do more things to help user using.

adrianreber:
@chen-zhuohan runc has a `--page-server` option during checkpointing. So runc could already transfer the pages to the destination host directly.

If you are looking at optimizations you should maybe also look at the compression algorithm of the `--export` option. I was thinking about adding the possibility to configure the compression algorithm for the exported archive as I think during `--export` a lot of time is spent in gzipping the tar archive. Maybe try other algorithms or test if uncompressed is faster.

Many years ago I tested NFS and I cannot recommend it for migration. It was always slower than checkpoint, rsync, restore. Another problem I had with NFS it is unclear how to synchronize the the two involved hosts in the migration. If you do the checkpoint on one system on NFS it is not as clear at which point in time the destination host can read all checkpoint files.

For optimization the page-server and maybe even image streaming server are the best. In the best case you would select a tmpfs as backing file-system to avoid any disk I/O.

chen-zhuohan:
> @chen-zhuohan runc has a `--page-server` option during checkpointing. So runc could already transfer the pages to the destination host directly.
> 
> If you are looking at optimizations you should maybe also look at the compression algorithm of the `--export` option. I was thinking about adding the possibility to configure the compression algorithm for the exported archive as I think during `--export` a lot of time is spent in gzipping the tar archive. Maybe try other algorithms or test if uncompressed is faster.
> 
> Many years ago I tested NFS and I cannot recommend it for migration. It was always slower than checkpoint, rsync, restore. Another problem I had with NFS it is unclear how to synchronize the the two involved hosts in the migration. If you do the checkpoint on one system on NFS it is not as clear at which point in time the destination host can read all checkpoint files.
> 
> For optimization the page-server and maybe even image streaming server are the best. In the best case you would select a tmpfs as backing file-system to avoid any disk I/O.

Thanks a lot, your comments provide many inspiration for us. My classmate still force on this topic, and may take page-server feature to podman. The open pr failed in CI, I don't know what happened. Could you give any advice?

PR #8819, opened 2020-12-23
Add pre-checkpoint and restore with previous

chen-zhuohan:
Add --pre-checkpoint option to checkpoint and --with-previous option to restore, for less dump-time when migrating

Now it only works with two-step checkpoint, the first is pre-checkpoint , the second is checkpoint with previous. It can be used like:

1. podman container checkpoint -P -e pre-checkpoint.tar.gz -l
2. scp pre-export.tar.gz destination
3. podman container checkpoint --with-previous -e checkpoint.tar.gz -l
4. scp export.tar.gz destination
5. ssh destination
6. podman container restore --import-previous=pre-checkpoint.tar.gz --import=checkpoint.tar.gz

Signed-off-by: Zhuohan Chen <chen_zhuohan@163.com>

### DIFF
--- a/cmd/podman/containers/checkpoint.go
+++ b/cmd/podman/containers/checkpoint.go
@@ -58,6 +58,9 @@ func init() {
 
 	flags.BoolVar(&checkpointOptions.IgnoreRootFS, "ignore-rootfs", false, "Do not include root file-system changes when exporting")
 	flags.BoolVar(&checkpointOptions.IgnoreVolumes, "ignore-volumes", false, "Do not export volumes associated with container")
+	flags.BoolVarP(&checkpointOptions.PreCheckPoint, "pre-checkpoint", "P", false, "Dump container's memory information only, leave the container running")
+	flags.BoolVar(&checkpointOptions.WithPrevious, "with-previous", false, "Checkpoint container with pre-checkpoint images")
+
 	validate.AddLatestFlag(checkpointCommand, &checkpointOptions.Latest)
 }
 
@@ -71,6 +74,9 @@ func checkpoint(cmd *cobra.Command, args []string) error {
 	}
 	if checkpointOptions.Export == "" && checkpointOptions.IgnoreVolumes {
 		return errors.Errorf("--ignore-volumes can only be used with --export")
+	}
+	if checkpointOptions.WithPrevious && checkpointOptions.PreCheckPoint {
+		return errors.Errorf("--with-previous can not be used with --pre-checkpoint")
 	}
 	responses, err := registry.ContainerEngine().ContainerCheckpoint(context.Background(), args, checkpointOptions)
 	if err != nil {
--- a/docs/source/markdown/podman-container-checkpoint.1.md
+++ b/docs/source/markdown/podman-container-checkpoint.1.md
@@ -58,11 +58,25 @@ This option must be used in combination with the **--export, -e** option.
 When this option is specified, the content of volumes associated with
 the container will not be included into the checkpoint tar.gz file.
 
+#### **--pre-checkpoint**, **-P**
+
+Dump the container's memory information only, leaving the container running. Later
+operations will supersede prior dumps. It only works on runc 1.0-rc3 or higher.
+
+#### **--with-previous**
+
+Check out the container with previous criu image files in pre-dump. It only works
+without **--pre-checkpoint** or **-P**. It only works on runc 1.0-rc3 or higher.
+
 ## EXAMPLE
 
 podman container checkpoint mywebserver
 
 podman container checkpoint 860a4b23
+
+podman container checkpoint -P -e pre-checkpoint.tar.gz -l
+
+podman container checkpoint --with-previous -e checkpoint.tar.gz -l
 
 ## SEE ALSO
 podman(1), podman-container-restore(1)
--- a/docs/source/markdown/podman-container-restore.1.md
+++ b/docs/source/markdown/podman-container-restore.1.md
@@ -48,6 +48,11 @@ Import a checkpoint tar.gz file, which was exported by Podman. This can be used
 to import a checkpointed container from another host. Do not specify a *container*
 argument when using this option.
 
+#### **--import-previous**
+
+Import a pre-checkpoint tar.gz file which was exported by Podman. This option
+must be used with **-i** or **--import**. It only works on runc 1.0-rc3 or higher.
+
 #### **--name**, **-n**
 
 This is only available in combination with **--import, -i**. If a container is restored
@@ -97,6 +102,8 @@ the content of associated volumes will not be restored.
 podman container restore mywebserver
 
 podman container restore 860a4b23
+
+podman container restore --import-previous pre-checkpoint.tar.gz --import checkpoint.tar.gz
 
 ## SEE ALSO
 podman(1), podman-container-checkpoint(1)
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -706,6 +706,13 @@ type ContainerCheckpointOptions struct {
 	// IgnoreVolumes tells the API to not export or not to import
 	// the content of volumes associated with the container
 	IgnoreVolumes bool
+	// Pre Checkpoint container and leave container running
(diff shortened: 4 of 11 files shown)


docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -1,65 +1 @@
...

podman container checkpoint mywebserver

podman container checkpoint 860a4b23

## SEE ALSO
podman(1), podman-container-restore(1)

## HISTORY
September 2018, Originally compiled by Adrian Reber <areber@redhat.com>
```
zhangguanzhang, line 1: Problem with line breaks

chen-zhuohan: Fixed

chen-zhuohan:
Something fail but I don't know why it happened, could someone help me?

yufeifly:
What if I want to pre-dump several times? Does it support that?

chen-zhuohan:
> What if I want to pre-dump several times? Does it support that?

Not yet. Now it only supports two-step checkpoint. It may be supported later.

yufeifly:
> Not yet. Now it only supports two-step checkpoint. It may be supported later.
 
and we cannot export the checkpoints of some type of containers, e.g. redis container. So I think we should support pre-checkpoint without exportation.

cmd/podman/containers/checkpoint.go
```diff
@@ -57,6 +57,9 @@ func init() {
 	_ = checkpointCommand.RegisterFlagCompletionFunc(exportFlagName, completion.AutocompleteDefault)
 
 	flags.BoolVar(&checkpointOptions.IgnoreRootFS, "ignore-rootfs", false, "Do not include root file-system changes when exporting")
+	flags.BoolVarP(&checkpointOptions.PreCheckPoint, "pre-checkpoint", "P", false, "Dump container's memory information only, leave the container running after this")
```
TomSweeneyRedHat, line 60:
```suggestion
	flags.BoolVarP(&checkpointOptions.PreCheckPoint, "pre-checkpoint", "P", false, "Dump container's memory information only, leave the container running")
```

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -52,12 +52,24 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
 to explicitly disable including changes to the root file-system into
 the checkpoint archive file.
 
+#### **--pre-checkpoint**, **-P**
+
+Dump container's memory information only, leave the container running after this. Later
```
TomSweeneyRedHat, line 57:
```suggestion
Dump the container's memory information only, leaving the container running. Later
```

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -52,12 +52,24 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
...
+operations will override the previous ones. It means a contaniner only has one pre-dump images.
```
TomSweeneyRedHat, line 58:
```suggestion
operations will supersede prior dumps.
```

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -52,12 +52,24 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
...
+
+#### **--with-previous**
+
+This only works without **--pre-checkpoint**. It will checkout the container with previous criu image files in pre-dump.
```
TomSweeneyRedHat, line 62: I'm not clear what criu is or how this operates.  First attempt, may adjust later.
```suggestion
Check out the container with previous criu image files in pre-dump.

```

docs/source/markdown/podman-container-restore.1.md
```diff
@@ -85,12 +85,19 @@ exported checkpoint with **--name, -n**.
 
 Using **--ignore-static-mac** tells Podman to ignore the MAC address if it was
 configured with **--mac-address** during container creation.
+
+#### **--import-previous**
+
+Import a pre-checkpoint tar.gz file if need, which was exported by Podman. The option
```
TomSweeneyRedHat, line 91:
```suggestion
Import a pre-checkpoint tar.gz file which was exported by Podman. This option
```

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -52,12 +52,24 @@ exported to a tar.gz file it is possible with the help of **--ignore-rootfs**
 to explicitly disable including changes to the root file-system into
 the checkpoint archive file.
 
+#### **--pre-checkpoint**, **-P**
```
TomSweeneyRedHat, line 55: Can you add an example below?

libpod/container_api.go
```diff
@@ -715,6 +722,12 @@ func (c *Container) Checkpoint(ctx context.Context, options ContainerCheckpointO
 		}
 	}
 
+	if options.WithPrevious {
```
TomSweeneyRedHat, line 725: I think you want a check to make sure WithPrevious and pre-checkpoint are not used at the same time.

docs/source/markdown/podman-container-checkpoint.1.md
chen-zhuohan: No problem.

chen-zhuohan: It has existed in 70-72 lines.

libpod/container_api.go
chen-zhuohan: Should I add it in cmd/podman/containers/checkpoints.go, or libpod/container_api.go?

docs/source/markdown/podman-container-restore.1.md
```diff
@@ -85,12 +85,19 @@ exported checkpoint with **--name, -n**.
 
 Using **--ignore-static-mac** tells Podman to ignore the MAC address if it was
 configured with **--mac-address** during container creation.
+
+#### **--import-previous**
```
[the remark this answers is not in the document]
chen-zhuohan: Ok.

rhatdan:
/approve
@adrianreber PTAL

adrianreber:
CRIU's pre-dump support does not work on aarch64, because aarch64 has no kernel support for dirty page tracking. All other architectures CRIU supports (pp64le, s390x, x86_64) can do dirty page tracking. This is something Podman could already test for to make sure pre-copy is not used on aarch64. Maybe also something for the documentation.

The whole pre-dump support has no test yet, can you include a test to the existing checkpoint tests in e2e?

chen-zhuohan:
> CRIU's pre-dump support does not work on aarch64, because aarch64 has no kernel support for dirty page tracking. All other architectures CRIU supports (pp64le, s390x, x86_64) can do dirty page tracking. This is something Podman could already test for to make sure pre-copy is not used on aarch64. Maybe also something for the documentation.
> 
> The whole pre-dump support has no test yet, can you include a test to the existing checkpoint tests in e2e?

chen-zhuohan:
> CRIU's pre-dump support does not work on aarch64, because aarch64 has no kernel support for dirty page tracking. All other architectures CRIU supports (pp64le, s390x, x86_64) can do dirty page tracking. This is something Podman could already test for to make sure pre-copy is not used on aarch64. Maybe also something for the documentation.
> 
> The whole pre-dump support has no test yet, can you include a test to the existing checkpoint tests in e2e?

Yes, I have tried to add two test, about export and not export with pre-checkpoint. Except to get your review.

adrianreber:
Tests look good. I would be happy to see this merged.

One thing were tests might fail is, that crun does not yet have support for pre-dumping. Only runc has pre-dump support today.

@chen-zhuohan You need to detect if you are running with crun to disable --pre-dump support in Podman or, better, add pre-dump support to crun.

libpod/container_api.go
TomSweeneyRedHat: My thought was here, but where ever you feel is best.

PR #8821, opened 2020-12-23
Containers should not get inheritable caps by default

rhatdan:
When I launch a container with --userns=keep-id the rootless processes
should have no caps by default even if I launch the container with
--privileged.  It should only get the caps if I specify by hand the
caps I want leaked to the process.

Currently we turn off capeff and capamb, but not capinh.  This patch
treats capinh the same way as capeff and capamb.

### DIFF
--- a/libpod/oci.go
+++ b/libpod/oci.go
@@ -151,8 +151,6 @@ type OCIRuntime interface {
 type ExecOptions struct {
 	// Cmd is the command to execute.
 	Cmd []string
-	// CapAdd is a set of capabilities to add to the executed command.
-	CapAdd []string
 	// Env is a set of environment variables to add to the container.
 	Env map[string]string
 	// Terminal is whether to create a new TTY for the exec session.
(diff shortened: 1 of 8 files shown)


rhatdan:
Helps fix https://github.com/containers/toolbox/issues/651

rhatdan:

baude:
lgtm

ashley-cui:
Unhappy tests :(

mheon:
I think this one needs further discussion. It is 100% not a fix for the reported issue, but it could still be a good idea.

mheon:
Good candidate for watercooler.

libpod/oci.go
```diff
@@ -151,8 +151,6 @@ type OCIRuntime interface {
 type ExecOptions struct {
 	// Cmd is the command to execute.
 	Cmd []string
-	// CapAdd is a set of capabilities to add to the executed command.
-	CapAdd []string
```
mheon, line 155: I think you need to retain this - `podman exec --privileged` can be run in containers that are not privileged.

PR #8828, opened 2020-12-24
Add --all to podman start

### LINKED ISSUE #8779: Add `podman system boot` for installing services
/kind feature

**Description**

Add a `podman system boot` command that installs a systemd service to run a pod on startup as the user who installed the service. 

This is a followup of https://github.com/containers/podman-compose/issues/254 where I had requested a feature of `podman-compose` to install systemd services for a pod via some built in command. The motivation here is that I've noticed users all installing systemd services for their pods in different ways making it difficult to manage. @rhatdan and @mheon discussed adding a `podman system boot` as a solution to the linked issue from `podman-compose`. I'm hoping opening the issue here (as suggested by @rhatdan) is the right thing to do to move this forward.

AWS/VirtualBox/physical

boaz0:
closes #8779 

### Progress

- [x] Add `--all` flag to restart all containers in `podman start`
- [x] ~Add to `--filter` to `podman start`. This flag allows to filter for specific attributes of the containers. For instance, a filter to match a specific restart policy.~ moved to #9689
- [x] Add tests to the changes above
- [x] Update `podman start` documents accordingly
- [ ] Ship a systemd service (similar to podman-auto-update.service) that package managers would install on the system.
- [x] ~podman systemm boot~

### DIFF
--- a/cmd/podman/containers/start.go
+++ b/cmd/podman/containers/start.go
@@ -57,6 +57,8 @@ func startFlags(cmd *cobra.Command) {
 	flags.BoolVarP(&startOptions.Interactive, "interactive", "i", false, "Keep STDIN open even if not attached")
 	flags.BoolVar(&startOptions.SigProxy, "sig-proxy", false, "Proxy received signals to the process (default true if attaching, false otherwise)")
 
+	flags.BoolVar(&startOptions.All, "all", false, "Start all containers regardless of their state or configuration")
+
 	if registry.IsRemote() {
 		_ = flags.MarkHidden("sig-proxy")
 	}
@@ -79,7 +81,7 @@ func init() {
 }
 
 func validateStart(cmd *cobra.Command, args []string) error {
-	if len(args) == 0 && !startOptions.Latest {
+	if len(args) == 0 && !startOptions.Latest && !startOptions.All {
 		return errors.New("start requires at least one argument")
 	}
 	if len(args) > 0 && startOptions.Latest {
@@ -87,6 +89,12 @@ func validateStart(cmd *cobra.Command, args []string) error {
 	}
 	if len(args) > 1 && startOptions.Attach {
 		return errors.Errorf("you cannot start and attach multiple containers at once")
+	}
+	if (len(args) > 0 || startOptions.Latest) && startOptions.All {
+		return errors.Errorf("either start all containers or the container(s) provided in the arguments")
+	}
+	if startOptions.Attach && startOptions.All {
+		return errors.Errorf("you cannot start and attach all containers at once")
 	}
 	return nil
 }
--- a/docs/source/markdown/podman-start.1.md
+++ b/docs/source/markdown/podman-start.1.md
@@ -38,6 +38,10 @@ to run containers such as CRI-O, the last started container could be from either
 
 Proxy received signals to the process (non-TTY mode only). SIGCHLD, SIGSTOP, and SIGKILL are not proxied. The default is *true* when attaching, *false* otherwise.
 
+#### **\-\-all**
+
+Start all the containers created by Podman, default is only running containers.
+
 ## EXAMPLE
 
 podman start mywebserver
--- a/test/system/045-start.bats
+++ b/test/system/045-start.bats
@@ -1,0 +1,43 @@
+#!/usr/bin/env bats   -*- bats -*-
+
+load helpers
+
+@test "podman start --all - start all containers" {
+    # Run a bunch of short-lived containers, with different --restart settings
+    run_podman run -d $IMAGE /bin/true
+    cid_none_implicit="$output"
+    run_podman run -d --restart=no $IMAGE /bin/false
+    cid_none_explicit="$output"
+    run_podman run -d --restart=on-failure $IMAGE /bin/true
+    cid_on_failure="$output"
+
+    # Run one longer-lived one.
+    run_podman run -d --restart=always $IMAGE sleep 20
+    cid_always="$output"
+
+    run_podman wait $cid_none_implicit $cid_none_explicit $cid_on_failure
+
+    run_podman start --all
+    is "$output" ".*$cid_none_implicit" "started: container with no --restart"
+    is "$output" ".*$cid_none_explicit" "started: container with --restart=no"
+    is "$output" ".*$cid_on_failure" "started: container with --restart=on-failure"
+    if [[ $output =~ $cid_always ]]; then
+        die "podman start --all restarted a running container"
+    fi
+
+    run_podman rm $cid_none_implicit $cid_none_explicit $cid_on_failure
+    run_podman stop -t 1 $cid_always
+    run_podman rm $cid_always
+}
+
+@test "podman start --all with incompatible options" {
+    expected="Error: either start all containers or the container(s) provided in the arguments"
+    run_podman 125 start --all 12333
+    is "$output" "$expected" "start --all, with args, throws error"
+    if ! is_remote; then
+        run_podman 125 start --all --latest
+        is "$output" "$expected" "podman start --all --latest"
+    fi
+}
+
+# vim: filetype=sh
--- a/pkg/domain/entities/containers.go
+++ b/pkg/domain/entities/containers.go
@@ -265,6 +265,7 @@ type ContainerExistsOptions struct {
 // ContainerStartOptions describes the val from the
 // CLI needed to start a container
 type ContainerStartOptions struct {
+	All         bool
 	Attach      bool
 	DetachKeys  string
 	Interactive bool
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -693,14 +693,17 @@ func (ic *ContainerEngine) ContainerExecDetached(ctx context.Context, nameOrID s
 func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []string, options entities.ContainerStartOptions) ([]*entities.ContainerStartReport, error) {
 	reports := []*entities.ContainerStartReport{}
 	var exitCode = define.ExecErrorCodeGeneric
-	ctrs, rawInputs, err := getContainersAndInputByContext(false, options.Latest, namesOrIds, ic.Libpod)
+	ctrs, rawInputs, err := getContainersAndInputByContext(options.All, options.Latest, namesOrIds, ic.Libpod)
 	if err != nil {
 		return nil, err
 	}
 	// There can only be one container if attach was used
 	for i := range ctrs {
 		ctr := ctrs[i]
-		rawInput := rawInputs[i]
+		rawInput := ctr.ID()
+		if !options.All {
+			rawInput = rawInputs[i]
+		}
 		ctrState, err := ctr.State()
 		if err != nil {
 			return nil, err
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -506,7 +506,7 @@ func startAndAttach(ic *ContainerEngine, name string, detachKeys *string, input,
 func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []string, options entities.ContainerStartOptions) ([]*entities.ContainerStartReport, error) {
 	reports := []*entities.ContainerStartReport{}
 	var exitCode = define.ExecErrorCodeGeneric
-	ctrs, err := getContainersByContext(ic.ClientCtx, false, false, namesOrIds)
+	ctrs, err := getContainersByContext(ic.ClientCtx, options.All, false, namesOrIds)
 	if err != nil {
 		return nil, err
 	}
@@ -514,9 +514,13 @@ func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []stri
 	// There can only be one container if attach was used
 	for i, ctr := range ctrs {
 		name := ctr.ID
+		rawInput := ctr.ID
+		if !options.All {
+			rawInput = namesOrIds[i]
+		}
 		report := entities.ContainerStartReport{
 			Id:       name,
-			RawInput: namesOrIds[i],
+			RawInput: rawInput,
 			ExitCode: exitCode,
 		}
 		ctrRunning := ctr.State == define.ContainerStateRunning.String()
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -602,9 +602,9 @@ func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []stri
 				reports = append(reports, &report)
 				continue
 			}
+			report.ExitCode = 0
+			reports = append(reports, &report)
 		}
-		report.ExitCode = 0
-		reports = append(reports, &report)
 	}
 	return reports, nil
 }


cmd/podman/system/boot.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/spf13/cobra"
+)
+
+var (
+	bootOptions     = entities.SystemPruneOptions{}
+	bootDescription = fmt.Sprintf(`
+  podman system boot
+
+        Start all stopped containers
```
rhatdan, line 19: Start all non running containers with a restart policy.

cmd/podman/system/boot.go
```diff
@@ -0,0 +1,73 @@
...
+		All:     true,
+	}
+	containers, err := registry.ContainerEngine().ContainerList(registry.GetContext(), listOpts)
+	if err != nil {
+		return err
+	}
+
+	cids := []string{}
+	for _, container := range containers {
+		if !container.Exited {
```
rhatdan, line 53: Should not start any containers with a Restart Policy of NO.

cmd/podman/system/boot.go
rhatdan: Or any that have had a MAX Recound > recount.

rhatdan (approved):
Not sure if this should be 
podman system boot 
or
podman containers boot.

cmd/podman/system/boot.go
boaz0: Sounds good but I am not sure where do I get this? Should I get this from the pod assigned to it?

rhatdan: It might be as simple as
if container.ShouldRestart()

docs/source/markdown/podman-system-boot.1.md
```diff
@@ -0,0 +1,20 @@
...
+
+This command is looking for stopped containers that should be started at boot time and starting them.
+
+The number of retries is set by the container's restart policy **MaximumRetryCount**.
+
+## SEE ALSO
+`podman(1)`, `containers.conf(5)`
+
+## HISTORY
+February 2020, Originally compiled by Boaz Shuster (bshuster at redhat dot com)
```
AkihiroSuda, line 20: nit: Feb 2020 -> Dec 2020 (or Jan 2021)

boaz0: Thanks

rhatdan:
This is not going to make the 3.0 release, moving to 3.1

rhatdan:
@boaz0 Could you rebase this please.

boaz0:
Just did

boaz0:
Ummm, I have no idea why openshift bot doesn't remove the tag, though.

boaz0:
Yea, that's what I am doing right now. Thanks for the support @rhatdan

cmd/podman/containers/start.go
```diff
@@ -57,6 +63,11 @@ func startFlags(cmd *cobra.Command) {
 	flags.BoolVarP(&startOptions.Interactive, "interactive", "i", false, "Keep STDIN open even if not attached")
 	flags.BoolVar(&startOptions.SigProxy, "sig-proxy", false, "Proxy received signals to the process (default true if attaching, false otherwise)")
 
+	flags.BoolVar(&startAll, "all", false, "Start all containers regardless to their state or configuration")
```
TomSweeneyRedHat, line 66: Soft suggestion
```suggestion
	flags.BoolVar(&startAll, "all", false, "Start all containers regardless of their state or configuration")
```

cmd/podman/containers/start.go
```diff
@@ -88,6 +99,15 @@ func validateStart(cmd *cobra.Command, args []string) error {
 	if len(args) > 1 && startOptions.Attach {
 		return errors.Errorf("you cannot start and attach multiple containers at once")
 	}
+	if (len(args) > 0 || startOptions.Latest) && startAll {
+		return errors.Errorf("either start all containers or the container(s) you provided in the arguments")
```
TomSweeneyRedHat, line 103:
```suggestion
		return errors.Errorf("either start all containers or the container(s)  provided in the arguments")
```

UebelAndre:
Hey, friendly ping here. Any updates 😅

boaz0:
@UebelAndre 
The code is ready to be reviewed. What is left for me to do is to update the docs and work on the service file that will be added to the podman package.

vrothberg (commented):
Thanks for working on it!

I think this needs to be broken into separate commits (or PRs if you wish). We further need to wire support into the remote client and the bindings.

`--all` is conceptually different from `--filter` and should be added separately. I think that `--filter` should not require `--all`.

Also, the testing matrix needs to be increased.

test/system/710-start.bats
```diff
@@ -0,0 +1,23 @@
...
+
+function teardown() {
+    basic_teardown
+}
+
+@test "podman start --all - start all containers" {
+    run_podman start --all
+    is "$output" "$cid_none
+$cid_always"
+}
```
TomSweeneyRedHat, line 21: It would probably be good to have a failure case or two here.  i.e.  `--all` and `--latest` in one line and/or `--all` and  a containerID or two.

boaz0: Right, I forgot about that. Thanks for mentioning it.

boaz0:
@TomSweeneyRedHat are the failing tests related to my change?

edsantiago:
@boaz0 it sure looks like it. What I'm seeing is:
```console
$ podman run -d first
cid1
$ podman run -d --restart=always second
cid2
$ podman start --all
cid2           <--- whereas you seemed to be expecting cid1 and cid2
```
Can you confirm that your `start --all` code is working as you expect?

Also, I'm going to be offering some review comments in a few moments.

test/system/045-start.bats
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function teardown() {
+    basic_teardown
+}
+
+@test "podman start --all - start all containers" {
+    run_podman start --all
+    is "$output" "$cid_none
+$cid_always"
```
edsantiago, line 20: This is currently failing; I can't diagnose that. But were it to pass: how certain are you about the order of CIDs? Is podman guaranteed to emit them in consistent order? If not, it gets trickier to test; ask if you need help.

Also, please take advantage of the three-argument form of `is`:
```
   is "$output" "$cid_none...." "podman start --all emits both CIDs"
```

test/system/045-start.bats
```diff
@@ -0,0 +1,33 @@
...
+}
+
+@test "podman start --all --latest - should fail" {
+    run_podman 125 start --all --latest
```
edsantiago, line 24: This is probably going to fail with podman-remote (I could check logs, but I'm lazy). And, since the error messages are the same, you could probably combine these two tests into one:
```bash
@test "podman start --all with incompatible options" {
    expect="Error: either blah blah"

    run_podman 125 start --all 12345
    is "$output" "$expect" "podman start --all + CID"

    if ! is_remote; then
        run_podman 125 start --all --latest
        is "$output" "$expect" "podman start --all --latest"
    fi
```

edsantiago (changes requested):
Recommendations for tests. setup/teardown are strong suggestions, the others less so.

TomSweeneyRedHat:
@boaz0 I think @edsantiago has answered the test questions.

boaz0:
@edsantiago your comments were so helpful!!! Thank you so much :bow:

docs/source/markdown/podman-start.1.md
```diff
@@ -38,6 +38,10 @@ to run containers such as CRI-O, the last started container could be from either
 
 Proxy received signals to the process (non-TTY mode only). SIGCHLD, SIGSTOP, and SIGKILL are not proxied. The default is *true* when attaching, *false* otherwise.
 
+#### **--all**, **-a**
```
edsantiago, line 41: For complicated reasons, you now have to backslash-escape double dashes: `\-\-all`. Look at the lines above for examples.

boaz0: Yeah I saw this in the CI :dango:

boaz0:
I still don't understand why this is happening:
```
hack/xref-helpmsgs-manpages
xref-helpmsgs-manpages: podman container start --help lists --all, but --all not in docs/source/markdown/podman-start.1.md
xref-helpmsgs-manpages: podman start --help lists --all, but --all not in docs/source/markdown/podman-start.1.md
```

"--all" does appear in podman-start.1.md.

mheon:
@edsantiago I'll point to you on this one - it looks like it ought to work?

docs/source/markdown/podman-start.1.md
```diff
@@ -38,6 +38,10 @@ to run containers such as CRI-O, the last started container could be from either
 
 Proxy received signals to the process (non-TTY mode only). SIGCHLD, SIGSTOP, and SIGKILL are not proxied. The default is *true* when attaching, *false* otherwise.
 
+#### **-\-\all**, **-a**
```
edsantiago, line 41: Please look really closely at this. Really, really, really closely. Compare it to the other lines, like line 37 or 32.

boaz0: :facepalm: 
THANKS

test/system/045-start.bats
```diff
@@ -0,0 +1,26 @@
...
+    run_podman start --all
+    is "$output" "$cid_none"
+    #[[ "$output" == *"$cid_none"* ]]
+    #[[ "$output" == *"$cid_always"* ]]
+}
+
+@test "podman start --all with incompatible options" {
+    expected="Error: either start all containers or the container(s) provided in the arguments"
+    run_podman 125 start --all 12333
+    cid_none="$output"
```
edsantiago, line 19: I think you meant to write `is "$output" "$expected" "start --all, with args, throws error"`

test/system/045-start.bats
```diff
@@ -0,0 +1,26 @@
...
+
+@test "podman start --all - start all containers" {
+    run_podman run -d $IMAGE /bin/true
+    cid_none="$output"
+    run_podman run -d --restart=always $IMAGE /bin/true
+    cid_always="$output"
+    run_podman start --all
+    is "$output" "$cid_none"
+    #[[ "$output" == *"$cid_none"* ]]
+    #[[ "$output" == *"$cid_always"* ]]
```
edsantiago, line 13: No need to leave these.

test/system/045-start.bats
```diff
@@ -0,0 +1,26 @@
...
+
+load helpers
+
+@test "podman start --all - start all containers" {
+    run_podman run -d $IMAGE /bin/true
+    cid_none="$output"
+    run_podman run -d --restart=always $IMAGE /bin/true
+    cid_always="$output"
+    run_podman start --all
+    is "$output" "$cid_none"
```
edsantiago, line 11: Shouldn't this be `$cid_always`?

edsantiago (changes requested):
Sorry to catch this so late; I didn't have time for a full review this morning.

test/system/045-start.bats
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bats   -*- bats -*-
+
+load helpers
+
+@test "podman start --all - start all containers" {
+    run_podman run -d $IMAGE /bin/true
+    cid_none="$output"
+    run_podman run -d --restart=always $IMAGE /bin/true
```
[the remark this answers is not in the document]
boaz0: Hi @edsantiago - yes and no.
`run_podman run -d --restart=always $IMAGE /bin/true` should be exited less than a second and be caught by `podman start --all` which try to start it again.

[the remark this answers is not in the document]
edsantiago: OK, the `--restart=always` container exits immediately. But it also restarts immediately, over and over and over. What will happen if `podman start --all` catches this container in the very very brief moment that it is in state Running? Will it restart it (and test will pass), or will it skip it, and test will fail because it doesn't see that CID? Based on my experiments with a `sleep 5` container (instead of `true`), I believe the latter will happen, therefore this test is racey. But I need confirmation from @containers/podman-maintainers .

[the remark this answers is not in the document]
boaz0: @edsantiago I see your point. Yeah I agree.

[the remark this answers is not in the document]
mheon: The Libpod APIs return an error on starting a started container. The higher-level start code assumes that starting a started container is a no-op and so it squashes this error and returns cleanly. You're operating entirely in cmd/podman, which should be above the level of the squash - I imagine that any start on a running container will add it to the list of started containers, having taken no action.

rhatdan:
LGTM
Thanks @boaz0 

edsantiago:
LGTM but tests are racey (see comments). The race seems very unlikely to trigger, but I'd really prefer race-free tests. 

Unfortunately, I can't think of a way. This has been on my mind for the past week, and every day I come back to revisit it, and every day decide I have no new ideas for fixing it.

test/system/045-start.bats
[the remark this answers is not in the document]
boaz0: Thanks a lot @edsantiago - that's very helpful.

edsantiago:
@boaz0 the test failures look real: somehow podman-remote is trying to restart the running container.

boaz0:
@edsantiago - that's why it failed :fearful:. I will add it to podman-remote. Thanks.

edsantiago:
@boaz0 exactly the same failure again

boaz0:
Um I'm not sure why it failed again. I did run it as podman-remote and it worked...

edsantiago:
Can you be more specific about what you ran, and what you mean by "it worked"?

On my Fedora 33 system, I can 100% reproduce the CI failure via:
```console
window1:
$ sudo bin/podman system service --timeout=0
```
```console
window2:
$ sudo env PODMAN=$(pwd)/bin/podman-remote bats test/system/045-start.bats
```
With `podman` instead of `podman-remote`, the test passes.

boaz0:
@edsantiago I removed the comment as I could reproduce it on my machine.
However, I think this is inconsistency between podman-remote and podman.
If I run `podman start $cid_always` it won't restart it while `podman-remote --url $PODMAN_SOCKET start $cid_always` will restart it.

boaz0:
Odd - looking at the code they all seem the same. I am investigating this further.

boaz0:
OK I think I found the error. @edsantiago can you confirm this:
https://github.com/containers/podman/pull/8828/files#diff-7458c1a055a350fe316f106abaf9936dd993da3e524faf64bad09f3a3ae01a68R606-R607
These two lines were supposed to be inside the `if !ctrRunning { ... }`. And because of that I will see containers that are not running in the output.

edsantiago:
@boaz0 I don't grok that area of code, sorry. @mheon or @rhatdan WDYT?

edsantiago (approved):

edsantiago:
Looks like you figured it out! (As in, "got tests passing" :-). LGTM but I would like close scrutiny of the new change from @containers/podman-maintainers 

You've put a lot of energy into this one. Thank you for your perseverance and patience.

rhatdan:
/lgtm
@boaz0 Thanks.

UebelAndre:
What a legend! Thank you @boaz0!!!

boaz0:
@UebelAndre I didn't do much. I am standing on the shoulders of giants @edsantiago and @rhatdan - they guided me and helped a lot. :bow: :pray: :clap: 

Good job team!

PR #8832, opened 2020-12-25
Fix e2e test for `podman build --logfile`

hshiina:
This is a tiny follow-up for #8810
Type casting is necessary to see if the logfile size is not equal to 0.

Signed-off-by: Hironori Shiina <Hironori.Shiina@fujitsu.com>

### DIFF
--- a/test/e2e/build_test.go
+++ b/test/e2e/build_test.go
@@ -72,9 +72,9 @@ var _ = Describe("Podman build", func() {
 
 		st, err := os.Stat(logfile)
 		Expect(err).To(BeNil())
-		Expect(st.Size()).To(Not(Equal(0)))
+		Expect(st.Size()).To(Not(Equal(int64(0))))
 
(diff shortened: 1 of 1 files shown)


test/e2e/build_test.go
```diff
@@ -72,9 +72,9 @@ var _ = Describe("Podman build", func() {
 
 		st, err := os.Stat(logfile)
 		Expect(err).To(BeNil())
-		Expect(st.Size()).To(Not(Equal(0)))
+		Expect(st.Size()).To(Not(Equal(int64(0))))
```
zhangguanzhang, line 75:
```suggestion
		Expect(st.Size()).To(Not(Equal(0)))
```

TomSweeneyRedHat: @zhangguanzhang do you have any objections to merging this with the code as is?  I think the new test with int64 is valid.
LGTM

Issue #4739, opened 2019-12-21
Remote snapshotter in podman


siscia:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description** Remote snapshotter in podman

We are really interested in the possibility to use a remote snapshotter like the one provide in containerd also in podman.

A remote snapshotter is a piece a containerd plugin which is provided as input the name of the layer that containerd need and it either mounts the correct directory or return an error.

All this process is manage by "user-level" code/plugin.

Is something like this even possible in podman_

baude:
care to contribute?

siscia:
Maybe I can find the time myself or maybe we can find some resource to have somebody working on it.
It is something interesting for podman? If I had a PR ready, it would be merged?
I haven't really find the time to explore the codebase, who is responsible for this particular part of the codebase?

AkihiroSuda:
cc @ktock

mheon:
I somewhat suspect this world be part of the containers/storage library.

ktock:
Thanks a lot for opening it! I'm keen to contribute to it.

I'm currently working on the implementation of remote snapshotter plugin which enables us to mount layers without pulling the actual contents. We can plug any filesystem into it so we currently support CRFS's stargz-based filesystem and we discuss to support CernVM-FS and other filesystems.
https://github.com/ktock/remote-snapshotter

Remote snapshotter currently supports [containerd's snapshotter API](https://github.com/containerd/containerd/blob/master/snapshots/snapshotter.go) but I think it's not hard to support [graphdriver API](https://github.com/containers/storage/blob/master/drivers/driver.go) as well.

I'll look deep into the codebase.

@siscia How do you think about this implementation strategy?

ktock:
I opened the discussion on https://github.com/containers/storage/issues/498 .

giuseppe:
low level bits are being worked in https://github.com/giuseppe/crfs-plugin and fuse-overlayfs.

The rest of the implementation, as @mheon said, should go into containers/storage

Other remote file systems can be added in a similar way to `crfs-plugin` that can be used with fuse-overlayfs to lookup files from the image/lower layers

rhatdan:
@giuseppe Any more progress on this?

siscia:
I am still trying to figure out what is the best approach for this.

I was expecting the storage to be layer-digest based, like each layer was indexed by its own hash.
Something like:
```
- storage
|- 123...
|- abc...
[...]

PR #8837, opened 2020-12-26
[wip] Support additional layer store (patch for podman)

ktock:
https://github.com/containers/podman/issues/4739

This enables podman to create containers using layers stored in a specified directory instead of pulling them from the registry. Leveraging this feature with remotely-mountable layers provided by stargz/zstd:chunked or CVMFS, podman can achieve lazy pulling.

That directory is named "additional layer store" (call ALS in this doc) and has the following structure.

```
<ALS root>
`-- <ID of layer1>
`-- <ID of layer2>
...
`-- .pool
```

Each layer is extracted and stored in ALS with a unique directory name (`<ID of layer1>`, `<ID of layer2>` in the above). There is `.pool` directory which is used for layer discovery described later.

### Changes in contianers/storage: https://github.com/containers/storage/pull/795

When the storage driver (e.g. overlay) is configured to use ALS, it uses layers in ALS with regarding each directory name as layer ID. In overlay driver, each layer directory in ALS is treated as the layer's `diff` directory, with assuming `lower` = "" and no link in the `l` directory.

When the `storage.Store` is configured to use ALS, this loads layers from ALS as read-only with assuming they don't have compressed digest nor uncompressed digest. This assumption should be the right thing just because exploded layers aren't archived nor compressed.

`storage.Store` also has layer discovery functionality which is indispensable for enabling lazy pulling. When a layer is queried with an ID via `Layer()` API with annotations (`key1=value1`, `key2=value2`...) and if that ID doesn't exist in ALS, `storage.Store` searches that layer from `.pool` directory with the following path.

```
<ALS root>/.pool/base64("key1=value1")/base64("key2=value2")/.../rootfs.<layer ID>
```

Each path element is base64 encoded key-value pair in the annotation.
The underlying filesystem (e.g. stargz/zstd:chunked-based filesystem or CVMFS) should show the exploded view of the target layer at that location. Example filesystem implementation (currently stargz-based) is https://github.com/ktock/stargz-snapshotter/tree/als-pool-example (this must be mounted under `<ALS root>/.pool`)

If `storage.Store` successfully accessed to this location, this creates the following symlink.

```
<ALS root>/<layer ID> -> <ALS root>/base64(key1=value1)/base64(key2=value2)/.../rootfs.<layer ID>
```

Then this layer (named `<layer ID>`) is accessible by the driver (with `<layer ID>`). So podman don't need to pull this layer from the registry.

### Changes in containers/image: https://github.com/containers/image/pull/1109

Now copier leverages this store. Every time this pulls images, it first tries to reuse blobs using `Layer()` API. This passes all annotations appended to the OCI descriptor of each layer and it also adds the following annotations by default for supporting registry-backed filesystems (e.g. stargz/zstd:chunked).

- `containers/image/target.layerdigest`: The digest of the target layer.
- `containers/image/target.reference`: The image reference of an image that contains the target layer.

When this copier successfully acquired that layer, it reused it without pulling.

### Changes in containers/podman: https://github.com/containers/podman/pull/8837

This patch also fixes libpod to eliminate unnecessary check of digest and size information of each layer
at runtime.

cc: @siscia @giuseppe

### DIFF
--- a/libpod/storage.go
+++ b/libpod/storage.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/containers/image/v5/image"
 	istorage "github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v2/libpod/define"
@@ -87,14 +88,19 @@ func (r *storageService) CreateContainerStorage(ctx context.Context, systemConte
 			return ContainerInfo{}, err
 		}
 		// Pull out a copy of the image's configuration.
-		image, err := ref.NewImage(ctx, systemContext)
+		src, err := ref.NewImageSource(ctx, systemContext)
 		if err != nil {
 			return ContainerInfo{}, err
 		}
-		defer image.Close()
+		defer src.Close()
+		ic, err := image.FromSource(ctx, systemContext, src)
+		if err != nil {
+			return ContainerInfo{}, err
+		}
+		defer ic.Close()
 
 		// Get OCI configuration of image
-		imageConfig, err = image.OCIConfig(ctx)
+		imageConfig, err = ic.OCIConfig(ctx)
 		if err != nil {
 			return ContainerInfo{}, err
 		}


AkihiroSuda:
How will this work with Podman CLI?

ktock:
Assuming `/tmp/storage` as the additional layer store and https://github.com/ktock/stargz-snapshotter/tree/als-pool-example is mounted at `/tmp/storage/.pool` (see also https://github.com/ktock/stargz-snapshotter/blob/als-pool-example/script/demo/run2.sh).

Image can be pulled with skipping downloading layers using `--storage-opt`:

```
podman --storage-opt "layerstore=/tmp/storage" pull ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz
```

This image can run using the same storage option:

```
podman --storage-opt "layerstore=/tmp/storage" run --rm -it ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz /bin/echo 'Hello, World!'
```

Image pulled in this way cannot be exported (e.g. podman saved).
For doing this, that image needs to be re-pulled without `--storage-opt`.

```
podman pull ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz
podman save ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz > /tmp/img.tar
```

libpod/storage.go
```diff
@@ -87,14 +88,18 @@ func (r *storageService) CreateContainerStorage(ctx context.Context, systemConte
...
 		// Pull out a copy of the image's configuration.
-		image, err := ref.NewImage(ctx, systemContext)
+		src, err := ref.NewImageSource(ctx, systemContext)
 		if err != nil {
 			return ContainerInfo{}, err
 		}
-		defer image.Close()
+		ic, err := image.FromSource(ctx, systemContext, src)
+		if err != nil {
+			return ContainerInfo{}, err
```
mtrmac, line 97: This leaks `src`.

mtrmac (commented):
I don’t understand this; `NewImage` is pretty much `NewImageSource+FromSource` already. How does this change behavior?

ktock:
This is for skipping [`*storageImageSource.getSize()`](https://github.com/containers/image/blob/122b16af75d13b9dacda1b04111fa7fdffde30c5/storage/storage_image.go#L1035-L1038) which includes checks of size >= 0 and uncompressedDigest != "" (exploded layers provided through additional layer store aren't archived nor compressed).
Additionally, the size information seems not to be used in `*storageService.CreateContainerStorage()`?

libpod/storage.go
```diff
@@ -87,14 +88,19 @@ func (r *storageService) CreateContainerStorage(ctx context.Context, systemConte
...
 		if err != nil {
 			return ContainerInfo{}, err
 		}
-		defer image.Close()
+		defer src.Close()
+		ic, err := image.FromSource(ctx, systemContext, src)
+		if err != nil {
+			return ContainerInfo{}, err
+		}
+		defer ic.Close()
```
mtrmac, line 100: `FromSource` “takes ownership” of `src`, so now this closes `src` twice.

mtrmac (commented):
> This is for skipping [`*storageImageSource.getSize()`](https://github.com/containers/image/blob/122b16af75d13b9dacda1b04111fa7fdffde30c5/storage/storage_image.go#L1035-L1038) which includes checks of size >= 0 and uncompressedDigest != "" (exploded layers provided through additional layer store aren't archived nor compressed).

Ah, thanks. But that just looks like a design issue to be fixed, rather than worked around here. At the very least CRI-O actually calls `types.Image.Size`, so it should continue to work — and in this case the image object has already been created locally, so there should have been an opportunity to create all the relevant metadata.

Also, `storage.Layer.UncompressedDigest` and the like is a part of the API, not something that can be just ignored and not filled in without reviewing the full ecosystem — and in this case, an essential value without which things like `podman push` would AFAICS stop working.

ktock:
Redesigning not to depend on this patch. (please see https://github.com/containers/image/pull/1109 and https://github.com/containers/storage/pull/795)

ktock:
After several redesigning, this PR is now not needed. 
(containers/image#1109 and containers/storage#795 are still needed)

Issue #8843, opened 2020-12-28
setting http_proxy env on [engine] table results in containers always having http_proxy


ahmadalli:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**
/kind bug

**Description**

**Steps to reproduce the issue:**

1. set `http_proxy` env on `[engine]` table on `/etc/containers/containers.conf
   ```toml
   env = [
        "http_proxy=http://1.2.3.4:5678",
        "https_proxy=http://1.2.3.4:5678",
        "no_proxy=localhost,127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16",
      ]
   ```
1. set `http_proxy` to false on `[containers]` table
   ```toml
   http_proxy = false
   ```
1. check the results
   ```bash
   podman run --rm docker.io/library/alpine printenv http_proxy
   ```

**Describe the results you received:**

the output of the last step would be `http://1.2.3.4:5678`

but if we add `--http-proxy=false` to the command, it works fine

```bash
podman run --rm --http-proxy=false docker.io/library/alpine printenv http_proxy
```

**Describe the results you expected:**

the last step shouldn't have any output

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.2
Built:        Thu Jan  1 03:30:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.22, commit: '
  cpus: 24
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: journald
  hostname: nebula6
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.4.0-58-generic
  linkmode: dynamic
  memFree: 29907927040
  memTotal: 67501182976
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.16.3-fd58-dirty
      commit: fd582c529489c0738e7039cbc036781d1d039014
      spec: 1.0.0
[...]

PR #8844, opened 2020-12-28
Use HTTPProxy settings from containers.conf

rhatdan:
This PR takes the settings from containers.conf and uses
them.  This works on the podman local but does not fix the
issue for podman remote or for APIv2.  We need a way
to specify optionalbooleans when creating containers.

Fixes: https://github.com/containers/podman/issues/8843


Luap99 (commented):
This does not work. The default should be true but if you leave it unset in the config file it defaults to false.

PR #8858, opened 2020-12-31
Expose Height/Width fields to decoder

### LINKED ISSUE #7102: APIv2: 'resize' resizes always to 80 characters

/kind bug

**Description**
Using [containers/resize](https://podman.readthedocs.io/en/latest/_static/api.html#operation/libpodResizeContainer) or [exec/resize](https://podman.readthedocs.io/en/latest/_static/api.html#operation/libpodResizeExec) to resize TTY session ignores `w` parameter and resizes always to 80 characters.

1. 
```
sudo podman run -dt fedora bash
7d481626f41cec219a377c7fb20be9e25d37f9965e9e7e868b9361952ddef597
```
2. `sudo podman attach 7d481626f41cec219a377c7fb20be9e25d37f9965e9e7e868b9361952ddef597` and just type more than 80 characters. You should be able to type as many, as you want.

3. run in different terminal `sudo curl -X POST --unix-socket /run/podman/podman.sock "http://d/v1.24/libpod/containers/7d481626f41cec219a377c7fb20be9e25d37f9965e9e7e868b9361952ddef597/resize?w=200&h=20"` and then in the one where you are still attached you can type only 80 characters and then the line will start to overwrite itself. Feels like the `w` is ignored and 80 is used as default value?

```
Version:      2.0.2
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

```
podman-2.0.2-1.fc32.x86_64
```

jwhonce:
Fixes #7102

### DIFF
--- a/pkg/api/handlers/compat/resize.go
+++ b/pkg/api/handlers/compat/resize.go
@@ -20,8 +20,8 @@ func ResizeTTY(w http.ResponseWriter, r *http.Request) {
 
 	// /containers/{id}/resize
 	query := struct {
-		height uint16 `schema:"h"`
-		width  uint16 `schema:"w"`
+		Height uint16 `schema:"h"`
+		Width  uint16 `schema:"w"`
 	}{
 		// override any golang type defaults
 	}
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

rhatdan:
Not sure how this fixes the issue, but change LGTM
/lgtm

Issue #8506, opened 2020-11-28
Start of rootless container in systemd hangs (Can't open PID file /run/user/1000/containers/overlay-containers/[...].pid (yet?) after start: Permission denied)


topas-rec:
/kind bug

**Description**

Generating systemd files for rootles container. Starting them makes the systemctl call not return. Though the container is started. The service status is still activating.

**Steps to reproduce the issue:**

1. Create a rootless container (with uidmap) and generate systemd file on Debian Bubster

2. Enable and start the service

**Describe the results you received:**
Starting them makes the systemctl call not return. Though the container is started. The service status is still activating.
```
tobias@tobias-pc:~$ systemctl --user start container-181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65.service
^C
tobias@tobias-pc:~$ podman ps -a
CONTAINER ID  IMAGE                               COMMAND               CREATED       STATUS             PORTS                 NAMES
181f37a4d457  docker.io/library/nextcloud:latest  apache2-foregroun...  12 hours ago  Up 12 seconds ago  0.0.0.0:8084->80/tcp  frosty_jang
tobias@tobias-pc:~$ systemctl --user status container-181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65.service
● container-181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65.service - Podman container-181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65.service
   Loaded: loaded (/home/tobias/.config/systemd/user/container-181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65.service; enabled; vendor preset: enabled)
   Active: activating (start) since Sat 2020-11-28 21:02:56 CET; 41s ago
     Docs: man:podman-generate-systemd(1)
  Process: 21849 ExecStart=/usr/bin/podman start 181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65 (code=exited, status=0/SUCCESS)
   CGroup: /user.slice/user-1000.slice/user@1000.service/container-181f37a4d4577fbcf1e7fc2cca6699b7a2906ec3b34134b43404542bda2ffa65.service
[...]

PR #8869, opened 2021-01-04
systemd: make rundir always accessible

giuseppe:
so that the PIDFile can be accessed also without being in the rootless
user namespace.

Closes: https://github.com/containers/podman/issues/8506


rhatdan:
LGTM

Issue #8870, opened 2021-01-04
Pull through API can fail with index error


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When trying to pull image with invalid tag through API often it fails with `index out of range` error.

**Steps to reproduce the issue:**
```
$ sudo curl -X POST -v --unix-socket /run/podman/podman.sock http://d/v1.24/libpod/images/pull?reference=fedora:foobar
*   Trying /run/podman/podman.sock:0...
* Connected to d (/run/podman/podman.sock) port 80 (#0)
> POST /v1.24/libpod/images/pull?reference=fedora:foobar HTTP/1.1
> Host: d
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Api-Version: 1.40
< Libpod-Api-Version: 2.0.0
< Server: Libpod/2.0.0 (linux)
< Date: Mon, 04 Jan 2021 12:01:17 GMT
< Transfer-Encoding: chunked
< 
{"stream":"Resolved short name \"fedora\" to a recorded short-name alias (origin: /etc/containers/registries.conf.d/shortnames.conf)\n"}
{"stream":"Trying to pull registry.fedoraproject.org/fedora:foobar...\n"}
{"stream":"  manifest unknown: manifest unknown\n"}
{}
{"cause":"runtime error: index out of range [-1]","message":"runtime error: index out of range [-1]","response":500}
* Connection #0 to host d left intact
```
 sometimes it however manages to end up with correct error message: (all is the same except the last line)
```
...
{"error":"Error initializing source docker://registry.fedoraproject.org/fedora:foobar: Error reading manifest foobar in registry.fedoraproject.org/fedora: manifest unknown: manifest unknown\n"}
* Connection #0 to host d left intact
```
I can surely reproduce this any time. Sometimes 5 times in a row all fail with index error, sometimes I see the expected error message a few times in a row. But a few retries and I hit it sooner or later.

**Describe the results you expected:**
Always end up with error code 200 and the expected error message.

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 15:37:50 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
[...]

PR #8876, opened 2021-01-04
libpod API: pull: fix channel race

vrothberg:
Fix a race condition in the pull endpoint caused by buffered channels.
Using buffered channels can lead to the context's cancel function to be
executed prior to the items being read from the channel.

Fixes: #8870


mheon:
LGTM

Issue #7387, opened 2020-08-20
`podman load ... [name:tag]` inconsistencies


vrothberg:
The optional `name:tag` for `podman load` is treated differently among the formats. `docker-archive` will apply it as tag to the loaded image while `oci-archive` seems to try to extract it from the image. Let's try to find a way to untangle the hairy ball and get it consistent (or clearly document the different behaviours).

The man pages need to be more explicit:
> You can also specify a name for the image if the archive does not contain a named reference, of if you want an additional name for the local image.

Background discussion: https://github.com/containers/podman/pull/6811/files#r472198035

zhangguanzhang:
```
podman save --format oci-archive alpine:latest > /tmp/FOO.tar
podman load -i /tmp/FOO.tar domain.com/NewName:tag   
```
after this, the image's name should be `domain.com/NewName:tag`? or just search `domain.com/NewName:tag` in the `/tmp/FOO.tar` and load it

vrothberg:
That really depends what we want to do.

I am also open to change `podman load` to not accept such an optional `name:tag`. If users want to load a specific image, they can use `podman pull` instead. That way, `podman load` would be ~identical to `docker load`.

rhatdan:
I think podman load should be identical to docker load.

vrothberg:
Proposal: let's wait first for Podman v2.1 and then remove the optional arguments and get that into Podman v3.0.

Users can always load+tag or use pull to extract a specific image.

vrothberg:
@mtrmac WDYT?

mtrmac:
Note that this is not just the CLI: changes might affect the two `/images/load` APIs (OTOH at least in Swagger the “reference” parameter seems completely undocumented).

> Users can always load+tag or use pull to extract a specific image.

Right now, the CLI load command prints the IDs(?) of loaded images only in a non-trivially-parseable text format.

---

_Conceptually_, I think it does make sense to offer _both_
- Extract only image $name from archive (as opposed to the `docker load` “overwrite an unpredictable set of images in local storage” behavior, making “load” insecure to use with not-100%-trusted archives)
- Tag the result with $dest-tag

OTOH, considering how broken it’s been (we provide sometimes one, sometimes the other, sometimes a broken version of one or the other, and we don’t document which one it is supposed to be) and how there haven’t been many reports complaining about this, removing the feature and waiting for an actual user motivated to contribute one of those features makes sense.

rhatdan:
[...]

PR #8877, opened 2021-01-04
Remove the ability to use [name:tag] in podman load command

rhatdan:
Docker does not support this, and it is confusing what to do if
the image has more then one tag.  We are dropping support for this
in podman 3.0

Fixes: https://github.com/containers/podman/issues/7387


TomSweeneyRedHat:
Changes LGTM,

But, you remove a few tests then the test run times out before it completes?  I've restarted.

rhatdan:
@QiWang19 @vrothberg @mtrmac PTAL

QiWang19 (commented):
LGTM assuming happy tests

cmd/podman/images/load.go
```diff
@@ -71,20 +71,6 @@ func loadFlags(cmd *cobra.Command) {
...
 func load(cmd *cobra.Command, args []string) error {
-	if len(args) > 0 {
-		ref, err := reference.Parse(args[0])
-		if err != nil {
-			return err
-		}
-		if t, ok := ref.(reference.Tagged); ok {
-			loadOpts.Tag = t.Tag()
-		} else {
-			loadOpts.Tag = "latest"
```
mtrmac, line 82: Shouldn’t the `loadOpts.Tag` field, and similarly all the way down (other than fixed APIs), be removed then, as well? Otherwise we won’t get rid of the ambiguity throughout the codebase.

giuseppe (approved):
/lgtm

PR #8878, opened 2021-01-04
Ensure we do not edit container config in Exec

mheon:
The existing code grabs the base container's process, and then modifies it for use with the exec session. This could cause errors in `podman inspect` or similar on the container, as the definition of its OCI spec has been changed by the exec session. The change never propagates to the DB, so it's limited to a single process, but we should still avoid it when possible - so deep-copy it before use.


TomSweeneyRedHat:
LGTM

PR #8899, opened 2021-01-06
Cirrus: Update Fedora & Ubuntu images

cevich:
Includes updates of several key podman dependencies.


edsantiago:
Cross-reference for posterity: https://github.com/containers/buildah/pull/2631

I predict that a lot of the capability subtests in  https://github.com/containers/podman/blob/23f25b8261cf2c0d83f8ce5b8251951a11a7e2c4/test/e2e/run_test.go will fail

cevich:
> Cross-reference for posterity

Oh...do you not see the automatic cross-references that github added (above your comment).

edsantiago:
I see them, but to my eye they're more prominent in a human-entered comment. Some day there'll be another capabilities change, and we'll need to spelunk to figure out how we solved it this time, and human text will catch my eye better than an automatic github xref.

cevich:
> I see them

Oh Okay.  I was just curious if that was something special to the Github WebUI or also translated into CLI tools.

edsantiago:
OK, this isn't something I'll be able to solve. Tests are passing. However, in the [int f33 root host log](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5762402127970304/html/int-podman-fedora-33-root-host.log.html), searching in-page for `CapBnd` shows that different podman commands show different results: some show the `08` value, some show the `a8` one. When tests expect `a8`, they get `a8`. I'm stepping away from this, sorry.

Issue #5278, opened 2020-02-20
Need test to make sure buildah doesn't list "-dev" in its output of podman info


TomSweeneyRedHat:
Add a test to make sure there is never the letters "-dev" in the podman info | grep -i buildah before we release a new podman. Don't want that gating PRs but only releases.

TomSweeneyRedHat:
From #5253 comments

edsantiago:
What difference is there in test environment between regular PR and release-gating? That is, how does the CI environment know "this is a release candidate"?

rhatdan:
Good question.  I guess it would be perfect if we could trigger that before pushing a release.

rhatdan:
Well we hit this, this past week.

edsantiago:
@cevich can you suggest a way to differentiate, within a CI run, between release candidates and regular builds?

cevich:
Yes, but it is (unfortunately) less than reliable because of ambiguity in both the definition of a "release" and inconsistent human behaviors WRT order of push operations and commits.  

Ref: https://cirrus-ci.org/guide/writing-tasks/#environment-variables

* The env. var. `$CIRRUS_TAG` will be set if the currently executing run was triggered because someone pushed a new or updated tag.  The `$CIRRUS_BRANCH` will reflect some random branch name if the tagged commit is on more than one.
* If the build was triggered by a push to a branch, then `$CIRRUS_PR` will be empty string and `$CIRRUS_BRANCH` will be the name of the branch.  `$CIRRUS_TAG` will be empty.

However, For the purpose of switching testing behavior on/off, it **might** be mostly okay to just check if `$CIRRUS_TAG` is set or not.   You can observe builds happening against tags, by using the special URL:
``https://cirrus-ci.com/github/containers/libpod/<tag name>``.

To my knowledge, that's the only place they show up.  If those builds fail, I believe they will post the typical 'post-merge failure' message on IRC.  That could be differentiated by adding a conditional on `$CIRRUS_TAG` being empty or not.  See `contrib/cirrus/notice_branch_failure.sh` and you'll find some more ideas in `contrib/cirrus/success.sh` (limited to PRs by 'only_if' on `success` task in `.cirrus.yml`).

rhatdan:
@cevich Any way we can move forward on this?

rhatdan:
@cevich Friendly Ping.

cevich:
[...]

Issue #8802, opened 2020-12-22
[ci]: ERROR: Releases must never contain '-dev' in output of 'podman info'  (contrib/cirrus/runner.sh:203 in _run_release())


zhangguanzhang:
https://cirrus-ci.com/task/5650026892361728
https://github.com/containers/podman/pull/8793/checks?check_run_id=1592712174
@edsantiago PTAL

edsantiago:
Sorry for the delay, I was on PTO. I can't figure out what this is referring to? It seems like that PR is merged, so you must've figured out the cause?

zhangguanzhang:
> Sorry for the delay, I was on PTO. I can't figure out what this is referring to? It seems like that PR is merged, so you must've figured out the cause?

Before that, almost every step in pr will report an error, you can check the related issue of the link

zhangguanzhang:
related https://github.com/containers/podman/pull/8951

PR #8900, opened 2021-01-06
Cirrus: Skip most tests on tag-push

### LINKED ISSUE #8538: Cirrus-CI Build validation fails on tag-push
/kind bug

When pushing a new tag, Cirrus-CI executes a special job.  A github "check suite" is created by Cirrus-CI, and associated with the tag in github.  However, this "check run" (a.k.a. "build" in Cirrus) is very hard to find in the github WebUI.  It's much easier to observe directly in the Cirrus-CI WebUI using a url of the form `https://cirrus-ci.com/github/containers/podman/<new tag>`.

[Specifically, for the v2.2.0 tag, this Cirrus-CI build failed in the validation step.](https://cirrus-ci.com/task/5693152994525184?command=main#L29)

1. Create a new tag
2. Push tag

The validation step fails due to a bad value being passed to the vbatts/git-validation tool.

The `validate` makefile target should pass for new tags

cevich:
Fixes #8538 #5278

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -199,8 +199,16 @@ function _run_altbuild() {
 }
 
 function _run_release() {
-    if bin/podman info |& grep -Eq -- '-dev'; then
-        die "Releases must never contain '-dev' in output of 'podman info'"
+    # TODO: These tests should come from code external to the podman repo.
+    # to allow test-changes (and re-runs) in the case of a correctible test
+    # flaw or flake at release tag-push time.  For now, the test is here
+    # given it's simplicity.
+
+    msg "Checking podman release (or potential release) criteria."
+    info_output=$(bin/podman info 2>&1)
+    if grep -q -- '-dev'<<<"$info_output"; then
+        die "Releases must never contain '-dev' in output of 'podman info':
+$info_output"
     fi
 }
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -67,6 +67,7 @@ gcp_credentials: ENCRYPTED[a28959877b2c9c36f151781b0a05407218cda646c7d047fc556e4
 ext_svc_check_task:
     alias: 'ext_svc_check'  # int. ref. name - required for depends_on reference
     name: "Ext. services"  # Displayed Title - has no other significance
+    skip: &tags "$CIRRUS_TAG != ''"  # Don't run on tags
     # Default/small container image to execute tasks with
     container: &smallcontainer
         image: ${CTR_FQIN}
@@ -90,7 +91,7 @@ ext_svc_check_task:
 automation_task:
     alias: 'automation'
     name: "Check Automation"
-    skip: &branch "$CIRRUS_PR == '' || $CIRRUS_TAG != ''" # Don't run for branches
+    skip: &branches_and_tags "$CIRRUS_PR == '' || $CIRRUS_TAG != ''" # Don't run on branches/tags
     container: *smallcontainer
     env:
         TEST_FLAVOR: automation
@@ -107,6 +108,7 @@ automation_task:
 smoke_task:
     alias: 'smoke'
     name: "Smoke Test"
+    skip: *tags
     container: &bigcontainer
         image: ${CTR_FQIN}
         # Leave some resources for smallcontainer
@@ -209,6 +211,7 @@ build_task:
 validate_task:
     name: "Validate $DISTRO_NV Build"
     alias: validate
+    skip: *tags
     depends_on:
         - ext_svc_check
         - automation
@@ -237,7 +240,7 @@ bindings_task:
     name: "Test Bindings"
     alias: bindings
     only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
-    skip: *branch
+    skip: *branches_and_tags
     depends_on:
         - build
     gce_instance: *standardvm
@@ -275,6 +278,7 @@ swagger_task:
 vendor_task:
     name: "Test Vendoring"
     alias: vendor
+    skip: *tags
     depends_on:
         - build
     container: *smallcontainer
@@ -373,6 +377,7 @@ osx_alt_build_task:
 docker-py_test_task:
     name: Docker-py Compat.
     alias: docker-py_test
+    skip: *tags
     only_if: *not_docs
     depends_on:
         - build
@@ -393,6 +398,7 @@ docker-py_test_task:
 unit_test_task:
     name: "Unit tests on $DISTRO_NV"
     alias: unit_test
+    skip: *tags
     only_if: *not_docs
     depends_on:
         - validate
@@ -410,6 +416,7 @@ unit_test_task:
 apiv2_test_task:
     name: "APIv2 test on $DISTRO_NV"
     alias: apiv2_test
+    skip: *tags
     depends_on:
         - validate
     gce_instance: *standardvm
@@ -437,6 +444,7 @@ apiv2_test_task:
 compose_test_task:
     name: "compose test on $DISTRO_NV"
     alias: compose_test
+    skip: *tags
     depends_on:
         - validate
     gce_instance: *standardvm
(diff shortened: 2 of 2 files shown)


cevich:
***Note:*** [I'm forcing Cirrus to think this PR was from a tag-push](https://github.com/containers/podman/pull/8900/commits/ffafbdc982eb9d57d3f77837de52e94980eb7655).  This correctly results in [a *Verify Release* task failure](https://cirrus-ci.com/task/5213822464032768?command=main#L6), as [is expected.](https://github.com/containers/podman/issues/5278)

edsantiago:
Here's the map, in case you find it useful. (It's what I'm primarily referring to for my review):
cirrus-map-pr8900

edsantiago:
There's one thing I find unintuitive: there are three `alt_build` tasks (`alt_build`, `osx_alt_build`, `static_alt_build`). Each of those has different SKIPs. I see no way to consolidate those (yes, I tried) but should they maybe share the same SKIPs?

cevich:
> Each of those has different SKIPs

Oh...actually, I think all of the 'alt build' tasks should run for a tag-push...my bad.

cevich:
(rebased and removed the `skip` from the `alt_build_task`)

edsantiago:
Friendly suggestion:
```diff
diff --git a/contrib/cirrus/runner.sh b/contrib/cirrus/runner.sh
index 7f9afd1fd..baf6f31fe 100755
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -199,8 +199,9 @@ function _run_altbuild() {
 }

 function _run_release() {
-    if bin/podman info |& grep -Eq -- '-dev'; then
-        die "Releases must never contain '-dev' in output of 'podman info'"
+    dev=$(bin/podman info |& grep -- -dev)
+    if [[ -n "$dev" ]]; then
+        die "Releases must never contain '-dev' in output of 'podman info' ($dev)"
     fi
 }

```

zhangguanzhang:
related https://github.com/containers/podman/issues/8802

cevich:
> Friendly suggestion:

@edsantiago this is also a valid approach, though I don't see a large benefit to justify changing the code at this time.  I'm assuming that before too long, all the release-processing/testing code will be both expanded and re-implemented.  This PR is just a stop-gap while the group solves the bigger release problems...in case a release needs to go out before improvements are in place.

edsantiago:
>though I don't see a large benefit to justify changing the code at this time.

Okay, I'll change "friendly suggestion" to "kind of important and insistent recommendation".

Here is the workflow:
1. I looked at the test failure
2. It told me "blah blah -dev blah blah"
3. It did not give any useful information whatsoever, so I had no way of knowing what it was complaining about, or what the `-dev` string was

Someone trying to debug this problem might not have the exact same setup as CI. Even if they do, it takes time and energy to check that out and rebuild. Including the offending `-dev` strings in the error output can save someone time. Please have consideration for future debuggers.

cevich:
> Please have consideration for future debuggers.

Oh your contention is about the message, not the checking itself...I'm more than willing to improve that aspect...

cevich:
@edsantiago [improved failure output](https://cirrus-ci.com/task/6151653336285184):

```
Checking podman release (or potential release) criteria.
************************************************
ERROR: Found unexpected '-dev' in output of 'podman info':
host:
  arch: amd64
  buildahVersion: 1.19.0-dev
...cut...
  OsArch: linux/amd64
  Version: 3.0.0-dev  (contrib/cirrus/runner.sh:211 in _run_release())
************************************************
```

contrib/cirrus/runner.sh
```diff
@@ -199,8 +199,16 @@ function _run_altbuild() {
 }
 
 function _run_release() {
-    if bin/podman info |& grep -Eq -- '-dev'; then
-        die "Releases must never contain '-dev' in output of 'podman info'"
+    # TODO: These tests should come from code external to the podman repo.
+    # to allow test-changes (and re-runs) in the case of a correctible test
+    # flaw or flake at release tag-push time.  For now, the test is here
+    # given it's simlicity.
```
edsantiago, line 205: Typo, simplicity. And I really don't understand this comment anyway, but trust that you do and it may help you in the future.

contrib/cirrus/runner.sh
```diff
@@ -199,8 +199,16 @@ function _run_altbuild() {
...
+
+    msg "Checking podman release (or potential release) criteria."
+    info_output=$(bin/podman info 2>&1)
+    if grep -Eq -- '-dev'<<<"$info_output"; then
+        die "Found unexpected '-dev' in output of 'podman info':
```
edsantiago, line 210: Could you leave my patch as it was? `podman info` is an enormous long string, and it is anti-helpful to include the whole thing. The reader only wants/needs to see the `-dev` strings.

BTW I deliberately removed the `-E` in my recommended patch. `-E` is not needed here, and its presence confuses the reader.

edsantiago (changes requested):

contrib/cirrus/runner.sh
cevich: I'm in favor of the more verbose output (unprocessed by `grep`) since (for now) it supports a human catching other release-affecting mistakes, like a wrong/unexpected buildah or conmon version.  At least until we get more/improved release tests in place.

contrib/cirrus/runner.sh
cevich: got it.

contrib/cirrus/runner.sh
edsantiago: I'm afraid I disagree. I cannot imagine your scenario happening in reality: 

Reader: "oh, look, a failure, let me look at the entire output, cross-reference it against my expected versions, oh, wow, an unexpected difference in conmon! Thank goodness for full output".

 What I think is more likely:

Reader: "oh look, a failure, what is the error? Do I actually have to look at ONE HUNDRED AND TWO LINES of crap just to see what the problem is? Okay, I'll use my browser search function, oh darn I used two dashes for --dev, that doesn't find anything, okay let's try again, oh there it is, all right I'll fix it"
(voiceover: what our debugger doesn't know is that there were *two* -devs in the output).

Please just show the relevant error output and only the relevant error output. No distractions please.

cevich:
Okay, tests are green.  I opened a Jira card WRT research and planning for an improved/better release process (and automation) RE: the two issues this PR will close when merged.

contrib/cirrus/runner.sh
edsantiago: I think you missed this (it's been a long week). It is not resolved. Please resolve it.

cevich: I hear what you're saying, but I think we're spending too much time on a relatively minor item.  I'm afraid of making your changes costing a lot of testing time.  In other words, I would need an additional run to verify (fake) `$CIRRUS_TAG` activation works (fails), another run to verify the manual-activation works (fails), then a third run to have a clean CI result for merging.  This is assuming there aren't any actual problems found that need fixing  :confounded: 

IMHO, it's simply not worth our time for a code-path that is used what...three-five times per year, maybe?   I'd rather spend time working improving the overall podman-release experience, and the myriad of other testing issues which need more urgent attention.

edsantiago: Then just remove the changes to runner.sh and resubmit. I'm not going to approve this as it is, it's made everything worse. I'll just fix it myself later.

cevich: What motivation have you provided, for me to approve a re-fix?

 I think for the greater good and grander-schemes...it's past time we agree to disagree.  Our foreheads will thank us for it :smiley:

edsantiago: The fact that I provided a simple, clean, tested fix... which you ignored in favor of a broken one. Come on, this is very simple. Please either use my fix to runner.sh (which, again, I tested fairly extensively) or undo your changes to runner.sh. I provided my reasoning above: dumping the entire `podman info` output is cruel and harmful.

cevich: Ed, we're simply not going to agree on the verbosity level.  But we also need to rise above perfectionism and pettiness, it doesn't help anybody. Since any changes will require re-testing anyway...and this test isn't a long-term solution...here's a compromise:

* I'll remove this automated check (here) and instead include it as a manual check in the release docs (short-term fix).
* I've got a jira card open already, as a "SPIKE" to fully spec. requirements for improved release-related automation and manual processes.  I'd like your assistance completing this (helping define the spec).
* I'd like you to take on the (assumed) followup card for adding (proper) automated release tests (long-term fix).

okay?

edsantiago (approved):
I'll just fix it myself.

PR #8905, opened 2021-01-07
Use HTTPProxy settings from containers.conf

rhatdan:
This PR takes the settings from containers.conf and uses
them.  This works on the podman local but does not fix the
issue for podman remote or for APIv2.  We need a way
to specify optionalbooleans when creating containers.

Fixes: https://github.com/containers/podman/issues/8843

### DIFF
--- a/pkg/specgen/generate/container.go
+++ b/pkg/specgen/generate/container.go
@@ -100,15 +100,9 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 	if err != nil {
 		return nil, err
 	}
-	// First transform the os env into a map. We need it for the labels later in
-	// any case.
-	osEnv, err := envLib.ParseSlice(os.Environ())
-	if err != nil {
-		return nil, errors.Wrap(err, "error parsing host environment variables")
-	}
 
 	// Get Default Environment from containers.conf
-	defaultEnvs, err := envLib.ParseSlice(rtc.GetDefaultEnv())
+	defaultEnvs, err := envLib.ParseSlice(rtc.GetDefaultEnvEx(s.EnvHost, s.HTTPProxy))
 	if err != nil {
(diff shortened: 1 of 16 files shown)


pkg/specgen/generate/container.go
```diff
@@ -100,13 +99,9 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
...
 		return nil, err
 	}
-	// First transform the os env into a map. We need it for the labels later in
-	// any case.
-	osEnv, err := envLib.ParseSlice(os.Environ())
-	if err != nil {
-		return nil, errors.Wrap(err, "error parsing host environment variables")
-	}
 
+	rtc.Containers.EnvHost = s.EnvHost
```
mheon, line 103: This doesn't seem right. We should not be modifying the runtime config once Podman is running - it's asking for all sorts of wierdness in `podman system service` when multiple calls come in at the same time.

rhatdan: This is data specific to the container, not to the general config.

rhatdan: I think I am mistaken, opened a PR for containers/common.

vrothberg (commented):
Needs tests to make sure the flag is really working as intended.

Do we really need to worry about the remote case?  If I understand it correctly, we just expose a `boolean` on the CLI which should also be treated in the remote API.  If set, the backend/server would append the proxy.  Or is it intended to be client side.

rhatdan:
There are httpd_proxy tests already, the only change here was to fix the default in containers.conf to match the reality. But the reality of this patch is just to get the buildah code updated.  I would rather get this in, and then add tests of containers.conf later.

vrothberg (commented):
LGTM

giuseppe (approved):
/lgtm

Issue #8501, opened 2020-11-27
podman stop causes other commands to hang


Hendrik-H:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. create an image with a startup script that just does a sleep infinity and has a trap for SIGTERM so that you have an image that does not stop but needs to be killed

2. start a container using this image

3. invoke something like `podman stop --time 3600 <stop-test-container>`. This should now hang as the container does not stop.

**Describe the results you received:**
commands like `podman container inspect <stop-test-container>` or `podman ps` do hang now until the `podman stop` call is canceled or finished

**Describe the results you expected:**
that the commands do not hang

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.14.9
Built:        Wed Sep 30 21:31:11 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.21-2.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 81d18b6c3ffc266abdef7ca94c1450e669a6a388'
  cpus: 12
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: journald
  hostname: p52
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.9.8-100.fc32.x86_64
  linkmode: dynamic
  memFree: 1059389440
  memTotal: 33335402496
  ociRuntime:
    name: runc
    package: containerd.io-1.3.7-3.1.fc32.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc10
      commit: dc9208a3303feef5b3839f4323d9beb36df0a9dd
      spec: 1.0.1-dev
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
[...]

PR #8906, opened 2021-01-07
container stop: release lock before calling the runtime

vrothberg:
Podman defers stopping the container to the runtime, which can take some
time.  Keeping the lock while waiting for the runtime to complete the
stop procedure, prevents other commands from acquiring the lock as shown
in #8501.

To improve the user experience, release the lock before invoking the
runtime, and re-acquire the lock when the runtime is finished.  Also
introduce an intermediate "stopping" to properly distinguish from
"stopped" containers etc.

Fixes: #8501

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -758,7 +758,7 @@ func (c *Container) isStopped() (bool, error) {
 		return true, err
 	}
 
-	return !c.ensureState(define.ContainerStateRunning, define.ContainerStatePaused), nil
+	return !c.ensureState(define.ContainerStateRunning, define.ContainerStatePaused, define.ContainerStateStopping), nil
 }
 
 // save container state to the database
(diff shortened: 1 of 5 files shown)


edsantiago:
I don't think you ran the BATS tests...

vrothberg:
> I don't think you ran the BATS tests...

That's not very helpful.  At the moment, the tests fail locally even on master on my machine.  Not sure what's going on.

edsantiago:
I'm sorry; I don't know how to help, because I don't understand the change. What first caught my eye was the `stop -t 50` followed by `ps`: there's no way that can work, because the container is run with `--rm` (unless you changed `stop` to not block, and it doesn't look like you did). Then I pulled your PR, ran tests, and even `podman stop basic` fails. I can't really help you debug it, but the tests are deliberately written to be simple so that it should be trivial to reproduce the failure. And indeed, I can easily manually reproduce the `basic` test failure.

mheon:
@vrothberg One suggestion: I think that the container remove code may need a small change to ensure that it treats stopping containers properly, and they can still be removed - I'm concerned about `podman stop` being killed by the user mid-way, and the container becoming stuck in a strange state, unable to be removed.

vrothberg:
Thanks, @edsantiago and @mheon!  Stuck in meetings for the rest of the day.

The test is definitely wrong as is.  I made some progress locally but will continue tomorrow morning.

edsantiago:
It's more than the test. Look at the existing `basic`. The new `stop` is leaving containers in `stopping` state, with status 0 instead of 137. IOW, something is really badly broken.

vrothberg:
> It's more than the test. Look at the existing `basic`. The new `stop` is leaving containers in `stopping` state, with status 0 instead of 137. IOW, something is really badly broken.

Yes, there's more work. Also kill behaved oddly in local tests.

libpod/container_internal.go
```diff
@@ -1284,10 +1284,47 @@ func (c *Container) stop(timeout uint) error {
...
 	if err := c.ociRuntime.StopContainer(c, timeout, all); err != nil {
 		return err
 	}
 
+	if !c.batched {
+		c.lock.Lock()
+		if err := c.syncContainer(); err != nil {
+			// If the container has already been removed (e.g., via
+			// the cleanup process), there's nothing left to do.
+			if errors.Cause(err) == define.ErrNoSuchCtr {
```
mheon, line 1309: Recommend also catching `define.ErrCtrRemoved`

mheon: (Same behaviour, it's a different way of saying the container has been removed)

vrothberg:
@baude @mheon @rhatdan this is looking good now :)

libpod/container_internal.go
```diff
@@ -758,7 +758,7 @@ func (c *Container) isStopped() (bool, error) {
 		return true, err
 	}
 
-	return !c.ensureState(define.ContainerStateRunning, define.ContainerStatePaused), nil
+	return !c.ensureState(define.ContainerStateRunning, define.ContainerStatePaused, define.ContainerStateStopping), nil
```
mheon, line 761: Note to self: in future we should require all of these `ensureState` invocations to not be inverted - will make them safer when we add states, as all valid states will have to be explicitly listed.

vrothberg: +1

vrothberg: Or add private functions for `isRunning` etc.

mheon:
LGTM

PR #8907, opened 2021-01-07
Fix build for mips architecture follow-up 

Luap99:
Follow-up to commit (1ad7966). The build on mips is still
failing because SIGWINCH was not defined in the signal pkg.
Also stat_t.Rdev is unit32 on mips so we need to typecast.

Second, add mips architecture to the cross build target.
I confirmed that the cross compile for mips works now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,11 @@ CROSS_BUILD_TARGETS := \
 	bin/podman.cross.linux.arm \
 	bin/podman.cross.linux.arm64 \
 	bin/podman.cross.linux.386 \
-	bin/podman.cross.linux.s390x
+	bin/podman.cross.linux.s390x \
+	bin/podman.cross.linux.mips \
+	bin/podman.cross.linux.mipsle \
+	bin/podman.cross.linux.mips64 \
+	bin/podman.cross.linux.mips64le
 
 .PHONY: all
 all: binaries docs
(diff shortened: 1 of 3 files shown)


mheon:
LGTM

Luap99:
@edsantiago @cevich I don't see the `local-cross` make target executed in CI. I think we should add this to prevent regression.

TomSweeneyRedHat:
@Luap99 do you want to do the Makefile change in this PR or should we merge this and do that in a follow up?

TomSweeneyRedHat:
LGTM

Luap99:
> @Luap99 do you want to do the Makefile change in this PR or should we merge this and do that in a follow up?

This can be merged. The makefile target exists I just need someone to make the CI use it to prevent a regression. This should be a task for @cevich or @edsantiago. I don't understand the `.cirrus.yml` stuff.

cevich:
> The makefile target exists I just need someone to make the CI use it to prevent a regression. This should be a task for @cevich or @edsantiago. I don't understand the .cirrus.yml stuff.

Sure I'm happy to help add this to CI @Luap99.  Does CI simply need to confirm that `make local-cross` runs w/o returning an error-exit code?

Luap99:
> Sure I'm happy to help add this to CI @Luap99. Does CI simply need to confirm that make local-cross runs w/o returning an error-exit code?

Yes

cevich:
Ref: https://github.com/containers/podman/pull/8939

PR #8910, opened 2021-01-07
Implement the rootless-cni-infra container imageless

### LINKED ISSUE #8709: Proposal: imageless rootless-cni-infra
**EDIT**: this is now tracked in https://github.com/containers/podman/pull/9423
- - -

/kind feature

**Description**

The current dependency on `quay.io/libpod/rootless-cni-infra` has several issues:
- Doesn't work for non-AMD64, unless we can have multi-arch image
- Doesn't work without Internet connectivity
- The origin of the image is unverifiable 

We can solve all these issues by making `rootless-cni-infra` to be imageless.

This can be implemented by executing an equivalent of the following commands in the user namespace:
```console
# mount --rbind / ./rootfs
# mount -t tmpfs none ./rootfs/run
# podman run -d --net=slirp4netns  --rootfs ./rootfs
```

Also we should consider reimplementing `rootless-cni-infra` shell script in Go with `pkg/reexec`.
This can be done separately to reduce complexity of PR, but maybe we should do the both at once to reduce extra migration steps.

**Challenge**

Supporting live upgrade from a previous version might be difficult. 
I think we can choose not to support live upgrade.

- - -

Related:
- https://github.com/containers/podman/pull/8415
- https://github.com/containers/podman/pull/8671

Luap99:
As proposed by Akihiro Suda make the rootless-cni-infra container use
the host rootfs instead of an image. This works by mounting the host
rootfs in the user namespace to `$runroot/rootless-cni-infra`
and use this as rootfs for the container.

Second, rewrite the rootless-cni-infra shell script in go to remove the
extra cnitool dependency which is not packaged anywhere. With that we
only need the same dependencies as rootful podman which should be
already installed.

Advantages:
- Works for all architectures podman supports.
- Works without internet connection.
- No extra maintainence of an extra image.

Disadvantages:
- Requires the dependencies to be available on the host (e.g. dnsname
plugin). The user may not have control over those.

Problems:
- It doesn't unmount the rootfs if the the rootless-cni-infra container
is stopped directly.

Also the image version did not respect the `--cni-config-dir` option
properly. It mounted the cni config dir only at container create time
but this option can be used on podman run commands which did not
worked if the rootless-cni-infra container was already running.
This is only possible with the rootfs version.

Live upgrading is possible. If the old infra container is still
running podman talks via the old api to the script. Once the
old infra container is deleted the new imageless infra container
will be created and podman can talk via the new api. A version
label is added to the container to distinguish between old and new.

Fixes #8709 


Luap99:
@rhatdan I need your selinux knowledge here. On fedora the rootless-cni-infra script is failing with the following error: 
```
failed to list chains: running [/usr/sbin/iptables -t nat -S --wait]: exit status 4: Fatal: can't open lock file /run/xtables.lock: Permission denied
```
This access is blocked by selinux. I see this AVC:
```
type=AVC msg=audit(1610099754.089:998): avc:  denied  { write } for  pid=6975 comm="iptables" name="/" dev="tmpfs" ino=1 scontext=unconfined_u:system_r:iptables_t:s0 tcontext=system_u:object_r:container_file_t:s0:c664,c817 tclass=dir permissive=0
```
The rooless-cni-infra container is created with selinux disabled so I don't understand why selinux is blocking access. `/run` is mounted as tmpfs into the container.

On the host the file has the following label:
```
$ sudo ls -lZ /run | grep xtables
-rw-------.  1 root    root    system_u:object_r:iptables_var_run_t:s0          0  8. Jan 10:51 xtables.lock
```

Interestingly enough it works on my test VM which has the fedora server spin installed and not workstation.
They are shipped with different `iptables` versions. Fedora workstation has `iptables v1.8.5 (legacy)` and fedora server has `iptables v1.8.5 (nf_tables)` both are fedora 33.

rhatdan:
The issue is the iptables command is attempting to create content inside of a directory labeled container_file_t.  Iptables command is confined, and not allowed to write container content.

rhatdan:
Is the /run within the container?  IE This is not the /run on the host?

libpod/rootless_cni_linux.go
```diff
@@ -255,46 +269,65 @@ func startRootlessCNIInfraContainer(ctx context.Context, r *Runtime) (*Container
...
-	inspectData, err := newImage.Inspect(ctx)
-	if err != nil {
+	// need writable /run
+	run := spec.Mount{
+		Destination: "/run",
+		Type:        "tmpfs",
+		Source:      "none",
+		Options:     []string{"rw", "nosuid", "nodev"},
+	}
+	g.AddMount(run)
```
Luap99, line 280: `/run` is mounted as tmpfs into the container

libpod/rootless_cni_linux.go
```diff
@@ -255,46 +269,65 @@ func startRootlessCNIInfraContainer(ctx context.Context, r *Runtime) (*Container
...
-	options = append(options, WithRestartPolicy("always"))
-	options = append(options, WithNetNS(nil, false, "slirp4netns", nil))
+	g.SetRootReadonly(true)
+
+	options := []CtrCreateOption{
+		WithRootFS(rootfs),
+		WithCtrNamespace(rootlessCNIInfraContainerNamespace),
+		WithName(rootlessCNIInfraContainerName),
+		WithPrivileged(true),
+		WithSecLabels([]string{"disable"}),
```
Luap99, line 326: selinux should be disabled for this container so I am not sure why selinux is blocking the access to this file

rhatdan:
Hopefully I will have time early next week to attempt this and figure out what is going on.

contrib/rootless-cni-infra/Readme.md
```diff
@@ -7,19 +7,16 @@ Infra container for CNI-in-slirp4netns.
 When a CNI network is specified for `podman run` in rootless mode, Podman launches the `rootless-cni-infra` container to execute CNI plugins inside slirp4netns.
 
 The infra container is created per user, by executing an equivalent of:
-`podman run -d --name rootless-cni-infra --pid=host --privileged -v $HOME/.config/cni/net.d:/etc/cni/net.d rootless-cni-infra`.
+`podman run -d --name rootless-cni-infra --pid=host --privileged --roofs /`.
```
AkihiroSuda, line 10: typo: roofs

libpod/rootless_cni_linux.go
```diff
@@ -255,46 +275,61 @@ func startRootlessCNIInfraContainer(ctx context.Context, r *Runtime) (*Container
...
+	// need writable /run
+	run := spec.Mount{
+		Destination: "/run",
+		Type:        "tmpfs",
+		Source:      "none",
+		Options:     []string{"rw", "nosuid", "nodev"},
+	}
+	g.AddMount(run)
+
+	// mount var as tmpfs
```
AkihiroSuda, line 288: Shouldn't this be a persistent volume?

libpod/rootless_cni_linux.go
```diff
@@ -308,14 +343,39 @@ func startRootlessCNIInfraContainer(ctx context.Context, r *Runtime) (*Container
...
 }
 
-func rootlessCNIInfraExec(c *Container, args ...string) (string, error) {
-	cmd := "rootless-cni-infra"
+func mountRootlessCNIINfraRootfs(rootfs string) error {
+	if err := os.MkdirAll(rootfs, 0700); err != nil {
+		return err
+	}
+	// bind mount the rootfs in the userns read only
+	if err := mount.Mount("/", rootfs, "bind", "rbind,rprivate,ro"); err != nil {
```
AkihiroSuda, line 351: It should be noted in comment lines that "rbind,ro" is just a "best effort" to make the tree read-only, as "ro" is not recursive.

pkg/rootless/cni/rootless_cni.go
```diff
@@ -0,0 +1,349 @@
...
+		stringPid := strconv.Itoa(pid)
+
+		if err := ioutil.WriteFile(pidfile, []byte(stringPid), 0700); err != nil {
+			exit(errors.Wrap(err, "failed to write pid file"))
+		}
+
+		// set the loopback adapter up
+		cmd := exec.Command("ip", "link", "set", "lo", "up")
+		if err := cmd.Run(); err != nil {
+			exit(err)
```
AkihiroSuda, line 137: should print error from `ip` command

AkihiroSuda: (or use netlink without shelling out `ip`)

libpod/rootless_cni_linux.go
Luap99: Good point. Is it possible to mount everything as readonly?

pkg/rootless/cni/rootless_cni.go
Luap99: I changed this to use netlink.

libpod/rootless_cni_linux.go
Luap99: No I added a comment to explain why

libpod/rootless_cni_linux.go
```diff
@@ -135,65 +182,38 @@ func getCNIPodName(c *Container) string {
...
+	var allocRes rootlesscni.AllocResult
+	allocResStr, err := rootlessCNIInfraExec(infra, stdin, "alloc", id, nw)
 	if err != nil {
 		return nil, err
 	}
-	var cniRes cnitypes.Result
-	if err := json.Unmarshal([]byte(cniResStr), &cniRes); err != nil {
-		return nil, errors.Wrapf(err, "unmarshaling as cnitypes.Result: %q", cniResStr)
+	if err := json.Unmarshal(allocResStr, &allocRes); err != nil {
+		return nil, errors.Wrapf(err, "unmarshaling as cnitypes.Result: %q", allocResStr)
```
AkihiroSuda, line 194: no longer `cnitypes.Result`

libpod/rootless_cni_linux.go
AkihiroSuda: Probably possible with `fuse-overlayfs`. We could also parse `/proc/self/mountinfo` and bind-mount each of the entries with `ro`, but that seems too complicated.

AkihiroSuda: Kernel patch for recursive read-only is here (unmerged): https://lore.kernel.org/linux-fsdevel/20210112220124.837960-7-christian.brauner@ubuntu.com/T/#u

rhatdan: Would be great to get that in.

Luap99:
I reworked this to support live migration from a previous version. If the old infra container is still running podman talks via the old api to the script. Once the old infra container is deleted the new imageless infra container will be created and podman can talk via the new api. A version label is added to the container to distinguish between old and new.

Luap99:
@rhatdan Have you looked at the selinux issue?

For now `sudo alternatives --set iptables /usr/sbin/iptables-nft` would work for fedora but this is a breaking change for users.

libpod/rootless_cni_linux.go
rhatdan: The problem is the iptables command is confined. The question is who launched it? And why did it transition.

Luap99:
@rhatdan 
I tried to look at the selinux issue and found the problem. Look at the following:
```bash
# selinux with container image
$ podman run alpine cat /proc/self/attr/current 
system_u:system_r:container_t:s0:c280,c796
$ podman run --security-opt label=disable alpine cat /proc/self/attr/current 
unconfined_u:system_r:spc_t:s0

# selinux with rootfs
$ mkdir -p ~/rootfs
$ podman unshare mount --rbind -r / ~/rootfs/
$ podman unshare mount -t tmpfs none ~/rootfs/run
$ podman run --rootfs ~/rootfs cat /proc/self/attr/current 
system_u:system_r:container_t:s0:c22,c289
$ podman run --security-opt label=disable --rootfs ~/rootfs cat /proc/self/attr/current 
unconfined_u:system_r:container_runtime_t:s0
# why container_runtime_t and not spc_t?

# set selinux labels manually
# spc_t fails
$ podman run --security-opt "label=user:unconfined_u" --security-opt "label=role:system_r" --security-opt "label=type:spc_t" --security-opt "label=level:s0" --rootfs ~/rootfs  cat /proc/self/attr/current 
{"msg":"exec container process `/usr/bin/cat`: Permission denied","level":"error","time":"2021-02-14T16:08:22.000844579Z"}
# unconfined_t works
$ podman run --security-opt "label=user:unconfined_u" --security-opt "label=role:system_r" --security-opt "label=type:unconfined_t" --security-opt "label=level:s0" --rootfs ~/rootfs  cat /proc/self/attr/current 
unconfined_u:system_r:unconfined_t:s0
```
I don't know if this is a bug or expected. Either way when I set unconfined_u:system_r:unconfined_t for the cni container it works.

rhatdan:
Do you have AVC Messages?

rhatdan:
ls -lZd ~/rootfs/

There are several issues here.  I am thinking that we could grab the "level" label off of the rootfs and set that to run the container with so that we run as container_t, that might be the best solution for this.  Check to make sure that rootfs is labeled container_file_t:MCS and then use the MCS.

rhatdan:
The reason you are seeing the difference on spc_t versus container_runtime_t is based on the label of ~/rootfs.

The reason spc_t is failing, is that there is an entrypoint fule on spc_t, So only certain labels are allowed to "transition" to spc_t.

The label of the cat command is probably not allowed as an entrypoint to spc_t.

Similar we have transition rules that say when podman running as container_runtime_t runs certain executables it will `transition` to spc_t, The label of cat within the rootfs does not have a transition rule, so the transition does not happen, and the process continues to run as container_runtime_t.

Luap99:
OK I don't think I can get this to work. There is a locking issue somewhere. While thinking about it I thought about not using a container at all and just create a netns in the podman user namespace. I have opened https://github.com/containers/podman/pull/9423 to implement this instead. I think this is better in the long run.

baude:
agree

PR #8911, opened 2021-01-07
CI: smoke test: insist on adding tests on PRs

edsantiago:
On each PR (other than CI:DOCS), check the list of git-touched
files, and abort if no tests are added. Include instructions
on how to bypass the check if tests really aren't needed.

Include a hardcoded exception list for PRs that only touch a
well-known subset of "safe" files: docs, .cirrus.yml, vendor.
This list is likely to need tuning, both right now and over time.

Add a test suite.

This is going to be controversial.

This is a github-only comment intended to bypass the new check: No Tests Needed

### DIFF
--- a/contrib/cirrus/pr-should-include-tests
+++ b/contrib/cirrus/pr-should-include-tests
@@ -1,0 +1,70 @@
+#!/bin/bash
+#
+# Intended for use in CI: check git commits, barf if no tests added.
+#
+
+# Docs-only changes are excused
+if [[ "${CIRRUS_CHANGE_TITLE}" =~ CI:DOCS ]]; then
+    exit 0
+fi
+
+# So are PRs where 'NO TESTS NEEDED' appears in the Github message
+if [[ "${CIRRUS_CHANGE_MESSAGE}" =~ NO.TESTS.NEEDED ]]; then
+    exit 0
+fi
+
+# HEAD should be good enough, but the CIRRUS envariable allows us to test
+head=${CIRRUS_CHANGE_IN_REPO:-HEAD}
+# Base of this PR. Here we absolutely rely on cirrus.
+base=$(git merge-base ${DEST_BRANCH:-master} $head)
+
+# This gives us a list of files touched in all commits, e.g.
+#    A    foo.c
+#    M    bar.c
+# We look for Added or Modified (not Deleted!) files under 'test'.
+if git diff --name-status $base $head | egrep -q '^[AM]\s+(test/|.*_test\.go)'; then
+    exit 0
+fi
+
+# Nothing changed under test subdirectory.
+#
+# This is OK if the only files being touched are "safe" ones.
+filtered_changes=$(git diff --name-status $base $head |
+                       awk '{print $2}'               |
+                       fgrep -vx .cirrus.yml          |
+                       fgrep -vx changelog.txt        |
+                       fgrep -vx go.mod               |
+                       fgrep -vx go.sum               |
+                       egrep -v  '^[^/]+\.md$'        |
+                       egrep -v  '^contrib/'          |
+                       egrep -v  '^docs/'             |
+                       egrep -v  '^hack/'             |
+                       egrep -v  '^vendor/'           |
+                       egrep -v  '^version/')
+if [[ -z "$filtered_changes" ]]; then
+    exit 0
+fi
+
+# One last chance: perhaps the developer included the magic '[NO TESTS NEEDED]'
+# string in an amended commit.
+if git log --format=%B ${base}..${head} | fgrep '[NO TESTS NEEDED]'; then
+   exit 0
+fi
+
+cat <<EOF
+$(basename $0): PR does not include changes in the 'tests' directory
+
+Please write a regression test for what you're fixing. Even if it
+seems trivial or obvious, try to add a test that will prevent
+regressions.
+
+If your change is minor, feel free to piggyback on already-written
+tests, possibly just adding a small step to a similar existing test.
+Every second counts in CI.
+
+If your commit really, truly does not need tests, you can proceed
+by adding '[NO TESTS NEEDED]' to the body of your commit message.
+Please think carefully before doing so.
+EOF
+
+exit 1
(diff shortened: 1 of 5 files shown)


edsantiago:
I did not include the magic 'no tests needed' string, so the first pass should fail smoke test. This is deliberate.

edsantiago:
Yay, [it failed](https://cirrus-ci.com/task/6031675102593024) (as expected):
```
 * f2a0009be "CI: smoke test: insist on adding tests on PRs" ... time="2021-01-07T21:20:39Z" level=info msg="[git] cmd: \"git --no-pager log --check f2a0009beb8148745ac8d2f23dc9bf46af43e56d^..f2a0009beb8148745ac8d2f23dc9bf46af43e56d -- . :(exclude)./vendor -- . :(exclude)docs/make.bat\""
PASS
pr-should-include-tests: PR does not include changes in the 'tests' directory

Please write a regression test for what you're fixing. Even if it
seems trivial or obvious, try to add a test that will prevent
regressions.

If your change is minor, feel free to piggyback on existing tests.
Every second counts in CI.

If your commit really, truly does not need tests, you can proceed
by adding 'No test needed' or 'No tests needed' to the body of
your commit message. Please think carefully before doing so.
```
I will now add the magic string to the github description and press Re-run.

edsantiago:
Interesting: updating the github message did not update `$CIRRUS_CHANGE_MESSAGE`. I must be misunderstanding how that works. Re-pushed with update to actual git commit, that worked.

@containers/podman-maintainers PTAL, and please give some thought to what other common cases we need to forgive.

Luap99:
Not all tests are stored in the test directory. The unit test files are called `*_test.go` and are located next to the source code. You should accept these files as well.

edsantiago:
Good point, thank you. Fixed and pushed.

contrib/cirrus/pr-should-include-tests
```diff
@@ -0,0 +1,64 @@
...
+
+Please write a regression test for what you're fixing. Even if it
+seems trivial or obvious, try to add a test that will prevent
+regressions.
+
+If your change is minor, feel free to piggyback on existing tests.
+Every second counts in CI.
+
+If your commit really, truly does not need tests, you can proceed
+by adding 'No test needed' or 'No tests needed' to the body of
```
TomSweeneyRedHat, line 60: Soft suggestion.  I'd change the required string to "[NO TESTS NEEDED]".  Hopefully, it will make the submitter think a bit more, and it will more likely stand out in the description for reviewers.

TomSweeneyRedHat:
I like the idea (may hate it later though when it bites me!  ;^) ), one suggestion for consideration.

contrib/cirrus/pr-should-include-tests
edsantiago: I like that. The brackets are going to be hard to check for cleanly, so I will look into it next week. Thank you!

vrothberg:
I'd love to have one "get out of jail" card similar to `[CI:DOCS]` that allows for merging without a test.  There are so many different kinds of changes that I fear we may miss one and have a red CI for an urgent change.  False positives can be annoying.

An alternative approach could be a GitHub action that drops a comment instead of blocking CI.

edsantiago:
You maybe missed the second sentence of the commit message?

vrothberg:
> You maybe missed the second sentence of the commit message?

I must have. Thanks :-)

LGTM

rhatdan:
Thanks @edsantiago I am going to hate this, but it is like taking your medicine, it is best for us all.
Should we wait to merge this until after 3.0 is released, though so we don't get hung up on this causing us a delay.

contrib/cirrus/pr-should-include-tests
```diff
@@ -0,0 +1,64 @@
...
+# Nothing changed under test subdirectory.
+#
+# This is OK if the only files being touched are "safe" ones.
+filtered_changes=$(git diff --name-status $base $head |
+                       awk '{print $2}'               |
+                       fgrep -vx .cirrus.yml          |
+                       fgrep -vx go.mod               |
+                       fgrep -vx go.sum               |
+                       egrep -v  '^docs/'             |
+                       egrep -v  '^vendor/')
```
mheon, line 38: Add to this: `version/`, `contrib/`, `hack/` - I need all of those for the release process, and I do not want to have to add tests for new releases.

mheon: Further, anything that ends in `.md` in the base directory. We should not be requiring this for readme changes.

rhatdan:
Also should ignore changes to version/version.go

Luap99 (commented):
Please document this in CONTRIBUTING.md and the PR template.

contrib/cirrus/pr-should-include-tests
edsantiago: Done; also changelog.txt.

contrib/cirrus/pr-should-include-tests
edsantiago: Done and tested.

edsantiago:
>Please document this in CONTRIBUTING.md and the PR template

CONTRIBUTING.md : done, thanks for the suggestion. PR template: not done, I'm actually reluctant to do so. I fear it would give new contributors license to just add the magic string without even thinking. I prefer to have CI fail outright (it will do so quickly), this forces them to go through the effort of a new push; and, I hope, to actually consider adding tests.

edsantiago:
>Should we wait to merge this until after 3.0 is released

As soon as this merges, we're going to have a lot of angry devs. Release timing will increase that anger. So, yes, please defer until the team thinks most appropriate.

rhatdan:
I am fine with doing this after day 3.0 ships.
We will have to react quickly and be there to help contrinbutors who just don't have the time to write tests.

vrothberg:
Sounds good to me. Great work, @edsantiago!

rhatdan:
/hold
do not merge until post 3.0

edsantiago:
Rebased, and updated `CONTRIBUTING.md` with instructions on bypassing the block. I again looked at the [PR template](https://github.com/containers/podman/blob/8c6df5e93e6941a7f50da651678751f7dfec900e/.github/PULL_REQUEST_TEMPLATE.md) and again opt not to include instructions there, since it's really just a short-and-sweet link to `CONTRIBUTING.md`.

And, added `tests-included` Makefile target, now required by `validate`. I'm not 100% sure this one will work as desired, I think it may result in false positives, but there's really no way to test it other than in the field. Please let me know if you see this trigger inappropriately.

PR #8915, opened 2021-01-08
Improve error message when the the podman service is not enabled

rhatdan:
Currently if server is not connected, we return an error message that
is confusing users on Mac and Windows boxes.  The hope here is to make
it a little easier to discover that a Podman service is required.

This message is similar to what Docker puts out so people might under
stand it better.

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -158,7 +158,7 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 
 	// Prep the engines
 	if _, err := registry.NewImageEngine(cmd, args); err != nil {
-		return err
+		return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure there is a Podman REST API service running.")
 	}
 	if _, err := registry.NewContainerEngine(cmd, args); err != nil {
 		return err


rhatdan:
@afbjorklund  PTAL

cmd/podman/root.go
```diff
@@ -158,7 +158,7 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 
 	// Prep the engines
 	if _, err := registry.NewImageEngine(cmd, args); err != nil {
-		return err
+		return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure their is a podman service running.")
```
TomSweeneyRedHat, line 161: I'm not completely sold on the "REST API" bit I've added, but I think it brings clarity.  Another wording perhaps?
```suggestion
		return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure there is a Podman REST API service running.")
```

afbjorklund:
Works for me.

`Error: Cannot connect to the Podman socket, make sure there is a Podman REST API service running.: Get "http://d/v2.0.0/libpod/_ping": dial unix ///run/user/1000/podman/podman.sock: connect: no such file or directory`

Similar enough to:

`Cannot connect to the Docker daemon at unix:///run/user/1000/docker.sock. Is the docker daemon running?`

rhatdan:

Luap99 (commented):
LGTM

Issue #8865, opened 2021-01-02
Errors are not reported when starting a pod from the REST API


fstephany:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Errors are not reported when starting a pod from the REST API. 

**Steps to reproduce the issue:**

1. Start a Pod using a port mapping (e.g., 8080:80).
2. Start another pod using the same port mapping as the previous one with the CLI. This result in an expected bind error:

```
$ podman pod start second-pod 
error starting container 47ab04855d52b81fc4245e4abb8a6f04933518e865d0aca695b65effc3193aee: rootlessport listen tcp 0.0.0.0:8080: bind: address already in use
Error: error starting container d6efdf7a7d2366e8af67deea417a98c3235fdcccbb8ecf99ca4babb23782e91c: a dependency of container d6efdf7a7d2366e8af67deea417a98c3235fdcccbb8ecf99ca4babb23782e91c failed to start: container state improper
```

3. Start a podman daemon:  ` $ podman system service -t 0 tcp:127.0.0.1:1234`
4. Try to start the same pod with the REST API:

```
$ curl -X POST -s 'http://localhost:1234/v2.0.0/libpod/pods/second-pod/start'
{"Errs":[{},{}],"Id":"3568556cd9b293a53f2ce3ff98d9bea58929a7f2ac41cebf575f958079dcf061"}
```

**Describe the results you received:**

The `Errs` property contains empty objects instead of the expected string (as described in [the reference](https://docs.podman.io/en/latest/_static/api.html#operation/startPod)).

**Output of `podman version`:**

Podman running on Fedora 33. 

```
$ podman version
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 15:37:50 2020
OS/Arch:      linux/amd64
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

No

rhatdan:
@jwhonce PTAL

baude:
we are getting a partial fail error back in this case.  but then we still return an OK.  That is easy to fix, but it also looks like the encoder is eating/throwing away the slice of errs.  i need @jwhonce to advise here and then fix should be easy.

PR #8917, opened 2021-01-08
Ensure that `podman play kube` actually reports errors

mheon:
In 2.2.x, we moved `play kube` to use the Start() API for pods, which reported errors in a different way (all containers are started in parallel, and then results reported as a block). The migration attempted to preserve compatibility by returning only one error, but that's not really a viable option as it can obscure the real reason that a pod is failing. Further, the code was not correctly handling the API's errors - Pod Start() will, on any container error, return a map of container ID to error populated for all container errors *and* return ErrPodPartialFail for overall error - the existing code did not handle the partial failure error and thus would never return container errors.

Refactor the `play kube` API to include a set of errors for containers in each pod, so we can return all errors that occurred to the frontend and print them for the user, and correct the backend code so container errors are actually forwarded.

### DIFF
--- a/pkg/domain/entities/play.go
+++ b/pkg/domain/entities/play.go
@@ -40,6 +40,9 @@ type PlayKubePod struct {
 	Containers []string
 	// Logs - non-fatal errors and log messages while processing.
 	Logs []string
+	// ContainerErrors - any errors that occurred while starting containers
+	// in the pod.
+	ContainerErrors []string
 }
 
 // PlayKubeReport contains the results of running play kube.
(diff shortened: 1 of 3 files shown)


mheon:

pkg/domain/entities/play.go
```diff
@@ -40,6 +40,9 @@ type PlayKubePod struct {
 	Containers []string
 	// Logs - non-fatal errors and log messages while processing.
 	Logs []string
+	// ContainerErrors - any errors that occurred while starting containers
+	// in the pod.
+	ContainerErrors []error
```
mheon, line 45: I'm deliberately keeping this separate from Logs, because Logs being set doesn't change exit code, and this does

rhatdan:
LGTM

pkg/domain/entities/play.go
Luap99: This will not work via remote for the same reason as https://github.com/containers/podman/issues/8865

mheon: Fixed

ashley-cui (commented):
LGTM

PR #8920, opened 2021-01-09
podman ps/pod ps add network filter and .Networks format placeholder

Luap99:
Add network filter for podman ps and pod ps
Allow to filter on the network name or full id.
For pod ps it will filter on the infra container networks.

Add Networks format placeholder to podman ps and pod ps
`podman ps --format {{.Networks}}` will show all connected networks for
this container. For `pod ps` it will show the infra container networks.


mheon:
LGTM

PR #8935, opened 2021-01-11
oci: keep LC_ env variables to conmon

giuseppe:
it is necessary for conmon to deal with the correct locale, otherwise
it uses C as a fallback.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1893567
Requires: https://github.com/containers/conmon/pull/215

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -1308,7 +1308,12 @@ func prepareProcessExec(c *Container, options *ExecOptions, env []string, sessio
 // configureConmonEnv gets the environment values to add to conmon's exec struct
 // TODO this may want to be less hardcoded/more configurable in the future
 func (r *ConmonOCIRuntime) configureConmonEnv(ctr *Container, runtimeDir string) ([]string, []*os.File) {
-	env := make([]string, 0, 6)
+	var env []string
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM

Any chance we can add a test?

rhatdan:
LGTM

haircommander:
I was wondering why my fix didn't work. thanks for picking this up @giuseppe :)

mheon:
I'm going to go ahead and merge - if a test gets added, we can do so in a further PR.
/lgtm

giuseppe:
> I'm going to go ahead and merge - if a test gets added, we can do so in a further PR.

I think it would be better to have an updated conmon first so we can add an e2e test

mheon:
/hold
Fair enough. Definitely need this in by Monday though.

rhatdan:
I have merged the fix in conmon, now we need @haircommander to cut a new release and then @cevich to create a new VM to testing.

rhatdan:
I have no problem with this being merged, so it will not effect anything, but it would be nice to make sure everything works correctly

haircommander:
I have to deal with the backwards compat problem I mentioned last week, then I'll cut a release.

PR #8937, opened 2021-01-11
Quiet v2.0 branch

cevich:
Running the full suite of tests is no longer needed on this branch according to @mheon.  Disable extensive testing and notifications, leaving just the bare-bones.  In case this is a mistake, the The Cirrus-Cron job will remain enabled, in order to keep the VM images from getting purged.  As long as the images are available, testing may be re-enabled if needed.


baude:
LGTM

cevich:
Ooof, I think this might need to be manually merged b/c the silly RDO failure.

mheon:
It passed all the required tasks. I pushed the button. Branch protection changes were not required.

PR #8942, opened 2021-01-12
Allow podman push to push manifest lists

rhatdan:
When doing a podman images, manifests lists look just like images, so
it is logical that users would assume that they can just podman push them
to a registry.  The problem is we throw out weird errors when this happens
and users need to somehow figure out this is a manifest list rather then
an image, and frankly the user will not understand the difference.

This PR will make podman push just do the right thing, by failing over and
attempting to push the manifest if it fails to push the image.


cmd/podman/images/push.go
```diff
@@ -147,5 +151,14 @@ func imagePush(cmd *cobra.Command, args []string) error {
 
 	// Let's do all the remaining Yoga in the API to prevent us from scattering
 	// logic across (too) many parts of the code.
-	return registry.ImageEngine().Push(registry.GetContext(), source, destination, pushOptions.ImagePushOptions)
+	err := registry.ImageEngine().Push(registry.GetContext(), source, destination, pushOptions.ImagePushOptions)
+	if err != nil {
+		if errors.Cause(err) != storage.ErrImageUnknown {
```
vrothberg, line 156: Having the error check here is unlikely to work in the remote client since we lose the type information over the wire.

I suggest something different and handle everything in `ImageEngine.Push` where we could do something as follows:
```golang
image := lookupLocalImage(...)
if isManifestList(image) && options.All {
    // Push manifest list and all images instead of just one image.
}
```

cmd/podman/manifest/push.go
```diff
@@ -41,7 +42,7 @@ func init() {
 		Parent:  manifestCmd,
 	})
 	flags := pushCmd.Flags()
-	flags.BoolVar(&manifestPushOpts.Purge, "purge", false, "remove the manifest list if push succeeds")
+	flags.BoolVar(&manifestPushOpts.Rm, "rm", false, "remove the manifest list if push succeeds")
```
vrothberg, line 45: This should be a separate change IMHO (non blocking).

docs/source/markdown/podman-push.1.md
```diff
@@ -53,6 +54,11 @@ Images are pushed from those stored in local image storage.
 
 ## OPTIONS
 
+#### **--all**
+
+If specified image is a manifest list or image index, push the images in addition to
```
vrothberg, line 59:
```suggestion
If the specified image is a manifest list or an image index, push the images in addition to
```

docs/source/markdown/podman-push.1.md
```diff
@@ -100,6 +106,10 @@ When writing the output image, suppress progress output
 
 Discard any pre-existing signatures in the image
 
+#### **--rm**
+
+When pushing a the manifest list or image index, delete them from local storage if pushing succeeds.
```
vrothberg, line 111:
```suggestion
When pushing a manifest list or an image index, delete them from local storage if pushing succeeds.
```

Will all included images be deleted as well?

cmd/podman/images/push.go
```diff
@@ -100,6 +103,7 @@ func pushFlags(cmd *cobra.Command) {
 	_ = cmd.RegisterFlagCompletionFunc(formatFlagName, common.AutocompleteManifestFormat)
 
 	flags.BoolVarP(&pushOptions.Quiet, "quiet", "q", false, "Suppress output information when pushing images")
+	flags.BoolVar(&pushOptions.Rm, "rm", false, "remove the manifest list if push succeeds")
```
vrothberg, line 106: I think this should really be a different PR (and discussion).  I have mixed feelings about it as I think users should really be using `podman manifest remove`.   Users may ask to remove normal images at some point and then want `--rm-force` etc.  I prefer to draw the lines early to not open too many doors for complexity.

vrothberg (commented):
I really like the `push --all` idea but am rather rejective toward the `--rm` (see comments).

rhatdan:
The --purge -> --rm came from @nalind in buildah, and I was just going for compatibility.

cmd/podman/images/push.go
rhatdan: You got me excited that an isManifestList object existed.

vrothberg: Ah, sorry!  There should be code checking for the manifest types though.

rhatdan: I actually rewrote a lot of this to be server side, so it should have better support for errors.Cause.

pkg/api/handlers/libpod/images.go
```diff
@@ -462,31 +457,21 @@ func PushImage(w http.ResponseWriter, r *http.Request) {
...
+	options := entities.ImagePushOptions{
+		Authfile: authfile,
+		Username: authConf.Username,
+		Password: authConf.Password,
+		Format:   query.Format,
 	}
 	if sys := runtime.SystemContext(); sys != nil {
-		dockerRegistryOptions.DockerCertPath = sys.DockerCertPath
-		dockerRegistryOptions.RegistriesConfPath = sys.SystemRegistriesConfPath
+		options.CertDir = sys.DockerCertPath
```
vrothberg, line 467: RegistriesConfPath must be set as well

pkg/domain/infra/abi/images.go
```diff
@@ -380,39 +380,15 @@ func (ir *ImageEngine) Push(ctx context.Context, source string, destination stri
 		signOptions,
 		&dockerRegistryOptions,
 		nil)
+	if err != nil && errors.Cause(err) != storage.ErrImageUnknown {
```
vrothberg, line 383: I still think that's wrong.  We're guessing here although we could know whether it's manifest list or image index. The `--all` flag is currently ignored as well. 

```golang
if options.All && imageIsAManifestListOrImageIndex {
   return ir.ManifestPush(...)
}
err = newImage.PushImageTo....
```
That would require to wire `--all` into the remote API but I think it's worth it.

The "guessing" may also yield very different error messages. A `podman push nonexistant` may now return something about manifest lists.

nalind: Or default to `--all` and have an option to push just the bare manifest list.  This stuff isn't intended to be set in stone on the buildah side until we're more sure we've gotten it right.

pkg/api/handlers/libpod/images.go
rhatdan: Neither of these fields are supported from the remote side, so they both should be dropped and allow the server to handle it.  I beleive.

cmd/podman/images/push.go
TomSweeneyRedHat: Docker doesn't have a similar option. I think I lean towards @vrothberg's thoughts here.

cmd/podman/images/push.go
```diff
@@ -75,6 +75,7 @@ func init() {
 func pushFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 
+	flags.BoolVar(&pushOptions.All, "all", false, "push all of the images referenced by the manifest list")
```
TomSweeneyRedHat, line 78: Docker has an `-a/--all-tags` option that pushes all the tags.  I'm leery about naming this `--all` especially if we shorten it to `-a`.  They also have a `manifest push` command.  Should we make "all" push tags and manifests?

vrothberg (commented):
Aren't we missing the wiring in the REST API? I'd feel more comfortable we could make this 3.1 thing.

rhatdan:
There is no new API here other then passing down the --all flag.

rhatdan:
The issue here is I am removing the bindings for manifest push, which would be a breaking API, and would need to wait a year.  There is limited risk here since no one has ever used this API since it is brand new and I don't want to release broken API.

vrothberg:
> The issue here is I am removing the bindings for manifest push, which would be a breaking API, and would need to wait a year. There is limited risk here since no one has ever used this API since it is brand new and I don't want to release broken API.

We don't need to remove these bindings IMHO.

pkg/domain/infra/abi/images.go
rhatdan: This is not ignored, it is wired into the podman push command.

pkg/domain/infra/abi/images.go
```diff
@@ -379,39 +379,15 @@ func (ir *ImageEngine) Push(ctx context.Context, source string, destination stri
...
+		// Image might be a manifest so attempt a manifest push
```
nalind, line 383: ... might be a manifest list...

rhatdan:
@giuseppe @vrothberg @saschagrunert PTAL
This is ready to go in, and we can finish the Release.
@vrothberg Removed --all as a option for podman push.

giuseppe (approved):
LGTM

cmd/podman/images/push.go
```diff
@@ -75,6 +75,8 @@ func init() {
 func pushFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 
+	// For now default All flag to true, for pushing of manifest lists
+	pushOptions.All = true
```
vrothberg, line 79: `.All` is set but not used in the push code. But we can merge as is.

vrothberg (commented):
LGTM

PR #8946, opened 2021-01-12
Expose security attribute errors with their own messages

JAORMX:
This creates error objects for runtime errors that might come from the
runtime. Thus, indicating to users that the place to debug should be in
the security attributes of the container.

When creating a container with a SELinux label that doesn't exist, we
get a fairly cryptic error message:

```
$ podman run --security-opt label=type:my_container.process -it fedora bash
Error: OCI runtime error: write file `/proc/thread-self/attr/exec`: Invalid argument
```

This instead handles any errors coming from LSM's `/proc` API and
enhances the error message with a relevant indicator that it's related
to the container's security attributes.

A sample run looks as follows:

```
$ bin/podman run --security-opt label=type:my_container.process -it fedora bash
Error: `/proc/thread-self/attr/exec`: OCI runtime error: unable to assign security attribute
```

With `debug` log level enabled it would be:

```
Error: write file `/proc/thread-self/attr/exec`: Invalid argument: OCI runtime error: unable to assign security attribute
```

Note that these errors wrap ErrOCIRuntime, so it's still possible to to
compare these errors with `errors.Is/errors.As`.

One advantage of this approach is that we could start handling these
errors in a more efficient manner in the future.

e.g. If a SELinux label doesn't exist (yet), we could retry until it
becomes available.

### DIFF
--- a/libpod/define/errors.go
+++ b/libpod/define/errors.go
@@ -2,6 +2,7 @@ package define
 
 import (
 	"errors"
+	"fmt"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


mheon:
/approve
@rhatdan PTAL

rhatdan:
@JAORMX Wouldn't this code be better in github.com/opencontainers/selinux?

rhatdan:
That way buildah, Docker, and other container engines could use it.

JAORMX:
> @JAORMX Wouldn't this code be better in https://github.com/opencontainers/selinux?

It would be! and that's the direction that I initially tried to take. However, in some flows we won't see that error when setting the label in exec. I would have expected https://github.com/opencontainers/selinux/blob/master/go-selinux/selinux_linux.go#L677 to throw an error when one uses a selinux label that doesn't exist. But that's not the case, seems that an error only manifests itself after `execve`, e.g. after conmon is spawned. And hence why I submitted this as it is. It was the only way I figured to capture the error.

rhatdan:
Ok, understood.
LGTM

rhatdan:

PR #8947, opened 2021-01-12
Fix problems reported by staticcheck

Luap99:
`staticcheck` is a golang code analysis tool. https://staticcheck.io/

This commit fixes a lot of problems found in our code. Common problems are:
- unnecessary use of fmt.Sprintf
- duplicated imports with different names
- unnecessary check that a key exists before a delete call

There are still a lot of reported problems in the test files but I have
not looked at those.

### DIFF
--- a/libpod/image/pull.go
+++ b/libpod/image/pull.go
@@ -11,7 +11,6 @@ import (
 	cp "github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/directory"
 	"github.com/containers/image/v5/docker"
-	"github.com/containers/image/v5/docker/archive"
 	dockerarchive "github.com/containers/image/v5/docker/archive"
 	ociarchive "github.com/containers/image/v5/oci/archive"
 	oci "github.com/containers/image/v5/oci/layout"
@@ -130,7 +129,7 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
 
 // getPullRefPairsFromDockerArchiveReference returns a slice of pullRefPairs
 // for the specified docker reference and the corresponding archive.Reader.
-func (ir *Runtime) getPullRefPairsFromDockerArchiveReference(ctx context.Context, reader *archive.Reader, ref types.ImageReference, sc *types.SystemContext) ([]pullRefPair, error) {
+func (ir *Runtime) getPullRefPairsFromDockerArchiveReference(ctx context.Context, reader *dockerarchive.Reader, ref types.ImageReference, sc *types.SystemContext) ([]pullRefPair, error) {
 	destNames, err := reader.ManifestTagsForReference(ref)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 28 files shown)


libpod/image/pull.go
```diff
@@ -70,7 +69,7 @@ type cleanUpFunc func() error
 type pullGoal struct {
 	refPairs     []pullRefPair
 	pullAllPairs bool                 // Pull all refPairs instead of stopping on first success.
-	cleanUpFuncs []cleanUpFunc        // Mainly used to close long-lived objects (e.g., an archive.Reader)
+	cleanUpFuncs []cleanUpFunc        // Mainly used to close long-lived objects (e.g., andockerarchive.Reader)
```
rhatdan, line 72: comment is screwed up.

libpod/image/pull.go
```diff
@@ -129,8 +128,8 @@ func (ir *Runtime) getSinglePullRefPairGoal(srcRef types.ImageReference, destNam
 }
 
 // getPullRefPairsFromDockerArchiveReference returns a slice of pullRefPairs
-// for the specified docker reference and the corresponding archive.Reader.
-func (ir *Runtime) getPullRefPairsFromDockerArchiveReference(ctx context.Context, reader *archive.Reader, ref types.ImageReference, sc *types.SystemContext) ([]pullRefPair, error) {
+// for the specified docker reference and the correspondingdockerarchive.Reader.
```
rhatdan, line 131: comment is screwed up.

rhatdan (commented):
Nice improvement but a couple of nits.

mheon:
LGTM once comments addressed

pkg/specgen/generate/validate.go
```diff
@@ -48,7 +48,7 @@ func verifyContainerResourcesCgroupV1(s *specgen.SpecGenerator) ([]string, error
 				warnings = append(warnings, "Your kernel does not support memory swappiness capabilities, or the cgroup is not mounted. Memory swappiness discarded.")
 				memory.Swappiness = nil
 			} else {
-				if *memory.Swappiness < 0 || *memory.Swappiness > 100 {
```
TomSweeneyRedHat, line 51: Why is the less than zero check removed?

Luap99: The swappiness type is uint64 and therefore cannot be less than zero.

TomSweeneyRedHat: I'd just be concerned that someone somewhere down the road decided to change the variable to a signed int.  I'd keep the check unless it causes real headaches.  Others may have differing opinions.

vrothberg (commented):
LGTM

pkg/specgen/generate/validate.go
Luap99: `Swappiness` is defined [here](https://github.com/opencontainers/runtime-spec/blob/7413a7f753e1bd9a6a9c6dc7f96f55888cbbd476/specs-go/config.go#L308). It is part of the OCI runtime specification. I highly doubt that this will ever be changed to a signed integer.

PR #8950, opened 2021-01-12
Exorcise Driver code from libpod/define

mheon:
The libpod/define code should not import any large dependencies, as it is intended to be structures and definitions only. It included the libpod/driver package for information on the storage driver, though, which brought in all of c/storage. Split the driver package so that define has the struct, and thus does not need to import Driver. And simplify the driver code while we're at it.

### DIFF
--- a/libpod/container_inspect.go
+++ b/libpod/container_inspect.go
@@ -49,7 +49,7 @@ func (c *Container) Inspect(size bool) (*define.InspectContainerData, error) {
 	return c.inspectLocked(size)
 }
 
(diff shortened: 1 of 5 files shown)


mheon:
If we were develop a `libpod/utils` package, `driver` is a prime candidate for consolidating into that package. I don't really want to put it in `pkg/util` because I feel like that might be used by `cmd/podman` or the bindings...

We really need to come up with consistent location and naming for CLI-safe and non-CLI-safe util functions.

vrothberg (commented):
LGTM

PR #8951, opened 2021-01-12
CI: fix broken diagnostic message for -dev check

edsantiago:
There's a CI check for the presence of "-dev" in podman-info output
(it should not appear). This test is unlikely to fail, but if it
ever does, the diagnostic output is unhelpful. This makes it helpful.

Tested via:

    $ ln -s /bin/echo ~/bin/msg
    $ ln -s /bin/echo ~/bin/die
    $ TEST_FLAVOR=release ./contrib/cirrus/runner.sh
    ...
    Releases must never contain '-dev' in output of 'podman info' ( buildahVersion: 1.19.0-dev
      Version: 3.0.0-dev)

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -203,16 +203,18 @@ function _run_altbuild() {
 
 function _run_release() {
     # TODO: These tests should come from code external to the podman repo.
-    # to allow test-changes (and re-runs) in the case of a correctible test
+    # to allow test-changes (and re-runs) in the case of a correctable test
     # flaw or flake at release tag-push time.  For now, the test is here
-    # given it's simplicity.
+    # given its simplicity.
+    msg "podman info:"
+    bin/podman info
 
     msg "Checking podman release (or potential release) criteria."
-    info_output=$(bin/podman info 2>&1)
-    if grep -q -- '-dev'<<<"$info_output"; then
-        die "Releases must never contain '-dev' in output of 'podman info':
-$info_output"
+    dev=$(bin/podman info |& grep -- -dev)
+    if [[ -n "$dev" ]]; then
+        die "Releases must never contain '-dev' in output of 'podman info' ($dev)"
     fi
+    msg "All OK"
 }
 
 logformatter() {


cevich (changes requested):
This is a cosmetic change which unnecessarily hides vital information from the developer at a potentially critical point in time:  There is no opportunity to "fix" a test run triggered by a tag-push.  We either provide all the potentially helpful detail, or we obscure.  I cannot imagine any situation where withholding this detail would be helpful...especially for an extremely rarely used test.

edsantiago:
@cevich we have a major misunderstanding. I would like to ask you to please take a step back, take some time, and look at this again tomorrow or a future day. I think you are gravely misunderstanding the point of this fix.

rhatdan:
LGTM

edsantiago:
Thanks Dan. I need to find a way to explain this to @cevich, and I've actually been struggling with it much of yesterday evening. Chris, I still don't understand your objection to this, which means I may not be addressing it. Here are my attempts to communicate the reasoning behind my patch:

* Simplicity. An error message should convey the information needed to identify and resolve the problem. 
* Kindness. It will be hard for a human being to find the actual error message among a hundred-plus-line info dump.
  * Even worse: Cirrus only shows the last ten or twenty lines of a failure. As the code is now, that means everyone seeing the failure will need to scan the Cirrus output, realize that it doesn't include anything useful, click the "Cirrus" link, scroll to the bottom, then tediously scroll up looking for the actual error message. This is not only cruel, it wastes time and energy.
* Error-proneness. Even if someone spends that time tracking down the error, they then need to understand what the problem is ("-dev" in output), then since few human beings can find strings that quickly in a hundred-line data dump, they need to engage their browser's in-page search function, type "dev", which will probably find way too many false hits, then retype "-dev" with the dash, find a match, and only then can they figure out the problem. **BUT** a human will probably stop at this point; and if there are two "-dev" matches, say buildah and podman, they will only find one, rebuild, and fail again. With my clean approach, all "-dev" strings are shown in one place.
* Human psychology. A hundred-line info dump is cognitive overload. Yes, theoretically a human could go through all the steps above, track down the "-dev" error, and in the process catch some other unrelated error in the info dump, and by doing so catch a problem that would have otherwise gone unnoticed. I humbly submit that the chances of this happening are near zero.

With my patch: should it trigger, the error will appear in the bottom of the Cirrus error page. It will give a description of the error ("-dev in output") and will show exactly the offending output (buildah, podman, any other future such tool). The developer will likely figure out the problem in a few seconds.

Does this address your objections?

cevich:
@edsantiago it's much simpler than that.  Borrowing you're bulleted format....

* We're not going to agree on the `podman info` output verbosity.  Your changes will obscure unlikely but critical output, at a critical moment.  A contrived example: `Error reading containers.conf: invalid format`.  But the output could literally be anything not containing '-dev' and still critical for a developer to observe.
* I want this to become dead-code (and removed) before too long.  I have a jira card open already, to investigate and improve the overall release process, including automation.  i.e. this task and test only addresses a (hopefully) short-term testing gap.
* Perhaps most important, I made several attempts at negotiating a compromise, all were dismissed out-of-hand in favor of playing games with PRs :cry: 
* We've already spent IMHO ***WAY*** too much time on this.  However, in good-faith, here's another compromise:  Log the output of `podman info` to a file, do your grep & test, then store the full log as an *artifact* for (just in case) examination.

cevich:
[Output example (from `Optional Release Test`)](https://cirrus-ci.com/task/4790214570803200) for reference.

edsantiago:
I'm trying really hard to understand.

We have a test here whose purpose is to fail if the string "-dev" appears in the output of "podman info". If "-dev" does not appear in the output of "podman info", the test passes. This has been the case for at least many months. Do we agree on this? I don't believe it's possible to disagree on this, but I'm just feeling a little out of touch with reality.

Now the hard part. You believe that, if the above test fails, there could be further information to be gleaned from the full output of "podman info", more than just the failing "-dev" strings. Is that accurate?

If you agree with the two statements above, will this revision I've just pushed satisfy your requirements? This runs "podman info" early on, before the test for "-dev". Should the "-dev" test fail, the error message will include the relevant "-dev" strings, and will appear at the bottom of the test, which is the desired objective. And, pass or fail, the full output of "podman info" is available in the same page for anyone to peruse should they so desire.

cevich:
> if the above test fails, there could be further information to be gleaned from the full output of "podman info", more than just the failing "-dev" strings. Is that accurate?

Worse, the *grep* test would ***PASS*** if it happens to not find `-dev` in an error message output.

Looking at changes...

cevich (approved):
This is an acceptable compromise.  Thanks.

LGTM

PR #8956, opened 2021-01-12
Makefile: add target to generate bindings

rhatdan:
Add a `.generate-bindings` make target that only runs in the absence of
the `.generate-bindings` file or when a `types.go` file below
`pkg/bindings` has changed.

It does not execute anything yet is just a template to wire in the
generation in a future change.

### DIFF
--- a/pkg/bindings/containers/types_list_options.go
+++ b/pkg/bindings/containers/types_list_options.go
@@ -12,8 +12,6 @@ import (
 
 /*
 This file is generated automatically by go generate.  Do not edit.
-
-Created 2020-12-18 13:33:20.199081744 -0600 CST m=+0.000270626
 */
 
 // Changed
--- a/pkg/bindings/generate/types_systemd_options.go
+++ b/pkg/bindings/generate/types_systemd_options.go
@@ -4,6 +4,7 @@ import (
 	"net/url"
 	"reflect"
 	"strconv"
+	"strings"
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
@@ -11,8 +12,6 @@ import (
 
 /*
 This file is generated automatically by go generate.  Do not edit.
-
-Created 2020-12-18 15:58:20.661450253 -0600 CST m=+0.000135779
 */
 
 // Changed
@@ -39,6 +38,7 @@ func (o *SystemdOptions) ToParams() (url.Values, error) {
 		if !o.Changed(fieldName) {
 			continue
 		}
+		fieldName = strings.ToLower(fieldName)
 		f := s.Field(i)
 		if reflect.Ptr == f.Kind() {
 			f = f.Elem()
--- a/pkg/bindings/generator/generator.go
+++ b/pkg/bindings/generator/generator.go
@@ -19,13 +19,10 @@ var bodyTmpl = `package {{.PackageName}}
 import (
 {{range $import := .Imports}}	{{$import}}
 {{end}}
-
 )
 
 /*
 This file is generated automatically by go generate.  Do not edit.
-
-Created {{.Date}}
 */
 
 // Changed
@@ -52,6 +49,7 @@ func (o *{{.StructName}}) ToParams() (url.Values, error) {
 		if !o.Changed(fieldName) {
 			continue
 		}
+		fieldName = strings.ToLower(fieldName)
 		f := s.Field(i)
 		if reflect.Ptr == f.Kind() {
 			f = f.Elem()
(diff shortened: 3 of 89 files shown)


rhatdan:
@baude @vrothberg @nalind PTAL

ashley-cui (commented):
I think this creates a dependency on goimports  in $PATH now? If a user doesn't have goimports in their path, the imports will not be correctly done. Looks like the build tests are failing because of this.

 I also would prefer it if we kept the generated time in the generated files.

Also, generating the bindings touches go.mod and go.sum and throws a bunch of inconsistent vendoring errors at me. Not entirely sure what's going on, but that might be my fault.

The files are correctly generated though, so that's good

contrib/cirrus/setup_environment.sh
```diff
@@ -177,6 +177,7 @@ case "$PODBIN_NAME" in
     *) die_unknown PODBIN_NAME
 esac
 
+make install.tools
```
cevich, line 180: Instead of this...try rebasing on https://github.com/containers/podman/pull/8967

The `install.tools` target is run during each of the "Build for XYZ" tasks and the entire $GOPATH is cached for that VM.  The change in my PR simply restores that cache for the `alt_build` task.  I did not do this previously out of fear of polluting something (call it paranoia).

pkg/bindings/generator/generator.go
```diff
@@ -239,7 +236,7 @@ func main() {
 				closed = true
 
 				// go fmt file
-				gofmt := exec.Command("gofmt", "-w", "-s", out.Name())
+				gofmt := exec.Command("go", "fmt", out.Name())
```
baude, line 239: why did you remove this?

pkg/bindings/generate/types_systemd_options.go
```diff
@@ -11,8 +11,6 @@ import (
 
 /*
 This file is generated automatically by go generate.  Do not edit.
-
-Created 2020-12-18 15:58:20.661450253 -0600 CST m=+0.000135779
```
baude, line 15: why did you remove this?

pkg/bindings/network/types_create_options.go
```diff
@@ -193,9 +191,9 @@ func (o *CreateOptions) WithIPRange(value net.IPNet) *CreateOptions {
 
 // GetIPRange
 func (o *CreateOptions) GetIPRange() net.IPNet {
-	var ipRange net.IPNet
+	var iPRange net.IPNet
```
baude, line 194: why?

pkg/bindings/containers/types_list_options.go
```diff
@@ -40,7 +37,6 @@ func (o *ListOptions) ToParams() (url.Values, error) {
 		if !o.Changed(fieldName) {
 			continue
 		}
-		fieldName = strings.ToLower(fieldName)
```
baude, line 43: this is likely your problem. looking into why.

pkg/bindings/network/types_create_options.go
rhatdan: go generate does this not, me.

pkg/bindings/generate/types_systemd_options.go
rhatdan: Because this will cause us to have to update the bindings every time you run create bindings even if the bindings did not change.

pkg/bindings/generator/generator.go
rhatdan: gofmt is not available in path on a lot of ci/cd images, go fmt seems to work, so I switched to it.

```
 go fmt [-n] [-x] [packages]

Fmt runs the command 'gofmt -l -w' on the packages named
by the import paths. It prints the names of the files that are modified.

For more about gofmt, see 'go doc cmd/gofmt'.
For more about specifying packages, see 'go help packages'.

The -n flag prints commands that would be executed.
The -x flag prints commands as they are executed.

The -mod flag's value sets which module download mode
to use: readonly or vendor. See 'go help modules' for more.

To run gofmt with specific options, run gofmt itself.

See also: go fix, go vet.
```

pkg/bindings/containers/types_list_options.go
rhatdan: Yup, this gets removed when you go generate, I tried to hard code this back in, but it caused other tools to break.
Strange that I can not get this same error to happen with podman-remote ps --last 1.

ashley-cui (commented):
LGTM assuming tests pass!

vrothberg (commented):
LGTM, thanks a lot @rhatdan !

umohnani8:
LGTM

PR #8957, opened 2021-01-13
Add 'MemUsageBytes' format option

### LINKED ISSUE #8945: Allow selection of IEC/JEDEC/non-SI units in `podman stats` memory output

/kind feature

**Description**

The output of `podman stats` was updated to use "human-readable" SI units (KB/MB/GB) when displaying output.  This makes sense in the case of storage where SI units are the default (... arguably for marketing reasons in this case, but then GbE is nominally 1,000,000,000bps rather than 1,073,741,824bps).

However, when it comes to memory sizes, this is not particularly helpful - *especially* when the display then doesn't match up with the interpretation of units used by the parameters of the various `--memory*` options.

For example, if I use `--memory 8g` then `podman stats` shows a `MEM USAGE / LIMIT` output of `8.59GB`... which isn't wrong, but doesn't aid the human-readability of the output compared to simply reading `8GiB`!

The best non-breaking way to fix this might be to add additional format options - there's already a `.MemUsage` template placeholder - could `.MemUsageBin`/`.MemUsageIEC`/`.MemUsageJEDEC` be added to select binary measurements?

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Git Commit:   a0d478edea7f775b7ce32f8eb1a01e75374486cb
Built:        Wed Dec  9 11:33:11 2020
OS/Arch:      linux/amd64
```

Yes

srcshelton:
Although storage is more human-readable when expressed in SI units,
IEC/JEDEC (Bytes) units are more pertinent for memory-related values
(and match the format of the --memory* command-line options).

(To prevent possible compatibility issues, the default SI display is
left unchanged)

Closes #8945

Signed-off-by: Stuart Shelton <stuart@shelton.me>


rhatdan:
Thanks @srcshelton 
LGTM
@vrothberg @mheon PTAL

saschagrunert (approved):
LGTM

Issue #8921, opened 2021-01-10
/etc/hosts duplicated every time after container restarted


changsijay:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`/etc/hosts` duplicated every time when container restarted

**Steps to reproduce the issue:**

```sh
sudo podman pod create --name demo-pod
sudo podman run --name=demo-container --detach --restart=always --pod=demo-pod nginx
sudo podman exec -it demo-container cat /etc/hosts | grep demo-container | wc
# get 1

sudo podman restart demo-container
sudo podman exec -it demo-container cat /etc/hosts | grep demo-container | wc
# get 2

sudo podman restart demo-container
sudo podman exec -it demo-container cat /etc/hosts | grep demo-container | wc
# get 3
```

**Describe the results you received:**

`/etc/hosts` keep duplicating `127.0.1.1 demo-pod demo-container` after I restart `demo-container`

**Describe the results you expected:**

`127.0.1.1 demo-pod demo-container` should not duplicated in `/etc/hosts`

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.6
Git Commit:   a0d478edea7f775b7ce32f8eb1a01e75374486cb
Built:        Wed Dec  9 04:48:23 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /usr/bin/conmon
    version: 'conmon version 2.0.22, commit: 9c34a8663b85e479e0c083801e89a2b2835228ed'
  cpus: 4
  distribution:
    distribution: arch
    version: unknown
  eventLogger: journald
  hostname: Macmini
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
  kernel: 5.10.5-arch1-1
  linkmode: dynamic
  memFree: 10837921792
  memTotal: 16686837760
  ociRuntime:
    name: runc
    package: Unknown
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc92
      commit: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
[...]

PR #8958, opened 2021-01-13
Fixes /etc/hosts duplicated every time after container restarted in a pod

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/8921
```
$ podman pod create --name p1
$ podman run -d --pod p1 --name c1 $IMG top
$ podman exec c1 sh -c "wc -l < /etc/hosts"
5
$ podman restart c1
$ podman exec c1 sh -c "wc -l < /etc/hosts"
6 <--- should be 5
```
after fix
```
$ ./bin/podman restart c1
$ podman exec c1 sh -c "wc -l < /etc/hosts"
6 <-- don't  change to 7
```

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1,6 +1,7 @@
 package libpod
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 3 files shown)


libpod/container_internal.go
```diff
@@ -1866,15 +1867,35 @@ func (c *Container) writeStringToStaticDir(filename, contents string) (string, e
 }
 
 // appendStringToRundir appends the provided string to the runtimedir file
-func (c *Container) appendStringToRundir(destFile, output string) (string, error) {
+func (c *Container) appendStringToRundir(destFile, output string, checkExist bool) (string, error) {
```
rhatdan, line 1870: Not sure we need the checkExists parameter, why not do this by default?  Do you envision this function being used in the future where something wants to just keep appending on the end of a file?

zhangguanzhang: I will change it to default

rhatdan:
LGTM

PR #8959, opened 2021-01-13
Use errors.Is for comparing ErrOCIRuntime errors

JAORMX:
Use errors.Is for comparing ErrOCIRuntime errors

This simplifies checking for ErrOCIRuntime type of errors by using go's
built-in `errors.Is` function.

In order to make this practical, some errors now wrap `ErrOCIRuntime`.
Namely: `ErrOCIRuntimePermissionDenied` and `ErrOCIRuntimeNotFound`.

Given that some errors might now wrap `ErrOCIRuntime`, the error
formatting in the podman command's output was adjusted to take this into
account.

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	goerrors "errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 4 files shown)


rhatdan:
I am a little concerned about doing this piecemail.  IE Should we switch from using errors.Cause-> errors.Is and all other uses of the "errors" modules.

This is going to be a far bigger change and needs to be discussed, I believe.  Definitely should not be done until post 3.0 release.

JAORMX:
ah... seems that I gotta fix some tests before this works.

> I am a little concerned about doing this piecemail. IE Should we switch from using errors.Cause-> errors.Is and all other uses of the "errors" modules.
> 
> This is going to be a far bigger change and needs to be discussed, I believe. Definitely should not be done until post 3.0 release.

Do you think this change is too disruptive? I can abandon this if it would be too much of a hassle for the code-base. I was just trying to do some clean-up. But if the cleanup dirties the code-base more than it really cleans up, then it's no use. If folks think it's worth to start moving away from the custom errors module and into golang's error functions, that's fine and I could try to help. It's fine if this waits post 3.0, it's not urgent.

rhatdan:
No I think it is time we started looking at this, but I want limited risk since we are hoping to release end of the week.

JAORMX:
> No I think it is time we started looking at this, but I want limited risk since we are hoping to release end of the week.

sounds good. Let's then continue this work until after that.

cmd/podman/root.go
```diff
@@ -1,6 +1,7 @@
 package main
 
 import (
+	goerrors "errors"
```
mheon, line 4: As this point, I don't think we can swap over to Golang's `errors` package without doing it everywhere throughout the stack - there is a high chance the error we're working with came from `github.com/pkg/errors` and was wrapped. Recommend you just use `errors.Cause()` instead.

PR #8961, opened 2021-01-13
play kube: set entrypoint when interpreting Command

rhatdan:
We now set Entrypoint when interpeting the image Entrypoint (or yaml.Command)
and Command when interpreting image Cmd (or yaml.Args)

This change is kind of breaking because now checking Config.Cmd won't return
the full command, but only the {cmd,args}.

Adapt the tests to this change as well


rhatdan:
Replaces: https://github.com/containers/podman/pull/8808

rhatdan:
@haircommander PTAL

haircommander:
LGTM

haircommander:
thanks for carrying this

PR #8973, opened 2021-01-14
[CI:DOCS] Cirrus: Upd. ext. service check host list

cevich:
Since CI doesn't depend heavily on installing packages at runtime
(there is some minor use) there's no need to exhaustively check
repository mirror hosts.  Remove them from the list.


mheon:
LGTM

edsantiago:
LGTM. Thanks, this has been a thorny issue for a long time.

Issue #7766, opened 2020-09-24
Fedora gating test failures


lsm5:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Fedora gating tests are currently failing.

**Steps to reproduce the issue:**

1. /usr/bin/podman run --uidmap 0:10000:10000 quay.io/libpod/testimage:20200902 true

2. Complete log at https://paste.centos.org/view/e6f829d2

**Describe the results you received:**

```
$ /usr/bin/podman run --uidmap 0:10000:10000 quay.io/libpod/testimage:20200902 true
Error: write to /proc/sys/net/ipv4/ping_group_range: Invalid argument: OCI runtime error
```

**Describe the results you expected:**

No error (I guess)

**Additional information you deem important (e.g. issue happens only occasionally):**

Consistent

**Output of `podman version`:**

```
Version:      2.1.0
API Version:  2.0.0
Go Version:   go1.15
Built:        Tue Sep 22 13:56:54 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 0f53fb68333bdead5fe4dc5175703e22cf9882ab'
  cpus: 8
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: hashirama.nanadai.me
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.10-300.fc33.x86_64
  linkmode: dynamic
  memFree: 20476694528
  memTotal: 33087016960
  ociRuntime:
    name: crun
    package: crun-0.14.1-5.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.14.1
      commit: 598ea5e192ca12d4f6378217d3ab1415efeddefa
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
[...]

PR #8975, opened 2021-01-14
system tests: disable ping tests on RHEL

edsantiago:
RHEL sets the net.ipv4.ping_group_range sysctl to "1 0",
basically preventing ping from working without setuid.
Disable the ping system test if we detect we're on RHEL.

### DIFF
--- a/test/system/200-pod.bats
+++ b/test/system/200-pod.bats
@@ -287,8 +287,16 @@ EOF
        "two containers cannot bind to same port"
 
     # make sure we can ping; failure here might mean that capabilities are wrong
-    run_podman run --rm --pod mypod $IMAGE ping -c1 127.0.0.1
-    run_podman run --rm --pod mypod $IMAGE ping -c1 $hostname
+    # NOTE: this does not work on RHEL, neither root nor rootless, because
+    #   the net.ipv4.ping_group_range sysctl defaults to "1 0". This allows
+    #   users to use /bin/ping, which is setuid, but prevents containers.
+    #   So, disable this test entirely on RHEL. (It's probably cleaner to
+    #   check sysctl here instead of hardcoding a RHEL exception, but this
+    #   is quick & safe enough.)
+    if ! egrep -q 'cpe:.*redhat.*enterprise' /etc/os-release; then
+        run_podman run --rm --pod mypod $IMAGE ping -c1 127.0.0.1
+        run_podman run --rm --pod mypod $IMAGE ping -c1 $hostname
+    fi
 
     # While the container is still running, run 'podman ps' (no --format)
     # and confirm that the output includes the published port


rhatdan:
LGTM
We should open an Issue with RHEL to change this default though.

edsantiago:
If someone wants to do that, I think the starting point might be [rhbz1740809](https://bugzilla.redhat.com/show_bug.cgi?id=1740809) (closed, f31). It references this very nice [requirements document](https://fedoraproject.org/wiki/Changes/EnableSysctlPingGroupRange). Unfortunately the BZ does not link to a kernel commit or any form of detail on how this was made to happen. It doesn't seem to be as simple as setting `net.ipv4.ping_group_range="0 0"` because doing so on a RHEL8.4 system makes no difference: rootless ping still doesn't work:

    $ podman run quay.io/libpod/testimage:20200929 ping -c 3 127.0.0.1
    PING 127.0.0.1 (127.0.0.1): 56 data bytes
    bind: Permission denied

    $ podman run -it ubi8 sh
    sh-4.4# dnf -y install iputils
    ...
    sh-4.4# ping -c 3 10.0.138.150
    PING 10.0.138.150 (10.0.138.150) 56(84) bytes of data.

    --- 10.0.138.150 ping statistics ---
    3 packets transmitted, 0 received, 100% packet loss, time 69ms

(Curiously -- but not curiously enough for me to spend any time on it -- the latter actually works with 127.0.0.1. Just, not with an address outside the lo interface).  Setting the sysctl to "0 2147483647" allows rootless ping to work.

Other reference points:
* [rhbz1882136](https://bugzilla.redhat.com/show_bug.cgi?id=1882136) - toolbox-related errors writing the sysctl
* #7766 - Related discussion on Fedora gating-test failures

vrothberg (commented):
LGTM

vrothberg:
> /hold

Cheers! :)

edsantiago:
Resubmitting: I tested this as *root* on RHEL, and it failed. Looks like ping doesn't work on containers in RHEL, even root! (Glad I tested). New diff is cleaner because there's no need for nested conditionals.

TomSweeneyRedHat:
@edsantiago Does RHEL not work at all, or only later versions?

edsantiago:
I have no idea. The `ping` test is new (#8535, November 2020) and it looks like it has never been tested on RHEL until now. Per IRC conversation with @rhatdan earlier today, he believes that with the RHEL settings of `net.ipv4.ping_group_range`, there's no way ping would work in a container. 

Allow me to mention, once again, how dangerous it is for us to run zero CI tests on RHEL.

edsantiago:
Oh boy... I'm sensing a great disturbance in the CI force coming soon: [build-each-commit test failed](https://cirrus-ci.com/task/5506549349089280) without my having done anything other than a new push. I've tried rebasing on master (post-#8956) and am crossing my fingers...

rhatdan:
We need to get Root to fix their broken sysctl.

rhatdan:
I am now thinking we might need to add back net_raw to RHEL.

edsantiago:
FWIW, that fixes ping on RHEL, both root and rootless:
```console
$ podman run --cap-add cap_net_raw quay.io/libpod/testimage:20200929 ping -c 3 127.0.0.1
PING 127.0.0.1 (127.0.0.1): 56 data bytes
64 bytes from 127.0.0.1: seq=0 ttl=64 time=0.055 ms
64 bytes from 127.0.0.1: seq=1 ttl=64 time=0.088 ms
64 bytes from 127.0.0.1: seq=2 ttl=64 time=0.062 ms

--- 127.0.0.1 ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
```
(Still can't ping the host IP on eth0, rootless, but that doesn't work on f33 either, and ISTR it will never work anyway due to rootless networking. Works fine as root).

rhatdan:
@edsantiago do you have the sysctls set?

edsantiago:
```
net.ipv4.ping_group_range = 1   0
```

rhatdan:
And this works in rootless mode?  How about if you do 
--net=host?

edsantiago:
No, `--net=host` breaks it again:
```console
$ podman run --net=host --cap-add cap_net_raw quay.io/libpod/testimage:20200929 ping -c 3 127.0.0.1
PING 127.0.0.1 (127.0.0.1): 56 data bytes
ping: permission denied (are you root?)
```
With ubi8:
```console
sh-4.4# ping -c 3 10.0.138.150
ping: socket: Operation not permitted
```

rhatdan:
Ok that makes some sense.  Although it could be revealing a bug in the kernel.
Bottom line is you can not send a ping on the host network from within a rootless container, without the "BIG" sysctl set.  Which makes sense from a security point of view.  You need a trusted setuid program with NET_ADMIN caps to be able to send ping packets.

When in a usernamespace with a controlled network, you can send ping packets with Namespaced NET_ADMIN

rhatdan:
Bottom line we have to talk to Product Management about what they want the defaults to be.

edsantiago:
@rhatdan I get nervous with "we need to" statements. Do you have someone in mind to do this? How do we guarantee that this gets properly assigned and resolved? Should I create an issue? a BZ?

rhatdan:
Sent an email to begin discussion.

edsantiago:
AIUI the cause of this problem is being addressed by re-adding CAP_NET_RAW to podman on RHEL8. Can we close this PR?

Issue #8734, opened 2020-12-15
Error: pod stats is not supported in rootless mode without cgroups v2


cevich:
/kind bug

**Description**

Nearly/all `podman pod stats` tests fail when running as a user, on a host using CGroupsV1 & `runc-1.0.0-146.rc92`.  The tests were globally skipped in the case of rootless + CGroupsV2.  However, apparently they also cannot function with CGv1 either.

As of #8312 this has been addressed by a blanket `SkipIfRootless()`. I am not certain if this is otherwise fixable or worth the effort, since `--cgroup-manager=cgroupfs` is required for rootless operation.

**Steps to reproduce the issue:**

1. Execute integration tests as a regular user

**Describe the results you received:**

[(typical failure example)](https://cirrus-ci.com/task/6062647370055680)

```
[+0284s] Podman pod stats 
[+0284s]   podman stats should run with no pods
[+0284s]   /var/tmp/go/src/github.com/containers/podman/test/e2e/pod_stats_test.go:42
[+0284s] 
[+0284s] [BeforeEach] Podman pod stats
[+0284s]   /var/tmp/go/src/github.com/containers/podman/test/e2e/pod_stats_test.go:19
[+0284s] [It] podman stats should run with no pods
[+0284s]   /var/tmp/go/src/github.com/containers/podman/test/e2e/pod_stats_test.go:42
[+0284s] Running: podman --storage-opt vfs.imagestore=/tmp/podman/imagecachedir --root /tmp/.../crio --runroot /tmp/.../crio-run --runtime runc --conmon /usr/bin/conmon --cni-config-dir /home/some20681dude/.config/cni/net.d --cgroup-manager cgroupfs --tmpdir /tmp/... --events-backend file --storage-driver vfs pod stats --no-stream
[+0284s] Error: pod stats is not supported in rootless mode without cgroups v2
```

**Describe the results you expected:**

Someone who knows more than me should examine `BeforeEach()` in `test/e2e/pod_stats_test.go` to ensure the skip-selection is correct.  Esp. after #8312 is merged.

mheon:
The logic here sounds inverted. The tests should run on Cgroups v2 + rootless. They are not expected to work on cgroups v1 + rootless.

cevich:
Sorry, I don't follow how that's any different from the issue title?

cevich:
Also...aren't you suppose to be on PTO? 😛

PR #8983, opened 2021-01-15
Turn on podman pod stats test for rootless cgroup v2

rhatdan:
These tests should only be blocked for cgroups V1

Fixes: https://github.com/containers/podman/issues/8734


rhatdan:
@edsantiago @cevich PTAL

test/e2e/pod_stats_test.go
```diff
@@ -177,7 +177,7 @@ var _ = Describe("Podman pod stats", func() {
 
 	It("podman stats on net=host post", func() {
 		// --net=host not supported for rootless pods at present
-		SkipIfRootlessCgroupsV1("Pause stats not supported in cgroups v1")
+		SkipIfRootless("Pause stats not supported in cgroups v1")
```
edsantiago, line 180: Shouldn't this `Skip` be removed entirely?

rhatdan: No, because, It will not work rootless, I believe. It fails locally.  Before the change above it was never being executed in cgroupV2 mode.

The issue is attempting to join the network namespace is causing a permission denied.

edsantiago: Oh - got it, I misread the change in `BeforeEach()`. Sorry.

Can you change the string description here, replacing "cgroups v1" with "rootless"?

rhatdan: Just grabbed the comment line from above.

edsantiago:
Thank you. LGTM, pending successful CI.

Issue #8842, opened 2020-12-27
Feature: --replace or else for "podman network create" and "podman volume create"


daniel-wtd:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Creating new volumes or networks can be done via `podman volume create <name>` or `podman network create <name>`. This will fail, if `<name>` already exists. For volumes, one can workaround this issue, since `podman volume run -v <name>:/foo` will only create the volume, if it is not existing. For networks, this does not work.

It would be awesome to see some cli additions, so `podman network create --replace <name>` or `podman network create --if-not.-exists <name>` can be used.

**Output of `podman version`:**

```
podman version 2.2.1
```

rhatdan:
Not sure what your goal here is?  Are you looking to have a script that runs multiple times but does not know if the network was already created?

Would `podman network create foobar` failing with an already EEXIST (17) satisfy your needs?

$ errno  -l | grep -i exis
EEXIST 17 File exists

daniel-wtd:
The goal is quite easy. podman containers will be started with systemd. For me it is only natural to do the network and volume creation the same way, alongside the containers. This is currently not achievable, since `ExecStart=` will fail, if the network is already existing.

rhatdan:
Couldn't you do something like 

ExecStartPre=podman network inspect FOOBAR > /dev/null || podman network create FOOBAR

daniel-wtd:
Sure, I can do this or provide a script or else.

It just feels like a missing feature/a workaround, since:

```
# This works and replaces existing pods with the same name
podman pod create --replace web

# This works and replaces existing containers with the same name
podman container run --name web01 --replace httpd:2

# This will fail, if net01 already exists
podman network create net01

# This will fail, if vol01 already exists
podman volume create vol01

# This will fail, if net01 does not exist
podman container run --name web01 --replace --network net01 httpd:2

# This works and creates vol01 on demand/if not existing
podman container create --volume vol01:/var/www/html httpd:2
```

Luap99:
Note that `replace` is not the same as `if not exists`. Replace will always remove before creating the container/pod. If not exists will do nothing if it already exists. Deleting a network is not safe if other containers are using it so I think you want the if not exists behaviour.

daniel-wtd:
@Luap99 completely correct. This is how `--volume` currently works and I would appreciate to see the same for `--network`.

[...]

PR #8987, opened 2021-01-16
Add --if-not-exists option to podman network create

Luap99:
If `--if-not-exists` is set `podman network create` will not error
if the a network with the same given name already exists.

This option is only supported for local podman client. Supporting
the remote client would require changes in the api endpoint.

This option can be used as `ExecStartPre` command in a systemd
service file. see #8842

### DIFF
--- a/cmd/podman/networks/create.go
+++ b/cmd/podman/networks/create.go
@@ -48,6 +48,9 @@ func networkCreateFlags(cmd *cobra.Command) {
 	flags.IPVar(&networkCreateOptions.Gateway, gatewayFlagName, nil, "IPv4 or IPv6 gateway for the subnet")
 	_ = cmd.RegisterFlagCompletionFunc(gatewayFlagName, completion.AutocompleteNone)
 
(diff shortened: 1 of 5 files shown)


rhatdan:
Bikeshedding a little, but I hate options with "-" embeded in them.

--ignore?
--exists?

rhatdan:
Currently we have commands

podman image exists
podman pod exists
podman container exists

We should add 
podman network exists
podman volume exists
podman manifest exists

Would that be better, then you could make a command like
podman network exists || podman network create ...

vrothberg:
Container/Pod creation also has a `--replace` flag which would remove existing objects in case of name conflicts.

Luap99:
@vrothberg Replace behaviour is not wanted. see #8842

Luap99:
@rhatdan I agree `podman network exists` would make sense.

vrothberg:
+1

Luap99:
podman network exists PR #9021

PR #8990, opened 2021-01-17
list the volumes to be deleted without forcing

### LINKED ISSUE #8913: RFE: Let `podman volume prune` show the volumes that are going to be removed
/kind feature

**Description**

Running `podman volume prune` you have to confirm the operation but you don't see which volumes actually will be removed after you confirm the command:

```
$ podman volume prune
WARNING! This will remove all volumes not used by at least one container.
Are you sure you want to continue? [y/N]
```

I was trapped by the issue #7862. This could have been avoided by showing the volumes that are going to be removed before you confirm the removal. Because the user would have the chance that there is a volume going to be removed that's currently in use by a running container instance. Possible data loss could be prevented this way.

Or at least add an option to the command that will show the volumes that are going to be removed.

1. `podman volume prune`

2. Hit 'y'

3.

```
$ podman volume prune
WARNING! This will remove all volumes not used by at least one container.
Are you sure you want to continue? [y/N]
```

```
$ podman volume prune
WARNING! This will remove all volumes not used by at least one container. The following volumes will be removed:
volume1
volume2
...
Are you sure you want to continue? [y/N]
```

```
podman version 3.0.0-dev
```

Yes

iwita:
This commit lists the volumes that are going to be removed when the command `podman volume prune` is issued. 
It is applied only in the cases when `--force` flag is **not** enabled.
Closes #8913


cmd/podman/volumes/prune.go
```diff
@@ -58,7 +59,18 @@ func prune(cmd *cobra.Command, args []string) error {
 	}
 	if !force {
 		reader := bufio.NewReader(os.Stdin)
-		fmt.Println("WARNING! This will remove all volumes not used by at least one container.")
+		fmt.Println("WARNING! This will remove all volumes not used by at least one container. The following volumes will be removed:")
+		listOptions.Filter, err = filters.ParseFilterArgumentsIntoFilters(filter)
```
rhatdan, line 63: This would list all volumes, not just the ones that would be removed.  podman volume prune is going to prune the volumes that are not in use. 

I believe you would need to have a filter to show volumes that are not in use.

iwita: Hi @rhatdan.
Yes, you are right. Didn't consider the actual use of the `prune` command. 
Now I added a function in `cmd/podman/volumes/prune.go` in order to get the used volumes and remove them from the available ones for pruning.

rhatdan:
I would prefer to add a filter command, Put this on the server side.

```
podman volume ls --filter unused
```
For example.  Then use this to fix list volumes

iwita:
> I would prefer to add a filter command, Put this on the server side.
> 
> ```
> podman volume ls --filter unused
> ```
> 
> For example. Then use this to fix list volumes

Thank you very much for your guidance.
I applied the needed changes by adding another filter in `pkg/domain/filters/volumes.go`
However, when I tried to use this newly created filter for the purposes of this PR, I found out that the `Volumes` function in https://github.com/containers/podman/blob/master/libpod/runtime_volume.go#L113 uses an OR statement when validating a Volume. Thus, when the custom **unused=1** filter is applied(internally) together with user-specified filters (e.g. volume labels), if either of them is true, the volume is printed (which is unwanted).

lsm5:
@iwita all commits need to have DCO/commit signoff else Smoke Test will complain.

PR #8995, opened 2021-01-17
[CI:DOCS] lint lint lint

vrothberg:
* golangci-lint: install to ./bin
* hack/install_golangci.sh: smarter install

Two commits to make the installation of golangci-lint smarter.  It'll now be installed into `./bin` to prevent sharing `$GOBIN` with other projects who may require a different version.

Since `varlink` has been removed, we can finally update golangci-lint to more recent versions.  I'll plan that soon unless somebody else is faster.


vrothberg:
@cevich, I see the compose test and the APIv2 test running despite `[CI:docs]`.  Is that known / desired?

vrothberg:

baude:
LGTM

PR #9004, opened 2021-01-18
Add binding options for container|pod exists

baude:
It turns out an options was added to container exists so it makes sense
to have pods and container exists calls have an optional structure for
options.

### DIFF
--- a/pkg/bindings/containers/types.go
+++ b/pkg/bindings/containers/types.go
@@ -237,3 +237,10 @@ type ExecStartAndAttachOptions struct {
 	// If false, stdout will not be attached
 	AttachInput *bool
 }
+
+//go:generate go run ../generator/generator.go ExistsOptions
+// ExistsOptions are optional options for checking if a container exists
+type ExistsOptions struct {
+	// External checks for containers created outside of Podman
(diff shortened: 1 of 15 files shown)


Luap99 (commented):
LGTM

vrothberg (commented):
LGTM

pkg/bindings/containers/types.go
```diff
@@ -237,3 +237,10 @@ type ExecStartAndAttachOptions struct {
 	// If false, stdout will not be attached
 	AttachInput *bool
 }
+
+//go:generate go run ../generator/generator.go ExistsOptions
+// ExistsOptions are optional options for checking if a container exists
+type ExistsOptions struct {
+	// External checks for containers created outside of Podman
+	External bool
```
vrothberg, line 245: This should be a bool

baude: fixed

Luap99:
@baude I just realized the image exists binding has to be updated as well.

baude:
> @baude I just realized the image exists binding has to be updated as well.

fixed up!

PR #9008, opened 2021-01-18
runner.sh : deal with bash 'set -e'

edsantiago:
Release trigger script failed[1] because the entire script
runs under 'set -e'; so a 'grep -- -dev' that finds no
results will cause a nonzero exit status and hence the
entire script to fail. Work around that.

 [1] https://cirrus-ci.com/task/4541290882793472

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -210,7 +210,8 @@ function _run_release() {
     bin/podman info
 
     msg "Checking podman release (or potential release) criteria."
-    dev=$(bin/podman info |& grep -- -dev)
+    # We're running under 'set -eo pipefail'; make sure this statement passes
+    dev=$(bin/podman info |& grep -- -dev || echo -n '')
     if [[ -n "$dev" ]]; then
         die "Releases must never contain '-dev' in output of 'podman info' ($dev)"
     fi


edsantiago:
@cevich PTAL, this should address the failure you linked in scrum today

cevich (approved):
LGTM.  Please port this over to the `v3.0` branch in a separate PR.  On that PR, please manually trigger the test in the PR to confirm it works.

cevich:
Ref: [Passing test on the `v3.0` branch](https://github.com/containers/podman/runs/1724313840)

cevich:
@edsantiago now that you've fully experienced the pain of testing this code in the intended CI context...perhaps you understand why I'm not happy having these checks in this repo.

What do you think about my idea of moving them into a script in `c/automation`, so the task in this repo. would simply pull the external code down and run it?

Advantages:
* We can leverage CI in `c/automation` to exercise the release code vs a mock-release of podman
* During podman release testing (i.e. after a tag-push), if something goes south, we can address it by modifying code in `c/automation` to fix it. Thus avoiding the need to push a new podman tag only to fix the release-test code.

edsantiago:
@cevich that seems like a high-effort solution to a low-risk problem, but I could definitely see it as a good idea if someone decides to add new `release` tests. That would be the ideal time to make this change, IMO, but if you feel like tackling it sooner I will be glad to review your changes.

cevich:
> if someone decides to add new release tests

@edsantiago yes, that was my thinking as well...and why it's just a "TODO" comment.  Though I'm also thinking bigger, in having more automation involved in (or running) the release process.  My Jira card is about exploring these ideas more fully with the group, to see what level of automation makes sense.

PR #9013, opened 2021-01-18
[v3.0] runner.sh : deal with bash 'set -e'

edsantiago:
Release trigger script failed[1] because the entire script
runs under 'set -e'; so a 'grep -- -dev' that finds no
results will cause a nonzero exit status and hence the
entire script to fail. Work around that.

 [1] https://cirrus-ci.com/task/4541290882793472

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -210,7 +210,8 @@ function _run_release() {
     bin/podman info
 
     msg "Checking podman release (or potential release) criteria."
-    dev=$(bin/podman info |& grep -- -dev)
+    # We're running under 'set -eo pipefail'; make sure this statement passes
+    dev=$(bin/podman info |& grep -- -dev || echo -n '')
     if [[ -n "$dev" ]]; then
         die "Releases must never contain '-dev' in output of 'podman info' ($dev)"
     fi


contrib/cirrus/runner.sh
```diff
@@ -210,7 +210,8 @@ function _run_release() {
     bin/podman info
 
     msg "Checking podman release (or potential release) criteria."
-    dev=$(bin/podman info |& grep -- -dev)
+    # We're running under 'set -eo pipefail'; make sure this statement passes
+    dev=$(bin/podman info |& grep -- -dev || echo -n '')
```
zhangguanzhang, line 214:
```suggestion
    dev=$(bin/podman info |& grep -- -dev ||  : )
```

zhangguanzhang: maybe this will be better

edsantiago: @zhangguanzhang thank you, why do you prefer that form? I chose `echo -n ''` because the expression is intended as a string capture, and I wanted to make it easily clear to a maintainer that the fallthrough is an empty string. `true` is logically equivalent, but could make a future maintainer stop to think it through. (At least, that was my thinking earlier today, and I still think it holds). What do you like about the `true` form?

zhangguanzhang: agree with you

Issue #8735, opened 2020-12-15
podman rm external-container: sometimes needs -f


edsantiago:
There seem to be two types of buildah containers: completed-successfully ones, and incomplete ones. `podman ps -a --storage` can see both, but can only `rm` the latter if `-f` is given.

Complete containers:
```console
$ buildah from quay.io/libpod/testimage:20200929
testimage-working-container
$ ./bin/podman rm testimage-working-container
testimage-working-container
$ echo $?
0     <----- yay it works
```

Incomplete containers, created by interrupting a build:
```console
$ printf "FROM quay.io/libpod/testimage:20200929\nRUN sleep 30\n"| timeout -v 5 podman build -t thiswillneverexist -
STEP 1: FROM quay.io/libpod/testimage:20200929
STEP 2: RUN sleep 30
timeout: sending signal TERM to command ‘podman’
$ echo $?
124        <--- expected

$ ./bin/podman rm testimage-working-container
Error: no container with name or ID testimage-working-container found: no such container
$ ./bin/podman ps -a --storage
CONTAINER ID  IMAGE                              COMMAND  CREATED         STATUS   PORTS   NAMES
57a7769d2ca2  quay.io/libpod/testimage:20200929  buildah  56 seconds ago  storage          testimage-working-container
$ ./bin/podman rm 57a7769d2ca2
Error: no container with name or ID 57a7769d2ca2 found: no such container
$ echo $?
1

$ ./bin/podman rm -f 57a7769d2ca2        <--- add "-f", it works
57a7769d2ca2
$ echo $?
0
```
master @ 9379ee97907266e5a96bd9c4ae178e0e572a6f2a

Terminology: "complete" and "incomplete" is my simpleminded attempt to find a way to describe this situation. My nomenclature may be incorrect or even misleading. Please go by the reproducer, not my words.

mheon:
Guess: the ones that won't `rm` without `-f` are mounted - I'd suspect a check of `ps` would show some extra fuse-overlay processes corresponding to them.

I think that not removing mounted containers without `-f` is probably correct - it doesn't catch all partially-failed containers from a `podman build` but it does make it a bit more difficult to accidentally remove a CRI-O container someone is actually using.

edsantiago:
Good guess:
```
[...]

PR #9014, opened 2021-01-18
Fix handling of container remove

rhatdan:
I found several problems with container remove

podman-remote rm --all
Was not handled

podman-remote rm --ignore
Was not handled

Return better errors when attempting to remove an --external container.
Currently we return the container does not exists, as opposed to container
is an external container that is being used.

This patch also consolidates the tunnel code to use the same code for
removing the container, as the local API, removing duplication of code
and potential problems.

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -14,7 +14,9 @@ import (
 	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/api/handlers"
 	"github.com/containers/podman/v2/pkg/api/handlers/utils"
+	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/domain/filters"
+	"github.com/containers/podman/v2/pkg/domain/infra/abi"
 	"github.com/containers/podman/v2/pkg/ps"
 	"github.com/containers/podman/v2/pkg/signal"
 	"github.com/docker/docker/api/types"
@@ -29,9 +31,11 @@ import (
 func RemoveContainer(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
-		Force bool `schema:"force"`
-		Vols  bool `schema:"v"`
-		Link  bool `schema:"link"`
+		All     bool `schema:"all"`
+		Force   bool `schema:"force"`
+		Ignore  bool `schema:"ignore"`
+		Link    bool `schema:"link"`
+		Volumes bool `schema:"v"`
 	}{
 		// override any golang type defaults
 	}
@@ -49,34 +53,31 @@ func RemoveContainer(w http.ResponseWriter, r *http.Request) {
 	}
 
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	// Now use the ABI implementation to prevent us from having duplicate
+	// code.
+	containerEngine := abi.ContainerEngine{Libpod: runtime}
 	name := utils.GetName(r)
-	con, err := runtime.LookupContainer(name)
-	if err != nil && errors.Cause(err) == define.ErrNoSuchCtr {
-		// Failed to get container. If force is specified, get the container's ID
-		// and evict it
-		if !query.Force {
+	options := entities.RmOptions{
+		All:     query.All,
+		Force:   query.Force,
+		Volumes: query.Volumes,
+		Ignore:  query.Ignore,
+	}
+	report, err := containerEngine.ContainerRm(r.Context(), []string{name}, options)
+	if err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
 			utils.ContainerNotFound(w, name, err)
 			return
 		}
 
-		if _, err := runtime.EvictContainer(r.Context(), name, query.Vols); err != nil {
-			if errors.Cause(err) == define.ErrNoSuchCtr {
-				logrus.Debugf("Ignoring error (--allow-missing): %q", err)
-				w.WriteHeader(http.StatusNoContent)
-				return
-			}
-			logrus.Warn(errors.Wrapf(err, "failed to evict container: %q", name))
-			utils.InternalServerError(w, err)
-			return
-		}
-		w.WriteHeader(http.StatusNoContent)
-		return
-	}
-
-	if err := runtime.RemoveContainer(r.Context(), con, query.Force, query.Vols); err != nil {
 		utils.InternalServerError(w, err)
 		return
 	}
+	if report[0].Err != nil {
+		utils.InternalServerError(w, report[0].Err)
+		return
+	}
+
 	utils.WriteResponse(w, http.StatusNoContent, nil)
 }
 
--- a/pkg/bindings/containers/types.go
+++ b/pkg/bindings/containers/types.go
@@ -122,6 +122,8 @@ type PruneOptions struct {
 //go:generate go run ../generator/generator.go RemoveOptions
 // RemoveOptions are optional options for removing containers
 type RemoveOptions struct {
+	All     *bool
+	Ignore  *bool
 	Force   *bool
 	Volumes *bool
 }
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -264,30 +264,30 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
 	reports := []*entities.RmReport{}
 
 	names := namesOrIds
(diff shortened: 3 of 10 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/8735

rhatdan:
@edsantiago PTAL

pkg/bindings/containers/types.go
```diff
@@ -122,6 +122,8 @@ type PruneOptions struct {
 //go:generate go run ../generator/generator.go RemoveOptions
 // RemoveOptions are optional options for removing containers
 type RemoveOptions struct {
+	All     *bool
+	Ignore  *bool
 	Force   *bool
```
edsantiago, line 127: I'm going to assume that Tom will ask you to sort those

edsantiago:
>podman-remote rm --all  Was not handled

That's quite a hole, I'm trying to figure out how the system tests didn't detect it. Thanks for catching it.

edsantiago:
Friendly suggestion for system tests:
```diff
diff --git a/test/system/040-ps.bats b/test/system/040-ps.bats
index 0447122b1..0ae8b0ce0 100644
--- a/test/system/040-ps.bats
+++ b/test/system/040-ps.bats
@@ -111,8 +111,11 @@ EOF
     run_podman ps --storage -a
     is "${#lines[@]}" "2" "podman ps -a --storage sees buildah container"

-    # This is what deletes the container
-    # FIXME: why doesn't "podman rm --storage $cid" do anything?
+    # We can't rm it without -f, but podman should issue a helpful message
+    run_podman 2 rm "$cid"
+    is "$output" "Error: container .* is mounted and cannot be removed without using force: container state improper" "podman rm <buildah container> without -f"
+
+    # With -f, we can remove it.
     run_podman rm -f "$cid"

     run_podman ps --storage -a
```
Is the exit status (2) intentional? I was surprised that it wasn't 125.

edsantiago:
OBTW CI is failing

pkg/api/handlers/compat/containers.go
```diff
@@ -29,9 +31,11 @@ import (
 func RemoveContainer(w http.ResponseWriter, r *http.Request) {
```
mheon, line 31: This is the compat endpoint. We shouldn't be modifying its parameters.

pkg/domain/infra/abi/containers.go
```diff
@@ -264,25 +264,27 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
...
+		report.Err = ic.Libpod.RemoveStorageContainer(ctr, options.Force)
+		switch errors.Cause(report.Err) {
+		case nil:
 			// remove container names that we successfully deleted
-			tmpNames = append(tmpNames, ctr)
-		} else {
 			reports = append(reports, &report)
+		case define.ErrNoSuchCtr:
+			// remove container names that does not exist
+			reports = append(reports, &report)
```
mheon, line 279: This seems dubious - `podman rm doesnotexist` needs to throw an error, and I'm not sure it does after this.

mheon: Also, I think this is actually unsafe - `RemoveStorageContainer` can potentially return `ErrNoSuchCtr` in cases where the container does not exist in c/storage *but still exists in Libpod*

pkg/domain/infra/abi/containers.go
```diff
@@ -264,25 +264,27 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
...
+		default:
+			if _, err := ic.Libpod.LookupContainer(ctr); errors.Cause(err) == define.ErrNoSuchCtr {
```
mheon, line 281: This bit seems unnecessary - I don't see any way we fail in a fashion that's not ErrNoSuchCtr before `RemoveStorageContainer` checks Libpod to see if the container exists.

rhatdan: The use case we are trying to fix #8735, is the case where we created a buildah container that is in use, and fails.  This would end up in this block, and now we check to see if it is a libpod container.  Since it is not, we return the error. 
Thus we end up with an error telling us the container is in use, and the user has to add --force.
In the current code we fail over to trying to remove a libpod container, and we report
container does not exist.

pkg/domain/infra/abi/containers.go
rhatdan: Ok, I am not sure that is possible, but I can change this to fall though, since it does not change the error.

pkg/api/handlers/compat/containers.go
rhatdan: This is the only endpoint we have for removing containers.  All of the Docker parameters still work, is it invalid to use additional libpod params on a compat endpoint?

@baude @jwhonce WDYT?

vrothberg: You can make it conditional which is something other endpoints are doing.  `utils.IsLibpodRequest(r)` returns a boolean that indicates if the request was made against a `libpod/...` endpoint.

pkg/domain/infra/abi/containers.go
mheon: I'm not sure you're right here - the check for whether a container is a Libpod container in `RemoveStorageContainer` happens before we try to remove the container, so we should already have gotten an `ErrCtrExists`

pkg/domain/infra/abi/containers.go
mheon: We should not be erroring in that case, we should be proceeding to normal `RemoveContainer` *without* appending the report.

rhatdan: Right it does not add the report now.

pkg/domain/infra/abi/containers.go
rhatdan: If you run @edsantiago test you will see this happen.

mheon: Then we need to catch this inside `RemoveStorageContainer` - why work around broken behavior when we can just fix the broken API

rhatdan: RemoveStorageContainer is not broken, it is returning a valid error, that we are returning to the user.  The problem is we are ignoring the error and then falling through to the libpod remove, which finds that there is no Container libpod container, and reports container does not exists.

rhatdan:
This is ready to merge.

pkg/domain/infra/abi/containers.go
```diff
@@ -264,30 +264,30 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
...
+		report.Err = ic.Libpod.RemoveStorageContainer(ctr, options.Force)
+		switch errors.Cause(report.Err) {
+		case nil:
 			// remove container names that we successfully deleted
-			tmpNames = append(tmpNames, ctr)
-		} else {
 			reports = append(reports, &report)
+		case define.ErrNoSuchCtr:
+			// There is still a potential this is a libpod container
+			tmpNames = append(tmpNames, ctr)
```
mheon, line 279: I think we duplicate ctr in tmpNames here - we go through the switch and then append again after the switch

rhatdan: No the case statement does not fall though to the default block.

jwhonce:
LGTM

pkg/api/handlers/compat/containers.go
```diff
@@ -49,34 +53,31 @@ func RemoveContainer(w http.ResponseWriter, r *http.Request) {
...
-	con, err := runtime.LookupContainer(name)
-	if err != nil && errors.Cause(err) == define.ErrNoSuchCtr {
-		// Failed to get container. If force is specified, get the container's ID
-		// and evict it
-		if !query.Force {
+	options := entities.RmOptions{
+		All:     query.All,
+		Force:   query.Force,
+		Volumes: query.Volumes,
+		Ignore:  query.Ignore,
```
TomSweeneyRedHat, line 64: nit, alpha order this plz.

TomSweeneyRedHat:
LGTM
I'm going to push this one over the finish line as I know you've another one tagging behind this.  Please fix my nit and address any remaining concerns from @mheon in that PR.

pkg/domain/infra/abi/containers.go
mheon: It doesn't need to fall through - the append happens again after the switch block ends.

rhatdan: Could you send a link to the exact line, you are talking about? Maybe I am missing it, but this is the default block, and there is no append after this.
https://github.com/containers/podman/blob/master/pkg/domain/infra/abi/containers.go#L280-L288

mheon: Nevermind, you're right, I'm misreading it

PR #9018, opened 2021-01-19
Add shell pipe support for "play kube"

### LINKED ISSUE #8996: "podman play kube" pipe input
/kind feature

It would be great if `podman play kube` could use input from pipe. I try to use kustomize with podman and would like to do something like `kustomize build my-overlay | podman play kube`.

ADIX7:
Closes #8996

Similar to `cat pod.json | kubectl apply -f -` this commit adds support
for input from shell pipe (ex `cat pod.json | podman play kube -`).
To use input form pipe one may use '-' as file name, the same way as
`kubectl apply -f -`

Signed-off-by: Ádám Kovács <kovacsadam07@outlook.hu>

### DIFF
--- a/pkg/bindings/play/play.go
+++ b/pkg/bindings/play/play.go
@@ -9,6 +9,7 @@ import (
 	"github.com/containers/podman/v2/pkg/auth"
 	"github.com/containers/podman/v2/pkg/bindings"
 	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/pkg/errors"
 )
 
 func Kube(ctx context.Context, path string, options *KubeOptions) (*entities.PlayKubeReport, error) {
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM, thanks for contributing!

TomSweeneyRedHat:
LGTM
Needs to be unset as a WIP before merging

ADIX7:
Your welcome! The documentation and tests are still needed. I guess they should be done before merging, right?

Luap99:
> The documentation and tests are still needed. I guess they should be done before merging, right?

Yes, please

Luap99 (commented):
This does not work with `podman-remote`. You have to make a similar change in `pkg/domain/infra/tunnel/play.go`

ADIX7:
What should be the new usage syntax? `kube [options] KUBEFILE|-` ? Or it should remain `kube [options] KUBEFILE` ?

rhatdan:
I think we can just leave this as kube [options] KUBEFILE  and then explain in the decription that "-" can be used as a kubefile name.

ADIX7:
I see https://github.com/containers/podman/pull/9420 went in, closing this one.

Issue #9000, opened 2021-01-18
make fails with undefined: jsoniter error


dc25:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Steps to reproduce the issue:**

1. Clone the podman repo. (per instructions)

2. make BUILDTAGS="selinux seccomp"

3. observe output as shown below

```

dave@coconut:/media/veracrypt1/repos/gitnc$ mkdir -p $GOPATH/src/github.com/containers/podman
dave@coconut:/media/veracrypt1/repos/gitnc$ git clone https://github.com/containers/podman/ $GOPATH/src/github.com/containers/podman
Cloning into '/home/dave/go/src/github.com/containers/podman'...
remote: Enumerating objects: 42, done.
remote: Counting objects: 100% (42/42), done.
remote: Compressing objects: 100% (33/33), done.
remote: Total 103186 (delta 12), reused 16 (delta 7), pack-reused 103144
Receiving objects: 100% (103186/103186), 66.58 MiB | 9.57 MiB/s, done.
Resolving deltas: 100% (72096/72096), done.
dave@coconut:/media/veracrypt1/repos/gitnc$ cd $GOPATH/src/github.com/containers/podman
dave@coconut:~/go/src/github.com/containers/podman$ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
dave@coconut:~/go/src/github.com/containers/podman$ git log | head
commit 5b3c7a52939275784c45c9747dd5864bd0581ff5
Merge: 00ab9cab0 caaaa2c5e
Author: OpenShift Merge Robot <openshift-merge-robot@users.noreply.github.com>
Date:   Sun Jan 17 15:38:30 2021 -0500

    Merge pull request #8995 from vrothberg/lint-lint-lint
    
    [CI:DOCS] lint lint lint

commit 00ab9cab01af125a01ea53b5f48c0d13943f85a1
dave@coconut:~/go/src/github.com/containers/podman$ make BUILDTAGS="selinux seccomp"
touch .gopathok
Podman is being compiled without the systemd build tag. Install libsystemd on 	Ubuntu or systemd-devel on rpm based distro for journald support.
[...]

PR #9020, opened 2021-01-19
make bindings generation more robust

vrothberg:
The Go gods did not shine upon us trying to understand what's going on
in #9000.  The symptom is that `go generate` did not add required
imports to a generated file, ultimately breaking subsequent compilation.

While it still remains unclear *why* Go is behaving like that, the
symptom disappears when `go generate` runs in module mode; that is
without `-mod=vendor` and without `GO111MODULE=off`.  This was
reproducible on two separate machines (Ubuntu and Fedora).

Also, when facing an unset GOPATH, set it to Go's default (i.e.,
$HOME/go) and make sure that GOBIN is in PATH since `goimports`
is required by `go generate`.

Fixes: #9000

@baude @rhatdan @edsantiago PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -464,6 +464,7 @@ podman-remote-%-release:
 	rm -f release.txt
 	$(MAKE) podman-remote-release-$*.zip
 
+.PHONY: generate-bindings
 generate-bindings:
 ifneq ($(shell uname -s), Darwin)
 	GO111MODULE=off $(GO) generate ./pkg/bindings/... ;
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ PODMAN_SERVER_LOG ?=
 
 # If GOPATH not specified, use one in the local directory
 ifeq ($(GOPATH),)
-export GOPATH := $(CURDIR)/_output
+export GOPATH := $(HOME)/go
 unexport GOBIN
 endif
 FIRST_GOPATH := $(firstword $(subst :, ,$(GOPATH)))
@@ -97,6 +97,8 @@ GOBIN := $(shell $(GO) env GOBIN)
 ifeq ($(GOBIN),)
 GOBIN := $(FIRST_GOPATH)/bin
 endif
+
+export PATH := $(PATH):$(GOBIN)
 
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 
@@ -466,7 +468,7 @@ podman-remote-%-release:
 BINDINGS_SOURCE = $(wildcard pkg/bindings/**/types.go)
 .generate-bindings: $(BINDINGS_SOURCE)
 ifneq ($(shell uname -s), Darwin)
-	$(GO) generate -mod=vendor ./pkg/bindings/... ;
+	GO111MODULE=off $(GO) generate ./pkg/bindings/... ;
 endif
 	touch .generate-bindings
 
--- a/Makefile
+++ b/Makefile
@@ -208,7 +208,7 @@ endif
 podman: bin/podman
 
 .PHONY: bin/podman-remote
-bin/podman-remote: .gopathok .generate-bindings $(SOURCES) go.mod go.sum ## Build with podman on remote environment
+bin/podman-remote: .gopathok $(SOURCES) go.mod go.sum ## Build with podman on remote environment
 	$(GO) build $(BUILDFLAGS) -gcflags '$(GCFLAGS)' -asmflags '$(ASMFLAGS)' -ldflags '$(LDFLAGS_PODMAN)' -tags "${REMOTETAGS}" -o $@ ./cmd/podman
 
 .PHONY: bin/podman-remote-static
@@ -279,7 +279,6 @@ clean: ## Clean artifacts
 		libpod/container_easyjson.go \
 		libpod/pod_easyjson.go \
 		.install.goimports \
-		.generate-bindings \
 		docs/build
 	make -C docs clean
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -98,6 +98,8 @@ ifeq ($(GOBIN),)
 GOBIN := $(FIRST_GOPATH)/bin
 endif
 
+export PATH := $(PATH):$(GOPATH)
```
Luap99, line 101: I think this should be GOBIN and not GOPATH

vrothberg: Good catch, thanks!

TomSweeneyRedHat:
LGTM

edsantiago:
I still can't get this to work on my f33 laptop. After the `go generate ./pkg/bindings/...` step, my tree is left in a bad state:
```diff
index 5aec4b479..46a57eae6 100644
--- a/pkg/bindings/play/types_kube_options.go
+++ b/pkg/bindings/play/types_kube_options.go
@@ -6,8 +6,8 @@ import (
        "strconv"
        "strings"

-       jsoniter "github.com/json-iterator/go"
-       "github.com/pkg/errors"
+       jsoniter "github.com/containers/libpod/src/github.com/json-iterator/go"
+       "github.com/containers/libpod/src/github.com/pkg/errors"
 )

 /*
```
Easily fixed via `git reset --hard`, but I have no idea how to debug this; and I suspect that if I'm running into this, others will too. I'm really uncomfortable with this `bindings` step running as part of a regular `make`.

edsantiago:
I still can't get this to work on my f33 laptop. After the `go generate ./pkg/bindings/...` step, my tree is left in a bad state:
```diff
index 5aec4b479..46a57eae6 100644
--- a/pkg/bindings/play/types_kube_options.go
+++ b/pkg/bindings/play/types_kube_options.go
@@ -6,8 +6,8 @@ import (
        "strconv"
        "strings"

-       jsoniter "github.com/json-iterator/go"
-       "github.com/pkg/errors"
+       jsoniter "github.com/containers/libpod/src/github.com/json-iterator/go"
+       "github.com/containers/libpod/src/github.com/pkg/errors"
 )

 /*
```
Easily fixed via `git reset --hard`, but I have no idea how to debug this; and I suspect that if I'm running into this, others will too. I'm really uncomfortable with this `bindings` step running as part of a regular `make`.

vrothberg:
@edsantiago thanks for checking!  I am not yet happy as well but suggest to go step by step.  Maybe we find a better way in the (hopefully) near future.

Luap99:
I also believe we should not run this for make or make binaries. All `pkg/bindings` files are committed. You only need to run `make .generate-bindings` when you changed something in `pkg/bindings/*/types.go`. I think this can be up to the individual developer in that case. It is also easy to enforce in CI, e.g. run generate bindings and check if the tree is still clean.

mheon:
My ideal situation would be that this `make` target is triggered by changes in those files, and otherwise does not fire.

vrothberg:
> My ideal situation would be that this `make` target is triggered by changes in those files, and otherwise does not fire.

In this case, we have to add the `.generate-bindings` file to Git.  But notice that we're a distributed team (different time zones) and that Make works on timestamps. So every time we update the file (and re-generate the bindings), there may be some false positives for a day.

vrothberg:
> I also believe we should not run this for make or make binaries. All `pkg/bindings` files are committed. You only need to run `make .generate-bindings` when you changed something in `pkg/bindings/*/types.go`. I think this can be up to the individual developer in that case. It is also easy to enforce in CI, e.g. run generate bindings and check if the tree is still clean.

I don't agree. History shows that if things are up to individual developers, things go wrong. I don't want to go back to the pre `make vendor` days.

We need to make sure that the generated code is _always_ in sync.

vrothberg:
> > I also believe we should not run this for make or make binaries. All `pkg/bindings` files are committed. You only need to run `make .generate-bindings` when you changed something in `pkg/bindings/*/types.go`. I think this can be up to the individual developer in that case. It is also easy to enforce in CI, e.g. run generate bindings and check if the tree is still clean.
> 
> I don't agree. History shows that if things are up to individual developers, things go wrong. I don't want to go back to the pre `make vendor` days.
> 
> We need to make sure that the generated code is _always_ in sync.

I just realize now that this is not mutually exclusive.  As @Luap99 said, we can easily enforce it in the CI.  That may be the smoothest way forward.

vrothberg:
Alright, the new CI job is working: https://github.com/containers/podman/pull/9020/checks?check_run_id=1729142191

vrothberg:
Updated as discussed:
* `make generate-bindings` is explicit now
* Added a CI check similar to `make vendor` and integrated that together with the vendor checks into "code consistency" task.

@edsantiago @Luap99 @mheon PTAL

vrothberg:
@cevich PTAL

Makefile
```diff
@@ -459,12 +460,10 @@ podman-remote-%-release:
 	rm -f release.txt
 	$(MAKE) podman-remote-release-$*.zip
 
-BINDINGS_SOURCE = $(wildcard pkg/bindings/**/types.go)
-.generate-bindings: $(BINDINGS_SOURCE)
+generate-bindings:
```
edsantiago, line 463: I think you need `.PHONY: generage-bindings` right above this

mheon:
LGTM once Ed's comment is addressed. Hoping we can get this landed soon, having to fight bindings generation is not fun.

Makefile
vrothberg: Thanks for checking!

Makefile
```diff
@@ -463,12 +464,11 @@ podman-remote-%-release:
 	rm -f release.txt
 	$(MAKE) podman-remote-release-$*.zip
 
-BINDINGS_SOURCE = $(wildcard pkg/bindings/**/types.go)
-.generate-bindings: $(BINDINGS_SOURCE)
+.PHONY: geenrate-bindings
```
edsantiago, line 467: Aw, man, sorry to do this, but that's a typo

vrothberg: Too many things in parallel ... sorry for my sloppiness

edsantiago:
Tests green.

/lgtm

There's a prior hold, remove it at will

vrothberg:
Thanks everybody!  I think we made some good improvements

Luap99:
I think we should backport this, right?

PR #9023, opened 2021-01-19
[CI:DOCS] fix go-md2man HTMLSpan warnings

edsantiago:
I'm tired of seeing these every time I run 'make':

   WARNING: go-md2man does not handle node type HTMLSpan

Cause: left-angle-brackets ( < ) in document source

Solution:
  1) backquote-escape those that need to be shown, usually
     ones referring to an argument or email address; or
  2) Actual HTML ( &lt;sup&gt; and &lt;a&gt; ) which are meant to be
     shown in generated HTML docs but can't be shown in
     man pages, we filter out via a sed expression.

### DIFF
--- a/docs/source/markdown/podman-system-service.1.md
+++ b/docs/source/markdown/podman-system-service.1.md
@@ -41,5 +41,5 @@ podman system service --timeout 5000
 podman(1), podman-system-service(1), podman-system-connection(1)
 
 ## HISTORY
-January 2020, Originally compiled by Brent Baude<bbaude@redhat.com>
-November 2020, Updated by Jhon Honce <jhonce at redhat.com>
+January 2020, Originally compiled by Brent Baude `<bbaude@redhat.com>`
+November 2020, Updated by Jhon Honce (jhonce at redhat dot com)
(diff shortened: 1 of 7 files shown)


mheon:
LGTM

docs/source/markdown/podman-system-service.1.md
```diff
@@ -41,5 +41,5 @@ podman system service --timeout 5000
 podman(1), podman-system-service(1), podman-system-connection(1)
 
 ## HISTORY
-January 2020, Originally compiled by Brent Baude<bbaude@redhat.com>
-November 2020, Updated by Jhon Honce <jhonce at redhat.com>
+January 2020, Originally compiled by Brent Baude `<bbaude@redhat.com>`
+November 2020, Updated by Jhon Honce `<jhonce@redhat.com>`
```
TomSweeneyRedHat, line 45: @jwhonce does this work for you?

edsantiago: Thanks for the comment, @TomSweeneyRedHat . `grep -R honce docs` shows that Jhon has an established convention for his email address in md files, so I've updated `podman-system-service.1.md` accordingly and re-pushed.

PR #9025, opened 2021-01-19
Ensure the Volumes field in Compat Create is honored

### LINKED ISSUE #8649: [APIv2]: container copy rewrite broke PUT /containers/{id}/archive endpoint

/kind bug

**Description**

Trying to find regressions by exercising the APIv2 trough [docker-py](https://github.com/docker/docker-py)'s test suite (see #5386) I came across a regressions in the [`ArchiveTest.test_copy_directory_to_container`](https://github.com/docker/docker-py/blob/master/tests/integration/api_container_test.py#L691) test.

It specifically broke with commit 0f496e4b81aadbc6e38cc02af67ddc58f1f3ef79 .

Assuming you have a tar archive for upload:

```shell
touch test.txt
tar cf test.tar test.txt
```

1. create a container 
```
curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/containers/create' -H 'Content-Type: application/json' -d '{"Tty": false, "OpenStdin": false, "StdinOnce": false, "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Cmd": ["ls", "-p", "/vol1"], "Image": "alpine:3.10", "Volumes": {"/vol1": {}}, "NetworkDisabled": false}'
```

2. upload the created archive to it
```
curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPUT 'http://localhost/v1.40/containers/a8922857d1daf9bbcff5b3703c42e31101ef9d0ed7b956e9f2359acea2449d99/archive?path=/vol1' --data-binary '@test.tar' -H 'Content-Type:'
```

The archive upload fails with:
```
200 {"cause":"destination must be a directory or stream when copying from a stream","message":"destination must be a directory or stream when copying from a stream","response":500}
```
**Note:** the `200` response code. :disappointed:

The contents of the archive should be extracted into the directory given in the `path` parameter (here: `"/vol1"`).

The 200 response code in case of an error is because of https://github.com/containers/podman/blob/0cccba834ff2939579f7ef4c340a9337dc9d7d2c/pkg/api/handlers/compat/containers_archive.go#L173 writing the `http.StatusOK` header out before the last fallible operation. 

The copy error happens in:
https://github.com/containers/podman/blob/0cccba834ff2939579f7ef4c340a9337dc9d7d2c/pkg/copy/copy.go#L120
called by
https://github.com/containers/podman/blob/0cccba834ff2939579f7ef4c340a9337dc9d7d2c/pkg/copy/copy.go#L32
called by
https://github.com/containers/podman/blob/0cccba834ff2939579f7ef4c340a9337dc9d7d2c/pkg/api/handlers/compat/containers_archive.go#L174
... incidentally exactly after setting the `http.StatusOK` header.

But the actual cause is that `destination` is invalid in
https://github.com/containers/podman/blob/0cccba834ff2939579f7ef4c340a9337dc9d7d2c/pkg/api/handlers/compat/containers_archive.go#L159

Because the `resolvedContainerPath` (something like: `"/home/riyad/podman/storage/overlay/02a2eb10376916256ef680acfbdf9296b48f3b4941a41d08e478618c228d68a9/merged/vol1"`) does not exist in
https://github.com/containers/podman/blob/0cccba834ff2939579f7ef4c340a9337dc9d7d2c/pkg/copy/item.go#L237
we have `statError != nil` (specifically `"file does not exist"`).
This leads us to fall through the conditions in L241 and L254 and enter the last block of code where we make sure the "parent" directory for `resolvedContainerPath` exists (but not the directory itself?). :shrug: 

All this worked before 0f496e4b81aadbc6e38cc02af67ddc58f1f3ef79.

```
Version:      3.0.0-dev
API Version:  3.0.0
Go Version:   go1.15.6
Git Commit:   9b3a81a002e570b8a49e60c3dd3feb65d742f286-dirty
Built:        Tue Dec  8 16:08:02 2020
OS/Arch:      linux/amd64
```

mheon:
Docker has, for unclear reasons, three separate fields in their Create Container struct in which volumes can be placed. Right now we support two of those - Binds and Mounts, which (roughly) correspond to `-v` and `--mount` respectively. Unfortunately, we did not support the third, `Volumes`, which is used for anonymous named volumes created by `-v` (e.g. `-v /test`). It seems that volumes listed here are *not* included in the remaining two from my investigation, so it should be safe to just append them into our handling of the `Binds` (`-v`) field.

Fixes #8649

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -3,6 +3,7 @@ package common
 import (
 	"fmt"
 	"net"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -383,8 +384,29 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	}
 
 	// volumes
-	if volumes := cc.HostConfig.Binds; len(volumes) > 0 {
-		cliOpts.Volume = volumes
+	volDestinations := make(map[string]bool)
+	for _, vol := range cc.HostConfig.Binds {
+		cliOpts.Volume = append(cliOpts.Volume, vol)
+		// Extract the destination so we don't add duplicate mounts in
+		// the volumes phase.
+		splitVol := strings.SplitN(vol, ":", 3)
+		switch len(splitVol) {
+		case 1:
+			volDestinations[vol] = true
+		default:
+			volDestinations[splitVol[1]] = true
+		}
+	}
+	// Anonymous volumes are added differently from other volumes, in their
+	// own special field, for reasons known only to Docker. Still use the
+	// format of `-v` so we can just append them in there.
+	// Unfortunately, these may be duplicates of existing mounts in Binds.
+	// So... We need to catch that.
+	for vol := range cc.Volumes {
+		if _, ok := volDestinations[filepath.Clean(vol)]; ok {
+			continue
+		}
+		cliOpts.Volume = append(cliOpts.Volume, vol)
 	}
 	if len(cc.HostConfig.BlkioWeightDevice) > 0 {
 		devices := make([]string, 0, len(cc.HostConfig.BlkioWeightDevice))
--- a/test/apiv2/20-containers.at
+++ b/test/apiv2/20-containers.at
@@ -237,3 +237,12 @@ t GET containers/$cid/json 200 \
 t DELETE containers/$cid 204
 t DELETE images/${MultiTagName}?force=true 200
 # vim: filetype=sh
+
+# Test Volumes field adds an anonymous volume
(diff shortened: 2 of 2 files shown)


mheon:
Apparently it is not a safe assumption that `Volumes` is distinct, because Compose likes to populate it as well as the other two. Fun!

rhatdan:
Looks like your test is failing.

mheon:
It is actually a different test that was broken by the addition of my test, which is confusing since I do clean up everything. I will take another look today.

mheon:
Looks like this is going to go green - @containers/podman-maintainers PTAL

vrothberg (commented):
LGTM

PR #9027, opened 2021-01-19
Podman volume exists

Luap99:
Add podman volume exists command with remote support.

### DIFF
--- a/docs/source/markdown/podman-volume-exists.1.md
+++ b/docs/source/markdown/podman-volume-exists.1.md
@@ -1,0 +1,43 @@
+% podman-volume-exists(1)
+
+## NAME
+podman\-volume\-exists - Check if the given volume exists
+
+## SYNOPSIS
+**podman volume exists** *volume*
+
+## DESCRIPTION
+**podman volume exists** checks if a volume exists. Podman will return an exit code
+of `0` when the volume is found. A `1` will be returned otherwise. An exit code of
+`125` indicates there was another issue.
+
+
+## OPTIONS
+
+#### **--help**, **-h**
+
+Print usage statement
+
+## EXAMPLE
+
+Check if a volume called `myvol` exists (the volume does actually exist).
+```
+$ podman volume exists myvol
+$ echo $?
+0
+$
+```
+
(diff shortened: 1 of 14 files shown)


rhatdan:
Looks like make validate does not like your man page.

pkg/api/handlers/libpod/volumes.go
```diff
@@ -203,3 +203,20 @@ func RemoveVolume(w http.ResponseWriter, r *http.Request) {
 	}
 	utils.WriteResponse(w, http.StatusNoContent, "")
 }
+
+// ExistsVolume check if a volume exists
+func ExistsVolume(w http.ResponseWriter, r *http.Request) {
+	runtime := r.Context().Value("runtime").(*libpod.Runtime)
```
rhatdan, line 209: I would prefer this to use the function in pkg/domain/infra/abi/volumes.go

We are going to be changing all of the handlers to use abi functions, to make sure podman-remote and other APIV2 tools use the same code as local podman.

Luap99: Ok, understood.

baude:
LGTM

docs/source/markdown/podman-volume-exists.1.md
```diff
@@ -0,0 +1,43 @@
...
+## NAME
+podman\-volume\-exists - Check if the given volume exists
+
+## SYNOPSIS
+**podman volume exists** *volume*
+
+## DESCRIPTION
+**podman volume exists** checks if a volume exists. Podman will return an exit code
+of `0` when the volume is found. A `1` will be returned otherwise. An exit code of
+`125` indicates there was an other issue.
```
edsantiago, line 12: 'another' is one word. (This is not worth a re-push. I mention it only if you need to resubmit for another reason).

rhatdan: U would rather fix this, since it is Human facing.
@Luap99 Fix this and then LGTM and we can merge.

Luap99:
This one is ready to merge.

edsantiago:
I was hoping someone else would jump in to say this, but I guess I'll be the bad guy. Could you squash please?

PR #9028, opened 2021-01-19
Accept and ignore 'null' as value for X-Registry-Auth

mlegenovic:
docker-client is a library written in Java and used in Eclipse to
speak with Docker API. When endpoint /images/search is called,
HTTP header attribute X-Registry-Auth has value "null". This is for
sure wrong but Docker tolerates this value, and call works. With this
patch call works also with Podman. #7857

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -297,7 +297,9 @@ func imageAuthToDockerAuth(authConfig types.DockerAuthConfig) dockerAPITypes.Aut
 func singleAuthHeader(r *http.Request) (map[string]types.DockerAuthConfig, error) {
 	authHeader := r.Header.Get(string(XRegistryAuthHeader))
 	authConfig := dockerAPITypes.AuthConfig{}
-	if len(authHeader) > 0 {
+	// Accept "null" and handle it as empty value for compatibility reason with Docker.
+	// Some java docker clients pass this value, e.g. this one used in Eclipse.
+	if len(authHeader) > 0 && authHeader != "null" {
 		authJSON := base64.NewDecoder(base64.URLEncoding, strings.NewReader(authHeader))
 		if err := json.NewDecoder(authJSON).Decode(&authConfig); err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat (commented):
Any tests that can be added.
@vrothberg WDYT?  I'm a bit leery about making this change, but I'm not seeing another way after going through the issue and this PR.

mlegenovic:
@TomSweeneyRedHat I am also not happy about the proposed change, that is why I took a look at docker/moby code to see how they do it (https://github.com/moby/moby/blob/master/api/server/router/distribution/distribution_routes.go#L37)
Comment there is: 

> for a search it is not an error if no auth was given, to increase compatibility with the existing api it is defaulting to be empty

It looks like they are parsing this value in various places in the code, and therefore they can make the behavior different from case to case. In Podman this code is centralized (of course much better, less copy/paste code), but it is more difficult to handle such exceptional cases.

I can of course create an issue in other project ([docker-client](https://github.com/dmandalidis/docker-client)), and let them solve the problem ("null" as value is definitely wrong), if you find this change quite ugly.

pkg/auth/auth.go
```diff
@@ -297,7 +297,7 @@ func imageAuthToDockerAuth(authConfig types.DockerAuthConfig) dockerAPITypes.Aut
 func singleAuthHeader(r *http.Request) (map[string]types.DockerAuthConfig, error) {
 	authHeader := r.Header.Get(string(XRegistryAuthHeader))
 	authConfig := dockerAPITypes.AuthConfig{}
-	if len(authHeader) > 0 {
+	if len(authHeader) > 0 && authHeader != "null" {
```
vrothberg, line 300: Please add a comment to both places that explains why need to ignore "null".

vrothberg (commented):
The change looks good to me.  There are other places in the code where we've made Java clients happy to remain compatible with Docker.

@jwhonce PTAL

vrothberg:
@mlegenovic can you run `make vendor` and update the commit?

vrothberg:
CI is failing but will be fixed with https://github.com/containers/podman/pull/9031. Once merged, please rebase.

rhatdan:
Thanks @mlegenovic 
/approve
/lgtm
Bug for bug compatible.

PR #9038, opened 2021-01-20
Add binding options for container|pod exists

baude:
It turns out an options was added to container exists so it makes sense
to have pods and container exists calls have an optional structure for
options.


vrothberg (commented):
LGTM

PR #9048, opened 2021-01-21
Fix Docker APIv2 container wait endpoint

### LINKED ISSUE #8697: APIv2 /containers/{id}/wait doesn't work with correct condition enumeration
Currently the used enumeration is here:
podman/libpod/define/containerstate.go

```go
const (
	ContainerStateUnknown ContainerStatus = iota
	ContainerStateConfigured ContainerStatus = iota
	ContainerStateCreated ContainerStatus = iota
	ContainerStateRunning ContainerStatus = iota
	ContainerStateStopped ContainerStatus = iota
	ContainerStatePaused ContainerStatus = iota
	ContainerStateExited ContainerStatus = iota
	ContainerStateRemoving ContainerStatus = iota
)
```

But docker uses different values:
```go
const (
	WaitConditionNotRunning WaitCondition = "not-running"
	WaitConditionNextExit   WaitCondition = "next-exit"
	WaitConditionRemoved    WaitCondition = "removed"
)
```

### LINKED ISSUE #9157: podman-remote returns invalid information to docker

/kind bug

**Description**

1. Run podman-remote

2. Start docker container

3.

~~~
podman run --rm -t  --security-opt label=disable -v /run/user/1234/podman/podman.sock:/var/run/docker.sock:z docker.io/library/docker "docker" "run" "-e" "POSTGRES_PASSWORD=mysecretpassword" "-p" "5432:5432" "-d" "docker.io/library/postgres:12"
0ec7122bd71c2895473f84b9420d2e8ca2f7d398e0943c5ac9fa8eb72e50ac71
ERRO[0000] error waiting for container: Error response from daemon: unknown container state: next-exit: invalid argument 
~~~

Information returned to docker should be compatible/valid for docker and not let the command fail, as the container is actually running.

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 15:37:50 2020
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 0f53fb68333bdead5fe4dc5175703e22cf9882ab'
  cpus: 24
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: xxx
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 110147
      size: 1
    - container_id: 1
      host_id: 120000
      size: 165536
    uidmap:
    - container_id: 0
      host_id: 110147
      size: 1
    - container_id: 1
      host_id: 120000
      size: 165536
  kernel: 5.9.16-200.fc33.x86_64
  linkmode: dynamic
  memFree: 6367051776
  memTotal: 67409752064
  ociRuntime:
    name: crun
    package: crun-0.17-1.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.17
      commit: 0e9229ae34caaebcb86f1fde18de3acaf18c6d9a
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/xxx/podman/podman.sock
  rootless: true
  slirp4netns:
    package: slirp4netns-1.1.8-1.fc33.x86_64
    version: |-
      slirp4netns version 1.1.8
      commit: d361001f495417b880f20329121e3aa431a8f90f
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 3
      libseccomp: 2.5.0
  swapFree: 35918303232
  swapTotal: 38071689216
  uptime: 433h 29m 37.43s (Approximately 18.04 days)
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - registry.centos.org
  - docker.io
store:
  configFile: /home/xxx/.config/containers/storage.conf
  containerStore:
    number: 1
    paused: 0
    running: 1
    stopped: 0
  graphDriverName: overlay
  graphOptions:
    overlay.mount_program:
      Package: fuse-overlayfs-1.4.0-1.fc33.x86_64
      Version: |-
        fusermount3 version: 3.9.3
        fuse-overlayfs: version 1.4
        FUSE library version 3.9.3
  graphRoot: /home/xxx/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 1046
  runRoot: /run/user/110147/containers
  volumePath: /home/xxx/.local/share/containers/storage/volumes
version:
  APIVersion: 2.1.0
  Built: 1607438270
  BuiltTime: Tue Dec  8 15:37:50 2020
  GitCommit: ""
  GoVersion: go1.15.5
  OsArch: linux/amd64
  Version: 2.2.1
```

```
podman-2.2.1-1.fc33.x86_64
podman-remote-2.2.1-1.fc33.x86_64
```

No

`docker version` from inside podman:

~~~
Client: Docker Engine - Community
 Version:           20.10.2
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:11:26 2020
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: linux/amd64/fedora-33
 Podman Engine:
  Version:          2.2.1
  APIVersion:       2.0.0
  Arch:             amd64
  BuildTime:        2020-12-08T15:37:50+01:00
  Experimental:     true
  GitCommit:        
  GoVersion:        go1.15.5
  KernelVersion:    5.9.16-200.fc33.x86_64
  MinAPIVersion:    2.0.0
  Os:               linux
 Engine:
  Version:          2.2.1
  API version:      1.40 (minimum version 1.24)
  Go version:       go1.15.5
  Git commit:       
  Built:            Tue Dec  8 15:37:50 2020
  OS/Arch:          linux/amd64
  Experimental:     true
~~~

matejvasek:
resolves #8697
resolves #9157
* container wait functions are now cancel-able via `context.Context`
* container wait functions can wait for multiple states, not just single one
* docker container wait endpoint now accepts docker conditions: `not-running`, `removed`, `next-exit`
* docker container wait endpoint now sends headers in advance before actual waiting

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/containers/podman/v2/libpod/define"
@@ -478,13 +479,15 @@ func (c *Container) RemoveArtifact(name string) error {
 }
 
 // Wait blocks until the container exits and returns its exit code.
-func (c *Container) Wait() (int32, error) {
-	return c.WaitWithInterval(DefaultWaitInterval)
+func (c *Container) Wait(ctx context.Context) (int32, error) {
+	return c.WaitWithInterval(ctx, DefaultWaitInterval)
 }
+
+var errWaitingCanceled = errors.New("waiting was canceled")
 
 // WaitWithInterval blocks until the container to exit and returns its exit
 // code. The argument is the interval at which checks the container's status.
-func (c *Container) WaitWithInterval(waitTimeout time.Duration) (int32, error) {
+func (c *Container) WaitWithInterval(ctx context.Context, waitTimeout time.Duration) (int32, error) {
 	if !c.valid {
 		return -1, define.ErrCtrRemoved
 	}
@@ -495,41 +498,111 @@ func (c *Container) WaitWithInterval(waitTimeout time.Duration) (int32, error) {
 	}
 	chWait := make(chan error, 1)
 
-	defer close(chWait)
+	go func() {
+		<-ctx.Done()
+		chWait <- errWaitingCanceled
+	}()
 
 	for {
-		// ignore errors here, it is only used to avoid waiting
+		// ignore errors here (with exception of cancellation), it is only used to avoid waiting
 		// too long.
-		_, _ = WaitForFile(exitFile, chWait, waitTimeout)
+		_, e := WaitForFile(exitFile, chWait, waitTimeout)
+		if e == errWaitingCanceled {
+			return -1, errWaitingCanceled
+		}
 
-		stopped, err := c.isStopped()
+		stopped, code, err := c.isStopped()
 		if err != nil {
 			return -1, err
 		}
 		if stopped {
-			return c.state.ExitCode, nil
+			return code, nil
 		}
 	}
 }
 
-func (c *Container) WaitForConditionWithInterval(waitTimeout time.Duration, condition define.ContainerStatus) (int32, error) {
+type waitResult struct {
+	code int32
+	err  error
+}
+
+func (c *Container) WaitForConditionWithInterval(ctx context.Context, waitTimeout time.Duration, conditions ...define.ContainerStatus) (int32, error) {
 	if !c.valid {
 		return -1, define.ErrCtrRemoved
 	}
-	if condition == define.ContainerStateStopped || condition == define.ContainerStateExited {
-		return c.WaitWithInterval(waitTimeout)
+
+	if len(conditions) == 0 {
+		panic("at least one condition should be passed")
 	}
-	for {
-		state, err := c.State()
-		if err != nil {
-			return -1, err
+
+	ctx, cancelFn := context.WithCancel(ctx)
+	defer cancelFn()
+
+	resultChan := make(chan waitResult)
+	waitForExit := false
+	wantedStates := make(map[define.ContainerStatus]bool, len(conditions))
+
+	for _, condition := range conditions {
+		if condition == define.ContainerStateStopped || condition == define.ContainerStateExited {
+			waitForExit = true
+			continue
 		}
-		if state == condition {
-			break
-		}
-		time.Sleep(waitTimeout)
+		wantedStates[condition] = true
 	}
-	return -1, nil
+
+	trySend := func(code int32, err error) {
+		select {
+		case resultChan <- waitResult{code, err}:
+		case <-ctx.Done():
+		}
+	}
+
+	var wg sync.WaitGroup
+
+	if waitForExit {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			code, err := c.WaitWithInterval(ctx, waitTimeout)
+			trySend(code, err)
+		}()
+	}
+
+	if len(wantedStates) > 0 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			for {
+				state, err := c.State()
+				if err != nil {
+					trySend(-1, err)
+					return
+				}
+				if _, found := wantedStates[state]; found {
+					trySend(-1, nil)
+					return
+				}
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(waitTimeout):
+					continue
+				}
+			}
+		}()
+	}
+
+	var result waitResult
+	select {
+	case result = <-resultChan:
+		cancelFn()
+	case <-ctx.Done():
+		result = waitResult{-1, errWaitingCanceled}
+	}
+	wg.Wait()
+	return result.code, result.err
 }
 
 // Cleanup unmounts all mount points in container and cleans up container storage
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -483,8 +483,6 @@ func (c *Container) Wait(ctx context.Context) (int32, error) {
 	return c.WaitWithInterval(ctx, DefaultWaitInterval)
 }
 
-var errWaitingCanceled = errors.New("waiting was canceled")
-
 // WaitWithInterval blocks until the container to exit and returns its exit
 // code. The argument is the interval at which checks the container's status.
 func (c *Container) WaitWithInterval(ctx context.Context, waitTimeout time.Duration) (int32, error) {
@@ -500,15 +498,15 @@ func (c *Container) WaitWithInterval(ctx context.Context, waitTimeout time.Durat
 
 	go func() {
 		<-ctx.Done()
-		chWait <- errWaitingCanceled
+		chWait <- define.ErrCanceled
 	}()
 
 	for {
 		// ignore errors here (with exception of cancellation), it is only used to avoid waiting
 		// too long.
 		_, e := WaitForFile(exitFile, chWait, waitTimeout)
-		if e == errWaitingCanceled {
-			return -1, errWaitingCanceled
+		if e == define.ErrCanceled {
+			return -1, define.ErrCanceled
 		}
 
 		stopped, code, err := c.isStopped()
@@ -599,7 +597,7 @@ func (c *Container) WaitForConditionWithInterval(ctx context.Context, waitTimeou
 	case result = <-resultChan:
 		cancelFn()
 	case <-ctx.Done():
-		result = waitResult{-1, errWaitingCanceled}
+		result = waitResult{-1, define.ErrCanceled}
 	}
 	wg.Wait()
 	return result.code, result.err
--- a/pkg/api/handlers/utils/containers.go
+++ b/pkg/api/handlers/utils/containers.go
@@ -23,8 +23,8 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
 	containerEngine := abi.ContainerEngine{Libpod: runtime}
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
-		Interval  string                 `schema:"interval"`
-		Condition define.ContainerStatus `schema:"condition"`
+		Interval  string                   `schema:"interval"`
+		Condition []define.ContainerStatus `schema:"condition"`
 	}{
 		// Override golang default values for types
 	}
@@ -33,7 +33,7 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
 		return 0, err
 	}
 	options := entities.WaitOptions{
-		Condition: define.ContainerStateStopped,
+		Condition: []define.ContainerStatus{define.ContainerStateStopped},
 	}
 	name := GetName(r)
 	if _, found := r.URL.Query()["interval"]; found {
--- a/pkg/api/handlers/utils/containers.go
+++ b/pkg/api/handlers/utils/containers.go
@@ -1,67 +1,230 @@
 package utils
 
 import (
+	"context"
+	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 
-	"github.com/containers/podman/v2/libpod"
-	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/domain/entities"
 	"github.com/containers/podman/v2/pkg/domain/infra/abi"
+
+	"github.com/containers/podman/v2/pkg/api/handlers"
+	"github.com/sirupsen/logrus"
+
+	"github.com/containers/podman/v2/libpod/define"
+
+	"github.com/containers/podman/v2/libpod"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
 )
 
-func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
-	var (
-		err      error
-		interval time.Duration
-	)
-	runtime := r.Context().Value("runtime").(*libpod.Runtime)
-	// Now use the ABI implementation to prevent us from having duplicate
-	// code.
-	containerEngine := abi.ContainerEngine{Libpod: runtime}
-	decoder := r.Context().Value("decoder").(*schema.Decoder)
-	query := struct {
-		Interval  string                   `schema:"interval"`
-		Condition []define.ContainerStatus `schema:"condition"`
-	}{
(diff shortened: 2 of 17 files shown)


pkg/api/handlers/utils/containers.go
```diff
@@ -27,32 +26,92 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
...
+func waitForOneOf(ctx context.Context, con *libpod.Container, stats ...define.ContainerStatus) (int32, error) {
+	newCtx, cancelFn := context.WithCancel(ctx)
+	ch := make(chan struct {
+		code int32
+		err  error
+	}, 1)
+
+	for _, s := range stats {
+		go func(s define.ContainerStatus) {
+			code, err := con.WaitForConditionWithInterval(newCtx, interval, s)
```
mheon, line 77: Probably better to rewrite `WaitForConditionWithInterval` to accept multiple conditions - Libpod API is allowed to change, and it seems a lot more logical to handle this there instead of spawning numerous goroutines to handle it.

pkg/api/handlers/utils/containers.go
```diff
@@ -27,32 +26,92 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
...
-	}
-	condition := define.ContainerStateStopped
+
+	var condition string
 	if _, found := r.URL.Query()["condition"]; found {
-		condition, err = define.StringToContainerStatus(query.Condition)
-		if err != nil {
-			InternalServerError(w, err)
+		condition = query.Condition
+		if condition != "next-exit" && condition != "removed" && condition != "not-running" && condition != "" {
```
mheon, line 33: I think the Libpod version of this has to continue accepting the old conditions as well to ensure backwards compat for that API.

matejvasek:
@mheon PTAL

test/apiv2/26-containersWait.at
```diff
@@ -0,0 +1,47 @@
...
+podman pull "${IMAGE}" &>/dev/null
+
+# Ensure clean slate
+podman rm -a -f &>/dev/null
+
+CTR="WaitTestingCtr"
+
+t POST "containers/nonExistent/wait?condition=next-exit" '' 404
+
+podman create --name "${CTR}" --entrypoint '["sleep", "0.5"]' "${IMAGE}"
```
matejvasek, line 18: I wish I could have just `--entrypoint '["true"]'` here, however such a container exits too fast. :disappointed:

pkg/api/handlers/utils/containers.go
```diff
@@ -1,58 +1,199 @@
...
+		code, err = waitNotRunning(ctx, container, interval)
+	default:
+		cnd, _ := define.StringToContainerStatus(condition)
+		code, err = container.WaitForConditionWithInterval(ctx, interval, cnd)
+	}
+	return code, err
+}
+
+func waitRemoved(ctx context.Context, con *libpod.Container, interval time.Duration) (int32, error) {
+	code, err := con.WaitForConditionWithInterval(ctx, interval, define.ContainerStateUnknown)
```
matejvasek, line 181: Here I am not waiting for specific state but for an error.

test/apiv2/26-containersWait.at
matejvasek: meaning the container is in `running` state for such a short duration we are not capable of detecting that

mheon:
@rhatdan I think this conflicts with some of the work you've been doing, and I think I prefer this version - it enables more functionality that we don't have yet.

rhatdan:
I disagree, I am fine with most of this, but it should all be implemented in the standard API code.  We should not have three different ways of doing the same function.  Too much risk of them drifting, as I have already seen with every single interface I have looked at.

matejvasek:
@rhatdan do you mean I should add the three new condition to the `define.ContainerStatus` `"enum"` and then handle it in `WaitForConditionWithInterval()` ?

mheon:
@rhatdan I disagree. Our API should not support the Docker-specific constants or the specifics of their operation mode. We can try and unify these but it's going to be a massive pain of conditional code

rhatdan:
My point being that podman-remote wait and podman wait should ultimately call into pkg/domain/infra/abi/ wait function.
If Docker API needs some fancy handling, I am fine with adding it.
I would like to get this merged along with the Stop/Remove/Wait patch that I have going into todays release.

mheon:
@rhatdan I don't believe the way Docker's Wait and our Wait work are compatible at this point. The handling for the custom condition codes is fundamentally different from the way our Wait code functions.

matejvasek:
I think that really different is `next-exit` condition. Seems to me that in `podman` `condition == state`. This is not true for `next-exit` since it describes _transition_ between two states instead of one particular state.

matejvasek:
Condition `not-running` is not that different what `podman` has. As for `removed`, there is no such a state since the container no longer exists :smile: ,  so I am basically waiting for `define.ErrNoSuchCtr`.

matejvasek:
Note that `next-exit` may fail for short living containers (even if polling interval is 1ns).

matejvasek:
I did experiment with condition variables it the shared memory as replacement for polling, but results weren't satisfactory.

matejvasek:
To make `next-exit` truly reliable we probably need to introduce some better inter-process notification mechanism. Any ideas what we could use?

rhatdan:
Let's get https://github.com/containers/podman/pull/9051 in and then rebase this on it.
Please review ^^

matejvasek:
@rhatdan I'll look at it and then rebase.

matejvasek:
@mheon 
Given the recent rework -- calling `abi.ContainerEngine` instead of direct call to `Wait*()` on container,
should I change
```go
type WaitOptions struct {
	Condition define.ContainerStatus
	Interval  time.Duration
	Latest    bool
}
```
into
```go
type WaitOptions struct {
	Conditions []define.ContainerStatus
	Interval   time.Duration
	Latest     bool
}
```
or should I prefer handling waiting for multiple states in handler code?

mheon:
That change LGTM

rhatdan:
@matejvasek Needs a rebase.

matejvasek:
@mheon @rhatdan PTAL

matejvasek:
Hmm there is some problem with having `Conditions` as a slice.
```
• Failure [1.761 seconds]
Podman containers 
/home/mvasek/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:18
  podman pause a paused container by name [It]
  /home/mvasek/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:118

  Expected
      <*errors.fundamental | 0xc0006b6060>: {                                                                                                                                                                                                                                 
          msg: "unknown slice type slice",                                                                                                                                                                                                                                    
          stack: [0xec148d, 0xead31f, 0xf64849, 0xf6fa05, 0xec7563, 0xec7177, 0xec6587, 0xecd6d1, 0xecce12, 0xef4951, 0xef4467, 0xef3c77, 0xef6446, 0xf009b8, 0xf00608, 0xfa6a31, 0x530e2f, 0x478ae1],                                                                        
      }                                                                                                                                                                                                                                                                       
  to be nil 
```

matejvasek:
@mheon only slices of strings seems to be supported in some generated code :disappointed: 
so 
```go
type WaitOptions struct {
	Conditions []define.ContainerStatus
	Interval   time.Duration
	Latest     bool
}
```
won't work.

matejvasek:
IMO it could work not only for `string` but also for any type implementing `stringer` interface.

matejvasek:
@rhatdan @mheon I am not sure if I understand all of this: Docker endpoint shall accept params as strings (e.g. "not-running"), and `/libpod` endpoints shall accept numerical values (enum int val)? Is that correct?

matejvasek:
This needs some more work, not really ready for review.

pkg/api/handlers/utils/containers.go
matejvasek: @mheon given the problem that arisen I am considering reverting the change that made `WaitForConditionWithInterval()` variadic.

matejvasek:
I create PR for fixing the generator https://github.com/containers/podman/pull/9203

matejvasek:
The commit `[DO NOT MERGE] Improve generator` must be removed before merging + code has to be merged after https://github.com/containers/podman/pull/9203 is merged.

matejvasek:
I added `[DO NOT MERGE] Improve generator` only temporarily to see if CI pass.

pkg/api/handlers/utils/containers.go
```diff
@@ -1,67 +1,229 @@
...
+			return
+		}
+	}
+
+	name := GetName(r)
+
+	exists, err := containerExists(ctx, name)
+
+	if err != nil {
+		InternalServerError(w, err)
```
matejvasek, line 61: Missing return statements.

matejvasek:
This is odd, APIv2 tests works for me.

matejvasek:
Some tests seems to fail randomly.

vrothberg (commented):
/hold

Please squash/rename the commits: we should not merge "touch" or "[DO NOT MERGE]" titled commits :)

matejvasek:
We will also need to backport https://github.com/containers/podman/pull/9203

matejvasek:
@rhatdan @mheon PTAL

matejvasek:
@riyad I think you know how to run some docker conformance tests, could you please run it against my `/containers/{id}/wait`, or give me some pointers how to do it myself?

pkg/bindings/containers/types.go
```diff
@@ -176,7 +176,7 @@ type UnpauseOptions struct{}
 //go:generate go run ../generator/generator.go WaitOptions
 // WaitOptions are optional options for waiting on containers
 type WaitOptions struct {
-	Condition *define.ContainerStatus
+	Condition []define.ContainerStatus
```
matejvasek, line 179: I deliberately choose not to change this to plural for preserving some degree of compatibility.
With this older client can call new server safely. Similarly new client can call older server, but only as long as the contains at most one element.

matejvasek: If I understand generator correctly uri param is derived just from member name. Maybe we can introduce some custom  tag. So it could look like:
```go
type WaitOptions struct {
	Conditions []define.ContainerStatus `uriparam:"condtion"`
	Interval   *string                  `uriparam:"interval"`
}
```
Meaning the name is changed, but on wire it's the same.

libpod/container_api.go
```diff
@@ -478,13 +479,15 @@ func (c *Container) RemoveArtifact(name string) error {
 }
 
 // Wait blocks until the container exits and returns its exit code.
-func (c *Container) Wait() (int32, error) {
-	return c.WaitWithInterval(DefaultWaitInterval)
+func (c *Container) Wait(ctx context.Context) (int32, error) {
```
matejvasek, line 482: I made the wait functions cancel-able. Useful for not leaking gorutines when client cancels request (closes connection abruptly). For CLI podman it's not used since context passed here is not canceled on SIGTERM or SIGINT. CLI chooses IMO rather brutal `os.Exit()`.

libpod/container_api.go
```diff
@@ -478,13 +479,15 @@ func (c *Container) RemoveArtifact(name string) error {
...
+	return c.WaitWithInterval(ctx, DefaultWaitInterval)
 }
 
+var errWaitingCanceled = errors.New("waiting was canceled")
```
mheon, line 486: Might be worth making this a bit more generic (`ErrCanceled = errors.New("cancelled by user")`) and putting it in `define/errors.go` for other parts of Podman to use

matejvasek: or maybe better yet use `context.Canceled`

libpod/container_api.go
```diff
@@ -495,41 +498,111 @@ func (c *Container) WaitWithInterval(waitTimeout time.Duration) (int32, error) {
 	}
 	chWait := make(chan error, 1)
 
-	defer close(chWait)
```
mheon, line 498: It seems like this could still be useful, given that `ctx.Done()` is a goroutine and not guaranteed to fire at the end of the function, but rather when the entire handler returns, AFAIK

libpod/container_api.go
```diff
@@ -495,41 +498,111 @@ func (c *Container) WaitWithInterval(waitTimeout time.Duration) (int32, error) {
...
+		case resultChan <- waitResult{code, err}:
+		case <-ctx.Done():
 		}
-		time.Sleep(waitTimeout)
 	}
-	return -1, nil
+
+	var wg sync.WaitGroup
+
+	if waitForExit {
```
mheon, line 562: It seems like both this and the `len(wantedStates) > 0` below can both be active at the same time - it looks like we don't need that, though, that goroutine seems to take care of the same cases this one does. Should we only do this `waitForExit` bit if the user is *only* waiting for `Stopped` and `Exited`, and return immediately after?

mheon:
Done reviewing. @rhatdan @jwhonce PTAL

libpod/container_api.go
matejvasek: Probably shouldn't hurt, but also it's not necessary to close a channel here.

libpod/container_api.go
matejvasek: As of now in practice only one is active. I future they might be both active. Waiting for terminal state differs from non-terminal. For instance if user condition is `[Stopped, Removing]`, both gorutines have to be active. One is waiting for `Stopped` and the other one is waiting for `Removing`.

matejvasek: The `len(wantedStates) > 0` gorutine is not taking care of non-terminal states.

mheon:

rhatdan:
Great work @matejvasek Thanks.
/lgtm

PR #9051, opened 2021-01-21
Switch podman stop/kill/wait handlers to use abi

rhatdan:
Change API Handlers to use the same functions that the
local podman uses.

At the same time:

 implement remote API for --all and --ignore flags for podman stop
 implement remote API for --all flags for podman stop

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -31,11 +31,11 @@ import (
 func RemoveContainer(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
-		All     bool `schema:"all"`
-		Force   bool `schema:"force"`
-		Ignore  bool `schema:"ignore"`
-		Link    bool `schema:"link"`
-		Volumes bool `schema:"v"`
+		Force         bool `schema:"force"`
+		Ignore        bool `schema:"ignore"`
+		Link          bool `schema:"link"`
+		DockerVolumes bool `schema:"v"`
+		LibpodVolumes bool `schema:"volumes"`
 	}{
 		// override any golang type defaults
 	}
@@ -46,10 +46,19 @@ func RemoveContainer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if query.Link && !utils.IsLibpodRequest(r) {
-		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
-			utils.ErrLinkNotSupport)
-		return
+	options := entities.RmOptions{
+		Force:  query.Force,
+		Ignore: query.Ignore,
+	}
+	if utils.IsLibpodRequest(r) {
+		options.Volumes = query.LibpodVolumes
+	} else {
+		if query.Link {
+			utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+				utils.ErrLinkNotSupport)
+			return
+		}
+		options.Volumes = query.DockerVolumes
 	}
 
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
@@ -57,12 +66,6 @@ func RemoveContainer(w http.ResponseWriter, r *http.Request) {
 	// code.
 	containerEngine := abi.ContainerEngine{Libpod: runtime}
 	name := utils.GetName(r)
-	options := entities.RmOptions{
-		All:     query.All,
-		Force:   query.Force,
-		Volumes: query.Volumes,
-		Ignore:  query.Ignore,
-	}
 	report, err := containerEngine.ContainerRm(r.Context(), []string{name}, options)
 	if err != nil {
 		if errors.Cause(err) == define.ErrNoSuchCtr {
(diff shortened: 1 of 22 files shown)


rhatdan:
@edsantiago more missing podman-remote options and missing tests.

vrothberg (commented):
LGTM

vrothberg:
`make generate-bindings` isn't happy in CI

vrothberg:
Get the same locally after `make generate-bindings`:
```diff
podman (rm) $ git diff
diff --git a/pkg/bindings/containers/types_stop_options.go b/pkg/bindings/containers/types_stop_options.go
index c86cc2f9ffa9..d4e2eda8383c 100644
--- a/pkg/bindings/containers/types_stop_options.go
+++ b/pkg/bindings/containers/types_stop_options.go
@@ -119,7 +119,7 @@ func (o *StopOptions) GetIgnore() bool {
        return *o.Ignore
 }

-// WithToimeout
+// WithTimeout
 func (o *StopOptions) WithTimeout(value uint) *StopOptions {
        v := &value
```

pkg/api/handlers/compat/containers.go
```diff
@@ -184,55 +182,38 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
...
-		return
-	}
+		if errors.Cause(err) == define.ErrNoSuchCtr {
+			utils.ContainerNotFound(w, name, err)
+			return
+		}
 
-	// If the Container is stopped already, send a 409
-	if state == define.ContainerStateStopped || state == define.ContainerStateExited {
-		utils.Error(w, fmt.Sprintf("Container %s is not running", name), http.StatusConflict, errors.New(fmt.Sprintf("Cannot kill Container %s, it is not running", name)))
```
mheon, line 216: We need to retain some form of this check so we can return `StatusConflict`

mheon: I think you can check if ContainerKill returned either `ErrCtrStateInvalid` or `ErrCtrStopped` though

pkg/api/handlers/compat/containers.go
```diff
@@ -184,55 +182,38 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrapf(err, "unable to kill Container %s", name))
+	if report[0].Err != nil {
+		utils.InternalServerError(w, report[0].Err)
 		return
 	}
 
-	// Docker waits for the container to stop if the signal is 0 or
-	// SIGKILL.
-	if !utils.IsLibpodRequest(r) && (signal == 0 || syscall.Signal(signal) == syscall.SIGKILL) {
```
mheon, line 230: I think we need to retain this.

pkg/api/handlers/compat/containers_stop.go
```diff
@@ -27,31 +34,25 @@ func StopContainer(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		utils.InternalServerError(w, errors.Wrapf(err, "unable to get state for Container %s", name))
-		return
+	options := entities.StopOptions{
+		All:     query.All,
+		Ignore:  query.Ignore,
+		Timeout: &query.Timeout,
 	}
-	// If the Container is stopped already, send a 304
-	if state == define.ContainerStateStopped || state == define.ContainerStateExited {
```
mheon, line 42: Need to retain this check so we can return `StatusNotModified`

pkg/api/handlers/compat/containers.go
```diff
@@ -193,46 +197,34 @@ func KillContainer(w http.ResponseWriter, r *http.Request) {
...
+		utils.InternalServerError(w, err)
 		return
 	}
 
-	signal := uint(sig)
-
-	err = con.Kill(signal)
-	if err != nil {
-		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrapf(err, "unable to kill Container %s", name))
+	if report[0].Err != nil {
```
mheon, line 223: We can't do this in cases where more than one result can be returned (e.g. All is specified). You are going to lose errors on everything except the first container, which is *HORRIBLE* user experience - we've had to deal with this in Pod commands and `play kube` and it makes debugging user issues impossible.

This is going to require a change in return type to return a struct containing all the errors, which will make it incompatible with Docker. I suppose we could use the magic check that determines if this is a Libpod or Compat request to handle it?

mheon: (As an aside, this is why we decided to handle `All` client side, not server side, initially - it makes things so much simpler because we can re-use handlers and not worry about this).

rhatdan: Removed all, although I think  in the long run this should be changed.

pkg/api/handlers/compat/containers.go
rhatdan: Fixed

pkg/api/handlers/compat/containers_stop.go
rhatdan: Yuck, Fixed.

pkg/api/handlers/compat/containers.go
```diff
@@ -242,6 +248,10 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) {
 	// /{version}/containers/(name)/wait
 	exitCode, err := utils.WaitContainer(w, r)
 	if err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
```
mheon, line 251: Is this necessary? Waiting on a container that has been removed by `--rm` will error now

pkg/api/handlers/compat/containers_restart.go
```diff
@@ -4,42 +4,60 @@ import (
...
-	if err != nil {
-		utils.ContainerNotFound(w, name, err)
-		return
-	}
 
-	timeout := con.StopTimeout()
-	if _, found := r.URL.Query()["t"]; found {
-		timeout = uint(query.Timeout)
+	options := entities.RestartOptions{
+		All:     query.All,
```
mheon, line 39: All is still hazardous to use, we're going to lose every error except the first one.

pkg/api/handlers/libpod/containers.go
```diff
@@ -148,6 +148,11 @@ func GetContainer(w http.ResponseWriter, r *http.Request) {
 func WaitContainer(w http.ResponseWriter, r *http.Request) {
 	exitCode, err := utils.WaitContainer(w, r)
 	if err != nil {
+		if errors.Cause(err) == define.ErrNoSuchCtr {
+			utils.ContainerNotFound(w, utils.GetName(r), err)
```
mheon, line 152: Hm. Should we error here? My concern is `--rm` - we could race against cleanup and error on a container being removed.

pkg/api/handlers/utils/containers.go
```diff
@@ -40,19 +49,15 @@ func WaitContainer(w http.ResponseWriter, r *http.Request) (int32, error) {
...
-	name := GetName(r)
-	con, err := runtime.LookupContainer(name)
+
+	report, err := containerEngine.ContainerWait(r.Context(), []string{name}, options)
 	if err != nil {
-		ContainerNotFound(w, name, err)
 		return 0, err
 	}
-	return con.WaitForConditionWithInterval(interval, condition)
+	return report[0].ExitCode, report[0].Error
```
mheon, line 62: Should double-check len(report) > 0 - better safe than segfaulting

pkg/api/handlers/libpod/containers.go
rhatdan: So just log it?

pkg/api/handlers/compat/containers_restart.go
rhatdan: I hope to fix this in the future.  But it will be a breaking change in the API.

pkg/api/handlers/compat/containers.go
rhatdan: Do you think we should at least log it. Or just ignore it.

pkg/api/handlers/libpod/containers.go
mheon: Yeah, logging seems right. Maybe someone will complain in the future and we'll have to do better, but for now this seems safest

rhatdan:

vrothberg (commented):
LGTM

Issue #9040, opened 2021-01-20
Podman does not create working directory from image


tvass:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman does not create working directory from image `registry.svc.ci.openshift.org/openshift/release:golang-1.15`.

**Steps to reproduce the issue:**

1.  Podman run the image `registry.svc.ci.openshift.org/openshift/release:golang-1.15`

```
$ podman run -it --rm registry.svc.ci.openshift.org/openshift/release:golang-1.15 /bin/bash
Error: chdir: No such file or directory: OCI not found
```
**Describe the results you received:**

The container is NOT run.

**Describe the results you expected:**

We expect to have bash executed inside the container.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

fc33 

```
[~]$ podman --version
podman version 2.2.1

```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 0f53fb68333bdead5fe4dc5175703e22cf9882ab'
  cpus: 8
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.9.16-200.fc33.x86_64
  linkmode: dynamic
  memFree: 2222854144
  memTotal: 16415924224
  ociRuntime:
    name: crun
    package: crun-0.16-3.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.16
      commit: eb0145e5ad4d8207e84a327248af76663d4e50dd
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    path: /run/user/1000/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
[...]

PR #9054, opened 2021-01-21
make sure the workdir exists on container mount

vrothberg:
A container's workdir can be specified via the CLI via `--workdir` and   
via an image config with the CLI having precedence.                      
                                                                         
Since images have a tendency to specify workdirs without necessarily     
shipping the paths with the root FS, make sure that Podman creates the   
workdir.  When specified via the CLI, do not create the path, but check  
for its existence and return a human-friendly error.                     
                                                                         

Fixes: #9040

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -776,3 +776,32 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
 	}
 	return c.shouldRestart()
 }
+
+// ResolvePath resolves the specified path on the root for the container.  The
+// root must either be the mounted image of the container or the already
+// mounted container storage.
+//
+// It returns the resolved root and the resolved path.  Note that the path may
+// resolve to the container's mount point or to a volume or bind mount.
+func (c *Container) ResolvePath(ctx context.Context, root string, path string) (string, string, error) {
+	logrus.Debugf("Resolving path %q (root %q) on container %s", path, root, c.ID())
+
+	// Minimal sanity checks.
+	if len(root)*len(path) == 0 {
+		return "", "", errors.Wrapf(define.ErrInternal, "ResolvePath: root (%q) and path (%q) must be non empty", root, path)
+	}
+	if _, err := os.Stat(root); err != nil {
+		return "", "", errors.Wrapf(err, "cannot locate root to resolve path on container %s", c.ID())
+	}
+
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return "", "", err
+		}
+	}
+
+	return c.resolvePath(root, path)
+}
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -174,25 +174,60 @@ func (c *Container) prepare() error {
 		return err
 	}
 
-	// Ensure container entrypoint is created (if required)
-	if c.config.CreateWorkingDir {
-		workdir, err := securejoin.SecureJoin(c.state.Mountpoint, c.WorkingDir())
-		if err != nil {
-			return errors.Wrapf(err, "error creating path to container %s working dir", c.ID())
-		}
-		rootUID := c.RootUID()
-		rootGID := c.RootGID()
+	// Make sure the workdir exists
+	if err := c.resolveWorkDir(); err != nil {
+		return err
+	}
 
-		if err := os.MkdirAll(workdir, 0755); err != nil {
-			if os.IsExist(err) {
-				return nil
+	return nil
+}
+
+// resolveWorkDir resolves the container's workdir and, depending on the
+// configuration, will create it, or error out if it does not exist.
+// Note that the container must be mounted before.
+func (c *Container) resolveWorkDir() error {
+	workdir := c.WorkingDir()
+
+	// If the specified workdir is a subdir of a volume or mount,
+	// we don't need to do anything.  The runtime is taking care of
+	// that.
+	if isPathOnVolume(c, workdir) || isPathOnBindMount(c, workdir) {
+		logrus.Debugf("Workdir %q resolved to a volume or mount", workdir)
+		return nil
+	}
+
+	_, resolvedWorkdir, err := c.resolvePath(c.state.Mountpoint, workdir)
+	if err != nil {
+		return err
+	}
+	logrus.Debugf("Workdir %q resolved to host path %q", workdir, resolvedWorkdir)
+
+	// No need to create it (e.g., `--workdir=/foo`), so let's make sure
+	// the path exists on the container.
+	if !c.config.CreateWorkingDir {
+		if _, err := os.Stat(resolvedWorkdir); err != nil {
+			if os.IsNotExist(err) {
+				return errors.Errorf("workdir %q does not exist on container %s", workdir, c.ID())
 			}
-			return errors.Wrapf(err, "error creating container %s working dir", c.ID())
+			// This might be a serious error (e.g., permission), so
+			// we need to return the full error.
+			return errors.Wrapf(err, "error detecting workdir %q on container %s", workdir, c.ID())
 		}
+	}
 
-		if err := os.Chown(workdir, rootUID, rootGID); err != nil {
-			return errors.Wrapf(err, "error chowning container %s working directory to container root", c.ID())
+	// Ensure container entrypoint is created (if required).
+	rootUID := c.RootUID()
+	rootGID := c.RootGID()
+
+	if err := os.MkdirAll(resolvedWorkdir, 0755); err != nil {
+		if os.IsExist(err) {
+			return nil
 		}
+		return errors.Wrapf(err, "error creating container %s workdir", c.ID())
+	}
+
+	if err := os.Chown(resolvedWorkdir, rootUID, rootGID); err != nil {
+		return errors.Wrapf(err, "error chowning container %s workdir to container root", c.ID())
 	}
 
 	return nil
--- a/libpod/container_path_resolution.go
+++ b/libpod/container_path_resolution.go
@@ -13,6 +13,11 @@ import (
 // resolveContainerPaths resolves the container's mount point and the container
 // path as specified by the user.  Both may resolve to paths outside of the
 // container's mount point when the container path hits a volume or bind mount.
+//
+// It returns a bool, indicating whether containerPath resolves outside of
+// mountPoint (e.g., via a mount or volume), the resolved root (e.g., container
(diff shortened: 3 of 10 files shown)


vrothberg:
@mheon PTAL

vrothberg:
@edsantiago PTAL

test/system/030-run.bats
```diff
@@ -589,4 +589,22 @@ json-file | f
     is "${lines[1]}" "$rand" "Container runs successfully despite warning"
 }
 
+# Regression test for #9040
+@test "podman run - implicitly create workdir" {
+    workdir=/$(random_string 10)
+    PODMAN_TIMEOUT=240 run_podman build -t workdir_test - << EOF
+FROM  $IMAGE
+WORKDIR $workdir
+EOF
```
edsantiago, line 598: What would you think of this instead?
```diff
diff --git a/test/system/070-build.bats b/test/system/070-build.bats
index 0e83a184b..c81918428 100644
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -148,7 +148,6 @@ EOF
     cat >$tmpdir/Containerfile <<EOF
 FROM $IMAGE
 LABEL $label_name=$label_value
-RUN mkdir $workdir
 WORKDIR $workdir

 # Test for #7094 - chowning of invalid symlinks
```

edsantiago (commented):
LGTM, with one suggestion. I'm always trying to encourage people to piggyback on existing tests. Every second we save is valuable, and `podman build` takes a long long time.

edsantiago:
Thinking back... wasn't there a good reason why podman decided not to implicitly create workdirs? @mheon @rhatdan @baude ?

test/system/030-run.bats
vrothberg: I copied the test from there and thought about it but thought you'd prefer to have it in `030-run` :smile:   I can certainly change the build test but keep the `--workdir=...` here :+1:

mheon:
@edsantiago Logic was to catch user error - a workdir not specified by the image, that does not exist, is probably a typo.

edsantiago:
Thanks @mheon. I remember that now.

Whichever way we go, we *need* to embed that decision in the code. If we go this route, please document it. If we reject this PR, I will add a system test that explicitly checks for failure and explains the reason in a test comment.

vrothberg:
Notes from the meeting: Consensus was on only creating the workdir if specified by the image.  When specified by the user, Podman should error out if it does not exist.

vrothberg:
Updated. PTanotherL

vrothberg:
Integration failure looks legit.  I will have a look tomorrow morning.

libpod/container_internal_linux.go
```diff
@@ -174,24 +174,35 @@ func (c *Container) prepare() error {
 		return err
 	}
 
-	// Ensure container entrypoint is created (if required)
+	workdir := c.WorkingDir()
+	resolvedWorkdir, err := securejoin.SecureJoin(c.state.Mountpoint, c.WorkingDir())
```
giuseppe, line 178: how would it work with volumes?

if we have something like `podman run -v /foo:/foo -w /foo/bar ...` the working dir must be looked up in the volume mount

vrothberg: Very good point.

To make this work, we can reuse the copy code in `pkg/domain/infra/abi/containers_stat.go` and move it into `libpod`.  But I think that should be a separate PR.

@mheon WDYT?

rhatdan:
Just out of curiosity, could you try this with a --read-only container?

vrothberg:
> Just out of curiosity, could you try this with a --read-only container?
```
# With --workdir
podman (fix-9040) $ ./bin/podman run  --rm --workdir=/foo/bar --read-only alpine pwd
Error: workdir "/foo/bar" does not exist on container a94d26e8b4fe7d29ca49395b4c9ff0edcdfa25010257642f749530b92d9a4e4d
# With workdir from the image
podman (fix-9040) $ ./bin/podman run  --rm --read-only test pwd
/i/do/not/exist
```

libpod/container_internal_linux.go
vrothberg: Actually, we need to tackle this now to make system tests happy.

test/system/030-run.bats
giuseppe: I think the tests are failing because the working dir is on a volume, so the check fails to detect it

rhatdan:
Excellent.

libpod/container_internal_linux.go
mheon: Would your intention be to replace the copy-up code wholesale, or just move the bit that determines where the path in a container lives?

libpod/container_internal_linux.go
```diff
@@ -174,24 +174,35 @@ func (c *Container) prepare() error {
...
 		rootUID := c.RootUID()
 		rootGID := c.RootGID()
 
-		if err := os.MkdirAll(workdir, 0755); err != nil {
+		if err := os.MkdirAll(resolvedWorkdir, 0755); err != nil {
 			if os.IsExist(err) {
 				return nil
 			}
-			return errors.Wrapf(err, "error creating container %s working dir", c.ID())
+			return errors.Wrapf(err, "error creating container %s workdir", c.ID())
```
TomSweeneyRedHat, line 191: It would probably be useful to include the value in resolveWorkdir in the error too.

libpod/container_internal_linux.go
```diff
@@ -174,24 +174,35 @@ func (c *Container) prepare() error {
...
-		if err := os.Chown(workdir, rootUID, rootGID); err != nil {
-			return errors.Wrapf(err, "error chowning container %s working directory to container root", c.ID())
+		if err := os.Chown(resolvedWorkdir, rootUID, rootGID); err != nil {
+			return errors.Wrapf(err, "error chowning container %s workdir to container root", c.ID())
+		}
+	} else {
+		// Otherwise, make sure the workdir is present.
+		if _, err := os.Stat(resolvedWorkdir); err != nil {
+			if os.IsNotExist(err) {
+				return errors.Errorf("workdir %q does not exist on container %s", workdir, c.ID())
```
TomSweeneyRedHat, line 201:
```suggestion
				return errors.Errorf("workdir %q does not exist on container %s", resolvedWorkdir, c.ID())
```

libpod/container_internal_linux.go
```diff
@@ -174,24 +174,35 @@ func (c *Container) prepare() error {
...
+			}
+			// This might be a serious error (e.g., permission), so
+			// we need to return the full error.
+			return errors.Wrapf(err, "error detecting workdir %q on container %s", workdir, c.ID())
```
TomSweeneyRedHat, line 205:
```suggestion
			return errors.Wrapf(err, "error detecting workdir %q on container %s", resolvedWorkdir, c.ID())
```

libpod/container_internal_linux.go
vrothberg: I don't think the resolved one is helpful.  That'll be the fully resolved path on the host which I think can easily confuse users.  I found nicer to return the erroneous input so users can match the error to their input.

TomSweeneyRedHat: Ignore this and the next if you wish.  I saw the workdir initialization removed from the old code, but missed it being added at the very top in this new change.  I'll trust you on which is the better to present here to the user and down below.

libpod/container_internal_linux.go
vrothberg: Just the latter.

vrothberg:
Merge me :) (assuming I am pleasing)

libpod/container_api.go
```diff
@@ -776,3 +776,21 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
 	}
 	return c.shouldRestart()
 }
+
+// ResolvePath resolves the specified path on the container.  The path may
+// resolve to the container's mount point or to a volume or bind mount.  It
+// returns the root (e.g., container, volume or bind mount) and the fully
+// resolved path which is either equal to or a subdirectory of the root.
+func (c *Container) ResolvePath(ctx context.Context, path string) (string, string, error) {
```
mheon, line 784: This should do the standard lock + sync if it's a public function

vrothberg: It's safe to use concurrently/

vrothberg: Or shall I lock and then `mount` and `unmount`?

mheon: My concern isn't really concurrency so much as removed containers - we could have the container removed out from under us midway through this, between Mount and Unmount, for example. Preference would be lock then `mount()` / `unmount()`

mheon: Although, looking at it more, I think a lot of the logic in `Unmount()` isn't duplicated in `unmount()`... I don't think it can be safely moved in there, either (we have two cases for unmounting - cases like this, where an external user wants to access the container's root filesystem - the code in Unmount() works there - and cases where Libpod wants to stop the container - the code in Unmount works there).

@vrothberg What would you think about making `ResolvePath()` be contingent on the container already being mounted? It seems like it would simplify the code, and potentially make things a bit more performant (we wouldn't need to call Mount more than once in the Copy code).

vrothberg: @mheon sounds good to me.  One early version of the change was exactly like that but I preferred to play it safe.  I'll go as suggested and repush.

vrothberg:
Ready to go

vrothberg:
Marking as backport for 3.0. Please raise your voice in case you have concerns.

rhatdan:
I want @mheon to approve and merge.

libpod/container_path_resolution.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// findBindMounts checks if the specified containerPath matches the destination
+// path of a Mount.  Returns a matching Mount or nil.
+func findBindMount(c *Container, containerPath string) *specs.Mount {
+	cleanedPath := filepath.Clean(containerPath)
+	for _, m := range c.Config().Spec.Mounts {
+		if m.Type != "bind" {
+			continue
+		}
```
giuseppe, line 142: to check for a bind mount, we'd need to look for `bind` or `rbind` in the options, the type is ignored by the runtime

mheon: I think we require that Type also be set for Libpod - at the very least, all of our existing code is guaranteed to do so.

vrothberg: Currently, i's detecting correctly when using `--mount=type=bind` etc.  It's pretty the code from `cp`.

If there's something missing, let's fix that in a separate PR.

mheon:
Oops, I think I was the only LGTM... Need one more to merge

PR #9055, opened 2021-01-21
[3.0] Set log driver for compatability containers

baude:
when using the compatibility api to create containers, now reflect the
use of k8s-file as json-file so that clients, which are
unaware of k8s-file, can work.  specifically, if the container is using
k8s-file as the log driver, we change the log type in container
inspection to json-file.  These terms are used interchangably in other
locations in libpod/podman.

this fixes log messages in compose as well.

[NO TESTS NEEDED]


vrothberg (commented):
LGTM

mheon:
Restarted one flake
/hold
/lgtm

PR #9062, opened 2021-01-21
[v3.0] [CI:DOCS] Cirrus: Fix running Validate task on branches

cevich:
Backport from master https://github.com/containers/podman/pull/9063

Followup to 4c60523

For whatever reason Cirrus-CI does not set `$CIRRUS_BASE_SHA` when
testing on a branch or tag.  If the branch happened to also not have
any successful previous runs, then `$EPOCH_TEST_COMMIT` was being set
to the error-flagging value `YOU_FOUND_A_BUG`.

Fix this by detecting an empty value when running on a branch or via
`hack/get_ci_vm.sh` (`$CIRRUS_TAG` will be empty) and setting
`$CIRRUS_BASE_SHA` based on the upstream branch HEAD.  Otherwise,
instead of using `YOU_FOUND_A_BUG`, simply reference the HEAD commit.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #9063, opened 2021-01-21
Cirrus: Fix running Validate task on branches

cevich:
Similar to v3.0 branch PR https://github.com/containers/podman/pull/9062

Followup to dbb9943

Despite skipping the `Smoke` check, it was observed on a *new* branch,
the `validate` task (specifically `git-validation`) will fail.  This
is because:

* `$CIRRUS_LAST_GREEN_CHANGE` will be empty on a new branch.
* `$CIRRUS_BASE_SHA` is always empty for runs triggered by branch-push
* `$EPOCH_TEST_COMMIT` will be set to `YOU_FOUND_A_BUG`.

Fix this by eliminating the `Smoke` task entirely, simplifying all
the `make validate` operations into the `validate` cirrus task.

Also, eliminate the `$CIRRUS_BUILD_ID` value as it's confusing and not
actually used anywhere.  It was formerly used for building VM images,
but this has moved to another repo entirely.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -46,7 +46,7 @@ env:
     #### Control variables that determine what to run and how to run it.
     #### N/B: Required ALL of these are set for every single task.
     ####
-    TEST_FLAVOR:             # int, sys, ext_svc, smoke, automation, etc.
+    TEST_FLAVOR:             # int, sys, ext_svc, validate, automation, etc.
     TEST_ENVIRON: host       # 'host' or 'container'
     PODBIN_NAME: podman      # 'podman' or 'remote'
     PRIV_NAME: root          # 'root' or 'rootless'
@@ -78,6 +78,25 @@ ext_svc_check_task:
     env:
         TEST_FLAVOR: ext_svc
         CTR_FQIN: ${FEDORA_CONTAINER_FQIN}
+    # NOTE: The default way Cirrus-CI clones is *NOT* compatible with
+    #       environment expectations in contrib/cirrus/lib.sh.  Specifically
+    #       the 'origin' remote must be defined, and all remote branches/tags
+    #       must be available for reference from CI scripts.
+    clone_script: &full_clone |
+          cd /
+          rm -rf $CIRRUS_WORKING_DIR
+          mkdir -p $CIRRUS_WORKING_DIR
+          git clone --recursive --branch=$DEST_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+          cd $CIRRUS_WORKING_DIR
+          git remote update origin
+          if [[ -n "$CIRRUS_PR" ]]; then # running for a PR
+              git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
+              git checkout pull/$CIRRUS_PR
+          else
+              git reset --hard $CIRRUS_CHANGE_IN_REPO
+          fi
+          make install.tools
+
     setup_script: &setup '$GOSRC/$SCRIPT_BASE/setup_environment.sh'
     main_script: &main '/usr/bin/time --verbose --output="$STATS_LOGFILE" $GOSRC/$SCRIPT_BASE/runner.sh'
     always: &runner_stats
@@ -97,43 +116,7 @@ automation_task:
         TEST_FLAVOR: automation
         CTR_FQIN: ${FEDORA_CONTAINER_FQIN}
         TEST_ENVIRON: container
-    setup_script: *setup
-    main_script: *main
-    always: *runner_stats
-
-
-# This task use to be called 'gating', however that name is being
-# used downstream for release testing.  Renamed this to avoid confusion.
-# All it does is run basic golang formatting and commit validation checks.
-smoke_task:
-    alias: 'smoke'
-    name: "Smoke Test"
-    skip: *branches_and_tags
-    container: &bigcontainer
-        image: ${CTR_FQIN}
-        # Leave some resources for smallcontainer
-        cpu: 6
-        memory: 22
-    env:
-        TEST_FLAVOR: 'smoke'
-        CTR_FQIN: "${FEDORA_CONTAINER_FQIN}"
-        TEST_ENVIRON: container
-    # This clone script is also used to initially populate gopath_cache (below)
-    clone_script: &full_clone |
-        cd /
-        rm -rf $CIRRUS_WORKING_DIR
-        mkdir -p $CIRRUS_WORKING_DIR
-        git clone --recursive --branch=$DEST_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
-        cd $CIRRUS_WORKING_DIR
-        git remote update origin
-        if [[ -n "$CIRRUS_PR" ]]; then # running for a PR
-            git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
-            git checkout pull/$CIRRUS_PR
-        else
-            git reset --hard $CIRRUS_CHANGE_IN_REPO
-        fi
-        cd $CIRRUS_WORKING_DIR
-        make install.tools
+    clone_script: *full_clone
     setup_script: *setup
     main_script: *main
     always: *runner_stats
@@ -211,11 +194,15 @@ build_task:
 validate_task:
     name: "Validate $DISTRO_NV Build"
     alias: validate
-    skip: *tags
+    # This task is primarily intended to catch human-errors early on, in a
+    # PR.  Skip it for branch-push, branch-create, and tag-push to improve
+    # automation reliability/speed in those contexts.  Any missed errors due
+    # to nonsequential PR merging practices, will be caught on a future PR,
+    # build or test task failures.
+    skip: *branches_and_tags
     depends_on:
         - ext_svc_check
         - automation
-        - smoke
         - build
     # golangci-lint is a very, very hungry beast.
     gce_instance: &bigvm
@@ -646,7 +633,6 @@ success_task:
     depends_on:
         - ext_svc_check
         - automation
-        - smoke
         - build
         - validate
         - bindings
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -80,10 +80,19 @@ CIRRUS_CI="${CIRRUS_CI:-false}"
 DEST_BRANCH="${DEST_BRANCH:-master}"
 CONTINUOUS_INTEGRATION="${CONTINUOUS_INTEGRATION:-false}"
 CIRRUS_REPO_NAME=${CIRRUS_REPO_NAME:-podman}
-# N/B: CIRRUS_BASE_SHA is empty on branch and tag push.
-CIRRUS_BASE_SHA=${CIRRUS_BASE_SHA:-${CIRRUS_LAST_GREEN_CHANGE:-YOU_FOUND_A_BUG}}
-CIRRUS_BUILD_ID=${CIRRUS_BUILD_ID:-$RANDOM$(date +%s)}  # must be short and unique
-
+# Cirrus only sets $CIRRUS_BASE_SHA properly for PRs, but $EPOCH_TEST_COMMIT
+# needs to be set from this value in order for `make validate` to run properly.
+# When running get_ci_vm.sh, most $CIRRUS_xyz variables are empty. Attempt
+# to accomidate both branch and get_ci_vm.sh testing by discovering the base
+# branch SHA value.
+# shellcheck disable=SC2154
+if [[ -z "$CIRRUS_BASE_SHA" ]] && [[ -z "$CIRRUS_TAG" ]]
+then  # Operating on a branch, or under `get_ci_vm.sh`
+    CIRRUS_BASE_SHA=$(git rev-parse ${UPSTREAM_REMOTE:-origin}/$DEST_BRANCH)
+elif [[ -z "$CIRRUS_BASE_SHA" ]]
+then  # Operating on a tag
+    CIRRUS_BASE_SHA=$(git rev-parse HEAD)
+fi
 # The starting place for linting and code validation
 EPOCH_TEST_COMMIT="$CIRRUS_BASE_SHA"
 
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -23,22 +23,6 @@ function _run_ext_svc() {
     $SCRIPT_BASE/ext_svc_check.sh
 }
 
-function _run_smoke() {
-    make gofmt
-
-    # There is little value to validating commits after tag-push
-    # and it's very difficult to automatically determine a starting commit.
-    # $CIRRUS_TAG is only non-empty when executing due to a tag-push
-    # shellcheck disable=SC2154
-    if [[ -z "$CIRRUS_TAG" ]]; then
-        # If PR consists of multiple commits, test that each compiles cleanly
-        make .gitvalidation
-
-        # PRs should include some way to test.
-        $SCRIPT_BASE/pr-should-include-tests
-    fi
-}
-
 function _run_automation() {
     $SCRIPT_BASE/cirrus_yaml_test.py
 
@@ -51,11 +35,14 @@ function _run_automation() {
 }
 
 function _run_validate() {
-    # Confirm compile via prior task + cache
-    bin/podman --version
-    bin/podman-remote --version
+    # git-validation tool fails if $EPOCH_TEST_COMMIT is empty
+    # shellcheck disable=SC2154
+    if [[ -n "$EPOCH_TEST_COMMIT" ]]; then
+        make validate
+    else
+        warn "Skipping git-validation since \$EPOCH_TEST_COMMIT is empty"
+    fi
 
-    make validate  # Some items require a build
 }
 
 function _run_unit() {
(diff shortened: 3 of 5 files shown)


rhatdan:
LGTM
@edsantiago PTAL

contrib/cirrus/runner.sh
```diff
@@ -24,18 +24,15 @@ function _run_ext_svc() {
...
-    # $CIRRUS_TAG is only non-empty when executing due to a tag-push
+    # git-validation tool fails if $EPOCH_TEST_COMMIT is empty
     # shellcheck disable=SC2154
-    if [[ -z "$CIRRUS_TAG" ]]; then
-        # If PR consists of multiple commits, test that each compiles cleanly
+    if [[ -n "$EPOCH_TEST_COMMIT" ]]; then
         make .gitvalidation
-
-        # PRs should include some way to test.
-        $SCRIPT_BASE/pr-should-include-tests
```
edsantiago, line 38: Why is this being removed?

contrib/cirrus/runner.sh
```diff
@@ -24,18 +24,15 @@ function _run_ext_svc() {
 }
 
 function _run_smoke() {
-    make gofmt
+    # Quick checks which do not require anything to be compiled
+    make gofmt lint validate.completions swagger-check tests-included
```
edsantiago, line 28: `make tests-included` should only be run on PRs. I realize there's a new `skip smoke` in `.cirrus.yml` but it might be helpful to include a loud comment here explaining why this is meaningful only in PRs and why it should be skipped on tags/branches.

contrib/cirrus/runner.sh
```diff
@@ -55,7 +52,8 @@ function _run_validate() {
     bin/podman --version
     bin/podman-remote --version
 
-    make validate  # Some items require a build
+    # Confirm --help and docs match (requires compiled binaries)
+    make man-page-check
```
edsantiago, line 56: Oh, no. This is a disaster waiting to happen. Please leave this as it was. It's perfectly fine to remove the `Quick checks` step above. This should not be committed under any circumstances as it is. (I really hope I don't need to spend time explaining why).

edsantiago (changes requested):
Some requests for explanations, and one absolute showstopper.

contrib/cirrus/lib.sh
```diff
@@ -80,13 +80,27 @@ CIRRUS_CI="${CIRRUS_CI:-false}"
...
+# N/B: Cirrus sets $CIRRUS_BASE_SHA properly for PRs, empty otherwise.
+# Discover the value when running on a branch or in get_ci_vm.sh
+# shellcheck disable=SC2154
+if [[ -z "$CIRRUS_BASE_SHA" ]] && [[ -z "$CIRRUS_TAG" ]]
+then
+    if git remote show | grep -q upstream; then
+        _remote="upstream"
+    else
+        _remote="origin"
+    fi
```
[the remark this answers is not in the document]
cevich: I intended the comment above to cover this: `# Discover the value when running on a branch or in`***`get_ci_vm.sh`***

For `get_ci_vm.sh`, most runtime `$CIRRUS_xyz` variables are not set automatically, they have to be discovered at runtime or through `lib.sh` defaults (like `$CIRRUS_BRANCH=$DEST_BRANCH`).  As a service to users, the example hook script renames "origin" to "upstream" before cloning their fork as "origin".  Outside of that context, "origin" will always be correct given [how the clone is setup](https://github.com/containers/podman/blob/65daf2eff891b34d2c5c28ad52763aa222cb7f83/.cirrus.yml#L122-L135).

TBH, I'm not very happy about running git-commands when loading `lib.sh`, but I can't think of another way to do this.  My goal was to allow running any of the VM-based cirrus-ci tasks using `get_ci_vm.sh` + hook.  Perhaps that goal is not practical.  If you have any suggestions, they would be very welcome.

contrib/cirrus/lib.sh
```diff
@@ -80,13 +80,27 @@ CIRRUS_CI="${CIRRUS_CI:-false}"
...
+    CIRRUS_BASE_SHA=$(git rev-parse $_remote/$CIRRUS_BRANCH)
```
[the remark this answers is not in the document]
cevich: Yes [there is a default](https://github.com/containers/podman/blob/65daf2eff891b34d2c5c28ad52763aa222cb7f83/contrib/cirrus/lib.sh#L80).  The `get_ci_vm.sh` takes care of scooping up all the global env. vars. in `.cirrus.yml`, including `$DEST_BRANCH`.

contrib/cirrus/lib.sh
```diff
@@ -80,13 +80,27 @@ CIRRUS_CI="${CIRRUS_CI:-false}"
...
+    unset _remote  # running under set -a
+elif [[ -z "$CIRRUS_BASE_SHA" ]]; then
+    # No way to reliably determine this for tag
+    CIRRUS_BASE_SHA=$(git rev-parse HEAD)
+fi
 # The starting place for linting and code validation
 EPOCH_TEST_COMMIT="$CIRRUS_BASE_SHA"
 
+CIRRUS_BUILD_ID=${CIRRUS_BUILD_ID:-$RANDOM$(date +%s)}  # must be short and unique
```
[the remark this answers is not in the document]
cevich: When running under `get_ci_vm.sh` is the only time `$RANDOM` would be used.

contrib/cirrus/runner.sh
cevich: Do you mean, `make tests-included` should only be run ***in CI*** for PRs?

contrib/cirrus/runner.sh
cevich: Sorry, I don't understand your concern.  My intention is to split the `make validate` targets across two tasks:

1. "Smoke" which doesn't require compiling anything, just quickly checks code for lint and formatting stuffs.
2. "Validate" which confirms the binary compilation and does more through early testing.

Compiling takes over 5-minutes, so the idea is to fail as early as possible for items that don't require it.  I could see combining the "Validate" and "Smoke" tasks into one, but the cost would be reduced parallelism and possibly longer feedback.

contrib/cirrus/runner.sh
cevich: > unintended consequences

I did not read any comment about this.

cevich:
@edsantiago hold off on responding please...I played around with simply eliminating "smoke" and I may be a simpler solution...

cevich:
@edsantiago how about this now, I simplified it quite a bit, expanded comments, and confirmed (manually) that `hack/get_ci_vm.sh` works properly when you `source contrib/cirrus/lib.sh` on the VM.

cevich:
Note to me: if this is acceptable, need to remote "Smoke" from the branch-protection list of required checks.

cevich:
Note: Observed ~7min from code push to when `make validate` started running.  Previously this was <2 minutes (but smoke did not run all the validate checks).

.cirrus.yml
```diff
@@ -102,43 +102,6 @@ automation_task:
     always: *runner_stats
 
 
-# This task use to be called 'gating', however that name is being
```
edsantiago, line 105: As a courtesy to future maintainers, could you remove `smoke` from line 49?

.cirrus.yml
```diff
@@ -190,7 +153,21 @@ build_task:
...
+          git clone --recursive --branch=$DEST_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+          cd $CIRRUS_WORKING_DIR
+          git remote update origin
+          if [[ -n "$CIRRUS_PR" ]]; then # running for a PR
+              git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
+              git checkout pull/$CIRRUS_PR
+          else
+              git reset --hard $CIRRUS_CHANGE_IN_REPO
+          fi
+          cd $CIRRUS_WORKING_DIR
```
edsantiago, line 169: I realize this is copy-paste, and the error is in the original, but please remove this unnecessary duplicate `cd`

contrib/cirrus/runner.sh
```diff
@@ -51,11 +35,14 @@ function _run_automation() {
 }
 
 function _run_validate() {
-    # Confirm compile via prior task + cache
-    bin/podman --version
-    bin/podman-remote --version
+    # git-validation tool fails if $EPOCH_TEST_COMMIT is empty
+    # shellcheck disable=SC2154
+    if [[ -n "$EPOCH_TEST_COMMIT" ]]; then
+        make validate
```
edsantiago, line 41: Confirmed that [this step executed correctly](https://cirrus-ci.com/task/6100705229406208?command=main#L0)

edsantiago (commented):
This is much better, thank you. I'll take a harder look tomorrow morning, my brain is tired now.

I see CI complaining about "Smoke Test", and assume that's what you mean by your "protection" comment, and that you know how to deal with that.

edsantiago:
New task-map, with `smoke` removed. 
cirrus-map-pr9063

saschagrunert (approved):
LGTM

.cirrus.yml
cevich: Of course, good idea.

.cirrus.yml
cevich: done

contrib/cirrus/runner.sh
cevich: oh...that reminds me.  We can override any of the `$CIRRUS_xyz` variables under the global `env:` section in `.cirrus.yml`.  That means it's possible to "fake" a branch or tag push in a PR (temporarily).  I'll do that next, so we have a proper test of all code-paths...

cevich:
Oh! Interesting, I'm glad I tested the branch-push code path.  [Somehow the VM doesn't know what 'origin/master' is](https://cirrus-ci.com/task/6482953154854912):

```
$GOSRC/$SCRIPT_BASE/setup_environment.sh
fatal: ambiguous argument 'origin/master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

Exit status: 128
```

@edsantiago do you understand what's going on here (when loading `lib.sh`)?  It must have something to with [how we populate the repository when running because of branch-push or tag-push.](https://github.com/containers/podman/blob/62680c4ddc7e9e4c5db932a48f9878c9bea7527f/.cirrus.yml#L157-L169)

cevich:
Ahh...wait a sec...this might simply be due to 'Check Automation' and 'Ext. services' use the default Cirrus-CI clone code, not our custom script...

cevich:
...alright, let's try just moving the custom clone, so all jobs use it.

cevich:
Ref: [Successful Cirrus-CI run w/ simulated branch-push variables](https://cirrus-ci.com/build/5529167754166272)

cevich:
Ready again for a peep (minor changes + rebase).

edsantiago:
@cevich please LMK if/when this is ready for review again. 

If you need to re-push, could you please remove the `smoke` line from https://github.com/containers/podman/blob/efcd48b9668612eb975a299b79e7b6673a5addce/contrib/cirrus/setup_environment.sh#L184

edsantiago:
Ah, looks like I was too late. OK, taking a new look.

cevich:
> please remove the smoke line from

...and there might be more, I honestly didn't search for it very hard (i.e. at all).

hack/get_ci_vm.sh
```diff
@@ -158,10 +158,11 @@ parse_args(){
     VM_IMAGE_NAME="$1"
 
     # Word-splitting is desirable in this case
-    # shellcheck disable=SC2207
+    # shellcheck disable=SC2207,SC2191
```
edsantiago, line 161: Is this SC2191 necessary after fixing line 164?

hack/get_ci_vm.sh
```diff
@@ -263,7 +264,7 @@ echo -e "Note: Script can be re-used in another terminal if needed."
 echo -e "${RED}(option to delete VM presented upon exiting).${NOR}"
 # TODO: This is fairly fragile, specifically the quoting for the remote command.
 echo '#!/bin/bash' > $TMPDIR/ssh
-echo "$SSH_CMD -- -t 'cd $GOSRC && exec env \"${ENVS[*]}\" bash -il'" >> $TMPDIR/ssh
+echo "$SSH_CMD -- -t 'cd $GOSRC && exec env ${ENVS[*]} bash -il'" >> $TMPDIR/ssh
```
edsantiago, line 267: This is beyond the scope of this review, but the above will fail pretty badly if any value in `ENV` has spaces. It's not worth the effort to deal with it properly, but I think a big fat warning (here and/or line 162) would be in order.

.cirrus.yml
```diff
@@ -211,11 +194,10 @@ build_task:
 validate_task:
     name: "Validate $DISTRO_NV Build"
     alias: validate
-    skip: *tags
+    skip: *branches_and_tags
```
edsantiago, line 197: This is way above my ability to review. I've spent too long on it already, and I *think* it's okay, but IMO it's a pretty big deal to disable `make validate` on branch actions (whatever those are). I would like double- and triple-checking from @containers/podman-maintainers . 

Team, the question is, can `make validate` safely be skipped in these CI circumstances? As a reminder, that includes: 

 https://github.com/containers/podman/blob/efcd48b9668612eb975a299b79e7b6673a5addce/Makefile#L651

@cevich if you wouldn't mind re-explaining what `branches` means, I think many of our non-CI-wizard members (myself included) would appreciate it.

edsantiago (approved):
Looks mostly good to me. Three suggestions (two inline, one mentioned earlier re: removing stray `smoke`). One request for deeper review from the team.

hack/get_ci_vm.sh
cevich: Yes, this area is more complex than it appears.  Without those disables, we get:

```

In hack/get_ci_vm.sh line 162:
        $(get_env_vars)
        ^-------------^ SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In hack/get_ci_vm.sh line 163:
        VM_IMAGE_NAME="$VM_IMAGE_NAME"
        ^------------^ SC2191: The = here is literal. To assign by index, use ( [index]=value ) with no spaces. To keep as literal, quote it.

Did you mean:
        "VM_IMAGE_NAME=""$VM_IMAGE_NAME"


In hack/get_ci_vm.sh line 164:
        UPSTREAM_REMOTE="upstream"
        ^--------------^ SC2191: The = here is literal. To assign by index, use ( [index]=value ) with no spaces. To keep as literal, quote it.

Did you mean:
        "UPSTREAM_REMOTE=""upstream"
```

I'll beef up the comment to explain more fully

hack/get_ci_vm.sh
cevich: Yep, and I've experienced it.  I'll take care of this in the beefed up comment on 162, since I need to touch that anyway.

.cirrus.yml
cevich: Correction: This will be skipped on branch-push (i.e. PR merge), creating a new branch,  and on tag-push.

Esp. for tag-push, we want automation to be as simple as possible, for improved automation reliability.   Any non-essential task failures that require code changes to correct...would necessarily need a new tag to be created (potentially repeating most of the release process).  

For a new branch and a branch update...I'm not sure how to distinguish between the two in a 'skip' statement, so they're lumped together.

IMHO, the danger here is caused by nonsequential merging of PRs.  In other words, PRs that pass `validate` individually, but may not if merged in a different order.  However, in this case, I believe we can rely on either the `build` task failing (i.e. in the daily cirrus-cron build) or a future PR (based on the branch) catching the problem - since `validate` would run in that context.

cevich: (@edsantiago I've just added a comment covering this in `.cirrus.yml`)

hack/get_ci_vm.sh
edsantiago: That's weird. Could it be a difference in the version of shellcheck? I ran this on my laptop before making my comment yesterday:
```console
$ contrib/cirrus/shellcheck.sh
WARNING: It does not appear that containers/automation was installed.
         Functionality of most of this library will be negatively impacted
         This contrib/cirrus/lib.sh was loaded by contrib/cirrus/shellcheck.sh
Shellcheck: PASS
$ rpm -q ShellCheck
ShellCheck-0.7.1-3.fc33.x86_64
```
(I wouldn't have made my comment if shellcheck had failed with SC2191) (the above, needless to say, was run on a tree with the `SC2191` removed from `hack/get_ci_vm.sh` but `SC2207` left in place, because that one is really needed).

Anyhow, it's not worth re-pushing for, but it sounds like we have a problem with ShellCheck, like we always do with linters. It might be worth adding a version dump to the failure message in CI some day.

.cirrus.yml
edsantiago: Thanks. I guess it SGTM but I'd still like confirmatory review from @containers/podman-maintainers 

All else LGTM. Thanks.

hack/get_ci_vm.sh
cevich: > Could it be a difference in the version of shellcheck?

Oh, that's probably exactly what it is.  In CI we run shellcheck from F33 whereas my laptop is F32.

> but it sounds like we have a problem with ShellCheck, like we always do with linters

The important one is what gets run in CI, but I agree, we gotta take the bad along with the good...at least until the bad exceeds the good :smile:

.cirrus.yml
cevich: nod

edsantiago:
@containers/podman-maintainers could someone PTAL at my question re: the changed `skip`?

I believe this is ready to merge, but I don't fully understand the branch/tag flows and would like a final blessing from someone who does.

edsantiago:
@cevich quick followup, please remember to:

>Note to me: if this is acceptable, need to remove "Smoke" from the branch-protection list of required checks.

cevich:
> need to remove "Smoke" from the branch-protection list

Already done, thanks for the reminder though.

PR #9070, opened 2021-01-22
Fixup search

rhatdan:
podman-remote search had some FIXMEs in tests that were failing.
So I reworked the search handler to use the local abi.  This
means the podman search and podman-remote search will use the
same functions.

While doing this, I noticed we were just outputing errors via
logrus.Error rather then returning them, which works ok for
podman but the messages get lost on podman-remote.  Changed
the code to actually return the error messages to the caller.

This allows us to turn on the remaining podman-remote FIXME
tests.


pkg/api/handlers/libpod/images.go
```diff
@@ -607,70 +608,29 @@ func SearchImages(w http.ResponseWriter, r *http.Request) {
...
-		options.InsecureSkipTLSVerify = types.NewOptionalBool(!query.TLSVerify)
+		options.SkipTLSVerify = types.NewOptionalBool(!query.TLSVerify)
 	}
-
-	_, authfile, key, err := auth.GetCredentials(r)
+	ir := abi.ImageEngine{Libpod: runtime}
+	reports, err := ir.Search(r.Context(), query.Term, options)
 	if err != nil {
-		utils.Error(w, "failed to retrieve repository credentials", http.StatusBadRequest, errors.Wrapf(err, "failed to parse %q header for %s", key, r.URL.String()))
+		utils.Error(w, "Something went wrong.", http.StatusNotFound, err)
```
mheon, line 631: We need to distinguish between Not Found and Internal Server Error here.

rhatdan: NotFound is not an error, it just returns nothing.
```
$ ./bin/podman search registry.access.redhat.com/danwalsh
$ ./bin/podman-remote  search registry.access.redhat.com/danwalsh
$
```

mheon: If NotFound is not an error, why are we returning it here? Docker API indicates the correct code is 500 - https://docs.docker.com/engine/api/v1.40/#operation/ImageSearch

ashley-cui (commented):
LGTM

pkg/api/handlers/libpod/images.go
rhatdan: Well podman search imagenotfound 

Does not return an error.

I don't think we should do podman search imagenotfound to return an error.

But if we have to, then I will need to do this for only compat endpoint.

baude:
i personally do not like the idea of combining the podman and compat endpoints when they have differently supported variables.  I think it is asking for problems in the future.  that said, lgtm

pkg/api/handlers/compat/images_search.go
```diff
@@ -29,67 +34,40 @@ func SearchImages(w http.ResponseWriter, r *http.Request) {
...
+		Filters:  filters,
+	}
+	if _, found := r.URL.Query()["tlsVerify"]; found {
+		options.SkipTLSVerify = types.NewOptionalBool(!query.TLSVerify)
+	}
+	ir := abi.ImageEngine{Libpod: runtime}
+	reports, err := ir.Search(r.Context(), query.Term, options)
 	if err != nil {
-		utils.BadRequest(w, "term", query.Term, err)
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, err)
```
mheon, line 62: This should be `StatusInternalServerError`

mheon: A Bad Parameters is handled above, line 33. This is a more general "things have gone wrong" error which is a 500.

These error codes are the best error handling we have in the HTTP API so we need to get them right.

vrothberg (commented):
LGTM

mheon:
I'm coming back to this as I write the release notes for v3.1.0-RC2, and I notice that we did not document the new options added to the handler in Swagger. The same can be said of the other reworks Dan did this patch cycle.

What's the thinking on these - are they internal-only and undocumented otherwise, or do we need to go back and find everything added and add docs? @jwhonce @baude @rhatdan

rhatdan:
These are internal and bug fixes.

PR #9071, opened 2021-01-23
Bump golang.org/x/crypto

TomSweeneyRedHat:
Bump golang.org.x/cyrpto to the latest

Partially addresses: https://bugzilla.redhat.com/show_bug.cgi?id=1910255, we'll need to do similar in other branches TBD.
This is a closed up a low level CVE.  Code in the crypto library has a vulnerability, but we don't call that code.  So this is just to be ultra-safe.

Signed-off-by: root <root@localhost.localdomain>


mheon:
Probably need to set your Git name/email - your signed-off-by looks wrong

TomSweeneyRedHat:
I reset my signature, missed that on the new vm.  I've repushed.

Issue #8313, opened 2020-11-12
Static "nix" build fails to compile podman


cevich:
/kind bug

**Description**

After fixing a Cirrus-CI caching problem (related to available VM disk space on boot), podman fails to compile.  I've temporarily made the static build task "optional", so the failure is currently ignored.

**Steps to reproduce the issue:**

1.  Execute a Cirrus-CI build on a branch or in a PR

**Describe the results you received:**

```
  go build -mod=vendor  -gcflags 'all=-trimpath=/tmp/nix-build-podman.drv-0/podman' -asmflags 'all=-trimpath=/tmp/nix-build-podman.drv-0/podman' -ldflags '-X github.com/containers/podman/v2/libpod/define.gitCommit=12dbc1ff7dd7468f0be22e597ed9e552ad545e20-dirty -X github.com/containers/podman/v2/libpod/define.buildInfo=315532800 -X github.com/containers/podman/v2/libpod/config._installPrefix=/usr/local -X github.com/containers/podman/v2/libpod/config._etcDir=/etc -s -w -linkmode external -extldflags "-static -lm"' -tags "static netgo osusergo exclude_graphdriver_btrfs exclude_graphdriver_devicemapper seccomp apparmor selinux" -o bin/podman ./cmd/podman
  # runtime/cgo
  In file included from /nix/store/ypih4394q488ljr421x8jak55vmr0ckn-glibc-2.32-dev/include/errno.h:28,
                   from cgo-gcc-prolog:30:
  /nix/store/ypih4394q488ljr421x8jak55vmr0ckn-glibc-2.32-dev/include/bits/errno.h:26:11: fatal error: linux/errno.h: No such file or directory
     26 | # include <linux/errno.h>
        |           ^~~~~~~~~~~~~~~
  compilation terminated.
  make: *** [Makefile:195: bin/podman] Error 2
error: build of '/nix/store/q0256ixhmc5myyxpgqimw3gp1rqji8pq-podman.drv' failed
```

[Ref. Cirrus task output from master](https://cirrus-ci.com/task/6579811101442048?command=main#L89)

**Describe the results you expected:**

Exit 0

**Additional information you deem important (e.g. issue happens only occasionally):**

[As part of the fix, please comment out the line which causes failures to be ignored](https://github.com/containers/podman/blob/master/.cirrus.yml#L336)

Otherwise, please feel free to mention me if you need any help with the automation-side of things

hswong3i:
[...]

PR #9082, opened 2021-01-25
Fix static build cache by using cachix

saschagrunert:
It looks like we always hit the caching issue in Cirrus CI described within https://github.com/containers/podman/issues/8313. A solution around that is to use cachix, which has been pre-populated from my local machine.

To push all (runtime and build) dependencies, we can leverage a pre-populated store by:
```
> nix-store -qR --include-outputs $(nix-instantiate nix/default.nix) | cachix push podman
```
The cache can be re-used by everybody to rapidly build static Podman binaries: https://app.cachix.org/cache/podman

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -344,14 +344,13 @@ static_alt_build_task:
         ALT_NAME: 'Static build'
         # Do not use 'latest', fixed-version tag for runtime stability.
         CTR_FQIN: "docker.io/nixos/nix:2.3.6"
-    # This is critical, it helps to avoid a very lengthy process of
-    # statically building every dependency needed to build podman.
-    # Assuming the dependency and build description hasn't changed,
-    # this cache ensures only the static podman binary is built.
-    nix_cache:
-        folder: '/var/cache/nix'
-        # Cirrus will calculate/use sha of this output as the cache key
-        fingerprint_script: echo "${IMAGE_SUFFIX}" && cat nix/*
+        # Authentication token for pushing the build cache to cachix.
+        # This is critical, it helps to avoid a very lengthy process of
+        # statically building every dependency needed to build podman.
+        # Assuming the pinned nix dependencies in nix/nixpkgs.json have not
+        # changed, this cache will ensure that only the static podman binary is
+        # built.
+        CACHIX_AUTH_TOKEN: ENCRYPTED[df0d4d0a67474e8ea49cc503221dcb912b7e2ba45c8ec4bf2e5fd9c49a18ac21c24bacee59b5393355ed9e4358d2baef]
     setup_script: *setup
     main_script: *main
     always: *binary_artifacts
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -191,15 +191,14 @@ function _run_altbuild() {
             req_env_vars CTR_FQIN
             [[ "$UID" -eq 0 ]] || \
                 die "Static build must execute nixos container as root on host"
-            mkdir -p /var/cache/nix
-            podman run -i --rm -v /var/cache/nix:/mnt/nix:Z \
-                $CTR_FQIN cp -rfT /nix /mnt/nix
-            podman run -i --rm -v /var/cache/nix:/nix:Z \
-                -v $PWD:$PWD:Z -w $PWD $CTR_FQIN \
-                nix --print-build-logs --option cores 4 --option max-jobs 4 \
-                    build --file ./nix/
-            # result symlink is absolute from container perspective :(
-            cp /var/cache/$(readlink result)/bin/podman ./  # for cirrus-ci artifact
+            podman run -i --rm \
+                -e CACHIX_AUTH_TOKEN \
+                -v $PWD:$PWD:Z -w $PWD $CTR_FQIN sh -c \
+                "nix-env -iA cachix -f https://cachix.org/api/v1/install && \
+                 cachix use podman && \
+                 nix-build nix && \
+                 nix-store -qR --include-outputs \$(nix-instantiate nix/default.nix) | grep -v podman | cachix push podman && \
+                 cp -R result/bin ."
             rm result  # makes cirrus puke
             ;;
         *)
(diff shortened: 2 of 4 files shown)


saschagrunert:
@cevich are we able to expose a secret as environment variable in Cirrus? This way we can automatically update the cache for each run.

.cirrus.yml
```diff
@@ -333,9 +333,6 @@ static_alt_build_task:
     only_if: *not_docs
     depends_on:
         - build
-    # Community-maintained task, may fail on occasion.  If so, uncomment
-    # the next line and file an issue with details about the failure.
-    allow_failures: $CI == $CI
```
vrothberg, line 338: Is this needed? The nix package is not something we committed to support officially.

saschagrunert: I reverted the change. The comment indicates that it was commented out, right?

cevich: That line simply always evaluates "true" (it has to be done that funny way in cirrus).  It means we run the task, but basically ignore the results.  Commenting out that line (for example) allows a PR to be merged if (for some external reason) it's failing: We don't want that to block, e.g. a podman release.  So yes, it's fine to just comment out the `allow_failures` but please also leave the comment-there in case somebody needs it.

cevich:
> @cevich are we able to expose a secret as environment variable in Cirrus? This way we can automatically update the cache for each run.

Yes this is easily possible, and I'm happy to set this up for you.  Just e-mail or private-message me the variable name and what you want it to contain.

saschagrunert:
> > @cevich are we able to expose a secret as environment variable in Cirrus? This way we can automatically update the cache for each run.
> 
> Yes this is easily possible, and I'm happy to set this up for you. Just e-mail or private-message me the variable name and what you want it to contain.

Thank you! Sent you a mail a few seconds ago.

cevich:
@saschagrunert you should have received a reply from me w/ the key.  Please let me know if you didn't get my mail and I can copy-paste my reply here instead.

saschagrunert:
Works, thanks again @cevich! :pray: 

This is ready for review, PTAL @containers/podman-maintainers

vrothberg (commented):
LGTM

.cirrus.yml
```diff
@@ -344,14 +344,8 @@ static_alt_build_task:
         ALT_NAME: 'Static build'
         # Do not use 'latest', fixed-version tag for runtime stability.
         CTR_FQIN: "docker.io/nixos/nix:2.3.6"
-    # This is critical, it helps to avoid a very lengthy process of
-    # statically building every dependency needed to build podman.
-    # Assuming the dependency and build description hasn't changed,
-    # this cache ensures only the static podman binary is built.
```
cevich, line 350: I think this comment is helpful for developers, who may not be privy to the history and detailed workings of this task.  Would you mind adding a similar comment, highlighting the role/importance of cache in this task?

saschagrunert: Sure, moved the comment to the `CACHIX_AUTH_TOKEN` variable definition.

contrib/cirrus/runner.sh
```diff
@@ -191,15 +191,14 @@ function _run_altbuild() {
...
-            # result symlink is absolute from container perspective :(
-            cp /var/cache/$(readlink result)/bin/podman ./  # for cirrus-ci artifact
+            podman run -i --rm \
+                -e CACHIX_AUTH_TOKEN \
+                -v $PWD:$PWD:Z -w $PWD $CTR_FQIN sh -c \
+                "nix-env -iA cachix -f https://cachix.org/api/v1/install && \
+                 cachix use podman && \
+                 nix-build nix && \
+                 nix-store -qR --include-outputs \$(nix-instantiate nix/default.nix) | grep -v podman | cachix push podman && \
+                 cp -R result/bin ."
```
cevich, line 201: Since this makes our CI dependent on an external 3rd party service, could you please add a test for it to `$SCRIPT_BASE/ext_svc_check.sh`?

I'm guessing cachix is https based?  If so you're in luck, since the script is already setup for this.  In this case, just add `cachix.org 443` to `$SCRIPT_BASE/required_host_ports.txt` and you're done.  If not, then a more specialized check may need to be added.

.cirrus.yml
```diff
@@ -344,14 +344,12 @@ static_alt_build_task:
...
-    nix_cache:
-        folder: '/var/cache/nix'
-        # Cirrus will calculate/use sha of this output as the cache key
-        fingerprint_script: echo "${IMAGE_SUFFIX}" && cat nix/*
+        # Authentication token for pushing the build cache to cachix.
+        # This is critical, it helps to avoid a very lengthy process of
+        # statically building every dependency needed to build podman.
+        # Assuming the dependency and build description hasn't changed,
+        # this cache ensures only the static podman binary is built.
+        CACHIX_AUTH_TOKEN: ENCRYPTED[df0d4d0a67474e8ea49cc503221dcb912b7e2ba45c8ec4bf2e5fd9c49a18ac21c24bacee59b5393355ed9e4358d2baef]
```
cevich, line 352: Thanks! This is better, but I think the last two lines in the comment are inaccurate, no? I'm fine if you prefer to just remove them

cevich (commented):
Thanks very much for working to fixup this task.  None of us has much knowledge of how this static build works, so we really depend on these updates.

Other than my two comments, this LGTM

.cirrus.yml
saschagrunert: I made the last two lines more clear. If we do not change the pinned nix dependencies, then we will always produce a cached and reproducible build.

contrib/cirrus/runner.sh
saschagrunert: Alright, added the check for the podman cache at `podman.cachix.org`.

.cirrus.yml
cevich: That's helpful, thanks.

contrib/cirrus/runner.sh
cevich: Great, I'm glad it was as simple as that.

cevich:
(I restarted the failed `int remote fedora-33 root host` task, no idea why it failed I've never seen that error before...hopefully it's just a flake)

PR #9084, opened 2021-01-25
Fix --arch and --os flags to work correctly

rhatdan:
Currently podman implements --override-arch and --overide-os
But Podman has made these aliases for --arch and --os.  No
reason to have to specify --override, since it is clear what
the user intends.

Currently if the user specifies an --override-arch field but the
image was previously pulled for a different Arch, podman run uses
the different arch.  This PR also fixes this issue.

Fixes: https://github.com/containers/podman/issues/8001

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -474,29 +474,29 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	)
 	_ = cmd.RegisterFlagCompletionFunc(oomScoreAdjFlagName, completion.AutocompleteNone)
 
-	overrideArchFlagName := "override-arch"
+	archFlagName := "arch"
 	createFlags.StringVar(
-		&cf.OverrideArch,
-		overrideArchFlagName, "",
+		&cf.Arch,
+		archFlagName, "",
 		"use `ARCH` instead of the architecture of the machine for choosing images",
 	)
-	_ = cmd.RegisterFlagCompletionFunc(overrideArchFlagName, completion.AutocompleteNone)
+	_ = cmd.RegisterFlagCompletionFunc(archFlagName, completion.AutocompleteArch)
 
-	overrideOSFlagName := "override-os"
+	osFlagName := "os"
 	createFlags.StringVar(
-		&cf.OverrideOS,
-		overrideOSFlagName, "",
+		&cf.OS,
+		osFlagName, "",
 		"use `OS` instead of the running OS for choosing images",
 	)
-	_ = cmd.RegisterFlagCompletionFunc(overrideOSFlagName, completion.AutocompleteNone)
+	_ = cmd.RegisterFlagCompletionFunc(osFlagName, completion.AutocompleteOS)
 
-	overrideVariantFlagName := "override-variant"
+	variantFlagName := "variant"
 	createFlags.StringVar(
-		&cf.OverrideVariant,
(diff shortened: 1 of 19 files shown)


rhatdan:
@lisongmin PTAL

cmd/podman/common/create.go
```diff
@@ -474,29 +474,29 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
...
+	archFlagName := "arch"
 	createFlags.StringVar(
-		&cf.OverrideArch,
-		overrideArchFlagName, "",
+		&cf.Arch,
+		archFlagName, "",
 		"use `ARCH` instead of the architecture of the machine for choosing images",
 	)
-	_ = cmd.RegisterFlagCompletionFunc(overrideArchFlagName, completion.AutocompleteNone)
+	_ = cmd.RegisterFlagCompletionFunc(archFlagName, completion.AutocompleteNone)
```
Luap99, line 483: If I remember correctly you added autocomplete functions for `arch` and `os`. It would be good if you add them here.

rhatdan: Thanks for reminding me.

rhatdan:

TomSweeneyRedHat:
LGTM 
assuming happy testst

edsantiago:
LGTM but all I can give it right now is a cursory once-over. FWIW I did run my usual `ack override-` and `ack '\.Override'` (`ack` is a nice `grep -r`) to look for strays: none found. Nice.

saschagrunert (approved):
LGTM

vrothberg:
/hold

I doubt that this change fixes https://github.com/containers/podman/issues/8001. Will test a bit.

cmd/podman/containers/create.go
```diff
@@ -237,17 +237,20 @@ func pullImage(imageName string) (string, error) {
...
 
-	if cliVals.Platform != "" {
-		if cliVals.OverrideArch != "" || cliVals.OverrideOS != "" {
-			return "", errors.Errorf("--platform option can not be specified with --override-arch or --override-os")
-		}
-		split := strings.SplitN(cliVals.Platform, "/", 2)
-		cliVals.OverrideOS = split[0]
-		if len(split) > 1 {
-			cliVals.OverrideArch = split[1]
+	if cliVals.Platform != "" || cliVals.Arch != "" || cliVals.OS != "" {
```
vrothberg, line 240: @rhatdan can you add a `// FIXME` that states that all this logic should go into the backend? While this fix works for the clients, it does not work for the server.

I'd prefer something similar to https://github.com/vrothberg/libpod/commit/6aa9622350853aa34ec3423f0ae1dd1fd730532e but I never managed to get back to it.  I would love to remove the `pull` entirely (as the backend does that already) but that would lose the progress bars for the remote client.

vrothberg:
> I doubt that this change fixes #8001. Will test a bit.

Looking good :+1:

cmd/podman/containers/create.go
rhatdan: When we do the containers/common image, could we take care of this there?

vrothberg: Yes we could do that. But that would be based on `libpod/image` which is currently not doing that it should. That's why I think that https://github.com/vrothberg/libpod/commit/6aa9622350853aa34ec3423f0ae1dd1fd730532e is closer to tackle the root of the problem.

PR #9090, opened 2021-01-25
[NO TESTS NEEDED] Ensure shutdown handler access is syncronized

mheon:
There was a potential race where two handlers could be added at the same time. Go Maps are not thread-safe, so that could do unpleasant things. Add a mutex to keep things safe.

Also, swap the order or Register and Start for the handlers in Libpod runtime created. As written, there was a small gap between Start and Register where SIGTERM/SIGINT would be completely ignored, instead of stopping Podman. Swapping the two closes  this gap.

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -180,19 +180,19 @@ func newRuntimeFromConfig(ctx context.Context, conf *config.Config, options ...R
 		}
 	}
 
+	if err := shutdown.Register("libpod", func(sig os.Signal) error {
+		os.Exit(1)
+		return nil
+	}); err != nil && errors.Cause(err) != shutdown.ErrHandlerExists {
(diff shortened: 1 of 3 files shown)


libpod/runtime.go
```diff
@@ -180,6 +180,13 @@ func newRuntimeFromConfig(ctx context.Context, conf *config.Config, options ...R
 		}
 	}
 
+	if err := shutdown.Register("libpod", func(sig os.Signal) error {
```
matejvasek, line 183: Similar code is in `server.Serve()`.

mheon: Good catch, fixed

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
/lgtm
/hold
on tests

Issue #8679, opened 2020-12-10
Need regression tests for podman build and mounts.conf


vrothberg:
I am currently looking at https://bugzilla.redhat.com/show_bug.cgi?id=1895027.  We regressed in the 1.6.X series on mounting secrets during builds.  It's fairly easy to reproduce with a Dockerfile:
```Dockerfile
FROM $IMAGE
RUN ls $MOUNT
```
With $MOUNT being the expected mount of the secret. For testing, we could setup a custom `mounts.conf` mounting a file.

Pulling in @edsantiago for input.

rhatdan:
@vrothberg @edsantiago No movement on this issue.

edsantiago:
So, this may be a dumb question, but, do we expect this to actually work in podman right now?
```console
# mkdir -p /tmp/podmantest/secrets
# echo /tmp/podmantest/secrets:/run/secrets >/tmp/podmantest/mounts.conf
# echo shhhh >/tmp/podmantest/secrets/file
# printf "FROM quay.io/libpod/testimage:20200929\nRUN ls -l /run/secrets\n" | podman --default-mounts-file=/tmp/podman/mounts.conf build -t foo -
STEP 1: FROM quay.io/libpod/testimage:20200929
STEP 2: RUN ls -l /run/secrets
total 0
STEP 3: COMMIT foo
--> 0acf9c0d708
0acf9c0d708255abb9ef5c1984c1558ca5d735b1fcd4619131a9b1c7254e18ed
```
It doesn't work via `podman run` either:
```console
# podman --default-mounts-file=/tmp/podman/mounts.conf run quay.io/libpod/testimage:20200929 ls -l /run/secrets
total 0
```
Am I misunderstanding the `--default-mounts-file` option?

rhatdan:
This worked for me as root.

```
# cat /tmp/podman/mounts.conf 
/tmp/dan:/run/secrets
# mkdir /tmp/dan/
# touch /tmp/dan/walsh
# podman --default-mounts-file=/tmp/podman/mounts.conf run quay.io/libpod/testimage:20200929 ls -l /run/secrets
total 0
-rw-r--r--    1 root     root             0 Jan 25 19:36 walsh
```

edsantiago:
@rhatdan the `podman run` error was a stupid typo on my end. I confirm that `podman run` works as expected with a custom mounts.conf. `podman build` does not (for me). Can you get `podman build` to see your secret?
```console
[...]

PR #9091, opened 2021-01-25
System tests: cover gaps from the last month

edsantiago:
- stop: test --all and --ignore (#9051)
- build: test /run/secrets (#8679, but see below)
- sensitive mount points: deal with 'stat' failures
- selinux: confirm useful diagnostics on unknown labels (#8946)

The 'build' test is intended as a fix for #8679, in which
'podman build' does not mount secrets from mounts.conf.
Unfortunately, as of this writing, 'podman build' does
not pass the --default-mounts-file option to buildah,
so there's no reasonable way to test this path. Still,
we can at least confirm /run/secrets on 'podman run'.

The /sys thing is related to #8949: RHEL8, rootless, cgroups v1.
It's just a workaround to get gating tests to pass on RHEL.

### DIFF
--- a/test/system/050-stop.bats
+++ b/test/system/050-stop.bats
@@ -30,6 +30,49 @@ load helpers
     run_podman rm $cid
 }
 
+# #9051 : podman stop --all was not working with podman-remote
+@test "podman stop --all" {
+    # Start three containers, create (without running) a fourth
+    run_podman run -d --name c1 $IMAGE sleep 20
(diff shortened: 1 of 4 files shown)


mheon:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
/lgtm
/hold
hold for happy tests

edsantiago:
Yay! Releasing hold so @rhatdan can proceed with #9094 

/hold cancel

Issue #6618, opened 2020-06-16
inspect command: Populate Bridge field in network settings


drewbailey:
/kind feature

**Populate bridge field in network settings when applicable**

**Steps to reproduce the issue:**

1. `sudo podman run -m=40m -p 8080:89 --network=bridge redis`

2. `sudo podman inspect <container>`

**Describe the results you received:**

```json
    "NetworkSettings": {
      "EndpointID": "",
      "Gateway": "10.88.0.1",
      "IPAddress": "10.88.0.239",
      "IPPrefixLen": 16,
      "IPv6Gateway": "",
      "GlobalIPv6Address": "",
      "GlobalIPv6PrefixLen": 0,
      "MacAddress": "62:86:4f:c2:28:01",
      "Bridge": "",
      "SandboxID": "",
      "HairpinMode": false,
      "LinkLocalIPv6Address": "",
      "LinkLocalIPv6PrefixLen": 0,
      "Ports": [
        {
          "hostPort": 8080,
          "containerPort": 89,
          "protocol": "tcp",
          "hostIP": ""
        }
      ],
      "SandboxKey": "/var/run/netns/cni-5ada0c78-0a5f-ea18-a950-7141c110f46b"
    },
```
**Describe the results you expected:**

Bridge to be populated with the cni id of the bridge in use

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
→ podman version
Version:            1.9.3
RemoteAPI Version:  1
Go Version:         go1.13.8
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
(paste your output here)
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
(paste your output here)
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

QiWang19:
Do you mean there should  be `Network` field Under "NetworkSettings"?

```
"Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "0df86146427ffaaf905bee17342ed31d012d761fc34a8788584427f53e1ac99b",
                    "EndpointID": "bd120b2e6fd090dbb180342956f9daa440ea5a3db78466dc1ed19bf3ba197dfd",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
[...]

PR #9092, opened 2021-01-25
Add default net info in container inspect

baude:
when inspecting a container that is only connected to the default
network, we should populate the default network in the container inspect
information.

Fixes: #6618

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -973,6 +973,13 @@ func (c *Container) getContainerNetworkInfo() (*define.InspectNetworkSettings, e
 		return nil, err
 	}
 
+	// If this a container with a netns and on the default network
+	if c.state.NetNS != nil && isDefault {
+		settings.Networks = make(map[string]*define.InspectAdditionalNetwork, len(networks))
+		cniNet := new(define.InspectAdditionalNetwork)
+		cniNet.NetworkID = c.runtime.netPlugin.GetDefaultNetworkName()
+		settings.Networks[c.runtime.netPlugin.GetDefaultNetworkName()] = cniNet
+	}
 	// We can't do more if the network is down.
 	if c.state.NetNS == nil {
 		// We still want to make dummy configurations for each CNI net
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

libpod/networking_linux.go
```diff
@@ -973,6 +973,13 @@ func (c *Container) getContainerNetworkInfo() (*define.InspectNetworkSettings, e
 		return nil, err
 	}
 
+	// If this a container with a netns and on the default network
+	if c.state.NetNS != nil && isDefault {
```
mheon, line 977: I don't think this is correct... Docker does not populate Networks when only the default is in use, last I checked? And, AFAIK, the original request was not about Networks, it was about populating the `Bridge` field in `InspectNetworkSettings`?

mheon: Hm. Nevermind, Docker is populating additional networks, even when only the default is specified. I stand corrected.

libpod/networking_linux.go
```diff
@@ -973,6 +973,13 @@ func (c *Container) getContainerNetworkInfo() (*define.InspectNetworkSettings, e
...
+		settings.Networks = make(map[string]*define.InspectAdditionalNetwork, len(networks))
+		cniNet := new(define.InspectAdditionalNetwork)
+		cniNet.NetworkID = c.runtime.netPlugin.GetDefaultNetworkName()
+		settings.Networks[c.runtime.netPlugin.GetDefaultNetworkName()] = cniNet
+	}
```
mheon, line 982: This block is still unnecessary, though. All your work here is undone by the `if len(networks) > 0` block below.

mheon: You also are going to need to modify the other use of `!isDefault` (L958) - the one that handles non-running containers.

test/e2e/inspect_test.go
```diff
@@ -443,4 +443,16 @@ var _ = Describe("Podman inspect", func() {
 		Expect(inspect.OutputToString()).To(Equal(`"{"80/tcp":[{"HostIp":"","HostPort":"8080"}]}"`))
 	})
 
+	It("Verify container inspect has default network", func() {
```
mheon, line 446: Please add another test with a container that is not running.

libpod/networking_linux.go
baude: docker does populate it, i checked.  also docker does not populate the bridge field apparently?
```sudo docker inspect foo | jq .[0].NetworkSettings.Bridge
""```

libpod/networking_linux.go
baude: we should talk about that ... look at the block just beneath this. it looks there is step prior where the map is setup. lets discuss in watercooler.

mheon:
Taking this over - #9109

PR #9095, opened 2021-01-25
podman-remote ps --external --pod --sort do not work.

rhatdan:
Fixup the bindings and the handling of the --external --pod and --sort
flags.

The --storage option was renamed --external, make sure we use
external up and down the stack.

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -78,7 +78,7 @@ func listFlagSet(cmd *cobra.Command) {
 	flags := cmd.Flags()
 
 	flags.BoolVarP(&listOpts.All, "all", "a", false, "Show all the containers, default is only running containers")
-	flags.BoolVar(&listOpts.Storage, "external", false, "Show containers in storage not controlled by Podman")
+	flags.BoolVar(&listOpts.External, "external", false, "Show containers in storage not controlled by Podman")
 
 	filterFlagName := "filter"
 	flags.StringSliceVarP(&filters, filterFlagName, "f", []string{}, "Filter output based on conditions given")
@@ -132,10 +132,10 @@ func checkFlags(c *cobra.Command) error {
 	}
 	cfg := registry.PodmanConfig()
 	if cfg.Engine.Namespace != "" {
-		if c.Flag("storage").Changed && listOpts.Storage {
-			return errors.New("--namespace and --storage flags can not both be set")
+		if c.Flag("storage").Changed && listOpts.External {
+			return errors.New("--namespace and --external flags can not both be set")
 		}
-		listOpts.Storage = false
+		listOpts.External = false
 	}
 
 	return nil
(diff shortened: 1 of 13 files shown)


mheon:
The change to Pod over the remote API is breaking - the responses used to
always contain pod information, now they only do if a new parameter is
passed.

cmd/podman/containers/ps.go
```diff
@@ -132,10 +132,10 @@ func checkFlags(c *cobra.Command) error {
 	}
 	cfg := registry.PodmanConfig()
 	if cfg.Engine.Namespace != "" {
-		if c.Flag("storage").Changed && listOpts.Storage {
+		if c.Flag("storage").Changed && listOpts.External {
 			return errors.New("--namespace and --storage flags can not both be set")
```
vrothberg, line 136:
```suggestion
			return errors.New("--namespace and --external flags can not both be set")
```

pkg/api/handlers/libpod/containers.go
```diff
@@ -63,11 +62,14 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
 		All       bool                `schema:"all"`
+		External  bool                `schema:"external"`
```
vrothberg, line 65: I think this requires swagger changes as well to document the new parameters. I couldn't find `external` mentioned.

Luap99:
podman-remote --sort and --pod work
can be tested with `diff <(bin/podman ps -a --sort id --pod) <(bin/podman-remote ps -a --sort id --pod)`

A Pod parameter should not be used for the bindings see https://github.com/containers/podman/pull/7223

rhatdan:
@Luap99 Thanks, I did not see that.  Added a comment to the code about this.

pkg/api/handlers/libpod/containers.go
rhatdan: Added.

cmd/podman/containers/ps.go
```diff
@@ -78,7 +78,7 @@ func listFlagSet(cmd *cobra.Command) {
 	flags := cmd.Flags()
 
 	flags.BoolVarP(&listOpts.All, "all", "a", false, "Show all the containers, default is only running containers")
-	flags.BoolVar(&listOpts.Storage, "external", false, "Show containers in storage not controlled by Podman")
+	flags.BoolVar(&listOpts.External, "external", false, "Show containers in storage not controlled by Podman")
```
mheon, line 81: This is being normalized so Storage is also accepted, right?

rhatdan: Yes that happened in an earlier PR.
https://github.com/containers/podman/blob/master/cmd/podman/utils/alias.go#L24-L25

rhatdan:
@containers/podman-maintainers This is ready to go in. PTAL

mheon:
LGTM

PR #9097, opened 2021-01-26
Bump golang.org/x/crypto v3.0

TomSweeneyRedHat:
Bumping golang.org/x/crypto to the latest.

Please double check me on this one.  This is the first time I've vendored somewhere other than the main branch and I'd some fun getting to this point.

Partially addresses: https://bugzilla.redhat.com/show_bug.cgi?id=1910255, we'll need to do similar in other branches TBD.
This is a closed up a low level CVE. Code in the crypto library has a vulnerability, but we don't call that code. So this is just to be ultra-safe.


vrothberg (commented):
LGTM
Restarted what look liked flaked jobs

Issue #8281, opened 2020-11-09
regression: podman-remote exec: no stderr


edsantiago:
This is an exact resurrection of #4096. It was working in v2.0, broken in v2.1. It was not caught because the CI test was disabled due to flake #7241 .

In window 1:
```console
$ ./bin/podman system service --timeout=0
```
Window 2:
```console
$ ./bin/podman-remote run -d --name foo alpine sleep 60
282efa747bd064007e594362a4697567226f8ebdb0b288e98f231ebdd464a7d7
$ ./bin/podman-remote exec foo /no/such/command
$ ./bin/podman-remote exec foo /etc
```
Note the missing output: there should be errors from each one. The errors actually appear in window 1 (server):
```
ERRO[0031] error attaching to container282efa747bd064007e594362a4697567226f8ebdb0b288e98f231ebdd464a7d7 exec session 018f23462f0c1d704a175f825c327dcce1ce3dba37bcfcbf8db236600f98ddc9: executable file `/no/such/command` not found in $PATH: No such file or directory: OCI not found
ERRO[0034] error attaching to container 282efa747bd064007e594362a4697567226f8ebdb0b288e98f231ebdd464a7d7 exec session 3b3852b4b3b6eca4e321385be65afc0fd52b061a37978f8ef029a367a7d7fcbc: open executable: Operation not permitted: OCI permission denied
```

edsantiago:
Whoever fixes this: please include the following patch in your PR:
```diff
diff --git a/test/system/075-exec.bats b/test/system/075-exec.bats
index edd7dedc4..19b4c1afc 100644
--- a/test/system/075-exec.bats
+++ b/test/system/075-exec.bats
@@ -6,8 +6,6 @@
 load helpers

 @test "podman exec - basic test" {
-    skip_if_remote "FIXME: pending #7241"
-
     rand_filename=$(random_string 20)
     rand_content=$(random_string 50)

```

rhatdan:
@mheon this seems like something you should be able to understand/fix?

zhangguanzhang:
https://github.com/containers/podman/commit/2ea9dac5e1d00b2820bd7156e3bea4b9fd98c1e6
this pr change

zhangguanzhang:
@mheon

rhatdan:
@mheon This seems pretty critical, I have just reproduced.  Any idea where this is happening?

mheon:
Honestly, no. Too many different errors happening at different places and times in the exec code.

[...]

Issue #8972, opened 2021-01-14
flake: remote: podman cp dir from host to container


edsantiago:
podman-remote cp tests are flaking; so far only on ubuntu-2004. Symptom is:
```console
# podman-remote --url unix:/tmp/podman.RlDd0r exec cpcontainer cat /usr//hostfile0
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: copy contents of dir to usr/ (cp -> ctr:/usr/)
#| expected: 'random-0-C2gGRg2DgJ'
#|   actual: ''
#\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Instances to date:

## sys: podman cp dir from host to container
* gce_instance:prior-ubuntu : sys remote ubuntu-2004 root host
  * PR #8960
    * [01-13 12:44](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6496656617111552/html/sys-remote-ubuntu-2004-root-host.log.html)
  * PR #8923
    * [01-11 06:52](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5012806921617408/html/sys-remote-ubuntu-2004-root-host.log.html)
  * PR #8905
    * [01-11 19:21](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6746605023920128/html/sys-remote-ubuntu-2004-root-host.log.html)
  * PR #8904
    * [01-06 19:44](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6111985689427968/html/sys-remote-ubuntu-2004-root-host.log.html)
  * PR #8813
    * [12-23 06:04](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6477341243211776/html/sys-remote-ubuntu-2004-root-host.log.html)
  * PR #8804
    * [12-22 12:18](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5076482294808576/html/sys-remote-ubuntu-2004-root-host.log.html)

Also seen in [nightly cron](https://cirrus-ci.com/task/6345347066757120)

[...]

PR #9105, opened 2021-01-26
remote exec: write conmon error on hijacked connection

vrothberg:
Make sure to write error from conmon on the hijacked http connection.
This fixes issues where errors were not reported on the client side,
for instance, when specified command was not found on the container.

To future generations: I am sorry.  The code is complex, and there are
many interdependencies among the concurrent goroutines.  I added more
complexity on top but I don't have a good idea of how to reduce
complexity in the available time.

Fixes: #8281

### DIFF
--- a/libpod/oci_conmon_exec_linux.go
+++ b/libpod/oci_conmon_exec_linux.go
@@ -126,20 +126,25 @@ func (r *ConmonOCIRuntime) ExecContainerHTTP(ctr *Container, sessionID string, o
 	}()
 
 	attachChan := make(chan error)
+	conmonPipeDataChan := make(chan conmonPipeData)
 	go func() {
 		// attachToExec is responsible for closing pipes
-		attachChan <- attachExecHTTP(ctr, sessionID, req, w, streams, pipes, detachKeys, options.Terminal, cancel, hijackDone, holdConnOpen)
+		attachChan <- attachExecHTTP(ctr, sessionID, req, w, streams, pipes, detachKeys, options.Terminal, cancel, hijackDone, holdConnOpen, execCmd, conmonPipeDataChan, ociLog)
 		close(attachChan)
 	}()
 
-	// Wait for conmon to succeed, when return.
-	if err := execCmd.Wait(); err != nil {
-		return -1, nil, errors.Wrapf(err, "cannot run conmon")
-	}
+	// NOTE: the channel is needed to communicate conmon's data.  In case
+	// of an error, the error will be written on the hijacked http
+	// connection such that remote clients will receive the error.
+	pipeData := <-conmonPipeDataChan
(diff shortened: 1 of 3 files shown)


edsantiago:
Thank you!

vrothberg:
@edsantiago this might be it.  The code is tricky ... lot's of spaghetti dependencies and I didn't help making it simpler. I'll have another look tomorrow morning with a fresh pair of eyes in hope to find a simpler solution without larger refactorings

edsantiago:
Flake #8972, restarted

vrothberg:
@containers/podman-maintainers PTAL, ready to merge from my side.

libpod/oci_conmon_exec_linux.go
```diff
@@ -488,9 +493,16 @@ func (r *ConmonOCIRuntime) startExec(c *Container, sessionID string, options *Ex
 }
 
 // Attach to a container over HTTP
-func attachExecHTTP(c *Container, sessionID string, r *http.Request, w http.ResponseWriter, streams *HTTPAttachStreams, pipes *execPipes, detachKeys []byte, isTerminal bool, cancel <-chan bool, hijackDone chan<- bool, holdConnOpen <-chan bool) (deferredErr error) {
+func attachExecHTTP(c *Container, sessionID string, r *http.Request, w http.ResponseWriter, streams *HTTPAttachStreams, pipes *execPipes, detachKeys []byte, isTerminal bool, cancel <-chan bool, hijackDone chan<- bool, holdConnOpen <-chan bool, execCmd *exec.Cmd, conmonPipeDataChan chan<- conmonPipeData, ociLog string) (deferredErr error) {
+	// NOTE: As you may notice, the attach code is quite complex.
+	// Many things happen concurrently and yet are interdependent.
+	// If you ever change this function, make sure to write to the
+	// conmonPipeDataChan in case of an error.
+
```
rhatdan, line 501: Would it be more elegant to catch the errors and do the conmonPipeData once?


defer func() { 
      if Err != nil {
 		conmonPipeDataChan <- conmonPipeData{-1, Err}
                return Err
     }
}
```

rhatdan (approved):

libpod/oci_conmon_exec_linux.go
vrothberg: Unfortunately that doesn't work unless we introduce another (thread safe) variable that would allow us to know whether we already wrote to the channel.

I can try to make it nicer if you want to.

rhatdan: I just worry abour a future deveopers adding a return error and not doing writing to the pipe.  For now, I will merge.

rhatdan:
LGTM

PR #9107, opened 2021-01-26
[3.0] disable dnsname when --internal

baude:
when doing a network creation, the dnsname plugin should be disabled
when the --internal bool is set.  a warning is displayed if this
happens and docs are updated.


TomSweeneyRedHat:
Changes LGTM
but tests are all kinds of angry

PR #9112, opened 2021-01-26
[CI:DOCS]Correct static API html path

baude:
Correct the path from the reference.rst to the static api documentation.


TomSweeneyRedHat:
LGTM

cevich:
***NOTE***: The v3.0 branch is not setup in github or RTD for a specialized `swagger.yml`.  However, the capabilities exist to do this if needed, but the effort needs to be coordinated.

PR #9113, opened 2021-01-26
Cirrus: Collect ginkgo node logs artifacts

cevich:
In rare cases, it's possible for one of the ginkgo processes to "hang".
When this occurs, the main output will contain this message:

``Ginkgo timed out waiting for all parallel nodes to report``

The only way to debug this was to look through concatenated printing
of the ginkgo node logs.  This is a tedious and daunting task,
requiring special search knowledge, facing a "wall of text".

Simplify the situation by collecting the node logs separately, as
individual files in a cirrus-artifact.  In this way, it's faster to
figure out which test "hung" by examining each log individually.  The
log file which does not have a pass/fail summary at the end,
indicates the last test hung (for whatever reason), and includes it's
output (if any).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -433,7 +433,6 @@ apiv2_test_task:
             path: ./*.html
             type: text/html
         package_versions_script: '$SCRIPT_BASE/logcollector.sh packages'
-        ginkgo_node_logs_script: '$SCRIPT_BASE/logcollector.sh ginkgo'
         df_script: '$SCRIPT_BASE/logcollector.sh df'
         audit_log_script: '$SCRIPT_BASE/logcollector.sh audit'
         journal_script: '$SCRIPT_BASE/logcollector.sh journal'
@@ -478,7 +477,11 @@ local_integration_test_task: &local_integration_test_task
     gopath_cache: *ro_gopath_cache
     setup_script: *setup
     main_script: *main
-    always: *logs_artifacts
+    always: &int_logs_artifacts
+        <<: *logs_artifacts
+        ginkgo_node_logs_artifacts:
+            path: ./test/e2e/ginkgo-node-*.log
+            type: text/plain
 
 
 # Nearly identical to `local_integration_test` except all operations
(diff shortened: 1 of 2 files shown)


cevich:
Ref.[ example `ginkgo_node_logs` item now listed under `artifacts` in integration testing tasks.](https://cirrus-ci.com/task/5809146270318592)

rhatdan:
LGTM

cevich:
@edsantiago PTAL...should be easy one.

edsantiago:
Unfortunate that Cirrus does not offer [an option to flatten the directory structure](https://github.com/cirruslabs/cirrus-ci-docs/issues/548).

Do you know why logs weren't collected for any of the `remote` jobs? Should they have been?

edsantiago:
I will assume that ginkgo logs don't exist for `remoteintegration`: https://cirrus-ci.com/task/5075101865476096?command=ginkgo_node_logs#L1
```
cat: '/var/tmp/go/src/github.com/containers/podman/test/e2e/ginkgo-node-*.log': No such file or directory
```
/lgtm
/hold

cevich:
> Unfortunate that Cirrus does not offer an option to flatten the directory structure.

No the only option is to move the around relative to `$CIRRUS_WORKING_DIR` :disappointed:

> I will assume that ginkgo logs don't exist for remoteintegration

Interesting...I believe it's because there is no `-debug` flag used for that target's command-line.  I can add that in a separate PR if you think it's important.

edsantiago:
Already done, #9200

cevich:
Oh!  Correction: This is because `ginkgo-remote` does not use multiple threads (nodes) for testing.  So the log file would be 100% identical to what you see for the regular output.

cevich:
Rebased w/ Ed's fix.

edsantiago:
Confirmed: [remote ginkgo logs now stored as artifacts](https://cirrus-ci.com/task/6013439815450624)

edsantiago (approved):

PR #9115, opened 2021-01-26
Switch podman image push handlers to use abi

rhatdan:
Change API Handlers to use the same functions that the
local podman uses.

At the same time:

Cleanup and pass proper bindings.  Remove cli options from
podman-remote push.  Cleanup manifest push.

### DIFF
--- a/docs/source/markdown/podman-push.1.md
+++ b/docs/source/markdown/podman-push.1.md
@@ -91,7 +91,7 @@ solely for scripting compatibility.
 #### **--format**, **-f**=*format*
 
 Manifest Type (oci, v2s1, or v2s2) to use when pushing an image to a directory using the 'dir:' transport (default is manifest type of source)
-Note: This flag can only be set when using the **dir** transport
+Note: This flag can only be set when using the **dir** transport. (Not available for remote commands)
 
 #### **--quiet**, **-q**
 
@@ -99,11 +99,11 @@ When writing the output image, suppress progress output
 
 #### **--remove-signatures**
 
-Discard any pre-existing signatures in the image
+Discard any pre-existing signatures in the image. (Not available for remote commands)
 
 #### **--sign-by**=*key*
 
-Add a signature at the destination using the specified key
+Add a signature at the destination using the specified key. (Not available for remote commands)
 
 #### **--tls-verify**=*true|false*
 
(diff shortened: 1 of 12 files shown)


vrothberg (commented):
Please use the `X-Registry-Auth` header to pass credentials as other endpoints do.  Passing the credentials via the parameters may be a security issue. There are convenience functions in `pkg/auth` to easily add and extract credentials on the client and server.

pkg/api/handlers/libpod/manifests.go
```diff
@@ -148,7 +148,9 @@ func ManifestPush(w http.ResponseWriter, r *http.Request) {
 		All         bool   `schema:"all"`
 		Destination string `schema:"destination"`
 		Format      string `schema:"format"`
+		Password    string `schema:"password"`
 		TLSVerify   bool   `schema:"tlsVerify"`
+		Username    string `schema:"username"`
```
vrothberg, line 153: Still some left.

vrothberg: I think the changes in this file can be reverted.

pkg/api/server/register_manifest.go
```diff
@@ -151,6 +151,14 @@ func (s *APIServer) registerManifestHandlers(r *mux.Router) error {
...
 	//    description: push all images
 	//    type: boolean
+	//  - in: query
+	//    name: username
+	//    type: string
+	//    description: username for authenticating against the registry.
+	//  - in: query
+	//    name: password
+	//    type: string
+	//    description: password for authenticating against the registry.
```
vrothberg, line 161: And these.

rhatdan:
@vrothberg @mheon PTAL
This is passing tests, and should be ready to merge.

vrothberg (commented):
LGTM, thanks @rhatdan !

docs/source/markdown/podman-push.1.md
```diff
@@ -91,19 +91,19 @@ solely for scripting compatibility.
 #### **--format**, **-f**=*format*
 
 Manifest Type (oci, v2s1, or v2s2) to use when pushing an image to a directory using the 'dir:' transport (default is manifest type of source)
-Note: This flag can only be set when using the **dir** transport
+Note: This flag can only be set when using the **dir** transport. (Not available for remote commands)
```
mheon, line 94: Suggest rewording all of these to "Not available with the remote client"

rhatdan: Mheon I am going to merge this and then change the man pages as a docs only change.

Issue #8577, opened 2020-12-03
podman docker swagger mismatches


tmds:
I'm running podman 2.1.1 and tried using it with a client generated from vendor/github.com/docker/docker/api/swagger.yaml.

These differences came up:

Command is not a string, but an array of strings.
```
        "Command": [
            "/bin/bash"
        ],
```

ports don't contain required PrivatePort and Type:
```
        "Ports": [
            {
                "hostPort": 3000,
                "containerPort": 3000,
                "protocol": "tcp",
                "hostIP": ""
            }
```

mounts is not an array of type Mount, but of strings.
```
        "Mounts": [
            "/root"
        ],
```

tmds:
I'm using `/containers/json` on `systemctl` `--user` `podman.socket`.

I'm using the docker swagger file. I tried using the https://storage.googleapis.com/libpod-master-releases/swagger-latest-master.yaml but the tool I'm using has an issue with it.

mheon:
@baude @jwhonce PTAL

jwhonce:
@tmds Which tool are you using? We've put a lot of work in the swagger  support and I am sorry to hear you are having issues.

tmds:
@jwhonce I'm using NSwag to generate a C# client. It may be a limitation of the tool, rather than an issue with the podman swagger file. I'll cc you on the issue in the NSwag repo.

tmds:
@jwhonce this is what I need to change to the swagger file to make NSwag happy:
```diff
diff --git a/swagger-latest-master.yaml b/swagger-latest-master.yaml
index d9b199f..1891a02 100644
--- a/swagger-latest-master.yaml
+++ b/swagger-latest-master.yaml
@@ -10300,7 +10300,8 @@ paths:
       - application/json
       responses:
         "200":
-          $ref: '#/definitions/IDResponse'
+          schema:
+            $ref: '#/definitions/IDResponse'
         "400":
           $ref: '#/responses/BadParamError'
         "404":
@@ -10329,7 +10330,8 @@ paths:
       - application/json
       responses:
         "200":
-          $ref: '#/definitions/IDResponse'
+          schema:
+            $ref: '#/definitions/IDResponse'
         "404":
           $ref: '#/responses/NoSuchManifest'
         "409":
@@ -10383,7 +10385,8 @@ paths:
       - application/json
       responses:
         "200":
-          $ref: '#/definitions/IDResponse'
+          schema:
+            $ref: '#/definitions/IDResponse'
         "400":
[...]

PR #9117, opened 2021-01-27
[NO TESTS NEEDED] swagger: add missing schema properties

tmds:
@jwhonce ptal.

### DIFF
--- a/pkg/api/server/register_manifest.go
+++ b/pkg/api/server/register_manifest.go
@@ -30,7 +30,8 @@ func (s *APIServer) registerManifestHandlers(r *mux.Router) error {
 	//   description: add all contents if given list
 	// responses:
 	//   200:
-	//     $ref: "#/definitions/IDResponse"
+	//     schema:
+	//       $ref: "#/definitions/IDResponse"
 	//   400:
 	//     $ref: "#/responses/BadParamError"
 	//   404:
@@ -96,7 +97,8 @@ func (s *APIServer) registerManifestHandlers(r *mux.Router) error {
 	//      $ref: "#/definitions/ManifestAddOpts"
 	// responses:
 	//   200:
-	//     $ref: "#/definitions/IDResponse"
+	//     schema:
+	//       $ref: "#/definitions/IDResponse"
 	//   404:
 	//     $ref: "#/responses/NoSuchManifest"
 	//   409:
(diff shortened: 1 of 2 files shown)


tmds:
cc @baude

zhangguanzhang:
thanks for contribute, and your commit messages should change to include `Sign-off-by $name ..` and `git push -f`

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/8577

rhatdan:
Do we need this change in all of the register_*go files?

tmds:
> Do we need this change in all of the register_*go files?

@rhatdan I think this covers all needed cases. Do you have some specific section in mind that still should be patched?

Is this good to merge?

rhatdan:
Could you repush to see if we get by the smoke test.

baude:
I think thew NO_TESTS_NEEDED needs to be in the body of your commit message.  Speaking of which, can you provide a commit message?

baude:
LGTM other than my comments about commit messages

jwhonce:
@tmds Please let me know what tool you are using, or why you need these changes? I'm curious if we need to add that property to all our responses.

tmds:
>  Please let me know what tool you are using, or why you need these changes? 

I'm using a .NET specific tool: https://github.com/RicoSuter/NSwag

Without the `schema` properties it considers the swagger invalid.
I'm not a swagger expert. Based on the docker swagger, and skimming through the swagger spec, I think the tool is probably right.

> I'm curious if we need to add that property to all our responses.

The other responses already include it, for example:
```yaml
  BadParamError:
    description: Bad parameter in request
    schema:
      properties:
        cause:
          description: API root cause formatted for automated parsing
          example: API root cause
          type: string
          x-go-name: Because
        message:
          description: human error message, formatted for a human to read
          example: human error message
          type: string
          x-go-name: Message
        response:
          description: http response code
          format: int64
          type: integer
          x-go-name: ResponseCode
      type: object

```

tmds:
Is this good to merge?

mheon:
Wait, CI is red
/hold

mheon:
@tmds I think you need a force-push to make the [NO TESTS NEEDED] label work - I recommend changing something trivial (punctuation in a description, maybe) and force-pushing an amended commit

TomSweeneyRedHat:
LGTM
assuming happy tests

rhatdan:
@tmds Could you rebase your PR and repush, so that the tests will notice the [NO TESTS NEEDED] flag and we can get this in.

tmds:
I've rebased.

PR #9121, opened 2021-01-27
[NO TESTS NEEDED] swagger: remove name wildcards

tmds:
@jwhonce @baude ptal

### DIFF
--- a/hack/swagger-check
+++ b/hack/swagger-check
@@ -178,9 +178,10 @@ sub handle_handle {
         or die "$ME: $path:$.: Cannot grok '$line'\n";
     my $endpoint = $1;
 
-    # FIXME: in older code, '{name:..*}' meant 'nameOrID'. As of 2020-02
-    # it looks like most of the '{name:..*}' entries are gone, except for one.
-###FIXME-obsolete?    $endpoint =~ s|\{name:\.\.\*\}|{nameOrID}|;
+    # Some function declarations require an argument of the form '{name:.*}'
+    # but the swagger (which gets derived from the comments) should not
+    # include them. Normalize all such args to just '{name}'.
+    $endpoint =~ s/\{name:\.\*\}/\{name\}/;
 
     # e.g. /auth, /containers/*/rename, /distribution, /monitor, /plugins
     return 1 if $line =~ /\.UnsupportedHandler/;
--- a/pkg/api/server/register_generate.go
+++ b/pkg/api/server/register_generate.go
@@ -8,7 +8,7 @@ import (
 )
 
 func (s *APIServer) registerGenerateHandlers(r *mux.Router) error {
-	// swagger:operation GET /libpod/generate/{name:.*}/systemd libpod GenerateSystemdLibpod
+	// swagger:operation GET /libpod/generate/{name}/systemd libpod GenerateSystemdLibpod
 	// ---
 	// tags:
 	//  - containers
@@ -17,7 +17,7 @@ func (s *APIServer) registerGenerateHandlers(r *mux.Router) error {
 	// description: Generate Systemd Units based on a pod or container.
 	// parameters:
 	//  - in: path
-	//    name: name:.*
+	//    name: name
 	//    type: string
 	//    required: true
 	//    description: Name or ID of the container or pod.
--- a/pkg/api/server/register_healthcheck.go
+++ b/pkg/api/server/register_healthcheck.go
@@ -8,7 +8,7 @@ import (
 )
 
 func (s *APIServer) registerHealthCheckHandlers(r *mux.Router) error {
-	// swagger:operation GET /libpod/containers/{name:.*}/healthcheck libpod ContainerHealthcheckLibpod
+	// swagger:operation GET /libpod/containers/{name}/healthcheck libpod ContainerHealthcheckLibpod
 	// ---
 	// tags:
 	//  - containers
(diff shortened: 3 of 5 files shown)


rhatdan:
Thanks @tmds 
I believe the wild cards were added for a reason, but I will let @jwhonce or @baude comment on that.

Please sign your commits.

git commit -a --amend -s
git push --force

tmds:
> I believe the wild cards were added for a reason, but I will let @jwhonce or @baude comment on that.

@jwhonce @baude do the wild cards serve some purpose?

tmds:
> @jwhonce @baude do the wild cards serve some purpose?

@jwhonce @baude ping.

tmds:
I've added the 'NO TESTS' and rebased.

Can this be merged, unless someone remembers why the wildcards are there?

rhatdan:
Seems like all of the tests do not like this change?

mheon:
The swagger errors look legitimate. I'm not sure about the rest.

tmds:
I think I'm to eagerly removing these wild cards.

I care about these:
```
// swagger:operation GET /libpod/generate/{name:.*}/systemd libpod libpodGenerateSystemd
```
And probably the test failures are due to changing these:
```
r.HandleFunc(VersionedPath("/libpod/generate/{name:.*}/systemd"), s.APIHandler(libpod.GenerateSystemd)).Methods(http.MethodGet)
```

tmds:
@mheon I've limited my change to comments. Can you take a look at the CI tests? I don't understand what they're trying to tell me.

tmds:
@mheon I've removed the changes to `vendor/github.com/docker/docker/api/types/types.go`.

mheon:
Ack. We should be good to merge, then.

mheon:
@jwhonce PTAL

tmds:
@jwhonce @mheon can you help me understand what CI is unhappy about?

edsantiago:
@tmds your change is incomplete. The CI failure is my last-minute validation script, it is complaining that the parameters are inconsistent between comments and code. For instance:
```
- 	// swagger:operation GET /images/{name}/get compat exportImage
+ 	// ................. ... /images/{name:.*}/get ......
  	r.Handle(VersionedPath("/images/{name:.*}/get"), s.APIHandler(compat.ExportImage)).Methods(http.MethodGet)
```
That is saying that the swagger comment claims `{name}`, but just a few lines below the actual code is registering `{name:.*}`. My code is trying to enforce that the swagger comment be identical to the `r.Handle()` code.

tmds:
> My code is trying to enforce that the swagger comment be identical to the r.Handle() code.

@edsantiago I'm trying to get rid of the `.*` that end up in the swagger. I think I need to remove them from the comments, but keep them in the `r.Handle` calls. I think we need to update the expectation of the test. Does that make sense?

edsantiago:
I don't know anything at all about swagger; I just know that for a long time we had inconsistencies between the code and the documentation, and my script fixed those. I will do whatver @jwhonce and @baude recommend.

tmds:
@jwhonce @baude I removed the `.*` from the comments, but kept them in the r.Handle calls. I think we need to update the expectation of the test. Does that make sense?

tmds:
ping @jwhonce @baude

jwhonce:
@tdms I've been trying to remember why we added the wildcards. There was a legal character in image names that we needed a greedy capture to include and gorilla mux by default did not include. Caused unexpected 404's as the there was no match.

The change propagated into comments so everything matched. 

Is this a stylistic change you want or is this blocking some tool?

tmds:
> Is this a stylistic change you want or is this blocking some tool?

It breaks code generation using https://github.com/RicoSuter/NSwag.
Can we remove them from the comments, so they don't end up in the swagger?

jwhonce:
@tmds I'll take a look at our validator to see how we can extend it to cover this use case.

edsantiago:
@tmds my apologies, I didn't realize you were waiting on me (because I don't understand this conversation thread at all).

If the goal is to remove dot-star from the comments, but leave it in the declaration thingies, please apply the following:
```diff
diff --git a/hack/swagger-check b/hack/swagger-check
index d20318305..45832d77d 100755
--- a/hack/swagger-check
+++ b/hack/swagger-check
@@ -186,9 +186,10 @@ sub handle_handle {
         or die "$ME: $path:$.: Cannot grok '$line'\n";
     my $endpoint = $1;

-    # FIXME: in older code, '{name:..*}' meant 'nameOrID'. As of 2020-02
-    # it looks like most of the '{name:..*}' entries are gone, except for one.
-###FIXME-obsolete?    $endpoint =~ s|\{name:\.\.\*\}|{nameOrID}|;
+    # Some function declarations require an argument of the form '{name:.*}'
+    # but the swagger comment itself can't have ':.*', some sort of tool
+    # thingy chokes on them (#9121). Normalize all such args to just '{name}'.
+    $endpoint =~ s/\{name:\.\*\}/\{name\}/;

     # e.g. /auth, /containers/*/rename, /distribution, /monitor, /plugins
     return 1 if $line =~ /\.UnsupportedHandler/;
```
Please revise the comment to be more meaningful to someone swagger-literate (and perhaps also be readable by those of us who are clueless).

Big request: you need to rebase, but please wait until #9952 merges before you re-push. Thank you.

tmds:
@edsantiago I've applied the changes. Let's see what CI makes of it.

edsantiago:
It's going to pass. BTW you can run the test manually in your own checked-out tree:
```console
$ ./hack/swagger-check
```
It runs in less than a second, and requires only `perl` and `File::Find`, which I believe are standard in all distros I know.

edsantiago:
LGTM but I'd like @jwhonce to have the final merge say-so. @tmds thank you for your patience!

tmds:
@edsantiago thanks for your assistance!

PR #9123, opened 2021-01-27
[CI:DOCS] swagger: add operationIds that match with docker

tmds:
@jwhonce @baude ptal.

I only added operationIds to to match with the docker ones.
You may want to add operationIds to the operations specific to podman too.

### DIFF
--- a/pkg/api/server/register_archive.go
+++ b/pkg/api/server/register_archive.go
@@ -8,10 +8,9 @@ import (
 )
 
 func (s *APIServer) registerArchiveHandlers(r *mux.Router) error {
-	// swagger:operation PUT /containers/{name}/archive compat putArchive
+	// swagger:operation PUT /containers/{name}/archive compat PutContainerArchive
 	// ---
 	//  summary: Put files into a container
-	//  operationId: PutContainerArchive
 	//  description: Put a tar archive of files into a container
 	//  tags:
 	//   - containers (compat)
@@ -53,10 +52,9 @@ func (s *APIServer) registerArchiveHandlers(r *mux.Router) error {
 	//    500:
 	//      $ref: "#/responses/InternalError"
 
-	// swagger:operation GET /containers/{name}/archive compat getArchive
+	// swagger:operation GET /containers/{name}/archive compat ContainerArchive
 	// ---
 	//  summary: Get files from a container
-	//  operationId: ContainerArchive
 	//  description: Get a tar archive of files from a container
 	//  tags:
 	//   - containers (compat)
--- a/pkg/api/server/register_archive.go
+++ b/pkg/api/server/register_archive.go
@@ -11,6 +11,7 @@ func (s *APIServer) registerArchiveHandlers(r *mux.Router) error {
 	// swagger:operation PUT /containers/{name}/archive compat putArchive
 	// ---
 	//  summary: Put files into a container
+	//  operationId: PutContainerArchive
 	//  description: Put a tar archive of files into a container
 	//  tags:
 	//   - containers (compat)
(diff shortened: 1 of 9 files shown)


mheon:
I think you need to add `[CI:DOCS]` to the title to disable full CI, since this is a docs-only change. You will probably need a force-push after that.

Otherwise LGTM

jwhonce:
@tmds If you add [NO TESTS NEEDED] to your commit message, the merge bot will not flag your PR as missing tests.

@baude This PR does show an issue where we have an `operationalId` conflict between the libpod API and the docker API, `SystemVersion` is duplicated. Opinion?

tmds:
> @baude This PR does show an issue where we have an operationalId conflict between the libpod API and the docker API, SystemVersion is duplicated. Opinion?

@jwhonce I don't see how there is a duplicate. It looks like distinct swagger sections?
```
        // swagger:operation GET /version compat CompatSystemVersion
        // ---
        // summary: Component Version information
+       // operationId: SystemVersion
        // tags:
        // - system (compat)
        // produces:
        // - application/json
        // responses:
        //   200:
        //    $ref: "#/responses/Version"
        r.Handle("/version", s.APIHandler(compat.VersionHandler)).Methods(http.MethodGet)
        r.Handle(VersionedPath("/version"), s.APIHandler(compat.VersionHandler)).Methods(http.MethodGet)
        // swagger:operation GET /libpod/version libpod SystemVersion
```

It does look like there is a duplicate for `ContainerChanges`:
```
        // swagger:operation GET /containers/{name}/changes libpod libpodChangesContainer
        // swagger:operation GET /libpod/containers/{name}/changes compat changesContainer
        // ---
        // tags:
        //   - containers
        //   - containers (compat)
        // summary: Report on changes to container's filesystem; adds, deletes or modifications.
+       // operationId: ContainerChanges
        // description: |
```
Is the fix to have separate swagger sections for these operations, and only include the `operationId` in one?

tmds:
@jwhonce can you take a look at my previous comment?

jwhonce:
@tmds I saw the conflict between your PR and https://github.com/containers/podman/blob/master/pkg/api/server/register_version.go#L23 regarding the operationId SystemVersion .

Since we have endpoints for the compatibility and libpod in one API we need to namespace them. To fix this across the API, we would need to remove Compat from the Compat<opertionalId> on the `swagger:operation` lines  and add Libpod<operationId> to the libpod operations.  If we make those changes, then the swagger  processor should gives us your proposed operationId's without having to add `operationId` to override the one given on the `swagger:operation` line. 

/cc @baude @rhatdan

tmds:
@jwhonce based on your comment, I've moved the `operationIds` to the end of the `swagger:operation` lines. Let me know if I should make additional changes.

tmds:
> Since we have endpoints for the compatibility and libpod in one API we need to namespace them. To fix this across the API, we would need to remove Compat from the Compat on the swagger:operation lines and add Libpod to the libpod operations. If we make those changes

Thinking out loud: you may want to generate separate swagger files, one for libpod, and one for compat. Both of them can use the same operationIds.

jwhonce:
@tmds I need to think on that suggestion. 

Immediate ideas:
 - It would fix this issue and supporting /swagger and /libpod/swagger endpoints is trivial.
 - Maintenance would be a bit painful since "like" endpoints would be in different files, /containers/json vs /libpod/containers/json
 - Generating a client using both API's would bring us back to this issue. Is that a real concern?

/cc @mheon @baude

vrothberg:
Friendly ping

rhatdan:
@jwhonce @baude @mheon What do we want to do with this?

jwhonce:
@rhatdan See https://github.com/containers/podman/pull/9944

jwhonce:
@tmds I included your commits in https://github.com/containers/podman/pull/9944 and rebased to master.

PR #9125, opened 2021-01-27
Implement Secrets

ashley-cui:
Implement podman secret create, inspect, ls, rm
Implement podman run/create --secret
Secrets are blobs of data that are sensitive.
Currently, the only secret driver supported is filedriver, which means creating a secret stores it in base64 unencrypted in a file.
After creating a secret, a user can use the --secret flag to expose the secret inside the container at /run/secrets/[secretname]
This secret will not be commited to an image on a podman commit

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -603,6 +603,14 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	)
 	_ = cmd.RegisterFlagCompletionFunc(sdnotifyFlagName, AutocompleteSDNotify)
 
+	secretFlagName := "secret"
+	createFlags.StringArrayVar(
+		&cf.Secrets,
+		secretFlagName, []string{},
+		"Add secret to container",
+	)
+	_ = cmd.RegisterFlagCompletionFunc(secretFlagName, AutocompleteSecrets)
+
 	securityOptFlagName := "security-opt"
 	createFlags.StringArrayVar(
 		&cf.SecurityOpt,
--- a/cmd/podman/secrets/create.go
+++ b/cmd/podman/secrets/create.go
@@ -1,0 +1,80 @@
+package secrets
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/containers/common/pkg/completion"
+	"github.com/containers/podman/v2/cmd/podman/common"
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/spf13/cobra"
+)
+
+var (
+	createCmd = &cobra.Command{
+		Use:   "create [options] SECRET FILE|-",
+		Short: "Create a new secret",
+		Long:  "Create a secret. Input can be a path to a file or \"-\" (read from stdin). Default driver is file (unencrypted).",
+		RunE:  create,
+		Args:  cobra.ExactArgs(2),
+		Example: `podman secret create mysecret /path/to/secret
+		printf "secretdata" | podman secret create mysecret -`,
+		ValidArgsFunction: common.AutocompleteSecretCreate,
+	}
+)
+
+var (
+	createOpts = entities.SecretCreateOptions{}
+)
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: createCmd,
+		Parent:  secretCmd,
+	})
+
+	flags := createCmd.Flags()
+
+	driverFlagName := "driver"
+	flags.StringVar(&createOpts.Driver, driverFlagName, "file", "Specify secret driver")
+	_ = createCmd.RegisterFlagCompletionFunc(driverFlagName, completion.AutocompleteNone)
+}
+
+func create(cmd *cobra.Command, args []string) error {
+	name := args[0]
+
+	var err error
+	path := args[1]
+
+	var reader io.Reader
+	if path == "-" || path == "/dev/stdin" {
+		stat, err := os.Stdin.Stat()
+		if err != nil {
+			return err
+		}
+		if (stat.Mode() & os.ModeNamedPipe) == 0 {
+			return errors.New("if `-` is used, data must be passed into stdin")
+
+		}
+		reader = os.Stdin
+	} else {
+		file, err := os.Open(path)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+		reader = file
+	}
+
+	report, err := registry.ContainerEngine().SecretCreate(context.Background(), name, reader, createOpts)
+	if err != nil {
+		return err
+	}
+	fmt.Println(report.ID)
+	return nil
+}
--- a/cmd/podman/secrets/list.go
+++ b/cmd/podman/secrets/list.go
@@ -1,0 +1,99 @@
+package secrets
+
+import (
+	"context"
+	"html/template"
+	"os"
+	"text/tabwriter"
+	"time"
+
+	"github.com/containers/common/pkg/completion"
+	"github.com/containers/common/pkg/report"
+	"github.com/containers/podman/v2/cmd/podman/common"
+	"github.com/containers/podman/v2/cmd/podman/parse"
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/cmd/podman/validate"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/docker/go-units"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	lsCmd = &cobra.Command{
+		Use:               "ls [options]",
+		Aliases:           []string{"list"},
+		Short:             "List secrets",
+		RunE:              ls,
+		Example:           "podman secret ls",
+		Args:              validate.NoArgs,
+		ValidArgsFunction: completion.AutocompleteNone,
+	}
+	listFlag = listFlagType{}
+)
+
+type listFlagType struct {
+	format    string
+	noHeading bool
+}
+
+func init() {
+	registry.Commands = append(registry.Commands, registry.CliCommand{
+		Mode:    []entities.EngineMode{entities.ABIMode, entities.TunnelMode},
+		Command: lsCmd,
+		Parent:  secretCmd,
+	})
+
+	flags := lsCmd.Flags()
+	formatFlagName := "format"
+	flags.StringVar(&listFlag.format, formatFlagName, "{{.ID}}\t{{.Name}}\t{{.Driver}}\t{{.CreatedAt}}\t{{.UpdatedAt}}\t\n", "Format volume output using Go template")
+	_ = lsCmd.RegisterFlagCompletionFunc(formatFlagName, common.AutocompleteJSONFormat)
+
+}
+
+func ls(cmd *cobra.Command, args []string) error {
+	responses, err := registry.ContainerEngine().SecretList(context.Background())
+	if err != nil {
+		return err
+	}
+	listed := make([]*entities.SecretListReport, 0, len(responses))
+	for _, response := range responses {
+		listed = append(listed, &entities.SecretListReport{
+			ID:        response.ID,
+			Name:      response.Spec.Name,
+			CreatedAt: units.HumanDuration(time.Since(response.CreatedAt)) + " ago",
+			UpdatedAt: units.HumanDuration(time.Since(response.UpdatedAt)) + " ago",
+			Driver:    response.Spec.Driver.Name,
+		})
+
+	}
+	return outputTemplate(cmd, listed)
+}
+
+func outputTemplate(cmd *cobra.Command, responses []*entities.SecretListReport) error {
+	headers := report.Headers(entities.SecretListReport{}, map[string]string{
+		"CreatedAt": "CREATED",
+		"UpdatedAt": "UPDATED",
+	})
+
+	row := report.NormalizeFormat(listFlag.format)
+	format := parse.EnforceRange(row)
+
+	tmpl, err := template.New("list secret").Parse(format)
+	if err != nil {
+		return err
+	}
+	w := tabwriter.NewWriter(os.Stdout, 12, 2, 2, ' ', 0)
+	defer w.Flush()
+
+	if cmd.Flags().Changed("format") && !parse.HasTable(listFlag.format) {
+		listFlag.noHeading = true
+	}
+
+	if !listFlag.noHeading {
+		if err := tmpl.Execute(w, headers); err != nil {
+			return errors.Wrapf(err, "failed to write report column headers")
+		}
+	}
+	return tmpl.Execute(w, responses)
+}
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -825,6 +825,16 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret. Can be specified multiple times.
+
+A secret is a blob of sensitive data which a container needs at runtime but
+should not be stored in the image or in source control, such as usernames and passwords,
+TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
+
+Secrets are managed using the `podman secret` command.
+
 #### **--security-opt**=*option*
 
 Security Options
@@ -1277,7 +1287,7 @@ b
 NOTE: Use the environment variable `TMPDIR` to change the temporary storage location of downloaded container images. Podman defaults to use `/var/tmp`.
 
 ## SEE ALSO
-**podman**(1), **podman-save**(1), **podman-ps**(1), **podman-attach**(1), **podman-pod-create**(1), **podman-port**(1), **podman-kill**(1), **podman-stop**(1),
+**podman**(1), **podman-secret**(1), **podman-save**(1), **podman-ps**(1), **podman-attach**(1), **podman-pod-create**(1), **podman-port**(1), **podman-kill**(1), **podman-stop**(1),
 **podman-generate-systemd**(1) **podman-rm**(1), **subgid**(5), **subuid**(5), **containers.conf**(5), **systemd.unit**(5), **setsebool**(8), **slirp4netns**(1), **fuse-overlayfs**(1), **proc**(5)**.
 
 ## HISTORY
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -877,6 +877,16 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret. Can be specified multiple times.
+
+A secret is a blob of sensitive data which a container needs at runtime but
+should not be stored in the image or in source control, such as usernames and passwords,
+TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
+
+Secrets are managed using the `podman secret` command
+
 #### **--security-opt**=*option*
 
 Security Options
--- a/docs/source/markdown/podman-secret-create.1.md
+++ b/docs/source/markdown/podman-secret-create.1.md
@@ -1,0 +1,43 @@
+% podman-secret-create(1)
+
+## NAME
+podman\-secret\-create - Create a new secret
+
+## SYNOPSIS
+**podman secret create** [*options*] *name* *file|-*
+
+## DESCRIPTION
+
+Creates a secret using standard input or from a file for the secret content.
+
+Create accepts a path to a file, or `-`, which tells podman to read the secret from stdin
+
+A secret is a blob of sensitive data which a container needs at runtime but
+should not be stored in the image or in source control, such as usernames and passwords,
+TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
+
+Secrets will not be commited to an image with `podman commit`, and will not be in the archive created by a `podman export`
+
+## OPTIONS
+
+#### **--driver**=*driver*
+
+Specify the secret driver (default **file**, which is unencrypted).
+
+#### **--help**
+
+Print usage statement.
+
+## EXAMPLES
+
+```
+$ podman secret create my_secret ./secret.json
+$ podman secret create --driver=file my_secret ./secret.json
+$ printf <secret> | podman secret create my_secret -
+```
+
+## SEE ALSO
+podman-secret (1)
+
+## HISTORY
+January 2021, Originally compiled by Ashley Cui <acui@redhat.com>
--- a/docs/source/markdown/podman-secret-inspect.1.md
+++ b/docs/source/markdown/podman-secret-inspect.1.md
@@ -1,0 +1,38 @@
+% podman-secret-inspect(1)
+
+## NAME
+podman\-secret\-inspect - Display detailed information on one or more secrets
+
+## SYNOPSIS
+**podman secret inspect** [*options*] *secret* [...]
+
+## DESCRIPTION
+
+Inspects the specified secret.
+
+By default, this renders all results in a JSON array. If a format is specified, the given template will be executed for each result.
+Secrets can be queried individually by providing their full name or a unique partial name.
+
+## OPTIONS
+
+#### **--format**=*format*
+
+Format secret output using Go template.
+
+#### **--help**
+
+Print usage statement.
+
+
+## EXAMPLES
+
+```
+$ podman secret inspect mysecret
+$ podman secret inspect --format "{{.Name} {{.Scope}}" mysecret
+```
+
+## SEE ALSO
+podman-secret(1)
+
+## HISTORY
+January 2021, Originally compiled by Ashley Cui <acui@redhat.com>
--- a/docs/source/markdown/podman-secret-ls.1.md
+++ b/docs/source/markdown/podman-secret-ls.1.md
@@ -1,0 +1,30 @@
+% podman-secret-ls(1)
+
+## NAME
+podman\-secret\-ls - List all available secrets
+
+## SYNOPSIS
+**podman secret ls** [*options*]
+
+## DESCRIPTION
+
+Lists all the secrets that exist. The output can be formatted to a Go template using the **--format** option.
+
+## OPTIONS
+
+#### **--format**=*format*
+
+Format secret output using Go template.
+
+## EXAMPLES
+
+```
+$ podman secret ls
+$ podman secret ls --format "{{.Name}}"
+```
+
+## SEE ALSO
+podman-secret(1)
+
+## HISTORY
+January 2021, Originally compiled by Ashley Cui <acui@redhat.com>
--- a/docs/source/markdown/podman-secret-rm.1.md
+++ b/docs/source/markdown/podman-secret-rm.1.md
@@ -1,0 +1,33 @@
+% podman-secret-rm(1)
+
+## NAME
+podman\-secret\-rm - Remove one or more secrets
+
+## SYNOPSIS
+**podman secret rm** [*options*] *secret* [...]
+
+## DESCRIPTION
+
+Removes one or more secrets.
+
+## OPTIONS
+
+#### **--all**, **-a**
+
+Remove all existing secrets.
+
+#### **--help**
+
+Print usage statement.
+
+## EXAMPLES
+
+```
+$ podman secret rm mysecret1 mysecret2
+```
+
+## SEE ALSO
+podman-secret(1)
+
+## HISTORY
+January 2021, Originally compiled by Ashley Cui <acui@redhat.com>
--- a/docs/source/markdown/podman-secret.1.md
+++ b/docs/source/markdown/podman-secret.1.md
@@ -1,0 +1,25 @@
+% podman-secret(1)
+
+## NAME
+podman\-secret - Manage podman secrets
+
+## SYNOPSIS
+**podman secret** *subcommand*
+
+## DESCRIPTION
+podman secret is a set of subcommands that manage secrets.
+
+## SUBCOMMANDS
+
+| Command | Man Page                                               | Description                                            |
+| ------- | ------------------------------------------------------ | ------------------------------------------------------ |
+| create  | [podman-secret-create(1)](podman-secret-create.1.md)   | Create a new secret                                    |
+| inspect | [podman-secret-inspect(1)](podman-secret-inspect.1.md) | Display detailed information on one or more secrets    |
+| ls      | [podman-secret-ls(1)](podman-secret-ls.1.md)           | List all available secrets                             |
+| rm      | [podman-secret-rm(1)](podman-secret-rm.1.md)           | Remove one or more secrets                             |
+
+## SEE ALSO
+podman(1)
+
+## HISTORY
+January 2021, Originally compiled by Ashley Cui <acui@redhat.com>
--- a/libpod/container_config.go
+++ b/libpod/container_config.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"time"
 
+	"github.com/containers/common/pkg/secrets"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/podman/v2/pkg/namespaces"
 	"github.com/containers/storage"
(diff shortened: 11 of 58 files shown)


ashley-cui:
This needs a new common, and also shouldn't be merged until after 3.0

cmd/podman/common/create.go
```diff
@@ -603,6 +603,14 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	)
 	_ = cmd.RegisterFlagCompletionFunc(sdnotifyFlagName, AutocompleteSDNotify)
 
+	secretFlagName := "secret"
+	createFlags.StringArrayVar(
+		&cf.Secrets,
+		secretFlagName, []string{},
+		"Add secret to contianer",
```
TomSweeneyRedHat, line 610:
```suggestion
		"Add secret to container",
```

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+	createCmd = &cobra.Command{
+		Use:   "create SECRET",
+		Short: "Create a new secret",
+		Long:  "Create a secret. Default driver is file (unencrypted).",
+		RunE:  create,
+		Args:  cobra.ExactArgs(2),
+		Example: `podman secret create mysecret /path/to/secret
+		printf "secretdata" | podman secret create mysecret -`,
```
TomSweeneyRedHat, line 24: I didn't test, but I'm not sure the example here is working as you expect.

docs/source/markdown/podman-create.1.md
```diff
@@ -824,6 +824,12 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret.
+
+Secrets are managed using the `podman secret` command
```
TomSweeneyRedHat, line 831: It would be good to have a line or two explaining what a secret is or does.
```suggestion
Secrets are managed using the `podman secret` command.
```

docs/source/markdown/podman-run.1.md
```diff
@@ -876,6 +876,12 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret.
+
+Secrets are managed using the `podman secret` command
```
TomSweeneyRedHat, line 883:
```suggestion
Secrets are managed using the `podman secret` command.
```

docs/source/markdown/podman-create.1.md
TomSweeneyRedHat: and/or perhaps add a link that the bottom to the new secret create man page

docs/source/markdown/podman-secret-create.1.md
```diff
@@ -0,0 +1,39 @@
...
+Creates a secret using standard input or from a file for the secret content.
+Secrets are a blob of sensitive data which a container needs at runtime but
+you don’t want to store in the image or in source control, such as Usernames and passwords,
+TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size)
+
+## OPTIONS
+
+#### **--driver**=*driver*
+
+Specify the secret driver (default file, which is unencrypted)
```
TomSweeneyRedHat, line 20:
```suggestion
Specify the secret driver (default file, which is unencrypted).
```

docs/source/markdown/podman-secret-create.1.md
```diff
@@ -0,0 +1,39 @@
...
+
+## SYNOPSIS
+**podman secret create** [*options*] *name* *file*
+
+## DESCRIPTION
+
+Creates a secret using standard input or from a file for the secret content.
+Secrets are a blob of sensitive data which a container needs at runtime but
+you don’t want to store in the image or in source control, such as Usernames and passwords,
+TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size)
```
TomSweeneyRedHat, line 14:
```suggestion
TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
```

docs/source/markdown/podman-secret-create.1.md
```diff
@@ -0,0 +1,39 @@
...
+podman\-secret\-create - Create a new secret
+
+## SYNOPSIS
+**podman secret create** [*options*] *name* *file*
+
+## DESCRIPTION
+
+Creates a secret using standard input or from a file for the secret content.
+Secrets are a blob of sensitive data which a container needs at runtime but
+you don’t want to store in the image or in source control, such as Usernames and passwords,
```
TomSweeneyRedHat, line 13: Please don't use pronouns in man pages.

```suggestion
should not be stored in the image or in source control, such as Usernames and passwords,
```

docs/source/markdown/podman-secret-create.1.md
```diff
@@ -0,0 +1,39 @@
...
+
+## OPTIONS
+
+#### **--driver**=*driver*
+
+Specify the secret driver (default file, which is unencrypted)
+
+#### **--help**
+
+Print usage statement
```
TomSweeneyRedHat, line 24:
```suggestion
Print usage statement.
```

docs/source/markdown/podman-secret-inspect.1.md
```diff
@@ -0,0 +1,38 @@
...
+Inspects the specified secret.
+By default, this renders all results in a JSON array. If a format is specified, the given template will be executed for each result.
+Secrets can be queried individually by providing their full name or a unique partial name.
+
+
+## OPTIONS
+
+#### **--format**=*format*
+
+Format secret output using Go template
```
TomSweeneyRedHat, line 20:
```suggestion
Format secret output using Go template.
```

docs/source/markdown/podman-secret-inspect.1.md
```diff
@@ -0,0 +1,38 @@
...
+
+#### **--help**
+
+Print usage statement
```
TomSweeneyRedHat, line 24:
```suggestion
Print usage statement.
```

docs/source/markdown/podman-secret-ls.1.md
```diff
@@ -0,0 +1,33 @@
...
+
+## NAME
+podman\-secret\-ls - List all the available secrets
+
+## SYNOPSIS
+**podman secret ls** [*options*]
+
+## DESCRIPTION
+
+Lists all the secrets that exist. The output can be formatted to a Go template using the **--format**
```
TomSweeneyRedHat, line 11:
```suggestion
Lists all the secrets that exist. The output can be formatted to a Go template using the **--format** option.
```

docs/source/markdown/podman-secret-rm.1.md
```diff
@@ -0,0 +1,34 @@
...
+
+## NAME
+podman\-secret\-rm - Remove one or more secrets
+
+## SYNOPSIS
+**podman volume rm** [*options*] *secret* [...]
+
+## DESCRIPTION
+
+Removes one or more secrets
```
TomSweeneyRedHat, line 11:
```suggestion
Removes one or more secrets.
```

docs/source/markdown/podman-secret-rm.1.md
```diff
@@ -0,0 +1,34 @@
...
+
+## OPTIONS
+
+#### **--all, -a**
+Remove all existing secrets
```
TomSweeneyRedHat, line 16:
```suggestion
Remove all existing secrets.
```

docs/source/markdown/podman-secret-rm.1.md
```diff
@@ -0,0 +1,34 @@
...
+
+Print usage statement
```
TomSweeneyRedHat, line 18:
```suggestion
Print usage statement

```

docs/source/markdown/podman-secret-rm.1.md
```diff
@@ -0,0 +1,34 @@
...
+#### **--help**
+
+Print usage statement
+
```
TomSweeneyRedHat, line 22: rm this blank line

docs/source/markdown/podman-secret-rm.1.md
```diff
@@ -0,0 +1,34 @@
...
+## OPTIONS
+
+#### **--all, -a**
+Remove all existing secrets
+
+Print usage statement
+#### **--help**
+
+Print usage statement
```
TomSweeneyRedHat, line 21:
```suggestion
Print usage statement.
```

TomSweeneyRedHat:
Need to update commands-demo.md  and https://github.com/containers/podman/blob/master/docs/source/Commands.rst, maybe another spot or two to get the man pages to show on docs.podman.io.

mheon:
We've already branched 3.0, so we're already safe to merge this, FYI

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/containers/common/pkg/completion"
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/spf13/cobra"
+)
+
+var (
+	createCmd = &cobra.Command{
+		Use:   "create SECRET",
```
Luap99, line 18: Wrong use line. Should be `create SECRET FILE`

cmd/podman/secrets/list.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+var (
+	lsCmd = &cobra.Command{
+		Use:     "ls [options]",
+		Aliases: []string{"list"},
+		Short:   "List secrets",
+		RunE:    ls,
+		Example: "podman secret ls",
+		Args:    cobra.ExactArgs(0),
```
Luap99, line 27: Add `ValidArgsFunction: completion.AutocompleteNone,`

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,76 @@
...
+		Short: "Create a new secret",
+		Long:  "Create a secret. Default driver is file (unencrypted).",
+		RunE:  create,
+		Args:  cobra.ExactArgs(2),
```
Luap99, line 22: You have to add a `ValidArgsFunction` for all commands. This one is used to provide suggestions on the cli if you type tab. You have to create a new one for the secrets. Look at https://github.com/containers/podman/blob/master/cmd/podman/common/completion.go

cmd/podman/common/create.go
```diff
@@ -603,6 +603,14 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
...
+	)
+
+	_ = cmd.RegisterFlagCompletionFunc(secretFlagName, completion.AutocompleteDefault)
```
Luap99, line 613: The newline should be below and not above. Also you should change the autocomplete function. see other comment

Luap99 (commented):
Some comments regarding the shell completion logic.

cmd/podman/secrets/create.go
vrothberg: FILE or `-` actually

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/containers/podman/v2/cmd/podman/registry"
+	"github.com/containers/podman/v2/pkg/domain/entities"
+	"github.com/spf13/cobra"
+)
+
+var (
+	createCmd = &cobra.Command{
+		Use:   "create SECRET",
+		Short: "Create a new secret",
+		Long:  "Create a secret. Default driver is file (unencrypted).",
```
vrothberg, line 20: I think it's worth mentioning that the input can either be a file or STDIN.

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,76 @@
...
+		if (stat.Mode() & os.ModeNamedPipe) == 0 {
+			return errors.New("if `-` is used, data must be passed into stdin")
+
+		}
+		reader = os.Stdin
+	} else {
+		reader, err = os.Open(path)
+		if err != nil {
+			return err
+		}
```
vrothberg, line 67: defer reader.Close()

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func create(cmd *cobra.Command, args []string) error {
+	name := args[0]
+
+	var err error
+	path := args[1]
+
+	var reader io.Reader
+	if path == "-" {
```
vrothberg, line 53: ` || path == "/dev/stdin" ` ... that's needed as we cannot close stdin below.

docs/source/markdown/podman-create.1.md
```diff
@@ -824,6 +824,12 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret.
```
vrothberg, line 829: Can be specified multiple times.

docs/source/markdown/podman-run.1.md
```diff
@@ -876,6 +876,12 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret.
```
vrothberg, line 881: Can be specified multiple times.

docs/source/markdown/podman-secret-create.1.md
vrothberg, line 13:
```suggestion
you don’t want to store in the image or in source control, such as usernames and passwords,
```

docs/source/markdown/podman-secret.1.md
```diff
@@ -0,0 +1,25 @@
...
+**podman secret** *subcommand*
+
+## DESCRIPTION
+podman secret is a set of subcommands that manage secrets.
+
+## SUBCOMMANDS
+
+| Command | Man Page                                               | Description                                                                    |
+| ------- | ------------------------------------------------------ | ------------------------------------------------------------------------------ |
+| create  | [podman-secret-create(1)](podman-secret-create.1.md)   | Create a secret from a file or STDIN as content                                |
```
vrothberg, line 16: I think we can trim some whitespaces in the last column.

libpod/container_internal_linux.go
```diff
@@ -1650,6 +1651,16 @@ rootless=%d
 		}
 	}
 
+	// This must come after secretMounts, since secretMounts creates the /run/secrets dir
+	if len(c.Secrets()) != 0 {
```
vrothberg, line 1655: The length check isn't necessary. `range` on an empty slice does nothing.

libpod/container_internal_linux.go
```diff
@@ -2367,3 +2378,37 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
 	}
 	return true, nil
 }
+
+// getSecret gets a secret to mount into the container
```
vrothberg, line 2382: What are the return values?

vrothberg: Although it might seem strange at first glance, I prefer using more specific function names. `getObject` is too generic and I wouldn't guess that files are being created.

Maybe we can name it `extractSecretAndPrepareMount`?  Just an example for a more expressive function name.

libpod/container_internal_linux.go
```diff
@@ -2367,3 +2378,37 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
...
+	manager, err := secrets.NewManager(secretsDir)
+	if err != nil {
+		return "", "", err
+	}
+	secr, data, err := manager.LookupSecretData(name)
+	if err != nil {
+		return "", "", err
+	}
+
+	secretFile := filepath.Join(c.config.StaticDir, secr.Name)
```
vrothberg, line 2394: This seems dangerous and subject to name conflicts (e.g., a secret named "containers").  I think we need to create these files in a subdir of `StaticDir`.

pkg/bindings/test/secrets_test.go
```diff
@@ -0,0 +1,105 @@
...
+		connText, err = bindings.NewConnection(context.Background(), bt.sock)
+		Expect(err).To(BeNil())
+	})
+
+	AfterEach(func() {
+
+		s.Kill()
+		bt.cleanup()
+	})
+
```
vrothberg, line 38: I think we should create more than one secret and also check how it behaves if there's no secret.

The intention is to make sure we exercise code paths that are subject to off-by-ones or nil derefs.

pkg/domain/infra/abi/secrets.go
```diff
@@ -0,0 +1,138 @@
...
+					Name: secret.Driver,
+				},
+			},
+		}
+		reports = append(reports, report)
+
+	}
+
+	return reports, errs, nil
+}
```
vrothberg, line 74: Empty line missing

pkg/domain/infra/abi/secrets.go
```diff
@@ -0,0 +1,138 @@
...
+				Driver: entities.SecretDriverSpec{
+					Name:    secret.Driver,
+					Options: secret.DriverOptions,
+				},
+			},
+		}
+		report = append(report, &reportItem)
+	}
+	return report, nil
+}
```
vrothberg, line 102: Empty line missing

test/e2e/secret_test.go
```diff
@@ -0,0 +1,215 @@
+package integration
```
vrothberg, line 1: Please add empty lines between the tests.

vrothberg (commented):
_Really_ good work, @ashley-cui!

This is a huge PR and in really good shape.  Once we have this merged, we can create a follow up card and work together with @edsantiago to create system tests in `test/system`.  Those are executed in the RHEL gating tests and important on the enterprise side.

libpod/container_config.go
```diff
@@ -146,6 +146,8 @@ type ContainerRootFSConfig struct {
 	// working directory if it does not exist. Some OCI runtimes do this by
 	// default, but others do not.
 	CreateWorkingDir bool `json:"createWorkingDir,omitempty"`
+	// Secrets lists secrets to mount into the container
+	Secrets []string `json:"secrets,omitempty"`
```
mheon, line 150: What is this - a list of UUID string we can get the secrets via?

libpod/container_internal_linux.go
```diff
@@ -2367,3 +2378,37 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
 	}
 	return true, nil
 }
+
+// getSecret gets a secret to mount into the container
+func (c *Container) getSecret(name string) (string, string, error) {
+	secretsDir := filepath.Join(c.runtime.store.GraphRoot(), "secrets")
```
mheon, line 2384: Make an accessor for this somewhere - we're going to need a way to determine where these live.

libpod/container_internal_linux.go
mheon: Concur, we need a subdir.

libpod/container_internal_linux.go
```diff
@@ -2367,3 +2378,37 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
...
+		return "", "", err
+	}
+	secr, data, err := manager.LookupSecretData(name)
+	if err != nil {
+		return "", "", err
+	}
+
+	secretFile := filepath.Join(c.config.StaticDir, secr.Name)
+
+	f, err := os.Create(secretFile)
```
mheon, line 2396: Replace this with `ioutil.WriteFile`

libpod/container_internal_linux.go
```diff
@@ -2367,3 +2378,37 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
...
+	if err != nil {
+		return "", "", errors.Wrapf(err, "unable to create %s", secretFile)
+	}
+	defer f.Close()
+	if err := f.Chown(c.RootUID(), c.RootGID()); err != nil {
```
mheon, line 2401: Are you sure this should be root? What if the container is run with `--user mheon`, for example?

libpod/container_config.go
ashley-cui: Yep, either the secret name or secret ID, both work. On a second thought though, I should prolly change it so that it only has secret name.

libpod/define/container_inspect.go
```diff
@@ -62,6 +62,8 @@ type InspectContainerConfig struct {
 	SystemdMode bool `json:"SystemdMode,omitempty"`
 	// Umask is the umask inside the container.
 	Umask string `json:"Umask,omitempty"`
+	// Secrets are the secrets mounted in the container
+	Secrets []string `json:"Secrets,omitempty"`
```
mheon, line 66: This feels like it should be an array of structs

mheon: Is there anything we can report about a secret besides its name?

libpod/container_internal_linux.go
mheon: We also need to handle errors better here. If a secret a container is using is removed, the container becomes unusable. Is that desirable, or should we proceed with a warning?

libpod/options.go
```diff
@@ -1687,6 +1687,17 @@ func WithUmask(umask string) CtrCreateOption {
 	}
 }
 
+// WithSecrets adds secrets to the container
+func WithSecrets(secrets []string) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+		ctr.config.Secrets = secrets
```
mheon, line 1696: We should validate that the secrets actually exist

libpod/container_internal_linux.go
ashley-cui: Not sure, I didn't think about that. Is there a UID function I can use for the case of --user?

libpod/define/container_inspect.go
ashley-cui: doing a podman secret inspect returns its name, id, creation time, and driver. That's the only information available on a secret

pkg/api/handlers/compat/secrets.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+	if reports[0].Err != nil {
+		utils.SecretNotFound(w, name, reports[0].Err)
+		return
+	}
+	utils.WriteResponse(w, http.StatusNoContent, nil)
+}
+
+func CreateSecret(w http.ResponseWriter, r *http.Request) {
```
mheon, line 70: This doesn't look like it's compatible with the Docker API - https://docs.docker.com/engine/api/v1.40/#operation/SecretCreate

Docker supports more parameters, uses no query parameters, and sends everything JSON-encoded in the request body.

If we're not actually Docker-compat, we should not be exposing the compat endpoints.

pkg/api/server/server.go
```diff
@@ -125,6 +125,7 @@ func newServer(runtime *libpod.Runtime, duration time.Duration, listener *net.Li
 		server.registerPluginsHandlers,
 		server.registerPodsHandlers,
 		server.RegisterSwaggerHandlers,
+		server.registerSecretHandlers,
```
mheon, line 128: Alpha order would put this above `RegisterSwaggerHandlers`

pkg/api/server/register_secrets.go
```diff
@@ -0,0 +1,192 @@
...
+	// responses:
+	//   '204':
+	//     description: no error
+	//   '404':
+	//     "$ref": "#/responses/NoSuchSecret"
+	//   '500':
+	//     "$ref": "#/responses/InternalError"
+	r.Handle(VersionedPath("/libpod/secrets/{name}"), s.APIHandler(compat.RemoveSecret)).Methods(http.MethodDelete)
+
+	/*
```
mheon, line 104: If we have compat endpoints, we need to at least stub out all the endpoints, even the ones we don't support. It looks like we're missing the Update endpoint.

libpod/container_config.go
mheon: ID seems a lot safer. I do not want people to be able to delete, and subsequently recreate, secrets and have the container use the new one - that seems like a security risk.

libpod/define/container_inspect.go
mheon: Is it worth including any of that when inspecting the container? Name and ID seem like things we'd want to have.

pkg/api/handlers/compat/secrets.go
ashley-cui: I think I just accidentally wrote extra code for the query params - the handler doesn't actually use it. The only thing missing is labels, which we decided not to implement as of now. I don't think docker actually accepts the template field, it looks to be a docs error, a copy of the driver field. I could either remove the docker-compat or do a logrus warning if someone tries to use labels, let me know the right way to go

libpod/define/container_inspect.go
ashley-cui: Name and ID sounds good, name definitely because the secret would be mounted at /run/secrets/name, and id to verify if its the actual secret, not one that has been deleted and remade with the same name

pkg/api/handlers/compat/secrets.go
mheon: We usually return an error when they pass a non-supported field

libpod/container_internal_linux.go
ashley-cui: Hmm I have to take another look at what I'm doing, in a create/start vs a run.

pkg/api/handlers/compat/secrets.go
ashley-cui: Alright, so then is removing the docker-compat the way to go, and making it a libpod-only endpoint? It would also make sense since docker secrets only exist in the swarm and podman secrets are different enough since they work in vanilla containers.

mheon: If the endpoint accepts the same parameters, we can keep the `compat` endpoint, as long as it returns a Bad Request when something we do not support is requested

ashley-cui:
Addressed most of the review comments, still needs a rebase and some lint-y stuff

cmd/podman/secrets/list.go
```diff
@@ -0,0 +1,95 @@
...
+			ID:        response.ID,
+			Name:      response.Spec.Name,
+			CreatedAt: units.HumanDuration(time.Since(response.CreatedAt)) + " ago",
+			UpdatedAt: units.HumanDuration(time.Since(response.UpdatedAt)) + " ago",
+			Driver:    response.Spec.Driver.Name,
+		})
+
+	}
+	return outputTemplate(cmd, listed)
+}
```
vrothberg, line 68: Empty line missing.  I am not sure why the linter isn't complaining about that.

docs/source/markdown/podman-create.1.md
```diff
@@ -824,6 +824,16 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
+#### **--secret**=*secret*
+
+Give the container access to a secret. Can be specified multiple times.
+
+A secret is blob of sensitive data which a container needs at runtime but
```
vrothberg, line 831:
```suggestion
A secret is a blob of sensitive data which a container needs at runtime but
```

vrothberg: And in other man pages

go.mod
```diff
@@ -71,3 +71,5 @@ require (
 	k8s.io/apimachinery v0.20.2
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 )
+
+replace github.com/containers/common => github.com/containers/common v0.33.1-0.20210119140603-b678ad9332bc
```
vrothberg, line 75: Is that still needed?

libpod/container_internal.go
```diff
@@ -2212,3 +2214,25 @@ func (c *Container) hasNamespace(namespace spec.LinuxNamespaceType) bool {
 	}
 	return false
 }
+
+// extractSecretToStorage copies a secret to storage to
```
vrothberg, line 2218: The comment seems incomplete.

libpod/container_internal_linux.go
```diff
@@ -1650,6 +1650,12 @@ rootless=%d
 		}
 	}
 
+	// This must come after secretMounts, since secretMounts creates the /run/secrets dir
```
vrothberg, line 1653: I think it's worth elaborating a bit further and explain how secrets and mounted into the container.

1. We get the data from the secrets manager.
2. We copy the data to a file in the container's static dir.
3. We mount the previously copied file(s) into `/run/secrets`

vrothberg (commented):
Mini nits. One more iteration and I am cool to merge :)

go.mod
ashley-cui: Oh yeah this was just for testing purposes, though this PR still won't pass tests without a new common.

cmd/podman/secrets/create.go
```diff
@@ -0,0 +1,79 @@
...
+var (
+	createCmd = &cobra.Command{
+		Use:   "create SECRET FILE or -",
+		Short: "Create a new secret",
+		Long:  "Create a secret. Input can be a path to a file or \"-\" (read from stdin). Default driver is file (unencrypted).",
+		RunE:  create,
+		Args:  cobra.ExactArgs(2),
+		Example: `podman secret create mysecret /path/to/secret
+		printf "secretdata" | podman secret create mysecret -`,
+		ValidArgsFunction: completion.AutocompleteNone,
```
Luap99, line 25: we should add a new autocomplete function here
```golang
func AutocompleteSecretCreate(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	if len(args) == 1 {
		return nil, cobra.ShellCompDirectiveDefault
	} 
	return nil, cobra.ShellCompDirectiveNoFileComp
}
```

cmd/podman/secrets/list.go
Luap99: `ValidArgsFunction` is still missing here

ashley-cui:
@Luap99  does ls need `ValidArgsFunction` ? ls doesn't take any args

Luap99:
> @Luap99 does ls need ValidArgsFunction ? ls doesn't take any args. Asking because I'm not sure :)

Yes `completion.AutocompleteNone` should be used to suppress any form of completion. The default is to allow path completion. There is also a test which requires that a `ValidArgsFunction` is set for all commands.

ashley-cui:
Update: everything looks ready but still won't pass tests without a new common :(

ashley-cui:
Hm, vendored in common master, looks like it worked without a new common version. Lets hope CI like this one :)

rhatdan:
I pushed a common with the journald change reverted.

ashley-cui:
Vendoring in the new common vendored in a lot of weird other stuff.. Let me know if that's okay

rhatdan:
Well I opened a vendoring for common, and it pulls in some vendors that common got on runc and friends, so this is not surprising.

Once my PR gets merged, you should easily just update to it.

vrothberg:
```
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv                                        
#| FAIL: podman secret create: --help has 'Options:' section but no '[options]' in synopsis 
```
The `cobra.Command.Use` string of `secret create` needs an [options] sub-string to pass `test/system/015-helps.bats`.

You can test it locally with `PODMAN=./bin/podman bats test/system/015-help.bats`

libpod/container_internal_linux.go
```diff
@@ -2368,3 +2385,30 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
...
+		if err := os.Chown(src, c.RootUID(), c.RootGID()); err != nil {
+			return err
+		}
+		c.state.BindMounts["/run/secrets"] = src
+	} else if err != nil {
+		return err
+	}
+
+	return nil
+}
```
vrothberg, line 2414:
```suggestion
		c.state.BindMounts["/run/secrets"] = src
		return nil
	}

	return err
}
```

Easier to read and extend.

libpod/container_internal_linux.go
```diff
@@ -2368,3 +2385,30 @@ func (c *Container) checkFileExistsInRootfs(file string) (bool, error) {
...
+
+		err = os.MkdirAll(src, 0644)
+		if err != nil {
+			return err
+		}
+		err = label.Relabel(src, c.config.MountLabel, false)
+		if err != nil {
+			return err
+		}
+		if err := os.Chown(src, c.RootUID(), c.RootGID()); err != nil {
```
vrothberg, line 2405: Non blocking nit: it would be nice to have one consistent way for error checks (`if err != nil` vs `if err := foo(); err != nil`).

ashley-cui:
SHE'S GREEN!!!!!

mheon:
LGTM

vrothberg (commented):
/lgtm

TomSweeneyRedHat:
Great job completing this one @ashley-cui!  :fireworks: :1st_place_medal:

ashley-cui:
Thanks @TomSweeneyRedHat !!

PR #9126, opened 2021-01-27
[No merge] swagger: add missing parameters

tmds:
@jwhonce @baude this PR gives an overview of parameters which are missing in podman swagger compared to docker.

This PR isn't meant for merging. You can go over the parameters and see which you want to add, and which you won't support.

### DIFF
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -21,6 +21,19 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//      name: name
 	//      type: string
 	//      description: container name
+	//    - in: body
+	//      name: body
+	//      description: Container to create
+	//      schema:
+	//        allOf:
+	//          - $ref: "#/definitions/ContainerConfig"
+	//          - type: object
+	//            properties:
+	//              HostConfig:
+	//                $ref: "#/definitions/HostConfig"
+	//              NetworkingConfig:
+	//                $ref: "#/definitions/NetworkingConfig"
+	//      required: true
 	//   responses:
 	//     201:
 	//       $ref: "#/responses/ContainerCreateResponse"
@@ -365,6 +378,10 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//    type: boolean
 	//    default: true
 	//    description: Stream the output
+	//  - in: query
+	//    name: one-shot
+	//    type: boolean
+	//    default: false
 	// produces:
 	// - application/json
 	// responses:
--- a/pkg/api/server/register_images.go
+++ b/pkg/api/server/register_images.go
@@ -31,18 +31,31 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//    type: string
 	//    description: needs description
 	//  - in: query
+	//    name: repo
+	//    description: the repository name for the created image
+	//    type: string
+	//  - in: query
 	//    name: tag
 	//    type: string
 	//    description: needs description
+	//  - in: query
+	//    name: message
+	//    description: commit message
+	//    type: string
 	//  - in: header
 	//    name: X-Registry-Auth
 	//    type: string
 	//    description: A base64-encoded auth configuration.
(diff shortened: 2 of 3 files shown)


pkg/api/server/register_containers.go
```diff
@@ -21,6 +21,19 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
...
+	//      schema:
+	//        allOf:
+	//          - $ref: "#/definitions/ContainerConfig"
+	//          - type: object
+	//            properties:
+	//              HostConfig:
+	//                $ref: "#/definitions/HostConfig"
+	//              NetworkingConfig:
+	//                $ref: "#/definitions/NetworkingConfig"
+	//      required: true
```
tmds, line 36: Without this parameter, you can't provide a configuration for the container you want to create.

pkg/api/server/register_images.go
```diff
@@ -624,6 +642,21 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//    description: |
 	//      output configuration TBD
 	//      (As of version 1.xx)
+	//  - name: inputStream
+	//    in: body
+	//    description: "A tar archive compressed with one of the following algorithms: identity (no compression), gzip, bzip2, xz."
```
tmds, line 647: This parameter seems to provide the build context. I don't think there is another parameter to provide it. So without it you can't build an image.

tmds:
@jwhonce @baude unless I misunderstand what the parameters are for, these missing parameters make it not possible to:
* build images, and
* create containers.

baude:
> @jwhonce @baude unless I misunderstand what the parameters are for, these missing parameters make it not possible to:
> 
>     * build images, and
> 
>     * create containers.

Well that certainly cannot be true as we create containers and build images in both the compat and libpod sides of the api.

@jwhonce what do you make of this?

mheon:
AFAIK: The parameters are there, but not documented, so an API client made using the swagger can't access them.

tmds:
How do we proceed with this?

A suggestion: can you go over this PR, and create specific issue(s) for adding the parameters and validating they work?

I think it's important to add these parameters because without them some really basic operations are not exposed through the swagger.

jwhonce:
Aligning the swagger to the as-built code is always good.  An audit of all the endpoints will take some time. After Podman v3.0 is released let's add a card to Jira (@tmds Our Jira instance will soon be public.) to do the audit and generate issues for the missing parameters. That is the pattern we've used in the past.

Any objections?

There are some issues with this PR regarding the references, once fixed I would merge.

tmds:
> After Podman v3.0 is released let's add a card to Jira (@tmds Our Jira instance will soon be public.) to do the audit and generate issues for the missing parameters. That is the pattern we've used in the past.

@jwhonce now that podman 3.0 is out, can you add these cards?

vrothberg:
Friendly ping

tmds:
ping @jwhonce

pkg/api/server/register_networks.go
```diff
@@ -57,6 +57,13 @@ func (s *APIServer) registerNetworkHandlers(r *mux.Router) error {
 	//    type: string
 	//    required: true
 	//    description: the name of the network
+	//  - in: query
+	//    name: verbose
+	//    type: boolean
+	//    default: false
+	//  - in: query
+	//    name: scope
+	//    type: string
```
jwhonce, line 66: See https://github.com/containers/podman/issues/10061

pkg/api/server/register_containers.go
jwhonce: agreed

jwhonce (commented):
Please see issues. As they are resolved I would like to update those areas of the swagger.

@tmds I also agreed to some changes that do not require code changes rather they comment existing behavior. Would you want to provide those changes in a PR that I can get merged for you?

rhatdan:
@tmds Since this PR has not been touched in a couple of months, I am going to close, Please reopen if you want to continue to work on it.

Issue #9111, opened 2021-01-26
Podman 2.2.1 localhost image pulls fail with connection refused on Ubuntu


smijolovic:
Podman version: 2.2.1

Built a debian image (debootstrap) and added it to the local system repo:

$ podman images
REPOSITORY               TAG          IMAGE ID      CREATED         SIZE
localhost/chaasm/debian  buster-slim  9293f2fa2849  10 minutes ago  75.3 MB

When attempting to use that image as the baseimage for the following command:
docker build --pull -t $(BUILD_IMAGE) -f $(TEMP_DIR)/$(CONFIG)/Dockerfile.build $(TEMP_DIR)/$(CONFIG) --log-level=debug

DEBU[0000] base: "localhost/chaasm/debian:buster-slim"  
DEBU[0000] FROM "localhost/chaasm/debian:buster-slim"   
STEP 1: FROM localhost/chaasm/debian:buster-slim
DEBU[0000] Loading registries configuration "/etc/containers/registries.conf" 
DEBU[0000] parsed reference into "[overlay@/home/chaasmadmin/.local/share/containers/storage+/run/user/5000/containers:overlay.mount_program=/usr/bin/fuse-overlayfs]@9293f2fa2849ee4cc875d09e7e5fbf7c46ec308456c5a0c54d0d2ce63d8555ed" 
DEBU[0000] registry "localhost" is not listed in registries configuration "/etc/containers/registries.conf", assuming it's not blocked 
DEBU[0000] parsed reference into "[overlay@/home/chaasmadmin/.local/share/containers/storage+/run/user/5000/containers:overlay.mount_program=/usr/bin/fuse-overlayfs]localhost/chaasm/debian:buster-slim" 
DEBU[0000] parsed reference into "[overlay@/home/chaasmadmin/.local/share/containers/storage+/run/user/5000/containers:overlay.mount_program=/usr/bin/fuse-overlayfs]localhost/chaasm/debian:buster-slim" 
DEBU[0000] copying "docker://localhost/chaasm/debian:buster-slim" to "localhost/chaasm/debian:buster-slim" 
DEBU[0000] Trying to access "localhost/chaasm/debian:buster-slim" 
DEBU[0000] Credentials not found                        
DEBU[0000] Using registries.d directory /etc/containers/registries.d for sigstore configuration 
DEBU[0000]  Using "default-docker" configuration        
[...]

PR #9127, opened 2021-01-27
podman build --pull: use correct policy

vrothberg:
The `--pull` flag should be using the "pull if newer" pull policy rather
than "pull always".  This aligns with what the help message states, what
Buildah does and, according to #9111, what was done before,

Also add a test to prevent future regressions.

Fixes: #9111

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -266,7 +266,7 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 
 	pullPolicy := imagebuildah.PullIfMissing
 	if c.Flags().Changed("pull") && flags.Pull {
-		pullPolicy = imagebuildah.PullAlways
+		pullPolicy = imagebuildah.PullIfNewer
 	}
 	if flags.PullAlways {
 		pullPolicy = imagebuildah.PullAlways
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -29,6 +29,29 @@ EOF
     run_podman rmi -f build_test
 }
 
+@test "podman build - basic test with --pull" {
+    rand_filename=$(random_string 20)
+    rand_content=$(random_string 50)
+
+    run_podman tag $IMAGE localhost/localonly
+
+    tmpdir=$PODMAN_TMPDIR/build-test
+    mkdir -p $tmpdir
+    dockerfile=$tmpdir/Dockerfile
+    cat >$dockerfile <<EOF
+FROM localhost/localonly
+RUN echo $rand_content > /$rand_filename
+EOF
+    # With --pull, Podman would try to pull a newer image but use the local one
+    # if present. See #9111.
+    run_podman build --pull -t build_test $tmpdir
+
+    run_podman run --rm build_test cat /$rand_filename
+    is "$output"   "$rand_content"   "reading generated file in image"
+
+    run_podman rmi -f build_test localhost/localonly
+}
+
 @test "podman build - global runtime flags test" {
     skip_if_remote "--runtime-flag flag not supported for remote"
 


rhatdan:
LGTM
@TomSweeneyRedHat PTAL

vrothberg:
The problem with "pull always" is that we'd fail if we cannot pull the image. Even if a local image exists.

smijolovic:
Great example of working as a community for "shift left" agile development.  Thanks a ton for addressing this quickly.  Can we get this in a release branch (2.2.2) as soon as all testing is complete?  I can then pull and test for the kubernetes build ET in our environment.

vrothberg:
> Great example of working as a community for "shift left" agile development. Thanks a ton for addressing this quickly. Can we get this in a release branch (2.2.2) as soon as all testing is complete? I can then pull and test for the kubernetes build ET in our environment.

Thanks! There won't be 2.2.2 but 3.0.0 is just around the corner. It should be released within the next two weeks.

TomSweeneyRedHat:
/hold
This change appears to be counter of the current man page (http://docs.podman.io/en/latest/markdown/podman-build.1.html).  I need to see if this change is correct, or if the man page is.

TomSweeneyRedHat:
This should not have merged right now....

umohnani8:
oops, I did not see your last comment - sorry!

TomSweeneyRedHat:
We need to talk about this before v3.0 is released.  Please see the explanation in #8210.  Although I would prefer to use PullifNewer for the pull option, it is NOT what Docker does.  That's why when we talked about it in late October early November, we went with the PullAlways option.

If we decide to adopt this change, the build man page needs to change.  I'll create an issue.

vrothberg:
> We need to talk about this before v3.0 is released. Please see the explanation in #8210. Although I would prefer to use PullifNewer for the pull option, it is NOT what Docker does. That's why when we talked about it in late October early November, we went with the PullAlways option.

That is what Docker does (I tested it).  Otherwise, all builds would require a network connection and local images could never be used. 

> If we decide to adopt this change, the build man page needs to change. I'll create an issue.

Thanks for the pointer, I'll update the man page.

TomSweeneyRedHat:
I triple checked myself this morning, and Docker does fail when `--pull` is used as it does a `PullAlways`.  If you don't use the `--pull` option, it does a `PullIfNewer` (or may `PullIfMissing`) and succeeds.

```
# docker version
 Version:           19.03.13

# docker build -t tom -f ~/Dockerfile.pull .
Sending build context to Docker daemon   12.8kB
Step 1/1 : FROM quay.io/tomsweeneyredhat/testing:fourth
fourth: Pulling from tomsweeneyredhat/testing
8e3ba11ec2a2: Pull complete 
4f4fb700ef54: Pull complete 
Digest: sha256:4a52f041edcf20b5bcc2442962cb8c75b88e3c4300a38ea79b98d22be511952d
Status: Downloaded newer image for quay.io/tomsweeneyredhat/testing:fourth
 ---> 72c09588141b
Successfully built 72c09588141b
Successfully tagged tom:latest

# docker images
REPOSITORY                         TAG                 IMAGE ID            CREATED             SIZE
simple-echo                        bob                 cf71716eb35f        14 hours ago        5.57MB
simple-echo                        latest              ef1f9a4fc8e2        14 hours ago        5.57MB
alpine                             3.12.1              d6e46aa2470d        3 months ago        5.57MB
tom                                latest              72c09588141b        2 years ago         4.41MB
quay.io/tomsweeneyredhat/testing   fourth              72c09588141b        2 years ago         4.41MB

## removed quay.io/tomsweeneyredhat/testing:fourth from quay.io web interface and kept it in local storage.

###  No --pull option, succeeds with PullIfNewer
# docker build -t tom -f ~/Dockerfile.pull .
Sending build context to Docker daemon   12.8kB
Step 1/1 : FROM quay.io/tomsweeneyredhat/testing:fourth
 ---> 72c09588141b
Successfully built 72c09588141b
Successfully tagged tom:latest
 
### Use --pull option, fails doing PullAlways
# docker build --pull -t tom -f ~/Dockerfile.pull .
Sending build context to Docker daemon   12.8kB
Step 1/1 : FROM quay.io/tomsweeneyredhat/testing:fourth
unknown: Tag fourth was deleted or has expired. To pull, revive via time machine

```

vrothberg:
@smijolovic, a quick update.  We decided to revert this change after a conversation balancing the pros and cons.

While older versions of Podman did not make pull errors fatal when a local image was found, Podman's goal is to remain compatible with Docker.  `docker build --pull` however will always pull the image and error if it could not be pulled.

I opened https://github.com/containers/podman/pull/9144 but also corrected the --help message to match the behaviour.

smijolovic:
With the testing above, it appears the removal of the --pull will supply the needed functionality to include a localhost cached image. Is that correct?

vrothberg:
> With the testing above, it appears the removal of the --pull will supply the needed functionality to include a localhost cached image. Is that correct?

Yes.  Removing `--pull` will use the local image and your script will work.

rhatdan:
Yes --pull tells the builder to check the registry for newer images.  If it can not check it returns an error.
If --pull is  not specified the the builder will check, but not fail if there is no access to the registry.

TomSweeneyRedHat:
Just tweaking @rhatdan's response slightly:

If --pull is not specified the builder will check, but not fail if there is no access to the registry and the image is in local storage.  If the image is not found locally or in the registry, then it will fail.

smijolovic:
Thanks a ton!  I totally understand the desire to want podman commands to have full feature parity with docker...I agree that is a sound choice.

It's easy enough to strip the pull command from a dockerfile for cached localhost images.

PR #9133, opened 2021-01-27
Cleanup bindings for image pull

rhatdan:
Remove bindings that are not handled over the API.

Leaving this one to not use image pull, since this would
break progress handling.  We should revisit this in the
future.


pkg/domain/infra/tunnel/images.go
```diff
@@ -106,16 +106,16 @@ func (ir *ImageEngine) Prune(ctx context.Context, opts entities.ImagePruneOption
 
 func (ir *ImageEngine) Pull(ctx context.Context, rawImage string, opts entities.ImagePullOptions) (*entities.ImagePullReport, error) {
 	options := new(images.PullOptions)
-	options.WithAllTags(opts.AllTags).WithAuthfile(opts.Authfile).WithCertDir(opts.CertDir).WithArch(opts.Arch).WithOS(opts.OS)
-	options.WithVariant(opts.Variant).WithPassword(opts.Password).WithPullPolicy(opts.PullPolicy)
+	options.WithAllTags(opts.AllTags).WithAuthfile(opts.Authfile).WithArch(opts.Arch).WithOS(opts.OS)
+	options.WithVariant(opts.Variant).WithPassword(opts.Password)
+	options.WithQuiet(opts.Quiet).WithUsername(opts.Username)
```
mheon, line 111: Shouldn't quiet be client side? I see no reason we shouldn't always send the info - the client can just decide not to print it

rhatdan: Currently this is being handled on the server side.  Seemed to risky to fix this now.

rhatdan:
This is ready to go in.

vrothberg (commented):
LGTM

QiWang19 (commented):
LGTM

PR #9136, opened 2021-01-27
Use internal abi function for pause/unpause

rhatdan:

### DIFF
--- a/pkg/api/handlers/compat/containers_pause.go
+++ b/pkg/api/handlers/compat/containers_pause.go
@@ -2,27 +2,50 @@ package compat
 
 import (
 	"net/http"
+	"os"
 
 	"github.com/containers/podman/v2/libpod"
+	"github.com/containers/podman/v2/libpod/define"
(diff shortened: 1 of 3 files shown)


pkg/api/handlers/compat/containers_pause.go
```diff
@@ -2,27 +2,50 @@ package compat
...
+	"github.com/gorilla/schema"
+	"github.com/pkg/errors"
 )
 
 func PauseContainer(w http.ResponseWriter, r *http.Request) {
+	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
+	containerEngine := abi.ContainerEngine{Libpod: runtime}
+	query := struct {
+		All bool `schema:"all"`
```
mheon, line 21: We can't do this without changing return type. It is an *absolutely terrible* user experience if we have All but do not return every error.

rhatdan: Yes I agree, and I am running out of time to fix the returns.

PR #8585, opened 2020-12-03
Add support for rootless network-aliases and static ip/mac

Luap99:
Make sure we pass the ip and mac address as CNI_ARGS to
the cnitool which is executed in the rootless-cni-infra container.

Fixes #7842

Make sure we pass the network aliases as capability args to the
cnitool in the rootless-cni-infra container. Also update the
dnsname plugin in the cni-infra container.

Fixes #8567

PR #9141, opened 2021-01-28
[v3.0] Add support for rootless network-aliases and static ip/mac

Luap99:
Backport #8585 

Make sure we pass the ip and mac address as CNI_ARGS to
the cnitool which is executed in the rootless-cni-infra container.

Make sure we pass the network aliases as capability args to the
cnitool in the rootless-cni-infra container. Also update the
dnsname plugin in the cni-infra container.


vrothberg (commented):
LGTM

PR #9149, opened 2021-01-28
Podman-remote push can support --format 

rhatdan:
Fix man page to document podman push --format fully.

Also found that push was not handling the tlsverify so fixed this.

### DIFF
--- a/docs/source/markdown/podman-push.1.md
+++ b/docs/source/markdown/podman-push.1.md
@@ -90,8 +90,7 @@ solely for scripting compatibility.
 
 #### **--format**, **-f**=*format*
 
-Manifest Type (oci, v2s1, or v2s2) to use when pushing an image to a directory using the 'dir:' transport (default is manifest type of source)
-Note: This flag can only be set when using the **dir** transport. (Not available for remote commands)
+Manifest Type (oci, v2s2, or v2s1) to use when pushing an image.
 
 #### **--quiet**, **-q**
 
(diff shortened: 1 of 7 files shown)


docs/source/markdown/podman-push.1.md
```diff
@@ -90,8 +90,9 @@ solely for scripting compatibility.
 
 #### **--format**, **-f**=*format*
 
-Manifest Type (oci, v2s1, or v2s2) to use when pushing an image to a directory using the 'dir:' transport (default is manifest type of source)
-Note: This flag can only be set when using the **dir** transport. (Not available for remote commands)
+Manifest Type (oci, v2s2, or v2s1) to use when pushing an image.
+Note: The **--format** flag can only be set when using the **dir** and
+**docker** transports. (Not available for remote commands)
```
mheon, line 95: Was this supposed to change the "Not available for remote commands" bit as well?

rhatdan: Nope. Although It now would work for them.

vrothberg: If it works for remote, can we drop the comment as well?

vrothberg: I think we should drop the "Note" entirely.  It also works with the `containers-storage` transport.  Other, such as `docker-archive` or `oci` or `oci-archive` will change the format implicitly.

rhatdan: Dropped and added support for --format to podman-remote.

vrothberg (commented):
LGTM

Issue #9132, opened 2021-01-27
Support dnsConfig in kube yaml with "podman play kube"


phlogistonjohn:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Support "dnsConfig" in pod YAML when running podman play kube.

AFAICT it doesn't show up in a code search at github and only appears in one issue (linked below) so I assume its not implemented yet so I chose feature. :-)

**Steps to reproduce the issue:**

1. I have a test set up where I run a specifc DNS server locally but that is not connected to the system resolv.conf. I can use `podman pod create` to establish this DNS server for my test pod.

2. The command `podman generate kube` doesn't reflect these added options

3. Following the k8s documentation to manaully add "dnsPolicy" and "dnsConfig" subsections does not seem to change the behavior of the containers in the pod.

**Describe the results you received:**

The dnsConfig subsection appears to be ignored.

**Describe the results you expected:**

Podman treats the dnsConfig options similarly to the podman pod create options `--dns`,  `--dns-opt` and `--dns-search`

**Additional information you deem important (e.g. issue happens only occasionally):**

I found the following by searching github: https://github.com/containers/podman/pull/5620#issuecomment-604439390
From this it was not clear that if these were already supported or simply should be supported in the future. 

FWIW I'm running this rootless. But since the `podman pod create` options worked rootless I would assume the kube yaml could apply to either rootless or rootful.

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.14.10
Built:        Tue Dec  8 09:37:43 2020
OS/Arch:      linux/amd64

```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-2.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 81d18b6c3ffc266abdef7ca94c1450e669a6a388'
  cpus: 8
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: journald
  hostname: popcorn
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 210000
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 210000
[...]

PR #9150, opened 2021-01-28
Honor custom DNS in play|generate kube

baude:
when creating kubernetes yaml from containers and pods, we should honor
any custom dns settings the user provided. in the case of generate kube,
these would be provided by --dns, --dns-search, and --dns-opt. if
multiple containers are involved in the generate, the options will be
cumulative and unique with the exception of dns-opt.

when replaying a kube file that has kubernetes dns information, we now
also add that information to the pod creation.

the options for dnspolicy is not enabled as there seemed to be no direct
correlation between kubernetes and podman.

Fixes: #9132

### DIFF
--- a/pkg/specgen/generate/kube/kube.go
+++ b/pkg/specgen/generate/kube/kube.go
@@ -3,6 +3,7 @@ package kube
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/containers/common/pkg/parse"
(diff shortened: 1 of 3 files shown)


pkg/specgen/generate/kube/kube.go
```diff
@@ -44,6 +45,27 @@ func ToPodGen(ctx context.Context, podName string, podYAML *v1.PodTemplateSpec)
...
+		}
+		// search domans
+		if domains := dnsConfig.Searches; len(domains) > 0 {
+			p.DNSSearch = domains
+		}
+		// dns options
+		if options := dnsConfig.Options; len(options) > 0 {
+			dnsOptions := make([]string, 0)
+			for _, opts := range options {
+				dnsOptions = append(dnsOptions, fmt.Sprintf("%s:%s", opts.Name, opts.Value))
```
mheon, line 65: I think we may want to make this conditional - if `Value == ""` we probably just want `Name` with no `:` after it

rhatdan:
LGTM

PR #9151, opened 2021-01-28
Docker ignores mount flags that begin with constency

rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1915332

```
According to the Docker docs, the consistency option should be ignored on Linux.

the possible values are 'cached', 'delegated', and 'consistent', but they should be ignored equally.

This is a widely used option in scripts run by developer machines, as this makes file I/O less horribly slow on MacOS.
```

### DIFF
--- a/cmd/podman/common/volumes.go
+++ b/cmd/podman/common/volumes.go
@@ -353,6 +353,10 @@ func getBindMount(args []string) (spec.Mount, error) {
 			default:
 				return newMount, errors.Wrapf(util.ErrBadMntOption, "%s mount option must be 'private' or 'shared'", kv[0])
 			}
+		case "consistency":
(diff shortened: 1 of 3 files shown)


mheon:
LGTM, but your tests are red

cmd/podman/common/volumes.go
```diff
@@ -353,6 +353,8 @@ func getBindMount(args []string) (spec.Mount, error) {
 			default:
 				return newMount, errors.Wrapf(util.ErrBadMntOption, "%s mount option must be 'private' or 'shared'", kv[0])
 			}
+		case "consistency":
```
vrothberg, line 356: Can you drop a comment here and the other location why we ignore this option?

vrothberg (commented):
/lgtm

Issue #9032, opened 2021-01-20
Building and installing podman from source unavailable on macos


ssbarnea:
At this moment there is no documentation on how someone can build podman and install it from source on macos. This is a real problem as it prevents testing of code from master or testing patches.

For example at https://github.com/containers/podman/issues/7806#issuecomment-743354169 it is mentioned a patch that merged more than month ago but which cannot really be confirmed as working or not because there is no clear documentation on how to build it from source.

Ideally building and installing from source should be same as on linux, as the makefile should detect the platform and try the different build path specific to that platform.

This approach should provide a consistent hacker/developer experience, probably even more important than documenting the process. Not only this, but this would enable easy testing on github where macos hosts are available.

rhatdan:
Sounds great.  Interested in opening a PR for the Makefile to work correctly by default on a MAC?

ssbarnea:
@rhatdan I would not mind helping with that task but at this moment there is zero documentation on how someone would even manually build podman from source on a mac and this makes testing impossible.

Shortly, show me some manual steps and I will automate them, or at least automate most of them.

rhatdan:
@baude @ashley-cui Could you help @ssbarnea out.

ashley-cui:
@ssbarnea I'll document but here's the gist:

`make podman-remote-release-darwin` creates a .zip file that contains podman for mac as well as all the docs for mac. 
`make podman-remote-darwin` creates the binary `podman-remote-darwin` which is just podman for mac (you should rename this to binary to just `podman`)
`make install-podman-remote-darwin-docs` creates the correct docs for macs.

rhatdan:
As I said in the other PR.  It would be very nice if the Makefile realized it is on a MAC and then just made the podman-remote binary as podman automajicly and then you used make install to install it.

PR #9161, opened 2021-01-29
[CI:DOCS]build instructions for macOS

baude:
add instructions on how to build podman on macOS.

big thanks to acui for help in getting this written down.

Fixes: #9032

### DIFF
--- a/build_osx.md
+++ b/build_osx.md
@@ -1,0 +1,55 @@
+# Building the Podman client on macOS
+
+The following describes the process for building the Podman client on macOS.
+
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM, it would be nice to make the Makefile smarter about running on a MAC and eliminate the need to do the special make commands.

make binary; make install

On a MAC this should make podman-remote-darwin as bin/podman and install would do the right thing.

ashley-cui (commented):
LGTM

PR #9163, opened 2021-01-29
Backports for v3.0 RC2

mheon:
Backports for v3.0.0-rc2


mheon:
RC commit now included

Issue #9165, opened 2021-01-30
It is not possible to start a container in a pod with --net=none


rudolfvesely:
/kind bug

**Description**

Dear Podman team,

I tested that rootless `podman run --net=none ...` works only when it's outside of a pod. I can't run a container inside of a rootless pod that was created with `--net=none`.

Please let me know if you need any additional information. Thank you.

Kind regards,

Rudolf Vesely

**Steps to reproduce the issue:**

``` sh
podman pod create --name project1 --net=none
```
~~~
5e1ac474318117ffa1aa853a3e4550c0898075c0224b4a43e73e9e01565aaea6
~~~
``` sh
podman run -d --pod=project1 --name=project1-web docker://docker.io/library/nginx:latest
```
~~~
Trying to pull docker://docker.io/library/nginx:latest...
Getting image source signatures
Copying blob a076a628af6f done
Copying blob d7f36f6fe38f done
Copying blob f72584a26f32 done
Copying blob 7125e4df9063 done
Copying blob 0732ab25fa22 done
Copying config f6d0b4767a done
Writing manifest to image destination
Storing signatures
ERRO[0009] error starting some container dependencies
ERRO[0009] "error stat'ing file `/home/testuser3/.config/cni/net.d`: No such file or directory: OCI not found"
Error: error starting some containers: internal libpod error
~~~

* now let's try again with `--net=none` when creating pod and also when running a container.

``` sh
podman pod rm --all
podman rm --all --force
```
``` sh
podman pod create --name project1 --net=none
```
~~~
16b8f369ce5465b7680ce6997e448cd131148b16b024d6c86d8f2a801733dc5c
~~~
``` sh
podman run -d --pod=project1 --name=project1-web --net=none docker://docker.io/library/nginx:latest
```
~~~
ERRO[0000] error starting some container dependencies
ERRO[0000] "error stat'ing file `/home/testuser3/.config/cni/net.d`: No such file or directory: OCI not found"
Error: error starting some containers: internal libpod error
~~~

**Describe the results you received:**

Error. Please see the output above.

**Describe the results you expected:**

Should run a container that has only localhost.

**Additional information you deem important (e.g. issue happens only occasionally):**

Tested on Fedora 33 on a fresh install.

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 14:37:50 2020
OS/Arch:      linux/amd64
```

[...]

PR #9168, opened 2021-01-31
Fix --network parsing for podman pod create

Luap99:
The `--network` flag is parsed differently for `podman pod create`.
This causes confusion and problems for users. The extra parsing
logic ignored unsupported network options such as `none`,
`container:...` and `ns:...` and instead interpreted them as cni
network names.

Tests are added to ensure the correct errors are shown.

### DIFF
--- a/cmd/podman/pods/create.go
+++ b/cmd/podman/pods/create.go
@@ -171,33 +171,6 @@ func create(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	createOptions.Net.Network = specgen.Namespace{}
-	if cmd.Flag("network").Changed {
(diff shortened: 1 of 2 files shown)


Luap99:
Partially fixes #9165

mheon:
LGTM. Pleasantly surprised the parsing works without that bit.

Luap99:
I should have mentioned that the parsing is already done in `common.NetFlagsToNetOptions(cmd)` some lines above. This is the same logic as podman create/run. This would successfully parse `none` and `container:` but these options are rejected by `pkg/specgen` with a nice error as you can see.

giuseppe (approved):
LGTM

PR #9173, opened 2021-01-31
Endpoint that lists containers does not return correct Status value

mlegenovic:
Eclipse and Intellij Docker plugin determines the state of the
container via the Status field, returned from /containers/json call.
Podman always returns empty string, and because of that, both IDEs
show the wrong state of the container.

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -22,6 +22,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/go-connections/nat"
+	"github.com/docker/go-units"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/compat/containers.go
```diff
@@ -274,6 +276,35 @@ func LibpodToContainer(l *libpod.Container, sz bool) (*handlers.Container, error
...
+			status += " (Paused)"
+		}
+	} else if state == define.ContainerStateRemoving {
+		status = "Removal In Progress"
+	} else if state == define.ContainerStateStopping {
+		status = "Stopping"
+	} else {
+		status = "Unknown"
+	}
+
```
mlegenovic, line 307: I took status-value logic from here https://github.com/moby/moby/blob/46cdcd206c56172b95ba5c77b827a722dab426c5/container/state.go#L69

mlegenovic:
Do not ask me why they use Status field to determine container state, and not State field which is much more suitable.

mheon:
LGTM
/approve

vrothberg (commented):
/lgtm
Thanks!

PR #9174, opened 2021-02-01
Make slirp MTU configurable (network_cmd_options)

bitstrings:
The mtu default value is currently forced to 65520.
This let the user control it using the config key network_cmd_options,
i.e.: network_cmd_options=["mtu=9000"]

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -638,6 +638,7 @@ Valid _mode_ values are:
 - **private**: create a new namespace for the container (default)
 - **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack. This is the default for rootless containers. It is possible to specify these additional options:
   - **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`). Default is false.
+  - **mtu=MTU**: Specify the MTU to use for this network. (Default is `65520`).
   - **cidr=CIDR**: Specify ip range to use for this network. (Default is `10.0.2.0/24`).
   - **enable_ipv6=true|false**: Enable IPv6. Default is false. (Required for `outbound_addr6`).
   - **outbound_addr=INTERFACE**: Specify the outbound interface slirp should bind to (ipv4 traffic only).
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -42,6 +43,9 @@ const (
 
 	// slirp4netnsDNS is the IP for the built-in DNS server in the slirp network
 	slirp4netnsDNS = "10.0.2.3"
+
+	// slirp4netnsMTU the default MTU override
+	slirp4netnsMTU = 65520
 )
 
 // Get an OCICNI network config
@@ -282,6 +286,7 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 	enableIPv6 := false
 	outboundAddr := ""
 	outboundAddr6 := ""
+	mtu := slirp4netnsMTU
 
 	if ctr.config.NetworkOptions != nil {
 		slirpOptions = append(slirpOptions, ctr.config.NetworkOptions["slirp4netns"]...)
@@ -345,6 +350,11 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 				}
 			}
 			outboundAddr6 = value
+		case "mtu":
+			mtu, err = strconv.Atoi(value)
+			if mtu < 68 || err != nil {
+				return errors.Errorf("invalid mtu %q", value)
+			}
 		default:
 			return errors.Errorf("unknown option for slirp4netns: %q", o)
 		}
@@ -358,8 +368,8 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 	if disableHostLoopback && slirpFeatures.HasDisableHostLoopback {
 		cmdArgs = append(cmdArgs, "--disable-host-loopback")
 	}
-	if slirpFeatures.HasMTU {
-		cmdArgs = append(cmdArgs, "--mtu", "65520")
+	if mtu > -1 && slirpFeatures.HasMTU {
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--mtu=%d", mtu))
 	}
 	if !noPivotRoot && slirpFeatures.HasEnableSandbox {
 		cmdArgs = append(cmdArgs, "--enable-sandbox")
--- a/test/e2e/run_networking_test.go
+++ b/test/e2e/run_networking_test.go
@@ -376,6 +376,13 @@ var _ = Describe("Podman run networking", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman run slirp4netns network with mtu", func() {
+		session := podmanTest.Podman([]string{"run", "--network", "slirp4netns:mtu=9000", ALPINE, "ip", "addr"})
+		session.Wait(30)
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring("mtu 9000"))
+	})
+
 	It("podman run slirp4netns network with different cidr", func() {
 		slirp4netnsHelp := SystemExec("slirp4netns", []string{"--help"})
 		Expect(slirp4netnsHelp.ExitCode()).To(Equal(0))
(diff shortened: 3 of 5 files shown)


giuseppe (approved):
LGTM

test/e2e/run_networking_test.go
```diff
@@ -376,6 +376,12 @@ var _ = Describe("Podman run networking", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman run slirp4netns network with mtu", func() {
+		session := podmanTest.Podman([]string{"run", "--network", "slirp4netns:mtu=9000", ALPINE, "ping", "-c1", "10.0.2.2"})
+		session.Wait(30)
+		Expect(session.ExitCode()).To(Equal(0))
+	})
```
Luap99, line 383: The test should check if the mtu was set correctly. For example run `ip addr` instead of `ping` in the container and check with
```golang
Expect(session.OutputToString()).To(ContainSubstring("mtu 9000"))
```

libpod/networking_linux.go
```diff
@@ -345,6 +347,11 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 				}
 			}
 			outboundAddr6 = value
+		case "mtu":
+			mtu, err = strconv.Atoi(value)
+			if mtu < 0 || err != nil {
```
Luap99, line 352: The minimum allowed mtu is `68`. Trying to set a value below that will fail with `Error: /usr/bin/slirp4netns failed: "cannot set MTU: Invalid argument\nWARNING: Support for seccomp is experimental\nchild failed(1)\nWARNING: Support for seccomp is experimental\n"`. You should error here if the value is below 68

Luap99 (commented):
Thanks for the contribution @bitstrings

You have to document that option here https://github.com/containers/podman/blob/master/docs/source/markdown/podman-create.1.md#--networkmode---net. and also in the `podman run` and `podman pod create` man page for the --network option

Luap99:
Also apply the following diff to add shell completion for that option:
```diff
diff --git a/cmd/podman/common/completion.go b/cmd/podman/common/completion.go
index c9a3c5e94..09dd74e20 100644
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -817,6 +817,7 @@ func AutocompleteNetworkFlag(cmd *cobra.Command, args []string, toComplete strin
                                "allow_host_loopback=": getBoolCompletion,
                                "cidr=":                nil,
                                "enable_ipv6=":         getBoolCompletion,
+                               "mtu=":                 nil,
                                "outbound_addr=":       nil,
                                "outbound_addr6=":      nil,
                                "port_handler=": func(_ string) ([]string, cobra.ShellCompDirective) {
```

libpod/networking_linux.go
bitstrings: I was hesitant, seeing `create.go` using 0+ as valid.

rhatdan:
Please update man pages with this information.

libpod/networking_linux.go
Luap99: I see we should fix this there as well.

libpod/networking_linux.go
```diff
@@ -282,6 +283,7 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 	enableIPv6 := false
 	outboundAddr := ""
 	outboundAddr6 := ""
+	mtu := -1
```
mheon, line 286: Should we keep the default of 65536 if unset? @giuseppe @AkihiroSuda

mheon: Sorry, 65520

AkihiroSuda: Yes

bitstrings: This seems wrong to me, why would podman force 65520 if my network is configured differently? Doesn't slirp honor the host mtu? After testing on multiple machines and network adapters, 65520 was always problematic vs the default mtu.

test/e2e/run_networking_test.go
bitstrings: Done. However, I modelled the test on current ones, i.e.: cdir and loopback. Maybe those need a touch up.

test/e2e/run_networking_test.go
```diff
@@ -376,6 +376,13 @@ var _ = Describe("Podman run networking", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman run slirp4netns network with mtu", func() {
+		session := podmanTest.Podman([]string{"run", "--network", "slirp4netns:mtu=9000", ALPINE, "ping", "-c1", "10.0.2.2"})
+		session.Wait(30)
+		Expect(session.ExitCode()).To(Equal(0))
+		Expect(session.OutputToString()).To(ContainSubstring("mtu=9000"))
+	})
```
Luap99, line 384:
```suggestion
	It("podman run slirp4netns network with mtu", func() {
		session := podmanTest.Podman([]string{"run", "--network", "slirp4netns:mtu=9000", ALPINE, "ip", "addr"})
		session.Wait(30)
		Expect(session.ExitCode()).To(Equal(0))
		Expect(session.OutputToString()).To(ContainSubstring("mtu 9000"))
	})
```
You have to run `ip addr` in the container and check if the output contains the string `mtu 9000`

libpod/networking_linux.go
Luap99: In general a higher MTU means higher throughput. However the MTU should be the same on each interface that the packet passes through. Physical NICs typically have an MTU of 1500. Some high-end devices operate with an MTU of 9000. If a packet with a higher MTU has to go through this NIC, it is repackaged into smaller packets. This causes the big performance drop you see. So if you are only communicating between the host and the container, you want an MTU that high. https://github.com/rootless-containers/slirp4netns#iperf3-netns---host That's why I think this MTU was chosen, although I don't understand why it is 65520 and not 65536.

bitstrings: Yes, the loopback device mtu is set at 65536 by default. But my point stands, podman forces a specific scenario based value which I don't think is sane.

I will, however, change the code if it's the way the project wants to go.

bitstrings:
> Please update man pages with this information.

Searching for relevant man page(s) - I can't seem to find anything I could update. I would need some assistance on this.

libpod/networking_linux.go
bitstrings: `slirp4netns` help:

-m, --mtu=MTU            specify MTU (default=1500, max=65521)

I guess it's the almost max?

Luap99:
@bitstrings The man page for podman create https://github.com/containers/podman/blob/master/docs/source/markdown/podman-create.1.md#--networkmode---net. You should add the slirp4netns options there. The same has to be done for podman run and podman pod create.

docs/source/markdown/podman-create.1.md
```diff
@@ -638,6 +638,7 @@ Valid _mode_ values are:
 - **private**: create a new namespace for the container (default)
 - **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack. This is the default for rootless containers. It is possible to specify these additional options:
   - **allow_host_loopback=true|false**: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`). Default is false.
+  - **mtu=MTU**: Specify MTU to use for this network. (Default is `65520`).
```
TomSweeneyRedHat, line 641:
```suggestion
  - **mtu=MTU**: Specify the MTU to use for this network. (Default is `65520`).
```
or better yet
```suggestion
  - **mtu=MTU**: Specify the Maximum Transmission Unit (MTU) to use for this network.   The MTU is the size of the largest Protocol Data Unit (PDU)  that can be communicated in a single network-layer transaction (Default is `65520`).
```
Lifted from Wikipedia.  Whatever you do here, do below too

libpod/networking_linux.go
AkihiroSuda: Changing the default can be discussed in a separate issue/PR with actual benchmark data.

AkihiroSuda: Actually 65520 is known not to be optimal on recent kernel with the default sysctl values: https://github.com/rootless-containers/slirp4netns/issues/128

Maybe we can adjust the default dynamically based on sysctl values. (Or rather adjust sysctl dynamically)

docs/source/markdown/podman-create.1.md
bitstrings: I am trying to follow what's there... I am in no position to decide which is best
Example:
```
  - **cidr=CIDR**: Specify ip range to use for this network. (Default is `10.0.2.0/24`).
```
So I have added the `the`.

libpod/networking_linux.go
bitstrings: Yes, anyways, I have made 65520 the default again.

docs/source/markdown/podman-create.1.md
rhatdan: I am fine with keeping it simple at this point.   If we want to get more data into this section, then we should expand them all.  This issue I have is this is a section that will not be used by the vast majority of people, and really only people who know what they are doing. (Not me...) I worry about adding too much noise for a seldom used feature.

PR #9178, opened 2021-02-01
Fix podman generate systemd --new special char handling

### LINKED ISSUE #9176: podman generate systemd --new can be broken by environment variable substitution and specifier expansion

/kind bug

**Description**

When you have a container with environment variable syntax in command or systemd specifiers, systemd will attempt to expand those result in broken container.

1. `podman run -d --name sleep alpine sh -c 'kill -0 $$ && sleep infinity'` 
2. `podman generate systemd --name --new sleep`
3. Start the generated service.

Container will crash with error `kill: illegal pid: $`. The reason is that systemd will expand environment variable before executing the command. Also, systemd uses `$$` to escape a single `$`, therefore the executed command will be no longer the same as the original command.

Container will start in the same way as original podman command.

> If the executable path is prefixed with ":", environment variable substitution (as described by the "Command Lines" section below) is not applied.

However, this issue is not only limited to literal `$` in command line being expanded environment variable as by systemd, literal `%` may also expanded by systemd as specifiers. For example, if I have `printf '%s' 'test'`, it will become `printf '/bin/sh' 'test'`.

Reference for how systemd expand command line:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=
https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 14:38:08 2020
OS/Arch:      linux/arm64
```

Luap99:
In a systemd unit dollar and percent signs are used for variables. A backslash
is used for escape sequences. If any of these characters are used in the create
command we have to properly escape them so systemd does not try to interpret them.

Fixes #9176


vrothberg (commented):
LGTM, nice work @Luap99 !

pkg/systemd/generate/common.go
```diff
@@ -60,13 +60,21 @@ func filterPodFlags(command []string) []string {
...
+// multiple ones. Also make sure to escape all characters which have a special
+// meaning to systemd -> $,% and \
+// see: https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines
+func escapeSystemdArguments(command []string) []string {
 	for i := range command {
+		command[i] = strings.ReplaceAll(command[i], "$", "$$")
+		command[i] = strings.ReplaceAll(command[i], "%", "%%")
 		if strings.ContainsAny(command[i], " \t") {
 			command[i] = strconv.Quote(command[i])
+		} else if strings.Contains(command[i], `\`) {
```
edsantiago, line 74: I am completely lost here. How can this possibly work as an `else`?

pkg/systemd/generate/common_test.go
```diff
@@ -46,10 +46,46 @@ func TestQuoteArguments(t *testing.T) {
...
+			[]string{`\\`},
+		},
+		{
+			[]string{"foo", `command with backslash \`},
+			[]string{"foo", `"command with backslash \\"`},
+		},
+		{
+			[]string{"foo", `command with two backslashs \\`},
+			[]string{"foo", `"command with two backslashs \\\\"`},
+		},
```
edsantiago, line 84: Yes, I see these tests, but I wonder if golang internal backslash processing is massaging those into equivalent strings? Is it possibly to add printfs to the test, or temporarily mismatch the strings to force a failure, and see what Go spits out as the test name?

edsantiago (commented):
I must be missing something really obvious

pkg/systemd/generate/common.go
edsantiago: OK I think I understand: `strconv.Quote()` seems to do backslash interpretation. If this is the case, could you add one more test case, one in which the test string contains no spaces or tabs? Thank you!

pkg/systemd/generate/common_test.go
Luap99: You can run the tests with `go test ./pkg/systemd/generate/` Add a extra char somewere and it will fail and give you the output.

pkg/systemd/generate/common.go
Luap99: There is a test case at line 74-75 in common_test.go.

edsantiago: My bad: I missed that, sorry.

One more question: will there ever be a condition in which we need to pass literals like `\n`, `\a`, `\007` ?

Luap99: I cannot think of any cases where this would be wanted.

PR #9180, opened 2021-02-01
[v3.0] Use cachix for static build & include podman-remote build

cevich:
Backport from master

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -344,14 +344,13 @@ static_alt_build_task:
         ALT_NAME: 'Static build'
         # Do not use 'latest', fixed-version tag for runtime stability.
         CTR_FQIN: "docker.io/nixos/nix:2.3.6"
-    # This is critical, it helps to avoid a very lengthy process of
-    # statically building every dependency needed to build podman.
-    # Assuming the dependency and build description hasn't changed,
-    # this cache ensures only the static podman binary is built.
-    nix_cache:
-        folder: '/var/cache/nix'
-        # Cirrus will calculate/use sha of this output as the cache key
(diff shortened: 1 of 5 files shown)


cevich:
@saschagrunert PTAL...Does the cache "key" need to be altered (on this branch) somehow, so the cache is distinct from the master branch?  *note* we have the `$DEST_BRANCH` variable available for use in `.cirrus.yml` and every other script, in case that helps.

saschagrunert:
> @saschagrunert PTAL...Does the cache "key" need to be altered (on this branch) somehow, so the cache is distinct from the master branch? _note_ we have the `$DEST_BRANCH` variable available for use in `.cirrus.yml` and every other script, in case that helps.

This looks fine from my side. The cache can be re-used, even if some derivations differ. Cachix will just put both into the same cache.

.cirrus.yml
```diff
@@ -335,7 +335,7 @@ static_alt_build_task:
         - build
     # Community-maintained task, may fail on occasion.  If so, uncomment
     # the next line and file an issue with details about the failure.
-    allow_failures: $CI == $CI
+    # allow_failures: $CI == $CI
```
saschagrunert, line 338: Do we want to disallow failures now?

saschagrunert (approved):

.cirrus.yml
cevich: I *think* so.  I did this on master as well.  The switch to cachix seems to have greatly improved the reliability of this job.  I'm puzzled why Cirrus's cache failed us...but "proof of the pudding is in the eating" :shrug:

cevich:
> Cachix will just put both into the same cache side by side with a different checksum.

Okay, glad to hear.  This is what I wasn't sure about.

PR #9188, opened 2021-02-01
Report StatusConflict on Pod opt partial failures

### LINKED ISSUE #8865: Errors are not reported when starting a pod from the REST API

jwhonce:
- When one or more containers in the Pod reports an error on an operation
report `StatusConflict` and report the error(s)

- `jsoniter.RegisterTypeEncoderFunc()` used to marshal error as string using `error.Error()`. This picks up any other places in API where `error` has been used.

- Update test framework to allow setting any flag when creating pods

Fixes #8865

### DIFF
--- a/pkg/api/handlers/compat/resize.go
+++ b/pkg/api/handlers/compat/resize.go
@@ -84,5 +84,5 @@ func ResizeTTY(w http.ResponseWriter, r *http.Request) {
 		// reasons.
 		status = http.StatusCreated
 	}
-	utils.WriteResponse(w, status, "")
+	w.WriteHeader(status)
 }
--- a/pkg/api/handlers/libpod/pods.go
+++ b/pkg/api/handlers/libpod/pods.go
@@ -139,19 +139,20 @@ func PodStop(w http.ResponseWriter, r *http.Request) {
 			logrus.Errorf("Error cleaning up pod %s container %s: %v", pod.ID(), id, err)
 		}
 	}
-	var errs []error //nolint
+
+	report := entities.PodStopReport{Id: pod.ID()}
 	for id, err := range responses {
-		errs = append(errs, errors.Wrapf(err, "error stopping container %s", id))
+		report.Errs = append(report.Errs, errors.Wrapf(err, "error stopping container %s", id))
 	}
-	report := entities.PodStopReport{
-		Errs: errs,
-		Id:   pod.ID(),
+
+	code := http.StatusOK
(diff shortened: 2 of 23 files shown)


baude:
code and idea LGTM, legit test failures

vrothberg (commented):
Code LGTM but I don't have enough knowledge judging whether `409 Conflict` is valid in all cases.

MDN [states](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409): "Conflicts are most likely to occur in response to a PUT request. For example, you may get a 409 response when uploading a file which is older than the one already on the server resulting in a version control conflict."

Does that include transient errors of `pod start`?

jwhonce:
@vrothberg The handler already used 409, I simply documented that fact and expanded it's usage to all instances of this condition. I believe 409 is more useful than 500 (that was returned in some cases) since a retry may resolve the issue, which is not implied by 500.

PR #9189, opened 2021-02-01
add macvlan as a supported network driver

baude:
instead of using the --macvlan to indicate that you want to make a
macvlan network, podman network create now honors the driver name of
*macvlan*.  Any options to macvlan, like the parent device, should be
specified as a -o option.  For example, -o parent=eth0.

the --macvlan option was marked as deprecated in the man page but is
still supported for the duration of 3.0.


vrothberg (commented):
LGTM

giuseppe (approved):
/lgtm

PR #9190, opened 2021-02-01
New 'make completions' target

edsantiago:
Generate shell completion files for podman and podman-remote
and for all known shells (bash, fish, zsh). Fix Readme file
in completions dir: it suggested running 'make completion',
singular; no such target existed. Since the install target
is plural, I choose to make the new target plural also.

This is intended for use in CI some day, in a check such as:

  (in contrib/cirrus/runner.sh):

    make completions
    SUGGESTION="run 'make completions' and commit all changes" ./hack/tree_status.sh

The goal would be to make sure that any new podman subcommands
or flags are accompanied by their corresponding shell helpers
on all commits.

IT IS NOT POSSIBLE TO ENFORCE THIS NOW. As I understand it,
Cobra is slow to incorporate community PRs, so the podman team
has chosen to sneak in completion files generated by an
unreleased and un-vendored version of Cobra. Running 'make
completions' right now would clobber those and result in
a diminished user experience.

I'm submitting this anyway as a stepping-stone toward that
future day when we can create such a CI hook.

[NO TESTS NEEDED]


edsantiago:
@Luap99 I would welcome your feedback on my commit message as well as the code. Very low priority. TIA.

rhatdan:
LGTM

Luap99 (commented):
LGTM

Luap99:
FYI, with the next cobra release we can add powershell completion.

edsantiago:
/hold cancel

(I had to duckduckgo powershell)

PR #9200, opened 2021-02-02
Makefile: refactor ginkgo * ginkgo-remote

edsantiago:
Create a new common target, ginkgo-run, with the main ginkgo
test incantation; ginkgo and ginkgo-remote now invoke that,
with a clearer distinction between their arguments.

Reason for this: 'make remoteintegration' was not generating
ginkgo logs in CI. Reason: '-debug' option was missing from
the ginkgo incantation. Reason: impossibility of maintaining
duplicate long-complicated lines. This PR distills the common
aspects, making the differences clearer between local & remote.

IMPORTANT NOTE: '-nodes 3' was also missing from remote ginkgo.
This PR adds it. If this was intentionally omitted, please let
me know (with reasons) so I can refactor that too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,13 +303,17 @@ localunit: test/goecho/goecho
 .PHONY: test
 test: localunit localintegration remoteintegration localsystem remotesystem  ## Run unit, integration, and system tests.
 
+.PHONY: ginkgo-run
+ginkgo-run:
+	$(GOBIN)/ginkgo -v $(TESTFLAGS) -tags "$(TAGS)" $(GINKGOTIMEOUT) -cover -flakeAttempts 3 -progress -trace -noColor -nodes 3 -debug test/e2e/. $(HACK)
+
 .PHONY: ginkgo
 ginkgo:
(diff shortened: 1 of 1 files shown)


cevich:
> Reason: '-debug' option was missing from the ginkgo incantation. 

Actually I'm pretty sure it's also the `-nodes 3` change as well.  Without that, the ginkgo output would be 100% the same as a single-node log file.  So ***caution*** is needed, I think we setup the remote tests to run single-threaded for a reason:  It will make debugging the "server" side of remote testing _MUCH_ harder (b/c centralized podman logging?).

Anyway...let's see what the actual CI jobs say about this change...

rhatdan:
LGTM

edsantiago:
Timing results, local vs remote:

| int | distro | root | host | local | remote |
| --- | --- | --- | --- | ---: | ---: |
| int | fedora-32 | root | host | 30:50 | 46:29 |
| int | fedora-33 | root | host | 30:33 | 40:28 |
| int | ubuntu-2004 | root | host | 38:23 | 01:10:30 |
| int | ubuntu-2010 | root | host | 36:10 | 51:29 |

The `-nodes 3` doesn't seem to have made any difference in remote.

PR #9203, opened 2021-02-02
[NO TESTS NEEDED] Improve binding generator

matejvasek:
This allows to have slices other than slices of strings in options structs.

Before this wasn't possible:
```go
type WaitOptions struct {
	Conditions []define.ContainerStatus // <- non-string slice
	Interval   time.Duration
	Latest     bool
}
```

### DIFF
--- a/pkg/bindings/generator/generator.go
+++ b/pkg/bindings/generator/generator.go
@@ -54,33 +54,19 @@ func (o *{{.StructName}}) ToParams() (url.Values, error) {
 		if reflect.Ptr == f.Kind() {
 			f = f.Elem()
 		}
-		switch f.Kind() {
-		case reflect.Bool:
-			params.Set(fieldName, strconv.FormatBool(f.Bool()))
-		case reflect.String:
-			params.Set(fieldName, f.String())
-		case reflect.Int, reflect.Int64:
-			// f.Int() is always an int64
-			params.Set(fieldName, strconv.FormatInt(f.Int(), 10))
-		case reflect.Uint, reflect.Uint64:
-			// f.Uint() is always an uint64
-			params.Set(fieldName, strconv.FormatUint(f.Uint(), 10))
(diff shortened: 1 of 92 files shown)


pkg/bindings/generator/generator.go
```diff
@@ -93,7 +79,10 @@ func (o *{{.StructName}}) ToParams() (url.Values, error) {
 			}
 
 			params.Set(fieldName, s)
+		default:
```
matejvasek, line 82: Mostly for debugging.

baude:
lgtm

rhatdan:
Could you remove draft.

matejvasek:
@rhatdan  [NO TESTS NEEDED] should be only in PR title, not it commit message? if so I'll remove that from the messages.

matejvasek:
@baude I removed debugging code (the default case of a switch), and fixed indentation on one place. Please lgtm again.

pkg/bindings/generator/generator.go
matejvasek: I removed it just in case.

pkg/bindings/util/util.go
```diff
@@ -0,0 +1,30 @@
+package util
+
+import (
+	"reflect"
+	"strconv"
+)
+
+func IsSimpleType(f reflect.Value) bool {
```
matejvasek, line 8: I couldn't come up with something better than `simple`, maybe scalar?

matejvasek:
Shall I squash commits?

jwhonce:
LGTM

vrothberg (approved):
/hold cancel

matejvasek:
@mheon this will alson need to be backported to v3 if we want to backport https://github.com/containers/podman/pull/9048

mheon:
Alright. Should be fine.

Issue #9167, opened 2021-01-31
podman create network macvlan does not create expected configuration


knumskull:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Creating a network using the command line tools, will not create the expected result.

**Steps to reproduce the issue:**

1. `podman network create --subnet 192.168.1.0/24 --gateway 192.168.1.254 --macvlan eth0 --opt vlan=10,mtu=1500 example-network`

2.

3.

**Describe the results you received:**
```
{
   "cniVersion": "0.4.0",
   "name": "example-network",
   "plugins": [
      {
         "type": "macvlan",
         "master": "eth0",
         "ipam": {
            "type": "dhcp"
         }
      }
   ]
}
```

**Describe the results you expected:**
```
{
   "cniVersion": "0.4.0",
   "name": "example-network",
   "plugins": [
      {
         "type": "macvlan",
         "master": "eth0",
         "ipam": {
            "type": "host-local",
            "ranges": [
              [{
                "subnet": "192.168.1.0/24",
                "gateway": "192.168.1.254"
              }]
            ],
            "routes": [
              {"dst": "0.0.0.0/0"}
            ]
         },
         "vlan": 10,
         "mtu": 1500
      }
   ]
}
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 3.0.0-dev
```

**Output of `podman info --debug`:**

```
host:
  arch: arm64
  buildahVersion: 1.19.2
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.aarch64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 66d50d2f6c81e28bacd99ee8e8d222f028eef63f'
  cpus: 4
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: rpi4.crazy.lab
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.9.16-200.fc33.aarch64
  linkmode: dynamic
  memFree: 4446306304
  memTotal: 8173101056
  ociRuntime:
    name: crun
    package: crun-0.16-3.fc33.aarch64
    path: /usr/bin/crun
    version: |-
      crun version 0.16
      commit: eb0145e5ad4d8207e84a327248af76663d4e50dd
      spec: 1.0.0
[...]

PR #9204, opened 2021-02-02
Honor network options for macvlan networks

baude:
when creating a macvlan network, we should honor gateway, subnet, and
mtu as provided by the user.

Fixes: #9167


rhatdan:
LGTM

vrothberg (approved):
/hold cancel

PR #9208, opened 2021-02-02
System test for #9096 (truncated stdout)

edsantiago:
This actually tests conmon, not podman; but that's the whole
point of system tests in the first place: if a problem exists,
we want to fail loudly, no matter whose fault it is.

(I can't get this to fail on my f33 laptop; OP on #9096 claims
it only fails on Ubuntu. We'll see what happens in CI).

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -610,4 +610,43 @@ json-file | f
     is "$output" "$randomcontent" "cat random content"
 }
 
+# https://github.com/containers/podman/issues/9096
+# podman exec may truncate stdout/stderr; actually a bug in conmon:
+# https://github.com/containers/conmon/issues/236
+@test "podman run - does not truncate or hang with big output" {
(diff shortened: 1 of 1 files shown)


edsantiago:
@giuseppe @haircommander could you take a quick look and LMK if my test logic here is meaningful? Does the podman log output path exercise a close-enough code path to that of plain stdout?

vrothberg (commented):
Very clever to use a log file :+1:

haircommander:
I've pulled this in to https://github.com/containers/podman/pull/8979 where I'm seeing failures from conmon 2.0.25. Hopefully this fails there, and updating to conmon 2.0.26 will fix it, verifying this issue is fixed

edsantiago:
@haircommander thank you! Although this PR has just merged, could you please follow up with your findings? (Right now I see #8979 stuck in all-red in `int` tests, which means the `sys` tests didn't get a chance to run)

PR #9209, opened 2021-02-02
Gating tests: diff test: workaround for RHEL8 failure

edsantiago:
RHEL8 rootless gating tests are inconsistently failing with:
```
   $ podman diff --format json -l
   #
   {"changed":["/etc"],"added":["/sys/fs","/sys/fs/cgroup","/pMOm1Q0fnN"],"deleted":["/etc/services"]}
   # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   # #|     FAIL: added
   # #| expected: '/pMOm1Q0fnN'
   # #|   actual: '/sys/fs'
   # #|         > '/sys/fs/cgroup'
   # #|         > '/pMOm1Q0fnN'
   # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Reason: PR #8561, I think (something to do with /sys on RHEL).

Workaround: ignore '/sys/fs' in diffs.

Signed-off-by: Ed Santiago <santiago@redhat.com>

PR #9210, opened 2021-02-03
Makefile: make bin/* real targets!

edsantiago:
Backstory: every time you run 'make podman' or even
just 'make', you get a full recompile. This is sub-ideal.

Cause: I don't really know. It looks complicated. #5017
introduced a .PHONY for bin/podman, for reasons not
explained in the PR. Then, much later, #5880 well-
intentionedly but improperly tweaked the 'find'
command used in defining SOURCES, adding a -prune
but without the corresponding and required -print.
Let's just say, it was an unfortunate cascade of events.

This PR fixes the SOURCES definition and removes the
highly-undesired .PHONY from podman & podman-remote,
making it so you can type 'make' and, oh joy, not
build anything if it's current. The way 'make' is
supposed to work.

Why fix this now? Because my PR (#9209) was failing in CI,
in the Validate step:

    Can't exec "./bin/podman": No such file or directory at hack/xref-helpmsgs-manpages line 223.

It failed even on Re-run, and only passed once I force-pushed
the PR (with no changes, just a new commit SHA). I have no idea
why bin/podman wasn't built, and I have zero interest in pursuing
that right now, but the proper solution is to add bin/podman as
a Makefile dependency for that particular test. So done.

While I'm at it, fix what is pretty clearly a typo in a .PHONY

And, finally, fix a go-md2man warning introduced in #9189

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,11 @@ PKG_MANAGER ?= $(shell command -v dnf yum|head -n1)
 # ~/.local/bin is not in PATH on all systems
 PRE_COMMIT = $(shell command -v bin/venv/bin/pre-commit ~/.local/bin/pre-commit pre-commit | head -n1)
 
-SOURCES = $(shell find . -path './.*' -prune -o -name "*.go")
+# This isn't what we actually build; it's a superset, used for target
+# dependencies. Basically: all *.go files, except *_test.go, and except
+# anything in a dot subdirectory. If any of these files is newer than
+# our target (bin/podman{,-remote}), a rebuild is triggered.
+SOURCES = $(shell find . -path './.*' -prune -o \( -name '*.go' -a ! -name '*_test.go' \) -print)
 
 BUILDFLAGS := -mod=vendor $(BUILDFLAGS)
 
@@ -195,7 +199,6 @@ test/goecho/goecho: .gopathok $(wildcard test/goecho/*.go)
 	$(GO) build $(BUILDFLAGS) -ldflags '$(LDFLAGS_PODMAN)' -o $@ ./test/goecho
 
 
-.PHONY: bin/podman
 bin/podman: .gopathok $(SOURCES) go.mod go.sum ## Build with podman
 # Make sure to warn in case we're building without the systemd buildtag.
 ifeq (,$(findstring systemd,$(BUILDTAGS)))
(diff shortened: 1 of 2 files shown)


edsantiago:
@containers/podman-maintainers I propose not merging this one without at least three lgtms. I've been careful, but this is a small change with deep ramifications; it warrants careful scrutiny.

Makefile
```diff
@@ -35,7 +35,7 @@ PKG_MANAGER ?= $(shell command -v dnf yum|head -n1)
 # ~/.local/bin is not in PATH on all systems
 PRE_COMMIT = $(shell command -v bin/venv/bin/pre-commit ~/.local/bin/pre-commit pre-commit | head -n1)
 
-SOURCES = $(shell find . -path './.*' -prune -o -name "*.go")
+SOURCES = $(shell find . -path './.*' -prune -o -name "*.go" -print)
```
saschagrunert, line 38: We did something similar in CRI-O as well, but excluded test files:
```
GO_FILES := $(shell find . -type f -name '*.go' -not -name '*_test.go')
```
How to get the best out of both `find` commands? :upside_down_face:

vrothberg (commented):
Changes LGTM, I like @saschagrunert's idea to exclude unit tests

rhatdan:
LGTM with @saschagrunert change.

edsantiago:
Changed as follows (whitespace for readability):
```diff
- find . -path './.*' -prune -o    -name "*.go"                           -print
+ find . -path './.*' -prune -o \( -name '*.go' -a ! -name '*_test.go' \) -print
```
Tested via:
```console
$ find (the old one) >/tmp/find.1
$ find (the new one) >/tmp/find.2
$ diff /tmp/find.[12] | egrep -v '^< .*_test\.go'
$ diff /tmp/find.[12] | less      (quick eyeball sanity check)
```
Thanks for the suggestion, @saschagrunert !

Since this got more complicated and fragile than I'd hoped, I added a comment.

vrothberg (commented):
LGTM

Makefile
```diff
@@ -35,7 +35,11 @@ PKG_MANAGER ?= $(shell command -v dnf yum|head -n1)
 # ~/.local/bin is not in PATH on all systems
 PRE_COMMIT = $(shell command -v bin/venv/bin/pre-commit ~/.local/bin/pre-commit pre-commit | head -n1)
 
-SOURCES = $(shell find . -path './.*' -prune -o -name "*.go")
+# This isn't what we actually build; it's a superset, used for target
+# dependencies. Basically: all *.go files, except *_test.go, and except
+# anything in a dot subdirectory. If any of these files is newer than
+# our target (bin/podman{,-remote}), a rebuild is triggered.
+SOURCES = $(shell find . -path './.*' -prune -o \( -name '*.go' -a ! -name '*_test.go' \) -print)
```
saschagrunert, line 42: For what would we need the `-prune` if we just use `-type f` and use only go files which are non tests? I guess the `-print` is obsolete in that case too, right?

edsantiago: The `-prune` is what excludes dot-directories; see #5880. You and I may not have dot-directories in our work environment, but the originator of that PR does, and it seems reasonable to me to exclude those.

saschagrunert: Makes sense.

saschagrunert (approved):
LGTM

Issue #9206, opened 2021-02-02
[APIv2] killing a container hangs after "Switch podman stop/kill/wait handlers to use abi"


riyad:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Trying to find regressions by exercising the APIv2 trough [docker-py](https://github.com/docker/docker-py)'s test suite (see #5386) I came across a regressions in the [`RestartContainerTest::test_restart_with_low_timeout`](https://github.com/docker/docker-py/blob/master/tests/integration/api_container_test.py#L1163) test.

Killing a container hangs with no result returned and the CPU spinning like crazy.

It specifically broke with commit 81a3ba36 .

**Steps to reproduce the issue:**

1. `curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/containers/create' -H 'Content-Type: application/json' -d '{"Tty": false, "OpenStdin": false, "StdinOnce": false, "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Cmd": ["sleep", "9999"], "Image": "alpine", "NetworkDisabled": false}'`

2. `curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/containers/398e20bb6c01742c1aebe09cd196efbfa6d82873561968fbf96dd4ca0b16198f/start'`

3. `curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/containers/398e20bb6c01742c1aebe09cd196efbfa6d82873561968fbf96dd4ca0b16198f/kill'`

**Describe the results you received:**

The kill call hangs with no response while the CPU starts spinning like crazy.

**Describe the results you expected:**

Kill the container and return result immediately.

**Additional information you deem important (e.g. issue happens only occasionally):**

Running `podman ps -a`  in another terminal while the kill call spins CPU shows that the container has stopped:

```
$ podman ps -a
CONTAINER ID  IMAGE                            COMMAND     CREATED        STATUS                      PORTS   NAMES
398e20bb6c01  docker.io/library/alpine:latest  sleep 9999  3 minutes ago  Exited (137) 2 minutes ago          modest_brown
```

**Output of `podman version`:**

```
Version:      3.0.0-dev
API Version:  3.0.0
Go Version:   go1.15.6
[...]

PR #9214, opened 2021-02-03
Fix invalid wait condition on kill

rhatdan:
When using the compatability tests on kill, the kill
function goes into an infinite wait loop taking all of the CPU.

This change will use the correct wait function and exit properly.

Fixes: https://github.com/containers/podman/issues/9206


vrothberg (commented):
LGTM

matejvasek (approved):

rhatdan:
This looks like the fixes were only merged into the main branch. so don't need to back port.  Will get these fixes in 3.1.

Issue #9211, opened 2021-02-03
'podman generate kube' confuses 'args' with 'command', and no support for 'command' 


nzjustin:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

In general (not specific to this issue) there is confusion between the Podman/Docker concepts of: 

'entrypoint' and 'command'

And the Kubernetes equivalents of:

'command' and 'args'

Podman/Docker 'entrypoint' = Kubernetes 'command'
Podman/Docker 'command' = Kubernetes 'args'

When you use `podman generate kube` on a pod/container with custom command arguments they are put as 'command' in the yml file when it should be 'args'. As a side effect of this, there is no support for a custom entrypoint. You either get the default one, or if you also specify a custom argument, you get the argument in place of the command as described previously.

**Steps to reproduce the issue:**

Create a pod with a container that has a custom (podman/docker) entrypoint and a custom (podman/docker) command:

`podman create --entrypoint /bin/sleep --name test --pod new:testpod centos:8 10s`

Generate a Kubernetes pod definition:

`podman generate kube testpod`

**Describe the results you received:**

The custom command argument is listed as the 'command' (where the entrypoint should be) in the yml output and, as a result the 'entrypoint' is not in the yml file at all.

```
# Generation of Kubernetes YAML is still under development!
#
# Save the output of this file and use kubectl create -f to import
# it into Kubernetes.
#
# Created with podman-2.0.5
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2021-02-03T00:54:24Z"
  labels:
    app: testpod
  name: testpod
spec:
  containers:
  - command:
    - 10s
    env:
    - name: PATH
      value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
    - name: TERM
      value: xterm
    - name: container
      value: podman
    - name: HOSTNAME
      value: testpod
    image: docker.io/library/centos:8
    name: test
    resources: {}
    securityContext:
      allowPrivilegeEscalation: true
      capabilities: {}
      privileged: false
      readOnlyRootFilesystem: false
      seLinuxOptions: {}
    workingDir: /
status: {}
```

**Describe the results you expected:**

I expected to see 

```
...
containers:
  - command:
    - /bin/sleep
  - args:
    - 10s
...
```

**Output of `podman version`:**

```
Version:      2.0.5
API Version:  1
Go Version:   go1.14.7
[...]

PR #9220, opened 2021-02-03
generate kube: handle entrypoint

vrothberg:
The spec of a Kube Container has a `Command` and `Args`.  While both are
slices, the `Command` is the counterpart of the entrypoint of a libpod
container.  Kube is also happily accepting the arguments to as following
items in the slice but it's cleaner to move those to `Args`.

Fixes: #9211

### DIFF
--- a/libpod/kube.go
+++ b/libpod/kube.go
@@ -353,22 +353,21 @@ func containerToV1Container(c *Container) (v1.Container, []v1.Volume, *v1.PodDNS
 		return kubeContainer, kubeVolumes, nil, err
 	}
 
-	containerCommands := c.Command()
-	kubeContainer.Name = removeUnderscores(c.Name())
+	// Handle command and arguments.
+	if ep := c.Entrypoint(); len(ep) > 0 {
+		// If we have an entrypoint, set the container's command as
+		// arguments.
+		kubeContainer.Command = ep
+		kubeContainer.Args = c.Command()
+	} else {
+		kubeContainer.Command = c.Command()
+	}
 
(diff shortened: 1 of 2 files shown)


vrothberg:
@haircommander PTAL

libpod/kube.go
```diff
@@ -353,22 +353,21 @@ func containerToV1Container(c *Container) (v1.Container, []v1.Volume, *v1.PodDNS
...
+		kubeContainer.Command = c.Command()
+	}
 
+	kubeContainer.Name = removeUnderscores(c.Name())
 	_, image := c.Image()
 	kubeContainer.Image = image
 	kubeContainer.Stdin = c.Stdin()
 
-	// prepend the entrypoint of the container to command
-	if ep := c.Entrypoint(); len(c.Entrypoint()) > 0 {
```
mheon, line 364: Do we have any tests verifying entrypoint from the image is properly used?

vrothberg: Good thinking.  I don't think so no.  I'll add another test :+1:

haircommander: I believe we do for play kube specifically, but I am not sure about generate

haircommander: stale github state had me answering a question that was already addressed :upside_down_face:

rhatdan:
LGTM

mheon:
Would like a final LGTM from @haircommander before merge

haircommander:
/lgtm

thanks for taking this @vrothberg

PR #9227, opened 2021-02-04
Bump github.com/gogo/protobuf/proto to v1.3.2

TomSweeneyRedHat:
Bumping github.com/gogo/protobuf/proto to v1.3.2 to prevent a low
likelyhood CVE.  Partially addresses:

https://bugzilla.redhat.com/show_bug.cgi?id=1924542

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/ghodss/yaml v1.0.0
 	github.com/godbus/dbus/v5 v5.0.3
+	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/google/shlex v0.0.0-20181106134648-c34317bd91bf
 	github.com/google/uuid v1.2.0
 	github.com/gorilla/mux v1.8.0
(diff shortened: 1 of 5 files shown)


rhatdan:
I think we should wait until APIMachinery gets updated, since that is what is pulling this in.  We have not confirmed that this is a Podman vulnerability, or if the vulnerable code paths are used by Podman.

TomSweeneyRedHat:
The two spots that I found that were pulling it in were vendored code, I'm not sure if we are or are not using it under there:  
```
vendor/github.com/Microsoft/hcsshim/Protobuild.toml:  "google/protobuf/descriptor.proto" = "github.com/gogo/protobuf/protoc-gen-gogo/descriptor"
vendor/github.com/containerd/cgroups/stats/v1/metrics.pb.go:	proto "github.com/gogo/protobuf/proto"
vendor/github.com/docker/docker/api/types/swarm/runtime/plugin.pb.go:import proto "github.com/gogo/protobuf/proto"
```
We also pull it in a number of files in `tests/tools`, but I'm not worried about those.

vrothberg:
Ping. Is the still desired? Consider a LGTM from me in case.

rhatdan:
I was hoping we would get this as a side effect of Kubernetes fixing their code, but so far no movement.

rhatdan:
We have decided that since we don't use the vulnerable code, to not take this PR.

Issue #9230, opened 2021-02-04
Podman v3.0 fails to run a container on a readonly rootfs


Luap99:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman run fails with `Error: error chowning container 01041b61625c64e34a849b974c5aa9dcc34377fd5025b9ee7fd629b2a0c5ad79 workdir to container root: chown /run/user/1000/test: read-only file system` when  I try to run a container with a read only rootfs. 

This worked with `v2.2.1`.

**Steps to reproduce the issue:**
```
$ ROOTFS=/run/user/1000/test
$ mkdir $ROOTFS
$ podman unshare mount --rbind -r / $ROOTFS
$ podman unshare mount -t tmpfs none $ROOTFS/run
$ podman run -i -t --rootfs $ROOTFS sleep 5
```

**Describe the results you received:**
```
Error: error chowning container 01041b61625c64e34a849b974c5aa9dcc34377fd5025b9ee7fd629b2a0c5ad79 workdir to container root: chown /run/user/1000/test: read-only file system
```

**Version**
```
$ podman version 
Version:      3.0.0-rc2
API Version:  3.0.0
Go Version:   go1.15.6
Git Commit:   510a983a752a136a5df0bdbff9d14db0490956b2
Built:        Thu Feb  4 13:51:00 2021
OS/Arch:      linux/amd64
```

PR #9231, opened 2021-02-04
fix logic when not creating a workdir

vrothberg:
When resolving the workdir of a container, we may need to create it unless
the user sets it explicitly on the command line.  Otherwise, we just do a
presence check.  Unfortunately, there was a missing return that lead us
to fall through into attempting to create and chown the workdir.  That
caused a regression when running on a read-only root fs.

Fixes: #9230

@Luap99 PTAL

It took me longer than I want to admit until I found an easy way to test it (doing the unshare+mount dance didn't work out well in the system tests).

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -608,6 +608,19 @@ json-file | f
     # a subdir of a volume.
     run_podman run --rm --workdir /IamNotOntheImage -v $testdir/content:/IamNotOntheImage/foo $IMAGE cat foo
     is "$output" "$randomcontent" "cat random content"
+
+    # Make sure that running on a read-only rootfs works (#9230).
+    if ! is_rootless && ! is_remote; then
+        # image mount is hard to test as a rootless user
+        # and does not work remotely
+        run_podman image mount $IMAGE
+        romount="$output"
+
+        run_podman run --rm --rootfs $romount echo "Hello world"
+        is "$output" "Hello world"
+
+        run_podman image unmount $IMAGE
+    fi
 }
 
 # https://github.com/containers/podman/issues/9096
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -213,6 +213,7 @@ func (c *Container) resolveWorkDir() error {
 			// we need to return the full error.
 			return errors.Wrapf(err, "error detecting workdir %q on container %s", workdir, c.ID())
 		}
+		return nil
 	}
 
 	// Ensure container entrypoint is created (if required).


Luap99 (commented):
LGTM

I can confirm that this fixes my issue.

test/system/030-run.bats
```diff
@@ -608,6 +608,16 @@ json-file | f
     # a subdir of a volume.
     run_podman run --rm --workdir /IamNotOntheImage -v $testdir/content:/IamNotOntheImage/foo $IMAGE cat foo
     is "$output" "$randomcontent" "cat random content"
+
+    # Make sure that running on a read-only rootfs works (#9230).
+    skip_if_rootless "image mounts are too hard to test rootless"
```
edsantiago, line 613: Please don't use `skip` in the middle of a test; that has the result of marking the entire test as skipped, which will confuse someone looking at logs. Instead, could you do:
```bash
    if ! is_rootless; then
        ...
```

edsantiago: Incidentally, a cheap way to test as rootless is `--rootfs /`. Unfortunately, that always fails in one of these two ways:
```console
$ ./bin/podman run --rm --rootfs / echo hi
Error: open `//.containerenv`: No such file or directory: OCI not found
```
or
```console
$ ./bin/podman run --rm --rootfs / echo hi
Error: mkdir `secrets`: Permission denied: OCI permission denied
```
(I don't know why it's never a consistent failure)

mheon:
LGTM once Ed's test issues are addressed

test/system/030-run.bats
edsantiago: Looks like you also have to check for remote:
```bash
    if ! is_rootless && ! is_remote; then
        ...
```
(reason: podman-remote doesn't handle `image mount`)

vrothberg: Thanks! Updated, also to skip if remote.

TomSweeneyRedHat:
LGTM
assuming happy tests

test/system/030-run.bats
```diff
@@ -608,6 +608,18 @@ json-file | f
     # a subdir of a volume.
     run_podman run --rm --workdir /IamNotOntheImage -v $testdir/content:/IamNotOntheImage/foo $IMAGE cat foo
     is "$output" "$randomcontent" "cat random content"
+
+    # Make sure that running on a read-only rootfs works (#9230).
+    if ! is_rootless && ! is_remote; then
+    	# image mount is hard to test as a rootless user
```
edsantiago, line 614: There's a mix of spaces and tabs on this line.

vrothberg: Need to fix my vimrc. Thank you!

Issue #9191, opened 2021-02-02
Open mounts all over the place on userdata/shm


xrow:
/kind bug

**Description**

It seems podman is not able to clean up layers. I have no pod runnning but hundreds of mounts. The system gets slow and akward. A reboot helps.

shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/b7d38b15781e1cdfd7a2801a9782ba08cec8830430269a240b1bff0c0dbdaea4/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/ad9982869d9083252ce5740850c91e09258ccbceab899d0f8b1684341b6ca440/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/32de96aa5678c7d1e828fc9749b2cc2af69fdb0181611a6403336981dcd0599a/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/ab57b11b29913f903cefb10b421cd44be06562bde1c7764b22987650bd0c33ae/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/f9ec808492d5f6f4d2611815612271e56460f62c55690a35aa22b7950faed307/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/1d4c2557fb49999ccafad42fbe322a6eafcc3b97084d656324340e65f736b387/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/6e34cb240ad464f33385ffcecde3d966e6e347021c758192706b1a5090b41540/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/6dc80979b859d678765d14441cebcfa4cdf0a9bd60e3bba79994d996bf48e653/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/56762539661e3db998423486908532e6f7c6600071b149ace96cb330bbf00ed1/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/fd0f46ec33abce37d4cde7cd691824968b77ee5f374e281266e3c17bea55af32/userdata/shm
shm                     64000        0     64000   0% /var/lib/containers/storage/overlay-containers/91487dcefa77870323cd65583806bfea9a93080c7189fabbae1fa92a1e3aa841/userdata/shm
[...]

PR #9240, opened 2021-02-05
Do not reset storage when running inside of a container

rhatdan:
Currently if the host shares container storage with a container
running podman, the podman inside of the container resets the
storage on the host. This can cause issues on the host, as
well as causes the podman command running the container, to
fail to unmount /dev/shm.

podman run -ti --rm --privileged -v /var/lib/containers:/var/lib/containers quay.io/podman/stable podman run alpine echo hello
	* unlinkat /var/lib/containers/storage/overlay-containers/a7f3c9deb0656f8de1d107e7ddff2d3c3c279c11c1635f233a0bffb16051fb2c/userdata/shm: device or resource busy
	* unlinkat /var/lib/containers/storage/overlay-containers/a7f3c9deb0656f8de1d107e7ddff2d3c3c279c11c1635f233a0bffb16051fb2c/userdata/shm: device or resource busy

Since podman is volume mounting in the graphroot, it will add a flag to
/run/.containerenv to tell podman inside of container whether to reset storage or not.

Since the inner podman is running inside of the container, no reason to assume this is a fresh reboot, so if "container" environment variable is set then skip
reset of storage.

Also added tests to make sure /run/.containerenv is runnig correctly.

Fixes: #9191

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -1,6 +1,7 @@
 package libpod
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
@@ -26,6 +27,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	"github.com/docker/docker/pkg/namesgenerator"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -622,9 +624,12 @@ func (r *Runtime) Shutdown(force bool) error {
 func (r *Runtime) refresh(alivePath string) error {
 	logrus.Debugf("Podman detected system restart - performing state refresh")
 
-	// First clear the state in the database
-	if err := r.state.Refresh(); err != nil {
-		return err
+	// Clear state of database if not running in container
+	if !graphRootMounted() {
+		// First clear the state in the database
+		if err := r.state.Refresh(); err != nil {
+			return err
+		}
 	}
 
 	// Next refresh the state of all containers to recreate dirs and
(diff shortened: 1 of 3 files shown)


rhatdan:
Setting [NO TESTS NEEDED] since,I do not know of an easy way to test this.  @edsantiago Any ideas.  We need a configured podman inside of a container to test.

edsantiago:
@rhatdan one of the CI steps is [integration tests in container](https://github.com/containers/podman/wiki/Cirrus-Dependency-Visualization) (see the aqua box near top center). Would it be possible to do something like:
* Create a $TMPDIR/fake-var-lib-containers
  * inside there, create a file that non-container podman would normally delete
* run `podman -v $TMPDIR/fake-var-lib-containers:/var/lib/containers run ....`
* Then the test forks: if running with `$container=SET`, confirm that the file is untouched; otherwise, confirm that file is removed

libpod/runtime.go
```diff
@@ -486,14 +486,17 @@ func makeRuntime(ctx context.Context, runtime *Runtime) (retErr error) {
...
+		// If the file doesn't exist, and we are not running inside of
+		// a container we need to refresh the state
 		// This will trigger on first use as well, but refreshing an
 		// empty state only creates a single file
 		// As such, it's not really a performance concern
-		if os.IsNotExist(err) {
-			doRefresh = true
-		} else {
-			return errors.Wrapf(err, "error reading runtime status file %s", runtimeAliveFile)
+		if os.Getenv("container") == "" {
```
mheon, line 494: We're just trading one set of issues for another here - if you actually *are* the first Podman call after a reboot, things are going to be massively messed up.

rhatdan: The first podman would have been the outer shell which would have done the reset.  I think the other cases are unlikely.  Podman running within Docker, CRI-O Buildah as the first time.

rhatdan:
@mheon @vrothberg @giuseppe @baude @umohnani8 @ashley-cui @TomSweeneyRedHat PTAL

mheon:
I'm still opposed on the grounds that this is potentially very harmful. If the user runs Podman in an environment where the `CONTAINER` environment variable is set and no refresh is performed, we can end up with deadlocks (refresh includes recreating the SHM locks and reacquiring lock ownership - without that any newly-created containers or locks will re-use existing lock IDs).

rhatdan:
But this is preventing a far more likely case which is actually a real world occurance where customers hit this by mounting their /var/lib/containers into a container, and it screwed up their database. The use case you are worried about is difficult for me to imagine.  Where as we have blogs describing that users should do this to share their images into privileged containers.

mheon:
New version LGTM

TomSweeneyRedHat:
As @mheon is hip with the changes,
LGTM

PR #9269, opened 2021-02-08
Allow path completion for podman create/run --rootfs

Luap99:
If the --rootfs flag is set podman create/run expect a host
path as first argument. The shell completion should provide
path completion in that case.

[NO TESTS NEEDED]
This can manually be verified with `podman run --rootfs [TAB]`.

### DIFF
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -426,6 +426,12 @@ func AutocompleteCreateRun(cmd *cobra.Command, args []string, toComplete string)
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
 	if len(args) < 1 {
+		// check if the rootfs flag is set
+		// if it is set to true provide directory completion
+		rootfs, err := cmd.Flags().GetBool("rootfs")
+		if err == nil && rootfs {
+			return nil, cobra.ShellCompDirectiveFilterDirs
+		}
 		return getImages(cmd, toComplete)
 	}
 	// TODO: add path completion for files in the image


mheon:
LGTM

cmd/podman/common/completion.go
```diff
@@ -426,6 +426,12 @@ func AutocompleteCreateRun(cmd *cobra.Command, args []string, toComplete string)
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
 	if len(args) < 1 {
+		// check if the rootfs flag is set
+		// if it is set to true we have to provide path completion
+		rootfs, err := cmd.Flags().GetBool("rootfs")
+		if err == nil && rootfs {
+			return nil, cobra.ShellCompDirectiveDefault
```
edsantiago, line 433: Just curious: why not `ShellCompDirectiveFilterDirs` ? Can `rootfs` ever be something other than a directory path? (e.g. tarball, image)?

rhatdan: No it has to be a directory.

Luap99: Good catch, I completely forgot that this option exists.

edsantiago:
Aw, phooey. I was just about to slash-lgtm when I remembered: this will have no effect unless we update the completion files themselves. @Luap99 since you're the only one who knows how to bring in the proper magic version of cobra, could you 

/hold

edsantiago:
Sorry, had to hold really fast

...could you do so and resubmit?

edsantiago:
PS `make completions` should now work

Luap99:
/hold cancel

@edsantiago There is no need to update the completion files. `bin/podman run --rootfs [TAB]` should work as expected now. Almost all completion logic exists in go so once we update the binary it should work. 

The completions files only needs to be updated for bug fixes in the scripts etc... For example the fish completion doesn't support `ShellCompDirectiveFilterDirs` and will use the default completion instead. This is fine but if `ShellCompDirectiveFilterDirs` is implemented for fish we would need to update the script.

PR #9270, opened 2021-02-08
Fix Docker APIv2 push endpoint

matejvasek:
Docker doesn't have destination parameter as libpod,
the "image name" path parameter is supposed to be destination.

### DIFF
--- a/pkg/api/handlers/compat/images_push.go
+++ b/pkg/api/handlers/compat/images_push.go
@@ -1,7 +1,6 @@
 package compat
 
 import (
-	"context"
 	"net/http"
 	"strings"
 
@@ -76,7 +75,15 @@ func PushImage(w http.ResponseWriter, r *http.Request) {
 	if _, found := r.URL.Query()["tlsVerify"]; found {
 		options.SkipTLSVerify = types.NewOptionalBool(!query.TLSVerify)
 	}
-	if err := imageEngine.Push(context.Background(), imageName, query.Destination, options); err != nil {
+
+	var destination string
+	if _, found := r.URL.Query()["destination"]; found {
+		destination = query.Destination
+	} else {
+		destination = imageName
+	}
+
+	if err := imageEngine.Push(r.Context(), imageName, destination, options); err != nil {
 		if errors.Cause(err) != storage.ErrImageUnknown {
 			utils.ImageNotFound(w, imageName, errors.Wrapf(err, "failed to find image %s", imageName))
 			return
--- a/test/apiv2/12-imagesMore.at
+++ b/test/apiv2/12-imagesMore.at
@@ -24,13 +24,10 @@ t GET libpod/images/$IMAGE/json 200 \
 
 # Run registry container
 podman run -d --name registry -p 5000:5000 quay.io/libpod/registry:2.6 /entrypoint.sh /etc/docker/registry/config.yml
+sleep 2
 
 # Push to local registry
-# FIXME: this is failing:
-#   "cause": "received unexpected HTTP status: 500 Internal Server Error",
-#   "message": "error pushing image \"localhost:5000/myrepo:mytag\": error copying image to the remote destination: Error writing blob: Error initiating layer upload to /v2/myrepo/blobs/uploads/ in localhost:5000: received unexpected HTTP status: 500 Internal Server Error",
-#   "response": 400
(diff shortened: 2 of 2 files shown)


mheon:
/approve
Sure, LGTM

vrothberg (commented):
Thanks! Can you add a commit description and ideally a test?

rhatdan:
LGTM but needs a test.

matejvasek:
@rhatdan @vrothberg test added (or fixed?) + git message improved.

test/apiv2/12-imagesMore.at
```diff
@@ -25,12 +25,7 @@ t GET libpod/images/$IMAGE/json 200 \
 # Run registry container
 podman run -d --name registry -p 5000:5000 quay.io/libpod/registry:2.6 /entrypoint.sh /etc/docker/registry/config.yml
 
-# Push to local registry
-# FIXME: this is failing:
-#   "cause": "received unexpected HTTP status: 500 Internal Server Error",
-#   "message": "error pushing image \"localhost:5000/myrepo:mytag\": error copying image to the remote destination: Error writing blob: Error initiating layer upload to /v2/myrepo/blobs/uploads/ in localhost:5000: received unexpected HTTP status: 500 Internal Server Error",
-#   "response": 400
-#t POST libpod/images/localhost:5000/myrepo:mytag/push\?tlsVerify\=false '' 200
+t POST "images/localhost:5000/myrepo/push?tlsVerify=false&tag=mytag" '' 200
```
edsantiago, line 28: I'm somewhat sure the test failure is a race condition: the podman registry is still in the process of coming up.

Could you add a `sleep 2` just before this, and see if it helps? If it does, I will follow up with a cleaner `wait-for-port` helper function.

And while you're at it, please revert line 28, I think that's a helpful comment.

matejvasek: ah my machine is too fast!

matejvasek: <b>works on my machine<sup>TM</sup></b>

edsantiago: Mine too. But our CI systems... a.r.e.. v.e.r.y.. s.l.o.w...  hence my suggestion. Thanks for understanding.

matejvasek: @edsantiago maybe it's not about fast machine, maybe I had registry running from previous run on my machine.

edsantiago: Well, the test passed on my laptop, and I didn't/don't have a registry (or anything) on :5000

matejvasek: Have you at least had `quay.io/libpod/registry:2.6` in local registry? It takes a while to download.

matejvasek: Oh while the image is being pulled it should be blocked, ignore the comment above.

edsantiago:
`int remote f33 root host` failed, but the corresponding `f32` passed, so I'm treating the former as a flake and have restarted. LGTM once tests pass. And, once this merges, I have my refactoring PR ready to go.

matejvasek:
> `int remote f33 root host` failed, but the corresponding `f32` passed, so I'm treating the former as a flake and have restarted. LGTM once tests pass. And, once this merges, I have my refactoring PR ready to go.

OK. Thanks for help with this @edsantiago .

edsantiago:
LGTM but I'd like @baude or @jwhonce to have the final say-so

vrothberg (commented):
LGTM

pkg/api/handlers/compat/images_push.go
```diff
@@ -76,7 +75,11 @@ func PushImage(w http.ResponseWriter, r *http.Request) {
 	if _, found := r.URL.Query()["tlsVerify"]; found {
 		options.SkipTLSVerify = types.NewOptionalBool(!query.TLSVerify)
 	}
-	if err := imageEngine.Push(context.Background(), imageName, query.Destination, options); err != nil {
+	destination := query.Destination
+	if destination == "" {
```
jwhonce, line 79:
```suggestion
	if _, found := r.URL.Query()["destination"]; !found {
```
Would be more correct if looking for a missing parameter.

matejvasek: @jwhonce fixed

PR #9273, opened 2021-02-08
[v3.0] Backport final breaking API change to v3.0

mheon:
I was hoping to avoid this until 3.1, but it's a potentially breaking API change so let's get it into 3.0.

This backports https://github.com/containers/podman/pull/9048. Unfortunately, this requires that we also backport an uncomfortably large amount of changes to API handler code from @rhatdan.

### DIFF
--- a/pkg/bindings/containers/types_resizetty_options.go
+++ b/pkg/bindings/containers/types_resizetty_options.go
@@ -1,7 +1,6 @@
 package containers
 
 import (
-	"fmt"
 	"net/url"
 	"reflect"
 	"strings"
@@ -69,8 +68,6 @@ func (o *ResizeTTYOptions) ToParams() (url.Values, error) {
 			}
 
 			params.Set(fieldName, s)
-		default:
-			panic(fmt.Sprintf("don't known how to handle %s", f.Type().String()))
 		}
 
 	}
--- a/pkg/bindings/containers/types_resizetty_options.go
+++ b/pkg/bindings/containers/types_resizetty_options.go
@@ -1,11 +1,12 @@
 package containers
 
 import (
+	"fmt"
 	"net/url"
 	"reflect"
-	"strconv"
 	"strings"
 
+	"github.com/containers/podman/v2/pkg/bindings/util"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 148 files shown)


mheon:

rhatdan:
LGTM

pkg/bindings/containers/types_resizetty_options.go
```diff
@@ -43,33 +43,19 @@ func (o *ResizeTTYOptions) ToParams() (url.Values, error) {
...
+		switch {
+		case util.IsSimpleType(f):
+			params.Set(fieldName, util.SimpleTypeToParam(f))
+		case f.Kind() == reflect.Slice:
+			for i := 0; i < f.Len(); i++ {
+				elem := f.Index(i)
+				if util.IsSimpleType(elem) {
+					params.Add(fieldName, util.SimpleTypeToParam(elem))
+				} else {
+					return nil, errors.New("slices must contain only simple types")
```
TomSweeneyRedHat, line 55: A lot of very similar code like this.  It would be good if we could manage to add a param or two to it and just have it in one place.

TomSweeneyRedHat: The last comment was for potential future work, not for this PR.

matejvasek: Isn't that a generated code?

matejvasek: But yeah, it could be improved by using some common interface. Or by generics if they will add them to go :smile:

TomSweeneyRedHat:
A relatively quick review and I'm leaning heavily on the CI tests saying life is good.
LGTM

pkg/bindings/containers/types_resizetty_options.go
rhatdan: It is generated code, but we are including it in the source, since everyone would have to generate it every time.
Getting to more shared code would be nice.

matejvasek: @rhatdan  Another approach would be generate `ToParams()` differently. Now all the `ToParams()` are the same, they use reflection. Maybe with using `ast.*` we could create more performant  non-reflection `ToParams()` specialized for each struct. `WithXYZ()` methods seems to be generated in such a way.

rhatdan: If you want to take a stab at that I am sure we would look at it.

PR #9274, opened 2021-02-08
Fix varlink GetVersion()

baude:
Fixed two small problems with the varlink GetVersion() call.  The first
was that an error was being eaten and not returned to users when a
string to int conversion was being done.  The second was that the
variable given to the string to int conversion was not something that
could be converted into an integer (i.e. 2.1.0).  Adjusted the
APIVersion to be only representative of the major number (i.e. 2).

Fixes bugzilla: #1925928

### DIFF
--- a/libpod/define/version.go
+++ b/libpod/define/version.go
@@ -42,7 +42,7 @@ func GetVersion() (Version, error) {
 		}
 	}
 	return Version{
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

lsm5 (approved):
I see :)

LGTM

PR #9275, opened 2021-02-08
Add missing params for podman-remote build 

rhatdan:
Currently we still have hard coded --isolation=chroot for podman-remote build.

Implement missing arguments for podman build

Implements
--jobs, --disable-compression, --excludes

Fixes:
MaxPullPushRetries
RetryDuration

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -10,9 +10,11 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"github.com/containers/buildah"
 	"github.com/containers/buildah/imagebuildah"
+	"github.com/containers/buildah/util"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v2/libpod"
 	"github.com/containers/podman/v2/pkg/api/handlers/utils"
@@ -20,6 +22,7 @@ import (
 	"github.com/containers/podman/v2/pkg/channel"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/gorilla/schema"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -65,6 +68,7 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		Annotations            string `schema:"annotations"`
 		BuildArgs              string `schema:"buildargs"`
 		CacheFrom              string `schema:"cachefrom"`
+		Compression            uint64 `schema:"compression"`
 		ConfigureNetwork       int64  `schema:"networkmode"`
 		CpuPeriod              uint64 `schema:"cpuperiod"`  // nolint
 		CpuQuota               int64  `schema:"cpuquota"`   // nolint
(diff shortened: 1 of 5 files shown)


test/system/070-build.bats
```diff
@@ -453,6 +453,40 @@ EOF
...
+@test "build with copy-from referencing the base image" {
+  _prefetch busybox
+  target=busybox-derived
+  target_mt=busybox-mt-derived
+  containerfile1=$tmpdir/Containerfile1
+    cat >$containerfile1 <<EOF
+FROM docker.io/library/busybox AS build
+RUN rm -f /bin/paste
+USER 1001
+COPY --from=docker.io/library/busybox /bin/paste /test/
```
edsantiago, line 465: Could I encourage you to `s;docker.io/library/;quay.io/libpod/;` ?

rhatdan: Yup cut and pasted from Buildah.

rhatdan:
Fixes: https://github.com/containers/podman/issues/9290

TomSweeneyRedHat:
The fedora-33 test failed during an update of ubutnu.  I've hit re-run.

pkg/api/handlers/compat/images_build.go
```diff
@@ -66,24 +68,28 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
...
 		ConfigureNetwork       int64  `schema:"networkmode"`
-		CpuPeriod              uint64 `schema:"cpuperiod"`  // nolint
-		CpuQuota               int64  `schema:"cpuquota"`   // nolint
-		CpuSetCpus             string `schema:"cpusetcpus"` // nolint
-		CpuShares              uint64 `schema:"cpushares"`  // nolint
+		CpuPeriod              uint64 `schema:"cpuperiod"`   // nolint
+		CpuQuota               int64  `schema:"cpuquota"`    // nolint
+		CpuSetCpus             string `schema:"cpusetcpus"`  // nolint
+		CpuShares              uint64 `schema:"cpushares"`   // nolint
+		Compression            uint64 `schema:"compression"` // nolint
```
TomSweeneyRedHat, line 75: nit Compression before ConfigureNetwork

TomSweeneyRedHat:
One nit of a nite, otherwise 
LGTM

vrothberg:
Restarted flakes

rhatdan:
I think the compose ones are real.  For some reason when the Docker API is used, the network is not being setup correctly.

rhatdan:
@vrothberg @giuseppe @mheon PTAL

pkg/api/handlers/compat/images_build.go
```diff
@@ -156,7 +164,15 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
...
 	}
 	format := buildah.Dockerv2ImageManifest
+	registry := query.Registry
+	isolation := buildah.IsolationChroot
+	/*
+		// FIXME, This is very broken.  Buildah will only work with chroot
+		isolation := buildah.IsolationDefault
+	*/
 	if utils.IsLibpodRequest(r) {
+		//		isolation = buildah.Isolation(query.Isolation)
```
giuseppe, line 174: should we drop this commented code?

rhatdan: Lets merge this.  I waiting to open a PR to fix this as soon as this PR merges.

vrothberg (commented):
/lgtm

test/system/070-build.bats
```diff
@@ -453,6 +453,45 @@ EOF
...
+    root_single_job=$output
+
+    run_podman image mount ${target_mt}
+    root_multi_job=$output
+
+    # Check that both the version with --jobs 1 and --jobs=N have the same number of files
+    test $(find $root_single_job -type f | wc -l) = $(find $root_multi_job -type f | wc -l)
+  fi
+}
+
```
edsantiago, line 494: @rhatdan @nalind I realize this is already merged, but I'm afraid this test is almost entirely a NOP. I've spent the last hour-plus trying to clean it up, and am forced to give up because I don't truly understand its purpose. Is it to test the `--jobs` option? To test `COPY --from`? I'm going to need to rewrite the whole thing, and I would be grateful for any hints on how to do so.

rhatdan: The basic idea, I believe since @nalind wrote it for Buildah, was to run the same build twice once with multiple --jobs and make sure that the output in the end was the same.

edsantiago: Thank you! In that case, are lines 476-477 (the first two builds) needed? Or for that matter `containerfile1`?

rhatdan: @nalind WDYT?

Rereading the tests, I stole this from, I am now not sure what it is testing.

nalind: The test was originally to ensure that when COPY --from was given an image name rather than a name we'd given to a build stage, that it would pull content from the unmodified image rather than a stage that was based on it, which was a bug that we'd fixed.
The parts that ran builds both with and without --jobs and counted the files in containers based on the resulting images, to check that the results from both were comparable, were mixed in later rather than added as a separate test.

Issue #8897, opened 2021-01-06
podman generate kube doesn't handle privileged container


fcrozat:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

podman generate kube is reporting an error if a container is privileged

**Steps to reproduce the issue:**

1. podman container create --name foobar --privileged registry.opensuse.org/opensuse/tumbleweed:latest

2. podman  generate kube foobar

**Describe the results you received:**
Error: linux devices: not yet implemented

**Describe the results you expected:**
# Generation of Kubernetes YAML is still under development!
#
# Save the output of this file and use kubectl create -f to import
# it into Kubernetes.
#
# Created with podman-2.2.0
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2021-01-06T14:17:11Z"
  labels:
    app: foobar
  name: foobar
spec:
  containers:
  - command:
    - /bin/bash
    env:
    - name: PATH
      value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
    - name: TERM
      value: xterm
    - name: container
      value: podman
    - name: HOSTNAME
    image: registry.opensuse.org/opensuse/tumbleweed:latest
    name: foobar
    resources: {}
    securityContext:
      allowPrivilegeEscalation: true
      capabilities: {}
      privileged: true
      readOnlyRootFilesystem: false
      seLinuxOptions: {}
    workingDir: /
status: {}
---
metadata:
  creationTimestamp: null
spec: {}
status:
  loadBalancer: {}

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

Version:      2.2.0
API Version:  2.1.0
Go Version:   go1.13.15
Built:        Wed Dec  2 01:00:00 2020
OS/Arch:      linux/amd64

**Output of `podman info --debug`:**

host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.21-1.1.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: unknown'
  cpus: 2
  distribution:
    distribution: '"opensuse-tumbleweed"'
    version: "20210102"
  eventLogger: journald
  hostname: localhost.localdomain
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.10.3-1-default
[...]

PR #9283, opened 2021-02-09
generate kube: do not set caps with --privileged

vrothberg:
Do not play with capabilities for privileged containers where all
capabilities will be set implicitly.

While the initital error report from #8897 is not reproducible,
privilged containers could not be executed due the bug at hand.

Fixes: #8897

### DIFF
--- a/libpod/kube.go
+++ b/libpod/kube.go
@@ -322,7 +322,8 @@ func containerToV1Container(c *Container) (v1.Container, []v1.Volume, *v1.PodDNS
 		return kubeContainer, kubeVolumes, nil, err
 	}
 
-	if len(c.config.Spec.Linux.Devices) > 0 {
+	// NOTE: a privileged container mounts all of /dev/*.
+	if !c.Privileged() && len(c.config.Spec.Linux.Devices) > 0 {
 		// TODO Enable when we can support devices and their names
 		kubeContainer.VolumeDevices = generateKubeVolumeDeviceFromLinuxDevice(c.Spec().Linux.Devices)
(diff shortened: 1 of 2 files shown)


baude:
LGTM

vrothberg:
Alright, hitting the same error as reported. Will have another look.

libpod/kube.go
```diff
@@ -625,13 +625,18 @@ func capAddDrop(caps *specs.LinuxCapabilities) (*v1.Capabilities, error) {
...
 func generateKubeSecurityContext(c *Container) (*v1.SecurityContext, error) {
-	priv := c.Privileged()
+	privileged := c.Privileged()
 	ro := c.IsReadOnly()
 	allowPrivEscalation := !c.config.Spec.Process.NoNewPrivileges
 
-	newCaps, err := capAddDrop(c.config.Spec.Process.Capabilities)
-	if err != nil {
-		return nil, err
+	var capabilities *v1.Capabilities
```
edsantiago, line 632: Shouldn't this be initialized? (I'm not a Gohead; I may be missing something obvious)

mheon:
Tests red. Otherwise LGTM

libpod/kube.go
vrothberg: That's no required here. Setting the container to privileged implies that all caps get set later on:
```
$ podman (fix-8897) $ podman top 5e5eb6e0fb43 capeff
EFFECTIVE CAPS
full
```

vrothberg:
This is looking good now :+1:

TomSweeneyRedHat:
LGTM
and all green if you want to cancel the hold @mheon

PR #9284, opened 2021-02-09
Support annotations from containers.conf

rhatdan:
Currently podman does not use the annotations specified in the
containers.conf. This PR fixes this.

### DIFF
--- a/pkg/specgen/generate/container.go
+++ b/pkg/specgen/generate/container.go
@@ -3,6 +3,7 @@ package generate
 import (
 	"context"
 	"os"
+	"strings"
 
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/podman/v2/libpod"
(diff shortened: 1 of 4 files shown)


rhatdan:
This is important for a blog I am writing on sharing supplemental groups into a container.

vrothberg (commented):
LGTM

pkg/specgen/generate/container.go
```diff
@@ -197,6 +198,15 @@ func CompleteSpec(ctx context.Context, r *libpod.Runtime, s *specgen.SpecGenerat
 		annotations[ann.ContainerType] = ann.ContainerTypeContainer
 	}
 
+	for _, v := range rtc.Containers.Annotations {
+		split := strings.SplitN(v, "=", 2)
+		k := split[0]
+		v := ""
+		if len(split) == 2 {
+			v = split[0]
```
Luap99, line 206:
```suggestion
			v = split[1]
```
wrong index

mheon:
Needs fix from @Luap99 then LGTM.

Concerned as to why tests are green despite the logic error.

rhatdan:
The =1 is not important to crun, if crun sees the value, it will allow the supplemental groups in.

mheon:
But this isn't the only important annotation that exists, and the test did not catch that the code was incorrect

rhatdan:
Good point now, the test looks for the value 1.

rhatdan:
I also added another PR here to fix a race condition in https://github.com/containers/podman/issues/9286.
This should help this PR get merged.

mheon:
LGTM once things go green

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #9289, opened 2021-02-09
apiv2 test fixes

edsantiago:
It's been a while since I last looked at these; some cruft
has crept in, generating noise and hence unreadable test
results. Clean it up:

 * remove pushd/popd in one subtest, replace with 'tar -C'.
   (Also remove confusing quotation marks). This removes
   spurious directory names from output.

 * in like(), show only first line of actual output.
   Some commands ('tree', 'generate kube') produce
   voluminous multi-line output, which is super useless
   and distracting when reading a test run.

 * Recognize that some queries will not generate output,
   e.g. HEAD requests and some POSTs. Deal with that.
   This fixes "curl.result.out: no such file" and "parse
   error" warnings.

 * In cleanup, 'podman rm -a' and 'rmi -af'; this gets
   rid of errors when deleting $WORKDIR. (EBUSY error
   when root, EPERM when rootless).

And, the original reason for poking in here: refactor the
wait-for-port part of start_server() into its own helper
function, so we can use it when starting a local registry
in 12-imagesMore. (Ref: #9270)

### DIFF
--- a/pkg/api/handlers/compat/images_push.go
+++ b/pkg/api/handlers/compat/images_push.go
@@ -1,7 +1,6 @@
 package compat
 
 import (
-	"context"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


edsantiago:
/hold

DO NOT MERGE (yet) - this PR is based on #9270. (I don't actually depend on it, in fact it's probably the converse, but I'd like that one to go in first because it's been in the pipeline for longer).

If for any reason #9270 fails CI or review I'll need to resubmit this, and will then decide whether to rebase on master (and have 9270 rebase here) or whether to rebase on a changed 9270. I am gambling on 9270 passing CI and review.

edsantiago:
Confirmed: much nicer cleaner [test log](https://api.cirrus-ci.com/v1/task/4841306260242432/logs/main.log)

TomSweeneyRedHat:
LGTM
assuming happy tests

edsantiago:
/hold cancel

#9270 is merged, so this should merge cleanly. Tests are green. @containers/podman-maintainers PTAL

PR #9291, opened 2021-02-09
hardening flags for fedora rpmbuilds

lsm5:
This commit sets the CGO_CFLAGS variable for hardening the Fedora rpm
binaries.
    
The flags used are the same as used in the official Fedora rpms.
Setting the flags in upstream spec would provide early warnings for
flag adjustments or other hardening issues.

@baude @cevich @rhatdan PTAL.

### DIFF
--- a/contrib/spec/podman.spec.in
+++ b/contrib/spec/podman.spec.in
@@ -380,6 +380,29 @@ tar zxf %{SOURCE1}
 %build
 mkdir _build
 pushd _build
+
+# These flags should work for all rpm distros and arches
+export CGO_CFLAGS="-O2 -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -ffat-lto-objects -fexceptions -fasynchronous-unwind-tables -fstack-protector-strong -fstack-clash-protection -D_GNU_SOURCE -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE"
+
+%if 0%{?fedora} || 0%{?rhel}
+# This flag is only present on RH-family distros
+export CGO_CFLAGS+=" -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1"
+%endif
+
+%ifnarch %{ix86}
+# Build fails on i686 with this flag
+export CGO_CFLAGS+=" -D_FILE_OFFSET_BITS=64"
+%endif
+
+%ifarch x86_64
+# Builds only on x86_64 with this flag
+export CGO_CFLAGS+=" -m64 -mtune=generic"
+%if 0%{?fedora} || 0%{?rhel} >= 8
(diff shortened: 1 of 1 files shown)


lsm5:
@edsantiago PTAL as well.

TomSweeneyRedHat:
LGTM
but definitely want someone with more spec background to review.

lsm5:
> LGTM
> but definitely want someone with more spec background to review.

yes, let's get @jnovy and @umohnani8 to look at this well.

edsantiago:
I feel embarrassingly unqualified to review this. All I can do is confirm that the flags match [those in fedpkg master](https://src.fedoraproject.org/rpms/podman/blob/rawhide/f/podman.spec#_421-427), which they do apart from the `ix86` exception. I don't know if `ix86` is the right arch macro, nor do I know if `+=` will work in this environment (i.e. can we safely assume that the rpm build script will run under bash?), and don't even know how to verify or test. I have to assume that if either doesn't work, we'll find out.

With those caveats, LGTM.

rhatdan:
@saschagrunert @vrothberg 
Does OpenSUSE use this spec file?  I would doubt they have the -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 file?

lsm5:
> I feel embarrassingly unqualified to review this. All I can do is confirm that the flags match [those in fedpkg master](https://src.fedoraproject.org/rpms/podman/blob/rawhide/f/podman.spec#_421-427), which they do apart from the `ix86` exception. I don't know if `ix86` is the right arch macro, nor do I know if `+=` will work in this environment (i.e. can we safely assume that the rpm build script will run under bash?), and don't even know how to verify or test. I have to assume that if either doesn't work, we'll find out.
> 
> With those caveats, LGTM.

Good point, I've yet to update rawhide branch. This was a copy-paste from the f33 branch where v3.0.0-rc3 koji build was failing for i686. You can see the f33 change here: https://src.fedoraproject.org/rpms/podman/blob/f33/f/podman.spec#_463-472

vrothberg:
> @saschagrunert @vrothberg
> Does OpenSUSE use this spec file? I would doubt they have the -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 file?

*SUSE is using their own spec file.

lsm5:
> @saschagrunert @vrothberg
> Does OpenSUSE use this spec file? I would doubt they have the -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 file?

Ack, I'll put these under a fedora/centos conditional so it won't bother other distros.

lsm5:
adjusted the flag conditionals with comments explaining why.

edsantiago:
LGTM. The comments are helpful, thank you.

If for any reason you need to resubmit, you may want to add the magic CI-colon-DOCS string to bypass unneeded CI.

vrothberg (commented):
/lgtm

PR #9294, opened 2021-02-09
Update golangci-lint

matejvasek:
Update golangci-lint.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,7 @@ run:
   deadline: 5m
   skip-dirs-use-default: true
   skip-dirs:
+    - dependencies
     - contrib
     - dependencies
     - test
@@ -17,6 +18,34 @@ linters:
   enable-all: true
   disable:
     # All these break for one reason or another
+    - nolintlint
+    - gocognit
+    - stylecheck
+    - testpackage
+    - goerr113
+    - exhaustivestruct
+    - errorlint
+    - wrapcheck
+    - paralleltest
+    - wsl
+    - godox
+    - tparallel
+    - gomnd
+    - nlreturn
+    - noctx
+    - nestif
+    - predeclared
+    - thelper
+    - ifshort
+    - staticcheck
+    - forbidigo
+    - exhaustive
+    - whitespace
+    - unparam
+    - gofumpt
+    - gci
+    - godot
+    - makezero
     - dupl
     - funlen
     - gochecknoglobals
--- a/pkg/domain/infra/abi/manifest.go
+++ b/pkg/domain/infra/abi/manifest.go
@@ -1,5 +1,3 @@
-// +build !remote
-
 package abi
 
 import (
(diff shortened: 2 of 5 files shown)


.golangci.yml
```diff
@@ -17,6 +18,40 @@ linters:
...
+    - whitespace
+    - unparam
+    - gofmt
+    - gofumpt
+    - goimports
+    - gci
+    - asciicheck
+    - godot
+    - makezero
+    - unused
```
matejvasek, line 54: These seems to crash linter, the about are just failing.

matejvasek, line 54: These seems to crash linter, the about are just failing.

pkg/domain/infra/abi/manifest.go
```diff
@@ -1,5 +1,3 @@
-// +build !remote
```
matejvasek, line 1: Without removing this the `tunnel` linting will fail.

matejvasek:
This update should lower memory requirements for linter. With the older version I saw 8.9GiB memory consumption. With the newer just 1.5GB.

.golangci.yml
```diff
@@ -17,6 +18,38 @@ linters:
   enable-all: true
   disable:
     # All these break for one reason or another
+    - nolintlint
```
matejvasek, line 21: I think that overtime we want to actually fix some of those errors instead of suppressing them.

matejvasek: I don't not mean `nolintlint` specifically here, mean all suppressed checkers.

.golangci.yml
```diff
@@ -17,6 +18,38 @@ linters:
...
+    - predeclared
+    - thelper
+    - ifshort
+    - staticcheck
+    - forbidigo
+    - govet
+    - exhaustive
+    - whitespace
+    - unparam
+    - gofmt
```
vrothberg, line 47: The list looks way to big to me. `gofmt` is a standard linter. If that doesn't work, we should really fix the underlying issue.

matejvasek: I enabled `gofmt` `goimports` and `govet`.

vrothberg: Thanks a lot!

vrothberg (commented):
LGTM, thanks!

Now we have a good base to start enabling more linters.

matejvasek:
One thing I don't understand: event without this PR there is a lint error:
```
conmon/cmd/conmon-config/conmon-config.go:33:13: unusedresult: result of fmt.Errorf call not used (govet)
                fmt.Errorf(err.Error())
```
however CI is not bothered by it.

matejvasek:
@vrothberg flake?

vrothberg:
> @vrothberg flake?

Restarted :) Looks like one

vrothberg:

pkg/domain/infra/abi/manifest.go
matejvasek: @rhatdan is this ok?

saschagrunert (approved):
LGTM

PR #9296, opened 2021-02-09
run buildah bud tests using podman

edsantiago:
Proof of concept, guaranteed to fail in CI (if it even makes it
as far as CI, which I doubt, because my .cirrus.yml changes
are a wild guess.)

Checking in just to get the conversation started.

### DIFF
--- a/test/buildah-bud/buildah-tests.diff
+++ b/test/buildah-bud/buildah-tests.diff
@@ -1,0 +1,192 @@
+From c85882a8f7fb6efbf4d59dfe8340bfbef57ccd48 Mon Sep 17 00:00:00 2001
+From: Ed Santiago <santiago@redhat.com>
+Date: Tue, 9 Feb 2021 17:28:05 -0700
+Subject: [PATCH] tweaks for running buildah tests under podman
+
+Signed-off-by: Ed Santiago <santiago@redhat.com>
+---
+ tests/bud.bats     | 26 ++++++++++++++++----------
+ tests/helpers.bash | 28 ++++++++++++++++++++++++----
+ 2 files changed, 40 insertions(+), 14 deletions(-)
+
+diff --git a/tests/bud.bats b/tests/bud.bats
+index 1efc3c58..9a39d594 100644
+--- a/tests/bud.bats
++++ b/tests/bud.bats
+@@ -4,7 +4,7 @@ load helpers
+
+ @test "bud with a path to a Dockerfile (-f) containing a non-directory entry" {
+   run_buildah 125 bud -f ${TESTSDIR}/bud/non-directory-in-path/non-directory/Dockerfile
+-  expect_output --substring "non-directory/Dockerfile: not a directory"
++  expect_output --substring "Error: context must be a directory:"
+ }
+
+ @test "bud with --dns* flags" {
+@@ -95,6 +95,7 @@ symlink(subdir)"
+ }
+
+ @test "bud-flags-order-verification" {
++  skip "N/A under podman"
+   run_buildah 125 bud /tmp/tmpdockerfile/ -t blabla
+   check_options_flag_err "-t"
+
+@@ -1324,13 +1325,13 @@ function _test_http() {
+ @test "bud with dir for file but no Dockerfile in dir" {
+   target=alpine-image
+   run_buildah 125 bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/empty-dir ${TESTSDIR}/bud/empty-dir
+-  expect_output --substring "no such file or directory"
++  expect_output --substring "Error: context must be a directory:"
+ }
+
+ @test "bud with bad dir Dockerfile" {
+   target=alpine-image
+   run_buildah 125 bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/baddirname ${TESTSDIR}/baddirname
+-  expect_output --substring "no such file or directory"
++  expect_output --substring "Error: context must be a directory:"
+ }
+
+ @test "bud with ARG before FROM default value" {
+@@ -1742,7 +1743,9 @@ _EOF
+   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test-img-2 --build-arg TEST=foo -f Dockerfile4 ${TESTSDIR}/bud/build-arg
+   run_buildah inspect -f '{{.FromImageID}}' test-img-2
+   argsid="$output"
+-  [[ "$argsid" != "$initialid" ]]
++  if [[ "$argsid" == "$initialid" ]]; then
++      die ".FromImageID of test-img-2 ($argsid) == same as test-img, it should be different"
++  fi
+
+   # Set the build-arg via an ENV in the local environment and verify that the cached layers are not used
+   export TEST=bar
+@@ -1795,6 +1798,7 @@ _EOF
+ }
+
+ @test "bud without any arguments should succeed" {
++  skip "does not work under podman"
+   cd ${TESTSDIR}/bud/from-scratch
+   run_buildah bud --signature-policy ${TESTSDIR}/policy.json
+ }
+@@ -1802,7 +1806,7 @@ _EOF
+ @test "bud without any arguments should fail when no Dockerfile exist" {
+   cd $(mktemp -d)
+   run_buildah 125 bud --signature-policy ${TESTSDIR}/policy.json
+-  expect_output --substring "no such file or directory"
++  expect_output "Error: no context directory and no Containerfile specified"
+ }
+
+ @test "bud with specified context should fail if directory contains no Dockerfile" {
+@@ -1815,16 +1819,17 @@ _EOF
+   DIR=$(mktemp -d)
+   mkdir -p "$DIR"/Dockerfile
+   run_buildah 125 bud --signature-policy ${TESTSDIR}/policy.json "$DIR"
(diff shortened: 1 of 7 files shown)


edsantiago:
What the heck was that about? (The 'Check Automation' failure with golang-ci-lint)? Did someone re-run it manually, or did it restart automatically? 

In any case, here's the new diagram. New step is dark purple near center right.
cirrus-map-test_buildah_bud

edsantiago:
Hey, it worked! (Well, "it ran!"): [log](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6651056966860800/html/bud-podman-fedora-33-root-host.log.html) [the remote test is hanging, no useful info in log, and I'm not going to stay up to wait]

The `--dns` failure (32) looks real. The symlink ones (78-81) are way over my head, might merit a look-see from someone knowledgeable. Some of the others are exit-code or error-message differences, I can work around those. (I'm thinking of adding a generic rule to change expected-status-code `2` to `125`; will try that tomorrow).

Some of the `skips`, as I found out by looking at the sources, happen because of a test like this:
```
   ...
   buildah do something
   buildah from that-image
}
```
That is: all the interesting stuff happens, then there's a final `buildah from` that doesn't actually do anything, simply confirms that the image exists. I need to cogitate on that because I have a hunch there's a simple way to detect or fix those; we could then greatly improve our coverage.

rhatdan:
podman image exists? replace podman from  for those use cases?

edsantiago:
I'm actually leaning toward some sort of `podman run` + `podman commit` combination, because in reviewing `bud.bats` I see a lot of cases like
```
   buildah from <img>
   buildah run <ctr> <something>
```
It's getting complicated really fast, though.

edsantiago:
I had a tiny insight and decided to simply use `buildah` itself for everything other than `bud`: that is, convert `buildah bud` to `podman build`, and keep everything else the same (`buildah from`, `inspect`, `images`, `run`). [Result](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5433052291661824/html/bud-podman-fedora-33-root-host.log.html). Pass count is now 387, with 94 failures which I'll now start looking at.

rhatdan:
SGTM

edsantiago:
Good news: given enough `skips`, we have [green tests](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5780713858400256/html/bud-podman-fedora-33-root-host.log.html). (Issues filed for the majority of those skips, and I expect most of those to be fixed).

Less good news: the `podman-remote build` work is really hard, and I keep running into snags (most recent one: dealing with the `--root` option which is a deeply ingrained part of buildah tests). How necessary is it to get these tests running under podman-remote?

rhatdan:
I am fine with skipping them for now.  The amount of bugs this effort has revealed thus far has been great.

test/buildah-bud/buildah-tests.diff
```diff
@@ -0,0 +1,192 @@
...
+
+ @test "bud with ARG before FROM default value" {
+@@ -1742,7 +1743,9 @@ _EOF
+   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test-img-2 --build-arg TEST=foo -f Dockerfile4 ${TESTSDIR}/bud/build-arg
+   run_buildah inspect -f '{{.FromImageID}}' test-img-2
+   argsid="$output"
+-  [[ "$argsid" != "$initialid" ]]
++  if [[ "$argsid" == "$initialid" ]]; then
++      die ".FromImageID of test-img-2 ($argsid) == same as test-img, it should be different"
++  fi
```
edsantiago, line 56: I plan to submit this as a change to buildah itself; will do so after this PR merges, as a test of the conflict handling when vendoring.

test/buildah-bud/buildah-tests.diff
```diff
@@ -0,0 +1,192 @@
...
++  skip "podman requires a directory, not a Dockerfile"
+   DIR=$(mktemp -d)
+   echo "FROM alpine" > "$DIR"/Dockerfile
+   run_buildah 0 bud --signature-policy ${TESTSDIR}/policy.json "$DIR"/Dockerfile
+@@ -1876,7 +1881,7 @@ _EOF
+
+ @test "bud-squash-hardlinks" {
+   _prefetch busybox
+-  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --squash ${TESTSDIR}/bud/layers-squash/Dockerfile.hardlinks
++  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --squash -f Dockerfile.hardlinks ${TESTSDIR}/bud/layers-squash
```
edsantiago, line 102: Am ambivalent about submitting this too as a buildah PR. It would help reduce the maintenance burden of carrying this diff, but OTOH it might be good to keep this as an extra test of buildah command-line handling

test/buildah-bud/buildah-tests.diff
```diff
@@ -0,0 +1,192 @@
...
+ }
+
+ @test "bud with additional directory of devices" {
+@@ -2023,6 +2028,7 @@ _EOF
+ }
+
+ @test "bud pull never" {
++  skip "FIXME: podman issue #9573"
```
edsantiago, line 110: Addressed by #9631, which is having trouble in CI

test/buildah-bud/buildah-tests.diff
```diff
@@ -0,0 +1,192 @@
...
++        fi
++        podman_or_buildah=${PODMAN_BINARY}
++
++        # podman always exits 125 where buildah exits 1 or 2
++        case $expected_rc in
++            1|2)   expected_rc=125 ;;
++        esac
++    fi
++    local cmd_basename=$(basename ${podman_or_buildah})
++
```
edsantiago, line 175: I'm actually quite pleased at how small a change is needed to this file.

Unfortunately, if we want to test podman-remote, it's a whole different story. Those changes are huge, and I choose to defer them for a future PR.

edsantiago (commented):
@containers/podman-maintainers @containers/buildah-maintainers PTAL. I'm about as happy with this as I'm likely to be.

I realize this is unorthodox and even nerve-racking, with unknown implications for each buildah revendor. Please review with care: not just the code, but considerations I may have missed. TIA.

test/buildah-bud/buildah-tests.diff
rhatdan: Yes I think we can do a better job of coordinating these errors between podman and buildah.

test/buildah-bud/buildah-tests.diff
rhatdan: I would like to see more consistency between the errors in buildah bud and podman build.

rhatdan (approved):
LGTM

rhatdan:
This is a huge increase of test coverage on podman build, nice work @edsantiago I think we should work to get some of these fixes back into buildah to make the tests easier to maintain.

rhatdan:
@vrothberg @cevich @TomSweeneyRedHat PTAL

vrothberg (commented):
/lgtm

Really nice work, Ed!

cevich:
@rhatdan @edsantiago Thought:  Maintaining these diffs will compound over time as podman and buildah branch.  Since updates to release branches do happen on occasion, it may become difficult for non-experts (in these tests) to produce working diffs for their changes and/or vendor updates.  This may unduly impact important security-related PRs (i.e. where these tests failing, blocks the merge).

Perhaps it would help to provide a `.cirrus.yml` "escape hatch" for this task, [similar to what we do for the static build task.](https://github.com/edsantiago/libpod/blob/3baa9da4edec72a70aec4aa25a29b4afd82ed689/.cirrus.yml#L330-L332).  A simple thing like that, where someone can un-comment it in a panic/emergency situation, to be addressed after the critical security change is merged.

edsantiago:
@cevich thanks for the reminder. I had wanted to include a bypass, but wanted to do it via a mechanism like `[CI:DOCS]` where the bypass is in the PR comments. Do you have a suggestion on how to do that? IIRC Cirrus yaml doesn't allow adding conditions (like `&& $CIRRUS_CHANGE_TITLE !=~ '.*SKIP:BUILDAH.*'`) to a star-anchor thing.
```
# allow_failures: $CI == $CI
```
This makes me uncomfortable because it's essentially the same as removing the entire buildah_bud task, except kind of wishy-washy: it's saying "we will never get around to fixing this". If that's the case, I think it's much better to rip out the buildah_bud task entirely.

cevich:
> IIRC Cirrus yaml doesn't allow adding conditions

Cirrus-CI allows this, we simply don't do it because adding it to work with the `*.CI:DOCS.*` feature, makes human-brains explode.

> `# allow_failures: $CI == $CI`

This is intended as temporary mechanism and reminder: "This task may become unstable, it's okay to skip it in extraordinary circumstances **but** if you do, please also open an issue."

Realistically, we're only running the full set of tasks on a 3-5 branches IIRC.  As they age, they join the majority of other defunct branches where I end up gutting `.cirrus.yml` down to only the 'smoke' test.  So the risk pool here is large enough to be annoying, but small enough I think having an escape-hatch-via-uncomment is probably fine.

Otherwise, something like this might work:

`    only_if: $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' && $CIRRUS_CHANGE_TITLE !=~ '.*CI:NOBUILDAH.*'`

...but it's very brain-hurty for me to read :confused:

PR #9297, opened 2021-02-10
Docker [APIv2] push sends digest in response body

matejvasek:
Push endpoint will return message containing image digest.


vrothberg (commented):
LGTM, thanks!

@jwhonce @baude PTAL

PR #9299, opened 2021-02-10
Add shell completion tests for secrets

Luap99:
Add the SECRET keyword to the shell completion test. Also update the
use line for podman secret create to use `NAME` instead of `SECRET`.
This matches the other commands such as network/volume create.

### DIFF
--- a/test/system/600-completion.bats
+++ b/test/system/600-completion.bats
@@ -157,7 +157,17 @@ function check_shell_completion() {
                     # resume
                     ;;&
 
-                *PATH* | *CONTEXT* | *KUBEFILE* | *COMMAND* | *ARG...* | *URI*)
+                *SECRET*)
+                    run_completion "$@" $cmd "${extra_args[@]}" ""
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM

@ashley-cui PTAL

test/system/600-completion.bats
```diff
@@ -274,6 +291,9 @@ function _check_completion_end() {
     check_shell_completion
 
     # cleanup
+    run_podman secret rm $random_secret_name
+    rm -f $secret_file
```
edsantiago, line 295: FYI for next time: cleanup isn't necessary here, because `$PODMAN_TMPDIR` is new for each test and wiped in teardown. My purpose was to make each test line be important, so the test writer/maintainer wouldn't have to worry about clutter.

edsantiago:
Nice work @Luap99. LGTM but will leave final approval to @ashley-cui.

ashley-cui:
LGTM 
/lgtm

PR #9304, opened 2021-02-10
[v1.0] Correct handling of capabilities

mheon:
Ensure that capabilities are properly handled for non-root users in privileged containers. We do not want to give full caps, but instead only CapInh and CapEff (others should be all-zeroes).


baude:
LGTM

PR #9305, opened 2021-02-10
[v1.6.4-rhel] Restrict caps of non-root processes in priv containers

mheon:
Non-root processes in privileged containers do not get all caps, and instead only have Bounding and Inherited set to all; the other caps are left empty. This behavior matches Docker and newer Podman releases.


baude:
LGTM

Issue #9303, opened 2021-02-10
Podman inspect returns invalid values for unlimited ulimits


mpolden:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. Run a container with unlimited core size:
   ```
   podman run --rm --name test --ulimit core=-1:-1 -it debian:latest /bin/bash
   ```

2. Inspect the container ulimits:
    ```
    podman inspect test | grep -B 1 -A 3 -F RLIMIT_CORE
    ```

**Describe the results you received:**

Inspect returns the following:

```json
                {
                    "Name": "RLIMIT_CORE",
                    "Soft": 18446744073709551615,
                    "Hard": 18446744073709551615
                },
```

**Describe the results you expected:**

```json
                {
                    "Name": "RLIMIT_CORE",
                    "Soft": -1,
                    "Hard": -1
                },
```

**Additional information you deem important (e.g. issue happens only occasionally):**

The returned value is 2^64-1, so this probably happens because Podman uses unsigned integers for these values and overflows.

**Output of `podman version`:**

```
Version:      2.0.5
API Version:  1
Go Version:   go1.14.7
Built:        Wed Sep 23 16:18:02 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.15.1
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.20-2.module+el8.3.0+8221+97165c3f.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.20, commit: 77ce9fd1e61ea89bd6cdc621b07446dd9e80e5b6'
  cpus: 24
  distribution:
    distribution: '"rhel"'
    version: "8.3"
  eventLogger: file
  hostname: <snip>
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 100
      size: 1
    uidmap:
    - container_id: 0
      host_id: 87591
      size: 1
  kernel: 4.18.0-240.10.1.el8_3.x86_64
  linkmode: dynamic
  memFree: 16772780032
  memTotal: 25032122368
  ociRuntime:
    name: runc
    package: runc-1.0.0-68.rc92.module+el8.3.0+8221+97165c3f.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.2-dev'
  os: linux
  remoteSocket:
[...]

PR #9309, opened 2021-02-10
Display correct value for unlimited ulimit

baude:
When doing a container inspect on a container with unlimited ulimits,
the value should be -1.  But because the OCI spec requires the ulimit
value to be uint64, we were displaying the inspect values as a uint64 as
well.  Simple change to display as an int64.

Fixes: #9303


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #9311, opened 2021-02-10
apiv2: handle docker-java clients pulling

deadNightTiger:
When docker-java calls images/create?fromImage=x, it expects two things for a successful response:
that both "error" and "errorDetail" are not set,
and that the "status" message contains one of five hard-coded strings ("Download complete" being one of them).

See:
https://github.com/docker-java/docker-java/blob/master/docker-java-api/src/main/java/com/github/dockerjava/api/model/PullResponseItem.java#L28-L39


baude:
@deadNightTiger you need to sign your commit and repush.

$ git commit --amend -s
$ git push -f ...

baude:
code lgtm

rhatdan:
/lgtm
Thanks @deadNightTiger

PR #9317, opened 2021-02-10
[WIP] ToParams() reflection-free experiment

matejvasek:
Attempt to make `ToParam()` method reflection-free by code-generation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -473,7 +473,7 @@ podman-remote-%-release:
 .PHONY: generate-bindings
 generate-bindings:
 ifneq ($(shell uname -s), Darwin)
-	GO111MODULE=off $(GO) generate ./pkg/bindings/... ;
+	GO111MODULE=on $(GO) generate ./pkg/bindings/... ;
 endif
 
 .PHONY: docker-docs
--- a/pkg/bindings/containers/types_execstartandattach_options.go
+++ b/pkg/bindings/containers/types_execstartandattach_options.go
@@ -5,11 +5,7 @@ import (
 	"io"
 	"net/url"
 	"reflect"
-	"strings"
-
-	"github.com/containers/podman/v2/pkg/bindings/util"
-	jsoniter "github.com/json-iterator/go"
-	"github.com/pkg/errors"
+	"strconv"
 )
 
 /*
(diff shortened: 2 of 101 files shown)


Makefile
```diff
@@ -473,7 +473,7 @@ podman-remote-%-release:
 .PHONY: generate-bindings
 generate-bindings:
 ifneq ($(shell uname -s), Darwin)
-	GO111MODULE=off $(GO) generate ./pkg/bindings/... ;
+	GO111MODULE=on $(GO) generate ./pkg/bindings/... ;
```
matejvasek, line 476: @vrothberg why was this changed to `off` before?

pkg/bindings/containers/types_execstartandattach_options.go
```diff
@@ -26,53 +22,35 @@ func (o *ExecStartAndAttachOptions) Changed(fieldName string) bool {
...
+
 	if o == nil {
 		return params, nil
 	}
-	json := jsoniter.ConfigCompatibleWithStandardLibrary
-	s := reflect.ValueOf(o)
-	if reflect.Ptr == s.Kind() {
-		s = s.Elem()
+
+	if o.OutputStream != nil {
```
matejvasek, line 30: Those options structs are supposed to be send over wire, right? How can one serialize `io.Writer/Reader`?

pkg/bindings/images/types_import_options.go
```diff
@@ -24,53 +19,27 @@ func (o *ImportOptions) Changed(fieldName string) bool {
...
 	if o == nil {
 		return params, nil
 	}
-	json := jsoniter.ConfigCompatibleWithStandardLibrary
-	s := reflect.ValueOf(o)
-	if reflect.Ptr == s.Kind() {
-		s = s.Elem()
+
+	if o.Changes != nil {
+		panic("*** GENERATOR DOESN'T IMPLEMENT THIS YET ***")
```
matejvasek, line 28: This crashes one of tests, however I believe it wasn't functioning correctly even before. `o.Changes` is `map[string]string` and therefore it was ignored before.

```go
lowerCaseKeys[iter.Key().Interface().(string)] = iter.Value().Interface().([]string)
```
If I understand it correcty `iter.Value().Interface().([]string)` will result in null if values happens to be just `string` instead of slice of strings.

matejvasek: Oh I see `o.Changes` is not `map[string]string`, it's `*[]string`! I don't see any reason to use `*[]string` since `[]string` is already nullable, weird.

matejvasek: The original code is removing pointer:
```golang
		if reflect.Ptr == f.Kind() {
			f = f.Elem()
		}
```

pkg/bindings/containers/types_execstartandattach_options.go
Luap99: They are not supposed to be send over wire. see: https://github.com/containers/podman/blob/afe4ce6b1ce6a5ba21472f51defd243bdd7750b3/pkg/bindings/containers/attach.go#L500-L519

matejvasek: @Luap99  So only some struct fields (structurally simple ones) are serialized. The other may be handled differently (e.g. post as JSON in body). How it's decided? Is it solely based on type. For instance `int` fields are include in uri params, but struct/interface are not?

pkg/bindings/images/types_import_options.go
matejvasek: Why is `map[string][]string` supported but `map[string]string` isn't?

pkg/bindings/containers/types_execstartandattach_options.go
Luap99: I have no idea. @baude wrote the bindings stuff

rhatdan:
Since the other PR is going in, is this still needed?

rhatdan:
@matejvasek What is happening with this PR?  Is this still something you are working on? Or should we close it?

Issue #9315, opened 2021-02-10
'podman ps --format json' no longer has a numeric UNIX timestamp in "Created"


debarshiray:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman ps --format json` no longer has a numeric UNIX timestamp in `Created`. In the release candidates for Podman 3.0.0, it's a string that looks like `"2021-01-19T19:17:40.179387486+01:00"`. Earlier, in Podman 2.2.1, it used to be a number that would get parsed as a `float64`.

Note that:

* `podman images --format json` continues to have a numeric UNIX timestamp in `Created`, and this particular encoding for containers was discussed during the Podman 2.x timeframe in https://github.com/containers/podman/issues/6594.

* Toolbox doesn't use the `CreatedAt` fields. So this string representation in `podman ps` can be moved to `CreatedAt` because that will make it consistent with `podman images`.

**Steps to reproduce the issue:**

1. `$ podman ps --all --format json | less`

**Describe the results you received:**

`Created` is a string.

**Describe the results you expected:**

`Created` should've been a numeric UNIX timestamp.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.0.0-rc3
API Version:  3.0.0
Go Version:   go1.15.7
Built:        Mon Feb  8 22:16:50 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.2
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 0f53fb68333bdead5fe4dc5175703e22cf9882ab'
  cpus: 4
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: kolache
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.10.13-200.fc33.x86_64
  linkmode: dynamic
  memFree: 124710912
  memTotal: 8102649856
  ociRuntime:
    name: crun
[...]

PR #9320, opened 2021-02-11
container ps json format miscue

baude:
when printing out json format, we mistakenly changed the Created field
output to be a time.time in a different commit.  This allows for
override of the Created field to be a unix ts as type int64.

Fixes: #9315

### DIFF
--- a/cmd/podman/containers/ps.go
+++ b/cmd/podman/containers/ps.go
@@ -142,11 +142,19 @@ func checkFlags(c *cobra.Command) error {
 }
 
 func jsonOut(responses []entities.ListContainer) error {
-	r := make([]entities.ListContainer, 0)
+	type jsonFormat struct {
+		entities.ListContainer
+		Created int64
+	}
+	r := make([]jsonFormat, 0)
 	for _, con := range responses {
 		con.CreatedAt = units.HumanDuration(time.Since(con.Created)) + " ago"
 		con.Status = psReporter{con}.Status()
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM

cmd/podman/containers/ps.go
```diff
@@ -142,11 +142,19 @@ func checkFlags(c *cobra.Command) error {
...
+		Created int64
+	}
+	r := make([]jsonFormat, 0)
 	for _, con := range responses {
 		con.CreatedAt = units.HumanDuration(time.Since(con.Created)) + " ago"
 		con.Status = psReporter{con}.Status()
-		r = append(r, con)
+		jf := jsonFormat{
+			ListContainer: con,
+			Created:       con.Created.UnixNano(),
```
debarshiray, line 155: This doesn't look correct. Earlier it used the be the UNIX timestamp in seconds.

Like I mentioned in https://github.com/containers/podman/issues/9315, Toolbox no longer uses the `CreatedAt` field for containers, because its different for images anyway. So maybe `CreatedAt` could be repurposed for a more precise time representation? That could possibly make it consistent with images.

debarshiray: Or, is there a way to programmatically differentiate between a timestamp in seconds and one in nanoseconds? Then Toolbox could handle that while parsing the JSON value.

PR #9321, opened 2021-02-11
[backport v3.0] hardening flags for fedora rpmbuilds

lsm5:
This commit sets the CGO_CFLAGS variable for hardening the Fedora rpm
binaries.

The flags used are the same as those in the official Fedora rpms.
Setting the flags in upstream spec would provide early warnings for
flag adjustments or other hardening issues.

(cherry picked from commit 21deafba85b21aa76ccd464c620dfa45085fc90f)

@rhatdan @mheon @vrothberg @TomSweeneyRedHat PTAL

### DIFF
--- a/contrib/spec/podman.spec.in
+++ b/contrib/spec/podman.spec.in
@@ -380,6 +380,29 @@ tar zxf %{SOURCE1}
 %build
 mkdir _build
 pushd _build
+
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM

PR #9324, opened 2021-02-11
[v2.2] Cirrus: Fix running Validate task on branches

cevich:
Followup to 6917272b

Despite skipping the `Smoke` check, it was observed on a *new* branch,
the `validate` task (specifically `git-validation`) will fail.  This
is because:

* `$CIRRUS_LAST_GREEN_CHANGE` will be empty on a new branch.
* `$CIRRUS_BASE_SHA` is always empty for runs triggered by branch-push
* `$EPOCH_TEST_COMMIT` will be set to `YOU_FOUND_A_BUG`.

Fix this by eliminating the `Smoke` task entirely, simplifying all
the `make validate` operations into the `validate` cirrus task.  Ensure
this task does not run when a new branch or tag is pushed.

Also, eliminate the `$CIRRUS_BUILD_ID` value as it's confusing and not
actually used anywhere.  It was formerly used for building VM images,
but this has moved to another repo entirely.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #9326, opened 2021-02-11
[v2.2.1-rhel] Cirrus: Fix running Validate task on branches

cevich:
Despite skipping the `Smoke` check, it was observed on a *new* branch,
the `validate` task (specifically `git-validation`) will fail.  This
is because:

* `$CIRRUS_LAST_GREEN_CHANGE` will be empty on a new branch.
* `$CIRRUS_BASE_SHA` is always empty for runs triggered by branch-push
* `$EPOCH_TEST_COMMIT` will be set to `YOU_FOUND_A_BUG`.

Fix this by eliminating the `Smoke` task entirely, simplifying all
the `make validate` operations into the `validate` cirrus task.  Ensure
this task does not run when a new branch or tag is pushed.

Also, eliminate the `$CIRRUS_BUILD_ID` value as it's confusing and not
actually used anywhere.  It was formerly used for building VM images,
but this has moved to another repo entirely.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,7 +6,7 @@ env:
     #### Global variables used for all tasks
     ####
     # Name of the ultimate destination branch for this CI run, PR or post-merge.
-    DEST_BRANCH: "v2.2"
+    DEST_BRANCH: "v2.2.1-rhel"
     # Overrides default location (/tmp/cirrus) for repo clone
     GOPATH: &gopath "/var/tmp/go"
     GOBIN: "${GOPATH}/bin"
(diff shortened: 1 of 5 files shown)


cevich:
Ref. fix for https://cirrus-ci.com/task/6628393229746176

rhatdan:
LGTM
@edsantiago PTAL

.cirrus.yml
```diff
@@ -6,7 +6,7 @@ env:
     #### Global variables used for all tasks
     ####
     # Name of the ultimate destination branch for this CI run, PR or post-merge.
-    DEST_BRANCH: "v2.2"
+    DEST_BRANCH: "v2.2.1-rhel"
```
edsantiago, line 9: The other PR (#9324) did not change this line. Are you fixing an oversight in the setup of this `-rhel` branch? If so, could you update the commit message accordingly?

edsantiago:
Reviewed by eyeballing diffs against #9063. In this case (unlike #9324) diffing against `master` proved impossible: too many unrelated changes. 

Mostly LGTM, with one concern about an unannotated branch-name change.

.cirrus.yml
cevich: yes

cevich:
@edsantiago PTAL, this should be good-to-go now.

PR #9331, opened 2021-02-11
Enable more golangci-lint linters

Luap99:
Enable the stylecheck, golint and whitespace linter.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,7 +39,6 @@ linters:
     - staticcheck
     - forbidigo
     - exhaustive
-    - whitespace
     - unparam
     - gofumpt
     - gci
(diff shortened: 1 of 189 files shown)


mheon:
Should we exclude the generated code in the bindings? I am concerned we'll overwrite all the changes there once the generator is re-run.

Maybe the generator should add a per-file nolint when it generates?

Otherwise LGTM

Luap99:
@mheon I fixed the lint issues in the generator and run make generate-bindings afterwards. I don't think we should add nolint unless there is a important reason.

mheon:
Works for me

vrothberg (commented):
LGTM, thanks!

PR #9302, opened 2021-02-10
utils: takes the longest path on cgroup v1

giuseppe:
now getCgroupProcess takes the longest path on cgroup v1, instead of
complaining if the paths are different.

This should help when --cgroups=split is used on cgroup v1 and the
process cgroups look like:

$ cat /proc/self/cgroup
11:pids:/user.slice/user-0.slice/session-4.scope
10:blkio:/
9:cpuset:/
8:devices:/user.slice
7:freezer:/
6:memory:/user.slice/user-0.slice/session-4.scope
5:net_cls,net_prio:/
4:hugetlb:/
3:cpu,cpuacct:/
2:perf_event:/

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #9332, opened 2021-02-12
[3.0] utils: takes the longest path on cgroup v1

giuseppe:
backport of: https://github.com/containers/podman/pull/9302

now getCgroupProcess takes the longest path on cgroup v1, instead of
complaining if the paths are different.

This should help when --cgroups=split is used on cgroup v1 and the
process cgroups look like:

$ cat /proc/self/cgroup
11:pids:/user.slice/user-0.slice/session-4.scope
10:blkio:/
9:cpuset:/
8:devices:/user.slice
7:freezer:/
6:memory:/user.slice/user-0.slice/session-4.scope
5:net_cls,net_prio:/
4:hugetlb:/
3:cpu,cpuacct:/
2:perf_event:/

Signed-off-by: Giuseppe Scrivano gscrivan@redhat.com

### DIFF
--- a/test/e2e/common_test.go
+++ b/test/e2e/common_test.go
@@ -600,6 +600,13 @@ func SkipIfNotRootless(reason string) {
 	}
 }
 
+func SkipIfNotSystemd(manager, reason string) {
+	checkReason(reason)
+	if manager != "systemd" {
+		ginkgo.Skip("[notSystemd]: " + reason)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
LGTM, restarted the flakes.

Great work, @giuseppe !

TomSweeneyRedHat:
LGTM

PR #9342, opened 2021-02-12
Enable more golangci-lint linters

Luap99:
Cleanup the golangci.yml file and enable more linters.

`pkg/spec` and `iopodman.io` is history. The vendor directory
is excluded by default. The dependencies dir was listed twice.

Fix the reported problems in `pkg/specgen` because that was also
excluded by `pkg/spec`.

Enable the structcheck, typecheck, varcheck, deadcode and depguard
linters.

### DIFF
--- a/pkg/specgen/generate/kube/seccomp.go
+++ b/pkg/specgen/generate/kube/seccomp.go
@@ -11,6 +11,7 @@ import (
 
 // KubeSeccompPaths holds information about a pod YAML's seccomp configuration
 // it holds both container and pod seccomp paths
+// nolint:golint
 type KubeSeccompPaths struct {
 	containerPaths map[string]string
 	podPath        string
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,14 +4,10 @@ run:
   deadline: 5m
   skip-dirs-use-default: true
   skip-dirs:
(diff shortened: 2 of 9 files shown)


mheon:
LGTM

pkg/specgen/generate/kube/seccomp.go
```diff
@@ -11,6 +11,7 @@ import (
 
 // KubeSeccompPaths holds information about a pod YAML's seccomp configuration
 // it holds both container and pod seccomp paths
+// nolint:golint
```
vrothberg, line 14: What was the linter complaining about here and at the others structs?

Luap99:
```
pkg/specgen/generate/kube/seccomp.go:14:6: type name will be used as kube.KubeSeccompPaths by other packages, and that stutters; consider calling this SeccompPaths (golint)
type KubeSeccompPaths struct {
```
However the variable `SeccompPaths` is already used in this package so I just added the nolint

vrothberg: Ah, thanks!

vrothberg (commented):
/hold cancel

PR #9343, opened 2021-02-12
change ps Created to unix

baude:
change from unixnano to unix for ps created --format json

[NO TESTS NEEDED]


rhatdan:
I believe this will bring us back to the original bug where two containers created back to back do not end up sorted correctly.

rhatdan:
Does this break? 
https://github.com/containers/podman/pull/8427

baude:
I dont think so ... we are only changing it for the presentation. i think it would have been sorted already.  So on the presentation side, it goes back ...

mheon:
Yeah, this ought to be fine, it's a JSON only change.
LGTM

Issue #8148, opened 2020-10-26
It's baaaack: podman images: Error: top layer [...] not found in layer tree


edsantiago:
Looks like a new variant of #7444, but am opening a new PR.

I seem to have gotten myself into a similar state again:
```console
$ ./bin/podman images
Error: top layer 1f832d5208105d5dde3f814d391ff7b4ddb557fd3bdbcb79418906242772dc73 of image a7a37f74ff864eec55891b64ad360d07020827486e30a92ea81d16459645b26a not found in layer tree

$ ./bin/podman images -a
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x15054fa]

goroutine 1 [running]:
github.com/containers/podman/v2/pkg/domain/infra/abi.(*ImageEngine).List(0xc0006c64c0, 0x1dc02a0, 0xc00001c1b0, 0xc000226d01, 0x2ad3878, 0x0, 0x0, 0xc000512000, 0xc0004afc88, 0xa1a754, ...)
        pkg/domain/infra/abi/images_list.go:52 +0x43a
github.com/containers/podman/v2/cmd/podman/images.images(0x2a1d900, 0xc0006cb0b0, 0x0, 0x1, 0x0, 0x0)
        cmd/podman/images/list.go:98 +0x10f
github.com/spf13/cobra.(*Command).execute(0x2a1d900, 0xc0000ba170, 0x1, 0x1, 0x2a1d900, 0xc0000ba170)
        vendor/github.com/spf13/cobra/command.go:850 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0x2a2c000, 0xc0000cc010, 0x184bce0, 0x2ad3878)
        vendor/github.com/spf13/cobra/command.go:958 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
        vendor/github.com/spf13/cobra/command.go:895
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        vendor/github.com/spf13/cobra/command.go:888
main.Execute()
        cmd/podman/root.go:88 +0xec
main.main()
        cmd/podman/main.go:78 +0x18c
```
Smoking gun seems to be: running BATS tests while also watching Tech Talk (heavy bandwidth hog). Test was pulling a huge image (`quay.io/libpod/fedora:31`), got interrupted by timeout, everything exploded after that. Nothing works any more.

[...]

PR #9346, opened 2021-02-12
[3.0] make layer-tree lookup errors non-fatal

vrothberg:
Internally, Podman constructs a tree of layers in containers/storage to
quickly compute relations among layers and hence images.  To compute the
tree, we intersect all local layers with all local images.  So far,
lookup errors have been fatal which has turned out to be a mistake since
it seems fairly easy to cause storage corruptions, for instance, when
killing builds.  In that case, a (partial) image may list a layer which
does not exist (anymore).  Since the errors were fatal, there was no
easy way to clean up and many commands were erroring out.

To improve usability, turn the fatal errors into warnings that guide the
user into resolving the issue.  In this case, a `podman system reset`
may be the approriate way for now.

[NO TESTS NEEDED] because I have no reliable way to force it.

[1] https://github.com/containers/podman/issues/8148#issuecomment-778253474

### DIFF
--- a/libpod/image/layer_tree.go
+++ b/libpod/image/layer_tree.go
@@ -5,6 +5,7 @@ import (
 
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

PR #9347, opened 2021-02-12
Fix superfluous response.WriteHeader call in WaitContainerLibpod()

edigaryev:
When the query decoding fails at the beginning of `WaitContainerLibpod()`, the `Error()` sets the header but doesn't returns after that.

This causes the execution flow to reach the `WriteResponse()` at the end of `WaitContainerLibpod()`, which attempts to set another header, thus causing the following error:

```
http: superfluous response.WriteHeader call from github.com/containers/podman/pkg/api/handlers/utils.WriteResponse (handler.go:124)
```


mheon:
LGTM

baude:
code lgtm

PR #9348, opened 2021-02-12
Refactor generated code

matejvasek:
Extracted common functionality to util function.

### DIFF
--- a/pkg/bindings/util/util.go
+++ b/pkg/bindings/util/util.go
@@ -1,8 +1,13 @@
 package util
 
 import (
+	"errors"
+	"net/url"
 	"reflect"
 	"strconv"
+	"strings"
+
+	jsoniter "github.com/json-iterator/go"
 )
 
 func IsSimpleType(f reflect.Value) bool {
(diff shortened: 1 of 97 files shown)


matejvasek:
Possible different approach is shown in: https://github.com/containers/podman/pull/9317

matejvasek:
Ah again this `nil != nil` golang s#it.

TomSweeneyRedHat:
@matejvasek if you can get this through tests, this will be the best PR of the year so far.  Very nice work.

matejvasek:
@TomSweeneyRedHat I think that the only reason it was failing was that `nil != nil` thing. Also you might checkout completely different approach to this: https://github.com/containers/podman/pull/9317 however it's much more complex / less practical, getting rid of reflection is not really worth it probably.

matejvasek:
@TomSweeneyRedHat are those flakes?

pkg/bindings/util/util.go
```diff
@@ -1,19 +1,24 @@
...
+	"net/url"
 	"reflect"
 	"strconv"
+	"strings"
+
+	jsoniter "github.com/json-iterator/go"
 )
 
-func IsSimpleType(f reflect.Value) bool {
+func isSimpleType(f reflect.Value) bool {
```
vrothberg, line 13: This is technically an API break.  While I doubt anybody is using it, I would sleep better keeping the two functions exported.

Retrospectively, it may should have been an `pkg/bindings/*internal*/util` where we can change whenever we want to.

vrothberg (commented):
```diff
+281 −5,213
```
Nice work!

vrothberg:
> @TomSweeneyRedHat are those flakes?

I restarted the failed jobs. It looked like flakes to me.

matejvasek:
Did just `sys podman fedora-32 rootless host` failed again.

matejvasek:
@vrothberg one of the two failing tests works not but the other one `sys podman fedora-32 rootless host` failed again :disappointed:

pkg/bindings/util/util.go
matejvasek: I introduced that just 10 ago so it's probably not used by anybody. I really should have added to some package with internal in its name.

vrothberg:
> @vrothberg one of the two failing tests works not but the other one `sys podman fedora-32 rootless host` failed again disappointed

I'll restarted another time :)  If it happens another time, I'll take a closer look.

pkg/bindings/util/util.go
vrothberg: Oh sorry, I should have checked if it's in v3.0!

In this case, could you move this code into `pkg/bindings/internal/util`?  This way, it's not part of the public API and we can change as much as we want without having to worry about potential external users.

vrothberg: Note that the code can be moved in another PR.  It's (almost) green, so we can merge this one.

matejvasek: It was merged 10 days ago to master, and then backported to v3. So it was released 3 days ago with v3.

Luap99:
@matejvasek Can you please squash your commits and add a proper commit message.

matejvasek:
Flake again.

vrothberg:
Restarted flakes

/hold
Holding the merge since I am still rather objecting to merge.  If we want to, please wait a while until v3.0 further stabilized to avoid additional work for backports.

vrothberg:
> Restarted flakes
> 
> /hold
> Holding the merge since I am still rather objecting to merge. If we want to, please wait a while until v3.0 further stabilized to avoid additional work for backports.

Apologies, I mixed tabs and meant to comment this in https://github.com/containers/podman/pull/9350

Issue #9365, opened 2021-02-14
kubic xUbuntu_18.04 broken default OCI runtime config? dockerfile RUN lines fail with default installation


buck2202:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Simple dockerfile builds fail on a default configuration install of podman 3 from the kubic xUbuntu_18.04 repository. The default oci runtime configuration seems broken

**Steps to reproduce the issue:**
I created a very dumb dockerfile which fails during its build
```
$ cat test_dockerfile 
FROM ubuntu
RUN set -x && apt-get -q update
$ podman build -t test -f test_dockerfile .
STEP 1: FROM ubuntu
STEP 2: RUN set -x && apt-get -q update
error running container: error creating container for [/bin/sh -c set -x && apt-get -q update]: : exec: "runc": executable file not found in $PATH
Error: error building at STEP "RUN set -x && apt-get -q update": error while running runtime: exit status 1
```

runc is not present, but the default runtime seems like it should be crun. The crun package is pulled in by podman:
```
$ which runc
$ which crun
/usr/bin/crun
$ podman info
*snip*
host:
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.17.6-58ef-dirty
      commit: fd582c529489c0738e7039cbc036781d1d039014
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
$ dpkg -S /usr/bin/crun
crun: /usr/bin/crun
$
$ apt show crun
Package: crun
Version: 100:0.17-4
Priority: optional
Section: devel
Maintainer: Lokesh Mandvekar <lsm5@fedoraproject.org>
Installed-Size: 1364 kB
Depends: libyajl2
Homepage: https://github.com/containers/crun.git
Download-Size: 231 kB
APT-Manual-Installed: no
APT-Sources: http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04  Packages
Description: OCI runtime written in C
$
$ apt show podman
Package: podman
Version: 100:3.0.0-1
Priority: optional
Section: devel
Maintainer: Lokesh Mandvekar <lsm5@fedoraproject.org>
Installed-Size: 85.2 MB
[...]

PR #9368, opened 2021-02-15
podman build: pass runtime to buildah

vrothberg:
Make sure that Podman's default OCI runtime is passed to Buildah in
`podman build`.  In theory, Podman and Buildah should use the same
defaults but the projects move at different speeds and it turns out
we caused a regression in v3.0.

Fixes: #9365


saschagrunert (approved):
LGTM

vrothberg:
CI failures should be fixed with #9370

vrothberg:
Need another lgtm

PR #9369, opened 2021-02-15
correct startup error message

vrothberg:
The error message when failing to create an image engine unconditionally
pointed to the Podman socket which is quite confusing when running
locally.

Move the error message to the point where the first ping to the service
fails.

[NO TESTS NEEDED]

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -158,7 +158,7 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 
 	// Prep the engines
 	if _, err := registry.NewImageEngine(cmd, args); err != nil {
-		return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure there is a Podman REST API service running.")
+		return err
 	}
 	if _, err := registry.NewContainerEngine(cmd, args); err != nil {
 		return err
--- a/pkg/bindings/connection.go
+++ b/pkg/bindings/connection.go
@@ -124,7 +124,7 @@ func NewConnectionWithIdentity(ctx context.Context, uri string, identity string)
 
 	ctx = context.WithValue(ctx, clientKey, &connection)
 	if err := pingNewConnection(ctx); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "cannot connect to the Podman socket, please verify that Podman REST API service is running")
 	}
 	return ctx, nil
 }


vrothberg:
Note that the error message is still quite broad as it assumes that _all_ errors would be caused by a remote-connection error but it's improving over the previous state.

saschagrunert (approved):
LGTM

rhatdan:
I was wondering where that error message was comping from.
/lgtm
/hold

vrothberg:
CI failures should be fixed with https://github.com/containers/podman/pull/9370

Luap99 (commented):
I think it would be better to move the error wrap to pkg/bindings here https://github.com/containers/podman/blob/30607d727895036d32aede44c1b4375849566433/pkg/bindings/connection.go#L126

Currently you can get something like this
```
$ bin/podman-remote --url abc  ps
Error: Cannot connect to the Podman socket, make sure there is a Podman REST API service running.: unable to create connection. "" is not a supported schema
```
If the url is invalid it is a client error and should not show something about the server.

vrothberg:
@Luap99 sounds good!  I'll update once CI is back green.

cmd/podman/root.go
```diff
@@ -158,9 +158,15 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 
 	// Prep the engines
 	if _, err := registry.NewImageEngine(cmd, args); err != nil {
-		return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure there is a Podman REST API service running.")
+		if registry.IsRemote() {
+			return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure there is a Podman REST API service running.")
```
TomSweeneyRedHat, line 162:
```suggestion
			return errors.Wrapf(err, "Cannot connect to the Podman socket, verify that there is a Podman REST API service running.")
```

cmd/podman/root.go
```diff
@@ -158,9 +158,15 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
...
+		}
+		return err
 	}
 	if _, err := registry.NewContainerEngine(cmd, args); err != nil {
+		if registry.IsRemote() {
+			return errors.Wrapf(err, "Cannot connect to the Podman socket, make sure there is a Podman REST API service running.")
```
TomSweeneyRedHat, line 168: If you take the above, same here.

mheon:
Very good catch! LGTM

pkg/bindings/connection.go
```diff
@@ -124,7 +124,7 @@ func NewConnectionWithIdentity(ctx context.Context, uri string, identity string)
 
 	ctx = context.WithValue(ctx, clientKey, &connection)
 	if err := pingNewConnection(ctx); err != nil {
-		return nil, err
+		return nil, logrus.Wrap(err, "cannot connect to the Podman socket, please verify that Podman REST API service is running")
```
Luap99, line 127:
```suggestion
		return nil, errors.Wrap(err, "cannot connect to the Podman socket, please verify that Podman REST API service is running")
```

PR #4601, opened 2019-11-29
Enable multi-platform rpm building

ssbarnea:
- make: fix python detection for multiple interpreters
- make: create generic `package` and `package-install` targets
- build_rpm.sh: move package installation into package-install
- build_rpm.sh: fix dnf/yum detection
- build_rpm.sh: install md2man rpm only on platfroms where is available
- build_rpm.sh: temporary skip packaging docs and debug on rhel-8
- docs: `make package-install`

This change is validated by [new CI jobs run by rdoproject](https://review.rdoproject.org/r/#/c/23943/). 

Fixes: #4627
Fixes: #4612

PR #9381, opened 2021-02-15
WIP: Cirrus: Make OSX and Static release archives

cevich:
CI During the v3.0 tag build did not produce release archives containing
documentation for all platforms.  Fix this by calling the release
related targets instead of the simple build targets.


cevich:
***Note to me:*** Confirm OSX and Static build tasks capture release *and* binary artifacts

cevich:
Oof..."homeless" and wrong arch errors :confounded:

cevich:
ugh...this is getting out of hand...our `Makefile` has some really serious problems:

```
make podman-remote-release-darwin.zip
...cut...
cmd/go: unsupported GOOS/GOARCH pair release-darwin.zip/amd64
```

:cry:

cevich:
[Cirrus Build tasks](https://cirrus-ci.com/build/6193534102405120) are now producing release-ready archives:
* `podman-release.tar.gz` (under gosrc) for F33, F32, U2010, and U2004 tasks.
* `podman-v3.1.0.msi` and `podman-remote-release-windows.zip` (under gosrc) for Windows
* `podman-remote-release-darwin.zip` (under gosrc) for OS-X
* `podman` and `podman-remote` (under binary/bin) for Static Linux

cevich:
@mheon PTAL - this fixes the missing OSX zip we noticed during the 3.0 release.  I'm a little nervous about the target rename, please let me know if you have a better idea.

mheon:
Why do we need the rename, for reference?

cevich:
> Why do we need the rename, for reference?

I tried to explain this in the commit message, but it's complex, so I probably failed.  I'll try again:

Wildcard targets are greedy (match as much as possible), and the order they appear is also significant.  Previously, the `podman-remote-darwin-release` target was not reachable.  References to it were matching `podman-remote-%` instead, with
the `%` scooping up `darwin-release` or `release-darwin.zip` (instead of just `darwin`).  Because the `%` is used for the `GOOS`  value, building the release archive on OSX was failing to compile.  Otherwise, calling `podman-remote-darwin` directly was working fine - which is probably why nobody noticed.

Since the `podman-remote-%` target is the most specific, it was simpler to rename it, than re-architect the `zip`, `tar.gz`, and `msi` targets (or massively re-arrange them).  I'm sure there are other solutions possible, but the PR diff (along with the bald-spot on my head) might be _much_ bigger.  This change was the quickest + simplest I could come up with, but the target rename might impact some people (I dunno, maybe not?).

@edsantiago speaks Makefile fairly well, maybe he has some idea for a simpler solution that gets us our tarballs, zips, msi's, and cross-compiled binaries.  Though, fair-warning: this is a rat's nest mapping exercise.

cevich:
Update: Had an idea this morning for (perhaps) a better way to tackle this problem.  I can start with a vastly simplified Makefile, where all targets simply touch a file.  Perhaps this will allow me to re-arrange the targets in such a way so that they all function as intended, without me wanting to gouge out my eyeballs.  Then use this as a template for updating the actual Makefile.

Ugh...still sounds like it will be quite a bit of work.  @mheon is there any pressure to solving this problem in the near-term (e.g. for an important release/branch)?

mheon:
Not really? And honestly, even if there were, I can build the OS X binaries myself without altering my own process. This obviously isn't ideal but it does work.

cevich:
> This obviously isn't ideal but it does work.

I feel the context here is important: An extremely long, detailed, manual podman release workflow.  Still your're right, there is a "plan b" available, so perhaps this isn't super urgent.  Thanks for the input.  I'll flip this back to `WIP`, dabble at it as I'm able, but focus on other items for now.

rhatdan:
@cevich what should we do with this PR, it is still in WIP state?

cevich:
Yes, still a WIP.  I'm procrastinating on it until I finish working on the new get_ci_vm since Makefile is so brain-hurty.  I'll resume work here next week most likely.

cevich:
Okay, I think this is ready now.  Though there are some ongoing google-cloud problems forcing me to babysit the tests :confounded:

Makefile
```diff
@@ -118,38 +118,27 @@ CROSS_BUILD_TARGETS := \
...
+all: binaries docs
 
 # Dereference variable $(1), return value if non-empty, otherwise raise an error.
 err_if_empty = $(if $(strip $($(1))),$(strip $($(1))),$(error Required variable $(1) value is undefined, whitespace, or empty))
 
+# Extract text following double-# for targets, as their description.
+_HLP_TGTS_RX = '^[[:print:]]+:.*?\#\# .*$$'
+_HLP_TGTS_CMD = grep -E $(_HLP_TGTS_RX) $(MAKEFILE_LIST)
+_HLP_TGTS_LEN = $(shell $(_HLP_TGTS_CMD) | cut -d : -f 1 | wc -L)
+_HLPFMT = "%-$(_HLP_TGTS_LEN)s %s\n"
```
vrothberg, line 189: Could we move these _inside_ the target?  That would prevent them from being executed outside.

vrothberg (commented):
Just one mini nit.

LGTM. Nice work, @cevich !

Makefile
```diff
@@ -118,38 +144,60 @@ CROSS_BUILD_TARGETS := \
 	bin/podman.cross.linux.mips64 \
 	bin/podman.cross.linux.mips64le
 
-.PHONY: all
-all: binaries docs
```
edsantiago, line 122: I strongly object to this change. No human being in the history of the world has wanted `make` to emit a help message. `make` should make. I've been typing `make` in this repo to build podman; I imagine that pretty much everyone who has ever built podman is also used to this. 

Please leave `all` as the default target. Anyone who wants help can type `make help`.

Makefile
edsantiago: It would seriously uglify the `help` target (because of the semicolon-backslash thing) but I agree: it's uncomfortable to have these run each time.

edsantiago (changes requested):
Mostly LGTM, and very nice work, thank you! I cannot approve the change to the default target though.

Makefile
cevich: I'm fine either way, the only reason I made this change was because the literal target "default" was not the actual default.  If this is the consensus, I'm happy to simply remove the target called "default".

hack/get_release_info.sh
```diff
@@ -14,11 +14,19 @@ valid_args() {
         cut -d '*' -f 1
 }
 
+# `git describe` does not reliably produce a useful version number.
+scrape_version() {
+    local versionfile='version/version.go'
+    local version_line=$(grep -m 1 'var Version =' $versionfile)
+    local version_string=$(cut -d '"' -f 2 <<<"$version_line")
+    echo "$version_string" | tr -d '[:space:]'
```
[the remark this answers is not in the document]
cevich: yep

cevich:
> it's uncomfortable to have these run each time.

I *think* they won't(?) because I made them `=` and not `:=`, they should only run upon reference.  I may be mistaken though.

cevich:
Confirmed, the `_HLP*` variables are only used on reference from the `help` target due to simple substitution at reference-time.  Example (`make target_one` does ***not*** produce a `foobarbaz` file)

```make
FOO = $(shell echo $$RANDOM)
BAR = one two $(FOO)
BAZ = $(shell echo "$(BAR)" | tee foobarbaz)

target_one:
        @echo "This is $@" | tee $@
        -cat foobarbaz

target_two:
        @echo "This is $(BAZ)" | tee $@
        -cat foobarbaz
```

I'll add a comment above the targets to make this situation more clear.

cevich:
Rebased and force-pushed with all'y'alls suggestions.  Hopefully the tests will remain green.

edsantiago (approved):
LGTM, thank you! One test flake (#9751), restarted.

Makefile
```diff
@@ -662,61 +794,46 @@ install.libseccomp.sudo:
...
-# installs them to /usr/local/bin/podman which is likely before. Always use
-# a full path to test installed podman or you risk to call another executable.
-.PHONY: package-install
-package-install: package  ## Install rpm packages
-	sudo ${PKG_MANAGER} -y install ${HOME}/rpmbuild/RPMS/*/*.rpm
-	/usr/bin/podman version
-	/usr/bin/podman info  # will catch a broken conmon
+.PHONY: clean
+clean: ## Clean all make artifacts
+	rm -rf \
```
jwhonce, line 821: Do you really want -f? Could ignoring rm errors cause any unexpected results?

Makefile
```diff
@@ -662,61 +794,46 @@ install.libseccomp.sudo:
...
+	# Remove related config files
+	rm -f ${DESTDIR}${ETCDIR}/cni/net.d/87-podman-bridge.conflist
+	rm -f ${DESTDIR}${TMPFILESDIR}/podman.conf
+	rm -f ${DESTDIR}${SYSTEMDDIR}/io.podman.socket
+	rm -f ${DESTDIR}${USERSYSTEMDDIR}/io.podman.socket
+	rm -f ${DESTDIR}${SYSTEMDDIR}/io.podman.service
+	rm -f ${DESTDIR}${SYSTEMDDIR}/podman.service
+	rm -f ${DESTDIR}${SYSTEMDDIR}/podman.socket
+	rm -f ${DESTDIR}${USERSYSTEMDDIR}/podman.socket
+	rm -f ${DESTDIR}${USERSYSTEMDDIR}/podman.service
```
jwhonce, line 817: Should -f be used? If uninstall is run shouldn't it report if some file cannot be removed?

Makefile
```diff
@@ -378,130 +584,88 @@ remoteapiv2:
...
 
-.PHONY: generate-bindings
-generate-bindings:
-ifneq ($(shell uname -s), Darwin)
-	GO111MODULE=off $(GO) generate ./pkg/bindings/... ;
-endif
+.PHONY: package
+package:  ## Build rpm packages
+	## TODO(ssbarnea): make version number predictable, it should not change
+	## on each execution, producing duplicates.
```
jwhonce, line 637: Should `scrape_version()` be exposed to take care of this TODO?

Makefile
edsantiago: `-f` is always a good idea, especially so if there's no `-r`. In this case, the goal is: "I want `make uninstall` to pass cleanly, even if I haven't run `make install`, or even if I ran `make install` with an older Makefile that didn't install some of the files I'm now removing. I do not want `rm` to fail if /foo/bar/podman.socket does not exist". `-f` will not hide errors:
```console
rootless$ /bin/rm -f /etc/passwd;echo $?
/bin/rm: cannot remove '/etc/passwd': Permission denied
1
```
The only thing `-f` does is "remove even if file is not writable, and shut up if file does not exist"

jwhonce: @edsantiago True `-f` does ignore missing files it also ignores all other errors. If the rm failed on an existing file, you have to intervene manually, which is why I called it out. Not that it's use is wrong, but can lead to unexpected results.

https://github.com/coreutils/coreutils/blob/master/src/rm.c#L334

Makefile
cevich: That wasn't a change I introduced...just looked, this is on master as well.  I can change it if there's good reason, but it wasn't really in-scope for my recipe adjustments (mainly targeted at the `podman-*-release` targets and getting non-linux binaries into their own subdir).

edsantiago: Please leave it. `rm -f` and `-rf` has been the Makefile convention for as long as I can remember. There is good reason for it.

Makefile
cevich: Oh, good catch.  I have no idea why specifically that TODO was added, but from the text it seems we could remove it.

cevich: This came in by https://github.com/containers/podman/pull/4601 

I looked through `contrib/build_rpm.sh` but don't see any references or direct use of `hack/get_release_info.sh` or scraping one of the `podman-*-release` target filenames.

@ssbarnea is this `TODO` in the Makefile `package:` target relevant to any of the release targets present in the Makefile or just something internal for the `package:` target?

Makefile
cevich: Ed's explanation (below) also gives me confidence we should leave this as-is.

Makefile
cevich: Yeah, I tend to agree with Ed.  I'm also not certain these have the same meaning for all contexts where the Makefile is used:

```
some_target:
    -rm -r /some /list /of /things

other_target:
    rm -rf /some /list /of /things
```

cevich: Note: I could see a case for doing `-rm -rf /some /list /of /things`, especially for "clean".  But I'm lazy and don't want to wait for all the tests to run again :grin:

Makefile
cevich: (Let's not hold up this PR for the question...it's not worth re-running all the tests just to remove a `TODO` - if necessary).

ssbarnea:
I hope no major changes come out of this from the behaviour point of view. What happens inside does bot concern me much.

I already use and parse that Makefile with https://pypi.org/project/mk/ as I consider it a good model of makefile that exposes/documented its targets.

cevich:
> I hope no major changes come out of this from the behaviour point of view.

Correct, this was my intention for the most part.  I don't know anything about `mk` but I can summarize what happened here in case that's helpful for you:

* Update the `help` target aesthetic appearance
* Fix the (literal) `default` target, to redirect to (actual) default `all` target
* Fix the various podman-*-release targets so they function for all OSes

HTH

ssbarnea:
@cevich Yeah, nothing to worry. I was able to easily fix it with https://github.com/pycontribs/mk/pull/65/files -- this being the only side effect. In fact it worked quite nice.

PR #9383, opened 2021-02-15
Fix an issue where copyup could fail with ENOENT

mheon:
This one is rather bizarre because it triggers only on some systems. I've included a CI test, for example, but I'm 99% sure we use images in CI that have volumes over empty directories, and the earlier patch to change copy-up implementation passed CI without complaint.

I can reproduce this on a stock F33 VM, but that's the only place I have been able to see it.

Regardless, the issue: under certain as-yet-unidentified environmental conditions, the copier.Get method will return an ENOENT attempting to stream a directory that is empty. Work around this by avoiding the copy altogether in this case.

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1617,6 +1617,17 @@ func (c *Container) mountNamedVolume(v *ContainerNamedVolume, mountpoint string)
 		if !srcStat.IsDir() {
 			return vol, nil
 		}
+		// Read contents, do not bother continuing if it's empty. Fixes
+		// a bizarre issue where something copier.Get will ENOENT on
+		// empty directories and sometimes it will not.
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests

edigaryev:
Care to add a simple test case from #9393?

vrothberg (commented):
LGTM

A rebase should turn CI green

mheon:
Rebased. I don't think we have any need of further tests - as I said, upstream CI doesn't catch this one, and we already have tests that use images with volumes.

ondrejbudai:
Will this be backported to RHEL 8.4? We're hitting the same issue in our CI.

rhatdan:
Yes

Issue #9387, opened 2021-02-15
WORKDIR changes the an existing folder's ownership to root:root


kkujala:
**Description**

I checked the bug reports I found one issue about folder ownership and WORKDIR: https://github.com/containers/buildah/issues/2323. That issue is about USER and WORKDIR, and in this case the folder is pre-created and chowned.

The podman and buildah work differently with WORKDIR than docker. The folder ownership (user:group) changes from my user (12345:12345) to root user (0:0) in case of buildah/podman. For docker the folder ownership does not change.

**Steps to reproduce the issue:**

1. Build the following Dockerfile either with buildah or podman, the result is the same. Also the option "--format docker" does not have an effect on the result.

```
FROM docker.io/centos:centos8

RUN : \
    && echo 'mygroup:x:12345:' >> /etc/group \
    && echo 'myuser:x:12345:12345:myuser:/home/myuser:/usr/sbin/nologin' \
        >> /etc/passwd \
    && mkdir --mode=0700 /home/myuser \
    && chown 12345:12345 /home/myuser \
    && :

USER 12345:12345
WORKDIR /home/myuser
```

The following build command can be used:
```
buildah bud --tag myuser_buildah .
```

2. Also build the same dockerfile with docker

```
docker build --tag myuser_docker .
```

**Describe the results you received:**

Results for buildah/podman:
```
$ podman run --rm -ti myuser_buildah ls -la /home
total 12
drwxr-xr-x  3 root root 4096 Feb 15 16:41 .
dr-xr-xr-x 17 root root 4096 Feb 15 16:41 ..
drwx------  2 root root 4096 Feb 15 16:41 myuser
```

Results for docker:
```
$ docker run --rm -ti myuser_docker ls -la /home
total 12
drwxr-xr-x 1 root   root    4096 Feb 15 15:52 .
drwxr-xr-x 1 root   root    4096 Feb 15 16:28 ..
drwx------ 2 myuser mygroup 4096 Feb 15 15:52 myuser
```

**Describe the results you expected:**
According to https://docs.docker.com/engine/reference/builder/#workdir
```
If the WORKDIR doesn’t exist, it will be created even if it’s not used in any subsequent Dockerfile instruction.
```
So in this case the folder the exists so it should not be re-created or modified.

For buildah/podman:
Expect that the /home/myuser be owned by myuser:mygroup, but because the root:root owns the /home/myuser, then the user myuser cannot access its own home folder.

For docker:
[...]

PR #9391, opened 2021-02-15
Don't chown workdir if it already exists

rhatdan:
Currently podman is always chowning the WORKDIR to root:root
This PR will return if the WORKDIR already exists.

Fixes: https://github.com/containers/podman/issues/9387

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -21,6 +21,7 @@ import (
 
 	cnitypes "github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ns"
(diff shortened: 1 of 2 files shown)


rhatdan:
@kkujala @vrothberg PTAL

mheon:
LGTM

kkujala:
Thank you for making this change. It looks good to me. Great that you added more checks in addition to the bug fixes. Maybe there could be one or more tests to cover those additional checks, if the existing test don't cover them already.

rhatdan:
@vrothberg PTAL

vrothberg (commented):
LGTM

Issue #9232, opened 2021-02-04
When using podman in gitlab ci, latest image is never pulled


rsommer:
/kind bug

**Description**

When using podman as docker replacement in gitlab ci, latest images are never pulled.

**Steps to reproduce the issue:**

1. Setup a new docker-runner with docker executor, using podman via podman.socket 

2. Execute a CI run on this runner using a simple image

3. Update the Image from another host, so that no local version is present

4. Execute another CI run -> the new image is not pulled

**Describe the results you received:**

The CI-runs always use the existing local image

**Describe the results you expected:**

Newer images usig a latest-tag are used.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.14
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.24, commit: '
  cpus: 4
  distribution:
    distribution: debian
    version: "10"
  eventLogger: journald
  hostname:citest01
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.19.0-14-amd64
  linkmode: dynamic
  memFree: 399601664
  memTotal: 4114087936
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.16.3-fd58-dirty
      commit: fd582c529489c0738e7039cbc036781d1d039014
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 2046816256
  swapTotal: 2046816256
  uptime: 1h 40m 10.27s (Approximately 0.04 days)
registries:
  search:
  - docker.io
  - quay.io
store:
  configFile: /etc/containers/storage.conf
  containerStore:
    number: 2
    paused: 0
    running: 0
    stopped: 2
  graphDriverName: overlay
[...]

PR #9397, opened 2021-02-16
images/create: always pull image

vrothberg:
The `images/create` endpoint should always attempt to pull a newer
image.  Previously, the local image was used which is not compatible
with Docker and caused issues in the Gitlab CI.

Fixes: #9232


vrothberg:

saschagrunert (approved):
LGTM

PR #9399, opened 2021-02-16
do not set empty $HOME

vrothberg:
Make sure to not set an empty $HOME for containers and let it default to
"/".

https://github.com/containers/crun/pull/599 is required to fully
address #9378.

Partially-Fixes: #9378


rhatdan:
LGTM

PR #9401, opened 2021-02-16
podman kill should report rawInput not container id

rhatdan:
Docker always reports back the users input, not the full
id, we should do the same.


rhatdan:
@vrothberg PTAL

vrothberg (commented):
LGTM

Issue #9408, opened 2021-02-17
podman-3.0.0 - runtime error using 'ps' `{{.Size}}` format


srcshelton:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman-3.0.0` outputs an error when attempting to use the `{{.Size}}` format to `podman ps`:

```
$ sudo podman ps -a --format '{{.Size}}'
Error: template: listContainers:1:13: executing "listContainers" at <.Size>: error calling Size: runtime error: invalid memory address or nil pointer dereference
```

**Output of `podman version`:**

```
Version:      3.0.0
API Version:  3.0.0
Go Version:   go1.15.7
Git Commit:   5b2585f5e91ca148f068cefa647c23f8b1ade622
Built:        Sun Feb 14 20:44:18 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.2
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: app-emulation/conmon-2.0.26
    path: /usr/bin/conmon
    version: 'conmon version 2.0.26, commit: 0e155c83aa739ef0a0540ec9f9d265f57f68038b'
  cpus: 8
  distribution:
    distribution: gentoo
    version: unknown
  eventLogger: file
  hostname: dellr330
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.10.16-gentoo
  linkmode: dynamic
  memFree: 1780682752
  memTotal: 33390886912
  ociRuntime:
    name: crun
    package: app-emulation/crun-0.17
    path: /usr/bin/crun
    version: |-
      crun version 0.17
      commit: 0e9229ae34caaebcb86f1fde18de3acaf18c6d9a
      spec: 1.0.0
      +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_AUDIT_WRITE,CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_MKNOD,CAP_NET_BIND_SERVICE,CAP_NET_RAW,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
[...]

PR #9409, opened 2021-02-17
podman ps --format '{{ .Size }}' requires --size option

rhatdan:
Podman -s crashes when the user specifies the '{{ .Size }}` format
on the podman ps command, without specifying the --size option.

This PR will stop the crash and print out a logrus.Error stating that
the caller should add the --size option.

Fixes: https://github.com/containers/podman/issues/9408


rhatdan:
@srcshelton @Luap99 PTAL

zhangguanzhang:
lgtm

PR #9339, opened 2021-02-12
[CI:DOCS] Update swagger doc for libpod container wait

matejvasek:
Signed-off-by: Matej Vasek <mvasek@redhat.com>

PR #9414, opened 2021-02-17
API: fix libpod's container wait endpoint condition conversion

### LINKED ISSUE #9351: API: container wait endpoint regression in 3.0.0
In https://github.com/containers/podman/pull/9048, which was included in the 3.0.0, the `condition` parameter was [changed to be a slice](), however, the assumptions behind the change explained in https://github.com/containers/podman/pull/9048#discussion_r570209609 does not seem to be valid. Two examples to demonstrate this:

Here's a Podman 3.0.0 with the fix from https://github.com/containers/podman/pull/9347 applied:

```
$ curl --unix-socket /tmp/podman.sock -d '' 'http://d/v1.0.0/libpod/containers/84138634fb08/wait?condition=running'
{"cause":"schema: error converting value for index 0 of \"condition\"","message":"failed to parse parameters for /v1.0.0/libpod/containers/84138634fb08/wait?condition=running: schema: error converting value for index 0 of \"condition\"","response":400}
```

However, in 2.2.1 everything works fine:

```
$ curl --unix-socket /tmp/podman.sock -d '' 'http://d/v1.0.0/libpod/containers/84138634fb08/wait?condition=running'                                                                          
-1
```

edigaryev:
Resolves #9351.

### DIFF
--- a/pkg/bindings/util/util.go
+++ b/pkg/bindings/util/util.go
@@ -12,8 +12,7 @@ import (
 )
 
 func IsSimpleType(f reflect.Value) bool {
-	switch f.Interface().(type) {
-	case fmt.Stringer:
+	if _, ok := f.Interface().(fmt.Stringer); ok {
 		return true
 	}
 
@@ -26,9 +25,8 @@ func IsSimpleType(f reflect.Value) bool {
 }
 
 func SimpleTypeToParam(f reflect.Value) string {
-	switch cast := f.Interface().(type) {
-	case fmt.Stringer:
-		return cast.String()
+	if s, ok := f.Interface().(fmt.Stringer); ok {
+		return s.String()
 	}
 
 	switch f.Kind() {
--- a/pkg/bindings/util/util.go
+++ b/pkg/bindings/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"errors"
+	"fmt"
 	"net/url"
 	"reflect"
 	"strconv"
@@ -11,14 +12,25 @@ import (
 )
 
 func IsSimpleType(f reflect.Value) bool {
+	switch f.Interface().(type) {
+	case fmt.Stringer:
+		return true
+	}
+
 	switch f.Kind() {
 	case reflect.Bool, reflect.Int, reflect.Int64, reflect.Uint, reflect.Uint64, reflect.String:
 		return true
 	}
+
 	return false
 }
 
 func SimpleTypeToParam(f reflect.Value) string {
+	switch cast := f.Interface().(type) {
+	case fmt.Stringer:
+		return cast.String()
+	}
+
 	switch f.Kind() {
 	case reflect.Bool:
 		return strconv.FormatBool(f.Bool())
@@ -31,6 +43,7 @@ func SimpleTypeToParam(f reflect.Value) string {
 	case reflect.String:
 		return f.String()
 	}
+
 	panic("the input parameter is not a simple type")
 }
 
--- a/test/apiv2/20-containers.at
+++ b/test/apiv2/20-containers.at
@@ -63,7 +63,7 @@ cid=$(jq -r '.Id' <<<"$output")
 # Prior to the fix in #6835, this would fail 500 "args must not be empty"
 t POST   libpod/containers/${cid}/start '' 204
 # Container should exit almost immediately. Wait for it, confirm successful run
-t POST   libpod/containers/${cid}/wait  '' 200 '0'
+t POST   libpod/containers/${cid}/wait?condition=stopped&condition=exited  '' 200 '0'
 t GET    libpod/containers/${cid}/json 200 \
   .Id=$cid \
   .State.Status~\\\(exited\\\|stopped\\\) \
--- a/test/apiv2/20-containers.at
+++ b/test/apiv2/20-containers.at
@@ -63,7 +63,7 @@ cid=$(jq -r '.Id' <<<"$output")
 # Prior to the fix in #6835, this would fail 500 "args must not be empty"
 t POST   libpod/containers/${cid}/start '' 204
 # Container should exit almost immediately. Wait for it, confirm successful run
-t POST   libpod/containers/${cid}/wait?condition=stopped&condition=exited  '' 200 '0'
+t POST   "libpod/containers/${cid}/wait?condition=stopped&condition=exited"  '' 200 '0'
 t GET    libpod/containers/${cid}/json 200 \
   .Id=$cid \
   .State.Status~\\\(exited\\\|stopped\\\) \
--- a/pkg/api/handlers/decoder.go
+++ b/pkg/api/handlers/decoder.go
@@ -6,6 +6,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/pkg/util"
 	"github.com/gorilla/schema"
 	"github.com/sirupsen/logrus"
@@ -19,6 +20,7 @@ func NewAPIDecoder() *schema.Decoder {
 	d.IgnoreUnknownKeys(true)
 	d.RegisterConverter(map[string][]string{}, convertURLValuesString)
 	d.RegisterConverter(time.Time{}, convertTimeString)
+	d.RegisterConverter(define.ContainerStatus(0), convertContainerStatusString)
 
 	var Signal syscall.Signal
 	d.RegisterConverter(Signal, convertSignal)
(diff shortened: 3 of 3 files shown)


edigaryev:
This approach might not work due to https://github.com/gorilla/schema/issues/157, but let's see if the tests pass.

mheon:
@jwhonce PTAL

matejvasek:
Doesn't this break API? I think that podman v3 is using numeric values in RESTful API.
https://github.com/containers/podman/pull/9339

matejvasek:
I think that change wasn't introduced https://github.com/containers/podman/pull/9048, but rather in https://github.com/containers/podman/pull/9051.

matejvasek:
@mheon @rhatdan @edigaryev ^^^

test/apiv2/20-containers.at
```diff
@@ -63,7 +63,7 @@ cid=$(jq -r '.Id' <<<"$output")
 # Prior to the fix in #6835, this would fail 500 "args must not be empty"
 t POST   libpod/containers/${cid}/start '' 204
 # Container should exit almost immediately. Wait for it, confirm successful run
-t POST   libpod/containers/${cid}/wait  '' 200 '0'
+t POST   libpod/containers/${cid}/wait?condition=stopped&condition=exited  '' 200 '0'
```
matejvasek, line 66: I think that libpod is sending decimal numbers (enum vals) now (podman v3), because `ToParams()` is called. If this should work we would need also to update code generator.

matejvasek:
Here https://github.com/containers/podman/pull/9051/files#diff-58b335c77eb3163bb2dec31c4e20ba5a62bafa143a1fbafe1a0e02dc1372fa31R327-R330

The https://github.com/containers/podman/issues/9351 is not related to the "slice thing".

test/apiv2/20-containers.at
matejvasek: The change to generator would be easy, but since podman v3 is out I don't know if that's good idea. To be clear I like descriptive string value much more than plain ints.

matejvasek:
@edigaryev please try whether this PR works with `podman-remote`.
I got:
```
podman-remote wait df5c8da64b80
Error: failed to parse parameters for /v2.0.0/libpod/containers/df5c8da64b803ec8587919d37dd55d2443fd4c1a81319e04b72f3ae08863d89e/wait?condition=4&interval=250ms: schema: error converting value for index 0 of "condition"
```
because `podman-remote` do `/v2.0.0/libpod/containers/df5c8da64b803ec8587919d37dd55d2443fd4c1a81319e04b72f3ae08863d89e/wait?condition=4&interval=250ms`

rhatdan:
If this is a bug we should fix it now.  No one is relying on this API yet, so I rather get it fixed.  We plan on releasing podman 3.0.1 very soon.

matejvasek:
> If this is a bug we should fix it now. No one is relying on this API yet, so I rather get it fixed. We plan on releasing podman 3.0.1 very soon.

@rhatdan  I think there was some (probably unnecessary) change to API. I asked about it https://github.com/containers/podman/pull/9048#issuecomment-771299272.

It was done as part of `Switch podman stop/kill/wait handlers to use abi` PR.
https://github.com/containers/podman/pull/9051/files#diff-58b335c77eb3163bb2dec31c4e20ba5a62bafa143a1fbafe1a0e02dc1372fa31R327-R330

matejvasek:
I am all in for using descriptive string instead of numbers, if it doesn't matter we change API between 3.0.0 and 3.0.1.

rhatdan:
I am fine with that.

test/apiv2/20-containers.at
matejvasek: Or maybe we shouldn't use generate `ToParams()` here at all, and revert it to manual uri parameter creation.

edigaryev:
> please try whether this PR works with `podman-remote`

Thanks for spotting this! Do I understand correctly that there very little `podman-remote` tests, so this doesn't get catched automatically?

matejvasek:
@edigaryev you may need to change:
pkg/bindings/containers/containers.go `func Wait(ctx context.Context, nameOrID string, options *WaitOptions) (int32, error)`

from:
```go
	params, err := options.ToParams()
```
back to:
```go
	params := url.Values{}
	if options.Changed("Condition") {
		params.Set("condition", options.GetCondition().String())
```

matejvasek:
@edigaryev unfortunately those test probably didn't run at all because of its dependency on some other task that failed first.

matejvasek:
we will see now.

matejvasek:
Or maybe we could improve `ToParams()` function, the "enum" in question `define.ContainerStatus` implements `type Stringer interface`, we could add special case for types implementing `Stringer` into https://github.com/containers/podman/blob/master/pkg/bindings/util/util.go

so `String()` would be called on it instead of `strconv.FormatInt()`

matejvasek:
`Test Bindings`

log contains:
```
[+0228s] • Failure [2.486 seconds]
[+0228s] Podman containers 
[+0228s] /var/tmp/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:18
[+0228s]   podman wait to pause|unpause condition [It]
[+0228s]   /var/tmp/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:273
[+0228s] 
[+0228s]   Expected
[+0228s]       <errorhandling.ErrorModel>: {
[+0228s]           Because: "schema: error converting value for index 0 of \"condition\"",
[+0228s]           Message: "failed to parse parameters for /v2.0.0/libpod/containers/140357e1bc9753fef84d650c2b4e72506e6220c93b3da20ae098ef559f9bd099/wait?condition=3: schema: error converting value for index 0 of \"condition\"",
[+0228s]           ResponseCode: 400,
[+0228s]       }
[+0228s]   to be nil
[+0228s] 
[+0228s]   /var/tmp/go/src/github.com/containers/podman/pkg/bindings/test/containers_test.go:282
```

test/apiv2/20-containers.at
edigaryev: Thanks for the suggestion, I've updated the generator here: https://github.com/containers/podman/pull/9414/commits/1e8f31bd16c831f13ba7c92447d99c8e0321fa9d.

pkg/bindings/util/util.go
```diff
@@ -11,14 +12,25 @@ import (
 )
 
 func IsSimpleType(f reflect.Value) bool {
+	switch f.Interface().(type) {
```
matejvasek, line 15: nit: probably could be:
```go
if _,  isStringer := f.Interface().(type); isStringer {
    return true
}
```

pkg/bindings/util/util.go
```diff
@@ -11,14 +12,25 @@ import (
...
 	switch f.Kind() {
 	case reflect.Bool, reflect.Int, reflect.Int64, reflect.Uint, reflect.Uint64, reflect.String:
 		return true
 	}
+
 	return false
 }
 
 func SimpleTypeToParam(f reflect.Value) string {
+	switch cast := f.Interface().(type) {
```
matejvasek, line 29: nit: probably could be:
```go
if s, ok := f.Interface().(Stringer); ok {
    return s.String()
}

File: test/apiv2/20-containers.at (hunk 1 above)
Comment (matejvasek) [inline test/apiv2/20-containers.at:66]: please put the URI into quotes
`t POST   "libpod/containers/${cid}/wait?condition=stopped&condition=exited"  '' 200 '0'`

Comment (matejvasek) [reply test/apiv2/20-containers.at:66]: > please put the URI into quotes
> `t POST "libpod/containers/${cid}/wait?condition=stopped&condition=exited" '' 200 '0'`

@edigaryev  It will fix the CI fail, the `&` causes troubles if not withing "".

Comment (edigaryev) [reply test/apiv2/20-containers.at]: Fixed in https://github.com/containers/podman/pull/9414/commits/3e168b19f23d3e9e4e9f25d0368b4d607968abb4.

File: pkg/bindings/util/util.go (hunk 3 above)
Comment (edigaryev) [reply pkg/bindings/util/util.go]: Changed in https://github.com/containers/podman/pull/9414/commits/e022c19753182cfb85f8f49354d493ee3a3147a3.

**Comment by matejvasek** [comment]
> This approach might not work due to [gorilla/schema#157](https://github.com/gorilla/schema/issues/157), but let's see if the tests pass.

I check it via debugger and the slice is de-serialized form the query just all right.

**Comment by matejvasek** [comment]
@edigaryev  OK, `Test Bindings` and `APIv2 tests` are passing. The rest should pass too I believe. Beware, there are flakes. So if some kind of test passes on `fedora` and it fails on `ubuntu` don't be scared and ask maintainers for restart.

I'm calling it a day

**Comment by matejvasek** [comment]
@mheon PTAL

**Comment by jwhonce** [comment]
LGTM

File: pkg/bindings/util/util.go [hunk 4]
```diff
@@ -11,14 +12,23 @@ import (
...
+
 	switch f.Kind() {
 	case reflect.Bool, reflect.Int, reflect.Int64, reflect.Uint, reflect.Uint64, reflect.String:
 		return true
 	}
+
 	return false
 }
 
 func SimpleTypeToParam(f reflect.Value) string {
```
matejvasek, line 27: Just occurred to me that the function could now look like:
```go
func SimpleTypeToParam(f reflect.Value) string {
	if IsSimpleType(f) {
		return fmt.Sprintf("%v", f.Interface())
	}
	panic("the input parameter is not a simple type")
}
```
But no need to change it now.

matejvasek:
@baude we made some changes to `ToParam()` function, I think you are original author. Please check it.

matejvasek (approved):

mheon:
/lgtm

Want this landed for 3.0.1

Issue #9249, opened 2021-02-05
We need to get logs testing to pass with journald driver.


rhatdan:
We attempted to switch default to journald, but tests were blowing up.

We need to fix these and to double the logging tests to run on both k8s_file and journald.

rhatdan:
@ashley-cui Once you get your secrets merged, I would like you to work on this, so we can change the default to journald.

ashley-cui:
closed via https://github.com/containers/podman/pull/9417

PR #9417, opened 2021-02-17
Fix journald logs

ashley-cui:
#9249

This PR contains a bunch of fixes to get logs testing on journald. I have a version of common that uses journald as default vendored in, I'll remove it if/when tests pass

This also needs to be updated later to have test coverage for k8s files once we switch over to default journald (otherwise k8s files wouldn't be tested at all)

### DIFF
--- a/libpod/container_log_linux.go
+++ b/libpod/container_log_linux.go
@@ -34,6 +34,8 @@ func (c *Container) readFromJournal(ctx context.Context, options *logs.LogOption
 	var config journal.JournalReaderConfig
 	if options.Tail < 0 {
 		config.NumFromTail = 0
+	} else if options.Tail == 0 {
+		config.NumFromTail = math.MaxUint64
 	} else {
 		config.NumFromTail = uint64(options.Tail)
 	}
--- a/libpod/container_log_linux.go
+++ b/libpod/container_log_linux.go
@@ -11,8 +11,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/podman/v2/libpod/define"
 	"github.com/containers/podman/v2/libpod/logs"
 	journal "github.com/coreos/go-systemd/v22/sdjournal"
+	"github.com/hpcloud/tail/watch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -65,8 +67,12 @@ func (c *Container) readFromJournal(ctx context.Context, options *logs.LogOption
 	if options.Tail == math.MaxInt64 {
 		r.Rewind()
 	}
+	state, err := c.State()
+	if err != nil {
+		return err
+	}
 
-	if options.Follow {
+	if options.Follow && state == define.ContainerStateRunning {
 		go func() {
 			done := make(chan bool)
 			until := make(chan time.Time)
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
Can we get the entrie logs_test to run on journal as well as k8s_file.  Is there a way to run through these tests twice, once for each type of log driver?

ashley-cui:
@rhatdan I'm not sure, I haven't played with ginkgo enough to know, but I can take a look and poke around with it. Worst comes to worst we just copy the file and add `--log-driver=k8s-file` in every test.

go.mod
```diff
@@ -10,8 +10,8 @@ require (
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.0
-	github.com/containers/buildah v1.19.4
-	github.com/containers/common v0.34.3-0.20210208115708-8668c76dd577
+	github.com/containers/buildah v1.19.0
+	github.com/containers/common v0.33.1-0.20210130125707-da196fb0f743
```
TomSweeneyRedHat, line 14: Do we really want to go from c/common 34 to 33 on master?

rhatdan: No I cut a new release today on common. we should move to that.

rhatdan:
@ashley-cui I am thinking of something like:
[logs_test.patch.txt](https://github.com/containers/podman/files/5999475/logs_test.patch.txt)

rhatdan:
Basically I wrap all of the logs_test.go that are not specific to journald and run each with all supported log-drivers in a loop.

go.mod
ashley-cui: @TomSweeneyRedHat  Oh this is just to force CI to use journald logging on the tests, the vendor commit will be dropped once the tests pass. The common here defaults to journald logging, which is what we want tests to pass on.

ashley-cui:
@rhatdan thanks! I'll take a look, that might be the solution :)

vrothberg (commented):
LGTM
Great work, @ashley-cui!

vrothberg:

libpod/container_log_linux.go
```diff
@@ -34,18 +35,24 @@ func (c *Container) readFromJournal(ctx context.Context, options *logs.LogOption
...
+	}
 	defaultTime := time.Time{}
 	if options.Since != defaultTime {
 		// coreos/go-systemd/sdjournal doesn't correctly handle requests for data in the future
 		// return nothing instead of falsely printing
 		if time.Now().Before(options.Since) {
 			return nil
 		}
-		config.Since = time.Since(options.Since)
+		config.Since = -time.Since(options.Since)
```
TomSweeneyRedHat, line 55: OK, I'll fess up to my ignorance.  What does the minus `-` sign do here?

ashley-cui: It just makes the value negative. The journald package we're using expects a negative `time.Duration` for times in the past and positive `time.Duration` for times in the future. Without the negative, a `--since 10m` ends up meaning "since 10 minutes in the future", which was why the `log --since` tests with journald weren't working

ashley-cui: I can edit the commit message or add a comment to say this

TomSweeneyRedHat: Thanks for the 411 @ashley-cui.  I'll merge now, but if you get a chance later, a small code comment there would be helpful.

TomSweeneyRedHat:
LGTM

test/e2e/logs_test.go
```diff
@@ -36,345 +36,341 @@ var _ = Describe("Podman logs", func() {
...
+			inspect.WaitWithDefaultTimeout()
+			if inspect.ExitCode() == 0 {
+				Expect(inspect.OutputToString()).To(Equal("exited"))
+				// TODO: add 2-second wait loop to confirm cleanup
+			} else {
+				Expect(inspect.ErrorToString()).To(ContainSubstring("no such container"))
+			}
+		})
+
+		It("podman logs with log-driver=none errors: "+log, func() {
```
rhatdan, line 262: This should not be within the loop, it should only be testing the "none" flag.

PR #9423, opened 2021-02-18
rootless cni without infra container

### LINKED ISSUE #8709: Proposal: imageless rootless-cni-infra

### LINKED ISSUE #9169: [RFE]Make docker-compose work with rootless podman
I've tried docker-compose with rootless podman, but that doesn't currently work:

```
siretart@x1:~/git/awesome-compose/gitea-postgres$ docker-compose -H unix:/run/user/1000/podman/podman.sock up
Pulling gitea (gitea/gitea:latest)...
76146dacacd305f5bbb4af41b6d107c06679db432fd1b5de4dd9577a56789dcf: pulling image () from docker.io/gitea/gitea:latest 
Pulling db (postgres:alpine)...
8c6053d81a45b455a79e7abbc5b4d8410735f060634d345f70b0530b1d50b2b6: pulling image () from docker.io/library/postgres:alpine 
Creating gitea-postgres_gitea_1 ... error
Creating gitea-postgres_db_1    ... 

Creating gitea-postgres_db_1    ... error

ERROR: for gitea-postgres_db_1  network connect is not enabled for rootless containers

ERROR: for gitea  network connect is not enabled for rootless containers

ERROR: for db  network connect is not enabled for rootless containers
ERROR: Encountered errors while bringing up the project.
```

What would it take to make network connect work for rootless mode? Is that on the roadmap?

_Originally posted by @siretart in https://github.com/containers/podman/discussions/9059#discussioncomment-325056_

Luap99:
Instead of creating an extra container create a network and mount
namespace inside the podman user namespace. This ns is used to
for rootless cni operations.
This helps to align the rootless and rootful network code path.
If we run as rootless we just have to set up a extra net ns and
initialize slirp4netns in it. The ocicni lib will be called in
that net ns.

This design allows allows easier maintenance, no extra container
with pause processes, support for rootless cni with --uidmap
and possibly more.

The biggest problem is backwards compatibility. I don't think
live migration can be possible. If the user reboots or restart
all cni containers everything should work as expected again.
The user is left with the rootless-cni-infa container and image
but this can safely be removed.

To make the existing cni configs work we need execute the cni plugins
in a extra mount namespace. This ensures that we can safely mount over
/run and /var which have to be writeable for the cni plugins without
removing access to these files by the main podman process. One caveat
is that we need to keep the netns files at `XDG_RUNTIME_DIR/netns`
accessible.

`XDG_RUNTIME_DIR/rootless-cni/{run,var}` will be mounted to `/{run,var}`.
To ensure that we keep the netns directory we bind mount this relative
to the new root location, e.g. XDG_RUNTIME_DIR/rootless-cni/run/user/1000/netns
before we mount the run directory. The run directory is mounted recursive,
this makes the netns directory at the same path accessible as before.

This also allows iptables-legacy to work because /run/xtables.lock is
now writeable.

Add rootless support for cni and --uidmap. This is supported with the new rootless cni logic.

Lastly enable network connect/disconnect for rootless users.

The docker-compose test has been changed to work rootless and also test for network connect.

Refer to the individual commits for more details.

Fixes #9169
Fixes #8709


Luap99:
This depends on https://github.com/cri-o/ocicni/pull/84

Luap99:
@edsantiago I need your input on the docker-compose test. I have two docker-compose test commits in here. The first one should enable `./test/compose/test-compose` to run rootless, however it is super messy. Theoretically it should just work when you change the socket path to a location writeable by the user but there are some serious issues with the test itself. All of the current test cases should be able to run rootless without modification. But it looks like running this as rootless is super flaky.  I added a `sleep 1` and some other changes to make it work. I would love if you could look into this I just changed things until it worked.

Only the second commit adds a test case which will fail as rootless with the current master. This PR will make it work.

Luap99:
@cevich @edsantiago I would like to have the rootless docker-compose functionality tested in CI. I already changed the docker-compose test script to make it work rootless but I do not know what I have to add to the .cirrus.yml file to test it in CI. It is already run as root.

edsantiago:
I'm no Cirrus expert, but I think @cevich has left for the day so here's my stab at it:
```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index e09db2a81..02ba5fc2c 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -440,7 +440,7 @@ apiv2_test_task:
         time_script: '$SCRIPT_BASE/logcollector.sh time'

 compose_test_task:
-    name: "compose test on $DISTRO_NV"
+    name: "compose test on $DISTRO_NV ($PRIV_NAME)"
     alias: compose_test
     only_if: *not_docs
     skip: *tags
@@ -450,6 +450,11 @@ compose_test_task:
     env:
         <<: *stdenvars
         TEST_FLAVOR: compose
+    matrix:
+        - env:
+              PRIV_NAME: root
+        - env:
+              PRIV_NAME: rootless
     clone_script: *noop  # Comes from cache
     gopath_cache: *ro_gopath_cache
     setup_script: *setup
```

edsantiago:
I'm seeing this (Build failed, gosrc something) in other PRs. I'm guessing Cirrus is down or having problems.

cevich:
@edsantiago has it, that should work fine (and looks like it does) for `.cirrus.yml`. However, the compose tests don't go through `dotest()` function in `contrib/cirrus/runner.sh`.  So `compose test on fedora-33 (rootless)` is in name only, we also need to teach the script how to actually run as the user...

cevich:
Okay, I think this should work (and support running other tests rootless in the future, if needed):

```diff
diff --git a/contrib/cirrus/runner.sh b/contrib/cirrus/runner.sh
index 507d22e13..f52e107cc 100755
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -288,18 +288,6 @@ dotest() {
         exec_container  # does not return
     fi;

-    # shellcheck disable=SC2154
-    if [[ "$PRIV_NAME" == "rootless" ]] && [[ "$UID" -eq 0 ]]; then
-        req_env_vars ROOTLESS_USER
-        msg "Re-executing runner through ssh as user '$ROOTLESS_USER'"
-        msg "************************************************************"
-        set -x
-        exec ssh $ROOTLESS_USER@localhost \
-                -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
-                -o CheckHostIP=no $GOSRC/$SCRIPT_BASE/runner.sh
-        # does not return
-    fi
-
     # containers/automation sets this to 0 for its dbg() function
     # but the e2e integration tests are also sensitive to it.
     unset DEBUG
@@ -340,6 +328,19 @@ msg "************************************************************"
 ((${SETUP_ENVIRONMENT:-0})) || \
     die "Expecting setup_environment.sh to have completed successfully"

+# shellcheck disable=SC2154
+if [[ "$PRIV_NAME" == "rootless" ]] && [[ "$UID" -eq 0 ]]; then
+    req_env_vars ROOTLESS_USER
+    msg "Re-executing runner through ssh as user '$ROOTLESS_USER'"
+    msg "************************************************************"
+    set -x
+    exec ssh $ROOTLESS_USER@localhost \
+            -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
+            -o CheckHostIP=no $GOSRC/$SCRIPT_BASE/runner.sh
+    # Does not return!
+fi
+# else: not running rootless, do nothing special
+
 cd "${GOSRC}/"

 handler="_run_${TEST_FLAVOR}"
```

contrib/cirrus/runner.sh
```diff
@@ -278,16 +279,8 @@ logformatter() {
...
-    local testsuite="$1"
-    req_env_vars testsuite CONTAINER TEST_ENVIRON PRIV_NAME
-
-    # shellcheck disable=SC2154
-    if ((CONTAINER==0)) && [[ "$TEST_ENVIRON" == "container" ]]; then
-        exec_container  # does not return
-    fi;
-
+# reexec the runner.sh scrip tas rootless user if necessary
+reexec_if_rootless() {
```
cevich, line 283: My proposed diff. will make rootless-mode available to all tests, but your way will work as well.  My thinking was: If two runner functions now need to run rootless, chances are good a third (and fourth and fifth and...) may be coming in the future.

contrib/cirrus/runner.sh
```diff
@@ -299,6 +292,19 @@ dotest() {
...
+dotest() {
+    local testsuite="$1"
+    req_env_vars testsuite CONTAINER TEST_ENVIRON PRIV_NAME
+
+    # shellcheck disable=SC2154
+    if ((CONTAINER==0)) && [[ "$TEST_ENVIRON" == "container" ]]; then
+        exec_container  # does not return
+    fi;
+
+    reexec_if_rootless
```
cevich, line 307: Issue in the original logic (not new): I don't ever see us running in a container first, then going rootless via ssh.  More than likely it would be the other way around - we become rootless first, then run tests in a container.

cevich (approved):
This LGTM assuming tests pass.  I would prefer my more general approach, but your way is fine too.  I don't really have a strong preference, only a minor inkling :smiley:

Luap99:
Well the new test failed. Most likely a flake.
I applied your diff @cevich.

baude:
LGTM as in looks GREAT to me!  well done

Issue #8874, opened 2021-01-04
Create Documentation for Networking Guide


fatherlinux:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Need a set of docs focused on creating and managing networks. Should cover 3-5 most common use cases when coming from Docker.

**Steps to reproduce the issue:**

1. Go to podman.io, click on Documentation, then click on search, and type "network" in the search box. 

**Describe the results you received:**

The only things which come up are  man pages. The man pages don't have all of the extra detail of a tutorial nor good examples. The man pages would augment this proposed documentation.

**Describe the results you expected:**
User should find a nice tutorial which covers common use cases when coming from Docker.

**Additional information you deem important (e.g. issue happens only occasionally):**
Like the manifest documentation RFE, I contemplated making this a RHEL documentation request, but I think it would be better upstream. Upstream first is the right approach for this.

baude:
@fatherlinux it would help us to cover the use cases.  i can only think of two. bridge and macvlan.

vrothberg:
I just saw on Twitter that CNI has fresh documentation which may be helpful: https://www.cni.dev/plugins/main/

rhatdan:
Baude this seems like something in your area.

baude:
that is a scary thought

PR #9425, opened 2021-02-18
[CI:DOCS]basic networking guide

baude:
created a basic networking guide for podman that outlines the common
network setups for podman.

Fixes: #8874


docs/tutorials/basic_networking.md
```diff
@@ -0,0 +1,285 @@
...
+```
+Now run the container.
+```
+$ podman run -dt --name webserver --net cni-podman1 -p 8081:80 quay.io/libpod/banner
+269fd0d6b2c8ed60f2ca41d7beceec2471d72fb9a33aa8ca45b81dc9a0abbb12
+```
+Note in the above run command, the container’s port 80 (where the Nginx server is
+running) was mapped to the host’s port 8080.  Port 8080 was chosen to demonstrate
+how the host and container ports can be mapped for external access.  The port could
+very well have been 80 as well.
```
Luap99, line 123: rootless users cannot bind ports < 1024

docs/tutorials/basic_networking.md
```diff
@@ -0,0 +1,285 @@
...
+there is no default network configuration provided.  You simply need to create a
+network, and the one will be created as a bridge network.
+
+```
+$ podman network create
+```
+
+When rootless containers are run with a CNI networking configuration, a “side-car”
+container for running CNI is also run. Do not remove this container while your rootless
+containers are running.  For rootfull containers, there is no “side-car” container
```
Luap99, line 89: Worth to add: if you remove this container (e.g by accident) all attached containers lose network connectivity. In order to restore the network connectivity all containers with networks must be restarted. This will automatically recreate the infra container.

docs/tutorials/basic_networking.md
```diff
@@ -0,0 +1,285 @@
...
+networks outside the laptop.  And outside clients can access the containers if the
+container is bound to a host port and the laptop firewall allows it.  Remember, unprivileged
+users must use ports 1024 through 65535 as lower ports require root privileges. (CAP_NET_BIND_SERVICE)
+Note: this can be adjusted using the `sysctl net.ipv4.ip_unprivileged_port_start`
+
+One of the drawbacks of slirp4netns is that the containers are completely isolated
+from each other.  Unlike the bridge approach, there is no virtual network.  For containers
+to communicate with each other, they can use the port mappings with the host system,
+or they can be put into a Pod where they share the same network namespace. See “Communicating
+between containers and pods'' for more information.
```
Luap99, line 216: link the header please

docs/tutorials/basic_networking.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Firewalls
+
+The role of a firewall will not impact the setup and configuration of networking,
+but it will impact traffic on those networks.  The most obvious is inbound network
+traffic to the container host, which is being passed onto containers usually with
+port mapping.  Depending on the firewall implementation, we have observed firewall
+ports being opened automatically due to running a container with a port mapping (for
+example).  If container traffic does not seem to work properly, check the firewall
+and allow traffic on ports the container is using.
```
Luap99, line 33: A common problem is that reloading the firewall deletes the cni iptables rules resulting in a loss of network connectivity for rootful containers. Podman v3 provides the podman network reload command to restore this without having to restart the container.
I think this can be added here.

Luap99 (commented):
This look very good just some notes which might be worth to add

TomSweeneyRedHat (commented):
The png files all appear to be empty, is that expected?  Are you planning to add a link in https://github.com/containers/podman/blob/master/docs/source/Tutorials.rst?

baude:
> The png files all appear to be empty, is that expected? Are you planning to add a link in https://github.com/containers/podman/blob/master/docs/source/Tutorials.rst?

i dont know why the files appear like that. maybe images are like that?

Luap99 (commented):
LGTM

TomSweeneyRedHat:
LGTM
IDK about the images being blank, you might be right @baude.  But you'd think GitHub would at least show a size and say something like "This is an image and can't be previewed", but ¯\_(ツ)_/¯

PR #9428, opened 2021-02-18
cp: treat "." and "/." correctly

vrothberg:
Make sure to treat "." and "/." correctly.  Both cases imply to copy the
contents of a directory in contrast to the directory.  This implies to
unset the KeepDirectoryNames options of the copiah package.

Previously, the code was performing a simple string suffix check which
is not enough since it would match files and directories ending with
".".


rhatdan:
LGTM

vrothberg:

Luap99 (commented):
/lgtm

PR #9440, opened 2021-02-19
vendor c/image v5.10.3

vrothberg:
* place shortnames in ~/.cache not ~/.config/.cache

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/containers/buildah v1.19.4
 	github.com/containers/common v0.34.3-0.20210208115708-8668c76dd577
 	github.com/containers/conmon v2.0.20+incompatible
-	github.com/containers/image/v5 v5.10.2
+	github.com/containers/image/v5 v5.10.3
 	github.com/containers/ocicrypt v1.1.0
 	github.com/containers/psgo v1.5.2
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM

vrothberg:
Argh, I think we need to wait. We either need c/storage 1.24.8 or 1.26.0. BUT we cannot use 1.24.8 without downgrading other dependencies and we cannot use 1.26.0 since that's bumping runc which currently doesn't work with crun as we need to wait for the libcap update to hit Fedora :^)

vrothberg:
@giuseppe do you know if the new libcap hit F33 and F32?

giuseppe:
F32 wasn't upgraded as far as I know. F33 yes, do we need to rebuild the images?

vrothberg:
> F32 wasn't upgraded as far as I know. F33 yes, do we need to rebuild the images?

The images usually do a `dnf update` before, so we should be good.

However, I don't see any other way than removing F32 from CI entirely.  Since we're not shipping Podman 3.0 there, we should be good.  @cevich @baude could we do this next week?

rhatdan:
We should move the images to f33 and f34.

rhatdan:
Dependabot is firing now.  We can close this one.

PR #9035, opened 2021-01-20
Make generate systemd --new robust against double curly braces

Luap99:
If the container create command contains an argument with double
curly braces the golang template parsing can fail since it tries
to interpret the value as variable. To fix this change the default
delimiter for the internal template to `{{{{`.

Fixes #9034

PR #9445, opened 2021-02-19
No header info for systemd generation

### LINKED ISSUE #8880: Add the --generate-timestamp option to podman generate systemd

/kind feature

**Description**

Add an option like `--generate-timestamp false` when generating a podman systemd unit file using the `podman generate systemd` command as seen in https://github.com/containers/podman/blob/master/pkg/systemd/generate/pods.go#L208

The default could still be set to true, but it will allow me to change this behaviour.

When running a command like `podman generate systemd --generate-timestamp false --name foobar` I expect the output to be somewhat like:

```
# container-foobar.service
# Mon Jan  4 23:32:51 CET 2021

[Unit]
Description=Podman container-foobar.service
Documentation=man:podman-generate-systemd(1)
Wants=network.target
After=network-online.target

[Service]
Restart=on-failure
ExecStart=/usr/bin/podman start foobar
ExecStop=/usr/bin/podman stop -t 10 foobar
ExecStopPost=/usr/bin/podman stop -t 10 foobar
PIDFile=/run/user/1000/containers/overlay-containers/720610392d571145fdd4f992cfc3be0e15ea283621b7f1a46f4d17fc6ce62f1f/userdata/conmon.pid
KillMode=none
Type=forking

[Install]
WantedBy=multi-user.target default.target
```

This feature would be usefull when using ansible's file copy to determine that the contents of the systemd unit file has changed. The timestamp makes it more difficult to do this and it will always result in a changed file. I can filter out line 2 or look for a timestamp on line 2 and filter it out, but that seems like a inconvenient solution.

```
podman version 2.2.1
```

```
not relevant
```

```
not relevant
```

Yes

Not relevant

jmguzik:
No header info for systemd generation: --no-header parameter
Fixes #8880
Closes #8880


pkg/systemd/generate/common.go
```diff
@@ -30,18 +30,23 @@ func validateRestartPolicy(restart string) error {
 	return errors.Errorf("%s is not a valid restart policy", restart)
 }
 
-const headerTemplate = `# {{{{.ServiceName}}}}.service
-# autogenerated by Podman {{{{.PodmanVersion}}}}
+const serviceTemplate = `# {{{{.ServiceName}}}}.service
```
vrothberg, line 33: I prefer to use a conditional, which is more template idiomatic:
```go
const headerTemplate = `# {{{{.ServiceName}}}}.service
{{{- if.GenerateHeader}}}
# autogenerated by Podman {{{{.PodmanVersion}}}}
{{{{- if .TimeStamp}}}}
# {{{{.TimeStamp}}}}
{{{{- end}}}}
{{{{- end}}}}

[Unit]
Description=Podman {{{{.ServiceName}}}}.service
Documentation=man:podman-generate-systemd(1)
Wants=network.target
After=network-online.target
`
```

vrothberg (commented):
Thanks for tackling it!

Just a couple of nits.  Could you also add two tests to `test/e2e/generate_systemd_test.go` that uses both `--no-header=true` and `--no-header=false`?

cmd/podman/generate/systemd.go
```diff
@@ -80,6 +81,9 @@ func systemd(cmd *cobra.Command, args []string) error {
 	if cmd.Flags().Changed("time") {
 		systemdOptions.StopTimeout = &systemdTimeout
 	}
+	if cmd.Flags().Changed("no-header") {
```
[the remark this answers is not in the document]
jmguzik: Wasn't sure if needed, thanks for your comment

pkg/systemd/generate/common.go
jmguzik: Hmm nice one, will see how it works

docs/source/markdown/podman-generate-systemd.1.md
```diff
@@ -32,6 +32,10 @@ Use the name of the container for the start, stop, and description in the unit f
 
 Using this flag will yield unit files that do not expect containers and pods to exist.  Instead, new containers and pods are created based on their configuration files.  The unit files are created best effort and may need to be further edited; please review the generated files carefully before using them in production.
 
+#### **--no-header**
+
+Do not add "autogenerated by" information and timestamp to the unit file.
```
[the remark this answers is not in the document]
jmguzik: > Just a couple of nits. Could you also add two tests to `test/e2e/generate_systemd_test.go` that uses both `--no-header=true` and `--no-header=false`?


@vrothberg --no-header currently similar to --new flag. It does not take =value. You just want to check what will happen or implement is as =value?

If first, what results do you expect? Error?

jmguzik:
> Just a couple of nits. Could you also add two tests to test/e2e/generate_systemd_test.go that uses both --no-header=true and --no-header=false?

@vrothberg --no-header currently similar to --new flag. It does not take =value. You just want to check what will happen or implement is as =value?

If first, what results do you expect? Error?

vrothberg:
> @vrothberg --no-header currently similar to --new flag. It does not take =value. You just want to check what will happen or implement is as =value?

Both take arguments such as `./bin/podman generate systemd --no-header=false --new=false $CTR`. Setting it false in the tests would just make sure that this value behaves as expected.  I don't think we test that yet for `--new=false` though.

pkg/systemd/generate/common.go
```diff
@@ -31,10 +31,12 @@ func validateRestartPolicy(restart string) error {
 }
 
 const headerTemplate = `# {{{{.ServiceName}}}}.service
```
rhatdan, line 33: What do you think about making this part of containers.conf and allowing users to edit it?  Then we would not need to add all of the plumbing for a new option and users could just modify the default template however they wanted?

@vrothberg Thoughts?

jmguzik: I would say it would be another feature - making this code useless.
It is architecture decision and you have to discuss this before I move further.
So @rhatdan , @vrothberg please let me know.

Luap99: I think supporting this kind of use case would be even more difficult. This would no longer allow us to do breaking changes to the template such as https://github.com/containers/podman/pull/9035.
I think adding a simple cli option is simple and intuitive for users.

jmguzik:
Generally, this feature is ready but remote test cases are failing. I suspect that something (probably one, two lines) is missing to enable the remote. I can't find it though - not super familiar with the codebase. Could I ask for a little help to spot the place @Luap99 @vrothberg ?

Luap99:
For remote you have to add the option here: https://github.com/containers/podman/blob/4a6582bd86572f0336c82577a0a4f315376c7179/pkg/bindings/generate/types.go#L12 and run `make generate-bindings` afterwards. Lastly, you have to add: WithNoHeaders here https://github.com/containers/podman/blob/4a6582bd86572f0336c82577a0a4f315376c7179/pkg/domain/infra/tunnel/generate.go#L12

pkg/systemd/generate/common.go
vrothberg: > What do you think about making this part of containers.conf and allowing users to edit it? Then we would not need to add all of the plumbing for a new option and users could just modify the default template however they wanted?

I think we always need a CLI counterpart for containers.conf options such that users can override defaults.  I have not strong feeling whether it's a good candidate for containers.conf.

But, I think it's something orthogonal to this change and we can add that a later point.

vrothberg: @jmguzik feel free to continue working on it.   It would not change the architecture or logic of this PR.   If we want a counterpart in the containers.conf, we can add it separately.

jmguzik: @vrothberg Ok, thanks for the clarification!

vrothberg:
In addition to what @Luap99 mentioned, there's a bit more to do. The Swagger documentation, the request routing and the handling of remote requests is all handled in `pkg/api/*`. So to make remote work, we also need to;

* Add the new `noHeader` remote parameter to the docs in `pkg/api/server/register_generate.go`.
* Supported the new parameter in the handler in `pkg/api/handlers/libpod/generate.go`.

It's mostly plumbing work to wire in the new parameter. The Podman backend is largely structured into two implementations of the same interfaces.  One for the native local Linux one (`pkg/domain/infra/abi`) and another one for the remote clients (`pkg/domain/infra/tunnel`).  The remote code is itself using the go-bindings we maintain directly in this repository in `pkg/bindings`.

@Luap99's comments cover the remote-client side. Mine above cover the server part.

pkg/systemd/generate/common.go
jmguzik: Works nicely, I admit didn't know about this (fairly new to go). However, I changed logic to GenerateNoHeader. Firstly, because the flag is called --no-header, and secondly because the default value will be false. Then, there is no need to add it to every structure and set explicitly true, as well as to tests. Hope it's ok for you.

jmguzik:
@vrothberg @Luap99 it would take me some time to find this. Many thanks for the insights. From my perspective feature is ready to be reviewed.

vrothberg:
> @vrothberg @Luap99 without help, it would take me some time to find this. Many thanks for the insights. From my perspective feature is ready to be reviewed.

LGTM.  Can you squash the four commits into one?

vrothberg:
Thanks for tackling it!  If you're interested in contributing more, feel free to have a look at the open issue or reach out to us directly. There is always work :)

jmguzik:
> LGTM. Can you squash the four commits into one?

Squashed :)

> Thanks for tackling it! If you're interested in contributing more, feel free to have a look at the open issue or reach out to us directly. There is always work :)

Sure :) My idea is to learn Go (my background is many years mainly C/C++). This is a good way. I will take a look in my free time, but I have still one PR open in buildah. I will try to finish this one first.

Luap99 (approved):
LGTM

PR #9456, opened 2021-02-21
Make binding util internal

matejvasek:
The functions are supposed to be called only from generated code


rhatdan:
LGTM

matejvasek:
rebased

vrothberg (commented):
/lgtm
/hold

@baude @mheon, I vote to backport to v3. Since we need to bump the go module to v3 in any case, we're in theory not breaking :^)

matejvasek:
Note that making those functions private is potentially braking for some users. Still the package is out one or two weeks so I doubt anybody is using them. We did more severe change between `3.0.0` and `3.0.1` https://github.com/containers/podman/pull/9414.

vrothberg:
> Note that making those functions private is potentially braking for some users. Still the package is out one or two weeks so I doubt anybody is using them. We did more severe change between `3.0.0` and `3.0.1` #9414.

Currently, Podman 3.0.X is still shipping a `go module v2`.  So we need to update the go module to v3 which would allow for getting this change in without really breaking users.

mheon:
Backport to v3 SGTM.

mheon:
Restarted two flakes.

PR #9458, opened 2021-02-21
Two usage variations of --new flag added to e2e

jmguzik:
During the work on https://github.com/containers/podman/pull/9445 it was discussed that --new flag do not have e2e tests for all the variations.
Additionally, one if in wrong place and similar, newly implemented flags do not work in all cases. I think it is good to add these tests for the completeness, but also future reference (to see all the possible options for this type of flag).


vrothberg (commented):
LGTM, thanks @jmguzik!

Luap99 (approved):
LGTM

jmguzik:
I think fail is not related to this change, however I do not have rights to re-run.

Issue #9252, opened 2021-02-07
podman stats not working, returns cgroup deleted


Kiritow:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman stats` not working. Using `root`, `podman stats` returns `Error: unable to load cgroup at /machine.slice/machine-libpod_pod_8cbc0cae539eb254c64e9fb5a613b151ed364c351083af55f5437fa4661e217f.slice/libpod-3b221f7e8b39f1e85111d88e7e8e864f7acbca8a6b880504cbd34793a72e8997.scope/init.scope: cgroup deleted`

**Steps to reproduce the issue:**

1. Reboot the machine, and login as root

2. Start all pods

3. `podman stats` fails.

**Describe the results you received:**

`podman stats` returns `Error: unable to load cgroup at /machine.slice/machine-libpod_pod_8cbc0cae539eb254c64e9fb5a613b151ed364c351083af55f5437fa4661e217f.slice/libpod-3b221f7e8b39f1e85111d88e7e8e864f7acbca8a6b880504cbd34793a72e8997.scope/init.scope: cgroup deleted`

**Describe the results you expected:**

`podman stats` should work fine.

**Additional information you deem important (e.g. issue happens only occasionally):**

`docker` is not installed on this machine.

`podman ps -a`:

```
CONTAINER ID  IMAGE                          COMMAND               CREATED       STATUS                   PORTS                     NAMES               POD ID        PODNAME
8a182186806b  localhost/frpc:latest          lsp-factorio-mcsm...  11 hours ago  Up 11 hours ago          0.0.0.0:34197->34197/udp  vigilant_bhaskara   c18bba00e974  lspfactorio
ae57b8694457  localhost/frpc:latest          lsp-factorio-dire...  11 hours ago  Up 11 hours ago          0.0.0.0:34197->34197/udp  optimistic_cerf     c18bba00e974  lspfactorio
3dabaaf7933e  localhost/lsp-factorio:latest                        11 hours ago  Up 11 hours ago          0.0.0.0:34197->34197/udp  stoic_kapitsa       c18bba00e974  lspfactorio
3db5b8127aa9  k8s.gcr.io/pause:3.2                                 11 hours ago  Up 11 hours ago          0.0.0.0:34197->34197/udp  c18bba00e974-infra  c18bba00e974  lspfactorio
[...]

PR #9464, opened 2021-02-22
cgroup: change cgroup deletion logic on v1

giuseppe:
do not raise an error if the cgroup exists at least on one
controller.

Previously it expected the cgroup to exists under all the
controllers.

[NO TESTS NEEDED]

Closes: https://github.com/containers/podman/issues/9252

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -331,18 +331,24 @@ func Load(path string) (*CgroupControl, error) {
 		control.additionalControllers = controllers
 	}
 	if !cgroup2 {
+		oneExists := false
+		// check that the cgroup exists at least under one controller
 		for name := range handlers {
 			p := control.getCgroupv1Path(name)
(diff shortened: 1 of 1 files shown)


pkg/cgroups/cgroups.go
```diff
@@ -331,17 +331,25 @@ func Load(path string) (*CgroupControl, error) {
...
-					return nil, ErrCgroupDeleted
+			if _, err := os.Stat(p); err == nil {
+				oneExists = true
+				break
+			}
+		}
+
+		// if there is no controller at all, raise an error
+		if !oneExists {
+			if os.IsNotExist(err) {
```
rhatdan, line 346: This err will always be nil, Correct? What err is this supposed to be checking?

giuseppe: my mistake, there should not be such check

mheon:
LGTM

PR #9467, opened 2021-02-22
[NO TESTS NEEDED] Turn on podman-remote build --isolation

rhatdan:
Currently podman only works with --isolation chroot.  This PR
fixes this by allowing the isolation mode to default to OCI and to
also allow users to pass the isolation mode into the containers.

The current tests for --isolation should cause this code to be tested.

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containers/buildah"
 	"github.com/containers/buildah/define"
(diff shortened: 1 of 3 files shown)


mheon:
I'm concerned about the Compat endpoint - it's the same codepath, and Docker does support that parameter but for an entirely different use (selecting backend implementation on Windows).

vrothberg (commented):
LGTM

rhatdan:
This one is fairly important, since remove builds will now be done locked down.

TomSweeneyRedHat:
LGTM
and happy green test buttons.

PR #9468, opened 2021-02-22
pr-should-include-tests: recognized "renamed" tests

edsantiago:
git tries to recognize renamed files. This isn't always
as helpful as intended. Turn it off, so we'll always see
files as 'A'dded.


vrothberg (commented):
LGTM, welcome back

Issue #9362, opened 2021-02-14
"podman cp" handles "/dev/stdout" as "-" (docker incompatibility)


eriksjolund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The command
`docker cp foobar:/root /dev/stdout` 
fails.
The same command with Podman does not fail, instead Podman writes a tar stream to _stdout_.
To match the Docker behaviour, Podman should fail too.

**Steps to reproduce the issue:**

#### Running with Podman

```
esjolund@laptop:~$ podman --version
podman version 3.0.0
esjolund@laptop:~$ podman create --name foobar docker.io/library/alpine:latest
6a4ea49f6bbc06572bbb117d6afeec54a1ab90af8645248ed0bfc462c06a0f72
esjolund@laptop:~$ rm -f /tmp/cmd.stdout
esjolund@laptop:~$ rm -f /tmp/cmd.stderr
esjolund@laptop:~$ podman cp foobar:/root /dev/stdout > /tmp/cmd.stdout 2> /tmp/cmd.stderr
esjolund@laptop:~$ echo $?
0
esjolund@laptop:~$ ls -l /tmp/cmd.std*
-rw-rw-r-- 1 esjolund esjolund    0 feb 14 18:55 /tmp/cmd.stderr
-rw-rw-r-- 1 esjolund esjolund 1536 feb 14 18:55 /tmp/cmd.stdout
esjolund@laptop:~$ cat /tmp/cmd.stdout | tar tvf -
drwx------ root/root         0 2021-01-28 22:51 root
esjolund@laptop:~$ rm -f /tmp/cmd.stdout
esjolund@laptop:~$ rm -f /tmp/cmd.stderr
esjolund@laptop:~$ podman cp foobar:/root /dev/stdout
root0000700000000000000000000000000014004630720011742 5ustar00rootroot00000000000000esjolund@laptop:~$ echo $?
0
esjolund@laptop:~$ 
```

#### Running with Docker

```
esjolund@laptop:~$ docker --version
Docker version 20.10.3, build 48d30b5
esjolund@laptop:~$ docker create --name foobar docker.io/library/alpine:latest
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
4c0d98bf9879: Pull complete 
Digest: sha256:08d6ca16c60fe7490c03d10dc339d9fd8ea67c6466dea8d558526b1330a85930
[...]

PR #9469, opened 2021-02-22
podman cp: /dev/std{in,out} fixes

vrothberg:
They shouldn't be treated specially to remain compatible with Docker.

Fixes: #9362

### DIFF
--- a/cmd/podman/containers/cp.go
+++ b/cmd/podman/containers/cp.go
@@ -121,7 +121,9 @@ func copyFromContainer(container string, containerPath string, hostPath string) 
 		return err
 	}
 
+	isStdout := false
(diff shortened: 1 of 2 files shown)


mheon:
I'm a little concerned by the potential for breaking existing users - it's unlikely, but this did work before. I suppose we can say it was never intentional and this is a bugfix?

mheon:
Code changes LGTM

vrothberg:
> I'm a little concerned by the potential for breaking existing users - it's unlikely, but this did work before. I suppose we can say it was never intentional and this is a bugfix?

Agreed.  It's not compatible with Docker's behavior, so it can be seen as a bug.

Issue #9479, opened 2021-02-22
podman rm -fa with continers that have --rm should not throw errors


baude:
When you have containers that were created with --rm and you perform a rm -fa, there shouldn't be errors thrown that the container is missing (because it is stopped before deletion).

/kind bug

Reproducer:

````
$ sudo podman run -dt --rm alpine top
$ sudo podman rm -fa

```

baude:
I'll grab this if nobody else does.

vrothberg:
I'll take a look :^)

vrothberg:
https://github.com/containers/podman/pull/9485

PR #9485, opened 2021-02-23
container removal: handle already removed containers

vrothberg:
Since commit d54478d8eaec, a container's lock is released before
attempting to stop it via the OCI runtime.  This opened the window
for various kinds of race conditions.  One of them led to #9479 where
the removal+cleanup sequences of a `run --rm` session overlapped with
`rm -af`.  Make both execution paths more robust by handling the case of
an already removed container.

Fixes: #9479

### DIFF
--- a/libpod/container_exec.go
+++ b/libpod/container_exec.go
@@ -954,18 +954,22 @@ func (c *Container) removeAllExecSessions() error {
 	}
 	// Delete all exec sessions
 	if err := c.runtime.state.RemoveContainerExecSessions(c); err != nil {
-		if lastErr != nil {
-			logrus.Errorf("Error stopping container %s exec sessions: %v", c.ID(), lastErr)
+		if errors.Cause(err) != define.ErrCtrRemoved {
+			if lastErr != nil {
+				logrus.Errorf("Error stopping container %s exec sessions: %v", c.ID(), lastErr)
+			}
+			lastErr = err
 		}
-		lastErr = err
 	}
 	c.state.ExecSessions = nil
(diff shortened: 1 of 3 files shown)


vrothberg:

libpod/container_exec.go
```diff
@@ -954,18 +954,22 @@ func (c *Container) removeAllExecSessions() error {
 	}
 	// Delete all exec sessions
 	if err := c.runtime.state.RemoveContainerExecSessions(c); err != nil {
-		if lastErr != nil {
-			logrus.Errorf("Error stopping container %s exec sessions: %v", c.ID(), lastErr)
+		if errors.Cause(err) != define.ErrCtrRemoved {
```
mheon, line 957: Should also check for `ErrNoSuchCtr`.

I wonder if we need a `IsCtrRemoved` function somewhere (`libpod/define` maybe?) to make this common check easy.

vrothberg: I don't think we can hit `ErrNoSuchCtr` since we're operating on this specific container already.

At least, I couldn't manage to hit this error.  Are you sure we can hit it there?

> I wonder if we need a IsCtrRemoved function somewhere (libpod/define maybe?) to make this common check easy.

Yes, I think it would be nice consolidate such state checks.

mheon: I think you may be right, since we're at most doing update calls on an existing container we probably can't trigger a `ErrNoSuchCtr`.

mheon:
LGTM

PR #9487, opened 2021-02-23
test: ignore named hierarchies for cgroups=split

giuseppe:
ignore named hierarchies for the --cgroups=split test as crun does not
set it.

Closes: https://github.com/containers/podman/pull/9302#issuecomment-784157272


giuseppe:
@chuanchang

It is not really an issue how we use it with systemd but I'll anyway address it in crun

rhatdan:
LGTM

vrothberg (commented):
/lgtm

PR #9491, opened 2021-02-23
system test image: build it multiarch

edsantiago:
buildah now supports running under emulation[1] as well as
creating manifests. Use those features to create a multiarch
testimage that can be used to test podman on other arches.

 [1] on Fedora 33, this requires the qemu-user-static package


edsantiago:
I think this is going to fail CI pretty badly. Lots of tests are failing on my laptop, most for reasons I can't even begin to understand. There's something weird about multiarch images. Also, the image is now almost three times its original size (15M, up from 6). This defeats my original purpose of using a tiny image. I can see why there's a `docker.io/ARCH/image` convention.

I'm going to try to focus on identifying the multiarch problems and seeing if I can fix them. ITM I would welcome advice from the team: should I (a) do that, or (b) forget about multiarch, and just build each arch separately and push it with a different tag, e.g. `quay.io/libpod/testimage:YYYYMMDD-{amd64,s390x,etc}` ?

edsantiago:
Weird. CI failed (as expected), but with only one failure (#9493). On my laptop I see many more failures, in networking and image and other tests. I'm starting to think my laptop may be haunted. Anyhow, this PR can't go anywhere until #9493 is resolved.

vrothberg (commented):
LGTM

TomSweeneyRedHat:
LGTM

PR #9495, opened 2021-02-23
Add '--group-add keep-groups': supplementary groups into container

rhatdan:
Currently we have rootless users who want to leak their groups access
into containers, but this group access is only able to be pushed in by
a hard to find OCI Runtime annotation.  This PR makes this option a lot
more visable and hides the complexity within the podman client.

This option is only really needed for local rootless users. It makes
no sense for remote clients, and probably makes little sense for
rootfull containers.

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -277,7 +277,7 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	createFlags.StringSliceVar(
 		&cf.GroupAdd,
 		groupAddFlagName, []string{},
-		"Add additional groups to join",
+		"Add additional groups to the primary container process. 'keep-groups' allows container processes to use suplementary groups.",
 	)
 	_ = cmd.RegisterFlagCompletionFunc(groupAddFlagName, completion.AutocompleteNone)
 
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -193,6 +193,25 @@ func createInit(c *cobra.Command) error {
 		val := c.Flag("entrypoint").Value.String()
 		cliVals.Entrypoint = &val
 	}
+
+	if c.Flags().Changed("group-add") {
+		groups := []string{}
+		for _, g := range cliVals.GroupAdd {
+			if g == "keep-groups" {
+				if len(cliVals.GroupAdd) > 1 {
+					return errors.New("the '--group-add keep-groups' option is not allowed with any other --group-add options")
+				}
+				if registry.IsRemote() {
+					return errors.New("the '--group-add keep-groups' option is not supported in remote mode")
+				}
+				cliVals.Annotation = append(cliVals.Annotation, "run.oci.keep_original_groups=1")
+			} else {
+				groups = append(groups, g)
+			}
+		}
+		cliVals.GroupAdd = groups
+	}
+
 	if c.Flags().Changed("pids-limit") {
 		val := c.Flag("pids-limit").Value.String()
 		pidsLimit, err := strconv.ParseInt(val, 10, 32)
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -261,8 +261,8 @@ Note: if _host_device_ is a symbolic link then it will be resolved first.
 The container will only store the major and minor numbers of the host device.
 
 Note: if the user only has access rights via a group, accessing the device
-from inside a rootless container will fail. The **crun**(1) runtime offers a
-workaround for this by adding the option **\-\-annotation run.oci.keep_original_groups=1**.
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the user's supplementary group access into the container.
 
 Podman may load kernel modules required for using the specified
 device. The devices that podman will load modules when necessary are:
@@ -361,9 +361,17 @@ GID map for the user namespace. Using this flag will run the container with user
 
 The following example maps uids 0-2000 in the container to the uids 30000-31999 on the host and gids 0-2000 in the container to the gids 30000-31999 on the host. `--gidmap=0:30000:2000`
 
-#### **\-\-group-add**=*group*
+#### **\-\-group-add**=*group|keep-groups*
 
-Add additional groups to run as
+Add additional groups to assign to primary user running within the container process.
+
+- `keep-groups` is a special flag that tells Podman to keep the supplementary group access.
+
+Allows container to use the user's supplementary group access. If file systems or
+devices are only accessible by the rootless user's group, this flag tells the OCI
+runtime to pass the group access into the container. Currently only available
+with the `crun` OCI runtime. Note: `keep-groups` is exclusive, you cannot add any other groups
+with this flag. (Not available for remote commands)
 
 #### **\-\-health-cmd**=*"command"* | *'["command", "arg1", ...]'*
 
@@ -634,7 +642,7 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **container:**_id_: reuse another container's network stack;
 - **host**: use the Podman host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
-- _network-id_: connect to a user-defined network, multiple networks should be comma separated;
+- _network-id_: connect to a user-defined network, multiple networks should be comma-separated;
 - **ns:**_path_: path to a network namespace to join;
 - **private**: create a new namespace for the container (default)
 - **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack. This is the default for rootless containers. It is possible to specify these additional options:
@@ -861,6 +869,8 @@ Security Options
 - `label=filetype:TYPE` : Set the label file type for the container files
 - `label=disable`       : Turn off label separation for the container
 
+Note: Labeling can be disabled for all containers by setting label=false in the **containers.conf** (`/etc/containers/containers.conf` or `$HOME/.config/containers/containers.conf`) file.
+
 - `mask=/path/1:/path/2` : The paths to mask separated by a colon. A masked path
   cannot be accessed inside the container.
 
@@ -869,12 +879,12 @@ Security Options
 - `seccomp=unconfined` : Turn off seccomp confinement for the container
 - `seccomp=profile.json` :  White listed syscalls seccomp Json file to be used as a seccomp filter
 
+- `proc-opts=OPTIONS` : Comma-separated list of options to use for the /proc mount. More details for the
+  possible mount options are specified in the **proc(5)** man page.
+
 - `unmask=ALL or /path/1:/path/2` : Paths to unmask separated by a colon. If set to **ALL**, it will
   unmask all the paths that are masked or made read only by default.
   The default masked paths are **/proc/acpi, /proc/kcore, /proc/keys, /proc/latency_stats, /proc/sched_debug, /proc/scsi, /proc/timer_list, /proc/timer_stats, /sys/firmware, and /sys/fs/selinux.**  The default paths that are read only are **/proc/asound, /proc/bus, /proc/fs, /proc/irq, /proc/sys, /proc/sysrq-trigger, /sys/fs/cgroup**.
-
-- `proc-opts=OPTIONS` : Comma separated list of options to use for the /proc mount. More details for the
-  possible mount options are specified at **proc(5)** man page.
 
 Note: Labeling can be disabled for all containers by setting label=false in the **containers.conf** (`/etc/containers/containers.conf` or `$HOME/.config/containers/containers.conf`) file.
 
@@ -1093,9 +1103,9 @@ Create a bind mount. If you specify, ` -v /HOST-DIR:/CONTAINER-DIR`, Podman
 bind mounts `/HOST-DIR` in the host to `/CONTAINER-DIR` in the Podman
 container. Similarly, `-v SOURCE-VOLUME:/CONTAINER-DIR` will mount the volume
 in the host to the container. If no such named volume exists, Podman will
-create one. The `OPTIONS` are a comma delimited list and can be: <sup>[[1]](#Footnote1)</sup>  (Note when using the remote client, the volumes will be mounted from the remote server, not necessarly the client machine.)
+create one. The `OPTIONS` are a comma-separated list and can be: <sup>[[1]](#Footnote1)</sup>  (Note when using the remote client, the volumes will be mounted from the remote server, not necessarly the client machine.)
 
-The _options_ is a comma delimited list and can be:
+The _options_ is a comma-separated list and can be:
 
 * **rw**|**ro**
 * **z**|**Z**
@@ -1185,7 +1195,7 @@ host into the container to allow speeding up builds.
 Content mounted into the container is labeled with the private label.
        On SELinux systems, labels in the source directory must be readable
 by the container label. Usually containers can read/execute `container_share_t`
-and can read/write `container_file_t`. If you can not change the labels on a
+and can read/write `container_file_t`. If you cannot change the labels on a
 source volume, SELinux container separation must be disabled for the container
 to work.
      - The source directory mounted into the container with an overlay mount
@@ -1245,10 +1255,14 @@ will convert /foo into a `shared` mount point. Alternatively one can directly
 change propagation properties of source mount. Say `/` is source mount for
 `/foo`, then use `mount --make-shared /` to convert `/` into a `shared` mount.
 
+Note: if the user only has access rights via a group, accessing the volume
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the user's supplementary group access into the container.
+
 #### **\-\-volumes-from**[=*CONTAINER*[:*OPTIONS*]]
 
 Mount volumes from the specified container(s). Used to share volumes between
-containers. The *options* is a comma delimited list with the following available elements:
+containers. The *options* is a comma-separated list with the following available elements:
 
 * **rw**|**ro**
 * **z**
@@ -1351,6 +1365,11 @@ $ podman create --name container1 -t -i fedora bash
 $ podman create --name container2 -t -i fedora bash
 $ podman create --name container3 --requires container1,container2 -t -i fedora bash
 $ podman start --attach container3
+
+### Configure keep supplemental groups for access to volume
+
+```
+$ podman create -v /var/lib/design:/var/lib/design --group-add keep-groups ubi8
 ```
 
 ### Rootless Containers
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -299,8 +299,8 @@ Note: if _host_device_ is a symbolic link then it will be resolved first.
 The container will only store the major and minor numbers of the host device.
 
 Note: if the user only has access rights via a group, accessing the device
-from inside a rootless container will fail. The **crun**(1) runtime offers a
-workaround for this by adding the option **\-\-annotation run.oci.keep_original_groups=1**.
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the user's supplementary group access into the container.
 
 Podman may load kernel modules required for using the specified
 device. The devices that Podman will load modules when necessary are:
@@ -405,9 +405,17 @@ Meaning **groupname** is initially mapped to gid **100000** which is referenced 
 above: The group **groupname** is mapped to group **100000** of the initial namespace then the
 **30000**st id of this namespace (which is gid 130000 in this namespace) is mapped to container namespace group id **0**. (groupname -> 100000 / 30000 -> 0)
 
-#### **\-\-group-add**=*group*
+#### **\-\-group-add**=*group|keep-groups*
 
-Add additional groups to run as
+Add additional groups to assign to primary user running within the container process.
+
+- `keep-groups` is a special flag that tells Podman to keep the supplementary group access.
+
+Allows container to use the user's supplementary group access. If file systems or
+devices are only accessible by the rootless user's group, this flag tells the OCI
+runtime to pass the group access into the container. Currently only available
+with the `crun` OCI runtime. Note: `keep-groups` is exclusive, you cannot add any other groups
+with this flag. (Not available for remote commands)
 
 #### **\-\-health-cmd**=*"command"* | *'["command", "arg1", ...]'*
 
@@ -670,7 +678,7 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **container:**_id_: reuse another container's network stack;
 - **host**: use the Podman host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
-- _network-id_: connect to a user-defined network, multiple networks should be comma separated;
+- _network-id_: connect to a user-defined network, multiple networks should be comma-separated;
 - **ns:**_path_: path to a network namespace to join;
 - **private**: create a new namespace for the container (default)
 - **slirp4netns[:OPTIONS,...]**: use **slirp4netns**(1) to create a user network stack. This is the default for rootless containers. It is possible to specify these additional options:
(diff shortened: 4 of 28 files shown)


rhatdan:
@mheon @vrothberg @baude @giuseppe WDYT?  This same issue has come up in multiple conversations, just today.  People are looking for this option and it is hidden in the crun man page noise.  I believe this is a very handy feature for rootless users and something we should make much more of a first class citizen.

@edsantiago I think we need to get clever about how to test this.  Basically need to setup a users group and a directory where the user launching podman only has access via group access.  Run one container without --groups have it fail, run same container with --groups option should succeed.

docs/source/markdown/podman-create.1.md
```diff
@@ -364,6 +364,12 @@ The following example maps uids 0-2000 in the container to the uids 30000-31999
 
 Add additional groups to run as
 
+#### **--groups**=*false*
+
+Allow container to use the user's supplementary group access. If file systems or
+devices are only accessible by the rootless users group, this flag tells the OCI
+runtim to pass the group access into the container. Currently only available with the crun OCI runtime. (Not available for remote commands)
```
vrothberg, line 371: Line break in this line is off.

rhatdan:
https://bugzilla.redhat.com/show_bug.cgi?id=1784889 Is an example of this.

rhatdan:
I just published a whole article on this.
https://www.redhat.com/sysadmin/supplemental-groups-podman-containers

vrothberg (commented):
_Love_ the idea.  I just wonder if `bool` is the right type since it's binary.  Can you imagine another setting where another value for `--groups` would be needed? @giuseppe WDYT?

docs/source/markdown/podman-create.1.md
edsantiago: type, runtim

edsantiago:
>I just wonder if bool is the right type

That's my initial reaction too: I can envision someone wanting to pass in `--groups=plugdev,cron` but not `wheel` (not that any files are ever owned by wheel, but, shrug).

Testing this is going to need some cleverness. It may require setup help from the framework, which means we'll have to coordinate between Cirrus, Fedora gating, and RHEL gating :(

vrothberg:
@edsantiago, that's a nice example.  Maybe we can have something ~symmetrical to `--user` where `--user=keep-id` has a special meaning and maps the current user into the container?  `--groups=keep-all`?

rhatdan:
We have to wait for @giuseppe to chime in, but I am not sure we can be that precise on the groups we want to leak.
I think crun is just not calling setgroups.

Looking at the man page this might be possible.
```
man setgroups
...
       int setgroups(size_t size, const gid_t *list);
```

cmd/podman/containers/create.go
```diff
@@ -189,6 +189,16 @@ func createInit(c *cobra.Command) error {
...
+	groups, err := c.Flags().GetBool("groups")
+	if err != nil {
+		return err
+
+	}
+	if groups {
+		cliVals.Annotation = append(cliVals.Annotation, "run.oci.keep_original_groups=1")
+
+	}
+
```
giuseppe, line 201: do we need to raise an error or at least a warning if the runtime is not crun?

giuseppe:
> We have to wait for @giuseppe to chime in, but I am not sure we can be that precise on the groups we want to leak.
> I think crun is just not calling setgroups.
> 
> Looking at the man page this might be possible.

yes, crun just skips calling `setgroups` so the original groups are maintained.

Should it perhaps be something under `--security-opt`? Like `--security-opt skip-setgroups=1`?

docs/source/markdown/podman-create.1.md
```diff
@@ -364,6 +364,13 @@ The following example maps uids 0-2000 in the container to the uids 30000-31999
 
 Add additional groups to run as
 
+#### **--groups**=*false*
```
mheon, line 367: Suggest rename to `--keep-groups` - that makes it much more clear what's going on.

docs/source/markdown/podman-create.1.md
```diff
@@ -364,6 +364,13 @@ The following example maps uids 0-2000 in the container to the uids 30000-31999
...
+
+Allow container to use the user's supplementary group access. If file systems or
```
mheon, line 369: "Allow the container to retain the supplementary groups of the user launching Podman"

cmd/podman/containers/create.go
rhatdan: I did not want to do that right away, since crun and others might get this support.  But I can add a Warning.

docs/source/markdown/podman-create.1.md
```diff
@@ -364,6 +364,13 @@ The following example maps uids 0-2000 in the container to the uids 30000-31999
...
+devices are only accessible by the rootless users group, this flag tells the OCI
```
mheon, line 370: "This ensures that the permissions of the container's root user will match that of the user launching Podman, ensuring that the container will be able to access files or devices the user can access via a supplemental group"

cmd/podman/containers/create.go
mheon: Concur, we need some sort of check that a compatible runtime is in use.

docs/source/markdown/podman-create.1.md
```diff
@@ -364,6 +364,13 @@ The following example maps uids 0-2000 in the container to the uids 30000-31999
...
+runtime to pass the group access into the container. Currently only available
+with the crun OCI runtime. (Not available for remote commands)
```
mheon, line 372: Does this work as root? Does it make any sense as root?

rhatdan:
@giuseppe Could we call the setgroups with specific groups to keep?

docs/source/markdown/podman-create.1.md
rhatdan: I hate options with embeded "-", why isn't --groups good enough, this is certainly what I would look for if I had this issue.

--groups=all or --groups=eng,staff would be my preference.

docs/source/markdown/podman-create.1.md
rhatdan: Maybe if we can specify the groups.  I don't know what it means to specify all groups.

```
# podman run --user=dan --groups=eng ...
```
Might make some sense.

docs/source/markdown/podman-create.1.md
mheon: I find `--groups` very confusing, because we already have `--group-add` which is quite similar but works on groups in the container. We need to distinguish between the two somehow.

rhatdan:
@giuseppe I am not crazy about  --security-opt? Like --security-opt skip-setgroups=1? since this is similar to --user and burying it under --security-opt just hides it.

I think we are seeing this is quite commonly needed.

edsantiago:
I've probably spent too much time on this already... but I can't get this to work at all, not with the new `--group=1` flag, nor even with manually specifying `--annotation`:
```console
$ ls -ld gt;ls -ldn gt
drwxrwx---. 1 root esm 30 Feb 23 14:12 gt/
drwxrwx---. 1 0 14904 30 Feb 23 14:12 gt/
$ ./bin/podman run --annotation run.oci.keep_original_groups=1 --rm --mount type=bind,source=$(pwd)/gt,destination=/gt registry.access.redhat.com/ubi8/ubi ls -l /gt
ls: cannot open directory '/gt': Permission denied
```
I've tried with `--userns=keep-id`, with `-v` instead of `--mount`, tried different images. Tried `:Z` which of course fails instantly. 

crun-0.17-1.fc33 which, as best I can tell, includes 6df930821d (the change to run.oci.keep_original_groups) (0.11)

I'm probably missing something very obvious. Any help appreciated; otherwise I'll take a fresh look in the morning.

rhatdan:
SELinux?

rhatdan:
Is ESM your primary group?  Try adding a sumplimental group "test", then add esm to 'test' and then execute
groups to make sure you have access to the directory outside of container.

Finally make sure you are running with crun.

edsantiago:
It's not SELinux. Or at least, not once I `chcon -t container_file_t /path/to/gt`. After that, if the `gt` directory is my primary group (`esm`), podman succeeds with or without the annotation. If I `chgrp` the directory to one of my secondary groups, I get EPERM but with no AVC (and it also fails with `setenforce 0`).

docs/source/markdown/podman-create.1.md
giuseppe: root in a user namespace would have the same issue and not able to map groups from the host (unless they are mapped in the user namespace)

giuseppe:
> @giuseppe Could we call the setgroups with specific groups to keep?

no we can't, we can either keep them all or just change to groups mapped in the user namespace.


> @giuseppe I am not crazy about --security-opt? Like --security-opt skip-setgroups=1? since this is similar to --user and burying it under --security-opt just hides it.

Personally I don't think it looks bad, we are already using security-opt for major features like seccomp and SELinux

mheon:
I don't mine `--security-opt` for this. It reduces potential confusion with the existing group flags.

rhatdan:
Ok I got out voted, switched man pages to reference `--security-opt keep-groups` Please review man pages to make sure everyone agrees.  Will work on changing the code, when I get some more time.

vrothberg:
To continue bike shedding: I would prefer `--keep-groups` over `--security-opt` as I find the lines of security a bit blurry in this context; SELinux or seccomp.  There is also ` --group-add` which we may take into account.  Having two, three, four different flags to alter group settings may be confusing to users _and_ smells risky to regress on.

Looking at `--user=keep-id` we could add a similar special value to `--group-add`?

rhatdan:
Lets discuss at standup.

rhatdan:
@edsantiago I might need some majic to get this to pass.  Basically I need the tests to run with root including a different group.

Can I run bats with sg wheel to put root in a non standard group.

@test "podman run podman in a different group" {
    sg wheel run_podman run alpine group
}

edsantiago:
Ugh. The problem is that `run_podman` is a bash function, and `sg` will have no idea how to run that. Let me cogitate.

rhatdan:
```
@test "podman --context emits reasonable output" {
    sg sshd "./bin/podman run --userns=auto --group-add keep-groups ${IMAGE} id"
    is "$output" ".*65534(nobody)"
}
```
This almost works.

edsantiago:
Well, you want a group that exists in both host and container, with the same GID. I'd say `sys` is a good candidate.

You could `sg sys "$PODMAN ..."` but please include a big "DO NOT DO THIS" comment.

giuseppe:
A hidden chroot(1) feature is the possibility to tweak UID/GID and additional groups before running a program (https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=commit;h=c45c51fe97193898f3909dcf5e4c0e117ab239a2):

```
# chroot --userspec=1000:1000 --groups 1,2,3 / id -G
1000 1 2 3
```
I think it makes easier to specify the numerical value so you don't have to worry if the group is defined on the host and in the container.

rhatdan:
I switched to using chroot. Thanks @giuseppe

test/system/170-run-userns.bats
```diff
@@ -0,0 +1,61 @@
...
+@test "podman --group-add without keep-groups {
+    run_podman run --group-add 457 $IMAGE id
+    if is_rootless; then
+        is "$output" ".*65534(nobody)" "Check group leaked into rootless mode"
+    else
+       is "$output" ".*457" "Check group leaked into container"
+    fi
+}
+
+@test "podman --group-add keep-groups plus added groups {
```
edsantiago, line 48: Missing end quote. I think this is going to fail pretty badly.

test/system/170-run-userns.bats
```diff
@@ -0,0 +1,61 @@
...
+
+function teardown() {
+    # A timeout or other error in 'build' can leave behind stale images
+    # that podman can't even see and which will cascade into subsequent
+    # test failures. Try a last-ditch force-rm in cleanup, ignoring errors.
+    run_podman '?' rm -a -f
+    run_podman '?' rmi -f build_test
+
+    basic_teardown
+}
```
edsantiago, line 61: As long as you're resubmitting, might as well remove this entire `teardown`; it's not necessary here

test/system/170-run-userns.bats
```diff
@@ -0,0 +1,51 @@
...
+    is "$output" ".*457" "Check group leaked into container"
+}
+
+@test "podman --group-add with keep-groups " {
+    if is_remote; then
+        run_podman run 125 --group-add keep-groups $IMAGE id
+        is "$output" ".*keep-groups not supported in remote mode" "Check keep-groups in remote mode"
+    else
+        run_podman run --group-add keep-groups $IMAGE id
+        is "$output" "uid=0(root) gid=0(root) groups=0(root)" "Check group leaked while root"
```
edsantiago, line 35: This is failing rootless (probably expected)

test/system/170-run-userns.bats
```diff
@@ -0,0 +1,51 @@
...
+    fi
+}
+
+@test "podman --group-add without keep-groups " {
+    run_podman run --group-add 457 $IMAGE id
+    if is_rootless; then
+        is "$output" ".*65534(nobody)" "Check group leaked into rootless mode"
```
edsantiago, line 42: This too is failing rootless:
```
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: Check group leaked into rootless mode
#| expected: '.*65534(nobody)'
#|   actual: 'uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video),457'
#\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rhatdan: Are you sure this is rootless?

rhatdan: oops never mind.

test/system/170-run-userns.bats
rhatdan: What is the actual?

edsantiago: on f33:
```
#|     FAIL: Check group leaked while root
#| expected: 'uid=0(root) gid=0(root) groups=0(root)'
#|   actual: 'uid=0(root) gid=0(root) groups=65534(nobody),65534(nobody),65534(nobody),0(root)'
```

test/system/170-run-userns.bats
```diff
@@ -0,0 +1,51 @@
...
+@test "podman --group-add without keep-groups while in a userns" {
+    skip_if_rootless "choot is not allowed in rootless mode"
+    skip_if_remote "--group-add keep-groups not supported in remote mode"
+    run chroot --groups 1234,5678 / ${PODMAN} run --uidmap 0:200000:5000 --group-add 457 $IMAGE id
+    is "$output" ".*457" "Check group leaked into container"
+}
+
+@test "podman --group-add with keep-groups " {
+    if is_remote; then
+        run_podman run 125 --group-add keep-groups $IMAGE id
```
edsantiago, line 31: Oops - should be `125 run`, not run 125

cmd/podman/common/create.go
```diff
@@ -277,7 +277,7 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	createFlags.StringSliceVar(
 		&cf.GroupAdd,
 		groupAddFlagName, []string{},
-		"Add additional groups to join",
+		"Add additional groups to the primary user process. 'keep-groups' allows container to use suplimentary groups.",
```
edsantiago, line 280: Should be "supplementary"

docs/source/markdown/podman-create.1.md
```diff
@@ -360,9 +360,17 @@ GID map for the user namespace. Using this flag will run the container with user
...
 
-Add additional groups to run as
+Add additional groups to assign to primary user running within the container process.
+
+- `keep-groups` is a special flag that tells Podman to keep the supplementary group access.
+
+Allows container to use the user's supplementary group access. If file systems or
+devices are only accessible by the rootless users group, this flag tells the OCI
+runtime to pass the group access into the container. Currently only available
+with the `crun` OCI runtime. Note: `keep-groups` is exclusive, you can not add any other groups
```
edsantiago, line 372: Language: "cannot", one word, not two

docs/source/markdown/podman-create.1.md
```diff
@@ -868,13 +878,13 @@ Security Options
 - `seccomp=unconfined` : Turn off seccomp confinement for the container
 - `seccomp=profile.json` :  White listed syscalls seccomp Json file to be used as a seccomp filter
 
+- `proc-opts=OPTIONS` : Comma separated list of options to use for the /proc mount. More details for the
```
edsantiago, line 881: Language: "Comma-separated", with a dash

docs/source/markdown/podman-create.1.md
```diff
@@ -1183,6 +1193,10 @@ will convert /foo into a `shared` mount point. Alternatively one can directly
 change propagation properties of source mount. Say `/` is source mount for
 `/foo`, then use `mount --make-shared /` to convert `/` into a `shared` mount.
 
+Note: if the user only has access rights via a group, accessing the volume
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the users supplementary group access into the container.
```
edsantiago, line 1198: Language: "user's", with an apostrophe

docs/source/markdown/podman-create.1.md
```diff
@@ -868,13 +878,13 @@ Security Options
 - `seccomp=unconfined` : Turn off seccomp confinement for the container
 - `seccomp=profile.json` :  White listed syscalls seccomp Json file to be used as a seccomp filter
 
+- `proc-opts=OPTIONS` : Comma separated list of options to use for the /proc mount. More details for the
+  possible mount options are specified at **proc(5)** man page.
```
edsantiago, line 882: Language: seems jarring. I know these two are moved from below, but if you have the chance to fix it, maybe it would read better as "...in \*\*proc(5)\*\*" or "...in the \*\*proc(5)** man page."

docs/source/markdown/podman-create.1.md
```diff
@@ -1278,6 +1292,11 @@ $ podman create --name container1 -t -i fedora bash
 $ podman create --name container2 -t -i fedora bash
 $ podman create --name container3 --requires container1,container2 -t -i fedora bash
 $ podman start --attach container3
+
+### Configure keep supplimental groups for access to volume
```
edsantiago, line 1296: Language: supplemental, no 'i'

docs/source/markdown/podman-run.1.md
```diff
@@ -298,8 +298,8 @@ Note: if _host_device_ is a symbolic link then it will be resolved first.
 The container will only store the major and minor numbers of the host device.
 
 Note: if the user only has access rights via a group, accessing the device
-from inside a rootless container will fail. The **crun**(1) runtime offers a
-workaround for this by adding the option **\-\-annotation run.oci.keep_original_groups=1**.
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the users supplementary group access into the container.
```
edsantiago, line 302: Language: "user's", with an apostrophe

docs/source/markdown/podman-run.1.md
```diff
@@ -404,9 +404,17 @@ Meaning **groupname** is initially mapped to gid **100000** which is referenced
...
-#### **\-\-group-add**=*group*
+#### **\-\-group-add**=*group|keep-groups*
 
-Add additional groups to run as
+Add additional groups to assign to primary user running within the container process.
+
+- `keep-groups` is a special flag that tells Podman to keep the supplementary group access.
+
+Allows container to use the user's supplementary group access. If file systems or
+devices are only accessible by the rootless users group, this flag tells the OCI
```
edsantiago, line 414: Again, missing apostrophe in "user's"

docs/source/markdown/podman-run.1.md
```diff
@@ -1264,6 +1280,10 @@ will convert /foo into a shared mount point. Alternatively, one can directly
 change propagation properties of source mount. Say, if _/_ is source mount for
 _/foo_, then use **mount --make-shared /** to convert _/_ into a shared mount.
 
+Note: if the user only has access rights via a group, accessing the volume
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the users supplementary group access into the container.
```
edsantiago, line 1285: missing apostrophe

docs/source/markdown/podman-run.1.md
```diff
@@ -1639,6 +1659,11 @@ Multiple containers can be required.
 $ podman create --name container1 -t -i fedora bash
 $ podman create --name container2 -t -i fedora bash
 $ podman run --name container3 --requires container1,container2 -t -i fedora bash
+
+### Configure keep supplimental groups for access to volume
```
edsantiago, line 1663: supplemental, no 'i'

edsantiago (commented):
LGTM, but just barely enough language issues to merit a re-push (IMO).

Please wait for further review feedback from @containers/podman-maintainers though, since I'm not qualified to review the primary code.

cmd/podman/common/create.go
rhatdan: Fixed

docs/source/markdown/podman-create.1.md
rhatdan: fixed

docs/source/markdown/podman-create.1.md
rhatdan: fixed, Plus a few more.

docs/source/markdown/podman-create.1.md
rhatdan: fixed

docs/source/markdown/podman-run.1.md
rhatdan: fixed

docs/source/markdown/podman-run.1.md
rhatdan: fixed

docs/source/markdown/podman-run.1.md
rhatdan: fixed

docs/source/markdown/podman-run.1.md
rhatdan: fixed

rhatdan:
@giuseppe @vrothberg PTAL

cmd/podman/common/create.go
```diff
@@ -277,7 +277,7 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	createFlags.StringSliceVar(
 		&cf.GroupAdd,
 		groupAddFlagName, []string{},
-		"Add additional groups to join",
+		"Add additional groups to the primary user process. 'keep-groups' allows container to use suplementary groups.",
```
mheon, line 280: Primary user process? Should this be "container process"?

docs/source/markdown/podman-create.1.md
```diff
@@ -261,8 +261,8 @@ Note: if _host_device_ is a symbolic link then it will be resolved first.
 The container will only store the major and minor numbers of the host device.
 
 Note: if the user only has access rights via a group, accessing the device
-from inside a rootless container will fail. The **crun**(1) runtime offers a
-workaround for this by adding the option **\-\-annotation run.oci.keep_original_groups=1**.
+from inside a rootless container will fail. Use the `--group-add keep-groups`
+flag to pass the users supplementary group access into the container.
```
mheon, line 265: "to pass the user's supplementary"

docs/source/markdown/podman-create.1.md
```diff
@@ -361,9 +361,17 @@ GID map for the user namespace. Using this flag will run the container with user
...
 The following example maps uids 0-2000 in the container to the uids 30000-31999 on the host and gids 0-2000 in the container to the gids 30000-31999 on the host. `--gidmap=0:30000:2000`
 
-#### **\-\-group-add**=*group*
+#### **\-\-group-add**=*group|keep-groups*
 
-Add additional groups to run as
+Add additional groups to assign to primary user running within the container process.
+
+- `keep-groups` is a special flag that tells Podman to keep the supplementary group access.
+
```
mheon, line 369: Should there be a newline here? This all seems like 1 paragraph.

cmd/podman/common/create.go
rhatdan: Fixed

docs/source/markdown/podman-create.1.md
rhatdan: fixed

docs/source/markdown/podman-create.1.md
rhatdan: fixed

rhatdan: Actually the line is needed so that the man page will be properly formatted

ashley-cui:
LGTM

andrewnicolalde:
> no we can't, we can either keep them all or just change to groups mapped in the user namespace.
@giuseppe 

I accept this is the route chosen, but even if only to document the rationale, could you explain it's not possible to only map specific groups into the container? Is there a technical reason?

rhatdan:
This is all the kernel allows,  We have an open request with the kernel to allow us to specify which Groups to leak, but it has not been accepted up stream.

@giuseppe Correct me if I am wrong.

giuseppe:
when running in a user namespace, `setgroups` is possible only with the groups that are part of the user namespace and the user namespace was created with a setuid helper (newuidmap/newgidmap).  It is not possible to reference groups that are not mapped into the user namespace.

If the call to `setgroups` is skipped, then the original groups are maintained.  So it is possible to either maintain the existing groups or use groups that are part of the user namespace (if it is created with newuidmap/newgidmap).

PR #9497, opened 2021-02-23
Network connect error if net mode is not bridge

### LINKED ISSUE #9496: network connect should error if netmode is not bridge

/kind bug

**Description**

Containers created with a network mode other than bridge should not work with network connect. (e.g. --network slirp4netns)

1. `sudo bin/podman run --name test --network slirp4netns -d alpine top`

2. `sudo bin/podman network connect cni-podman3 test`
 This command actually works and connects to the cni network but does not keep it after a restart. Also inspect is broken with `asked to join 2 CNI network(s) [cni-podman3 podman], but have information on 1 network(s)`

In my opinion only bridge should be allowed and all other net modes must be rejected by network connect/disconnect.

Luap99:
Only the the network mode bridge supports cni networks.
Other network modes cannot use network connect/disconnect
so we should throw a error.

Fixes #9496

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -1134,6 +1134,11 @@ func (w *logrusDebugWriter) Write(p []byte) (int, error) {
 
 // NetworkDisconnect removes a container from the network
 func (c *Container) NetworkDisconnect(nameOrID, netName string, force bool) error {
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

TomSweeneyRedHat:
LGTM

PR #9516, opened 2021-02-25
[NO TESTS NEEDED] Shrink the size of podman bindings

rhatdan:

### DIFF
--- a/libpod/runtime_img.go
+++ b/libpod/runtime_img.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 
+	buildahDefine "github.com/containers/buildah/define"
 	"github.com/containers/buildah/imagebuildah"
 	"github.com/containers/image/v5/directory"
 	"github.com/containers/image/v5/docker/reference"
@@ -165,7 +166,7 @@ func (r *Runtime) newImageBuildCompleteEvent(idOrName string) {
 }
 
 // Build adds the runtime to the imagebuildah call
-func (r *Runtime) Build(ctx context.Context, options imagebuildah.BuildOptions, dockerfiles ...string) (string, reference.Canonical, error) {
+func (r *Runtime) Build(ctx context.Context, options buildahDefine.BuildOptions, dockerfiles ...string) (string, reference.Canonical, error) {
 	if options.Runtime == "" {
 		// Make sure that build containers use the same runtime as Podman (see #9365).
 		conf := util.DefaultContainerConfig()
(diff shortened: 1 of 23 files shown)


rhatdan:
@cevich Still getting the same errors, I don't believe the VM was updated.

cevich:
FWIW [this error is all over the F33 logs, ](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6048676733779968/html/int-podman-fedora-33-root-container.log.html) in nearly/every test:

```
time="2021-03-13T11:19:02Z" level=warning msg="Error updating pod 2b8...3f4 conmon cgroup PID limit: open /sys/fs/cgroup/libpod_parent/2b8...3f4/conmon/pids.max: no such file or directory"```

**Comment by rhatdan** [comment]
@giuseppe PTAL, would this be causing all of the failures?

**Comment by giuseppe** [comment]
> `` time="2021-03-13T11:19:02Z" level=warning msg="Error updating pod 2b8...3f4 conmon cgroup PID limit: open /sys/fs/cgroup/libpod_parent/2b8...3f4/conmon/pids.max: no such file or directory"``` ``

I don't think so, it is just a warning when we cannot set the cgroup limits for conmon

**Comment by rhatdan** [comment]
@containers/podman-maintainers This one is ready to merge PTAL

File: libpod/runtime_img.go [hunk 1]
```diff
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 
+	buildah "github.com/containers/buildah/define"
```
TomSweeneyRedHat, line 12: Is `github.com/containers/podman/v3/libpod/define` going away eventually?  If not, rather than buildah here and just define elsewhere, I think this should be `bdefine` or something that's easily identified as being different than Podman's define.

vrothberg: I concur with Tom.  I would prefer to alias _all_ imports of this package as `buildahDefine`.  My eyes are used to read `define` as the Podman package.

rhatdan: I am waiting for my other pull-never  gets merged, before fixing this, in case of conflicts.

rhatdan:
Ok they are all buildahDefine now.

libpod/runtime_img.go
rhatdan: Done.

vrothberg (commented):
LGTM

PR #9525, opened 2021-02-25
[NO TESTS NEEDED] Prune and debloat

vrothberg:
* pkg/terminal: use c/storage/pkg/homedir
* turn hidden --trace into a NOP

While the homedir change is simple, the --trace changes may be more controversial.  Please speak up in case you want to keep it.  I _believe_ nobody is using it anymore and there are many parts that are not traced at all, so I think it's OK.  But I may very well miss something.

We get 2MB of binary size back.


rhatdan:
LGTM

TomSweeneyRedHat:
I'm not seeing any documentation for `--trace` anywhere in our man pages, and just wanted to double-check that you concur.
The changes LGTM and we had said we would remove this in v2.2.1.  However, at this point, I would not backport this to 3.0.  This should show up in Podman 3.1.
LGTM given the above.

vrothberg:

Issue #9500, opened 2021-02-24
No way to control /etc/hosts in podman play kube


gcs278:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

podman play kube does not have YAML or runtime options to set `--no-hosts` (like `podman pod create`). Therefore, it always the same /etc/hosts as the host, even if you have isolated (non-host) networking. Kubernetes does not shared the /etc/hosts file by default.

This is a problem for me since I need to isolate the /etc/hosts file and have no way to control it.

**Steps to reproduce the issue:**

1. podman play kube <YAML>

**Describe the results you received:**

POD Containers always has /etc/hosts from the host

**Describe the results you expected:**

Ability to control whether /etc/hosts is the same as the host's /etc/hosts

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2
Go Version:   go1.14.7
Built:        Mon Feb  8 16:19:06 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.21-3.module_el8.4.0+581+8bd31022.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 4cff974a37907e701aa2e31ad480ccc6f3f586f5'
  cpus: 10
  distribution:
    distribution: '"rhel"'
    version: "8.2"
  eventLogger: file
  hostname: rhel82
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.18.0-193.el8.x86_64
  linkmode: dynamic
  memFree: 1504854016
  memTotal: 3884625920
  ociRuntime:
    name: runc
    package: runc-1.0.0-65.rc10.module+el8.2.0+5762+aaee29fb.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.1-dev'
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  rootless: false
  slirp4netns:
    executable: ""
    package: ""
    version: ""
  swapFree: 3776471040
  swapTotal: 4190105600
  uptime: 2195h 35m 53.54s (Approximately 91.46 days)
  search:
  - registry.access.redhat.com
  - registry.redhat.io
  - docker.io
[...]

PR #9527, opened 2021-02-26
Enable no_hosts from containers.conf

rhatdan:
Since we have no good way to enable this on the server side, we will
just allow it to be set on the client side. This should solve almost all
cases.

Partially fixes: https://github.com/containers/podman/issues/9500


rhatdan:

TomSweeneyRedHat:
LGTM

vrothberg (commented):
/lgtm

PR #9537, opened 2021-02-28
Validate passed in timezone from tz option

TomSweeneyRedHat:
Erik Sjolund reported an issue where a badly formated file
could be passed into the `--tz` option and then the date in the container
would be badly messed up:

```
erik@laptop:~$ echo Hello > file.txt
erik@laptop:~$ podman run --tz=../../../home/erik/file.txt --rm -ti
docker.io/library/alpine cat /etc/localtime
Hello
erik@laptop:~$ podman --version
podman version 3.0.0-rc1
erik@laptop:~$
```
This fix checks to make sure the TZ passed in is a valid
value and then proceeds with the rest of the processing.

This was first reported as a potential security issue, but it
was thought not to be.   However, I thought closing the hole
sooner rather than later would be good.

### DIFF
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -1412,7 +1412,28 @@ USER mail`
 	})
 
 	It("podman run --tz", func() {
-		session := podmanTest.Podman([]string{"run", "--tz", "foo", "--rm", ALPINE, "date"})
+		testDir := filepath.Join(podmanTest.RunRoot, "tz-test")
+		err := os.MkdirAll(testDir, 0755)
+		Expect(err).To(BeNil())
+
+		tzFile := filepath.Join(testDir, "tzfile.txt")
+		file, err := os.Create(tzFile)
+		Expect(err).To(BeNil())
+
+		_, err = file.WriteString("Hello")
+		Expect(err).To(BeNil())
+		file.Close()
+
+		badTZFile := fmt.Sprintf("../../../%s", tzFile)
+		session := podmanTest.Podman([]string{"run", "--tz", badTZFile, "--rm", ALPINE, "date"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Not(Equal(0)))
+		Expect(session.ErrorToString()).To(ContainSubstring("error finding timezone for container"))
+
+		err = os.Remove(tzFile)
+		Expect(err).To(BeNil())
+
+		session = podmanTest.Podman([]string{"run", "--tz", "foo", "--rm", ALPINE, "date"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Not(Equal(0)))
 
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1503,16 +1503,24 @@ func (c *Container) makeBindMounts() error {
 	}
 
 	// Make /etc/localtime
-	if c.Timezone() != "" {
+	ctrTimezone := c.Timezone()
+	if ctrTimezone != "" {
+		// validate the format of the timezone specified if it's not "local"
+		if ctrTimezone != "local" {
+			_, err = time.LoadLocation(ctrTimezone)
+			if err != nil {
+				return errors.Wrapf(err, "error finding timezone for container %s", c.ID())
+			}
+		}
 		if _, ok := c.state.BindMounts["/etc/localtime"]; !ok {
 			var zonePath string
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
@eriksjolund FYI, thanks again for the report

test/e2e/run_test.go
```diff
@@ -1376,7 +1376,26 @@ USER mail`
...
+		file, err := os.Create(tzFile)
+		Expect(err).To(BeNil())
+		defer file.Close()
+
+		_, err = file.WriteString("Hello")
+		Expect(err).To(BeNil())
+
+		session := podmanTest.Podman([]string{"run", "--tz", ("../../../" + tzFile), "--rm", ALPINE, "date"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Not(Equal(0)))
```
Luap99, line 1393: Can you check for the correct error message here with `Expect(session.ErrorToString()).To(ContainSubstring("error finding timezone for container"))`. Otherwise the non zero exit code could come from the date command.

test/e2e/run_test.go
```diff
@@ -1376,7 +1376,27 @@ USER mail`
 	})
 
 	It("podman run --tz", func() {
-		session := podmanTest.Podman([]string{"run", "--tz", "foo", "--rm", ALPINE, "date"})
+		testDir := filepath.Join(podmanTest.RunRoot, "tz-test")
+		err := os.MkdirAll(testDir, 0755)
+		Expect(err).To(BeNil())
+
+		tzFile := filepath.Join(testdir, "tzfile.txt")
```
edsantiago, line 1383: typo, `s/dir/Dir/`

test/e2e/run_test.go
```diff
@@ -1376,7 +1376,27 @@ USER mail`
...
+		file, err := os.Create(tzFile)
+		Expect(err).To(BeNil())
+		defer file.Close()
+
+		_, err = file.WriteString("Hello")
+		Expect(err).To(BeNil())
```
edsantiago, line 1389: Don't you need a `file.Close()` here? Without it, you can't be guaranteed that the `Hello` will be flushed when `podman run` tries to read it.

test/e2e/run_test.go
```diff
@@ -1376,7 +1376,27 @@ USER mail`
...
+
+		session := podmanTest.Podman([]string{"run", "--tz", ("../../../" + tzFile), "--rm", ALPINE, "date"})
```
edsantiago, line 1391: Doesn't Go have some sort of path joiner? Would it make sense to use that instead of string concatenation?

edsantiago (commented):
minor nits

test/e2e/run_test.go
TomSweeneyRedHat: fixed and ty!

test/e2e/run_test.go
TomSweeneyRedHat: I've dropped the defer, thx.

test/e2e/run_test.go
TomSweeneyRedHat: 7 actually, this is the cheap and cheezy one.  I'll work up another approach

test/e2e/run_test.go
rhatdan: You have to write before close.

test/e2e/run_test.go
```diff
@@ -1412,7 +1412,29 @@ USER mail`
...
 	It("podman run --tz", func() {
-		session := podmanTest.Podman([]string{"run", "--tz", "foo", "--rm", ALPINE, "date"})
+		testDir := filepath.Join(podmanTest.RunRoot, "tz-test")
+		err := os.MkdirAll(testDir, 0755)
+		Expect(err).To(BeNil())
+
+		tzFile := filepath.Join(testDir, "tzfile.txt")
+		file, err := os.Create(tzFile)
+		Expect(err).To(BeNil())
+		file.Close()
```
rhatdan, line 1422: Delete this line.

test/e2e/run_test.go
TomSweeneyRedHat: Yep, copied it below, removed the defer above and should have killed it entirely.

TomSweeneyRedHat:
Oh my Goodness, after way too many tries, finally all green test buttons.
All comments from previous reviews have been addressed.

rhatdan:
LGTM

PR #9542, opened 2021-03-01
podman rmi: handle corrupted storage better

vrothberg:
The storage can easily be corrupted when a build or pull process (or any
process *writing* to the storage) has been killed.  The corruption
surfaces in Podman reporting that a given layer could not be found in
the layer tree.  Those errors must not be fatal but only logged, such
that the image removal may continue.  Otherwise, a user may be unable to
remove an image.

[NO TESTS NEEDED] as I do not yet have a reliable way to cause such a
storage corruption.

Reported-in: https://github.com/containers/podman/issues/8148#issuecomment-787598940

### DIFF
--- a/libpod/image/layer_tree.go
+++ b/libpod/image/layer_tree.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

PR #9552, opened 2021-03-01
[v3.0.1-rhel] Fix RHBZ#1932083

mheon:
This is a forward-port of the original changes in #8671 to the
new v3.0.1-rhel branch.


mheon:
Test failures are expected, I need to mark some tests as root-only since we no longer have an image to test with.

mheon:
Re-pushed, let's see if this goes green.

PR #9555, opened 2021-03-01
APIv2 tests: make more maintainable

edsantiago:
While I wasn't looking, some completely unreadable cruft
crept in here, and it's totally my fault: I never knew
you could pass JSON to a GET query. Everyone who DID
know that, did so, but had to URL-escape it into a
completely gobbledygook mess to make curl happy.

Solution: trivial, do the URL-escaping in 't' itself. I
just never realized that was needed.

I'm so sorry. I hope this helps.

### DIFF
--- a/test/apiv2/30-volumes.at
+++ b/test/apiv2/30-volumes.at
@@ -45,18 +45,17 @@ t GET libpod/volumes/json 200 \
     .[0].Name~.* \
     .[0].Mountpoint~.* \
     .[0].CreatedAt~[0-9]\\{4\\}-[0-9]\\{2\\}-[0-9]\\{2\\}.*
-# -G --data-urlencode 'filters={"name":["foo1"]}'
-t GET libpod/volumes/json?filters=%7B%22name%22%3A%5B%22foo1%22%5D%7D 200 length=1 .[0].Name=foo1
-# -G --data-urlencode 'filters={"name":["foo1","foo2"]}'
-t GET libpod/volumes/json?filters=%7B%22name%22%3A%20%5B%22foo1%22%2C%20%22foo2%22%5D%7D 200 length=2 .[0].Name=foo1 .[1].Name=foo2
-# -G --data-urlencode 'filters={"name":["nonexistent"]}'
-t GET libpod/volumes/json?filters=%7B%22name%22%3A%5B%22nonexistent%22%5D%7D 200 length=0
-# -G --data-urlencode 'filters={"label":["testlabel"]}'
-t GET libpod/volumes/json?filters=%7B%22label%22:%5B%22testlabel%22%5D%7D 200 length=2
-# -G --data-urlencode 'filters={"label":["testlabel=testonly"]}'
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM

test/apiv2/30-volumes.at
```diff
@@ -45,18 +45,17 @@ t GET libpod/volumes/json 200 \
...
-# -G --data-urlencode 'filters={"label":["testlabel"]}'
-t GET libpod/volumes/json?filters=%7B%22label%22:%5B%22testlabel%22%5D%7D 200 length=2
-# -G --data-urlencode 'filters={"label":["testlabel=testonly"]}'
-t GET libpod/volumes/json?filters=%7B%22label%22:%5B%22testlabel=testonly%22%5D%7D 200 length=1
-# -G --data-urlencode 'filters={"label":["testlabel1=testonly"]}'
-t GET libpod/volumes/json?filters=%7B%22label%22:%5B%22testlabel1=testonly%22%5D%7D 200 length=1
+t GET libpod/volumes/json?filters='{"name":["foo1"]}' 200 \
+  length=1 \
+  .[0].Name=foo1
+t GET libpod/volumes/json?filters='{"name":%20["foo1",%20"foo2"]}' 200 \
```
Luap99, line 51: Are the `%20` here intentional?

Luap99 (commented):
One question otherwise LGTM.

test/apiv2/30-volumes.at
edsantiago: You have a knack for finding the interesting questions!

Are you asking: "Why didn't you change those to spaces, since the quoting should make spaces OK?" If so, it's because I just felt really uncomfortable about using spaces in a query string - especially about possibly encouraging people to do so. Spaces in strings in shell scripts are a common problem, one that often wastes developer time. I chose to leave those as `%20` and trust that most people are by now used to interpreting that mentally as a space.

Are you asking: "Is there really a reason for including spaces in that filter?" If so, I have no idea. Those date to the original commit, 532bce4 by @acui. I simply left them in place. Ashley, did you intentionally include those spaces? Perhaps as a test of the server JSON parsing?

TBH I think the most readable solution would be to remove the spaces entirely, but if someone ever files a bug of the form "podman rejects spaces in input JSON" I would feel very stupid. So I think this is a reasonable tradeoff. I'd like to hear differing opinions though.

Luap99: > Why didn't you change those to spaces, since the quoting should make spaces OK?

I would prefer this but you know more about bash so I completely trust you when you think this is better. Maybe add comment that the `%20` in there are intentional.

edsantiago: Oops - too late - sorry, I was working on something else. I have a mild preference not to submit a full PR just for this comment change, can we agree to leave this be?

And FYI the issue with spaces is that, if one is careless, it's very likely that `"string with spaces"` (one string, one token) can at some point in a script be converted to `string with spaces` (three separate tokens) and those kinds of errors can be hard to track down.

Luap99: > can we agree to leave this be?

Sure no problem

PR #9558, opened 2021-03-01
entities: change the VolumeInfo CreatedAt type to time.Time

TristanCacqueray:

### DIFF
--- a/pkg/domain/entities/volumes.go
+++ b/pkg/domain/entities/volumes.go
@@ -2,6 +2,7 @@ package entities
 
 import (
 	"net/url"
+	"time"
 
 	"github.com/containers/podman/v3/libpod/define"
 	docker_api_types "github.com/docker/docker/api/types"
@@ -34,7 +35,7 @@ type VolumeConfigResponse struct {
 type VolumeInfo struct {
 
 	// Date/Time the volume was created.
-	CreatedAt string `json:"CreatedAt,omitempty"`
+	CreatedAt time.Time `json:"CreatedAt,omitempty"`
 
 	// Name of the volume driver used by the volume. Only supports local driver
 	// Required: true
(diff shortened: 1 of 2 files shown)


TristanCacqueray:
@mheon The intent is to make the CreatedAt attribute type similar to the other CreatedAt so that they can be easily compared. If I understand correctly, the json encoding converts this to a string and this shouldn't break the API (unless CreatedAt is not a date).

rhatdan:
Looks like you need a go.fmt.

You also need to add a test.

rhatdan:
/approve 
LGTM

rhatdan:

vrothberg (commented):
Thanks! LGTM but I want a final head nod from @mheon or @jwhonce/@baude.  Does it impact the compat API?

jwhonce (changes requested):
I need to dig deeper into all the code, but it appears the VolumeInfo structure was abandoned in favor of github.com/docker/docker/api/types/volume/Volume and libpod/define/InspectVolumeData.

If that is true, I would suggest we delete the orphaned structs and current the swagger documentation. This does not resolve the data type/format issue.

rhatdan:
@mheon WDYT?

mheon:
The abandoning sounds correct - I did that as part of the volume driver work (we have reimplemented inspecting volumes, completely parallel implementations, in at least five places, which I replaced with Libpod's version). I didn't remove this because I suspected it was still in use, but if we still compile cleanly after, the struct can be torn out without concern.

rhatdan:
@TristanCacqueray @mheon @What is the state of this PR?

mheon:
Needs to be replaced with a new PR that removes this struct, at a minimum (it's currently unused) and possibly adjusts the types used by volume inspect.

PR #9575, opened 2021-03-02
Rewrite Rename backend in a more atomic fashion

mheon:
Move the core of renaming logic into the DB. This guarantees a lot more atomicity than we have right now (our current solution, removing the container from the DB and re-creating it, is *VERY* not atomic and prone to leaving a corrupted state behind if things go wrong. Moving things into the DB allows us to remove most, but not all, of this - there's still a potential scenario where the c/storage rename fails but the Podman rename succeeds, and we end up with a mismatched state.


mheon:
I need to think of a few tests for this after lunch - we unblocked renaming containers with active exec sessions, for example.

mheon:
Test added

mheon:
Looks like this is going green. @containers/podman-maintainers PTAL

rhatdan:
LGTM

Luap99 (commented):
LGTM

One question: Right now we do not update update the dnsname plugin dns name. Do you think we should call network reload as last step? (Not for this PR)

mheon:
Tide is blocked on something, I'll check for what.

@Luap99 A full reload is a little expensive, it would be better if we could just update dnsname - we need some reworking of it anyways to fix the flaw where it can only handle a single network the container is connected to, we can handle this at the same time when we get around to that.

PR #9580, opened 2021-03-02
Fix support for podman build --timestamp

rhatdan:
Currently podman is ignoreing the build --timestamp flag.
This PR fixes this for local and remote clients.

Fixes: https://github.com/containers/podman/issues/9569


rhatdan:
@edsantiago PTAL

mheon:
LGTM

edsantiago:
LGTM, and I've confirmed that this also has the (desired) effect of reproducible image IDs. Thanks for the quick work.

Issue #9316, opened 2021-02-10
[APIv2] secrets are missing Version field


riyad:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Trying to find regressions by exercising the APIv2 trough [docker-py](https://github.com/docker/docker-py)'s test suite (see #5386) I came across a failing assertion [`SecretAPITest::test_inspect_secret`](https://github.com/docker/docker-py/blob/master/tests/integration/api_secret_test.py#L41) test.

It seems secrets returned from `GET /secrets` and `GET /secret/{id}` endpoints are missing the (non-optional) `Version` field.

**Steps to reproduce the issue:**

1. create a secret
```shell
$ curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/secrets/create' -H "Content-Type: application/json" -d '{"Data": "Zm9vIGJhcg==", "Name": "dockerpy_test1"}'
{"ID":"f76cb9091ea531994cd9e9438"}
```

2. inspect the secret
```shell
$ curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XGET 'http://localhost/v1.40/secrets/f76cb9091ea531994cd9e9438' 
{"ID":"f76cb9091ea531994cd9e9438","CreatedAt":"2021-02-10T23:57:25.547915167+01:00","UpdatedAt":"2021-02-10T23:57:25.547915167+01:00","Spec":{"Name":"dockerpy_test1","Driver":{"Name":"file","Options":null}}}
```

**Describe the results you received:**

The secrets returned don't have a `Version` field.

**Describe the results you expected:**

The secrets returned should have a `Version` field.

see Docker API docs: https://docs.docker.com/engine/api/v1.40/#operation/ConfigInspect

mheon:
@ashley-cui PTAL

ashley-cui:
hmm, didn't we decide that it doesn't make sense to have a version field since docker only responds with a version field since it has a swarm manager that deals with secrets? but i guess since its compat, should i wire in the version number anyway?

vrothberg:
> hmm, didn't we decide that it doesn't make sense to have a version field since docker only responds with a version field since it has a swarm manager that deals with secrets? but i guess since its compat, should i wire in the version number anyway?

Yes, that what we thought initially.  I am also not sure which version to return.  @jwhonce @baude WDYT?

rhatdan:
[...]

PR #9583, opened 2021-03-02
Add version field to secret compat list/inspect api

ashley-cui:
Docker api expects secrets endpoint to have a version field. So, the
version field is added into the compat endpoint only. The version field
is always 1, since Docker uses the version to keep track of updates to
the secret, and currently we cannot update a secret.

Closes https://github.com/containers/podman/issues/9316


mheon:
LGTM

Issue #9582, opened 2021-03-02
kata: running `podman --runtime /path/to/kata/binary` doesn't automatically apply the correct SELinux labels


fidencio:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When running `sudo podman --runtime /usr/bin/kata-runtime ...`, on Fedora 33, the SELinux labels are not correctly applied to the qemu & virtiofsd processes, leading to the pod not starting

**Steps to reproduce the issue:**

1. Ensure you have SELinux set as enforcing
2. Install kata-runtime package on Fedora (`sudo dnf install -y kata-runtime`)
3. Run `sudo podman --runtime /usr/bin/kata-runtime ...`

**Describe the results you received:**
Error: container create failed (no logs from conmon): EOF

**Describe the results you expected:**
Container would be created, in the same as if the command was called as: `sudo podman --runtime kata ...`.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 3.1.0-dev (but happens with the Fedora stock version as well).
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.6
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.21-3.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: 0f53fb68333bdead5fe4dc5175703e22cf9882ab'
  cpus: 4
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: machado
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.9.16-200.fc33.x86_64
  linkmode: dynamic
  memFree: 916963328
  memTotal: 8213188608
  ociRuntime:
    name: crun
    package: crun-0.16-3.fc33.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.16
      commit: eb0145e5ad4d8207e84a327248af76663d4e50dd
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
[...]

PR #9584, opened 2021-03-02
libpod: Apply SELinux KVM label if process contains "kata" in its name

fidencio:
We currently only apply the label when the user calls `podman --runtime
kata ...`, which may not happen in several different cases, including
testing binaries which are not part of the PATH.

In order to avoid forcing the user to pass `--security-opt
label=type:...`, let's follow a similar logic than the one implemented
on the CRI-O side and apply the KVM label to the processes in case the
runtime binary contains "kata" as part of its name.

Fixes: #9582

Signed-off-by: Fabiano Fidêncio <fabiano@fidencio.org>

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -467,7 +467,7 @@ func (c *Container) setupStorage(ctx context.Context) error {
 
 	processLabel := containerInfo.ProcessLabel
 	switch {
-	case c.ociRuntime.SupportsKVM():
+	case c.ociRuntime.SupportsKVM(), strings.Contains(strings.ToLower(filepath.Base(c.ociRuntime.Path())), "kata"):
 		processLabel, err = selinux.KVMLabel(processLabel)
 		if err != nil {
 			return err


brianredbeard:
At a minimum, shouldn't this do a bit more checking than a partial string match?

fidencio:
> At a minimum, shouldn't this do a bit more checking than a partial string match?

I'm open to suggestions, @brianredbeard.

By the way, here you can find what's done on CRI-O side:
https://github.com/cri-o/cri-o/blob/091b0208c3a4659ad5f2f66011a7be012d51d775/server/sandbox_run_linux.go#L864-L867

mheon:
This seems unnecessary? We added the `runtime_supports_kvm` config table already for this purpose. Runtimes requiring this should be marked in the config file. Doing it based on a string match instead seems questionable.

fidencio:
> This seems unnecessary? We added the `runtime_supports_kvm` config table already for this purpose. Runtimes requiring this should be marked in the config file. Doing it based on a string match instead seems questionable.

Okay, no biggie.

@rhatdan, we discussed this over the IRC, but seems we better **not** address this case.  At least now it's documented (as in, issue and discussions) that this is not something that was missed and it's the intended behaviour.

rhatdan:
Should we go the other way though.  Basically look at kata paths, and if we get a match then we do the SELinux labels.

# kata = [
#            "/usr/bin/kata-runtime",
#            "/usr/sbin/kata-runtime",
#            "/usr/local/bin/kata-runtime",
#            "/usr/local/sbin/kata-runtime",
#            "/sbin/kata-runtime",
#            "/bin/kata-runtime",
#            "/usr/bin/kata-qemu",
#            "/usr/bin/kata-fc",

Issue #9564, opened 2021-03-02
compatibility api should implement /auth endpoint


troyready:
/kind feature

**Description**

The docker api [/auth endpoint](https://docs.docker.com/engine/api/v1.40/#operation/SystemAuth) should be implemented.

**Steps to reproduce the issue:**

1. Post to /auth endpoint: `curl -X POST -s --unix-socket $XDG_RUNTIME_DIR/podman/podman.sock  http://localhost/auth`

**Describe the results you received:**
`{"cause":"","message":"Path /auth is not supported","response":0}`

**Additional information you deem important (e.g. issue happens only occasionally):**
This will allow the use of the [python SDK for docker login function](https://docker-py.readthedocs.io/en/stable/client.html#docker.client.DockerClient.login), which [uses this endpoint](https://github.com/docker/docker-py/blob/4.4.4/docker/api/daemon.py#L149)

**Output of `podman version`:**

```
Version:      3.0.0
API Version:  3.0.0
Go Version:   go1.15.8
Built:        Wed Dec 31 16:00:00 1969
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: 'conmon: /usr/bin/conmon'
    path: /usr/bin/conmon
    version: 'conmon version 2.0.25, commit: unknown'
  cpus: 8
  distribution:
    distribution: debian
    version: unknown
  eventLogger: journald
  hostname: stargazer
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.10.0-3-amd64
  linkmode: dynamic
  memFree: 29001703424
  memTotal: 50156310528
  ociRuntime:
    name: runc
    package: 'runc: /usr/bin/runc'
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0~rc93+ds1
      commit: 1.0.0~rc93+ds1-2
      spec: 1.0.2-dev
      go: go1.15.8
      libseccomp: 2.5.1
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #9589, opened 2021-03-03
add /auth for docker compatibility

troyready:
This endpoint just validates credentials:
https://github.com/moby/moby/blob/v20.10.4/api/swagger.yaml#L7936-L7977

Fixes: #9564

### DIFF
--- a/pkg/api/handlers/compat/auth.go
+++ b/pkg/api/handlers/compat/auth.go
@@ -1,0 +1,51 @@
+package compat
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	DockerClient "github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/podman/v3/pkg/api/handlers/utils"
+	"github.com/containers/podman/v3/pkg/domain/entities"
+	"github.com/containers/podman/v3/pkg/registries"
+	docker "github.com/docker/docker/api/types"
+	"github.com/pkg/errors"
+)
+
+func Auth(w http.ResponseWriter, r *http.Request) {
+	var authConfig docker.AuthConfig
+	err := json.NewDecoder(r.Body).Decode(&authConfig)
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrapf(err, "failed to parse request"))
+		return
+	}
+
+	skipTLS := types.NewOptionalBool(false)
+	if strings.HasPrefix(authConfig.ServerAddress, "http://localhost/") || strings.HasPrefix(authConfig.ServerAddress, "http://localhost:") {
+		// support for local testing
+		skipTLS = types.NewOptionalBool(true)
+	}
+
+	fmt.Println("Authenticating with existing credentials...")
+	sysCtx := types.SystemContext{
+		AuthFilePath:                "",
+		DockerCertPath:              "",
+		DockerInsecureSkipTLSVerify: skipTLS,
+		SystemRegistriesConfPath:    registries.SystemRegistriesConfPath(),
+	}
+	if err := DockerClient.CheckAuth(context.Background(), &sysCtx, authConfig.Username, authConfig.Password, authConfig.ServerAddress); err == nil {
+		utils.WriteResponse(w, http.StatusOK, entities.AuthReport{
+			IdentityToken: "",
+			Status:        "Login Succeeded",
+		})
+	} else {
+		utils.WriteResponse(w, http.StatusBadRequest, entities.AuthReport{
+			IdentityToken: "",
+			Status:        "login attempt to " + authConfig.ServerAddress + " failed with status: " + err.Error(),
+		})
+	}
+}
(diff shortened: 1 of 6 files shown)


troyready:
The core element of this is working now (tested and verified with AWS ECR).

I'm not sure how to proceed with a unit test for it. If I understand correctly, it would mean setting up a docker registry.

Would appreciate a second set of eyes on the testing for an idea of how to go forward.

mheon:
@edsantiago might have an answer on testing

edsantiago:
Ugh. This is a solved problem for our system tests in CI, but not for apiv2 tests. It's probably about a 2- to 3-hour effort to set it up for apiv2 tests; unfortunately I won't have that time until mid-March.

rhatdan:
LGTM
I am fine with getting this in, without testing in CI, since it is a feature we are missing.

rhatdan:
@baude @jwhonce PTAL

edsantiago:
Quick comment: I'm working on a testing framework for this. It's not ready, but one thing I've found is that this might not be compatible with docker. The [docker API documentation](https://docs.docker.com/engine/api/v1.41/#operation/SystemAuth) implies that `serveraddress` should be a full URL, of the form `https://host:port/v2`. That doesn't work with your code; your code works if I pass in just `host:port`. Can you look into this? (I might be mistaken).

edsantiago:
Also: the `tlsVerify(false)` code is never triggering; I can't figure out how to make it trigger (possibly because `serveraddress` is getting set internally somehow to `https://localhost` which will never match `http://localhost`)

Also: on successful auth I get `{"IdentityToken":"","Status":"Login Succeeded"}` - shouldn't `IdentityToken` be non-null?

edsantiago:
@troyready try this:
```console
$ make          ! (in your source tree, with these changes)
$ git pr 9669   ! requires git-extras
$ ./test/apiv2/test-apiv2 auth
Generating a RSA private key
.......................................++++
......++++
writing new private key to '/dev/shm/test-apiv2.tmp.33vdZT/registry/auth/domain.key'
-----
ok 1 [60-auth] POST /v1.40/auth [-d {"username":"uRIU5jdv","password":"WrOnGPassWord","serveraddress":"localhost:5066/"}] : status=400
ok 2 [60-auth] POST /v1.40/auth [-d {"username":"uRIU5jdv","password":"WrOnGPassWord","serveraddress":"localhost:5066/"}] : .Status ('login attempt to localhost:5066/ failed with status: unable to retrieve auth token: invalid username/password: unauthorized') ~ .* invalid username/password
ok 3 [60-auth] POST /v1.40/auth [-d {"username":"uRIU5jdv","password":"pVhgy0Bi","serveraddress":"localhost:5066/"}] : status=200
ok 4 [60-auth] POST /v1.40/auth [-d {"username":"uRIU5jdv","password":"pVhgy0Bi","serveraddress":"localhost:5066/"}] : .Status=Login Succeeded
not ok 5 [60-auth] POST /v1.40/auth [-d {"username":"uRIU5jdv","password":"pVhgy0Bi","serveraddress":"localhost:5066/"}] : .IdentityToken
#  expected: ~ [a-zA-Z0-9]
#    actual:
1..5
```
That will run my tests against your PR, with a local registry. It will fail, because See Above. To iterate:
```console
$ git co your-working-branch
edit-edit-edit
$ make
$ git co pr/9669
```
HTH.

(Postscript: the results above partly-succeed only because I hardcoded `skipTLS=false`)

troyready:
Thanks for looking into it @edsantiago !

> The docker API documentation implies that serveraddress should be a full URL, of the form https://host:port/v2. That doesn't work with your code; your code works if I pass in just host:port. Can you look into this? (I might be mistaken).

I think I ran into the same issue, where getting the http (no `s`) through from the api wasn't working (mangling it like pre-pending a `https//` or something akin to that). So as is, the `tlsVerify(false)` conditional isn't really useful.

To be clear, this seems to be specific to testing (which I think is the only scenario where TLS would be avoided -- this doesn't appear to be an issue when using it with a trusted `https://` address (my actual integration tests were run against an AWS ECR).

> Also: on successful auth I get {"IdentityToken":"","Status":"Login Succeeded"} - shouldn't IdentityToken be non-null?

The IdentityToken appears to be optional; when testing the same API using Docker against AWS ECR, I get the same empty string.

troyready:
Thanks again @edsantiago -- with the new test setup I was able to figure out what was going wrong. Latest commit fixes the issue: tests are working now, and I've confirmed it still works with a normal live (trusted TLS) repo as well.

(I'm pretty sure the zuul build failure is unrelated to changes here)

edsantiago:
/hold

@troyready could you please rebase, then update your added `.at` file to reflect the new parameter format (space-separated)? Now that #9686 has merged, if your PR merges it will break all of CI for everyone. I apologize for the inconvenience but hope that the new POST param format makes things easier.

troyready:
Sure thing, done @edsantiago

edsantiago:
Tests LGTM but I'll leave final approval to @jwhonce or @baude. Thank you for your efforts on this.

>IdentityToken appears to be optional

I still find this weird, and pointless, but hey, if it matches docker, shrug. Thank you for testing against docker.

TomSweeneyRedHat:
LGTM

rhatdan:
Great work @troyready Thanks.
/lgtm
/approve
/hold cancel

PR #9592, opened 2021-03-03
Numerous buildah fixes found by Ed's testing of buildah tests against podman.

rhatdan:
Add support for podman build --ignorefile
podman build --build-arg should fall back to environment
Handle podman build --dns-search
Fix podman build --pull-never

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -318,7 +318,12 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 			if len(av) > 1 {
 				args[av[0]] = av[1]
 			} else {
-				delete(args, av[0])
+				// check if the env is set in the local environment and use that value if it is
+				if val, present := os.LookupEnv(av[0]); present {
+					args[av[0]] = val
+				} else {
+					delete(args, av[0])
+				}
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


cmd/podman/images/build.go
```diff
@@ -1,7 +1,9 @@
 package images
 
 import (
+	"fmt"
```
edsantiago, line 4: Looks like you need to drop this

test/system/070-build.bats
```diff
@@ -358,10 +362,89 @@ Labels.$label_name | $label_value
...
+test*
+!test2*
+subdir
+!*/sub1*
+EOF
+
+        # Build an image. For .dockerignore
+        local -a ignoreflag
+        if [[ $ignorefile != ".dockerignore" ]]; then
+            ignoreflag="--ignorefile $ignorefile"
```
edsantiago, line 418: Oops. I guess this has to be `--ignorefile $tmpdir/$ignorefile`. (Sorry, for obvious reasons I was unable to test this yesterday. I had assumed that podman/buildah would treat paths as relative to the context directory; and TBH I kind of am surprised that it doesn't. So maybe this is a code bug. Anyhow, here's the reason).

TomSweeneyRedHat:
The testing needs more help....

edsantiago:
Looks like dns-search isn't actually working?

edsantiago:
LGTM!

vrothberg (commented):
/lgtm

Nice work!

Issue #7228, opened 2020-08-04
CI: flake: read unixpacket [conmon path]


edsantiago:
As predicted, we're starting to see flakes in CI due to podman-remote tests being enabled. This one manifests as:
```
$ [some podman command]
[+0242s] # read unixpacket @->/var/run/libpod/socket/99a264c5e5fcaed668d0413741c63bfb90bdcdb02e1ad4ad109d7f6fe064fd84/attach: read: connection reset by peer
```
Seen in [special testing rootless](https://github.com/containers/podman/pull/7224/checks?check_run_id=946573502) in #7224 

Seen in [fedora 31 test](https://github.com/containers/podman/pull/7222/checks?check_run_id=946592904) in #7222

I'm sorry not to be able to provide more information. I tried to reproduce this one last week and was unable to.

edsantiago:
It is possible that this and #7195 are the same problem, but until there's evidence for that let's leave them as separate issues.

edsantiago:
* fedora-32 : test fedora-32
  * PR #7081
    * [08-04 16:39](https://api.cirrus-ci.com/v1/task/5219124764737536/logs/system_test.log)

edsantiago:
Wow - a [triple failure](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5594465345732608/html/integration_test.log.html#t--podman-run-user-capabilities-test--1)

edsantiago:
See https://github.com/containers/podman/issues/7195#issuecomment-673651534 for a reproducer.

I'm still not convinced that this is the same as #7195: #7195 is "no output received, and sometimes there are other errors". This one is "read unixpacket error". The above reproducer triggers this error _sometimes_, but not reliably. Right now #7195 is the higher priority; until that one is understood, I still lean slightly toward keeping these two separate. (But as always I'm willing to listen to counterarguments).

edsantiago:
Whoa - just now seeing this in [podman gating tests](https://jenkins-continuous-infra.apps.ci.centos.org/blue/organizations/jenkins/fedora-rawhide-build-pipeline/detail/fedora-rawhide-build-pipeline/15088/pipeline/#step-177-log-1104), which don't even pretend to test podman-remote:
```
$ /usr/bin/podman run --rm build_test cat /subtest/myfile1
[...]

Issue #3302, opened 2019-06-11
[intermittent]: podman run -a stdin&stderr: read unixpacket: connection reset by peer


edsantiago:
This is a combination that I can find no realistic use for, but even so I think merits attention because it might present itself in other more real-world situations:
```
# echo true | podman run -a stdin -a stderr --tty=false alpine sh
Error: error attaching to container 22e7b45a00508b0e60a27288bb6c745100a2ee42335800de1d8e246cfe24bd48: read unixpacket @->/var/run/libpod/socket/22e7b45a00508b0e60a27288bb6c745100a2ee42335800de1d8e246cfe24bd48/attach: read: connection reset by peer
```
Reproduces maybe one in five attempts; the rest of the time it succeeds. It also fails with `echo ls /` and `echo ls /sdf`, but does not (seem to) fail with `</dev/null` (redirection, not pipe).

`--log-level=debug` adds nothing of value AFAICT—both pass and fail look identical to my eye—but I will provide on request.

podman-1.4.0-2.fc29 and fc30; but I think I've seen it before, just never taken time to pursue it. I can look in logs if necessary.

rhatdan:
Must be a race condition.

edsantiago:
Still present in podman-1.4.2-1.fc29 and fc30.

There's another failure which I think must be related. Instead of `sh`, run something that outputs to stderr:
```
# echo hi | podman run -a stdin -a stderr --interactive --tty=false  alpine ls /nonesuch
ls: /nonesuch: No such file or directory
```
At this point it hangs - AFAICT forever. `^D` has no effect. `podman ps` in a separate window shows the container running. `^C` yields:
```
^CERRO[0000] container not running
container not running
ERRO[0178] Error forwarding signal 2 to container 0af91336dd3700b2acd9889e92cd91e558a016510efc9dc5f5bd669ef921d69a: error sending si
gnal to container 0af91336dd3700b2acd9889e92cd91e558a016510efc9dc5f5bd669ef921d69a: `/usr/bin/runc kill 0af91336dd3700b2acd9889e92cd
91e558a016510efc9dc5f5bd669ef921d69a 2` failed: exit status 1
```
"container not running" even though a few seconds ago `podman ps` showed it running and, right now, after the `^C`, `podman ps` still shows it running. But:
```
# podman exec 0af9 date
ERRO[0000] exec failed: cannot exec a container that has stopped
exec failed: cannot exec a container that has stopped
Error: exit status 1
# podman ps
[...]

PR #9593, opened 2021-03-03
podman cp: support copying on tmpfs mounts

vrothberg:
Traditionally, the path resolution for containers has been resolved on
the *host*; relative to the container's mount point or relative to
specified bind mounts or volumes.

While this works nicely for non-running containers, it poses a problem
for running ones.  In that case, certain kinds of mounts (e.g., tmpfs)
will not resolve correctly.  A tmpfs is held in memory and hence cannot
be resolved relatively to the container's mount point.  A copy operation
will succeed but the data will not show up inside the container.

To support these kinds of mounts, we need to join the *running*
container's mount namespace (and PID namespace) when copying.

Note that this change implies moving the copy and stat logic into
`libpod` since we need to keep the container locked to avoid race
conditions.  The immediate benefit is that all logic is now inside
`libpod`; the code isn't scattered anymore.

Further note that Docker does not support copying to tmpfs mounts.

Tests have been extended to cover *both* path resolutions for running
and created containers.  New tests have been added to exercise the
tmpfs-mount case.

@edsantiago, I'd love your input especially on the test changes

### DIFF
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -2,6 +2,7 @@ package libpod
 
 import (
 	"context"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -349,10 +350,6 @@ func (c *Container) Mount() (string, error) {
 		}
 	}
 
-	if c.state.State == define.ContainerStateRemoving {
-		return "", errors.Wrapf(define.ErrCtrStateInvalid, "cannot mount container %s as it is being removed", c.ID())
-	}
-
 	defer c.newContainerEvent(events.Mount)
 	return c.mount()
 }
@@ -367,7 +364,6 @@ func (c *Container) Unmount(force bool) error {
 			return err
 		}
 	}
-
 	if c.state.Mounted {
 		mounted, err := c.runtime.storageService.MountedContainerImage(c.ID())
 		if err != nil {
@@ -847,31 +843,59 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
 	return c.shouldRestart()
 }
 
-// ResolvePath resolves the specified path on the root for the container.  The
-// root must either be the mounted image of the container or the already
-// mounted container storage.
-//
-// It returns the resolved root and the resolved path.  Note that the path may
-// resolve to the container's mount point or to a volume or bind mount.
-func (c *Container) ResolvePath(ctx context.Context, root string, path string) (string, string, error) {
-	logrus.Debugf("Resolving path %q (root %q) on container %s", path, root, c.ID())
-
-	// Minimal sanity checks.
-	if len(root)*len(path) == 0 {
-		return "", "", errors.Wrapf(define.ErrInternal, "ResolvePath: root (%q) and path (%q) must be non empty", root, path)
-	}
-	if _, err := os.Stat(root); err != nil {
-		return "", "", errors.Wrapf(err, "cannot locate root to resolve path on container %s", c.ID())
-	}
-
+// CopyFromArchive copies the contents from the specified tarStream to path
+// *inside* the container.
+func (c *Container) CopyFromArchive(ctx context.Context, containerPath string, tarStream io.Reader) (func() error, error) {
 	if !c.batched {
 		c.lock.Lock()
 		defer c.lock.Unlock()
 
 		if err := c.syncContainer(); err != nil {
-			return "", "", err
+			return nil, err
 		}
 	}
 
-	return c.resolvePath(root, path)
+	return c.copyFromArchive(ctx, containerPath, tarStream)
+}
+
+// CopyToArchive copies the contents from the specified path *inside* the
+// container to the tarStream.
+func (c *Container) CopyToArchive(ctx context.Context, containerPath string, tarStream io.Writer) (func() error, error) {
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return nil, err
+		}
+	}
+
+	return c.copyToArchive(ctx, containerPath, tarStream)
+}
+
+// Stat the specified path *inside* the container and return a file info.
+func (c *Container) Stat(ctx context.Context, containerPath string) (*define.FileInfo, error) {
+	if !c.batched {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if err := c.syncContainer(); err != nil {
+			return nil, err
+		}
+	}
+
+	var mountPoint string
+	var err error
+	if c.state.Mounted {
+		mountPoint = c.state.Mountpoint
+	} else {
+		mountPoint, err = c.mount()
+		if err != nil {
+			return nil, err
+		}
+		defer c.unmount(false)
+	}
+
+	info, _, _, err := c.stat(ctx, mountPoint, containerPath)
+	return info, err
 }
--- a/libpod/container_copy_linux.go
+++ b/libpod/container_copy_linux.go
@@ -1,0 +1,266 @@
+// +build linux
+
+package libpod
+
+import (
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	buildahCopiah "github.com/containers/buildah/copier"
+	"github.com/containers/buildah/pkg/chrootuser"
+	"github.com/containers/buildah/util"
+	"github.com/containers/podman/v3/libpod/define"
+	"github.com/containers/storage"
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/docker/docker/pkg/archive"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+func (c *Container) copyFromArchive(ctx context.Context, path string, reader io.Reader) (func() error, error) {
+	var (
+		mountPoint   string
+		resolvedRoot string
+		resolvedPath string
+		unmount      func()
+		err          error
+	)
+
+	// Make sure that "/" copies the *contents* of the mount point and not
+	// the directory.
+	if path == "/" {
+		path = "/."
+	}
+
+	// Optimization: only mount if the container is not already.
+	if c.state.Mounted {
+		mountPoint = c.state.Mountpoint
+		unmount = func() {}
+	} else {
+		// NOTE: make sure to unmount in error paths.
+		mountPoint, err = c.mount()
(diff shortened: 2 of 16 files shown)


vrothberg:
@TomSweeneyRedHat FWIW, I would _love_ to get this into 3.0 for two reasons: 1) It's a really thing and customers are asking for it (we can consider it a bug), 2) it will make backports much easier (there are still some bugs we need to iron out in cp).

Is a backport for RHEL possible?

libpod/container_copy_linux.go
```diff
@@ -0,0 +1,237 @@
...
+	decompressed, err := archive.DecompressStream(reader)
+	if err != nil {
+		c.unmount(false)
+		return nil, err
+	}
+
+	idMappings, idPair, err := getIDMappingsAndPair(c, mountPoint)
+	if err != nil {
+		c.unmount(false)
+		return nil, err
```
rhatdan, line 60: Should this have `decompressed.Close()`

rhatdan (approved):
Nice work.

libpod/container_copy_linux.go
vrothberg: Good catch!  Yes, we need to close in this return path as well.

libpod/container_copy_linux.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (c *Container) copyFromArchive(ctx context.Context, path string, reader io.Reader) (func() error, error) {
+	// Make sure that "/" copies the *contents* of the mount point and not
+	// the directory.
+	if path == "/" {
+		path = "/."
+	}
+
+	// NOTE: make sure to unmount in error paths.
+	mountPoint, err := c.mount()
```
mheon, line 34: Probably should be conditional on the container not being mounted (c.state.Mounted)

vrothberg: Nice idea!  That should speed things up.  Added that to all places.

edsantiago:
Just an FYI: in one of my test runs, I encountered a flake:
```
 ✗ podman cp file from host to container
...
$ podman start -a cpcontainer
   random-0-qvvUDTQtME
   Error: unable to start container f0191a2a6a86a3e78252b8d095a50a24b21980ea80f8c71de8df9c98703b9428: error attaching to container f0191a2a6a86a3e78252b8d095a50a24b21980ea80f8c71de8df9c98703b9428: read unixpacket @->/proc/self/fd/13/attach: read: connection reset by peer
   [ rc=125 (** EXPECTED 0 **) ]
```
"read unixpacket" has a horrible history of flakiness, usually relating to conmon: #7228, #3302. This was on my laptop, f33, conmon-2.0.26-1.fc33, 5.10.19-200.fc33.

edsantiago:
May I make a friendly suggestion? This shaves over one minute from the test's run time (126 seconds down to 58):
[cp-fast.diff.txt](https://github.com/containers/podman/files/6076975/cp-fast.diff.txt)

Reason: `podman rm -f` has to wait those stupid ten seconds, and there's no `--do-it-right-now-dammit` option.

test/system/065-cp.bats
```diff
@@ -68,6 +72,38 @@ load helpers
...
+    srcdir=$PODMAN_TMPDIR/cp-test-file-host-to-ctr
+    mkdir -p $srcdir
+    content=$(random_string 20)
+    echo $content > $srcdir/file
+
+    # RUNNING container
+    run_podman run -d --mount type=tmpfs,dst=/tmp --name cpcontainer $IMAGE sleep infinity
+    run_podman cp $srcdir/file cpcontainer:/tmp
+    run_podman exec cpcontainer cat /tmp/file
+    is "$output" "${content}" "cp to container's tmpfs"
```
edsantiago, line 98: A lot of people—myself first among those—are simple monkeys who, when we see an error message, grep directly for it in the test sources to find the relevant place. When error messages are identical in multiple places, brainless monkeys like me get confused.

Would you consider changing this to "cp to running container's tmpfs", and line 105 to "cp to created container's tmpfs"?

test/system/065-cp.bats
```diff
@@ -68,6 +72,38 @@ load helpers
...
+        run_podman rm -f cpcontainer
+    done < <(parse_table "$tests")
+
+    run_podman rmi -f $cpimage
+}
+
+@test "podman cp file from host to container tmpfs mount" {
+    srcdir=$PODMAN_TMPDIR/cp-test-file-host-to-ctr
+    mkdir -p $srcdir
+    content=$(random_string 20)
```
edsantiago, line 91: I've developed a new habit when writing tests like this: adding placeholder markers to random content.
```bash
   content=tmpfile-content-$(random_string 20)
```
The main benefit is, when there's a test failure, the `tmpfile-content` text makes it very easy for a maintainer to see the error in better context: to understand what the expectation is and where it's set. Also importantly, when reading a long error log, it makes it easier to see which random string is which, and ignore the red herrings.

test/system/065-cp.bats
```diff
@@ -110,8 +149,23 @@ load helpers
...
+    run_podman create --name cpcontainer --workdir=/srv $cpimage
+    while read id src dest dest_fullname description; do
+        # dest may be "''" for empty table cells
+        if [[ $dest == "''" ]];then
+            unset dest
+        fi
+        run_podman cp cpcontainer:$src "$srcdir$dest"
+        run cat $srcdir$dest_fullname
+        is "$output" "${randomcontent[$id]}" "$description (cp ctr:$src to \$srcdir$dest)"
+        rm $srcdir/$dest_fullname
```
edsantiago, line 164: Consistency: slash is not needed.

test/system/065-cp.bats
```diff
@@ -110,8 +149,23 @@ load helpers
...
+    # Created container
+    run_podman create --name cpcontainer --workdir=/srv $cpimage
+    while read id src dest dest_fullname description; do
+        # dest may be "''" for empty table cells
+        if [[ $dest == "''" ]];then
+            unset dest
+        fi
+        run_podman cp cpcontainer:$src "$srcdir$dest"
+        run cat $srcdir$dest_fullname
+        is "$output" "${randomcontent[$id]}" "$description (cp ctr:$src to \$srcdir$dest)"
```
edsantiago, line 163: Hint:
```bash
    (remove the "run cat")
    is "$(< $srcdif$dest_fullname)" "${random...."
```
`$(< PATH)` is an idiomatic equivalent to `$(cat PATH)`

vrothberg: Nice!  I'll get that consistent across the copy tests.

test/system/065-cp.bats
vrothberg: I'll add that. That can be super helpful!

test/system/065-cp.bats
vrothberg: Sure!

vrothberg:
> May I make a friendly suggestion? This shaves over one minute from the test's run time (126 seconds down to 58):
> [cp-fast.diff.txt](https://github.com/containers/podman/files/6076975/cp-fast.diff.txt)
> 
> Reason: `podman rm -f` has to wait those stupid ten seconds, and there's no `--do-it-right-now-dammit` option.

Yes, I was afraid of that.  I will add ` podman kill` before to cut things short :^)

vrothberg:
NOTE: we currently cannot copy tty devices from a rootless container. One `e2e` test is disabled for now. I chatted with @nalind and he has an idea how to tackle that in the copier package.

I think we can merge. It's a very minor corner case and I will be adding a couple more commits to address other `cp` issues tomorrow.

vrothberg:
Updated.  Thanks for the super helpful reviews!

libpod/container_api.go
```diff
@@ -847,31 +843,57 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
...
-// ResolvePath resolves the specified path on the root for the container.  The
-// root must either be the mounted image of the container or the already
-// mounted container storage.
-//
-// It returns the resolved root and the resolved path.  Note that the path may
-// resolve to the container's mount point or to a volume or bind mount.
-func (c *Container) ResolvePath(ctx context.Context, root string, path string) (string, string, error) {
-	logrus.Debugf("Resolving path %q (root %q) on container %s", path, root, c.ID())
+// CopyFromArchive copies the contents of the specified tarStream to path.
+func (c *Container) CopyFromArchive(ctx context.Context, path string, tarStream io.Reader) (func() error, error) {
```
mheon, line 847: Can you make it more clear that this one copies into the container, the other one copies out of it? Updating the comments on both could help.

libpod/container_api.go
```diff
@@ -847,31 +843,57 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
...
+
+		if err := c.syncContainer(); err != nil {
+			return nil, err
+		}
 	}
 
+	return c.copyToArchive(ctx, path, tarStream)
+}
+
+// Stat the specified path and return file info.
```
mheon, line 874: Can you make it clear that this stats the path *in the container*?

libpod/container_api.go
```diff
@@ -847,31 +843,57 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
...
+func (c *Container) Stat(ctx context.Context, path string) (*define.FileInfo, error) {
 	if !c.batched {
 		c.lock.Lock()
 		defer c.lock.Unlock()
```
mheon, line 878: Should we consider making the container unlock after initial checks? Right now, `podman ps` while copying will block, which is not good - but at the same time, a container that stops while we are copying would be able to unmount the container from under us.

vrothberg: I think it's okay for `Stat` since it's fast, but I will look into optimizing it for the two copy cases.

vrothberg:
@edsantiago 

> read: connection reset by peer

Just hit that in the F33 root system tests as well.

vrothberg:
Repushing. Many jobs timed out (Cc @cevich).

vrothberg:
> > May I make a friendly suggestion? This shaves over one minute from the test's run time (126 seconds down to 58):
> > [cp-fast.diff.txt](https://github.com/containers/podman/files/6076975/cp-fast.diff.txt)
> > Reason: `podman rm -f` has to wait those stupid ten seconds, and there's no `--do-it-right-now-dammit` option.
> 
> Yes, I was afraid of that. I will add ` podman kill` before to cut things short :^)

Unfortunately, `start -a`seems to be too flaky/broken. I'll change the copy tests but it's concerning.

mheon:
We really ought to add `--timeout` to `podman rm`

rhatdan:
I agree podman rm --timeout 0 -f is needed.

giuseppe (approved):
/lgtm

Issue #9594, opened 2021-03-03
Podman 3.0.1 with docker-compose segfaults when files with wildcard don't exist


sshnaidm:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. Run docker-compose build on attached docker-compose.yaml file

```
version: '3.6'

services:
  container:
    build: .
    container_name: sidecar

```

and Dockerfile:

```
FROM python:3.6

COPY *.json some_file /usr/local/bin/
CMD bash
```
where `some_file` **does** exist, but no `*.json` files in the folder.

**Describe the results you received:**

```
Building container
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
STEP 1: FROM python:3.6
STEP 2: COPY some_file /usr/local/bin/
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x5614104c64f5]
 
goroutine 1 [running]:
panic(0x561410cc4d00, 0x561411cfdce0)
        /usr/lib/golang/src/runtime/panic.go:1064 +0x471 fp=0xc0002fb410 sp=0xc0002fb358 pc=0x56140f2d9ea1
runtime.panicmem(...)
        /usr/lib/golang/src/runtime/panic.go:212
runtime.sigpanic()
        /usr/lib/golang/src/runtime/signal_unix.go:717 +0x3de fp=0xc0002fb440 sp=0xc0002fb410 pc=0x56140f2f12fe
github.com/containers/podman/cmd/podman/images.build(0x561411d21820, 0xc000144050, 0x1, 0x5, 0x0, 0x0)
        /home/abuild/rpmbuild/BUILD/podman-3.0.1/_build/src/github.com/containers/podman/cmd/podman/images/build.go:274 +0x4e5 fp=0xc0002fbd20 sp=0xc0002fb440 pc=0x5614104c64f5
github.com/containers/podman/vendor/github.com/spf13/cobra.(*Command).execute(0x561411d21820, 0xc000134020, 0x5, 0x5, 0x561411d21820, 0xc000134020)
[...]

PR #9595, opened 2021-03-03
Fix possible panic with podman build --iidfile

Luap99:
Make sure we check for errors before trying to write the iidfile.

[NO TESTS NEEDED]

Ref #9594


baude:
lgtm

mheon:
LGTM once the flake goes green

PR #9599, opened 2021-03-03
Add tests for selinux kvm/init labels

rhatdan:
spc_t tests should be able to run rootless as well.

### DIFF
--- a/test/system/410-selinux.bats
+++ b/test/system/410-selinux.bats
@@ -77,13 +77,14 @@ function check_label() {
 @test "podman selinux: inspect kvm labels" {
     skip_if_no_selinux
     skip_if_remote "runtime flag is not passed over remote"
-    if [ ! -e /usr/bin/kata-runtime ]; then
-        skip "kata-runtime not available"
-    fi
 
-    run_podman create --runtime=kata --name myc $IMAGE
+    tmpdir=$PODMAN_TMPDIR/kata-test
+    mkdir -p $tmpdir
+    KATA=${tmpdir}/kata-runtime
+    ln -s /bin/true ${KATA}
+    run_podman create --runtime=${KATA} --name myc $IMAGE
     run_podman inspect --format='{{ .ProcessLabel }}' myc
-    is "$output" ".*container_kvm_t.*"
+    is "$output" ".*container_kvm_t"
 }
 
 # pr #6752


rhatdan:
@edsantiago PTAL

test/system/410-selinux.bats
```diff
@@ -74,6 +74,15 @@ function check_label() {
     check_label "--security-opt label=level:s0:c1,c2" "container_t" "s0:c1,c2"
 }
 
+@test "podman selinux: inspect kvm labels" {
+    skip_if_no_selinux
+    skip_if_remote "runtime flag is not passed over remote"
+
+    run_podman create --runtime=kata --name myc $IMAGE
```
edsantiago, line 81: Error: default OCI runtime "kata" not found: invalid argument

rhatdan: If you touch /usr/bin/kata-runtime does the test work?

edsantiago: Almost. This works: `ln -s true /usr/bin/kata-runtime`. Point is, this new test is going to fail in CI (and Fedora/RHEL gating) because of the missing runtime. It might be prudent to:
```bash
    if ! -x /usr/bin/kata-runtime; then
        skip "kata runtime unavailable"
    fi
```

(And I forgot to mention: your rootless changes work fine. Is that a new feature, or was I mistaken when I disabled those for rootless?)

rhatdan: I think it is a fairly new feature.

BTW, I think I can fake this test out by doing

ln -s /bin/true  $TMPDIR/kata-runtime

Then set the runtime to this, which should trigger the code to do the right thing, once my other kata PR gets merged.

rhatdan:
@edsantiago PTAL, this just turns on a test that we currently ignore.

TomSweeneyRedHat:
LGTM

ashley-cui (commented):
/lgtm

edsantiago:
Day late dollar short, but why did you revert the `skip_if_rootless` changes? Did they fail CI in some environment? Unfortunately, the commit message still mentiones "spc_t tests should be able to run rootless as well".

rhatdan:
A separate PR got those I believe.

```
$ grep skip_if_rootless 410-selinux.bats 
$ 
```

PR #9600, opened 2021-03-03
[v2.2] Cirrus: Skip frequent flake on v2.2 branch

cevich:
**Derpends on:** #9620

This test frequently flakes on at least one platform during daily
cirrus-cron runs on the **v2.2** branch.  Since this is not happening on
other versions, assume it's some kind of difficult-to-find/fix race
condition.  Skip the test on all platforms to cut down on noise.

Example: https://cirrus-ci.com/task/4715979050582016?command=main#L75
(Same failure observed on Ubuntu as well)

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -544,6 +544,7 @@ json-file | f
 }
 
 @test "podman run with --net=host and --port prints warning" {
+    skip "Frequently flakes on multiple platforms on v2.2 branch"
     run_podman run -d --rm -p 8080 --net=host $IMAGE ls > /dev/null
     is "$output" ".*Port mappings have been discarded as one of the Host, Container, Pod, and None network modes are in use"
 }


edsantiago:
This is #7340. Unfortunately, your `skip` isn't going to help: this bug manifests itself in pretty much any test.

cevich:
>this bug manifests itself in pretty much any test.

For the last three days, it's consistently hit this one exact test on Fedora or Ubuntu.  Whatever the cause is, we just want it to stop :grin: 

@edsantiago  Do you think it would help to backport any/all of the "fix" PRs (mentioned in that issue) to this branch?

edsantiago:
>backport any/all of the "fix" PRs

That's a risk evaluation I have no idea how to make: I don't know what this branch is used for, nor how important it is, nor how difficult it would be to backport #7561 and #7574. 

> Whatever the cause is, we just want it to stop

Oh, understood, and I'm fine with that. I should've been more clear: this PR will not fix anything. It will be whack-a-mole. It may silence the error in one particular test, but the error will recur (in integration tests too). I was just trying to offer a heads-up.

I would've lgtm'ed this morning, but tests were red. I would lgtm right now, but tests are still red.

cevich:
> That's a risk evaluation I have no idea how to make

If you need to ask...it's probably not worth the risk.  We can keep this in mind if it does turn out to be wack-a-mole.

> I would've lgtm'ed this morning, but tests were red. 

I wanted to check with you first before spending time investigating that.  No idea why they all failed.

cevich:
Closing in favor of https://github.com/containers/podman/pull/9664

PR #9620, opened 2021-03-04
Fix images_test using inconsistent alpine:latest

cevich:
At some point, somehow the `quay.io/libpod/apline:latest` image changed.
This is causing integration tests to fail on this branch.  Mainly
these failures are caused by missed-expectations w/ the test image. Fix
these test by using a specific, fixed, named image tag instead of
'latest' (which can change unpredictably).

### DIFF
--- a/test/e2e/images_test.go
+++ b/test/e2e/images_test.go
@@ -188,23 +188,23 @@ WORKDIR /test
 	})
 
 	It("podman images filter since image", func() {
-		dockerfile := `FROM quay.io/libpod/alpine:latest
+		dockerfile := `FROM quay.io/libpod/alpine:3.2
 `
 		podmanTest.BuildImage(dockerfile, "foobar.com/before:latest", "false")
 		result := podmanTest.Podman([]string{"images", "-q", "-f", "since=quay.io/libpod/alpine:latest"})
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

mheon:
We should start thinking of a deprecation of the v2.2 branch upstream in the next month or so.

cevich:
> We should start thinking of a deprecation of the v2.2 branch upstream in the next month or so.

I'm all for it, just say the word.

mheon:
As I think about it more, I think we can do it now. We have to retain v2.2.1-rhel until RHEL 8.4 releases upstream, but I don't see us doing another 2.2.x release now that 3.0.x is out.

mheon:
@cevich Should we formalize this somewhere? We only support branch X until a newer branch, Y, is created and has a release done from it?

cevich:
> @cevich Should we formalize this somewhere? We only support branch X until a newer branch, Y, is created and has a release done from it?

I think it's a great idea.  That would help me out, and set community expectations.  Maybe we go one more step and also update the README.md to indicate the release is not supported?

cevich:
Closing in favor of https://github.com/containers/podman/pull/9664

PR #9625, opened 2021-03-05
Bump containersnetworking/plugins to v0.9.1

TomSweeneyRedHat:
Bumps containersnetworking/plugins to v0.9.1 to address the issue in https://bugzilla.redhat.com/show_bug.cgi?id=1932399.

A container might not be able to be exit if it has IPV6.  But that said, I'm not 100% sure we need this for Podman in 8.4.  @mheon @rhatdan, thoughts?
 

rhatdan:
This should go into the main branch first and then cherry-pick back here, if necessary.

mheon:
I don't think this is necessary, but I don't think it hurts anything either.

rhatdan:
Since this is not adding value and only adds risk, I am going to close.  RHEL will get this in RHEL8.4.0.2

Issue #9573, opened 2021-03-02
podman build: --pull-never is a NOP


edsantiago:
```console
$ printf "FROM quay.io/libpod/alpine:nosuchtag\n" | ./bin/podman build --pull-never -
STEP 1: FROM quay.io/libpod/alpine:nosuchtag
Error: error creating build container: Error initializing source docker://quay.io/libpod/alpine:nosuchtag: Error reading manifest nosuchtag in quay.io/libpod/alpine: manifest unknown: manifest unknown
```
Unexpected: it actually tried to pull. Compare to buildah:
```
STEP 1: FROM quay.io/libpod/alpine:nosuchtag
error creating build container: pull policy is "never" but "quay.io/libpod/alpine:nosuchtag" could not be found locally
```

rhatdan:
I will trade you one test for one fix. :^)

edsantiago:
One test: [pull-never.patch.txt](https://github.com/containers/podman/files/6071963/pull-never.patch.txt)

(exit status and error message checks are a WAG; please edit as appropriate)

PR #9631, opened 2021-03-05
Fix podman build --pull-never

rhatdan:
Fixes: https://github.com/containers/podman/issues/9573

### DIFF
--- a/cmd/podman/images/build.go
+++ b/cmd/podman/images/build.go
@@ -304,6 +304,21 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 		return nil, err
 	}
 
+	pullFlagsCount := 0
+	if c.Flag("pull").Changed {
+		pullFlagsCount++
+	}
+	if c.Flag("pull-always").Changed {
+		pullFlagsCount++
+	}
+	if c.Flag("pull-never").Changed {
+		pullFlagsCount++
+	}
+
+	if pullFlagsCount > 1 {
+		return nil, errors.Errorf("can only set one of 'pull' or 'pull-always' or 'pull-never'")
+	}
+
 	pullPolicy := define.PullIfMissing
 	if c.Flags().Changed("pull") && flags.Pull {
 		pullPolicy = define.PullAlways
@@ -313,7 +328,7 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 	}
 
 	if flags.PullNever {
-		pullPolicy = define.PullIfMissing
+		pullPolicy = define.PullNever
 	}
 
 	args := make(map[string]string)
--- a/contrib/cirrus/setup_environment.sh
+++ b/contrib/cirrus/setup_environment.sh
@@ -36,6 +36,9 @@ do
     fi
 done
 
+# Make sure cni network plugins directory exists
+mkdir -p /etc/cni/net.d
+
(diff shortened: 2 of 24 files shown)


rhatdan:
This PR Requires https://github.com/containers/buildah/pull/3062

go.mod
```diff
@@ -11,19 +11,19 @@ require (
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.0
-	github.com/containers/buildah v1.19.7
-	github.com/containers/common v0.35.0
+	github.com/containers/buildah v1.19.2-0.20210305112159-898eea8fbefd
```
TomSweeneyRedHat, line 14: Shouldn't this be 1.20 something?  I saw the commit in master, maybe  I missed the 1.19 commit?

cmd/podman/images/build.go
```diff
@@ -307,7 +323,7 @@ func buildFlagsWrapperToOptions(c *cobra.Command, contextDir string, flags *buil
 	}
 
 	if flags.PullNever {
-		pullPolicy = imagebuildah.PullIfMissing
+		pullPolicy = imagebuildah.PullNever
```
TomSweeneyRedHat, line 326: OH OUCH!

go.mod
rhatdan: We have not created a 1.20 yet, so I just put in the current commit id.  Since we have not updated the main branch beyond v1.19.2 this is what it show up as.
IE I set the version to
898eea8fbefd73993406610311cad8fcab045bad

And then make vendor-in-container changes it to the above.  BTW I think it might be time to cut a new buildah release.

cevich:
> what are running inside of a privileged container?

Should just be the ones that have `container` at the end of the name.  i.e. `int podman fedora-33 root container`

edsantiago:
I'm seeing `Files changed: 131`. Is that deliberate? I don't remember the earlier iteration.

rhatdan:
@edsantiago This is all vendoring in a new version of Buildah.  If we had a new release this PR would be much smaller.

edsantiago:
OK. I'm confused about the vendor downgrade from buildah 1.19.8 to .2.something, but will leave that for others to worry about.

If you need to repush for any reason, what would you think of deleting these lines from the `bud` test?

https://github.com/containers/podman/blob/af91f27d85f4104267eea493f8588f8c6a2d01e3/test/buildah-bud/buildah-tests.diff#L106-L113

rhatdan:
Sure.

edsantiago:
bud test failing:
```
 + git clone -q --branch v1.19.2-0.20210325094324-0c0ce48c77e8 https://github.com/containers/buildah test-buildah-v1.19.2-0.20210325094324-0c0ce48c77e8
 fatal: Remote branch v1.19.2-0.20210325094324-0c0ce48c77e8 not found in upstream origin
```
This is probably my fault: my bud test assumes that whatever is in `go.mod` can be checked out as a tag from the buildah repo. It looks like I need some way of detecting magic strings like `v<xyz>-something-sha`. What should be the exact git clone command for a string like that?

rhatdan:
@cevich We should have a new major release soon, in buildah, so there should be no rush.

umohnani8:
LGTM

PR #9641, opened 2021-03-06
[ci skip] Bad formatting fix in build documentation

jmguzik:
Spotted bad formatting when searching in podman build doc. Visible here (grep for Building an multi-architecture image ...):
http://docs.podman.io/en/latest/markdown/podman-build.1.html


TomSweeneyRedHat:
OOPS!  Yep, good catch, TYVM @jmguzik 
LGTM

PR #9649, opened 2021-03-07
Allow users to generate a kubernetes yaml off non running containers

rhatdan:
Currently if you attempt to create a kube.yaml file off of a non running
container where the container runs as a specific User, the creation
fails because the storage container is not mounted. Podman is supposed to
read the /etc/passwd entry inside of the container but since the
container is not mounted, the c.State.Mountpoint == "".  Podman
incorrectly attempts to read /etc/passwd on the host, and fails if the
specified user is not in the hosts /etc/passwd.

This PR mounts the storage container, if it was not mounted so the read
succeeds.

### DIFF
--- a/libpod/kube.go
+++ b/libpod/kube.go
@@ -676,8 +676,18 @@ func generateKubeSecurityContext(c *Container) (*v1.SecurityContext, error) {
 			return nil, errors.Wrapf(err, "unable to sync container during YAML generation")
 		}
 
+		mountpoint := c.state.Mountpoint
+		if mountpoint == "" {
+			var err error
+			mountpoint, err = c.mount()
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to mount %s mountpoint", c.ID())
+			}
+			defer c.unmount(false)
+		}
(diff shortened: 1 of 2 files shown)


libpod/kube.go
```diff
@@ -676,7 +676,16 @@ func generateKubeSecurityContext(c *Container) (*v1.SecurityContext, error) {
 			return nil, errors.Wrapf(err, "unable to sync container during YAML generation")
 		}
 
+		if c.state.Mountpoint == "" {
+			mountPoint, err := c.mount()
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to mount %s mountpoint", c.ID())
+			}
+			defer c.unmount(false)
+			c.state.Mountpoint = mountPoint
```
mheon, line 685: Don't do this - could interfere with state elsewhere. We should only do this when we do a full mount as part of starting the container.

rhatdan:

saschagrunert (approved):
LGTM

vrothberg (commented):
/lgtm

Nice fix!

PR #9651, opened 2021-03-07
[NO TESTS NEEDED] podman-remote build does not support volumes

rhatdan:
Remove --volume option from podman-remote since it is
not supported, also add information to podman-build man page
indicating options not supported over remote connections.

### DIFF
--- a/docs/source/markdown/podman-build.1.md
+++ b/docs/source/markdown/podman-build.1.md
@@ -86,7 +86,7 @@ resulting image's configuration.
 #### **--cache-from**
 
 Images to utilize as potential cache sources. Podman does not currently support
-caching so this is a NOOP.
+caching so this is a NOOP. (This option is not available with the remote Podman client)
 
 #### **--cap-add**=*CAP\_xxx*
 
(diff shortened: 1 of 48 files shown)


docs/source/markdown/podman-build.1.md
```diff
@@ -86,7 +86,7 @@ resulting image's configuration.
 #### **--cache-from**
 
 Images to utilize as potential cache sources. Podman does not currently support
-caching so this is a NOOP.
+caching so this is a NOOP. (Not available for remote commands)
```
mheon, line 89: Suggest "This option is not available with the remote Podman client"

rhatdan: We have this comment all over the podman-run and podman-create man page, so I will make the change everywhere.  Want to get @TomSweeneyRedHat agreement though so we don't bikeshed.

TomSweeneyRedHat: @mheon's suggestion SGTM

jwhonce:
LGTM

TomSweeneyRedHat:
Hmm, I could have sworn I LGTM'd this earlier, I'll do so now.
LGTM

PR #9653, opened 2021-03-07
exec/start api: setup streams to avoid invalid argument exception

TristanCacqueray:


TristanCacqueray:
Oops sorry, this is not necessary when the exec config attach the output.

PR #9665, opened 2021-03-08
[NO TESTS NEEDED] Compat API: Avoid trying to create volumes if they already exist

psgreco:
Trying to mkdir when the destination exists, can create problems like it was experienced with unix sockets, avoid trying to mkdir if we already know the volume exists

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -439,6 +439,10 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 		if !filepath.IsAbs(vol) {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


mheon:
Can you add a signoff (`git commit --amend -s`) and re-push? You'll probably also need to add `[NO TESTS NEEDED]` to the title.

psgreco:
> Can you add a signoff (`git commit --amend -s`) and re-push? You'll probably also need to add `[NO TESTS NEEDED]` to the title.

Done

rhatdan:
Could you rebase and push again.  It seems to still want tests.

mheon:
LGTM

mheon:

Issue #9656, opened 2021-03-08
Document CONTAINERS_CONF and CONTAINERS_STORAGE_CONF


martinetd:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

I was looking for a way to override the /etc/containers config files if required on a read-only system and stumbled upon CONTAINERS_CONF and CONTAINERS_STORAGE_CONF

Are these safe to use? (that is, can I rely on these working and not going away with minor version updates -- I see they're used in some tests so probably safe enough)
If so would it be worth documenting these?

Thanks!

vrothberg:
Thanks for reaching out.  Yes, they are safe to use with Podman but I am not sure why they're not documented. @rhatdan may know.  It may have simply slipped through when adding the env vars.

Luap99:
It is documented in `containers.conf(5)` https://github.com/containers/common/blob/master/docs/containers.conf.5.md#files

martinetd:
uh, sorry :| Now I need to know why `man -K CONTAINERS_CONF` didn't find this, I can confirm the installed man page has it.. But that's another problem.

Thanks!

vrothberg:
I don't consider it documented - if only half of the request since CONTAINERS_STORAGE_CONF is not mentioned.  I think that the Podman docs should mention it.

martinetd:
There's one line about `CONTAINERS_STORAGE_CONF` there as well:
If the CONTAINERS_STORAGE_CONF path environment variable is set, this path is used for the storage.conf file rather than the default.  This is primarily used for testing.

vrothberg:
I was a bit vague, apologies.  What I want is that it's mentioned in `man podman`.  It would be more approachable.

martinetd:
Ah, sure :)

btw if anyone is curious, to answer myself: man -K searches for literals in the man page sources, so in this case I should have use `man -K 'CONTAINERS\_CONF'`... I had also grepped through the repo but go only has go sources from vendored dependencies and not a clone of the whole other repos -- either way I agree it can be made more easily discoverable, thanks for the follow up!

martinetd:
@rhatdan (thanks for the PR)
While we're here, there is also REGISTRIES_CONFIG_PATH that I cannot find documented anywhere (this time not even in containers.conf(5) man page) -- it might already be too late to rename it to CONTAINERS_REGISTRIES_CONF to be homogeneous? but might as well be documented at the same time

[...]

PR #9668, opened 2021-03-08
Document CONTAINERS_CONF/CONTAINERS_STORAGE_CONF Env variables

rhatdan:
Fixes: https://github.com/containers/podman/issues/9656

### DIFF
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -279,6 +279,8 @@ Distributions ship the `/usr/share/containers/containers.conf` file with their d
 
 Podman uses builtin defaults if no containers.conf file is found.
 
+If the **CONTAINERS_CONF** environment variable is set, then its value is used for the containers.conf file rather than the default.
+
 **mounts.conf** (`/usr/share/containers/mounts.conf`)
 
     The mounts.conf file specifies volume mount directories that are automatically mounted inside containers when executing the `podman run` or `podman start` commands. Administrators can override the defaults file by creating `/etc/containers/mounts.conf`.
@@ -295,6 +297,8 @@ When Podman runs in rootless mode, the file `$HOME/.config/containers/mounts.con
 
     Non root users of Podman can create the `$HOME/.config/containers/registries.conf` file to be used instead of the system defaults.
 
+    If the **CONTAINERS_REGISTRIES_CONF** environment variable is set, then its value is used for the registries.conf file rather than the default.
+
 **storage.conf** (`/etc/containers/storage.conf`, `$HOME/.config/containers/storage.conf`)
 
     storage.conf is the storage configuration file for all tools using containers/storage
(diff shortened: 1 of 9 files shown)


rhatdan:

ashley-cui (commented):
LGTM

vrothberg (commented):
Can we drop the "This is primarily used for testing." notes?  I imagine this could confuse users whether it's safe to use these environment variables or not.

martinetd:
Repeating comment I wrote in #9656 as per suggestion (but this could very well be a different PR)

> While we're here, there is also REGISTRIES_CONFIG_PATH that I cannot find documented anywhere (this time not even in containers.conf(5) man page) -- it might already be too late to rename it to CONTAINERS_REGISTRIES_CONF to be homogeneous? but might as well be documented at the same time

rhatdan:
Since REGISTRIES_CONFIG_PATH was never used I am changing it to use the CONTANERS_REGISTISTRIES_CONF and documenting it.

docs/source/markdown/podman.1.md
```diff
@@ -303,8 +307,10 @@ When Podman runs in rootless mode, the file `$HOME/.config/containers/mounts.con
 
     When Podman runs in rootless mode, the file `$HOME/.config/containers/storage.conf` is used instead of the system defaults.
 
+    If the **CONTAINERS_STORAGE_CONF** environment variable is set, the its value is used for the storage.conf file rather than the default.
+
 ## Rootless mode
-Podman can also be used as non-root user.  When podman runs in rootless mode, a user namespace is automatically created for the user, defined in /etc/subuid and /etc/subgid.
+Podman can also be used as non-root user. When podman runs in rootless mode, a user namespace is automatically created for the user, defined in /etc/subuid and /etc/subgid.
```
TomSweeneyRedHat, line 313:
```suggestion
Podman can also be used as non-root user. When Podman runs in rootless mode, a user namespace is automatically created for the user, defined in /etc/subuid and /etc/subgid.
```

pkg/registries/registries.go
```diff
@@ -24,7 +24,7 @@ var userRegistriesFile = filepath.Join(os.Getenv("HOME"), ".config/containers/re
 // FIXME: This should be centralized in a global SystemContext initializer inherited throughout the code,
 // not haphazardly called throughout the way it is being called now.
 func SystemRegistriesConfPath() string {
-	if envOverride := os.Getenv("REGISTRIES_CONFIG_PATH"); len(envOverride) > 0 {
```
vrothberg, line 27: That's a breaking change. Can we continue supporting `REGISTRIES_CONFIG_PATH` but error out if both are set? I'd sleep better.

vrothberg (commented):
I think we need to continue supporting the old variable. We don't know who may rely on it already.

pkg/registries/registries.go
rhatdan: It was never documented and was buried in the code.

vrothberg: Yet it was used in many tests for a very long time.  I would feel very uncomfortable to merge as I have no idea who might already rely on it.

vrothberg (commented):
LGTM, thank you!

pkg/registries/registries.go
martinetd: Rather than warn if both are set, emit a warning if the old one is set recommending to use new name and saying it'll be deprecated in a future version -- then remove it later™ ?

(That said I'm fine with anything, but I agree people like me could have found about it and started using it already even if it's not documented... Although I do believe you can't complain if undocumented features break, which is why I had asked in the first place :D )

PR #9669, opened 2021-03-08
apiv2 tests: add helpers to start/stop a local registry

edsantiago:
...and a rudimentary set of /auth tests for PR#9589 (disabled).

This simply adds a new start_registry() helper function that
allocates a random unused port, pulls a registry image, creates
a local certificate + random username + random password, and
fires everything up. Since none of this is (yet) used in CI,
this is very low risk.

The only infinitessimally-risky change is using a dedicated
subdirectory of $WORKDIR (instead of $WORKDIR itself) as
the podman root. This fixes a dumb oversight on my part:
the workdir has grown to be used for much more than just
podman root; this change removes clutter and makes it
easier for humans to debug in cases of problems.

### DIFF
--- a/test/apiv2/60-auth.at
+++ b/test/apiv2/60-auth.at
@@ -1,0 +1,29 @@
+# -*- sh -*-
+#
+# registry-related tests
+#
+
+start_registry
+
+# FIXME FIXME FIXME: remove the 'if false' for use with PR 9589
+if false; then
+
+# FIXME FIXME: please forgive the horrible POST params format; I have an
+#              upcoming PR which should fix that.
+
+# Test with wrong password. Confirm bad status and appropriate error message
+t POST /v1.40/auth "\"username\":\"${REGISTRY_USERNAME}\",\"password\":\"WrOnGPassWord\",\"serveraddress\":\"localhost:$REGISTRY_PORT/\"" \
+  400 \
+  .Status~'.* invalid username/password'
+
+# Test with the right password. Confirm status message and reasonable token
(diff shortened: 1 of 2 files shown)


troyready:
I'm having trouble running this (in the context of testing another PR #9589 ). Seems like it doesn't find any tests:

```
$ ./test/apiv2/test-apiv2 auth
Generating a RSA private key
.........................................++++
.........................................++++
writing new private key to '/tmp/test-apiv2.tmp.XUkD2D/registry/auth/domain.key'
-----
1..0

```
(trying it on debian testing aka bullseye, for context)

It wasn't immediately apparent to me what's causing it. If it could be some local system issue, happy to look into it after this gets merged but wanted to mention it now (while it's fresh on your mind) in case you might know offhand what would cause it.

edsantiago:
@troyready hand-edit `test/apiv2/60-auth.at`, remove the `if` and `fi`.

The confusion is my fault: I was happy with this PR, so I force-pushed it with the disabling conditional, with the goal of getting this merged soon so you can rebase on it.

edsantiago:
Failure is a flake, a new one: [timeout in podman build](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6514698670571520/html/sys-remote-ubuntu-2004-root-host.log.html). No possible way that it's my fault, so I've restarted.

edsantiago:
@containers/podman-maintainers PTAL. This is currently almost-entirely a NOP: it adds functionality intended for testing #9589 and potentially other auth-related endpoints. There's one minor functional change, for maintainability, but I've triple-convinced myself that it's safe.

CI is flaking, but it's my bedtime and I'm not hanging out to wait for it.

TomSweeneyRedHat:
As usual, really nice work @edsantiago 
LGTM
and happy green test buttons.

Issue #9670, opened 2021-03-08
API: Copy/Archive API Implementation


pzl:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

I noticed that the compat API has [Archive PUT](https://github.com/containers/podman/blob/master/pkg/api/handlers/compat/containers_archive.go#L23) (aka copy?) implemented.  but libpod API [does not](https://github.com/containers/podman/blob/master/pkg/api/handlers/libpod/copy.go#L10).

Is there anything this is waiting for? Can it use a similar call (`ContainerCopyFromArchive`) or `AddArtifact` or anything similar?

I was trying to use `github.com/containers/libpod/v2/pkg/bindings` to put a file into a container (where mounting was not preferable) and saw that I was not able to via the libpod API (of which `pkg/bindings` seems to be a client) but I seem to able able to via the compat API.

If this is a simple fix of mostly copy-over implementation, happy to submit a PR for that. But I didn't know what the thinking was surrounding `libpod/copy.go` here

vrothberg:
I opened https://github.com/containers/podman/pull/9671 to clean up the dead code.

PR #9671, opened 2021-03-09
rm pkg/api/handlers/libpod/copy.go

vrothberg:
Remove the file since it only contains dead code.  The archive endpoints
are shared between the libpod and the compat API and both use the compat
package.

[NO TESTS NEEDED] since we're removing dead code.

Fixes: #9670


vrothberg:

saschagrunert (approved):
LGTM

Issue #9657, opened 2021-03-08
Setting --storage-opt overrides unrelated settings


martinetd:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Overriding storage options on command-line seems to override more than what I would have expected, e.g. in rootless mode this doesn't work:
```
podman --storage-opt additionalimagestore=/tmp/podman_store image list
```

I had to set the overlay mount program again like this:
```
podman --storage-opt additionalimagestore=/tmp/podman_store --storage-opt overlay.mount_program=/usr/bin/fuse-overlayfs image list
```

Note that I have no /etc/containers/storage.conf nor $HOME/.config/containers/storage.conf, so the mount_program comes from some default value when not setting any storage-opt.

**Steps to reproduce the issue:**

1. podman --root /tmp/podman_store pull docker.io/alpine

2. podman --storage-opt additionalimagestore=/tmp/podman_store image list

**Describe the results you received:**

`Error: kernel does not support overlay fs: 'overlay' is not supported over btrfs at "/tmp/podmanroot2/overlay": backing file system is unsupported for this graph driver`

**Describe the results you expected:**

```
REPOSITORY                TAG     IMAGE ID      CREATED      SIZE     R/O
docker.io/library/alpine  latest  28f6e2705743  2 weeks ago  5.88 MB  true
```

**Additional information you deem important (e.g. issue happens only occasionally):**

Might be related to #9650

**Output of `podman version`:**

```
Version:      3.1.0-dev
API Version:  3.1.0-dev
Go Version:   go1.15.8
Git Commit:   b7c00f2cc03499d5d385a7aa7e8cd35d0ab994d7
Built:        Mon Mar  8 09:56:35 2021
OS/Arch:      linux/amd64
```

vrothberg:
@rhatdan @giuseppe PTAL

giuseppe:
not sure what is the best way to handle this.

If we don't override all the storage opts, how would it be possible to drop opts that were already specified in the config file?

At least now it is not possible to set an empty value for, e.g. mount_program:

```
$ podman --storage-opt overlay.mount_program="" --storage-opt additionalimagestore=/tmp/podman_store image list
Error: overlay: can't stat program : stat : no such file or directory
```

martinetd:
Good point, hadn't thought of testing dropping settings.

[...]

PR #9676, opened 2021-03-09
options: append CLI graph driver options

giuseppe:
if --storage-opt are specified on the CLI append them after what is
specified in the configuration files instead of overriding it.

Closes: https://github.com/containers/podman/issues/9657

[NO TESTS NEEDED]


mheon:
LGTM

rhatdan:
What happens if I want to override one?

Issue #9672, opened 2021-03-09
podman load with invalid input has 0 (success) exit code


martinetd:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature
(not quite sure there :D I swear I'm not just trying to use podman randomly like a fuzzer...!)

**Description**

if podman load failed to parse the input I believe it should error out not just in messages but also in its exit code
(mostly-unrelated: it's a shame saved images can't be signed somehow, or am I missing something?)

**Steps to reproduce the issue:**

```
$ echo garbled input | podman load
  open /var/tmp/podman877531368/manifest.json: not a directory
  open /var/tmp/podman877531368/index.json: not a directory
Loaded image(s): 

$ echo $?
0
```

I'm -very bad- at go, but pulling a debugger out it looks like a simple assignment problem in libpod/runtime_img.go (err goes out of scope after the if, so `saveErr = err` only ever saves "nil")?
With this diff:
```
diff --git a/libpod/runtime_img.go b/libpod/runtime_img.go
index 6e1105b9e3f5..76cd03dc0ebd 100644
--- a/libpod/runtime_img.go
+++ b/libpod/runtime_img.go
@@ -316,7 +316,8 @@ func() (types.ImageReference, error) {
        } {
                src, err := referenceFn()
                if err == nil && src != nil {
-                       if newImages, err := r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer); err == nil {
+                       newImages, err := r.ImageRuntime().LoadFromArchiveReference(ctx, src, signaturePolicy, writer)
+                       if err == nil {
                                return getImageNames(newImages), nil
                        }
                        saveErr = err
```
I get:
```
$ ./bin/podman load -i foo
  open foo/manifest.json: not a directory
  open foo/index.json: not a directory
Error: error pulling image: unable to pull oci:foo:: 1 error occurred:
	* Error initializing source oci:foo:: open foo/index.json: not a directory

$ echo $?
125
```

The error only displaying the last handler's error isn't necessarily the most obvious for users but at least an error bubbles up as it looks like the code originally intended. Happy to submit that as a PR if that makes sense (I'm not sure if this project has a policy to always refer to an issue for each PR? Some people don't like PRs that hadn't been discussed first)

**Output of `podman version`:**

```
Version:      3.1.0-dev
API Version:  3.1.0-dev
Go Version:   go1.15.8
[...]

PR #9677, opened 2021-03-09
podman load: fix error handling

vrothberg:
Make sure to properly return loading errors and to set the exit code
accordingly.

Fixes: #9672

### DIFF
--- a/test/system/120-load.bats
+++ b/test/system/120-load.bats
@@ -26,6 +26,13 @@ verify_iid_and_name() {
     is "$new_img_name" "$1"   "Name & tag of restored image"
 }
 
+@test "podman load invalid file" {
+    # Regression test for #9672 to make sure invalid input yields errors.
+    invalid=$PODMAN_TMPDIR/invalid
+    echo "I am an invalid file and should cause a podman-load error" > $invalid
+    run_podman 125 load -i $invalid
+}
+
 @test "podman save to pipe and load" {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

test/system/120-load.bats
```diff
@@ -26,6 +26,13 @@ verify_iid_and_name() {
     is "$new_img_name" "$1"   "Name & tag of restored image"
 }
 
+@test "podman load invalid file" {
+    # Regression test for #9672 to make sure invalid input yields errors.
+    invalid=$PODMAN_TMPDIR/invalid
+    echo "I am an invalid file and should cause a podman-load error" > $invalid
+    run_podman 125 load -i $invalid
```
edsantiago, line 33: Wish I'd seen this earlier. In future, I encourage you to check the error message. I'll open a PR to add that.

edsantiago: >Shouldn't you be doing yoga on a mountain peak? :)

What makes you think I'm not? :wink: 

Ugly or not, the reason for error-message checking is to save yourself from typos or syntax errors:
```bash
    run_podman load -i $imvalid      <--- misspelled variable name, yields "flag needs an argument"
    run_podman load -i $nonexistent_path  <--- yields "no such file or directory"
```
...or other mistakes that throw an error for other reasons. (None of those are likely in this case, the code looks good, but an error-message check will greatly reassure future maintainers)

Issue #9675, opened 2021-03-09
Get a 500 over the rest api during deletion of unkown containers


codemaker219:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

With upgrade to podman 3.0.1 we will get a 500 if we try to delete forcefully a container that does not exists. Before it was a 204 (what also is not the best response code)

**Steps to reproduce the issue:**

1. Start the rest api `podman system service tcp:0.0.0.0:8099 -t0`

2. Try to delete a container forcefully that does not exists `curl -s -vvv -X DELETE http://127.0.0.1:8099/v1.40.0/libpod/containers/unkown?force=true`

**Describe the results you received:**
A 500 responsecode

**Describe the results you expected:**
A 404 responsecode

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.14.12
Built:        Tue Mar  2 21:31:11 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug:**

```
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.26-4.el8.5.2.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.26, commit: 6ce9825c91d32f9890be05932c8e5a39c6a8ae2a'
  cpus: 8
  distribution:
    distribution: '"centos"'
    version: "8"
  eventLogger: journald
  hostname: localhost.localdomain
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 4.18.0-240.15.1.el8_3.x86_64
  linkmode: dynamic
  memFree: 9583484928
  memTotal: 16601899008
  ociRuntime:
    name: crun
    package: crun-0.18-3.el8.5.2.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.18
      commit: 808420efe3dc2b44d6db9f1a3fac8361dde42a95
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #9681, opened 2021-03-09
Removing a non existing container API should return 404

rhatdan:
Currently we were overwrapping error returned from removal
of a non existing container.

$ podman rm bogus -f
Error: failed to evict container: "": failed to find container "bogus" in state: no container with name or ID bogus found: no such container

Removal of wraps gets us to.

./bin/podman rm bogus -f
Error: no container with name or ID "bogus" found: no such container

Finally also added quotes around container name to help make it standout
when you get an error, currently it gets lost in the error.


rhatdan:
Fixes: https://github.com/containers/podman/issues/9675

TomSweeneyRedHat:
All kinds of red unhappiness in tests @rhatdan

vrothberg (commented):
/lgtm

Issue #9650, opened 2021-03-07
[Volume Plugins] Specifying options resets Volume Driver to local.


thephoenixofthevoid:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

If using volume plugin with specified options, Driver (--driver) argument is completely ignored.

**Steps to reproduce the issue:**

1. Forked `docker-volume-sshfs` and changed paths to socket (it has been hardcoded).

2. Added a record in `containers.conf`

3. Started a plugin. Then tried to create a volume without options and specifying them. 

**Describe the results you received:**

For `podman volume create --driver testplugin` as well as using API endpoint `/libpod/volumes/create` with `{ Driver: "testplugin" }` got:
```
$ ./podman-volume-ssh 
INFO[0000] listening on /tmp/podman-plugins-test.sock   
2021/03/07 14:59:13 Entering go-plugins-helpers getPath
ERRO[0054] volume 784484bf997f98d0b04a07525eb12560da1740f6139cc0e7463132427496937a not found 
2021/03/07 14:59:13 Entering go-plugins-helpers createPath
ERRO[0054] 'sshcmd' option required         
```
API Endpoint response:
```
{
  cause: "'sshcmd' option required",
  message: `error creating volume "9874028d86d0e6404b0dcda74cd46da441dcc0b2a6650bb0f8fef56da8bc0eb4" in plugin testplugin: error on /VolumeDriver.Create on volume 9874028d86d0e6404b0dcda74cd46da441dcc0b2a6650bb0f8fef56da8bc0eb4 in volume plugin testplugin: 'sshcmd' option required`,
  response: 500
}
```
Which is correct actually. But.

For `podman volume create --driver testplugin --opt sshcmd=<whatever>` as well as using API endpoint `/libpod/volumes/create` with `{ Driver: "testplugin", Options: { sshcmd: <whatever> } }` got:
```
<NOTHING REACHES PLUGIN ENDPOINT>
```
and the response:
```
{
  cause: 'invalid argument',
  message: 'error running volume create option: unrecognized volume option "sshcmd" is not supported with local driver: invalid argument',
  response: 500
}
```

**Describe the results you expected:**

Expected to create a volume that corresponds to a remote sshfs.

**Additional information you deem important (e.g. issue happens only occasionally):**

This might be related: https://github.com/containers/podman/blob/d92b94677cc816a1e157802836195430b731d015/cmd/podman/volumes/create.go#L77

[...]

PR #9685, opened 2021-03-10
[NO TESTS NEEDED] Fix (experimental) Volume Plugin in rootless mode

thephoenixofthevoid:
According to https://github.com/containers/podman/issues/9650 there are few issues when using volume plugins with rootless containers.

The reason behind these is local-driver-specific code being executed for all cases. This is a minimal fix to address these issues.

Please also note that:
1. In fact, there is a need to clearly separate local-driver-specific code from plugin-related, because there are little to none logic that should be executed in both cases.
2. The removed check is actually already correctly implemented here
https://github.com/containers/podman/blob/d92b94677cc816a1e157802836195430b731d015/libpod/runtime_volume_linux.go#L66-L77
3. These kind of checks is actually implementation specific, so it seems natural to keep them close to the implementation.

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -69,7 +69,7 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		BuildArgs              string `schema:"buildargs"`
 		CacheFrom              string `schema:"cachefrom"`
 		Compression            uint64 `schema:"compression"`
-		ConfigureNetwork       int64  `schema:"networkmode"`
+		ConfigureNetwork       string `schema:"networkmode"`
 		CpuPeriod              uint64 `schema:"cpuperiod"`  // nolint
 		CpuQuota               int64  `schema:"cpuquota"`   // nolint
 		CpuSetCpus             string `schema:"cpusetcpus"` // nolint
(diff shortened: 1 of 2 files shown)


mheon:
Git validation is complaining that your commits don't have a signoff - `git commit --amend -s` should add one.

I think it will also complain that you don't have tests. You can add a test (a slightly-changed version (option added) of https://github.com/containers/podman/blob/master/test/e2e/volume_plugin_test.go#L53-L85 would work) or add `[NO TESTS NEEDED]` to your PR title before you re-push.

/approve

thephoenixofthevoid:
At this point no new tests are needed, because there is no reliable and stable volume plugin yet. Using unstable volume plugin is rather harmful, as such test is much more likely to break due to unrelated issue, than due to regression of this fix.

mheon:
Sorry - it's now complaining about your commit subject length:
`  - FAIL - commit subject exceeds 90 characters`

edsantiago:
The second line of a git commit must be empty. @thephoenixofthevoid please insert an empty line between `Fix rootless...` and `In a case of`.

thephoenixofthevoid:
Oh, no, it failed again.

mheon:
Ah, this one's simple. Just need to rebase against latest master - it's a known test failure due to timezone change that we've fixed already.

mheon:
@thephoenixofthevoid Did you mean to close?

thephoenixofthevoid:
No. It happened as a result of rebase automatically.

mheon:
Hm. Git is saying that there are 0 new commits - I think your new commit got lost somehow?

thephoenixofthevoid:
Luckly, I have patch files, and will recreate everything.

thephoenixofthevoid:
https://github.com/containers/podman/pull/9808

thephoenixofthevoid:
That was completely unexpected, that github closes PR if by chance/by mistake commits get excluded from branch.

PR #9686, opened 2021-03-10
apiv2 tests: finally fix POST as originally intended

edsantiago:
When I originally wrote this code I had no idea what POST
would look like so I did a sloppy job, deferring making it
usable. Now that we have some real-world examples in place,
I have a better understanding of what params look like and
how to make tests more readable/maintainable. (Deferring isn't
always bad: one of my early ideas was to separate params using
commas; that would've been a disaster because some JSON values,
such as arrays, include commas).

This commit implements a better way of dealing with POST:

  * The main concept is still 'key=value'
    * When value is a JSON object (dictionary, array), it
      can be quoted.
    * Multiple params are simply separated by spaces.
      The 3-digit HTTP code is a prominent, readable separator
      between POST params and expected results. The parsing
      code is a little uglier, but test developers need
      never see that. The important thing is that writing
      tests is now easier.
  * POST params can be empty (this removes the need for a
    useless '')

I snuck in one unrelated change: one of the newly-added
tests, .NetworkSettings, was failing when run rootless
(which is how I test on my setup). I made it conditional.

### DIFF
--- a/test/apiv2/20-containers.at
+++ b/test/apiv2/20-containers.at
@@ -31,12 +31,16 @@ t GET libpod/containers/json?all=true 200 \
   .[0].ExitCode=0 \
   .[0].IsInfra=false
 
-# Test compat API for Network Settings
+# Test compat API for Network Settings (.Network is N/A when rootless)
+network_expect=
+if root; then
+    network_expect='.[0].NetworkSettings.Networks.podman.NetworkID=podman'
+fi
 t GET /containers/json?all=true 200 \
   length=1 \
   .[0].Id~[0-9a-f]\\{64\\} \
   .[0].Image=$IMAGE \
-  .[0].NetworkSettings.Networks.podman.NetworkID=podman
+  $network_expect
 
 # Make sure `limit` works.
 t GET libpod/containers/json?limit=1 200 \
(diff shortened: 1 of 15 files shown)


edsantiago:
**Review Suggestions**: I realize this is a mess to review. If I may make a couple of suggestions:
* Look at the first few ''-only (quote-only) changes. If you're happy with those, train your eye to ignore them because they're all the same and they're all mostly noise.
* For the real changes, the ones with real params, ignore the old value (gobbledygook clutter) and just focus on the new value: is it readable? Is it correct?
* Give a once-over to the parsing code in `test-apiv2` but if you're not a bash person, just assume that if the tests pass, the code is good. (I know. That's never good advice).

test/apiv2/20-containers.at
```diff
@@ -31,12 +31,16 @@ t GET libpod/containers/json?all=true 200 \
...
+network_expect=
+if root; then
+    network_expect='.[0].NetworkSettings.Networks.podman.NetworkID=podman'
+fi
 t GET /containers/json?all=true 200 \
   length=1 \
   .[0].Id~[0-9a-f]\\{64\\} \
   .[0].Image=$IMAGE \
-  .[0].NetworkSettings.Networks.podman.NetworkID=podman
+  $network_expect
```
edsantiago, line 43: This is the one non-param-related change, the one I snuck in. All it does is, if running rootless, skip the `.NetworkSettings` check but run the others.

test/apiv2/README.md
```diff
@@ -52,9 +52,14 @@ Notes:
...
 
-* When method is POST, the argument after the endpoint must be a series
-of POST arguments in the form 'key=value', separated by commas. `t` will
-convert those to JSON form for passing to the server.
+* When method is POST, the argument(s) after the endpoint may be a series
+of POST parameters in the form 'key=value', separated by spaces:
+     t POST myentrypoint 200                                 ! no params
+     t POST myentrypoint id=$id 200                          ! just one
+     t POST myentrypoint id=$id filter='{"foo":"bar"}' 200   ! two, with json
+     t POST myentrypoint name=$name badparam='["foo","bar"]' 500  ! etc...
```
vrothberg, line 60: Would `foo="value with space"` be counted as one argument?

edsantiago: Oh, good question. I was careful to do arg processing the "correct" way, so yes, it should work, but I don't have any explicit tests for it. I think it might be time to add an internal test suite for that helper. Unfortunately, my sneak-in-some-work-time for today has run out! I'll get to that when I can. Thanks for pointing it out.

vrothberg: Thank you, LGTM :)

vrothberg (commented):
LGTM

PR #9689, opened 2021-03-10
add restart-policy to container filters & --filter to podman start

boaz0:
This patch adds two things:
1) restart-policy to the container filters - you can filter containers by their restart policy
2) `--filter` to the `podman start` command - you can filter what containers to start

by adding these two you can start only containers that has a restart-policy of "on-failure" for example.

//cc @rhatdan


cmd/podman/containers/start.go
```diff
@@ -57,6 +58,8 @@ func startFlags(cmd *cobra.Command) {
 	flags.BoolVarP(&startOptions.Interactive, "interactive", "i", false, "Keep STDIN open even if not attached")
 	flags.BoolVar(&startOptions.SigProxy, "sig-proxy", false, "Proxy received signals to the process (default true if attaching, false otherwise)")
 
+	flags.StringSliceVarP(&filters, "filter", "f", []string{}, "Filter output based on conditions given")
```
rhatdan, line 61: Remove blank line above.

cmd/podman/containers/start.go
```diff
@@ -108,7 +111,35 @@ func start(cmd *cobra.Command, args []string) error {
...
-	responses, err := registry.ContainerEngine().ContainerStart(registry.GetContext(), args, startOptions)
+	for _, f := range filters {
+		split := strings.SplitN(f, "=", 2)
+		if len(split) == 1 {
+			return errors.Errorf("invalid filter %q", f)
+		}
+		listOpts.Filters[split[0]] = append(listOpts.Filters[split[0]], split[1])
+	}
+	listOpts.All = true
+	ctrs, listErr := registry.ContainerEngine().ContainerList(registry.GetContext(), listOpts)
```
rhatdan, line 122: just use `err`

cmd/podman/containers/start.go
```diff
@@ -108,7 +111,35 @@ func start(cmd *cobra.Command, args []string) error {
...
+	for _, ctr := range ctrs {
+		for _, nameOrId := range args {
+			if nameOrId == ctr.ID {
+				containers = append(containers, nameOrId)
+				continue
+			}
+			for _, name := range ctr.Names {
+				if name == nameOrId {
+					containers = append(containers, nameOrId)
+					continue
```
rhatdan, line 136: Should this be a break?

rhatdan (approved):

TomSweeneyRedHat:
All kinds of test unhappiness @boaz0

boaz0:
Yeah, that's odd. I will look at this today. Thanks.

test/system/710-start.bats
```diff
@@ -0,0 +1,30 @@
...
+
+function setup() {
+    basic_setup
+    run_podman run -d $IMAGE /bin/true
+    cid_none="$output"
+    run_podman run -d --restart=always $IMAGE /bin/true
+    cid_always="$output"
+}
+
+function teardown() {
```
edsantiago, line 13: Please add `podman rm $cid_none $cid_always` here; this will reduce test-log clutter.

test/system/710-start.bats
```diff
@@ -0,0 +1,30 @@
...
+    basic_teardown
+}
+
+@test "podman start --filter - start only containers that match the filter" {
+    run_podman start --filter restart-policy=always $cid_none $cid_always
+    is "$output" "$cid_always"
```
edsantiago, line 19: As a courtesy to someone reading a test log, could you add a third argument here (and line 22 and 27) with a short description of what this error means? E.g., `"CID of restart-policy=always container"`

test/system/710-start.bats
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bats   -*- bats -*-
```
edsantiago, line 1: Would you consider renaming to `045-start.bats`? The purpose of the naming scheme isn't "keep appending to the end", it's "what is important?" In this case, `start` is much much more important than networking, completion, play, and so on. `45` seems like a good place for it, right before `stop`. The fact that there wasn't already a `start` file is purely an oversight on my part.

edsantiago (commented):
A few suggestions. Thanks for this work!

boaz0:
ping @TomSweeneyRedHat I am not sure the CI failure is related to my changes.

cmd/podman/containers/start.go
```diff
@@ -108,7 +111,35 @@ func start(cmd *cobra.Command, args []string) error {
 		startOptions.Stdout = os.Stdout
 	}
 
-	responses, err := registry.ContainerEngine().ContainerStart(registry.GetContext(), args, startOptions)
+	for _, f := range filters {
```
rhatdan, line 114: This should only be done if a filters were specified.

boaz0: Where is that start protocol?

rhatdan: Yes?

rhatdan:
@boaz0 Please rebase, I would love to get this in before we release podman 3.2

boaz0:
No problem! @rhatdan

TomSweeneyRedHat:
LGTM
And

Happy green test buttons!

rhatdan:
Great job @boaz0 Thanks

/lgtm

boaz0:
This is me right now:

Thank you so much guys!!! I couldn't have done it without your help and support.

PR #9697, opened 2021-03-11
sdnotify tests: try real hard to kill socat processes

edsantiago:
podman gating tests are hanging in the new Fedora CI setup;
long and tedious investigation suggests that 'socat' processes
are being left unkilled, which then causes BATS to hang when
it (presumably) runs a final 'wait' in its end cleanup.

The two principal changes are to exec socat in a subshell
with fd3 closed, and to pkill its child processes before
killing the process itself. I don't know if both are needed.
The pkill definitely is; the exec may just be superstition.
Since I've wasted more than a day of PTO time on this, I'm
okay with a little superstition. What I do know is that with
these two changes, my reproducer fails to reproduce in over
one hour of trying (normally it fails within 5 minutes).

### DIFF
--- a/test/system/260-sdnotify.bats
+++ b/test/system/260-sdnotify.bats
@@ -42,14 +42,22 @@ function _start_socat() {
     _SOCAT_LOG="$PODMAN_TMPDIR/socat.log"
 
     rm -f $_SOCAT_LOG
-    socat unix-recvfrom:"$NOTIFY_SOCKET",fork \
-          system:"(cat;echo) >> $_SOCAT_LOG" &
+    # Execute in subshell so we can close fd3 (which BATS uses).
+    # This is a superstitious ritual to try to avoid leaving processes behind,
+    # and thus prevent CI hangs.
+    (exec socat unix-recvfrom:"$NOTIFY_SOCKET",fork \
+          system:"(cat;echo) >> $_SOCAT_LOG" 3>&-) &
     _SOCAT_PID=$!
 }
 
 # Stop the socat background process and clean up logs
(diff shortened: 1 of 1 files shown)


mheon:
Sure, LGTM

rhatdan:
LGTM

edsantiago:
Update: changed `pkill` to `pkill ... || true`, because (sigh) f33 and ubuntu do not leave behind a set of stray socat processes. Only rawhide does:
```
  [parent socat process has been killed]
  ├─socat,358891 unix-recvfrom:/tmp/podman_bats.OBYLSk/conmon.sock,fork system:(cat;echo) >> /tmp/podman_bats.OBYLSk/socat.log
  │   └─socat,358893 unix-recvfrom:/tmp/podman_bats.OBYLSk/conmon.sock,fork system:(cat;echo) >> /tmp/podman_bats.OBYLSk/socat.log
  │       └─sh,358895 -c (cat;echo) >> /tmp/podman_bats.OBYLSk/socat.log
  │           └─sh,358897 -c (cat;echo) >> /tmp/podman_bats.OBYLSk/socat.log
  │               └─cat,358900
```
Anyhow, `pkill -P pid`, on f33 and ubuntu, will error out because `pid` has no children. Only rawhide has child socats. I have no idea what's going on.

on f33: socat-1.7.4.1-1.fc33 5.10.19-200.fc33
on rawhide: socat-1.7.4.1-2.fc34 5.12.0-0.rc2.165.fc35

TomSweeneyRedHat:
LGTM

PR #9710, opened 2021-03-14
Network prune filters for http api (compat and libpod)

jmguzik:
/kind feature
Prune filters are missing not implemented in HTTP API. This PR tackles the topic in compat and libpod API.

### DIFF
--- a/libpod/network/netconflist.go
+++ b/libpod/network/netconflist.go
@@ -5,8 +5,11 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/containernetworking/cni/libcni"
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/podman/v3/pkg/network"
 	"github.com/containers/podman/v3/pkg/util"
 	"github.com/pkg/errors"
@@ -222,24 +225,7 @@ func IfPassesFilter(netconf *libcni.NetworkConfigList, filters map[string][]stri
 
 		case "label":
 			// matches all labels
-			labels := GetNetworkLabels(netconf)
-		outer:
-			for _, filterValue := range filterValues {
-				filterArray := strings.SplitN(filterValue, "=", 2)
-				filterKey := filterArray[0]
-				if len(filterArray) > 1 {
-					filterValue = filterArray[1]
-				} else {
-					filterValue = ""
-				}
-				for labelKey, labelValue := range labels {
-					if labelKey == filterKey && (filterValue == "" || labelValue == filterValue) {
-						result = true
-						continue outer
-					}
-				}
-				result = false
-			}
+			result = matchPruneLabelFilters(netconf, filterValues)
 
 		case "driver":
 			// matches only for the DefaultNetworkDriver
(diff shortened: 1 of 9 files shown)


libpod/network/netconflist.go
```diff
@@ -267,3 +254,82 @@ func IfPassesFilter(netconf *libcni.NetworkConfigList, filters map[string][]stri
...
+func getCreatedTimestamp(config *config.Config, netconf *libcni.NetworkConfigList) (*time.Time, error) {
+	networkConfigPath, err := GetCNIConfigPathByNameOrID(config, netconf.Name)
+	if err != nil {
+		return nil, err
+	}
+	f, err := os.Stat(networkConfigPath)
+	if err != nil {
+		return nil, err
+	}
+	stat := f.Sys().(*syscall.Stat_t)
```
jmguzik, line 332: @containers/podman-maintainers Windows and OS X cross builds are failing because of syscall.Stat_t. As far as I know, there is no way to get timestamp created from data structures as in the case of the containers. What would be your recommended way to pass cross builds? Making arch-specific code here?

jmguzik:
It seems there is timezone issue (maybe due to time change in USA):
```
           Expected
               <string>: '04 EDT'
           to contain substring
               <string>: EST
         
```

libpod/network/netconflist.go
Luap99: I will open a PR to fix this. We should not use `libpod/network` client side.

jmguzik: @Luap99 Ok, thanks. Let me know and then I will do a rebase.

Luap99: You can try with #9716

jmguzik:
sys remote ubuntu-2004 root host fails because of some podman build issues. I think it is unrelated. I would be grateful if somebody with rights would re-run.

libpod/network/netconflist.go
jmguzik: done

libpod/network/netconflist.go
```diff
@@ -268,3 +255,82 @@ func IfPassesFilter(netconf *libcni.NetworkConfigList, filters map[string][]stri
 	}
 	return result, nil
 }
+
+// ComputeUntilTimestamp extracts unitil timestamp from filters
+func ComputeUntilTimestamp(filter string, filterValues []string) (time.Time, error) {
```
Luap99, line 260: Move this function to `pkg/util`

libpod/network/netconflist.go
```diff
@@ -268,3 +255,82 @@ func IfPassesFilter(netconf *libcni.NetworkConfigList, filters map[string][]stri
...
+	return time.Unix(seconds, nanoseconds), nil
+}
+
+// IfPassesPruneFilter filters NetworkListReport and returns true if the prune filter match the given config
+func IfPassesPruneFilter(config *config.Config, netconf *libcni.NetworkConfigList, f map[string][]string) (bool, error) {
+	for key, filterValues := range f {
+		switch strings.ToLower(key) {
+		case "label":
+			return matchPruneLabelFilters(netconf, filterValues), nil
+		case "until":
```
Luap99, line 282: I think it would be good to also add the until filter for network ls

Luap99 (commented):
Make sure to update the api docs in `pkg/api/server/register_networks.go`.
Also squash your commits please.

libpod/network/netconflist.go
jmguzik: @Luap99 please be certain about this location. In the coming days I will add a unification PR which will remove code duplication. Maybe it is worth to create a new dir that could be shared between libpod and compat endpoints? Please comment on this.

libpod/network/netconflist.go
jmguzik: It is api extension then. Docker http does not allow this:
https://docs.docker.com/engine/api/v1.41/#operation/NetworkList

jmguzik: Let me know if you really want it.

libpod/network/netconflist.go
Luap99: I thought about moving it to this file pkg/util/utils.go. I do not want `libpod/network` to be imported by non network related packages like you do in pkg/domain/filters/containers.go.

Removing code duplication between compat and libpod would definitely be great.

jmguzik: pkg/util/utils.go? or pkg/util/filter_helpers.go? There will be quite a few functions there coming (depending on my free time of course to do this).

libpod/network/netconflist.go
Luap99: I don't thinks its needed when docker doesn't have it and no one asked for it.

libpod/network/netconflist.go
Luap99: If you want to add more a new file would is good.

jmguzik: moved to pkg/util/filters.go this file will grow once this one gets merged.

jmguzik:
> Make sure to update the api docs in `pkg/api/server/register_networks.go`.
> Also squash your commits please.

Commits squashed and API docs updated.

Luap99 (approved):
LGTM

Luap99:

PR #9711, opened 2021-03-14
Fix for volumes prune in http compat api when using filters

jmguzik:
/kind bug

1. Docker: scenario => when the user makes mistake when using filters, then no negative consequences
```
$ docker volume create v1
v1
$ docker volume create v2
v2
$ curl -gG -XPOST --unix-socket /var/run/docker.sock  http://localhost/v1.41/volumes/prune?filters=dssd21 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    67  100    67    0     0  67000      0 --:--:-- --:--:-- --:--:-- 67000
{
  "message": "invalid character 'd' looking for beginning of value"
}


```

2. Podman (compat API): scenario => mistake, then there are negative consequences
```
$ bin/podman volume create v1
v1
$ bin/podman volume create v2
v2
$ curl -gG -XPOST --unix-socket /run/user/1000/podman/podman.sock  http://d/v3.0.0/volumes/prune?filters=dssd21 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    50  100    50    0     0   2173      0 --:--:-- --:--:-- --:--:--  2173
{
  "VolumesDeleted": [
    "v1",
    "v2"
  ],
  "SpaceReclaimed": 0
}
$ bin/podman volume create v1bis
v1bis
$ bin/podman volume create v2bis
v2bis
$ curl -gG -XPOST --unix-socket /run/user/1000/podman/podman.sock  http://d/v3.0.0/volumes/prune?filters={garba1ge | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    56  100    56    0     0   2666      0 --:--:-- --:--:-- --:--:--  2666
{
  "VolumesDeleted": [
    "v1bis",
    "v2bis"
  ],
  "SpaceReclaimed": 0
}

```
3. After fix:
```
$ curl -gG -XPOST --unix-socket /run/user/1000/podman/podman.sock  http://d/v3.0.0/volumes/prune?filters=dssd21 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   155  100   155    0     0   151k      0 --:--:-- --:--:-- --:--:--  151k
{
  "cause": "invalid character 'd' looking for beginning of value",
  "message": "Decode(): invalid character 'd' looking for beginning of value",
  "response": 500
}

```

Generally, this code:
```
	query := struct {
		Filters map[string][]string `schema:"filters"`
	}{
		// override any golang type defaults
	}

	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
```
Is very broken in similar situations. Sometimes it does not have negative consequences (for GET list, just inconsistency) but for prune user instead of 500 user receives 200 with success... having all things wiped out.

I can do similar PRs for compat API (images, containers, etc). Libpod is also affected, but since there are no compat constraints there, maybe this is by design?

I also added fixes for HTTP codes, according to this: https://docs.docker.com/engine/api/v1.41/ . There are many 400 (StatusBadRequest) while there is no example of 400 in docker API. Another story is that current flow never gets to if conditions in questions since `if err := decoder.Decode(&query, r.URL.Query()); err != nil....` is broken. Probably it returns nil instead of error when analyzing garbage filter input.

### DIFF
--- a/test/apiv2/30-volumes.at
+++ b/test/apiv2/30-volumes.at
@@ -86,6 +86,10 @@ t DELETE libpod/volumes/foo1 404 \
     .message~.* \
     .response=404
 
+# Prune volumes - bad filter input
+t POST volumes/prune?filters='garb1age}' 500 \
+    .cause="invalid character 'g' looking for beginning of value"
+
 ## Prune volumes with label matching 'testlabel1=testonly'
 t POST libpod/volumes/prune?filters='{"label":["testlabel1=testonly"]}' 200
 t GET libpod/volumes/json?filters='{"label":["testlabel1=testonly"]}' 200 length=0
--- a/pkg/api/handlers/compat/volumes.go
+++ b/pkg/api/handlers/compat/volumes.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/containers/podman/v3/libpod"
@@ -31,7 +32,7 @@ func ListVolumes(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
-		utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+		utils.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError,
 			errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
 		return
 	}
@@ -40,7 +41,7 @@ func ListVolumes(w http.ResponseWriter, r *http.Request) {
 	// happily parse them for us.
 	for filter := range query.Filters {
 		if filter == "opts" {
-			utils.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest,
+			utils.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError,
 				errors.Errorf("unsupported libpod filters passed to docker endpoint"))
 			return
 		}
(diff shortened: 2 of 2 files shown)


jmguzik:
It seems there is a timezone issue (maybe due to a time change in the USA):
```
           Expected
               <string>: '05 EDT'
           to contain substring
               <string>: EST
         
```

test/apiv2/30-volumes.at
```diff
@@ -86,6 +86,9 @@ t DELETE libpod/volumes/foo1 404 \
     .message~.* \
     .response=404
 
+# Prune volumes - bad filter input
+t POST volumes/prune?filters='garb1age}' 500
+
```
edsantiago, line 91: Friendly suggestion:
```diff
-t POST volumes/prune?filters='garb1age}' 500
+t POST volumes/prune?filters='garb1age}' 500 \
+  .cause="invalid character 'g' looking for beginning of value"
```
It is a good habit when checking for failures to confirm that the failure is happening for the reason you expect, and not, e.g., for a syntax error or other unrelated reason.

edsantiago (commented):
One suggestion. And, you need to rebase onto current master in order to pick up a fix for a bug that's breaking CI.

test/apiv2/30-volumes.at
jmguzik: Applied friendly suggestion and rebased. Thanks!

TomSweeneyRedHat:
LGTM

rhatdan:
Thanks @jmguzik 
/lgtm
/approve

PR #9716, opened 2021-03-15
Do not leak libpod package into the remote client

Luap99:
1. Split libpod/network package 
The `libpod/network` package should only be used on the backend and not the
client. The client used this package only for two functions so move them
into a new `pkg/network` package.
2. Do not leak libpod package into the remote client
Some packages used by the remote client imported the libpod package.
This is not wanted because it adds unnecessary bloat to the client and
also causes problems with platform specific code(linux only), see #9710.

The solution is to move the used functions/variables into extra packages
which do not import libpod.

This change shrinks the remote client size more than 6MB compared to the
current master.

[NO TESTS NEEDED]
I have no idea how to test this properly but with #9710 the cross
compile should fail.


vrothberg (commented):
LGTM

mheon:
Damn fine work, LGTM

Luap99:
Tests are green. Merge me

Issue #9379, opened 2021-02-15
build: cannot tar container entries: 0 errors occurred


edsantiago:
Another unfortunate flake that happens in multiple tests, so I can't easily analyze it:
```
Running: podman-remote [options] build --layers=true -t test --file /tmp/podman_test768134587/Dockerfile /tmp/podman_test768134587
time="2021-02-14T09:34:12-06:00" level=error msg="cannot tar container entries [/tmp/podman_test768134587/Dockerfile /tmp/podman_test768134587] error: 0 errors occurred:\n\t\n\n"
Error: 0 errors occurred:
...
Exit code is 125, expected 0 (or maybe the other way around, I can never understand ginkgo diagnostics)
```
This is happening all over the place, and is now our number-one flake. It is not easily characterized other than "podman-remote build". It seems to have started on February 8. I recommend treating this as very high priority due to the number of flakes yesterday.

## Podman rmi [It] podman rmi with cached images
* gce_instance:fedora : int remote fedora-33 root host
  * PR #9348
    * [02-14 10:45](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6389174224814080/html/int-remote-fedora-33-root-host.log.html#t--podman-rmi-with-cached-images--1)
  * PR #9284
    * [02-09 14:13](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5693393412030464/html/int-remote-fedora-33-root-host.log.html#t--podman-rmi-with-cached-images--1)

## Podman rmi [It] podman image rm - concurrent with shared layers
* gce_instance:fedora : int remote fedora-33 root host
  * PR #9348
    * [02-14 10:45](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6389174224814080/html/int-remote-fedora-33-root-host.log.html#t--podman-image-rm---concurrent-with-shared-layers--1)
    * [02-14 10:45](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6389174224814080/html/int-remote-fedora-33-root-host.log.html#t--podman-image-rm---concurrent-with-shared-layers--2)
[...]

PR #9717, opened 2021-03-15
[NO TESTS NEEDED] pkg/bindings/images.Build(): fix a race condition in error reporting

nalind:
In nTar(), don't return the error value when the goroutine that's populating the error value can continue running long after nTar() returns.  Instead, wrap the Close() method of the pipe that we're returning in a function that collects those errors, along with any error we get from closing the pipe, and returns them from Close() wrapper.
In Build(), if the Close() method returns an error, at least log it.

Should help with https://github.com/containers/podman/issues/9379.


vrothberg (commented):
LGTM
Thank you!

edsantiago:
Thank you @nalind! You may need to force-push to rerun CI; looks like Cirrus can't pick up my title change.

nalind:
@edsantiago no problem, looks like I needed to rebase anyway.

TomSweeneyRedHat:
LGTM

nalind:
Rebased.

Issue #9567, opened 2021-03-02
CI: Ubuntu 2004 is slow


edsantiago:
I've updated my test-timings script to have it interpret system test results. Ubuntu is slow, especially 2004:

(All of these are remote, from #9556. It takes just a few seconds to get results for any PR, so just ask if you want others)

 |                                                                        | fedora<br/>33  | ubuntu<br/>2004  | ubuntu<br/>2010 |
 | ---------------------------------------------------------------------    | -----:    | -----:    | -----: |
 | 20 podman run - basic tests                                            |     3.00  |   **5.00** |   **5.00** |
 | 33 podman run --log-driver                                             |     4.00  |   **9.00** |     5.00  |
 | 48 podman ps --filter                                                  |     2.00  |   **7.00** |   **4.00** |
 | 51 podman stop --all                                                   |     5.00  |   **9.00** |     6.00  |
 | 53 podman stop - can trap signal                                       |     5.00  |   **9.00** |     6.00  |
 | 65 podman cp file from host to container                               |    19.00  |  **76.00** |    26.00  |
 | 66 podman cp file from container to host                               |    15.00  |  **95.00** |  **25.00** |
 | 67 podman cp dir from host to container                                |    17.00  |  **142.00** |  **35.00** |
 | 68 podman cp dir from container to host                                |    14.00  |  **107.00** |  **28.00** |
 | 69 podman cp file from host to container volume                        |     1.00  |  **34.00** |   **6.00** |
 | 70 podman cp file from host to container mount                         |     2.00  |  **29.00** |   **6.00** |
 | 71 podman cp * - wildcard copy multiple files from container to host   |     1.00  |  **18.00** |   **5.00** |
 | 72 podman cp - will not recognize symlink pointing into host space     |     1.00  |  **14.00** |   **5.00** |
 | 73 podman cp - will not expand globs in host space (#3829)             |     1.00  |  **13.00** |   **4.00** |
 | 74 podman cp - will not expand wildcard                                |     1.00  |  **13.00** |   **5.00** |
 | 75 podman cp into container: weird symlink expansion                   |     3.00  |  **28.00** |  **13.00** |
 | 76 podman cp into a subdirectory matching GraphRoot                    |     3.00  |  **22.00** |   **8.00** |
 | 77 podman cp from stdin to container                                   |    13.00  |  **41.00** |  **24.00** |
 | 78 podman cp from container to stdout                                  |    12.00  |  **38.00** |  **21.00** |
[...]

PR #9719, opened 2021-03-15
System test cleanup

edsantiago:
- cp test: clean up stray image

- build test: add workaround for #9567 (ultra-slow ubuntu).
  We're seeing CI flakes (timeouts) due to ubuntu 2004 being
  absurdly slow. Workaround: double our timeout on one specific
  test when ubuntu + remote.

- build test: clean up new copy-from test (from #9275).
  The test was copy-pasted from buildah system tests, without
  really adapting for podman environment (e.g. it was using
  images that we don't use here, and would cause pulls, which
  will cause flakes). Rewrite test so it references only $IMAGE,
  remove some confusing/unnecessary stuff, selectively run
  parts of it even when rootless or remote, and add a
  test to confirm that copy-from succeeded.

- load test: add error-message test to new load-invalid (#9672).
  Basically, make sure the command fails for the right reason.

- play test (kube): use $IMAGE, not alpine; and add pause-image
  cleanup to teardown()

- apiv2 mounts test: add a maintainability comment in a tricky
  section of code; and tighten up the mount point test.

### DIFF
--- a/test/system/065-cp.bats
+++ b/test/system/065-cp.bats
@@ -370,6 +370,7 @@ load helpers
     is "${lines[0]}" "${randomcontent[0]}" "eval symlink - created container"
     is "${lines[1]}" "${randomcontent[1]}" "eval symlink - created container"
     run_podman rm -f cpcontainer
+    run_podman rmi $cpimage
 }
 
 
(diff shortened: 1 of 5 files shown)


edsantiago:
Rebased to incorporate time zone fix from #9716

TomSweeneyRedHat:
LGTM

test/system/065-cp.bats
```diff
@@ -370,6 +370,7 @@ load helpers
     is "${lines[0]}" "${randomcontent[0]}" "eval symlink - created container"
     is "${lines[1]}" "${randomcontent[1]}" "eval symlink - created container"
     run_podman rm -f cpcontainer
+    run_podman rmi $cpimage
```
vrothberg, line 373: Thanks for cleaning up :)

Issue #8798, opened 2020-12-21
Socket truncated with long user names using default $XDG_RUNTIME_DIR


fool65c:
**Is this a BUG REPORT or FEATURE REQUEST?**

/kind bug

Socket truncated with long user names using default $XDG_RUNTIME_DIR

**Steps to reproduce the issue:**

1. have a long username such as `11111111`

2. run `podman run -it ubuntu`

**Describe the results you received:**
podman fails to attach to the running container

```
DEBU[0003] Received: 20102
INFO[0003] Got Conmon PID as 20092
DEBU[0003] Created container 7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e in OCI runtime
DEBU[0003] Attaching to container 7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e
DEBU[0003] connecting to socket /var/run/user/11111111/libpod/tmp/socket/7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e/a
DEBU[0003] ExitCode msg: "failed to connect to container's attach socket: /var/run/user/11111111/libpod/tmp/socket/7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e/a: dial unixpacket /var/run/user/11111111/libpod/tmp/socket/7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e/a: connect: no such file or directory"
Error: failed to connect to container's attach socket: /var/run/user/11111111/libpod/tmp/socket/7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e/a: dial unixpacket /var/run/user/11111111/libpod/tmp/socket/7e19533164c776b5965dfb17891efd6d2eda50059190d00e5e2765c60323d38e/a: connect: no such file or directory
```

**Describe the results you expected:**
user able to attach to the container

**Additional information you deem important (e.g. issue happens only occasionally):**

This appears to be because of a system limit `addr.sun_path` because the socket is truncated:
https://github.com/containers/podman/blob/master/libpod/oci_attach_linux.go#L187

Based on `addr.sun_path` here:
https://github.com/containers/podman/blob/b68b6f334d4a83de35d68cc359365b348e72ad6b/libpod/oci_attach_linux_cgo.go#L6

Is this a system setting we can change?

[...]

PR #9722, opened 2021-03-16
[2.2] oci: use /proc/self/fd/FD to open unix socket

giuseppe:
requested here: https://bugzilla.redhat.com/show_bug.cgi?id=1939247

instead of opening directly the UNIX socket path, grab a reference to
it through a O_PATH file descriptor and use the fixed size string
"/proc/self/fd/%d" to open the UNIX socket.  In this way it won't hit
the 108 chars length limit.

Closes: https://github.com/containers/podman/issues/8798

(cherry picked from commit fdbc278868fa0e9ea470d77857a7905811d2faa0)


vrothberg (approved):
LGTM

PR #9699, opened 2021-03-11
Delete all containers and pods between tests

jwhonce:
New tearDown() deletes all pods and containers between tests

Signed-off-by: Jhon Honce <jhonce@redhat.com>

PR #9732, opened 2021-03-17
[v3.0.1-rhel] podman cp: ignore EPERMs in rootless mode

vrothberg:
Ignore permission errors when copying from a rootless container.
TTY devices inside rootless containers are owned by the host's
root user which is "nobody" inside the container's user namespace
rendering us unable to even read them.

Enable the integration test which was temporarily disabled for rootless
users.

This one slipped through in the backports.  Last bit needed to fix https://bugzilla.redhat.com/show_bug.cgi?id=1936927.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.0
-	github.com/containers/buildah v1.19.7
+	github.com/containers/buildah v1.19.8
 	github.com/containers/common v0.33.4
 	github.com/containers/conmon v2.0.20+incompatible
 	github.com/containers/image/v5 v5.10.5
--- a/test/e2e/containers_conf_test.go
+++ b/test/e2e/containers_conf_test.go
@@ -311,7 +311,7 @@ var _ = Describe("Podman run", func() {
 		session = podmanTest.Podman([]string{"run", ALPINE, "date", "+'%H %Z'"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		Expect(session.OutputToString()).To(ContainSubstring("EST"))
+		Expect(session.OutputToString()).To(Or(ContainSubstring("EST"), ContainSubstring("EDT")))
 
 		// Umask
 		session = podmanTest.Podman([]string{"run", "--rm", ALPINE, "sh", "-c", "umask"})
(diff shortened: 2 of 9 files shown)


vrothberg:
Had to add a third commit to fix the containers.conf timezone e2e test.  I do not know _why_ that worked before since New York yields EDT, not EST.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.0
-	github.com/containers/buildah v1.19.7
```
TomSweeneyRedHat, line 13: This will give us a higher version of Buildah vendored into Podman than we have available on RHEL (v1.19.6).

test/e2e/containers_conf_test.go
```diff
@@ -311,7 +311,7 @@ var _ = Describe("Podman run", func() {
 		session = podmanTest.Podman([]string{"run", ALPINE, "date", "+'%H %Z'"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		Expect(session.OutputToString()).To(ContainSubstring("EST"))
+		Expect(session.OutputToString()).To(ContainSubstring("EDT"))
```
TomSweeneyRedHat, line 314: Won't this change/fail in the fall when we go back to EST?  Seems like a check that needs to be modified....

edsantiago: DO NOT MERGE THIS. Use 8de560703 instead.

TomSweeneyRedHat:
@rhatdan @mheon @baude Please put this at the top of your list for reviews this morning.

go.mod
vrothberg: Not sure what that implies.  We need `.7` to fix the issue.

test/e2e/containers_conf_test.go
vrothberg: Done, thanks!

mheon:
I can live with Buildah patch release not matching. If it was a minor release I'd be more concerned.

LGTM, but CI is red even after your patch.

vrothberg:
> LGTM, but CI is red even after your patch.

The APIv2 tests flake. I want to wait with restarting until the other tests turn green.

edsantiago:
Maybe this would be a good time to cherry-pick #9699

go.mod
TomSweeneyRedHat: Just concerned that if you run Buildah you could get different behavior than running `podman build`.  I'm probably over-fretting.

TomSweeneyRedHat:
LGTM
and all happy green test buttons
@Jnovy FYI
@mheon, want to press the shiny merge button?

PR #9740, opened 2021-03-17
System test: Add tests for podman auto-update

ypu:
Add some tests for podman auto-update and also fix one bug in basic_setup().

### DIFF
--- a/test/system/255-auto-update.bats
+++ b/test/system/255-auto-update.bats
@@ -1,0 +1,279 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# Tests for automatically update images for containerized services
+#
+
+load helpers
+
+UNIT_DIR="/usr/lib/systemd/system"
+SNAME_FILE=$BATS_TMPDIR/services
+
+function setup() {
+    skip_if_remote "systemd tests are meaningless over remote"
+    skip_if_rootless
+
+    basic_setup
+}
+
+function teardown() {
+    while read line; do
+        if [[ "$line" =~ "podman-auto-update" ]]; then
+            echo "Stop timer: $line.timer"
+            systemctl stop $line.timer
+            systemctl disable $line.timer
+        else
+            systemctl stop $line
+        fi
+        rm -f $UNIT_DIR/$line.{service,timer}
+    done < $SNAME_FILE
+
+    rm -f $SNAME_FILE
+    run_podman ? rmi quay.io/libpod/alpine:latest
+    run_podman ? rmi quay.io/libpod/alpine_nginx:latest
+    run_podman ? rmi quay.io/libpod/localtest:latest
+    basic_teardown
+}
+
+# This functions is used for handle the basic step in auto-update related
+# tests. Including following steps:
+#   1. Generate a random container name and echo it to output.
+#   2. Tag the fake image before test
+#   3. Start a container with io.containers.autoupdate
+#   4. Generate the service file from the container
+#   5. Remove the origin container
+#   6. Start the container from service
+function generate_service() {
+    target_img_basename=$1
+    autoupdate=$2
+
+    # Please keep variable name for cname and ori_image. The
+    # scripts will use them directly in following tests.
+    cname=c_$(random_string)
+    target_img="quay.io/libpod/$target_img_basename:latest"
+    run_podman tag $IMAGE $target_img
+    if [[ -n "$autoupdate" ]]; then
+        label="--label io.containers.autoupdate=$autoupdate"
+    else
+        label=""
+    fi
+    run_podman run -d --name $cname $label $target_img top -d 120
+
+    run_podman generate systemd --new $cname
+    echo "$output" > "$UNIT_DIR/container-$cname.service"
+    echo "container-$cname" >> $SNAME_FILE
+    run_podman rm -f $cname
+
+    systemctl daemon-reload
+    systemctl start container-$cname
+    systemctl status container-$cname
+
+    run_podman inspect --format "{{.Image}}" $cname
+    ori_image=$output
+}
+
+function _wait_service_ready() {
+    local sname=$1
+
+    local timeout=6
+    while [[ $timeout -gt 1 ]]; do
+        run systemctl is-active $sname
+        if [[ $output == "active" ]]; then
+            return
+        fi
+        sleep 1
+        let timeout=$timeout-1
+    done
+
+    # Print serivce status as debug information before failed the case
+    systemctl status $sname
+    die "Timed out waiting for $sname to start"
+}
+
+function _confirm_update() {
+    local sname=$1
+
+    local timeout=6
+    last_log=""
+    while [[ $timeout -gt 1 ]]; do
+        run journalctl -u $sname -n 10
+        if [[ "$output" == "$last_log" ]]; then
+            return
+        fi
+        last_log=$output
+        sleep 1
+        let timeout=$timeout-1
+    done
+
+    die "Timed out waiting for $sname to update"
+}
+
+# This test can fail in dev. environment because of SELinux.
+# quick fix: chcon -t container_runtime_exec_t ./bin/podman
+@test "podman auto-update - label io.containers.autoupdate=image" {
+    run_podman images
+    generate_service alpine image
+
+    _wait_service_ready container-$cname.service
+    run_podman ps -a
+    run_podman auto-update
+    is "$output" "Trying to pull.*" "Image is updated."
+    run_podman ps -a
+    _confirm_update container-$cname.service
+    run_podman inspect --format "{{.Image}}" $cname
+    [[ "$output" != "$ori_image" ]]
+}
+
+@test "podman auto-update - label io.containers.autoupdate=disabled" {
+    generate_service alpine disabled
+
+    _wait_service_ready container-$cname.service
+    run_podman ps -a
+    run_podman auto-update
+    is "$output" "" "Image is not updated with disabled."
+    run_podman ps -a
+    _confirm_update container-$cname.service
+    run_podman inspect --format "{{.Image}}" $cname
+    is "$output" "$ori_image" "Image hash should not changed."
+}
+
+@test "podman auto-update - label io.containers.autoupdate=fakevalue" {
+    fakevalue=$(random_string)
+    generate_service alpine $fakevalue
+
+    _wait_service_ready container-$cname.service
+    run_podman ps -a
+    run_podman ? auto-update
+    is "$output" ".*invalid auto-update policy.*" "invalid policy setup"
+    run_podman ps -a
+    _confirm_update container-$cname.service
+    run_podman inspect --format "{{.Image}}" $cname
+    is "$output" "$ori_image" "Image hash should not changed."
+}
+
+@test "podman auto-update - label io.containers.autoupdate=local" {
+    generate_service localtest local
+    podman commit --change CMD=/bin/bash $cname quay.io/libpod/localtest:latest
+
+    _wait_service_ready container-$cname.service
+    run_podman ps -a
+    run_podman auto-update
+    run_podman ps -a
+    _confirm_update container-$cname.service
+    run_podman inspect --format "{{.Image}}" $cname
+    [[ "$output" != "$ori_image" ]]
+}
+
+@test "podman auto-update with multiple services" {
+    fakevalue=$(random_string)
+    run_podman inspect --format "{{.Id}}" $IMAGE
+    img_id="$output"
+    cnames=()
+    local -A expect_update
+    local -A will_update=([image]=1 [registry]=1 [local]=1)
+
+    for auto_update in image registry "" disabled "''" $fakevalue local
+    do
+        img_base="alpine"
+        if [[ $auto_update == "registry" ]]; then
+            img_base="alpine_nginx"
+        elif [[ $auto_update == "local" ]]; then
+            img_base="localtest"
+        fi
+        generate_service $img_base $auto_update
+        cnames+=($cname)
+        if [[ $auto_update == "local" ]]; then
+            local_cname=$cname
+        fi
+        if [[ -n "$auto_update" && -n "${will_update[$auto_update]}" ]]; then
+            expect_update[$cname]=1
+        fi
+    done
+
+    # Only check the last service is started. Previous services should already actived.
+    _wait_service_ready container-$cname.service
+    run_podman commit --change CMD=/bin/bash $local_cname quay.io/libpod/localtest:latest
+    run_podman ? auto-update
+    update_log=$output
+    for cname in "${cnames[@]}"; do
+        _confirm_update container-$cname.service
+    done
+    count=0
+    while read line; do
+       if [[ "$line" =~ "Trying to pull" ]]; then
+           ((count+=1))
+       fi
+    done <<< "$update_log"
+    is "$update_log" ".*invalid auto-update policy.*" "invalid policy setup"
+    is "$update_log" ".*1 error occurred.*" "invalid policy setup"
+    is "$count" "2" "There are two images being updated from registry."
+
+    for cname in "${!expect_update[@]}"; do
+
+        is "$update_log" ".*$cname.*" "container with auto-update policy image updated"
+    done
+
+    for cname in "${cnames[@]}"; do
+        run_podman inspect --format "{{.Image}}" $cname
+        if [[ -n "${expect_update[$cname]}" ]]; then
+            [[ "$output" != "$img_id" ]]
+        else
+            is "$output" "$img_id" "Image should not be changed."
+        fi
+    done
+}
+
+@test "podman auto-update using systemd" {
+    generate_service alpine image
+
+    cat >$UNIT_DIR/podman-auto-update-$cname.timer <<EOF
+[Unit]
+Description=Podman auto-update testing timer
+
+[Timer]
+OnCalendar=*-*-* *:*:0/2
+Persistent=true
+
+[Install]
+WantedBy=timers.target
+EOF
+    cat >$UNIT_DIR/podman-auto-update-$cname.service <<EOF
+[Unit]
+Description=Podman auto-update testing service
+Documentation=man:podman-auto-update(1)
+Wants=network.target
+After=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/podman auto-update
+
+[Install]
+WantedBy=multi-user.target default.target
+EOF
+
+    echo "podman-auto-update-$cname" >> $SNAME_FILE
+    systemctl enable --now podman-auto-update-$cname.timer
+    systemctl list-timers --all
+
+    count=0
+    failed_start=1
+    while [ $count -lt 120 ]; do
+        run journalctl -n 15 -u podman-auto-update-$cname.service
+        if [[ "$output" =~ "Finished Podman auto-update testing service" ]]; then
+            failed_start=0
+            break
+        fi
+        ((count+=1))
+        sleep 1
+    done
+    echo $output
+
+    _confirm_update container-$cname.service
+    run_podman inspect --format "{{.Image}}" $cname
+    if [[ $failed_start == 1 ]]; then
+       die "Failed to get podman auto-update service finished"
+    fi
+    [[ "$output" != "$ori_image" ]]
+}
+
+# vim: filetype=sh


test/system/255-auto-update.bats
```diff
@@ -0,0 +1,212 @@
...
+        die "Error starting systemd unit $SERVICE_NAME, output: $output"
+    fi
+
+    run systemctl status container-$cname
+    if [ $status -ne 0 ]; then
+        die "Non-zero status of systemd unit $SERVICE_NAME, output: $output"
+    fi
+}
+
+@test "podman auto-update - label io.containers.autoupdate=image" {
```
TomSweeneyRedHat, line 48: IDK what's up, but this test isn't able to start up systems in CI.

ParkerVR (commented):
Other than this looks good

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,213 @@
...
+    sleep 2
+    run_podman ps -a
+    run_podman auto-update
+    is "$output" "" "Image is not updated with disabled."
+    run_podman ps -a
+    sleep 2
+    run_podman inspect --format "{{.Image}}" $cname
+    systemctl stop container-$cname
+    rm -f $UNIT_DIR/container-$cname.service
+    is "$output" "$ori_image" "Image hash should not changed."
```
edsantiago, line 84: Please assign a temporary named variable after line 81 to hold `$output`. Testing `$output` so far after `run` is going to cause confusion and possibly unrelated errors if a future maintainer adds an unrelated `run` command in between. Same applies to various other places in the code. 

Even better, create a `teardown()` function and do such cleanup there.

test/system/helpers.bash
```diff
@@ -61,6 +61,11 @@ function basic_setup() {
         fi
     done
 
+    run_podman images --all --format '{{.Repository}}:{{.Tag}}'
```
edsantiago, line 64: Can you describe (in comments, and in your commit message) why you believe this is necessary and/or a bug? I'm afraid I don't understand what this solves.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,213 @@
...
+function generate_service() {
+    fake_img=$1
+    target_img=$2
+    cmd=$3
+    label=$4
+
+    cname=$(random_string)
+    echo $cname
+    run_podman tag $fake_img $target_img
+    if [ "$4" ]; then
```
edsantiago, line 26: You have a named variable (`$label`); please use it (and use explicit `-n`)

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,213 @@
...
+        run_podman run -d --name $cname --label "io.containers.autoupdate=$label" $target_img $cmd
```
edsantiago, line 27: ...and, even better: instead of this duplication, do something like
```bash
if [[ -n $label ]]; then
    label="--label io.containers.autoupdate=$label"
fi
run_podman run -d --name $cname $label $target_img $cmd
```

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,213 @@
...
+    fi
+
+    run_podman generate systemd --new $cname
+    echo "$output" > "$UNIT_DIR/container-$cname.service"
+    run_podman rm -f $cname
+
+    systemctl daemon-reload
+
+    run systemctl start container-$cname
+    if [ $status -ne 0 ]; then
```
edsantiago, line 39: Another reason for `teardown()`. This could leave the system in a broken state.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,212 @@
...
+}
+
+@test "podman auto-update - label io.containers.autoupdate=image" {
+    run generate_service $IMAGE quay.io/libpod/alpine:latest top image
+    cname=${lines[0]}
+    echo $output
+    run_podman inspect --format "{{.Image}}" $cname
+    ori_image=$output
+
+    sleep 2
```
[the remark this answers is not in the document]
edsantiago: I second @ParkerVR's concern. If possible, please use `wait_for_ready` instead.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,213 @@
...
+@test "podman auto-update using systemd" {
+    run generate_service $IMAGE quay.io/libpod/alpine:latest top image
+    cname=${lines[0]}
+    echo $output
+    run_podman inspect --format "{{.Image}}" $cname
+    ori_image=$output
+
+    sleep 2
+    cp $UNIT_DIR/podman-auto-update.timer $UNIT_DIR/podman-auto-update.timer-bk
+    sed -i 's/OnCalendar=.*/OnCalendar=*-*-* *:*:0\/2/g' $UNIT_DIR/podman-auto-update.timer
```
edsantiago, line 183: I really have to encourage you not to do this. It's one thing to add scratch system files; quite another to muck with existing distro-installed and -needed files. I strongly urge you to find a different way to do this, such as creating a `podman-test-randomname.timer` in `/etc/systemd` instead.

edsantiago (changes requested):
This sort of system-file muckery makes me incredibly nervous: please remember that these tests run on developer laptops, not just ad hoc VMs in CI. Please take every possible precaution to not harm a real-world system.

test/system/255-auto-update.bats
ypu: Seems my testing env on Fedora 32 has different systemd set up directory. Already update the code to fix this. Thanks for the comments.

test/system/255-auto-update.bats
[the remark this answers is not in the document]
ypu: There are two types of sleep in this test. One is wait for the service fully up and another is wait for the container back after we update the image. In my test env, sleep 2 makes it stable and not failed because we check some date too soon. This is also used in the 250-systemd.bats tests. And I am afraid that we can not use wait_for_ready function here as that function only designed to check the container output but not suitable for this situation.

test/system/255-auto-update.bats
ypu: OK. May need a global array in the scripts to handle the service name and the cleanup steps. A teardown function may make the code more easy to understand and also make sure the services we start will be cleanup and removed from the test machine. Thanks a lot for the advice.

test/system/helpers.bash
ypu: Here is my commit message for this problem. I met it several times when the tag related tests failed. Didn't put it inside the code because it is a little hard to me to describe it in one sentence.

>  System test: Reset found_needed_image after clean up image if needed
> There are some tests that tag the desired images with new image
> name. And if these tests fail in some steps and don't clean up
> the environment. The tagged image will exist with the same id as
> the desired images. So in the setup steps, the desired images
> will be also removed as we try to remove the tagged image with
> its id. This will fail the tests right behind it. So add a check
> step after the clean up to fix it.

> Signed-off-by: Yiqiao Pu <ypu@redhat.com>

test/system/255-auto-update.bats
ypu: OK. Thanks for your comments. Will update this.

test/system/255-auto-update.bats
ypu: OK. Will set up an timer for testing. Thanks for your advice.

ypu:
Hi @edsantiago 
The code is updated. Can you help to review it again? Thanks a lot.

test/system/helpers.bash
edsantiago: I think I understand the problem you mention. If so, this here is not the right solution. Can you PTAL at #10008 and see if that resolves your problem?

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+
+function setup() {
+    skip_if_remote "systemd tests are meaningless over remote"
+    skip_if_rootless
+
+    basic_setup
+}
+
+function teardown() {
+    while read line; do
```
edsantiago, line 19: This is unnecessary and misleading: `teardown` runs after every `@test`, not once per `.bats` file. From what I see below, `$SNAME_FILE` only gets written to once per test, therefore it can only contain at most one line. A better way to implement this would be:
```bash
setup() {
   ...
   TEST_SERVICE_NAME=...
...
teardown() {
   if [[ -n "$TEST_SERVICE_NAME" ]]; then
      systemctl disable --now $TEST_SERVICE_NAME
      rm -f $UNIT_DIR/$TEST_SERVICE_NAME
      systemctl daemon-reload
   fi
```
...with suitable changes in the tests themselves.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+    basic_teardown
+}
+
+function generate_service() {
+    fake_img=$1
+    target_img=$2
+    cmd=$3
+    label=$4
+
+    cname=$(random_string)
```
edsantiago, line 40: Adding a short static string can be a courtesy to the next person trying to debug a test failure (possibly you yourself):
```
   cname=c_$(random_string)
```
Then if you see `c_sdfsdfsdf` in test output you immediately know "this is the container name". If you just saw `sdfsdf`, it would be harder to know.

And the `echo` below, see my note on line 71

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+    echo $cname
+    run_podman tag $fake_img $target_img
+    if [[ -n "$label" ]]; then
+        label="--label io.containers.autoupdate=$label"
+    fi
```
edsantiago, line 45: Eeek. Please do not reuse variables like that, it is super confusing. Instead:
```bash
   label_opt=
   if [[ -n "$label" ]]; then
      label_opt="--label ..."
   fi
```

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+        die "Non-zero status of systemd unit $SERVICE_NAME, output: $output"
+    fi
+}
+
+# This test can fail in dev. environment because of SELinux.
+# quick fix: chcon -t container_runtime_exec_t ./bin/podman
+@test "podman auto-update - label io.containers.autoupdate=image" {
+    run_podman images
+    run generate_service $IMAGE quay.io/libpod/alpine:latest top image
+    cname=${lines[0]}
```
edsantiago, line 71: Whoa..... this is super dangerous and fragile. It (undocumentedly!) relies on the `echo` on line 41. There is no mention of this dependency anywhere. A future maintainer could inadvertently change the behavior of `generate_service`, and everything would break. A better way to do this would be:
```bash
   cname=$(generate_service ...)
```
And then, in `generate_service()`: (1) clearly state, in a top-level function comment, that the function outputs container name to stdout and nothing else; (2) guarantee that the function does so; (3) add a line comment just above the `echo`, repeating that this is used by the caller.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+        fi
+        rm -f $UNIT_DIR/$line.{service,timer}
+    done < $SNAME_FILE
+
+    rm -f $SNAME_FILE
+    basic_teardown
+}
+
+function generate_service() {
+    fake_img=$1
```
edsantiago, line 35: Useless function parameter: it is always `$IMAGE`. Please remove it.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+    target_img=$2
+    cmd=$3
```
edsantiago, line 37: Useless function parameter: it is always `top`. Please remove it.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+        rm -f $UNIT_DIR/$line.{service,timer}
+    done < $SNAME_FILE
+
+    rm -f $SNAME_FILE
+    basic_teardown
+}
+
+function generate_service() {
+    fake_img=$1
+    target_img=$2
```
edsantiago, line 36: Almost-useless function parameter: it is (almost) always `quay.io/libpod/alpine:latest`. There is only one exception, `alpine_nginx`, and there is no explanation of why that one is different.

If they can always all be the same, please remove this parameter.

If there is an important reason for using `alpine_nginx` in one instance, then (1) add a comment explaining the one-time use of nginx, and (2) please just make the function accept the basename (`alpine`, `alpine_nginx`) and have the function add the registry/tag:
```
    target_img=quay.io/libpod/${target_img_basename}:latest
```

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+    cname_default="${lines[0]}"
+    cnames+=($cname_default)
+    echo $output
+
+    run generate_service $IMAGE quay.io/libpod/alpine:latest top disabled
+    cname_disabled="${lines[0]}"
+    cnames+=($cname_disabled)
+    echo $output
+
+    run generate_service $IMAGE quay.io/libpod/alpine:latest top ""
```
edsantiago, line 149: This is identical to line 139. Is it really necessary?

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+    cname_empty="${lines[0]}"
+    cnames+=($cname_empty)
+    echo $output
+
+    run generate_service $IMAGE quay.io/libpod/alpine:latest top $fakevalue
+    cname_fakevalue="${lines[0]}"
+    cnames+=($cname_fakevalue)
+    echo $output
```
edsantiago, line 157: All these `echo`s are not useful.

Also, you could accomplish the same thing with much less duplication by:
```bash
   for label in image "" disabled $fakevalue; do
      cnames+=$(generate_service alpine $label)
   done
```

edsantiago (changes requested):
Thank you for all your work on this. Unfortunately this cannot merge as it is. I've added some feedback and suggestions. There are still more issues I need to comment on, but I've already spent many hours reviewing this and I need to move on to other work.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,243 @@
...
+
+    sleep 2
+    run_podman ps -a
+    run_podman auto-update
+    is "$output" "Trying to pull.*" "Image is updated."
+    run_podman ps -a
+    sleep 2
+    run_podman inspect --format "{{.Image}}" $cname
+    [[ "$output" != "$ori_image" ]]
+}
```
edsantiago, line 84: Tests should clean up after themselves: please `run_podman rmi $UPDATE_IMAGE` here and in all tests, where `UPDATE_IMAGE` (or variable of your choice) is defined at the top as a way to not hardcode `quay.io/libpod/alpine:latest` everywhere.

test/system/255-auto-update.bats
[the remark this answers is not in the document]
edsantiago: This is another good habit to develop: every time you see `sleep`, you think: "this is something that will flake".

Please try this approach instead:
```bash
function _confirm_update() {
    local cname=$1
    local old_iid=$2

    # Image has already been pulled, so this shouldn't take too long
    local timeout=5
    while [[ $timeout -gt 0 ]]; do
        run_podman '?' inspect --format "{{.Image}}" $cname
        if [[ $status != 0 ]]; then
            if [[ $output =~ "no such object:" ]]; then
                # this is ok, it just means the container is being restarted
                :
            fi
        elif [[ $output != $old_iid ]]; then
            return
        fi
        sleep 1
    done

    die "Timed out waiting for $cname to update; old IID=$old_iid"
}
```

test/system/helpers.bash
edsantiago: @ypu, #10008 has merged. Please drop this part of your commit, and rebase on current master. (A good idea anyway, since this PR is forked from a very old commit). Thank you!

edsantiago:
@ypu ping, I hope you will get back to this PR. It is good work, it just needs a little cleanup to be ready to merge!

ypu:
@edsantiago Yes. Will work on it. Blocked with some errata tasks and just come back from a vacation. Thanks a lot for your advises about how to make the scripts better. Will try to fix it as soon as possible.

test/system/255-auto-update.bats
ypu: Hi @edsantiago there is a test case "podman auto-update with multiple services" which request we start several services once inside one test. And I tried to use global variable, but seems bats has it's own way to handle variables. So the original way for transfer variables between functions is not working. We can not change the var inside a function. So I had to store them in a file.

test/system/255-auto-update.bats
ypu: Yes. I want to test if the command can handle different value for this label, and empty value is one of them.

test/system/255-auto-update.bats
ypu: OK. Will make a loop for this.

ypu: And the line of echo $output will help tester get the output from the functions. So we will get useful information from the test log to analyses the test failed in which step.

ypu:
Hi @edsantiago 
I update the code. Add two wait_for functions for checking the service and image update status. Can you help to check this again? Thanks a lot.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,308 @@
...
+
+@test "podman auto-update - label io.containers.autoupdate=disabled" {
+    run generate_service alpine disabled
+    cname=${lines[0]}
+    echo $output
+    run_podman inspect --format "{{.Image}}" $cname
+    ori_image=$output
```
edsantiago, line 141: There's a lot of duplication of this five-line stanza. Did you know that bash variables aren't scoped? That is, you can define a variable in a function, and the caller will be able to see it. (You can't use `run` though). So if you do the `podman inspect + ori_image=` inside `generate_service`, you can make all these five-line repeats become just one line.

edsantiago (changes requested):
This isn't really ready for review. You have one function that does not do anything; and you need to do some serious rethinking about the multiple-services test because you need to preserve IIDs somehow so your `confirm_update` in line 224 will work. It looks like a good test, you just need to find a way to make it work reliably. Thank you again!

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,308 @@
...
+    local timeout=5
+    last_log=""
+    while [[ $timeout -gt 0 ]]; do
+        if [[ "$output" == "$last_log" ]]; then
+            return
+        fi
+        last_log=$output
+        sleep 1
+        let timeout=$timeout-1
+    done
```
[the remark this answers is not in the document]
ypu: I just check if there is any output from the service log to check if the update complete or not. The service log will record the steps as stop container, start container etc. If the update finished, the service log will not change any more. By this way this function can be used for check the status for services that will not update the image too in our testing. And I will check the image id in the tests separately inside the case.

test/system/255-auto-update.bats
```diff
@@ -0,0 +1,308 @@
...
+
+    run systemctl start container-$cname
+    if [ $status -ne 0 ]; then
+        die "Error starting systemd unit $SERVICE_NAME, output: $output"
+    fi
+
+    run systemctl status container-$cname
+    if [ $status -ne 0 ]; then
+        die "Non-zero status of systemd unit $SERVICE_NAME, output: $output"
+    fi
```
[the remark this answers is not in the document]
ypu: Thanks. Will use the commands directly. Sorry I mess up the command line output from bats and ginkgo. Will correct this kind of mistake in other place too.

And for the status check. There is a possible that the service start but failed as there is something wrong in the container or service file. So we'd better keep it here as a debug message for failed tests.

test/system/255-auto-update.bats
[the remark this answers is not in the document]
edsantiago: Can you please take a closer look? I think you will find that you are not actually setting `$output` anywhere. I don't know what you mean by "service log" but there is no log checking being done here.

test/system/255-auto-update.bats
ypu: Thanks a lot for bring this out. Then I think I can get ride of "echo $cname" inside the function too. Will update the code.

test/system/255-auto-update.bats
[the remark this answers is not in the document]
ypu: Oops. Just remove the command for checking it "run journalctl -u $sname -n 10"  by mistake when I remove some debug echos. Will update it.

ypu:
Hi @edsantiago Can you help to check it again. Thanks a lot for all your help.

test/system/255-auto-update.bats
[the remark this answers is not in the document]
edsantiago: I think you forgot to update this.

edsantiago:
Since you need to resubmit anyway, could I encourage you to clean up the two `...[@] =~` instances? Those are risky and hard to maintain. Using an associative array is one option:

```diff
diff --git a/test/system/255-auto-update.bats b/test/system/255-auto-update.bats
index 42627eb02..392ead137 100644
--- a/test/system/255-auto-update.bats
+++ b/test/system/255-auto-update.bats
@@ -167,8 +167,8 @@ function _confirm_update() {
     run_podman inspect --format "{{.Id}}" $IMAGE
     img_id="$output"
     cnames=()
-    cnames_update=()
-    update_type=("image" "registry" "local")
+    local -A expect_update
+    local -A will_update=([image]=1 [registry]=1 [local]=1)


     for auto_update in image registry "" disabled "''" $fakevalue local
@@ -184,8 +184,8 @@ function _confirm_update() {
         if [[ $auto_update == "local" ]]; then
             local_cname=$cname
         fi
-        if [[ -n "$auto_update" && "${update_type[@]}" =~ "$auto_update" ]]; then
-            cnames_update+=($cname)
+        if [[ -n "$auto_update" && -n "${will_update[$auto_update]}" ]]; then
+            expect_update[$cname]=1
         fi
     done

@@ -207,13 +207,13 @@ function _confirm_update() {
     is "$update_log" ".*1 error occurred.*" "invalid policy setup"
     is "$count" "2" "There are two images being updated from registry."

-    for cname in "${cnames_update[@]}"; do
+    for cname in "${!expect_update[@]}"; do
         is "$update_log" ".*$cname.*" "container with auto-update policy image updated"
     done

     for cname in "${cnames[@]}"; do
         run_podman inspect --format "{{.Image}}" $cname
-        if [[ "${cnames_update[@]}" =~ "$cname" ]]; then
+        if [[ -n "${expect_update[$cname]}" ]]; then
             [[ "$output" != "$img_id" ]]
         else
             is "$output" "$img_id" "Image is not changed."
```

edsantiago (approved):

edsantiago:
Several unnerving flakes, but I have reason to believe it was a quay.io hiccup.

/lgtm

Thanks for your hard work, @ypu!

Issue #8916, opened 2021-01-08
pipe to rootless podman container hangs


someone-somenet-org:
/kind bug

**Description**
running ``echo "asdf" | podman run --rm -it --entrypoint /bin/cat centos:7`` hangs indefinitely.

**Describe the results you expected:**
Expected getting printed "asdf", the cointainer to exit and get back to shell.

**Output of `podman version`:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.14.10
Built:        Tue Dec  8 15:37:43 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.18.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.22-2.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.22, commit: c93fb161acdf210143641c432bd8c448788bb093'
  cpus: 6
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: journald
  hostname: [redacted]
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 10100
      size: 1
    - container_id: 1
      host_id: 1010200000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 10102
      size: 1
    - container_id: 1
      host_id: 1010200000
      size: 65536
  kernel: 5.9.16-100.fc32.x86_64
  linkmode: dynamic
  memFree: 783147008
  memTotal: 8236179456
  ociRuntime:
    name: crun
    package: crun-0.16-3.fc32.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.16
      commit: eb0145e5ad4d8207e84a327248af76663d4e50dd
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    path: /run/user/10102/podman/podman.sock
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.1.8-1.fc32.x86_64
    version: |-
      slirp4netns version 1.1.8
      commit: d361001f495417b880f20329121e3aa431a8f90f
      libslirp: 4.3.1
      SLIRP_CONFIG_VERSION_MAX: 3
      libseccomp: 2.5.0
  swapFree: 8338272256
[...]

PR #9747, opened 2021-03-18
Check if stdin is a term in --interactive --tty mode

rhatdan:
If you are attempting to run a container in interactive mode, and want
a --tty, then there must be a terminal in use.

Docker exits right away when a user specifies to use a --interactive and
--TTY but the stdin is not a tty.

Currently podman will pull the image and then fail much later.

Discussion in : https://github.com/containers/podman/issues/8916

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -17,6 +17,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


edsantiago:
Wow! (re: failing CI). 

There are scads of CI tests that unnecessarily add `-ti`. Those can be fixed - but the prevalence makes me suspect that this is a pattern that will be found in the wild, and hence this PR will break existing users. Proceed with caution.

mheon:
We should make it a logrus.Errorf and not a breaking error for now. We can make it an actual error in Podman 4.0 sometime next year.

cmd/podman/containers/run.go
```diff
@@ -106,6 +107,10 @@ func run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if cliVals.TTY && cliVals.Interactive && !terminal.IsTerminal(int(os.Stdin.Fd())) {
+		return errors.New("the input device is not a TTY")
+	}
+
```
giuseppe, line 113: I am fine with this check, it is the same `tty` does, but it won't help in the case where the output is sent to a pipe like:

`podman run -ti ... | another-command`

Should we have another check for stdout?

rhatdan: Docker does not do anything in this case.

rhatdan:
Switched to logrus.Warnf.

edsantiago:
LGTM, and tests are green!

PR #9748, opened 2021-03-18
add a dependabot config to automate vendoring

vrothberg:
While dependabot has turned out great to automate updating dependencies,
a major painpoint was that we had to manually run `make vendor` for each
and every commit.  It was causing noise.

Adding the config file to `.github/dependabot.yml` will take of also
updating the `./vendor` tree.  `containers/common` is using this config
for a while successfully.

@saschagrunert @edsantiago @rhatdan PTAL


vrothberg:
Let's do this for the other projects as well.

mheon:
Will it also sign its commits? That's another one we noticed

saschagrunert:
> Will it also sign its commits? That's another one we noticed

Yes, this bug should be fixed on their side.

saschagrunert (approved):
LGTM

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
```
TomSweeneyRedHat, line 7: Not that we probably would, but what happens if we get 12 updates? Do the last two get queued up for "by hand" submission?  Or am I misinterpreting this?  If I'm on, I think I'd feel more comfortable at 15 or 20.

TomSweeneyRedHat:
LGTM
but one question for rumination or my enlightenment.

.github/dependabot.yml
vrothberg: That should be scheduled for the next day. @saschagrunert, did you set 10 for a specific reason or was it a gut feeling?

TomSweeneyRedHat: Thanks @saschagrunert .  So if it doesn't create number 11 and 12, is a notification sent out or will it just do those the next day?

PR #9749, opened 2021-03-18
Generate Kubernetes PersistentVolumeClaims from named volumes

### LINKED ISSUE #5788: Play kube support for volumes created with `podman volume create`

jwillikers:
Fixes #5788

This commit adds support for named volumes in podman-generate-kube.
Named volumes are output in the YAML as PersistentVolumeClaims.
To avoid naming conflicts, the volume name is suffixed with "-pvc".
This commit adds a corresponding suffix for host path mounts.
Host path volumes are suffixed with "-host".

### DIFF
--- a/docs/source/markdown/podman-generate-kube.1.md
+++ b/docs/source/markdown/podman-generate-kube.1.md
@@ -6,9 +6,13 @@ podman-generate-kube - Generate Kubernetes YAML based on a pod or container
 **podman generate kube** [*options*] *container...* | *pod*
 
 ## DESCRIPTION
-**podman generate kube** will generate Kubernetes Pod YAML (v1 specification) from Podman one or more containers or a single pod. Whether
+**podman generate kube** will generate Kubernetes Pod YAML (v1 specification) from Podman from one or more containers or a single pod. Whether
 the input is for containers or a pod, Podman will always generate the specification as a Pod. The input may be in the form
 of a pod or one or more container names or IDs.
+
+Volumes appear in the generated YAML according to two different volume types. Bind-mounted volumes become *hostPath* volume types and named volumes become *persistentVolumeClaim* volume types. Generated *hostPath* volume types will be one of three subtypes depending on the state of the host path: *DirectoryOrCreate* when no file or directory exists at the host, *Directory* when host path is a directory, or *File* when host path is a file. The value for *claimName* for a *persistentVolumeClaim* is the name of the named volume registered in Podman.
+
+Potential name conflicts between volumes are avoided by using a standard naming scheme for each volume type. The *hostPath* volume types are named according to the path on the host machine, replacing forward slashes with hyphens less any leading and trailing forward slashes. The special case of the filesystem root, `/`, translates to the name `root`. Additionally, the name is suffixed with `-host` to avoid naming conflicts with *persistentVolumeClaim* volumes. Each *persistentVolumeClaim* volume type uses the name of its associated named volume suffixed with `-pvc`.
 
 Note that the generated Kubernetes YAML file can be used to re-run the deployment via podman-play-kube(1).
 
@@ -25,7 +29,7 @@ random port is assigned by Podman in the specification.
 
 ## EXAMPLES
 
-Create Kubernetes Pod YAML for a container called `some-mariadb` .
+Create Kubernetes Pod YAML for a container called `some-mariadb`.
 ```
 $ sudo podman generate kube some-mariadb
 # Generation of Kubernetes YAML is still under development!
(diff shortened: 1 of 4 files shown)


mheon:
/approve
LGTM on my end, nice work!

mheon:
Also @haircommander PTAL

haircommander:
in general LGTM!

one note I have: often folks want to be able for generate -> play -> generate to create the same YAML. Appending -host/-pvc without doing so in play will break this. If it turns out to be too cumbersome to fix this, I am fine not, but would be cool to handle this as well

jwillikers:
@haircommander 

I've tested using generate -> play -> generate for both a named volume and a bind-mounted volume.
The generated volume names are identical when this is done.
Of course, appending `-host` does alter the name compared to previous versions of Podman.

The volume names that receive a suffix are made from either the host path or the name of the named volume, depending on the type of volume.
This should lead to the volume names always being the same name for every named volume and host path.

Does that address or your concern or did I misunderstand you?
I'm not against using another method to address potential name collisions.
I chose the suffix method for simplicity.

libpod/kube.go
```diff
@@ -505,9 +502,34 @@ func libpodMountsToKubeVolumeMounts(c *Container) ([]v1.VolumeMount, []v1.Volume
...
+	vm.ReadOnly = ro
+
+	pvc := v1.PersistentVolumeClaimVolumeSource{ClaimName: v.Name, ReadOnly: ro}
+	vs := v1.VolumeSource{}
+	vs.PersistentVolumeClaim = &pvc
+	vo := v1.Volume{Name: name, VolumeSource: vs}
+
+	return vm, vo
+}
+
```
TomSweeneyRedHat, line 532: Are there man pages changes necessary to at least note this new ability?

jwillikers: There's barely any mention at all related to volumes in either the podman-play-kube(1) or podman-generate-kube(1) man pages.
The corresponding change to add PersistentVolumeClaims support to podman-play-kube didn't update the man pages.
This would probably require creating sections describing how Podman handles volumes for both podman-play-kube(1) and podman-generate-kube(1).
It's definitely something I would have found helpful to have documentation on.
If that's considered within the scope of this PR, I can write-up those sections.

TomSweeneyRedHat: @jwillikers I'd really like to see the info there in the man pages about the volumes.  It would be awesome if you could provide it.  That said, if you want to create an issue to that effect, then I'm fine merging this in now, or you could add the examples to this PR.  I'll let you make the call.  

If you go the issue route, please add a link to it in a comment here.  Thanks!

jwillikers: Thanks for getting back to me on that. I don't have a problem adding it to this PR, I should be able to take care of that later this afternoon. 👍

jwillikers: @TomSweeneyRedHat I've updated the documentation for podman-play-kube(1) and podman-generate-kube(1).
However, in doing so it brought up the fact that there are now multiple ways to specify named volumes with `podman play kube`, through a relative path using the `hostPath` volume type _or_ a `persistentVolumeClaim` volume type. I documented both, but wasn't sure if the documentation should prefer one method over the other.

docs/source/markdown/podman-generate-kube.1.md
```diff
@@ -6,10 +6,14 @@ podman-generate-kube - Generate Kubernetes YAML based on a pod or container
...
 of a pod or one or more container names or IDs.
 
+Volumes appear in the generated YAML according to two different volume types. Bind-mounted volumes become *hostPath* volume types and named volumes become *persistentVolumeClaim* volume types. Generated *hostPath* volume types will be one of three subtypes depending on the state of the host path: *DirectoryOrCreate* when no file or directory exists at the host, *Directory* when host path is a directory, or *File* when host path is a file. The value for *claimName* for a *persistentVolumeClaim* is the name of the named volume registered in Podman.
+
+Potential name conflicts between volumes are avoided by using a standard naming scheme for each volume type. *hostPath* volume types are named according to the path on the host machine, replacing forward slashed with hyphens less any leading and trailing forward slashes. The special case of the filesystem root, `/`, translates to the name `root`. Additionally, the name is suffixed with `-host` to avoid naming conflicts with *persistentVolumeClaim* volumes. Each *persistentVolumeClaim* volume type uses the name of its associated named volume suffixed with `-pvc`.
```
TomSweeneyRedHat, line 15:
```suggestion
Potential name conflicts between volumes are avoided by using a standard naming scheme for each volume type. The *hostPath* volume types are named according to the path on the host machine, replacing forward slashed with hyphens less any leading and trailing forward slashes. The special case of the filesystem root, `/`, translates to the name `root`. Additionally, the name is suffixed with `-host` to avoid naming conflicts with *persistentVolumeClaim* volumes. Each *persistentVolumeClaim* volume type uses the name of its associated named volume suffixed with `-pvc`.
```

TomSweeneyRedHat:
@jwillikers thanks!  The doc all looks good to me except one little nit.  However, I'd ask @baude to peruse/approve it as he's the knowledge expert in this arena.  Thanks again.

docs/source/markdown/podman-generate-kube.1.md
jwillikers: Thanks for pointing this out. I've made the requested change.

mheon:
Let's get this landed.
/lgtm

PR #9757, opened 2021-03-18
Cleanup /libpod/images/load handler

jwhonce:
* Remove orphaned code
* Add meaningful error from LoadImageFromSingleImageArchive() when
  heuristic fails to determine payload format
* Correct swagger to output correct types and headers

### DIFF
--- a/libpod/runtime_img.go
+++ b/libpod/runtime_img.go
@@ -313,15 +313,23 @@ func (r *Runtime) LoadImageFromSingleImageArchive(ctx context.Context, writer io
 		func() (types.ImageReference, error) {
 			return layout.NewReference(inputFile, "")
 		},
+		func() (types.ImageReference, error) {
+			// This item needs to be last to break out of loop and report meaningful error message
+			return nil,
+				errors.New("payload does not match any of the supported image formats (oci-archive, oci-dir, docker-archive, docker-dir)")
+		},
 	} {
 		src, err := referenceFn()
-		if err == nil && src != nil {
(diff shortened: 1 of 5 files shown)


libpod/runtime_img.go
```diff
@@ -313,15 +313,22 @@ func (r *Runtime) LoadImageFromSingleImageArchive(ctx context.Context, writer io
 		func() (types.ImageReference, error) {
 			return layout.NewReference(inputFile, "")
 		},
+		func() (types.ImageReference, error) {
+			// This item needs to be last to break out of loop and report meaningful error message
+			return nil, errors.New("loader failed to determine payload format, could be corrupted")
```
vrothberg, line 318: How about: `the specified input does not match any of the supported image formats (oci-archive, oci-dir, docker-archive, docker-dir)`?

vrothberg: Can you extend the `"podman load invalid file"` test in `test/systemd/120-load.bats` to string match the error message?

libpod/runtime_img.go
```diff
@@ -313,15 +313,22 @@ func (r *Runtime) LoadImageFromSingleImageArchive(ctx context.Context, writer io
 		func() (types.ImageReference, error) {
 			return layout.NewReference(inputFile, "")
 		},
+		func() (types.ImageReference, error) {
```
vrothberg, line 316: :1st_place_medal: love the idea!

I think @edsantiago will like it as well; the new error message is so much more user friendly than before.

TomSweeneyRedHat:
LGTM
assuming happy tests

jwhonce:

vrothberg (commented):
/lgtm

Issue #9129, opened 2021-01-27
"podman play kube" yaml multiple resources


ADIX7:
/kind feature

Add support for multiple resources in yaml used by `podman play kube` like kubernetes does have.

baude:
interested in contributing?

ADIX7:
Unfortunately I won't have time to do this in the near future.

mheon:
Can you provide more details here? An example YAML file with the exact resources you want us to support identified, perhaps?

ADIX7:
I mean to concate multiple yaml-s generated by `podman generate kube` separeted by `---` just like in a kubernetes yaml.
I would like to use this feature with https://github.com/containers/podman/issues/8996 to do something like `kustomize build my-overlay | podman play kub -`. The pipe features is WIP but kustomize create a yaml with multiple resources separeted by `---`.

As for this issue:
For example if I create 2 pod like this:

podman create --pod new:nginxpod1 --name nginx1 nginx1
podman create --pod new:nginxpod2 --name nginx1 nginx2

And generate kubernetes yamls

Pod1: `podman generate kube nginxpod1 > pod1.yaml`

*pod1.yaml*
```
# Generation of Kubernetes YAML is still under development!
#
# Save the output of this file and use kubectl create -f to import
# it into Kubernetes.
#
# Created with podman-2.2.1
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2021-01-27T20:08:05Z"
  labels:
    app: nginxpod1
  name: nginxpod1
spec:
  containers:
  - command:
    - nginx
    - -g
    - daemon off;
    env:
    - name: PATH
      value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
    - name: TERM
      value: xterm
    - name: container
      value: podman
    - name: NGINX_VERSION
      value: 1.19.6
    - name: NJS_VERSION
      value: 0.5.0
    - name: PKG_RELEASE
      value: 1~buster
    - name: HOSTNAME
      value: nginxpod1
    image: docker.io/library/nginx:latest
    name: nginx1
    resources: {}
    securityContext:
      allowPrivilegeEscalation: true
      capabilities: {}
      privileged: false
      readOnlyRootFilesystem: false
      seLinuxOptions: {}
    workingDir: /
  restartPolicy: Never
status: {}
---
metadata:
  creationTimestamp: null
spec: {}
status:
  loadBalancer: {}

```

[...]

PR #9759, opened 2021-03-19
Support multi doc yaml for generate/play kube

EduardoVega:
Fix: #9129 

This PR adds the following for podman play / generate kube
- podman play kube
  - Reads every document in a multi doc yaml file
  - Fails if any of the documents is invalid
  - Creates a podman pod if document is a valid kubernetes pod or deployment
  - Skips any other valid kubernetes kind
 - podman generate kube
   - Accepts multiple podman pods
   - Generates a multi doc yaml file

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9
 	golang.org/x/sys v0.0.0-20210315160823-c6e025ad8005
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
-	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	k8s.io/api v0.20.5
 	k8s.io/apimachinery v0.20.5
 )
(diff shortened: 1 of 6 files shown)


zhangguanzhang:
should also handle  the  configmap   and volumes

EduardoVega:
@zhangguanzhang I can add configmaps. For volumes (k8s persistent volume claims) I think there is some work in progress for that. I can track all the issues/prs that are related to it and check their status because that should be a dependency for this change.

rhatdan:
Nice work @EduardoVega LGTM

rhatdan:

rhatdan:
@vrothberg @giuseppe @saschagrunert PTAL

go.mod
```diff
@@ -67,6 +67,7 @@ require (
 	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9
 	golang.org/x/sys v0.0.0-20210315160823-c6e025ad8005
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
+	gopkg.in/yaml.v2 v2.4.0
```
vrothberg, line 70: Can we use `yaml.v3` below?  I'd like to avoid vendoring multiple versions of the same library.

EduardoVega: It should work with `yaml.v3`. I have just changed the code.

vrothberg (commented):
LGTM, thank you!

PR #9758, opened 2021-03-18
Fix volumes and networks list/prune filters in http api

jmguzik:
This is the continuation of work started in #9711. It turns out
that list/prune commands for volumes in libpod/compat api have
very dangerous error handling when broken filter input is supplied.
Problem also affects network list/prune in libpod. This commit
unifies filter handling across libpod/compat api and adds sanity
apiv2 testcases.

Signed-off-by: Jakub Guzik <jakubmguzik@gmail.com>

/kind bug
/kind cleanup

Below applies to libpod HTTP network prune, volume prune and volume list (both libpod and compat http)
1. Docker: scenario => when the user makes mistake when using filters, then no negative consequences
```
$ docker volume create v1
v1
$ docker volume create v2
v2
$ curl -gG -XPOST --unix-socket /var/run/docker.sock  http://localhost/v1.41/volumes/prune?filters=dssd21 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    67  100    67    0     0  67000      0 --:--:-- --:--:-- --:--:-- 67000
{
  "message": "invalid character 'd' looking for beginning of value"
}


```

2. Podman (compat API): scenario => mistake, then there are negative consequences
```
$ bin/podman volume create v1
v1
$ bin/podman volume create v2
v2
$ curl -gG -XPOST --unix-socket /run/user/1000/podman/podman.sock  http://d/v3.0.0/volumes/prune?filters=dssd21 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    50  100    50    0     0   2173      0 --:--:-- --:--:-- --:--:--  2173
{
  "VolumesDeleted": [
    "v1",
    "v2"
  ],
  "SpaceReclaimed": 0
}
$ bin/podman volume create v1bis
v1bis
$ bin/podman volume create v2bis
v2bis
$ curl -gG -XPOST --unix-socket /run/user/1000/podman/podman.sock  http://d/v3.0.0/volumes/prune?filters={garba1ge | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    56  100    56    0     0   2666      0 --:--:-- --:--:-- --:--:--  2666
{
  "VolumesDeleted": [
    "v1bis",
    "v2bis"
  ],
[...]

PR #9773, opened 2021-03-20
Fix containers list/prune http api filter behaviour

jmguzik:
The problem described in #9711 and followed by #9758 affects
containers as well. When user provides wrong filter input, error
message should occur, not fallback to full list/prune command.
This change fixes the issue. Additionally, there are error message
fixes for docker http api compat.


zhangguanzhang:
also need update the doc. otherwise LGTM

jmguzik:
> also need update the doc. otherwise LGTM

Three is no functional change, just fix of the not working functionality. Which docs i need to update in that case?

jwhonce:
LGTM

Luap99 (approved):

PR #9785, opened 2021-03-22
Unification of label and until filters across list/prune endpoints

jmguzik:
All label and until filters are using a similar match function. This PR unifies them and adds UT.
/kind cleanup

### DIFF
--- a/pkg/util/filters.go
+++ b/pkg/util/filters.go
@@ -11,7 +11,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ComputeUntilTimestamp extracts unitil timestamp from filters
+// ComputeUntilTimestamp extracts until timestamp from filters
 func ComputeUntilTimestamp(filter string, filterValues []string) (time.Time, error) {
 	invalid := time.Time{}
 	if len(filterValues) != 1 {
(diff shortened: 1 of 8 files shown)


pkg/util/filters.go
```diff
@@ -93,3 +93,27 @@ func PrepareFilters(r *http.Request) (*map[string][]string, error) {
...
+				matched = true
+				break
+			}
+		}
+		if !matched {
+			return false
+		}
+	}
+	return true
+}
```
Luap99, line 119: I think we can make this function a bit cleaner, like this:
```golang
outer:
	for _, filterValue := range filterValues {
		filterArray := strings.SplitN(filterValue, "=", 2)
		filterKey := filterArray[0]
		if len(filterArray) > 1 {
			filterValue = filterArray[1]
		} else {
			filterValue = ""
		}
		for labelKey, labelValue := range labels {
			if labelKey == filterKey && (filterValue == "" || labelValue == filterValue) {
				continue outer
			}
		}
		return false
	}
return true
```

jmguzik: I personally do not like continue outer formula, so I have chosen a version from containers, however, it is indifferent to me since both do the same.
Seems there are some CI problems and at least part is not related to this commit so I will continue once that gets sorted.

jmguzik:
I decided to add also until to this PR in a separate commit.

rhatdan:
Nice cleanup, thanks @jmguzik 
LGTM

Luap99 (approved):

PR #9788, opened 2021-03-23
[CI:DOCS] rootless: Tell the user what was led to the error, not just what it is

andrewshadura:
Users coming e.g. from Docker do not always read the manual and expect podman to not require sudo or uidmap, for them the default message is not very helpful:

    Error: Cannot connect to the Podman socket, make sure there is a Podman REST API service running.:
    cannot find newuidmap: exec: "newuidmap": executable file not found in $PATH

Adding a bit more context to this would help to nudge them into the right direction and tell them what to look for in the documentation:

    cannot find newuidmap required for rootless mode (try running as root?): exec: "newuidmap": executable file not found in $PATH

Signed-off-by: Andrej Shadura `<andrew.shadura@collabora.co.uk>`

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -116,7 +116,7 @@ func tryMappingTool(uid bool, pid int, hostID int, mappings []idtools.IDMap) err
 	}
 	path, err := exec.LookPath(tool)
 	if err != nil {
-		return errors.Wrapf(err, "cannot find %s", tool)
+		return errors.Wrapf(err, "command required for rootless mode with multiple IDs")
 	}
 
 	appendTriplet := func(l []string, a, b, c int) []string {
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -203,12 +203,12 @@ Rootless Podman requires the newuidmap and newgidmap programs to be installed.
 
 #### Symptom
 
-If you are running Podman or buildah as a not root user, you get an error complaining about
+If you are running Podman or Buildah as a rootless user, you get an error complaining about
 a missing newuidmap executable.
 
 ```
 podman run -ti fedora sh
-cannot find newuidmap: exec: "newuidmap": executable file not found in $PATH
+command required for rootless mode with multiple IDs: exec: "newuidmap": executable file not found in $PATH
 ```
 
 #### Solution


mheon:
Running as root there doesn't seem like the right suggestion? If `newuidmap` is missing, it needs to be installed.

andrewshadura:
> Running as root there doesn't seem like the right suggestion? If `newuidmap` is missing, it needs to be installed.

Well, it is if they didn’t intend to run rootless.

wladmis:
> Running as root there doesn't seem like the right suggestion? If `newuidmap` is missing, it needs to be installed.

To be clarified running as root does not mean that there will be root inside of container. This can be helpful when userns is not used.

mheon:
But very few people *didn't* mean to run rootless. Rootless is one of Podman's most-used features, and I'd say we have more rootless users than root users. This error message is nonsensical for them, the majority of our users.

andrewshadura:
How about `try running as root if you didn’t want rootless mode`?

andrewshadura:
Or do you have a different phrasing in mind? I think it’s important to point out both options exist.

mheon:
@TomSweeneyRedHat @rhatdan You two might have opinions here?

pkg/rootless/rootless_linux.go
```diff
@@ -105,7 +105,7 @@ func tryMappingTool(uid bool, pid int, hostID int, mappings []idtools.IDMap) err
 	}
 	path, err := exec.LookPath(tool)
 	if err != nil {
-		return errors.Wrapf(err, "cannot find %s", tool)
+		return errors.Wrapf(err, "cannot find %s required for rootless mode (try running as root?)", tool)
```
rhatdan, line 108: This is stuttering.

How about just  errors.Wrapf(err, "command required for rootless mode")

Which will result in 

command required for rootless mode:  exec: "newuidmap": executable file not found in $PATH

troubleshooting.md
```diff
@@ -203,12 +203,12 @@ Rootless Podman requires the newuidmap and newgidmap programs to be installed.
 
 #### Symptom
 
-If you are running Podman or buildah as a not root user, you get an error complaining about
+If you are running Podman or buildah as a non-root user, you get an error complaining about
```
TomSweeneyRedHat, line 206:
```suggestion
If you are running Podman or Buildah as a rootless user, you get an error complaining about
```

TomSweeneyRedHat:
Beyond @rhatdan's comment, I'd a couple small nits in the text.  Otherwise, TYVM @andrewshadura for the PR!

giuseppe:
newuidmap/newgidmap are required only for rootless mode with multiple IDs available.

It is still possible to use rootless but you are limited to just one user, together with ignore_chown_errors

pkg/rootless/rootless_linux.go
giuseppe: I'd make it clearer as: `errors.Wrapf(err, "command required for rootless mode with multiple IDs")` or similar

andrewshadura:
> newuidmap/newgidmap are required only for rootless mode with multiple IDs available.
> 
> It is still possible to use rootless but you are limited to just one user, together with ignore_chown_errors

How? I don’t think that’s documented very well, or at all?

giuseppe:
> How? I don’t think that’s documented very well, or at all?

yes, unfortunately this part is not documented properly, but you can try adding: `podman --storage-opt ignore_chown_errors ...` when a single ID is available and all the files will be owned by just one user.

It is not done automatically (and I think it shouldn't) because potentially it can lead to security problems when the files ownership inside the container matters.

rhatdan:
Right, we recommend this only for specific use cases like HPC environments. 
Bottom line is you will only be able to run with a single UID within the container and that UID will be fully privileged to do read/write anything in the container.  If this works for you then you can use it.

Also pushing these types of containers to container registries, might cause some issues.

rhatdan:
@andrewshadura Still working on this?

pkg/rootless/rootless_linux.go
```diff
@@ -116,7 +116,7 @@ func tryMappingTool(uid bool, pid int, hostID int, mappings []idtools.IDMap) err
 	}
 	path, err := exec.LookPath(tool)
 	if err != nil {
-		return errors.Wrapf(err, "cannot find %s", tool)
+		return errors.Wrapf(err, "command required for rootless mode with multiple IDs")
```
rhatdan, line 119: I would still have the name of the command liste.
return errors.Wrapf(err, "command %q required for rootless mode with multiple IDs", tool)

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

Issue #9290, opened 2021-02-09
[APIv2] building container with isolation=default started failing after #9246


riyad:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Trying to find regressions by exercising the APIv2 trough [docker-py](https://github.com/docker/docker-py)'s test suite (see #5386) I came across a regressions in the [`BuildTest::test_build_isolation`](https://github.com/docker/docker-py/blob/master/tests/integration/api_build_test.py#L185) test.

Building containers with `isolation=default` started failing with commit 19507d0ffe8cda0a69f056838556f471fd9e61fa.

**Steps to reproduce the issue:**

1.
```shell
cat > Dockerfile <<EOF
FROM scratch
CMD sh -c "echo \'Deaf To All But The Song\'
EOF
```

2. `tar cf test.tar Dockerfile`

3. `curl -sS --unix-socket /var/run/user/1000/podman/podman.sock -XPOST 'http://localhost/v1.40/build?t=isolation&isolation=default'  -H "Content-Type: application/tar" --data-binary '@test.tar'`

**Describe the results you received:**

Run against commit 19507d0ffe8cda0a69f056838556f471fd9e61fa:
```json
{"cause":"schema: error converting value for \"isolation\"","message":"schema: error converting value for \"isolation\"","response":400}
```

**Describe the results you expected:**

Run against commit 2bf13219f587d769400f26aeaed05930c34ce3d7:
```json
{"stream":"STEP 1: FROM scratch\n"}
{"stream":"STEP 2: CMD sh -c \"echo \\'Deaf To All But The Song\\'\n"}
{"stream":"STEP 3: COMMIT isolation\n"}
{"stream":"Getting image source signatures\n"}
{"stream":"Copying blob sha256:4ca545ee6d5db5c1170386eeb39b2ffe3bd46e5d4a73a9acbebc805f19607eb3\n"}
{"stream":"Copying config sha256:fe9067e46ed83a50d10c196a74bed82e66fc11a714240b001f1e70c82d6eaec9\n"}
{"stream":"Writing manifest to image destination\n"}
{"stream":"Storing signatures\n"}
{"stream":"--\u003e fe9067e46ed\n"}
{"stream":"fe9067e46ed83a50d10c196a74bed82e66fc11a714240b001f1e70c82d6eaec9\n"}
{"stream":"Successfully built fe9067e46ed8\n"}
```

mheon:
[...]

PR #9790, opened 2021-03-23
fix: build endpoint for compat API

matejvasek:
resolves: #9290

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -481,7 +481,7 @@ func parseNetworkConfigurationPolicy(network string) buildah.NetworkConfiguratio
 	}
 }
 
-func parseLibPodIsolation(isolation string) buildah.Isolation {
+func parseLibPodIsolation(isolation string) buildah.Isolation { // nolint
 	if val, err := strconv.Atoi(isolation); err == nil {
 		return buildah.Isolation(val)
 	}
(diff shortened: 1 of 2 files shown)


rhatdan:
I have a similar PR, but have not been able to get back to this.  Some of this code is in the master branch of Buildah, now I believe.

pkg/api/handlers/compat/images_build.go
```diff
@@ -205,9 +205,15 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		isolation := buildah.IsolationDefault
 	*/
 	if utils.IsLibpodRequest(r) {
-		//		isolation = buildah.Isolation(query.Isolation)
+		// isolation = parseLibPodIsolation(query.Isolation)
```
TomSweeneyRedHat, line 208: is this comment of any value?

TomSweeneyRedHat:
Changes LGTM, but I'm not sure if we should go with this or @rhatdan's WIP?

pkg/api/handlers/compat/images_build.go
matejvasek: If I understand it correctly the comment should be removed at some point see TODO above.

rhatdan: Yes I have a PR that is attempting to remove it.
I think we just merge this and I will continue to rework my PRs

PR #9791, opened 2021-03-23
Updated schema:resource to schema:network to comply with JSON API

kellen-dunham:
The libpod JSON API handler for play kube was not decoding the URL query parameters correctly as per the API spec. Changing "reference" to "network" fixed the issue.


rhatdan:
Sadly you need to squash these together, and we need a test added.

rhatdan:
@kellen-dunham Thanks, change LGTM, but need to get it past tests.

TomSweeneyRedHat:
Change LGTM
@kellen-dunham I don't think you signed your commit which our CI system requires.  Can you please do a `git commit --amed -s`, fetch, rebase and push?

PR #9794, opened 2021-03-23
Cirrus: Reduce to bare-minimum testing

cevich:
Now that the v3.1 branch is in process, there is no need for extensive
testing of v3.0.  Dial it down to the bare-minimum to reduce overall
maintenance burden.


mheon:
LGTM

PR #9795, opened 2021-03-23
Bump to v3.2.0-dev

mheon:
Bump version to reflect that we now have a v3.1 branch.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -27,7 +27,7 @@ const (
 // NOTE: remember to bump the version at the top
 // of the top-level README.md file when this is
 // bumped.
-var Version = semver.MustParse("3.1.0-dev")
+var Version = semver.MustParse("3.2.0-dev")
 
 // See https://docs.docker.com/engine/api/v1.40/
 // libpod compat handlers are expected to honor docker API versions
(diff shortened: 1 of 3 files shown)


Luap99:
Tests are not happy

version/version.go
```diff
@@ -38,7 +38,7 @@ var Version = semver.MustParse("3.1.0-dev")
 var APIVersion = map[Tree]map[Level]semver.Version{
 	Libpod: {
 		CurrentAPI: Version,
-		MinimalAPI: semver.MustParse("3.0.0"),
+		MinimalAPI: semver.MustParse("3.2.0-dev"),
```
Luap99, line 41: I don't think we have to bump the minimal API version. @jwhonce PTAL

mheon: The tests start blowing up if we don't - the Current API version is set to v3.2.0-dev, and according to semantic versioning, v3.2.0 is more recent than v3.2.0-dev (suffixes indicate pre-release). Unless we bump this. the minimum API version is lower than the current API version.

Luap99: I don't think so, looking at the changes I can see that you changed this from 3.0.0 so it worked together with the 3.1.0-dev version.

jwhonce: @Luap99 @mheon I would suggest that MimimalAPI be n-1 (Current release version) at least at until the end of the release cycle for a given branch. As we stabilize the API, I could see a day where the drift between Current and Minimal grows.

PR #9808, opened 2021-03-24
[NO TESTS NEEDED] Fix rootless volume plugins

thephoenixofthevoid:
According to https://github.com/containers/podman/issues/9650 there are few issues when using volume plugins with rootless containers.

The reason behind these is local-driver-specific code being executed for all cases. This is a minimal fix to address these issues.

Please also note that:
1. In fact, there is a need to clearly separate local-driver-specific code from plugin-related, because there are little to none logic that should be executed in both cases.
2. The removed check is actually already correctly implemented here
https://github.com/containers/podman/blob/d92b94677cc816a1e157802836195430b731d015/libpod/runtime_volume_linux.go#L66-L77
3. These kind of checks is actually implementation specific, so it seems natural to keep them close to the implementation.

At this point no new tests are needed, because there is no reliable and stable volume plugin yet. Using unstable volume plugin is rather harmful, as such test is much more likely to break due to unrelated issue, than due to regression of this fix.

### DIFF
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -1577,8 +1577,6 @@ func WithVolumeLabels(labels map[string]string) VolumeCreateOption {
 }
 
 // WithVolumeOptions sets the options of the volume.
-// If the "local" driver has been selected, options will be validated. There are
-// currently 3 valid options for the "local" driver - o, type, and device.
 func WithVolumeOptions(options map[string]string) VolumeCreateOption {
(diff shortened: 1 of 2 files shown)


mheon:
/approve
LGTM again.

Would love to get this landed in time for 3.1.0, but if not we'll probably be cutting a 3.1.1 shortly thereafter.

PR #9816, opened 2021-03-25
Correct json encoding field name for Namespace type

jwhonce:
[NO TESTS NEEDED]

* When using the **Namespace** type, the field Value was json encoded
  with the name "string" vs "value"
* No tests included as fields are not reported via API


rhatdan:
LGTM

rhatdan:

PR #9554, opened 2021-03-01
Cirrus: Use Fedora 34beta images

cevich:
Signed-off-by: Chris Evich <cevich@redhat.com>

PR #9817, opened 2021-03-25
image unit tests - make them pass on F34 with enforced short-name mode

vrothberg:
Helps bring https://github.com/containers/podman/pull/9554 forward. Please refer to the individual commits.

### DIFF
--- a/libpod/image/testdata/registries.conf
+++ b/libpod/image/testdata/registries.conf
@@ -1,0 +1,4 @@
+short-name-mode="enforcing"
+
+[aliases]
+"busybox"="docker.io/library/busybox"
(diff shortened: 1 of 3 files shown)


libpod/image/testdata/registries.conf
```diff
@@ -0,0 +1,4 @@
+short-name-mode="enforcing"
+
+[aliases]
+"busybox"="docker.io/library/busybox"
```
cevich, line 4: Is going to respect our proxy over to using quay.io instead of docker for images?

Just checking since the rate-limiting is a very real problem.

vrothberg: Very good question! Yes, we will continue pulling from `quay.io`.

rhatdan:
LGTM

ashley-cui (commented):
LGTM

PR #9818, opened 2021-03-25
system tests: new interactive tests

edsantiago:
socat can create a dummy PTY that we can manipulate. This
lets us run a variety of tests that we couldn't before,
involving "run -it", and stty, and even "load" with no args.

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -668,15 +668,4 @@ json-file | f
     is "$output" ".*HOME=/.*"
 }
 
-@test "podman run --tty -i failure with no tty" {
-    run_podman run --tty -i --rm $IMAGE echo hello < /dev/null
-    is "$output" ".*The input device is not a TTY.*"
-
-    run_podman run --tty=false -i --rm $IMAGE echo hello < /dev/null
-    is "$output" "hello"
-
-    run_podman run --tty -i=false --rm $IMAGE echo hello < /dev/null
-    is "$output" "hello"
-}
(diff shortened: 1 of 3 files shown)


rhatdan:
Nice work @edsantiago 
LGTM

vrothberg (commented):
That is _really_ cool stuff, @edsantiago !

edsantiago:
Thank you @vrothberg ! Do you think there's a way to make it work with the shortnames prompt? I tried for a while before submitting but could never get the prompt, only the fallback. (Perhaps because stdout is a pipe - it has to be?) Anyhow, if you can think of any other tests that could benefit from this, please LMK.

vrothberg:
> Do you think there's a way to make it work with the shortnames prompt? I tried for a while before submitting but could never get the prompt, only the fallback. (Perhaps because stdout is a pipe - it has to be?)

That would be fantastic. If stdout is a pipe, that may be the issue. The prompt wants stdin and stdout to be a terminal.

> Anyhow, if you can think of any other tests that could benefit from this, please LMK.

`podman login` could be a good candidate.  It may prompt for user/pw.

Luap99:
@edsantiago I just saw one of the tests flake in #9423, see https://cirrus-ci.com/task/5159880362295296?logs=main#L195

edsantiago:
@Luap99 thank you. I was seeing that during development, it's a race condition that I'll need to chase and file a bug for.

PR #9810, opened 2021-03-25
Fix filters list/prune in image http compat/libpod api endpoints

jmguzik:
Signed-off-by: Jakub Guzik <jakubmguzik@gmail.com>

/kind bug
As described in previous commits and PRs, in certain situations prune/list filter handling is broken (see #9773, #9711 and #9758 ). The problem affects images as well. This is an attempt to fix it.

Adding also handling of dangling filter in compat API (was unimplemented).

PR #9822, opened 2021-03-25
Fix list pods filter handling in libpod api

jmguzik:
/kind bug

I found one more affected by wrong filter handling in case of bad input. Its pods list HTTP API. I think this could be the last one in regards to the issue described in previous commits (see #9773, #9711, #9758, #9810).


rhatdan:
Thanks @jmguzik 
/approve
LGTM

PR #9829, opened 2021-03-26
[NO TESTS NEEDED] Use same function podman-remote rmi as podman

rhatdan:
Make sure fixes that go into local podman commands also work in
podman-remote, by using the same function.

Since this is just a rewrite of existing code, existing tests should
handle it.

### DIFF
--- a/pkg/api/handlers/compat/images_remove.go
+++ b/pkg/api/handlers/compat/images_remove.go
@@ -4,7 +4,10 @@ import (
 	"net/http"
 
 	"github.com/containers/podman/v3/libpod"
+	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/pkg/api/handlers/utils"
+	"github.com/containers/podman/v3/pkg/domain/entities"
+	"github.com/containers/podman/v3/pkg/domain/infra/abi"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Looks like a tty test failed, not sure how from this change.
Code LGTM otherwise.

pkg/api/handlers/compat/images_remove.go
```diff
@@ -30,28 +33,32 @@ func RemoveImage(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		utils.ImageNotFound(w, name, errors.Wrapf(err, "failed to find image %s", name))
-		return
+	imageEngine := abi.ImageEngine{Libpod: runtime}
+
+	options := entities.ImageRemoveOptions{
+		Force: query.Force,
 	}
+	report, rmerrors := imageEngine.Remove(r.Context(), []string{name}, options)
+	if rmerrors != nil && rmerrors[0] != nil {
```
Luap99, line 42: This check can panic. `rmerrors` could be a empty slice (not nil) so access to the first element will panic. You would need to use `len(rmerrors) > 0`, this also covers the nil check.

Issue #9831, opened 2021-03-26
podman-remote run -it: failed to resize TTY: container in wrong state "stopped|created"


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$  ./bin/podman-remote run --rm -it quay.io/libpod/alpine echo hello
hello
WARN[0000] failed to resize TTY: container "87e8cefa1dd42c9a819b76ae758d12a3e2358696320e0f178e0db3deabbb1195" in wrong state "stopped"
WARN[0000] Container 87e8cefa1dd42c9a819b76ae758d12a3e2358696320e0f178e0db3deabbb1195 does not exist: no container with name or ID "87e8cefa1dd42c9a819b76ae758d12a3e2358696320e0f178e0db3deabbb1195" found: no such container
```
The errors are not deterministic. Sometimes the state is "created" instead of "stopped". Sometimes the second message is an error instead of warning:
```console
$  ./bin/podman-remote run --rm -it quay.io/libpod/alpine echo hello
hello
WARN[0000] failed to resize TTY: container "49980f36f6f2c0adddcdc4949199c2b58020074322617dad897e3188d7756f2c" in wrong state "stopped"
ERRO[0000] Error removing container 49980f36f6f2c0adddcdc4949199c2b58020074322617dad897e3188d7756f2c: error looking up container "49980f36f6f2c0adddcdc4949199c2b58020074322617dad897e3188d7756f2c" mounts: layer not known
```
(Even though it says ERROR, exit code is always 0).

Sometimes there's no second line at all, only the first.

master @ 9e23e0b3e3b219cbdc42fac4f843d6d2ec97421b

edsantiago:
Forgot to mention: there's a simple workaround, just add a `sleep`:
```console
$ ./bin/podman-remote run --rm -it quay.io/libpod/alpine sh -c 'sleep 0.3;echo hello'
hello
```
0.2 is not quite good enough, and 0.1 is even less good.

edsantiago:
For reference: [flake failure in CI](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5159880362295296/html/sys-remote-ubuntu-2010-root-host.log.html)

PR #9833, opened 2021-03-26
Remove resize race condition

rhatdan:
Since podman-remote resize requests can come in at random times, this
generates a real potential for race conditions. We should only be
attempting to resize TTY on running containers, but the containers can
go from running to stopped at any time, and returning an error to the
caller is just causing noice.

This change will basically ignore requests to resize terminals if the
container is not running and return the caller to success.  All other
callers will still return failure.

Fixes: https://github.com/containers/podman/issues/9831

### DIFF
--- a/test/system/450-interactive.bats
+++ b/test/system/450-interactive.bats
@@ -57,9 +57,6 @@ function teardown() {
 
     # ...and make sure stty under podman reads that.
     # FIXME: 'sleep 1' is needed for podman-remote; without it, there's
-    # a race condition resulting in the following warning:
-    #   WARN[0000] failed to resize TTY: container "xx" in wrong state "stopped"
-    # (also "created")
     run_podman run -it --name mystty $IMAGE sh -c 'sleep 1;stty size' <$PODMAN_TEST_PTY
     is "$output" "$rows $cols" "stty under podman reads the correct dimensions"
 }
--- a/pkg/api/handlers/compat/resize.go
+++ b/pkg/api/handlers/compat/resize.go
@@ -19,8 +19,9 @@ func ResizeTTY(w http.ResponseWriter, r *http.Request) {
 
 	// /containers/{id}/resize
 	query := struct {
-		Height uint16 `schema:"h"`
-		Width  uint16 `schema:"w"`
+		Height           uint16 `schema:"h"`
+		Width            uint16 `schema:"w"`
+		IgnoreNotRunning bool   `schema:"running"`
 	}{
 		// override any golang type defaults
 	}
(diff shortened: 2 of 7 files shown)


rhatdan:
@edsantiago PTAL, this is going to fail because of missing tests, do you have a handy test for this?
@mheon WDYT of this?

mheon:
LGTM

edsantiago:
@rhatdan yes I do, and it's my favorite kind of test, the one that consists entirely of deletions!
```diff
diff --git a/test/system/450-interactive.bats b/test/system/450-interactive.bats
index d047b9f25..a2db39492 100644
--- a/test/system/450-interactive.bats
+++ b/test/system/450-interactive.bats
@@ -56,11 +56,7 @@ function teardown() {
     stty rows $rows cols $cols <$PODMAN_TEST_PTY

     # ...and make sure stty under podman reads that.
-    # FIXME: 'sleep 1' is needed for podman-remote; without it, there's
-    # a race condition resulting in the following warning:
-    #   WARN[0000] failed to resize TTY: container "xx" in wrong state "stopped"
-    # (also "created")
-    run_podman run -it --name mystty $IMAGE sh -c 'sleep 1;stty size' <$PODMAN_TEST_PTY
+    run_podman run -it --name mystty $IMAGE stty size <$PODMAN_TEST_PTY
     is "$output" "$rows $cols" "stty under podman reads the correct dimensions"
 }

```
Unfortunately, it is now triggering something else:
```
   #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   #|     FAIL: stty under podman reads the correct dimensions
   #| expected: '63 30'
'  #|   actual: 'stty: standard input
   #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This is a flake, about 50% of the time it passes (root and rootless), the rest it fails, always the same message. Before I try tracking it down further: does that message ("stty: standard input") mean anything to you? (He says, hoping there's an obvious quick fix)

edsantiago:
The `stty: standard input` message does not happen if I leave the `sleep` in. (Or at least it doesn't happen over many retries). If you can't figure out the cause, just leave the `sleep` in place and change the comment:
```bash
# FIXME: 'sleep 1' is needed for podman-remote; without it, there's
# a race condition resulting in 'stty: standard input` instead of dimensions.
```
I can file an issue later. 

(Point being, by changing the comment, (a) you don't need NO TESTS NEEDED, (b) the comment is closer to reality, therefore helping a future developer)

mheon:
My best guess is that it means the terminal never had a size assigned to it - but that's just a guess.

edsantiago:
Hmmm I'm still seeing the stty failure with 2d525684185a5c2a2352b275db9b8e23444e5087. There are three podman-remote system tests, so I'm guessing one will pass but I do hope that one or two fail. I don't think this can go in as it is, without the stty error fixed.

test/system/450-interactive.bats
```diff
@@ -57,9 +57,6 @@ function teardown() {
 
     # ...and make sure stty under podman reads that.
     # FIXME: 'sleep 1' is needed for podman-remote; without it, there's
-    # a race condition resulting in the following warning:
-    #   WARN[0000] failed to resize TTY: container "xx" in wrong state "stopped"
-    # (also "created")
     run_podman run -it --name mystty $IMAGE sh -c 'sleep 1;stty size' <$PODMAN_TEST_PTY
```
edsantiago, line 60: A line got dropped here but it's not important enough to be worth resubmitting. I'll file an issue next week for the new flake.

edsantiago:
/lgtm

because it's causing too many CI flakes

PR #9835, opened 2021-03-26
[NO TESTS NEEDED] Fix for kernel without CONFIG_USER_NS

ntkme:
I have an embedded device for which the kernel was compiled without `CONFIG_USER_NS`.

Podman 2.2. was working fine with `runc` in root mode, but Podman 3.x is breaking due to a change introduced by https://github.com/containers/podman/commit/ecedda63a6488162c9aad2a99c1ada172340ac7f:

https://github.com/containers/podman/blob/ecedda63a6488162c9aad2a99c1ada172340ac7f/libpod/container_internal_linux.go#L532-L537

The root cause is that the code added above crashing at line 532 due to `/proc/self/uid_map` and `/proc/self/gid_map` not exists in the embedded kernel I'm using.

Here is an example where `runc` handles such kernel without userns support:

https://github.com/containers/podman/blob/c81e273835303664d62c4e958a7f34877c05ed49/vendor/github.com/opencontainers/runc/libcontainer/system/linux.go#L97-L107

This PR follows the same pattern that `runc` uses.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -607,10 +607,16 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 
 	availableUIDs, availableGIDs, err := rootless.GetAvailableIDMaps()
 	if err != nil {
-		return nil, err
+		if os.IsNotExist(err) {
+			// The kernel-provided files only exist if user namespaces are supported
+			logrus.Debugf("user or group ID mappings not available: %s", err)
+		} else {
+			return nil, err
+		}
+	} else {
+		g.Config.Linux.UIDMappings = rootless.MaybeSplitMappings(g.Config.Linux.UIDMappings, availableUIDs)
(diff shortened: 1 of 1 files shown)


libpod/container_internal_linux.go
```diff
@@ -607,10 +607,11 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 
 	availableUIDs, availableGIDs, err := rootless.GetAvailableIDMaps()
 	if err != nil {
-		return nil, err
+		// The kernel-provided files only exist if user namespaces are supported
```
rhatdan, line 610: You should be checking the specific error here. Probably ENOENT.  Theoretically you could get other errors line EPERM. Also probably should logrus.Debug this.

ntkme: Thanks for the quick feedback. Added check for specific error as well as a debug message.

TomSweeneyRedHat:
LGTM, thx @ntkme I'd like a head nod from @giuseppe

ntkme:
One test failed but the failure looks totally unrelated.

rhatdan:
LGTM

PR #9838, opened 2021-03-27
Add problematic volume name to kube play error messages

xordspar0:
When kube play fails to create a volume, it should say which volume had
the problem so the user doesn't have to guess. For the following pod
spec:

	apiVersion: v1
	kind: Pod
	metadata:
	  name: mypod
	spec:
	  containers:
	    - name: myfrontend
	      image: nginx
	      volumeMounts:
	      - mountPath: "/var/www/html"
		name: mypd
	  volumes:
	    - name: mypd
	      hostPath:
		path: /var/blah

podman will now report:

	Error: failed to create volume "mypd": error in parsing HostPath
	in YAML: error checking path "/var/blah": stat /var/blah: no such
	file or directory

### DIFF
--- a/pkg/specgen/generate/kube/volume.go
+++ b/pkg/specgen/generate/kube/volume.go
@@ -116,7 +116,7 @@ func InitializeVolumes(specVolumes []v1.Volume) (map[string]*KubeVolume, error) 
 	for _, specVolume := range specVolumes {
 		volume, err := VolumeFromSource(specVolume.VolumeSource)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrapf(err, "failed to create volume %q", specVolume.Name)
 		}
 
 		volumes[specVolume.Name] = volume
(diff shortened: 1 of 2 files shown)


pkg/specgen/generate/kube/volume.go
```diff
@@ -116,7 +116,7 @@ func InitializeVolumes(specVolumes []v1.Volume) (map[string]*KubeVolume, error)
 	for _, specVolume := range specVolumes {
 		volume, err := VolumeFromSource(specVolume.VolumeSource)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrapf(err, "failed to create volume %#v", specVolume.Name)
```
Luap99, line 119:
```suggestion
			return nil, errors.Wrapf(err, "failed to create volume %q", specVolume.Name)
```
Don't use `%#v` for strings. `%q` will add quotes.

xordspar0: Good point. Done.

Luap99 (approved):
LGTM

rhatdan:
Thanks @xordspar0 
/lgtm

PR #9842, opened 2021-03-27
Add support for env from secrets in play kube

AlbanBedel:
This serie add support for the `env.valueFrom.secretKeyRef` and `envFrom.secretRef` elements in play kube. While at it I noticed that podman treat env references as optional, but K8S treat them as mandatory unless the `optional` parameter if set true. So this serie also add support for the `optional` parameter in configMapRef and fix the behavior when it is not present to match K8S.

Note that K8S secrets are dictionaries of binary data, so the value stored in the secret must be a JSON object that contain values in base64 encoding. This match the data type of the `data` field in a K8S secret object.

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/containers/common/pkg/secrets"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v3/libpod"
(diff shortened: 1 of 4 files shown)


rhatdan:
Thanks @AlbanBedel 
Tests do not look happy.
@ashley-cui @umohnani8 @haircommander PTAL

ashley-cui (commented):
LGTM, thanks @AlbanBedel!

haircommander:
LGTM!

PR #9844, opened 2021-03-28
Add machine support for more Linux distros

afbjorklund:
Since `qemu-kvm` symlink is only available in Fedora

Fixes:

```
Error: exec: "qemu-kvm": executable file not found in $PATH
```

KVM is still hard-coded.

### DIFF
--- a/pkg/machine/qemu/options_linux_amd64.go
+++ b/pkg/machine/qemu/options_linux_amd64.go
@@ -1,11 +1,14 @@
 package qemu
 
 var (
-	QemuCommand = "qemu-kvm"
+	QemuCommand = "qemu-system-x86_64"
 )
 
 func (v *MachineVM) addArchOptions() []string {
-	opts := []string{"-cpu", "host"}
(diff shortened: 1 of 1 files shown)


afbjorklund:
Hmm, need a machine on other architecture:

```
qemu-system-aarch64: -accel kvm: No machine specified, and there is no default
```

rhatdan:
Thanks @afbjorklund 
Please squash commits, and needs [NO TESTS NEEDED]
@baude @ashley-cui PTAL

afbjorklund:
I think I will split the `arm64` out in a separate PR, it was in a very half-baked state.

i.e. all builds disabled, obsolete hardcoded path, etc

afbjorklund:
>  Please squash commits, and needs [NO TESTS NEEDED]

Ah, you mean to the actual commit itself and not just to the PR

baude:
@afbjorklund i still had this in mind to do but thanks for the PR!

PR #9850, opened 2021-03-28
An improvement for is() function in case of missing testname

aliaslanturk:
Change is an improvement and cosmetic correction. 

Previously error returning from is() would print as (FAIL line in particular):
image

Improvement changes the FIXME text as (without cosmetic corrections):
image

Finally with cosmetic corrections:
image

PersonalNote:  I think [CI:DOCS] would be required. However, I am trying to learn and adjust. Any suggestion/criticism/edit is welcomed. :)

Signed-off-by: Ali Aslanturk <ilaaslanturk@gmail.com>

### DIFF
--- a/test/system/helpers.bash
+++ b/test/system/helpers.bash
@@ -384,7 +384,11 @@ function die() {
 function is() {
     local actual="$1"
     local expect="$2"
-    local testname="${3:-FIXME}"
+    #FUNCNAME is used to determine function caller to be utilized in case of missing testname.
+    #sed substitutes are used for cosmetic corrections.	
+    local callerchain="${FUNCNAME[1]}"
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
Thanks for contributing!

@edsantiago PTAL

edsantiago:
@aliaslanturk thank you for your PR, and my humble apologies for having left that FIXME unfixed for over two years.

Your approach is clever, but it offers information (the test name) that is already known. What would you think of #9857 instead? This displays the most recently run `podman` command with its arguments, offering finer granularity into what the `is()` test is checking. This is the approach used in `buildah` system tests, and has proven fairly useful when developers forget to include a test-name argument.

edsantiago:
Closing, as #9857 has merged. @aliaslanturk thank you again for taking the time to contribute!

PR #9851, opened 2021-03-28
[CI:DOCS] Fix long option format in man pages

TomSweeneyRedHat:
Currently long options like `--tag` look to have only one long
dash instead of two dashes in the man pages on docs.podman.io.

When cut/pasted, only one dash is included, and worse yet, searches
for two dashes fail.

@eriksjolund discovered this probable fix, unfortunately it's not
easy to test without merging.

### DIFF
--- a/docs/remote-docs.sh
+++ b/docs/remote-docs.sh
@@ -70,7 +70,7 @@ function html_fn() {
         local link=$(sed -e 's?.so man1/\(.*\)?\1?' <$dir/links/${file%.md})
         markdown=$dir/$link.md
     fi
-    pandoc --ascii --lua-filter=docs/links-to-html.lua -o $TARGET/${file%%.*}.html $markdown
+    pandoc --ascii -f gfm --lua-filter=docs/links-to-html.lua -o $TARGET/${file%%.*}.html $markdown
 }
 
 # Run 'podman help' (possibly against a subcommand, e.g. 'podman help image')


TomSweeneyRedHat:
@eriksjolund PTAL

eriksjolund:
@TomSweeneyRedHat Looks good to me.
Yes, that should match what I added in my experiment:
https://github.com/containers/podman.io/issues/373#issuecomment-808705527

SweeneyTom:
LGTM
I'm going to push this through, if it doesn't work, I'll back it out.

TomSweeneyRedHat:
@rhatdan @baude @mheon please `/lgtm` this.

Luap99:
/hold
@TomSweeneyRedHat pandoc is only used to create the podman remote docs for windows.

It will not fix the problem you mention. The docs on docs.podman.io are build with sphinx. Use this to build locally:
```
cd docs
make html
```
Although this does not work correctly for me.

TomSweeneyRedHat:
@Luap99 drats, thanks for the pointer.  @eriksjolund back to the drawing board.

Luap99:
I found a possible solution. 
```
change
#### **--detach-keys**=*sequence*
to
#### **\-\-detach-keys**=*sequence*
```

Luap99:
I will create a PR. I checked that this change still renders fine on github and the linux man pages.

Issue #9251, opened 2021-02-06
Rootless API not usable right after image is deleted


marusak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

In Cockpit tests we saw one test failing way too many times so I played with it. It is timing issue, possibly some race as sleeping just one second resolves this.

**Steps to reproduce the issue:**

Prerequisites:
- When you start make sure that rootless service is not running
- Make sure that you don't have fedora:33 image (nor fedora:latest or anything which points to the same sha. `podman rmi fedora:33` in the following script needs to physically remove it, if it just untags it, then it works just fine).

1. Put the following script into file, `chmod u+x` it and run it as normal user.
```
set -x

systemctl --user enable --now podman.socket

echo "With stopped service it times out"
podman pull fedora:33
podman rmi fedora:33
timeout 5s curl --unix-socket /run/user/1000/podman/podman.sock http://d/v1.24/libpod/info

echo "Even though we never got reply from ^ the service is now running"
systemctl --user is-active podman

echo "Now I do the same, without timeout as it is going to fail"
podman pull fedora:33
podman rmi fedora:33
curl --unix-socket /run/user/1000/podman/podman.sock http://d/v1.24/libpod/info

echo "If I sleep just a tiny bit after doing `podman rmi` it works fine"
podman pull fedora:33
podman rmi fedora:33
sleep 1
curl --unix-socket /run/user/1000/podman/podman.sock http://d/v1.24/libpod/info
```

**Describe the results you received:**

The first API access just hangs and is killed by timeout.
The second API access fails.
The  third API access succeeds.

**Describe the results you expected:**

All API calls succeed.

**Output of `podman version`:**

```
[admin@m1 ~]$ podman version
Version:      3.0.0-dev
API Version:  3.0.0
Go Version:   go1.16rc1
Built:        Wed Feb  3 07:07:25 2021
OS/Arch:      linux/amd64
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-3.0.0-0.204.dev.gita086f60.fc34.x86_64
```
But this is not current regression in podman-3, we see this on current F33 as well.

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

Yes

mheon:
@baude @jwhonce PTAL

baude:
I think my results are correct?

```$ ./test.sh 
[...]

PR #9855, opened 2021-03-29
service: use LISTEN_FDS

giuseppe:
if LISTEN_FDS is specified by systemd, we need to use the first fd after the std files (so fd=3) to read from the activation socket
instead of manually opening the UNIX socket.

[NO TESTS NEEDED]

Closes: https://github.com/containers/podman/issues/9251

### DIFF
--- a/cmd/podman/system/service_abi.go
+++ b/cmd/podman/system/service_abi.go
@@ -24,16 +24,28 @@ func restService(opts entities.ServiceOptions, flags *pflag.FlagSet, cfg *entiti
 	)
 
 	if opts.URI != "" {
-		fields := strings.Split(opts.URI, ":")
-		if len(fields) == 1 {
-			return errors.Errorf("%s is an invalid socket destination", opts.URI)
+		if os.Getenv("LISTEN_FDS") != "" {
+			// If it is activated by systemd, use the first LISTEN_FD (3)
+			// instead of opening the socket file.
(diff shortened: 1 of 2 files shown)


cmd/podman/system/service_abi.go
```diff
@@ -24,16 +24,28 @@ func restService(opts entities.ServiceOptions, flags *pflag.FlagSet, cfg *entiti
 	)
 
 	if opts.URI != "" {
-		fields := strings.Split(opts.URI, ":")
-		if len(fields) == 1 {
-			return errors.Errorf("%s is an invalid socket destination", opts.URI)
+		if os.Getenv("LISTEN_FDS") != "" {
+			// If it is activated by systemd, use the first LISTEN_FD (3)
+			// instead of opening the socket file.
+			f := os.NewFile(uintptr(3), "podman.sock")
```
rhatdan, line 30: Should this be hard coded to 3 or should we use the value of LISTEN_FDS?

rhatdan (approved):

cmd/podman/system/service_abi.go
giuseppe: LISTEN_FDS says how many fds there are but we always need the first one.

It is the first FD after 0, 1, 2 so it can be hardcoded to 3

rhatdan:
LGTM

giuseppe:
@jwhonce do we need to backport this fix?

PR #9856, opened 2021-03-29
[CI:DOCS] Fix long option format on docs.podman.io

Luap99:
Escape the two dashes, otherwise they are combined into one long dash.
I tested that this change is safe and still renders correctly on github
and with the man pages.

This commit also contains a small change to make it build locally.
Assuming you have the dependencies installed you can do:
```
cd docs
make html
```
Preview the html files in docs/build/html with
`python -m http.server 8000 --directory build/html`.

Fixes containers/podman.io#373


Luap99:
@TomSweeneyRedHat @eriksjolund PTAL

mheon:
LGTM, as long as the manpages still look OK

rhatdan:
I checked the man pages.
LGTM, although this will be a PITA.
/lgtm

edsantiago:
@TomSweeneyRedHat already on my TODO list as of this morning; sorry for not having it done already (podman-buildah-bud is keeping me really busy!)

Luap99:
I had to change the  `hack/xref-helpmsgs-manpages` to make the test happy so it should already fail if you do not escape it.

edsantiago:
Submitted #9870.  @Luap99 I saw your change but was concerned that it would just anger developers, not help them. "Yes, I documented that", they will scream. With #9870 the reason for the error will be clearer. I hope.

Issue #9753, opened 2021-03-18
Podman 3.0.1 :: As a non-root user inside a Fedora-33 docker-ce container, podman(1) fails with: "Error: cannot setup namespace using newuidmap: exit status 1". [Note: This was once working] ...


nmvega:
Hello Friends:

Environment:
- `Fresh install of Fedora-33 running inside a docker-ce container`
- `podman(1) version 3.0.1 is being used (installed via dnf(1) during "docker build [...]")`
- `Default podman configuration files (beneath /etc/) have not been modified.`
- `jdoe, the non-root user, has not implemented any podman(1) configuration overrides.`

Also:
```
jdoe@docker-ce-container$ grep jdoe /etc/subuid /etc/subgid
/etc/subuid:jdoe:100000:65536
/etc/subgid:jdoe:100000:65536
```
Under the above environment and configuration, `podman(1)` use to work, but now fails. All sub-commands fail with the following error:

```
jdoe@docker-ce-container$ podman version
Error: cannot setup namespace using newuidmap: exit status 1
```
As far as I can tell, the only command that works is:
```
jdoe@docker-ce-container$ podman -v
podman version 3.0.1
```
The only thing that I can think of as far as why this once worked but now fails, is that I freshly rebuilt the enclosing docker container (after many months -- just to freshen it), and perhaps it installed a newer major version of `podman(1) v3.x`; and this gave rise to this new issue.

Note that I didn't change anything in the `Dockerfile`; I simply rebuilt it.

Any ideas friends? Thank you in advance!

mheon:
@rhatdan PTAL

rhatdan:
Could you reinstall the shadow-utils (newuidmap) within the container.  And see if it works.  For some reason the file capabilities are not set by default.

nmvega:
@rhatdan Absolutely. When I bring my cluster back up (likely tomorrow evening), I'll perform a `dnf -y reinstall shadow-utils` and let you know if that resolved the issue. If there is any other output that you would like to see (before or after), feel free to let me know. Hopefully it's something relatively simple either way. =:) Thank you.

nmvega:
@rhatdan Hello Dan:

There is positive progress after running `dnf -y reinstall shadow-utils`.

I'm now able to `pull` and `ls` images; ... operations that I couldn't previously perform. However, there's now a pesky little `WARN`ing that appears after every sub-command:

```
jdoe@docker-ce-container$ podman pull docker.io/library/mariadb:latest
[...]

PR #9858, opened 2021-03-29
[NO TESTS NEEDED] Drop Warning to Info, if cgroups not mounted

rhatdan:
If I run podman within a container without /sys/fs/cgroups mounted,
we get this warning, which the user can not do much about.  I don't
see a reason for requiring this warning to show by default, so we
should drop it to Info.

User can not do anything about it either, and code works well with
the somewhat expected error.

Helps fix: https://github.com/containers/podman/issues/9753


mheon:
LGTM

c-goes:
Thanks for this. This will be useful for WSL2 use because often there is no systemd available.

PR #9863, opened 2021-03-29
Containers prune endpoint should use only prune filters

jmguzik:
Containers endpoints for HTTP compat and libpod APIs allowed usage of list HTTP
endpoint filter funcs. Documentation in case of libpod and compat API does not allow that.
This commit aligns code with the documentation.

/kind bug


rhatdan:
LGTM
/approve
Thanks @jmguzik 

rhatdan:
@jmguzik Your prune tests are failing.

jmguzik:
> @jmguzik Your prune tests are failing.

This test is for bindings and checking filters that are not supposed to be working with the prune endpoint. I will try to adapt it.

rhatdan:
@jmguzik Your bindings test seems to be causing lots of flakes in the CI/CD System
@edsantiago FYI

pkg/bindings/test/containers_test.go
```diff
@@ -550,21 +550,28 @@ var _ = Describe("Podman containers ", func() {
...
-		pruneResponse, err = containers.Prune(bt.conn, new(containers.PruneOptions).WithFilters(filtersIncorrect))
+		pruneResponse, err := containers.Prune(bt.conn, new(containers.PruneOptions).WithFilters(filtersIncorrect))
 		Expect(err).To(BeNil())
 		Expect(len(reports.PruneReportsIds(pruneResponse))).To(Equal(0))
 		Expect(len(reports.PruneReportsErrs(pruneResponse))).To(Equal(0))
 
 		// Valid filter params container should be pruned now.
 		filters := map[string][]string{
-			"name": {"top"},
+			"until": {"0s"},
```
rhatdan, line 574: This change here is causing flakes?  Any ideas why? Tests locally this seems to work fine.

jmguzik: I hope https://github.com/containers/podman/pull/9902 will fix the issue. I shouldn't use until here. Sorry for that.

PR #9876, opened 2021-03-29
APIv2 basic test: relax APIVersion check

edsantiago:
It is tedious and error-prone to update the 'APIVersion=<exact>'
test every time there's a minor bump. Change the test so it
confirms only the major version.


rhatdan:
LGTM

PR #9878, opened 2021-03-30
[NO TESTS NEEDED] Remove semantic version suffices from API calls

jwhonce:
[NO TESTS NEEDED]

When using the bindings do not include the pre-release or build
metadata in the URL for the service. This breaks older services, while
not providing that much additional functionality.

Not sure how to test this without writing a mock service.


baude:
LGTM

PR #9881, opened 2021-03-30
Silence podman network reload erros with iptables-nft

Luap99:
Make sure we do not display the expected error when using podman network
reload. This is already done for iptables-legacy however iptables-nft
creates a slightly different error message so check for this as well.
The error is logged at info level.

[NO TESTS NEEDED] The test VMs do not use iptables-nft so there is no
way to test this. It is already tested for iptables-legacy.


baude:
lgtm

Issue #9874, opened 2021-03-29
podman-remote build: -v is a NOP


edsantiago:
It does check for invalid path:
```console
$ ../bin/podman-remote build -v /foo:/bar .
Error: error checking path "/foo": stat /foo: no such file or directory
```
...but apart from that, it seems to be a NOP. Rather than fill out a detailed issue with a custom-crafted Containerfile and logs, I'm going to take a step back and ask: what does `-v` mean in the context of podman-remote, anyway? I'm tempted to think it should be intercepted with a fatal you-seem-to-be-unclear-on-the-concept error.

rhatdan:
Well it should be hidden.

Luap99:
Yes it should work but I only tested with podman create/run and never podman build.

rhatdan:
This is now documented.

Issue #8473, opened 2020-11-25
With remote service, podman run --mount complains that path does not exist on client.


nlfiedler:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

With a remote connection defined, invoke `podman run` with the `--mount` option and it complains that the source path does not exist. True, but I am using a connection to remote host, so I would expect it to understand that the path in question is on the *remote* host, not the client.

**Steps to reproduce the issue:**

1. `podman system connection add --identity ~/.ssh/id_rsa microvm nfiedler@192.168.1.128`
2. `podman run --name=minio -d -e MINIO_ACCESS_KEY=minio -e MINIO_SECRET_KEY='keyboard cat' --publish '9000:9000' --mount type=bind,source=/mnt/tank/packs,destination=/data minio/minio --compat server /data`

**Describe the results you received:**

```
Error: error checking path "/mnt/tank/packs": stat /mnt/tank/packs: no such file or directory
```

**Describe the results you expected:**

No error reported, and the container would be running on the remote host.

**Additional information you deem important (e.g. issue happens only occasionally):**

Happens every time.

**Output of `podman version`:**

```
Client:
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Built:        Fri Sep 25 18:07:22 2020
OS/Arch:      darwin/amd64

Server:
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.13.15
Built:        Sun Oct 25 17:00:00 2020
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.16.1
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.21-1.1.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.21, commit: unknown'
  cpus: 4
  distribution:
    distribution: '"opensuse-microos"'
    version: "20201121"
  eventLogger: journald
  hostname: yubaba
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 100
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1001
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
  kernel: 5.9.8-2-default
  linkmode: dynamic
  memFree: 2565263360
[...]

PR #9882, opened 2021-03-30
Document --volume from podman-remote run/create client

rhatdan:
[NO TESTS NEEDED] This PR is mainly documentation and some code cleanup.

Also cleanup and consolidate handling of other hanlding of podman-remote
hidden options.

Fixes: #9874

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -765,11 +765,15 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	)
 	_ = cmd.RegisterFlagCompletionFunc(mountFlagName, AutocompleteMountFlag)
 
+	volumeDesciption := "Bind mount a volume into the container"
+	if registry.IsRemote() {
+		volumeDesciption = "Bind mount a volume into the container. Volume src will be on the server machine, not the client"
+	}
 	volumeFlagName := "volume"
 	createFlags.StringArrayVarP(
 		&cf.Volume,
 		volumeFlagName, "v", volumes(),
-		"Bind mount a volume into the container",
+		volumeDesciption,
 	)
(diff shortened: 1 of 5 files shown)


Luap99:
This is not correct. `podman-remote run --volume` is supported and should work. It will just not mount the host path on the client instead it will always mount the host path on the server. This should be documented but do not hide --volume from remote.

rhatdan:
Why would we want to support this?

rhatdan:
podman-remote main use case in on Macs and Windows, and remote machines, I don't see this as a common behaviour and set's bad expectations.

Luap99:
Well first there are users who use this, see https://github.com/containers/podman/issues/8473. Second, how do you provide persistent storage for your containers when you disable the --volume flag. (I know you just hide them, but still documenting that this doesn't work is wrong.)

edsantiago:
`--volume` is kind of special in that it can also refer to named volumes, which (I think?) should be able to work with podman-remote (confirmed: `podman-remote volume xxx` seems to work). I agree with @Luap99 that the flag shouldn't be marked hidden, but I tentatively disagree about path mounts: I don't think `-v /fs:/fs` is meaningful in the remote case. (BUt I'm listening, and could be convinced otherwise)

Luap99:
`-v /path:/path` is needed for remote. You can have only one storage location with named volumes. When your container needs data from different partitions you cannot use named volumes, also you might want to mount files from the host into the container.
Docker works exactly the same.
The `--volume` is without a doubt one of the most used flags for run/create. Hiding or removing it will confuse users even more and saying it is not supported is just wrong.

rhatdan:
Beaten into submission.

rhatdan:
I changed the podman-remote help to reflect this information as well.

./bin/podman-remote run --help | grep -- -v,
  -v, --volume stringArray                       Bind mount a volume into the container. Volume src will be on the server machine, not the client

TomSweeneyRedHat:
LGTM

PR #9887, opened 2021-03-30
buildah bud tests under podman-remote

edsantiago:
A huge ton of skips: podman-remote build just doesn't
really work at all. Lots of issues filed, a few more
left to go. Submitting this just to see how it does in CI.


edsantiago:
--arch/--os/--platform don't work. --log-rusage either. Will file issues for those later today.

rhatdan:
@edsantiago Any issues still blocking this?

edsantiago:
Yes. #10154 ("stream dropped") causes 3-5 flakes per run; we can't even dream of merging until that issue is fixed.

edsantiago:
OMG it passed!! (with #11067)

edsantiago:
@containers/podman-maintainers, @containers/build-maintainers, I believe this is ready for review. It's more complicated than I would've liked: the big diffs to `helpers.bash` are more likely to cause conflicts each time we vendor in a new buildah. I don't have any good ideas on how to solve that.

Despite the complexity, I think this is worth having. This remote test has already caught a large number of podman-remote bugs, and I think it's likely to catch more in the future.

vrothberg (commented):
LGTM
Thanks for doing this, Ed! Quite a herculean task.

vrothberg:
@edsantiago please drop the `WIP: ` if you want to merge

rhatdan:
LGTM
Is there stuff we could do on the Buildah side to make the breaking vendors less likely?

edsantiago:
The only thing to do, really, is be careful when touching `helpers.bash`. But that's really hard, because real problems come up in buildah-land that can only be addressed by tweaking that file.

Where my mind keeps going is: eliminate the diff, put the podman and podman-remote support directly in the buildah repo itself, but that just seems loathsome: buildah really shouldn't know about the kludges done in podman CI. I'd like to do it by some hook- or plugin-based mechanism, but my explorations there just lead to more complexity.

PR #9892, opened 2021-03-30
Trim white space from /top endpoint results

jwhonce:
Versions of the ps command have additional spaces between fields, this
manifests as the container asking to run "top" and API reporting "top "
as a process.

Endpoint and tests updated to check that "top" is reported.

There is no libpod specialized endpoint to update.


mheon:
LGTM

PR #9895, opened 2021-03-30
Don't relabel volumes if running in a privileged container

rhatdan:
Docker does not relabel this content, and openstack is running
containers in this manner.  There is a penalty for doing this
on each container, that is not worth taking on a disable SELinux
container.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -376,8 +376,14 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 			case "z":
 				fallthrough
 			case "Z":
-				if err := label.Relabel(m.Source, c.MountLabel(), label.IsShared(o)); err != nil {
-					return nil, err
+				if c.MountLabel() != "" {
+					if c.ProcessLabel() != "" {
+						if err := label.Relabel(m.Source, c.MountLabel(), label.IsShared(o)); err != nil {
+							return nil, err
+						}
+					} else {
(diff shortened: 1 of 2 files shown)


rhatdan:
@tunacado PTAL

libpod/container_internal_linux.go
```diff
@@ -380,8 +380,14 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
...
 			case "Z":
-				if err := label.Relabel(m.Source, c.MountLabel(), label.IsShared(o)); err != nil {
-					return nil, err
+				if c.MountLabel() != "" {
+					if c.ProcessLabel() != "" {
+						if err := label.Relabel(m.Source, c.MountLabel(), label.IsShared(o)); err != nil {
+							return nil, err
+						}
+					} else {
+						logrus.Infof("Ignoring label request on volumes for label disabled containers %q", m.Source)
```
mheon, line 389: Suggest reword to `"Not relabeling volume %q in container %s as SELinux is disabled", m.Source, c.ID()`

TomSweeneyRedHat:
LGTM once @mheon's comment is addressed.

tunacado:
LGTM

TomSweeneyRedHat:
@rhatdan unhappy red tests.

rhatdan:
@tunacado @dustymabe Has pointed out that this breaks compatibility with Docker, so I need to revert this PR.

PR #9896, opened 2021-03-30
Handle podman-remote --arch, --platform, --os

rhatdan:
Podman remote should be able to handle remote specification of
arches.

Requires: https://github.com/containers/buildah/pull/3116

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/checkpoint-restore/go-criu v0.0.0-20190109184317-bdb7599cd87b
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
-	github.com/containers/buildah v1.20.0
+	github.com/containers/buildah v1.20.1-0.20210402144408-36a37402d0c8
 	github.com/containers/common v0.35.4
 	github.com/containers/conmon v2.0.20+incompatible
 	github.com/containers/image/v5 v5.10.5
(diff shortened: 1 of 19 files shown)


rhatdan:
@edsantiago PTAL

edsantiago:
Something not quite right: it's not working with just plain `--os`, no other flags:
```console
$ echo "FROM scratch" | ./bin/podman-remote build -t foo --os=windows -
Error: failed to parse query parameter 'platform': "windows": invalid platform syntax for "windows" (use OS/ARCH)
```
It works fine if I append `--arch=foo` or if I use `--platform=foo/windows`.

rhatdan:
@edsantiago `$ echo "FROM scratch" | ./bin/podman-remote build -t foo --os=windows -`
Should work now.

rhatdan:

edsantiago:
Confirmed: passes buildah-bud tests (on my local setup. They won't pass in this PR due to the `tag-YMD-SHA` vendor string, which I don't know how to grok.

rhatdan:
@edsantiago I am going to cut a Buildah test so we can get this to pass.

edsantiago:
@rhatdan try rebasing on #9943 to test the custom buildah version. Unfortunately, you will need to do some manual conflict resolution, which means you will need to be the guinea pig for testing  https://github.com/containers/podman/blob/626b7ebd74dbd2614c4e6e9f99514404c50e806a/test/buildah-bud/README.md

edsantiago:
(Spoiler alert: whether you cut a new buildah, or rebase on my PR, there are going to be conflicts because of the buildah system-test PR I merged last week. I will gladly help with the conflict resolution, but I hope someone else does the brunt of the work so we can test my documentation).

edsantiago:
I don't see any changes to `test/buildah-bud/buildah-tests.diff`, which (I suspect) means you didn't read and follow `test/buildah-bud/README.md`, which (I suspect) means this PR is going to fail in the `bud` step with a merge conflict.

edsantiago:
>error: corrupt patch at line 130

Sigh.

edsantiago:
I got past that hurdle by manually editing line 134:
```
from:  @@ -2729,1 +2733,1 @@ _EOF     <--- '1' makes no sense! How did that happen?
to:    @@ -2729,6 +2733,7 @@ _EOF
```
...but then ran into a different snag, and I have to give up for the day. @rhatdan I'm sorry, obviously my code is not ready for prime time. I will spend more effort polishing it up this week.

edsantiago:
Fixed in #9962.

PR #9898, opened 2021-03-30
[CI:DOCS] Makefile: Fix make install.docker regression

Foxboron:
This should fix the fact that `make install.docker` would fail as `$(BINDIR)` was never created after splitting up the routine in 3908c00799fe2af1a12c9c4f4be8b49dbdecd9be.

I also introduced a `install.docker-full` to install the binary and documentation. But I can remove that commit if upstream thinks it's not needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -558,6 +558,9 @@ install.docker-docs-nobuild:
 .PHONY: install.docker-docs
 install.docker-docs: docker-docs install.docker-docs-nobuild
 
+.PHONY: install.docker-full
+install.docker-full: install.docker install.docker-docs
+
 .PHONY: install.systemd
 ifneq (,$(findstring systemd,$(BUILDTAGS)))
 install.systemd:
--- a/Makefile
+++ b/Makefile
@@ -545,13 +545,14 @@ install.cni:
 
 .PHONY: install.docker
 install.docker:
+	install ${SELINUXOPT} -d -m 755 $(DESTDIR)$(BINDIR)
 	install ${SELINUXOPT} -m 755 docker $(DESTDIR)$(BINDIR)/docker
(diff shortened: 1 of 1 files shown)


mheon:
Edited title to set CI:DOCS. Can you force-push to the branch? That should fix CI.

Foxboron:
Thanks! Found a unexpected error while actually testing out the patches (:upside_down_face:) I'll do a force push when I figure it out.

Foxboron:
I think 61e626c859a60b19419a858b15e76cd4881e24c3 either introduced a race-condition or made it more obvious. Locally it works fine while testing because of `-j1` but `-j2` is going to race the manpage building and installation.

<pre>
λ libpod » make -j2 DESTDIR=./tmp-podmandocker install.docker-docs
if [ ! -x "/usr/bin/go-md2man" ]; then \
		env GO111MODULE=off go get -u github.com/cpuguy83/go-md2man; \
fi
mkdir -p docs/build/man
install  -d -m 755 ./tmp-podmandocker/usr/local/share/man/man1
install  -m 644 docs/build/man/docker*.1 -t ./tmp-podmandocker/usr/local/share/man/man1
install: cannot stat 'docs/build/man/docker*.1': No such file or directory
make: *** [Makefile:555: install.docker-docs-nobuild] Error 1
</pre>

I love Makefiles :disappointed:

Foxboron:
The Makefile for podman isn't really in a great state. There is a *lot* of workarounds simply because most build targets does not know their input/output files which leads to things being rerun for no good reason. I tried fixing up parts of this but it's going to be a lot of work which I'm personally not motivated for. 

I'll package `podman` for Arch Linux with `-j1` wherever needed and I think it would probably be a good idea to write on the README that the Makefile might not work when specifying multiple jobs.

mheon:
LGTM
/approve

mheon:
Can you file an issue about the makefile bug? I'm not a makefile expert myself, but having this documented will encourage us to fix it.

PR #9902, opened 2021-03-31
Fix bindings prune containers flaky test

jmguzik:
In #9863 prune containers filter params were narrowed to support only those
required by http API. name filter in bindings was replaced by until filter,
which is not a good match, as until filters are causing tests to be flaky.

### DIFF
--- a/pkg/bindings/test/containers_test.go
+++ b/pkg/bindings/test/containers_test.go
@@ -568,15 +568,6 @@ var _ = Describe("Podman containers ", func() {
 		Expect(err).To(BeNil())
 		Expect(len(reports.PruneReportsIds(pruneResponse))).To(Equal(0))
 		Expect(len(reports.PruneReportsErrs(pruneResponse))).To(Equal(0))
-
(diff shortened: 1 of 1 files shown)


jmguzik:
@rhatdan FYI

rhatdan:
That works, but I would love to get a working tests in.
LGTM to fix flakes for now.
/approve

jmguzik:
And unitil is added by me, previously name filter was here.

vrothberg:
@edsantiago may have an idea how to make it pass

jmguzik:
@vrothberg my alternative idea for sleep until would be something like 0ns or 1ns and keeping fingers crossed that it is enough.
I would suggest however to remove this one and I can experiment further once @edsantiago will provide input.

edsantiago:
@vrothberg @jmguzik sorry, this is way above my head. I don't know ginkgo at all.

rhatdan:
/approve
/lgtm
Lets work on this in a separate PR.  For now I want to fixup CI so tests can merge.

PR #9904, opened 2021-03-31
podman machine shell completion

Luap99:
Add shell completion for machine names.

[NO TESTS NEEDED]
I would like to add one to the shell completion test however
using podman machine init is to expensive.


Luap99:
@baude @ashley-cui PTAL

Luap99:
Updated to fix the cross compile.

ashley-cui (commented):
LGTM

Issue #9459, opened 2021-02-21
[3.0.1] podman build started failing with "runc: executable file not found in $PATH" although runc is in a recognized location


riyad:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman build ...` with a Containerfile started failing with `exec: "runc": executable file not found in $PATH`.
Indeed `runc` is not in $PATH but in one of the locations listed in the `engine.runtimes.runc` config option. `podman info` picks it up correctly (see below).

* Running `buildah bud ...` (with runc configured) works correctly (buildah 1.19.6)
* Changing the runtime to crun (which also happens to be in $PATH) makes it work
* Installing Podman 3.0.0 (i.e. downgrading from 3.0.1) makes it work again

**Steps to reproduce the issue:**

1. Make sure podman uses runc

2.
```shell
cat > Containerfile <<EOF
FROM alpine

RUN echo "does this work?"
EOF
```

3.
 ```shell
$ podman build -f Containerfile 
STEP 1: FROM alpine
STEP 2: RUN echo "does this work?"
error running container: error creating container for [/bin/sh -c echo "does this work?"]: : exec: "runc": executable file not found in $PATH
Error: error building at STEP "RUN echo "does this work?"": error while running runtime: exit status 1
```

**Describe the results you received:**

`podman build ...` fails with an unexpected error:
 ```shell
$ podman build -f Containerfile 
STEP 1: FROM alpine
STEP 2: RUN echo "does this work?"
error running container: error creating container for [/bin/sh -c echo "does this work?"]: : exec: "runc": executable file not found in $PATH
Error: error building at STEP "RUN echo "does this work?"": error while running runtime: exit status 1
```

**Describe the results you expected:**

Should build a container as it did before and as it does with buildah
```shell
$ podman build --no-cache -f Containerfile         
STEP 1: FROM alpine
STEP 2: RUN echo "does this work?"
does this work?
STEP 3: COMMIT
--> 80dd1ca0a2c
80dd1ca0a2cab2b15edb1464ab12362dff495048f5cf57402abd8dbeea326948
```
```shell
$ buildah bud --no-cache -f Containerfile
STEP 1: FROM alpine
STEP 2: RUN echo "does this work?"
does this work?
STEP 3: COMMIT
Getting image source signatures
Copying blob cb381a32b229 skipped: already exists  
Copying blob 2c456cc9967d done  
Copying config ed22a73939 done  
[...]

PR #9906, opened 2021-03-31
Should send the OCI runtime path not just the name to buildah

rhatdan:
[NO TESTS NEEDED] Mainly because I have no idea how we would test
this.

Fixes: https://github.com/containers/podman/issues/9459

### DIFF
--- a/libpod/runtime_img.go
+++ b/libpod/runtime_img.go
@@ -168,9 +168,7 @@ func (r *Runtime) newImageBuildCompleteEvent(idOrName string) {
 // Build adds the runtime to the imagebuildah call
 func (r *Runtime) Build(ctx context.Context, options buildahDefine.BuildOptions, dockerfiles ...string) (string, reference.Canonical, error) {
 	if options.Runtime == "" {
-		// Make sure that build containers use the same runtime as Podman (see #9365).
-		conf := util.DefaultContainerConfig()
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

libpod/runtime_img.go
```diff
@@ -168,9 +168,7 @@ func (r *Runtime) newImageBuildCompleteEvent(idOrName string) {
 // Build adds the runtime to the imagebuildah call
 func (r *Runtime) Build(ctx context.Context, options buildahDefine.BuildOptions, dockerfiles ...string) (string, reference.Canonical, error) {
 	if options.Runtime == "" {
-		// Make sure that build containers use the same runtime as Podman (see #9365).
-		conf := util.DefaultContainerConfig()
-		options.Runtime = conf.Engine.OCIRuntime
+		options.Runtime = r.GetOCIRuntimePath()
```
TomSweeneyRedHat, line 171: Could the default runtime ever differ from what's set in the containerConfig structure?  If so, I'm not sure we'd want to do this.

TomSweeneyRedHat:
One question, if I'm off base with it, then LGTM.

libpod/runtime_img.go
rhatdan: They are stored in the same structure.  We read it originally from the containers.conf, but if the user overrides it, we replace the value in the same structure.

Issue #9806, opened 2021-03-24
Docker Compose CLI is not supported (missing top level context CLI command)


troyready:
/kind feature

**Description**

The [Docker Compose CLI](https://github.com/docker/compose-cli) (NOTE: this is related to but *not* docker-compose) sits in front of the traditional docker command and extends docker-compose capability to "cloud" deployments. Essentially, `docker-compose up` to test locally, `docker compose up` to deploy to (e.g.) AWS ECS.

It relies on a [context command](https://docs.docker.com/engine/reference/commandline/context/) (added in docker v19 from what I can see). Additional info on it is [available here](https://docs.docker.com/engine/context/working-with-contexts/).

The overall workflow is [fairly well walked through here.](https://aws.amazon.com/blogs/containers/deploy-applications-on-amazon-ecs-using-docker-compose/)

**Steps to reproduce the issue:**

1. Unpack [docker compose-cli](https://github.com/docker/compose-cli) into PATH

2. Symlink `com.docker.cli` to `podman`

3. Execute `docker --context default ps`

**Describe the results you received:**

`Error: unknown flag: --context`

**Describe the results you expected:**

List of containers

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.15.8
Built:        Wed Dec 31 16:00:00 1969
OS/Arch:      linux/amd64
```

vrothberg:
Thanks for reaching and apologies for the longer silence; we've been really busy.  The `--context` option along with the `context` commands relate to Docker Swarm which is something Podman does not support.

In your specific case `--context=default` is something that Podman could support (i.e., default context).  I also think that Podman could do a better job at error reporting and report that `docker context` isn't supported because of Swarm.

@containers/podman-maintainers PTAL

PR #9907, opened 2021-03-31
Add support for podman --context default

rhatdan:
This is a noop but helps with scripting and docker-compose.

Fixes: https://github.com/containers/podman/issues/9806

### DIFF
--- a/cmd/podman/root.go
+++ b/cmd/podman/root.go
@@ -180,6 +180,10 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
 		os.Setenv("TMPDIR", "/var/tmp")
 	}
 
+	context := cmd.Root().LocalFlags().Lookup("context")
+	if context.Value.String() != "default" {
+		return errors.New("Podman does not support swarm, the only --context value allowed is \"default\"")
+	}
 	if !registry.IsRemote() {
 		if cmd.Flag("cpu-profile").Changed {
 			f, err := os.Create(cfg.CPUProfile)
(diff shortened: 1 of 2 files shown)


cmd/podman/root.go
```diff
@@ -269,6 +269,9 @@ func rootFlags(cmd *cobra.Command, opts *entities.PodmanConfig) {
 	lFlags.StringVar(&opts.URI, urlFlagName, uri, "URL to access Podman service (CONTAINER_HOST)")
 	_ = cmd.RegisterFlagCompletionFunc(urlFlagName, completion.AutocompleteDefault)
 
+	// Context option added just for compatibily with DockerCLI.
+	lFlags.String("context", "", "Name of the context to use to connect to the daemon (This flag is a NOOP and provided solely for scripting compatibility.)")
```
vrothberg, line 273: default should be "default"

vrothberg (commented):
I think we need to check the value of and reject anything else than "default" with an error indicating that we don't support Swarm

cmd/podman/root.go
rhatdan: I actually had that originally but was not crazy about it since it showed on the help message.  But I can add it back.

TomSweeneyRedHat:
All kinds of unhappy tests.

cmd/podman/root.go
vrothberg: I think we should hide the flag and not document it.  It doesn't add value since it only accepts one value.  But I think it adds great value for cases as in the initial issue.

rhatdan: SGTM

vrothberg (commented):
LGTM, thanks!

ashley-cui (commented):
/lgtm

PR #9909, opened 2021-03-31
[CI:DOCS] Add transport and destination info to manifest doc

w4tsn:
Initially I was missing transport information on podman manifest add.
Especially the `containers-storage` transport which references the
local image store. Had a use case where this came in quite handy and it
is not stated anywhere else in the docs. Suppose it does not make sense
for podman pull & push.

I've only added containers-storage and docker transports for
manifest add since I know those work. Maybe others work too.

I then also added the destination section to manifest push as it is done
in podman push & pull. I've added all transports here, but I don't know
if all are supported. Please review.

I feel the podman manifest documentation is a bit *underwhelming*. I hope
this is a step to improve this situation also for others.

### DIFF
--- a/docs/source/markdown/podman-manifest-add.1.md
+++ b/docs/source/markdown/podman-manifest-add.1.md
@@ -4,7 +4,7 @@
 podman\-manifest\-add - Add an image to a manifest list or image index
 
 ## SYNOPSIS
-**podman manifest add** [*options*] *listnameorindexname* *imagename*
+**podman manifest add** [*options*] *listnameorindexname* [*transport*]:*imagename*
 
 ## DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


docs/source/markdown/podman-manifest-add.1.md
```diff
@@ -79,6 +79,20 @@ Specify the variant which the list or index records for the image.  This option
 is typically used to distinguish between multiple entries which share the same
 architecture value, but which expect different versions of its instruction set.
 
+## Transport
+
+ Multiple transports are supported:
```
rhatdan, line 84: What one is the default?

TomSweeneyRedHat: I believe it's a [registry](https://github.com/containers/podman/blob/00792f8c95f804c446f8ae8e0e105a240074198e/docs/source/markdown/podman-push.1.md#destination)

w4tsn: I just verified this. Building an image locally with localhost/myImage, then trying to add it with `manifest add mylist:v1 localhost/myImage` throws an error that `docker://localhost/myImage` yields in a connection refused on `https://localhost/v2/`

rhatdan: Sorry, I meant the documentation should identify the the default transport.

rhatdan:
Thanks @w4tsn 
LGTM
/approve

TomSweeneyRedHat:
LGTM

Issue #9852, opened 2021-03-29
"Error: lock … is not a read-only lock" failure with `additionalimagestores` set


srcshelton:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Possibly related to #7309, if I have `additionalstorage` set in `storage.conf` I always get, e.g.:

```
$ sudo podman image pull --root /space/podman/images helloworld:latest 
Error: lock "/space/podman/images/overlay-images/images.lock" is not a read-only lock
```

(and the same with `--storage-opt=overlay.mountopt=nodev` also specified, as suggested in the above issue)

I've tried creating `images.lock` (and `layers.lock`) as empty files, and also copying over the (~64byte?) files from the default storage location.

If I temporarily edit `containers.conf` to swap the `graphRoot` and `additionalstorage` paths and then re-run the failing command (which is actually:

```
sudo podman container commit sys-build sys-build:latest
```

) and remove the `--root` option then the operation succeeds, but in the original default storage location rather than the original additional storage location.

What is the correct process to follow if I wish to maintain multiple storage locations to differentiate between (effectively) ephemeral images and images which should have longer life-spans?

**Output of `podman version`:**

```
Version:      3.1.0-rc2
API Version:  3.1.0-rc2
Go Version:   go1.16.2
Git Commit:   1b56ea2d9df82cbba2679f646c077881fefb49d6
Built:        Sat Mar 27 00:25:10 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.8
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: app-emulation/conmon-2.0.27
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: 65fad4bfcb250df0435ea668017e643e7f462155'
  cpus: 8
  distribution:
    distribution: gentoo
    version: unknown
  eventLogger: file
  hostname: dellr330
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.11.10-gentoo
  linkmode: dynamic
  memFree: 2663763968
  memTotal: 33390899200
  ociRuntime:
    name: crun
    package: app-emulation/crun-0.18
    path: /usr/bin/crun
    version: |-
      crun version 0.18
      commit: 808420efe3dc2b44d6db9f1a3fac8361dde42a95
      spec: 1.0.0
[...]

PR #9911, opened 2021-03-31
Allow users to override default storage opts with --storage-opt

rhatdan:
We define in the man page that this overrides the default storage
options, but the code was appending to the existing options.

This PR also makes a change to allow users to specify --storage-opt="".
This will turn off all storage options.

https://github.com/containers/podman/issues/9852

### DIFF
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -148,7 +148,7 @@ specify additional options via the `--storage-opt` flag.
 
 #### **\-\-storage-opt**=*value*
 
-Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all.
+Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all. If you specify --storage-opt="", no storage options will be used.
 
 #### **\-\-syslog**=*true|false*
 
--- a/test/system/005-info.bats
+++ b/test/system/005-info.bats
@@ -53,4 +53,13 @@ store.imageStore.number   | 1
 
 }
 
+@test "podman info --storage-opt='' " {
+    skip_if_remote "--storage-opt flag is not supported for remote"
+    skip_if_rootless "storage opts are required for rootless running"
+    run_podman --storage-opt='' info
+    # Note this will not work in rootless mode, unless you specify
+    # storage-driver=vfs, until we have kernels that support rootless overlay
+    # mounts.
+    is "$output" ".*graphOptions: {}" "output includes graphOptions: {}"
+}
 # vim: filetype=sh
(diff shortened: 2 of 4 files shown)


rhatdan:
@srcshelton PTAL

test/system/005-info.bats
```diff
@@ -53,4 +53,8 @@ store.imageStore.number   | 1
 
 }
 
+@test "podman info --storage-opt="" " {
```
edsantiago, line 56: Suggestion: change "" (doublequote doublequote) to '' (singlequote singlequote) for log readability.

test/system/005-info.bats
```diff
@@ -53,4 +53,8 @@ store.imageStore.number   | 1
...
+    run_podman --storage-opt="" info
```
edsantiago, line 57: This completely blows up on my laptop (rootless, that is; works fine as root of course):
```
Error: kernel does not support overlay fs: 'overlay' is not supported over btrfs at "/home/esm/.local/share/containers/storage/overlay": backing file system is unsupported for this graph driver
```

docs/source/markdown/podman.1.md
```diff
@@ -148,7 +148,7 @@ specify additional options via the `--storage-opt` flag.
 
 #### **\-\-storage-opt**=*value*
 
-Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all.
+Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all. If you specify the --storage-opt="", no storage options will be used.
```
edsantiago, line 151: Unnecessary `the`

TomSweeneyRedHat:
Unhappy red tests.
LGTM once @edsantiago 's comments are addressed.

srcshelton:
> @srcshelton PTAL

So, to confirm, this replaces the `--root` argument, or complements it?

If `storage.conf` defines the graphRoot directory as `/a` and an additionalStorage location of `/b`, then what command-line/environment variable configuration would I need to be able to pull or build an image directly into directory`/b`?

rhatdan:
Complements it.  Basically --root follows the settings in storage.conf.  --storage-opt will ignore the storage.conf settings.

srcshelton:
> Complements it. Basically --root follows the settings in storage.conf. --storage-opt will ignore the storage.conf settings.

I'm afraid that I'm not sure I understand how these two options are intended to be used… would be be able to give an example command-line which either pulls or builds a container definition into a directory declared as an additionalStorage location but which it not the main graphRoot?

`podman --root /path/to/additional/storage --storage-opt="" pull image:tag` or similar?

(Is the intent that `storage-opt` can be used without a value in order to clear the `storage.conf` settings, and then `root` causes the additionalStorage directory - if specified - to be the location the image is written to?)

test/system/005-info.bats
rhatdan: Fixed

rhatdan:
@srcshelton Yes, exactly you use case.

rhatdan:

srcshelton:
Is there any use-case for `--root` which doesn't require `--storage-opt=""` to clear any pre-existing configuration (… perhaps running without any `storage.conf` whereby everything has to be passed on the command-line)?

Should `podman` then implicitly clear any defined `storage` options if `--root` is used, even if `--storage-opt` isn't specified?

PR #9912, opened 2021-04-01
Recreate until container prune tests for bindings

jmguzik:
#9902 follow-up.
/kind bug
**Please double-check that, but it seems until filter was wrongly implemented.**

Docker API https://docs.docker.com/engine/api/v1.41/#operation/ContainerPrune states:
```
until=<timestamp> Prune containers created before this timestamp. 
The <timestamp> can be Unix timestamps, date formatted
timestamps, or Go duration strings (e.g. 10m, 1h30m) computed 
relative to the daemon machine’s time.
```
More or less the same can be found in podman API. In the code, after was used instead of until. That caused tests to flake heavily.
If you approve the code change, I will look for similar errors, because until is probably badly implemented in other places.

Added also test for list containers with until filter. Dates are now set as Unix timestamp way in future and in past, so tests should not flake.


mheon:
Ooops. Good catch.

PR #9917, opened 2021-04-01
fix machine naming conventions

baude:
try to align the machine commands and their usage descriptions.

[NO TESTS NEEDED]

### DIFF
--- a/cmd/podman/machine/list.go
+++ b/cmd/podman/machine/list.go
@@ -28,7 +28,7 @@ var (
 		Use:     "list [options]",
 		Aliases: []string{"ls"},
 		Short:   "List machines",
-		Long:    "List Podman managed virtual machines.",
+		Long:    "List managed virtual machines.",
 		RunE:    list,
 		Args:    validate.NoArgs,
 		Example: `podman machine list,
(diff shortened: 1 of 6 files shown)


cmd/podman/machine/list.go
```diff
@@ -28,7 +28,7 @@ var (
 		Use:     "list [options]",
 		Aliases: []string{"ls"},
 		Short:   "List machines",
-		Long:    "List Podman managed virtual machines.",
+		Long:    "List machines.",
```
TomSweeneyRedHat, line 31: @rhatdan @baude, I don't understand why you're reverting this.

baude: i believe @rhatdan preferred the more direct use

rhatdan: I am fine with the long discription. as managed virtual machines.  Better to have the long description to have more data then the short one.

rhatdan: Although I agree with removing Podman.

rhatdan: There should be consistency though, so above should be "Initialize a virtual machine"

rhatdan:
LGTM

baude:
merge me

Issue #9280, opened 2021-02-09
rootless service not work with systemd socket activation


pendulm:
/kind bug

**Description**

podman system service works broken in rootless mode, not get socket from systemd user session, and a lot of noise fill in the system log because system service restart continuously. 

**Steps to reproduce the issue:**

Run command as a **non-root user**, so podman  run in rootless mode

1.  $ systemctl --user start podman.socket

2. $ nc -U /run/user/$UID/podman/podman.sock

3. $ journalctl --user -f  watch the log

**Describe the results you received:**
`level=info msg="using API endpoint: 'unix:/run/user/1000/podman/podman.sock'"` in the log, which mean socket activation not works, and systemd pull up podman service all the time

**Describe the results you expected:**
we should see `level=info msg="using systemd socket activation to determine API endpoint"` in the log

**Additional information you deem important (e.g. issue happens only occasionally):**
run podman system service in root mode is OK, service get unix socket correctly from systemd

**Output of `podman version`:**

```
Version:      3.0.0-dev
API Version:  3.0.0
Go Version:   go1.16rc1
Built:        Wed Feb  3 12:07:25 2021
OS/Arch:      linux/amd64
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
podman-3.0.0-0.204.dev.gita086f60.fc34.x86_64
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide?**

Yes

**Additional environment details (AWS, VirtualBox, physical, etc.):**
System: Fedora rawhide

afbjorklund:
I think this is normal, since the podman service has a 5 second timeout by default.

`podman system service --timeout 5000`

pendulm:
I dig into this bug further, and I found the root in [pkg/systemd/activation.go](https://github.com/containers/podman/blob/19507d0ffe8cda0a69f056838556f471fd9e61fa/pkg/systemd/activation.go#L10)

In rootless mode, podman fork&exec a podman process in new user-namespace,  and systemd pass LISTEN_PID as pid of the parent, so this check breaks:
```go
	if err != nil || p != os.Getpid() {
		return false
	}
```

But in root mode,  no subprocess be forked, so socket activation works

pendulm:
> I think this is normal, since the podman service has a 5 second timeout by default.
> 
> `podman system service --timeout 5000`
> 
> EDIT: Never mind, you meant that it fails to hand over the socket _within_ that timeout ?

[...]

Issue #9484, opened 2021-02-23
[RFE] bindings.NewConnection should implement a "retry"


cjeanneret:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**
Hello there,
From time to time, when calling the ```bindings.NewConnection```, it fails because the socket is not temporary not available:
```
cannot connect to the Podman socket, please verify that Podman REST API service is running: Get "http://d/v2.0.0/libpod/_ping": dial unix /run/user/1000/podman/podman.sock: connect: no such fi
le or directory
```
When I re-run my app right after this failure, it's all green. If the NewConnection could implement a kind of retry with a configurable timeout (or number of retries), that would be really neat.
Of course, my own app could implement it - but maybe this could also interest other users.

Thank you!

Cheers,

C.

vrothberg:
Thanks, @cjeanneret!

@jwhonce, would some kind of exponential back off make sense?

cjeanneret:
In the meanwhile, I'm considering using some 3rd party retry lib, such as github.com/avast/retry-go
I'm already using it for another part in my code, so it's "ok". Still - apparently the podman socket seems to be flapping, so getting this as an embedded feature would be great :).

rhatdan:
This issue is caused by mishandling of socket activation, I believe and I think we have a PR to fix this.

PR #9928, opened 2021-04-03
Fix rootless socket activation

pendulm:
Move socket activation check into init() and set global condition, So rootless setup could use this condition in parent and child, child podman should adjust LISTEN_PID to its self PID.  

This fix test for following scenarios:
1) no container and pause process, start system service in a new namespace
2) join existed pause namespace, start system service
3) join existed container namespace while pause process die,  start system service

when use Shortcut re-exec, I think the logic is still OK.

Closes: #9280

Signed-off-by: pendulm <lonependulm@gmail.com>

### DIFF
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -61,6 +61,10 @@ static int open_files_max_fd;
 static fd_set *open_files_set;
 static uid_t rootless_uid_init;
 static gid_t rootless_gid_init;
+static bool do_socket_activation = false;
+static char *saved_systemd_listen_fds;
+static char *saved_systemd_listen_pid;
+static char *saved_systemd_listen_fdnames;
 
 static int
 syscall_setresuid (uid_t ruid, uid_t euid, uid_t suid)
@@ -242,6 +246,10 @@ static void __attribute__((constructor)) init()
 {
   const char *xdg_runtime_dir;
   const char *pause;
+  const char *listen_pid;
+  const char *listen_fds;
+  const char *listen_fdnames;
+
   DIR *d;
 
   pause = getenv ("_PODMAN_PAUSE");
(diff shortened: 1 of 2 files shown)


rhatdan:
Thanks @pendulm 

Could you squash your commits into one commit, and then make sure there is a line break between the heading
```
Fix rootless socket activation

...
```
Then tests should pass validation.

@giuseppe PTAL

pendulm:
@rhatdan modified according to your sugguestion.

pkg/rootless/rootless_linux.c
```diff
@@ -293,6 +301,24 @@ static void __attribute__((constructor)) init()
...
+
+    if (listen_pid != NULL && listen_fds != NULL && strtol(listen_pid, NULL, 10) == getpid())
+      {
+        size_t len_pid = strlen(listen_pid) + 1;
+        size_t len_fds = strlen(listen_fds) + 1;
+        size_t len_fdnames = strlen(listen_fdnames) + 1;
+
+        // save systemd socket environment for rootless child
+        do_socket_activation = true;
+        char * save_buf = (char *) calloc (len_pid + len_fds + len_fdnames, 1);
```
giuseppe, line 316: can we add a check for `calloc` failures?

giuseppe: also there is no need to use a single buffer here.  Let's split it to multiple assignments and use `strdup` instead. Each of the `strdup` must be checked for errors.

giuseppe (commented):
thanks for the patch.  Just left some comments

rhatdan:
@baude @ashley-cui This might be the issue you were seeing in the podman-machine testing.

baude:
LGTM

rhatdan:
I don't believe the calloc checks were ever added. @pendulm could you open a PR to fix @giuseppe suggestions?

mheon:
Oh - sorry. I thought @giuseppe added those as review comments on the PR, and since I didn't see any I assumed they'd been fixed.

pendulm:
@rhatdan sorry for comment lately. I followed @giuseppe's second suggestion, use `strdup` to eliminate  `calloc` and check for errors, so no further PR is needed.

rhatdan:
Fixed: https://github.com/containers/podman/issues/9484

rhatdan:
Ok great.

PR #8787, opened 2020-12-20
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/podman/commit/04624679014630f9d7d486be6262faac1fcacc39#commitcomment-45284288

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/podman/commit/83328c1aa180d27ccc45ae030679eca30274f212

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

---
I understand that changes.txt is generated from commit messages, but commit messages have typos, which make them fairly misleading. It's trivial for me to drop changes for this file if requested (or any other file(s)).

PR #9932, opened 2021-04-03
[CI:DOCS] Fix spelling rootfull -> rootful

eriksjolund:
If I understand correctly the spelling "rootful" should be prefered over "rootfull" (see
https://github.com/containers/podman/pull/8787#issuecomment-749026276).

This PR fixes
./docs/tutorials/basic_networking.md

There is one file that still has the wrong spelling 
./vendor/github.com/containers/common/pkg/config/containers.conf

```
[esjolund@laptop podman]$ grep rootfull ./vendor/github.com/containers/common/pkg/config/containers.conf
#       rootfull "unix://run/podman/podman.sock (Default)
#       remote rootfull ssh://root@10.10.1.136:22/run/podman/podman.sock
[esjolund@laptop podman]$ 
```
But I leave those out because they are in vendored code.

The recommendation "_Do not change vendored code._" can be seen here:
https://github.com/containers/podman/pull/8787#discussion_r546678504

Signed-off-by: Erik Sjölund <erik.sjolund@gmail.com>

### DIFF
--- a/docs/tutorials/basic_networking.md
+++ b/docs/tutorials/basic_networking.md
@@ -7,15 +7,15 @@
 It seems once people master the basics of containers, networking is one of the first
 aspects they begin experimenting with.  And in regards to networking, it takes very
 little experimentation before ending up on the deep end of the pool.  The following
-guide shows the most common network setups for Podman rootfull and rootless containers.
+guide shows the most common network setups for Podman rootful and rootless containers.
 Each setup is supported with an example.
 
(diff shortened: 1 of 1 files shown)


rhatdan:
The decision was the other way around rootfull is correct just like rootless.

eriksjolund:
@rhatdan I see, both spellings can be used. No need for this PR then. I'll close it.

PR #9934, opened 2021-04-04
Update nix pin with `make nixpkgs`

hswong3i:
- Bugfix `make nixpkgs` which pin with branch `nixos-20.09`
  - Code lint with `nixpkgs-fmt`
  - Code sync between x86\_64 and aarch64

Signed-off-by: Wong Hoi Sing Edison <hswong3i@pantarei-design.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,8 +242,11 @@ bin/podman.cross.%: .gopathok
 # Update nix/nixpkgs.json its latest stable commit
 .PHONY: nixpkgs
 nixpkgs:
-	@nix run -f channel:nixos-20.09 nix-prefetch-git -c nix-prefetch-git \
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
Thanks!

To pass CI, can you add `[NO TESTS NEEDED]` to the body of your commit message?  We have a CI job to enforce adding tests which has some false positives.

hswong3i:
> [NO TESTS NEEDED]

Just force pushed, does the commit message format correct?

vrothberg:
> > [NO TESTS NEEDED]
> 
> Just force pushed, does the commit message format correct?

That should work.  We usually put it in the _body_ rather than in the title. Thanks!
LGTM

mheon:
Personally prefer it in the commit title, so it doesn't clutter commit messages.

PR #9935, opened 2021-04-04
Add support for play/generate kube PersistentVolumeClaims and Podman volumes

EduardoVega:

This PR provides the ability to support Kubernetes PersistentVolumeClaims for podman play or generate kube. A k8s PersistentVolumeClaim represents a Podman volume, only the PVC name is required to create the volume. Podman volume options can be set using PVC annotations.

This is based on this PR #8266 and related to this Issue #5788.

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/containers/common/pkg/secrets"
@@ -41,6 +42,12 @@ func (ic *ContainerEngine) PlayKube(ctx context.Context, path string, options en
 	documentList, err := splitMultiDocYAML(content)
 	if err != nil {
 		return nil, err
+	}
+
+	// sort kube kinds
+	documentList, err = sortKubeKinds(documentList)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to sort kube kinds in %q", path)
 	}
 
 	// create pod on each document if it is a pod or deployment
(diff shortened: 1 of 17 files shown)


TomSweeneyRedHat:
@EduardoVega looks like you've some trailing space issues.
Otherwise this LGTM

rhatdan:
/approve
@umohnani8 @haircommander PTAL

pkg/domain/infra/abi/play.go
```diff
@@ -313,6 +334,68 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
...
+		libpod.WithVolumeName(name),
+		libpod.WithVolumeLabels(pvcYAML.GetLabels()),
+	}
+
+	// Get pvc annotations and create remaining podman volume options if available.
+	// These are podman volume options that do not match any of the persistent volume claim
+	// attributes, so they can be configured using annotations since they will not affect k8s.
+	for k, v := range pvcYAML.GetAnnotations() {
+		switch k {
+		case "podman.volume/driver":
```
haircommander, line 360: 2 questions for consideration:
- should these be kept as constants anywhere?
- often, these special annotations follow the format of a url: `io.podman.volume` does it make sense to apply that here?

haircommander:
A couple of questions but none of them blocking. This looks great @EduardoVega, I especially like the idea about using annotations to get special podman options

pkg/domain/infra/abi/play.go
EduardoVega: 1. Let me change that.
2. You're correct. The url format does not provide any special functionality but it is better to stick to the k8s syntax. What about something like this `volume.podman.io/driver` based on `node.kubernetes.io/instance-type`?

EduardoVega: This has been changed.

rhatdan:
Thanks @EduardoVega

PR #9941, opened 2021-04-05
Fix rootlesskit port forwarder with custom slirp cidr

### LINKED ISSUE #9828: rootless kit port forwarder broken with custom slirp4netns cidr

/kind bug

**Description**

You cannot use a custom slipr4netns cidr together with rootless kit port forwarder. The problem is that the rooltess kit source ip is hard coded to the slirp4netns ip `10.0.2.100`.

The slirp4netns port forwarder works: `podman run --rm --network slirp4netns:cidr=10.0.3.0/24,port_handler=slirp4netns  -p 8080:80 nginx`

1. `podman run --rm --network slirp4netns:cidr=10.0.3.0/24  -p 8080:80 nginx`

2. `curl 127.0.0.1:8080`

curl hangs

curl should get the default nginx page

The same issue as https://github.com/containers/podman/issues/9065 just that this doesn't involve cni networking.

```
since v3.0.0
```

Luap99:
The source ip for the rootlesskit port forwarder was hardcoded to the
standard slirp4netns ip. This is incorrect since users can change the
subnet used by slirp4netns with `--network slirp4netns:cidr=10.5.0.0/24`.
The container interface ip is always the .100 in the subnet. Only when
the rootlesskit port forwarder child ip matches the container interface
ip the port forwarding will work.

Fixes #9828

### DIFF
--- a/libpod/networking_slirp4netns.go
+++ b/libpod/networking_slirp4netns.go
@@ -312,7 +312,7 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 		if netOptions.isSlirpHostForward {
 			return r.setupRootlessPortMappingViaSlirp(ctr, cmd, apiSocket)
 		}
-		return r.setupRootlessPortMappingViaRLK(ctr, netnsPath)
+		return r.setupRootlessPortMappingViaRLK(ctr, netnsPath, netOptions.cidr)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@Luap99 looks like some tests aren't happy with these changes.

rhatdan:
LGTM
@mheon @giuseppe @AkihiroSuda PTAL

libpod/networking_slirp4netns.go
```diff
@@ -391,6 +391,16 @@ func (r *Runtime) setupRootlessPortMappingViaRLK(ctr *Container, netnsPath strin
 	}
 
 	childIP := slirp4netnsIP
```
AkihiroSuda, line 393: Can we remove this const `slirp4netnsIP = "10.0.2.100"`?

Luap99: If the slirp cidr is omitted, we still need the default ip.

AkihiroSuda: If we have the default CIDR string (10.0.2.0/24), we no longer need the default IP string.

Luap99: OK but than we just changed the ip with a subnet.

giuseppe (approved):
LGTM

PR #9943, opened 2021-04-05
buildah-bud tests: handle go pseudoversions, plus...

edsantiago:
Handle go pseudoversions, e.g. a custom non-released buildah
used during testing of a PR. This will be something like:

   v1.20.1-0.20210402144408-36a37402d0c8

...and it makes it impossible (AFAIK) to do a shallow checkout;
we need to do a full clone of buildah, then git-checkout the
SHA (last element of the long string above).

FIXME: this is great for testing, but we almost certainly
want some way to block this PR from merging, don't we?

And, while testing this, found and fixed two bugs:

  - quote "$failhint" when echoing it on failure; otherwise
    we lose original whitespace.

  - invoke git-am with --reject! This makes it SO MUCH EASIER
    to identify the failing part of our patch!

### DIFF
--- a/test/buildah-bud/run-buildah-bud-tests
+++ b/test/buildah-bud/run-buildah-bud-tests
@@ -72,7 +72,7 @@ function die() {
 
 # From here on out, any unexpected abort will try to offer helpful hints
 failhint=
-trap 'if [[ $? != 0 ]]; then if [[ -n $failhint ]]; then echo;echo "***************************************";echo $failhint;echo;echo "Please see $BUD_TEST_DIR_REL/README.md for advice";fi;fi' 0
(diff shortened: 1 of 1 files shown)


mheon:
Sure, LGTM

rhatdan:
LGTM If it passes.

edsantiago:
Oh, it'll pass (correction: has passed). The test will be on a PR like your #9896 which brings in a non-tagged buildah. (I tested it on my end, of course.)

rhatdan:
I added this patch to my PR with the vendored version of Podman, if the tests pass, we can merge both.

edsantiago:
Fixed in #9962

PR #9946, opened 2021-04-05
[CI:DOCS] Restore rootless-cni-infra contrib info

TomSweeneyRedHat:
PR #9423 removed the contrib/rootless-cni-infra directory and
it's files and should not have.  The information in that tree
is required for Podman v3.0 and earlier users, and primarily
those in RHEL 8.4.0 and 8.3.1.

This PR puts those files back and adds a note to the README.md
noting the changes are not needed for versions after Podman v3.0.

### DIFF
--- a/contrib/rootless-cni-infra/Containerfile
+++ b/contrib/rootless-cni-infra/Containerfile
@@ -1,0 +1,36 @@
+ARG GOLANG_VERSION=1.15
+ARG ALPINE_VERSION=3.12
+ARG CNI_VERSION=v0.8.0
+ARG CNI_PLUGINS_VERSION=v0.8.7
+ARG DNSNAME_VERSION=v1.1.1
+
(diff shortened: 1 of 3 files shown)


mheon:
NACK. This is unnecessary. We are pointing people at the long-term support branches, not master, for the image. We should not continue to maintain this readme and image.

TomSweeneyRedHat:
I'll buy the NACK when you show me where we're pointing people to find this info....  Someone running into this on Podman 2.2 I don't think will find it unless it's on master.

mheon:
We point them at the specific release branch. From our 2.2.1 RHEL branch - https://github.com/containers/podman/commit/4007554f97553055131df12f9a8d7b162d094fb8 points people to https://github.com/containers/podman/blob/v2.2.1-rhel/contrib/rootless-cni-infra/ - that's not master, that's the v2.2.1-rhel branch. We did this deliberately to ensure they pulled a version of the infra image that matched their Podman. v3.0.1 is the same.

PR #9950, opened 2021-04-06
podman unshare: add --rootless-cni to join the ns

Luap99:
Add a new --rootless-cni option to podman unshare to also join the
rootless-cni network namespace. This is useful if you want to connect
to a rootless container via IP address. This is only possible from the
rootless-cni namespace and not from the host namespace. This option also
helps to debug problems in the rootless-cni namespace.

### DIFF
--- a/cmd/podman/system/unshare.go
+++ b/cmd/podman/system/unshare.go
@@ -12,9 +12,10 @@ import (
 )
 
 var (
+	unshareOptions     = entities.SystemUnshareOptions{}
 	unshareDescription = "Runs a command in a modified user namespace."
 	unshareCommand     = &cobra.Command{
-		Use:                   "unshare [COMMAND [ARG...]]",
+		Use:                   "unshare [options] [COMMAND [ARG...]]",
 		DisableFlagsInUseLine: true,
 		Short:                 "Run a command in a modified user namespace",
 		Long:                  unshareDescription,
@@ -33,6 +34,7 @@ func init() {
 	})
 	flags := unshareCommand.Flags()
 	flags.SetInterspersed(false)
+	flags.BoolVar(&unshareOptions.RootlessCNI, "rootless-cni", false, "Join the rootless network namespace used for CNI networking")
 }
 
 func unshare(cmd *cobra.Command, args []string) error {
@@ -49,5 +51,5 @@ func unshare(cmd *cobra.Command, args []string) error {
 		args = []string{shell}
 	}
 
-	return registry.ContainerEngine().Unshare(registry.Context(), args)
+	return registry.ContainerEngine().Unshare(registry.Context(), args, unshareOptions)
 }
--- a/pkg/domain/infra/abi/system.go
+++ b/pkg/domain/infra/abi/system.go
@@ -390,13 +390,25 @@ func unshareEnv(graphroot, runroot string) []string {
 		fmt.Sprintf("CONTAINERS_RUNROOT=%s", runroot))
 }
 
-func (ic *ContainerEngine) Unshare(ctx context.Context, args []string) error {
-	cmd := exec.Command(args[0], args[1:]...)
-	cmd.Env = unshareEnv(ic.Libpod.StorageConfig().GraphRoot, ic.Libpod.StorageConfig().RunRoot)
-	cmd.Stdin = os.Stdin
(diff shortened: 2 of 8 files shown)


cmd/podman/system/unshare.go
```diff
@@ -33,6 +34,7 @@ func init() {
 	})
 	flags := unshareCommand.Flags()
 	flags.SetInterspersed(false)
+	flags.BoolVar(&unshareOptions.RootlessCNI, "rootless-cni", false, "Join the rootless-cni namespace")
```
mheon, line 37: Suggest rewording description to "Join the rootless network namespace used for CNI networking"

pkg/domain/infra/abi/system.go
```diff
@@ -390,13 +390,25 @@ func unshareEnv(graphroot, runroot string) []string {
...
+		cmd.Stderr = os.Stderr
+		return cmd.Run()
+	}
+
+	if options.RootlessCNI {
+		rootlesscni, err := ic.Libpod.GetRootlessCNINetNs(true)
+		if err != nil {
+			return err
+		}
+		defer rootlesscni.Cleanup(ic.Libpod)
```
mheon, line 408: Is there a risk that we wipe out the rootless CNI namespace if we have two unshare sessions both joined to it, and one exits before the other?

Luap99: Yes that is problematic. What would be the proper way to fix this? Cleanup currently checks that there are no running containers with bridge networking and that `var/lib/cni/results` is empty. Otherwise cleanup is skipped.

mheon: Hm. We could try reference counting, but I'm concerned we could get into a state where we never clean up if a `podman unshare` gets SIGKILL'd before it can indicate it's done.

The easiest way is probably to guarantee there's only 1 `podman unshare` shell with `--rootless-cni` active at a time using a lockfile. I don't think we'll run into cases where more than 1 session is needed very often?

Luap99: Yeah I think we should ensure that we always do cleanup. Otherwise we will not only leak the namespace but also the slirp process.

Only allowing one session feels a little odd. The only thing that will happen to the other unshare session is that we lose network connectivity because we kill the slirp process. The namespace stays alive until the process exits only the bind mount is removed bind cleanup.  Also as long as at least one container with CNI networking is running we can safely use as many sessions as we want.

I think we can live with this problem.

mheon: OK, that ought to be fine, then.

cmd/podman/system/unshare.go
Luap99: Done

mheon:
Compose test failing, potential flake? Restarted.

LGTM

Luap99:
> Compose test failing, potential flake? Restarted.

Yes I have to fix this ASAP

pkg/domain/infra/abi/system.go
giuseppe: could we just join the network namespace without creating it?

That would also solve the issue that we don't have to cleanup on exit?

Luap99: We could do this but it doesn't solve the problem. We could skip the cleanup here but if all running cni containers are stopped we will perform the cleanup and then all running unshare sessions would lose network connectivity.
Also the reason I choose to create the ns here is because this should open up more use cases. I thought about using this to create the ns and do some setup stuff, e.g. create a interface used by cni.

mheon:
Needs a rebase

Luap99:
Updated. I added a note to the man page that using more than one session can have unexpected results. I also added a extra check to prevent a possible panic.

TomSweeneyRedHat:
LGTM
and happy green test buttons.

pkg/domain/infra/abi/system.go
giuseppe: could we split the two cases, perhaps `--rootless-cni=create|join`?  I think having an option to just join the network namespace, and lose connectivity if there are no other containers running is still a valid use case.

Luap99: There is only one rootless cni namespace per user. We will always join if it exists.

PR #9944, opened 2021-04-05
[CI:DOCS] Set all swagger operation id's to be compatible

jwhonce:
* Libpod operation id's changed to better match compatible id

Supersedes #9123 by included all commits. 

Additionally, corrects a duplicated ID that prevented that PR#9123 from merging, and rebased to current master.

Signed-off-by: Tom Deseyn <tom.deseyn@gmail.com>
Signed-off-by: Jhon Honce <jhonce@redhat.com>

Issue #9951, opened 2021-04-06
[pedantic] swagger operation ID nits


edsantiago:
Thank you for #9944! It hugely improves consistency in swagger, which is something my `hack/swagger-check` script has sorely wanted to check. I've updated my script to recognize the new operation-name format, and added a bunch of special cases, but I think the following might be actual bugs:

--------------------

https://github.com/containers/podman/blob/f143de9db6861cd8b8d6fc4e03f2826041307982/pkg/api/server/register_containers.go#L777

Shouldn't that be `ContainerShowmountedLibpod`?

--------------
https://github.com/containers/podman/blob/f143de9db6861cd8b8d6fc4e03f2826041307982/pkg/api/server/register_containers.go#L1471-L1472

This one is weird: two consecutive swagger lines, the first one tagged `libpod` but not a `/libpod` endpoint, the second tagged `compat` but with a `/libpod` endpoint. This just looks completely off to me. If it's intentional, just say so, and I'll hardcode an exception.

------------------------

https://github.com/containers/podman/blob/f143de9db6861cd8b8d6fc4e03f2826041307982/pkg/api/server/register_play.go#L11

Going by the rules and conventions, that should be `PlayKube`

edsantiago:
There are also these three which are annoying but I suppose I'll have to special-case:
```
       ContainersStatsAllLibpod
       ImageGetAll
       ImageDeleteAllLibpod
```
(I suppose the `All` is maybe helpful to humans, but it's spurious and arbitrary)

jwhonce:
@edsantiago `All` came in from the compatibility requirements, the idea is those operations are "one or more". OperationIds need to be unique across the API which does lead to some naming weirdness.  #9953 includes all your recommendations.

PR #9952, opened 2021-04-06
[CI:DOCS] swagger-check: compare operations

edsantiago:
Until now we've only compared operations when called with the
non-default --pedantic flag, because there were way too many
exceptions.

With the merge of #9944 the rules have become much cleaner.
Still not perfect, but it's now possible to have simple
general rules with a (semi-)manageable list of exceptions.

### DIFF
--- a/hack/swagger-check
+++ b/hack/swagger-check
@@ -59,11 +59,6 @@ says 'libpod'.
 
 OPTIONS:
 
-  --pedantic     Compare operation names (the last part of swagger comment).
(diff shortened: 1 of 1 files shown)


edsantiago:
Tagged as WIP because I'd like to see #9951 addressed first. (If those need fixing, I'm happy to do so in this PR).

jwhonce:
LGTM

edsantiago:

PR #9955, opened 2021-04-06
System tests: special case for RHEL: require runc

edsantiago:
As discussed in watercooler 2021-04-06: make sure that RHEL
and CentOS are using runc. Using crun is probably a packaging
error that should be caught early.

### DIFF
--- a/test/system/005-info.bats
+++ b/test/system/005-info.bats
@@ -53,6 +53,27 @@ store.imageStore.number   | 1
 
 }
 
+# 2021-04-06 discussed in watercooler: RHEL must never use crun, even if
+# using cgroups v2.
+@test "podman info - RHEL8 must use runc" {
+    local osrelease=/etc/os-release
+    test -e $osrelease || skip "Not a RHEL system (no $osrelease)"
+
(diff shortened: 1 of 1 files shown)


test/system/005-info.bats
```diff
@@ -53,6 +53,21 @@ store.imageStore.number   | 1
 
 }
 
+# 2021-04-06 discussed in watercooler: RHEL must never use crun, even if
+# using cgroups v2.
+@test "podman info - RHEL must use runc" {
```
rhatdan, line 58: Can you make this only for RHEL < 9.0?

mheon:
LGTM once comment from @rhatdan is addressed

test/system/005-info.bats
edsantiago: Done. New changes pushed, with a FIXME because I have no idea what to expect on CentOS. (Since we don't run system tests on it, I don't expect this to ever trigger, but if anyone reading this knows what `/etc/os-release` looks like on CentOS, I'd appreciate seeing `$VERSION_ID`, TIA)

vrothberg (commented):
/lgtm
/hold

PR #9956, opened 2021-04-06
[CI:DOCS] Update podman image Dockerfile to support Podman in container

rhatdan:
[NO TEST NEEDED] Can not test this in CI/CD system since it needs to be
merged in order for the Dockerfiles to even work.

Modified the /etc/subuid and /etc/subgid to be able to run in rootless
containers.  The Range can not be the same as on the host.

Add /home/podman/.config/containers/containers.conf to automatically
mount /proc on /proc while inside of the container. This prevents
additional permissions being required that are blocked when not in
--privileged mode.

Setup volumes for /var/lib/containers and
/home/podman/.local/share/containwers

This will prevent the errors where people are doing overlay on overlay.

### DIFF
--- a/contrib/podmanimage/stable/Dockerfile
+++ b/contrib/podmanimage/stable/Dockerfile
@@ -11,9 +11,21 @@ FROM registry.fedoraproject.org/fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN useradd podman; yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN dnf -y update; yum -y reinstall shadow-utils; \
+yum -y install podman fuse-overlayfs --exclude container-selinux; \
+rm -rf /var/cache /var/log/dnf* /var/log/yum.*
 
-ADD https://raw.githubusercontent.com/containers/libpod/master/contrib/podmanimage/stable/containers.conf /etc/containers/
+RUN useradd podman; \
+echo podman:10000:5000 > /etc/subuid; \
+echo podman:10000:5000 > /etc/subgid;
+
+VOLUME /var/lib/containers
+VOLUME /home/podman/.local/share/containers
+
+ADD https://raw.githubusercontent.com/containers/libpod/master/contrib/podmanimage/stable/containers.conf /etc/containers/containers.conf
+ADD https://raw.githubusercontent.com/containers/libpod/master/contrib/podmanimage/stable/podman-containers.conf /home/podman/.config/containers/containers.conf
+
+RUN chown podman:podman -R /home/podman
 
 # chmod containers.conf and adjust storage.conf to enable Fuse storage.
 RUN chmod 644 /etc/containers/containers.conf; sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' /etc/containers/storage.conf
--- a/contrib/podmanimage/testing/Dockerfile
+++ b/contrib/podmanimage/testing/Dockerfile
@@ -13,9 +13,19 @@ FROM registry.fedoraproject.org/fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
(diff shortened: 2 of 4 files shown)


rhatdan:
@TomSweeneyRedHat @mheon @vrothberg PTAL

mheon:
LGTM for me. Compose test is failing elsewhere, likely unrelated.

contrib/podmanimage/stable/Dockerfile
```diff
@@ -11,9 +11,21 @@ FROM registry.fedoraproject.org/fedora:latest
...
+RUN useradd podman; \
+echo podman:10000:5000 > /etc/subuid; \
+echo podman:10000:5000 > /etc/subgid;
+
+VOLUME /var/lib/containers
+VOLUME /home/podman/.local/share/containers
+
+ADD https://raw.githubusercontent.com/containers/libpod/master/contrib/podmanimage/stable/containers.conf /etc/containers/containers.conf
+ADD https://raw.githubusercontent.com/containers/libpod/master/contrib/podmanimage/stable/podman-containers.conf /home/podman/.config/containers/containers.conf
+
```
vrothberg, line 27: Can we make the source path relative?  This way, we'd always use the local files.  The upstream files may unintentionally break the "stable" images after a rebuild.

rhatdan: @TomSweeneyRedHat PTAL
I believe this is the only way this works on Quay.io, but I could be mistaken.  @TomSweeneyRedHat set this up originally.

contrib/podmanimage/testing/Dockerfile
```diff
@@ -13,9 +13,19 @@ FROM registry.fedoraproject.org/fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN useradd podman; yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux --enablerepo updates-testing; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN podman; yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux --enablerepo updates-testing; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
```
eriksjolund, line 16:
```suggestion
RUN yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux --enablerepo updates-testing; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
```

contrib/podmanimage/stable/Dockerfile
TomSweeneyRedHat: I was unable to figure a way to make it source.  quay.io doesn't pull the files over and runs this in corn or something like that within their servers.

rhatdan:
@giuseppe @umohnani8 PTAL

contrib/podmanimage/stable/Dockerfile
```diff
@@ -11,9 +11,21 @@ FROM registry.fedoraproject.org/fedora:latest
...
-RUN useradd podman; yum -y update; yum -y reinstall shadow-utils; yum -y install podman fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN dnf -y update; yum -y reinstall shadow-utils; \
+yum -y install podman fuse-overlayfs --exclude container-selinux; \
+rm -rf /var/cache /var/log/dnf* /var/log/yum.*
 
-ADD https://raw.githubusercontent.com/containers/libpod/master/contrib/podmanimage/stable/containers.conf /etc/containers/
+RUN useradd podman; \
+echo podman:10000:5000 > /etc/subuid; \
+echo podman:10000:5000 > /etc/subgid;
+
```
giuseppe, line 21: Just something to think about, not blocking for this PR but that we probably want to address in future.

With such ranges specified:
- If the outer Podman doesn't run in a user namespace, all the inner Podman will end up using the same IDs.
- If the outer Podman runs in a user namespace, there won't likely be enough IDs to satisfy this request.

If we want to address the first scenario, we could probably fix it at runtime, and have the entrypoint script generate these entries with a random initial ID.

giuseppe (approved):
just an inline question, but:

LGTM

umohnani8:
LGTM

contrib/podmanimage/stable/Dockerfile
rhatdan: Yes I think in the long run we want to make this more flexible, some how, But for now, we want to get to the point that we have an image which people can experiment with.

PR #9958, opened 2021-04-06
Volumes prune endpoint should use only prune filters

jmguzik:
Volumes endpoints for HTTP compat and libpod APIs allowed
usage of list HTTP endpoint filter funcs. Documentation in
case of compat API does not allow that. This commit aligns
code with the documentation and also ligns libpod with compat API.

/kind bug

### DIFF
--- a/pkg/domain/filters/volumes.go
+++ b/pkg/domain/filters/volumes.go
@@ -75,7 +75,25 @@ func GenerateVolumeFilters(filters url.Values) ([]libpod.VolumeFilter, error) {
 					return dangling
 				})
 			default:
-				return nil, errors.Errorf("%q is in an invalid volume filter", filter)
+				return nil, errors.Errorf("%q is an invalid volume filter", filter)
+			}
+		}
+	}
+	return vf, nil
+}
+
+func GeneratePruneVolumeFilters(filters url.Values) ([]libpod.VolumeFilter, error) {
+	var vf []libpod.VolumeFilter
+	for filter, v := range filters {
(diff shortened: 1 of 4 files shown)


jmguzik:
Docker API prune filters:
https://docs.docker.com/engine/api/v1.41/#operation/VolumePrune
The same is stated in podman compat.

rhatdan:
/approve
LGTM

pkg/domain/filters/volumes.go
```diff
@@ -81,3 +81,21 @@ func GenerateVolumeFilters(filters url.Values) ([]libpod.VolumeFilter, error) {
...
+	for filter, v := range filters {
+		for _, val := range v {
+			switch filter {
+			case "label":
+				filter := val
+				vf = append(vf, func(v *libpod.Volume) bool {
+					return util.MatchLabelFilters([]string{filter}, v.Labels())
+				})
+			default:
+				return nil, errors.Errorf("%q is in an invalid volume filter", filter)
```
mheon, line 96: Small suggestion: "%q is an invalid volume prune filter"

mheon:
One nit otherwise LGTM

pkg/domain/filters/volumes.go
jmguzik: Nicely caught. I corrected this in the original filter function and here as well.

jmguzik:
> One nit otherwise LGTM

Done

TomSweeneyRedHat:
LGTM
/hold cancel

PR #9959, opened 2021-04-06
rootless cni add /usr/sbin to PATH if not present

Luap99:
The CNI plugins need access to iptables in $PATH. On debian /usr/sbin
is not added to $PATH for rootless users. This will break rootless
cni completely. To prevent breaking existing users add /usr/sbin to
$PATH in podman if needed.


mheon:
LGTM

PR #9961, opened 2021-04-07
Fix flaking rootless compose test

Luap99:
The compose port test is flaking with an empty curl result. The curl retry
does not work properly. Given the the tests never expect an empty result
lets just wait one second and retry again.
Unfortunately there is no way for me to actually verify if this will fix
the flake.

### DIFF
--- a/test/compose/test-compose
+++ b/test/compose/test-compose
@@ -163,7 +163,13 @@ function test_port() {
     local op="$2"                # '=' or '~'
     local expect="$3"            # what to expect from curl output
 
-    local actual=$(curl --retry 10 --retry-all-errors -s http://127.0.0.1:$port/)
+    local actual=$(curl --retry 3 --retry-all-errors -s http://127.0.0.1:$port/)
+    # The test is flaking with an empty result. The curl retry doesn't solve this.
+    # If the result is empty sleep one second and try again.
+    if [[ "$actual" == "" ]]; then
+        sleep 1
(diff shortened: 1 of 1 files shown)


test/compose/test-compose
```diff
@@ -163,7 +163,13 @@ function test_port() {
     local op="$2"                # '=' or '~'
     local expect="$3"            # what to expect from curl output
 
-    local actual=$(curl --retry 10 --retry-all-errors -s http://127.0.0.1:$port/)
+    local actual=$(curl --retry 3 --retry-all-errors -s http://127.0.0.1:$port/)
+    # The test is flaking with an empty result. The curl retry doesn't solve this.
+    # If the result is empty sleep one second and try again.
+    if [[ "$actual" == "" ]]; then
+        sleep 1
+        local actual=$(curl --retry 3 --retry-all-errors -s http://127.0.0.1:$port/)
```
edsantiago, line 171: You probably don't want the `local` here

edsantiago:
LGTM. One nit that shouldn't affect anything. Two flakes, which I've restarted. Thanks for looking into this.

test/compose/test-compose
Luap99: Thanks for clarifying this. I wasn't sure if local has to be set again.

edsantiago: It has scope within the function, not the block. This usage is a little jarring, but not worth resubmitting for.

PR #9968, opened 2021-04-07
[CI:DOCS] Adjust libpod API Container Wait documentation to the code

### LINKED ISSUE #9960: Inconsistencies in Container Wait API documentation

/kind bug

**Description**

The HTTP API reference for libpod/containers/{name}/wait at https://docs.podman.io/en/latest/_static/api.html#operation/ContainerWaitLibpod says that the output for `200` is content type `application/json` and of the form:
```
{
    "Error": 
    {
        "Message": "string"
    },
    "StatusCode": 0
}
```
However, when I try on an instance of podman 3.0.1, the reply is a single integer, which seems to me consistent with the code here:
https://github.com/containers/podman/blob/382b5b81d24870abe400d14296e4f5ef47c99d29/pkg/api/handlers/utils/containers.go#L138

Is that possibly a bug in the documentation?

I also realized looking into the `WaitContainerLibpod` function, that there is  an undocumented parameter, `interval`. Should that possibly be documented?

pablofsf:
Closes #9960

As of the issue, what is currently documented as the response for `/libpod/containers/{name}/wait` is not actually what the code is doing. At the same time, `interval` property was left undocumented.

### DIFF
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -1194,11 +1194,22 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//       - removing
 	//       - stopping
 	//    description: "Conditions to wait for. If no condition provided the 'exited' condition is assumed."
+	//  - in: query
+	//    name: interval
+	//    type: string
+	//    default: "250ms"
+	//    description: Time Interval to wait before polling for completion.
 	// produces:
 	// - application/json
+	// - text/plain
 	// responses:
 	//   200:
-	//     $ref: "#/responses/ContainerWaitResponse"
+	//     description: Status code
+	//     schema:
+	//       type: integer
+	//       format: int32
+	//     examples:
+	//       text/plain: 137
 	//   404:
 	//     $ref: "#/responses/NoSuchContainer"
 	//   500:


pkg/api/server/register_containers.go
```diff
@@ -1194,11 +1194,22 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//       - removing
 	//       - stopping
 	//    description: "Conditions to wait for. If no condition provided the 'exited' condition is assumed."
+	//  - in: query
+	//    name: interval
+	//    type: string
+	//    default: "250ms"
+	//    description: Time Interval to wait before polling for completion.
 	// produces:
 	// - application/json
```
mheon, line 1203: Shouldn't this be one or the other? @jwhonce

baude:
you need to sign your commit ... git commit --amend -s following by git push -f ...

pablofsf:
@baude, sorry, forgot to push it after reading the CONTRIBUTION.md and fixing it...

baude:
lgtm

pkg/api/server/register_containers.go
jwhonce: @mheon I do not understand your question.

PR #9969, opened 2021-04-07
Update manifest API endpoints

jwhonce:
* Add validation for manifest name
* Always return an array for manifests even if empty

When c/image is refactored to include manifests, these endpoints should
be revisited.


rhatdan:
LGTM

PR #9972, opened 2021-04-08
Add host.containers.internal entry into container's etc/hosts

### LINKED ISSUE #5651: hostname for container gateway
/kind feature

**Description**

Provide a dns entry to containers pointing to their gateway.

Context : as a developer, one of my setup involves a service running on the host machine, that should be proxied by nginx running in a container.

I could make it work by writing the container's static gateway ip in nginx's config.

It would have been nice to have a standard hostname for the gateway (I believe docker has `host.docker.internal` for that). Or did I miss some easy alternative ?

```
Version:            1.8.2
RemoteAPI Version:  1
Go Version:         go1.14
Git Commit:         028e3317eb1494b9b2acba4a0a295df80fae66cc
Built:              Sat Mar 21 15:30:34 2020
OS/Arch:            linux/amd64```

### PR DESCRIPTION
**Pull request by bblenard**
This change adds the entry `host.containers.internal` to the `/etc/hosts`
file within a new containers filesystem. The ip address is determined by
the containers networking configuration and points to the gateway address
for the containers networking namespace.

Closes #5651

Signed-off-by: Baron Lenardson <lenardson.baron@gmail.com>

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1358,6 +1358,34 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
 	return c.save()
 }
 
+// Retrieves a container's "root" net namespace container dependency.
+func (c *Container) getRootNetNsDepCtr() (depCtr *Container, err error) {
+	containersVisited := map[string]int{c.config.ID: 1}
+	nextCtr := c.config.NetNsCtr
+	for nextCtr != "" {
+		// Make sure we aren't in a loop
+		if _, visited := containersVisited[nextCtr]; visited {
+			return nil, errors.New("loop encountered while determining net namespace container")
+		}
+		containersVisited[nextCtr] = 1
+
+		depCtr, err = c.runtime.state.Container(nextCtr)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error fetching dependency %s of container %s", c.config.NetNsCtr, c.ID())
+		}
+		// This should never happen without an error
+		if depCtr == nil {
+			break
+		}
+		nextCtr = depCtr.config.NetNsCtr
+	}
+
+	if depCtr == nil {
+		return nil, errors.New("unexpected error depCtr is nil without reported error from runtime state")
+	}
+	return depCtr, nil
+}
+
 // Make standard bind mounts to include in the container
 func (c *Container) makeBindMounts() error {
 	if err := os.Chown(c.state.RunDir, c.RootUID(), c.RootGID()); err != nil {
@@ -1396,24 +1424,9 @@ func (c *Container) makeBindMounts() error {
 			// We want /etc/resolv.conf and /etc/hosts from the
 			// other container. Unless we're not creating both of
 			// them.
-			var (
-				depCtr  *Container
-				nextCtr string
-			)
-
-			// I don't like infinite loops, but I don't think there's
-			// a serious risk of looping dependencies - too many
-			// protections against that elsewhere.
-			nextCtr = c.config.NetNsCtr
-			for {
-				depCtr, err = c.runtime.state.Container(nextCtr)
-				if err != nil {
-					return errors.Wrapf(err, "error fetching dependency %s of container %s", c.config.NetNsCtr, c.ID())
-				}
-				nextCtr = depCtr.config.NetNsCtr
-				if nextCtr == "" {
-					break
-				}
+			depCtr, err := c.getRootNetNsDepCtr()
+			if err != nil {
+				return errors.Wrapf(err, "error fetching network namespace dependency container for container %s", c.ID())
 			}
 
 			// We need that container's bind mounts
@@ -1698,7 +1711,12 @@ func (c *Container) generateResolvConf() (string, error) {
 		nameservers = resolvconf.GetNameservers(resolv.Content)
 		// slirp4netns has a built in DNS server.
 		if c.config.NetMode.IsSlirp4netns() {
-			nameservers = append([]string{slirp4netnsDNS}, nameservers...)
+			slirp4netnsDNS, err := GetSlirp4netnsDNS(c.slirp4netnsSubnet)
+			if err != nil {
+				logrus.Warn("failed to determine Slirp4netns DNS: ", err.Error())
+			} else {
+				nameservers = append([]string{slirp4netnsDNS.String()}, nameservers...)
+			}
 		}
 	}
 
@@ -1779,7 +1797,12 @@ func (c *Container) getHosts() string {
 	if c.Hostname() != "" {
 		if c.config.NetMode.IsSlirp4netns() {
 			// When using slirp4netns, the interface gets a static IP
-			hosts += fmt.Sprintf("# used by slirp4netns\n%s\t%s %s\n", slirp4netnsIP, c.Hostname(), c.config.Name)
+			slirp4netnsIP, err := GetSlirp4netnsGateway(c.slirp4netnsSubnet)
+			if err != nil {
+				logrus.Warn("failed to determine slirp4netnsIP: ", err.Error())
+			} else {
+				hosts += fmt.Sprintf("# used by slirp4netns\n%s\t%s %s\n", slirp4netnsIP.String(), c.Hostname(), c.config.Name)
+			}
 		} else {
 			hasNetNS := false
 			netNone := false
@@ -1802,6 +1825,36 @@ func (c *Container) getHosts() string {
 			}
 		}
 	}
+
+	// Add gateway entry
+	var depCtr *Container
+	if c.config.NetNsCtr != "" {
+		// ignoring the error because there isn't anything to do
+		depCtr, _ = c.getRootNetNsDepCtr()
+	} else if len(c.state.NetworkStatus) != 0 {
+		depCtr = c
+	} else {
+		depCtr = nil
+	}
+
+	if depCtr != nil {
+		for _, pluginResultsRaw := range depCtr.state.NetworkStatus {
+			pluginResult, _ := cnitypes.GetResult(pluginResultsRaw)
+			for _, ip := range pluginResult.IPs {
+				hosts += fmt.Sprintf("%s host.containers.internal\n", ip.Gateway)
+			}
+		}
+	} else if c.config.NetMode.IsSlirp4netns() {
+		gatewayIP, err := GetSlirp4netnsGateway(c.slirp4netnsSubnet)
+		if err != nil {
+			logrus.Warn("failed to determine gatewayIP: ", err.Error())
+		} else {
+			hosts += fmt.Sprintf("%s host.containers.internal\n", gatewayIP.String())
+		}
+	} else {
+		logrus.Debug("network configuration does not support host.containers.internal address")
+	}
+
 	return hosts
 }
 
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -37,15 +37,11 @@ import (
 )
 
 const (
-	// slirp4netnsIP is the IP used by slirp4netns to configure the tap device
-	// inside the network namespace.
-	slirp4netnsIP = "10.0.2.100"
-
-	// slirp4netnsDNS is the IP for the built-in DNS server in the slirp network
-	slirp4netnsDNS = "10.0.2.3"
-
 	// slirp4netnsMTU the default MTU override
 	slirp4netnsMTU = 65520
+
+	// default slirp4ns subnet
+	defaultSlirp4netnsSubnet = "10.0.2.0/24"
 
 	// rootlessCNINSName is the file name for the rootless network namespace bind mount
 	rootlessCNINSName = "rootless-cni-ns"
@@ -360,15 +356,20 @@ func (r *Runtime) GetRootlessCNINetNs(new bool) (*RootlessCNI, error) {
 				}
 
 				// build a new resolv.conf file which uses the slirp4netns dns server address
-				resolveIP := slirp4netnsDNS
+				resolveIP, err := GetSlirp4netnsDNS(nil)
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to determine default slirp4netns DNS address")
+				}
+
 				if netOptions.cidr != "" {
 					_, cidr, err := net.ParseCIDR(netOptions.cidr)
 					if err != nil {
 						return nil, errors.Wrap(err, "failed to parse slirp4netns cidr")
 					}
-					// the slirp dns ip is always the third ip in the subnet
-					cidr.IP[len(cidr.IP)-1] = cidr.IP[len(cidr.IP)-1] + 3
-					resolveIP = cidr.IP.String()
+					resolveIP, err = GetSlirp4netnsDNS(cidr)
+					if err != nil {
+						return nil, errors.Wrapf(err, "failed to determine slirp4netns DNS address from cidr: %s", cidr.String())
+					}
 				}
 				conf, err := resolvconf.Get()
 				if err != nil {
@@ -377,7 +378,7 @@ func (r *Runtime) GetRootlessCNINetNs(new bool) (*RootlessCNI, error) {
 				searchDomains := resolvconf.GetSearchDomains(conf.Content)
 				dnsOptions := resolvconf.GetOptions(conf.Content)
 
-				_, err = resolvconf.Build(filepath.Join(cniDir, "resolv.conf"), []string{resolveIP}, searchDomains, dnsOptions)
+				_, err = resolvconf.Build(filepath.Join(cniDir, "resolv.conf"), []string{resolveIP.String()}, searchDomains, dnsOptions)
 				if err != nil {
 					return nil, errors.Wrap(err, "failed to create rootless cni resolv.conf")
 				}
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: libpod/container_internal_linux.go [hunk 1]
```diff
@@ -1381,23 +1401,12 @@ func (c *Container) makeBindMounts() error {
 			// other container. Unless we're not creating both of
 			// them.
 			var (
-				depCtr  *Container
-				nextCtr string
+				depCtr *Container
```
mheon, line 1404: Can drop this block and make `depCtr, err = c.getRootNetNsDepCtr()` a `:=`

Will be a little cleaner

libpod/networking_linux.go
```diff
@@ -36,7 +36,7 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const (
+var (
```
mheon, line 39: Can you keep this block `const` and then add a separate `var` block below for the slirp4netns bit?

mheon:
I'd like @AkihiroSuda or someone else familiar with slirp to verify the slirp address calculation bits. Overall looks good, few small comments.

TomSweeneyRedHat:
Nothing over @mheon's comments.  @giuseppe if you have a spare eyeball for this, that would be great.

libpod/container_internal_linux.go
```diff
@@ -1342,6 +1342,26 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
 	return c.save()
 }
 
+// Retrieves a container's "root" net namespace container dependency.
```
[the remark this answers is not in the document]
bblenard: I don't think this can be null unless there is an error, although I could be wrong. So would you like a note about it returning an nil on error?

test/system/500-networking.bats
```diff
@@ -95,6 +95,12 @@ load helpers
     run_podman rm $cid
 }
 
+@test "podman run with slirp4ns assigns correct gateway address to host.gateway.internal" {
+    run_podman run --network slirp4netns:cidr=192.168.0.0/24,allow_host_loopback=true \
+                $IMAGE grep 192.168.0.2 /etc/hosts
+    is "$output"   "192.168.0.2 host.gateway.internal"   "host.gateway.internal should be the cidr+2 address"
+}
```
AkihiroSuda, line 102: I'm not feeling this behavior is correct.
The CIDR+2 address isn't useful when slirp4netns is launched with `--disable-host-loopback`.
And `--disable-host-loopback` should not be removed, as exposing bare 127.0.0.1 to containers might result in container-breakout.

I suggest exposing the "real" host IP regardless to rootful/rootless.

AkihiroSuda (changes requested):

test/system/500-networking.bats
bblenard: So what would be the alternative? Are you suggesting to derive all interface ip addresses not including `lo`?

The issue (#5651) was to add behavior that reflects docker's `host.docker.internal`. 

I had a look through some of the docker code and as far as I can tell that is a configuration option given to the daemon so if a user asks for the host gateway address to be added to their container it is added based on that configuration [docker-ce reference](https://github.com/docker/docker-ce/blob/cc71e4fd9a9933078d79daaf69c8422c1bfa1618/components/engine/cmd/dockerd/config.go#L67)

So should podman have a similar configuration option?

AkihiroSuda: > Are you suggesting to derive all interface ip addresses not including lo?

The first interface (eth0) would be fine

libpod/container_internal_linux.go
```diff
@@ -1786,6 +1795,31 @@ func (c *Container) getHosts() string {
...
 	}
+
+	// Add gateway entry
+	var depCtr *Container
+	if c.config.NetNsCtr != "" {
+		// ignoring the error because there isn't anything to do
+		depCtr, _ = c.getRootNetNsDepCtr()
+	} else if len(c.state.NetworkStatus) != 0 {
+		depCtr = c
+	} else {
```
rhatdan, line 1806: This last block is not necessary.  depCtr will be nil by default.

libpod/container_internal_linux.go
```diff
@@ -1786,6 +1795,31 @@ func (c *Container) getHosts() string {
...
+		for _, pluginResultsRaw := range depCtr.state.NetworkStatus {
+			pluginResult, _ := cnitypes.GetResult(pluginResultsRaw)
+			for _, ip := range pluginResult.IPs {
+				hosts += fmt.Sprintf("%s host.gateway.internal\n", ip.Gateway)
+			}
+		}
+	} else if c.config.NetMode.IsSlirp4netns() {
+		hosts += fmt.Sprintf("%s host.gateway.internal\n", slirp4netnsGateway)
+	} else {
+		logrus.Debug("unexpected network configuration while determining host.gateway.internal address")
```
rhatdan, line 1820: Is this unexpected, or would this just be the situation if I run a container with `podman run --net=none ...`

IE Is this worth logging?

rhatdan:
@bblenard Thanks for working on this.

test/system/500-networking.bats
bblenard: I'm not sure about this solution.

There can be so many different interfaces in so many different configurations that trying to derive an ip address from one of the interfaces seems unstable. I don't necessarily disagree that exposing the loopback interface to the container could result in a container breakout but this change doesn't give that access it simply puts a `/etc/hosts` entry for that interface.

Like I said above. I think the only viable alternative is to make the interface a configurable option similar to dockerd's `--host-gateway-ip` launch option.

libpod/container_internal_linux.go
bblenard: This would be the case for `podman run --net=none` so I suppose it isn't that unexpected. Should I just quietly continue or change up the debug message?

libpod/container_internal_linux.go
bblenard: I'm not sure syntactically how to get rid of that block :grimacing: 

I added it because I wanted to ensure mutual exclusion between shared container network namespace configurations and containers with network namespaces created for them. Basically I didn't want the situation where

```
        var depCtr *Container
	if c.config.NetNsCtr != "" { // **TRUE**
		// ignoring the error because there isn't anything to do
		depCtr, _ = c.getRootNetNsDepCtr()
	} 
	if len(c.state.NetworkStatus) != 0 { // **TRUE**
		depCtr = c
	}
```

But If that is not possible / something to worry about I can make them both `if`'s and remove the unnecessary block

libpod/networking_linux.go
bblenard: I have changed this I just don't want to kick off another CICD run until I get the other things sorted out.

libpod/container_internal_linux.go
bblenard: https://github.com/containers/podman/pull/9972#discussion_r615246254

libpod/container_internal_linux.go
bblenard: @rhatdan bump just in case you didn't see my response

libpod/container_internal_linux.go
bblenard: @rhatdan just in case you didn't see my response.

test/system/500-networking.bats
bblenard: @AkihiroSuda bump in case you didn't see my last response

libpod/container_internal_linux.go
rhatdan: Yes, that is fine.

TomSweeneyRedHat:
@rhatdan can you merge this if it LGTY?  I think there was one last "is this OK" type of question for you in the review comments on this one.

test/system/500-networking.bats
AkihiroSuda: Having an option SGTM

AkihiroSuda:
Bikeshedding:  For consistency with `host.docker.internal`, the corresponding value for Podman should be `host.podman.internal` or `host.containers.internal` rather than `host.gateway.internal`?

rhatdan:
Yes lets go with `host.containers.internal` and then we can merge.

bblenard:
I just pushed up some code and assuming the tests all pass there are just a few small things I want to make sure everyone is on the same page about:

1) [this thread](https://github.com/containers/podman/pull/9972#discussion_r610340456). @AkihiroSuda mentioned their concerns about using the slirpns gateway address, we discussed a bit and contemplated using a configuration option to specify an ip address for the `host.containers.internal` address. @rhatdan / @mheon / @TomSweeneyRedHat thoughts? If that is the route we want to go I'll have to rework most of this code :)

2) @rhatdan I wasn't sure what ["Yes, that is fine."](https://github.com/containers/podman/pull/9972#discussion_r620705065) meant exactly so I left the code as it was. I can change it if I mis-interpreted your response though.

AkihiroSuda:
The commit message ( host.gateway.internal ) seems to need to be updated

bblenard:
> The commit message ( host.gateway.internal ) seems to need to be updated

Fixed f5e7b40d66646b3e6517d989d8078110172808de

rhatdan:
LGTM

rhatdan:

libpod/networking_slirp4netns.go
```diff
@@ -314,9 +314,53 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
...
+		expectedGatewayIP, err := addToIP(ipv4, ipv4network, uint32(2))
+		if err != nil {
+			return errors.Wrapf(err, "error calculating expected gateway ip for slirp4netns")
+		}
+		expectedDNSIP, err := addToIP(ipv4, ipv4network, uint32(3))
+		if err != nil {
+			return errors.Wrapf(err, "error calculating expected dns ip for slirp4netns")
+		}
+		// Save off calculated address into globals. I don't like this either
+		slirp4netnsIP = expectedIP.String()
```
Luap99, line 337: I don't think we should use globals, this will fail when used with the podman service.

libpod/networking_slirp4netns.go
```diff
@@ -314,9 +314,53 @@ func (r *Runtime) setupSlirp4netns(ctr *Container) error {
 		}
 		return r.setupRootlessPortMappingViaRLK(ctr, netnsPath)
 	}
+
+	// Calculate Slirp addresses now that we are pretty sure the command executed
```
Luap99, line 318: This block has to above `if  havePortMapping...`

libpod/networking_slirp4netns.go
bblenard: Okay, I'll look at moving away from the globals. I think when I initially wrote this those were constant values that I changed to variables so I didn't even think about if they **should** be globals. Thanks for catching that :)

libpod/networking_slirp4netns.go
bblenard: I'll fix that once I figure out how to move away from globals like you suggested above.

libpod/networking_slirp4netns.go
Luap99: I think you can add a new field (maybe called slirp4netnsSubnet) to the container struct here https://github.com/containers/podman/blob/663ea964733f5d7ccef9b457b33e97ebdec79e70/libpod/container.go#L97
and add methods like GetSlirp4netnsIP, GetSlirp4netnsGateway and GetSlirp4netnsDNS which returns the correct ip calculated from the stored subent.

bblenard: Fixed: https://github.com/containers/podman/commit/3ce696e05426bc03a541ba6f36f8ee48080c0c66#diff-b50e7e875575fcd700a1431a314b378fc179af9a4d1da93cbb62a42d10949c0aR335

I made the helper functions generatl libpod functions that take subnets instead of Container functions that reference internal struct members because I didn't think it was a container specific thing. I also essentially reverted your changes here: https://github.com/containers/podman/commit/f99b7a314bf6d285b92a528d1a0800c35dfdc603 since I had a merge conflict and my changes covered what you did with that commit as far as I could tell.

libpod/networking_slirp4netns.go
bblenard: Fixed: c0335e77819e23be94bb55d883e18bdd6b742195

Luap99 (commented):
Thanks this looks good. One small nit.

libpod/container_internal_linux.go
```diff
@@ -1779,7 +1797,12 @@ func (c *Container) getHosts() string {
 	if c.Hostname() != "" {
 		if c.config.NetMode.IsSlirp4netns() {
 			// When using slirp4netns, the interface gets a static IP
-			hosts += fmt.Sprintf("# used by slirp4netns\n%s\t%s %s\n", slirp4netnsIP, c.Hostname(), c.config.Name)
+			slirp4netnsIP, err := GetSlirp4netnsGateway(c.slirp4netnsSubnet)
+			if err != nil {
+				logrus.Error("failed to determine slirp4netnsIP: ", err.Error())
```
rhatdan, line 1802: This should be a warning not an error,  If the user can not do anything about it, then it should just warn.

Luap99:
@edsantiago PTAL at the system tests.

libpod/container_internal_linux.go
```diff
@@ -1698,7 +1711,12 @@ func (c *Container) generateResolvConf() (string, error) {
 		nameservers = resolvconf.GetNameservers(resolv.Content)
 		// slirp4netns has a built in DNS server.
 		if c.config.NetMode.IsSlirp4netns() {
-			nameservers = append([]string{slirp4netnsDNS}, nameservers...)
+			slirp4netnsDNS, err := GetSlirp4netnsDNS(c.slirp4netnsSubnet)
+			if err != nil {
+				logrus.Error("failed to determine Slirp4netns DNS: ", err.Error())
```
rhatdan, line 1716: Warning not Error.

libpod/container_internal_linux.go
```diff
@@ -1802,6 +1825,36 @@ func (c *Container) getHosts() string {
...
+		for _, pluginResultsRaw := range depCtr.state.NetworkStatus {
+			pluginResult, _ := cnitypes.GetResult(pluginResultsRaw)
+			for _, ip := range pluginResult.IPs {
+				hosts += fmt.Sprintf("%s host.containers.internal\n", ip.Gateway)
+			}
+		}
+	} else if c.config.NetMode.IsSlirp4netns() {
+		gatewayIP, err := GetSlirp4netnsGateway(c.slirp4netnsSubnet)
+		if err != nil {
+			logrus.Error("failed to determine gatewayIP: ", err.Error())
```
rhatdan, line 1850: Warning not error.

rhatdan:
LGTM once Error->Warnin

test/system/500-networking.bats
```diff
@@ -162,6 +162,24 @@ load helpers
     done
 }
 
+@test "podman run with slirp4ns assigns correct gateway address to host.containers.internal" {
+    run_podman run --network slirp4netns:cidr=192.168.0.0/24 \
```
edsantiago, line 166: Just curious: what problem does `random_rfc1918_subnet()` have that prevents you from using it?

bblenard: I wasn't aware of that function. I added it to the tests I wrote :+1:

libpod/networking_linux.go
```diff
@@ -360,15 +356,19 @@ func (r *Runtime) GetRootlessCNINetNs(new bool) (*RootlessCNI, error) {
 				}
 
 				// build a new resolv.conf file which uses the slirp4netns dns server address
-				resolveIP := slirp4netnsDNS
+				resolveIP, err := GetSlirp4netnsDNS(nil)
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to determine default slirp4netns DNS address")
+				}
+
```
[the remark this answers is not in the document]
bblenard: Fixed: ef6027b2db9d7ebd4a294af002c08c0c93d0e5db

libpod/container_internal_linux.go
bblenard: Fixed: ef6027b2db9d7ebd4a294af002c08c0c93d0e5db

libpod/container_internal_linux.go
bblenard: Fixed: ef6027b2db9d7ebd4a294af002c08c0c93d0e5db

libpod/container_internal_linux.go
bblenard: Fixed: ef6027b2db9d7ebd4a294af002c08c0c93d0e5db

Luap99 (approved):
LGTM

test/system/500-networking.bats
edsantiago: Thanks! One suggestion, though: I find string concatenation easier to read than stripping. Would you consider:
```
    subnet=$(random_rfc1918_subnet)
    ... :cidr="$subnet.0/24"
    .... grep "$subnet"
    ... nameserger "${subnet}.3"
```
That removes the need for all those complicated fragile `%.*`, and makes life much easier for future maintainers.

bblenard: Sounds good. Fixed: c8dfcce6db0adfa2f91d979271e8c2725eb753a6

test/system/500-networking.bats
```diff
@@ -162,6 +162,27 @@ load helpers
     done
 }
 
+@test "podman run with slirp4ns assigns correct gateway address to host.containers.internal" {
+    CIDR="$(random_rfc1918_subnet)"
```
edsantiago, line 166: Much better, thank you, but I'm kind of uncomfortable now with the variable name: a CIDR is defined as a full subnet, slash, and mask. A future maintainer may find this a stumbling point, and waste time (e.g. on line 169) wondering how a CIDR can have `.2` appended. This is about a 3-out-of-10 gripe, I'm not going to insist on you changing it, but I'd like to bring it to the attention of others in case someone else on the team has a strong opinion either way.

rhatdan:
Thanks @bblenard Nice work.

bblenard:
> Thanks @bblenard Nice work.

Thanks! I've been enjoying working on podman :)

PR #9977, opened 2021-04-08
buildah-bud tests: simplify

edsantiago:
Experience this week has shown that managing .diff files
is too difficult for humans, and too fragile. Opportunities
for errors abound. So, let's try to minimize the diffs.

We can't eliminate the diffs to helpers.bash: those are
true code changes that are absolutely required for running
tests using podman instead of buildah. We need to carry
those ourselves: they are not appropriate for the buildah
repo itself.

What we can do is simplify the patching of bud.bats. That
is fragile, because bud.bats changes often, and context-
sensitive git patch files can easily get confused.

Recognizing that the changes to bud.bats fall under two types:

  - tests that are skipped
  - tests in which podman error messages differ from buildah's

...we now have a new script, apply-podman-deltas, which
is (I hope) much user-friendlier. It understands two directives:

  errmsg - alter the expected error message
  skip   - skip a test

Both operate based on a bats test name. The test name must
match exactly. These directives use 'sed' to update bud.bats.
If any directive fails, the script will keep going (so you
get as many errors as possible in a run), then exits failure.

Instructions (README.md) now explain the process for dealing
with all expected test failures.

(Sneak checkin: add '--filter=NAME' option to test runner,
allowing for targeted and much shorter test runs).


test/buildah-bud/README.md
```diff
@@ -1,5 +1,4 @@
-buildah-bud tests under podman
-==============================
+# buildah-bud tests under podman
```
edsantiago, line 1: Review recommendation for this file: '...' menu (top right) -> View file, instead of trying to read the diffs

test/buildah-bud/apply-podman-deltas
```diff
@@ -0,0 +1,157 @@
+#!/bin/bash
+#
+# *** NOTE TO READER: Please skip down to "user-customizable section" below!
```
edsantiago, line 3: Review recommendation: this advice is suitable for reviewers, too. If you want, give a once-over to the bash cruft immediately below, but what I really truly care about is the actual user-maintainable `errmsg` and `skip` directives. I want to know: do you find that readable? Maintainable?

test/buildah-bud/buildah-tests.diff
```diff
@@ -1,130 +1,13 @@
...
 From: Ed Santiago <santiago@redhat.com>
 Date: Tue, 9 Feb 2021 17:28:05 -0700
 Subject: [PATCH] tweaks for running buildah tests under podman
 
 Signed-off-by: Ed Santiago <santiago@redhat.com>
 ---
- tests/bud.bats     | 22 +++++++++++++---------
  tests/helpers.bash | 28 ++++++++++++++++++++++++----
- 2 files changed, 37 insertions(+), 13 deletions(-)
+ 1 file changed, 24 insertions(+), 4 deletions(-)
```
edsantiago, line 9: You don't need to review this diff file, either. The important part is: there are no longer any diffs to `bud.bats`, only `helpers.bash`.

edsantiago (commented):
This is a lot to review; here are some suggestions for making it easier.

rhatdan:
Since I am the only one who has gone through this. I will say LGTM.
I like the change as it will simplify what has to be done.

/lgtm

We will truly know the next time buildah updates.

Issue #9979, opened 2021-04-09
```podman stats``` is broken after some time.

### ISSUE THREAD

**Issue by lemniskett** [containers/podman#9979]
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

```podman stats``` is broken after some time, with error : 
```
Error: unable to obtain cgroup stats: strconv.ParseUint: parsing "26228283908": value out of range
```
this doesn't happen when I start containers recently.
**Steps to reproduce the issue:**

1. Run containers

2. Check with ```podman stats```

3. It works

4. Wait a few hours

5. Check with ```podman stats```

6. It doesn't work

**Describe the results you received:**

Error :
```
Error: unable to obtain cgroup stats: strconv.ParseUint: parsing "26718118422": value out of range
```
**Describe the results you expected:**
```podman stats``` show stats of containers

**Additional information you deem important (e.g. issue happens only occasionally):**
This issue seems to be only in ARMv7 OS. this doesn't happen when I reinstall with Aarch64 OS.

**Output of `podman version`:**

```
Version:      3.1.0
API Version:  3.1.0
Go Version:   go1.16.2
Git Commit:   9f09fb62cba8f44c18eda84db3e72aab3ae76046-dirty
Built:        Wed Mar 31 22:33:31 2021
OS/Arch:      linux/arm
```

**Output of `podman info --debug`:**

```
host:
  arch: arm
  buildahVersion: 1.20.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: /usr/bin/conmon is owned by conmon 1:2.0.27-1
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: 65fad4bfcb250df0435ea668017e643e7f462155'
  cpus: 4
  distribution:
    distribution: archarm
    version: unknown
  eventLogger: journald
  hostname: mokoupi
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.10.27-1-ARCH
  linkmode: dynamic
  memFree: 298610688
  memTotal: 1015156736
  ociRuntime:
    name: crun
    package: /usr/bin/crun is owned by crun 0.19-1
    path: /usr/bin/crun
    version: |-
      crun version 0.19
[...]

PR #9982, opened 2021-04-09
cgroups: force 64 bits to ParseUint

giuseppe:
[NO TESTS NEEDED]

force bitsSize==64 so that the string is always parsed to a uint64
instead of using the native int size, that could be not big enough on
32 bits arches.

Closes: https://github.com/containers/podman/issues/9979


vrothberg (commented):
LGTM

Issue #4854, opened 2020-01-13
Improve logging


mheon:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Podman's debug-level logs are starting to get out of hand. We've been abusing the DEBUG level for everything that is below WARN, ignoring the INFO and TRACE levels entirely, and it's seriously affecting readability of our logs. Further, we should investigate improvements to log tagging if a debug-level log is printed; including the part of the project the log originated from (libpod, containers/storage, etc) would aid clarity significantly, as would including filename and line number in the printed logs.

**Steps to reproduce the issue:**

1. `podman run --log-level=debug -t -i --rm fedora bash`

**Describe the results you received:**

A massive block of mostly-undifferentiated text which makes it very difficult to tell what's going on

**Describe the results you expected:**

Differentiated logging levels that allow me to easily tell when critical operations happen via INFO logs, and omit minute debugging information that would should only be in TRACE logs.

baude:
@mheon make a card and lets remove from here. wdyt ?

rhatdan:
@mheon did you make a card?

mheon:
I'll make a card, but I'd prefer not to close this. This could easily be taken by a community contributor.

rhatdan:
@sujil02 Could you investigate this?  At least look at cleaning up a some of the Warning messages.

sujil02:
Hi, @mheon @rhatdan I looked into this. Understanding through the regular stack trace I retagged a bunch load of debug messages that referred to using or creating resources to Info level.

Referring to the log tagging with the file name and line number,  Logrus does support this however enabling it kind of clutters the logs messages with a lot of text. I suggest adding a flag for the --log-level to explicitly enable it is that a wise option?  wdyt Need opinions?

rhatdan:
@sujil02 Please hand this over to @ParkerVR 
I want to get to the point of running by default at level Warning.

sujil02:
> Hi, @mheon @rhatdan I looked into this. Understanding through the regular stack trace I retagged a bunch load of debug messages that referred to using or creating resources to Info level.
> 
> Referring to the log tagging with the file name and line number, Logrus does support this however enabling it kind of clutters the logs messages with a lot of text. I suggest adding a flag for the --log-level to explicitly enable it is that a wise option? wdyt Need opinions?

[...]

PR #9995, opened 2021-04-12
Fix message about runtime to show only the actual runtime

rhatdan:
Currently the debug line shows every runtime up until it finds
the correct one, confusing users on which runtime it is using.

[NO TESTS NEEDED] Since this is just debug.

Triggered by https://github.com/containers/podman/issues/4854

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -130,8 +130,8 @@ func newConmonOCIRuntime(name string, paths []string, conmonPath string, runtime
 			continue
 		}
 		foundPath = true
+		logrus.Tracef("found runtime %q", runtime.path)
 		runtime.path = path
-		logrus.Debugf("using runtime %q", path)
 		break
 	}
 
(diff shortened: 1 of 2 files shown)


libpod/oci_conmon_linux.go
```diff
@@ -131,9 +131,9 @@ func newConmonOCIRuntime(name string, paths []string, conmonPath string, runtime
 		}
 		foundPath = true
 		runtime.path = path
-		logrus.Debugf("using runtime %q", path)
```
mheon, line 134: Maybe keep this, but reword to "Found runtime %q"?

rhatdan: This Issue was based on people complaining about too much info in debugging.  Should we drop this down to trace?

mheon: SGTM

giuseppe (approved):
LGTM

saschagrunert (approved):
LGTM

PR #9996, opened 2021-04-12
cgroup: do not set cgroup parent when rootless and cgroupfs

giuseppe:
do not set the cgroup parent when running as rootless with cgroupfs,
even if cgroup v2 is used.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1947999


mheon:
LGTM

PR #9999, opened 2021-04-12
[CI:DOCS] Correct status code for /pods/create

jwhonce:
Swagger documentation reported that the API endpoint /pods/create
returned 200 while the as-built code returned 201. 201 is more
correct so documentation updated.

Tests already checked for 201 so no updated needed.


mheon:
LGTM

Issue #9983, opened 2021-04-09
poststop hook executed twice


chethanah:
/kind bug

**Description**
- poststop hook executes twice 

**Consider the following hooks example for podman**
- Hooks config `prestart-hook.json` and `poststop-hook.json`
```bash
# ls /etc/containers/oci/hooks.d/
poststop-hook.json  prestart-hook.json

# jq .  /etc/containers/oci/hooks.d/prestart-hook.json
{
  "version": "1.0.0",
  "hook": {
    "path": "/root/podman-hook/prestart-hook.sh"
  },
  "when": {
    "annotations": {
      "testhook": "true"
    }
  },
  "stages": [
    "prestart"
  ]
}

# jq .  /etc/containers/oci/hooks.d/poststop-hook.json
{
  "version": "1.0.0",
  "hook": {
    "path": "/root/podman-hook/poststop-hook.sh"
  },
  "when": {
    "annotations": {
      "testhook": "true"
    }
  },
  "stages": [
    "poststop"
  ]
}
```
- `prestart-hook.sh` and `poststop-hook.sh` are executed in prestart and poststop stages respectively. Writes logs to `prestart-hook.log` and `poststop-hook.log`
```sh
root@home:~/podman-hook# cat prestart-hook.sh 
#!/bin/bash 
echo "----Prestart" >> /root/podman-hook/prestart-hook.log
date +%s >> /root/podman-hook/prestart-hook.log
#cat - | jq . >> /root/podman-hook/prestart-hook.log

root@home:~/podman-hook# cat poststop-hook.sh 
#!/bin/bash 
echo "----Poststop" >> /root/podman-hook/poststop-hook.log
date +%s >> /root/podman-hook/poststop-hook.log
#cat - | jq . >> /root/podman-hook/poststop-hook.log

# chmod 755 prestart-hook.sh poststop-hook.sh
```
- Launch container, with `--rm` option
```bash
# podman --hooks-dir="/etc/containers/oci/hooks.d/" run --annotation=testhook=true -it --rm  docker.io/library/bash
bash-5.1# ls
bin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
bash-5.1#
exit
```
- Upon inspecting the `prestart-hook.log` and `poststop-hook.log`, we can notice that the poststop hook has been executed twice.
```bash
# cat prestart-hook.log
----Prestart
1617936054

# cat poststop-hook.log
----Poststop
1617936057
[...]

PR #10000, opened 2021-04-12
Do not delete container twice

rhatdan:
10 lines above we had

	// Set ContainerStateRemoving
	c.state.State = define.ContainerStateRemoving

Which causes the state to not be the two checked states.  Since the
c.cleanup call already deleted the OCI state, this meant that we were
calling cleanup, and hence the postHook hook twice.

Fixes: https://github.com/containers/podman/issues/9983

[NO TESTS NEEDED] Since it would be difficult to tests this.  Main tests
should handle that the container is being deleted successfully.

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -616,20 +616,6 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force, remo
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

vrothberg (commented):
/lgtm

PR #10008, opened 2021-04-13
System tests: setup: better cleanup of stray images

edsantiago:
Fix a corner case in basic_setup(), where we rmi stray images.
If a test tags $IMAGE and fails to rmi by tag name, cleanup
could rmi both tag name and IID, wiping out the desired image:

   podman tag $IMAGE foo
   ...
   cleanup: rmi foo $FOO_IID   [this removes $IMAGE!]

Solution: rmi by name, but only rmi by IID if != $IMAGE.

TOTH to ypu for bringing this to my attention.


edsantiago:
@ypu PTAL

rhatdan:
LGTM

ypu:
Thanks @edsantiago LGTM and problem resolved :)

Issue #10009, opened 2021-04-13
`GO111MODULE=off go build` failure


lsm5:
https://github.com/containers/podman/blob/481556cbee53fb18e327a71726a0380481ffb767/pkg/domain/entities/types/types.go#L1

This works just fine in the regular case, but fails whenever GO111MODULE=off is used, like rpm and deb package cases.

```
$ GO111MODULE=off CGO_ENABLED=1 go build -o bin/podman ./cmd/podman
pkg/domain/entities/system.go:8:2: code in directory /home/lsm5/go/src/github.com/containers/podman/pkg/domain/entities/types expects import "github.com/docker/docker/api/types"
```

@rhatdan @vrothberg @baude @mheon  PTAL.

@cevich Going forward, we should also include a CI job for `GO111MODULE=off go build`. I can look at doing a PR for that.

lsm5:
```
$ GO111MODULE=off CGO_ENABLED=1 go build -o bin/podman ./cmd/podman
pkg/domain/entities/system.go:8:2: code in directory /home/lsm5/go/src/github.com/containers/podman/pkg/domain/entities/types expects import "github.com/docker/docker/api/types"

```

lsm5:
So, if you remove `// import "github.com/docker/docker/api/types` from pkg/domain/entities/types/{auth.go,types.go}` the build works fine with GO111MODULE=off. Just wanna confirm if that's the right fix or if it breaks something else.

lsm5:
sends PR..

PR #10010, opened 2021-04-13
fix build with with GO111MODULE=off

lsm5:
Fix build with GO111MODULE=off
    
Distro builds on Fedora and Kubic projects use GO111MODULE=off
by default which are currently failing. This commit fixes it and
going forward, podman CI will also indicate failures in rpm builds.

conmon build has also been removed from podman.spec.in because the COPR
for which it was provided has been discontinued.

Fixes: #10009

@mheon @vrothberg @rhatdan @baude @cevich PTAL

### DIFF
--- a/pkg/domain/entities/types/auth.go
+++ b/pkg/domain/entities/types/auth.go
@@ -1,4 +1,5 @@
-package types // import "github.com/docker/docker/api/types"
+// copied from github.com/docker/docker/api/types
+package types
 
 // AuthConfig contains authorization information for connecting to a Registry
 type AuthConfig struct {
--- a/contrib/spec/podman.spec.in
+++ b/contrib/spec/podman.spec.in
@@ -12,9 +12,9 @@
 %global debug_package   %{nil}
 %endif
 
-# %if ! 0% {?gobuild:1}
-%define gobuild(o:) go build -tags="$BUILDTAGS" -ldflags "${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n')" -a -v -x %{?**};
-#% endif
+%if ! 0%{?gobuild:1}
+%define gobuild(o:) go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n') -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld  '" -a -v -x %{?**};
+%endif
 
 # podman hack directory
 %define hackdir %{_builddir}/%{repo}-%{shortcommit0}
@@ -29,12 +29,6 @@
 %global git0 https://%{provider}.%{provider_tld}/%{project}/%{repo}
 %global commit0 #COMMIT#
 %global shortcommit0 %(c=%{commit0}; echo ${c:0:8})
-
-# People want conmon packaged with the copr rpm
-%global import_path_conmon      github.com/containers/conmon
-%global git_conmon      https://%{import_path_conmon}
-%global commit_conmon   41877362fc4685d55e0473d2e4a1cbe5e1debee0
-%global shortcommit_conmon %(c=%{commit_conmon}; echo ${c:0:7})
 
 Name: podman
 %if 0%{?fedora}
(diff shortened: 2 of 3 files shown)


lsm5:
let me add a CI job for GO111MODULE=off

mheon:
This seems OK, LGTM

vrothberg (commented):
LGTM

lsm5:
Let's hold off on lgtm until we have a green CI.

cevich:
@lsm5 I'm curious why this is needed.  My understanding is `GO111MODULE=off` is only necessary when using really, really, old versions of golang, no?

---

CI test-job wise, this sounds like something ripe for inclusion in `alt_build_task`.  This is a kind of catch-all for "Confirm building works in *" scenarios.  To add something new, just toss in a new matrix item to that task with a distinct `$ALT_NAME` value.  Then pick that up with a [new entry in the runner script case block.](https://github.com/containers/podman/blob/b6b0b6e8bd0f8c75fe9411f722a43f9305c6137a/contrib/cirrus/runner.sh#L208-L250)

lsm5:
> @lsm5 I'm curious why this is needed. My understanding is `GO111MODULE=off` is only necessary when using really, really, old versions of golang, no?

hmm, I never bothered to check honestly but I could dig in if necessary on that. FWIW, the default fedora `%gobuild` definition is:

```
$ rpm --eval %gobuild

  # https://bugzilla.redhat.com/show_bug.cgi?id=995136#c12
     GO111MODULE=off \
  go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n') -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld  '" -a -v -x ;
```

so I'm guessing there hasn't been any change on newer versions.

> CI test-job wise, this sounds like something ripe for inclusion in `alt_build_task`. This is a kind of catch-all for "Confirm building works in *" scenarios. To add something new, just toss in a new matrix item to that task with a distinct `$ALT_NAME` value. Then pick that up with a [new entry in the runner script case block.](https://github.com/containers/podman/blob/b6b0b6e8bd0f8c75fe9411f722a43f9305c6137a/contrib/cirrus/runner.sh#L208-L250)

So, given that the spec file is only used for checking if rpm builds fine, I changed the spec file directly without a specific CI job. Things should hopefully be changing very soon when we have the packit integration in place.

lsm5:
@cevich oh also, if we don't have GO111MODULE=off it ends up fetching all the vendored dependencies which is a no-go for non-networked koji buiilds.

cevich:
> https://bugzilla.redhat.com/show_bug.cgi?id=995136

That BZ is from 2016.

> if we don't have GO111MODULE=off it ends up fetching all the vendored dependencies which is a no-go for non-networked koji buiilds.

Oh...right, yeah, golang really wants to be connected to the internet :disappointed:

cevich:
> Things should hopefully be changing very soon when we have the packit integration in place.

Oh okay, yeah, feel free to replace that RPM-build job then.  My understand is it's mainly there for package release benefit - noticing problems early.  So if there's a better way to do that, go for it.

lsm5:
`Test build RPM` is successful with current commit, so I'm gonna push another version with `[NO TESTS NEEDED]` to handle `Validate fedora-34beta Build`. Going forward, I think there should be exceptions for cases where no tests are added/modified but tests are still required to pass.

lsm5:
ready for lgtm

pkg/domain/entities/types/auth.go
```diff
@@ -1,4 +1,4 @@
-package types // import "github.com/docker/docker/api/types"
```
rhatdan, line 1: Why did you change these?

lsm5: that's the main thing that breaks build with `GO111MODULE=off`. Please let me know if there's a better option to handle that.

mheon:
I think the comment from @rhatdan is addressed
/lgtm

lsm5:
we also need a `/hold cancel` i guess.

rhatdan:
I thought those were there to indicate where we copied the code from .  Perhaps change the import to copied from?

lsm5:
> I thought those were there to indicate where we copied the code from . Perhaps change the import to copied from?

@rhatdan I can give that a try.

lsm5:
@rhatdan added a comment on top of the package line. I wish golang didn't act funny with comments.

lsm5:
anybody know how to rerun the rdoproejct.org/github-check job? It doesn't fail locally for me on centos8

edsantiago:
rdoproject always fails. We don't care: it doesn't block merging.

lsm5:
> rdoproject always fails. We don't care: it doesn't block merging.

thanks @edsantiago . If someone could  please `/hold cancel`, this should be good for merge.

PR #10014, opened 2021-04-13
Remove in-memory state implementation

mheon:
We originally added this in the *very early* days of Podman, before a proper persistent state was written, so we had something to test with. It was retained after the original SQLite state (and current BoltDB state) were written so it could be used for testing Libpod in unit tests with no requirement for on-disk storage. Well, such unit tests never materialized, and if we were to write some now the requirement to have a temporary directory for storing data on disk is not that bad. I can basically guarantee there are no users of this in the wild because, even if you managed to figure out how to configure it when we don't document it, it's completely unusable with Podman since all your containers and pods will disappear every time Podman exits.

Given all this, and since it's an ongoing maintenance burden I no longer wish to deal with, let's just remove it.


mheon:
Shaves about 1mb off non-remote Podman binary size, which isn't terrible for the amount of code removed.

vrothberg (commented):
LGTM

saschagrunert:
The removal of the registrar package is a breaking change for us on the CRI-O side.

mheon:
You might want to throw it directly in your codebase, then - I don't think we want to keep it here if no part of Podman is using it.

PR #10017, opened 2021-04-13
compose test: try to get useful data from flakes

edsantiago:
docker-compose test continues to flake even after #9961.
Let's try to get some useful data from the failures, by:

  * adding -S (--show-error) to curl. With just -s (--silent),
    curl is completely quiet. With -S, it displays errors.
    (Not in TAP form, but I'm OK with that)

  * oops, adding safety checks to the fix from #9961 (it
    was inadvertently clobbering the curl exit status)

And, as long as I'm in this code: logformatter was not
highlighting these results, because the '1..N' TAP line
needs to be spit out at the end. Have test-compose emit
a 'TAP' header <http://testanything.org/> and make
logformatter recognize it.

### DIFF
--- a/test/compose/test-compose
+++ b/test/compose/test-compose
@@ -163,18 +163,21 @@ function test_port() {
     local op="$2"                # '=' or '~'
     local expect="$3"            # what to expect from curl output
 
-    local actual=$(curl --retry 3 --retry-all-errors -s http://127.0.0.1:$port/)
-    # The test is flaking with an empty result. The curl retry doesn't solve this.
-    # If the result is empty sleep one second and try again.
-    if [[ "$actual" == "" ]]; then
-        sleep 1
-        local actual=$(curl --retry 3 --retry-all-errors -s http://127.0.0.1:$port/)
-    fi
+    # -s -S means "silent, but show errors"
+    local actual=$(curl --retry 3 --retry-all-errors -s -S http://127.0.0.1:$port/)
     local curl_rc=$?
+
+    # FIXME 2021-04-13: test is flaking, curl succeeds but returns empty result.
+    # Could it be that the container is not actually ready? Wait, and retry.
+    if [[ $curl_rc -eq 0 && -z "$actual" ]]; then
+        sleep 1
+        echo "# Retrying curl:"
+        actual=$(curl --retry 3 --retry-all-errors -s -S http://127.0.0.1:$port/)
+        curl_rc=$?
+    fi
+
     if [ $curl_rc -ne 0 ]; then
         _show_ok 0 "$testname - curl failed with status $curl_rc"
-###        docker-compose down >>$logfile 2>&1
-###        exit 1
     fi
 
     case "$op" in
@@ -285,6 +288,10 @@ fi
 
 # Too hard to precompute the number of tests; just spit it out at the end.
 n_tests=0
+
+# We aren't really TAP 13; this helps logformatter recognize our output as BATS
+echo "TAP version 13"
+
 for t in ${tests_to_run[@]}; do
     testdir="$(dirname $t)"
     testname="$(basename $testdir)"
--- a/contrib/cirrus/logformatter
+++ b/contrib/cirrus/logformatter
@@ -243,10 +243,16 @@ END_HTML
             $cirrus_task = $1;
         }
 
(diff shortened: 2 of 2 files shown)


test/compose/test-compose
```diff
@@ -163,18 +163,20 @@ function test_port() {
...
-        local actual=$(curl --retry 3 --retry-all-errors -s http://127.0.0.1:$port/)
+        actual=$(curl --retry 3 --retry-all-errors -s -S http://127.0.0.1:$port/)
+        curl_rc=$?
     fi
-    local curl_rc=$?
+
     if [ $curl_rc -ne 0 ]; then
         _show_ok 0 "$testname - curl failed with status $curl_rc"
-###        docker-compose down >>$logfile 2>&1
-###        exit 1
```
edsantiago, line 177: Removed cruft.

edsantiago:
For posterity, this was my motivation:

* instance:GCEInstance : compose test on fedora-34beta (rootless)
  * PR #9996
    * [04-12 12:31](https://api.cirrus-ci.com/v1/task/6310475005689856/logs/main.log)

(Flake in a very recent PR).

FWIW, just from testing on my laptop, I see these once in a while:
```
curl: (7) Failed to connect to 127.0.0.1 port 5001: Connection refused
```
This tells me that `docker-compose` makes no guarantees about when the service will actually be up; which in turn tells me that we need to look for a robust container-is-actually-ready mechanism. Any ideas?

Luap99:
Thanks for looking into this. 

> This tells me that docker-compose makes no guarantees about when the service will actually be up; which in turn tells me that we need to look for a robust container-is-actually-ready mechanism. Any ideas?

docker-compose doesn't care just like podman doesn't care. The container process is started whether the service in the container is ready is never known.

I think the bug is somewhere in the rootless port forwarding logic. Would you mind turning on log level debug for the podman service and use the following diff?
```diff
diff --git a/libpod/networking_slirp4netns.go b/libpod/networking_slirp4netns.go
index 72ab3c919..de3aced62 100644
--- a/libpod/networking_slirp4netns.go
+++ b/libpod/networking_slirp4netns.go
@@ -410,6 +410,7 @@ outer:
                TmpDir:    ctr.runtime.config.Engine.TmpDir,
                ChildIP:   childIP,
        }
+       logrus.Debugf("rootless port config %v", cfg)
        cfgJSON, err := json.Marshal(cfg)
        if err != nil {
                return err
```

edsantiago:
>Would you mind turning on log level debug for the podman service and use the following diff?

As in: do that and keep running tests in this PR (without merging) until it fails? Or as in: do that, and merge the PR? Either way, what do you hope to gain by that? The server log (`$WORKDIR/server.log`) is reset for each test; and even if it wasn't, it is not preserved or displayed. But I don't remember this code that well, what am I missing?

Luap99:
Yes, if we do not display the log it is useless.

How often does it fail locally for you?

edsantiago:
Followup: colorized results: [root](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6424825422938112/html/compose-podman-fedora-34beta-root-host.log.html) and [rootless](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5017450539384832/html/compose-podman-fedora-34beta-rootless-host.log.html).

I find it interesting that root only fails with ECONNREFUSED, rootless only with ECONNRESET. Is that meaningful?

edsantiago:
>How often does it fail locally for you?

Unfortunately, it doesn't. I've never seen it fail locally; and my flake-analyze script sees very few of these in CI. That's why the goal of this PR is to see if `curl` emits something, anything useful when it flakes.

Luap99:
> I find it interesting that root only fails with ECONNREFUSED, rootless only with ECONNRESET. Is that meaningful?

I would say this is somewhat expected. As root we use iptables for port forwarding. As rootless we use rooltesskit. We should get ECONNREFUSED if the container process did not bind the port yet. I assume since rootlesskit already binds the port on the host but the container process did not bind the port yet it will respond with ECONNRESET.

Luap99:
LGTM, lets see the output when this actually flakes.

PR #10021, opened 2021-04-14
Fix invalid expression in save command

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/10024

### DIFF
--- a/test/e2e/save_test.go
+++ b/test/e2e/save_test.go
@@ -111,6 +111,24 @@ var _ = Describe("Podman save", func() {
 		Expect(save.ExitCode()).To(Equal(0))
 	})
 
+	It("podman save to directory with --compress but not use docker-dir and oci-dir", func() {
+		if rootless.IsRootless() && podmanTest.RemoteTest {
+			Skip("Requires a fix in containers image for chown/lchown")
+		}
+		outdir := filepath.Join(podmanTest.TempDir, "save")
+
+		save := podmanTest.Podman([]string{"save", "--compress", "--format", "docker-archive", "-o", outdir, ALPINE})
+		save.WaitWithDefaultTimeout()
+		// should not be 0
+		Expect(save.ExitCode()).ToNot(Equal(0))
+
+		save = podmanTest.Podman([]string{"save", "--compress", "--format", "oci-archive", "-o", outdir, ALPINE})
+		save.WaitWithDefaultTimeout()
+		// should not be 0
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM but I'd prefer to add a test for it.

zhangguanzhang:
> LGTM but I'd prefer to add a test for it.

cobra's `Args` method verifies this, so no unit testing is required
https://github.com/containers/podman/blob/4a22d6117aeb2f2b9fa120328bba857f7fd54d47/cmd/podman/images/save.go#L33-L45

zhangguanzhang:
> LGTM but I'd prefer to add a test for it.

all had done

test/e2e/save_test.go
```diff
@@ -111,6 +111,24 @@ var _ = Describe("Podman save", func() {
 		Expect(save.ExitCode()).To(Equal(0))
 	})
 
+	It("podman save to directory with --compress but not use docker-dir and oci-dir", func() {
+		if rootless.IsRootless() && podmanTest.RemoteTest {
+			Skip("Requires a fix in containers image for chown/lchown")
+		}
```
vrothberg, line 117: You can use `SkipIfRootless("...")` instead.

vrothberg: Can you elaborate on what needs to be fixed?  If there is no open issue for that yet, could you open one and reference it as `#issueNumber` in the skip message?

zhangguanzhang: > You can use `SkipIfRootless("...")` instead.

This condition is copy from this file, I think is right

zhangguanzhang: > Can you elaborate on what needs to be fixed? If there is no open issue for that yet, could you open one and reference it as `#issueNumber` in the skip message?
Fixes https://github.com/containers/podman/issues/10024

vrothberg (commented):
LGTM

test/e2e/save_test.go
TomSweeneyRedHat: @zhangguanzhang yes, your test is correct, but if you could convert to the SkipIfRootless as @vrothberg showed, it would be appreciated.  That way if we ever need to quickly find our rootless tests, we have one convenient tag to do so.

zhangguanzhang: @TomSweeneyRedHat  These two conditions are parallel logical AND, I think SkipIfRootless should not be used

Issue #9742, opened 2021-03-18
podman-remote save to directories does not work


vrothberg:
/kind bug

**Description**

`podman-remote save` to `oci-dir` and `docker-dir` does not work. The client exits with 0 but the created directories are empty.

Note: it seems that {docker,oci}-dir are not exercised in the tests which must be addressed as well.

**Steps to reproduce the issue:**

1. `podman-remote save --format=oci-dir -o oci 28f6e2705743`

2. `ls oci`

**Describe the results you received:**

The created directory `oci` is empty. The `podman system service` logs the following error:
> ERRO[0013] unable to copy to response: "read /var/tmp/save247875906: is a directory"

The client, however, does not error out and succeeds with exit code 0.

**Output of `podman version`:**
v3.0.1 and HEAD

PR #10025, opened 2021-04-14
Fixes podman-remote save to directories does not work

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/9742

### DIFF
--- a/pkg/api/handlers/libpod/images.go
+++ b/pkg/api/handlers/libpod/images.go
@@ -270,6 +270,16 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// if format is dir, server will save to an archive
+	// the client will unArchive after receive the archive file
+	// so must convert is at here
+	switch query.Format {
+	case define.OCIManifestDir:
+		query.Format = define.OCIArchive
+	case define.V2s2ManifestDir:
+		query.Format = define.V2s2Archive
+	}
+
 	switch query.Format {
 	case define.V2s2Archive, define.OCIArchive:
 		tmpfile, err := ioutil.TempFile("", "api.tar")
--- a/test/e2e/save_test.go
+++ b/test/e2e/save_test.go
@@ -79,7 +79,7 @@ var _ = Describe("Podman save", func() {
 	})
 
 	It("podman save to directory with oci format", func() {
-		if rootless.IsRootless() && podmanTest.RemoteTest {
+		if rootless.IsRootless() {
 			Skip("Requires a fix in containers image for chown/lchown")
 		}
 		outdir := filepath.Join(podmanTest.TempDir, "save")
@@ -90,7 +90,7 @@ var _ = Describe("Podman save", func() {
 	})
 
 	It("podman save to directory with v2s2 docker format", func() {
-		if rootless.IsRootless() && podmanTest.RemoteTest {
+		if rootless.IsRootless() {
 			Skip("Requires a fix in containers image for chown/lchown")
 		}
 		outdir := filepath.Join(podmanTest.TempDir, "save")


pkg/api/handlers/libpod/images.go
```diff
@@ -244,6 +244,7 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
 		Compress   bool     `schema:"compress"`
+		Remote     bool     `schema:"remote"`
```
rhatdan, line 247: Why is this needed.  It is Remote by definition?

pkg/api/handlers/libpod/images.go
```diff
@@ -270,6 +271,18 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// if format is dir and use podman-remote to request, should save to an archive
+	// client will unArchive after receive the archive file
+	// so must convert is at here
+	if query.Remote {
+		if query.Format == define.OCIManifestDir {
```
rhatdan, line 278: Would it be better to just do this on the client side and not change the server side at all.

zhangguanzhang: Server will save to dir, not respone in on file.and client will untar the archive file

rhatdan: I am saying that the client checks to see if it is remote and sends across the 

		query.Format = define.OCIArchive
                 or
		query.Format = define.V2s2Archive

And then dumps it locally.

zhangguanzhang: > I am saying that the client checks to see if it is remote and sends across the
> 
> ```
> 	query.Format = define.OCIArchive
>              or
> 	query.Format = define.V2s2Archive
> ```
> 
> And then dumps it locally.
if format is an archive, will return at these lines
https://github.com/containers/podman/blob/df6c7c2ae5fb77a557d5616f8b422c45a43fe0ff/pkg/domain/infra/tunnel/images.go#L270-L280
if is a dir, will recreive an archive of the dir, and untar at this line
https://github.com/containers/podman/blob/df6c7c2ae5fb77a557d5616f8b422c45a43fe0ff/pkg/domain/infra/tunnel/images.go#L299
The logic already exists, so I don’t need to add it

pkg/bindings/images/images.go
```diff
@@ -143,6 +143,7 @@ func Export(ctx context.Context, nameOrIDs []string, w io.Writer, options *Expor
 	for _, ref := range nameOrIDs {
 		params.Add("references", ref)
 	}
+	params.Add("remote", "true")
```
vrothberg, line 146: This parameter shouldn't be needed as it's always remote.

zhangguanzhang: Should I remove all the remote parameter?

zhangguanzhang: Should I remove all the remote parameter in all files?

vrothberg: Thanks, yes please remove it.

pkg/api/handlers/libpod/images.go
```diff
@@ -270,6 +270,16 @@ func ExportImages(w http.ResponseWriter, r *http.Request) {
...
 
+	// if format is dir, server will save to an archive
+	// the client will unArchive after receive the archive file
+	// so must convert is at here
+	if query.Format == define.OCIManifestDir {
+		query.Format = define.OCIArchive
+	}
+	if query.Format == define.V2s2ManifestDir {
+		query.Format = define.V2s2Archive
+	}
```
vrothberg, line 281: Non-blocking: a `switch query.Format {...}` may be more idiomatic.

LGTM

zhangguanzhang: had done

TomSweeneyRedHat:
LGTM
and happy test buttons

rhatdan:
Someone will have to explain to me why this is being done on the server side?  Seems like the client knows what it is doing, and the server is not doing what the caller requested.  If I write a python client and ask for a directory, I will not have the majic available to convert, and will end up with a mess.  Or am I missing something?

zhangguanzhang:
> Someone will have to explain to me why this is being done on the server side? Seems like the client knows what it is doing, and the server is not doing what the caller requested. If I write a python client and ask for a directory, I will not have the majic available to convert, and will end up with a mess. Or am I missing something?

server side always return file bytes in one file：
https://github.com/containers/podman/blob/6ff56ab50a10e2350013c4ff1bf4b8d5d7b5aa87/pkg/api/handlers/libpod/images.go#L230-L236

rhatdan:
Well outvoted and it does fix a bug, so merging.
/lgtm

Issue #10029, opened 2021-04-14
podman-remote build: can fail but exit 0 (success)


edsantiago:
[ Forked from #9751 ]

podman-remote build sometimes fails but returns exit status 0. Trivial reproducer (assumes podman system service running elsewhere):
```console
$ cat Containerfile
FROM quay.io/libpod/nosuchimage:nosuchtag
RUN echo hi
$ while :;do ../bin/podman-remote build -t build_test --pull-never . && break;done
...runs for a minute or two, then

STEP 1: FROM quay.io/libpod/nosuchimage:nosuchtag
Error: error creating build container: pull policy is "never" but "quay.io/libpod/nosuchimage:nosuchtag" could not be found locally

STEP 1: FROM quay.io/libpod/nosuchimage:nosuchtag
$ <--- back to prompt, with exit status 0. This should never happen.
```
master @ 14ec3a64d, also in podman-3.1.0-3.fc33

edsantiago:
Flake. The above reproducer works fine for the first 50-100 iterations, it just fails randomly. (Same in CI).

mheon:
Alright, it's a race then. Damn, I was hoping this one was going to be simple.

rhatdan:
I might have a fix.

edsantiago:
@rhatdan I'm sorry... no-go. It takes a lot longer than before (30 seconds instead of about 10), but the race is still there.

PR #10030, opened 2021-04-14
Fix flake on failed podman-remote build

rhatdan:
We have a race condition where podman build can fail
but still return an exit code of 0.  This PR ensures
that as soon as the build fails, the failed flag is set
eliminating the race.

Fixes: https://github.com/containers/podman/issues/10029

[NO TESTS NEEDED] Tests of failed builds are already in place, and
the elimination of the race should be enough.

### DIFF
--- a/pkg/api/handlers/compat/images_build.go
+++ b/pkg/api/handlers/compat/images_build.go
@@ -462,12 +462,17 @@ func BuildImage(w http.ResponseWriter, r *http.Request) {
 		buildOptions.Timestamp = &ts
 	}
 
+	var (
+		imageID string
+		failed  bool
+	)
+
 	runCtx, cancel := context.WithCancel(context.Background())
-	var imageID string
 	go func() {
 		defer cancel()
 		imageID, _, err = runtime.Build(r.Context(), buildOptions, query.Dockerfile)
 		if err != nil {
+			failed = true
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago I ran your tests for a few minutes and could not get this PR to fail.

edsantiago:
@rhatdan I'm sorry... no-go. I'm still seeing it fail, anywhere from 7 seconds to 208 seconds. New reproducer shows how long it takes:
```console
$ t0=$SECONDS;while :;do echo $(($SECONDS - $t0));../bin/podman-remote build -t build_test --pull-never . && break;done
```

edsantiago:
@Luap99 we got a [compose flake!](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4855067355381760/html/compose-podman-fedora-34beta-rootless-host.log.html) Rootless, and `curl` shows one ECONNRESET followed by consistent ECONNREFUSED. To me that implies that the container has died. The timestamps are consistent with that. I don't think any amount of retrying is going to help here; we need a different way to debug.

(Sorry to hijack the thread, Dan)

mheon:
This LGTM, even if it's not a complete fix for the problem - the change on its own does seem to make things better.

edsantiago:
/lgtm
/hold
Le mieux est l'ennemi du bien

PR #10031, opened 2021-04-14
compose test: ongoing efforts to diagnose flakes

edsantiago:
Yay, we got a failure with the new code (#10017). It shows
one ECONNRESET followed by a lot of ECONNREFUSED over an 8-second
period (actually 15s because of the second curl retry).

My hunch: the container itself is dying. No amount of retrying
will get anything to work. So, instead of the curl retry, if
curl fails, run 'docker-compose logs' and hope that it tells us
something useful when the test flakes again.

Also: DUH! Bitten by one of the most common bash pitfalls.
Checking exit status after 'local' will always be zero.
Split the declaration and the action into separate lines.

Also: if curl fails, return immediately. There's no point in
running the string output comparison.

Also: in _show_ok(), don't emit "actual/expect" messages
if both strings are empty.

### DIFF
--- a/test/compose/test-compose
+++ b/test/compose/test-compose
@@ -136,8 +136,11 @@ function _show_ok() {
     local expect=$3
     local actual=$4
     printf "${red}not ok $count $testname${reset}\n"
-    printf "${red}#  expected: %s${reset}\n" "$expect"
-    printf "${red}#    actual: ${bold}%s${reset}\n" "$actual"
+    # Not all errors include actual/expect
+    if [[ -n "$expect" || -n "$actual" ]]; then
+        printf "${red}#  expected: %s${reset}\n" "$expect"
+        printf "${red}#    actual: ${bold}%s${reset}\n" "$actual"
+    fi
 
     echo    "not ok $count $testname" >>$LOG
     echo    "  expected: $expect"                     >>$LOG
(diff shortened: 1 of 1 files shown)


edsantiago:
[The flaking test run](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4855067355381760/html/compose-podman-fedora-34beta-rootless-host.log.html), for posterity.

I tested the new code by hardcoding `port=12345` just before the `curl`. It triggered as expected, gave some logs (successful, because the containers are of course working fine), and returned.

I suppose we just merge this and see what happens on a test failure. @Luap99, @baude, any other suggestions for what we can run in addition to / instead of `docker-compose logs`?

test/compose/test-compose
```diff
@@ -164,20 +167,15 @@ function test_port() {
...
-        echo "# Retrying curl:"
-        actual=$(curl --retry 3 --retry-all-errors -s -S http://127.0.0.1:$port/)
-        curl_rc=$?
-    fi
-
     if [ $curl_rc -ne 0 ]; then
-        _show_ok 0 "$testname - curl failed with status $curl_rc"
+        _show_ok 0 "$testname - curl (port $port) failed with status $curl_rc"
+        # FIXME: is this useful? What else can we do to diagnose?
+        docker-compose logs
```
Luap99, line 177: docker-compose logs should be good but can you add `podman ps -a`. It would be good to know if the container is still active. Also `ss -tulpn` should show us if the port is binded by the rootlesskit port forwarder.

edsantiago: Done, thank you!

test/compose/test-compose
```diff
@@ -164,20 +167,22 @@ function test_port() {
...
-    fi
-
     if [ $curl_rc -ne 0 ]; then
-        _show_ok 0 "$testname - curl failed with status $curl_rc"
+        _show_ok 0 "$testname - curl (port $port) failed with status $curl_rc"
+        # FIXME: is this useful? What else can we do to diagnose?
+        echo "# docker-compose logs:"
+        docker-compose logs
+        echo "# podman ps -a:"
+        $PODMAN_BIN ps -a
```
Luap99, line 180: Sry I just noticed, this wont work because of the custom --root ... You have to use the podman function

edsantiago: Oh, good catch, thanks. Re-pushed.

Luap99 (commented):
LGTM

PR #10032, opened 2021-04-14
Recognize "trace" logging, and use it for lone errors at exit

nalind:
* "trace" is a valid logrus debugging level, so we should be able to tell the library to display messages logged at that level.
* change `pkg/errorhandling.JoinErrors()` to not throw away the backtrace info in errors that were wrapped using `github.com/pkg/errors` if the `go-multierror` it's formatting only contains one error
* if we're printing an error at exit, and we're at trace level or higher, use %+v instead of %v when formatting the error for display, which will show the stack trace if there is one.

The net result is that `./bin/podman rmi no-such-image` still produces
```
Error: 1 error occurred:
	* unable to find a name and tag match for no-such-image in repotags: no such image
```
but `./bin/podman --log-level=trace rmi no-such-image` now ends with
```
Error: no such image
unable to find a name and tag match for no-such-image in repotags
github.com/containers/podman/v3/libpod/image.findImageInRepotags
	libpod/image/utils.go:48
github.com/containers/podman/v3/libpod/image.(*Runtime).getLocalImage
	libpod/image/image.go:489
github.com/containers/podman/v3/libpod/image.(*Runtime).NewFromLocal
	libpod/image/image.go:136
github.com/containers/podman/v3/pkg/domain/infra/abi.(*ImageEngine).Remove
	pkg/domain/infra/abi/images.go:641
github.com/containers/podman/v3/cmd/podman/images.rm
	cmd/podman/images/rm.go:56
github.com/spf13/cobra.(*Command).execute
	vendor/github.com/spf13/cobra/command.go:852
github.com/spf13/cobra.(*Command).ExecuteC
	vendor/github.com/spf13/cobra/command.go:960
github.com/spf13/cobra.(*Command).Execute
	vendor/github.com/spf13/cobra/command.go:897
github.com/spf13/cobra.(*Command).ExecuteContext
	vendor/github.com/spf13/cobra/command.go:890
main.Execute
	cmd/podman/root.go:90
main.main
	cmd/podman/main.go:38
runtime.main
	/usr/lib/golang/src/runtime/proc.go:225
runtime.goexit
	/usr/lib/golang/src/runtime/asm_amd64.s:1371
```

### DIFF
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -21,7 +21,7 @@ var (
 	// ChangeCmds is the list of valid Change commands to passed to the Commit call
 	ChangeCmds = []string{"CMD", "ENTRYPOINT", "ENV", "EXPOSE", "LABEL", "ONBUILD", "STOPSIGNAL", "USER", "VOLUME", "WORKDIR"}
 	// LogLevels supported by podman
-	LogLevels = []string{"debug", "info", "warn", "warning", "error", "fatal", "panic"}
+	LogLevels = []string{"trace", "debug", "info", "warn", "warning", "error", "fatal", "panic"}
 )
 
 type completeType int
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

rhatdan:

giuseppe (approved):
LGTM

Issue #3761, opened 2019-08-08
podman image digest showing incorrect value in some cases


dustymabe:
/kind bug

**Description**

podman inspect sometimes shows different digest than what is correct.

**Steps to reproduce the issue:**

1. Look at the `30-x86_64` tag from the [fedora/fedora repo on quay.io](https://quay.io/repository/fedora/fedora?tab=tags) and notice the current tag is at `sha256:5bc93c7ca1c526b2a73f7c97eae15638f40dfef4b44d528f4d0374302fcb9f2b`. 

2. `sudo podman pull quay.io/fedora/fedora:30-x86_64`

3. `sudo podman inspect quay.io/fedora/fedora:30-x86_64 | jq '.[]["Digest"]'`

**Describe the results you received:**

```
$ sudo podman inspect quay.io/fedora/fedora:30-x86_64 | jq '.[]["Digest"]'
"sha256:9a1bdd9da2e0723b1a4f6f016311c3228c0fadb867ef6fdd9496ab665c1d5130"
```

**Describe the results you expected:**

```
$ sudo podman inspect quay.io/fedora/fedora:30-x86_64 | jq '.[]["Digest"]'
"sha256:5bc93c7ca1c526b2a73f7c97eae15638f40dfef4b44d528f4d0374302fcb9f2b"
```

**Additional information you deem important (e.g. issue happens only occasionally):**

This is the weird part. I've noticed this is not consistent. For example:

- My Home Fedora 29 Workstation machine (`podman-1.4.4-4.fc29.x86_64`) **DOES** have the problem
- My Work Laptop Fedora 29 Silverblue Machine (`podman-1.4.4-4.fc29.x86_64) **DOES NOT** have the problem
- My Spare Laptop Fedora 30 Siverblue Machine (`podman-1.4.4-4.fc30.x86_64`)  **DOES NOT** have the problem
- My coworkers Laptop Fedora 30 Silverblue Machine (`podman-1.4.4-4.fc30.x86_64`) **DOES** have the problem

**Output of `podman version`:**

```
$ sudo podman version
Version:            1.4.4
RemoteAPI Version:  1
Go Version:         go1.11.11
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
$ sudo podman info --debug
debug:
  compiler: gc
  git commit: ""
  go version: go1.11.11
  podman version: 1.4.4
host:
  BuildahVersion: 1.9.0
  Conmon:
[...]

PR #10033, opened 2021-04-14
rmi: don't break when the image is missing a manifest

nalind:
In `libpod/image/Image.Remove()`, don't go looking for the image's parent before actually deleting the image.  If the attempt to find the image's parent fails for any reason, we're basically stuck, when we could fail in a way that at least gets the caller closer to where they wanted to be.

In `libpod/Runtime.RemoveImage()`, if we can't determine if an image has children because of some error, assume that it doesn't instead of bailing out.

Combined, these changes allow removal of an image that's lost its manifest for whatever reason, as we're seeing in https://bugzilla.redhat.com/show_bug.cgi?id=1884043.

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -617,7 +617,8 @@ func (i *Image) TopLayer() string {
 func (i *Image) Remove(ctx context.Context, force bool) error {
 	parent, err := i.GetParent(ctx)
 	if err != nil {
-		return err
+		logrus.Warnf("error determining parent of image: %v, ignoring the error", err)
+		parent = nil
 	}
 	if _, err := i.imageruntime.store.DeleteImage(i.ID(), true); err != nil {
 		return err
--- a/test/system/330-corrupt-images.bats
+++ b/test/system/330-corrupt-images.bats
@@ -1,0 +1,134 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# All tests in here perform nasty manipulations on image storage.
+#
+
+load helpers
+
+###############################################################################
+# BEGIN setup/teardown
+
+# Create a scratch directory; this is what we'll use for image store and cache
+if [ -z "${PODMAN_CORRUPT_TEST_WORKDIR}" ]; then
+    export PODMAN_CORRUPT_TEST_WORKDIR=$(mktemp -d --tmpdir=${BATS_TMPDIR:-${TMPDIR:-/tmp}} podman_corrupt_test.XXXXXX)
+fi
+
+PODMAN_CORRUPT_TEST_IMAGE_FQIN=quay.io/libpod/alpine@sha256:634a8f35b5f16dcf4aaa0822adc0b1964bb786fca12f6831de8ddc45e5986a00
+PODMAN_CORRUPT_TEST_IMAGE_ID=961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
+
+# All tests in this file (and ONLY in this file) run with a custom rootdir
+function setup() {
+    skip_if_remote "none of these tests run under podman-remote"
+    _PODMAN_TEST_OPTS="--root ${PODMAN_CORRUPT_TEST_WORKDIR}/root"
+}
+
+function teardown() {
+    # No other tests should ever run with this custom rootdir
+    unset _PODMAN_TEST_OPTS
+
+    is_remote && return
+
+    # Clean up
+    umount ${PODMAN_CORRUPT_TEST_WORKDIR}/root/overlay || true
+    if is_rootless; then
+        run_podman unshare rm -rf ${PODMAN_CORRUPT_TEST_WORKDIR}/root
+    else
+        rm -rf ${PODMAN_CORRUPT_TEST_WORKDIR}/root
+    fi
+}
+
+# END   setup/teardown
+###############################################################################
+# BEGIN primary test helper
+
+# This is our main action, invoked by every actual test. It:
+#    - creates a new empty rootdir
+#    - populates it with our crafted test image
+#    - removes [ manifest, blob ]
+#    - confirms that "podman images" throws an error
+#    - runs the specified command (rmi -a -f, prune, reset, etc)
+#    - confirms that it succeeds, and also emits expected warnings
+function _corrupt_image_test() {
+    # Run this test twice: once removing manifest, once removing blob
+    for what_to_rm in manifest blob; do
+        # I have no idea, but this sometimes remains mounted
+        umount ${PODMAN_CORRUPT_TEST_WORKDIR}/root/overlay || true
+        # Start with a fresh storage root, load prefetched image into it.
+        /bin/rm -rf ${PODMAN_CORRUPT_TEST_WORKDIR}/root
+        mkdir -p ${PODMAN_CORRUPT_TEST_WORKDIR}/root
+        run_podman load -i ${PODMAN_CORRUPT_TEST_WORKDIR}/img.tar
+        # "podman load" restores it without a tag, which (a) causes rmi-by-name
+        # to fail, and (b) causes "podman images" to exit 0 instead of 125
+        run_podman tag ${PODMAN_CORRUPT_TEST_IMAGE_ID} ${PODMAN_CORRUPT_TEST_IMAGE_FQIN}
+
+        # shortcut variable name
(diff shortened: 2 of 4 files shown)


mheon:
Tests are red, but changes LGTM
/approve
@vrothberg PTAL

nalind:
I missed that we're not able to correctly calculate an image's "parent" after we delete it, which is what's breaking the tests.
/retitle WIP: rmi: don't break when the image is missing a manifest

rhatdan:
Can't we just make the parent not existing and expected error and just log it was debugf or warnf.  And continue on?

libpod/image/image.go
```diff
@@ -615,14 +615,14 @@ func (i *Image) TopLayer() string {
 // TODO: the force param does nothing as of now. Need to move container
 // handling logic here eventually.
 func (i *Image) Remove(ctx context.Context, force bool) error {
-	parent, err := i.GetParent(ctx)
-	if err != nil {
-		return err
```
rhatdan, line 620: If you just log.Warnf(err) here, everything else will work fine, I believe.

nalind: Tried it, and it still failed with another file-not-found I haven't chased down yet.  It's mixed in with another error, so trace logging still only gives me a backtrace that points to `pkg/errorhandling.JoinErrors()`.  Frustrating, that.

nalind: Never mind, `rmi -a` apparently doesn't use the same code path as `rmi`.

nalind:
/retitle rmi: don't break when the image is missing a manifest

nalind:
Updated tests to not try `podman system reset` when testing the remote client, which doesn't implement `system reset`, and added tests that try removing the config blob instead of the manifest.

test/system/010-images.bats
```diff
@@ -241,4 +241,94 @@ Labels.created_at | 20[0-9-]\\\+T[0-9:]\\\+Z
...
+	# the image's ID is known for the image with the digest we're about to pull
+	image=quay.io/libpod/alpine@sha256:634a8f35b5f16dcf4aaa0822adc0b1964bb786fca12f6831de8ddc45e5986a00
+	id=961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
+	run_podman pull $image
+	run_podman info -f '{{.Store.GraphRoot}}'
+	graphroot="$output"
+	rm -v ${graphroot}/*-images/$id/=$(echo -n "sha256:$id" | base64 -w0)
+	run_podman 125 images
+	run_podman rmi -f $id
+	if [[ "$PODBIN_NAME" != remote ]] ; then
```
rhatdan, line 300: Why do you need to reset storage?  And if you don't reset it on remote, why isn't it broken then?

nalind: Hmm, yeah, that should probably just be a different test.

rhatdan:
@edsantiago PTAL at the tests.

edsantiago:
Looking. I don't like the duplication, will offer a suggestion later. And I *REALLY* don't like the `system reset`... but don't see a way around it.

First things first though: the final test barfs (root) with:
```
...
   [ rc=125 (expected) ]
   # /home/esm/src/atomic/2018-02.podman/libpod/bin/podman system reset -f
   time="2021-04-19T14:45:02-06:00" level=warning msg="error determining parent of image: error reading image \"961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4\" as image: error locating item named \"manifest\" for image with ID \"961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4\": file does not exist, ignoring the error"
   A storage.conf file exists at /etc/containers/storage.conf
   You should remove this file if you did not modified the configuration.
   time="2021-04-19T14:45:02-06:00" level=error msg="unlinkat /var/lib/containers/storage/overlay-containers/1d9ac53d745d3ed1ce5d981b072ee002f52d23b7a8a0471f376aa7044044a0b0/userdata/shm: device or resource busy"
   [ rc=125 (** EXPECTED 0 **) ]
   #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   #| FAIL: exit code is 125; expected 0
   #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

nalind:
Why is there a container present?  The test doesn't create one.

edsantiago:
There are no containers: `images -a` and `ps -a` are empty. There was a stray mount, probably a leftover from some prior test run. `umount /var/lib/containers/.../shm` makes the failure go away.

edsantiago:
Tests need a pretty major rewrite; I'm going to have to finish it tomorrow, sorry. (Quick rundown: too much duplication, and we need a `teardown()` because any failure anywhere in a test will leave the entire system unusable. This is podman, where tests use `/var/lib/containers`, not buildah which uses a custom root for each test.)

edsantiago:
Help, please. To minimize `podman pull`s, I'm trying `podman save/load`, and that completely ruins everything. Basically, if I `podman load` a saved image, and `rm .../manifest`, `podman images` still complains but now it exits 0 (success) instead of 125:
```bash
#!/bin/bash

# SETUP
img=quay.io/libpod/alpine@sha256:634a8f35b5f16dcf4aaa0822adc0b1964bb786fca12f6831de8ddc45e5986a00
id=961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4

tmpdir=$(mktemp -d --tmpdir podman_manifest_test.XXXXXXX)
mkdir $tmpdir/root

PODMAN="./bin/podman --root $tmpdir/root"

$PODMAN pull -q $img
$PODMAN save -o $tmpdir/img.tar $img

# Test 1
echo
echo "Test 1: this 'podman images' will fail"
rm -v $tmpdir/root/*-images/$id/manifest
$PODMAN images
echo status=$? expect=125

umount $tmpdir/root/overlay

echo "--------------------------------------------------------------"
rm -rf $tmpdir/root
mkdir $tmpdir/root
$PODMAN load -q -i $tmpdir/img.tar

echo
echo "Test 2: this 'podman images' will succeed????"
rm -v $tmpdir/root/*-images/$id/manifest
$PODMAN images
echo status=$? expect=125
```
Is this the same as #3761? Is it even a bug? Whatever it is: how can I work around this?

nalind:
Apparently we weren't treating this as an error for images with no names attached to them.  Manually untagging the image after pulling it also sidesteps the error.  `images -a` still produces an error when it fails to retrieve labels from the image's configuration blob, which we can't find without consulting the manifest.

nalind:
Conversely, manually tagging the image after re-loading it but before removing its manifest will re-trigger the errors.

edsantiago:
That works - thank you!

edsantiago:
@nalind would you consider this instead? (Save as [330-corrupt-images.bats](https://github.com/containers/podman/files/6345333/330-corrupt-images.bats.txt))

I choose to disable entirely for podman-remote, because there's no sane way to restart the server for each test. If the code paths in podman-remote are sufficiently different that they MUST be tested, lmk, I can (grumblingly) make this work with the default root.

nalind:
I'm not particularly invested in testing with the remote-client configuration (Red Hat bug 1884043 is concerned with local podman on an OpenShift node), but the CI configuration was running the ones I was adding with it.  Removed the tests I was adding in favor of yours, rebasing, and repushing.

nalind:
Rebased.

test/system/330-corrupt-images.bats
```diff
@@ -0,0 +1,134 @@
...
+@test "podman corrupt images - rmi -f -a" {
+    _corrupt_image_test "rmi -f -a"
+}
+
+@test "podman corrupt images - image prune" {
+    _corrupt_image_test "image prune -a -f"
+}
+
+@test "podman corrupt images - system reset" {
+    _corrupt_image_test "image prune -a -f"
```
edsantiago, line 120: Oops. The nitwit you copied this from (who prefers to remain anonymous right now) seems to have forgotten to actually fix this line. Sorry.

edsantiago:
/lgtm
/hold

The aforementioned nitwit has privately told me that he will volunteer to clean up the 'prune' in a followup PR.

vrothberg (commented):
/lgtm

PR #10034, opened 2021-04-14
Fix flake on failed podman-remote build : try 2

rhatdan:
This time we are checking if the function actually succeeded,
otherwise we will report an error.

Also if we did not get the id, report unexpected failure.


edsantiago:
Nice!  What I'm seeing now is:
```
Error: stream dropped, unexpected failure
```
instead of the `pull policy is "never"` one. The important factors, though, are (1) it exited nonzero (as it should), and (2) it seems to take a lot longer to fail: 202s, 182s, 48s

LGTM!

rhatdan:

vrothberg (commented):
LGTM

PR #10045, opened 2021-04-15
compose test: diagnose flakes v3

Luap99:
From the debug output we know that rootlesskit does not bind the port
correctly. The rootlesskit port forwarder has a quite a few debug
statements so lets see the debug log when the test fails. Also check
if it binded the port inside the rootless cni namespace.

### DIFF
--- a/test/compose/test-compose
+++ b/test/compose/test-compose
@@ -173,15 +173,16 @@ function test_port() {
 
     if [ $curl_rc -ne 0 ]; then
         _show_ok 0 "$testname - curl (port $port) failed with status $curl_rc"
-        # FIXME: is this useful? What else can we do to diagnose?
-        echo "# docker-compose logs:"
-        docker-compose logs
         echo "# podman ps -a:"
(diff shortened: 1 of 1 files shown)


test/compose/test-compose
```diff
@@ -229,7 +231,7 @@ function start_service() {
         sleep 1
         _timeout=$(( $_timeout - 1 ))
     done
-    cat $WORKDIR/server.log
+    # cat $WORKDIR/server.log
```
edsantiago, line 234: Why remove this?

Luap99: This will clutter the log unreadable because it is always printed for each test run.

edsantiago:
LGTM but some sort of nix thing is down, causing all tests to fail. Restarted, with fingers crossed.

test/compose/test-compose
edsantiago: Check again. It only triggers if podman server doesn't come up.

Luap99: You are correct, I missed the early return there.

PR #10050, opened 2021-04-16
Fix possible panic in libpod/image/prune.go

Luap99:
podman image prune paniced locally for me. The error handling was not
done correctly and we could end up with a nil pointer dereference.

[NO TEST NEEDED] I have no idea how I could force an error in img.Size().


giuseppe (approved):
LGTM

PR #10056, opened 2021-04-16
[NO TESTS NEEDED] Make an advanced layer diff function private

mheon:
Noticed this while I was poking around in the runtime doing DB work. The signature of this function makes me a bit uncomfortable (why should we let people apply arbitrary diffs to layers? Seems like a good way to break things...) and it's completely unused, so make it private to remove from our public API.

### DIFF
--- a/libpod/diff.go
+++ b/libpod/diff.go
@@ -1,8 +1,6 @@
 package libpod
 
 import (
-	"io"
-
 	"github.com/containers/podman/v3/libpod/layers"
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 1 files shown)


rhatdan:
That validation code is telling you to just remove the function.

mheon:
Works for me. Re-pushed with a complete removal.

TomSweeneyRedHat:
@mheon all kinds of test unhappiness

rhatdan:
Needs [NO TESTS NEEDED]
This should be default when only code is being removed.

Issue #9088, opened 2021-01-25
auto update without registry on local images


vrothberg:
/kind feature

**Description**

After a private chat with a user on auto updates, it became clear that auto updates could also be applied without having a container registry. For instance, when images built locally, the use case remains the same. Currently, auto updates only work when the image on the remote registry has changed.

There a couple of things to think about:
* Currently, containers configured for auto updates must use fully-qualified image references. Short names are not supported. Should this change? It seems _very_ common to just use short names locally.
* Should we introduce a new auto-update policy?  Currently it's called "image" (and maybe should be aliased to "registry"). The local one could be called "local". Then we could easily support short names.

@fatherlinux @mrguitar ... do you have feedback suggestions?

mrguitar:
The majority of the "production" use cases seem to prefer separating builds from runtime environments. That said, I like this idea and it seems really useful for certain types of use cases. This would be valuable for my home systems and I can see other development-ish use cases where this would be helpful.

Agreed that short names are common for local use. I don't have a strong opinion on good options here, but would requiring "localhost/[name]" be an acceptable compromise? requiring a local policy feels cumbersome, but it's possible that's a better option.

fatherlinux:
+1 Ben. Also, since we have the secure short names feature now (or will), I think this could probably be done in a way has defense in depth and prevents an attack from a man in the middle.

I could also see this being used in an environment with really slow internet connection. It might be cheaper to deliver a Dockerfile/Containerfile out and have the image build from some local cache of packages. It's probably an edge case, but could be really convenient if you happen to be a user with *that* edge case.

rhatdan:
@vrothberg Is this is issue you wanted @ParkerVR to look at?

ParkerVR:
@rhatdan touched based with @vrothberg on this and will discuss more in depth tmmr

vrothberg:
Implemented in #10063.

PR #10063, opened 2021-04-16
Autoupdate Local

ParkerVR:
https://github.com/containers/podman/issues/9088

Introduces a new auto-update policy: "local"

Although it should probably be using NewFromLocal() and comparing image digests, the current method is to compare creation dates for autoupdate - Though I'd like to discuss before changing this system

I tested this using a modified version of the docs autoupdate instructions & it worked 'successfully'
(it required the ctr to be stopped before updating it- is this proper?)

Todo List:
        Digest-Based Diff Check ?
        Refactor/Merge similar functions
        Write Tests
        Alias "image" policy as "registry" 

Will build off of https://github.com/containers/podman/pull/9740 to write tests


pkg/autoupdate/autoupdate.go
```diff
@@ -282,6 +332,21 @@ func newerImageAvailable(runtime *libpod.Runtime, img *image.Image, origName str
...
+// newerLocalImageAvailable returns true if the container and local image are different
+func newerLocalImageAvailable(ctr *libpod.Container, img *image.Image) (bool, error) {
+
+	//ctr_digest := ctr.Digest().String()
+	//img_digest := img.Digest().String()
+
+	//return ctr_digest == img_digest, nil
+
+	img_created := img.Created()
+	ctr_created := ctr.CreatedTime()
```
vrothberg, line 344: I think we should continue comparing digests rather than using time stamps.  The latter is prone to errors.

pkg/autoupdate/autoupdate.go
```diff
@@ -174,6 +183,36 @@ func AutoUpdate(runtime *libpod.Runtime, options Options) ([]string, []error) {
 		}
 	}
 
+	// Autoupdate Local Iteration
+	for imageID, containers := range localMap {
```
vrothberg, line 187: That adds quite some duplicate code.  I think we should have only loop and make the `needsUpdate` check conditional on the policy.  We could achieve that by merging the maps into one.

`map[imageUpdate][]*libpod.Container` with
```golang
type imageUpdate struct {
    image string
    policy string
}
```

That allows for bundling an image with a policy:
```golang
for imageUpdate, containers := range myMap {
    switch imageUpdate.policy {
    case ...:
    case ...:
    }
}
```

ParkerVR:
I have a working edition of the code and made some changes to refactor - new struct including the policy and changing the naming convention of the original autoupdate to 'registry'

Here is my manual test/results:
[auto-update-local-test.txt](https://github.com/containers/podman/files/6362176/auto-update-local-test.txt)

pkg/autoupdate/autoupdate.go
```diff
@@ -33,15 +33,25 @@ type Policy string
 const (
 	// PolicyDefault is the default policy denoting no auto updates.
 	PolicyDefault Policy = "disabled"
-	// PolicyNewImage is the policy to update as soon as there's a new image found.
-	PolicyNewImage = "image"
+	// PolicyRegistryImage is the policy to update as soon as there's a new image found.
+	PolicyRegistryImage = "image"
```
vrothberg, line 37: Could you change it to "registry"?

I would love us to use this term in the future as it's less ambiguous than "image".  But (as you already did :+1:) we still need to support "image" for backwards compat.

pkg/autoupdate/autoupdate.go
```diff
@@ -33,15 +33,25 @@ type Policy string
...
 	"":         PolicyDefault,
 	"disabled": PolicyDefault,
-	"image":    PolicyNewImage,
+	"image":    PolicyRegistryImage,
+	"registry": PolicyRegistryImage,
+	"local":    PolicyLocalImage,
+}
+
+// Struct for tying a container to it's autoupdate policy
+type PolicyContainer struct {
```
vrothberg, line 52: I think this should be a private type.

pkg/autoupdate/autoupdate.go
```diff
@@ -198,14 +233,14 @@ func AutoUpdate(runtime *libpod.Runtime, options Options) ([]string, []error) {
 
 // imageContainersMap generates a map[image ID] -> [containers using the image]
 // of all containers with a valid auto-update policy.
-func imageContainersMap(runtime *libpod.Runtime) (map[string][]*libpod.Container, []error) {
+func imageContainersMap(runtime *libpod.Runtime) (map[string][]PolicyContainer, []error) {
```
vrothberg, line 236: Can you extend the comment of this function? I think it'll be helpful for future readers to understand the reasoning behind this structure:

* `policyContainer` to bundle a policy with a container
* Mapping from image to a slice of policy containers since it's possible that multiple containers refer to the same image but with different policies.

Thinking about the upper points. If the slice has `[policy A, policy B, policy A]` ... I think it would make sense to further bundle the items together such that we have to run a given policy only _once_ instead of for each item individually.

A `PolicyContainer` could look as follows:
```golang
type policyMapper struct {
   map[Policy][]*libpod.Container
}
```

Now, when executing the policies we have direct mapping `image -> policy -> containers using this policy`.

pkg/autoupdate/autoupdate.go
```diff
@@ -230,22 +261,38 @@ func imageContainersMap(runtime *libpod.Runtime) (map[string][]*libpod.Container
...
+			policyMap, exists := containerMap[id]
+			if !exists {
+				policyMap = policyMapper{
+					pMap: make(map[Policy][]*libpod.Container),
+				}
+			}
+			policyMap.pMap[policy] = append(policyMap.pMap[policy], ctr)
+			containerMap[id] = policyMap
+			// Now we know that `ctr` is configured for auto updates.
+		} else {
```
vrothberg, line 276: If we invert the branch logic, we can get rid of one branch:
```golang
if policy == PolicyDefault {
    continue
}
id, _ := ...
```

vrothberg (commented):
Looking good!

Along with tests, we also need to extend the man pages a bit.

I think that some may argue that having different policies for one image may be wrong (configuration error).  But I haven't made my mind up on that.  @ParkerVR, what do you think?  Do you think we should (in theory) yield an error if image `foo` as containers configured with `registry` and `local`?

ParkerVR:
@vrothberg to address your concern with multiple configurations, I think while this _could_ be implemented I don't see a realistic use and as we just use a difference in digests it doesn't seem particularly rational to implement - so, I'm adding an error case.
On that note, is it even possible to use multiple labels like this- 'io.containers.autoupdate=local', 'io.containers.autoupdate=registry' would not overwrite each other?

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -9,21 +9,24 @@ podman-auto-update - Auto update containers according to their auto-update polic
...
 `podman auto-update` looks up containers with a specified "io.containers.autoupdate" label (i.e., the auto-update policy).
 
-If the label is present and set to "image", Podman reaches out to the corresponding registry to check if the image has been updated.
+If the label is present and set to "registry", Podman reaches out to the corresponding registry to check if the image has been updated.
 An image is considered updated if the digest in the local storage is different than the one of the remote image.
 If an image must be updated, Podman pulls it down and restarts the systemd unit executing the container.
 
+The registry policy requires a requires a fully-qualified image reference (e.g., quay.io/podman/stable:latest) to be used to create the container.
+This enforcement is necessary to know which image to actually check and pull.
+If an image ID was used, Podman would not know which image to check/pull anymore.
```
vrothberg, line 18: Note that "registry" and "image" refer to the same policy for backwards compatibility.

test/system/250-systemd.bats
```diff
@@ -42,7 +42,7 @@ function teardown() {
 
     cname=$(random_string)
     # See #7407 for --pull=always.
-    run_podman create --pull=always --name $cname --label "io.containers.autoupdate=image" $IMAGE top
+    run_podman create --pull=always --name $cname --label "io.containers.autoupdate=local" $IMAGE top
```
vrothberg, line 45: Please create a new test for the `local` policy.  `--pull=always` and local conflict a bit and the `TODO` not below is a rather important reminder of open work items.

vrothberg (commented):
Besides the nits in the system tests, LGTM

mheon:
LGTM

Issue #10066, opened 2021-04-17
"podman ps --filter ancestor=" matches a truncated image name. Docker does not.


eriksjolund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

A truncated image name will give a match when using `docker ps --filter ancestor=`. 
This works differently on Docker.

**Steps to reproduce the issue:**

1.

```
[erik@laptop ~]$ podman ps  --filter ancestor=docker.io/library/alpin
CONTAINER ID  IMAGE   COMMAND  CREATED  STATUS  PORTS   NAMES
[erik@laptop ~]$ podman ps  --filter ancestor=docker.io/library/alpine
CONTAINER ID  IMAGE   COMMAND  CREATED  STATUS  PORTS   NAMES
[erik@laptop ~]$ podman run --rm -d  docker.io/library/alpine:latest sh -c "sleep 10000"
fe93777dc162c4bc34e2a102dc2408d58875ee777d9f0499747251f0fc6bf672
[erik@laptop ~]$ podman ps  --filter ancestor=docker.io/library/alpin
CONTAINER ID  IMAGE                            COMMAND            CREATED        STATUS            PORTS   NAMES
fe93777dc162  docker.io/library/alpine:latest  sh -c sleep 10000  4 seconds ago  Up 5 seconds ago          optimistic_galois
[erik@laptop ~]$ podman ps  --filter ancestor=docker.io/library/alpine
CONTAINER ID  IMAGE                            COMMAND            CREATED        STATUS            PORTS   NAMES
fe93777dc162  docker.io/library/alpine:latest  sh -c sleep 10000  7 seconds ago  Up 8 seconds ago          optimistic_galois
[erik@laptop ~]$ 
```

**Describe the results you received:**

```
[erik@laptop ~]$ podman ps  --filter ancestor=docker.io/library/alpin
CONTAINER ID  IMAGE                            COMMAND            CREATED        STATUS            PORTS   NAMES
fe93777dc162  docker.io/library/alpine:latest  sh -c sleep 10000  4 seconds ago  Up 5 seconds ago          optimistic_galois
[erik@laptop ~]$ 
```

**Describe the results you expected:**

```
[erik@laptop ~]$ podman ps  --filter ancestor=docker.io/library/alpin
CONTAINER ID  IMAGE                            COMMAND            CREATED        STATUS            PORTS   NAMES
[erik@laptop ~]$ 
```

[...]

PR #10074, opened 2021-04-18
Fix podman ps --filter ancestor to match exact ImageName/ImageID

flouthoc:
Reference Issue: https://github.com/containers/podman/issues/10066

Following PR brings parity between how `docker` and `podman` behaves to `ps --filter ancestor=`. Fix for https://github.com/containers/podman/issues/10066

### DIFF
--- a/pkg/domain/filters/containers.go
+++ b/pkg/domain/filters/containers.go
@@ -83,7 +83,19 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
 		return func(c *libpod.Container) bool {
 			for _, filterValue := range filterValues {
 				containerConfig := c.Config()
-				if strings.Contains(containerConfig.RootfsImageID, filterValue) || strings.Contains(containerConfig.RootfsImageName, filterValue) {
+				var imageTag string
+				var imageNameWithoutTag string
+				// Compare with ImageID, ImageName
+				// Will match ImageName if running image has tag latest for other tags exact complete filter must be given
+				imageNameSlice := strings.SplitN(containerConfig.RootfsImageName, ":", 2)
+				if len(imageNameSlice) == 2 {
+					imageNameWithoutTag = imageNameSlice[0]
+					imageTag = imageNameSlice[1]
+				}
+
+				if (containerConfig.RootfsImageID == filterValue) ||
+					(containerConfig.RootfsImageName == filterValue) ||
+					(imageNameWithoutTag == filterValue && imageTag == "latest") {
 					return true
 				}
 			}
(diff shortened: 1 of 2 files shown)


Luap99:
Thanks @flouthoc
Please add a test. This is important to prevent future regressions. I think you can add a test case here https://github.com/containers/podman/blob/master/test/e2e/ps_test.go#L258.

rhatdan:
Thanks @flouthoc 
Please add a test as @Luap99 suggested.

@edsantiago This PR seemed to have gotten past your test checker?  Any ideas?

edsantiago:
@rhatdan see the commit message :-)

flouthoc:
@edsantiago @rhatdan @Luap99  I think [NO TEST NEEDED] only allows me to bypass the part of adding additional tests for fix. The CI tests would still kick-in right ? But anyways let me add tests and remove that.

edsantiago:
@flouthoc you are correct: CI will run all tests, but will forgive you for not including a new regression test in your commit. The string is misnamed (my mistake at design time), it should have been 'no **new** tests needed'.

flouthoc:
@edsantiago Thanks  :pray:  will add regression for this

eriksjolund:
When running

```
docker run --rm -d  docker.io/library/alpine:3.12  sh -c "sleep 10000" 
docker ps --filter ancestor=docker.io/library/alpine
```
there is no match.

I haven't built this PR to test it but my gut feeling by looking at the code tells me there
will be a match when running this test in Podman.

```
podman run --rm -d  docker.io/library/alpine:3.12  sh -c "sleep 10000" 
podman ps --filter ancestor=docker.io/library/alpine
```
(3.12 is an old Alpine version).

I might be wrong as I don't know so much about Golang.

pkg/domain/filters/containers.go
```diff
@@ -83,7 +83,10 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
 		return func(c *libpod.Container) bool {
 			for _, filterValue := range filterValues {
 				containerConfig := c.Config()
-				if strings.Contains(containerConfig.RootfsImageID, filterValue) || strings.Contains(containerConfig.RootfsImageName, filterValue) {
+				// Compare with ImageID, ImageName and ImangeName ignoring tag
```
giuseppe, line 86: nit: typo in ImangeName

rhatdan:
@flouthoc Can you fix the issues found so we can merge this PR?

flouthoc:
@rhatdan Sure can i get few days. I was recently tested +ve for Covid. Feeling much better now so i should be able to wrap this in few days.

flouthoc:
@rhatdan @giuseppe I have added the test case for the fix provided in the PR.  

@eriksjolund You are right `docker` only matches on the basis of imagename if tag is `latest` for other tags full explicit match is needed. I have fixed that also in this PR.

rhatdan:
LGTM
/approve

TomSweeneyRedHat:
LGTM
@flouthoc nice work, hope you're on the upswing health wise

PR #10075, opened 2021-04-18
Add local-image-proxy command to proxy container image fetches

cgwalters:
This command runs a webserver that can fetch manifests and blobs;
it's a bit like an OCI/Docker Distribution API except we can fetch
from multiple registries.  It may be more useful to think of it
as a streaming HTTP version of `podman pull`.

The primary intended use case is for ostree+containers integration.
What we want is to fetch and *stream* the underlying tarball from a container
image that will be unpacked and used for operating system updates.
The ostree code here will do things like validate the embedded
OSTree signatures inside the tarball.

Design constraints/goals:

 - This should not read or write to the containers/image storage; we
   want to stream the result and have other code unpack it
   into the target ostree repo.  I don't want to have to think about
   any potential interactions between the operating system update mechanism/storage
   and container storage.
 - I want to stream instead of pulling the whole image to a temporary
   directory and then re-parse it.
 - But, crucially we *do* want to reuse any configuration in `/etc/containers/registries.conf`
   particularly around mirroring.  In particular for e.g. OpenShift
   it must work to set up Image Content Source Policy to pull from
   a local registry mirror of the CoreOS update stream.
 - We want to reuse the containers/image default handling for
   authentication (`~/.docker/config.json` etc.)
 - I'd really rather not vendor in all of containers/image into
   another program and have to keep it up to date
 - ostree is in C and Rust, adding Go in-process would be a *huge* change
 - By having this out of process, it can easily be runnable fully unprivileged under e.g. a transient
   uid to fully ensure it has no effect on the system and can be sandboxed.

It's of course a bonus here if we also support containers signing, etc.

The expected primary use case is that ostree tooling allocates a `socketpair()`
and passes it down to skopeo, which will run a webserver on that anonymous
socketpair.

For ease of debugging/testing, we also support listening on a localhost TCP port.

More information here: https://github.com/ostreedev/ostree-rs-ext/tree/b84955aa170bf4b5518603547f33d1aec7719f44#pulling-an-ostree-container-directly

### DIFF
--- a/cmd/podman/local_image_proxy.go
+++ b/cmd/podman/local_image_proxy.go
@@ -1,0 +1,198 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/containers/image/v5/pkg/blobinfocache"
+	"github.com/containers/image/v5/transports"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/podman/v3/cmd/podman/registry"
+	"github.com/containers/podman/v3/pkg/domain/entities"
+	"github.com/containers/podman/v3/pkg/registries"
+	"github.com/opencontainers/go-digest"
+	"github.com/spf13/cobra"
+)
+
+type localImageProxyOptions struct {
+	tlsVerify bool
+	// sockFd is file descriptor for a socketpair()
+	sockFd int
+	// portNum is a port to use for TCP
+	portNum int
+}
+
+var (
+	opts                   localImageProxyOptions
+	localImageProxyCommand = &cobra.Command{
+		Use:     "local-image-proxy [options]",
+		Short:   "Interactive proxy for fetching container images",
+		Long:    "Run a local webserver that proxies OCI Distribution HTTP requests to fetch manifests and blobs",
+		RunE:    runLocalImageProxy,
+		Args:    cobra.MaximumNArgs(0),
+		Example: `podman local-image-proxy --sockfd N`,
+	}
+)
+
+func init() {
+	// Note that the local and the remote client behave the same: both
+	// store credentials locally while the remote client will pass them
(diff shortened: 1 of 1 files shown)


cgwalters:
Other notes

- Marking this as draft because it's just something I hacked up quickly and it seems to work, but I'm looking for feedback before I do a lot more work
- AFAICS, the containers/image code only validates the blob sha256 and diffID uncompressed sha256 when copying between backends; so doing that stuff seems like it'd involve an API that is basically "pull to a `Writer` but do validation"?  Or hmm, maybe we can expose an API that saves to `docker-archive`?
- I didn't actually test that this honors registry mirroring etc.
- I also don't know the podman test suite at all and this obviously needs one

cgwalters:
Hmm and a deeper problem here is I really want this command to be runnable inside a container; all it does is HTTP requests.  But it looks like all of the containers/common machinery entangles things like "parse `/etc/containers/` stuff" with "find conmon and load storage".  OK, looks like I can kind of work around this by passing `--root $(mktemp -d)`.

rhatdan:
Should this be a `podman system` or `podman image` command? Not a big fan of the local-image-proxy name.

podman system proxy?

podman image proxy?

mheon:
This feels a bit more Skopeo than Podman... A special transport so `skopeo copy` can directly produce the output format you want?

cgwalters:
I actually originally wrote this patch for skopeo, then changed my mind because it feels like the long term direction is to avoid having lots of binaries vendoring and re-exporting the containers/ library stack.  We need to keep podman and cri-o up to date for example but skopeo isn't currently critical path for anything.

That said, I agree it does more clearly fit into skopeo so...happy to close this and reopen there if that's the consensus.

cgwalters:
OK will move to skopeo.

cgwalters:
This initiative moved to https://github.com/cgwalters/container-image-proxy for now, but I still plan to propose for skopeo at some point probably soon.

PR #10079, opened 2021-04-19
runtime: create userns when CAP_SYS_ADMIN is not present

giuseppe:
when deciding to create a user namespace, check for CAP_SYS_ADMIN instead of looking at the euid.

This is useful for running podman inside podman.

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -4,6 +4,7 @@ package rootless
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -18,6 +19,7 @@ import (
 
 	"github.com/containers/podman/v3/pkg/errorhandling"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/unshare"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
@@ -67,6 +69,15 @@ func IsRootless() bool {
 			}
 		}
 		isRootless = os.Geteuid() != 0 || os.Getenv("_CONTAINERS_USERNS_CONFIGURED") != ""
+		if !isRootless {
+			hasCapSysAdmin, err := unshare.HasCapSysAdmin()
+			if err != nil {
+				logrus.Warnf("failed to read CAP_SYS_ADMIN presence for the current process")
+			}
+			if err == nil && !hasCapSysAdmin {
+				isRootless = true
+			}
+		}
 	})
 	return isRootless
 }
@@ -142,8 +153,12 @@ func tryMappingTool(uid bool, pid int, hostID int, mappings []idtools.IDMap) err
 // namespace of the specified PID without looking up its parent.  Useful to join directly
 // the conmon process.
 func joinUserAndMountNS(pid uint, pausePid string) (bool, int, error) {
-	if os.Geteuid() == 0 || os.Getenv("_CONTAINERS_USERNS_CONFIGURED") != "" {
-		return false, -1, nil
+	hasCapSysAdmin, err := unshare.HasCapSysAdmin()
+	if err != nil {
+		return false, 0, err
+	}
+	if hasCapSysAdmin || os.Getenv("_CONTAINERS_USERNS_CONFIGURED") != "" {
+		return false, 0, nil
 	}
 
 	cPausePid := C.CString(pausePid)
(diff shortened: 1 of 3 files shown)


giuseppe:
related: https://github.com/containers/common/pull/499

cmd/podman/root.go
```diff
@@ -207,7 +207,13 @@ func persistentPreRunE(cmd *cobra.Command, args []string) error {
...
 	// 3) command doesn't require Parent Namespace
 	_, found := cmd.Annotations[registry.ParentNSRequired]
-	if !registry.IsRemote() && rootless.IsRootless() && !found {
+
+	hasCapSysAdmin, err := rootless.HasCapSysAdmin()
+	if err != nil {
+		return err
+	}
+
+	if !registry.IsRemote() && !hasCapSysAdmin && !found {
```
Luap99, line 216: The same change must be made here https://github.com/containers/podman/blob/69085570f7ebbb3768e963e2a6a31d7bb9b4ca16/cmd/podman/common/completion.go#L46

pkg/rootless/rootless.go
```diff
@@ -9,8 +9,32 @@ import (
...
 )
 
+var (
+	hasCapSysAdminOnce sync.Once
+	hasCapSysAdminRet  bool
+	hasCapSysAdminErr  error
+)
+
+// HasCapSysAdmin returns whether the current process has CAP_SYS_ADMIN.
+func HasCapSysAdmin() (bool, error) {
```
rhatdan, line 22: I think we need this in storage/pkg/unshare so that buildah could use it.

pkg/rootless/rootless_linux.go
```diff
@@ -180,17 +194,40 @@ func GetConfiguredMappings() ([]idtools.IDMap, []idtools.IDMap, error) {
...
+func copyMappings(from, to string) error {
+	content, err := ioutil.ReadFile(from)
+	if err != nil {
+		return err
+	}
+	// Both runc and crun check whether the current process is in a user namespace
+	// by looking up 4294967295 in /proc/self/uid_map.  If the mappings would be
+	// copied as they are, the check in the OCI runtimes would fail.  So just split
+	// it in two different ranges.
+	if bytes.Contains(content, []byte("4294967295")) {
```
rhatdan, line 218: Very cute.

pkg/rootless/rootless.go
giuseppe: buildah has already the equivalent check, I'll refactor podman and buildah to use the common code

mheon:
Concern: a lot of our code is wired to use different paths for rootless, but I think we really only want to do that for non-EUID 0 rootless - root + non SYS_ADMIN can (and IMO should) still use standard Podman paths.

pkg/rootless/rootless.go
giuseppe: https://github.com/containers/storage/pull/877

rhatdan:
This would only cause this in rare conditions like podman inside of a container.  We will have to study this, and see if it is an issue with using /root/.config/containers rather then /etc/containers ...

giuseppe:
reworked to use the code from containers/storage

giuseppe:
tests are green now

pkg/rootless/rootless_linux.go
```diff
@@ -67,6 +69,15 @@ func IsRootless() bool {
 			}
 		}
 		isRootless = os.Geteuid() != 0 || os.Getenv("_CONTAINERS_USERNS_CONFIGURED") != ""
+		if !isRootless {
+			hasCapSysAdmin, err := unshare.HasCapSysAdmin()
+			if err != nil {
+				logrus.Errorf("failed to read CAP_SYS_ADMIN presence for the current process")
```
rhatdan, line 75: This should be a Warnf, since the process does not fail.

rhatdan (approved):
One small change and then LGTM

rhatdan:

mheon:
Would we consider this a breaking change for Podman in a unprivileged container when run as root, given that the paths in use potentially changed? I know this isn't really a supported configuration, but I imagine someone may have figured out how to get it working?

I'm not opposed to merging if it is, but we should have release notes and troubleshooting guide ready if so.

rhatdan:
No, because rootfull podman within a container and with overlay is currently broken. and would not work.  I see this as more of a bugfix.  Basically you can not mount any thing without SYS_ADMIN or a user namespace.  So only way to get this to work would be if the users had created a user namepspace for podman to use, and this change would not break that.

rhatdan:
@giuseppe fix my Error->Warn and we can get this merged.

giuseppe:
> @giuseppe fix my Error->Warn and we can get this merged.

fixed and pushed a new version

rhatdan:
LGTM

rhatdan:

PR #10081, opened 2021-04-19
Add support for CDI device configuration

sjug:
Thanks to Renaud from Nvidia.


libpod/options.go
```diff
@@ -293,6 +293,17 @@ func WithHooksDir(hooksDirs ...string) RuntimeOption {
 	}
 }
 
+// WithCDI sets the devices to check for for CDI configuration.
+func WithCDI(devices []string) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+		ctr.cdiDevices = devices
```
mheon, line 302: This won't survive the initial start of the container, or even the first start if you use `podman create`

sjug: Hey @mheon, isn't this the same pattern for every other config function in here? How can it survive?

sjug: Thank you @mheon, moved cdiDevices to `ctr.config` for permanence as per your recommendation.

TomSweeneyRedHat:
The CI won't be happy until you add a test.  
Quick glance, the code LGTM

rhatdan:
Needs a rebase.

sjug:
Test added, but the non-CDI device path isn't happy yet.

pkg/specgen/generate/container_create.go
```diff
@@ -136,13 +138,44 @@ func MakeContainer(ctx context.Context, rt *libpod.Runtime, s *specgen.SpecGener
...
 
+func extractCDIDevices(s *specgen.SpecGenerator) []libpod.CtrCreateOption {
+	devs := make([]spec.LinuxDevice, 0, len(s.Devices))
+	var cdiDevs []string
+	var options []libpod.CtrCreateOption
+
+	for _, device := range s.Devices {
+		isCDIDevice, err := cdi.HasDevice(device.Path)
+		if err != nil {
+			// do nothing
```
giuseppe, line 161: should we at least log the error (warning or debug level)?

pkg/specgen/generate/container_create.go
```diff
@@ -136,13 +138,44 @@ func MakeContainer(ctx context.Context, rt *libpod.Runtime, s *specgen.SpecGener
...
+		}
+		if isCDIDevice {
```
giuseppe, line 163: in case of errors, what is the value for isCDIDevice?

Would it be safer to use `if err == nil && isCDIDevice`?

pkg/specgen/generate/container_create.go
sjug: This "error" will occur whenever the `--device` passed is not a CDI device.
Agreed it's worth logging somewhere, added.

zvonkok:
@RenaudWasTaken PTAL

pkg/specgen/generate/container_create.go
sjug: It can still be true and have an error, added suggestion condition.

pkg/specgen/generate/container_create.go
```diff
@@ -136,13 +138,44 @@ func MakeContainer(ctx context.Context, rt *libpod.Runtime, s *specgen.SpecGener
...
 }
 
+func extractCDIDevices(s *specgen.SpecGenerator) []libpod.CtrCreateOption {
+	devs := make([]spec.LinuxDevice, 0, len(s.Devices))
+	var cdiDevs []string
+	var options []libpod.CtrCreateOption
+
+	for _, device := range s.Devices {
+		isCDIDevice, err := cdi.HasDevice(device.Path)
+		if err != nil {
```
RenaudWasTaken, line 160: Should we bubble an error up the stack here?

RenaudWasTaken:
cc @umohnani8

umohnani8:
LGTM

PR #10093, opened 2021-04-20
[CI:DOCS] Fix Markdown table layout bugs

eriksjolund:
Interestingly GitHub's renderer of Markdown displays the table just fine here:
https://github.com/containers/podman/blob/master/docs/source/markdown/podman-run.1.md

but on the web page
https://docs.podman.io/en/latest/markdown/podman-run.1.html
the table is instead shown with the text:

First mapping step: | host UID | intermediate UID | | - | - | | UID for the user starting Podman | 0 | | 1st subordinate UID for the user starting Podman | 1 | | 2nd subordinate UID for the user starting Podman | 2 | | 3rd subordinate UID for the user starting Podman | 3 | | nth subordinate UID for the user starting Podman | n |

Comparing other "good" tables with this "bad" table led me to 
believe that a newline is needed just before the table.

podman-create.1.md had the same problem.

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -1014,6 +1014,7 @@ The first mapping step is derived by Podman from the contents of the file
 _/etc/subuid_ and the UID of the user calling Podman.
 
 First mapping step:
+
 | host UID                                         | intermediate UID |
 | -                                                |                - |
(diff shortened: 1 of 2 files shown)


jwhonce:
LGTM

zhangguanzhang (approved):

PR #10102, opened 2021-04-21
corrupt-image test: fix an oops

edsantiago:
Followup to #10033: actually implement the system reset test.
And, just out of paranoia, extend the warning-message check.


mheon:
LGTM

Issue #6412, opened 2020-05-27
Container Timeout


npmccallum:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

I want to use `podman run --rm ...` to run a container that is removed on exit. I *also* want the container to be forcibly killed if it is still running after `n` seconds.

Yes, I can write a podman state manager to call `podman stop`. But that requires me to do a lot of work. It is also racy and bug prone.

I could use the `timeout` command (from coreutils) to send a signal. But podman has only two modes for signal handling. The default mode forwards the signal to pid 1 in the container. However, pid 1 could just ignore the signal to bypass the timeout. If I use `--sig-proxy=false` then podman doesn't forward the signal to the container. But it also doesn't stop the container. Therefore, the container bypasses the timeout.

I tried looking at the `--timeout` option for `conmon`, but that [doesn't do what we need](https://github.com/containers/conmon/issues/162).

I see two ways forward.

1. Add support for `--sig-proxy=stop`. This mode would not proxy the signal to the container and would instead terminate the container (and, implicitly, do the `--rm`). Then timeout support could be implemented using the `timeout` utility from coreutils.

2. Add a new option for `--timeout=n` which would cause `podman run --rm --timeout=30` to forcibly shut down the container and remove it after 30 seconds. I think I would prefer this option since it doesn't require another process.

rhatdan:
Would the container get the SIGSTOP (Actually --stop-signal value) signal or SIGKILL?  
Most likely I just answered my own question.  Since if you want SIGKILL you can just add --stop-signal=kill

If I restart the container, does it run for another 30 seconds?

rhatdan:
@mheon @vrothberg @baude WDYT?

rhatdan:
We would need to wire this into conmon. since it is the only thing left running when you run with --detatch.
@haircommander FYI

vrothberg:
This sounds completely reasonable to me.
Naming nit: to prevent confusion, I suggest to name it `--rm-timeout`.

mheon:
[...]

PR #10119, opened 2021-04-22
Add podman run --timeout option

rhatdan:
This option allows users to specify the maximum amount of time to run
before conmon sends the kill signal to the container.

Fixes: https://github.com/containers/podman/issues/6412

### DIFF
--- a/cmd/podman/common/create.go
+++ b/cmd/podman/common/create.go
@@ -651,7 +651,7 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	createFlags.UintVar(
 		&cf.StopTimeout,
 		stopTimeoutFlagName, containerConfig.Engine.StopTimeout,
-		"Timeout (in seconds) to stop a container. Default is 10",
+		"Timeout (in seconds) that containers stopped by user command have to exit. If exceeded, the container will be forcibly stopped via SIGKILL.",
 	)
 	_ = cmd.RegisterFlagCompletionFunc(stopTimeoutFlagName, completion.AutocompleteNone)
 
@@ -696,6 +696,14 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 		`Run container in systemd mode ("true"|"false"|"always")`,
 	)
 	_ = cmd.RegisterFlagCompletionFunc(systemdFlagName, AutocompleteSystemdFlag)
+
+	timeoutFlagName := "timeout"
+	createFlags.UintVar(
+		&cf.Timeout,
+		timeoutFlagName, 0,
+		"Maximum length of time a container is allowed to run. The container will be killed automatically after the time expires.",
+	)
+	_ = cmd.RegisterFlagCompletionFunc(timeoutFlagName, completion.AutocompleteNone)
 
 	tmpfsFlagName := "tmpfs"
 	createFlags.StringArrayVar(
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -951,6 +951,12 @@ The `container_manage_cgroup` boolean must be enabled for this to be allowed on 
 
 `setsebool -P container_manage_cgroup true`
 
+#### **\-\-timeout**=*seconds*
+
+Maximimum time a container is allowed to run before conmon sends it the kill
+signal.  By default containers will run until they exit or are stopped by
+`podman stop`.
+
 #### **\-\-tmpfs**=*fs*
 
 Create a tmpfs mount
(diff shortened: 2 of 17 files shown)


cmd/podman/common/create.go
```diff
@@ -651,7 +651,7 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	createFlags.UintVar(
 		&cf.StopTimeout,
 		stopTimeoutFlagName, containerConfig.Engine.StopTimeout,
-		"Timeout (in seconds) to stop a container. Default is 10",
+		"Timeout (in seconds) to stop a container, when calling Podman stop. Default is 10",
```
mheon, line 654: Timeout (in seconds) that containers stopped by user command have to exit. If exceeded, the container will be forcibly stopped via SIGKILL.

cmd/podman/common/create.go
```diff
@@ -697,6 +697,14 @@ func DefineCreateFlags(cmd *cobra.Command, cf *ContainerCLIOpts) {
 	)
 	_ = cmd.RegisterFlagCompletionFunc(systemdFlagName, AutocompleteSystemdFlag)
 
+	timeoutFlagName := "timeout"
+	createFlags.UintVar(
+		&cf.Timeout,
+		timeoutFlagName, 0,
+		"Timeout (in seconds) to kill a container if it does not complete, by default container will not be stopped",
```
mheon, line 704: Instead of "by default..." - "A timeout of 0 means the container will not be automatically stopped"

mheon: Maybe should reword the first bit, too - "Maximum length of time a container can run. They will be stopped automatically afterwards.

libpod/define/container_inspect.go
```diff
@@ -529,6 +533,10 @@ type InspectContainerHostConfig struct {
 	IOMaximumBandwidth uint64 `json:"IOMaximumBandwidth"`
 	// CgroupConf is the configuration for cgroup v2.
 	CgroupConf map[string]string `json:"CgroupConf"`
+	// Timeout is time before container is killed by conmon
+	Timeout uint `json:"Timeout"`
```
mheon, line 537: These should be in config or hostconfig, not both

libpod/options.go
```diff
@@ -758,6 +758,20 @@ func WithStopTimeout(timeout uint) CtrCreateOption {
 	}
 }
 
+// WithTimeout sets the time to after initial stop signal is sent to the
```
mheon, line 761: I think this is wrong - "sets the maximum time a container is allowed to run"

pkg/specgen/specgen.go
```diff
@@ -83,6 +83,11 @@ type ContainerBasicConfig struct {
 	// instead.
 	// Optional.
 	StopTimeout *uint `json:"stop_timeout,omitempty"`
+	// Timeout is a maximum time in seconds the container will run before
+	// main process is sent SIGKILL.
```
mheon, line 87: So is this immediate SIGKILL, or does it use the normal stop logic (SIGTERM -> timeout -> SIGKILL)?

rhatdan: We are using the conmon --timeout flag, which is defined as

       -T, --timeout Kill container after specified timeout in seconds.

Which to me means the KILL signal gets sent.  I don't think we should make it more complicated then this.  If users want more complicated they should do it via bash.

cmd/podman/common/create.go
rhatdan: Fixed

cmd/podman/common/create.go
rhatdan: Fixed

libpod/define/container_inspect.go
rhatdan: Oops from a previous version.

docs/source/markdown/podman-create.1.md
```diff
@@ -951,6 +951,12 @@ The `container_manage_cgroup` boolean must be enabled for this to be allowed on
 
 `setsebool -P container_manage_cgroup true`
 
+#### **\-\-timeout**=*seconds*
+
+Maximimum time a container is allowed to run before conmon sends it the kill
+signal.  By default containers will run until they exit or are stopped by
+`podman stop`.
+
```
TomSweeneyRedHat, line 959: We point at conmon here and a few other places in this and the other man pages, but we don't explain it's relationship with Podman anywhere.  Not for this PR, but we should probably do so.

In the shorter term, the common(8) man page now exists.  Can you add a a ptr to that and to the other manpages in here please?

rhatdan: https://github.com/containers/podman/pull/10134

rhatdan:

umohnani8:
LGTM

ashley-cui (commented):
/lgtm

kblin:
Cool, thanks a lot!

PR #10142, opened 2021-04-26
[CI:DOCS] Fix logic for pushing stable multi-arch images

cevich:
The intention is to only push an image if there is ***NOT*** an existing
tag.  The original logic for this condition was inverted.

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -77,7 +77,7 @@ jobs:
                       docker://$PODMAN_QUAY_REGISTRY/stable | \
                       jq -r '.Tags[]')
             PUSH="false"
-            if fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
                 PUSH="true"
             fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/multi-arch-build.yaml
```diff
@@ -77,7 +77,7 @@ jobs:
                       docker://$PODMAN_QUAY_REGISTRY/stable | \
                       jq -r '.Tags[]')
             PUSH="false"
-            if fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
```
edsantiago, line 80: I still like my original version better. Would you consider:
```bash
    TAG_FOUND=$(fgrep -x "$VERSION" <<<"$ALLTAGS")
    if [[ -z $TAG_FOUND ]]; then
```
I also find line 88 confusing: it should really just read `::true`, not `::${PUSH}`. My brain resents having to dereference `$PUSH` when there is only one possible value for it.

cevich: > Would you consider:

I did consider it, but concluded this is going to be TOMato:tomATO.

> My brain resents having to dereference $PUSH when there is only one possible value for it.

Oh yes I agree that would improve readability, I'll fix it.

cevich: Okay, fixed.

edsantiago: Shrug. Okay then, would you consider eliminating the unnecessary `if` duplication? The complicated duplicate statements could more readable be collapsed into one, and you can get rid of the unnecessary `PUSH` variable.
```bash
    if whatever; then
        echo "Will push $FQIN"
```

rhatdan:
/approve
LGTM

TomSweeneyRedHat:
LGTM

Issue #10116, opened 2021-04-22
podman pull doesn't use default --cert-dir as indicated in the manual


micah:
/kind bug

**Description**

The manual page for podman pull details this option:

```
   --cert-dir=path
       Use certificates at path (*.crt, *.cert, *.key) to connect to the registry.  Default certificates directory is /etc/containers/certs.d. (Not available for remote commands)
```
however, it doesn't appear that `--cert-dir=/etc/containers/certs.d` is actually the default for this command. 

**Steps to reproduce the issue:**

1. Create a registry with a certificate signed by a private CA (for example by deploying https://github.com/rpardini/docker-registry-proxy)

2. put the CA certificate in /etc/containers/certs.d/additional.crt

3. attempt to pull from that registry
4. try again, but this time pass `-certs-dir=/etc/containers/certs.d` to the pull command

**Describe the results you received:**

I have a CA cert in `/etc/containers/certs.d/additional.crt` and podman pull, without the flag specified, complains about the validity of the x509 certificate:

```
Error: Error initializing source docker://registry.example.com/docker/grafana:master: error pinging docker registry registry.example.com Get "https://registry.example.com/v2/": x509: certificate signed by unknown authority
```

**Describe the results you expected:**

If I pass manually `--cert-dir=/etc/containers/certs.d` then the pull does manage to validate the certificate and pull properly:

```
podman pull --cert-dir=/etc/containers/certs.d registry.example.com/docker/grafana:master
Trying to pull registry.example.com/docker/grafana:master...
Getting image source signatures
Copying blob 74dd2fd0d41c skipped: already exists  
Copying blob a076a628af6f [--------------------------------------] 0.0b / 0.0b
Copying blob 22c2d8e413a3 [--------------------------------------] 0.0b / 0.0b
Copying config 19d65b7b68 done  
Writing manifest to image destination
Storing signatures
19d65b7b68ed4b0b3c223b861d6f27cb502215d367b8c072579266af532841d9
root@host3:/home/vagrant# 
```

I would expect that I would not need to pass `--cert-dir` as this is supposedly the default option. 

**Additional information you deem important (e.g. issue happens only occasionally):**

I also see no place where I can configure this option as a default other than having to put it in every pull command.

**Output of `podman version`:**

```
root@host3:/home/vagrant# podman version
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.14
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
```

[...]

PR #10147, opened 2021-04-27
migrate Podman to containers/common/libimage

vrothberg:
Migrate the Podman code base over to `common/libimage` which replaces
`libpod/image` and a lot of glue code entirely.

Note that I tried to leave bread crumbs for changed tests.

Miscellaneous changes:

 * Some errors yield different messages which required to alter some
   tests.

 * I fixed some pre-existing issues in the code.  Others were marked as
   `//TODO`s to prevent the PR from exploding.

 * The `NamesHistory` of an image is returned as is from the storage.
   Previously, we did some filtering which I think is undesirable.
   Instead we should return the data as stored in the storage.

 * Touched handlers use the ABI interfaces where possible.

 * Local image resolution: previously Podman would match "foo" on
   "myfoo".  This behaviour has been changed and Podman will now
   only match on repository boundaries such that "foo" would match
   "my/foo" but not "myfoo".  I consider the old behaviour to be a
   bug, at the very least an exotic corner case.

 * Futhermore, "foo:none" does *not* resolve to a local image "foo"
   without tag anymore.  It's a hill I am (almost) willing to die on.

 * `image prune` prints the IDs of pruned images.  Previously, in some
   cases, the names were printed instead.  The API clearly states ID,
   so we should stick to it.

 * Compat endpoint image removal with _force_ deletes the entire not
   only the specified tag.


vrothberg:
Note: this does _not_ yet implement an event system in `libimage`.  I want to tackle that _after_ this PR has been merged. Then, I'll update Buildah to the latest `libimage`.

vrothberg:
@mheon @baude PTAL.

I now massaging the changes until they pass CI. But early feedback is good feedback.

vrothberg:
@jwhonce if you have some cycles to check why `podman/test/e2e/create_test.go:634` is failing, I'd love to have your eyes on it.

I cannot reproduce locally and this seems to be the last remote fart.

vrothberg:
Now all tests _but_ the buildah-bud one should pass.  The reason for that is that saving/loading is now smarter than before which breaks exactly one bud test.

Once the updates to libimage are merged into c/common, we can merge it into Buildah and then vendor buildah and c/common here.

vrothberg:
Note: this should fix https://github.com/containers/podman/issues/10116.

vrothberg:
Events for image operations will be added in a follow-up PR.

vrothberg:
@mheon, can you have a look at the libpod changes?

mheon:
Looks fine to me

test/system/070-build.bats
```diff
@@ -698,7 +699,7 @@ EOF
     # we're happy.
     if ! is_remote; then
         is "$output" \
-           ".* pull policy is .never. but .* could not be found locally" \
+           "Error: error creating build container: quay.io/libpod/nosuchimage:nosuchtag: image not known" \
```
edsantiago, line 702: This is causing the rootless CI failures. And as I look at it closely, I think this is a regression and a bug: this test should never emit "image not known" as a failure. You should not be touching this line. If you are touching this line, it means your PR is breaking `--pull-never`.

vrothberg: Thanks for checking, will address it now.  It's not a bug but the error message has changed (everything behind `podman image *` is new).  If you want to keep the error message, please let me know and I will change it with the next iteration of libimage.

vrothberg:
> Looks fine to me

Thanks for checking!

test/system/070-build.bats
edsantiago: "image not known" is ambiguous: I can't tell if it means "I do not have that image locally, and cannot check because of --pull-never", or "I do not have that image locally, and I went and checked anyway, and did not find it in the registry".

vrothberg:
@containers/podman-maintainers should turn green now

test/system/070-build.bats
```diff
@@ -393,9 +393,9 @@ Labels.$label_name | $label_value
        "image tree: third line"
     is "${lines[3]}" "Image Layers" \
        "image tree: fourth line"
-    is "${lines[4]}"  "...  ID: [0-9a-f]\{12\} Size: .* Top Layer of: \[$IMAGE]" \
+    is "${lines[4]}"  ".* ID: [0-9a-f]\{12\} Size: .* Top Layer of: \[localhost/build_test:latest]" \
        "image tree: first layer line"
-    is "${lines[-1]}" "...  ID: [0-9a-f]\{12\} Size: .* Top Layer of: \[localhost/build_test:latest]" \
+    is "${lines[-1]}" ".* ID: [0-9a-f]\{12\} Size: .* Top Layer of: \[$IMAGE]" \
        "image tree: last layer line"
```
edsantiago, line 399: Just a comment, not a change request: this seems weird, like a deliberate reordering/reversal of image tree. Isn't this a breaking change? Will users be affected?

rhatdan:
Fixes: https://github.com/containers/podman/issues/10116

edsantiago:
Fix for bud test:
```diff
diff --git a/test/buildah-bud/apply-podman-deltas b/test/buildah-bud/apply-podman-deltas
index 9f6f38190..57cca7a04 100755
--- a/test/buildah-bud/apply-podman-deltas
+++ b/test/buildah-bud/apply-podman-deltas
@@ -137,7 +137,8 @@ skip "podman requires a directory, not a Dockerfile" \

 # ...or due to Ed's laziness
 skip "Too much effort to spin up a local registry" \
-     "bud with encrypted FROM image"
+     "bud with encrypted FROM image" \
+     "bud --authfile"

 # ...or due to a fundamental arg-parsing difference between buildah and podman
 # which we could and perhaps should fix in the buildah repo via:
@@ -146,6 +147,10 @@ skip "Too much effort to spin up a local registry" \
 skip "FIXME FIXME FIXME: argument-order incompatible with podman" \
      "bud-squash-hardlinks"

+skip "FIXME FIXME FIXME we'll figure these out later" \
+     "bud-multi-stage-nocache-nocommit" \
+     "bud with --cgroup-parent"
+
 ###############################################################################
 # BEGIN tests which are skipped due to actual podman bugs.
 skip "FIXME: podman #9915" \
```
[or, download [buildah-bud-skip.diff.txt](https://github.com/containers/podman/files/6423855/buildah-bud-skip.diff.txt)]

The `--authfile` one is a permanent skip. The other two will have to be fixed in buildah itself: both are newly-added or -edited tests, and I guessI didn't review them properly for working under podman. That is a problem for later (I will take it on).

test/system/070-build.bats
vrothberg: Sorry for not mentioning that in the commit message here.  I only mentioned it in the last libimage commit (https://github.com/containers/common/commit/c53283f5fc9a7bdbf727478bc5e6e32ce490524a).

It was a deliberate reordering.  I had to rewrite image tree to make it use the internal layer tree (used to compute parent-child relationships).  When looking closer at it, the previous order did not feel intuitive as the top layer was at the very bottom.  In practice, however, layers are put on top of each other.  It also came with the nice benefit of making the code much simpler.

I see the output of `image tree` as something like the pull/push progress bars which may change over time (rather cosmetic) but others may feel different.

I will make it a PS in today's meeting.

vrothberg:
Thank you very much, @edsantiago!

vrothberg:
```
[+0100s] not ok 8 bud with --layers and --no-cache flags
[...]
[+0100s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+0100s] # #|     FAIL: buildah inspect --format {{index .Docker.History 2}} test1
[+0100s] # #| expected: 'FROM localhost/alpine:latest'
[+0100s] # #|   actual: '{2021-05-05 07:37:21.000305695 +0000 UTC  /bin/sh -c mkdir /hello FROM docker.io/library/alpine:latest false}'
```

One last to go.  Once Podman is updated in the Buildah CI VMs, we'll run into the same issue there.  Some background: previously `podman save` would tag `alpine` as `localhost/alpine` in the tar archive (even if the local image is actually docker.io/library/alpine).  This has been fixed in libimage, hence the error.

While it changes previous behavior, I see it as fixing a bug.

test/system/070-build.bats
vrothberg: I did some more detailed testing of `image tree` and found some other issues in `--whatrequires`.  I will address them in a follow-up PR.

vrothberg:
Ready to be merged

libpod/define/errors.go
```diff
@@ -12,15 +12,6 @@ var (
 	// ErrNoSuchPod indicates the requested pod does not exist
 	ErrNoSuchPod = errors.New("no such pod")
 
-	// ErrNoSuchImage indicates the requested image does not exist
-	ErrNoSuchImage = errors.New("no such image")
```
rhatdan, line 16: This is a breaking change to the API, should we atlease set these constants to point at the libimage constants?

rhatdan (approved):
LGTM
Except for constants in define directory.

libpod/define/errors.go
vrothberg: Which API do you mean?

rhatdan:
As Nancy Pelosi famously said ""We have to pass the bill so that you can find out what is in it." 

Meaning this PR is so huge we probably have to merge it and hope that the CI/CD system was good enough to catch regressions, then we need to cut a beta release and request as many users as possible use it, to find if we have regressions.

libpod/define/errors.go
vrothberg: `libimage` forwards all errors from `containers/storage`, so it would be `storage.ErrImageUnknown` but there's really no user of that error in libpod/Podman.

rhatdan: If we change this then we should bump the version of podman/libpod to v3, if we are following the rules.

vrothberg:
> As Nancy Pelosi famously said ""We have to pass the bill so that you can find out what is in it."
> 
> Meaning this PR is so huge we probably have to merge it and hope that the CI/CD system was good enough to catch regressions, then we need to cut a beta release and request as many users as possible use it, to find if we have regressions.

Totally agree.  It's too big to review entirely.  I think we can cut a first RC next week.  I want the team to test manually before.

Issue #9803, opened 2021-03-24
Podman API incompatibility findings for Docker API v1.40


JBartscher:
Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)

/kind bug

**Description**
### Podman API incompabilities with Docker API

Podman provides an API which aims to be compatible to the **v1.40** Docker API. We found some incompatibles and wanted to share them with you.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
▶ podman version                    
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.15.2
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64

```

**Output of `podman info --debug`:**

```
▶ podman info --debug
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: journald
  hostname: jbartscher-Latitude-3410
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.8.0-45-generic
  linkmode: dynamic
  memFree: 6258999296
  memTotal: 33423282176
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.18.1-7931a-dirty
      commit: 7931a1eab0590eff4041c1f74e2844b297c31cea
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #10149, opened 2021-04-27
Allow docker volume create API to pass without name

rhatdan:
The Docker API does not require Volume name to be specified when
creating a volume.

Fixes: https://github.com/containers/podman/issues/9803

### DIFF
--- a/pkg/api/handlers/compat/volumes.go
+++ b/pkg/api/handlers/compat/volumes.go
@@ -96,11 +96,17 @@ func CreateVolume(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// See if the volume exists already
-	existingVolume, err := runtime.GetVolume(input.Name)
-	if err != nil && errors.Cause(err) != define.ErrNoSuchVolume {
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/compat/volumes.go
```diff
@@ -96,11 +96,17 @@ func CreateVolume(w http.ResponseWriter, r *http.Request) {
...
-	// See if the volume exists already
-	existingVolume, err := runtime.GetVolume(input.Name)
-	if err != nil && errors.Cause(err) != define.ErrNoSuchVolume {
-		utils.InternalServerError(w, err)
-		return
+	var (
+		existingVolume *libpod.Volume
+		err            error
+	)
+	if len(input.Name) != 0 || utils.IsLibpodRequest(r) {
```
mheon, line 103: It's safe to drop the `IsLibpodRequest` bit, no reason this shouldn't work everywhere.

rhatdan: Fixed.

TomSweeneyRedHat:
LGTM

PR #10150, opened 2021-04-27
[CI:DOCS] Update container image docs + fix unstable execution

cevich:
Update the order of image documentation to be from most to least stable.
Similarly, avoid depending on execution of upstream podman, when
building/pushing.  It's easily possible for this build to function but
execution to fail due to some partially implemented feature.


cevich:
@TomSweeneyRedHat PTAL, this changes behavior to favor pushing a possibly non-functional 'upstream' podman image vs not pushing anything ([example failure](https://github.com/containers/podman/runs/2445972598?check_suite_focus=true#step:7:53)).  Since nobody's watching these daily builds, I figured it was better to have builds/pushes happening than potentially hosting a stale 'upstream' image for indefinite time periods.

contrib/podmanimage/README.md
```diff
@@ -16,11 +16,12 @@ default to `/`.
...
-  * quay.io/podman/stable:version - This image is built manually using a Fedora based container.  An RPM is first pulled from the [Fedora Updates System](https://bodhi.fedoraproject.org/) and the image is built from there.  For more details, see the Containerfile used to build it, [podmanimage/stable/manual/Containerfile](stable/manual/Containerfile).
+  * `quay.io/containers/podman` - This image is built using the latest stable version of Podman in a Fedora based container.  Built with [podmanimage/stable/Dockerfile](stable/Dockerfile).
+  * `quay.io/podman/stable` - This image is built using the latest stable version of Podman in a Fedora based container.  Built with [podmanimage/stable/Dockerfile](stable/Dockerfile).
+  * `quay.io/podman/stable:<version>` - This image is built and pushed daily, when a new podman version becomes available in Fedora.  An RPM is first pulled from the [Fedora Updates System](https://bodhi.fedoraproject.org/) and the image is built from there.  For more details, see the Containerfile used to build it, [podmanimage/stable/manual/Containerfile](stable/manual/Containerfile).
```
TomSweeneyRedHat, line 21: Just double checking, just `quay.io/containers/podman:latest` and `quay.io/podman/stable:latest` are pushed daily.  The `quay.io/podman/stable:<version>` should only be pushed if not present on the repo....

cevich: >  just quay.io/containers/podman:latest and quay.io/podman/stable:latest are pushed daily

Actually, I've been struggling with this.  We only push a tag 'master' and not 'latest'.  I can update the docs, and/or the code to also push 'latest' if you like.

> The quay.io/podman/stable:<version> should only be pushed if not present on the repo.

Correct, [that's this bit of code here.](https://github.com/containers/podman/pull/10150/files#diff-691cbdaf78b3bc338d4c6458c8d7eccae7f3a89afbc1ed27674bbf1fe53aa33bR75-R83).  The build happens daily, but we only push if the podman version (as reported by podman) doesn't have a tag.

TomSweeneyRedHat: @cevich, I'd not push a master tag, if we can, I'd let that whither on the vine.  Let's just push the `:latest` tag.
I'll go suggest a change to the `:<version>` doc here, it's not currently clear when that gets pushed.

cevich: Okay, fixed both.  PTAL when you get a chance.

cevich:
@TomSweeneyRedHat sorry, it just occurred to me: For stable, we never push a `latest` tag, only the versioned tag (if it doesn't exist).  Fix this can be done but it's a bit tricky, I'll need to do some refactoring...

cevich:
...done.  I also tweaked the docs to show we always push a 'latest' tag for all images.

rhatdan:
LGTM
/approve

.github/workflows/multi-arch-build.yaml
```diff
@@ -65,59 +66,51 @@ jobs:
...
-              echo "Will push $FQIN"
-              echo "::set-output name=podman_push::true"
-              echo "::set-output name=podman_fqin::${FQIN}"
+            # quay.io/podman/stable:vX.X.X and :latest
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+              # Only push if version tag does not exist
+              FQIN="$PODMAN_QUAY_REGISTRY/stable:$VERSION,$PODMAN_QUAY_REGISTRY/stable:latest"
             else
-              echo "Not pushing, $FQIN already exists."
+              FQIN="$PODMAN_QUAY_REGISTRY/stable:latest"
```
edsantiago, line 84: I've been trying hard to understand this, and unless I'm missing something fundamental, I think this is wrong. IIUC the logic here is:
```
if there's a new tag vX.Y.Z
    push podman/stable:vX.Y.Z
else
    push podman/stable:latest
```
Is my understanding correct? If so, I really think this is wrong. It violates POLA. What I believe should happen is:
```
if there's a new tag vX.Y.Z
    push podman/stable:vX.Y.Z
    push the same image as podman/stable:latest
```
That is: `stable:latest` should always be identical to a tagged image.

Or have I misunderstood something?

cevich: Thanks for taking a closer look, it's easily possible I've got it wrong.  However, the description in the README.md is correct:

1. If new version (doesn't exist in quay), push it ***and*** the `latest` tag[1]
2. Otherwise, only push `latest` tag.

That said, I'll go double-check this section and fix/add comments accordingly.

[1] Note: It's *possible* a new out-of-order version is pushed to master, in which case "latest" will briefly be a lie (until the next day, when the job runs again).  I'm simply ignoring this corner-case for now.

cevich: Okay, it's a bigger diff now, but I think it's much clearer this way.  PTAL @edsantiago

.github/workflows/multi-arch-build.yaml
```diff
@@ -65,69 +72,69 @@ jobs:
...
-              echo "::set-output name=podman_fqin::${FQIN}"
-            else
-              echo "Not pushing, $FQIN already exists."
+            # New image? Push quay.io/podman/stable:vX.X.X and :latest
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+              # Assume version-tag is also the most up to date (i.e. "latest")
+              FQIN="$PODMAN_QUAY_REGISTRY/stable:$VERSION,$PODMAN_QUAY_REGISTRY/stable:latest"
+            else # Not a new version-tagged image
+              # Assume other contents changed, so this is the "new" latest.
+              FQIN="$PODMAN_QUAY_REGISTRY/stable:latest"
```
edsantiago, line 93: This is what bothers me. As a naïve end user, I do not want "other contents changed". I do not want `stable:latest` to be some unpredictable unknown random thing that differs from `stable:<most-recent-tag>`. I want `stable` to be... stable.

I realize I'm arriving late into this game, and may not have full context of what these images are used for, and there may be something obvious I'm missing. But this really seems like a huge deal to me, and I need to understand our intention before I can continue reviewing.

cevich: Oh I see what you mean now.  So to answer (maybe) the most important question (@TomSweeneyRedHat please correct me):

> what these images are used for

Nothing and (probably) nobody uses them.  How do we know this?  There was a serious problem in the prior implementation, where the `podman` image was actually built using the buildah repo. (yep, you read that correctly).  This went on for months and months, and we got nary a single complaint :grin: 

This github action is a royal PITA to debug.  Look, this is the 3rd PR just to get it kind of working properly.  It can only actually execute on master :confounded:   Adding any more script newfangles at this point scares me as a months-long endeavor :cry: 

So here's what I propose:

Let's get this (imperfect) PR in, and any more required to make things sort-of functional.  Then I need to think about fixing buildah, and skopeo which both already has a substantially similar (and worse broken) implementations.  Part of that journey should include simplifying a lot of this script-in-YAML into shared code it's not duplicated 3x.  That then would be a fantastical place to sneak in tweaks for the proper handling of "latest" as you describe.

Sound reasonable?

cevich:
[Oof.  Have I mentioned how much I dislike working with github-actions?](https://github.com/containers/podman/actions/runs/798759907)

cevich:
Fixed by #10183

.github/workflows/multi-arch-build.yaml
rhatdan: The stable version of this container image is just built with the released version of Podman for the current Fedora.

Testing would be just the image that is in updates-testing, which means most of the time they are the same.

upstream would have the latest podman built from the upstream bits.

cevich: That's Correct @rhatdan .  The issue @edsantiago raised is regarding the "latest" tagged image.  He believes they should match the most recent "version-tagged" image.  As-written currently, the "latest" images are built and pushed daily.

[For example, in this most recent successful run]( https://github.com/containers/podman/actions/runs/808065332), The `quay.io/podman/stable:latest` image is now F34 w/ podman v3.1.2 (and will update daily).  Whereas the most recent "version-tagged" image, `quay.io/podman/stable:v3.1.2` is using F33 (and will never be updated).

I'm avoiding "fixing" this for now, because the automation is already very complicated and difficult to debug.  I'm also afraid of several important corner-cases that need handling.  Mainly, we would never get a "new" latest-tag, despite upstream Dockerfile changes or new OS security updates, unless/until a new podman version is released - which might be weeks.  

So, we agreed to merged this as-is (w/ daily updates/pushing of "latest") and I agreed to address the issue separately.

cevich: I opened a jira card to track the "fix this later" intention.

PR #10155, opened 2021-04-27
Use seccomp_profile as default profile if defined in containers.conf

pablofsf:
I bumped into the issue that pointing `seccomp_profile` in containers.conf to a user-defined profile completely ignored my settings. Instead, podman kept using the profile in `/usr/share/containers/seccomp.json`. This PR fixes that issue, by considering the runtime configuration of higher priority than the default settings.

### DIFF
--- a/libpod/define/info.go
+++ b/libpod/define/info.go
@@ -17,6 +17,7 @@ type SecurityInfo struct {
 	DefaultCapabilities string `json:"capabilities"`
 	Rootless            bool   `json:"rootless"`
 	SECCOMPEnabled      bool   `json:"seccompEnabled"`
+	SECCOMPProfilePath  string `json:"seccompProfilePath"`
 	SELinuxEnabled      bool   `json:"selinuxEnabled"`
 }
 
--- a/libpod/info.go
+++ b/libpod/info.go
@@ -87,6 +87,12 @@ func (r *Runtime) hostInfo() (*define.HostInfo, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "error getting hostname")
 	}
+
+	seccompProfilePath, err := DefaultSeccompPath()
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting Seccomp profile path")
+	}
+
 	info := define.HostInfo{
 		Arch:           runtime.GOARCH,
 		BuildahVersion: buildah.Version,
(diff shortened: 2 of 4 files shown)


pablofsf:
Checks are failing due to the lack of tests. Should I actually add tests for this small fix? I guess it could be nice to add one test to avoid future regressions, but not sure where exactly to do so.

rhatdan:
I would prefer this patch, and then add a test to make sure info shows the correct path.
[seccomp.patch.txt](https://github.com/containers/podman/files/6387320/seccomp.patch.txt)

pablofsf:
Thank you for your fast reply! That would make sense. I also took more time than necessary to figure out that podman was using the wrong path, which this should fix. I'll apply it and figure out the tests for it.

rhatdan:
@pablofsf Thanks.

pablofsf:
@rhatdan tests are fixed now. I skipped container and remote tests, as I could not see they would add any value.

rhatdan:
How about this for the test to test/e2e/containers_conf_test.go
```
	It("podman info seccomp profile path", func() {
		SkipIfInContainer("Storage needs to be configured")
		configPath := filepath.Join(podmanTest.TempDir, "containers.conf")
		os.Setenv("CONTAINERS_CONF", configPath)

		profile := filepath.Join(podmanTest.TempDir, "seccomp.json")
		containersConf := []byte(fmt.Sprintf("[containers]\nseccomp_profile=\"%s\"", profile))
		err = ioutil.WriteFile(configPath, containersConf, os.ModePerm)
		Expect(err).To(BeNil())

		if IsRemote() {
			podmanTest.RestartRemoteService()
		}

		session := podmanTest.Podman([]string{"info", "--format", "{{.Host.Security.SECCOMPProfilePath}}"})
		Expect(session.OutputToString()).To(Equal(profile))
	})
```

rhatdan:
/approve
LGTM, once the test is replaced.

pablofsf:
@rhatdan, it is added now, although I have removed the `SkipIfInContainer`, as I understand that using `podmanTest.Podman` then that is unnecessary.

Also, it has my `Signed-off by:`, but feels like it shouldn't as all the changes were actually implemented by you :)

rhatdan:
Don't worry, I am fine with you getting all the credit.  Thanks for bringing this to our attention.
LGTM

rhatdan:

PR #10165, opened 2021-04-28
system tests: networking: fix another race condition

edsantiago:
Our current #1 flake; this is a simple one: we can't run 'logs'
on a container run with '-d --rm'.


mheon:
LGTM

edsantiago:
/hold cancel
thanks

PR #10170, opened 2021-04-28
[NO TESTS NEEDED] Check if another VM is running on machine start

ashley-cui:
Only one VM can be up at a time. If another VM is running, or the current VM is running, error out on a podman machine start

[NO TESTS NEEDED]

### DIFF
--- a/cmd/podman/machine/start.go
+++ b/cmd/podman/machine/start.go
@@ -7,6 +7,7 @@ import (
 	"github.com/containers/podman/v3/pkg/domain/entities"
 	"github.com/containers/podman/v3/pkg/machine"
 	"github.com/containers/podman/v3/pkg/machine/qemu"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


rhatdan:
What happens if I have other VMs running on the machine for other purposes?

ashley-cui:
@rhatdan Shouldn't affect it. This only looks for VMs in `~/.config/containers/podman/machine`

rhatdan:
Excellent.

rhatdan:
LGTM, Except you need to put a comment into the `podman machine start` man page.

rhatdan:
Also if the VM exits for some other reason the podman machine stop, does the config get cleaned up?  Or does the qemu check if the container actually still exists?
Don't want a reboot to trigger, VM already running, when it is not.

ashley-cui:
@rhatdan It only checks the config to check where the VM's `QMPMonitor.Address`. It then pings the address to see if the VM in actually running or not. The config and address is persistent, but the ping will only succeed if the VM is running.

ashley-cui:
@baude PTAL

rhatdan:
LGTM

Issue #10173, opened 2021-04-29
`--cgroup-parent` has no effect


yaqwsx:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman creates container's cgroup under parent's cgroup instead of the one specified via `--cgroup-parent`.

**Steps to reproduce the issue:**

Follow this script:

```.sh
#!/usr/bin/env bash

set -e

# Create a systemd delegated scope
busctl call --user \
    org.freedesktop.systemd1 /org/freedesktop/systemd1 \
    org.freedesktop.systemd1.Manager StartTransientUnit \
    'ssa(sv)a(sa(sv))' TEST.scope \
    fail 4 PIDs au 1 $$ \
    Delegate b 1 MemoryAccounting b 1 CPUAccounting b 1 0
CGROUP=$(cat /proc/$$/cgroup | cut -d":" -f3)
echo "Base cgroup: $CGROUP"
FSBASE=/sys/fs/cgroup$CGROUP

# Create manager subgroup
mkdir $FSBASE/manager
echo $$ > $FSBASE/manager/cgroup.procs

# Enable controllers
echo "+cpu +memory" > $FSBASE/cgroup.subtree_control

# Create group for container
mkdir $FSBASE/container
echo "+cpu +memory" > $FSBASE/container/cgroup.subtree_control

# Create a container
CONNAME=$(podman container create \
            --cgroup-parent $CGROUP/container --cgroup-manager cgroupfs \
            ubuntu:20.04 /bin/bash -c 'echo XXX; sleep 5')
echo "Starting a container:"
podman container start $CONNAME

# Get the process from container
PID=$(ps ax | grep "echo XXX" | head -n1 | cut -d" " -f2)
echo "We get the cgroup of the parent process instead of the specified one via --cgroup-parent:"
cat /proc/$PID/cgroup
echo "But the configuration is correct:"
podman inspect $CONNAME | grep "CgroupParent"
```

**Describe the results you received:**

Podman creates container's cgroup under parent's cgroup instead of the one specified via `--cgroup-parent`.

**Describe the results you expected:**

Podman creates the container's cgroup under the cgroup specified via `---cgroup-parent`

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.15.2
Built:        Thu Jan  1 01:00:00 1970
[...]

PR #10177, opened 2021-04-29
cgroup: always honor --cgroup-parent

giuseppe:
if --cgroup-parent is specified, always honor it without doing any
detection whether cgroups are supported or not.

Closes: https://github.com/containers/podman/issues/10173

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -2224,8 +2224,19 @@ func (c *Container) getOCICgroupPath() (string, error) {
 	}
 	cgroupManager := c.CgroupManager()
 	switch {
-	case (rootless.IsRootless() && (cgroupManager == config.CgroupfsCgroupsManager || !unified)) || c.config.NoCgroups:
+	case c.config.NoCgroups:
 		return "", nil
+	case (rootless.IsRootless() && (cgroupManager == config.CgroupfsCgroupsManager || !unified)):
+		if c.config.CgroupParent == CgroupfsDefaultCgroupParent {
+			// old versions of podman were setting the CgroupParent to CgroupfsDefaultCgroupParent
+			// by default.  Avoid breaking these versions and check whether the cgroup parent is
+			// set to the default and in this case enable the old behavior.  It should not be a real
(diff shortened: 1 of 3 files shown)


libpod/container_internal_linux.go
```diff
@@ -2224,12 +2224,11 @@ func (c *Container) getOCICgroupPath() (string, error) {
 	}
 	cgroupManager := c.CgroupManager()
 	switch {
+	case c.config.CgroupParent != "":
```
mheon, line 2227: I think we should avoid this for the systemd cgroup manager

libpod/container_internal_linux.go
```diff
@@ -2225,7 +2225,7 @@ func (c *Container) getOCICgroupPath() (string, error) {
 	cgroupManager := c.CgroupManager()
 	switch {
 	case (rootless.IsRootless() && (cgroupManager == config.CgroupfsCgroupsManager || !unified)) || c.config.NoCgroups:
-		return "", nil
+		return c.config.CgroupParent, nil
```
mheon, line 2228: Do we risk breaking existing rootless containers that have CgroupParent set to the default?

giuseppe: good point.

Not sure what is the best way to deal with it.  Should we stop setting it if it is the default value (perhaps until podman 4.x)?

mheon: Is there any way we can do an access check on the cgroup, see if we have the ability to use it without error? If not, then your suggestion is definitely the best way forward.

giuseppe: I think that will be error prone and difficult to get right.  We won't only have to check if the current user has permissions to the cgroup, but that any ID mapped in the user namespace has it.  And that will be even more complicated with cgroup v1.

Looking more into it, I don't think it will be a problem if we just check for `c.config.CgroupParent == CgroupfsDefaultCgroupParent` as that cgroup is usually owned by root.

What do you think about the new version?

mheon: Alright, this SGTM

mheon:
LGTM

PR #10178, opened 2021-04-29
[v2.2.1-rhel] rmi: don't break when the image is missing a manifest

nalind:
From 3.1 and later: in `libpod/image/Image.Remove()`, don't go looking for the image's parent before actually deleting the image.  If the attempt to find the image's parent fails for any reason, we're basically stuck, when we could fail in a way that at least gets the caller closer to where they wanted to be.

From 3.1 and later: in `libpod/Runtime.RemoveImage()`, if we can't determine if an image has children because of some error, assume that it doesn't instead of bailing out.

From 3.1 and later: in pkg/domain/infra/abi.ImageEngine.Remove(), when attempting to remove all images, if we encounter an error checking if a given image has children, log a warning, and assume that it doesn't have any instead of failing, which would leave us unable to remove the image without resetting everything.

New for this branch: in `libpod/image/layerTree.toOCI()`, if we can't read the image's configuration in OCI format, log the error and return a `nil` pointer instead of returning the error.

Combined, these changes allow removal of an image that's lost its manifest for whatever reason, as we're seeing in https://bugzilla.redhat.com/show_bug.cgi?id=1884043.

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -623,7 +623,8 @@ func (i *Image) TopLayer() string {
 func (i *Image) Remove(ctx context.Context, force bool) error {
 	parent, err := i.GetParent(ctx)
 	if err != nil {
-		return err
+		logrus.Warnf("error determining parent of image: %v, ignoring the error", err)
+		parent = nil
 	}
 	if _, err := i.imageruntime.store.DeleteImage(i.ID(), true); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


edsantiago:
System tests pass on f33 (I ran them on my laptop, because this branch doesn't run any actual tests)

mheon:
We should let Jindrich know about this one, given it's only targeted for Openshift, not RHEL.

vrothberg (commented):
LGTM

PR #10185, opened 2021-05-01
Add filepath glob support to --security-opt unmask

rhatdan:
Want to allow users to specify --security-opt unmask=/proc/*.
This allows us to run podman within podman more securely, then
specifing umask=all, also gives the user more flexibilty.

### DIFF
--- a/contrib/podmanimage/stable/containers.conf
+++ b/contrib/podmanimage/stable/containers.conf
@@ -5,6 +5,7 @@ ipcns="host"
 utsns="host"
 cgroupns="host"
 cgroups="disabled"
+log_driver = "k8s_file"
 [engine]
 cgroup_manager = "cgroupfs"
 events_logger="file"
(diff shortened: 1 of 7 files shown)


TomSweeneyRedHat:
@rhatdan still fighting the CI test system

rhatdan:

ashley-cui (commented):
LGTM

contrib/podmanimage/stable/containers.conf
```diff
@@ -5,6 +5,7 @@ ipcns="host"
 utsns="host"
 cgroupns="host"
 cgroups="disabled"
+log_driver = "k8s_file"
```
vrothberg, line 8: Is this change intentional?

pkg/util/utils.go
```diff
@@ -76,6 +76,24 @@ func ParseRegistryCreds(creds string) (*types.DockerAuthConfig, error) {
 	}, nil
 }
 
+func ShouldMask(mask string, unmask []string) bool {
```
vrothberg, line 79: Since there's only one callers, can we move it into `pkg/specgen/generate`?

vrothberg: Code comments and unit tests would be nice.

contrib/podmanimage/stable/containers.conf
rhatdan: Yes it is a separate PR, Could add it as a fully separate from this, but I figured I would sneak it in while I was thinking about it.

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #10195, opened 2021-05-03
[CI:DOCS] Fix variable reference typo. in multi-arch image action

cevich:
Bug introduced by #10150

Also, in case of failure of one matrix-leg, do not terminate execution
of all others.  There are many reasons why an item could fail (i.e.
temporary networking problem).  Since the job runs periodically,
we can simply allow the subsequent run to cover for any missed images
pushes due to sporadic job failures.

Ref: https://github.com/containers/podman/runs/2490406959?check_suite_focus=true#step:12:27

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -24,6 +24,8 @@ jobs:
 
     # build several images (upstream, testing, stable) in parallel
     strategy:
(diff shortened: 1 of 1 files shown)


cevich:
Manually triggered workflow with this change: https://github.com/containers/podman/actions/runs/808065332

PR #10200, opened 2021-05-03
system test image: add arm64v8 image

edsantiago:
The RHEL multi-arch team informed me that we were missing
aarch64; add it, using the new name (arm64v8).

(This is from last week, so the image date tag does not
match today's date. I was waiting for confirmation that
things were working).


mheon:
LGTM

vrothberg (commented):
/lgtm

PR #10201, opened 2021-05-03
WIP: Cirrus: Update to XFS root in Fedora VM images

cevich:
***COMPLETELY experimental; DO NOT MERGE***

Ref: https://github.com/containers/automation_images/pull/69


cevich:
Failure Summary:

* [int podman fedora-34 root host](https://cirrus-ci.com/task/4648012543688704) - [Ginkgo node hung running checkpoint-pause test.](https://api.cirrus-ci.com/v1/artifact/task/4648012543688704/ginkgo_node_logs/test/e2e/ginkgo-node-1.log)
*

rhatdan:
Could you check to see if reflink is enabled?
```
xfs_info /var | grep ref
         =                       reflink=1    bigtime=0
```

cevich:
```
~/dev/podman (kickstart_fedora|✔) $ hack/get_ci_vm.sh int podman fedora-34 root host

# Initializing get_ci_vm
...cut...
# Accessing instance cevich-fedora-c6202812959817728
+ Entering Cirrus-CI environment
+ Changing into /var/tmp/go/src/github.com/containers/podman
+ Dropping into a bash login shell inside Cirrus-CI 'int podman fedora-34 root host' task environment
[root@cevich-fedora-c6202812959817728 podman]# xfs_info /var | grep ref
xfs_info: cannot open /var: Is a directory
[root@cevich-fedora-c6202812959817728 podman]# xfs_info / | grep ref
         =                       reflink=1    bigtime=0
[root@cevich-fedora-c6202812959817728 podman]#
```

rhatdan:
If /tmp is on the / OS then reflinks are enabled and look like they are enabled by default, which is goodness.

cevich:
> If /tmp is on the / OS then reflinks are enabled and look like they are enabled by default, which is goodness.

Yep, so either our bits aren't taking advantage as expected, or the affect isn't very substantial, or both.  Still, I'm a bit more concerned about the test failures I mentioned above.  They all strike me as things which shouldn't fail due to underlying FS changing.  Albeit, I also don't know the details, maybe it makes sense to someone?

rhatdan:
Yes I have no idea what is causing the timeouts.

cevich:
The network-create failures are even more strange: `Error: container create failed (no logs from conmon): EOF`

I wonder if maybe these are all race-conditions in the tests which are simply being exposed by timing changes.  If so, the failures are pretty consistent.  I got basically the same failures running these jobs multiple times and with BTRFS also :confused:

cevich:
Closing this for now, since there's a LOT of additional complexity in the image-build workflow for apparently minimal performance gain.  Happy to revisit/re-open this at a later date if there is a need.

PR #10219, opened 2021-05-05
add libimage events

vrothberg:
libimage now supports events which `libpod.Runtime` now uses for image
events.

Need to merge https://github.com/containers/common/pull/524 first.

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/containers/common/libimage"
 	"github.com/containers/common/pkg/config"
@@ -68,17 +69,18 @@ type Runtime struct {
 	storageConfig storage.StoreOptions
 	storageSet    storageSet
 
-	state             State
-	store             storage.Store
-	storageService    *storageService
-	imageContext      *types.SystemContext
-	defaultOCIRuntime OCIRuntime
-	ociRuntimes       map[string]OCIRuntime
-	runtimeFlags      []string
-	netPlugin         ocicni.CNIPlugin
-	conmonPath        string
-	libimageRuntime   *libimage.Runtime
-	lockManager       lock.Manager
+	state                  State
+	store                  storage.Store
+	storageService         *storageService
+	imageContext           *types.SystemContext
+	defaultOCIRuntime      OCIRuntime
+	ociRuntimes            map[string]OCIRuntime
+	runtimeFlags           []string
+	netPlugin              ocicni.CNIPlugin
+	conmonPath             string
+	libimageRuntime        *libimage.Runtime
+	libimageEventsShutdown chan bool
+	lockManager            lock.Manager
 
 	// doRenumber indicates that the runtime should perform a lock renumber
 	// during initialization.
@@ -214,6 +216,8 @@ func newRuntimeFromConfig(ctx context.Context, conf *config.Config, options ...R
 	if err := makeRuntime(ctx, runtime); err != nil {
 		return nil, err
 	}
+
+	runtime.libimageEventsShutdown = make(chan bool)
 
 	return runtime, nil
 }
@@ -680,6 +684,62 @@ func (r *Runtime) GetConfig() (*config.Config, error) {
 	return config, nil
 }
 
+// libimageEventsMap translates a libimage event type to a libpod event status.
+var libimageEventsMap = map[libimage.EventType]events.Status{
+	libimage.EventTypeImagePull:    events.Pull,
+	libimage.EventTypeImagePush:    events.Push,
+	libimage.EventTypeImageRemove:  events.Remove,
+	libimage.EventTypeImageLoad:    events.LoadFromArchive,
+	libimage.EventTypeImageSave:    events.Save,
+	libimage.EventTypeImageTag:     events.Tag,
+	libimage.EventTypeImageUntag:   events.Untag,
+	libimage.EventTypeImageMount:   events.Mount,
+	libimage.EventTypeImageUnmount: events.Unmount,
+}
+
+// libimageEvents spawns a goroutine in the background which is listenting for
+// events on the libimage.Runtime.  The gourtine will be cleaned up implicitly
+// when the main() exists.
+func (r *Runtime) libimageEvents() {
+	toLibpodEventStatus := func(e *libimage.Event) events.Status {
+		status, found := libimageEventsMap[e.Type]
+		if !found {
+			return "Unknown"
+		}
+		return status
+	}
+
+	go func() {
+		eventChannel := r.libimageRuntime.EventChannel()
+
+		for {
+			// Make sure to read and write all events before
+			// checking if we're about to shutdown.
+			for len(eventChannel) > 0 {
+				libimageEvent := <-eventChannel
+				e := events.Event{
+					ID:     libimageEvent.ID,
+					Name:   libimageEvent.Name,
+					Status: toLibpodEventStatus(libimageEvent),
+					Time:   libimageEvent.Time,
+					Type:   events.Image,
+				}
+				if err := r.eventer.Write(e); err != nil {
+					logrus.Errorf("unable to write image event: %q", err)
+				}
+			}
+
+			select {
+			case <-r.libimageEventsShutdown:
+				return
+
+			default:
+				time.Sleep(100 * time.Millisecond)
+			}
+		}
+	}()
+}
+
 // DeferredShutdown shuts down the runtime without exposing any
 // errors. This is only meant to be used when the runtime is being
 // shutdown within a defer statement; else use Shutdown
@@ -719,7 +779,11 @@ func (r *Runtime) Shutdown(force bool) error {
 	// If no store was requested, it can be nil and there is no need to
 	// attempt to shut it down
 	if r.store != nil {
-		if _, err := r.store.Shutdown(force); err != nil {
+		// Wait for the events to be written.
+		r.libimageEventsShutdown <- true
+
+		// Note that the libimage runtime shuts down the store.
+		if err := r.libimageRuntime.Shutdown(force); err != nil {
 			lastError = errors.Wrapf(err, "error shutting down container storage")
 		}
 	}
@@ -845,6 +909,8 @@ func (r *Runtime) configureStore() error {
 		return err
 	}
 	r.libimageRuntime = libimageRuntime
+	// Run the libimage events routine.
+	r.libimageEvents()
 
 	return nil
 }
--- a/test/system/090-events.bats
+++ b/test/system/090-events.bats
@@ -25,3 +25,39 @@ load helpers
     run_podman events --filter type=container --filter container=$cname --filter event=start --stream=false
     is "$output" "$expect" "filtering just by label"
 }
+
+@test "image events" {
+    skip_if_remote "FIXME: remove events on podman-remote seem to be broken"
+    pushedDir=$PODMAN_TMPDIR/dir
+    mkdir -p $pushedDir
+
+    tarball=$PODMAN_TMPDIR/ball.tar
+
+    run_podman image inspect --format "{{.ID}}" $IMAGE
+    imageID="$output"
+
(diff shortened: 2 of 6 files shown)


vrothberg:

@edsantiago PTAL at the system test.

test/system/090-events.bats
```diff
@@ -25,3 +25,35 @@ load helpers
...
+    skip_if_remote "FIXME: -remote is missing many events"
+    pushedDir=$PODMAN_TMPDIR/dir
+    mkdir -p $pushedDir
+
+    tarball=$PODMAN_TMPDIR/ball.tar
+
+    run_podman image inspect --format "{{.ID}}" $IMAGE
+    imageID="$output"
+
+    tag="registry.com/image:newtag"
```
edsantiago, line 40: I'd really like to see some use of `random_string` here (probably with `| tr A-Z a-z` to ensure lower-case. Some day I should fix random_string so it does that automatically). Reason: with a fixed name, we could be seeing results of a prior test run.

test/system/090-events.bats
```diff
@@ -25,3 +25,35 @@ load helpers
...
+    run_podman rmi $tag
+
+    run_podman events --stream=false --filter type=image
+    is "$output" ".*image push.* $imageID dir:$pushedDir.*"
+    is "$output" ".*image save.* $imageID $tarball.*"
+    is "$output" ".*image loadfromarchive.* $tarball.*"
+    is "$output" ".*image pull.* docker-archive:$tarball.*"
+    is "$output" ".*image tag.* $imageID $tag.*"
+    is "$output" ".*image untag.* $imageID $tag.*"
+    is "$output" ".*image remove.* $imageID $tag.*"
```
edsantiago, line 58: There's no order guarantee here, but I suppose that's not really worth complicating things for (as in, ensuring that the messages appear in the log in the same order as the actions).

vrothberg: Good point! I'll address it. The order should be checked.

test/system/090-events.bats
vrothberg: Will keep that in mind for the future. Thanks, Ed!

libpod/runtime.go
```diff
@@ -677,6 +677,53 @@ func (r *Runtime) GetConfig() (*config.Config, error) {
...
+		case libimage.EventTypeImageUntag:
+			return events.Untag
+		case libimage.EventTypeImageMount:
+			return events.Mount
+		case libimage.EventTypeImageUnmount:
+			return events.Unmount
+		default:
+			return "Unknown"
+		}
+	}
```
TomSweeneyRedHat, line 707: A couple of questions.  If this is going to hang around for a while and be used multiple times, might it be better to create a map and reference that?  If you decide to stick with the switch, are these in the order that you think most likely to be found/used?  If the ordering isn't a concern for perf reasons, then can you alphabetize them?

vrothberg: Great idea using a map!  I think alphabetic sorting makes sense if there's no relation among the items.  In that case, I prefer tag/untag or mount/unmout, for instance, to stick together.  But I don't feel strongly about it.

vrothberg:

test/system/090-events.bats
```diff
@@ -25,3 +25,35 @@ load helpers
...
+    run_podman rmi $tag
+
+    run_podman events --stream=false --filter type=image
+    is "$output" ".*image push.* $imageID dir:$pushedDir.*" \
+        ".*image save.* $imageID $tarball.*" \
+        ".*image loadfromarchive.* $tarball.*" \
+        ".*image pull.* docker-archive:$tarball.*" \
+        ".*image tag.* $imageID $tag.*" \
+        ".*image untag.* $imageID $tag.*" \
+        ".*image remove.* $imageID $tag.*"
```
edsantiago, line 58: This does not do what you think it does (and, sadly, I was not clever enough to add an error check to `is()` such that it would catch this). What I think you want is:
```bash
   is "$output" ".*image push $imageID dir:$pushedDir
.*image save $imageID $tarball
.*image loadfromarchive *$tarball
.*image pull *docker-archive:$tarball
.*image tag $imageID $tag
.*image untag $imageID $tag:latest
.*image tag $imageID $tag
.*image remove $imageID $tag.*" \
       "podman events"
```
(please copy-paste from the above, because I made a lot of changes including adding a missing event, removing a lot of unwanted `.*`s, and adding a missing `:latest`).

test/system/090-events.bats
```diff
@@ -25,3 +25,35 @@ load helpers
...
+    run_podman push $IMAGE dir:$pushedDir
+    run_podman save $IMAGE -o $tarball
+    run_podman load -i $tarball
+    run_podman pull docker-archive:$tarball
+    run_podman tag $IMAGE $tag
+    run_podman untag $IMAGE $tag
+    run_podman tag $IMAGE $tag
+    run_podman rmi $tag
+
+    run_podman events --stream=false --filter type=image
```
edsantiago, line 51: I would urge you please to add `--since`:
```bash
    t0=$(date --iso-8601=seconds)
    run_podman etc etc etc
    ...
    run_podman events --stream=false --filter type=image --since=$t0
```
Without that, the `events` command takes a long time on my laptop, which has only been up for a week or so; it would be pretty bad on a system with a multi-week uptime.

test/system/090-events.bats
```diff
@@ -25,3 +25,35 @@ load helpers
     run_podman events --filter type=container --filter container=$cname --filter event=start --stream=false
     is "$output" "$expect" "filtering just by label"
 }
+
+@test "image events" {
+    # `remove events` seem to be broken...
```
edsantiago, line 30: Suggestion: move the quote, to make this "# 'remove' events seem". That makes it a little less ambiguous: with the quotes as they are, and the `skip` immediately below, my first thought was "is that a typo for 'remote'?"

Even better: "'remove' events under podman-remote seem ..."

edsantiago (changes requested):
Suggestions for test; otherwise LGTM

test/system/090-events.bats
vrothberg: Thanks a lot, Ed!

test/system/090-events.bats
vrothberg: Lovely, thanks!

vrothberg:
Updated. Thanks, @edsantiago, for your guidance on the tests!

edsantiago (approved):

edsantiago:
Events seem to be out of order:
```
# #|     FAIL: podman events
# #| expected: '.*image push aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 dir:/tmp/podman_bats.kacprR/dir
# .*image save aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 /tmp/podman_bats.kacprR/ball.tar
# .*image loadfromarchive */tmp/podman_bats.kacprR/ball.tar
# .*image pull *docker-archive:/tmp/podman_bats.kacprR/ball.tar
# .*image tag aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb
# .*image untag aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb:latest
# .*image tag aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb
# .*image remove aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb.*'
# #|   actual: '2021-05-17 13:17:02.496094502 +0000 UTC image push aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 dir:/tmp/podman_bats.kacprR/dir'
# #|         > '2021-05-17 13:17:02.722939591 +0000 UTC image save aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 /tmp/podman_bats.kacprR/ball.tar'
# #|         > '2021-05-17 13:17:02.903253475 +0000 UTC image loadfromarchive  /tmp/podman_bats.kacprR/ball.tar'
# #|         > '2021-05-17 13:17:03.020686738 +0000 UTC image pull  docker-archive:/tmp/podman_bats.kacprR/ball.tar'
# #|         > '2021-05-17 13:17:03.19948181 +0000 UTC image untag aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb:latest'
# #|         > '2021-05-17 13:17:03.277832913 +0000 UTC image tag aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb'
# #|         > '2021-05-17 13:17:03.357413079 +0000 UTC image remove aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb:latest'
# #|         > '2021-05-17 13:17:03.357449542 +0000 UTC image untag aadc32e2a62610a251826bd9a5a0337bf84f62aeee982cb555ac0fa106c30aa2 registry.com/nwowcxnklb:latest'
```
Explanation: expected pull-tag-untag, got pull-untag-tag. Source: [ubuntu-2104](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5462055848247296/html/sys-podman-ubuntu-2104-root-host.log.html) but it looks like all system tests failed.

edsantiago:
The [fedora-34-root](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4899105894825984/html/sys-podman-fedora-34-root-host.log.html) failure is a missing 'remove' log. I think there's a problem with missing log entries.

vrothberg:
I have a suspicion: our old friend, the journal.  I am forcing to use the file backend for now to see if that resolves the issue.  If that does resolve the issue, I am going to open an issue so we get the journal backend fixed (similar to logs).

test/system/090-events.bats
```diff
@@ -25,3 +25,37 @@ load helpers
...
+    run_podman --events-backend=file push $IMAGE dir:$pushedDir
+    run_podman --events-backend=file save $IMAGE -o $tarball
+    run_podman --events-backend=file load -i $tarball
+    run_podman --events-backend=file pull docker-archive:$tarball
+    run_podman --events-backend=file tag $IMAGE $tag
+    run_podman --events-backend=file untag $IMAGE $tag
+    run_podman --events-backend=file tag $IMAGE $tag
+    run_podman --events-backend=file rmi $tag
+
+    run_podman events --stream=false --filter type=image --since $t0
```
edsantiago, line 51: I think you're going to need `--events-backend=file` here too

edsantiago:
Yuk. In [f34 root](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6664617599434752/html/sys-podman-fedora-34-root-host.log.html) there are **two** missing `tag` logs.

vrothberg:
Okay, it also fails with the "file" eventer (tag events not shown). Again it works locally.

This changes my _suspicion_ a bit: it could be a more general issue in the events rather than a bug in a specific driver. Maybe there is race condition when _reading_ the events in `podman events` such that we eat certain events?

edsantiago:
In other logs (sorry, I don't feel like linking) I'm seeing missing `remove`... which leads me to wonder if your `skip_if_remote` is truly needed, or if that was just an early indicator of the problem we're seeing now? Anyhow, it seems indisputable that there are missing log entries no matter which log driver is used.

vrothberg:
It seems like the problem is not the _reading_ but the _writing_ of events.  I debugged in a CI VM and noticed that some events are not showing up in the file.

vrothberg:
A first step is https://github.com/containers/common/pull/561 but it looks like there are still consistency issue on the Podman side.

edsantiago:
@vrothberg needs a whitespace fix

vrothberg:

PR #10220, opened 2021-05-05
podman: set volatile storage flag for --rm containers

giuseppe:
volatile containers are a storage optimization that disables *sync()
syscalls for the container rootfs.

If a container is created with --rm, then automatically set the
volatile storage flag as anyway the container won't persist after a
reboot or machine crash.

[NO TESTS NEEDED]

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -451,6 +451,8 @@ func (c *Container) setupStorage(ctx context.Context) error {
 		options.MountOpts = newOptions
 	}
 
+	options.Volatile = c.config.Volatile
+
 	c.setupStorageMapping(&options.IDMappingOptions, &c.config.IDMappings)
 
 	containerInfo, err := c.runtime.storageService.CreateContainerStorage(ctx, c.runtime.imageContext, c.config.RootfsImageName, c.config.RootfsImageID, c.config.Name, c.config.ID, options)
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -2318,3 +2318,16 @@ func WithPodSlirp4netns(networkOptions map[string][]string) PodCreateOption {
 		return nil
 	}
 }
+
+// WithVolatile sets the volatile flag for the container storage.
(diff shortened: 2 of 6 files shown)


libpod/container_internal.go
```diff
@@ -451,6 +451,10 @@ func (c *Container) setupStorage(ctx context.Context) error {
 		options.MountOpts = newOptions
 	}
 
+	if c.config.Volatile {
```
rhatdan, line 454: options.Volatile = c.config.Volatile

rhatdan (approved):

libpod/container_internal.go
giuseppe: fixed

rhatdan:
LGTM

libpod/options.go
```diff
@@ -2318,3 +2318,15 @@ func WithPodSlirp4netns(networkOptions map[string][]string) PodCreateOption {
 		return nil
 	}
 }
+
+// WithVolatile sets the volatile flag for the container storage.
```
mheon, line 2322: Can you add a warning here that this is potentially dangerous to use and can result in data loss?

mheon:
Missing tunnel implementation

pkg/domain/infra/abi/containers.go
```diff
@@ -834,6 +834,11 @@ func (ic *ContainerEngine) ContainerRun(ctx context.Context, opts entities.Conta
 	for _, w := range warn {
 		fmt.Fprintf(os.Stderr, "%s\n", w)
 	}
+
+	if opts.Rm {
```
mheon, line 838: Probably better to put this in cmd/podman/common/specgen.go - that way it's not just `podman run` for the ABI implementation, but run and create for both local and remote

giuseppe: thanks for the hint.  Fixed now

libpod/options.go
giuseppe: sure

Issue #10106, opened 2021-04-21
Support environment variable secrets


ashley-cui:
/kind feature

**Description**

We want to support environment variable secrets, where an environment variable would exist inside the container but not get committed inside an image.

mheon:
Is there a risk the current secrets are committed? I thought we had a guarantee they were not.

ashley-cui:
No, they're not committed, we just want to be able to do the same for some env variables

rhatdan:
Current secrets just support sticking a file into a container.  In this case we insert the secrets as environment variables, which do not get committed to the container. they are only given to the OCI Runtime.

Imagine running a container with a database in it that requires $PASSWORD to be set.

PR #10221, opened 2021-05-05
Add support for environment variable secrets

ashley-cui:
Env var secrets are env vars that are set inside the container but not
commited to and image. Also support reading from env var when creating a
secret.

Fixes: https://github.com/containers/podman/issues/10106

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -639,11 +639,15 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 		}
 		s.RestartPolicy = splitRestart[0]
 	}
+
+	s.Secrets, s.EnvSecrets, err = parseSecrets(c.Secrets)
+	if err != nil {
+		return err
+	}
 	s.Remove = c.Rm
 	s.StopTimeout = &c.StopTimeout
 	s.Timezone = c.Timezone
 	s.Umask = c.Umask
-	s.Secrets = c.Secrets
 	s.PidFile = c.PidFile
 
 	return nil
@@ -770,4 +774,71 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
 		td[split[0]] = specs.LinuxThrottleDevice{Rate: rate}
 	}
 	return td, nil
+}
+
+func parseSecrets(secrets []string) ([]string, map[string]string, error) {
+	secretParseError := errors.New("error parsing secret")
+	var mount []string
+	envs := make(map[string]string)
+	for _, val := range secrets {
+		source := ""
+		secretType := ""
+		target := ""
+		split := strings.Split(val, ",")
+
+		// --secret mysecret
+		if len(split) == 1 {
+			source = val
+			mount = append(mount, source)
+			continue
+		}
+		// --secret mysecret,opt=opt
+		if !strings.Contains(split[0], "=") {
+			source = split[0]
+			split = split[1:]
+		}
+		// TODO: implement other secret options
+		for _, val := range split {
+			kv := strings.SplitN(val, "=", 2)
+			if len(kv) < 2 {
+				return nil, nil, errors.Wrapf(secretParseError, "option %s must be in form option=value", val)
+			}
+			switch kv[0] {
+			case "source":
+				source = kv[1]
+			case "type":
+				if secretType != "" {
+					return nil, nil, errors.Wrap(secretParseError, "cannot set more tha one secret type")
+				}
+				if kv[1] != "mount" && kv[1] != "env" {
+					return nil, nil, errors.Wrapf(secretParseError, "type %s is invalid", kv[1])
+				}
+				secretType = kv[1]
+			case "target":
+				target = kv[1]
+			default:
+				return nil, nil, errors.Wrapf(secretParseError, "option %s invalid", val)
+			}
+		}
+
+		if secretType == "" {
+			secretType = "mount"
+		}
+		if source == "" {
+			return nil, nil, errors.Wrapf(secretParseError, "no source found %s", val)
+		}
+		if secretType == "mount" {
+			if target != "" {
+				return nil, nil, errors.Wrapf(secretParseError, "target option is invalid for mounted secrets")
+			}
+			mount = append(mount, source)
+		}
+		if secretType == "env" {
+			if target == "" {
+				target = source
+			}
+			envs[target] = source
+		}
+	}
+	return mount, envs, nil
 }
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -840,7 +840,7 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
-#### **\-\-secret**=*secret*
+#### **\-\-secret**=*secret*[,opt=opt ...]
 
 Give the container access to a secret. Can be specified multiple times.
 
@@ -848,12 +848,17 @@ A secret is a blob of sensitive data which a container needs at runtime but
 should not be stored in the image or in source control, such as usernames and passwords,
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
-secret value will still remain.
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an environment variable within the container.
+Secrets are written in the container at the time of container creation, and modifying the secret using `podman secret` commands
+after the container is created will not affect the secret inside the container.
 
-Secrets are managed using the `podman secret` command.
+Secrets and its storage are managed using the `podman secret` command.
+
+Secret Options
+
+- `type=mount|env`    : How the secret will be exposed to the container. Default mount.
+- `target=target`     : Target of secret. Defauts to secret name.
 
 #### **\-\-security-opt**=*option*
 
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -892,7 +892,7 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
-#### **\-\-secret**=*secret*
+#### **\-\-secret**=*secret*[,opt=opt ...]
 
 Give the container access to a secret. Can be specified multiple times.
 
@@ -900,12 +900,17 @@ A secret is a blob of sensitive data which a container needs at runtime but
 should not be stored in the image or in source control, such as usernames and passwords,
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
-secret value will still remain.
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an environment variable within the container.
+Secrets are written in the container at the time of container creation, and modifying the secret using `podman secret` commands
+after the container is created will not affect the secret inside the container.
 
-Secrets are managed using the `podman secret` command
+Secrets and its storage are managed using the `podman secret` command.
+
+Secret Options
+
+- `type=mount|env`    : How the secret will be exposed to the container. Default mount.
+- `target=target`     : Target of secret. Defauts to secret name.
 
 #### **\-\-security-opt**=*option*
 
--- a/cmd/podman/secrets/create.go
+++ b/cmd/podman/secrets/create.go
@@ -2,15 +2,16 @@ package secrets
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/containers/common/pkg/completion"
 	"github.com/containers/podman/v3/cmd/podman/common"
 	"github.com/containers/podman/v3/cmd/podman/registry"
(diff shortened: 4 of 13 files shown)


docs/source/markdown/podman-run.1.md
```diff
@@ -892,20 +892,26 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
+If a secret is deleted using `podman secret rm`, the container will still have access to the secret.
+If a secret is deleted and another secret is created with the same name, the secret inside the container will not change; the old
 secret value will still remain.
 
-Secrets are managed using the `podman secret` command
+Secrets are managed using the `podman secret` command.
+
+Secret Options
+
+- `type=mount|env`    : How the secret will be exposed to the container. Default mount.
```
mheon, line 913: It feels like this should be decided at secret creation time?

ashley-cui: I think it makes more sense for it to be an option in container run/create since mount/env doesn't change the substance of the secret, but only how the secret is used in relation to the container. More discussion here: https://github.com/containers/common/pull/516

rhatdan: @vrothberg and I had this discussion before and went back and forth on it, we came to this conclusion, to allow users to specify how they want to share the secrets.

cmd/podman/common/specgen.go
```diff
@@ -771,3 +775,63 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
...
+		split := strings.Split(val, ",")
+
+		// --secret mysecret
+		if len(split) == 1 {
+			source = val
+			mount = append(mount, source)
+			continue
+		}
+		// --secret mysecret,opt=opt
+		maybeSource := strings.SplitN(split[0], "=", 2)
```
rhatdan, line 797:
```
if strings.Index(split[0],"=") == -1 {
		source = split[0]
		split = split[1:]
}
```

cmd/podman/common/specgen.go
```diff
@@ -771,3 +775,63 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
...
+			split = split[1:]
+		}
+		// TODO: implement other secret options
+		for _, val := range split {
+			kv := strings.SplitN(val, "=", 2)
+			switch kv[0] {
+			case "source":
+				source = kv[1]
+			case "type":
+				if kv[1] == "mount" || kv[1] == "env" {
```
rhatdan, line 809: If it does not match one of these, you should throw an error.

Also if secretType != "" then you should throw an error.

cmd/podman/common/specgen.go
```diff
@@ -771,3 +775,63 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
...
+					secretType = kv[1]
+				}
+			case "target":
+				// Target is an environment variable only option
+				envOnly = true
```
rhatdan, line 814: Why do I need envOnly, can I just check if target!=""?

cmd/podman/common/specgen.go
```diff
@@ -771,3 +775,63 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
...
+			default:
+				return nil, nil, errors.Wrapf(secretParseError, "option %s invalid", val)
+			}
+		}
+		if source == "" {
+			return nil, nil, errors.Wrapf(secretParseError, "no source found %s", val)
+		}
+		if secretType == "mount" {
+			if envOnly {
+				return nil, nil, errors.Wrapf(secretParseError, "invalid mount secret option")
```
rhatdan, line 825: Tell me I can not have a target with a mount type.

cmd/podman/common/specgen.go
rhatdan: Why not support target=/PATH/TO/SECRET for mount types?

I don't see any target tests below.  I also don't see any failure tests.

rhatdan (approved):

cmd/podman/common/specgen.go
ashley-cui: I was going to support that in another PR, along with UID, GID, and Mode for mounted secrets, since it touches on a lot of other stuff in the mount secret backend

ashley-cui: I'll add some failure tests though

cmd/podman/common/specgen.go
```diff
@@ -771,3 +775,67 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
...
+		if !strings.Contains(split[0], "=") {
+			source = split[0]
+			split = split[1:]
+		}
+		// TODO: implement other secret options
+		for _, val := range split {
+			kv := strings.SplitN(val, "=", 2)
+			switch kv[0] {
+			case "source":
+				source = kv[1]
```
TomSweeneyRedHat, line 805: I think there's a slight chance you could not have a second entry in the slice, perhaps:
```suggestion
				source = kv[1:]
```
to be ultrasafe?

TomSweeneyRedHat: ditto below

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,27 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
 
 Note that this feature is experimental and may change in the future.
 
-#### **\-\-secret**=*secret*
+#### **\-\-secret**=*secret*[**,**opt=opt]
```
TomSweeneyRedHat, line 843: I'm not sure what you're trying to show here, but it comes out as the below, but all bolded:
```
--secret=secret[**,**opt=opt]
```

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,27 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
 
 A secret is a blob of sensitive data which a container needs at runtime but
 should not be stored in the image or in source control, such as usernames and passwords,
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an enviornment variable.
```
TomSweeneyRedHat, line 852:
```suggestion
When secrets are specified as type `env`, the secret will be set as an environment variable within the container.
```

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,27 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
+If a secret is deleted using `podman secret rm`, the container will still have access to the secret.
+If a secret is deleted and another secret is created with the same name, the secret inside the container will not change; the old
 secret value will still remain.
```
TomSweeneyRedHat, line 855:
```suggestion
secret value will still remain within the container.
```

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,27 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
 should not be stored in the image or in source control, such as usernames and passwords,
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an enviornment variable.
+If a secret is deleted using `podman secret rm`, the container will still have access to the secret.
+If a secret is deleted and another secret is created with the same name, the secret inside the container will not change; the old
```
TomSweeneyRedHat, line 854: Probably my misunderstanding, but where is the secret that is being deleted if it is not within the container?  I think more clarification is needed here as by reading this, it's not clear to me.  (Although it might be end of day fried brain here too.)

docs/source/markdown/podman-run.1.md
```diff
@@ -892,20 +892,26 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
+- `target=target`     : Target of secret. Defauts to secret name.
```
TomSweeneyRedHat, line 914: if you use any comments from above, please apply here too.

docs/source/markdown/podman-create.1.md
ashley-cui: It's being deleted from the secrets store, so, `podman secret create` creates a secret that is available for a container to use, and `podman secret rm` deletes it. This doesn't affect a secret inside a container if the container was created with that secret.

TomSweeneyRedHat: A "for example" sentence or two explaining how a tls cert might be deleted, and then recreated on the host.  But does podman secret rm really delete the cert?

TomSweeneyRedHat: Ah, the secret store is what I was missing and there's no mention of it in the man page here.  A quick sentence or two about that being the place secrets are stored created to and deleted from on the host would probably be good.

ashley-cui: I tried to clarify it a little in my latest push, let me know if its clearer

cmd/podman/common/specgen.go
ashley-cui: I think this is okay since strings.split takes 2 as an arg so it splits it into 2?

TomSweeneyRedHat: Doh, yep, it's fine.  I completely messed the kv < 2 len check.

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,28 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
-secret value will still remain.
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an environment variable within the container.
+If a secret is deleted using `podman secret rm`, the secret will be deleted from the secret store,
+but secret data will sill be availible inside the container. `
```
TomSweeneyRedHat, line 854:
```suggestion
but secret data will still be available inside the container.
```

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,28 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
 Note that this feature is experimental and may change in the future.
 
-#### **\-\-secret**=*secret*
+#### **\-\-secret**=*secret*[,opt=opt ...]
 
 Give the container access to a secret. Can be specified multiple times.
 
 A secret is a blob of sensitive data which a container needs at runtime but
 should not be stored in the image or in source control, such as usernames and passwords,
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
```
TomSweeneyRedHat, line 849: You now talk about the secret store below, but it was never intro'd elsewhere.  IDK if this in spot on or not, but could you add something like at the end of this paragraph?

When a secret is created, it is placed in the secret store under `/run/secrets` for later reference.

Feel free to munge that further if necessary.

docs/source/markdown/podman-create.1.md
```diff
@@ -840,21 +840,28 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
-secret value will still remain.
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an environment variable within the container.
+If a secret is deleted using `podman secret rm`, the secret will be deleted from the secret store,
+but secret data will sill be availible inside the container. `
+If a secret is deleted and another secret is created with the same name, the secret data inside the container will not change;
```
TomSweeneyRedHat, line 855:
```suggestion
If a secret is deleted and another secret is created with the same name in the secret store, the secret data inside the container will not change;
```

docs/source/markdown/podman-create.1.md
ashley-cui: Hmm, I think its best not to get too much into how the secret is stored, since the storage isn't in the container or `/run/secrets`. This is the doc for the --secret flag for a secret to be used in a container, and not for how secrets are managed. 
I think the thing that's valuable here in the docs is that secrets are written in the container at the time of container creation, and modifying the secret using podman secret commands after the container is created with the secret will not affect the secret inside the container. I've made edits, let me know if this is clearer.

docs/source/markdown/podman-create.1.md
```diff
@@ -840,20 +840,25 @@ Specify the policy to select the seccomp profile. If set to *image*, Podman will
...
 TLS certificates and keys, SSH keys or other important generic strings or binary content (up to 500 kb in size).
 
-Secrets are copied and mounted into the container when a container is created. If a secret is deleted using
-`podman secret rm`, the container will still have access to the secret. If a secret is deleted and
-another secret is created with the same name, the secret inside the container will not change; the old
-secret value will still remain.
+When secrets are specified as type `mount`, the secrets are copied and mounted into the container when a container is created.
+When secrets are specified as type `env`, the secret will be set as an environment variable within the container.
+Secrets are written in the container at the time of container creation, and modifying the secret using `podman secret` commands
+after the container is created will not affect the secret inside the container.
```
TomSweeneyRedHat, line 854: :star: to @ashley-cui I like  how you've worked this up, tyvm and well done.

TomSweeneyRedHat:
LGTM
assuming happy tests

rhatdan:
Does this handle

podman run --secret mysecret,type=mount or --secret mysecret,type=env?

ashley-cui:
> Does this handle
> 
> podman run --secret mysecret,type=mount or --secret mysecret,type=env?

Both

PR #10222, opened 2021-05-05
podman image tree: restore previous behavior

vrothberg:
The initial version of libimage changed the order of layers which has      
now been restored to remain backwards compatible.                          
                                                                           
Further changes:                                                           
                                                                           
 * Fix a bug in the journald logging which requires to strip trailing      
   new lines from the message.  The system tests did not pass due to       
   empty new lines.  Triggered by changing the default logger to           
   journald in containers/common.                                          
                                                                           
 * Fix another bug in the journald logging which embedded the container    
   ID inside the message rather than the specifid field.  That surfaced    
   in a preceeding whitespace of each log line which broke the system      
   tests.                                                                  
                                                                           
 * Alter the system tests to make sure that the k8s-file and the           
   journald logging drivers are executed.                                  
                                                                           
 * A number of e2e tests have been changed to force the k8s-file driver    
   to make them pass when running inside a root container.                 
                                                                           
 * Increase the timeout in a kill test which seems to take longer now.     
   Reasons are unknown.  Tests passed earlier and no signal-related        
   changes happend.  It may be CI VM flake since some system tests but     
   other flaked.                                                           
                                                                           
### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/buildah v1.20.2-0.20210504130217-903dc56408ac
-	github.com/containers/common v0.37.2-0.20210503193405-42134aa138ce
+	github.com/containers/common v0.38.1-0.20210510140555-24645399a050
 	github.com/containers/conmon v2.0.20+incompatible
 	github.com/containers/image/v5 v5.12.0
 	github.com/containers/ocicrypt v1.1.1
--- a/go.sum
+++ b/go.sum
@@ -197,8 +197,9 @@ github.com/containernetworking/plugins v0.9.1 h1:FD1tADPls2EEi3flPc2OegIY1M9pUa9
 github.com/containernetworking/plugins v0.9.1/go.mod h1:xP/idU2ldlzN6m4p5LmGiwRDjeJr6FLK6vuiUwoH7P8=
 github.com/containers/buildah v1.20.2-0.20210504130217-903dc56408ac h1:rPQTF+1lz+F4uTZgfk2pwqGcEEg9mPSWK58UncsqsrA=
 github.com/containers/buildah v1.20.2-0.20210504130217-903dc56408ac/go.mod h1:0hqcxPCNk/lit/SwBQoXXymCbp2LUa07U0cwrn/T1c0=
-github.com/containers/common v0.37.2-0.20210503193405-42134aa138ce h1:e7VNmGqwfUQkw+D5bms262x1HYqxfN9/+t5SoaFnwTk=
 github.com/containers/common v0.37.2-0.20210503193405-42134aa138ce/go.mod h1:JjU+yvzIGyx8ZsY8nyf7snzs4VSNh1eIaYsqoSKBoRw=
+github.com/containers/common v0.38.1-0.20210510140555-24645399a050 h1:o3UdnXpzCmJwto4y+addBH2NXZObuZ0tXA7COZXNMnQ=
+github.com/containers/common v0.38.1-0.20210510140555-24645399a050/go.mod h1:64dWQkAgrd2cxVh9eDOxJ/IgPOulVFg6G7WLRDTrAuA=
 github.com/containers/conmon v2.0.20+incompatible h1:YbCVSFSCqFjjVwHTPINGdMX1F6JXHGTUje2ZYobNrkg=
 github.com/containers/conmon v2.0.20+incompatible/go.mod h1:hgwZ2mtuDrppv78a/cOBNiCm6O0UMWGx1mu7P00nu5I=
 github.com/containers/image/v5 v5.11.1/go.mod h1:HC9lhJ/Nz5v3w/5Co7H431kLlgzlVlOC+auD/er3OqE=
--- a/libpod/container_log_linux.go
+++ b/libpod/container_log_linux.go
@@ -8,12 +8,12 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"strings"
 	"time"
 
 	"github.com/containers/podman/v3/libpod/define"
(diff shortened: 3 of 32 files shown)


vrothberg:

@edsantiago this restores the previous order of layers.

rhatdan:
LGTM

vrothberg:
Looks like the journald change broke e2e tests, having a look now.

vrothberg:
@mheon @rhatdan, I am running out of time today and am on PTO tomorrow.  The errors look related to your changes in c/common.  Feel free to take over this PR (the image-tree related parts are done and green).

vrothberg:
Currently failing. Requires https://github.com/containers/common/pull/540.

vrothberg:
This turned out to be more work.  Fixed bugs in the journald driver and updates tests.

@edsantiago, please take a look at the changes.

vrothberg:

edsantiago:
I'm really kind of neurotic about duplication. Would you consider a friendly amendment?

[035.diff.txt](https://github.com/containers/podman/files/6466192/035.diff.txt)

vrothberg:
> I'm really kind of neurotic about duplication. Would you consider a friendly amendment?

Absolutely. It looks great, thanks a lot!

edsantiago:
```
# $ podman exec 97109c3ce4e6efe077b26124eb217d18708e15147841f9b054c7b66edbedd3b6 touch /stop
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #| FAIL: Timed out waiting for DONE from container
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This is really bad. The fact that this PR is mucking with logging makes me suspect that there's something broken in logging.

edsantiago:
And no, bumping the timeout to 15 seconds is not a solution

vrothberg:
> ```
> # $ podman exec 97109c3ce4e6efe077b26124eb217d18708e15147841f9b054c7b66edbedd3b6 touch /stop
> # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
> # #| FAIL: Timed out waiting for DONE from container
> # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> This is really bad. The fact that this PR is mucking with logging makes me suspect that there's something broken in logging.

Still trying to get a feeling for it.  Increasing the timeout worked for most cases but not rootless.  I don't yet see a relation to logging in this specific case of exec.

vrothberg:
> And no, bumping the timeout to 15 seconds is not a solution

I'd appreciate help.  I wanted to fix image-tree and ended up fixing things that are very unrelated but merged untested in c/common :see_no_evil:

edsantiago:
Something is broken in logging. The timeout has nothing to do with anything; this should trigger in less than a second. If it doesn't, you can bump the timeout to an hour, it will not change anything, the test will still fail.

Maybe it's the change to journald logging? Maybe journald logging doesn't work? Maybe you can try:
```diff
--- a/test/system/130-kill.bats
+++ b/test/system/130-kill.bats
@@ -8,7 +8,7 @@ load helpers
 @test "podman kill - test signal handling in containers" {
     # Start a container that will handle all signals by emitting 'got: N'
     local -a signals=(1 2 3 4 5 6 8 10 12 13 14 15 16 20 21 22 23 24 25 26 64)
-    run_podman run -d $IMAGE sh -c \
+    run_podman run --log-driver=k8s-file -d $IMAGE sh -c \
         "for i in ${signals[*]}; do trap \"echo got: \$i\" \$i; done;
         echo READY;
         while ! test -e /stop; do sleep 0.05; done;
```
If that fixes it, then the problem is with the journald driver, and I kind of think it is super important to find and fix that bug.

vrothberg:
Thanks, Ed!

> Maybe you can try:

Yes.  I did that in the last push.

> If that fixes it, then the problem is with the journald driver, and I kind of think it is super important to find and fix that bug.

Agreed.  I find it curious thought that it started flaking today.  The only thing that changed since yesterday is the diff you proposed.

edsantiago:
Basically, what seems to be happening is that `podman logs -f CONTAINER` is not spitting out the final output line from a container. 

UPDATE: I have a reproducer.
```console
$ while :;do ./bin/podman run --log-driver=journald -d --name foo quay.io/libpod/testimage:20210427 sh -c 'echo hi;sleep 2;echo bye';./bin/podman logs -f foo;./bin/podman rm foo;done
caf4dce7ce3ac5d08da339a8c0ecd4ec97df498cd472fbff61182352e71cf831
hi     <---- there is no subsequent bye!
caf4dce7ce3ac5d08da339a8c0ecd4ec97df498cd472fbff61182352e71cf831
08f58203f738027300d651fed861f1673eb8a1a953b869c788d905664b709938
hi
bye
```

vrothberg:
Thank you so much, @edsantiago!  That gives me something to chew on.

vrothberg:
@edsantiago, I had a look and analyzed it.  The race condition is nasty and not easy to fix.  I did what I could in the given time but opened https://github.com/containers/podman/issues/10323 to track it and dropped a FIXME in the code (see https://github.com/containers/podman/pull/10222/files#diff-20cc30e1cdf302ef7404e5923eada3912c68c8b8943c0a7a0a834b29236eba69R92).

FWIW, I opened https://github.com/containers/common/pull/546 earlier today to revert the journald defaulting in c/common.

vrothberg:
@rhatdan FYI

edsantiago:
Reluctant LGTM. This journald thing is making me reeeeeeeeally uncomfortable.

vrothberg:
Thanks! Yes, I am uncomfortable as well. It’ll get reverted in common soon.

edsantiago:
Tests green. @containers/podman-maintainers PTAL

PR #10223, opened 2021-05-05
rootless: improve automatic range split

giuseppe:
sort.Search returns the smallest index, so provide the available IDs
in decreasing order.

It fixes an issue when splitting the current mappings over multiple
available IDs.


mheon:
LGTM

PR #10230, opened 2021-05-05
fix: docker APIv2 `images/get`

### LINKED ISSUE #10229: Docker APIv2 `images/get` endpoint not working.
To reproduce:
```
curl "http://localhost:1234/v1.38/images/get?names=quay.io/mvasek/faas-quarkus-jvm-bp:v999" -v       
*   Trying ::1:1234...
* Connected to localhost (::1) port 1234 (#0)
> GET /v1.38/images/get?names=quay.io/mvasek/faas-quarkus-jvm-bp:v999 HTTP/1.1
> Host: localhost:1234
> User-Agent: curl/7.76.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< Api-Version: 1.40
< Content-Type: application/json
< Libpod-Api-Version: 3.2.0-dev
< Server: Libpod/3.2.0-dev (linux)
< Date: Wed, 05 May 2021 18:47:59 GMT
< Content-Length: 121
< 
{"cause":"unspported format \"\" for saving images","message":"unspported format \"\" for saving images","response":500}
* Connection #0 to host localhost left intact
```

This has been introduced in `migrate Podman to containers/common/libimage` a278195.

matejvasek:
Fixing docker APIv2 `images/get/` endpoint.
Image format must be set to "docker-archive" not empty string.

resolves #10229

### DIFF
--- a/test/apiv2/10-images.at
+++ b/test/apiv2/10-images.at
@@ -151,5 +151,6 @@ t DELETE libpod/images/test1:latest 200
 #      actual: {"cause":"no such image","message":"unable to find a name and tag match for busybox in repotags: no such image","response":500}
 #
 #t GET images/get?names=alpine,busybox 200 '[POSIX tar archive]'
+t GET "images/get?names=alpine" 200 '[POSIX tar archive]'
 
 # vim: filetype=sh
--- a/test/apiv2/10-images.at
+++ b/test/apiv2/10-images.at
@@ -148,6 +148,7 @@ podman pull busybox
 #is "$img_cnt" 2 "number of images in tar archive"
 # TODO getting multiple images is not supported yet
 # once it is supported replace the test below by the tests above
-t GET "images/get?names=alpine&names=busybox" 501
+t GET "images/get?names=alpine&names=busybox" 501 \
+  .cause="getting multiple image is not supported yet"
 
 # vim: filetype=sh
--- a/test/apiv2/10-images.at
+++ b/test/apiv2/10-images.at
@@ -140,17 +140,14 @@ t GET images/json?filters='{"reference":["test1"]}' 200 length=1
 
 t DELETE libpod/images/test1:latest 200
 
-# Export more than one image
-# FIXME FIXME FIXME, this doesn't work:
-#   not ok 64 [10-images] GET images/get?names=alpine,busybox : status
-#    expected: 200
-#      actual: 500
-#    expected: 200
-#  not ok 65 [10-images] GET images/get?names=alpine,busybox : output
-  #  expected: [POSIX tar archive]
-#      actual: {"cause":"no such image","message":"unable to find a name and tag match for busybox in repotags: no such image","response":500}
-#
-#t GET images/get?names=alpine,busybox 200 '[POSIX tar archive]'
 t GET "images/get?names=alpine" 200 '[POSIX tar archive]'
+
+podman pull busybox
+#t GET "images/get?names=alpine&names=busybox" 200 '[POSIX tar archive]'
+#img_cnt=$(tar xf "$WORKDIR/curl.result.out" manifest.json -O | jq "length")
+#is "$img_cnt" 2 "number of images in tar archive"
+# TODO getting multiple images is not supported yet
+# once it is supported replace the test below by the tests above
+t GET "images/get?names=alpine&names=busybox" 501
 
 # vim: filetype=sh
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@vrothberg PTAL

test/apiv2/10-images.at
```diff
@@ -151,5 +151,6 @@ t DELETE libpod/images/test1:latest 200
 #      actual: {"cause":"no such image","message":"unable to find a name and tag match for busybox in repotags: no such image","response":500}
 #
 #t GET images/get?names=alpine,busybox 200 '[POSIX tar archive]'
+t GET "images/get?names=alpine" 200 '[POSIX tar archive]'
```
edsantiago, line 154: Why not just uncomment line 153, and get rid of the huge FIXME which is no longer applicable?

edsantiago (changes requested):
Whether you go with `alpine,busybox` or just `alpine`, please remove the enormous FIXME comment block - it is now misleading, since you've fixed the bug. Thank you!

TomSweeneyRedHat:
For future reference, for the sake of the person creating the release notes, it's really nice to have a small description, even if it's mostly cut/pasted from the issue, in the description field rather than just the link to the issue.

LGTM once @edsantiago 's comment is addressed.

test/apiv2/10-images.at
matejvasek: @edsantiago because I know that if there is only one name it works correctly. If multiple names are provided only first one is taken account.

matejvasek: `t GET images/get?names=alpine,busybox 200 '[POSIX tar archive]'` would pass, we should have some better assertion than just content-type.

matejvasek: I can change the `TODO`, it should say that this function for now works only with single parameter.

matejvasek: Is there simple way to assert content of non-JSON body?

matejvasek: hmm also seems that `dockerd` accepts querystring like `?names=alpine&names=busybox`, not comma separated

matejvasek: but the documenation says it's comma separated array of string, https://docs.docker.com/engine/api/v1.41/#operation/ImageGet, weird

edsantiago: Aaaaaargh! Too many comments to reply to! :smile:

The results are saved in `$WORKDIR/curl.result.out`. You can see `test/apiv2/44-mounts.at` for an example of reading that. Maybe you could run `tar tf` and then do a `like` on that.

I don't know anything about what this endpoint is supposed to do. If the two-argument form doesn't work, please update the comments to say so.

My only request is, please don't leave a confusing obsolete FIXME comment in there. Thank you.

edsantiago: Here's a good test:
```bash
t GET images/get?names=$IMAGE 200 '[POSIX tar archive]'
tar_tf=$(tar tf $WORKDIR/curl.result.out)
like "$tar_tf" ".*$iid\.json" "IID of test image found in tarball"
```
(`$iid` is defined way at top). Customizing that to deal with multi-arg form is left as an exercise for the reader.

matejvasek:
@edsantiago PTAL

test/apiv2/10-images.at
```diff
@@ -140,16 +140,14 @@ t GET images/json?filters='{"reference":["test1"]}' 200 length=1
...
-#t GET images/get?names=alpine,busybox 200 '[POSIX tar archive]'
+t GET "images/get?names=alpine" 200 '[POSIX tar archive]'
+
+podman pull busybox
+#t GET "images/get?names=alpine&names=busybox" 200 '[POSIX tar archive]'
+#img_cnt=$(tar xf "$WORKDIR/curl.result.out" manifest.json -O | jq "length")
+#is "$img_cnt" 2 "number of images in tar archive"
+# TODO getting multiple images is not supported yet
+# once it is supported replace the test below by the tests above
+t GET "images/get?names=alpine&names=busybox" 501
```
edsantiago, line 151: It's a good practice to check error messages, not just codes. Suggestion:
```diff
 t GET "images/get?names=alpine&names=busybox" 501 \      <-- add backslash, plus next line
+  .cause="getting multiple image is not supported yet"
```

edsantiago (approved):
LGTM; I will leave approval in the hands of those more familiar with `images.go`.

One suggestion, not worth re-pushing just for that though.

matejvasek:
@edsantiago I found out how to fix the multi image download, but maybe I should do it in separate PR

edsantiago:
I am totally a fan of small, clean, reviewable PRs!

vrothberg:
Changes LGTM but a friendly reminder to keep the commit message descriptive.

PR #10234, opened 2021-05-05
cgroup: fix rootless --cgroup-parent with pods

giuseppe:
extend to pods the existing check whether the cgroup is usable when
running as rootless with cgroupfs.

commit 17ce567c6827abdcd517699bc07e82ccf48f7619 introduced the
regression.

[NO TESTS NEEDED] -- we already have tests, they're just not being run. #10237 hopes to fix that.

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -2216,6 +2216,17 @@ func (c *Container) generatePasswdAndGroup() (string, string, error) {
 	return passwdPath, groupPath, nil
 }
 
+func isRootlessCgroupSet(cgroup string) bool {
+	// old versions of podman were setting the CgroupParent to CgroupfsDefaultCgroupParent
+	// by default.  Avoid breaking these versions and check whether the cgroup parent is
+	// set to the default and in this case enable the old behavior.  It should not be a real
+	// problem because the default CgroupParent is usually owned by root so rootless users
+	// cannot access it.
+	// This check might be lifted in a future version of Podman.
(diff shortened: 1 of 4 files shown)


edsantiago:
Needs a test, please. Unfortunately, IIUC, this only happens on rootless + cgroups v1 + runc? If that's true, then we're in trouble because we don't test that in CI.

@containers/podman-maintainers can we add Prior-ubuntu + rootless to our CI matrix? I do not ask this lightly: I ask because that's the only way we can catch RHEL problems early.

rhatdan:
SGTM

edsantiago:
@giuseppe please re-push (maybe with a whitespace-only change to the commit message). I've edited the github description to include the magic [NO TESTS NEEDED] string, and as soon as this merges I will follow up with #10237 to make sure this doesn't happen again.

giuseppe:
> Needs a test, please. Unfortunately, IIUC, this only happens on rootless + cgroups v1 + runc? If that's true, then we're in trouble because we don't test that in CI.

in this case I like more the runc behavior.  If the path is explicitly set in the configuration file (as Podman did before this PR), then the error should not be ignored even in rootless mode on cgroup v1.  I've opened a PR for crun to fix it: https://github.com/containers/crun/pull/658

edsantiago:
LGTM. Tests are passing in #10237, which I rebased onto this PR (they failed catastrophically before I rebased). @containers/podman-maintainers PTAL and let's get this in ASAP.

PR #10236, opened 2021-05-05
Add envars to the generated systemd unit

### LINKED ISSUE #10101: generated systemd units lack values of environment variables referenced
/kind bug

**Description**

When generating systemd units using `--new`, the value of environment variables which were only referenced in the original `create` command is lost.

```shell
$ export FOO=bar
$ podman create --name=test --env FOO alpine
$ podman inspect test | grep bar
$ podman generate systemd -n --new test | grep bar
```

"bar" is found for the `inspect` command, but not found in the systemd unit.

I expected the value of `FOO` being saved somewhere, either in the systemd unit itself, or a in separate file referenced by the systemd unit.

```
$ podman version 
Version:      3.1.0-dev
API Version:  3.1.0-dev
Go Version:   go1.16.1
Built:        Fri Mar 26 19:32:03 2021
OS/Arch:      linux/amd64
```

```
podman-3.1.0-0.13.module_el8.5.0+733+9bb5dffa.x86_64
```

Luap99:
The with --new generated systemd unit loses the environment variables
when the create command only contains the key without the value. Since
podman tries to lookup those values from the environment the unit can
fail.

This commits ensures that we will add the environment variables to the
unit file when this is the case. The container environment variables are
looked up in the container spec.

Fixes #10101


TomSweeneyRedHat:
Changes LGTM, but want someone with more systems knowledge to look (@vrothberg?).  The fedora-34 root test failed, looked to be a network flake.  I've restarted.

eriksjolund:
Quote from `man sysemd.exec`:  _Specifier expansion is  performed_

I wonder if the character `%` needs to be escaped. Maybe it's already done?


Here is a an example of specifier expansion:



```
[esjolund@laptop user]$ cat test.service 
[Unit]
Description=Demonstrating specifier expansion

[Service]
Type=oneshot
Environment="FOOBAR=%s"
ExecStart=/bin/env
RemainAfterExit=true
ExecStop=/bin/true
StandardOutput=file:/home/esjolund/stdoutfile
[Install]
WantedBy=multi-user.target

[esjolund@laptop user]$ systemctl --user daemon-reload 
[esjolund@laptop user]$ systemctl --user stop test.service 
[esjolund@laptop user]$ rm -f ~/stdoutfile  
[esjolund@laptop user]$ systemctl --user start test.service
[esjolund@laptop user]$ cat ~/stdoutfile | grep FOOBAR
FOOBAR=/bin/bash
[esjolund@laptop user]$ 
```

`FOOBAR=%` was replaced by `FOOBAR=/bin/bash`

Luap99:
@eriksjolund You are right I need to escape `%`, thanks for the hint.

mheon:
LGTM

Issue #9997, opened 2021-04-12
Ubuntu rootless podman cp / fails to stat cgroups file


cevich:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

PR #9905 enables rootless testing under cgroupsV2 enabled Ubuntu 2010 (with `crun`).  During one particular CI run, [this failure mode was observed](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5313776456040448/html/int-podman-ubuntu-2010-rootless-host.log.html#t--podman-cp-the-root-directory-from-the-ctr-to-an-existing-directory-on-the-host--1).  It seems to me, that the behavior or `cp` may be incorrect: Is it ***ever*** reasonable to copy a cgroups virtual file from inside a container to the host?   

**Steps to reproduce the issue:**

1. Execute rootless CI testing on Ubuntu 2010

2. Repeat (possibly a billion times) until this failure occurs :cry: 

**Describe the results you received:**

```
           podman cp the root directory from the ctr to an existing directory on the host 
           /var/tmp/go/src/github.com/containers/podman/test/e2e/cp_test.go:214
         
         [BeforeEach] Podman cp
           /var/tmp/go/src/github.com/containers/podman/test/e2e/cp_test.go:26
         [It] podman cp the root directory from the ctr to an existing directory on the host 
           /var/tmp/go/src/github.com/containers/podman/test/e2e/cp_test.go:214
         Running: podman [options] run --name copyroottohost -d quay.io/libpod/alpine:latest top
         b1f46c239b73b8e07fcfe08a03054eae512a841c5b0553e3e0cff67aa40a3292
         Running: podman [options] exec copyroottohost touch /dummy.txt
         Running: podman [options] cp copyroottohost:/ /tmp/212797057
         Error: 1 error occurred:
         	* error copying from container: copier: get: "/"("/"): copier: get: error reading "/sys/kernel/slab/dentry/cgroup/dentry(45891:podman-240179.scope)": lstat /sys/kernel/slab/dentry/cgroup/dentry(45891:podman-240179.scope): no such file or directory
         [AfterEach] Podman cp
           /var/tmp/go/src/github.com/containers/podman/test/e2e/cp_test.go:36
         Running: podman [options] stop -a --time 0
         b1f46c239b73b8e07fcfe08a03054eae512a841c5b0553e3e0cff67aa40a3292
[...]

PR #9905, opened 2021-03-31
WIP: Cirrus: Add Ubuntu to rootless integration testing

cevich:
Fixes #9508

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #10237, opened 2021-05-05
CI: run rootless tests under ubuntu

edsantiago:
Reason: to catch errors before they surface in RHEL.

One of the Ubuntus is specially crafted to run with cgroups v1
and runc. Although this isn't quite the same as RHEL, it's as
close as we can come in our CI environment, and I suspect it
would have caught #10234 (a regression).

Sorry, team.

[NO TESTS NEEDED] - we already have tests, this PR will make them run

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -539,7 +539,7 @@ rootless_integration_test_task:
     skip: *branches_and_tags
     depends_on:
         - unit_test
-    matrix: *fedora_vm_axis
+    matrix: *platform_axis
     gce_instance: *standardvm
     timeout_in: 90m
     env:
@@ -614,7 +614,7 @@ rootless_system_test_task:
     only_if: *not_docs
     depends_on:
       - rootless_integration_test
-    matrix: *fedora_vm_axis
+    matrix: *platform_axis
     gce_instance: *standardvm
     env:
         TEST_FLAVOR: sys
--- a/test/system/001-basic.bats
+++ b/test/system/001-basic.bats
@@ -10,15 +10,17 @@ function setup() {
     :
 }
 
-@test "podman --context emits reasonable output" {
-    run_podman 125 --context=swarm version
-    is "$output" "Error: Podman does not support swarm, the only --context value allowed is \"default\"" "--context=default or fail"
-
-    run_podman --context=default version
-}
+#### DO NOT ADD ANY TESTS HERE! ADD NEW TESTS AT BOTTOM!
 
 @test "podman version emits reasonable output" {
     run_podman version
+
+    # FIXME FIXME FIXME: #10248: nasty message on Ubuntu cgroups v1, rootless
+    if [[ "$output" =~ "overlay test mount with multiple lowers failed" ]]; then
+        if is_rootless; then
+            lines=("${lines[@]:1}")
+        fi
+    fi
 
     # First line of podman-remote is "Client:<blank>".
     # Just delete it (i.e. remove the first entry from the 'lines' array)
@@ -40,6 +42,17 @@ function setup() {
     fi
 }
 
+
+@test "podman --context emits reasonable output" {
+    # All we care about here is that the command passes
+    run_podman --context=default version
+
+    # This one must fail
+    run_podman 125 --context=swarm version
+    is "$output" \
+       "Error: Podman does not support swarm, the only --context value allowed is \"default\"" \
+       "--context=default or fail"
+}
 
 @test "podman can pull an image" {
     run_podman pull $IMAGE
--- a/test/e2e/common_test.go
+++ b/test/e2e/common_test.go
@@ -605,13 +605,6 @@ func SkipIfRootlessCgroupsV1(reason string) {
 	}
 }
 
-func SkipIfUnprivilegedCPULimits() {
-	info := GetHostDistributionInfo()
-	if isRootless() && info.Distribution == "fedora" {
-		ginkgo.Skip("Rootless Fedora doesn't have permission to set CPU limits")
-	}
-}
-
 func SkipIfRootless(reason string) {
 	checkReason(reason)
 	if os.Geteuid() != 0 {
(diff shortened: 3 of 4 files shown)


edsantiago:
Additions are marked with ellipse outline
cirrus-map-ci_rootless_add_ubuntu

Luap99:
Dup of #9905?

edsantiago:
Oh wow, good catch. Almost a duplicate: differences are (1) I intend to run system tests in addition to integration tests, and (2) I intend to devote myself to merging this urgently, ASAP.

cevich:
>  merging this urgently, ASAP.

@edsantiago since #9905 is dedicated to integration tests, what about limiting this PR to only the system-tests?  Reason being, my PR is blocked by an tricky 'podman cp' issue buried in buildah code.  Based on experience so far, I do not see that issue being fixed "urgently, ASAP" :disappointed: So I think if you limit this PR to just system-tests there can be some realized testing benefit while I wait for a fix.

edsantiago:
I want this merged today.

rhatdan:
I am just going to merge if it passes
/lgtm
/approve
/hold

cevich:
> I want this merged today.

I understand, but it's not worth subjecting people to random failures of `Podman cp` tests!

.cirrus.yml
```diff
@@ -539,7 +539,7 @@ rootless_integration_test_task:
     skip: *branches_and_tags
     depends_on:
         - unit_test
-    matrix: *fedora_vm_axis
+    matrix: *platform_axis
```
cevich, line 542: I would respectfully ask this specific change be dropped from this PR and left up to #9905

cevich:
> if it passes

Clarification: Issue #9997 is a high-likelihood flake.  I'm not opposed to having a `skip()` added to all the `podman cp` tests to avoid it...but that has consequences: Hiding MORE regressions on Ubuntu until #9997 is fixed (i.e. `podman cp` and `buildah copy` are very popular).

edsantiago:
@cevich if I understand your concern correctly, you believe that enabling rootless integration tests on ubuntu (not clear if 2010 or 2104) will result in a high number of flakes due to #9997. Is that correct?

If that's your concern, I will still make a case for merging my PR ASAP. First: #9905 has languished for over a month. In that time, we've had a regression that broke RHEL, that should have been caught. I want to prevent regressions, and I want to do so right now, today. Second: flakes that don't affect anyone don't get fixed. So, counterintuitively, I **want** flakes. I want people to suffer, because suffering is the only way to motivate people to fix flakes. #9997 has languished a long time too. If we start suffering because CI starts failing, the pressure will be on to find and fix it. (And, third, at the time I write this, integration tests have passed on this PR. They also passed 100% on my previous submission. Maybe #9997 is somehow already fixed by another commit?)

I will honor your `/hold`, and will not be the one to release it. I would appreciate an independent team member's resolution on this.

(And if I've misunderstood your concern, I apologize, and would welcome clarification)

test/system/001-basic.bats
```diff
@@ -10,16 +10,18 @@ function setup() {
     :
 }
 
-@test "podman --context emits reasonable output" {
-    run_podman 125 --context=swarm version
-    is "$output" "Error: Podman does not support swarm, the only --context value allowed is \"default\"" "--context=default or fail"
-
-    run_podman --context=default version
-}
```
edsantiago, line 18: I simply copied this below, adding a few comments and reordering. Reason: I want 'podman version emits reasonable output' to be the very absolute first thing we test.

cevich:
> Maybe #9997 is somehow already fixed by another commit?

It's certainly possible.  It's also easily possible the flake-rate changed due to timing changes.  

>  I want people to suffer, because suffering is the only way to motivate people to fix flakes.

While arguably effective psychological manipulation in some cases, I don't believe it's helpful as leverage for a technical decision. My opinion on that aside, we have PR's blocked by issues for a substantially similar reason:  It brings more eyeballs to a problem.

Worse...now that (apparently) the flake-rate has reduced, merging this (and by extension #9905) would actually ***LOWER*** visibility.  That would leave us with zero blocked PRs and a [hard to reproduce] issue lying around, among a hundred others to be ignored.

> misunderstood your concern

That's also part of the issue...it's an unintentional cross-device access race affecting podman and buildah, with not well understood implications - including security.  There are device-locks involved due to this access, so it's possible for this fault to negatively affect other system services and containers.

The copy problem also affects Fedora (and every other distro), we simply haven't observed it there yet.

In any case...my point is, I don't know what the implications are of ignoring #9997.  I'm open to adding a `skip()` or ignoring the problem (and merging) if the experts deem it unworthy of holding up new testing.

Lastly if you care to take a peep yourself, the [error is coming from this line](https://github.com/containers/buildah/blob/master/copier/copier.go#L1169).  It seems to be caused by a TOCTOU race on an in-copy-context symlink pointing to an out-of-context cross-device file, when the file disappears (or otherwise becomes inaccessible) immediately prior to being lstat()'ed.

edsantiago:
>I don't know what the implications are of ignoring #9997

I'm sorry, I must not have communicated properly. My intention is not to ignore that flake, it is to embrace it: give it higher visibility. Make it fail often, collect data in my flake tracker, see what patterns emerge that might help identify and fix it. (And, of course, provide stronger motivation to do so).

cevich:
> I'm sorry, I must not have communicated properly.

Nope you did fine, I understand the intent.  I should probably call the underlying problem a "race" not a "flake", since it's not completely innocent.  After looking and understanding the code more fully...there are indeed at least two TOCTOU races we happened by luck/accident to hit in the podman tests.

But I'm worried since it seems to be reproducing ***less*** now, if we merge these PRs then the problem may become easier to ignore - difficult to reproduce and lost in the shuffle of issues.

[In any case, I think Nalin and I are honing in on a fix.](https://github.com/containers/podman/issues/9997#issuecomment-833860160).  Once it's addressed in buildah, it's just a quick (hopefully) re-vendor to bring it over here.

cevich:
Hmmm....taking a step back...My motivation to not merging new testing with possible random breakage is to spare humans the agony of debugging.  If the problem is reproducing less now (for whatever reason) that does mean the risk of annoying humble humans is also probably lower.

How about let's do this:  I trust Nalin will figure something out eventually, and buildah will get fixed.  Let's give until Monday afternoon to sort it out, and if there are significant snags (like buildah test failures or problems vendoring into podman) we can just merge this and cross fingers.

Could that be a compromise?

rhatdan:
/hold cancel
Well see how the weekend goes.

PR #10238, opened 2021-05-05
Fix infinite loop in isPathOnVolume

### LINKED ISSUE #10216: Podman run blocks in `isPathOnVolume` function forever

/kind bug

**Description**

There is a bug in current [isPathOnVolume](https://github.com/containers/podman/blob/master/libpod/container_path_resolution.go#L125) function which in some cases leads to infinite loop which causes some commands to stuck. This happens cause `filepath.Dir(dest)` could return `.` in some cases and calling same method again would return same result (`.`).

1. Create a Docker image using this `Dockerfile`:
```
cat Dockerfile 
FROM alpine:3.12
VOLUME ['/etc/foo']
WORKDIR /etc/foo
CMD ["echo", "test"]
```
it can be compiled using any tool (I used original `docker build`).
**NOTE**: that `VOLUME` uses single quotes `'` and not `"`.

2. Pull that image into your system and make it available for `podman`

3. Try to start that container using that image:
```
podman run testimg
```
and this would stuck forever in the loop.

`podman run testimg` got stuck consuming cpu resources.

It should either fail or succeed. Interestingly spec like this works fine in a `docker`, it also used to work in a `podman` version 2.

**Extra information**

Basically using single quotes causes `vol.Dest` to become `[/etc/foo]` string and calling `filepath.Dir` on it will eventually return `[` and then `.` so will never match `/`.

I skipped some fields cause these are not really necessary cause issue is really issue to reproduce.

bacher09:
[filepath.Dir](https://golang.org/pkg/path/filepath/#Dir) in some cases returns `.` symbol and calling this function again returns same result. In such cases this function never returns and causes some operations to stuck forever.

Closes #10216

### DIFF
--- a/libpod/container_path_resolution.go
+++ b/libpod/container_path_resolution.go
@@ -128,7 +128,7 @@ func isPathOnVolume(c *Container, containerPath string) bool {
 		if cleanedContainerPath == filepath.Clean(vol.Dest) {
 			return true
 		}
-		for dest := vol.Dest; dest != "/"; dest = filepath.Dir(dest) {
+		for dest := vol.Dest; dest != "/" && dest != "."; dest = filepath.Dir(dest) {
 			if cleanedContainerPath == dest {
 				return true
 			}
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -921,6 +921,17 @@ USER mail`, BB)
 		Expect(session.OutputToString()).To(ContainSubstring("mail root"))
 	})
 
+	It("podman run with incorect VOLUME", func() {
+		dockerfile := fmt.Sprintf(`FROM %s
+VOLUME ['/etc/foo']
+WORKDIR /etc/foo`, BB)
+		podmanTest.BuildImage(dockerfile, "test", "false")
+		session := podmanTest.Podman([]string{"run", "--rm", "test", "echo", "test"})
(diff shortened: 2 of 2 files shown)


libpod/container_path_resolution.go
```diff
@@ -128,7 +128,7 @@ func isPathOnVolume(c *Container, containerPath string) bool {
 		if cleanedContainerPath == filepath.Clean(vol.Dest) {
 			return true
 		}
-		for dest := vol.Dest; dest != "/"; dest = filepath.Dir(dest) {
+		for dest := vol.Dest; dest != "/" && dest != "."; dest = filepath.Dir(dest) {
```
TomSweeneyRedHat, line 131: IDK, but wonder if it would be better to add at line 134.5

```
if dest == "." {
    return false}
}
```

vrothberg: To me it looks good next to the check for `"/"`.

vrothberg (commented):
Thanks for the fix.

Our CI enforces adding tests. Could you add a test with your reproducer to `test/e2e/run_test.go`?

rhatdan:
LGTM

rhatdan:

rhatdan:
Thanks @bacher09 
/lgtm
/hold cancel

PR #10243, opened 2021-05-06
channel: simplify implementation

giuseppe:
do not use recover() to prevent writing to a closed channel.  There is
already a lock, use it as well for Close and let Write check if the
channel is still active.


rhatdan:
LGTM

rhatdan:

Issue #10245, opened 2021-05-06
docker-compose cannot set static ip when use ipam


zhangguanzhang:
compose file content:
```yaml
version: "3.2"
services:
  emby:
    image: docker.io/library/alpine:latest
    networks:
      test:
        ipv4_address: 10.1.0.253
    tty: true
    command: ["top"]

networks:
  test:
    driver: bridge
    ipam:
      driver: default
      config:
        - subnet: 10.1.0.0/24
```

use docker:
```
$ docker-compose up -d
Creating network "root_test" with driver "bridge"
Creating root_emby_1 ... done
$ docker inspect root_emby_1 --format '{{ .NetworkSettings.Networks.root_test.IPAddress }}'
10.1.0.253
```
use the latest code to compile,  podman:
```
$ ./bin/podman  system service -t 0 unix:///var/run/docker.sock --log-level debug
# in windows 2
$ docker-compose up -d
$ podman inspect root_emby_1 --format '{{ .NetworkSettings.Networks.root_test.IPAddress }}'
10.1.0.2  <--------- should be 10.1.0.253
```

Issue #10240, opened 2021-05-06
Docker Compose support setting static IP


jose-pr:
/kind bug

**Description**

Static IP configuration in a docker-compose file not setting the ip.

**Steps to reproduce the issue:**

1.
```
version: "3.2"
   
services:
  emby:
    image: docker.io/emby/embyserver:latest 
    networks:
      default:
        ipv4_address: 10.7.88.2
    environment:
      UID: 0
      GUID: 0
networks:
  default:
    external:
        name: podman
```

2.
docker-compose  up
3.
```
 podman inspect emby_emby_1 | grep -i ip
"IPAddress": "",
"IPPrefixLen": 0,
"IPv6Gateway": "",
"GlobalIPv6Address": "",
"GlobalIPv6PrefixLen": 0,
"LinkLocalIPv6Address": "",
"LinkLocalIPv6PrefixLen": 0,
        "IPAddress": "10.7.88.9",
        "IPPrefixLen": 24,
        "IPv6Gateway": "",
        "GlobalIPv6Address": "",
        "GlobalIPv6PrefixLen": 0,
        "IPAMConfig": null,
```
**Describe the results you expected:**

I expected the IPAddress to be 10.7.88.2

**Output of `podman version`:**

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16
Built:        Thu Apr 22 09:11:28 2021
OS/Arch:      linux/amd64```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.11.12-300.fc34.x86_64
  linkmode: dynamic
  memFree: 272363520
  memTotal: 33620488192
  ociRuntime:
    name: crun
    package: crun-0.19.1-2.fc34.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.19.1
      commit: 1535fedf0b83fb898d449f9680000f729ba719f5
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
[...]

PR #10246, opened 2021-05-06
Fixes docker-compose cannot set static ip when use ipam

zhangguanzhang:

Fixes: https://github.com/containers/podman/issues/10245
FIXes: https://github.com/containers/podman/issues/10240


Luap99 (commented):
LGTM

Issue #10258, opened 2021-05-07
Docker compatibility: pull of uncached/new image has final status that doesn't indicate success to docker-java


candrews:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**
This issue only happens for images that podman doesn't already have cached. If podman already has the image, it returns a final status that `docker-java` interprets as success.

1. Use docker-java (such as through testcontainers) to pull an image that is not already in podman's image cache

**Describe the results you received:**
docker-java fails with this exception:
`com.github.dockerjava.api.exception.DockerClientException: Could not pull image: Pull complete`

**Describe the results you expected:**
No exception should be thrown

**Additional information you deem important (e.g. issue happens only occasionally):**
The problem is that podman's last status when pulling an image isn't one of that statuses that `docker-java` checks for, which are:
* `Download complete`
* `Image is up to date`
* `Downloaded newer image`
* `this image was pulled from a legacy registry`
* `: downloaded`

 https://github.com/docker-java/docker-java/blob/3.2.7/docker-java-api/src/main/java/com/github/dockerjava/api/model/PullResponseItem.java#L33

I ran the same test with Docker and with Podman so I could compare the data requests/responses.
Docker transcript: 
[docker.log](https://github.com/containers/podman/files/6442570/docker.log)
Podman transcript:
[podman.log](https://github.com/containers/podman/files/6442571/podman.log)

Notice that docker's pull transcript ends with a status of `Status: Downloaded newer image for alpine:3.5` which matches one o the patterns `docker-java` looks for:
```
{"status":"Download complete","progressDetail":{},"id":"8cae0e1ac61c"}\r
\r
< 2021/05/07 10:53:18.350530  length=191 from=1276 to=1466
b9\r
{"status":"Extracting","progressDetail":{"current":32768,"total":1971694},"progress":"[\\u003e                                                  ]  32.77kB/1.972MB","id":"8cae0e1ac61c"}\r
\r
< 2021/05/07 10:53:18.471037  length=192 from=1467 to=1658
ba\r
{"status":"Extracting","progressDetail":{"current":360448,"total":1971694},"progress":"[=========\\u003e                                         ]  360.4kB/1.972MB","id":"8cae0e1ac61c"}\r
\r
< 2021/05/07 10:53:18.498898  length=193 from=1659 to=1851
bb\r
[...]

PR #10259, opened 2021-05-07
Report Download complete in Compatibility mode

rhatdan:
Fixes: https://github.com/containers/podman/issues/10258

[NO TESTS NEEDED] Difficult to create tests for this.


rhatdan:
@mheon Should we just always return "Download complete"?

mheon:
I'm fine with always returning "Download complete". What the Java library is doing here is honestly nonsensical, IMO - string-arsing human readable output should not be expected to produce sane results.

jwhonce:
@mheon Comes from a poor API

/lgtm
/hold

PR #10265, opened 2021-05-07
fix: compat API "images/get" for multiple images

matejvasek:
Fix Docker APIv2 `images/get` endpoint to be able to download multiple images.


rhatdan:
LGTM

vrothberg (commented):
A late LGTM, thanks for the fix!

Please make sure to add a commit message. "fix API" will not aid in understanding the commit without reading the code, which ca be tiresome when browsing the git history.

Issue #10263, opened 2021-05-07
mount command does not work inside podman containers


vikas-goel:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When mount command is run inside containers managed by Podman, the command shows nothing. The command works and shows all the mount points when the same containers are managed by Docker.

In the absence of mount command output, the application inside the existing containers fail to work.

In Docker managed container, `/etc/mtab` is a symbolic link to `/proc/mounts`. Whereas, in Podman the symbolic link is missing. When the symbolic link is created manually in Podman managed container, the `mount` command works same as it does in Docker environment.

Podman environment
```
bash-4.2# ls -l /etc/mtab
-rw-r--r--. 1 root root 0 May  6 17:09 /etc/mtab
bash-4.2# mount
bash-4.2#
bash-4.2# ln -sf /proc/mount /etc/mtab
bash-4.2# mount
overlay on / type overlay (rw,relatime,context=system_u:object_r:container_file_ ...
```

Docker environment
```
bash-4.2# ls -l /etc/mtab
lrwxrwxrwx. 1 root root 12 Mar 28 01:16 /etc/mtab -> /proc/mounts
bash-4.2# mount
overlay on / type overlay (rw,relatime,context=system_u:object_r:container_file_t ...
```

**Steps to reproduce the issue:**

1. Load a Red Hat based container image with `mount` package installed in Podman environment

2. Run a container using the container image

3. Run `mount` command inside the container

**Describe the results you received:**
The mount command shows no output

**Describe the results you expected:**
The mount command should show the mount points as available in /proc/mounts

**Additional information you deem important (e.g. issue happens only occasionally):**
The behavior is consistent and reproducible 100% of time

**Output of `podman version`:**

```
Version:      3.0.2-dev
API Version:  3.0.0
Go Version:   go1.15.7
Built:        Tue Mar  2 07:10:06 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.26-1.module+el8.4.0+10198+36d1d0e3.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.26, commit: 0a5175681bdd52b99f1f0f442cbba8f8c126a1c9'
  cpus: 8
  distribution:
[...]

PR #10270, opened 2021-05-07
Create the /etc/mtab file if does not exists

rhatdan:
We should create the /etc/mtab->/proc/mountinfo link
so that mount command will work within the container.

Docker does this by default.

Fixes: https://github.com/containers/podman/issues/10263

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1530,6 +1530,16 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
 		}()
 	}
 
+	// If /etc/mtab does not exist in container image, then we need to
+	// create it, so that mount command within the container will work.
+	mtab := filepath.Join(mountPoint, "/etc/mtab")
+	if err := os.MkdirAll(filepath.Dir(mtab), 0755); err != nil {
+		return "", errors.Wrap(err, "error creating mtab directory")
+	}
+	if err = os.Symlink("/proc/mounts", mtab); err != nil && !os.IsExist(err) {
+		return "", err
+	}
+
 	// Request a mount of all named volumes
 	for _, v := range c.config.NamedVolumes {
 		vol, err := c.mountNamedVolume(v, mountPoint)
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
All kinds of red unhappy tests @rhatdan

mheon:
Can we do this with a bind mount of the link? As is, we’re creating a file
that will show up in diff, be committed with user changes, etc

giuseppe (approved):
LGTM

rhatdan:
@mheon We already do this, with /etc/resolv.conf and /etc/hosts, if they do not exist in the base image.  We could create an empty file for the inode and mount over it, but I am not sure this is better then setting up the image with the symbolic link, which BTW almost all images have by default, which is why we never caught this until now.  We can talk at standup about potentially special handling the additional mount inodes we add to images.

mheon:
We bind-mount hosts, resolv, and most others (with a few notable
exceptions) - it seems more sane than creating files in the container’s own
RW layer unless there is a good reason not to

libpod/container_internal.go
```diff
@@ -1530,6 +1530,18 @@ func (c *Container) mountStorage() (_ string, deferredErr error) {
...
 	}
 
+	// If /etc/mtab does not exist in container image, then we need to
+	// create it, so that mount command within the container will work.
+	mtab := filepath.Join(mountPoint, "/etc/mtab")
+	if err := os.MkdirAll(filepath.Dir(mtab), 0755); err != nil {
+		return "", errors.Wrap(err, "error creating mtab directory")
+	}
+	if _, err := os.Lstat(mtab); err != nil {
+		if err = os.Symlink("/proc/mounts", mtab); err != nil {
```
giuseppe, line 1540: could this be simplified as?

```
if err = os.Symlink("/proc/mounts", mtab); err != nil && !os.IsExist(err) {
	return "", err
}
```

giuseppe (approved):
just a nit that could be addressed later,

LGTM

TomSweeneyRedHat:
LGTM
but I'll let @mheon hit the merge button based on his last comment.

PR #10271, opened 2021-05-07
fix: improved "containers/{name}/wait" endpoint

### LINKED ISSUE #10256: Improve Docker APIv2 container wait reliability
As of now `podman` uses polling to wait for a condition. This approach is however unreliable.

There is an issue when waiting for status change. There is a condition called `next-exit`, it basically waits for transition `running->not-running`. It's possible that for short living containers that the `running` state is never observed by polling. Even non-sleeping busy waiting.

We need to figure how to solve this.

matejvasek:
Previous implementation wasn't working with short living containers.
Polling missed the very brief moment the container was running.

resolves #10256


pkg/api/handlers/utils/containers.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/containers/podman/v3/libpod/events"
+
```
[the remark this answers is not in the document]
rhatdan: LGTM

mheon:
Looks like this could leak goroutines - think you need to close the channel
once you have the event so event reading terminates

matejvasek:
> Looks like this could leak goroutines - think you need to close the channel once you have the event so event reading terminates

@mheon  Good catch. Closing the eventChannel would cause `panic` as the gorutine would try to write to closed channel. Correct way to terminate the event watcher is to cancel the passed context. The context here is done once the HTTP request is ended. So in the end there is no leak. However I will add explicit cancellation for clarity.

pkg/api/handlers/utils/containers.go
```diff
@@ -202,12 +204,35 @@ func waitRemoved(ctrWait containerWaitFn) (int32, error) {
 	return code, err
 }
 
-func waitNextExit(ctrWait containerWaitFn) (int32, error) {
-	_, err := ctrWait(define.ContainerStateRunning)
-	if err != nil {
+func waitNextExit(ctx context.Context, containerName string) (int32, error) {
+	runtime := ctx.Value("runtime").(*libpod.Runtime)
+	var containerEngine entities.ContainerEngine = &abi.ContainerEngine{Libpod: runtime}
```
vrothberg, line 209:
```suggestion
	containerEngine := abi.ContainerEngine{Libpod: runtime}
```

pkg/api/handlers/utils/containers.go
```diff
@@ -202,12 +204,35 @@ func waitRemoved(ctrWait containerWaitFn) (int32, error) {
...
+	select {
+	case evt, ok := <-eventChannel:
+		if ok {
+			return int32(evt.ContainerExitCode), nil
+		}
+		select {
+		case err := <-errChannel:
+			return -1, err
+		default:
+			return -1, nil
```
vrothberg, line 231: This should be an error, for instance, `internal error: unexpected empty data waiting for container`.

It should not happen but if it happens, we know where to look.

pkg/api/handlers/utils/containers.go
```diff
@@ -202,12 +204,35 @@ func waitRemoved(ctrWait containerWaitFn) (int32, error) {
...
+	runtime := ctx.Value("runtime").(*libpod.Runtime)
+	var containerEngine entities.ContainerEngine = &abi.ContainerEngine{Libpod: runtime}
+	eventChannel := make(chan *events.Event)
+	errChannel := make(chan error)
+	opts := entities.EventsOptions{
+		EventChan: eventChannel,
+		Filter:    []string{"event=died", fmt.Sprintf("container=%s", containerName)},
+		Stream:    true,
+	}
+	ctx, cancel := context.WithCancel(ctx)
```
vrothberg, line 217: Please add a comment to explain why the cancel is needed (i.e., we need to stop streaming once we found the event).

pkg/api/handlers/utils/containers.go
```diff
@@ -202,12 +204,35 @@ func waitRemoved(ctrWait containerWaitFn) (int32, error) {
...
 }
 
-func waitNextExit(ctrWait containerWaitFn) (int32, error) {
-	_, err := ctrWait(define.ContainerStateRunning)
-	if err != nil {
+func waitNextExit(ctx context.Context, containerName string) (int32, error) {
+	runtime := ctx.Value("runtime").(*libpod.Runtime)
+	var containerEngine entities.ContainerEngine = &abi.ContainerEngine{Libpod: runtime}
+	eventChannel := make(chan *events.Event)
+	errChannel := make(chan error)
```
vrothberg, line 211: Should both channels be closed in a deferred statement?

pkg/api/handlers/utils/containers.go
matejvasek: I deliberately used explicit type so I just call interface method, not method specific to the impl struct.

pkg/api/handlers/utils/containers.go
matejvasek: AFAIK channels don't need to be closed. It's sometime useful (e.g. to be able to end `for` loop over channel), but that's not the case here.

pkg/api/handlers/utils/containers.go
matejvasek: I will rework this code.

matejvasek:
@mheon I tried to compile `podman` without `systemd` support and this is not working. Any idea why?

pkg/api/handlers/utils/containers.go
matejvasek: @vrothberg PTAL

pkg/api/handlers/utils/containers.go
```diff
@@ -202,12 +203,30 @@ func waitRemoved(ctrWait containerWaitFn) (int32, error) {
...
+	defer cancel()
+	go func() {
+		errChannel <- containerEngine.Events(ctx, opts)
+	}()
+
+	evt, ok := <-eventChannel
+	if ok {
+		return int32(evt.ContainerExitCode), nil
+	} else {
+		return -1, <-errChannel
```
vrothberg, line 228: What if we get `nil` from the `errChannel`?

Since we're streaming, the container _must_ die at some point.  So if we reach line 228, somethings off and I think we need a custom error for that (see previous comment).

matejvasek: >What if we get nil from the errChannel?

I thin that is all right, isn't it? It happens if the request was canceled.

>Since we're streaming, the container must die at some point.

But the container may die after request cancelation.

The line 228 can be reached if the context was canceled (e.g. client closed connection) in which case err is nil.

matejvasek: we reach 228 if `ok == false` which happens if the `eventChannel` has been closed, and the channel is closed by `containerEngine.Events()` just before the function exits.

matejvasek: does it make sense @vrothberg ?

pkg/api/handlers/utils/containers.go
```diff
@@ -202,12 +203,29 @@ func waitRemoved(ctrWait containerWaitFn) (int32, error) {
...
+	defer cancel()
+	go func() {
+		errChannel <- containerEngine.Events(ctx, opts)
+	}()
+
+	evt, ok := <-eventChannel
+	if ok {
+		return int32(evt.ContainerExitCode), nil
+	}
+	return -1, <-errChannel
```
vrothberg, line 228: Please, let's discuss the right approach before changing the code.

I still think we should return a custom error here. `errChannel` could return `nil` in line 228 and I believe that it must be an error that clearly identifies the (code) location.  The channel should not be closed since we're streaming, so we must receive the exit code of the container at some point.  To me, line 228 should not happen unless there's an internal error.

matejvasek: > The channel should not be closed since we're streaming, so we must receive the exit code of the container at some point.

But it can be closed. It happens if `ctx` was canceled.

> To me, line 228 should not happen unless there's an internal error.

It can can happen and it's completely all right. It happens if `ctx` was canceled.

The `ctx` is canceled if request is canceled.
Example:
Your run `curl "http://localhost:1234/containers/${CTR}/wait?condition=next-exit" -v -XPOST`
but before the `curl` exits you pres `ctrl+c`.

vrothberg: Thank you, yes that makes sense!  Could you add a code comment there explaining the reasoning?

matejvasek: @vrothberg I added a comment before the final return, is it sufficient?

vrothberg (commented):
LGTM, thank you!

Issue #9869, opened 2021-03-29
podman-remote build: --rm=false is ignored


edsantiago:
In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ cat Dockerfile
FROM alpine
RUN date
RUN echo hello
RUN echo goodbye
$ ../bin/podman-remote build --force-rm=false --layers --rm=false -t foo .
STEP 1: FROM alpine
STEP 2: RUN date
Mon Mar 29 18:28:51 UTC 2021
--> b9328ccce3e
STEP 3: RUN echo hello
hello
--> f5b01f9cb4c
STEP 4: RUN echo goodbye
goodbye
STEP 5: COMMIT foo
--> 9a41527cdb1
9a41527cdb1dc71359791d4db8d5c542c1f4974b0341c3915ebd4b751d27577e
$ buildah containers
CONTAINER ID  BUILDER  IMAGE ID     IMAGE NAME                       CONTAINER NAME
```
That is: empty output from `buildah containers`. Expected multiple non-deleted layers:
```console
$ buildah containers
CONTAINER ID  BUILDER  IMAGE ID     IMAGE NAME                       CONTAINER NAME
fa2db0a99cfd     *     302aba9ce190 docker.io/library/alpine:latest  alpine-working-container
5fb1c74a92a2     *     fa3e1b440db3                                  fa3e1b440db32af589c022c212375db64359d7381ed64380ab9f7b095e8b7380-working-container
fd7777c2756d     *     d9c3da18bc3e                                  d9c3da18bc3e5f3db3fcd183cf0d54fa109e136754335cc5c5df2211e6e78da9-working-container
```

edsantiago:
I'm pretty confused. This is not fixed, even in 8c0df1c44 with buildah-v1.20.1-0.20210402144408-36a37402d0c8. And looking at #9884, I see:
```
        if options.RemoveIntermediateCtrs {
		params.Set("rm", "1")
	}
```
...but this issue relates to `--rm=false`, and I'm wondering if there's a default somewhere (I can't find it) such that the above is a NOP, and what we need is to override the default `1`? Reopening.

edsantiago:
Still present in 54bed1025d07bc5f77ee4e1e7f942157e211ec0a

Issue #10154, opened 2021-04-27
podman build: stream dropped, unexpected failure


edsantiago:
Looks like an unintended consequence of #10034 . I'm seeing a lot of these failures in my buildah-bud-tests-under-podman-remote testing (#9887):
```
# podman-remote build ....
STEP 1: FROM <some image>
STEP 2: something-somethng
Error: stream dropped, unexpected failure
[rc=125]
```
Logs, all from fedora34-beta:
* [first attempt](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6745318667583488/html/bud-remote-fedora-34beta-root-host.log.html#t--00118)
* [second attempt](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5257019318861824/html/bud-remote-fedora-34beta-root-host.log.html#t--00118)
* [third attempt](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5265760818823168/html/bud-remote-fedora-34beta-root-host.log.html#t--00090) (this one has three separate failures)

Four of the five failures are in tests that expect rc=125; but one of them is in a test that expects rc=0.

This is based on master @ 476c76f58 as of yesterday afternoon (Apr 26)

edsantiago:
Two more failures in [my next rerun](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5742129533157376/html/bud-remote-fedora-34beta-root-host.log.html) (plus one more flake which I'm just going to stick my head in the sand about)

edsantiago:
[Still happening](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6011994407960576/html/bud-remote-fedora-34-root-host.log.html)

edsantiago:
Update: although it _mostly_ happens in the `bud with copy-from with bad from flag in Dockerfile with --layers` test, not always: the "third attempt" and "my next rerun" links, above, include failures in different tests.

This bug makes it impossible to merge #9887

edsantiago:
[Another one](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6509655984701440/html/bud-remote-fedora-34-root-host.log.html), in bud-http-Dockerfile test

edsantiago:
[...]

PR #10292, opened 2021-05-10
Fix podman-remote build --rm=false ...

rhatdan:
Fixes: https://github.com/containers/podman/issues/9869

### DIFF
--- a/pkg/bindings/images/build.go
+++ b/pkg/bindings/images/build.go
@@ -127,6 +127,8 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
 	}
 	if options.RemoveIntermediateCtrs {
 		params.Set("rm", "1")
+	} else {
+		params.Set("rm", "0")
 	}
 	if len(options.From) > 0 {
 		params.Set("from", options.From)


edsantiago:
I sort-of have tests for it, in #9887 (enable podman-remote for buildah-bud tests). But I can't merge that until #10154 (flake: stream dropped) is fixed, because that's a very-easy-to-trigger flake and it would cause failures in pretty much every single  CI run.

mheon:
LGTM with a test

rhatdan:
@mheon Since @edsantiago Thinks he has tests for this, and it is clearly correct, I am going to merge, then we can work on fixing the blockage to @edsantiago tests.

edsantiago:
There's only one `bud.bats` test that triggers this, `bud with --rm flag`. I've pulled Dan's PR, removed the `skip` from my podman-remote buildah-bud tests, and confirmed that the test now passes.

If we can get some love for #10154, I can then push to merge #9887, and we'll have tests. I'm OK with merging this PR as it is.

rhatdan:
@edsantiago I have given more hate then love to #10154, and have no idea why this is dropping.  I will attempt to take another look, but could probably use some help from @nalind and @jwhonce To see if they have any ideas.

edsantiago:
>more hate then love

Ah, there's your problem: hate leads to the dark side. Before you know it you'll be hobnobbing with Sith lords and zipping around in a TIE fighter. (Which, come to think of it, sounds kind of neat).

TomSweeneyRedHat:
LGTM
will let @mheon press the shiny merge button.

mheon:
Only have email even

PR #10318, opened 2021-05-12
fix network restart always test

Luap99:
The added test in 30544f225e73 is flaking. Podman inspect is always
working so we have to check the pid instead.

### DIFF
--- a/test/system/500-networking.bats
+++ b/test/system/500-networking.bats
@@ -88,8 +88,9 @@ load helpers
     # Wait for container to restart
     retries=20
     while :;do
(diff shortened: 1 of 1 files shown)


Luap99:
@edsantiago PTAL, the log from the flake: https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4680692446527488/html/sys-podman-ubuntu-2010-root-host.log.html

giuseppe (approved):
LGTM

edsantiago:
Oops. Sorry about that, I should've tested better.

/lgtm

PR #10324, opened 2021-05-12
[NO TESTS NEEDED] Print "extracting" only on compressed file

ashley-cui:
We should only print "extracting compressed file" when the file is
actually compressed

[NO TESTS NEEDED] because its podman machine


rhatdan:
/approve
LGTM

Issue #9985, opened 2021-04-09
Container created with --pid host returns operation not permitted on copy in F34


JayDoubleu:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Container created with --pid host returns `operation not permitted` error when trying to copy file into container.

**Steps to reproduce the issue:**

1. Create container with `podman run  --ipc host --name pidtest --pid host --privileged --security-opt label=disable --ulimit host --user root:root -it registry.fedoraproject.org/fedora:34 /bin/bash`

2. Try copying file into container `touch /tmp/test && podman cp /tmp/test pidtest:/tmp/test`

**Describe the results you received:**
`Error: "/tmp/test" could not be found on container pidtest: operation not permitted`

**Describe the results you expected:**
File should be copied and no errors returned.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version
Version:      3.1.0
API Version:  3.1.0
Go Version:   go1.16
Built:        Tue Mar 30 14:29:36 2021
OS/Arch:      linux/amd64

```

**Output of `podman info --debug`:**

```
podman info --debug
host:
  arch: amd64
  buildahVersion: 1.20.0
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-1.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 24
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: thinkD
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.11.11-300.fc34.x86_64
  linkmode: dynamic
  memFree: 26587058176
  memTotal: 33655853056
  ociRuntime:
    name: crun
    package: crun-0.18-5.fc34.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.18
      commit: 808420efe3dc2b44d6db9f1a3fac8361dde42a95
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
[...]

PR #10327, opened 2021-05-13
Fix problem copying files when container is in host pid namespace

rhatdan:
When attempting to copy files into and out of running containers
within the host pidnamespace, the code was attempting to join the
host pidns again, and getting an error. This was causing the podman
cp command to fail. Since we are already in the host pid namespace,
we should not be attempting to join.  This PR adds a check to see if
the container is in NOT host pid namespace, and only then attempts to
join.

Fixes: https://github.com/containers/podman/issues/9985

### DIFF
--- a/libpod/container_copy_linux.go
+++ b/libpod/container_copy_linux.go
@@ -237,21 +237,32 @@ func (c *Container) joinMountAndExec(ctx context.Context, f func() error) error 
 		}
 		defer mountFD.Close()
 
-		pidFD, err := getFD(PIDNS)
+		inHostPidNS, err := c.inHostPidNS()
 		if err != nil {
-			errChan <- err
+			errChan <- errors.Wrap(err, "checking inHostPidNS")
 			return
 		}
-		defer pidFD.Close()
+		var pidFD *os.File
+		if !inHostPidNS {
+			pidFD, err = getFD(PIDNS)
+			if err != nil {
+				errChan <- err
+				return
+			}
+			defer pidFD.Close()
+		}
+
 		if err := unix.Unshare(unix.CLONE_NEWNS); err != nil {
 			errChan <- err
 			return
 		}
-		if err := unix.Setns(int(pidFD.Fd()), unix.CLONE_NEWPID); err != nil {
-			errChan <- err
-			return
-		}
 
+		if pidFD != nil {
+			if err := unix.Setns(int(pidFD.Fd()), unix.CLONE_NEWPID); err != nil {
+				errChan <- err
+				return
+			}
+		}
 		if err := unix.Setns(int(mountFD.Fd()), unix.CLONE_NEWNS); err != nil {
 			errChan <- err
 			return
--- a/libpod/container_inspect.go
+++ b/libpod/container_inspect.go
@@ -892,3 +892,26 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
 
 	return hostConfig, nil
 }
+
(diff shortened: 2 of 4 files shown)


giuseppe (approved):
I guess we join the pidns to not break `/proc`?

LGTM

libpod/container_inspect.go
```diff
@@ -890,3 +890,26 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
 
 	return hostConfig, nil
 }
+
+func (c *Container) hostPidNS() (bool, error) {
+	if c.config.PIDNsCtr != "" {
+		return false, nil
```
mheon, line 896: Technically, we could go further here, and call hostPidNS() on that container - it could potentially be a pid=host container that we joined the namespace of?

I don't know if this is particularly valuable though.

TomSweeneyRedHat:
@rhatdan changes LGTM, but it appears you've a gofmt error somewhere.

test/e2e/cp_test.go
```diff
@@ -91,6 +91,49 @@ var _ = Describe("Podman cp", func() {
 		Expect(roundtripContent).To(Equal(originalContent))
 	})
 
+	// Copy a file to the container, then back to the host in --pid=host
+	It("podman cp --pid=host file", func() {
```
vrothberg, line 95: Apologies for noticing the PR just now:  Would you consider moving the test logic to the system tests?  When rewriting the copy backend, I made sure to have a good coverage in the system tests since they're used in gating and also by QE.

libpod/container_inspect.go
```diff
@@ -890,3 +890,25 @@ func (c *Container) generateInspectContainerHostConfig(ctrSpec *spec.Spec, named
 
 	return hostConfig, nil
 }
+
+func (c *Container) hostPidNS() (bool, error) {
```
vrothberg, line 894: I read `hostPidNS()` like it's returning the host's PID NS.

```suggestion
// Return true if the container is running in the host's PID NS.
func (c *Container) inHostPidNS() (bool, error) {
```

libpod/container_copy_linux.go
```diff
@@ -237,21 +237,32 @@ func (c *Container) joinMountAndExec(ctx context.Context, f func() error) error
...
 		if err != nil {
-			errChan <- err
+			errChan <- errors.Wrap(err, "checking hostPidNS")
 			return
 		}
-		defer pidFD.Close()
-		if err := unix.Unshare(unix.CLONE_NEWNS); err != nil {
-			errChan <- err
-			return
+		var pidFD *os.File
```
vrothberg, line 245: Do we need all the unsharing and the limbo above if `hostPidNS == true`?  I wonder if we couldn't just do a:
```golang
if hostPidNS {
     errChan <- f()
     return
}
```

... or even do that before line 219 outside of the go routine.

rhatdan: We still need to join the mount namespace?

vrothberg: Absolutely. Thanks for clarifying!

vrothberg (commented):
LGTM

TomSweeneyRedHat:
Still fighting with the test system on this one @rhatdan

test/e2e/cp_test.go
vrothberg, line 95: To make ubuntu happy.

```suggestion
	It("podman cp --pid=host file", func() {
	SkipIfRootlessCgroupsV1("Not supported for rootless + CGroupsV1")
```

PR #10254, opened 2021-05-06
Add support for cli network prune --filter flag

jmguzik:
Missing cli functionality added.
/kind feature

PR #10328, opened 2021-05-13
Several shell completion fixes

Luap99:
- fix network filters
- add prune filters
- pod create --share support comma separated namespaces

Also fix the volume prune man page which listed invalid prune filters.

### DIFF
--- a/cmd/podman/common/completion.go
+++ b/cmd/podman/common/completion.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/podman/v3/cmd/podman/registry"
 	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/pkg/domain/entities"
+	"github.com/containers/podman/v3/pkg/network"
 	"github.com/containers/podman/v3/pkg/registries"
 	"github.com/containers/podman/v3/pkg/rootless"
 	systemdDefine "github.com/containers/podman/v3/pkg/systemd/define"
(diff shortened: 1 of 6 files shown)


Luap99:
This is the second time this week that dependabot breaks the CI. Both after #10303 and #10316 the Test Code Consistency check was failing due an change in the go.sum file. How can the original PR pass this test but break for all future PRs until the missing line is added to go.sum?

@vrothberg @rhatdan @mheon PTAL

TomSweeneyRedHat:
@Luap99 all kinds of test unhappiness.

Luap99:
@TomSweeneyRedHat The tests are failing because podman.io is down ;(

jmguzik:
@Luap99 @TomSweeneyRedHat 
I opened similar PR a couple of days ago. Could you check how much this ovelaps?
https://github.com/containers/podman/pull/10254

Luap99:
@jmguzik This does not interfere. I merged your PR and rebased this one to also add shell completion for network prune --filter.

rhatdan:
LGTM
@edsantiago PTAL

docs/source/markdown/podman-volume-prune.1.md
```diff
@@ -21,14 +21,9 @@ Do not prompt for confirmation.
 
 #### **--filter**
 
-Filter volumes to be pruned. Volumes can be filtered by the following attributes:
+Filter volumes to be pruned. Volumes can be filtered by the following attribute:
```
edsantiago, line 24: Is this true? I seem to be able to use every one of those (the removed entries) as an argument to `podman volume prune --filter`.

Luap99: You are right. This was apparently only removed from the compat api endpoint.

edsantiago:
LGTM. If you have to push again (please don't re-push just for this), I think future maintainers might appreciate a checkin comment that includes something like "prune --labels and --until are only stubs; implementing actual completion is TBI".

TomSweeneyRedHat:
LGTM
assuming happy tests.

PR #10331, opened 2021-05-13
Update go.sum

font:
Running `make vendor` to add a missing dependency hash.


TomSweeneyRedHat:
LGTM but would like @vrothberg to give this an eyeball.  I'm not sure why past vendors didn't make this change?

mheon:
LGTM. Dependabot seems broken at the moment, which is a damn pity because it was working so well.

giuseppe (approved):
/lgtm
/hold

Issue #9371, opened 2021-02-15
SELinux volume permission issues when playing kube file (rootless)


ghost:
/kind bug

**Description**

As mentioned in https://github.com/containers/podman/issues/8710 I have been having issues playing a kube file that I generated with podman. The problem seems to persist (I'm not 100% sure it's still the same issue).
Now on v3.0.0 I have permission issues with the volumes.

**Steps to reproduce the issue:**

SELinux is set to "enforcing".

_Scenario A: running everything manually:_

1. `podman pod create pod -n testpod`
2. create a simple Dockerfile that writes into a file on a mounted volume: 
```
FROM docker.io/centos:centos8
USER root
CMD while true; do echo $(date) >> /testdir/testfile && sleep 3 ; done
```
3. build it then run the container in the pod: `podman run -d --rm --name testcontainer --pod testpod -v $(pwd)/testdir:/testdir:Z testimage`
4. let's look at the testfile:
```
 tail $(pwd)/testdir/testfile
Mon Feb 15 13:04:13 UTC 2021
Mon Feb 15 13:04:16 UTC 2021
Mon Feb 15 13:04:19 UTC 2021
Mon Feb 15 13:04:22 UTC 2021
Mon Feb 15 13:04:25 UTC 2021
Mon Feb 15 13:04:28 UTC 2021
```
So far so good.

5. Remove the testfile: `rm -f $(pwd)/testdir/testfile`

_Scenario B: using podman generate:_
1. `podman pod create pod -n testpod`
2. create a simple Dockerfile: 
```
FROM docker.io/centos:centos8
USER root
CMD while true; do echo $(date) >> /testdir/testfile && sleep 3 ; done
```
3. build it then run the container in the pod: `podman run -d --rm --name testcontainer --pod testpod -v $(pwd)/testdir:/testdir:Z testimage`
4. generate a YAML file while the container is running: `podman generate kube <pod id> > kube.yaml`
5. the kube.yaml file looks like this:
```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2021-02-15T12:40:29Z"
  labels:
    app: testpod
  name: testpod
spec:
  containers:
  - command:
    - /bin/sh
    - -c
    - while true; do echo $(date) >> /testdir/testfile && sleep 3 ; done
    env:
    - name: PATH
      value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
    - name: TERM
      value: xterm
    - name: container
[...]

PR #10339, opened 2021-05-14
Support automatic labeling of kube volumes

rhatdan:
Allow users to specify options on the volume mount path.
This will trigger relabels of user specifies :z,:Z
Also will handle User Relabels if the user specifies :U

Fixes: https://github.com/containers/podman/issues/9371

### DIFF
--- a/test/system/700-play.bats
+++ b/test/system/700-play.bats
@@ -51,18 +51,40 @@ spec:
       seLinuxOptions:
          level: "s0:c1,c2"
       readOnlyRootFilesystem: false
+    volumeMounts:
+    - mountPath: /testdir:z
+      name: home-podman-testdir
     workingDir: /
+  volumes:
+  - hostPath:
+      path: TESTDIR
+      type: Directory
+    name: home-podman-testdir
 status: {}
 "
 
+RELABEL="system_u:object_r:container_file_t:s0"
+
 @test "podman play with stdin" {
-    echo "$testYaml" > $PODMAN_TMPDIR/test.yaml
+    TESTDIR=$PODMAN_TMPDIR/testdir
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@rhatdan lint is complaining about trailing whitespace.

giuseppe (approved):
LGTM

test/system/700-play.bats
```diff
@@ -51,18 +51,40 @@ spec:
...
 
+RELABEL="system_u:object_r:container_file_t:s0"
+
 @test "podman play with stdin" {
-    echo "$testYaml" > $PODMAN_TMPDIR/test.yaml
+    TESTDIR=$PODMAN_TMPDIR/testdir
+    mkdir -p $TESTDIR
+    echo "$testYaml" | sed "s|TESTDIR|${TESTDIR}|g" > $PODMAN_TMPDIR/test.yaml
     run_podman play kube - < $PODMAN_TMPDIR/test.yaml
+    if [ -e /usr/sbin/selinuxenabled -a /usr/sbin/selinuxenabled ]; then
```
edsantiago, line 73: This does not do what you think it does (the `/usr/sbin/selinuxenabled` on the right-hand side is not executed, it is an empty-string check (equivalent to `test -n /usr/sbin/selinuxenabled`). I can fix this in a followup PR if you need to merge this urgently.

rhatdan: We have this same check in other locations.  But let's fix it elsewhere so we can get this merged and move along.

TomSweeneyRedHat:
LGTM
(and dang, @ashley-cui beat me to the shiny merge button once again!)

Issue #10190, opened 2021-05-03
feature: podman auto-update --prune


vrothberg:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

`podman auto-update` should have a way to clean up dangling images.  As mentioned on Twitter [1], auto updates may contribute to the disk running out of space over time.

I am not yet entirely sure whether there should be a `--prune` option or if we should extend the auto-update systemd unit to run `podman image prune`.  At the moment, I have a preference towards extending the systemd unit as the design seems cleaner.

[1] https://twitter.com/r_isc_y/status/1388981737011793921

vrothberg:
@mrguitar, I think that's a rather useful feature on the Edge.  What's your take on it?

PR #10341, opened 2021-05-14
auto-update service: prune images

vrothberg:
Extend the systemd auto-update service to prune images after an update
has run.  As reported by a user [1], auto updates can over time cause
the disk to run out of space.  With Edge being a target use case, we
need to make sure that systems can run without much supervision, so
let's make sure to run `podman image prune` to clean up dangling images.

[1] https://twitter.com/r_isc_y/status/1388981737011793921

Fixes: #10190

@containers/podman-maintainers @mrguitar PTAL


Luap99 (commented):
LGTM

rhatdan:
This is fine, but at some point we might want to attempt rollback.
/lgtm

PR #10347, opened 2021-05-15
Rootless: Fix user access to secrets mount

flouthoc:
`Non-root` user cannot access mounted secrets while performing `podman run`.
Ref issue: https://github.com/containers/common/issues/555

`Idea`:
* `/run/secrets` dir mounts with ownership `0:0` even if user is specified via `--user`.
* Allow  `execute (x)` on `/run/secrets` if env is `rootless` or owner's uid is not `0`

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -2339,13 +2339,33 @@ func (c *Container) createSecretMountDir() error {
 		oldUmask := umask.Set(0)
 		defer umask.Set(oldUmask)
 
-		if err := os.MkdirAll(src, 0644); err != nil {
+		if err := os.MkdirAll(src, 0755); err != nil {
 			return err
 		}
 		if err := label.Relabel(src, c.config.MountLabel, false); err != nil {
 			return err
 		}
-		if err := os.Chown(src, c.RootUID(), c.RootGID()); err != nil {
+		userForSecretMount := c.RootUID()
+		groupForSecretMount := c.RootGID()
+		if c.User() != "" {
+			splitUser := strings.SplitN(c.config.User, ":", 2)
+			user := splitUser[0]
+			group := splitUser[0]
+			if len(splitUser) > 1 {
+				group = splitUser[1]
+			}
+			uid, errUID := strconv.Atoi(user)
+			gid, errGID := strconv.Atoi(group)
+			// Make sure user:group were parsed
+			// If not then go with RootUid, RootGid
+			if errUID == nil {
+				userForSecretMount = uid
+			}
+			if errGID == nil {
+				groupForSecretMount = gid
+			}
+		}
+		if err := os.Chown(src, userForSecretMount, groupForSecretMount); err != nil {
 			return err
 		}
 		c.state.BindMounts["/run/secrets"] = src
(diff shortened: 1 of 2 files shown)


libpod/container_internal_linux.go
```diff
@@ -2339,13 +2339,39 @@ func (c *Container) createSecretMountDir() error {
 		oldUmask := umask.Set(0)
 		defer umask.Set(oldUmask)
 
-		if err := os.MkdirAll(src, 0644); err != nil {
+		srcMountPermission := 0644
+		// If env is rootless give directory execute permission
+		// Non-root ( uid != 0 user can only access dir if it has `x`
+		if rootless.IsRootless() {
+			srcMountPermission = 0744
```
rhatdan, line 2346: Permissions should just be 755. In rootful and rootless. 644 makes no sense.

flouthoc: @rhatdan  I am confused about `755` , I just escalated mount to `644` -> `744` for `non-root` user so that only mount owner could access this dir. I thought thats what author thought initially , if we are going to bump this to `755` all the created users on container would have access to secrets. Not sure but could this be a potential security flaw since they are mounted secrets ? WDYT ?

But I can change this to  `755` if that is expected behavior.

flouthoc: @rhatdan I verified this with `docker swarm` it has analogous behavior, in places even tighter permissions. For instance when owner is root , permissions is `444` and adds execute when non-root user is specified or in some use-cases people are doing it manually. https://stackoverflow.com/questions/53385930/docker-secrets-with-non-root-user

I have added a test for rootless user and it seems to be passing also I have verified this manually just pasting output here
```bash
SECRET="mysecret"
echo a_secret | podman secret create "$SECRET" -
$ podman run --secret "$SECRET" --rm --user 200:200 -i -t docker.io/library/debian sh
$ cat /run/secrets/mysecret
a_secret
$ 
```

flouthoc: @rhatdan Let me know if this needs to be closed as discussed here https://github.com/containers/common/issues/555

rhatdan: My point is the directory where the secret is stored and all of the paths up to it should be (root:root) 0755 so that all users within the container can search through the directories to get to the secret file.  The file with the secrets should be 0444 so all users of the container can read it.

I guess you could argue that the secret `file` should be owned by the default user and be 0600, but I am fine with making it readable by all users.

I see the secret file being something that can be used by processes within the container, but not something that should be saved to the image if I was building a container image.

flouthoc: @rhatdan Ah my bad i misunderstood few things. I agree `0755` for `secrets` dir makes sense. However actual  file underneath should be `0444`. I have done the required changes for mounted `run/secrets` dir.

But as discussed containers/common#555 this PR might overlap a bit with what @ashley-cui is already doing. Not sure what needs to be done.  But this PR would definitely fix  issue being faced in containers/common#555

ashley-cui: If you don't mind, I'd be in favor of closing this PR and using the one I have coming up, since I'll have to move all these changes anyway

rhatdan: No problem.  @flouthoc Thanks for the effort.  Please work with @ashley-cui and her PR since she is in charge of this issue.  And make sure it matches your expectations.

ashley-cui: @flouthoc Thanks for giving this a shot though!! I'll @ you when I open my PR do you can take a look :)

flouthoc: @ashley-cui @rhatdan  Sure no issues. 😄

PR #10356, opened 2021-05-16
podman network reload add rootless support

Luap99:
Allow podman network reload to be run as rootless user. While it is
unlikely that the iptable rules are flushed inside the rootless cni
namespace, it could still happen. Also fix podman network reload --all
to ignore errors when a container does not have the bridge network mode,
e.g. slirp4netns.


mheon:
LGTM

rhatdan (approved):
/lgtm

PR #10359, opened 2021-05-16
Update troubleshooting.md

infiniteregrets:
I have been experimenting with this idea of nested containers, and I faced some problems/errors while doing so that really confused me (probably because I was new to some concepts) for a while, but I think what I have added, although trivial, can be helpful to someone. If I made any errors or wrote something that is incorrect then please do let me know (: 
Thank you!

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -342,7 +342,7 @@ or `docker logout` to remove the credentials from the authentication file.
 
 ### 13) Running Podman inside a container causes container crashes and inconsistent states
 
-Running Podman in a container and forwarding some, but not all, of the required host directories can cause inconsistent container behavior.
+- Running Podman in a container and forwarding some, but not all, of the required host directories can cause inconsistent container behavior.
 
 #### Symptom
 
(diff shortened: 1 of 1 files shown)


troubleshooting.md
```diff
@@ -356,7 +356,44 @@ If you do mount in the host's `/var/lib/containers/storage`, however, you must a
 Not doing this will cause Podman in the container to detect that temporary files have been cleared, leading it to assume a system restart has taken place.
 This can cause Podman to reset container states and lose track of running containers.
 
-For running containers on the host from inside a container, we also recommend the [Podman remote client](remote_client.md), which only requires a single socket to be mounted into the container.
+For running containers on the host from inside a container, we also recommend the [Podman remote client](remote_client.md), which only requires a single socket to be mounted into the container.  
+
+
+- Running Podman containers inside a Docker container
```
rhatdan, line 362: There is no such thing as a Docker container
Running Podman containers inside of a container launched by Docker.

I would rather not get this specific with this particular use case.  We are planning and writing a large document on running Podman within a container and this will cover lots of different use cases including podman in podman, podman in docker, podman in kubernetes.  Once this document is released, we can add specific use cases.

Then we can just point this troubleshooting document at that document. 

We plan on releasing the document with podman 3.2 release.

infiniteregrets:
> Running Podman containers inside of a container launched by Docker.
My bad, thanks for correcting me, sometimes it's just comfortable to say a "Docker Container" so I just went with it. But yeah, that would be incorrect. 

> Then we can just point this troubleshooting document at that document.
Thanks! 
 
> We plan on releasing the document with podman 3.2 release.
That would be really nice!

troubleshooting.md
TomSweeneyRedHat: I'll echo @rhatdan's thinking.  We try to keep the solutions here to be a few sentences, I would prefer your solution here to be a quick explanation and then a link to another doc.  I'll also take Dan's thought a step further and say there's no "Podman container" either.  So perhaps:
`Running Podman in a container controlled by Docker`  or perhaps `being run by Docker`

infiniteregrets: Thanks for clarifying!

vrothberg:
I am going to close this PR as the podman-in-* doc will be released soon.

Issue #10350, opened 2021-05-16
podman-3.2.0_rc1: `podman image prune` and `podman system prune` now remove all images


srcshelton:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

With `podman-3.1.2` and below, `podman image prune` or `podman system prune` would only remove untagged images.

With `podman-3.2.0_rc1`, the same commands remove **all** images not linked to a running container.

This is unhelpful as it clears libraries of cached (and correctly tagged) images.  If this was an intentional change rather than a bug, then it would appear preferable to introduce it as a new option for the `prune` sub-commands, as the current situation is bound to lead to unintended data-loss.

I also note in passing that this may be related to a lack of clarity in the command help text: the `podman image prune` help text reads:

```
$ podman image prune --help
Remove unused images

Description:
  Removes all unnamed images from local storage.

  If an image is not being used by a container, it will be removed from the system.
```

… which appears unclear - which is the intended behaviour: to remove unnamed images, or to remove _all_ images 
not being used by an active container?

The `podman system prune` help text is even less clear:

```
$ podman system prune --help
Remove unused data

Description:
  
        podman system prune

        Remove unused data
```

… which doesn't specify what is classified as 'unused', or what items will be considered.

**Steps to reproduce the issue:**

1. Pull some images

2. `podman image prune -f`

3. `podman image ls`

**Describe the results you received:**

All images removed.

**Describe the results you expected:**

Untagged images removed, but previously pulled images retained (as per pre-`3.2.0_rc1` behaviour).

**Output of `podman version`:**

```
Version:      3.2.0-rc1
API Version:  3.2.0-rc1
Go Version:   go1.16.2
Git Commit:   cb30558bdf8c8f0481c4f05774105fffe0db597a
Built:        Thu May  6 10:55:51 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.2-dev
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: app-emulation/conmon-2.0.27
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: 65fad4bfcb250df0435ea668017e643e7f462155'
  cpus: 8
  distribution:
    distribution: gentoo
    version: unknown
  eventLogger: file
  hostname: dellr330
  idMappings:
    gidmap: null
    uidmap: null
[...]

PR #10363, opened 2021-05-17
image prune: remove unused images only with `--all`

vrothberg:
Fix a regression in `podman image prune` where unused images were
accidentally removed even when `--all=false`.  Extend and partially
rewrite the e2e tests to make sure we're not regressing again in the
future.

Fixing the aforementioned issue revealed another issue in the default
prune filter.  While prune should remove all "dangling" images (i.e.,
those without tag), it removed only "intermediate" ones; dangling images
without children.  Remove the mistaken comment from the libimage
migration.

Also clarify the help message and man page.

Fixes: #10350

### DIFF
--- a/cmd/podman/images/prune.go
+++ b/cmd/podman/images/prune.go
@@ -15,10 +15,8 @@ import (
 )
 
 var (
-	pruneDescription = `Removes all unnamed images from local storage.
-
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

Issue #10210, opened 2021-05-04
Secrets should support UID, GID, mode options


ashley-cui:
/kind feature

**Description**
docker supports:
--secret source=app-key,target=app,uid=1000,gid=1001,mode=0400 
so we should too

**Output of `podman version`:**

Version:      3.2.0-dev
API Version:  3.2.0-dev
Go Version:   go1.15.8
Git Commit:   476c76f580d5cd092ff958765af36857b2a68d6c-dirty
Built:        Tue May  4 15:37:55 2021
OS/Arch:      linux/amd64

Issue #10364, opened 2021-05-15
rootless: secret not accessible to non root


sbourlon:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Hello podman team,

My container cannot access secrets when running as non root.

**Steps to reproduce the issue:**

```sh
#!/bin/sh
set -o xtrace

SECRET="mysecret"
echo a_secret | podman secret create "$SECRET" -
podman run --secret "$SECRET" --rm docker.io/library/debian sh -c "id && cat /run/secrets/$SECRET"
podman run --secret "$SECRET" --rm --user 200:200 docker.io/library/debian sh -c "id && cat /run/secrets/$SECRET"
podman secret rm "$SECRET"
```

**Describe the results you received:**

The user 200:200 received the error `cat: /run/secrets/mysecret: Permission denied` when accessing the secret.

```
+ SECRET=mysecret
+ echo a_secret
+ podman secret create mysecret -
200255d821a5ef5b0a3e8b7b6
+ podman run --secret mysecret --rm docker.io/library/debian sh -c 'id && cat /run/secrets/mysecret'
uid=0(root) gid=0(root) groups=0(root)
a_secret
+ podman run --secret mysecret --rm --user 200:200 docker.io/library/debian sh -c 'id && cat /run/secrets/mysecret'
uid=200(200) gid=200(200) groups=200(200)
cat: /run/secrets/mysecret: Permission denied
+ podman secret rm mysecret
200255d821a5ef5b0a3e8b7b6
```

**Describe the results you expected:**

The user 200:200 should access the secret.

```
+ SECRET=mysecret
+ echo a_secret
+ podman secret create mysecret -
200255d821a5ef5b0a3e8b7b6
+ podman run --secret mysecret --rm docker.io/library/debian sh -c 'id && cat /run/secrets/mysecret'
uid=0(root) gid=0(root) groups=0(root)
a_secret
+ podman run --secret mysecret --rm --user 200:200 docker.io/library/debian sh -c 'id && cat /run/secrets/mysecret'
uid=200(200) gid=200(200) groups=200(200)
a_secret
+ podman secret rm mysecret
200255d821a5ef5b0a3e8b7b6
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16.3
[...]

PR #10366, opened 2021-05-17
Support uid,gid,mode options for secrets

ashley-cui:
Support UID, GID, Mode options for mount type secrets. Also, change
default secret permissions to 444 so all users can read secret.

Fixes: https://github.com/containers/podman/issues/10210, https://github.com/containers/podman/issues/10364

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -778,20 +778,30 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
 	return td, nil
 }
 
-func parseSecrets(secrets []string) ([]string, map[string]string, error) {
+func parseSecrets(secrets []string) ([]specgen.Secret, map[string]string, error) {
 	secretParseError := errors.New("error parsing secret")
-	var mount []string
+	var mount []specgen.Secret
 	envs := make(map[string]string)
 	for _, val := range secrets {
+		// mount only tells if user has set an option that can only be used with mount secret type
+		mountOnly := false
 		source := ""
 		secretType := ""
 		target := ""
+		var uid, gid uint32
+		// default mode 444 octal = 292 decimal
+		var mode uint32 = 292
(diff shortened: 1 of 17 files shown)


cmd/podman/common/specgen.go
```diff
@@ -778,28 +778,38 @@ func parseThrottleIOPsDevices(iopsDevices []string) (map[string]specs.LinuxThrot
...
 	envs := make(map[string]string)
 	for _, val := range secrets {
+		// mount only tells if user has set an option that can only be used with mount secret type
+		mountOnly := false
 		source := ""
 		secretType := ""
 		target := ""
+		var uid, gid uint32
+		// default mode 444 octal = 292 decimal
+		var mode uint32 = 292
```
ashley-cui, line 793: @flouthoc, @rhatdan  this line should solve the permissions issue, it defaults the secret to 444, which is what docker does. The default UID and GID for a secret inside a container is still 0 though, even if the container is run using --user, because that's what docker does, but it still allows other users to read the secret. All paths leading up to the secret are 755.  Let me know if this looks okay/

rhatdan: This should be in Common, since buildah will also need this correct?

ashley-cui: Buildah already has something like this, though I would argue it shouldn't be in the current common, since the secrets package in common only understands how to store secrets, and should be agnostic to how the secrets are used. I'm open to possibly have a secretsUtil package under the secrets package in common for functions like this parse function and a copy secret to dir function, but I think @vrothberg has a strong opinion on keeping common secrets usage-agnostic.

rhatdan: Ok, I would say we should move these helper functions then.

rhatdan:
LGTM

rhatdan:
@flouthoc @vrothberg @containers/podman-maintainers PTAL

flouthoc:
LGTM

PR #10370, opened 2021-05-17
Add script for identifying commits in release branches

mheon:
One of the worst parts of a Podman release is writing the release notes. It requires manually going through all merged commits since the last release, figuring out what was actually done, and writing a small blurb about what was fixed. The worst part of this is the difficulty in finding the commits that were actually included in previous releases - our extensive backports to prior releases mean that there are usually dozens of commits that were included in a prior release, but do not have a matching SHA (as the original author did not do the backport, and often the commit required massaging to cherry-pick in).

This script automates the job of finding commits in one release branch that are not in another, with filtering to remove most cherry-picked commits. It makes my life a lot easier during releases, so I figured I'd include it in hack/ so anyone else stuck with the enjoyable task of writing release notes can have a slightly easier life.

The script is written in absolutely terrible Ruby and its performance is absolutely terrible, but you only need to run it once per major release and a 30-second wait to generate the list of commits to include isn't bad.

### DIFF
--- a/hack/branch_commits.rb
+++ b/hack/branch_commits.rb
@@ -1,0 +1,98 @@
+#!/usr/bin/ruby
+
+require 'set'
+
+# Get commits in one branch, but not in another, accounting for cherry-picks.
+# Accepts two arguments: base branch and old branch. Commits in base branch that
+# are not in old branch will be reported.
+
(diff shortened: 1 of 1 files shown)


edsantiago:
LGTM

PR #10371, opened 2021-05-17
fix: response of containers wait endpoint

matejvasek:
The `Error` part of response must be nil if no error occurred.
Before this commit a zero value for the struct was returned.

### DIFF
--- a/pkg/api/handlers/types.go
+++ b/pkg/api/handlers/types.go
@@ -98,7 +98,7 @@ type BuildResult struct {
 
 type ContainerWaitOKBody struct {
 	StatusCode int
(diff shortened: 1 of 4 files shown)


matejvasek:
> The Error part of response must be nil if no error occurred.

It can be also omitted, but I don't know how to assert that.

matejvasek:
@mheon @jwhonce PTAL

TomSweeneyRedHat:
Changes LGTM, but it looks like you need a rebase @matejvasek

matejvasek:
@TomSweeneyRedHat rebased

Issue #10338, opened 2021-05-14
Add ability to publish other ports in container restore


new-sashok724:
/kind feature

Hello! We use podman checkpoint/restore to speed up the launch of our TCP application in multiple instances.
We run container with application, then when it launched checkpoint it to tar, and then restore it from tar with different names.

**Steps to reproduce the issue:**
1. `podman run -d -p 10000:$port $image`
2. Wait until it started
3. `podman container checkpoint -R -e $container.tar.gz $container`
4. `podman container restore --import $container.tar.gz -n $container2`

**Describe the results you received:**
Container restore fails because published port is already used by original container.
I have to manually modify .tar.gz archive to change the port, which even if automated will be longer than just starting new instance without checkpoint.

Also it would be good to have ability to checkpoint just to uncompressed tar archive, or to directory, and maybe just to publish port after container is created.

**Describe the results you expected:**
There's argument in restore command to publish other ports (-p)

**Output of `podman version`:**
podman version 3.1.2

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide? (https://github.com/containers/podman/blob/master/troubleshooting.md)**
Yes

rhatdan:
@adrianreber @mheon would this even be possible?

new-sashok724:
It's already possible by modifying config.dump in tar archive. It's just too inconvenient

eriksjolund:
Just an idea, I don't know how realistic it is:

Some imaginary command-line options

* --export-to-stdout  write tar stream to stdout (uncompressed)
* --import-from-stdin read tar stream from stdout (uncompressed)

```
podman container checkpoint -R --export-to-stdout $container | custom_script_to_adjust_port_number | podman container restore --import-from-stdin -n $container2
```
(Note the command-line options `--export-to-stdout` and `--import-from-stdin` do not exist as of now).


I would guess the performance would be less of a problem, because the adjustment of the port number could be done without  the extra decompress and compress steps.

new-sashok724:
I think option to import/export to/from directory would be even more convenient

adrianreber:
> It's already possible by modifying config.dump in tar archive. It's just too inconvenient

If it is just a change in `config.dump` then it would work more or less like the `--name` option. To be honest I never tested port forwarding when implementing checkpoint/restore. I just tested it now and I see that it works. Cool. I think it should be possible to add the `-p` as during container creation also to restore. I will try to implement something and open a pull request next week.

[...]

PR #10381, opened 2021-05-18
Add --publish to container restore

adrianreber:
Fixes: #10338

During container restore the published ports were recreated the same they were during initial container creation. With the changes from this PR it is now possible to change the published ports during restore.

It was already possible to change the published ports during restore by editing the files in the container checkpoint archive. This PR adds the command-line option `--publish` to `restore` with the same semantics as `--publish` during `run`.

### DIFF
--- a/cmd/podman/containers/restore.go
+++ b/cmd/podman/containers/restore.go
@@ -36,9 +36,7 @@ var (
 	}
 )
 
-var (
-	restoreOptions entities.RestoreOptions
-)
+var restoreOptions entities.RestoreOptions
 
 func init() {
 	registry.Commands = append(registry.Commands, registry.CliCommand{
@@ -66,10 +64,17 @@ func init() {
 	flags.BoolVar(&restoreOptions.IgnoreStaticIP, "ignore-static-ip", false, "Ignore IP address set via --static-ip")
 	flags.BoolVar(&restoreOptions.IgnoreStaticMAC, "ignore-static-mac", false, "Ignore MAC address set via --mac-address")
 	flags.BoolVar(&restoreOptions.IgnoreVolumes, "ignore-volumes", false, "Do not export volumes associated with container")
+
+	flags.StringSliceP(
+		"publish", "p", []string{},
+		"Publish a container's port, or a range of ports, to the host (default [])",
+	)
+	_ = restoreCommand.RegisterFlagCompletionFunc("publish", completion.AutocompleteNone)
+
 	validate.AddLatestFlag(restoreCommand, &restoreOptions.Latest)
 }
 
-func restore(_ *cobra.Command, args []string) error {
+func restore(cmd *cobra.Command, args []string) error {
 	var errs utils.OutputErrors
 	if rootless.IsRootless() {
 		return errors.New("restoring a container requires root")
@@ -88,6 +93,17 @@ func restore(_ *cobra.Command, args []string) error {
 	}
 	if restoreOptions.Name != "" && restoreOptions.TCPEstablished {
 		return errors.Errorf("--tcp-established cannot be used with --name")
+	}
+
+	inputPorts, err := cmd.Flags().GetStringSlice("publish")
+	if err != nil {
+		return err
+	}
+	if len(inputPorts) > 0 {
+		restoreOptions.PublishPorts, err = common.CreatePortBindings(inputPorts)
+		if err != nil {
+			return err
+		}
 	}
 
 	argLen := len(args)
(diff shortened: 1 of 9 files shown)


docs/source/markdown/podman-container-restore.1.md
```diff
@@ -95,6 +95,14 @@ This option must be used in combination with the **--import, -i** option.
 When restoring containers from a checkpoint tar.gz file with this option,
 the content of associated volumes will not be restored.
 
+#### **--publish**, **-p**
+
+Replaces the ports that the container publishes, as configured during initial
```
TomSweeneyRedHat, line 100:
```suggestion
Replaces the ports that the container publishes, as configured during the initial
```

docs/source/markdown/podman-container-restore.1.md
```diff
@@ -95,6 +95,14 @@ This option must be used in combination with the **--import, -i** option.
...
+container start, with a new set of port forwarding rules.
+
+For more details please see **podman run --publish**.
+
+
```
TomSweeneyRedHat, line 105: it would be good to add an example, and we probably should add `podman-run(1)` to the See Also list below

rhatdan:
LGTM Other then @TomSweeneyRedHat fixups.

TomSweeneyRedHat:
I should have noted earlier, other than the man page nits, LGTM.

docs/source/markdown/podman-container-restore.1.md
adrianreber: Changed.

adrianreber:
I applied both man page suggestions.

TomSweeneyRedHat:
I love our CI at times, doc nits cleaned up, now the CI is barfing....  Looks like a network hiccup, will try restarting.

cmd/podman/containers/restore.go
```diff
@@ -67,10 +65,18 @@ func init() {
 	flags.BoolVar(&restoreOptions.IgnoreStaticIP, "ignore-static-ip", false, "Ignore IP address set via --static-ip")
 	flags.BoolVar(&restoreOptions.IgnoreStaticMAC, "ignore-static-mac", false, "Ignore MAC address set via --mac-address")
 	flags.BoolVar(&restoreOptions.IgnoreVolumes, "ignore-volumes", false, "Do not export volumes associated with container")
+
+	publishFlagName := "publish"
```
mheon, line 69: Nit: if you make any other changes, there's no reason to keep this - you can remove the variable and replace `publishFlagName` in the `StringSliceP` with "publish"

pkg/checkpoint/checkpoint_restore.go
```diff
@@ -95,6 +96,23 @@ func CRImportCheckpoint(ctx context.Context, runtime *libpod.Runtime, restoreOpt
 		newName = true
 	}
 
+	if len(restoreOptions.PublishPorts) > 0 {
+		ctrConfig.PortMappings = []ocicni.PortMapping{}
+		for _, pp := range restoreOptions.PublishPorts {
+			mapping := ocicni.PortMapping{
```
mheon, line 102: This actually needs to be a loop - a `specgen.PortMapping` represents a range of ports (the `Range` field), and we need to generate one `ocicni.PortMapping` for each.

mheon: You may want to use https://github.com/containers/podman/blob/master/pkg/specgen/generate/ports.go#L27 (it's in a problematic package right now - would introduce an import loop - but I don't see why it can't be moved to `/pkg/specgen` and that ought to be OK.

adrianreber: Thanks for the pointer to existing code. I was grepping a lot but I was not able to find it.

About 'This actually needs to be a loop'. It is a loop. At least I hope it is. Line 101 has `for _, pp := range restoreOptions.PublishPorts {` That should loop over all port mappings, right?

cmd/podman/containers/restore.go
adrianreber: It is also used in line 75 `restoreCommand.RegisterFlagCompletionFunc(publishFlagName`. This also a copy from `podman run` or create. No problem removing the variable.

pkg/checkpoint/checkpoint_restore.go
mheon: My original thought with the "needs to be a loop" comment was you needed a nested loop - loop over `PublishedPorts` and then loop over each port's `Range` to generate a number of ocicni PortMapping structs for each port range in PublishedPorts. The function I linked should remove that need and do the work itself.

adrianreber: Switched to the existing function. Thanks for the pointer.

cmd/podman/containers/restore.go
adrianreber: removed the `publish` String

adrianreber:
I addressed the review comments and CI seems happy.

rhatdan:
@mheon PTAL and Merge.

Issue #9609, opened 2021-03-04
Podman play kube ignores USER (rootless)


ghost:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The `podman play kube` command does not use the user specified in the Dockerfile

**Steps to reproduce the issue:**

1. Build simple image with USER step:
```
FROM docker.io/centos

RUN useradd -c 'testuser' -m -p '!' -u 10000 -s '/bin/bash' testuser

USER testuser

CMD tail -f /dev/null
```

2. Write yaml file that uses this image:
```
---

apiVersion: v1
kind: Pod
metadata:
  labels:
    app: pod
  name: pod
spec:
  containers:
  - name: nginx
    image: localhost/testimage
    env:
     - name: SOME_ENV
       value: "VALUE"
    ports:
    - containerPort: 111111
      hostPort: 11111
      protocol: TCP
    volumeMounts:
    - mountPath: /home/podman/testdir
      name:  testdir

  volumes:
  - name: testdir
    hostPath:
      path: /home/podman/testdir
      type: Directory
```

3. Play the file: `podman play kube test.yaml`
4. Check who is the user in the running container:
```
podman exec pod-nginx whoami
root
```

5. Run the same image without `podman play kube`:
```
podman run -ti localhost/testimage whoami
testuser
```

**Describe the results you received:**
The user should be the same ("testuser" in this case) with or without `podman play kube`.

**Describe the results you expected:**
The user is "root" when I use `podman play kube`.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.14.12
Built:        Mon Mar  1 15:31:01 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.26-4.el8.5.1.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.26, commit: c30e47104fbc00a7381abe56ad8d54ab151f3fde'
  cpus: 4
  distribution:
    distribution: '"centos"'
    version: "8"
  eventLogger: journald
  hostname: mymachine
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 795
      size: 1
    - container_id: 1
[...]

PR #10383, opened 2021-05-18
Handle image user and exposed ports in podman play kube

rhatdan:
Currently if a user runs an image with a user specified or
exposed ports with podman play kube, the fields are ignored.

Fixed: https://github.com/containers/podman/issues/9609

### DIFF
--- a/pkg/specgen/generate/kube/kube.go
+++ b/pkg/specgen/generate/kube/kube.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/common/pkg/secrets"
 	ann "github.com/containers/podman/v3/pkg/annotations"
 	"github.com/containers/podman/v3/pkg/specgen"
+	"github.com/containers/podman/v3/pkg/specgen/generate"
 	"github.com/containers/podman/v3/pkg/util"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
Test system is not at all happy with you today @rhatdan

rhatdan:

baude:
LGTM

pkg/specgen/generate/kube/kube.go
```diff
@@ -182,6 +183,21 @@ func ToSpecGen(ctx context.Context, opts *CtrSpecGenOptions) (*specgen.SpecGener
...
+		if s.User == "" {
+			s.User = imageData.Config.User
+		}
+
+		exposed, err := generate.GenExposedPorts(imageData.Config.ExposedPorts)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(s.Expose) == 0 {
```
TomSweeneyRedHat, line 195: should we regenerate `if len(exposed) != len(s.Expose)` too?

pkg/specgen/generate/kube/kube.go
```diff
@@ -182,6 +183,19 @@ func ToSpecGen(ctx context.Context, opts *CtrSpecGenOptions) (*specgen.SpecGener
 		if imageData.Config.WorkingDir != "" {
 			s.WorkDir = imageData.Config.WorkingDir
 		}
+		if s.User == "" {
+			s.User = imageData.Config.User
+		}
```
TomSweeneyRedHat, line 188: Should we really be setting the user to the value found in the image?  That sounds like it might get twisted, or am I misreading this?

PR #10385, opened 2021-05-18
[Techinal Debt][NO TESTS NEEDED] Cleanup ABI vs. Tunnel CLI commands

jwhonce:
This commit cleans up two issues:
* Most commands support all EngineModes so default to that. Let
  outlayers declare their intent.
* Use cobra.Annotations to set supported EngineMode. This simplies
  instantiating commands as there is now one method to communicate a
  commands requirements rather than two.

Additionally:
* Combined aliased commands into one file
* Fixed aliased commands where Args field did not match
* Updated examples in README.md for writing commands


cmd/podman/containers/mount.go
```diff
@@ -25,27 +25,28 @@ var (
 `
 
 	mountCommand = &cobra.Command{
+		Annotations: map[string]string{
+			registry.UnshareNSRequired: "",
+			registry.ParentNSRequired:  "",
+			registry.EngineMode:        registry.ABIMode,
```
mheon, line 31: You sure about this one? I remember losing an argument with @baude that this one should be ABI-only

rhatdan: Yes mount makes no sense on a remote server.

jwhonce: @mheon The command was limited to just ABI support and continues to be so. Not listing an EngineMode means the command supports all modes.

cmd/podman/main.go
```diff
@@ -42,38 +43,41 @@ func main() {
...
-					}
-				} else {
-					_, found = c.Command.Annotations[registry.ParentNSRequired]
-					if rootless.IsRootless() && found {
-						c.Command.RunE = func(cmd *cobra.Command, args []string) error {
-							return fmt.Errorf("cannot run command %q in rootless mode", cmd.CommandPath())
-						}
-					}
+		if supported, found := c.Command.Annotations[registry.EngineMode]; found {
+			if !strings.Contains(cfg.EngineMode.String(), supported) {
```
Luap99, line 47: Why contains? Should this be `cfg.EngineMode.String() != supported`?

cmd/podman/images/list.go
```diff
@@ -49,6 +48,19 @@ var (
...
+		Use:               "images [options] [IMAGE]",
+		Args:              imageListCmd.Args,
+		Short:             imageListCmd.Short,
+		Long:              imageListCmd.Long,
+		RunE:              imageListCmd.RunE,
+		ValidArgsFunction: imageListCmd.ValidArgsFunction,
+		Example: `podman images --format json
+  podman images --sort repository --format "table {{.ID}} {{.Repository}} {{.Tag}}"
+  podman images --filter dangling=true`,
+		DisableFlagsInUseLine: true,
```
Luap99, line 61: DisableFlagsInUseLine can be removed it is set for all commands in main

cmd/podman/main.go
jwhonce: @Luap99 I did contains to support any future engines.

cmd/podman/containers/unmount.go
```diff
@@ -20,11 +20,12 @@ var (
   An unmount can be forced with the --force flag.
 `
 	unmountCommand = &cobra.Command{
-		Use:     "unmount [options] CONTAINER [CONTAINER...]",
-		Aliases: []string{"umount"},
-		Short:   "Unmounts working container's root filesystem",
-		Long:    description,
-		RunE:    unmount,
+		Annotations: map[string]string{registry.EngineMode: registry.ABIMode},
```
rhatdan, line 23: Is this saying unmount is supported in podman-remote?

jwhonce: No, only "local"

rhatdan: What I find confusing is that Annotations indicates which engines to use, now rather then Mode, is that basically the change that is going on?

cmd/podman/images/inspect.go
```diff
@@ -12,6 +12,7 @@ import (
 var (
 	// Command: podman image _inspect_
 	inspectCmd = &cobra.Command{
+
```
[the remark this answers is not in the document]
jwhonce: Global sed bit me...

TomSweeneyRedHat:
LGTM
and that's some serious sandbox scrubbing @jwhonce !
All happy green test buttons and a shiny merge button with another LGTM

Issue #10306, opened 2021-05-11
podman info - print available cgroup controllers


aojea:
Opening a new issue with from:

_Originally posted by @afbjorklund in https://github.com/containers/podman/issues/10289#issuecomment-837840788_

> If i understand correctly, they had some different original requirements anyway:
> 
> ```
> 		// Unlike `docker info`, `podman info` does not print available cgroup controllers.
> 		// So we parse "cgroup.subtree_control" file by ourselves.
> ```
> 
> So maybe they were more interested in `cgroupVersion`, and some new fields ?
> 
> ```
> 	MemoryLimit     bool     `json:"MemoryLimit"`
> 	PidsLimit       bool     `json:"PidsLimit"`
> 	CPUShares       bool     `json:"CPUShares"`
> ```
> 
> https://github.com/kubernetes-sigs/kind/issues/2233
>

afbjorklund:
The complete list is available at github.com/docker/docker/pkg/sysinfo

        v.MemoryLimit = sysInfo.MemoryLimit
        v.SwapLimit = sysInfo.SwapLimit
        v.KernelMemory = sysInfo.KernelMemory
        v.KernelMemoryTCP = sysInfo.KernelMemoryTCP
        v.OomKillDisable = sysInfo.OomKillDisable
        v.CPUCfsPeriod = sysInfo.CPUCfsPeriod
        v.CPUCfsQuota = sysInfo.CPUCfsQuota
        v.CPUShares = sysInfo.CPUShares
        v.CPUSet = sysInfo.Cpuset
        v.PidsLimit = sysInfo.PidsLimit

rhatdan:
Care to open a PR to fix?

rhatdan:
Do you want podman info to report all of this information?

afbjorklund:
> Do you want podman info to report all of this information?

The request from `kind` was for three new fields.  Maybe add swap too ?

```go
// dockerInfo corresponds to `docker info --format '{{json .}}'`
type dockerInfo struct {
        CgroupDriver    string   `json:"CgroupDriver"`  // "systemd", "cgroupfs", "none"
        CgroupVersion   string   `json:"CgroupVersion"` // e.g. "2"
        MemoryLimit     bool     `json:"MemoryLimit"`
        PidsLimit       bool     `json:"PidsLimit"`
        CPUShares       bool     `json:"CPUShares"`
        SecurityOptions []string `json:"SecurityOptions"`
}
```

[...]

PR #10387, opened 2021-05-18
Podman info add support for status of standard available cgroup controllers

flouthoc:
Following PR adds support for reflecting availability status of cgroup controllers to `info`.  Patch attempts to resolve https://github.com/containers/podman/issues/10306

### DIFF
--- a/libpod/define/info.go
+++ b/libpod/define/info.go
@@ -23,21 +23,22 @@ type SecurityInfo struct {
 
 // HostInfo describes the libpod host
 type HostInfo struct {
-	Arch           string           `json:"arch"`
-	BuildahVersion string           `json:"buildahVersion"`
-	CgroupManager  string           `json:"cgroupManager"`
-	CGroupsVersion string           `json:"cgroupVersion"`
-	Conmon         *ConmonInfo      `json:"conmon"`
-	CPUs           int              `json:"cpus"`
-	Distribution   DistributionInfo `json:"distribution"`
-	EventLogger    string           `json:"eventLogger"`
-	Hostname       string           `json:"hostname"`
-	IDMappings     IDMappings       `json:"idMappings,omitempty"`
-	Kernel         string           `json:"kernel"`
-	MemFree        int64            `json:"memFree"`
-	MemTotal       int64            `json:"memTotal"`
-	OCIRuntime     *OCIRuntimeInfo  `json:"ociRuntime"`
-	OS             string           `json:"os"`
+	Arch              string           `json:"arch"`
+	BuildahVersion    string           `json:"buildahVersion"`
+	CgroupManager     string           `json:"cgroupManager"`
+	CGroupsVersion    string           `json:"cgroupVersion"`
+	CgroupControllers []string         `json:"cgroupControllers"`
+	Conmon            *ConmonInfo      `json:"conmon"`
+	CPUs              int              `json:"cpus"`
+	Distribution      DistributionInfo `json:"distribution"`
+	EventLogger       string           `json:"eventLogger"`
+	Hostname          string           `json:"hostname"`
+	IDMappings        IDMappings       `json:"idMappings,omitempty"`
+	Kernel            string           `json:"kernel"`
+	MemFree           int64            `json:"memFree"`
+	MemTotal          int64            `json:"memTotal"`
+	OCIRuntime        *OCIRuntimeInfo  `json:"ociRuntime"`
+	OS                string           `json:"os"`
 	// RemoteSocket returns the UNIX domain socket the Podman service is listening on
 	RemoteSocket *RemoteSocket          `json:"remoteSocket,omitempty"`
 	RuntimeInfo  map[string]interface{} `json:"runtimeInfo,omitempty"`
--- a/libpod/info.go
+++ b/libpod/info.go
@@ -93,20 +93,33 @@ func (r *Runtime) hostInfo() (*define.HostInfo, error) {
 		return nil, errors.Wrapf(err, "error getting Seccomp profile path")
 	}
 
+	// CGroups version
+	unified, err := cgroups.IsCgroup2UnifiedMode()
+	if err != nil {
+		return nil, errors.Wrapf(err, "error reading cgroups mode")
+	}
+
+	// Get Map of all available controllers
+	availableControllers, err := cgroups.GetAvailableControllers(nil, unified)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting available cgroup controllers")
+	}
+
 	info := define.HostInfo{
-		Arch:           runtime.GOARCH,
-		BuildahVersion: buildah.Version,
-		CgroupManager:  r.config.Engine.CgroupManager,
-		Linkmode:       linkmode.Linkmode(),
-		CPUs:           runtime.NumCPU(),
-		Distribution:   hostDistributionInfo,
-		EventLogger:    r.eventer.String(),
-		Hostname:       host,
-		IDMappings:     define.IDMappings{},
-		Kernel:         kv,
-		MemFree:        mi.MemFree,
-		MemTotal:       mi.MemTotal,
-		OS:             runtime.GOOS,
+		Arch:              runtime.GOARCH,
+		BuildahVersion:    buildah.Version,
+		CgroupManager:     r.config.Engine.CgroupManager,
+		CgroupControllers: availableControllers,
+		Linkmode:          linkmode.Linkmode(),
+		CPUs:              runtime.NumCPU(),
+		Distribution:      hostDistributionInfo,
+		EventLogger:       r.eventer.String(),
+		Hostname:          host,
+		IDMappings:        define.IDMappings{},
+		Kernel:            kv,
+		MemFree:           mi.MemFree,
+		MemTotal:          mi.MemTotal,
+		OS:                runtime.GOOS,
 		Security: define.SecurityInfo{
 			AppArmorEnabled:     apparmor.IsEnabled(),
 			DefaultCapabilities: strings.Join(r.config.Containers.DefaultCapabilities, ","),
@@ -120,11 +133,6 @@ func (r *Runtime) hostInfo() (*define.HostInfo, error) {
 		SwapTotal:   mi.SwapTotal,
 	}
 
-	// CGroups version
-	unified, err := cgroups.IsCgroup2UnifiedMode()
-	if err != nil {
-		return nil, errors.Wrapf(err, "error reading cgroups mode")
-	}
 	cgroupVersion := "v1"
 	if unified {
 		cgroupVersion = "v2"
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -128,26 +128,116 @@ func init() {
 // getAvailableControllers get the available controllers
 func getAvailableControllers(exclude map[string]controllerHandler, cgroup2 bool) ([]controller, error) {
 	if cgroup2 {
-		return nil, fmt.Errorf("getAvailableControllers not implemented yet for cgroup v2")
+		controllers := []controller{}
+		subtreeControl := cgroupRoot + "/cgroup.subtree_control"
+		// rootless cgroupv2: check available controllers for current user ,systemd or servicescope will inherit
+		if rootless.IsRootless() {
+			userSlice, err := getCgroupPathForCurrentProcess()
+			if err != nil {
+				return controllers, err
+			}
+			//userSlice already contains '/' so not adding here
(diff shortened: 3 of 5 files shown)


libpod/define/info.go
```diff
@@ -27,6 +27,9 @@ type HostInfo struct {
 	BuildahVersion string           `json:"buildahVersion"`
 	CgroupManager  string           `json:"cgroupManager"`
 	CGroupsVersion string           `json:"cgroupVersion"`
+	MemoryLimit    bool             `json:"memoryLimit"`
```
rhatdan, line 30: I think this would be better in a subgroup,  CGroupInfo {
     CPUShares...
     MemoryLimit ...
     PidsLimit ...
}

And Alphabetized.

flouthoc: @rhatdan Sure makes sense, much cleaner i'll do the required changes.

pkg/cgroups/cgroups.go
```diff
@@ -128,7 +128,25 @@ func init() {
 // getAvailableControllers get the available controllers
 func getAvailableControllers(exclude map[string]controllerHandler, cgroup2 bool) ([]controller, error) {
 	if cgroup2 {
-		return nil, fmt.Errorf("getAvailableControllers not implemented yet for cgroup v2")
+		controllers := []controller{}
+		// Get caller uid to confirm if its invoked by root or non-root
+		uid := os.Geteuid()
+		subtreeControl := cgroupRoot + "/cgroup.subtree_control"
+		if uid != 0 {
+			subtreeControl = fmt.Sprintf("%s/user.slice/user-%d.slice/user@%d.service/cgroup.subtree_control", cgroupRoot, uid, uid)
```
AkihiroSuda, line 136: Not robust

https://github.com/kubernetes-sigs/kind/issues/2254#issuecomment-842888408

AkihiroSuda (changes requested):

pkg/cgroups/cgroups.go
flouthoc: @AkihiroSuda ah I think i'll get prefix from group manager, that should resolve this. Thanks for the review :) i'll try to fix these.

libpod/info.go
```diff
@@ -93,10 +93,25 @@ func (r *Runtime) hostInfo() (*define.HostInfo, error) {
...
+	availableControllers, err := cgroups.GetAvailableControllers(nil, unified)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting available cgroup controllers")
+	}
+
 	info := define.HostInfo{
 		Arch:           runtime.GOARCH,
 		BuildahVersion: buildah.Version,
 		CgroupManager:  r.config.Engine.CgroupManager,
+		MemoryLimit:    availableControllers["memory"],
```
AkihiroSuda, line 112: Wondering it may make more sense to just return `CgroupControllers []string` here.

(Docker-compatible REST API should emulate Docker REST API, though)

flouthoc: @AkihiroSuda `docker info` spits boolean for separate controllers so i tried keeping it same. But i guess a common string for all available controllers would be better. WDYT ?

pkg/cgroups/cgroups.go
giuseppe: can't we read those from the current cgroup so we don't have to worry about systemd/cgroupfs or rootless/rootful?

giuseppe: probably for root it still makes sense to read them from the root cgroup

libpod/info.go
AkihiroSuda: Yes, for Podman CLI, returning the common string slice SGTM.
Docker-compatible REST API should retain Docker-compatible booleans.

pkg/cgroups/cgroups.go
flouthoc: @giuseppe Yeah I think for rootless i could just get available controllers from current cgroup. That should be more standard.

libpod/info.go
flouthoc: @AkihiroSuda I tried this but dumping all the available controllers is making `info` too noisy, for instance on my machine even single liner has too much content. I guess docker only prints standard controllers maybe because this is too verbose but i am cool with this.
On my machine `info` shows this
```bash
cgroupControllers:
  - blkio
  - cpu
  - cpu,cpuacct
  - cpuacct
  - cpuset
  - devices
  - freezer
  - hugetlb
  - memory
  - net_cls
  - net_cls,net_prio
  - net_prio
  - perf_event
  - pids
  - rdma
  - systemd
  - unified
```

libpod/define/info.go
flouthoc: @rhatdan This is resolved. But format might change as there is a discussion going below for `CgroupControllers []string`

pkg/cgroups/cgroups.go
flouthoc: @AkihiroSuda @giuseppe PTAL for `rootless cgroupv2` now we are marking the controller availability as `false` if it got failed during `initialization` of cgroup. This seems to be more simple and wouldn't require handling caveats of systemd/cgroupfs unit name.

rhatdan:
Please update the man page with the new output.

pkg/cgroups/cgroups.go
flouthoc: @AkihiroSuda @giuseppe PTAL made some changes I think current checking available controller for current user is better. `Service` and `scope` would anyways inherit this. Here is the output for `rootless` / `cgroupv2` from my machine.

```bash
host:
  arch: amd64
  buildahVersion: 1.20.2-dev
  cgroupInfo:
    cpuShares: false
    memoryLimit: true
    pidsLimit: true
  cgroupManager: systemd
  cgroupVersion: v2
```

flouthoc:
@rhatdan Updated Man pages.

libpod/info.go
AkihiroSuda: For cgroup v1, the right way is to parse /proc/cgroups

flouthoc: @AkihiroSuda Would still give the same output as i have all controllers enabled on v1 
```bash
$ cat /proc/cgroups
#subsys_name	hierarchy	num_cgroups	enabled
cpuset	0	230	1
cpu	0	230	1
cpuacct	0	230	1
blkio	0	230	1
memory	0	230	1
devices	0	230	1
freezer	0	230	1
net_cls	0	230	1
perf_event	0	230	1
net_prio	0	230	1
hugetlb	0	230	1
pids	0	230	1
rdma	0	230	1
```

AkihiroSuda: not "same" :)

flouthoc: @AkihiroSuda ah i get it. So i guess logic for existing `getAvailableControllers` https://github.com/containers/podman/blob/master/pkg/cgroups/cgroups.go#L129 for cgroupv1 needs to be changed or i can add a new function but dont think redundant code for similar task is a good idea. @rhatdan @giuseppe @AkihiroSuda  If we can mutually agree upon having `CgroupControllers []string` in `info` then i think we can proceed with removing old logic in `master` and use `/proc/cgroups` for cgroupv1

pkg/cgroups/cgroups.go
```diff
@@ -123,12 +124,38 @@ func init() {
...
 // getAvailableControllers get the available controllers
 func getAvailableControllers(exclude map[string]controllerHandler, cgroup2 bool) ([]controller, error) {
 	if cgroup2 {
-		return nil, fmt.Errorf("getAvailableControllers not implemented yet for cgroup v2")
+		controllers := []controller{}
+		subtreeControl := cgroupRoot + "/cgroup.subtree_control"
+		// rootless cgroupv2: check available controllers for current user , systemd or servicescope will inherit
+		if rootless.IsRootless() {
+			uid := rootless.GetRootlessUID()
+			subtreeControl = fmt.Sprintf("%s/user.slice/user-%d.slice/cgroup.subtree_control", cgroupRoot, uid)
```
AkihiroSuda, line 145: IIUC the controllers there are not always available for rootless users.
Reading the cgorup.subtree_control of Podman process itself might be better (it is still not robust, either, but at least better...)

AkihiroSuda: cc @giuseppe @kolyshkin

flouthoc: We cant use subtree created by podman process as it wouldn't exist , if `podman info` is called prior to creating any container.

flouthoc: also i think podman cleans up it entries as soon as running container exits. Not sure.

giuseppe: yes the problem is that some controllers are not enabled for the rootless user (and we don't really know what controllers are enabled until we do the equivalent of `systemd-run --scope --user` with systemd).  So I agree with @AkihiroSuda that reading the current enabled controllers is a best effort way of doing it, which should be fine in most cases

libpod/info.go
giuseppe: cgroupv1 is not really supported for rootless, even if manually chowned as the delegation is unsafe.

For root, we can parse `/proc/cgroups` or `/proc/self/cgroups` but for rootless we can just force every controller to be disabled.

pkg/cgroups/cgroups.go
flouthoc: @giuseppe @AkihiroSuda This is resolved in latest commit.

libpod/info.go
flouthoc: @giuseppe @AkihiroSuda Thank a lot . I have added changes, this is resolved in latest commit

libpod/define/info.go
flouthoc: @rhatdan Made some changes in latest commit , added `CgroupControllers []string` instead. Could you please review that.

flouthoc:
@giuseppe @AkihiroSuda @rhatdan I have addressed all the comments in latest commit. Could you guys please take a look.

flouthoc:
Missing test is a flake i think.

pkg/cgroups/cgroups.go
flouthoc: @giuseppe @AkihiroSuda Thank a lot . I have added changes, this is resolved in latest commit

flouthoc:
It was a flake force push solved it. 😃

AkihiroSuda (approved):

flouthoc:
yay!!! thanks @AkihiroSuda  waiting for @giuseppe and @rhatdan 's approval.

pkg/cgroups/cgroups.go
```diff
@@ -128,28 +128,118 @@ func init() {
...
+			if err != nil {
+				return controllers, err
+			}
+			//userSlice already contains '/' so not adding here
+			basePath := cgroupRoot + userSlice
+			subtreeControl = fmt.Sprintf("%s/cgroup.subtree_control", basePath)
+		}
+		subtreeControlBytes, err := ioutil.ReadFile(subtreeControl)
+		if err != nil {
+			return nil, fmt.Errorf("failed while reading controllers for cgroup v2 from %q: %+v", subtreeControl, err)
```
giuseppe, line 145: let's use `errors.Wrapf()` here

pkg/cgroups/cgroups.go
```diff
@@ -128,28 +128,118 @@ func init() {
 // getAvailableControllers get the available controllers
 func getAvailableControllers(exclude map[string]controllerHandler, cgroup2 bool) ([]controller, error) {
 	if cgroup2 {
-		return nil, fmt.Errorf("getAvailableControllers not implemented yet for cgroup v2")
+		controllers := []controller{}
+		subtreeControl := cgroupRoot + "/cgroup.subtree_control"
+		// rootless cgroupv2: check available controllers for current user , systemd or servicescope will inherit
```
giuseppe, line 133: nit: extra space before "systemd"

mheon:
Restarted a single flake.
/approve
LGTM

flouthoc:
@mheon Thanks 😄

flouthoc:
@mheon How do we restart single flake ? There is still one check which has a flake.

pkg/cgroups/cgroups.go
flouthoc: Resolved in latest commit.

pkg/cgroups/cgroups.go
flouthoc: Resolved in latest commit.

giuseppe (approved):
LGTM

mheon:
/lgtm

@flouthoc I click the "Details" link on the test in question. The Github summary page it takes you to should have a "Retest" button if the test is red.

Issue #10393, opened 2021-05-18
[additionalimagestores] Simplify `--storage-opt`/`--root` syntax


srcshelton:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Currently, the `--root` option appears to have little functionality without also specifying `--storage-opt=` to clear the existing storage configuration (which includes the default root).

If use of `--root` implied `--storage-opt= --root` then `podman`'s syntax would be clearer.

Note, however, that `podman` currently doesn't accept, e.g.

```
--storage-opt= --storage-opt="overlay.imagestore=<dir2>" --root <dir1>
```

(… as an error is generated), so either this would need to be relaxed or `--root` should only imply `--storage-opt=` if no additional `--storage-opt` value is specified.

A further enhancement would be to provide an option such as `--swap-root <dir>` or `--root-swap <dir>` which would be equivalent to:

```
podman --storage-opt="overlay.imagestore=${GRAPH_ROOT}" --root "${IMAGE_ROOT}"
```

(e.g. operate on a specified `additiopalimagestore` location but retain access to resources in the default `graphRoot` store)

Finally, since multiple `additionalimagestore` locations can be defined, the above syntax would be ideal to represent the case there one of the additional stores should be promoted to being the main store, whilst retaining access to both the original main store and all other defined additional stores.

See https://github.com/containers/podman/issues/9852 for further context.

PR #10400, opened 2021-05-19
Clear the storage-options from the graphdriver if users specifies --root

rhatdan:
Fixes: https://github.com/containers/podman/issues/10393

Currently if a user specifies a --root flag to override the location of
the container storage, we still enforce the storage-opts from
storage.conf. This causes issues with people trying to intereact with
the additional stores feature, and then forces them to use the obscure
--storage-opt="" option. I belive this should be the default and we
already do this when the user specifies the --storage-driver option.


rhatdan:
@srcshelton PTAL

edsantiago:
That was quick - thank you!

mheon:
Sure, LGTM

TomSweeneyRedHat:
LGTM
@mheon A shiny merge button if you want to push it

PR #10405, opened 2021-05-19
Always spawn a cleanup process with exec

mheon:
We were previously only doing this for detached exec. I don't know why we did that, but I don't see any reason not to extend it to all exec sessions - it guarantees that we will always clean up exec sessions, even if the original `podman exec` process died.

[NO TESTS NEEDED] because I don't really know how to test this one.

### DIFF
--- a/libpod/container_exec.go
+++ b/libpod/container_exec.go
@@ -761,6 +761,9 @@ func (c *Container) Exec(config *ExecConfig, streams *define.AttachStreams, resi
 	}
 	exitCode := session.ExitCode
 	if err := c.ExecRemove(sessionID, false); err != nil {
+		if errors.Cause(err) == define.ErrNoSuchExecSession {
+			return exitCode, nil
+		}
 		return -1, err
 	}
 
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM

vrothberg (commented):
/lgtm

mheon:
I think the errors are legit - I need to handle the already-removed race condition in the exec code.

mheon:
Alright, now I realize why I didn't do this originally. We now have a race where `podman exec` needs to pick up the exit code of the exec session from the container, but the cleanup process is wiping the exec session (and said exit code) at the same time.

We can potentially work around this by having a special event for exec session exit (Docker uses `exec_die` for this) and using logic similar to what `podman run` uses when the container is removed before we can read its exit code. I'll get this coded up tomorrow.

mheon:
Re-pushed with a fix. Added an `exec_died` event, which we now use for retrieving exit codes on exec sessions being removed before we can retrieve them.

vrothberg:
Restarted the two jobs. Looked like flakes.

mheon:
I think they are, but they seem extremely consistent.

mheon:
/hold cancel

Oh dear, it went green. @containers/podman-maintainers Anyone want to drop an LGTM?

PR #10416, opened 2021-05-20
pkg/systemd: don't require LISTEN_FDNAMES for socket activation

tych0:
LISTEN_FDNAMES is optional, the docs for sd_listen_fds() says:

    This information is read from the $LISTEN_FDNAMES variable, which
    **may** contain a colon-separated list of names.

emphasis mine (indeed, the cited coreos code also suggests it is optional).

This actually results in bug, since the default
/contrib/systemd/system/podman.socket file doesn't set a
FileDescriptorName=. podman when run with this systemd configuration
*always* starts in unix socket mode since SocketActivated() will return
false because the name is missing.

The bug is a race with a very small window: between when podman does the
unlink() and when it re-binds the socket later in the code, requests made
during this time will fail since nothing is listening. There's another
small race when the service stops and systemd realizes it and starts
listening again.

However, small this window we managed to hit it :).

Let's fix this by ignoring LISTEN_FDNAMES. Since the code in
cmd/podman/system/service_abi.go:restService() ignores this value anyway
when setting up the socket activated stuff, there's no real loss here.

Signed-off-by: Tycho Andersen <tycho@tycho.pizza>

### DIFF
--- a/pkg/systemd/activation.go
+++ b/pkg/systemd/activation.go
@@ -25,11 +25,5 @@ func SocketActivated() bool {
 	if err != nil || nfds == 0 {
 		return false
 	}
-
-	// "github.com/coreos/go-systemd/v22/activation" will use and validate this variable's
-	// value. We're just providing a fast fail
-	if _, found = os.LookupEnv("LISTEN_FDNAMES"); !found {
-		return false
-	}
 	return true
 }
--- a/pkg/systemd/activation_test.go
+++ b/pkg/systemd/activation_test.go
@@ -1,0 +1,32 @@
+package systemd
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSocketActivated(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.False(SocketActivated())
+
+	// different pid
+	assert.NoError(os.Setenv("LISTEN_PID", "1"))
+	assert.False(SocketActivated())
+
+	// same pid no fds
+	assert.NoError(os.Setenv("LISTEN_PID", fmt.Sprintf("%d", os.Getpid())))
+	assert.NoError(os.Setenv("LISTEN_FDS", "0"))
+	assert.False(SocketActivated())
+
+	// same pid some fds
+	assert.NoError(os.Setenv("LISTEN_FDS", "1"))
+	assert.True(SocketActivated())
+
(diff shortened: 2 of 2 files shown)


mheon:
PR is complaining about a lack of tests. I don't believe we have existing tests for socket activation (though @edsantiago may correct me there), so you will probably have to add `[NO TESTS NEEDED]` to the PR title and re-push.

mheon:
/approve
@giuseppe PTAL

tych0:
Thanks, I think it shouldn't be too hard to write some tests for this, so I'll go ahead and do that later today. Have to run for now though.

giuseppe (approved):
LGTM

eriksjolund:
Quote from `man systemd.socket` (regarding `FileDescriptorName=` in Systemd version 248) :
_If this setting is not used, the file descriptor name defaults to the name of the socket unit, including its .socket suffix_

One way to interpret that sentence is that LISTEN_FDNAMES would always be set.
(I'm a bit confused about whether LISTEN_FDNAMES is optional or not)

Anyway if we assume that LISTEN_FDNAMES is optional maybe this code needs to be adjusted too?
https://github.com/containers/podman/blob/4b3fc0c67a1eb4970b55828e1e1ac1c9679ef656/pkg/rootless/rootless_linux.c#L314
The if-statement in the following lines does not treat the environment variable LISTEN_FDNAMES as optional.
Another thing that I noticed is that `strdup(listen_fdnames)` may be given a NULL pointer. I've not verified it but it seems to be that way just by reading the code.

tych0:
> One way to interpret that sentence is that LISTEN_FDNAMES would always be set. (I'm a bit confused about whether LISTEN_FDNAMES is optional or not)

All I can say is that on systemd 219 (centos 7) it is not set, and that some man pages seem to think it is optional while others do not :)

I pushed a basic smoke test for the patch as-is, though.

eriksjolund:
I'm not sure that the code that I mentioned (in _rootless_linux.c_) is directly related to this PR. I'll need to investigate it a bit more. I could open up a new issue or PR if needed.

vrothberg (commented):
LGTM, thanks @tych0!

Could you rebase?

@jwhonce could this be the mysterious issue we saw rarely but didn't manage to track down yet?

tych0:
Sorry,l just realized you had asked for a rebase :). Done.

vrothberg:
Thanks for the fix, @tych0 !

Issue #10423, opened 2021-05-20
podman-remote run --rm: Error removing container / error looking up container / mounts: layer not known


edsantiago:
This looks kind of serious, mostly because it's so easily and often reproduced:
```console
[ assumes podman system service running elsewhere]
$ while :;do ./bin/podman-remote run --rm alpine date;echo $?;done
Thu May 20 21:39:28 UTC 2021
0
...
Thu May 20 21:39:33 UTC 2021
ERRO[0000] Error removing container b3286c58744903d1e5b7d679ca9168ca16c2d5d1a6084b3aa3919c41cbe28f8e: error looking up container "b3286c58744903d1e5b7d679ca9168ca16c2d5d1a6084b3aa3919c41cbe28f8e" mounts: layer not known
...
Thu May 20 21:39:35 UTC 2021
WARN[0000] Container 6b4d7060b9466bd7c0e806758b3c2366227ff415784541882ea324c0bfb1b9e2 does not exist: container 6b4d7060b9466bd7c0e806758b3c2366227ff415784541882ea324c0bfb1b9e2 does not exist in database: no such container
0
...
Thu May 20 21:39:37 UTC 2021
ERRO[0000] Error removing container 20c17035174b364f7e3d983ee0af899eb7af635c5a8d2d09b48b44932d606086: error looking up container "20c17035174b364f7e3d983ee0af899eb7af635c5a8d2d09b48b44932d606086" mounts: layer not known
```
Sometimes it's ERRO, sometimes WARN. In all cases, exit status is 0 (no error). 

master @ e48aa8c82f7f26260f6911bb3ea99ff54c294675, root and rootless

rhatdan:
This is the fix in Buildah to show log messages on the client side, these messages used to be just on the server side.  I think what is happening, is a race condition between the container being removed in the API and the container being removed via conmon.

In either case we should drop the remove error on container does not exist to debug it should not be warning or error.

edsantiago:
Ah, thanks for the analysis. I would recommend making this a top priority; otherwise, podman-remote tests in CI are going to start flaking uncontrollably.

rhatdan:
I ran your test with the patch for a minute and saw no errors.

edsantiago:
#10427 and #10421 have merged, I believe this is fixed.

PR #10427, opened 2021-05-21
Drop container does not exist on removal to debugf

rhatdan:
We have race conditions where a container can be removed
by two different processes when running podman --remove rm.

It can be cleaned up in the API or by the conmon executing
podman container cleanup.

When we fail to remove a container that does not exists we should
not be printing errors or warnings, we should just debug the fact.

[NO TESTS NEEDED] Since this is a race condition it is difficult to
test.

### DIFF
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -275,7 +275,7 @@ func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string,
 		case nil:
 			// remove container names that we successfully deleted
 			reports = append(reports, &report)
-		case define.ErrNoSuchCtr:
+		case define.ErrNoSuchCtr, define.ErrCtrExists:
 			// There is still a potential this is a libpod container
 			tmpNames = append(tmpNames, ctr)
 		default:
--- a/pkg/domain/infra/tunnel/containers.go
+++ b/pkg/domain/infra/tunnel/containers.go
@@ -21,6 +21,7 @@ import (
 	"github.com/containers/podman/v3/pkg/errorhandling"
 	"github.com/containers/podman/v3/pkg/specgen"
 	"github.com/containers/podman/v3/pkg/util"
+	"github.com/containers/storage/types"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -580,7 +581,7 @@ func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []stri
 						if err := containers.Remove(ic.ClientCtx, ctr.ID, removeOptions); err != nil {
 							if errorhandling.Contains(err, define.ErrNoSuchCtr) ||
 								errorhandling.Contains(err, define.ErrCtrRemoved) {
-								logrus.Warnf("Container %s does not exist: %v", ctr.ID, err)
+								logrus.Debugf("Container %s does not exist: %v", ctr.ID, err)
 							} else {
 								logrus.Errorf("Error removing container %s: %v", ctr.ID, err)
 							}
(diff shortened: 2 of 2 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/10423

edsantiago:
Hold on please, I just saw:
```console
$ while :;do ./bin/podman-remote run --rm alpine true;done                                                                                     pr/10427
ERRO[0000] Error removing container 63170103bcc92cde7a8c3454f463def82e39196b3e588f75e183725f28433cf0: refusing to remove "63170103bcc92cde7a8c3454f463def82e39196b3e588f75e183725f28433cf0" as it exists in libpod as container 63170103bcc92cde7a8c3454f463def82e39196b3e588f75e183725f28433cf0: container already exists
```
Looks like https://github.com/containers/podman/blob/e48aa8c82f7f26260f6911bb3ea99ff54c294675/libpod/runtime_cstorage.go#L97

rhatdan:
@edsantiago Try again.

libpod/runtime_ctr.go
```diff
@@ -645,10 +645,14 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force, remo
 		}
 	} else {
 		if err := r.state.RemoveContainer(c); err != nil {
-			if cleanupErr == nil {
-				cleanupErr = err
+			if err == define.ErrNoSuchCtr {
```
mheon, line 648: I think this is still worth an error. If we got this far, we did it *holding the lock* - there is 0 chance the container was removed from under us. If this happens something very strange and very bad is going on - someone is messing around in the DB without obeying our locks.

rhatdan: Ok some of this was just guessing.  I will remove this check.

rhatdan: Actually @mheon @edsantiago reported in the original issue that he was getting.
WARN[0000] Container 6b4d7060b9466bd7c0e806758b3c2366227ff415784541882ea324c0bfb1b9e2 does not exist: container 6b4d7060b9466bd7c0e806758b3c2366227ff415784541882ea324c0bfb1b9e2 does not exist in database: no such container

Which looks like this is happening.

mheon: Then we have a serious bug, because that should never happen.

rhatdan: Ok I reverted, let's see if this error is reproducible, with the other fixes.

edsantiago: I've pulled your latest change (6ca721cccd98bf9936b33d690b28a52cd4213a18), rebuilt, and am running tests. No failures in the last two minutes, which is good. I need to be AFK periodically this morning but will check in as time allows.

edsantiago: Still running with no errors (20+ minutes). I call that good. Gotta head out now, will check back in in 90m or so.

TomSweeneyRedHat:
LGTM
but you're still fighting with the tests

edsantiago:
One of the [flakes](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4697591129571328/html/int-podman-fedora-33-root-container.log.html#t--podman-pod-prune-removes-a-pod-with-a-stopped-container--2) looks like #7139

libpod/runtime_ctr.go
edsantiago: Still no failures. I call this good.

edsantiago:
[The other flake](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5893859780591616/html/sys-podman-ubuntu-2010-rootless-host.log.html#t--00057) also looks like #7139

Issue #10420, opened 2021-05-20
--gpu option does not work with podman


asm582:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. Install podam

2. Run: 
`docker run --gpus all -it --rm tensorflow/tensorflow:latest-gpu \
   python -c "import tensorflow as tf; print(tf.reduce_sum(tf.random.normal([1000, 1000])))"`

**Describe the results you received:*

```
$ docker run --gpus all -it --rm tensorflow/tensorflow:latest-gpu \
  python -c "import tensorflow as tf; print(tf.reduce_sum(tf.random.normal([1000, 1000])))"
Error: unknown flag: --gpus
```

**Describe the results you expected:**
Since podman is a drop in replacement for docker I was expecting this command to be working

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
$ podman --version
podman version 1.9.3
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.13.4
  podmanVersion: 1.9.3
host:
  arch: amd64
  buildahVersion: 1.14.9
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.22-3.module+el8.3.1+9857+68fb1526.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.22, commit: a40e3092dbe499ea1d85ab339caea023b74829b9'
  cpus: 32
  distribution:
    distribution: '"rhel"'
    version: "8.3"
  eventLogger: file
  hostname: ccc3-06.sl.cloud9.ibm.com
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1001
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1001
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
  kernel: 4.18.0-240.15.1.el8_3.x86_64
  memFree: 535668387840
  memTotal: 540650921984
  ociRuntime:
    name: runc
    package: runc-1.0.0-70.rc92.module+el8.3.1+9857+68fb1526.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.2-dev'
  os: linux
  rootless: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
[...]

PR #10430, opened 2021-05-21
Add podman run --gpus flag for compatibility

sjug:
People who use docker scripts with Podman see failures
if they use `podman run --gpus` flag. This is a noop flag.

Resolves https://github.com/containers/podman/issues/10420

### DIFF
--- a/cmd/podman/containers/run.go
+++ b/cmd/podman/containers/run.go
@@ -77,6 +77,11 @@ func runFlags(cmd *cobra.Command) {
 	flags.StringVar(&runOpts.DetachKeys, detachKeysFlagName, containerConfig.DetachKeys(), "Override the key sequence for detaching a container. Format is a single character `[a-Z]` or a comma separated sequence of `ctrl-<value>`, where `<value>` is one of: `a-cf`, `@`, `^`, `[`, `\\`, `]`, `^` or `_`")
 	_ = cmd.RegisterFlagCompletionFunc(detachKeysFlagName, common.AutocompleteDetachKeys)
 
+	gpuFlagName := "gpus"
+	flags.String(gpuFlagName, "", "This is a Docker specific option and is a NOOP")
+	_ = cmd.RegisterFlagCompletionFunc(gpuFlagName, completion.AutocompleteNone)
+	_ = flags.MarkHidden("gpus")
+
 	if registry.IsRemote() {
 		_ = flags.MarkHidden("preserve-fds")
 		_ = flags.MarkHidden("conmon-pidfile")
@@ -203,6 +208,9 @@ func run(cmd *cobra.Command, args []string) error {
 		if len(rmErrors) > 0 {
 			logrus.Errorf("%s", errorhandling.JoinErrors(rmErrors))
 		}
+	}
+	if cmd.Flag("gpus").Changed {
+		logrus.Info("--gpus is a Docker specific option and is a NOOP")
 	}
 	return nil
 }
--- a/test/e2e/run_device_test.go
+++ b/test/e2e/run_device_test.go
@@ -113,4 +113,10 @@ var _ = Describe("Podman run device", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 		Expect(session.OutputToString()).To(Equal("/dev/kmsg1"))
 	})
+
+	It("podman run --gpus noop", func() {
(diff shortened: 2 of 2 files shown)


sjug:
Looks like I need to either not have docs (man pages) or not make the flag hidden.
Which is the preferred outcome?

cmd/podman/containers/run.go
```diff
@@ -77,6 +77,11 @@ func runFlags(cmd *cobra.Command) {
 	flags.StringVar(&runOpts.DetachKeys, detachKeysFlagName, containerConfig.DetachKeys(), "Override the key sequence for detaching a container. Format is a single character `[a-Z]` or a comma separated sequence of `ctrl-<value>`, where `<value>` is one of: `a-cf`, `@`, `^`, `[`, `\\`, `]`, `^` or `_`")
 	_ = cmd.RegisterFlagCompletionFunc(detachKeysFlagName, common.AutocompleteDetachKeys)
 
+	gpuFlagName := "gpus"
+	flags.String(gpuFlagName, "", "This is a Docker specific option and is a NOOP")
```
mheon, line 81: I wonder if it's worth printing a log message (Info level, probably) if this is actually set, notifying the user it has no effect

sjug: Added info level log message if the flag is set; message follows description of flag.

rhatdan: Hide it, no reason for it other then backwards compatibility with Docker CLI.

sjug: @rhatdan Did you mean hide the flag from `--help`?

rhatdan: Hide the command from --help

docs/source/markdown/podman-run.1.md
```diff
@@ -405,6 +405,14 @@ Meaning **groupname** is initially mapped to gid **100000** which is referenced
 above: The group **groupname** is mapped to group **100000** of the initial namespace then the
 **30000**st id of this namespace (which is gid 130000 in this namespace) is mapped to container namespace group id **0**. (groupname -> 100000 / 30000 -> 0)
 
+#### **--gpus**=*all|device=*
+
+Since containerd does not support executing prestart hooks, Docker introduced the `--gpu` flag which is a hardcoded call to the NVIDIA prestart hook.
```
TomSweeneyRedHat, line 410:
```suggestion
Since containerd does not support executing prestart hooks, Docker introduced the `--gpu` flag, which is a hardcoded call to the NVIDIA prestart hook.
```

docs/source/markdown/podman-run.1.md
```diff
@@ -405,6 +405,14 @@ Meaning **groupname** is initially mapped to gid **100000** which is referenced
...
+
+To support automatic detection of GPUs install OCI pre-start hook provided by the `nvidia-container-toolkit` package on RHEL/Fedora.
```
TomSweeneyRedHat, line 412:
```suggestion
To support automatic detection of GPUs, install the OCI pre-start hook provided by the `nvidia-container-toolkit` package on RHEL/Fedora.
```

docs/source/markdown/podman-run.1.md
```diff
@@ -405,6 +405,14 @@ Meaning **groupname** is initially mapped to gid **100000** which is referenced
...
+
+This flag is a NOOP and provided solely for scripting compatibility.
```
TomSweeneyRedHat, line 414:
```suggestion
This flag is a NOOP and is provided solely for scripting compatibility.
```

sjug:
Thanks for the review @TomSweeneyRedHat, it seems the problem I'm running into now is that the CI does not allow me to mark the flags hidden while also adding documentation to the man files. However, I do think that the documentation should be added in some place.
AFAICT without the OCI hooks installed the GPU support will not work, so users adding this flag may be curious why it doesn't work for them.

rhatdan:
@sjug Don't document them.  We don't want users to know about them.  The only goal is if they have a script built with docker in mind, or with podman-docker installed, it completes successfully, but we don't want to have users thinking the option is useful with podman.

TomSweeneyRedHat:
@sjug one test failed with absolutely no bread crumbs, just restarted.
Otherwise LGTM

giuseppe (approved):
/lgtm

Issue #10261, opened 2021-05-07
docker compatibility: Container Inspect returns {} for NetworkSettings.Networks


candrews:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**
Java 11 (or later) is a prerequisite for running this test.

1. Clone https://github.com/candrews/podman-testcontainers-issue

2. Run `./test-podman.sh`

**Describe the results you received:**
The test fails. Looking at the stdout of the failing test:
```
[main] ERROR 🐳 [curlimages/curl:7.76.0] - Log output from the failed container:
curl: (6) Could not resolve host: host.testcontainers.internal
```

**Describe the results you expected:**
The test should pass; run `./test-docker.sh` to run against docker and observe the test passing.

**Additional information you deem important (e.g. issue happens only occasionally):**

Here's a log of the socket when run against docker (which works):
[docker.log](https://github.com/containers/podman/files/6443423/docker.log)

And here's a log of the same test run against podman (which fails):
[podman.log](https://github.com/containers/podman/files/6443424/podman.log)

**Output of `podman version`:**

```
podman version 3.2.0-rc1
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.2-dev
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: craigatwork
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.11.18-300.fc34.x86_64
  linkmode: dynamic
  memFree: 584716288
  memTotal: 16647507968
  ociRuntime:
    name: crun
    package: crun-0.19.1-2.fc34.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.19.1
      commit: 1535fedf0b83fb898d449f9680000f729ba719f5
      spec: 1.0.0
[...]

PR #10447, opened 2021-05-24
Add the option of Rootless CNI networking by default

mheon:
When the containers.conf field "NetNS" is set to "Bridge" and the "RootlessNetworking" field is set to "cni", Podman will now handle rootless in the same way it does root - all containers will be joined to a default CNI network, instead of exclusively using slirp4netns. 

If no CNI default network config is present for the user, one will be auto-generated (this also works for root, but it won't be nearly as common there since the package should already ship a config).

I eventually hope to remove the "NetNS=Bridge" bit from containers.conf, but let's get something in for Brent to work with.

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/podman/v3/cmd/podman/registry"
 	"github.com/containers/podman/v3/pkg/api/handlers"
 	"github.com/containers/podman/v3/pkg/cgroups"
@@ -140,7 +141,7 @@ func stringMaptoArray(m map[string]string) []string {
 
 // ContainerCreateToContainerCLIOpts converts a compat input struct to cliopts so it can be converted to
 // a specgen spec.
-func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroupsManager string) (*ContainerCLIOpts, []string, error) {
+func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, rtc *config.Config) (*ContainerCLIOpts, []string, error) {
 	var (
 		capAdd     []string
 		cappDrop   []string
@@ -248,7 +249,7 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	}
 
 	// netMode
-	nsmode, _, err := specgen.ParseNetworkNamespace(string(cc.HostConfig.NetworkMode))
+	nsmode, _, err := specgen.ParseNetworkNamespace(string(cc.HostConfig.NetworkMode), true)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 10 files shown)


Luap99:
Sorry I'm late, but why do we need two configuration fields? Shouldn't netns=bridge and netns=slirp4netns be enough. I don't see the need for the RootlessNetworking field?!

libpod/runtime.go
```diff
@@ -532,6 +533,13 @@ func makeRuntime(ctx context.Context, runtime *Runtime) (retErr error) {
 		}
 	}
 
+	// If we need to make a default network - do so now.
```
Luap99, line 536: This should happen before `ocicni.InitCNI` (line 465).

mheon:
@Luap99 The idea was to keep `NetNS=private` in all cases, and just adjust `RootlessNetworking` to set the default behavior. I'm beginning to doubt that it's necessary, though. Fortunately, I think we can remove the field without adverse effect in the future - TOML doesn't really care about field names it doesn't know about.

Luap99:
OK, one last point. When the compat api is used as rootless we default to slirp4netns, see #10261. We could not swap the default to bridge since there was no default cni config. With this patch we could change this. However, the config is only created when `rootless_networking="cni"` is set in containers.conf. I would argue that we should always create the default cni config since this would allow us to default to bridge in the compat api without forcing the user to change his config file.

mheon:
I was waffling on that one, but if we have a good reason for it to be universal, I'm happy to make the change.

mheon:
@Luap99 Made default network creation by default. I left the RootlessNetworking bits intact for now - mainly because I don't see an easier way of setting the default for rootless if the user sets `--net=private`  or some other default value (or the equivalent in containers.conf). I think this is probably worth further discussion, and we have time to do that since this won't ship in 3.2. Test has been added. This is probably good to merge if it goes green, barring further discussion.

mheon:
We also probably want to have a discussion about containers.conf stability - I feel like it's not held up to the same standards of stability as the rest of the API because we can remove config options without much consequence. So even if this did ship, IMO we still have the ability to remove it in the future without a major version bump...

cmd/podman/common/create_opts.go
```diff
@@ -248,7 +249,7 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, cgroup
 	}
 
 	// netMode
-	nsmode, _, err := specgen.ParseNetworkNamespace(string(cc.HostConfig.NetworkMode))
+	nsmode, _, err := specgen.ParseNetworkNamespace(string(cc.HostConfig.NetworkMode), rtc.Containers.RootlessNetworking == "cni")
```
Luap99, line 252: always set this to true for compat api, this should fix #10261

libpod/container.go
```diff
@@ -1168,7 +1167,7 @@ func (c *Container) Networks() ([]string, bool, error) {
 func (c *Container) networks() ([]string, bool, error) {
 	networks, err := c.runtime.state.GetNetworks(c)
 	if err != nil && errors.Cause(err) == define.ErrNoSuchNetwork {
-		if len(c.config.Networks) == 0 && !rootless.IsRootless() {
+		if len(c.config.Networks) == 0 && !c.config.NetMode.IsSlirp4netns() {
```
Luap99, line 1170: I think this should be `IsBridge()` to not match net=none

Luap99:
@mheon you have to rebase

mheon:
Done.

Luap99 (commented):
LGTM

TomSweeneyRedHat:
LGTM

Issue #10443, opened 2021-05-24
Socket activation works with TCP socket but not with Unix Domain socket


eriksjolund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I tested socket activation with _/usr/lib/systemd/systemd-socket-proxyd_ in four different ways.

| Running in Podman |     Socket type  | Systemd type  | Working? |
| ---                     |       ---               | -- | --              |
| yes                    | TCP socket    | Type=forking | yes  |
| yes                    | Unix Domain socket | Type=forking | no |
| no                     |   TCP socket    |  Type=exec |  yes          |
| no                     |   Unix Domain socket | Type=exec | yes |

The tests were performed in Systemd user services (i.e. running rootless).

I also tried adding the  _--privileged_ flag but it didn't help.

**Steps to reproduce the issue:**

#### TCP socket with Podman : Success 

```
[erikdev@laptop ~]$ mkdir systemd_container
[erikdev@laptop ~]$ cd systemd_container/
[erikdev@laptop systemd_container]$ emacs -nw Dockerfile
[erikdev@laptop systemd_container]$ cat Dockerfile
FROM registry.fedoraproject.org/fedora:34
RUN dnf -y install systemd && dnf clean all
CMD ["/usr/lib/systemd/systemd-socket-proxyd","fedoramagazine.org:443"]

[erikdev@laptop systemd_container]$ podman build -t systemd .
STEP 1: FROM registry.fedoraproject.org/fedora:34
STEP 2: RUN dnf -y install systemd && dnf clean all
--> Using cache ed085871341244571e0f8d655039e5fa1f6d47b90054e9665dfa866b51bf6253
--> ed085871341
STEP 3: CMD ["/usr/lib/systemd/systemd-socket-proxyd","fedoramagazine.org:443"]
--> Using cache 035bd01dfa88ab15f3e637ddc8a257d00b575dd490e48985304093ed9c178a21
STEP 4: COMMIT systemd
--> 035bd01dfa8
035bd01dfa88ab15f3e637ddc8a257d00b575dd490e48985304093ed9c178a21
[erikdev@laptop systemd_container]$ cd ~/.config/systemd/user
[erikdev@laptop user]$ podman create --name testing localhost/systemd
d69c504e11c880532c66bf9adb699499a880dbd481a1df7cb7f7b0fd1f3d3605
[erikdev@laptop user]$ podman generate systemd --new --name testing > testing.service
[...]

PR #10448, opened 2021-05-24
Pass Systemd LISTEN_* environment to the container

rhatdan:
If a container is running within a systemd service and it is socket
activated, we need to leak the LISTEN_* environment variables into the
container.

Fixes: https://github.com/containers/podman/issues/10443

### DIFF
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -1055,6 +1055,16 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		}
 	}
 
+	if l, ok := os.LookupEnv("LISTEN_FDS"); ok {
+		listenFds, err := strconv.ParseUint(l, 10, 64)
+		if err != nil {
+			logrus.Warnf("Error LISTEN_FDS environment variable for %s not an int %v", ctr.ID(), err)
+		}
+
+		if uint(listenFds) > ctr.config.PreserveFDs {
+			ctr.config.PreserveFDs = uint(listenFds)
+		}
+	}
 	if ctr.config.PreserveFDs > 0 {
 		args = append(args, formatRuntimeOpts("--preserve-fds", fmt.Sprintf("%d", ctr.config.PreserveFDs))...)
 	}
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -648,6 +648,18 @@ func (c *Container) generateSpec(ctx context.Context) (*spec.Spec, error) {
 		g.AddProcessEnv("HOSTNAME", hostname)
 	}
 
+	for _, lEnv := range []string{"LISTEN_PID", "LISTEN_FDS", "LISTEN_FDNAMES"} {
+		if val, ok := os.LookupEnv(lEnv); ok {
+			// The primary process within the container will be PID=1, so this
+			// value needs to be reset
+			if lEnv == "LISTEN_PID" {
+				g.AddProcessEnv(lEnv, "1")
+				continue
+			}
+			g.AddProcessEnv(lEnv, val)
+		}
+	}
+
 	if c.config.UTSNsCtr != "" {
 		if err := c.addNamespaceContainer(&g, UTSNS, c.config.UTSNsCtr, spec.UTSNamespace); err != nil {
 			return nil, err
--- a/test/system/250-systemd.bats
+++ b/test/system/250-systemd.bats
@@ -148,4 +148,43 @@ function service_cleanup() {
     service_cleanup
 }
 
+@test "podman pass run LISTEN environment " {
+    tmpdir=$PODMAN_TMPDIR/build-test
+    subdir=$tmpdir/subdir
+    run_podman run --hostname=host1 --rm $IMAGE printenv
+    std_output=$output
+
+    export LISTEN_PID="100" LISTEN_FDS="1" LISTEN_FDNAMES="listen_fdnames"
+    run_podman run --hostname=host1 --rm $IMAGE printenv
+    if is_remote; then
+	is "$output" "$std_output" "LISTEN Environment did not pass"
+    else
+	is "$output" "$std_output
+LISTEN_PID=1
+LISTEN_FDS=1
+LISTEN_FDNAMES=listen_fdnames" "LISTEN Environment passed"
+    fi
+    unset LISTEN_PID LISTEN_FDS LISTEN_FDNAMES
+}
+
+@test "podman pass start LISTEN environment " {
+    tmpdir=$PODMAN_TMPDIR/build-test
+    subdir=$tmpdir/subdir
+    run_podman run --hostname=host1 --rm $IMAGE printenv
+    std_output=$output
+
+    run_podman create --name=test --hostname=host1 --rm $IMAGE printenv
+    export LISTEN_PID="100" LISTEN_FDS="1" LISTEN_FDNAMES="listen_fdnames"
+    run_podman start --attach test
+    if is_remote; then
+	is "$output" "$std_output" "LISTEN Environment did not pass"
+    else
+	is "$output" "$std_output
+LISTEN_PID=1
+LISTEN_FDS=1
+LISTEN_FDNAMES=listen_fdnames" "LISTEN Environment passed"
+    fi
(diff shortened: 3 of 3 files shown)


rhatdan:
@eriksjolund I would love to work with you on this.  I think this PR is needed to move this forward.

Basically this PR passes the environment variables into the container so PID1 of the container can use them.  We force the LISTEN_PID=1 since systemd will originally set it to the Podman PID.

I am hoping that conmon and crun/runc handle down the open file descriptors correctly.  If you could build a simple test script to test this, I could finish it up.

rhatdan:
@giuseppe PTAL

rhatdan:
@haircommander @giuseppe I believe we have to update the  --preservefds as well when executing conmon.

eriksjolund:
@rhatdan Sure, I'll create a simple test script. I could try it out during the weekend.

libpod/oci_conmon_linux.go
```diff
@@ -1051,6 +1051,16 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
 		}
 	}
 
+	if l, ok := os.LookupEnv("LISTEN_FDS"); ok {
+		listenFds, err := strconv.ParseUint(l, 10, 64)
+		if err != nil {
+			logrus.Warnf("Error LISTEN_FDS environment variable not an int %v", err)
```
jwhonce, line 1057: Would adding the container id to the warning provide context?

cmd/podman/system/service_abi.go
```diff
@@ -82,6 +82,12 @@ func restService(opts entities.ServiceOptions, flags *pflag.FlagSet, cfg *entiti
 		}
 	}()
 
+	// If the podman service is setup as socket activated, we don't want any containers
+	// to inherit these settings from the service.
+	for _, lEnv := range []string{"LISTEN_PID", "LISTEN_FDS", "LISTEN_FDNAMES"} {
+		os.Unsetenv(lEnv)
```
jwhonce, line 88: These are also unset in the coreos/go-systemd library when the service is started. See https://github.com/containers/podman/blob/d9eb1269257490dddfae3fde2fe57c3009f94287/vendor/github.com/coreos/go-systemd/v22/activation/files_unix.go#L39 that is called during NewServer...() before Serve()

libpod/oci_conmon_linux.go
rhatdan: Added

cmd/podman/system/service_abi.go
rhatdan: Except I don't see activation.Files(true) ever being called.

jwhonce: From `activation.Listeners()` which is called in NewServer...()

https://github.com/containers/podman/blob/d9eb1269257490dddfae3fde2fe57c3009f94287/vendor/github.com/coreos/go-systemd/v22/activation/listeners.go#L29

rhatdan:
@eriksjolund PTAL

rhatdan:
@giuseppe @mheon @vrothberg @edsantiago PTAL, Anyone have a simple test to see if this works?

eriksjolund:
Just some thoughts. (Disclaimer: sending this quite late in the evening so it might not be crystal clear).

_crun_  takes the LISTEN_FDS information from the environment
```
grep -r LISTEN_FDS .
./src/create.c:  if (getenv ("LISTEN_FDS"))
./src/create.c:    crun_context.preserve_fds += strtoll (getenv ("LISTEN_FDS"), NULL, 10);
./src/run.c:  if (getenv ("LISTEN_FDS"))
./src/run.c:    crun_context.preserve_fds += strtoll (getenv ("LISTEN_FDS"), NULL, 10);
./src/exec.c:  if (getenv ("LISTEN_FDS"))
./src/exec.c:    crun_context.preserve_fds += strtoll (getenv ("LISTEN_FDS"), NULL, 10);
```
That looks good.

I think we need to keep the preserve_fds information and LISTEN_FDS information apart in the Podman and Conmon code.

In the source code for Podman and Conmon the term preserve_fds could for instance be "reserved" to always mean the user-provided argument __--preserve-fds__ from `podman run` and `podman exec`. It would be a constant value  that would not change.

One thing to think about

Quote from: https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html
```
#define SD_LISTEN_FDS_START 3
```

In other words. The first LISTEN_FDS file descriptor needs to come directly after stdin, stdout, stderr.

libpod/oci_conmon_linux.go
```diff
@@ -1055,6 +1055,16 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
...
 
+	if l, ok := os.LookupEnv("LISTEN_FDS"); ok {
+		listenFds, err := strconv.ParseUint(l, 10, 64)
+		if err != nil {
+			logrus.Warnf("Error LISTEN_FDS environment variable for %s not an int %v", ctr.ID(), err)
+		}
+
+		if uint(listenFds) > ctr.config.PreserveFDs {
+			ctr.config.PreserveFDs = uint(listenFds)
+		}
```
eriksjolund, line 1066:
```
		if uint(listenFds) > ctr.config.PreserveFDs {
			ctr.config.PreserveFDs = uint(listenFds)
		}
```
seems to be 
(pseudo code)
```
ctr.config.PreserveFDs := max(listenFds, ctr.config.PreserveFDs)
```

I don't understand this code. Why should it be a max value?

rhatdan: I am just guessing here, that if the user specified --preserve_fds=1 and LISTEN_FDS=2, then we need to leak at least 2 FDs into the container.  But I am not really sure what should be the behaviour.  Perhaps we should ignore the LISTEN_FDS and force the user to tell us how many FDS to leak.

rhatdan:
I don't want to add another option about leaked FDs into conmon, since conmon would not know which one to take priority with.
@giuseppe @vrothberg @haircommander WDYT?

libpod/oci_conmon_linux.go
eriksjolund: At first sight runc and crun are doing the same:
See the runc example:
```
 LISTEN_PID=$pid_of_runc LISTEN_FDS=3 runc run --preserve-fds 5 <container>
```

from https://github.com/opencontainers/runc/blob/master/docs/terminals.md

-------------


Quote: _basically, it can just leave the FDs and $LISTEN_FDS untouched_
https://systemd.io/CONTAINER_INTERFACE/

I guess runc and crun could have ignored LISTEN_FDS and just make it the responsibility of 
the caller to specify a --preserve-fds  argument that includes the LISTEN_FDS.

But that is not how it is done right now.

rhatdan: So we should do nothing with them except leak them into the OCI Runtimes.  Will crun and runc rename the LISTEN_PID to pid1?

eriksjolund:
@rhatdan Regarding the test scripts, I gave it a try but I haven't finished it yet.
I wanted to do a simple server that echoed the input back.

Here is some code that might be useful if anyone is trying to do the same
```
$ egrep -r 'echo.sock|5555' . | grep service
./oneshot-test.service:ExecStart=bash -c "echo abc | socat tcp:localhost:5555 - > %t/reply.txt"
./echo.service:#ExecStart=socat TCP4-LISTEN:5555,fork EXEC:/bin/cat
./echo.service:ExecStart=socat UNIX-LISTEN:%t/echo.sock,fork EXEC:/bin/cat
./socket-activated-echo.service: #ExecStart=/usr/lib/systemd/systemd-socket-proxyd localhost:5555
./socket-activated-echo.service: ExecStart=/usr/lib/systemd/systemd-socket-proxyd %t/echo.sock
```
I don't remember what was not working. (Adding `podman` to the commands would be the next step).

rhatdan:
@eriksjolund any progress?

eriksjolund:
@rhatdan No, I have been busy with other things.

libpod/oci_conmon_linux.go
vrothberg: That's correct.  crun and runc preserve the FDS but only runc sets LISTEN_PID=1 (Cc: @giuseppe @flouthoc).

vrothberg:
Closing, @rhatdan aksed me to take over.

PR #10451, opened 2021-05-25
tests: update CI images

giuseppe:
preparation work for: https://github.com/containers/common/pull/573

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,17 +30,20 @@ env:
     PRIOR_UBUNTU_NAME: "ubuntu-2010"
 
     # Google-cloud VM Images
+    # TODO: At the time of this comment, an selinux-policy regression is blocking use of updated
+    # Fedora VM images: https://bugzilla.redhat.com/show_bug.cgi?id=1965743
+    IMAGE_SUFFIX_UBUNTU: "c5521575421149184"
     IMAGE_SUFFIX: "c5348179051806720"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
-    UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
-    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "prior-ubuntu-${IMAGE_SUFFIX}"
+    UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX_UBUNTU}"
+    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "prior-ubuntu-${IMAGE_SUFFIX_UBUNTU}"
 
     # Container FQIN's
     FEDORA_CONTAINER_FQIN: "quay.io/libpod/fedora_podman:${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CONTAINER_FQIN: "quay.io/libpod/prior-fedora_podman:${IMAGE_SUFFIX}"
-    UBUNTU_CONTAINER_FQIN: "quay.io/libpod/ubuntu_podman:${IMAGE_SUFFIX}"
-    PRIOR_UBUNTU_CONTAINER_FQIN: "quay.io/libpod/prior-ubuntu_podman:${IMAGE_SUFFIX}"
+    UBUNTU_CONTAINER_FQIN: "quay.io/libpod/ubuntu_podman:${IMAGE_SUFFIX_UBUNTU}"
+    PRIOR_UBUNTU_CONTAINER_FQIN: "quay.io/libpod/prior-ubuntu_podman:${IMAGE_SUFFIX_UBUNTU}"
 
     ####
     #### Control variables that determine what to run and how to run it.
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -921,17 +921,6 @@ USER mail`, BB)
 		Expect(session.OutputToString()).To(ContainSubstring("mail root"))
 	})
 
-	It("podman run with incorect VOLUME", func() {
-		dockerfile := fmt.Sprintf(`FROM %s
-VOLUME ['/etc/foo']
-WORKDIR /etc/foo`, BB)
-		podmanTest.BuildImage(dockerfile, "test", "false")
-		session := podmanTest.Podman([]string{"run", "--rm", "test", "echo", "test"})
-		session.WaitWithDefaultTimeout()
-		Expect(session.ExitCode()).To(Equal(0))
-		Expect(session.OutputToString()).To(ContainSubstring("test"))
-	})
-
 	It("podman run --volumes-from flag", func() {
 		vol := filepath.Join(podmanTest.TempDir, "vol-test")
 		err := os.MkdirAll(vol, 0755)
--- a/utils/utils_supported.go
+++ b/utils/utils_supported.go
@@ -172,7 +172,7 @@ func moveUnderCgroup(cgroup, subtree string, processes []uint32) error {
 		if len(processes) > 0 {
 			for _, pid := range processes {
 				if _, err := f.Write([]byte(fmt.Sprintf("%d\n", pid))); err != nil {
-					logrus.Warnf("Cannot move process %d to cgroup %q", pid, newCgroup)
+					logrus.Warnf("Cannot move process %d to cgroup %q: %v", pid, newCgroup, err)
 				}
 			}
 		} else {
@@ -185,7 +185,7 @@ func moveUnderCgroup(cgroup, subtree string, processes []uint32) error {
 					continue
 				}
 				if _, err := f.Write(pid); err != nil {
-					logrus.Warnf("Cannot move process %s to cgroup %q", string(pid), newCgroup)
+					logrus.Warnf("Cannot move process %s to cgroup %q: %v", string(pid), newCgroup, err)
 				}
 			}
 		}
--- a/utils/utils_supported.go
+++ b/utils/utils_supported.go
@@ -172,7 +172,7 @@ func moveUnderCgroup(cgroup, subtree string, processes []uint32) error {
 		if len(processes) > 0 {
 			for _, pid := range processes {
 				if _, err := f.Write([]byte(fmt.Sprintf("%d\n", pid))); err != nil {
-					logrus.Warnf("Cannot move process %d to cgroup %q: %v", pid, newCgroup, err)
+					logrus.Debugf("Cannot move process %d to cgroup %q: %v", pid, newCgroup, err)
 				}
 			}
 		} else {
@@ -185,7 +185,7 @@ func moveUnderCgroup(cgroup, subtree string, processes []uint32) error {
 					continue
 				}
 				if _, err := f.Write(pid); err != nil {
-					logrus.Warnf("Cannot move process %s to cgroup %q: %v", string(pid), newCgroup, err)
+					logrus.Debugf("Cannot move process %s to cgroup %q: %v", string(pid), newCgroup, err)
 				}
 			}
 		}


giuseppe:
we need an updated runc.  v1.0-rc95 has the feature we need

rhatdan:
@cevich PTAL

cevich:
> we need an updated runc. v1.0-rc95 has the feature we need

Sorry only just seeing this now after being on PTO last week.  Okay, I'll build new VM images...[okay, opened a PR.](https://github.com/containers/automation_images/pull/72)

cevich:
@giuseppe okay, give image suffix `c6704734020042752` a try.

```
$ podman run -it --rm quay.io/libpod/fedora_podman:c6704734020042752 rpm -q runc
Trying to pull quay.io/libpod/fedora_podman:c6704734020042752...
Copying blob 1ab2777b591f done
Copying blob 76d9e8b34382 done
Copying config 3584509b8e done
Storing signatures
runc-1.0.0-378.rc95.fc34.x86_64
```

rhatdan:
@giuseppe needs a rebase.  @cevich I think we need new VMs to get latest runc.

giuseppe:
I am hitting https://bugzilla.redhat.com/show_bug.cgi?id=1965743 now

cevich:
> I am hitting https://bugzilla.redhat.com/show_bug.cgi?id=1965743 now

Ugh, that's frustrating.  If it ain't one thing...it's another thing :cry: 

> we will need new images once that is fixed

Sure thing, happy to.  I left open [the image-build PR](https://github.com/containers/automation_images/pull/72) so when it's time, poke me and I can smash the re-run button.  That way the image-id won't change, it'll rebuild and overwrite.

giuseppe:
the `cri-o-runc` deb package on ubuntu:2010 is still too old, and I think it is also vulnerable to the last runc security issue.

@lsm5 FYI

giuseppe:
```
$ podman run --rm quay.io/libpod/prior-ubuntu_podman:c6704734020042752 dpkg-query -s cri-o-runc
Package: cri-o-runc
Status: install ok installed
Priority: extra
Section: devel
Installed-Size: 11743
Maintainer: Lokesh Mandvekar <lsm5@fedoraproject.org>
Architecture: amd64
Version: 1.0.0~rc93.1
Depends: libc6 (>= 2.28), libseccomp2 (>= 2.4.1)
Breaks: docker.io (<< 1.12)
Description: Open Container Project - runtime
 "runc" is a command line client for running applications packaged according
 to the Open Container Format (OCF) and is a compliant implementation of
 the Open Container Project specification.
 This package is a fork of the "runc' package, specifically for cri-o.
Homepage: https://github.com/opencontainers/runc

```

haircommander:
you may want an apt update, I updated to rc95 the other day: https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/cri-o-runc

giuseppe:
Thanks but I think we are not using that repository, as I don't see any possible upgrade.

@cevich do you think we can move to using that packages for our images?

cevich:
> Thanks but I think we are not using that repository, as I don't see any possible upgrade.

Yep, you got it.

> you may want an apt update, I updated to rc95 the other day:

We have a special 'testing' kubik repo. @lsm5 made just for CI. It's *suppose* to be even more bleeding edge.  @haircommander do you know how to check if maybe the build failed or never happened in that repo?

haircommander:
oh it definitely never happened. lemme see how easy it'll be to mirror my changes

haircommander:
builds are in the oven now :bread:

cevich:
Oh goody, I'm glad you could figure it out.

@lsm5 - un-ping :smile:

giuseppe:
the failures don't seem related to the seccomp change

rhatdan:
@giuseppe needs a rebase.

giuseppe:
@adrianreber it looks like some checkpoint tests are failing with runc-1.0-rc95 on Ubuntu:

```
[+1588s] Running: /var/tmp/go/src/github.com/containers/podman/bin/podman --storage-opt vfs.imagestore=/tmp/podman/imagecachedir --root /tmp/podman_test174508096/crio --runroot /tmp/podman_test174508096/crio-run --runtime /usr/lib/cri-o-runc/sbin/runc --conmon /usr/libexec/podman/conmon --cni-config-dir /etc/cni/net.d --cgroup-manager systemd --tmpdir /tmp/podman_test174508096 --events-backend file --storage-driver vfs container checkpoint --with-previous d634e5cabd4879629628bb0b1775dce9c701f79ae2a63acf4f740d288f5cee68
[+1588s] time="2021-06-10T11:08:49Z" level=error msg="--parent-path must be relative"
[+1588s] Error: `/usr/lib/cri-o-runc/sbin/runc checkpoint --image-path /tmp/podman_test174508096/crio/vfs-containers/d634e5cabd4879629628bb0b1775dce9c701f79ae2a63acf4f740d288f5cee68/userdata/checkpoint --work-path /tmp/podman_test174508096/crio/vfs-containers/d634e5cabd4879629628bb0b1775dce9c701f79ae2a63acf4f740d288f5cee68/userdata --parent-path /tmp/podman_test174508096/crio/vfs-containers/d634e5cabd4879629628bb0b1775dce9c701f79ae2a63acf4f740d288f5cee68/userdata/pre-checkpoint d634e5cabd4879629628bb0b1775dce9c701f79ae2a63acf4f740d288f5cee68` failed: exit status 1
```

Not sure if it is a regression in runc, or an issue in Podman

adrianreber:
That is from a fix in runc: https://github.com/opencontainers/runc/pull/2913

pre-checkpointing only works correctly with a relative path

The reason why it did not fail before is that CRIU used to silently ignore if the parent path was not a relative path. CRIU just switches to non-precheckpointing mode. So the test was definitely always wrong in Podman. At least it was not doing what it was supposed.

Newer versions of CRIU (unreleased) will fail. There was also a check introduced in runc which is why this breaks now here.

Looking how to fix it now.

adrianreber:
This boils down to something like this:
```
diff --git a/libpod/oci_conmon_linux.go b/libpod/oci_conmon_linux.go
index 3da49b85f..58e0a284b 100644
--- a/libpod/oci_conmon_linux.go
+++ b/libpod/oci_conmon_linux.go
@@ -760,7 +760,7 @@ func (r *ConmonOCIRuntime) CheckpointContainer(ctr *Container, options Container
        // imagePath is used by CRIU to store the actual checkpoint files
        imagePath := ctr.CheckpointPath()
        if options.PreCheckPoint {
-               imagePath = ctr.PreCheckPointPath()
+               imagePath = "../pre-checkpoint"
        }
        // workPath will be used to store dump.log and stats-dump
        workPath := ctr.bundlePath()
```

I will open a PR to fix this.

adrianreber:
See #10639 for a fix.

giuseppe:
@rhatdan @cevich PTAL

.cirrus.yml
```diff
@@ -30,17 +30,18 @@ env:
     PRIOR_UBUNTU_NAME: "ubuntu-2010"
 
     # Google-cloud VM Images
+    IMAGE_SUFFIX_UBUNTU: "c5521575421149184"
```
cevich, line 33: This could lead to trouble (and maintenance headaches).  Could you add a comment explaining why the same set of images can't be used for both Fedora and Ubuntu?

Also, the `meta_task` will need updating to block the Ubuntu images from being pruned.

test/e2e/run_test.go
```diff
@@ -922,6 +922,7 @@ USER mail`, BB)
 	})
 
 	It("podman run with incorect VOLUME", func() {
+		Skip("FIXME: it breaks with the latest runc release")
```
cevich, line 925: It would be helpful if you mention the issue number open for this in the skip message.  There is an open issue on it, right?

cevich: Clarification:  "helpful" as in - the skipped test will not be forgotton about (and abandoned) if there's an open issue tracking it.  Also, having the issue # in the skip message is helpful for people to quickly find the how/who/whyfor the test is being skipped.

cevich (commented):
I hate to ask for re-running all the tests *just* to update the `meta_task` (and add a comment).  Unfortunately that's the only mechanism we have to track the "in-use" status of VM images.  I'm okay if you'd prefer to fix it in a followup `[CI:DOCS]` PR.

What's the reason why the new Fedora image can't also be used?

test/e2e/run_test.go
giuseppe: thanks for spotting it.  I've dropped this patch for now as we are using newer images, let's see if it is still failing

giuseppe:
> What's the reason why the new Fedora image can't also be used?

the Fedora images have a selinux-policy regression that causes all tests to fail

cevich:
> the Fedora images have a selinux-policy regression that causes all tests to fail

Gotcha.  So the Fedora/Ubuntu image split sounds like a temporary fix.  Would you mind adding a comment such as:

```yaml
    # TODO: At the time of this comment, an selinux-policy regression is blocking use of updated
    # Fedora VM images: <BZ/issue link>
```

That way it can be un-split when the issue/bug is fixed and images need updating again.

giuseppe:
addressed the comments :arrow_up:

rhatdan:
So we revert this PR or at least the separation of the VMs once selinux-policy gets updated.

cevich:
LGTM

cevich:
> So we revert this PR or at least the separation of the VMs once selinux-policy gets updated.

Just the one `.cirrus.yml: use c5521575421149184 for Ubuntu` commit.

But if this takes longer than 30-days, there's a very not-obvious problem WRT image pruning.  I'll open a followup PR so the tests don't all need to be re-run for it (here).

cevich:
...in case you want to include it in the one commit (would be convenient for reverting), this is what I'm thinking:

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index a403889c6..afeb5903d 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -674,10 +674,12 @@ meta_task:
     env:
         # Space-separated list of images used by this repository state
         IMGNAMES: >-
-            ${FEDORA_CACHE_IMAGE_NAME}
-            ${PRIOR_FEDORA_CACHE_IMAGE_NAME}
-            ${UBUNTU_CACHE_IMAGE_NAME}
-            ${PRIOR_UBUNTU_CACHE_IMAGE_NAME}
+            "fedora-${IMAGE_SUFFIX}"
+            "prior-fedora-${IMAGE_SUFFIX}"
+            "ubuntu-${IMAGE_SUFFIX}"
+            "prior-ubuntu-${IMAGE_SUFFIX}"
+            "ubuntu-${IMAGE_SUFFIX_UBUNTU}"
+            "prior-ubuntu-${IMAGE_SUFFIX_UBUNTU}"
         BUILDID: "${CIRRUS_BUILD_ID}"
         REPOREF: "${CIRRUS_REPO_NAME}"
         GCPJSON: ENCRYPTED[3a198350077849c8df14b723c0f4c9fece9ebe6408d35982e7adf2105a33f8e0e166ed3ed614875a0887e1af2b8775f4]
```

giuseppe:
thanks for the suggestion.  I've amended the change in the last version

cevich:
oof, my bad...

...in case you want to include it in the one commit (would be convenient for reverting), this is what I'm thinking:

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index a403889c6..026843503 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -678,6 +678,8 @@ meta_task:
             ${PRIOR_FEDORA_CACHE_IMAGE_NAME}
             ${UBUNTU_CACHE_IMAGE_NAME}
             ${PRIOR_UBUNTU_CACHE_IMAGE_NAME}
+            ubuntu-${IMAGE_SUFFIX}
+            prior-ubuntu-${IMAGE_SUFFIX}
         BUILDID: "${CIRRUS_BUILD_ID}"
         REPOREF: "${CIRRUS_REPO_NAME}"
         GCPJSON: ENCRYPTED[3a198350077849c8df14b723c0f4c9fece9ebe6408d35982e7adf2105a33f8e0e166ed3ed614875a0887e1af2b8775f4]
```

cevich:
Okay, welp, since we're re-spinning tests I can close my PR then.  Sorry for the noise/churn, I was trying to help avoid needing to re-run all the tests while there's the annoying system-test flake (assuming it's still a thing).

giuseppe:
looks like the test doesn't like the last diff:

```

ERROR: (gcloud.compute.images.update) HTTPError 400: Invalid value for field 'image': '"ubuntu-c5348179051806720"'. Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}'
�[1;33mDetected update error for '"ubuntu-c5348179051806720"'�[0m
ERROR: (gcloud.compute.images.update) HTTPError 400: Invalid value for field 'image': '"prior-ubuntu-c5348179051806720"'. Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}'
�[1;33mDetected update error for '"prior-ubuntu-c5348179051806720"'�[0m
�[1;31mERROR: ERROR: Failed to update one or more image timestamps:  "ubuntu-c5348179051806720" "prior-ubuntu-c5348179051806720"�[0m
```

giuseppe:
reverted it back, and the CI is smart to turn immediately green :-)

rhatdan:
LGTM

PR #10452, opened 2021-05-25
Fix race condition in running ls container in a pod

rhatdan:
All of the tests has an assumption that RunLsContainer and RunLsContainerInPod completes
the container before returning.  But since the container is running
in back ground mode, the container could be still running before tools
attempt to remove it. Removing the "-d" from the command fixes the
container to match the assumption.

### DIFF
--- a/test/e2e/common_test.go
+++ b/test/e2e/common_test.go
@@ -408,7 +408,14 @@ func (p *PodmanTestIntegration) RunLsContainer(name string) (*PodmanSessionInteg
 	podmanArgs = append(podmanArgs, "-d", ALPINE, "ls")
 	session := p.Podman(podmanArgs)
 	session.WaitWithDefaultTimeout()
-	return session, session.ExitCode(), session.OutputToString()
+	if session.ExitCode() != 0 {
+		return session, session.ExitCode(), session.OutputToString()
+	}
+	cid := session.OutputToString()
+
+	wsession := p.Podman([]string{"wait", cid})
+	wsession.WaitWithDefaultTimeout()
+	return session, wsession.ExitCode(), cid
 }
 
 // RunNginxWithHealthCheck runs the alpine nginx container with an optional name and adds a healthcheck into it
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago I am thinking this is the cause of most of the `podman pod` flakes.

vrothberg (commented):
LGTM

test/e2e/common_test.go
```diff
@@ -431,7 +438,14 @@ func (p *PodmanTestIntegration) RunLsContainerInPod(name, pod string) (*PodmanSe
 	podmanArgs = append(podmanArgs, "-d", ALPINE, "ls")
 	session := p.Podman(podmanArgs)
 	session.WaitWithDefaultTimeout()
-	return session, session.ExitCode(), session.OutputToString()
+	if session.ExitCode() != 0 {
+		return session, session.ExitCode(), session.OutputToString()
+	}
+	cid := session.OutputToString()
+
+	wsession = p.Podman([]string{"wait", cid})
```
edsantiago, line 446: Looks like this should be `:=`

rhatdan: Face slap.

rhatdan:
I think the podman start --all flake can be fixed just by adding this patch.  Will open a PR once this gets merged.

```
diff --git a/test/system/045-start.bats b/test/system/045-start.bats
index 542f9d1c2..3e0118dba 100644
--- a/test/system/045-start.bats
+++ b/test/system/045-start.bats
@@ -25,6 +25,8 @@ load helpers
         die "podman start --all restarted a running container"
     fi
 
+    run_podman wait $cid_none_implicit $cid_none_explicit $cid_on_failure
+
     run_podman rm $cid_none_implicit $cid_none_explicit $cid_on_failure
     run_podman stop -t 1 $cid_always
     run_podman rm $cid_always
```

rhatdan:
@edsantiago WDYT?

edsantiago:
The `start --all` flake is #10454, the one you and @mheon are actively discussing right now.

PR #10453, opened 2021-05-25
[CI:DOCS] docs: --cert-dir: point to containers-certs.d(5)

vrothberg:
Point to containers-certs.d(5) for details on the default paths, the
lookup logic and the structure of these directories.  Previously, the
man pages stated that the default path would be in `/etc/containers/...`
which is not entirely and a red herring for users (see #10116).


vrothberg:

rhatdan:
LGTM

Issue #10459, opened 2021-05-25
Data races related to libimage events


matejvasek:
After introduction of libimage events I can see dataraces caught by sanitizer:

```
podman pull quay.io/boson/faas-nodejs-builder                                                                                            ST 15   post-img-event 
==================
WARNING: DATA RACE
Write at 0x00c00001fd48 by main goroutine:
  github.com/containers/podman/v3/libpod.newRuntimeFromConfig()
      libpod/runtime.go:220 +0xb84
  github.com/containers/podman/v3/libpod.NewRuntime()
      libpod/runtime.go:163 +0x1dc
  github.com/containers/podman/v3/pkg/domain/infra.getRuntime()
      pkg/domain/infra/runtime_libpod.go:240 +0x38ed
  github.com/containers/podman/v3/pkg/domain/infra.GetRuntime.func1()
      pkg/domain/infra/runtime_libpod.go:79 +0x22f
  sync.(*Once).doSlow()
      /usr/lib/golang/src/sync/once.go:68 +0x137
  sync.(*Once).Do()
      /usr/lib/golang/src/sync/once.go:59 +0x7d
  github.com/containers/podman/v3/pkg/domain/infra.GetRuntime()
      pkg/domain/infra/runtime_libpod.go:78 +0x1b4
  github.com/containers/podman/v3/pkg/domain/infra.NewLibpodImageRuntime()
      pkg/domain/infra/runtime_proxy.go:24 +0xf8
  github.com/containers/podman/v3/pkg/domain/infra.NewImageEngine()
      pkg/domain/infra/runtime_abi.go:33 +0x175
  github.com/containers/podman/v3/cmd/podman/registry.NewImageEngine()
      cmd/podman/registry/registry.go:56 +0x144
  main.persistentPreRunE()
      cmd/podman/root.go:157 +0xfce
  github.com/spf13/cobra.(*Command).execute()
      vendor/github.com/spf13/cobra/command.go:831 +0xed4
  github.com/spf13/cobra.(*Command).ExecuteC()
      vendor/github.com/spf13/cobra/command.go:960 +0xac6
  github.com/spf13/cobra.(*Command).Execute()
      vendor/github.com/spf13/cobra/command.go:897 +0x56
  github.com/spf13/cobra.(*Command).ExecuteContext()
      vendor/github.com/spf13/cobra/command.go:890 +0xaf
  main.Execute()
      cmd/podman/root.go:90 +0xb9
  main.main()
[...]

PR #10466, opened 2021-05-26
libimage-events channel: fix data race

vrothberg:
Fix a data race between creating and using the libimage-events channel.

[NO TESTS NEEDED] since it really depends on the scheduler and we
couldn't hit the race so far.

Fixes: #10459

### DIFF
--- a/libpod/runtime.go
+++ b/libpod/runtime.go
@@ -217,8 +217,6 @@ func newRuntimeFromConfig(ctx context.Context, conf *config.Config, options ...R
 		return nil, err
 	}
 
-	runtime.libimageEventsShutdown = make(chan bool)
-
 	return runtime, nil
 }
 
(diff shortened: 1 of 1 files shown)


vrothberg:

vrothberg:
The start --all flake is hitting this PR hard.

libpod/runtime.go
```diff
@@ -701,6 +699,8 @@ var libimageEventsMap = map[libimage.EventType]events.Status{
 // events on the libimage.Runtime.  The gourtine will be cleaned up implicitly
 // when the main() exists.
 func (r *Runtime) libimageEvents() {
+	r.libimageEventsShutdown = make(chan bool)
```
matejvasek, line 702: Wouldn't be best do this at very creation of runtime like `newRuntimeFromConfig()` and maybe some tests? Then we might omit the nil checks because it's always initialized.

matejvasek:
Also there is similar problem with `eventChannel` in containers/common/libimage. IMHO the channel should be set once in `RuntimeFromStore()`.

vrothberg:
> Also there is similar problem with `eventChannel` in containers/common/libimage. IMHO the channel should be set once in `RuntimeFromStore()`.

That's only hypothetical. EventChannel is not meant to be used concurrently (see the comment). It wouldn't hurt to do it for sure.

Do you mind doing opening PRs to fix it?

rhatdan:
LGTM
@mheon @baude PTAL

PR #10476, opened 2021-05-26
Ensure that container still exists when removing

mheon:
After #8906, there is a potential race condition in container removal of running containers with `--rm`. Running containers must first be stopped, which was changed to unlock the container to allow commands like `podman ps` to continue to run while stopping; however, this also means that the cleanup process can potentially run before we re-lock, and remove the container from under us, resulting in error messages from `podman rm`. The end result is unchanged, the container is still cleanly removed, but the `podman rm` command will seem to have failed.

Work around this by pinging the database after we stop the container to make sure it still exists. If it doesn't, our job is done and we can exit cleanly.

[NO TESTS NEEDED] because the race doesn't trigger easily and I can't make a test to reproduce it on demand anywhere.

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -581,6 +581,15 @@ func (r *Runtime) removeContainer(ctx context.Context, c *Container, force, remo
 		if err := c.stop(c.StopTimeout()); err != nil && errors.Cause(err) != define.ErrConmonDead {
 			return errors.Wrapf(err, "cannot remove container %s as it could not be stopped", c.ID())
 		}
+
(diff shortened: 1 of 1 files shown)


rhatdan:
Is this to fix THE race?
LGTM

mheon:
Nope. Separate BZ, unfortunately.

Would not be surprised if it fixes some races in CI, though.

TomSweeneyRedHat:
FWIW, addresses https://bugzilla.redhat.com/show_bug.cgi?id=1964852
LGTM

vrothberg (commented):
/hold cancel

LGTM

Issue #10435, opened 2021-05-21
segfault when LISTEN_PID and LISTEN_FDS are set but not LISTEN_FDNAMES


eriksjolund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
A segmentation fault can be triggered by setting the environment variables 
`LISTEN_PID` and `LISTEN_FDS` but not `LISTEN_FDNAMES` and then `podman run ...`

It seems the relevant code is 
https://github.com/containers/podman/blob/fe0595683a05c0f0307e07f0cce29f6ae0b3d310/pkg/rootless/rootless_linux.c#L314

**Steps to reproduce the issue:**

1. Create the file _bug.bash_ with this content
    ```
    #!/bin/bash
    export LISTEN_PID=$$
    export LISTEN_FDS=1
    exec podman run --rm registry.fedoraproject.org/fedora:34 /bin/true
    ```

2. Run `bash bug.sh`

**Describe the results you received:**

```
Segmentation fault (core dumped)
```

**Output of `podman version`:**

```
Version:      3.2.0-dev
API Version:  3.2.0-dev
Go Version:   go1.15.3
Git Commit:   fe0595683a05c0f0307e07f0cce29f6ae0b3d310
Built:        Thu May 20 14:11:25 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug | grep -v hostname:`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.2-dev
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: Unknown
    path: /var/data/podman-installations/build-podman_601505070e4cf66a37d42a81156e5508d0cfb79c__centos8__fe0595683a05c0f0307e07f0cce29f6ae0b3d310__v2.0.27__v0.9.1__1.15.3__0.19.1__v1.1.10__v1.5.0/bin/conmon
    version: 'conmon version 2.0.27, commit: 65fad4bfcb250df0435ea668017e643e7f462155'
  cpus: 8
  distribution:
    distribution: '"centos"'
    version: "8"
  eventLogger: journald
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 100
      size: 1
    - container_id: 1
      host_id: 231072
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1626
      size: 1
    - container_id: 1
      host_id: 231072
      size: 65536
[...]

PR #10481, opened 2021-05-27
rootless: fix SIGSEGV, Make `LISTEN_FDNAMES` optional

flouthoc:
Reproduce SIGSEGV

```bash
#!/bin/bash
export LISTEN_PID=$$
export LISTEN_FDS=1
exec podman run --rm registry.fedoraproject.org/fedora:34 /bin/true
```
**[NO TESTS NEEDED]:** I think test for this would be a basic `run` with `LISTEN_PID` and `LISTEN_FDS` set and since no new feature or significant behavior change is being added so putting more tests might not be necessary. But sure let me know if having new tests here makes sense.
Closes: https://github.com/containers/podman/issues/10435

### DIFF
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -311,10 +311,10 @@ static void __attribute__((constructor)) init()
         do_socket_activation = true;
         saved_systemd_listen_pid = strdup(listen_pid);
         saved_systemd_listen_fds = strdup(listen_fds);
-        saved_systemd_listen_fdnames = strdup(listen_fdnames);
+        if (listen_fdnames != NULL)
+          saved_systemd_listen_fdnames = strdup(listen_fdnames);
         if (saved_systemd_listen_pid == NULL
-                || saved_systemd_listen_fds == NULL
-                || saved_systemd_listen_fdnames == NULL)
+                || saved_systemd_listen_fds == NULL)
           {
             fprintf (stderr, "save socket listen environments error: %s\n", strerror (errno));
             _exit (EXIT_FAILURE);
@@ -700,7 +700,9 @@ reexec_userns_join (int pid_to_join, char *pause_pid_file_path)
       sprintf (s, "%d", getpid());
       setenv ("LISTEN_PID", s, true);
       setenv ("LISTEN_FDS", saved_systemd_listen_fds, true);
-      setenv ("LISTEN_FDNAMES", saved_systemd_listen_fdnames, true);
+      // Setting fdnames is optional for systemd_socket_activation
+      if (saved_systemd_listen_fdnames != NULL)
+        setenv ("LISTEN_FDNAMES", saved_systemd_listen_fdnames, true);
     }
 
   setenv ("_CONTAINERS_USERNS_CONFIGURED", "init", 1);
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

rhatdan:
@giuseppe shouldn't we be saving the ones that are set.  LISTEN_FDNAMES is optional, so we should just handle that case.

giuseppe:
> @giuseppe shouldn't we be saving the ones that are set. LISTEN_FDNAMES is optional, so we should just handle that case.

ah right, yes in that case we need to handle `LISTEN_FDNAMES` to be null.

@flouthoc could you please fix it in the patch?

flouthoc:
@rhatdan @giuseppe Sure makes sense, we could make this optional instead.

flouthoc:
@giuseppe @rhatdan I have made requested changes. Could you please take a look ?

rhatdan:
LGTM

eriksjolund:
Thanks @flouthoc for fixing the issue!

It seems there is no reference in the commit message to the Github issue.

https://github.com/containers/podman/blob/master/CONTRIBUTING.md
_PRs that fix issues should include a reference like Closes #XXXX in the commit message so that GitHub will automatically close the referenced issue when the PR is merged._

Another thing, regarding:
"Ref: #10435"
in the [comment](https://github.com/containers/podman/pull/10481#issue-654526199).
_Ref_ does not seem to be one of the keywords that links the pull request to an issue:
https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

flouthoc:
@eriksjolund thanks for pointing this out 😄 .Adding it now Closes: #10435

eriksjolund:
Thanks @flouthoc Could you also add 
```
Closes #10435
```
in the commit message? (The information will then be preserved in the commit log. Another advantage is that it will help GitHub provide a convenient hyperlink to the issue when displaying the Git commit on their website)

Running `git log` I see that there are also other styles that works as well, for instance

```
Closes: https://github.com/containers/podman/issues/10435
```

flouthoc:
@eriksjolund Done. Let me know if anything else needs to be done. 🙂

eriksjolund:
@flouthoc Thanks, to me it looks fine. I now see the GitHub hyperlink in https://github.com/flouthoc/podman/commit/2addc0f90d7134158f52124fc8d0ae75691d947b

pkg/rootless/rootless_linux.c
```diff
@@ -311,10 +311,10 @@ static void __attribute__((constructor)) init()
         do_socket_activation = true;
         saved_systemd_listen_pid = strdup(listen_pid);
         saved_systemd_listen_fds = strdup(listen_fds);
-        saved_systemd_listen_fdnames = strdup(listen_fdnames);
+        if (listen_fdnames != NULL)
+          saved_systemd_listen_fdnames = strdup(listen_fdnames);
```
eriksjolund, line 315:
```suggestion
        saved_systemd_listen_fdnames = NULL;
        if (listen_fdnames != NULL)
          saved_systemd_listen_fdnames = strdup(listen_fdnames);
```

Wouldn't it otherwise lead to undefined behavior when it's later used? I guess the initializing could be done in different ways (for instance at the same time as it is declared).



Another thing I thought about is that there is no `free()` of the `strdup()` strings.

Quote from `man strdup()`:
_Memory for the new string is obtained with malloc(3), and can be freed with free(3)._

Maybe that is not so important though.

PR #10489, opened 2021-05-27
create libimage-events channel in main routine

vrothberg:
Move the creation of the channel outside of the sub-routine to fix a
data race between writing the channel (implicitly by calling
EventChannel()) and using that channel in libimage.

[NO TESTS NEEDED]

Fixes: #10459

@matejvasek @containers/podman-maintainers PTAL


matejvasek:
lgtm

PR #10508, opened 2021-05-31
Better error handing for images/create compat api

matejvasek:
If server cannot resolve short name due to enforcing policy send
non-2xx status code.

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/common/libimage"
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/shortnames"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v3/libpod"
 	"github.com/containers/podman/v3/pkg/api/handlers"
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/compat/images.go
```diff
@@ -231,6 +232,12 @@ func CreateImageFromImage(w http.ResponseWriter, r *http.Request) {
 
 	fromImage := mergeNameAndTagOrDigest(query.FromImage, query.Tag)
 
+	_, err := shortnames.Resolve(runtime.SystemContext(), fromImage)
```
vrothberg, line 235: Can you drop a comment and elaborate why do an early check here?

Future readers may feel inclined to drop the check.

vrothberg (commented):
Code LGTM

pkg/api/handlers/compat/images.go
```diff
@@ -231,6 +232,14 @@ func CreateImageFromImage(w http.ResponseWriter, r *http.Request) {
 
 	fromImage := mergeNameAndTagOrDigest(query.FromImage, query.Tag)
 
+	// without this early check this function would return 200 but reported error via body stream soon after
+	// it's better to let caller know early via HTTP status code that request cannot be processed
+	_, err := shortnames.Resolve(runtime.SystemContext(), fromImage)
+	if err != nil {
+		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrap(err, "failed to resolve image name"))
```
matejvasek, line 239: Maybe `403 Forbidden` would be better?

vrothberg: That's @jwhonce question :) He's the Yoda of REST APIs.

matejvasek: Thinking about it I cannot be sure that this failed due to enforcing mode. Maybe it should be internal server error after all.

rhatdan:
/approve
/lgtm
Will allow @jwhonce Merge.

Issue #10493, opened 2021-05-28
Use Systemd specifiers in the line "RequiresMountsFor=" from "podman generate systemd --new"


eriksjolund:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

`podman generate systemd --new`  does not make use of Systemd specifiers in the line that starts with `RequiresMountsFor=`.

By changing for instance

```
RequiresMountsFor=/home/tmptest/.local/share/containers/storage /run/user/1003/containers
```
to
```
RequiresMountsFor=%h/.local/share/containers/storage %t/containers
```
the unit file would be made more portable and could be reused by another user. (If the other user would have a non-standard Podman configuration it would not work but that would be an edge case).

The list of available Systemd specifiers can be seen in _Table 4. Specifiers available in unit files_ in 
https://www.freedesktop.org/software/systemd/man/systemd.unit.html

**Steps to reproduce the issue:**

1. Create the new user _tmptest_ and log in to that account

2. Create the container _foobar_

    ```
    [tmptest@laptop ~]$ podman create --name foobar docker.io/library/alpine
    Trying to pull docker.io/library/alpine:latest...
    Getting image source signatures
    Copying blob 540db60ca938 done  
    Copying config 6dbb9cc540 done  
    Writing manifest to image destination
    Storing signatures
    71bf99fa5a7c6056ca5a203b0734162d1289db0eef6b17a9c55d6483b7d1bcac
    [tmptest@laptop ~]$ 
    ```

3. Generate a Systemd user service

    ```
    [tmptest@laptop ~]$ podman generate systemd --new --name foobar
    # container-foobar.service
    # autogenerated by Podman 3.1.2
    # Fri May 28 07:19:37 CEST 2021

    [Unit]
    Description=Podman container-foobar.service
    Documentation=man:podman-generate-systemd(1)
    Wants=network.target
    After=network-online.target
    RequiresMountsFor=/home/tmptest/.local/share/containers/storage /run/user/1003/containers

    [Service]
    Environment=PODMAN_SYSTEMD_UNIT=%n
    Restart=on-failure
    TimeoutStopSec=70
    ExecStartPre=/bin/rm -f %t/container-foobar.pid %t/container-foobar.ctr-id
[...]

PR #10509, opened 2021-05-31
[v3.2] generate systemd: make mounts portable

vrothberg:
Commit 748826fc88fc fixed a bug where slow mounting of the runroot was
causing issues when the units are started at boot.  The fix was to add
the container's runroot to the required mounts; the graph root has been
added as well.

Hard-coding the run- and graphroot to the required mounts, however,
breaks the portability of units generated with --now.  Those units are
intended to be running on any machine as, theoreticaly, any user.

Make the mounts portable by using the `%t` macro for the run root.
Since the graphroot's location varies across root and ordinary users,
drop it from the list of required mounts.  The graphroot was not causing
issues.

Fixes: #10493


vrothberg:

mheon:
Sure, LGTM

PR #10510, opened 2021-05-31
Use request context instead of background context

matejvasek:
This fixes a goroutine leak.
Previously if a request was cancelled the image pulling goroutine kept running regardless.

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -1,7 +1,6 @@
 package compat
 
 import (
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
Please update the commit message to describe what the problem is and what the solution.  It's very hard to match the code changes to the intention.

matejvasek:
@vrothberg better now?

vrothberg (commented):
LGTM, thanks

Issue #10507, opened 2021-05-31
The events API is not Docker-compatible when using the same filter multiple times


ahmed-mez:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Podman's events API doesn't handle the case where the same filter is used multiple times like Docker does.

Docker documentation: https://docs.docker.com/engine/reference/commandline/events/

> Using the same filter multiple times will be handled as a OR; for example --filter container=588a23dac085 --filter container=a8f7720b8c22 will display events for container 588a23dac085 OR container a8f7720b8c22

However, Podman will still apply an AND instead of an OR for the example above.

**Steps to reproduce the issue:**

1. Run two containers

```bash
podman run -d nginx
80fe6eb5f585dde552491f3b0c77751a19045e99211c0d72c539833859f9b5d3

podman run -d redis
1c62ec17cb44e183f83db5261551bd0d5c0db75e0973e8bbf32c97e41b5700a1

podman ps
CONTAINER ID  IMAGE                           COMMAND               CREATED         STATUS             PORTS   NAMES
80fe6eb5f585  docker.io/library/nginx:latest  nginx -g daemon o...  13 seconds ago  Up 13 seconds ago          cranky_banzai
1c62ec17cb44  docker.io/library/redis:latest  redis-server          5 seconds ago   Up 5 seconds ago           gracious_khorana
```

2. Query events for both containers (doesn't work)

```bash
podman events --since 5m --filter "container=80fe6eb5f585" --filter "container=1c62ec17cb44"
# No events shown!
```

3. Query events for one container (works)

```bash
podman events --since 5m --filter "container=1c62ec17cb44"
2021-05-31 09:36:09.923083952 +0000 UTC container create 1c62ec17cb44e183f83db5261551bd0d5c0db75e0973e8bbf32c97e41b5700a1 (image=docker.io/library/redis:latest, name=gracious_khorana)
2021-05-31 09:36:10.200379849 +0000 UTC container init 1c62ec17cb44e183f83db5261551bd0d5c0db75e0973e8bbf32c97e41b5700a1 (image=docker.io/library/redis:latest, name=gracious_khorana)
2021-05-31 09:36:10.212573077 +0000 UTC container start 1c62ec17cb44e183f83db5261551bd0d5c0db75e0973e8bbf32c97e41b5700a1 (image=docker.io/library/redis:latest, name=gracious_khorana)
```

**Describe the results you received:**

[...]

PR #10514, opened 2021-06-01
events: support disjunctive filters

vrothberg:
While different filters are applied in conjunction, the same filter (but
with different values) should be applied in disjunction.  This allows,
for instance, to query the events of two containers.

Fixes: #10507

### DIFF
--- a/test/system/090-events.bats
+++ b/test/system/090-events.bats
@@ -61,3 +61,32 @@ load helpers
 .*image remove $imageID $tag.*" \
        "podman events"
 }
+
+function _events_disjunctive_filters() {
+    local backend=$1
+
+    # Regression test for #10507: make sure that filters with the same key are
+    # applied in disjunction.
+    t0=$(date --iso-8601=seconds)
+    run_podman $backend run --name foo --rm $IMAGE ls
+    run_podman $backend run --name bar --rm $IMAGE ls
+    run_podman $backend events --stream=false --since=$t0 --filter container=foo --filter container=bar --filter event=start
+    is "$output" ".* container start .* name=foo.*
+.* container start .* name=bar.*"
+}
+
+@test "events with disjunctive filters - file" {
+    skip_if_remote "remote does not support --events-backend"
+    _events_disjunctive_filters --events-backend=file
+}
+
+@test "events with disjunctive filters - journald" {
+    skip_if_remote "remote does not support --events-backend"
+    _events_disjunctive_filters --events-backend=journald
+}
+
+@test "events with disjunctive filters - default" {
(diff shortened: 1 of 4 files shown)


vrothberg:
Would love to have it in v3.2.

vrothberg:

mheon:
Little late for 3.2, unfortunately

test/system/090-events.bats
```diff
@@ -61,3 +61,34 @@ load helpers
...
+    _events_disjunctive_filters --events-backend=file
+}
+
+@test "events with disjunctive filters - journald" {
+    skip_if_remote "remote does not support --events-backend"
+    _events_disjunctive_filters --events-backend=journald
+}
+
+@test "events with disjunctive filters - remote only" {
+    skip "FIXME: remote events seems to eat events"
```
edsantiago, line 89: Could you include the issue number here? That will make it easy for my `find-obsolete-skips` script to flag about this when the issue is fixed.

test/system/090-events.bats
```diff
@@ -61,3 +61,34 @@ load helpers
...
+    # Since podman-remote does not support the --events-backend flag, do a
+    # remote-only test with the default events backend.
+    skip_if_local "test is intended to only run on remote"
```
edsantiago, line 92: Why the skip? What harm does it do to test the default events backend when running local? It actually seems like a kind of important thing to test.

test/system/090-events.bats
vrothberg: Good idea! I am not sure if there is an issue for it yet. Will have a look and open a new one if needed.

vrothberg: https://github.com/containers/podman/issues/10529

vrothberg:
@edsantiago @rhatdan PTAL

test/system/090-events.bats
edsantiago: >The default is implicitly tested by the two tests above

Quick clarification: *Today's* default is *explicitly* tested. Tomorrow, someone may change the default to `--events-backend=something-new`. Or, the default might actually not be implemented correctly, or it might break in the future. Paranoia is a good thing sometimes: it would be really embarrassing to have the default case break, and ship!

vrothberg: > Or, the default might actually not be implemented correctly, or it might break in the future. Paranoia is a good thing sometimes: it would be really embarrassing to have the default case break, and ship!

Sounds good. Changed :+1:

vrothberg:
Ready

edsantiago:
tests LGTM. I should've griped about the use of fixed non-random strings, but that's not worth a re-push.

PR #10515, opened 2021-06-01
[v3.0.1] set version to v3.0.1

vrothberg:
The v3.0.1 branch included the bump to v3.0.2-dev which indicates the
state in between two releases.

Fixes: bugzilla.redhat.com/show_bug.cgi?id=1966538

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -8,7 +8,7 @@ import (
 // NOTE: remember to bump the version at the top
 // of the top-level README.md file when this is
 // bumped.
-var Version = semver.MustParse("3.0.2-dev")
+var Version = semver.MustParse("3.0.1")
 
 // APIVersion is the version for the remote
 // client API.  It is used to determine compatibility


vrothberg:
/hold
Not yet acked.

mheon:
I wonder if this would cause any RPM issues - is the package that went out tagged as 3.0.2, not 3.0.1? If so we probably need to bump to 3.0.2 here to avoid a situation where we've "downgraded" the RHEL package.

vrothberg:
> I wonder if this would cause any RPM issues - is the package that went out tagged as 3.0.2, not 3.0.1? If so we probably need to bump to 3.0.2 here to avoid a situation where we've "downgraded" the RHEL package.

The package is podman-3.0.1-6.module+el8.4.0+10607+f4da7515.ppc64le, so I think we're good.

mheon:
OK, good.

TomSweeneyRedHat:
LGTM
@mheon @rhatdan, OK to cancel the hold on this?  Green buttons everywhere.

TomSweeneyRedHat:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1966538

PR #10517, opened 2021-06-01
API one-shot query implementation/handling

cdoern:
containers_stats.go was missing the handling of the one-shot query parameter. Modified the code to handle the incoming query as well as return an error if there is a mismatch of stream and one-shot (stream must be false if one shot is true). One shot skips the preCPU stats and returns a singular cycle's statistics. Also added documentation in register_containers.go

### DIFF
--- a/pkg/api/handlers/compat/containers_stats.go
+++ b/pkg/api/handlers/compat/containers_stats.go
@@ -22,12 +22,17 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 
 	query := struct {
-		Stream bool `schema:"stream"`
+		Stream  bool `schema:"stream"`
+		OneShot bool `schema:"one-shot"` //added schema for one shot
 	}{
 		Stream: true,
 	}
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
+		return
+	}
+	if query.Stream && query.OneShot { // mismatch. one-shot can only be passed with stream=false
+		utils.Error(w, "invalid combination of stream and one-shot", http.StatusBadRequest, define.ErrInvalidArg)
 		return
 	}
 
@@ -56,6 +61,16 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	coder := json.NewEncoder(w)
+	// Write header and content type.
+	w.WriteHeader(http.StatusOK)
+	w.Header().Add("Content-Type", "application/json")
+	if flusher, ok := w.(http.Flusher); ok {
+		flusher.Flush()
+	}
+
+	// Setup JSON encoder for streaming.
+	coder.SetEscapeHTML(true)
 	var preRead time.Time
 	var preCPUStats CPUStats
 	if query.Stream {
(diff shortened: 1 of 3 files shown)


zhangguanzhang:
should rebase to one commit and `git commit -s`

cdoern:
> should rebase to one commit and `git commit -s`

thank you, my first PR as an intern so I wasn't sure how to properly format.

pkg/api/handlers/compat/containers_stats.go
```diff
@@ -22,14 +22,19 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 
 	query := struct {
-		Stream bool `schema:"stream"`
+		Stream  bool `schema:"stream"`
+		oneShot bool `schema:"one-shot"` //added schema for one shot
```
jwhonce, line 26:
```suggestion
		OneShot bool `schema:"one-shot"` //added schema for one shot
```
Field needs to be exported.

pkg/api/handlers/compat/containers_stats.go
```diff
@@ -22,14 +22,19 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
...
 	}{
 		Stream: true,
 	}
 	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
 		return
 	}
+	if query.Stream && query.oneShot { // mismatch. one-shot can only be passed with no-stream=true
```
jwhonce, line 34:
```suggestion
	if query.Stream && query.oneShot { // mismatch. one-shot can only be passed with stream=false
```
`no-stream` is not a valid query parameter.

pkg/api/handlers/compat/containers_stats.go
```diff
@@ -22,14 +22,19 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
...
+		utils.Error(w, "invalid combination of stream and one-shot", http.StatusConflict, define.ErrInvalidArg)
```
jwhonce, line 35:
```suggestion
		utils.Error(w, "invalid combination of stream and one-shot", http.StatusBadRequest, define.ErrInvalidArg)
```

pkg/api/handlers/compat/containers_stats.go
```diff
@@ -56,8 +61,22 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
...
+	coder := json.NewEncoder(w)
+	// Write header and content type.
+	w.WriteHeader(http.StatusOK)
+	w.Header().Add("Content-Type", "application/json")
+	if flusher, ok := w.(http.Flusher); ok {
+		flusher.Flush()
+	}
+
+	// Setup JSON encoder for streaming.
+	//coder := json.NewEncoder(w)
```
jwhonce, line 73:
```suggestion
```

pkg/api/handlers/compat/containers_stats.go
```diff
@@ -56,8 +61,22 @@ func StatsContainer(w http.ResponseWriter, r *http.Request) {
...
+	coder.SetEscapeHTML(true)
 	var preRead time.Time
 	var preCPUStats CPUStats
+	if query.oneShot { // if one shot response skip initial read and return one round results
+		goto streamLabel
+	}
```
jwhonce, line 79:
```suggestion
```
Is this test needed? From your check above, `query.Stream` must be false if `query.OneShot` is true.

jwhonce (changes requested):

jwhonce:
@cdoern You also need to add some tests.

pkg/api/handlers/compat/containers_stats.go
cdoern: good point, I think I accidentally left this one in from when I had implemented it slightly differently.

cdoern:
> @cdoern You also need to add some tests.

how would I go about doing that?

rhatdan:
Nice work @cdoern 
/approve

jwhonce (approved):

TomSweeneyRedHat:
LGTM
will let @jwhonce push the shiny merge button.

Issue #10495, opened 2021-05-28
API list networks returns "null" instead of empty array when used with no networks


arctic-alpaca:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The [list volumes API endpoint](https://docs.podman.io/en/latest/_static/api.html#operation/VolumeListLibpod) returns `[]` when called with no volumes, [list networks](https://docs.podman.io/en/latest/_static/api.html#operation/NetworkListLibpod) returns `null` when used with no networks. According to the API docs, the behavior should be similar.

**Steps to reproduce the issue:**

1. ``podman system service unix:///home/`whoami`/testing.sock --log-level=debug --time=500``

2. ```
   curl --unix-socket /home/`whoami`/testing.sock http://d/v3.2.0/libpod/networks/json
   null
   ```

**Describe the results you received:**
`null`


**Describe the results you expected:**
`[]` - an emtpy array.


**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.2.0-dev
API Version:  3.2.0-dev
Go Version:   go1.15.10
Git Commit:   6ff56ab50a10e2350013c4ff1bf4b8d5d7b5aa87-dirty
Built:        Tue Jan  1 01:00:00 1980
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.1-dev
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 12
  distribution:
    distribution: debian
    version: "10"
  eventLogger: file
  hostname: DESKTOP
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.4.72-microsoft-standard-WSL2
  linkmode: static
  memFree: 12503470080
  memTotal: 13346447360
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.18.1-7931a-dirty
[...]

PR #10518, opened 2021-06-01
  [NO TESTS NEEDED] Fixes API list networks returns "null" instead of empty array when used with no networks

zhangguanzhang:
API list networks returns "null" instead of empty array when used with no networks
Fixes: https://github.com/containers/podman/issues/10495

[NO TESTS NEEDED]


rhatdan:
/approve
LGTM

PR #10519, opened 2021-06-01
Use request context instead of background

matejvasek:
This prevents goroutine leak:
If background context were used then push operation would continue even
if client aborted request by closing connection.

### DIFF
--- a/pkg/api/handlers/compat/images_push.go
+++ b/pkg/api/handlers/compat/images_push.go
@@ -1,7 +1,6 @@
 package compat
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM, thanks!

pkg/api/handlers/compat/images_push.go
```diff
@@ -160,43 +145,50 @@ loop: // break out of for/select infinite loop
 			}
 			report.ID = e.Artifact.Digest.Encoded()[0:12]
 			if err := enc.Encode(report); err != nil {
-				errorWriter.Write([]byte(err.Error()))
+				logrus.Warnf("Failed to json encode error %q", err.Error())
```
mheon, line 148: Minor nit, but if this needs other changes, can you swap this to an `Errorf`?

rhatdan:
/lgtm
Lets do this in a separate PR

PR #10524, opened 2021-06-01
[CI:DOCS] Fix link error

rugk:
After all, maybe using numbers in the FAQ is not such a good idea for having persistent links… :wink:


rhatdan:
/approve
LGTM
Thanks @rugk

Issue #10145, opened 2021-04-27
Document the single ID rootless mode


andrewshadura:
The single ID rootless mode is not documented properly. It is possible to use rootless, but limited to just one user, together with `ignore_chown_errors`:
> `podman --storage-opt ignore_chown_errors ...` when a single ID is available and all the files will be owned by just one user.
> 
> It is not done automatically (and I think it shouldn't) because potentially it can lead to security problems when the files ownership inside the container matters.

_Originally posted by @giuseppe in https://github.com/containers/podman/issues/9788#issuecomment-806684711_

rhatdan:
Interested in opening a PR?

Should this be documented in the podman run man page, or in some other location?

PR #10525, opened 2021-06-01
[CI:DOCS] Add documentation on ignore_chown_errors

rhatdan:
fixes: https://github.com/containers/podman/issues/10145


zhangguanzhang:
title should start with `[CI:DOCS]`

zhangguanzhang:
the step may be need re-run

rhatdan:
Thanks @zhangguanzhang Already fixed, Since the docs still need to be worked on I will repush, once this is reviewed.

docs/source/markdown/podman.1.md
```diff
@@ -336,9 +336,11 @@ Images are pulled under `XDG_DATA_HOME` when specified, otherwise in the home di
...
+The Overlay file system (OverlayFS) is not supported with older kernels in rootless mode.  The fuse-overlayfs package is a tool that provides the functionality of OverlayFS in user namespace that allows mounting file systems in rootless environments.  It is recommended to install the fuse-overlayfs package.  In rootless mode Podman will automatically use the fuse-overlafs program as the mount_program if installed, as long as the $HOME/.config/containers/storage.conf file was not previously created.  If storage.conf exists in the homedir, add `mount_program = "/usr/bin/fuse-overlayfs"` under `[storage.options.overlay]` to enable this feature.
```
edsantiago, line 343: Any time I see "older", "newer", "recent", I ask myself: how will this read in six/twelve months? Would it be safe to say here, "with kernels prior to 5.12.9"?

docs/source/markdown/podman.1.md
```diff
@@ -336,9 +336,11 @@ Images are pulled under `XDG_DATA_HOME` when specified, otherwise in the home di
 
 Currently the slirp4netns package is required to be installed to create a network device, otherwise rootless containers need to run in the network namespace of the host.
 
+In certain enviroments like HPC (High Performance Computing) users are not allowed to take advantage of the additional UIDs and GIDs from the /etc/subuid and /etc/subgid systems.  Even in this environment you are allowed to use rootless podman with a single UID.  You need to set the `ignore_chown_errors` option in the /etc/containers/storage.conf or in ~/.config/containers/storage.conf. This option tells podman when pulling an image to ignore chown errors when it attempts to change a file in a container image to match the non root UID in the image. This means all files get saved as the user's UID. Note this could cause issues when running the container.
```
TomSweeneyRedHat, line 339:
```suggestion
In certain environments like HPC (High Performance Computing), users cannot take advantage of the additional UIDs and GIDs from the /etc/subuid and /etc/subgid systems.  However, in this environment, rootless Podman can operate with a single UID.  To make this work, set the `ignore_chown_errors` option in the /etc/containers/storage.conf or in ~/.config/containers/storage.conf files. This option tells Podman when pulling an image to ignore chown errors when attempting to change a file in a container image to match the non-root UID in the image. This means all files get saved as the user's UID. Note this could cause issues when running the container.
```

docs/source/markdown/podman.1.md
TomSweeneyRedHat, line 343:
```suggestion
The Overlay file system (OverlayFS) is not supported with kernels prior to 5.12.9 in rootless mode.  The fuse-overlayfs package is a tool that provides the functionality of OverlayFS in user namespace that allows mounting file systems in rootless environments.  It is recommended to install the fuse-overlayfs package.  In rootless mode, Podman will automatically use the fuse-overlayfs program as the mount_program if installed, as long as the $HOME/.config/containers/storage.conf file was not previously created.  If storage.conf exists in the homedir, add `mount_program = "/usr/bin/fuse-overlayfs"` under `[storage.options.overlay]` to enable this feature.
```

rhatdan: No because it will be backported into RHEl8.

rhatdan: How about if we just say `is not supported in all kernels ...`

TomSweeneyRedHat (commented):
LGTM
and happy green test buttons

PR #10526, opened 2021-06-01
[CI:DOCS] UPDATE MANPAGE_SYNTAX (commit,attach,auto-update)

Procyhon:
Updated version for the MANPAGE_SYNTAX and adaption of the syntax for
the manpages of podman-commit, podman-attach, and podman-auto-update.


vrothberg (commented):
LGTM, thank you!

docs/source/markdown/podman-commit.1.md
```diff
@@ -9,58 +9,52 @@ podman\-commit - Create new image based on the changed container
...
-#### **--include-volumes**
+#### **--include-volumes** [=**false** | *true*]
 
-Include in the committed image any volumes added to the container by the `--volume` or `--mount` options to the `podman create` and `podman run` commands.
+Include in the committed image any volumes added to the container by the **--volume** or **--mount** OPTIONS to the [**podman create**](podman-create.1.md) and [**podman run**](podman-run.1.md) commands.
 
 #### **--message**, **-m**=*message*
 
-Set commit message for committed image.  The message field is not supported in _oci_ format.
+Set commit message for committed image.\
```
rhatdan, line 44: This line break "\" ends up showing up in the man pages.

docs/source/markdown/podman-commit.1.md
```diff
@@ -9,58 +9,52 @@ podman\-commit - Create new image based on the changed container
...
 
 #### **--iidfile**=*ImageIDfile*
 
 Write the image ID to the file.
 
-#### **--include-volumes**
+#### **--include-volumes** [=**false** | *true*]
 
-Include in the committed image any volumes added to the container by the `--volume` or `--mount` options to the `podman create` and `podman run` commands.
+Include in the committed image any volumes added to the container by the **--volume** or **--mount** OPTIONS to the [**podman create**](podman-create.1.md) and [**podman run**](podman-run.1.md) commands.
```
rhatdan, line 40: Sadly these [] end up showing up in man page, but I am willing to live with this.

docs/source/markdown/podman-commit.1.md
```diff
@@ -9,58 +9,52 @@ podman\-commit - Create new image based on the changed container
...
 
-Can be set multiple times
+Can be set multiple times.
 
-#### **--format**, **-f**=*format*
+#### **--format**, **-f** =**oci** | *docker*
 
-Set the format of the image manifest and metadata.  The currently supported formats are _oci_ and _docker_.  If
-not specifically set, the default format used is _oci_.
+Set the format of the image manifest and metadata.  The currently supported formats are **oci** and *docker*. The default is **oci**.
```
rhatdan, line 32: ``` *docker* should be **docker**```

docs/source/markdown/podman-attach.1.md
```diff
@@ -9,48 +9,52 @@ podman\-attach - Attach to a running container
...
-The attach command allows you to attach to a running container using the container's ID
-or name, either to view its ongoing output or to control it interactively.
+**podman attach** allows you to attach to a running *container* using the *container's name* or *ID*, either to view its ongoing output or to control it interactively.
 
-You can detach from the container (and leave it running) using a configurable key sequence. The default
+You can detach from the *container* (and leave it running) using a configurable key sequence. The default
 sequence is `ctrl-p,ctrl-q`.
 Configure the keys sequence using the **--detach-keys** option, or specifying
-it in the **containers.conf** file: see **containers.conf(5)** for more information.
+it in the `containers.conf` file: see [**containers.conf(5)**](https://github.com/containers/common/blob/master/docs/containers.conf.5.md) for more information.
```
rhatdan, line 17: This link is showing up in the actual man page.

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -7,44 +7,41 @@ podman-auto-update - Auto update containers according to their auto-update polic
...
-This variable is now set by all systemd units generated by `podman-generate-systemd` and is set to `%n` (i.e., the name of systemd unit starting the container).
+At container-creation time, Podman looks up the `PODMAN_SYSTEMD_UNIT` environment variables and stores it verbatim in the container's label.
+This variable is now set by all systemd units generated by [**podman-generate-systemd**](podman-generate-systemd.1.md) and is set to `%n` (i.e., the name of systemd unit starting the container).
 This data is then being used in the auto-update sequence to instruct systemd (via DBUS) to restart the unit and hence to restart the container.
 
-Note that `podman auto-update` relies on systemd. The systemd units are expected to be generated with `podman-generate-systemd --new`, or similar units that create new containers in order to run the updated images.
+Note that **podman auto-update** relies on systemd. The systemd units are expected to be generated with [**podman-generate-systemd --new**](podman-generate-systemd.1.md#--new), or similar units that create new containers in order to run the updated images.
```
rhatdan, line 30: Link shows up in the man page.

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -7,44 +7,41 @@ podman-auto-update - Auto update containers according to their auto-update polic
...
-
+Podman ships with a `podman-auto-update.service` systemd unit. This unit is triggered daily at midnight by the `podman-auto-update.timer` systemd timer.  The timer can be altered for custom time-based updates if desired.  The unit can further be invoked by other systemd units (e.g., via the dependency tree) or manually via **systemctl start podman-auto-update.service**.
 
 ## OPTIONS
 
 #### **--authfile**=*path*
 
-Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
-If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+Path of the authentication file. Default is `${XDG_RUNTIME_DIR}/containers/auth.json`, which is set using [**podman login**](podman-login.1.md).
```
rhatdan, line 41: Brackets around [podman login] show up in man page.

rhatdan (approved):

rhatdan:
@Procyhon Thanks this is great, but Please review the man page that is generated.  Most users of Podman are going to be using the man page, rather then viewing them in markdown, since go-md2man is not perfect, we need to make sure that the generated man pages still look good, and do not have embeded links viewable to the user.  In most cases I would prefer the markdown to be worse then make the man pages look bad.

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
 **podman subcommand command** [*optional*] *value* [*value* ...]
 
 ## DESCRIPTION
-**podman command** is always the beginning of the DESCRIPTION section. Putting the command as the first part of the DESCRIPTION ensures uniformity. All commands mentioned in a text retain their appearance and form.\
+**podman command** is always the beginning of the DESCRIPTION section. Putting the command as the first part of the DESCRIPTION ensures uniformity. All commands mentioned in a text retain their appearance and form. Commands that are quoted from other manpages have to be linked to those. None podman commands aren't to be linked but should be listed under SEE ALSO. This ensures clarity.\
```
edsantiago, line 32: I can't grok `None podman commands`. Is that supposed to be `Note`?

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
 
-#### **--option**, **-o**
 
-OPTIONS can be put after the command in two different ways. Eather the long version with **--option** or as the short version **-o**. If there are two ways to write an OPTION they are separated by a comma. If there are 2 (two) versions of one command the long version is always shown in front.
+#### **--version**, **-v** [=**false** | *true*]
 
-#### **--exit**
+OPTIONS can be put after the command in two different ways. Eather the long version with **--option** or as the short version **-o**. If there are two ways to write an OPTION they are separated by a comma. If there are 2 (two) versions of one command the long version is always shown in front. If the OPTION is boolean, it is not requiered to set an argument. In this case the equal sign is put in brackets `[]` with the arguments, in difference to arguments that are not boolean as shown with **--answer**. The default argument that is shown for a boolean OPTION is given when the OPTION is not used in the command. If the OPTION is used in the command and no argument is specified, the OPTION will always become `true`.\
```
edsantiago, line 42: There is no reason for `2 (two)`, and it is awkward. Standard usage is to simply use the spelled-out number: "If there are two ..."

edsantiago: Typo: `requiered` should be `required`

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
+
+#### **--exit** [=**true** | *false*]
 
 An example of an OPTION that has only one possible structure. Thus, it cannot be executed by the extension **-e**.
 
-#### **--answer**=, **-a**=**_active_** | *disable*
+#### **--answer**=, **-a**=**active** | *disable*
 
-OPTIONS that accept 2 possible arguments as inputs are shown above. If there is a default argument that is selected when no special input is made, it is shown in **_bold italics_**. It must always be ensured that the standard argument is in the first place after the OPTION. In this example, there are 2 (two) different versions to execute the command. Both versions of the OPTION have to be shown with the arguments behind them.
+An example for an OPTIONS that accept 2 possible arguments as inputs is shown above. If there is a default argument that is selected when the OPTION is not used in the command, it is shown in **bold**. It must always be ensured that the standard argument is in the first place after the OPTION. In this example, there are 2 (two) different versions to execute the command. Both versions of the OPTION have to be shown with the arguments behind them. The default is **active**.
```
edsantiago, line 51: I can't parse this either. Maybe you mean "An example for an OPTION that accepts ..."?

Also, same comment about `2 (two)`.

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
 
 #### **--status**=**good** | *better* | *best*
 
-This is an example for 3 (three) arguments behind an OPTION. If the number of arguments is higher than 3 (three), the arguments are **not** listed after the equal sign. The arguments have to be explained in a table like in **--test**=**_test_** regardless of the number of arguments.
+This is an example for 3 (three) arguments behind an OPTION. If the number of arguments is higher than 3 (three), the arguments are **not** listed after the equal sign. The arguments have to be explained in a table like in **--test**=**_test_** regardless of the number of arguments. The default is **good**.
```
edsantiago, line 55: "greater than", not "higher than". And, same thing with the `3 (three)`.

docs/source/markdown/podman-attach.1.md
```diff
@@ -9,48 +9,52 @@ podman\-attach - Attach to a running container
 **podman container attach** [*options*] *container*
 
 ## DESCRIPTION
-The attach command allows you to attach to a running container using the container's ID
-or name, either to view its ongoing output or to control it interactively.
+**podman attach** allows you to attach to a running *container* using the *container's name* or *ID*, either to view its ongoing output or to control it interactively.
```
edsantiago, line 12: I realize this isn't your change, but I bet @TomSweeneyRedHat is going to jump in here and say something about not using "you"...

docs/source/markdown/podman-attach.1.md
```diff
@@ -9,48 +9,52 @@ podman\-attach - Attach to a running container
...
-Instead of providing the container name or ID, use the last created container. If you use methods other than Podman
-to run containers such as CRI-O, the last started container could be from either of those methods. (This option is not available with the remote Podman client)
+Instead of providing the *container name* or *ID*, use the last created *container*. If you use methods other than Podman to run containers such as CRI-O, the last started *container* could be from either of those methods. The default is **false**.\
+*IMPORTANT: This option is not available with the remote Podman client*
 
-#### **--no-stdin**
+#### **--no-stdin** [=**false** | *true*]
 
-Do not attach STDIN. The default is false.
+Do not attach STDIN. The default is **false**.
```
edsantiago, line 31: This isn't your change, but this is one of those cases that is nearly impossible for a human brain to read unambiguously. First, I don't believe that the default is false -- that makes no sense. Second, it really needs clarification. I think this should be something like "--no-stdin [=true|false" .... By default, **podman attach** includes stdin." But I think this needs clarification from other team members, because in my testing I can't actually get stdin to work in podman attach no matter what I do.

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -7,44 +7,41 @@ podman-auto-update - Auto update containers according to their auto-update polic
...
+The label `image` is an alternative to `registry` maintained for backwards compatibility.
 An image is considered updated if the digest in the local storage is different than the one of the remote image.
 If an image must be updated, Podman pulls it down and restarts the systemd unit executing the container.
 
 The registry policy requires a fully-qualified image reference (e.g., quay.io/podman/stable:latest) to be used to create the container.
 This enforcement is necessary to know which image to actually check and pull.
 If an image ID was used, Podman would not know which image to check/pull anymore.
 
-Alternatively, if the autoupdate label is set to "local", Podman will compare the image a container is using to the image with it's raw name in local storage.
+Alternatively, if the autoupdate label is set to `local`, Podman will compare the image a container is using to the image with it's raw name in local storage.
```
edsantiago, line 21: Not your change, but please remove the incorrect apostrophe from `it's`

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -7,44 +7,41 @@ podman-auto-update - Auto update containers according to their auto-update polic
...
 If an image is updated locally, Podman simply restarts the systemd unit executing the container.
 
-If "io.containers.autoupdate.authfile" label is present, Podman reaches out to corresponding authfile when pulling images.
+If `io.containers.autoupdate.authfile` label is present, Podman reaches out to the corresponding authfile when pulling images.
 
-At container-creation time, Podman looks up the "PODMAN_SYSTEMD_UNIT" environment variables and stores it verbatim in the container's label.
-This variable is now set by all systemd units generated by `podman-generate-systemd` and is set to `%n` (i.e., the name of systemd unit starting the container).
+At container-creation time, Podman looks up the `PODMAN_SYSTEMD_UNIT` environment variables and stores it verbatim in the container's label.
```
edsantiago, line 26: Not your change, but please remove the `s` from `environment variables`.

docs/source/markdown/podman-commit.1.md
Procyhon: I found a workaround. I will just put the `**` in front and in the back of `[]`.
Example: \**[podman create]**

docs/source/markdown/podman-commit.1.md
Procyhon: The only reason i used `\` instead of the normal way, which is putting 2 spaces at the end of the line and adding a new line, is because the Validate fedora-34 Build test gives me an error because of whitespaces at the end of a line. `go-md2man` only accepts line breaks in the form of whitespace in the end and a new line. So it should be impossible to have a line break at this point. If there is truely no option for it, I will change it in general and come up with something new.

docs/source/markdown/podman-commit.1.md
Procyhon: Since **oci** is the default and *docker* is only a possible argument that can be used instead of it, it is right that it is *italic*. Thats how it is defined in the MANPAGE_SYNTAX in general.

docs/source/markdown/podman-attach.1.md
Procyhon: Maybe we can filter them out with a regex before. In this case we could also leave the `**` in the `[]`. But sadly I am not familiar with regex.

docs/MANPAGE_SYNTAX.md
Procyhon: It was an idea. I wanted to see if this a good way to react on reading numbers wrong, but I guess you are right. In the end, it is just useless.

docs/MANPAGE_SYNTAX.md
Procyhon: Grammar error on my side.

docs/source/markdown/podman-attach.1.md
Procyhon: I do believe that the default is false. Because it is --**no**-stdin and not just --stdin. https://github.com/containers/podman/blob/52dae693da0df1447b7f5210a4c842d5c5a8a401/cmd/podman/containers/attach.go#L52

I thought about how to express it alot, so I came up with this definition that is in the MANPAGE_SYNTAX:
If the OPTION is boolean, it is not requiered to set an argument. In this case the equal sign is put in brackets `[]` with the arguments, in difference to arguments that are not boolean as shown with **--answer**. The default argument that is shown for a boolean OPTION is given when the OPTION is not used in the command. If the OPTION is used in the command and no argument is specified, the OPTION will always become `true`.

edsantiago: Sorry, I wasn't clear. The ambiguity is in the meaning of "default": A reader of the man page, seeing the following documentation string:
```
    **--no-stdin** [=**false** | *true*]    [...] The default is **false**
```
...might have trouble understanding exactly how the flag operates:
* Does "default" apply to **the flag**? That is, if I, the end  user, specify `--no-stdin`, is that the same as `--no-stdin=false`, and hence, I need to explicitly specify `--no-stdin=true` to get the desired effect? Or,
* Does "default" apply to **podman itself**? That is, is the man page saying that `no-stdin` is false by default, hence podman does attach stdin? 
* 
Double negatives are evil. This is a horrible situation that needs clarification. IMO I would completely remove the `[=**false** | *true*]` string from the flag description, and remove "The default is false". They add clutter and confusion. Less is more:
```
   #### **--no-stdin**
   Do not attach STDIN.
```

docs/MANPAGE_SYNTAX.md
edsantiago: OH! I think you mean "Non-podman commands" then? Is that it?

Procyhon: Yes I believe it is.

docs/source/markdown/podman-attach.1.md
Procyhon: > Does "default" apply to podman itself? That is, is the man page saying that no-stdin is false by default, hence podman does attach stdin?

Yes this is what I mean. In the end it is optinial if you want to put an argument behind the option. This is why the whole thing is in `[]` including the equal sign. But since the default value of boolean options is sometimes `true` and sometimes `false` I thought it needed clarification. I had the choice to define what is default for me. 

> Does "default" apply to the flag? That is, if I, the end user, specify --no-stdin, is that the same as --no-stdin=false

No, thats is why I said in the MANPAGE_SYNTAX that if you type a option without putting anything behind it, it is always true. I believe this is the way in which it is currently working. 

> Double negatives are evil. This is a horrible situation that needs clarification.

That was my starting point and I think that to. But it not on me to change the option itself but in this case I would like to have it changed. 

>I would completely remove the [=**false** | *true*] string from the flag description, and remove "The default is false". They add clutter and confusion.

I though since there are these different cases of options that sometimes default to `true` and sometimes to `false` it should be mentioned. So you know what option will be in command automaticly if you dont specifiy them.

docs/source/markdown/podman-attach.1.md
rhatdan: We would need to do some sed magic.  @edsantiago @nalind Any ideas?

edsantiago: there's already some sed magic in the Makefile, for dealing with man pages. See `MANPAGES`, line 430. I'm not sure what needs to be done here but I'll take a look in the morning.

docs/source/markdown/podman-commit.1.md
Procyhon: This seems to work for the man pages but not for the markdown pages.

docs/source/markdown/podman-attach.1.md
Procyhon: It seems like links to other man pages or config files can only work in either markdown or in the man page. It would be really sad to dispense the idea of having commands linked to the correct .md files. But at the current stage until a solution is found for it, I will revert it.

Procyhon:
> @Procyhon Thanks this is great, but Please review the man page that is generated. Most users of Podman are going to be using the man page, rather then viewing them in markdown, since go-md2man is not perfect, we need to make sure that the generated man pages still look good, and do not have embeded links viewable to the user. In most cases I would prefer the markdown to be worse then make the man pages look bad.

@rhatdan I had a look at the manpages in relation to the markdown and I have to agree that go-md2man is far from optimal. Is there another way to result this. Maybe one structure for md and one for the manpages. Tho this would result in each command having two files. But I think this is not the intention. Until there is a solution, I'll leave things like line breaks and links out of the files, since go-md2man can't work with them.

rhatdan:
I think we can work on preprocessing the markdown with sed, or open up PR for go-md2man, but I am not sure how active that project is at this point.

docs/source/markdown/podman-attach.1.md
edsantiago: If I understand the problem correctly, perhaps this could solve it:
```diff
diff --git a/Makefile b/Makefile
index 15d6d9fb6..70ba76e49 100644
--- a/Makefile
+++ b/Makefile
@@ -428,8 +428,10 @@ pkg/api/swagger.yaml: .gopathok
        make -C pkg/api

 $(MANPAGES): %: %.md .install.md2man docdir
-       @sed -e 's/\((podman[^)]*\.md)\)//g' -e 's/\[\(podman[^]]*\)\]/\1/g' \
-               -e 's;<\(/\)\?\(a\|a\s\+[^>]*\|sup\)>;;g' $<  | \
+       @sed -e 's/\((podman[^)]*\.md)\)//g' \
+            -e 's/\[\(podman[^]]*\)\]/\1/g' \
+            -e 's/\[\(\*[^]]*\)\](http[^)]\+)/\1/g' \
+            -e 's;<\(/\)\?\(a\|a\s\+[^>]*\|sup\)>;;g' $<  | \
        $(GOMD2MAN) -in /dev/stdin -out $(subst source/markdown,build/man,$@)

 .PHONY: docdir
```

docs/source/markdown/podman-attach.1.md
edsantiago: >No, thats is why I said in the MANPAGE_SYNTAX that if you type a option without putting anything behind it, it is always true. 

I interpret that as "Please read a long complicated document before you are allowed to understand our man pages".  I don't think that's fair to end users, who are simply trying to solve a problem by finding one small relevant piece in a large documentation set. I feel pretty strongly about removing the `[=false]` thing from the description, I think if you look at it from the perspective of a normal user, one without your background and expertise, you will agree that it is too confusing.

Procyhon:
I have no idea why the links don't work. If I use go-md2man myself on the existing podman.1.md it gives me the wrong output in which you are able to see the links in bad formatting. If I look in the existing manpage for podman.1.md, the links are not visible, which is strange, since I used the same source code. Maybe there was something changed in the way that go-md2man acts with links since the podman.1.md file was last changed.

docs/source/markdown/podman-attach.1.md
Procyhon: You are completely right. For the normal user it is very confusing. I will remove it.

Just in general, is there no where this fits. Cause the only way currently for users to find out about it, is if they look into the code itself. I don't have a problem with that, but I have the feeling it should be mentioned somewhere. But you are right, maybe not in the manpages.

docs/source/markdown/podman-attach.1.md
Procyhon: I meant, that the whole link should be filtered out in the manpage. In the way it is now it's just put in a less-than sign and greater-than sign. So everything that is between the less-than sign and greater-than sign should go.

The `[]` aren't a problem anymore. 

It would also be helpful to replace the `\` at the end of the line with two spaces. In this way the desired line break can be achieved.

edsantiago: >everything that is between the less-than sign and greater-than sign should go.

My patch converts `[**markup**](link)` to just `**markup**`, which isn't exactly the same thing, but has the same end effect: the `<link>` is gone from the man page:
```diff
diff -ur docs/build/man.ante/podman-attach.1 docs/build/man/podman-attach.1
--- docs/build/man.ante/podman-attach.1 2021-06-03 10:46:32.118576644 -0600
+++ docs/build/man/podman-attach.1      2021-06-03 13:50:10.629289923 -0600
@@ -19,8 +19,7 @@
 You can detach from the \fIcontainer\fP (and leave it running) using a configurable key sequence. The default
 sequence is \fB\fCctrl\-p,ctrl\-q\fR\&.
 Configure the keys sequence using the \fB\-\-detach\-keys\fP option, or specifying
-it in the \fB\fCcontainers.conf\fR file: see \fBcontainers.conf(5)\fP
-\[la]https://github.com/containers/common/blob/master/docs/containers.conf.5.md\[ra] for more information.
+it in the \fB\fCcontainers.conf\fR file: see \fBcontainers.conf(5)\fP for more information.

 .SH OPTIONS
 .SS \fB\-\-detach\-keys\fP=\fBsequence\fP
```
Is that not good enough?

docs/MANPAGE_SYNTAX.md
TomSweeneyRedHat, line 32:
```suggestion
**podman command** is always the beginning of the DESCRIPTION section. Putting the command as the first part of the DESCRIPTION ensures uniformity. All commands mentioned in the text retain their appearance and form. Commands that are quoted from other manpages have to be linked to those manpages. Non-podman commands are not to be linked but should be listed under SEE ALSO. This ensures clarity.\
```

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
 Example sentence: The command **podman command** is an example command.\
-It should also be specified if the command can only be run as root. In addition, it should be described when a command or OPTION cannot be executed with the remote client. For a command, this should be done in the DESCRIPTION part. For the OPTIONS, it should be done in the DESCRIPTION of the specified OPTION. Do not use pronouns in the man pages, especially the word `you`.
+It should also be specified if the command can only be run as root. In addition, it should be described when a command, OPTION, or other content cannot be executed with the remote client or in combination with other commands, OPTIONS, or contents. In this case the following sentence is put at the end of the first chapter of an command, OPTION, or content: *IMPORTANT: This option/command/other is not available with the command/OPTION/content/remote Podman client*. For a command, this should be done in the DESCRIPTION part. For the OPTIONS, it should be done in the DESCRIPTION of the specified OPTION. Do not use pronouns in the man pages, especially the word `you`.
```
TomSweeneyRedHat, line 34:
```suggestion
It should also be specified if the command can only be run as root. In addition, it should be described when a command, OPTION, or other content cannot be executed with the remote client or in combination with other commands, OPTIONS, or content. In this case, the following sentence is put at the end of the first chapter of a command, OPTION, or content: *IMPORTANT: This option/command/other is not available with the command/OPTION/content/remote Podman client*. For a command, this should be done in the DESCRIPTION section. For the OPTIONS, it should be done in the DESCRIPTION of the specified OPTION. Do not use pronouns in the man pages, especially the word `you`.
```

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
 
 ## OPTIONS
-All flags are referred to as OPTIONS. The term flags should not be used. All OPTIONS are listed in this section. OPTIONS that appear in descriptions of other OPTIONS and sections retain their appearance, for example: **--exit**. Each OPTION should be explained to the fullest extend below the OPTION itself. Each OPTION is behind an H4-header (`####`).
+All flags are referred to as OPTIONS. The term flags should not be used. All OPTIONS are listed in this section. OPTIONS that appear in descriptions of other OPTIONS and sections retain their appearance, for example: **--exit**. OPTIONS that are quoted from other manpages have to be linked to those with the command to execute them. OPTIONS within the same manpage don't have to be linked. This ensures clarity. Each OPTION should be explained to the fullest extend below the OPTION itself. Each OPTION is behind an H4-header (`####`). If the OPTION has a default argument, is has to be written in the description of the OPTION. If the OPTION is also not available with the remote client, the sentence about the default argument should the second to last sentence.
```
TomSweeneyRedHat, line 37: It would be good to put an example of a link for an option from another manpage.

TomSweeneyRedHat, line 37:
```suggestion
All flags are referred to as OPTIONS. The term flags should not be used. All OPTIONS are listed in this section. OPTIONS that appear in descriptions of other OPTIONS and sections retain their appearance, for example: **--exit**. OPTIONS that are quoted from other manpages have to be linked to those with the command to execute them. OPTIONS within the same manpage don't have to be linked. This ensures clarity. Each OPTION should be explained to the fullest extent below the OPTION itself. Each OPTION is behind an H4-header (`####`). If the OPTION has a default argument, it has to be written in the description of the OPTION. If the OPTION is also not available with the remote client, the sentence about the default argument should the second to last sentence.
```

docs/MANPAGE_SYNTAX.md
TomSweeneyRedHat, line 42:
```suggestion
OPTIONS can be put after the command in two different ways. Eather the long version with **--option** or as the short version **-o**. If there are two ways to write an OPTION, they are separated by a comma. If there are two versions of one command the long version is always shown in front. If the OPTION is a boolean, it is not required to set an argument. In this case, the equal sign is put in brackets `[]` with the arguments.  This differs for arguments that are not boolean, as shown with **--answer** below. The default argument for a boolean OPTION is given when the OPTION is not used in the command. If the OPTION is used in the command and no argument is specified, the OPTION will always become `true`.\
```

docs/MANPAGE_SYNTAX.md
TomSweeneyRedHat, line 51:
```suggestion
The "answer" option above is an example of an OPTION that accepts two possible arguments as inputs. If there is a default argument that is selected when the OPTION is not used in the command, it is shown in **bold**. It must always be ensured that the standard argument is in the first position after the OPTION. In this example, there are two different ways to execute the command. Both possible OPTIONs have to be shown with the arguments following them. The default is **active**.
```

TomSweeneyRedHat: I think this is what you were getting at.  ^^

docs/MANPAGE_SYNTAX.md
TomSweeneyRedHat, line 55:
```suggestion
This is an example of three arguments following an OPTION. If the number of arguments is higher than three, the arguments are **not** listed after the equal sign. The arguments have to be shown in a table like the one below for **--test**=**_test_**, regardless of the number of arguments. The default value is shown as **good**.
```

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
+This is an example for 3 (three) arguments behind an OPTION. If the number of arguments is higher than 3 (three), the arguments are **not** listed after the equal sign. The arguments have to be explained in a table like in **--test**=**_test_** regardless of the number of arguments. The default is **good**.
 
-#### **--test**=**_test_**
+#### **--test**=**test**
 
-OPTIONS that are followed by an equal sign include an argument after the equal sign in *italic*. If there is a default argument, that is used if the OPTION is not specified in the **command**, the argument after the eqaul sign is displayed in **bold**. All arguments must be listed and explained in the text below the OPTION.
+OPTIONS that are followed by an equal sign include an argument after the equal sign in **bold**. If there is a default argument, that is used if the OPTION is not specified in the command, the argument after the eqaul sign is displayed in **bold**. All arguments must be listed and explained in the text below the OPTION. The default is **example one**.
```
TomSweeneyRedHat, line 59:
```suggestion
OPTIONS that are followed by an equal sign include an argument after the equal sign in **bold**. If there is a default argument, that is used if the OPTION is not specified in the command, the argument after the eqaul sign is displayed in **bold**. All arguments must be listed and explained in the text below the OPTION. The default value is shown as **example one**.
```

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
 
 | Argument           | Description                                                                 |
 | -                  | -                                                                           |
-| **_example one_**  | This argument is the standard argument if the OPTION is not specified.      |
+| **example one**    | This argument is the standard argument if the OPTION is not specified.      |
```
TomSweeneyRedHat, line 63: standard argument or default argument.  I see it's pre-existing...   Hm...

docs/MANPAGE_SYNTAX.md
```diff
@@ -29,45 +29,51 @@ podman\-command - short description
...
 | *example four*     | Example: Can be combined with **--exit**.                                   |
 | *example five*     | The fifth description                                                       |
 
 The table shows an example for a listing of arguments. The contents in the table should be aligned left. If the content in the table conflicts with this, it can be aligned in a way that supports the understanding of the content. If there is a standard argument, it **must** listed as the first entry in the table.
 
 If the number of arguments is smaller than 4 (four) they have to be listed behind the OPTION as seen in the OPTION **--status**.
 
+#### **--problem**=*container*
+
+OPTIONS that are followed by an equal sign that is followed by an unspecified argument, have no standard argument. If this OPTION is written with an equal sign and the argument is left empty, there will be no error, but the OPTION will be ignored. The meaning of the argument is described preferably in `one` word after the equal sign in *italic* format.
```
TomSweeneyRedHat, line 75: I'm a bit lost on this one.  I'm seeing `*container*` here, does this description pertain to `--problem`?
```suggestion
OPTIONS that are followed by an equal sign that is then followed by an unspecified argument, have no standard argument. If this OPTION is written with an equal sign and the argument is left empty, there will be no error, but the OPTION will be ignored. The meaning of the argument is described preferably in `one` word after the equal sign in *italic* format.
```

docs/source/markdown/podman-attach.1.md
TomSweeneyRedHat: indeed, 'you' is not good!
```suggestion
**podman attach** attaches to a running *container* using the *container's name* or *ID*, to either view its ongoing output or to control it interactively.
```

docs/source/markdown/podman-attach.1.md
```diff
@@ -9,48 +9,52 @@ podman\-attach - Attach to a running container
 **podman container attach** [*options*] *container*
 
 ## DESCRIPTION
-The attach command allows you to attach to a running container using the container's ID
-or name, either to view its ongoing output or to control it interactively.
+**podman attach** allows you to attach to a running *container* using the *container's name* or *ID*, either to view its ongoing output or to control it interactively.
 
-You can detach from the container (and leave it running) using a configurable key sequence. The default
+You can detach from the *container* (and leave it running) using a configurable key sequence. The default
```
TomSweeneyRedHat, line 14:
```suggestion
The *container* can be detached (and leave it running) using a configurable key sequence. The default
```

docs/source/markdown/podman-attach.1.md
Procyhon: It works for http and https but no for relative links like `(podman.1.md)`. The link is gone, but the `[]` are still there, around the command.

docs/source/markdown/podman-auto-update.1.md
TomSweeneyRedHat, line 26:
```suggestion
At container-creation time, Podman looks up the `PODMAN_SYSTEMD_UNIT` environment variable and stores it verbatim in the container's label.
```

docs/source/markdown/podman-attach.1.md
Procyhon: @edsantiago @TomSweeneyRedHat @rhatdan @vrothberg 
What is the general opinion about the topic above. In the end, I find it really important. So it should be mentioned somewhere. I could live with leaving out the arguments `false` and `true` in the option line. But I would at least like to specifiy the default value in the description with a short sentence.

docs/source/markdown/podman-auto-update.1.md
TomSweeneyRedHat, line 41: I think you can't have the `**` in the bracket.  Perhaps
```suggestion
Path of the authentication file. Default is `${XDG_RUNTIME_DIR}/containers/auth.json`, which is set using **[podman login](podman-login.1.md)**.
```

docs/source/markdown/podman-auto-update.1.md
```diff
@@ -7,44 +7,41 @@ podman-auto-update - Auto update containers according to their auto-update polic
...
 #### **--authfile**=*path*
 
-Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
-If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+Path of the authentication file. Default is `${XDG_RUNTIME_DIR}/containers/auth.json`, which is set using [**podman login**](podman-login.1.md).
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using **docker login**.
 
-Note: You can also override the default path of the authentication file by setting the REGISTRY\_AUTH\_FILE
-environment variable. `export REGISTRY_AUTH_FILE=path`
+Note: There is also an option to override the default path of the authentication file by setting the `REGISTRY_AUTH_FILE` environment variable. This can be done with **export REGISTRY_AUTH_FILE=_path_**.
```
TomSweeneyRedHat, line 44:
```suggestion
Note: There is also the option to override the default path of the authentication file by setting the `REGISTRY_AUTH_FILE` environment variable. This can be done with **export REGISTRY_AUTH_FILE=_path_**.
```

TomSweeneyRedHat (commented):
@Procyhon I left a lot of comments, but a really great start, thank you!

docs/source/markdown/podman-attach.1.md
Procyhon: Quick overview for the visual errors for links in the manpages:

- `**[podman](podman.1.md)**` results in **podman**
 *for commands alone everything is working*


- `**[podman-generate-systemd --new](podman-generate-systemd.1.md#--new)**` results in 
**podman-generate-systemd --new**(podman-generate-systemd.1.md#--new)
*first possibility for command with an option*

- `[**podman-generate-systemd --new**](podman-generate-systemd.1.md#--new)` results in
 **podman-generate-systemd --new** <podman-generate-systemd.1.md#--new>
*second possibility for command with an option*
It seems like links that include an option don't work. I believe it is do to the space between the command and option.

- `**[containers.conf(5)](https://github.com/containers/common/blob/master/docs/containers.conf.5.md)**` results in
containers.conf(5) \<https://github.com/containers/common/blob/master/docs/containers.conf.5.md>
*first possibility for a link with http/https*

- `[**containers.conf(5)**](https://github.com/containers/common/blob/master/docs/containers.conf.5.md)` results in
**containers.conf(5)**
*second possibility for a link with http/https*
Since the `**` should be on the outside of the whole structure, it should be switched at the http/https links. 

Thank for the help with the regex. :)

Procyhon: Can someone confirm, that the last `-e` line added to the makefile works for deleting the last two whitespaces in the `.md` file and creating a line break: 
```
$(MANPAGES): %: %.md .install.md2man docdir
	@sed -e 's/\((podman[^)]*\.md)\)//g' \
         -e 's/\[\(podman[^]]*\)\]/\1/g' \
         -e 's/\[\(\*[^]]*\)\](http[^)]\+)/\1/g' \
         -e 's;<\(/\)\?\(a\|a\s\+[^>]*\|sup\)>;;g' \
         -e 's/[[:blank:]]+/\n/g' $<  | \
	$(GOMD2MAN) -in /dev/stdin -out $(subst source/markdown,build/man,$@)
```

nalind: Just dropping in, but would `[[:blank:]]+` not match any sequence of two or more blank characters anywhere on any line?

Procyhon: Yep i think so, but my first thought was also wrong. I wanted to replace the `\` with two spaces. And i believe this works:

```
$(MANPAGES): %: %.md .install.md2man docdir
	@sed -e 's/\((podman[^)]*\.md)\)//g' \
         -e 's/\[\(podman[^]]*\)\]/\1/g' \
         -e 's/\[\(\*[^]]*\)\](http[^)]\+)/\1/g' \
         -e 's;<\(/\)\?\(a\|a\s\+[^>]*\|sup\)>;;g' \
         -e 's/\\$//  /g' $<  | \
	$(GOMD2MAN) -in /dev/stdin -out $(subst source/markdown,build/man,$@)
```

Procyhon: I edited the regex a little more. It should work as intended now.

Procyhon:
Applied all changes. Should be ready now.

Issue #10522, opened 2021-06-01
v3.2: AVC with volumes


edsantiago:
Test `030-run.bats:podman run - check workdir` is failing in podman-3.2.0-0.2.rc3.fc34 under podman and podman-remote, root and rootless. It's an AVC. I don't know why CI didn't catch it.
```console
$ mkdir /tmp/mypodman
$ echo hi >/tmp/mypodman/file
$ ./bin/podman run --rm -v /tmp/mypodman:/mypodman alpine cat /mypodman/file
cat: can't open '/mypodman/file': Permission denied
```
AVC is:
```
type=AVC msg=audit(1622571374.261:211727): avc:  denied  { open } for  pid=2099144 comm="cat" path="/mypodman/file" dev="tmpfs" ino=1491929 scontext=system_u:system_r:container_t:s0:c380,c660 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=file permissive=0
```
Easily reproduced on master @ cbffdddce6d741eac4f3efa132016aba99683500

container-selinux-2.162.1-3.fc34.noarch and also -2.162.2-1.fc34

mheon:
@rhatdan PTAL

edsantiago:
It works with `:Z`, as expected. This is a regression, because it's causing the following test to fail:

https://github.com/containers/podman/blob/cbffdddce6d741eac4f3efa132016aba99683500/test/system/030-run.bats#L602-L604

That test dates to 2021-01-21. I don't know why the test doesn't use `:Z` or why it ever worked without it. Perhaps something special about `--workdir`?

edsantiago:
This is also manifesting in a lot of the `500-networking.bats` tests, which use `-v` (without `:Z`) to mount a file for the http server to serve.

edsantiago:
Problem is **selinux-policy-34.9-1.fc34**.  Everything works fine with 34.**8**-1.fc34.

Although this is not podman's fault, it's something we will need to deal with.

rhatdan:
This definitely looks like a valid error, and I have no idea why this worked in the past.  We should definitely not allow users to mount random content into SELinux containers and have it work.  Might be that the test systems previously did not have SELinux enabled, or had this /tmp/mypodman labeled correctly.  I don't see this as a bug in podman, but with an error in the test.

edsantiago:
[...]

PR #10527, opened 2021-06-01
System tests: add :Z to volume mounts

edsantiago:
selinux-policy-34.9-1.fc34 breaks a behavior we've relied on
since (at least) January 2020:

   - Revert "Add permission open to files_read_inherited_tmp_files()
     interface"

That's probably the correct thing to do, but it breaks our
existing tests. Solution: add ':Z' where needed.

Tested on Ed's laptop, which has the offending selinux-policy
as of 2021-05-31. Tests pass root and rootless. (I mention
this because tests will obviously pass in CI, which has a
much older selinux-policy).

Also: add a 'podman rmi' for cleanup in one test, to avoid
noise in test logs.

Fixes: #10522

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -600,12 +600,12 @@ json-file | f
     echo "$randomcontent" > $testdir/content
 
     # Workdir does not exist on the image but is volume mounted.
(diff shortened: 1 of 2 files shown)


edsantiago:

This will need to be cherrypicked into v3.2 (and probably more); I will do so tomorrow.

Issue #10188, opened 2021-04-01
VOLUME in Dockerfile created with wrong permissions + timestamp


ensc:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

A `VOLUME` within Dockerfile is created with wrong permissions.  E.g.

```
FROM fedora:33

RUN ls -lad /var/tmp

VOLUME /var/tmp

RUN ls -lad /var/tmp
```

and then `docker run ... ls -lad /var/tmp` shows three different modes:

```
drwxrwxrwt. 2 root root 4096 Feb 17 07:48 /var/tmp       <<< from original image
drwxrwxrwx. 2 root root 4096 Apr  1 15:30 /var/tmp       <<< in `podman build` after VOLUME
drwxr-xr-x. 2 root root 4096 Apr  1 15:31 /var/tmp       <<< in generated image
```

Doing the same with moby-engine-19.03.13 shows

```
drwxrwxrwt. 2 root root 4096 Feb 17 07:48 /var/tmp
drwxrwxrwt. 2 root root 4096 Feb 17 07:48 /var/tmp
drwxrwxrwt. 2 root root 4096 Feb 17 07:48 /var/tmp
```

I expect the moby-engine behavior:

- `/vat/tmp` keeps 01777 perms
- timestamps do not change

**Steps to reproduce the issue:**

1. run 'podman build' with Dockerfile above

2. start the generated image and look at `/var/tmp`

**Output of `podman version`:**

```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.15.8
Built:        Fri Feb 19 17:56:17 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.26-1.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.26, commit: 777074ecdb5e883b9bec233f3630c5e7fa37d521'
  cpus: 8
  distribution:
    distribution: fedora
    version: "33"
  eventLogger: journald
  hostname: ensc-virt.intern.sigma-chemnitz.de
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 505
      size: 1
    - container_id: 1
      host_id: 505000000
      size: 1000000
    uidmap:
    - container_id: 0
      host_id: 505
[...]

PR #10531, opened 2021-06-02
Fix permissions on initially created named volumes

rhatdan:
Permission of volume should match the directory it is being mounted on.

Fixes: https://github.com/containers/podman/issues/10188

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1062,7 +1062,7 @@ func (c *Container) init(ctx context.Context, retainRetries bool) error {
 	}
 
 	for _, v := range c.config.NamedVolumes {
-		if err := c.chownVolume(v.Name); err != nil {
+		if err := c.fixVolumePermissions(v); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


rhatdan:
@mheon PTAL

libpod/container_internal.go
```diff
@@ -1680,11 +1680,11 @@ func (c *Container) mountNamedVolume(v *ContainerNamedVolume, mountpoint string)
 	return vol, nil
 }
 
-// Chown the specified volume if necessary.
-func (c *Container) chownVolume(volumeName string) error {
-	vol, err := c.runtime.state.Volume(volumeName)
+// Fix ownership and permissions the specified volume if necessary.
```
vrothberg, line 1683:
```suggestion
// Fix ownership and permissions of the specified volume if necessary.
```

libpod/container_internal.go
```diff
@@ -1734,6 +1734,15 @@ func (c *Container) chownVolume(volumeName string) error {
 		if err := os.Lchown(mountPoint, uid, gid); err != nil {
 			return err
 		}
+
+		st, err := os.Lstat(filepath.Join(c.state.Mountpoint, v.Dest))
```
vrothberg, line 1738: Could you add a comment here? Maybe also mention #10188 for historical reasons.

test/e2e/run_test.go
```diff
@@ -904,6 +904,18 @@ USER bin`, BB)
 		Expect(session.ExitCode()).To(Equal(100))
 	})
 
+	It("podman run with named volume", func() {
```
vrothberg, line 907: Can you drop a comment that points to #10188?

vrothberg (commented):
Just nits, code LGTM.

Does it really fully fix #10188?  It looks like there's something in Buildah as well.

test/e2e/run_test.go
rhatdan: Why, this fixes the permissions of the volume created, I don't believe we care about the timestamps.

libpod/container_internal.go
rhatdan: Sure,

TomSweeneyRedHat:
All kinds of test unhappiness @rhadan

test/e2e/run_test.go
rhatdan: Ok It now handled timestamps.

rhatdan:
Fixes: https://github.com/containers/podman/issues/10606

TomSweeneyRedHat:
LGTM
I restarted one test that was failing, it appears we may have a network flake flying around.

PR #10534, opened 2021-06-02
[v.3.2] events: support disjunctive filters

vrothberg:
While different filters are applied in conjunction, the same filter (but
with different values) should be applied in disjunction.  This allows,
for instance, to query the events of two containers.

Fixes: #10507
Backport-of: commit 37f39eefee72ec4fb6c6bd71642e9d384c448387


vrothberg:

Luap99 (commented):
LGTM

PR #10546, opened 2021-06-03
Add CORS support

boaz0:


cmd/podman/system/service_abi.go
```diff
@@ -72,7 +72,7 @@ func restService(opts entities.ServiceOptions, flags *pflag.FlagSet, cfg *entiti
 	}
 
 	infra.StartWatcher(rt)
-	server, err := api.NewServerWithSettings(rt, opts.Timeout, listener)
+	server, err := api.NewServerWithSettings(rt, opts.Timeout, listener, opts.CorsHeaders)
```
rhatdan, line 75: Should we change the API to just pass in opts?

pkg/api/server/server.go
```diff
@@ -34,10 +34,12 @@ type APIServer struct {
 	context.CancelFunc               // Stop APIServer
 	idleTracker        *idle.Tracker // Track connections to support idle shutdown
 	pprof              *http.Server  // Sidecar http server for providing performance data
+	CorsHeaders        string        // Inject CORS headers to each request
 }
 
 // Number of seconds to wait for next request, if exceeded shutdown server
 const (
+	DefaultCorsHeaders       = "*"
```
rhatdan, line 42: Should this be "*" or just ""?

rhatdan (approved):
A couple of questions.

pkg/api/server/server.go
boaz0: I will change it, it should be empty string `""`

cmd/podman/system/service_abi.go
boaz0: I thought about it and I think so. It will be easier.

cmd/podman/system/service.go
```diff
@@ -39,7 +39,8 @@ Enable a listening service for API access to Podman commands.
 	}
 
 	srvArgs = struct {
-		Timeout int64
+		Timeout     int64
+		corsHeaders string
```
mheon, line 43: I think this needs to be Capitalized (first letter, at least) so the value can be set by the CLI package

boaz0: Right, good catch

boaz0:
@rhatdan I am now figuring out a way to test this. It seems like I will have to come up with a solution from scratch.

pkg/api/server/handler_api.go
```diff
@@ -63,6 +63,12 @@ func (s *APIServer) APIHandler(h http.HandlerFunc) http.HandlerFunc {
 			w.Header().Set("Libpod-API-Version", lv)
 			w.Header().Set("Server", "Libpod/"+lv+" ("+runtime.GOOS+")")
 
+			if s.CorsHeaders != "" {
+				w.Header().Set("Access-Control-Allow-Origin", s.CorsHeaders)
+				w.Header().Set("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept, X-Registry-Auth")
```
jwhonce, line 68: Do we need "Connection" and "Upgrade" in this header list?
We do need "X-Registry-Config".

jwhonce (changes requested):
One question and one change request.

TomSweeneyRedHat:
Changes LGTM once tests and @jwhonce 's concerns are addressed

TomSweeneyRedHat:
gack, hit the wrong button, didn't mean to close.

jwhonce (approved):

PR #10549, opened 2021-06-03
remote: always send resize before the container starts

### LINKED ISSUE #9859: race: podman-remote run -it : stty: standard input
Almost certainly related to #9831 / #9833

In window 1:
```console
$ ./bin/podman system service --timeout=0
```
In window 2:
```console
$ ./bin/podman-remote run -it --rm quay.io/libpod/testimage:20210223 stty size
stty: standard input
WARN[0000] Container 980c16fd4dd648821d6a1029f1d93a52f380250e4e516fc10c2c311c7caf25d9 does not exist: no container with name or ID "980c16fd4dd648821d6a1029f1d93a52f380250e4e516fc10c2c311c7caf25d9" found: no such container
```
The "Container does not exist" warning is correlated with `--rm`.

Once in a while, the `stty` will succeed:
```console
$ ./bin/podman-remote run -it quay.io/libpod/testimage:20210223 stty size
60 160
```
[UPDATED 2021-03-29 because I forgot to mention the workaround]
Workaround: add a `sleep`:
```console
$ ./bin/podman-remote run -it --rm quay.io/libpod/testimage:20210223 sh -c 'sleep 1;stty size'
60 160
```

Luap99:
There is race condition in the remote client attach logic. Because the
resize api call was handled in an extra goroutine the container was
started before the resize call happend. To fix this we have to call
resize in the same goroutine as attach. When the first resize is done
start a goroutine to listen on SIGWINCH in the background and resize
again if the signal is received.

Fixes #9859

### DIFF
--- a/pkg/api/handlers/compat/resize.go
+++ b/pkg/api/handlers/compat/resize.go
@@ -46,20 +46,13 @@ func ResizeTTY(w http.ResponseWriter, r *http.Request) {
 			utils.ContainerNotFound(w, name, err)
 			return
 		}
-		if state, err := ctnr.State(); err != nil {
-			utils.InternalServerError(w, errors.Wrapf(err, "cannot obtain container state"))
-			return
-		} else if state != define.ContainerStateRunning && !query.IgnoreNotRunning {
-			utils.Error(w, "Container not running", http.StatusConflict,
-				fmt.Errorf("container %q in wrong state %q", name, state.String()))
-			return
-		}
-		// If container is not running, ignore since this can be a race condition, and is expected
 		if err := ctnr.AttachResize(sz); err != nil {
-			if errors.Cause(err) != define.ErrCtrStateInvalid || !query.IgnoreNotRunning {
+			if errors.Cause(err) != define.ErrCtrStateInvalid {
 				utils.InternalServerError(w, errors.Wrapf(err, "cannot resize container"))
-				return
+			} else {
+				utils.Error(w, "Container not running", http.StatusConflict, err)
 			}
+			return
 		}
 		// This is not a 204, even though we write nothing, for compatibility
 		// reasons.
--- a/pkg/api/server/register_containers.go
+++ b/pkg/api/server/register_containers.go
@@ -1364,6 +1364,8 @@ func (s *APIServer) registerContainersHandlers(r *mux.Router) error {
 	//     $ref: "#/responses/ok"
 	//   404:
 	//     $ref: "#/responses/NoSuchContainer"
+	//   409:
+	//     $ref: "#/responses/ConflictError"
 	//   500:
 	//     $ref: "#/responses/InternalError"
 	r.HandleFunc(VersionedPath("/libpod/containers/{name}/resize"), s.APIHandler(compat.ResizeTTY)).Methods(http.MethodPost)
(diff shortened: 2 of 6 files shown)


Luap99:
@edsantiago PTAL I think this fixes your issue.

edsantiago:
Running this infinite loop for the past five minutes, and it's passing so far:
```console
$ while :;do bin/podman-remote run -it --rm quay.io/libpod/testimage:20210427 stty size;done
```
LGTM, thank you! And thank you for identifying and fixing the appropriate test!

rhatdan:
@Luap99 If we don't have a thread watching for resize, then if the terminal actually get's resized, the container will now realize it.  Have you tried this out with this change.

Luap99:
@rhatdan Yes the extra thread is still created and resizes when SIGWINCH is received. Only the first resize is made from the attach thread.

rhatdan:
LGTM

pkg/api/handlers/compat/resize.go
```diff
@@ -71,14 +62,6 @@ func ResizeTTY(w http.ResponseWriter, r *http.Request) {
 			utils.SessionNotFound(w, name, err)
 			return
 		}
-		if state, err := ctnr.State(); err != nil {
```
mheon, line 74: I think this bit is still reasonable to hit - exec sessions in a non-running container will never happen.

Luap99: I added this back. It looks `podman exec -it con1 stty size` is broken as well but for both local and remote.

pkg/api/handlers/compat/resize.go
```diff
@@ -46,14 +46,6 @@ func ResizeTTY(w http.ResponseWriter, r *http.Request) {
...
-			utils.InternalServerError(w, errors.Wrapf(err, "cannot obtain container state"))
-			return
-		} else if state != define.ContainerStateRunning && !query.IgnoreNotRunning {
-			utils.Error(w, "Container not running", http.StatusConflict,
-				fmt.Errorf("container %q in wrong state %q", name, state.String()))
-			return
-		}
 		// If container is not running, ignore since this can be a race condition, and is expected
 		if err := ctnr.AttachResize(sz); err != nil {
 			if errors.Cause(err) != define.ErrCtrStateInvalid || !query.IgnoreNotRunning {
```
mheon, line 51: I think we probably need to catch `define.ErrCtrStateInvalid` and return a `http.StatusConflict` down here. It should be a lot more forgiving than the above check, which only allows Running (we can also resize Created containers, and probably Paused containers as well, I forget) but ensures we're still returning good HTTP status codes.

mheon:
One more nit then LGTM

pkg/api/handlers/compat/resize.go
Luap99: Not sure about the http status code, this is used for the compat api as well and the [docker api](https://docs.docker.com/engine/api/v1.41/#operation/ContainerResize) does not specify 409.

Luap99: fixed

mheon:
I think the compose test flaked - restarted

/lgtm
/hold

rhatdan:
Nice work @Luap99.  This is a very old issue that @edsantiago has wanted fix for a while.

Issue #5572, opened 2020-03-20
systemd not always producing the %t from systemd file


ikke-t:
/kind bug

**Description**

I followed the blog to use conmon-pidfile and cidfile from [this blog](https://www.redhat.com/sysadmin/podman-shareable-systemd-services). However, systemd doesn't fills %t for podman user containers.

ping @vrothberg 

**Steps to reproduce the issue:**

1. Ansible does this systemd file:

```
cat /etc/systemd/system/ha-container-pod.service 
[Unit]
Description=ha Podman Container
After=network.target

[Service]
Type=simple
TimeoutStartSec=15
ExecStartPre=-/usr/bin/rm -f /%t/%n-pid /%t/%n-cid
User=redhatter

ExecStart=/usr/bin/podman run --name ha \
  --rm -p 8123:8123/tcp -v "/var/lib/containers/storage/homeassistant-config:/config:Z" --hostname="ha.localnet" --memory=1024M \
  --conmon-pidfile  /%t/%n-pid --cidfile /%t/%n-cid \
  homeassistant/home-assistant:stable

ExecStop=/usr/bin/sh -c "/usr/bin/podman stop -t "15" `cat /%t/%n-cid`"
ExecStop=/usr/bin/sh -c "/usr/bin/podman rm -f `cat /%t/%n-cid`"
Restart=on-failure
RestartSec=30
KillMode=none
PIDFile=/%t/%n-pid

[Install]
WantedBy=multi-user.target

```

2 add user ```sudo useradd redhatter```
3. reload systemd ```sudo systemctl daemon-reload```
4. start container:  sudo systemctl start ha-container-pod.service

**Describe the results you received:**

```
sudo journalctl -eu ha-container-pod.service
Mar 20 17:12:57 gr8.localdomain systemd[1]: Started ha Podman Container.
Mar 20 17:12:57 gr8.localdomain podman[11501]: Error: error opening cidfile //run/ha-container-pod.service-cid
Mar 20 17:12:57 gr8.localdomain systemd[1]: ha-container-pod.service: Main process exited, code=exited, status=125/n/a
Mar 20 17:12:57 gr8.localdomain systemd[1]: ha-container-pod.service: Failed with result 'exit-code'.
Mar 20 17:13:27 gr8.localdomain systemd[1]: ha-container-pod.service: Service RestartSec=30s expired, scheduling restart.
[...]

PR #10557, opened 2021-06-04
systemd/generate: change type to notify

vrothberg:
Change the type of units generated with --new from "forking" to
"notify".  This brings Podman closer to systemd and opens up
Podman to a number of use cases (see #5572).

Units generated without --new remain with `type=forking`.  I
experimented a bit with adding a `--sdnotify` flag to `podman start` but
it doesn't really work well since we're competing with the default
sdnotify mode set during container creation.

Fixes: #5572

### DIFF
--- a/pkg/systemd/generate/common.go
+++ b/pkg/systemd/generate/common.go
@@ -60,7 +60,7 @@ func filterPodFlags(command []string, argCount int) []string {
 	return processed
 }
 
-// filterCommonContainerFlags removes --conmon-pidfile, --cidfile and --cgroups from the specified command.
+// filterCommonContainerFlags removes --sdnotify, --rm and --cgroups from the specified command.
 // argCount is the number of last arguments which should not be filtered, e.g. the container entrypoint.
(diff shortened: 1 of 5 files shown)


giuseppe (approved):
LGTM

vrothberg:
@rhatdan PTAL

vrothberg:
Want to tackle some issue in the edge space.  Mostly polishing the systemd integration and anything related to journald.

vrothberg:

Luap99 (commented):
LGTM

mheon:
Is this ready? I thought the conmon-based SDNotify implementation was still unmerged, and the default OCI runtime implementation has massive issues and should never be used?

vrothberg:
> Is this ready? I thought the conmon-based SDNotify implementation was still unmerged, and the default OCI runtime implementation has massive issues and should never be used?

I think it's ready.  With `--sdnotify=conmon`, conmon isn't much involved as Podman sends the main PID and ready messages.

Issue #10529, opened 2021-06-02
remote events seems to lose data


vrothberg:
`podman-remote events` seems to lose data that the local client displays reliably (detected in https://github.com/containers/podman/pull/10514).  For reproducers, look for this issue's number in `test/system`.

PR #10558, opened 2021-06-04
remote events: fix --stream=false

vrothberg:
Fix a bug in remote events where only one event would be sent if when
streaming is turned off.  The source of the bug was that the handler
attempted to implement the streaming logic and did it wrong.  The fix is
rather simple by removing this logic from the handler and let the events
backend handle streaming.

Fixes: #10529


vrothberg:
Repushed to enable all events tests for remote :+1:

giuseppe (approved):
LGTM

PR #10561, opened 2021-06-04
remote events: support labels

vrothberg:
Certain event meta data was lost when converting the remote events to
libpod events and vice versa.  Enable the skipped system tests for
remote.


umohnani8:
LGTM

PR #10567, opened 2021-06-04
Add support for selectable checkpoint archive compression algorithm

adrianreber:
The checkpoint archive compression was hardcoded to `archive.Gzip`.

There have been requests to make the used compression algorithm selectable. There was especially the request to not compress the checkpoint archive to be able to create faster checkpoints when not compressing it.

This also changes the default from `gzip` to `zstd`. This change should not break anything as the restore code path automatically handles whatever compression the user provides during restore.

### DIFF
--- a/docs/source/markdown/podman-container-checkpoint.1.md
+++ b/docs/source/markdown/podman-container-checkpoint.1.md
@@ -10,31 +10,9 @@ podman\-container\-checkpoint - Checkpoints one or more running containers
 Checkpoints all the processes in one or more containers. You may use container IDs or names as input.
 
 ## OPTIONS
-#### **--keep**, **-k**
-
-Keep all temporary log and statistics files created by CRIU during checkpointing. These files
-are not deleted if checkpointing fails for further debugging. If checkpointing succeeds these
-files are theoretically not needed, but if these files are needed Podman can keep the files
-for further analysis.
-
 #### **--all**, **-a**
 
 Checkpoint all running containers.
-
-#### **--latest**, **-l**
-
-Instead of providing the container name or ID, checkpoint the last created container. (This option is not available with the remote Podman client)
-
-#### **--leave-running**, **-R**
-
-Leave the container running after checkpointing instead of stopping it.
-
-#### **--tcp-established**
-
-Checkpoint a container with established TCP connections. If the checkpoint
-image contains established TCP connections, this options is required during
-restore. Defaults to not checkpointing containers with established TCP
-connections.
 
 #### **--export**, **-e**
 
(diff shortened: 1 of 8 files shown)


pkg/domain/entities/containers.go
```diff
@@ -178,6 +178,7 @@ type CheckpointOptions struct {
 	TCPEstablished bool
 	PreCheckPoint  bool
 	WithPrevious   bool
+	Compression    int64
```
jwhonce, line 181:
```suggestion
	Compression    archive.Compression
```
Why not use the compression type vs. converting to and from int64?

jwhonce (commented):
@adrianreber Thanks for the PR. One question/suggestion.

pkg/domain/entities/containers.go
adrianreber: Initially I was using `archive.Compression`, but then I thought that it seems a bit much to include another package just to track 3 integer values. As I am also using this in my attempt to bring checkpoint/restore to CRI-O and I am including `libpod.CheckpointOptions` there and so `int64` seemed simpler. But `archive` is already part of Podman and CRI-O so that it doesn't really make a difference. It just felt like a too large dependency for 3 integer values. But I can easily change it back to `archive.Compression`.

jwhonce: In for a penny, in for a pound, you are already including the package so I would prefer the code being type safe. Thanks.

adrianreber: Switched everything to `archive.Compression`.

Luap99 (commented):
Please remove the make completions commit. For now we have to live with the current scripts as they are much better. When cobra releases a new version with the necessary bugfixes we can start using the auto generated ones.

adrianreber:
completion commit removed

docs/source/markdown/podman-container-checkpoint.1.md
```diff
@@ -56,11 +50,33 @@ This option must be used in combination with the **--export, -e** option.
...
+#### **--latest**, **-l**
+
+Instead of providing the container name or ID, checkpoint the last created container. (This option is not available with the remote Podman client)
+
 #### **--pre-checkpoint**, **-P**
 
 Dump the container's memory information only, leaving the container running. Later
 operations will supersede prior dumps. It only works on runc 1.0-rc3 or higher.
 
+#### **--leave-running**, **-R**
```
rhatdan, line 69: Could you alphabetize the options.  --leave-running should come before --pre-checkpoint.

adrianreber: I missed that one. Thanks. Fixed.

rhatdan:
/approve
LGTM

vrothberg (commented):
/lgtm

PR #10588, opened 2021-06-07
[CI:DOCS] converted query parameter for credentials to header parameter.

cdoern:

Edited docs based on the fact that [we added support for X-Registry-Auth](https://github.com/containers/podman/pull/6207) and the [libpod pull code](https://github.com/containers/podman/blob/v2.2/pkg/api/handlers/libpod/images_pull.go#L76) seems to only parse the header instead of the credentials query parameter. If interested, we could add handling for a credentials query parameter as well.

### DIFF
--- a/pkg/api/server/register_images.go
+++ b/pkg/api/server/register_images.go
@@ -948,10 +948,6 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//     description: "Mandatory reference to the image (e.g., quay.io/image/name:tag)"
 	//     type: string
 	//   - in: query
-	//     name: credentials
-	//     description: "username:password for the registry"
-	//     type: string
-	//   - in: query
 	//     name: Arch
 	//     description: Pull image for the specified architecture.
 	//     type: string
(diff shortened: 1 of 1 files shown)


rhatdan:
@jwhonce @baude PTAL

jwhonce:
@vrothberg Was there a use case you intended for the `credentials` query parameter?

vrothberg:
> @vrothberg Was there a use case you intended for the `credentials` query parameter?

I think that comes from a time from before we had the x-registry-... header plumbed.

cdoern:
@vrothberg @jwhonce is this good to merge or do we not want to switch to x-reg? It seems to me like `credentials` just is not used otherwise so if you wanted I could implement it on our end instead of just sending it in the header.

pkg/api/server/register_images.go
```diff
@@ -972,6 +968,10 @@ func (s *APIServer) registerImagesHandlers(r *mux.Router) error {
 	//     name: allTags
 	//     description: Pull all tagged images in the repository.
 	//     type: boolean
+	//   - in: header
+	//     name: X-Registry-Auth
+	//     description: "base-64 encoded auth config. Must include either: username, password, email and server address OR simply just an identity token."
```
TomSweeneyRedHat, line 973:
```suggestion
	//     description: "base-64 encoded auth config. Must include the following four values: username, password, email and server address OR simply just an identity token."
```

PR #10589, opened 2021-06-07
Add support for pod inside of user namespace.

umohnani8:
Add the --userns flag to podman pod create and keep
track of the userns setting that pod was created with
so that all containers created within the pod will inherit
that userns setting.

Specifically we need to be able to launch a pod with
--userns=keep-id

### DIFF
--- a/cmd/podman/containers/create.go
+++ b/cmd/podman/containers/create.go
@@ -184,6 +184,9 @@ func createInit(c *cobra.Command) error {
 	if c.Flag("cpu-quota").Changed && c.Flag("cpus").Changed {
 		return errors.Errorf("--cpu-quota and --cpus cannot be set together")
 	}
+	if c.Flag("pod").Changed && !strings.HasPrefix(c.Flag("pod").Value.String(), "new:") && c.Flag("userns").Changed {
+		return errors.Errorf("--userns and --pod cannot be set together")
+	}
 
 	noHosts, err := c.Flags().GetBool("no-hosts")
 	if err != nil {
@@ -309,6 +312,12 @@ func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions
 	if len(podName) < 1 {
 		return nil, errors.Errorf("new pod name must be at least one character")
 	}
+
+	userns, err := specgen.ParseUserNamespace(cliVals.UserNS)
+	if err != nil {
+		return nil, err
+	}
+
 	createOptions := entities.PodCreateOptions{
 		Name:          podName,
 		Infra:         true,
@@ -318,6 +327,7 @@ func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions
 		Cpus:          cliVals.CPUS,
 		CpusetCpus:    cliVals.CPUSetCPUs,
 		Pid:           cliVals.PID,
+		Userns:        userns,
 	}
 	// Unset config values we passed to the pod to prevent them being used twice for the container and pod.
 	s.ContainerBasicConfig.Hostname = ""
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -1123,9 +1123,9 @@ Podman allocates unique ranges of UIDs and GIDs from the `containers` subpordina
 
   Valid `auto`options:
 
-  - *gidmapping*=_HOST_GID:CONTAINER_GID:SIZE_: to force a GID mapping to be present in the user namespace.
+  - *gidmapping*=_CONTAINER_GID:HOST_GID:SIZE_: to force a GID mapping to be present in the user namespace.
   - *size*=_SIZE_: to specify an explicit size for the automatic user namespace. e.g. `--userns=auto:size=8192`. If `size` is not specified, `auto` will estimate a size for the user namespace.
-  - *uidmapping*=_HOST_UID:CONTAINER_UID:SIZE_: to force a UID mapping to be present in the user namespace.
+  - *uidmapping*=_CONTAINER_UID:HOST_UID:SIZE_: to force a UID mapping to be present in the user namespace.
 
 **container:**_id_: join the user namespace of the specified container.
 
--- a/docs/source/markdown/podman-pod-create.1.md
+++ b/docs/source/markdown/podman-pod-create.1.md
@@ -164,6 +164,19 @@ podman generates a UUID for each pod, and if a name is not assigned
 to the container with **--name** then a random string name will be generated
 for it. The name is useful any place you need to identify a pod.
 
+#### **--userns**=*mode*
+
+Set the user namespace mode for all the containers in a pod. It defaults to the **PODMAN_USERNS** environment variable. An empty value ("") means user namespaces are disabled.
+
+Valid _mode_ values are:
+
+- *auto[:*_OPTIONS,..._*]*: automatically create a namespace. It is possible to specify these options to `auto`:
+  - *gidmapping=*_CONTAINER_GID:HOST_GID:SIZE_ to force a GID mapping to be present in the user namespace.
+  - *size=*_SIZE_: to specify an explicit size for the automatic user namespace. e.g. `--userns=auto:size=8192`. If `size` is not specified, `auto` will estimate a size for the user namespace.
+  - *uidmapping=*_CONTAINER_UID:HOST_UID:SIZE_ to force a UID mapping to be present in the user namespace.
+- *host*: run in the user namespace of the caller. The processes running in the container will have the same privileges on the host as any other process launched by the calling user (default).
+- *keep-id*: creates a user namespace where the current rootless user's UID:GID are mapped to the same values in the container. This option is ignored for containers created by the root user.
+
 ## EXAMPLES
 
 ```
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -1181,9 +1181,9 @@ Podman allocates unique ranges of UIDs and GIDs from the `containers` subpordina
 
   Valid `auto`options:
 
-  - *gidmapping*=_HOST_GID:CONTAINER_GID:SIZE_: to force a GID mapping to be present in the user namespace.
+  - *gidmapping*=_CONTAINER_GID:HOST_GID:SIZE_: to force a GID mapping to be present in the user namespace.
   - *size*=_SIZE_: to specify an explicit size for the automatic user namespace. e.g. `--userns=auto:size=8192`. If `size` is not specified, `auto` will estimate a size for the user namespace.
-  - *uidmapping*=_HOST_UID:CONTAINER_UID:SIZE_: to force a UID mapping to be present in the user namespace.
+  - *uidmapping*=_CONTAINER_UID:HOST_UID:SIZE_: to force a UID mapping to be present in the user namespace.
 
 **container:**_id_: join the user namespace of the specified container.
 
--- a/libpod/define/pod_inspect.go
+++ b/libpod/define/pod_inspect.go
@@ -105,6 +105,8 @@ type InspectPodInfraConfig struct {
 	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
 	// Pid is the PID namespace mode of the pod's infra container
 	PidNS string `json:"pid_ns,omitempty"`
+	// UserNS is the usernamespace that all the containers in the pod will join.
+	UserNS string `json:"userns,omitempty"`
 }
 
 // InspectPodContainerInfo contains information on a container in a pod.
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -956,8 +956,9 @@ func WithUserNSFrom(nsCtr *Container) CtrCreateOption {
 		}
 
 		ctr.config.UserNsCtr = nsCtr.ID()
-		ctr.config.IDMappings = nsCtr.config.IDMappings
-
+		if err := JSONDeepCopy(nsCtr.IDMappings(), &ctr.config.IDMappings); err != nil {
+			return err
+		}
 		g := generate.Generator{Config: ctr.config.Spec}
 
 		g.ClearLinuxUIDMappings()
@@ -968,7 +969,6 @@ func WithUserNSFrom(nsCtr *Container) CtrCreateOption {
 		for _, gidmap := range nsCtr.config.IDMappings.GIDMap {
 			g.AddLinuxGIDMapping(uint32(gidmap.HostID), uint32(gidmap.ContainerID), uint32(gidmap.Size))
 		}
-		ctr.config.IDMappings = nsCtr.config.IDMappings
 		return nil
 	}
 }
@@ -2423,6 +2423,24 @@ func WithVolatile() CtrCreateOption {
 		}
 
 		ctr.config.Volatile = true
+
+		return nil
+	}
+}
+
+// WithPodUserns sets the userns for the infra container in a pod.
+func WithPodUserns(userns specgen.Namespace) PodCreateOption {
+	return func(pod *Pod) error {
+		if pod.valid {
+			return define.ErrPodFinalized
+		}
+
+		if !pod.config.InfraContainer.HasInfraContainer {
+			return errors.Wrapf(define.ErrInvalidArg, "cannot configure pod userns as no infra container is being created")
+		}
+
+		pod.config.InfraContainer.Userns = userns
+
 		return nil
 	}
 }
--- a/libpod/runtime_pod_infra_linux.go
+++ b/libpod/runtime_pod_infra_linux.go
@@ -8,7 +8,9 @@ import (
 
 	"github.com/containers/common/pkg/config"
 	"github.com/containers/podman/v3/libpod/define"
+	"github.com/containers/podman/v3/pkg/namespaces"
 	"github.com/containers/podman/v3/pkg/rootless"
+	"github.com/containers/podman/v3/pkg/specgen"
 	"github.com/containers/podman/v3/pkg/util"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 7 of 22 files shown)


umohnani8:
@rhatdan PTAL

libpod/options.go
```diff
@@ -2339,6 +2340,18 @@ func WithVolatile() CtrCreateOption {
 		}
 
 		ctr.config.Volatile = true
+	}
```
rhatdan, line 2343: There is a compilation bug here.missing a return nil

rhatdan:
I applied this to the master branch and I am getting weird errors?

/bin/podman pod create --userns=keep-id --name dan1
ERRO[0000] Error freeing pod lock after failed creation: no such file or directory 
Error: error adding Infra Container: please set User explicitly via WithUser() instead of in OCI spec directly: invalid argument

libpod/options.go
umohnani8: Fixed

umohnani8:
@rhatdan should be fixed now. I rebased before pushing it up and didn't realise the rebase broke it xD

rhatdan:
Still looks like it does not work, and tests are not happy.

```
 $ ./bin/podman pod create --userns=keep-id --name dan1
cb862419ce6a0b79b5876cbe32e2acad8cefb90d99ccfdd8864f828da5134f96
$ ./bin/podman run --pod dan1 fedora id
ERRO[0000] error starting some container dependencies   
ERRO[0000] "open /proc/sys/net/ipv4/ping_group_range: Permission denied: OCI permission denied" 
Error: error starting some containers: internal libpod error
```

cmd/podman/containers/create.go
```diff
@@ -324,3 +339,20 @@ func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions
 	s.Pod = podName
 	return registry.ContainerEngine().PodCreate(context.Background(), createOptions)
 }
+
+// inheritPodUserNS gets the userns option that the pod was created with and sets that
+// for the container being created (if the container is being created inside a pod).
+func inheritPodUserNS(s *specgen.SpecGenerator) error {
```
rhatdan, line 345: I think this should happen on the server side.  If I launch a container within a pod via the API, I want this to happen also.

pkg/specgen/generate/namespaces.go
```diff
@@ -506,3 +469,46 @@ func GetNamespaceOptions(ns []string) ([]libpod.PodCreateOption, error) {
 	}
 	return options, nil
 }
+
+func SetupUserNS(idmappings *storage.IDMappingOptions, userns specgen.Namespace, g *generate.Generator) error {
```
rhatdan, line 473: Can't this SetupUserNS be combined with the setupUserNS above?

rhatdan (approved):

cmd/podman/containers/create.go
```diff
@@ -324,3 +339,20 @@ func createPodIfNecessary(s *specgen.SpecGenerator, netOpts *entities.NetOptions
...
+		return nil
+	}
+	inspectOptions := entities.PodInspectOptions{
+		NameOrID: s.Pod,
+	}
+	data, err := registry.ContainerEngine().PodInspect(context.Background(), inspectOptions)
+	if err != nil {
+		return errors.Errorf("couldn't inspect pod %q", s.Pod)
+	}
+	s.UserNS, err = specgen.ParseUserNamespace(data.InfraConfig.UserNS)
```
giuseppe, line 356: should the effect be something like `WithUserNSFrom()` and join the user namespace for the infra container?

rhatdan:
Fixes: https://github.com/containers/podman/issues/7706

umohnani8:
This is ready now. @rhatdan @mheon @giuseppe PTAL

rhatdan:
Tests are not happy?

rhatdan:
Code looks good and tests look great.

libpod/define/pod_inspect.go
```diff
@@ -103,6 +103,8 @@ type InspectPodInfraConfig struct {
 	CPUQuota int64 `json:"cpu_quota,omitempty"`
 	// CPUSetCPUs contains linux specific CPU data for the container
 	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
+	// UserNS is the usernamespace that all the containers in the pod will join.
+	UserNS string
```
mheon, line 107: Should this be omitempty?

libpod/options.go
```diff
@@ -2355,6 +2356,20 @@ func WithVolatile() CtrCreateOption {
 		}
 
 		ctr.config.Volatile = true
+
+		return nil
+	}
+}
+
+// WithInfraUserns sets the userns for the infra container in a pod.
+func WithInfraUserns(userns specgen.Namespace) PodCreateOption {
```
mheon, line 2365: Suggest renaming to "WithPodUserNS(...)`

libpod/runtime_pod_infra_linux.go
```diff
@@ -146,11 +146,32 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
...
+	mapopt, err := util.ParseIDMapping(namespaces.UsernsMode(p.config.InfraContainer.Userns.String()), []string{}, []string{}, "", "")
+	if err != nil {
+		return nil, err
+	}
+	user, err := specgen.SetupUserNS(mapopt, p.config.InfraContainer.Userns, &g)
+	if err != nil {
+		return nil, err
+	}
+	if user != "" {
+		options = append(options, WithUser(user))
```
mheon, line 172: Do we need this? I think this is only for keep-id and it doesn't really make sense for infra containers - other containers in the pod won't inherit it.

umohnani8: The pod fails to create without this when using keep-id. Should I make it a condition to only do this when userns is set to keep-id?
This is the error I get 
```
Error: error adding Infra Container: please set User explicitly via WithUser() instead of in OCI spec directly: invalid argument
```

libpod/options.go
umohnani8: Fixed

libpod/define/pod_inspect.go
umohnani8: Fixed

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -152,6 +152,19 @@ podman generates a UUID for each pod, and if a name is not assigned
 to the container with **--name** then a random string name will be generated
 for it. The name is useful any place you need to identify a pod.
 
+#### **--userns**=*mode*
+
+Set the user namespace mode for all the containers in a pod. It defaults to the **PODMAN_USERNS** environment variable. An empty value ("") means user namespaces are disabled.
+
+Valid _mode_ values are:
+
+- **auto[:**_OPTIONS,..._**]**: automatically create a namespace. It is possible to specify these options to `auto`:
```
TomSweeneyRedHat, line 161: Unless I'm misinterpreting https://raw.githubusercontent.com/containers/podman/main/docs/MANPAGE_SYNTAX.md, I think these and the ones following should have a single and not a double asterisk

```suggestion
- *auto[:*_OPTIONS,..._*]*: automatically create a namespace. It is possible to specify these options to `auto`:
```

docs/source/markdown/podman-run.1.md
```diff
@@ -1160,9 +1160,9 @@ Set the user namespace mode for the container. It defaults to the **PODMAN_USERN
 Valid _mode_ values are:
 
 - **auto[:**_OPTIONS,..._**]**: automatically create a namespace. It is possible to specify these options to `auto`:
-  - **gidmapping=**_HOST_GID:CONTAINER_GID:SIZE_: to force a GID mapping to be present in the user namespace.
+  - **gidmapping=**_CONTAINER_GID:HOST_GID:SIZE_: to force a GID mapping to be present in the user namespace.
   - **size=**_SIZE_: to specify an explicit size for the automatic user namespace. e.g. `--userns=auto:size=8192`. If `size` is not specified, `auto` will estimate a size for the user namespace.
-  - **uidmapping=**_HOST_UID:CONTAINER_UID:SIZE_: to force a UID mapping to be present in the user namespace.
+  - **uidmapping=**_CONTAINER_UID:HOST_UID:SIZE_: to force a UID mapping to be present in the user namespace.
```
TomSweeneyRedHat, line 1165: ditto previous

docs/source/markdown/podman-create.1.md
```diff
@@ -1087,9 +1087,9 @@ Set the user namespace mode for the container. It defaults to the **PODMAN_USERN
 Valid _mode_ values are:
 
 - **auto[:**_OPTIONS,..._**]**: automatically create a namespace. It is possible to specify these options to `auto`:
-  - **gidmapping=**_HOST_GID:CONTAINER_GID:SIZE_: to force a GID mapping to be present in the user namespace.
+  - **gidmapping=**_CONTAINER_GID:HOST_GID:SIZE_: to force a GID mapping to be present in the user namespace.
```
TomSweeneyRedHat, line 1090: See next files comment

libpod/runtime_pod_infra_linux.go
mheon: You should be able to clear the User field in the OCI spec (set it to "" or root:root) and that will clear the error

libpod/runtime_pod_infra_linux.go
```diff
@@ -146,11 +146,32 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
...
+	if err != nil {
+		return nil, err
+	}
+	_, err = specgen.SetupUserNS(mapopt, p.config.InfraContainer.Userns, &g)
+	if err != nil {
+		return nil, err
+	}
+	// if user != "" {
+	// 	options = append(options, WithUser(user))
+	// }
```
giuseppe, line 173: do we need this code (and uncomment it) or can we drop it?

giuseppe (commented):
could we also add a check to error out when the container joins a pod that uses a user namespace and it also specifies an uidmapping/gidmapping?

e.g. `podman run --pod pod-with-userns --uidmapping 0:0:1000` must fail

docs/source/markdown/podman-pod-create.1.md
umohnani8: I am not sure, but all over the doc options like these are in bold vs italics (which is what single asterisk does). I do think the bold is easier to read than the italics.

rhatdan:
> could we also add a check to error out when the container joins a pod that uses a user namespace and it also specifies an uidmapping/gidmapping?
> 
> e.g. `podman run --pod pod-with-userns --uidmapping 0:0:1000` must fail

I think we should block setting of uidmappings or any modification of the user namespace when joining a pod.  If you are going to be in a Pod, then you can not manipulate the user namespace.

giuseppe:
> I think we should block setting of uidmappings or any modification of the user namespace when joining a pod. If you are going to be in a Pod, then you can not manipulate the user namespace via the container creation. You have to modify the Pod

Agreed.  At the moment we are limited by the oci runtime and specs.  At the kernel level, there is nothing that prevents joining the pod user namespace and then create a new child namespace.

rhatdan:
Sure you can do this within a user namespace, but I think there will be unexpected behavior of a pod if we allow users to specify multiple userns.  Imagine two containers within a Pod attempting to share a volume each one in a different USERNS.  The complexity is to great, for questionable improvements in Security.

test/e2e/pod_create_test.go
```diff
@@ -559,4 +560,217 @@ ENTRYPOINT ["sleep","99999"]
 		podJSON := podInspect.InspectPodToJSON()
 		Expect(podJSON.CPUSetCPUs).To(Equal(in))
 	})
+
+	It("podman pod create with --userns=keep-id", func() {
+		podName := "testPod"
+		podCreate := podmanTest.Podman([]string{"pod", "create", "--userns", "keep-id", "--name", podName})
+		podCreate.WaitWithDefaultTimeout()
+		Expect(podCreate.ExitCode()).To(Equal(0))
```
edsantiago, line 568: This pattern is anti-recommended, please see #10997. Could you fix it (and all instances) to use the recommended form (`Expect(podCreate).To(Exit(0))`)? Thank you!

edsantiago (changes requested):

docs/source/markdown/podman-pod-create.1.md
umohnani8: Fixed

libpod/runtime_pod_infra_linux.go
umohnani8: Fixed

test/e2e/pod_create_test.go
umohnani8: Fixed

pkg/specgen/generate/namespaces.go
```diff
@@ -175,6 +175,10 @@ func namespaceOptions(ctx context.Context, s *specgen.SpecGenerator, rt *libpod.
 		if pod == nil || infraCtr == nil {
 			return nil, errNoInfra
 		}
+		// Inherit the user from the infra container if it is set
+		if infraCtr.User() != "" {
```
umohnani8, line 179: @giuseppe @mheon @rhatdan does it make sense to allow containers to set `--user` when being created with `--pod`? For the `--userns=keep-id` case, the container wasn't being created with the calling user even though the uid_map was correctly set in the container, so I had to explicitly inherit the user as well from the infra container.
I am not sure if this is the right approach, but if it is, should we be able to set `--user` when creating a container in a pod with userns set?

Before explicitly setting the user to the infra user
```
podman exec -it 5be9d20f7cb14aad964fb7e8f6cc81acd3a60b8822d19631df7496f0e1b83ab7 sh
sh-5.1# id
uid=0(root) gid=0(root) groups=0(root)
sh-5.1# cat /proc/self/uid_map 
         0          1       1000
      1000          0          1
      1001       1001      64536
```

After setting it (which is what we should be getting I believe)
```
✗ ./bin/podman run --pod foo -it fedora sh
sh-5.1$ id
uid=1000(1000) gid=1000(1000) groups=1000(1000)
sh-5.1$ cat /proc/self/uid_map 
         0          1       1000
      1000          0          1
      1001       1001      64536
```

mheon: Yes, we definitely want to allow `--user` and `--pod` to be used together; having a DB server running as user `mysql` and a web server running as user `httpd` in the same pod sounds like a very common use-case.

umohnani8: Okay fixed it, here is an example output of the options
```
➜  podman git:(pod-userns) ✗ ./bin/podman pod create --userns keep-id --name foo
ae9120da729b2cf56b129fe5641f362ae06dcac6b8f3c9294f947e5f93254a13
➜  podman git:(pod-userns) ✗ ./bin/podman run --pod foo -it fedora sh           
sh-5.1$ id
uid=1000(1000) gid=1000(1000) groups=1000(1000)
sh-5.1$ 
exit
➜  podman git:(pod-userns) ✗ ./bin/podman run --pod foo -it --user 2000 fedora sh
sh-5.1$ id
uid=2000(2000) gid=0(root) groups=0(root)
sh-5.1$ 
```

rhatdan: LGTM

TomSweeneyRedHat:
@umohnani8 tests still aren't hip with you.
Otherwise LGTM

libpod/runtime_pod_infra_linux.go
umohnani8: @mheon we want containers inside a pod to inherit the user from the infra container when --userns=keep-id and the container was not created with the --user flag. This kind of links to the discussion we had in https://github.com/containers/podman/pull/10589#discussion_r677718549.
I think for that case, we should leave this as is instead of always adding `root:root` to the infra container, wdyt?

umohnani8:
@giuseppe @rhatdan any idea why the userns tests are unhappy only on ubuntu-2010 - https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4754045987454976/html/int-podman-ubuntu-2010-rootless-host.log.html#t--podman-pod-create-with---userns-auto--1

libpod/runtime_pod_infra_linux.go
mheon: If we're using it to have other containers inherit it, that's probably OK, but I'm still worried that running infra images as non-root users may not be very durable with user-supplied infra images.

rhatdan: Could this be a runc versus crun issue?

rhatdan:
@vrothberg @umohnani8 Could this be a runc versus crun issue?

vrothberg:
> @vrothberg @umohnani8 Could this be a runc versus crun issue?

Probably but I had to read the code in detail. @giuseppe does the below error sound familiar to you?

> podman [options] run --pod testPod0 quay.io/libpod/alpine:latest cat /proc/self/uid_map
         Error: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "sysfs" to rootfs at "/sys" caused: mount through procfd: operation not permitted: OCI permission denied

giuseppe:
how does the generated `config.json` file look like with the failing test?

giuseppe:
I see it generates:

```
    "namespaces": [
      {
        "type": "pid"
      },
      {
        "type": "network",
        "path": "/proc/122177/ns/net"
      },
      {
        "type": "ipc",
        "path": "/proc/122177/ns/ipc"
      },
      {
        "type": "uts",
        "path": "/proc/122177/ns/uts"
      },
      {
        "type": "mount"
      },
      {
        "type": "cgroup"
      },
      {
        "type": "user"
      }
    ],
```

The user namespace should also have the path set to the infraContainer as for the network, ipc and uts namespaces.

umohnani8:
Tests are finally passing now, can I please get some reviews here @rhatdan @mheon @giuseppe

libpod/runtime_pod_infra_linux.go
umohnani8: @mheon I will open up a follow PR for the user work. Setting the user to root:root always for the infra container is causing some other issues. Can we get this in for now while I debug that?

giuseppe (approved):
LGTM

rhatdan:
/lgtm
Congratulations @umohnani8 I started working on this last November, and you finally pushed it over the finish line.  Great work.
This was a very difficult problem.

rlpowell:
Thank you so much!

As a Fedora user, how can I track when this will be available there?  Without pulling down pre-release packages, I mean.

rhatdan:
podman 3.4 or later, we missed the cut, I believe for podman 3.3.

francoism90:
Has this been removed? I'm keep getting:
```
Error: --userns and --pod cannot be set together
```
podman version 4.4.4

giuseppe:
> Error: --userns and --pod cannot be set together

this is a different error.  If you are joining an existing user namespace (with `--pod`), you cannot also create a new one with `--userns`

francoism90:
@giuseppe Thanks for your reply and sorry for posting on the wrong PR.

If I understand correctly, and this to work, is to use idsmapping instead, right?

giuseppe:
it depends, not sure what you'd like to get.  What is the expectation you've from the command you've used?

Were you trying to create a new user namespace?

rhatdan:
The container you are adding is going to join the pod's user namespace.  We don't allow containers inside of the same pod to run with different user namespaces, because too much stuff will break in unexpected ways.

rhatdan:
If you are using podman-compose you should bring the issue up there.

PR #10592, opened 2021-06-07
auto-update tests: various fixes

edsantiago:
Followup to #9740. Nice work, but the _confirm_update() helper
was dangerously broken and I just wasn't able to communicate
that. Given the time zone difference, and my weekly time cost
in reviewing, it's easier for me to fix it myself. (The problem
is that the function was a complete NOP, which would lead to
flakes).

Also: got rid of some clutter, restructured a few minor places
for maintainability.

### DIFF
--- a/test/system/255-auto-update.bats
+++ b/test/system/255-auto-update.bats
@@ -43,12 +43,12 @@ function teardown() {
 #   5. Remove the origin container
 #   6. Start the container from service
 function generate_service() {
-    target_img_basename=$1
-    autoupdate=$2
+    local target_img_basename=$1
+    local autoupdate=$2
 
-    # Please keep variable name for cname and ori_image. The
-    # scripts will use them directly in following tests.
-    cname=c_$(random_string)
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

test/system/255-auto-update.bats
```diff
@@ -89,91 +90,87 @@ function _wait_service_ready() {
...
-        if [[ "$output" == "$last_log" ]]; then
+    local cname=$1
+    local old_iid=$2
+
+    # Image has already been pulled, so this shouldn't take too long
+    local timeout=5
+    while [[ $timeout -gt 0 ]]; do
+        run_podman '?' inspect --format "{{.Image}}" $cname
+        if [[ $status != 0 ]]; then
+            if [[ $output =~ (no such object|does not exist in database): ]]; then
```
edsantiago, line 103: This was the hardest part to get right. Sometimes it's one of those errors, sometimes the other.

vrothberg (commented):
/lgtm

Issue #10564, opened 2021-06-04
API endpoint "network prune" returns array of objects instead of array of stringfor field "Bytes"


arctic-alpaca:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
Hi,
the [API reference for network list ](https://docs.podman.io/en/latest/_static/api.html#operation/NetworkListLibpod) describes field `Bytes` as `Array of integers <uint8>`, boht for the initial result and the `Plugins` field. The API returns a base64 encoded string instead.

**Steps to reproduce the issue:**

1. ``podman system service unix:///home/`whoami`/testing.sock --log-level=debug --time=500``

2. `podman network create testns`

3. ``` 
   curl --unix-socket /home/`whoami`/testing.sock "http://d/v3.0.0/libpod/networks/json"
[...]

PR #10597, opened 2021-06-08
Fix network prune api docs

Luap99:
The api doc used wrong response examples for both the compat and libpod
network prune endpoints. Change the doc so that it matches the actual
return values. Also fix the endpoints to return an empty array instead
of null when no networks are removed.

Fixes: #10564


vrothberg (commented):
LGTM

Issue #10570, opened 2021-06-05
Rootful podman cni dns resolution defaults to 8.8.8.8, are there ways for it to auto detect the ones the host is using? 


KCSesh:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

**Disclaimer: I don't understand all of the network details of the host I am accessing in an offsite location.** 

I am accessing a host through a VPN which has podman. 
The host seems to have a local nameserver running: 

Host file: 
```
cat /etc/resolv.conf
# Generated by NetworkManager
nameserver 127.0.0.53
```

If I were to change this file to use the nameserver 8.8.8.8, the host will not properly do DNS. And simple commands like: `curl google.com` stop working. 

So when I start a Podman container, it does not have DNS because the /etc/resolv.conf looks like this by default: 

```
[root@fc20b160c79a /]# cat /etc/resolv.conf
nameserver 8.8.8.8
nameserver 8.8.4.4
nameserver 2001:4860:4860::8888
nameserver 2001:4860:4860::8844
```

If I do the same test with Docker, docker seems to be able to fetch the DNS hosts that 127.0.0.53, and docker containers run with DNS working on the exact same host with no changes. 

This is the docker: /etc/resolve.conf
Full IP's are omitted. 
```
sudo docker run --rm  -it fedora
[root@b7809e842f47 /]# cat /etc/resolv.conf
# This file is managed by man:systemd-resolved(8). Do not edit.
#
# This is a dynamic resolv.conf file for connecting local clients directly to
# all known uplink DNS servers. This file lists all configured search domains.
#
# Third party programs must not access this file directly, but only through the
# symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a different way,
# replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 10.---.--.--
nameserver 10.---.--.--
# Too many DNS servers configured, the following entries may be ignored.
nameserver 10.---.---.--
nameserver 10.---.--.--
search nvidia.com 
```

I I manually set the DNS of my podman container with the runtime flag of `--dns=10.---.--.--` and DNS is fixed in the podman container. 

I would like to know if there is a plan to make the CNI detect the actual DNS instead of what seems to be a hardcoding of `8.8.8.8`

**Steps to reproduce the issue:**

[...]

PR #10598, opened 2021-06-08
Improve systemd-resolved detection

Luap99:
When 127.0.0.53 is the only nameserver in /etc/resolv.conf assume
systemd-resolved is used. This is better because /etc/resolv.conf does
not have to be symlinked to /run/systemd/resolve/stub-resolv.conf in
order to use systemd-resolved.

[NO TESTS NEEDED]

Fixes: #10570

### DIFF
--- a/libpod/container_internal_linux.go
+++ b/libpod/container_internal_linux.go
@@ -1650,22 +1650,20 @@ func (c *Container) generateResolvConf() (string, error) {
 		}
 	}
 
-	// Determine the endpoint for resolv.conf in case it is a symlink
-	resolvPath, err := filepath.EvalSymlinks(resolvConf)
+	contents, err := ioutil.ReadFile(resolvConf)
 	// resolv.conf doesn't have to exists
 	if err != nil && !os.IsNotExist(err) {
 		return "", err
 	}
 
-	// Determine if symlink points to any of the systemd-resolved files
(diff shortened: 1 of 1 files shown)


libpod/container_internal_linux.go
```diff
@@ -1650,22 +1650,20 @@ func (c *Container) generateResolvConf() (string, error) {
...
-	// resolv.conf doesn't have to exists
-	if err != nil && !os.IsNotExist(err) {
-		return "", err
+	ns := resolvconf.GetNameservers(contents)
+	// check if systemd-resolved is used, assume it is used when 127.0.0.53 is the only nameserver
+	if len(ns) == 1 && ns[0] == "127.0.0.53" {
+		// read the actual resolv.conf file for systemd-resolved
+		contents, err = ioutil.ReadFile("/run/systemd/resolve/resolv.conf")
+		if err != nil {
+			return "", err
```
mheon, line 1665: Maybe improve the error here - `errors.Wrapf(err, "detected that systemd-resolved is in use, but could no locate real resolv.conf")` or similar

mheon:
One nit, otherwise LGTM

libpod/container_internal_linux.go
Luap99: fixed

Issue #10596, opened 2021-06-08
podman logs -f ends early


Hendrik-H:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

A container traps SIGINT and SIGTERM, logs a few things, waits a few seconds, logs more and then exits. A parallel `podman logs -f ...` for that container does not capture all the logs and ends early when the container is stopped using `podman stop`. I believe I also had a case that a bit was missing when stopping with `ctrl-c` but I can not reproduce that anymore.

Test Dockerfile:
```bash
FROM fedora:34
ADD script.sh /
ENTRYPOINT ["/script.sh"]
```

Test Script:
```bash
#!/bin/bash

sigIntHandler() {
    echo "received SIGINT"
    echo "going to exit soon"
    sleep 5
    echo "exiting now"
    exit 1
}

sigTermHandler() {
    echo "received SIGTERM"
    echo "going to exit soon"
    sleep 5
    echo "exiting now"
    exit 1
}

trap sigIntHandler  SIGINT
trap sigTermHandler SIGTERM
trap

echo "started"
for i in {1..30}
do
    sleep 1
done
echo "done"
```

**Steps to reproduce the issue:**

1. build a test image using the above files: `podman build -t=podman-test .`
2. start a test container: `podman run -it --name test --rm podman-test`
3. follow the logs: `podman logs -f test`
4. stop the container using `podman stop --time 10 test`

**Describe the results you received:**
When using `podman stop` the logs stop straight away in the `podman logs -f` terminal and some logs that are still coming out in the `podman run` terminal are missing. When you try to restart `podman logs -f` while the container is in the `stopping` state you can not follow the logs anymore.

**Describe the results you expected:**
`podman logs -f` to log out all logs until the container is fully stopped and not just until it is starting to stop.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.15.11
Built:        Tue May 11 15:53:47 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc33.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 12
  distribution:
[...]

PR #10600, opened 2021-06-08
logs: k8s-file: fix race

vrothberg:
Fix a race in the k8s-file logs driver.  When "following" the logs,
Podman will print the container's logs until the end.  Previously,
Podman logged until the state transitioned into something non-running
which opened up a race with the container still running, possibly in
the "stopping" state.

To fix the race, log until we've seen the wait event for the specific
container.  In that case, conmon will have finished writing all logs to
the file, and Podman will read it until EOF.

Further tweak the integration tests for testing `logs -f` on a  running
container.  Previously, the test only checked for one of two lines
stating that there was a race.  Indeed the race was in using `run --rm`
where a log file may be removed before we could fully read it.

Fixes: #10596

### DIFF
--- a/libpod/container_log.go
+++ b/libpod/container_log.go
@@ -4,11 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"time"
 
 	"github.com/containers/podman/v3/libpod/define"
+	"github.com/containers/podman/v3/libpod/events"
 	"github.com/containers/podman/v3/libpod/logs"
-	"github.com/hpcloud/tail/watch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


libpod/container_log.go
```diff
@@ -94,28 +93,34 @@ func (c *Container) readFromLogFile(ctx context.Context, options *logs.LogOption
...
+			errorChannel := make(chan error)
+			eventOptions := events.ReadOptions{
+				EventChannel: eventChannel,
+				Filters:      []string{"event=died", "container=" + c.ID()},
+				FromStart:    true,
+				Stream:       true,
 			}
+			go func() {
+				errorChannel <- c.runtime.Events(ctx, eventOptions)
+			}()
```
Luap99, line 123: I don't understand this. You never read the event channel and this will just hang forever and never exit.

Luap99 (changes requested):
I don't think this works.

libpod/container_log.go
vrothberg: Absolutely, thanks!  Took a phone call and thought I added the remaining bits :^)

vrothberg:
> Also take a look at the following for a possible test:
> 
> https://github.com/containers/podman/blob/9a3a7327fdafaac66c99130a6729e4bcde8df0b0/test/e2e/logs_test.go#L186-L193

Good catch.  With the recent fixes in the journald logger and this PR, we should be good to go.  I'll take a look.

Luap99 (approved):
LGTM

maybe-sybr:
@mheon I've also hit the issue introduced by 84b55eec2 in 3.2.1 (picked as c3f6ef63a) and would appreciate if you could include this patch in 3.2.2 :)

vrothberg:
> @mheon I've also hit the issue introduced by [84b55ee](https://github.com/containers/podman/commit/84b55eec27967d1a528c4f49f100393668638e1a) in 3.2.1 (picked as [c3f6ef6](https://github.com/containers/podman/commit/c3f6ef63a27ae23f79995c4019d552ed4dfdd406)) and would appreciate if you could include this patch in 3.2.2 :)

Roger :) I backported a fix for the fix yesterday to the v3.2 branch, so it'll be fixed in 3.2.2.

PR #10603, opened 2021-06-08
implemented verbose and scope as possible

cdoern:

partial implementation of scope and verbose for /networks/{id} query. Not much else is possible for these two. Returns Bad Request is scope is anything other than local. The values docker uses for verbose output are not implemented in podman from what I can tell, and implementing would not save any space/time.

### DIFF
--- a/pkg/api/handlers/compat/networks.go
+++ b/pkg/api/handlers/compat/networks.go
@@ -28,19 +28,24 @@ import (
 func InspectNetwork(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 
-	// FYI scope and version are currently unused but are described by the API
-	// Leaving this for if/when we have to enable these
-	// query := struct {
-	//	scope   string
-	//	verbose bool
-	// }{
-	//	// override any golang type defaults
-	// }
-	// decoder := r.Context().Value("decoder").(*schema.Decoder)
-	// if err := decoder.Decode(&query, r.URL.Query()); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/api/handlers/compat/networks.go
```diff
@@ -28,19 +28,24 @@ import (
...
-	//	verbose bool
-	// }{
-	//	// override any golang type defaults
-	// }
-	// decoder := r.Context().Value("decoder").(*schema.Decoder)
-	// if err := decoder.Decode(&query, r.URL.Query()); err != nil {
-	//	utils.Error(w, "Something went wrong.", http.StatusBadRequest, errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
-	//	return
-	// }
+	//scope is only used to see if th euser passes any illegal value, verbose is not used but implemented
```
jwhonce, line 31:
```suggestion
	// scope is only used to see if the user passes any illegal value, verbose is not used but implemented
```

pkg/api/handlers/compat/networks.go
```diff
@@ -28,19 +28,24 @@ import (
...
+	// for compatability purposes only.
+	query := struct {
+		scope   string
+		verbose bool
+	}{
+		// override any golang type defaults
```
jwhonce, line 37: Given the code below should scope be defaulted to "local"?

jwhonce (changes requested):
Nit on comments and one question on logic.

You also need to either tweak existing or add tests, using these query parameters.

pkg/api/handlers/compat/networks.go
cdoern: okay, if I do that should I then remove the http bad request check? or leave it even though it'll be redundant.

jwhonce (approved):

jwhonce:
LGTM

PR #10604, opened 2021-06-08
[v3.0.1-rhel] rmi: don't break when the image is missing a manifest

nalind:
In libpod/image.Image.Remove(), if the attempt to find the image's parent fails for any reason, log a warning and proceed as though it didn't have one instead of failing, which would leave us unable to remove the image without resetting everything.

In libpod/Runtime.RemoveImage(), if we can't determine if an image has children, log a warning, and assume that it doesn't have any instead of failing, which would leave us unable to remove the image without resetting everything.

In pkg/domain/infra/abi.ImageEngine.Remove(), when attempting to remove all images, if we encounter an error checking if a given image has children, log a warning, and assume that it doesn't have any instead of failing, which would leave us unable to remove the image without resetting everything.

Cherry-picked from https://github.com/containers/podman/pull/10033, with an additional hunk adding $_PODMAN_TEST_OPTS to the run_podman() test function.

### DIFF
--- a/libpod/image/image.go
+++ b/libpod/image/image.go
@@ -622,7 +622,8 @@ func (i *Image) TopLayer() string {
 func (i *Image) Remove(ctx context.Context, force bool) error {
 	parent, err := i.GetParent(ctx)
 	if err != nil {
-		return err
+		logrus.Warnf("error determining parent of image: %v, ignoring the error", err)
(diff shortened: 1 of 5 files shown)


mheon:
In the future, please include the BZ number this addresses in the commit message

PR #123, opened 2017-12-12
\t was not being recognized as tab in --format

umohnani8:
When doing kpod images --format "{{.ID}}\t{{.Tag}}"
the "\t" was being passed in as a string of "\" and "t"
instead of as a tab character.

Signed-off-by: umohnani8 <umohnani@redhat.com>

PR #10610, opened 2021-06-09
Edited compat handling code for containers/json status and added python tests

### LINKED ISSUE #10457: Healthcheck is ignored in compose files when adding labels to volumes
/kind bug

**Description**

It appears that podman is ignoring the `healtcheck` configuration in a compose file if you add a `:Z` or a `:z` label to a volume.

1. `git clone https://github.com/os-autoinst/openQA.git && cd container/webui && sed -i 's|./workdir/db:/var/lib/postgresql/data|./workdir/db:/var/lib/postgresql/data:z|' docker-compose.yaml`

2. `docker-compose up`

```ShellSession
$ docker-compose up
Creating webui_db_1            ... done
Creating webui_webui_db_init_1 ... done

ERROR: for webui  Service "db" is missing a healthcheck configuration
ERROR: Encountered errors while bringing up the project.
```

Note that not adding the `:Z` label is not a solution, because it only makes the docker compose fail later (and the `db` container is actually not working at that point as it can't write `./workdir/db/`):
```
Creating webui_webui_db_init_1 ... done
Creating webui_db_1            ... done
Creating webui_webui_1         ... done
Creating webui_webui_2         ... done

ERROR: for scheduler  Service "webui" is missing a healthcheck configuration

ERROR: for livehandler  Service "webui" is missing a healthcheck configuration

ERROR: for gru  Service "webui" is missing a healthcheck configuration

ERROR: for websockets  Service "webui" is missing a healthcheck configuration
ERROR: Encountered errors while bringing up the project.
```

The compose should succeed. 

- This is always reproducible.
- tried with both docker-compose 1.28.6 from Fedora's repositories and the latest release 1.29.2 installed via pip.

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16.3
Built:        Wed May 12 21:27:59 2021
OS/Arch:      linux/amd64
```

```
host:
  arch: amd64
  buildahVersion: 1.20.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 4
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: Cerberus
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 10000
      size: 65536
    - container_id: 65537
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 10000
      size: 65536
    - container_id: 65537
      host_id: 100000
      size: 65536
  kernel: 5.12.5-300.fc34.x86_64
  linkmode: dynamic
  memFree: 198840320
  memTotal: 8259194880
  ociRuntime:
    name: crun
    package: crun-0.19.1-2.fc34.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.19.1
      commit: 1535fedf0b83fb898d449f9680000f729ba719f5
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    selinuxEnabled: true
  slirp4netns:
    package: slirp4netns-1.1.9-1.fc34.x86_64
    version: |-
      slirp4netns version 1.1.8+dev
      commit: 6dc0186e020232ae1a6fcc1f7afbc3ea02fd3876
      libslirp: 4.4.0
      SLIRP_CONFIG_VERSION_MAX: 3
      libseccomp: 2.5.0
  swapFree: 14715969536
  swapTotal: 16676544512
  uptime: 9h 52m 50.38s (Approximately 0.38 days)
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - docker.io
  - quay.io
store:
  configFile: /home/dan/.config/containers/storage.conf
  containerStore:
    number: 3
    paused: 0
    running: 0
    stopped: 3
  graphDriverName: overlay
  graphOptions:
    overlay.mount_program:
      Package: fuse-overlayfs-1.5.0-1.fc34.x86_64
      Version: |-
        fusermount3 version: 3.10.3
        fuse-overlayfs: version 1.5
        FUSE library version 3.10.3
  graphRoot: /home/dan/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 120
  volumePath: /home/dan/.local/share/containers/storage/volumes
version:
  APIVersion: 3.1.2
  Built: 1620847679
  BuiltTime: Wed May 12 21:27:59 2021
  GitCommit: ""
  GoVersion: go1.16.3
  OsArch: linux/amd64
  Version: 3.1.2
```

```
podman-3.1.2-3.fc34.x86_64
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide? (https://github.com/containers/podman/blob/master/troubleshooting.md)**

Yes

I have followed this guide to setup podman with compose: https://www.redhat.com/sysadmin/podman-docker-compose

cdoern:

Health Check is not handled in the compat LibpodToContainerJSON function within containers.go. Added parsing and handling for this status check and also edited some of the list container filter code.

fixes #10457

### DIFF
--- a/pkg/api/handlers/compat/containers.go
+++ b/pkg/api/handlers/compat/containers.go
@@ -403,6 +403,24 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON, 
 		state.Status = define.ContainerStateCreated.String()
 	}
 
+	state.Health = &types.Health{
+		Status:        inspect.State.Healthcheck.Status,
+		FailingStreak: inspect.State.Healthcheck.FailingStreak,
+	}
+
+	log := inspect.State.Healthcheck.Log
+
+	for _, item := range log {
+		res := &types.HealthcheckResult{}
+		s, _ := time.Parse(time.RFC3339Nano, item.Start)
+		e, _ := time.Parse(time.RFC3339Nano, item.End)
+		res.Start = s
+		res.End = e
+		res.ExitCode = item.ExitCode
+		res.Output = item.Output
+		state.Health.Log = append(state.Health.Log, res)
+	}
+
 	formatCapabilities(inspect.HostConfig.CapDrop)
 	formatCapabilities(inspect.HostConfig.CapAdd)
 
--- a/test/apiv2/python/rest_api/test_v2_0_0_container.py
+++ b/test/apiv2/python/rest_api/test_v2_0_0_container.py
@@ -14,6 +14,13 @@ class ContainerTestCase(APITestCase):
         obj = r.json()
         self.assertEqual(len(obj), 1)
 
+    def test_list_filters(self):
+        r = requests.get(self.podman_url + "/v1.40/containers/json?filters%3D%7B%22status%22%3A%5B%22running%22%5D%7D")
+        self.assertEqual(r.status_code, 200, r.text)
+        payload = r.json()
+        containerAmnt = len(payload)
+        self.assertGreater(containerAmnt, 0)
+
     def test_list_all(self):
         r = requests.get(self.uri("/containers/json?all=true"))
         self.assertEqual(r.status_code, 200, r.text)
@@ -24,6 +31,22 @@ class ContainerTestCase(APITestCase):
         self.assertEqual(r.status_code, 200, r.text)
         self.assertId(r.content)
         _ = parse(r.json()["Created"])
+
+        r = requests.post(
+            self.podman_url + "/v1.40/containers/create?name=topcontainer",
+            json={"Cmd": ["top"], "Image": "alpine:latest"},
+        )
+        self.assertEqual(r.status_code, 201, r.text)
+        payload = r.json()
+        container_id = payload["Id"]
+        self.assertIsNotNone(container_id)
+
+        r = requests.get(self.podman_url + f"/v1.40/containers/{payload['Id']}/json")
+        self.assertEqual(r.status_code, 200, r.text)
+        self.assertId(r.content)
+        out = r.json()
+        state = out["State"]["Health"]
+        self.assertIsInstance(state, dict)
 
     def test_stats(self):
         r = requests.get(self.uri(self.resolve_container("/containers/{}/stats?stream=false")))


pkg/api/handlers/compat/containers.go
```diff
@@ -95,9 +95,10 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 	decoder := r.Context().Value("decoder").(*schema.Decoder)
 	query := struct {
-		All   bool `schema:"all"`
-		Limit int  `schema:"limit"`
-		Size  bool `schema:"size"`
+		All     bool   `schema:"all"`
+		Limit   int    `schema:"limit"`
+		Size    bool   `schema:"size"`
+		Filters string `schema:"filters"`
```
Luap99, line 101: Filters should already work. They are handled differently, check below for `util.PrepareFilters(r)`

test/apiv2/python/rest_api/test_v2_0_0_container.py
```diff
@@ -14,6 +14,10 @@ def test_list(self):
         obj = r.json()
         self.assertEqual(len(obj), 1)
 
+    def test_list_filters(self):
+        r = requests.get(self.podman_url + "/v1.40/containers/json?filters={'status':'running'}")
```
mheon, line 18: Any chance we can check the output of the query, see that there are (or are not?) running containers?

pkg/api/handlers/compat/containers.go
cdoern: good point, I will look into why it isn't picking up on status then. was thinking it was a simple missing query param, looks like it might be deeper or just an issue a few lines later in the compat code.

mheon:
Does this actually fix #10457 - I don't see any HC-related changes, just filter ones?

cdoern:
@mheon I am pretty sure the issue is originating with the lack of handling for both status and health filters in the compat code upon inspect. I am planning on taking another look in the morning specifically at the Health filter to see if that is also just not handled.

Luap99:
@cdoern I just noticed, could please include the PR description in the commit message. Also the `fixes #123` should be in the commit message. This makes it easier to track the commits in the local git history and the sign-off-by line should include your real name not the nick.

pkg/api/handlers/compat/containers.go
```diff
@@ -126,6 +126,12 @@ func ListContainers(w http.ResponseWriter, r *http.Request) {
 	// the all setting and always deal with all containers.
 	if len((*filterMap)["status"]) > 0 {
 		all = true
+		stat, err := filters.GenerateContainerFilterFuncs("status", (*filterMap)["status"], runtime) // pass actual value that the user wants
+		if err != nil {
+			utils.InternalServerError(w, err)
+			return
+		}
+		filterFuncs = append(filterFuncs, stat)
```
Luap99, line 134: I don't think this is necessary, the filter functions should have already been generated above.

Luap99 (commented):
You say this fixes #10457 but I don't see why this would fix it. There is no test for #10457

pkg/api/handlers/compat/containers.go
cdoern: correct me if I am wrong, but I interpreted this part of the code as only passing status to the generateFilters function is all is also true which is not the expected behavior? or is this taken care of further up in the code?

cdoern:
> @cdoern I just noticed, could please include the PR description in the commit message. Also the `fixes #123` should be in the commit message. This makes it easier to track the commits in the local git history and the sign-off-by line should include your real name not the nick.

will do, sorry for the formatting issues

cdoern:
> You say this fixes #10457 but I don't see why this would fix it. There is no test for #10457

Added further fixing for health check specifically in most recent commit.

pkg/api/handlers/compat/containers.go
Luap99: When filters are set they are already parsed before, see line 114-123. So we should not parse them again.

pkg/api/handlers/compat/containers.go
```diff
@@ -403,6 +403,32 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
...
+	state.Health.FailingStreak = inspect.State.Healthcheck.FailingStreak
+
+	//item := define.HealthCheckLog{}
+
+	var item (*types.HealthcheckResult)
+
+	log := inspect.State.Healthcheck.Log
+
+	for ind := range log {
+		s, err := time.Parse("2019-12-22T10:59:05.6385933Z", log[ind].Start)
```
mheon, line 417: I'm ~90% sure this is RFC3339Nano which should already be defined in the Go time package

pkg/api/handlers/compat/containers.go
```diff
@@ -403,6 +403,32 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
...
+
+	state.Health.FailingStreak = inspect.State.Healthcheck.FailingStreak
+
+	//item := define.HealthCheckLog{}
+
+	var item (*types.HealthcheckResult)
+
+	log := inspect.State.Healthcheck.Log
+
+	for ind := range log {
```
mheon, line 416: Generally we don't bother with index variables in loops - `for _, item := range log {` will get you the actual healthcheck structure as `item`, no need to index into the array every time

cdoern: Thanks, I'll change to that. I was just doing this loop to convert between `string `and `time.Time` since simply assigning one Log entry to the other wouldn't work because of the incompatible libpod/docker types.

pkg/api/handlers/compat/containers.go
cdoern: yep, you're right. Any idea why on APIv2 test these lines parsing the time (I think) seem to be causing an issue? The test spits out: 
` not ok 726 [44-mounts] GET containers/7d2916da153cfbf7866eb530a68f9d499f4896810a068410e5674ae8263dd3cb/json : status
         #  expected: 200
         #    actual: 500
         #  response: {"cause":"runtime error: invalid memory address or nil pointer dereference","message":"runtime error: invalid memory address or nil pointer dereference","response":500}`

mheon: I think you need a line number - is the server spitting anything out during this? I'd try and run the test locally and verify output of the server.

TomSweeneyRedHat:
@edsantiago can you take a peak at the failing test here?  It's failing on a bad release version and I don't see any changes here that should have triggered that.

test/apiv2/python/rest_api/test_v2_0_0_container.py
```diff
@@ -25,6 +32,22 @@ def test_inspect(self):
...
+        r = requests.post(
+            self.podman_url + "/v1.40/containers/create?name=topcontainer",
+            json={"Cmd": ["top"], "Image": "alpine:latest"},
+        )
+        self.assertEqual(r.status_code, 201, r.text)
+        payload = r.json()
+        container_id = payload["Id"]
+        self.assertIsNotNone(container_id)
+
+        r = requests.get(self.podman_url + f"/v1.40/containers/{payload['Id']}/json")
```
TomSweeneyRedHat, line 44: would it be worthwhile changing "v1.40" to a variable?  Seems like you have it in at least a few places.

edsantiago:
This test should never run. I have to assume that someone manually pressed the 'Trigger' button. Whoever did that, live and learn, don't do it again.

You can ignore the test result. It does not gate merging.

cdoern:
> This test should never run. I have to assume that someone manually pressed the 'Trigger' button. Whoever did that, live and learn, don't do it again.
> 
> You can ignore the test result. It does not gate merging.

that was me sorry, one test failed due to a flake and cancelled all of the others so I must've accidentally restarted that one.

edsantiago:
No worries! To restart a failed test, just press the 'Re-run' link in the left-hand panel. You do not need to restart any others: once the flaked test succeeds, the others will eventually run even though the @#$% UI says "Canceled". It's completely anti-intuitive, just one of those things we have to get used to.

test/apiv2/python/rest_api/test_v2_0_0_container.py
cdoern: @TomSweeneyRedHat seems like a good idea to me, should probably run it by @jwhonce first. Would probably have to edit api_testcase.py and just add it to the end of the podman_url variable in there.

TomSweeneyRedHat:
LGTM

rhatdan:
@mheon @Luap99 PTAL

pkg/api/handlers/compat/containers.go
```diff
@@ -403,6 +403,26 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
 		state.Status = define.ContainerStateCreated.String()
 	}
 
+	state.Health = &types.Health{}
+	state.Health.Status = inspect.State.Healthcheck.Status
+	state.Health.FailingStreak = inspect.State.Healthcheck.FailingStreak
+
+	//item := define.HealthCheckLog{}
```
mheon, line 410: Can you trim this bit?

pkg/api/handlers/compat/containers.go
```diff
@@ -403,6 +403,26 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
...
+
+	var res *types.HealthcheckResult = &types.HealthcheckResult{}
```
mheon, line 412: This looks a little dubious - why is it being created only once? I think every time the for loop below runs, you're overwriting the contents, and appending the same entry to the array again?

I would move this inside the loop for safety

pkg/api/handlers/compat/containers.go
```diff
@@ -403,6 +403,24 @@ func LibpodToContainerJSON(l *libpod.Container, sz bool) (*types.ContainerJSON,
...
 
+	state.Health = &types.Health{
+		Status:        inspect.State.Healthcheck.Status,
+		FailingStreak: inspect.State.Healthcheck.FailingStreak,
+	}
+
+	log := inspect.State.Healthcheck.Log
+
+	for _, item := range log {
+		var res *types.HealthcheckResult = &types.HealthcheckResult{}
```
rhatdan, line 414: Can this be stated as

res:=&types.HealthcheckResult{}

cdoern: yep, pushing that change now.

PR #10618, opened 2021-06-09
System tests: deal with crun 0.20.1

edsantiago:
crun 0.20.1 changed an error message that we relied on. Deal
with it by accepting the old and new message.

Also (unrelated): sneak in some doc fixes to get rid of
nasty go-md2man warnings that have crept into man pages.


edsantiago:
Fedora gating tests are going to fail without this.

edsantiago:

PR #10548, opened 2021-06-03
API images/create added missing parameters platform, message, repo 

cdoern:
Signed-off-by: cdoern <cbdoer23@g.holycross.edu>

Added support for the query parameters platform, message, and repo. Also edited documentation and formatted the request string as binary in register_images.go. CreateImageFromSrc within pkg/api/handlers/compat/images.go has some issues with supporting platform due to imageEngine.Import/the ImageImportOptions struct not supporting a platform parameter. 

I will most likely open a separate issue to handle this as it deals more with our side rather than the docker API. You can see [here](https://github.com/docker/engine/blob/7c6a9484ee37b28fa314f716581bdb268cd78ce2/api/server/router/image/image_routes.go#L26) that Docker supports platform in both the fromSrc and fromImage requests.

PR #10622, opened 2021-06-09
compat import imageFromSrc support for platform query parameter

### LINKED ISSUE #10566: Podman does not support importing image from source with a platform specified

/kind feature

**Description**

While implementing the API endpoint for [/images/create](https://github.com/containers/podman/pull/10548) I noticed that CreateImageFromSrc within pkg/api/handlers/compat/images.go has some issues with supporting platform due to imageEngine.Import/the ImageImportOptions struct not supporting a platform parameter. Suggesting that we modify the ImageImportOptions struct as well as the abi image.go code to support handling the platform as [docker does ](https://github.com/docker/engine/blob/7c6a9484ee37b28fa314f716581bdb268cd78ce2/api/server/router/image/image_routes.go#L26). I can then go back and open up another PR with this fix as well as the API endpoint for supporting platform parsing.

I will most likely need to open up an Issue/PR on containers/common as well due to the fact that the import functions in podman/pkg/domain/infra/abi use libimage migration and end up calling the import function in that directory.

```
podman version 3.1.2
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide? (https://github.com/containers/podman/blob/master/troubleshooting.md)**

Yes

cdoern:

Building off of [this PR](https://github.com/containers/common/pull/606) as well as [this](https://github.com/containers/podman/pull/10548) initial addition, made necessary changes in the entities ImageImportOptions struct as well as the compat image.go code in order to pass an OS and Architecture while importing an image from source as [docker does](https://github.com/docker/engine/blob/7c6a9484ee37b28fa314f716581bdb268cd78ce2/api/server/router/image/image_routes.go#L88). Also edited the python tests to pass an os/arch pair in the query.

fixes #10566 

P.S. sorry for the duplicate pr, computer has been having issues today and all of my git repos got scrambled so I had to nuke a lot and start over

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -166,10 +166,11 @@ func CreateImageFromSrc(w http.ResponseWriter, r *http.Request) {
 	runtime := r.Context().Value("runtime").(*libpod.Runtime)
 
 	query := struct {
-		FromSrc string   `schema:"fromSrc"`
-		Changes []string `schema:"changes"`
-		Message string   `schema:"message"`
-		Repo    string   `shchema:"repo"`
+		Changes  []string `schema:"changes"`
+		FromSrc  string   `schema:"fromSrc"`
+		Message  string   `schema:"message"`
+		Platform string   `schema:"platform"`
+		Repo     string   `shchema:"repo"`
 	}{
 		// This is where you can override the golang default value for one of fields
 	}
@@ -192,9 +193,21 @@ func CreateImageFromSrc(w http.ResponseWriter, r *http.Request) {
 			utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to write temporary file"))
 		}
 	}
+
+	platformSpecs := strings.Split(query.Platform, "/")
+	opts := entities.ImageImportOptions{
+		Source:    source,
+		Changes:   query.Changes,
+		Message:   query.Message,
+		Reference: query.Repo,
+		OS:        platformSpecs[0],
+	}
+	if len(platformSpecs) > 1 {
+		opts.Architecture = platformSpecs[1]
+	}
+
 	imageEngine := abi.ImageEngine{Libpod: runtime}
-	// TODO: add support for ImageImportOptions to take a platform parameter. Also import https://github.com/opencontainers/image-spec/tree/master/specs-go/v1 either here or within imageEngine.Import to get default platform
-	report, err := imageEngine.Import(r.Context(), entities.ImageImportOptions{Source: source, Changes: query.Changes, Message: query.Message, Reference: query.Repo})
+	report, err := imageEngine.Import(r.Context(), opts)
 	if err != nil {
 		utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "unable to import tarball"))
 		return
--- a/pkg/domain/entities/images.go
+++ b/pkg/domain/entities/images.go
@@ -271,8 +271,10 @@ type ImageLoadReport struct {
 }
 
 type ImageImportOptions struct {
+	Architecture    string
 	Changes         []string
 	Message         string
+	OS              string
 	Quiet           bool
 	Reference       string
 	SignaturePolicy string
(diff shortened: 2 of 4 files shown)


pkg/api/handlers/compat/images.go
```diff
@@ -179,6 +180,18 @@ func CreateImageFromSrc(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	// fromSrc – Source to import. The value may be a URL from which the image can be retrieved or - to read the image from the request body. This parameter may only be used when importing an image.
+
+	platformSpecs := strings.Split(query.Platform, "/")
+
+	addOS := true
```
rhatdan, line 186: Does this work.
addOS := len(platformSpecs) == 0 
addArch := len(platformSpecs) > 1

rhatdan: I don't think you need addOS and you should just move this code down below where you are setting the OS and Arch.

pkg/api/handlers/compat/images.go
```diff
@@ -192,9 +205,22 @@ func CreateImageFromSrc(w http.ResponseWriter, r *http.Request) {
 			utils.Error(w, "Something went wrong.", http.StatusInternalServerError, errors.Wrap(err, "failed to write temporary file"))
 		}
 	}
+
```
rhatdan, line 208:
```
 	opts := entities.ImageImportOptions{}
 	platformSpecs := strings.Split(query.Platform, "/")
	opts.OS = platformSpecs[0]
        if len(platformSpecs) > 1 {
    		opts.Architecture = platformSpecs[1]
        }
```

cdoern: right, this makes sense. Will change now

rhatdan:
/approve
LGTM

cdoern:
Thought I'd try :)

pkg/api/handlers/compat/images.go
```diff
@@ -166,10 +166,11 @@ func CreateImageFromSrc(w http.ResponseWriter, r *http.Request) {
...
 	query := struct {
-		FromSrc string   `schema:"fromSrc"`
-		Changes []string `schema:"changes"`
-		Message string   `schema:"message"`
-		Repo    string   `shchema:"repo"`
+		FromSrc  string   `schema:"fromSrc"`
+		Changes  []string `schema:"changes"`
+		Message  string   `schema:"message"`
+		Repo     string   `shchema:"repo"`
+		Platform string   `schema:"platform"`
```
TomSweeneyRedHat, line 173: Put these in alpha order please

pkg/domain/entities/images.go
```diff
@@ -278,6 +278,8 @@ type ImageImportOptions struct {
 	SignaturePolicy string
 	Source          string
 	SourceIsURL     bool
+	OS              string
+	Architecture    string
```
TomSweeneyRedHat, line 282: ditto alpha

test/apiv2/python/rest_api/test_v2_0_0_image.py
```diff
@@ -89,14 +89,9 @@ def test_pull(self):
 
     def test_create(self):
         r = requests.post(
-            self.podman_url + "/v1.40/images/create?fromImage=alpine&platform=linux/amd64/v8",
-            timeout=15,
-        )
+            self.podman_url + "/v1.40/images/create?fromImage=alpine&platform=linux/amd64/v8", timeout=15)
```
TomSweeneyRedHat, line 92: made similar comment in another PR, is it worthwhile to move the "v1.40" to a variable?

TomSweeneyRedHat:
@cdoern you have to go to the dark side before you get the power to `/lgtm` your own PRs!

test/apiv2/python/rest_api/test_v2_0_0_image.py
cdoern: Looking at it again, we could do it but then we would eventually have to edit both this file and api_testcase.py. I guess we could create a local variable in this file?

jwhonce: I would prefer the version number be expicate for these compat tests.

PR #10629, opened 2021-06-10
[CI:DOCS] podman-run.1.md:detach-keys: spell the default value just once

dilyanpalauzov:

### DIFF
--- a/docs/source/markdown/podman-run.1.md
+++ b/docs/source/markdown/podman-run.1.md
@@ -281,11 +281,9 @@ it in the **containers.conf** file: see **containers.conf(5)** for more informat
 
 #### **--detach-keys**=*sequence*
 
-Specify the key sequence for detaching a container. Format is a single character `[a-Z]` or one or more `ctrl-<value>` characters where `<value>` is one of: `a-z`, `@`, `^`, `[`, `,` or `_`. Specifying "" will disable this feature. The default is *ctrl-p,ctrl-q*.
+Specify the key sequence for detaching a container. Format is a single character `[a-Z]` or one or more `ctrl-<value>` characters where `<value>` is one of: `a-z`, `@`, `^`, `[`, `,` or `_`. Specifying "" will set the sequence to the default value of *ctrl-p,ctrl-q*.
 
 This option can also be set in **containers.conf**(5) file.
-
-Specifying "" will disable this feature. The default is **ctrl-p,ctrl-q**.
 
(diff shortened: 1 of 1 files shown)


rhatdan:
Where is it specified elsewhere?

BTW You have to sign your commits.

git commit -a --amend -s
git push --force

dilyanpalauzov:
The short description of **--detach-keys** spells the default value twice.  The second mention is three lines after the first one.

rhatdan:
Ok, But I find it harder to find now.  I think we should reorganize the --detach-keys definition up a few lines, to include the defaults under the option.  But this is not a hard call on my part.  @TomSweeneyRedHat WDYT?

docs/source/markdown/podman-run.1.md
```diff
@@ -281,7 +281,7 @@ it in the **containers.conf** file: see **containers.conf(5)** for more informat
 
 #### **--detach-keys**=*sequence*
 
-Specify the key sequence for detaching a container. Format is a single character `[a-Z]` or one or more `ctrl-<value>` characters where `<value>` is one of: `a-z`, `@`, `^`, `[`, `,` or `_`. Specifying "" will disable this feature. The default is *ctrl-p,ctrl-q*.
+Specify the key sequence for detaching a container. Format is a single character `[a-Z]` or one or more `ctrl-<value>` characters where `<value>` is one of: `a-z`, `@`, `^`, `[`, `,` or `_`. Specifying "" will disable this feature.
```
TomSweeneyRedHat, line 284:
```suggestion
Specify the key sequence for detaching a container. Format is a single character `[a-Z]` or one or more `ctrl-<value>` characters where `<value>` is one of: `a-z`, `@`, `^`, `[`, `,` or `_`. Specifying "" will set the sequence to the default value of *ctrl-p,ctrl-q*
```

TomSweeneyRedHat:
I see `<ctrl-p>, <ctrl-q>` only once in `--detach` and once in `--detach-keys`.  I think we should have one reference for each option.  I think many people if it wasn't in both would miss it.  But I did leave a suggestion for reworking it that might help satisfy your concern.

docs/source/markdown/podman-run.1.md
dilyanpalauzov: The proposed wording `Specifying "" will set the sequence to the default value of *ctrl-p,ctrl-q*` does change the current meaning.  I have updated the patch with the new meaning.

TomSweeneyRedHat:
LGTM
TYVM @dilyanpalauzov !

PR #10631, opened 2021-06-10
[v3.2] rootless: fix fast join userns path

giuseppe:
commit ab886328357184cd0a8375a5dedf816ba91789f9 changed the path for
the pause.pid file but didn't update the same path in the C code.
This prevented Podman to take the fast path when the userns is already
created and to join it without re-execing itself.

Fix the path in the C code as well so we can join the rootless
user+mount namespace without having to re-exec Podman.

[NO TESTS NEEDED]

(cherry picked from commit 11badab046d32c0dce40b28707ee3bb72678e36e)


mheon:
LGTM

vrothberg (commented):
/lgtm

PR #10632, opened 2021-06-10
[v3.0.1-rhel] rootless: fix fast join userns path

giuseppe:
commit ab886328357184cd0a8375a5dedf816ba91789f9 changed the path for
the pause.pid file but didn't update the same path in the C code.
This prevented Podman to take the fast path when the userns is already
created and to join it without re-execing itself.

Fix the path in the C code as well so we can join the rootless
user+mount namespace without having to re-exec Podman.

[NO TESTS NEEDED]

(cherry picked from commit 11badab046d32c0dce40b28707ee3bb72678e36e)

### DIFF
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -302,7 +302,7 @@ static void __attribute__((constructor)) init()
       uid_t uid;
       gid_t gid;
       char path[PATH_MAX];
-      const char *const suffix = "/libpod/pause.pid";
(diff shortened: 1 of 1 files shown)


mheon:
3.0.1 is our LTS branch, zstream only at this point - next RHEL release will be of the 3.2 branch. Did we want to merge this anyways? Will need a BZ to justify the merge if so.

giuseppe:
ok this is not a security issue.  We can close the PR.  Thanks!

PR #10637, opened 2021-06-10
[3.0.1-rhel] handle corrupted images

vrothberg:
While various execution paths in Podman already handle corrupted
images, `podman-{create,image exists,run}` did not.

Some corruptions can only be detected when accessing the individual
data.  A reliable way of accessing such data is accessing its layers.
Hence, an image will only be listed to exist if a) it has been found
and b) can be inspected.  If the inspection fails, the image will be
reported to not exists but without an error; the error will only be
logged.  This allows for properly recovering and pull the image, even
in `podman-{create,run}`.

Podman will now behave as follows:
```
$ ./bin/podman run --rm nginx echo "it works!"
ERRO[0000] Image nginx exists in local storage but may be corrupted: layer not known
Resolved "nginx" as an alias (/home/vrothberg/.cache/containers/short-name-aliases.conf)
Trying to pull docker.io/library/nginx:latest...
Copying blob 351ad75a6cfa skipped: already exists
Copying blob febe5bd23e98 skipped: already exists
Copying blob 30afc0b18f67 skipped: already exists
Copying blob 596b1d696923 skipped: already exists
Copying blob 8283eee92e2f skipped: already exists
Copying blob 69692152171a done
Copying config d1a364dc54 done
Storing signatures
it works!
```

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1966872

### DIFF
--- a/test/system/335-corrupt-layers.bats
+++ b/test/system/335-corrupt-layers.bats
@@ -1,0 +1,90 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# All tests in here perform nasty manipulations on image storage.
+#
+
+load helpers
+
+###############################################################################
+# BEGIN setup/teardown
+
+# Create a scratch directory; this is what we'll use for image store and cache
+if [ -z "${PODMAN_CORRUPT_TEST_WORKDIR}" ]; then
+    export PODMAN_CORRUPT_TEST_WORKDIR=$(mktemp -d --tmpdir=${BATS_TMPDIR:-${TMPDIR:-/tmp}} podman_corrupt_test.XXXXXX)
+fi
+
+# All tests in this file (and ONLY in this file) run with a custom rootdir
+function setup() {
+    skip_if_remote "none of these tests run under podman-remote"
+    _PODMAN_TEST_OPTS="--storage-driver=vfs --root ${PODMAN_CORRUPT_TEST_WORKDIR}/root"
+}
+
+function teardown() {
+    # No other tests should ever run with this custom rootdir
+    unset _PODMAN_TEST_OPTS
+
+    is_remote && return
+
+    # Clean up
+    umount ${PODMAN_CORRUPT_TEST_WORKDIR}/root/overlay || true
+    if is_rootless; then
+        run_podman unshare rm -rf ${PODMAN_CORRUPT_TEST_WORKDIR}/root
+    else
+        rm -rf ${PODMAN_CORRUPT_TEST_WORKDIR}/root
+    fi
+}
+
+# END   setup/teardown
+###############################################################################
+# BEGIN primary test helper
+
+# This is our main action, invoked by every actual test. It:
+#    - creates a new empty rootdir
+#    - populates it with our crafted test image
+#    - removes [ manifest, blob ]
+#    - confirms that "podman images" throws an error
+#    - runs the specified command (rmi -a -f, prune, reset, etc)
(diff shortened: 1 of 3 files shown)


vrothberg:
/hold
Still want to add a test.

vrothberg:
@edsantiago PTAL at the system test.

mheon:
Hold this one until we clarify if this needs a 8.4.0 zstream

vrothberg:
I think we're ready to go, are we?

test/system/335-corrupt-layers.bats
```diff
@@ -0,0 +1,88 @@
...
+#    - runs the specified command (rmi -a -f, prune, reset, etc)
+#    - confirms that it succeeds, and also emits expected warnings
+function _corrupt_image_test() {
+    run_podman info --format "{{.Store.GraphRoot}}"
+    indexPath=$output/*-layers/layers.json
+
+    run_podman pull $IMAGE
+    run_podman image exists $IMAGE
+
+    run jq 'del(.[0])' $indexPath > $indexPath
```
edsantiago, line 55: /hold

I don't think this does what you think it does. I need time to look into it.

vrothberg: Good catch.  I see it now.  Writing back like this will empty the entire file.  It yields effectively the same error but is indeed not what I inteded.

vrothberg: @edsantiago how about:
``bash
run jq 'del(.[0])' $indexPath > $indexPath.tmp
run mv $indexPath.tmp $indexPath
```

Comment (edsantiago) [reply test/system/335-corrupt-layers.bats:55]: Closer, but still won't work: `run` is evil, it's a super-misdirecting problem with BATS. You should almost never use `run`. So, yes, do the above but remove all `run`.

Comment (edsantiago) [reply test/system/335-corrupt-layers.bats:55]: ...except that the `jq` is malformed, so that will fail too.

Comment (edsantiago) [reply test/system/335-corrupt-layers.bats:55]:
```diff
diff --git a/test/system/335-corrupt-layers.bats b/test/system/335-corrupt-layers.bats
index 54eee9992..3bd67de8d 100644
--- a/test/system/335-corrupt-layers.bats
+++ b/test/system/335-corrupt-layers.bats
@@ -52,7 +52,9 @@ function _corrupt_image_test() {
     run_podman pull $IMAGE
     run_podman image exists $IMAGE

-    run jq 'del(.[0])' $indexPath > $indexPath
+    indexTmp="$(realpath $indexPath).tmp.$$"
+    jq 'del(.[0])' $indexPath > $indexTmp
+    mv -f $indexTmp $indexPath
     # A corrupted image won't be marked as existing.
     run_podman 1 image exists $IMAGE
```

vrothberg: Done.  As always, thanks for you guidance, @edsantiago!

edsantiago:
Thanks @vrothberg . Two things still bother me: `podman pull` (which is going to cause flakes), and the fact that `layers.json` is empty so we don't test the one-good-image, one-bad-one case. But I can't think of an easy way to fix either one right now, so we can defer that.

/hold cancel

edsantiago:
LGTM but I'm placing a hold again due to @mheon's earlier z-stream comment. Release if/when ready.

/hold

TomSweeneyRedHat:
LGTM
and I think the hold should be canceled, but I'll leave that to @mheon

PR #10639, opened 2021-06-10
Fix pre-checkpointing

adrianreber:
Unfortunately --pre-checkpointing never worked as intended and recent
changes to runc have shown that it is broken.

To create a pre-checkpoint CRIU expects the paths between the
pre-checkpoints to be a relative path. If having a previous checkpoint
it needs the be referenced like this: --prev-images-dir ../parent

Unfortunately Podman was giving runc (and CRIU) an absolute path.

Unfortunately, again, until March 2021 CRIU silently ignored if
the path was not relative and switch back to normal checkpointing.

This has been now fixed in CRIU and runc and running pre-checkpoint
with the latest runc fails, because runc already sees that the path is
absolute and returns an error.

This commit fixes this by giving runc a relative path.

This commit also fixes a second pre-checkpointing error which was just
recently introduced.

So summarizing: pre-checkpointing never worked correctly because CRIU
ignored wrong parameters and recent changes broke it even more.

Now both errors should be fixed.

### DIFF
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -42,6 +42,7 @@ const (
 	// name of the directory holding the artifacts
 	artifactsDir      = "artifacts"
 	execDirPermission = 0755
+	preCheckpointDir  = "pre-checkpoint"
 )
 
 // rootFsSize gets the size of the container's root filesystem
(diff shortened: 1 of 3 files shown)


libpod/container_internal.go
```diff
@@ -42,6 +42,8 @@ const (
 	// name of the directory holding the artifacts
 	artifactsDir      = "artifacts"
 	execDirPermission = 0755
+	preCheckpointDir  = "pre-checkpoint"
+
```
giuseppe, line 46: extra line here

adrianreber: gone

mheon:
LGTM

giuseppe (approved):
LGTM

PR #10646, opened 2021-06-10
System tests: the continuing multiarch saga

edsantiago:
TL;DR podman needs "arm64" as arch, not "arm64v8".

Unexpurgated version: docker.io publishes ${ARCH}/alpine for
several values of ARCH. Unfortunately, the arm64 one is
called "arm64v8", which is sensible, but podman needs the
--arch value of the manifest to be exactly "arm64". So we
need to special-case this value in our loop. Do so, and
build/publish a new 20210610 testimage. Use that in tests
moving forward.

And, since we need to jump through the same hoops to build
the nonlocal image, include it in the build loop instead
of as a tacked-on comment. Try to be helpful by determining
the next-available numeric tag.

And: don't push anything by default. Instead, just tell
the user what buildah-push commands to run.

And: refactor $PODMAN_NONLOCAL_IMAGE_TAG, to make it easier
for the RHEL-arch-testing folx to override using envariables
instead of inplace-sed. (Not that they should ever need to
override again, because this is the final multiarch commit
that should be forevermore perfect and need no further commits
ever again).

And, finally, bump up to latest alpine/busybox images.


rhatdan:
LGTM

vrothberg (commented):
/lgtm

PR #10654, opened 2021-06-11
Fix network connect race with docker-compose

Luap99:
Network connect/disconnect has to call the cni plugins when the network
namespace is already configured. This is the case for `ContainerStateRunning`
and `ContainerStateCreated`. This is important otherwise the network is
not attached to this network namespace and libpod will throw errors like
`network inspection mismatch...` This problem happened when using
`docker-compose up` in attached mode.

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -1090,7 +1090,7 @@ func (c *Container) NetworkDisconnect(nameOrID, netName string, force bool) erro
 	}
 
 	c.newNetworkEvent(events.NetworkDisconnect, netName)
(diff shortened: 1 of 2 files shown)


Luap99:
@mheon PTAL

mheon:
Good catch, LGTM

mheon:
You caught a flake:
`[Fail] Podman search [It] podman search attempts HTTP if registry is in registries.insecure and force secure is false`

Been seeing this one a lot, I suspect we'll have to chase this down next week (CC @edsantiago)

Restarted

PR #10683, opened 2021-06-15
Fix resize race with podman exec -it

### LINKED ISSUE #10560: podman exec --tty resizes only after the process is started
/kind bug

**Description**

`podman exec -it` only resizes the terminal after the exec process is started. This issues exists in both local and remote podman.

1. `podman run -dit --name test alpine top`

2. `podman exec -it test stty size`

`stty: standard input`

two numbers, the real terminal size
`19 194`

Luap99:
When starting a process with `podman exec -it` the terminal is resized
after the process is started. To fix this allow exec start to accept the
terminal height and width as parameter and let it resize right before
the process is started.

Fixes #10560

### DIFF
--- a/pkg/api/server/register_exec.go
+++ b/pkg/api/server/register_exec.go
@@ -269,10 +269,16 @@ func (s *APIServer) registerExecHandlers(r *mux.Router) error {
 	//      properties:
 	//        Detach:
 	//          type: boolean
-	//          description: Detach from the command. Not presently supported.
+	//          description: Detach from the command.
 	//        Tty:
 	//          type: boolean
-	//          description: Allocate a pseudo-TTY. Presently ignored.
+	//          description: Allocate a pseudo-TTY.
+	//        h:
+	//          type: integer
+	//          description: Height of the TTY session in characters. Tty must be set to true to use it.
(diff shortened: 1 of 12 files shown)


pkg/api/server/register_exec.go
```diff
@@ -272,7 +272,13 @@ func (s *APIServer) registerExecHandlers(r *mux.Router) error {
 	//          description: Detach from the command. Not presently supported.
```
mheon, line 272: This is definitely supported, I don't know why the docs say it isn't.

edsantiago:
Uh-oh. My flake catcher detected that the new test [flaked (sys podman ubuntu-2104 root host)](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6429990330826752/html/sys-podman-ubuntu-2104-root-host.log.html):
```
# # podman exec -it mystty stty size
# stty: standard input

# #|     FAIL: stty under podman exec reads the correct dimensions
# #| expected: '41 62'
# #|   actual: 'stty: standard input
'
```
This is really really bad: it means that the test is not reliable, and is going to start flaking in CI.

PR #10684, opened 2021-06-15
add correct slirp ip to /etc/hosts

Luap99:
The container name should have the slirp interface ip set in /etc/hosts
and not the gateway ip. Commit c8dfcce6db0a introduced this regression.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1972073


mheon:
LGTM. Usual flakes are happening.

maybe-sybr:
Will this be backported onto a v3.2.2, @mheon ? I only just noticed this breakage in my rabbitmq containers and may have to versionlock exclude the 3.2.0 and 3.2.1 packages in prod.

mheon:
Yes

PR #10691, opened 2021-06-16
[CI:DOCS] Cirrus: Prevent BZ1965743 workaround pruning

cevich:
A hidden non-obvious corner-case of temporary changes introduced by
https://github.com/containers/podman/pull/10451 could be unintended
pruning of some ubuntu images.  This could be impactful if for some
reason the `.cirrus.yml: use c5521575421149184 for Ubuntu` commit is
reverted beyond 30-days (the disused image-prune interval) and the _old_
images are needed (for an unforeseen reason).

Mitigate this by temporarily including the old images in the timestamp
updating task. This commit may be reverted (and the problem ignored)
if new VM images are built and deployed for all OS's (i.e. replacing
the Fedora/Ubuntu tag split workaround needed for the BZ).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -673,11 +673,15 @@ meta_task:
         image: quay.io/libpod/imgts:$IMAGE_SUFFIX
     env:
         # Space-separated list of images used by this repository state
+        # TODO: Protect commonly tagged ubuntu images from puning in case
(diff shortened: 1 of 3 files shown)


cevich:
Closing...changes incorporated into #10451

cevich:
Changes in 10451 reverted, so we do need this then.

PR #10692, opened 2021-06-16
[CI:DOCS] Fix multi-arch image build clone:failure

cevich:
A suspected recent change in docker (in github-actions Ubuntu
environment) results in a error:

```
cannot clone: Operation not permitted
Error: cannot re-exec process
```

Fix this by using podman to execute the container instead of docker.

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -72,7 +72,10 @@ jobs:
         if: matrix.source == 'stable'
         id: sniff_test
         run: |
-          VERSION_OUTPUT="$(docker run localhost:5000/podman/${{ matrix.source }} \
+          podman pull --tls-verify=false \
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM, love getting Docker out of our CI whenever possible.

cevich:
Fudged-workflow testing of this [shows it fixed the problem](https://github.com/containers/podman/pull/10692/checks?check_run_id=2842531563#step:7:34).  Removing the fudging, and this will be ready to go...

cevich:
...note to self: after this merges, need to manually run and/or check on the workflow to make sure it pushed v3.2.0

TomSweeneyRedHat:
LGTM

cevich:
Just following up, @rhatdan the run with this change [successfully pushed a `v3.2.0` (and `latest`) image](https://github.com/containers/podman/actions/runs/945646120).

PR #10696, opened 2021-06-16
Podman Stats additional features

### LINKED ISSUE #9258: Additional `podman stats` metrics

/kind feature

**Description**

It would be really helpful if additional column(s) could be added to those available when specifying custom `podman stats` output to indicate:

* maximum/high-water-mark memory usage;
* Average CPU usage (since started in in a given period);
* total CPU time;
* average I/O reads/second;
* average I/O writes/second;
* etc.

... of each container.

cdoern:
Added Avg CPU calculation and CPU up time to podman stats. Adding different feature sets in different PRs, CPU first. 

The CPU track array is used to keep track of all CPU usage values to tally up. AvgCPU contains the average CPU percent usage which is calculated using cgroupStats.CPU.Usage.Total. Lastly, the Start and UpTime stats keep track of when the CPU tracking was started and how long the CPU has been active for since that tracking began.

resolves #9258

### DIFF
--- a/cmd/podman/containers/stats.go
+++ b/cmd/podman/containers/stats.go
@@ -146,7 +146,9 @@ func stats(cmd *cobra.Command, args []string) error {
 func outputStats(reports []define.ContainerStats) error {
 	headers := report.Headers(define.ContainerStats{}, map[string]string{
 		"ID":            "ID",
+		"UpTime":        "CPU TIME",
 		"CPUPerc":       "CPU %",
+		"AVGCPU":        "Avg CPU %",
 		"MemUsage":      "MEM USAGE / LIMIT",
 		"MemUsageBytes": "MEM USAGE / LIMIT",
 		"MemPerc":       "MEM %",
@@ -166,7 +168,7 @@ func outputStats(reports []define.ContainerStats) error {
 	if report.IsJSON(statsOptions.Format) {
 		return outputJSON(stats)
 	}
-	format := "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\n"
+	format := "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\t{{.UpTime}}\t{{.AVGCPU}}\n"
 	if len(statsOptions.Format) > 0 {
 		format = report.NormalizeFormat(statsOptions.Format)
 	}
@@ -200,6 +202,14 @@ func (s *containerStats) ID() string {
 
 func (s *containerStats) CPUPerc() string {
 	return floatToPercentString(s.CPU)
+}
+
+func (s *containerStats) AVGCPU() string {
+	return floatToPercentString(s.AvgCPU)
+}
+
+func (s *containerStats) Up() string {
+	return (s.UpTime.String())
 }
 
 func (s *containerStats) MemPerc() string {
@@ -257,7 +267,9 @@ func outputJSON(stats []containerStats) error {
 	type jstat struct {
 		Id         string `json:"id"` // nolint
 		Name       string `json:"name"`
+		CPUTime    string `json:"cpu_time"`
 		CpuPercent string `json:"cpu_percent"` // nolint
+		AverageCPU string `json:"avg_cpu"`
 		MemUsage   string `json:"mem_usage"`
 		MemPerc    string `json:"mem_percent"`
 		NetIO      string `json:"net_io"`
@@ -269,7 +281,9 @@ func outputJSON(stats []containerStats) error {
 		jstats = append(jstats, jstat{
 			Id:         j.ID(),
 			Name:       j.Name,
+			CPUTime:    j.Up(),
 			CpuPercent: j.CPUPerc(),
+			AverageCPU: j.AVGCPU(),
 			MemUsage:   j.MemUsage(),
 			MemPerc:    j.MemPerc(),
 			NetIO:      j.NetIO(),
--- a/libpod/stats.go
+++ b/libpod/stats.go
@@ -56,7 +56,11 @@ func (c *Container) GetContainerStats(previousStats *define.ContainerStats) (*de
 
 	previousCPU := previousStats.CPUNano
 	now := uint64(time.Now().UnixNano())
+	stats.Duration = cgroupStats.CPU.Usage.Total
+	stats.UpTime = time.Duration(stats.Duration)
 	stats.CPU = calculateCPUPercent(cgroupStats, previousCPU, now, previousStats.SystemNano)
+	stats.AvgCPU = calculateAvgCPU(stats.CPU, previousStats.AvgCPU, previousStats.DataPoints)
+	stats.DataPoints = previousStats.DataPoints + 1
 	stats.MemUsage = cgroupStats.Memory.Usage.Usage
 	stats.MemLimit = getMemLimit(cgroupStats.Memory.Usage.Limit)
 	stats.MemPerc = (float64(stats.MemUsage) / float64(stats.MemLimit)) * 100
@@ -126,4 +130,10 @@ func calculateBlockIO(stats *cgroups.Metrics) (read uint64, write uint64) {
 		}
 	}
 	return
+}
+
+// calculateAvgCPU calculates the avg CPU percentage given the previous average and the number of data points.
+func calculateAvgCPU(statsCPU float64, prevAvg float64, prevData int64) float64 {
+	avgPer := ((prevAvg * float64(prevData)) + statsCPU) / (float64(prevData) + 1)
+	return avgPer
 }
--- a/test/e2e/stats_test.go
+++ b/test/e2e/stats_test.go
@@ -83,6 +83,17 @@ var _ = Describe("Podman stats", func() {
 		Expect(session.ExitCode()).To(Equal(0))
 	})
 
+	It("podman stats only output CPU data", func() {
+		session := podmanTest.RunTopContainer("")
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		session = podmanTest.Podman([]string{"stats", "--all", "--no-stream", "--format", "\"{{.ID}} {{.UpTime}} {{.AVGCPU}}\""})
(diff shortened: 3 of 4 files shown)


cmd/podman/containers/stats.go
```diff
@@ -146,7 +146,9 @@ func stats(cmd *cobra.Command, args []string) error {
 func outputStats(reports []define.ContainerStats) error {
 	headers := report.Headers(define.ContainerStats{}, map[string]string{
 		"ID":            "ID",
+		"UpTime":        "CPU TIME (S)",
```
mheon, line 149: Is UpTime really correect? I've always heard that as the total duration a system has been running for, not cumulative CPU time?

cdoern: This is the result of podman stats 
```
ID            NAME            CPU TIME (S)     CPU %   AVG CPU %  MEM USAGE / LIMIT  MEM %   NET IO   BLOCK IO  PIDS
f10ae061aa6b  thirsty_colden  14m1.261643561s  1.28%   6.60%      729.4MB / 16.02GB  4.55%   -- / --  -- / --   117
```
I made it so that the CPU time pulls the container creation time if the variable is time's zero value. I assumed that CPU time just basically meant up time of the container? If I was wrong I can go back and adjust accordingly.

mheon: No, CPU time is the number of seconds or CPU time the container has actively used, not seconds it has been alive.

cdoern: @mheon I _think_ most recent commit fixed this. I didn't realize cgroupStats.CPU.Usage.Total was a duration in ns of how much up time the group had. I switched around some things with avg CPU % given this information as well.

libpod/stats.go
```diff
@@ -56,7 +56,11 @@ func (c *Container) GetContainerStats(previousStats *define.ContainerStats) (*de
 
 	previousCPU := previousStats.CPUNano
 	now := uint64(time.Now().UnixNano())
+	stats.Duration = cgroupStats.CPU.Usage.Total
+	stats.UpTime = time.Duration(stats.Duration)
 	stats.CPU = calculateCPUPercent(cgroupStats, previousCPU, now, previousStats.SystemNano)
+	stats.AvgCPU = calculateAvgCPU(stats.CPU, previousStats.CPUTrack)
+	stats.CPUTrack = append(previousStats.CPUTrack, float64(cgroupStats.CPU.Usage.Total))
```
giuseppe, line 63: any risk that this can grow indefinitely or use outdated info (or should the average computed using all the available data points)?

If we need all data points, could we just store the current average and number of data points to calculate the incremental average?

cdoern: Good idea, i will store number of datapoints and the current average I am pretty sure I can do `((oldAvg * numDataPoints) + newData) / (numDataPoints + 1)`

giuseppe (approved):
LGTM

libpod/stats.go
```diff
@@ -127,3 +131,10 @@ func calculateBlockIO(stats *cgroups.Metrics) (read uint64, write uint64) {
 	}
 	return
 }
+
+// calculateAvgCPU calculates the avg CPU percentage given the previous average and the number of data points.
+func calculateAvgCPU(statsCPU float64, prevAvg float64, prevData int64) float64 {
+	total := prevAvg
```
rhatdan, line 137: Why add a temporary variable?  Why not just use prevAvg?

test/e2e/stats_test.go
```diff
@@ -75,6 +75,15 @@ var _ = Describe("Podman stats", func() {
 	})
 
 	It("podman stats only output cids", func() {
+		session := podmanTest.RunTopContainer("")
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+		session = podmanTest.Podman([]string{"stats", "--all", "--no-stream", "--format", "\"{{.ID}} {{.UpTime}} {{.AVGCPU}}\""})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
rhatdan, line 83: Can you look for the header in the the output?

rhatdan (approved):

cmd/podman/containers/stats.go
```diff
@@ -166,7 +168,7 @@ func outputStats(reports []define.ContainerStats) error {
 	if report.IsJSON(statsOptions.Format) {
 		return outputJSON(stats)
 	}
-	format := "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\n"
+	format := "{{.ID}}\t{{.Name}}\t{{.UpTime}}\t{{.CPUPerc}}\t{{.AVGCPU}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\n"
```
Luap99, line 171: Changing the default format would break docker compat. Not sure if this is desired. @mheon @rhatdan WDYT?

rhatdan: I am not sure I would classify this as a breaking change to compat, but it would probbaly be best if we added the new fields to the end of the output, that way, if someone built a script for docker that expected the second field to be CPUPerc, we would not break it.

cdoern: Do you want me to switch around the ordering? I feel like that would make the output look strange though to have CPU data all scattered throughout.

test/e2e/stats_test.go
cdoern: @rhatdan do you mean something like

 `session = podmanTest.Podman([]string{"stats", "--all", "--no-stream", "--format", "\"{{.ID}} {{.UpTime}} {{.AVGCPU}}\""})`
`out := session.OutputToStringArray()`
`Expect(len(out)).To(BeNumerically("==", 3))`

or would that not work? It seems that none of the stats tests actually validate the output too much.

cmd/podman/containers/stats.go
TomSweeneyRedHat: Unfortunately, I think we need to @cdoern

TomSweeneyRedHat: Maybe add an `--alpha` flag which would then show the stats in alphabetical order left to right?  Or if you really want to go off the rails, make it all selectable by the caller.

test/e2e/stats_test.go
rhatdan: No I just want you to make sure the headings you expect show up in the output.
	Expect(session.LineInOutputContains("UpTime").To(BeTrue())
	Expect(session.LineInOutputContains("AVGCPU").To(BeTrue())

cmd/podman/containers/stats.go
cdoern: @TomSweeneyRedHat working on an alpha flag right now, should it simply just be an if/else around `format:=...` with two different options?

Luap99: @TomSweeneyRedHat I don't see the value for an extra alpha flag, the user can always change the output format with `--format`

mheon: Tend to agree - I don't see a strong case for this until and unless we get a user request to add it.

cmd/podman/containers/stats.go
```diff
@@ -75,6 +76,7 @@ func statFlags(cmd *cobra.Command) {
 
 	flags.BoolVar(&statsOptions.NoReset, "no-reset", false, "Disable resetting the screen between intervals")
 	flags.BoolVar(&statsOptions.NoStream, "no-stream", false, "Disable streaming stats and only pull the first result, default setting is false")
+	flags.BoolVar(&statsOptions.Alpha, "alpha", false, "Decides whether or not to alphabetize the output")
```
rhatdan, line 79: I thought we decided to not add the --alpha flag?

cmd/podman/containers/stats.go
rhatdan: Ok, so we are not going to move the new data to the end of stats?  And just tell users if they want a specific format to use --format.  I guess I am good with this.  Make the default look good for the user.

baude:
lgtm

cmd/podman/containers/stats.go
```diff
@@ -166,7 +168,8 @@ func outputStats(reports []define.ContainerStats) error {
 	if report.IsJSON(statsOptions.Format) {
 		return outputJSON(stats)
 	}
-	format := "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\n"
+	format := ""
+	format = "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\t{{.UpTime}}\t{{.AVGCPU}}\n"
```
jwhonce, line 172:
```suggestion
	format := "{{.ID}}\t{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}\t{{.PIDS}}\t{{.UpTime}}\t{{.AVGCPU}}\n"
```

cmd/podman/containers/stats.go
TomSweeneyRedHat: @Luap99 The value in my eye is just typing `--alpha` rather than working it out in the `--format` option.  YMMV.  I think it would be a good addition, but not a biggy.

srcshelton (commented):
I've patched this on top of `podman-3.2.2`, and I'm always getting exactly the same value for `.CPUPerc` and the new `.AVGCPU` at all times - is this expected to be fully hooked-up at this point, or have I missed something?

(Also, if the manpages/docs aren't auto-generated from the code, could a doc update for these additions please also be added?)

cdoern:
@srcshelton running now and I can't seem to recreate this. Can I see how you're running this/your output?

srcshelton:
> I've patched this on top of `podman-3.2.2`, and I'm always getting exactly the same value for `.CPUPerc` and the new `.AVGCPU` at all times - is this expected to be fully hooked-up at this point (or dependent on another patch on top of `3.2.2` to function), or have I missed something?

I've realised that this is likely because I'm running `podman stats` with the `--no-stream` option in a loop, to integrate other data into the eventual output.

I'd suggest that the '{{.AVGCPU}}' option _must_ gather data persistently beyond the `podman` invocation which outputs the current counters!

PR #10703, opened 2021-06-17
do not merge - CI test for containers/common/pull/630

vrothberg:


vrothberg:
Ah. I think we need to revert bump of c/image in c/common.

Issue #10297, opened 2021-05-10
Support filtering connections from slirp4netns 


KCSesh:
/kind feature

**Description**

Is there any support for filtering connections from slirp4netns? 
https://github.com/rootless-containers/slirp4netns/blob/master/slirp4netns.1.md#filtering-connections

I would like to [allow_host_loopback](http://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net)
But then setup additional iptable rules to **not** allow for total host access. i.e. allow loopback to host on a specific port and set this on a per container basis. 

Is there any way to do this today? 
I imagine a custom prestart hook could likely be written to enter the network namespace and setup iptable rules. 
The only problem is knowing the network namespace path at hook execution time.

mheon:
I don't believe we can do this in the container, and Rootless Podman does not have permission to modify the system firewall, so you may have to make these changes manually. @AkihiroSuda @giuseppe Can you confirm?

KCSesh:
Is there a way I can manually make these changes @mheon from the host using root?

Luap99:
This should work to join the container network namespace:
```
$ podman run --name test -d alpine top
# Note this only works when the container is running or you used podman container init test before
$ nspath=$(podman container inspect --format "{{.NetworkSettings.SandboxKey}}" test)
# This command executes your shell in the user and network namespace
$ podman unshare nsenter --net=$nspath
# Here should be able to use iptables
$ iptables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
```

Note: If you run into `Fatal: can't open lock file /run/xtables.lock: Permission denied` set the  `XTABLES_LOCKFILE` envar to a user writeable location, e.g. `XTABLES_LOCKFILE=/tmp/xtables.lock`.

Also the rules won't persist if the container is stopped and started again.

basilgello:
I am using the `createContainer` OCI hook to add iptables rules.

On simple containers like Debian it works like a charm, but on some containers Podman manages to lauch slirp4netns right before startContainer, but not before createContainer hook is called.

I guess we must ensure that createContainer hook is called after slirp4netns is started.

Also the SandboxKey from the alpine example is non-empty only if no `--userns=keep-id` specified.

rhatdan:
@mheon @Luap99 what is going on with this one?

[...]

PR #3310, opened 2019-06-12
rootless: Rearrange setup of rootless containers ***CIRRUS: TEST IMAGES***

gabibeyer:
In order to run Podman with VM-based runtimes unprivileged, the
network must be set up prior to the container creation. Therefore
this commit modifies Podman to:
  1. create a network namespace
  2. pass the netns persistent mount path to the slirp4netns
     to create the tap inferface
  3. pass the netns path to the OCI spec, so the runtime can
     enter the netns

Closes #2897

Signed-off-by: Gabi Beyer <gabrielle.n.beyer@intel.com>

PR #10704, opened 2021-06-17
Implement 'networksetup' extension stage

basilgello:
This PR addresses the class of issues described in #10297

The `createContainer` OCI hooks are executed after OCI runtime creates container but it is not guaranteed that the network like Slirp4netns will be ready at the time of hook invocation.

To be able to set `iptables(8)` or `nftables(8)` rules based on gateway provided by Slirp4netns, one needs to hook before the root is pivoted but after network is ready.

### DIFF
--- a/docs/source/markdown/podman.1.md
+++ b/docs/source/markdown/podman.1.md
@@ -64,9 +64,14 @@ For the bind-mount conditions, only mounts explicitly requested by the caller vi
 
 If `--hooks-dir` is unset for root callers, Podman and libpod will currently default to `/usr/share/containers/oci/hooks.d` and `/etc/containers/oci/hooks.d` in order of increasing precedence.  Using these defaults is deprecated, and callers should migrate to explicitly setting `--hooks-dir`.
 
-Podman and libpod currently support an additional `precreate` state which is called before the runtime's `create` operation.  Unlike the other stages, which receive the container state on their standard input, `precreate` hooks receive the proposed runtime configuration on their standard input.  They may alter that configuration as they see fit, and write the altered form to their standard output.
+Podman and libpod currently support the following additional states:
+* `precreate`
+These hooks are called before the runtime's `create` operation. Unlike the other stages, which receive the container state on their standard input, `precreate` hooks receive the proposed runtime configuration on their standard input.  They may alter that configuration as they see fit and write the altered form to their standard output.
 
 **WARNING**: the `precreate` hook lets you do powerful things, such as adding additional mounts to the runtime configuration.  That power also makes it easy to break things.  Before reporting libpod errors, try running your container with `precreate` hooks disabled to see if the problem is due to one of your hooks.
+
+* `networksetup`
+These hooks are called after Podman finishes setting the network up. The hook path resolves in the runtime namespace, but the hook executes in the container namespace. These hooks are useful to set network restrictions via `iptables(8)` or `nftables(8)` or perform other tasks inside a container with the network ready but before the root is pivoted.
 
 #### **--identity**=*path*
 
--- a/libpod/container_internal.go
+++ b/libpod/container_internal.go
@@ -1090,10 +1090,23 @@ func (c *Container) runNetworkSetupHooks(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	netnsPath := ""
+	if !c.config.PostConfigureNetNS && c.state.NetNS != nil {
+		netnsPath = c.state.NetNS.Path()
+	}
+	if netnsPath == "" && c.state.PID > 0 {
+		netnsPath = fmt.Sprintf("/proc/%d/ns/net", c.state.PID)
+	}
+	if netnsPath == "" {
+		return fmt.Errorf("container %s: both netns path is empty and container PID is zero", c.ID())
+	}
+
 	for i, hook := range allHooks["networksetup"] {
+		hook := hook
 		logrus.Debugf("container %s: invoke networksetup hook %d, path %s", c.ID(), i, hook.Path)
 		var stderr, stdout bytes.Buffer
-		hookErr, err := exec.Run(ctx, &hook, state, &stdout, &stderr, exec.DefaultPostKillTimeout)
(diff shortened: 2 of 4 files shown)


libpod/container_internal.go
```diff
@@ -1042,6 +1042,91 @@ func (c *Container) cniHosts() string {
 	return hosts
 }
 
+// Run networksetup hooks
+func (c *Container) runNetworkSetupHooks(ctx context.Context) error {
```
mheon, line 1046: Can you refactor this to share code with `setupOCIHooks`? They look largely identical.

libpod/container_internal.go
```diff
@@ -1042,6 +1042,91 @@ func (c *Container) cniHosts() string {
...
+		Status:      "networksetup",
+		Bundle:      c.bundlePath(),
+		Annotations: c.config.Spec.Annotations,
+	})
+	if err != nil {
+		return err
+	}
+
+	netnsPath := ""
+	if !c.config.PostConfigureNetNS && c.state.NetNS != nil {
```
mheon, line 1095: It seems like this hook stage isn't actually guaranteed to run after network init, given that `PostConfigureNetNS` exists. That seems a little questionable - I suppose the limitation could be documented, but it's still counterintuitive.

basilgello: That was the question that I had in mind... To ensure the hook fires after the network init in any case, should I make the PR to OCI spec and rework this PR to pass the hook similar to `createContainer` one?

Usually, what chances are that an OCI specification proposal gets approved in the reasonable timeframe?

basilgello: By "OCI spec" I meant OCI spec + libcontainer + runc + (maybe) crun PRs. Thinking about kata and runsc I do not use...

basilgello: I don't see any example where `PostConfigureNetNS` is set. Can you point the one, please?

mheon: Main use of it is with user namespaces in root containers - we need to configure the network after the user namespace is created.

basilgello: > I don't see any example of Podman invocation where `PostConfigureNetNS` is set. Can you point the one, please?

OK, I see from https://github.com/containers/podman/pull/3310 that the primary use-case for `PostConfigureNetNS` is VM OCI runtimes like Kata. Still, the question: if slirp4netns starts routing packets, how different is behavior with Kata from the behavior with rootless Podman? I want to have a clear understanding of the issue before I can make the limitation documented or overcome it.

docs/source/markdown/podman.1.md
```diff
@@ -64,10 +64,15 @@ For the bind-mount conditions, only mounts explicitly requested by the caller vi
 
 If `--hooks-dir` is unset for root callers, Podman and libpod will currently default to `/usr/share/containers/oci/hooks.d` and `/etc/containers/oci/hooks.d` in order of increasing precedence.  Using these defaults is deprecated, and callers should migrate to explicitly setting `--hooks-dir`.
 
-Podman and libpod currently support an additional `precreate` state which is called before the runtime's `create` operation.  Unlike the other stages, which receive the container state on their standard input, `precreate` hooks receive the proposed runtime configuration on their standard input.  They may alter that configuration as they see fit, and write the altered form to their standard output.
+Podman and libpod currently supports the following additional states:
```
TomSweeneyRedHat, line 67:
```suggestion
Podman and libpod currently support the following additional states:
```

docs/source/markdown/podman.1.md
```diff
@@ -64,10 +64,15 @@ For the bind-mount conditions, only mounts explicitly requested by the caller vi
...
+* `precreate`
+These hooks are called before the runtime's `create` operation. Unlike the other stages, which receive the container state on their standard input, `precreate` hooks receive the proposed runtime configuration on their standard input.  They may alter that configuration as they see fit, and write the altered form to their standard output.
```
TomSweeneyRedHat, line 69:
```suggestion
These hooks are called before the runtime's `create` operation. Unlike the other stages, which receive the container state on their standard input, `precreate` hooks receive the proposed runtime configuration on their standard input.  They may alter that configuration as they see fit and write the altered form to their standard output.
```

docs/source/markdown/podman.1.md
```diff
@@ -64,10 +64,15 @@ For the bind-mount conditions, only mounts explicitly requested by the caller vi
...
 
 **WARNING**: the `precreate` hook lets you do powerful things, such as adding additional mounts to the runtime configuration.  That power also makes it easy to break things.  Before reporting libpod errors, try running your container with `precreate` hooks disabled to see if the problem is due to one of your hooks.
 
+* `networksetup`
+These hooks are called after Podman finishes setting the network up. The hook path resolves in runtime namespace, but the hook executes in the container namespace. These hooks are useful to set network restrictions via `iptables(8)` or `nftables(8)` or perform other tasks inside container with network ready but before the root is pivoted.
```
TomSweeneyRedHat, line 74:
```suggestion
These hooks are called after Podman finishes setting the network up. The hook path resolves in the runtime namespace, but the hook executes in the container namespace. These hooks are useful to set network restrictions via `iptables(8)` or `nftables(8)` or perform other tasks inside a container with the network ready but before the root is pivoted.
```

libpod/container_internal.go
basilgello: Done!

rhatdan:
You need to add tests, or if tests can not be added, add [NO TESTS NEEDED]

rhatdan:
@basilgello Are you still interested in this?

basilgello:
> Are you still interested in this?

Yes, as I mentiined in #10297 I felt a bit stuck first time because on vacation, I coukdnt find where did I answer to @mheon :) But now I found that place

rhatdan:
@basilgello are you still working on this?

rhatdan:
I am going to close this for since @basilgello does not seem to be working on it any longer. Reopen, if you want to make progress.

PR #10709, opened 2021-06-17
Cirrus: Update Ubuntu VM images

cevich:
These images include a fix for:
https://github.com/containers/common/issues/631

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -32,8 +32,8 @@ env:
     # Google-cloud VM Images
     # TODO: At the time of this comment, an selinux-policy regression is blocking use of updated
     # Fedora VM images: https://bugzilla.redhat.com/show_bug.cgi?id=1965743
-    IMAGE_SUFFIX_UBUNTU: "c5521575421149184"
-    IMAGE_SUFFIX: "c5348179051806720"
+    IMAGE_SUFFIX_UBUNTU: "c6534244118822912"   # new
+    IMAGE_SUFFIX: "c6534244118822912"          # old
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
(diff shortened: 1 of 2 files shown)


cevich:
bdflush syscall permission denied problem from buildah PR is [reproducing here also.](https://cirrus-ci.com/task/6152440987779072?logs=main#L272) using freshly built image c4805484248039424.

Trying the image I built 3-days earlier: c4635821094469632

cevich:
`c4635821094469632` [does not reproduce the problem](https://github.com/containers/podman/pull/10709/checks?check_run_id=2850856668)

cevich:
Hrmmmm, it passed, that was unexpected.

cevich:
Ugh, this is really strange.  [Fedora 33/34 think the /dev/kvm device doesn't exist](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4687202643869696/html/int-podman-fedora-33-rootless-host.log.html) and the [Ubuntus throw an error inside the container.](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5250152597291008/html/int-podman-ubuntu-2010-rootless-host.log.html#t--podman-run-device-host-device-with---privileged--1)

cevich:
Okay right...the Fedora problem is explainable - we're using really old images due to need of an SELinux-related updated (IIRC, already working it's way through the system).  Ignoring this for now.

The Ubuntu failures are a true anomaly though.  I did go into a VM and verify the device (on the host) does exist.  @mheon these VM's have just had nested-virt enabled on them (previously it was disabled).  I'm guessing this is why the Ubuntu test failures got missed: The device node wasn't there, now it is (test skips if file is missing).  Is this even a valid test to run in rootless mode?  Maybe should I switch the test to use a less-special device, like `/dev/random` or something?

rhatdan:
I think the SELinux issues should be fixed in Fedora now.

cevich:
> I think the SELinux issues should be fixed in Fedora now.

Oh great, thanks for reminding me.  I'll revert that change.

cevich:
Oof, what did I do :confused:

cevich:
Re-opened as https://github.com/containers/podman/pull/10829

PR #10711, opened 2021-06-17
Fix documentation of the --format option of podman push

mtrmac:
It affects all transports; and without --format, we try several manifest formats.


mheon:
Sure, LGTM

PR #10716, opened 2021-06-17
Podman Pod Create --cpus and --cpuset-cpus flags

cdoern:
Added logic and handling for two new Podman pod create Flags.

--cpus specifies the total number of cores on which the pod can execute, this
is a combination of the period and quota for the CPU.

 --cpuset-cpus is a string value which determines of these available cores,
how many we will truly execute on.

### DIFF
--- a/cmd/podman/pods/create.go
+++ b/cmd/podman/pods/create.go
@@ -5,9 +5,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"runtime"
+	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/containers/common/pkg/completion"
+	"github.com/containers/common/pkg/sysinfo"
 	"github.com/containers/podman/v3/cmd/podman/common"
 	"github.com/containers/podman/v3/cmd/podman/parse"
 	"github.com/containers/podman/v3/cmd/podman/registry"
@@ -16,6 +20,7 @@ import (
 	"github.com/containers/podman/v3/pkg/errorhandling"
 	"github.com/containers/podman/v3/pkg/specgen"
 	"github.com/containers/podman/v3/pkg/util"
+	"github.com/docker/docker/pkg/parsers"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -54,6 +59,14 @@ func init() {
 	flags.SetInterspersed(false)
 
 	common.DefineNetFlags(createCommand)
+
+	cpusetflagName := "cpuset-cpus"
+	flags.StringVar(&createOptions.CpusetCpus, cpusetflagName, "", "CPUs in which to allow execution")
+	_ = createCommand.RegisterFlagCompletionFunc(cpusetflagName, completion.AutocompleteDefault)
+
+	cpusflagName := "cpus"
+	flags.Float64Var(&createOptions.Cpus, cpusflagName, 0.000, "set amount of CPUs for the pod")
+	_ = createCommand.RegisterFlagCompletionFunc(cpusflagName, completion.AutocompleteDefault)
 
 	cgroupParentflagName := "cgroup-parent"
 	flags.StringVar(&createOptions.CGroupParent, cgroupParentflagName, "", "Set parent cgroup for the pod")
@@ -185,6 +198,46 @@ func create(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	numCPU := sysinfo.NumCPU()
+	if numCPU == 0 {
+		numCPU = runtime.NumCPU()
+	}
+	if createOptions.Cpus > float64(numCPU) {
+		createOptions.Cpus = float64(numCPU)
+	}
+	copy := createOptions.CpusetCpus
+	cpuSet := createOptions.Cpus
+	if cpuSet == 0 {
+		cpuSet = float64(sysinfo.NumCPU())
+	}
+	ret, err := parsers.ParseUintList(copy)
+	copy = ""
+	if err != nil {
+		errors.Wrapf(err, "could not parse list")
+	}
+	var vals []int
+	for ind, val := range ret {
+		if val {
+			vals = append(vals, ind)
+		}
+	}
+	sort.Ints(vals)
+	for ind, core := range vals {
+		if core > int(cpuSet) {
+			if copy == "" {
+				copy = "0-" + strconv.Itoa(int(cpuSet))
+				createOptions.CpusetCpus = copy
+				break
+			} else {
+				createOptions.CpusetCpus = copy
+				break
+			}
+		} else if ind != 0 {
+			copy += "," + strconv.Itoa(core)
+		} else {
+			copy = "" + strconv.Itoa(core)
+		}
+	}
 	response, err := registry.ContainerEngine().PodCreate(context.Background(), createOptions)
 	if err != nil {
 		return err
--- a/docs/source/markdown/podman-pod-create.1.md
+++ b/docs/source/markdown/podman-pod-create.1.md
@@ -23,6 +23,22 @@ Add a host to the /etc/hosts file shared between all containers in the pod.
 
 Path to cgroups under which the cgroup for the pod will be created. If the path is not absolute, the path is considered to be relative to the cgroups path of the init process. Cgroups will be created if they do not already exist.
 
+#### **--cpus**=*amount*
+
+Set the total number of CPUs delegated to the pod. Default is 0.000 which indicates that there is no limit on computation power.
+
+#### **--cpuset-cpus**=*amount*
+
+Limit the CPUs to support execution. First CPU is numbered 0. Unlike --cpus this is of type string and parsed as a list of numbers
+
+Format is 0-3,0,1
+
+Examples of the List Format:
+
+0-4,9           # bits 0, 1, 2, 3, 4, and 9 set
+0-2,7,12-14     # bits 0, 1, 2, 7, 12, 13, and 14 set
+
+
 #### **--dns**=*ipaddr*
 
 Set custom DNS servers in the /etc/resolv.conf file that will be shared between all containers in the pod. A special option, "none" is allowed which disables creation of /etc/resolv.conf for the pod.
--- a/libpod/container_validate.go
+++ b/libpod/container_validate.go
@@ -131,6 +131,5 @@ func (c *Container) validate() error {
 	if c.config.User == "" && (c.config.Spec.Process.User.UID != 0 || c.config.Spec.Process.User.GID != 0) {
 		return errors.Wrapf(define.ErrInvalidArg, "please set User explicitly via WithUser() instead of in OCI spec directly")
 	}
-
 	return nil
 }
--- a/libpod/define/pod_inspect.go
+++ b/libpod/define/pod_inspect.go
@@ -51,6 +51,12 @@ type InspectPodData struct {
 	// Containers gives a brief summary of all containers in the pod and
 	// their current status.
 	Containers []InspectPodContainerInfo `json:"Containers,omitempty"`
+	// CPUPeriod contains the CPU period of the pod
+	CPUPeriod uint64 `json:"cpu_period,omitempty"`
+	// CPUQuota contains the CPU quota of the pod
+	CPUQuota int64 `json:"cpu_quota,omitempty"`
+	// CPUSetCPUs contains linux specific CPU data for the pod
+	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
 }
 
 // InspectPodInfraConfig contains the configuration of the pod's infra
@@ -91,6 +97,12 @@ type InspectPodInfraConfig struct {
 	Networks []string
 	// NetworkOptions are additional options for each network
 	NetworkOptions map[string][]string
+	// CPUPeriod contains the CPU period of the pod
+	CPUPeriod uint64 `json:"cpu_period,omitempty"`
+	// CPUQuota contains the CPU quota of the pod
+	CPUQuota int64 `json:"cpu_quota,omitempty"`
+	// CPUSetCPUs contains linux specific CPU data for the container
+	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
 }
 
 // InspectPodContainerInfo contains information on a container in a pod.
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/cri-o/ocicni/pkg/ocicni"
+	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -559,7 +560,6 @@ func WithMaxLogSize(limit int64) CtrCreateOption {
 		if ctr.valid {
 			return define.ErrRuntimeFinalized
 		}
-
 		ctr.config.LogSize = limit
 
 		return nil
@@ -867,7 +867,6 @@ func WithMountNSFrom(nsCtr *Container) CtrCreateOption {
 		if err := checkDependencyContainer(nsCtr, ctr); err != nil {
 			return err
 		}
-
 		ctr.config.MountNsCtr = nsCtr.ID()
 
 		return nil
@@ -2356,6 +2355,45 @@ func WithVolatile() CtrCreateOption {
 		}
 
 		ctr.config.Volatile = true
+		return nil
+	}
+}
+
+// WithPodCPUPAQ takes the given cpu period and quota and inserts them in the proper place.
+func WithPodCPUPAQ(period uint64, quota int64) PodCreateOption {
+	return func(pod *Pod) error {
+		if pod.valid {
+			return define.ErrPodFinalized
+		}
+		if pod.CPUPeriod() != 0 && pod.CPUQuota() != 0 {
+			pod.config.InfraContainer.ResourceLimits.CPU = &specs.LinuxCPU{
+				Period: &period,
+				Quota:  &quota,
+			}
+		} else {
+			pod.config.InfraContainer.ResourceLimits = &specs.LinuxResources{}
+			pod.config.InfraContainer.ResourceLimits.CPU = &specs.LinuxCPU{
+				Period: &period,
+				Quota:  &quota,
+			}
+		}
+		return nil
+	}
+}
+
+// WithPodCPUSetCPUS computes and sets the Cpus linux resource string which determines the amount of cores, from those available,  we are allowed to execute on
+func WithPodCPUSetCPUs(inp string) PodCreateOption {
+	return func(pod *Pod) error {
+		if pod.valid {
+			return define.ErrPodFinalized
+		}
+		if pod.ResourceLim().CPU.Period != nil {
+			pod.config.InfraContainer.ResourceLimits.CPU.Cpus = inp
+		} else {
+			pod.config.InfraContainer.ResourceLimits = &specs.LinuxResources{}
+			pod.config.InfraContainer.ResourceLimits.CPU = &specs.LinuxCPU{}
+			pod.config.InfraContainer.ResourceLimits.CPU.Cpus = inp
+		}
 		return nil
 	}
 }
--- a/libpod/pod.go
+++ b/libpod/pod.go
@@ -1,12 +1,14 @@
 package libpod
 
 import (
+	"context"
 	"net"
 	"time"
 
 	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/libpod/lock"
 	"github.com/cri-o/ocicni/pkg/ocicni"
+	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 )
 
@@ -91,25 +93,26 @@ type podState struct {
 // Generally speaking, aside from those two exceptions, these options will set
 // the equivalent field in the container's configuration.
 type InfraContainerConfig struct {
-	ConmonPidFile      string               `json:"conmonPidFile"`
-	HasInfraContainer  bool                 `json:"makeInfraContainer"`
-	NoNetwork          bool                 `json:"noNetwork,omitempty"`
-	HostNetwork        bool                 `json:"infraHostNetwork,omitempty"`
-	PortBindings       []ocicni.PortMapping `json:"infraPortBindings"`
-	StaticIP           net.IP               `json:"staticIP,omitempty"`
-	StaticMAC          net.HardwareAddr     `json:"staticMAC,omitempty"`
-	UseImageResolvConf bool                 `json:"useImageResolvConf,omitempty"`
-	DNSServer          []string             `json:"dnsServer,omitempty"`
-	DNSSearch          []string             `json:"dnsSearch,omitempty"`
-	DNSOption          []string             `json:"dnsOption,omitempty"`
-	UseImageHosts      bool                 `json:"useImageHosts,omitempty"`
-	HostAdd            []string             `json:"hostsAdd,omitempty"`
-	Networks           []string             `json:"networks,omitempty"`
-	ExitCommand        []string             `json:"exitCommand,omitempty"`
-	InfraImage         string               `json:"infraImage,omitempty"`
-	InfraCommand       []string             `json:"infraCommand,omitempty"`
-	Slirp4netns        bool                 `json:"slirp4netns,omitempty"`
-	NetworkOptions     map[string][]string  `json:"network_options,omitempty"`
+	ConmonPidFile      string                `json:"conmonPidFile"`
+	HasInfraContainer  bool                  `json:"makeInfraContainer"`
+	NoNetwork          bool                  `json:"noNetwork,omitempty"`
+	HostNetwork        bool                  `json:"infraHostNetwork,omitempty"`
+	PortBindings       []ocicni.PortMapping  `json:"infraPortBindings"`
+	StaticIP           net.IP                `json:"staticIP,omitempty"`
+	StaticMAC          net.HardwareAddr      `json:"staticMAC,omitempty"`
+	UseImageResolvConf bool                  `json:"useImageResolvConf,omitempty"`
+	DNSServer          []string              `json:"dnsServer,omitempty"`
+	DNSSearch          []string              `json:"dnsSearch,omitempty"`
+	DNSOption          []string              `json:"dnsOption,omitempty"`
+	UseImageHosts      bool                  `json:"useImageHosts,omitempty"`
+	HostAdd            []string              `json:"hostsAdd,omitempty"`
+	Networks           []string              `json:"networks,omitempty"`
+	ExitCommand        []string              `json:"exitCommand,omitempty"`
+	InfraImage         string                `json:"infraImage,omitempty"`
+	InfraCommand       []string              `json:"infraCommand,omitempty"`
+	Slirp4netns        bool                  `json:"slirp4netns,omitempty"`
+	NetworkOptions     map[string][]string   `json:"network_options,omitempty"`
+	ResourceLimits     *specs.LinuxResources `json:"resource_limits,omitempty"`
 }
 
 // ID retrieves the pod's ID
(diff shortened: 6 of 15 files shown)


libpod/options.go
```diff
@@ -548,14 +553,78 @@ func WithRuntimeFlags(runtimeFlags []string) RuntimeOption {
...
+					totalNew := "0-" + max
+					res = totalNew
+				}
+			}
+			if j != len(res)-1 {
+				if (string(res[j+1]) != "-") && (string(res[j+1]) != ",") && wholeNum != "-" && wholeNum != "," {
+					wholeNum += string(res[j+1])
+				}
+			}
+		}
```
jwhonce, line 582: @cdoern Let's chat tomorrow about your goal here and how to improve the code performance.

libpod/options.go
```diff
@@ -548,14 +553,78 @@ func WithRuntimeFlags(runtimeFlags []string) RuntimeOption {
...
+func WithCPUsCtr(num float64) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrRuntimeFinalized
+		}
+		if ctr.config.Pod != "" {
+			podID := ctr.config.Pod
+			pod, err := ctr.runtime.state.Pod(podID)
+			if err != nil {
+				return errors.New("could not get cpu data")
```
jwhonce, line 598:
```suggestion
				return errors.Wrap(err, "could not get cpu data")
```
Using https://pkg.go.dev/github.com/pkg/errors#Wrap or Wrapf will pass error information to caller vs throwing it away.

libpod/options.go
```diff
@@ -2342,3 +2410,51 @@ func WithVolatile() CtrCreateOption {
...
+					totalNew := "0-" + max
+					inp = totalNew
+				}
+			}
+			if j != len(inp)-1 {
+				if (string(inp[j+1]) != "-") && (string(inp[j+1]) != ",") && wholeNum != "-" && wholeNum != "," {
+					wholeNum += string(inp[j+1])
+				}
+			}
+		}
```
jwhonce, line 2454: We will cover this from above.

libpod/pod.go
```diff
@@ -70,6 +71,10 @@ type PodConfig struct {
 
 	// ID of the pod's lock
 	LockID uint32 `json:"lockID"`
+	//
```
jwhonce, line 74: We need comments for the documentation not empty comments.

libpod/pod.go
```diff
@@ -110,6 +115,8 @@ type InfraContainerConfig struct {
 	InfraCommand       []string             `json:"infraCommand,omitempty"`
 	Slirp4netns        bool                 `json:"slirp4netns,omitempty"`
 	NetworkOptions     map[string][]string  `json:"network_options,omitempty"`
+	CPUs               float64              `json:"cpus,omitempty"`
+	ResourceLimits     *spec.LinuxResources `json:"resource_limits"`
```
jwhonce, line 119: Given these are new fields both should be marked omitempty.

pkg/specgen/generate/pod_create.go
```diff
@@ -54,6 +56,36 @@ func createPodOptions(p *specgen.PodSpecGenerator, rt *libpod.Runtime) ([]libpod
...
+						totalNew := "0-" + max
+						p.ResourceLimits.CPU.Cpus = totalNew
+					}
+				}
+				if j != len(p.ResourceLimits.CPU.Cpus)-1 {
+					if (string(p.ResourceLimits.CPU.Cpus[j+1]) != "-") && (string(p.ResourceLimits.CPU.Cpus[j+1]) != ",") && wholeNum != "-" && wholeNum != "," {
+						wholeNum += string(p.ResourceLimits.CPU.Cpus[j+1])
+					}
+				}
+			}
```
jwhonce, line 82: Ditto from above.

jwhonce (changes requested):

libpod/options.go
cdoern: Okay, thank you @jwhonce! I just implemented a slightly better way using regex, I basically want to check if the cpuset-cpus string contains a CPU core value greater than the total number of cores available. If it does, I reset the value to `0-availableCores`. I left the other code in a comment just in case we wan to look at it.

libpod/pod.go
cdoern: Sorry, I must have missed some comments before committing, should be fixed now.

mheon:
I think the approach here is wrong. We should be applying these to the Pod cgroup, which the containers are children of - not to the individual containers of the pod.

mheon:
I wonder if we can pass these into the infra container, have it create the cgroup, and use the infra container's cgroup as parent for all other containers in the pod.

@giuseppe Any reason we couldn't use a container's cgroup as a cgroup parent for other containers? I'd like to use v2's resource inheritance to set overall resource limits for the pod.

cdoern:
> I think the approach here is wrong. We should be applying these to the Pod cgroup, which the containers are children of - not to the individual containers of the pod.

could you point me in the right direction for taking this approach? I see that the pod has a `CgroupParent` field but I don't see anything about setting the resources of the cgroup

libpod/container_validate.go
```diff
@@ -128,8 +128,10 @@ func (c *Container) validate() error {
 
 	// If User in the OCI spec is set, require that c.config.User is set for
 	// security reasons (a lot of our code relies on c.config.User).
-	if c.config.User == "" && (c.config.Spec.Process.User.UID != 0 || c.config.Spec.Process.User.GID != 0) {
-		return errors.Wrapf(define.ErrInvalidArg, "please set User explicitly via WithUser() instead of in OCI spec directly")
+	if c.config.Spec != nil {
+		if c.config.User == "" && (c.config.Spec.Process.User.UID != 0 || c.config.Spec.Process.User.GID != 0) {
+			return errors.Wrapf(define.ErrInvalidArg, "please set User explicitly via WithUser() instead of in OCI spec directly")
```
TomSweeneyRedHat, line 133: Is an end user going to know what WithUser() is?

libpod/options.go
```diff
@@ -2342,3 +2423,64 @@ func WithVolatile() CtrCreateOption {
...
+						inp = totalNew
+					}
+				}
+				if j != len(inp)-1 {
+					if (string(inp[j+1]) != "-") && (string(inp[j+1]) != ",") && wholeNum != "-" && wholeNum != "," {
+						wholeNum += string(inp[j+1])
+					}
+				}
+			}
+		*/
```
TomSweeneyRedHat, line 2480: If not needed, yank this or add a todo.

libpod/container_validate.go
cdoern: Not sure, this line was in there beforehand (I think?) I added the `c.config.Spec != nil` line so a de-reference error doesn't happen

libpod/options.go
cdoern: Will do, I think @jwhonce wanted to take a look at my different implementations so I left them in for a bit.

libpod/options.go
```diff
@@ -2342,3 +2440,82 @@ func WithVolatile() CtrCreateOption {
...
+		if pod.valid {
+			return define.ErrPodFinalized
+		}
+		ret, err := parsers.ParseUintList(inp)
+		if err != nil {
+			errors.Wrapf(err, "could not parse list")
+		}
+		for i := 0; i < len(ret); i++ {
+			if ret[i] {
+				if i > runtime.NumCPU() {
```
jwhonce, line 2471:
```suggestion
				if i > sysinfo.NumCPU() {
```
I believe this is more compatible, only gotcha is you need to fail back to runtime.NumCPU() when sysinfo.NumCPU() returns 0

jwhonce (changes requested):

libpod/container_validate.go
```diff
@@ -128,8 +128,10 @@ func (c *Container) validate() error {
 
 	// If User in the OCI spec is set, require that c.config.User is set for
 	// security reasons (a lot of our code relies on c.config.User).
-	if c.config.User == "" && (c.config.Spec.Process.User.UID != 0 || c.config.Spec.Process.User.GID != 0) {
-		return errors.Wrapf(define.ErrInvalidArg, "please set User explicitly via WithUser() instead of in OCI spec directly")
+	if c.config.Spec != nil {
```
mheon, line 131: Woah. Where is this happening? `c.config.Spec` should never be nil?

libpod/container_validate.go
mheon: If this line pops up, whoever is writing code against Libpod messed up, not the user.

libpod/define/pod_inspect.go
```diff
@@ -51,6 +53,10 @@ type InspectPodData struct {
 	// Containers gives a brief summary of all containers in the pod and
 	// their current status.
 	Containers []InspectPodContainerInfo `json:"Containers,omitempty"`
+	// CPUs contains the total number of executable CPUs in the pod
+	CPUs float64 `json:"cpus,omitempty"`
+	// ResourceLimits contains linux specific CPU data for the pod
+	ResourceLimits *spec.LinuxResources `json:"resource_limits,omitempty"`
```
mheon, line 59: I wouldn't throw this entire struct in Inspect - we should output fields from it, but the struct itself is not intended to be human-readable

libpod/options.go
```diff
@@ -548,14 +553,108 @@ func WithRuntimeFlags(runtimeFlags []string) RuntimeOption {
 
 // Container Creation Options
 
+// WithCpuSetCpusCtr computes and sets the Cpus linux resource string which determines the amount of cores, from those available,  we are allowed to execute on
+func WithCPUSetCPUsCtrInfra(res string) CtrCreateOption {
```
mheon, line 557: I would prefer to have this just take an array of ints, and have the string parsing be done in the command line - that's generally been our approach with Libpod.

mheon: Also - why do we need the per-container options here? The container code should not need changing. The fact that they are in the pod cgroup should be sufficient.

libpod/options.go
```diff
@@ -2342,3 +2440,82 @@ func WithVolatile() CtrCreateOption {
 		return nil
 	}
 }
+
+// WithCpusCtr computes the CPU period and quota given the CPU core count for a pod
+func WithCPUs(num float64) PodCreateOption {
```
mheon, line 2445: Suggest renaming to `WithPodCPUs` and `WithPodCPUSetCPUs`

libpod/pod.go
```diff
@@ -70,6 +72,12 @@ type PodConfig struct {
 
 	// ID of the pod's lock
 	LockID uint32 `json:"lockID"`
+
+	//ResourceLimits contains linux specific CPU data for the pod
+	ResourceLimits *spec.LinuxResources `json:"resource_limits,omitempty"`
+
+	// CPUs to use within the cpuset. Default is to use any CPU available.
+	CPUs float64 `json:"cpus,omitempty"`
```
mheon, line 80: Why is this separate from ResourceLimits? I believe that CPUs is a part of ReousrceLimits?

libpod/pod.go
```diff
@@ -128,6 +138,16 @@ func (p *Pod) Namespace() string {
 	return p.config.Namespace
 }
 
+// ResourceLim returns the cpuset resource limits for the pod
+func (p *Pod) ResourceLim() *spec.LinuxResources {
+	return p.config.ResourceLimits
```
mheon, line 143: Return a copy of the struct, so it can't be altered from outside Libpod

libpod/pod.go
```diff
@@ -209,6 +229,27 @@ func (p *Pod) CgroupPath() (string, error) {
 		return "", err
 	}
 
+	if !p.HasInfraContainer() {
```
mheon, line 232: We should check `p.state.CgroupPath` here first, and if it's set, we can return it immediately.

libpod/pod.go
```diff
@@ -209,6 +229,27 @@ func (p *Pod) CgroupPath() (string, error) {
...
+		if err != nil {
+			return "", errors.Wrapf(err, "could not get infra")
+		}
+		if ctr != nil {
+			ctr.Start(context.Background(), false)
+			cgroupPath, err := ctr.CGroupPath()
+			if err != nil {
+				return "", errors.Wrapf(err, "could not get container cgroup")
+			}
+			p.state.CgroupPath = cgroupPath
```
mheon, line 249: We need a `p.save()` after this to save the state

libpod/runtime_pod_infra_linux.go
```diff
@@ -200,7 +207,7 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
 	if len(p.config.InfraContainer.ConmonPidFile) > 0 {
 		options = append(options, WithConmonPidFile(p.config.InfraContainer.ConmonPidFile))
 	}
-
+	//g.Config.Linux.Resources = p.config.InfraContainer.ResourceLimits
```
mheon, line 210: This is more correct than using options - the spec should be set directly.

libpod/options.go
cdoern: @mheon is this true for infra as well? If so I can snip a lot of code I just wasn't sure

libpod/pod.go
cdoern: Here I was just passing CPUs along as it is a float64 vs the resourceConfig.CPU.Cpus is a string and of a different format `CPUs = 4.0` vs `CPU.Cpus=0-2,4`

libpod/define/pod_inspect.go
```diff
@@ -51,6 +51,10 @@ type InspectPodData struct {
 	// Containers gives a brief summary of all containers in the pod and
 	// their current status.
 	Containers []InspectPodContainerInfo `json:"Containers,omitempty"`
+	// CPUs contains the total number of executable CPUs in the pod
+	CPUs float64 `json:"cpus,omitempty"`
```
jwhonce, line 55:
```suggestion
	// CPUs contains the total number of executable CPUs in the pod
	CPUs int `json:"cpus,omitempty"`
```
Partial CPU's isn't supported, Correct? NumCPU() signature is int.

libpod/define/pod_inspect.go
```diff
@@ -91,6 +95,10 @@ type InspectPodInfraConfig struct {
 	Networks []string
 	// NetworkOptions are additional options for each network
 	NetworkOptions map[string][]string
+	// CPUs contains the total number of executable CPUs in the container
+	CPUs float64
```
jwhonce, line 99:
```suggestion
	// CPUs contains the total number of executable CPUs in the container
	CPUs int
```
See above.

libpod/options.go
```diff
@@ -2342,3 +2341,33 @@ func WithVolatile() CtrCreateOption {
 		return nil
 	}
 }
+
+// WithCpusCtr computes the CPU period and quota given the CPU core count for a pod
```
jwhonce, line 2345:
```suggestion
// WithPodCPUs computes the CPU period and quota given the CPU core count for a pod
```

libpod/pod.go
```diff
@@ -128,6 +138,17 @@ func (p *Pod) Namespace() string {
...
 }
 
+// ResourceLim returns the cpuset resource limits for the pod
+func (p *Pod) ResourceLim() *spec.LinuxResources {
+	resCopy := p.config.ResourceLimits
+	return resCopy
+}
+
+// CPUs returns the specified number of CPUs for the pod
+func (p *Pod) CPUs() float64 {
```
jwhonce, line 148:
```suggestion
func (p *Pod) CPUs() int {
```
NumCPU() signature is int, not float64

pkg/domain/entities/pods.go
```diff
@@ -116,13 +118,35 @@ type PodCreateOptions struct {
 	Name               string
 	Net                *NetOptions
 	Share              []string
+	Cpus               float64
+	CpusetCpus         string
```
jwhonce, line 122:
```suggestion
	Cpus               int
	CpusetCpus         string
```

pkg/specgen/podspecgen.go
```diff
@@ -155,6 +157,14 @@ type PodSpecGenerator struct {
...
 	PodNetworkConfig
 	PodCgroupConfig
+	PodResourceConfig
+}
+
+type PodResourceConfig struct {
+	// ResourceLimits contains linux specific CPU data for the pod
+	ResourceLimits *spec.LinuxResources `json:"resource_limits,omitempty"`
+	// CPUs to use within the cpuset. Default is to use any CPU available.
+	CPUs float64 `json:"cpus,omitempty"`
```
jwhonce, line 167:
```suggestion
	CPUs int `json:"cpus,omitempty"`
```

pkg/specgen/specgen.go
```diff
@@ -470,6 +470,8 @@ type ContainerResourceConfig struct {
 	// that are used to configure cgroup v2.
 	// Optional.
 	CgroupConf map[string]string `json:"unified,omitempty"`
+	// CPUs to use within the cpuset. Default is to use any CPU available.
+	CPUs float64 `json:"cpus,omitempty"`
```
jwhonce, line 474:
```suggestion
	CPUs int `json:"cpus,omitempty"`
```

cmd/podman/pods/create.go
```diff
@@ -185,6 +197,43 @@ func create(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	numCPU := sysinfo.NumCPU()
+	if numCPU == 0 {
+		numCPU = runtime.NumCPU()
+	}
+	if createOptions.Cpus > float64(numCPU) {
+		createOptions.Cpus = float64(numCPU)
+	}
```
jwhonce, line 206:
```suggestion
	if createOptions.Cpus > (numCPU) {
		createOptions.Cpus = (numCPU)
	}
```

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -23,6 +23,22 @@ Add a host to the /etc/hosts file shared between all containers in the pod.
 
 Path to cgroups under which the cgroup for the pod will be created. If the path is not absolute, the path is considered to be relative to the cgroups path of the init process. Cgroups will be created if they do not already exist.
 
+#### **--cpus**=*amount*
+
+Set the total number of CPUs delegated to the pod. Default is 0.000 which indicates that there is no limit on computation power.
+
+#### **--cpuset-cpus**=*amount*
+
+Sets the total number of available CPUs to allow execution in. Unlike --cpus this is of type string and parsed as a list of numbers
```
jwhonce, line 32:
```suggestion
Limit the CPUs to support execution. First CPU is numbered 0. Unlike --cpus this is of type string and parsed as a list of numbers
```

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -23,6 +23,22 @@ Add a host to the /etc/hosts file shared between all containers in the pod.
 
 Path to cgroups under which the cgroup for the pod will be created. If the path is not absolute, the path is considered to be relative to the cgroups path of the init process. Cgroups will be created if they do not already exist.
 
+#### **--cpus**=*amount*
+
+Set the total number of CPUs delegated to the pod. Default is 0.000 which indicates that there is no limit on computation power.
```
jwhonce, line 28:
```suggestion
Set the total number of CPUs delegated to the pod. Default is 0 which indicates that there is no limit on computation power.
```

jwhonce (changes requested):
Looking better. I believe field types should be int rather than float64, as given by the input type.

libpod/define/pod_inspect.go
cdoern: @jwhonce I could be wrong but @rhatdan told me float64 in the doc he shared w/ me. The default is 0.000, I guess you can specify a portion of a core?

rhatdan: Could have been a cut and paste issue.  But take a look at this field in containers?

cdoern: @rhatdan we figured it out for the most part. Decided to read in the CPUs as a float but then immediately use it to calculate the quota and period and just store those as uints/ints in the intermediate structs.

rhatdan: SGTM

libpod/runtime_pod_infra_linux.go
```diff
@@ -200,7 +197,15 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
 	if len(p.config.InfraContainer.ConmonPidFile) > 0 {
 		options = append(options, WithConmonPidFile(p.config.InfraContainer.ConmonPidFile))
 	}
-
+	newRes := new(spec.LinuxCPU)
+	if p.config.InfraContainer.CPUPeriod != 0 && p.config.InfraContainer.CPUQuota != 0 {
+		newRes.Period = &p.config.InfraContainer.CPUPeriod
+		newRes.Quota = &p.config.InfraContainer.CPUQuota
+	}
+	if p.config.InfraContainer.ResourceLimits.CPU.Cpus != "" {
```
rhatdan, line 205: No need for this check since I assume newRes.Cpus by default is ""

libpod/runtime_pod_linux.go
```diff
@@ -50,6 +50,16 @@ func (r *Runtime) NewPod(ctx context.Context, options ...PodCreateOption) (_ *Po
 		pod.config.Name = name
 	}
 
+	if (pod.config.InfraContainer.CPUPeriod == 0 || pod.config.InfraContainer.CPUQuota == 0) && (pod.config.CPUPeriod != 0 && pod.config.CPUQuota != 0) {
+		pod.config.InfraContainer.CPUPeriod, pod.config.InfraContainer.CPUQuota = pod.config.CPUPeriod, pod.config.CPUQuota
+	}
+
+	pod.config.InfraContainer.ResourceLimits = &spec.LinuxResources{}
+	pod.config.InfraContainer.ResourceLimits.CPU = &spec.LinuxCPU{}
+	if pod.config.InfraContainer.ResourceLimits != nil && pod.config.InfraContainer.ResourceLimits.CPU.Cpus == "" && pod.config.ResourceLimits != nil && pod.config.ResourceLimits.CPU.Cpus != "" {
```
rhatdan, line 59: You just set the two fields above so the checks in this line are not necessary.

Just set the value.

libpod/runtime_pod_linux.go
```diff
@@ -208,6 +218,7 @@ func (r *Runtime) removePod(ctx context.Context, p *Pod, removeCtrs, force bool)
 		}
 
 		// New resource limits
+
```
rhatdan, line 221: Delete blank line.

pkg/domain/entities/pods.go
```diff
@@ -116,13 +118,35 @@ type PodCreateOptions struct {
...
 }
 
 type PodCreateReport struct {
 	Id string //nolint
 }
 
-func (p PodCreateOptions) ToPodSpecGen(s *specgen.PodSpecGenerator) {
+func getCPULimits(p *PodCreateOptions) *specs.LinuxCPU {
```
rhatdan, line 129: Why not 
```
func (p *PodCreateOptions)CPULimits *specs.LinuxCPU {
```

pkg/domain/entities/pods.go
```diff
@@ -156,6 +180,21 @@ func (p PodCreateOptions) ToPodSpecGen(s *specgen.PodSpecGenerator) {
 
 	// Cgroup
 	s.CgroupParent = p.CGroupParent
+
+	// Resource config
+	cpuDat := getCPULimits(p)
```
rhatdan, line 185: cpuDat = p.CPULimits()

pkg/specgen/generate/pod_create.go
```diff
@@ -54,6 +54,13 @@ func createPodOptions(p *specgen.PodSpecGenerator, rt *libpod.Runtime) ([]libpod
 	if len(p.Name) > 0 {
 		options = append(options, libpod.WithPodName(p.Name))
 	}
+	if p.CPUPeriod != 0 || p.CPUQuota != 0 {
+		options = append(options, libpod.WithPodCPUPAQ((p.CPUPeriod), (p.CPUQuota)))
+		//options = append(options, libpod.WithPodCPUs(p.CPUs))
```
rhatdan, line 59: Don't leave commented out code.

test/e2e/pod_create_test.go
```diff
@@ -514,5 +517,47 @@ ENTRYPOINT ["sleep","99999"]
 		create.WaitWithDefaultTimeout()
 		Expect(create.ExitCode()).To(BeZero())
 	})
+	It("podman pod create --cpus", func() {
```
rhatdan, line 520: Add empty line

test/e2e/pod_create_test.go
```diff
@@ -514,5 +517,47 @@ ENTRYPOINT ["sleep","99999"]
...
+
+		podInspect := podmanTest.Podman([]string{"pod", "inspect", podName})
+		podInspect.WaitWithDefaultTimeout()
+		Expect(podInspect.ExitCode()).To(Equal(0))
+		podJSON := podInspect.InspectPodToJSON()
+		Expect(podJSON.CPUPeriod).To(Equal(period))
+		Expect(podJSON.CPUQuota).To(Equal(quota))
+
+	})
+	It("podman pod create --cpuset-cpus", func() {
```
rhatdan, line 541: Add empty line

test/e2e/pod_create_test.go
```diff
@@ -514,5 +517,47 @@ ENTRYPOINT ["sleep","99999"]
...
+		contCreate := podmanTest.Podman([]string{"container", "create", "--name", ctrName, "--pod", podName, "alpine"})
+		contCreate.WaitWithDefaultTimeout()
+		Expect(podCreate.ExitCode()).To(Equal(0))
+
+		podInspect := podmanTest.Podman([]string{"pod", "inspect", podName})
+		podInspect.WaitWithDefaultTimeout()
+		Expect(podInspect.ExitCode()).To(Equal(0))
+		podJSON := podInspect.InspectPodToJSON()
+		Expect(podJSON.CPUSetCPUs).To(Equal(in))
+
```
rhatdan, line 560: Remove empty line

libpod/runtime_pod_linux.go
cdoern: Sorry, that was leftover from testing what was causing nil pointer error b/c ResourceLimits and ResourceLimits.CPU are both pointers.

cdoern: > You just set the two fields above so the checks in this line are not necessary.
> 
> Just set the value.

changing it to `if p.config.InfraContainer.ResourceLimits != nil && p.config.ResourceLimits.CPU != nil` since it you do a simple `podman pod create` ResourceLimits.CPU (I think) will be a nil pointer to a nil pointer

libpod/runtime_pod_infra_linux.go
cdoern: > No need for this check since I assume newRes.Cpus by default is ""

changing it to `if p.config.InfraContainer.ResourceLimits != nil && p.config.ResourceLimits.CPU != nil` since if you do a simple `podman pod create` ResourceLimits.CPU (I think) will be a nil pointer to a nil pointer

libpod/pod.go
```diff
@@ -70,6 +72,14 @@ type PodConfig struct {
 
 	// ID of the pod's lock
 	LockID uint32 `json:"lockID"`
+
+	//ResourceLimits contains linux specific CPU data for the pod
+	ResourceLimits *spec.LinuxResources `json:"resource_limits,omitempty"`
```
mheon, line 77: This should be in InfraConfig

libpod/pod.go
```diff
@@ -70,6 +72,14 @@ type PodConfig struct {
...
+
+	// CPU period of the cpuset, determined by --cpus
+	CPUPeriod uint64 `json:"cpu_period,omitempty"`
```
mheon, line 80: Do we need to store these separately? They're already in ResourceLimits

libpod/pod.go
```diff
@@ -128,6 +141,22 @@ func (p *Pod) Namespace() string {
 	return p.config.Namespace
 }
 
+// ResourceLim returns the cpuset resource limits for the pod
+func (p *Pod) ResourceLim() *spec.LinuxResources {
+	resCopy := p.config.ResourceLimits
```
mheon, line 146: Use JSONDeepCopy for this

mheon: Basically, this is just making a copy of the pointer, we want a copy of what it points to

libpod/pod.go
cdoern: I was keeping them around just to be explicit and also helps avoid more handling of nil pointers with a generic podman pod create @mheon

libpod/pod.go
cdoern: in addition to podConfig, correct? if I remove this it breaks pretty much everything i've written. PodConfig uses this to pass to InfraConfig down the line.

mheon: Why do we need two copies? When we do this for other infra container fields, we store a single copy in `InfraConfig` - duplicating this in two places in the Pod structure in the DB seems unnecessary?

libpod/pod.go
mheon: How so? It seems like we ought to be able to avoid having the uints if we just nil-check ResourceLimits before accessing it?

libpod/pod.go
cdoern: Yep you're right, just  had to change my options.go code to populate the infra fields instead.

libpod/pod.go
cdoern: ok, I will just nil check. Should I keep the getter functions for quota and period?

mheon: Sure, they can just look at the ResourceLimits bit. Best part, if you put the null checks in there, and use the getters/setters everywhere - you only need to add them once

libpod/pod.go
```diff
@@ -128,6 +131,48 @@ func (p *Pod) Namespace() string {
...
+		return nil
+	}
+	if resCopy != nil && resCopy.CPU != nil {
+		return resCopy
+	}
+	pe := uint64(0)
+	q := int64(0)
+	c := ""
+	empty := &specs.LinuxResources{
+		CPU: &specs.LinuxCPU{Period: &pe, Quota: &q, Cpus: c},
```
jwhonce, line 147:
```suggestion
		CPU: &specs.LinuxCPU{},
```
nit, not need to populate the fields with go Zero-values.

jwhonce (commented):
LGTM

libpod/pod.go
cdoern: @jwhonce Period and Quota are pointers, does this affect access (with a nil pointer deref) if i don't initialize them?

cdoern: nevermind, just had to change a line that checked `*.CPU.Period`'s value to checking if `.CPU.Period != nil`

cdoern:
@rhatdan @mheon PTAL

PR #10728, opened 2021-06-19
support container to container copy

### LINKED ISSUE #7370: support container-to-container copy
```
podman cp v2ray:/etc/ssl/certs/ca-certificates.crt caddy:/etc/ssl/certs/ca-certificates.crt
Error: invalid arguments v2ray:/etc/ssl/certs/ca-certificates.crt, caddy:/etc/ssl/certs/ca-certificates.crt you must use just one container
```
There seems to be no harm in implementing this function.

infiniteregrets:
This would fix #7370. I haven't done anything yet, but this seems to work:
```
[mehul@fedora bin]$ ./podman container exec 59aef1582431 /bin/sh -c 'mkdir test'
[mehul@fedora bin]$ ./podman container cp ffae1af127e0:/tmp 59aef1582431:/test
[mehul@fedora bin]$ ./podman container exec 59aef1582431 /bin/sh -c 'cd test && ls'
tmp
```
I am missing some cases where a file to file copy would occur or if the path is a symbolic link and read/write should be recursive. Once I am sure about how everything works I will fix that. I think checking whether a path exists on a container and getting containerinfo can be speeded up by doing so in a separate thread

I am unable to test much locally because my laptop gets really hot and laggy, so I am just pushing this here. 
Thanks!
Signed-off-by: Mehul Arora <aroram18@mcmaster.ca>


rhatdan:
You will need to add tests for this.

@vrothberg PTAL

cmd/podman/containers/cp.go
```diff
@@ -113,6 +116,55 @@ func doCopy(funcA func() error, funcB func() error) error {
...
+	}
+
+	fromContainerInfo, err := registry.ContainerEngine().ContainerStat(registry.GetContext(), fromContainer, sourcePath)
+	if err != nil {
+		return errors.Wrapf(err, "%q could not be found on container %s", sourcePath, fromContainer)
+	}
+
+	toContainerInfo, err := registry.ContainerEngine().ContainerStat(registry.GetContext(), toContainer, destPath)
+	if err != nil {
+		return errors.Wrapf(err, "%q could not be found on container %s", destPath, toContainer)
```
vrothberg, line 135: That should not be an error per-se.  If I do `cp 1:/foo.txt 2:/foo.txt` then "foo.txt" should be created at the destination.  Have a look at the copy functions above for the wiring.

vrothberg (commented):
Thanks!

Yes, we need _a lot of_ tests for this one.  Look at the tests in https://github.com/containers/podman/blob/master/test/system/065-cp.bats which have a decent coverage of all corner cases and the basic copy functionality.

cmd/podman/containers/cp.go
infiniteregrets: I am on it, I didn't understand why we are checking for this condition here: https://github.com/containers/podman/blob/master/cmd/podman/containers/cp.go#L266
If possible, could you please give an example

vrothberg: If we do a `cp 1:/dir 2:/tmp/foo/` `foo` must exist. If we do a `cp 1:/dir 2:/tmp/foo` `foo` would created if it doesn't exist.

The condition enforces that rule.  The `podman-cp` man pages mention these rules.

infiniteregrets: Thanks for adding those comments btw, they are very helpful! I will work on the missing cases soon

vrothberg: Happy to help! Feel free to ping me here on IRC at any time.  I am reachable during regular CEST working hours.

infiniteregrets: Sorry for the late reply, I didn't get time to work on this. 
I also haven't been able to get IRC to work for some reason. But anyway, I will ask here 
So I am confused about the case where a file exists on one container say ok.txt and then it is being copied like:
`podman container cp distracted_tesla:ok.txt pedantic_hertz:ok.txt` or `podman container cp distracted_tesla:ok.txt pedantic_hertz:abdcd` 
Error:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x166ae92]

goroutine 1 [running]:
```
How will I go about handling this? Can you point me in the right direction in the existing code?

And then, say a file exists in both containers then this error occurs:
```
[mehul@fedora bin]$ ./podman container cp distracted_tesla:ok.txt pedantic_hertz:/
[mehul@fedora bin]$ ./podman container cp distracted_tesla:ok.txt pedantic_hertz:ok.txt
Error: 1 error occurred:
        * error copying to container: copier: put: /ok.txt (/ok.txt): exists but is not a directory
```
Not completely sure how it is being handled for container to host and host to container copy but is the buildah package playing a role in this? Any hints?

Thanks!

vrothberg: > Sorry for the late reply, I didn't get time to work on this.
> I also haven't been able to get IRC to work for some reason. But anyway, I will ask here
> So I am confused about the case where a file exists on one container say ok.txt and then it is being copied like:
> `podman container cp distracted_tesla:ok.txt pedantic_hertz:ok.txt` or `podman container cp distracted_tesla:ok.txt pedantic_hertz:abdcd`
> Error:
> 
> ```
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x166ae92]
> 
> goroutine 1 [running]:
> ```
> 
> How will I go about handling this? Can you point me in the right direction in the existing code?

Below `goroutine 1 [...]` should be a stack trace that points to the exact line of code of the nil derref.

> And then, say a file exists in both containers then this error occurs:
> 
> ```
> [mehul@fedora bin]$ ./podman container cp distracted_tesla:ok.txt pedantic_hertz:/
> [mehul@fedora bin]$ ./podman container cp distracted_tesla:ok.txt pedantic_hertz:ok.txt
> Error: 1 error occurred:
>         * error copying to container: copier: put: /ok.txt (/ok.txt): exists but is not a directory
> ```
> 
> Not completely sure how it is being handled for container to host and host to container copy but is the buildah package playing a role in this? Any hints?

container->host and host->container seems to allows that but it should fail. Looks like you found a bug :)

vrothberg: I can tackle that bug. Thanks for finding that!

vrothberg: Opened https://github.com/containers/podman/pull/10772

infiniteregrets: Thank you for fixing that!

infiniteregrets: @vrothberg a bit confused about git, how do I get the changes you made to the branch on my fork? 
would it be like checking out to main and then pulling the changes and checking out to my branch `c2c-copy` and then merging main to my branch? 
wouldn't that mean losing the changes I made?

vrothberg: What I usually do is the following (`upstream` is the remote branch pointing to `github.com/containers/podman`):
```
$ git remote update
$ git pull --rebase upstream master
```

Note that while rebasing you may have to resolve merge conflicts.

infiniteregrets: been on this for hours now, cant seem to figure this. I was able to update the c2c-copy branch to the recent changes on podman but the changes you made yesterday don't reflect? even git log shows your merged PR

I think I might have to just re-add those changes and continue

infiniteregrets: Nevermind, I figured it out. Didn't have to do anything manually

infiniteregrets: @vrothberg so I am not sure how we will go about renaming files for container-container copy in a case like:
`podman container cp distracted_tesla:ok.txt pedantic_hertz:ok.txt`
In host->container or container->host I can see that the buildah copier package is handling that for us but how do we do this for this case?

I was able to fix the segmentation fault 
```
[mehul@fedora bin]$ ./podman cp c1:file.txt c2:/skbxlswx
[mehul@fedora bin]$ ./podman exec c2 /bin/sh -c "ls"
bin
dev
etc
file.txt
home
proc
root
run
sys
tmp
usr
var
```

cmd/podman/containers/cp.go
```diff
@@ -113,6 +116,82 @@ func doCopy(funcA func() error, funcB func() error) error {
...
+	if err := containerMustExist(toContainer); err != nil {
+		return err
+	}
+
+	fromContainerInfo, err := registry.ContainerEngine().ContainerStat(registry.GetContext(), fromContainer, sourcePath)
+	if err != nil {
+		return errors.Wrapf(err, "%q could not be found on container %s", sourcePath, fromContainer)
+	}
+
+	var toContainerBaseName string
```
cdoern, line 133: can you change to `toContainerBaseName := ""` removes an unnecessary line.

cmd/podman/containers/cp.go
```diff
@@ -113,6 +116,82 @@ func doCopy(funcA func() error, funcB func() error) error {
...
+		return errors.New("destination must be a directory when copying a directory")
+	}
+
+	fromContainerTarget, toContainerTarget := fromContainerInfo.LinkTarget, toContainerInfo.LinkTarget
+	if !toContainerInfo.IsDir {
+		toContainerTarget = filepath.Dir(destPath)
+	}
+	reader, writer := io.Pipe()
+
+	fromContainerCopy := func() error {
```
cdoern, line 168: potentially could consolidate `toContainerCopy` and `fromContainer` copy into a function you call twice that takes specific parameters. Seems slightly redundant to have nearly identical code twice.

cmd/podman/containers/cp.go
cdoern: Also, I am seeing you are getting an ineffectual assignment error here. Where are you using `toContainerBaseName` after setting it?

cmd/podman/containers/cp.go
infiniteregrets: there might be some changes to the functions, so depending on what the final outcome looks like i'll refactor the code

cmd/podman/containers/cp.go
infiniteregrets: > can you change to toContainerBaseName := "" removes an unnecessary line.

sure, will make the changes once I figure everything out 

> Also, I am seeing you are getting an ineffectual assignment error here. Where are you using toContainerBaseName after setting it?

I am not completely sure tbh, I had to do _ = toContainerBaseName to get rid of that unused variable error

cdoern: ok @infiniteregrets you're getting that error because you set the base name and never actually use it. the `_ =` is just a clever way to trick golang. You should use that name somewhere in your copy code.

infiniteregrets: @cdoern I am using it? 
```
if toContainerInfo != nil {
			toContainerBaseName = filepath.Base(toContainerInfo.LinkTarget)
		} else {
			toContainerBaseName = filepath.Base(destPath)
		}
```		
This looks like a shadowing issue? Im not sure but I'll try to figure it. Thanks!

vrothberg:
Please create a new PR against the new `main` branch. @rhatdan, the master->main migration closes on PRs against master.

PR #10736, opened 2021-06-21
read secret config from config file if no user data.

trusch:
**What**

feat: read secret config from config file if the user hasn't entered
explicit config values

feat: allow to specify `--driver-opts opt1=val1,opt2=val2` in the secret
create command to allow overriding the default values

fix: show driver options in `podman secret inspect`

**Why**

containers/common@0.40.0 contains a new secret section in the definition of the containers.conf file so this PR reads the secret config when creating a new secret and uses its content as default values when the user did not supply any overrides (the user can specifiy the driver at runtime and since this PR also the driver options)

This also contains a little fix that includes the driver options in the output of `podman secret inspect`. Until now it didn't hurt to not have this information there, because it was only a non-configurable root directory for the file driver, but now there is also the pass driver with more interesting config options.

### DIFF
--- a/cmd/podman/secrets/create.go
+++ b/cmd/podman/secrets/create.go
@@ -47,7 +47,7 @@ func init() {
 	cfg := registry.PodmanConfig()
 
 	flags.StringVar(&createOpts.Driver, driverFlagName, cfg.Secrets.Driver, "Specify secret driver")
-	flags.StringToStringVar(&createOpts.Opts, optsFlagName, cfg.Secrets.Opts, "Specify driver specific options")
+	flags.StringToStringVar(&createOpts.DriverOpts, optsFlagName, cfg.Secrets.Opts, "Specify driver specific options")
 	_ = createCmd.RegisterFlagCompletionFunc(driverFlagName, completion.AutocompleteNone)
 	_ = createCmd.RegisterFlagCompletionFunc(optsFlagName, completion.AutocompleteNone)
 
--- a/cmd/podman/secrets/create.go
+++ b/cmd/podman/secrets/create.go
@@ -42,8 +42,14 @@ func init() {
 	flags := createCmd.Flags()
 
 	driverFlagName := "driver"
-	flags.StringVar(&createOpts.Driver, driverFlagName, "file", "Specify secret driver")
+	optsFlagName := "driver-opts"
+
+	cfg := registry.PodmanConfig()
+
+	flags.StringVar(&createOpts.Driver, driverFlagName, cfg.Secrets.Driver, "Specify secret driver")
+	flags.StringToStringVar(&createOpts.Opts, optsFlagName, cfg.Secrets.Opts, "Specify driver specific options")
 	_ = createCmd.RegisterFlagCompletionFunc(driverFlagName, completion.AutocompleteNone)
+	_ = createCmd.RegisterFlagCompletionFunc(optsFlagName, completion.AutocompleteNone)
 
 	envFlagName := "env"
 	flags.BoolVar(&env, envFlagName, false, "Read secret data from environment variable")
--- a/pkg/api/handlers/libpod/secrets.go
+++ b/pkg/api/handlers/libpod/secrets.go
@@ -40,7 +40,7 @@ func CreateSecret(w http.ResponseWriter, r *http.Request) {
 	}
 
 	opts.Driver = query.Driver
-	opts.Opts = query.DriverOpts
+	opts.DriverOpts = query.DriverOpts
 
 	ic := abi.ContainerEngine{Libpod: runtime}
 	report, err := ic.SecretCreate(r.Context(), query.Name, r.Body, opts)
--- a/pkg/api/handlers/libpod/secrets.go
+++ b/pkg/api/handlers/libpod/secrets.go
@@ -1,7 +1,9 @@
 package libpod
 
 import (
+	"encoding/json"
 	"net/http"
+	"reflect"
 
 	"github.com/containers/podman/v3/libpod"
 	"github.com/containers/podman/v3/pkg/api/handlers/utils"
@@ -16,9 +18,17 @@ func CreateSecret(w http.ResponseWriter, r *http.Request) {
 		runtime = r.Context().Value("runtime").(*libpod.Runtime)
 		decoder = r.Context().Value("decoder").(*schema.Decoder)
 	)
+
+	decoder.RegisterConverter(map[string]string{}, func(str string) reflect.Value {
+		res := make(map[string]string)
+		json.Unmarshal([]byte(str), &res)
+		return reflect.ValueOf(res)
+	})
+
 	query := struct {
-		Name   string `schema:"name"`
-		Driver string `schema:"driver"`
+		Name       string            `schema:"name"`
+		Driver     string            `schema:"driver"`
+		DriverOpts map[string]string `schema:"driveropts"`
 	}{
 		// override any golang type defaults
 	}
@@ -28,7 +38,9 @@ func CreateSecret(w http.ResponseWriter, r *http.Request) {
 			errors.Wrapf(err, "failed to parse parameters for %s", r.URL.String()))
 		return
 	}
+
 	opts.Driver = query.Driver
+	opts.Opts = query.DriverOpts
 
 	ic := abi.ContainerEngine{Libpod: runtime}
 	report, err := ic.SecretCreate(r.Context(), query.Name, r.Body, opts)
--- a/pkg/domain/entities/secrets.go
+++ b/pkg/domain/entities/secrets.go
@@ -12,6 +12,7 @@ type SecretCreateReport struct {
 
 type SecretCreateOptions struct {
 	Driver string
+	Opts   map[string]string
 }
 
 type SecretListRequest struct {
--- a/pkg/domain/entities/secrets.go
+++ b/pkg/domain/entities/secrets.go
@@ -11,8 +11,8 @@ type SecretCreateReport struct {
 }
 
 type SecretCreateOptions struct {
-	Driver string
-	Opts   map[string]string
+	Driver     string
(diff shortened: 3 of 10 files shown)


jwhonce:
LGTM, once tests are passing

ashley-cui:
This looks like it's going to break the remote API for libpod. I'm also a little wary of allowing a user to specify driver opts, and if we do, the flag name should be called `--driver-opts`, or rather something like `--driver=drivertype,opt1=opt`. Do you have a specific use case for this in the passdriver?

@vrothberg wdyt?

mheon:
How will this break the API? It looks like a straight addition, not a change

ashley-cui:
@mheon when designing the secrets libpod api I pretty naively assumed that all opts could be passed in using url params. Since the opts are a map, they have to be encoded in the body of the request (which is what we do for everything else, just not secrets). The body of the request currently doesn't take any structured data, just a io.Reader.

trusch:
@ashley-cui the options are a map, but one from string to string, so encoding it as URL params should be straight forward. Basically it's a set of key value pairs so it could be just encoded as is `?driver=file&root=/foo/bar&opt2=val2`

trusch:
@ashley-cui I totally forgot about implementing the serialisation of the opts. Can you point me to the place where it needs to be added? Also I'd like to know how to selectively run only the secret related e2e tests. I wasn't able to do so by now, and running all the tests after each iteration doesn't seem to be an option ;)

ashley-cui:
@trusch I think that that would work! I'd also be a little more comfortable with something like `driver=file&driveropts=opt%3Dval` and parsing and decoding into a map in the handler, so we can know all these opts are driveropts, and not other opts we may need to support in the future. It's hairy, but it doesn't break api, and its extendable for the future :)

trusch:
@ashley-cui I was looking for the place where the map[string]string gets converted to the url parameters to get passed to a remote podman instance. But you are right the receiving side also has to understand this change! Should I add those changes to this PR or is it ok to do it in a follow up PR? Also thanks for the hint about how to run the tests selectively, it saved a ton of time.

trusch:
@ashley-cui could you kick off the CI one more time? All the tests are fixed and the commits are properly signed off, but the CI got stuck setting up the gopath.

trusch:
I can't figure out why CI is complaining. If I run the e2e tests locally they pass and they are the only tests that I touched. In some of the checks the CI dies on setting up the gopath, but in other it explicitly complains about the secret tests but no hints about whats wrong with them.

rhatdan:
```
[+1397s]   Expected
[+1397s]       <string>: map[opt1:val path:/tmp/podman_test916977372/crio/secrets/filedriver]
[+1397s]   to equal
[+1397s]       <string>: map[opt1:val]
[+1397s] 
[+1397s]   /var/tmp/go/src/github.com/containers/podman/test/e2e/secret_test.go:54
```
Looks like the test is finding an additional secret driver.

trusch:
@rhatdan thanks for pointing this out, it actually make sense to me. It's passing locally for me because my containers.conf is different than the default and the additional path option is coming from some special handling for the file driver.

TomSweeneyRedHat:
@trusch the CI tests aren't happy

trusch:
@TomSweeneyRedHat They are failing because the libpod api is not updated to work with the new --opt flag in the secret create command. I think I did the neccessary changes now (locally). Can you or somebody else give me a hint how to run the e2e tests with podman-remote? When I run them like `go test -v test/e2e/libpod_suite_test.go test/e2e/common_test.go test/e2e/config.go test/e2e/config_amd64.go test/e2e/secret_test.go` the tests are passing because its just using the traditional control flow and not the one via the api.

rhatdan:
@ashley-cui PTAL

ashley-cui (commented):
I think there's a plumbing issue in the remote client. You can manually test the remote client using `./bin/podman system service -t0` and `./bin/podman --remote secret create` I don't know how to run the tests remotely, if anyone else does, please let us know!

cmd/podman/secrets/create.go
```diff
@@ -42,8 +42,11 @@ func init() {
 	flags := createCmd.Flags()
 
 	driverFlagName := "driver"
-	flags.StringVar(&createOpts.Driver, driverFlagName, "file", "Specify secret driver")
+	optsFlagName := "opts"
+	flags.StringVar(&createOpts.Driver, driverFlagName, "", "Specify secret driver")
+	flags.StringToStringVar(&createOpts.Opts, optsFlagName, map[string]string{}, "Specify driver specific options")
```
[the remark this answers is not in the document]
trusch: done

pkg/domain/infra/abi/secrets.go
```diff
@@ -17,15 +17,36 @@ func (ic *ContainerEngine) SecretCreate(ctx context.Context, name string, reader
...
+	}
+
+	var (
+		driver = cfg.Secrets.Driver
+		opts   = make(map[string]string)
+	)
+
+	for k, v := range cfg.Secrets.Opts {
+		opts[k] = v
+	}
```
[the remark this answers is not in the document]
trusch: done

pkg/domain/infra/tunnel/secrets.go
```diff
@@ -11,7 +11,10 @@ import (
 )
 
 func (ic *ContainerEngine) SecretCreate(ctx context.Context, name string, reader io.Reader, options entities.SecretCreateOptions) (*entities.SecretCreateReport, error) {
-	opts := new(secrets.CreateOptions).WithDriver(options.Driver).WithName(name)
+	opts := new(secrets.CreateOptions).
+		WithDriver(options.Driver).
+		WithOpts(options.Opts).
+		WithName(name)
 	created, _ := secrets.Create(ic.ClientCtx, reader, opts)
```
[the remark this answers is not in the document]
trusch: done

cmd/podman/secrets/create.go
```diff
@@ -42,8 +42,11 @@ func init() {
 	flags := createCmd.Flags()
 
 	driverFlagName := "driver"
-	flags.StringVar(&createOpts.Driver, driverFlagName, "file", "Specify secret driver")
+	optsFlagName := "opts"
```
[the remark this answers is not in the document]
trusch: done

trusch:
@ashley-cui Thanks for your hints! I found out how to run the e2e tests using the remote client:
```
go test -v **test/e2e/libpod_suite_remote_test.go** test/e2e/common_test.go test/e2e/config.go test/e2e/config_amd64.go test/e2e/secret_test.go
```
I'm still trying to find out why the remote case isn't working though.

rhatdan:
LGTM

trusch:
@ashley-cui I rebased one last time. The last CI run was nearly green, except one integration test (ubuntu2021 rootless) which failed on network and volume management and others with ugly panics. I guess that the node was broken or something.

pkg/domain/entities/secrets.go
```diff
@@ -12,6 +12,7 @@ type SecretCreateReport struct {
 
 type SecretCreateOptions struct {
 	Driver string
+	Opts   map[string]string
```
ashley-cui, line 15: I think this should be consistent throughout
```suggestion
	DriverOpts   map[string]string
```

pkg/domain/infra/abi/secrets.go
```diff
@@ -17,15 +17,30 @@ func (ic *ContainerEngine) SecretCreate(ctx context.Context, name string, reader
 	if err != nil {
 		return nil, err
 	}
-	driverOptions := make(map[string]string)
 
+	// set defaults from config for the case they are not set by an upper layer
+	// (-> i.e. tests that talk directly to the api)
+	cfg, err := ic.Libpod.GetConfig()
```
ashley-cui, line 23: I still think defaults should be set only in `cmd/podman` and not here in the abi. Most commands have defaults set in `cmd/podman` and I'd like to keep that consistency and not read config files in the abi.

ashley-cui (commented):
A few more things, thanks for all your work!

pkg/domain/infra/abi/secrets.go
trusch: Before this PR (in the old line 23) we checked if the driver was not specified and then set the hard coded default "file" since this was the only valid driver when this code was written, so we already set a default here. This is needed, because in `tests/apiv2/50-secrets.at` the very first test sends down `POST secrets/create Name=mysecret Data=c2VjcmV0` and expects that the secret is created. If we would not set defaults in the handler, this test would fail because the secret manager wouldn't know which driver to use. Since the new source for the default secret driver is no longer hard-coded but given in the (default) config, we need to reference it here. Otherwise we need to make the "driver" field in the payload required and this would be a very breaking change to the API.

Also I found 8 other uses of `ic.Libpod.GetConfig()` in the `abi` package.

pkg/domain/entities/secrets.go
trusch: changed.

ashley-cui:
LGTM

pkg/api/handlers/libpod/secrets.go
```diff
@@ -16,9 +18,17 @@ func CreateSecret(w http.ResponseWriter, r *http.Request) {
 		runtime = r.Context().Value("runtime").(*libpod.Runtime)
 		decoder = r.Context().Value("decoder").(*schema.Decoder)
 	)
+
+	decoder.RegisterConverter(map[string]string{}, func(str string) reflect.Value {
+		res := make(map[string]string)
+		json.Unmarshal([]byte(str), &res)
+		return reflect.ValueOf(res)
+	})
```
jwhonce, line 26: @trusch NIT, can you add a comment that if this coventer code is needed elsewhere it should be moved to pkg/api/handlers/decoder.go. Thanks.

jwhonce (approved):
/lgtm

PR #10741, opened 2021-06-21
Do not use inotify for OCICNI

### LINKED ISSUE #10686: v3.2.1 (on Gentoo): Error: error configuring CNI network plugin: failed to create new watcher too many open files
rootless podman v3.2.1 fails on my Gentoo laptop, no matter what I run:
```console
$ podman info
Error: error configuring CNI network plugin: failed to create new watcher too many open files
```
It fails very early: not even `podman info` nor `podman system reset` get past this point. `strace` shows a smoking gun here:
```
newfstatat(AT_FDCWD, "/home/esm/.config/cni/net.d", {st_mode=S_IFDIR|0755, st_size=4096, ...}, 0) = 0
inotify_init1(IN_CLOEXEC)               = -1 EMFILE (Too many open files)
```
Per watercooler discussion 2021-06-15 the source of the error is in CNI plugins. The root cause is almost certainly something on my end, possibly my use of `syncthing` (which also periodically gripes with inotify-related errors). I'm not asking the podman team to spend much time on this, my goal is to file a placeholder to help others who run into this. (For now, the only solution is to revert to 3.1.2).

Luap99:
Podman does not need to watch the cni config directory. If a network is
not found in the cache, OCICNI will reload the networks anyway and thus
even podman system service should work as expected.

Also include a change to not mount a "new" /var by default in the
rootless cni ns, instead try to use /var/lib/cni first and then the
parent dir. This allows users to store cni configs under /var/... which
is the case for the CI compose test.

Fixes #10686

### DIFF
--- a/libpod/networking_linux.go
+++ b/libpod/networking_linux.go
@@ -46,6 +46,9 @@ const (
 
 	// rootlessCNINSName is the file name for the rootless network namespace bind mount
 	rootlessCNINSName = "rootless-cni-ns"
+
+	// persistentCNIDir is the directory where the CNI files are stored
+	persistentCNIDir = "/var/lib/cni"
 )
 
 // Get an OCICNI network config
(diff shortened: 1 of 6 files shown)


Luap99:
@edsantiago Please test if this works for you.

Luap99:
@mheon PTAL. This should be ready

libpod/networking_linux.go
```diff
@@ -150,14 +150,31 @@ func (r *RootlessCNI) Do(toRun func() error) error {
...
-		varDir := filepath.Join(r.dir, "var")
+		// cni plugins need access to /var/lib/cni and /run
+		varDir := ""
+		varTarget := "/var/lib/cni"
+		// we can only mount to a target dir which exists, check /var/lib/cni recursively
+		// while we could always use /var there are cases where a user might store the cni
+		// configs under /var/custom and this would break
+		for {
+			if _, err := os.Stat(varTarget); err == nil {
+				varDir = filepath.Join(r.dir, varTarget[1:])
```
mheon, line 161: Suggest that a `strings.TrimPrefix` might be a little better than that array slicing - probably a bit slower, but it's safer (guarantees we don't trim a non-slash character if this code ever changes)

libpod/networking_linux.go
```diff
@@ -150,14 +150,31 @@ func (r *RootlessCNI) Do(toRun func() error) error {
 			}
 		}
 
-		// cni plugins need access to /var and /run
-		runDir := filepath.Join(r.dir, "run")
-		varDir := filepath.Join(r.dir, "var")
+		// cni plugins need access to /var/lib/cni and /run
+		varDir := ""
+		varTarget := "/var/lib/cni"
```
rhatdan, line 155: Can you make this a constant, so that it can be used below.

libpod/networking_linux.go
```diff
@@ -385,7 +402,7 @@ func (r *Runtime) GetRootlessCNINetNs(new bool) (*RootlessCNI, error) {
 
 				// create cni directories to store files
 				// they will be bind mounted to the correct location in a extra mount ns
-				err = os.MkdirAll(filepath.Join(cniDir, "var"), 0700)
+				err = os.MkdirAll(filepath.Join(cniDir, "var/lib/cni"), 0700)
```
rhatdan, line 405: Need to use constant.

rhatdan (approved):

rhatdan:
LGTM

rhatdan:

Issue #10675, opened 2021-06-14
podman logs -f: misses final container output


edsantiago:
This one smells a little like #10003 but it's podman-local, not -remote.
```
$ podman run -d quay.io/libpod/testimage:20210427 sh -c 'for i in 1 2 3 4 5 6 8 10 12 13 14 15 16 20 21 22 23 24 25 26 64; do trap "echo got: $i" $i; done;echo READY;while ! test -e /stop; do sleep 0.05; done;echo DONE'
$ podman kill -s X $cid
[lots more podman-kills]
$ podman exec $cid touch /stop
...container fails to echo DONE...
```
(The above is not a real reproducer, just an overview of the problem)

## sys: podman kill - test signal handling in containers
* fedora-33 : sys podman fedora-33 rootless host
  * PR #10635
    * [06-11 11:14](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4808706966683648/html/sys-podman-fedora-33-rootless-host.log.html)
  * PR #10628
    * [06-10 11:16](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6007134157537280/html/sys-podman-fedora-33-rootless-host.log.html)
  * PR #10611
    * [06-09 12:59](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5129405514121216/html/sys-podman-fedora-33-rootless-host.log.html)
* fedora-34 : sys podman fedora-34 rootless host
  * PR #10654
    * [06-11 16:25](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6147039093325824/html/sys-podman-fedora-34-rootless-host.log.html)
    * [06-11 13:17](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6185071246835712/html/sys-podman-fedora-34-rootless-host.log.html)
  * PR #10634
    * [06-10 05:44](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6718132841611264/html/sys-podman-fedora-34-rootless-host.log.html)
  * PR #10611
    * [06-09 14:53](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5645402952171520/html/sys-podman-fedora-34-rootless-host.log.html)
[...]

PR #10742, opened 2021-06-21
 logs: k8s-file: restore poll sleep

vrothberg:
Commit 84b55ee attempted to fix a race waiting for the container
died event.  Previously, Podman slept for duration of the polling
frequence which I considerred to be a mistake.  As it turns out, I was
mistaken since the file logger will, in fact, NOT read until EOF and
then stop logging but stop logging immediately _after_ it woke up.

[NO TESTS NEEDED] as the race condition cannot be hit reliably.

Fixes: #10675

### DIFF
--- a/libpod/container_log.go
+++ b/libpod/container_log.go
@@ -74,7 +74,7 @@ func (c *Container) readFromLogFile(ctx context.Context, options *logs.LogOption
 			}
 			nll, err := logs.NewLogLine(line.Text)
 			if err != nil {
-				logrus.Error(err)
+				logrus.Errorf("Error getting new log line: %v", err)
 				continue
 			}
 			if nll.Partial() {
@@ -100,10 +100,10 @@ func (c *Container) readFromLogFile(ctx context.Context, options *logs.LogOption
 			// read the file until EOF.
 			tailError := t.StopAtEOF()
 			if tailError != nil && fmt.Sprintf("%v", tailError) != "tail: stop at eof" {
-				logrus.Error(tailError)
+				logrus.Errorf("Error stopping logger: %v", tailError)
 			}
-			if errors.Cause(err) != define.ErrNoSuchCtr {
-				logrus.Error(err)
+			if err != nil && errors.Cause(err) != define.ErrNoSuchCtr {
+				logrus.Errorf("Error getting container state: %v", err)
 			}
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


vrothberg:
Needs a backport to v3.2.

go.mod
```diff
@@ -42,6 +42,7 @@ require (
 	github.com/mattn/go-colorable v0.1.8 // indirect
 	github.com/moby/term v0.0.0-20201216013528-df9cb8a40635
 	github.com/mrunalp/fileutils v0.5.0
+	github.com/nxadm/tail v1.4.8
```
Luap99, line 45: This looks wrong, podman uses `github.com/hpcloud/tail`

libpod/container_log.go
```diff
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/libpod/events"
 	"github.com/containers/podman/v3/libpod/logs"
+	"github.com/nxadm/tail/watch"
```
Luap99, line 12: same here

vrothberg: Good catch!  Looks like my vim plugin chose the wrong dep.  Thanks!

libpod/container_log.go
```diff
@@ -93,17 +95,20 @@ func (c *Container) readFromLogFile(ctx context.Context, options *logs.LogOption
...
+		// getting its state, instruct the logger to read the file
+		// until EOF.
 		state, err := c.State()
 		if err != nil || state != define.ContainerStateRunning {
-			// If the container isn't running or if we encountered
-			// an error getting its state, instruct the logger to
-			// read the file until EOF.
+			// Make sure to wait at least for the poll duration
+			// before stopping the file logger (see #10675).
+			time.Sleep(watch.POLL_DURATION)
```
TomSweeneyRedHat, line 105: I'm always worried about putting a hard sleep into a function like this, but I'm not thinking of a better way atm.

vrothberg: Same here. I _really_ did not like it which is why I removed it previously.  I read the docs of the library we're using but they did not reflect reality :(

Luap99 (commented):
LGTM

TomSweeneyRedHat:
LGTM

PR #10743, opened 2021-06-21
[v3.2]  logs: k8s-file: restore poll sleep

vrothberg:
Commit 84b55ee attempted to fix a race waiting for the container
died event.  Previously, Podman slept for duration of the polling
frequency which I considered to be a mistake.  As it turns out, I was
mistaken since the file logger will, in fact, NOT read until EOF and
then stop logging but stop logging immediately _after_ it woke up.

[NO TESTS NEEDED] as the race condition cannot be hit reliably.

Fixes: #10675


vrothberg:
Wrong branch.

PR #10749, opened 2021-06-21
Add CNI rootless networking troubleshooting for v2.2.1

TomSweeneyRedHat:
A CNI container image is required for rootless networking
in V2.2.1 (RHEL 8.3.1) and earlier.  Add a note in the
troubleshooting guide with a pointer to the documenation
for that.

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -156,7 +156,7 @@ When rootless Podman attempts to execute a container on a non exec home director
 
 #### Symptom
 
-If you are running Podman or buildah on a home directory that is mounted noexec,
+If you are running Podman or Buildah on a home directory that is mounted noexec,
 then they will fail. With a message like:
 
 ```
@@ -726,3 +726,23 @@ And then re-add the connection (removing the old one if necessary):
 And now this should work:
 
 `podman-remote info`
+
+---
+### 28) Rootless CNI networking fails in RHEL with Podman v2.2.1 to v3.0.1.
+
+A failure is encountered when trying to use networking on a rootless
+container in Podman v2.2.1 through v3.0.1 on RHEL.  This error does not
+occur on other Linux Distributions.
+
+#### Symptom
+
+A rootless container is created using a CNI network, but the `podman run` command
+returns an error that an image must be built.
+
+#### Solution
+
+In order to use a CNI network in a rootless container on RHEL,
+an Infra container image for CNI-in-slirp4netns must be created.  The
+instructions for building the Infra container image can be found for
+v2.2.1 [here](https://github.com/containers/podman/tree/v2.2.1-rhel/contrib/rootless-cni-infra),
+and for v3.0.1 [here](https://github.com/containers/podman/tree/v3.0.1-rhel/contrib/rootless-cni-infra).


troubleshooting.md
```diff
@@ -726,3 +726,21 @@ And then re-add the connection (removing the old one if necessary):
 And now this should work:
 
 `podman-remote info`
+
+---
+### 28) CNI networking fails in Podman v2.2.1 and earlier.
```
mheon, line 731: This is only RHEL. Anything that is not RHEL will pull an image automatically to accomplish this. Also, version number is wrong - I believe 2.2.1 was the earliest that offered rootless CNI functionality, and 3.1 is the last that required the image, with 3.2 moving to the network namespace approach that does not require external dependencies (beyond CNI itself)

troubleshooting.md
```diff
@@ -726,3 +726,21 @@ And then re-add the connection (removing the old one if necessary):
...
+#### Symptom
+
+Two rootless containers are created on the host but are not able to communicate
+with each other via their virtual network devices.
+
+#### Solution
+
+In order to use a CNI network in a rootless container in Podman v2.2.1 and earlier,
+an Infra container image for CNI-in-slirp4netns must be created.  The
+instructions for building the Infra container image can be found [here](https://github.com/containers/podman/tree/v2.2.1-rhel/contrib/rootless-cni-infra).
```
mheon, line 746: This is only applicable to v2.2.1. The other version we shipped on RHEL with this functionality, v3.0.1, requires a newer version of the image from https://github.com/containers/podman/tree/v3.0.1-rhel/contrib/rootless-cni-infra

Generally speaking, it's a hard requirement to match the image version to the Podman version, but since non-RHEL Podman can only pull the image and cannot have it manually built, we only need to mention these two.

mheon:
I question whether this is necessary, since the Podman command will, on requiring this image, print an error message containing the same link you're pointing to here.

TomSweeneyRedHat:
I've repushed with I think the changes that you wanted @mheon.  Please let me know if further tweaking is necessary.  We've had a jira card hanging around for a while for this.  Beyond that, I think it would be good to add to the guide, but can be talked out of it easily and just delete the card.

troubleshooting.md
```diff
@@ -726,3 +726,23 @@ And then re-add the connection (removing the old one if necessary):
...
+### 28) CNI networking fails in RHEL with Podman v2.2.1 to v3.0.1.
+
+A failure is encountered when trying to use networking on a rootless
+container in Podman v2.2.1 through v3.0.1 on RHEL.  This error does not
+occur on other Linux Distributions.
+
+#### Symptom
+
+Two rootless containers are created on the host but are not able to communicate
+with each other via their virtual network devices.
```
mheon, line 740: I would suggest "A rootless container is created using a CNI network, but the `podman run` command returns an error that an image must be built"

TomSweeneyRedHat: I've made your line the headliner in the symptom section, but said "Additionally..." and kept the old line as it was part of the original problem reports.

troubleshooting.md
```diff
@@ -726,3 +726,25 @@ And then re-add the connection (removing the old one if necessary):
 And now this should work:
 
 `podman-remote info`
+
+---
+### 28) CNI networking fails in RHEL with Podman v2.2.1 to v3.0.1.
```
mheon, line 731: "Rootless CNI Networking fails"

troubleshooting.md
mheon: This won't cause a failure when two containers are already created - this completely blocks the creation of containers in a CNI network until the image is built.

TomSweeneyRedHat:
Updated.  "Once more unto the breach!".

Issue #10579, opened 2021-06-07
Docker API incompatibility for "volume prune": "until" is an invalid volume filter


w4tsn:
/kind bug

**Description**

The docker API supports the `until` filter in volume prunes. This is not stated in the official documentation, interestingly enough, but I'm using this flag on the docker cli client without problems. E.g. `until=24h`.

**Steps to reproduce the issue:**

Run `podman volume prune --filter until="24h"

**Describe the results you received:**

Got an error message from the "docker" API socket

```
An unexpected docker error occurred: 500 Server Error: Internal Server Error (\"failed to parse filters for until=24h: \"until\" is an invalid volume filter\")
```

**Describe the results you expected:**

Compatibility of the "until" filter for volume prune, as it is supported by docker - if wider docker API compatibility is desired here.

**Output of `podman version`:**

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16.3
Built:        Wed May 12 21:27:59 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 4
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: backend05.othermo.de
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.12.8-300.fc34.x86_64
  linkmode: dynamic
  memFree: 15457230848
  memTotal: 16779862016
  ociRuntime:
    name: crun
    package: crun-0.19.1-3.fc34.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.19.1
      commit: 1535fedf0b83fb898d449f9680000f729ba719f5
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #10756, opened 2021-06-22
Add support for volume prune until filter to http api

jmguzik:
As stated in #10579 docker silently implements until filter for volume prune.
This commit adds initial support to the HTTP API, both libpod and compat.
It enables further work on that issue, such as adding cli support in the future.

Relates to #10579


rhatdan:
LGTM
/approve
@jwhonce PTAL

mheon:
LGTM

PR #10765, opened 2021-06-23
Create /etc/mtab with the correct ownership

umohnani8:
Create the /etc and /etc/mtab directories with the
correct ownership based on what the UID and GID is
for the container. This was causing issue when starting
the infra container with userns as the /etc directory
wasn't being created with the correct ownership.


umohnani8:
@rhatdan @mheon PTAL
@giuseppe thanks for helping find this!

rhatdan:
/approve
LGTM

giuseppe (approved):
LGTM

PR #10772, opened 2021-06-24
cp: do not allow dir->file copying

vrothberg:
Fix a bug in `podman-cp` to forbid copying directories to files.
Previously, the directory was copied to the parent directory of the file
which is wrong.


vrothberg:
A late one for 3.2. 

Luap99 (commented):
LGTM

PR #10788, opened 2021-06-26
support pulling multiple images sequentially in a single podman pull 

### LINKED ISSUE #10726: Podman should support pulling multiple images in a single command

/kind feature

**Description**
Currently podman pull command accepts only one argument and it would be really nice if it could support pulling images in a single command.

```
[mehul@fedora bin]$ ./podman pull alpine busybox
Error: accepts 1 arg(s), received 2
```
If someone wants to pull multiple images and if they are not familiar with bash scripting then they might have to do it manually by calling the command over and over. Instead it would be nice to have something like:
```
podman pull alpine busybox fedora ubuntu
```

```
[mehul@fedora bin]$ ./podman --version
podman version 3.3.0-dev
```
I think this could be a useful feature, but I am not completely sure about it's use case except I felt that it was redundant to type the command again when I was working on something.  I can try to work on this if this is agreed upon! 

Thank you

infiniteregrets:
Closes #10726 
Adds support for pulling multiple images sequentially. 

Signed-off-by: Mehul Arora <aroram18@mcmaster.ca>

### DIFF
--- a/cmd/podman/images/pull.go
+++ b/cmd/podman/images/pull.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v3/cmd/podman/common"
 	"github.com/containers/podman/v3/cmd/podman/registry"
+	"github.com/containers/podman/v3/cmd/podman/utils"
 	"github.com/containers/podman/v3/pkg/domain/entities"
 	"github.com/containers/podman/v3/pkg/util"
 	"github.com/pkg/errors"
@@ -32,8 +33,8 @@ var (
 
 	// Command: podman pull
 	pullCmd = &cobra.Command{
-		Use:               "pull [options] IMAGE",
-		Args:              cobra.ExactArgs(1),
+		Use:               "pull [options] IMAGE [IMAGE...]",
+		Args:              cobra.MinimumNArgs(1),
 		Short:             "Pull an image from a registry",
 		Long:              pullDescription,
 		RunE:              imagePull,
@@ -154,17 +155,16 @@ func imagePull(cmd *cobra.Command, args []string) error {
 	}
 	// Let's do all the remaining Yoga in the API to prevent us from
 	// scattering logic across (too) many parts of the code.
-	pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), args[0], pullOptions.ImagePullOptions)
-	if err != nil {
-		return err
+	var errs utils.OutputErrors
+	for _, arg := range args {
+		pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		for _, img := range pullReport.Images {
+			fmt.Println(img)
+		}
 	}
-
-	if len(pullReport.Images) > 1 {
-		fmt.Println("Pulled Images:")
-	}
-	for _, img := range pullReport.Images {
-		fmt.Println(img)
-	}
-
-	return nil
+	return errs.PrintErrors()
 }
--- a/test/e2e/pull_test.go
+++ b/test/e2e/pull_test.go
@@ -35,6 +35,23 @@ var _ = Describe("Podman pull", func() {
 
 	})
 
+	It("podman pull multiple images with/without tag/digest", func() {
+		session := podmanTest.Podman([]string{"pull", "busybox:musl", "alpine", "alpine:latest", "quay.io/libpod/cirros", "quay.io/libpod/testdigest_v2s2@sha256:755f4d90b3716e2bf57060d249e2cd61c9ac089b1233465c5c2cb2d7ee550fdb"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+
+		session = podmanTest.Podman([]string{"pull", "busybox:latest", "docker.io/library/ibetthisdoesnotexistfr:random", "alpine"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(125))
+		expectedError := "Error initializing source docker://ibetthisdoesnotexistfr:random"
+		found, _ := session.ErrorGrepString(expectedError)
+		Expect(found).To(Equal(true))
+
+		session = podmanTest.Podman([]string{"rmi", "busybox", "alpine", "testdigest_v2s2", "quay.io/libpod/cirros"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
+	})
+
 	It("podman pull from docker a not existing image", func() {
 		session := podmanTest.Podman([]string{"pull", "ibetthisdoesntexistthere:foo"})
 		session.WaitWithDefaultTimeout()
@@ -385,7 +402,6 @@ var _ = Describe("Podman pull", func() {
 		session := podmanTest.Podman([]string{"pull", "--all-tags", "k8s.gcr.io/pause"})
 		session.WaitWithDefaultTimeout()
 		Expect(session.ExitCode()).To(Equal(0))
-		Expect(session.LineInOutputStartsWith("Pulled Images:")).To(BeTrue())
 
 		session = podmanTest.Podman([]string{"images"})
 		session.WaitWithDefaultTimeout()
--- a/docs/source/markdown/podman-pull.1.md
+++ b/docs/source/markdown/podman-pull.1.md
@@ -4,24 +4,20 @@
 podman\-pull - Pull an image from a registry
 
 ## SYNOPSIS
-**podman pull** [*options*] *source*
+**podman pull** [*options*] *source* [*source*...]
 
-**podman image pull** [*options*] *source*
+**podman image pull** [*options*] *source* [*source*...]
 
 **podman pull** [*options*] [*transport*]*name*[:*tag*|@*digest*]
 
 **podman image pull** [*options*] [*transport*]*name*[:*tag*|@*digest*]
 
 ## DESCRIPTION
-Copies an image from a registry onto the local machine. The **podman pull** command pulls an
-image.  If the image reference in the command line argument does not contain a registry, it is referred to as a`short-name` reference. If the image is a 'short-name' reference, Podman will prompt the user for the specific container registry to pull the image from, if an alias for the short-name has not been specified in the short-name-aliases.conf.  If an image tag is not specified, **podman pull** defaults to the image with the **latest** tag (if it exists) and pulls it. After the image is pulled, podman will print the full image ID.  **podman pull** can also pull an image using its digest **podman pull** *image*@*digest*. **podman pull** can be used to pull images from archives and local storage using different transports.
-
-## Image storage
-Images are stored in local image storage.
+podman pull copies an image from a registry onto the local machine. The command can pull one or more images.  If the image reference in the command line argument does not contain a registry, it is referred to as a`short-name` reference. If the image is a 'short-name' reference, Podman will prompt the user for the specific container registry to pull the image from, if an alias for the short-name has not been specified in the `short-name-aliases.conf`.  If an image tag is not specified, **podman pull** defaults to the image with the **latest** tag (if it exists) and pulls it. After the image is pulled, podman will print the full image ID.  **podman pull** can also pull images using a digest **podman pull** *image*@*digest* and can also be used to pull images from archives and local storage using different transports.
+*IMPORTANT: Images are stored in local image storage.*
 
 ## SOURCE
-
- SOURCE is the location from the container image is pulled from. It supports all transports from `containers-transports(5)`. If no transport is specified, the input is subject to short-name resolution and the `docker` (i.e., container registry) transport is used.  For remote clients, `docker` is the only supported transport.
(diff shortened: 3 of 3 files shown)


infiniteregrets:
@rhatdan PTAL

cmd/podman/images/pull.go
```diff
@@ -154,17 +159,28 @@ func imagePull(cmd *cobra.Command, args []string) error {
...
+			fmt.Print("\n")
+			_, err = registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
+			if err != nil {
+				return err
+			}
+		}
 	}
 
-	if len(pullReport.Images) > 1 {
+	/*if len(pullReport.Images) > 1 {
```
mtrmac, line 178: This needs finishing.

mtrmac (commented):
> If an image exists, it is skipped. Not sure why this wasn't present before but podman tries to pull an existing image:

If the `:latest` tag on a repo changes, why _wouldn’t_ `podman pull` verify that it has the updated version?

cmd/podman/images/pull.go
infiniteregrets: > If the :latest tag on a repo changes, why wouldn’t podman pull verify that it has the updated version?

My bad, I didn't realize. Will fix that

infiniteregrets: > This needs finishing.

I forgot to ask, I didn't understand what the portion of code I commented is for?

cmd/podman/images/pull.go
```diff
@@ -32,8 +32,13 @@ var (
 
 	// Command: podman pull
 	pullCmd = &cobra.Command{
-		Use:               "pull [options] IMAGE",
-		Args:              cobra.ExactArgs(1),
+		Use: "pull [options] IMAGE",
+		Args: func(cmd *cobra.Command, args []string) error {
```
vrothberg, line 36: Can you use `Args: cobra.MinimumNArgs(1)` instead?  This will do the same but be consistent with the error messages of other commands.

infiniteregrets: Oh right, thanks!

cmd/podman/images/pull.go
jwhonce: @infiniteregrets For each pull, you need to print out the image ids that were pulled.

jwhonce (changes requested):

cmd/podman/images/pull.go
infiniteregrets: @jwhonce I was a bit confused because, "Pulled Images:" was never getting printed before only the signature was. I have made the changes, please let me know if this is okay now?

cmd/podman/images/pull.go
```diff
@@ -154,15 +154,20 @@ func imagePull(cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		return err
+	pulledImages := make([]string, len(args))
+	for i, arg := range args {
+		pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
+		if err != nil {
+			return err
+		}
+		pulledImages[i] = pullReport.Images[0]
+		fmt.Print("\n")
```
vrothberg, line 164: I assume the newline was added to have a blank line between each image pull but it will break `-q`.  I suggest to just remove the line.

cmd/podman/images/pull.go
```diff
@@ -154,15 +154,20 @@ func imagePull(cmd *cobra.Command, args []string) error {
...
-	pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), args[0], pullOptions.ImagePullOptions)
-	if err != nil {
-		return err
+	pulledImages := make([]string, len(args))
+	for i, arg := range args {
+		pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
+		if err != nil {
+			return err
+		}
+		pulledImages[i] = pullReport.Images[0]
```
vrothberg, line 163: `pullReport.Images` can have more than one image (see `podman pull -a`) so we should do as follows:
```go
var pulledImages []string
for i, arg := .... {
   ....
   pulledImages = append(pulledImages, pullReport.Images...)
}
```
This will append all pulled images.

```

File: test/e2e/pull_test.go [hunk 5]
```diff
@@ -35,6 +35,24 @@ var _ = Describe("Podman pull", func() {
 
 	})
 
+	It("podman pull multiple images with tag", func() {
+		session := podmanTest.Podman([]string{"pull", "busybox:latest", "alpine:3"})
+		session.WaitWithDefaultTimeout()
+		Expect(session.ExitCode()).To(Equal(0))
```
vrothberg, line 41: We should also check the output string and make sure that the images are correctly reported as pulled.

vrothberg: Look for `GrepString(...)` in the other tests.

cmd/podman/images/pull.go
infiniteregrets: Sure Ill remove it, but can we not do something like:
```go
if !pullOptions.Quiet {
    fmt.Print("\n")
}
```

vrothberg: The output of `pull --all foo` doesn't have these blank lines, so it would not be consistent.  But I do not feel strongly.  If you like the blank lines, feel free to add them.

cmd/podman/images/pull.go
```diff
@@ -154,15 +154,22 @@ func imagePull(cmd *cobra.Command, args []string) error {
...
-		return err
+	pulledImages := make([]string, 0)
+	for _, arg := range args {
+		pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
+		if err != nil {
+			return err
+		}
+		pulledImages = append(pulledImages, pullReport.Images...)
+		if !pullOptions.Quiet {
+			fmt.Print("\n")
```
rhatdan, line 165: Is this correct?  If I pull a bunch of images with quiet, doesn't it print the ImageID for each image I pull? Why doesn't that need a \n for more then one?

rhatdan: I would expect the output to be
```
$ podman pull ubi8-micro fedora
Trying to pull registry.access.redhat.com/ubi8-micro:latest...
Copying blob c07a48b26be2 skipped: already exists  
Copying blob 4f4fb700ef54 [--------------------------------------] 0.0b / 0.0b
Copying config f390b26f6a done  
Storing signatures
f390b26f6a00561dbacfc5991a39a6845ff7968f8d18a319b68cc9c0f97dea73
Trying to pull registry.fedoraproject.org/fedora:latest...
Copying blob ca594071dfff skipped: already exists  
Copying config abec9a7a7d done  
Storing signatures
abec9a7a7dc640768e6dc51b6a5728e470411615c62e9ff46215206bde816772
```
Rather then

```
$ podman pull ubi8-micro fedora
Trying to pull registry.access.redhat.com/ubi8-micro:latest...
Copying blob c07a48b26be2 skipped: already exists  
Copying blob 4f4fb700ef54 [--------------------------------------] 0.0b / 0.0b
Copying config f390b26f6a done  
Storing signatures
Trying to pull registry.fedoraproject.org/fedora:latest...
Copying blob ca594071dfff skipped: already exists  
Copying config abec9a7a7d done  
Storing signatures
Pulled Images:
f390b26f6a00561dbacfc5991a39a6845ff7968f8d18a319b68cc9c0f97dea73
abec9a7a7dc640768e6dc51b6a5728e470411615c62e9ff46215206bde816772
```

cmd/podman/images/pull.go
```diff
@@ -154,15 +154,22 @@ func imagePull(cmd *cobra.Command, args []string) error {
...
 	// Let's do all the remaining Yoga in the API to prevent us from
 	// scattering logic across (too) many parts of the code.
-	pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), args[0], pullOptions.ImagePullOptions)
-	if err != nil {
-		return err
+	pulledImages := make([]string, 0)
+	for _, arg := range args {
+		pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
+		if err != nil {
+			return err
```
rhatdan, line 161: When we try to do multiple operations at once, we complete the tasks, and print the errors.  Look how podman stop works.
If I say

podman pull alpine bogus fedora ubi

I would want alpine fedora and ubi to be pulled, and then the command fail with bogus does not exist.

rhatdan (approved):

cmd/podman/images/pull.go
mtrmac: I don’t feel strongly about this, but the second one seems _much_ easier to copy&paste. (Of course users should not copy&paste non-API command output, but then why are the IDs printed at all?)

rhatdan: Because docker printed them, and they make sense when you do it with a -q option on one image.

cmd/podman/images/pull.go
infiniteregrets: I think that is possible because the ContainerStop function accepts an array of strings and here it's just a single string. And passing an array makes it possible to get an accumulated response but in this case we would have to change the code for Pull in various places. 
But if you think that this is possible, then this PR is editable by maintainers. So please make the change. Thanks!

cmd/podman/images/pull.go
infiniteregrets:
```
❯ docker pull -q busybox
docker.io/library/busybox:latest
```
Docker doesnt print the IDs for -q flag. 
So let me know what the desired behavior is and I will make the changes accordingly

cmd/podman/images/pull.go
rhatdan: No you just need something like:
```
var errs utils.OutputErrors
for _, arg := range args {
		pullReport, err := registry.ImageEngine().Pull(registry.GetContext(), arg, pullOptions.ImagePullOptions)
		if err != nil {
                        errs = append(errs, err)
			continue
              }
    ...
    }
    return errs.PrintErrors()
}

File: cmd/podman/images/pull.go (hunk 6 above)
Comment (rhatdan) [reply cmd/podman/images/pull.go:165]: I would prefer to print the image name on success, but I think we have set the precedent of printing the ID.  WDYT @vrothberg @mheon ?

Comment (vrothberg) [reply cmd/podman/images/pull.go:165]: I think we should do what @rhatdan suggested in https://github.com/containers/podman/pull/10788#discussion_r660486435.

File: cmd/podman/images/pull.go (hunk 7 above)
Comment (infiniteregrets) [reply cmd/podman/images/pull.go]: I have updated the code, but why arent the checks running?

Comment (vrothberg) [reply cmd/podman/images/pull.go]: Yes, it looks like a `master` -> `main` issue. Hm ... maybe try closing and re-opening. Then repush?

Comment (infiniteregrets) [reply cmd/podman/images/pull.go]: @vrothberg I fixed that, but some tests are failing. Am I missing something?

Comment (vrothberg) [reply cmd/podman/images/pull.go]: Tests are currently running.  I'll wait for them to finish and will have a look later.

Comment (infiniteregrets) [reply cmd/podman/images/pull.go]: Looks like it is unable to resolve the tag I am using for the alpine image
[Error: Error initializing source docker://alpine:edge: Error reading manifest edge in quay.io/libpod/alpine: manifest unknown: manifest unknown](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4715965364568064/html/int-podman-fedora-34-rootless-host.log.html)
But this works just fine locally for me?

File: test/e2e/pull_test.go [hunk 8]
```diff
@@ -35,6 +35,42 @@ var _ = Describe("Podman pull", func() {
 
 	})
 
+	It("podman pull multiple images with tag", func() {
+		session := podmanTest.Podman([]string{"pull", "busybox:latest", "alpine:edge"})
```
vrothberg, line 39: `alpine:edge` does not exist on our mirror on quay.  Please use `alpine`.  Then you can also remove the `"podman pull multiple images without tag"` test.

cmd/podman/images/pull.go
```diff
@@ -154,17 +155,19 @@ func imagePull(cmd *cobra.Command, args []string) error {
...
+			errs = append(errs, err)
+			continue
+		}
+		for _, img := range pullReport.Images {
+			fmt.Println(img)
+		}
 	}
-	for _, img := range pullReport.Images {
-		fmt.Println(img)
+	if len(errs) > 0 {
```
rhatdan, line 169: No need for this check, if errs.PrintErrors will return nil if there are no errors.
LGTM

rhatdan:
Nice work @infiniteregrets.  Thanks
LGTM

vrothberg (commented):
/lgtm
/hold

Thanks!

mheon:
Test failures seem consistent

cmd/podman/images/pull.go
```diff
@@ -33,7 +34,7 @@ var (
 	// Command: podman pull
 	pullCmd = &cobra.Command{
 		Use:               "pull [options] IMAGE",
```
Luap99, line 36: You have to add this, this will fix the failing test.
Also please update the synopsis in the man page.
```suggestion
		Use:               "pull [options] IMAGE [IMAGE...]",
```

infiniteregrets:
hi @Luap99 let me know if there is anything to be changed in docs

mheon:
LGTM

vrothberg (approved):
/lgtm

Thank you!

Issue #10755, opened 2021-06-22
make `podman system reset` handle external containers


dustymabe:
/kind feature

**Description**

If I have "external" containers (like ones created by `buildah` or `podman build` then `podman system reset` will error:

```
[core@pcengines ~]$ podman system reset

WARNING! This will remove:
        - all containers
        - all pods
        - all images
        - all build cache
Are you sure you want to continue? [y/N] y
ERRO[0002] 31 errors occurred:
	* Image used by 769d78b7025b2b92072d2875f92474474dd36284a1d1cdb0480365b7ec534cc2: image is in use by a container
	* Image used by 4f834bf4657dbb9652a3e3be17b8595f32b5af28c2e2ebd4dcceeafc779449b9: image is in use by a container
	* Image used by 381b0e9504ed095875ac2093e67c14ce37edeb2881ecea0c4e49edb06a8f276b: image is in use by a container
	* Image used by 9fc744b584ce8ffcb26f0d593ecd1201d3fb17bbdd534607af7f56830800590b: image is in use by a container
	* Image used by ed0e6ed6ae5f0edcf6f7843f2903e26595a7efb64ddd31cd07617c05522c3bc5: image is in use by a container
	* Image used by 30748298264d7ccccf43ec43b875ce2189111dd47cd0947559f1ee0a7c7f9900: image is in use by a container
	* Image used by c949a4f5154931d24803966885bb2b01a8648470e0620876fb0e5dfc7de050a5: image is in use by a container
	* Image used by aa4bc49440ea59a757c04e22d410a62adf4ff062fe4a79e450067784cfdf23e3: image is in use by a container
	* Image used by d05b12db0a97d74a27558cb596091cd1547dc6303b718e97c9c2a8f87e0a1cc3: image is in use by a container
	* Image used by 00e39cc9be1d49e5e62313b2f3f62dfaa026303decb5fc9f2ae69004e597657d: image is in use by a container
	* Image used by ca865c272f46497d98f06706755978e7e822ce8263d899b604b11101b6cbda7c: image is in use by a container
	* Image used by 932c6ea5f622e8812ea3b471687990f5517aa181a6e42570f012fc25cbbc4b0d: image is in use by a container
	* Image used by ab9dc123c9c192b5b775887a252c6d7711ab0b50c74dacfaf50c64f1a458bb27: image is in use by a container
[...]

PR #10789, opened 2021-06-26
reset: remove external containers on podman system reset

flouthoc:
Following PR allows `podman system reset` to remove all the external containers with a **safety prompt** when `--force` is not provided. This makes sure that `podman system reset` performs a complete nuke of all the containers.

Attempts a feature patch for issue: https://github.com/containers/podman/issues/10755

Example 
```bash
flouthoc@system:~/src/stable/glibc$ podman system reset
WARNING! This will remove:
        - all containers
        - all pods
        - all images
        - all build cache
WARNING! The following external containers will be purged:
	- dbfc83909cbb (busybox-working-container)
	- c37fb04008bd (busybox-working-container-1)
Are you sure you want to continue? [y/N] y
```
* `--force` silently removes everything without any prompt.
Example 2
```bash
flouthoc@system:~/src/stable/glibc$ podman system reset --force
```

### DIFF
--- a/cmd/podman/system/reset.go
+++ b/cmd/podman/system/reset.go
@@ -45,16 +45,29 @@ func init() {
 }
 
 func reset(cmd *cobra.Command, args []string) {
+	// Get all the external containers in use
+	listCtn, _ := registry.ContainerEngine().ContainerListExternal(registry.Context())
+	listCtnIds := make([]string, 0, len(listCtn))
+	for _, externalCtn := range listCtn {
+		listCtnIds = append(listCtnIds, externalCtn.ID)
+	}
 	// Prompt for confirmation if --force is not set
 	if !forceFlag {
 		reader := bufio.NewReader(os.Stdin)
-		fmt.Print(`
+		fmt.Println(`
 WARNING! This will remove:
         - all containers
         - all pods
         - all images
-        - all build cache
-Are you sure you want to continue? [y/N] `)
+        - all build cache`)
+		if len(listCtn) > 0 {
+			fmt.Println(`WARNING! The following external containers will be purged:`)
+			// print first 12 characters of ID and first configured name alias
+			for _, externalCtn := range listCtn {
+				fmt.Printf("	- %s (%s)\n", externalCtn.ID[0:12], externalCtn.Names[0])
+			}
+		}
+		fmt.Print(`Are you sure you want to continue? [y/N] `)
 		answer, err := reader.ReadString('\n')
 		if err != nil {
 			logrus.Error(err)
@@ -65,6 +78,8 @@ Are you sure you want to continue? [y/N] `)
 		}
 	}
 
+	// Purge all the external containers with storage
+	registry.ContainerEngine().ContainerRm(registry.Context(), listCtnIds, entities.RmOptions{Force: true, All: true, Ignore: true, Volumes: true})
 	// Shutdown all running engines, `reset` will hijack repository
 	registry.ContainerEngine().Shutdown(registry.Context())
 	registry.ImageEngine().Shutdown(registry.Context())
--- a/pkg/domain/entities/engine_container.go
+++ b/pkg/domain/entities/engine_container.go
@@ -32,6 +32,7 @@ type ContainerEngine interface {
 	ContainerInspect(ctx context.Context, namesOrIds []string, options InspectOptions) ([]*ContainerInspectReport, []error, error)
 	ContainerKill(ctx context.Context, namesOrIds []string, options KillOptions) ([]*KillReport, error)
 	ContainerList(ctx context.Context, options ContainerListOptions) ([]ListContainer, error)
+	ContainerListExternal(ctx context.Context) ([]ListContainer, error)
 	ContainerLogs(ctx context.Context, containers []string, options ContainerLogsOptions) error
 	ContainerMount(ctx context.Context, nameOrIDs []string, options ContainerMountOptions) ([]*ContainerMountReport, error)
 	ContainerPause(ctx context.Context, namesOrIds []string, options PauseUnPauseOptions) ([]*PauseUnpauseReport, error)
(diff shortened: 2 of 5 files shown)


flouthoc:
Added [NO TESTS NEEDED] for a check will add relevant tests after initial review if needed.

flouthoc:
I think failing tests are flake. I'll re-kick CI

cmd/podman/system/reset.go
```diff
@@ -65,6 +65,34 @@ Are you sure you want to continue? [y/N] `)
 		}
 	}
 
+	// Get all the external containers in use
+	listCtn, _ := registry.ContainerEngine().ContainerListExternal(registry.Context())
+	listCtnIds := make([]string, 0, len(listCtn))
+	for _, externalCtn := range listCtn {
+		listCtnIds = append(listCtnIds, externalCtn.ID)
```
rhatdan, line 72: Please specify Names or both.  These IDs mean very little to Human beings, where as they might mean more if they saw:

alpine-working-container

vrothberg:
I have a weak preference to have only _one_ prompt. In fact, to keep things as they are but in case there are external containers to display them as follows:
```
flouthoc@system:~/src/stable/glibc$ podman system reset

WARNING! This will remove:
        - all containers (5 external containers)
        - all pods
        - all images
        - all build cache
Are you sure you want to continue? [y/N] y
```

IMO, `system reset` is a "nuke all" but we already remove external containers with `rmi -f $image`.

flouthoc:
@vrothberg Sure sounds good to me. Just a slight doubt how would human-users know about which external containers are getting purged. Its fine if it is generally acceptable that `podman system reset` will nuke everything.

cmd/podman/system/reset.go
flouthoc: @rhatdan Added both of them but this might change as @vrothberg has a better suggestion as being discussed below.
```bash 
WARNING! Following External Containers will be purged
CONTAINER ID  NAME
c7d50cbfff74  busybox-working-container
```

rhatdan:
I agree only give one prompt, but output the external containers as well.

mheon:
/approve
Concur that a single prompt would be fine - we should just list the external containers before the existing "proceed" prompt

cmd/podman/system/reset.go
```diff
@@ -65,6 +65,37 @@ Are you sure you want to continue? [y/N] `)
...
+	}
+	if len(listCtn) > 0 {
+		if !forceFlag {
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Println(`
+WARNING! Following External Containers will be purged`)
```
TomSweeneyRedHat, line 78:
```suggestion
WARNING! The Following External Containers will be purged`)
```

TomSweeneyRedHat:
I'm not sure why the release test is failing with `dev` shouldn't be in the version for the main branch....  @edsantiago ?

flouthoc:
@TomSweeneyRedHat Thanks resolved your feedback on latest commit.
@vrothberg @rhatdan @mheon Added a single prompt with list of external containers if any as suggested.

Example
```bash
flouthoc@system:~/src/stable/glibc$ podman system reset
WARNING! This will remove:
        - all containers
        - all pods
        - all images
        - all build cache
WARNING! Following External Containers will be purged
CONTAINER ID  NAME
c7d50cbfff74  busybox-working-container
3ab9923a9f8b  busybox-working-container-1
bd371cdddd29  busybox-working-container-2
Are you sure you want to continue? [y/N] y
```

cmd/podman/system/reset.go
```diff
@@ -45,16 +45,31 @@ func init() {
...
+		fmt.Println(`
 WARNING! This will remove:
         - all containers
         - all pods
         - all images
-        - all build cache
-Are you sure you want to continue? [y/N] `)
+        - all build cache`)
+		if len(listCtn) > 0 {
+			fmt.Println(`WARNING! The Following External Containers will be purged`)
```
vrothberg, line 64:
```suggestion
			fmt.Println(`WARNING! The following external containers will be purged:`)
```

vrothberg:
Thanks, @flouthoc!

Visually, I would find it more appealing if the the two lists would have the same format. Maybe something as below?
```
flouthoc@system:~/src/stable/glibc$ podman system reset
WARNING! This will remove:
        - all containers
        - all pods
        - all images
        - all build cache
WARNING! Following External Containers will be purged
        - c7d50cbfff74  (busybox-working-container)
        - 3ab9923a9f8b  (busybox-working-container-1)
        - bd371cdddd29  (busybox-working-container-2)
Are you sure you want to continue? [y/N] y
```

Let's wait how the others think.

edsantiago:
> I'm not sure why the release test is failing with `dev` shouldn't be in the version for the main branch.... @edsantiago ?

@TomSweeneyRedHat that happens when someone manually presses the "Trigger" button for that test, usually because they see that it didn't run and they believe it should run. It shouldn't. It's an optional test that doesn't actually run and shouldn't run. If you've read HHGTTG, it's the equivalent of a big button that, when pressed, lights up and says "do not press this button again".

flouthoc:
@edsantiago Oh actually i triggered that test by mistake. 😢 😭

rhatdan:
@flouthoc Make @vrothberg suggestion and then we can merge.

flouthoc:
@vrothberg Updated with suggested changes in lastest commit
```bash
flouthoc@system:~/src/stable/glibc$ podman system reset
WARNING! This will remove:
        - all containers
        - all pods
        - all images
        - all build cache
WARNING! The following external containers will be purged:
	- dbfc83909cbb (busybox-working-container)
	- c37fb04008bd (busybox-working-container-1)
Are you sure you want to continue? [y/N] y
```

vrothberg (approved):
LGTM

PR #10804, opened 2021-06-28
Implement --archive flag for podman cp

### LINKED ISSUE #10801: podman cp chowns files
Sub-comman `podman cp` doesn't preserve uid/git from tar. It deliberately chowns everything to uid/gid the container is running under at the moment. `docker` however preserve uid/gid. Since `podman` is drop in replacement we should do that too.
If not for the `cp` CLI sub-comman then an least for docker APIv2.

matejvasek:
Implement --archive flag for `podman cp`

This PR implements for direction when a container is a destination, not a source.

resolves #10801

### DIFF
--- a/cmd/podman/containers/cp.go
+++ b/cmd/podman/containers/cp.go
@@ -28,13 +28,13 @@ var (
   You can copy from the container's file system to the local machine or the reverse, from the local filesystem to the container. If "-" is specified for either the SRC_PATH or DEST_PATH, you can also stream a tar archive from STDIN or to STDOUT. The CONTAINER can be a running or stopped container. The SRC_PATH or DEST_PATH can be a file or a directory.
 `
 	cpCommand = &cobra.Command{
-		Use:               "cp [CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
+		Use:               "cp [options] [CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
 		Short:             "Copy files/folders between a container and the local filesystem",
 		Long:              cpDescription,
 		Args:              cobra.ExactArgs(2),
 		RunE:              cp,
 		ValidArgsFunction: common.AutocompleteCpCommand,
-		Example:           "podman cp [CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
+		Example:           "podman cp [options] [CONTAINER:]SRC_PATH [CONTAINER:]DEST_PATH",
 	}
 
 	containerCpCommand = &cobra.Command{
@@ -50,12 +50,14 @@ var (
 
 var (
 	cpOpts entities.ContainerCpOptions
+	chown  bool
 )
 
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Deprecated...")
 	flags.BoolVar(&cpOpts.Pause, "pause", true, "Deprecated")
+	flags.BoolVarP(&chown, "archive", "a", true, `Chown copied files to the primary uid/gid of the destination container.`)
 	_ = flags.MarkHidden("extract")
 	_ = flags.MarkHidden("pause")
 }
@@ -378,7 +380,7 @@ func copyToContainer(container string, containerPath string, hostPath string) er
 			target = filepath.Dir(target)
 		}
 
-		copyFunc, err := registry.ContainerEngine().ContainerCopyFromArchive(registry.GetContext(), container, target, reader)
+		copyFunc, err := registry.ContainerEngine().ContainerCopyFromArchive(registry.GetContext(), container, target, reader, entities.CopyOptions{Chown: chown})
 		if err != nil {
 			return err
 		}
--- a/docs/source/markdown/podman-cp.1.md
+++ b/docs/source/markdown/podman-cp.1.md
@@ -4,9 +4,9 @@
 podman\-cp - Copy files/folders between a container and the local filesystem
 
 ## SYNOPSIS
-**podman cp** [*container*:]*src_path* [*container*:]*dest_path*
+**podman cp** [*options*] [*container*:]*src_path* [*container*:]*dest_path*
 
-**podman container cp** [*container*:]*src_path* [*container*:]*dest_path*
+**podman container cp** [*options*] [*container*:]*src_path* [*container*:]*dest_path*
 
 ## DESCRIPTION
 Copy the contents of **src_path** to the **dest_path**. You can copy from the container's filesystem to the local machine or the reverse, from the local filesystem to the container.
@@ -60,6 +60,13 @@ Using `-` as the *src_path* streams the contents of STDIN as a tar archive. The 
 Note that `podman cp` ignores permission errors when copying from a running rootless container.  The TTY devices inside a rootless container are owned by the host's root user and hence cannot be read inside the container's user namespace.
 
 ## OPTIONS
+
+#### **--archive**, **-a**
+
+Archive mode (copy all uid/gid information).
+When set to true, files copied to a container will have changed ownership to the primary uid/gid of the container.
+When set to false, maintain uid/gid from archive sources instead of changing them to the primary uid/gid of the destination container.
+The default is *true*.
 
 ## ALTERNATIVES
 
--- a/libpod/container_api.go
+++ b/libpod/container_api.go
@@ -840,7 +840,7 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
 
 // CopyFromArchive copies the contents from the specified tarStream to path
 // *inside* the container.
-func (c *Container) CopyFromArchive(ctx context.Context, containerPath string, tarStream io.Reader) (func() error, error) {
+func (c *Container) CopyFromArchive(ctx context.Context, containerPath string, chown bool, tarStream io.Reader) (func() error, error) {
 	if !c.batched {
 		c.lock.Lock()
 		defer c.lock.Unlock()
@@ -850,7 +850,7 @@ func (c *Container) CopyFromArchive(ctx context.Context, containerPath string, t
 		}
 	}
 
-	return c.copyFromArchive(ctx, containerPath, tarStream)
+	return c.copyFromArchive(ctx, containerPath, chown, tarStream)
 }
 
 // CopyToArchive copies the contents from the specified path *inside* the
--- a/pkg/api/handlers/compat/containers_archive.go
+++ b/pkg/api/handlers/compat/containers_archive.go
@@ -9,6 +9,7 @@ import (
 	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/pkg/api/handlers/utils"
 	"github.com/containers/podman/v3/pkg/copy"
+	"github.com/containers/podman/v3/pkg/domain/entities"
 	"github.com/containers/podman/v3/pkg/domain/infra/abi"
 	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
@@ -92,11 +93,13 @@ func handleHeadAndGet(w http.ResponseWriter, r *http.Request, decoder *schema.De
 
 func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, runtime *libpod.Runtime) {
 	query := struct {
-		Path string `schema:"path"`
+		Path  string `schema:"path"`
+		Chown bool   `schema:"copyUIDGID"`
 		// TODO handle params below
 		NoOverwriteDirNonDir bool `schema:"noOverwriteDirNonDir"`
-		CopyUIDGID           bool `schema:"copyUIDGID"`
-	}{}
+	}{
+		Chown: utils.IsLibpodRequest(r), // backward compatibility
+	}
 
 	err := decoder.Decode(&query, r.URL.Query())
 	if err != nil {
@@ -107,7 +110,7 @@ func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, 
 	containerName := utils.GetName(r)
 	containerEngine := abi.ContainerEngine{Libpod: runtime}
 
-	copyFunc, err := containerEngine.ContainerCopyFromArchive(r.Context(), containerName, query.Path, r.Body)
+	copyFunc, err := containerEngine.ContainerCopyFromArchive(r.Context(), containerName, query.Path, r.Body, entities.CopyOptions{Chown: query.Chown})
 	if errors.Cause(err) == define.ErrNoSuchCtr || os.IsNotExist(err) {
 		// 404 is returned for an absent container and path.  The
 		// clients must deal with it accordingly.
--- a/pkg/bindings/containers/archive.go
+++ b/pkg/bindings/containers/archive.go
@@ -50,11 +50,21 @@ func Stat(ctx context.Context, nameOrID string, path string) (*entities.Containe
 }
 
 func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader) (entities.ContainerCopyFunc, error) {
+	return CopyFromArchiveWithOptions(ctx, nameOrID, path, reader, nil)
+}
+
+// CopyFromArchiveWithOptions FIXME: remove this function and make CopyFromArchive accept the option as the last parameter in podman 4.0
+func CopyFromArchiveWithOptions(ctx context.Context, nameOrID string, path string, reader io.Reader, options *CopyOptions) (entities.ContainerCopyFunc, error) {
 	conn, err := bindings.GetClient(ctx)
 	if err != nil {
 		return nil, err
 	}
-	params := url.Values{}
+
+	params, err := options.ToParams()
+	if err != nil {
+		return nil, err
+	}
+
 	params.Set("path", path)
 
 	return func() error {
--- a/pkg/domain/entities/containers.go
+++ b/pkg/domain/entities/containers.go
@@ -160,6 +160,13 @@ type CommitOptions struct {
 	Writer         io.Writer
 }
 
+type CopyOptions struct {
+	// If used with ContainerCopyFromArchive and set to true
+	// it will change ownership of files from the source tar archive
+	// to the primary uid/gid of the destination container.
+	Chown bool
+}
+
 type CommitReport struct {
 	Id string //nolint
 }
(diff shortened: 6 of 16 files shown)


cmd/podman/containers/cp.go
```diff
@@ -50,12 +50,14 @@ var (
...
 var (
 	cpOpts entities.ContainerCpOptions
+	chown  bool
 )
 
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Deprecated...")
 	flags.BoolVar(&cpOpts.Pause, "pause", true, "Deprecated")
+	flags.BoolVarP(&chown, "archive", "a", false, "Archive mode (copy all uid/gid information)")
```
matejvasek, line 60: On `docker` this is by default false so no `chown` is done. Since `podman` should be drop in replacement for `docker` I made it false by default here, however this might be surprising for current users of `podman`.

pkg/domain/entities/engine_container.go
```diff
@@ -20,7 +20,7 @@ type ContainerEngine interface {
 	ContainerCheckpoint(ctx context.Context, namesOrIds []string, options CheckpointOptions) ([]*CheckpointReport, error)
 	ContainerCleanup(ctx context.Context, namesOrIds []string, options ContainerCleanupOptions) ([]*ContainerCleanupReport, error)
 	ContainerCommit(ctx context.Context, nameOrID string, options CommitOptions) (*CommitReport, error)
-	ContainerCopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader) (ContainerCopyFunc, error)
+	ContainerCopyFromArchive(ctx context.Context, nameOrID, path string, chown bool, reader io.Reader) (ContainerCopyFunc, error)
```
matejvasek, line 23: Is it OK to change this API? If not I could sneak in the boolean in context.

pkg/api/handlers/compat/containers_archive.go
```diff
@@ -92,11 +92,13 @@ func handleHeadAndGet(w http.ResponseWriter, r *http.Request, decoder *schema.De
...
 	query := struct {
-		Path string `schema:"path"`
+		Path       string `schema:"path"`
+		CopyUIDGID bool   `schema:"copyUIDGID"`
 		// TODO handle params below
 		NoOverwriteDirNonDir bool `schema:"noOverwriteDirNonDir"`
-		CopyUIDGID           bool `schema:"copyUIDGID"`
-	}{}
+	}{
+		CopyUIDGID: utils.IsLibpodRequest(r), // backward compatibility
```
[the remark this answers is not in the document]
matejvasek: Also useful for users of `Podman Golang bindings`

rhatdan:
You should also add the other `follow-link` option

  -L, --follow-link   Always follow symbol link in SRC_PATH

I am not sure what it does, but we need to have it implemented.

test/apiv2/23-containersArchive.at
```diff
@@ -72,6 +72,20 @@ if [ "$(tar -xf "${TMPD}/body.tar" hello.txt  --to-stdout)" != "Hello" ]; then
...
 fi
 
+# test if uid/gid was set correctly in the server
+uidngid=$($PODMAN_BIN --root $WORKDIR/server_root exec "${CTR}" stat -c "%u:%g" "/tmp/hello.txt")
+if [[ "${uidngid}" != "1042:1043" ]]; then
+  echo -e "${red}NOK: UID/GID of the file doesn't match.${nc}" 1>&2;
+  ARCHIVE_TEST_ERROR="1"
+fi
+
+# TODO: uid/gid should be also preserved on way back (GET request)
```
matejvasek, line 82: This is however task for another PR.

matejvasek:
> You should also add the other `follow-link` option
> 
> -L, --follow-link Always follow symbol link in SRC_PATH
> 
> I am not sure what it does, but we need to have it implemented.

I would prefer this to be done in another PR.

cmd/podman/containers/cp.go
```diff
@@ -50,12 +50,17 @@ var (
...
+	chown  bool
 )
 
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Deprecated...")
 	flags.BoolVar(&cpOpts.Pause, "pause", true, "Deprecated")
+	flags.BoolVarP(&chown, "archive", "a", false, `Archive mode (copy all uid/gid information).
+When set to true, files copied to a container will have changed ownership to the user of the container.
+When set to false, files will preserve original uid/gid form the source.
```
rhatdan, line 62: s/form/from/g

cmd/podman/containers/cp.go
```diff
@@ -50,12 +50,17 @@ var (
...
 	cpOpts entities.ContainerCpOptions
+	chown  bool
 )
 
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Deprecated...")
 	flags.BoolVar(&cpOpts.Pause, "pause", true, "Deprecated")
+	flags.BoolVarP(&chown, "archive", "a", false, `Archive mode (copy all uid/gid information).
+When set to true, files copied to a container will have changed ownership to the user of the container.
```
rhatdan, line 61: How about:

"Maintain uid/gid from archive sources instead of changing them to the primary uid/gid of the container"

@TomSweeneyRedHat WDYT?

docs/source/markdown/podman-cp.1.md
```diff
@@ -61,6 +61,13 @@ Note that `podman cp` ignores permission errors when copying from a running root
 
 ## OPTIONS
 
+#### **--archive**, **-a**
+
+Archive mode (copy all uid/gid information).
+When set to true, files copied to a container will have changed ownership to the user of the container.
+When set to false, files will preserve original uid/gid form the source.
```
rhatdan, line 68: s/form/from/g

libpod/container_api.go
```diff
@@ -840,7 +840,7 @@ func (c *Container) ShouldRestart(ctx context.Context) bool {
 
 // CopyFromArchive copies the contents from the specified tarStream to path
 // *inside* the container.
-func (c *Container) CopyFromArchive(ctx context.Context, containerPath string, tarStream io.Reader) (func() error, error) {
+func (c *Container) CopyFromArchive(ctx context.Context, containerPath string, chown bool, tarStream io.Reader) (func() error, error) {
```
rhatdan, line 843: Lets do this with a options structure, So that when we open a PR for links, we don't have to change the interface again.

```
struct CopyArchiveOptions {
           Chown bool,
}
```

pkg/domain/entities/engine_container.go
rhatdan: I think it is better to change the API, but lets do it with a struct, so we can add fields in the future.

rhatdan (approved):

cmd/podman/containers/cp.go
matejvasek: @vrothberg shall I change the default value to `true` to make it backward compatible with older `podman`?

vrothberg: Thanks for checking. I think we should remain backwards compat with Podman, so `true` may be a good way forward.

Other may feel differently though. @rhatdan @mheon WDYT?

vrothberg:
> > You should also add the other `follow-link` option
> > -L, --follow-link Always follow symbol link in SRC_PATH
> > I am not sure what it does, but we need to have it implemented.
> 
> I would prefer this to be done in another PR.

Yes, this should be another PR.

pkg/api/handlers/compat/containers_archive.go
```diff
@@ -92,11 +94,13 @@ func handleHeadAndGet(w http.ResponseWriter, r *http.Request, decoder *schema.De
 
 func handlePut(w http.ResponseWriter, r *http.Request, decoder *schema.Decoder, runtime *libpod.Runtime) {
 	query := struct {
-		Path string `schema:"path"`
+		Path       string `schema:"path"`
+		CopyUIDGID bool   `schema:"copyUIDGID"`
```
vrothberg, line 98: Can we name it `chown`?

Otherwise, `chown`, `archive` and `copyUIDGID` all refer to the same thing.

pkg/bindings/containers/archive.go
```diff
@@ -49,13 +50,16 @@ func Stat(ctx context.Context, nameOrID string, path string) (*entities.Containe
...
+func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader, options *CopyOptions) (entities.ContainerCopyFunc, error) {
 	conn, err := bindings.GetClient(ctx)
 	if err != nil {
 		return nil, err
 	}
 	params := url.Values{}
 	params.Set("path", path)
+	if options != nil && options.Chown != nil {
+		params.Set("copyUIDGID", strconv.FormatBool(*options.Chown))
+	}
```
vrothberg, line 62: Can you use `params.WithChown(options.Chown)` (see other bindings)?

pkg/domain/entities/containers.go
```diff
@@ -160,6 +160,10 @@ type CommitOptions struct {
 	Writer         io.Writer
 }
 
+type CopyOptions struct {
```
vrothberg, line 163: PLease add comments.

vrothberg (commented):
Thanks!  Looking like a good direction.

cmd/podman/containers/cp.go
```diff
@@ -50,12 +50,17 @@ var (
...
 )
 
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Deprecated...")
 	flags.BoolVar(&cpOpts.Pause, "pause", true, "Deprecated")
+	flags.BoolVarP(&chown, "archive", "a", false, `Archive mode (copy all uid/gid information).
+When set to true, files copied to a container will have changed ownership to the primary uid/gid of the container.
+When set to false, maintain uid/gid from archive sources instead of changing them to the primary uid/gid of the container.
+The default is false.`)
```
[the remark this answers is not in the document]
matejvasek: Any suggestion how to trim it? The `docker` says only `Archive mode (copy all uid/gid information).` but that is IMHO totally contraintuitive -- if `-a=true` the it does chown it, DOESN'T copies uid/gid from the archve!

pkg/api/handlers/compat/containers_archive.go
matejvasek: We could.

pkg/bindings/containers/archive.go
matejvasek: @vrothberg don't you mean to use `ToParams()` ?

cmd/podman/containers/cp.go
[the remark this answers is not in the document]
vrothberg: How about "Chown copied files to the primary uid/gid of the container"?

cmd/podman/containers/cp.go
matejvasek: I made it true by default

cmd/podman/containers/cp.go
```diff
@@ -50,12 +50,14 @@ var (
...
 var (
 	cpOpts entities.ContainerCpOptions
+	chown  bool
 )
 
 func cpFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.BoolVar(&cpOpts.Extract, "extract", false, "Deprecated...")
 	flags.BoolVar(&cpOpts.Pause, "pause", true, "Deprecated")
+	flags.BoolVarP(&chown, "archive", "a", true, `Chown copied files to the primary uid/gid of the container.`)
```
vrothberg, line 60: One more: can we say "destination container"?  There's another PR up for adding container-to-container copy which may cause confusion.

Also please drop the trailing dot.

docs/source/markdown/podman-cp.1.md
```diff
@@ -61,6 +61,13 @@ Note that `podman cp` ignores permission errors when copying from a running root
 
 ## OPTIONS
 
+#### **--archive**, **-a**
+
+Archive mode (copy all uid/gid information).
+When set to true, files copied to a container will have changed ownership to the primary uid/gid of the container.
```
vrothberg, line 67:
```suggestion
When set to true, files copied to a container will have changed ownership to the primary uid/gid of the destination container.
```

pkg/bindings/containers/archive.go
```diff
@@ -49,12 +49,17 @@ func Stat(ctx context.Context, nameOrID string, path string) (*entities.Containe
 	return statReport, finalErr
 }
 
-func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader) (entities.ContainerCopyFunc, error) {
+func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader, options *CopyOptions) (entities.ContainerCopyFunc, error) {
```
vrothberg, line 52: Sorry for noticing just now.  This would break the bindings which is something we cannot do at the moment without having to bump Podman to 4.0.

vrothberg (commented):
I just realized it's a breaking change to the bindings.

We should be able to handle everything on the client side in cmd/podman though.  Both, copier.Put and copier.Get allow for chowning.  In order to get the data, I think we need to extend `podman container inspect` with the necessary information about the user.

The only drawback is that the libpod/archive endpoint would _not_ chown by default anymore, which is a risk I am willing to take.

matejvasek:
>I just realized it's a breaking change to the bindings.

@vrothberg 
There should be some compatibility:
User can use old `podman-remote` or an application compiled with old `Binding` lib it should work against new server.

vrothberg:
> > I just realized it's a breaking change to the bindings.
> 
> @vrothberg
> There should be some compatibility:
> User can use old `podman-remote` or an application compiled with old `Binding` lib it should work against new server.

Very fair point. Seems like we'd break one way or another :(

@baude @rhatdan WDYT?

matejvasek:
> Very fair point. Seems like we'd break one way or another :(

Too bad that Go doesn't have function overloading :disappointed:

vrothberg:
> > Very fair point. Seems like we'd break one way or another :(
> 
> Too bad that Go doesn't have function overloading disappointed

An alternative solution would be to add a new function to the bindings which supports the new option.  Add a clear comment that this will be removed with 4.0 and should not be used by any outside user.  It's a trick but sounds like a good compromise at the moment.  We can move on with the work and existing (bindings) users won't suddenly break.

matejvasek:
>An alternative solution would be to add a new function to the bindings which supports the new option. Add a clear comment that this will be removed with 4.0 and should not be used by any outside user. It's a trick but sounds like a good compromise at the moment. We can move on with the work and existing (bindings) users won't suddenly break.

@vrothberg 
or abuse variadic args:
`func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader, options ...*CopyOptions) (entities.ContainerCopyFunc, error)`  making the option argument optional


just kidding :laughing: 

I'll do as you proposed.

matejvasek:
>More sane alternative might be options pattern

also I think such an API is more convenient to use.
I based it on: https://pkg.go.dev/github.com/manifestival/manifestival?utm_source=godoc#Manifest.Apply

pkg/api/handlers/compat/containers_archive.go
```diff
@@ -5,6 +5,8 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/containers/podman/v3/pkg/domain/entities"
+
```
vrothberg, line 9: Please remove the blank line.

vrothberg:
> More sane alternative might be options pattern (although it's not used in rest of the API which would be little bit inconsistent):

That would be a breaking change we want to avoid.  Adding a second function, e.g., `CopyFromArchiveWithOptions` with a comment that clearly indicates that this function will be removed with Podman 4.0 would work.

matejvasek:
>That would be a breaking change we want to avoid.

@vrothberg 
Would it be? 
`func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader, options ...CopyOption) (entities.ContainerCopyFunc, error)`

The options are there via variadic args meaning that they are entirely optional so the function is still callable old way, not breaking existing code. Or Am I wrong?

matejvasek:
I think I should be OK if pkg is distributed as source code.

vrothberg:
> I think I should be OK if pkg is distributed as source code.

That would work as well. Nice idea. Please make sure to drop a `// FIXME: drop variadic args with Podma 4.0`` in the code.

matejvasek:
@vrothberg @rhatdan PTAL

matejvasek:
>That would work as well. Nice idea. Please make sure to drop a `// FIXME: drop variadic args with Podman 4.0`` in the code.

I thought we would keep the variadic. If they are to be deleted then I rather do original suggestion.

vrothberg:
> > That would work as well. Nice idea. Please make sure to drop a `// FIXME: drop variadic args with Podman 4.0`` in the code.
> 
> I thought we would keep the variadic. 

Yes, for consistency with other bindings.

> If they are to be deleted then I rather do original suggestion.

OK, sounds good to me. Thank you!

pkg/bindings/containers/archive.go
```diff
@@ -50,11 +50,21 @@ func Stat(ctx context.Context, nameOrID string, path string) (*entities.Containe
 }
 
 func CopyFromArchive(ctx context.Context, nameOrID string, path string, reader io.Reader) (entities.ContainerCopyFunc, error) {
+	return CopyFromArchiveWith(ctx, nameOrID, path, reader, nil)
+}
+
+// CopyFromArchiveWith FIXME: remove this function and make CopyFromArchive accept the option as the last parameter in podman 4.0
+func CopyFromArchiveWith(ctx context.Context, nameOrID string, path string, reader io.Reader, options *CopyOptions) (entities.ContainerCopyFunc, error) {
```
vrothberg, line 57:
```suggestion
func CopyFromArchiveWithOptions(ctx context.Context, nameOrID string, path string, reader io.Reader, options *CopyOptions) (entities.ContainerCopyFunc, error) {
```

vrothberg (commented):
Other than the naming nit: LGTM
@mheon PTAL

rhatdan:
/lgtm
@matejvasek Could you open a PR with containers/common to add --archive as a option, so users who want the Docker defaults can change the way podman works.

rhatdan:
BTW @matejvasek nice work and thanks.

matejvasek:
> @matejvasek Could you open a PR with containers/common to add --archive as a option, so users who want the Docker defaults can change the way podman works.

to containers/common? where?

rhatdan:
Yes https://github.com/containers/common/pkg/config

vrothberg:
To add a new field to containers.conf which, if set, would overwrite the default value of `--archive`.

I think it should go under the "[engine]" table.  Naming is always the hardest ... "chown_copied_files"?

rhatdan:
Yes this one will need a good description.

ocafebabe:
@matejvasek are there still plans to implement "follow-link" (-L) like docker?

matejvasek:
@ocafebabe I haven't seen need for it so far (for my use), but feel free to submit PR.

Issue #10813, opened 2021-06-29
podman system connection list: please sort


edsantiago:
The output of `podman system connection list` is random. This is causing [flakes](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4612888552800256/html/int-podman-ubuntu-2010-root-host.log.html#t--default--1). Please make it sort deterministically.

mheon:
@ashley-cui @baude PTAL

PR #10819, opened 2021-06-29
[NO TESTS NEEDED] Make system connection ls deterministic

ashley-cui:
Sort system connection ls by name, making the output deterministic. Previously, we were just iterating through a map, which caused CI flakes.

[NO TESTS NEEDED] I think, since its just about deterministic listing. 

Closes: #10813


mheon:
LGTM
/approve

PR #10836, opened 2021-07-01
podman diff accept two images or containers

### LINKED ISSUE #10649: Incorrect documentation examples for podman-image-diff

/kind bug

**Description**
In the documentation at http://docs.podman.io/en/latest/markdown/podman-image-diff.1.html
this example: `podman diff --format json redis:old redis:alpine` is used. But this does not work with the current version of podman. 

```
:/$ podman image diff alpine:latest alpine:3
:/$ Error: accepts 1 arg(s), received 2 
```
```
:/$ podman diff alpine:latest alpine:3
:/$ Error: `podman diff` accepts at most one argument 
```
I think when a single image is mentioned the diff should be done between the image and its parent layer and when two are mentioned the diff should be between them. 

The JSON output when `--format=json` flag is used is not indented and `JSON` is not accepted as a valid option, which in my opinion should be. This would just be a minor fix in the regular expression here: https://github.com/containers/podman/blob/0c38ac44efcaa83f98c727408fbba78456f9edd8/vendor/github.com/containers/common/pkg/report/validate.go#L5

```
podman version 3.1.2
```

Please provide correct documentation examples, and if there are any potential bug fixes then let me know! I would love to try working on them

Luap99:
First, make podman diff accept optionally a second argument. This allows
the user to specify a second image/container to compare the first with.
If it is not set the parent layer will be used as before.

Second, podman container diff should only use containers and podman
image diff should only use images. Previously, podman container diff
would use the image when both an image and container with this name
exists.

To make this work two new parameters have been added to the api. If they
are not used the previous behaviour is used. The same applies to the
bindings.

Fixes #10649

### DIFF
--- a/docs/source/markdown/podman-container-diff.1.md
+++ b/docs/source/markdown/podman-container-diff.1.md
@@ -1,0 +1,54 @@
+% podman-container-diff(1)
+
+## NAME
+podman\-container\-diff - Inspect changes on a container's filesystem
+
+## SYNOPSIS
+**podman container diff** [*options*] *container* [*container*]
+
+## DESCRIPTION
+Displays changes on a container's filesystem. The container will be compared to its parent layer or the second argument when given.
+
+The output is prefixed with the following symbols:
+
+| Symbol | Description |
+|--------|-------------|
+| A | A file or directory was added.   |
+| D | A file or directory was deleted. |
+| C | A file or directory was changed. |
+
+## OPTIONS
+
+#### **--format**
+
+Alter the output into a different format. The only valid format for **podman container diff** is `json`.
+
+#### **--latest**, **-l**
+
+Instead of providing the container name or ID, use the last created container. If you use methods other than Podman
+to run containers such as CRI-O, the last started container could be from either of those methods. (This option is not available with the remote Podman client)
+
+## EXAMPLE
+
+```
+# podman container diff container1
+C /usr
+C /usr/local
+C /usr/local/bin
+A /usr/local/bin/docker-entrypoint.sh
+```
+
+```
+$ podman container diff --format json container1 container2
+{
+     "added": [
+          "/test"
+     ]
+}
+```
+
+## SEE ALSO
+**[podman(1)](podman.1.md)**, **[podman-container(1)](podman-container.1.md)**
+
+## HISTORY
+July 2021, Originally compiled by Paul Holzinger <pholzing@redhat.com>
--- a/docs/source/markdown/podman-diff.1.md
+++ b/docs/source/markdown/podman-diff.1.md
@@ -4,18 +4,24 @@
 podman\-diff - Inspect changes on a container or image's filesystem
 
 ## SYNOPSIS
-**podman diff** [*options*] *name*
-
-**podman container diff** [*options*] *name*
+**podman diff** [*options*] *container|image* [*container|image*]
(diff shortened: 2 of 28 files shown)


infiniteregrets:
Thanks @Luap99, I kind of understood where I was going wrong. 
Can you add this table to docs for image diff:

```
The output is prefixed with the following symbols:

| Symbol | Description |
|--------|-------------|
| A | A file or directory was added. |
| D | A file or directory was deleted. |
| C | A file or directory was changed. |
```

Luap99:
@infiniteregrets I added this all three diff man pages.

infiniteregrets:
Looks good!

libpod/define/diff.go
```diff
@@ -0,0 +1,26 @@
+package define
+
+// extra type to use as enum
+type DiffType uint8
+
+const (
+	// only diff containers
+	DiffContainer DiffType = 1 << iota
```
mheon, line 8: If this is a bitfield, it seems like it would be better to manually specify - 0b1, 0b10, etc.

mheon:
One nit otherwise LGTM

libpod/define/diff.go
Luap99: `1 << iota` is basically just saying to shift the bit one step for each entry. It should be less error prone than writing this manually especially when you add new entries, However I understand that this syntax is confusing if you are not familiar with it. I can remove it if you want.

docs/source/markdown/podman-container-diff.1.md
```diff
@@ -0,0 +1,54 @@
...
+|--------|-------------|
+| A | A file or directory was added.   |
+| D | A file or directory was deleted. |
+| C | A file or directory was changed. |
+
+## OPTIONS
+
+#### **--format**
+
+Alter the output into a different format. The only valid format for diff is `json`.
```
TomSweeneyRedHat, line 24:
```suggestion
Alter the output into a different format. The only valid format for **podman diff** is `json`.
```

docs/source/markdown/podman-container-diff.1.md
```diff
@@ -0,0 +1,54 @@
...
+$ podman container diff --format json container1 container2
+{
+     "added": [
+          "/test"
+     ]
+}
+```
+
+## SEE ALSO
+podman(1), podman-container(1)
```
TomSweeneyRedHat, line 51:
```suggestion
**[podman(1)](podman.1.md)**, **[podman-container(1)](podman-container.1.md)**
```

docs/source/markdown/podman-diff.1.md
```diff
@@ -46,8 +49,13 @@ A /usr/local/bin/docker-entrypoint.sh
...
 ```
 
+```
+$ podman diff container1 image1
+A /test
+```
+
 ## SEE ALSO
-podman(1)
+podman(1), podman-container(1), podman-image(1)
```
TomSweeneyRedHat, line 58: Did you want container or container-diff here?  I went with container-diff, but maybe both?
```suggestion
**[podman(1)](podman.1.md)**, **[podman-container-diff(1)](podman-container-diff.1.md)**, **[podman-image(1)](podman-image.1.md)**
```

Luap99:
I had to repush to fix a test.

Issue #10826, opened 2021-06-30
[placeholder] podman images: "failed to kill PID"


edsantiago:
Weird one. I've only seen it once (rawhide, rootless, podman-3.3.0-0.20.dev.git599b7d7.fc35) and have not been able to reproduce:
```
podman corrupt images - system reset test
     (from function `is' in file /usr/share/podman/test/system/helpers.bash, line 456,
    from function `_corrupt_image_test' in file /usr/share/podman/test/system/330-corrupt-images.bats, line 84,
    in test file /usr/share/podman/test/system/330-corrupt-images.bats, line 120)
     `_corrupt_image_test "system reset -f"' failed with status 125
   umount: /tmp/podman_corrupt_test.z4I9ih/root/overlay: No such file or directory
   $ podman load -i /tmp/podman_corrupt_test.z4I9ih/img.tar
   Getting image source signatures
   Copying blob sha256:03901b4a2ea88eeaad62dbe59b072b28b6efa00491962b8741081c5df50c65e0
   Copying config sha256:961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
   Writing manifest to image destination
   Storing signatures
   Loaded image(s): sha256:961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
   $ podman tag 961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4 quay.io/libpod/alpine@sha256:634a8f35b5f16dcf4aaa0822adc0b1964bb786fca12f6831de8ddc45e5986a00
   removed '/tmp/podman_corrupt_test.z4I9ih/root/vfs-images/961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4/manifest'
   $ podman images
   Error: error retrieving label for image "961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4": you may need to remove the image to resolve the error: fallback error checking whether image is a manifest list: error locating item named "manifest" for image with ID "961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4": file does not exist: error locating item named "manifest" for image with ID "961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4": file does not exist
   [ rc=125 (expected) ]
   $ podman system reset -f
[...]

PR #10837, opened 2021-07-01
podman: ignore ESRCH from kill

giuseppe:
Closes: https://github.com/containers/podman/issues/10826

[NO TESTS NEEDED] Fixes a race condition

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -268,7 +268,9 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (_ boo
 		}
 		if retErr != nil && pid > 0 {
 			if err := unix.Kill(pid, unix.SIGKILL); err != nil {
-				logrus.Errorf("failed to kill %d", pid)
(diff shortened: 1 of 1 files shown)


pkg/rootless/rootless_linux.go
```diff
@@ -394,7 +396,9 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (_ boo
 			}
 
 			if err := unix.Kill(int(pidC), s.(unix.Signal)); err != nil {
-				logrus.Errorf("failed to kill %d", int(pidC))
+				if err != unix.ESRCH {
+					logrus.Errorf("failed to kill %d: %v", int(pidC), err)
```
edsantiago, line 400: One habit that has served me well over the years is: always write unique error messages, so it's possibly to identify the exact place in the code where a failure happened. Would it make sense here to disambiguate these two, by adding a short relevant context descriptor to each error string?

giuseppe: good idea, made them different

mheon:
LGTM

PR #10869, opened 2021-07-07
[CI:DOCS] [3.2] podman search: clarify that results depend on implementation

vrothberg:
Clarify in the man page that podman-search is not a generally reliable
way of determining the presence/existence of an image.  The results of
the v1 and the v2 endpoints depend on the implementation of each
registry; the semantics are not really specified.  Some registries may
not support search at all as it's not part of the OCI dist spec.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1978556


rhatdan:
LGTM
@mtrmac @TomSweeneyRedHat PTAL

mtrmac (commented):
LGTM. The original PR is https://github.com/containers/podman/pull/10842 for the record.

PR #10870, opened 2021-07-07
manifest push --rm: use libimage for removal

vrothberg:
Use libimage for removing the manifest instead of going directly
through the store.

[NO TESTS NEEDED]


mheon:
LGTM, thanks!

Issue #9949, opened 2021-04-06
Dry-run, download-only, notifiy-only modes for podman auto-update


jpf91:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

It'd be great to have much more configuration options for podman auto-update: I have quite some services which I don't want updated automatically. These are things such as IPA client containers which perform a join to the IPA directory which should not happen unattended (can easily break depending on DNS and network state). Because of this, I have systemd services which should not be restarted if an update is available. I also have services which could be restarted but wouldn't apply the new configuration anyway, as they start persistent containers. To summarize, it'd be great to have these operation modes:

* Only check if updates are available. Print a list of affected container names. Provide some way to react to this, so that users can send this information via email.
* Check for updates and pull, but do not restart containers
* Check for updates, pull and restart containers

in general, it'd be great to have some more verbose output from auto-update. (Even if it's just for testing whether the labels have been set properly).

I've programmed a small hack to do / demonstrate all this:
```bash
#!/bin/bash

IGNORE_LABEL=1
VERBOSE=1
DOWNLOAD=0
SEND_UPDATE_MAIL=0

if [ -f /etc/podman-download-updates.conf ]; then
    source /etc/podman-download-updates.conf
fi 

CONTAINERS=$(podman ps --format {{.ID}})
DOWNLOAD_IMAGES=()
CONTAINER_UPDATES=()

echo "Scanning running containers:"
for CID in $CONTAINERS; do
    IMAGE=$(podman inspect $CID --format {{.ImageName}})
    IMAGE_ID=$(podman inspect $CID --format {{.ImageID}})
    NAME=$(podman inspect $CID --format {{.Name}})
    LABELS=$(podman inspect $CID --format {{.Config.Labels}})

    echo $LABELS | grep -q "io.containers.autoupdate:image"
    IS_ENABLED=$?

    if [[ $IS_ENABLED -eq 0 || IGNORE_LABEL -eq 1 ]]; then
        test $VERBOSE -eq 1 && echo "* $NAME($CID) checking for updates"
        
        TIME_REG=$(skopeo inspect docker://$IMAGE 2> /dev/null | jq '.Created')
        TIME_LOCAL=$(skopeo inspect containers-storage:$IMAGE 2> /dev/null | jq '.Created')
        TIME_CONTAINER=$(skopeo inspect containers-storage:$IMAGE_ID 2> /dev/null | jq '.Created')

[...]

PR #10880, opened 2021-07-08
auto-update: make output more user friendly

vrothberg:
The rather raw and scarce output of `podman auto-update` has been a                         
thorn in my eyes for a longer while.  So far, Podman would only print                       
updated systemd units, one per line, without further formatting.                            
                                                                                            
Motivated by issue #9949 which is asking for some more useful                               
information in combination with a dry-run feature, I sat down and                           
reflected which information may come in handy.                                              
                                                                                            
Running `podman auto-update` will now look as follows:                                      
                                                                                            
```                                                                                         
$ podman auto-update                                                                        
Trying to pull [...]                                                                        
                                                                                            
UNIT                    CONTAINER            IMAGE                   POLICY      UPDATED    
container-test.service  08fd34e533fd (test)  localhost:5000/busybox  registry    false      
```                                                                                         
                                                                                            
Also refactor the spaghetti code in the backend a bit to make it easier                     
to digest and maintain.                                                                     
                                                                                            
For easier testing and for the sake of consistency with other commands                      
listing output, add a `--format` flag.                                                                                                    
                                                                                            
The man page will get an overhaul in a follow up commit.                                    
                                                                                            
### DIFF
--- a/cmd/podman/auto-update.go
+++ b/cmd/podman/auto-update.go
@@ -50,7 +50,9 @@ func autoUpdate(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("`%s` takes no arguments", cmd.CommandPath())
 	}
 	report, failures := registry.ContainerEngine().AutoUpdate(registry.GetContext(), autoUpdateOptions)
-	if report != nil {
+	if report != nil && len(report.Units) > 0 {
+		// Make it more obvious to users what the output means.
+		fmt.Println("\nRestarted the following systemd units:")
 		for _, unit := range report.Units {
 			fmt.Println(unit)
 		}
--- a/cmd/podman/auto-update.go
+++ b/cmd/podman/auto-update.go
@@ -1,10 +1,15 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/containers/common/pkg/auth"
 	"github.com/containers/common/pkg/completion"
+	"github.com/containers/common/pkg/report"
+	"github.com/containers/podman/v3/cmd/podman/common"
 	"github.com/containers/podman/v3/cmd/podman/registry"
 	"github.com/containers/podman/v3/pkg/domain/entities"
 	"github.com/containers/podman/v3/pkg/errorhandling"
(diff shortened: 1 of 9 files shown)


vrothberg:
@edsantiago, I would love your thoughts on the proposed changes.

cmd/podman/auto-update.go
```diff
@@ -49,11 +49,36 @@ func autoUpdate(cmd *cobra.Command, args []string) error {
...
-		for _, unit := range report.Units {
-			fmt.Println(unit)
+	if report == nil {
+		return errorhandling.JoinErrors(failures)
+	}
+
+	if len(report.Containers) > 0 {
+		fmt.Println("Scanned the following containers:")
+		for _, c := range report.Containers {
+			fmt.Printf(" * ID: %s, name: %s, image: %s, unit: %s, updated: %v", c.ID[:12], c.Name, c.Image, c.SystemdUnit, c.Updated)
```
mheon, line 61: Could we do table output on this and Images? I don't know if our table writer supports doing more than one, but it seems like it would be a lot cleaner, output-wise.

edsantiago:
LGTM in principle, but there's a segv that needs fixing, and tests need some work too. I'll look again on Monday. Thanks for taking this on!

cmd/podman/auto-update.go
vrothberg: Maybe we can squeeze everything into one table:
```
UNIT               CONTAINER         IMAGE            UPDATED        PENDING-UPDATE
unit-name          container-name    raw-image-name   {false,true}   {false,true}
```

While the anticipated `--json` would just spit out all data as raw JSON (including digests and IDs etc).

What do you think?

mheon: I think that's probably a better solution - very obvious what happened when it's printed this way

vrothberg:
@edsantiago @mheon, updated.  I like it much more like that :+1:

TomSweeneyRedHat:
LGTM and happy green test buttons.
@edsantiago thoughts?

edsantiago:
Still LGTM but I will leave up to others to slash-lgtm it

baude:
alright, ill bite ...
/lgtm

PR #10881, opened 2021-07-08
Remove GetStore function from Libpod

mheon:
We should not be exposing the store outside of Libpod. We want to encapsulate it as an internal implementation detail - there's no reason functions outside of Libpod should directly be manipulating container storage. Convert the last use to invoke a method on Libpod instead, and remove the function.

[NO TESTS NEEDED] as this is just a refactor.


baude:
ok, ill bite ... LGTM

Issue #10795, opened 2021-06-27
Only the first `docker-compose up -d` succeeds


x-yuri:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The first `docker-compose up -d` succeeds, but the following ones fail. The workaround is to do `docker-compose down` before `up -d`.

**Steps to reproduce the issue:**

`docker-compose.yml`:

```yaml
version: '3'
services:
    app:
        build: .
        command: sleep 1000
```

`Dockerfile`:

```dockerfile
FROM alpine
WORKDIR /app
```

```
$ sudo /usr/local/bin/docker-compose up -d
Creating network "a1_default" with the default driver
Creating a1_app_1 ... 
Creating a1_app_1 ... done
```

Change 1000 to 2000 in `docker-compose.yml`.

```
$ sudo /usr/local/bin/docker-compose up -d
Recreating a1_app_1 ... 
[29100] Failed to execute script docker-compose

ERROR: for a1_app_1  'NoneType' object has no attribute 'get'

ERROR: for app  'NoneType' object has no attribute 'get'
Traceback (most recent call last):
  File "docker-compose", line 3, in <module>
  File "compose/cli/main.py", line 81, in main
  File "compose/cli/main.py", line 203, in perform_command
  File "compose/metrics/decorator.py", line 18, in wrapper
  File "compose/cli/main.py", line 1186, in up
  File "compose/cli/main.py", line 1182, in up
  File "compose/project.py", line 702, in up
  File "compose/parallel.py", line 108, in parallel_execute
  File "compose/parallel.py", line 206, in producer
  File "compose/project.py", line 688, in do
  File "compose/service.py", line 581, in execute_convergence_plan
  File "compose/service.py", line 503, in _execute_convergence_recreate
  File "compose/parallel.py", line 108, in parallel_execute
  File "compose/parallel.py", line 206, in producer
  File "compose/service.py", line 496, in recreate
  File "compose/service.py", line 615, in recreate_container
  File "compose/service.py", line 334, in create_container
  File "compose/service.py", line 922, in _get_container_create_options
  File "compose/service.py", line 962, in _build_container_volume_options
  File "compose/service.py", line 1549, in merge_volume_bindings
  File "compose/service.py", line 1579, in get_container_data_volumes
AttributeError: 'NoneType' object has no attribute 'get'

$ sudo /usr/local/bin/docker-compose ps
        Name             Command      State     Ports
[...]

PR #10893, opened 2021-07-09
Add container config to compat image inspect

baude:
With docker-compose, there is a use case where you can `docker-compose
up -d`, then change a file like docker-compose.yml and run up again.
This requires a ContainerConfig with at least Volumes be populated in
the inspect data.  This PR adds just that.

Fixes: #10795


vrothberg (commented):
LGTM

rhatdan:
I would figure we would need to further populate that structure, but for now
/lgtm
/hold

mheon:
LGTM
/hold cancel

PR #10894, opened 2021-07-09
podman pod create --pid flag

cdoern:
added support for --pid flag. User can specify ns:file, pod, private, or host pid namespace values. container returns an error since you cannot point the ns of the pods infra container
to a container outside of the pod.

Signed-off-by: cdoern <cdoern@redhat.com>

### DIFF
--- a/docs/source/markdown/podman-pod-create.1.md
+++ b/docs/source/markdown/podman-pod-create.1.md
@@ -120,6 +120,14 @@ Add a DNS alias for the container. When the container is joined to a CNI network
 
 Disable creation of /etc/hosts for the pod.
 
+#### **--pid**=*pid*
+
+Set the PID mode for the pod. The default is to create a private PID namespace for the pod. Requires the PID namespace to be shared via --share.
+
+    host: use the host’s PID namespace for the pod
+    ns: join the specified PID namespace
+    private: create a new namespace for the pod (default)
+
 #### **--pod-id-file**=*path*
 
 Write the pod ID to the file.
--- a/libpod/define/pod_inspect.go
+++ b/libpod/define/pod_inspect.go
@@ -103,6 +103,8 @@ type InspectPodInfraConfig struct {
 	CPUQuota int64 `json:"cpu_quota,omitempty"`
 	// CPUSetCPUs contains linux specific CPU data for the container
 	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
+	// Pid is the PID namespace mode of the pod's infra container
+	PidNS string `json:"pid_ns,omitempty"`
 }
 
 // InspectPodContainerInfo contains information on a container in a pod.
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -16,6 +16,7 @@ import (
 	"github.com/containers/podman/v3/libpod/events"
 	"github.com/containers/podman/v3/pkg/namespaces"
 	"github.com/containers/podman/v3/pkg/rootless"
+	"github.com/containers/podman/v3/pkg/specgen"
 	"github.com/containers/podman/v3/pkg/util"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
@@ -2394,6 +2395,25 @@ func WithPodCPUSetCPUs(inp string) PodCreateOption {
 			pod.config.InfraContainer.ResourceLimits.CPU = &specs.LinuxCPU{}
 			pod.config.InfraContainer.ResourceLimits.CPU.Cpus = inp
 		}
+		return nil
+	}
+}
+
+func WithPodPidNS(inp specgen.Namespace) PodCreateOption {
+	return func(p *Pod) error {
+		if p.valid {
+			return define.ErrPodFinalized
+		}
+		if p.config.UsePodPID {
+			switch inp.NSMode {
+			case "container":
+				return errors.Wrap(define.ErrInvalidArg, "Cannot take container in a different NS as an argument")
+			case "host":
+				p.config.UsePodPID = false
+			}
+			p.config.InfraContainer.PidNS = inp
+		}
+
 		return nil
 	}
 }
--- a/libpod/pod.go
+++ b/libpod/pod.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/libpod/lock"
+	"github.com/containers/podman/v3/pkg/specgen"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
@@ -97,6 +98,7 @@ type InfraContainerConfig struct {
 	HasInfraContainer  bool                  `json:"makeInfraContainer"`
 	NoNetwork          bool                  `json:"noNetwork,omitempty"`
 	HostNetwork        bool                  `json:"infraHostNetwork,omitempty"`
+	PidNS              specgen.Namespace     `json:"infraPid,omitempty"`
 	PortBindings       []ocicni.PortMapping  `json:"infraPortBindings"`
 	StaticIP           net.IP                `json:"staticIP,omitempty"`
 	StaticMAC          net.HardwareAddr      `json:"staticMAC,omitempty"`
@@ -168,6 +170,11 @@ func (p *Pod) CPUQuota() int64 {
 		return *resCopy.CPU.Quota
 	}
 	return 0
+}
+
+// PidMode returns the PID mode given by the user ex: pod, private...
+func (p *Pod) PidMode() string {
+	return string(p.config.InfraContainer.PidNS.NSMode)
 }
 
 // Labels returns the pod's labels
--- a/libpod/pod_api.go
+++ b/libpod/pod_api.go
@@ -541,6 +541,7 @@ func (p *Pod) Inspect() (*define.InspectPodData, error) {
 		infraConfig.CPUPeriod = p.CPUPeriod()
 		infraConfig.CPUQuota = p.CPUQuota()
 		infraConfig.CPUSetCPUs = p.ResourceLim().CPU.Cpus
+		infraConfig.PidNS = p.PidMode()
 
 		if len(p.config.InfraContainer.DNSServer) > 0 {
 			infraConfig.DNSServer = make([]string, 0, len(p.config.InfraContainer.DNSServer))
--- a/libpod/runtime_pod_infra_linux.go
+++ b/libpod/runtime_pod_infra_linux.go
@@ -145,6 +145,18 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
 		if len(p.config.InfraContainer.ExitCommand) > 0 {
 			options = append(options, WithExitCommand(p.config.InfraContainer.ExitCommand))
 		}
+
+		if p.config.UsePodPID && p.config.InfraContainer.PidNS.NSMode != "host" {
+			g.AddOrReplaceLinuxNamespace(string(spec.LinuxNamespaceType("pid")), p.config.InfraContainer.PidNS.Value)
+		} else if p.config.InfraContainer.PidNS.NSMode == "host" {
+			newNS := []spec.LinuxNamespace{}
+			for _, entry := range g.Config.Linux.Namespaces {
+				if entry.Type != spec.LinuxNamespaceType("pid") {
+					newNS = append(newNS, entry)
+				}
+			}
+			g.Config.Linux.Namespaces = newNS
+		}
 	}
 	g.SetRootReadonly(true)
 	g.SetProcessArgs(infraCtrCommand)
--- a/libpod/runtime_pod_linux.go
+++ b/libpod/runtime_pod_linux.go
@@ -116,6 +116,7 @@ func (r *Runtime) NewPod(ctx context.Context, options ...PodCreateOption) (_ *Po
 	if pod.config.UsePodCgroup {
 		logrus.Debugf("Got pod cgroup as %s", pod.state.CgroupPath)
 	}
+
 	if !pod.HasInfraContainer() && pod.SharesNamespaces() {
 		return nil, errors.Errorf("Pods must have an infra container to share namespaces")
 	}
--- a/pkg/domain/entities/pods.go
+++ b/pkg/domain/entities/pods.go
@@ -118,6 +118,7 @@ type PodCreateOptions struct {
 	Name               string
 	Net                *NetOptions
 	Share              []string
+	Pid                string
 	Cpus               float64
 	CpusetCpus         string
 }
@@ -144,6 +145,18 @@ func (p *PodCreateOptions) CPULimits() *specs.LinuxCPU {
 		return cpu
 	}
 	return cpu
+}
+
+func setNamespaces(p *PodCreateOptions) ([4]specgen.Namespace, error) {
+	allNS := [4]specgen.Namespace{}
+	if p.Pid != "" {
+		pid, err := specgen.ParseNamespace(p.Pid)
+		if err != nil {
+			return [4]specgen.Namespace{}, err
+		}
+		allNS[0] = pid
+	}
+	return allNS, nil
 }
 
 func (p *PodCreateOptions) ToPodSpecGen(s *specgen.PodSpecGenerator) error {
@@ -177,6 +190,14 @@ func (p *PodCreateOptions) ToPodSpecGen(s *specgen.PodSpecGenerator) error {
 	s.DNSOption = p.Net.DNSOptions
 	s.NoManageHosts = p.Net.NoHosts
 	s.HostAdd = p.Net.AddHosts
+
+	namespaces, err := setNamespaces(p)
+	if err != nil {
+		return err
+	}
+	if !namespaces[0].IsDefault() {
+		s.Pid = namespaces[0]
+	}
 
 	// Cgroup
 	s.CgroupParent = p.CGroupParent
--- a/pkg/specgen/generate/pod_create.go
+++ b/pkg/specgen/generate/pod_create.go
@@ -102,6 +102,10 @@ func createPodOptions(p *specgen.PodSpecGenerator, rt *libpod.Runtime) ([]libpod
 		options = append(options, libpod.WithInfraCommand(p.InfraCommand))
 	}
 
+	if !p.Pid.IsDefault() {
+		options = append(options, libpod.WithPodPidNS(p.Pid))
+	}
+
 	switch p.NetNS.NSMode {
 	case specgen.Default, "":
 		if p.NoInfra {
--- a/pkg/specgen/podspecgen.go
+++ b/pkg/specgen/podspecgen.go
@@ -57,6 +57,10 @@ type PodBasicConfig struct {
 	// (e.g. `podman generate systemd --new`).
 	// Optional.
 	PodCreateCommand []string `json:"pod_create_command,omitempty"`
+	// Pid sets the process id namespace of the pod
+	// Optional (defaults to private if unset). This sets the PID namespace of the infra container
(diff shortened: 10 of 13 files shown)


rhatdan:
What does "pod" mean?  Are you allowing pod's to join other pods pid namespace?

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -120,6 +120,15 @@ Add a DNS alias for the container. When the container is joined to a CNI network
...
 Disable creation of /etc/hosts for the pod.
 
+#### **--pid**=*pid*
+
+Set the PID mode for the pod Default is to create a private PID namespace for the pod
+
+    host: use the host’s PID namespace for the pod
+    ns: join the specified PID namespace
+    private: create a new namespace for the pod (default)
+    pod: The pod's PID namespace will be used for all containers
```
mheon, line 130: I would just omit this... We should still accept it, but it doesn't need to but documented

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -120,6 +120,15 @@ Add a DNS alias for the container. When the container is joined to a CNI network
 
 Disable creation of /etc/hosts for the pod.
 
+#### **--pid**=*pid*
+
+Set the PID mode for the pod Default is to create a private PID namespace for the pod
```
mheon, line 125: Should mention this requires the PID namespace be shared

libpod/define/pod_inspect.go
```diff
@@ -57,6 +57,8 @@ type InspectPodData struct {
 	CPUQuota int64 `json:"cpu_quota,omitempty"`
 	// CPUSetCPUs contains linux specific CPU data for the pod
 	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
+	// Pid contains the process id NS for the pod
```
mheon, line 60: "Pid is the PID namespace mode of the pod's infra container"

libpod/define/pod_inspect.go
```diff
@@ -103,6 +105,8 @@ type InspectPodInfraConfig struct {
 	CPUQuota int64 `json:"cpu_quota,omitempty"`
 	// CPUSetCPUs contains linux specific CPU data for the container
 	CPUSetCPUs string `json:"cpuset_cpus,omitempty"`
+	// Pid contains the process id namespace information for the container
+	Pid string `json:"pid,omitempty"`
```
mheon, line 109: This only needs to be in one place - I think here is probably the more sensible one

libpod/options.go
```diff
@@ -2397,3 +2398,14 @@ func WithPodCPUSetCPUs(inp string) PodCreateOption {
 		return nil
 	}
 }
+
+func WithPid(inp specgen.Namespace) PodCreateOption {
```
mheon, line 2402: Rename to `WithPodPidNS(...)`

libpod/options.go
```diff
@@ -2397,3 +2398,14 @@ func WithPodCPUSetCPUs(inp string) PodCreateOption {
...
+	return func(pod *Pod) error {
+		if pod.valid {
+			return define.ErrPodFinalized
+		}
+		pod.config.Pid = inp
```
mheon, line 2407: We need to require that `UsePodPID` is set in the pod configuration - we have to be sharing the pod PID namespace for this to be set.

libpod/pod.go
```diff
@@ -72,6 +73,8 @@ type PodConfig struct {
 
 	// ID of the pod's lock
 	LockID uint32 `json:"lockID"`
+	// Pid contains the pod's process id namespace
+	Pid specgen.Namespace `json:"pid,omitempty"`
```
mheon, line 77: This should only be in the infra container configuration, no need to duplicate

libpod/pod.go
```diff
@@ -170,6 +174,16 @@ func (p *Pod) CPUQuota() int64 {
 	return 0
 }
 
+// Pid returns the pod process id namespace value
```
mheon, line 177: This isn't getting the PID, but rather the path of the infra container's PID file?

It's also not really related to the PID namespace

libpod/pod_api.go
```diff
@@ -541,6 +541,9 @@ func (p *Pod) Inspect() (*define.InspectPodData, error) {
 		infraConfig.CPUPeriod = p.CPUPeriod()
 		infraConfig.CPUQuota = p.CPUQuota()
 		infraConfig.CPUSetCPUs = p.ResourceLim().CPU.Cpus
+		if p.config.Pid.NSMode != "" {
+			infraConfig.Pid = p.Pid()
```
mheon, line 545: Why is this needed?

libpod/runtime_pod_infra_linux.go
```diff
@@ -145,6 +146,34 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
 		if len(p.config.InfraContainer.ExitCommand) > 0 {
 			options = append(options, WithExitCommand(p.config.InfraContainer.ExitCommand))
 		}
+		if p.config.Pid.NSMode != "private" && p.config.Pid.NSMode != "" {
+			path := p.config.Pid.Value
+			ns := spec.LinuxNamespace{}
+			pidVal := p.config.Pid.NSMode
+			if pidVal != "" {
+				if pidVal == "container" {
```
mheon, line 154: This validation should be done in the option (`WithPid()` so we only have to do it once and we can reject invalid configurations at creation time, not runtime.

libpod/runtime_pod_infra_linux.go
```diff
@@ -145,6 +146,34 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
...
+					err := errors.New("Invalid Argument")
+					return nil, errors.Wrap(err, "Cannot take container in a different NS as an argument")
+				}
+				if pidVal == "ns" {
```
mheon, line 158: What is this if statement doing? The PID namespace and PID file are not related.

libpod/runtime_pod_infra_linux.go
```diff
@@ -145,6 +146,34 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
...
+					path = p.config.Pid.Value
+					_, err := os.Stat("path")
+					if err != nil {
+						return nil, err
+					}
+					options = append(options, WithPidFile(path))
+				}
+			}
+			if pidVal != "host" {
```
mheon, line 167: I think you probably want to structure this as a `switch` with a blog for each valid option

libpod/runtime_pod_infra_linux.go
```diff
@@ -145,6 +146,34 @@ func (r *Runtime) makeInfraContainer(ctx context.Context, p *Pod, imgName, rawIm
...
+				ns.Type = "pid"
+				ns.Path = path
+				g.AddOrReplaceLinuxNamespace(string(ns.Type), ns.Path)
+			} else if pidVal == "host" {
+				g.Config.Linux.Namespaces = []spec.LinuxNamespace{}
```
mheon, line 172: This wipes out *all* namespaces, not just the PID namespace. You need to loop through namespaces and make a new array containing everything except the PID namespace.

libpod/runtime_pod_linux.go
```diff
@@ -116,6 +116,10 @@ func (r *Runtime) NewPod(ctx context.Context, options ...PodCreateOption) (_ *Po
 	if pod.config.UsePodCgroup {
 		logrus.Debugf("Got pod cgroup as %s", pod.state.CgroupPath)
 	}
+
+	//	if !pod.state.Pid.IsDefault(){
```
mheon, line 120: This looks like it can be removed

pkg/domain/entities/pods.go
```diff
@@ -146,6 +147,17 @@ func (p *PodCreateOptions) CPULimits() *specs.LinuxCPU {
 	return cpu
 }
 
+func setNamespaces(p *PodCreateOptions) (specgen.Namespace, error) {
```
mheon, line 150: Suggest renaming to "getPIDNamespace" or, alternatively, making this return a struct which identifies which namespace is which - right now this is only useful for setting the PID namespace, but eventually we'll end up with four-ish settable namespaces.

pkg/specgen/generate/pod_create.go
```diff
@@ -102,6 +102,13 @@ func createPodOptions(p *specgen.PodSpecGenerator, rt *libpod.Runtime) ([]libpod
 		options = append(options, libpod.WithInfraCommand(p.InfraCommand))
 	}
 
+	if !p.Pid.IsDefault() {
+		options = append(options, libpod.WithPid(p.Pid))
+		if p.Pid.NSMode == "pod" {
```
mheon, line 107: This bit doesn't seem necessary?

pkg/domain/entities/pods.go
```diff
@@ -178,6 +190,14 @@ func (p *PodCreateOptions) ToPodSpecGen(s *specgen.PodSpecGenerator) error {
 	s.NoManageHosts = p.Net.NoHosts
 	s.HostAdd = p.Net.AddHosts
 
+	pidR, err := setNamespaces(p)
```
mheon, line 193: Minor nit, but slightly more descriptive variable names would be nice - I don't really like having a function with `pid` and `pidR`

pkg/specgen/podspecgen.go
```diff
@@ -57,6 +57,8 @@ type PodBasicConfig struct {
 	// (e.g. `podman generate systemd --new`).
 	// Optional.
 	PodCreateCommand []string `json:"pod_create_command,omitempty"`
+	// Pid sets the process id namespace of the pod
```
mheon, line 60: Add a line underneath - "Optional (defaults to private if unset)". And this should mention that it sets the PID namespace of the infra container, that will then be shared with the entire pod if PID namespace sharing is enabled.

libpod/options.go
cdoern: > We need to require that `UsePodPID` is set in the pod configuration - we have to be sharing the pod PID namespace for this to be set.

Should I require it by checking and erroring out if not set, or warn the user and set it for them?

libpod/pod.go
cdoern: Yeah I was unsure here of what would be helpful here @mheon, because the namespace just contains `pod` or `host` or whatever the user passed or a ns file if specified. What would this be expected to return?

libpod/pod_api.go
cdoern: If I can fix `Pid()` to return the proper value I guess it won't be but right now it is never "empty" so the file path is always in inspect even when user didn't specify `--pid`

libpod/options.go
mheon: Good question. I was leaning towards just erroring out unless the user passes `--share` as well to specify that the PID namespace is shared, but that seems like a bad user experience. @rhatdan Should setting `--pid` automatically cause the pod to share the PID namespace?

libpod/pod.go
mheon: Returning the mode of the namespace is what I would expect - we want to tell the user how the pod was configured. The PID of the pod's init container isn't that valuable.

libpod/runtime_pod_infra_linux.go
cdoern: Okay @mheon , to streamline this and preserve the originally passed values I might make another InfraContainerConfig value that is an actual `specs.LinuxNameSpace{}`. This will allow me to be able to print the proper mode in podInspect but also not have to parse anything again in `runtime_pod_infra_linux.go`

libpod/options.go
```diff
@@ -2397,3 +2400,32 @@ func WithPodCPUSetCPUs(inp string) PodCreateOption {
...
+				err := errors.New("Invalid Argument")
+				return errors.Wrap(err, "Cannot take container in a different NS as an argument")
+			case "path":
+				fmt.Println("hello")
+				ns.Path = inp.Value
+			case "private":
+				ns.Type = spec.PIDNamespace
+				ns.Path = ""
+			}
+			fmt.Println(ns)
```
mheon, line 2424: Extra debug

libpod/options.go
```diff
@@ -2397,3 +2400,32 @@ func WithPodCPUSetCPUs(inp string) PodCreateOption {
...
 }
+
+func WithPodPidNS(inp specgen.Namespace) PodCreateOption {
+	return func(p *Pod) error {
+		if p.valid {
+			return define.ErrPodFinalized
+		}
+		if p.config.UsePodPID {
+			ns := specs.LinuxNamespace{}
+			ns.Type = "pid"
```
mheon, line 2411: Is there a reason we need to recreate the namespace?

libpod/options.go
```diff
@@ -2397,3 +2400,32 @@ func WithPodCPUSetCPUs(inp string) PodCreateOption {
...
+			fmt.Println(inp)
+			switch inp.NSMode {
+			case "container":
+				err := errors.New("Invalid Argument")
```
mheon, line 2415: use define.ErrInvalidArg instead

libpod/options.go
cdoern: I think `specs.LinuxNameSpace{}` is the type that the infra generator takes versus `specgen.NameSpace`, we need to switch over at some point and this is behind the scenes enough.

libpod/options.go
cdoern: whoops!

libpod/options.go
mheon: Ahhh, I missed the type change. Hm. My concern would be forwards-compat - we need to make sure that pods from before this change, without a valid namespace in the DB, continue to work.

cdoern: well,  the `specgen.ParseNamespace()` function I use in pkg/domain/entities/pods.go takes the type specgen.NameSpace so the actual validation and parsing relies more on the specgen version this is just a conversion. This switch will only execute when someone explicitly says `--pid=...` so would forwards-compat be a problem?

cdoern:
> What does "pod" mean? Are you allowing pod's to join other pods pid namespace?

pod means that the containers will default to using the pod's pid ns which is similar to private (the default). I added this because the `specgen.parsenamespace()` function utilizes this value so we are going to accept.

libpod/pod.go
```diff
@@ -97,6 +98,8 @@ type InfraContainerConfig struct {
 	HasInfraContainer  bool                  `json:"makeInfraContainer"`
 	NoNetwork          bool                  `json:"noNetwork,omitempty"`
 	HostNetwork        bool                  `json:"infraHostNetwork,omitempty"`
+	Pid                specgen.Namespace     `json:"infraPid,omitempty"`
+	LinuxPidNS         specs.LinuxNamespace  `json:"LinuxPIDNS,omitempty"`
```
mheon, line 102: Do we need both of these? I would imagine we can get the mode just by looking at the `specs.LinuxNamespace` - or, alternatively, generate the `specs.LinuxNamespace` from the `specgen.Namespace` when creating the infra container?

cdoern: @mheon I kept both of them in order to parse the Pid NS before runtime. The types are different as `specgen.Namespace` contains info like `host` or `pod` which I do use again at runtime but `specs.LinuxNamespace` is the actual NS struct containing `pid` and the proper path. Not sure if I can parse before runtime and not have both of these but I will look.

cdoern: fixed

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -120,6 +120,14 @@ Add a DNS alias for the container. When the container is joined to a CNI network
 
 Disable creation of /etc/hosts for the pod.
 
+#### **--pid**=*pid*
+
+Set the PID mode for the pod Default is to create a private PID namespace for the pod. Requires PID NS to be shared.
```
TomSweeneyRedHat, line 125: I'm having a hard time parsing this.  Maybe?
```suggestion
Set the PID mode for the pod.  The default is to create a private PID namespace for the pod. Requires the PID namespace to be shared.
```

docs/source/markdown/podman-pod-create.1.md
```diff
@@ -120,6 +120,14 @@ Add a DNS alias for the container. When the container is joined to a CNI network
 
 Disable creation of /etc/hosts for the pod.
 
+#### **--pid**=*pid*
```
TomSweeneyRedHat, line 123: I was thinking "process ID (pid) when reading this.  I'll let others over-rule me, but I'd rather the option be named something else, in case we ever manipulate the pid later down the road.  Perhaps `--pid-mode`, regardless, at least  change the variable:
```suggestion
#### **--pid**=*pid-mode*
```

libpod/pod.go
TomSweeneyRedHat: Similar comment here for me, I'd change the variable to indicate it's pidmode and not the pid to help make the support of this more clear months down the line.

docs/source/markdown/podman-pod-create.1.md
cdoern: I feel like `--pid-ns` might be more accurate if anything @TomSweeneyRedHat. Only thing is, [container create](http://docs.podman.io/en/latest/markdown/podman-create.1.html#pid-pid) calls it `--pid`

libpod/pod.go
cdoern: good idea @TomSweeneyRedHat. I will change to make it more explicit that it returns the mode passed by the user.

docs/source/markdown/podman-pod-create.1.md
TomSweeneyRedHat: Ugh.  They should be the same in both places, and it looks like Docker has gone with `--pid` too, so we probably ought to stay with `--pid`.  Rather than `*pid-mode*` at line 123, I'd be fine with `*pid-ns*` if you prefer for the variable

cdoern: okay @TomSweeneyRedHat I am good with keeping it as `--pid`, probably best so people don't get confused as to what this flag is compared to container create and docker

libpod/pod.go
```diff
@@ -97,6 +98,7 @@ type InfraContainerConfig struct {
 	HasInfraContainer  bool                  `json:"makeInfraContainer"`
 	NoNetwork          bool                  `json:"noNetwork,omitempty"`
 	HostNetwork        bool                  `json:"infraHostNetwork,omitempty"`
+	Pid                specgen.Namespace     `json:"infraPid,omitempty"`
```
vrothberg, line 101: I think this should be called `PidNS` to prevent mistaking it for a process ID.  Same at other places in the code.

But I do agree that `--pid` makes sense on the CLI to remain consistent with `container create`.

test/e2e/pod_create_test.go
```diff
@@ -559,4 +559,37 @@ ENTRYPOINT ["sleep","99999"]
...
+		podCreate = podmanTest.Podman([]string{"pod", "create", "--pid", ns, "--share", "pid"})
+		podCreate.WaitWithDefaultTimeout()
+		Expect(podCreate.ExitCode()).To(Equal(0))
+
+		ns = "container:randomfakeid"
+
+		podCreate = podmanTest.Podman([]string{"pod", "create", "--pid", ns, "--share", "pid"})
+		podCreate.WaitWithDefaultTimeout()
+		Expect(podCreate).Should(ExitWithError())
+
```
vrothberg, line 593: Will the pid ns now be displayed in `podman inspect`? If so, it would be nice to exercise that here in the test.

vrothberg (commented):
LGTM in general. Just two comments.

test/e2e/pod_create_test.go
cdoern: @vrothberg we decided to output the `NSMode` the user gave us as it is more helpful, aka: `path`, `pod`, `private`... I will check for that here, thanks for calling that out!

cdoern:
sorry had to re-push to fix a test, ran locally and all green. should be good now @baude

Issue #10302, opened 2021-05-10
podman run --device-cgroup-rule option is not honored


vikas-goel:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When `--device-cgroup-rule=b 7:* rmw` option is passed to `podman run` command along with `CAP_MKNOD` capability, the container is expected to be able to setup loop devices. However, podman is not honoring the option. The same option works with Docker.

**Steps to reproduce the issue:**

1. Start a podman container with `--device-cgroup-rule=b 7:* rmw --device /dev/loop-control:/dev/loop-control:rwm --cap-add CAP_MKNOD` option.

2. Log into the container (`podman exec -it <container-name> bash`)

3. Create a virtual block device using `dd` command and apply file-system using `mkfs -t xfs`

4. Create a loop device using `mknod` command

5. Attach the virtual block device using `losetup` command

**Describe the results you received:**
`losetup` fails with "failed to set up loop device: Operation not permitted" error.
```
[root@flex-vm-02 ~]# podman container inspect tme-mas-02 | grep -i device-cgroup-rule
                "--device-cgroup-rule=b 7:* rmw",
[root@flex-vm-02 ~]# podman exec -it tme-mas-02 bash
bash-4.2# dd if=/dev/zero of=/mnt/nblogs/vg.img count=208896
208896+0 records in
208896+0 records out
106954752 bytes (107 MB) copied, 0.412903 s, 259 MB/s
bash-4.2# mkfs -t xfs /mnt/nblogs/vg.img
meta-data=/mnt/nblogs/vg.img     isize=512    agcount=4, agsize=6528 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=0, sparse=0
data     =                       bsize=4096   blocks=26112, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
log      =internal log           bsize=4096   blocks=855, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
bash-4.2# losetup -f
/dev/loop1
bash-4.2# mknod /dev/loop1 b 7 1
bash-4.2# mkdir /mnt/vg
bash-4.2# losetup -f /mnt/nblogs/vg.img
[...]

PR #10895, opened 2021-07-09
Support DeviceCgroupRules to actually get added.

rhatdan:
Fixes: https://github.com/containers/podman/issues/10302

### DIFF
--- a/pkg/specgen/specgen.go
+++ b/pkg/specgen/specgen.go
@@ -239,6 +239,9 @@ type ContainerStorageConfig struct {
 	// Devices are devices that will be added to the container.
 	// Optional.
 	Devices []spec.LinuxDevice `json:"devices,omitempty"`
+	// DeviceCGroupRule are device cgroup rules that allow containers
+	// to use additional types of devices.
+	DeviceCGroupRule []spec.LinuxDeviceCgroup `json:"device_cgroup_rule,omitempty"`
 	// IpcNS is the container's IPC namespace.
 	// Default is private.
 	// Conflicts with ShmSize if not set to private.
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
Lots of Red test unhappiness.
The code LGTM

pkg/specgen/specgen.go
```diff
@@ -239,6 +239,9 @@ type ContainerStorageConfig struct {
 	// Devices are devices that will be added to the container.
 	// Optional.
 	Devices []spec.LinuxDevice `json:"devices,omitempty"`
+	// DeviceCGroupRule are device cgroup rules that allow containers
+	// to use additional types of devices.
+	DeviceCGroupRule []string `json:"device_cgroup_rule,omitempty"`
```
mheon, line 244: Can we do the parsing on the frontend, so we can pass a struct here (including the type, path, and access permissions)?

rhatdan: That would require all users of the API to do the processing. I believe this is something that Docker supports as well.
WDYT @jwhonce @baude

mheon: Generally speaking, we try to parse in the frontend to the greatest extent possible, so the server does not have to do much validation of the contents of specgen.

TomSweeneyRedHat:
LGTM
But I'd like @mheon to eyeball and merge if his concern has been answered.

cdoern:
LGTM, just a general note: wondering if in the future we should look to combine all `[]spec.` elements into a singular entity since one spec entity should contain all of the things we are referencing?

PR #10910, opened 2021-07-12
Add support for checkpoint/restore into and out of pods

adrianreber:
This adds support to checkpoint containers out of pods and restore container into pods.

It is only possible to restore a container into a pod if it has been checkpointed out of pod. It is also not possible to restore a non pod container into a pod.

The main reason this does not work is the PID namespace. If a non pod container is being restored in a pod with a shared PID namespace, at least one process in the restored container uses PID 1 which is already in use by the infrastructure container. If someone tries to restore a container from a pod with a shared PID namespace without a shared PID namespace it will also fail because the resulting PID namespace will not have a PID 1.


libpod/container_internal.go
```diff
@@ -437,6 +437,7 @@ func (c *Container) setupStorage(ctx context.Context) error {
 		}
 		options.Flags["ProcessLabel"] = c.config.ProcessLabel
 		options.Flags["MountLabel"] = c.config.MountLabel
+		logrus.Debugf("before create container storage c.config.MountLabel %s", c.config.MountLabel)
```
mheon, line 440: Extra debug?

libpod/container_internal_linux.go
```diff
@@ -901,7 +901,25 @@ func (c *Container) addNamespaceContainer(g *generate.Generator, ns LinuxNS, ctr
 }
 
 func (c *Container) exportCheckpoint(options ContainerCheckpointOptions) error {
-	if len(c.Dependencies()) > 0 {
+	if len(c.Dependencies()) == 1 {
+		// Check if the dependency is an infra container. If it is we can checkpoint
+		// the container out of the Pod.
+		if c.config.Pod == "" {
+			return errors.Errorf("Cannot export checkpoints of containers with dependencies")
```
mheon, line 908: First letter should not be capitalized

libpod/container_internal_linux.go
```diff
@@ -901,7 +901,25 @@ func (c *Container) addNamespaceContainer(g *generate.Generator, ns LinuxNS, ctr
...
+
+		pod, err := c.runtime.state.Pod(c.config.Pod)
+		if err != nil {
+			return errors.Wrapf(err, "container %s is in pod %s, but pod cannot be retrieved", c.ID(), c.config.Pod)
+		}
+		if err := pod.updatePod(); err != nil {
+			return err
+		}
+		if c.Dependencies()[0] != pod.state.InfraContainerID {
+			return errors.Errorf("Cannot export checkpoints of containers with dependencies")
```
mheon, line 919: First letter should not be capitalized

libpod/container_internal_linux.go
```diff
@@ -901,7 +901,25 @@ func (c *Container) addNamespaceContainer(g *generate.Generator, ns LinuxNS, ctr
...
+		// the container out of the Pod.
+		if c.config.Pod == "" {
+			return errors.Errorf("Cannot export checkpoints of containers with dependencies")
+		}
+
+		pod, err := c.runtime.state.Pod(c.config.Pod)
+		if err != nil {
+			return errors.Wrapf(err, "container %s is in pod %s, but pod cannot be retrieved", c.ID(), c.config.Pod)
+		}
+		if err := pod.updatePod(); err != nil {
```
mheon, line 915: Need to lock the pod before this

libpod/container_internal_linux.go
```diff
@@ -901,7 +901,25 @@ func (c *Container) addNamespaceContainer(g *generate.Generator, ns LinuxNS, ctr
...
+			return err
+		}
+		if c.Dependencies()[0] != pod.state.InfraContainerID {
```
mheon, line 918: Probably easier to drop the updatePod and instead just call `pod.InfraContainerID()` to get the ID

libpod/container_internal_linux.go
```diff
@@ -1247,6 +1275,79 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
...
+		if err != nil {
+			return errors.Wrapf(err, "pod %q cannot be retrieved", options.Pod)
+		}
+
+		infraContainer, err := pod.InfraContainer()
+		if err != nil {
+			return errors.Wrapf(err, "cannot retrieved infra container from pod %q", options.Pod)
+		}
+
+		if err := infraContainer.syncContainer(); err != nil {
```
mheon, line 1291: Need to lock the infra container before this.

libpod/container_internal_linux.go
```diff
@@ -1247,6 +1275,79 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
...
+			return errors.Wrapf(err, "Error syncing infrastructure container %s status", infraContainer.ID())
+		}
+		if infraContainer.state.State != define.ContainerStateRunning {
+			if err := infraContainer.initAndStart(ctx); err != nil {
+				return errors.Wrapf(err, "Error starting infrastructure container %s status", infraContainer.ID())
+			}
+		}
+
+		if c.config.IPCNsCtr != "" {
```
mheon, line 1300: This isn't really matching the way we do pods - the pod itself stores what namespaces it shares. I would not expect good results if you restore a container from a pod that shared the PID namespace, into a pod that does not share the PID namespace.

mheon: Also, for all of these - you need to ensure that the container config field is set to the ID of the new pod's infra container. I think that the DB will not allow you to add a container to the DB if this isn't true, but we should make sure.

libpod/oci_conmon_linux.go
```diff
@@ -1064,6 +1064,43 @@ func (r *ConmonOCIRuntime) createOCIContainer(ctr *Container, restoreOptions *Co
...
+			// the runtime to change process and mount labels
+			pod, err := ctr.runtime.LookupPod(restoreOptions.Pod)
+			if err != nil {
+				return errors.Wrapf(err, "pod %q cannot be retrieved", restoreOptions.Pod)
+			}
+			infraContainer, err := pod.InfraContainer()
+			if err != nil {
+				return errors.Wrapf(err, "cannot retrieved infra container from pod %q", restoreOptions.Pod)
+			}
+			if err := infraContainer.syncContainer(); err != nil {
```
mheon, line 1078: Need to lock the infra container before doing this

mheon: Actually, do we need to do this at all? The Mount and Process label should not require a sync

libpod/container_internal_linux.go
```diff
@@ -1247,6 +1275,79 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
 		}
 	}
 
+	if options.Pod != "" {
```
mheon, line 1278: We need to make sure that the container config for the restored container has the ID of the new pod in it.

mheon:
I'm somewhat doubtful that checkpointing a container and restoring it into a different pod will work well - there's the potential for configuration differences that the container itself can't really detect but would result in major issues (a missing shared namespace being most notable)

libpod/container_internal_linux.go
adrianreber: I added checks to make sure only container are restored into Pods with the same namespace shared as the checkpointed container.

libpod/container_internal_linux.go
adrianreber: I think I solved also this.

adrianreber:
@mheon Thanks for your review. I added a check to only restore containers into a pod that has the same namespaces shared as the original pod out of which the container was checkpointed. I also updated all container metadata to no longer reference the original pod, but updated it with the new pod.

rhatdan:
Should we have a podman pod checkpoint and podman pod restore?

libpod/container_internal_linux.go
```diff
@@ -1247,6 +1276,82 @@ func (c *Container) restore(ctx context.Context, options ContainerCheckpointOpti
...
+		if err != nil {
+			return errors.Wrapf(err, "cannot retrieved infra container from pod %q", options.Pod)
+		}
+
+		infraContainer.lock.Lock()
+		if err := infraContainer.syncContainer(); err != nil {
+			infraContainer.lock.Unlock()
+			return errors.Wrapf(err, "Error syncing infrastructure container %s status", infraContainer.ID())
+		}
+		infraContainer.lock.Unlock()
```
mheon, line 1297: This should move down a bit, just under the `initAndStart()` on the infra container

adrianreber:
> Should we have a podman pod checkpoint and podman pod restore?

Yes, that was my goal. This is just the first step towards that.

libpod/container_internal_linux.go
adrianreber: done

test/e2e/checkpoint_test.go
```diff
@@ -977,4 +979,164 @@ var _ = Describe("Podman checkpoint", func() {
...
+			}
+			// Create a pod
+			session := podmanTest.Podman([]string{
+				"pod",
+				"create",
+				"--share",
+				share,
+			})
+			session.WaitWithDefaultTimeout()
+			Expect(session.ExitCode()).To(Equal(0))
```
edsantiago, line 1015: This pattern is anti-recommended, please see #10997. Could you fix it (and all instances) to use the recommended form? Thank you!

edsantiago (changes requested):
Recommended form is `Expect(session).To(Exit(0))`

test/e2e/checkpoint_test.go
adrianreber: done

libpod/oci_conmon_linux.go
adrianreber: I simplified this part.

vrothberg:
> Recommended form is `Expect(session).To(Exit(0))`

That's still open, after that we can merge.

adrianreber:
> > Recommended form is `Expect(session).To(Exit(0))`
> 
> That's still open, after that we can merge.

Oops, I thought I changed it. I will take a look later today.

adrianreber:
> > > Recommended form is `Expect(session).To(Exit(0))`
> > 
> > 
> > That's still open, after that we can merge.
> 
> Oops, I thought I changed it. I will take a look later today.

I pushed but did not commit previously. But now I committed and pushed.

edsantiago (approved):

vrothberg (commented):
LGTM, thank you @adrianreber!

Issue #10262, opened 2021-05-07
Data not copying up to volumes for containers created but not started


vikas-goel:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When volume and container with this volume are created by docker, that volume gets populated (copied up from the image) and can then be consumed by another container using `--volumes-from`. It does not seem to work with podman. With podman, the workaround is to run `podman container init` after creating the container as described in https://github.com/containers/podman/issues/8041.

Running `podman container init` gives error same as mentioned in https://github.com/containers/podman/issues/8041#issuecomment-710080732. The command, however, does the job.

I am using Docker Compose to create data volume container and then adding the same as a dependency for the service/application in the same Docker Compose configuration YAML file.

With Podman 3.0.2 and later version, I am using podman-docker package to manage my podman containers using Docker Compose. As the said podman command (podman container init) for the workaround is very specific to podman, Docker Compose does not have option to do the same.

Can we have a podman (global) configuration option, say `setup-volume-mount`, that can stay turned off by default. When the option is turned on, the ```podman create --name <name> <container-image>``` behaves same as corresponding `docker` command. That is, the podman should internally prepare the mount, i.e. copy the data without requiring additional commands (podman container init). The error described in https://github.com/containers/podman/issues/8041#issuecomment-710080732 should be ignored.

**Steps to reproduce the issue:**
Same as described in https://github.com/containers/podman/issues/8041

**Describe the results you received:**
Same as described in https://github.com/containers/podman/issues/8041

**Describe the results you expected:**
Same as described in https://github.com/containers/podman/issues/8041

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Version:      3.0.2-dev
API Version:  3.0.0
Go Version:   go1.15.7
Built:        Tue Mar  2 07:10:06 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.26-1.module+el8.4.0+10198+36d1d0e3.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.26, commit: 0a5175681bdd52b99f1f0f442cbba8f8c126a1c9'
  cpus: 8
  distribution:
    distribution: '"rhel"'
    version: "8.4"
  eventLogger: file
[...]

PR #10911, opened 2021-07-12
Copy the content from the underlying image into the newly created volume.

vikas-goel:
The copy happens if prepare_volume_on_create is set to true in containers.conf file.

Docker does this by default.

Fixes: #10262

Signed-off-by: Vikas Goel <vikas.goel@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,6 +66,6 @@ require (
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/sys v0.0.0-20210603125802-9665404d3644
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
-	k8s.io/api v0.21.2
+	k8s.io/api v0.21.3
 	k8s.io/apimachinery v0.21.3
 )
--- a/go.sum
+++ b/go.sum
@@ -1428,12 +1428,11 @@ honnef.co/go/tools v0.0.1-2020.1.4/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9
 k8s.io/api v0.20.1/go.mod h1:KqwcCVogGxQY3nBlRpwt+wpAMF/KjaCc7RpywacvqUo=
 k8s.io/api v0.20.4/go.mod h1:++lNL1AJMkDymriNniQsWRkMDzRaX2Y/POTUi8yvqYQ=
 k8s.io/api v0.20.6/go.mod h1:X9e8Qag6JV/bL5G6bU8sdVRltWKmdHsFUGS3eVndqE8=
-k8s.io/api v0.21.2 h1:vz7DqmRsXTCSa6pNxXwQ1IYeAZgdIsua+DZU+o+SX3Y=
-k8s.io/api v0.21.2/go.mod h1:Lv6UGJZ1rlMI1qusN8ruAp9PUBFyBwpEHAdG24vIsiU=
+k8s.io/api v0.21.3 h1:cblWILbLO8ar+Fj6xdDGr603HRsf8Wu9E9rngJeprZQ=
+k8s.io/api v0.21.3/go.mod h1:hUgeYHUbBp23Ue4qdX9tR8/ANi/g3ehylAqDn9NWVOg=
 k8s.io/apimachinery v0.20.1/go.mod h1:WlLqWAHZGg07AeltaI0MV5uk1Omp8xaN0JGLY6gkRpU=
 k8s.io/apimachinery v0.20.4/go.mod h1:WlLqWAHZGg07AeltaI0MV5uk1Omp8xaN0JGLY6gkRpU=
 k8s.io/apimachinery v0.20.6/go.mod h1:ejZXtW1Ra6V1O5H8xPBGz+T3+4gfkTCeExAHKU57MAc=
-k8s.io/apimachinery v0.21.2/go.mod h1:CdTY8fU/BlvAbJ2z/8kBwimGki5Zp8/fbVuLY8gJumM=
 k8s.io/apimachinery v0.21.3 h1:3Ju4nvjCngxxMYby0BimUk+pQHPOQp3eCGChk5kfVII=
 k8s.io/apimachinery v0.21.3/go.mod h1:H/IM+5vH9kZRNJ4l3x/fXP/5bOPJaVP/guptnZPeCFI=
 k8s.io/apiserver v0.20.1/go.mod h1:ro5QHeQkgMS7ZGpvf4tSMx6bBOgPfE+f52KwvXfScaU=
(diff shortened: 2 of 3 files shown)


libpod/runtime_ctr.go
```diff
@@ -460,6 +460,11 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (_ *Contai
 	} else if err := r.state.AddContainer(ctr); err != nil {
 		return nil, err
 	}
+
+	if err := ctr.prepare(); err != nil {
```
mheon, line 464: As discussed before, this *cannot* be the default. There will be significant performance implications from it.

Also, this must be following by a `ctr.cleanup()` to ensure we do not leak mountpoints and network configurations.

vikas-goel: I will add a `defer ctr.cleanup()` statement. After checking the `prepare()` code I thought calling it again will be a no-op whether it is invoked using `podman container init` or `podman run`.
For my knowledge, how will be create performance implications?

mheon: This adds another mount and unmount to the early-init code of Podman, plus a setup and teardown of the full CNI network stack. These are the two most expensive operations during container creation, and forcing the `podman run` command to do them twice is going to significantly increase runtime in typical cases.

The core issue here is that not all calls to container creation are `podman run`. It is perfectly legal for someone to `podman create` a container, wait a long time (years, even), and then `podman start` said container. We cannot keep the container's mounts and network namespace alive during this time - the end result of container creation must be a container with no resources consumed. This is why we must clean up after the initial mount of the container when doing this copy-up, and why it cannot be the default.

vikas-goel: > the end result of container creation must be a container with no resources consumed

Thank you @mheon . Agree with this.

libpod/runtime_ctr.go
```diff
@@ -460,6 +460,25 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (_ *Contai
...
+
+	// Prepare container for volume copy-up if user opted to.
+	if r.config.Containers.PrepareOnCreate {
+		// A data volume container may not have executable file.
+		// Calling init() will result in failure due to executable file
+		// check. Instead, call prepare() to avoid the failure.
+		defer func() {
+			if retErr != nil {
+				if err := ctr.cleanup(ctx); err != nil {
+					logrus.Errorf("error cleaning up container %s: %v", ctr.ID(), err)
```
cdoern, line 472: we should probably return this error instead of simply logging it.

cdoern (changes requested):

libpod/runtime_ctr.go
vikas-goel: This is happening in a `defer` call where the dependency is on the function return error `retErr`. This piece of code handling is similar to how is handled elsewhere. I got this code from `libpod/container_internal.go`
```
 249 func (c *Container) handleRestartPolicy(ctx context.Context) (_ bool, retErr error) {
        ...
 277         defer func() {
 278                 if retErr != nil {
 279                         if err := c.cleanup(ctx); err != nil {
 280                                 logrus.Errorf("error cleaning up container %s: %v", c.ID(), err)
 281                         }
 282                 }
 283         }()
 284         if err := c.prepare(); err != nil {
 285                 return false, err
 286         }
 287
```

libpod/runtime_ctr.go
```diff
@@ -460,6 +460,25 @@ func (r *Runtime) setupContainer(ctx context.Context, ctr *Container) (_ *Contai
 	} else if err := r.state.AddContainer(ctr); err != nil {
 		return nil, err
 	}
+
+	// Prepare container for volume copy-up if user opted to.
+	if r.config.Containers.PrepareOnCreate {
```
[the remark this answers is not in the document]
vikas-goel: WIP https://github.com/containers/common/pull/676

Luap99 (commented):
I think we should split the network setup out from this, you only need storage. CNI setup is expensive. I also think in case of slirp4netns it will leak the process. cleanup() will not kill the slirp process.
Does `c.mountStorage()` and `c.cleanupStorage()` works for you?

vikas-goel:
> I think we should split the network setup out from this, you only need storage. CNI setup is expensive. I also think in case of slirp4netns it will leak the process. cleanup() will not kill the slirp process.
> Does `c.mountStorage()` and `c.cleanupStorage()` works for you?

That will work. Thank you, @Luap99 . I had looked up that option earlier but thought there was some complexity. Not sure why. Checking again after your comment, it seems to be fine.

Currently, I am blocked and waiting for someone to merge the https://github.com/containers/common/pull/678 PR before I can make this change.

vikas-goel:
/c/common

vikas-goel:
Hi @mheon ,
| The updated c/common will have to be vendored into Podman to be used once this is merged.

I am not familiar with the command/steps. Could you help me on this one?

mheon:
In the top-level directory for the Podman git repo, run:
`go get github.com/containers/common@$COMMITHASH` (where $COMMITHASH is the hash of the commit you want to vendor)

Then run `make vendor-in-container` to finalize the changes and commit the result.

Please note that you cannot have symlinks in the path of your working directory when you run these - has to be an absolute path or the commands will misbehave.

vikas-goel:
Thank you @mheon . Looks like a new version of common is not needed per your instruction.
The build is failing for something in containers/buildah

mheon:
I think it is asking for Common - from the tests:
```
libpod/runtime_ctr.go:466:24: r.config.Containers.PrepareVolumeOnCreate undefined (type "github.com/containers/common/pkg/config".ContainersConfig has no field or method PrepareVolumeOnCreate)
```

vikas-goel:
> I think it is asking for Common - from the tests:
> 
> ```
> libpod/runtime_ctr.go:466:24: r.config.Containers.PrepareVolumeOnCreate undefined (type "github.com/containers/common/pkg/config".ContainersConfig has no field or method PrepareVolumeOnCreate)
> ```

Correct. That I comment I made was for vendor PR failure.

vikas-goel:
Now that vendor c/common is done, could someone add `/ok-to-test` message or start a `/retest`? It requires a "trusted user" review.

mheon:
I think you need to rebase on top of latest main and force-push - that will pick up the change and force CI to rerun

vikas-goel:
@mheon , @rhatdan ,
Do we have a data volume container image in the pipeline to that I can use to write a test case?

Or, do you think creating a configuration file with `prepare_volume_on_create = true` and creating a container will be good enough?

mheon:
Copy-up triggers on any volume, including volumes from `-v` - our general test is to use `-v` to bind-mount a host directory over a directory in the container that we know contains files.

Issue #10908, opened 2021-07-12
podman-remote build, with .containerignore, should (but does not) skip .dockerignore


edsantiago:
podman and buildah, if they see `.containerignore`, will read it and only it. `podman-remote`, OTOH, will also look for and read `.dockerignore`:
```console
mkdir zzz;cd zzz;printf "FROM quay.io/libpod/testimage:20210610\nCOPY ./ ./\n" >Dockerfile
$ echo 'test1*' >.containerignore
$ echo 'test2*' >.dockerignore
$ touch test1 test2
$ ../bin/podman build -q -t foo .;../bin/podman run --rm foo ls
700dae51df0fecf3cf2fbc59343c36f505a5ff87074494180d59dfb258faa9f9
Dockerfile
pause
test2            <---- good. .dockerfile was ignored
testimage-id

$ ../bin/podman-remote build -q -t foo .;../bin/podman run --rm foo ls
707107350bb1bcc1f7b4b29ca857f2b5649a5ab962e4da26cc2696b6100b9b36
Dockerfile
pause                <--- BAD: test2 should be listed here
testimage-id
```

Issue #10907, opened 2021-07-12
podman-remote build: complicated .dockerignore bug


edsantiago:
A confusing mix of wildcards in `.dockerignore` does not work the way it should in podman-remote:
```console
$ mkdir zzz;cd zzz
$ printf "FROM quay.io/libpod/testimage:20210610\nCOPY ./ ./\nCOPY subdir ./\n" >Dockerfile
$ printf "*\nsubdir\n\!*/sub1*\n" >.dockerignore
$ mkdir subdir; touch subdir/sub1.txt
$ ../bin/podman-remote build .
STEP 1/3: FROM quay.io/libpod/testimage:20210610
STEP 2/3: COPY ./ ./
--> c3555113657
STEP 3/3: COPY subdir ./
COMMIT
--> 5f9f29bd454
5f9f29bd45409e016dfccdea718348e8788c6e0d4de9ed3d35920a3c70c93dc3
```
It succeeds, but should not. `podman` (non-remote) and buildah fail with:
```
STEP 3/3: COPY subdir ./
Error: error building at STEP "COPY subdir ./": no items matching glob "/home/esm/src/atomic/2018-02.podman/libpod/zzz/subdir" copied (1 filtered out): no such file or directory
```

PR #10913, opened 2021-07-12
 podman-remote build use .containerignore over .dockerignore 

rhatdan:
$ mkdir zzz;cd zzz
$ printf "FROM quay.io/libpod/testimage:20210610\nCOPY ./ ./\nCOPY subdir ./\n" >Dockerfile
$ printf "*\nsubdir\n\!*/sub1*\n" >.dockerignore
$ mkdir subdir; touch subdir/sub1.txt
$ ../bin/podman-remote build .

Should fail, but succeeds because we are not sending over the
.dockerignore file to the server side.  This PR will send the
.dockerignore so the server side and use it.

Fixes: #10907

Also if both .containerignore and .dockerignore in the context
directory, podman-remote should prefer .containerignore and not use
.dockerignore.

Fixes: #10908

### DIFF
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -851,7 +851,7 @@ EOF
     run_podman rmi -f build_test
 }
 
-@test "podman build -f test " {
+@test "podman build -f test" {
     tmpdir=$PODMAN_TMPDIR/build-test
     subdir=$tmpdir/subdir
     mkdir -p $subdir
@@ -875,6 +875,44 @@ EOF
     run_podman build -t build_test -f $containerfile1 -f $containerfile2 $tmpdir
     is "$output" ".*$IMAGE" "Containerfile2 is also passed to server"
     run_podman rmi -f build_test
+}
+
+@test "podman build .dockerignore failure test" {
+    tmpdir=$PODMAN_TMPDIR/build-test
+    subdir=$tmpdir/subdir
+    mkdir -p $subdir
+
+    cat >$tmpdir/.dockerignore <<EOF
+*
+subdir
+!*/sub1*
+EOF
+    cat >$tmpdir/Containerfile <<EOF
+FROM $IMAGE
+COPY ./ ./
+COPY subdir ./
+EOF
+    run_podman 125 build -t build_test $tmpdir
+    is "$output" ".*Error: error building at STEP \"COPY subdir ./\"" ".dockerignore was ignored"
+}
+
+@test "podman build .containerignore and .dockerignore test" {
+    tmpdir=$PODMAN_TMPDIR/build-test
+    mkdir -p $tmpdir
+    touch $tmpdir/test1 $tmpdir/test2
+    cat >$tmpdir/.containerignore <<EOF
+test2*
+EOF
+    cat >$tmpdir/.dockerignore <<EOF
+test1*
+EOF
+    cat >$tmpdir/Containerfile <<EOF
+FROM $IMAGE
+COPY ./ /tmp/test/
+RUN ls /tmp/test/
+EOF
+    run_podman build -t build_test $tmpdir
+    is "$output" ".*test1" "test1 should exists in the final image"
 }
 
 function teardown() {
--- a/pkg/bindings/images/build.go
+++ b/pkg/bindings/images/build.go
@@ -302,7 +302,7 @@ func Build(ctx context.Context, containerFiles []string, options entities.BuildO
 	tarContent := []string{options.ContextDirectory}
 	newContainerFiles := []string{}
 
-	dontexcludes := []string{"!Dockerfile", "!Containerfile"}
+	dontexcludes := []string{"!Dockerfile", "!Containerfile", "!.dockerignore", "!.containerignore"}
 	for _, c := range containerFiles {
 		if c == "/dev/stdin" {
 			content, err := ioutil.ReadAll(os.Stdin)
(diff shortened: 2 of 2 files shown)


rhatdan:
@edsantiago PTAL

rhatdan:
Fixes: #10908

edsantiago:
I don't think this solves the problem; it actually breaks in a different way. I'll update later, after f2f.

edsantiago:
Yes, this doesn't seem to fix anything. The fault is mine: I got lazy in writing the reproducer, and wrote something that probably didn't create the correct dockerignore file because of shell expansions. Try this for your dockerignore:
```
$ cat >zzz/.dockerignore <<EOF
*
subdir
!*/sub1*
EOF
```
With that file, `podman build` fails (as it should) but `podman-remote build` succeeds (bad, #10907)

edsantiago:
That seems to have done it, thanks. Passes on my laptop. New CI test run [in progress](https://cirrus-ci.com/task/6334189655031808)

edsantiago:
LGTM! All dockerignore tests are [now passing](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6334189655031808/html/bud-remote-fedora-34-root-host.log.html). (Failures are the usual stream-dropped bug, #10154)

test/system/070-build.bats
```diff
@@ -887,4 +887,61 @@ function teardown() {
...
+COPY ./ ./
+COPY subdir ./
+EOF
+    run_podman 125 build -t build_test $tmpdir
+    is "$output" ".*Error: error building at STEP \"COPY subdir ./\": no items matching glob" "Expected to fail"
+}
+
+@test "podman build .containerignore and .dockerignore test " {
+    tmpdir=$PODMAN_TMPDIR/build-test
+    mkdir -p $subdir
```
edsantiago, line 911: I think you forgot: 
```
    subdir=$tmpdir/subdir
```

test/system/070-build.bats
```diff
@@ -887,4 +887,61 @@ function teardown() {
...
+    cat >$tmpdir/Containerfile <<EOF
+FROM $IMAGE
+COPY ./ ./test/
+RUN ls /tmp/test/
+EOF
+    run_podman build -t build_test $tmpdir
+    is "$output" ".*test2" "test2 should exists in the DockerfileExpected to fail"
+}
+
+@test "podman build .containerignore and .dockerignore test " {
```
edsantiago, line 928: Could you give this test a unique name? (It's the same as line 909)

edsantiago (changes requested):

test/system/070-build.bats
```diff
@@ -887,4 +887,61 @@ function teardown() {
...
+    cat >$tmpdir/.dockerignore <<EOF
+test2*
+EOF
+    cat >$tmpdir/Containerfile <<EOF
+FROM $IMAGE
+COPY ./ ./test/
+RUN ls /tmp/test/
+EOF
+    run_podman build -t build_test $tmpdir
+    is "$output" ".*test2" "test2 should exists in the DockerfileExpected to fail"
```
edsantiago, line 925: Oops, hit Submit too soon. Also, should be 'exist' (no ess). And, I think the "Expected to fail" is a copy-paste error?

test/system/070-build.bats
```diff
@@ -887,4 +887,61 @@ function teardown() {
...
+    cat >$tmpdir/.dockerignore <<EOF
+test1*
+EOF
+    cat >$tmpdir/Containerfile <<EOF
+FROM $IMAGE
+COPY ./ /tmp/test/
+RUN ls /tmp/test/
+EOF
+    run_podman build -t build_test $tmpdir
+    is "$output" ".*test1" "test1 should exists in the DockerfileExpected to fail"
```
edsantiago, line 944: same thing (exist, Expected to fail)

test/system/070-build.bats
rhatdan: Removed since it was a duplicate of prior.

test/system/070-build.bats
rhatdan: Fixed

test/system/070-build.bats
rhatdan: Fixed

test/system/070-build.bats
edsantiago: ^^

test/system/070-build.bats
```diff
@@ -877,6 +877,44 @@ EOF
...
+    cat >$tmpdir/.dockerignore <<EOF
+test1*
+EOF
+    cat >$tmpdir/Containerfile <<EOF
+FROM $IMAGE
+COPY ./ ./test/
+RUN ls /tmp/test/
+EOF
+    run_podman build -t build_test $tmpdir
+    is "$output" ".*test1" "test1   should exists in the final image" "Confirm only .containerignore is used"
```
edsantiago, line 915: Too many arguments to `is()`: the last one ("Confirm only...") is ignored. Could you pick one or the other between that and "test1 should exist"?

TomSweeneyRedHat:
All kinds of red test unhappiness @rhatdan

test/system/070-build.bats
```diff
@@ -887,4 +887,61 @@ function teardown() {
     basic_teardown
 }
 
+@test "podman build .dockerignore failure test " {
```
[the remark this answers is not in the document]
rhatdan: Not sure what you mean?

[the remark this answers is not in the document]
edsantiago: Could you remove the space between `failure test` and the closing double-quote?

Also, still two unresolved bugs from my previous feedback; I'll repeat them

test/system/070-build.bats
edsantiago: Again, this one seems to have slipped

edsantiago:
oh, you fixed them

edsantiago (approved):

edsantiago:
Looks like the podman-remote error is different:
```
#| expected: '.*Error: error building at STEP "COPY subdir ./": no items matching glob'
#|   actual:   'Error: error building at STEP "COPY subdir ./": checking on sources under "/var/tmp/libpod_builder446168272/build": copier: stat: "/subdir": no such file or directory'
```
My simpleminded workaround would be something like:
```bash
    expect_msg="no items matching glob"
    if is_remote; then expect_msg="checking on sources under .*/subdir.*no such file or directory";fi
```

edsantiago:
/lgtm

Thanks @rhatdan !

PR #10915, opened 2021-07-12
[3.2] Backport: Support uid,gid,mode options for secrets

ashley-cui:
This somehow slipped from 3.2, but it contains a bugfix for users to be able to read secrets, so I think it should be backported?

Support UID, GID, Mode options for mount type secrets. Also, change
default secret permissions to 444 so all users can read secret.


rhatdan:
Not sure if we are shipping a 3.2.3, but this LGTM
/lgtm
/approve

Issue #10935, opened 2021-07-15
podman start --attach, a container created with --rm, does not rm the container


edsantiago:
...at least, not immediately. There's a race condition, and the following test is actually broken:
https://github.com/containers/podman/blob/4c38202bd1f79e0c6dc4d5533e3412a3761192c4/test/e2e/start_test.go#L67-L78

Trivial reproducer:
```console
$ while :;do cid=$(./bin/podman create --rm quay.io/libpod/testimage:20210610 foo);./bin/podman start --attach $cid;./bin/podman container exists $cid && break;done
```
Will spit out a few iterations of errors, then break (because `podman container exists` succeeded).

Seems to be root & rootless

vrothberg:
I am taking a look. Thanks for the reproducer, @edsantiago !

vrothberg:
Small fix: `s/foo/ls/` in the reproducer.  `foo` will bark immediately as it doesn't exist on the image.

edsantiago:
That's the point -- I was going with a reproducer of the exact same e2e test

vrothberg:
For `ls` I think the race is natural; only `run` will wait for the container to be cleaned up.

My gut feeling is that this should be true for `foo` as well; should podman wait if the command succeeded (if so, for how long)?  But I'll have a look at the git history and see what Docker does.

vrothberg:
Opened https://github.com/containers/podman/pull/10942.

PR #10942, opened 2021-07-15
podman start: remove containers configured for auto removal

vrothberg:
Make sure that containers configured for auto removal
(e.g., via `podman create --rm`) are removed in `podman start`
if starting the container failed.

Fixes: #10935

### DIFF
--- a/pkg/domain/infra/abi/containers.go
+++ b/pkg/domain/infra/abi/containers.go
@@ -261,6 +261,24 @@ func (ic *ContainerEngine) ContainerRestart(ctx context.Context, namesOrIds []st
 	return reports, nil
 }
 
+func (ic *ContainerEngine) removeContainer(ctx context.Context, ctr *libpod.Container, options entities.RmOptions) error {
+	err := ic.Libpod.RemoveContainer(ctx, ctr, options.Force, options.Volumes)
+	if err == nil {
+		return nil
+	}
+	logrus.Debugf("Failed to remove container %s: %s", ctr.ID(), err.Error())
+	switch errors.Cause(err) {
+	case define.ErrNoSuchCtr:
+		if options.Ignore {
+			logrus.Debugf("Ignoring error (--allow-missing): %v", err)
+			return nil
+		}
+	case define.ErrCtrRemoved:
+		return nil
+	}
+	return err
+}
+
 func (ic *ContainerEngine) ContainerRm(ctx context.Context, namesOrIds []string, options entities.RmOptions) ([]*entities.RmReport, error) {
 	reports := []*entities.RmReport{}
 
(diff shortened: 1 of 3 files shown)


pkg/domain/infra/abi/containers.go
```diff
@@ -261,6 +261,24 @@ func (ic *ContainerEngine) ContainerRestart(ctx context.Context, namesOrIds []st
...
+	}
+	logrus.Debugf("Failed to remove container %s: %s", ctr.ID(), err.Error())
+	switch errors.Cause(err) {
+	case define.ErrNoSuchCtr:
+		if options.Ignore {
+			logrus.Debugf("Ignoring error (--allow-missing): %v", err)
+			return nil
+		}
+	case define.ErrCtrRemoved:
+		return nil
```
cdoern, line 277: if the container is already removed should we log to user/return an error instead of nil?

pkg/domain/infra/abi/containers.go
```diff
@@ -838,6 +844,11 @@ func (ic *ContainerEngine) ContainerStart(ctx context.Context, namesOrIds []stri
 				}
 				report.Err = errors.Wrapf(err, "unable to start container %q", ctr.ID())
 				reports = append(reports, report)
+				if ctr.AutoRemove() {
+					if err := ic.removeContainer(ctx, ctr, entities.RmOptions{}); err != nil {
```
cdoern, line 848: just thinking, should we default to `--ignore` in the `RmOptions{}` since either way we log that the container isn't in existence? Maybe change the log to explicitly say the container doesn't exist but ignoring.

pkg/domain/infra/abi/containers.go
rhatdan: If I ask for a container to be removed and it is removed, then do I care.  Let's not add noice, when there is a potential race condition.  If we want to log this, then do it at the Debug level.

pkg/domain/infra/abi/containers.go
rhatdan: SGTM

pkg/domain/infra/abi/containers.go
vrothberg: I think we're cool as it, mostly because removal is racy; e.g., with the cleanup process.

pkg/domain/infra/abi/containers.go
vrothberg: `--ignore` is useful when doing  a `podman container rm --ignore notSureExists`.  But in this case, we know that the container exists. So I think that setting `RmOptions.Ignore = true` may suppress otherwise valid errors - in this specific case.

rhatdan:
LGTM

mheon:
LGTM

edsantiago:
If for some reason you need to resubmit, could you remove this skip?

https://github.com/containers/podman/blob/main/test/e2e/start_test.go#L68

vrothberg:
> If for some reason you need to resubmit, could you remove this skip?
> 
> https://github.com/containers/podman/blob/main/test/e2e/start_test.go#L68

Done :checkered_flag:

edsantiago:
Test failures look real. Is there a separate code path for podman-remote?

vrothberg:
```
[+0510s]   Expected
[+0510s]       <int>: 0
[+0510s]   to be greater than exit code:
[+0510s]       <int>: 0
```

:thinking:

vrothberg:
> Test failures look real. Is there a separate code path for podman-remote?

Yes, the remote client missed the error code which this test was exercising.  Should be good now; at least worked locally.

PR #10947, opened 2021-07-15
System tests: fix a multiarch problem

edsantiago:
The multi-image load test was failing on non-x86_64 arch,
because the images used by the test (:00000000, :20200902)
did not have manifests for the given arch.

Solution: all we need are two nonlocal images. Use the
predefined NONLOCAL_IMAGE for one, and a new :multiimage
tag (manually created, currently == :20210610) for the other.
Document, so if/when RHEL adds new supported arches, the
test will fail but a maintainer will have a clue what to do.

Also, as long as I'm in here: add 'image prune -f' to teardown
in build.bats, to avoid seeing lots of red "stray image"
warnings in test logs.

### DIFF
--- a/test/python/docker/compat/test_containers.py
+++ b/test/python/docker/compat/test_containers.py
@@ -206,6 +206,7 @@ class TestContainers(unittest.TestCase):
         self.assertEqual(len(ctnrs), 1)
 
     def test_copy_to_container(self):
+        self.skipTest("FIXME: #10948 - test is broken")
         ctr: Optional[Container] = None
         try:
             test_file_content = b"Hello World!"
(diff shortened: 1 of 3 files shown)


edsantiago:
Fixes: https://gitlab.com/cki-project/kernel-tests/-/issues/609

edsantiago:
Important heads-up: I've taken the liberty of disabling the flaky APIv2 test. Issue is tracked in #10948, whoever fixes the flake can reenable the test.

rhatdan:
LGTM

vrothberg (approved):
/lgtm

Issue #10948, opened 2021-07-15
APIv2 test: test_copy_to_container: False is not true


edsantiago:
I'm seeing this flake on almost every PR:
```
[+0215s] ======================================================================
[+0215s] FAIL: test_copy_to_container (compat.test_containers.TestContainers)
[+0215s] ----------------------------------------------------------------------
[+0215s] Traceback (most recent call last):
[+0215s]   File "/var/tmp/go/src/github.com/containers/podman/test/python/docker/compat/test_containers.py", line 232, in test_copy_to_container
[+0215s]     self.assertTrue(out.startswith(b'1042:1043'), "assert correct uid/gid")
[+0215s] AssertionError: False is not true : assert correct uid/gid
[+0215s] 
[+0215s] ----------------------------------------------------------------------
```
For example: [here](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5246553477611520/html/apiv2-podman-fedora-34-root-host.log.html) 

When fixed, could you please make the error message a little bit more helpful than "false is not true"?

edsantiago:
@matejvasek PTAL

matejvasek:
@edsantiago I'll look into it, my guess is that there is something before expected string "1042:1043". Actually the output should only contain "1042:1043" and nothing else, but when running in CI there was some superfluous output so I changed exact match to `startswith`. If it's the case changing it to `contains` would help.

edsantiago:
Here is [useful data](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5795692124831744/html/apiv2-podman-fedora-34-root-host.log.html). The failure is:
```
[+0218s] ======================================================================
[+0218s] FAIL: test_copy_to_container (compat.test_containers.TestContainers)
[+0218s] ----------------------------------------------------------------------
[+0218s] Traceback (most recent call last):
[+0218s]   File "/var/tmp/go/src/github.com/containers/podman/test/python/docker/compat/test_containers.py", line 232, in test_copy_to_container
[+0218s]     self.assertEqual(out.rstrip(), b'1042:1043')
[+0218s] AssertionError: b'1042:1043\n[conmon:d]: exec with attach is waiting[86 chars]rent' != b'1042:1043'
[+0218s] 
[+0218s] ======================================================================
[+0218s] FAIL: test_mount_preexisting_dir (compat.test_containers.TestContainers)
[...]

PR #10949, opened 2021-07-15
Better assert for test

matejvasek:
resolves #10948

Signed-off-by: Matej Vasek <mvasek@redhat.com>

PR #10950, opened 2021-07-15
APIv2 (python) tests: fix flake

edsantiago:
Python tests were flaking because they behave differently
when $DEBUG is set. It looks like something in CI sets
that envariable.

Solution: do not use $DEBUG as a debug trigger, use a
properly-named custom variable that is unlikely to be
set accidentally.

Also: get rid of AssertTrue(), which gives no visibility
into what happened. Write in proper form that can emit
useful diagnostics on failure.

Fixes: #10948

### DIFF
--- a/test/python/docker/__init__.py
+++ b/test/python/docker/__init__.py
@@ -27,7 +27,7 @@ class Podman(object):
         # No support for tmpfs (/tmp) or extfs (/var/tmp)
         # self.cmd.append("--storage-driver=overlay")
 
-        if os.getenv("DEBUG"):
+        if os.getenv("PODMAN_PYTHON_TEST_DEBUG"):
             self.cmd.append("--log-level=debug")
             self.cmd.append("--syslog=true")
 
--- a/test/python/docker/compat/test_containers.py
+++ b/test/python/docker/compat/test_containers.py
@@ -140,7 +140,7 @@ class TestContainers(unittest.TestCase):
 
     def test_remove_container_without_force(self):
         # Validate current container count
-        self.assertTrue(len(self.client.containers.list()), 1)
+        self.assertEqual(len(self.client.containers.list()), 1)
 
         # Remove running container should throw error
         top = self.client.containers.get(TestContainers.topContainerId)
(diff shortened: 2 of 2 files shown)


edsantiago:
/hold

Looks like my fix worked. I need to resubmit this with the `.cirrus.yml` changes removed (those were to avoid running lots of jobs). I will wait until #10947 merges, rebase, and remove the `Skip()`.

matejvasek:
@edsantiago disabling the debug seems to be good solution, but maybe better would be to demux stdout/stder:
```diff
diff --git a/test/python/docker/compat/test_containers.py b/test/python/docker/compat/test_containers.py
index 38ac5b59f..7c5a52454 100644
--- a/test/python/docker/compat/test_containers.py
+++ b/test/python/docker/compat/test_containers.py
@@ -226,10 +226,12 @@ class TestContainers(unittest.TestCase):
 
             buff.seek(0)
             ctr.put_archive("/tmp/", buff)
-            ret, out = ctr.exec_run(["stat", "-c", "%u:%g", "/tmp/a.txt"])
+            out: bytes
+            err: bytes
+            ret, (out, err) = ctr.exec_run(["stat", "-c", "%u:%g", "/tmp/a.txt"], demux=True)
 
             self.assertEqual(ret, 0)
-            self.assertTrue(out.startswith(b'1042:1043'), "assert correct uid/gid")
+            self.assertEqual(out, b"1042:1043\n", "assert copied file has correct uid/gid")
 
             ret, out = ctr.exec_run(["cat", "/tmp/a.txt"])
             self.assertEqual(ret, 0)
```

edsantiago:
@matejvasek that's much more pythony than I could've ever done; I will include it in my next push. Thank you!

matejvasek:
I included it in https://github.com/containers/podman/pull/10949 but maybe I'll close the PR.

edsantiago:
And, on second thought, I'm not going to do that. I'm really neurotic when I test: I want to do exact matches, not "close enough". There is no circumstance in which we should accept anything other than `1042:1043`. There is no circumstance in which we should accept anything on stderr. Anything on stderr should trigger a test failure.

cdoern:
LGTM

PR #10951, opened 2021-07-15
Randomize the auto-update of podman containers

rugk:
This makes sure, that the podman auto-update is not executed exactly at midnight for the same time always.
If many things do the same and many services use this keyword and react at midnight, this can cause a lot of stress to a server.

Thus, this adds a 900s/15min delay.

As [the arch wiki says](https://wiki.archlinux.org/title/Systemd/Timers#Realtime_timer):
> Special event expressions like daily and weekly refer to specific start times and thus any timers sharing such calendar events will start simultaneously. Timers sharing start events can cause poor system performance if the timers' services compete for system resources. The RandomizedDelaySec option in the [Timer] section avoids this problem by randomly staggering the start time of each timer. See systemd.timer(5).

### DIFF
--- a/contrib/systemd/auto-update/podman-auto-update.timer
+++ b/contrib/systemd/auto-update/podman-auto-update.timer
@@ -3,6 +3,7 @@ Description=Podman auto-update timer
 
 [Timer]
 OnCalendar=daily
+RandomizedDelaySec=900
 Persistent=true
 
 [Install]


rhatdan:
LGTM
@vrothberg PTAL
/approve

vrothberg (commented):
LGTM, thanks for the nice explanation!

/lgtm

PR #10952, opened 2021-07-15
Fix example; the --name option does not accept a parameter

jberkus:
Signed-off-by: Josh Berkus <josh@agliodbs.com>


mheon:
This doesn't seem right - `podman create --name` definitely takes an argument?

jberkus:
@mheon podman-generate-systemd --name does not. You can read the rest of the docs, or you can test it yourself.

Personally, I would like it to take a parameter, but it currently doesn't.

docs/source/markdown/podman-generate-systemd.1.md
```diff
@@ -64,7 +64,7 @@ Set the systemd unit name separator between the name/id of a container/pod and t
 Generate a systemd unit file for a container running nginx with an *always* restart policy and 1-second timeout to stdout. Note that the **RequiresMountsFor** option in the **Unit** section ensures that the container storage for both the GraphRoot and the RunRoot are mounted prior to starting the service. For systems with container storage on disks like iSCSI or other remote block protocols, this ensures that Podman is not executed prior to any necessary storage operations coming online.
 
 ```
-$ podman create --name nginx nginx:latest
+$ podman create --name nginx:latest
```
rhatdan, line 67: --name does take a value.

This is creating a container named nginx off of the nginx:latest image.

rhatdan:
@vrothberg PTAL, it would be nice if --name and --name=value worked, but this does not work in cobra.

Luap99:
> @vrothberg PTAL, it would be nice if --name and --name=value worked, but this does not work in cobra.

You can do this with `flags.Lookup("name").NoOptDefVal = "SOMEVALUE"`. If you set this you can do both `--name` and `--name=value`. The only problem is that you cannot set `NoOptDefVal` to an empty string.
https://play.golang.org/p/aXEkhTA_6c1

vrothberg:
Let's first agree whether there's a problem and which it is.

By default, `podman generate systemd` will generate units (and file names) based on the IDs of containers and pods. Using `--name` uses the names instead of the IDs.  Note that this is purely a boolean switch.

The change here is wrong as `podman create --name=xxx` works and the parameter is mandatory.

@jberkus, I hope this explanation makes sense.

> @vrothberg PTAL, it would be nice if --name and --name=value worked, but this does not work in cobra.

I _really_ do not like it due to the ambiguity of  `--name=foobar` on a POD; we cannot control the names of the containers.  At the moment, it only changes the mode between using either IDs or names and I think we should keep it that way.

rhatdan:
I guess the question is what is @jberkus goal?

jberkus:
Oh!  I misread the example as being podman generate systemd.  I get it now, closing.

Issue #10655, opened 2021-06-11
generated systemd unit files need to pull in network-online.target


srd424:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Generated systemd unit files need to pull in `network-online.target`; at the moment they have `After=network-online.target`, but as per the [systemd documentation](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/#cutthecraphowdoimakesurethatmyservicestartsafterthenetworkisreallyonline), they should have `Wants=network-online.target` as well.

In my particular case, I've seen podman starting containers before the host has finished using DHCP to configure the main interface, which in turns means the host resolv.conf has not been correctly populated and is therefore not correctly configured in the containers either.

 **Steps to reproduce the issue:**

(my particular instance of this problem is a race condition so hard to create a simple / scriptable reproducer; hopefully examination of systemd docs and podman generated unit files make this "obvious to inspection.")

**Describe the results you received:**

Started containers cannot always access DNS.

**Describe the results you expected:**

Started containers should reliably be able to access the network.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**
```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.15.2
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```
**Output of `podman info --debug`:**
```
host:
  arch: amd64
  buildahVersion: 1.20.1
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 4
  distribution:
    distribution: ubuntu
    version: "20.04"
  eventLogger: journald
  hostname: asr-ocihost
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.4.0-74-generic
  linkmode: dynamic
  memFree: 2218684416
  memTotal: 3826343936
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.19.1.3-9b83-dirty
      commit: 33851ada2cc9bf3945915565bf3c2df97facb92c
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
[...]

PR #10962, opened 2021-07-16
systemd: require network*-online*.target

vrothberg:
Require the network to be online in all (generated) systemd units to
make sure that containers and Podman run only after the network has been
fully configured.

Fixes: #10655


Luap99 (commented):
LGTM

PR #10964, opened 2021-07-16
[v3.2] vendor containers/common@v0.38.16

vrothberg:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/buildah v1.21.3
-	github.com/containers/common v0.38.15
+	github.com/containers/common v0.38.16
 	github.com/containers/conmon v2.0.20+incompatible
 	github.com/containers/image/v5 v5.12.0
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
I don't know if this is a candidate for RHEL 8.4.0.2, Podman 3.2.3, but I'm leery about merging this one in, especially without man page changes, just before we release....

rhatdan:
Why would you want man page changes, I believe this was breaking a previous non libimage version of Podman.  Basically auto-update was attempting to pull images from localhost/image.

TomSweeneyRedHat:
Unless I'm completely misunderstanding the change, this changes the way --pull works dependent upon the image that's being requested.

http://docs.podman.io/en/latest/markdown/podman-build.1.html#pull

Doesn't mention anything about what happens when `localhost` is part of the value passed into the image.

vrothberg:
I hope to find time next week to revisit a bit all the multi-arch issues and write to down exactly what happens when.  That will be a good exercise to find some improvements but also to massage that into the man pages.

FWIW, nothing has changed in Podman.  `podman pull` has always enforced --pull=always when custom arch, os or variant were specified.  Buildah behaved slightly different prior to libimage.

PR #10970, opened 2021-07-17
Copy the content from the underlying image into the newly created volume

vikas-goel:
Fixes: #10262

Signed-off-by: Vikas Goel <vikas.goel@gmail.com>

### DIFF
--- a/libpod/runtime_ctr.go
+++ b/libpod/runtime_ctr.go
@@ -47,6 +47,32 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
 	return r.newContainer(ctx, rSpec, options...)
 }
 
+func (r *Runtime) PrepareVolumeOnCreateContainer(ctx context.Context, ctr *Container) error {
+	// Copy the content from the underlying image into the newly created
+	// volume if configured to do so.
+	if !r.config.Containers.PrepareVolumeOnCreate {
+		return nil
+	}
+
+	defer func() {
+		if err := ctr.cleanupStorage(); err != nil {
+			logrus.Errorf("error cleaning up container storage %s: %v", ctr.ID(), err)
+		}
+	}()
+
+	mountPoint, err := ctr.mountStorage()
+	if err == nil {
+		// Finish up mountStorage
+		ctr.state.Mounted = true
+		ctr.state.Mountpoint = mountPoint
+		if err = ctr.save(); err != nil {
+			logrus.Errorf("Error saving container %s state: %v", ctr.ID(), err)
+		}
+	}
+
+	return err
+}
+
 // RestoreContainer re-creates a container from an imported checkpoint
 func (r *Runtime) RestoreContainer(ctx context.Context, rSpec *spec.Spec, config *ContainerConfig) (*Container, error) {
 	r.lock.Lock()
--- a/pkg/specgen/generate/container_create.go
+++ b/pkg/specgen/generate/container_create.go
@@ -153,7 +153,15 @@ func MakeContainer(ctx context.Context, rt *libpod.Runtime, s *specgen.SpecGener
 	if err != nil {
 		return nil, err
 	}
-	return rt.NewContainer(ctx, runtimeSpec, options...)
+
+	ctr, err := rt.NewContainer(ctx, runtimeSpec, options...)
+	if err != nil {
+		return ctr, err
+	}
+
+	// Copy the content from the underlying image into the newly created
+	// volume if configured to do so.
+	return ctr, rt.PrepareVolumeOnCreateContainer(ctx, ctr)
 }
 
 func extractCDIDevices(s *specgen.SpecGenerator) []libpod.CtrCreateOption {
--- a/test/e2e/container_create_volume_test.go
+++ b/test/e2e/container_create_volume_test.go
@@ -1,0 +1,127 @@
+package integration
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	. "github.com/containers/podman/v3/test/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
+)
+
+func buildDataVolumeImage(pTest *PodmanTestIntegration, image, data, dest string) {
+	// Create a dummy file for data volume
+	dummyFile := filepath.Join(pTest.TempDir, data)
+	err := ioutil.WriteFile(dummyFile, []byte(data), 0644)
+	Expect(err).To(BeNil())
+
+	// Create a data volume container image but no CMD binary in it
+	containerFile := fmt.Sprintf(`FROM scratch
+CMD doesnotexist.sh
+ADD %s %s/
+VOLUME %s/`, data, dest, dest)
+	pTest.BuildImage(containerFile, image, "false")
+}
+
+func createContainersConfFile(pTest *PodmanTestIntegration) {
+	configPath := filepath.Join(pTest.TempDir, "containers.conf")
+	containersConf := []byte(fmt.Sprintf("[containers]\nprepare_volume_on_create = true\n"))
+	err := ioutil.WriteFile(configPath, containersConf, os.ModePerm)
+	Expect(err).To(BeNil())
+
+	// Set custom containers.conf file
+	os.Setenv("CONTAINERS_CONF", configPath)
+	if IsRemote() {
+		pTest.RestartRemoteService()
+	}
+}
+
+func checkDataVolumeContainer(pTest *PodmanTestIntegration, image, cont, dest, data string) {
+	create := pTest.Podman([]string{"create", "--name", cont, image})
+	create.WaitWithDefaultTimeout()
+	Expect(create).Should(Exit(0))
+
+	inspect := pTest.InspectContainer(cont)
+	Expect(len(inspect)).To(Equal(1))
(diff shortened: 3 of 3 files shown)


vikas-goel:
@mheon , @rhatdan , @Luap99 , @TomSweeneyRedHat ,
I am facing two issues, one related to test case and another about the code change. I am posting them separately. I will appreciate any guidance.

vikas-goel:
I am using `config/containers-volume.conf` in test case but the execution failing to find the file. I have tried creating the file runtime in tmp directory instead of static file, that doesn't help either.

```
[+0935s] Running: /var/tmp/go/src/github.com/containers/podman/bin/podman --storage-opt vfs.imagestore=/tmp/podman/imagecachedir --root /tmp/podman_test791424325/crio --runroot /tmp/podman_test791424325/crio-run --runtime crun --conmon /usr/bin/conmon --cni-config-dir /etc/cni/net.d --cgroup-manager cgroupfs --tmpdir /tmp/podman_test791424325 --events-backend file --storage-driver vfs create --name ctr-copy dataimg
[+0935s] time="2021-07-19T15:47:35Z" level=error msg="Error streaming contents of container c15a6d0d258da0180568a3a4a502dc19e9c9a8984b48bedeb52d292e1cfa3608 directory for volume copy-up: error in copier subprocess: time=\"2021-07-19T15:47:35Z\" level=error msg=\"finding config on system: CONTAINERS_CONF file: stat config/containers-volume.conf: no such file or directory\"\n"
[+0935s] Error: error copying up to volume d2e89f0d1df08c3999078ff13be71a999bf6d507468b348956ebe0ce360e9f71: error in copier subprocess: time="2021-07-19T15:47:35Z" level=error msg="finding config on system: CONTAINERS_CONF file: stat config/containers-volume.conf: no such file or directory"
```

libpod/runtime_ctr.go
```diff
@@ -47,6 +47,32 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
...
 }
 
+func (r *Runtime) PrepareVolumeOnCreateContainer(ctx context.Context, ctr *Container) (retErr error) {
+	// Copy the content from the underlying image into the newly created
+	// volume if configured to do so.
+	if !r.config.Containers.PrepareVolumeOnCreate {
+		return nil
+	}
+
+	defer func() {
```
rhatdan, line 57: Since it's only the mountstorage that can fail, I don't think you need to cleanup storage at all.

libpod/runtime_ctr.go
```diff
@@ -47,6 +47,32 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
...
+			}
+		}
+	}()
+
+	mountPoint, err := ctr.mountStorage()
+	if err == nil {
+		// Finish up mountStorage
+		ctr.state.Mounted = true
+		ctr.state.Mountpoint = mountPoint
+		logrus.Infof("Container %s storage %s mounted successfully", ctr.Name(), mountPoint)
```
rhatdan, line 70: This should be Debug, not Info.

vikas-goel:
I am facing an issue where removing the container and volume created using `prepare_volume_on_create = true` show the following error message.
```
# ~/go/src/github.com/containers/podman/bin/podman rm -afv
Error: error removing container 58865407f8f7c1858704375a173a1cc93da4280cdb079aad44be536069540eff root filesystem: 2 errors occurred:
        * unlinkat /var/lib/containers/storage/overlay-containers/58865407f8f7c1858704375a173a1cc93da4280cdb079aad44be536069540eff/userdata/shm: device or resource busy
        * unlinkat /var/lib/containers/storage/overlay-containers/58865407f8f7c1858704375a173a1cc93da4280cdb079aad44be536069540eff/userdata/shm: device or resource busy
```

If I run `container init` and `container cleanup` after creating the container with `prepare_volume_on_create = true` option, then the error does not happen.

pkg/specgen/generate/container_create.go
```diff
@@ -153,7 +153,17 @@ func MakeContainer(ctx context.Context, rt *libpod.Runtime, s *specgen.SpecGener
...
-	return rt.NewContainer(ctx, runtimeSpec, options...)
+
+	ctr, err := rt.NewContainer(ctx, runtimeSpec, options...)
+	if err != nil {
+		return ctr, err
+	}
+
+	// Copy the content from the underlying image into the newly created
+	// volume if configured to do so.
+	verr := rt.PrepareVolumeOnCreateContainer(ctx, ctr)
```
rhatdan, line 164: Jusy use err.
I don't see the data being copied up in PrepareVolumeOnCreateContainer?

rhatdan (approved):

pkg/specgen/generate/container_create.go
vikas-goel: Using `err` was compilation error. I had to change that.
`PrepareVolumeOnCreateContainer()` calls `mountStorage()` that internally calls `mountNamedVolume()` which does the copy.

libpod/runtime_ctr.go
vikas-goel: Sure, next commit will take care of that.

libpod/runtime_ctr.go
vikas-goel: The `prepare()` function that calls `mountStorage()` does call `cleanupStorage()` afterwards. I am wondering whether not calling `cleanupStorage()` will result in any leak for the containers that are just created but never started/stopped.

libpod/runtime_ctr.go
```diff
@@ -47,6 +47,32 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
 	return r.newContainer(ctx, rSpec, options...)
 }
 
+func (r *Runtime) PrepareVolumeOnCreateContainer(ctx context.Context, ctr *Container) (retErr error) {
```
mheon, line 50: Why did you move to a separate function for this, as opposed to keeping it part of container creation?

vikas-goel: When I had this code originally in `setupContainer()`, then `podman create` would hang indefinitely. Only a process kill would make the operation abort. I guess some of the deferred calls in the `setupContainer()` need to be executed before calling the volume copy.

So, I made the change to look it somewhat like `podman create` followed by `podman init` so as to follow the same business logic (code path).

mheon:
@vikas-goel I've identified the problem. The issue is that `mountStorage()` is not setting `c.state.Mounted` which `unmountStorage()` checks before doing anything. After mounting storage you need to do a `c.state.Mounted = true` and then a `c.save()`.

vikas-goel:
> @vikas-goel I've identified the problem. The issue is that `mountStorage()` is not setting `c.state.Mounted` which `unmountStorage()` checks before doing anything. After mounting storage you need to do a `c.state.Mounted = true` and then a `c.save()`.

That works. The missing save() was the reason. Thank you @mheon .

TomSweeneyRedHat:
LGTM
assuming happy tests

vikas-goel:
Assigned to @TomSweeneyRedHat for merge

libpod/runtime_ctr.go
```diff
@@ -47,6 +47,33 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
...
+	// Copy the content from the underlying image into the newly created
+	// volume if configured to do so.
+	if !r.config.Containers.PrepareVolumeOnCreate {
+		return nil
+	}
+
+	defer func() {
+		if retErr != nil {
```
Luap99, line 58: There is no need to check for retErr, it is never used in this function.

libpod/runtime_ctr.go
```diff
@@ -47,6 +47,33 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
 	return r.newContainer(ctx, rSpec, options...)
 }
 
+func (r *Runtime) PrepareVolumeOnCreateContainer(ctx context.Context, ctr *Container) (retErr error) {
```
Luap99, line 50:
```suggestion
func (r *Runtime) PrepareVolumeOnCreateContainer(ctx context.Context, ctr *Container) error {
```

libpod/runtime_ctr.go
```diff
@@ -47,6 +47,33 @@ func (r *Runtime) NewContainer(ctx context.Context, rSpec *spec.Spec, options ..
...
+			}
+		}
+	}()
+
+	mountPoint, err := ctr.mountStorage()
+	if err == nil {
+		// Finish up mountStorage
+		ctr.state.Mounted = true
+		ctr.state.Mountpoint = mountPoint
+		ctr.save()
```
Luap99, line 70: You have to check the error from save().

vikas-goel:
Not sure why some of the checks are cancelled and the status say "Some checks were not successful".

vikas-goel:
Hi @TomSweeneyRedHat , could you help merge the PR is no other feedback?

pkg/specgen/generate/container_create.go
rhatdan: This should work.
```
err = rt.PrepareVolumeOnCreateContainer(ctx, ctr)
```

vikas-goel: I changed the code to not worry about storing in variable. Please take another look.

rhatdan:
Thanks @vikas-goel

vikas-goel:
Thank you @rhatdan , @mheon , @TomSweeneyRedHat , @Luap99

PR #10973, opened 2021-07-17
Support size options on builtin volumes

rhatdan:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1982164

### DIFF
--- a/libpod/runtime_volume_linux.go
+++ b/libpod/runtime_volume_linux.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/podman/v3/libpod/define"
 	"github.com/containers/podman/v3/libpod/events"
 	volplugin "github.com/containers/podman/v3/libpod/plugin"
+	"github.com/containers/storage/drivers/quota"
 	"github.com/containers/storage/pkg/stringid"
 	pluginapi "github.com/docker/go-plugins-helpers/volume"
 	"github.com/pkg/errors"
@@ -68,7 +69,7 @@ func (r *Runtime) newVolume(ctx context.Context, options ...VolumeCreateOption) 
 		// Validate options
 		for key := range volume.config.Options {
 			switch key {
-			case "device", "o", "type", "UID", "GID":
+			case "device", "o", "type", "UID", "GID", "SIZE", "INODES":
 				// Do nothing, valid keys
 			default:
 				return nil, errors.Wrapf(define.ErrInvalidArg, "invalid mount option %s for driver 'local'", key)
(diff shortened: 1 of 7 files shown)


rhatdan:
@mheon PTAL

mheon:
Nice and simple, LGTM

libpod/volume.go
```diff
@@ -49,6 +49,8 @@ type VolumeConfig struct {
 	UID int `json:"uid"`
 	// GID the volume will be created as.
 	GID int `json:"gid"`
+	// Size maximum of the volume.
+	Size int64 `json:"gid"`
```
Luap99, line 53: Should we store this as uint64 given that `SetQuota()` accepts uint64?
Also the json field name is wrong `s/gid/size/`.

libpod/runtime_volume_linux.go
```diff
@@ -106,6 +107,26 @@ func (r *Runtime) newVolume(ctx context.Context, options ...VolumeCreateOption)
...
+		}
+		quota := quota.Quota{}
+		if volume.config.Size > 0 || volume.config.Inodes > 0 {
+			if !projectQuotaSupported {
+				return nil, errors.New("Volume options size and inodes not supported. Filesystem does not support Project Quota")
+			}
+			quota.Size = volume.config.Size
+			quota.Inodes = volume.config.Inodes
+		}
+		if projectQuotaSupported {
```
mtrmac, line 124:
```suggestion
```
Just like this? We don’t need to set quota (and potentially conflict with XFS project IDs).

mtrmac: don’t need to set quota … if there isn’t a quota request.

mtrmac: … actually, all of this, including the quota support check, can be wrapped inside the `volume.config.{Size,Inodes} > 0` condition.

/me steps away to regain focus.

rhatdan:
@giuseppe @vrothberg @TomSweeneyRedHat @saschagrunert @baude @jwhonce PTAL

giuseppe (approved):
LGTM

saschagrunert (approved):
LGTM

TomSweeneyRedHat:
@mheon @rhatdan, unless I missed it, it doesn't look like this was cherry-picked into the v3.3 branch.  Can we make that so, so we get this bug fixed in RHEL 8.5?

mheon:
We're feature-frozen at this point, and this counts as a feature. Will that cause us problems?

TomSweeneyRedHat:
@mheon it has an associated BZ, so I think it's fair game.

Issue #10977, opened 2021-07-18
docker compatibility --platform doesn't work


fugkco:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Seems using the `--platform` tag via docker doesn't pull the expected image.

**Steps to reproduce the issue:**

1. Use podman to pull an arm64 image, and notice that it works fine:
   ```
   $ podman pull --platform linux/arm64 docker.io/node:12-buster-slim
   ...
   $ podman inspect docker.io/library/node:12-buster-slim --format '{{.Architecture}}'
   arm64
   $ docker inspect docker.io/library/node:12-buster-slim --format '{{.Architecture}}'
   arm64
   $ podman rmi node:12-buster-slim
   ```

2. Compare with to pulling with docker, which pulls the amd64:
   ```
   $ docker pull --platform linux/arm64 docker.io/node:12-buster-slim
   ...
   $ podman inspect docker.io/library/node:12-buster-slim --format '{{.Architecture}}'
   amd64
   $ docker inspect docker.io/library/node:12-buster-slim --format '{{.Architecture}}'
   amd64
   ```

**Describe the results you received:**
Pulled image is `amd64`

**Describe the results you expected:**
Expect docker pull to pull the platform requested, in above case `arm64`.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
$ podman version
Version:      3.2.2
API Version:  3.2.2
Go Version:   go1.14.7
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64

```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.21.0
  cgroupControllers: []
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: 'conmon: /usr/libexec/podman/conmon'
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: pop
    version: "21.04"
  eventLogger: journald
  hostname: pop-os
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.11.0-7620-generic
[...]

PR #10979, opened 2021-07-19
compat: image create: handle platform correctly

vrothberg:
Handle the platform parameter correctly.  The parameter was only parsed
in presence of credentials and the code was a bit complex.  Also add a
regression test.

Fixes: #10977

### DIFF
--- a/pkg/api/handlers/compat/images.go
+++ b/pkg/api/handlers/compat/images.go
@@ -266,41 +266,26 @@ func CreateImageFromImage(w http.ResponseWriter, r *http.Request) {
 	}
 	defer auth.RemoveAuthfile(authfile)
 
-	platformSpecs := strings.Split(query.Platform, "/") // split query into its parts
-
-	addOS := true // default assume true due to structure of if/else below
-	addArch := false
-	addVariant := false
-
-	if len(platformSpecs) > 1 { // if we have two arguments then we have os and arch
-		addArch = true
-		if len(platformSpecs) > 2 { // if we have 3 arguments then we have os arch and variant
-			addVariant = true
-		}
-	} else if len(platformSpecs) == 0 {
-		addOS = false
-	}
-
 	pullOptions := &libimage.PullOptions{}
 	pullOptions.AuthFilePath = authfile
 	if authConf != nil {
 		pullOptions.Username = authConf.Username
 		pullOptions.Password = authConf.Password
 		pullOptions.IdentityToken = authConf.IdentityToken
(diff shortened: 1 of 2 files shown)


test/apiv2/10-images.at
```diff
@@ -45,6 +45,10 @@ t POST "images/create?fromImage=alpine" 200 .error~null .status~".*Download comp
 
 t POST "images/create?fromImage=alpine&tag=latest" 200
 
+# 10977 - handle platform parameter correctly
+t POST "images/create?fromImage=alpine&platform=linux/arm64" 200
+is "$($PODMAN_BIN inspect --format '{{.Architecture}}' alpine)" "arm64" "platform parameter works"
```
vrothberg, line 50: Note: `$PODMAN_BIN` was necessary to actually get the output of the command.

rhatdan:
LGTM

pkg/api/handlers/compat/images.go
```diff
@@ -266,41 +266,26 @@ func CreateImageFromImage(w http.ResponseWriter, r *http.Request) {
...
-				pullOptions.Variant = platformSpecs[2]
-			}
-		}
 	}
 	pullOptions.Writer = os.Stderr // allows for debugging on the server
 
-	progress := make(chan types.ProgressProperties)
+	// Handle the platform.
+	platformSpecs := strings.Split(query.Platform, "/")
+	if len(platformSpecs) > 1 {
```
cdoern, line 280: I think you can pass an OS (technically) without a platform even though it doesn't make much sense that is why I checked explicitly for length =1, 2 or 3

test/apiv2/10-images.at
```diff
@@ -45,6 +45,11 @@ t POST "images/create?fromImage=alpine" 200 .error~null .status~".*Download comp
 
 t POST "images/create?fromImage=alpine&tag=latest" 200
 
+# 10977 - handle platform parameter correctly
+t POST "images/create?fromImage=alpine&platform=linux/arm64" 200
+t GET  "images/alpine/json" 200
+is "$(echo $output | jq .Architecture)" "\"arm64\"" "platform parameter works"
```
Luap99, line 51: Can you please use the following like the other tests:
```suggestion
t GET  "images/alpine/json" 200 \
  .Architecture=arm64
```
This looks much better in the test log.

pkg/api/handlers/compat/images.go
vrothberg: Good catch, I'll fix that.

Issue #10052, opened 2021-04-16
compose flake: rootlesskit port forwarder not setup correctly


Luap99:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The rootless compose test flaking regularly in CI. The problem is that the port forwarding with rootlesskit is not working. Rootlesskit does not seem to bind the port on the host.
A example test failure with somw debug information can be seen here: https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5302424773591040/html/compose-podman-fedora-34beta-rootless-host.log.html
Relevant log lines:
```
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport: time=\"2021-04-16T04:34:23-05:00\" level=info msg=\"starting parent driver\"\n"
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport: time=\"2021-04-16T04:34:23-05:00\" level=info msg=\"opaque=map[builtin.readypipepath:/run/user/30212/libpod/tmp/rootlessport761613945/.bp-ready.pipe builtin.socketpath:/run/user/30212/libpod/tmp/rootlessport761613945/.bp.sock]\"\n"
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport: time=\"2021-04-16T04:34:23-05:00\" level=info msg=\"starting child driver in child netns (\\\"/proc/self/exe\\\" [containers-rootlessport-child])\"\n"
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport: time=\"2021-04-16T04:34:23-05:00\" level=info msg=\"waiting for initComplete\"\n"
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport: time=\"2021-04-16T04:34:23-05:00\" level=info msg=\"initComplete is closed; parent and child established the communication channel\"\n"
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport: time=\"2021-04-16T04:34:23-05:00\" level=info msg=\"exposing ports [{5000 5000 tcp }]\"\n"
time="2021-04-16T04:34:23-05:00" level=debug msg="rootlessport is ready"
```
A successful run looks like this:
```
time="2021-04-16T13:33:14+02:00" level=debug msg="rootlessport: time=\"2021-04-16T13:33:14+02:00\" level=info msg=\"starting parent driver\"\n"
[...]

PR #10982, opened 2021-07-19
Flake Fix: Wait before connecting to container port

cevich:
It was observed during periodic testing, this test can fail due to the
container process being not fully running and listening on the expected
port:

```
[+1069s] not ok 220 podman networking: port with --userns=keep-id
[+1069s] # (in test file test/system/500-networking.bats, line 144)
[+1069s] #   `echo "$teststring" | nc 127.0.0.1 $myport' failed
[+1069s] # # /var/tmp/go/src/github.com/containers/podman/bin/podman rm
--all --force
[+1069s] # # /var/tmp/go/src/github.com/containers/podman/bin/podman ps
--all --external --format {{.ID}} {{.Names}}
[+1069s] # # /var/tmp/go/src/github.com/containers/podman/bin/podman
images --all --format {{.Repository}}:{{.Tag}} {{.ID}}
[+1069s] # quay.io/libpod/testimage:20210610 9f9ec7f2fdef
[+1069s] # # /var/tmp/go/src/github.com/containers/podman/bin/podman run
-d --userns=keep-id -p 127.0.0.1:54322:54322
quay.io/libpod/testimage:20210610 nc -l -n -v -p 54322
[+1069s] #
252c562c9a3c96892d867d1d72fb52b2efdfe62855ebedbccd2d281c472c2988
[+1069s] # Ncat: No route to host.
```

Fix this by using a new `wait_for_port()` function (thanks @edsantiago)
before attempting to communicate with the service.

### DIFF
--- a/test/system/500-networking.bats
+++ b/test/system/500-networking.bats
@@ -139,6 +139,8 @@ load helpers
                    $IMAGE nc -l -n -v -p $myport
         cid="$output"
 
+        wait_for_port 127.0.0.1 $myport
+
         # emit random string, and check it
         teststring=$(random_string 30)
         echo "$teststring" | nc 127.0.0.1 $myport
--- a/test/system/helpers.bash
+++ b/test/system/helpers.bash
@@ -278,6 +278,24 @@ function wait_for_ready {
     wait_for_output 'READY' "$@"
 }
 
+###################
+#  wait_for_port  #  Returns once port is available on host
+###################
+function wait_for_port() {
+    local host=$1                      # Probably "localhost"
+    local port=$2                      # Numeric port
+    local _timeout=${3:-5}              # Optional; default to 5 seconds
+
+    # Wait
+    while [ $_timeout -gt 0 ]; do
+        { exec 3<> /dev/tcp/$host/$port; } &>/dev/null && return
+        sleep 1
+        _timeout=$(( $_timeout - 1 ))
+    done
+
+    die "Timed out waiting for $host:$port"
+}
+
 # END   podman helpers
 ###############################################################################
 # BEGIN miscellaneous tools


test/system/500-networking.bats
```diff
@@ -139,6 +139,9 @@ load helpers
                    $IMAGE nc -l -n -v -p $myport
         cid="$output"
 
+        # Workaround: Process execution, Socket & port binding isn't instant.
+        sleep 1s
```
edsantiago, line 143: Suggestion: copy and use https://github.com/containers/podman/blob/0a0ade3cc00c3779bbf68ddd103d3efd10b5c25b/test/apiv2/test-apiv2#L439-L454

cevich: Oh! Okay, that's better, thanks!

cevich: I tweaked the `die` message slightly, before adding it into `helpers.bash`.

test/system/helpers.bash
```diff
@@ -278,6 +278,24 @@ function wait_for_ready {
...
+function wait_for_port() {
+    local host=$1                      # Probably "localhost"
+    local port=$2                      # Numeric port
+    local timeout=${3:-5}              # Optional; default to 5 seconds
+
+    # Wait
+    while [ $timeout -gt 0 ]; do
+        { exec 3<> /dev/tcp/$host/$port; } &>/dev/null && return
+        sleep 1
+        _timeout=$(( $_timeout - 1 ))
```
Luap99, line 293: The variable name is misspelled.
```suggestion
        timeout=$(( $timeout - 1 ))
```

Luap99 (commented):
I do not think no route to host is a timing issue but it is worth a try.

test/system/helpers.bash
cevich: @edsantiago I copy-pasted that part, was there a source for this code that also needs a typo-fixed?

cevich:
> I do not think no route to host is a timing issue but it is worth a try.

Hmm true...maybe CNI processing/setup delay?  In any case, if it ain't there in 5-seconds, something "else" is broken which we also want to know about.  Waiting for the port to open is certainly improved over my original `sleep 1s` fix :grin:

test/system/helpers.bash
edsantiago: https://github.com/containers/podman/pull/10984

cevich: Oh goodie...want me to fix this one to use the same `_timeout` style?

cevich: nevermind, an unrelated test failed anyway.  I fixed this to match the `_timeout` style.

edsantiago:
Yeah... this is unlikely to fix the `No route to host` problem, but it's still a good thing to do.

FWIW this and a bunch of other rootless networking flakes are being ~~tracked~~ ignored in #10052.

Issue #10832, opened 2021-07-01
Intermediate layers for tagged images are incorrectly pruned


Enchufa2:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

`podman-image-prune` incorrectly prunes intermediate layers for tagged images.

**Steps to reproduce the issue:**

Consider the following Dockerfile:

```
FROM scratch
ENV test1=test1
ENV test2=test2
```

then,

```bash
$ podman build . -t test
STEP 1: FROM scratch
STEP 2: ENV test1=test1
--> f106d67818d
STEP 3: ENV test2=test2
STEP 4: COMMIT test
--> 3775b6bda38
Successfully tagged localhost/test:latest
3775b6bda38dc3cf77b00e540706c367bc7d6fff606bbc7eab75842fcb2f41a8
$ podman image prune 
WARNING! This will remove all dangling images.
Are you sure you want to continue? [y/N] y
f106d67818dd8ee583d3fb3052a87705ac8c953c9c9e3ce453b51a9da57e1271
```

**Describe the results you received:**

As you can see, intermediate layer `f106d67818d` was removed.

**Describe the results you expected:**

It shouldn't be removed. FWIW, docker doesn't remove it.

**Output of `podman version`:**

```
Version:      3.2.1
API Version:  3.2.1
Go Version:   go1.16.3
Built:        Mon Jun 14 21:12:29 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.21.0
  cgroupControllers: []
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: ****
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.12.11-300.fc34.x86_64
  linkmode: dynamic
  memFree: 6931976192
  memTotal: 16467206144
  ociRuntime:
    name: crun
[...]

PR #10983, opened 2021-07-19
refine dangling checks

vrothberg:
By proxy by vendoring containers/common. Previously, a "dangling" image
was an untagged image; just a described in the Docker docs. The
definition of dangling has now been refined to an untagged image without
children to be compatible with Docker.

Further update a redundant image-prune test.

Fixes: #10998
Fixes: #10832


vrothberg:

go.mod
```diff
@@ -69,3 +69,5 @@ require (
 	k8s.io/api v0.21.3
 	k8s.io/apimachinery v0.21.3
 )
+
+replace github.com/containers/common => github.com/vrothberg/common v0.0.3-0.20210720091711-39e7214a081e
```
rhatdan, line 73: Looks like this is still a WIP?

pkg/domain/infra/abi/images_list.go
```diff
@@ -30,12 +30,16 @@ func (ir *ImageEngine) List(ctx context.Context, opts entities.ImageListOptions)
 		for j, d := range img.Digests() {
 			digests[j] = string(d)
 		}
+		isDangling, err := img.IsDangling(ctx)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error checking if image %q is dangling", img.ID())
```
rhatdan, line 35: NIT: I don't think we should start errors with "error", Podman prepends Error: on all Errors, I don't want to end up with 
Error: error checking if image ...

go.mod
vrothberg: Thank you for catching that!  I missed to remove that line.

pkg/domain/infra/abi/images_list.go
vrothberg: I'll tackle that in a separate PR.  There more such cases in the code that I can tackle at once.

rhatdan:
LGTM

vrothberg:
@containers/podman-maintainers @Luap99 PTAL

TomSweeneyRedHat:
Does `podman images --filter dangling=true` work appropriately with these changes?

Also the man page for `podman rmi` defines an untagged image as a dangling one, does that man page need changing?

vrothberg:
> Does `podman images --filter dangling=true` work appropriately with these changes?

Yes.

> Also the man page for `podman rmi` defines an untagged image as a dangling one, does that man page need changing?

Good catch, thanks!  Fixed :+1:

vrothberg:

TomSweeneyRedHat (commented):
LGTM
and happy green test buttons.

vrothberg:
merge me :)

saschagrunert (approved):
/lgtm

PR #10985, opened 2021-07-19
system test: auto-update: multiarch fixes, and cleanup

edsantiago:
auto-update test is failing on non-x86_64 arch:

   https://gitlab.com/cki-project/kernel-tests/-/issues/614#note_630010734

Reason: test was relying on alpine_nginx image; this is
a custom libpod image that only has a x86_64 version.

Solution: use busybox instead. Busybox was copied from
docker.io using skopeo copy --all, so it has all arches.

Also: allow tests to run as rootless, they were being
skipped for no clear reason.

Also: use --files and --name option in generate-systemd. Before,
output went to stdout, resulting in completely unreadable logs
on test failure.

### DIFF
--- a/test/system/255-auto-update.bats
+++ b/test/system/255-auto-update.bats
@@ -5,13 +5,20 @@
 
 load helpers
 
-UNIT_DIR="/usr/lib/systemd/system"
+DASHUSER=""
+UNIT_DIR="/run/systemd/system"
 SNAME_FILE=$BATS_TMPDIR/services
 
 function setup() {
     skip_if_remote "systemd tests are meaningless over remote"
-    skip_if_rootless
 
+    if is_rootless; then
+        test -n "${XDG_RUNTIME_DIR}" || skip "\$XDG_RUNTIME_DIR is unset"
+        UNIT_DIR="${XDG_RUNTIME_DIR}/systemd/user"
+        mkdir -p $UNIT_DIR
+        # Why isn't systemd smart enough to figure this out on its own?
+        DASHUSER="--user"
+    fi
     basic_setup
 }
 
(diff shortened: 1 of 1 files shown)


edsantiago:
Deliberately submitting as two separate commits

edsantiago:
Confirmation: auto-update tests [ran and passed](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5392754197921792/html/sys-podman-ubuntu-2010-rootless-host.log.html#t--00177) (even on Ubuntu, which was my main worry)

test/system/255-auto-update.bats
```diff
@@ -5,31 +5,37 @@
 
 load helpers
 
+DASHUSER=""
 UNIT_DIR="/usr/lib/systemd/system"
```
Luap99, line 9: Not related to your changes but can we please store the unit files in `/run/systemd/system/` for root and `$XDG_RUNTIME_DIR/systemd/user/` for rootless. Using a tmp dir should prevent the files from dangling for ever if the test got interrupted. `/usr/lib/systemd/system` is used by the package manager I do not want some random test to mess with files in there.

Luap99: see `systemd.unit(5)`

vrothberg (commented):
Other than @Luap99's comment, LGTM

test/system/255-auto-update.bats
edsantiago: @Luap99 **thank you**! That has long been a top concern of mine, but I hadn't found that particular section of the systemd man pages. Fixed, and I will follow up separately with a fix for the other tests that use `/lib/systemd`.

Issue #10969, opened 2021-07-16
Mount error when using ':z' on podman 3.2


pascapascapascapasca:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When I start up a pod with a container that has a shared mount (`:z`), the mount fails. I also see this error in the journal: `kernel: overlayfs: unrecognized mount option "volatile" or missing value`.

I only get this issue since I've updated to podman 3.2. Other mount types (`:Z` and `:ro`) seem to work fine.

I tried switching runtimes between `runc` and `crun` and both failed.

I'm on AlmaLinux 8.4.

**Steps to reproduce the issue:**

1. Start pod with `podman play kube` with this definition. I use this one as example, but, it happens to all my pods with `:z` mounts. The `config.yml` file exists on disk at the right location.

```
---
apiVersion: v1
kind: Pod
metadata:
  name: etcd-pod
  labels:
    app.kubernetes.io/name: etcd
    app.kubernetes.io/instance: <...>
    group: <...>
spec:
  hostNetwork: true
  restartPolicy: Never
  volumes:
    - name: etcd
      hostpath:
        path: /var/lib/etcd
  containers:
    - name: etcd
      image: gcr.io/etcd-development/etcd:v3.4.15
      volumeMounts:
        - name: etcd
          mountPath: /etcd:z
      env:
        - name: ECTDCTL_API
          value: 3
      command:
        - /usr/local/bin/etcd
        - --config-file
        - /etcd/config.yml
 ```

2. Start container with `podman start -a etcd-pod-etcd`

**Describe the results you received:**

The container does not mount the volume and the command fails.

```
E | etcdmain: error verifying flags, open /etcd/config.yml: no such file or directory. See 'etcd --help'.
```

I also see `kernel: overlayfs: unrecognized mount option "volatile" or missing value` in the journal.

**Describe the results you expected:**

Volume should be mounted with `:z`.

**Output of `podman version`:**

```
Version:      3.2.2
API Version:  3.2.2
Go Version:   go1.15.13
Built:        Fri Jul  9 20:40:27 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:                                                                                                                                                                                                       [45/5776]
  arch: amd64
  buildahVersion: 1.21.0
  cgroupControllers:
  - cpuset
  - cpu
  - cpuacct
  - blkio
  - memory
  - devices
  - freezer
  - net_cls
[...]

PR #10992, opened 2021-07-20
Fix handling of selinux labels in podman play kube

rhatdan:
Fixes: https://github.com/containers/podman/issues/10969

[NO TESTS NEEDED] We added tests for this, but they don't seem to be
running.  If I run the local system tests, they fail with the current
Podman and work with this version.

### DIFF
--- a/pkg/specgen/generate/kube/kube.go
+++ b/pkg/specgen/generate/kube/kube.go
@@ -276,10 +276,11 @@ func ToSpecGen(ctx context.Context, opts *CtrSpecGenOptions) (*specgen.SpecGener
 			return nil, err
 		}
 
+		volume.MountPath = dest
 		switch volumeSource.Type {
 		case KubeVolumeTypeBindMount:
 			mount := spec.Mount{
-				Destination: dest,
+				Destination: volume.MountPath,
 				Source:      volumeSource.Source,
 				Type:        "bind",
 				Options:     options,
@@ -287,7 +288,7 @@ func ToSpecGen(ctx context.Context, opts *CtrSpecGenOptions) (*specgen.SpecGener
 			s.Mounts = append(s.Mounts, mount)
 		case KubeVolumeTypeNamed:
 			namedVolume := specgen.NamedVolume{
-				Dest:    dest,
+				Dest:    volume.MountPath,
 				Name:    volumeSource.Source,
 				Options: options,
 			}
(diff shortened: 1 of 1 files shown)


umohnani8:
LGTM

edsantiago:
>We added tests for this, but they don't seem to be running.

Can you help me understand why this is happening? If you point me at a specific test, I can try to track down prior CI runs and see if they're skipped. If the tests are running and passing, that would be bad, no?

rhatdan:
700-play.bats

@test "podman play with stdin" {
    TESTDIR=$PODMAN_TMPDIR/testdir
    mkdir -p $TESTDIR
    echo "$testYaml" | sed "s|TESTDIR|${TESTDIR}|g" > $PODMAN_TMPDIR/test.yaml
    run_podman play kube - < $PODMAN_TMPDIR/test.yaml
    if [ -e /usr/sbin/selinuxenabled -a /usr/sbin/selinuxenabled ]; then
       run ls -Zd $TESTDIR
       is "$output" ${RELABEL} "selinux relabel should have happened"
    fi
    run_podman pod rm -f test_pod
}

I would guess these tests are running on disabled SELinux machines.  When I ran them locally they blew up.

edsantiago:
> When I ran them locally they blew up.

Do you still have the command you ran, and the output? I've never seen that test fail, and I often run full bats runs on my laptop (with SELinux enforcing, needless to say). I also confirmed that the test is not being skipped in CI. I think there might be something we're not actually testing.

rhatdan:
I ran them locally on F34 as root yesterday and it failed, with
podman-3.2.2-1.fc34.x86_64

edsantiago:
```console
# bats --filter stdin /usr/share/podman/test/system/700-play.bats
 ✓ podman play with stdin

1 test, 0 failures
# rpm -q podman
podman-3.2.2-1.fc34.x86_64
```
These tests pass all the time in gating. I run them manually, often, on my laptop and on 1mt VMs. If it failed in your setup, I need to know how you invoked it and what the error was. I have to assume it's a problem with your setup or your test invocation.

More importantly, I would like to write a test for this PR, because I don't like PRs going in without tests. I'm trying to figure out a reproducer from #10969 but am having trouble reducing it to something small.

acastong:
@rhatdan, is this a change we could consider porting over to 3.2.x?

mheon:
We're done with 3.2.x releases at this point. We are looking towards a 3.3.0 something in early August.

edsantiago:
We have a problem. We have no tests for this bug fix, no way to catch a regression. 

I've spent several hours this week trying to reproduce the failure described in #10969, and can't come up with a test case that fails under pre-10992 podman. That makes me think our 700-play.bats tests are incomplete; I just don't know how to fix them. And I'm giving up because this has become counterproductive for me. Am writing this note because some day this may come back to bite us.

PR #10995, opened 2021-07-20
system tests: cleaner, safer use of systemd

edsantiago:
First and foremost: use ephemeral (/run, $XDG) directories
for systemd unit files, so as not to vandalize a working system.

Second, refactor common systemd-related functionality into
a new helper file, loaded by the systemd-related tests.
Shared functionality includes:

  * setting $XDG_RUNTIME_DIR if unset and rootless
  * setting $UNIT_DIR for use by tests
  * new systemctl() and journalctl() functions, which
    include "--user" when rootless (why can't systemd
    figure this out on its own?)

### DIFF
--- a/test/system/250-systemd.bats
+++ b/test/system/250-systemd.bats
@@ -4,17 +4,10 @@
 #
 
 load helpers
+load helpers.systemd
 
 SERVICE_NAME="podman_test_$(random_string)"
 
-SYSTEMCTL="systemctl"
-UNIT_DIR="/usr/lib/systemd/system"
-if is_rootless; then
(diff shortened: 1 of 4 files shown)


vrothberg (commented):
LGTM
Love it! Thanks to you and @Luap99.

Luap99:
>       new systemctl() and journalctl() functions, which
>       include "--user" when rootless (why can't systemd
>       figure this out on its own?)

@edsantiago  You made me curious, so I did a little search: https://github.com/systemd/systemd/issues/18419

PR #10996, opened 2021-07-20
Implemented --until flag for Libpod's Container Logs

cdoern:
Compat containers/logs was missing actual usage of until query param.
This led me to implement the until param for libpod's container logs as well. Added e2e tests and documentation.

Until is a parameter which shows logs up until a certain timestamp. The default, similar to `--since`, is zero which means all logs. Support added for all log formats that libpod supports.

Signed-off-by: cdoern <cdoern@redhat.com>

### DIFF
--- a/docs/source/markdown/podman-logs.1.md
+++ b/docs/source/markdown/podman-logs.1.md
@@ -39,6 +39,14 @@ strings (e.g. 10m, 1h30m) computed relative to the client machine's time. Suppor
 time stamps include RFC3339Nano, RFC3339, 2006-01-02T15:04:05, 2006-01-02T15:04:05.999999999, 2006-01-02Z07:00,
 and 2006-01-02.
 
+#### **--until**=*TIMESTAMP*
+
+Show logs until TIMESTAMP. The --until option can be Unix timestamps, date formatted timestamps, or Go duration
+strings (e.g. 10m, 1h30m) computed relative to the client machine's time. Supported formats for date formatted
+time stamps include RFC3339Nano, RFC3339, 2006-01-02T15:04:05, 2006-01-02T15:04:05.999999999, 2006-01-02Z07:00,
+and 2006-01-02.
+
+
 #### **--tail**=*LINES*
 
 Output the specified number of LINES at the end of the logs.  LINES must be an integer.  Defaults to -1,
@@ -74,6 +82,17 @@ podman logs --tail 2 b3f2436bdb97
 # Server initialized
 ```
 
+To view all containers logs:
+```
+podman logs -t --since 0 myserver
+
+1:M 07 Aug 14:10:09.055 # Server can't set maximum open files to 10032 because of OS error: Operation not permitted.
+1:M 07 Aug 14:10:09.055 # Current maximum open files is 4096. maxclients has been reduced to 4064 to compensate for low ulimit. If you need higher maxclients increase 'ulimit -n'.
+1:M 07 Aug 14:10:09.056 * Running mode=standalone, port=6379.
+1:M 07 Aug 14:10:09.056 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
+1:M 07 Aug 14:10:09.056 # Server initialized
+```
+
 To view a containers logs since a certain time:
 ```
 podman logs -t --since 2017-08-07T10:10:09.055837383-04:00 myserver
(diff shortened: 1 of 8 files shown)


libpod/container_log_linux.go
```diff
@@ -152,17 +152,22 @@ func (c *Container) readFromJournal(ctx context.Context, options *logs.LogOption
 
 			if !afterTimeStamp {
 				entryTime := time.Unix(0, int64(entry.RealtimeTimestamp)*int64(time.Microsecond))
-				if entryTime.Before(options.Since) {
+				if entryTime.Before(options.Since) && options.Until.IsZero() {
 					continue
 				}
 				afterTimeStamp = true
+				beforeTimeStamp = true
```
vrothberg, line 159: Can you elaborate why after and before are set?

libpod/container_log_linux.go
```diff
@@ -152,17 +152,22 @@ func (c *Container) readFromJournal(ctx context.Context, options *logs.LogOption
...
 			}
-			if beforeTimeStamp {
+
+			if !beforeTimeStamp {
```
vrothberg, line 162: Don't we need to check "before" for each `entry` and exit once we passed it?

vrothberg (commented):
Nice work, @cdoern!

Do we need to change some swagger docs for the libpod endpoint?

Could you also extend the system tests in `test/system/035-logs.bats`?  Those are executed in the RHEL gating tests.

libpod/container_log_linux.go
cdoern: the `afterTimeStamp` path was already there so I tried to stick as closely to that for the `beforeTimeStamp` path as possible. Seems to me like they are used further down to indicate when to skip the log line via a `continue` and when to print it if the time is accurate.

libpod/container_log_linux.go
cdoern: My thinking here is that if we are before the time stamp we want to print the log line. From what I can tell, the current implementation of `since` causes you to skip anything that doesn't apply to you. I was trying to mimic that here and if we are before the `until` flag then that line does in fact apply to us.

libpod/container_log_linux.go
cdoern: isn't that the proper functionality? if you set since 3m, you are looking for all logs since 3 minutes ago. Until is the strict bound in this situation. The since integration tests still pass when until is equal to 0m. I could be really misunderstanding something here but from my point of view it seems to work?

cdoern: nevermind, I see what you mean now!!! Since until is always (in a valid case) going to be later than since, we want to only set both `aftetTimeStamp = true` and `beforeTimeStamp = true` when we have a valid until case.

vrothberg: Let's make this a PS in today's meeting and maybe play a bit with Docker.

To my current understanding, `since` sets a lower bound, `until` sets an upper one; no logs outside the range should be printed.

cdoern: sgtm @vrothberg I think my most recent push fixes all of the issues and I removed the need for the two booleans. Also added system tests.

cdoern:
@rhatdan @jwhonce PTAL

docs/source/markdown/podman-logs.1.md
```diff
@@ -39,6 +39,14 @@ strings (e.g. 10m, 1h30m) computed relative to the client machine's time. Suppor
 time stamps include RFC3339Nano, RFC3339, 2006-01-02T15:04:05, 2006-01-02T15:04:05.999999999, 2006-01-02Z07:00,
 and 2006-01-02.
 
+#### **--until**=*TIMESTAMP*
```
TomSweeneyRedHat, line 42: Maybe it's my warped mind, but when I saw this flag, I thought it would act like tail on the log for the specified time period, not show the logs from the "last" time period.  

Is there precedence for "until" vs "last" or "prior" or some such?

TomSweeneyRedHat:
The code/doc LGTM, I'm just not sold on the on the option name.

docs/source/markdown/podman-logs.1.md
rhatdan: Sadly this is what Docker calls i.
```
docker logs --help
...
      --until string   Show logs before a timestamp (e.g.
                       2013-01-02T13:23:37Z) or relative (e.g. 42m for 42
                       minutes)
```

PR #10997, opened 2021-07-20
e2e tests: prevent 'Expect(ExitCode())' pattern

edsantiago:
Followup to #10932: add a validation check to prevent introduction
of new 'Expect(foo.ExitCode()).To(...)' patterns. If such use is
absolutely necessary -- there is one such instance in the code
already -- require that the assertion include a description.

Also: clean up instances that were introduced since the merging
of #10932.

Also: fix one remaining instance in run_exit_test.go: it had
a FIXME comment mentioning a race condition, but unfortunately
there was no issue or bug ID, hence no way to know if the race
is fixed or not. We will assume it is.


rhatdan:
LGTM

PR #11006, opened 2021-07-21
[v3.2.3-rhel] Remove a test that is nonfunctional on RHEL8

mheon:
We don't support the journald events backend on 8, so remove a test that forces its use.

Partial fix for RHBZ1955166

### DIFF
--- a/test/system/090-events.bats
+++ b/test/system/090-events.bats
@@ -81,6 +81,7 @@ function _events_disjunctive_filters() {
 
 @test "events with disjunctive filters - journald" {
     skip_if_remote "remote does not support --events-backend"
+    skip_if_journald_unavailable "system does not support journald events"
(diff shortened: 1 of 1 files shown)


mheon:

test/system/090-events.bats
```diff
@@ -79,11 +79,6 @@ function _events_disjunctive_filters() {
     _events_disjunctive_filters --events-backend=file
 }
 
-@test "events with disjunctive filters - journald" {
-    skip_if_remote "remote does not support --events-backend"
```
edsantiago, line 83: Why not just `skip_if_journald_unavailable` ?

mheon: Because I didn't know that existed :smile: 

Concur that's better. I'll also make this change against main.

PR #11007, opened 2021-07-21
Ensure journald events tests only run where supported

mheon:
Forward-port #11006 to main branch, so this doesn't bite us in the future.

### DIFF
--- a/test/system/090-events.bats
+++ b/test/system/090-events.bats
@@ -81,6 +81,7 @@ function _events_disjunctive_filters() {
 
 @test "events with disjunctive filters - journald" {
     skip_if_remote "remote does not support --events-backend"
+    skip_if_journald_unavailable "system does not support journald events"
     _events_disjunctive_filters --events-backend=journald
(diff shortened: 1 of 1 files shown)


test/system/090-events.bats
```diff
@@ -81,6 +81,7 @@ function _events_disjunctive_filters() {
 
 @test "events with disjunctive filters - journald" {
     skip_if_remote "remote does not support --events-backend"
+    skip_if_journald_unavailable "system does not support journald events"
```
edsantiago, line 84: Absolutely not worth resubmitting for; but should you need to do this again in future, I'd suggest adding "rootless" to the skip message. Potential benefit is ultra-unlikely, though: some sort of system bug causing journald to be unavailable as root, _and_ someone looking at the test logs, _and_ that someone happening to notice the "skip" message for this test (skip messages aren't normally prominent), _and_ that someone noticing the word "rootless" in a rootful test.

PR #11009, opened 2021-07-21
Upgrade test: add new baseline

edsantiago:
Add a new reference point against which to test upgrades.

And, remove some FIXME comments, thanks to @cevich for
clarifying those questions (offline, some months ago).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -643,6 +643,8 @@ upgrade_test_task:
               PODMAN_UPGRADE_FROM: v2.0.6
         - env:
               PODMAN_UPGRADE_FROM: v2.1.1
+        - env:
+              PODMAN_UPGRADE_FROM: v3.1.2
     gce_instance: *standardvm
     env:
         TEST_FLAVOR: upgrade_test


edsantiago:
Tagged as WIP because I just guessed at the version to use as baseline. The full list of tagged podman images is:
```console
$ bin/podman search --list-tags quay.io/podman/stable | awk '$2 ~ /^v/ { print $2}' | sort | column -c 75
v1.4.2  v1.5.0  v1.6    v1.9.0  v2.0.2  v2.1.1  v3.1.2  v3.2.1
v1.4.4  v1.5.1  v1.6.2  v1.9.1  v2.0.6  v2.2.1  v3.2.0  v3.2.2
```
Corrections welcome.

cevich:
> just guessed at the version to use as baseline

Suggestion: Perhaps this should follow (or resemble) the list of releases we test on a daily basis via cirrus-cron?  Since those releases still also have "alive and kicking" CI setups w/ stable VM images to play on if needed.  You should have access to see [this list, scroll to the bottom of the configuration page.](https://cirrus-ci.com/settings/repository/6707778565701632)

Note: I just recently pruned the cirrus-cron list with @mheon help/guidance.  It's probably not too late to re-activate testing on branches if needed to align with upgrade testing (assuming y'all think this is a good idea).

edsantiago:
I haven't heard from anyone else, so I'll add my (possibly invalid) 2¢:

>Perhaps this should follow (or resemble) the list of releases we test on a daily basis via cirrus-cron?

I think there's a subtle difference between the two: the daily-cron list is a set of podman branches that are (or should be) receiving maintenance updates; the list in the upgrade test is a set of podman versions that _have been_ supported, and, like it or not, may be out in the field for a long time before someone upgrades to a newer version. As such, the cron list will change (additions and removals) over time, whereas the upgrade-test list should only grow, never or rarely remove, and only grow with a carefully curated list. So I think they need to remain separate -- but I may be wrong.

cevich:
Right, "Have been supported" vs. "Currently supported".  Seems a reasonable augment.  Though then with that logic, doesn't it also follow: When we drop a cirrus-cron branch, it needs to be *added* into the upgrade-testing?

I'm not sure how it could be done, but perhaps a more accurate rubric is: Maintain upgrade testing across podman versions which need extra scrutiny (due to the upgrade itself being "complex" or "tricky").  i.e. those with substantial/significant differences in config-files and runtime database schema.

But ya...I think we could benefit from some extra input/opinions here from @containers/podman-maintainers (when time is available).

vrothberg:
Can we make this a PS today?  I feel like I have some knowledge gap and am interested in filling it.

rhatdan:
@edsantiago What is going on with this PR?

edsantiago:
I believe this is still an important and necessary PR, I just have had no response to my question ("what should the actual version string be")

cevich:
Yes, and we need to get rid of those comments.   I attempted to consolidate this task into one VM + for-loop: It failed horribly.  These tests actually need to run against a clean-VM each time.  The other one about only running it from the daily cirrus-cron is also somewhat dangerous...one of these tests breaking in a PR will (hopefully) prevent introducing a new upgrade problem onto the branch.  Even if it never catches anything, that peace-of-mind is worth the extra $0.02 of CPU-time.

So...what should the actual version string be?

edsantiago:
@containers/podman-maintainers in the interests of just getting this in, even if it's not perfect, I've removed the WIP. PTAL and merge if it's good enough.

mheon:
LGTM on my end.

vrothberg (approved):
/lgtm

Issue #6480, opened 2020-06-03
Handle initContainer for Pods


durandx:
/kind feature

A nice feature to have in podman could be handle initContainer in the same way that K8S handle it.

The main problem is that container images and architecture has to be modified between pods in K8s and Podman : K8s can use a dedicated initContainer to properly initialise a pod whereas in podman we have to combine initialisation and core runtime in a single container. This limit the reuse and compatiblity between K8S and podman even for very simple pods.

mheon:
Creating the init containers seems fairly easy, and I think we can order pod startup to ensure that they are the first things to run in the pod without a problem (we already have a built-in startup dependency model), but the behaviour where Kube waits until an init container has *cleanly exited* to start the next one is problematic - all our dependency ordering is based on the dependency moving to the "running" state, and from a glance at the code this won't be easy to fix.

We'd probably need to add the concept of init containers to Libpod pods, and modify the `podman pod start` command to automatically handle this, to make this happen. This does sound useful outside of `play kube` so I'm not opposed to doing it this way.

@haircommander Thoughts?

haircommander:
Theoretically, we could spoof init containers on podman side, and not expose to libpod.
the work flow could be
- loop through and create all init containers
- wait until each have exited
- clear the pod of all but infra container
- create and start all of the normal containers as is happening now

mheon:
If there's an expectation that init containers run more than once - IE, on each start of the pod - I think libpod is a good place. Kube doesn't really hit this, given they never restart a stopped pod, just recreate, but we will.

haircommander:
> If there's an expectation that init containers run more than once - IE, on each start of the pod - I think libpod is a good place. Kube doesn't really hit this, given they never restart a stopped pod, just recreate, but we will.

yeah we'd need to restart the init container process if we restarted the pod



> We'll also have to modify pod status - a pod where everything except the init containers is running is still a running pod, where I think we'd consider it partially stopped.

I think a pod state: initializing would be clear for when the init containers are running. I think the containers could be ignored (or they could be cleaned up) in figuring out if the pod is running with little problem

rhatdan:
@haircommander Are you working on this?

haircommander:
I am not currently working on this!

rhatdan:
@ryanchpowell Want to take a look?

ryanchpowell:
@rhatdan sounds good, looking into!

rhatdan:
This issue is still out there.

[...]

PR #11011, opened 2021-07-21
implement init containers in podman

baude:
this is the first pass at implementing init containers for podman pods.
init containersare made popular by k8s as a way to run setup for pods
before the pods standard containers run.

unlike k8s, we support two styles of init containers: always and
oneshot.  always means the container stays in the pod and starts
whenever a pod is started.  this does not apply to pods restarting.
oneshot means the container runs onetime when the pod starts and then is
removed.

### DIFF
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -659,6 +659,8 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	s.PidFile = c.PidFile
 	s.Volatile = c.Rm
 
+	// Initcontainers
+	s.InitContainerType = c.InitContainerType
 	return nil
 }
 
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -447,6 +447,21 @@ content that disappears when the container is stopped.
 
 Run an init inside the container that forwards signals and reaps processes.
 
+#### **--init-ctr**=*type* (pods only)
+
+When using pods, create an init style container, which is run after the infra container is started
+but before regular pod containers are started.  Init containers are useful for running
+setup operations for the pod's applications.
+
+Valid values for `init-ctr` type are *always* or *oneshot*.  The *always* value
+means the container will run with each and every `pod start`, whereas the *oneshot*
+value means is will ony run once when the pod is started and then the container is
+removed.
+
+Init containers are only run on pod `start`.  Restarting a pod will not execute any init
+containers should they be present.  Furthermore, init containers can only be created in a
+pod when that pod is not running.
+
 #### **--init-path**=*path*
 
 Path to the container-init binary.
--- a/libpod/container_config.go
+++ b/libpod/container_config.go
@@ -375,4 +375,7 @@ type ContainerMiscConfig struct {
 	CDIDevices []string `json:"cdiDevices,omitempty"`
 	// EnvSecrets are secrets that are set as environment variables
 	EnvSecrets map[string]*secrets.Secret `json:"secret_env,omitempty"`
+	// InitContainerType specifies if the container is an initcontainer
+	// and if so, what type: always or oneshot are possible non-nil entries
+	InitContainerType string `json:"init_container_type,omitempty"`
 }
--- a/libpod/options.go
+++ b/libpod/options.go
@@ -1768,6 +1768,21 @@ func WithPidFile(pidFile string) CtrCreateOption {
 	}
 }
 
+// WithInitCtrType indicates the container is a initcontainer
+func WithInitCtrType(containerType string) CtrCreateOption {
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+		// Make sure the type is valid
+		if containerType == define.OneShotInitContainer || containerType == define.AlwaysInitContainer {
+			ctr.config.InitContainerType = containerType
+			return nil
+		}
+		return errors.Errorf("%s is invalid init container type", containerType)
+	}
+}
+
 // Pod Creation Options
 
 // WithInfraImage sets the infra image for libpod.
--- a/libpod/pod_api.go
+++ b/libpod/pod_api.go
@@ -12,6 +12,45 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// startInitContainers starts a pod's init containers.
+func (p *Pod) startInitContainers(ctx context.Context) error {
+	initCtrs, err := p.initContainers()
+	if err != nil {
+		return err
+	}
+	// Now iterate init containers
+	for _, initCon := range initCtrs {
+		if err := initCon.Start(ctx, true); err != nil {
+			return err
+		}
+		// Check that the init container waited correctly and the exit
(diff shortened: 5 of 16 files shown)


baude:
i dont have the bandwidth to run the e2e tests so pushing here to test.

docs/source/markdown/podman-create.1.md
```diff
@@ -443,6 +443,21 @@ content that disappears when the container is stopped.
 
 Run an init inside the container that forwards signals and reaps processes.
 
+#### **--init-ctr**=*type*
```
mheon, line 446: This needs to be more clear that this only makes sense for pods

libpod/options.go
```diff
@@ -1754,6 +1754,18 @@ func WithPidFile(pidFile string) CtrCreateOption {
 	}
 }
 
+// WithInitCtrType indicates the container is a initcontainer
+func WithInitCtrType(containerType string) CtrCreateOption {
```
mheon, line 1758: In ctr_validate.go - can you validate that InitContainerType is only set if Pod is also set?

libpod/pod_api.go
```diff
@@ -34,26 +62,28 @@ func (p *Pod) Start(ctx context.Context) (map[string]error, error) {
 		return nil, define.ErrPodRemoved
 	}
 
+	ctrErrors := make(map[string]error)
+	ctrsVisited := make(map[string]bool)
```
mheon, line 66: This can move back where it was

libpod/pod_api.go
```diff
@@ -449,12 +479,18 @@ func (p *Pod) Status() (map[string]define.ContainerStatus, error) {
 	if !p.valid {
 		return nil, define.ErrPodRemoved
 	}
-
+	noInitCtrs := make([]*Container, 0)
```
mheon, line 482: Should do `make([]*Container, 0, len(allCtrs))` (and do this a few lines down so allCtrs exists)

baude: > Should do `make([]*Container, 0, len(allCtrs))` (and do this a few lines down so allCtrs exists)

I don't believe this can be done because not allCtrs go into InitCtrs.

mheon:
Having less containers than max capacity of the array is fine

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/6480

rhatdan:
@zhangguanzhang Are you reviewing the code?

zhangguanzhang:
I test it on my machine, it works ok.

baude:
review or merge me please

libpod/pod_api.go
```diff
@@ -12,6 +12,45 @@ import (
...
 )
 
+// startInitContainers starts a pod's init containers.
+func (p *Pod) startInitContainers(ctx context.Context) error {
+	initCtrs, err := p.initContainers()
+	if err != nil {
+		return err
+	}
+	// Nothing to do ?
+	if initCtrLen := len(initCtrs); initCtrLen < 1 {
```
haircommander, line 22: nit: why allocate a variable here?

baude: thats how I roll ... no, actually previous version of the code actually used the variable but that code was removed. so this could be simplified to a simple if statement, absolutely.

haircommander:
one drive-by nit, otherwise LGTM.

something to explore may be extending the pod state (status?) to support an initializing state (while init containers are running) so a `pod ps` shows it's initializing. `kubectl` does so by writing `Init:0/2` if the two init containers are initializing.

cmd/podman/common/specgen.go
```diff
@@ -659,6 +659,10 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	s.PidFile = c.PidFile
 	s.Volatile = c.Rm
 
+	// Initcontainers
```
rhatdan, line 662: Replace this block with simply
 		s.InitContainerType = c.InitContainerType

libpod/pod_api.go
```diff
@@ -12,6 +12,45 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// startInitContainers starts a pod's init containers.
+func (p *Pod) startInitContainers(ctx context.Context) error {
+	initCtrs, err := p.initContainers()
+	if err != nil {
+		return err
+	}
+	// Nothing to do ?
```
rhatdan, line 21: This check is unecessary.

rhatdan (approved):
A couple of small changes.

libpod/container_config.go
```diff
@@ -375,4 +375,7 @@ type ContainerMiscConfig struct {
 	CDIDevices []string `json:"cdiDevices,omitempty"`
 	// EnvSecrets are secrets that are set as environment variables
 	EnvSecrets map[string]*secrets.Secret `json:"secret_env,omitempty"`
+	// InitContainerType specifies if the container is an initcontainer
+	// and if so, what type: always or oneshot are possible non-nil entries
+	InitContainerType string `json:"init_container_type"`
```
mheon, line 380: This should be `omitempty`

libpod/options.go
```diff
@@ -1768,6 +1768,17 @@ func WithPidFile(pidFile string) CtrCreateOption {
...
+	return func(ctr *Container) error {
+		if ctr.valid {
+			return define.ErrCtrFinalized
+		}
+		ctr.config.InitContainerType = containerType
```
mheon, line 1777: Can you validate that the type is one of the InitContainerTypes constants here?

rhatdan:
LGTM

libpod/pod_api.go
```diff
@@ -12,6 +12,41 @@ import (
...
+		if err != nil {
+			return err
+		}
+		if rc != 0 {
+			return errors.Errorf("init container %s exited with code %d", initCon.ID(), rc)
+		}
+		// If the container is an oneshot init container, we need to remove it
+		// after it runs
+		if initCon.Config().InitContainerType == define.OneShotInitContainer {
+			if err := p.runtime.removeContainer(ctx, initCon, false, false, true); err != nil {
```
mheon, line 38: You need to lock initCon before this

mheon: (Make sure to assign it to a local variable, because defers in loops like this have issues if you do a `defer initCon.lock.Unlock()`

flouthoc:
@baude @mheon @zhangguanzhang @rhatdan Just a suggestion not sure if it makes sense,  how about renaming ****--init-ctr**=*type*  **`oneshot`** to **`once`** i have seen **once** being used very often in **init-container extensions** for example: https://github.com/hashicorp/consul-template/blob/master/docs/modes.md#once-mode but i like `oneshot` also. :smiley: .

sshnaidm:
> @baude @mheon @zhangguanzhang @rhatdan @haircommander Just a suggestion not sure if it makes sense, how about renaming ****--init-ctr**=_type_ **`oneshot`** to **`once`** i have seen **once** being used very often in **init-container extensions** for example: https://github.com/hashicorp/consul-template/blob/master/docs/modes.md#once-mode but i like `oneshot` also. smiley .

I'd agree, **once** is a more common term.

rhatdan:
once is better.

PR #11013, opened 2021-07-21
[CI:DOCS] Add notes to flags not supported on cgroups V2

hshiina:
Clarify what flags are not supported on cgroups V2 in documentation.
The following flags of `create` and `run` commands are not supported:
- --cpu-rt-period
- --cpu-rt-runtime
- --kernel-memory
- --memory-swappiness

Signed-off-by: Hironori Shiina <shiina.hironori@jp.fujitsu.com>

### DIFF
--- a/docs/source/markdown/podman-create.1.md
+++ b/docs/source/markdown/podman-create.1.md
@@ -181,6 +181,8 @@ Limit the CPU real-time period in microseconds
 
 Limit the container's Real Time CPU usage. This flag tell the kernel to restrict the container's Real Time CPU usage to the period you specify.
 
+This flag is not supported on cgroups V2 systems.
+
 #### **--cpu-rt-runtime**=*microseconds*
 
 Limit the CPU real-time runtime in microseconds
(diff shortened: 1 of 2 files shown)


docs/source/markdown/podman-create.1.md
```diff
@@ -181,6 +181,8 @@ Limit the CPU real-time period in microseconds
 
 Limit the container's Real Time CPU usage. This flag tell the kernel to restrict the container's Real Time CPU usage to the period you specify.
 
+This flag is not supported on cgroups V2.
```
mheon, line 184: Can you add "systems" to the end - so "not supported on cgroups v2 systems."

mheon:
@TomSweeneyRedHat PTAL

rhatdan:
LGTM
Would love a head nod from @giuseppe to confirm the lack of support of these options.

rhatdan:
BTW Do we give a decent error when a user attempts one of these flags?   Something like this option is not supported on cgroup v2 systems, as opposed to some random file object does not exist in the cgroup file system?

mheon:
Podman itself won't give any error - we rely on the OCI runtime. Which probably means it's the latter.

rhatdan:
Well Podman could check if it is running in cgroupv2 and then report errors on the unsupported options.  Although this would not work  as easily on the --remote.

ashley-cui (commented):
/lgtm

PR #10861, opened 2021-07-05
Add prune until filter test for podman volume cli

jmguzik:
This commit follows work started in #10756. Changes made in #11015
enabled cli support for volume prune --filter until. Adding e2e test
 closes #10579.

Fixes #10579

PR #11015, opened 2021-07-21
Add until filter to volume ls filters list

jmguzik:
As a conclusion of a discussion in #10861, until filter is added
by this commit to volume ls filters.


mheon:
/approve
LGTM

Issue #10994, opened 2021-07-20
podman-3.2.3 import result image is empty


babykart:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

podman-3.2.3 import result image is empty 

**Steps to reproduce the issue:**

1.
```
wget https://mirror.init7.net/gentoo/releases/amd64/autobuilds/current-stage3-amd64-hardened+nomultilib/stage3-amd64-hardened+nomultilib-20210630T214504Z.tar.xz
```

2.
```
xzcat stage3-amd64-hardened+nomultilib-20210630T214504Z.tar.xz | bzip2 | podman import - kubler-gentoo/stage3-amd64-hardened-nomultilib:20210630T214504Z
```

3.

**Describe the results you received:**

```
Getting image source signatures
Copying blob e3b0c44298fc done
Copying config 4c17b62dd9 done
Writing manifest to image destination
Storing signatures
sha256:ad8d30dbb5cc8c1c93d668cc6688f0301179270a3e19fc58c1c4f77a55604b31
```

```
podman images
localhost/kubler-gentoo/stage3-amd64-hardened-nomultilib  latest            ad8d30dbb5cc  2 minutes ago      1.09 kB
```

**Describe the results you expected:**

```
Getting image source signatures
Copying blob ceeec99a2999 done
Copying config e08265be09 done
Writing manifest to image destination
Storing signatures
e08265be09ba19106937104d11f71a3927d1b031584b4ed0522883fcf8232b9d
```

```
podman images
localhost/kubler-gentoo/stage3-amd64-hardened-nomultilib  latest            e08265be09ba  4 minutes ago       905 MB
```

**Additional information you deem important (e.g. issue happens only occasionally):**

It appears to be relative to the size of the archive because if I try a smaller archive, it works perfectly with 3.2.X

```
localhost/kubler-gentoo/portage                           latest            174eb8a7c4a7  2 hours ago     372 MB
```

**Output of `podman version`:**

```
Version:      3.2.3
API Version:  3.2.3
Go Version:   go1.16.6
Git Commit:   1e6fd46e91b21342f9454cf8105a92b90e398c52
Built:        Tue Jul 20 16:21:49 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.21.3
  cgroupControllers:
  - cpuset
  - cpu
  - cpuacct
  - blkio
[...]

PR #11018, opened 2021-07-22
import: write stdin to tmp file

vrothberg:
If importing an archive via stdin write it to a temporary file such that
the temporary file can be opened multiple times later on.  Otherwise, we
may end up with an empty image.

Add system tests for `podman import` exercising the basics.

Fixes: #10994

### DIFF
--- a/cmd/podman/images/import.go
+++ b/cmd/podman/images/import.go
@@ -3,6 +3,9 @@ package images
 import (
 	"context"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/containers/common/pkg/completion"
@@ -97,6 +100,22 @@ func importCon(cmd *cobra.Command, args []string) error {
 	default:
 		return errors.Errorf("too many arguments. Usage TARBALL [REFERENCE]")
 	}
+
+	if source == "-" {
+		outFile, err := ioutil.TempFile("", "podman")
+		if err != nil {
+			return errors.Errorf("error creating file %v", err)
+		}
+		defer os.Remove(outFile.Name())
+		defer outFile.Close()
+
+		_, err = io.Copy(outFile, os.Stdin)
+		if err != nil {
+			return errors.Errorf("error copying file %v", err)
+		}
+		source = outFile.Name()
+	}
+
 	errFileName := parse.ValidateFileName(source)
 	errURL := parse.ValidURL(source)
 	if errURL == nil {
--- a/cmd/podman/parse/net.go
+++ b/cmd/podman/parse/net.go
@@ -180,9 +180,12 @@ func ValidateFileName(filename string) error {
 
 // ValidURL checks a string urlStr is a url or not
 func ValidURL(urlStr string) error {
-	_, err := url.ParseRequestURI(urlStr)
+	url, err := url.ParseRequestURI(urlStr)
 	if err != nil {
-		return errors.Wrapf(err, "invalid url path: %q", urlStr)
+		return errors.Wrapf(err, "invalid url %q", urlStr)
+	}
+	if url.Scheme == "" {
+		return errors.Errorf("invalid url %q: missing scheme", urlStr)
 	}
 	return nil
 }
--- a/test/system/125-import.bats
+++ b/test/system/125-import.bats
@@ -1,0 +1,45 @@
+#!/usr/bin/env bats   -*- bats -*-
+#
+# tests for podman import
+#
+
+load helpers
+
+@test "podman import" {
+    local archive=$PODMAN_TMPDIR/archive.tar
+    local random_content=$(random_string 12)
+    # Generate a random name and tag (must be lower-case)
+    local random_name=x0$(random_string 12 | tr A-Z a-z)
+    local random_tag=t0$(random_string 7 | tr A-Z a-z)
+    local fqin=localhost/$random_name:$random_tag
+
+    run_podman run --name import $IMAGE sh -c "echo ${random_content} > /random.txt"
+    run_podman export import -o $archive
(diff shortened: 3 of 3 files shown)


vrothberg:
@rhatdan @TomSweeneyRedHat shall we create a BZ for this one?

It's a regression in 3.2 where archives cannot be imported via stdin.

rhatdan:
If we want to get it fixed in rhel8.4.0.2 then it needs a bugzilla.

cmd/podman/images/import.go
```diff
@@ -97,6 +101,22 @@ func importCon(cmd *cobra.Command, args []string) error {
 	default:
 		return errors.Errorf("too many arguments. Usage TARBALL [REFERENCE]")
 	}
+
+	if source == "-" {
+		outFile, err := ioutil.TempFile(util.Tmpdir(), "podman")
```
rhatdan, line 106: I don't think you need the util.TmpDir() here, why not just pass ""?

rhatdan: Can't you just set source to "/dev/stdin"?

vrothberg: > Can't you just set source to "/dev/stdin"?

That's what we did before which caused the regression.  This code is directly copied from `load`; I think `util.TempDir()` is an extra security net for TMPDIR and falling back to /var/tmp.

vrothberg:
> If we want to get it fixed in rhel8.4.0.2 then it needs a bugzilla.

Do you want it fixed in 8.4.0.2?

cmd/podman/images/import.go
rhatdan: So ioutil.TempFile does not use $TMPDIR?

rhatdan:
I am not sure if there is time.  If so then yes.

cmd/podman/images/import.go
rhatdan: https://pkg.go.dev/os#TempDir
Says that ioutil.TmpFile will end up using TMPDIR if set, but fall back to /tmp.
But we already force TMPDIR to always be set.

https://github.com/containers/podman/blob/main/cmd/podman/root.go#L178

edsantiago:
Tests LGTM, and I agree that there's no sane way to write them in a table or for-loop.

I have one enormous concern: your new test passes on podman *without* your patch:
```console
$ git co main
$ make
...
$ git co pr/11018
$ ./bats import
# bats  test/system/125-import.bats
 ✓ podman import

1 test, 0 failures

--------------------------------------------------
$ bats  test/system/125-import.bats
 ✓ podman import

1 test, 0 failures
```
It also passes with `PODMAN=/usr/bin/podman ./bats import` (podman-3.2.2-1.fc34). This means that your test is not testing what you think it is testing.

edsantiago:
I tried the obvious fix:
```diff
-    run_podman import -q - < $archive
+    run_podman import -q - <(cat $archive)
```
...but that just causes podman (unpatched & patched) to hang. Probably one of those BATS weirdnesses. I need to think about this. For now, please do not merge.

edsantiago:
Suggestion for test:
```diff
diff --git a/test/system/125-import.bats b/test/system/125-import.bats
index 596884668..507bbab56 100644
--- a/test/system/125-import.bats
+++ b/test/system/125-import.bats
@@ -25,8 +25,9 @@ load helpers
     run_podman rmi -f $iid

     # Simple import via stdin
-    run_podman import -q - < $archive
-    iid="$output"
+    cat $archive | run_podman import -q -
+    run_podman images -a --format '{{.Id}}' --sort created
+    iid="${lines[0]}"
     run_podman run -t --rm $iid cat /random.txt
     is "$output" "$random_content" "simple import via stdin"
     run_podman rmi -f $iid
@@ -39,8 +40,7 @@ load helpers
     run_podman rmi -f $fqin

     # Tagged import via stdin
-    run_podman import -q - $fqin < $archive
-    iid="$output"
+    cat $archive | run_podman import -q - $fqin
     run_podman run -t --rm $fqin cat /random.txt
     is "$output" "$random_content" "tagged import via stdin"
     run_podman rmi -f $fqin
```

edsantiago:
Solved it! This is the solution:
```diff
diff --git a/test/system/125-import.bats b/test/system/125-import.bats
index 596884668..c53711618 100644
--- a/test/system/125-import.bats
+++ b/test/system/125-import.bats
@@ -25,7 +25,7 @@ load helpers
     run_podman rmi -f $iid

     # Simple import via stdin
-    run_podman import -q - < $archive
+    run_podman import -q - < <(cat $archive)
     iid="$output"
     run_podman run -t --rm $iid cat /random.txt
     is "$output" "$random_content" "simple import via stdin"
@@ -33,14 +33,12 @@ load helpers

     # Tagged import
     run_podman import -q $archive $fqin
-    iid="$output"
     run_podman run -t --rm $fqin cat /random.txt
     is "$output" "$random_content" "tagged import"
     run_podman rmi -f $fqin

     # Tagged import via stdin
-    run_podman import -q - $fqin < $archive
-    iid="$output"
+    run_podman import -q - $fqin < <(cat $archive)
     run_podman run -t --rm $fqin cat /random.txt
     is "$output" "$random_content" "tagged import via stdin"
     run_podman rmi -f $fqin
```
(I had forgotten one `<`).

With this fix, the test now fails in podman-without-your-patch:
```
podman run -t --rm <sha> cat /random.txt
   Error: executable file `cat` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

vrothberg:
@edsantiago, _thank you_!

vrothberg:

TomSweeneyRedHat:
LGTM
@edsantiago ?

PR #11021, opened 2021-07-22
Add IPVLAN network support to podman network.

ananthb:
podman network learns the modified `--driver=ipvlan` option which will create an ipvlan network using the relevant CNI plugin.\
Additionally, both macvlan and ipvlan drivers gain the ability to set network mode via the `macvlan_mode` and `ipvlan_mode` driver options.

### DIFF
--- a/libpod/network/config.go
+++ b/libpod/network/config.go
@@ -22,9 +22,9 @@ const (
 	// LockFileName is used for obtaining a lock and is appended
 	// to libpod's tmpdir in practice
 	LockFileName = "cni.lock"
-	// PluginTypeMacVLAN
+	// PluginTypeMacVLAN is to create macvlan networks.
 	PluginTypeMacVLAN = "macvlan"
-	// PluginTypeIPVLAN
+	// PluginTypeIPVLAN is used to create ipvlan networks
 	PluginTypeIPVLAN = "ipvlan"
 )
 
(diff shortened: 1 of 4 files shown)


Luap99 (commented):
It looks like there is no difference between a macvlan config and ipvlan config other than the plugin name. I think it is better to just add a new parameter to createMacvlan() and set PluginType to `ipvlan`

TomSweeneyRedHat:
It would be nice to have at least a little bit of description on what this PR does in the description.  You'll eventually need to add some tests too so the CI won't bark at you.

ananthb:
Yep @TomSweeneyRedHat, I'm working on it.

ananthb:
@Luap99 yeah that makes sense. I also want to add macvlan_mode and ipvlan_mode support like docker. These will go in `podman network create -d ipvlan -o ipvlan_mode=l2` or `podman network create -d macvlan -o mode macvlan_mode=...`. The modes will be different between ipvlan and macvlan but that should be easy to account for with one function creating both.

ananthb:
Hey @Luap99, I'm re-using code from macvlan networks for ipvlan. If this seems ok, I'll add tests to validate macvlan and ipvlan network options similar to how bridge network options are being validated right now. Does this make sense?

Luap99:
Thanks @ananthb. Sorry for not getting back earlier. I am currently reworking the network code (see #11232) and this code will go away soon so I am not sure if this worth merging. I think we can easily add support for the IPVLAN driver to new code once #11232 is merged.

ananthb:
Oh! I'll close this PR then.

PR #11025, opened 2021-07-22
[CI:DOCS] Fix handling of shadow-utils

rhatdan:
There seems to be a bug in rpm, where it fails silently if you specify
rpm --restore --quiet shadow-utils.

rpm --restore shadow-utils 2> /dev/null

Does the right thing.

[NO TESTS NEEDED] Might add tests from buildah, once we have them
working correctly.

### DIFF
--- a/contrib/podmanimage/stable/Dockerfile
+++ b/contrib/podmanimage/stable/Dockerfile
@@ -11,7 +11,7 @@ FROM registry.fedoraproject.org/fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN dnf -y update; rpm --restore --quiet shadow-utils; \
+RUN dnf -y update; rpm --restore shadow-utils 2>/dev/null; \
 yum -y install podman fuse-overlayfs --exclude container-selinux; \
 rm -rf /var/cache /var/log/dnf* /var/log/yum.*
 
--- a/contrib/podmanimage/upstream/Dockerfile
+++ b/contrib/podmanimage/upstream/Dockerfile
@@ -17,7 +17,7 @@ ENV GOPATH=/root/podman
 # to the container.
 # Finally remove the podman directory and a few other packages
 # that are needed for building but not running Podman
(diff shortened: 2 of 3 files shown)


rhatdan:
@nalind PTAL

contrib/podmanimage/upstream/Dockerfile
```diff
@@ -17,7 +17,7 @@ ENV GOPATH=/root/podman
 # to the container.
 # Finally remove the podman directory and a few other packages
 # that are needed for building but not running Podman
-RUN yum -y update; rpm --restore --quiet shadow-utils; yum -y install --exclude container-selinux \
+RUN dnf -y update; rpm --restore shadow-utils 2>/dev/null;  yum -y install --exclude container-selinux \
```
nalind, line 20: Is the change from `yum` to `dnf` here intentional?  The other call on this line, and in the ones in the other Dockerfiles, aren't changed.

nalind:
This should work, but `--setcaps` doesn't produce error messages that need to be silenced, so it may be a better choice.

contrib/podmanimage/stable/Dockerfile
```diff
@@ -11,7 +11,7 @@ FROM registry.fedoraproject.org/fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN dnf -y update; rpm --restore --quiet shadow-utils; \
+RUN dnf -y update; rpm --restore shadow-utils 2>/dev/null; \
```
TomSweeneyRedHat, line 14: and this probably ought to be yum too, although it's preexisting.

contrib/podmanimage/upstream/Dockerfile
rhatdan: No this is just a cut and paste error, although not sure why these are inconsistent.  I will leave it as yum.

rhatdan:
I was going with --restore, since it seemed closer to the reinstall.  IE it would fix other potential issues in the permissions of the package.

TomSweeneyRedHat:
LGTM
and happy green test buttons

vrothberg (commented):
/lgtm

PR #11026, opened 2021-07-22
wait_for_port: use filehandle 5, not 3

edsantiago:
...because 3 is reserved by BATS. Using 3 causes a very
unpleasant failure mode.


edsantiago:
@containers/podman-maintainers please treat as high priority, this is silently disabling one of our tests.

rhatdan:
LGTM If tests pass.

edsantiago:
This approach is not going to work.

Issue #11031, opened 2021-07-24
podman generates redundant rules (which are rejected by runtime)


kolyshkin:
This issue was reported quite a few times (for the last month: #10735, #10885, #11005, #11029).

TL;DR: when trying to start container, you get an error like

> OCI runtime error: container_linux.go:380: starting container process caused: error adding seccomp filter rule for syscall bdflush: permission denied: OCI permission denied

or

> OCI runtime error: container_linux.go:380: starting container process caused: error adding seccomp filter rule for syscall bdflush: requested action matches default action of filter

or something similar.

One particular cause of this is, since containers-common-4:1-20 rpm in F34 
 (see commit https://src.fedoraproject.org/rpms/containers-common/c/0b37657a909b2892af242e2875b39e5496d7ca42, possibly same in RHEL7 and 8), seccomp.json specifies
```
"defaultErrnoRet": 38,
```
plus a number of syscalls with
```
"action": "SCMP_ACT_ERRNO",
"errnoRet": 1
```

This comes from https://github.com/containers/common/pull/573 (which is all good).

The problem is, containers/common is also used by podman to generate config.json for container to be started, and before https://github.com/containers/common/pull/573 it silently ignores `defaultErrnoRet`, not propagating it from system's seccomp.json to container's config.json. As a result, we have set of rules like this:

```json
{
  "defaultAction": "SCMP_ACT_ERRNO",
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    {
      "names": [
        "bdflush",
        "io_pgetevents",
        "kexec_file_load",
        "kexec_load",
        .....
      ],
      "action": "SCMP_ACT_ERRNO",
      "errnoRet": 1
    },
```

Note there's no `defaultErrnoRet`, and this defaults to `1` (aka EPERM). As a result, we have a set of rules (for bdflush etc.) which have the same action as the defaultAction. Unfortunately, seccomp (and libseccomp-golang) rejects those, and so does runc (before https://github.com/opencontainers/runc/pull/3109). OTOH crun skips those redundant rules (since https://github.com/containers/crun/commit/08229f3fb904c5ea19a7d9).

As a result of all this, crun works, while other runtimes don't.

From the practical perspective, it will be faster to fix this in podman. To do so for v3.2, we need https://github.com/containers/common/pull/689 and, once merged and a release is cut, vendor it to v3.2.

[...]

PR #11033, opened 2021-07-24
[3.2] seccomp: propagate defaultErrnoRet

kolyshkin:
This fixes the issue of not propagating `defaultErrnoRet` from
system's seccomp.json to container's config.json, which can
result in redundant seccomp rules and inability to start a container
when using some runtimes (notably runc up to and including 1.0.1).

The fix is to vendor in containers/common with the fixed pkg/seccomp
which no longer ignores defaultErrnoRet.

Fixes: https://github.com/containers/podman/issues/11031

### Proposed changelog entry
```markdown
Bugfixes:
 * Fixed a bug of ignoring the `defaultErrnoRet` field in system's seccomp.json,
   which may cause an "error adding seccomp filter rule for syscall" error
   and a failure to start a container when using `--runtime=runc` (#11031).
 ```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/buildah v1.21.3
-	github.com/containers/common v0.38.16
+	github.com/containers/common v0.38.18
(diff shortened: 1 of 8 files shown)


kolyshkin:
no longer a draft

PR #11039, opened 2021-07-25
[3.2.3-rhel] seccomp: propagate defaultErrnoRet

kolyshkin:
_(same as https://github.com/containers/podman/pull/11033 but for v3.2.3-rhel branch)_

This fixes the issue of not propagating defaultErrnoRet from
system's seccomp.json to container's config.json, which can
result in redundant seccomp rules and inability to start a container
when using some runtimes (notably runc up to and including 1.0.1).

The fix is to vendor in containers/common with the fixed pkg/seccomp
which no longer ignores defaultErrnoRet.

Fixes:  #11031
Fixes: bugzilla.redhat.com/1987049
### Proposed changelog entry
```markdown
Bugfixes:
 * Fixed a bug of ignoring the `defaultErrnoRet` field in system's seccomp.json,
   which may cause an "error adding seccomp filter rule for syscall" error
   and a failure to start a container when using `--runtime=runc` (#11031).
 ```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/buildah v1.21.3
-	github.com/containers/common v0.38.16
+	github.com/containers/common v0.38.18
 	github.com/containers/conmon v2.0.20+incompatible
 	github.com/containers/image/v5 v5.12.0
 	github.com/containers/ocicrypt v1.1.1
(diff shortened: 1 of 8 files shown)


mheon:
Do we have a new-enough c/common in 8.4.0.2 to require this? If so, I think
we definitely need a BZ spun up to get this landed.

kolyshkin:
> Do we have a new-enough c/common in 8.4.0.2 to require this? If so, I think
> we definitely need a BZ spun up to get this landed.

I assume we don't. As long as the vendored version is used for building podman in rhel (which I believe is how things are), we do not care about what version of containers-common rpm is installed.

vrothberg:
Yes, we need a BZ to merge any PR into the -rhel branch (also to start rebuilding packages). @kolyshkin, would you open one?

/hold

kolyshkin:
> Yes, we need a BZ to merge any PR into the -rhel branch (also to start rebuilding packages). @kolyshkin, would you open one?

Done, please see bugzilla.redhat.com/1987049

Description updated.

vrothberg (commented):
/lgtm

Let's wait for an ACK in the BZ before merge. Thank you, Kir!

rhatdan:
/hold cancel
Looks like we have acks.

PR #11045, opened 2021-07-26
[v3.2.3-rhel] play kube: support capitalized pull policy

vrothberg:
Pull policies in K8s yaml may be capitalized, so lower them before
parsing.

Fixes: bugzilla.redhat.com/show_bug.cgi?id=1985905

### DIFF
--- a/pkg/domain/infra/abi/play.go
+++ b/pkg/domain/infra/abi/play.go
@@ -275,7 +275,10 @@ func (ic *ContainerEngine) playKubePod(ctx context.Context, podName string, podY
 		// registry on localhost.
 		pullPolicy := config.PullPolicyNewer
 		if len(container.ImagePullPolicy) > 0 {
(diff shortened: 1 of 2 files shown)


vrothberg:
/hold
Need to wait for an ACK in the BZ before we can merge.

vrothberg:
@TomSweeneyRedHat @rhatdan PTAL

Have a qa ack on it.

cdoern:
LGTM, but is a fully capitalized or camel capitalized pull allowed because this will read that as valid as well.

vrothberg:
> LGTM, but is a fully capitalized or camel capitalized pull allowed because this will read that as valid as well.

Probably only the camel case (see https://docs.openshift.com/container-platform/4.1/openshift_images/managing_images/image-pull-policy.html). Very good catch, @cdoern!

vrothberg:
Looks like we moved it over to be fixed with 3.3, so I can close here.  Will massage libimage parsing a bit to accept the camel cases

PR #11048, opened 2021-07-26
Fixed Healthcheck formatting, string to []string

### LINKED ISSUE #10617: podman docker-compose healthchecks issue
I am running podman 3.1.2-dev on a custom embedded Linux device.  I have a docker-compose.yml file with an application named '**app1**', which is a simple Python3 Flask webserver.  Podman successfully brings up the docker-compose environment, but my healthchecks always return error, regardless of what I use for timeouts, intervals, retries, etc. My defined healthcheck works when I exec into the container and run them manually. When I inspect the application with '**podman inspect app1**', I can see 2 '**CMD-SHELL**' instances defined. I believe it should only have mine and perhaps this is the issue.

Output of **podman inspect app1**:

```
"Healthcheck": {
    "Test": [
        "CMD-SHELL",
        "CMD-SHELL /usr/bin/wget http://127.0.0.1 || exit 1"
    ],
    "StartPeriod": 300000000000,
    "Interval": 30000000000,
    "Timeout": 60000000000,
    "Retries": 5
},

"Healthcheck": {
        "Status": "unhealthy",
        "FailingStreak": 21,
        "Log": [
            {
                "Start": "2021-06-09T14:53:29.192875476Z",
                "End": "2021-06-09T14:53:31.211062949Z",
                "ExitCode": 1,
                "Output": ""
            },
            {
                "Start": "2021-06-09T14:54:07.933818178Z",
                "End": "2021-06-09T14:54:10.368276033Z",
                "ExitCode": 1,
                "Output": ""
            },
            {
                "Start": "2021-06-09T14:54:47.115168651Z",
                "End": "2021-06-09T14:54:49.119256984Z",
                "ExitCode": 1,
                "Output": ""
            },
            {
                "Start": "2021-06-09T14:55:26.183508777Z",
                "End": "2021-06-09T14:55:28.038064193Z",
                "ExitCode": 1,
                "Output": ""
            },
            {
                "Start": "2021-06-09T14:56:02.115129058Z",
                "End": "2021-06-09T14:56:03.41780082Z",
                "ExitCode": 1,
                "Output": ""
            }
        ]
    }
},
```

Manually running healthchecks:

```
root@robot:~# podman healthcheck run app1
unhealthy
root@robot:~# 
root@robot:~# podman exec -it app1 sh
/ # 
/ # /usr/bin/wget http://127.0.0.1 || exit 1
Connecting to 127.0.0.1 (127.0.0.1:80)
saving to 'index.html'
index.html           100% |*******************************************************************************|    39  0:00:00 ETA
'index.html' saved
/ # 
/ # echo $?
0
/ # exit
root@robot:~# 

root@robot:~# podman --version
podman version 3.2.1-dev
root@robot:~# 
```

docker-compose.yml file:

```
services:
  app1:
    container_name: app1
    hostname: app1
    image: app1:1.0
    build: ./app1
    init: true
    entrypoint:
      - "python3"
      - "/app/app.py"
    security_opt:
      - seccomp=unconfined
    tty: true
    healthcheck:
      test: /usr/bin/wget http://127.0.0.1 || exit 1
      interval: 30s
      timeout: 60s
      retries: 5
      start_period: 300s
    restart: always
    volumes:
      - ./app1:/app
      - ./shared_data:/shared_data
    environment:
      HEALTHCHECK_DEBUG: 1
    cpu_count: 1
    cpus: 0.95
    cpu_percent: 10
    mem_limit: 25m
    networks:
      - robot_internal_net
```

cdoern:
Compat healthcheck tests are of the format []string but podman's were of
the format string. Converted podman's to []string at the specgen level since it has the same effect
and removed the incorrect parsing of compat healthchecks.

fixes #10617

### DIFF
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -521,7 +521,9 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, rtc *c
 		for _, str := range cc.Config.Healthcheck.Test {
 			finCmd = finCmd + str + " "
 		}
-		finCmd = finCmd[:len(finCmd)-1]
+		if len(finCmd) > 1 {
+			finCmd = finCmd[:len(finCmd)-1]
+		}
 		cliOpts.HealthCmd = finCmd
 		cliOpts.HealthInterval = cc.Config.Healthcheck.Interval.String()
 		cliOpts.HealthRetries = uint(cc.Config.Healthcheck.Retries)
--- a/cmd/podman/common/create_opts.go
+++ b/cmd/podman/common/create_opts.go
@@ -517,7 +517,12 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, rtc *c
 		cliOpts.OOMKillDisable = *cc.HostConfig.OomKillDisable
 	}
 	if cc.Config.Healthcheck != nil {
-		cliOpts.HealthCmd = strings.Join(cc.Config.Healthcheck.Test, " ")
+		finCmd := ""
+		for _, str := range cc.Config.Healthcheck.Test {
+			finCmd = finCmd + str + " "
+		}
+		finCmd = finCmd[:len(finCmd)-1]
+		cliOpts.HealthCmd = finCmd
 		cliOpts.HealthInterval = cc.Config.Healthcheck.Interval.String()
 		cliOpts.HealthRetries = uint(cc.Config.Healthcheck.Retries)
 		cliOpts.HealthStartPeriod = cc.Config.Healthcheck.StartPeriod.String()
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -516,7 +516,6 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 			if len(con) != 2 {
 				return fmt.Errorf("invalid --security-opt 1: %q", opt)
 			}
-
 			switch con[0] {
 			case "apparmor":
 				s.ContainerSecurityConfig.ApparmorProfile = con[1]
@@ -655,20 +654,32 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	return nil
 }
 
-func makeHealthCheckFromCli(inCmd string, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
-	var cmdArr []string
-	cmdSplitPrelim := strings.SplitN(inCmd, " ", 2)
-	if cmdSplitPrelim[0] == "CMD-SHELL" {
-		cmdArr = cmdSplitPrelim
-	} else if cmdSplitPrelim[0] != "CMD" && cmdSplitPrelim[0] != "none" { // if it isnt a cmd-shell or a cmd, it is a command
-		cmdArr = []string{"CMD-SHELL"}
-		cmdArr = append(cmdArr, inCmd)
-	} else {
-		cmdArr = strings.Fields(inCmd)
+func makeHealthCheckFromCli(inCmd, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+	cmdArr := []string{}
+	isArr := true
+	err := json.Unmarshal([]byte(inCmd), &cmdArr) // array unmarshalling
+	if err != nil {
+		cmdArr = strings.SplitN(inCmd, " ", 2) // default for compat
+		isArr = false
 	}
 	// Every healthcheck requires a command
 	if len(cmdArr) == 0 {
 		return nil, errors.New("Must define a healthcheck command for all healthchecks")
+	}
+	concat := ""
+	if cmdArr[0] == "CMD" || cmdArr[0] == "none" { // this is for compat, we are already split properly for most compat cases
+		cmdArr = strings.Fields(inCmd)
+	} else if cmdArr[0] != "CMD-SHELL" { // this is for podman side of things, wont contain the keywords
+		if isArr && len(cmdArr) > 1 { // an array of consecutive commands
+			cmdArr = append([]string{"CMD"}, cmdArr...)
+		} else { // one singular command
+			if len(cmdArr) == 1 {
+				concat = cmdArr[0]
+			} else {
+				concat = strings.Join(cmdArr[0:], " ")
+			}
+			cmdArr = append([]string{"CMD-SHELL"}, concat)
+		}
 	}
 
 	if cmdArr[0] == "none" { // if specified to remove healtcheck
--- a/cmd/podman/common/specgen.go
+++ b/cmd/podman/common/specgen.go
@@ -655,26 +655,29 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	return nil
 }
 
-func makeHealthCheckFromCli(inCmd, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+func makeHealthCheckFromCli(inCmd string, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+	var cmdArr []string
+	cmdSplitPrelim := strings.SplitN(inCmd, " ", 2)
+	if cmdSplitPrelim[0] == "CMD-SHELL" {
+		cmdArr = cmdSplitPrelim
+	} else if cmdSplitPrelim[0] != "CMD" && cmdSplitPrelim[0] != "none" { // if it isnt a cmd-shell or a cmd, it is a command
+		cmdArr = []string{"CMD-SHELL"}
+		cmdArr = append(cmdArr, inCmd)
+	} else {
+		cmdArr = strings.Fields(inCmd)
+	}
 	// Every healthcheck requires a command
-	if len(inCmd) == 0 {
+	if len(cmdArr) == 0 {
 		return nil, errors.New("Must define a healthcheck command for all healthchecks")
 	}
 
-	// first try to parse option value as JSON array of strings...
-	cmd := []string{}
-
-	if inCmd == "none" {
-		cmd = []string{"NONE"}
-	} else {
-		err := json.Unmarshal([]byte(inCmd), &cmd)
-		if err != nil {
-			// ...otherwise pass it to "/bin/sh -c" inside the container
-			cmd = []string{"CMD-SHELL", inCmd}
-		}
+	if cmdArr[0] == "none" { // if specified to remove healtcheck
+		cmdArr = []string{"NONE"}
 	}
+
+	// healthcheck is by default an array, so we simply pass the user input
 	hc := manifest.Schema2HealthConfig{
-		Test: cmd,
+		Test: cmdArr,
 	}
 
 	if interval == "disable" {
--- a/test/e2e/run_test.go
+++ b/test/e2e/run_test.go
@@ -1205,21 +1205,21 @@ USER mail`, BB)
 	})
 
 	It("podman run with bad healthcheck retries", func() {
-		session := podmanTest.Podman([]string{"run", "-dt", "--health-cmd", "[\"foo\"]", "--health-retries", "0", ALPINE, "top"})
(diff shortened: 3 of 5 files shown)


rhatdan:
LGTM
But tests are failing.

test/e2e/run_test.go
```diff
@@ -1205,21 +1205,21 @@ USER mail`, BB)
 	})
 
 	It("podman run with bad healthcheck retries", func() {
-		session := podmanTest.Podman([]string{"run", "-dt", "--health-cmd", "[\"foo\"]", "--health-retries", "0", ALPINE, "top"})
+		session := podmanTest.Podman([]string{"run", "-dt", "--health-cmd", "foo", "--health-retries", "0", ALPINE, "top"})
```
jwhonce, line 1208:
```suggestion
		session := podmanTest.Podman([]string{"run", "-dt", "--health-cmd", "foo", "--health-retries", "0", ALPINE, "top"})
```
With this change does the health-retries error out as expected or the health-cmd not being an array?

cmd/podman/common/specgen.go
```diff
@@ -655,26 +655,20 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	return nil
 }
 
-func makeHealthCheckFromCli(inCmd, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+func makeHealthCheckFromCli(inCmd string, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+	cmdArr := strings.Split(inCmd, " ")
```
jwhonce, line 659:
```suggestion
	cmdArr := strings.Fields(inCmd)
```
Will handle more inputs correctly.

cdoern:
@jwhonce @rhatdan I am realizing now that this would be much simpler and compat if we required users (like docker) to specify `CMD-SHELL` (one command to read in as a whole) vs `CMD` (arg1, arg2 etc). Would this be something you would like me to add? I have a few fixes ready to push for this right now but they are contingent on that.

test/e2e/run_test.go
cdoern: healthCmd is read in as a string and then converted to an array so as long as you pass it a single string with no quotes: `foo` or a long string with quotes `"foo bar 123"` the errors don't change

TomSweeneyRedHat:
LGTM
and happy green test buttons

test/e2e/run_test.go
Luap99: This is a breaking change. Why should `[\"foo\"]` no longer be valid?

cdoern: I can change the parsing so [] is valid. Personally it doesn't make sense to me why we would take an array if our healthchecks on the cmd level are strings, [""] and "" are just as easy to input

Luap99: The point is this is breaking users who already use the `[""]` syntax. Breaking changes should be limited to major releases. It is also documented in the man page that this syntax is valid.

vrothberg (commented):
/lgtm
/hold

test/e2e/run_test.go
cdoern: just changed it so `[""]` is valid @Luap99

Luap99: I think you have to use `json.Unmarshal` as this will handle special escape sequences and more. Also I do not think just removing the first bracket will not work for `["some","long","command"]`

Luap99:
Please do not merge until https://github.com/containers/podman/pull/11048#discussion_r677442979 is resolved.

test/e2e/run_test.go
cdoern: It removes both brackets. The array is technically a string so long commands wont matter (I don't think) as it reads it as one consecutive string.

Luap99: `strings.Fields(inCmd)` will only split at whitespaces but in this case it has to split at the comma and also remove the quotes. Therefore you should use `json.Unmarshal`.

cdoern: resolving now.

cmd/podman/common/specgen.go
```diff
@@ -655,26 +655,41 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
+		isArr = true
+		err := json.Unmarshal([]byte(inCmd), cmdArr)
+		if err != nil {
+			return nil, err
+		}
+		fmt.Println(cmdArr)
+		cmdSplit = *cmdArr
+		fmt.Println(cmdSplit)
+	}
+	if cmdSplit[0] != "CMD" && cmdSplit[0] != "none" { // if it isnt a cmd-shell or a cmd, it is a command
```
Luap99, line 672: Also need to check for `CMD-SHELL`.

cmd/podman/common/specgen.go
```diff
@@ -655,26 +655,41 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
+	cmdSplit := strings.SplitN(inCmd, " ", 2)
+	if strings.Contains(inCmd, "[") { // given an array, need to unmarshall it into a pointer
+		isArr = true
+		err := json.Unmarshal([]byte(inCmd), cmdArr)
+		if err != nil {
+			return nil, err
+		}
+		fmt.Println(cmdArr)
+		cmdSplit = *cmdArr
+		fmt.Println(cmdSplit)
```
Luap99, line 670: I think this is leftover from debugging, please remove both print statements.

cmd/podman/common/create_opts.go
```diff
@@ -517,7 +517,12 @@ func ContainerCreateToContainerCLIOpts(cc handlers.CreateContainerConfig, rtc *c
 		cliOpts.OOMKillDisable = *cc.HostConfig.OomKillDisable
 	}
 	if cc.Config.Healthcheck != nil {
-		cliOpts.HealthCmd = strings.Join(cc.Config.Healthcheck.Test, " ")
+		finCmd := ""
+		for _, str := range cc.Config.Healthcheck.Test {
+			finCmd = finCmd + str + " "
+		}
+		finCmd = finCmd[:len(finCmd)-1]
```
mheon, line 524: Don't think this is safe for all commands - Manpages say that a length-1 command is allowed to be specified without brackets, so `podman run --health-cmd "./my_script,sh"` is perfectly legal, and will not work with this.

test/e2e/run_test.go
```diff
@@ -1212,14 +1212,14 @@ USER mail`, BB)
 	})
 
 	It("podman run with bad healthcheck timeout", func() {
-		session := podmanTest.Podman([]string{"run", "-dt", "--health-cmd", "[\"foo\"]", "--health-timeout", "0s", ALPINE, "top"})
+		session := podmanTest.Podman([]string{"run", "-dt", "--health-cmd", "foo", "--health-timeout", "0s", ALPINE, "top"})
```
mheon, line 1215: Why change this? Is the syntax no longer legal?

cmd/podman/common/specgen.go
```diff
@@ -655,26 +655,53 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 	return nil
 }
 
-func makeHealthCheckFromCli(inCmd, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+func makeHealthCheckFromCli(inCmd string, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
```
mheon, line 658: Why was the signature changed? Looks like addition of duplicate `string` specifiers, nothing functional

cmd/podman/common/specgen.go
```diff
@@ -655,26 +655,53 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
+	cmdArr := &[]string{}
```
mheon, line 659: The double-indirection here is kind of gross - would be better to specify as `[]string{}` and then add the `&` only to the `json.Unmarshal` call below.

cmd/podman/common/specgen.go
```diff
@@ -655,26 +655,53 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
+	if cmdSplit[0] != "CMD" && cmdSplit[0] != "none" && cmdSplit[0] != "CMD-SHELL" { // if it isnt a cmd-shell or a cmd, it is a command
+		if isArr && len(cmdSplit) > 1 { // an array of consecutive commands
+			cmdSplit = append([]string{"CMD"}, cmdSplit...)
+		} else { // one singular command
+			if len(cmdSplit) == 1 {
+				concat = cmdSplit[0]
+			} else {
+				concat = strings.Join(cmdSplit[1:], " ")
+			}
+			cmdSplit = append([]string{"CMD-SHELL"}, concat)
```
mheon, line 680: I'm confused as to the necessity of this. We've split the command apart on whitespace, but now we're concatenating it together again? Seems unnecessary. Seems like it would be easier to just check for certain prefixes? Especially given that I don't think that `CMD-SHELL` inside a JSON array is actually valid.

cdoern: This is for the purpose of compat vs non compat. The compat side will always include `CMD` or `CMD-SHELL` while most podman users will not. This is similar to how this was done before, there are just more codepaths now, I think all compat healthchecks were broken and duplicating `CMD-SHELL` twice

cmd/podman/common/specgen.go
cdoern: Typo sorry. I was originally doing `[]string` on the cmd level but changed it back to `string`

test/e2e/run_test.go
cdoern: No it is, I just thought It would be good to alternate the tests with strings and arrays to test both

cmd/podman/common/create_opts.go
cdoern: fixed

cmd/podman/common/specgen.go
```diff
@@ -656,25 +655,52 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
 }
 
 func makeHealthCheckFromCli(inCmd, interval string, retries uint, timeout, startPeriod string) (*manifest.Schema2HealthConfig, error) {
+	cmdArr := []string{}
+	isArr := false
+	cmdSplit := strings.SplitN(inCmd, " ", 2) // default for compat
```
mheon, line 660: The logic here is reverse of what it should be - do a JSON unmarshal first, and if that returns an error attempt to parse as a string.

cmd/podman/common/specgen.go
```diff
@@ -656,25 +655,52 @@ func FillOutSpecGen(s *specgen.SpecGenerator, c *ContainerCLIOpts, args []string
...
+	if strings.Contains(inCmd, "[") {         // given an array, need to unmarshall it into a pointer
+		isArr = true
+		err := json.Unmarshal([]byte(inCmd), &cmdArr)
+		if err != nil {
+			return nil, err
+		}
+		cmdSplit = cmdArr
+	}
+	concat := ""
+	if cmdSplit[0] != "CMD" && cmdSplit[0] != "none" && cmdSplit[0] != "CMD-SHELL" { // if it isnt a cmd-shell or a cmd, it is a command
```
mheon, line 670: What if the result is len(0)? This will panic

cmd/podman/common/specgen.go
mheon: Sure, but that doesn't address the core of my concern - there's no reason I can see to split the command on whitespace, then recombine it again into a single string for the typical CLI case. It seems like you really have two separate cases - CLI input as a JSON array, and CLI input as a raw string - and both of them can be parsed differently, because I don't believe that something like `--health-cmd '["CMD-SHELL", "echo hello world"]'` is valid.

cdoern: hmmm okay I see what you are saying, this introduces an issue with docker though, since all of their commands have to have either cmd-shell or cmd when passed to this function.

cmd/podman/common/specgen.go
cdoern: an empty string does not panic and shows no healthcheck on inspect since no cmd was specified

TomSweeneyRedHat:
LGTM
and happy green test buttons
@Luap99 can you do the merge honors if your concerns have been addressed?

Issue #10701, opened 2021-06-16
podman exec: periodic 30-second CPU-hogging hangs


edsantiago:
```console
$ ./bin/podman run -d --name foo quay.io/libpod/testimage:20210610 top
0af6b438059a10b47bc78d15226433c3a794cdb32754c608dc1756ad9dd5901d
$ while :;do ./bin/podman exec foo date;done
Wed Jun 16 20:20:57 UTC 2021
Wed Jun 16 20:20:58 UTC 2021
Wed Jun 16 20:20:58 UTC 2021
```
Now just keep watching, it might take a minute or so. When you hear your CPU fan go loud, hit RETURN so you can mark the spot where it hangs. The next entry will be timestamped 30 seconds after the previous. Example:
```console
Wed Jun 16 20:21:13 UTC 2021
Wed Jun 16 20:21:13 UTC 2021

Wed Jun 16 20:21:41 UTC 2021

Wed Jun 16 20:22:11 UTC 2021
Wed Jun 16 20:22:12 UTC 2021
```
master @ 2509a81c343314fa452a0215d05e9d74ab4ec15c, root and rootless.

rhatdan:
One would think we are leaking something.  Perhaps mount points.

Have you tried this as root?  If this works fine as root and not as rootless, that might point to something with fuse-overlay.

edsantiago:
Yep, tried root and rootless (see final line of OP). No leaked mounts that I can see (I even checked while the exec hang is in place; no new mounts)

mheon:
Can't reproduce. Had it running for the last 5 minutes without any effect? F34, latest main branch.

edsantiago:
Oh, weird. It reproduces within seconds for me as root, and ~1 minute rootless. Also f34, 5.11.17-300.fc34. With root I can't even get three iterations of the loop without a hang.

edsantiago:
```console
$ sudo ./bin/podman info
host:
  arch: amd64
  buildahVersion: 1.21.1
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: f.edsantiago.com
  idMappings:
[...]

PR #11049, opened 2021-07-26
support container to container copy

vrothberg:
Implement container to container copy.  Previously data could only be
copied from/to the host.

Fixes: #7370
Co-authored-by: Mehul Arora <aroram18@mcmaster.ca>

### DIFF
--- a/cmd/podman/containers/cp.go
+++ b/cmd/podman/containers/cp.go
@@ -82,7 +82,9 @@ func cp(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if len(sourceContainerStr) > 0 {
+	if len(sourceContainerStr) > 0 && len(destContainerStr) > 0 {
+		return copyContainerToContainer(sourceContainerStr, sourcePath, destContainerStr, destPath)
+	} else if len(sourceContainerStr) > 0 {
 		return copyFromContainer(sourceContainerStr, sourcePath, destPath)
 	}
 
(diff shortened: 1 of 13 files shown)


vrothberg:

cmd/podman/containers/cp.go
```diff
@@ -115,6 +117,110 @@ func doCopy(funcA func() error, funcB func() error) error {
...
+		return err
+	}
+
+	sourceContainerInfo, err := registry.ContainerEngine().ContainerStat(registry.GetContext(), sourceContainer, sourcePath)
+	if err != nil {
+		return errors.Wrapf(err, "%q could not be found on container %s", sourcePath, sourceContainer)
+	}
+
+	var destContainerBaseName string
+	destContainerInfo, destContainerInfoErr := registry.ContainerEngine().ContainerStat(registry.GetContext(), destContainer, destPath)
```
rhatdan, line 135: NIT: Why not just use `err` here?

cmd/podman/containers/cp.go
```diff
@@ -115,6 +117,110 @@ func doCopy(funcA func() error, funcB func() error) error {
...
+		// its base path as it may have changed due to symlink
+		// evaluations.
+		destContainerBaseName = filepath.Base(destContainerInfo.LinkTarget)
+	}
+
+	if sourceContainerInfo.IsDir && !destContainerInfo.IsDir {
+		return errors.New("destination must be a directory when copying a directory")
+	}
+
+	sourceContainerTarget, destContainerTarget := sourceContainerInfo.LinkTarget, destContainerInfo.LinkTarget
```
rhatdan, line 170: This is strange syntax, why not break into two lines?

rhatdan (approved):

cmd/podman/containers/cp.go
rhatdan: I see below you used the err, but it is very hard to follow.  Could you create a boolean variable with a name indicating what you intend.

edsantiago:
It looks like the ten-minute problem is caused by #10701: there are a lot of `podman exec`s in these tests, and each one takes 25-30 seconds.

cmd/podman/containers/cp.go
```diff
@@ -115,6 +117,110 @@ func doCopy(funcA func() error, funcB func() error) error {
...
+		}
+		if err := copyFunc(); err != nil {
+			return errors.Wrap(err, "error copying to container")
+		}
+		return nil
+	}
+
+	return doCopy(sourceContainerCopy, destContainerCopy)
+}
+
```
TomSweeneyRedHat, line 223: Does the file system ensure that we don't write over a container we don't own?

vrothberg: Yes.  We can only use containers in our local storage.

cmd/podman/containers/cp.go
vrothberg: Can I tackle this and the other comment in another PR?  Now it's consistent with the other two copy functions but I can spin up a follow-up PR trying to simplify things a bit more.

vrothberg:
> It looks like the ten-minute problem is caused by #10701: there are a lot of `podman exec`s in these tests, and each one takes 25-30 seconds.

Yes, the tests are noticeably slow.  They used to  _much_ faster.  I have absolutely no idea what may be the cause, so I'll try bisecting.

vrothberg:
> Yes, the tests are noticeably slow. They used to run _much_ faster. I have absolutely no idea what may be the cause, so I'll try bisecting.

I do not notice a difference between Podman v3.0.1-rhel and v3.2.3 on my machine.  Could it be conmon?

vrothberg:
> > Yes, the tests are noticeably slow. They used to run _much_ faster. I have absolutely no idea what may be the cause, so I'll try bisecting.
> 
> I do not notice a difference between Podman v3.0.1-rhel and v3.2.3 on my machine. Could it be conmon?

Scratch that.  Found a reproducer and the difference is very noticeable.

vrothberg:
I stop spamming here now.  Let's continue over in #10701.

I think we should wait for #10701 to be fixed before merging this PR.

vrothberg:
With the commit from https://github.com/containers/podman/pull/11055, the execution time of the cp tests on my machine dropped from ~8 min down to 3min 40sec.

I will still look into reducing the number of tests a bit.

vrothberg:
Rebased on top of the exec fixes.  I also took a shot at reducing the exec's in the cp tests and cut down the execution time by around 25 percent on my machine.

edsantiago:
The exec changes cut test time on my laptop from 3:46 to 3:36 (4% improvement). How did you get 25%? Are you comparing to old-slow-broken-exec time?

vrothberg:
> The exec changes cut test time on my laptop from 3:46 to 3:36 (4% improvement). How did you get 25%? Are you comparing to old-slow-broken-exec time?

Old & slow: > 8min
This PR without cutting down execs: 3min 40sec
This with with cutting down execs: 2min 50sec

edsantiago:
LGTM!

rhatdan:
/lgtm
Thanks @infiniteregrets and @vrothberg

PR #1905, opened 2018-11-30
Pick registry to login from full image name as well

umohnani8:
podman login reg.io/username/image works as well now. It picks
the registry and checks for authentication, if none exist it
will prompt for username and password.

Fixes https://github.com/containers/libpod/issues/1690
Waiting for https://github.com/containers/image/pull/535 to get in. Will fix up the vendor.conf and remove the WIP once that is done.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

Issue #1690, opened 2018-10-22
the issue about podman login


coderwangke:
**Steps to reproduce the issue:**

#### steps
docker : docker login docker.io/wangke007/buildah-test:latest 
podman : podman login docker.io/wangke007/buildah-test:latest 

#### result
docker get Login Succeeded, but podman return a issue:
error authenticating creds for "docker.io/wangke007/buildah-test:latest": pinging docker registry returned: error pinging registry docker.io/wangke007/buildah-test:latest, response code 404 (Not Found)

rhatdan:
What version of podman?
@umohnani8 @mtrmac @TomSweeneyRedHat PTAL

umohnani8:
So docker picks up the registry you are trying to log into from a registry/repo/image path as well. Should be an easy fix, but curious to why would you ever want to do that and not just do podman login docker.io?

mtrmac:
Yeah—this looks potentially misleading, suggesting that `podman login docker.io/ns1` and `podman login docker.io/ns2` maintained two independent login records, which does not happen.

rhatdan:
Is this just docker login always attempting to login as docker.io?  If you did 

docker login quay.io/wangke007/buildah-test:latest

Does it do the right thing?

umohnani8:
@rhatdan yeah it asks for credential for quay.io then. Also doing it with the full path, it doesn't care about what user you put in the path, it will still ask for username and password again. Like I can do `docker login docker.io/umohnani8/blah` and then login as `umohnani` instead.

umohnani8:
@mtrmac @rhatdan  so what do we want do about this?

rhatdan:
I think we should match docker's behaviour and grab the registry from the image name.

PR #11054, opened 2021-07-27
Add `--accept-repositories` integration tests

saschagrunert:
This adds the integration tests for the repository or namespaced registry feature introduced in c/common.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/buildah v1.21.1-0.20210721171232-54cafea4c933
-	github.com/containers/common v0.41.1-0.20210721172332-291287e9d060
+	github.com/containers/common v0.41.1-0.20210730122913-cd6c45fd20e3
 	github.com/containers/conmon v2.0.20+incompatible
 	github.com/containers/image/v5 v5.14.0
 	github.com/containers/ocicrypt v1.1.2
--- a/cmd/podman/login.go
+++ b/cmd/podman/login.go
@@ -52,6 +52,7 @@ func init() {
 	loginOptions.Stdin = os.Stdin
 	loginOptions.Stdout = os.Stdout
 	loginOptions.AcceptUnspecifiedRegistry = true
+	loginOptions.AcceptRepositories = true
 }
 
 // Implementation of podman-login.
--- a/cmd/podman/logout.go
+++ b/cmd/podman/logout.go
@@ -43,6 +43,7 @@ func init() {
 
 	logoutOptions.Stdout = os.Stdout
 	logoutOptions.AcceptUnspecifiedRegistry = true
+	logoutOptions.AcceptRepositories = true
 }
 
 // Implementation of podman-logout.
--- a/go.sum
+++ b/go.sum
@@ -242,8 +242,8 @@ github.com/containernetworking/plugins v0.9.1/go.mod h1:xP/idU2ldlzN6m4p5LmGiwRD
 github.com/containers/buildah v1.21.1-0.20210721171232-54cafea4c933 h1:jqO3hDypBoKM5be+fVcqGHOpX2fOiQy2DFEeb/VKpsk=
 github.com/containers/buildah v1.21.1-0.20210721171232-54cafea4c933/go.mod h1:9gspFNeUJxIK72n1IMIKIHmtcePEZQsv0tjo+1LqkCo=
 github.com/containers/common v0.41.1-0.20210721112610-c95d2f794edf/go.mod h1:Ba5YVNCnyX6xDtg1JqEHa2EMVMW5UbHmIyEqsEwpeGE=
-github.com/containers/common v0.41.1-0.20210721172332-291287e9d060 h1:HgGff2MeEKfYoKp2WQFl9xdsgP7KV8rr/1JZRIuPXmg=
-github.com/containers/common v0.41.1-0.20210721172332-291287e9d060/go.mod h1:Ba5YVNCnyX6xDtg1JqEHa2EMVMW5UbHmIyEqsEwpeGE=
+github.com/containers/common v0.41.1-0.20210730122913-cd6c45fd20e3 h1:lHOZ+G5B7aP2YPsbDo4DtALFAuFG5PWH3Pv5zL2bC08=
+github.com/containers/common v0.41.1-0.20210730122913-cd6c45fd20e3/go.mod h1:UzAAjDsxwd4qkN1mgsk6aspduBY5bspxvKgwQElaBwk=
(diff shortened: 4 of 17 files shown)


saschagrunert:
Test are running towards green. Do we need a c/common release or are we fine vendoring the latest master here @vrothberg?

vrothberg:
> Test are running towards green. Do we need a c/common release or are we fine vendoring the latest master here @vrothberg?

I am cool with vendoring commits during development.  We just need to make sure to vendor release before releasing Podman.

docs/source/markdown/podman-logout.1.md
```diff
@@ -21,6 +21,10 @@ All the cached credentials can be removed by setting the **all** flag.
 
 ## OPTIONS
 
+#### **--accept-repositories**
+
+Allow namespaces or repositories rather than just registries.
```
vrothberg, line 26: Can you give an example directly here what that means?  I think that many users will have troubles imagining the semantics.

docs/source/markdown/podman-login.1.md
```diff
@@ -28,6 +28,10 @@ For more details about format and configurations of the auth.json file, please r
 
 ## OPTIONS
 
+#### **--accept-repositories**
+
+Allow namespaces or repositories rather than just registries.
```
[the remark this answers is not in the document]
saschagrunert: Yep, updated the docs to provide an example.

docs/source/markdown/podman-logout.1.md
saschagrunert: Yep, updated the docs to provide an example.

saschagrunert:

docs/source/markdown/podman-login.1.md
```diff
@@ -28,6 +28,12 @@ For more details about format and configurations of the auth.json file, please r
 
 ## OPTIONS
 
+#### **--accept-repositories**
```
rhatdan, line 31: I still question why is this an option?  Why would I care? I don't like to add options that know one is ever going to choose?  Should this option default to true?  If so who would ever set it to false?

TomSweeneyRedHat: I think my question may mirror@rhatdan's.  Do we need this option at all?  If a user specifies `podman login quay.io/username/image` could we just log into that rather than `quay.io.`?

saschagrunert: Follow-up of https://github.com/containers/common/pull/659#discussion_r673842324, https://github.com/containers/common/pull/659#issuecomment-872999764

@vrothberg @mtrmac would it make sense to automatically switch on the flag if a repository is provided?

vrothberg: In theory it is a breaking change if we did.  If @rhatdan wants to take the risk, I am OK with it.

rhatdan:
```
 podman login quay.io/rhatdan/podman
Username: rhatdan
Password: 
Login Succeeded!
```
Currently works.  But this is undefined behavior, I believe.

With the new change will push quay.io/rhatdan/buildah fail?

saschagrunert: Right now it will only write `quay.io` into auth.json, where with the new behavior (and the flag enabled) it will write `quay.io/rhatdan/podman`.

rhatdan: Meaning the podman push of quay.io/rhatdan/buildah will fail because the repo is not in the quay.io/rhatdan/podman.

Currently the login to quay.io/rhatdan/podman writes to quay.io/rhatdan, correct?

saschagrunert: > Meaning the podman push of quay.io/rhatdan/buildah will fail because the repo is not in the quay.io/rhatdan/podman.

No, it will fallback to the credentials previously written, aka `quay.io` .

Ref: https://github.com/containers/image/blob/9e1d7b48f1de2ea1ffe61a74916f6b0bcd605abb/pkg/docker/config/config.go#L677-L697

> Currently the login to quay.io/rhatdan/podman writes to quay.io/rhatdan, correct?

Currently, a login to `quay.io/rhatdan/podman` writes only `quay.io`.

After this merge, everything works as before without breaking existing behavior if the flag is not being used.

mtrmac: (Stripping repo from the argument)
> But this is undefined behavior, I believe.

Well, it’s an explicitly made change , in https://github.com/containers/podman/pull/1905 , for https://github.com/containers/podman/issues/1690 .

---

We have added the opt-in flag exactly because this is a breaking change. It would make no sense to _both_ have the flag and have it default to on; the impacted users would have to add the flag to revert to the previous behavior, so they might just as well change the non-option argument.

(It might make sense to force the flag to on, without having it in the CLI at all. That would be equally a breaking change (and something c/common does not currently allow, I think), but at least the users wouldn’t be bothered with the existence of the flag.

Overall I don’t think the breaking change is the best option — Podman does have user reports of users using the repo syntax, and expecting it to work in a way compatible with Docker. So the opt-in not to break users makes more sense to me. But it’s a judgement call; I could see the argument that many users passing a repo without the opt-in flag are likely to be making a mistake with the old semantics.)

saschagrunert:
@rhatdan PTAL at the latest discussion points

rhatdan:
I say no option and opt in to new change.  IE Breaking change,  If we get complaints, then we can add the flag and the ability to manage this in containers/common.

mtrmac:
> I say no option and opt in to new change.

OK, that requires a new parameter to `c/common/pkg/auth.Get{Login,Logout}Flags`. (Or hard-coding the behavior for all callers and dropping the support for the flag entirely? Either way, a c/common change.)

saschagrunert:
Change in c/common: https://github.com/containers/common/pull/700

go.mod
```diff
@@ -69,3 +69,5 @@ require (
 	k8s.io/api v0.21.3
 	k8s.io/apimachinery v0.21.3
 )
+
+replace github.com/containers/common => github.com/saschagrunert/common v0.40.2-0.20210730110303-fc86f687be82
```
saschagrunert, line 73: TODO: remove and vendor latest c/common master if https://github.com/containers/common/pull/700 got merged

saschagrunert:
/hold cancel
This is ready for review and possibly merge

TomSweeneyRedHat:
LGTM

mtrmac:
Any opinions on whether Buildah and Skopeo should opt into the new behavior as well? (For the upcoming releases?)

[Cc: @TomSweeneyRedHat for possible timing interaction, especially vs. a Buildah release before vendoring into Podman — I don’t think it would be a problem to deliver this CLI change in Podman only to minimize last-minute rush, but I might be mistaken about that.]

mtrmac (commented):
Code LGTM, but I’d prefer someone more knowledgeable to check the timing vs. vendor dance and releases.

rhatdan:
Yes Buildah and Skopeo should also work the same way.
/lgtm

saschagrunert:
Changed the PR to reflect the actual change.

Issue #10946, opened 2021-07-15
iscsi and volume


snail59:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
I have an LVM logical volume on top of an iscsi LUN, mounted on a directory ( let's call it /foo/bar ) :
iscsi -> LVM -> logical volume -> /foo/bar

I ran something like podman run -v /foo/bar:/my_container_directory my_image .
Only /foo was mounted on /my_container_directory  in the container, not /foo/bar.

**Steps to reproduce the issue:**

1. add an iscsi target to your machine

2. create an lvm volume or a filesystem on this volume

3. try to bind it to a container

**Describe the results you received:**
The mount command run in the container shows that /foo is mounted on /my_container_directory in the container

**Describe the results you expected:**
I expected /foo/bar to be mounted on /my_container_directory in the container, instead of /foo.

**Additional information you deem important (e.g. issue happens only occasionally):**
I tried this with docker and it worked fine.
**Output of `podman version`:**

```
Version:      3.2.1
API Version:  3.2.1
Go Version:   go1.16.5
OS/Arch:      linux/amd64

```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
*  app-emulation/podman
      Latest version available: 3.2.1
      Latest version installed: 3.2.1
      Size of files: 10 503 KiB
      Homepage:      https://github.com/containers/podman/
      Description:   Library and podman tool for running OCI-based containers in Pods
      License:       Apache-2.0 BSD BSD-2 CC-BY-SA-4.0 ISC MIT MPL-2.0
This is on gentoo
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide? (https://github.com/containers/podman/blob/master/troubleshooting.md)**

Yes/No/Maybe I don't know

**Additional environment details (AWS, VirtualBox, physical, etc.):**

Physical

Luap99:
Can you try with `-v /foo/bar:/my_container_directory:rbind`.

> If you want to recursively mount a volume and all of its submounts into a container, then you can use the rbind option. By default the bind option is used, and submounts of the source directory will not be mounted into the container.

From https://github.com/containers/podman/blob/main/docs/source/markdown/podman-create.1.md#--volume--vsource-volumehost-dircontainer-diroptions

snail59:
I tried, it does not work either.
Interesting facts are that: 
[...]

PR #11056, opened 2021-07-27
rootless: check that / is mounted as shared

giuseppe:
if the root mount '/' is not mounted as MS_SHARED, print a
warning, otherwise new mounts that are created in the host won't be
propagated to the rootless mount namespace.

Closes: https://github.com/containers/podman/issues/10946

[NO TESTS NEEDED]

### DIFF
--- a/pkg/rootless/rootless_linux.go
+++ b/pkg/rootless/rootless_linux.go
@@ -14,11 +14,13 @@ import (
 	"os/user"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
 	"unsafe"
 
 	"github.com/containers/podman/v3/pkg/errorhandling"
 	"github.com/containers/storage/pkg/idtools"
+	pmount "github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/unshare"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM

pkg/rootless/rootless_linux.go
```diff
@@ -235,6 +237,24 @@ func becomeRootInUserNS(pausePid, fileToRead string, fileOutput *os.File) (_ boo
...
+			isShared := false
+			for _, o := range strings.Split(m.Optional, ",") {
+				if strings.HasPrefix(o, "shared:") {
+					isShared = true
+					break
+				}
+			}
+			if !isShared {
+				logrus.Warningf("%q is not a shared mount, this could cause issues or missing mounts with rootless containers", m.Mountpoint)
+			}
```
Luap99, line 254: Should there be a break here? There is no point in looping over the remaining mount points.

giuseppe: thanks for spotting it.  Fixed and pushed a new version

rhatdan:
Is this going to warn all the time?  If I have a system that is built with no / as private, will podman constantly complain?

giuseppe:
> Is this going to warn all the time? If I have a system that is built with no / as private, will podman constantly complain?

just the first time when we create the user+mount namespace.

rhatdan:
so the first time after use login, everytime.

rhatdan:
/lgtm
Might need an option eventually to turn this off,  we shall see.

PR #11061, opened 2021-07-27
command: migrate doesn't move process to cgroup

giuseppe:
add a new annotation for the "system migrate" command to not move the pause process to a separate cgroup.
    
The operation is not needed since "system migrate" destroys the pause process, so there won't be any process left to move to a cgroup.

[NO TESTS NEEDED]


vrothberg (commented):
LGTM

rhatdan:
Did this fix an issue?

/lgtm

PR #11067, opened 2021-07-28
remote build: fix streaming and error handling

vrothberg:
Address a number of issues in the streaming logic in remote build, most
importantly an error in using buffered channels on the server side.

The pattern below does not guarantee that the channel is entirely read
before the context fires.
```
for {
	select {
		case <- bufferedChannel:
		...
		case <- ctx.Done():
		...
	}
}
```
Fixes: #10154

... had to create a new PR. GitHub did not like to reopen after I pushed to the branch.

### DIFF
--- a/test/system/070-build.bats
+++ b/test/system/070-build.bats
@@ -749,16 +749,9 @@ RUN echo $random_string
 EOF
 
     run_podman 125 build -t build_test --pull-never $tmpdir
-    # FIXME: this is just ridiculous. Even after #10030 and #10034, Ubuntu
-    # remote *STILL* flakes this test! It fails with the correct exit status,
-    # but the error output is 'Error: stream dropped, unexpected failure'
-    # Let's just stop checking on podman-remote. As long as it exits 125,
-    # we're happy.
-    if ! is_remote; then
-        is "$output" \
-           ".*Error: error creating build container: quay.io/libpod/nosuchimage:nosuchtag: image not known" \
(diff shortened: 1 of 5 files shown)


vrothberg:
@edsantiago @rhatdan @jwhonce PTAL

test/system/070-build.bats
```diff
@@ -749,16 +749,9 @@ RUN echo $random_string
...
-    # FIXME: this is just ridiculous. Even after #10030 and #10034, Ubuntu
-    # remote *STILL* flakes this test! It fails with the correct exit status,
-    # but the error output is 'Error: stream dropped, unexpected failure'
-    # Let's just stop checking on podman-remote. As long as it exits 125,
-    # we're happy.
-    if ! is_remote; then
-        is "$output" \
-           ".*Error: error creating build container: quay.io/libpod/nosuchimage:nosuchtag: image not known" \
-           "--pull-never fails with expected error message"
-    fi
```
edsantiago, line 761: oh, nice catch. I had forgotten all about this skip.

edsantiago (approved):
With this PR, #9887 passes.

To understand what that means, go look at that PR. I'll wait. Scroll up to April 27, when I filed #10154. Look at all the force-pushes I've done since then. All of those failed with multiple flakes. This is the first flake-free run in exactly three months. So, in short, LGTM. Nice work, @vrothberg, and thank you!

PR #11072, opened 2021-07-28
Remove ReadHeaderTimeout

matejvasek:
Effectively sets timeout to infinity.
This is needed in order to make `podman` work with `pack`.
The `pack` CLI is keeping one connection for prolonged time.
Closing the connection breaks `pack`'s functionality.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -90,11 +90,10 @@ func newServer(runtime *libpod.Runtime, duration time.Duration, listener *net.Li
 
 	server := APIServer{
 		Server: http.Server{
-			Handler:           router,
-			ReadHeaderTimeout: 20 * time.Second,
-			IdleTimeout:       duration * 2,
-			ConnState:         idle.ConnState,
-			ErrorLog:          log.New(logrus.StandardLogger().Out, "", 0),
+			Handler:     router,
+			IdleTimeout: duration * 2,
+			ConnState:   idle.ConnState,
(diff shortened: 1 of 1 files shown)


matejvasek:
There is a tool for which the timeout posses a problem.

matejvasek:
The tool is `pack` for building buildpacks.
I am already tying to address it here https://github.com/buildpacks/lifecycle/pull/684.

vrothberg (commented):
Change LGTM.  In case of a repush, please update the commit message with an explanation why the change is necessary.

rhatdan:
Please repush with and explanation of why you want this change,  We need @jwhonce to describe why he added it in the first place.  I guess this would allow someone to connect to the service and just hang it?  Is my assumption correct?

matejvasek:
@jwhonce PTAL

vrothberg (commented):
/lgtm
/hold

vrothberg:
@mheon @jwhonce I would love to get this into 3.3.

jwhonce:
@rhatdan Correct, without the timeout, we as open to a DOS. The attacker would just open n sockets and never close them until the service hits the configured rlimit.

matejvasek:
@jwhonce but even with 20 second limit the attacker still could keep opening a lot of connection right? also how often are user exposing the socket to outer internet?

matejvasek:
Am in for reverting this as soon as this is resolved https://github.com/buildpacks/lifecycle/issues/686

Issue #10575, opened 2021-06-06
The command Podman system service creates 2 processes that are hard to die.


PavelSosin-320:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Every execution of the command podman system service by a rootless user with default parameters results in the creation of 2 processes owned by the user but only one Linux Domain socket. The 1st process is zombi.
**Steps to reproduce the issue:**

1. Login as a rootless user
2. Execute podman system service -t 0 --> no output???? No pid file??????
2.  Execute ps and get 2 "podman" processes
3. Check Socket existence ls $XDG_RUNTIME_DIR/podman/ podman.sock ---> exists!
3. Kill the 1st process - nothing happens
4. Try another way:
 podman system service -t 0 &
[1] 16712
[pavel@MSI-wsl user]$ ps
    PID TTY          TIME CMD
  14055 pts/6    00:00:00 bash
  16712 pts/6    00:00:00 podman
  16722 pts/6    00:00:00 podman

5. kill 16722; The output is [1]+  Exit 1                  podman system service -t 0

**Describe the results you received:**
2 processes are created, one of them is zombi.

**Describe the results you expected:**
Only one "Leading" process is created and its PID is placed into XDG_RUNTIME_DIR/podman/podman.pid
To be used later for kill $(<%XDG_RUNTIE_DIR/podman/podman.pid
**Additional information you deem important (e.g. issue happens only occasionally):**
Pevent management podman service using systend,
**Output of `podman version`:**
podman version
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.15.11
Built:        Tue May 11 16:53:47 2021
OS/Arch:      linux/amd64

```
(paste your output here)
```

**Output of `podman info --debug`:**

```
(paste your output here)
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
(paste your output here)
```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide? (https://github.com/containers/podman/blob/master/troubleshooting.md)**

Yes/No

**Additional environment details (AWS, VirtualBox, physical, etc.):**
![image](https://user-images.githubusercontent.com/54495912/120926290-f2021d00-c6e4-11eb-9fcd-ed9e528048ec.png)

mheon:
Replicates on Linux. It's not the pause process - I see that too, but there are also two `podman system service` processes and a zombie from the `podman system service` invocation.

[...]

PR #11073, opened 2021-07-29
rootless: avoid zombie process on first launch

giuseppe:
avoid a zombie process if on the first launch Podman creates a long
living process, such as "podman system service -t 0".

The `r` variable was overriden thus causing the waitpid to fail and
not clean up the intermediate process.

Closes: https://github.com/containers/podman/issues/10575

[NO TESTS NEEDED]

### DIFF
--- a/pkg/rootless/rootless_linux.c
+++ b/pkg/rootless/rootless_linux.c
@@ -465,38 +465,43 @@ reexec_in_user_namespace_wait (int pid, int options)
 static int
 create_pause_process (const char *pause_pid_file_path, char **argv)
 {
-  int r, p[2];
+  pid_t pid;
+  int p[2];
 
   if (pipe (p) < 0)
-    _exit (EXIT_FAILURE);
+    return -1;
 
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM, thank you @giuseppe !

vrothberg:

rhatdan:
/lgtm
/hold
Don't know what is going on with the CI system.

PavelSosin-320:
@giuseppe Hopefully, PR 10575 will calm the following issue:
 
_systemd[133]: Starting PodmanAPI Service for Rootless users...
podman[279]: 2021-08-03 08:19:42.6111998 +0300 IDT m=+0.180908901 system refresh
systemd[133]: podman.service: Got notification message from PID 279, but reception only permitt>_
that appears when rootless user starts podman using systemd unit as "Notify" type service:
[Service]
Type=notify
...
Indeed, the main process is Main PID: 269 (podman), So systemd doesn't expect any other notifications.

Issue #10361, opened 2021-05-17
support the `container:...` notation for `ps --filter network=...`


dsedivec:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I can't find any reasonable way to ask for all containers that are using another container's network.

**Steps to reproduce the issue:**

1. `podman run --name=alpha --net=testnet -d docker.io/alpine sleep 1h`

2. `podman run --name=bravo --net=container:alpha -d docker.io/alpine sleep 1h`

3. `podman container ps --filter network=container:alpha`

**Describe the results you received:**

No containers listed

**Describe the results you expected:**

Container `bravo` listed (and maybe `alpha` too, I'm not that picky)

**Additional information you deem important (e.g. issue happens only occasionally):**

`podman container ps --filter network=testnet` only lists `alpha`, not `bravo`.  `podman container ps --filter network=container:$(podman inspect --format {{.ID}} alpha)` doesn't list anything.

`podman container ps --format '{{.Names}} {{.Networks}}'` tells me that it thinks `bravo` is in the `podman` network, which seems wrong.

**Output of `podman version`:**

```
podman version
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16
Built:        Thu Apr 22 06:11:28 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.20.1
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.27-2.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.27, commit: '
  cpus: 8
  distribution:
    distribution: fedora
    version: "34"
  eventLogger: journald
  hostname: mmhmm
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.11.13-300.fc34.x86_64
  linkmode: dynamic
  memFree: 782872576
  memTotal: 16676921344
  ociRuntime:
    name: runc
    package: runc-1.0.0-377.rc93.fc34.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc93
      commit: 48389904331fdebba5e2db73147c067560a51d80
      spec: 1.0.2-dev
      go: go1.16
      libseccomp: 2.5.0
  os: linux
  remoteSocket:
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #11075, opened 2021-07-29
ps: support the `container...` notation for `ps --filter network=...`

flouthoc:
Hi team,

Following PR extends `ps --filter` to support docker like:
 `podman container ps --filter network=container:<CtrIdorCtrName>`.

Closes: https://github.com/containers/podman/issues/10361

### DIFF
--- a/libpod/container.go
+++ b/libpod/container.go
@@ -1173,6 +1173,46 @@ func (c *Container) Networks() ([]string, bool, error) {
 	return c.networks()
 }
 
+// NetworkMode gets the configured network mode for the container.
+// Get actual value from the database
+func (c *Container) NetworkMode() string {
+	networkMode := ""
+	ctrSpec := c.config.Spec
+
+	switch {
+	case c.config.CreateNetNS:
+		// We actually store the network
+		// mode for Slirp and Bridge, so
+		// we can just use that
+		networkMode = string(c.config.NetMode)
+	case c.config.NetNsCtr != "":
+		networkMode = fmt.Sprintf("container:%s", c.config.NetNsCtr)
+	default:
+		// Find the spec's network namespace.
+		// If there is none, it's host networking.
+		// If there is one and it has a path, it's "ns:".
+		foundNetNS := false
+		for _, ns := range ctrSpec.Linux.Namespaces {
+			if ns.Type == spec.NetworkNamespace {
+				foundNetNS = true
+				if ns.Path != "" {
+					networkMode = fmt.Sprintf("ns:%s", ns.Path)
+				} else {
+					// We're making a network ns,  but not
+					// configuring with Slirp or CNI. That
+					// means it's --net=none
+					networkMode = "none"
+				}
+				break
+			}
+		}
+		if !foundNetNS {
+			networkMode = "host"
+		}
+	}
+	return networkMode
+}
+
 // Unlocked accessor for networks
 func (c *Container) networks() ([]string, bool, error) {
 	networks, err := c.runtime.state.GetNetworks(c)
--- a/pkg/domain/filters/containers.go
+++ b/pkg/domain/filters/containers.go
@@ -211,6 +211,36 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
 		}, nil
 	case "network":
 		return func(c *libpod.Container) bool {
+			networkMode := c.NetworkMode()
+			// support docker like `--filter network=container:<IDorName>`
+			// check if networkMode is configured as `container:<ctr>`
+			// peform a match against filter `container:<IDorName>`
+			// networks is already going to be empty if `container:<ctr>` is configured as Mode
+			if strings.HasPrefix(networkMode, "container:") {
+				networkModeContainerPart := strings.SplitN(networkMode, ":", 2)
+				if len(networkModeContainerPart) < 2 {
+					return false
+				}
+				networkModeContainerID := networkModeContainerPart[1]
+				for _, val := range filterValues {
+					if strings.HasPrefix(val, "container:") {
+						filterNetworkModePart := strings.SplitN(val, ":", 2)
+						if len(filterNetworkModePart) < 2 {
+							return false
+						}
+						filterNetworkModeIDorName := filterNetworkModePart[1]
+						filterID, err := r.LookupContainerID(filterNetworkModeIDorName)
+						if err != nil {
+							return false
+						}
+						if filterID == networkModeContainerID {
+							return true
+						}
+					}
+				}
+				return false
+			}
+
 			networks, _, err := c.Networks()
 			// if err or no networks, quick out
 			if err != nil || len(networks) == 0 {
--- a/test/e2e/ps_test.go
+++ b/test/e2e/ps_test.go
@@ -172,6 +172,43 @@ var _ = Describe("Podman ps", func() {
 		Expect(fullCid).To(Equal(result.OutputToStringArray()[0]))
 	})
 
+	It("podman ps --filter network=container:<name>", func() {
+		ctrAlpha := "alpha"
+		container := podmanTest.Podman([]string{"run", "-dt", "--name", ctrAlpha, ALPINE, "top"})
+		container.WaitWithDefaultTimeout()
+		Expect(container).Should(Exit(0))
+
+		ctrBravo := "bravo"
+		containerBravo := podmanTest.Podman([]string{"run", "-dt", "--network", "container:alpha", "--name", ctrBravo, ALPINE, "top"})
+		containerBravo.WaitWithDefaultTimeout()
+		Expect(containerBravo).Should(Exit(0))
+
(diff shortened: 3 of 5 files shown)


rhatdan:
@mheon @Luap99 PTAL

libpod/container.go
```diff
@@ -1173,6 +1173,24 @@ func (c *Container) Networks() ([]string, bool, error) {
 	return c.networks()
 }
 
+// NetworkMode gets the configured network mode for the container.
+// Get actual value from the database
+func (c *Container) NetworkMode() string {
```
Luap99, line 1178: I think you copied the logic from inspect. It would make sense to use the complete inspect logic here and use this function in inspect to prevent code duplication.

pkg/domain/filters/containers.go
```diff
@@ -212,6 +212,24 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
 	case "network":
 		return func(c *libpod.Container) bool {
 			networks, _, err := c.Networks()
```
Luap99, line 214: Can you move this down to line 233.

pkg/domain/filters/containers.go
```diff
@@ -212,6 +212,24 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
...
+			// support docker like `--filter network=container:<IDorName>`
+			// check if networkMode is configured as `container:<ctr>`
+			// peform a match against filter `container:<IDorName>`
+			// networks is already going to be empty if `container:<ctr>` is configured as Mode
+			if strings.Contains(networkMode, "container:") {
+				networkModeContainerID := strings.SplitN(networkMode, ":", 2)[1]
+				for _, val := range filterValues {
+					if strings.Contains(val, "container:") {
+						filterNetworkModeIDorName := strings.SplitN(val, ":", 2)[1]
+						filterID, _ := r.LookupContainerID(filterNetworkModeIDorName)
```
Luap99, line 226: Please handle the error.

pkg/domain/filters/containers.go
```diff
@@ -212,6 +212,24 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
...
+						if filterID == networkModeContainerID {
+							return true
+						}
+					}
+				}
```
Luap99, line 231: You can return false here, if the network mode is `container:` we do not have to check network names.

test/e2e/ps_test.go
```diff
@@ -172,6 +172,24 @@ var _ = Describe("Podman ps", func() {
...
+		container := podmanTest.Podman([]string{"run", "-dt", "--name", ctrAlpha, ALPINE, "top"})
+		container.WaitWithDefaultTimeout()
+		Expect(container).Should(Exit(0))
+
+		ctrBravo := "bravo"
+		containerBravo := podmanTest.Podman([]string{"run", "-dt", "--network", "container:alpha", "--name", ctrBravo, ALPINE, "top"})
+		containerBravo.WaitWithDefaultTimeout()
+		Expect(containerBravo).Should(Exit(0))
+
+		result := podmanTest.Podman([]string{"ps", "-a", "--format", "table {{.Names}}", "--filter", "network=container:alpha"})
```
Luap99, line 186: Can you add another test with `container:<ID>` instead of name.

libpod/container.go
flouthoc: @Luap99 thanks this is resolved in latest commit.

pkg/domain/filters/containers.go
flouthoc: Resolved in latest commit.

test/e2e/ps_test.go
flouthoc: Resolved in latest commit.

pkg/domain/filters/containers.go
flouthoc: resolved :)

pkg/domain/filters/containers.go
flouthoc: resolved in latest commit.

Luap99 (approved):
LGTM

libpod/container.go
```diff
@@ -1173,6 +1173,54 @@ func (c *Container) Networks() ([]string, bool, error) {
 	return c.networks()
 }
 
+// NetworkMode gets the configured network mode for the container.
+// Get actual value from the database
+func (c *Container) NetworkMode(ctrSpec *spec.Spec) (string, error) {
```
mheon, line 1178: Why does this accept a spec? The container struct already contains a spec, and nothing you're accessing here requires something that would be overwritten from the initial user-provided spec. Just using `c.config.Spec` is simpler and safer.

pkg/domain/filters/containers.go
```diff
@@ -211,6 +211,31 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
...
 		return func(c *libpod.Container) bool {
+			networkMode, err := c.NetworkMode(nil)
+			if err != nil {
+				return false
+			}
+			// support docker like `--filter network=container:<IDorName>`
+			// check if networkMode is configured as `container:<ctr>`
+			// peform a match against filter `container:<IDorName>`
+			// networks is already going to be empty if `container:<ctr>` is configured as Mode
+			if strings.Contains(networkMode, "container:") {
```
mheon, line 222: Why `strings.Contains()` instead of `strings.HasPrefix()`?

pkg/domain/filters/containers.go
```diff
@@ -211,6 +211,31 @@ func GenerateContainerFilterFuncs(filter string, filterValues []string, r *libpo
...
+				networkModeContainerID := strings.SplitN(networkMode, ":", 2)[1]
```
mheon, line 223: Probably safer to do a length check here to ensure we don't panic.

pkg/domain/filters/containers.go
mheon: As I look at this more - it really looks as though the check for `container:` needs to be integrated into the `filterValues` loop below, instead of done separately up here.

libpod/container.go
flouthoc: @mheon Following function is also going to be used by `generateInspectContainerHostConfig` in inspect code chunk. I think for both `inspect` and `ps --filter` their exists a state where config does not exist **(e.g., because the container was never started)** for such cases I think we can read data from db.

However for inspect block `ctrSpec` is already fetched from state so i thought would be more optimal to reuse the `ctrSpec` as argument inoder to reduce lookup cost.

pkg/domain/filters/containers.go
flouthoc: This is resolved in latest commit.

pkg/domain/filters/containers.go
flouthoc: This is resolved in latest commit.

flouthoc:
@mheon I have resolved all your feedback comments in latest commit. I have a doubt on first feedback , could you please take a look and correct me if i am wrong somewhere.

flouthoc:
Re-basing with main to add fix for https://github.com/containers/podman/pull/11077

libpod/container.go
mheon: `c.config.Spec` is always valid; there is never a point at which it is not populated. It is not the final spec that is used for the container (we use it to generate the final spec), but it is very close - the changes made to produce a final spec are minimal. As such it is safe to use here - no part of what you are accessing will change during final spec generation.

TomSweeneyRedHat:
LGTM
and happy green test buttons
I'll let @mheon press the final merge button as I'm not certain all of his concerns were addressed.

flouthoc:
@TomSweeneyRedHat thanks :) there is one change left which @mheon suggested. @mheon made relevant changes will only try to read spec from db if `c.config.Spec` is nil for added safety does it looks fine to you ?

libpod/container.go
```diff
@@ -1173,6 +1173,58 @@ func (c *Container) Networks() ([]string, bool, error) {
...
 
+// NetworkMode gets the configured network mode for the container.
+// Get actual value from the database
+func (c *Container) NetworkMode() (string, error) {
+	var err error
+	networkMode := ""
+	ctrSpec := c.config.Spec
+
+	// fetch spec from existing state if nil provided.
+	if ctrSpec == nil {
```
mheon, line 1184: This will never be true, you can drop the check - we will never create a container with a nil state.

mheon: Further - we create the spec that `specFromState()` would retrieve *from* `c.config.Spec` so that will also fail if `c.config.Spec` does not exist

flouthoc: @mheon thanks, I have removed the check in latest commit.

flouthoc:
Just a sec cleaning error handling for removed block.

flouthoc:
@mheon This is resolved could you please review again.

PR #11080, opened 2021-07-29
system tests: fix race in stop test

edsantiago:
In the unlock/timeout test, on slow systems, 'podman ps' could
catch the container before the just-backgrounded 'podman stop'
sends the signal. Wait for signal ack from container before
we inspect it.

Also: If I understand the test correctly, it wasn't actually
checking that 'ps' could grab the lock while the container
was exiting. Add a check.

### DIFF
--- a/test/system/050-stop.bats
+++ b/test/system/050-stop.bats
@@ -119,10 +119,30 @@ load helpers
     # the container's status.
 
     run_podman run --name stopme -d $IMAGE sh -c \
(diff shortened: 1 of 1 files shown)


edsantiago:
For reference: race condition caused flake in [multiarch testing](https://gitlab.com/cki-project/kernel-tests/-/issues/628#note_638860131)

vrothberg (commented):
LGTM

PR #11086, opened 2021-07-30
Fix auto-update system test for older systemd

hshiina:
If the systemd version is older than this [PR](https://github.com/systemd/systemd/pull/14851), the systemd uses
'Started' when a oneshot service finishes.

Signed-off-by: Hironori Shiina <shiina.hironori@jp.fujitsu.com>

### DIFF
--- a/test/system/255-auto-update.bats
+++ b/test/system/255-auto-update.bats
@@ -261,7 +261,8 @@ EOF
     systemctl enable --now podman-auto-update-$cname.timer
     systemctl list-timers --all
 
-    local expect='Finished Podman auto-update testing service'
+    # While systemd v245 and later uses 'Finished', older versions uses 'Started' for oneshot services
+    local expect='(Finished|Started) Podman auto-update testing service'
     local failed_start=failed
     local count=0
     while [ $count -lt 120 ]; do


mheon:
LGTM
@edsantiago PTAL

test/system/255-auto-update.bats
```diff
@@ -261,7 +261,8 @@ EOF
     systemctl enable --now podman-auto-update-$cname.timer
     systemctl list-timers --all
 
-    local expect='Finished Podman auto-update testing service'
+    # While newer systemd uses 'Finished', older systemd uses 'Started' for oneshot services
```
edsantiago, line 264: A good habit to develop: never use "new", "recent", "old" in a code comment. A future maintainer, either tomorrow or in five years, will have no possible way of knowing what that means. Could you track down a systemd version tag and use that instead?

edsantiago (commented):
LGTM - thank you for identifying that! One suggestion for the benefit of future maintainers.

TomSweeneyRedHat:
I'd like to get @edsantiago 's nit addressed before merging, otherwise LGTM

hshiina:
Thank you for the review.

This message was changed at systemd v245: https://github.com/systemd/systemd/commit/eda0cbf07186d16a160bd1d810613586fdbdf587

PR #11094, opened 2021-08-02
Bump to v4.0.0-dev

mheon:
Bump the main branch's `-dev` version to differentiate between it and the v3.3 branch.


mheon:

vrothberg:
Are we certain that the next version will be 4.0.0?  I thought that would be early next year.

mheon:
@vrothberg If we have to branch again before 4.0, we can change the development version of the branch to 3.5.0-dev.

We should discuss whether we want to allow breaking changes in main at this point, though. We could hold off for another few months, or we could make sure they are clearly marked and revert if and when we need a new branch.

PR #11096, opened 2021-08-02
Use static path for gvproxy

baude:
Given that we do not want to support gvproxy for anything other than
podman machine, we have decided to use a static path of
/usr/libexec/podman/gvproxy instead of a lookpath.

[NO TESTS NEEDED]

### DIFF
--- a/pkg/machine/qemu/machine.go
+++ b/pkg/machine/qemu/machine.go
@@ -603,9 +603,9 @@ func CheckActiveVM() (bool, string, error) {
 // startHostNetworking runs a binary on the host system that allows users
 // to setup port forwarding to the podman virtual machine
 func (v *MachineVM) startHostNetworking() error {
-	binary, err := exec.LookPath(machine.ForwarderBinaryName)
-	if err != nil {
-		return err
+	binary := filepath.Join("/usr/lib/podman/", machine.ForwarderBinaryName)
+	if _, err := os.Stat(binary); os.IsNotExist(err) {
+		return errors.Errorf("unable to find %s", binary)
 	}
 	// Listen on all at port 7777 for setting up and tearing
 	// down forwarding


pkg/machine/qemu/machine.go
```diff
@@ -603,9 +603,9 @@ func CheckActiveVM() (bool, string, error) {
 // startHostNetworking runs a binary on the host system that allows users
 // to setup port forwarding to the podman virtual machine
 func (v *MachineVM) startHostNetworking() error {
-	binary, err := exec.LookPath(machine.ForwarderBinaryName)
-	if err != nil {
-		return err
+	binary := filepath.Join("/usr/libexec/podman/", machine.ForwarderBinaryName)
+	if _, err := os.Stat(binary); os.IsNotExist(err) {
+		return errors.Errorf("unable to find %s", binary)
```
rhatdan, line 608: Should we call back to the default, if the /usr/libexec/podman does not work?

I am not sure libexec is common on other distributions, specifically Ubuntu.

@cgwalters WDYT?

cgwalters: Yeah, the most portable is e.g. `/usr/lib/podman`.  If you try out e.g. `podman run --rm -ti docker.io/debian:stable` then e.g. `apt install systemd/stable` and `dpkg -L systemd` I notice that systemd's helper binaries are in `/lib/systemd`.   I forgot Debian stable hasn't even done UsrMove yet!  Amazing, like a glacier is amazing.

baude: we could call findexec, but given what we are trying to do here, it would seem to be shooting ourselves in the foot.  we can certainly try /usr/lib/podman/gvproxy as well.

rhatdan: Ok let's use /usr/lib/podman.

vrothberg: catatonit is placed in /usr/libexec/**podman**/catatonit.  We may reconsider moving them into the same directory.  Just want to make sure that we're not spreading our config files and binary helpers at all possible places.

rhatdan: I agree, we should move that also.  But not in this PR.

vrothberg: Thanks! Also note that the path to `catatonit` is configurable in containers.conf. openSUSE, for instance, sets it to `init_path = "/usr/bin/catatonit"`

rhatdan:
LGTM

PR #11101, opened 2021-08-02
Fix handling of user specified container labels

rhatdan:
Currently we override the SELinux labels specified by the user
if the container is runing a kata container or systemd container.

This PR fixes to use the label specified by the user.

Fixes: https://github.com/containers/podman/issues/11100

### DIFF
--- a/test/system/410-selinux.bats
+++ b/test/system/410-selinux.bats
@@ -50,6 +50,18 @@ function check_label() {
     check_label "--systemd=always" "container_init_t"
 }
 
+@test "podman selinux: init container with --security-opt type" {
+    check_label "--systemd=always --security-opt=label=type:spc_t" "spc_t"
+}
+
+@test "podman selinux: init container with --security-opt level&type" {
+    check_label "--systemd=always --security-opt=label=level:s0:c1,c2 --security-opt=label=type:spc_t" "spc_t" "s0:c1,c2"
+}
+
+@test "podman selinux: init container with --security-opt level" {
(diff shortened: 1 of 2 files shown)


test/system/410-selinux.bats
```diff
@@ -50,6 +50,14 @@ function check_label() {
     check_label "--systemd=always" "container_init_t"
 }
 
+@test "podman selinux: init container with --security-opt type" {
+    check_label "--systemd=always --security-opt=label=type:spc_t" "spc_t"
```
edsantiago, line 54: This test fails (root and rootless) on my laptop: gets `container_init_t` instead of `spc_t`.

test/system/410-selinux.bats
```diff
@@ -50,6 +50,14 @@ function check_label() {
...
+}
+
+@test "podman selinux: init container with --security-opt level" {
+    check_label "--systemd=always --security-opt=label=level:s0:c1,c2" "container_init_t"
```
edsantiago, line 58: This test passes, but it also passes in podman without your PR. (IOW, this is a test of functionality already present in podman. Which is fine by me, I like extra tests, but I wanted to make sure you weren't expecting this to be a test for your fix).

rhatdan: I want to make sure my changes do not change this behavior. IE I want to make sure we still get container_init_t if we change the level without changing the type.

edsantiago:
Tests LGTM now, thanks

vrothberg (commented):
/lgtm
/hold

